PR #1734, opened 2018-10-02
Adding a soft stale config so that we can enable probot/stale to keep…

admc:
… our backlog relevant

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,25 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 120
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned 
+  - area/security
+  - area/tls
+  - area/docs
+  - good first issue
+  - benchmarking
+  - help wanted
+  - epic
+  - rfc
+  - roadmap 
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,25 @@
...
+  - roadmap 
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 14 days if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: true
```
klingerf: From looking at the [probot source](https://github.com/probot/stale/blob/e285e96ff5f3bd6a10202f6179e61b5754740425/lib/schema.js#L31-L33), it appears that `closeComment` needs to be either a comment string or false. As written, I think this will close issues with the message "true". I'd recommend changing to `closeComment: false`.

.github/stale.yml
```diff
@@ -0,0 +1,25 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 120
```
klingerf: I vote to set this to `daysUntilStale: 90`. As far as I can tell we'll still mark the same 4 issues as stale initially at either 90 or 120 days, so might as well tighten it up somewhat:

https://github.com/linkerd/linkerd2/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+updated%3A%3C2018-06-03

https://github.com/linkerd/linkerd2/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+updated%3A%3C2018-07-03

(excluding the 2 `area/docs` issues that appear in those results)

PR #1749, opened 2018-10-10
Added support for json output in `linkerd stat`

alpeb:
Howdy, first commit into this project here! :-P 

I decided to have the json be an array of flat objects, repeating `namespace` in each object. Another option could have been to group things per namespace. Let me know what you think.
Also, for those cases where a value is not defined (represented with `-` in the traditional output) I'm using `null`.


grampelberg:
Finally, I've been wanting this forever!

alpeb:
No problemo. I've just added the json pretty-printing.
As for the latency units, they're always in milliseconds. I preferred to keep a numerical primitive in there, but I have no issue changing that to a "x`ms`" string if you deem that's what the output consumers would be expecting.

grampelberg:
I'm 100% behind it being numerical, only thought was to do it like:

```json
[
  {
    "namespace": "linkerd",
    "kind": "deployment",
    "name": "ca",
    "meshed": "1/1",
    "success": 1,
    "rps": 0.3,
    "latency_p50": {
        "unit": "ms",
        "value": 1
    },
    "latency_p95": 2,
    "latency_p99": 2,
    "tls": 0
  }
]
```

Looking at it, that's not great either. As long as the units are only ever ms, I'm good.

alpeb:
Yeah, and actually those fields are actually called `latency_ms_p50` for example elsewhere in the system. We can adopt the same naming here...

cli/cmd/stat.go
```diff
@@ -307,6 +327,79 @@ func printStatTable(stats map[string]*row, resourceType string, w *tabwriter.Wri
...
+
+// Using pointers there where the value is NA and the corresponding json is null
+type jsonStats struct {
+	Namespace   string   `json:"namespace"`
+	Kind        string   `json:"kind"`
+	Name        string   `json:"name"`
+	Meshed      string   `json:"meshed"`
+	Success     *float64 `json:"success"`
+	Rps         *float64 `json:"rps"`
+	Latency_p50 *uint64  `json:"latency_p50"`
```
siggy: agree re: your `latency_ms_p50` comment, i think it's best to include the unit somewhere.

cli/cmd/stat.go
```diff
@@ -307,6 +327,79 @@ func printStatTable(stats map[string]*row, resourceType string, w *tabwriter.Wri
...
+	Tls         *float64 `json:"tls"`
+}
+
+func printStatJson(statTables map[string]map[string]*row, resourceTypes []string, w *tabwriter.Writer) {
+	var entries []*jsonStats
+	for _, resourceType := range resourceTypes {
+		if stats, ok := statTables[resourceType]; ok {
+			sortedKeys := sortStatsKeys(stats)
+			for _, key := range sortedKeys {
+				resourceTypePrefix := resourceType
```
adleong: Given that there is a `kind` field, I don't think we need to prefix resource names by type.

cli/cmd/stat.go
```diff
@@ -307,6 +327,79 @@ func printStatTable(stats map[string]*row, resourceType string, w *tabwriter.Wri
...
+						name,
+						stats[key].meshed,
+						&stats[key].successRate,
+						&stats[key].requestRate,
+						&stats[key].latencyP50,
+						&stats[key].latencyP95,
+						&stats[key].latencyP99,
+						&stats[key].tlsPercent,
+					}
+				} else {
```
siggy: slightly less verbose and more explicit:
```golang
entry := &jsonStats{
	Namespace: namespace,
	Kind:      resourceType,
	Name:      name,
	Meshed:    stats[key].meshed,
}
if stats[key].rowStats != nil {
	entry.Success = &stats[key].successRate
	entry.Rps = &stats[key].requestRate
	entry.Latency_p50 = &stats[key].latencyP50
	entry.Latency_p95 = &stats[key].latencyP95
	entry.Latency_p99 = &stats[key].latencyP99
	entry.Tls = &stats[key].tlsPercent
}
```

alpeb:
@siggy I merged your `moar-stat-tests` branch here but it appears the following signature change is breaking some other tests:
https://github.com/linkerd/linkerd2/blob/d09b1e8ecbbc0e432f953e839cf0805317409aac/controller/api/public/test_helper.go#L130

alpeb:
Thanks @siggy . Agreed on those var name changes. I've pushed my branch again. And thanks everyone else for your input! Linkerd rules! :1st_place_medal:

PR #1750, opened 2018-10-10
Make room for columns in `linkerd top`

alpeb:
Not only there wasn't enough padding, but also column data were bleeding out into other columns.
I set the column width dynamically upon each table update, taking into account the row with the longest data so everything fits.

### DIFF
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -348,30 +348,44 @@ func renderHeaders(withSource bool) {
 	}
 }
 
+func max(i, j int) int {
+	if i > j {
+		return i
+	}
+	return j
+}
+
 func renderTableBody(table *[]tableRow, withSource bool) {
 	sort.SliceStable(*table, func(i, j int) bool {
 		return (*table)[i].count > (*table)[j].count
 	})
+	adjustedColumnWidths := columnWidths
+	for _, row := range *table {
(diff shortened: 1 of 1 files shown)


cli/cmd/top.go
```diff
@@ -348,30 +348,44 @@ func renderHeaders(withSource bool) {
...
+		adjustedColumnWidths[1] = max(adjustedColumnWidths[1], runewidth.StringWidth(row.destination))
+		adjustedColumnWidths[3] = max(adjustedColumnWidths[3], runewidth.StringWidth(row.by))
+
+	}
 	for i, row := range *table {
 		x := 0
 		if withSource {
-			tbprint(x, i+headerHeight, row.source)
-			x += columnWidths[0] + 1
+			tbprint(x, i+headerHeight, runewidth.Truncate(row.source, adjustedColumnWidths[0], ""))
```
adleong: If we are adjusting the column width to the maximum runewidth in that column, is this truncation needed?

alpeb:
Oops you're right. That was a leftover from a previous iteration. I've just removed it.

cli/cmd/top.go
alpeb: Oops you're right. That was a leftover from a previous iteration. I've just removed it.

alpeb:
@adleong Cool. I've forced-pushed that last commit again with the signoff :+1:

wmorgan:
Thanks @alpeb!

PR #1754, opened 2018-10-10
Update pinned proxy SHA to latest master

hawkw:
This PR updates the proxy SHA the build is pinned. This is in order to
track dependency updates in the proxy for the upcoming edge release.

### DIFF
--- a/bin/docker-build-proxy
+++ b/bin/docker-build-proxy
@@ -14,6 +14,6 @@ rootdir="$( cd $bindir/.. && pwd )"
 . $bindir/_tag.sh
 
 # Default to a pinned commit SHA of the proxy.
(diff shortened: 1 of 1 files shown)


bin/docker-build-proxy
```diff
@@ -14,6 +14,6 @@ rootdir="$( cd $bindir/.. && pwd )"
 . $bindir/_tag.sh
 
 # Default to a pinned commit SHA of the proxy.
-PROXY_VERSION="${PROXY_VERSION:-977ff25}"
+PROXY_VERSION="${PROXY_VERSION:- 51db6f8}"
```
klingerf: Thanks for bumping this! Looks like you've include an extra space character that shouldn't be there.

hawkw: Agh, thanks @klingerf!

Issue #1662, opened 2018-09-17
Improve web docker image build time


klingerf:
If feels like the web docker image builds pretty slowly at this point. We should investigate speeding it up. Am including some timing information and other observations from a recent build of that image.

```
Step 10/28 : COPY web/app $PACKAGEDIR
 ---> 614631acf28e
```

The docker image that we build has 28 layers. Any changes in the `web/app` directory invalidate layer 10, which requires rebuilding layers 11-28. We should see if we can invalidate fewer layers when the source code in `web/app` changes.

```
Step 13/28 : RUN $ROOT/bin/web setup --pure-lockfile
 ---> Running in f5028daf4a95
yarn install v1.7.0
[1/4] Resolving packages...
[2/4] Fetching packages...
(node:15) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
info fsevents@1.2.4: The platform "linux" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning "antd > rc-editor-mention@1.1.6" has unmet peer dependency "immutable@^3.7.4".
warning " > karma-webpack@2.0.6" has incorrect peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0".
warning "karma-webpack > webpack-dev-middleware@1.12.2" has incorrect peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0".
warning " > react-test-renderer@15.6.1" has incorrect peer dependency "react@^15.6.1".
warning " > babel-preset-react-app@3.1.1" has unmet peer dependency "babel-runtime@^6.23.0".
warning " > postcss-cssnext@3.1.0" has unmet peer dependency "caniuse-lite@^1.0.30000697".
[4/4] Building fresh packages...
Done in 62.81s.
Removing intermediate container f5028daf4a95
 ---> 0006fe7c387d
```

Running `yarn --pure-lockfile` takes more than 1 minute, and we run it every time any of the js or css in `web/app` changes. Maybe we can only run it when `yarn.lock` changes?

```
Step 15/28 : RUN $ROOT/bin/web build
 ---> Running in 0db83504e396
yarn run v1.7.0
$ /go/src/github.com/linkerd/linkerd2/web/app/node_modules/.bin/webpack
[...]

PR #1757, opened 2018-10-11
Improve docker layer caching for web image

klingerf:
Our web Dockerfile was previously setup to reinstall all yarn dependencies and rebuild the go binary any time any of our css or js files changed.

In this branch I've tweaked the layers in the Dockerfile so that yarn dependencies are only rebuilt when package.json or yarn.lock changes, and the go binary is only rebuilt when go source files change.

I also realized that our production web image included all of our unbundled javascript and css files, which aren't used when running the app in production, so I've removed them.

And in the process I change the file layout of the web image to match the defaults expected by the web server, which means we can remove the `-static-dir` and `-template-dir` flags from our install configs.

Testing locally, prior to make this change, making a modification to `web/app/js/index.js` would cause the docker build to take about 3m15s. After making this change, it rebuilds in about 1m5s.

This branch doesn't modify webpack, which accounts for the bulk of the remaining time required to build the web image (relates to #1662).

### DIFF
--- a/cli/cmd/testdata/install_default.golden
+++ b/cli/cmd/testdata/install_default.golden
@@ -348,8 +348,6 @@ spec:
       containers:
       - args:
         - -api-addr=api.linkerd.svc.cluster.local:8085
-        - -static-dir=/dist
-        - -template-dir=/templates
         - -uuid=deaab91a-f4ab-448a-b7d1-c832a2fa0a60
         - -controller-namespace=linkerd
         - -log-level=info
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -38,7 +38,9 @@ RUN CGO_ENABLED=0 GOOS=linux go build -o web/web ./web
 
 ## package it all up
 FROM gcr.io/linkerd-io/base:2017-10-30.01
-COPY LICENSE /linkerd/LICENSE
+WORKDIR /linkerd
+
+COPY LICENSE .
 COPY --from=golang /go/src/github.com/linkerd/linkerd2/web/web .
 RUN mkdir -p app
 COPY --from=webpack-bundle /go/src/github.com/linkerd/linkerd2/web/app/dist app/dist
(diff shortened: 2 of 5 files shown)


cli/cmd/testdata/install_default.golden
```diff
@@ -348,8 +348,6 @@ spec:
       containers:
       - args:
         - -api-addr=api.linkerd.svc.cluster.local:8085
-        - -static-dir=/dist
```
siggy: 👍

web/Dockerfile
```diff
@@ -35,9 +39,10 @@ RUN CGO_ENABLED=0 GOOS=linux go build -o web/web ./web
 ## package it all up
 FROM gcr.io/linkerd-io/base:2017-10-30.01
 COPY LICENSE /linkerd/LICENSE
-COPY --from=golang /go/src/github.com/linkerd/linkerd2/web .
-RUN mkdir -p ./dist
-COPY --from=webpack-bundle /go/src/github.com/linkerd/linkerd2/web/app/dist ./dist
+COPY --from=golang /go/src/github.com/linkerd/linkerd2/web/web .
```
siggy: tioli: what do you think about moving `web`, `app`, and `templates` to the `linkerd` top-level directory? that would make it clearer what is our stuff vs. OS stuff, and also avoid any possible top-level naming conflicts going forward.

```dockerfile
## package it all up
FROM gcr.io/linkerd-io/base:2017-10-30.01
WORKDIR /linkerd
COPY LICENSE /linkerd/LICENSE
COPY --from=golang /go/src/github.com/linkerd/linkerd2/web/web /linkerd/web
COPY --from=webpack-bundle /go/src/github.com/linkerd/linkerd2/web/app/dist /linkerd/app/dist
COPY web/templates /linkerd/templates
```

klingerf: This is a great idea! Will definitely do it as part of this branch. It was bugging me that we were dumping everything into the filesystem root.

PR #1752, opened 2018-10-10
Add generated Kubernetes client for ServiceProfile custom resource

adleong:
To support reading and writing of the ServiceProfile custom resource, we add a codegen'd Kubernetes client for this resource.

* Adding the ServiceProfile type and related boilerplate to /controller/gen/apis/serviceprofile. This boilerplate also contains directives that control how codegen works.
* A script in /hack which invokes codegen that generates Kubernetes client machinery for interacting with ServiceProfile resources. The majority of the generated code lives in /controller/gen/client.
* The above-mentioned generated code.

PR #1608, opened 2018-09-07
Implement getProfiles method in destination service

adleong:
We implement the getProfiles method in the destination service.  This method returns a stream of destination profiles for a given authority.  It does this by looking up the service for that authority and watching the `linkerd.io/service-profile` annotation on that service.  It then looks up the ServiceProfile custom resource with that name.  When either the annotation or the ServiceProfile resource changes, a new value is sent on the stream.

This PR includes:

* Adding a ServiceProfile Custom Resource Definition to `linkerd install`
* Adding the ServiceProfile type and related boilerplate to `/pkg/apis/serviceprofile`.  This boilerplate also contains directives that control how codegen works.
* A script in `/hack` which invokes codegen that generates Kubernetes client machinery for interacting with ServiceProfile resources.  The majority of the generated code lives in `/pkg/client`.
* The above-mentioned generated code.
* Updates to the Kubernetes utilility code in `/controller/k8s` to support interacting with ServiceProfiles.
* A watch based implementation of the getProfiles method in the destination service, similar to the implementation of get.
* An update to the destination client script that allows querying the getProfiles method.

PR #1758, opened 2018-10-11
Add ServiceProfile support to k8s utilities

adleong:
Updates to the Kubernetes utility code in `/controller/k8s` to support interacting with ServiceProfiles.

This makes use of the code generated client added in #1752


adleong:
Note to reviewers: this is a subset of the previously reviewed https://github.com/linkerd/linkerd2/pull/1608

controller/k8s/clientset.go
```diff
@@ -1,14 +1,15 @@
...
+	spclient "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	// Load all the auth plugins for the cloud providers.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
-func NewClientSet(kubeConfig string) (*kubernetes.Clientset, error) {
+func NewClientSet(kubeConfig string) (*kubernetes.Clientset, *spclient.Clientset, error) {
```
klingerf: Since most processes that need a kubernetes clientset are not also going to need an sp clientset, I'd prefer to have 2 separate methods in this file: `NewClientSet` and `NewSpClientSet`. You could create a separate helpful func for config parsing that's shared between the two methods.

controller/k8s/api_test.go
```diff
@@ -172,19 +170,23 @@ metadata:
...
+metadata:
+  name: namespace1`
+
+			ns2 := `
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: namespace2`
+
+			api, err := NewFakeAPI("namespace1", ns1, ns2)
```
klingerf: 👍 👍 👍 Much nicer!

controller/k8s/api.go
```diff
@@ -198,6 +225,8 @@ func (api *API) GetObjects(namespace, restype, name string) ([]runtime.Object, e
 		return api.getRCs(namespace, name)
 	case k8s.Service:
 		return api.getServices(namespace, name)
+	case k8s.ServiceProfile:
+		return api.getServiceProfiles(namespace, name)
```
klingerf: I don't think we need to add this to `GetObjects`, since `GetObjects` is only used right now for fetching resource targets used in CLI stat and tap commands. I'd remove this case, and remove the `getServiceProfiles` func that you're adding.

controller/cmd/tap/main.go
```diff
@@ -34,7 +34,8 @@ func main() {
 		restrictToNamespace = *controllerNamespace
 	}
 	k8sAPI := k8s.NewAPI(
-		clientSet,
+		k8sClient,
+		spClient,
```
klingerf: Right, and related to my comment about `NewSpClientSet`, let's just pass in nil for this argument in every `NewAPI` call, until we have a process that's actually going to watch and fetch SPs.

controller/k8s/clientset.go
```diff
@@ -9,6 +10,32 @@ import (
...
+	config, err := parseConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return clientset, nil
```
klingerf: There's no need for the nil check now -- you can just return the result of NewForConfig directly, since it's return signature is the same as this func.

```
	return kubernetes.NewForConfig(config)
```

controller/k8s/clientset.go
```diff
@@ -9,6 +10,32 @@ import (
...
+	config, err := parseConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	spclientset, err := spclient.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	return spclientset, nil
```
klingerf: And here:

```
	return spclient.NewForConfig(config)
```

PR #1759, opened 2018-10-12
Implement getProfiles method in destination service

adleong:
We implement the getProfiles method in the destination service. This method returns a stream of destination profiles for a given authority. It does this by looking up the ServiceProfile resource in the controller namespace named `<svc>.<ns>` where `<svc>` is the name of the service and `<ns>` is the namespace of the service.

This PR includes:
* Adding a ServiceProfile Custom Resource Definition to linkerd install
* A watch based implementation of the getProfiles method in the destination service, similar to the implementation of get.
* An update to the destination client script that allows querying the getProfiles method.


adleong:
Note to reviewers: this is a subset of the already reviewed https://github.com/linkerd/linkerd2/pull/1608

controller/destination/profile_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+
+	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
+	splisters "github.com/linkerd/linkerd2/controller/gen/client/listers/serviceprofile/v1alpha1"
+	"github.com/linkerd/linkerd2/controller/k8s"
+	log "github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/cache"
+)
+
+const profileAnnotation = "linkerd.io/service-profile"
```
klingerf: Looks like this const is unused. Remove it?

controller/util/http.go
```diff
@@ -0,0 +1,41 @@
...
+				Registered: httpPb.HttpMethod_Registered(value),
+			},
+		}
+	}
+	return &httpPb.HttpMethod{
+		Type: &httpPb.HttpMethod_Unregistered{
+			Unregistered: strings.ToUpper(method),
+		},
+	}
+}
```
klingerf: These two funcs are identical to the `parseScheme` and `parseMethod` funcs in `controller/tap/server.go`. Can you remove them from that file and update it to import this package and call these methods instead?

controller/destination/profile_watcher_test.go
```diff
@@ -0,0 +1,100 @@
...
+		service          profileId
+		expectedProfiles []*sp.ServiceProfileSpec
+	}{
+		{
+			name: "service profile",
+			k8sConfigs: []string{`
+apiVersion: linkerd.io/v1alpha1
+kind: ServiceProfile
+metadata:
+  name: foobar.ns
```
klingerf: Is this a valid name for a service profile? Should service profiles use the same validation that the k8s services use for the name? If I try to name a service "foobar.ns", I get:

```
The Service "foobar.ns" is invalid: metadata.name: Invalid value: "foobar.ns": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

klingerf: Ah, I was misunderstanding where these resources are supposed to live. @adleong explained that they all live in the control plane namespace, and the name reflects the name _and_ namespace of the service to which they apply. 👍

controller/destination/profile_watcher_test.go
```diff
@@ -0,0 +1,100 @@
...
+  namespace: linkerd
+spec:
+  routes:
+  - condition:
+      path: "/x/y/z"
+    responses:
+    - condition:
+        status:
+          min: 500
```
klingerf: I think this needs a `max` field in order to be considered valid.

Issue #1756, opened 2018-10-11
proxy-injector produces invalid patch in GKE 10.7 cluster


klingerf:
### What is the issue?

In a GKE 10.7 cluster, the admission controller that's installed when installing linkerd with the `--proxy-auto-inject` flag set produces a patch to a pod template spec that's invalid, causing the apply to fail, with:

```
Error from server (Invalid): error when creating "STDIN": Deployment.apps "emoji" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"linkerd.io/control-plane-ns":"linkerd", "linkerd.io/proxy-deployment":"emoji"}: `selector` does not match template `labels`
```

### How can it be reproduced?

This can be reproduced on a GKE 10.7 cluster, using the emojivoto app.

1. Install linkerd version `edge-18.10.2`:

```
linkerd install --tls optional --proxy-auto-inject | kubectl apply -f -
```

2. Wait for it to start

```
linkerd check
```

3. Install the emojivoto app, which fails with:

```
curl -s https://run.linkerd.io/emojivoto.yml | kubectl apply -f -
namespace "emojivoto" configured
service "emoji-svc" unchanged
service "voting-svc" unchanged
service "web-svc" unchanged
Error from server (Invalid): error when creating "STDIN": Deployment.apps "emoji" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"linkerd.io/control-plane-ns":"linkerd", "linkerd.io/proxy-deployment":"emoji"}: `selector` does not match template `labels`
Error from server (Invalid): error when creating "STDIN": Deployment.apps "voting" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"linkerd.io/control-plane-ns":"linkerd", "linkerd.io/proxy-deployment":"voting"}: `selector` does not match template `labels`
Error from server (Invalid): error when creating "STDIN": Deployment.apps "web" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"linkerd.io/control-plane-ns":"linkerd", "linkerd.io/proxy-deployment":"web"}: `selector` does not match template `labels`
Error from server (Invalid): error when creating "STDIN": Deployment.apps "vote-bot" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"linkerd.io/proxy-deployment":"vote-bot", "linkerd.io/control-plane-ns":"linkerd"}: `selector` does not match template `labels`
```

### Environment

- Kubernetes Version:

```
[...]

PR #1765, opened 2018-10-15
Proxy injector must preserve the original pod template labels and annotations

ihcsim:
This PR updates the JSON patch generated by the proxy injector admission webhook to ensure that:

* all original labels and annotations are preserved
* the mutated deployment's `spec.template.metadata.labels` matches its `spec.selector.matchLabels` which seems to be a requirement in newer version of Kubernetes.

It resolves issue #1756.

```
$ bin/linkerd version
Client version: git-e08642ee
Server version: unavailable

$ curl -s https://run.linkerd.io/emojivoto.yml | kubectl apply -f -
namespace "emojivoto" created
deployment.apps "emoji" created
service "emoji-svc" created
deployment.apps "voting" created
service "voting-svc" created
deployment.apps "web" created
service "web-svc" created
deployment.apps "vote-bot" created

$ kubectl -n emojivoto get deploy --show-labels
NAME       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE       LABELS
emoji      1         1         1            1           59s       app=emoji-svc,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=emoji
vote-bot   1         1         1            1           58s       app=vote-bot,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=vote-bot
voting     1         1         1            1           59s       app=voting-svc,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=voting
web        1         1         1            1           59s       app=web-svc,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=web

$ kubectl -n emojivoto get po --show-labels
NAME                        READY     STATUS    RESTARTS   AGE       LABELS
emoji-dc94dcd6-wqwtn        2/2       Running   0          1m        app=emoji-svc,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=emoji,pod-template-hash=87508782
vote-bot-55f47cd896-7hxqt   2/2       Running   0          1m        app=vote-bot,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=vote-bot,pod-template-hash=1190378452
voting-64479996d8-zjkv8     2/2       Running   0          1m        app=voting-svc,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=voting,pod-template-hash=2003555284
web-56669d856d-sb76s        2/2       Running   0          1m        app=web-svc,linkerd.io/control-plane-ns=linkerd,linkerd.io/proxy-deployment=web,pod-template-hash=1222584128
```


controller/proxy-injector/webhook.go
```diff
@@ -152,10 +152,13 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
 	patch.addVolume(caBundle)
 	patch.addVolume(tlsSecrets)
 
-	patch.addPodLabel(map[string]string{
-		k8sPkg.ControllerNSLabel:    w.controllerNamespace,
-		k8sPkg.ProxyDeploymentLabel: deployment.ObjectMeta.Name,
-	})
+	labels := map[string]string{}
```
adleong: Would it be simpler to just mutate the `deployment.Spec.Template.Labels` map rather than allocating a new one?

controller/proxy-injector/fake/data/inject-enabled-response.yaml
```diff
@@ -68,6 +68,6 @@ request:
...
...bGF0ZS9zcGVjL3ZvbHVtZXMiLCJ2YWx1ZSI6W119LHsib3AiOiJhZGQiLCJwYXRoIjoiL3NwZWMvdGVtcGxhdGUvc3BlYy92b2x1bWVzLzAiLCJ2YWx1ZSI6eyJuYW1lIjoibGlua2VyZC10cnVzdC1hbmNob3JzIiwiY29uZmlnTWFwIjp7Im5hbWUiOiJsaW5rZXJkLWNhLWJ1bmRsZSIsIm9wdGlvbmFsIjp0cnVlfX19LHsib3AiOiJhZGQiLCJwYXRoIjoiL3NwZWMvdGVtcGxhdGUvc3BlYy92b2x1bWVzLzAiLCJ2YWx1ZSI6eyJuYW1lIjoibGlua2VyZC1zZWNyZXRzIiwic2VjcmV0Ijp7InNlY3JldE5hbWUiOiJuZ2lueC1kZXBsb3ltZW50LXRscy1saW5rZXJkLWlvIiwib3B0aW9uYWwiOnRydWV9fX0seyJvcCI6ImFkZCIsInBhdGgiOiIvc3BlYy90ZW1wbGF0ZS9tZXRhZGF0YS9sYWJlbHMiLCJ2YWx1ZSI6eyJhcHAiOiJuZ2lueCIsImxpbmtlcmQuaW8vYXV0by1pbmplY3QiOiJlbmFibGVkIiwibGlua2VyZC5pby9jb250cm9sLXBsYW5lLW5zIjoibGlua2VyZCIsImxpbmtlcmQuaW8vcHJveHktZGVwbG95bWVudCI6Im5naW54In19LHsib3AiOiJhZGQiLCJwYXRoIjoiL3NwZWMvc2VsZWN0b3IvbWF0Y2hMYWJlbHMiLCJ2YWx1ZSI6eyJhcHAiOiJuZ2lueCIsImxpbmtlcmQuaW8vYXV0by1pbmplY3QiOiJlbmFibGVkIiwibGlua2VyZC5pby9jb250cm9sLXBsYW5lLW5zIjoibGlua2VyZCIsImxpbmtlcmQuaW8vcHJveHktZGVwbG95bWVudCI6Im5naW54In19LHsib3AiOiJhZGQiLCJwYXRoIjoiL3NwZWMvdGVtcGxhdGUvbWV0YWRhdGEvYW5ub3RhdGlvbnMiLCJ2YWx1ZSI6eyJjcmVhdGVkLWJ5IjoiaXNpbSIsImxpbmtlcmQuaW8vY3JlYXRlZC1ieSI6ImxpbmtlcmQvcHJveHktaW5qZWN0b3IgdjE4LjguNCIsImxpbmtlcmQuaW8vcHJveHktdmVyc2lvbiI6InYxOC44LjQifX1d
```
adleong: Does the decoded version of this live in the source tree somewhere?  How is this generated?

controller/proxy-injector/webhook.go
adleong: I just mean mutating the `deployment.Spec.Template.Labels` map in memory and then calling `patch.addPodLabels(deployment.Spec.Template.Labels)` rather than allocating a new map.  I figured this might be simpler and more direct.  
But what you have here is fine too, I was just curious.

controller/proxy-injector/fake/data/inject-enabled-response.yaml
ihcsim: It's currently generated manually, where I just encode the [log output](https://github.com/linkerd/linkerd2/blob/2534462e77889e4a5e16133820e6dfd45144dee6/controller/proxy-injector/webhook.go#L85) from `go test`, then past it here. Open to feedback on better ways to manage this.

controller/proxy-injector/webhook.go
ihcsim: Ah.. gotcha. Yes, you are right 👍 It's ok to mutate `deployment` here. If this PR isn't merged after I get off work today, I can change it 😄

controller/proxy-injector/patch.go
```diff
@@ -72,6 +73,12 @@ func (p *Patch) addPodLabel(label map[string]string) {
 		Path:  patchPathPodLabel,
 		Value: label,
 	})
+
+	p.patchOps = append(p.patchOps, &patchOp{
+		Op:    "add",
+		Path:  patchPathSelectorMatchLabels,
+		Value: label,
+	})
```
klingerf: I'm not sure that we should modify the spec's selector labels. I tried reverting this change, and your branch still fixes the issue that I encountered on GKE. Plus if we do decide to start modifying selector labels, I'd want to make sure that we also update the `linkerd inject` implementation to match.

And those issues aside, based on my reading of kubernetes/kubernetes#50719, I'm pretty sure the right behavior is for us to treat the selector labels as immutable. How would you feel about reverting this change?

ihcsim: hmmm... that's interesting. I thought the error is caused by a mismatch between the selector labels and the pod template labels. The k8s docs say,

> Label selector for pods. Existing ReplicaSets whose pods are selected by this will be the ones affected by this deployment. **It must match the pod template's labels.**

I'll need to do some more testing on this.

ihcsim: I did some more testing on this by changing the `apiVersion` of the emojivoto deployments to `apps/v1`. It looks like this change isn't required, as you pointed out. Agree with treating the selector labels as immutable for all the reasons pointed out in kubernetes/kubernetes#50808. I have reverted it.

controller/proxy-injector/fake/data/inject-enabled-response.yaml
adleong: Maybe just a comment in this file about how to generate this?  The comment should get discarded when this is unmarshalled by the test, right?

Issue #1225, opened 2018-06-28
Add Grafana dashboard for authorities


rmars:
As of #1181 we can query statsummaries for authorities.
We can now add a grafana dashboard for the authority stats, similar to how we do for deployments, replication controllers, etc.

We should also link to the Grafana dashboard from the Web UI tables.

alpeb:
Thanks for the clarification @rmars . I've created #1772 accordingly. Please let me know if it makes sense :-)

rmars:
Fixed by #1772

PR #1772, opened 2018-10-17
Add Grafana dashboard for Authorities

alpeb:
Proposal for #1225

Signed-off-by: Alejandro Pedraza <alejandro.pedraza@gmail.com>

### DIFF
--- a/grafana/dashboards/authority.json
+++ b/grafana/dashboards/authority.json
@@ -1,0 +1,1237 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 1,
+  "id": 10,
+  "iteration": 1539744617561,
+  "links": [],
+  "panels": [
+    {
+      "content": "<div style=\"display: flex; align-items: center\">\n  <img src=\"https://linkerd.io/images/identity/favicon/linkerd-favicon.png\" style=\"height:32px;\"/>&nbsp;\n  <span style=\"font-size: 32px\">au/$authority</span>\n</div>",
+      "gridPos": {
+        "h": 2.4,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 2,
+      "links": [],
+      "mode": "html",
+      "title": "",
+      "transparent": true,
+      "type": "text"
+    },
+    {
+      "cacheTimeout": null,
+      "colorBackground": false,
+      "colorValue": false,
+      "colors": [
+        "#d44a3a",
+        "rgba(237, 129, 40, 0.89)",
+        "#299c46"
+      ],
+      "datasource": "prometheus",
+      "decimals": null,
+      "format": "percentunit",
+      "gauge": {
+        "maxValue": 1,
+        "minValue": 0,
+        "show": true,
+        "thresholdLabels": false,
+        "thresholdMarkers": true
+      },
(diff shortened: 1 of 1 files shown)


grampelberg:
Awesome!

grafana/dashboards/authority.json
```diff
@@ -0,0 +1,1237 @@
...
+      "targets": [
+        {
+          "expr": "sum(irate(request_total{namespace=\"$namespace\", authority=\"$authority\", direction=\"inbound\", tls=\"true\"}[30s])) by (authority, deployment)",
+          "format": "time_series",
+          "intervalFactor": 1,
+          "legendFormat": "🔒au/{{authority}}",
+          "refId": "A"
+        },
+        {
+          "expr": "sum(irate(request_total{namespace=\"$namespace\", authority=\"$authority\", direction=\"inbound\", tls!=\"true\"}[30s])) by (authority, deployment)",
```
klingerf: Authorities are a bit of a special case for our stats, since there's no notion of a `dst_authority` label -- the `authority` label is both the source and the destination, in some sense.

That said, in this row, I think we want to display stats for all deployments that are sending traffic to the authority that's stored in the `$authority` variable. The only way to query for that information is to use outbound stats, rather than inbound stats, and group by "deployment" instead of authority. Something like:

```
sum(irate(request_total{namespace="$namespace", authority="$authority", direction="outbound", tls!="true"}[30s])) by (deployment)
```

That will produce a list of deployments that are sending traffic to the authority that is being displayed in this dashboard. We'd want to make the same modification for all charts in this row.

grafana/dashboards/authority.json
```diff
@@ -0,0 +1,1237 @@
...
+          "format": "time_series",
+          "intervalFactor": 1,
+          "legendFormat": "au/{{authority}}",
```
klingerf: Related to my comment above, once we start grouping by outbound deployment, we should modify this label to display the deployment instead of the authority: "deploy/{{deployment}}".

grafana/dashboards/authority.json
```diff
@@ -0,0 +1,1237 @@
...
+      "targets": [
+        {
+          "expr": "sum(irate(request_total{namespace=\"$namespace\", authority=\"$authority\", direction=\"inbound\", tls=\"true\"}[30s])) by (authority, pod)",
+          "format": "time_series",
+          "intervalFactor": 1,
+          "legendFormat": "🔒po/{{pod}}",
+          "refId": "A"
+        },
+        {
+          "expr": "sum(irate(request_total{namespace=\"$namespace\", authority=\"$authority\", direction=\"inbound\", tls!=\"true\"}[30s])) by (authority, pod)",
```
klingerf: Likewise for pods, we'd also need to use upstream stats to find all pods that are sending traffic to this authority:

```
sum(irate(request_total{namespace="$namespace", authority="$authority", direction="outbound", tls!="true"}[30s])) by (pod)
```

And we'd want to make the same modification for all charts in this row.

grafana/dashboards/authority.json
```diff
@@ -0,0 +1,1237 @@
...
+        "tagsQuery": "",
+        "type": "query",
+        "useTags": false
+      },
+      {
+        "allValue": null,
+        "current": {
+          "text": "forum.topmachina.com",
+          "value": "forum.topmachina.com"
+        },
```
klingerf: This is unfortunately an annoying part of exporting Grafana dashboard -- the export includes the currently selected dropdown value for all template variables. Since most folks won't have this authority in their installations, we need to manually remove them:

```
        "current": {},
```

grafana/dashboards/authority.json
```diff
@@ -0,0 +1,1237 @@
...
+    "linkerd"
+  ],
+  "templating": {
+    "list": [
+      {
+        "allValue": null,
+        "current": {
+          "text": "default",
+          "value": "default"
+        },
```
klingerf: Same here:

```
        "current": {},
```

PR #1773, opened 2018-10-17
Add linkerd profile --template command

adleong:
Add a new CLI command: `linkerd profile --template` which outputs a sample service profile yaml.  Users can edit this sample and then `kubectl apply` it to add a service profile.  The sample serves as "documentation by example" of what service profiles may contain.

Example usage:
```bash
linkerd profile --template web-svc.emojivoto > web-svc-profile.yaml
# edit web-svc-profile.yaml in your favorite editor
kubectl apply -f web-svc-profile.yaml
```

### DIFF
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -2,10 +2,9 @@ package cmd
 
 import (
 	"bytes"
-	"fmt"
+	"errors"
 	"io"
 	"os"
-	"strings"
 	"text/template"
 
 	"github.com/linkerd/linkerd2/cli/profile"
@@ -18,10 +17,21 @@ type templateConfig struct {
 	ServiceName           string
 }
 
+type profileOptions struct {
+	namespace string
+	template  bool
+}
+
+func newProfileOptions() *profileOptions {
+	return &profileOptions{
+		namespace: "default",
+		template:  false,
+	}
+}
+
 func newCmdProfile() *cobra.Command {
 
-	var namespace string
-	var template bool
+	options := newProfileOptions()
 
 	cmd := &cobra.Command{
 		Use:   "profile [flags] --template (SERVICE)",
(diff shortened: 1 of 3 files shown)


adleong:
@grampelberg totally.  This all started while I was writing docs for service profiles and it started to feel like having an editable template to point to would be helpful.  The main tricky bit that the template helps with is getting the name and namespace just right.  The service profile resource needs to be in the controller namespace but must be named `<svc-name>.<svc-namespace>`.

`--template` doesn't do anything yet, but I'm imagining that there will be different ways to generate service profiles in the future:

```
linkerd profile --template
linkerd profile --swagger=myapi.swagger
linkerd profile --proto=myapi.proto
linkerd profile --interactive
```

grampelberg:
Ahhh, that explains a lot, thank you!

Is it a hard requirement to have the profile in the *controller* namespace? That seems like an odd decision.

adleong:
Not a hard requirement but it's the way things are currently implemented.

This was intended to simplify things when one service is calling another across namespaces.  For example, if A is calling B, where should we look for B's service profile?  If we look in A's namespace, then we potentially need to duplicate B's service profile for each calling namespace.  If we look in B's namespace then there is no way for A to create a service profile on B if A doesn't have write access to B's namespace.  If we look in both, then we need some precedence rules for deciding which to use.

By putting the service profile in the controller namespace, we sidestep the issue.

grampelberg:
Hmm, I'll ask folks some questions and installs. As a service owner, it is likely you also own the controller namespace. As a platform owner, it is unlikely any service owners have access. 

Purely from a k8s resource perspective, I would assume that it should be in the same namespace as the service. Obviously. that only makes sense when it is a profile for *my* service. Everything falls down a little bit when a profile is used in anger. We've talked about using "external" services before, maybe that's the solution (I'm thinking about stuff like having a profile for `github.com` as well).

Feels like the right decision for now, lemme play with it a little bit and get some feedback from other folks to see if we should change it around to something else entirely.

adleong:
:+1::+1::+1:

cli/cmd/profile.go
```diff
@@ -0,0 +1,94 @@
...
+
+func buildConfig(namespace, service string) (templateConfig, error) {
+
+	serviceParts := strings.Split(service, ".")
+
+	serviceName := serviceParts[0]
+	serviceNamespace := "default"
+
+	if len(serviceParts) >= 2 && namespace != "" && serviceParts[1] != namespace {
+		return templateConfig{}, fmt.Errorf("service namespace cannot be specified by namespace flag (%s) and by service name (%s)", namespace, serviceParts[1])
```
klingerf: Ahh, regarding my comment about the example above -- my preference is to only support namespaces provided via the `--namespace` flag. That will be more consistent with how we treat stat and tap targets. I'd just simplify this to:

```diff
diff --git a/cli/cmd/profile.go b/cli/cmd/profile.go
index 93244a30..ca2cad37 100644
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -2,10 +2,8 @@ package cmd
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"os"
-	"strings"
 	"text/template"
 
 	"github.com/linkerd/linkerd2/cli/profile"
@@ -39,11 +37,7 @@ func newCmdProfile() *cobra.Command {
 		`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			config, err := buildConfig(namespace, args[0])
-			if err != nil {
-				return err
-			}
-			return renderProfileTemplate(config, os.Stdout)
+			return renderProfileTemplate(buildConfig(namespace, args[0]), os.Stdout)
 		},
 	}
 
@@ -53,32 +47,15 @@ func newCmdProfile() *cobra.Command {
 	return cmd
 }
 
-func buildConfig(namespace, service string) (templateConfig, error) {
-
-	serviceParts := strings.Split(service, ".")
-
-	serviceName := serviceParts[0]
-	serviceNamespace := "default"
-
-	if len(serviceParts) >= 2 && namespace != "" && serviceParts[1] != namespace {
-		return templateConfig{}, fmt.Errorf("service namespace cannot be specified by namespace flag (%s) and by service name (%s)", namespace, serviceParts[1])
-	}
-
-	if len(serviceParts) >= 2 {
-		serviceNamespace = serviceParts[1]
-	}
-
-	if namespace != "" {
-		serviceNamespace = namespace
+func buildConfig(namespace, service string) *templateConfig {
+	return &templateConfig{
+		ControlPlaneNamespace: controlPlaneNamespace,
+		ServiceNamespace:      namespace,
+		ServiceName:           service,
 	}
-	return templateConfig{
-		controlPlaneNamespace,
-		serviceNamespace,
-		serviceName,
-	}, nil
 }
 
-func renderProfileTemplate(config templateConfig, w io.Writer) error {
+func renderProfileTemplate(config *templateConfig, w io.Writer) error {
 	template, err := template.New("profile").Parse(profile.Template)
 	if err != nil {
 		return err
```

(note that the diff also modifies this method to a return a pointer to the template struct, rather than the struct itself)

cli/profile/template.go
```diff
@@ -0,0 +1,70 @@
+package profile
+
+// Template provides the base template for the `linkerd profile --template` command.
+const Template = `### Service Profile for {{.ServiceName}}.{{.ServiceNamespace}} ###
```
klingerf: TIOLI -- "ServiceProfile" instead of "Service Profile"?

cli/profile/template.go
```diff
@@ -0,0 +1,70 @@
...
+  - name: '/authors/{id}'
+
+    # Each route must define a condition.  All requests that match the
+    # condition will be counted as belonging to that route.  If a request
+    # matches more than one route, the first match wins.
+    condition:
+      # The simplest condition is a path regular expression.
+      path: '/authors/\d+'
+
+      # This is a condition that checks that request method.
```
klingerf: Hmm, maybe "...that checks _the_ request method."?

cli/cmd/profile.go
adleong: just for my education, why is returning a pointer to the struct preferred over returning the struct directly?

Issue #1669, opened 2018-09-17
Figure out interaction between sidebar state and main page


rmars:
Should the main page reflect the namespace selected in the sidebar?

For e.g. if a user selects a namespace and then goes to /pods, should the pod list be filtered to that namespace? 

Related: #1542

dadjeibaah:
It looks like if we want this to work, we may need to do some work in the dashboard URL scheme to have some state into what namespace a user is working in and therefore dictate what the dashboard displays in the sidebar and in the main page. Each component in the dashboard would then have the ability to filter its information based on the current namespace in the URL. The URL scheme could look like this:

`/namespaces/_all`
The  Overview page, Tap, Top pages and all Resource List pages would display metrics from all namespaces. We us the `_all` url segment to represent all namespaces to avoid clashing with a namespace `all`.

`/namespaces/<namespace>/<resourceType>`
This would filter the Overview page to only show <namespace>, Tap and Top pages would have the namespace field populated with <namespace> and all resource lists would only show metrics from <namespace>

`/namespaces/<namespace>/<resourceType>/<resourceName>`
This would show the resource under that namespace. There was the idea that if we click on a resource name that is from a different namespace or if the sidebar namespace is changed to a different namespace, we would notify a user that they would no longer be working in <namespace> with a modal. This seems jarring and we may want a better way to handle this. We may not even need to notify the user that that are changing namespaces and take them straight to the overview page.

grampelberg:
> We may not even need to notify the user that that are changing namespaces and take them straight to the overview page.

That's my vote. If you change the namespace selector it just bumps you out to the overview page.

PR #1637, opened 2018-09-12
[RFC] Bikeshed the tap and top icons

adleong:
I find the tap and top icons a bit strange.  Using the filter icon for tap is weird because we already use the filter icon for filtering columns.  The caret-up icon looks weird to me for top because it looks like something that is click to expand.

Unfortunately, the alternatives aren't super great either.  Here are some possibilities I could think of:

For tap:
* `eye` (although this sometimes indicates show/hide... and it also just looks weird)
* `phone`
* `shower` (too whimsical?)

For top:
* `table` (looks like a calendar)
* `list-ol`
* `arrow-up`

Thoughts?

Signed-off-by: Alex Leong <alex@buoyant.io>

PR #1500, opened 2018-08-21
Add kubernetes style sidebar

dadjeibaah:
Linkerd CLI's "look and feel" is similar to Kubernetes `kubectl` CLI. Linkerd's dashboard can be extended to match Kubernetes dashboard UI.

This PR serves as a starting point for this work. The new sidebar shows all resources from all namespaces on initial page load. Resources can be filtered to show only items in a given namespace.  The sidebar displays authority, deployment, service and, pod resources. We may need to think about whether it is necessary to show all resources types. Some resources, i.e. authorities, contain a large cardinality of resource details and may not be very useful to a user.

screen shot 2018-08-23 at 5 29 12 pm

fixes #1449

PR #1776, opened 2018-10-18
Move the dashboard's component library from antd to material-ui

### LINKED ISSUE #1408: Switch web component library from Ant to Material-UI
This will be a tracking ticket for the move. 

Switch the component pattern library we use in the frontend from Ant to [Material-UI](https://github.com/mui-org/material-ui).

All of our current components and layouts are based on ant. We'll need to find the equivalents of those in material, and resolve any feature discrepancies between the two. From my looking at it thus far, 

Components that need to be switched over, in rough order of effort:
- Table - Sorting seems to be there, but I think we'll need to implement our own filtering https://material-ui.com/api/table/#table
- Layout
  - Grid layout  (Row, Col) - move to https://material-ui.com/layout/grid/
  - Sider - https://material-ui.com/demos/drawers/#drawer
  - Nav bar - move to https://material-ui.com/demos/app-bar/#app-bar
- Menu - https://material-ui.com/demos/menus/
- Form components  - https://material-ui.com/demos/selection-controls/#checkboxes
  - Switch
  - Radio
- Modal - https://material-ui.com/utils/modal/
- Tooltip - https://material-ui.com/demos/tooltips/
- Spin (spinner) - https://material-ui.com/demos/progress/#progress
- Button - https://material-ui.com/api/button
- Icon - https://material-ui.com/style/icons/

rmars:
Switch the component library from [antd](https://ant.design/) to [material-ui](https://material-ui.com/).

This branch changes all uses of antd components to their closest equivalent in material. There is still a lot of polish that needs to go into the look of individual components, but since the major component rewrites are done, I think it's time to put tis up for review and get it in so that further work can be done in smaller branches. 

## Changes in this branch
- replace all uses of antd with material-ui components
- slight modifications of eslint rules
- restructuring of app components to be rendered under the Navigation component
- deleted most of our css (replaced with material's inline styles)
- pinned package versions in `package.json` (mostly removing `^`)

## Follow-up work
Here is a list of known UI polish in the current branch (I will ticket each one, as well as the results of code-review of this branch, to be done in follow up branches):
- Navigation/Sidebar #1781
  - I removed the list of resources per namespace and the namespace dropdown in the sidebar added in #1500 for simplicity. we should re-add it. I restructured the Navigation to be part of a larger component which renders the page as a subcomponent, which should make future work on #1669 easier
  - Side navigation doesn’t go all the way down on the Overview page
  - The icons for the side navigation were chosen without a lot of thought. Revisit this, as done in #1637. Resources that can help: https://material.io/tools/icons/?icon=home&style=baseline, https://materialdesignicons.com/. You can also use FontAwesome: (https://material-ui.com/style/icons/#font-icons)
- Theming and colours (#1782): for the charts and app-bar I did a very cursory colour selection. 
  - We should review the colour choices and pick a coherent set  (https://material.io/tools/color should help)
  - The theming in `index.js` should be moved into its own file as we expand it
- Tables (#1783) throughout the app need to be revisited and neatened
  - fix table widths, currently the tables are too wide, the columns are too narrow
  - add sorting to the table columns
  - add filtering to the tables
- Popover (#1784): popover on src dst column in the top tables goes away too fast for you to click anything on it, clicking on the links doesn’t work
- Octopus graph (#1785)
  - unmeshed resources and collapsed resources need to be full column width
  - octopus arms need arrow alignment to new card widths
- TapQueryForm: need form validation
- Tooltips: add arrows to tooltips
-  Tooling: upgrade to babel7 (#1786)

Fixes #1408

### DIFF
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -201,13 +201,13 @@ class NavigationBase extends React.Component {
       <div className={classes.root}>
         <AppBar
           position="absolute"
-          className={classNames(classes.appBar, this.state.drawerOpen && classes.appBarShift)}>
+          className={classNames(classes.appBar, {[classes.appBarShift]: this.state.drawerOpen} )}>
           <Toolbar disableGutters={!this.state.drawerOpen}>
             <IconButton
               color="inherit"
               aria-label="Open drawer"
               onClick={this.handleDrawerOpen}
-              className={classNames(classes.menuButton, this.state.drawerOpen && classes.hide)}>
+              className={classNames(classes.menuButton, {[classes.hide]: this.state.drawerOpen} )}>
               <MenuIcon />
             </IconButton>
 
@@ -220,7 +220,7 @@ class NavigationBase extends React.Component {
         <Drawer
           variant="permanent"
           classes={{
-            paper: classNames(classes.drawerPaper, !this.state.drawerOpen && classes.drawerPaperClose),
+            paper: classNames(classes.drawerPaper, {[classes.drawerPaperClose]: !this.state.drawerOpen} ),
           }}
           open={this.state.drawerOpen}>
           <div className={classNames(classes.linkerdLogoContainer, classes.toolbar)}>


web/app/js/components/ExpandableTable.jsx
```diff
@@ -0,0 +1,138 @@
...
+                        numeric={c.isNumeric}>
+                        {c.render(d)}
+                      </TableCell>
+                    ))
+                  }
+                </TableRow>
+              </React.Fragment>
+
+            );
+          })}
```
siggy: nit: indent

web/app/js/components/util/Progress.jsx
```diff
@@ -0,0 +1,33 @@
+import LinearProgress from '@material-ui/core/LinearProgress';
+import { withStyles } from '@material-ui/core/styles';
+
+const colorLookup = {
+  good: {
+    colorPrimary: '#c8e6c9', // background bar color (lighter)
```
siggy: per https://material-ui.com/style/color/, `#c8e6c9` is documented as `green[100]`:

```js
import red from '@material-ui/core/colors/green';
...
colorPrimary: green[100], // background bar color (lighter)
```

is it advisable/appropriate to reference them this way?

web/app/js/components/Navigation.jsx
```diff
@@ -0,0 +1,329 @@
...
+    );
+  }
+  render() {
+    const { classes, ChildComponent } = this.props;
+
+    return (
+      <div className={classes.root}>
+        <AppBar
+          position="absolute"
+          className={classNames(classes.appBar, this.state.drawerOpen && classes.appBarShift)}>
```
dadjeibaah: More of a readability nit: I am not entirely sure how `classNames` interprets statements like `this.state.drawerOpen && classes.appBarShift` but I found it a little difficult to comprehend. I tried rewriting the statement like so:
```jsx
className={classNames(classes.appBar, {[`${classes.appBarShift}`]: this.state.drawerOpen})}>
```
And that seemed to work the same way.

web/app/js/components/Navigation.jsx
```diff
@@ -0,0 +1,329 @@
...
+      </MenuItem>
+    );
+  }
+  render() {
+    const { classes, ChildComponent } = this.props;
+
+    return (
+      <div className={classes.root}>
+        <AppBar
+          position="absolute"
```
dadjeibaah: Could this line be moved to the `appBar` style object?

rmars: This is part of `AppBar`'s API:
https://material-ui.com/api/app-bar/#appbar

web/app/js/components/Navigation.jsx
rmars: Hahahah, what that line is actually doing is a combination of cool es6 tricks:

First, it's using the `&&` operand to return the string name of the class if `drawerOpen` is true.
(The js `&&` operand will return the latter of the provided values if the first expression is truthy). Example:
```
>> false && "foo"
>> false

>> true && "foo"
>> "foo"
```

Then, it's using the fancy object assignment to assign `appBarShift`'s className to the object if provided:
```
>> let a = "a_string";
>> let obj = { a }

>> obj
{a: "a_string"}
```

So this is equivalent to providing an object whose key is `classes.appBarShift` and value is `this.state.drawerOpen` as you've written. I'll rewrite it to be less trick-y!

dadjeibaah: Oh my gosh! so much sorcery! That is pretty cool. it's a TIOLI so feel free to do whatever you think is best.

rmars: Hahaha, yeah I think readable > cool, so I'll change them to more traditional args. 
It's a pattern I see popping up a lot in the material examples, so maybe we can change later.

web/app/js/components/BaseTable.jsx
```diff
@@ -1,31 +1,78 @@
...
-// https://github.com/ant-design/ant-design/blob/master/components/table/Table.tsx#L348
-export default class BaseTable extends Table {
-  constructor(props) {
-    super(props);
-    Table.prototype.toggleSortOrder = this.toggleSortOrder;
-  }
+const styles = theme => ({
+  root: {
+    width: '100%',
+    marginTop: theme.spacing.unit * 3,
```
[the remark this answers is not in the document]
rmars: done!

PR #1777, opened 2018-10-18
Use target for dataplane checks

### LINKED ISSUE #1513: Update linkerd check to accept a target for data plane checks
Once the ability to run data plane checks exists (#1512), it would be nice for check to accept a target kubernetes resource to run checks against, rather than running against all proxies that are configured to talk to the control plane. For instance:

```
linkerd check --proxy deploy/web -n emojivoto
```

Would check all proxies on all pods in the web deployment in the emojivoto namespace.

alenkacz:
When performing data plane checks make it possible to specify the target for those checks.

Fixes #1513


klingerf:
@grampelberg Are you picturing both a `--selector` flag and the ability to specify an owner target directly, or just the `--selector` flag? Is it possible to use a selector to filter by pod owners, in the absence of user-defined labels?

The list of data plane pods that we check is fetched from the public API's `ListPods` endpoint. I agree that doing the filtering server side might be a bit cleaner, and could be useful in other contexts where we call `ListPods`. We could modify the `ListPodsRequest` message as follows:

```
message ListPodsRequest {
  string namespace = 1; // DEPRECATED
  ResourceSelection selector = 2;
}
```

Where `ResourceSelection` is defined as:

```
message ResourceSelection {
  // Identifies a Kubernetes resource.
  Resource resource = 1;

  // A string-formatted Kubernetes label selector as passed to `kubectl get
  // --selector`.
  //
  // XXX in the future this may be superceded by a data structure that more
  // richly describes a parsed label selector.
  string label_selector = 2;
}
```

If we do that, I think we should deprecate the "namespace" field, in favor of the `ResourceSelection.Resource` field, which includes a namespace. One big benefit of this change is that it would make the `ListPods` API behave a lot more similarly to how resource selection works in the `Tap` and `StatSummary` APIs.

@alenkacz Sorry for the back and forth here. If it works for you, I could make the proposed modifications to `ListPods`, and then once that merges, you could use the updated `ListPods` functionality in this branch, to add the owner target and `--selector` flag. You're also welcome to make the `ListPods` change yourself, but I know you've already done one refactor in service of this change, and I don't want to hold you up even more.

grampelberg:
@klingerf I'd originally pictured only having `--selector`. As I went and look at `kubectl` more, it feels like having both is still a good idea.

From a UX perspective, I think this is pretty solid as is and my selector can be handled as a follow up.

alenkacz:
@klingerf that's ok, that happens :) I am in no rush landing this. I am fine with updating `ListPods` in another PR so I'll follow up and prepare the ListPods change.

Thanks for the feedback!

cli/cmd/check.go
```diff
@@ -62,11 +65,15 @@ non-zero exit code.`,
...
 	cmd.PersistentFlags().BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
-	cmd.PersistentFlags().BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
+	cmd.PersistentFlags().StringVar(&options.targetProxyResource, "proxy", options.targetProxyResource, "Only run data-plane checks, to determine if the data plane is healthy")
 	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for all tests to pass")
 	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	cmd.PersistentFlags().BoolVar(&options.singleNamespace, "single-namespace", options.singleNamespace, "When running pre-installation checks (--pre), only check the permissions required to operate the control plane in a single namespace")
 
+	if cmd.Flags().Changed("proxy") {
```
dadjeibaah: This unfortunately will not ever be true because the check is run before `flag.Parse()` is executed. and `cmd.Flags().Changed` documentation describes the function as:
```
// Changed returns true if the flag was explicitly set during Parse() and false
// otherwise
```
It looks like moving this to `validate()` would have the same effect.
```diff
diff --git a/cli/cmd/check.go b/cli/cmd/check.go
index ad1c8a17..305967ef 100644
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -137,6 +137,10 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 }
 
 func (o *checkOptions) validate() error {
+	if o.targetProxyResource != "" {
+		o.dataPlaneOnly = true
+	}
+
 	if o.preInstallOnly && o.dataPlaneOnly {
 		return errors.New("--pre and --proxy flags are mutually exclusive")
 	}

```

pkg/healthcheck/healthcheck.go
```diff
@@ -919,7 +937,7 @@ func validateDataPlanePods(pods []*pb.Pod, targetNamespace string) error {
 	return nil
 }
 
-func validateDataPlanePodReporting(pods []*pb.Pod) error {
+func validateDataPlanePodReporting(pods []*pb.Pod, proxyTarget string) error {
```
dadjeibaah: Looks like the `proxyTarget` variable is unused. Would this be used to check for data plane reporting for a specific proxy Target?

cli/cmd/check.go
```diff
@@ -132,6 +140,14 @@ func (o *checkOptions) validate() error {
 	if o.preInstallOnly && o.dataPlaneOnly {
 		return errors.New("--pre and --proxy flags are mutually exclusive")
 	}
+	ownerParts := strings.Split(o.targetProxyResource, "/")
```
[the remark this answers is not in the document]
alenkacz: good catch, thanks

cli/cmd/check.go
alenkacz: 🤔 chmm, but your proposed solution does not end up doing what I want it to be doing because I need for --proxy to work even without any target specified. So basically for it to keep working as it is working now when --proxy is just a boolean flag. Without target specified, it will target everything.

Anyway it looks like I should be able to do it in `configureAndRunChecks` which is after `Parse` call so I fixed that.

pkg/healthcheck/healthcheck.go
alenkacz: Good catch. No, we don't need it here because the `proxyTarget` is now used directly when querying pods. See the `getDataPlanePods`.

cli/cmd/check.go
dadjeibaah: Ah great. That works even better! 👍

cli/cmd/check.go
```diff
@@ -55,22 +58,26 @@ non-zero exit code.`,
...
-			return configureAndRunChecks(stdout, options)
+			return configureAndRunChecks(cmd, stdout, options)
 		},
 	}
 
 	cmd.Args = cobra.NoArgs
 	cmd.PersistentFlags().StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
 	cmd.PersistentFlags().BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
-	cmd.PersistentFlags().BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
+	cmd.PersistentFlags().StringVar(&options.targetProxyResource, "proxy", options.targetProxyResource, "Only run data-plane checks, to determine if the data plane is healthy")
```
klingerf: I'd like for the target of the data plane check to be optional. In order to do that, we need to leave this as a boolean flag, and instead update the command itself to take an optional additional argument. Here's a diff of what that would look like against this file on master:

```diff
diff --git a/cli/cmd/check.go b/cli/cmd/check.go
index fcf10c21..09bb3ff4 100644
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -53,13 +53,16 @@ non-zero exit code.`,
 
   # Check that the Linkerd data plane proxies in the "app" namespace are up and running
   linkerd check --proxy --namespace app`,
-		Args: cobra.NoArgs,
+		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return configureAndRunChecks(stdout, options)
+			var target string
+			if len(args) == 1 {
+				target = args[0]
+			}
+			return configureAndRunChecks(stdout, options, target)
 		},
 	}
 
-	cmd.Args = cobra.NoArgs
 	cmd.PersistentFlags().StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
 	cmd.PersistentFlags().BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	cmd.PersistentFlags().BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
@@ -70,7 +73,10 @@ non-zero exit code.`,
 	return cmd
 }
 
-func configureAndRunChecks(w io.Writer, options *checkOptions) error {
+func configureAndRunChecks(w io.Writer, options *checkOptions, target string) error {
+	if target != "" {
+		options.dataPlaneOnly = true
+	}
 	err := options.validate()
 	if err != nil {
 		return fmt.Errorf("Validation error when executing check command: %v", err)
@@ -110,6 +116,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 		KubeContext:           kubeContext,
 		APIAddr:               apiAddr,
 		VersionOverride:       options.versionOverride,
+		TargetProxyResource:   target,
 		RetryDeadline:         time.Now().Add(options.wait),
 	})
 
```

pkg/healthcheck/healthcheck.go
```diff
@@ -749,7 +750,21 @@ func (hc *HealthChecker) checkNamespace(ctx context.Context, namespace string, s
 
 func (hc *HealthChecker) getDataPlanePods(ctx context.Context) ([]*pb.Pod, error) {
 	req := &pb.ListPodsRequest{}
-	if hc.DataPlaneNamespace != "" {
+	if hc.Options.TargetProxyResource != "" {
+		parts := strings.Split(hc.Options.TargetProxyResource, "/")
+		ownerType, ownerName := parts[0], parts[1]
+		ownerKind, err := k8s.CanonicalResourceNameFromFriendlyName(ownerType)
+		if err != nil {
+			return nil, err
```
klingerf: Hmm, I'm not exactly sure how to remedy this, but it's slightly problematic for this func to return a validation error to checks that are configured to retry. It means that the check will retry validation errors for up to 5 minutes, even if though they have no chance of succeeding. Maybe the best approach would be to add a new check prior to this check that just validates the target, and make that check non-retryable and fatal? That would cause the check command to exit immediately in the event of an invalid target, allowing the user to re-run the command with a valid target.

alenkacz: well we should never hit this because we validate all this right after check command is processed here https://github.com/linkerd/linkerd2/pull/1777/files#diff-45365b61a96eb768eecd4c352cc736c1R143

Do you think that is not enough?

cli/cmd/check.go
alenkacz: hmm so you think my approach with `cmd.Flags().Changed("proxy")` is not correct? Hmm, yes, you're right, I just tested it. I don't know why am I testing this now and not earlier, sorry 🙈 For my solution it says "Error: flag needs an argument: --proxy"

I'll try your approach

alenkacz: OK, I have some challenges making this work, I'll have to get back to this later.

Btw. will this work also in this way? `linkerd check --proxy deploy/web -n emojivoto` because I am afraid it won't

klingerf: @alenkacz I think that should work. Cobra allows args to be interspersed between flags like that.

pkg/healthcheck/healthcheck.go
klingerf: Hmm, in that case, can we ensure that the `TargetProxyResource` field contains the canonical name, so that we don't have to call `CanonicalResourceNameFromFriendlyName` here, which could return an error?

cli/cmd/check.go
alenkacz: hmm when I did that change locally, cobra is telling me 'deployment/web' is unknown command. Any ideas until I dive deeper? As a part of that endeavor I wanted to update cobra anyway, it seems linkerd is using pretty old version. Thoughts on both?

pkg/healthcheck/healthcheck.go
alenkacz: yeah, you're right. I'll do that

wmorgan:
@alenkacz don't give up!

PR #1800, opened 2018-10-23
Send metric labels in profile API

adleong:
Bump our dependency on the Linkerd2-proxy-api and start sending the route name as a metric label.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -9,7 +9,7 @@ required = [
 
 [[constraint]]
   name = "github.com/linkerd/linkerd2-proxy-api"
-  version = "v0.1.2"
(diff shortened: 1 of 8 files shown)


Gopkg.toml
```diff
@@ -9,7 +9,7 @@ required = [
 
 [[constraint]]
   name = "github.com/linkerd/linkerd2-proxy-api"
-  version = "v0.1.2"
+  revision = "9f5c62278bc10b7bf9bb9facba8ea46b904bc0e8"
```
klingerf: Since this commit is already on master, my preference is to create a v0.1.3 tag in the linkerd2-proxy-api repo, and pin the version to that tag.

Issue #1720, opened 2018-09-27
Allow configuration of cluster base domain


icedream:
Currently linkerd expects the cluster to use `cluster.local` as its base domain name, however more often than not a Kubernetes admin will decide to change that base domain to something else. This shouldn't break deployments, however in the case of linkerd2 it does as every generated configuration, from `linkerd install` to `linkerd inject`, assumes connection to happen via `cluster.local`-based domain names.

A workaround for this issue is to pipe generated configurations into `sed 's/cluster\.local/custom\.domain/g`. [This has been documented shortly in this pull request for the website](https://github.com/linkerd/website/pull/36) but I think it would be good for usage sanity to introduce a CLI option like `--cluster-domain=…` or a similar way to configure the base domain and inject it into the [respective template](https://github.com/linkerd/linkerd2/blob/b5ff29c8aa26a92e8e3a348fdc24b610e278e7e8/cli/install/template.go).

Not sure about this but this may also affect linkerd CLI functionality outside of generating templates. (Checks? Proxies?)

ctaggart:
There is a workaround at https://github.com/linkerd/website/pull/36/files#diff-8f8a7890b984337e6c7f4646339d49a2R121
```
linkerd install | sed s/cluster\.local/name\.of\.your\.cluster/ | kubectl apply -f -
``

**Comment by icedream** [comment]
@ctaggart I mentioned that in the issue description already.

**Comment by olix0r** [comment]
I should note for whoever implements this: up until very recently, the proxy would not do any special discovery (including metric labeling) for services that do not use the default domain suffix. When we add an installation configuration, it should configure the appropriate proxy environment variables:

https://github.com/linkerd/linkerd2-proxy/blob/87fb677cdf5e1598b89b24c3f2adb6ccb6e1f01a/src/app/config.rs#L199-L220

**Comment by grampelberg** [comment]
To pull this all the way through, the domain should be made configurable and be passed to both `install`, `inject` and `profile` (a global CLI flag that will end up being used in multiple places).

**Comment by benley** [comment]
Is the mentioned workaround above known to work?  I just tried deploying with it and still ended up with errors, presumably because of the x509 certs pre-generated in the kubectl manifests that `linkerd install` produces:

```
time="2019-05-16T20:28:12Z" level=info msg="running version stable-2.3.0"
[...]

PR #1808, opened 2018-10-24
Use fqdn for service profile name

adleong:
Service profiles must be named in the form `"<service>.<namespace>"`.  This is inconsistent with the fully normalized domain name that the proxy sends to the controller.  It also does not permit creating service profiles for non-Kubernetes services.

We switch to requiring that service profiles must be named with the FQDN of their service.  For Kubernetes services, this is `"<service>.<namespace>.svc.cluster.local"`.

This change alone is not sufficient for allowing service profile for non-Kubernetes services because the k8s resolver will ignore any DNS names which are not Kubernetes services.  Further refactoring of the resolver will be required to allow looking up non-Kubernetes service profiles in Kuberenetes.

### DIFF
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -15,6 +15,7 @@ type templateConfig struct {
 	ControlPlaneNamespace string
 	ServiceNamespace      string
 	ServiceName           string
+	ClusterZone           string
 }
 
 type profileOptions struct {
@@ -68,6 +69,7 @@ func buildConfig(namespace, service string) *templateConfig {
 		ControlPlaneNamespace: controlPlaneNamespace,
 		ServiceNamespace:      namespace,
 		ServiceName:           service,
+		ClusterZone:           "svc.cluster.local",
 	}
 }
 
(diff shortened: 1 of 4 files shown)


cli/cmd/profile.go
```diff
@@ -68,6 +69,7 @@ func buildConfig(namespace, service string) *templateConfig {
 		ControlPlaneNamespace: controlPlaneNamespace,
 		ServiceNamespace:      namespace,
 		ServiceName:           service,
+		ClusterZone:           "svc.cluster.local",
```
dadjeibaah: I am not familiar with the background for this change but should we be cautious to not assume that the ClusterZone is `svc.cluster.local`? This is related to #1720 and I was thinking maybe this could be another command line flag for the profile command or maybe there is a better way to handle cases were the base cluster domain is not `cluster.local`.

olix0r: Let's keep that configuration out of scope for now. There are a number of places where this string is hardcoded.  Ultimately, I think this setting should be discoverable (rather than configurable).

Issue #1781, opened 2018-10-18
MaterialUI followup: Navigation/Sidebar


rmars:
The dashboard was converted to material-ui in #1776.

Things to be done:
- [x] #1810 I removed the list of resources per namespace and the namespace dropdown in the sidebar added in #1500 for simplicity. we should re-add it. I restructured the Navigation to be part of a larger component which renders the page as a subcomponent, which should make future work on #1669 easier
- [x] #1827 Side navigation doesn’t go all the way down on the Overview page
- [x] #1830 The icons for the side navigation were chosen without a lot of thought. Revisit this, as done in #1637. Resources that can help: https://material.io/tools/icons/?icon=home&style=baseline, https://materialdesignicons.com/. You can also use FontAwesome: (https://material-ui.com/style/icons/#font-icons)

PR #1810, opened 2018-10-25
Re-implement sidebar resource selectors

siggy:
Part of #1781

## 1st commit

nav-res

## 2nd commit

nav-res3

## 3rd commit

nav-res4

### DIFF
--- a/web/app/js/components/NavigationResource.jsx
+++ b/web/app/js/components/NavigationResource.jsx
@@ -1,6 +1,5 @@
 import {
   Collapse,
-  IconButton,
   ListItemIcon,
   ListItemText,
   MenuItem,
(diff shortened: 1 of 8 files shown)


web/app/js/components/NavigationResources.jsx
```diff
@@ -0,0 +1,105 @@
...
+      if (_.has(data, '[1]')) {
+        nsMetrics = processMultiResourceRollup(data[1]);
+      }
+    }
+
+    return (
+      <MenuList dense component="div" disablePadding>
+        <NavigationResource type="authorities" />
+        <NavigationResource type="deployments" metrics={allMetrics.deployment} />
+        <NavigationResource type="namespaces" metrics={nsMetrics.namespace} />
```
rmars: if I click on the `linkerd` namespace, then click back and forth between `emojivoto` and `linkerd`, the uri changes, but the component doesn't actually re-render. It does for the other resource types though.

web/app/js/components/NavigationResource.jsx
```diff
@@ -0,0 +1,145 @@
...
+        <ListItemText primary={friendlyTitle(type).plural} />
+      </MenuItem>
+    );
+  }
+
+  subMenu() {
+    const { api, classes } = this.props;
+
+    return (
+      <MenuList dense component="div" disablePadding>
```
rmars: what do you think about making the first subMenu Item "All Pods" (etc), rather than navigating to the pods list page  when you click to expand the submenu? I find it a little disconcerting if I'm, for example, in the middle of a top page, and I just want to see what resources are available in the sidebar, but I get navigated off the current page.

web/app/js/components/util/SuccessRateMiniChart.jsx
```diff
@@ -1,37 +1,32 @@
...
   render() {
-    const { sr, classes } = this.props;
+    const { sr } = this.props;
 
     return (
       <Grid container spacing={8}>
         <Grid item>{metricToFormatter["SUCCESS_RATE"](sr)}</Grid>
         <Grid item>{_.isNil(sr) ? null :
-        <div className={classNames("success-rate-dot", classes[getSuccessRateClassification(sr)])} />}
+        <SuccessRateDot sr={sr} />
```
rmars: sweet, thanks for DRYing this!

web/app/js/components/NavigationResource.jsx
siggy: @rmars agree it's disorienting when navigating the resource tree causes the page to change. in the second commit, i tried adding a small button, visible on hover, that provides the "all" link you described, while keeping the rest of the menu item clickable for navigation only. let me know what you think.

rmars: hmmm do you have objections to just adding an "All <resource>s" option as the first option in the list? I think it's weird to have another icon just pop up that you don't know what it does (even if it's the list icon that Resources uses). I feel like having an "All <resource>" option makes it very clear as to what is contained in the option. 

Having options to click an adjacent icon as well as click a list item also feels weird in that it divides the actions you could take. This introduces more complication than the simple "I click on a menu and then a submenu" flow. What would be the motivations for breaking this flow?

siggy: @rmars Added an "All" list item as you suggested. You're right a random button appearing is a bit confusing. I was trying to avoid having a list item at the top of the menu that's different from the individual resources below it, as I haven't seen that pattern in a tree navigation UI before. That said, this makes the code way simpler, and is the most explicit for the user. Thanks for the thorough eyes!

rmars: yeah, I do wonder how most people actually do this! thanks for adding!

siggy: ooh, thanks for finding these! pop-up style is pretty cool. something to consider down the road.

PR #1813, opened 2018-10-25
Remove the proxy-api container

adleong:
A container called `proxy-api` runs in the Linkerd2 controller pod.  This container listens on port 8086 and serves the proxy-api but does nothing other than forward gRPC requests to the destination container which listens on port 8089.

We remove the proxy-api container altogether and change the destination container to listen on port 8086 instead of 8089.  The result is that clients still use the proxy-api by connecting to `proxy-api.<ns>.svc.cluster.local:8086` but the controller has one fewer containers.  This results in a simpler system that is easier to reason about.


olix0r:
Should we call the _destination_ container _proxy-api_? Perhaps as a followup?

controller/cmd/destination/main.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func main() {
-	addr := flag.String("addr", "127.0.0.1:8089", "address to serve on")
+	addr := flag.String("addr", ":8086", "address to serve on")
```
dadjeibaah: Is there a reason we are no longer using `127.0.0.1`? I am sure it's not a big deal but just curious

adleong: Yeah, I didn't make this change at first and it took me hours to figure out what was wrong.

This causes the destination server to listen on all interfaces instead of just localhost.  If it only listens on localhost then the proxy can't connect to it.

dadjeibaah: Ah, I get it. Thanks!

controller/api/proxy/server.go
```diff
@@ -1,96 +1,151 @@
...
-type (
-	server struct {
-		destinationClient destination.DestinationClient
+type server struct {
+	k8sAPI    *k8s.API
+	resolvers []streamingDestinationResolver
+	enableTLS bool
+}
+
+// The Destination service serves service discovery information to the proxy.
```
klingerf: Should this be:

> // The proxy-api services serves...

controller/script/destination-client/main.go
```diff
@@ -18,12 +18,12 @@ import (
 func main() {
 	rand.Seed(time.Now().UnixNano())
 
-	addr := flag.String("addr", ":8089", "address of destination service")
+	addr := flag.String("addr", ":8086", "address of destination service")
```
klingerf: "address of destination service" => "address of the proxy API" or something similar? On a related note, maybe we should rename this directory from `destination-client` to `proxy-api-client`?

adleong: I've gone back and forth on this.  You can think of this as a client of the the destination gRPC service, which is served by the proxy-api process.  So I can see it either way.  What do you think?

Issue #1792, opened 2018-10-20
Dashboard: Add sorting and filtering to Metric Tables


rmars:
If we don't want to implement the sorting/filtering ourselves, https://github.com/gregnb/mui-datatables looks like a good option.

klingerf:
Related to this issue, I noticed that the top tables in the UI are no longer sorted in descending order by request count:

![image](https://user-images.githubusercontent.com/9226/47454442-1a32ec80-d784-11e8-9fc2-05038edc5779.png)

When support for sorting is added, we should restore the default sort by count for these tables.

klingerf:
Sorting was implemented in #1814, but filtering still needs to happen.

scottcarol:
@rmars @grampelberg What do you think of this proposed UI for filtering tables, inspired by Material-UI [table demo](https://material-ui.com/demos/tables/)?

Top image is existing UI.

Bottom image is proposed UI. It moves the table title inside the table box itself. The filter field is on the top right. I think the size of the filter field could be a little smaller, but want feedback from you on overall look and feel.

![currentlinkerdtable](https://user-images.githubusercontent.com/2289389/52655547-409ea400-2ea9-11e9-90ec-7b79c6d50357.png)

![proposed linkerd table](https://user-images.githubusercontent.com/2289389/52655558-44322b00-2ea9-11e9-96b3-793e07510d0d.png)

grampelberg:
I like it, here's the k8s dashboard for reference (which appears to do it the same way).

<img width="1006" alt="screenshot 2019-02-12 09 37 11" src="https://user-images.githubusercontent.com/47992/52655775-dcc8ab00-2ea9-11e9-9eab-0360494cdb1a.png">
<img width="1017" alt="screenshot 2019-02-12 09 37 17" src="https://user-images.githubusercontent.com/47992/52655776-dcc8ab00-2ea9-11e9-8435-6e800b22ac71.png">

scottcarol:
@grampelberg Thanks, good call on referencing k8s dashboard!

PR #1814, opened 2018-10-25
Re-add sortable column headers to tables in web UI

klingerf:
This branch updates `BaseTable` to include support for sortable columns, and re-adds the `sortable` field to columns that were previously sortable. As part of this change, I've updated `BaseTable` to more closely resemble antd's implementation of the `dataIndex`, `key`, `render` and `sortable` fields.

Relates to #1792.

### DIFF
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -7,7 +7,6 @@ import TableCell from '@material-ui/core/TableCell';
 import TableHead from '@material-ui/core/TableHead';
 import TableRow from '@material-ui/core/TableRow';
 import TableSortLabel from '@material-ui/core/TableSortLabel';
-import Tooltip from '@material-ui/core/Tooltip';
 import _ from 'lodash';
 import { withStyles } from '@material-ui/core/styles';
 
@@ -19,6 +18,12 @@ const styles = theme => ({
     overflowX: 'auto',
   },
   table: {},
+  activeSortIcon: {
+    opacity: 1,
+  },
+  inactiveSortIcon: {
+    opacity: 0.4,
+  },
 });
 
 class BaseTable extends React.Component {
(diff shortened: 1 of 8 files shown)


web/app/js/components/BaseTable.jsx
```diff
@@ -19,59 +21,125 @@ const styles = theme => ({
...
+  }
+
+  renderHeaderCell = (col, order, orderBy) => {
+    if (col.sorter) {
+      return (
+        <TableCell
+          key={col.key || col.dataIndex}
+          numeric={col.isNumeric}
+          sortDirection={orderBy === col.dataIndex ? order : false}>
+          <Tooltip
```
rmars: I feel like an updown icon that's present all the time is a better visual indicator of whether a column is Sortable than the tooltip, since it's more discoverable. I'm not sure the tooltip gives us that much here.

screen shot 2018-10-25 at 4 45 18 pm

I couldn't find one on the default material icons page, I did find this one under "sort":
https://material.io/tools/icons/?icon=sort&style=baseline

web/app/js/components/BaseTable.jsx
```diff
@@ -19,59 +21,125 @@ const styles = theme => ({
...
+          key={col.key || col.dataIndex}
+          numeric={col.isNumeric}>
+          {col.title}
+        </TableCell>
+      );
+    }
+  }
+
+  render() {
+    const { classes, tableRows, tableColumns, tableClassName, rowKey, padding} = this.props;
```
rmars: tiny nit, `}` spacing inconsistent with `{` spacing

web/app/js/components/BaseTable.jsx
rmars: Oooh I found https://material.io/tools/icons/?icon=unfold_more&style=baseline

PR #1817, opened 2018-10-26
Update CHANGES.md for edge-18.10.4 release

rmars:


CHANGES.md
```diff
@@ -1,3 +1,14 @@
+## edge-18.10.4
+* CLI
```
klingerf: Can you add an extra space between the header and the first bullet here? Also, we should call out the material switchover more prominently. It's a Big Deal!!!

```
## edge-18.10.4

This releases includes a major redesign of the web frontend to make use of the
Material design system. Additional features that leverage the new design are
coming soon. This release also includes the following changes:

* CLI
  * **Fixed** Fixed an issue with the `--registry` install flag not accepting
    hosts with ports (thanks, @alenkacz!)
* Web UI
  * **New** Added a new Grafana dashboard for authorities (thanks, @alpeb!)
```

CHANGES.md
```diff
@@ -1,3 +1,14 @@
...
+  * **New** adds the concept of service profiles to the controller
```
wmorgan: Is there a user-facing impact of this change that we can describe?

CHANGES.md
```diff
@@ -1,3 +1,14 @@
...
+* Web UI
+  * **New** Revamped the look of the dashboard by switching component libraries
```
wmorgan: How about: Major look and feel update to the Linkerd dashboard by switching component libraries from antd to material-ui

CHANGES.md
rmars: Yeah, I think we should hold off on the bullet point about service profiles until we have a user-facing feature!

PR #1827, opened 2018-10-30
Fix sidebar not rendering to end of page

siggy:
Also make main content independently scrollable.

Part of #1781

scrollz2


siggy:
@rmars good catch! this fix was also a good excuse to define all styles terms of `theme.spacing.unit`.

PR #1828, opened 2018-10-30
Improve ergonomics of service profile spec

adleong:
We make several changes to the service profile spec to make service profiles more ergonomic and to make them more consistent with the destination profile API.

* Allow multiple fields to be simultaneously set on a RequestMatch or ResponseMatch condition.  Doing so is equivalent to combining the fields with an "all" condition.
* Rename "responses" to "response_classes"
* Change "IsSuccess" to "is_failure"


olix0r:
This appears to be good, from what I can tell... but, it would be pretty helpful to have an example Yaml file in the repo so we can really see a before/after...  Perhaps we should store the API spec for linkerd's API itself?

adleong:
It was my intent that [template.go](https://github.com/linkerd/linkerd2/pull/1828/files#diff-f00cc02e90fcc303b0cbd4a7f69b7e46) serve as an example of the yaml structure in the repo.  The diff on that file should demonstrate what has changed.

PR #1830, opened 2018-10-30
Replace some sidebar icons with Font Awesome

siggy:
This replaces a couple of the MaterialUI icons introduced in #1776 with
their original counterparts in Font Awesome, but wrapped in a MaterialUI
`Icon` tag. Also fix Linkerd logo padding in sidebar.

Part of #1781.

screen shot 2018-10-30 at 1 59 57 pm

### DIFF
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -122,8 +122,8 @@ const styles = theme => {
       paddingRight: `${contentPadding}px`,
     },
     shrinkIcon: {
-      fontSize: "20px",
-      paddingLeft: "2px",
+      fontSize: "18px",
+      paddingLeft: "3px",
     }
   };
 };
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -121,6 +121,10 @@ const styles = theme => {
       paddingLeft: `${contentPadding}px`,
       paddingRight: `${contentPadding}px`,
     },
+    shrinkIcon: {
+      fontSize: "20px",
+      paddingLeft: "2px",
+    }
   };
 };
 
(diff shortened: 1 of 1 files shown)


siggy:
@rmars Per your comment in https://github.com/linkerd/linkerd2/issues/1781#issue-371688811, I browsed through the MaterialUI and Font Awesome icons, and couldn't find better fits than what we had before the MaterialUI switch, so I re-added the Font Awesome icons we had for Tap and Top. The Tap microscope now matches the links on the Tap page.

For Service Mesh, this was the best I could find, which I don't think is as good as the existing Cloud icon:
https://fontawesome.com/icons/vector-square?style=solid

siggy:
@rmars modified the two font awesome icons and one material icon to be the same width as most of the others. the material icons vary slightly by height and width, so the resources icon is still a bit smaller, and the cloud icon has less height. let me know what you think...

## before

before

## after

after

web/app/js/components/Navigation.jsx
```diff
@@ -115,13 +114,17 @@ const styles = theme => {
...
-      marginLeft: `-${navLogoWidth-theme.spacing.unit/2}px`,
+      marginLeft: `-${navLogoWidth+theme.spacing.unit/2}px`,
       transition: leavingFn(['margin', 'opacity']),
     },
     navMenuItem: {
       paddingLeft: `${contentPadding}px`,
       paddingRight: `${contentPadding}px`,
     },
+    shrinkIcon: {
+      fontSize: "20px",
```
rmars: one nit: can we reduce this ever so slightly, to 18px?

siggy: size updated, also bumped the padding to make the icons align through their centers:

screen shot 2018-10-31 at 1 18 55 pm

PR #1834, opened 2018-10-31
Rewrite octopus arm code to be more parameterized and flexible.

rmars:
As a result, displays better in the material UI version of the dashboard.
Also adds Success rate to data displayed on neighbour nodes.

I also limited the width of the chart so it doesn't expand super wide when in fullscreen. There's still about a 10px gap if you do expand it, but I think that could be easily fixed later.

# Before
screen shot 2018-10-31 at 3 12 36 pm

# After
screen shot 2018-10-31 at 3 26 54 pm

### DIFF
--- a/web/app/css/styles.css
+++ b/web/app/css/styles.css
@@ -59,8 +59,8 @@ div.octopus-graph-container {
 }
 
 div.octopus-graph {
-  max-width: 980px;
-  min-width: 980px;
+  max-width: 974px;
+  min-width: 974px;
   margin-left: auto;
   margin-right: auto;
 }
--- a/web/app/css/styles.css
+++ b/web/app/css/styles.css
@@ -55,6 +55,7 @@ div.success-rate-dot {
 
 div.octopus-graph-container {
   overflow-x: scroll;
+  padding: 16px 0;
 }
 
 div.octopus-graph {
--- a/web/app/css/styles.css
+++ b/web/app/css/styles.css
@@ -59,12 +59,15 @@ div.octopus-graph-container {
 }
 
 div.octopus-graph {
-  max-width: 1000px;
-  min-width:1000px;
+  max-width: 980px;
+  min-width: 980px;
   margin-left: auto;
   margin-right: auto;
 }
 
-.octopus-body-node {
-  width: 232px;
+.octopus-body-node.neighbor {
+  width: 220px;
+}
+.octopus-body-node.main {
+  width: 244px;
 }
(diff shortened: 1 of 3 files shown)


web/app/js/components/util/OctopusArms.jsx
```diff
@@ -1,65 +1,74 @@
...
-  } else {
-    y1 = svgHeight / 2 ;
-  }
-
-  let svgPaths = svgArrow(x1, y1, width, height, "up");
+  let translate = isOutbound ? null : `translate(0, ${svgHeight / 2 + (isEven ? 0 : halfBoxHeight) + inboundAlignment})`;
 
   return (
-    <g key={`up-arrow-${height}`} id="downstream-up" fill="none" stroke="none" strokeWidth="1">
+    <g key={`up-arrow-${height}`} id="downstream-up" fill="none" strokeWidth="1">
```
siggy: `up` and `down` return pretty similar `g` components, what do think about factoring those out?

in `up`:

```js
return g(height, arm, translate, null, "up");
```

in `down`:

```js
  return g(height, arm, translate, reflect, "down");
```

and then...
```js
const g = (height, arm, translate, reflect, direction) => {
  return (
    <g key={`${direction}-arrow-${height}`} id={`downstream-${direction}`} fill="none" strokeWidth="1">
      <path
        d={arm.arrowPath}
        stroke={arrowColor}
        transform={`${translate} ${reflect}`}
        strokeOpacity={strokeOpacity} />
      <circle
        cx={arm.circle.cx}
        cy={arm.circle.cy}
        transform={`${translate} ${reflect}`}
        fill={arrowColor}
        r="4" />
      <polyline
        points={arm.arrowHead}
        stroke={arrowColor}
        strokeLinecap="round"
        transform={`${translate} ${reflect}`} />
    </g>
  );
};
```

rmars:
yeahhh I think that we don't actually want the responsive features of the grid for this particular graph, so I've made it a fixed width (and also centered it). This will allow us to just pick one width for the arrows, and stick with that.

web/app/css/styles.css
```diff
@@ -52,3 +52,18 @@ div.success-rate-dot {
...
   }
 }
+
+div.octopus-graph-container {
+  overflow-x: scroll;
+}
+
+div.octopus-graph {
+  max-width: 1126px;
+  min-width:1126px;
```
siggy: not sure if this is the cause, but on laptop-size screen, getting weird sizing / overflow issues:

screen shot 2018-10-31 at 6 57 15 pm

screen shot 2018-10-31 at 6 56 55 pm

rmars: thanks @siggy ! turns out I'd been viewing the dashboard at 90% zoom in chrome all along. I've reduced the sizes, should be good now!

Issue #622, opened 2018-03-26
Validate that linkerd images are accessible to cluster before running integration tests


klingerf:
Per https://github.com/runconduit/conduit/pull/608/files#r176890012, we should update the `bin/test-run` script to validate that all of the images required for running integration tests are accessible to the Kubernetes cluster where the tests are going to run, prior to initiating the tests. Offhand, I don't know the best way to validate that without deploying a pod that actually pulls all of the images.

klingerf:
We fixed this issue in a different way as part of #1835 -- closing.

PR #1835, opened 2018-10-31
Increase retry timeout for retryable tests, refactor RetryFor

### LINKED ISSUE #1735: integration tests: 30s retry timeout too short
When tap_test.go is setting up, it will wait 30s for the test target to become available:

https://github.com/linkerd/linkerd2/blob/dccccebd79e1fb041194e7acdb61640bd22dcb3b/test/tap/tap_test.go#L87-L98

If the test is being run on a non-local cluster, it can spend 20-30s just pulling docker images, let alone running anything.

Can we increase these timeouts safely?  If not, can we make them configurable?

dadjeibaah:
When running integration tests in a Kubernetes cluster that sometimes takes a little longer to get pods ready, the integration tests fail tests too early because most tests have a retry timeout of 30 seconds. 

This PR bumps up this retry timeout for `TestInstall` to 3 minutes. This gives the test enough time to download any new docker images that it needs to complete succesfully and also reduces the need to have large timeout values for subsequent tests. This PR also refactors `CheckPods` to check that all containers in a pods for a deployment are in a`Ready` state. This helps also helps in ensuring that all docker images have been downloaded and the pods are in a good state.

Tests were run on the community cluster and all were successful.

Fixes #1735, #622 

### DIFF
--- a/test/egress/testdata/proxy.yaml
+++ b/test/egress/testdata/proxy.yaml
@@ -165,7 +165,7 @@ spec:
 apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
-  name: egress-test-http-get
+  name: egress-test-not-www-get
 spec:
   replicas: 1
   selector:
--- a/testutil/kubernetes_helper.go
+++ b/testutil/kubernetes_helper.go
@@ -20,10 +20,11 @@ import (
 // Kubernetes API using the environment's configured kubeconfig file.
 type KubernetesHelper struct {
 	clientset *kubernetes.Clientset
+	retryFor  func(func() error) error
 }
 
 // NewKubernetesHelper creates a new instance of KubernetesHelper.
-func NewKubernetesHelper() (*KubernetesHelper, error) {
+func NewKubernetesHelper(retryFor func(func() error) error) (*KubernetesHelper, error) {
 	rules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := &clientcmd.ConfigOverrides{}
 	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
@@ -39,6 +40,7 @@ func NewKubernetesHelper() (*KubernetesHelper, error) {
 
 	return &KubernetesHelper{
 		clientset: clientset,
+		retryFor:  retryFor,
 	}, nil
 }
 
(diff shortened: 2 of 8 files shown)


testutil/test_helper.go
```diff
@@ -15,6 +15,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const DefaultRetryTimeout = 2 * time.Minute
```
siggy: since `RetryFor` is always called with this default timeout, consider removing it as a parameter until we need it.

testutil/test_helper.go
```diff
@@ -93,6 +95,31 @@ func NewTestHelper() *TestHelper {
 	return testHelper
 }
 
+// RetryFor retries a given function every second until the function returns
+// without an error, or a timeout is reached. If the timeout is reached, it
+// returns the last error received from the function.
+func RetryFor(timeout time.Duration, fn func() error) error {
```
siggy: 👍 for removing this from `*TestHelper`

klingerf: My preference is to revert this change, and leave `RetryFor` as a function on the `TestHelper` struct. One of the main motivations behind having the `TestHelper` struct is to prevent the proliferation of a bunch of global helper functions.

To add retryability to the `KubernetesHelper` struct, I recommend passing the `RetryFor` func as a parameter when initializing that struct. That way it's available to both structs but only defined in one place. Something like this:

```diff
diff --git a/testutil/kubernetes_helper.go b/testutil/kubernetes_helper.go
index 55fe82ed..ffcbf68c 100644
--- a/testutil/kubernetes_helper.go
+++ b/testutil/kubernetes_helper.go
@@ -5,6 +5,7 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	coreV1 "k8s.io/api/core/v1"
@@ -20,10 +21,11 @@ import (
 // Kubernetes API using the environment's configured kubeconfig file.
 type KubernetesHelper struct {
 	clientset *kubernetes.Clientset
+	RetryFor  func(time.Duration, func() error) error
 }
 
 // NewKubernetesHelper creates a new instance of KubernetesHelper.
-func NewKubernetesHelper() (*KubernetesHelper, error) {
+func NewKubernetesHelper(retryFor func(time.Duration, func() error) error) (*KubernetesHelper, error) {
 	rules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := &clientcmd.ConfigOverrides{}
 	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
@@ -39,6 +41,7 @@ func NewKubernetesHelper() (*KubernetesHelper, error) {
 
 	return &KubernetesHelper{
 		clientset: clientset,
+		RetryFor:  retryFor,
 	}, nil
 }
 
diff --git a/testutil/test_helper.go b/testutil/test_helper.go
index 47aa711f..4f56165a 100644
--- a/testutil/test_helper.go
+++ b/testutil/test_helper.go
@@ -80,7 +80,7 @@ func NewTestHelper() *TestHelper {
 	}
 	testHelper.version = strings.TrimSpace(version)
 
-	kubernetesHelper, err := NewKubernetesHelper()
+	kubernetesHelper, err := NewKubernetesHelper(testHelper.RetryFor)
 	if err != nil {
 		exit(1, "error creating kubernetes helper: "+err.Error())
 	}
```

(For a bit of additional background: we originally had only one `*Helper` struct, but decided to split it into two based on the types of helpers they were providing, as a way to keep things organized. I'm not actually sure we benefit all that much from having two separate structs, however, so in the long run we may end up collapsing them back into a single struct, if it turns out there's a lot of overlap among the helpers. But in the near term passing `RetryFor` as a param seems like an ok solution to me.)

testutil/test_helper.go
dadjeibaah: Agreed, I will make that change in the next commit.

testutil/kubernetes_helper.go
```diff
@@ -137,31 +143,37 @@ func (h *KubernetesHelper) getPods(namespace string) (map[string]coreV1.PodPhase
 // CheckPods checks that a deployment in a namespace contains the expected
 // number of pods in the Running state.
 func (h *KubernetesHelper) CheckPods(namespace string, deploymentName string, replicas int) error {
```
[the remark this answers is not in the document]
dadjeibaah: TIL: Pod in a Running state doesn't mean that ALL containers are ready. Thanks for this @klingerf!

test/egress/testdata/proxy.yaml
```diff
@@ -165,7 +165,7 @@ spec:
 apiVersion: apps/v1beta1
 kind: Deployment
 metadata:
-  name: egress-test-http-get
+  name: egress-test-not-www-get
```
klingerf: Ugh are you serious? This test has been reconfiguring an existing deployment the whole time?! Wow. Thanks for finding and fixing it!

dadjeibaah: I was trying to figure out why the tests kept failing and it turned out to be this. 😩

PR #1836, opened 2018-10-31
add control plane check before running proxy test assertion

### LINKED ISSUE #1733: Integration tests fail when check retries
This check command was actually successful, but because it retried, the integration test is considered a failure. Is there something that we can do to prevent the integration tests failing in cases like this?

```
--- FAIL: TestCheckProxy (10.23s)
	install_test.go:237: Received unexpected output
		Expected:
		kubernetes-api: is running the minimum Kubernetes API version..............[ok]
		linkerd-api: control plane namespace exists................................[ok]
		linkerd-api: control plane pods are ready..................................[ok]
		linkerd-api: can initialize the client.....................................[ok]
		linkerd-api: can query the control plane API...............................[ok]
		linkerd-api[kubernetes]: control plane can talk to Kubernetes..............[ok]
		linkerd-api[prometheus]: control plane can talk to Prometheus..............[ok]
		linkerd-data-plane: data plane namespace exists............................[ok]
		linkerd-data-plane: data plane proxies are ready...........................[ok]
		linkerd-data-plane: data plane proxy metrics are present in Prometheus.....[ok]
		linkerd-version: can determine the latest version..........................[ok]
		linkerd-version: cli is up-to-date.........................................[ok]
		linkerd-version: data plane is up-to-date..................................[ok]
		
		Status check results are [ok]
		
		Actual:
		kubernetes-api: is running the minimum Kubernetes API version..............[ok]
		linkerd-api: control plane namespace exists................................[ok]
		linkerd-api: control plane pods are ready..................................[ok]
		linkerd-api: can initialize the client.....................................[ok]
		linkerd-api: can query the control plane API...............................[ok]
		linkerd-api[kubernetes]: control plane can talk to Kubernetes..............[ok]
		linkerd-api[prometheus]: control plane can talk to Prometheus..............[ok]
		linkerd-data-plane: data plane namespace exists............................[ok]
		linkerd-data-plane: data plane proxies are ready...........................[retry] -- The "linkerd-proxy" container in the "ver-stacked-tls-smoke-test/smoke-test-gateway-796c7c56d-lzr7j" pod is not ready
		linkerd-data-plane: data plane proxies are ready...........................[retry] -- The "linkerd-proxy" container in the "ver-stacked-tls-smoke-test/smoke-test-gateway-796c7c56d-lzr7j" pod is not ready
		linkerd-data-plane: data plane proxies are ready...........................[ok]
		linkerd-data-plane: data plane proxy metrics are present in Prometheus.....[ok]
		linkerd-version: can determine the latest version..........................[ok]
		linkerd-version: cli is up-to-date.........................................[ok]
		linkerd-version: data plane is up-to-date..................................[ok]
		
		Status check results are [ok]
```

dadjeibaah:
This PR adds a check before the `TestCheckProxy` executes the CLI command to make sure the control plane has installed all its pods. This avoids the situation where the test would fail because the pods are retrying the data plane check.  The PR's base is #1835 but will switch to master once that PR merges.

Relates to PR #1835 
Fixes #1733 

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -131,7 +131,12 @@ func TestCheckPostInstall(t *testing.T) {
 	var out string
 	var err error
 	overallErr := TestHelper.RetryFor(30*time.Second, func() error {
-		out, _, err = TestHelper.LinkerdRun("check", "--expected-version", TestHelper.GetVersion())
+		out, _, err = TestHelper.LinkerdRun(
+			"check",
+			"--expected-version",
+			TestHelper.GetVersion(),
+			"--wait=0",
+			)
 		return err
 	})
 	if overallErr != nil {
@@ -220,14 +225,21 @@ func TestInject(t *testing.T) {
 
 func TestCheckProxy(t *testing.T) {
 	prefixedNs := TestHelper.GetTestNamespace("smoke-test")
-	out, _, err := TestHelper.LinkerdRun(
-		"check",
-		"--proxy",
-		"--expected-version",
-		TestHelper.GetVersion(),
-		"--namespace",
-		prefixedNs,
-	)
+	var out string
+	err := TestHelper.RetryFor(2*time.Minute, func() error {
+		var err error
+		out, _, err = TestHelper.LinkerdRun(
+			"check",
+			"--proxy",
+			"--expected-version",
+			TestHelper.GetVersion(),
+			"--namespace",
+			prefixedNs,
+			"--wait=0",
+		)
+		return err
+	})
+
 	if err != nil {
 		t.Fatalf("Check command failed\n%s", out)
 	}


kleimkuhler:
LGTM!

test/install_test.go
```diff
@@ -206,13 +206,25 @@ func TestInject(t *testing.T) {
...
+
+	// Tests Pods and Deployments
+	for deploy, replicas := range linkerdDeployReplicas {
+		if err := TestHelper.CheckPods(TestHelper.GetLinkerdNamespace(), deploy, replicas); err != nil {
+			t.Error(fmt.Errorf("Error validating pods for deploy [%s]:\n%s", deploy, err))
+		}
+		if err := TestHelper.CheckDeployment(TestHelper.GetLinkerdNamespace(), deploy, replicas); err != nil {
+			t.Error(fmt.Errorf("Error validating deploy [%s]:\n%s", deploy, err))
+		}
+	}
```
klingerf: This code unfortunately won't fix #1733. The failure described in that issue is caused by the linkerd-proxy containers in the "smoke-test" namespace not being in the ready state by the time `linkerd check --proxy` is run. But this code is only checking the pods in the "linkerd" namespace, not the "smoke-test" namespace.

Since you're adding the `--wait=0` flag, I think the easiest fix for the issue is to just retry running the check command until it succeeds. Something like this:

```
	var out string
	err := TestHelper.RetryFor(30*time.Second, func() error {
		var err error
		out, _, err = TestHelper.LinkerdRun(
			"check",
			"--proxy",
			"--expected-version",
			TestHelper.GetVersion(),
			"--namespace",
			prefixedNs,
			"--wait=0",
		)
		return err
	})
```

That's pretty similar to how `TestCheckPostInstall` is already setup in this file. Ideally this test and that test would be almost identical, save for the `--proxy` and `--namespace` flags. Speaking of which, we should add the `--wait=0` flag to the `TestCheckPostInstall` test, too.

test/install_test.go
```diff
@@ -220,14 +225,21 @@ func TestInject(t *testing.T) {
...
-	out, _, err := TestHelper.LinkerdRun(
-		"check",
-		"--proxy",
-		"--expected-version",
-		TestHelper.GetVersion(),
-		"--namespace",
-		prefixedNs,
-	)
+	var out string
+	err := TestHelper.RetryFor(2*time.Minute, func() error {
```
klingerf: I think we should use a tighter timeout here, since I would consider it a bug if it takes the proxies 2 minutes to reach the ready state. For the other check test, we're using 30 seconds. Mind updating this timeout to match?

dadjeibaah: Per our conversation, these timeouts will be revisited in #1835.

Issue #1832, opened 2018-10-31
proxy: Predictable response classification logic


olix0r:
When the proxy calculates latency, it includes a "classification". That is, it includes labels that indicate whether the response should be considered a success or failure.

Because the proxy, rightly, measures latency as it processes the first data frame (or EOS), this means that we must classify responses eagerly, before we know whether the rest of the response completes successfully.

Alternatively, in some cases (like for gRPC streams), we cannot know the classification until the stream completes, and as such we cannot update the latency in a timely fashion.  This can make metrics data misleading, as it may hide or misrepresent latency in the system.

To resolve this---and to simplify classification semantics for service profiles--I propose the following:

1. The proxy continues to report latency as time-to-first-byte.
2. Response latencies will _not_ include classification.
3. Response counts will include classification labels, and will only be incremented once a stream completes.

This allows us to change the classification API so that only complete streams are classified.

klingerf:
Omitting `classification` labels from latency metrics works for me. I've verified that the controller only consumes `classification` labels for the `response_total` metric.

olix0r:
@klingerf Thanks for checking. I think we should change `tcp_close_total` requests to stop using `classification` -- the majority of sockets will close with some error code, though this is in no way a failure. I'll look at making this change.

hawkw:
This seems correct to me 👍

PR #1839, opened 2018-11-01
grafana: Show TCP closes by errno

olix0r:
linkerd/linkerd2-proxy#116 removes the `classification` label for the
`tcp_close_total` metric because TCP sockets that close with an error
do not actually indicate any sort of failure -- many graceful shutdown
situations can still cause a socket error.

This change uses the `errno` label to enumerate tcp_close_total metrics.

Relates to #1832

### DIFF
--- a/grafana/dashboards/health.json
+++ b/grafana/dashboards/health.json
@@ -694,7 +694,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",classification=\"failure\"}",
+              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",errno=\"\"}",
               "format": "time_series",
               "intervalFactor": 1,
               "refId": "A"
--- a/grafana/dashboards/health.json
+++ b/grafana/dashboards/health.json
@@ -694,9 +694,10 @@
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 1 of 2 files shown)


grafana/dashboards/health.json
```diff
@@ -694,7 +694,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",classification=\"failure\"}",
+              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",errno=\"\"}",
```
klingerf: Ah, hmm, this chart is a bit off:

image

In order to display failures, I think your match condition needs to be `errno!=\"\"`, not `errno=\"\"`.

And this chart doesn't include a legend, which is why the tooltip is so long. If you update this chart to include a legendFormat attr on line 700, that should fix it.

```diff
diff --git a/grafana/dashboards/health.json b/grafana/dashboards/health.json
index ed22d43c..2cf904cb 100644
--- a/grafana/dashboards/health.json
+++ b/grafana/dashboards/health.json
@@ -694,9 +694,10 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",errno=\"\"}",
+              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",errno!=\"\"}",
               "format": "time_series",
               "intervalFactor": 1,
+              "legendFormat": "{{peer}} {{errno}}",
               "refId": "A"
             }
           ],
```

Issue #1487, opened 2018-08-17
Allow stat summary to query for multiple resources


rmars:
`kubectl get` allows you to list multiple resources at once, for example:
```
$ kubectl get deploy/web ns/one-em po/voting-88d6d67f4-xbgsl -n emojivoto
NAME         DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deploy/web   2         2         2            2           8m

NAME        STATUS    AGE
ns/one-em   Active    6m

NAME                        READY     STATUS    RESTARTS   AGE
po/voting-88d6d67f4-xbgsl   2/2       Running   0          8m
```

We should update `linkerd stat` to also accept multiple resources and query for those stats.

Issue #911, opened 2018-05-08
Modify the Stat endpoint to accept multiple resource types


rmars:
As of #893 the ServiceMesh page gets stats from the /stat endpoint. It has to make one call per resource type, but it would be nice if this could be handled by a single api call.

Issue #683, opened 2018-04-05
Add `kubectl get`-style resource parsing to `conduit stat` CLI command


rmars:
The queries proposed in #627 require more sophisticated parsing than the stat summary endpoints currently do. (Current implementation in #671 just takes a singular resource and name).

Add parsing to the CLI to take queries such as the following and generate a series of `StatSummaryRequests`.

```
  # Stat all replication controllers and services together.
  conduit stat rc,services

  # Stat one or more resources by their type and names.
  conduit stat rc/web service/frontend pods/web-pod-13je7
```

klingerf:
As part of this change, we should also enable kubectl-style resource parsing for the `--to*` and `--from*` flags.

For example, as it's currently implemented, if you wanted to query all of the namespaces that send traffic to the "foo" deployment in the default namespace, you'd use:

```
conduit stat ns --to foo --to-resource deployment --to-namespace default
```

But we should make the `--to` flag resource-aware, such that you can instead run:

```
conduit stat ns --to deploy/foo --to-namespace default
```

siggy:
Moving this into 0.5.0 milestone. This task is mostly complete. What remains is handing comma-separated lists of resource strings.

PR #1841, opened 2018-11-04
Allow stat summary to query for multiple resources

alpeb:
Howdy folks,

This will allow us to query multiple resources of different types in a single call (#1487), like
```
linkerd stat deploy/web po/emoji-694f94cdff-j44qs
```
or multiple resources of the same type by specifying the type at the beginning:
```
linkerd stat po emoji-694f94cdff-j44qs vote-bot-546746cb47-fswth voting-5f6b976c9f-krtsm
```
- The first commit prepares `util.BuildResource` to be able to deal with multiple resources.
- The second commit updates the `linkerd stat` cli help text.
- The third commit is the real implementation, consisting on firing parallel requests to the controller api, one for each resource. This stroke me as a much simpler approach than generalizing the controller api to accept multiple resources. I couldn't notice any performance degradation through preliminary testing.

Still pending is merging the tables in the results; currently each resource is producing a separate table even if they're of the same resource type. And adding more test cases for this of course :-) I'll wait for your feedback before I tackle those.

### DIFF
--- a/controller/api/util/api_utils.go
+++ b/controller/api/util/api_utils.go
@@ -255,7 +255,7 @@ func validateResources(resType string, args []string) error {
 		}
 	}
 	if len(set) < len(args) {
-		return errors.New("supplied duped resources")
+		return errors.New("cannot supply duplicate resources")
 	}
 	if all && len(args) > 1 {
 		return errors.New("'all' can't be supplied alongside other resources")
--- a/controller/api/util/api_utils.go
+++ b/controller/api/util/api_utils.go
@@ -205,27 +205,57 @@ func validateFromResourceType(resourceType string) (string, error) {
 
 // BuildResource parses input strings, typically from CLI flags, to build a
 // Resource object for use in the protobuf API.
-func BuildResource(namespace string, args ...string) (pb.Resource, error) {
+// It's the same as BuildResources but only admits one arg and only returns one resource
+func BuildResource(namespace, arg string) (pb.Resource, error) {
+	res, err := BuildResources(namespace, []string{arg})
+	return res[0], err
+}
+
+// BuildResources parses input strings, typically from CLI flags, to build a
+// slice of Resource objects for use in the protobuf API.
+// It's the same as BuildResource but it admits any number of args and returns multiple resources
+func BuildResources(namespace string, args []string) ([]pb.Resource, error) {
 	switch len(args) {
 	case 0:
-		return pb.Resource{}, errors.New("No resource arguments provided")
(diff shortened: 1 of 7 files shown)


grampelberg:
Ooooh, that's cool!

alpeb:
Awesome, thanks for the detailed feedback :1st_place_medal: 
I'll get back to you in a couple of days with all this addressed.

alpeb:
Alright, the next batch of changes is in :raised_hands: 
I had to make some refactorings to existing code along the way which I believe improved things.
Looking forward for your feedback :black_nib:

controller/api/util/api_utils.go
```diff
@@ -205,27 +205,78 @@ func validateFromResourceType(resourceType string) (string, error) {
...
+	set := make(map[string]bool)
+	all := false
+	for _, arg := range args {
+		set[arg] = true
+		if arg == k8s.All {
+			all = true
+		}
+	}
+	if len(set) < len(args) {
+		return errors.New("supplied duped resources")
```
klingerf: Hmm, I'd word this as "cannot supply duplicate resources".

pkg/k8s/k8s.go
```diff
@@ -22,6 +22,8 @@ const (
 	StatefulSet           = "statefulset"
 )
 
+var SortedRes = []string{Authority, DaemonSet, Deployment, Namespace, Pod, ReplicationController, ReplicaSet, Service, ServiceProfile, StatefulSet}
```
klingerf: How would you feel about calling this `AllResources`, and adding a comment that the contents should be sorted? I think that would make it clearer that this must be updated when we add new resources in the future. I'd also split this only multiple lines:

```
// AllResources is a sorted list of all resources defined as constants above.
var SortedRes = []string{
	Authority,
	DaemonSet,
	Deployment,
	...
```

controller/api/util/api_utils.go
alpeb: Agreed :+1:

pkg/k8s/k8s.go
alpeb: Yep, that'd be more explicit.

alpeb:
Thanks @klingerf for your review :ok_hand: 
I've just pushed your suggested changes :checkered_flag:

siggy:
FWIW this is an awesome change, thank you @alpeb !

This completes the last bit of #683, and relates to #911.

Playing with this a bit, realized we can now also do:
```bash
$ linkerd stat deploy/ po/
```

Issue #1745, opened 2018-10-08
Add a CLI command for displaying per-route stats


adleong:
Implement a CLI command which displays a live updating table of metrics for each route for a given resource.  This is very similar to the existing `linkerd top` command with some key differences:
* the underlying data is from Prometheus instead of from tap
* this data is only available if the resources have a service profile
* if the data is unavailable, a warning should be displayed telling the user to add a service profile or to use `linkerd tap table` instead

If `linkerd top` is renamed to `linkerd tap table` as described in https://github.com/linkerd/linkerd2/issues/1744 then this command could be called `linkerd top`.  Otherwise, it could be called `linkerd routes` or something similar.

This command should have a flag to add columns from source and destination resource and to split out the data accordingly.

adleong:
Fixed by #1860

PR #1843, opened 2018-11-06
Add routes command

adleong:
This PR defines the API that the CLI will use to get route stats data from Prometheus.  This will power the CLI command described in https://github.com/linkerd/linkerd2/issues/1745.

I am posting this PR early to share context.  I do not indent to merge this independently of the implementation which is currently blocked until both inbound and outbound route metrics are available in the proxy.


dadjeibaah:
Thanks for adding this @adleong! This is super helpful in thinking about the routes CLI command.

PR #1863, opened 2018-11-15
Add new iptable rule to for outbound traffic

### LINKED ISSUE #1585: Requests fail when src & dst are the same
Hello,

When making HTTP1.1 requests where the `src` and `dst` are the same (a pod sending a request to itself) the proxy responds with a 500 status code. If the request is sent to a different pod in the deployment, everything works fine. If you send the request on the loopback address rather than the service DNS name, that is also fine. Is this expected?

```
$ linkerd version
Client version: v18.8.4
Server version: v18.8.4
```
Using the default sidecar generated from `linkerd inject`

Thanks

dadjeibaah:
When requests from a pod send requests to itself, the proxy properly redirects traffic from the originating container in the pod through the outbound listener of the proxy. Once the request ends on the inbound side of the proxy, it skips the proxy and calls the original container that made the request. This can cause problems for containers that serve HTTP as the proxy naively tries to initiate an HTTP/2 connection to the destination of a request.  (See #1585 for a concrete example)

This PR adds a new iptable rule, coupled with a proxy [change](https://github.com/linkerd/linkerd2-proxy/pull/122) ensure that requests from a that occur in the aforementioned scenario, always redirect to the inbound listener of the proxy first.

fixes #1585

### DIFF
--- a/proxy-init/iptables/iptables.go
+++ b/proxy-init/iptables/iptables.go
@@ -82,7 +82,7 @@ func addOutgoingTrafficRules(commands []*exec.Cmd, firewallConfiguration Firewal
 	// Ignore traffic from the proxy
 	if firewallConfiguration.ProxyUid > 0 {
 		log.Printf("Ignoring uid %d", firewallConfiguration.ProxyUid)
-		// Redirect calls originating from from the proxy destined for a app container e.g. app -> proxy(outbound) -> proxy(inbound) -> app
+		// Redirect calls originating from the proxy destined for an app container e.g. app -> proxy(outbound) -> proxy(inbound) -> app
 		commands = append(commands, makeRedirectChainForOutgoingTraffic(outputChainName, redirectChainName, firewallConfiguration.ProxyUid,"redirect-non-loopback-local-traffic"))
 		commands = append(commands, makeIgnoreUserId(outputChainName, firewallConfiguration.ProxyUid, "ignore-proxy-user-id"))
 	} else {
--- a/proxy-init/iptables/iptables.go
+++ b/proxy-init/iptables/iptables.go
@@ -79,11 +79,11 @@ func addOutgoingTrafficRules(commands []*exec.Cmd, firewallConfiguration Firewal
 
 	commands = append(commands, makeCreateNewChain(outputChainName, "redirect-common-chain"))
 
-	// Redirect calls originating from to the proxy
-	commands = append(commands, makeRedirectChainForOutgoingTraffic(outputChainName, redirectChainName, "redirect-non-loopback-local-traffic"))
 	// Ignore traffic from the proxy
 	if firewallConfiguration.ProxyUid > 0 {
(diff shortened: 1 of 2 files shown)


proxy-init/iptables/iptables.go
```diff
@@ -241,6 +244,20 @@ func makeJumpFromChainToAnotherForAllProtocols(chainName string, targetChain str
 		"--comment", formatComment(comment))
 }
 
+func makeRedirectChainForOutgoingTraffic(chainName string, redirectChainName string, comment string) *exec.Cmd {
+	return exec.Command("iptables",
+		"-t", "nat",
+		"-A", chainName,
+		"-m", "owner",
+		"--uid-owner", "2102",
```
klingerf: This user ID is configurable when running `linkerd inject --proxy-uid 1234 -` -- unfortunately we can't hardcode it here. But the configured UID is passed into this process and stored in the `ProxyUid` field of the `FirewallConfiguration` struct in this file. So you should be able to use that. See my suggestion above.

proxy-init/iptables/iptables.go
```diff
@@ -73,12 +73,15 @@ func formatComment(text string) string {
...
 	outputChainName := "PROXY_INIT_OUTPUT"
+	redirectChainName := "PROXY_INIT_REDIRECT"
 	executeCommand(firewallConfiguration, makeFlushChain(outputChainName))
 	executeCommand(firewallConfiguration, makeDeleteChain(outputChainName))
 
 	commands = append(commands, makeCreateNewChain(outputChainName, "redirect-common-chain"))
 
-	// Ingore traffic from the proxy
+	// Redirect calls originating from to the proxy
+	commands = append(commands, makeRedirectChainForOutgoingTraffic(outputChainName, redirectChainName, "redirect-non-loopback-local-traffic"))
```
klingerf: Since this rule is tied to the proxy UID, we should handle it similarly to the `makeIgnoreUserId` rule on line 87. I'd move this rule to line 86, and then pass `firewallConfiguration.ProxyUid` as a param to `makeRedirectChainForOutgoingTraffic`.

proxy-init/iptables/iptables.go
```diff
@@ -73,14 +73,17 @@ func formatComment(text string) string {
...
 	executeCommand(firewallConfiguration, makeFlushChain(outputChainName))
 	executeCommand(firewallConfiguration, makeDeleteChain(outputChainName))
 
 	commands = append(commands, makeCreateNewChain(outputChainName, "redirect-common-chain"))
 
-	// Ingore traffic from the proxy
+	// Ignore traffic from the proxy
 	if firewallConfiguration.ProxyUid > 0 {
 		log.Printf("Ignoring uid %d", firewallConfiguration.ProxyUid)
+		// Redirect calls originating from from the proxy destined for a app container e.g. app -> proxy(outbound) -> proxy(inbound) -> app
```
klingerf: Super minor typo nits: "from from" => "from" and "a app" => "an app"

PR #1867, opened 2018-11-20
Add --open-api option to linkerd profiles command

adleong:
The `--open-api` flag is an alternative to the `--template` flag for the `linkerd profile` command.  It reads an OpenAPI specification file (also called a swagger file) and uses it to generate a corresponding service profile.

```
$ linkerd profile --open-api pet.swagger pet
kind: ServiceProfile
metadata:
  creationTimestamp: null
  name: pet.default.svc.cluster.local
  namespace: linkerd
spec:
  routes:
  - condition:
      method: GET
      path: ^/sample/pet/[^/]*$
    name: GET /sample/pet/{petId}
  - condition:
      method: POST
      path: ^/sample/pet$
    name: POST /sample/pet
```

### DIFF
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -3,12 +3,20 @@ package cmd
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"regexp"
 	"text/template"
 
+	"github.com/ghodss/yaml"
+	"github.com/go-openapi/spec"
 	"github.com/linkerd/linkerd2/cli/profile"
+	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
 	"github.com/spf13/cobra"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type templateConfig struct {
@@ -18,15 +26,21 @@ type templateConfig struct {
 	ClusterZone           string
 }
 
+var pathParamRegex = regexp.MustCompile(`\\{[^\}]*\\}`)
+
 type profileOptions struct {
+	name      string
 	namespace string
 	template  bool
+	openAPI   string
 }
 
 func newProfileOptions() *profileOptions {
 	return &profileOptions{
+		name:      "",
 		namespace: "default",
 		template:  false,
+		openAPI:   "",
 	}
 }
 
@@ -35,30 +49,55 @@ func newCmdProfile() *cobra.Command {
 	options := newProfileOptions()
 
 	cmd := &cobra.Command{
-		Use:   "profile [flags] --template (SERVICE)",
+		Use:   "profile [flags] (--template | --open-api file) (SERVICE)",
 		Short: "Output template service profile config for Kubernetes",
 		Long: `Output template service profile config for Kubernetes.
 		
-		This outputs a service profile template for the given service.  Edit the
-		template and then apply it with kubectl to add a service profile to a
-		service.
+		This outputs a service profile for the given service.
+		
+		If the --template flag is specified, it outputs a service profile template.
+		Edit the template and then apply it with kubectl to add a service profile to
+		a service.
 
 		Example:
 		linkerd profile -n emojivoto --template web-svc > web-svc-profile.yaml
 		# (edit web-svc-profile.yaml manually)
 		kubectl apply -f web-svc-profile.yaml
+
+		If the --open-api flag is specified, it reads the given OpenAPI
+		specification file and outputs a corresponding service profile.
+
+		Example:
+		linkerd profile -n emojivoto --open-api web-svc.swagger web-svc | kubectl apply -f -
 		`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if !options.template {
-				return errors.New("only template mode is currently supported, please run with --template")
+			options.name = args[0]
+
+			if !options.template && options.openAPI == "" {
+				return errors.New("You must specify exactly one of --template or --open-api")
 			}
 
-			return renderProfileTemplate(buildConfig(options.namespace, args[0]), os.Stdout)
+			if options.template {
+				if options.openAPI != "" {
+					return errors.New("You must specify exactly one of --template or --open-api")
+				}
+				return renderProfileTemplate(buildConfig(options.namespace, args[0]), os.Stdout)
+			}
+
+			if options.openAPI != "" {
+				if options.template {
+					return errors.New("You must specify exactly one of --template or --open-api")
+				}
+				return renderOpenAPI(options, os.Stdout)
+			}
+
+			return errors.New("You must specify exactly one of --template or --open-api")
 		},
 	}
 
 	cmd.PersistentFlags().BoolVar(&options.template, "template", options.template, "Output a service profile template")
+	cmd.PersistentFlags().StringVar(&options.openAPI, "open-api", options.openAPI, "Output a service profile based on the given OpenAPI spec file")
 	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace of the service")
 
 	return cmd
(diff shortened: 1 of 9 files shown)


cli/cmd/profile.go
```diff
@@ -87,3 +126,129 @@ func renderProfileTemplate(config *templateConfig, w io.Writer) error {
...
+			spec := mkRouteSpec(path, pathRegex, http.MethodPost, item.Post.Responses)
+			routes = append(routes, spec)
+		}
+		if item.Put != nil {
+			spec := mkRouteSpec(path, pathRegex, http.MethodPut, item.Put.Responses)
+			routes = append(routes, spec)
+		}
+	}
+
+	profile.Spec.Routes = routes
```
siggy: i think it's useful to make the output of `profile --open-api` commands stable. the `swagger.Paths.Paths` map above will be intentionally randomized by Go. you can stabilize it here:
```go
sort.Slice(routes, func(i, j int) bool {
	return routes[i].Name < routes[j].Name
})
```

cli/cmd/profile.go
```diff
@@ -87,3 +126,129 @@ func renderProfileTemplate(config *templateConfig, w io.Writer) error {
...
+		failure := false
+		if status >= 500 {
+			failure = true
+		}
+		classes = append(classes, &sp.ResponseClass{
+			Condition: cond,
+			IsFailure: !failure,
+		})
+	}
+	return classes
```
siggy: similarly for output stability:
```go
sort.Slice(classes, func(i, j int) bool {
	return classes[i].Condition.Status.Min < classes[j].Condition.Status.Min
})
```
we could maybe go a bit farther than this and sort on `IsFailure` and `Max`, but this worked with my limited testing:
```bash
diff <(bin/go-run cli profile --open-api <(curl https://petstore.swagger.io/v2/swagger.json) pet) <(bin/go-run cli profile --open-api <(curl https://petstore.swagger.io/v2/swagger.json) pet)
```

cli/cmd/profile.go
```diff
@@ -35,30 +49,55 @@ func newCmdProfile() *cobra.Command {
...
+		linkerd profile -n emojivoto --open-api web-svc.swagger web-svc | kubectl apply -f -
 		`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if !options.template {
-				return errors.New("only template mode is currently supported, please run with --template")
+			options.name = args[0]
+
+			if !options.template && options.openAPI == "" {
+				return errors.New("You must specify exactly one of --template or --open-api")
```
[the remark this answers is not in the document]
adleong: I think the first return I had was redundant so I removed it.  But generally I wanted to strucutre this in an extendable way where more mutually exclusive options can be added.  Each option checks that no other options are set, and if no options are set, the command fails via the final return.

cli/cmd/profile.go
```diff
@@ -87,3 +126,129 @@ func renderProfileTemplate(config *templateConfig, w io.Writer) error {
...
+		TypeMeta: meta_v1.TypeMeta{
+			APIVersion: "linkerd.io/v1alpha1",
+			Kind:       "ServiceProfile",
+		},
+	}
+
+	routes := make([]*sp.RouteSpec, 0)
+	for path, item := range swagger.Paths.Paths {
+		pathRegex := pathToRegex(path)
+		if item.Delete != nil {
```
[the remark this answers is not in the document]
adleong: I don't think we need it for anything.  All we're really doing is replacing the path params with regex wildcards

cli/cmd/profile.go
```diff
@@ -87,3 +126,129 @@ func renderProfileTemplate(config *templateConfig, w io.Writer) error {
...
+	classes := make([]*sp.ResponseClass, 0)
+	for status := range responses.StatusCodeResponses {
+		cond := &sp.ResponseMatch{
+			Status: &sp.Range{
+				Min: uint32(status),
+				Max: uint32(status),
+			},
+		}
+		failure := false
+		if status >= 500 {
```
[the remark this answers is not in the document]
adleong: I like 500 because then you don't have to remember that InternalServerError is the lowest 5xx error.

cli/cmd/profile.go
adleong: I went with sorting the path keys prior to constructing routes

cli/cmd/profile.go
adleong: I wend with sorting the status keys prior to constructing response classes

cli/cmd/profile.go
```diff
@@ -35,30 +50,51 @@ func newCmdProfile() *cobra.Command {
...
 	cmd := &cobra.Command{
-		Use:   "profile [flags] --template (SERVICE)",
+		Use:   "profile [flags] (--template | --open-api file) (SERVICE)",
 		Short: "Output template service profile config for Kubernetes",
 		Long: `Output template service profile config for Kubernetes.
 		
-		This outputs a service profile template for the given service.  Edit the
-		template and then apply it with kubectl to add a service profile to a
-		service.
+		This outputs a service profile for the given service.
```
klingerf: Not directly related to your change, but can you un-indent this entire string block? Otherwise the help text is weirdly indented when printed by the CLI:

```
Output template service profile config for Kubernetes.
		
		This outputs a service profile for the given service.
		
		If the --template flag is specified, it outputs a service profile template.
...
```

See some of the other files in this directory for an example of how they handle the long description.

cli/cmd/profile.go
```diff
@@ -87,3 +123,144 @@ func renderProfileTemplate(config *templateConfig, w io.Writer) error {
...
+				Max: uint32(status),
+			},
+		}
+		failure := false
+		if status >= 500 {
+			failure = true
+		}
+		classes = append(classes, &sp.ResponseClass{
+			Condition: cond,
+			IsFailure: !failure,
```
klingerf: Isn't this the inverse of what we want? If I use this example swagger definition:

```
swagger: "2.0"
info:
  title: Sample API
  description: API description in Markdown.
  version: 1.0.0
host: api.example.com
basePath: /v1
schemes:
  - https
paths:
  /users:
    get:
      summary: Returns a list of users.
      description: Optional extended description in Markdown.
      produces:
        - application/json
      responses:
        200:
          description: OK
```

Then I end up with:

```
    response_classes:
    - condition:
        status:
          max: 200
          min: 200
      is_failure: true
```

But shouldn't that be `is_failure: false`?

klingerf: Also, I think this would be easier to reason about if you condensed it a bit:

```diff
diff --git a/cli/cmd/profile.go b/cli/cmd/profile.go
index 391966ef..6ccce642 100644
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -253,13 +253,9 @@ func toRspClasses(responses *spec.Responses) []*sp.ResponseClass {
 				Max: uint32(status),
 			},
 		}
-		failure := false
-		if status >= 500 {
-			failure = true
-		}
 		classes = append(classes, &sp.ResponseClass{
 			Condition: cond,
-			IsFailure: !failure,
+			IsFailure: status >= 500,
 		})
 	}
 	return classes
```

cli/cmd/profile.go
```diff
@@ -3,12 +3,21 @@ package cmd
...
 	"io"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"regexp"
+	"sort"
 	"text/template"
 
+	"github.com/ghodss/yaml"
+	"github.com/go-openapi/spec"
```
klingerf: Looks like this library only supports OpenAPI v2. Maybe the CLI should abort with an error if you pass it a v3 spec?

adleong: Hmmm... yeah the readme says v2 but all my testing has been with v3 specs.  🤷‍♂️

cli/cmd/profile.go
```diff
@@ -35,30 +50,50 @@ func newCmdProfile() *cobra.Command {
 	options := newProfileOptions()
 
 	cmd := &cobra.Command{
-		Use:   "profile [flags] --template (SERVICE)",
+		Use:   "profile [flags] (--template | --open-api file) (SERVICE)",
 		Short: "Output template service profile config for Kubernetes",
 		Long: `Output template service profile config for Kubernetes.
```
klingerf: Maybe it's worth tweaking these short descriptions to not include the word "template" now? Especially for the long description, this help output ready a bit awkwardly to me:

```
$ bin/go-run cli help profile
Output template service profile config for Kubernetes.
		
This outputs a service profile for the given service.
		
If the --template flag is specified, ...
```

Issue #1737, opened 2018-10-03
Per-route historical metrics


grampelberg:
Linkerd has metrics at the service level. These are valuable for understanding the general health of a service. Unfortunately, it is usually a specific route that is having problems. While it is possible to run `top` and get some metrics today, these are gathered in real time and not stored in prometheus.

It should be possible to see per-route metrics alongside the existing per-service metrics (success rate, latency, throughput). These can improve the time to fixing issues and provide visibility into what is really happening. 

## User Stories

- As a service owner, I would like to see per-route metrics in my dashboards so that I can quickly see any endpoints that are operating outside my SLO.
- As a service owner, I would like to see a list of all the routes in my service and sort that list by success rate, so that I can quickly see what is currently failing.
- As a service owner, I would like to have per-route metrics aggregated by URL parameters such as user id, so that I can quickly see what code path is being taken.
- As a service owner, I would like to persist per-route metrics so that I can use them to debug historical issues.

klingerf:
Related to #187.

grampelberg:
Also related #1418

ctaggart:
Any update on per-route metrics? Is it stubbed out in the UI and not implemented on the backend as of linkerd edge 18.11.2 as it [appears here](https://github.com/linkerd/linkerd2/issues/1451#issuecomment-442258315)?

grampelberg:
@ctaggart it is on master (so it'll be in the next edge release). I believe most of the stuff you need is in 18.11.2 but the CLI pieces to actually see what is going on didn't land until earlier this week.

PR #1868, opened 2018-11-22
Add the top routes feature to the dashboard UI

rmars:
Adds a (currently not displayed in sidebar, but available at `/routes`) page to mirror the current functionality of `linkerd routes`. So far, this is just a barebones form and table, but it works.

Adds a `api/routes` path and handler to receive `TopRoutes` requests from the web. (for example, run the books app and add a service profile, and go to `http://localhost:8084/api/routes?resource_name=webapp&namespace=default&from_type=deploy&from_namespace=default`)

Follow up branches will include:
- a reasonable error message if there is no data, asking the user to create a service profile
- form validation
- service and namespace autocompletion

screen shot 2018-11-21 at 4 11 08 pm

Relates to  #1737

### DIFF
--- a/web/app/js/components/TopRoutesTable.jsx
+++ b/web/app/js/components/TopRoutesTable.jsx
@@ -3,16 +3,12 @@ import BaseTable from './BaseTable.jsx';
 import PropTypes from 'prop-types';
 import React from 'react';
 import SuccessRateMiniChart from './util/SuccessRateMiniChart.jsx';
-import _ from 'lodash';
 
 const routesColumns = [
   {
     title: "Route",
     dataIndex: "route",
-    render: d => {
-      return _.isEmpty(d.route) ? "UNKNOWN" : d.route;
-    },
-    sorter: (a, b) => (a.route || "").localeCompare(b.route)
+    sorter: (a, b) => (a.route).localeCompare(b.route)
   },
   {
     title: "Success Rate",
--- a/web/app/js/components/TopRoutesTable.jsx
+++ b/web/app/js/components/TopRoutesTable.jsx
@@ -1,0 +1,75 @@
+import { metricToFormatter, numericSort } from './util/Utils.js';
+import BaseTable from './BaseTable.jsx';
+import PropTypes from 'prop-types';
+import React from 'react';
+import SuccessRateMiniChart from './util/SuccessRateMiniChart.jsx';
+import _ from 'lodash';
+
+const routesColumns = [
+  {
+    title: "Route",
+    dataIndex: "route",
+    render: d => {
+      return _.isEmpty(d.route) ? "UNKNOWN" : d.route;
+    },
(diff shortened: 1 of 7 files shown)


wmorgan:
This is awesome to see. I think we should make a call on whether this goes as a separate page or whether it replaces the top table on the service page. cc @grampelberg

grampelberg:
@adleong what do you think?

Having the real time data there is valuable for any service that doesn't have a profile (external APIs for example). However, the services with profiles will have higher quality data.

1. Replace the stream with profile data. I'm concerned this will end up hiding most service's data. It would end up hiding any client data that doesn't have a profile (and any server data that doesn't have a profile). Without any data on how many folks would add profiles to their own (and other's) services, this feels like a big step backwards.

1. Show the stream when there isn't a profile, show the profile data when there is one. Feels a little bit better than the previous option. It doesn't solve the client profile issue and will end up changing the look and feel of the table (columns, updates ...).

1. Show both and let users decide. I think this is the best option of the bunch. We can hide the table for profile data if there isn't one associated and stick the stream data below it.

@adleong this has got me thinking about a separate problem. Today, we do a good job of showing all the server and client traffic (with the stream). What are we doing to show that some data isn't being captured by a profile and show that you can improve it? I'd hate to just silently drop it ....

adleong:
@grampelberg I think I agree with you that option 3 seems the most desirable.  Perhaps with a "live" tab and a  "metrics" tab, or something like that.  Defaulting to showing whichever tab has better data.

I'm not sure exactly what you mean by data that isn't captured by a profile.  If there are requests which don't match any of the routes in a profile, they get assigned to the "default route" which shows up as [UNKNOWN].

wmorgan:
I'm ok with option #3 if that doesn't overly complicate @rmars's life. One tab for live data, one tab for metrics, and the metrics ta can show a message about service profiles if they're not there yet. Something like that?

rmars:
Yeah, 3 sounds fine, and is probably the easiest for now anyway. Tabs could be useful, as the page is getting long anyways!

Note that the implementation in this PR just puts it out there as a separate page (like /tap and /top) and is not currently findable in the UI. We can modularize this like top and insert it into the resource detail pages as well!

web/app/js/components/TopRoutesTable.jsx
```diff
@@ -0,0 +1,75 @@
...
+import _ from 'lodash';
+
+const routesColumns = [
+  {
+    title: "Route",
+    dataIndex: "route",
+    render: d => {
+      return _.isEmpty(d.route) ? "UNKNOWN" : d.route;
+    },
+    sorter: (a, b) => (a.route || "").localeCompare(b.route)
```
siggy: This sorts `UNKNOWN` first by default. Any way to sort it to the bottom?

screen shot 2018-11-27 at 3 43 02 pm

rmars: fixed!

Issue #1022, opened 2018-05-25
Add an option to not request stats from prometheus from the stat API


rmars:
Or provide a purely resource-listing endpoint.

Some components of the web UI just want a list of resources (e.g. the Sidebar requires a list of namespaces) without any stat information. 
Currently we get that list for the Sidebar by querying the Stat API, but we don't actually need any of the prometheus stats from this call, we just want a list of resource names. It would be nice to set a flag so that the stat endpoint knows it doesn't have to hit prometheus for these calls.

If the Stat API were to be modified to accept requests for specific metrics, we could do this via passing in metric: none or something like that.

PR #1871, opened 2018-11-22
Add parameter to stats API to skip retrieving Prometheus stats

alpeb:
This new parameter (defaults to false) avoids fetching stats from Prometheus, that are not used in some cases.

So currently this request
`/api/tps-reports?resource_type=namespace&all_namespaces=true&window=1m`
returns something like
```
{                                   
   "ok" : {                                 
      "statTables" : [                      
         {                               
            "podGroup" : {              
               "rows" : [                    
                  {                
                     "resource" : {      
                        "name" : "linkerd",    
                        "type" : "namespace",
                        "namespace" : ""
                     },
                     "timeWindow" : "1m",
                     "errorsByPod" : {},
                     "runningPodCount" : "4", 
                     "stats" : {         
                        "successCount" : "1626",
                        "latencyMsP50" : "43", 
                        "latencyMsP99" : "99",
                        "latencyMsP95" : "94",
                        "failureCount" : "0",
                        "tlsRequestCount" : "0"
                     },                     
                     "failedPodCount" : "0",
                     "meshedPodCount" : "4"
                  },
                 ...
```
And now with the extra param:
`/api/tps-reports?resource_type=namespace&all_namespaces=true&skip_stats=true&window=1m`
it returns something like
```
{
   "ok" : {
      "statTables" : [
         {
            "podGroup" : {
               "rows" : [
                 {
                     "runningPodCount" : "4",
                     "errorsByPod" : {},
                     "timeWindow" : "1m",
                     "stats" : null,
                     "failedPodCount" : "0",
                     "resource" : {
                        "type" : "namespace",
                        "name" : "linkerd",
                        "namespace" : ""
                     },
                     "meshedPodCount" : "4"
                  },
                 ...
```

@rmars please let me know if this works out for #1022, before I proceed adding some tests.


proto/public.proto
```diff
@@ -294,6 +294,8 @@ message StatSummaryRequest {
     Resource to_resource   = 4;
     Resource from_resource = 5;
   }
+
+  bool skip_stats = 6;  // true if we want to skip stats from Prometheus (used in the dashboard sidebar)
```
rmars: I think we can remove `(used in the dashboard sidebar)` from the comment, since the comments in the proto tend to be descriptions of what the fields signify

alpeb:
Thanks @rmars for the review :+1: 
I've pushed a new test, which required some preliminary changes/fixes in the test helpers.
Prometheus queries being generated were actually not checked, so I fixed that. Also we were not checking when actually _no_ queries should be generated. When I added support for that it turns out a previous test case was generating queries when it shouldn't. I didn't fix that one, but I flagged it in the code.
Let me know what you think :eyes:

alpeb:
Regarding my previous comment, those Prometheus queries were actually expected, so I added them, and it all makes sense now :musical_note:

controller/api/util/api_utils.go
```diff
@@ -143,6 +144,7 @@ func BuildStatSummaryRequest(p StatsRequestParams) (*pb.StatSummaryRequest, erro
 			},
 		},
 		TimeWindow: window,
+		SkipStats:  p.SkipStats,
```
rmars: we've since made some slight adjustments to this in master; we've now split this into `StatsSummaryRequestParams` and `StatsBaseRequestParams` - I'd recommend putting SkipStats in `StatsSummaryRequestParams`

rmars:
@alpeb sorry about the delay in getting back to you on this! I think if you merge master and resolve the conflicts this should be good to go!

controller/api/util/api_utils.go
alpeb: Sounds good :+1:

PR #1873, opened 2018-11-26
tap: Use nil-safe protobuf accessors

olix0r:
The tap server accesses protobuf fields directly instead of using the
`Get*()` accessors. The accessors are necessary to prevent dereferencing
a nil pointer and crashing the tap service.

### DIFF
--- a/controller/tap/server.go
+++ b/controller/tap/server.go
@@ -482,25 +482,29 @@ func indexPodByIP(obj interface{}) ([]string, error) {
 // Since errors encountered while hydrating metadata are non-fatal and result
 // only in missing labels, any errors are logged at the WARN level.
 func (s *server) hydrateEventLabels(ev *public.TapEvent) {
-	labels := ev.GetSourceMeta().GetLabels()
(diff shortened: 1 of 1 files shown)


controller/tap/server.go
```diff
@@ -482,16 +482,25 @@ func indexPodByIP(obj interface{}) ([]string, error) {
 // Since errors encountered while hydrating metadata are non-fatal and result
 // only in missing labels, any errors are logged at the WARN level.
 func (s *server) hydrateEventLabels(ev *public.TapEvent) {
-	err := s.hydrateIPLabels(ev.Source.Ip, ev.SourceMeta.Labels)
+	labels := ev.GetSourceMeta().GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
```
adleong: Does this work?  I would expect that this creates an empty map but doesn't set it into the source meta struct.

olix0r: oh _that's_ what this is doing.  Yes, good catch. I'll try again.

PR #1876, opened 2018-11-27
Add ListServices to public api

rmars:
Add a barebones `ListServices` endpoint, in support of autocomplete for services.
As we develop service profiles, this endpoint could probably be used to describe more aspects of services (like, if there were some way to check whether a service profile was enabled or not).

Accessible from the web UI via `http://localhost:8084/api/services`

Note that we could probably hack the existing stat api (especially with skip_stats being added in #1871) to return service information, but we'd need to request `resource_type=all&skip_stats=true`. Also, because stats on services don't make sense for outbound, we have restrictions on that api, which means we won't get a list of all possible services. I'm open to changing the implementation of this, though.

```
$ curl http://localhost:8084/api/services?namespace=default 
{
  "services": [
    {
      "name": "webapp",
      "namespace": "default"
    },
    {
      "name": "mysql",
      "namespace": "default"
    },
    {
      "name": "authors",
      "namespace": "default"
    },
    {
      "name": "books",
      "namespace": "default"
    },
    {
      "name": "kubernetes",
      "namespace": "default"
    }
  ]
}
```


controller/api/public/grpc_server_test.go
```diff
@@ -19,13 +19,26 @@ type listPodsExpected struct {
 	res     pb.ListPodsResponse
 }
 
+type listServicesExpected struct {
```
siggy: 🙏

controller/api/public/grpc_server_test.go
```diff
@@ -174,3 +187,84 @@ spec:
...
+
+		if aSvc.Name != bSvc.Name || aSvc.Namespace != bSvc.Namespace {
+			return false
+		}
+	}
+
+	return true
+}
+func TestListServices(t *testing.T) {
+	t.Run("Successfully performs a query based on resource type", func(t *testing.T) {
```
klingerf: Super tiny nit: I think this string was copied from `TestListPods`, but it's not quite right in this context. Maybe it could be "Successfully queries for services" or something similar?

rmars: ah yeah totally!

PR #1886, opened 2018-11-28
Modularize TopRoutes, add autocomplete for services

rmars:
Separates out the querying and table display of route data, so that this module can be easily placed in other places in the UI.
Adds usability improvements to the routes query form at `/routes`:
- displays CLI equivalent
- adds dropdown with populated options for service / namespace

As part of this work, made TapQueryCliCmd more generic, so it can work for other CLI commands besides tap/top. 

screen shot 2018-11-28 at 2 06 00 pm


web/app/js/components/TopRoutes.jsx
```diff
@@ -113,49 +112,102 @@ class TopRoutes extends React.Component {
...
+            <Grid item xs={6} md={3}>
+              { this.renderServiceDropdown() }
+            </Grid>
           </Grid>
-          <Grid item xs={6} md={3}>
-            { this.renderTextInput("From namespace", "from_namespace", "Resource namespace") }
+
+          <Grid item container spacing={8} alignItems="center">
+            <Grid item xs={6} md={3}>
+              { this.renderTextInput("From", "from_name", "Resource name") }
```
siggy: could/should these 3 text inputs be reworked into 2 fields, `From Namespace` and `From Resource`, similar tap/top forms? also, add autocomplete?

screen shot 2018-11-29 at 10 03 00 am

rmars: yes I think so! will delegate that to a polish followup

rmars: filed #1894

PR #1893, opened 2018-11-29
Add ability to download a service profile template from the web ui

rmars:
Adds an endpoint, at `/profiles/new` that allows you to input a service name and namespace, and download a service profile yaml template. This will enable future work, where we can add more of the yaml customization via a form in the dashboard, and use that data to help the user configure routes.

To test, you can go to, for example, `http://localhost:8084/profiles/new?service=foo&namespace=bar`


web/srv/server.go
```diff
@@ -111,6 +111,7 @@ func NewServer(addr, templateDir, staticDir, uuid, controllerNamespace, webpackD
 	server.router.GET("/api/services", handler.handleApiServices)
 	server.router.GET("/api/tap", handler.handleApiTap)
 	server.router.GET("/api/routes", handler.handleApiTopRoutes)
+	server.router.GET("/profiles/new", handler.handleProfileDownload)
```
klingerf: Would you mind moving this to the "webapp routes" section above? I don't think it belongs here in the "webapp api routes" section, since it's not a /api route.

web/srv/api_handlers.go
```diff
@@ -242,3 +246,39 @@ func (h *handler) handleApiTap(w http.ResponseWriter, req *http.Request, p httpr
...
 }
+
+func (h *handler) handleProfileDownload(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+	service := req.FormValue("service")
+	namespace := req.FormValue("namespace")
+
+	if service == "" || namespace == "" {
+		err := fmt.Errorf("Service and namespace must be provided to create a new profile")
+		log.Error(err.Error())
+		renderJsonError(w, err, http.StatusBadRequest)
```
klingerf: I wouldn't expect these error messages to be rendered as JSON, since this endpoint doesn't return JSON for successful responses. Can you instead do something like:

```suggestion
		http.Error(w, err.Error(), http.StatusBadRequest)
```

web/srv/api_handlers.go
```diff
@@ -242,3 +246,39 @@ func (h *handler) handleApiTap(w http.ResponseWriter, req *http.Request, p httpr
 		}
 	}
 }
+
+func (h *handler) handleProfileDownload(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
```
klingerf: Related to my comment about the route below, maybe this should be defined in `web/srv/handlers.go` instead?

web/srv/api_handlers.go
```diff
@@ -3,14 +3,18 @@ package srv
...
 	"io"
 	"net/http"
+	"text/template"
 	"time"
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	"github.com/gorilla/websocket"
 	"github.com/julienschmidt/httprouter"
+	profileCmd "github.com/linkerd/linkerd2/cli/cmd"
```
klingerf: I'd like to avoid having this package import the `cli/cmd` package. My preference is to move the code that's shared between the CLI and the web server to a separate package in the `pkg` directory.

In fact, as written the web docker image won't build, since we don't add the `cli` directory before building the web binary. All code that the web server depends on needs to live in `web/srv`, `controller`, `pkg`, or `vendor`.

web/srv/api_handlers.go
```diff
@@ -242,3 +246,39 @@ func (h *handler) handleApiTap(w http.ResponseWriter, req *http.Request, p httpr
...
+		return
+	}
+
+	profileYaml := &bytes.Buffer{}
+	err = template.Execute(profileYaml, tmplParams)
+	if err != nil {
+		log.Error(err.Error())
+		renderJsonError(w, err, http.StatusInternalServerError)
+		return
+	}
```
klingerf: You could reuse more of the code from `cli/cmd/profile.go` if you do something like:

```diff
diff --git a/cli/cmd/profile.go b/cli/cmd/profile.go
index 3b1a1330..3123ab9e 100644
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -54,7 +54,7 @@ func newCmdProfile() *cobra.Command {
 				return errors.New("only template mode is currently supported, please run with --template")
 			}
 
-			return renderProfileTemplate(BuildConfig(options.namespace, args[0]), os.Stdout)
+			return RenderProfileTemplate(BuildConfig(options.namespace, args[0]), os.Stdout)
 		},
 	}
 
@@ -73,7 +73,7 @@ func BuildConfig(namespace, service string) *ProfileTemplateConfig {
 	}
 }
 
-func renderProfileTemplate(config *ProfileTemplateConfig, w io.Writer) error {
+func RenderProfileTemplate(config *ProfileTemplateConfig, w io.Writer) error {
 	template, err := template.New("profile").Parse(profile.Template)
 	if err != nil {
 		return err
diff --git a/web/srv/api_handlers.go b/web/srv/api_handlers.go
index 84a18829..1194f757 100644
--- a/web/srv/api_handlers.go
+++ b/web/srv/api_handlers.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"text/template"
 	"time"
 
 	"github.com/golang/protobuf/jsonpb"
@@ -14,7 +13,6 @@ import (
 	"github.com/gorilla/websocket"
 	"github.com/julienschmidt/httprouter"
 	profileCmd "github.com/linkerd/linkerd2/cli/cmd"
-	"github.com/linkerd/linkerd2/cli/profile"
 	"github.com/linkerd/linkerd2/controller/api/util"
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/k8s"
@@ -259,19 +257,12 @@ func (h *handler) handleProfileDownload(w http.ResponseWriter, req *http.Request
 	}
 
 	tmplParams := profileCmd.BuildConfig(namespace, service)
-
-	template, err := template.New("profile").Parse(profile.Template)
-	if err != nil {
-		log.Error(err.Error())
-		renderJsonError(w, err, http.StatusInternalServerError)
-		return
-	}
-
 	profileYaml := &bytes.Buffer{}
-	err = template.Execute(profileYaml, tmplParams)
+
+	err := profileCmd.RenderProfileTemplate(tmplParams, profileYaml)
 	if err != nil {
 		log.Error(err.Error())
-		renderJsonError(w, err, http.StatusInternalServerError)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
```

(you'd want to move that `RenderProfileTemplate` func to somewhere in `pkg` though, per my other suggestion)

web/srv/api_handlers.go
```diff
@@ -242,3 +246,39 @@ func (h *handler) handleApiTap(w http.ResponseWriter, req *http.Request, p httpr
...
 	}
 }
+
+func (h *handler) handleProfileDownload(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+	service := req.FormValue("service")
+	namespace := req.FormValue("namespace")
+
+	if service == "" || namespace == "" {
+		err := fmt.Errorf("Service and namespace must be provided to create a new profile")
+		log.Error(err.Error())
```
siggy: i realize we're not consistent about this in the codebase, but `log.Error(err)` will do the same thing here.

cli/cmd/test_helpers.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/ghodss/yaml"
+	"github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func GenServiceProfile(service, namespace string) v1alpha1.ServiceProfile {
```
siggy: 🙏

cli/cmd/profile.go
```diff
@@ -54,7 +54,7 @@ func newCmdProfile() *cobra.Command {
 				return errors.New("only template mode is currently supported, please run with --template")
 			}
 
-			return renderProfileTemplate(buildConfig(options.namespace, args[0]), os.Stdout)
+			return renderProfileTemplate(BuildConfig(options.namespace, args[0]), os.Stdout)
```
siggy: fyi this may get conflicty with #1867

pkg/profiles/profiles.go
```diff
@@ -298,3 +308,27 @@ func ValidateResponseMatch(rspMatch *sp.ResponseMatch) error {
 
 	return nil
 }
+
+func BuildConfig(namespace, service, controlPlaneNamespace string) *ProfileTemplateConfig {
```
siggy: sorry i didn't catch this earlier... since `BuildConfig` is only ever called immediately before `RenderProfileTemplate`, consider removing it, and also making `ProfileTemplateConfig` private...

```go
func RenderProfileTemplate(namespace, service, controlPlaneNamespace string, w io.Writer) error {
	config := &profileTemplateConfig{
		ControlPlaneNamespace: controlPlaneNamespace,
		ServiceNamespace:      namespace,
		ServiceName:           service,
		ClusterZone:           "svc.cluster.local",
	}

	template, err := template.New("profile").Parse(Template)
	if err != nil {
		return err
	}
	buf := &bytes.Buffer{}
	err = template.Execute(buf, config)
	if err != nil {
		return err
	}

	_, err = w.Write(buf.Bytes())
	return err
}
```

rmars: do you think it's worth keeping them separate so we can integrate validation into the `BuildConfig` function, and only pass in validated template params to `RenderProfileTemplate`? That would be the only reason I would have for not integrating it

siggy: i think if we found that building/validating the template struct became complex enough that it warranted its own function (and testing), it would make sense to separate it out. until we do those things though, i'd tend to keep the code simpler. all that said, i don't feel too strongly here, totally ok to ship either way.

rmars: done!

cli/cmd/profile.go
```diff
@@ -78,7 +76,7 @@ Example:
 				if options.openAPI != "" {
 					return errors.New("You must specify exactly one of --template or --open-api")
 				}
-				return renderProfileTemplate(buildConfig(options.namespace, options.name), os.Stdout)
+				return profiles.RenderProfileTemplate(options.namespace, args[0], controlPlaneNamespace, os.Stdout)
```
klingerf: TIOLI -- the `options.name` field was added in #1867, so you can use that here instead of `args[0]`.

pkg/profiles/template.go
```diff
@@ -50,7 +50,6 @@ spec:
         status:
           min: 500
           max: 599
-
```
klingerf: Inadvertent newline removal?

web/srv/handlers_test.go
```diff
@@ -7,8 +7,11 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/ghodss/yaml"
 	"github.com/julienschmidt/httprouter"
+	helpers "github.com/linkerd/linkerd2/cli/cmd"
```
klingerf: This still introduces a dependency from the `web/srv` package to the `cli/cmd`, which I'd like to avoid, even if it's only for tests. Can you instead move the `cli/cmd/test_helpers.go` file that you're adding to `pkg/profiles/test_helper.go`?

Also, I'd call to `test_helper.go` (singular), instead of `test_helpers.go` (plural). Right now this branch is introducing a new file called `cli/cmd/test_helpers.go`, and `cli/cmd/test_helper.go` already exists.

rmars: ah yes, good point! thanks!

cli/cmd/profile_test.go
```diff
@@ -2,20 +2,17 @@ package cmd
 
 import (
 	"bytes"
-	"reflect"
 	"testing"
 
 	"github.com/ghodss/yaml"
 	"github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	profiles "github.com/linkerd/linkerd2/pkg/profiles"
```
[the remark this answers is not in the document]
rmars: oops good catch!!!

Issue #1747, opened 2018-10-09
Grafana pod in CrashLoopBackOff periodically


chzbrgr71:
### What is the issue?
Grafana pod in CrashLoopBackOff periodically. If I delete the pod, it will start working, but eventually crashes again.

### How can it be reproduced?
Running Linkerd stable-2.0.0
After a period of time, Grafana fails and I can only get it working again by deleting the pod.
This has happened consistently with multiple clusters in my demo environment. 

### Logs, error output, etc
Looks like the grafana liveliness probe failed:
```
Events:
  Type     Reason     Age               From                               Message
  ----     ------     ----              ----                               -------
  Warning  Unhealthy  5m (x9 over 6m)   kubelet, aks-nodepool1-34340261-2  Liveness probe failed: Get http://10.244.3.4:3000/api/health: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

I see these errors on the linkerd-proxy container in the grafana pod:
```
ERR! proxy={server=in listen=0.0.0.0:4143 remote=10.244.2.4:36412} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111)
```

#### `linkerd check` output

```text
kubernetes-api: can initialize the client..................................[ok]
kubernetes-api: can query the Kubernetes API...............................[ok]
kubernetes-api: is running the minimum Kubernetes API version..............[ok]
linkerd-api: control plane namespace exists................................[ok]
linkerd-api: control plane pods are ready..................................[retry] -- The "grafana" pod's "grafana" container is not ready
...etc.
```

### Environment

- Kubernetes Version: 1.10.7
- Cluster Environment: AKS
- Host OS: ubuntu
- Linkerd version: stable-2.0.0

grampelberg:
@chzbrgr71 are you running with TLS turned on?

chzbrgr71:
Yes. This cluster is setup with TLS on. I think the other clusters did as well, but not sure.

grampelberg:
Hmmm, I've not been able to replicate this yet with:

- AKS cluster @ 1.10.7, 1.11.3
- `linkerd install --tls=optional`
- emojivoto
- Viewing deploy/emoji in grafana

Any other tips?

chzbrgr71:
I have a potential clue as I am about to demo this in front of a large room. :-) 

[...]

PR #1168, opened 2018-06-20
Add controller admin servers and readiness probes

klingerf:
This branch introduces a new package, `pkg/admin`, which our go processes can use to run an admin server. The admin server has the following three endpoints:

* `/metrics` -- prometheus metrics
* `/ping` -- liveliness check
* `/ready` -- readiness check

The `/ping` endpoint is a no-op at this point, but we might want to make that more sophisticated as part of #1116.

For controller components that sync with Kubernetes, the `/ready` endpoint doesn't return ready until the caches are fully synced.

I've also updated our install configs to configure the checks accordingly.

Fixes #1013.

PR #1121, opened 2018-06-14
Add readiness/liveness checks for third party components

grampelberg:
Any possible issues with the third party control plane components can wedge the services.

Take the best practices for prometheus/grafana and add them to our template. See #1116

PR #1899, opened 2018-11-30
Adjust proxy, Prometheus, and Grafana probes

### LINKED ISSUE #1804: Adjust controller and proxy livenessProbe and readinessProbe
The values for `livenessProbe` and `readinessProbe` set by `linkerd install` and `linkerd inject` are more permissive than Kubernetes defaults. While this increases the likelihood all pods will come up healthy without pod restart, it also increases startup time, and may paper over other issues related to slow startup.

## Default probe values

```yaml
livenessProbe:
  initialDelaySeconds: 0
  periodSeconds: 10
  timeoutSeconds: 1
  successThreshold: 1
  failureThreshold: 3
readinessProbe:
  initialDelaySeconds: 0
  periodSeconds: 10
  timeoutSeconds: 1
  successThreshold: 1
  failureThreshold: 3
```

### grafana

```yaml
readinessProbe:
  failureThreshold: 10
  initialDelaySeconds: 30
  periodSeconds: 10
  timeoutSeconds: 30
```

### prometheus

```yaml
livenessProbe:
  initialDelaySeconds: 30
  timeoutSeconds: 30
readinessProbe:
  initialDelaySeconds: 30
  timeoutSeconds: 30
```

### public-api, destination, proxy-api, tap

```yaml
livenessProbe:
  initialDelaySeconds: 10
readinessProbe:
  failureThreshold: 7
```

### proxy
```yaml
livenessProbe:
  initialDelaySeconds: 10
readinessProbe:
  initialDelaySeconds: 10
```

## Proposal

Set `failureThreshold: 10` on all controller and proxy pods, use Kubernetes defaults for all other settings. This should decrease time to readiness, while still permitting 100 seconds for pods to come up.

```yaml
livenessProbe:
  initialDelaySeconds: 0
  periodSeconds: 10
  timeoutSeconds: 1
  successThreshold: 1
  failureThreshold: 10
readinessProbe:
  initialDelaySeconds: 0
  periodSeconds: 10
  timeoutSeconds: 1
  successThreshold: 1
  failureThreshold: 10
```

siggy:
High `readinessProbe.initialDelaySeconds` values delayed the controller's
readiness by up to 30s, preventing cli commands from succeeding shortly after
control plane deployment.

Decrease `readinessProbe.initialDelaySeconds` in the proxy, Prometheus, and
Grafana to the default 0s. Also change `linkerd check` controller pod ordering
to: controller, prometheus, web, grafana.

Detailed probe changes:
- proxy
  - decrease `readinessProbe.initialDelaySeconds` from 10s to 0s
- prometheus
  - decrease `readinessProbe.initialDelaySeconds` from 30s to 0s
  - decrease `readinessProbe.timeoutSeconds` from 30s to 1s
  - decrease `livenessProbe.timeoutSeconds` from 30s to 1s
- grafana
  - decrease `readinessProbe.initialDelaySeconds` from 30s to 0s
  - decrease `readinessProbe.timeoutSeconds` from 30s to 1s
  - decrease `readinessProbe.failureThreshold` from 10 to 3

Fixes #1804


siggy:
Relates to #1121 and #1168.

cli/install/template.go
```diff
@@ -401,14 +401,11 @@ spec:
           httpGet:
             path: /-/ready
             port: 9090
-          initialDelaySeconds: 30
-          timeoutSeconds: 30
```
klingerf: Given the [readiness probe defaults](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes), won't this only give the prometheus pod 20 seconds to start? Is that enough time in clusters that are resource-constrained? How would you feel about setting `failureThreshold: 7` on this probe, to give prometheus a full minute to start?

cli/install/template.go
```diff
@@ -572,10 +569,6 @@ spec:
           httpGet:
             path: /api/health
             port: 3000
-          initialDelaySeconds: 30
-          timeoutSeconds: 30
-          failureThreshold: 10
-          periodSeconds: 10
```
klingerf: Same question here as above -- won't this only give the grafana pod 20 seconds to start? Maybe we can set `failureThreshold: 7` on this probe as well?

klingerf: Unrelated to your change, but I realized the the grafana pod's liveliness probe does not include an initial delay, whereas the prometheus pod's liveliness probe includes `initialDelaySeconds: 30`. Would you mind updating the grafana pod's liveliness probe to match? I actually wonder if the fact that we haven't previously had an initial delay for the grafana pod's liveliness probe could be the cause of #1747.

cli/install/template.go
siggy: I believe pods start as unready and then become ready as soon as they pass their readiness probe. Since the default `successThreshold` is 1, starting the readiness probe immediately will not delay a pod's readiness at startup time, nor will early readiness failures put the pod at risk of restart.

cli/install/template.go
siggy: Good idea around setting `initialDelaySeconds: 30` for Grafana liveness, PR updated.

We should keep this technique in mind if we encounter liveness failures/restarts at startup time, for control or data plane. There's not much harm in increasing `livenessProbe.initialDelaySeconds`.

PR #1902, opened 2018-11-30
tap: Include route labels in tap events

olix0r:
https://github.com/linkerd/linkerd2-proxy-api/pull/17 adds route labels to the proxy's Tap API.

https://github.com/linkerd/linkerd2-proxy/pull/147 adds support to the proxy to emit these labels.

This change alters the controller's Tap service to include route labels when translating tap events, modifies the public API to include route metadata in responses, and modifies the `tap` CLI command to include `rt_` labels in tap output (when `-o wide` is used).


Gopkg.toml
```diff
@@ -9,7 +9,7 @@ required = [
 
 [[constraint]]
   name = "github.com/linkerd/linkerd2-proxy-api"
-  version = "v0.1.3"
+  revision = "de20a798ba70fcc27f91054088786b68fe0d9dcc" # FIXME
```
olix0r: We should not merge this until the linkerd2-proxy-api repo has a 0.1.4 release with this change.

PR #1907, opened 2018-12-03
Display the correct tap/top command in the Tap/Top pages

rmars:
Previously, we were passing in "tap" as the command name for both the tap and top forms, resulting in the equivalent CLI command always being `linkerd tap` regardless of whether you were in the Tap or Top view.

Fix this to correctly pass in tap or top depending on the page.


rmars:
Ah good catch! Fixed!

PR #1918, opened 2018-12-04
Add a tabbed view to the resource detail page for Top and Routes

### LINKED ISSUE #1896: Add routes to the resource detail page

It is possible to get real per-route metrics now, instead of the "real time" solution we have with tap. The resource detail page (deployments for example) should be updated.

Add a route table to the resource detail page above the current `top` table. For the empty state, link to documentation on how to add a service profile to your service.

1. Replace the stream with profile data. I'm concerned this will end up hiding most service's data. It would end up hiding any client data that doesn't have a profile (and any server data that doesn't have a profile). Without any data on how many folks would add profiles to their own (and other's) services, this feels like a big step backwards.

1. Show the stream when there isn't a profile, show the profile data when there is one. Feels a little bit better than the previous option. It doesn't solve the client profile issue and will end up changing the look and feel of the table (columns, updates ...).

rmars:
Adds tabs to the resource detail view to show data from Top and Routes.

Once https://github.com/linkerd/linkerd2/pull/1921 is merged I'm going to rebase this branch on master. But I've already switched the querying that the resource detail pages use so it works against that branch. So if you'd like to run this branch, you should merge in the api changes from that branch.

It would be good to have a way to figure out whether there's a service profile available for that deploy. That way we could default to the Top Routes tab if the service has a profile.

Feel free to nitpick the tab names. They're currently called "Sampled Route Data" (Top) and "Live Route Data" (Top Routes).

Fixes #1896 
screen shot 2018-12-03 at 5 22 37 pm
screen shot 2018-12-03 at 5 24 37 pm


web/app/js/components/ResourceDetail.jsx
```diff
@@ -217,69 +219,62 @@ export class ResourceDetailBase extends React.Component {
...
-              startTap={true}
-              updateNeighbors={this.updateNeighborsFromTapData}
-              maxRowsToDisplay={10} />
-          </div>
-        }
+        <TopRoutesTabs
+          query={query}
+          pathPrefix={this.props.pathPrefix}
+          updateNeighbors={this.updateNeighborsFromTapData}
+          disableTop={!this.state.resourceIsMeshed} />
```
siggy: i don't see `disableTop` used?

web/app/js/components/TopRoutesTabs.jsx
```diff
@@ -0,0 +1,131 @@
...
+class TopRoutesTabs extends React.Component {
+  state = {
+    value: 0,
+  };
+
+  handleChange = (_event, value) => {
+    this.setState({ value });
+  };
+
+  handleChangeIndex = index => {
```
siggy: `handleChangeIndex ` unused?

web/app/js/components/ResourceDetail.jsx
rmars: ah good catch, it is supposed to be used by TopRoutesTabs. fixed!

web/app/js/components/TopRoutesTabs.jsx
rmars: oops, fixed!

rmars:
> With books I can see that sometimes it displays a "not found" message, not sure what's going on.

question, if you paste the command that it generates into the CLI, what does it give you?

siggy:
@rmars Sorry, when I said `With books I can see that sometimes it displays a "not found" message, not sure what's going on` ... I meant that the behavior of showing a "not found" message feels correct, but in the emojivoto screenshot preceding it, there's no "not found" message, and I couldn't reconcile the difference in behavior.

wmorgan:
Just a comment on the tab labels... I think it makes more for the top table to be "Live calls", and for the metrics tab to be "Route metrics". I also would love to see a link to the service profile docs in "no service profile" text, e.g. "No route metrics found. [Create a service profile](https://linky.link) for this service with the `linkerd profile` command."

rmars:
@wmorgan that sounds good. also, now that we can download service profiles from the web app, i'm planning to have a little popup that you can configure a service profile from as well. once we have documentation, I can also link to that!

rmars:
Now that routes for any resource type can be queried (#1921), I've updated this branch to query for the actual service being viewed in ResourceDetail. So the top module there should reflect inbound calls to that resource.

I still need to update the `/routes` page to not be service specific, but I'm going to do that in a separate branch.

rmars:
@grampelberg see https://github.com/linkerd/linkerd2/pull/1918#issuecomment-444274619 and https://github.com/linkerd/linkerd2/pull/1918#issue-235650306

rmars:
@grampelberg what kind of subtle colour treatment would you suggest? more like, greyed out and unnoticeable, or red like error message?

yeah, I can move the CLI queries to the bottom of the table instead

wmorgan:
This is my favorite branch

grampelberg:
@rmars you beat me to the comment, you're on top of it.

Let's leave the `[UNKNOWN]` line the way it is for now, I'll get a design opinion and open up an issue.

web/app/js/components/TopRoutesTable.jsx
```diff
@@ -10,6 +10,11 @@ const routesColumns = [
     dataIndex: "route",
     sorter: (a, b) => (a.route).localeCompare(b.route)
   },
+  {
+    title: "Service",
```
klingerf: Let's call this column "Authority", and if it's straightforward, can we also add a tooltip explaining what an authority is?

web/app/js/components/TopRoutesModule.jsx
```diff
@@ -29,26 +28,14 @@ class TopRoutesBase extends React.Component {
...
-
   render() {
     const {data} = this.props;
     let metrics = processTopRoutesResults(_.get(data, '[0].routes.rows', []));
 
     return (
       <React.Fragment>
         {this.banner()}
-        {_.isEmpty(metrics) ? this.configureProfileMsg() : null}
+        {_.isEmpty(metrics) ? <ConfigureProfilesMsg /> : null}
```
klingerf: Can we update this module to display a spinner if `props.loading` is true, instead of displaying the `ConfigureProfilesMsg`? Or if it's difficult to add a spinner, maybe we can just suppress the `ConfigureProfilesMsg` component while `props.loading` is true.

rmars:
CLI actually less correct :P there's a bug there that keys all the rows by UNKNOWN. what we're actually seeing here is route rows from different authorities

web/app/js/components/BaseTable.jsx
```diff
@@ -91,6 +94,8 @@ class BaseTable extends React.Component {
         </TableCell>
       );
     }
+
+    return _.isNil(col.tooltip) ? tableCell : <Tooltip key={col.key || col.dataIndex} placement="top" title="FOOBAR">{tableCell}</Tooltip>;
```
klingerf: Should this be `title="{col.tooltip}"`?

web/app/js/components/BaseTable.jsx
```diff
@@ -77,12 +79,13 @@ class BaseTable extends React.Component {
             direction={active ? order : col.defaultSortOrder || 'asc'}
             classes={{icon: active ? classes.activeSortIcon : classes.inactiveSortIcon}}
             onClick={this.createSortHandler(col)}>
-            {col.title}
+            {_.isEmpty(col.tooltip) ? col.title : <Tooltip title={col.tooltip} placement="top"><React.Fragment>{col.title}</React.Fragment></Tooltip>}
+            {/* <Tooltip title={col.tooltip} placement="top"><React.Fragment>{col.title}</React.Fragment></Tooltip> */}
```
klingerf: Remove this comment?

web/app/js/components/BaseTable.jsx
```diff
@@ -6,6 +6,7 @@ import TableBody from '@material-ui/core/TableBody';
 import TableCell from '@material-ui/core/TableCell';
 import TableHead from '@material-ui/core/TableHead';
 import TableRow from '@material-ui/core/TableRow';
+import Tooltip from '@material-ui/core/Tooltip';
 import TableSortLabel from '@material-ui/core/TableSortLabel';
```
klingerf: Lines 9 and 10 need to be swapped in order to sort these imports alphabetically and make eslint happy.

PR #1925, opened 2018-12-05
proxy-api: Expose a flag to control auto-h2-upgrade

olix0r:
When debugging issues, it's helpful to disable HTTP/2 upgrading to
simplify diagnostics.

This change adds an `enable-h2-ugprade` flag to _proxy-api_. When this
flag is set to false, the proxy-api will not suggest that meshed
endpoints are upgraded to use HTTP/2.

As a follow-up, a flag should be added to `install` to control how the
proxy-api is initialized.

### DIFF
--- a/controller/cmd/proxy-api/main.go
+++ b/controller/cmd/proxy-api/main.go
@@ -18,6 +18,7 @@ func main() {
 	metricsAddr := flag.String("metrics-addr", ":9996", "address to serve scrapable metrics on")
 	kubeConfigPath := flag.String("kubeconfig", "", "path to kube config")
 	k8sDNSZone := flag.String("kubernetes-dns-zone", "", "The DNS suffix for the local Kubernetes zone.")
+	enableH2Upgrade := flag.Bool("enable-h2-upgrade", true, "Enable transparently upgraded HTTP2 connections among pods in the service mesh")
 	enableTLS := flag.Bool("enable-tls", false, "Enable TLS connections among pods in the service mesh")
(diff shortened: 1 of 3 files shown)


olix0r:
This need not make it in before this week's Edge release...

controller/cmd/proxy-api/main.go
```diff
@@ -18,6 +18,7 @@ func main() {
 	metricsAddr := flag.String("metrics-addr", ":9996", "address to serve scrapable metrics on")
 	kubeConfigPath := flag.String("kubeconfig", "", "path to kube config")
 	k8sDNSZone := flag.String("kubernetes-dns-zone", "", "The DNS suffix for the local Kubernetes zone.")
+	enableH2Upgrade := flag.Bool("enable-h2-upgrade", true, "Enable transparently upgraded HTTP2 connections among pods in the service mesh")
```
klingerf: Related to @siggy's comment on #1925 -- my preference is for boolean flags to default to false, for no other reason than the fact that if you wanted to actually disable protocol upgrade using this flag, you'd have to specify it as `-enable-h2-upgrade=false`. If you inverted it, then you could disable protocol upgrade with `-disable-h2-upgrade` (no argument). And that would be more consistent with the flag that you're adding in #1926.

olix0r: My intent here was that the service exposes several enable- flags, and it would be inconsistent to expose otherwise. What if the default value changes? Changing the flags is an "API breakage" at that point.

PR #1926, opened 2018-12-05
install: Add a -disable-h2-upgrade flag

olix0r:
The proxy-api service _always_ suggests that two meshed pods communicate
via HTTP/2 (i.e. via transparent protocol upgrading, if necessary).
This can complicate debugging and diagnostics at times, so it's
important that we have a way to deploy linkerd without this auto-upgrade
behavior.

This change adds a `-disable-h2-upgrade` flag to the `linkerd install`
command that disables transparent upgrading for the whole cluster.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -57,6 +57,7 @@ type installConfig struct {
 	ProxyBindTimeout                 string
 	SingleNamespace                  bool
 	EnableHA                         bool
+	EnableH2Upgrade                  bool
 }
 
 type installOptions struct {
@@ -65,6 +66,7 @@ type installOptions struct {
 	proxyAutoInject    bool
 	singleNamespace    bool
 	highAvailability   bool
+	disableH2Upgrade   bool
 	*proxyConfigOptions
 }
 
@@ -81,6 +83,7 @@ func newInstallOptions() *installOptions {
 		proxyAutoInject:    false,
 		singleNamespace:    false,
 		highAvailability:   false,
+		disableH2Upgrade:   false,
 		proxyConfigOptions: newProxyConfigOptions(),
 	}
 }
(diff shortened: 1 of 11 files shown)


olix0r:
Depends on #1925. Not urgent.

cli/cmd/install.go
```diff
@@ -108,6 +111,7 @@ func newCmdInstall() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&options.proxyAutoInject, "proxy-auto-inject", options.proxyAutoInject, "Experimental: Enable proxy sidecar auto-injection webhook (default false)")
 	cmd.PersistentFlags().BoolVar(&options.singleNamespace, "single-namespace", options.singleNamespace, "Experimental: Configure the control plane to only operate in the installed namespace (default false)")
 	cmd.PersistentFlags().BoolVar(&options.highAvailability, "ha", options.highAvailability, "Experimental: Enable HA deployment config for the control plane")
+	cmd.PersistentFlags().BoolVar(&options.disableH2Upgrade, "disable-h2-upgrade", options.disableH2Upgrade, "Prevents the controller from instructing proxies to perform transparent HTTP/2 ugprading")
```
siggy: with have `--disable-h2-upgrade` on `linkerd install` and `--enable-h2-upgrade` on `proxy-api`. can we make these the same? prefer `--disable-h2-upgrade` as it implies `true` as the default.

also, `ugprading`

olix0r: My thought process was the following:

proxy-api has several `enable-*` flags, so adding a disable- flag there feels inconsistent.

However, when a user is interacting with `install`, it feels awkward to expose a flag `-enable-h2-upgrade`, since this is the default behavior.  I could be convinced that the `install` flag should be enable-* instead of disable-* if others find that to feel like the right UX.

cc @grampelberg

klingerf: I definitely agree that the public-facing install flag should be `--disable-h2-upgrade`, as you currently have it implemented in this branch. I'm much less concerned about what we call the proxy-api flag, since most users won't interact with that directly anyway.

PR #1921, opened 2018-12-04
Rework routes command to accept any resource

adleong:
We rework the routes command so that it can accept any Kubernetes resource, making it act much more similarly to the stat command.

When no additional flags are given, the routes command displays all incoming routes for the given resource.  Depending on the resource, this may includes routes from multiple different services:

```
$ bin/linkerd routes ns/default
ROUTE                                                      SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TLS
DELETE /authors/{id}.json   authors.default.svc.cluster.local:7001   100.00%   0.3rps         100ms         290ms         298ms    0%
DELETE /books/{id}.json       books.default.svc.cluster.local:7002   100.00%   0.6rps          17ms          92ms          98ms    0%
GET /                        webapp.default.svc.cluster.local:7000   100.00%   0.3rps         250ms         295ms         299ms    0%
GET /authors.json           authors.default.svc.cluster.local:7001   100.00%   0.3rps          10ms          29ms          30ms    0%
GET /authors/{id}            webapp.default.svc.cluster.local:7000   100.00%   0.3rps          75ms         185ms         197ms    0%
GET /authors/{id}.json      authors.default.svc.cluster.local:7001   100.00%   0.8rps           9ms         255ms         291ms    0%
GET /books.json               books.default.svc.cluster.local:7002   100.00%   1.0rps          35ms         180ms         196ms    0%
GET /books/{id}              webapp.default.svc.cluster.local:7000   100.00%   0.6rps         100ms         850ms         970ms    0%
GET /books/{id}.json          books.default.svc.cluster.local:7002   100.00%   1.3rps          12ms          36ms          39ms    0%
HEAD /authors/{id}.json     authors.default.svc.cluster.local:7001    50.48%   1.8rps           9ms         120ms         184ms    0%
POST /authors                webapp.default.svc.cluster.local:7000   100.00%   0.3rps         350ms         395ms         399ms    0%
POST /authors.json          authors.default.svc.cluster.local:7001   100.00%   0.3rps         150ms         195ms         199ms    0%
POST /authors/{id}/delete    webapp.default.svc.cluster.local:7000   100.00%   0.3rps         150ms         285ms         297ms    0%
POST /books                  webapp.default.svc.cluster.local:7000    50.00%   1.2rps         112ms         255ms         291ms    0%
POST /books.json              books.default.svc.cluster.local:7002    52.05%   1.2rps          50ms         750ms         950ms    0%
[...]

PR #1928, opened 2018-12-05
Support service profiles for external authorities

adleong:
This is branched off of https://github.com/linkerd/linkerd2/pull/1921

Add support for service profiles created on external (non-service) authorities.  For example, this allows you to create a service profile named `linkerd.io` which will apply to calls made to `linkerd.io`.

This is done by changing the `LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES` to `.` so that the proxy will attempt to lookup a service profile for any authority.  We provide the `--disable-external-profiles` proxy flag to revert this behavior in case it is a problem.

We also refactor the proxy-api implementation of GetProfiles so that it does the profile lookup, regardless of if the authority looks like a Kubernetes service name or not.  To simplify this, support for multiple resolves (which was unused) was removed.


grampelberg:
Best feature ever.

terminus.yaml
```diff
@@ -0,0 +1,81 @@
+## TERMINUS
```
klingerf: Looks like this file was checked in accidentally -- mind removing it?

adleong: WHOOPS, thanks.

Issue #1887, opened 2018-11-28
Allow for operation on clusters that restrict NET_ADMIN


grampelberg:
### What problem are you trying to solve?

Many production clusters are locked down and do not allow arbitrary pods to have `cap_net_admin`. This makes it, unfortunately, impossible to use linkerd in these environments.

### How should the problem be solved?

It would be nice to have an option (addon?) that does the iptables configuration outside of a pod's initContainer. This would allow an operator with elevated privileges to add the solution to the cluster and have arbitrary pods have minimal privileges.

A good solution would be creating a CNI plugin that does the iptables setup. A DaemonSet would be added to the cluster which adds the CNI plugin to the set. Each time a pod is scheduled, the iptables rules would be added as part of network setup. This allows for them to be in place when the pod starts up (there might be some oddities with initContainers?).

Take a look at istio/cni for some good previous work.

### Any alternatives you've considered?

- DaemonSet controller that updates iptables - Obviously the easiest solution, there are real sync problems and issues with pods starting that don't have rules in place (making policy a little difficult).
- HTTP_PROXY options - Would only work for outgoing traffic, which isn't great.

### How would users interact with this feature?

This should be optional and not default. It would be an awesome use of addons:

```bash
linkerd add secure-cluster-helper
```

There are definitely some big open questions about the interactions between this and things like the auto-inject feature.

mmack:
Great to see this picked up! If you need testers let me know...

In my opinion, even if it is not required, skipping root images and avoiding cap_net_admin priviledges is always a good thing and should be the default. Or are there any downsides other than deploying a deamonset for management?

grampelberg:
@mmack there are definitely tradeoffs. After digging into this some more, it looks like going the CNI route is more sustainable than a daemonset controller (for obvious sync reasons). Unfortunately, not every environment lets you add a new CNI plugin.

Ideally, this *would* be the default though, just need to make sure it works in most cases and has clear guard rails. Some `check` integration would help quite a bit.

mmack:
@grampelberg so the plan is to add a second CNI? My primary is calico atm...

grampelberg:
@mmack that looks like the best solution I can find so far. You add a new CNI plugin to the chain that simply sets the iptables rules up (and cleans them up on teardown).

PR #1929, opened 2018-12-05
add securityContext with runAsUser: {{.ControllerUID}} to the various cont…

codeman9:
…ainers in the install template

Signed-off-by: Cody Vandermyn <kodeman@kodemanindustries.com>


codeman9:
It is unclear to me why this test is failing in the single_namespace case. For some reason, `{{.ProxyUID}}` is returning `0` there.

cli/install/template.go
```diff
@@ -415,6 +423,8 @@ spec:
             cpu: 300m
             memory: 300Mi
         {{- end }}
+        securityContext:
+          runAsUser: 12102
```
klingerf: When I try to install the prometheus pod with this user ID set, the pod ends up in CrashLoopBackoff, with the following error:

```
level=error ts=2018-12-06T00:57:56.2975388Z caller=main.go:617 err="opening storage failed: mkdir data/: permission denied"
```

Based on my reading of prometheus/prometheus#3441, it looks like the pod is already configured to run as a non-root user:

```
$ docker run --rm --entrypoint id prom/prometheus:v2.4.0
uid=65534(nobody) gid=65534(nogroup)
```

And setting that user ID here seems to fix the issue:

```
        securityContext:
          runAsUser: 65534
```

cli/install/template.go
```diff
@@ -807,6 +822,8 @@ spec:
             path: /ready
             port: 9995
           failureThreshold: 7
+        securityContext:
+          runAsUser: 12102
```
klingerf: The proxy-injector pod listens on 443, and it fails to bind with this user ID set:

```
time="2018-12-06T01:28:39Z" level=fatal msg="listen tcp :443: bind: permission denied"
```

I'm not exactly sure how to fix that.

codeman9:
@klingerf Would you also suggest adding in the ability and a default for the "web" pods that is different or just keep that the same as the `controllerUID`?

cli/install/template.go
codeman9: So here, I will just hard-code this value in the template.

cli/install/template.go
codeman9: Can we change the port that the service listens to? The container can still be on 443, but the service can be 8443 or something else?

codeman9: I changed the port and it looks like the proxy-injector pod start up now. Also, I verified that auto injection works on my minikube installation.

cli/install/template.go
```diff
@@ -154,6 +154,8 @@ spec:
         {{.CreatedByAnnotation}}: {{.CliVersion}}
     spec:
       serviceAccount: linkerd-controller
+      securityContext:
+        runAsUser: {{.ControllerUID}}
```
klingerf: I don't think we can set `securityContext` at the pod level, and instead we'll need to see it at the container level. Setting it at the pod level applies to all containers in the pod, including init containers, and as far as I can tell the `linkerd-init` init container will exit silently with an error if it's not run as root.

To repro with this branch, I ran:

```
bin/fast-build
bin/linkerd install | kubectl apply -f -
kubectl -n linkerd logs deploy/linkerd-controller linkerd-init
2018/12/07 23:38:38 Tracing this script execution as [1544225918]
2018/12/07 23:38:38 State of iptables rules before run:
2018/12/07 23:38:38 > iptables -t nat -vnL
2018/12/07 23:38:38 < iptables v1.4.21: can't initialize iptables table `nat': Permission denied (you must be root)
Perhaps iptables or your kernel needs to be upgraded.

2018/12/07 23:38:38 Aborting firewall configuration
Error: exit status 3
Usage:
  proxy-init [flags]

Flags:
  -h, --help                                help for proxy-init
      --inbound-ports-to-ignore intSlice    Inbound ports to ignore and not redirect to proxy. This has higher precedence than any other parameters.
  -p, --incoming-proxy-port int             Port to redirect incoming traffic (default -1)
      --outbound-ports-to-ignore intSlice   Outbound ports to ignore and not redirect to proxy. This has higher precedence than any other parameters.
  -o, --outgoing-proxy-port int             Port to redirect outgoing traffic (default -1)
  -r, --ports-to-redirect intSlice          Port to redirect to proxy, if no port is specified then ALL ports are redirected
  -u, --proxy-uid int                       User ID that the proxy is running under. Any traffic coming from this user will be ignored to avoid infinite redirection loops. (default -1)
      --simulate                            Don't execute any command, just print what would be executed
```

Moving this `securityContext` attribute to each of the containers fixes the issue, but that also means that the `linkerd-init` container will continue as run as root, in which case this branch won't be a complete fix for #1930. But I'm not sure it's going to be possible to run the init container under a different UID, since it mucks with iptables.

cli/install/template.go
```diff
@@ -788,7 +803,7 @@ spec:
         - "-log-level={{.ControllerLogLevel}}"
         ports:
         - name: proxy-injector
-          containerPort: 443
+          containerPort: 8443
```
klingerf: 👍 Great, looks like this fixes the issue that we were having with this container.

cli/install/template.go
klingerf: Note that the `linkerd-init` container already sets its security context as follows:

```
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
          privileged: false
```

codeman9: Good catch! I have updated the initContainer to `runAsUser: 0`, `runAsGroup: 0` and `runAsNonRoot: false`. This looks like it fixes the startup problem you saw. I will amend the issue: #1930 to recognize that the `initContainer`s will need root no matter what, until #1887 is resolved.

cli/cmd/testdata/install_single_namespace_output.golden
```diff
@@ -804,7 +820,13 @@ spec:
             add:
             - NET_ADMIN
           privileged: false
+          runAsGroup: 0
+          runAsNonRoot: false
+          runAsUser: 0
         terminationMessagePolicy: FallbackToLogsOnError
+      securityContext:
+        runAsGroup: 472
+        runAsUser: 472
```
klingerf: Since these IDs are container-specific, my preference is to set them on the container itself, rather than on the pod. Can I trouble you to update this PR so that all of the new security contexts are set per-container instead of per-pod?

PR #1933, opened 2018-12-05
Add a form to create and download service profiles in the web UI

rmars:
This form will be displayed as the call to action if no route metrics are found.

It would be nice to get dropdowns for namespace / service like in #1937 but I don't know if we want to wait for all that to get in before release.

screen shot 2018-12-05 at 11 19 57 am
screen shot 2018-12-05 at 11 04 28 am


rmars:
Agreeeed!

PR #1934, opened 2018-12-05
Add input validation for profile command

### LINKED ISSUE #1878: `profile` has no input validation
### What is the issue?

```bash
$  linkerd -n emojivoto profile --template svc/web-svc
### ServiceProfile for svc/web-svc.emojivoto ###
kind: ServiceProfile
metadata:
  name: svc/web-svc.emojivoto.svc.cluster.local
  namespace: linkerd
```

I would expect this to error out, as `/` is not a valid character.

siggy:
Fixes #1878


cli/cmd/profile.go
```diff
@@ -34,15 +35,38 @@ type profileOptions struct {
...
+		outputs++
+	}
+	if options.openAPI != "" {
+		outputs++
+	}
+	if outputs != 1 {
+		return errors.New("You must specify exactly one of --template or --open-api")
+	}
+
+	if msgs := path.IsValidPathSegmentName(options.name); len(msgs) != 0 {
```
adleong: TIOLI: I was a bit confused by this until I looked up the contract of IsValidPathSegmentName.  Renaming `msgs` to `errs` might make this more clear.

cli/cmd/profile.go
```diff
@@ -34,15 +35,38 @@ type profileOptions struct {
...
+func newProfileOptions() profileOptions {
+	return profileOptions{
 		name:      "",
 		namespace: "default",
 		template:  false,
 		openAPI:   "",
 	}
 }
 
+func validateOptions(options profileOptions) error {
```
klingerf: TIOLI -- for some of our other commands, the convention has been to add a `validate()` func on the options struct itself, rather than passing it as a param. Something like:

```diff
diff --git a/cli/cmd/profile.go b/cli/cmd/profile.go
index c68bf5e9..4bfcfa7e 100644
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -44,7 +44,7 @@ func newProfileOptions() profileOptions {
 	}
 }
 
-func validateOptions(options profileOptions) error {
+func (options *profileOptions) validate() error {
 	outputs := 0
 	if options.template {
 		outputs++
@@ -96,7 +96,7 @@ Example:
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.name = args[0]
 
-			err := validateOptions(options)
+			err := options.validate()
 			if err != nil {
 				return err
 			}
```

cli/cmd/profile_test.go
```diff
@@ -30,3 +32,61 @@ func TestParseProfile(t *testing.T) {
...
+	options = newProfileOptions()
+	options.openAPI = "openAPI"
+	err = validateOptions(options)
+	if err != nil {
+		t.Fatalf("validateOptions returned unexpected error: %s (expected: %s) for options: %+v", err, exp, options)
+	}
+
+	options = newProfileOptions()
+	options.template = true
+	options.name = "service.name"
```
klingerf: This is an invalid service name, unfortunately:

```
$ kubectl create service clusterip service.name --clusterip="None"
The Service "service.name" is invalid: metadata.name: Invalid value: "service.name": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

Maybe we can find that regex in the kubernetes codebase and use that?

klingerf:
Ah, and I think you need to run `bin/update-go-deps-shas` on this branch, given the change to `Gopkg.lock`.

cli/cmd/profile_test.go
siggy: Good catch, I'll use something around `IsDNS1035Label` in the k8s codebase.

cli/cmd/profile.go
```diff
@@ -34,15 +35,42 @@ type profileOptions struct {
 	openAPI   string
 }
 
-func newProfileOptions() *profileOptions {
-	return &profileOptions{
+func newProfileOptions() profileOptions {
```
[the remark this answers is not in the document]
siggy: With Go, I typically use structs unless there is an explicit reason to use a pointer, such as modification by a callee, or large memory footprint. My understanding is that pointers in Go tend to put more pressure on GC, specifically escape analysis, and that outweighs memory allocation caused by copying small structs.

That said, I can change this back to be consistent with the other cli commands, though if we just want consistency I'd be inclined to move all these `new*Options` off of pointers.

A few references:
https://github.com/golang/go/wiki/CodeReviewComments#pass-values
https://segment.com/blog/allocation-efficiency-in-high-performance-go-services/
https://medium.com/@vCabbage/go-are-pointers-a-performance-optimization-a95840d3ef85
https://flaviocopes.com/golang-methods-receivers/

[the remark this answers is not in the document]
klingerf: Right, my strong preference is to keep this consistent across all of our commands. Plus, I find pointers easier to reason about, and those performance concerns don't apply in the same way to the CLI. Until the CLI is suffering from performance issues, I would really like to keep things the way they are.

PR #1937, opened 2018-12-05
Show Top Routes in sidebar, change form to query any resource

rmars:
Now that #1921 has merged, we can query for top routes for any resource, not just services.

This PR adds a dropdown for all resources to the Top Routes query form.
It also adds a link to the Top Routes page in the sidebar.

For the sake of expediency, I've taken out all other fields. We'll need to add the fields to query the `--to` fields later.

screen shot 2018-12-05 at 2 00 21 pm


web/app/js/components/Navigation.jsx
```diff
@@ -234,6 +234,7 @@ class NavigationBase extends React.Component {
             { this.menuItem("/overview", "Overview", <HomeIcon />) }
             { this.menuItem("/tap", "Tap", <Icon className={classNames("fas fa-microscope", classes.shrinkIcon)} />) }
             { this.menuItem("/top", "Top", <Icon className={classNames("fas fa-stream", classes.shrinkIcon)} />) }
+            { this.menuItem("/routes", "Top Routes", <Icon className={classNames("fas fa-random", classes.shrinkIcon)} />) }
```
klingerf: 😻 I like your choice of icon!

PR #1941, opened 2018-12-06
Add the create new service profile button by default at /routes

### LINKED ISSUE #1940: Display Create New Profile button in more places in the web UI
Currently the call to action is only displayed when there are no metrics for a route under a resource.

We should display this button on the `/routes` page, and we should also display it if a resource has no configured routes (i.e. all routes are UNKNOWN)

rmars:
Also display the "Create service profile" button when all route metrics are `UNKNOWN`

Also fixes a console error I noticed from props validation.

Fixes #1940

screen shot 2018-12-05 at 4 35 27 pm
screen shot 2018-12-05 at 4 35 01 pm


web/app/js/components/TopRoutesModule.jsx
```diff
@@ -41,12 +41,14 @@ class TopRoutesBase extends React.Component {
...
+    let noServiceProfile = _.isEmpty(metrics) || _.every(metrics, m => m.route === "UNKNOWN");
+    let showCallToAction = !loading && noServiceProfile;
 
     return (
       <React.Fragment>
         {this.loading()}
         {this.banner()}
-        { !loading && _.isEmpty(metrics) ? <ConfigureProfilesMsg /> : null}
+        { showCallToAction ? <ConfigureProfilesMsg /> : null}
         <TopRoutesTable rows={metrics} />
```
klingerf: I found it a bit surprising that on the top routes page, for a service with traffic but no service profile, we display both the configure profile message and the routes table with one row of metrics in it:

image

I'd expect for us to either display the table or the call to action, but not both. And now that I see this in action, I think I prefer to see the routes table with an unknown route, since there's some value in the metrics data, even for unknown routes.

Am just thinking out loud here, but maybe, instead of displaying the call to action, we include some sort of popover on the UNKNOWN text that explains that the route is unknown because the service does not include a service profile, and we prompt the user to create the profile from the popover. Or... something like that. Am not sure the best way to go about this, unfortunately.

klingerf: A similar thing is happening on the resource detail page for resources that have traffic but no associated service profile:

image

rmars: yeah, I was thinking about changing the text of the message to read "No named routes found" which would explain why the user is seeing all UNKNOWNS.

other idea: display that message at the *bottom* of the table with a message like:
`All routes are UNKNOWN. Have you created a Service Profile? <button>Create one now</button>`

rmars: screen shot 2018-12-06 at 11 19 03 am

klingerf: @rmars Yeah! That looks like a good improvement. I think there's another state for the table, however, which is when there is actually no traffic to the service at all. In that case, we can't infer whether or not the service has a service profile or not, so I think we should just display some sort of "No data" message, without prompting the user to create a service profile (this is sort of the opposite of today's behavior). Thoughts?

rmars: screen shot 2018-12-06 at 11 44 36 am

PR #1952, opened 2018-12-06
Add --routes flag to linkerd top

adleong:
Add a `--routes` flag to the `top` command which show requests by route instead of by path.  

This also includes an overhaul of the live table rendering code to be more flexible and better accommodate the various combinations of columns that may be visible or hidden.

### DIFF
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -30,6 +30,7 @@ type topOptions struct {
 	authority   string
 	path        string
 	hideSources bool
+	routes      bool
 }
 
 type topRequest struct {
@@ -49,9 +50,25 @@ func (id topRequestID) String() string {
 	return fmt.Sprintf("%s->%s(%d)", id.src, id.dst, id.stream)
 }
 
+type tableColumn struct {
+	header string
+	width  int
+	// Columns with key=true will be treated as the primary key for the table.
+	// In other words, if two rows have equal values for the key=true columns
+	// then those rows will be merged.
+	key bool
+	// If true, render this column.
+	display bool
+	// If true, set the width to the widest value in this column.
+	flexible   bool
+	rightAlign bool
+	value      func(tableRow) string
+}
+
 type tableRow struct {
-	by          string
+	path        string
 	method      string
+	route       string
 	source      string
 	destination string
 	count       int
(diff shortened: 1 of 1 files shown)


cli/cmd/top.go
```diff
@@ -175,7 +342,7 @@ func getTrafficByResourceFromAPI(w io.Writer, client pb.ApiClient, req *pb.TapBy
 	go recvEvents(rsp, requestCh, done)
 	go pollInput(done)
 
-	renderTable(requestCh, done, !options.hideSources)
+	renderTable(requestCh, done)
```
siggy: unrelated to this PR, but just realized `options` is no longer needed as a param to this function.

cli/cmd/top.go
```diff
@@ -62,11 +79,145 @@ type tableRow struct {
...
+		key:        false,
+		display:    true,
+		flexible:   false,
+		rightAlign: true,
+		value: func(r tableRow) string {
+			return fmt.Sprintf("%.2f%%", 100.0*float32(r.successes)/float32(r.successes+r.failures))
+		},
+	}
+
+	columns = []*tableColumn{
```
siggy: i realize we already had this pattern, but this introduces a lot of global vars into the `cmd` package. consider hiding it in a function, something like...

```go
func mkColumns(options topOptions) []tableColumn {
	sourceColumn := tableColumn{...}
	...
	return []tableColumn{sourceColumn, ...}
}

...

return getTrafficByResourceFromAPI(os.Stdout, validatedPublicAPIClient(time.Time{}), req, mkColumns(options))
```

... you could even go a step further and create a new `topTable` type, something like...

```go
type topTable struct{
	columns []tableColumn
	rows []tableRow
}

func newTopTable(options topOptions) (*topTable) {}
func (t *topTable) insert(row tableRow) {}
func (t *topTable) render() {}
```

cli/cmd/top.go
```diff
@@ -286,65 +459,77 @@ func tableInsert(table *[]tableRow, req topRequest, withSource bool) {
 		}
 	}
 
+	successes := 0
+	failures := 0
+	if success {
+		successes++
+	} else {
+		failures++
+	}
```
siggy: maybe this is slightly more explicit ?

```go
successes := 0
failures := 0
if success {
	successes = 1
} else {
	failures = 1
}
```

adleong:
@siggy I've created a topTable struct to move the column definitions out of the global namespace.  I kept the columns as named fields in the struct because I want to be able to refer to them by name.  The result is a bit more verbose, but it doesn't clutter up the global namespace as much.  Let me know what you think.

adleong:
@siggy fantastic!

PR #1966, opened 2018-12-08
Add rules_file loading into prometheus

grampelberg:


cli/install/template.go
```diff
@@ -434,6 +434,9 @@ data:
       scrape_timeout: 10s
       evaluation_interval: 10s
 
+    rule_files:
+      - /etc/prometheus/*_rules.yml
```
siggy: nit: mind removing this indent to match the rest of the file?

PR #1974, opened 2018-12-12
Remove namespace definition from --single-namespace installs

### LINKED ISSUE #1889: Remove namespace definition from `--single-namespace`
### What is the issue?

When using `--single-namespace`, the namespace definition shows up:

```yaml
kind: Namespace
apiVersion: v1
metadata:
  name: linkerd
```

While this makes sense for global installs, in the `--single-namespace` world, it requires extra permissions (assuming it was created, patch). That's not great! It should just be removed from the install output.

klingerf:
Users installing linkerd in single-namespace mode likely don't have cluster-wide permissions to create namespaces, so in this branch I'm removing the namespace definition from the `linkerd install --single-namespace` output, and I'm updating `linkerd check --single-namespace` to require that the namespace already exist, and not require that the user has permissions to create namespaces.

Fixes #1889.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -200,19 +200,18 @@ func (hc *HealthChecker) addKubernetesAPIChecks() {
 }
 
 func (hc *HealthChecker) addLinkerdPreInstallChecks() {
+	roleType := "ClusterRole"
+	roleBindingType := "ClusterRoleBinding"
+
 	if hc.SingleNamespace {
+		roleType = "Role"
+		roleBindingType = "RoleBinding"
+
 		hc.checkers = append(hc.checkers, &checker{
 			category:    LinkerdPreInstallCategory,
 			description: "control plane namespace exists",
 			check: func() error {
-				exists, err := hc.kubeAPI.NamespaceExists(hc.httpClient, hc.ControlPlaneNamespace)
-				if err != nil {
-					return err
-				}
-				if !exists {
-					return fmt.Errorf("The \"%s\" namespace must exist before installing with --single-namespace", hc.ControlPlaneNamespace)
-				}
-				return nil
+				return hc.checkNamespace(hc.ControlPlaneNamespace, true)
 			},
 		})
 	} else {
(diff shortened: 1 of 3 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -200,28 +200,47 @@ func (hc *HealthChecker) addKubernetesAPIChecks() {
...
 
-	hc.checkers = append(hc.checkers, &checker{
-		category:    LinkerdPreInstallCategory,
-		description: "can create Namespaces",
-		check: func() error {
-			return hc.checkCanCreate("", "", "v1", "Namespace")
-		},
-	})
+	if !hc.SingleNamespace {
+		hc.checkers = append(hc.checkers, &checker{
```
rmars: tioli you could put this in the if/else right above, to avoid splitting out the logic for "if single namespace" vs "if not single namespace"?

dadjeibaah: In addition to @rmars's comment, do you think it is possible to DRY this check up a little bit? Something like this?
```diff
diff --git a/pkg/healthcheck/healthcheck.go b/pkg/healthcheck/healthcheck.go
index 49a479c9..082f5f8f 100644
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -200,35 +200,30 @@ func (hc *HealthChecker) addKubernetesAPIChecks() {
 }
 
 func (hc *HealthChecker) addLinkerdPreInstallChecks() {
+	namespaceCheck := func(msg string) func() error {
+		return func() error {
+		exists, err := hc.kubeAPI.NamespaceExists(hc.httpClient, hc.ControlPlaneNamespace)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			return fmt.Errorf(msg, hc.ControlPlaneNamespace)
+		}
+		return nil
+		}
+	}
+
 	if hc.SingleNamespace {
 		hc.checkers = append(hc.checkers, &checker{
 			category:    LinkerdPreInstallCategory,
 			description: "control plane namespace exists",
-			check: func() error {
-				exists, err := hc.kubeAPI.NamespaceExists(hc.httpClient, hc.ControlPlaneNamespace)
-				if err != nil {
-					return err
-				}
-				if !exists {
-					return fmt.Errorf("The \"%s\" namespace must exist before installing with --single-namespace", hc.ControlPlaneNamespace)
-				}
-				return nil
-			},
+			check:       namespaceCheck("The \"%s\" namespace must exist before installing with --single-namespace"),
 		})
 	} else {
 		hc.checkers = append(hc.checkers, &checker{
 			category:    LinkerdPreInstallCategory,
 			description: "control plane namespace does not already exist",
-			check: func() error {
-				exists, err := hc.kubeAPI.NamespaceExists(hc.httpClient, hc.ControlPlaneNamespace)
-				if err != nil {
-					return err
-				}
-				if exists {
-					return fmt.Errorf("The \"%s\" namespace already exists", hc.ControlPlaneNamespace)
-				}
-				return nil
-			},
+			check:       namespaceCheck("The \"%s\" namespace already exists"),
 		})
 	}
```

klingerf: All good suggestions -- have addressed them in the latest commit. Turns out we already had a `checkNamespace` func in this file that I could repurpose for these checks.

Issue #1877, opened 2018-11-27
Linkerd2 install in OpenShift


kinash000:
I have error after install linkerd2 in minishift:
OpenShift v3.9
Kubernetes v1.9.1
Linkerd stable-2.0.0

After installation, the collection of metrics does not work:
`linkerd stat pods --all-namespaces`
```
NAMESPACE NAME MESHED SUCCESS RPS LATENCY_P50 LATENCY_P95 LATENCY_P99 TLS
default docker-registry-1-xk7mz 0/1 - - - - - -
default router-1-d9f9k 0/1 - - - - - -
emojivoto emoji-8466d478db-8gsnf 1/1 - - - - - -
emojivoto vote-bot-77f8478f6b-g6f5j 1/1 - - - - - -
emojivoto voting-858656b5c8-gzv6l 1/1 - - - - - -
emojivoto web-68656ddd64-b7g2j 1/1 - - - - - -
kube-system kubernetes-dashboard-6c664cf6c5-knnlh 0/1 - - - - - -
linkerd controller-d6f584b7d-gntjg 1/1 - - - - - -
linkerd grafana-d57f845dd-5k8l4 1/1 - - - - - -
linkerd prometheus-6b979dc67d-mqrss 1/1 - - - - - -
linkerd web-5bc657776c-bpkwc 1/1 - - - - - -
openshift-web-console webconsole-5f567f6dc5-vjqbg 0/1 - - - - - -
```

But i try installed Linkerd in Katakoda's OpenShift and all works.

Install commands:
```
curl -sL https://run.linkerd.io/install | sh
export PATH=$PATH:$HOME/.linkerd2/bin
oc adm policy add-scc-to-user privileged -z linkerd-controller -n linkerd
oc adm policy add-scc-to-user privileged -z linkerd-prometheus -n linkerd
oc adm policy add-scc-to-user privileged -z default -n linkerd
linkerd install | oc apply -f - 
```

PR #1978, opened 2018-12-12
Create new service accounts for linkerd-web and linkerd-grafana. Chan…

codeman9:
…ge 'serviceAccount:' to 'serviceAccountName:'

### DIFF
--- a/cli/install/template.go
+++ b/cli/install/template.go
@@ -155,7 +155,7 @@ spec:
       annotations:
         {{.CreatedByAnnotation}}: {{.CliVersion}}
     spec:
-      serviceAccount: linkerd-controller
+      serviceAccountName: linkerd-controller
       containers:
       - name: public-api
         ports:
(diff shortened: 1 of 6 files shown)


cli/install/template.go
```diff
@@ -275,6 +275,14 @@ spec:
     shortNames:
     - sp
 
+### Service Account Web ###
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-web
+  namespace: linkerd
```
klingerf: Whoops, I realize I didn't specify this correctly in my comment on #1877. Sorry about that. We should use the `{{.Namespace}}` template variable here to make sure these are created in the correct namespace.

```
  namespace: {{.Namespace}}
```

cli/install/template.go
```diff
@@ -520,6 +529,14 @@ data:
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
 
+### Service Account Grafana ###
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-grafana
+  namespace: linkerd
```
klingerf: Same comment here as above:

```
  namespace: {{.Namespace}}
```

PR #1980, opened 2018-12-13
Disable serivce profiles in single-namespace mode

### LINKED ISSUE #1847: Issues with `linkerd install --linkerd-namespace [$namespace] -—single-namespace | kubectl apply -f -` when execute with limited role.
### What is the issue?
Because I am operating from an user within a enterprise version of kubernetes clusters, I only have namespace level permission. With the help from @rmars & @wmorgan, I was able to try out the `--single-namespace` feature in edge release. However, there seems to be sign of bugs and I will attempt to document them.  

### How can it be reproduced?
[the namespace I operate within is called `clickstream`]
I ran `linkerd install --linkerd-namespace clickstream —single-namespace | kubectl apply -f -`

### Logs, error output, etc

paste the link here.)
https://gist.github.com/jimcal/937ddfb95e97ce8da4c51e7fef737bda

#### `linkerd check` output

```text
kubernetes-api: is running the minimum Kubernetes API version..............[ok]
kubernetes-setup: control plane namespace does not already exist...........[FAIL] -- The "clickstream" namespace already exists
kubernetes-setup: can create Namespaces....................................[FAIL] -- Missing permissions to create Namespace

Status check results are [FAIL]
```

### Environment

- Kubernetes Version: Client Version: v1.9.6 / Server Version: v1.10.8
- Cluster Environment: Enterprise self administrated Kubernetes platform on AWS 
- Host OS: Container Linux by CoreOS 1855.4.0 (Rhyolite)
- Linkerd version: edge-18.11.1

### Additional context
Related issue https://github.com/linkerd/linkerd2/issues/1846

### LINKED ISSUE #1890: RBAC issues with service profiles

With the introduction of service profiles, the required RBAC for installing linkerd has grown larger. In particular, these two definitions are troublesome:

```yaml
kind: Role
metadata:
  name: linkerd-linkerd-controller
  namespace: linkerd
rules:
- apiGroups: ["linkerd.io"]
  resources: ["serviceprofiles"]

---

metadata:
  name: serviceprofiles.linkerd.io
  namespace: linkerd
  annotations:
    linkerd.io/created-by: linkerd/cli edge-18.11.2
spec:
  group: linkerd.io
  version: v1alpha1
  scope: Namespaced
  names:
    plural: serviceprofiles
    singular: serviceprofile
    kind: ServiceProfile
    shortNames:
    - sp
```

It would be nice to have an `--exclude-service-profile` option for `install`. These should be default, but optionally be disabled.

This feels like moving further down the addon path. Adding more and more options to `install` is starting to feel like it is going to require some in-depth understanding to know which options to use. These should *definitely* be on by default, but maybe a better option would be to allow users to list which addons they'd like to install, perhaps:

```bash
linkerd install --addons=grafana,serviceprofile,inject
```

These could also be added post install via:

```bash
linkerd addon serviceprofile
```

klingerf:
Service profiles are backed by a CRD, which requires cluster-wide permissions to create. Since users running in single-namespace mode don't have cluster-wide permissions, I'm disabling service profiles when linkerd is installed with the `--single-namespace` flag. This branch conditionally removes the CRD from the install YAML, adjusts RBAC permissions accordingly, modifies pre-flight permission checks to stop checking for permission to create CRDs, updates the destination service to not watch sp resources and to close `GetProfile` streams, and updates the web UI to not prompt users to create service profiles.

Fixes #1847.
Fixes #1890.

### DIFF
--- a/web/app/js/components/ConfigureProfilesMsg.jsx
+++ b/web/app/js/components/ConfigureProfilesMsg.jsx
@@ -140,7 +140,9 @@ class ConfigureProfilesMsg extends React.Component {
   render() {
     const { showAsIcon } = this.props;
 
-    if (showAsIcon) {
+    if (this.props.singleNamespace === "true") {
+      return null;
+    } else if (showAsIcon) {
       return this.renderDownloadProfileForm();
     } else {
       return (
@@ -161,7 +163,8 @@ ConfigureProfilesMsg.propTypes = {
     prefixedUrl: PropTypes.func.isRequired,
   }).isRequired,
   classes: PropTypes.shape({}).isRequired,
-  showAsIcon: PropTypes.bool
+  showAsIcon: PropTypes.bool,
+  singleNamespace: PropTypes.string.isRequired,
 };
 
 ConfigureProfilesMsg.defaultProps = {
--- a/cli/cmd/testdata/install_default.golden
+++ b/cli/cmd/testdata/install_default.golden
@@ -23,7 +23,7 @@ rules:
   resources: ["deployments", "replicasets"]
   verbs: ["list", "get", "watch"]
 - apiGroups: [""]
-  resources: ["pods", "endpoints", "services", "namespaces", "replicationcontrollers"]
+  resources: ["pods", "endpoints", "services", "replicationcontrollers", "namespaces"]
   verbs: ["list", "get", "watch"]
 - apiGroups: ["linkerd.io"]
   resources: ["serviceprofiles"]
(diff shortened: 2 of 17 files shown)


web/app/js/components/ConfigureProfilesMsg.jsx
```diff
@@ -140,7 +140,9 @@ class ConfigureProfilesMsg extends React.Component {
   render() {
     const { showAsIcon } = this.props;
 
-    if (showAsIcon) {
+    if (this.props.singleNamespace === "true") {
```
rmars: 😭

klingerf: Heh, yeah, I cried a bit too when I wrote this...

Issue #217, opened 2018-01-26
Improve Go Report Card


siggy:
We are getting an A+ on Go Report Card, but there is still a lot to improve:
~https://goreportcard.com/report/github.com/runconduit/conduit~
https://goreportcard.com/report/github.com/linkerd/linkerd2

We should consider these changes, and also add a badge to `README.md`.

klingerf:
Hmm, looks like almost all of the issues in the repo could be found with the [golint](https://github.com/golang/lint) tool. We should get in the habit of running that.

pcalcado:
Should we create a `precommit.sh` file or something that does a standardized `go test ./... && go vet ./... && golint`? Also could be the script travis runs

klingerf:
Heh, and this one:

> func name will be used as version.VersionFlag by other packages, and that stutters; consider calling this Flag

pcalcado:
I am very easygoing with these things, I can definitely live with the silly ones if it gives us a standardised and well-supported way to check for more interesting things.

rafaeljesus:
IMHO I am not really about skipping these lint errors, it's pretty handy for writing idiomatic golang code ^.^

Also it might be required if you decide to add it at [awesome-go](https://github.com/avelino/awesome-go) list

siggy:
Taking this a bit further, #2284 introduced a [.golangci.yml](https://github.com/linkerd/linkerd2/blob/5b836526b8018ae1315df87005a42a493b31ba55/.golangci.yml) file, which specifies which golangci-lint checkers to run.

To help improve the codebase further, we should work towards reducing the number of checkers in the `disable` list. Essentially repeat this pattern:
1. remove an item from the `disable` list
1. run `bin/lint`, observe errors
1. fix errors until `bin/lint` succeeds

grampelberg:
@siggy this issue doesn't feel particularly actionable at this point (and there's a ton of awesome work that's already been done). WDYT about closing this out and opening up directed issues for what's remaining?

siggy:
@grampelberg I'm fine with closing this one.

Note that we lost our perfect gofmt score recently:
https://goreportcard.com/report/github.com/linkerd/linkerd2#gofmt
This may have happened when we changed our lint config, probably worth a new issue to track this.

grampelberg:
I was wondering about that. Thanks!

PR #1982, opened 2018-12-13
Address linter errors

radu-matei:
First of all, thanks a lot for this awesome project!

This PR begins to address #217, and it addresses all issues that are not comment related, and some obvious comment issues. However, it does _not_ add canonical comments to all exported types - for two reasons - a lot of them are missing, and I don't have the necessary expertise to add any value to the comments.

Thus, this PR addresses all other linter issues, which have been checked with [`golangci-lint`](https://github.com/golangci/golangci-lint), a nifty tool similar to [`gometalinter`](https://github.com/alecthomas/gometalinter), but significantly faster.

There is still one linter error (see below) where the fix really depends on how we want to use the package - so I didn't want to make any assumptions here - as this PR doesn't make any fundamental changes.

```
linkerd2: $ golangci-lint run --disable-all -E golint
pkg/filesonly/filesonly.go:15:29: exported func FileSystem returns unexported type filesonly.fileSystem, which can be annoying to use (golint)
func FileSystem(dir string) fileSystem {
                            ^
```

TODO:

- there was a comment on adding a `precommit.sh` script  that would also be used by Travis - I'd recommend a Makefile - if it's the route we want to take, I'd happy to follow-up with another PR

- decide how to approach the remaining linter error

- decide how to approach the comment related linter errors.


klingerf:
Thanks for submitting this, @radu-matei! Looks like a huge improvement, and we'll review it more thoroughly soon.

wmorgan:
@radu-matei very cool, thanks for kicking this off!

radu-matei:
Hey, thanks for the review!
Will update the branch later today and ping back when it's ready for a re-review.

Thanks!

PR #1988, opened 2018-12-14
Update CHANGES.md for the edge-18.12.3 release

rmars:


CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+  * Fix proxy-api handling of named target ports in service configs
+  * Add parameter to stats API to skip retrieving prometheus stats (thanks,
+    @alpeb!)
+* Web UI
+  * Adjust label for unknown routes in route tables, add tooltip
+  * Update Top Routes page to persist form settings in URL
+  * Add button to create new service profiles on Top Routes page
+  * Fix CLI commands displayed when linkerd is running in non-default namespace
+* Proxy
+  * Disable protocol detection for server-speaks-first protocols on default
```
hawkw: I'd rephrase this as something like "Proxies with TLS enabled now honor ports which skip protocol detection"; this change fixes both the default skipped ports _and_ any user-configured ones.

hawkw: (also it's not actually on master yet, due to issues w/ the proxy CI build)

rmars: we'll wait until this gets in, update the proxy version and then ship the release! (or if this change won't get in before eod Friday, we'll release without it, and remove this bullet point).

hawkw: #1989 updates the pinned proxy version to pick up this change :)

PR #1995, opened 2018-12-19
Start reducing webpack bundle size, fix minor bugs

rmars:
- Adds bundle analyzer so we can see the composition of our bundle.
- Stop importing all the locales of moment.js
- Use named imports for all places we use material-ui components
- Fix a bug where, due to the controller components being relabled `linkerd-`, the service mesh page wasn't showing the correct statuses.
- Fixes some eslint warnings on tests

Next up will be fixing our lodash imports, and moving away from postcss

Running `NODE_ENV=production webpack -p --progress`:
bundle size (before):
index_bundle.js (1.42 MiB)

bundle size (after):
index_bundle.js (1.2 MiB)

### DIFF
--- a/web/app/js/components/ServiceMesh.jsx
+++ b/web/app/js/components/ServiceMesh.jsx
@@ -39,12 +39,12 @@ const getPodClassification = pod => {
 };
 
 const componentsToDeployNames = {
-  "Grafana" : "grafana",
-  "Prometheus": "prometheus",
-  "Proxy API": "controller",
-  "Public API": "controller",
-  "Tap": "controller",
-  "Web UI": "web"
+  "Grafana" : "linkerd-grafana",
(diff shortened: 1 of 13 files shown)


web/app/js/components/ServiceMesh.jsx
```diff
@@ -39,12 +39,12 @@ const getPodClassification = pod => {
...
-  "Grafana" : "grafana",
-  "Prometheus": "prometheus",
-  "Proxy API": "controller",
-  "Public API": "controller",
-  "Tap": "controller",
-  "Web UI": "web"
+  "Grafana" : "linkerd-grafana",
+  "Prometheus": "linkerd-prometheus",
+  "Proxy API": "linkerd-controller",
+  "Public API": "linkerd-controller",
```
siggy: nice catch!

PR #2008, opened 2018-12-19
Fix update neighbors function not being passed in to top module

rmars:
I realized that at some point we stopped showing unmeshed upstreams in the octopus graph.
Turns out we were passing in the wrong prop name for the update function.

Before:
screen shot 2018-12-19 at 3 10 58 pm
screen shot 2018-12-19 at 3 09 06 pm

After:
screen shot 2018-12-19 at 3 07 02 pm
screen shot 2018-12-19 at 3 08 48 pm


grampelberg:
This would have helped with some debugging yesterday, yay!

Issue #2013, opened 2018-12-20
remove proxy bind timeout


olix0r:
The bind timeout currently does not constrain discovery, but is effectively an outbound response timeout.

With https://github.com/linkerd/linkerd2-proxy/pull/165, we have user-configurable response timeouts.

Let's remove the bind timeout entirely and figure out proper service discovery semantics separately.

olix0r:
the CLI will also need to change to remove this option from install/inject

hawkw:
@olix0r as far as removing the CLI options for the bind timeout goes, do you think we should continue to accept those flags and _warn_ the user that they no longer do anything (and remove them from the docs), or completely remove them so that invoking install or inject with `--proxy-bind-timeout` will _fail_? 

I'm wondering if a "soft" removal (the former option) might be a better choice to avoid breaking anyone's existing scripts etc. Also, if we eventually re-introduce a timeout that _does_ constrain discovery rather than outbound responses, we might want to re-use the `--proxy-bind-timeout` flag and/or env var, so temporarily breaking and then re-adding them might not be ideal. What do you think is the proper UX around this?

olix0r:
I think it would be especially confusing if we change the semantics of an existing flag later.

I think I'm mostly in favor of removing the flag now that we'll have timeouts in service profiles; so that, if we do choose to add the flag later, existing scripts won't have the semantics changed out from nder them.

If we feel it's important to deprecate this incrementally, that's an option; though it may not be worth it. cc @grampelberg

grampelberg:
I'm behind a hard failure, we can document why it changed if folks ask. My suspicion is that it won't be an issue for something like this.

hawkw:
Okay, that makes sense to me, thanks for the input.

olix0r:
FWIW, It does look like the bind timeout is at least in-part removed in https://github.com/linkerd/linkerd2-proxy/pull/165 (I hadn't looked at the PR closely before opening this).

hawkw:
@olix0r ah, in that case, I'll just look at that branch and make sure it's completely removed, I suppose.

I've opened #2017 to remove the `--proxy-bind-timeout` flag from `install` and `inject`.

seanmonstar:
The support in the proxy was just removed, though this may wish to track the CLI options also?

grampelberg:
I'm fine addressing it in the changelog.

Issue #1748, opened 2018-10-09
Use same code path for CLI injection and admission controller injection


klingerf:
Once #1714 merges, it will be possible to install linkerd with an admission controller that automatically injects new deployments. The admission controller uses a YAML template from template.go to produce the injected sidecar configuration, whereas the CLI inject subcommand builds its configuration using Kubernetes objects. The net result is the same, but these two implementations could drift over time. We should standardize injection to use a single code path. cc @ihcsim.

alenkacz:
@klingerf which of the two options for we want to lean to? Any preferences?

ihcsim:
I think the proxy and proxy-init container specs generated in [`cli/cmd/inject.go`](https://github.com/linkerd/linkerd2/blob/38dfc5308f01c4af247beaa58fc83db9ebdf10ad/cli/cmd/inject.go#L361-L362) should be used to populate the [config map](https://github.com/linkerd/linkerd2/blob/38dfc5308f01c4af247beaa58fc83db9ebdf10ad/cli/install/template.go#L833-L948) in template.go during `install`, instead of hard-coding it as seen in template.go.

klingerf:
That's right -- when linkerd is installed with `--proxy-auto-inject` enabled, the install YAML includes a `proxy-injector-sidecar-config` ConfigMap. Right now, the contents of that ConfigMap are templated to mirror the output of running `linkerd inject` on a pod spec, but any changes to inject also need to be reflected in the template. It would be better to produce the template using the inject code itself.

Taking this one step further, however, I think an additional improvement could be to stop storing auto-inject YAML in the ConfigMap altogether, and instead store an object containing inject options, which we could then use to produce the YAML at auto-inject time. That would allow us to further standardize with the standard inject code path, by sharing this struct:

https://github.com/linkerd/linkerd2/blob/38dfc5308f01c4af247beaa58fc83db9ebdf10ad/cli/cmd/root.go#L133-L153

This would be a pretty sizable change, however.

olix0r:
What do folks think about putting the entire inject logic into the public-api service? such that the `inject` command and the webhook can make an API call that takes yaml/json and receives yaml/json. This is especially useful if cluster-wide policy is set on the controller that needs to influence how injection is done. Relates to https://github.com/linkerd/linkerd2/issues/1999

klingerf:
[...]

PR #2017, opened 2018-12-20
Remove proxy bind timeout from CLIs

hawkw:
This branch removes the `--proxy-bind-timeout` flag from the 
`linkerd inject` and `linkerd install` CLI commands, and the
`LINKERD2_PROXY_BIND_TIMEOUT` environment variable from their output.
This is in preparation for removing that timeout from the proxy (as
described in #2013). 

I thought it was prudent to remove this from the CLIs before removing it
from the proxy, so we can't create a situation where the CLIs produce
output that results in broken proxy containers.


cli/install/template.go
```diff
@@ -981,8 +981,6 @@ data:
     env:
     - name: LINKERD2_PROXY_LOG
       value: warn,linkerd2_proxy=info
-    - name: LINKERD2_PROXY_BIND_TIMEOUT
-      value: {{.ProxyBindTimeout}}
```
klingerf: This change also needs to be reflected in the test fixtures for the proxy-injector package.

```
./controller/proxy-injector/fake/data/config-proxy.yaml:4:- name: LINKERD2_PROXY_BIND_TIMEOUT
./controller/proxy-injector/fake/data/inject-sidecar-container-spec.yaml:4:- name: LINKERD2_PROXY_BIND_TIMEOUT
```

And updating those files will require re-encoding the patches in these files:

```
controller/proxy-injector/fake/data/inject-no-labels-response.yaml
controller/proxy-injector/fake/data/inject-enabled-response.yaml
```

You can do that by copying the test output and passing it through `base64`. Apologies this isn't more straightforward -- we'll improve upon it as part of #1748.

hawkw:
Now that linkerd/linkerd2-proxy#165 has merged, I think this is unblocked.

cli/install/template.go
hawkw: Thanks, I think c569f65b58af23e7e7a61f6b34831cdf1c346b77 should address this?

PR #2018, opened 2018-12-20
Add go linting to CI config

klingerf:
This branch starts running [golint](https://github.com/golang/lint) on every CI build. I'm expecting this PR to fail CI with 8 lint warning, which I'll clean up in a subsequent commit. This is a follow-up to #1982.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -eu
+
+cd "$(pwd -P)"
+
+bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rootdir="$( cd $bindir/.. && pwd )"
+
+cd "$rootdir"
+
+# install golint from vendor
+go install ./vendor/golang.org/x/lint/golint
+
+# use `go list` to exclude packages in vendor, ignore uncommented warnings
+out=$(go list ./... | xargs golint | grep -v 'should have comment') || true
+
+if [ -n "$out" ]; then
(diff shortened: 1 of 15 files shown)


bin/lint
```diff
@@ -0,0 +1,23 @@
...
+bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rootdir="$( cd $bindir/.. && pwd )"
+
+cd "$rootdir"
+
+# install golint from vendor
+go install ./vendor/golang.org/x/lint/golint
+
+# use `go list` to exclude packages in vendor, ignore uncommented warnings
+out=$(go list ./... | xargs golint | grep -v 'should have comment') || true
```
siggy: 👍

klingerf:
CI passed 🎉  -- this is now ready for review.

bin/lint
radu-matei: An alternative to this would be to use `golangci-lint run --disable-all -E golint ` - this ignores all comment related linter errors, and in the future we could enable even more linters by adding [a linter config file](https://github.com/golangci/golangci-lint/blob/master/.golangci.yml). If `golint` is the only one needed, then this is perfect.

klingerf: Wow! I had no idea that the go linting rabbit hole goes so deep :)

I wouldn't be opposed to adding additional linting configurations going forward, but I think `golint` minus comments suffices for our use case right now. That said, I'd happily review a PR that improved upon it. Thanks for the suggestion!

Issue #216, opened 2018-01-26
Improve GoDoc


siggy:
Godoc.org automatically crawls our repo and renders docs:
~https://godoc.org/github.com/runconduit/conduit~
https://godoc.org/github.com/linkerd/linkerd2

We could yield more thorough documentation by conforming to GoDoc's commenting:
https://blog.golang.org/godoc-documenting-go-code

We could also add a badge to `README.md`:
~https://godoc.org/github.com/runconduit/conduit?tools~
https://godoc.org/github.com/linkerd/linkerd2?tools

cpretzer:
I'm going to start looking in to this. Are the conduit links in the original Issue still applicable?

Or are these the docs that should be improved: 
https://godoc.org/github.com/linkerd/linkerd2

This is the updated badge link:
https://godoc.org/github.com/linkerd/linkerd2?tools

grampelberg:
This would be awesome!!!

siggy:
@cpretzer Thanks! You're correct about the deprecated conduit URLs. I've updated the ticket description accordingly.

cpretzer:
@grampelberg @siggy I've started working through this and given that it's a non-trivial amount of work, I plan to document files in a piecemeal manner.

This means that there will be several pull requests and to manage that, I'm going to use separate branches of my fork for each pull request and name them "216_go_doc_<year>_<month>_<day>".

Does that sound reasonable to you? I'm happy to work in a single branch on my side if that makes more sense from your perspective.

siggy:
@cpretzer We’ve got a PR up now, #2023, that introduces basic comments for exported symbols, a subset of this ticket. Recommend keeping an eye on that PR to avoid duplicating work. Thanks for jumping in!

cpretzer:
@siggy I appreciate the heads up. Once #2023 is accepted, I'll merge master to my fork and merge the changes to my branch.

I've been focusing on the comment documentation and hadn't even looked at the linting stuff yet

ArthurSens:
Hello there, do you still need help with this? 

If that's the case, I would like to work on it... but I think I would need some guidance. 
Can I reach to someone to talk about this issue?

cpretzer:
@ArthurSens sure, that would be great. The `#contributors` channel in [Linkerd Slack](https://slack.linkerd.io) is a great place to ask for guidance.

coderfornow:
is this available ?

cpretzer:
sure if you're willing to help, that would be great :)

PR #2023, opened 2018-12-22
Enable lint check for comments

siggy:
Part of #217. Follow up from #1982 and #2018.

A subsequent commit will fix the ci failure.

### 2nd commit

Address all comment-related linter errors.

This change addresses all comment-related linter errors by doing the
following:
- Add comments to exported symbols
- Make some exported symbols private
- Recommend via TODOs that some exported symbols should should move or
  be removed

This PR does not:
- Modify, move, or remove any code
- Modify existing comments

### DIFF
--- a/controller/api/proxy/client.go
+++ b/controller/api/proxy/client.go
@@ -5,6 +5,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+// NewClient creates a new gRPC client to the Destination service.
+// TODO: consider moving this into destination-client, or removing altogether.
 func NewClient(addr string) (pb.DestinationClient, *grpc.ClientConn, error) {
 	conn, err := grpc.Dial(addr, grpc.WithInsecure())
 	if err != nil {
--- a/controller/api/public/client.go
+++ b/controller/api/public/client.go
@@ -190,7 +190,8 @@ func newClient(apiURL *url.URL, httpClientToUse *http.Client, controlPlaneNamesp
 	}, nil
 }
 
-// NewInternalClient creates a new Public API client intended to run inside a Kubernetes cluster.
+// NewInternalClient creates a new Public API client intended to run inside a
+// Kubernetes cluster.
 func NewInternalClient(controlPlaneNamespace string, kubeAPIHost string) (pb.ApiClient, error) {
 	apiURL, err := url.Parse(fmt.Sprintf("http://%s/", kubeAPIHost))
 	if err != nil {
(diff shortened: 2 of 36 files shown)


siggy:
Relates to #216.

controller/api/proxy/client.go
```diff
@@ -5,6 +5,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+// NewClient creates a new gRPC client to the Destination service.
+// TODO: consider moving this into destination-client, or removing altogether.
```
klingerf: 👍 Looks like this is only called from the destination-client script, so I'm definitely in favor of removing this file and moving the code directly into the script. But doesn't necessarily have to happen as part of this branch.

controller/api/public/client.go
```diff
@@ -190,6 +190,7 @@ func newClient(apiURL *url.URL, httpClientToUse *http.Client, controlPlaneNamesp
 	}, nil
 }
 
+// NewInternalClient creates a new Public API client intended to run inside a Kubernetes cluster.
```
klingerf: Mind wrapping this at 80 characters? Same goes for the comment on line 203 in this file.

controller/api/public/test_helper.go
```diff
@@ -110,37 +121,41 @@ func (a *MockAPITapByResourceClient) Recv() (*pb.TapEvent, error) {
...
 //
 
-type MockProm struct {
+type mockProm struct {
 	Res             model.Value
 	QueriesExecuted []string // expose the queries our Mock Prometheus receives, to test query generation
 	rwLock          sync.Mutex
 }
 
+// PodCounts is a test helper struct that for representing data in a
```
klingerf: Typo here. Maybe it should be: // PodCounts is a test helper struct that _is used_ for representing data in a

controller/api/util/api_utils.go
```diff
@@ -535,6 +547,8 @@ func routeLabels(event *pb.TapEvent) string {
 	return out
 }
 
+// RenderTapEvent renders a Public API TapEvent to a string.
+// TODO: consider moving this into cli/cmd/tap.go.
```
klingerf: 👍 Ah, right, this was originally moved here when we were rendering tap output as plaintext in the web UI, but we could move it back now that the web has its own rendering.

controller/k8s/api.go
```diff
@@ -140,62 +143,71 @@ func (api *API) Sync() {
...
 
+// Pod provides access to a shared informer and lister for Pods.
 func (api *API) Pod() coreinformers.PodInformer {
 	if api.pod == nil {
 		panic("Pod informer not configured")
 	}
 	return api.pod
 }
 
+// RC provides access to a shared informer and lister for ReplicaSets.
```
klingerf: Typo here: `ReplicaSets` => `ReplicationControllers`

controller/api/proxy/client.go
siggy: Sounds good, will address in a subsequent PR.

controller/api/util/api_utils.go
siggy: Will do in a subsequent PR.

PR #2025, opened 2018-12-27
Introduce resource selector and deprecate namespace field for ListPods

alenkacz:
This is first follow-up for #1777

We introduce ResourceSelector for `ListPods` method and deprecate namespace. Current check code is now using the non-deprecated way of querying `ListPods`.


alenkacz:
cc. @klingerf

controller/api/public/grpc_server_test.go
```diff
@@ -186,6 +189,131 @@ spec:
...
+			k8sAPI,
+			"linkerd",
+			[]string{},
+		)
+
+		k8sAPI.Sync()
+
+		_, _ = fakeGrpcServer.ListPods(context.TODO(), &pb.ListPodsRequest{
+			Selector: &pb.ResourceSelection{
+				Resource: &pb.Resource{
```
dadjeibaah: More of a nit than anything else: I think it would be good to also set the `Namespace` field here to something other than `testnamespace` to further prove that the `Name` field is used instead of `Namespace` when `Type` is Namespace.

controller/api/public/grpc_server.go
```diff
@@ -72,9 +72,23 @@ func (s *grpcServer) ListPods(ctx context.Context, req *pb.ListPodsRequest) (*pb
...
+		return nil, fmt.Errorf("cannot set both namespace and resource in the request. These are mutually exclusive")
+	}
+
 	nsQuery := ""
+	namespace := ""
 	if req.GetNamespace() != "" {
-		nsQuery = fmt.Sprintf("namespace=\"%s\"", req.GetNamespace())
+		namespace = req.GetNamespace()
+	}
+	if req.GetSelector() != nil && req.GetSelector().GetResource().GetNamespace() != "" {
```
klingerf: The intent of the new "selector" field is to filter the list of pods by their owner. For example, if I issue the following `ListPods` request, I'd expect to only get back pods owned by the emoji deployment:

```
req := &pb.ListPodsRequest{
		Selector: &pb.ResourceSelection{
			Resource: &pb.Resource{
				Namespace: "emojivoto",
				Type:      "deployment",
				Name:      "emoji",
			},
		},
	}
```

Maybe we could add a comment to the protobuf definition that indicates the intent of the "selector" field? Specifically, if "selector.resource" is present, it filters by owner resource, whereas if "selector.label_selector" is present, it filters by pod labels (see my other comment about label selectors).

We can't query the kubernetes API by owner, but we're already populating the "owner" field for all pods that are returned by the API, and we can do the filtering there. Something like:

```diff
diff --git a/controller/api/public/grpc_server.go b/controller/api/public/grpc_server.go
index 6026ce1a..1740946d 100644
--- a/controller/api/public/grpc_server.go
+++ b/controller/api/public/grpc_server.go
@@ -76,16 +76,16 @@ func (s *grpcServer) ListPods(ctx context.Context, req *pb.ListPodsRequest) (*pb
 		return nil, fmt.Errorf("cannot set both namespace and resource in the request. These are mutually exclusive")
 	}
 
+	targetOwner := req.GetSelector().GetResource()
+
 	nsQuery := ""
 	namespace := ""
 	if req.GetNamespace() != "" {
 		namespace = req.GetNamespace()
-	}
-	if req.GetSelector() != nil && req.GetSelector().GetResource().GetNamespace() != "" {
-		namespace = req.GetSelector().GetResource().GetNamespace()
-	}
-	if req.GetSelector() != nil && req.GetSelector().GetResource().GetType() == pkgK8s.Namespace {
-		namespace = req.GetSelector().GetResource().GetName()
+	} else if targetOwner.GetNamespace() != "" {
+		namespace = targetOwner.GetNamespace()
+	} else if targetOwner.GetType() == pkgK8s.Namespace {
+		namespace = targetOwner.GetName()
 	}
 	if namespace != "" {
 		nsQuery = fmt.Sprintf("namespace=\"%s\"", namespace)
@@ -124,6 +124,17 @@ func (s *grpcServer) ListPods(ctx context.Context, req *pb.ListPodsRequest) (*pb
 			continue
 		}
 
+		ownerKind, ownerName := s.k8sAPI.GetOwnerKindAndName(pod)
+		if targetOwner.GetNamespace() != "" && targetOwner.GetNamespace() != pod.GetNamespace() {
+			continue
+		}
+		if targetOwner.GetType() != "" && targetOwner.GetType() != ownerKind {
+			continue
+		}
+		if targetOwner.GetName() != "" && targetOwner.GetName() != ownerName {
+			continue
+		}
+
 		updated, added := reports[pod.Name]
 
 		status := string(pod.Status.Phase)
@@ -160,7 +171,6 @@ func (s *grpcServer) ListPods(ctx context.Context, req *pb.ListPodsRequest) (*pb
 			ProxyVersion:        proxyVersion,
 		}
 
-		ownerKind, ownerName := s.k8sAPI.GetOwnerKindAndName(pod)
 		namespacedOwnerName := pod.Namespace + "/" + ownerName
 
 		switch ownerKind {
```

controller/api/public/grpc_server.go
```diff
@@ -94,7 +108,6 @@ func (s *grpcServer) ListPods(ctx context.Context, req *pb.ListPodsRequest) (*pb
 	}
 
 	var pods []*k8sV1.Pod
-	namespace := req.GetNamespace()
 	if namespace != "" {
 		pods, err = s.k8sAPI.Pod().Lister().Pods(namespace).List(labels.Everything())
```
klingerf: Should we also populate the label selector used here based on the "selector" object that's sent with the request?

For example, if I issue the following `ListPods` request, I'd expect to only get back only those pods with the `app: emoji-svc` label:

```
	req := &pb.ListPodsRequest{
		Selector: &pb.ResourceSelection{
			LabelSelector: "app=emoji-svc",
		},
	}
```

You can use `labels.Parse` to parse the label selector string. Something like:

```diff
diff --git a/controller/api/public/grpc_server.go b/controller/api/public/grpc_server.go
index 1740946d..522d2f21 100644
--- a/controller/api/public/grpc_server.go
+++ b/controller/api/public/grpc_server.go
@@ -107,11 +107,20 @@ func (s *grpcServer) ListPods(ctx context.Context, req *pb.ListPodsRequest) (*pb
 		}
 	}
 
+	labelSelector := labels.Everything()
+	if s := req.GetSelector().GetLabelSelector(); s != "" {
+		var err error
+		labelSelector, err = labels.Parse(s)
+		if err != nil {
+			return nil, fmt.Errorf("invalid label selector \"%s\": %s", s, err)
+		}
+	}
+
 	var pods []*k8sV1.Pod
 	if namespace != "" {
-		pods, err = s.k8sAPI.Pod().Lister().Pods(namespace).List(labels.Everything())
+		pods, err = s.k8sAPI.Pod().Lister().Pods(namespace).List(labelSelector)
 	} else {
-		pods, err = s.k8sAPI.Pod().Lister().List(labels.Everything())
+		pods, err = s.k8sAPI.Pod().Lister().List(labelSelector)
 	}
 
 	if err != nil {
```

controller/api/public/grpc_server.go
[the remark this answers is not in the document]
alenkacz: oh, I see that @klingerf made very similar comment, I just missed that because it was not part of this thread 🙈

alenkacz: I thought about adding owner filtering in the next step once we merge this. I tend to like to keep my PRs small. But I can make that change in this PR if it makes more sense to you...

controller/api/public/grpc_server_test.go
alenkacz: @dadjeibaah this is actually very ambiguous :) Now that you're pointing it out... I think that having both namespace and type: Namespace set in the same request should be probably prohibited, right? Because it's not really obvious which one would be used...

controller/api/public/grpc_server_test.go
```diff
@@ -186,6 +189,131 @@ spec:
...
+		err = verifyPromQueries(&mockProm, "testnamespace")
+
+		if err != nil {
+			t.Fatalf("Prometheus queries don't match. Got error: %s", err)
+		}
+	})
+}
+
+type expectedProm struct {
+	expectedPrometheusQueries []string
```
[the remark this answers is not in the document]
alenkacz: yes 👼 thanks

controller/api/public/grpc_server_test.go
dadjeibaah: That's right. I think you actually have done that in the test above called `Fails for both resource and namespace set in request` now that I look at this more So we may not need this. I wasn't looking hard enough :man_facepalming:

controller/gen/common/healthcheck/healthcheck.pb.go
```diff
@@ -1,11 +1,13 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // source: common/healthcheck.proto
 
-package healthcheck // import "github.com/linkerd/linkerd2/controller/gen/common/healthcheck"
```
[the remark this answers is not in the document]
alenkacz: I actually think this is correct. I re-run it locally and it yielded the same results. Then I went to check those files on master and did a `git blame` and it was me who touched those line last times so I think that last time I probably run with old protoc? But I am happy to hear if I am doing something wrong :-)

controller/api/public/grpc_server.go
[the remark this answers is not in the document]
siggy: @alenkacz Ah makes sense re: specifying an owner of the pod, thanks!

controller/api/public/grpc_server_test.go
```diff
@@ -186,6 +189,128 @@ spec:
 			}
 		}
 	})
+
+	t.Run("Fails for both resource and namespace set in request", func(t *testing.T) {
```
klingerf: Hooray for tests! Since the existing test for the `ListPods` endpoint is already setup to use a [table driven test](https://github.com/golang/go/wiki/TableDrivenTests), I'd prefer to add new test cases to the table, instead of new tests entirely. Unfortunately, however, the table driven test setup is a bit lacking for the tests that you're adding. But you should feel free to tweak it to satisfy your use case. For instance, to convert this specific test to a table driven test, I'd do:

```diff
diff --git a/controller/api/public/grpc_server_test.go b/controller/api/public/grpc_server_test.go
index 969e5f23..0708e93b 100644
--- a/controller/api/public/grpc_server_test.go
+++ b/controller/api/public/grpc_server_test.go
@@ -3,6 +3,7 @@ package public
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sort"
 	"strings"
 	"testing"
@@ -19,7 +20,8 @@ type listPodsExpected struct {
 	err     error
 	k8sRes  []string
 	promRes model.Value
-	res     pb.ListPodsResponse
+	req     *pb.ListPodsRequest
+	res     *pb.ListPodsResponse
 }
 
 type listServicesExpected struct {
@@ -42,7 +44,11 @@ func (bs ByService) Len() int           { return len(bs) }
 func (bs ByService) Swap(i, j int)      { bs[i], bs[j] = bs[j], bs[i] }
 func (bs ByService) Less(i, j int) bool { return bs[i].Name <= bs[j].Name }
 
-func listPodResponsesEqual(a pb.ListPodsResponse, b pb.ListPodsResponse) bool {
+func listPodResponsesEqual(a *pb.ListPodsResponse, b *pb.ListPodsResponse) bool {
+	if a == nil || b == nil {
+		return a == b
+	}
+
 	if len(a.Pods) != len(b.Pods) {
 		return false
 	}
@@ -72,7 +78,7 @@ func listPodResponsesEqual(a pb.ListPodsResponse, b pb.ListPodsResponse) bool {
 }
 
 func TestListPods(t *testing.T) {
-	t.Run("Successfully performs a query based on resource type", func(t *testing.T) {
+	t.Run("Queries to the ListPods endpoint", func(t *testing.T) {
 		expectations := []listPodsExpected{
 			listPodsExpected{
 				err: nil,
@@ -142,7 +148,8 @@ spec:
       pod-template-hash: hash-not-meshed
 `,
 				},
-				res: pb.ListPodsResponse{
+				req: &pb.ListPodsRequest{},
+				res: &pb.ListPodsResponse{
 					Pods: []*pb.Pod{
 						&pb.Pod{
 							Name:            "emojivoto/emojivoto-meshed",
@@ -161,6 +168,25 @@ spec:
 					},
 				},
 			},
+			listPodsExpected{
+				err: fmt.Errorf("cannot set both namespace and resource in the request. These are mutually exclusive"),
+				promRes: model.Vector{
+					&model.Sample{
+						Metric:    model.Metric{"pod": "emojivoto-meshed"},
+						Timestamp: 456,
+					},
+				},
+				k8sRes: []string{},
+				req: &pb.ListPodsRequest{
+					Namespace: "test",
+					Selector: &pb.ResourceSelection{
+						Resource: &pb.Resource{
+							Type: pkgK8s.Pod,
+						},
+					},
+				},
+				res: nil,
+			},
 		}
 
 		for _, exp := range expectations {
@@ -179,52 +205,17 @@ spec:
 
 			k8sAPI.Sync()
 
-			rsp, err := fakeGrpcServer.ListPods(context.TODO(), &pb.ListPodsRequest{})
-			if err != exp.err {
+			rsp, err := fakeGrpcServer.ListPods(context.TODO(), exp.req)
+			if !reflect.DeepEqual(err, exp.err) {
 				t.Fatalf("Expected error: %s, Got: %s", exp.err, err)
 			}
 
-			if !listPodResponsesEqual(exp.res, *rsp) {
-				t.Fatalf("Expected: %+v, Got: %+v", &exp.res, rsp)
+			if !listPodResponsesEqual(exp.res, rsp) {
+				t.Fatalf("Expected: %+v, Got: %+v", exp.res, rsp)
 			}
 		}
 	})
 
-	t.Run("Fails for both resource and namespace set in request", func(t *testing.T) {
-		k8sAPI, err := k8s.NewFakeAPI("", make([]string, 0)...)
-		if err != nil {
-			t.Fatalf("NewFakeAPI returned an error: %s", err)
-		}
-
-		fakeGrpcServer := newGrpcServer(
-			&mockProm{Res: model.Vector{
-				&model.Sample{
-					Metric:    model.Metric{"pod": "emojivoto-meshed"},
-					Timestamp: 456,
-				},
-			}},
-			tap.NewTapClient(nil),
-			k8sAPI,
-			"linkerd",
-			[]string{},
-		)
-
-		k8sAPI.Sync()
-
-		_, err = fakeGrpcServer.ListPods(context.TODO(), &pb.ListPodsRequest{
-			Namespace: "test",
-			Selector: &pb.ResourceSelection{
-				Resource: &pb.Resource{
-					Type: pkgK8s.Pod,
-				},
-			},
-		})
-		expectedError := fmt.Errorf("cannot set both namespace and resource in the request. These are mutually exclusive")
-		if err.Error() != expectedError.Error() {
-			t.Fatalf("Expected error: %s, Got: %s", expectedError, err)
-		}
-	})
-
 	t.Run("Reads namespace from resource namespace field", func(t *testing.T) {
 		k8sAPI, err := k8s.NewFakeAPI("", make([]string, 0)...)
 		if err != nil {
```

Hopefully with a few more modifications to that setup, we could satisfy the two other tests that you're adding as well.

PR #2026, opened 2018-12-29
Disallow both pre and proxy flags for check

### LINKED ISSUE #1690: Add validation that will disallow both `pre` and `proxy` checks
Right now, it's possible to issue the following command for linkerd check `linkerd check --pre --proxy` but these two are actually mutually exclusive as it's apparent from the documentation. What might be even more confusing it when you execute it as this `linkerd check --proxy --pre` because only pre-install are executed - that's given by the order of the if-clauses in the codebase https://github.com/linkerd/linkerd2/blob/master/cli/cmd/check.go#L75

I think it makes sense to add validation and if someone tries to trigger both just end with failure.

Another possible solution might be to allow running `pre` and `proxy` through the same invocation but not sure if that's desirable.

alenkacz:
Fixes #1690

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -57,8 +57,8 @@ non-zero exit code.`,
   # Check that the Linkerd data plane proxies in the "app" namespace are up and running
   linkerd check --proxy --namespace app`,
 		Args: cobra.NoArgs,
-		Run: func(cmd *cobra.Command, args []string) {
-			configureAndRunChecks(options)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return configureAndRunChecks(options)
 		},
 	}
 
@@ -73,7 +73,11 @@ non-zero exit code.`,
 	return cmd
 }
 
-func configureAndRunChecks(options *checkOptions) {
+func configureAndRunChecks(options *checkOptions) error {
+	err := options.validate()
+	if err != nil {
+		return fmt.Errorf("Validation error when executing check command: %v", err)
+	}
 	checks := []healthcheck.Checks{healthcheck.KubernetesAPIChecks}
 
 	if options.preInstallOnly {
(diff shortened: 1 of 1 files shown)


cli/cmd/check.go
```diff
@@ -103,14 +107,21 @@ func configureAndRunChecks(options *checkOptions) {
 
 	success := runChecks(os.Stdout, hc)
 
-	fmt.Println("")
```
siggy: This `Println` separates the final result from the checks above it.

```
...
linkerd-version: control plane is up-to-date...............................[ok]

Status check results are [ok]
```

This would have been caught by our integration tests, but unfortunately those do not run on every PR, only master merge.

To run integration tests (make sure you have access to a kubernetes cluster, and can create:
```bash
bin/docker-build-cli-bin
bin/test-run `pwd`/bin/linkerd checktest
```

... or to run a relevant subset:
```bash
bin/docker-build-cli-bin
go test test/install_test.go --linkerd `pwd`/target/cli/darwin/linkerd --integration-tests --linkerd-namespace checktest
```

and then clean up:
```bash
bin/test-cleanup checktest
```

cli/cmd/check.go
```diff
@@ -103,14 +107,21 @@ func configureAndRunChecks(options *checkOptions) {
...
 	}
 
 	fmt.Printf("Status check results are %s\n", okStatus)
+
+	return nil
+}
+
+func (o *checkOptions) validate() error {
+	if o.preInstallOnly && o.dataPlaneOnly {
+		return fmt.Errorf("--pre and --proxy flags are mutually exclusive")
```
siggy: though we're not perfectly consistent with this across the codebase, i'd recommend `errors.New()` if you don't need variable substitution.

cli/cmd/check.go
alenkacz: LOL! I'll push it back with a comment then :D That totally looked like something left my mistake

alenkacz:
@siggy both should be fixed

cli/cmd/check.go
```diff
@@ -103,6 +108,7 @@ func configureAndRunChecks(options *checkOptions) {
 
 	success := runChecks(os.Stdout, hc)
 
+	// this empty line separates final results from the checks list in the output
```
siggy: 👍

PR #2030, opened 2019-01-02
Reduce webpack bundle size: import d3 modules individually

rmars:
Previously, we were importing all of d3, but we don't need a lot of it. This branch switches our imports to only the things we need.

Follows advice here:
https://www.giacomodebidda.com/how-to-import-d3-plugins-with-webpack/

Before:
screen shot 2019-01-02 at 2 24 15 pm

After:
screen shot 2019-01-02 at 3 32 21 pm

```
Running `NODE_ENV=production` yarn webpack produces:
On master:
Assets:
  index_bundle.js (1.19 MiB)

On rmars/no-lo:
Assets:
  index_bundle.js (1.14 MiB)

On this branch:
Assets:
  index_bundle.js (859 KiB)
```


web/app/js/components/NetworkGraph.jsx
```diff
@@ -1,17 +1,35 @@
...
 import _isEmpty from 'lodash/isEmpty';
 import _map from 'lodash/map';
 import _uniq from 'lodash/uniq';
+import { drag } from 'd3-drag';
+import { format } from 'd3-format';
 import { metricsPropType } from './util/MetricUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 import withREST from './util/withREST.jsx';
 
+// create a Object with only the subset of functions/submodules/plugins that we need
```
siggy: 👍

web/app/js/components/NetworkGraph.jsx
```diff
@@ -37,6 +55,9 @@ export class NetworkGraphBase extends React.Component {
 
   constructor(props) {
     super(props);
+
+    // https://github.com/d3/d3-zoom/issues/32
+    d3.getEvent = (() => require("d3-selection" ).event).bind(this);
```
siggy: nit: `require("d3-selection")`

web/app/js/components/NetworkGraph.jsx
```diff
@@ -37,6 +55,9 @@ export class NetworkGraphBase extends React.Component {
 
   constructor(props) {
     super(props);
+
+    // https://github.com/d3/d3-zoom/issues/32
```
siggy: 👀

web/app/js/components/NetworkGraph.jsx
rmars: thanks!

PR #2031, opened 2019-01-02
Fix reporting of injected resources

### LINKED ISSUE #1676: ensure linkerd inject provides stderr feedback for list inputs
`linkerd inject` can take a list of k8s objects, but it appears to only provide stderr feedback to one of the list elements:
```bash
kubectl get -n emojivoto deploy -o yaml \
  | linkerd inject - \
  | kubectl apply -f -
...
hostNetwork: pods do not use host networking...............................[ok]
sidecar: pods do not have a proxy or initContainer already injected........[ok]
supported: at least one resource injected..................................[ok]
udp: pod specs do not include UDP ports....................................[ok]

Summary: 1 of 1 YAML document(s) injected
  deployment/web
```

alpeb:
When reporting resources provided as a list, it was picking just one
item from the list and ignoring the rest.

Fixes #1676

Signed-off-by: Alejandro Pedraza <alejandro.pedraza@gmail.com>

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -390,16 +390,13 @@ func InjectYAML(in io.Reader, out io.Writer, report io.Writer, options *injectOp
 			return err
 		}
 
-		ir := injectReport{}
-		result, err := injectResource(bytes, options, &ir)
+		result, err := injectResource(bytes, options, &injectReports)
 		if err != nil {
 			return err
 		}
 
 		out.Write(result)
 		out.Write([]byte("---\n"))
-
-		injectReports = append(injectReports, ir)
 	}
 
 	generateReport(injectReports, report)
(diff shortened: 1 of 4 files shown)


grampelberg:
This PR gets me *so* excited. Now my demos will make more sense!

alpeb:
Thanks for the review! :+1: 
@siggy  Indeed, passing that slice by ref felt a little weird. I've incorporated your changes.
@klingerf I have enhanced the test as per your instructions :tada:

cli/cmd/inject.go
```diff
@@ -566,20 +571,20 @@ func injectResource(bytes []byte, options *injectOptions, report *injectReport)
...
 		// in the list (instead of just marshaling the injected pod template).
-
-		// TODO: generate an injectReport per list item
-		return injectList(bytes, options, report)
-
+		return injectList(bytes, options)
 	}
 
+	report := injectReport{}
+	report.name = fmt.Sprintf("%s/%s", strings.ToLower(meta.Kind), om.Name)
```
klingerf: TIOLI -- you could set the name field here on initialization, rather than as a separate operation:

```
	report := injectReport{
		name: fmt.Sprintf("%s/%s", strings.ToLower(meta.Kind), om.Name),
	}
```

alpeb: Good catch :+1:

PR #2032, opened 2019-01-02
Followups from lint/comment changes

siggy:
This is a followup branch from #2023:
- delete `proxy/client.go`, move code to `destination-client`
- move `RenderTapEvent` and `GetPercentTLS` from `util` to `cmd`

### DIFF
--- a/controller/api/util/api_utils.go
+++ b/controller/api/util/api_utils.go
@@ -8,7 +8,6 @@ import (
 
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/k8s"
-	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


controller/api/util/api_utils.go
```diff
@@ -672,14 +494,3 @@ func GetSuccessRate(stats *pb.BasicStats) float64 {
 	}
 	return float64(success) / float64(success+failure)
 }
-
-// GetPercentTLS calculates the percent of traffic that is TLS, from Public API
-// BasicStats.
-// TODO: consider moving this into `/cli/cmd/stat.go`.
-func GetPercentTLS(stats *pb.BasicStats) float64 {
```
klingerf: TIOLI -- Even though this func is only called in one place, I'm inclined to leave it defined here, with the `GetRequestRate` and `GetSuccessRate` functions, since they feel so similar. Thoughts? Or maybe we could move all three functions to `cli/cmd/root.go`? They seem like a trio to me.

siggy: Yeah, went back and forth on this. These 3 do feel related. The only difference being `GetPercentTLS` is only called from `stat`, while the other two are called by both `stat` and `routes`.

I'm onboard with moving these to `root.go`. Was slightly concerned about polluting that file in a grab-bag of utils, but I suppose that's at least better than a grab-bag under `controller/api/util/api_utils.go`.

PR #2037, opened 2019-01-03
Surface logs from control plane pods 

### LINKED ISSUE #1994: Surface controller logs in dashboard

When things go wrong in the controller, the controller will usually log an error.  However, these errors are not surfaced to users who are interacting with Linkerd primarily through the CLI and web UI.  Furthermore, even if you are looking for the logs, it can be difficult to know which pod and which container's logs to look at.

Consider adding a "logs" page to the Linkerd web dashboard which displays aggregated logs from all the controller containers.  Ideally with filter controls to select which container (default: all) and what log level (default: error).

A handcrafted `kubectl` command to get logs from the desired containers and an artisanal grep command to select out the appropriate log levels.

dadjeibaah:
When debugging control plane issues or issues pertaining to the linkerd proxy, it can be cumbersome to get logs from affected containers in quickly. 

This PR adds a new command `logs` to the Linkerd CLI to surface log lines from any container within linkerd's control plane. This feature relies heavily on [stern](https://github.com/wercker/stern), which already provides this behavior. This PR integrates this package into the Linkerd CLI to allow users to quickly retrieve logs whenever they run into issues when using Linkerd.  As a follow-up issue, it would be useful to filter logs by log level to avoid users from being inundated with log line output.

```bash
➜  linkerd2 ✗ bin/linkerd logs -h
Tail logs from containers in the Linkerd control plane.

Usage:
  linkerd logs [flags]

Examples:
  # Tail logs from all containers in the prometheus control plane component
  linkerd logs --control-plane-component prometheus

  # Tail logs from the linkerd-proxy container in the grafana control plane component
  linkerd logs --control-plane-component grafana --container linkerd-proxy

  # Tail logs from the linkerd-proxy container in the controller component beginning with the last two lines
  linkerd logs --control-plane-component controller --container linkerd-proxy --tail 2

  # Tail logs from the linkerd-proxy container in the controller component showing timestamps for each line
  linkerd logs --control-plane-component controller --container linkerd-proxy --timestamps


Flags:
  -c, --container string                 Tail logs from the specified container. Options are 'public-api', 'proxy-api', 'tap', 'destination', 'prometheus', 'grafana' or 'linkerd-proxy'
      --control-plane-component string   Tail logs from the specified control plane component. Default value (empty string) causes this command to tail logs from all resources marked with the 'linkerd.io/control-plane-component' label selector
  -h, --help                             help for logs
  -s, --since duration                   Duration of how far back logs should be retrieved (default 48h0m0s)
      --tail int                         Last number of log lines to show for a given container. -1 does not show previous log lines (default -1)
  -t, --timestamps                       Print timestamps for each given log line

Global Flags:
      --api-addr string            Override kubeconfig and communicate directly with the control plane at host:port (mostly for testing)
      --context string             Name of the kubeconfig context to use
      --kubeconfig string          Path to the kubeconfig file to use for CLI requests
  -l, --linkerd-namespace string   Namespace in which Linkerd is installed [$LINKERD_NAMESPACE] (default "linkerd")
      --verbose                    Turn on debug logging
```

```bash
➜  linkerd2  ✗ bin/linkerd logs --control-plane-component web --container web --timestamps
+ linkerd linkerd-web-5856d7cd96-qln4n › web
linkerd linkerd-web-5856d7cd96-qln4n web 2019-01-15T18:34:27.2293628Z time="2019-01-15T18:34:27Z" level=info msg="running version stable-2.1.0"
linkerd linkerd-web-5856d7cd96-qln4n web 2019-01-15T18:34:27.2309416Z time="2019-01-15T18:34:27Z" level=info msg="starting admin server on :9994"
linkerd linkerd-web-5856d7cd96-qln4n web 2019-01-15T18:34:27.2328459Z time="2019-01-15T18:34:27Z" level=info msg="starting HTTP server on :8084"
```

fixes #1994

### DIFF
--- a/cli/Dockerfile-bin
+++ b/cli/Dockerfile-bin
@@ -1,5 +1,5 @@
 ## compile binaries
-FROM gcr.io/linkerd-io/go-deps:844900ed as golang
+FROM gcr.io/linkerd-io/go-deps:8c5ab859 as golang
 WORKDIR /go/src/github.com/linkerd/linkerd2
 COPY cli cli
 COPY controller/k8s controller/k8s
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -1,5 +1,5 @@
 ## compile controller services
-FROM gcr.io/linkerd-io/go-deps:844900ed as golang
+FROM gcr.io/linkerd-io/go-deps:8c5ab859 as golang
 WORKDIR /go/src/github.com/linkerd/linkerd2
 COPY controller/gen controller/gen
 COPY pkg pkg
--- a/proxy-init/Dockerfile
+++ b/proxy-init/Dockerfile
@@ -1,5 +1,5 @@
 ## compile proxy-init utility
-FROM gcr.io/linkerd-io/go-deps:844900ed as golang
+FROM gcr.io/linkerd-io/go-deps:8c5ab859 as golang
 WORKDIR /go/src/github.com/linkerd/linkerd2
 COPY ./proxy-init ./proxy-init
 RUN CGO_ENABLED=0 GOOS=linux go install -v ./proxy-init/
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -26,7 +26,7 @@ COPY web/app .
 RUN $ROOT/bin/web build
 
 ## compile go server
-FROM gcr.io/linkerd-io/go-deps:844900ed as golang
+FROM gcr.io/linkerd-io/go-deps:8c5ab859 as golang
 WORKDIR /go/src/github.com/linkerd/linkerd2
 RUN mkdir -p web
 COPY web/main.go web


cli/cmd/logs.go
```diff
@@ -0,0 +1,208 @@
...
+						Stream()
+
+					if err != nil {
+						return
+					}
+
+					defer stream.Close()
+
+					bufReader := bufio.NewReader(stream)
+					bytes, err := bufReader.ReadBytes('\n')
```
siggy: i think this needs to continually read:
```go
bufReader := bufio.NewReader(stream)
bytes := []byte{}
loglineID := fmt.Sprintf("[%s %s]", p, c)

for {
	bytes, err = bufReader.ReadBytes('\n')
	if err != nil {
		fmt.Printf("Log reader returned: %s\n", err)
		return
	}
	lineRead <- fmt.Sprintf("%s %s", colorPicker.pick(loglineID).Color(loglineID), string(bytes))
}
```

Gopkg.toml
```diff
@@ -58,6 +58,9 @@ required = [
 [[constraint]]
   name = "github.com/go-openapi/spec"
   version = "v0.17.2"
+[[constraint]]
+  name = "github.com/ttacon/chalk"
+  version = "v0.1"
```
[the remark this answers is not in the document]
grampelberg: I've used color in the past and been happy with it.

cli/cmd/logs.go
```diff
@@ -0,0 +1,208 @@
...
+		filterOpts,
+	}, nil
+}
+
+func newCmdLogs() *cobra.Command {
+
+	var containerFilter string
+
+	cmd := &cobra.Command{
+		Use:   "logs (COMPONENT) [flags]",
```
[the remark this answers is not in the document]
adleong: Chances are that the user won't know the pod name (if they did, they could just use `kubectl logs`).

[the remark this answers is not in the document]
adleong: I think it would be super useful to have an argument or flag to filter by log level.  I'd even go so far as to have WARN be the default.  I think the most common use-case is that "something's wrong and I don't know where to look" and starting with WARN logs is probably more useful than INFO.

Gopkg.toml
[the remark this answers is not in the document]
dadjeibaah: @siggy @grampelberg when I looked at Color, it appeared to no longer be maintained. So, I decided to use chalk.

cli/cmd/logs.go
```diff
@@ -0,0 +1,208 @@
...
+	availableColors []chalk.Color
+	lastUsedColor   int
+}
+
+func (c *ColorPicker) pick(id string) chalk.Color {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if color, ok := c.m[id]; !ok {
+		if c.lastUsedColor > len(c.availableColors)-1 {
+			c.lastUsedColor = 1
```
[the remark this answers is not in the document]
dadjeibaah: I think this might have been left over in a refactor I did. Thanks for catching that.

grampelberg:
- Could we more closely match the `kubectl logs` paradigm? In particular:

    ```bash
    -f, --follow=false: Specify if the logs should be streamed.
    --since=0s: Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to
    all logs. Only one of since-time / since may be used.
    --tail=-1: Lines of recent log file to display. Defaults to -1 with no selector, showing all
    log lines otherwise 10, if a selector is provided.
    --timestamps=false: Include timestamps on each line in the log output
    ```

- How are you thinking about integrating this in with proxy logs eventually? It would be nice to   maintain a single command for both control plane and proxy logs.

    ```bash
    --control-plane=true: Show control plane logs
    --proxy=false: Show proxy logs
    --data-plane=false: Show data plane logs
    ```

    None of these are a great solution, maybe `--proxy` isn't required at all?

- It looks like we can just use stern wholesale what do you think about using that instead of doing it ourselves? They've got some good functionality there that I suspect will make our lives better in the long run.

@adleong should we do anything around log filtering here? It feels like that should just be handled server side (thinking about proxy logs in particular). I'd hate to have a cli option for DEBUG logs, but have the server side not configured for it. If we could dynamically configure everything, then that would be *amazing*. Unfortunately, I don't think some of the control plane components could ever be made to support it (looking at you grafana and prometheus).

adleong:
I think it's definitely useful to be able to filter client side.  With the server components having a default log level of INFO and the CLI defaulting to filtering down to WARN.  Perhaps a nice compromise would be that we can detect the log levels of the server components (by parsing their ENV or by adding an annotation perhaps) and have the CLI exit with an error if you try to specify a log level higher than what the server is configured for.  Things are a bit more complex with the proxy because log levels can be set per module, but I think as long as the default log level is INFO, it should work okay.

My preference is to focus on the most common basic use case.  `kubectl logs` is always there as a more advanced tool if you need to get more granular or dive deeper.

grampelberg:
@adleong that makes sense, let's make sure to tackle that as a follow-up issue though.

cli/cmd/logs.go
```diff
@@ -0,0 +1,199 @@
...
+
+	selector, err := labels.Parse(c.labelSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	if c.labelSelector == "" {
+		config.LabelSelector = labels.Everything()
+	}
+	config.LabelSelector = selector
```
siggy: consider avoiding `labels.Parse` if `c.labelSelector == ""`
```go
if c.labelSelector == "" {
	config.LabelSelector = labels.Everything()
} else {
	selector, err := labels.Parse(c.labelSelector)
	if err != nil {
		return nil, err
	}

	config.LabelSelector = selector
}
```

cli/cmd/logs.go
```diff
@@ -0,0 +1,199 @@
...
+	)
+
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case a := <-added:
```
siggy: i think we can follow more closely to stern here:
```go
go func() {
	for p := range added {
```

siggy:
FWIW confirmed issue goes away with stern 1.6.0, but we lose containerState filtering, templating, and color functionality:
https://github.com/linkerd/linkerd2/commit/fce60baa802a75b5906fa470e0bb71315e931d1e

grampelberg:
```bash
  linkerd logs --pod-filter linkerd-controller.* --namespace linkerd
```

I would prefer to stick to the kubectl/stern standard here:

```bash
linkerd logs -l "linkerd.io/control-plane-component: controller" --namespace linkerd
```

Any reason to use this:

```bash
  -c, --container-filter string   Regex string to use for filtering log lines by container name
```

Instead of:

```bash
  -c, --container string           Container name when multiple containers in pod (default ".*")
```

It doesn't appear that we're watching k8s, if I start a pod after `linkerd logs`, it doesn't show up in a running list. How hard would it be to get that behavior?

dadjeibaah:
@grampelberg 
> would prefer to stick to the kubectl/stern standard here:
> linkerd logs -l "linkerd.io/control-plane-component: controller" --namespace linkerd

This is an issue because we already have a global flag tied to the `-l` 
```bash
-l, --linkerd-namespace string   Namespace in which Linkerd is installed [$LINKERD_NAMESPACE] (default "linkerd")
```
> Any reason to use this:
>   -c, --container-filter string   Regex string to use for filtering log lines by container name

I used   `-c, --container-filter` mostly for uniformity. I thought it would be better to have that match up with `--pod-filter`. Maybe we don't need the suffix `filter`? 

> It doesn't appear that we're watching k8s, if I start a pod after linkerd logs...

Hmm, when I run this command. 
```bin/go-run cli/main.go  logs --pod-filter linkerd-controller.*  --container-filter public-api --namespace linkerd```

Then delete the controller pod, I see the new pod come up in the logs. What log query did you run when you ran into this?

**Comment by grampelberg** [comment]
> This is an issue because we already have a global flag tied to the -l

Well that's a bummer. @klingerf what do you think about replacing the meaning of `-l`. The long form (`--selector`) is still valuable if we don't want to use the short form for this.

> I used -c, --container-filter mostly for uniformity. I thought it would be better to have that match up with --pod-filter. Maybe we don't need the suffix filter?

I was honestly just trying to match `stern` and k8s (which are `-c` and `--container`). I don't feel particularly passionate about it though.

> What log query did you run when you ran into this?

```bash
$ bin/go-run cli logs
$ kubectl -n linkerd delete po --all
```

grampelberg:
👍

cli/cmd/logs.go
```diff
@@ -0,0 +1,208 @@
...
+
+	var containerFilter string
+
+	cmd := &cobra.Command{
+		Use:   "logs (COMPONENT) [flags]",
+		Short: "Prints logs for controller components",
+		Long:  `Prints logs for controller components`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts, err := newLogOptions(args, containerFilter, kubeconfigPath, kubeContext)
+
```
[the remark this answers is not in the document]
dadjeibaah: I would like to wait for PR #2033 to land since that will allow me to only check that the Kubernetes API is reachable. `validatePublicAPIClient looks like it does that and checks to see that the control plane is up, which we don't really need to check for with this particular command.

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+type logFlags struct {
+	containerFilter string
+	namespace       string
+	podFilter       string
+	containerState  string
+	labelSelector   string
+	sinceSeconds    time.Duration
+	tail            int64
+	timestamps      bool
+	follow          bool
```
klingerf: Is `follow` unused?

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+//This code replicates most of the functionality in https://github.com/wercker/stern/blob/master/cmd/cli.go
+type logCmdOpts struct {
+	clientset *kubernetes.Clientset
+	*stern.Config
+}
+
+type logFlags struct {
```
klingerf: For consistency with the other commands in this directory, this struct should be called `logsOptions`, and you should provide a `newLogsOptions` func that initializes the `logsOptions` struct with all of the default values, rather than hardcoding them when you add the flags, below. Roughly:

```diff
diff --git a/cli/cmd/logs.go b/cli/cmd/logs.go
index b0065eb5..eb46adac 100644
--- a/cli/cmd/logs.go
+++ b/cli/cmd/logs.go
@@ -22,7 +22,7 @@ type logCmdOpts struct {
 	*stern.Config
 }
 
-type logFlags struct {
+type logsOptions struct {
 	containerFilter string
 	namespace       string
 	podFilter       string
@@ -34,22 +34,35 @@ type logFlags struct {
 	follow          bool
 }
 
-func (c *logFlags) NewSternConfig() (*stern.Config, error) {
+func newLogsOptions() *logsOptions {
+	return &logsOptions{
+		containerFilter: "",
+		namespace:       controlPlaneNamespace,
+		podFilter:       "",
+		containerState:  "running",
+		labelSelector:   "",
+		sinceSeconds:    48 * time.Hour,
+		tail:            -1,
+		timestamps:      false,
+	}
+}
+
+func (o *logsOptions) NewSternConfig() (*stern.Config, error) {
 	config := &stern.Config{}
 
-	podFilterRgx, err := regexp.Compile(c.podFilter)
+	podFilterRgx, err := regexp.Compile(o.podFilter)
 	if err != nil {
 		return nil, err
 	}
 	config.PodQuery = podFilterRgx
 
-	containerFilterRgx, err := regexp.Compile(c.containerFilter)
+	containerFilterRgx, err := regexp.Compile(o.containerFilter)
 	if err != nil {
 		return nil, err
 	}
 	config.ContainerQuery = containerFilterRgx
 
-	containerState, err := stern.NewContainerState(c.containerState)
+	containerState, err := stern.NewContainerState(o.containerState)
 	if err != nil {
 		return nil, err
 	}
@@ -68,28 +81,28 @@ func (c *logFlags) NewSternConfig() (*stern.Config, error) {
 	}
 	config.Template = tmpl
 
-	if c.labelSelector == "" {
+	if o.labelSelector == "" {
 		config.LabelSelector = labels.Everything()
 	} else {
-		selector, err := labels.Parse(c.labelSelector)
+		selector, err := labels.Parse(o.labelSelector)
 		if err != nil {
 			return nil, err
 		}
 		config.LabelSelector = selector
 	}
 
-	if c.tail != -1 {
-		config.TailLines = &c.tail
+	if o.tail != -1 {
+		config.TailLines = &o.tail
 	}
 
-	config.Since = c.sinceSeconds
-	config.Timestamps = c.timestamps
-	config.Namespace = c.namespace
+	config.Since = o.sinceSeconds
+	config.Timestamps = o.timestamps
+	config.Namespace = o.namespace
 
 	return config, nil
 }
 
-func newLogOptions(flags *logFlags, kubeconfigPath, kubeContext string) (*logCmdOpts, error) {
+func newLogOptions(options *logsOptions, kubeconfigPath, kubeContext string) (*logCmdOpts, error) {
 	kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext)
 	if err != nil {
 		return nil, err
@@ -100,7 +113,7 @@ func newLogOptions(flags *logFlags, kubeconfigPath, kubeContext string) (*logCmd
 		return nil, err
 	}
 
-	c, err := flags.NewSternConfig()
+	c, err := options.NewSternConfig()
 	if err != nil {
 		return nil, err
 	}
@@ -112,8 +125,7 @@ func newLogOptions(flags *logFlags, kubeconfigPath, kubeContext string) (*logCmd
 }
 
 func newCmdLogs() *cobra.Command {
-
-	flags := &logFlags{}
+	options := newLogsOptions()
 
 	cmd := &cobra.Command{
 		Use:   "logs [flags]",
@@ -126,7 +138,7 @@ func newCmdLogs() *cobra.Command {
   linkerd logs --pod-filter linkerd-grafana.* --container linkerd-proxy --namespace linkerd
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts, err := newLogOptions(flags, kubeconfigPath, kubeContext)
+			opts, err := newLogOptions(options, kubeconfigPath, kubeContext)
 
 			if err != nil {
 				return err
@@ -136,14 +148,14 @@ func newCmdLogs() *cobra.Command {
 		},
 	}
 
-	cmd.PersistentFlags().StringVarP(&flags.containerFilter, "container", "c", "", "Regex string to use for filtering log lines by container name")
-	cmd.PersistentFlags().StringVar(&flags.labelSelector, "label-selector", "", "kubernetes label selector to retrieve logs from resources that match")
-	cmd.PersistentFlags().StringVar(&flags.containerState, "container-state", "running", "Show logs from containers that are in a specific container state")
-	cmd.PersistentFlags().StringVarP(&flags.namespace, "namespace", "n", controlPlaneNamespace, "String to retrieve logs from pods in the given namespace")
-	cmd.PersistentFlags().StringVarP(&flags.podFilter, "pod-filter", "p", "", "Regex string to use for filtering log lines by pod name")
-	cmd.PersistentFlags().DurationVarP(&flags.sinceSeconds, "since", "s", 48*time.Hour, "Duration of how far back logs should be retrieved")
-	cmd.PersistentFlags().Int64Var(&flags.tail, "tail", -1, "Last number of log lines to show for a given container. -1 does not show any previous log lines")
-	cmd.PersistentFlags().BoolVarP(&flags.timestamps, "timestamps", "t", false, "Print timestamps for each given log line")
+	cmd.PersistentFlags().StringVarP(&options.containerFilter, "container", "c", options.containerFilter, "Regex string to use for filtering log lines by container name")
+	cmd.PersistentFlags().StringVar(&options.labelSelector, "label-selector", options.labelSelector, "kubernetes label selector to retrieve logs from resources that match")
+	cmd.PersistentFlags().StringVar(&options.containerState, "container-state", options.containerState, "Show logs from containers that are in a specific container state")
+	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "String to retrieve logs from pods in the given namespace")
+	cmd.PersistentFlags().StringVarP(&options.podFilter, "pod-filter", "p", options.podFilter, "Regex string to use for filtering log lines by pod name")
+	cmd.PersistentFlags().DurationVarP(&options.sinceSeconds, "since", "s", options.sinceSeconds, "Duration of how far back logs should be retrieved")
+	cmd.PersistentFlags().Int64Var(&options.tail, "tail", options.tail, "Last number of log lines to show for a given container. -1 does not show any previous log lines")
+	cmd.PersistentFlags().BoolVarP(&options.timestamps, "timestamps", "t", options.timestamps, "Print timestamps for each given log line")
 
 	return cmd
 }
```

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.PersistentFlags().StringVarP(&flags.podFilter, "pod-filter", "p", "", "Regex string to use for filtering log lines by pod name")
+	cmd.PersistentFlags().DurationVarP(&flags.sinceSeconds, "since", "s", 48*time.Hour, "Duration of how far back logs should be retrieved")
+	cmd.PersistentFlags().Int64Var(&flags.tail, "tail", -1, "Last number of log lines to show for a given container. -1 does not show any previous log lines")
+	cmd.PersistentFlags().BoolVarP(&flags.timestamps, "timestamps", "t", false, "Print timestamps for each given log line")
+
+	return cmd
+}
+
+func runLogOutput(opts *logCmdOpts) error {
+
```
klingerf: Mind removing these newlines at the beginning of each func in this file? We typically don't include them (go is long enough as is 😉).

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+	podFilter       string
+	containerState  string
+	labelSelector   string
+	sinceSeconds    time.Duration
+	tail            int64
+	timestamps      bool
+	follow          bool
+}
+
+func (c *logFlags) NewSternConfig() (*stern.Config, error) {
```
klingerf: I don't think this func needs to be exported publicly for use outside of the `cli` package. Also, I'd typically call a func that instantiates a struct `new*`, but since this one already has a different type of struct as its receiver, I'd call it `toSternConfig`.

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+
+			if err != nil {
+				return err
+			}
+
+			return runLogOutput(opts)
+		},
+	}
+
+	cmd.PersistentFlags().StringVarP(&flags.containerFilter, "container", "c", "", "Regex string to use for filtering log lines by container name")
```
adleong: This flag is great because it effectively lets you select the control plane component that you care about.  e.g. `linkerd logs -c public-api`.  

We could also consider changing this to not accept a regex and also updating the doc string to say that this must be one of "public-api", "private-api", "tap", "destination", "prometheus", "grafana", or "linkerd-proxy".  This could help with discovery.  This would also allow us to add tab completion as a future improvement.

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.PersistentFlags().StringVar(&flags.labelSelector, "label-selector", "", "kubernetes label selector to retrieve logs from resources that match")
```
adleong: I'd consider removing this flag unless there's a common and compelling use-case.

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.PersistentFlags().StringVar(&flags.containerState, "container-state", "running", "Show logs from containers that are in a specific container state")
+	cmd.PersistentFlags().StringVarP(&flags.namespace, "namespace", "n", controlPlaneNamespace, "String to retrieve logs from pods in the given namespace")
+	cmd.PersistentFlags().StringVarP(&flags.podFilter, "pod-filter", "p", "", "Regex string to use for filtering log lines by pod name")
```
adleong: Potentially useful in an HA control plane, but I think when you get to the point that you're pulling logs from specific pods, you've graduated past `linkerd logs` and should be using `stern` or `kubectl logs`.  I'd consider removing this.

cli/cmd/logs.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.PersistentFlags().DurationVarP(&flags.sinceSeconds, "since", "s", 48*time.Hour, "Duration of how far back logs should be retrieved")
```
adleong: this is cool and handy

cli/cmd/logs.go
```diff
@@ -0,0 +1,224 @@
...
+		Short: "Tail logs from containers in the linkerd control plane",
+		Long:  `Tail logs from containers in the linkerd control plane`,
+		Example: `  # Tail logs from all containers in the prometheus control plane component
+  linkerd logs --control-plane-component prometheus
+
+  # Tail logs from the linkerd-proxy container in the grafana control plane component
+  linkerd logs --control-plane-component grafana --container linkerd-proxy
+
+  # Tail logs from the linkerd-proxy container in the controller component with timestamps
+  linkerd logs --control-plane-component controller --container linkerd-proxy --tail true
```
siggy: `--tail true` should be some integer.

the `linkerd logs --tail N` behavior is already correctly matching stern's, but the help message for the default `-1` doesn't match.

```bash
stern --help
...
--tail int                   The number of lines from the end of the logs to show. Defaults to -1, showing all logs. (default -1)
```

Gopkg.toml
```diff
@@ -55,6 +55,10 @@ required = [
 [[constraint]]
   name = "github.com/go-openapi/spec"
   version = "v0.17.2"
+
+[[constraint]]
+  name = "github.com/wercker/stern"
+  revision = "b04b5491222d9743529cb859a20d34ce2bb763af"
```
siggy: mind adding a comment here, something like:
```
# pin to 1.6.0 until https://github.com/wercker/stern/issues/96 is resolved
```

cli/cmd/logs.go
dadjeibaah: Oops. I am confusing that with `follow` which is done automatically. Good catch.

cli/cmd/logs.go
```diff
@@ -0,0 +1,224 @@
...
+		for _, container := range pod.Spec.Containers {
+			containers = append(containers, container.Name)
+		}
+	}
+	return controlPlaneComponents, containers
+}
+
+func newLogCmdConfig(options *logsOptions, kubeconfigPath, kubeContext string) (*logCmdConfig, error) {
+	// Check that we can call the Kubernetes API
+	_ := cliPublicAPIClient()
```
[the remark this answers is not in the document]
klingerf: I'm actually on the fence about needing this validation. It seems like the `linkerd logs` command might be most helpful when the control plane is not functioning as expected. And in that case we may not successfully be able to connect to the API, and the command will exit with an error here. But I could see the value in fetching container logs, even if the public API is inaccessible. So maybe just omit this?

cli/cmd/logs.go
```diff
@@ -0,0 +1,224 @@
...
+	}, nil
+}
+
+func newCmdLogs() *cobra.Command {
+	options := newLogsOptions()
+
+	cmd := &cobra.Command{
+		Use:   "logs [flags]",
+		Short: "Tail logs from containers in the linkerd control plane",
+		Long:  `Tail logs from containers in the linkerd control plane`,
```
klingerf: For consistency with the help text from other commands (e.g. `dashboard`), we should use capital-L "Linkerd" in both the short and long descriptions here.

cli/cmd/logs.go
```diff
@@ -0,0 +1,224 @@
...
+		Example: `  # Tail logs from all containers in the prometheus control plane component
+  linkerd logs --control-plane-component prometheus
+
+  # Tail logs from the linkerd-proxy container in the grafana control plane component
+  linkerd logs --control-plane-component grafana --container linkerd-proxy
+
+  # Tail logs from the linkerd-proxy container in the controller component with timestamps
+  linkerd logs --control-plane-component controller --container linkerd-proxy --tail 2
```
klingerf: The description of this command and the command itself don't match here. The description says "with timestamps", but the command doesn't include a `--timestamps` flag. It does include a `--tail` flag.

cli/cmd/logs_test.go
```diff
@@ -0,0 +1,49 @@
...
+			expectedErrStr: "container [not-tap] does not exist in control plane [linkerd]",
+			testContainer:  "not-tap",
+		},
+	}
+
+	for _, tt := range flagTestCases {
+		t.Run("container logs selection", func(t *testing.T) {
+			flags := &logsOptions{controlPlaneComponent: tt.testComponent, container: tt.testContainer}
+			_, err := flags.toSternConfig(tt.components, tt.containers)
+			if err != nil && err.Error() != tt.expectedErrStr {
```
klingerf: I think this if condition won't catch the case where something is expected to return an error but instead it returns nil. For instance, if your test case sets `expectedErrStr: "foo"`, but the call to `toSternConfig` returns nil, that test case will still pass, even though it didn't get the error that was expected.

cli/cmd/logs.go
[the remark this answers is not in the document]
dadjeibaah: Very good point. TIL. I thought calling`cliPublicAPIClient()` without any assignment variable would lead other readers to believe that the method was side-effect-y but it looks like it goes against the `go vet`'s suggestion. 

Also good point on the check it may not be necessary. 👍

cli/cmd/logs.go
```diff
@@ -0,0 +1,225 @@
...
+  linkerd logs --control-plane-component prometheus
+
+  # Tail logs from the linkerd-proxy container in the grafana control plane component
+  linkerd logs --control-plane-component grafana --container linkerd-proxy
+
+  # Tail logs from the linkerd-proxy container in the controller component beginning with the last two lines
+  linkerd logs --control-plane-component controller --container linkerd-proxy --tail 2
+
+  # Tail logs from the linkerd-proxy container in the controller component showing timestamps for each line
+  linkerd logs --control-plane-component controller --container linkerd-proxy --timestamps true
```
siggy: sorry i did not catch this the first time, you can just do `--timestamps`, the `true` is redundant.

PR #2038, opened 2019-01-03
Add support for retries

adleong:
This change adds the `isRetryable` property to routes in a service profile.  This causes the proxy to retry failed requests to retryable routes whenever it is able to.  A retry budget may also be added to the service profile which can limit the number of retries.  The `linkerd profile --template` command has been updated to describe these new properties.

The `linkerd routes` command has been updated to differentiate between actual and effective RPS and success rate.  This data is only available for outgoing queries i.e. when the `--to` flag is used.  Effective metrics are the metrics as observed by the client whereas actual metrics are the metrics which describe the requests and responses which are actually sent on the wire.  Adding retries will usually increase the actual RPS and increase the effective success rate while leaving the actual success rate and effective RPS unchanged.

```
$linkerd routes deploy/books --to svc/authors
ROUTE                     SERVICE   EFFECTIVE_SUCCESS   EFFECTIVE_RPS   ACTUAL_SUCCESS   ACTUAL_RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
HEAD /authors/{id}.json   authors              68.28%          2.4rps           45.21%       3.6rps          14ms          40ms          48ms
[UNKNOWN]                 authors               0.00%          0.0rps            0.00%       0.0rps           0ms           0ms           0ms
```


adleong:
## How I tested

1. Install Linkerd control plane and booksapp
2. Run `linkerd routes deploy/webapp --to svc/books`, notice: no useful data
3. Apply service profiles from swagger:
```
curl https://run.linkerd.io/booksapp/webapp.swagger | linkerd profile --open-api - webapp | kubectl apply -f -
curl https://run.linkerd.io/booksapp/authors.swagger | linkerd profile --open-api - authors | kubectl apply -f -
curl https://run.linkerd.io/booksapp/books.swagger | linkerd profile --open-api - books | kubectl apply -f -
```
4. Run `linkerd routes deploy/webapp --to svc/books`: Notice SR on the POST and PUT routes is ~50%
5. Run `linkerd routes deploy/books --to svc/authors`: Notice SR on the HEAD route is ~50%
6. Download the authors profile: `kubectl -n linkerd get sp/authors.default.svc.cluster.local -o yaml > authors-profile.yaml`
7. Edit the HEAD route to be retryable:
```
  - condition:
      method: HEAD
      pathRegex: /authors/[^/]*\.json
    name: HEAD /authors/{id}.json
    isRetryable: true
```
8. Apply the new profile `kubectl apply -f authors-profile.yaml`
9. Run `linkerd routes deploy/books --to svc/authors` notice effective SR go to 100%
10. Run `linkerd routes deploy/webapp --to svc/books` notice SR go to 100%

pkg/profiles/profiles.go
```diff
@@ -18,6 +20,45 @@ type profileTemplateConfig struct {
...
+	routes := make([]*pb.Route, 0)
+	for _, route := range profile.Routes {
+		pbRoute, err := ToRoute(route)
+		if err != nil {
+			return nil, err
+		}
+		routes = append(routes, pbRoute)
+	}
+	budget := DefaultRetryBudget
+	if profile.RetryBudget != nil {
```
seanmonstar: Sorry if this is addressed elsewhere and I couldn't find it: would this be an appropriate place to validate the budget arguments? Would an error here be more visible to a user?

The proxy will already disregard invalid budgets and log at WARN level, which should be visible, but only if the user is looking at the proxy logs...

adleong: Good point.  Probably a good idea to validate here too.  Unfortunately, the suffers from the same problem that validation errors will only be visible if you're looking at the logs.  

We may be able to some limited validation with CRD schema validation too...

seanmonstar: Is it possible for `linkerd inject` or other related command to do the validation then?

grampelberg: It'd be a great addition to the `inject` checks.

grampelberg:
@adleong I can't get `stat` to work on this branch. Installed booksapp, looked at stat and there's nothing there.

adleong:
@grampelberg uh oh.  can you share a screenshot of the stat command and output?

grampelberg:
```bash
$ bin/go-run cli -n booksapp stat deploy
NAME      MESHED   SUCCESS   RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TLS
authors      1/1         -     -             -             -             -     -
books        1/1         -     -             -             -             -     -
traffic      1/1         -     -             -             -             -     -
webapp       3/3         -     -             -             -             -     -
```

grampelberg:
I've replicated my stat bug on stable, so it isn't your code! I'm going to go figure out what the deal is there now .....

jon-walton:
is it feasible to read the `IsRetryable` from an openapi spec? The spec supports extensions (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#specificationExtensions) as part of the operation object (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#operationObject).

We could use something like `x-linkerd-is-retryable`

The reason I ask this is that our openapi spec is auto generated based on the routes in the controllers. It would be annoying to maintain a separate list of routes and their linkerd settings to then patch the json before passing into the linkerd cli

grampelberg:
What do you think about reporting effective by default and adding a `-o wide` option?

```
bin/go-run cli -n booksapp routes deploy/books --to svc/authors
ROUTE                     SERVICE   SUCCESS   RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
HEAD /authors/{id}.json   authors   100.00%  2.2rps         5ms          16ms          19ms
[UNKNOWN]                 authors     0.00%  0.0rps         0ms           0ms           0ms

bin/go-run cli -n booksapp routes deploy/books --to svc/authors -o wide
ROUTE                     SERVICE   SUCCESS   RPS   ACTUAL_SUCCESS  ACTUAL_RPS  LATENCY_P50   LATENCY_P95   LATENCY_P99
HEAD /authors/{id}.json   authors   100.00%  2.2rps         55.14%      4.0rps          5ms          16ms          19ms
[UNKNOWN]                 authors     0.00%  0.0rps          0.00%      0.0rps          0ms           0ms           0ms
```

adleong:
@jon-walton I love this idea!  It should be pretty simple to add this to the `linkerd profile --open-api` command.  Would you mind opening an issue?

adleong:
@grampelberg I dig it.  ⛏

adleong:
@grampelberg I implemented your `-o wide` suggestion.  PTAL

grampelberg:
Oh yeah, I like that!

controller/api/public/top_routes.go
```diff
@@ -13,6 +13,7 @@ import (
 
 const (
 	routeReqQuery             = "sum(increase(route_response_total%s[%s])) by (%s, dst, classification)"
+	actualRouteReqQuery       = "sum(increase(route_actual_response_total%s[%s])) by (%s, dst, classification)"
```
siggy: at the prometheus metrics level and `BasicStats` level, we're re-using the existing response keys for "effective" stats, and adding new keys for "actual". higher up in the abstractions, around `jsonRouteStats` and cli table `headers`, we're instead adding both new "effective" and "actual" keys, and leaving the existing keys empty. what do you think about modifying the higher-level abstractions to match these lower-level ones, such that we'd remove "effective" altogether?

Issue #1696, opened 2018-09-21
use port-forward for linkerd dashboard


siggy:
The `linkerd dashboard` command uses a kubernetes proxy to allow the user to connect from their laptop to the Linkerd dashboard and Grafana.

Instead, port-forward from the user's machine. This has several advantages:
- simple tcp forwarding vs. http proxy
- kubernetes proxy inspects and rewrites http
- Linkerd dashboard and Grafana may be served at `localhost:1234/` root, rather than a long proxy URL, this would help solve the `root_url` Grafana issue discussed in #1692

olix0r:
Part of the goal for using a proxy to the API master is to leverage the TLS infrastructure between the CLI and kubernetes master. If we're going to pursue this, we'll need to ensure that we can secure the connection to the control plane.

grampelberg:
I'm pretty sure that proxy -> api server uses the same TLS infrastructure. The api server -> control plane connection, however, would be a different beast.

olix0r:
I believe @briansmith has plans for securing the connection between the K8S API server and the control plane.  My point is, we shouldn't change this without understanding how the change impacts the TLS plan.

grampelberg:
@briansmith any insight? What would we lose by going to port-forward instead of proxy?

briansmith:
> * simple tcp forwarding vs. http proxy
> * kubernetes proxy inspects and rewrites http

What problems are we running into where the above two points are problematic? I'm not sure we can do anything that depends on there not being an HTTP proxy between us and the dashboard because some networking configurations might insert HTTP proxies (e.g. anti-virus) that we can't control or avoid.

> * Linkerd dashboard and Grafana may be served at `localhost:1234/` root, rather than a long proxy URL, this would help solve the `root_url` Grafana issue discussed in #1692

Like I mentioned in #1692, the main reason we use the k8s proxy and the reason why we have/had plans to expose the control plane as a k8s API extension server is so that we can integrate with the k8s security mechanisms for authentication and authorizing access to the control plane. That is, k8s currently ensures that our control plane is only accessed from within the cluster and/or by authenticated users from outside the cluster. IIUC, supporting the use of just a simple port forward would disable these security protections.

grampelberg:
> IIUC, supporting the use of just a simple port forward would disable these security protections.

[...]

PR #2039, opened 2019-01-04
Proxy grafana requests through web service

### LINKED ISSUE #2003: web service should proxy requests to grafana service
## Feature Request

The web component of the control plane should provide a reverse proxy that proxies requests to Grafana, such that Grafana dashboards can be loaded directly from the web UI.

Right now the web UI and grafana dashboards are served by two different services. This isn't a problem when viewing the dashboards through the kube service proxy, but it will be difficult to maintain if we switch to using port forwarding as described in #1696.

The web service should run a reverse proxy for grafana requests, and forward them to the grafana service.

None offhand, but we should consider anything that avoids having multiple port-forwards.

From the user's perspective, the experience should be mostly the same, but with different URLs in their browser.

klingerf:
This branch forwards all web request to `/grafana/*` to the grafana service. This allows us to serve both the web UI and grafana on the same port, which is a requirement for #1696. This also allows us to simplify some of the frontend code around linking from web to grafana. Fixes #2003.


grampelberg:
<3 @klingerf, looking forward to this one! Mind opening a couple issues on the website side of things to get the docs updated? That and the katacoda scenario will get updated/simplified.

klingerf:
@siggy Ok, fixed those issues. @grampelberg I opened linkerd/website#92 for the docs update. It's kinda vague, so feel free to elaborate as you see fit.

grafana/dashboards/authority.json
```diff
@@ -1129,7 +1129,7 @@
...
...el = parts[0];\n    versionElem.innerHTML += 'Running Linkerd ' + parts[1] + ' (' + parts[0] + ')' + '.<br>';\n  } else {\n    versionElem.innerHTML += 'Running Linkerd ' + localVersion + '.<br>';\n  }\n  var uuid = main.getAttribute(\"data-uuid\");\n\n  fetch('https://versioncheck.linkerd.io/version.json?version='+localVersion+'&uuid='+uuid+'&source=grafana', {\n    credentials: 'include',\n    headers: {\n      \"Content-Type\": \"application/json; charset=utf-8\",\n    },\n  })\n  .then(response => response.json())\n  .then(json => {\n    if (!channel || !json[channel]) {\n      versionElem.innerHTML += 'Version check failed.'\n    } else if (json[channel] === localVersion) {\n      versionElem.innerHTML += 'Linkerd is up to date.';\n    } else {\n      parts = json[channel].split(\"-\", 2);\n      if (parts.length === 2) {\n        versionElem.innerHTML += \"A new \"+parts[0]+\" version (\"+parts[1]+\") is available.\"\n      } else {\n        versionElem.innerHTML += \"A new version (\"+json[channel]+\") is available.\"\n      }\n      versionElem.innerHTML += \" <a href='https://versioncheck.linkerd.io/update' target='_blank'>Update now</a>.\";\n    }\n  });\n});\n</script>",
```
siggy: oof, thank you / sorry you had to do this.

PR #2028, opened 2019-01-02
Reduce webpack bundle size: import lodash by module

rmars:
This branch replaces imports of all of lodash (`import _ from 'lodash'` with per function imports e..g `import _isNil from 'lodash/isNil'`. This reduces the bundle size from~520 kb to ~160kb.

I've also taken the opportunity to replace our use of lodash functions with native js functions where it makes sense (e.g. some maps, concats, sizes).

Before:
screen shot 2019-01-02 at 12 31 18 pm

After:
screen shot 2019-01-02 at 12 29 49 pm

PR #2043, opened 2019-01-04
Fix lodash error on namespace landing page

klingerf:
This is a follow-up to #2028. As part of that branch, we stopped including the lodash "shorthands" feature set as part of our webpack bundle, which means we can no longer use the `_.property`, `_.matches` and `_.matchesProperty` shorthands. Fortunately we were only doing this in one place, and there are eslint rules to prevent us from using them going forward. This branch fixes the invalid shorthand and adds the new eslint rules.

### DIFF
--- a/web/app/.eslintrc
+++ b/web/app/.eslintrc
@@ -1,7 +1,8 @@
 {
   "plugins": [
     "react",
-    "promise"
+    "promise",
+    "lodash"
   ],
   "parser": "babel-eslint",
   "parserOptions": {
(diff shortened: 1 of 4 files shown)


web/app/.eslintrc
```diff
@@ -34,6 +35,9 @@
     "jsx-a11y/click-events-have-key-events": 0,
     "jsx-quotes": 2,
     "keyword-spacing": 2,
+    "lodash/matches-shorthand": [2, "never"],
```
rmars: 👍

Issue #2047, opened 2019-01-07
Dashboard command not working on Debian


tmlbl:
### What is the issue?

On Debian, running the `linkerd2 dashboard` command fails to open the dashboard. Using `kubectl proxy` appears to work fine.

### How can it be reproduced?

Cluster was created on AWS with `kops` using the following command:

```bash
kops create cluster \
	--zones us-west-2a \
	--kubernetes-version 1.10.11 \
	--image kope.io/k8s-1.8-debian-stretch-amd64-hvm-ebs-2018-02-08 \
	--networking amazon-vpc-routed-eni \
	--node-count 3 \
	$CLUSTER_NAME
```

LinkerD was installed with the flags `--tls=optional --proxy-auto-inject`

### Logs, error output, etc

```
Linkerd dashboard available at:
http://127.0.0.1:40309/api/v1/namespaces/linkerd/services/linkerd-web:http/proxy/
Grafana dashboard available at:
http://127.0.0.1:40309/api/v1/namespaces/linkerd/services/linkerd-grafana:http/proxy/
Opening Linkerd dashboard in the default browser
Fontconfig warning: "/etc/fonts/fonts.conf", line 100: unknown element "blank"
ATTENTION: default value of option force_s3tc_enable overridden by environment.
[27892:28073:0107/131232.027227:ERROR:object_proxy.cc(615)] Failed to call method: org.freedesktop.Notifications.GetCapabilities: object_path= /org/freedesktop/Notifications: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.Notifications was not provided by any .service files
[27892:27892:0107/131232.368502:ERROR:account_tracker.cc(242)] AccessTokenFetched error: Invalid credentials (1).
[27892:27892:0107/131244.151191:ERROR:http_bridge.cc(126)] Not implemented reached in virtual void syncer::HttpBridgeFactory::OnSignalReceived()
```

#### `linkerd check` output

```text
kubernetes-api: can initialize the client..................................[ok]
kubernetes-api: can query the Kubernetes API...............................[ok]
kubernetes-api: is running the minimum Kubernetes API version..............[ok]
linkerd-api: control plane namespace exists................................[ok]
linkerd-api: control plane pods are ready..................................[ok]
linkerd-api: can initialize the client.....................................[ok]
linkerd-api: can query the control plane API...............................[ok]
linkerd-api[kubernetes]: control plane can talk to Kubernetes..............[ok]
[...]

PR #2052, opened 2019-01-08
Setup port-forwarding for linkerd dashboard command

### LINKED ISSUE #1696: use port-forward for linkerd dashboard

klingerf:
This branch switches the `linkerd dashboard` command from using the kubernetes service proxy to access the web UI, to instead using port-forward to access the web pod directly. This simplifies the URLs displayed in the dashboard, and it removes our direct dependency on "k8s.io/kubernetes" 🎉 

As part of this change, I fixed a bug with `linkerd dashboard`, where the proxy would not start until after the dashboard had been successfully loaded in a browser (relates to #2047). And I've updated the integration tests to use port-forwarding instead of proxying to access pods running in the test environment.

Fixes #1696.


grampelberg:
Woah, now that's an awesome PR.

klingerf:
Note that there's still a lot of code in the `web/` directory for loading the web UI through the service proxy. If we decide that we're not going to support viewing the web UI via `kubectl proxy`, then we could rip all of that out and it would clean things up a lot. @grampelberg, do you have a preference one way or another?

grampelberg:
@klingerf I'm totally behind removing support for viewing the dashboard it via `kubectl proxy`.

cli/cmd/dashboard.go
```diff
@@ -20,19 +21,25 @@ const (
 
 	// showURL displays dashboard URLs without opening a browser.
 	showURL = "url"
+
+	// webDeployment is the name of the web UI's deployment in the install config
+	webDeployment = "linkerd-web"
+
+	// webPort is the port on which the web UI is served in the web pod
+	webPort = 8084
```
siggy: mind adding a note that `webDeployment` and `webPort` must match values in `template.go`? perhaps eventually these should be shared constants.

pkg/k8s/portforward.go
```diff
@@ -0,0 +1,149 @@
...
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
```
siggy: 👍

pkg/k8s/portforward.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	req := clientset.CoreV1().RESTClient().Post().
+		Resource("pods").
+		Namespace(namespace).
+		Name(podName).
+		SubResource("portforward")
+
+	if localPort == 0 {
+		localPort, err = getLocalPort()
```
siggy: is there a potential timing issue where we call `getLocalPort` in `NewPortForward`, and then that `localPort` is taken before `Run()` is called?

is the reason we're calling `getLocalPort` here to enable `URLFor` to work immediately after calling `NewPortForward`?

pkg/k8s/portforward.go
```diff
@@ -0,0 +1,149 @@
...
+func (pf *PortForward) Run() error {
+	transport, upgrader, err := spdy.RoundTripperFor(pf.config)
+	if err != nil {
+		return err
+	}
+
+	ports := []string{fmt.Sprintf("%d:%d", pf.localPort, pf.remotePort)}
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, pf.method, pf.url)
+
+	fw, err := portforward.New(dialer, ports, pf.stopCh, pf.readyCh, ioutil.Discard, ioutil.Discard)
```
siggy: consider `os.Stdout, os.Stderr`, perhaps when `verbose` is enabled?

Gopkg.toml
```diff
@@ -60,26 +56,18 @@ required = [
...
-
 [[override]]
   name = "github.com/json-iterator/go"
   revision = "f2b4162afba35581b6d4a50d3b8f34e33c144682"
 
 [[override]]
   name = "github.com/Azure/go-autorest"
   revision = "1ff28809256a84bb6966640ff3d0371af82ccba4"
+
+[[override]]
```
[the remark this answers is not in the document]
klingerf: This is a result of kubernetes/apimachinery#46 / golang/dep#1207. We could try upgrading our k8s.io deps to latest and removing these overrides, but I didn't want to open up that can of worms.

pkg/k8s/portforward.go
```diff
@@ -103,10 +107,17 @@ func (pf *PortForward) Run() error {
...
+	if pf.emitLogs {
+		out = os.Stdout
+		errOut = os.Stderr
+	}
+
 	ports := []string{fmt.Sprintf("%d:%d", pf.localPort, pf.remotePort)}
 	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, pf.method, pf.url)
 
-	fw, err := portforward.New(dialer, ports, pf.stopCh, pf.readyCh, ioutil.Discard, ioutil.Discard)
+	fw, err := portforward.New(dialer, ports, pf.stopCh, pf.readyCh, out, errOut)
```
siggy: nice!

```bash
Forwarding from 127.0.0.1:51328 -> 8084
Forwarding from [::1]:51328 -> 8084
Linkerd dashboard available at:
http://127.0.0.1:51328
Grafana dashboard available at:
http://127.0.0.1:51328/grafana
Handling connection for 51328
```

PR #2058, opened 2019-01-10
Hash JS bundle to expire cache

### LINKED ISSUE #1996: Expire web dashboard bundle

When updating to a new version of Linkerd, the dashboard does not expire the old web bundle. This forces me to do a hard refresh on the page to get it working (lots of errors)

Update webpack to add a hash for the bundle, so that when there's a new bundle the browser pulls it down.

It'd be possible to set expiration headers in the web service, just feels more fragile than letting webpack do its thing.

siggy:
JavaScript assets could be cached across Linkerd releases, showing an
out of date ui, or a broken page.

Modify the webpack build pipeline to add a hash to the JS bundle
filename. Move all logic around webpack-dev-server state from Go into
JS, via a templatized index_bundle.js file, generated at build time.
Disable caching of index_bundle.js in Go, via a `Cache-Control` header.

Fixes #1996


web/app/webpack.config.js
```diff
@@ -64,10 +72,17 @@ module.exports = smp.wrap({
...
+      filename: "index_bundle.js.out",
+      template: 'lodash/index_bundle.js.lodash.tmpl',
+    }),
     new LodashModuleReplacementPlugin({
       // 'chain': true,
       'collections': true,
       'paths': true
-    })
+    }),
+    new WebpackMv()
```
rmars: tioli we can name this WebpackMvPlugin to match what seems to be the style of the other plugin names

web/templates/app.tmpl.html
```diff
@@ -13,10 +13,6 @@
 
 {{ define "script-tags" }}
   {{ if not .Contents.Error }}
-    {{ if .Context.WebpackDevServer }}
-      <script type="text/javascript" src="{{.Context.WebpackDevServer}}/dist/index_bundle.js" async></script>
-    {{else}}
-      <script type="text/javascript" src="{{.Contents.PathPrefix}}dist/index_bundle.js" async></script>
-    {{end}}
+    <script type="text/javascript" src="{{.Contents.PathPrefix}}dist/index_bundle.js" async></script>
```
rmars: 😻

klingerf: I think I'm not understanding the setup, but I was expecting the name of this file to change, based on this sentence from the description of this PR:

> Modify the webpack build pipeline to add a hash to the JS bundle filename

Isn't this the JS bundle? I see that you've set cache control headers on this file, so maybe that fixes the issue?

rmars: This isn't the JS bundle - it's a js file that points to the hashed bundle. so the file that actually contains our js would be the hashed bundle and this file, when generated, would look something like:
```
// based on https://gist.github.com/necolas/1025811
(function(doc, script) {
  let js,
    fjs = doc.getElementsByTagName(script)[0],
    add = function(url, id) {
      if (doc.getElementById(id)) {return;}
      js = doc.createElement(script);
      js.src = url;
      id && (js.id = id);
      fjs.parentNode.insertBefore(js, fjs);
    };

  
    let root = fjs.src.substring(0, fjs.src.indexOf("dist"));
    add(root+'dist/main.17249b45c7d23228c299.js');
  

}(document, 'script'));

```

klingerf: Ah, yep, I was definitely not understanding what was going on here. Thanks for the explanation!

web/app/webpack.config.js
```diff
@@ -1,22 +1,30 @@
...
 const path = require('path');
+const CleanWebpackPlugin = require('clean-webpack-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
 const LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
+const WebpackMv = require('./webpack-mv.js');
 
 // analyze plugin speeds
 const SpeedMeasurePlugin = require("speed-measure-webpack-plugin");
-const smp = new SpeedMeasurePlugin();
+// disable by default until this is fixed:
```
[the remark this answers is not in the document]
siggy: i'm totally fine removing `SpeedMeaurePlugin` or disabling it as-is. i haven't used it much, so if you're leaning towards removing it, i'm onboard with that.

web/app/webpack-mv.js
```diff
@@ -0,0 +1,21 @@
...
+      if (compilation.hooks.htmlWebpackPluginAfterHtmlProcessing !== undefined) {
+        compilation.hooks.htmlWebpackPluginAfterHtmlProcessing.tapAsync(
+          'webpack-mv',
+          function(htmlPluginData, cb) {
+            var out = htmlPluginData.plugin.childCompilationOutputName;
+            htmlPluginData.plugin.childCompilationOutputName = out.substring(0, out.indexOf('.out'));
+            cb(null, htmlPluginData);
+          }
+        );
+      }
```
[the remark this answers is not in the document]
siggy: ah good idea about printing a warning. though, if i remove `SpeedMeaurePlugin`, this becomes moot...

web/templates/app.tmpl.html
siggy: Thanks for the explanation @rmars. This conversation reaffirms my thinking that this `app.tmpl.html` -> `index_bundle.js` -> `main.17249b45c7d23228c299.js` dependency chain is overly complex and not very canonical. I think we can roll forward with these changes for now, but I've filed #2074 (thanks @rmars for the idea!) as a way out of this.

siggy:
@rmars re: `Entrypoint undefined = index_bundle.js.out`, unfortunately its a known issue with `HtmlWebpackPlugin`:
https://github.com/jantimon/html-webpack-plugin/issues/895#issuecomment-379006202

web/app/webpack.config.js
```diff
@@ -64,10 +66,17 @@ module.exports = smp.wrap({
   },
   plugins: [
     // new BundleAnalyzerPlugin(), // uncomment to analyze bundle size
+    new CleanWebpackPlugin(['dist']),
+    new HtmlWebpackPlugin({
+      inject: false,
+      filename: "index_bundle.js.out",
+      template: 'index_bundle.js.lodash.tmpl',
```
klingerf: Sorry, one more clarifying question: why does this file have "lodash" in the name? Is there anything lodash-specific about the file or the `HtmlWebpackPlugin` that's using it? Would it make sense to rename it to `index_bundle.js.tmpl`?

siggy: Hmm, `index_bundle.js.lodash.tmpl` is a "lodash template", which is what `HtmlWebpackPlugin` defaults to:
https://github.com/jantimon/html-webpack-plugin/blob/master/docs/template-option.md#1-dont-set-any-loader

That said, according to the lodash docs, this should be equivalent to ES2015 or ERB:
https://lodash.com/docs/4.17.11#templateSettings

I don't feel super strongly about this, was erring on the side explicitness. During development I was trying several non-default `HtmlWebpackPlugin` templating systems, and it was helpful to know what type of templating each file was written for.

klingerf: Ahh, ok, thanks for the additional explanation. Definitely no need to change it. I didn't realize that lodash supported it's own templating and that's what we're using. TIL.

web/app/webpack.config.js
```diff
@@ -64,10 +66,17 @@ module.exports = smp.wrap({
   },
   plugins: [
```
rmars: I think the order of the plugins matters (https://stackoverflow.com/questions/41470771/webpack-does-the-order-of-plugins-matter); not that it matters for the plugins we currently have, but maybe the generation of the index_bundle and the moving of it should be in adjacent steps?

also, wdyt of adding comments in the `plugins` section so that it's more clear what each step does?

```suggestion
  plugins: [
    // new BundleAnalyzerPlugin(), // uncomment to analyze bundle size
    new CleanWebpackPlugin(['dist']),
    new LodashModuleReplacementPlugin({
      // 'chain': true,
      'collections': true,
      'paths': true
    }),
    // compile the bundle with hashed filename into index_bundle.js.out
    new HtmlWebpackPlugin({
      inject: false,
      filename: "index_bundle.js.out",
      template: 'index_bundle.js.lodash.tmpl',
    }),
    // move /dist/index_bundle.js.out to /dist/index_bundle.js
    new WebpackMvPlugin()
  ]
```

Issue #1875, opened 2018-11-27
Profiles feel broken until the first stats are collected


grampelberg:
### What problem are you trying to solve?

After adding a service profile to my cluster, it took quite awhile (60s+) to see the stats in `routes`. In that time, I had no idea whether the profile was valid, whether data was being collected at all and if I'd see anything. Instead, I just saw the following message:

```bash
$ linkerd -n emojivoto routes web-svc
No traffic found.  Does the service have a service profile?  You can create one with the `linkerd profile` command.
```

### How should the problem be solved?

It would be nice to know that there is a profile:

```bash
$ linkerd -n emojivoto routes web-svc
ROUTE       SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TLS
/api/list     0.00%   0.0rps           0ms           0ms           0ms    0%
/api/vote     0.00%   0.0rps           0ms           0ms           0ms    0%
[UNKNOWN]     0.00%   0.0rps           0ms           0ms           0ms    0%
```

adleong:
I think the ideal solution here is to have an API that lists the existence of a profile for the given authorities and lists the routes.  This allows us to:
* populate the full list of routes, even the ones that are getting no traffic
* display a call-to-action when the service profile does not exist

PR #2063, opened 2019-01-10
Add display chips to the resource detail page to provide more detail

rmars:
Trying this out: having "meshed" and "no traffic" badges on the resource detail pages to display a little more info to the user.

This branch also contains some  code cleanup in ResourceDetail.jsx

screen shot 2019-01-09 at 1 19 39 pm
screen shot 2019-01-09 at 1 18 18 pm

### DIFF
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -68,8 +68,8 @@ export class ResourceDetailBase extends React.Component {
       resourceMetrics: [],
       podMetrics: [], // metrics for all pods whose owner is this resource
       neighborMetrics: {
-        upstream: {},
-        downstream: {}
+        upstream: [],
+        downstream: []
       },
       unmeshedSources: {},
       resourceIsMeshed: true,
@@ -200,14 +200,23 @@ export class ResourceDetailBase extends React.Component {
       return <Spinner />;
     }
 
-    let { resourceName, resourceType, namespace } = this.state;
+    let {
+      resourceName,
+      resourceType,
+      namespace,
+      resourceMetrics,
+      unmeshedSources,
+      resourceIsMeshed,
+      neighborMetrics
+    } = this.state;
+
     let query = {
       resourceName,
       resourceType,
       namespace
     };
 
-    let unmeshed = _filter(this.state.unmeshedSources, d => d.type === this.state.resourceType)
+    let unmeshed = _filter(unmeshedSources, d => d.type === resourceType)
       .map(d => _merge({}, emptyMetric, d, {
         unmeshed: true,
         pods: {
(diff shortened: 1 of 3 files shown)


web/app/js/components/ResourceDetail.jsx
```diff
@@ -21,6 +21,9 @@ import _reduce from 'lodash/reduce';
 import { processNeighborData } from './util/TapUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 
+// if there has been no traffic for some time, show a warning
+const showNoTrafficMsgDelay = 1000;
```
dadjeibaah: I think it would be good to be more explicit about how long of a duration `showNoTrafficMsgDelay` is. What do you thinking about changing it to `showNoTrafficMsgDelayMs`?

dadjeibaah: Also, should the duration be longer? What criteria should we use to pick the right duration? Is this variable used for debugging purposes to indicate if something is wrong or is it purely informative?

rmars: whoooooops I changed it to 1 sec for testing and forgot to change it back. I think a 5-10 second delay would be more appropriate!

rmars: basically I don't want it to be flickering if there's intermittent traffic (low RPS from the service) but I do want it to appear soon enough so that the user sees it before they leave the empty page in confusion

grampelberg: Should we have an API for this? It would be nice to know that we got metrics at *some* point, even though they're not in the current window. Feels related-ish to #1875.

rmars: yes! same idea. it would be nice to know if we got metrics at some point. we could add this into the metrics response somehow.... or maybe put this in the get api

rmars:
@dadjeibaah which meshed resources box did you mean? the octopus graph boxes? or the tables?

web/app/js/components/ResourceDetail.jsx
klingerf: I agree -- seems like we could introduce an API that does a historical prometheus query for this information. Trying to determine based on current traffic seems difficult, and it can lead to the flickering issue that @rmars mentioned:



screenrecording20190111at2

rmars:
@dadjeibaah yeah it would be cool to add some sort of visual indicator to the octopus graph boxes that would indicate "meshed" status better. Currently if a box appears with a traffic bar, that resource is meshed; if it is unmeshed but is still sending traffic, it appears in the unmeshed box. This is pretty subtle, would be better to call it out.

I think at the top level of the page it might still be good to indicate whether the resource you are viewing is meshed or not / has traffic or not so I'd still leave the top indicator (unless it gets really crowded after updating the octopus graph views).

I've updated the variable name and the delay timing in case @siggy wants to merge this PR before releasing :)

rmars:
@dadjeibaah @klingerf @grampelberg what are your thoughts on merging this and implementing via api as a follow-up vs just waiting until we have an api? 
I've raised the timeout to 6 seconds so there won't be (obvious) flickering unless the rps is that low

klingerf:
@rmars Yep, that totally works for me, and sorry I didn't weigh in pre-merge. This looks good!

grampelberg:
@rmars same from me, incremental progress is the best of the progresses =)

PR #2070, opened 2019-01-11
Wire up grafana proxying in web dev environment

klingerf:
Now that the web service proxies requests to grafana (#2039), we can display grafana dashboards in the web dev environment by adding an additional port-forward to the grafana pod at startup. This will allow us to access grafana at http://localhost:8084/grafana, and the icons in the web UI will link correctly.

### DIFF
--- a/bin/web
+++ b/bin/web
@@ -14,7 +14,10 @@ USAGE: web <command>
   * build - build the assets
   * dev - run a dev server (with live reloading). Options:
       -p $DEV_PORT
-  * port-forward - forward the controller from a running k8s cluster to localhost
+  * port-forward-controller - forward the public-api container from the
+      controller pod in a running k8s cluster to localhost:8085
+  * port-forward-grafana - forward the grafana container from the
+      grafana pod in a running k8s cluster to localhost:3000
   * run - run a local server (sans. reloading)
   * setup - get the environment setup for development
       Note: any command line options are passed on to yarn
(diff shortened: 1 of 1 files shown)


alpeb:
I'm testing `bin/web port-forward-grafana` and it appears something's going on:

image
Grafana is returning 404 for urls like `http://localhost:3000/grafana/public/build/manifest.4f5454f867a0cc2fe8dd.js` presumably because of the `/grafana` part for which there is no proxy in this case that can trim that out.
(Although I didn't understand why there's need for that trimming given we have `/grafana` in `root_url`)

bin/web
```diff
@@ -54,21 +66,34 @@ function build {
...
 }
 
-function port-forward {
+function port-forward-controller {
+  nc -z localhost 8085 || \
+    kubectl --namespace=linkerd port-forward $(get-pod controller) 8085:8085
+}
+
+function port-forward-grafana {
   nc -z localhost 8085 || \
```
siggy: s/8085/3000

siggy:
@alpeb regarding those 404s, are you running a recent master (post #2039) as your deployed linkerd on k8s?

alpeb:
@siggy yep. Does `bin/web port-forward-grafana` work for you?

siggy:
@alpeb ah, i don't think `bin/web port-forward-grafana` is intended to be run on its own, but rather in the context of `bin/web run`, or at least next to a web dashboard Go process running on localhost. as you pointed out, Grafana is configured to be served at `/grafana`, which only works when the web process it proxying for you via https://github.com/linkerd/linkerd2/blob/a27bb2e0cea7f52eae9310b5b1b158d60e4463ed/web/srv/grafana_proxy.go#L21

@klingerf maybe a document fixup indicating the context required to run `bin/web port-forward-grafana` independently, or, don't give devs the option to?

klingerf:
Ah, good feedback. I've removed the `port-forward-grafana` command in favor of a more generic `port-forward` command. Hopefully that will help to prevent confusion going (port-)forward ;)

PR #2078, opened 2019-01-14
 Modify all health checks to be specified via enums

siggy:
Modify all health checks to be specified via enums

The set of health checks to be executed were dependent on a combination
of check enums and boolean options.

This change modifies the health checks to be governed strictly by a set
of enums. This change does not add or remove any checks, but rather
moves checks into more granular categories, such that any set of checks
that are toggle-able are defined together under a single category.

This is a first step in cleaning up the `linkerd check` code, and moving towards #1471.

Next steps:
- tightly couple category IDs to names
- tightly couple checks to their parent categories
- programmatic control over check ordering

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

Issue #1471, opened 2018-08-16
Revisit linkerd check output format


klingerf:
Per #1081, we should streamline the linkerd check output to be less verbose when checks are passing, and more verbose when checks are failing. We should also group the checks as Pre-Install and Post-Install. Relates to #1470.

rmars:
Step 1: get some design feedback on the overall UX of the CLI

wmorgan:
Related: #1653, #1656

siggy:
As part of this work, we should consider embellishing cli output with emoji/color/lol, as mentioned in TGIK8s:

emoji and color in linkerd check:
https://youtu.be/1dXJ34-KLe8?t=2739

ascii art / color / lolgopher for linkerd dashboard:
https://youtu.be/1dXJ34-KLe8?t=2797
https://github.com/kris-nova/lolgopher

grampelberg:
Also related, #1739.

grampelberg:
Here's a proposal for updated formating:

```bash
$ linkerd check

kubernetes-api
---------------
can initialize the client                                                     ✔️
can query the Kubernetes API                                                  ✔️

RBAC
----
can create Namespaces                                                         ❌
    Missing permissions to create Namespace
    See https://linkerd.io/2/troubleshooting#rbac for hints

Version
-------
cli is up-to-date                                                             ✔️
data plane is up-to-date                                                      ⚠️
    You are running 18.11.1 and the latest version is 18.11.2
    See https://linkerd.io/2/troubleshooting#version-mismatch for hints
```

In addition, it would be nice to add a spinner when we are retrying (instead of outputting another line). Something like https://github.com/briandowns/spinner.

siggy:
Relates to #1653 and #1739.

PR #2080, opened 2019-01-14
Group checkers by category

siggy:
The `linkerd check` command organized the various checks via loosely
coupled category IDs, category names, and checkers themselves, all with
ordering defined by consumers of this code.

This change removes category IDs in favor of category names, groups all
checkers by category, and enforces ordering at the `HealthChecker`
level.

Part of #1471, depends on #2078.


grampelberg:
This looks great to me. My only commentary would be around format, but I assume that'll happen in another PR =)

PR #2098, opened 2019-01-17
Refactor `install`'s `render` output to be helm compatible

kleimkuhler:
# Problem
In order to refactor `install` to allow for a more flexible configuration, we
should start with the format of the YAML that it renders. Using the Helm
YAML format will make it easier add flexible configuration options in the
future. Currently, the rendered template that `install` produces does not
follow this format.

# Solution
Use the internals that Helm itself uses to render an inject template that
follows the same formatting rules. Helm's `template` cmd provides a good
outline of what is needed to make Linkerd's `install` cmd work as if it was
a Chart.

# Validation
There are no new tests, but there may not be anything to test at this stage.
This is a WIP PR towards the ultimate goal of `install` allowing a more
flexible configuration.

However, `install` now uses all the Helm `template` internals and therefore
satisfies the needed properties for Helm Charts.


cli/cmd/install.go
```diff
@@ -205,56 +219,155 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+
 	if config.EnableTLS {
-		tlsTemplate, err := template.New("linkerd").Parse(install.TLSTemplate)
-		if err != nil {
-			return err
-		}
-		err = tlsTemplate.Execute(buf, config)
-		if err != nil {
-			return err
+		if err := loadTemplate(chrt, tlsTemplateName, tlsTemplatePath); err != nil {
```
grampelberg: It'll be important from a chart perspective to handle all the configuration inside of the chart and not load separate templates.

cli/cmd/install.go
```diff
@@ -205,56 +219,155 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+	chrt.Metadata = metadata
 
-	return InjectYAML(buf, w, ioutil.Discard, injectOptions)
+	return nil
 }
 
-func (options *installOptions) validate() error {
-	if _, err := log.ParseLevel(options.controllerLogLevel); err != nil {
-		return fmt.Errorf("--controller-log-level must be one of: panic, fatal, error, warn, info, debug")
+func loadTemplate(chrt *chart.Chart, tmplName string, tmplPath string) error {
```
grampelberg: Is there a helm helper to just load a whole directory (or chart) at once?

cli/cmd/install.go
```diff
@@ -205,56 +219,155 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+	caps := &chartutil.Capabilities{
+		APIVersions:   chartutil.DefaultVersionSet,
+		KubeVersion:   chartutil.DefaultKubeVersion,
+		TillerVersion: tversion.GetVersionProto(),
 	}
 
-	return options.proxyConfigOptions.validate()
+	renderOpts := renderutil.Options{
+		ReleaseOptions: chartutil.ReleaseOptions{
+			Name:      "release-name",
```
grampelberg: We'll want this to be `linkerd` (and then include the release name in the chart as that is the best practice standard).

cli/cmd/install.go
```diff
@@ -205,56 +219,155 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+			IsInstall: true,
+			IsUpgrade: false,
+			Time:      timeconv.Now(),
+			Namespace: "default",
```
grampelberg: Shouldn't this be off the cli args?

cli/cmd/install.go
```diff
@@ -205,56 +219,155 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+		},
+		KubeVersion: "",
+	}
+
+	rawValues, err := yaml.Marshal(config)
+	if err != nil {
+		return chartutil.Values{}, err
+	}
+
+	chrtConfig := &chart.Config{Raw: string(rawValues), Values: map[string]*chart.Value{}}
```
grampelberg: We're going to definitely want to map the cli args to values here.

cli/cmd/install.go
kleimkuhler: This is fixed.

cli/cmd/install.go
kleimkuhler: This is fixed.

kleimkuhler:
@grampelberg @klingerf I've addressed most of what we talked about after talking yesterday. I can make some additional changes for certain structs to be populated with CLI args, but I also am hesitant in making this a large PR. If we can talk about more logical chunks that make sense, it is currently in a passing state.

cli/cmd/install.go
kleimkuhler: @grampelberg Yes, I think some of these fields could be.

```
Name: "linkerd"
Namespace: controlPlaneNamespace
```

`controlPlaneNamespace` is set in _root.go_

`IsInstall` and `IsUpgrade` are what I am not sure about. With the current way that we install Linkerd, is it not always considered an install? Even when we upgrade, we still run the install command.

cli/cmd/install.go
kleimkuhler: We're not doing anything different with the CLI arguments so they should not be passed in here as it currently stands. `installOptions` have already been validated when passed into `render()` and then just like before, we still just pass them through to `InjectYAML()`

grampelberg:
I'm getting this on `dep ensure`:

```
grouped write of manifest, lock and vendor: failed to export github.com/aokoli/goutils: fatal: failed to unpack tree object 41ac8693c5c10a92ea1ff5ac3a7f95646f6123b0
```

Seems weird, am I doing something wrong?

kleimkuhler:
> I'm getting this on `dep ensure`:
> 
> ```
> grouped write of manifest, lock and vendor: failed to export github.com/aokoli/goutils: fatal: failed to unpack tree object 41ac8693c5c10a92ea1ff5ac3a7f95646f6123b0
> ```
> Seems weird, am I doing something wrong?

Make sure you are running `bin/dep` instead of just `dep`. If you are doing that, trying deleting `vendor/` and then running it again.

cli/cmd/install.go
```diff
@@ -205,34 +214,64 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
 	}
-	buf := &bytes.Buffer{}
-	err = template.Execute(buf, config)
+
+	chrtConfig := &chart.Config{Raw: string(rawValues), Values: map[string]*chart.Value{}}
+
+	// Create chart and render templates
+	gp := os.Getenv("GOPATH")
+	cp := path.Join(gp, "/src/github.com/linkerd/linkerd2/cli/install")
+	chrt, err := chartutil.Load(cp)
```
kleimkuhler: @grampelberg @klingerf This fixes the `bin/go-run cli install` issue. However, this doesn't seem like the proper way to parameterize the path that Linkerd is installed. Do you know a more idiomatic way to get the install path? It is tough to consider using string constants like we currently do, because `Load()` is expecting a directory that contains a `Chart.yaml` and subdirectories such as `templates/`.

grampelberg: Wouldn't we want to use https://github.com/gobuffalo/packr for this instead of dealing with the path?

kleimkuhler: We will be addressing a follow-up change to the build scripts so that we can investigate use of packr. It will change how we build linkerd, so we are trying to keep the changes in separate PRs.

grampelberg:
👍 works for me!

kleimkuhler:
Who should provide review sign-off for this?

Gopkg.lock
```diff
@@ -977,6 +1054,24 @@
   pruneopts = ""
   revision = "4242d8e6c5dba56827bb7bcf14ad11cda38f3991"
 
+[[projects]]
+  digest = "1:a96ffeefef3ae5a7eb243ce0ac10a0d9b844336e8a360cc1404b2e7903166d55"
+  name = "k8s.io/helm"
```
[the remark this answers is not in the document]
kleimkuhler: Ah good cool - will do.

cli/cmd/install.go
```diff
@@ -78,9 +84,13 @@ type installOptions struct {
 }
 
 const (
-	prometheusProxyOutboundCapacity = 10000
 	defaultControllerReplicas       = 1
 	defaultHAControllerReplicas     = 3
+	prometheusProxyOutboundCapacity = 10000
```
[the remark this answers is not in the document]
kleimkuhler: Sure. As a quick explanation, I was just alphabetizing them. It doesn't seem that their original order is in the order that they are used either. Not that there needs to be one, but any reason to revert it besides keeping the diff smaller?

cli/cmd/install.go
```diff
@@ -209,34 +219,72 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+	if _, err := log.ParseLevel(options.controllerLogLevel); err != nil {
+		return fmt.Errorf("--controller-log-level must be one of: panic, fatal, error, warn, info, debug")
+	}
+
+	if options.proxyAutoInject && options.singleNamespace {
+		return fmt.Errorf("The --proxy-auto-inject and --single-namespace flags cannot both be specified together")
+	}
+
+	return options.proxyConfigOptions.validate()
+}
```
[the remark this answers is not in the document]
kleimkuhler: Similar explanation here: moved `validate()` below `validateAndBuildConfig()` because it is only used by that function and `render()` is used after.

cli/cmd/install.go
```diff
@@ -209,34 +219,72 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
-	err = template.Execute(buf, config)
+
+	chrtConfig := &chart.Config{Raw: string(rawValues), Values: map[string]*chart.Value{}}
+
+	// Load chart files
+	files := []*chartutil.BufferedFile{}
+	files = append(files, &chartutil.BufferedFile{Name: "Chart.yaml", Data: install.Chart})
+	files = append(files, &chartutil.BufferedFile{Name: baseTemplateName, Data: install.BaseTemplate})
+	files = append(files, &chartutil.BufferedFile{Name: tlsTemplateName, Data: install.TLSTemplate})
+	files = append(files, &chartutil.BufferedFile{Name: proxyInjectorTemplateName, Data: install.ProxyInjectorTemplate})
```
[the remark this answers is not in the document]
kleimkuhler: Ah nice!

cli/install/template.go
```diff
@@ -1,46 +1,52 @@
 package install
 
-// Template provides the base template for the `linkerd install` command.
-const Template = `{{ if not .SingleNamespace -}}
+// Chart provides the Chart.yaml file needed to validate chart metadata
+var Chart = []byte(`apiVersion: "v1"
+name: "linkerd"
+version: 0.1.0`)
```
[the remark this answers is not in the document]
kleimkuhler: Yep I can add in `appVersion` for this. I can also add a few additional chart values that would be helpful since this should not just be a _minimal_ working chart.

cli/cmd/install.go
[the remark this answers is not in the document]
klingerf: Yep, I had a similar motivation behind both of these comments. It's no problem to rearrange things, but it's safer to do that in a separate PR that doesn't include functional changes.

cli/install/template.go
```diff
@@ -940,37 +941,37 @@ spec:
...
-          secretName: {{.ProxyInjectorTLSSecret}}
+          secretName: {{.Values.ProxyInjectorTLSSecret}}
           optional: true
       - name: proxy-spec
         configMap:
           name: linkerd-proxy-injector-sidecar-config
 
----
 ### Proxy Injector Service Account ###
+---
```
alpeb: Having the `---` after the comment results in the comment not being included in the final output. You did the same in other places, but only here is the comment being stripped out.

kleimkuhler: Yep this seemed to be an effect of the template engine consuming the next comment if the previous resource ends with a `status: {}`. The entire `template.go` is a mix of comments being consumed or displayed like this.

When I changed all the template comments to appear above the `---`, this was the result of me doing that. I'm not sure why it happens; I didn't investigate too much since we are modularizing these templates soon anyways (hopefully).

To reduce the diff, I did revert that change so thanks for pointing it out. It was not necessary!

PR #2099, opened 2019-01-17
Add cli command to install linkerd service profiles

### LINKED ISSUE #1901: Introduce routes into default control plane install
## Feature Request

We recently introduced Routes support into the [cli](#1860) and [dashboard](#1868).Displaying routes is still a few steps involving deploying an app with an associated service profile.

The output from `linkerd install` should include service profiles with route information for the control plane components, to better demonstrate and dogfood this new functionality.

Eventually this could be autogenerated from our proto files, but hard-coding for now is reasonable.

siggy:
This introduces a `linkerd install-sp` command to install service
profiles into the linkerd control plane. It installs one service profile
for each of controller-api, proxy-api, prometheus, and grafana.

Fixes #1901

### DIFF
--- a/cli/cmd/install-sp.go
+++ b/cli/cmd/install-sp.go
@@ -1,0 +1,49 @@
+package cmd
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"text/template"
+
+	"github.com/linkerd/linkerd2/cli/installsp"
+	"github.com/spf13/cobra"
+)
+
+type installSPConfig struct {
+	Namespace string
+}
+
+func newCmdInstallSP() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install-sp [flags]",
+		Short: "Output Kubernetes configs to install Linkerd Service Profiles",
+		Long: `Output Kubernetes configs to install Linkerd Service Profiles.",
+
+This command installs Service Profiles into the Linkerd control plane. A
+cluster-wide Linkerd control-plane is a prerequisite. To confirm Service Profile
+support, verify "kubectl api-versions" outputs "linkerd.io/v1alpha1".`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return renderSP(os.Stdout, controlPlaneNamespace)
+		},
+	}
+
+	return cmd
+}
+
+func renderSP(w io.Writer, namespace string) error {
+	template, err := template.New("linkerd").Parse(installsp.Template)
+	if err != nil {
+		return err
+	}
+	buf := &bytes.Buffer{}
+	err = template.Execute(buf, map[string]string{"Namespace": namespace})
+	if err != nil {
+		return err
+	}
+
+	w.Write(buf.Bytes())
+	w.Write([]byte("---\n"))
+
+	return nil
+}
--- a/cli/installsp/template.go
+++ b/cli/installsp/template.go
@@ -4,11 +4,9 @@ package installsp
 const Template = `apiVersion: linkerd.io/v1alpha1
 kind: ServiceProfile
 metadata:
-  creationTimestamp: null
   name: linkerd-controller-api.{{.Namespace}}.svc.cluster.local
   namespace: {{.Namespace}}
 spec:
-  retryBudget: null
   routes:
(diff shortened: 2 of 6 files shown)


grampelberg:
Is there any reason to not just do this by default as part of `linkerd install`?

siggy:
@grampelberg i'd love to include this as part of `linkerd install`, unfortunately there's a race condition where service profile support via the `linkerd.io/v1alpha1` apiextension (also installed via `linkerd install`) is not available by the time `kubectl apply` gets around to installing the service profiles themselves.

fwiw, istio works around this by having all CRDs / apiextensions installed in a preliminary step:
https://github.com/knative/serving/issues/2195
https://istio.io/docs/setup/kubernetes/quick-start/#installation-steps

i opted to keep the default `linkerd install` as a single step, which i hope will have the added benefit of educating users that installing service profiles is a thing you do, generally, after you've gotten linkerd up and running.

cli/installsp/template.go
```diff
@@ -0,0 +1,135 @@
+package installsp
+
+// Template provides the base template for the `linkerd install-sp` command.
+const Template = `apiVersion: linkerd.io/v1alpha1
+kind: ServiceProfile
+metadata:
+  creationTimestamp: null
```
adleong: I think you can omit all of the null fields.

siggy:
@adleong i did consider the approach of the control-plane posting its own service profiles. i opted for this approach due to the added complexity and added permissions it would require.

adleong:
That's a good point.

I love that our install is just a single command, but I also love that our control plane has no write permissions :sob:

But giving the control plane write permissions just on the service profile resource type doesn't seem too crazy though...

grampelberg:
@adleong let's handle that as a separate issue/PR. @siggy convinced me that it is a good idea to have a distinct command as is (really helps educate what service profiles are and gives the solid impression of dogfooding).

cli/installsp/template.go
```diff
@@ -0,0 +1,127 @@
+package installsp
+
+// Template provides the base template for the `linkerd install-sp` command.
+const Template = `apiVersion: linkerd.io/v1alpha1
+kind: ServiceProfile
+metadata:
+  name: linkerd-controller-api.{{.Namespace}}.svc.cluster.local
+  namespace: {{.Namespace}}
```
codeman9: If the command is executed before the `linkerd install` command, the `linkerd` namespace (or whatever namespace was chosen) won't exist. Will k8s accept this .yaml in that case? Or will it error out and say that `namespace doesn't exist`?

grampelberg: `linkerd install-sp` would be called *after* `linkerd install`. The service profiles are just nice to have (as they get the per-route metrics).

rmars:
@dadjeibaah yeahh the routes metrics problem should be fixed in #2112

siggy:
@dadjeibaah re: `linkerd install --control-plane-sp`, we definitely considered that approach, i like how it avoids introducing a new command. i opted for `install-sp` to keep `install` strictly for installing the control plane.

cli/cmd/install-sp.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/spf13/cobra"
+)
+
+type installSPConfig struct {
+	Namespace string
+}
+
+func newCmdInstallSP() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install-sp [flags]",
```
adleong: Consider a more verbose command name for clarity like `install-profiles` or `install-service-profiles`.

dadjeibaah: Ditto. I like `install-service-profiles` as it fully describes what the command is going to do when I run it.

siggy: I agree these names better convey what's going on. I just implemented it, and the length feels out of sync with the rest of the commands (`install-sp` is already tied with `completion` for longest command).

Given that rest of our subcommands are all extremely short, and probably require the user to run `linkerd` or `linkerd subcmd --help` to understand what each does anyway, I'm inclined to leave it as-is. For comparison, kubectl's longest cmd is `rolling-update`, and most are much shorter and a single word.

PR #2112, opened 2019-01-18
Update Top Routes in dashboard to use new API, add `to` query options

rmars:
The recent routes API changes caused the Top Routes tab to stop working, as it wasn't looking for the changed structure of the response. This PR updates that page to accept the new API response.

This PR also adds `to` fields to the Top Routes query form, so that the equivalent of `linkerd routes deploy --to deploy/authors` will work in the dashboard.

screen shot 2019-01-18 at 3 32 16 pm

### DIFF
--- a/web/app/js/components/TopRoutes.jsx
+++ b/web/app/js/components/TopRoutes.jsx
@@ -1,4 +1,5 @@
 import { UrlQueryParamTypes, addUrlProps } from 'react-url-query';
+
 import Button from '@material-ui/core/Button';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
@@ -24,6 +25,7 @@ import _pick from 'lodash/pick';
 import _uniq from 'lodash/uniq';
 import _upperFirst from 'lodash/upperFirst';
 import { groupResourcesByNs } from './util/MetricUtils.jsx';
+import { tapResourceTypes } from './util/TapUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 import { withStyles } from '@material-ui/core/styles';
 
@@ -31,12 +33,19 @@ const topRoutesQueryProps = {
   resource_name: PropTypes.string,
   resource_type: PropTypes.string,
   namespace: PropTypes.string,
+  to_name: PropTypes.string,
+  to_type: PropTypes.string,
+  to_namespace: PropTypes.string,
 };
 const topRoutesQueryPropType = PropTypes.shape(topRoutesQueryProps);
 
 const urlPropsQueryConfig = _mapValues(topRoutesQueryProps, () => {
   return { type: UrlQueryParamTypes.string };
 });
+
+const toResourceName = (query, typeKey, nameKey) => {
+  return `${query[typeKey] || ""}${!query[nameKey] ? "" : "/"}${query[nameKey] || ""}`;
+};
 
 const styles = theme => ({
   root: {
@@ -61,6 +70,9 @@ class TopRoutes extends React.Component {
       resource_name: '',
       resource_type: '',
       namespace: '',
+      to_name: '',
+      to_type: '',
+      to_namespace: ''
     },
   }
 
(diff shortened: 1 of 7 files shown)


web/app/js/components/TopRoutes.jsx
```diff
@@ -24,20 +25,28 @@ import _pick from 'lodash/pick';
...
+  to_namespace: PropTypes.string,
 };
 const topRoutesQueryPropType = PropTypes.shape(topRoutesQueryProps);
 
 const urlPropsQueryConfig = _mapValues(topRoutesQueryProps, () => {
   return { type: UrlQueryParamTypes.string };
 });
 
+const toResourceName = (query, typeKey, nameKey) => {
+  return `${query[typeKey] || ""}${!query[nameKey] ? "" : "/"}${query[nameKey] || ""}`;
```
siggy: 👍

web/app/js/components/TopRoutes.jsx
```diff
@@ -241,48 +267,53 @@ class TopRoutes extends React.Component {
...
-    let key = "resource_name";
     let servicesWithPrefix = services
       .filter(s => s.namespace === query.namespace)
       .map(svc => `service/${svc.name}`);
     let otherResources = resourcesByNs[query.namespace] || [];
 
-    let dropdownOptions = servicesWithPrefix.concat(otherResources).sort();
-    let dropdownVal = _isEmpty(query.resource_name) || _isEmpty(query.resource_type) ? "" :
-      query.resource_type + "/" + query.resource_name;
+    let dropdownOptions = servicesWithPrefix
```
siggy: i think the lines immediately above here need to reference `query.to_namespace` when creating the `To Resource` dropdown, otherwise the dropdown items are filtered by the wrong namespace:

screen shot 2019-01-22 at 10 40 48 am

web/app/js/components/util/CliQueryUtils.js
```diff
@@ -1,9 +1,11 @@
 import _compact from 'lodash/compact';
 import _isNil from 'lodash/isNil';
 
-const topRoutesDisplayOrder = [
+const topRoutesDisplayOrder = query => _compact([
   "namespace",
-];
+  "toResource",
+  _isNil(query.to_type) ? null : "toNamespace",
```
siggy: i don't see a way to clear the "To Namespace" once it's been selected? not a big deal, maybe for a subsequent PR?

web/app/js/components/TopRoutes.jsx
```diff
@@ -155,22 +167,26 @@ class TopRoutes extends React.Component {
...
-    this.handleUrlUpdate("resource_name", resource_name);
-    this.handleUrlUpdate("resource_type", resource_type);
+    let [resourceType, resourceName] = resource.split("/");
+
+    resourceName = resourceName || "";
+    query[nameKey] = resourceName;
+    this.handleUrlUpdate(nameKey, resourceName);
+
+    query[typeKey] = resourceType;
+    this.handleUrlUpdate(typeKey, resourceType);
```
siggy: nit, maybe for another PR, what about modifying `handleUrlUpdate` to take all of `state.query`, and then calling it once at the end of both `handleNamespaceSelect` and `handleResourceSelect`?

web/app/js/components/util/CliQueryUtils.js
rmars: yeah, I think it would be good to add a Clear button similar to Tap and Top; I'll follow up with another PR

web/app/js/components/TopRoutes.jsx
rmars: ahhh good catch!

rmars:
@adleong great meta-questions!
- it could be sufficient to have these in the CLI. I do like the web options such as sorting by column, the ability to add links so we can click around to other resource types based on table output. ideally we would try to make the entering of input on this form easier though. perhaps we can make a query to see which services have service profiles installed, and try to present those options instead? I agree that perusing the form options won't tell you much right away
- I still need to add Clear functionality to that form. currently the only way to clear the inputs are to clear the url params
- perhaps in the clear follow up branch, I can add indicators that the `to` fields are optional

it would be awesome to make this page a little more use case driven. perhaps when the user arrives on this page, they instead see a list of services that have SPs installed, ordered by success rate, and can do a one-click to start querying them? or maybe just a little NUX about what service profiles are and how they are being used?

rmars:
I've filed #2138 to keep track of those improvements. I'm going to merge this so that the Top Routes tabs on the ResourceDetail pages don't remain broken

PR #2087, opened 2019-01-15
Update check and inject output

siggy:
The outputs of the `check` and `inject` commands did not vary much
between successful and failed executions, and were a bit verbose and
challenging to parse.

Reorganize output of `check` and `inject` commands, to provide more
output when errors occur, and less output when successful.

Specific changes:

`linkerd check`
- visually group checks by category
- introduce `hintURL`'s, to provide doc links when checks fail
- add spinners when retrying, remove addtional retry lines
- colored emojis to indicate success/warning/failure

`linkerd inject`
- modify default output to mirror `kubectl apply`
- only output non-successful inject reports
- support `--verbose` flag to output all inject reports

Fixes #1471, #1653, #1656, #1739

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

# Example outputs

Note the unicode symbols are rendered with color in the terminal:

screen shot 2019-01-15 at 3 35 08 pm

```
✔
✘
⚠
```

## `linkerd check` with retries

check-spinner4

## `linkerd check --pre`

```
$ bin/linkerd check --pre --expected-version=$(bin/root-tag)
kubernetes-api
--------------
✔ can initialize the client
✔ can query the Kubernetes API

kubernetes-version
------------------
✔ is running the minimum Kubernetes API version

pre-kubernetes-cluster-setup
----------------------------
✘ control plane namespace does not already exist
    The "linkerd" namespace already exists
✔ can create Namespaces
✔ can create ClusterRoles
✔ can create ClusterRoleBindings
✔ can create CustomResourceDefinitions

pre-kubernetes-setup
--------------------
✔ can create ServiceAccounts
✔ can create Services
✔ can create Deployments
✔ can create ConfigMaps

linkerd-version
---------------
✔ can determine the latest version
✔ cli is up-to-date

Status check results are ✘
```

## `linkerd check`
```
$ bin/linkerd check --expected-version=$(bin/root-tag)
kubernetes-api
--------------
✔ can initialize the client
✔ can query the Kubernetes API

kubernetes-version
------------------
✔ is running the minimum Kubernetes API version

linkerd-existence
-----------------
✔ control plane namespace exists
✔ controller pod is running
✔ can initialize the client
✔ can query the control plane API

linkerd-api
-----------
✔ control plane pods are ready
✔ can query the control plane API
✔ [kubernetes] control plane can talk to Kubernetes
✔ [prometheus] control plane can talk to Prometheus

linkerd-service-profile
-----------------------
✔ no invalid service profiles

[...]

PR #2114, opened 2019-01-20
Fix Unicode and color output for Windows

siggy:
The default font in Windows console did not support the Unicode
characters recently added to check and inject commands. Also the color
library the linkerd cli depends on was not being used in a
cross-platform way.

Replace the existing Unicode characters used in `check` and `inject`
with characters available in most fonts, including Windows console.
Similarly replace the spinner used in `check` with one that uses
characters available in most fonts.

Modify `check` and `inject` to use `color.Output` and `color.Error`,
which wrap `os.Stdout` and `os.Stderr`, and perform special
tranformations when on Windows.

Add a `--no-color` option to `linkerd logs`. While stern uses the same
color library that `check`/`inject` use, it is not yet using the
`color.Output` API for Windows support. That issue is tracked at:
https://github.com/wercker/stern/issues/69

Relates to https://github.com/linkerd/linkerd2/pull/2087

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -55,7 +55,7 @@ non-zero exit code.`,
   linkerd check --proxy --namespace app`,
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return configureAndRunChecks(options)
+			return configureAndRunChecks(stdout, options)
 		},
 	}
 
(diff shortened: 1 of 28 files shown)


alpeb:
Looks good on my crappy windows 10 box with Powershell

whatsapp image 2019-01-22 at 1 40 42 pm

grampelberg:
Looks great to me.

PR #2122, opened 2019-01-22
Follow up to #2054: move magic strings into constants

alpeb:


siggy:
Follow up to #2054.

cli/cmd/install_test.go
```diff
@@ -39,6 +39,8 @@ func TestRender(t *testing.T) {
 		CreatedByAnnotation:              "CreatedByAnnotation",
 		ProxyAPIPort:                     123,
 		EnableTLS:                        true,
+		TLSTrustAnchorVolumeName:         "linkerd-trust-anchors",
+		TLSSecretsVolumeName:             "linkerd-secrets",
```
siggy: recommend setting these to something like `TLSTrustAnchorVolumeName`/`TLSSecretsVolumeName`, to disambiguate that the templating is using the injected values and not the constants.

alpeb: Ok gotcha :+1:

Issue #2090, opened 2019-01-15
ServiceProfile for grpc routes regex


balchua:
When working with grpc based routes, often the grpc endpoint is designated as /package-name/MethodName.  In cases like this com.example.User/GetUserById, one needs to escape the forward slash, resulting pathRegex should be com.example.User\/GetUserById.  ServiceProfile documentation is lacking, perhaps when it is available, we could add it there.

klingerf:
Thanks for reporting this, @balchua. We're working on a feature that will allow you to generate service profile templates directly from protobuf definitions, which should hopefully help to prevent some of the confusion. There's more info in #1425.

adleong:
Hi @balchua!  Can you give me an example of a `pathRegex` and a path that doesn't match it, but should?  Forward slash (`/`) is not a special character and shouldn't need to be escaped.

balchua:
What worked for me in the end is escaping forward slash.  Which also makes me wonder since it's not a special character like you said.  I tried to play with different regex, the one i posted above is the exact sample i used.  
I forgot to mention that i even applied the initial regex and i can still see the service profile stats.

adleong:
@balchua here is the service profile that I'm testing with:

```
$ kubectl -n linkerd get sp/voting-svc.emojivoto.svc.cluster.local -oyaml
apiVersion: linkerd.io/v1alpha1
kind: ServiceProfile
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"linkerd.io/v1alpha1","kind":"ServiceProfile","metadata":{"annotations":{},"name":"voting-svc.emojivoto.svc.cluster.local","namespace":"linkerd"},"spec":{"routes":[{"condition":{"method":"POST","pathRegex":"/emojivoto\\.v1\\.VotingService/VotePoop"},"name":"grpc-vote"}]}}
  creationTimestamp: "2019-01-25T18:00:16Z"
  generation: 1
  name: voting-svc.emojivoto.svc.cluster.local
  namespace: linkerd
  resourceVersion: "158237"
  selfLink: /apis/linkerd.io/v1alpha1/namespaces/linkerd/serviceprofiles/voting-svc.emojivoto.svc.cluster.local
  uid: 123fda02-20cb-11e9-b480-025000000001
spec:
  routes:
  - condition:
      method: POST
      pathRegex: /emojivoto\.v1\.VotingService/VotePoop
    name: grpc-vote
```

[...]

PR #2128, opened 2019-01-23
Add --proto flag to linkerd profile command to read protobuf files

### LINKED ISSUE #1425: Read Service Profile from protobuf
#1418 describes Service Profiles which describe how a service is expected to behave.  For gRPC services, it should be possible to generate a profile from that service's protobuf IDL file.  Service behaviors can be inferred from the proto rpc signatures (e.g. methods with `stream` responses should report latency differently than ones with unary responses) and from annotations in the proto file:

```
service Destination {
  // @Retryable
  rpc Get(GetDestination) returns (stream Update) {}
}
```

e.g. the `//@Retryable` annotation means that the method is safe to be retried.

adleong:
Fixes #1425 

```
$ curl -sL https://raw.githubusercontent.com/BuoyantIO/emojivoto/master/proto/Emoji.proto | linkerd profile -n emojivoto --proto - emoji-svc
kind: ServiceProfile
metadata:
  creationTimestamp: null
  name: emoji-svc.emojivoto.svc.cluster.local
  namespace: linkerd
spec:
  routes:
  - condition:
      method: POST
      pathRegex: /emojivoto.v1.EmojiService/ListAll
    name: ListAll
  - condition:
      method: POST
      pathRegex: /emojivoto.v1.EmojiService/FindByShortcode
    name: FindByShortcode
```


grampelberg:
Doesn't the path need to be escaped, ala. #2090?

cli/cmd/root.go
```diff
@@ -80,7 +81,7 @@ func init() {
 	RootCmd.AddCommand(newCmdInject())
 	RootCmd.AddCommand(newCmdInstall())
 	RootCmd.AddCommand(newCmdLogs())
-	RootCmd.AddCommand(newCmdProfile())
+	RootCmd.AddCommand(profile.NewCmdProfile(controlPlaneNamespace))
```
siggy: for consistency with the rest of the cmd's, how about leaving `cmd/profile.go`, and then having that pass `controlPlaneNamespace` to the `cmd/profile` package?

or, going a step further, maybe it makes sense to move everything under `cmd/profile` down into `pkg/profiles`? this would make `renderOpenAPI` and `renderProto` consistent with `RenderProfileTemplate`

related, you could avoid making `profileOptions` public by a slight change to the render functions:
```go
func RenderOpenAPI(input, name, namespace string, controlPlaneNamespace string, w io.Writer) error {}
func RenderProto(input, name, namespace string, controlPlaneNamespace string, w io.Writer) error {}
```

adleong:
@grampelberg yes, good catch!  Forward slashes don't need to be, but other special characters do, such as `.`.

cli/cmd/profile/profile.go
```diff
@@ -0,0 +1,129 @@
...
+If the --template flag is specified, it outputs a service profile template.
+Edit the template and then apply it with kubectl to add a service profile to
+a service.
+
+Example:
+  linkerd profile -n emojivoto --template web-svc > web-svc-profile.yaml
+  # (edit web-svc-profile.yaml manually)
+  kubectl apply -f web-svc-profile.yaml
+
+If the --open-api flag is specified, it reads the given OpenAPI
```
siggy: i realize some of this was already here, but for consistency with the other commands, everything under `Example:` should be indented. consider combining everything under a single `Examples:` header, to match the way cobra prints `Usage:`

```go
		Long: `
...
Examples:
  linkerd profile -n emojivoto --template web-svc > web-svc-profile.yaml
  # (edit web-svc-profile.yaml manually)
  kubectl apply -f web-svc-profile.yaml

  If the --open-api flag is specified, it reads the given OpenAPI
  specification file and outputs a corresponding service profile.

  linkerd profile -n emojivoto --open-api web-svc.swagger web-svc | kubectl apply -f -

  If the --proto flag is specified, it reads the given protobuf specification
  file and outputs a corresponding service profile.

  linkerd profile -n emojivoto --proto Voting.proto vote-svc | kubectl apply -f -`,
```

cli/cmd/profile/proto.go
```diff
@@ -0,0 +1,78 @@
...
+	proto.Walk(definition, handle)
+
+	profile := sp.ServiceProfile{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      fmt.Sprintf("%s.%s.svc.cluster.local", options.name, options.namespace),
+			Namespace: controlPlaneNamespace,
+		},
+		TypeMeta: meta_v1.TypeMeta{
+			APIVersion: "linkerd.io/v1alpha1",
+			Kind:       "ServiceProfile",
```
siggy: this `linkerd.io/v1alpha1` / `ServiceProfile` pair occurs 6 times around the codebase, consider a constant.

PR #2131, opened 2019-01-23
Don't show the tap icon for requests from sources that are not meshed

### LINKED ISSUE #1723: don't display tap icon when source resource isn't meshed
Currently the "top" table in a service profile page has a tap icon next to every row. However, because not all upstream resources are necessarily meshed, we can't actually tap all requests. The tap icon should only be shown for requests that can be tapped, i.e. those requests where the source is meshed.

alpeb:
Fixes #1723

For testing, I injected all the emojivoto deployments but vote-bot. You can see the requests coming from vote-bot are not tappable anymore:

image

### DIFF
--- a/web/app/js/components/TapLink.jsx
+++ b/web/app/js/components/TapLink.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, path}) => {
+const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, path, disabled}) => {
   let params = {
     autostart: "true",
     namespace,
@@ -12,7 +12,7 @@ const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, pa
   };
   let queryStr = Object.entries(params).map(([k, v]) => `${k}=${v}`).join("&");
 
-  if (resource === "") {
+  if (disabled) {
     return <i className="fas fa-microscope tapGrayed" />;
   }
 
@@ -24,6 +24,7 @@ const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, pa
 };
 
 TapLink.propTypes = {
+  disabled: PropTypes.bool.isRequired,
   namespace: PropTypes.string.isRequired,
   path: PropTypes.string.isRequired,
   PrefixedLink: PropTypes.func.isRequired,
--- a/web/app/js/components/TapLink.jsx
+++ b/web/app/js/components/TapLink.jsx
@@ -24,7 +24,7 @@ const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, pa
 };
 
 TapLink.propTypes = {
-  disabled: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool,
   namespace: PropTypes.string.isRequired,
   path: PropTypes.string.isRequired,
   PrefixedLink: PropTypes.func.isRequired,
@@ -32,5 +32,9 @@ TapLink.propTypes = {
   toNamespace: PropTypes.string.isRequired,
   toResource: PropTypes.string.isRequired,
 };
+
+TapLink.defaultProps = {
+  disabled: false
+}
 
 export default TapLink;
--- a/web/app/js/components/TapLink.jsx
+++ b/web/app/js/components/TapLink.jsx
@@ -12,6 +12,10 @@ const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, pa
   };
   let queryStr = Object.entries(params).map(([k, v]) => `${k}=${v}`).join("&");
 
+  if (resource === "") {
+    return <i className="fas fa-microscope tapGrayed" />;
+  }
+
   return (
     <PrefixedLink to={`/tap?${queryStr}`}>
       <i className="fas fa-microscope" />
--- a/web/app/js/components/TopModule.jsx
+++ b/web/app/js/components/TopModule.jsx
@@ -1,4 +1,4 @@
-import { processTapEvent, setMaxRps, wsCloseCodes } from './util/TapUtils.jsx';
+import { processNeighborData, processTapEvent, setMaxRps, wsCloseCodes } from './util/TapUtils.jsx';
 
 import ErrorBanner from './ErrorBanner.jsx';
 import Percentage from './util/Percentage.js';
@@ -27,8 +27,8 @@ class TopModule extends React.Component {
       resource: PropTypes.string
     }),
     startTap: PropTypes.bool.isRequired,
-    updateNeighbors: PropTypes.func,
-    updateTapClosingState: PropTypes.func
+    updateTapClosingState: PropTypes.func,
+    updateUnmeshedSources: PropTypes.func
   }
 
   static defaultProps = {
@@ -38,8 +38,8 @@ class TopModule extends React.Component {
     // - un-ended tap results, pre-aggregation into the top counts
     // - aggregated top rows
     maxRowsToStore: 50,
-    updateNeighbors: _noop,
     updateTapClosingState: _noop,
+    updateUnmeshedSources: _noop,
     query: {
       resource: ""
     }
@@ -51,6 +51,7 @@ class TopModule extends React.Component {
     this.topEventIndex = {};
     this.throttledWebsocketRecvHandler = _throttle(this.updateTapEventIndexState, 500);
     this.updateTapClosingState = this.props.updateTapClosingState;
+    this.unmeshedSources = {};
 
     this.state = {
       error: null,
@@ -97,7 +98,6 @@ class TopModule extends React.Component {
 
   onWebsocketRecv = e => {
     this.indexTapResult(e.data);
-    this.props.updateNeighbors(e.data);
     this.throttledWebsocketRecvHandler();
   }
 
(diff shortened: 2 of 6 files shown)


grampelberg:
Why can't we tap the vote-bot requests?

alpeb:
@grampelberg From what I understand, the `tap` API accepts a resource and optionally a target (TO) resource. Currently it doesn't have the option to specify a FROM.
So in order to tap requests from `vote-bot` to `web`, we're not tapping `web` but actually `vote-bot`:

`linkerd tap deployment/vote-bot --namespace emojivoto --to deployment/web --to-namespace emojivoto`

That requires `vote-bot` to be meshed, which is not in the example above.

grampelberg:
`--from` sounds like a pretty valuable filter for tap. @adleong you wouldn't happen to have any history there would you?

adleong:
I believe the problem is that we don't have source metadata for requests.  We can tap requests coming into the web deployment, but we can't tell which of them came from the vote-bot.

web/app/js/components/TopModule.jsx
```diff
@@ -237,7 +239,8 @@ class TopModule extends React.Component {
     }
 
     if (d.base.proxyDirection === "INBOUND") {
-      this.props.updateNeighbors(d.requestInit.source, _get(d, "requestInit.sourceMeta.labels"), null);
+      let unmeshedSources = this.props.updateNeighbors(d.requestInit.source, _get(d, "requestInit.sourceMeta.labels"), null);
+      topResults[eventKey].meshed = !_has(unmeshedSources, d.base.source.owner);
```
rmars: this check doesn't quite work when the resource we're doing the Top on is a `pod`. E.g. with web meshed, if you go to the Web's Resource Detail page for pod (e.g. http://localhost:8084/namespaces/emojivoto/pods/web-6b4675554d-ldtzm) you'll see that the tap icon continues to appear for all rows.

I think you'd need something like:


```suggestion
      let isPod = this.props.query.resource.split("/")[0] === "pod";
      if (isPod) {
        topResults[eventKey].meshed = !_has(unmeshedSources, "pod/" + d.base.source.pod);
      } else {
        topResults[eventKey].meshed = !_has(unmeshedSources, d.base.source.owner);
      }
```

alpeb: Ok, understood :+1:

alpeb:
@rmars thanks for the detailed feedback. You provided above a diff for `Top.jsx`, but your comment states we should move `updateNeighborsFromTapData` into `TopModule`. Was that a mistake, or am I totally misunderstanding?
On the other hand, I think we still need `updateNeighborsFromTapData` to be in `ResourceDetail.jsx` because `<Octupus>` uses `unmeshedSources` which is populated as a side-effect of having that function live in that class...

rmars:
@alpeb yes! indeed! I was saying that once we implemented the above functionality in Top (i.e. my diff), we would notice that both `Top` and `ResourceDetail` pages would have similar `updateNeighborsFromTapData` methods, and that maybe we could then consider moving the `updateNeighborsFromTapData` functionality into `TopModule` so that the logic is in one place. 

(`TopModule` would still need to pass the unmeshedSources it collected to its parent component, though, since ResourceDetail needs it to render the Octopus graph. But the calling of the `processNeighborData` could be put in one place.) It might be that it doesn't make sense to consolidate them, though, and leaving two `updateNeighborsFromTapData` would be fine too!

alpeb:
Thanks for the clarification @rmars.

I moved `updateNeighborsFromTapData` into `TopModule` and now the Top and ResourceDetail pages are properly hiding the tap icons.

Now `<Top>` is receiving as an attribute the `updateUnmeshedSources` function instead of `updateNeighbors`, which is used to propagate the `unmeshedSources` object into the parent components.                                                                                             

The logic for updating the `meshed` property has been updated to take into account pods, like you suggested.

Let me know what you think :thinking:

web/app/js/components/TopModule.jsx
```diff
@@ -97,7 +98,6 @@ class TopModule extends React.Component {
 
   onWebsocketRecv = e => {
     this.indexTapResult(e.data);
-    this.props.updateNeighbors(e.data);
```
alpeb: It appears this wasn't doing anything (since not sending anything as a secong arg) so I removed it.

web/app/js/components/TopModule.jsx
```diff
@@ -26,8 +27,8 @@ class TopModule extends React.Component {
       resource: PropTypes.string
     }),
     startTap: PropTypes.bool.isRequired,
-    updateNeighbors: PropTypes.func,
-    updateTapClosingState: PropTypes.func
+    updateTapClosingState: PropTypes.func,
+    updateUnmeshedSources: PropTypes.func
```
rmars: since we now depend on this to show the tap icon in the TopTable, I think we should make this `isRequired`

web/app/js/components/TopModule.jsx
rmars: good catch!

web/app/js/components/TopModule.jsx
```diff
@@ -253,6 +261,14 @@ class TopModule extends React.Component {
     });
   }
 
+  updateNeighborsFromTapData = (source, sourceLabels) => {
+    // store this outside of state, as updating the state upon every websocket event received
+    // is very costly and causes the page to freeze up
+    let resourceType = _isNil(this.props.query.resource) ? "" : this.props.query.resource.split("/")[0];
+    this.unmeshedSources = processNeighborData(source, sourceLabels, this.unmeshedSources, resourceType);
+    if (this.props.updateUnmeshedSources) {this.props.updateUnmeshedSources(this.unmeshedSources);}
```
rmars: you don't need this check here, since the current default value of this func is `_noop` so it won't do anything. but also, if we make this prop required, we won't need to worry about checking it here!

web/app/js/components/util/TapUtils.jsx
```diff
@@ -354,7 +354,9 @@ export const tapLink = (d, resourceType, PrefixedLink) => {
   let namespace = d.sourceLabels.namespace;
   let resource = "";
 
-  if (_has(d.sourceLabels, resourceType)) {
+  if (!d.meshed) {
+    return null;
```
rmars: TIOLI we could return a greyed out, non-clickable Tap icon for resources that aren't tappable

web/app/js/components/TopModule.jsx
alpeb: When building  `<TopModule>` in the `/top` page I'm not passing any function because there's no octopus that needs an `unmeshedSources` object to be available.

web/app/js/components/TopModule.jsx
alpeb: ah that makes sense :+1:

web/app/js/components/util/TapUtils.jsx
alpeb: will do :sparkles:

alpeb:
@rmars please check my comment about the required function. I've pushed your other suggestions, including the grayed-out icon, which looks much better. Not sure though what are the project's standards regarding css...

web/app/js/components/TopModule.jsx
rmars: ah, right! I thought we needed this to keep track of the unmeshed upstreams for the table, but we're keeping track of them in TopModule

web/app/js/components/util/TapUtils.jsx
```diff
@@ -352,14 +352,19 @@ export const srcDstColumn = (d, resourceType, ResourceLink) => {
 
 export const tapLink = (d, resourceType, PrefixedLink) => {
   let namespace = d.sourceLabels.namespace;
-  let resource = "";
+  let resource;
```
rmars: could we set namespace/resource to "" here instead, so we don't have to deal with the subsequent cases?

```suggestion
  let namespace = d.sourceLabels.namespace || "";
  let resource = "";
```

That also kind of got me thinking, that instead of forcing these to be "" to get around the fact that namespace/resource are required by TapLink, that we could instead make those prop types be not required, but have a default value of "".

Another random thought I had was maybe we should instead of doing that, add a `disabled` prop that would control whether the tap icon was greyed out. Just trying to think of whatever is clearest, it's not that important though :)

alpeb: I like the idea of `disable`, the intention would be clearer :+1:

alpeb:
@rmars In my last push I addressed your comment, and I also fixed a missing scenario where the tap icon should be grayed-out (when `d.base.source.owner` is empty) :microscope:

web/app/js/components/TapLink.jsx
```diff
@@ -24,6 +24,7 @@ const TapLink = ({PrefixedLink, namespace, resource, toNamespace, toResource, pa
 };
 
 TapLink.propTypes = {
+  disabled: PropTypes.bool.isRequired,
```
rmars: TIOLI: I think `disabled` shouldn't be required; and that it should default to `false` since that'd probably be the most common use case; this doesn't matter much for this instance since `tapLink` is the only place we actually create a `TapLink`....

```
TapLink.defaultProps = {
  disabled: false
}
```

web/app/js/components/TopModule.jsx
```diff
@@ -244,7 +245,7 @@ class TopModule extends React.Component {
       if (isPod) {
         topResults[eventKey].meshed = !_has(this.unmeshedSources, "pod/" + d.base.source.pod);
       } else {
-        topResults[eventKey].meshed = !_has(this.unmeshedSources, d.base.source.owner);
+        topResults[eventKey].meshed = !_isEmpty(d.base.source.owner) && !_has(this.unmeshedSources, d.base.source.owner);
```
rmars: good catch!

web/app/js/components/TapLink.jsx
alpeb: no problemo :+1:

PR #2139, opened 2019-01-23
Add a --tap flag to the linkerd profile command

### LINKED ISSUE #2041: Create service profile from tap data
Add a `linkerd profile --tap (TARGET) (SERVICE)` command which generates a service profile based on tap data to the given service.

e.g. `linkerd profile --tap deploy/books books`

This command initiates a tap to the target resource and records requests for a (configurable) amount of time.  After that time has elapsed, it outputs a service profile based on the requests it saw.  For each distinct (method, path) pair, it creates a route.

Users may want to manually edit the output to consolidate routes when there are path params.

rmars:
Adds the ability to generate a service profile by running a tap for a configurable amount of time, and using the route results from the routes seen during the tap.

e.g. ` bin/go-run cli profile web --tap deploy/web -n emojivoto --tap-duration 2s`

Currently, this branch just takes all the routes it sees during that window. Which in the case of a high rps, very rest-ful app, (e.g. books) can be a lot. So the user has to consolidate the routes manually. It would definitely make sense to consolidate the paths next.

Fixes #2041


cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
...
+			Name:      fmt.Sprintf("%s.%s.svc.cluster.local", options.name, options.namespace),
+			Namespace: controlPlaneNamespace,
+		},
+		TypeMeta: meta_v1.TypeMeta{
+			APIVersion: "linkerd.io/v1alpha1",
+			Kind:       "ServiceProfile",
+		},
+	}
+
+	log.Debug("Running `linkerd tap %s --namespace %s`", options.tap, options.namespace)
```
siggy: `log.Debugf`

cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
 
 	cmd.PersistentFlags().BoolVar(&options.template, "template", options.template, "Output a service profile template")
 	cmd.PersistentFlags().StringVar(&options.openAPI, "open-api", options.openAPI, "Output a service profile based on the given OpenAPI spec file")
+	cmd.PersistentFlags().StringVar(&options.tap, "tap", options.tap, "Output a service profile based on tap data for the given target resource")
```
adleong: What do you think about a flag that limits the maximum number of routes to output?  By setting a default of something like 100, we can safeguard against out-of-control profiles for high traffic REST-y services.

cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
...
+		},
+	}
+
+	log.Debug("Running `linkerd tap %s --namespace %s`", options.tap, options.namespace)
+	client := cliPublicAPIClient()
+	requestParams := util.TapRequestParams{
+		Resource:  options.tap,
+		Namespace: options.namespace,
+	}
+	req, err := util.BuildTapByResourceRequest(requestParams)
```
siggy: should this filter on `proxy=in`? another way to say it, should it do this:
```bash
bin/linkerd tap deploy --to deploy/web -n emojivoto
```

when i generate a profile from `deploy/web`, i'm getting expected routes like `/api/list`, but i'm also getting `/emojivoto.v1.EmojiService/ListAll`, which i wouldn't expect on `web.emojivoto.svc.cluster.local`:
```bash
$ bin/go-run cli profile web --tap deploy/web -n emojivoto --tap-duration 2s
kind: ServiceProfile
metadata:
  creationTimestamp: null
  name: web.emojivoto.svc.cluster.local
  namespace: linkerd
spec:
  retryBudget: null
  routes:
  - condition:
      method: GET
      pathRegex: /api/list
    name: GET /api/list
  - condition:
      method: GET
      pathRegex: /api/vote
    name: GET /api/vote
  - condition:
      method: POST
      pathRegex: /emojivoto\.v1\.EmojiService/FindByShortcode
    name: POST /emojivoto.v1.EmojiService/FindByShortcode
  - condition:
      method: POST
      pathRegex: /emojivoto\.v1\.EmojiService/ListAll
    name: POST /emojivoto.v1.EmojiService/ListAll
  - condition:
      method: POST
      pathRegex: /emojivoto\.v1\.VotingService/VoteGirl
    name: POST /emojivoto.v1.VotingService/VoteGirl
  - condition:
      method: POST
      pathRegex: /emojivoto\.v1\.VotingService/VotePoop
    name: POST /emojivoto.v1.VotingService/VotePoop
```

cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
...
+
+	for _, path := range sortMapKeys(routesMap) {
+		routes = append(routes, routesMap[path])
+	}
+	return routes
+}
+
+func recordRoutesFromTap(tapClient pb.Api_TapByResourceClient, tapDuration time.Duration) map[string]*sp.RouteSpec {
+	routes := make(map[string]*sp.RouteSpec)
+	timerChannel := make(chan string, 1)
```
siggy: slightly more canonical:
```go
timerChannel := make(chan struct{}, 1)
```

cli/cmd/profile.go
rmars: yeah that's a good idea!

cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
...
+
+func getPathDataFromTap(event *pb.TapEvent) *sp.RouteSpec {
+	switch ev := event.GetHttp().GetEvent().(type) {
+	case *pb.TapEvent_Http_RequestInit_:
+		path := ev.RequestInit.GetPath()
+		if path == "/" {
+			return nil
+		}
+		return mkRouteSpec(
+			path,
```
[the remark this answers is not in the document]
rmars: wait I thought that `mkRouteSpec` sets the name to include the method? (in `Name:            fmt.Sprintf("%s %s", method, path),`)? or is this referring to something else?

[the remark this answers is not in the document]
adleong: oh, you're totally right.  ignore me.

cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
...
+	timerChannel := make(chan string, 1)
+	go func() {
+		time.Sleep(tapDuration)
+		timerChannel <- "done"
+	}()
+	done := ""
+
+	for {
+		log.Debug("Waiting for data...")
+		event, err := tapClient.Recv()
```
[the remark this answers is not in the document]
rmars: ahhh yeah, that's a good point. I think this restructuring would make sense!

cli/cmd/profile.go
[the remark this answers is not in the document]
rmars: reading https://gobyexample.com/timers makes me think that Sleep is the right choice here? I could be swayed though!

cli/cmd/profile.go
```diff
@@ -118,11 +147,121 @@ Example:
 
 	cmd.PersistentFlags().BoolVar(&options.template, "template", options.template, "Output a service profile template")
 	cmd.PersistentFlags().StringVar(&options.openAPI, "open-api", options.openAPI, "Output a service profile based on the given OpenAPI spec file")
+	cmd.PersistentFlags().StringVar(&options.tap, "tap", options.tap, "Output a service profile based on tap data for the given target resource")
+	cmd.PersistentFlags().StringVarP(&options.tapDuration, "tap-duration", "t", options.tapDuration, "Duration over which tap data is collected (for example: \"10s\", \"1m\", \"10m\")")
```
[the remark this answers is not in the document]
rmars: this is much better! ty!

cli/cmd/profile.go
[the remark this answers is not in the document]
adleong: by my reading of it, you can just do 
```
timer := time.NewTimer(tapDuration)
```

and then in your select:
```
case  <-timer.C:
```

which lets you get rid of the manual goroutine with the sleep

cli/cmd/profile.go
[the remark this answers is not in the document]
rmars: great catch!

rmars: ahhhh. I think that makes sense.

rmars: we'd run into the problem of missing traffic between different resource types though.
what if we did `linkerd tap namespace --to deploy/web -n emojivoto` instead?

thoughts? cc @adleong

adleong: I think we want to look at inbound traffic for this, not outbound.  Is it possible to still use `linkerd tap deploy/web` but post-filter the results to only consider proxy=in?

cli/cmd/profile.go
```diff
@@ -94,6 +111,16 @@ Example:
 If the --open-api flag is specified, it reads the given OpenAPI
 specification file and outputs a corresponding service profile.
 
+Example:
+	linkerd profile --tap deploy/books --tap-duration 10s books > book-svc-profile.yaml
+	# (edit book-svc-profile.yaml manualy)
```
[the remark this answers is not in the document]
dadjeibaah: The example text for `--tap` needs a bit of editing to align with the rest of the text.
```diff
diff --git a/cli/cmd/profile.go b/cli/cmd/profile.go
index 9090c951..4f599b05 100644
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -112,9 +112,9 @@ If the --open-api flag is specified, it reads the given OpenAPI
 specification file and outputs a corresponding service profile.

 Example:
-       linkerd profile --tap deploy/books --tap-duration 10s books > book-svc-profile.yaml
-       # (edit book-svc-profile.yaml manualy)
-       kubectl apply -f book-svc-profile.yaml
+  linkerd profile --tap deploy/books --tap-duration 10s books > book-svc-profile.yaml
+  # (edit book-svc-profile.yaml manualy)
+  kubectl apply -f book-svc-profile.yaml

```

cli/cmd/profile.go
```diff
@@ -118,11 +147,149 @@ Example:
...
+	w.Write(output)
+	return nil
+}
+
+func routeSpecFromTap(tapClient pb.Api_TapByResourceClient, tapDuration time.Duration, routeLimit int) []*sp.RouteSpec {
+	routes := make([]*sp.RouteSpec, 0)
+	routesMap := make(map[string]*sp.RouteSpec)
+
+	tapEventChannel := make(chan *pb.TapEvent, 10)
+	timerChannel := make(chan struct{}, 1)
```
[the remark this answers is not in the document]
dadjeibaah: I was just about to comment something similar to siggy's suggestion. 😄. in addition, I think you can also simplify the routeLimit check so that it wraps the `getPathDataFromTap(...)` method. So something like this:
```diff
diff --git a/cli/cmd/profile.go b/cli/cmd/profile.go
index 9090c951..26c90857 100644
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
+               if len(routesMap) <= routeLimit {
                        routeSpec := getPathDataFromTap(event)
-
-                       if len(routesMap) > routeLimit {
-                               stopTap = true
-                               break
-                       }
-
                        if routeSpec != nil {
                                routesMap[routeSpec.Name] = routeSpec
                        }
```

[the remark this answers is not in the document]
rmars: changed the structure of this per siggy's suggestion; I think the check/break is a little simpler now!

cli/cmd/profile.go
```diff
@@ -109,6 +136,8 @@ Example:
 				return profiles.RenderProfileTemplate(options.namespace, options.name, controlPlaneNamespace, os.Stdout)
 			} else if options.openAPI != "" {
 				return renderOpenAPI(options, os.Stdout)
+			} else if options.tap != "" {
+				return renderTapOutputProfile(options, controlPlaneNamespace, os.Stdout)
 			}
 
 			// we should never get here
```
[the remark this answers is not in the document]
dadjeibaah: Yea, I see your point. I was working off of the assumption that there would be no way to reach `return errors.New("Unexpected error")`. If we tried to do that we would need to not provide any options for the profile command, however, `options.validate()` also runs similar checks so we would return an error before we run into the subsequent checks in question.

On the other hand, I see that doing return `errors.New("Unexpected error")` is beneficial because it represents an undesired state for the command. 

That's a long way of saying I don't have a preference 😂I could go either way now that I think about it.

pkg/profiles/tap.go
```diff
@@ -0,0 +1,142 @@
...
+	routes := make([]*sp.RouteSpec, 0)
+	routesMap := make(map[string]*sp.RouteSpec)
+
+recvLoop:
+	for {
+		select {
+		case <-ctx.Done():
+			break recvLoop
+		default:
+			log.Debug("Waiting for data...")
```
dadjeibaah: What do you think about removing this log line and adding a debug log line that displays the routeSpec this function was able to create after receiving the tap event. This might provide useful information in the case where we need to observe what routeSpecs are being created based on the data we are getting from tap.

pkg/profiles/tap.go
```diff
@@ -0,0 +1,142 @@
...
+			event, err := tapClient.Recv()
+
+			if err == io.EOF {
+				break recvLoop
+			}
+			if err != nil {
+				fmt.Fprintln(os.Stderr, err)
```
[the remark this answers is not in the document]
rmars: I think the reason this is here is so that it prints out the error even if you don't have a log level set? if I do this instead, it fails silently without printing the error (the other cli/ commands use this pattern as well)

pkg/profiles/tap.go
rmars: I think having the routespec data could be useful! I think this is here so we know we've started a tap, but we are waiting to receive (could be useful if we're waiting a while without any results). added another log line for logging the routespec.

pkg/profiles/tap.go
```diff
@@ -0,0 +1,140 @@
...
+			Name:      fmt.Sprintf("%s.%s.svc.cluster.local", name, namespace),
+			Namespace: controlPlaneNamespace,
+		},
+		TypeMeta: ServiceProfileMeta,
+	}
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(tapDuration))
+	defer cancel()
+
+	tapClient, err := client.TapByResource(ctx, tapReq)
```
siggy: apologies, my previous recommendation caused a race condition where the tapClient could error out on `tapClient.Recv()` with a scary error message. i think a better approach is to strictly rely on tapClient.Recv() hitting the deadline, and hiding the error if we know it was caused the by deadline. let me know what you think:
```diff
diff --git a/pkg/profiles/tap.go b/pkg/profiles/tap.go
index 01bde4d0..779f12be 100644
--- a/pkg/profiles/tap.go
+++ b/pkg/profiles/tap.go
@@ -6,6 +6,7 @@ import (
        "io"
        "os"
        "sort"
+       "strings"
        "time"

        "github.com/ghodss/yaml"
@@ -62,43 +63,39 @@ func tapToServiceProfile(client pb.ApiClient, tapReq *pb.TapByResourceRequest, n
                return profile, err
        }

-       routes := routeSpecFromTap(ctx, tapClient, routeLimit)
+       routes := routeSpecFromTap(tapClient, routeLimit)

        profile.Spec.Routes = routes

        return profile, nil
 }

-func routeSpecFromTap(ctx context.Context, tapClient pb.Api_TapByResourceClient, routeLimit int) []*sp.RouteSpec {
+func routeSpecFromTap(tapClient pb.Api_TapByResourceClient, routeLimit int) []*sp.RouteSpec {
        routes := make([]*sp.RouteSpec, 0)
        routesMap := make(map[string]*sp.RouteSpec)

-recvLoop:
        for {
-               select {
-               case <-ctx.Done():
-                       break recvLoop
-               default:
-                       log.Debug("Waiting for data...")
-                       event, err := tapClient.Recv()
-
-                       if err == io.EOF {
-                               break recvLoop
-                       }
-
-                       if err != nil {
+               log.Debug("Waiting for data...")
+               event, err := tapClient.Recv()
+
+               if err != nil {
+                       // expected errors when hitting the tapDuration deadline
+                       if err != io.EOF &&
+                               !strings.HasSuffix(err.Error(), context.DeadlineExceeded.Error()) &&
+                               !strings.HasSuffix(err.Error(), "http2: response body closed") {
                                fmt.Fprintln(os.Stderr, err)
-                               break recvLoop
                        }

-                       routeSpec := getPathDataFromTap(event)
-                       log.Debugf("Created route spec: %v", routeSpec)
+                       break
+               }
+
+               routeSpec := getPathDataFromTap(event)
+               log.Debugf("Created route spec: %v", routeSpec)

-                       if routeSpec != nil {
-                               routesMap[routeSpec.Name] = routeSpec
-                               if len(routesMap) > routeLimit {
-                                       break recvLoop
-                               }
+               if routeSpec != nil {
+                       routesMap[routeSpec.Name] = routeSpec
+                       if len(routesMap) > routeLimit {
+                               break
                        }
                }
        }
```

rmars: oh interesting. yeah I see what you mean. I guess it does simplify the looping if we do is this way!__

cli/cmd/profile.go
```diff
@@ -82,6 +92,10 @@ Examples:
   Edit the template and then apply it with kubectl to add a service profile to
   a service:
 
+  If the --tap flag is specified, it runs linkerd tap target for --tap-duration seconds,
+  and creates a profile for the SERVICE based on the requests seen in that window
+
+Example:
```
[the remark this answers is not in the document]
adleong: I think this text is in the wrong place: it got between the description of the `--template` flag and the example of using `--template`.  I think this should move down to be right above the `--tap` example.  And, to be consistent with the rest of this section, I don't think it needs the `Example:` header.

pkg/profiles/tap.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	w.Write(output)
+	return nil
+}
+
+func tapToServiceProfile(client pb.ApiClient, tapReq *pb.TapByResourceRequest, namespace, name, controlPlaneNamespace string, tapDuration time.Duration, routeLimit int) (sp.ServiceProfile, error) {
+	profile := sp.ServiceProfile{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      fmt.Sprintf("%s.%s.svc.cluster.local", name, namespace),
+			Namespace: controlPlaneNamespace,
```
adleong: FYI this will semantically conflict with https://github.com/linkerd/linkerd2/pull/2200.  Depending on which change lands first, someone will need to resolve this to point at the service's namespace.

pkg/profiles/tap.go
```diff
@@ -0,0 +1,136 @@
...
+	routesMap := make(map[string]*sp.RouteSpec)
+
+	for {
+		log.Debug("Waiting for data...")
+		event, err := tapClient.Recv()
+
+		if err != nil {
+			// expected errors when hitting the tapDuration deadline
+			if err != io.EOF &&
+				!strings.HasSuffix(err.Error(), context.DeadlineExceeded.Error()) &&
```
adleong: Is there a way to check the type or value of the error object instead of comparing the string message?

pkg/profiles/tap.go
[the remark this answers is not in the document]
adleong: Out of scope for this issue, but maybe we should investigate this separately.  It seems weird that there isn't a default log level??

pkg/profiles/tap.go
```diff
@@ -0,0 +1,142 @@
...
+				break recvLoop
+			}
+
+			routeSpec := getPathDataFromTap(event)
+
+			if routeSpec != nil {
+				routesMap[routeSpec.Name] = routeSpec
+			}
+
+			if len(routesMap) > routeLimit {
```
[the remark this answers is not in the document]
adleong: Does this actually allow `routeLimit + 1` routes?

pkg/profiles/tap.go
siggy: This unfortunate string check was my suggestion. The error object that contains this string is private:
https://github.com/golang/net/blob/351d144fa1fc0bd934e2408202be0c29f25e35a0/http2/transport.go#L2008

adleong: :sob:

pkg/profiles/tap.go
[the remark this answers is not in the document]
siggy: logrus's default is `InfoLevel`. in the cli, we expose a `--verbose` flag, when disabled sets `PanicLevel`, when enabled, sets `DebugLevel`.

i agree with @rmars implementation here, that this error should always print regardless of log level, and should not print like a log message, but rather a cli output message.

adleong:
If I specify a duration that's short enough to not capture any events, there's a weird error message:

```
$ go run cli/main.go profile --tap deploy/webapp webapp --tap-duration 1ms
Error: Post https://localhost:6443/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/TapByResource: context deadline exceeded
```

pkg/profiles/tap.go
rmars: 😭

cli/cmd/profile.go
```diff
@@ -94,6 +104,19 @@ Examples:
...
+  # (edit book-svc-profile.yaml manually)
+  kubectl apply -f book-svc-profile.yaml
+
+  The command will run linkerd tap deploy/books for tap-duration seconds, and then create
+  a service profile for the books service with routes prepopulated from the tap data.
+  For high RPS, high-route-cardinality services, use tap-route-limit to limit the number of
+  routes in the output profile.
+
+  linkerd profile -n emojivoto --open-api web-svc.swagger web-svc | kubectl apply -f -,
   linkerd profile -n emojivoto --proto Voting.proto vote-svc | kubectl apply -f -`,
```
siggy: move this line up to right below `If the --proto flag...`

rmars: good catch! okay I /think/ the examples section is ok now

pkg/profiles/tap.go
```diff
@@ -60,6 +61,10 @@ func tapToServiceProfile(client pb.ApiClient, tapReq *pb.TapByResourceRequest, n
 
 	tapClient, err := client.TapByResource(ctx, tapReq)
 	if err != nil {
+		if strings.HasSuffix(err.Error(), context.DeadlineExceeded.Error()) {
```
adleong: is it possible to do a object equality comparison instead of a string comparison for this one?

rmars: I think you have to do the string suffix comparison because the error that gets returned here (err) is `Post https://localhost:6443/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/TapByResource: context deadline exceeded` whereas the value of `context.DeadlineExceeded.Error()` is only `context deadline exceeded`

PR #2140, opened 2019-01-24
Fix issue where linkerd check would panic if there was a pod with a replicationcontroller ID in the control plane namespace 

### LINKED ISSUE #2084: Panic when running linkerd check with the linkerd-namespace flag set to default
This is kind of weird bug and only came across it by accident. When running Linkerd check with a namespace that doesn't have linkerd installed in it I get a panic.
```bash
➜  linkerd2 git:(master) ✗ bin/go-run cli check -l default
kubernetes-version: is running the minimum Kubernetes API version..........[ok]
linkerd-existence: control plane namespace exists..........................[ok]
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/linkerd/linkerd2/pkg/healthcheck.getPodStatuses(0xc4201a6000, 0xe, 0x13, 0x73646f702f746c)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/pkg/healthcheck/healthcheck.go:866 +0x4d6
github.com/linkerd/linkerd2/pkg/healthcheck.checkControllerRunning(0xc4201a6000, 0xe, 0x13, 0x7, 0xc4201a6000)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/pkg/healthcheck/healthcheck.go:905 +0x51
github.com/linkerd/linkerd2/pkg/healthcheck.(*HealthChecker).addLinkerdControlPlaneExistenceChecks.func2(0xc4203d7120, 0x38)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/pkg/healthcheck/healthcheck.go:433 +0x1b4
github.com/linkerd/linkerd2/pkg/healthcheck.(*HealthChecker).runCheck(0xc4205e2e80, 0xc4203d70e0, 0xc420245ac8, 0xc420321601)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/pkg/healthcheck/healthcheck.go:667 +0xa3
github.com/linkerd/linkerd2/pkg/healthcheck.(*HealthChecker).RunChecks(0xc4205e2e80, 0xc420245ac8, 0x0)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/pkg/healthcheck/healthcheck.go:640 +0x1ab
github.com/linkerd/linkerd2/cli/cmd.runChecks(0x28064e0, 0xc4200ce008, 0xc4205e2e80, 0xc4205e2d80)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/cli/cmd/check.go:179 +0x94
github.com/linkerd/linkerd2/cli/cmd.configureAndRunChecks(0xc4203bae80, 0xc420257c28, 0xc420257be0)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/cli/cmd/check.go:130 +0x76f
github.com/linkerd/linkerd2/cli/cmd.newCmdCheck.func1(0xc4205e8480, 0xc4204c3740, 0x0, 0x2, 0x0, 0x0)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/cli/cmd/check.go:62 +0x42
github.com/linkerd/linkerd2/vendor/github.com/spf13/cobra.(*Command).execute(0xc4205e8480, 0xc4204c36a0, 0x2, 0x2, 0xc4205e8480, 0xc4204c36a0)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/vendor/github.com/spf13/cobra/command.go:698 +0x4db
github.com/linkerd/linkerd2/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x31212c0, 0x0, 0x102e540, 0x23b85ed)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/vendor/github.com/spf13/cobra/command.go:783 +0x415
github.com/linkerd/linkerd2/vendor/github.com/spf13/cobra.(*Command).Execute(0x31212c0, 0x23c0a98, 0xc4200ae058)
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/vendor/github.com/spf13/cobra/command.go:736 +0x39
main.main()
	/Users/dennis/workspace/go/src/github.com/linkerd/linkerd2/cli/main.go:10 +0x4b
```

Pods in that namespace where from the booksapp.yaml
```bash
authors-7dbc46647b-h7hrb   2/2       Running   1          28d
books-558b45bbb-xcx8g      2/2       Running   0          28d
hello-6c666bd4f9-c6w96     3/3       Running   0          6h
hello-6c666bd4f9-s794c     3/3       Running   0          6h
hello-6c666bd4f9-tlcsk     3/3       Running   0          6h
hello-stpb9                2/2       Running   0          2m
...
```

The code that experienced the panic was here https://github.com/linkerd/linkerd2/blob/master/pkg/healthcheck/healthcheck.go#L866.

It looks like the `getPodStatuses` method assumes that all pod names will be of type `linkerd-xyz-abc-def` based off of the result of strings.Split. This method is only used when dealing  getting pod statuses from control plane pods. We probably could add a check that ensures that whatever pod we are working with in this method is indeed a control plane pod.

dadjeibaah:
When running Linkerd check with a control plane namespace that may contain an additional pod with a replication controller ID for pod names instead of a replicaSet ID, the check command panics because of an "index out of bounds" error.

This PR adds a check to make sure that, when parsing pod names during the `checkControllerRunning` healthcheck, the check ensures that the pod name can be split into three substrings on '-'. This prevents the panic from occurring when the check encounters a replication controller ID pod name.

Fixes #2084 

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -802,7 +802,14 @@ func getPodStatuses(pods []v1.Pod) map[string][]v1.ContainerStatus {
 	for _, pod := range pods {
 		if pod.Status.Phase == v1.PodRunning {
 			parts := strings.Split(pod.Name, "-")
-			name := strings.Join(parts[1:len(parts)-2], "-")
+			var name string
+			// All control plane pods  should have a name that results in 3 substrings string.Split on '-'
+			if len(parts) == 3 {
+				name = strings.Join(parts[1:len(parts)-2], "-")
+			} else {
+				name = strings.Join(parts, "-")
+			}
+
 			if _, found := statuses[name]; !found {
 				statuses[name] = make([]v1.ContainerStatus, 0)
 			}


pkg/healthcheck/healthcheck.go
```diff
@@ -802,7 +802,14 @@ func getPodStatuses(pods []v1.Pod) map[string][]v1.ContainerStatus {
 	for _, pod := range pods {
 		if pod.Status.Phase == v1.PodRunning {
 			parts := strings.Split(pod.Name, "-")
-			name := strings.Join(parts[1:len(parts)-2], "-")
+			var name string
+			// All control plane pods  should have a name that results in 3 substrings string.Split on '-'
```
siggy: controller pods should split into 4 parts because we recently added the `linkerd-` prefix (`linkerd-controller-b8c4c48c8-q85d2`)

in addition to checking on the length, i think it's safer to just ignore pods we don't recognize, consider something like:
```go
for _, pod := range pods {
	if pod.Status.Phase == v1.PodRunning && strings.HasPrefix(pod.Name, "linkerd-") {
		parts := strings.Split(pod.Name, "-")
		// All control plane pods should have a name that results in 4 substrings string.Split on '-'
		if len(parts) == 4 {
			name := parts[1]
			if _, found := statuses[name]; !found {
				statuses[name] = make([]v1.ContainerStatus, 0)
			}
			statuses[name] = append(statuses[name], pod.Status.ContainerStatuses...)
		}
	}
}
```

pkg/healthcheck/healthcheck.go
```diff
@@ -802,7 +802,14 @@ func getPodStatuses(pods []v1.Pod) map[string][]v1.ContainerStatus {
...
+			if len(parts) >= 3 {
+				name = strings.Join(parts[1:len(parts)-2], "-")
+			} else {
+				name = strings.Join(parts, "-")
```
klingerf: If we know that this isn't a control plane pod, we should just skip it altogether. I think this works:

```
			parts := strings.Split(pod.Name, "-")
			if len(parts) < 3 {
				// If this pod doesn't have a replicaset ID in it's name, then it's
				// not owned by the control plane and we can skip it
				continue
			}
			name := strings.Join(parts[1:len(parts)-2], "-")
```

dadjeibaah:
Great suggestions @siggy @klingerf I'll make those changes in the next commit

pkg/healthcheck/healthcheck.go
klingerf: @siggy Jinx! And not to bikeshed this to death, but some of our control plane pods will actually have 5 parts in their name, e.g. "linkerd-proxy-injector-5ffc675b8b-8zfrf".

There's a related issue as well, which is that if you install something else into your linkerd namespace and it's not healthy, that will cause this check to fail. So given that we're now using the "linkerd-" prefix everywhere, maybe we can fix both issues with something like:

```
	for _, pod := range pods {
		if pod.Status.Phase == v1.PodRunning {
			if strings.HasPrefix(pod.Name, "linkerd-") {
				parts := strings.Split(pod.Name, "-")
				// All control plane pods should have a name that results in at least 4
				// substrings when string.Split on '-'
				if len(parts) >= 4 {
					name := strings.Join(parts[1:len(parts)-2], "-")
					if _, found := statuses[name]; !found {
						statuses[name] = make([]v1.ContainerStatus, 0)
					}
					statuses[name] = append(statuses[name], pod.Status.ContainerStatuses...)
				}
			}
		}
	}
```

siggy: @klingerf that all makes sense! i will continue this bikeshed to say we should stick with the single line `if` i had proposed, to decrease indent:
```
if pod.Status.Phase == v1.PodRunning && strings.HasPrefix(pod.Name, "linkerd-") {
```

dadjeibaah: I like the Prefix check. It felt kinda strange checking for a pod that was not part of the components of the control plane.

klingerf: omg @siggy I didn't even see that your solution _already included_ the prefix check. nevermind me -- carry on.

PR #2141, opened 2019-01-24
Add a less scary message to the user when retries are still in progress

### LINKED ISSUE #2082: improve `linkerd install` retry messages
In a Docker for Mac install of linkerd edge-12.1.1, I see this output from running `linkerd check` shortly after installing:

<pre>
linkerd-existence: can query the control plane API.........................[retry] -- Unexpected API response: 503 Service Unavailable
linkerd-existence: can query the control plane API.........................[retry] -- Unexpected API response: 503 Service Unavailable
linkerd-existence: can query the control plane API.........................[ok]
</pre>

"Unexpected API response" seems overly scary, not to mention inaccurate, for a 100% expected situation.

rmars:
When a failing check is being retried, we show the current err to the user. This can sometimes be unnecessarily alarming, as in the case of the control plane starting up.

If a failing check is in the process of being retried, wait to show the final error message until the retries have completed.

Fixes #2082

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -15,6 +15,7 @@ import (
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/profiles"
 	"github.com/linkerd/linkerd2/pkg/version"
+	log "github.com/sirupsen/logrus"
 	authorizationapi "k8s.io/api/authorization/v1beta1"
 	"k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -608,6 +609,9 @@ func (hc *HealthChecker) runCheck(categoryID CategoryID, c *checker, observer ch
 
 		if err != nil && time.Now().Before(c.retryDeadline) {
 			checkResult.Retry = true
+			checkResult.Err = errors.New("waiting for check to complete")
+			log.Debugf("Retrying on error: %s", err.Error())
```
siggy: you can do: `log.Debugf("Retrying on error: %s", err)` (we're probably not consistent around the codebase with this)

PR #2142, opened 2019-01-24
Check RoleBindings for specified single namespace only

### LINKED ISSUE #2117: linkerd check --single-namespace Missing permissions to create Role/RoleBinding
### What is the issue?

`linkerd check` reports missing permissions, though `linkerd install` seems to work ok

### How can it be reproduced?

Latest release edge-19.1.2, use single-namespace, not full cluster permissions

#### `linkerd check` output

`linkerd check --pre -l $EKSNAMESPACE --single-namespace`
 
can create Roles -> Missing permissions to create Role
can create RoleBindings -> Missing permissions to create RoleBinding
everything else passes
 
`linkerd install -l $EKSNAMESPACE --single-namespace | kubectl -n $EKSNAMESPACE apply -f -`

### Environment

- Kubernetes Version: 1.10
- Cluster Environment: EKS
- Host OS:
- Linkerd version: edge-19.1.2 

### Possible solution

These lines dont seems to consider the namespace with hc.ControlPlaneNamespace?
 
https://github.com/linkerd/linkerd2/blob/master/pkg/healthcheck/healthcheck.go#L298
https://github.com/linkerd/linkerd2/blob/master/pkg/healthcheck/healthcheck.go#L305

### Additional context

rmars:
Previously, we were doing the creation checks for both Roles/RoleBindings and ClusterRoles/ClusterRoleBindings for all namespaces, but in `--single-namespace` mode, we only need to check that these can be created in the control plane namespace.

Fixes #2117

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -295,14 +295,14 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "can create Roles",
 					hintURL:     "https://linkerd.io/2/getting-started/#step-0-setup",
 					check: func() error {
-						return hc.checkCanCreate("", "rbac.authorization.k8s.io", "v1beta1", "Role")
+						return hc.checkCanCreate(hc.ControlPlaneNamespace, "rbac.authorization.k8s.io", "v1beta1", "Role")
 					},
 				},
 				{
 					description: "can create RoleBindings",
 					hintURL:     "https://linkerd.io/2/getting-started/#step-0-setup",
 					check: func() error {
-						return hc.checkCanCreate("", "rbac.authorization.k8s.io", "v1beta1", "RoleBinding")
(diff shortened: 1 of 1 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -295,14 +295,14 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "can create Roles",
 					hintURL:     "https://linkerd.io/2/getting-started/#step-0-setup",
 					check: func() error {
-						return hc.checkCanCreate("", "rbac.authorization.k8s.io", "v1beta1", "Role")
+						return hc.checkCanCreate(hc.ControlPlaneNamespace, "rbac.authorization.k8s.io", "v1beta1", "Role")
```
siggy: gaah this was totally me in https://github.com/linkerd/linkerd2/pull/2078/files#diff-d4056ff163bcf2aeacefb2a34164563cR327 ... thanks for fixing !

PR #2149, opened 2019-01-25
Add support for timeouts in service profiles

### LINKED ISSUE #2042: Allow setting timeout in service profile route
Add the ability to set a timeout in the route section of a service profile.

```
routes:
- name: launch
  condition:
    pathRegex: '/launch'
  timeout: 200ms
```

adleong:
Fixes #2042 

Adds a new field to service profile routes called `timeout`.  Any requests to that route which take longer than the given timeout will be aborted and a 504 response will be returned instead.  If the timeout field is not specified, a default timeout of 1 second is used.


adleong:
How I tested:

* Install Linkerd and booksapp
* Add service profiles for booksapp from swagger definitions
* Notice that calls on a particular route from webapp to books have 100% success rate: 
```
$ bin/linkerd routes deploy/webapp --to svc/books | grep 'GET /books.json'
```
* Edit the books service profile and add an aggressive timeout:
```
...
  routes:
  - condition:
      method: GET
      pathRegex: /books\.json
    name: GET /books.json
    timeout: 10ms # pick a timeout that will make some requests fail
...
```
* Check the stats again and see requests are hitting the timeout and failing:
```
$ bin/linkerd routes deploy/webapp --to svc/books | grep 'GET /books.json'
GET /books.json             books    73.08%   0.9rps           7ms          10ms          10ms
```
(Note that the higher quantiles might show values larger than the timeout if your timeout is near a histogram bucket boundry.  because math)
* See the latency of all responses.  They should all be under (or just barely over) the timeout:
```
bin/linkerd tap deploy/webapp --to svc/books --path /books.json --method GET | grep latency
```

pkg/profiles/profiles.go
```diff
@@ -80,11 +84,16 @@ func ToRoute(route *sp.RouteSpec) (*pb.Route, error) {
 		}
 		rcs = append(rcs, pbRc)
 	}
+	timeout, err := time.ParseDuration(route.Timeout)
+	if err != nil {
+		timeout = 1 * time.Second
```
siggy: maybe encode this somewhere as a default constant?

siggy: consider:
```go
log.Errorf("failed to parse duration for route %+v: %s", route, err)
```

pkg/profiles/template.go
```diff
@@ -73,6 +73,11 @@ spec:
       # successes or failures.
       isFailure: true
 
+    # A route can definte a request timeout.  Any requests to this route that
+    # exceed the timeout will be canceled.  If unspecified, the default timeout
+    # is '1s' (one second).
+    # timeout: 250ms
```
siggy: i really like the way these templates are documented. a related ergonomics question: should the outputs of `--proto` and `--open-api` flags produce similar output and/or default values?

adleong: I'm not sure.  The more extensive comments/documentation we have in each of these outputs, the harder it will be to keep them all up to date and in sync.  So far I've been treating --template as the authoritative "documentation by example" and the other outputs as basically generated code.

pkg/profiles/profiles.go
```diff
@@ -89,11 +97,17 @@ func ToRoute(route *sp.RouteSpec) (*pb.Route, error) {
 		}
 		rcs = append(rcs, pbRc)
 	}
+	timeout, err := time.ParseDuration(route.Timeout)
+	if err != nil {
+		log.Errorf("failed to parse duration for route %s: %s", route.Name, err.Error())
```
siggy: you can just use `err` rather than `err.Error()`.

pkg/profiles/profiles.go
```diff
@@ -89,11 +97,17 @@ func ToRoute(route *sp.RouteSpec) (*pb.Route, error) {
 		}
 		rcs = append(rcs, pbRc)
 	}
+	timeout, err := time.ParseDuration(route.Timeout)
```
klingerf: Can we also update the "no invalid service profiles" check to validate that the `Timeout` field contains a valid duration?

pkg/profiles/profiles.go
```diff
@@ -43,8 +44,18 @@ var (
 		Routes:      []*pb.Route{},
 		RetryBudget: &DefaultRetryBudget,
 	}
+	// DefaultRouteTimeout is the default timeout for routes that do not specify
+	// one.
+	DefaultRouteTimeout = 1 * time.Second
```
klingerf: Before we removed the bind timeout in #2017, we effectively had a default timeout of 10 seconds. I'm a bit worried that setting a default timeout of 1 second will be too aggressive for some installations, especially those that were previously reliant on the 10 second time. How would you feel about setting this default to 10 seconds to match?

PR #2159, opened 2019-01-25
Add hint URLs for all checks

siggy:
linkerd/website#105 introduced a FAQ page, providing resolutions for all
`linkerd check` failures.

Update each check to reference its corresponding section in the FAQ.


pkg/healthcheck/healthcheck.go
```diff
@@ -469,16 +492,21 @@ func (hc *HealthChecker) allCategories() []category {
...
 						return hc.latestVersions.Match(hc.serverVersion)
 					},
 				},
 				{
 					description: "control plane and cli versions match",
+					hintURL:     "https://linkerd.io/2/faq/#l5d-version-control",
 					warning:     true,
 					check: func() error {
-						return version.Match(hc.serverVersion, version.Version)
+						if hc.serverVersion != version.Version {
```
dadjeibaah: What is the reason for switching to a string comparison over the `version.Match`?

siggy: Good question! `version.Match` is specifically written to return an informative error when doing a check against a version channel retrieved online:
https://github.com/linkerd/linkerd2/blob/c9ac77cd7c0f8b5211b93375fd436436f4ecec39/pkg/version/version.go#L35-L62

...and, in answering this question, I realized that this implied `version.Match` doesn't make sense outside the context of `Channels.Match`, so I've just made `version.Match` private, thanks!

hawkw:
This is awesome! :D

PR #2160, opened 2019-01-26
Update client-go to 1.13.1

### LINKED ISSUE #2145: Linkerd CLI can't connect to clusters that use self-signed certs
## Bug Report

Users are [reporting](https://discourse.linkerd.io/t/using-self-signed-k8s-api-certificate/801) that the `linkerd` CLI can't connect to clusters that are setup with self-signed certs, but the `kubectl` CLI is able to successfully connect. This indicates that `linkerd` is not handling credentials in the same way the `kubectl` is handling credentials, but it should be.

### What is the issue?

On a cluster with self-signed certs, API calls return "certificate signed by unknown authority".

### How can it be reproduced?

Setup a cluster with self-signed certs (I haven't tried to replicate this yet).

### `linkerd check` output

```text
kubernetes-api
--------------
✔ can initialize the client
✘ can query the Kubernetes API
    Get REDACTED/version: x509: certificate signed by unknown authority
```

grampelberg:
There's a new cluster config option (`certificate-authority-data`) supported by the latest version of client-go. This updates to the latest and fixes the problem.

Fixes #2145

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -67,11 +67,3 @@ required = [
 [[constraint]]
   name = "github.com/wercker/stern"
   revision = "b04b5491222d9743529cb859a20d34ce2bb763af" # pin to 1.6.0 until https://github.com/wercker/stern/issues/96 is resolved
-#
-# k8s.io/client-go dependency fixes
-# taken from https://github.com/kubernetes/client-go/blob/kubernetes-1.11.1/Godeps/Godeps.json
-#
-
-[[override]]
-  name = "github.com/Azure/go-autorest"
-  version = "v11.3.2"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -22,15 +22,15 @@ required = [
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.11.1"
+  version = "kubernetes-1.13.1"
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.11.1"
+  version = "kubernetes-1.13.1"
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  version = "kubernetes-1.11.1"
+  version = "kubernetes-1.13.1"
 
 [[constraint]]
   name = "k8s.io/helm"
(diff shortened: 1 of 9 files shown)


Gopkg.toml
```diff
@@ -72,14 +72,6 @@ required = [
 # taken from https://github.com/kubernetes/client-go/blob/kubernetes-1.11.1/Godeps/Godeps.json
```
siggy: you can remove this comment as it referred to the deps below it. consider removing the `go-autorest` reference altogether from this file, since we don't directly depend on it.

grampelberg: I thought I needed `go-autorest` in there (definitely had to do something with it to get everything to compile). Now that it is in the lockfile, doesn't appear to be required so I'll chalk it up to issues on my end.

grampelberg:
Well, at least it worked this time. Now to find where in the world that cannot create log error is coming from and why it fixes itself.

klingerf:
@siggy's right about that klog error. With klog, we should now be able to send those logs to `/dev/null`, which wasn't possible with glog. But that's likely a more involved change. How would you feel about using `kubernetes-1.12.5` for this branch, which predates the klog change? And then we could switch to klog in a separate PR?

pkg/flags/flags.go
```diff
@@ -7,14 +7,16 @@ import (
...
-// overrides the default flag for glog logging, which we can't disable. This
+// ConfigureAndParse adds flags that are common to all go processes. This
 // func calls flag.Parse(), so it should be called after all other flags have
 // been configured.
 func ConfigureAndParse() {
-	// override glog's default configuration
+	// override klog's default configuration and log to stderr instead of a log
+	// file.
+	klog.InitFlags(nil)
 	flag.Set("logtostderr", "true")
```
siggy: is `flag.Set("logtostderr", "true")` still necessary if we're no longer using glog?

pkg/flags/flags.go
```diff
@@ -7,14 +7,16 @@ import (
...
 
-// ConfigureAndParse adds flags that are common to all go processes, and
-// overrides the default flag for glog logging, which we can't disable. This
+// ConfigureAndParse adds flags that are common to all go processes. This
 // func calls flag.Parse(), so it should be called after all other flags have
 // been configured.
 func ConfigureAndParse() {
-	// override glog's default configuration
+	// override klog's default configuration and log to stderr instead of a log
+	// file.
```
siggy: the tiniest of nits: remove period or capitalize first word.

pkg/flags/flags.go
```diff
@@ -7,14 +7,15 @@ import (
...
 )
 
-// ConfigureAndParse adds flags that are common to all go processes, and
-// overrides the default flag for glog logging, which we can't disable. This
+// ConfigureAndParse adds flags that are common to all go processes. This
 // func calls flag.Parse(), so it should be called after all other flags have
 // been configured.
 func ConfigureAndParse() {
-	// override glog's default configuration
+	// Override klog's default configuration and log to stderr instead of a file
```
siggy: [something snarky about sentence definitions]

grampelberg:
Okay, working for me, comments aren't sentences and log lines appear to be identical to before.

Follow up work:

- #2168 
- #2169

PR #2163, opened 2019-01-28
Use `ca.NewCA()` for generating certs and keys for the proxy injector

### LINKED ISSUE #2095: Simplify certificate distribution for webhooks
## Background
Proxy auto-inject is done by a mutating webhook and it is likely that Linkerd will add more webhooks in the future (for example, a validating webhook for service profile validation).  The configuration and initialization of these webhooks is complicated because of the way that TLS credentials are distributed.

When the webhook is created, the webhook container posts a `MutatingWebhookConfiguration` object to the Kubernetes API.  The Linkerd CA notices this and creates TLS credentials for the webhook service and distributes them to to the webhook as a secret.  The webhook then waits until the secret is mounted and then uses the given credentials to serve over TLS.

## Problem

This approach is complex and difficult to follow.  We would like for the webhook to be able to serve TLS without all of this credential management.  Ideally:
* There would be little to no special case code in the CA for handling webhooks
* There would be little to no code in the webhook for managing certificates

alpeb:
Use `ca.NewCA()` for generating certs and keys for the proxy injector instead of grabbing the trust anchor from the linkerd-ca-bundle ConfigMap, as well as avoiding having to watch linkerd-proxy-injector-service-tls-linkerd-io secrets for the cert and key used for the injector's TLS server.

This is just a POC. If this approach is accepted, the next step is to remove all the code that was dealing with those secrets.

Fixes #2095


olix0r:
Should the `ca` logic be put somewhere in `pkg/` to indicate it's a library? Is the `controller/ca` package generic library code or code related to our current CA.  I'd opt to depend on generic code, rather than code related to our CA, which will have to change substantially in an upcoming change.

This isn't a blocker, just an organizational note.

alpeb:
@ver yea, `ca.go` is pretty generic so we can move it into `/pkg`. The more specific stuff dealing with the resources watching and secrets updating leaves in `/controller/ca/controller.go`.

controller/cmd/proxy-injector/main.go
```diff
@@ -40,7 +41,12 @@ func main() {
 		log.Fatalf("failed to mount the ca bundle: %s", err)
 	}
 
-	webhookConfig, err := injector.NewWebhookConfig(k8sClient, *controllerNamespace, *webhookServiceName, k8sPkg.MountPathTLSTrustAnchor)
+	rootCA, err := ca.NewCA()
```
klingerf: 👍 to @olix0r's suggesting of moving the `CA` struct and this func to `pkg/tls` or somewhere similar.

Note that there's already a very small `pkg/tls` package that provides the `PEMEncodeCert` and `PEMEncodeKey` funcs, and incidentally those functions are called from the `tlsConfig` function in controller/proxy-injector/server.go, which you're modifying as part of this branch. I think it would make sense to consolidate `controller/ca/ca.go` and `pkg/tls` into a single package as part of this change.

controller/proxy-injector/server.go
```diff
@@ -76,24 +77,17 @@ func (w *WebhookServer) Shutdown() error {
...
-	if err != nil {
-		return nil, err
-	}
-
-	keyBytes, err := ioutil.ReadFile(keyFile)
-	if err != nil {
-		return nil, err
-	}
+func tlsConfig(certFile, keyFile string, rootCA *ca.CA) (*tls.Config, error) {
+	dnsName := "linkerd-proxy-injector.linkerd.svc"
```
klingerf: This DNS name can change based on the namespace where Linkerd is installed. So it would be better to use the `controllerNamespace` variable that's passed into `NewWebhookServer` when constructing this string.

controller/proxy-injector/webhook_config.go
```diff
@@ -3,9 +3,10 @@ package injector
 import (
 	"bytes"
 	"encoding/base64"
-	"io/ioutil"
 	"text/template"
 
+	"github.com/linkerd/linkerd2/controller/ca"
+
```
klingerf: Extraneous newline.

controller/proxy-injector/server.go
klingerf: Or better yet, you could use the `TLSIdentity` from `pkg/k8s. I haven't tried it, but I think something like this would work:

```diff
diff --git a/controller/proxy-injector/server.go b/controller/proxy-injector/server.go
index bdeb9c32..a67622a2 100644
--- a/controller/proxy-injector/server.go
+++ b/controller/proxy-injector/server.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 
 	"github.com/linkerd/linkerd2/controller/ca"
+	"github.com/linkerd/linkerd2/pkg/k8s"
 	pem "github.com/linkerd/linkerd2/pkg/tls"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
@@ -22,7 +23,7 @@ type WebhookServer struct {
 
 // NewWebhookServer returns a new instance of the WebhookServer.
 func NewWebhookServer(client kubernetes.Interface, resources *WebhookResources, addr, controllerNamespace, certFile, keyFile string, rootCA *ca.CA) (*WebhookServer, error) {
-	c, err := tlsConfig(certFile, keyFile, rootCA)
+	c, err := tlsConfig(certFile, keyFile, rootCA, controllerNamespace)
 	if err != nil {
 		return nil, err
 	}
@@ -77,8 +78,14 @@ func (w *WebhookServer) Shutdown() error {
 	return w.Server.Shutdown(context.Background())
 }
 
-func tlsConfig(certFile, keyFile string, rootCA *ca.CA) (*tls.Config, error) {
-	dnsName := "linkerd-proxy-injector.linkerd.svc"
+func tlsConfig(certFile, keyFile string, rootCA *ca.CA, controllerNamespace string) (*tls.Config, error) {
+	tlsIdentity := k8s.TLSIdentity{
+		Name:                "linkerd-proxy-injector",
+		Kind:                k8s.Service,
+		Namespace:           controllerNamespace,
+		ControllerNamespace: controllerNamespace,
+	}
+	dnsName := tlsIdentity.ToDNSName()
 	certAndPrivateKey, err := rootCA.IssueEndEntityCertificate(dnsName)
 
 	certPEM, err := pem.PEMEncodeCert(certAndPrivateKey.Certificate)
```

controller/proxy-injector/server.go
```diff
@@ -76,24 +77,17 @@ func (w *WebhookServer) Shutdown() error {
...
+	certAndPrivateKey, err := rootCA.IssueEndEntityCertificate(dnsName)
 
-	certPEM, err := pem.PEMEncodeCert(certBytes)
+	certPEM, err := pem.PEMEncodeCert(certAndPrivateKey.Certificate)
```
klingerf: Right, per my comment above, maybe we can provide the functionality of `PEMEncodeCert` and `PEMEncodeKey` funcs directly from the CA itself.

controller/proxy-injector/server.go
alpeb: yep, will certainly do :+1:

alpeb:
My last push is broken into 4 commits like so:
- Moved ca.go into `/pkg/tls`
- Merged the smaller tls.go with ca.go and addressed @klingerf's comments
- Cleaned up and de-coupled the CA controller and the proxy-injector
- Removed the need to use `--tls` during install to have the proxy injector work
- Updated tests

Let me know what you think :pray:

pkg/tls/ca.go
```diff
@@ -167,6 +175,30 @@ func (ca *CA) createTemplate(publicKey *ecdsa.PublicKey) x509.Certificate {
 	}
 }
 
+// PEMEncodeCert returns the PEM encoding of cert.
+func PEMEncodeCert(cert []byte) ([]byte, error) {
```
klingerf: Since this func is only ever called on the `Certificate` field of the `CertificateAndPrivateKey` struct, we could do some information hiding and attach this function directly to that struct. Something like:

```
func (cpk *CertificateAndPrivateKey) EncodedCertificate() ([]byte, error) {
	buf := &bytes.Buffer{}
	err := pem.Encode(buf, &pem.Block{Type: "CERTIFICATE", Bytes: cpk.Certificate})
	return buf.Bytes(), err
}
```

Note that I also recommend renaming it to `EncodedCertificate`, to parallel the naming of the `Certificate` field in the struct.

pkg/tls/ca.go
```diff
@@ -167,6 +175,30 @@ func (ca *CA) createTemplate(publicKey *ecdsa.PublicKey) x509.Certificate {
...
+	buf := &bytes.Buffer{}
+	err := pem.Encode(buf, &pem.Block{Type: "CERTIFICATE", Bytes: cert})
+	return buf.Bytes(), err
+}
+
+// PEMEncodeKey returns the PEM encoding of key.
+func PEMEncodeKey(key []byte, keyType string) ([]byte, error) {
```
klingerf: Same comment as above for this encoded private key func. Also, it looks like we were only ever calling this func with `KeyTypeECDSA`. Since we're not using it, let's rip out support for different key types:

```
func (cpk *CertificateAndPrivateKey) EncodedPrivateKey() ([]byte, error) {
	buf := &bytes.Buffer{}
	err := pem.Encode(buf, &pem.Block{Type: "EC PRIVATE KEY", Bytes: cpk.PrivateKey})
	return buf.Bytes(), err
}
```

That will also allow you to remove the `KeyType*` consts at the top of this file.

controller/proxy-injector/fake/factory.go
```diff
@@ -151,72 +150,3 @@ func (f *Factory) Volume(filename string) (*corev1.Volume, error) {
 
 	return &volume, nil
 }
-
-// CATrustAnchors creates a fake CA trust anchors and returns the name of the
-// temporary file. Caller is responsible for deleting the file once it's done.
-func (f *Factory) CATrustAnchors() (string, error) {
```
klingerf: 😻 Wow, this is such a nice cleanup.

controller/proxy-injector/server.go
```diff
@@ -76,24 +77,23 @@ func (w *WebhookServer) Shutdown() error {
...
-	if err != nil {
-		return nil, err
+func tlsConfig(rootCA *pkgTls.CA, controllerNamespace string) (*tls.Config, error) {
+	tlsIdentity := k8s.TLSIdentity{
+		Name:                "linkerd-proxy-injector",
+		Kind:                k8s.Service,
+		Namespace:           controllerNamespace,
+		ControllerNamespace: controllerNamespace,
 	}
+	dnsName := tlsIdentity.ToDNSName()
```
adleong: What DNS name does this evaluate to?  `linkerd-proxy-injector.<controllerNS>.svc`?  Is that the DNS name that the k8s API will call this by?

cli/install/template.go
```diff
@@ -879,9 +879,6 @@ spec:
         - "ca"
         - "-controller-namespace={{.Values.Namespace}}"
         - "-single-namespace={{.Values.SingleNamespace}}"
-        {{- if and .Values.EnableTLS .Values.ProxyAutoInjectEnabled }}
-        - "-proxy-auto-inject={{ .Values.ProxyAutoInjectEnabled }}"
```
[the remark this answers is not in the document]
alpeb: ah, nice catch!

controller/proxy-injector/server.go
alpeb: That's right. Here's the k8s piece that prepares the client for that call:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/util/webhook/client.go#L151

pkg/tls/ca.go
alpeb: ok that looks good :+1:

pkg/tls/ca.go
alpeb: :cool:

pkg/tls/ca.go
```diff
@@ -11,6 +12,12 @@ import (
 	"time"
 )
 
+// Supported TLS key types.
+const (
+	KeyTypeRSA   = "rsa"
+	KeyTypeECDSA = "ecdsa"
+)
+
```
klingerf: These constants are no longer used and can be removed.

controller/proxy-injector/webhook_config_test.go
```diff
@@ -22,13 +21,12 @@ func TestCreateOrUpdate(t *testing.T) {
 		t.Fatal("Unexpected error: ", err)
 	}
 
-	trustAnchorsPath, err := factory.CATrustAnchors()
+	rootCA, err := tls.NewCA()
 	if err != nil {
-		t.Fatal("Unexpected error: ", err)
+		log.Fatalf("failed to create root CA: %s", err)
```
klingerf: This should be `t.Fatalf` to actually fail the test. Calling `log.Fatalf` will cause the test to exit immediately, but the error won't be printed, since all log output is being discarded by the `log.SetOutput(ioutil.Discard)` call on line 17 in this file.

```suggestion
		t.Fatalf("failed to create root CA: %s", err)
```

pkg/tls/ca.go
alpeb: oops, thanks!

controller/proxy-injector/webhook_config_test.go
alpeb: oh indeed. Nice catch.

PR #2170, opened 2019-01-29
WIP: Mute klog by default

### LINKED ISSUE #2168: Handle innocuous error messages during control plane startup

When the controller containers start up, they log a lot of connection refused lines. This is a timing issue between the proxy and the controller container. As the proxy is not yet ready to receive traffic, the connections get retried.

Unfortunately, this scares users and can provide misleading information around debugging. In the past, it was not possible to opt-out of these lines because of glog.

The underlying problem here is that a `Sync` is kicked off before the proxy is ready to forward traffic. This feels like a general linkerd problem, and possibly something that could help our users out.

I would like to see a way for sidecars to know when they can start sending traffic out. It feels like this could be a readiness probe for the proxy. Then, the controller code could wait for the probe to pass and continue starting.

### Alternatives considered

- Mute klog - there are valid messages that will just make debugging more difficult if klog is completely muted.
- Toggle klog around sync - feels pretty magical and fragile.

grampelberg:
Fixes #2168 

@siggy I really liked your idea to dynamically change the log level (it is possible). Unfortunately, the more I played around with it, the more magical it felt. I'd hate to have users expect a certain behavior, but only in the right context around log levels. It feels like a better solution to have it set globally and just document that you should turn it up if there are issues.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -14,9 +14,9 @@ import (
 // func calls flag.Parse(), so it should be called after all other flags have
 // been configured.
 func ConfigureAndParse() {
-	// override klog's default configuration and log to stderr instead of a file
+	// override klog's default configuration and send everything to /dev/null
 	klog.InitFlags(nil)
-	flag.Set("logtostderr", "true")
+	flag.Set("log_file", "/dev/null")
 
 	logLevel := flag.String("log-level", log.InfoLevel.String(),
 		"log level, must be one of: panic, fatal, error, warn, info, debug")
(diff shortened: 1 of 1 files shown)


grampelberg:
Don't look at this yet, I'm missing a magical config setting.

pkg/flags/flags.go
```diff
@@ -34,6 +34,12 @@ func setLogLevel(logLevel string) {
 		log.Fatalf("invalid log-level: %s", logLevel)
 	}
 	log.SetLevel(level)
+
+	// configure klog to be very verbose to stderr when debug logs are requested
+	if level == log.DebugLevel {
+		flag.Set("v", "3")
```
siggy: i'm not sure what level 3 means.

experimenting a bit with `kubectl`:
```bash
$ kubectl options|grep "log level"
  -v, --v=0: log level for V logs
```
kubectl provides additional info up to level 9:
```bash
$ diff <(kubectl version -v 8 2>&1 | wc -l) <(kubectl version -v 9 2>&1 | wc -l)
1c1
<       23
---
>       20
```

Some detail on these log levels:
https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md

I'd love to hook into some proper constants for this, and mapping our log levels to theirs would be ideal.

grampelberg:
@siggy thinking this through, you're right. The better solution would be to wait to sync until the proxy is ready to let us through to the API server. I'll close this out and update the tickets =)

PR #2189, opened 2019-02-01
Introduce go generate to embed static templates

### LINKED ISSUE #2153: Use packr to embed static template files into Linkerd binary
In order to allow installing Linkerd through [helm](https://github.com/helm/helm), the templates must be static files. Currently, the templates are `[]byte`s found in https://github.com/linkerd/linkerd2/blob/master/cli/install/template.go

If we changed the templates to static files and continued building Linkerd the way that we currently do, those files would need to be read directly off disk. This is not always an option depending on the way Linkerd was installed.

Using [packr](https://github.com/gobuffalo/packr) will allow us to build Linkerd with static files so that this is not an issue.

kleimkuhler:
# Problem
In order to switch Linkerd template rendering to use `.yaml` files, static
assets must be bundled in the Go binary for use by `linkerd install`.

# Solution
The solution should not affect the local development process of building and
testing.

[vfsgen](https://github.com/shurcooL/vfsgen) generates Go code that statically
implements the provided `http.FileSystem`. Paired with `go generate` and Go
[build tags](https://golang.org/pkg/go/build/), we can continue to use the
template files on disk when developing with no change required.

In `!prod` Go builds, the `cli/static/templates.go` file provides a
`http.FileSystem` to the local templates. In `prod` Go builds, `go generate
./cli` generates `cli/static/generated_templates.gogen.go` that statically
provides the template files.

When built with `-tags prod`, the executable will be built with the staticlly
generated file instead of the local files.

# Validation
The binaries were compiled locally with `bin/docker-build`. The binaries were
then tested with `bin/test-run (pwd)/target/cli/darwin/linkerd`. All tests
passed.

No change was required to successfully run `bin/go-run cli install`. No change
was required to run `bin/linkerd install`.

Fixes #2153


.gitignore
```diff
@@ -13,4 +13,5 @@ web/app/yarn-error.log
 .protoc
 .gorun
 .dep*
+**.gogen*
```
kleimkuhler: `go generate ./cli` generates the file `cli/static/generated_templates.gogen.go`. This will ignore that file if we are testing `go generate` in development.

cli/main.go
```diff
@@ -1,3 +1,5 @@
+//go:generate go run static/generate.go
```
kleimkuhler: `go:generate` is relative to the directory of the file it is annotated in. Therefore, `static/generate/.go` is valid.

cli/static/generate.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/linkerd/linkerd2/cli/static"
+	"github.com/shurcooL/vfsgen"
+)
+
+func main() {
+	err := vfsgen.Generate(static.Templates, vfsgen.Options{
+		Filename:     "static/generated_templates.gogen.go",
+		PackageName:  "static",
+		BuildTags:    "prod",
```
kleimkuhler: This produces the static Go file with the `// +build prod` tag. This tag is used in the `go build -tags prod` command.

cli/static/templates.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"net/http"
+	"os"
+	"path"
+)
+
+// Templates that will be rendered by `linkerd install`. This is only used on
+// dev builds so we can assume $GOPATH is set properly.
+var Templates http.FileSystem = http.Dir(path.Join(os.Getenv("GOPATH"), "src/github.com/linkerd/linkerd2/chart"))
```
kleimkuhler: We can assumed `$GOPATH` is set here because of the `// +build !prod` flag. It is only used in development where `go env` can be assumed to be valid.

kleimkuhler:
> this is looking good! i'd expect a bit of merge foo with #2163 and #2187.

I keep getting a lot of merge conflicts with moving the templates around. I'll continue adjusting until we can get this merged in!

cli/static/generate.go
```diff
@@ -0,0 +1,22 @@
+// +build ignore
+
+package main
+
+import (
+	log "github.com/sirupsen/logrus"
+
+	"github.com/linkerd/linkerd2/cli/static"
+	"github.com/shurcooL/vfsgen"
```
klingerf: If you remove the newline in the imports here, `gofmt` will sort the logrus import between the static and vfsgen imports.

Issue #2011, opened 2018-12-20
bug: reduced timeouts on API breaks checks on slow clusters


radu-matei:
related: #553, #586 

Hey!
While I clearly understand the motivation behind the 5 seconds timeout on checks, the fact that this is not configurable breaks the following use case - I have a dev cluster made of burstable instances - this means that most of the time, they are really, really slow, which makes the Kubernetes API server respond in most cases in a little over 5 seconds (and because it is a cheap dev cluster, I suspect there are other people using them in the same way).

Now, because of the change in #586  (and because the `--wait` flag for `linkerd check` is rather unclear in functionality - `Retry and wait for some checks`), it means that there is no way for me to successfully `linkerd check --pre` on this type of clusters.

One solution would be to add a new flag for this, or re-use the existing `--wait` flag (although I'd avoid changing the behaviour of a flag) - in any case, the default behaviour would still be dictated by #586 , thus not affecting the majority of clusters where response times are much lower.

I'd be more than glad to open a PR for this, but before I do that, I'd like an opinion first.

Thanks!

Edit: as I'm playing around the repo, it feels like the timeout for the Kuberentes API should be a global flag, as it also impacts installs and other configs.

klingerf:
@radu-matei Thanks for bringing this to our attention, and for all of the additional info about your setup. I agree that the timeout should be configurable via a global command line flag, and I'd be more than happy to review that change.

For the `check` command specifically, my preference is to use a separate flag instead of the `--wait` flag, since the `--wait` timeout spans all requests made in the process of completing a check run, whereas this timeout would apply to individual requests. Maybe we could call it `--request-timeout` and default it to 5 seconds?

Unfortunately, the 5-second timeout is hardcoded throughout `pkg/...`. It would be great if we could consolidate all of these as part of the change:

```
./pkg/k8s/api.go:	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
./pkg/k8s/api.go:	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
./pkg/k8s/api.go:	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
./pkg/healthcheck/healthcheck.go:			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
./pkg/version/version.go:	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
[...]

PR #2191, opened 2019-02-01
Consolidate timeouts for `linkerd check`

alpeb:
- Moved the creation of contexts from inside the methods targeted by the checks into a single place in the runCheck() and runCheckRPC() methods where the context is built using a hard-coded timeout of 30 seconds (we decided it's not necessary to expose this as a flag).
- Some checks rely on k8s' client-go, which doesn't allow passing along contexts, but it let's us setting the Timeout manually.
- Reworded the description for the --wait option.

How I tested this
------------------------
I introduced a sleep of 25s in the `Version` API and then I ran `linkerd check --verbose --wait 80s`.
We can see the "can query the control plane API" check (which uses the Version API) succeed after 25s:
image

Then I changed that sleep to 35s and now the check fails after 30s and it gets retried until we go past the 80s overall timeout:
image

Whenever the check call fails (as opposed to timing out), the check fails immediately (and is retried if `fatal` is false). I tested this just by removing the CP and running the checks. So it seems this wouldn't mean that increasing the timeout per-check would affect how we do retries. Although I would like to see this get better validated under more scenarios.

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -63,7 +63,7 @@ non-zero exit code.`,
 	cmd.PersistentFlags().StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
 	cmd.PersistentFlags().BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	cmd.PersistentFlags().BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
-	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Retry and wait for some checks to succeed if they don't pass the first time")
+	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for the totally of tests to pass")
 	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	cmd.PersistentFlags().BoolVar(&options.singleNamespace, "single-namespace", options.singleNamespace, "When running pre-installation checks (--pre), only check the permissions required to operate the control plane in a single namespace")
 
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -63,7 +63,7 @@ non-zero exit code.`,
 	cmd.PersistentFlags().StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
 	cmd.PersistentFlags().BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	cmd.PersistentFlags().BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
-	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for the totally of tests to pass")
+	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for all tests to pass")
 	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	cmd.PersistentFlags().BoolVar(&options.singleNamespace, "single-namespace", options.singleNamespace, "When running pre-installation checks (--pre), only check the permissions required to operate the control plane in a single namespace")
 
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -104,6 +104,7 @@ const (
 var (
 	maxRetries        = 60
 	retryWindow       = 5 * time.Second
+	requestTimeout    = 30 * time.Second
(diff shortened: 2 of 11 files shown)


cli/cmd/check.go
```diff
@@ -63,7 +63,7 @@ non-zero exit code.`,
 	cmd.PersistentFlags().StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
 	cmd.PersistentFlags().BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	cmd.PersistentFlags().BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
-	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Retry and wait for some checks to succeed if they don't pass the first time")
+	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for the totally of tests to pass")
```
siggy: How about `Maximum allowed time for all tests to pass`

pkg/healthcheck/healthcheck.go
```diff
@@ -222,22 +223,29 @@ func (hc *HealthChecker) allCategories() []category {
...
+					check: func(ctx context.Context) (err error) {
 						hc.httpClient, err = hc.kubeAPI.NewClient()
 						if err != nil {
-							return
+							return err
 						}
-						hc.kubeVersion, err = hc.kubeAPI.GetVersionInfo(hc.httpClient)
-						return
+						hc.kubeVersion, err = hc.kubeAPI.GetVersionInfo(ctx, hc.httpClient)
+						return err
```
siggy: recommend either using a named return (`func() (err error)`), or explicitly returning vars (`return err`), but not both. i have a _slight_ preference for explicitly returning vars, but care more that a function only implements one or the other.

a bit more info:
https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

pkg/healthcheck/healthcheck.go
```diff
@@ -222,22 +223,29 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "can initialize the client",
 					hintURL:     "https://linkerd.io/2/faq/#k8s-api",
 					fatal:       true,
-					check: func() (err error) {
+					check: func(_ context.Context) (err error) {
```
siggy: _slightly_ less verbose: `check: func(context.Context)`

pkg/version/channels.go
```diff
@@ -59,20 +58,17 @@ func (c Channels) Match(actualVersion string) error {
...
 }
 
-func getLatestVersions(client *http.Client, url string, uuid string, source string) (Channels, error) {
+func getLatestVersions(ctx context.Context, client *http.Client, url string, uuid string, source string) (Channels, error) {
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
 		return Channels{}, err
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
siggy: just confirming, that by delegating context creation to the healthcheck caller, we're increasing this timeout from 5s to 30s. i think this is ok?

pkg/healthcheck/healthcheck.go
```diff
@@ -601,8 +607,10 @@ func (hc *HealthChecker) Add(categoryID CategoryID, description string, hintURL
 			checkers: []checker{
 				checker{
 					description: description,
-					check:       check,
-					hintURL:     hintURL,
+					check: func(_ context.Context) error {
+						return check()
```
siggy: i think for consistency in testing (which i think there are a few to fix ;) ), this should be:
```go
func (hc *HealthChecker) Add(categoryID CategoryID, description string, hintURL string, check func(context.Context) error) {
...
	check: check,
```

pkg/healthcheck/healthcheck.go
```diff
@@ -104,6 +104,7 @@ const (
 var (
 	maxRetries        = 60
```
siggy: i just realized `maxRetries` is not used, mind removing?

cli/cmd/check.go
alpeb: Yep, sounds better :notes:

pkg/healthcheck/healthcheck.go
alpeb: Agreed. I missed that what we had originally was named returns. I'll revert it to how it was to avoid increasing the diff with an unnecessary change.

pkg/healthcheck/healthcheck.go
alpeb: oh, nice!

alpeb:
Fixes #2011

pkg/version/channels.go
alpeb: That's correct. To cater for the case of slow clusters like described in #2011

pkg/k8s/portforward.go
```diff
@@ -56,7 +58,9 @@ func NewPortForward(
 		return nil, err
 	}
 
-	pods, err := kubeAPI.GetPodsByNamespace(client, namespace)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	pods, err := kubeAPI.GetPodsByNamespace(ctx, client, namespace)
```
klingerf: Just curious about the choice of a 5-second timeout for this call when setting up the port-forward. Maybe it should be 30 seconds as well? Before we get to this code when running `linkerd dashboard`, we've already made other API calls, all with a 30 second timeout. If we're trying to talk to a cluster with really slow response times, my guess is that we would need a 30 second timeout here.

alpeb: I only changed the timeouts relevant to the checks, but yeah, I agree we can have this one be 30s as well.

PR #2196, opened 2019-02-03
Use the proper controller identity when configuring pods with TLS

olix0r:
Since 37ae423b, deployments have been prefixed with `linkerd-`; however
the inject logic was not changed to take this into consideration when
constructing the controller's identity.

This means that the proxy's client to the control plane has been unable to
establish TLS'd communcation to the proxy-api. Previously, the proxy would
silently fall back to plaintext, but in master this behavior recently changed to
be stricter, so this bug will prevent the proxy from connecting to proxy-api
in any way.


klingerf:
Ah, whoops, I missed the CI failure though. Thanks @dadjeibaah!

PR #2202, opened 2019-02-05
Reorganize install/inject config structs

klingerf:
This is a follow-up to #2193, which added configuration options for inbound/outbound keepalives to the `proxyConfigOptions` struct. It wasn't clear from the existing code, but the fields in that struct should map directly to command line flags. Since these configuration options don't have corresponding command line flags, I'm instead moving them to use a constant, and adding comments to explain how command line flags are configured. This also fixes an install issue that was incorrectly mapping the inbound keepalive to the outbound keepalive in the install template.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -222,60 +222,64 @@ func getPercentTLS(stats *pb.BasicStats) float64 {
 	return float64(stats.TlsRequestCount) / float64(reqTotal)
 }
 
+// proxyConfigOptions holds values for command line flags that apply to both the
+// install and inject commands. All fields in this struct should have
+// corresponding flags added in the addProxyConfigFlags func later in this file.
 type proxyConfigOptions struct {
-	linkerdVersion             string
-	proxyImage                 string
-	initImage                  string
-	dockerRegistry             string
-	imagePullPolicy            string
-	inboundPort                uint
-	outboundPort               uint
-	inboundAcceptKeepaliveMs   uint
-	outboundConnectKeepaliveMs uint
-	ignoreInboundPorts         []uint
-	ignoreOutboundPorts        []uint
-	proxyUID                   int64
-	proxyLogLevel              string
-	proxyAPIPort               uint
-	proxyControlPort           uint
-	proxyMetricsPort           uint
-	proxyCPURequest            string
-	proxyMemoryRequest         string
-	proxyOutboundCapacity      map[string]uint
-	tls                        string
-	disableExternalProfiles    bool
-	noInitContainer            bool
+	linkerdVersion          string
+	proxyImage              string
+	initImage               string
+	dockerRegistry          string
+	imagePullPolicy         string
+	inboundPort             uint
+	outboundPort            uint
+	ignoreInboundPorts      []uint
+	ignoreOutboundPorts     []uint
+	proxyUID                int64
+	proxyLogLevel           string
+	proxyAPIPort            uint
+	proxyControlPort        uint
+	proxyMetricsPort        uint
+	proxyCPURequest         string
+	proxyMemoryRequest      string
+	tls                     string
+	disableExternalProfiles bool
+	noInitContainer         bool
+
+	// proxyOutboundCapacity is a special case that's only used for injecting the
+	// proxy into the control plane install, and as such it does not have a
+	// corresponding command line flag.
+	proxyOutboundCapacity map[string]uint
 }
 
 const (
 	optionalTLS           = "optional"
 	defaultDockerRegistry = "gcr.io/linkerd-io"
+	defaultKeepaliveMs    = 10000
 )
 
 func newProxyConfigOptions() *proxyConfigOptions {
 	return &proxyConfigOptions{
-		linkerdVersion:             version.Version,
-		proxyImage:                 defaultDockerRegistry + "/proxy",
-		initImage:                  defaultDockerRegistry + "/proxy-init",
-		dockerRegistry:             defaultDockerRegistry,
-		imagePullPolicy:            "IfNotPresent",
-		inboundPort:                4143,
-		outboundPort:               4140,
-		ignoreInboundPorts:         nil,
-		ignoreOutboundPorts:        nil,
-		inboundAcceptKeepaliveMs:   10000,
-		outboundConnectKeepaliveMs: 10000,
-		proxyUID:                   2102,
-		proxyLogLevel:              "warn,linkerd2_proxy=info",
-		proxyAPIPort:               8086,
-		proxyControlPort:           4190,
-		proxyMetricsPort:           4191,
-		proxyOutboundCapacity:      map[string]uint{},
-		proxyCPURequest:            "",
-		proxyMemoryRequest:         "",
-		tls:                        "",
-		disableExternalProfiles:    false,
-		noInitContainer:            false,
+		linkerdVersion:          version.Version,
+		proxyImage:              defaultDockerRegistry + "/proxy",
+		initImage:               defaultDockerRegistry + "/proxy-init",
+		dockerRegistry:          defaultDockerRegistry,
+		imagePullPolicy:         "IfNotPresent",
+		inboundPort:             4143,
+		outboundPort:            4140,
+		ignoreInboundPorts:      nil,
+		ignoreOutboundPorts:     nil,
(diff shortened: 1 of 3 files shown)


cli/cmd/root.go
```diff
@@ -222,60 +222,64 @@ func getPercentTLS(stats *pb.BasicStats) float64 {
...
+	// proxyOutboundCapacity is a special case that's only used for injecting the
+	// proxy into the control plane install, and as such it does not have a
+	// corresponding command line flag.
+	proxyOutboundCapacity map[string]uint
 }
 
 const (
 	optionalTLS           = "optional"
 	defaultDockerRegistry = "gcr.io/linkerd-io"
+	defaultKeepaliveMs    = 10000
```
siggy: what do you think about defining this as a `time.Duration`?

```diff
diff --git a/cli/cmd/inject.go b/cli/cmd/inject.go
index b1371054..2757ed7a 100644
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -256,8 +256,8 @@ func injectPodSpec(t *v1.PodSpec, identity k8s.TLSIdentity, controlPlaneDNSNameO
 				Name:      PodNamespaceEnvVarName,
 				ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "metadata.namespace"}},
 			},
-			{Name: "LINKERD2_PROXY_INBOUND_ACCEPT_KEEPALIVE", Value: fmt.Sprintf("%dms", defaultKeepaliveMs)},
-			{Name: "LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE", Value: fmt.Sprintf("%dms", defaultKeepaliveMs)},
+			{Name: "LINKERD2_PROXY_INBOUND_ACCEPT_KEEPALIVE", Value: fmt.Sprintf("%s", defaultKeepalive)},
+			{Name: "LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE", Value: fmt.Sprintf("%s", defaultKeepalive)},
 		},
 		LivenessProbe:  &proxyProbe,
 		ReadinessProbe: &proxyProbe,
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index a5bd8a35..ebfe411e 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/linkerd/linkerd2/cli/install"
@@ -51,8 +52,8 @@ type installConfig struct {
 	OutboundPort                     uint
 	IgnoreInboundPorts               string
 	IgnoreOutboundPorts              string
-	InboundAcceptKeepaliveMs         uint
-	OutboundConnectKeepaliveMs       uint
+	InboundAcceptKeepalive           time.Duration
+	OutboundConnectKeepalive         time.Duration
 	ProxyAutoInjectEnabled           bool
 	ProxyInjectAnnotation            string
 	ProxyInjectDisabled              string
@@ -205,8 +206,8 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
 		OutboundPort:                     options.outboundPort,
 		IgnoreInboundPorts:               strings.Join(ignoreInboundPorts, ","),
 		IgnoreOutboundPorts:              strings.Join(ignoreOutboundPorts, ","),
-		InboundAcceptKeepaliveMs:         defaultKeepaliveMs,
-		OutboundConnectKeepaliveMs:       defaultKeepaliveMs,
+		InboundAcceptKeepalive:           defaultKeepalive,
+		OutboundConnectKeepalive:         defaultKeepalive,
 		ProxyAutoInjectEnabled:           options.proxyAutoInject,
 		ProxyInjectAnnotation:            k8s.ProxyInjectAnnotation,
 		ProxyInjectDisabled:              k8s.ProxyInjectDisabled,
diff --git a/cli/cmd/install_test.go b/cli/cmd/install_test.go
index 8c28d105..eb4ef4be 100644
--- a/cli/cmd/install_test.go
+++ b/cli/cmd/install_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"testing"
+	"time"
 )
 
 func TestRender(t *testing.T) {
@@ -55,8 +56,8 @@ func TestRender(t *testing.T) {
 		ControllerUID:                    2103,
 		InboundPort:                      4143,
 		OutboundPort:                     4140,
-		InboundAcceptKeepaliveMs:         10000,
-		OutboundConnectKeepaliveMs:       10000,
+		InboundAcceptKeepalive:           10 * time.Second,
+		OutboundConnectKeepalive:         10 * time.Second,
 		ProxyControlPort:                 4190,
 		ProxyMetricsPort:                 4191,
 		ProxyInitImage:                   "ProxyInitImage",
@@ -91,8 +92,8 @@ func TestRender(t *testing.T) {
 		ProxyUID:                         2102,
 		ControllerUID:                    2103,
 		EnableTLS:                        true,
-		InboundAcceptKeepaliveMs:         10000,
-		OutboundConnectKeepaliveMs:       10000,
+		InboundAcceptKeepalive:           10 * time.Second,
+		OutboundConnectKeepalive:         10 * time.Second,
 		TLSTrustAnchorConfigMapName:      "TLSTrustAnchorConfigMapName",
 		ProxyContainerName:               "ProxyContainerName",
 		TLSTrustAnchorFileName:           "TLSTrustAnchorFileName",
diff --git a/cli/cmd/root.go b/cli/cmd/root.go
index b5ef572e..35af20e1 100644
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -255,7 +255,7 @@ type proxyConfigOptions struct {
 const (
 	optionalTLS           = "optional"
 	defaultDockerRegistry = "gcr.io/linkerd-io"
-	defaultKeepaliveMs    = 10000
+	defaultKeepalive      = 10 * time.Second
 )
 
 func newProxyConfigOptions() *proxyConfigOptions {
diff --git a/cli/install/template.go b/cli/install/template.go
index b64ef6aa..d126b192 100644
--- a/cli/install/template.go
+++ b/cli/install/template.go
@@ -1076,13 +1076,13 @@ data:
       valueFrom:
         fieldRef:
           fieldPath: metadata.namespace
-    {{- if .Values.InboundAcceptKeepaliveMs }}
+    {{- if .Values.InboundAcceptKeepalive }}
     - name: LINKERD2_PROXY_INBOUND_ACCEPT_KEEPALIVE
-      value: {{.Values.InboundAcceptKeepaliveMs}}ms
+      value: {{.Values.InboundAcceptKeepalive}}
     {{- end}}
-    {{- if .Values.OutboundConnectKeepaliveMs }}
+    {{- if .Values.OutboundConnectKeepalive }}
     - name: LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE
-      value: {{.Values.OutboundConnectKeepaliveMs}}ms
+      value: {{.Values.OutboundConnectKeepalive}}
     {{- end}}
     - name: LINKERD2_PROXY_TLS_TRUST_ANCHORS
       value: /var/linkerd-io/trust-anchors/{{.Values.TLSTrustAnchorFileName}}
```

cli/cmd/root.go
```diff
@@ -325,24 +329,27 @@ func (options *proxyConfigOptions) taggedProxyInitImage() string {
 	return fmt.Sprintf("%s:%s", image, options.linkerdVersion)
 }
 
+// addProxyConfigFlags adds command line flags for all fields in the
+// proxyConfigOptions struct. To keep things organized, the flags should be
```
siggy: 👍

cli/cmd/root.go
klingerf: Yeah! I like this a lot. Will update.

klingerf: Ah, hmm, it looks like the formatting of the duration is unpredictable. When I apply this diff and run the tests, in some places I see the new duration represented as `10s` and in other places I see it as `1e+10`, which wreaks havoc with the tests. I'm going to leave it as is for now.

olix0r: Yeah, this was intentionally not a time.Duration because the proxy doesn't support reading the full range of go-formatted durations.  Values like 1s+10ms would break the proxy, so we just noramlize it into millis

PR #2206, opened 2019-02-05
Build request urls in the most generic possible way

### LINKED ISSUE #2132: Kubernetes server must not have a `/` suffix
### What is the issue?
I'm getting 404 response when the `server` field in my kubeconfig has a `/` suffix.
E.g. 'server: https://1.2.3.4/'

Removing the trailing `/` solves the problem.

Note that `kubectl` has no problem with exactly the same kubeconfig file.

### How can it be reproduced?

Edit kubeconfig file, add trailing slash to server endpoint.

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
$ linkerd check
kubernetes-api: can query the Kubernetes API...............................[FAIL] -- Unexpected Kubernetes API response: 404 Not Found
```

### Environment

- Kubernetes Version: 1.10.9
- Cluster Environment: GKE
- Host OS:  macos
- Linkerd version: 2.1

alpeb:
Fixes #2132

### DIFF
--- a/pkg/k8s/testdata/config.test
+++ b/pkg/k8s/testdata/config.test
@@ -16,6 +16,10 @@ clusters:
     certificate-authority-data: cXVlIHBhcmFkYSBhdHJhc2FkYQ==
     server: https://162.128.50.10
   name: cluster4
+- cluster:
+    certificate-authority-data: cXVlIHBhcmFkYSBhdHJhc2FkYQ==
+    server: https://162.128.50.11/
+    name: clusterTrailingSlash
 contexts:
 - context:
     cluster: cluster3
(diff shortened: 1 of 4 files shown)


alpeb:
Some checks requests were failing when the cluster server entry in `~/.kube/config` had a trailing slash, like `server: https://192.168.99.100:8443/`
We were using string concatenation and `fmt.Sprintf` to generate some URLs, so I changed that to use `url.ResolveReference` instead.

ihcsim:
LGTM.

alpeb:
Thanks @kleimkuhler and @siggy for the suggestions. I've consolidated the dupe code, and added a test around a config with a trailing slash.

pkg/k8s/testdata/config.test
```diff
@@ -16,6 +16,10 @@ clusters:
     certificate-authority-data: cXVlIHBhcmFkYSBhdHJhc2FkYQ==
     server: https://162.128.50.10
   name: cluster4
+- cluster:
+    certificate-authority-data: cXVlIHBhcmFkYSBhdHJhc2FkYQ==
+    server: https://162.128.50.11/
```
siggy: nice!

PR #2089, opened 2019-01-15
When injecting, perform an uninject as a first step

alpeb:
Fixes #1970

The fixture `inject_emojivoto_already_injected.input.yml` is no longer
rejected, so I created the corresponding golden file.

Note that we'll still forbid injection over resources already injected
with third party meshes (Istio, Contour), so now we have
`HasExisting3rdPartySidecars()` to detect that.

PR #2209, opened 2019-02-05
Update auto-injector to require opt-in by namespace or pod

### LINKED ISSUE #1862: Flexible auto-injector targeting
## Feature Request

I propose to consider more flexible auto-injector targeting. 

Auto-injecting is an awesome feature, but we want to have more control over it. For example, we want to auto-inject only for namespaces **a** and **b** without labeling all other existing and future namespaces with `linkerd.io/auto-inject: disabled`. And in addition, to have even more granular control, we may also want to require an explicit annotation on the pod template like `linkerd.io/auto-inject: enabled`.

Maybe we can add support for additional flags for **install** command that will enable auto-injection only on explicitly annotated namespaces and pods.

As discussed per https://github.com/linkerd/linkerd2/issues/1820#issuecomment-438066895

klingerf:
This branch updates the control plane's proxy-injector process to only inject pods that have the `linkerd.io/inject: enabled` annotation on the pod or namespace. This is a breaking change from the existing behavior, of injecting all pods unless the pod or namespace has the `linkerd.io/auto-inject: disabled` label, but it will provide more configurability for auto-inject in the long run.

Fixes #1862.


klingerf:
There's a corresponding docs change in linkerd/website#127, which describes the new behavior once this PR merges.

controller/proxy-injector/webhook_test.go
```diff
@@ -61,26 +62,58 @@ func TestMutate(t *testing.T) {
 	}
 }
 
-func TestIgnore(t *testing.T) {
-	ns, err := factory.Namespace("namespace-kube-public.yaml")
+func TestShouldInject(t *testing.T) {
+	nsEnabled, err := factory.Namespace("namespace-kube-public.yaml")
```
siggy: naming nit: mind calling these files `namespace-inject-enabled.yaml` and `namespace-inject-disabled.yaml`

controller/proxy-injector/webhook.go
```diff
@@ -212,14 +212,19 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
...
-//   linkerd.io/inject annotation set to "enabled"; or
+//   "enabled", and the deployment's pod spec does not have the
+//   linkerd.io/inject annotation set to "disabled"; or
 // - the deployment's pod spec has the linkerd.io/inject annotation set to
-//   "disabled"
-func (w *Webhook) ignore(ns string, deployment *appsv1.Deployment) (bool, error) {
+//   "enabled"
+func (w *Webhook) shouldInject(ns string, deployment *appsv1.Deployment) (bool, error) {
+	if healthcheck.HasExistingSidecars(&deployment.Spec.Template.Spec) {
+		return false, nil
```
siggy: this may be out of scope for this PR, but i think putting the `HasExistingSidecars` first changes the semantics a bit? this would prevent re-injection of pod following a Linkerd control-plane upgrade.

if we do want to support the upgrade case, perhaps we need to modify `HasExistingSidecars` to distinguish between a `linkerd-proxy` and other proxies?

alpeb: We briefly did that in the `uninject before inject` PR but then rolled that back, the reason being already injected pods might contained special settings that might get lost during a proxy upgrade
https://github.com/linkerd/linkerd2/pull/2089#discussion_r248463876

ihcsim: Good points. At the moment, this webhook is only triggered when new deployments are created. Yes, as part of the story for data plane upgrade in an auto-inject env, `HasExistingSidecars` will need to be a bit smarter.

PR #2218, opened 2019-02-06
Improve ServiceProfile validation in linkerd check

### LINKED ISSUE #2190: Improve validation of Service profiles in linkerd check
Issue #2075 describes how an validating admission webhook can be used to do extensive validation of service profiles.  In the meantime before that issue is complete, we should add more thorough validation of service profiles the `linkerd check`.  The validation of service profiles done by `linkerd check` today is very limited.

In particular, `linkerd check` should validate that each service profile:
* Does not contain any unrecognized fields
* Contains all of the required fields
* Recursively validate the structure of the request and response conditions.

siggy:
The `linkerd check` command was doing limited validation on
ServiceProfiles.

Make ServiceProfile validation more complete, specifically validate:
- types of all fields
- presence of required fields
- presence of unknown fields
- recursive fields

Also move all validation code into a new `Validate` function in the
profiles package.

Validation of field types and required fields is handled via
`yaml.UnmarshalStrict` in the `Validate` function. This motivated
migrating from github.com/ghodss/yaml to a fork, sigs.k8s.io/yaml.

Fixes #2190

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1336,6 +1336,7 @@
     "k8s.io/helm/pkg/renderutil",
     "k8s.io/helm/pkg/timeconv",
     "k8s.io/klog",
+    "sigs.k8s.io/yaml",
   ]
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/chart/templates/base.yaml
+++ b/chart/templates/base.yaml
@@ -286,12 +286,16 @@ spec:
           - routes
           properties:
             retryBudget:
+              required:
+              - minRetriesPerSecond
+              - retryRatio
+              - ttl
               type: object
               properties:
-                retryRatio:
-                  type: number
                 minRetriesPerSecond:
                   type: integer
+                retryRatio:
+                  type: number
                 ttl:
                   type: string
             routes:
--- a/chart/templates/base.yaml
+++ b/chart/templates/base.yaml
@@ -285,6 +285,15 @@ spec:
           required:
           - routes
           properties:
+            retryBudget:
+              type: object
+              properties:
+                retryRatio:
+                  type: number
+                minRetriesPerSecond:
+                  type: integer
+                ttl:
+                  type: string
             routes:
               type: array
               items:
@@ -294,6 +303,8 @@ spec:
                 - condition
                 properties:
                   name:
+                    type: string
+                  timeout:
                     type: string
                   condition:
                     type: object
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -102,9 +102,8 @@ const (
 )
 
 var (
-	retryWindow       = 5 * time.Second
-	requestTimeout    = 30 * time.Second
-	clusterZoneSuffix = []string{"svc", "cluster", "local"}
+	retryWindow    = 5 * time.Second
+	requestTimeout = 30 * time.Second
 )
 
 type checker struct {
@@ -827,47 +826,19 @@ func (hc *HealthChecker) validateServiceProfiles() error {
 	}
 
 	for _, p := range svcProfiles.Items {
-		nameParts := strings.Split(p.Name, ".")
-		if len(nameParts) != 2+len(clusterZoneSuffix) {
-			return fmt.Errorf("ServiceProfile \"%s\" has invalid name (must be \"<service>.<namespace>.svc.cluster.local\")", p.Name)
+		service, namespace, err := profiles.ValidateName(p.Name)
+		if err != nil {
+			return err
 		}
-		for i, part := range nameParts[2:] {
-			if part != clusterZoneSuffix[i] {
-				return fmt.Errorf("ServiceProfile \"%s\" has invalid name (must be \"<service>.<namespace>.svc.cluster.local\")", p.Name)
-			}
-		}
-		service := nameParts[0]
-		namespace := nameParts[1]
-		_, err := hc.clientset.Core().Services(namespace).Get(service, meta_v1.GetOptions{})
+
+		_, err = hc.clientset.Core().Services(namespace).Get(service, meta_v1.GetOptions{})
 		if err != nil {
 			return fmt.Errorf("ServiceProfile \"%s\" has unknown service: %s", p.Name, err)
 		}
-		for _, route := range p.Spec.Routes {
-			if route.Name == "" {
-				return fmt.Errorf("ServiceProfile \"%s\" has a route with no name", p.Name)
-			}
-			if route.Timeout != "" {
-				_, err := time.ParseDuration(route.Timeout)
-				if err != nil {
-					return fmt.Errorf("ServiceProfile \"%s\" has a route with an invalid timeout: %s", p.Name, err)
-				}
-			}
-			if route.Condition == nil {
-				return fmt.Errorf("ServiceProfile \"%s\" has a route with no condition", p.Name)
-			}
-			err = profiles.ValidateRequestMatch(route.Condition)
-			if err != nil {
-				return fmt.Errorf("ServiceProfile \"%s\" has a route with an invalid condition: %s", p.Name, err)
-			}
-			for _, rc := range route.ResponseClasses {
-				if rc.Condition == nil {
-					return fmt.Errorf("ServiceProfile \"%s\" has a response class with no condition", p.Name)
-				}
-				err = profiles.ValidateResponseMatch(rc.Condition)
-				if err != nil {
-					return fmt.Errorf("ServiceProfile \"%s\" has a response class with an invalid condition: %s", p.Name, err)
-				}
-			}
+
+		err = profiles.ValidateSP(p)
+		if err != nil {
+			return err
 		}
 	}
 	return nil
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -836,7 +836,14 @@ func (hc *HealthChecker) validateServiceProfiles() error {
 			return fmt.Errorf("ServiceProfile \"%s\" has unknown service: %s", p.Name, err)
 		}
 
-		err = profiles.ValidateSP(p)
+		// TODO: remove this check once we implement ServiceProfile validation via a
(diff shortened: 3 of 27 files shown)


pkg/profiles/profiles.go
```diff
@@ -306,6 +315,90 @@ func ToRequestMatch(reqMatch *sp.RequestMatch) (*pb.RequestMatch, error) {
...
+				return fmt.Errorf("ServiceProfile \"%s\" has a response class with an invalid condition: %s", serviceProfile.Name, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+// ValidateName validates that a ServiceProfile's name is of the form:
+// <service>.<namespace>.svc.cluster.local
```
adleong: We can do this in a follow-up but this validation will have to be loosened to support external names.  We should change this to only validate that the name is a valid DNS name.

Gopkg.lock
```diff
@@ -1334,6 +1334,7 @@
     "k8s.io/helm/pkg/renderutil",
     "k8s.io/helm/pkg/timeconv",
     "k8s.io/klog",
+    "sigs.k8s.io/yaml",
```
kleimkuhler: Now that we have a direct dependency on `sigs.k8s.io/yaml`, should we add it to `Gopkg.yaml` so that we can lock the version?

```
[[constraint]]
  name = "github.com/kubernetes-sigs/yaml"
  version = "v1.1.0"
```

pkg/healthcheck/healthcheck.go
```diff
@@ -827,47 +826,19 @@ func (hc *HealthChecker) validateServiceProfiles() error {
...
-			}
+		service, namespace, err := profiles.ValidateName(p.Name)
+		if err != nil {
+			return err
 		}
-		service := nameParts[0]
-		namespace := nameParts[1]
-		_, err := hc.clientset.Core().Services(namespace).Get(service, meta_v1.GetOptions{})
+
+		_, err = hc.clientset.Core().Services(namespace).Get(service, meta_v1.GetOptions{})
```
kleimkuhler: We should be explicit about the version of `CoreClient` here

```suggestion
		_, err = hc.clientset.CoreV1().Services(namespace).Get(service, meta_v1.GetOptions{})
```

Gopkg.lock
siggy: Unless we strictly need to pin to a version, I'd rather let `sigs.k8s.io/yaml` float with the version `client-go`/`kubernetes` pulls in. We previously weren't specifying `github.com/ghodss/yaml` in `Gopkg.toml`. Hopefully this gets merged back into `github.com/ghodss/yaml` at some point.

pkg/profiles/profiles.go
siggy: Filed #2220 to track.

Gopkg.lock
kleimkuhler: Oh that makes sense to let `client-go`/`kubernetes` pull it in. Sounds good!

chart/templates/base.yaml
```diff
@@ -285,6 +285,15 @@ spec:
...
           properties:
+            retryBudget:
+              type: object
+              properties:
+                retryRatio:
+                  type: number
+                minRetriesPerSecond:
+                  type: integer
+                ttl:
+                  type: string
```
alpeb: I'm not sure if `retryRatio` and `minRetriesPerSecond` are mandatory, but it seems `ttl` is, otherwise this would fail:
https://github.com/linkerd/linkerd2/blob/master/pkg/profiles/profiles.go#L61-L77
so we'd need something like:
```
         properties:
            retryBudget:
              type: object
              required:
              - ttl
              properties:
                retryRatio:
                  type: number
                minRetriesPerSecond:
                  type: integer
                ttl:
                  type: string
```

alpeb: ... or better yet refactor `profiles.ToServiceProfile()` to apply a default when this field is missing.

pkg/profiles/profiles.go
```diff
@@ -306,6 +315,90 @@ func ToRequestMatch(reqMatch *sp.RequestMatch) (*pb.RequestMatch, error) {
...
+
+	for _, route := range serviceProfile.Spec.Routes {
+		if route.Name == "" {
+			return fmt.Errorf("ServiceProfile \"%s\" has a route with no name", serviceProfile.Name)
+		}
+		if route.Timeout != "" {
+			_, err := time.ParseDuration(route.Timeout)
+			if err != nil {
+				return fmt.Errorf("ServiceProfile \"%s\" has a route with an invalid timeout: %s", serviceProfile.Name, err)
+			}
```
alpeb: Looks like we could have a similar duration validation for `retryBudget.ttl`

chart/templates/base.yaml
adleong: For the numeric fields (retryRatio and minRetriesPerSecond) it is impossible to distinguish between an explicit 0 value and the field not being present.  Therefore, for those fields, it is impossible to apply a default when the field is missing.  For consistency, I think the ttl field should work the same way.  

I think what that means is that all three fields should be required.

pkg/profiles/profiles.go
```diff
@@ -306,6 +315,90 @@ func ToRequestMatch(reqMatch *sp.RequestMatch) (*pb.RequestMatch, error) {
 	}, nil
 }
 
+// ValidateSP takes a v1alpha1.ServiceProfile object, marshals to bytes, and
+// wraps `Validate`.
+func ValidateSP(serviceProfile sp.ServiceProfile) error {
```
adleong: Is this now unused?  Should it be removed?

pkg/profiles/profiles.go
```diff
@@ -306,6 +315,90 @@ func ToRequestMatch(reqMatch *sp.RequestMatch) (*pb.RequestMatch, error) {
...
+// This function validates:
+// - types of all fields
+// - presence of required fields
+// - presence of unknown fields
+// - recursive fields
+func Validate(data []byte) error {
+	var serviceProfile sp.ServiceProfile
+	err := yaml.UnmarshalStrict([]byte(data), &serviceProfile)
+	if err != nil {
+		return fmt.Errorf("failed to validate ServiceProfile: %s", err)
```
adleong: It would be helpful to have the name of the service profile in this error message.  It's not available here since the unmarshal failed, but it is available in the health check this gets called from.  We could have the health wrap the error message with the profile name and also remove the profile name from the other errors in this method.

siggy: I opted to modify the healthcheck to include the ServiceProfile name, but leave the underlying errors as-is. Although this means error messages may print the name twice, I am assuming when we start calling `profiles.Validate` from a `ValidatingAdmissionWebhook`, we will want the name included in the error message.

pkg/profiles/profiles.go
```diff
@@ -380,6 +370,18 @@ func Validate(data []byte) error {
 		}
 	}
 
+	rb := serviceProfile.Spec.RetryBudget
+	if rb != nil {
+		if rb.MinRetriesPerSecond == 0 || rb.RetryRatio == 0 || rb.TTL == "" {
```
adleong: I think we need to allow zero values here.  (don't get me started about how mad I am that you can't differentiate between 0 and unset 😡)

A more appropriate validation would be that `minRetriesPerSecond` and `retryRatio` cannot be negative.

siggy: @adleong My assumption was that because these are now required fields, they should not be set to zero?

Related, the `if rb != nil {` check above this should handle the case where no `RetryBudget` is set.

siggy: Another related question: is it possible to make these fields `*int`, etc, to indicate whether it was set?

adleong: It must be possible to explicitly set them to 0.  i.e.

```
retryBudget:
  retryRatio: 0.0
  minRetriesPerSecond: 0
  ttl: 10m
```

adleong: > Another related question: is it possible to make these fields `*int`, etc, to indicate whether it was set?

I don't know.  Is this an existing pattern?  I will investigate but in the meantime, I think this validation needs to allow 0 values.

siggy: Agreed that is not a change I'd want to make in this PR, but it is an existing pattern:
https://github.com/linkerd/linkerd2/blob/a5628780f7dbeff47549c7270af5549ec53c0434/cli/cmd/routes.go#L229-L242

pkg/profiles/profiles.go
```diff
@@ -372,8 +372,12 @@ func Validate(data []byte) error {
 
 	rb := serviceProfile.Spec.RetryBudget
 	if rb != nil {
-		if rb.MinRetriesPerSecond == 0 || rb.RetryRatio == 0 || rb.TTL == "" {
-			return fmt.Errorf("ServiceProfile \"%s\" RetryBudget missing fields (requires minRetriesPerSecond, retryRatio, ttl)", serviceProfile.Name)
+		if rb.RetryRatio < 0 {
+			return fmt.Errorf("ServiceProfile \"%s\" RetryBudget RetryRatio must be greater than zero: %f", serviceProfile.Name, rb.RetryRatio)
```
adleong: Nit on the error message: "RetryBudget RetryRatio must be greater than or equal to zero:" or "etryBudget RetryRatio must be non-negative:"

pkg/profiles/profiles.go
```diff
@@ -372,8 +372,12 @@ func Validate(data []byte) error {
...
+		}
+
```
adleong: We should also validate that MinRetriesPerSecond is non-negative also

siggy: `MinRetriesPerSecond` is a `uint32` so will be validating via `yaml.UnmarshalStrict`:
```
failed to validate ServiceProfile: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal number -5 into Go struct field RetryBudget.minRetriesPerSecond of type uint32
```

adleong: fantastic!

PR #2221, opened 2019-02-07
Modify klog to be verbose when controller log-level is set to debug

### LINKED ISSUE #2171: Map klog levels to controller log levels

Currently, changing `--controller-log-level` does not change klog's level.

Map the `-v` levels of klog to `panic, fatal, error, warn, info, debug`. Kubernetes documents the [log levels](https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md).

Perhaps:

- `-v 0` - panic, fatal, error, warn
- `-v 1` - info
- `-v 2` - info
- `-v 3` - debug
- `-v 4` - debug
- `-v 5` - Not sure, maybe we should add a trace level? @siggy found differences between logging at `-v 8` even. When does it stop?

Note, along with `-v` for log levels, klog also uses `stderrthreshold`. The possible values for that are info, warning, error, fatal. These can also be mapped to the linkerd log levels.

dadjeibaah:
The controller logs innocuous messages when control plane proxies aren't ready to route requests during startup from each control plane component. i.e. tap, public-api and proxy-api. Setting the log level in the control plane to `INFO` would not hide these log messages and would still show up on control plane startup.

This PR modifies `klogs` initial flag set to route innocuous logs to `/dev/null` if the controller log level is set to INFO. If set to debug, we output all loglines to stderr.

Fixes #2171 #2168 

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -3,6 +3,7 @@ package flags
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 
 	"github.com/linkerd/linkerd2/pkg/version"
@@ -14,10 +15,13 @@ import (
 // func calls flag.Parse(), so it should be called after all other flags have
 // been configured.
 func ConfigureAndParse() {
-	// override klog's default configuration and log to stderr instead of a file
-	klog.InitFlags(nil)
-	flag.Set("logtostderr", "true")
+	// -stderrthreshold=FATAL forces klog to only log FATAL errors to stderr.
+	// -logtostderr to false to not log to stderr by default.
+	var klogFlags = flag.NewFlagSet("klog", flag.ExitOnError)
 
+	klog.InitFlags(klogFlags)
+	klogFlags.Set("stderrthreshold", "FATAL")
+	klogFlags.Set("logtostderr", "false")
 	logLevel := flag.String("log-level", log.InfoLevel.String(),
 		"log level, must be one of: panic, fatal, error, warn, info, debug")
 	printVersion := flag.Bool("version", false, "print version and exit")
@@ -34,6 +38,13 @@ func setLogLevel(logLevel string) {
 		log.Fatalf("invalid log-level: %s", logLevel)
 	}
 	log.SetLevel(level)
+
+	klog.SetOutput(ioutil.Discard)
+	// Anything lower than the INFO level according to log is sent to /dev/null
+	if level == log.DebugLevel {
+		// Set stderr to INFO severity see https://github.com/kubernetes/klog/issues/23
+		klog.SetOutputBySeverity("INFO", os.Stderr)
+	}
 }
 
 func maybePrintVersionAndExit(printVersion bool) {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -3,7 +3,6 @@ package flags
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"os"
 
 	"github.com/linkerd/linkerd2/pkg/version"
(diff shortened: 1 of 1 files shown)


pkg/flags/flags.go
```diff
@@ -14,10 +15,13 @@ import (
 // func calls flag.Parse(), so it should be called after all other flags have
 // been configured.
 func ConfigureAndParse() {
-	// override klog's default configuration and log to stderr instead of a file
-	klog.InitFlags(nil)
-	flag.Set("logtostderr", "true")
+	// -stderrthreshold=FATAL forces klog to only log FATAL errors to stderr.
+	// -logtostderr to false to not log to stderr by default.
```
grampelberg: You'll want to set `-v` to 0 here as well.

pkg/flags/flags.go
```diff
@@ -34,6 +38,13 @@ func setLogLevel(logLevel string) {
 		log.Fatalf("invalid log-level: %s", logLevel)
 	}
 	log.SetLevel(level)
+
+	klog.SetOutput(ioutil.Discard)
```
grampelberg: No reason to do this. But you will want to set the log file to `/dev/null`.

pkg/flags/flags.go
```diff
@@ -34,6 +38,13 @@ func setLogLevel(logLevel string) {
...
+	// Anything lower than the INFO level according to log is sent to /dev/null
+	if level == log.DebugLevel {
+		// Set stderr to INFO severity see https://github.com/kubernetes/klog/issues/23
+		klog.SetOutputBySeverity("INFO", os.Stderr)
```
grampelberg: Don't forget a `-v` of 8 or 10.

grampelberg: Just `-stderrthreshold` to `info` and that's good enough.

pkg/flags/flags.go
dadjeibaah: While working on this I found that any flags that were set in this method would not have any effect on the logs because we would be modifying flags after `flags.Parse()` was called. So setting the `log_file` to `/dev/null` would not work here.  Setting the output to `ioutil.DIscard`seemed to work.

pkg/flags/flags.go
dadjeibaah: Same deal here as my previous comment pointed out.

pkg/flags/flags.go
grampelberg: Looks like it is because you created the second flag set. This works:

```
func ConfigureAndParse() {
	// -stderrthreshold=FATAL forces klog to only log FATAL errors to stderr.
	// -logtostderr to false to not log to stderr by default.

	klog.InitFlags(nil)
	flag.Set("stderrthreshold", "FATAL")
	flag.Set("logtostderr", "false")
	flag.Set("log_file", "/dev/null")
	flag.Set("v", "0")
	logLevel := flag.String("log-level", log.InfoLevel.String(),
		"log level, must be one of: panic, fatal, error, warn, info, debug")
	printVersion := flag.Bool("version", false, "print version and exit")

	flag.Parse()

	setLogLevel(*logLevel)
	maybePrintVersionAndExit(*printVersion)
}

func setLogLevel(logLevel string) {
	level, err := log.ParseLevel(logLevel)
	if err != nil {
		log.Fatalf("invalid log-level: %s", logLevel)
	}
	log.SetLevel(level)

	// Anything lower than the INFO level according to log is sent to /dev/null
	if level == log.DebugLevel {
		flag.Set("stderrthreshold", "INFO")
		flag.Set("v", "10")
	}
}
```

pkg/flags/flags.go
```diff
@@ -34,6 +35,12 @@ func setLogLevel(logLevel string) {
 		log.Fatalf("invalid log-level: %s", logLevel)
 	}
 	log.SetLevel(level)
+
+	if level == log.DebugLevel {
+		flag.Set("stderrthreshold", "INFO")
+		flag.Set("logtostderr", "true")
+		flag.Set("v", "10")
```
klingerf: TIL you can call `flag.Set` _after_ `flag.Parse` and it will still work. I wasn't expecting that, but have verified that this implementation does what you want it to do.

grampelberg: Honestly, it frightens me a little bit. There be dragons there =)

PR #2222, opened 2019-02-07
CNI: Removed unnecessary tolerations; added created-by annotation.

codeman9:

In our cluster, we have some nodes that are tainted with `NoSchedule` and this means that only "master"-level pods will be installed on those nodes (pods that tolerate `NoSchedule`). The linkerd-cni plugin only needs to be installed and executed on nodes that will have "user" pods on them so the output .yaml does not need any special tolerations. Installing the linkerd-cni plugin on nodes that will never have "user" pods scheduled on them by default is unnecessary.

I also removed the `scheduler.alpha.kubernetes.io/critical-pod: ''` based on this documentation which mentions that the annotation is only effective if the pod is in the `kube-system` namespace: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

### DIFF
--- a/cli/install/cni-template.go
+++ b/cli/install/cni-template.go
@@ -106,6 +106,8 @@ metadata:
   namespace: {{.Namespace}}
   labels:
     k8s-app: linkerd-cni
+  annotations:
+    {{.CreatedByAnnotation}}: {{.CliVersion}}
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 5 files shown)


cli/install/cni-template.go
```diff
@@ -118,25 +120,10 @@ spec:
     metadata:
       labels:
         k8s-app: linkerd-cni
-      annotations:
```
siggy: As a matter of convention, we typically add the `CreatedByAnnotation` on both the pod template as well as the parent object:
https://github.com/linkerd/linkerd2/blob/66070c26f4c5639fd94f99345b2cfba066b9134e/chart/templates/base.yaml#L151-L160

codeman9: 👍

PR #2237, opened 2019-02-08
Validate service profiles in all namespaces

### LINKED ISSUE #2220: Loosen ServiceProfile name validation to support all DNS names
The ServiceProfile validation code expects names of the form `<service>.<namespace>.svc.cluster.local`. This should be loosened to support any DNS name.

Based on https://github.com/linkerd/linkerd2/pull/2218/files#r254501947

adleong:
Fixes #2220 

The service profile validation which is part of `linkerd check` only validates service profiles in the Linkerd namespace.  Due to a recent change, service profiles now can exist in any namespace.

Update the logic so that service profiles in all namespaces are validated.  
Additionally:
* Generate a separate check for each service profile
* Relax validation of service profile names to support external names

```
linkerd-service-profile
-----------------------
√ can list service profiles
√ service profile webapp.default.svc.cluster.local in namespace alex is valid
‼ service profile authors.default.svc.cluster.local in namespace default is valid
    authors.default.svc.cluster.local: failed to validate ServiceProfile: error unmarshaling JSON: while decoding JSON: json: unknown field "pathRegexxxxxx"
    see https://linkerd.io/2/faq/#l5d-sp for hints
√ service profile linkerd.io in namespace default is valid
```

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/linkerd/linkerd2/controller/api/public"
-	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
 	spclient "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
 	healthcheckPb "github.com/linkerd/linkerd2/controller/gen/common/healthcheck"
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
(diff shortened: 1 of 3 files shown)


adleong:
I think it's super useful to list all the service profiles in the check output as a confirmation that they were created and validated successfully.  Otherwise, a successful check might leave you wondering, "were my profiles valid, or were they not found at all?"  As a user, I'd be tempted to introduce intentional errors to my profiles just to check that the validator is finding them.

But I'll defer to previous precedent for verbosity.

grampelberg:
@adleong you're totally right that it is useful to list all the profiles. Check doesn't feel like the right place for it right now though, I'd love to see a new command that gives really verbose feedback with explanations of how to fix it. That way, you can run check, see that something is wrong and go run the detailed debugging command. As a general rule, I like the idea of having check point to more detailed debugging commands.

PR #2239, opened 2019-02-09
Introduce golangci-lint tooling, fixes

siggy:
`golangci-lint` performs numerous checks on Go code, including golint,
ineffassign, govet, and gofmt.

This change modifies `bin/lint` to use `golangci-lint`, and replaces
usage of golint and govet.

Also perform a one-time gofmt cleanup:
- `gofmt -s -w controller/`
- `gofmt -s -w pkg/`

Part of #217

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,6 +1,5 @@
 required = [
   "github.com/golang/protobuf/protoc-gen-go",
-  "golang.org/x/lint/golint",
   "k8s.io/code-generator/cmd/client-gen",
   "k8s.io/code-generator/cmd/deepcopy-gen",
   "k8s.io/code-generator/cmd/defaulter-gen",
--- a/bin/lint
+++ b/bin/lint
@@ -2,6 +2,8 @@
 
 set -eu
 
+lintversion=1.14.0
+
 cd "$(pwd -P)"
 
 bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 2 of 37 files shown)


controller/proxy-injector/server.go
```diff
@@ -86,6 +86,9 @@ func tlsConfig(rootCA *pkgTls.CA, controllerNamespace string) (*tls.Config, erro
 	}
 	dnsName := tlsIdentity.ToDNSName()
 	certAndPrivateKey, err := rootCA.IssueEndEntityCertificate(dnsName)
+	if err != nil {
```
kleimkuhler: Oo good catch here.

Gopkg.toml
```diff
@@ -1,5 +1,6 @@
 required = [
   "github.com/golang/protobuf/protoc-gen-go",
+  "github.com/gordonklaus/ineffassign",
```
klingerf: I see that this project doesn't do releases, but mind pinning it to a fixed sha in this file?

bin/ineffassign
```diff
@@ -0,0 +1,15 @@
...
+
+bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rootdir="$( cd $bindir/.. && pwd )"
+
+cd "$rootdir"
+
+# install ineffassign from vendor
+go install ./vendor/github.com/gordonklaus/ineffassign
+
+ineffassign ./
```
klingerf: TIOLI This feels really similar to linting to me. How would you feel about updating the `go/lint` script to also run this tool, instead of adding a separate script? I can imagine that in almost all cases it would make sense to run them together.

siggy: Good point. I started down this path, then thought about what other tools we could combine here. I landed on [golangci-lint](https://github.com/golangci/golangci-lint/), which provides golint, ineffassign, govet, and a bunch of other linters, and it's fast. It will also let us incrementally enable linters as we go. Let me know what you think.

klingerf: Awesome! Yeah, using golangci-lint works for me. In fact, @radu-matei suggested that a while ago, here: https://github.com/linkerd/linkerd2/pull/2018#discussion_r243457453. I'm all for switching to that setup.

bin/lint
```diff
@@ -2,22 +2,32 @@
...
+if [ ! -f "$lintbin" ]; then
+  curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/v$lintversion/install.sh | sh -s -- -b . v$lintversion
+  mv ./golangci-lint${exe} $lintbin
 fi
 
-echo "all clean!"
+$lintbin run \
+  --enable gofmt,golint \
+  --disable deadcode,errcheck,gosimple,staticcheck,structcheck,unused,varcheck \
+  "$@"
```
klingerf: Am not sure exactly what's going on, but `bin/lint` no longer detects missing comments on publicly-exported structs and functions, which it previously caught.

I've modified my local branch as follows:

```diff
diff --git a/controller/proxy-injector/webhook.go b/controller/proxy-injector/webhook.go
index 1a25ff0d..5a55fae2 100644
--- a/controller/proxy-injector/webhook.go
+++ b/controller/proxy-injector/webhook.go
@@ -25,9 +25,6 @@ const (
 	envVarKeyProxyID                    = "LINKERD2_PROXY_ID"
 )
 
-// Webhook is a Kubernetes mutating admission webhook that mutates pods admission
-// requests by injecting sidecar container spec into the pod spec during pod
-// creation.
 type Webhook struct {
 	client              kubernetes.Interface
 	deserializer        runtime.Decoder
```

When I run `bin/lint`, I get:

```bash
$ bin/lint 
$ echo $?
0
```

Whereas if I run the old command, I get:

```
$ go list ./... | xargs golint
/Users/kl/workspace/go/src/github.com/linkerd/linkerd2/controller/proxy-injector/webhook.go:28:6: exported type Webhook should have comment or be unexported
```

I guess they're using different version of golint?

siggy: Oh wow nice catch. Looking.

PR #2242, opened 2019-02-09
Bump Prometheus to v2.7.1, Grafana to 5.4.3

siggy:


grampelberg:
@siggy awesome! Is there anything in those changelogs we should surface for the release?

siggy:
@grampelberg Oh hmm, I hadn't intended for these changes to be in 2.2, but if you'd like them included, I can dig into the changelogs.

PR #2245, opened 2019-02-10
Update check command hint URLs to new alias

siggy:
The existing hint URLs printing by `linkerd check` pointed to locations
that would change if the linkerd.io website was reorganized.

linkerd/website#148 introduces an alias for hint URLs at
https://linkerd.io/checks/. This is the corresponding change to update
`linkerd check` output.

Depends on linkerd/website#148, relates to linkerd/website#146.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -101,6 +101,11 @@ const (
 	LinkerdDataPlaneChecks CategoryID = "linkerd-data-plane"
 )
 
+// HintBaseURL is the base URL on the linkerd.io website that all check hints
+// point to. Each check adds its own `hintAnchor` to specify a location on the
+// page.
+const HintBaseURL = "https://linkerd.io/checks/#"
+
 var (
 	retryWindow    = 5 * time.Second
 	requestTimeout = 30 * time.Second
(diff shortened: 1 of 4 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -111,8 +116,9 @@ type checker struct {
 	// when the check is executed
 	description string
 
-	// hintURL provides a pointer to more information about the check
-	hintURL string
+	// hintAnchor, when appended to `HintBaseURL`, provides a URL to more
+	// information about the check
+	hintAnchor string
```
dadjeibaah: TIOLI: What are your thoughts on renaming this to `hintSection`, to me, the name `hintAnchor`contains HTML implementation detail which may not be necessary in this case.

rmars: hm, imo the name `hintAnchor` is very descriptive as to what it is, and it seems helpful for someone who is not familiar with the check code to know exactly how that piece fits in with the hint page

siggy: heh, ok i get both perspectives. i'm going to revert back to `hintAnchor`, as i do have a slight preference for communicating that this is intended as a URL anchor.

PR #2246, opened 2019-02-11
Fix auto-inject config when TLS is disabled

### LINKED ISSUE #2236: Automatic proxy injection still appears to require TLS
### What is the issue?
When using proxy auto-injection (`linkerd install --proxy-auto-inject`), liveness and readiness checks are failing.
### Logs, error output, etc
```
ERR! admin={bg=tls-config} linkerd2_proxy::transport::tls::config error loading /var/linkerd-io/identity/certificate.crt: No such file or directory (os error 2)
WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io("/var/linkerd-io/identity/certificate.crt", Some(2)), falling back
ERR! proxy={server=in listen=0.0.0.0:4143 remote=100.96.64.1:47810} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:9200)
```
#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ controller pod is running
√ can initialize the client

linkerd-api
-----------

linkerd-service-profile
-----------------------
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √
```

### Environment

- Kubernetes Version: 1.10.11
- Cluster Environment: kops
- Linkerd version: 19.2.1

### Additional context
The failing health check:
```
httpGet:
  path: /_cluster/health
  port: 9200
  scheme: HTTP
```

klingerf:
As of #2163, TLS is no longer required when installing linkerd with automatic proxy injection, but the injected linkerd configs were still referencing TLS volumes. This change updates the proxy-injector to only include those volumes when the control plane is installed with auto-inject and TLS enabled.

Fixes #2236.


controller/proxy-injector/fake/data/inject-no-init-container-request.json
```diff
@@ -1,94 +0,0 @@
-{
```
klingerf: This text fixture file was not being used anywhere, so I'm removing it.

controller/proxy-injector/webhook_config.go
```diff
@@ -23,11 +23,10 @@ type WebhookConfig struct {
 	trustAnchor         []byte
 	configTemplate      *template.Template
 	k8sAPI              kubernetes.Interface
-	noInitContainer     bool
 }
 
 // NewWebhookConfig returns a new instance of initiator.
-func NewWebhookConfig(client kubernetes.Interface, controllerNamespace, webhookServiceName string, noInitContainer bool, rootCA *tls.CA) (*WebhookConfig, error) {
+func NewWebhookConfig(client kubernetes.Interface, controllerNamespace, webhookServiceName string, rootCA *tls.CA) (*WebhookConfig, error) {
```
klingerf: The webhook config template does not use the value of `noInitContainer` when rendering the template, so it doesn't need to be passed into this func or set on the `WebhookConfig` struct.

Issue #1999, opened 2018-12-19
Fetch `inject` configuration from control plane


grampelberg:
### What problem are you trying to solve?

As the configuration for `install` continues to grow, it is becoming difficult to know what flags to set in `inject` to match the control plane features. As an example, see #1887 and #1891 .

### How should the problem be solved?

Update the `inject` command to pull default configuration from the control plane before actually doing the inject.

### Any alternatives you've considered?

- A local configuration file (see #1892 and #1604). This will work for the single situation, but not for multiple users of the same control plane.

klingerf:
@olix0r is working on this in  #2434, but the implementation will change once the config endpoint has been added (#2455). Assigning to him.

alpeb:
Closed via #2535

PR #2249, opened 2019-02-11
Honour install's `proxy-log-level` flag when autoinjecting proxies

### LINKED ISSUE #2248: Proxy log level hard-coded when auto-injecting
`LINKERD2_PROXY_LOG` is hard-coded to `warn,linkerd2_proxy=info` here:

https://github.com/linkerd/linkerd2/blob/f6e75ec83a7c15e423ce21894f49841924da2987/chart/templates/proxy_injector.yaml#L162

It should take the value passed as a parameter in `linkerd install`.

alpeb:
Fixes #2248

It was hard-coded to `warn-linkerd2_proxy=info`.
Still, something that will have to be handled more generically in #1999 

Also still to do in a separate PR would be the syntax validation of this flag value.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -56,6 +56,7 @@ type installConfig struct {
 	ProxyAutoInjectEnabled           bool
 	ProxyInjectAnnotation            string
 	ProxyInjectDisabled              string
+	ProxyLogLevel                    string
 	ProxyUID                         int64
 	ProxyMetricsPort                 uint
 	ProxyControlPort                 uint
@@ -210,6 +211,7 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
 		ProxyAutoInjectEnabled:           options.proxyAutoInject,
 		ProxyInjectAnnotation:            k8s.ProxyInjectAnnotation,
 		ProxyInjectDisabled:              k8s.ProxyInjectDisabled,
+		ProxyLogLevel:                    options.proxyLogLevel,
 		ProxyUID:                         options.proxyUID,
 		ProxyMetricsPort:                 options.proxyMetricsPort,
 		ProxyControlPort:                 options.proxyControlPort,
--- a/chart/templates/proxy_injector.yaml
+++ b/chart/templates/proxy_injector.yaml
@@ -159,7 +159,7 @@ data:
   {{.Values.ProxySpecFileName}}: |
     env:
     - name: LINKERD2_PROXY_LOG
-      value: warn,linkerd2_proxy=info
+      value: {{.Values.ProxyLogLevel}}
     - name: LINKERD2_PROXY_CONTROL_URL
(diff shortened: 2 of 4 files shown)


cli/cmd/install.go
```diff
@@ -56,6 +56,7 @@ type installConfig struct {
 	ProxyAutoInjectEnabled           bool
 	ProxyInjectAnnotation            string
 	ProxyInjectDisabled              string
+	ProxyLogLevel                    string
```
siggy: since `installConfig` already contains a `proxyConfigOptions`, which itself contains a `proxyLogLevel`, is adding the field directly to `installConfig` necessary?

alpeb: Unless I'm missing something, `installConfig` doesn't contain `proxyConfigOptions`. That's `installOptions`

siggy: Ah you're right! I misread this.

kleimkuhler:
> Also still to do in a separate PR would be the syntax validation of this flag value.

Good point here. Something I'm not finding easily is: what is valid syntax for `proxy-log-level`?

`func (options *installOptions) validate` could just have another check like this:
```
if _, err := log.ParseLevel(options.proxyLogLevel); err != nil {
    return fmt.Errorf("--proxy-log-level must be one of: panic, fatal, error, warn, info, debug")
}
```
but those values are not valid according to the default.

alpeb:
@kleimkuhler yeah, maybe @hawkw can explain the syntax (why two levels?) and list the valid values :wink: 
In the meantime, I'll merge this.

hawkw:
@alpeb and @kleimkuhler:

I think it would definitely be possible to validate proxy log levels on the CLI, but I don't think "one of: panic, fatal, error, warn, info, debug" is sufficient to accept all valid log levels. 

A quick breakdown of the syntax for setting proxy log levels: 
+ `PROXY_LOG_LEVEL` is a comma-separated list of _log directives_
+ A *log directive_ consists of either:
  - a *level* (e.g. `info`), which sets the global log level, or 
  - a *module path* (e.g. `foo` or `foo::bar::baz`),
  - a *module path*  followed by an equals sign and a *level* (e.g. `foo=warn` or `foo::bar::baz=debug`), which sets the log level for that module
+ A *level* is one of:
  - trace
  - debug
  - info
  - warn
  - error
+ A *module path* represents the path to a Rust module. it consists of one or more *module name*s, separated by `::`
+ A *module name* starts with a letter, and consists of alphanumeric characters and `_`

We can't easily have a list of valid module paths, since there's potentially a module path for every source code file in the proxy codebase or any of its dependencies. However, we _can_ validate that they're syntactically correct.

We could perhaps test that the log level matches a regex like 
```
(trace|debug|warn|info|error)|(\w|::)+|((\w|::)+=(trace|debug|warn|info|error)?)(?:,|$)*
```
(apologies for my mediocre regex skills)

For more information on the log level format, see https://docs.rs/env_logger/0.6.0/env_logger/#enabling-logging

PR #2256, opened 2019-02-11
Add validator for proxy log-level

alpeb:
Follow-up to #2249

### DIFF
--- a/cli/cmd/install_test.go
+++ b/cli/cmd/install_test.go
@@ -174,17 +174,48 @@ func TestValidate(t *testing.T) {
 		}
 	})
 
-	t.Run("Rejects invalid log level", func(t *testing.T) {
+	t.Run("Rejects invalid controller log level", func(t *testing.T) {
 		options := newInstallOptions()
 		options.controllerLogLevel = "super"
 		expected := "--controller-log-level must be one of: panic, fatal, error, warn, info, debug"
 
 		err := options.validate()
 		if err == nil {
-			t.Fatalf("Expected error, got nothing")
+			t.Fatal("Expected error, got nothing")
 		}
 		if err.Error() != expected {
 			t.Fatalf("Expected error string\"%s\", got \"%s\"", expected, err)
+		}
+	})
+
+	t.Run("Properly validates proxy log level", func(t *testing.T) {
+		testCases := []struct {
+			input string
+			valid bool
+		}{
+			{"", false},
+			{"info", true},
+			{"somemodule", true},
+			{"linkerd2_proxy=debug", true},
+			{"linkerd2_proxy=foobar", false},
+			{"linker2d_proxy,std::option", true},
+			{"warn,linkerd2_proxy=info", true},
+			{"warn,linkerd2_proxy=foobar", false},
+		}
+
+		options := newInstallOptions()
+		for _, tc := range testCases {
+			options.proxyLogLevel = tc.input
+			err := options.validate()
+			if tc.valid && err != nil {
+				t.Fatalf("Error not expected: %s", err)
+			}
+			if !tc.valid && err == nil {
+				t.Fatalf("Expected error string \"%s is not a valid proxy log level\", got nothing", tc.input)
+			}
+			if !tc.valid && err.Error() != fmt.Sprintf("%s is not a valid proxy log level", tc.input) {
+				t.Fatalf("Expected error string \"%s is not a valid proxy log level\", got \"%s\"", tc.input, err)
+			}
 		}
 	})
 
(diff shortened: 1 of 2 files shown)


cli/cmd/root.go
```diff
@@ -43,8 +43,20 @@ var (
 	// These regexs are not as strict as they could be, but are a quick and dirty
 	// sanity check against illegal characters.
 	alphaNumDash              = regexp.MustCompile("^[a-zA-Z0-9-]+$")
-	alphaNumDashDot           = regexp.MustCompile("^[\\.a-zA-Z0-9-]+$")
-	alphaNumDashDotSlashColon = regexp.MustCompile("^[\\./a-zA-Z0-9-:]+$")
+	alphaNumDashDot           = regexp.MustCompile(`^[\.a-zA-Z0-9-]+$`)
+	alphaNumDashDotSlashColon = regexp.MustCompile(`^[\./a-zA-Z0-9-:]+$`)
+
+	// Full Rust log level syntax at
+	// https://docs.rs/env_logger/0.6.0/env_logger/#enabling-logging
```
siggy: 👍

cli/cmd/root.go
```diff
@@ -43,8 +43,20 @@ var (
 	// These regexs are not as strict as they could be, but are a quick and dirty
 	// sanity check against illegal characters.
 	alphaNumDash              = regexp.MustCompile("^[a-zA-Z0-9-]+$")
-	alphaNumDashDot           = regexp.MustCompile("^[\\.a-zA-Z0-9-]+$")
-	alphaNumDashDotSlashColon = regexp.MustCompile("^[\\./a-zA-Z0-9-:]+$")
+	alphaNumDashDot           = regexp.MustCompile(`^[\.a-zA-Z0-9-]+$`)
+	alphaNumDashDotSlashColon = regexp.MustCompile(`^[\./a-zA-Z0-9-:]+$`)
```
siggy: curious: why the backticks on `alphaNumDashDot` and `alphaNumDashDotSlashColon` but not on `alphaNumDash`?

cli/cmd/install_test.go
```diff
@@ -174,20 +174,51 @@ func TestValidate(t *testing.T) {
...
 		if err == nil {
-			t.Fatalf("Expected error, got nothing")
+			t.Fatal("Expected error, got nothing")
 		}
 		if err.Error() != expected {
 			t.Fatalf("Expected error string\"%s\", got \"%s\"", expected, err)
 		}
 	})
 
+	t.Run("Properly validates proxy log level", func(t *testing.T) {
```
hawkw: TIOLI: maybe the test name should be more consistent with the controller log level test? Like
```suggestion
	t.Run("Rejects invalid proxy log level", func(t *testing.T) {
```

cli/cmd/install_test.go
```diff
@@ -174,20 +174,51 @@ func TestValidate(t *testing.T) {
...
+			valid bool
+		}{
+			{"", false},
+			{"info", true},
+			{"somemodule", true},
+			{"linkerd2_proxy=debug", true},
+			{"linkerd2_proxy=foobar", false},
+			{"linker2d_proxy,std::option", true},
+			{"warn,linkerd2_proxy=info", true},
+			{"warn,linkerd2_proxy=foobar", false},
```
hawkw: It would be nice to see some test cases in here with characters that are never valid in a proxy log level string.

cli/cmd/root.go
```diff
@@ -315,6 +327,10 @@ func (options *proxyConfigOptions) validate() error {
 		return fmt.Errorf("--tls must be blank or set to \"%s\"", optionalTLS)
 	}
 
+	if !validProxyLogLevel.MatchString(options.proxyLogLevel) {
+		return fmt.Errorf("%s is not a valid proxy log level", options.proxyLogLevel)
```
hawkw: Perhaps this error should provide a link to https://docs.rs/env_logger/0.6.0/env_logger/#enabling-logging as a reference?

cli/cmd/root.go
alpeb: @siggy backticks relieve us from having to escape bakslashes, so we don't end up with `\\`, which we didn't have in `alphaNumDash`. But perhaps as a best practice we should always use backticks in all regexes.

cli/cmd/install_test.go
alpeb: I called it like that because it no only rejects invalid values, but it also tests for valid values :grin:

cli/cmd/install_test.go
alpeb: will do :+1:

PR #2257, opened 2019-02-11
Remove auto proxy inject 'Mutate' function tests

### LINKED ISSUE #2201: proxy injector webhook tests are too fragile
Every time the proxy's generated yaml changes, controller/proxy-injector/webhook_test.go tests fail, because these tests rely on hardcoded base64 values describing the pod manifest.

These test failures are extremely hard to diagnose without knowing how this all hangs together.

We should figure out a testing strategy that is more flexible to unrelated changes.

At a worst case, we may want to remove these tests if they prove too costly to maintain.

ihcsim:
The way these tests compare the hard-coded base64-encoded JSON patches with those generated by the proxy injector, is extremely brittle. Changing any of the proxy configuration causes these tests to break, even though the proxy injector itself isn't affected.

Also, the AdmissionRequest and AdmissionResponse types are "boundary objects" that are largely irrelevant to our code.

Fixes #2201 


ihcsim:
> Will this PR have new `Mutate` webhook tests

I've been debating about this. At this point, I think I won't add any more tests to the webhook simply because any additional tests will just end up doing the same thing i.e. comparing hard-coded strings. The webhook doesn't care about the proxy specs and configs; it just treats everything as YAML (for request body) and JSON (for response body). It doesn't even validate the response JSON patches; it simply sends the patches back to the k8s api server, and let it re-assemble the resource manifest.

klingerf:
Thanks for tackling this! I agree that maintaining these base64 strings is untenable. I haven't given this too much thought, but how much work do you think it would be to have the tests validate the responses automatically using the CLI code that generates injected YAML?

We could still remove the base64 response fixtures, fixing #2201, but we'd be able to assert that the patch matches what we'd expect to get from running `linkerd inject` manually. That would have the added benefit of guaranteeing that our two inject code paths are in sync, and it would have helped to detect issues like the one #2267.

Anyway, if it's an enormous amount of work, we can just remove the tests outright, but it seems like an opportunity to increase our coverage.

ihcsim:
@klingerf IIUC, you're talking about validating the [webhook inject output](https://github.com/linkerd/linkerd2/blob/1cde1f083c5f245170f75e1974f3b20f7fa34d9c/controller/proxy-injector/webhook.go#L104) against the output generated by [these CLI code](https://github.com/linkerd/linkerd2/blob/1cde1f083c5f245170f75e1974f3b20f7fa34d9c/cli/cmd/inject.go#L391), right? I feel then we're getting into #1748, because I'd think we need to extract the CLI inject code into `pkg`, before we can use it to validate the webhook responses.

controller/proxy-injector/webhook_test.go
```diff
@@ -15,62 +15,6 @@ var (
...
-		webhook      *Webhook
-		title        string
-		requestFile  string
-		responseFile string
-	}{
-		{defaultWebhook, "no labels", "inject-empty-request.json", "inject-empty-response.yaml"},
-		{defaultWebhook, "inject enabled", "inject-enabled-request.json", "inject-enabled-response.yaml"},
-		{defaultWebhook, "inject disabled", "inject-disabled-request.json", "inject-disabled-response.yaml"},
-		{noInitContainerWebhook, "inject no-init-container", "inject-enabled-request.json", "inject-no-init-container-response.yaml"},
-		{tlsDisabledWebook, "inject without tls", "inject-enabled-request.json", "inject-enabled-tls-disabled-response.yaml"},
```
klingerf: You can also remove the inject-enabled-tls-disabled-response.yaml fixture file as part of this branch (sorry, I snuck it in as part of #2246).

Issue #2223, opened 2019-02-07
Notify of TCP traffic in the CLI/GUI


grampelberg:
### What problem are you trying to solve?

When a service is using either a binary protocol or HTTPS (and terminating it), there are no metrics. This can be pretty confusing in the CLI/GUI as the stats just don't show up. In particular, HTTPS comes up as a common question.

### How should the problem be solved?

Replace the `meshed` column with a `protocol detected` (name needs work) column. This way, you'll know that we're seeing traffic, just that it can't be parsed and there won't be metrics for it.

grampelberg:
@siggy what do you think about adding something like this as a check? Feels wrong to do a warning or error, but it also is something nice to have. Perhaps this suggests a different command or something that shows TCP metrics...

grampelberg:
Related to #2224.

siggy:
@grampelberg i definitely agree just showing no stats isn't a good user experience. perhaps a modification to the `stat` command, or a separate command could help with that, maybe even with some hints if we know it's only tcp metrics.

rmars:
@grampelberg TCP stats are now available in the CLI and in several dashboard pages. Is this ticket still something we should do? If there's TCP only stats, the http-based tables do still look bare, but you can scroll down and still see some metrics. We could possibly add a TCP connections column to all the metrics tables so that they don't look so weird. We could alternatively add a note that says something like "No http metrics detected but TCP metrics present, please scroll down"

grampelberg:
@rmars the core problem is that folks go looking for stats, let's say for an HTTPS service, and are confused/surprised that nothing is there. For the GUI, a little notification that says there are TCP metrics with a link would be great (whether there are HTTP one or not, it is nice to teach users about other functionality in the platform).

For the CLI, it is a little bit tougher. Perhaps adding a column to `stat` that says there's *something* and a clear command for the CLI to run?

rmars:
@grampelberg For the CLI additional column, how about a simple `TCP detected` that just says y/n? That would at least give some indication when there are no http stats present. The criterion could be tcp open connections > 0.

grampelberg:
Maybe just show the open connections count?

siggy:
@rmars Is this good to close?

rmars:
I think enough has been completed for the UIs to be less confusing when there's TCP traffic and no HTTP traffic. There's still more work that could be done in showing TCP traffic, but we should open a more specific issue when we decide what that work would entail.

PR #2262, opened 2019-02-12
Modify StatSummary to also return TCP stats

rmars:
This uses the existing statsummary query infrastructure to handle things like inbound/outbound, and request labels.

The following api query will produce these prometheus queries:
`http://localhost:8084/api/tps-reports?resource_type=deployment&namespace=linkerd&resource_name=linkerd-controller&tcp_stats=true`
```
sum(increase(tcp_open_total{deployment="linkerd-controller", direction="inbound", namespace="linkerd"}[1m])) by (namespace, deployment)
sum(increase(tcp_read_bytes_total{deployment="linkerd-controller", direction="inbound", namespace="linkerd"}[1m])) by (namespace, deployment)
sum(increase(response_total{deployment="linkerd-controller", direction="inbound", namespace="linkerd"}[1m])) by (namespace, deployment, classification, tls)
```

For now, I added a table at the bottom of the ResourceDetail page, but we could better incorporate this into different views.

screen shot 2019-02-11 at 5 31 40 pm

Part of #2223


controller/api/public/stat_summary.go
```diff
@@ -184,6 +188,14 @@ func (s *grpcServer) k8sResourceQuery(ctx context.Context, req *pb.StatSummaryRe
 		if !ok {
 			continue
 		}
+
+		var tcpStats *pb.TcpStats
+		if !req.TcpStats {
+			tcpStats = nil
+		} else {
+			tcpStats = tcpMetrics[key]
```
siggy: default value for a pointer is `nil`, so this should work:
```go
var tcpStats *pb.TcpStats
if req.TcpStats {
	tcpStats = tcpMetrics[key]
}
```

ihcsim:
Super excited about this PR! The code change LGTM (based on the little that I know about the `public` and `web`). A few questions:

1. Is the plan to add the TCP tables to other views in this PR?
1. The latency columns of pods on the Overview page show `---`, which I think makes more sense than the `0s`, since we aren't picking up any latency metrics from Prometheus, right?
1. Do we need to add these TCP metrics to the grafana dashboard too?

rmars:
1. I think this PR will serve to introduce the proto api and wire up the quering in StatSummary. But I'll add TCP stats to other views in a separate PR (didn't want to make this too big!).
2. Yeah, the Overview page latencies aren't tcp ones? (I'm not quite sure what you're asking)
3. These metrics should already be in the grafana dashboards:
```
grafana/dashboards/daemonset.json
709:              "expr": "tcp_close_total{namespace=\"$namespace\", daemonset=\"$daemonset\", direction=\"inbound\",errno!=\"\"}",
794:              "expr": "tcp_open_connections{namespace=\"$namespace\", daemonset=\"$daemonset\", direction=\"inbound\"}",
870:              "expr": "tcp_connection_duration_ms_bucket{namespace=\"$namespace\", daemonset=\"$daemonset\", direction=\"inbound\"}",
1574:              "expr": "tcp_close_total{namespace=\"$namespace\", daemonset=\"$daemonset\", direction=\"outbound\",errno!=\"\"}",
1659:              "expr": "tcp_open_connections{namespace=\"$namespace\", daemonset=\"$daemonset\", direction=\"outbound\"}",
1735:              "expr": "tcp_connection_duration_ms_bucket{namespace=\"$namespace\", daemonset=\"$daemonset\", direction=\"outbound\"}",

grafana/dashboards/deployment.json
709:              "expr": "tcp_close_total{namespace=\"$namespace\", deployment=\"$deployment\", direction=\"inbound\",errno!=\"\"}",
794:              "expr": "tcp_open_connections{namespace=\"$namespace\", deployment=\"$deployment\", direction=\"inbound\"}",
870:              "expr": "tcp_connection_duration_ms_bucket{namespace=\"$namespace\", deployment=\"$deployment\", direction=\"inbound\"}",
1574:              "expr": "tcp_close_total{namespace=\"$namespace\", deployment=\"$deployment\", direction=\"outbound\",errno!=\"\"}",
1659:              "expr": "tcp_open_connections{namespace=\"$namespace\", deployment=\"$deployment\", direction=\"outbound\"}",
1735:              "expr": "tcp_connection_duration_ms_bucket{namespace=\"$namespace\", deployment=\"$deployment\", direction=\"outbound\"}",

grafana/dashboards/health.json
697:              "expr": "tcp_close_total{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\",errno!=\"\"}",
781:              "expr": "tcp_open_connections{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\"}",
857:              "expr": "tcp_connection_duration_ms_bucket{deployment=\"$deployment\", namespace=\"$namespace\", direction=\"inbound\"}",

grafana/dashboards/statefulset.json
709:              "expr": "tcp_close_total{namespace=\"$namespace\", statefulset=\"$statefulset\", direction=\"inbound\",errno!=\"\"}",
794:              "expr": "tcp_open_connections{namespace=\"$namespace\", statefulset=\"$statefulset\", direction=\"inbound\"}",
870:              "expr": "tcp_connection_duration_ms_bucket{namespace=\"$namespace\", statefulset=\"$statefulset\", direction=\"inbound\"}",
1574:              "expr": "tcp_close_total{namespace=\"$namespace\", statefulset=\"$statefulset\", direction=\"outbound\",errno!=\"\"}",
1659:              "expr": "tcp_open_connections{namespace=\"$namespace\", statefulset=\"$statefulset\", direction=\"outbound\"}",
1735:              "expr": "tcp_connection_duration_ms_bucket{namespace=\"$namespace\", statefulset=\"$statefulset\", direction=\"outbound\"}",
```

proto/public.proto
```diff
@@ -339,6 +340,15 @@ message BasicStats {
   uint64 actual_failure_count = 8;
 }
 
+message TcpStats {
+  // number of currently open connections
+  uint64 open_connections = 1;
```
siggy: thoughts around including these tcp stats in `BasicStats`?

taking that a step further, what about always including tcp stats in stats queries? pros and cons that come to mind:
- always include
  - simpler code
  - aligns with flatter telemetry model in proxies/prometheus/grafana
- require `tcp_stats=true`
  - potentially less pressure on prometheus?
  - allows for more/different types of stats going forward?
  - maybe tcp stats aren't a thing for some stats queries?

ihcsim:
> 2. the Overview page latencies aren't tcp ones

On the ResourceDetail page, the latency shows up as 0s. If there are no tcp latency data, then I think they should be `---`.

image

> 3. These metrics should already be in the grafana dashboards

On my end, the grafana dashboard isn't picking up the statefulset and its pods. I think like @klingerf said last week, the labels (under Settings > Variables) need to be updated.

web/app/js/components/MetricsTable.jsx
```diff
@@ -14,7 +14,78 @@ import _isNil from 'lodash/isNil';
...
+const TcpStatColumns = [
+  {
+    title: "Connections",
+    dataIndex: "tcp.openConnections",
+    isNumeric: true,
+    render: d => metricToFormatter["NO_UNIT"](d.tcp.openConnections),
+    sorter: (a, b) => numericSort(a.tcp.openConnections, b.tcp.openConnections)
+  },
+  {
+    title: "Read Bytes",
```
siggy: consider something to indicate this is bytes over the last time window? `Read Bytes/min`, assuming our time window is hard-coded?

proto/public.proto
rmars: Yeah totally, I did think about that! So currently, I'm adding 3 new prometheus queries (and when we add duration it will be 4) which is doubling the number of queries sent per stat request. I wasn't sure whether that was worth it, unless we were committed to always displaying TCP stats everywhere (since our current views don't use them, it seemed like a lot to double the queries for every case, when we don't surface the stats in many places). However, I could also see us incorporating TCP stats into normal stat views, where I agree it could make sense to always have them.

web/app/js/components/MetricsTable.jsx
rmars: yeah, good point!

rmars:
@ihcsim ahhh yeah I see what you mean, on the redis statefulsets page! Looking into it now!

rmars:
@ihcsim 90542f0 should fix the issue! We were returning an empty pb.BasicStats{} struct even if there were no stats, which the proto to json renderer converts to "0" default values. Setting this to `nil` instead will cause the json to render `null` and not `"0"`

web/app/js/components/MetricsTable.jsx
rmars: (we could also consider doing Read Bytes / sec and just dividing by the metric window)

proto/public.proto
siggy: Test results show no significant increase in resource usage when always querying for TCP metrics. The only variation between baseline and the "always TCP" case is an increase in queries to Prometheus, which is expected.

# Results

## Prometheus mem/cpu/latency

### Baseline

baseline prometheus

### Always tcp

always-tcp prometheus

## Prometheus linkerd-proxy stats

### Baseline

baseline prom-proxy-metrics

### Always tcp

always-tcp prom-proxy-metrics

# Test Setup

- Linkerd built off of this branch
- deployed to 2 namespaces, `l5d-baseline`, and `l5d-always-tcp`
- 2 emojivoto namespaces created, one for each linkerd. each had 10 replicas per-deployment, so each Linkerd was managing 40 injected pods
- 15-node AKS cluster, 4 vcpus, 14 GB memory per node
- 15 hour run, followed by evaluation of metrics in Prometheus

## Scripts

These scripts were run from a Kubernetes node in a tight bash `while... do ` loop.

### Baseline

```bash
#!/bin/bash

curl -s -o /dev/null "http://10.0.137.105:8084/api/tps-reports?resource_type=all&all_namespaces=true&window=1m"
curl -s -o /dev/null "http://10.0.137.105:8084/api/tps-reports?resource_type=namespace&all_namespaces=true&window=1m"
curl -s -o /dev/null "http://10.0.137.105:8084/api/tps-reports?resource_type=pod&namespace=emojivoto-baseline&tcp_stats=true&window=1m"
curl -s -o /dev/null "http://10.0.137.105:8084/api/tps-reports?resource_type=deployment&namespace=emojivoto-baseline&resource_name=emoji&window=1m"
curl -s -o /dev/null "http://10.0.137.105:8084/api/tps-reports?resource_type=all&all_namespaces=true&to_name=emoji&to_type=deployment&to_namespace=emojivoto-baseline&window=1m"
curl -s -o /dev/null "http://10.0.137.105:8084/api/tps-reports?resource_type=all&all_namespaces=true&from_name=emoji&from_type=deployment&from_namespace=emojivoto-baseline&window=1m"
```

### Always tcp

```
#!/bin/bash

curl -s -o /dev/null "http://10.0.122.86:8084/api/tps-reports?resource_type=all&all_namespaces=true&window=1m&tcp_stats=true"
curl -s -o /dev/null "http://10.0.122.86:8084/api/tps-reports?resource_type=namespace&all_namespaces=true&window=1m&tcp_stats=true"
curl -s -o /dev/null "http://10.0.122.86:8084/api/tps-reports?resource_type=pod&namespace=emojivoto-always-tcp&tcp_stats=true&window=1m"
curl -s -o /dev/null "http://10.0.122.86:8084/api/tps-reports?resource_type=deployment&namespace=emojivoto-always-tcp&resource_name=emoji&window=1m&tcp_stats=true"
curl -s -o /dev/null "http://10.0.122.86:8084/api/tps-reports?resource_type=all&all_namespaces=true&to_name=emoji&to_type=deployment&to_namespace=emojivoto-always-tcp&window=1m&tcp_stats=true"
curl -s -o /dev/null "http://10.0.122.86:8084/api/tps-reports?resource_type=all&all_namespaces=true&from_name=emoji&from_type=deployment&from_namespace=emojivoto-always-tcp&window=1m&tcp_stats=true"
```

rmars: based on this info, I'm going to move TcpStats into BasicStats, and always query for TCP stats on stat queries

rmars:
Okay so after thinking about this some more I think I'm back to thinking about the two things @siggy mentioned in https://github.com/linkerd/linkerd2/pull/2262#discussion_r256152913

1. Should we always query for TCP stats on every stat request?
2. Should TcpStats be integrated into the BasicStats struct?

### Pros/Cons for querying for tcp stats on every stat request
Pros: 
- simplifies logic, we won't need to check for `TcpStats: true` in stat requests
- Doubling the number of Prometheus queries doesn't seem to affect stat query performance in the tests we've done

Cons:
- feels icky to query for a bunch of data that we show in very few places
- in very large deployments in the future, with multiple users viewing the linkerd dashboard, this could be a problem

### Pros/Cons for integrating TcpStats into BasicStats
Pros:
- if we are always requesting tcp stats, it seems like they belong in "BasicStats"
- Simpler structure of returned data

Cons:
- If a resource is tcp-only and doesn't send HTTP traffic, we'll encounter the problem fixed in [90542f0](https://github.com/linkerd/linkerd2/commit/90542f07fd9ba2a5e6a7e7a03898cda22d238a69), which is that we can't differentiate between zero and null values for TcpStats. Similarly, if we don't have TcpStats for an object it would be better to return a nil TcpStat
- The current BasicStats object would better be named HttpStats as it's closer to that. So putting them together would be like mashing together different types of stats

rmars:
My current thinking is to leave the implementation as is and stick with non-default tcp stat fetching, and a separate struct, and revisit later, if necessary. cc/ @siggy @klingerf

web/app/js/components/MetricsTable.jsx
```diff
@@ -37,97 +108,60 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink) => {
...
+    },
+    sorter: (a, b) => (getResourceDisplayName(a) || "").localeCompare(getResourceDisplayName(b))
+  };
+
+  let columns = [nameColumn];
+  if (isTcpTable) {
+    columns = columns.concat(TcpStatColumns);
+  } else {
+    columns = columns.concat(httpStatColumns);
+    columns = columns.concat(grafanaColumn);
```
siggy: i just realized while testing this that we don't expose tcp-level stats on the `Linkerd Pod` Grafana dashboard. if/when we add tcp metrics to that dashboard, perhaps we should add the Grafana column when `isTcpTable == true`, what do you think?

klingerf:
@rmars Apologies for the delay in responding. That all sounds good to me. I prefer to keep TCP and HTTP stats in separate structs, and I'm good with only populating TCP stats on request.

web/app/js/components/MetricsTable.jsx
rmars: true! filed https://github.com/linkerd/linkerd2/issues/2329

rmars:
@siggy great spot on the Pod view; That was a bug, I've fixed it!

web/app/js/components/ResourceDetail.jsx
```diff
@@ -154,15 +154,24 @@ export class ResourceDetailBase extends React.Component {
...
-          }
+        let resourceName = resource.namespace + "/" + resource.name;
+        let podMetricsForResource;
+
+        if (resource.type === "pod") {
+          podMetricsForResource = _filter(podMetrics, pod => pod.name === resource.name);
+        } else {
+          let podBelongsToResource = _reduce(podListRsp.pods, (mem, pod) => {
+            if (_get(pod, resourceTypeToCamelCase(resource.type)) === resourceName) {
+              mem[pod.name] = true;
```
siggy: realizing this code was already in place, but would you mind adding a comment about the structure of `pod.name`? something like...
```diff
diff --git a/web/app/js/components/ResourceDetail.jsx b/web/app/js/components/ResourceDetail.jsx
index 6744dbdd..cccfc054 100644
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -157,21 +157,24 @@ export class ResourceDetailBase extends React.Component {
         let resourceName = resource.namespace + "/" + resource.name;
         let podMetricsForResource;

+        // note pod.name in podMetrics != pod.name in podListRsp
         if (resource.type === "pod") {
+          // pod.name == emoji-85977956cf-5v46h
           podMetricsForResource = _filter(podMetrics, pod => pod.name === resource.name);
         } else {
           let podBelongsToResource = _reduce(podListRsp.pods, (mem, pod) => {
             if (_get(pod, resourceTypeToCamelCase(resource.type)) === resourceName) {
+              // pod.name == emojivoto/emoji-85977956cf-5v46h
               mem[pod.name] = true;
             }

             return mem;
           }, {});

+          // pod.name == emoji-85977956cf-5v46h
           podMetricsForResource = _filter(podMetrics, pod => podBelongsToResource[pod.namespace + "/" + pod.name]);
         }

-
         let resourceIsMeshed = true;
         if (!_isEmpty(this.state.resourceMetrics)) {
           resourceIsMeshed = _get(this.state.resourceMetrics, '[0].pods.meshedPods') > 0;
```

web/srv/api_handlers.go
```diff
@@ -108,6 +108,11 @@ func (h *handler) handleAPIStat(w http.ResponseWriter, req *http.Request, p http
 	if req.FormValue("skip_stats") == "true" {
 		skipStats = true
 	}
+	tcpStats := false
+	if req.FormValue("tcp_stats") == "true" {
+		tcpStats = true
+	}
```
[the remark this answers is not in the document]
rmars: omg, so much simpler!

web/app/js/components/MetricsTable.jsx
```diff
@@ -14,7 +14,78 @@ import _isNil from 'lodash/isNil';
 import { processedMetricsPropType } from './util/MetricUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 
-const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink) => {
+const TcpStatColumns = [
```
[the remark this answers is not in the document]
rmars: omg whoops yes, the capital T was unintentional

controller/api/public/prometheus.go
```diff
@@ -135,10 +138,15 @@ func (s *grpcServer) getPrometheusMetrics(ctx context.Context, requestQueryTempl
...
+			query = fmt.Sprintf(requestQueryTemplate, labels, groupBy)
+		} else {
+			query = fmt.Sprintf(requestQueryTemplate, labels, timeWindow, groupBy)
+		}
+
 		go func(typ promType, template string) {
-			// success/failure counts
-			requestsQuery := fmt.Sprintf(template, labels, timeWindow, groupBy)
-			resultVector, err := s.queryProm(ctx, requestsQuery)
+			resultVector, err := s.queryProm(ctx, query)
```
klingerf: I'm pretty sure we want to continue to build the `query` variable inside this goroutine, or otherwise we risk running into a closure issue [like the one that's described here](https://blog.cloudflare.com/a-go-gotcha-when-closures-and-goroutines-collide/). Can we do something like this:

```diff
diff --git a/controller/api/public/prometheus.go b/controller/api/public/prometheus.go
index 884e49ca..a1143307 100644
--- a/controller/api/public/prometheus.go
+++ b/controller/api/public/prometheus.go
@@ -138,14 +138,13 @@ func (s *grpcServer) getPrometheusMetrics(ctx context.Context, requestQueryTempl
 
 	// kick off asynchronous queries: request count queries + 3 latency queries
 	for pt, requestQueryTemplate := range requestQueryTemplates {
-		var query string
-		if pt == promTCPConnections {
-			query = fmt.Sprintf(requestQueryTemplate, labels, groupBy)
-		} else {
-			query = fmt.Sprintf(requestQueryTemplate, labels, timeWindow, groupBy)
-		}
-
 		go func(typ promType, template string) {
+			var query string
+			if typ == promTCPConnections {
+				query = fmt.Sprintf(template, labels, groupBy)
+			} else {
+				query = fmt.Sprintf(template, labels, timeWindow, groupBy)
+			}
 			resultVector, err := s.queryProm(ctx, query)
 
 			resultChan <- promResult{
```

rmars: great catch! as talked about in person, I've passed in the query string to the goroutine instead!

PR #2272, opened 2019-02-12
Modify the grafana variable queries to use a tcp-based metric

rmars:
Currently, we use request_total for the variable query to determine the names in the grafana dropdowns. We should use a tcp based metric instead, so that non-http traffic will also count towards something getting populated in there.
Uses tcp_open_total instead of the http-based request_total to query for grafana variables

screen shot 2019-02-12 at 3 12 30 pm


klingerf:
All of the `dst_*` labels are hydrated using the destination response from the proxy API. That API is only called for known protocols, so my guess is that only protocol-specific stats include `dst_*` labels.

Inbound stats should have the labels we need, however, since those are generated by the prometheus scrape config. Maybe we can adjust all of the inbound queries to use TCP metrics, and keep all of the outbound queries on HTTP metrics?

hawkw:
@siggy I believe @klingerf is correct that `dst_` labels will only be present on HTTP/H2 metrics

siggy:
@klingerf Thanks for the explanation, and I agree it makes sense to use the TCP metrics on inbound queries only.

grampelberg:
Is it possible a user would get confused over this (TCP server, HTTP client)? Trying to think through the use cases and having a hard time, mostly because I'm not 100% up to speed on the grafana side of things.

siggy:
@grampelberg The queries in question are not user-facing. They are for generating sets of Grafana variables, to power the templating in the dashboards. For example, the Grafana template variable:
```
label_values(tcp_open_total{daemonset!=\"\"}, namespace)
```
...returns a list of every `daemonset` Prometheus knows about (via `linkerd-proxy`). We can then use that list to generate a section in the `Linkerd Daemonset` dashboard for each known daemonset.

rmars:
Okay! Thanks for the info, all! I've updated it so that only the user facing dropdowns (the queries with no `dst_`s in them) use `tcp_open_total`

grafana/dashboards/daemonset.json
```diff
@@ -2133,7 +2133,7 @@
         "multi": false,
         "name": "namespace",
         "options": [],
-        "query": "label_values(request_total{daemonset!=\"\"}, namespace)",
+        "query": "label_values(tcp_open_total{daemonset!=\"\"}, namespace)",
```
klingerf: Per in-person convo, it would be great to use `process_start_time_seconds` for these queries, since that will populate with all instances that have reported metrics to prometheus, rather than just those that have made network connections.

grafana/dashboards/statefulset.json
```diff
@@ -2133,7 +2133,7 @@
         "multi": false,
         "name": "namespace",
         "options": [],
-        "query": "label_values(request_total{statefulset!=\"\"}, namespace)",
+        "query": "label_values(tcp_open_total{statefulset!=\"\"}, namespace)",
```
siggy: `process_start_time_seconds`?

rmars: ahhhh thanks!

rmars:
@siggy yeah that makes sense! I've updated `top-line` and `health` to also use it!

Issue #1700, opened 2018-09-22
embellish cli output with emoji/color/lol


siggy:
as mentioned in TGIK8s:

emoji and color in `linkerd check`:
https://youtu.be/1dXJ34-KLe8?t=2739

ascii art / color / lolgopher for `linkerd dashboard`:
https://youtu.be/1dXJ34-KLe8?t=2797
https://github.com/kris-nova/lolgopher

klingerf:
@siggy This feels like a dupe of #1471 -- would you mind posting a comment there and closing this?

PR #2274, opened 2019-02-13
Embellish dashboard cli output with Linkerd logo

siggy:
This introduces a Linkerd color logo to the `linkerd dashboard` command,
to better indicate to the user that the terminal session is occupied by
a Linkerd dashboard port-forward.

Part of #1700

screen shot 2019-02-12 at 5 11 56 pm

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -408,6 +408,14 @@
   version = "v1.1"
 
 [[projects]]
+  branch = "master"
+  digest = "1:db3e74e12ad83aaa22a0e270573b923d8034efb6328a116c08d3e50f9f36508f"
+  name = "github.com/kris-nova/lolgopher"
+  packages = ["."]
+  pruneopts = ""
+  revision = "313b3abb0d9bb53e3c43230132061a7a9cb100fc"
+
+[[projects]]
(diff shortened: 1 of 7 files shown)


olix0r:
FWIW, I'm generally opposed to tools outputting ascii banners -- they eat up terminal real estate and frequently look terrible on unexpected terminal sizes/configurations. This might be more appropriate on a `linkerd thanks|credits` command -- i.e. a less-frequently-used command where whimsy might be less disruptive. Just my 0.2

siggy:
I'm open to finding a different spot for this. I personally find it useful for my own workflow, as I have lots of sessions open and this makes it easier to visually locate where I'm running the dashboard.

Re: unexpected terminal sizes/configurations, this color library, lolgopher, uses https://github.com/mattn/go-colorable under the hood, the same one we use for `linkerd check`. It's cross-platform and degrades properly when color is not supported.

grampelberg:
@siggy what do you mean by "visually locate where I'm running the dashboard"? Just that you've got it in the foreground and forget where it is? I'm wondering if we could make that experience better...

siggy:
@grampelberg yup, i have 10 terminal tabs open, sometimes several windows. i like running things in the foreground to more easily catch logging/errors at a glance.

PR #2195, opened 2019-02-03
Introduce Discovery API and endpoints command

siggy:
The Proxy API service lacked introspection of its internal state.

Introduce a new gRPC Discovery API, implemented by two servers:
1) Proxy API Server: returns a snapshot of discovery state
2) Public API Server: pass-through to the Proxy API Server

Also wire up a new `linkerd endpoints` command.

Fixes #2165

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

```bash
$ bin/linkerd endpoints -n emojivoto
IP            PORT   POD                      VERSION   WEIGHT   IDENTITY
10.1.10.134   8080   emoji-6664f85687-cr8vt   1564063   1        emoji-svc.emojivoto
10.1.10.135   8080   voting-dc698fcdc-9rq7k   1564092   1        voting-svc.emojivoto
10.1.10.136   80     web-78b6759fc7-r5bpb     1564073   1        web-svc.emojivoto

$ bin/linkerd endpoints
NAMESPACE   IP            PORT   POD                                   VERSION   WEIGHT   IDENTITY
emojivoto   10.1.10.134   8080   emoji-6664f85687-cr8vt                1564063   1        emoji-svc.emojivoto
emojivoto   10.1.10.135   8080   voting-dc698fcdc-9rq7k                1564092   1        voting-svc.emojivoto
emojivoto   10.1.10.136   80     web-78b6759fc7-r5bpb                  1564073   1        web-svc.emojivoto

NAMESPACE   IP            PORT   POD                                   VERSION   WEIGHT   IDENTITY
linkerd     10.1.10.130   8085   linkerd-controller-d566b445-x8llk     1564058   1        linkerd-controller-api.linkerd
linkerd     10.1.10.133   3000   linkerd-grafana-869d98557c-6xscx      1564069   1        linkerd-grafana.linkerd
linkerd     10.1.10.132   9090   linkerd-prometheus-6745fc66ff-fv7k6   1564116   1        linkerd-prometheus.linkerd
```

PR #2275, opened 2019-02-13
Add an Endpoints view to the web dashboard

rmars:
In #2195 we introduced `linkerd endpoints` on the CLI. I would like similar information to be on the web.

This PR adds an api endpoint at `/api/endpoints`, and introduces a new page with a table that is the same as the CLI table, available at `/endpoints` but currently not linked in the nav.

screen shot 2019-02-19 at 4 58 20 pm

### DIFF
--- a/web/app/js/components/Debug.jsx
+++ b/web/app/js/components/Debug.jsx
@@ -22,7 +22,8 @@ const endpointColumns = [
   },
   {
     title: "IP",
-    dataIndex: "ip"
+    dataIndex: "ip",
+    sorter: (a, b) => (a.ip).localeCompare(b.ip)
   },
   {
     title: "Port",
@@ -102,12 +103,12 @@ class Debug extends React.Component {
         {this.banner()}
         <Typography variant="h6">Endpoints</Typography>
         <Typography>
-        This table allow you to see Linkerd&#39;s service discovery state. It provides
-        debug information about the internal state of the
-        control-plane&#39;s proxy-api container. Note that this cache of service discovery
-        information is populated on-demand via linkerd-proxy requests. No endpoints
-        will be found  until a linkerd-proxy begins routing
-        requests.
+          This table allows you to see Linkerd&#39;s service discovery state. It
+          provides debug information about the internal state of the
+          control-plane&#39;s proxy-api container. Note that this cache of service
+          discovery information is populated on-demand via linkerd-proxy requests.
+          No endpoints will be found  until a linkerd-proxy begins routing
+          requests.
         </Typography>
 
         <BaseTable
--- a/web/app/js/components/Debug.jsx
+++ b/web/app/js/components/Debug.jsx
@@ -9,6 +9,7 @@ import _get from 'lodash/get';
 import _reduce from 'lodash/reduce';
 import { apiErrorPropType } from './util/ApiHelpers.jsx';
 import { numericSort } from './util/Utils.js';
+import Typography from '@material-ui/core/Typography';
 import { withContext } from './util/AppContext.jsx';
 import withREST from './util/withREST.jsx';
 
@@ -45,7 +46,7 @@ const endpointColumns = [
     sorter: (a, b) => (a.service).localeCompare(b.service)
   }
 ];
-class Endpoints extends React.Component {
+class Debug extends React.Component {
   static defaultProps = {
     error: null
   }
@@ -99,6 +100,15 @@ class Endpoints extends React.Component {
       <React.Fragment>
         {this.loading()}
         {this.banner()}
+        <Typography variant="h6">Endpoints</Typography>
+        <Typography>
+        This table allow you to see Linkerd's service discovery state. It provides
+        debug information about the internal state of the
+        control-plane's proxy-api container. Note that this cache of service discovery
+        information is populated on-demand via linkerd-proxy requests. No endpoints
+        will be found  until a linkerd-proxy begins routing
+        requests.
+        </Typography>
 
         <BaseTable
           tableRows={rows}
@@ -114,6 +124,6 @@ class Endpoints extends React.Component {
 }
 
 export default withREST(
-  withContext(Endpoints),
+  withContext(Debug),
   ({api}) => [api.fetch("/api/endpoints")]
 );
--- a/web/app/js/components/Debug.jsx
+++ b/web/app/js/components/Debug.jsx
@@ -4,12 +4,12 @@ import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import React from 'react';
 import Spinner from './util/Spinner.jsx';
+import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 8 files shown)


web/app/js/components/Endpoints.jsx
```diff
@@ -0,0 +1,110 @@
...
+          padding="dense" />
+
+      </React.Fragment>
+    );
+  }
+}
+
+export default withREST(
+  withContext(Endpoints),
+  ({api}) => [api.fetchMetrics("/api/endpoints")]
```
siggy: not a huge deal but this is adding a `window=1m` param on the URL, which is not needed/used by `/api/endpoints`.

web/app/js/components/Endpoints.jsx
```diff
@@ -0,0 +1,110 @@
...
+    sorter: (a, b) => numericSort(a.addr.port, b.addr.port)
+  },
+  {
+    title: "Pod",
+    dataIndex: "name",
+    sorter: (a, b) => (a.name).localeCompare(b.name)
+  },
+  {
+    title: "Resource Version",
+    dataIndex: "pod.resourceVersion"
```
siggy: tioli perhaps sorting by resource version could be useful in determining pod age / service discovery staleness?

web/srv/handlers.go
```diff
@@ -19,7 +20,7 @@ type (
 
 	handler struct {
 		render              renderTemplate
-		apiClient           pb.ApiClient
+		apiClient           public.APIClient
```
siggy: 👍

grampelberg:
Definitely not a blocker for this PR, but would it be possible to have links for pods/resources/services? Pods and services should link to the detail views and the resource could maybe pull the actual YAML?

web/app/js/components/Endpoints.jsx
rmars: good catch! I think `api.fetch` is better here

web/app/js/components/Endpoints.jsx
rmars: ohhh that's interesting. I'll add it!

rmars:
@grampelberg  oh yeah, that's a good idea, we can easily link to the detail views!

web/app/js/components/Endpoints.jsx
```diff
@@ -29,11 +31,13 @@ const endpointColumns = [
...
     dataIndex: "name",
-    sorter: (a, b) => (a.name).localeCompare(b.name)
+    sorter: (a, b) => (a.name).localeCompare(b.name),
+    render: d => <Link to={`/namespaces/${d.namespace}/pods/${d.name}`}>{d.name}</Link>
   },
   {
     title: "Resource Version",
-    dataIndex: "pod.resourceVersion"
+    dataIndex: "pod.resourceVersion",
+    sorter: (a, b) => (a.pod.resourceVersion).localeCompare(b.resourceVersion)
```
siggy: are these numbers stored as strings? will this sort numerically?

screen shot 2019-02-13 at 4 48 42 pm

rmars: yeah they are, but maybe I should set `numeric: true` on the `localeCompare` so that they are sorted as numbers

web/srv/server.go
```diff
@@ -107,8 +108,10 @@ func NewServer(
 	server.router.GET("/namespaces/:namespace/replicationcontrollers/:replicationcontroller", handler.handleIndex)
 	server.router.GET("/tap", handler.handleIndex)
 	server.router.GET("/top", handler.handleIndex)
+	server.router.GET("/endpoints", handler.handleIndex)
 	server.router.GET("/routes", handler.handleIndex)
 	server.router.GET("/profiles/new", handler.handleProfileDownload)
+
```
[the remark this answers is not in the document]
rmars: yeah, some of the other sections were space separated!

web/app/js/components/Endpoints.jsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  render() {
+    const { data } = this.props;
+    let results = _get(data, '[0].servicePorts', {});
+    let rows = _reduce(results, (mem, svc, svcName) => {
+      let pods = [];
+      _each(svc.portEndpoints, info => {
+        info.podAddresses.forEach(podAddress => {
+          podAddress.service = svcName;
```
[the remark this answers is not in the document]
rmars: great point! updated

rmars:
@dadjeibaah Yeah! I think this table could fit in the service mesh page! I think it could also be part of a `/debug` page, if we have more things to put in that page! I feel like the service mesh page is more "Overviewy" and this is more debug info, but we could also just try this on that page and see how it feels!

grampelberg:
My preference is for a `/debug` page. The mesh page needs a big revamp for it to be valuable again.

web/app/js/components/Debug.jsx
```diff
@@ -0,0 +1,129 @@
...
+        {this.loading()}
+        {this.banner()}
+        <Typography variant="h6">Endpoints</Typography>
+        <Typography>
+        This table allow you to see Linkerd&#39;s service discovery state. It provides
+        debug information about the internal state of the
+        control-plane&#39;s proxy-api container. Note that this cache of service discovery
+        information is populated on-demand via linkerd-proxy requests. No endpoints
+        will be found  until a linkerd-proxy begins routing
+        requests.
```
klingerf: You have two spaces here between "found" and "until". Might as well move "requests" up a line while you're at it.

web/app/js/components/Debug.jsx
```diff
@@ -0,0 +1,129 @@
...
+  {
+    title: "Namespace",
+    dataIndex: "namespace",
+    render: d => <Link to={"/namespaces/" + d.namespace}>{d.namespace}</Link>,
+    sorter: (a, b) => (a.namespace).localeCompare(b.namespace)
+  },
+  {
+    title: "IP",
+    dataIndex: "ip"
+  },
```
klingerf: TIOLI, "IP" is the only column that's not sortable. Would it make sense to add a sorter for it? I think it's also totally fine for it to not be sortable, but in that case I'd recommend removing the sorter from the "Port" column, since it feels pretty similar to this column.

web/app/js/components/Debug.jsx
```diff
@@ -0,0 +1,129 @@
...
+      return mem.concat(pods);
+    }, []);
+
+    return (
+      <React.Fragment>
+        {this.loading()}
+        {this.banner()}
+        <Typography variant="h6">Endpoints</Typography>
+        <Typography>
+        This table allow you to see Linkerd&#39;s service discovery state. It provides
```
klingerf: Typo here: `allow` => `allows`

web/app/js/components/Debug.jsx
rmars: reindented this section!

PR #2279, opened 2019-02-13
Issue 2276: Do not log error when timeout is blank

### LINKED ISSUE #2276: proxy-api logs errors when route timeout isn't set
When a route does not specify a timeout, the proxy-api defaults to the default timeout and logs an error:

```
time="2019-02-13T16:29:12Z" level=error msg="failed to parse duration for route POST /io.linkerd.proxy.destination.Destination/GetProfile: time: invalid duration "
```

We should not log any error when the timeout is blank.  Additionally, having service profile name and namespace in logs like this could be helpful.

### Environment

- Linkerd version: stable-2.2.0

kleimkuhler:
# Problem

When a route does not specify a timeout, the proxy-api defaults to the default
timeout and logs an error:

```
time="2019-02-13T16:29:12Z" level=error msg="failed to parse duration for route POST /io.linkerd.proxy.destination.Destination/GetProfile: time: invalid duration"
```

# Solution

We now check if a route timeout is blank. If it is not set, it is set to
`DefaultRouteTimeout`. If it is set, we try to parse it into a `Duration`.

A request was made to improve logging to include the service profile and
namespace as well.

# Validation

With valid service profiles installed, edit the `.yaml` to include an invalid
`timeout`:

```
...
name: GET /
timeout: foo
```

We should now see the following errors:

```
proxy-api time="2019-02-13T22:27:32Z" level=error msg="failed to parse duration for route 'GET /' in service profile 'webapp.default.svc.cluster.local' in namespace 'default': time: invalid duration foo"
```

This error does not show up when `timeout` is blank.

Fixes #2276

### DIFF
--- a/pkg/profiles/profiles.go
+++ b/pkg/profiles/profiles.go
@@ -105,10 +105,13 @@ func ToRoute(route *sp.RouteSpec) (*pb.Route, error) {
 		}
 		rcs = append(rcs, pbRc)
 	}
-	timeout, err := time.ParseDuration(route.Timeout)
-	if err != nil {
-		log.Errorf("failed to parse duration for route %s: %s", route.Name, err)
-		timeout = DefaultRouteTimeout
+	timeout := DefaultRouteTimeout
+	if route.Timeout != "" {
+		timeout, err = time.ParseDuration(route.Timeout)
+		if err != nil {
+			log.Errorf("failed to parse duration for route %s: %s", route.Name, err)
+			timeout = DefaultRouteTimeout
+		}
 	}
 	ret := pb.Route{
 		Condition:       cond,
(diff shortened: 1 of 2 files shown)


pkg/profiles/profiles.go
```diff
@@ -105,10 +105,19 @@ func ToRoute(route *sp.RouteSpec) (*pb.Route, error) {
...
-		log.Errorf("failed to parse duration for route %s: %s", route.Name, err)
-		timeout = DefaultRouteTimeout
+	timeout := DefaultRouteTimeout
+	if route.Timeout != "" {
+		timeout, err = time.ParseDuration(route.Timeout)
+		if err != nil {
+			log.Errorf(
+				"failed to parse duration for route '%s' in service profile '%s' in namespace '%s': %s",
+				route.Name,
+				profile.ObjectMeta.Name,
```
adleong: I might be mistaken, but I think you can just do `profile.Name` and `profile.Namespace`.

kleimkuhler: > I might be mistaken, but I think you can just do `profile.Name` and `profile.Namespace`.

Right - thanks!

PR #2281, opened 2019-02-14
Rename linkerd-proxy-api to linkerd-destination

olix0r:
Up until now, the proxy-api controller service has been the sole service
that the proxy communicates with, implementing the majoriry of the API
defined in the `linkerd2-proxy-api` repo. But this is about to change:
linkerd/linkerd2-proxy-api#25 introduces a new Identity service; and
this service must be served outside of the existing proxy-api service
in the linkerd-controller deployment (so that it may run under a
distinct service account).

With this change, the "proxy-api" name becomes less descriptive. It's no
longer "the service that serves the API for the proxy," it's "the
service that serves the Destination API to the proxy." Therefore, it
seems best to bite the bullet and rename this to be the "destination"
service.


olix0r:
Note that this shouldn't even be a backwards incompatible change when upgrading an existing cluster: as long as the linkerd-proxy-api service isn't actively _deleted_ (which it shouldn't be in most workflows), then extant running proxies should Just Work, because the linkerd-proxy-api and linkerd-destination service will both point at linkerd-controller pods on the same port. @adleong

controller/api/destination/server.go
```diff
@@ -62,7 +62,7 @@ func NewServer(
 	s := prometheus.NewGrpcServer()
 
 	// this server satisfies 2 gRPC interfaces:
-	// 1) linkerd2-proxy-api/destination.Destination (proxy-facing)
+	// 1) linkerd2-destination/destination.Destination (proxy-facing)
```
klingerf: Maybe this is an inadvertent rename? Am pretty sure "linkerd2-proxy-api" refers to the name of the repo here, which isn't changing.

controller/api/public/grpc_server_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/duration"
-	"github.com/linkerd/linkerd2/controller/api/proxy"
+	"github.com/linkerd/linkerd2/controller/api/destination"
```
klingerf: Oof, why does this test depend on the proxy package? Looks like this dependency was introduced in #2195, when we added the `linkerd endpoints` command. My preference would be to move it to a separate discovery package and have both destination and public depend on that. But I'm happy to do that in a follow-up branch. In the meantime, once you change the destination package name as suggested above, you'll need to use the new package name in this file, like so:

```diff
diff --git a/controller/api/public/grpc_server_test.go b/controller/api/public/grpc_server_test.go
index f30030cc..cf3c8b6c 100644
--- a/controller/api/public/grpc_server_test.go
+++ b/controller/api/public/grpc_server_test.go
@@ -549,7 +549,7 @@ func TestEndpoints(t *testing.T) {
 			}
 			k8sAPI.Sync()
 
-			discoveryClient, gRPCServer, proxyAPIConn := proxy.InitFakeDiscoveryServer(t, k8sAPI)
+			discoveryClient, gRPCServer, proxyAPIConn := destination.InitFakeDiscoveryServer(t, k8sAPI)
 			defer gRPCServer.GracefulStop()
 			defer proxyAPIConn.Close()
 
```

controller/cmd/public-api/main.go
```diff
@@ -25,7 +25,7 @@ func main() {
 	kubeConfigPath := flag.String("kubeconfig", "", "path to kube config")
 	prometheusURL := flag.String("prometheus-url", "http://127.0.0.1:9090", "prometheus url")
 	metricsAddr := flag.String("metrics-addr", ":9995", "address to serve scrapable metrics on")
-	proxyAPIAddr := flag.String("proxy-api-addr", "127.0.0.1:8086", "address of proxy-api service")
+	proxyAPIAddr := flag.String("destination-addr", "127.0.0.1:8086", "address of destination service")
```
klingerf: Can we also rename this variable to `destinationAPIAddr`? And then for completeness later in this file, `proxyAPIConn` to `destinationAPIConn`?

olix0r:
thank you, @klingerf !

Issue #2288, opened 2019-02-14
Refactor `linkerd inject` to make use of the new endpoint for injecting the proxy


alpeb:
Depends on #2286

klingerf:
Just a quick clarifying comment here: we should update `linkerd inject` to make use of the new endpoint, but for `linkerd install` we'll still need to use a library to inject the control plane install config, since the control plane won't be available at the time that the install config is generated.

alpeb:
@klingerf thanks for the clarification, that makes sense.

alpeb:
This has been superseded by #2447

Issue #2286, opened 2019-02-14
Expose a new endpoint for injection in the proxy auto-injector


alpeb:
Expose a new endpoint in the control plane that performs the proxy injection, given the new stored config (#1999). This will be used by both `linkerd install/inject` (#2288) and the auto-injector (#2289).

It should receive the resource yaml in the request, and return the injected yaml, or the JSON patch.
A request param should determine the type of the response.
As a followup, this endpoint should take into account the annotations/labels of the namespace and the resource itself for overriding the config (#2287)

alpeb:
We can provide this as a new endpoint in the controller api, or as part of the auto-injector, in which case we need to lift the current restriction of only installing the auto-injector when `--proxy-auto-inject` is provided at install time.

ihcsim:
My thought is to re-use the existing proxy-injector controller, and change the effect of the `--proxy-auto-inject` flag such that, if enabled, then the [admission controller webhook](https://github.com/linkerd/linkerd2/blob/c78f1053505fc95391142c77548356098fbc19de/controller/cmd/proxy-injector/main.go#L42-L63) is installed. Otherwise, it only serves the new endpoint.

klingerf:
👍 to @ihcsim's comment. Originally I was thinking we should add a separate endpoint that's called by both the CLI and the auto-injector, but I think we'll gain a lot by instead forcing the CLI to use the auto-inject endpoint directly (although not by way of the `linkerd-proxy-injector` Service that the webhook uses -- we'll instead need to expose that endpoint via the public API).

By forcing the CLI to use the auto-inject endpoint, it guarantees that CLI injection and auto-injection will maintain feature parity. We won't be able to add additional functionality to CLI inject without also making that functionality available to auto-inject users.

Also, since auto-inject requires modifications in JSON patch format but the CLI outputs fully-patched YAML, I feel like we'd end up with a lot of conditional formatting logic in an endpoint that provides both formats. Instead, if we only support JSON patch format, the CLI can leverage k8s.io code directly to apply those patches, and we won't need an inject implementation in our codebase that produces fully-patched YAML.

[...]

Issue #2029, opened 2019-01-02
Add option to inject to set resource limits


coleca:
## Feature Request
Currently there is an option to set the memory and CPU requests via `--proxy-memory` and `--proxy-cpu` but there is no way to set the limits without hand-editing the manifests.

### What problem are you trying to solve?
Setting the resource limits to prevent runaway usage in case of a memory or CPU leak.

### How should the problem be solved?
Add in an option `--proxy-memory-limit` and `--proxy-cpu-limit` to set the limits.

What do you want to happen? Add any considered drawbacks.
Probably best to not overload the existing CLI options for requests.

### Any alternatives you've considered?
No

Is there another way to solve this problem that isn't as good a solution?
Manually adjusting the manifests.

### How would users interact with this feature?
CLI

If you can, explain how users will be able to use this. Maybe some sample CLI
output?

`helm get manifest mychart |linkerd inject --proxy-cpu-limit=100m --proxy-cpu=100m --proxy-memory-limit=128Mi --proxy-memory=100Mi - |kubectl apply -f -`

Would create output:

`        resources:
          limits:
            cpu: 100m
            memory: 128Mi
          requests:
            cpu: 100m
            memory: 100Mi`

grampelberg:
Also related to: #2001 #1999

digmunhoz:
It's a really great idea!

I'd like just to say that it maybe would be interesting if the parameters to set the `requests` and `limits` use the same nomenclature which Kubernetes uses.

Something like: --proxy-memory-requests and --proxy-memory-limits.

I believe that this can avoid some misunderstanding.

TwinProduction:
This is also important if a pod needs to have a specific `qosClass`. Because `linkerd` doesn't currently allow inject to set the resource limit, it prevents a pod from having a `qosClass` of `Guaranteed`, even if all other containers in the pod meet the requirements.

grampelberg:
TIL, I've not played around with the `qosClass` stuff. Another great reason to set limits as well.

PR #2291, opened 2019-02-14
Draft on CLI install/inject config protobuf

### LINKED ISSUE #2285: Protobuf for configs to be stored in the control plane
Create protobuf (and generated) files for:

- [x] global config
- [x] proxy config
- [x] endpoints for persisting global and proxy configs

ihcsim:
This new protobuf defines the structure of the centralized global and proxy configurations provided during the CLI install and inject. It will be consumed by the CLI `inject` and the auto-injector.

Fixes #2285 


proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+
+message GlobalConfig {
+  string api_addr = 1;
+  string linkerd_namespace = 2;
+  string linkerd_version = 3;
+  string cluster_domain = 4;
+  bool no_init_container = 5;
+  bool ha = 6;
+  bool auto_inject = 7;
+}
```
alpeb: Any particular reason why not all the control plane configs are listed here?

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  Resource resource = 10;
+
+  TLS tls = 11;
+
+  int64 proxy_uid = 12;
+  LogLevel log_level = 13;
+  bool disable_external_profiles = 14;
+  map<string, uint32> proxy_outbound_capacity = 15;
+}
```
alpeb: I think we can simplify here by removing the `proxy_` prefixes, given we're already inside `ProxyConfig`.

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+message CNI {
+  Image image = 1;
+  LogLevel log_level = 2;
+  Port control_port = 3;
+
+  string dest_cni_bin_dir = 4;
+  string dest_cni_net_dir = 5;
+
+  ProxyConfig proxy_config = 6;
+}
```
alpeb: Nice, I had forgotten about CNI :+1:

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  Port api_port = 3;
+  Port control_port = 4;
+  repeated Port ignore_inbound_ports = 5;
+  repeated Port ignore_outbound_ports = 6;
+  Port inbound_port = 7;
+  Port metrics_port = 8;
+  Port outbound_port = 9;
+
+  Resource resource = 10;
```
klingerf: Right now `linkerd inject` lets you set resource requests, but there's also an open issue (#2029) for setting resource limits. It would be good to be more specific here about what this field represents. How about:

```
  ResourceRequirements resource_requests = 10;
```

(I recommend using `ResourceRequirements` to match the Kubernetes API)

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  TLS tls = 11;
```
klingerf: Per comment above, let's move this to the global config.

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  int64 proxy_uid = 12;
+  LogLevel log_level = 13;
+  bool disable_external_profiles = 14;
+  map<string, uint32> proxy_outbound_capacity = 15;
```
klingerf: Ah, this map is only required when injecting the control plane -- we don't expose any way for `linkerd inject` or auto-inject users to set it. Let's remove it from this definition, but take into account that we'll need to support this when implementing #2288.

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+}
+
+message Image{
+  string image_name = 1;
+  string pull_policy = 2;
+  string registry = 3;
+  string version = 4;
```
klingerf: This field duplicates the `linkerd_version` field from the `GlobalConfig` struct that you're adding. I think it could make sense in either struct, but would rather not duplicate this information. Come to think of it, maybe we don't need a version field in either of these structs? All of our control plane binaries are compiled to include the version that they're running, so perhaps the right behavior is to always use the compiled version from `pkg/version`, unless it's overridden by a `linkerd.io/inject-version` (or similar) annotation on the namespace or pod spec.

proto/config/config.proto
ihcsim: I'm still not 100% sure if all control plane configs are _global_. I guess as starter, we can include all the configs, and then add and trim as we go.

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
+syntax = "proto3";
+
+package linkerd2.config;
+
+option go_package = "github.com/linkerd/linkerd2/controller/gen/config";
+
+// ConfigStore provides an abstraction over the underlying configuration storge.
+service ConfigStore {
+  // Return the JSON-serialized global config.
+  rpc GetGlobalConfig(GetParam) returns (GlobalConfig) {}
```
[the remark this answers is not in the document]
alpeb: Just to confirm, the new config store endpoint will handle all the JSON (un)marshalling when persisting/retrieving the config to/from the ConfigMap, and consumers of this endpoint will only have to deal with the `GlobalConfig` struct and not with JSON. Is that correct?

alpeb:
@ihcsim I think you also need to add `controller/gen/config` here:
https://github.com/linkerd/linkerd2/blob/df0357505ce73453ed186726d747d22269714992/bin/protoc-go.sh#L7

proto/config/config.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  // Return the JSON-serialized proxy config.
+  rpc GetProxyConfig(GetParam) returns (ProxyConfig) {}
+
+  // Set the global config. It overwrites the existing value.
+  rpc SetGlobalConfig(GlobalConfig) returns (Result) {}
+
+  // Set the proxy config. It overwrites the existing value.
+  rpc SetProxyConfig(ProxyConfig) returns (Result) {}
+}
```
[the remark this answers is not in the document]
ihcsim: Yes, the intent is to provide an API to insulate all the consumers from whatever goes on at the k8s API and config storage level.

proto/config/config.proto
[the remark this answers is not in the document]
ihcsim: Following our comment at https://github.com/linkerd/linkerd2/issues/2286#issuecomment-463750562, I think the only consumer of this endpoint right now is the proxy-injector. And yes, the proxy-injector will be given a `GlobalConfig` or `ProxyConfig` struct, and act on it to create JSON patches to be returned to the CLI and the admission controller webhook.

proto/config/config.proto
ihcsim: I think as long as `--tls=optional` is still a thing, we will need this at the proxy level too, right?

proto/config/config.proto
ihcsim: (I think our last comments overlapped by less than a minute). I agree with starting by keeping the config to the minimal and increment as needed, simply because of the unknown effects of changing some of them during runtime (if we decided to watch them for reasons). 

Regarding `auto_inject` and others like `controller_log_level`, `disable_h2_upgrade`, `ha` etc., which aren't needed for install/inject, I was (jumping the :gun: again) hoping for a mechanism to record all the initial installation configurations for, who knows, if they would be helpful (?) for future debugging. But let's keep the struct small for now.

proto/config/config.proto
```diff
@@ -0,0 +1,89 @@
...
+  Port control_port = 4;
+  repeated Port ignore_inbound_ports = 5;
+  repeated Port ignore_outbound_ports = 6;
+  Port inbound_port = 7;
+  Port metrics_port = 8;
+  Port outbound_port = 9;
+
+  ResourceRequirements resource = 10;
+
+  TLS tls_enabled = 11;
```
klingerf: In the current auto-inject implementation, whether or not injected proxies use TLS is only configurable at control plane install time, and it applies globally to all injected deployments. In that case, it seems like we can use the TLS setting from `GlobalConfig` to satisfy the current behavior, without also need a TLS setting in this struct. It's entirely possible I'm forgetting something about the current implementation though.

proto/config/config.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message Port {
+  uint32 port = 1;
+}
+
+message ResourceRequirements {
+  float request_cpu = 1;
+  float request_memory = 2;
+}
```
alpeb: Can you please change these to strings? Kubernetes accepts some suffixes like `m`, `M` or `Mi`.

alpeb:
@ihcsim It looks like you removed the `version` field from both the global and the Image messages. We'll need that to build the tag for the image.

klingerf:
@alpeb Sorry about that, it was my request. See here for more details: https://github.com/linkerd/linkerd2/pull/2291#discussion_r257053223 In the auto-injector we should be able to use pkg/version.Version to get the running version, which I think is preferable to reading it from the config file.

proto/config/config.proto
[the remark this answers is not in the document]
olix0r: Let's think about this a little... is the go-generated Service code going to be the best API for interacting with the config?

When will the `Set` calls _ever_ make sense in this?

I think we care about protobuf _mostly_ for having reliable serialization; but we don't have any need to serialize requests/repsonses over a network; so this service abstraction seems unnecessary to me.

At the very least, it seems like something we can add separately from identifying the core data types, which is most important right now.

proto/config/config.proto
```diff
@@ -0,0 +1,89 @@
...
+  rpc SetGlobalConfig(GlobalConfig) returns (Result) {}
+
+  // Set the proxy config. It overwrites the existing value.
+  rpc SetProxyConfig(ProxyConfig) returns (Result) {}
+}
+
+message GlobalConfig {
+  string linkerd_namespace = 1;
+  bool cni_enabled = 2;
+  bool tls_enabled = 3;
```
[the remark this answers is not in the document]
olix0r: Not as a boolean ;)

We'll need to store things like the identity trust anchors, though...

I'd say that a boolean is fine for now, if we need to support the existing config surface area, but it will _definitely_ be deprecated in the near future (and so you may opt to use a high-numbered field ID).

proto/config/config.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message LogLevel {
+  string level = 1;
+}
+
+enum TLS {
+  NONE = 0;
+  OPTIONAL = 1;
+}
```
olix0r: Let's avoid introducing this type if we can. As I mentioned above, this is going to chagne substantially in the 2.3 release.

proto/config/config.proto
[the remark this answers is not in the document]
olix0r: Perhaps it should be a goal for the "API" in our applications to _hide_ the actual protobuf structs (which may change over time) and instead expose a higher-level interface that can be satisfied by reading the protobuf types.

For instance, having `config.IsCNIEnabled()` would be nice, so that I don't have to worry about whether that config is stored in global or inject config... etc.

proto/config/config.proto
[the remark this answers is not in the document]
ihcsim: Shall we replace the TLS enum (defined below) with a `message` so that we can introduce additional properties like identity trust anchors etc. into it, without affecting the `GlobalConfig` at this level?

proto/config/config.proto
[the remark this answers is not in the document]
ihcsim: I do think having some form of "API" over the configs is better than each container having to mount the config map, performs its own JSON serialization/deserialization and dealing with the protobuf types directly.

And yes, it's does make more sense to add something like this to the shared library, instead of over the network. I will remove this endpoint.

Also, since install, inject and auto-inject only work with JSON blobs, we can introduce individual accessors like `config.IsCNIEnabled()` later when the needs arise, to limit this PR to only what is currently needed.

olix0r:
Okay, in order to figure out how to proceed regarding `tls_enabled` and `enum TLS`...

In an upcoming change, we will remove the CA (as done on identity/master) and the "identity" service will be part of the default control plane installation. The identity service will need some configuration of its own (possibly from the global config... tbd), but the inject controller needs the following information:

- the "trust domain" -- a string, probably stored in the global config
- the "trust anchors" -- a blob of certificates; probably should be stored in a dedicated configmap so changes to this can be audited/acled separately from the base linkerd config.

There will also have to be an annotation that causes the proxy to opt-out of identity, but this probably _only_ needs to be an annotation on the target pods. We can support namespace/parent overrides, as well...

So, with this in mind, I expect us to end up with _at least_ something like the following...

```proto3
message GlobalConfig {
...
  IdentityContext identity_context = NN;
}

message IdentityContext {
  string trust_domain = 1;
}
```

So, if we want to introduce this message type now (maybe without the `trust_domain` until we need it), we could use the default null value to indicate what is currently "tls=disabled" and use it as set as "tls=optional" -- this is obviously a hack, but this should be very temporary (and documented).

Alternatively, if this is too subtle, we may want to just have a tls_enabled boolean that we put on a high-numbered field that we'll immediately deprecate...

ihcsim:
@olix0r That works for us. I've added an empty `IdentityContext` message, with a comment mentioning that null means "tls=disabled" and non-null means "tls=optional". Also, the config store service is removed.

PR #2292, opened 2019-02-14
Ignoring resources with empty kind field.

### LINKED ISSUE #2273: Inject output shows empty YAML
### What is the issue?

For [this YAML](https://gist.github.com/grampelberg/110db89003542231403364a5c58bda19), the inject output is:

```bash
secret "release-name-traefik-default-cert" skipped
configmap "release-name-traefik" skipped
serviceaccount "release-name-traefik" skipped
clusterrole "release-name-traefik" skipped
clusterrolebinding "release-name-traefik" skipped
service "release-name-traefik-dashboard" skipped
service "release-name-traefik" skipped
deployment "release-name-traefik" injected
ingress "release-name-traefik-dashboard" skipped
 "" skipped
 "" skipped
 "" skipped
 "" skipped
```

I would expect the last four lines to not be there at all.

Pothulapati:
Fixes #2273 

A Check is added to ignore empty resources in the `processYAML()` function used by both `inject..go` and `uninject.go`

@grampelberg @ihcsim


cli/cmd/endpoints.go
```diff
@@ -59,6 +59,7 @@ func newCmdEndpoints() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "endpoints [flags]",
+		Aliases: []string{"ep"},
```
klingerf: @Pothulapati Thanks for tackling this! Since it addresses a separate issue from #2273, can I trouble you to put it in a separate PR?

Pothulapati: Yes, It was a mistake. I unexpectedly pushed it to the same branch.

klingerf: No prob! Am glad to see it's such a straightforward fix.

Pothulapati: Done :)

ihcsim:
@Pothulapati This PR looks good! Would you mind updating the test fixtures in the `cli/cmd/testdata` folder, just to make sure our tests cover this change? One idea is to update the `inject_emojivoto_deployment.input.yml` and `inject_emojivoto_list.input.yml` files with additional lines that get translated into an empty `Kind`. If your code works, none of the report outputs will break, but you will need to update the response YAML files.

Pothulapati:
@ihcsim I wrote the tests for yaml's with multiple resources along with empty resources, but for `kind: List`  with empty resources like [here](https://gist.github.com/Pothulapati/7618a786141346c518d45ac0e6028af1) , The output YAML replaces the empty elemts with `- null`. 

Kubectl throws a error, both with null List elements or comments like shown in the example (i.e input), Should Linkerd CLI have a condition and not emit null values and fix it? Or should we go with the CLI being in sync with the kubectl yaml's?

ihcsim:
It's ok to let kubectl catch and error out. The Linkerd CLI shouldn't change the user's experience by auto-magically "fixes" a broken YAML. If a YAML failed with kubectl, it should also fail when the Linkerd CLI is added into the mix.

Pothulapati:
@ihcsim Added empty resources to the `inject_emojivoto_deployment.input.yml` and `inject_emojivoto_list.input.yml` . 

Also had to update multiple output files, as the same `inject_emojivoto_deployment.input.yml` is used as input in multiple tests. :)

ihcsim:
@siggy That is @Pothulapati original suggestion actually. My preference is to just let the YAML pass through our CLI, and let `kubectl` decide for itself what YAML is acceptable or not. If we do decide to perform the above check in `inject.go`, a similar check will need to be added to `uninject.go` too.

ihcsim:
@Pothulapati Apologies for the back-and-forth. I just chatted with @siggy. So we will still let `kubectl` failed if the YAML contains some null values in the list, but we will update the inject report to output the detection of the missing `kind` field. Can you make the change as @siggy suggested above? Also, remember to include this change in [`uninject.go`](https://github.com/linkerd/linkerd2/blob/12527169b87c7788584474bd9486b2f926727ca3/cli/cmd/uninject.go#L111-L113). Thanks!

Pothulapati:
@ihcsim Sure, Just updated My Code, 
Also, Sould I create new testdata files to check for Empty field resources?
 It feels a bit awkward to change multiple output files, as the `inject_emojivoto_deployment.yml` input file is being used for different tests. 
Thanks :+1:

ihcsim:
> It feels a bit awkward to change multiple output files

Yes, you are right. Since the latest change to the `generateReport()` method does require you to update multiple `.golden` and `.report` test files if we re-use the `inject_emojivoto_deployment_input.yml` file. It might be easier to just create new, separate testdata files.

PR #2296, opened 2019-02-15
Generate CLI docs for usage by the website

grampelberg:
Replaces #2278.


BUILD.md
```diff
@@ -427,3 +428,9 @@ build_architecture
   }
 build_architecture
 </details>
+
+# Generating CLI docs
```
dadjeibaah: How would we use the generated YAML from the command with the `linkerd/website` repo? I think we may need additional documentation to describe the `doc` command's If that doesn't exist already somewhere.

grampelberg: It hasn't been wired up yet, but will look something like https://github.com/linkerd/website/blob/master/linkerd.io/data/cli.yaml. My preference would be to have all the documentation on how to do it in the website repo (since that's the end display case).

grampelberg:
@dadjeibaah it is nice to have a reference page for CLI. There's a couple that have been done by hand, this will get all the docs and make sure we're in sync from a version perspective.

cli/cmd/doc.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	var doc cmdDoc
+	if err := yaml.Unmarshal(buf.Bytes(), &doc); err != nil {
+		return err
+	}
+
+	// Cobra names start with linkerd, strip that off for the docs.
+	doc.Name = strings.Replace(doc.Name, "linkerd ", "", 1)
```
klingerf: Would be slightly more accurate to use `TrimPrefix` here:

```suggestion
	doc.Name = strings.TrimPrefix(doc.Name, "linkerd ")
```

BUILD.md
```diff
@@ -427,3 +428,9 @@ build_architecture
...
+
+The [documentation](https://linkerd.io/2/reference/cli/) for the CLI
+tool is partially generated from YAML. This can be generated by running the
+`doc` command.
```
klingerf: Maybe `` `linkerd doc` command. ``, for clarity?

cli/cmd/doc.go
```diff
@@ -0,0 +1,91 @@
...
+	Name             string
+	Synopsis         string      `yaml:",omitempty"`
+	Description      string      `yaml:",omitempty"`
+	Options          []cmdOption `yaml:",omitempty"`
+	InheritedOptions []cmdOption `yaml:"inherited_options,omitempty"`
+	Example          string      `yaml:",omitempty"`
+	SeeAlso          []string    `yaml:"see_also,omitempty"`
+}
+
+var cmdList []cmdDoc
```
klingerf: In generally I'm anti-global variables, especially in this package, which includes all of our CLI subcommands. It seems like you could remove this one pretty easily by having `generateDocs` return the list that it populates. Something like:

```diff
diff --git a/cli/cmd/doc.go b/cli/cmd/doc.go
index b172b908..dc744339 100644
--- a/cli/cmd/doc.go
+++ b/cli/cmd/doc.go
@@ -28,8 +28,6 @@ type cmdDoc struct {
 	SeeAlso          []string    `yaml:"see_also,omitempty"`
 }
 
-var cmdList []cmdDoc
-
 func newCmdDoc() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:    "doc",
@@ -37,7 +35,8 @@ func newCmdDoc() *cobra.Command {
 		Short:  "Generate YAML documentation for the Linkerd CLI",
 		Args:   cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := generateDocs(RootCmd); err != nil {
+			cmdList, err := generateDocs(RootCmd)
+			if err != nil {
 				return err
 			}
 
@@ -57,35 +56,39 @@ func newCmdDoc() *cobra.Command {
 
 // generateDocs takes a command and recursively walks the tree of commands,
 // adding each as an item to cmdList.
-func generateDocs(cmd *cobra.Command) error {
+func generateDocs(cmd *cobra.Command) ([]cmdDoc, error) {
+	var cmdList []cmdDoc
+
 	for _, c := range cmd.Commands() {
 		if !c.IsAvailableCommand() || c.IsAdditionalHelpTopicCommand() {
 			continue
 		}
 
-		if err := generateDocs(c); err != nil {
-			return err
+		subList, err := generateDocs(c)
+		if err != nil {
+			return cmdList, err
 		}
+		cmdList = append(cmdList, subList...)
 	}
 
 	var buf bytes.Buffer
 
 	if err := doc.GenYaml(cmd, io.Writer(&buf)); err != nil {
-		return err
+		return cmdList, err
 	}
 
 	var doc cmdDoc
 	if err := yaml.Unmarshal(buf.Bytes(), &doc); err != nil {
-		return err
+		return cmdList, err
 	}
 
 	// Cobra names start with linkerd, strip that off for the docs.
-	doc.Name = strings.Replace(doc.Name, "linkerd ", "", 1)
+	doc.Name = strings.TrimPrefix(doc.Name, "linkerd ")
 
 	// Don't include the root command.
 	if doc.Name != "linkerd" {
 		cmdList = append(cmdList, doc)
 	}
 
-	return nil
+	return cmdList, nil
 }
```

Mind updating?

cli/cmd/doc.go
```diff
@@ -0,0 +1,91 @@
...
+			if err := generateDocs(RootCmd); err != nil {
+				return err
+			}
+
+			out, err := yaml.Marshal(cmdList)
+			if err != nil {
+				return err
+			}
+
+			fmt.Println(fmt.Sprintf("%s", out))
```
klingerf: Super minor, but I'd typically write this as:

```suggestion
			fmt.Println(string(out))
```

grampelberg:
@klingerf great suggestions! That global was bothering me too =)

PR #2297, opened 2019-02-15
Adding filter feature to web UI

### LINKED ISSUE #1792: Dashboard: Add sorting and filtering to Metric Tables

scottcarol:
Fixes #1792.

This PR adds filter functionality to the web UI via an optional Material-UI `<Toolbar>` on the top of the table which contains the table's title and a text input. This toolbar only shows if the `enableFilter={true}` prop is passed down from the parent component. 

`MetricsTable` and `TopRoutesTable` components were modified to include a `filter()` method on certain column types. `filter()` returns a comma-separated string of table cell data that can be filtered by user input from `BaseTable`.

The PR also includes additions to the `MetricsTable` test and adds tests for `BaseTable` and `TopRoutesTable`.

Note: The previous Ant-based UI allowed certain tables to be filtered by individual table column; this capacity is not part of this PR but can be added later if useful.


web/app/js/components/MetricsTable.jsx
```diff
@@ -161,26 +162,30 @@ class MetricsTable extends React.Component {
...
 
     let columns = columnDefinitions(resource, showNsColumn, api.PrefixedLink);
     let rows = preprocessMetrics(metrics);
     return (
       <BaseTable
+        enableFilter={true}
         tableRows={rows}
         tableColumns={columns}
         tableClassName="metric-table"
+        title={title}
```
siggy: A side effect of adding a filter box to a table without a title is that it creates a bit of empty space:
screen shot 2019-02-18 at 11 45 14 am

Perhaps we should consider moving the section titles into the tables themselves, like you did in `NamespaceLanding.jsx`, or, something else? Also this is totally fine to do in a subsequent PR. @rmars thoughts?

web/app/js/components/MetricsTable.test.js
```diff
@@ -28,6 +28,39 @@ describe('Tests for <MetricsTable>', () => {
     expect(table.props().tableColumns).toHaveLength(10);
   });
 
+  it('if enableFilter is true, user can filter rows by search term', () => {
```
siggy: 🙏

web/app/js/components/BaseTable.jsx
```diff
@@ -54,14 +63,37 @@ class BaseTable extends React.Component {
...
+    this.setState({ filterBy: filterBy });
+  }
+
+  generateRows = (tableRows, tableColumns, order, orderBy, filterBy) => {
+    if (!orderBy && !filterBy) {
       return tableRows;
     }
 
     let col = _find(tableColumns, d => d.dataIndex === orderBy);
     let sorted = tableRows.sort(col.sorter);
```
siggy: i think this could error out if `orderBy === null`, though i can't reproduce it?

consider sorting only when `orderBy !== null`. also consider consolidating the return statements into a single one at the end, something like (this bit of totally untested code):
```diff
diff --git a/web/app/js/components/BaseTable.jsx b/web/app/js/components/BaseTable.jsx
index 2d9690e2..6ecfcd18 100644
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -73,15 +73,16 @@ class BaseTable extends React.Component {
       return tableRows;
     }

-    let col = _find(tableColumns, d => d.dataIndex === orderBy);
-    let sorted = tableRows.sort(col.sorter);
-    if (!filterBy) {
-      return order === 'desc' ? sorted.reverse() : sorted;
-    } else {
+    let rows = tableRows;
+    if (orderBy) {
+      let col = _find(tableColumns, d => d.dataIndex === orderBy);
+      rows = rows.sort(col.sorter);
+    }
+    if (filterBy) {
       let filteredRows = [];
       let swapWildCard = /[*]/g; // replace "*"" in filterBy with wildcard
       let regexString = new RegExp(filterBy.replace(swapWildCard, ".+"), "i");
-      let extractedTableRowText = sorted.reduce((acc, row) => {
+      let extractedTableRowText = rows.reduce((acc, row) => {
         if (col.filter) {
           acc.push(col.filter(row));
         }
@@ -89,11 +90,13 @@ class BaseTable extends React.Component {
       },[]);
       extractedTableRowText.forEach((row, rowIndex) => {
         if (row.match(regexString)) {
-          filteredRows.push(sorted[rowIndex]);
+          filteredRows.push(rows[rowIndex]);
         }
       });
-      return order === 'desc' ? filteredRows.reverse() : filteredRows;
+      rows = filteredRows;
     }
+
+    return order === 'desc' ? rows.reverse() : rows;
   }

   renderHeaderCell = (col, order, orderBy) => {
```

web/app/js/components/BaseTable.jsx
```diff
@@ -54,14 +63,37 @@ class BaseTable extends React.Component {
...
+      let swapWildCard = /[*]/g; // replace "*"" in filterBy with wildcard
+      let regexString = new RegExp(filterBy.replace(swapWildCard, ".+"), "i");
+      let extractedTableRowText = sorted.reduce((acc, row) => {
+        if (col.filter) {
+          acc.push(col.filter(row));
+        }
+        return acc;
+      },[]);
+      extractedTableRowText.forEach((row, rowIndex) => {
+        if (row.match(regexString)) {
```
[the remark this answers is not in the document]
scottcarol: Great catch!

web/app/js/components/MetricsTable.jsx
scottcarol: Agreed, where titles occur are not consistent throughout the UI. Let me brainstorm with @rmars about a consistent approach.

web/app/js/components/MetricsTable.jsx
```diff
@@ -42,6 +42,7 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink) => {
       title: isMultiResourceTable ? "Resource" : friendlyTitle(resource).singular,
       dataIndex: "name",
       isNumeric: false,
+      filter: d => d.name + "," + d.namespace,
```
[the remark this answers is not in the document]
scottcarol: The Deployments-->All view does show namespace name, but you're right, you shouldn't be able to filter by unseen data in your example. Will change!  

deploymentsview

web/app/js/components/TopEventTable.jsx
```diff
@@ -34,6 +34,7 @@ const topColumns = (resourceType, ResourceLink, PrefixedLink) => [
     dataIndex: "count",
     isNumeric: true,
     defaultSortOrder: "desc",
+    filter: d => d.path + ',' + d.httpMethod,
```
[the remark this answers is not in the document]
scottcarol: Thanks for the input, I restructured how the filter() method worked so going to resolve this now!

web/app/js/components/BaseTable.jsx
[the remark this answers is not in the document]
scottcarol: Thanks again 🎈

web/app/js/components/MetricsTable.jsx
scottcarol: I agree, @rmars are you cool with me doing a separate PR for this?

web/app/js/components/BaseTable.jsx
scottcarol: Took @siggy's suggestion, thank you!

web/app/js/components/BaseTable.jsx
```diff
@@ -54,14 +63,37 @@ class BaseTable extends React.Component {
...
     }
 
     let col = _find(tableColumns, d => d.dataIndex === orderBy);
     let sorted = tableRows.sort(col.sorter);
-    return order === 'desc' ? sorted.reverse() : sorted;
+    if (!filterBy) {
+      return order === 'desc' ? sorted.reverse() : sorted;
+    } else {
+      let filteredRows = [];
+      let swapWildCard = /[*]/g; // replace "*"" in filterBy with wildcard
```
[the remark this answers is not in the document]
scottcarol: Good 👁 !

scottcarol:
@rmars Thanks so much for the feedback on structure - tell me if the new structure makes sense to you. When a user enters a search term, `generateRows()` in `BaseTable` filters the `tableColumns` for columns that have a `filter()` method (specified in `MetricsTable`, `TopRoutesTable` and `TopEventTable`.

Mapping through the table rows, we return an array of searchable text strings corresponding to each `tableRow`. If there is a match with the search term, the original row is added to a new `filteredRows` array.

One other change from the original PR, I switched the icon and UI from an exposed text field (my original idea) to the k8s dashboard UI (click on a filter icon to expose the input) for consistency's sake.

web/app/js/components/BaseTable.test.js
```diff
@@ -0,0 +1,94 @@
...
+  it("renders the table with sample data", () => {
+    let extraProps = _merge({}, defaultProps, {
+      tableRows: [{
+        name: "authors",
+        namespace: "default",
+        key: "default-deployment-authors",
+        successRate: 0.6956521739130435,
+        requestRate: 6.9,
+        latency: {P50: 4, P95: 17, P99: 66},
+        pods: {totalPods: "1", meshedPods: "1"},
```
rmars: if we're only including two columns for the table definition, do we need all the data in this object? it kinda makes it harder to see what's being tested. maybe include only data for columns that are defined? 
another thing to do could be to have a generate fake data function that will generate this datum to avoid having a bunch of these data objects in here (mentioned below too).

web/app/js/components/TopEventTable.jsx
```diff
@@ -16,17 +16,28 @@ const topColumns = (resourceType, ResourceLink, PrefixedLink) => [
   },
   {
     title: "Name",
+    filter: d => {
+      if (d.path === "/ping") {
+        return "";
+      } else if (d.direction === "INBOUND") {
+        return "deploy/" + d.source.owner.split('/')[1];
```
rmars: this section is a little too application specific (not all applications have a `/ping` endpoint, or have deployments that we're Top-ping).

I think we should filter on whatever the visible name is for this column, e.g. `ns/emojivoto` or `po/web-1234`
To do this, we should return the filter text as whatever is displayed. We'll need a little bit of refactoring so this isn't repetitive:
```diff
diff --git a/web/app/js/components/TopEventTable.jsx b/web/app/js/components/TopEventTable.jsx
index bf4673e7..d98fc3ca 100644
--- a/web/app/js/components/TopEventTable.jsx
+++ b/web/app/js/components/TopEventTable.jsx
@@ -1,10 +1,11 @@
-import { directionColumn, srcDstColumn, tapLink } from './util/TapUtils.jsx';
-import { formatLatencySec, numericSort } from './util/Utils.js';
+import { directionColumn, extractDisplayName, srcDstColumn, tapLink } from './util/TapUtils.jsx';
+import { formatLatencySec, numericSort, toShortResourceName } from './util/Utils.js';

 import BaseTable from './BaseTable.jsx';
 import PropTypes from 'prop-types';
 import React from 'react';
 import SuccessRateMiniChart from './util/SuccessRateMiniChart.jsx';
+import _isEmpty from 'lodash/isEmpty';
 import _isNil from 'lodash/isNil';
 import { withContext } from './util/AppContext.jsx';

@@ -17,13 +18,11 @@ const topColumns = (resourceType, ResourceLink, PrefixedLink) => [
   {
     title: "Name",
     filter: d => {
-      if (d.path === "/ping") {
-        return "";
-      } else if (d.direction === "INBOUND") {
-        return "deploy/" + d.source.owner.split('/')[1];
-      } else if (d.direction === "OUTBOUND") {
-        return "deploy/" + d.destination.owner.split('/')[1];
-      }
+      let [labels, display] = extractDisplayName(d);
+
+      return _isEmpty(labels[resourceType]) ?
+        display.str :
+        toShortResourceName(resourceType) + "/" + labels[resourceType];
     },
     key: "src-dst",
     render: d => srcDstColumn(d, resourceType, ResourceLink)
diff --git a/web/app/js/components/util/TapUtils.jsx b/web/app/js/components/util/TapUtils.jsx
index fe7c9e78..3b95cdec 100644
--- a/web/app/js/components/util/TapUtils.jsx
+++ b/web/app/js/components/util/TapUtils.jsx
@@ -312,8 +312,7 @@ export const directionColumn = d => (
   </Tooltip>
 );

-export const srcDstColumn = (d, resourceType, ResourceLink) => {
-
+export const extractDisplayName = d => {
   let display = {};
   let labels = {};

@@ -324,6 +323,11 @@ export const srcDstColumn = (d, resourceType, ResourceLink) => {
     display = d.destination;
     labels = d.destinationLabels;
   }
+  return [labels, display];
+};
+
+export const srcDstColumn = (d, resourceType, ResourceLink) => {
+  let [labels, display] = extractDisplayName(d);
```

web/app/js/components/MetricsTable.jsx
scottcarol: Created issue #2403 for this!

web/app/js/components/BaseTable.jsx
```diff
@@ -54,14 +68,44 @@ class BaseTable extends React.Component {
...
+      let filteredRows = [];
+      let columnsToFilter = tableColumns.filter(col => col.filter);
+      let textToSearch = rows.map(row => {
+        let string = "";
+        columnsToFilter.forEach(col => {
+          string+= col.filter(row);
+        });
+        return string;
+      });
+      textToSearch.forEach((row, rowIndex) => {
```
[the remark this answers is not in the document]
scottcarol: Amazing, great points.

web/app/js/components/BaseTable.test.js
```diff
@@ -0,0 +1,94 @@
...
+      {
+        dataIndex: "namespace",
+        title: "Namespace",
+        sorter: (a, b) => (a.namespace || "").localeCompare(b.namespace),
+        defaultOrderBy: "asc"
+      }],
+      enableFilter: true
+    });
+    const component = mount(routerWrap(BaseTable, extraProps));
+
```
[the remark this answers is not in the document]
scottcarol: Sounds good!

web/app/js/components/TopEventTable.jsx
scottcarol: Super helpful context and suggestion, thanks! 🌞

web/app/js/components/BaseTable.test.js
scottcarol: Agreed re: unused data, I removed it from the tests and will build a generate fake data function in a separate PR.

web/app/js/components/util/TapUtils.jsx
```diff
@@ -312,18 +312,14 @@ export const directionColumn = d => (
...
+  return [labels, display];
+};
 
-  if (d.direction === "INBOUND") {
-    display = d.source;
-    labels = d.sourceLabels;
-  } else {
-    display = d.destination;
-    labels = d.destinationLabels;
-  }
```
klingerf: We still need to check the direction of the event here before extracting the name. If this is an inbound event, then we want to use the source, whereas as written you're always using the destination. I think this fixes it:

```diff
diff --git a/web/app/js/components/util/TapUtils.jsx b/web/app/js/components/util/TapUtils.jsx
index d8b60243..3b95cdec 100644
--- a/web/app/js/components/util/TapUtils.jsx
+++ b/web/app/js/components/util/TapUtils.jsx
@@ -313,8 +313,16 @@ export const directionColumn = d => (
 );
 
 export const extractDisplayName = d => {
-  let display = d.destination || {};
-  let labels = d.destinationLabels || {};
+  let display = {};
+  let labels = {};
+
+  if (d.direction === "INBOUND") {
+    display = d.source;
+    labels = d.sourceLabels;
+  } else {
+    display = d.destination;
+    labels = d.destinationLabels;
+  }
   return [labels, display];
 };
 
```

scottcarol: Thanks for the catch, my new commit should resolve this.

web/app/js/components/BaseTable.jsx
```diff
@@ -100,13 +137,42 @@ class BaseTable extends React.Component {
...
+          <FilterListIcon
+            className={classes.toolbarIcon}
+            onClick={this.handleFilterToggle} />}
+        {this.state.showFilter &&
+          <CloseIcon
+            className={classes.toolbarIcon}
+            onClick={this.handleFilterToggle} />}
+      </Toolbar>
+    );
+    return toolbar;
```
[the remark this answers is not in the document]
scottcarol: Good point!

scottcarol:
> I'm observing a super weird issue where, when I install the books app and go to the top routes page, the order of the sections bounces around a lot. -- @rmars 

Well spotted! I was able to reproduce this in the edge release by running `linkerd dashboard` with the CLI, so I don't think it's related to this branch. I'll file an issue for it.

Issue #2217, opened 2019-02-06
proxy-api: Only return TLS identities when dst pod is known to participate in identity


olix0r:
The proxy-api returns a TLS identity whenever the pod is annotated with a controller namespace:

https://github.com/linkerd/linkerd2/blob/50fbcc60b50a367209672a8b6e34398955d4344a/controller/api/proxy/endpoint_listener.go#L202-L220

Then, the proxy checks the namespace to determine whether TLS should be initiated:

https://github.com/linkerd/linkerd2-proxy/blob/4b0e50e8fc488f535a4688447efa89cd650fec92/src/transport/tls/identity.rs#L26-L33

This is fragile and, in fact, the only reason that the proxy has to know anything about kubernetes namespaces; and the K8sPodIdentity type is the only kubernetes specific-logic that has bled into the proxy api.

At the very least, the controller check should be made server-side, so that the proxy-api only returns identities of pods that it expects the client to be able to validate.

Following that, the proxy should stop performing this check and, eventually, the controllerNS value should be deprecated in the API and server should stop setting it.  And, preferably, the entire K8sPodIdentity type should be removed from the API.

Furthermore, the inject process should be modified to set an annotation on pods to explicitly indicate whether the pod should participate in identity. The proxy-api should also check this annotation before returning an identity in a resolution.

klingerf:
I'm pretty sure this issue is the same as #1274, but phrased a bit differently. Just wanted to pass it along in case you wanted to see some prior art for implementation details.

PR #2302, opened 2019-02-15
identity/master: Disable identity hinting

olix0r:
The destination service's identity hinting is going to change
substantially with the new identity system.

While we're in a transitionary state, let's just disable identity
hinting entirely until the new system is in place.

Relates to #2217


controller/cmd/destination/main.go
```diff
@@ -21,7 +21,6 @@ func main() {
 	kubeConfigPath := flag.String("kubeconfig", "", "path to kube config")
 	k8sDNSZone := flag.String("kubernetes-dns-zone", "", "The DNS suffix for the local Kubernetes zone.")
 	enableH2Upgrade := flag.Bool("enable-h2-upgrade", true, "Enable transparently upgraded HTTP2 connections among pods in the service mesh")
-	enableTLS := flag.Bool("enable-tls", false, "Enable TLS connections among pods in the service mesh")
```
klingerf: We also need to stop setting this flag in `chart/templates/base.yaml`

adleong:
Does this change cause TLS to no longer be used?  Should we remove the `--tls` flag or make it fail loudly as part of this change?

adleong:
Oh, the base branch here is `identity/master`.  Nevermind.

PR #2305, opened 2019-02-16
Fix handling of kubeconfig server urls that include paths

### LINKED ISSUE #2304: CLI is dropping paths from kubeconfig server URLs
When the `linkerd` CLI loads the API server URL from a kubeconfig file, it strips the path from the URL. This means that our CLI can't successfully connect to clusters where the API server is served from a non-root path (e.g. Rancher clusters). We must fix this by preserving the path, as described in the client-go rest package documentation:

```
// If a URL is given then the (optional) Path of that URL represents a prefix that must
// be appended to all request URIs used to access the apiserver. This allows a frontend
// proxy to easily relocate all of the apiserver endpoints.
```

klingerf:
In #2206, we adjusted the way in which we build Kubernetes API URLs to handle API servers that include trailing slashes. Unfortunately in the process we started dropping the paths from API servers that include paths. This change semi-reverts the fix from #2206, to fix both issues. In the process, I've cleaned up some misleadingly-named variables and functions, and fixed some indentation issues in our test kubeconfig file.

Fixes #2304.

### DIFF
--- a/pkg/k8s/api.go
+++ b/pkg/k8s/api.go
@@ -121,12 +121,12 @@ func (kubeAPI *KubernetesAPI) getPods(ctx context.Context, client *http.Client, 
 }
 
 // URLFor generates a URL based on the Kubernetes config.
-func (kubeAPI *KubernetesAPI) URLFor(namespace string, extraPathStartingWithSlash string) (*url.URL, error) {
-	return generateKubernetesAPIBaseURLFor(kubeAPI.Host, namespace, extraPathStartingWithSlash)
+func (kubeAPI *KubernetesAPI) URLFor(namespace, path string) (*url.URL, error) {
+	return generateKubernetesAPIURLFor(kubeAPI.Host, namespace, path)
 }
 
 func (kubeAPI *KubernetesAPI) getRequest(ctx context.Context, client *http.Client, path string) (*http.Response, error) {
-	endpoint, err := BuildURL(kubeAPI.Host, path)
+	endpoint, err := generateKubernetesURL(kubeAPI.Host, path)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/k8s/api_test.go
+++ b/pkg/k8s/api_test.go
@@ -1,7 +1,6 @@
 package k8s
 
 import (
-	"fmt"
 	"testing"
 )
 
@@ -11,15 +10,24 @@ func TestKubernetesApiUrlFor(t *testing.T) {
 
 	t.Run("Returns base config containing k8s endpoint listed in config.test", func(t *testing.T) {
 		tests := []struct {
-			server      string
 			kubeContext string
+			expected    string
 		}{
-			{"https://55.197.171.239", ""},
-			{"https://162.128.50.11", "clusterTrailingSlash"},
+			{
+				kubeContext: "",
+				expected:    "https://55.197.171.239/api/v1/namespaces/some-namespace/some/extra/path",
+			},
+			{
+				kubeContext: "clusterTrailingSlash",
+				expected:    "https://162.128.50.11/api/v1/namespaces/some-namespace/some/extra/path",
+			},
+			{
+				kubeContext: "clusterWithPath",
+				expected:    "https://162.128.50.12/k8s/clusters/c-fhjws/api/v1/namespaces/some-namespace/some/extra/path",
+			},
 		}
 
 		for _, test := range tests {
-			expected := fmt.Sprintf("%s/api/v1/namespaces/%s%s", test.server, namespace, extraPath)
 			api, err := NewAPI("testdata/config.test", test.kubeContext)
 			if err != nil {
 				t.Fatalf("Unexpected error creating Kubernetes API: %+v", err)
@@ -28,8 +36,8 @@ func TestKubernetesApiUrlFor(t *testing.T) {
 			if err != nil {
 				t.Fatalf("Unexpected error generating URL: %+v", err)
 			}
-			if actualURL.String() != expected {
-				t.Fatalf("Expected generated URL to be [%s], but got [%s]", expected, actualURL.String())
+			if actualURL.String() != test.expected {
+				t.Fatalf("Expected generated URL to be [%s], but got [%s]", test.expected, actualURL.String())
 			}
 		}
 	})
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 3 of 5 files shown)


pkg/k8s/api.go
```diff
@@ -121,12 +121,12 @@ func (kubeAPI *KubernetesAPI) getPods(ctx context.Context, client *http.Client,
...
 // URLFor generates a URL based on the Kubernetes config.
-func (kubeAPI *KubernetesAPI) URLFor(namespace string, extraPathStartingWithSlash string) (*url.URL, error) {
-	return generateKubernetesAPIBaseURLFor(kubeAPI.Host, namespace, extraPathStartingWithSlash)
+func (kubeAPI *KubernetesAPI) URLFor(namespace, path string) (*url.URL, error) {
+	return generateKubernetesAPIURLFor(kubeAPI.Host, namespace, path)
 }
 
 func (kubeAPI *KubernetesAPI) getRequest(ctx context.Context, client *http.Client, path string) (*http.Response, error) {
-	endpoint, err := BuildURL(kubeAPI.Host, path)
+	endpoint, err := generateKubernetesURL(kubeAPI.Host, path)
```
siggy: `generateKubernetesURL` is only called from `generateKubernetesAPIURLFor` and `getRequest`. looking at the callers to `getRequest`, they are constructing URLs of the form `"/api/v1/namespaces/"+namespace...` (with one exception for `/version`). what do you think about removing `generateKubernetesURL`, doing all `/api/v1/namespaces/` concatenating in one place, and then implementing a separate method for `/version`? something like:

```diff
diff --git a/pkg/k8s/api.go b/pkg/k8s/api.go
index 9f02baf3..a3243641 100644
--- a/pkg/k8s/api.go
+++ b/pkg/k8s/api.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/version"
@@ -38,7 +39,7 @@ func (kubeAPI *KubernetesAPI) NewClient() (*http.Client, error) {
 
 // GetVersionInfo returns version.Info for the Kubernetes cluster.
 func (kubeAPI *KubernetesAPI) GetVersionInfo(ctx context.Context, client *http.Client) (*version.Info, error) {
-	rsp, err := kubeAPI.getRequest(ctx, client, "/version")
+	rsp, err := kubeAPI.version(ctx, client)
 	if err != nil {
 		return nil, err
 	}
@@ -77,7 +78,7 @@ func (kubeAPI *KubernetesAPI) CheckVersion(versionInfo *version.Info) error {
 
 // NamespaceExists validates whether a given namespace exists.
 func (kubeAPI *KubernetesAPI) NamespaceExists(ctx context.Context, client *http.Client, namespace string) (bool, error) {
-	rsp, err := kubeAPI.getRequest(ctx, client, "/api/v1/namespaces/"+namespace)
+	rsp, err := kubeAPI.getRequest(ctx, client, namespace, "")
 	if err != nil {
 		return false, err
 	}
@@ -92,11 +93,11 @@ func (kubeAPI *KubernetesAPI) NamespaceExists(ctx context.Context, client *http.
 
 // GetPodsByNamespace returns all pods in a given namespace
 func (kubeAPI *KubernetesAPI) GetPodsByNamespace(ctx context.Context, client *http.Client, namespace string) ([]v1.Pod, error) {
-	return kubeAPI.getPods(ctx, client, "/api/v1/namespaces/"+namespace+"/pods")
+	return kubeAPI.getPods(ctx, client, namespace)
 }
 
-func (kubeAPI *KubernetesAPI) getPods(ctx context.Context, client *http.Client, path string) ([]v1.Pod, error) {
-	rsp, err := kubeAPI.getRequest(ctx, client, path)
+func (kubeAPI *KubernetesAPI) getPods(ctx context.Context, client *http.Client, namespace string) ([]v1.Pod, error) {
+	rsp, err := kubeAPI.getRequest(ctx, client, namespace, "/pods")
 	if err != nil {
 		return nil, err
 	}
@@ -125,8 +126,22 @@ func (kubeAPI *KubernetesAPI) URLFor(namespace, path string) (*url.URL, error) {
 	return generateKubernetesAPIURLFor(kubeAPI.Host, namespace, path)
 }
 
-func (kubeAPI *KubernetesAPI) getRequest(ctx context.Context, client *http.Client, path string) (*http.Response, error) {
-	endpoint, err := generateKubernetesURL(kubeAPI.Host, path)
+func (kubeAPI *KubernetesAPI) getRequest(ctx context.Context, client *http.Client, namespace, path string) (*http.Response, error) {
+	endpoint, err := generateKubernetesAPIURLFor(kubeAPI.Host, namespace, path)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest("GET", endpoint.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.Do(req.WithContext(ctx))
+}
+
+func (kubeAPI *KubernetesAPI) version(ctx context.Context, client *http.Client) (*http.Response, error) {
+	endpoint, err := url.Parse(strings.TrimSuffix(kubeAPI.Host, "/") + "/version")
 	if err != nil {
 		return nil, err
 	}
diff --git a/pkg/k8s/k8s.go b/pkg/k8s/k8s.go
index 4de44f37..e5457782 100644
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -56,20 +56,12 @@ var StatAllResourceTypes = []string{
 }
 
 func generateKubernetesAPIURLFor(serverURL, namespace, path string) (*url.URL, error) {
-	if !strings.HasPrefix(path, "/") {
-		return nil, fmt.Errorf("path must start with a /, got [%s]", path)
+	if path != "" && !strings.HasPrefix(path, "/") {
+		return nil, fmt.Errorf("path must be empty or start with a /, got [%s]", path)
 	}
 
 	fullPath := "/api/v1/namespaces/" + namespace + path
-	return generateKubernetesURL(serverURL, fullPath)
-}
-
-func generateKubernetesURL(serverURL, path string) (*url.URL, error) {
-	if !strings.HasPrefix(path, "/") {
-		return nil, fmt.Errorf("path must start with a /, got [%s]", path)
-	}
-
-	return url.Parse(strings.TrimSuffix(serverURL, "/") + path)
+	return url.Parse(strings.TrimSuffix(serverURL, "/") + fullPath)
 }
 
 // GetConfig returns kubernetes config based on the current environment.
diff --git a/pkg/k8s/k8s_test.go b/pkg/k8s/k8s_test.go
index 3a51035f..10c72c73 100644
--- a/pkg/k8s/k8s_test.go
+++ b/pkg/k8s/k8s_test.go
@@ -33,31 +33,6 @@ func TestGenerateKubernetesApiBaseUrlFor(t *testing.T) {
 	})
 }
 
-func TestGenerateBaseKubernetesUrl(t *testing.T) {
-	t.Run("Generates correct URL when all elements are present", func(t *testing.T) {
-		serverURL := "gopher://some-server.example.com:661"
-
-		url, err := generateKubernetesURL(serverURL, "/api/v1/")
-		if err != nil {
-			t.Fatalf("Unexpected error starting proxy: %v", err)
-		}
-
-		expectedURLString := "gopher://some-server.example.com:661/api/v1/"
-		if url.String() != expectedURLString {
-			t.Fatalf("Expected generated URl to be [%s], but got [%s]", expectedURLString, url.String())
-		}
-	})
-
-	t.Run("Return error if invalid host and port", func(t *testing.T) {
-		serverURL := "ftp://some-server.exampl     e.com:666"
-		_, err := generateKubernetesURL(serverURL, "/api/v1/")
-
-		if err == nil {
-			t.Fatalf("Expected error when trying to generate URL with extra path without leading slash, got nothing")
-		}
-	})
-}
-
 func TestGetConfig(t *testing.T) {
 	t.Run("Gets host correctly form existing file", func(t *testing.T) {
 		config, err := GetConfig("testdata/config.test", "")
```

klingerf: I'm ok with having `generateKubernetesAPIURLFor` and `generateKubernetesURL` co-exist in the code base, since `generateKubernetesAPIURLFor` wraps `generateKubernetesURL` and layers on additional functionality.

But I agree that hand-building API urls when `generateKubernetesAPIURLFor` could do that for us feels like a waste. So I could see the value in refactoring `getRequest` to accept a fully-formed URL, allowing the callers to first call `generateKubernetesAPIURLFor` and `generateKubernetesURL` as needed, but that feels like a bigger refactor, and I'm not sure it buys us much. I'm also a bit nervous updating `generateKubernetesAPIURLFor` to allow empty paths, since I feel like that could have unintended consequences. So I'm going to leave this as is for now, but we could consider a larger refactor later.

PR #2306, opened 2019-02-16
[WIP] Implement Resource limit

### LINKED ISSUE #2029: Add option to inject to set resource limits

TwinProduction:
Closes #2029: Add option to inject to set resource limits

This is also a breaking change, because the flag `proxy-cpu` has been renamed to `proxy-cpu-request` and the flag `proxy-memory` to `proxy-memory-request`. 
That change was necessary to clearly differentiate request from limit, since there are now two new flags, `proxy-cpu-limit` and `proxy-memory-limit`.


TwinProduction:
I did a oopsie

PR #2307, opened 2019-02-16
Implement resource limit

### LINKED ISSUE #2029: Add option to inject to set resource limits

TwinProduction:
Closes #2029: Add flags to the `inject` subcommand for setting resource limits

This is also a breaking change, because the flag `proxy-cpu` has been renamed to `proxy-cpu-request` and the flag `proxy-memory` to `proxy-memory-request`. 
That change was necessary to clearly differentiate request from limit, since there are now two new flags, `proxy-cpu-limit` and `proxy-memory-limit`.


TwinProduction:
Depending on what gets merged first, #2291 is impacted, more specifically `ResourceRequirements` in `config.proto`

cli/cmd/root.go
```diff
@@ -368,8 +384,10 @@ func addProxyConfigFlags(cmd *cobra.Command, options *proxyConfigOptions) {
...
 	cmd.PersistentFlags().UintVar(&options.proxyMetricsPort, "metrics-port", options.proxyMetricsPort, "Proxy port to serve metrics on")
-	cmd.PersistentFlags().StringVar(&options.proxyCPURequest, "proxy-cpu", options.proxyCPURequest, "Amount of CPU units that the proxy sidecar requests")
-	cmd.PersistentFlags().StringVar(&options.proxyMemoryRequest, "proxy-memory", options.proxyMemoryRequest, "Amount of Memory that the proxy sidecar requests")
+	cmd.PersistentFlags().StringVar(&options.proxyCPURequest, "proxy-cpu-request", options.proxyCPURequest, "Amount of CPU units that the proxy sidecar requests")
+	cmd.PersistentFlags().StringVar(&options.proxyMemoryRequest, "proxy-memory-request", options.proxyMemoryRequest, "Amount of Memory that the proxy sidecar requests")
+	cmd.PersistentFlags().StringVar(&options.proxyCPULimit, "proxy-cpu-limit", options.proxyCPULimit, "Maximum amount of CPU units that the proxy sidecar can use")
+	cmd.PersistentFlags().StringVar(&options.proxyMemoryLimit, "proxy-memory-limit", options.proxyMemoryLimit, "Maximum amount of Memory that the proxy sidecar can use")
```
alpeb: Those descriptions look good :+1: 
I agree with the flags name changes, but could you leave the previous options `proxy-cpu` and `proxy-memory` for backwards compatibility (with a deprecation message in the description)? And then you can use `cmd.PersistentFlags().MarkHidden` to hide them from the help screen.
Both old and new flags can point to `&options.proxyCPURequest` and `&options.proxyMemoryRequest` and it should work.

cli/cmd/root.go
```diff
@@ -314,13 +318,25 @@ func (options *proxyConfigOptions) validate() error {
...
+		if _, err := k8sResource.ParseQuantity(options.proxyCPULimit); err != nil {
+			return fmt.Errorf("Invalid cpu limit '%s' for --proxy-cpu-limit flag", options.proxyCPULimit)
+		}
+	}
+
+	if options.proxyMemoryLimit != "" {
+		if _, err := k8sResource.ParseQuantity(options.proxyMemoryLimit); err != nil {
+			return fmt.Errorf("Invalid memory limit '%s' for --proxy-memory-limit flag", options.proxyMemoryLimit)
 		}
 	}
```
alpeb: To be more bulletproof, it would be great if you could add a validation to check requests are not greater than limits, if both are set.

alpeb:
It would also be great if you could update the test that has `requests` entries to also include `limits`.
That'd mean adding the limits after this:
https://github.com/linkerd/linkerd2/blob/98c2686c9575e120363bca3418e24b0b719627b0/cli/cmd/inject_test.go#L61-L64
and updating `/cli/cmd/testdata/inject_emojivoto_pod_with_requests.golden.yml`, if I'm not mistaking.

cli/cmd/root.go
TwinProduction: Fixed with a0083b6, and I didn't know cobra had that feature. Pretty handy.

cli/cmd/root.go
TwinProduction: Fixed with 35eb8fe2743ad6c46eb235f1b301715ee1b9ea11, let me know what you think

TwinProduction:
@alpeb I added the tests as you suggested

cli/cmd/root.go
```diff
@@ -314,13 +318,39 @@ func (options *proxyConfigOptions) validate() error {
...
+	}
+
+	if options.proxyCPULimit != "" {
+		cpuLimit, err := k8sResource.ParseQuantity(options.proxyCPULimit)
+		if err != nil {
+			return fmt.Errorf("Invalid cpu limit '%s' for --proxy-cpu-limit flag", options.proxyCPULimit)
+		}
+		if options.proxyCPURequest != "" {
+			// Not checking for error because option proxyCPURequest was already validated
+			if cpuRequest, _ := k8sResource.ParseQuantity(options.proxyCPURequest); cpuRequest.Value() > cpuLimit.Value() {
```
alpeb: Unfortunately `.Value()` doesn't catch problems here because it rounds values like `100m` to `1`. So something like this won't be caught:
` inject --proxy-cpu-request 200m --proxy-cpu-limit 100m`

You want to use `.MilliValue()` instead for doing the comparison.

TwinProduction: Oops, my bad.

Fixed.

cli/cmd/root.go
```diff
@@ -314,13 +318,39 @@ func (options *proxyConfigOptions) validate() error {
...
+	}
+
+	if options.proxyMemoryLimit != "" {
+		memoryLimit, err := k8sResource.ParseQuantity(options.proxyMemoryLimit)
+		if err != nil {
+			return fmt.Errorf("Invalid memory limit '%s' for --proxy-memory-limit flag", options.proxyMemoryLimit)
+		}
+		if options.proxyMemoryRequest != "" {
+			// Not checking for error because option proxyMemoryRequest was already validated
+			if memoryRequest, _ := k8sResource.ParseQuantity(options.proxyMemoryRequest); memoryRequest.MilliValue() > memoryLimit.MilliValue() {
```
alpeb: Sorry I wasn't clearer on my previous comment. CPU values are fractions of 1 (like 100m means 10% of one CPU core) so we should use `MilliValue`. Memory on the other hand is an integer, in bytes. So we can still use `Value` for memory, while saving us some cycles.

TwinProduction: Though I have no problem making that change, isn't it better to opt for consistency for something that would save an insignificant amount of cycles?

Regardless, fixed with 35a4d13

alpeb:
Thanks again for all your work. This will make it into today's Edge release :+1:

Issue #2154, opened 2019-01-25
Break up install templates by component


kleimkuhler:
Currently, the install [templates](https://github.com/linkerd/linkerd2/blob/master/cli/install/template.go) are monolithic. In order to progress towards a more modular template rendering process during `install`, it will be beneficial to work on smaller components.

PR #2313, opened 2019-02-17
Split install template into component-specific files

olix0r:
chart/templates/base.yaml is nearly 800 lines and contains the
kubernetes configurations for the marjority of the control plane.
Furthermore, its contents are not particularly organized (for example,
the prometheus RBAC bindings are in the middle of the controller's
configuration).

The size and complexity of this file makes it especially daunting to
introduce new functionality.

In order to make the situation easier to understand and change, this
splits base.yaml into several new template files: namespace, controller,
serviceprofile, and prometheus, and grafana. The `tls.yaml` template has
been renamed `ca.yaml`, since it installs the `linkerd-ca` resources.

This change also makes the comments uniform, adding a "header" to each
logical component.


alpeb:
Fixes #2154

cli/cmd/install.go
```diff
@@ -286,13 +316,35 @@ func render(config installConfig, w io.Writer, options *installOptions) error {
 
 	// Merge templates and inject
 	var buf bytes.Buffer
-	bt := path.Join(renderOpts.ReleaseOptions.Name, baseTemplateName)
-	if _, err := buf.WriteString(renderedTemplates[bt]); err != nil {
+	t := path.Join(renderOpts.ReleaseOptions.Name, nsTemplateName)
```
olix0r: Should this be guarded by `if !config.SingleNamespace` ?  (I'm not really sure what this does or when/why it should be skipped.

kleimkuhler: This is already guarded by the template syntax `{{ if not .Values.SingleNamespace -}}` in `namespace.yaml`. If `SingleNamespace` is not set, then that template will not be rendered.

PR #2315, opened 2019-02-18
test: Overwrite .golden files with actual output

olix0r:
When changing the installation templates, it's can be pretty
time-intensive to repair the .golden output files.

This change modifies the install tests to overwrite each .golden file
with the generated output when the `INSTALL_TEST_OVERWRITE_GOLDEN`
environment variable is set.


grampelberg:
Mind adding this to `BUILD.md`? I'm sure to forget it exists.

cli/cmd/inject_test.go
```diff
@@ -38,6 +38,11 @@ func testUninjectAndInject(t *testing.T, tc injectYAML) {
 	if expectedOutput != actualOutput {
 		t.Errorf("Result mismatch.\nExpected: %s\nActual: %s", expectedOutput, actualOutput)
 	}
+	if actualOutput != expectedOutput && updateFixtures {
```
klingerf: Thanks for adding the inject fixtures. There are multiple tests in this file, and for this flag to be effective, we should update all of them. It might be worth refactoring the fixture logic a bit so that we can do the rewrites in one place.

BUILD.md
```diff
@@ -434,3 +434,14 @@ build_architecture
...
+
+# Updating install templates
+
+Install configuration is stored in `chart/templates/*.yaml` as HELM-templated
+YAML. When these files are changed, tests will fail until the files in
+`cli/cmd/testdata/install_*.golden` are updated to reflect the change. These
+golden files can be automatically regenerated with the command:
+
+```sh
+go test ./cli/... -args -update
```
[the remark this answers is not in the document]
olix0r: I can only get this to work with the `-args`

olix0r:
OK, I've updated this, cleaning up inject test util code; also, i've replaced manual error messages with `diffCompare` in the inject test.

Finally, i've tested that modifying the inject logic produces the proper results.

cli/cmd/inject_test.go
```diff
@@ -34,19 +34,21 @@ func testUninjectAndInject(t *testing.T, tc injectYAML) {
 	}
 
 	actualOutput := output.String()
-	expectedOutput := readOptionalTestFile(t, tc.goldenFileName)
-	if expectedOutput != actualOutput {
-		t.Errorf("Result mismatch.\nExpected: %s\nActual: %s", expectedOutput, actualOutput)
+	expectedOutput := readTestdata(t, tc.goldenFileName)
+	if actualOutput != expectedOutput {
+		writeTestdataIfUpdate(t, tc.goldenFileName, output.Bytes())
+		diffCompare(t, expectedOutput, actualOutput)
```
klingerf: Method signature for `diffCompare` is:

```
func diffCompare(t *testing.T, actual string, expected string) {
```

So this should be
```suggestion
		diffCompare(t, actualOutput, expectedOutput)
```

In order for the diff to be colored correctly. Same goes for the rest of the `diffCompare` calls in this file.

cli/cmd/inject_test.go
```diff
@@ -34,19 +34,21 @@ func testUninjectAndInject(t *testing.T, tc injectYAML) {
...
 	reportFileName := tc.reportFileName
 	if verbose {
 		reportFileName += ".verbose"
 	}
-	expectedReport := readOptionalTestFile(t, reportFileName)
+	expectedReport := readTestdata(t, reportFileName)
 	if expectedReport != actualReport {
-		t.Errorf("Result mismatch.\nExpected: %s\nActual: %s", expectedReport, actualReport)
+		writeTestdataIfUpdate(t, reportFileName, report.Bytes())
+		diffCompare(t, expectedReport, actualReport)
```
olix0r: blegh, need to update these signatures

cli/cmd/inject_test.go
```diff
@@ -294,19 +299,22 @@ func testReadFromFolder(t *testing.T, resourceFolder string, expectedFolder stri
 		t.Fatal("Unexpected error. Exit code from runInjectCmd: ", exitCode)
 	}
 
-	expected := readOptionalTestFile(t, filepath.Join(expectedFolder, "injected_nginx_redis.yaml"))
+	expectedFile := filepath.Join(expectedFolder, "injected_nginx_redis.yaml")
+	expected :, expectedFile= readTestdata(t)
```
klingerf: Typo here -- looks like it should be:

```suggestion
	expected := readTestdata(t, expectedFile)
```

But that causes a different compilation error on 305, in the call to `writeTestdataIfUpdate`.

olix0r: i thought i would very cleverly reorder the arguments with one cool vscode command ;( thanks for catching

olix0r:
@klingerf I decided to go the other way and use the compact diff syntax by default and enable the long output via flag/env... having spent the past weekend pouring over 10K lines of terminal text trying to find misplaced spaces --- the color diffs are becoming less helpful when we have so much output.  The compact diffs, on the other hand, at least give clear coordinates into the golden file.

cli/cmd/inject_test.go
```diff
@@ -201,25 +192,17 @@ func testInjectCmd(t *testing.T, tc injectCmd) {
...
 		t.Fatalf("Expected exit code to be %d but got: %d", tc.exitCode, exitCode)
 	}
-
-	actualStdOutResult := outBuffer.String()
-	expectedStdOutResult := readOptionalTestFile(t, tc.stdOutGoldenFileName)
-	if expectedStdOutResult != actualStdOutResult {
-		t.Errorf("Result mismatch.\nExpected: %s\nActual: %s", expectedStdOutResult, actualStdOutResult)
+	if tc.stdOutGoldenFileName != "" {
+		testDiff(t, tc.stdOutGoldenFileName, outBuffer.String())
 	}
```
klingerf: For parity with the previous implementation, this should assert the output buffer is empty if no golden output file is given. Something like:

```diff
diff --git a/cli/cmd/inject_test.go b/cli/cmd/inject_test.go
index def9298b..76126848 100644
--- a/cli/cmd/inject_test.go
+++ b/cli/cmd/inject_test.go
@@ -194,6 +194,8 @@ func testInjectCmd(t *testing.T, tc injectCmd) {
 	}
 	if tc.stdOutGoldenFileName != "" {
 		testDiff(t, tc.stdOutGoldenFileName, outBuffer.String())
+	} else if outBuffer.Len() != 0 {
+		t.Fatalf("Expected no standard output, but got: %s", outBuffer)
 	}
 
 	stdErrGoldenFileName := tc.stdErrGoldenFileName
```

cli/cmd/test_helper.go
```diff
@@ -1,54 +1,62 @@
...
-		t.Fatalf("Unexpected output:\n%+v", dmp.DiffPrettyText(diffs))
+	// prettyDiff is set by the `-verbose-diff` flag.
+	prettyDiff bool
+)
 
-		diffs = dmp.DiffCleanupSemantic(diffs)
-		patches := dmp.PatchMake(diffs)
-		patchText := dmp.PatchToText(patches)
-		t.Fatalf("Unexpected output:\n%+v", patchText)
+// readTesdtata reads a file and return the contents of that file as a string.
```
klingerf: Typo here: `return` => `returns`

PR #2317, opened 2019-02-18
Add discovery package for managing discovery API

klingerf:
This is a follow-up to #2195, which introduced the discovery API for the public and proxy APIs, and #2281, which renamed the proxy API to the destination API. The original goal of this change was to update the public API package to stop importing the destination API package, by introducing a shared discovery API package on which they can both depend. Here's a full list of changes in this branch:

* Add a `controller/api/discovery` package for building the discovery client and providing test helpers
* Update public API main and discovery-client main to use the client builder from the discovery package
* Rename all imports of the discovery protobuf to `pb` or `discoveryPb`, which is more consistent with how we handle `tapPb` and `healthcheckPb`
* Update public API tests to use new discovery mock client, instead of using the client builder from the destination package
* Remove the `InitFakeDiscoveryServer` helper from the destination package, since it is only used in one test now
* Fix a bunch of remaining references to the proxy API in comments and variables

### DIFF
--- a/cli/cmd/endpoints_test.go
+++ b/cli/cmd/endpoints_test.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"testing"
 
+	"github.com/linkerd/linkerd2/controller/api/discovery"
 	"github.com/linkerd/linkerd2/controller/api/public"
 )
 
@@ -52,11 +53,11 @@ func TestEndpoints(t *testing.T) {
 }
 
 func testEndpointsCall(exp endpointsExp, t *testing.T) {
-	mockClient := &public.MockAPIClient{}
-
-	response := public.GenEndpointsResponse(exp.identities)
-
-	mockClient.EndpointsResponseToReturn = &response
+	mockClient := &public.MockAPIClient{
+		MockDiscoveryClient: &discovery.MockDiscoveryClient{
+			EndpointsResponseToReturn: discovery.GenEndpointsResponse(exp.identities),
+		},
+	}
 
 	endpoints, err := requestEndpointsFromAPI(mockClient)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


cli/cmd/endpoints_test.go
```diff
@@ -52,11 +53,11 @@ func TestEndpoints(t *testing.T) {
...
 
 func testEndpointsCall(exp endpointsExp, t *testing.T) {
-	mockClient := &public.MockAPIClient{}
-
-	response := public.GenEndpointsResponse(exp.identities)
-
-	mockClient.EndpointsResponseToReturn = &response
+	mockClient := &public.MockAPIClient{
+		MockDiscoveryClient: &discovery.MockDiscoveryClient{
+			EndpointsResponseToReturn: discovery.GenEndpointsResponse(exp.identities),
```
siggy: 👍

controller/api/destination/server.go
```diff
@@ -63,9 +63,9 @@ func NewServer(
 
 	// this server satisfies 2 gRPC interfaces:
 	// 1) linkerd2-proxy-api/destination.Destination (proxy-facing)
-	// 2) controller/discovery.Api (controller-facing)
+	// 2) controller/discovery.Destination (controller-facing)
```
siggy: `controller/discovery.Discovery`

controller/api/public/grpc_server_test.go
```diff
@@ -10,9 +10,8 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/duration"
-	"github.com/linkerd/linkerd2/controller/api/destination"
```
siggy: is this the change intended to `update the public API package to stop importing the destination API package, by introducing a shared discovery API package on which they can both depend`?

klingerf: That's right. My overall goal was to stop having this test file call `destination.InitFakeDiscoveryServer`, and I did that by way of introducing `discovery.MockDiscoveryClient` instead.

PR #2319, opened 2019-02-18
CLI help updates: non-experimental auto-inject

klingerf:
A few tweaks to our CLI help text, as follows:

* Removes the "experimental" text from the `--proxy-auto-inject` install flag (fixes #2311)
* ~~Unhides the `linkerd install-cni` command and `--linkerd-cni-enabled` flag and ensures that they're labeled as experimental~~
* Rewords the short and long descriptions for the `linkerd completion` command

PR #2321, opened 2019-02-18
CLI help updates: unhide install-cni

klingerf:
This is just the CNI changes that were split out from #2319.

### DIFF
--- a/cli/cmd/install-cni-plugin.go
+++ b/cli/cmd/install-cni-plugin.go
@@ -90,15 +90,14 @@ func newCmdInstallCNIPlugin() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "install-cni [flags]",
-		Short: "Output Kubernetes configs to install the Linkerd CNI Plugin",
-		Long: `Output Kubernetes configs to install the Linkerd CNI Plugin."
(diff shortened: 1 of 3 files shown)


cli/cmd/install-cni-plugin.go
```diff
@@ -90,15 +90,14 @@ func newCmdInstallCNIPlugin() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "install-cni [flags]",
-		Short: "Output Kubernetes configs to install the Linkerd CNI Plugin",
-		Long: `Output Kubernetes configs to install the Linkerd CNI Plugin."
+		Short: "Output Kubernetes configs to install the Linkerd CNI Plugin (experimental)",
```
siggy: i don't know how much we care about this, but this makes `install-cni` the only command description that prints beyond 80 chars when you type `linkerd`:

screen shot 2019-02-18 at 3 28 27 pm

cli/cmd/install-cni-plugin.go
```diff
@@ -90,15 +90,14 @@ func newCmdInstallCNIPlugin() *cobra.Command {
...
+		Long: `Output Kubernetes configs to install the Linkerd CNI Plugin (experimental)."
```
siggy: this was an existing nit, but mind removing the trailing `"`?

Issue #2289, opened 2019-02-14
Refactor the auto-injector to make use of the new shared lib for injecting the proxy


alpeb:
Depends on #2286

alpeb:
Closed via #2334

PR #2334, opened 2019-02-20
Injection consolidation

alpeb:
I'm pushing first the changes related to `patch.go`:

- Created the pkg/inject package to hold the new injection shared lib.
- Moved patch.go there.
- Added the ability to add pod labels and annotations without having to
specify the already existing ones. For that purpose, added `patch.addPodLabelsRoot()` and `addPodAnnotationsRoot()`, and repurposed `addPodLabels()` into `addPodLabel()`, and `addPodAnnotations()` into  `addPodAnnotation()`
- Removed `addDeploymentLabels()` because I couldn't find the need for
it. At least `linkerd inject` wasn't setting those labels. Please
confirm in case I'm wrong.

Still to-do: have `patch_test.go` depend on fixtures separate from the
ones under `controller/proxy-injector`

Ref #1748 and #2289


klingerf:
Looks like the functionality from pkg/inject/patch.go is still used in controller/proxy-injector/webhook.go, so you'll need to update all of the `(p *Patch)` functions in patch.go to be public, and then update webhook.go to work with the public package.

alpeb:
@klingerf the patching has been encapsulated under `/pkg/inject` which will be the shared lib so `webhook.go` won't have to deal with it. I'll push all that in a few hours :wink: :sweat_smile:

alpeb:
`linkerd inject`, `linkerd uninject` and auto-injection are passing the smoke tests with this new approach. I still need to fix all the unit tests though :persevere: 

Here's a detailed list of the changes:

- Removed linkerd-proxy-injector-sidecar-config ConfigMap and the corresponding volumes in the linkerd-proxy-injector deployment
 - Extracted from `/cli/cmd/inject.go` and `/cli/cmd/inject_util.go` the core methods doing the workload parsing and injection, and moved them into `/pkg/inject/inject.go`. The CLI files should now deal only with strictly CLI concerns, and applying the json patch returned by the new lib.
- Proceeded analogously with `/cli/cmd/uninject.go` and `/pkg/inject/uninject.go`.
- Refactored the webhook to use the new injection lib.
- The `InjectReport` struct and helping methods were moved into `/pkg/inject/report.go`
    
The only network call `/pkg/inject/inject.go` does is inside `shouldInject()` where it retrieves the namespace annotations from kubernetes. I think we need to review how that client is created in there.
    
Caveats:
    
- Currently passing aroung GlobalConfig and ProxyConfig and using their GRPC methods directly, until we come up with an abstraction for all that.
- The webhook has those configs hard-coded for the moment.
- I removed support for parsing yamls with List containing multiple workloads because it was not an easy refactor given the recursion. I'll re-add that later on.
- Todo: fix all the tests.
 - I've made a note to remove later on the bad v1 import alias that VSCode is adding.

Eager to hear your feedback! :pray:

cli/cmd/inject.go
```diff
@@ -509,18 +227,36 @@ func (resourceTransformerInject) generateReport(injectReports []injectReport, ou
...
-		}
-	}
-	return false
-}
-
-func injectDisabled(t *metaV1.ObjectMeta) bool {
-	return t.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled
+// TODO: this is just a temporary function to convert command-line options to GlobalConfig
+// and ProxyConfig, until we come up with an abstraction over those GRPC structs
+func injectOptionsToConfigs(options *injectOptions) (*pb.GlobalConfig, *pb.ProxyConfig) {
```
ihcsim: :+1: We can remove this once #2344 is merged.

pkg/inject/report.go
```diff
@@ -0,0 +1,32 @@
+package inject
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Report contains the Kind and Name for a given workload along with booleans
+// describing the result of the injection transformation
+type Report struct {
```
ihcsim: Is there a way to consolidate the report here and the `generateReport()` in the CLI? Do they serve different purposes?

alpeb: `Report` captures the common functionality leveraged by the implementations of `generateReport()` that are different for cli inject and cli uninject. That's why I put `Report` in the lib, and the `generateReport()` functions in their respective cli files.

pkg/inject/inject.go
```diff
@@ -0,0 +1,571 @@
...
+// objMeta provides a generic struct to parse the names of Kubernetes objects
+type objMeta struct {
+	metaV1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+}
+
+// ResourceConfig contains both the raw (bytes) and parsed information for a given workload
+type ResourceConfig struct {
+	bytes           []byte
+	obj             interface{}
+	om              objMeta
```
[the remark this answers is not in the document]
alpeb: Good catch. This is the result from the refactoring I did a few weeks ago when I introduced `uninject`. I think I should be able to merge these two.

pkg/inject/inject.go
```diff
@@ -0,0 +1,571 @@
...
+	meta            metaV1.TypeMeta
+	podSpec         *v1.PodSpec
+	objectMeta      *metaV1.ObjectMeta
+	dnsNameOverride string
+	k8sLabels       map[string]string
+}
+
+// NewResourceConfig creates and initializes a ResourceConfig
+func NewResourceConfig(bytes []byte, request *admissionv1beta1.AdmissionRequest) (*ResourceConfig, error) {
```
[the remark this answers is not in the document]
alpeb: Right, `AdmissionRequest` from the webhook contains the metadata (kind, name, namespace) separated from the workload bytes.
I agree that function signature didn't feel too kosher. Definitely better to split this in two. I'll see how I can incorporate your suggestions :+1:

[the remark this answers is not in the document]
ihcsim: I forgot to mention that I notice the `globalConfig` and `proxyConfig` are being passed into a few methods, so perhaps they can be stored as the struct's attributes. The `ResourceConfig` gets re-created during CLI install and auto-inject anyway, so I don't think the configs will become stale.

pkg/inject/inject.go
```diff
@@ -0,0 +1,571 @@
...
+	return nil
+}
+
+// Given a PodSpec, update the PodSpec in place with the sidecar
+// and init-container injected. If the pod is unsuitable for having them
+// injected, return false.
+func injectPodSpec(t *v1.PodSpec, patch *Patch, identity k8s.TLSIdentity, controlPlaneDNSNameOverride string, globalConfig *pb.GlobalConfig, proxyConfig *pb.ProxyConfig, report *Report) bool {
+	report.HostNetwork = t.HostNetwork
+	report.Sidecar = healthcheck.HasExistingSidecars(t)
+	report.Udp = checkUDPPorts(t)
```
ihcsim: TIOLI: I think the report-related code don't need to be here. Similar observation for `injectObjectMeta()` and `uninjectPodSpec()`. If all these report code are moved to the parent `Transform()` and `Uninject()` methods, we can potentially create inject-specific and uninject-specific reports in the `Report` struct, and returned them to the CLI. Hence, trimming down the current `generateReport()` functions.

pkg/inject/inject.go
```diff
@@ -0,0 +1,571 @@
...
+// YamlMarshalObj returns the yaml for the workload in conf
+func (conf *ResourceConfig) YamlMarshalObj() ([]byte, error) {
+	return yaml.Marshal(conf.obj)
+}
+
+// Transform parses conf.bytes and (if pertinent) returns the json patch
+// for injecting the init and proxy containers and all the necessary metadata.
+// It also returns a Report with the results of the injection.
+// Note that conf.bytes should only contain a single YAML (not multiple separated by ---)
+func (conf *ResourceConfig) Transform(globalConfig *pb.GlobalConfig, proxyConfig *pb.ProxyConfig) ([]byte, []Report, error) {
```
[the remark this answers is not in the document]
alpeb: `Uninject()` does indeed return the uninjected yaml. `Transform()` returns the json patch for the callers to apply it. What do you think about calling it `getInjectPatch()`?

cli/cmd/inject.go
```diff
@@ -96,7 +78,9 @@ sub-folders, or coming from stdin.`,
 				return err
 			}
 
-			exitCode := uninjectAndInject(in, stderr, stdout, options)
+			globalConfig, proxyConfig := injectOptionsToConfigs(options)
```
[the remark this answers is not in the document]
alpeb: I assumed that network (and the merging with cli flags) call would happen before these functions get called, right before this snippet. Do you have a different place in mind?

pkg/inject/inject.go
```diff
@@ -0,0 +1,571 @@
...
+		patch.addInitContainer(initContainer)
+	}
+
+	return true
+}
+
+// Given a ObjectMeta, update ObjectMeta in place with the new labels and
+// annotations.
+func (conf *ResourceConfig) injectObjectMeta(t *metaV1.ObjectMeta, patch *Patch, globalConfig *pb.GlobalConfig, proxyConfig *pb.ProxyConfig, report *Report) error {
+	res, err := conf.shouldInject()
```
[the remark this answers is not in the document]
alpeb: In the previous implementation the cli was checking for the `linkerd.io/inject` annotation and it also checked whether the workload was already injected, which now is done inside `shouldInject()`. Now the cli is also checking for the annotation at the namespace level, which I think doesn't hurt. We could introduce some kind of `-force` flag to ignore these from the cli.

pkg/inject/inject.go
alpeb: Agreed. It appears it wouldn't be to hard to extract that out :+1:

alpeb:
Last batch is in:

- Removed most of those extra config params in many functions by moving the config into `ResourceConfig`.
- Have the CLI and the webhook use different methods when using the shared lib.

Still pending: move the report code out of the lower-level inject methods

alpeb:
My last push rebases to #2344, and also:
- Added `Registry` to GlobalConfig; needed at injection time to honour the `--registry` flag
- Proxy and init images were being tagged in two different places

alpeb:
In the last push I Fix unit tests and:
- Applied #2379
- Added `Version` to GlobalConfig, to honour CLI flag
- Updated text fixtures
- Fixed uninjection and reporting issues
- Removed more vars/consts from the no longer used webhook pod spec template
- Removed a couple of tests from `webhook_test.go` that are no longer relevant
- Removed network dependency on `shouldInject()`

pkg/inject/inject.go
alpeb: This was partially addressed in #2379 and back-ported here

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+func (conf *ResourceConfig) YamlMarshalObj() ([]byte, error) {
+	return yaml.Marshal(conf.obj)
+}
+
+// PatchForYaml parses bytes and (if pertinent) returns the json patch
+// for injecting the init and proxy containers and all the necessary metadata.
+// It also returns a Report with the results of the injection.
+// Note that bytes should only contain a single YAML (not multiple separated by ---)
+func (conf *ResourceConfig) PatchForYaml(bytes []byte) ([]byte, []Report, error) {
+	if err := yaml.Unmarshal(bytes, &conf.meta); err != nil {
```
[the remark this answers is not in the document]
alpeb: `ParseMetaAndYaml()` takes a different path at the end, but still I'll extract the common code between `PatchForYaml()`  and `ParseMetaAndYaml()` in a separate function.

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+// PatchForAdmissionRequest parses request.ObjectRaw and (if pertinent) returns the json patch
+// for injecting the init and proxy containers and all the necessary metadata.
+func (conf *ResourceConfig) PatchForAdmissionRequest(request *admissionv1beta1.AdmissionRequest) ([]byte, error) {
+	conf.meta = metaV1.TypeMeta{Kind: request.Kind.Kind}
+	conf.objectMeta = &objMeta{metaV1.ObjectMeta{Name: request.Name, Namespace: request.Namespace}}
+	patch, _, err := conf.getPatch(request.Object.Raw)
+	return patch, err
+}
+
+func (conf *ResourceConfig) getPatch(bytes []byte) ([]byte, []Report, error) {
```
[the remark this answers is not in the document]
alpeb: Can you provide a bit more details? Not sure I got the intent of the suggestion...

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+	defaultKeepaliveMs    = 10000
+)
+
+// objMeta provides a generic struct to parse the names of Kubernetes objects
+type objMeta struct {
+	metaV1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+}
+
+// ResourceConfig contains the parsed information for a given workload
+type ResourceConfig struct {
```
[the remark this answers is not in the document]
alpeb: Given `globalConfig` and `proxyConfig` and the other fields in `ResourceConfig` are used all over `inject.go`, if we split them apart that'd mean the one that is no longer in the receiver will have to be passed around. Or did I misunderstand your point?

pkg/inject/inject.go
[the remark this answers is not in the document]
ihcsim: I think if possible, `pkg/inject` should only export two functions; one for inject and one for un-inject.

```
// takes a k8s resource YAML manifest and returns JSON proxy patch.
func ProxyPatch(yaml io.Reader, meta metaV1.ObjectMeta, rc *ResourceConfig) (patch io.Reader, error)

// takes a k8s resource YAML manifest and returns unmeshed YAML.
func Uninject(yaml io.Reader, meta metaV1.ObjectMeta) (uninjected io.Reader, error)
```

IIUC, the details in `PatchForYaml()` and `PatchForAdmissionRequest()` aren't necessary in this lib. For example, this lib shouldn't need to know about webhook-specifics like admission request. I think we can let webhook extract the `objectMeta` and `request.Object.Raw`.

pkg/inject/inject.go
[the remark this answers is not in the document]
ihcsim: Following my [comment](https://github.com/linkerd/linkerd2/pull/2334#discussion_r260481629) on the exported functions, if we also change `parse()` return values to include a struct that contains all the non-config fields, then we can simplify the `ResourceConfig` struct to just config-related fields and methods.

cli/cmd/inject.go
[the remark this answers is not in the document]
ihcsim: Per _slack_ discussion, the configs will be read by the proxy injector as a volume mount. The only network call that will happen here is to the proxy injector's new "inject" endpoint.

pkg/inject/inject.go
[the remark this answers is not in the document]
alpeb: I started working on a refactoring but it got tedious rapidly... For the moment I've just made some tiny vars renaming for things to be a little clearer.

pkg/inject/inject.go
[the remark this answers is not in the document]
alpeb: Ok I've moved the AdmissionRequest dependency from pkg/inject/inject.go, and both the webhook and CLI inject will consume `GetPatch` directly. Also note that pkg/inject/uninject.go only exposes `Uninject()`.

controller/proxy-injector/webhook.go
```diff
@@ -102,105 +92,40 @@ func (w *Webhook) decode(data []byte) (*admissionv1beta1.AdmissionReview, error)
 }
 
 func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admissionv1beta1.AdmissionResponse, error) {
-	var deployment appsv1.Deployment
-	if err := yaml.Unmarshal(request.Object.Raw, &deployment); err != nil {
-		return nil, err
-	}
-	log.Infof("working on %s/%s %s..", request.Kind.Version, strings.ToLower(request.Kind.Kind), deployment.ObjectMeta.Name)
+	log.Debugf("request object bytes: %v", string(request.Object.Raw))
```
siggy: you can use `%s` since you are converting to string.

controller/proxy-injector/webhook_test.go
```diff
@@ -2,30 +2,57 @@ package injector
...
+		OutboundPort:            &config.Port{Port: 4140},
+		Resource:                &config.ResourceRequirements{RequestCpu: "", RequestMemory: "", LimitCpu: "", LimitMemory: ""},
+		ProxyUid:                2102,
+		LogLevel:                &config.LogLevel{Level: "warn,linkerd2_proxy=info"},
+		DisableExternalProfiles: false,
+	}
 )
 
+func confNsEnabled() *inject.ResourceConfig {
+	return inject.NewResourceConfig(globalConfig, proxyConfig).WithNsAnnotations(map[string]string{"linkerd.io/inject": "enabled"})
```
siggy: `{ProxyInjectAnnotation: ProxyInjectEnabled}`

pkg/inject/inject.go
```diff
@@ -0,0 +1,541 @@
...
+	LocalhostDNSNameOverride = "localhost."
+	// ControlPlanePodName default control plane pod name.
+	ControlPlanePodName = "linkerd-controller"
+	// PodNamespaceEnvVarName is the name of the variable used to pass the pod's namespace.
+	PodNamespaceEnvVarName = "LINKERD2_PROXY_POD_NAMESPACE"
+	// PrometheusImage is the docker image and tag for the Prometheus instance used in the control plane
+	PrometheusImage                 = "prom/prometheus:v2.7.1"
+	prometheusProxyOutboundCapacity = 10000
+
+	defaultDockerRegistry = "gcr.io/linkerd-io"
```
siggy: mind sharing this with `defaultDockerRegistry` in `root.go`? same with `defaultKeepaliveMs`.

pkg/inject/inject.go
```diff
@@ -0,0 +1,541 @@
...
+// WithMeta enriches ResourceConfig with extra metadata
+func (conf *ResourceConfig) WithMeta(kind, namespace, name string) *ResourceConfig {
+	conf.meta = metav1.TypeMeta{Kind: kind}
+	conf.objMeta = objMeta{&metav1.ObjectMeta{Name: name, Namespace: namespace}}
+	return conf
+}
+
+// WithNsAnnotations enriches ResourceConfig with the namespace annotations, that can
+// be used in shouldInject()
+func (conf *ResourceConfig) WithNsAnnotations(m map[string]string) *ResourceConfig {
```
siggy: these are slick!

pkg/inject/inject.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+
+	// Special case if the caller specifies that
+	// LINKERD2_PROXY_OUTBOUND_ROUTER_CAPACITY be set on the pod.
+	// We key off of any container image in the pod. Ideally we would instead key
+	// off of something at the top-level of the PodSpec, but there is nothing
+	// easily identifiable at that level.
+	// This is currently only used by the Prometheus pod in the control-plane.
+	for _, container := range conf.podSpec.Containers {
+		if container.Image == PrometheusImage {
```
siggy: will this set `LINKERD2_PROXY_OUTBOUND_ROUTER_CAPACITY` into the control-plane's prometheus, or any prometheus pod matching this image name? it's probably ok if that's the case, but would be good to call out.

proto/config/config.proto
```diff
@@ -4,17 +4,19 @@ package linkerd2.config;
 
 option go_package = "github.com/linkerd/linkerd2/controller/gen/config";
 
-message GlobalConfig {
+message Global {
```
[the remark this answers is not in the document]
alpeb: We were using it just to do some serialization on 2.2.1, which persisted its JSON in a new ConfigMap, which should be oblivious to this naming change.

pkg/inject/inject.go
alpeb: Good catch. This will be set on any proxy that gets injected into any Prometheus (it is already like that in master). I'll modify the comment accordingly.

pkg/k8s/labels.go
```diff
@@ -105,14 +97,6 @@ const (
 	// when injecting a proxy with TLS enabled.
 	TLSSecretsVolumeName = "linkerd-secrets"
 
-	// TLSTrustAnchorVolumeSpecFileName is the name (key) within the
-	// proxy-injector ConfigMap that contains the trust anchors volume spec.
-	TLSTrustAnchorVolumeSpecFileName = "linkerd-trust-anchors.yaml"
```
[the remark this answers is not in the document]
alpeb: Correct.

alpeb:
@siggy Agreed on the benefits of keeping ordering. The culprit is this guy:
https://github.com/linkerd/linkerd2/blob/acc0a42bde4ae4f39f4db5bba9623d6dd039ab91/cli/cmd/inject.go#L122
I'll see if there's something we can do with that lib, or if a different lib is required.

alpeb:
I overrode `yaml.JSONToYAML()` with a function that preserves the field order of workloads as declared in the k8s api structs. As a result this branch diff was reduced substantially: 
image

I also simplified a bit the `inject.parse()` function.

It seems this branch is blacklisted from CI. @siggy could you please help me re-enabling it?

cli/cmd/root.go
```diff
@@ -305,6 +306,32 @@ func newProxyConfigOptions() *proxyConfigOptions {
...
+func newConfig() configs {
+	globalConfig := &config.Global{
+		LinkerdNamespace: defaultNamespace,
+		CniEnabled:       false,
+		Registry:         inject.DefaultDockerRegistry,
+		Version:          version.Version,
+		IdentityContext:  nil,
+	}
+	proxyConfig := &config.Proxy{
+		ProxyImage:              &config.Image{ImageName: inject.DefaultDockerRegistry + "/proxy", PullPolicy: "IfNotPresent"},
```
siggy: i'm not seeing docker image tags on the `linkerd-config` `proxy` obecjt. do these need to use `taggedProxyImage` and `taggedProxyInitImage`?

edge-19.2.5:

```yaml
  proxy: |
    {"proxyImage":{"imageName":"gcr.io/linkerd-io/proxy:edge-19.2.5",...
```

this branch:

```yaml
  proxy: |
    {"proxyImage":{"imageName":"gcr.io/linkerd-io/proxy",...
```

pkg/inject/inject.go
```diff
@@ -0,0 +1,585 @@
...
+	PrometheusImage                 = "prom/prometheus:v2.7.1"
+	prometheusProxyOutboundCapacity = 10000
+	// DefaultDockerRegistry is the default registry to pull the proxy and init container images from
+	DefaultDockerRegistry = "gcr.io/linkerd-io"
+	// DefaultKeepaliveMs is used in the proxy configuration for remote connections
+	DefaultKeepaliveMs = 10000
+)
+
+var injectableKinds = []string{"Deployment", "ReplicationController",
+	"ReplicaSet", "Job", "DaemonSet", "StatefulSet", "Pod"}
```
klingerf: Whoops, one more comment. Lowercase versions of all of these strings already exist in `pkg/k8s/k8s.go`, so I'd prefer to use those if possible. Something like this:

```diff
diff --git a/pkg/inject/inject.go b/pkg/inject/inject.go
index c64a2c47..8955bc31 100644
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -33,8 +33,15 @@ const (
 	defaultKeepaliveMs = 10000
 )
 
-var injectableKinds = []string{"Deployment", "ReplicationController",
-	"ReplicaSet", "Job", "DaemonSet", "StatefulSet", "Pod"}
+var injectableKinds = []string{
+	k8s.DaemonSet,
+	k8s.Deployment,
+	k8s.Job,
+	k8s.Pod,
+	k8s.ReplicaSet,
+	k8s.ReplicationController,
+	k8s.StatefulSet,
+}
 
 // objMeta provides a generic struct to parse the names of Kubernetes objects
 type objMeta struct {
@@ -120,7 +127,7 @@ func (conf *ResourceConfig) GetPatch(bytes []byte) ([]byte, []Report, error) {
 	}
 
 	var patch *Patch
-	if conf.meta.Kind == "Pod" {
+	if strings.ToLower(conf.meta.Kind) == k8s.Pod {
 		patch = NewPatchPod()
 	} else {
 		patch = NewPatchDeployment()
@@ -166,7 +173,7 @@ func (conf *ResourceConfig) GetPatch(bytes []byte) ([]byte, []Report, error) {
 // KindInjectable returns true if the resource in conf can be injected with a proxy
 func (conf *ResourceConfig) KindInjectable() bool {
 	for _, kind := range injectableKinds {
-		if conf.meta.Kind == kind {
+		if strings.ToLower(conf.meta.Kind) == kind {
 			return true
 		}
 	}
@@ -174,20 +181,20 @@ func (conf *ResourceConfig) KindInjectable() bool {
 }
 
 func (conf *ResourceConfig) getFreshWorkloadObj() runtime.Object {
-	switch conf.meta.Kind {
-	case "Deployment":
+	switch strings.ToLower(conf.meta.Kind) {
+	case k8s.Deployment:
 		return &v1beta1.Deployment{}
-	case "ReplicationController":
+	case k8s.ReplicationController:
 		return &v1.ReplicationController{}
-	case "ReplicaSet":
+	case k8s.ReplicaSet:
 		return &v1beta1.ReplicaSet{}
-	case "Job":
+	case k8s.Job:
 		return &batchv1.Job{}
-	case "DaemonSet":
+	case k8s.DaemonSet:
 		return &v1beta1.DaemonSet{}
-	case "StatefulSet":
+	case k8s.StatefulSet:
 		return &appsv1.StatefulSet{}
-	case "Pod":
+	case k8s.Pod:
 		return &v1.Pod{}
 	}
 
@@ -235,8 +242,8 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		return nil
 	}
 
-	switch conf.meta.Kind {
-	case "Deployment":
+	switch strings.ToLower(conf.meta.Kind) {
+	case k8s.Deployment:
 		v := obj.(*v1beta1.Deployment)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
@@ -250,7 +257,7 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.podLabels[k8s.ProxyDeploymentLabel] = v.Name
 		conf.complete(&v.Spec.Template)
 
-	case "ReplicationController":
+	case k8s.ReplicationController:
 		v := obj.(*v1.ReplicationController)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
@@ -260,7 +267,7 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.podLabels[k8s.ProxyReplicationControllerLabel] = v.Name
 		conf.complete(v.Spec.Template)
 
-	case "ReplicaSet":
+	case k8s.ReplicaSet:
 		v := obj.(*v1beta1.ReplicaSet)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
@@ -270,7 +277,7 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.podLabels[k8s.ProxyReplicaSetLabel] = v.Name
 		conf.complete(&v.Spec.Template)
 
-	case "Job":
+	case k8s.Job:
 		v := obj.(*batchv1.Job)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
@@ -280,7 +287,7 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.podLabels[k8s.ProxyJobLabel] = v.Name
 		conf.complete(&v.Spec.Template)
 
-	case "DaemonSet":
+	case k8s.DaemonSet:
 		v := obj.(*v1beta1.DaemonSet)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
@@ -290,7 +297,7 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.podLabels[k8s.ProxyDaemonSetLabel] = v.Name
 		conf.complete(&v.Spec.Template)
 
-	case "StatefulSet":
+	case k8s.StatefulSet:
 		v := obj.(*appsv1.StatefulSet)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
@@ -300,7 +307,7 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.podLabels[k8s.ProxyStatefulSetLabel] = v.Name
 		conf.complete(&v.Spec.Template)
 
-	case "Pod":
+	case k8s.Pod:
 		v := obj.(*v1.Pod)
 		if err := yaml.Unmarshal(bytes, v); err != nil {
 			return err
```

I went ahead and added that change to #2422.

pkg/inject/inject.go
alpeb: Agreed, I had that misinterpreted. This is a nice leverage of the `resourceTransformerInject` struct :+1:

pkg/inject/inject.go
```diff
@@ -0,0 +1,585 @@
...
+		obj = appsv1.StatefulSet{}
+	case "Pod":
+		obj = v1.Pod{}
+	}
+	return obj
+}
+
+// JSONToYAML is a replacement for the same function in sigs.k8s.io/yaml
+// that does conserve the field order as portrayed in k8s' api structs
+func (conf *ResourceConfig) JSONToYAML(bytes []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
alpeb: @siggy The original problem was that when applying the JSON patch with `jsonpatch.Apply()`, that lib uses a map under the hood to unmarshal the original JSON before applying the patch, which disrupted the order.
We're keeping that, and the following step is to convert the patched JSON into YAML. I can't recall if k8s' `yaml.JSONToYAML()` kept the ordering, but that doesn't matter since we already have a disordered JSON at this point.
So our `conf.JSONToYAML()` uses its knowledge of the workload's type to pass it into k8s' `yaml.Unmarshal()` to build that typed workload object fully filled (actually I'm gonna change this and use the std lib `json.Unmarshal()` instead) which kinda "restores" the order according to the struct, and then I'm finally ready to transform it into YAML.

pkg/inject/inject.go
```diff
@@ -0,0 +1,585 @@
...
+		conf.podLabels[k8s.ProxyDaemonSetLabel] = v.Name
+		conf.complete(&v.Spec.Template)
+
+	case appsv1.StatefulSet:
+		if err := yaml.Unmarshal(bytes, &v); err != nil {
+			return err
+		}
+
+		conf.obj = &v
+		conf.podLabels[k8s.ProxyStatefulSetLabel] = v.Name
```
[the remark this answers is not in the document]
alpeb: Yep, that's the reason :point_up:

cli/cmd/root.go
alpeb: That's right. Note however that this introduces a BC issue against the two two previous Edge releases I think. We could deal with this with some extra logic inside `taggedProxyImage()` and `taggedProxyInitImage()`. I also need to cater for the absence of `Global.Version` in `injectObjectMeta()`.
Either that, or force users to run `linkerd install` if they have any of those Edges. @siggy @klingerf what do you think?

controller/proxy-injector/webhook.go
```diff
@@ -102,221 +92,56 @@ func (w *Webhook) decode(data []byte) (*admissionv1beta1.AdmissionReview, error)
...
 
-	ns := request.Namespace
-	if ns == "" {
-		ns = corev1.NamespaceDefault
-	}
-	log.Infof("resource namespace: %s", ns)
+	jsonUnmarshaler := jsonpb.Unmarshaler{}
 
-	inject, err := w.shouldInject(ns, &deployment)
+	globalConfigJSON, err := ioutil.ReadFile(k8s.MountPathGlobalConfig)
```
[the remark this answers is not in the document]
alpeb: Sorry, that felt through the cracks. This looks fine. I guess as part of that follow-up at some point we mentioned having some sort of API to abstract the protobuf types, and move there things like `taggedProxyImage()` and the building of `inboundSkipPorts`.

pkg/inject/inject.go
```diff
@@ -0,0 +1,585 @@
...
+	*metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+}
+
+// ResourceConfig contains the parsed information for a given workload
+type ResourceConfig struct {
+	globalConfig    *config.Global
+	proxyConfig     *config.Proxy
+	nsAnnotations   map[string]string
+	meta            metav1.TypeMeta
+	obj             interface{}
```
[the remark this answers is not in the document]
alpeb: My intuition told me there had to be a higher abstraction for all the workload types :+1: 
Any particular reason though why not using the type switch? It alleviates us from doing that casting. In this case it would have to be pointers like `case *v1.Pod:`

cli/cmd/testdata/inject-filepath/expected/injected_nginx.yaml
```diff
@@ -1,20 +1,17 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  creationTimestamp: null
```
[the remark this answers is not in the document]
alpeb: :+1: :+1:

cli/cmd/install.go
```diff
@@ -24,60 +25,56 @@ import (
...
+	ProxyResourceLimitMemory    string
+	SingleNamespace             bool
+	EnableHA                    bool
+	ControllerUID               int64
+	ProfileSuffixes             string
+	EnableH2Upgrade             bool
+	NoInitContainer             bool
+	GlobalConfig                string
+	ProxyConfig                 string
 }
```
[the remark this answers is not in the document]
alpeb: :taco: for the kick-ass bash!

cli/cmd/inject_util.go
```diff
@@ -121,29 +64,31 @@ func ProcessYAML(in io.Reader, out io.Writer, report io.Writer, options *injectO
...
-	rt.generateReport(injectReports, report)
+	rt.generateReport(reports, report)
 
 	return nil
 }
 
-func processList(b []byte, options *injectOptions, rt resourceTransformer) ([]byte, []injectReport, error) {
-	var sourceList corev1.List
+// TODO: Temporarily disabled processing Lists of resources because the recursion was too hard
+// to refactor. I'll come back to this later.
```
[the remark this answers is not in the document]
alpeb: Nope, Lists can't be injected now. I'll fix this before we merge.

proto/config/config.proto
```diff
@@ -4,17 +4,19 @@ package linkerd2.config;
...
   string linkerd_namespace = 1;
   bool cni_enabled = 2;
+  string registry = 3;
```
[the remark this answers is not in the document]
alpeb: Looks good to me :star2: 
I've just added a small comment in your PR about this.

pkg/inject/inject.go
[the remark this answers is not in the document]
klingerf: Ahh, yep, I couldn't get the type switch working last night, but now I realize it's because I was using object types instead of pointer types. Will push this change to that branch momentarily.

cli/cmd/uninject_test.go
```diff
@@ -22,11 +22,12 @@ func TestUninjectYAML(t *testing.T) {
 			goldenFileName: "inject_emojivoto_deployment.input.yml",
 			reportFileName: "inject_emojivoto_deployment_uninject.report",
 		},
-		{
+		// TODO: fix Lists
```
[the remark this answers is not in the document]
alpeb: I believe lists are the only important item left in this PR. I'll get it done in a few :hand:

alpeb:
Thanks a lot for the detailed review @klingerf ! I'm almost done replying to your comments. I'll get back to you about the auto-injector injecting even when the annotation is not there...

cli/cmd/root.go
klingerf: It's no problem from my perspective to force users to reinstall the full control plane when upgrading to the next edge release, which will overwrite the old configmap. I definitely prefer that approach to adding compatibility code at this point.

klingerf:
@alpeb Ok great -- have pushed my changes to #2422. When you're happy with that, feel free to merge into this branch before starting on the lists fix. I can also investigate the proxy-injector injecting into un-annotated namespaces issue.

alpeb:
> going to do a another pass now, but fwiw i'm still seeing [#2334 (review)](https://github.com/linkerd/linkerd2/pull/2334#pullrequestreview-209322930)

Yes, I'm looking into that one. Something's fishy for sure. We don't inject labels at the Deployment level (only at its template->spec level) but I discovered in https://github.com/linkerd/linkerd2/issues/2377 that Kubernetes is doing that automatically. I checked that's the case even for webhook-injected deployments in master, but it's not doing it in this branch...

cli/cmd/uninject.go
```diff
@@ -57,7 +52,7 @@ sub-folders, or coming from stdin.`,
 				return err
 			}
 
-			exitCode := runUninjectCmd(in, os.Stderr, os.Stdout, nil)
+			exitCode := runUninjectCmd(in, os.Stderr, os.Stdout, configs{nil, nil})
```
siggy: i think this can just be `configs{}`

cli/cmd/uninject.go
```diff
@@ -66,51 +61,40 @@ sub-folders, or coming from stdin.`,
...
-	}
+func (rt resourceTransformerUninject) transform(bytes []byte) ([]byte, []inject.Report, error) {
+	conf := inject.NewResourceConfig(rt.global, rt.proxy)
 
-	report := injectReport{
-		kind: strings.ToLower(conf.meta.Kind),
-		name: conf.om.Name,
+	report, err := conf.ParseMetaAndYaml(bytes)
+	if err != nil {
+		return bytes, []inject.Report{}, err
```
siggy: `return nil, nil, err`

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+	return conf.objMeta.ObjectMeta != nil, nil
+}
+
+// GetPatch returns the JSON patch containing the proxy and init containers specs, if any
+func (conf *ResourceConfig) GetPatch(bytes []byte) ([]byte, []Report, error) {
+	report := newReport(conf)
+	log.Infof("working on %s %s..", strings.ToLower(conf.meta.Kind), conf.objMeta.Name)
+
+	if err := conf.parse(bytes); err != nil {
+		return nil, []Report{report}, err
```
siggy: `return nil, nil, err`

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+func (conf *ResourceConfig) KindInjectable() bool {
+	for _, kind := range injectableKinds {
+		if strings.ToLower(conf.meta.Kind) == kind {
+			return true
+		}
+	}
+	return false
+}
+
+func (conf *ResourceConfig) getFreshWorkloadObj() runtime.Object {
```
siggy: mind adding a comment that the objects supported here should match `injectableKinds`?

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+
+func (conf *ResourceConfig) taggedProxyInitImage() string {
+	return fmt.Sprintf("%s:%s",
+		conf.proxyConfig.GetProxyInitImage().GetImageName(),
+		conf.globalConfig.GetVersion())
+}
+
+// shouldInject determines whether or not the given deployment should be
+// injected. A deployment shouldn't be injected if it contains any known sidecars
+// or is on a HostNetwork or the pod is not annotated with "linkerd.io/inject: disabled".
```
siggy: should this read `or the pod is not annotated with "linkerd.io/inject: disabled"` (remove the `not`?)
Maybe easier read as:
```
// A deployment shouldn't be injected if:
// - it contains any known sidecars; or
// - is on a HostNetwork; or
// - the pod is annotated with "linkerd.io/inject: disabled".
```

pkg/inject/inject.go
```diff
@@ -0,0 +1,595 @@
...
+//   linkerd.io/inject annotation set to "disabled"; or
+// - the deployment's pod spec has the linkerd.io/inject annotation set to "enabled"
+func (conf *ResourceConfig) shouldInject(r Report) bool {
+	if !r.Injectable() {
+		return false
+	}
+
+	podAnnotation := conf.objMeta.Annotations[k8s.ProxyInjectAnnotation]
+	if conf.nsAnnotations == nil {
+		return true
```
siggy: should this check be removed? if `conf.nsAnnotations == nil` we shouldn't return true unless `podAnnotation == k8s.ProxyInjectEnabled`?

pkg/inject/inject.go
```diff
@@ -0,0 +1,594 @@
...
+	if err := yaml.Unmarshal(bytes, &conf.objMeta); err != nil {
+		return false, err
+	}
+	return conf.objMeta.ObjectMeta != nil, nil
+}
+
+// GetPatch returns the JSON patch containing the proxy and init containers specs, if any
+func (conf *ResourceConfig) GetPatch(bytes []byte) ([]byte, []Report, error) {
+	report := newReport(conf)
+	log.Infof("working on %s %s..", strings.ToLower(conf.meta.Kind), conf.objMeta.Name)
```
klingerf: I noticed that this log line is printed as:

```
time="2019-03-01T21:39:06Z" level=info msg="working on deployment .."
```

Looks like the proxy injector is calling `WithMeta` with these fields:

```
WithMeta(request.Kind.Kind, request.Namespace, request.Name)
```

But I think the issue is that `request.Namespace` and `request.Name` could be blank, based on this documentation:

```
	// Name is the name of the object as presented in the request.  On a CREATE operation, the client may omit name and
	// rely on the server to generate the name.  If that is the case, this method will return the empty string.
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,5,opt,name=name"`
	// Namespace is the namespace associated with the request (if any).
	// +optional
	Namespace string `json:"namespace,omitempty" protobuf:"bytes,6,opt,name=namespace"`
```

The previous implementation parsed `request.Object.Raw` as a Deployment object and used the deployment's name instead of `request.Name`. But if you took that approach, then you'd end up parsing the object meta twice -- once in webhook.go, and once in the `parse` func in this file.

Off the top of my head I can't think of a great way to fix this. If you moved the log line after the `parse` call, that would work, but it also suggests that we're not attaching the right metadata when calling `WithMeta`.

pkg/inject/inject.go
klingerf: Yeah, I think this must be the source of https://github.com/linkerd/linkerd2/pull/2334#discussion_r261472515, but I can't figure out what's up.

Only the auto-inject webhook calls `WithNsAnnotations`, so for the CLI inject code path, I'd expect `conf.nsAnnotations` to be `nil`. For CLI inject, we only care about pod annotations, so I'd expect this return statement to be:

```
		return podAnnotation != k8s.ProxyInjectDisabled
```

But it looks like it's also OK to `return true` as written, since the `Injectable` call above this is already checking that the `linkerd.io/inject: disabled` annotation is not set. This is pretty subtle though, so if there was some way to make this easier to reason about, that would be great.

For the auto-inject use case, we're expecting `conf.nsAnnotations` to not be `nil`, since we've called `WithNsAnnotations`, in which case we'll proceed to the namespace annotation check later in this method. But my guess is that's it's still possible for `conf.nsAnnotations` to be `nil` even after calling `WithNsAnnotations`, if the namespace doesn't actually have any annotations. If that were the case (untested), then that would explain why the proxy injector is injecting into namespaces that don't contain the `linkerd.io/inject: enabled` annotation.

Anyway, we might need to refactor this slightly to make it easier to understand how the inject decision is made.

controller/proxy-injector/webhook_test.go
```diff
@@ -2,30 +2,60 @@ package injector
...
+	return inject.
+		NewResourceConfig(globalConfig, proxyConfig).
+		WithNsAnnotations(map[string]string{k8s.ProxyInjectAnnotation: k8s.ProxyInjectEnabled})
+}
+
+func confNsDisabled() *inject.ResourceConfig {
+	return inject.NewResourceConfig(globalConfig, proxyConfig).WithNsAnnotations(map[string]string{})
+}
+
 func TestShouldInject(t *testing.T) {
```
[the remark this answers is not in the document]
alpeb: @klingerf I haven't been able to reproduce this. To clarify the scenario, you mean pods and their namespace, that both don't have any linkerd annotation set?

alpeb:
@siggy @klingerf I've addressed the labels issue (alongside with `Created-by` not being properly set) in #2432

pkg/inject/inject.go
alpeb: I've pushed #2433 as a suggestion to try making this more tractable. Although the functionality is the same, as I couldn't repro [#2334 (comment)](https://github.com/linkerd/linkerd2/pull/2334#discussion_r261472515).

pkg/inject/inject.go
alpeb: @klingerf We also parse the meta twice in CLI inject (even before this PR) for the very purpose of extracting the Kind and Name. Given the `AdmissionRequest`'s Name is useless, I'll pre-parse `request.Object.Raw` as well to extract it.
I was gonna make a PR for this but I pushed it by mistake to this branch :man_shrugging:

pkg/inject/inject.go
siggy: I also tried to repro this without success:
```bash
bin/linkerd install --proxy-auto-inject | kubectl apply -f -
curl https://run.linkerd.io/emojivoto.yml | kubectl apply -f -
```

Noting that emojivoto.yml does not have any `annotations` sections, it appears `conf.nsAnnotations` (coming from `namespace.GetAnnotations()`) is an empty `map[string]string` and not `nil`.

All that said, I'm onboard with making this code a bit more explicit, as I wouldn't want to rely on this subtlety when deciding whether to auto inject something. Also onboard with hashing this out in #2433 (thanks for posting that @alpeb).

controller/proxy-injector/webhook.go
```diff
@@ -107,15 +112,18 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
 
 	conf := inject.NewResourceConfig(globalConfig, proxyConfig).
 		WithNsAnnotations(nsAnnotations).
-		WithMeta(request.Kind.Kind, request.Namespace, request.Name)
-	patchJSON, reports, err := conf.GetPatch(request.Object.Raw)
+		WithKind(request.Kind.Kind)
+	supportedResource, err := conf.ParseMeta(request.Object.Raw)
```
siggy: naming nit: if the `bool` returned by `ParseMeta` strictly indicates whether the payload was empty, recommend just calling this `empty` or similar.

controller/proxy-injector/webhook.go
```diff
@@ -89,6 +89,11 @@ func (w *Webhook) decode(data []byte) (*admissionv1beta1.AdmissionReview, error)
 func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admissionv1beta1.AdmissionResponse, error) {
 	log.Debugf("request object bytes: %s", request.Object.Raw)
 
+	admissionResponse := &admissionv1beta1.AdmissionResponse{
```
siggy: mind moving this down to directly above the first time it's returned? i'm extra paranoid about myself or other devs accidentally returning this object with `Allowed: true` when we shouldn't.

pkg/inject/inject.go
klingerf: Hmm, ok, maybe it's something about my environment, but I can't repro the same behavior with `edge-19.2.5` -- it only happens on this branch.

In my environment, I have 2 namespaces configured, `kltest1` and `kltest2`, as follows:

```
$ kubectl get ns/kltest1 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    linkerd.io/inject: enabled
  creationTimestamp: 2019-02-12T17:18:23Z
  name: kltest1
  resourceVersion: "110658"
  selfLink: /api/v1/namespaces/kltest1
  uid: 3382bbc5-2eea-11e9-a483-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active

$ kubectl get ns/kltest2 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: 2019-02-12T17:18:24Z
  name: kltest2
  resourceVersion: "110623"
  selfLink: /api/v1/namespaces/kltest2
  uid: 3447222e-2eea-11e9-a483-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

Only `kltest1` has the `linkerd.io/inject: enabled` annotation.

I then install `edge-19.2.5` with:

```
linkerd install --proxy-auto-inject --tls optional | kubectl apply -f -
```

And then apply the booksapp config in both namespaces:

```
curl https://run.linkerd.io/booksapp.yml | kubectl apply -f - -n kltest1
curl https://run.linkerd.io/booksapp.yml | kubectl apply -f - -n kltest2
```

That results in only the pods in the `kltest1` namespace being injected:

```
kubectl get po --all-namespaces | grep kltest
kltest1       authors-55d8c75c67-2sbj7                     2/2       Running   0          23m
kltest1       books-6bc5d6c547-nlhqm                       2/2       Running   0          23m
kltest1       traffic-5fccc89fbd-hgkns                     2/2       Running   0          23m
kltest1       webapp-766f6d4977-5hwrb                      2/2       Running   0          23m
kltest1       webapp-766f6d4977-nfnkb                      2/2       Running   0          23m
kltest1       webapp-766f6d4977-rvkfw                      2/2       Running   0          23m
kltest2       authors-6b757cd4fd-f4dn9                     1/1       Running   0          23m
kltest2       books-7c79648958-wm8s7                       1/1       Running   0          23m
kltest2       traffic-66b46679f9-xpr5q                     1/1       Running   0          23m
kltest2       webapp-58d887ff9d-5wxbg                      1/1       Running   0          23m
kltest2       webapp-58d887ff9d-kjwj4                      1/1       Running   0          23m
kltest2       webapp-58d887ff9d-zknjj                      1/1       Running   0          23m
```

---

Then I tear everything down:

```
curl https://run.linkerd.io/booksapp.yml | kubectl delete -f - -n kltest2
curl https://run.linkerd.io/booksapp.yml | kubectl delete -f - -n kltest1
linkerd install --proxy-auto-inject --tls optional | kubectl delete -f -
kubectl delete mutatingwebhookconfigurations/linkerd-proxy-injector-webhook-config
kubectl get cm,secret -n kltest1 -oname | grep linkerd | xargs kubectl delete -n kltest1
kubectl get cm,secret -n kltest2 -oname | grep linkerd | xargs kubectl delete -n kltest2

```

And reinstall from this branch:

```
bin/docker-build
bin/linkerd install --proxy-auto-inject --tls optional | kubectl apply -f -
```

And then apply the booksapp config in both namespaces:

```
curl https://run.linkerd.io/booksapp.yml | kubectl apply -f - -n kltest1
curl https://run.linkerd.io/booksapp.yml | kubectl apply -f - -n kltest2
```

And that unfortunately results in pods in both namespaces being injected:

```
kubectl get po --all-namespaces | grep kltest
kltest1       authors-5c7d5c6c7d-xk8c2                     2/2       Running   0          1m
kltest1       books-79fdf678b-qddp7                        2/2       Running   0          1m
kltest1       traffic-59949bdb85-h9k4m                     2/2       Running   0          1m
kltest1       webapp-9f7b88c7b-lm2pd                       2/2       Running   0          1m
kltest1       webapp-9f7b88c7b-nmkxd                       2/2       Running   0          1m
kltest1       webapp-9f7b88c7b-q7zlc                       2/2       Running   0          1m
kltest2       authors-5c7d5c6c7d-q8j4j                     2/2       Running   0          1m
kltest2       books-79fdf678b-qlc2h                        2/2       Running   0          1m
kltest2       traffic-59949bdb85-6mxmb                     2/2       Running   0          1m
kltest2       webapp-9f7b88c7b-j2k86                       2/2       Running   0          1m
kltest2       webapp-9f7b88c7b-prfrz                       2/2       Running   0          1m
kltest2       webapp-9f7b88c7b-r56sr                       2/2       Running   0          1m
```

---

So it seems like something related to this branch, but as I said, having stared at this code for a while, I can't figure out where the issue is coming from. I'd be fine with merging this branch and following up with a fix separately, so as not to delay merging longer than needed.

siggy: I'm unable to reproduce on this branch. Confirming versions:

```bash
$ bin/linkerd version
Client version: git-54b6c038
Server version: git-54b6c038

$ kubectl version --short
Client Version: v1.10.7
Server Version: v1.10.11
```

```bash
$ kubectl get ns kltest1 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
      {"apiVersion":"v1","kind":"Namespace","metadata":{"annotations":{"linkerd.io/inject":"enabled"},"name":"kltest1","namespace":""}}
    linkerd.io/inject: enabled
  creationTimestamp: 2019-03-04T19:39:11Z
  name: kltest1
  resourceVersion: "334201"
  selfLink: /api/v1/namespaces/kltest1
  uid: 2f1b39ea-3eb5-11e9-a352-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active

$ kubectl get ns kltest2 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
      {"apiVersion":"v1","kind":"Namespace","metadata":{"annotations":{},"name":"kltest2","namespace":""}}
  creationTimestamp: 2019-03-04T19:39:09Z
  name: kltest2
  resourceVersion: "334170"
  selfLink: /api/v1/namespaces/kltest2
  uid: 2de9c7bd-3eb5-11e9-a352-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

```bash
bin/linkerd install --proxy-auto-inject --tls optional | kubectl apply -f -
bin/linkerd check
curl https://run.linkerd.io/booksapp.yml | kubectl apply -f - -n kltest1
curl https://run.linkerd.io/booksapp.yml | kubectl apply -f - -n kltest2
```

```bash
$ kubectl get po --all-namespaces | grep kltest
kltest1       authors-5c7d5c6c7d-7cs86                     2/2       Running   0          2m
kltest1       books-79fdf678b-p2zpw                        2/2       Running   0          2m
kltest1       traffic-59949bdb85-vvv8x                     2/2       Running   0          2m
kltest1       webapp-9f7b88c7b-hp8c9                       2/2       Running   0          2m
kltest1       webapp-9f7b88c7b-stb5q                       2/2       Running   0          2m
kltest1       webapp-9f7b88c7b-xjqjc                       2/2       Running   0          2m
kltest2       authors-6b757cd4fd-5nd5q                     1/1       Running   0          2m
kltest2       books-7c79648958-pv2hp                       1/1       Running   0          2m
kltest2       traffic-66b46679f9-9ztnj                     1/1       Running   0          2m
kltest2       webapp-58d887ff9d-2g5z7                      1/1       Running   0          2m
kltest2       webapp-58d887ff9d-cxxnp                      1/1       Running   0          2m
kltest2       webapp-58d887ff9d-pmxtd                      1/1       Running   0          2m
```

klingerf: @siggy Can you try testing against a namespace that has 0 annotations? It looks like your `kltest2` namespace still includes the `kubectl.kubernetes.io/last-applied-configuration` annotation. Likewise, I'll try testing against a namespace with at least 1 non-`linkerd.io/inject` annotation.

siggy: Confirmed reproduction. I was using `kubectl apply` (which adds `kubectl.kubernetes.io/last-applied-configuration`) rather than `kubectl create` when creating the namespaces. When the namespace objects do not have any annotations on them, both apps get injected.

### kltest1-ns.yml

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: kltest1
  annotations:
    linkerd.io/inject: enabled
```

### kltest2-ns.yml

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: kltest2
```

## Use `kubectl create` to create namespaces

```bash
$ kubectl create -f kltest1-ns.yml
$ kubectl create -f kltest2-ns.yml

$ kubectl get ns kltest1 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    linkerd.io/inject: enabled
  creationTimestamp: 2019-03-04T19:51:50Z
  name: kltest1
  resourceVersion: "335737"
  selfLink: /api/v1/namespaces/kltest1
  uid: f3da6503-3eb6-11e9-a352-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active

$ kubectl get ns kltest2 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: 2019-03-04T19:51:53Z
  name: kltest2
  resourceVersion: "335744"
  selfLink: /api/v1/namespaces/kltest2
  uid: f52f9c3c-3eb6-11e9-a352-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

## Observe both namespaces injected.

```bash
$ kubectl get po --all-namespaces | grep kltest
kltest1       authors-5c7d5c6c7d-fhqwm                     0/2       Init:0/1          0          21s
kltest1       books-79fdf678b-mhxxr                        0/2       Init:0/1          0          18s
kltest1       traffic-59949bdb85-2cnw6                     0/2       Pending           0          16s
kltest1       webapp-9f7b88c7b-fqq7x                       0/2       PodInitializing   0          21s
kltest1       webapp-9f7b88c7b-k5vsm                       0/2       Init:0/1          0          21s
kltest1       webapp-9f7b88c7b-n6kfh                       0/2       Init:0/1          0          21s
kltest2       authors-5c7d5c6c7d-l44hw                     0/2       Pending           0          8s
kltest2       books-79fdf678b-rmr7d                        0/2       Pending           0          7s
kltest2       traffic-59949bdb85-9jds6                     0/2       Pending           0          6s
kltest2       webapp-9f7b88c7b-nqxkw                       0/2       Pending           0          8s
kltest2       webapp-9f7b88c7b-qcjbg                       0/2       Pending           0          8s
kltest2       webapp-9f7b88c7b-w7vwt                       0/2       Pending           0          8s
```

klingerf: Sure enough, if I apply a config in the `kltest3` namespace, which includes a miscellaneous annotation:

```
kubectl get ns/kltest3 -oyaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    kltest: "3"
  creationTimestamp: 2019-03-04T19:52:40Z
  name: kltest3
  resourceVersion: "574251"
  selfLink: /api/v1/namespaces/kltest3
  uid: 11bb4b1e-3eb7-11e9-8d00-025000000001
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

The pods in that namespace are not inject:

```
kltest3       authors-6b757cd4fd-hx7ls                     1/1       Running   0          1m
kltest3       books-7c79648958-nwqxt                       1/1       Running   0          1m
kltest3       traffic-66b46679f9-9p6gw                     1/1       Running   0          1m
kltest3       webapp-58d887ff9d-qc2gf                      1/1       Running   0          1m
kltest3       webapp-58d887ff9d-qjtlz                      1/1       Running   0          1m
kltest3       webapp-58d887ff9d-vkvxw                      1/1       Running   0          1m
```

So it seems to only be an issue when the target namespace has 0 annotations.

klingerf: I can check to see if #2433 fixes the issue, once it has been rebased.

siggy: Confirmed #2433 fixes the issue (also found a simpler repro via `kubectl create`:

# #2334

```bash
$ bin/linkerd install --proxy-auto-inject --tls optional | kubectl apply -f -
$ curl https://run.linkerd.io/emojivoto.yml | kubectl create -f -
$ kubectl -n emojivoto get po
NAME                        READY     STATUS    RESTARTS   AGE
emoji-7d4b89647b-9pxbg      2/2       Running   0          34s
vote-bot-846d8846fb-s9vxw   2/2       Running   0          33s
voting-5f54984cdf-nrhvs     2/2       Running   0          34s
web-6457545bcd-8vmw9        2/2       Running   0          34s
```

# #2433

```bash
$ bin/linkerd install --proxy-auto-inject --tls optional | kubectl apply -f -
$ curl https://run.linkerd.io/emojivoto.yml | kubectl create -f -
$ kubectl -n emojivoto get po
NAME                        READY     STATUS    RESTARTS   AGE
emoji-f449cd7d8-dlt8c       1/1       Running   0          7s
vote-bot-644ddf8596-62582   1/1       Running   0          6s
voting-6cfd45c845-fm9gc     1/1       Running   0          7s
web-6cddf54f66-dxbtz        1/1       Running   0          7s
```

cli/cmd/inject.go
```diff
@@ -106,321 +92,92 @@ sub-folders, or coming from stdin.`,
...
+	patchJSON, err := p.Marshal()
+	if patchJSON == nil {
+		return bytes, reports, nil
 	}
-
-	initArgs := []string{
-		"--incoming-proxy-port", fmt.Sprintf("%d", options.inboundPort),
-		"--outgoing-proxy-port", fmt.Sprintf("%d", options.outboundPort),
-		"--proxy-uid", fmt.Sprintf("%d", options.proxyUID),
+	if err != nil {
```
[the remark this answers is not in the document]
alpeb: You're right, I'll fix that.

PR #2344, opened 2019-02-21
Install the Linkerd global and proxy config maps

### LINKED ISSUE #2298: Have `Linkerd install` generate a new ConfigMap for general and proxy configs
Have `linkerd install` output an additional `ConfigMap` populated with the json serialization of `pb.GlobalConfig` and `pb.ProxyConfig` that were declared in #2291.

ihcsim:
This change creates a config map, named `linkerd-config`, to store the global and proxy configurations, during `linkerd install`.

Test cases as follows:
```
# basic: expect default configs in config map
$ bin/linkerd install  | kubectl apply -f -

# custom Linkerd namespace: expect global namespace to be linkerd-custom.
$ bin/linkerd install -l linkerd-custom | kubectl apply -f -

# single namespace: expect default global and proxy configs.
$ bin/linkerd install --single-namespace | kubectl apply -f -

# tls: expect (empty) identity context to be defined in global config. 
$ bin/linkerd install --tls=optional | kubectl apply -f -

# cni: expect cni-enabled to be true in global config.
$ bin/linkerd install --linkerd-cni-enabled | kubectl apply -f -

# custom configs: expect all custom proxy configs to be persisted.
$ bin/linkerd install --api-port=7000 --control-port=7001 --inbound-port=7002 --metrics-port=7003 --outbound-port=7004 --proxy-cpu-limit=0.5 --proxy-cpu-request=0.1 --proxy-log-level="info,linkerd2_proxy=debug" --proxy-memory-limit=512M --proxy-memory-request=64M --proxy-uid=9999 --skip-inbound-ports=4222,6222 --skip-outbound-ports=6777,6778 | kubectl apply -f -
```

Fixes #2298 


ihcsim:
> we'll have to change it to `DestinationApiPort`

Got it. I will update the protobuf file (unless it has already been updated in another PR).

cli/cmd/install.go
```diff
@@ -331,3 +348,69 @@ func readIntoBytes(filename string) ([]byte, error) {
...
+	for _, port := range options.ignoreOutboundPorts {
+		ignoreOutboundPorts = append(ignoreOutboundPorts, &config.Port{Port: uint32(port)})
+	}
+
+	return &config.ProxyConfig{
+		ProxyImage: &config.Image{
+			ImageName:  options.taggedProxyImage(),
+			PullPolicy: options.imagePullPolicy,
+			Registry:   options.dockerRegistry,
+		},
```
klingerf: This produces a config that looks like:

```
  "proxyImage": {
    "imageName": "gcr.io/linkerd-io/proxy:dev-47a1f2fb-kl",
    "pullPolicy": "IfNotPresent",
    "registry": "gcr.io/linkerd-io"
  },
```

My original thinking was that we would represent the image as:

```
  "proxyImage": {
    "imageName": "proxy",
    "pullPolicy": "IfNotPresent",
    "registry": "gcr.io/linkerd-io"
  },
```

No registry or version in the image name. The registry already exists as a separate field, and the version should be determined by the control plane components that load this config, rather than the CLI that's used to generate it.

cli/cmd/install.go
```diff
@@ -182,6 +186,17 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
 		profileSuffixes = "svc.cluster.local."
 	}
 
+	jsonMarshaler := jsonpb.Marshaler{}
```
klingerf: In other places where we use the jsonpb marshaler, we use:

```
jsonpb.Marshaler{EmitDefaults: true, Indent: "  "}
```

I think we should use `EmitDefaults` here too, but I'm on the fence about `Intent`.

cli/cmd/install.go
```diff
@@ -331,3 +348,69 @@ func readIntoBytes(filename string) ([]byte, error) {
...
+	return &config.ProxyConfig{
+		ProxyImage: &config.Image{
+			ImageName:  k8s.ProxyImageName,
+			PullPolicy: options.imagePullPolicy,
+			Registry:   options.dockerRegistry,
+		},
+		ProxyInitImage: &config.Image{
+			ImageName:  k8s.ProxyInitImageName,
+			PullPolicy: options.imagePullPolicy,
+			Registry:   options.dockerRegistry,
```
klingerf: For parity with the existing implementation, we unfortunately can't just use the value of the `--docker-registry` flag here. If the value of `--proxy-image` includes a non-default registry, we use that instead. For instance, if I ran:

```
linkerd install --registry foo.com --proxy-image bar.com/custom-proxy
```

That should produce:

```
&config.Image{
	ImageName: "custom-proxy",
	PullPolicy: "IfNotPresent",
	Registry: "bar.com",
}
```

But as written it uses "foo.com" for the registry.

klingerf: (sorry, hit submit too soon)

Same comment goes for the proxy-init image. We should probably refactor `taggedProxyImage` and `taggedProxyInitImage` to make the registry more accessible.

ihcsim: Per discussion with @klingerf, we will remove the registry field to avoid having to workaround edge cases like fully-qualified image name, versus user-provided Linkerd version etc.

Issue #2164, opened 2019-01-28
RFC: Break up linkerd install into stages


siggy:
This RFC proposes modification to the `linkerd install` flow, dividing Linkerd
installation into stages grouped by permission level.

# Current State

Installation of the full suite of components requires three commands, with
varying combinations of flags:

## `linkerd install-cni`

- ClusterRole
  - linkerd-cni
- ClusterRoleBinding
  - linkerd-cni
- Namespace
  - linkerd
- ServiceAccount
  - linkerd-cni
- DaemonSet
  - linkerd-cni
- ConfigMap
  - linkerd-cni-config

## `linkerd install`

- ClusterRole
  - linkerd-linkerd-controller
  - linkerd-linkerd-prometheus
- ClusterRoleBinding
  - linkerd-linkerd-controller
  - linkerd-linkerd-prometheus
- Namespace
  - linkerd
- CustomResourceDefinition
  - serviceprofiles.linkerd.io
- ServiceAccount
  - linkerd-controller
  - linkerd-grafana
  - linkerd-prometheus
  - linkerd-web
- ConfigMap
  - linkerd-grafana-config
  - linkerd-prometheus-config
- Service
  - linkerd-controller-api
  - linkerd-grafana
  - linkerd-prometheus
  - linkerd-proxy-api
  - linkerd-proxy-injector
  - linkerd-web
- Deployment
  - linkerd-controller
  - linkerd-grafana
  - linkerd-prometheus
  - linkerd-web

## `linkerd install --tls optional`

- ClusterRole
  - linkerd-linkerd-ca
- ClusterRoleBinding
  - linkerd-linkerd-ca
- ServiceAccount
  - linkerd-ca
- ConfigMap
  - linkerd-ca-bundle
- Deployment
  - linkerd-ca

## `linkerd install --tls optional --proxy-auto-inject`

- ClusterRole
  - linkerd-linkerd-proxy-injector
- ClusterRoleBinding
  - linkerd-linkerd-proxy-injector
- ConfigMap
  - linkerd-proxy-injector-sidecar-config
- Service
  - linkerd-proxy-injector
- Deployment
  - linkerd-proxy-injector

### `linkerd install-sp`

- ServiceProfile
  - linkerd-controller-api
  - linkerd-proxy-api
  - linkerd-prometheus
  - linkerd-grafana

# Proposal: Multi-stage installation

All components listed above, grouped by privilege, `--admin` and `--user`:

```bash
$ linkerd install --help
[...]

PR #2349, opened 2019-02-22
Authorization-aware control-plane components

siggy:
The control-plane components relied on a `--single-namespace` param,
passed from `linkerd install` into each individual component, to
determine which namespaces they were authorized to access, and whether
to support ServiceProfiles. This command-line flag was redundant given
the authorization rules encoded in the parent `linkerd install` output,
via [Cluster]Role[Binding]s.

Modify the control-plane components to query Kubernetes at startup to
determine which namespaces they are authorized to access, and whether
ServiceProfile support is available. This allows removal of the
`--single-namespace` flag on the components.

Also update `bin/test-cleanup` to cleanup the ServiceProfile CRD.

TODO:
- Remove `--single-namespace` flag on `linkerd install`, part of #2164

### DIFF
--- a/bin/test-cleanup
+++ b/bin/test-cleanup
@@ -28,3 +28,9 @@ if ! clusterroles=$(kubectl get clusterroles -oname | grep -E "/linkerd-$linkerd
 else
   kubectl delete $clusterroles
 fi
+
+if ! crd=$(kubectl get crd/serviceprofiles.linkerd.io -oname); then
+  echo "no serviceprofile crd found for [$linkerd_namespace]" >&2
+else
+  kubectl delete $crd
+fi
--- a/controller/k8s/api.go
+++ b/controller/k8s/api.go
@@ -71,13 +71,64 @@ type API struct {
 	namespace         string
 }
 
-// NewAPI takes a Kubernetes client and returns an initialized API
+// InitializeAPI creates Kubernetes clients and returns an initialized API wrapper.
+func InitializeAPI(kubeConfig string, namespace string, resources ...APIResource) (*API, error) {
+	k8sClient, err := NewClientSet(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// check for cluster-wide vs. namespace-wide access
+	clusterAccess, err := k8s.ClusterAccess(k8sClient, namespace)
+	if err != nil {
+		return nil, err
+	}
+	restrictToNamespace := ""
+	if !clusterAccess {
+		log.Warnf("Not authorized for cluster-wide access, limiting access to \"%s\" namespace", namespace)
+		restrictToNamespace = namespace
+	}
+
+	// check for need and access to ServiceProfiles
+	var spClient *spclient.Clientset
+	idxSP := 0
+	needSP := false
+	for i := range resources {
+		if resources[i] == SP {
+			needSP = true
+			idxSP = i
+			break
+		}
+	}
+	if needSP {
+		serviceProfiles, err := k8s.ServiceProfilesAccess(k8sClient)
+		if err != nil {
+			return nil, err
+		}
+		if serviceProfiles {
+			spClient, err = NewSpClientSet(kubeConfig)
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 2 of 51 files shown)


grampelberg:
This is awesome! I recently found `can-i --list` and had my mind blown.

controller/k8s/api.go
```diff
@@ -71,26 +71,79 @@ type API struct {
...
+	}
+
+	// check for cluster-wide vs. namespace-wide access
+	clusterAccess, err := k8s.ClusterAccess(k8sClient, namespace)
+	if err != nil {
+		return nil, err
+	}
+	restrictToNamespace := ""
+	if !clusterAccess {
+		log.Warnf("Not authorized for cluster-wide access, limiting access to \"%s\" namespace", namespace)
```
adleong: Not something to address right now, but possibly something to think about going forward as we refine the install process: 

We could potentially do authz checks individually for each namespace in the cluster and watch all namespaces that we have access to.  This would be helpful if you want a Linkerd control plane to manage a data plane that spans more than one namespace, but you don't have full cluster access.

controller/k8s/api.go
```diff
@@ -71,26 +71,79 @@ type API struct {
...
+	idxSP := 0
+	needSP := false
+	for i := range resources {
+		if resources[i] == SP {
+			needSP = true
+			idxSP = i
+			break
+		}
+	}
+	if needSP {
```
adleong: This works, but I would also be okay with removing this logic and just saying that the service profile CRD and permissions are *required*.  That puts us in an intermediate state where users without cluster access can't install Linkerd without help from someone who does have cluster access to install the CRD, but that should be resolved once we have the install flow broken in two stages.

TIOLI

siggy: Per the [product org's requirement 3](https://github.com/linkerd/linkerd2/issues/2164#issuecomment-458265137):
```
3. Linkerd should still be usable in a locked down environment without elevated permissions, even if that mode has fewer features.
```

siggy:
@adleong Good catch on `ignore-namespaces`, I did not noticed that it was set to `kube-system` by default (I was looking for `--ignore-namespaces` in the `linkerd install` output). I'll put it back.

grampelberg:
@siggy what's the implication of us ignoring `kube-system`?

siggy:
@grampelberg I assume we ignore `kube-system` by default since we're not expecting it to be in the mesh. I think this particular setting only applies to the Public APIs `ListPods` endpoint.

To be clear, this PR is not changing anything related to `ignore-namespaces`, I had accidentally removed it thinking it was dead code, it's not.

grampelberg:
Makes total sense, I was just trying to remember why it'd been done in the past and whether it was still needed or not (so totally irrelevant to this PR).

bin/test-cleanup
```diff
@@ -28,3 +28,9 @@ if ! clusterroles=$(kubectl get clusterroles -oname | grep -E "/linkerd-$linkerd
 else
   kubectl delete $clusterroles
 fi
+
+if ! crd=$(kubectl get crd/serviceprofiles.linkerd.io -oname); then
```
klingerf: The service profile CRD is global to the whole cluster, in that all linkerd installs in the cluster rely on the same CRD being present. If I run this cleanup script in a cluster where I have linkerd installed in a different, non-test namespace, that linkerd install will break, which seems problematic. Would you mind reverting this change?

siggy: @klingerf Sure will revert. Should this motivate that we someone scope and/or name the ServiceProfile CRD to a specific Linkerd control-plane, similar to ClusterRoles?

klingerf: Ideally yes, but in practice we can't change the name of our CRDs without also changing the name/group of the ServiceProfile resource itself, since Kubernetes requires that CRDs be named as follows:

> must be spec.names.plural+"."+spec.group

So a name like `linkerd-l5d-integration-serviceprofiles.linkerd.io` is invalid.

siggy: 😞

PR #2360, opened 2019-02-23
Increase integration test timeouts

siggy:
The integration tests occasionally timeout in ci when talking to
Kubernetes and Linkerd:
https://travis-ci.org/linkerd/linkerd2/jobs/497300669#L972
https://travis-ci.org/linkerd/linkerd2/jobs/497329339#L7284

Increase `linkerd check --wait` from `0` to `30s`.
Increase `HTTPGetURL` timeout from 30s to 1 minute.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,12 @@ branches:
     - master
     - /^(edge|stable)-\d+\.\d+\.[\w-]+$/ # build version tags
 
-# TESTING integration tests prior to master merge
 stages:
   - name: test
   - name: docker-deploy
-    # if: type != pull_request
+    if: type != pull_request
   - name: integration-test
-    # if: type != pull_request
+    if: type != pull_request
 
 jobs:
   include:
@@ -110,8 +109,7 @@ jobs:
       after_success:
         - bin/docker-push-deps
         - bin/docker-push $LINKERD_TAG
-        # TESTING integration tests prior to master merge
-        # - bin/docker-retag-all $LINKERD_TAG master && bin/docker-push master
+        - bin/docker-retag-all $LINKERD_TAG master && bin/docker-push master
         - target/cli/linux/linkerd install --linkerd-version=$LINKERD_TAG |tee linkerd.yml
         - kubectl -n linkerd apply -f linkerd.yml --prune --selector='linkerd.io/control-plane-component'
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,12 +9,13 @@ branches:
...
 
+# TESTING integration tests prior to master merge
 stages:
   - name: test
   - name: docker-deploy
-    if: type != pull_request
+    # if: type != pull_request
   - name: integration-test
-    if: type != pull_request
+    # if: type != pull_request
```
siggy: note: these changes are to make integration tests run on this PR prior to master merge. will back them out once integration tests are validated.

test/install_test.go
```diff
@@ -131,7 +131,7 @@ func TestVersionPostInstall(t *testing.T) {
 }
 
 func TestCheckPostInstall(t *testing.T) {
-	cmd := []string{"check", "--expected-version", TestHelper.GetVersion(), "--wait=0"}
+	cmd := []string{"check", "--expected-version", TestHelper.GetVersion(), "--wait=1m"}
```
klingerf: We were previously using `--wait=0` here, since the check output would be different if any of the checks had to retry, and the subsequent golden file check would fail. It looks like that's still the case. If I run `linkerd check` while the control plane is installing and dump the output to a text file, that file contains a bunch of control characters that are printed when checks are retried:

```
linkerd-existence
-----------------
| controller pod is running -- waiting for check to complete [K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K/ controller pod is running -- waiting for check to complete [K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K[K- controller pod is running -- waiting for check to complete 
```

Am pretty sure those control characters will interfere with the `ValidateOutput` call on line 146. Would you mind reverting the changes to the `--wait` flag in files file, and instead wrapping the those tests with `TestHelper.RetryFor` to achieve the retry behavior that you want?

siggy: Good catch @klingerf. I optimistically assumed the output would ignore the spinners. Will revert.

PR #2362, opened 2019-02-23
Enable more linters

siggy:
Enables the following linters:
- `deadcode` finds unused code.
- `depguard` checks if package imports are in a list of acceptable
  packages.
- `misspell` finds commonly misspelled English words in comments.
- `nakedret` finds naked returns in functions greater than a specified
  function length.
- `varcheck` finds unused global variables and constants.

Our repo was already passing these linters so no fixes needed.

Also explicitly list all linters enabled by default, for reference.

Part of #217

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,7 +20,7 @@ linters:
   # - dupl
   # - gochecknoglobals
   # - gochecknoinits
-  # - goconst # TODO
+  # - goconst
   # - gocyclo
   # - goimports
   # - gosec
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -6,9 +6,26 @@ linters:
...
   - gofmt
   - golint
+  - depguard
+  - misspell
+  - nakedret
+  # TODO: enable more linters!
+  # - dupl
+  # - gochecknoglobals
+  # - gochecknoinits
+  # - goconst # TODO
```
olix0r: double TODO ;)

PR #2368, opened 2019-02-24
lint: Enable unconvert

siggy:
unconvert removes unnecessary type conversions:
https://github.com/mdempsky/unconvert

Part of #217

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,6 +14,7 @@ linters:
   - misspell
   - nakedret
   - typecheck
+  - unconvert
   - unused
   - varcheck
   # TODO: enable more linters!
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -14,6 +14,7 @@ linters:
   - misspell
   - nakedret
   - typecheck
+  - unconvert
```
rmars: should we be adding comments to each of these saying what they do? for eslint you can just go to the eslint site and look up good documentation for a rule by name, but these are harder to find. for example, if I google "go lint unconvert" a direct link to https://github.com/mdempsky/unconvert doesn't even show up in the first page of results.

siggy: good idea! i've updated `.golangci.yml` with some helpful pointers.

PR #2371, opened 2019-02-24
destination: Only return TLS identities when appropriate

### LINKED ISSUE #2217: proxy-api: Only return TLS identities when dst pod is known to participate in identity

olix0r:
As described in #2217, the controller returns TLS identities for results even when the destination pod may not be able to participate in identity requester: specifically, the other pod may not have the same controller namespace or it may not be injected with identity.

This change introduces a new annotation, `linkerd.io/identity-mode` that is set when injecting pods (via both CLI and webhook). This annotation is _always_ added.

The destination service now only returns TLS identities when this annotation is set to `optional` on a pod and the destination pod uses the same controller. These semantics are expected to change before the 2.3 release.

Fixes #2217


ihcsim:
LGTM in terms of adding the new annotation to the CLI inject and webhook.

@klingerf @alpeb Thoughts?

controller/proxy-injector/webhook.go
```diff
@@ -166,6 +166,13 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
 		}
 		patch.addVolume(caBundle)
 		patch.addVolume(tlsSecrets)
+		patch.addPodAnnotations(map[string]string{
+			k8sPkg.IdentityModeAnnotation: k8sPkg.IdentityModeOptional,
```
siggy: i'm not seeing `linkerd.io/identity-mode` show up on auto-injected pods. i'm wondering if `patch.addPodAnnotations` is blowing away previous calls, rather than adding them? note that we only call `addPodLabels` and similar once, and while we call `addVolume` twice, it's implemented with a `/spec/template/spec/volumes/-`, and first requires calling `addVolume`

repro:

```bash
$ bin/linkerd install --tls optional --proxy-auto-inject

$ curl  curl https://run.linkerd.io/emojivoto.yml > emojivoto.yml
# add linkerd.io/inject: enabled annotation to emojivoto namespace
$ cat emojivoto.yml | kubectl apply

$ kubectl -n emojivoto get po -o jsonpath='{.items[*].metadata.annotations}'
map[linkerd.io/created-by:linkerd/proxy-injector git-9a915b45 linkerd.io/proxy-version:git-9a915b45] map[linkerd.io/created-by:linkerd/proxy-injector git-9a915b45 linkerd.io/proxy-version:git-9a915b45] map[linkerd.io/created-by:linkerd/proxy-injector git-9a915b45 linkerd.io/proxy-version:git-9a915b45 voting-pod-foo:voting-pod-bar] map[linkerd.io/created-by:linkerd/proxy-injector git-9a915b45 linkerd.io/proxy-version:git-9a915b45]
```

alpeb: @siggy That's correct. I fixed that in #2334. I'll try to extract a diff from there that can be applied here.

olix0r:
Thanks for all the feedback,

I realized that I don't think we _read_ this annotation while injecting -- i.e. we require that the CLI flag be set even if the annotation is present. I think we shouldn't require setting the flag if the annotation is set... Will follow up.

controller/proxy-injector/webhook.go
alpeb: Not the same I did in the other branch, but this will do for this one:
https://gist.github.com/alpeb/2ad3f5df7038eb2822a7901c18db4b34

pkg/k8s/labels.go
```diff
@@ -75,6 +75,20 @@ const (
...
+
+	// IdentityModeDisabled is assigned to IdentityModeAnnotation to
+	// disable the proxy from participating in automatic identity.
+	IdentityModeDisabled = "disabled"
+
+	// IdentityModeOptional is assigned to IdentityModeAnnotation to
+	// optionally configure the proxy to participate in automatic identity.
+	//
+	// Will be deprecated soon.
+	IdentityModeOptional = "optional"
```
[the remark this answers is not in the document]
olix0r: Actually, I'd prefer to do this later... these constants are just here to reflect the annotation values. I'll be tackling the CLI flags in a followup.

olix0r:
On further thought, I don't think this annotation should impact normal inject (yet). Let's do that in a followup.

controller/proxy-injector/webhook.go
```diff
@@ -153,6 +153,9 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
 		patch.addInitContainer(proxyInit)
 	}
 
+	if deployment.Spec.Template.Annotations == nil {
```
siggy: remove the second copy of this about 45 lines below.

controller/api/destination/endpoint_listener.go
```diff
@@ -204,46 +207,44 @@ func (l *endpointListener) toAddrSet(addresses []*updateAddress) *pb.AddrSet {
...
+	if l.enableTLS && controllerNS == l.controllerNS &&
+		pod.Annotations[pkgK8s.IdentityModeAnnotation] == pkgK8s.IdentityModeOptional {
+		name := pkgK8s.TLSIdentity{
+			Name:                ownerName,
+			Kind:                ownerKind,
+			Namespace:           pod.Namespace,
+			ControllerNamespace: controllerNS,
+		}.ToDNSName()
+
+		l.log.Debugf("getAddrMetadata(%+v): Owner: %s/%s DNS Name: %s Labels: %+v", pod, ownerKind, ownerName, name, labels)
```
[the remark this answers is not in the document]
siggy: (bump)

controller/api/destination/endpoint_listener_test.go
```diff
@@ -248,6 +257,57 @@ func TestEndpointListener(t *testing.T) {
 		}
 	})
 
+	t.Run("Does not send TlsIdentity to for other meshes", func(t *testing.T) {
+		expectedPodName := "pod1"
+		expectedPodNamespace := "this-namespace"
+		expectedControllerNamespace := "other-linkerd-namespace"
+		expectedPodDeployment := "pod-deployment"
```
siggy: goconst lint error, define this similar to `thisNS`.

controller/api/destination/endpoint_listener.go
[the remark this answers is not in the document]
olix0r: I'm going to skip the controller namespace, since that will never change between log lines.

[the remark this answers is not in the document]
olix0r: Ok, I took a bigger whack at improving debug logging by consolidating the logic in endpointListener.Update:

Now, on updates, we log both additions and removals (previously, only additions were logged). Furthermore, we print the pod name and IP (before the IP wasn't represented as a dotted quad), and then include the entire protobuf weightedaddr struct (rather than a few select fields):

```
time="2019-02-26T16:53:48Z" level=debug msg="Update: add=1; remove=0" component=endpoint-listener namespace=linkerd service=linkerd-prometheus
time="2019-02-26T16:53:48Z" level=debug msg="Update: add: pod=linkerd-prometheus-7bbc899687-nd9zt; addr=10.1.1.160; addr:<ip:<ipv4:167838112 > port:9090 > weight:1 metric_labels:<key:\"control_plane_ns\" value:\"linkerd\" > metric_labels:<key:\"deployment\" value:\"linkerd-prometheus\" > metric_labels:<key:\"pod\" value:\"linkerd-prometheus-7bbc899687-nd9zt\" > metric_labels:<key:\"pod_template_hash\" value:\"7bbc899687\" > protocol_hint:<h2:<> > " component=endpoint-listener namespace=linkerd service=linkerd-prometheus
```

controller/api/destination/endpoint_listener.go
```diff
@@ -131,29 +134,37 @@ func (l *endpointListener) SetServiceID(id *serviceID) {
...
+
 	if len(remove) > 0 {
-		update := &pb.Update{
-			Update: &pb.Update_Remove{
-				Remove: l.toAddrSet(remove),
-			},
+		// If pods were removed, send the list of IP addresses.
+		set := &pb.AddrSet{}
+		for _, a := range remove {
+			l.log.Debugf("Update: remove: pod=%s; addr=%s", a.pod.Name, a.pod.Status.PodIP)
```
siggy: This panics when a.pod is nil:
https://travis-ci.org/linkerd/linkerd2/jobs/498829807#L1169

This is interesting. Other than this log line, `updateAddress.pod` is not required on the `remove` param. In the longer term we may want to consider modifying this method signature to be explicit about it:
```go
func (l *endpointListener) Update(add []*updateAddress, remove []*net.TcpAddress)
```

In the short term I suppose you could do a nil check here, or update the tests to always include pods in the `remove` slice, or both. My reading of the non-test code paths indicates we can always expect the pod to be present.

Either way it may be helpful to call this out in the top-line function comment.

olix0r: nice catch!

olix0r:
Because the logging changes ended up rattailing, i've spun those off into https://github.com/linkerd/linkerd2/pull/2389

controller/api/destination/endpoint_listener_test.go
```diff
@@ -248,11 +261,62 @@ func TestEndpointListener(t *testing.T) {
 		}
 	})
 
+	t.Run("Does not send TlsIdentity to for other meshes", func(t *testing.T) {
```
siggy: `Does not send TlsIdentity for other meshes`

PR #2372, opened 2019-02-24
inject: Use k8s templating for identities

olix0r:
linkerd/linkerd2-proxy#199 changes the proxy to accept
`LINKERD2_PROXY_LOCAL_IDENTITY` without namespace templating.

This changes inject to set LOCAL_IDENTITY instead of POD_IDENTITY and,
in the case of cli inject, uses kubernetes templating to materialize the
identity into the environment.

### DIFF
--- a/chart/templates/proxy_injector.yaml
+++ b/chart/templates/proxy_injector.yaml
@@ -168,10 +168,6 @@ data:
       value: tcp://0.0.0.0:{{.Values.InboundPort}}
     - name: LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES
       value: {{.Values.ProfileSuffixes}}
-    - name: LINKERD2_PROXY_POD_NAMESPACE
-      valueFrom:
-        fieldRef:
-          fieldPath: metadata.namespace
     {{- if .Values.InboundAcceptKeepaliveMs }}
     - name: LINKERD2_PROXY_INBOUND_ACCEPT_KEEPALIVE
       value: {{.Values.InboundAcceptKeepaliveMs}}ms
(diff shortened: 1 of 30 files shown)


olix0r:
requires https://github.com/linkerd/linkerd2/pull/2371 as well

chart/templates/proxy_injector.yaml
```diff
@@ -168,10 +168,6 @@ data:
       value: tcp://0.0.0.0:{{.Values.InboundPort}}
     - name: LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES
       value: {{.Values.ProfileSuffixes}}
-    - name: LINKERD2_PROXY_POD_NAMESPACE
-      valueFrom:
-        fieldRef:
-          fieldPath: metadata.namespace
```
alpeb: We'll stop using `proxy_injector.yaml` in #2334 but it appears you missed replacing this with `K8S_NS`?

controller/proxy-injector/fake/data/config-proxy.yaml
```diff
@@ -1,66 +1,62 @@
 env:
-- name: LINKERD2_PROXY_LOG
-  value: warn,linkerd2_proxy=info
```
alpeb: Looks like your editor is inserting some extra spaces that changed all the lines. Same for inject-sidecar-container-spec.yaml.

olix0r: eh, thanks. i'll fix that.

cli/cmd/inject.go
```diff
@@ -237,6 +236,7 @@ func injectPodSpec(t *v1.PodSpec, identity k8s.TLSIdentity, controlPlaneDNSNameO
 	if options.disableExternalProfiles {
 		profileSuffixes = "svc.cluster.local."
 	}
+	identity.Namespace = fmt.Sprintf("$(%s)", PodNamespaceEnvVarName)
```
siggy: why not just set this when the `identity` struct is instantiated, prior to calling `injectPodSpec`?

olix0r: the intent is that this is where the environment is actually instrumented, so this should  be where that templating logic belongs. if it's encoded where the identity struct is instantiated, it seems like a random implementation detail that we have to assume is correct.

More generally, that identity struct is going to be removed/changed substantially over the next few PRs.

olix0r:
I'm going to pause this branch for now since it's not a hard blocker for identity work. I'll revisit this when polishing.

Issue #2337, opened 2019-02-20
Focus install by role and stage


grampelberg:
### Task Breakdown

- [x] #2719 introduce 2 new subcommands: `admin` and `user`
- [x] #2719 finalize stage naming
- [x] #2719 templatize stages
- [x] #2723 corresponding `linkerd upgrade` updates
- [x] #2765 corresponding `linkerd check` updates
- [x] #2474 `--single-namespace` removal
- [x] #2719 unit tests for subcommands
- [x] #2723 integration tests for subcommands

### Tasks we may or may not do

- [ ] disable `linkerd install`
- [ ] `install-sp` integration
- [ ] `install-cni` integration

---

# Previous

### What problem are you trying to solve?

Today, Linkerd has a all-in-one install via `linkerd install`. As features have been added, this is becoming to be a larger problem because:

- CRDs and their resource definitions cannot be applied at the same time. This has created a new `install-sp` command that users need to run after `install` if they would like service profiles for the control plane. While control plane service profiles are optional right now, it is likely that they will be required in the future to support new functionality.
- Teams run on locked down clusters where there are two different individuals. One is a cluster admin that can add things such as CRDs and ClusterRoles and one is a namespace admin that can add Roles and Deployments. There is a hypothesis that cluster admins do not want to operate the components of Linkerd, simply audit that namespace admins won't hurt the cluster.
- The CNI plugin introduces a step where it must be installed before the control plane components themselves.

### How should the problem be solved?

To address these issues, the install should be separated into multiple phases. By using `install` subcommands to implement these phases, `linkerd check --pre` can be integrated into the install experience. As a tldr:

```bash
$ linkerd install config | kubectl apply -f -
$ linkerd install components | kubectl apply -f -
```

#### Pre-install experience

- Move `check --pre` into the existing `install` command.
- Order `linkerd-version` checks first instead of last.

For users without `cluster-admin`:

```bash
$ linkerd install
linkerd-version
---------------
√ can determine the latest version
√ cli is up-to-date

kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version

pre-kubernetes-cluster-setup
----------------------------
√ control plane namespace does not already exist
× can create Namespaces
× can create ClusterRoles
× can create ClusterRoleBindings
× can create CustomResourceDefinitions

pre-kubernetes-setup
--------------------
√ can create ServiceAccounts
√ can create Services
[...]

PR #2401, opened 2019-02-27
Introduce multi-stage install

siggy:
The existing `linkerd install` requires the user to have cluster-wide
permissions, to install CRDs, ClusterRoles, etc.

This change introduces two install subcommands, specific to the user's
privilege level:
- `linkerd install admin`
  - ServiceProfile CRD
  - ClusterRole[Bindings]s
  - control-plane Namespace
- `linkerd install user`
  - ServiceAccounts
  - Role[Binding]s
  - control-plane ConfigMaps
  - control-plane Deployments
  - control-plane Services

The existing `linkerd install` retains existing behavior, modulo some
reordering of components, to render as a concatenation of `admin` and
`user`.

TODO:
- enable/disable entire template files based on install flags
- corresponding `linkerd check` updates
- `--single-namespace` removal
- unit and integration tests for subcommands
- `install-sp` integration
- `install-cni` integration

Part of #2337, relates to #2164

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -23,6 +23,8 @@ import (
 )
 
 type installConfig struct {
+	stage string
+
 	Namespace                   string
 	ControllerImage             string
 	WebImage                    string
@@ -71,18 +73,12 @@ type installOptions struct {
 }
 
 const (
+	admin = "admin"
+	user  = "user"
+
 	prometheusProxyOutboundCapacity = 10000
 	defaultControllerReplicas       = 1
 	defaultHAControllerReplicas     = 3
-
-	nsTemplateName             = "templates/namespace.yaml"
-	controllerTemplateName     = "templates/controller.yaml"
-	webTemplateName            = "templates/web.yaml"
-	prometheusTemplateName     = "templates/prometheus.yaml"
-	grafanaTemplateName        = "templates/grafana.yaml"
-	serviceprofileTemplateName = "templates/serviceprofile.yaml"
-	caTemplateName             = "templates/ca.yaml"
-	proxyInjectorTemplateName  = "templates/proxy_injector.yaml"
 )
 
 func newInstallOptions() *installOptions {
@@ -100,13 +96,30 @@ func newInstallOptions() *installOptions {
 
 func newCmdInstall() *cobra.Command {
 	options := newInstallOptions()
-
 	cmd := &cobra.Command{
-		Use:   "install [flags]",
-		Short: "Output Kubernetes configs to install Linkerd",
-		Long:  "Output Kubernetes configs to install Linkerd.",
+		Use:       "install [admin|user] [flags]",
+		Args:      cobra.OnlyValidArgs,
+		ValidArgs: []string{"admin", "user"},
+		Short:     "Output Kubernetes configs to install Linkerd",
+		Long: `Output Kubernetes configs to install Linkerd.
+
+This command provides Kubernetes configs necessary to install the Linkerd
+control-plane.`,
+		Example: `  # Default install.
+  linkerd install | kubectl apply -f -
+
+  # Installation may also be broken up into two stages by user privilege.
+  linkerd install admin | kubectl apply -f -
+  linkerd install user | kubectl apply -f -
+
+  # Install Linkerd into a non-default namespace.
+  linkerd install admin -l linkerdtest | kubectl apply -f -
+  linkerd install user -l linkerdtest | kubectl apply -f -
+
+  # Install Linkerd with only namespace-wide privileges.
+  linkerd install user --single-namespace -l myapp | kubectl apply -f -`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			config, err := validateAndBuildConfig(options)
+			config, err := validateAndBuildConfig(args, options)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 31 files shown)


cli/cmd/install.go
```diff
@@ -126,13 +122,30 @@ func newInstallOptions() *installOptions {
 
 func newCmdInstall() *cobra.Command {
 	options := newInstallOptions()
-
 	cmd := &cobra.Command{
-		Use:   "install [flags]",
-		Short: "Output Kubernetes configs to install Linkerd",
-		Long:  "Output Kubernetes configs to install Linkerd.",
+		Use:       "install [admin|user] [flags]",
```
adleong: I find these names a bit misleading, we're not installing an admin or a user.  I'd expect this to be something more explicit like `linkerd install prepare` and `linkerd install control` or `linkerd install control-plane`.  

But if the shed has sailed on these names then :+1:

cli/cmd/testdata/install_single_namespace_output.golden
```diff
@@ -1,6 +1,12 @@
+
```
adleong: is this whitespace intentional?

siggy:
The goal of a single `linkerd install` command would be strictly as a concatenation of the two subcommands (we could even add a test to ensure this stays true). Here are the tradeoffs (@grampelberg would like your thoughts here):

### multi-stage install only
- one install flow for all users
- the control-plane's `ServiceProfile`s may be installed at `kubectl apply`-time, rather than programmatically at startup

### multi-stage install and single `linkerd install`
- installation remains a single command for users/developers with cluster-wide access
- installing `ServiceProfile`s programmatically at startup eliminates timing issues between install stages. i know as a developer i would run:
    ```
    (linkerd install admin && linkerd install user) | kubectl apply -f -
    ```
   ... and it would break half the time

For the purpose of this PR, let's keep the single command available. Of the remaining TODOs, it will only hold up `install-sp` integration.

Clarifying some related requirements, if we assume that `linkerd install admin` is always run prior to `linkerd install user`, we can eliminate all Role[Binding]s in the 2nd stage, and also simplify the Kubernetes client builder code to always assume cluster-wide access and ServiceProfile CRD availability.

Regarding the `admin` / `user` naming. My goals with the name were to convey:
- privilege level associated with each command
- ordering

`linkerd install control-plane` as a 2nd stage clearly conveys what is being installed. `prepare` feels a bit optional and vague as a 1st stage. Maybe `linkerd install cluster`? We can also bikeshed this after the PR merges.

/cc @adleong @olix0r @wmorgan

grampelberg:
> Leaving linkerd install as a concatenation of the two subcommands makes me nervous. What happens when we want to integrate install-sp into the second subcommand?

I think this is a solid small step in the right direction. We definitely need to figure out how to handle `install-sp`, but it doesn't have to happen in the PR. @siggy was planning on doing some extra digging to understand if it would be possible for a single install command that includes the service profiles.

I'm still moderately concerned about adding the service profiles at run time, given the added magic that is there. As the control plane profiles are still optional as of right now, it might not be a problem we *need* to solve in the short term until there's a solid reason for requiring them.

> linkerd install control-plane as a 2nd stage clearly conveys what is being installed. prepare feels a bit optional and vague as a 1st stage. Maybe linkerd install cluster? We can also bikeshed this after the PR merges.

I don't mind bikeshedding post-merge, but I'd like to have the first edge release this is included in to have some more consensus first. My suggestion is still:

- `linkerd install config`
- `linkerd install components`

olix0r:
I'll echo Thomas's concerns about creating resources at runtime when we don't strictly need to. I'm +1 on merging and resolving these considerations before 2.3 is released. The `install` command being broken into two steps doesn't strike me as all that onerous, personally.

cli/cmd/testdata/install_single_namespace_output.golden
siggy: it's expected but not ideal. subsequent PRs will remove single namespace template wrapping, and introduce templatizing via stage, we can clean this up then.

cli/cmd/install.go
siggy: i'm not strictly tied to admin/user, but we can decide after merge.

siggy:
@olix0r @grampelberg Captured a few more tasks in #2337, laying them out here for reference, and assuming they all can be hashed out post-merge:
- disable single command `linkerd install`
- sort out stage naming
- in-line service profiles (in general i agree the fewer objects we magically deploy, the better. ideally `(linkerd install admin && linkerd install user) | kubectl delete -f -` would remove all of Linkerd).

I want to call out again that once we in-line service profiles into the second stage install, this command will start breaking some of the time:
```bash
(linkerd install admin && linkerd install user) | kubectl apply -f -
```
We can certainly address this via docs, but it will affect how we script and test Linkerd installation.

grampelberg:
@siggy that all sounds great to me. Are you planning on updating `check` post-merge? Feels like we should settle on the subcommand naming first.

I'll get something together so we can get a consensus on the naming!

siggy:
@grampelberg I'm assuming the tasks listed in #2337 fall out of scope for this PR.

grampelberg:
@siggy that's what I get for forgetting about that list. Thank you! Looks great to me =)

siggy:
closing for now, will either reopen for 2.4 release, or start fresh.

Issue #2224, opened 2019-02-07
Show TCP metrics


grampelberg:
### What problem are you trying to solve?

For binary protocols or services using HTTPS, it can be confusing to see zero stats in the CLI and GUI. There are TCP stats, they're just only accessible in the grafana dashboards right now.

### How should the problem be solved?

Introduce a CLI and GUI panel for TCP metrics that shows whether the protocol has been detected or not and the general metrics. See #2223 for the other side of this.

```
$ linkerd tcp
```

### Any alternatives you've considered?

Maybe as a flag to `stat`? The main reason for not going that route is that the headers will be totally different.

grampelberg:
What is the list of TCP metrics that we have?

klingerf:
@grampelberg Here's what I see for TCP metrics in a pod running MySQL:

* **tcp_open_total** Total count of opened connections
* **tcp_open_connections** Number of currently-open connections
* **tcp_read_bytes_total** Total count of bytes read from peers
* **tcp_write_bytes_total** Total count of bytes written to peers
* **tcp_close_total** Total count of closed connections
* **tcp_connection_duration_ms** Connection lifetimes

<details><summary>Full metrics output</summary>

```
# HELP tcp_open_total Total count of opened connections
# TYPE tcp_open_total counter
tcp_open_total{direction="inbound",peer="src",tls="no_identity",no_tls_reason="not_http"} 24
tcp_open_total{direction="inbound",peer="dst",tls="no_identity",no_tls_reason="not_http"} 24
# HELP tcp_open_connections Number of currently-open connections
# TYPE tcp_open_connections gauge
tcp_open_connections{direction="inbound",peer="src",tls="no_identity",no_tls_reason="not_http"} 21
tcp_open_connections{direction="inbound",peer="dst",tls="no_identity",no_tls_reason="not_http"} 21
# HELP tcp_read_bytes_total Total count of bytes read from peers
# TYPE tcp_read_bytes_total counter
tcp_read_bytes_total{direction="inbound",peer="src",tls="no_identity",no_tls_reason="not_http"} 281477
tcp_read_bytes_total{direction="inbound",peer="dst",tls="no_identity",no_tls_reason="not_http"} 1198068
# HELP tcp_write_bytes_total Total count of bytes written to peers
# TYPE tcp_write_bytes_total counter
tcp_write_bytes_total{direction="inbound",peer="src",tls="no_identity",no_tls_reason="not_http"} 1198068
[...]

PR #2408, opened 2019-02-27
Show TCP stats in `linkerd stat` when -o wide or -o json is used

rmars:
Show TCP stats in the linkerd stat output. They are not shown by default, but will be queried when using `-o wide` or `-o json`.

Also display read/write bytes as bytes per sec in the CLI and dashboard.

```
$ go run cli/main.go stat deploy -n linkerd -o wide
NAME                 MESHED   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TLS   TCP CONNECTIONS   READ BYTES/SEC   WRITE BYTES/SEC
linkerd-controller      1/1   100.00%   0.9rps           1ms           8ms          10ms    0%                18         979.9bps          968.5bps
linkerd-grafana         1/1   100.00%   0.3rps         250ms         485ms         497ms    0%                 2         347.9bps          344.1bps
linkerd-prometheus      1/1   100.00%   0.2rps           1ms           3ms           3ms    0%                 7          51.5bps           47.7bps
linkerd-web             1/1   100.00%   0.3rps           2ms           3ms           3ms    0%                 2         332.2bps          328.4bps
```

Part of #2224


cli/cmd/stat.go
```diff
@@ -413,6 +453,12 @@ func printStatJSON(statTables map[string]map[string]*row, w *tabwriter.Writer) {
 					entry.LatencyMSp95 = &stats[key].latencyP95
 					entry.LatencyMSp99 = &stats[key].latencyP99
 					entry.TLS = &stats[key].tlsPercent
+
+					if options.showTCPStats {
```
rmars: I'm actually wondering whether we should just query tcp stats by default if `-o json` is specified, so that we don't have these weird null values

grampelberg: I was thinking maybe `-o wide` and `-o json`? I'd rather manage it that way than a new flag. In the future, we'll be able to do something like `kubectl` does with custom columns.

rmars: yeah I think auto including it with `-o wide` makes sense

cli/cmd/stat.go
```diff
@@ -326,9 +337,19 @@ func printSingleStatTable(stats map[string]*row, resourceType string, w *tabwrit
...
 
+	if showTCPStats(options, resourceType) {
+		headers = append(headers, []string{
+			"TCP CONNECTIONS",
+			"READ BYTES",
+			"WRITE BYTES",
+		}...)
+	}
+
+	headers = append(headers, "\t") // trailing \t is required to format last column
```
siggy: this renders an extra column in the header, instead how about:
```go
headers[len(headers)-1] = headers[len(headers)-1] + "\t" // trailing \t is required to format last column
```

adleong:
Are the read bytes and write bytes a rate or a total?  Windowed over some time period?  Or total over lifetime of the current connections?  Or total over the lifetime of the pod?  Do these counters reset if Prometheus restarts (unlike our other rate-based stats which only look at a recent window of data and can therefore recover from a Prometheus restart)?

cli/cmd/stat.go
adleong: I think I'm agreeing with what you're both saying but I think we should always include all fields in json output, since width is not really a concern.

rmars:
According to https://linkerd.io/2/reference/proxy-metrics/,  ```tcp_read_bytes_total: A counter of the total number of received bytes. This is updated when the connection closes.``` (So total over the connection lifetime).

The query we do to get the stats is windowed, though, (over the time_window specified in the stat query) so it looks like this:
```sum(increase(tcp_read_bytes_total{direction="inbound", namespace="default"}[1m])) by (namespace, deployment)```

Because they're windowed, I believe they'd recover like our other stats.

File: cli/cmd/stat.go [hunk 3]
```diff
@@ -452,7 +452,7 @@ func printStatJSON(statTables map[string]map[string]*row, w *tabwriter.Writer, o
 					entry.LatencyMSp99 = &stats[key].latencyP99
 					entry.TLS = &stats[key].tlsPercent
 
-					if showTCPStats(options, resourceType) {
+					if resourceType != k8s.Authority {
```
siggy: i like that we're no longer passing `statOptions` into this function, but it does mean the logic to determine whether to print TCP stats lives in 2 places. i'd prefer to continue calling `showTCPStats` here, as it takes into account `outputFormat` already.

adleong:
@rmars in that case, this is really bytes per window, right?  I feel like we should call that out in the column name somehow (bytes per 10m) or convert to bytes per second like we do for requests.

rmars:
@adleong yeah that's a great point. Currently the column is called `READ BYTES` so we could either do `READ BYTES/MIN` or just divide the bytes by 60 and display them as `63472bps` or `63472b/sec`. The column headers will probably be the clearest (though for RPS we do call the column RPS and also have the unit RPS in there). If there's no preference I will go the rps route.

grampelberg:
@rmars with zero context, my preference would be BPS for consistency.

web/app/js/components/util/Utils.js
```diff
@@ -55,7 +55,7 @@ export const metricToFormatter = {
   "SUCCESS_RATE": m => _isNil(m) ? "---" : successRateFormatter(m),
   "LATENCY": formatLatencyMs,
   "UNTRUNCATED": m => styleNum(m, "", false),
-  "BYTES": m => _isNil(m) ? "---" : styleNum(m, "B", true),
+  "BYTES": m => _isNil(m) ? "---" : styleNum(m, "bps", true),
```
siggy: naming nit: `bps` is typically bits per second. consider `B/s` 
https://en.wikipedia.org/wiki/Data-rate_units

rmars: oooh thanks!

rmars: changing this to B/s as it seems more standard
cc @grampelberg

PR #2414, opened 2019-02-28
Introduce logging and restart integration tests

### LINKED ISSUE #2348: Integration tests should validate control plane container logs

When manually vetting a Linkerd build, it's easy to forget to look in the control plane's logs for warnings or errors that might be indicative of a problem. It's frequently the case that control plane components are spewing warnings or errors, but we don't notice them until after the code has already been released (e.g. #2331).

Update `test/install_test.go` to start validating control plane logs, after the smoke test has run. In a fresh control plane install, there should not be any warnings or errors (if there are, we should fix that). So it should be possible for us to fail integration tests if any control plane component (including linkerd-proxy) emits warnings or errors.

Manual validation, but we often forget to do that.

siggy:
The integration tests deploy complete Linkerd environments into
Kubernetes, but do not check if the components are logging errors or
restarting.

Introduce integration tests to validation that all expected
control-plane containers (including `linkerd-proxy` and `linkerd-init`)
are found, logging no errors, and not restarting.

Fixes #2348

# Test examples

## TestLogs

Prior to adding the `knownProxyErrors` check, linkerd-proxy fails with TLS:
```bash
$ go test github.com/linkerd/linkerd2/test -run ^TestLogs$ -enable-tls -linkerd `pwd`/target/cli/darwin/linkerd -integration-tests
--- FAIL: TestLogs (11.01s)
    install_test.go:306: Found error in web/linkerd-proxy log: l5d-integration-tls linkerd-web-5cbdb887fd-hqszx linkerd-proxy ERR! admin={bg=tls-config} linkerd2_proxy::transport::tls::config error loading /var/linkerd-io/trust-anchors/trust-anchors.pem: No such file or directory (os error 2)
FAIL
FAIL	github.com/linkerd/linkerd2/test	11.111s
```

Modified:
```go
proxyRegex := regexp.MustCompile(fmt.Sprintf("%s (ERR|WARN|DBUG)", proxyContainer))
```

Returns:
```bash
$ go test github.com/linkerd/linkerd2/test -run ^TestLogs$ -linkerd `pwd`/target/cli/darwin/linkerd -integration-tests
--- FAIL: TestLogs (2.04s)
    install_test.go:313: Found error in controller/linkerd-proxy log: l5d-integration linkerd-controller-d67b9b4c9-7ks4s linkerd-proxy DBUG proxy={server=out listen=127.0.0.1:4140 remote=10.1.2.24:48158} linkerd2_proxy::transport::metrics::io client connection open
FAIL
FAIL	github.com/linkerd/linkerd2/test	2.288s
```

Modified:
```go
controllerRegex := regexp.MustCompile("level=(panic|fatal|error|warn|info)")
```

Returns:
```bash
$ go test github.com/linkerd/linkerd2/test -run ^TestLogs$ -linkerd `pwd`/target/cli/darwin/linkerd -integration-tests
--- FAIL: TestLogs (0.50s)
    install_test.go:306: Found error in controller/destination log: l5d-integration linkerd-controller-d67b9b4c9-7ks4s destination time="2019-02-27T22:29:51Z" level=info msg="running version git-9c0537c3"
FAIL
FAIL	github.com/linkerd/linkerd2/test	0.654s
```

## TestRestarts

On a recent master, with unexpected restarts:

```bash
$ kubectl -n l5d-integration get po
NAME                                 READY     STATUS    RESTARTS   AGE
linkerd-controller-d67b9b4c9-7ks4s   4/4       Running   3          3h
linkerd-grafana-86b96ddb67-kjzts     2/2       Running   0          3h
linkerd-prometheus-f4f6986db-7skmj   2/2       Running   0          3h
linkerd-web-c47b5b94-kqkdp           2/2       Running   1          3h
```

```
$ go test github.com/linkerd/linkerd2/test -run ^TestRestarts$ -linkerd `pwd`/target/cli/darwin/linkerd -integration-tests
--- FAIL: TestRestarts (0.36s)
    install_test.go:338: Found '1' restarts on controller/destination
FAIL
FAIL	github.com/linkerd/linkerd2/test	0.619s
```

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -57,6 +57,24 @@ var (
 		`linkerd-proxy ERR! admin={bg=tls-config} linkerd2_proxy::transport::tls::config error loading /var/linkerd-io/trust-anchors/trust-anchors.pem: No such file or directory (os error 2)`,
 		`linkerd-proxy WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io("/var/linkerd-io/identity/certificate.crt", Some(2)), falling back`,
 		`linkerd-proxy WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io("/var/linkerd-io/trust-anchors/trust-anchors.pem", Some(2)), falling back`,
+
+		// k8s hitting `:3000/api/health` before grafana is ready
+		`linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:3000)`,
+
+		// k8s hitting `:9994/ready` before web is ready
+		`linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:9994)`,
+
+		// k8s hitting `:9995/ready` before public-api is ready
+		`linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:9995)`,
+
+		// k8s hitting `:9996/ready` before destination is ready
+		`linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:9996)`,
+
+		// k8s hitting `:9997/ready` before ca is ready
+		`linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:9997)`,
+
+		// k8s hitting `:9998/ready` before tap is ready
+		`linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:9998)`,
 	}
 )
 
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -49,33 +49,24 @@ var (
 		"linkerd-web":        {1, []string{"web"}},
 	}
 
-	// linkerd-proxy logs some errors when TLS is enabled, remove these once
+	// Linkerd commonly logs these errors during testing, remove these once
 	// they're addressed.
-	knownProxyErrors = []string{
-		`linkerd2_proxy::control::serve_http error serving metrics: Error { kind: Shutdown, cause: Os { code: 107, kind: NotConnected, message: "Transport endpoint is not connected" } }`,
-		`linkerd-proxy ERR! admin={bg=tls-config} linkerd2_proxy::transport::tls::config error loading /var/linkerd-io/identity/certificate.crt: No such file or directory (os error 2)`,
-		`linkerd-proxy ERR! admin={bg=tls-config} linkerd2_proxy::transport::tls::config error loading /var/linkerd-io/trust-anchors/trust-anchors.pem: No such file or directory (os error 2)`,
-		`linkerd-proxy WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io("/var/linkerd-io/identity/certificate.crt", Some(2)), falling back`,
-		`linkerd-proxy WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io("/var/linkerd-io/trust-anchors/trust-anchors.pem", Some(2)), falling back`,
+	knownErrorsRegex = regexp.MustCompile(strings.Join([]string{
+		// TLS not ready at startup
+		`.*-tls linkerd-(ca|controller|grafana|prometheus|web)-.*-.* linkerd-proxy ERR! admin={bg=tls-config} linkerd2_proxy::transport::tls::config error loading /var/linkerd-io/identity/certificate\.crt: No such file or directory \(os error 2\)`,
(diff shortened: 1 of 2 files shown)


ihcsim:
@siggy Re-running the integration test on my local (with edge-19.2.5) the second time, I'm seeing these proxy errors. Do you know if they are expected? 
```
--- FAIL: TestLogs (12.50s)
    install_test.go:329: Found error in grafana/linkerd-proxy log: l5d-integration-tls linkerd-grafana-7b6577675c-45j8x linkerd-proxy ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.12:42282} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:3000)
```

```
--- FAIL: TestLogs (12.00s)
    install_test.go:329: Found error in controller/destination log: l5d-integration-single-namespace linkerd-controller-6c4f74466b-7vlh6 destination time="2019-03-01T00:36:35Z" level=warning msg="Not authorized for cluster-wide access, limiting access to \"l5d-integration-single-namespace\" namespace"
```

If yes, then we'll have to update the `knownProxyErrors` slice. The pods in both the `l5d-integration-tls` and `l5d-integration-single-namespace` namespaces look healthy with no restarts.

siggy:
@ihcsim Nice catches! It looks like the tests are working _too_ well. ;)

I hadn't seen that proxy error before. Curious that it references `127.0.0.1:3000`, something Grafana related? Had the cluster been up and running for a while? Were you accessing the dashboard?

That single namespace warning in controller/destination is expected when running in single-namespace mode, and now I'm trying to determine how this ever passed on my system. It's possible I had permissive ClusterRoles installed that allowed this to pass. I'll have a look.

ihcsim:
> Had the cluster been up and running for a while? Were you accessing the dashboard?

I spin up a new VM for that run. Nothing else was going on besides the intg tests. FWIW, I re-ran the integration test on another VM with more CPUs and memory, the grafana problem disappeared, but got a different error:
```
--- FAIL: TestLogs (11.00s)
    install_test.go:329: Found error in web/web log: l5d-integration-tls linkerd-web-7f9746c947-cqfnn web time="2019-03-01T05:55:31Z" level=error msg="Post http://linkerd-controller-api.l5d-integration-tls.svc.cluster.local:8085/api/v1/Version: context canceled"
```
Sounds like happens only on my machine.

But the single namespace warning was definitely still there.

siggy:
@ihcsim I just reproduced the grafana 127.0.0.1:3000 error on my system. The `/api/v1/Version: context canceled` issue is also something that happens on ci occasionally. I'll keep digging, thanks for testing!

siggy:
@ihcsim Update on each of the issues you were seeing:
- `Connection refused (os error 111) (address: 127.0.0.1:`
  - these are k8s readiness checks hitting the containers before they are ready, i've added an explicit known error for each control-plane component
- `Not authorized for cluster-wide access`
  - i'm not able to repro this, but this check is going away soon, in a follow up to #2401
- `/api/v1/Version`
  - not able to repro this. it could be from failed `linkerd check` commands, but i'm not sure how those could should on the `web` container. i'll keep an eye out for this.

test/install_test.go
```diff
@@ -261,3 +300,85 @@ func TestCheckProxy(t *testing.T) {
...
+			outputStream, err := TestHelper.LinkerdRunStream(
+				"logs", "--no-color",
+				"--control-plane-component", deploy,
+				"--container", container,
+			)
+			if err != nil {
+				t.Fatalf("Error running command:\n%s", err)
+			}
+			defer outputStream.Stop()
+			outputLines, _ := outputStream.ReadUntil(100, 10*time.Second)
```
klingerf: For control plane containers that don't log at least 100 lines under normal testing conditions (e.g `tap` and `web`), this is going to add an additional 30 seconds to each test run (10 seconds each for standard install, tls, and single-namespace). And for other containers that log more than 100 lines, we won't be inspecting all of the logs emitted by the container. I think it would be better to fetch more log lines over a shorter window. How about:

```suggestion
			outputLines, _ := outputStream.ReadUntil(10000, time.Second)
```

Also, is skipping the error check here intentional since `outputLines` will be empty in the event of an error, and you're already checking that? Might be worth adding a comment.

test/install_test.go
```diff
@@ -261,3 +300,85 @@ func TestCheckProxy(t *testing.T) {
...
+		deploy := strings.TrimPrefix(deploy, "linkerd-")
+		containers := append(spec.containers, proxyContainer, initContainer)
+
+		for _, container := range containers {
+			selector := fmt.Sprintf("linkerd.io/control-plane-component=%s", deploy)
+			containerStatus := "containerStatuses"
+			if container == initContainer {
+				containerStatus = "initContainerStatuses"
+			}
+			output := fmt.Sprintf("jsonpath='{.items[*].status.%s[?(@.name==\"%s\")].restartCount}'", containerStatus, container)
```
klingerf: Ah interesting. Instead of shelling out to Kubernetes for the restart check, I was originally envisioning that we'd add a new test helper to do this programmatically, in `testutil/kubernetes_helper.go`, similar to what `CheckPods` is doing. In fact, how would you feel about just updating the `CheckPods` to also validate that none of the containers have restarted, and then calling that here after the rest of the tests in this file have run? In might take a small refactor, but I think it would be useful generally, and could be handy in our other tests that call `CheckPods`.

test/install_test.go
siggy: Yeah swallowing the error in case the timeout get hits (which now it consistently does). I've set the params to `(10000, 2*time.Second)`, with `(10000, 2*time.Second)` I was seeing it timeout before receiving any log lines.

PR #2422, opened 2019-03-01
Address review feedback on injection consolidation

klingerf:
This branch addresses my feedback on #2334, and it's expected to merge into that PR before that PR merges to master.


cli/cmd/root.go
```diff
@@ -399,6 +392,14 @@ func (options *proxyConfigOptions) enableTLS() bool {
 	return options.tls == optionalTLS
 }
 
+func (options *proxyConfigOptions) proxyImageOverride() string {
+	return strings.Replace(options.proxyImage, defaultDockerRegistry, options.dockerRegistry, 1)
+}
+
+func (options *proxyConfigOptions) initImageOverride() string {
+	return strings.Replace(options.initImage, defaultDockerRegistry, options.dockerRegistry, 1)
+}
```
alpeb: How about having just one function that would cater for any image (would be called when building the proxy, init and cni plugin images)
```
func imageOverride(name string) string {
	return strings.Replace(name, defaultDockerRegistry, options.dockerRegistry, 1)
}
```

klingerf: Good call! I added this in 40786464. And I restored the type switch in inject.go in db3e5e58.

PR #2424, opened 2019-03-01
Manual Helm chart updates

### LINKED ISSUE #2402: Create manual helm chart

We're slowly working on getting `install` to use an actual helm chart. Unfortunately, #2385 is currently blocking the process. This is because the control plane has proxies injected and the only way to inject proxies currently is via the `inject` code which isn't usable from helm.

In the mean time, there's no reason why the current templates, https://github.com/linkerd/linkerd2/tree/master/chart can't be used. They'll just need to have the proxy manually added to the YAML. This new chart can then be submitted as a PR to https://github.com/helm/charts for use by anyone.

cohix:
Add proxy and init containers to helm chart

Problem:
Helm chart does not inject sidecar or init containers for control plane pods.

Solution:
Added container specs for proxy and init containers to the helm chart manifests for control, web, prometheus, and grafana to allow for helm install. This also adds a default `values.yaml` file so that helm install can be used without any `--set`s (because you would need about 50 of them...)

Validation:
`helm template . | kubectl apply -f -` creates a working linkerd control plane. Tested on Docker for Mac k8s.

Note: this uses the current edge build, I am not sure if this is the best approach or not, advice appreciated.

Fixes #2402


chart/templates/controller.yaml
```diff
@@ -203,6 +200,67 @@ spec:
         {{- end }}
         securityContext:
           runAsUser: {{.Values.ControllerUID}}
+      - name: linkerd-proxy
```
grampelberg: I would like to see these as a template as they're not being modified on a per-resource basis.

grampelberg: @klingerf @siggy how does this interact with the CLI doing inject as well? Should these be behind an if when the CLI is rendering these files?

chart/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+InboundPort: 4143
+Namespace: linkerd
+NoInitContainer: false
+OutboundConnectKeepaliveMs: 10000
+OutboundPort: 4140
+ProfileSuffixes: .
+PrometheusImage: prom/prometheus:v2.7.1
+PrometheusVolumeName: data
+ProxyAutoInjectEnabled: false
+ProxyConfig: '{"proxyImage":{"imageName":"gcr.io/linkerd-io/proxy:edge-19.2.5","pullPolicy":"IfNotPresent"},"proxyInitImage":{"imageName":"gcr.io/linkerd-io/proxy-init:edge-19.2.5","pullPolicy":"IfNotPresent"},"destinationApiPort":{"port":8086},"controlPort":{"port":4190},"ignoreInboundPorts":[],"ignoreOutboundPorts":[],"inboundPort":{"port":4143},"metricsPort":{"port":4191},"outboundPort":{"port":4140},"resource":{"requestCpu":"","requestMemory":"","limitCpu":"","limitMemory":""},"proxyUid":"2102","logLevel":{"level":"warn,linkerd2_proxy=info"},"disableExternalProfiles":false}'
```
grampelberg: Can this come from a file instead? Maybe as YAML that helm renders as json? The ugly json string is hard to reason about and version. I think this'll work:

```
{{ .Values.ProxyConfig | toJson }}
```

chart/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+SingleNamespace: false
+TLSCertFileName: certificate.crt
+TLSIdentityVolumeSpecFileName: linkerd-secrets.yaml
+TLSPrivateKeyFileName: private-key.p8
+TLSSecretsVolumeName: linkerd-secrets
+TLSTrustAnchorConfigMapName: linkerd-ca-bundle
+TLSTrustAnchorFileName: trust-anchors.pem
+TLSTrustAnchorVolumeName: linkerd-trust-anchors
+TLSTrustAnchorVolumeSpecFileName: linkerd-trust-anchors.yaml
+UUID: 1e0f82c4-0b28-42b9-ba3c-c7787b8b7d58
```
grampelberg: This will need to be auto-generated as it is unique per control plane.

chart/values.yaml
```diff
@@ -0,0 +1,49 @@
+CliVersion: linkerd/cli edge-19.2.5
```
grampelberg: It would be nice to group and comment on these. I don't think it is a requirement for this PR as I can imagine the `install` code would change. That said, I'd love to see something like:

```yaml
images:
    proxy: gcr.io/linkerd-io/proxy:edge-19.2.5
    controller: gcr.io/linkerd-io/controller:edge-19.2.5

tls:
    certFileName: certificate.crt

proxy:
    autoInjectEnabled: false
    containerName: linkerd-proxy
```

What do you think @cohix create a separate issue for that?

chart/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+ControllerComponentLabel: linkerd.io/control-plane-component
+ControllerImage: gcr.io/linkerd-io/controller:edge-19.2.5
+ControllerLogLevel: info
+ControllerReplicas: 1
+ControllerUID: 2103
+CreatedByAnnotation: linkerd.io/created-by
+DestinationAPIPort: 8086
+EnableH2Upgrade: true
+EnableHA: false
```
grampelberg: It doesn't look like this is working as intended (replicas are staying at 1). Is that managed differently in the CLI?

chart/values.yaml
grampelberg: Is it possible to have a CLI command that dumps these? I'm thinking something similar to `linkerd doc`? Again, not something required for this PR to land, but maybe a follow up issue?

chart/values.yaml
```diff
@@ -0,0 +1,49 @@
+CliVersion: linkerd/cli edge-19.2.5
+ControllerComponentLabel: linkerd.io/control-plane-component
+ControllerImage: gcr.io/linkerd-io/controller:edge-19.2.5
+ControllerLogLevel: info
+ControllerReplicas: 1
+ControllerUID: 2103
+CreatedByAnnotation: linkerd.io/created-by
```
grampelberg: @alpeb @ihcsim what happens when an end user changes this? Would it break their install? I imagine there's some other possibilities in here that could break things ...

chart/values.yaml
ihcsim: Something like `helm inspect values`?

chart/values.yaml
ihcsim: Looks like this flag has no effects on the replicas count in the [controller chart](https://github.com/linkerd/linkerd2/blob/d850b02b667dfdbbf08f4dcccf2046c5165d8209/chart/templates/controller.yaml#L102).

alpeb: Right, the CLI sets the controller's replicas programatically. This needs to be modified to be done in the chart directly instead. Something like: 
```
kind: Deployment
metadata:
  name: linkerd-controller
  namespace: {{.Values.Namespace}}
  labels:
    {{.Values.ControllerComponentLabel}}: controller
  annotations:
    {{.Values.CreatedByAnnotation}}: {{.Values.CliVersion}}
spec:
  replicas: {{- if .Values.EnableHA }}{{.Values.ControllerHAReplicas}}{{- else }}{{.Values.ControllerReplicas}}{{- end }}
```
where `ControllerHAReplicas` would be a new Value.

chart/values.yaml
grampelberg: @ihcsim other way around I think. I'd like to have CI auto-update `chart/values.yaml` based off the config struct in the code itself.

chart/values.yaml
alpeb: Yeah, it seems it's not necessary (nor safe) to expose all these vars. Possible alternatives that come to mind:
- Put all those that are "internal" under a separate section, trusting the user won't touch them
- Move our current charts out of the `chart` directory, and populate the `chart` directory during build time with charts with their internal values replaced.

Maybe there's some Helm trickery I'm not aware of, that can help here as well...

grampelberg: For now, I think it is okay. We should just add comments to this file and group the ones that are internal. The end goal would be to have a build time creation with the correct values.

grampelberg:
@cohix this is starting to get hairy, there's so many changes going on with the `chart` directory that I'm afraid merging is going to be hellacious. Instead, what do you think about creating a `helm` directory for all these changes that is everything you've updated here? That way,  *most* of my change requests aren't an issue and we can get this in and ready to use for folks!

cohix:
@grampelberg sounds good to me. I've been out of commission for a few days due to Wisdom teeth surgery, but I'll do that tomorrow and get the rest of the comments addressed.

grampelberg:
@cohix giving me painful flashbacks to having my own wisdom teeth removed. Take your time!

I think the only comment that is valuable without any code changes on the CLI side would be moving the proxy/init containers to their own templates so there isn't any copy/pasting for those sections.

chart/templates/controller.yaml
cohix: Alright I added the proxy and init containers as sub-templates and then converted all the templates to `include` them instead of having them inline.

chart/values.yaml
cohix: I added the `_config.json` file that includes the `config.global` and `config.proxy` blocks, which are now `include`d in the ConfigMap template.

chart/values.yaml
cohix: I still need to figure out the best way to generate this.

chart/values.yaml
cohix: I re-organised the values.yaml to group them like you mentioned.

chart/values.yaml
cohix: I added this.

chart/values.yaml
cohix: I created an `Internal` section of `values.yaml` and added a comment. Which other values do we think belong in there?

helm/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+  InjectDisabled: disabled
+  LogLevel: warn,linkerd2_proxy=info
+  MetricsPort: 4191
+  SpecFileName: proxy.yaml
+  UID: 2102
+
+# These values are internal to the operation of linkerd,
+# and should not be changed unless you are entirely sure of what you're doing.
+Internal:
+  Internal.CreatedByAnnotation: linkerd.io/created-by
```
grampelberg: Should be:

```yaml
Internal:
  CreatedByAnnotation: xxx
```

grampelberg:
@cohix I found a bunch of other stuff not working, pushed a commit that fixes that:

https://github.com/linkerd/linkerd2/tree/grampelberg/chart-fixes

Feel free to cherry-pick on over to your branch.

PR #2428, opened 2019-03-02
Require cluster-wide k8s API access

siggy:
linkerd/linkerd2#2349 removed the `--single-namespace` flag, in favor of
runtime detection of cluster vs. namespace access, and also
ServiceProfile availability. This maintained control-plane support for
running in these two states.

This change requires control-plane components have cluster-wide
Kubernetes API access and ServiceProfile availability, and will error
out if not. Once #2349 merges, stage 1 install will be a requirement for
a successful stage 2 install.

Part of #2337

### DIFF
--- a/controller/k8s/api.go
+++ b/controller/k8s/api.go
@@ -2,6 +2,7 @@ package k8s
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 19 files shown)


grampelberg:
Should this be called out as a breaking change in the release notes?

siggy:
@grampelberg Yes it should be called out as a breaking change. This is part of a larger "remove single namespace support" task, so we may want to just frame it in that context.

controller/k8s/api.go
```diff
@@ -499,16 +471,11 @@ func GetNamespaceOf(obj runtime.Object) (string, error) {
 }
 
 // getNamespaces returns the namespace matching the specified name. If no name
-// is given, it returns all namespaces, unless the API was configured to only
-// work with a single namespace, in which case it returns that namespace. Note
-// that namespace reads are not cached.
+// is given, it returns all namespaces. Note that namespace reads are not
+// cached.
```
klingerf: Doesn't have to happen as part of this PR, but I originally removed caching for namespace reads when I implemented single-namespace, so in theory we could re-add caching now. Check out https://github.com/linkerd/linkerd2/commit/46c887ca0#diff-45847bbca5cd0ae2b05da10f9f5ab505 to see how it was previously setup.

Issue #2377, opened 2019-02-25
Uninject does not remove linkerd.io labels


grampelberg:
### What is the issue?

Running `uninject`, I would expect all the labels that linkerd adds to be removed.

### How can it be reproduced?

```bash
kubectl get --all-namespaces daemonset,deploy,job,statefulset -l "linkerd.io/control-plane-ns" -o yaml | linkerd uninject - | kubectl apply -f -
```

Labels are still there.

### Environment

- Linkerd version: stable-2.2.0

alpeb:
We only inject annotations and labels at the pod spec template level, so that's what `uninject` is currently removing. But it appears that when the deployment we injected gets applied in Kubernetes, the labels (not the annotations) get copied from the pod spec template into the Deployment metadata. I'll see if I can have `uninject` get rid of those as well.

grampelberg:
That's useful to be aware of in relation to #2376.

klingerf:
@alpeb Good find! This also relates to #2342.

alpeb:
While we're at it, we're gonna fix this while also removing _all_ the labels and annotations that start with "linkerd.io/", not just the list of hard-coded ones we currently have.

PR #2432, opened 2019-03-02
Injection labels fixes

alpeb:
Follow-up to #2334 

Have `GetPatch()` return `*Patch` instead of its marshalled version, so that the consuming side can contribute to the patch; the CLI and the webhook now are adding their own `Created-by` labels, and the webhook will set the Deployment-level labels that Kuberenetes automatically adds in the CLI scenario (ref #2377).


alpeb:
In my mind it's useful to know whether CLI or webhook injected something for debugging purposes... Any other opinions?

ihcsim:
> In my mind it's useful to know whether CLI or webhook injected something for debugging purposes

Yup, totally fine. What you have here is good.

klingerf:
Yep, I like having different `created-by` annotations based on whether the pod was injected by the CLI or auto-injector. I glance at bc1dca8d, and it looks good to me, pending other reviewers' comments. If it's too much of a pain to rebase this branch, I recommend just pushing a separate commit to #2334 with this change, I can review it there.

alpeb:
I'm moving this patch into #2334 , after having applied your suggestions.

PR #2429, opened 2019-03-02
Remove `--api-port` flag from the cli

paranoidaditya:
Closes #2351

Signed-off-by: Aditya Sharma <hello@adi.run>

PR #2434, opened 2019-03-02
wip: Fetch coniguration from the control plane during inject

olix0r:
This change alters inject to fetch the control plane's configuration. Further testing is required.

While doing this, I also tore out the destination-api-port configuration, which I don't believe should exist; though we may want to back this out into a separate change.


alpeb:
:cool:
Just fyi, we already have an external contribution in progress for the api-port config removal at #2429 
Also you might wanna branch off of `alpeb/injection-consolidation` instead of `alpeb/injection-consolidation-conditions-checker` which is gonna get merged into the former soon.

pkg/config/config.go
```diff
@@ -30,17 +37,39 @@ func Proxy() (*pb.Proxy, error) {
...
-		log.Errorf("error unmarshaling %s: %s", filepath, err)
-		return err
+		return fmt.Errorf("Error unmarshaling config: %s", err)
 	}
 
 	return nil
 }
+
+// Fetch the configuration from the kubernetes API.
+func Fetch(k corev1.ConfigMapInterface) (global *pb.Global, proxy *pb.Proxy, err error) {
```
ihcsim: Curious as to which consumers need this? We are hoping that all the consumer pods can just mount the config map and rely on the `Global()` and `Proxy()` functions above to get access to the configs.

olix0r: The CLI needs access to the cluster's configuration. Doing so during inject unblocks [identity work](https://github.com/linkerd/linkerd2/pull/new/ver%2Fidentity-inject); but even after inject is moved into a controller, install will need to be able to inspect the cluster's configuration. Any tooling to manage identity configuration will also need to be able to fetch the global config.

ihcsim: :ok_hand: Just wanna confirm whether this is a stop-gap solution, or something that we will want to keep around. Sounds like the latter.

olix0r:
This has been superseded by a #2455. Inject will take advantage of this in an upcoming PR.

PR #2437, opened 2019-03-03
Fix Tab start button disabled state #2347

### LINKED ISSUE #2347: Tap start button remains disabled even if new resource is selected
### What is the issue?

If you run a Top query on a resource that does not return data, the buttons will be disabled even if you choose a new resource. An Unexpected close error also occurs in the terminal.

### How can it be reproduced?

From Web UI, click Top. 
Enter a namespace (default) and a resource (statefulsets) that will not load data. Click 'Start'. The buttons will be disabled (expected) but will stay disabled if you choose a new resource (unexpected). You have to reload the page to make a new query.

### Logs, error output, etc

In the terminal:
ERRO[0025] Unexpected close error: websocket: close 1011 (internal server error): no pods found for statefulset/

#### `linkerd check` output

```text
Status check results are √
```

### Environment

- Kubernetes Version: v1.10.11
- Linkerd version: Client version: edge-19.2.4
Server version: edge-19.2.4

### Context

May be related to #1679

liquidslr:
Fixes #2347 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 target
 tmp.discovery
 **/.idea
+**/.vscode
 **/cmake-*
 **/CMakeLists.txt
 *.iml
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 target
 tmp.discovery
 **/.idea
-**/.vscode
 **/cmake-*
 **/CMakeLists.txt
 *.iml
--- a/web/app/js/components/Top.jsx
+++ b/web/app/js/components/Top.jsx
@@ -133,7 +133,7 @@ class Top extends React.Component {
   handleTapStop = () => {
     this.setState({
       tapRequestInProgress: false,
-      tapIsClosing: true
+      tapIsClosing: false
     });
   }
 
--- a/web/app/js/components/Top.jsx
+++ b/web/app/js/components/Top.jsx
@@ -133,7 +133,7 @@ class Top extends React.Component {
   handleTapStop = () => {
     this.setState({
       tapRequestInProgress: false,
-      tapIsClosing: false
+      tapIsClosing: true,
     });
   }
 
@@ -144,13 +144,15 @@ class Top extends React.Component {
     });
   }
 
-  updateTapClosingState(isTapClosing) {
+  updateTapClosingState(isTapClosing, isTapRequestInProgress) {
     this.setState({
-      tapIsClosing: isTapClosing
+      tapIsClosing: isTapClosing,
+      tapRequestInProgress: isTapRequestInProgress
     });
   }
 
   render() {
+
     return (
       <div>
         {!this.state.error ? null :
--- a/web/app/js/components/Top.jsx
+++ b/web/app/js/components/Top.jsx
@@ -133,7 +133,7 @@ class Top extends React.Component {
   handleTapStop = () => {
     this.setState({
       tapRequestInProgress: false,
-      tapIsClosing: true,
+      tapIsClosing: true
     });
   }
 
@@ -144,15 +144,14 @@ class Top extends React.Component {
     });
   }
 
-  updateTapClosingState(isTapClosing, isTapRequestInProgress) {
+  updateTapClosingState() {
     this.setState({
-      tapIsClosing: isTapClosing,
-      tapRequestInProgress: isTapRequestInProgress
+      tapRequestInProgress: false,
+      tapIsClosing: false
     });
   }
 
   render() {
-
     return (
       <div>
         {!this.state.error ? null :
--- a/web/app/js/components/TopModule.jsx
+++ b/web/app/js/components/TopModule.jsx
@@ -21,6 +21,7 @@ import { withContext } from './util/AppContext.jsx';
 
 const WS_NORMAL_CLOSURE = 1000;
 const WS_ABNORMAL_CLOSURE = 1006;
+
 class TopModule extends React.Component {
   static propTypes = {
     maxRowsToDisplay: PropTypes.number,
--- a/web/app/js/components/TopModule.jsx
+++ b/web/app/js/components/TopModule.jsx
@@ -103,7 +103,7 @@ class TopModule extends React.Component {
   }
 
   onWebsocketClose = e => {
-    this.updateTapClosingState(false);
+    this.updateTapClosingState(false,false);
     /* We ignore any abnormal closure since it doesn't matter as long as
     the connection to the websocket is closed. This is also a workaround
     where Chrome browsers incorrectly displays a 1006 close code
--- a/web/app/js/components/TopModule.jsx
+++ b/web/app/js/components/TopModule.jsx
@@ -19,6 +19,8 @@ import _throttle from 'lodash/throttle';
 import _values from 'lodash/values';
 import { withContext } from './util/AppContext.jsx';
 
+const WS_NORMAL_CLOSURE = 1000;
+const WS_ABNORMAL_CLOSURE = 1006;
 class TopModule extends React.Component {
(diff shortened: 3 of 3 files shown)


liquidslr:
@scottcarol  please have a look, Also I added.vscode in gitignore

.gitignore
```diff
@@ -2,6 +2,7 @@
 target
 tmp.discovery
 **/.idea
+**/.vscode
```
scottcarol: Hi @liquidslr, We typically try to avoid making changes to files that don't specifically pertain to the PR. In this case I would make a [global .gitignore file ](https://help.github.com/en/articles/ignoring-files) on your machine so that this file is unchanged.

liquidslr: Sure @scottcarol I will revert the change

web/app/js/components/Top.jsx
```diff
@@ -133,7 +133,7 @@ class Top extends React.Component {
   handleTapStop = () => {
     this.setState({
       tapRequestInProgress: false,
-      tapIsClosing: true
+      tapIsClosing: false
```
scottcarol: Thanks for jumping in to this issue. See @klingerf's [comment in issue #2347](https://github.com/linkerd/linkerd2/issues/2347#issuecomment-469353306) - the issue is not solely a UI issue - it may be related to websockets. 

I would suggest starting by investigating the `Unexpected close error` in the terminal - why we are seeing this error, what is being sent and what is being received, and how it compares to the websockets spec.

liquidslr: Yeah I read it a few minutes ago, I would take a look at the issue again

grampelberg: From the product perspective, I'd like to have the client do a contextual banner in the top/tap fields that explains why the query a user tried to do didn't work. This would be different than the popup in the lower left hand corner, which feels like it is for unknown errors.

I'm sure this could either be implemented by having structured errors that can be switched between (known vs unknown) or as a new message type. @rmars @scottcarol what do you think is the best approach?

rmars: @grampelberg yeah, that would be useful. I think a new state field, like `displayError` or something could be used to track errors we can display in a helpful manner to the user, and leave `this.state.error` for unknown errors (which it currently is).

We could have structured errors too, though I think that error handling is ever so slightly different in each file, so it'd be a bit more annoying (or require more refactoring).

For this particular case, would this be for query errors? (e.g. having a malformed query) or would it be for displaying the websocket errors? I've been wanting to add more form validation for a while, this could also go into that sort of banner.

grampelberg: Hmmm, are we getting too complicated? The scope of this issue feels like it should just make sure you can start a tap again if any error occurs. A follow up ticket would be to make the error a little more contextual.

rmars: Ah yeah, sure! I don't think that needs to be addressed as part of this issue.
Filed https://github.com/linkerd/linkerd2/issues/2461 to track.

scottcarol: Is an empty return to a query always an error? In the situation I described in #2347, I ran a Top query on a resource that didn't return data. Are there cases where the combination of namespace/resources available in the select menus would not produce a result from the API, and is that something we could check before sending the request?

screen shot 2019-03-06 at 2 41 56 pm

liquidslr: I checked for several cases and noticed that the query which does not returns data are the one's causing problem and every time no data is returned an error is thrown as i mentioned earlier that it has been implemented in such a way. And about knowing whether namespace/resources available in the select menus would not produce a result without sending a request, i am not sure how to implement that.

web/app/js/components/Top.jsx
```diff
@@ -144,13 +144,15 @@ class Top extends React.Component {
     });
   }
 
-  updateTapClosingState(isTapClosing) {
+  updateTapClosingState(isTapClosing, isTapRequestInProgress) {
     this.setState({
-      tapIsClosing: isTapClosing
+      tapIsClosing: isTapClosing,
+      tapRequestInProgress: isTapRequestInProgress
```
scottcarol: This is a minor nit, but for consistency's sake I would switch the order of lines 149 and 150 to match the order of `handleTapStop`, to make it more readable later.

web/app/js/components/Top.jsx
```diff
@@ -144,13 +144,15 @@ class Top extends React.Component {
...
     });
   }
 
   render() {
+
```
scottcarol: You added an extra space here, which you don't need if you are returning right below the `render()`. See [React Component docs code example](https://reactjs.org/docs/react-component.html)

web/app/js/components/TopModule.jsx
```diff
@@ -103,7 +103,7 @@ class TopModule extends React.Component {
   }
 
   onWebsocketClose = e => {
-    this.updateTapClosingState(false);
+    this.updateTapClosingState(false,false);
```
scottcarol: You want to add a space before the second `false` to match our style.

web/app/js/components/TopModule.jsx
```diff
@@ -103,7 +103,7 @@ class TopModule extends React.Component {
   }
 
   onWebsocketClose = e => {
```
rmars: I think we may want to update this method to be

```
  onWebsocketClose = e => {
    this.updateTapClosingState();
    /* We ignore any abnormal closure since it doesn't matter as long as
    the connection to the websocket is closed. This is also a workaround
    where Chrome browsers incorrectly displays a 1006 close code
    https://github.com/linkerd/linkerd2/issues/1630
    */
    if (e.code !== WS_NORMAL_CLOSURE && e.code !== WS_ABNORMAL_CLOSURE) {
      this.setState({
        error: {
          error: `Websocket close error [${e.code}: ${wsCloseCodes[e.code]}] ${e.reason ? ":" : ""} ${e.reason}`
        }
      });
    }
  }
```
We'll need to add these consts near the top of the file:
```
const WS_NORMAL_CLOSURE = 1000;
const WS_ABNORMAL_CLOSURE = 1006;
```

That will allow us to pass any errors on to the default banner handling, which I think is better than nothing until we implement the contextual banner discussed in #2461

web/app/js/components/Top.jsx
```diff
@@ -144,13 +144,15 @@ class Top extends React.Component {
     });
   }
 
-  updateTapClosingState(isTapClosing) {
+  updateTapClosingState(isTapClosing, isTapRequestInProgress) {
```
rmars: We only call this from one place, and it's to set both of these to false. We should consider removing the args in this case and just having
```
  updateTapClosingState() {
    this.setState({
      tapIsClosing: false,
      tapRequestInProgress: false
    });
  }
```

liquidslr:
@rmars @scottcarol  incorporated the suggestions made,  please have a look now :-)

web/app/js/components/TopModule.jsx
```diff
@@ -19,6 +19,8 @@ import _throttle from 'lodash/throttle';
 import _values from 'lodash/values';
 import { withContext } from './util/AppContext.jsx';
 
+const WS_NORMAL_CLOSURE = 1000;
+const WS_ABNORMAL_CLOSURE = 1006;
```
scottcarol: My only styling nit - I'd add a line underneath line 23, since we typically add a new line under blocks of `const` declarations for readability (see `ErrorModal.jsx` for an example).

liquidslr: Sure  :+1: done !

scottcarol:
Merged to master, thanks again @liquidslr, great job! 🚂

Issue #2134, opened 2019-01-23
Sidebar should poll for new meshed resources


rmars:
When new resources are added to the mesh, they don't show up in the sidebar unless the whole page is refreshed. We should add polling to the sidebar. 

We should also think about how the polling will affect the appearance of SR indicators in the sidebar and main change (the out of sync issue mentioned in #1542). I would rather the page at least refresh periodically, though, because the navigation aspect is useful.

Relevant past tickets: #1669 #1542

scottcarol:
Closed by #2440

PR #2440, opened 2019-03-04
Refactor code to make sidebar in sync with the main view (#2134)

liquidslr:
This PR fixes issue #2134, also mentioned by # 1542. The main reason for out of sync behavior was the component was not re-rendering although it received new data as props which is now fixed. Now any change in resources will be reflected in the sidebar in sync with the main view.

cc @rmars


web/app/js/components/NavigationResources.jsx
```diff
@@ -19,18 +19,35 @@ const styles = () => ({
...
     super(props);
-    this.state = {open: false};
+    this.state = {
+      open: false,
+      allMetrics: {},
+      nsMetrics: {}
+    };
+  }
+
+  componentWillReceiveProps(prevProps, nextProps) {
```
rmars: The changes you made in NavigationResource.jsx were enough to fix the bug, and moving the setting of this state into a `componentWillReceiveProps` isn't necessary.  Reverting this to the way it was, I still see new resources being added and removed in the sidebar.

(Also note that `componentWillReceiveProps` will be deprecated in the future, so we should try to avoid introducing them. https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops)

liquidslr: Yeah, I shouldn't have done that because data is already coming as props, so every time props change component will re-render.

liquidslr:
@rmars thanks for the detailed code review, please have a look now!

Issue #2355, opened 2019-02-22
CNI configuration is not cleaned up on deletion


grampelberg:
### What is the issue?

Removing the CNI DaemonSet from my cluster does not clean the plugin or configuration up.

### How can it be reproduced?

```bash
linkerd install-cni --dest-cni-bin-dir /home/kubernetes/bin | kubectl delete -f -
```

### Logs, error output, etc

(If the output is long, please create a [gist](https://gist.github.com/) and
paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version: 1.12.5-gke.5
- Cluster Environment: GKE
- Host OS: COS
- Linkerd version: stable-2.2.0

siggy:
FWIW I've observed this working sometimes but not all the time. In the failed cases the pod just exits before `cleanup()` executes.

Here's output from a success cleanup:
```bash
$ kubectl -n linkerd logs -f pod/linkerd-cni-8qgc7
Wrote linkerd CNI binaries to /host/opt/cni/bin
Using CNI config template from CNI_NETWORK_CONFIG environment variable.
      "k8s_api_root": "https://__KUBERNETES_SERVICE_HOST__:__KUBERNETES_SERVICE_PORT__",
      "k8s_api_root": "https://10.96.0.1:__KUBERNETES_SERVICE_PORT__",
CNI config: {
  "name": "linkerd-cni",
  "type": "linkerd-cni",
  "log_level": "info",
  "policy": {
      "type": "k8s",
      "k8s_api_root": "https://10.96.0.1:443",
      "k8s_auth_token": "__SERVICEACCOUNT_TOKEN__"
  },
  "kubernetes": {
      "kubeconfig": "/etc/cni/net.d/ZZZ-linkerd-cni-kubeconfig"
  },
  "linkerd": {
    "incoming-proxy-port": 4143,
    "outgoing-proxy-port": 4140,
    "proxy-uid": 2102,
    "ports-to-redirect": [],
    "inbound-ports-to-ignore": [4190,4191],
    "outbound-ports-to-ignore": [],
    "simulate": false
  }
}
Created CNI config /host/etc/cni/net.d/10-default.conflist
Done configuring CNI. Sleep=true


SIGTERM received, simply exiting...
Removing linkerd-cni artifacts.
Removing linkerd-cni config: /host/etc/cni/net.d/10-default.conflist
[...]

PR #2441, opened 2019-03-04
Update how the CNI script cleans itself up

codeman9:
Due to spacing updates, the following:

```
        lifecycle:
          preStop:
            exec:
              command: ["kill","-15","1"]
```

is hidden in the change list for cni-template.go

Since this is an intermittent problem, it would be great to get some more extensive testing on this to see if it resolves: #2355


grampelberg:
This works for me! I went back and tested on the stable version and it fails 100% of the time on GKE for me. So, problem solved, far as I can tell =)

PR #2442, opened 2019-03-04
Uninject: remove labels at the top level

### LINKED ISSUE #2377: Uninject does not remove linkerd.io labels

alpeb:
Fixes #2377

In inject's `ResourceConfig`, renamed `objMeta` to `templateMeta` since
it really points to the pod template metadata. And created a new field
`objMeta` that really points to the main workload (e.g. Deployment) metadata.

Refactored uninject to clean up the labels at both `objMeta` and
`templateMeta`. Also it will remove all the labels and annotations that
start with "linkerd.io" except for the "linkerd.io/inject" annotation.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -55,8 +55,8 @@ type ResourceConfig struct {
 	nsAnnotations         map[string]string
 	meta                  metav1.TypeMeta
 	obj                   runtime.Object
-	objMeta               *metav1.ObjectMeta
-	templateMeta          objMeta
+	workLoadMeta          *metav1.ObjectMeta
+	podMeta               objMeta
 	podLabels             map[string]string
 	podSpec               *v1.PodSpec
 	dnsNameOverride       string
@@ -114,10 +114,10 @@ func (conf *ResourceConfig) ParseMeta(bytes []byte) (bool, error) {
 	if err := yaml.Unmarshal(bytes, &conf.meta); err != nil {
 		return false, err
 	}
-	if err := yaml.Unmarshal(bytes, &conf.templateMeta); err != nil {
+	if err := yaml.Unmarshal(bytes, &conf.podMeta); err != nil {
 		return false, err
 	}
-	return conf.templateMeta.ObjectMeta != nil, nil
+	return conf.podMeta.ObjectMeta != nil, nil
 }
 
 // GetPatch returns the JSON patch containing the proxy and init containers specs, if any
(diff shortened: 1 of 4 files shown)


grampelberg:
```bash
curl https://run.linkerd.io/emojivoto.yml | bin/go-run cli inject --tls=optional - | bin/go-run cli uninject -
```

Worked for me!

pkg/inject/uninject.go
```diff
@@ -49,28 +57,18 @@ func (conf *ResourceConfig) uninjectPodSpec(report *Report) {
...
 	for key, val := range t.Labels {
-		keep := true
-		for _, label := range k8s.InjectedLabels {
-			if key == label {
-				keep = false
-				break
-			}
-		}
-		if keep {
+		if !strings.HasPrefix(key, k8s.Prefix) {
```
siggy: 👍

pkg/inject/inject.go
```diff
@@ -55,7 +55,8 @@ type ResourceConfig struct {
 	nsAnnotations         map[string]string
 	meta                  metav1.TypeMeta
 	obj                   runtime.Object
-	objMeta               objMeta
+	objMeta               *metav1.ObjectMeta
+	templateMeta          objMeta
```
siggy: i'm struggling to parse a field named `objMeta` and another field of type `objMeta`. mind naming them a bit more explicitly? maybe something like:
```go
workLoadMeta  *metav1.ObjectMeta
podMeta       objMeta
```

alpeb: You're right, that's confusing. I'll apply your suggestion :+1:

alpeb:
Thank you all for the feedback :+1: 
@klingerf That's a good catch. And an odd Kubernetes behaviour. Looks like we'll need some kind of `uninstall` subcommand for this, which could also take care of removing the CRD, roles, etc.

grampelberg:
@alpeb I was chatting with @olix0r about that earlier this morning. A `linkerd remove | kubectl delete -f -` would be a great alternative to something like `kubectl delete --all-namespaces -l ...`.

Totally in agreement with @klingerf that it should be a follow up =)

PR #2451, opened 2019-03-06
Re-add pre-install permission checks

klingerf:
This fixes a small issue from #2421, where we inadvertently removed some of the pre-flight RBAC checks.

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -90,6 +90,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 		} else {
 			checks = append(checks, healthcheck.LinkerdPreInstallClusterChecks)
 		}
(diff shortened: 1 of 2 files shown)


cli/cmd/check.go
```diff
@@ -90,6 +90,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 		} else {
 			checks = append(checks, healthcheck.LinkerdPreInstallClusterChecks)
 		}
+		checks = append(checks, healthcheck.LinkerdPreInstallChecks)
```
siggy: nit: there's no functional difference, but i find it helpful to order these the same way they are defined as constants (and output via `linkerd check`). mind moving `LinkerdPreInstallChecks` below `LinkerdPreInstallCapabilityChecks`?

klingerf: Ah, yep, fixed.

PR #2455, opened 2019-03-06
Public API endpoint `Config()`

alpeb:
Retrieves Global and Proxy configurations.
This can be used in #2434 as an alternative to calling the Kubernetes
API directly.

TODO: A unit test. The endpoint relies on the mounted volume, so I don't know
yet how to mock that.


controller/api/public/testdata/proxy.conf.json
```diff
@@ -0,0 +1,40 @@
+{
+    "proxyImage": {
+      "imageName": "gcr.io/linkerd-io/proxy",
+      "pullPolicy": "IfNotPresent"
+    },
+    "proxyInitImage": {
+      "imageName": "gcr.io/linkerd-io/proxy-init",
+      "pullPolicy": "IfNotPresent"
+    },
+    "destinationApiPort": {
```
klingerf: Just a heads up that once you rebase this branch and pick up #2429, you'll need to remove this field from this fixture file.

proto/public.proto
```diff
@@ -6,6 +6,8 @@ import "google/protobuf/duration.proto";
 
 import "common/healthcheck.proto";
 
+import "config/config.proto";
```
klingerf: 👍

controller/api/public/grpc_server.go
```diff
@@ -63,19 +67,27 @@ func newGrpcServer(
...
+		mountPathGlobalConfig: pkgK8s.MountPathGlobalConfig,
+		mountPathProxyConfig:  pkgK8s.MountPathProxyConfig,
 	}
 
 	pb.RegisterApiServer(prometheus.NewGrpcServer(), grpcServer)
 
 	return grpcServer
 }
 
+func (s *grpcServer) WithConfigPaths(global, proxy string) *grpcServer {
```
klingerf: Since this is only used in tests, my preference would be to remove this method so that people can't misuse it, and instead just set these fields directly in the test. Something like:

```diff
diff --git a/controller/api/public/grpc_server.go b/controller/api/public/grpc_server.go
index 43ff9edb..d8b0a121 100644
--- a/controller/api/public/grpc_server.go
+++ b/controller/api/public/grpc_server.go
@@ -82,12 +82,6 @@ func newGrpcServer(
 	return grpcServer
 }
 
-func (s *grpcServer) WithConfigPaths(global, proxy string) *grpcServer {
-	s.mountPathGlobalConfig = global
-	s.mountPathProxyConfig = proxy
-	return s
-}
-
 func (*grpcServer) Version(ctx context.Context, req *pb.Empty) (*pb.VersionInfo, error) {
 	return &pb.VersionInfo{GoVersion: runtime.Version(), ReleaseVersion: version.Version, BuildDate: "1970-01-01T00:00:00Z"}, nil
 }
diff --git a/controller/api/public/grpc_server_test.go b/controller/api/public/grpc_server_test.go
index 95124a9f..4ebf424b 100644
--- a/controller/api/public/grpc_server_test.go
+++ b/controller/api/public/grpc_server_test.go
@@ -586,7 +586,9 @@ func TestConfig(t *testing.T) {
 			k8sAPI,
 			"linkerd",
 			[]string{},
-		).WithConfigPaths("testdata/global.conf.json", "testdata/proxy.conf.json")
+		)
+		fakeGrpcServer.mountPathGlobalConfig = "testdata/global.conf.json"
+		fakeGrpcServer.mountPathProxyConfig = "testdata/proxy.conf.json"
 
 		k8sAPI.Sync()
 
```

controller/api/public/grpc_server.go
```diff
@@ -220,6 +232,19 @@ func (s *grpcServer) SelfCheck(ctx context.Context, in *healthcheckPb.SelfCheckR
...
+	global, err := config.Global(s.mountPathGlobalConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving global config - %s", err)
+	}
+	proxy, err := config.Proxy(s.mountPathProxyConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving proxy config - %s", err)
+	}
+	rsp := &configPb.All{Global: global, Proxy: proxy}
+	return rsp, nil
```
klingerf: TIOLI:

```
	return &configPb.All{Global: global, Proxy: proxy}, nil
```

(no need for the `rsp` var)

chart/templates/controller.yaml
```diff
@@ -115,6 +115,9 @@ spec:
           containerPort: 8085
         - name: admin-http
           containerPort: 9995
+        volumeMounts:
+          - name: config
```
siggy: unindent?

controller/proxy-injector/webhook.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/linkerd/linkerd2/pkg/config"
 	"github.com/linkerd/linkerd2/pkg/inject"
+	pkgK8s "github.com/linkerd/linkerd2/pkg/k8s"
```
siggy: nit: no need to rename this as we're not importing any other `k8s` packages.

controller/api/public/grpc_server.go
alpeb: Makes sense :+1:

ihcsim:
LGTM too :+1:

pkg/config/config.go
```diff
@@ -7,23 +7,22 @@ import (
...
+func Global(filepath string) (*pb.Global, error) {
 	config := &pb.Global{}
-	err := unmarshalConfig(k8s.MountPathGlobalConfig, config)
+	err := unmarshalConfig(filepath, config)
 	return config, err
 }
 
 // Proxy returns the Proxy protobuf config from the linkerd-config ConfigMap
-func Proxy() (*pb.Proxy, error) {
+func Proxy(filepath string) (*pb.Proxy, error) {
```
olix0r: tioli: if this only will ever be called with one path, perhaps go back to using consts, but consts defined _in this package_? Or is there a value to having these paths defined as variables in every service that consumes them?

alpeb: @olix0r I parametrized this so that the unit test can call pointing to a test fixture file.

PR #2466, opened 2019-03-07
cli: Always display stat tables for all routes

### LINKED ISSUE #2323: "linkerd routes" fails due to non-.default.svc.cluster.local Host header
Most of our traffic comes via an AWS ELB (created using `Service` of `type: LoadBalancer` with annotations - see below). Therefore the `Host` header in requests is not generally in the `foo.default.svc.cluster.local` format. 

This means that even _after_ adding a `ServiceProfile` and running `linkerd routes` we see this error...
```
linkerd routes deploy/foo
Error: No Service Profiles found for selected resources 
```

Looking at the code, it seems that we're silently ignoring errors from `topRoutesFor` (see https://github.com/linkerd/linkerd2/blob/master/controller/api/public/top_routes.go#L72), which would give more context to the problem, and then falling back to this outputted error due to `len(tables) == 0`.

`linkerd routes deploy/foo` began working once we logged into an adjacent pod and made a `curl` request to the target service passing the host header as `-H 'Host: foo.default.svc.cluster.local'`.

Here's an example of our Service definition...

```
---
apiVersion: v1
kind: Service
metadata:
  name: foo
  labels:
    app: foo
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: ${AWS_CERT_ID}
    service.beta.kubernetes.io/aws-load-balancer-ssl-ports: https
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: http
    external-dns.alpha.kubernetes.io/hostname: ${APP_EXTERNAL_HOSTNAME}
spec:
  type: LoadBalancer
  ports:
  - port: 443
    protocol: TCP
    name: https
    targetPort: app-port
  selector:
    app: foo
```

### LINKED ISSUE #2328: Routes error message is misleading
### What is the issue?

When there is no traffic to a service, running `linkerd routes` produces:

```bash
linkerd routes deploy/foo
Error: No Service Profiles found for selected resources
```

The error message should be `Error: No traffic observed over the last X minutes, see ... for more details`

kleimkuhler:
## Problem

When an object has no previous route metrics, we do not generate a table for
that object.

The reasoning behind this was for reducing output of the following command:

```
$ linkerd routes deploy --to deploy/foo
```

For each deployment object, if it has no previous traffic to `deploy/foo`, then
a table would not be generated for it.

However, the behavior we see with that indicates there is an error even when a
Service Profile is installed:

```
$ linkerd routes deploy deploy/foo
Error: No Service Profiles found for selected resources
```

## Solution

Always generate a stat table for the queried resource object.

## Validation

I deployed [booksapp](https://github.com/buoyantIO/booksapp) with the `traffic`
deployment removed and Service Profiles installed.

Without the fix, `linkerd routes deploy/webapp` displays an error because there
has been no traffic to `deploy/webapp` without the `traffic` deployment.

With the fix, the following output is generated:

```
ROUTE                       SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
GET /                        webapp     0.00%   0.0rps           0ms           0ms           0ms
GET /authors/{id}            webapp     0.00%   0.0rps           0ms           0ms           0ms
GET /books/{id}              webapp     0.00%   0.0rps           0ms           0ms           0ms
POST /authors                webapp     0.00%   0.0rps           0ms           0ms           0ms
POST /authors/{id}/delete    webapp     0.00%   0.0rps           0ms           0ms           0ms
POST /authors/{id}/edit      webapp     0.00%   0.0rps           0ms           0ms           0ms
POST /books                  webapp     0.00%   0.0rps           0ms           0ms           0ms
POST /books/{id}/delete      webapp     0.00%   0.0rps           0ms           0ms           0ms
POST /books/{id}/edit        webapp     0.00%   0.0rps           0ms           0ms           0ms
[DEFAULT]                    webapp     0.00%   0.0rps           0ms           0ms           0ms
```

Closes #2328

### DIFF
--- a/controller/api/public/top_routes.go
+++ b/controller/api/public/top_routes.go
@@ -154,10 +154,6 @@ func (s *grpcServer) topRoutesFor(ctx context.Context, req *pb.TopRoutesRequest,
 		}
 	}
 
-	if len(profiles) == 0 {
-		return nil, fmt.Errorf("No ServiceProfiles found for Services of %s/%s", requestedResource.Type, requestedResource.Name)
-	}
-
 	metrics, err := s.getRouteMetrics(ctx, req, profiles, targetResource)
 	if err != nil {
 		return nil, err
@@ -274,10 +270,8 @@ func (s *grpcServer) getRouteMetrics(ctx context.Context, req *pb.TopRoutesReque
 		}
 	}
 
-	err = processRouteMetrics(results, timeWindow, table)
-	if err != nil {
-		return nil, err
-	}
+	processRouteMetrics(results, timeWindow, table)
+
 	return table, nil
 }
 
@@ -311,12 +305,9 @@ func renderLabels(labels model.LabelSet, services []string) string {
 	return fmt.Sprintf("{%s}", strings.Join(pairs, ", "))
 }
 
-func processRouteMetrics(results []promResult, timeWindow string, table indexedTable) error {
-	samples := 0
+func processRouteMetrics(results []promResult, timeWindow string, table indexedTable) {
 	for _, result := range results {
 		for _, sample := range result.vec {
-			samples++
-
 			route := string(sample.Metric[model.LabelName("rt_route")])
 			dst := string(sample.Metric[model.LabelName("dst")])
 			dst = strings.Split(dst, ":")[0] // Truncate port, if there is one.
@@ -355,6 +346,4 @@ func processRouteMetrics(results []promResult, timeWindow string, table indexedT
 			}
 		}
 	}
-
-	return nil
 }
--- a/controller/api/public/top_routes.go
+++ b/controller/api/public/top_routes.go
@@ -355,8 +355,6 @@ func processRouteMetrics(results []promResult, timeWindow string, table indexedT
 			}
 		}
 	}
-	if samples == 0 {
-		return errors.New("No samples")
-	}
+
 	return nil
 }
--- a/controller/k8s/api.go
+++ b/controller/k8s/api.go
@@ -787,7 +787,7 @@ func (api *API) GetServiceProfileFor(svc *corev1.Service, clientNs string) *spv1
 		}
 	}
 	// Not found; return default.
-	log.Infof("no Service Profile found for '%s' -- using default", dst)
+	log.Debugf("no Service Profile found for '%s' -- using default", dst)
 	return &spv1alpha1.ServiceProfile{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: dst,
--- a/controller/k8s/api.go
+++ b/controller/k8s/api.go
@@ -787,6 +787,7 @@ func (api *API) GetServiceProfileFor(svc *corev1.Service, clientNs string) *spv1
 		}
 	}
 	// Not found; return default.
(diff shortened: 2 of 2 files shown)


controller/api/public/top_routes.go
```diff
@@ -355,8 +355,6 @@ func processRouteMetrics(results []promResult, timeWindow string, table indexedT
 			}
 		}
 	}
-	if samples == 0 {
-		return errors.New("No samples")
-	}
+
```
siggy: now that we're not checking `samples`, you can just remove the variable altogether. and going a step further, this function can no longer return an error, so remove the return type.

grampelberg:
@kleimkuhler how are these two use cases being handled?

- No traffic, no profile
- Traffic, no profile 

With `[default]` there all the time now, it feels like we should warn that there's no service profile and still show `[default]` for both of these cases.

kleimkuhler:
@grampelberg Great question!

Previously, we would return `Error: No Service Profiles found for selected resources` for a resource that had no Service Profiles installed.

Now that we do not consider empty stats an error, we use the default Service Profile returned by [`GetServiceProfileFor`](https://github.com/linkerd/linkerd2/blob/375a00efc606b2135b6ca59c12a862b0b50d08b3/controller/k8s/api.go#L791) and display the following:

```
ROUTE       SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
[DEFAULT]    webapp    80.95%   6.7rps          30ms          97ms         175ms
```

In order for a user to see a more detailed `linkerd routes ...` command, a Service Profile must be installed for that service.

I added an additional log line when a default Service Profile is used so that a user can see there is opportunity to install one of their own.

controller/k8s/api.go
```diff
@@ -787,6 +787,7 @@ func (api *API) GetServiceProfileFor(svc *corev1.Service, clientNs string) *spv1
 		}
 	}
 	// Not found; return default.
+	log.Infof("no Service Profile found for '%s' -- using default", dst)
```
siggy: a quick check of `linkerd logs` shows this contributing to the majority of log lines. consider `log.Debugf`

grampelberg:
@kleimkuhler I must be doing something wrong, I was going to take a peek at the warning and got this:

```bash
$ bin/go-run cli routes -n linkerd svc/linkerd-web
Error: No Service Profiles found for selected resources
```

Nothing special, far as I'm aware.

kleimkuhler:
@grampelberg Let me know if you still get the error after a fresh install of the control plane!

grampelberg:
@kleimkuhler works for me. I like the concept of having `no Service Profile found` be part of info. That said, I'll defer to @siggy as the value doesn't feel substantial.

kleimkuhler:
@grampelberg When clicking around on the dashboard, I see why @siggy has concern about it being logged at `info` instead of `debug`. It fills up a significant portion of the output. However, at `debug` it could be hard to notice if you are not looking for it.

Considering both, I like keeping it at `debug` because a user can pretty easily investigate how to get more information out of the `routes` command, and learn about Service Profiles from that.

grampelberg:
@kleimkuhler yup, works for me!

Issue #2447, opened 2019-03-05
CLI Inject With Proxy Config Overrides Annotations


ihcsim:
Provide a mechanism to override the configurations of a proxy during `linkerd inject`. By default, if no flags are provided, the proxy will inherit the configurations specified in the `linkerd-config` config map during CLI inject. If present, the config flags should be converted into annotations on the pod spec before being passed to the proxy-injector. The proxy-injector will then perform the configs overrides per #2287.

Depending on whether the YAML defines a single pod or a workflow resource, the CLI will have to insert the annotations at different levels (i.e. `metadata.annotations` and `spec.template.metadata.annotations`).

Depends on #2287.

klingerf:
I just wanted to clarify that once we've added the new public API config endpoint (#2455) and updated the CLI to call that endpoint (#2434), the CLI can apply flag overrides directly to the config before calling the inject library. There won't be any need to convert the flags to annotations on the pod spec, and this change won't be blocked on needing the inject library to honor pod spec annotations (#2287).

ihcsim:
Per discussion with @klingerf, the step to convert the CLI inject proxy options to annotations is still necessary, to ensure that subsequent updates (via `kubectl apply` etc.) don't override the existing proxy configs. The current override logic will always look for config annotations, before falling back to defaults defined in the `linkerd-config` config map.

Issue #2342, opened 2019-02-21
autoinject: Inject Pods rather than Deployments


olix0r:
Currently, the autoinjector service watches deployments and adds proxies to deployments. However, deployments are only created/updated when a user invokes `kubectl apply`... As we introduce injector config that might be dynamic (i.e. a user may change it at runtime), this means that existing deployments would have no way of using this configuration as the deployments pods are (re-)created.

What if, instead of getting notifications about deployment updates, we watch for _Pod_ creation? I think this will enable the injector to apply configuration on pod creation (i.e. if i delete a pod of an existing deployment, a new pod will be created and get injected anew with config).

cc @grampelberg @ihcsim

klingerf:
This seems like a good idea, and could be a fix for #1751. It would be worth verifying this assumption:

> if i delete a pod of an existing deployment, a new pod will be created and get injected anew with config

We can setup a webhook that watches pods and logs when the webhook is called, to figure out exactly when that happens in a pod lifecycle. I suspect it might only be called once per apply, and subsequent incarnations of the pod will just use the previously applied and modified yaml, instead of re-calling the webhook.

I also wanted to note that some of our inject code relies on the pod's parent object. For example, when I inject a deployment called "hello", the inject code adds the `linkerd.io/proxy-deployment: hello` label to the pod. This is required for generating prometheus labels. If we're only watching pods, we won't receive the pod's parent object as part of the admission request, and we'll need to come up with a different way to derive the owner from the standalone pod object. We already have code to do this by way of pod owner references, but we should validate that that successfully returns the pod owner for all of our inject use cases:

https://github.com/linkerd/linkerd2/blob/43d29d629e2416c827aff0883812404bf8e99da5/controller/k8s/api.go#L274-L293

olix0r:
>  I suspect it might only be called once per apply, and subsequent incarnations of the pod will just use the previously applied and modified yaml, instead of re-calling the webhook.

I'm pretty confident that this _has to_ work: a pod doesn't actually exist until it's instantiated.  I.e. if i delete one pod in a deployment, a new pod is _created_ (with a new name), so there can't be a pod resource in the kubernetes API until then... But I certainly agree it's important to prototype this and to figure out what other metadata we have access to in the pod spec we receive from the webhook.

olix0r:
[...]

Issue #2339, opened 2019-02-20
New Annotation To Trigger Automatic Proxy Version Upgrade


ihcsim:
### What problem are you trying to solve?
Support a new annotation (`linkerd.io/upgrade:true`) to trigger automatic proxy version upgrade using the proxy injector during workload updates. This can help to streamline the users' proxy version upgrade process, without the needs for CLI interaction. Currently, the proxy injector will skip all workloads that are already injected with the Linkerd proxy.

Continuation of #2260.
Depends on #2289.

### How should the problem be solved?
The logic around the [`healthcheck.HasExistingSidecars()`](https://github.com/linkerd/linkerd2/blob/5dcb9086f54e4cd290b499b8ba7285c01fdcc50c/controller/proxy-injector/webhook.go#L228) needs to be a bit smarter, by checking for the existent of the `linkerd.io/upgrade: true` annotation, and comparing the current vs latest proxy version. The latest proxy version can be retrieved from the proxy config. See #2298.

Once the upgrade is completed, the annotation needs to be removed.

We also need to consider version rollback mechanism if the version upgrade failed, and not leave the workload in a limbo state.

### Any alternatives you've considered?
Stick with the manual upgrade process using the CLI.

### How would users interact with this feature?
Via the `linkerd.io/upgrade: true` annotation.

olix0r:
@ihcsim do you think something like this is still valuable or necessary given the auto-inject and config changes we've made?

I can see a use case for this, but, rather than linkerd drive the actual rolling of pods, i think we'd want to be able to trigger a deployment operator of some sort to cause a new rollout. In that scenario, I'd imagine the operator would watch the linkerd control plane's config object and, when it changes, trigger a (gradual, guarded) redeploy.

ihcsim:
It will be super cool to be able to do what you said through a deployment operator. I can totally see how a smart (configurable?) auto-upgrade will be super useful for large data planes. Gonna keep this issue open for now.

PR #2471, opened 2019-03-07
Support Auto-Inject Config Overrides Via Annotations

### LINKED ISSUE #2287: Proxy Config Overrides Via Annotation During Auto-Inject
Provide a mechanism to override proxy configurations during auto-inject via pod annotations. When the proxy-injector receives an inject request, it determines the inject options in the following order:

* if an annotation for that option is present on the pod, it uses that
* otherwise if an annotation for that option is present on the pod's namespace, it uses that
* if no annotations are present, it uses the default from the linkerd-config config map

ihcsim:
This PR introduces a ~`useOverridesOrDefaults()`~ `getOverrideOrDefault()` internal method to the `pkg/inject` library, to perform proxy configuration overrides. It looks up the runtime value of a particular proxy configuration in the following places, and returns immediately when a match is found:

1. pod template annotations
1. owner namespace annotations
1. the `linkerd-config` config map (containing defaults created during installation)

The set of configurations that can be overridden is determined a predefined list of annotations defined in [pkg/k8s/labels.go](https://github.com/linkerd/linkerd2/blob/0ddeb3e50514e71963609941a6959fc8535c2404/pkg/k8s/labels.go#L231-L248). Configuration overrides support for the CLI inject is tracked under #2447.

Test cases can be found in this [gist](https://gist.github.com/ihcsim/b457b9a799cb4ea56f7ce8441838b511).

### Assumptions

* The `useOverridesOrDefault()` method assumes that the  `config.ResourceConfig.nsAnnotations` and `config.ResourceConfig.podMeta` fields always have the correct namespace annotations and pod template annotations.

### Limitations

* Only works with `Deployment` workload, per #2342.

### WIP

* ~Perform configuration overrides when injected workload is updated (i.e. although the proxy injector won't inject the proxy into the YAML manifest, it still needs to check for config annotations.)~

Fixes #2287. 


grampelberg:
Looking through the test cases, I'm having a bit of an existential crisis. The annotations when you're just overriding a single thing seems okay. Seeing the wall of annotations for overrides gets me worried after experiencing the crazy set of annotations most ingress controllers try to support. So, is there a better way to do this that has some more structure?

As this ends up on the injection side, the only options I can come up with are:

- New CRD that uses selectors to target the overrides. This will be particularly painful to version between releases, introduces another CRD that isn't super generic/useful elsewhere.
- Something with a ConfigMap. Feels as bad or worse than the annotations, unstructured and separate.

I went looking at kustomize again, but it doesn't look like the right solution to this either.

Any other ideas? I'm not sure there are many other possibilities.

pkg/inject/inject_test.go
```diff
@@ -0,0 +1,397 @@
...
+					k8s.ProxyLimitCPUAnnotation:                "1.5",
+					k8s.ProxyLimitMemoryAnnotation:             "256",
+					k8s.ProxyUIDAnnotation:                     "8500",
+					k8s.ProxyLogLevelAnnotation:                "debug,linkerd2_proxy=debug",
+					k8s.ProxyDisableExternalProfilesAnnotation: "true"},
+			},
+		}
+	)
+
+	for _, testCase := range testCases {
```
siggy: 🙏

pkg/inject/inject_test.go
```diff
@@ -0,0 +1,412 @@
...
+			resourceConfig.podMeta.Annotations = podAnnotations
+
+			for _, annotation := range proxyConfigAnnotations {
+				actual := resourceConfig.getOverrideOrDefault(annotation)
+				expected, exists := expectedOverrides[annotation]
+				if !exists {
+					t.Errorf("Expected annotation %q to exist", annotation)
+				}
+
+				if !reflect.DeepEqual(expected, actual) {
```
alpeb: If both are strings, why not simply use `!=` ?

ihcsim:
> I was just wondering about the usage of `strconv.FormatUint` vs simply `strconv.Itoa` in `getOverrideOrDefault`

It's just a personal nit. Underneath the hood, `strconv.Itoa` uses ` FormatInt(int64(i), 10)`. Since the ports are declared as `uint`, I prefer not to have to convert an `uint` into an `int`, in order to use `strconv.Itoa`.

siggy:
@ihcsim i think the TLS issue i was seeing could be related to `proxy.linkerd.io/ignore-outbound-ports: "8079,8080"`, will keep testing.

ihcsim:
@siggy Yeah, same observation here. If I replaced `8079:8080` with something random like `9999`, TLS works again. Is there anything special about 8080?

siggy:
@ihcsim yup, emoji-svc and voting-svc both use 8080 ;)
```
$ kubectl -n emojivoto get svc
NAME         TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)        AGE
emoji-svc    ClusterIP      None             <none>        8080/TCP       42m
voting-svc   ClusterIP      None             <none>        8080/TCP       42m
web-svc      LoadBalancer   10.108.179.230   localhost     80:30909/TCP   42m
```

pkg/k8s/labels.go
```diff
@@ -82,6 +82,69 @@ const (
 	// in service identity.
 	IdentityModeAnnotation = Prefix + "/identity-mode"
 
+	/*
+	 * Proxy config annotations
+	 */
+
+	// ProxyImageAnnotation can be used to override the proxyImage config.
+	ProxyImageAnnotation = "proxy.linkerd.io/proxy-image"
```
klingerf: We should talk about the format of this label in more detail. I had originally anticipated all linkerd-related annotations would be prefixed with "linkerd.io" and this format breaks that assumption.

pkg/inject/inject.go
```diff
@@ -565,16 +591,95 @@ func (conf *ResourceConfig) AddRootLabels(patch *Patch) {
...
 }
 
 func (conf *ResourceConfig) taggedProxyInitImage() string {
 	return fmt.Sprintf("%s:%s",
-		conf.proxyConfig.GetProxyInitImage().GetImageName(),
+		conf.getOverrideOrDefault(k8s.ProxyInitImageAnnotation),
 		conf.globalConfig.GetVersion())
 }
 
+func (conf *ResourceConfig) getOverrideOrDefault(annotation string) string {
```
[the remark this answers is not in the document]
ihcsim: :+1: I actually did something very similar in my "override on update" commit, which will be included in this PR momentarily.

pkg/inject/inject.go
```diff
@@ -170,14 +180,20 @@ func (conf *ResourceConfig) GetPatch(
...
 
 		report.update(conf)
 		if shouldInject(conf, report) {
-			conf.injectPodSpec(patch, identity)
+			if err := conf.injectPodSpec(patch, identity); err != nil {
+				return nil, nil, err
+			}
 			conf.injectObjectMeta(patch)
+		} else {
+			if err := conf.overrideProxyConfigs(patch); err != nil {
```
siggy: just want to confirm this behavior for my own understanding. if `shouldInject == false`, but has a `linkerd-proxy` in the pod, `overrideProxyConfigs` will override the existing proxy settings in the pod.

does this mean if i specify some proxy config options during manual injection, those settings will be overridden by auto-inject?

ihcsim: > just want to confirm this behavior for my own understanding. if shouldInject == false, but has a linkerd-proxy in the pod, overrideProxyConfigs will override the existing proxy settings in the pod.

That's correct. The current behaviour is that it will always perform an override (even if the old vs. new values are the same), either by using values found in the config annotations, or the defaults from the config map.

> does this mean if i specify some proxy config options during manual injection, those settings will be overridden by auto-inject?

Those config options should be preserved. I still need to confirm with @klingerf on how to preserve them in light of his comment in [2447](https://github.com/linkerd/linkerd2/issues/2447#issuecomment-470320637), if we decided not to store these config options as annotations.

pkg/inject/inject.go
```diff
@@ -170,14 +182,17 @@ func (conf *ResourceConfig) GetPatch(
...
 			ControllerNamespace: conf.globalConfig.GetLinkerdNamespace(),
 		}
 
 		report.update(conf)
 		if shouldInject(conf, report) {
 			conf.injectPodSpec(patch, identity)
 			conf.injectObjectMeta(patch)
+		} else {
+			log.Infof("skipped auto-inject on %s/%s", strings.ToLower(conf.meta.Kind), report.Name)
+			conf.overrideProxyConfigs(patch)
```
klingerf: It looks like this fixes the issue where a previously-injected deployment is re-injected when its annotations change. But I'm a bit nervous that this could have unintended consequences, especially since we're calling it on all resources that fail the `shouldInject` check, not just those with a linkerd sidecar. Plus the way in which we override containers is totally separate from the main inject code path. We might want to consider changing `shouldInject` to consider pods with existing linkerd proxies to be injectable, so that they go through the same code path, but I haven't given it enough thought. I hate to ask, but would it be possible to undo this part of this PR, and add support for honoring updated annotations in a separate PR? Even if we undo this change, this PR will still suffice for override annotations on create, which is a great start, and I just want to give the update use case a bit more thought, possibly in the larger context of #2339.

ihcsim: Momentarily omitting this part sounds reasonable to me. At least per demo discussion, we are in agreement that the "override on updates" is a legit use case. I am hesitant to extend the `shouldInject` criteria, because IMO `shouldInject` and `overrideProxyConfigs` are two different tasks/responsibilities/features, with different resulting JSON patches. We'll chat more about it in the follow-up PR.

cli/cmd/inject_test.go
```diff
@@ -177,6 +177,12 @@ func TestUninjectAndInject(t *testing.T) {
 			reportFileName:   "inject_emojivoto_deployment.report",
 			testInjectConfig: noInitContainerConfig,
 		},
+		{
+			inputFileName:    "inject_emojivoto_deployment_config_overrides.input.yml",
+			goldenFileName:   "inject_emojivoto_deployment_config_overrides.golden.yml",
+			reportFileName:   "inject_emojivoto_deployment.report",
+			testInjectConfig: defaultConfig,
+		},
```
klingerf: Oooh, we could also use these fixtures in uinject_test.go to validate that the `config.linkerd.io` annotations are preserved when uninjecting.

```diff
diff --git a/cli/cmd/uninject_test.go b/cli/cmd/uninject_test.go
index e7fdb7c7..91a7bb2c 100644
--- a/cli/cmd/uninject_test.go
+++ b/cli/cmd/uninject_test.go
@@ -63,6 +63,11 @@ func TestUninjectYAML(t *testing.T) {
 			goldenFileName: "inject_emojivoto_deployment.input.yml",
 			reportFileName: "inject_emojivoto_deployment_uninject.report",
 		},
+		{
+			inputFileName:  "inject_emojivoto_deployment_config_overrides.golden.yml",
+			goldenFileName: "inject_emojivoto_deployment_config_overrides.input.yml",
+			reportFileName: "inject_emojivoto_deployment_uninject.report",
+		},
 		{
 			inputFileName:  "inject_emojivoto_pod_tls.golden.yml",
 			goldenFileName: "inject_emojivoto_pod.input.yml",
```

But for that to work, you have to tweak the annotations in the input file slightly, as follows:

```diff
diff --git a/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml b/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml
index 5f67c697..4cd6bf59 100644
--- a/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml
+++ b/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml
@@ -12,17 +12,17 @@ spec:
   strategy: {}
   template:
     metadata:
+      annotations:
+        config.linkerd.io/metrics-port: "9998"
+        config.linkerd.io/proxy-cpu-limit: "1"
+        config.linkerd.io/proxy-cpu-request: "0.5"
+        config.linkerd.io/proxy-memory-limit: 256Mi
+        config.linkerd.io/proxy-memory-request: 64Mi
+        config.linkerd.io/skip-inbound-ports: 7777,8888
+        config.linkerd.io/skip-outbound-ports: "9999"
       creationTimestamp: null
       labels:
         app: web-svc
-      annotations:
-        config.linkerd.io/skip-inbound-ports: 7777,8888
-        config.linkerd.io/skip-outbound-ports: 9999
-        config.linkerd.io/metrics-port: 9998
-        config.linkerd.io/proxy-cpu-limit: 1.0
-        config.linkerd.io/proxy-cpu-request: 0.5
-        config.linkerd.io/proxy-memory-request: 64Mi
-        config.linkerd.io/proxy-memory-limit: 256Mi
     spec:
       containers:
       - env:
```

pkg/inject/inject.go
```diff
@@ -577,16 +544,354 @@ func (conf *ResourceConfig) AddRootLabels(patch *Patch) {
...
+			break
+		}
+	}
+}
+
 func (conf *ResourceConfig) taggedProxyImage() string {
-	return fmt.Sprintf("%s:%s",
-		conf.proxyConfig.GetProxyImage().GetImageName(),
-		conf.globalConfig.GetVersion())
+	return fmt.Sprintf("%s:%s", conf.proxyImage(), conf.globalConfig.GetVersion())
```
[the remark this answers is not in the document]
ihcsim: Created #2497 to track this.

pkg/inject/inject.go
ihcsim: Created #2498 to track this.

Issue #1751, opened 2018-10-10
Update proxy-injector to support non-deployment workloads


klingerf:
Once #1714 merges, the proxy-injector process can be used to auto-inject deployment configurations that are applied to the cluster. The process does not support other workloads that include pod specs, such as statefulsets, daemonsets, and individual pod definitions. We should update the process to support these.

reegnz:
I just encountered this issue with StatefulSet, verified that it also doesn't work with DaemonSet.

I think the best approach would be to move towards injecting pods like suggested in #2342

alpeb:
Fixed via #2472

PR #2472, opened 2019-03-07
Have the Webhook react to pod creation/update only

### LINKED ISSUE #2342: autoinject: Inject Pods rather than Deployments

alpeb:
This was already working almost out-of-the-box, just had to:

- Change the webhook config so it watches pods instead of deployments
- Grant some extra ClusterRole permissions
- Add the piece that figures what's the OwnerReference and add the label
for it

Tested fine when applying new yamls for Deployments and StatefulSets.

Fixes #2342 and #1751


alpeb:
This is ready for review now.
One caveat: we're not copying the pod labels into its parent workload (deployment, job, daemonset or statefulset). When using the CLI, that is done automatically by Kubernetes when creating the workload with `kubectl apply`. And we were doing that manually in the webhook when we were listening to Deployments.
Doing that now that we're only dealing with pod creation/update events would require hitting the Kubernetes API to update those labels, and additional permissions for the proxy-injector ClusterRole.
But if I'm not mistaking those labels aren't really useful. If I'm mistaking, let me know :wink:

olix0r:
If this is particularly difficult to fix, I should note that the `identity/master` branch changes the ID structure such that no parent information need be reference by the injector.

alpeb:
Thanks for the detailed check @siggy! (I'll have to get into the habit of testing things with TLS enabled...)

I tried fixing the `LINKERD2_PROXY_ID` value by setting it to the pod's "generated name", which would end up something like `vote-bot-c78795449-.pod.$LINKERD2_PROXY_POD_NAMESPACE.linkerd-managed.linkerd.svc.cluster.local`, but that still doesn't fix the proxy error. Having it refer to the parent deployment would require changes that will need to be discarded afterwards per @olix0r's comment above, so I'm leaving it as is for now.

As for the missing serviceaccount mount paths, I believe the cause is as follows:

TL;DR: The injected sidecar + initContainer don't pass through the Service Account admission controller, so we have to manually add the missing mount paths.

The Service Account admission controller is the one in charge of automatically adding those mounts when a pod enters the system, and that's always run _before_ our mutating admission webhook.

When we were dealing with Deployments, we injected our sidecar and initContainer into the Deployment spec, then that created a Pod which went through the Service Account admission controller and got its mount paths set.

Now that we listen to Pod creation only, the Pod we receive in the webhook has already gone through the Service Account admission controller, which is confirmed by the mount path being correctly set in the uninjected container. We then proceed to inject our containers, but the Pod won't go again through the Service Account admission controller, so the mount path for our new containers won't get set.

As a solution I will be manually adding those mount paths.

alpeb:
Following up on our conversation, I've disabled `--tls optional` to avoid hitting these errors. This means this branch will depend on the Identity branch to get merged first.

alpeb:
I've just rebased now that the `--tls` flag in master has been dropped, so this branch can be reviewed without any disclaimers. The plan is to merge this before #2500, as it simplifies the webhook logic.

controller/proxy-injector/webhook.go
```diff
@@ -148,3 +159,36 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
 
 	return admissionResponse, nil
 }
+
+func (w *Webhook) parentRefLabel(ns string, ref metav1.OwnerReference) (string, string, error) {
```
klingerf: We already have a `GetOwnerKindAndName` func for finding pod owners in controller/k8s/api.go, and my preference would be to use that here.

https://github.com/linkerd/linkerd2/blob/4eb89bb8c22fa161be801038c090c3796d60f025/controller/k8s/api.go#L328-L347

In order to do that, you'll need to wire through a `k8s.API` object from `controller/cmd/proxy-injector/main.go` to this file. My suggestion would be to create it in main.go as follows:

```
	k8sAPI, err := k8s.InitializeAPI(*kubeconfig, k8s.NS, k8s.RS)
```

So that it automatically watches namespaces and replicasets. That will also allow us to replace our uncached namespace reads in this file with cached reads (e.g. `k8sAPI.NS().Lister().Get(request.Namespace)`).

chart/templates/proxy_injector.yaml
```diff
@@ -81,6 +81,9 @@ rules:
 - apiGroups: [""]
   resources: ["namespaces"]
   verbs: ["get"]
+- apiGroups: ["", "extensions"]
+  resources: ["replicationcontrollers", "replicasets"]
+  verbs: ["get"]
```
klingerf: If you take my `GetOwnerKindAndName` suggestion below, then the proxy injector won't need get privileges for replicationcontrollers, but it will need list and watch privileges for replicasets. I'd change this to:

```
- apiGroups: ["extensions"]
  resources: ["replicasets"]
```

pkg/k8s/labels.go
```diff
@@ -183,6 +183,10 @@ const (
 	// MountPathBase is the base directory of the mount path.
 	MountPathBase = "/var/run/linkerd"
 
+	// MountPathServiceAccount is the default path where Kuberenetes stores
+	// the service account token
+	MountPathServiceAccount = "/var/run/secrets/kubernetes.io/serviceaccount"
```
klingerf: Are the service account volume mount changes in this diff related to the pod watch change that you're making, or are they maybe a result of a merge issue with the identity stuff?

alpeb: @klingerf it's related to the missing mount paths issue I referred to in this [comment](https://github.com/linkerd/linkerd2/pull/2472#issuecomment-472115321)

controller/proxy-injector/webhook.go
```diff
@@ -129,11 +131,20 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
...
+		if err != nil {
+			return nil, err
+		}
+		p.AddPodLabel(k, v)
+	} else {
+		// When adding workloads through `kubectl apply` the spec template labels are
+		// automatically copied to the workload's main metadata section.
+		// This doesn't happen when adding labels through the webhook. So we manually
+		// add them to remain consistent.
+		conf.AddRootLabels(p)
```
ihcsim: IIUC, if the workload is a single pod, this will just re-add the labels to its `metadata`, right? I *think* issuing `add` patches to re-add existing labels won't break anything, at least, that's the behaviour with `kubectl`.

pkg/k8s/labels.go
klingerf: Ah, I totally missed that, thanks. Will go back and take a look.

ihcsim:
One TIOLI: I notice there is still a `patch = NewPatchDeployment()` in [`GetPatch()`](https://github.com/linkerd/linkerd2/blob/967a8e1d27a861fb3b476e5fe19081b4be5d31ec/pkg/inject/inject.go#L173). We definitely still need it to support CLI inject of different kinds of workloads. I wonder if we can make it more generic, since those patches work for other non-workload types. (For a second, I thought that only works for `Deployment` and we somehow broke CLI inject :laughing:).

One suggestion is to just replace the two constructors with a more general `NewPatch()` constructor, and use a `kind` parameter to decide which paths to use.

controller/proxy-injector/webhook.go
alpeb: @klingerf I did it like that because the admission request gets me the `ownerReferences` JSON outside of the pod spec. But I can simply manufacture a proper `corev1.Pod` settings by setting those refs manually, and pass it to `GetOwnerKindAndName` (whose caching properties I was unaware of, so that'll be an important gain). On the other hand, `GetOwnerKindAndName` doesn't consider Replication Controllers; should I just add support for that in that function?

controller/proxy-injector/webhook.go
alpeb: I think the `else` part of that snippet and `AddRootLabels()` don't make sense anymore. That was for updating the `Deployment` metadata, which is no longer there since we're watching Pods. You agree?

controller/proxy-injector/webhook.go
klingerf: Good questions. We don't call `GetOwnerKindAndName` in too many places, so I'd be OK with changing its method signature to make it more accommodating for your use case. Am pretty sure we could get by with passing `pod.ObjectMeta` to that function, instead of the full `Pod` object.

`GetOwnerKindAndName` already supports ReplicationControllers. All types of workloads can be represented in the `OwnerReference` `Kind` field. If the `Kind` is set to "ReplicationController", we'll return the `Kind` and `Name` fields without needing to look up the full ReplicationController object. Only if the `Kind` is set to "ReplicaSet" do we need to do an additional lookup to get the full ReplicaSet object, so that we can get its owner references, which will give us the Deployment. For an example of how it works, see the `GetOwnerKindAndName` tests, here:

https://github.com/linkerd/linkerd2/blob/master/controller/k8s/api_test.go#L588

alpeb:
@ihcsim yeah, `NewPatchDeployment` is not a generic-enough name hehe. I'll implement your suggestion based on 'Kind` :+1:

controller/proxy-injector/webhook.go
alpeb: Gotcha. I was under the impression that `ReplicationControllers` could themselves have an owner reference, but that's not the case :grimacing:

controller/proxy-injector/webhook.go
ihcsim: Yeah, I think removing it makes sense. I tested this further with `kubectl`. If I tried to re-annotate a workload with the same value, it errors out if there isn't an `--overwrite` flag, which leads me to think sending an `add` patch again will trigger the same error.

```
k annotate pod nginx-overrides-5d5b675ddd-cznwh linkerd.io/config=enabled
error: --overwrite is false but found the following declared annotation(s): 'linkerd.io/config' already has a value (enabled)
```

ihcsim:
In addition to the issue with auto-injecting single pod without labels, there is another minor one. The log `msg` is missing the workload name:
```
proxy-injector time="2019-03-20T22:38:55Z" level=info msg="patch generated for: Pod"
```
I think the `ResourceConfig.String()` method needs to look for the workload name in `podSpec`, instead of `workLoad`.

alpeb:
> I think the `ResourceConfig.String()` method needs to look for the workload name in `podSpec`, instead of `workLoad`.

Actually `podSpec` doesn't hold the pod's name either, so we have to rely on the admission request data. I'll fix that.

alpeb:
Thank you guys for all your feedback. Here's a new push with everything addressed.

@klingerf I'm now using `GetOwnerKindAndName()` as you suggested, although I finally didn't get to change its signature because it implied lots of changes elsewhere so it wasn't worth it.

@ihcsim Given the payload received by the webhook is a little different in regards to metadata, I had to get rid of the `ResourceConfig.String()` method, and I'm letting the report produce that string (I also dropped the namespace from there for simplicity).

Sorry I had to rebase one more time, to keep in sync with the Identity changes.

controller/proxy-injector/webhook.go
```diff
@@ -150,3 +158,28 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
 
 	return admissionResponse, nil
 }
+
+func (w *Webhook) getLabelForParent(conf *inject.ResourceConfig) (string, string, error) {
```
ihcsim: I feel this belongs in the `case *v1.Pod` of the `inject.Parse()` method. All the other workloads have this label set there.

controller/proxy-injector/webhook_test.go
```diff
@@ -153,6 +154,158 @@ func TestGetPatch(t *testing.T) {
...
+			conf := confNsEnabled()
+			nonEmpty, err := conf.ParseMeta(b, "")
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !nonEmpty {
+				t.Fatalf("Unexpected empty result from ParseMeta()")
+			}
+
+			if err := conf.Parse(b); err != nil {
```
ihcsim: I don't think we should export this function just for the test. Can you use `inject.ParseMetaAndYaml()` here?

pkg/inject/inject.go
```diff
@@ -150,22 +137,27 @@ func (conf *ResourceConfig) YamlMarshalObj() ([]byte, error) {
...
+	return &r, conf.Parse(bytes)
 }
 
 // ParseMeta extracts metadata from bytes.
-// It returns false if the workload's payload is empty
-func (conf *ResourceConfig) ParseMeta(bytes []byte) (bool, error) {
+// If bytes doesn't contain the namespace (webhook case) then it must be
+// provided in the second argument.
+// It returns false if the workload's payload is empty.
+func (conf *ResourceConfig) ParseMeta(bytes []byte, ns string) (bool, error) {
```
ihcsim: Per _slack_ discussion, the second argument isn't necessary since it's not being used anywhere.

pkg/inject/inject.go
```diff
@@ -362,10 +355,20 @@ func (conf *ResourceConfig) complete(template *v1.PodTemplateSpec) {
 	conf.pod.Meta = &template.ObjectMeta
 }
 
+// GetPod returns the parsed object as a *v1.Pod variable
+func (conf *ResourceConfig) GetPod() (*v1.Pod, error) {
```
ihcsim: Probably not needed if you moved the logic of `webhook.getLabelForParent()` into this package.

pkg/inject/report.go
```diff
@@ -27,21 +27,25 @@ func newReport(conf *ResourceConfig) Report {
 	var name string
 	if m := conf.pod.Meta; m != nil {
 		name = m.Name
+		if name == "" {
+			name = m.GenerateName
```
ihcsim: Not sure how useful this will be, as it only returns the prefix of the pod name like `linkerd-controller-7c6b8c586b-`. Will there be cases where the pod name will be empty? The parent workload name, if non-empty, might be more useful.

pkg/inject/inject.go
```diff
@@ -351,7 +340,11 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 
 		conf.workload.obj = v
 		conf.pod.spec = &v.Spec
+		ns := conf.pod.Meta.Namespace
 		conf.pod.Meta = &v.ObjectMeta
+		if ns != "" {
+			conf.pod.Meta.Namespace = ns
```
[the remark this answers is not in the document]
alpeb: The namespace is missing in the `AdmissionRequest` metadata, so the idea here is to fill it in. It occurred to me that was a natural place to put it (already supported in the `Meta` struct), instead of augmenting `ResourceConfig`.
@ihcsim made me realize we're not using namespace anywhere in the injection logic for pods (which is the webhook case, which is the case where the namespace has to be "filled-in") so we can get rid of all this.

_However_:

We're using the namespace when injecting Deployments through the CLI [here](https://github.com/linkerd/linkerd2/blob/alpeb/webhook-pods/pkg/inject/inject.go#L272-L277) to be able to set `destinationDNSOverride`. 
So we could actually be needing to use namespaces for this same purpose in the particular case of having the webhook inject a linkerd controller pod that wasn't injected to begin with. We're currently not supporting that case, although it would be a very rare and non-default one. Do you think it's worth to consider it?

chart/templates/proxy_injector.yaml
```diff
@@ -80,7 +80,13 @@ rules:
   verbs: ["create", "get", "delete"]
 - apiGroups: [""]
   resources: ["namespaces"]
-  verbs: ["get"]
+  verbs: ["list", "get", "watch"]
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["list"]
```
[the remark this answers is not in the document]
siggy: @klingerf Good catch. Agree it doesn't make sense for the proxy injector to have "list pods" privileges just to pass this check. Ideally the authorization code should perform a `SelfSubjectAccessReview` check based on the requirements of each individual component.

I think this is fine for now, but we should fix it so we can remove "list pods" capability from proxy injector, have filed #2559 to track.

controller/proxy-injector/webhook_test.go
```diff
@@ -153,6 +154,158 @@ func TestGetPatch(t *testing.T) {
...
+				pod: `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: emoji-svc
+  namespace: emojivoto
+  ownerReferences:                                                      
+  - apiVersion: apps/v1beta2
+    kind: Deployment                                                    
+    name: emoji-deploy`,
```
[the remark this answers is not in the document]
alpeb: Indeed! :blush:

chart/templates/proxy_injector.yaml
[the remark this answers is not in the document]
klingerf: @siggy Awesome, thanks for writing that up.

pkg/inject/inject.go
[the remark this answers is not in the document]
klingerf: Ahhh, ok great, thanks for the additional explanation. That all makes sense. I'm in favor of reverting these changes, and not worrying about the case where we might auto-inject the proxy into the control plane. We should assume we're always generating injected YAML as part of the install path.

pkg/inject/inject.go
alpeb: `inject.go` is not dependent on `k8s.API`, which is needed in `getLabelForParent()`. That's why I kept the pod object retrieval logic separated from the logic that retrieves its ownership, which is only used in the webhook.

pkg/inject/report.go
alpeb: The pod name appears to be always empty in the admission request metadata, and we only have access to that `GenerateName`, which is actually the ReplicaSet's name (plus a dash). I think this is a bit more informative (`linkerd-controller-7c6b8c586b-`) than the Deployment's name (`linkerd-controller`).

controller/proxy-injector/webhook.go
alpeb: True. That would require adding the 'k8s.API` into `inject.go`'s `ResourceConfig`, which is probably fine. What do you think @siggy and/or @klingerf ?

klingerf: As a general rule, none of the package in `./pkg/...` should depend on any of the packages in `./controller/...`. So I'd prefer not to add a new field in `ResourceConfig` that's of type `k8s.API`. That said, we could create a new field that holds a func that produces the label, and pass the func in when setting up the config.... or something along those lines.

ihcsim: Sorry, I was hoping that the `AdmissionRequest` would have given us the info we need, without needing to use `k8s.API`. If this is too much work, it's probably not worth it.

controller/proxy-injector/webhook.go
```diff
@@ -131,17 +135,21 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
...
+		p.AddPodLabel(key, name)
+	}
 
 	patchJSON, err := p.Marshal()
 	if err != nil {
 		return nil, err
 	}
-	log.Infof("patch generated for: %s", conf)
+	// TODO: refactor GetPatch() so it only returns one report item
+	r := reports[0]
```
siggy: tioli: consider a `len(reports) != 1` sanity check, to ensure we don't panic.

controller/proxy-injector/webhook.go
```diff
@@ -150,3 +158,28 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
...
+		case pkgK8s.Job:
+			return pkgK8s.ProxyJobLabel, name, nil
+		case pkgK8s.DaemonSet:
+			return pkgK8s.ProxyDaemonSetLabel, name, nil
+		case pkgK8s.StatefulSet:
+			return pkgK8s.ProxyStatefulSetLabel, name, nil
+		}
+		return "", "", fmt.Errorf("unsupported parent kind \"%s\"", kind)
+	}
+	return "", "", nil
```
siggy: tioli, maybe slightly clearer implementing all cases via the switch?
```go
kind, name := w.k8sAPI.GetOwnerKindAndName(pod)
switch kind {
case pkgK8s.Deployment:
	return pkgK8s.ProxyDeploymentLabel, name, nil
case pkgK8s.ReplicationController:
	return pkgK8s.ProxyReplicationControllerLabel, name, nil
case pkgK8s.ReplicaSet:
	return pkgK8s.ProxyReplicaSetLabel, name, nil
case pkgK8s.Job:
	return pkgK8s.ProxyJobLabel, name, nil
case pkgK8s.DaemonSet:
	return pkgK8s.ProxyDaemonSetLabel, name, nil
case pkgK8s.StatefulSet:
	return pkgK8s.ProxyStatefulSetLabel, name, nil
case pkgK8s.Pod:
	return "", "", nil
default:
	return "", "", fmt.Errorf("unsupported parent kind \"%s\"", kind)
}
```

pkg/inject/inject.go
```diff
@@ -565,15 +588,11 @@ func (conf *ResourceConfig) injectObjectMeta(patch *Patch) {
 		patch.addPodAnnotation(k8s.IdentityModeAnnotation, k8s.IdentityModeDisabled)
 	}
 
-	for k, v := range conf.pod.labels {
-		patch.addPodLabel(k, v)
+	if len(conf.pod.Meta.Labels) == 0 {
```
siggy: should this be... ?

```go
if len(conf.pod.labels) != 0 {
	if len(conf.pod.Meta.Labels) == 0 {
		patch.addPodLabelsRoot()
	}
	for k, v := range conf.pod.labels {
		patch.AddPodLabel(k, v)
	}
}
```

controller/proxy-injector/webhook.go
```diff
@@ -159,27 +151,9 @@ func (w *Webhook) inject(request *admissionv1beta1.AdmissionRequest) (*admission
...
-			return pkgK8s.ProxyReplicaSetLabel, name, nil
-		case pkgK8s.Job:
-			return pkgK8s.ProxyJobLabel, name, nil
-		case pkgK8s.DaemonSet:
-			return pkgK8s.ProxyDaemonSetLabel, name, nil
-		case pkgK8s.StatefulSet:
-			return pkgK8s.ProxyStatefulSetLabel, name, nil
-		}
-		return "", "", fmt.Errorf("unsupported parent kind \"%s\"", kind)
+func (w *Webhook) ownerRetriever(ns string) inject.OwnerRetrieverFunc {
```
alpeb: who said go cannot haz curry? :cat:

controller/proxy-injector/webhook_test.go
```diff
@@ -154,158 +153,6 @@ func TestGetPatch(t *testing.T) {
 	})
 }
 
-func TestGetLabelForParent(t *testing.T) {
```
alpeb: Given we no longer have `GetLabelForParent()` I removed this test, but created #2560 to complete the webhook tests, that we can address in the 2.3 testing sprint.

pkg/inject/inject.go
```diff
@@ -352,6 +347,25 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.workload.obj = v
 		conf.pod.spec = &v.Spec
 		conf.pod.Meta = &v.ObjectMeta
+
+		if conf.ownerRetriever != nil {
```
klingerf: 👍 I like this updated approach

pkg/inject/patch_test.go
```diff
@@ -34,16 +34,14 @@ func TestPatch(t *testing.T) {
 		createdBy           = "linkerd/cli v18.8.4"
 	)
 
-	actual := NewPatchDeployment()
+	actual := NewPatch(k8sPkg.Deployment)
 	actual.addContainer(sidecar)
 	actual.addInitContainerRoot()
 	actual.addInitContainer(init)
-	//actual.addVolumeRoot()
-	//actual.addVolume(secrets)
```
klingerf: I'd actually recommend uncommenting these lines, instead of removing them. And also uncommenting lines 16-21 and 49-50 of this file, so that we can start verifying that the volume patches are working correctly. You'd need to create a new secret, however, in order for this to work. See https://github.com/linkerd/linkerd2/commit/91c5f07650c3dc25e6cab0134e6a3185e8ec20b4#diff-4940a4b4cb3bf0d6f5d4bf9090e2ea6e, which shows when the original tests were removed.

pkg/inject/inject.go
```diff
@@ -352,6 +347,25 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
...
+				conf.pod.labels[k8s.ProxyReplicationControllerLabel] = name
+			case k8s.ReplicaSet:
+				conf.pod.labels[k8s.ProxyReplicaSetLabel] = name
+			case k8s.Job:
+				conf.pod.labels[k8s.ProxyJobLabel] = name
+			case k8s.DaemonSet:
+				conf.pod.labels[k8s.ProxyDaemonSetLabel] = name
+			case k8s.StatefulSet:
+				conf.pod.labels[k8s.ProxyStatefulSetLabel] = name
+			case k8s.Pod:
```
siggy: just confirming this empty case is here for completeness?

alpeb: The idea was to not do anything in that case. I'd better just remove it.

PR #2476, opened 2019-03-08
Add "Community" menu item to dashboard that displays linkerd.io content

scottcarol:
Fulfills #2327.

This PR creates a "Community" menu item on the dashboard sidebar that, when clicked, displays an iFrame of a page on `linkerd.io`. A yellow badge appears on the menu item if there has been an update since the user last clicked the "Community" menu item. This is calculated by comparing a date in the user's `localStorage` to a JSON feed at `linkerd.io/dashboard/index.json`.

### DIFF
--- a/web/app/js/components/Community.jsx
+++ b/web/app/js/components/Community.jsx
@@ -1,0 +1,20 @@
+import Iframe from 'react-iframe';
+import React from 'react';
+
+function Community() {
+  return (
+    <React.Fragment>
+      <Iframe
+        url="https://linkerd.io/dashboard/"
+        id="myId"
+        className="myClassname"
+        position="inherit"
+        display="block"
+        height="-webkit-fill-available"
+        width="-webkit-fill-available"
+        border="none" />
+    </React.Fragment>
+  );
+}
+
+export default Community;
--- a/web/app/js/components/Community.jsx
+++ b/web/app/js/components/Community.jsx
@@ -6,8 +6,6 @@ function Community() {
     <React.Fragment>
       <Iframe
         url="https://linkerd.io/dashboard/"
-        id="myId"
-        className="myClassname"
         position="inherit"
         display="block"
         height="-webkit-fill-available"
--- a/web/app/js/components/Community.jsx
+++ b/web/app/js/components/Community.jsx
@@ -1,17 +1,15 @@
 import Iframe from 'react-iframe';
 import React from 'react';
 
-function Community() {
+const Community = () => {
   return (
-    <React.Fragment>
-      <Iframe
-        url="https://linkerd.io/dashboard/"
-        position="inherit"
-        display="block"
-        height="100vh"
-        border="none" />
-    </React.Fragment>
+    <Iframe
+      url="https://linkerd.io/dashboard/"
+      position="inherit"
+      display="block"
+      height="100vh"
+      border="none" />
   );
-}
+};
 
 export default Community;
--- a/web/app/js/components/Community.jsx
+++ b/web/app/js/components/Community.jsx
@@ -8,8 +8,7 @@ function Community() {
         url="https://linkerd.io/dashboard/"
         position="inherit"
         display="block"
-        height="-webkit-fill-available"
-        width="-webkit-fill-available"
+        height="100vh"
         border="none" />
     </React.Fragment>
   );
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -34,6 +34,7 @@ import { withStyles } from '@material-ui/core/styles';
 import yellow from '@material-ui/core/colors/yellow';
 
 const jsonFeedUrl = "https://linkerd.io/dashboard/index.json";
+const localStorageKey = "linkerd-updates-last-clicked";
 
 const styles = theme => {
   const drawerWidth = theme.spacing.unit * 31;
@@ -184,7 +185,7 @@ class NavigationBase extends React.Component {
       .then(rsp => rsp.data.date)
       .then(rsp => {
         if (rsp.length > 0) {
-          let lastClicked = localStorage["linkerd-updates-last-clicked"];
+          let lastClicked = localStorage[localStorageKey];
           if (!lastClicked) {
             this.setState({ hideUpdateBadge: false });
           } else {
@@ -207,10 +208,8 @@ class NavigationBase extends React.Component {
 
   handleCommunityClick = () => {
     let lastClicked = new Date();
-    localStorage.setItem("linkerd-updates-last-clicked", lastClicked);
-    if (this.state.hideUpdateBadge === false) {
-      this.setState({ hideUpdateBadge: true });
-    }
+    localStorage.setItem(localStorageKey, lastClicked);
+    this.setState({ hideUpdateBadge: true });
   }
 
   handleDrawerClick = () => {
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -27,10 +27,13 @@ import SentimentVerySatisfiedIcon from '@material-ui/icons/SentimentVerySatisfie
 import Toolbar from '@material-ui/core/Toolbar';
 import Typography from '@material-ui/core/Typography';
 import Version from './Version.jsx';
+import _maxBy from 'lodash/maxBy';
 import classNames from 'classnames';
 import { withContext } from './util/AppContext.jsx';
 import { withStyles } from '@material-ui/core/styles';
 import yellow from '@material-ui/core/colors/yellow';
+
+const jsonFeedUrl = "https://linkerd.io/dashboard/index.json";
 
 const styles = theme => {
   const drawerWidth = theme.spacing.unit * 31;
@@ -176,7 +179,6 @@ class NavigationBase extends React.Component {
   }
 
   fetchLatestCommunityUpdate() {
-    let jsonFeedUrl = "https://linkerd.io/dashboard/index.json";
     this.communityUpdatesPromise = fetch(jsonFeedUrl)
       .then(rsp => rsp.json())
       .then(rsp => rsp.data.date)
@@ -186,10 +188,10 @@ class NavigationBase extends React.Component {
           if (!lastClicked) {
             this.setState({ hideUpdateBadge: false });
           } else {
-            lastClicked = new Date(lastClicked);
-            let latestCommunityUpdate = rsp.map(update => new Date(update.date))
-              .sort((a, b) => b - a)[0];
-            if (latestCommunityUpdate > lastClicked) {
+            let lastClickedDateObject = new Date(lastClicked);
+            let latestArticle = _maxBy(rsp, update => update.date);
+            let latestArticleDateObject = new Date(latestArticle);
+            if (latestArticleDateObject > lastClickedDateObject) {
               this.setState({ hideUpdateBadge: false });
             }
           }
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -181,15 +181,16 @@ class NavigationBase extends React.Component {
       .then(rsp => rsp.data.date)
       .then(rsp => {
         if (rsp.length > 0) {
-          let lastUpdated = rsp.map(update => new Date(update.date))
-            .sort((a, b) => b - a);
-          let lastChecked = localStorage["linkerd-updates-last-checked"];
-          if (!lastChecked) {
+          let lastClicked = localStorage["linkerd-updates-last-checked"];
+          if (!lastClicked) {
             localStorage.setItem("linkerd-updates-last-checked", new Date());
             this.setState({ hideUpdateBadge: false });
-          }
-          if (lastUpdated[0] > lastChecked) {
-            this.setState({ hideUpdateBadge: false });
+          } else {
+            let latestCommunityUpdate = rsp.map(update => new Date(update.date))
+              .sort((a, b) => b - a)[0];
+            if (latestCommunityUpdate > lastClicked) {
+              this.setState({ hideUpdateBadge: false });
+            }
           }
         }
       }).catch(this.handleApiError);
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -186,6 +186,7 @@ class NavigationBase extends React.Component {
           if (!lastClicked) {
             this.setState({ hideUpdateBadge: false });
           } else {
+            lastClicked = new Date(lastClicked);
             let latestCommunityUpdate = rsp.map(update => new Date(update.date))
               .sort((a, b) => b - a)[0];
             if (latestCommunityUpdate > lastClicked) {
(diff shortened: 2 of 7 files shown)


scottcarol:
@grampelberg A follow up on our colors conversation, the generated Material-UI `secondary` color was a reddish-pink so I went with yellow.

grampelberg:
Looks like firefox needs a little love:

Screenshot 2019-03-08 13 25 06

web/app/js/components/Navigation.jsx
```diff
@@ -122,6 +124,9 @@ const styles = theme => {
       fontSize: "18px",
       paddingLeft: "3px",
       paddingRight: "3px",
+    },
+    badge: {
+      backgroundColor: "#ffeb3b",
```
siggy: consider using material ui colors (i think it's equivalent)?
```diff
diff --git a/web/app/js/components/Navigation.jsx b/web/app/js/components/Navigation.jsx
index cc0bee28..5457bc26 100644
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -30,6 +30,7 @@ import Version from './Version.jsx';
 import classNames from 'classnames';
 import { withContext } from './util/AppContext.jsx';
 import { withStyles } from '@material-ui/core/styles';
+import yellow from '@material-ui/core/colors/yellow';

 const styles = theme => {
   const drawerWidth = theme.spacing.unit * 31;
@@ -126,7 +127,7 @@ const styles = theme => {
       paddingRight: "3px",
     },
     badge: {
-      backgroundColor: "#ffeb3b",
+      backgroundColor: yellow[500],
     }
   };
 };
```

web/app/js/components/Navigation.jsx
```diff
@@ -245,6 +298,14 @@ class NavigationBase extends React.Component {
               <ListItemIcon><LibraryBooksIcon /></ListItemIcon>
               <ListItemText primary="Documentation" />
             </ListItem>
+            { this.menuItem("/community", "Community",
+              <Badge
+                classes={{ badge: classes.badge }}
+                invisible={this.state.hideUpdateBadge}
+                badgeContent="1">
+                <SentimentVerySatisfiedIcon fontSize="small" />
```
siggy: change this to `medium` to align the menu items?

Screen Shot 2019-03-08 at 2 30 11 PM

web/app/js/components/Navigation.jsx
```diff
@@ -131,8 +136,10 @@ class NavigationBase extends React.Component {
     super(props);
     this.api = this.props.api;
     this.handleApiError = this.handleApiError.bind(this);
+    this.handleCommunityClick = this.handleCommunityClick.bind(this);
 
     this.state = this.getInitialState();
+    this.state.hideUpdateBadge = true;
```
siggy: should setting `hideUpdateBadge` just be part of `getInitialState`?

web/app/js/components/Navigation.jsx
```diff
@@ -166,12 +174,42 @@ class NavigationBase extends React.Component {
...
+      .then(rsp => {
+        if (rsp.length > 0) {
+          let lastClicked = localStorage["linkerd-updates-last-checked"];
+          if (!lastClicked) {
+            localStorage.setItem("linkerd-updates-last-checked", new Date());
+            this.setState({ hideUpdateBadge: false });
+          } else {
+            let latestCommunityUpdate = rsp.map(update => new Date(update.date))
+              .sort((a, b) => b - a)[0];
+            if (latestCommunityUpdate > lastClicked) {
```
siggy: is this doing a string comparison? would this be easier as an integer timestamp?
```js
new Date().getTime()
1552084942276
```

scottcarol:
fantastic catch @grampelberg - thanks!

web/app/js/components/Navigation.jsx
```diff
@@ -166,12 +174,42 @@ class NavigationBase extends React.Component {
...
+  fetchLatestCommunityUpdate() {
+    let jsonFeedUrl = "https://linkerd.io/dashboard/index.json";
+    this.communityUpdatesPromise = fetch(jsonFeedUrl)
+      .then(rsp => rsp.json())
+      .then(rsp => rsp.data.date)
+      .then(rsp => {
+        if (rsp.length > 0) {
+          let lastClicked = localStorage["linkerd-updates-last-checked"];
+          if (!lastClicked) {
+            localStorage.setItem("linkerd-updates-last-checked", new Date());
```
siggy: just want to confirm that we're intending to set `linkerd-updates-last-checked` on the date of the last page load, not the date of the last time the `Community` nav item was clicked. if so, considering renaming `lastClicked` to `lastChecked`.

related to this, users will typically view this page via `linkerd dashboard`, so the port will be different each time, causing the local storage to be empty and the badge to always show up. we've talked about pinning the dashboard port rather than random port assignment, but that change is probably outside the scope of this PR. i'm fine with this as-is just wanted to call it out.

web/app/js/components/Navigation.jsx
```diff
@@ -180,22 +218,37 @@ class NavigationBase extends React.Component {
...
-      <MenuItem
-        component={Link}
-        to={api.prefixLink(path)}
-        className={classes.navMenuItem}
-        selected={isCurrentPage(path)}>
-        <ListItemIcon>{icon}</ListItemIcon>
-        <ListItemText primary={title} />
-      </MenuItem>
-    );
+    if (isCommunity) {
```
siggy: consider passing in `onClick`, that way you won't have to if/else based on `isCommunity`:
```js
menuItem(path, title, icon, onClick) {
...
    return (
      <MenuItem
        component={Link}
        onClick={onClick}
```

web/app/js/components/Navigation.jsx
scottcarol: Good 👁, I just removed the attribute to restore the default size.

web/app/js/components/Navigation.jsx
scottcarol: good point!

web/app/js/components/Navigation.jsx
scottcarol: On lines 189-190, I'm creating a map of `Date` instances from the returned JSON dates, then passing a `compareFunction` into `sort()` that subtracts them to get a value that is either negative, positive or zero. 

This is so I can sort in descending order (latest-earliest) which requires me to use a `compareFunction`.

Using `getTime()` on the `Date` instances would work too, but since it would add an extra step and I would still need a `compareFunction` I'm inclined to keep as-is.

web/app/js/components/Navigation.jsx
scottcarol: re: `linkerd-updates-last-checked`, thanks for the questions, i realized i was overcomplicating the logic. i've now simplified so that the only place where `localStorage` is set is in `handleCommunityClick()`. 

re: port assignments, thanks for pointing this out. 🔍🕵️  as we talked about in person, will make a separate PR for this.

web/app/js/components/Navigation.jsx
scottcarol: Great tip, thanks - I didn't realize that React would skip the `onClick` attribute if no value is supplied to the `menuItem` function - solves the problem of the extra code block!

web/app/js/components/Navigation.jsx
scottcarol: Thanks, am now using the color library from Material UI ✅

web/app/js/components/Navigation.jsx
scottcarol: Addressed in PR #2486

web/app/js/components/Navigation.jsx
siggy: Hmm, I was mostly curious about `lastClicked`, which `localStorage["linkerd-updates-last-clicked"]`, which I think is a string? I'm not totally sure what happens in `latestCommunityUpdate > lastClicked` if its comparing a `Date` with a string?

scottcarol: You are totally right, thanks for clarifying! Let me fix now.

siggy: Confirmed this all works, thanks for the fix.

For a bit more context, I'm generally paranoid about storing date strings and opt for integer timestamps in UTC. JS's `Date.getTime()` is milliseconds since the unix epoch, not quite the same as unix timestamps (seconds since epoch), but close. It eliminates format-specific parsing or consideration for timezones, and makes sorting/comparing easier, particularly across languages.

Going forward we'll want to be mindful that we're syndicating https://linkerd.io/dashboard/index.json timestamps as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), and storing [ECMA-262](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString#Description) in `localstorage`. Again this is all pretty academic, just want to be intentional about supporting this.

grampelberg:
@scottcarol can we get the content updated on the `linkerd.io` side before the next edge release? I'd hate for edge users to see the current lorem ipsum =)

scottcarol:
@grampelberg Definitely, was on my list to check with you before merging - agreed about no lorem ipsum. Let me know when you have the `linkerd.io` content good to go - I'm also waiting on a review from @klingerf.

grampelberg:
@scottcarol I was kinda hoping you'd take a first whack at the content, because I'm lazy =)

web/app/js/components/Navigation.jsx
```diff
@@ -245,6 +288,14 @@ class NavigationBase extends React.Component {
               <ListItemIcon><LibraryBooksIcon /></ListItemIcon>
               <ListItemText primary="Documentation" />
             </ListItem>
+            { this.menuItem("/community", "Community",
+              <Badge
+                classes={{ badge: classes.badge }}
+                invisible={this.state.hideUpdateBadge}
+                badgeContent="1">
```
klingerf: It seems like we could set the number in the badge to be the actual number of updates since the user last clicked, based on the timestamp of each of the updates in the JSON feed. How would you feel about something like?

```diff
diff --git a/web/app/js/components/Navigation.jsx b/web/app/js/components/Navigation.jsx
index 2434bd03..365ef498 100644
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -27,7 +27,7 @@ import SentimentVerySatisfiedIcon from '@material-ui/icons/SentimentVerySatisfie
 import Toolbar from '@material-ui/core/Toolbar';
 import Typography from '@material-ui/core/Typography';
 import Version from './Version.jsx';
-import _maxBy from 'lodash/maxBy';
+import _filter from 'lodash/filter';
 import classNames from 'classnames';
 import { withContext } from './util/AppContext.jsx';
 import { withStyles } from '@material-ui/core/styles';
@@ -149,7 +149,7 @@ class NavigationBase extends React.Component {
     return {
       drawerOpen: true,
       helpMenuOpen: false,
-      hideUpdateBadge: true,
+      communityUpdateCount: 0,
       latestVersion: '',
       isLatest: true,
       namespaceFilter: "all"
@@ -186,14 +186,13 @@ class NavigationBase extends React.Component {
         if (rsp.length > 0) {
           let lastClicked = localStorage["linkerd-updates-last-clicked"];
           if (!lastClicked) {
-            this.setState({ hideUpdateBadge: false });
+            this.setState({ communityUpdateCount: rsp.length });
           } else {
             let lastClickedDateObject = new Date(lastClicked);
-            let latestArticle = _maxBy(rsp, update => update.date);
-            let latestArticleDateObject = new Date(latestArticle);
-            if (latestArticleDateObject > lastClickedDateObject) {
-              this.setState({ hideUpdateBadge: false });
-            }
+            let newArticles = _filter(rsp, update => {
+              return new Date(update.date) > lastClickedDateObject;
+            });
+            this.setState({ communityUpdateCount: newArticles.length });
           }
         }
       }).catch(this.handleApiError);
@@ -208,9 +207,7 @@ class NavigationBase extends React.Component {
   handleCommunityClick = () => {
     let lastClicked = new Date();
     localStorage.setItem("linkerd-updates-last-clicked", lastClicked);
-    if (this.state.hideUpdateBadge === false) {
-      this.setState({ hideUpdateBadge: true });
-    }
+    this.setState({ communityUpdateCount: 0 });
   }
 
   handleDrawerClick = () => {
@@ -291,8 +288,8 @@ class NavigationBase extends React.Component {
             { this.menuItem("/community", "Community",
               <Badge
                 classes={{ badge: classes.badge }}
-                invisible={this.state.hideUpdateBadge}
-                badgeContent="1">
+                invisible={this.state.communityUpdateCount === 0}
+                badgeContent={this.state.communityUpdateCount}>
                 <SentimentVerySatisfiedIcon />
               </Badge>, this.handleCommunityClick
               ) }
```

web/app/js/components/Community.jsx
```diff
@@ -0,0 +1,17 @@
...
+function Community() {
+  return (
+    <React.Fragment>
+      <Iframe
+        url="https://linkerd.io/dashboard/"
+        position="inherit"
+        display="block"
+        height="100vh"
+        border="none" />
+    </React.Fragment>
```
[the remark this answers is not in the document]
scottcarol: Oops forgot to remove that! Good catch.

web/app/js/components/Community.jsx
```diff
@@ -0,0 +1,17 @@
+import Iframe from 'react-iframe';
+import React from 'react';
+
+function Community() {
```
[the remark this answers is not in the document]
scottcarol: Thanks!

web/app/js/components/Navigation.jsx
```diff
@@ -122,6 +128,9 @@ const styles = theme => {
       fontSize: "18px",
       paddingLeft: "3px",
       paddingRight: "3px",
+    },
+    badge: {
+      backgroundColor: yellow[500],
```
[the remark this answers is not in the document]
scottcarol: 🌈

web/app/js/components/Navigation.jsx
scottcarol: Thanks for these suggestions, and I will save this to look at in the future when we improve this feature. Initially we are hard-coding the number of updates to "1" in order to not overwhelm the users, but this may change!

web/app/js/components/Navigation.jsx
```diff
@@ -166,12 +178,41 @@ class NavigationBase extends React.Component {
       }).catch(this.handleApiError);
   }
 
+  fetchLatestCommunityUpdate() {
+    this.communityUpdatesPromise = fetch(jsonFeedUrl)
+      .then(rsp => rsp.json())
+      .then(rsp => rsp.data.date)
+      .then(rsp => {
+        if (rsp.length > 0) {
+          let lastClicked = localStorage["linkerd-updates-last-clicked"];
```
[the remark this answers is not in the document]
scottcarol: 👍

scottcarol:
Awesome, thanks all for review feedback. @grampelberg let me know when `linkerd.io/dashboard` content is good to go and I'll merge this to master 🌈

PR #2478, opened 2019-03-08
Stop background processes on failure

grampelberg:
`bin/go-run` failed for me locally, ended up with a bunch in the background I was unaware of and nothing worked. This'll make that whole process a little easier =)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -247,12 +247,14 @@ These commands assume working [Go](https://golang.org) and
 
 #### First time setup
 
-Install [Yarn](https://yarnpkg.com) and use it to install dependencies:
+1. Install [Yarn](https://yarnpkg.com) and use it to install dependencies:
 
-```bash
-brew install yarn
-bin/web setup
-```
+    ```bash
+    brew install yarn
+    bin/web setup
+    ```
+
+1. Install Linkerd on a Kubernetes cluster.
 
 #### Run web standalone
 
@@ -265,8 +267,6 @@ The web server will be running on `localhost:8084`.
 #### Webpack dev server
 
 To develop with a webpack dev server:
-
-1. Install Linkerd on a Kubernetes cluster.
 
 1. Start the development server.
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -264,13 +264,30 @@ The web server will be running on `localhost:8084`.
 
 #### Webpack dev server
 
-To develop with a webpack dev server, run:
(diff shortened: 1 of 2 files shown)


BUILD.md
```diff
@@ -264,13 +264,30 @@ The web server will be running on `localhost:8084`.
 
 #### Webpack dev server
 
-To develop with a webpack dev server, run:
+To develop with a webpack dev server:
 
-```bash
-bin/web dev
-```
+1. Install Linkerd on a Kubernetes cluster.
```
alpeb: Shouldn't this item go above instead, under the Web heading, as a prerequisite to both web standalone and webpack dev server?

bin/web
```diff
@@ -7,6 +7,10 @@ ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/.. && pwd )"
 DEV_PORT=8080
 export NODE_ENV=${NODE_ENV:=development}
 
+# Stop everything in the process group (in the background) whenever the
+# parent process experiences an error and exits.
+trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
```
alpeb: **Nice**. This forced me to dig into quite some bash docs to fully understand, and it's awesome.
I also realized that even when there are no errors, when you killed `bin/web dev` the processes for `kubectl port-forward` remained in the background, and this fixes that as well :+1:

alpeb: Turns out CI is not happy about it because when `/bin/web` finishes it returns status 0.
Getting rid of the `EXIT` in this trap fixes that, but it seems that interferes with trapping the errors from other scripts (maybe because of `set -e` in those other scripts?).
Not sure what to do in this case. Maybe it's not necessary for CI to run a naked `/bin/web`...

grampelberg: @alpeb I've been playing around and seeing what to do, the problem is that it signals the parent as well as the children. Just need to figure out how to limit to the children only...

grampelberg:
@alpeb realized that this doesn't need to be there for the whole script. Moved it into `dev` and looks like CI passes =)

alpeb:
> @alpeb realized that this doesn't need to be there for the whole script. Moved it into `dev` and looks like CI passes =)

Oh of course, makes sense :+1: I guess it also needs to go under `run` naturally.

grampelberg:
@alpeb good call, it looks like dev could mostly just be folded into run and DRY it up a little!

PR #2479, opened 2019-03-09
Fix service profiles quoted filename download issue (#2473)

### LINKED ISSUE #2473: dashboard service profiles have quoted filenames
Generating and downloading a ServiceProfile from the `/routes` page in the dashboard yields a filename wrapped in quotes (`'test-profile.yml'` rather than `test-profile.yml`).

```bash
$ ls|grep test-profile
'test-profile.yml'
```
screen shot 2019-03-07 at 4 49 52 pm

liquidslr:
Fixes #2473

Generating and downloading a ServiceProfile from the /routes page in the dashboard yield a filename wrapped in quotes ('test-profile.yml' rather than test-profile.yml). This PR fixes this issue.

cc @siggy

### DIFF
--- a/web/srv/handlers.go
+++ b/web/srv/handlers.go
@@ -78,7 +78,7 @@ func (h *handler) handleProfileDownload(w http.ResponseWriter, req *http.Request
 		return
 	}
 
-	dispositionHeaderVal := fmt.Sprintf("attachment; filename='%s-profile.yml'", service)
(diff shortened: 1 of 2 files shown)


alpeb:
Good fix @liquidslr :star: 

While testing this I realized the popup isn't validating its fields, so I created #2490; that one might be a quick hit, if you're interested in tackling React issues :wink:

liquidslr:
Sure @alpeb would definitely like to work on it after I am finished with other issues that I am currently working on

PR #2486, opened 2019-03-12
Specify a default port for the Linkerd dashboard

scottcarol:
This PR changes the default behavior of `getLocalPort()` in `portforward.go`. Originally, we randomly assigned the port for the dashboard. Now, we specify a default port and check whether it is available. If the default port is unavailable, we resume our original behavior of assigning a random port.

The reason for this change is that the community updates navbar feature in #2476 checks the user's `localStorage` to calculate whether to display an "unread" badge on the dashboard. Since `localStorage` does not persist across ports, each time a user runs the dashboard command, they are on a new port with empty `localStorage`. This would mean the "unread" badge in #2476 would always be displayed, rendering it meaningless.

Note: Although this PR fixes most use cases, if a user is running multiple `linkerd dashboard` instances, they will still have empty `localStorage` on all additional instances after the first.

### DIFF
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -20,6 +20,8 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
+const defaultPort = 50750
+
 // PortForward provides a port-forward connection into a Kubernetes cluster.
 type PortForward struct {
 	method     string
@@ -185,21 +187,17 @@ func (pf *PortForward) URLFor(path string) string {
 // If that port is taken, it binds to a free ephemeral port and returns the
 // port number.
 func getLocalPort() (int, error) {
-	addr := "127.0.0.1:50750"
-	ln, err := net.Listen("tcp", addr)
-	if err != nil {
-		if opError, ok := err.(*net.OpError); ok {
-			if opError.Op == "listen" &&
-				opError.Net == "tcp" &&
-				opError.Addr.String() == addr &&
-				opError.Err.Error() == "bind: address already in use" {
-				ln, err = net.Listen("tcp", ":0")
-			}
-		}
-		if err != nil {
-			return 0, err
-		}
+	defaultAddr := fmt.Sprintf("127.0.0.1:%d", defaultPort)
+	if ln, err := net.Listen("tcp", defaultAddr); err == nil {
+		ln.Close()
+		return defaultPort, nil
 	}
+
+	ln, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return 0, err
+	}
+
 	defer ln.Close()
 
 	// get port
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -180,11 +180,25 @@ func (pf *PortForward) URLFor(path string) string {
 	return fmt.Sprintf("http://127.0.0.1:%d%s", pf.localPort, path)
 }
 
-// getLocalPort binds to a free ephemeral port and returns the port number.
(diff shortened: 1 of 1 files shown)


grampelberg:
We could do something on the linkerd.io side to fix this as well. It feels overly complex though.

scottcarol:
@grampelberg Interesting idea! We could load a cookie from the iFrame, since cookies do persist across ports -- however, it would be classified as a third-party cookie from a remote domain, and most browsers give users the option to block those. Any other methods coming to mind on the linkerd.io side?

I'm inclined to stay with this default port solution for now, and once the "Community" section goes live and we're seeing usage statistics, make another pass at improving it. Thoughts?

grampelberg:
Oh, the iframe is an even better solution than what I was thinking of! Can just use some javascript on each side and do message passing (https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage).

The default port solution is just fine for now, I'm on board with doing something incremental and improving over time.

pkg/k8s/portforward.go
```diff
@@ -180,11 +180,25 @@ func (pf *PortForward) URLFor(path string) string {
...
 }
 
-// getLocalPort binds to a free ephemeral port and returns the port number.
+// getLocalPort is used by dashboard.go to select a port for the dashboard.
+// It first checks the availability of the default port, defined in addr.
+// If that port is taken, it binds to a free ephemeral port and returns the
+// port number.
 func getLocalPort() (int, error) {
-	ln, err := net.Listen("tcp", ":0")
+	addr := "127.0.0.1:60606"
```
siggy: TIOLI: how about port `50750`, which translates to 50L5D in pager code, which translates to 50 Linkerd? #DreamOfThe90s

scottcarol: ✅

pkg/k8s/portforward.go
```diff
@@ -180,11 +180,25 @@ func (pf *PortForward) URLFor(path string) string {
...
-	ln, err := net.Listen("tcp", ":0")
+	addr := "127.0.0.1:50750"
+	ln, err := net.Listen("tcp", addr)
 	if err != nil {
-		return 0, err
+		if opError, ok := err.(*net.OpError); ok {
+			if opError.Op == "listen" &&
+				opError.Net == "tcp" &&
+				opError.Addr.String() == addr &&
+				opError.Err.Error() == "bind: address already in use" {
```
klingerf: Seeing all of these hardcoded strings makes me nervous that this code would break during an upgrade, and I don't think we need this level of error checking. If the initial call to `net.Listen` fails with an error, let's assume that the failure was a result of the port being in use, and always fall back to listening on an ephemeral port. If it wasn't a result of the port being in use, I'd expect the listen call on the ephemeral port to return the same error, at which point we can return that.

It's also the case that if we do successfully bind to 50750, then we already know the port and can return it immediately, instead of needing to read it from the listener on line 206. I recommend setting the function up like this:

```
func getLocalPort() (int, error) {
	defaultAddr := fmt.Sprintf("127.0.0.1:%d", defaultPort)
	if _, err := net.Listen("tcp", defaultAddr); err == nil {
		return defaultPort, nil
	}

	ln, err := net.Listen("tcp", ":0")
	if err != nil {
		return 0, err
	}
	defer ln.Close()
	...
```

That would require you to define a `defaultPort` const at the top of this file, like this:

```
const defaultPort = 50750
```

scottcarol: Sounds good, I just pushed the updated branch, which passes all unit tests.

PR #2495, opened 2019-03-13
Update CHANGES.md for the edge-19.3.2 release

klingerf:
Changelog notes for the edge-19.3.2 release. Am planning to have #2471, #2476, #2494, and linkerd/website#219 merged before starting the release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,20 +2,22 @@
 
 * Controller
   * **Breaking change** Removed support for running the control plane in
-    single-namespace mode, and removed the `--single-namespace` install flag
+    single-namespace mode, which was severely limited in the number of features
+    it supported due to not having access to cluster-wide resources
   * Updated automatic proxy injection and CLI injection to support overriding
     inject defaults via pod spec annotations
   * Added a new public API endpoint for fetching control plane configuration
 * CLI
   * **Breaking change** Removed the `--api-port` flag from the `inject` and
-    `install` commands (thanks, @paranoidaditya)
+    `install` commands, since there's no benefit to running the control plane's
+    destination API on a non-default port (thanks, @paranoidaditya)
   * Introduced the `linkerd metrics` command for fetching proxy metrics
   * Updated the `linkerd routes` command to display rows for routes that are not
     receiving any traffic
   * Updated the `linkerd dashboard` command to serve the dashboard on a fixed
-    port
+    port, allowing it to leverage browser local storage for user settings
 * Web UI
-  * Added a Community page to surface news and updates from linkerd.io
+  * **New** Added a Community page to surface news and updates from linkerd.io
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,26 @@
...
+  * Updated automatic proxy injection and CLI injection to support overriding
+    inject defaults via pod spec annotations
+  * Added a new public API endpoint for fetching control plane configuration
+* CLI
+  * **Breaking change** Removed the `--api-port` flag from the `inject` and
+    `install` commands (thanks, @paranoidaditya)
+  * Introduced the `linkerd metrics` command for fetching proxy metrics
+  * Updated the `linkerd routes` command to display rows for routes that are not
+    receiving any traffic
+  * Updated the `linkerd dashboard` command to serve dashboard on a fixed port
```
alpeb: Should this be "to serve _the_ dashboard"?

CHANGES.md
```diff
@@ -1,3 +1,27 @@
+## edge-19.3.2
+
+* Controller
+  * **Breaking change** Removed support for running the control plane in
```
wmorgan: I would like to add a little explanation here, like "This simplifies things considerably, and instead, you can install Linkerd in a parallel namespace" or something like that?

CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+    single-namespace mode, and removed the `--single-namespace` install flag
+  * Updated automatic proxy injection and CLI injection to support overriding
+    inject defaults via pod spec annotations
+  * Added a new public API endpoint for fetching control plane configuration
+* CLI
+  * **Breaking change** Removed the `--api-port` flag from the `inject` and
```
wmorgan: Also would love to see a brief explanation here

CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+    `install` commands (thanks, @paranoidaditya)
+  * Introduced the `linkerd metrics` command for fetching proxy metrics
+  * Updated the `linkerd routes` command to display rows for routes that are not
+    receiving any traffic
+  * Updated the `linkerd dashboard` command to serve the dashboard on a fixed
```
wmorgan: Is there a user-facing benefit?

CHANGES.md
klingerf: That's sort of the idea. The main benefit here is that single-namespace mode was feature poor to the point that it wasn't useful, and there was a large overhead to maintaining it, so we removed it. That said, if a user required single-namespace mode, we don't have another alternative to offer them at this point. Will see if I can wordsmith something like that.

klingerf: How about?

```
  * **Breaking change** Removed support for running the control plane in
    single-namespace mode, which was severely limited in the number of features
    it supported due to not having access to cluster-wide resources
```

wmorgan: LGTM

PR #2500, opened 2019-03-14
Handle configs override for meshed workloads

ihcsim:
This PR is a different take to handle configs override while updating meshed workloads. 

### Assumptions
The proxy injector is only interested in ~two~ three types of workloads:

* unmeshed workload with the `linkerd.io/inject: enabled` annotation
* meshed workload with the `config.linkerd.io/*` annotations
* NEW: unmeshed workload with the `linkerd.io/inject: enabled` and `config.linkerd.io/*` annotations

Everything else should be a passthrough.

### Implementations
* Introduces a check to verify if a workload should be injected and/or config-overridden, OR if it just needs to be config-overridden. If neither, the rest of the injection process will be skipped.
* Separated out the logic to generate the different patches for unmeshed and meshed workloads.

PS The whole thing probably doesn't work yet and the unit tests are failing. Just thought I will share it to get some early feedback.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -181,8 +181,6 @@ func (conf *ResourceConfig) AddRootLabels(patch *Patch) {
 
 // GetPatch returns the JSON patch containing the proxy and init containers specs, if any
 func (conf *ResourceConfig) GetPatch() (*Patch, error) {
-	log.Infof("received %s", conf)
-
 	// If we don't inject anything into the pod template then output the
 	// original serialization of the original object. Otherwise, output the
 	// serialization of the modified object.
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -162,7 +162,7 @@ func (conf *ResourceConfig) GetPatch(
 	shouldInject func(*ResourceConfig, Report) bool,
 ) (*Patch, []Report, error) {
 	report := newReport(conf)
-	log.Infof("received %s/%s", strings.ToLower(conf.meta.Kind), report.Name)
+	log.Infof("received %s", conf)
 
 	if err := conf.parse(bytes); err != nil {
 		return nil, nil, err
@@ -189,6 +189,7 @@ func (conf *ResourceConfig) GetPatch(
 
 		report.update(conf)
 		if !shouldInject(conf, report) && !shouldOverrideConfig(conf) {
+			log.Infof("skipping %s", conf)
 			return &Patch{}, []Report{report}, nil
 		}
 
@@ -517,7 +518,7 @@ func (conf *ResourceConfig) setProxyInitConfigs() *v1.Container {
 	initContainer.ImagePullPolicy = conf.proxyImagePullPolicy()
 	initContainer.Args = conf.proxyInitArgs()
 
-	log.Debugf("after overrides (%s): %+v\n", k8s.ProxyContainerName, initContainer)
+	log.Debugf("after overrides (%s): %+v\n", k8s.InitContainerName, initContainer)
 	return initContainer
 }
 
@@ -835,6 +836,10 @@ func ShouldInjectCLI(_ *ResourceConfig, r Report) bool {
 // - the workload's pod spec has the linkerd.io/inject annotation set to "enabled"
 func ShouldInjectWebhook(conf *ResourceConfig, r Report) bool {
 	if !r.Injectable() {
+		return false
+	}
+
+	if createdBy, exists := conf.podMeta.Annotations[k8s.CreatedByAnnotation]; exists && strings.Contains(createdBy, k8s.CreatedByCLI) {
 		return false
 	}
 
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -59,7 +59,9 @@ var (
 		k8s.StatefulSet,
 	}
 
-	ErrUnsupportedResourceType = fmt.Errorf("Unsupported resource type")
+	// ErrUnsupportedResourceKind is the error used to indicate an supported
+	// resource kind in the input YAML.
+	ErrUnsupportedResourceKind = fmt.Errorf("Unsupported resource kind")
 )
 
 // objMeta provides a generic struct to parse the names of Kubernetes objects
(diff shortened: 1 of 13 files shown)


pkg/inject/inject.go
```diff
@@ -148,6 +149,13 @@ func (conf *ResourceConfig) ParseMeta(bytes []byte) (bool, error) {
 	return conf.podMeta.ObjectMeta != nil, nil
 }
 
+// AddRootLabels adds all the pod labels into the root workload (e.g. Deployment)
+func (conf *ResourceConfig) AddRootLabels(patch *Patch) {
```
ihcsim: Just grouping all the exported methods together.

pkg/inject/inject.go
```diff
@@ -187,15 +188,41 @@ func (conf *ResourceConfig) GetPatch(
 		}
 
 		report.update(conf)
+		if !shouldInject(conf, report) && !shouldOverrideConfig(conf) {
+			return &Patch{}, []Report{report}, nil
+		}
```
alpeb: +1 one for the early returns

pkg/inject/inject.go
```diff
@@ -187,15 +188,41 @@ func (conf *ResourceConfig) GetPatch(
...
+
+		// populate the proxy's configurable properties with either overrides or
+		// defaults
+		proxy := conf.setProxyConfigs(identity)
```
alpeb: Say `shouldInject()` returned false because `HostNetwork` is true, yet the user added config annotations unknowingly. This will inject the proxy when it shouldn't. I think `shouldInject()` needs to be split into different methods to be able to distinguish  when pods should not be injected at all, and when they are already injected.

alpeb: This might also help clarifying a bit the logic to decide when to inject/override inside `GetPatch()`, which is currently not obvious IMO.

pkg/inject/proxy_patch.go
```diff
@@ -0,0 +1,146 @@
...
+
+func newProxyPatch(proxy *v1.Container, identity k8s.TLSIdentity, config *ResourceConfig) *Patch {
+	patch := patchKind(config.meta.Kind)
+
+	// Special case if the caller specifies that
+	// LINKERD2_PROXY_OUTBOUND_ROUTER_CAPACITY be set on the pod.
+	// We key off of any container image in the pod. Ideally we would instead key
+	// off of something at the top-level of the PodSpec, but there is nothing
+	// easily identifiable at that level.
+	// Currently this will bet set on any proxy that gets injected into a Prometheus pod,
```
scottcarol: tiny nit: should be `this will be set`

pkg/inject/inject.go
ihcsim: Yes, definitely a bug in the `if` statement. As a matter of fact, it breaks `linkerd inject`, because the received YAML will already have a proxy spec, and if config annotations are present, the bug you described will occur, leading to some downstream complication in the YAML (i.e. the CLI calls `GetPatch()` once, and then the webhook calls it again when the deployment is created.) :boom:

I'm not sure how splitting up the `shouldInject()` will help, as I imagine it will require some clever ways to chain these different methods together, to satisfy both the webhook and CLI code path.

What I ended up doing was to perform the `shouldInject()` checks earlier in both the CLI and webhhook. So if `shouldInject()` failed, we won't even call the `GetPatch()` function. This also helps to keep `GetPatch()` to focus only on generating JSON patches, and the decision on whether the configs need to be overridden.

The commit is at dde345f75aeeff75712406c39804e0c53705efb7 This seems to work for the webhook, but I'm now battling with how the CLI reports unit tests. Maybe you can take a look at that commit to see if it makes sense? Thanks!

alpeb: > What I ended up doing was to perform the shouldInject() checks earlier in both the CLI and webhhook. So if shouldInject() failed, we won't even call the GetPatch() function. This also helps to keep GetPatch() to focus only on generating JSON patches, and the decision on whether the configs need to be overridden.

Yes, this looks much better to me :+1: 
Note though that your `shouldInject()` functions will return false whenever there's already a proxy attached because `healthcheck.HasExistingSidecars()` doesn't distinguish between linkerd proxies and 3rd party proxies, so this will prevent config overriding. That should be an easy fix :wink:

pkg/inject/report.go
```diff
@@ -47,10 +54,10 @@ func (r Report) Injectable() bool {
...
 func (r *Report) update(conf *ResourceConfig) {
-	r.InjectDisabled = conf.podMeta.ObjectMeta.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled
-	r.HostNetwork = conf.podSpec.HostNetwork
-	r.Sidecar = healthcheck.HasExistingSidecars(conf.podSpec)
-	r.UDP = checkUDPPorts(conf.podSpec)
+	r.InjectDisabled = conf.pod.meta.ObjectMeta.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled
+	r.HostNetwork = conf.pod.spec.HostNetwork
+	r.Sidecar = healthcheck.HasExistingSidecars(conf.pod.spec)
+	r.UDP = checkUDPPorts(conf.pod.spec)
 }
```
klingerf: This func now duplicates a bunch of code that exists in pkg/inject/inject.go. Can you instead write it like:

```
func (r *Report) update(conf *ResourceConfig) {
	r.InjectDisabled = conf.InjectDisabled()
	r.HostNetwork = conf.PodUsingHostNetwork()
	r.Sidecar = conf.HasExistingProxy()
	r.UDP = checkUDPPorts(conf.pod.spec)
}
```

pkg/k8s/labels.go
```diff
@@ -61,6 +61,14 @@ const (
 	// (e.g. linkerd/cli v2.0.0).
 	CreatedByAnnotation = Prefix + "/created-by"
 
+	// CreatedByCLI is used by the CreatedByAnnotation to indicate that the
+	// proxy is injected by the CLI.
+	CreatedByCLI = "linkerd/cli"
```
klingerf: This is unused?

pkg/inject/inject.go
```diff
@@ -814,35 +821,58 @@ func (conf *ResourceConfig) proxyOutboundSkipPorts() string {
...
+
+// HasExistingProxy returns true if the pod already has a proxy sidecar.
+func (conf *ResourceConfig) HasExistingProxy() bool {
+	return conf.pod.spec != nil && healthcheck.HasExistingSidecars(conf.pod.spec)
+}
+
+// ShouldOverrideConfig returns true if the workload has existing sidecars and
+// config override annotations.
+func (conf *ResourceConfig) ShouldOverrideConfig() bool {
+	return healthcheck.HasExistingSidecars(conf.pod.spec) && hasOverrideAnnotations(conf.pod.meta)
```
klingerf: Rather than only doing the override if at least `config.linkerd.io` annotation is present, I think it would better to use something like `k8s.IsMeshed` to determine if the pod is inject, and then always override if inject, even if it doesn't contain override annotations. My thinking is that eventually we might allow users to change the contents of the proxy configmap after install, and we won't pick up the change if we're only looking for override annotations.

ihcsim:
Here's a summary of the discussion with @klingerf @alpeb:

* the `ShouldInject` checks will remain separated in the CLI and webhook, as seen in this PR
* the `ParseMetaAndYaml` will be used to generate the `report` early before any proxy injection, as seen in this PR
* the webhook's `ShouldInject` will be updated to check only for the presence of 3rd party proxies, implying that any YAML manifests with existing Linkerd proxy will be "re-injected" to support config override
* the `report` struct will be extended to capture the value of any `linkerd.io/inject: disabled`. This struct will be used to decide whether a proxy injection should happen
* put this PR on-hold, and proceed with #2447, so that when the webhook intercepts YAML manifests submitted by the CLI, the proxy will be "re-injected" with the correct values from the config annotations.

ihcsim:
See #2590.

PR #2501, opened 2019-03-14
Replacing native sort with lodash stable sortBy for table rows

### LINKED ISSUE #2413: Implement stable sort for web UI
Currently we are using the native JavaScript `sort()` method to sort columns in our web UI. This leads to unexpected behavior when a table is re-rendered, where order of the rows may not be preserved in `BaseTable` despite no change to the `orderBy` or `order` values.

Proposed solution: 

Switch from native `sort` to Lodash `_sortBy` which is a stable sort. Rewrite the `sorter()` methods in `MetricsTable` and other corresponding components, so that instead of returning a `compareFunction` they simply return the value to sort by.

Example:

```sorter: d => d.name```

Then in the `BaseTable` function use:

```rows = _sortBy(tableRows, [row => row.sort()]]);```

This would be a lot of line changes to `MetricsTable` and the other table components but I think it would switch everything to stable sorting.

@rmars @siggy thoughts welcome!

Screenshots:

Initial view

prechange

After state change (clicking the filter button, no change to data)

postchange

### PR DESCRIPTION
**Pull request by scottcarol**
Fixes #2413.

Currently we are using the native JavaScript `sort()` method to sort columns in our web UI. This leads to unexpected behavior when a table is re-rendered, where order of the rows may not be preserved despite no change to the `orderBy` or values. lodash's `_sortBy` is stable so every time you sort by a specific type, you'll get the same result.

**To Test Current Behavior Locally**

While running BooksApp, open the dashboard and visit path 
```
/top?namespace=default&resource=namespace%2Fdefault
```
Start, then stop, the Top query. Click on the filter icon, which triggers a state update. You will likely see the order of the rows change due to unstable `sort()` being used on the default `orderBy` value for this view, `count`. (If you don't see it immediately, start and stop the query a few more times - the unstable behavior means the order is not predictable.)

**Note**

In testing my changes, I saw that some of our tests are passing in an `undefined` into the `BaseTable` component for `col.sorter` (line 89).

The original JS native `sort()` ignores a `compareFunction` that is `undefined`, so the tests were not failing. However, lodash's `_sortBy` runs `col.sorter()` on each row so the `undefined` was causing the tests to fail.

To remedy this, I include a check for `col.sorter` on line 89 of `BaseTable` and am opening a separate issue for fixing the `undefined` data passed into the tests.

### DIFF
--- a/web/app/js/components/MetricsTable.jsx
+++ b/web/app/js/components/MetricsTable.jsx
@@ -152,7 +152,7 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink, isTcpTab
         </Grid>
       );
     },
-    sorter: d => getResourceDisplayName(d) ? getResourceDisplayName(d) : -1
+    sorter: d => getResourceDisplayName(d) || -1
   };
 
   let columns = [nameColumn];
--- a/web/app/js/components/MetricsTable.jsx
+++ b/web/app/js/components/MetricsTable.jsx
@@ -1,4 +1,4 @@
-import { displayName, friendlyTitle, metricToFormatter, numericSort } from './util/Utils.js';
+import { displayName, friendlyTitle, metricToFormatter } from './util/Utils.js';
 import BaseTable from './BaseTable.jsx';
 import ErrorModal from './ErrorModal.jsx';
 import GrafanaLink from './GrafanaLink.jsx';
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: web/app/js/components/MetricsTable.jsx [hunk 1]
```diff
@@ -44,44 +44,42 @@ const httpStatColumns = [
...
   },
   {
     title: "TLS",
     dataIndex: "tlsRequestPercent",
     isNumeric: true,
     render: d => _isNil(d.tlsRequestPercent) || d.tlsRequestPercent.get() === -1 ? "---" : d.tlsRequestPercent.prettyRate(),
-    sorter: (a, b) => numericSort(
-      a.tlsRequestPercent ? a.tlsRequestPercent.get() : -1,
-      b.tlsRequestPercent ? b.tlsRequestPercent.get() : -1)
+    sorter: d => d.tlsRequestPercent ? d.tlsRequestPercent.get() : -1
```
dadjeibaah: TIOLI: This statement is equivalent as well d.tlsRequestPercent || -1

scottcarol:
@siggy Thanks for the spot on the filter order -- just realized there's a bug in `BaseTable` that resets the ordering of the rows if you are filtering - the below `tableRows` should be `rows`. I can go ahead and fix this in this branch since everything is related to row order.

```      
let filteredRows = tableRows.filter(row => {
```

web/app/js/components/MetricsTable.jsx
rmars: I don't think this is true, since `tlsRequestPercent` is a `Percentage` so to get the value out of it, you have to call `get()`

web/app/js/components/MetricsTable.jsx
```diff
@@ -154,7 +152,7 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink, isTcpTab
         </Grid>
       );
     },
-    sorter: (a, b) => (getResourceDisplayName(a) || "").localeCompare(getResourceDisplayName(b))
+    sorter: d => getResourceDisplayName(d) ? getResourceDisplayName(d) : -1
```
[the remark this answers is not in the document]
scottcarol: Good point @dadjeibaah

scottcarol:
@rmars That sort/filter bug should be fixed, would you mind trying my updated branch locally to doublecheck?

PR #2507, opened 2019-03-15
Adding text wrap to expanded Tap request

### LINKED ISSUE #1932: Dashboard: tap details, path should wrap
### What is the issue?
Can't see the full path, and it doesn't scroll

### How can it be reproduced?
Rap details page with emojivoto
screen shot 2018-12-05 at 11 07 07 am

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version:
- Host OS:
- Linkerd version:

scottcarol:
Fixes #1932.

This PR adds text wrap to the tap details page so the full request path can be seen.

### DIFF
--- a/web/app/js/components/ExpandableTable.jsx
+++ b/web/app/js/components/ExpandableTable.jsx
@@ -21,6 +21,9 @@ const styles = theme => ({
     marginTop: theme.spacing.unit * 3,
     overflowX: 'auto',
   },
+  expandedWrap: {
+    wordBreak: `break-word`
+  },
   table: {
     minWidth: 700
   },
@@ -107,7 +110,7 @@ class ExpandableTable extends React.Component {
           aria-labelledby="form-dialog-title">
           <DialogTitle id="form-dialog-title">Request Details</DialogTitle>
           <DialogContent>
-            {expandedRowRender(this.state.datum)}
+            {expandedRowRender(this.state.datum, classes.expandedWrap)}
           </DialogContent>
           <DialogActions>
             <Button onClick={this.handleDialogClose} color="primary">Close</Button>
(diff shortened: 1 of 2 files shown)


web/app/js/components/ExpandableTable.jsx
```diff
@@ -21,6 +21,9 @@ const styles = theme => ({
     marginTop: theme.spacing.unit * 3,
     overflowX: 'auto',
   },
+  expandedWrap: {
+    wordBreak: `break-word`
```
siggy: Mostly for my own understanding... I noticed that `break-word` is supposedly not compatible with Firefox and IE:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#Browser_compatibility

...but I guess through the magic of, react?, it removes that property and correctly breaks the word in Firefox:
firefox

... and funny enough, with Chrome supporting `break-word`, it does a slightly worse job of breaking the string:
chrome

scottcarol: @siggy, great thoughts/questions - my understanding is that typically, Material-UI [adds its own magic to unsupported browser features](https://material-ui.com/getting-started/supported-platforms/), but in this case, the word-wrap for this specific scenario was only broken in Chrome (not Firefox) so no extra CSS was needed to make it FF-compatible. (Pssst, automated cross-browser testing is in our future which will help with these comparisons. 🔮🤞) 

BUT, in digging into the CSS to figure out how Material-UI was handling Firefox, @rmars and I spotted some other layout problems and saw some opportunities to improve this modal - so thanks for the questions! Merging this PR now and reworking the modal in a separate PR.

PR #2510, opened 2019-03-17
Add check for unschedulable pods and psp issues (#2465)

### LINKED ISSUE #2465: Add check for unschedulable pods

On clusters, it is possible for the control plane to be unschedulable. This could be because there aren't enough resources in the cluster or there is a PSP that does not allow the pods to launch.

There should be two checks added:

- Control plane pods are in pending and have events such as `no nodes available`.
- The top level resources were created but no pods have been yet, and have events blocking them (PSP issues).

Instead of showing the errors in `check`, for both of these links to the troubleshooting docs should be provided with commands that explain how to debug scheduling issues.

liquidslr:
Fixes #2465

On clusters, it is possible for the control plane to be unschedulable due to unavailability or deletion of nodes. Also it also possible for the pods to be unable to start due to pod security policies. This PR adds checks for the above two conditions.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -964,8 +964,8 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
 func checkUnschedulablePods(pods []corev1.Pod) error {
 	var podName []string
 	for _, pod := range pods {
-		for _, node := range pod.Status.Conditions {
-			if node.Reason == "Unschedulable" {
+		for _, condition := range pod.Status.Conditions {
+			if condition.Reason == "Unschedulable" {
 				podName = append(podName, pod.Name)
 			}
 		}
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -360,7 +360,7 @@ func (hc *HealthChecker) allCategories() []category {
 					check: func(ctx context.Context) error {
 						var err error
 						var controlPlaneReplicaSet []v1beta1.ReplicaSet
-						controlPlaneReplicaSet, err = hc.kubeAPI.GetReplicaSets(ctx, hc.httpClient, hc.ControlPlaneNamespace)
+						controlPlaneReplicaSet, err = hc.kubeAPI.GetReplicaSetByNamespace(ctx, hc.httpClient, hc.ControlPlaneNamespace)
 						if err != nil {
 							return err
 						}
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -972,7 +972,7 @@ func checkUnschedulablePods(pods []corev1.Pod) error {
 	}
 
 	if len(errors) > 0 {
-		return fmt.Errorf("%s\n", strings.Join(errors, "\n    "))
+		return fmt.Errorf("%s", strings.Join(errors, "\n    "))
 	}
 
 	return nil
@@ -989,7 +989,7 @@ func checkControlPlaneReplicaSets(rst []v1beta1.ReplicaSet) error {
 	}
 
 	if len(errors) > 0 {
-		return fmt.Errorf("%s\n", strings.Join(errors, "\n   "))
+		return fmt.Errorf("%s", strings.Join(errors, "\n   "))
 	}
 
 	return nil
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -17,6 +17,7 @@ import (
 	"github.com/linkerd/linkerd2/pkg/version"
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	v1beta1 "k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sVersion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
@@ -163,14 +164,15 @@ type HealthChecker struct {
 	*Options
 
 	// these fields are set in the process of running checks
-	kubeAPI          *k8s.KubernetesAPI
-	httpClient       *http.Client
-	clientset        kubernetes.Interface
-	kubeVersion      *k8sVersion.Info
-	controlPlanePods []corev1.Pod
-	apiClient        public.APIClient
-	latestVersions   version.Channels
-	serverVersion    string
+	kubeAPI                *k8s.KubernetesAPI
+	httpClient             *http.Client
+	clientset              kubernetes.Interface
+	kubeVersion            *k8sVersion.Info
+	controlPlanePods       []corev1.Pod
+	controlPlaneReplicaSet []v1beta1.ReplicaSet
+	apiClient              public.APIClient
+	latestVersions         version.Channels
+	serverVersion          string
 }
 
 // NewHealthChecker returns an initialized HealthChecker
@@ -350,6 +352,32 @@ func (hc *HealthChecker) allCategories() []category {
 					fatal:       true,
 					check: func(ctx context.Context) error {
 						return hc.checkNamespace(ctx, hc.ControlPlaneNamespace, true)
+					},
+				},
+				{
+					description: "correct pod secuirty policies",
+					hintAnchor:  "l5d-existence-psp",
+					fatal:       true,
+					check: func(ctx context.Context) error {
+						var err error
+						hc.controlPlaneReplicaSet, err = hc.kubeAPI.GetReplicaSetByNamespace(ctx, hc.httpClient, hc.ControlPlaneNamespace)
+						if err != nil {
+							return err
+						}
+						return checkPodSecurityPolicies(hc.controlPlaneReplicaSet)
+					},
+				},
+				{
+					description: "no unschedulable pods",
+					hintAnchor:  "l5d-existence-unschedulable-pods",
+					fatal:       true,
+					check: func(ctx context.Context) error {
+						var err error
+						hc.controlPlanePods, err = hc.kubeAPI.GetPodsByNamespace(ctx, hc.httpClient, hc.ControlPlaneNamespace)
+						if err != nil {
+							return err
+						}
+						return checkUnschedulablePods(hc.controlPlanePods)
 					},
 				},
 				{
@@ -835,10 +863,12 @@ func getPodStatuses(pods []corev1.Pod) map[string][]corev1.ContainerStatus {
 	statuses := make(map[string][]corev1.ContainerStatus)
 
 	for _, pod := range pods {
+
 		if pod.Status.Phase == corev1.PodRunning && strings.HasPrefix(pod.Name, "linkerd-") {
 			parts := strings.Split(pod.Name, "-")
 			// All control plane pods should have a name that results in at least 4
 			// substrings when string.Split on '-'
+
 			if len(parts) >= 4 {
 				name := strings.Join(parts[1:len(parts)-2], "-")
 				if _, found := statuses[name]; !found {
@@ -928,6 +958,29 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
 
 	if errMsg != "" {
 		return fmt.Errorf(errMsg)
+	}
+
+	return nil
+}
+
+func checkUnschedulablePods(pods []corev1.Pod) error {
+	for _, pod := range pods {
+		for _, node := range pod.Status.Conditions {
+			if node.Status == "False" && node.Message == "no nodes available to schedule pods" {
+				return fmt.Errorf("No node available for the pod %s", pod.Name)
+			}
+		}
+	}
+	return nil
+}
+
+func checkPodSecurityPolicies(rst []v1beta1.ReplicaSet) error {
+	for _, rs := range rst {
+		for _, r := range rs.Status.Conditions {
+			if strings.Contains(r.Message, "unable to validate against any pod security policy") {
+				return fmt.Errorf("Invalid pod secuirty policy for the replicaset %s", rs.Name)
+			}
+		}
 	}
 
 	return nil
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -354,7 +354,7 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
-					description: "correct pod security policies",
+					description: "control plane components ready",
 					hintAnchor:  "l5d-existence-psp",
 					fatal:       true,
 					check: func(ctx context.Context) error {
@@ -962,35 +962,34 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
 }
 
 func checkUnschedulablePods(pods []corev1.Pod) error {
-	var podName []string
+	var errors []string
 	for _, pod := range pods {
 		for _, condition := range pod.Status.Conditions {
 			if condition.Reason == corev1.PodReasonUnschedulable {
-				podName = append(podName, pod.Name)
+				errors = append(errors, fmt.Sprintf("%s: %s", pod.Name, condition.Message))
 			}
 		}
 	}
 
-	if len(podName) != 0 {
-		pods := strings.Join(podName, ",")
-		return fmt.Errorf("Unable to schedule pods: %s", pods)
+	if len(errors) > 0 {
+		return fmt.Errorf("%s\n", strings.Join(errors, "\n    "))
 	}
 
 	return nil
 }
 
 func checkControlPlaneReplicaSets(rst []v1beta1.ReplicaSet) error {
-	var errors string
+	var errors []string
 	for _, rs := range rst {
 		for _, r := range rs.Status.Conditions {
 			if r.Type == v1beta1.ReplicaSetReplicaFailure && r.Status == corev1.ConditionTrue {
-				errors += fmt.Sprintf("%s: %s, %s\n", rs.Name, r.Reason, r.Message)
+				errors = append(errors, fmt.Sprintf("%s: %s", r.Reason, r.Message))
 			}
 		}
 	}
 
-	if errors != "" {
-		return fmt.Errorf("errors in creating control plane replicas %s ", errors)
+	if len(errors) > 0 {
+		return fmt.Errorf("%s\n", strings.Join(errors, "\n   "))
 	}
 
 	return nil
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -164,15 +164,14 @@ type HealthChecker struct {
 	*Options
 
 	// these fields are set in the process of running checks
-	kubeAPI                *k8s.KubernetesAPI
-	httpClient             *http.Client
-	clientset              kubernetes.Interface
-	kubeVersion            *k8sVersion.Info
-	controlPlanePods       []corev1.Pod
-	controlPlaneReplicaSet []v1beta1.ReplicaSet
-	apiClient              public.APIClient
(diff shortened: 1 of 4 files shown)


liquidslr:
I will be shortly making a pr in the linkerd website to update with the solutions of the above two checks, till the time @grampelberg  can you pleas review this.

liquidslr:
@alpeb here's what did to produce the psp  issue
 
 1. ssh in to the vm/minikube
2. `sudo vi /etc/kubernetes/manifests/kube-apiserver.yaml`
3. `    - --enable-admission-plugins=PodSecuirtyPolicy,...`
4.  Save and exit vm/minikube and add  [PSP](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example) in local kuberentes   and [RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/)

pkg/k8s/api.go
```diff
@@ -149,3 +151,34 @@ func NewAPI(configPath, kubeContext string) (*KubernetesAPI, error) {
 
 	return &KubernetesAPI{Config: config}, nil
 }
+
+// GetReplicaSetByNamespace returns all replicasets in a given namespace
+func (kubeAPI *KubernetesAPI) GetReplicaSetByNamespace(ctx context.Context, client *http.Client, namespace string) ([]v1beta1.ReplicaSet, error) {
+	return kubeAPI.getReplicaSets(ctx, client, "/apis/extensions/v1beta1/namespaces/"+namespace+"/replicasets")
+}
```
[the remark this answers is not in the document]
liquidslr: yeah but if I remove this function then I will have to hard code the namespace which is linkerd in this case, will that be fine ?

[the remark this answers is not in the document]
liquidslr: @alpeb i get your point but if i remove this function i will have to harcode the namespace name, will that be fine ?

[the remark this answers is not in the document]
liquidslr: ok understood i passed in the namespace and created the path inside 👍

liquidslr:
@alpeb I have updated with the changes suggested please review now.

liquidslr:
@alpeb can you help me figure out why Travis is failing, I am unable to figure out

siggy:
@liquidslr looks like `GetReplicaSets` is not a thing, maybe you want `GetReplicaSetByNamespace`? ...
```bash
$ go test ./...
# github.com/linkerd/linkerd2/pkg/healthcheck
pkg/healthcheck/healthcheck.go:363:47: hc.kubeAPI.GetReplicaSets undefined (type *"github.com/linkerd/linkerd2/pkg/k8s".KubernetesAPI has no field or method GetReplicaSets)
FAIL	github.com/linkerd/linkerd2/cli/cmd [build failed]
```

pkg/k8s/api.go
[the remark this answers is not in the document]
alpeb: @liquidslr It appears both functions are still there... maybe the refactor got lost during rebasing?
My idea was to refactor it something like this:
```diff
diff --git a/pkg/k8s/api.go b/pkg/k8s/api.go
index a9983801..9acecabf 100644
--- a/pkg/k8s/api.go
+++ b/pkg/k8s/api.go
@@ -151,12 +151,9 @@ func NewAPI(configPath, kubeContext string) (*KubernetesAPI, error) {
        return &KubernetesAPI{Config: config}, nil
 }
 
-// GetReplicaSetByNamespace returns all replicasets in a given namespace
-func (kubeAPI *KubernetesAPI) GetReplicaSetByNamespace(ctx context.Context, client *http.Client, namespace string) ([]v1beta1.ReplicaSet, error) {
-       return kubeAPI.getReplicaSets(ctx, client, "/apis/extensions/v1beta1/namespaces/"+namespace+"/replicasets")
-}
-
-func (kubeAPI *KubernetesAPI) getReplicaSets(ctx context.Context, client *http.Client, path string) ([]v1beta1.ReplicaSet, error) {
+// GetReplicaSetsByNamespace returns all replicasets in a given namespace
+func (kubeAPI *KubernetesAPI) GetReplicaSetsByNamespace(ctx context.Context, client *http.Client, namespace string) ([]v1beta1.ReplicaSet, error) {
+       path := "/apis/extensions/v1beta1/namespaces/" + namespace + "/replicasets"
        rsp, err := kubeAPI.getRequest(ctx, client, path)
        if err != nil {
                return nil, err
```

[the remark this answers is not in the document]
liquidslr: I did the same thing but this resulted in travis to fail (couldn't figure out why) so I reverted it  back to the previous state

liquidslr:
@alpeb please review now

alpeb:
@liquidslr thank you very much for addressing my feedback and the Minikube tip! Apologies for not having been able to fully review this PR, as we've been busy preparing version 2.3... I should provide you with some feedback again in a few days.

liquidslr:
sure will do that, I created the issue of unschedulable pods by deleting the node, will handle the other case as well, also it will be great if you could tell how to inc the resource CPU request value for some pod

alpeb:
For testing resource requests, you could for example edit this line:
https://github.com/linkerd/linkerd2/blob/4562ed0660db191cf6f978d736e43addae8bdde6/chart/templates/controller.yaml#L128
changing it from `20m` to `20`. Then you install linkerd with `bin/go-run cli install --ha | k apply -f -` and the controller will ask for 20 cores, which is gonna fail :fire:

liquidslr:
@alpeb i tried to incorporate both cases of unschedulable-pods  arising due to node deletion and other one due to Insufficient cpu. Please have a look :eyes:

pkg/healthcheck/healthcheck.go
```diff
@@ -962,24 +962,41 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
...
 			if strings.Contains(r.Message, "unable to validate against any pod security policy") {
-				return fmt.Errorf("Invalid pod security policy for the replicaset %s", rs.Name)
+				replicaSets = append(replicaSets, rs.Name)
 			}
 		}
 	}
 
+	if len(replicaSets) != 0 {
+		for _, name := range replicaSets {
+			fmt.Println(name)
```
alpeb: Please build the list of replicasets names as a comma-separated list to be included in the same error message produced below on line 998

alpeb: (same for `checkUnschedulablePods` above)

pkg/healthcheck/healthcheck.go
```diff
@@ -962,24 +962,41 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
 }
 
 func checkUnschedulablePods(pods []corev1.Pod) error {
+	var podName []string
 	for _, pod := range pods {
 		for _, node := range pod.Status.Conditions {
-			if node.Status == "False" && node.Message == "no nodes available to schedule pods" {
-				return fmt.Errorf("No node available for the pod %s", pod.Name)
+			if strings.Contains(node.Message, "Insufficient cpu") || (node.Status == "False" && node.Message == "no nodes available to schedule pods") {
```
alpeb: There are other possible failures, like insufficient memory. I believe it's better to just check for `Reason == "FailedSheduling"` instead of checking for the error message strings, so we can cover any possible failure.

liquidslr: i am not  getting reason as Reason == "FailedSheduling" instead i am getting node.Reason == "ContainersNotReady", should i check for it

alpeb: Could you please clarify how can I reproduce that? Also it would be helpful to get the output of `kubectl -n linkerd get events` right after the install fails.

liquidslr: i was able to check for the reason in the two cases it comes out to be the same i.e.  node.Reason == "Unschedulable"

liquidslr: > Also it would be helpful to get the output of `kubectl -n linkerd get events` right after the install fails.

as if the output of `kubectl -n linkerd get events` after the check ?

alpeb: Reason `Unschedulable` is actually fine. One last minor change: could you rename the `node` var in that loop to something like `condition`?

liquidslr: something  like `isNodeAvailable`  ?

liquidslr: or `nodeAvailability`

liquidslr: sorry i misread

pkg/healthcheck/healthcheck.go
```diff
@@ -932,3 +960,39 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
...
+		for _, condition := range pod.Status.Conditions {
+			if condition.Reason == "Unschedulable" {
+				podName = append(podName, pod.Name)
+			}
+		}
+	}
+
+	if len(podName) != 0 {
+		pods := strings.Join(podName, ",")
+		return fmt.Errorf("No node available for the pods %s", pods)
```
grampelberg: Can we change this to `Unable to schedule pods: %s`?

pkg/healthcheck/healthcheck.go
```diff
@@ -932,3 +960,39 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
...
+		for _, r := range rs.Status.Conditions {
+			if strings.Contains(r.Message, "unable to validate against any pod security policy") {
+				replicaSets = append(replicaSets, rs.Name)
+			}
+		}
+	}
+
+	if len(replicaSets) != 0 {
+		replicasets := strings.Join(replicaSets, ",")
+		return fmt.Errorf("Invalid pod security policy for the replicasets %s", replicasets)
```
grampelberg: How about `Pod security policy does not allow creation of pods: %s`?

liquidslr: Something like `Pod security policy does not allow creation of pods for the repliasets %s` ?

pkg/healthcheck/healthcheck.go
liquidslr: sure, meanwhile i created an issue[https://github.com/linkerd/website/issues/272](url) on the linkerd website, will make a pr soon, just needed some help in what solutions should i write.

pkg/healthcheck/healthcheck.go
```diff
@@ -932,3 +960,39 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
 
 	return nil
 }
+
+func checkUnschedulablePods(pods []corev1.Pod) error {
+	var podName []string
+	for _, pod := range pods {
+		for _, condition := range pod.Status.Conditions {
+			if condition.Reason == "Unschedulable" {
```
ihcsim: Any reasons why we don't check the pod [phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) instead? I'd think phases like [`Failed`, `Unknown`, `CrashLoopBackOff`](https://godoc.org/k8s.io/api/core/v1#PodPhase) etc. all imply that the pod isn't scheduled.

pkg/healthcheck/healthcheck.go
```diff
@@ -932,3 +960,39 @@ func validateDataPlanePodReporting(pods []*pb.Pod) error {
...
+
+	if len(podName) != 0 {
+		pods := strings.Join(podName, ",")
+		return fmt.Errorf("No node available for the pods %s", pods)
+	}
+
+	return nil
+}
+
+func checkPodSecurityPolicies(rst []v1beta1.ReplicaSet) error {
```
ihcsim: How about instead of explicitly single out psp error, we only have the `checkUnschedulablePods()` above, where we check for the pod's phase. If the phase isn't one of `succeeded`, `completed`, `running` or `pending`, then we output the status, conditions and reasons?

pkg/healthcheck/healthcheck.go
```diff
@@ -352,6 +353,33 @@ func (hc *HealthChecker) allCategories() []category {
...
 					},
 				},
+				{
+					description: "correct pod security policies",
+					hintAnchor:  "l5d-existence-psp",
+					fatal:       true,
+					check: func(ctx context.Context) error {
+						var err error
+						var controlPlaneReplicaSet []v1beta1.ReplicaSet
+						controlPlaneReplicaSet, err = hc.kubeAPI.GetReplicaSets(ctx, hc.httpClient, hc.ControlPlaneNamespace)
```
ihcsim: Any reasons why we want to use replica sets, not just pods?

ihcsim: With replica set, it will be useful to compare (desired) [`Replicas`](https://godoc.org/k8s.io/api/apps/v1#ReplicaSetSpec) with [`ReadyReplicas`](https://godoc.org/k8s.io/api/apps/v1#ReplicaSetStatus).

alpeb: @ihcsim PSP can forbid a pod to be created, so we can only check the failure in the Replicaset.

pkg/healthcheck/healthcheck.go
alpeb: For the reason stated in my comment below, we need special handling for PSP blockages that looks into Replicasets.

alpeb: .. but I agree `checkUnschedulablePods` could be more generic and return the actual error string, not a hard-coded one.

pkg/healthcheck/healthcheck.go
liquidslr: i didn't knew that these all also imply that the pod isn't scheduled, if that is the case then probably we can check pod phase here itself, but i think it would be better to keep the two checks separate  i.e. the one occurring due to node failure and the others that might occur due to `Init:ImagePullBackOff` or `CrashLoopBackOff` , probably another check can be added  for the later

pkg/healthcheck/healthcheck.go
liquidslr: yes we can have one more check as mentioned in the above comment, just to keep things organised

pkg/healthcheck/healthcheck.go
liquidslr: also because pod phase only returns pending, running, failed or unknown we will have to check pod status and condition to know the actual reason of pod being unscheduled

pkg/healthcheck/healthcheck.go
ihcsim: > PSP can forbid a pod to be created

That makes sense, since the PSP is an admission controller.

In that case, can we check for the `v1beta1.ReplicaSetReplicaFailure` error instead? Looking at the [doc](https://godoc.org/k8s.io/api/apps/v1#ReplicaSetConditionType), the replica set condition will have this `Type` if the _pods fails to be created due to insufficient quota, limit ranges, pod security policy, node selectors, etc. or deleted due to kubelet being down or finalizers are failing_.

```diff
diff --git a/pkg/healthcheck/healthcheck.go b/pkg/healthcheck/healthcheck.go
index e7f4247f..a24c118c 100644
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -976,19 +976,18 @@ func checkUnschedulablePods(pods []corev1.Pod) error {
 	return nil
 }
 
-func checkPodSecurityPolicies(rst []v1beta1.ReplicaSet) error {
-	var replicaSets []string
+func checkControlPlaneReplicaSets(rst []v1beta1.ReplicaSet) error {
+	var errors string
 	for _, rs := range rst {
 		for _, r := range rs.Status.Conditions {
-			if strings.Contains(r.Message, "unable to validate against any pod security policy") {
-				replicaSets = append(replicaSets, rs.Name)
+			if r.Type == v1beta1.ReplicaSetReplicaFailure && r.Status == v1beta1.ConditionTrue {
+				errors += fmt.Sprintf("%s: %s, %s\n", rs.Name, r.Reason, r.Message)
 			}
 		}
 	}
 
-	if len(replicaSets) != 0 {
-		replicasets := strings.Join(replicaSets, ",")
-		return fmt.Errorf("Invalid pod security policy for the replicasets %s", replicasets)
+	if errors != "" {
+		return fmt.Errorf("errors in creating control plane replicas: ", errors)
 	}
 
 	return nil
```

liquidslr: > ```diff
> v1beta1.ConditionTrue
> ```
i couldn't find this in the [doc](https://godoc.org/k8s.io/api/apps/v1beta1)

pkg/healthcheck/healthcheck.go
ihcsim: Yeah, checking the phase does feel a bit cumbersome. What do you think of just:

```
if condition.Type = corev1.PodReasonUnscheduleable
```

replacing the literal with the [`corev1.PodReasonUnscheduleable`](https://godoc.org/k8s.io/api/core/v1#PodConditionType) constant.

liquidslr: yeah this can be done 👍

pkg/healthcheck/healthcheck.go
ihcsim: Sorry, it's `corev1.ConditionTrue`.

liquidslr: i thiink it should be `corev1.ConditionTrue `

liquidslr: The output of the above code is something like this

```
× correct pod security policies
    errors in creating control plane replicas linkerd-controller-8647d7d878: FailedCreate, pods "linkerd-controller-8647d7d
878-" is forbidden: unable to validate against any pod security policy: []
linkerd-controller-cd8f8d4c5: FailedCreate, pods "linkerd-controller-cd8f8d4c5-" is forbidden: unable to validate against a
ny pod security policy: []
linkerd-grafana-7f8b57bd95: FailedCreate, pods "linkerd-grafana-7f8b57bd95-" is forbidden: unable to validate against any p
od security policy: []
linkerd-identity-67cfbdf4d5: FailedCreate, pods "linkerd-identity-67cfbdf4d5-" is forbidden: unable to validate against any pod security policy: []
linkerd-prometheus-64b89cd5cd: FailedCreate, pods "linkerd-prometheus-64b89cd5cd-" is forbidden: unable to validate against any pod security policy: []
linkerd-sp-validator-8465d9789d: FailedCreate, pods "linkerd-sp-validator-8465d9789d-" is forbidden: unable to validate against any pod security policy: []
linkerd-web-6cd6965f77: FailedCreate, pods "linkerd-web-6cd6965f77-" is forbidden: unable to validate against any pod security policy: []
 
    see https://linkerd.io/checks/#l5d-existence-psp for hints

Status check results are ×
```

I dont think that we need to show Reason and Message for each pods as we know it is same for all the pods i.e. `Invalid pod security policy`

ihcsim: Yup, fine by me.

ihcsim:
@liquidslr We are so close! The integration tests are failing. Would you mind fixing them? Then we should be good to go.

```diff
diff --git a/test/testdata/check.golden b/test/testdata/check.golden
index af4d04ec..2d12b17b 100644
--- a/test/testdata/check.golden
+++ b/test/testdata/check.golden
@@ -11,6 +11,8 @@ kubernetes-version
 linkerd-existence
 -----------------
 √ control plane namespace exists
+√ control plane components ready
+√ no unschedulable pods
 √ controller pod is running
 √ can initialize the client
 √ can query the control plane API
diff --git a/test/testdata/check.proxy.golden b/test/testdata/check.proxy.golden
index 91a53723..c9079371 100644
--- a/test/testdata/check.proxy.golden
+++ b/test/testdata/check.proxy.golden
@@ -11,6 +11,8 @@ kubernetes-version
 linkerd-existence
 -----------------
 √ control plane namespace exists
+√ control plane components ready
+√ no unschedulable pods
 √ controller pod is running
 √ can initialize the client
 √ can query the control plane API
```

PR #2515, opened 2019-03-18
Remove `--tls=optional` and `linkerd-ca`

olix0r:
The proxy's TLS implementation has changed to use a new _Identity_ controller.

In preparation for this, the `--tls=optional` CLI flag has been removed
from install and inject; and the `ca` controller has been deleted. Metrics
and UI treatments for TLS have **not** been removed, as they will continue to
be valuable for the new Identity system.

With the removal of the old identity scheme, the Destination service's proxy
ID field is now set with an opaque string (e.g. `ns:emojivoto`) to enable
locality awareness.


olix0r:
i've successfully run integration tests against this branch.

olix0r:
cc @adleong for proxy_id value change

cli/cmd/root.go
```diff
@@ -417,7 +408,6 @@ func addProxyConfigFlags(cmd *cobra.Command, options *proxyConfigOptions) {
 	cmd.PersistentFlags().StringVar(&options.proxyMemoryRequest, "proxy-memory-request", options.proxyMemoryRequest, "Amount of Memory that the proxy sidecar requests")
 	cmd.PersistentFlags().StringVar(&options.proxyCPULimit, "proxy-cpu-limit", options.proxyCPULimit, "Maximum amount of CPU units that the proxy sidecar can use")
 	cmd.PersistentFlags().StringVar(&options.proxyMemoryLimit, "proxy-memory-limit", options.proxyMemoryLimit, "Maximum amount of Memory that the proxy sidecar can use")
-	cmd.PersistentFlags().StringVar(&options.tls, "tls", options.tls, "Enable TLS; valid settings: \"optional\"")
```
siggy: should `tls` also be removed from `proxyConfigOptions`?

cli/cmd/install_test.go
```diff
@@ -21,33 +21,30 @@ func TestRender(t *testing.T) {
...
+		CreatedByAnnotation:      "CreatedByAnnotation",
+		ProxyContainerName:       "ProxyContainerName",
+		ProxyAutoInjectEnabled:   true,
+		ProxyInjectAnnotation:    "ProxyInjectAnnotation",
+		ProxyInjectDisabled:      "ProxyInjectDisabled",
+		ControllerUID:            2103,
+		EnableH2Upgrade:          true,
+		NoInitContainer:          false,
+		GlobalConfig:             "GlobalConfig",
+		ProxyConfig:              "ProxyConfig",
```
siggy: about 30 lines below this there's a `noInitContainerWithProxyAutoInjectOptions.tls = "optional"`. intentional?

controller/api/destination/server.go
```diff
@@ -96,11 +94,17 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
-	// <deployment>.deployment.<namespace>.linkerd-managed.linkerd.svc.cluster.local
-	if len(proxyID) >= 3 {
-		proxyNS = proxyID[2]
+	parts := strings.Split(dest.ProxyId, ":")
+	// ns:<namespace>
+	if len(parts) == 2 && parts[0] == "ns" {
+		proxyNS = parts[1]
+	} else {
+		parts = strings.Split(dest.ProxyId, ".")
+		// <deployment>.deployment.<namespace>.linkerd-managed.linkerd.svc.cluster.local
```
siggy: would this happen if there's an older proxy / injected deployment? if so, should we log it?

test/install_test.go
```diff
@@ -53,19 +53,15 @@ var (
...
-		`.*-tls linkerd-(ca|controller|grafana|prometheus|web)-.*-.* linkerd-proxy WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io\("/var/linkerd-io/identity/certificate\.crt", Some\(2\)\), falling back`,
-		`.*-tls linkerd-(ca|controller|grafana|prometheus|web)-.*-.* linkerd-proxy WARN admin={bg=tls-config} linkerd2_proxy::transport::tls::config error reloading TLS config: Io\("/var/linkerd-io/trust-anchors/trust-anchors\.pem", Some\(2\)\), falling back`,
 
-		`.*-tls linkerd-(ca|controller|grafana|prometheus|web)-.*-.* linkerd-proxy WARN proxy={server=in listen=0\.0\.0\.0:4143} rustls::session Sending fatal alert AccessDenied`,
-		`.*-tls linkerd-(ca|controller|grafana|prometheus|web)-.*-.* linkerd-proxy ERR! proxy={server=in listen=0\.0\.0\.0:4143 remote=.*} linkerd2_proxy::proxy::http::router service error: an IO error occurred: Connection reset by peer (os error 104)`,
+		`.*-tls linkerd-(controller|grafana|prometheus|web)-.*-.* linkerd-proxy WARN proxy={server=in listen=0\.0\.0\.0:4143} rustls::session Sending fatal alert AccessDenied`,
```
siggy: you can remove all of these `.*-tls` log lines unless you're planning to re-use them later.

cli/cmd/testdata/inject_emojivoto_deployment_tls.golden.yml
```diff
@@ -126,14 +107,5 @@ spec:
           runAsNonRoot: false
           runAsUser: 0
         terminationMessagePolicy: FallbackToLogsOnError
-      volumes:
-      - configMap:
-          name: linkerd-ca-bundle
-          optional: true
-        name: linkerd-trust-anchors
```
siggy: i still see references to `linkerd-trust-anchors`, should these go away completely? same question for `linkerd-trust-anchors`, `linkerd-secrets`, `/var/linkerd-io/identity`, and `/var/linkerd-io/trust-anchors`

controller/api/destination/server.go
olix0r: It's really only during upgrade that this is necessary -- i.e. when an older proxy is talking to a newer control plane. If we're worried more generally about older proxies staying around, there are probably better ways to flag this than in the destination service logs.  That said, will add a comment here to remove this in the next release.

controller/proxy-injector/fake/data/config-proxy.yaml
```diff
@@ -1,66 +1,49 @@
 env:
```
klingerf: Sigh, this file and `config-proxy-init.yaml` are completely unused, so you could remove both of them, and their corresponding consts in `factory.go`:

```diff
diff --git a/controller/proxy-injector/fake/factory.go b/controller/proxy-injector/fake/factory.go
index 10511799..76551299 100644
--- a/controller/proxy-injector/fake/factory.go
+++ b/controller/proxy-injector/fake/factory.go
@@ -14,8 +14,6 @@ import (
 const (
 	DefaultControllerNamespace = "linkerd"
 	DefaultNamespace           = "default"
-	FileProxySpec              = "fake/data/config-proxy.yaml"
-	FileProxyInitSpec          = "fake/data/config-proxy-init.yaml"
 	DefaultNoInitContainer     = false
 )
 
```

PR #2521, opened 2019-03-18
Introduce the Identity controller implementation

olix0r:
This change introduces a new Identity service implementation for the
`ioo.linkerd.proxy.identity.Identity` gRPC service.

The `pkg/identity` contains a core, abstract implentation of the service
(generic over both the CA and (Kubernetes) Validator interfaces).

`controller/identity` inclues a concrete implementation that uses the
Kubernetes TokenReview API to validate serviceaccount tokens when
issuing certificates.

This change does **NOT** alter installation or runtime to include the
identity service. This will be included in a follow-up.


pkg/identity/service.go
```diff
@@ -0,0 +1,180 @@
...
+
+// Register registers an identity service implementation in the provided gRPC
+// server.
+func Register(g *grpc.Server, s *Service) {
+	pb.RegisterIdentityServer(g, s)
+}
+
+// Certify validates identity and signs certificates.
+func (s *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.CertifyResponse, error) {
+	// Extract the relevant info form the request.
```
olix0r: "from the request"

pkg/tls/ca.go
```diff
@@ -66,14 +71,14 @@ const (
 	// DefaultClockSkewAllowance indicates the maximum allowed difference in clocks
 	// in the network.
 	//
-	// Allow an two hours of clock skew.
+	// Allow an two minutes of clock skew.
```
olix0r: Phrasing. Also 2m is probably excessive

controller/cmd/identity/main.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	csa := 0 * time.Minute
+	if pbd := idctx.GetClockSkewAllowance(); pbd != nil {
+		if d, err := ptypes.Duration(pbd); err == nil {
+			csa = d
+		}
+	}
+
+	il := 24 * time.Hour
```
olix0r: constify DefaultIssuanceLifetime

pkg/identity/service.go
```diff
@@ -0,0 +1,187 @@
...
+
+// NewService creates a new identity service.
+func NewService(v Validator, i tls.Issuer) *Service {
+	return &Service{v, i}
+}
+
+// Register registers an identity service implementation in the provided gRPC
+// server.
+func Register(g *grpc.Server, s *Service) {
+	pb.RegisterIdentityServer(g, s)
```
klingerf: This doesn't compile for me. Our pattern in most of our other gRPC servers is to include a `NewServer` initialization func that handles registration and returns `*grpc.Server` and `net.Listener`. For instance:

https://github.com/linkerd/linkerd2/blob/81f645da665bcec99ccfc26233d7d84a56ce41f3/controller/tap/server.go#L430-L452

Can we do something like that here?

pkg/identity/service.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	crts := crt.ExtractRaw()
+	if len(crts) == 0 {
+		panic("We seem to have lost the certificate while signing it?")
+	}
+
+	// Bundle issuer crt with certificate so the trust path to the root can be verified.
+	log.Infof("certifying %s until %s", tokIdentity, crt.Certificate.NotAfter)
+	validUntil, err := ptypes.TimestampProto(crt.Certificate.NotAfter)
```
klingerf: There's a similar protobuf compilation issue here with this timestamp:

```
pkg/identity/service.go:127:3: cannot use validUntil (type *"github.com/linkerd/linkerd2/vendor/github.com/golang/protobuf/ptypes/timestamp".Timestamp) as type *"github.com/linkerd/linkerd2-proxy-api/vendor/github.com/golang/protobuf/ptypes/timestamp".Timestamp in field value
```

controller/identity/domain.go
```diff
@@ -0,0 +1,41 @@
...
+	if !isLabel(ns) {
+		return "", fmt.Errorf("Namespace account must be a label: '%s'", ns)
+	}
+
+	id := fmt.Sprintf("%s.%s.%s.identity.%s.%s", nm, ns, typ, d.controlNamespace, d.domain)
+	return id, nil
+}
+
+func isLabel(p string) bool {
+	return p != "" && !strings.Contains(p, ".")
```
klingerf: Let's use [`IsDNS1123Label`](https://godoc.org/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Label) from apimachinery instead of writing this validation ourselves. We import the "k8s.io/apimachinery/pkg/util/validation" a few places in the codebase already.

controller/identity/validator.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	kauthnApi "k8s.io/api/authentication/v1"
+	kauthzApi "k8s.io/api/authorization/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	kauthn "k8s.io/client-go/kubernetes/typed/authentication/v1"
+	kauthz "k8s.io/client-go/kubernetes/typed/authorization/v1"
+
```
klingerf: No extra newline in imports

controller/identity/validator.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := checkAccess(k8s.AuthorizationV1()); err != nil {
+		return nil, err
+	}
+
+	authn := k8s.AuthenticationV1()
+	return &K8sTokenValidator{authn, domain}, nil
+}
+
+func init() {
+	var _ identity.Validator = &K8sTokenValidator{}
```
klingerf: I can't figure out what this `init()` function is doing. Mind adding a comment to explain it?

pkg/identity/service.go
olix0r: This pkg module is intentionally kubernetes-agnostic; and it exposes this helper so that the caller doesn't need to import the protobuf library just for this one call.

It would be a shame to make this module pull in a kubernetes dependency just for this builder. But perhaps the controller module should export a constructor like this?

controller/identity/validator.go
olix0r: it's asserting that the type implements the interface

pkg/identity/service.go
olix0r: The poiny about kubernetes isn't exactly relevant I guess.. but the larger point is that the `pkg/` module is just a "library" and doesn't have any grpc.Server/net.Listener dependencies -- it only deals in the interfaces.  I think the `controller/identity` package should be responsible for this, if we care to split out of the main...

controller/identity/validator.go
olix0r: The intent was to group the repo-local packages separately from external deps.

pkg/identity/service.go
olix0r: Hmm.. How are you getting that? I can't reproduce this...
```
:; bin/lint ./pkg/...    
:; go version
go version go1.11.5 darwin/amd64
```

pkg/identity/service.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	// Create a certificate
+	crt, err := s.i.IssueEndEntityCrt(csr)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	crts := crt.ExtractRaw()
+	if len(crts) == 0 {
+		panic("We seem to have lost the certificate while signing it?")
```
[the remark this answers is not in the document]
olix0r: Reaching this indicates a programming error in the Issuer implementation, so the intent is to (1) be fragile to programmer errors and (2) be defensive against incorrect interface implementations.

controller/identity/validator.go
olix0r: I wonder if this is something we can enforce with a lint...

olix0r:
@klingerf I've addressed most of your feedback. I can't reproduce the compilation errors you're seeing, so I have not fixed those. And I would prefer to keep `pkg/identity` independent of application concerns like binding the service to a socket, prometheus, etc.

pkg/identity/service.go
olix0r: My thought process is the following: pkg/identity implements the grpc interface in terms of two other interfaces: Validator, and Issuer.  The controller, which handles configuration and runtime logic, builds a validator and Issuer.  I structured this specifically this way because I know of another user that wants to implement another identity service. Providing a clear library interface for implementing this seemed preferable to bundling the logical service and the concrete server.

pkg/identity/service.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	if len(csr.URIs) > 0 {
+		return errors.New("Cannot validate URIs")
+	}
+
+	return nil
+}
+
+func (e NotAuthenticated) Error() string {
+	_ = e // satisfy linter
```
klingerf: Am pretty sure this would work for the purpose of linting:

```
func (NotAuthenticated) Error() string {
	return "Authentication token could not be authenticated"
}
```

controller/identity/validator.go
klingerf: Ah, thanks for the explanation. It would be more conventional to have the return type of `NewK8sTokenValidator` be `identity.Validator`, which would ensure that `K8sTokenValidator` satisfies the interface. I'm thinking something like:

```diff
diff --git a/controller/identity/validator.go b/controller/identity/validator.go
index 76554444..e6115ad3 100644
--- a/controller/identity/validator.go
+++ b/controller/identity/validator.go
@@ -29,7 +29,7 @@ type K8sTokenValidator struct {
 func NewK8sTokenValidator(
 	k8s k8s.Interface,
 	domain *TrustDomain,
-) (*K8sTokenValidator, error) {
+) (identity.Validator, error) {
 	if err := checkAccess(k8s.AuthorizationV1()); err != nil {
 		return nil, err
 	}
@@ -38,11 +38,6 @@ func NewK8sTokenValidator(
 	return &K8sTokenValidator{authn, domain}, nil
 }
 
-func init() {
-	// Assert that the struct implements the interface.
-	var _ identity.Validator = &K8sTokenValidator{}
-}
-
 // Validate accepts kubernetes bearer tokens and returns a DNS-form linkerd ID.
 func (k *K8sTokenValidator) Validate(_ context.Context, tok []byte) (string, error) {
 	// TODO: Set/check `audience`
```

pkg/identity/service.go
klingerf: Ok, that all makes sense. Since this implementation deviates from the implementation of the rest of our gRPC services in the project, it's adding complexity to the codebase. We can certainly justify the complexity, as it exists to satisfy some external use case, but I wanted to at least note it.

controller/identity/domain.go
```diff
@@ -0,0 +1,41 @@
...
+// TrustDomain is a namespace for identities.
+type TrustDomain struct {
+	controlNS, domain string
+}
+
+// NewTrustDomain creates a new identity namespace.
+func NewTrustDomain(controlNS, domain string) (*TrustDomain, error) {
+	if errs := validation.IsDNS1123Label(controlNS); len(errs) > 0 {
+		for _, e := range errs {
+			return nil, fmt.Errorf("Invalid label '%s': %s", controlNS, e)
```
alpeb: I believe the convention in go is to have error strings not be capitalized, as they're usually embedded into other strings.

controller/identity/domain.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+	}
+	if domain == "" {
```
alpeb: This check isn't needed as the regex inside `IsDNS1123Label()` doesn't allow empty strings.

controller/identity/validator.go
```diff
@@ -0,0 +1,98 @@
...
+// access check fails.
+func NewK8sTokenValidator(
+	k8s k8s.Interface,
+	domain *TrustDomain,
+) (*K8sTokenValidator, error) {
+	if err := checkAccess(k8s.AuthorizationV1()); err != nil {
+		return nil, err
+	}
+
+	authn := k8s.AuthenticationV1()
```
alpeb: We could have this guy be declared at the beginning of this function to avoid calling `AuthenticationV1()` twice.

pkg/identity/service.go
```diff
@@ -0,0 +1,186 @@
...
+
+// Certify validates identity and signs certificates.
+func (s *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.CertifyResponse, error) {
+	// Extract the relevant info from the request.
+	reqIdentity, tok, csr, err := checkRequest(req)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	if err = checkCSR(csr, reqIdentity); err != nil {
+		log.Debug()
```
alpeb: Is this supposed to be empty?

pkg/identity/service.go
```diff
@@ -0,0 +1,186 @@
...
+	// DefaultIssuanceLifetime is the default lifetime of certificates issued by
+	// the identity service.
+	DefaultIssuanceLifetime = 24 * time.Hour
+)
+
+type (
+	// Service implements the gRPC service in terms of a Validator and Issuer.
+	Service struct {
+		v Validator
+		i tls.Issuer
```
alpeb: TIOLI: Would you mind having the `Service` struct have a bit more descriptive field names?

olix0r: Sure. I always feel a bit silly having field names that are just the type names: the code ends up reading as `s.issuer.Issue()` and `s.validator.Validate()` which isn't really any clearer than `s.i.Issue()` and s.v.Validate`, imo... perhaps I should just embed the interfaces?  Either way, i'm not opposed to clarifying it.

controller/identity/domain.go
olix0r: But the domain isn't restricted to be just a label... it needs to be a domain suffix.

pkg/identity/service.go
klingerf: Ooh, big 👍 to embedding both of these interfaces in the `Service` struct.

controller/identity/domain.go
olix0r: Changed this to use a kubernetes validator

controller/identity/validator.go
olix0r: One is `AuthorizationV1` and the other is `AuthenticationV1`

pkg/identity/service.go
```diff
@@ -64,58 +64,58 @@ func Register(g *grpc.Server, s *Service) {
...
-func (s *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.CertifyResponse, error) {
+func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.CertifyResponse, error) {
 	// Extract the relevant info from the request.
 	reqIdentity, tok, csr, err := checkRequest(req)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 	if err = checkCSR(csr, reqIdentity); err != nil {
-		log.Debug()
+		log.Debugf("requster sent invalid CSR: %s", err)
```
alpeb: _requester_

PR #2524, opened 2019-03-19
Update to identity-capable proxy

olix0r:
The new proxy has changed its configuration as follows:

- `LISTENER` urls are now `LISTEN_ADDR` addresses;
- `CONTROL_URL` is now `DESTINATION_SVC_ADDR`;
- `*_NAMESPACE` vars are no longer needed;
- The `PROXY_ID` is now the `DESTINATION_CONTEXT`;
- The "metrics" port is now the "admin" port, since it serves more than
  just metrics;
- A readiness probe now checks a dedicated /ready endpoint eagerly.


pkg/inject/inject.go
```diff
@@ -336,71 +341,70 @@ func (conf *ResourceConfig) complete(template *v1.PodTemplateSpec) {
...
 			},
 			{
-				Name:  envVarProxyInboundAcceptKeepAlive,
+				Name:  envOutboundConnectKeepAlive,
 				Value: fmt.Sprintf("%dms", defaultKeepaliveMs),
 			},
 			{
-				Name:  envVarProxyOutboundConnectKeepAlive,
-				Value: fmt.Sprintf("%dms", defaultKeepaliveMs),
+				Name:      "K8S_NS",
```
ihcsim: Nit: make this as a `const`, to be consistent with the others env vars?

pkg/inject/inject.go
```diff
@@ -422,35 +426,39 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
 		}
 	}
 
-	if conf.globalConfig.GetIdentityContext() != nil {
-		log.Warn("TLS configuration exists but is currently not available")
+	sidecar.Env = append(sidecar.Env, v1.EnvVar{
+		Name:  "LINKERD2_PROXY_IDENTITY_DISABLED",
```
ihcsim: Nit: Declare this as a `const`?

pkg/inject/inject.go
```diff
@@ -656,15 +664,24 @@ func (conf *ResourceConfig) proxyUID() int64 {
...
+
+func (conf *ResourceConfig) proxyLivenessProbe() *v1.Probe {
 	return &v1.Probe{
 		Handler: v1.Handler{
 			HTTPGet: &v1.HTTPGetAction{
 				Path: "/metrics",
-				Port: intstr.IntOrString{
-					IntVal: metricsPort,
-				},
+				Port: intstr.IntOrString{IntVal: conf.proxyAdminPort()},
```
ihcsim: Should the `Path` be renamed to `/admin` too?

pkg/k8s/labels.go
```diff
@@ -144,14 +144,18 @@ const (
...
+
 	// IdentityModeDisabled is assigned to IdentityModeAnnotation to
 	// disable the proxy from participating in automatic identity.
 	IdentityModeDisabled = "disabled"
 
 	// IdentityModeOptional is assigned to IdentityModeAnnotation to
 	// optionally configure the proxy to participate in automatic identity.
 	//
-	// Will be deprecated soon.
+	// Deprecated.
```
ihcsim: The convention for this is `Deprecated: <reason>`. See [here](https://github.com/golang/go/wiki/Deprecated).

pkg/k8s/labels.go
```diff
@@ -179,12 +183,12 @@ const (
 	 */
 
 	// MountPathBase is the base directory of the mount path
```
klingerf: Super, minor, but please add a trailing "." to this comment as well, for consistency with the changes that you're making on lines 188 and 191 in this file.

pkg/inject/inject.go
olix0r: I chose not to make this one a const since it is effectively a "temp variable" that is only used  for templating. Is there a better way to connote this? perhaps call it `_k8s_ns` or something?

pkg/inject/inject.go
```diff
@@ -422,35 +426,39 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
...
 	}
 
-	if conf.globalConfig.GetIdentityContext() != nil {
-		log.Warn("TLS configuration exists but is currently not available")
+	sidecar.Env = append(sidecar.Env, v1.EnvVar{
+		Name:  "LINKERD2_PROXY_IDENTITY_DISABLED",
+		Value: "Identity configuration is unavailable",
+	})
+	if idctx := conf.globalConfig.GetIdentityContext(); idctx != nil {
+		log.Warn("Ignoring Identity configuration.")
```
ihcsim: Curious as to what will this do eventually? I assume it won't stay as `log.Warn()`.

pkg/inject/inject.go
olix0r: No, we want the liveness probe to check the metrics endpoint still

pkg/inject/inject.go
olix0r: https://github.com/linkerd/linkerd2/blob/ver/identikit/pkg/inject/inject.go#L484-L554

pkg/inject/inject.go
ihcsim: Sure, I think prefixing this and the [rest](https://github.com/linkerd/linkerd2/blob/ver/identikit/pkg/inject/inject.go#L484-L554) with an underscore will communicate to future readers that this is intentional.

pkg/k8s/labels.go
olix0r: Actually, just going to delete this, as there's no reason to use it going forward.

PR #2526, opened 2019-03-19
install: Introduce the Identity controller

olix0r:
https://github.com/linkerd/linkerd2/pull/2521 introduces an "Identity" controller, but there is no way to include it in linkerd installation.

This change alters the `install` flow as follows:
- An Identity service is _always_ installed;
- Issuer credentials may be specified via the CLI;
- If no Issuer credentials are provided, they are generated each time `install` is called.
- It's possible to override the credential generation logic, especially for tests;
- Proxies are NOT configured to use the identity service.


cli/cmd/install.go
```diff
@@ -135,8 +182,89 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+		identity = identityOverride
+	} else if idopts := options.identityOptions; idopts != nil {
+		trustDomain := options.identityOptions.trustDomain
+		if trustDomain == "" {
+			return nil, errors.New("Trust domain must be specified")
+		}
+		issuerName := fmt.Sprintf("identity.%s.%s", controlPlaneNamespace, trustDomain)
+
+		// Load signing material from options...
+		if idopts.trustPEMFile != "" || idopts.crtPEMFile != "" || idopts.keyPEMFile != "" {
```
olix0r: these flags need to be added to the CLI...

cli/cmd/install.go
```diff
@@ -135,8 +182,89 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+			if err := creds.Verify(roots, issuerName); err != nil {
+				return nil, fmt.Errorf("Credentials cannot be validated: %s", err)
+			}
+
+			identity = &installIdentityConfig{
+				TrustDomain:     idopts.trustDomain,
+				TrustAnchorsPEM: trustAnchorsPEM,
+				Issuer: &issuerConfig{
+					ClockSkewAllowance:  options.identityOptions.clockSkewAllowance.String(),
+					IssuanceLifetime:    options.identityOptions.issuanceLifetime.String(),
```
klingerf: Let's use `idopts` here as well:

```
					ClockSkewAllowance:  idopts.clockSkewAllowance.String(),
					IssuanceLifetime:    idopts.issuanceLifetime.String(),
```

cli/cmd/install.go
```diff
@@ -135,8 +182,89 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("Failed to create root certificate for identity: %s", err)
+			}
+
+			identity = &installIdentityConfig{
+				TrustDomain:     trustDomain,
+				TrustAnchorsPEM: root.Cred.Crt.EncodeCertificatePEM(),
+				Issuer: &issuerConfig{
+					ClockSkewAllowance:  options.identityOptions.clockSkewAllowance.String(),
+					IssuanceLifetime:    options.identityOptions.issuanceLifetime.String(),
```
klingerf: And here:

```
					ClockSkewAllowance:  idopts.clockSkewAllowance.String(),
					IssuanceLifetime:    idopts.issuanceLifetime.String(),
```

cli/cmd/install.go
```diff
@@ -241,7 +371,7 @@ func render(config installConfig, w io.Writer, options *installOptions) error {
 
 	injectOptions := newInjectOptions()
 
-	*injectOptions.proxyConfigOptions = *options.proxyConfigOptions
+	injectOptions.proxyConfigOptions = options.proxyConfigOptions
```
klingerf: This func modifies `injectOptions.proxyConfigOptions`, which will now also modify `options.proxyConfigOptions` if you make this change. Was it intended? The previous setup ensured that this func did not modify the `options` param that's passed to it. We could add a comment to that effect to make this clearer.

cli/cmd/install.go
```diff
@@ -280,59 +433,77 @@ func readIntoBytes(filename string) ([]byte, error) {
...
+		if err != nil {
+			csa = defaultIdentityClockSkewAllowance
+		}
+
+		identityContext = &pb.IdentityContext{
+			TrustDomain:        id.TrustDomain,
+			TrustAnchorsPem:    id.TrustAnchorsPEM,
+			IssuanceLifetime:   ptypes.DurationProto(il),
+			ClockSkewAllowance: ptypes.DurationProto(csa),
+		}
```
klingerf: The code inside this if condition is identical to the code that constructs the `IdentityContext` on lines 388-406 above. Consider moving it to a shared func. Something like:

```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 52d8b328..07ce852f 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -384,25 +384,7 @@ func render(config installConfig, w io.Writer, options *installOptions) error {
 	// injectOptionsToConfigs does NOT set an identity context if none exists,
 	// since it can't be enabled at inject-time if it's not enabled at
 	// install-time.
-	if config.Identity != nil {
-		id := config.Identity
-		il, err := time.ParseDuration(id.Issuer.IssuanceLifetime)
-		if err != nil {
-			il = defaultIdentityIssuanceLifetime
-		}
-
-		csa, err := time.ParseDuration(id.Issuer.ClockSkewAllowance)
-		if err != nil {
-			csa = defaultIdentityClockSkewAllowance
-		}
-
-		pbConfig.global.IdentityContext = &pb.IdentityContext{
-			TrustDomain:        id.TrustDomain,
-			TrustAnchorsPem:    id.TrustAnchorsPEM,
-			IssuanceLifetime:   ptypes.DurationProto(il),
-			ClockSkewAllowance: ptypes.DurationProto(csa),
-		}
-	}
+	pbConfig.global.IdentityContext = config.Identity.toIdentityContext()
 
 	return processYAML(&buf, w, ioutil.Discard, resourceTransformerInject{
 		configs: pbConfig,
@@ -434,31 +416,11 @@ func readIntoBytes(filename string) ([]byte, error) {
 }
 
 func globalConfig(options *installOptions, id *installIdentityConfig) *pb.Global {
-	var identityContext *pb.IdentityContext
-	if id != nil {
-		il, err := time.ParseDuration(id.Issuer.IssuanceLifetime)
-		if err != nil {
-			il = defaultIdentityIssuanceLifetime
-		}
-
-		csa, err := time.ParseDuration(id.Issuer.ClockSkewAllowance)
-		if err != nil {
-			csa = defaultIdentityClockSkewAllowance
-		}
-
-		identityContext = &pb.IdentityContext{
-			TrustDomain:        id.TrustDomain,
-			TrustAnchorsPem:    id.TrustAnchorsPEM,
-			IssuanceLifetime:   ptypes.DurationProto(il),
-			ClockSkewAllowance: ptypes.DurationProto(csa),
-		}
-	}
-
 	return &pb.Global{
 		LinkerdNamespace: controlPlaneNamespace,
 		CniEnabled:       options.noInitContainer,
 		Version:          options.linkerdVersion,
-		IdentityContext:  identityContext,
+		IdentityContext:  id.toIdentityContext(),
 	}
 }
 
@@ -509,3 +471,26 @@ func proxyConfig(options *installOptions) *pb.Proxy {
 		DisableExternalProfiles: options.disableExternalProfiles,
 	}
 }
+
+func (id *installIdentityConfig) toIdentityContext() *pb.IdentityContext {
+	if id == nil {
+		return nil
+	}
+
+	il, err := time.ParseDuration(id.Issuer.IssuanceLifetime)
+	if err != nil {
+		il = defaultIdentityIssuanceLifetime
+	}
+
+	csa, err := time.ParseDuration(id.Issuer.ClockSkewAllowance)
+	if err != nil {
+		csa = defaultIdentityClockSkewAllowance
+	}
+
+	return &pb.IdentityContext{
+		TrustDomain:        id.TrustDomain,
+		TrustAnchorsPem:    id.TrustAnchorsPEM,
+		IssuanceLifetime:   ptypes.DurationProto(il),
+		ClockSkewAllowance: ptypes.DurationProto(csa),
+	}
+}
```

cli/cmd/testdata/install_ha_output.golden
```diff
@@ -5,6 +5,213 @@ metadata:
...
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        securityContext:
+          runAsUser: 2103
+        volumeMounts:
+        - mountPath: /var/run/linkerd/config
+          name: config
+        - mountPath: /var/run/linkerd/identity/issuer
```
olix0r: This should probably be `/var/linkerd-io` or something...

olix0r: No, disregard. I actually changed the base path in this PR so it's fine.

klingerf: Ah, that changed in #2524. I thought it was intentional.

```diff
-       // MountPathBase is the base directory of the mount path
-       MountPathBase = "/var/linkerd-io"
+       // MountPathBase is the base directory of the mount path.
+       MountPathBase = "/var/run/linkerd"
```

olix0r: It was intentional, ultimately. I just had expected to flip that in the next PR ;)

chart/templates/identity.yaml
```diff
@@ -0,0 +1,134 @@
...
+          containerPort: 8080
+        - name: admin-http
+          containerPort: 9990
+        image: {{.ControllerImage}}
+        imagePullPolicy: {{.ImagePullPolicy}}
+        args:
+        - "identity"
+        - "-log-level={{.ControllerLogLevel}}"
+        volumeMounts:
+        - mountPath: /var/run/linkerd/config
```
alpeb: Do you mind updating this path in `controller.yaml` as well? It still has the old `/var/linkerd-io/config`, which now breaks the auto-injector and the config gRPC endpoint.

cli/cmd/install.go
```diff
@@ -135,8 +227,86 @@ func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
...
+			if err != nil {
+				return nil, err
+			}
+			trustAnchorsPEM := string(trustb)
+			roots, err := tls.DecodePEMCertPool(trustAnchorsPEM)
+			if err != nil {
+				return nil, err
+			}
+
+			issuerName := "" // TODO restrict issuer name?
```
ihcsim: By _restrict issuer name_, are you thinking about eventually support whitelisted issuers?

chart/templates/identity.yaml
```diff
@@ -0,0 +1,136 @@
...
+apiVersion: extensions/v1beta1
+metadata:
+  name: linkerd-identity
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: identity
+  annotations:
+    {{.CreatedByAnnotation}}: {{.CliVersion}}
+spec:
+  replicas: {{.IdentityReplicas}}
```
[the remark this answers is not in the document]
klingerf: Whoa, good catch @ihcsim. The field doesn't exist, so we're setting this to null here and then it's being stripped when we inject the install config. Am surprised the helm renderer doesn't blow-up on unknown fields.

cli/cmd/install.go
olix0r: well, when we generate an issuer, we give it the name `identity.NS.TRUST-DOMAIN...` -- we could potentially enforce that naming for issuers, but it's not high-priority.

cli/cmd/install.go
```diff
@@ -99,7 +144,9 @@ func newCmdInstall() *cobra.Command {
 		Short: "Output Kubernetes configs to install Linkerd",
 		Long:  "Output Kubernetes configs to install Linkerd.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			config, err := validateAndBuildConfig(options)
+			// TODO check with a config already exists in the API and fail if it does.
```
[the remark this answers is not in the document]
olix0r: When you're generating an issuer, if you're installing over an existing controller, you'll break everything such that existing pods can't communicate with newer pods/control plane instances, since the trust root will change.  We need to make `install` fail if it would likely overwrite an existing install; and we need to introduce a new `upgrade` command that consumes the existing config to generate an updated set of resources (i.e. reusing the existing issuer material).

chart/templates/identity.yaml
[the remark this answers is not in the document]
olix0r: Fixed by https://github.com/linkerd/linkerd2/pull/2526/commits/9810b9c729a778e6583d3ecfe2340d2e8e441db5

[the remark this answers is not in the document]
klingerf: Fix looks good, but I think you need to update the install fixtures to include a `replicas:` field.

PR #2529, opened 2019-03-20
Add identity pod to check, web, and integration tests

klingerf:
This branch updates our integration tests, `linkerd check`, and the web UI to work with the identity pod that was added in #2526.

### DIFF
--- a/test/get/get_test.go
+++ b/test/get/get_test.go
@@ -33,10 +33,11 @@ var (
 	}
 
 	linkerdPods = map[string]int{
-		"linkerd-grafana":    1,
-		"linkerd-web":        1,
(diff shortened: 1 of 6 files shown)


test/get/get_test.go
```diff
@@ -33,10 +33,11 @@ var (
 	}
 
 	linkerdPods = map[string]int{
+		"linkerd-controller": 1,
 		"linkerd-grafana":    1,
-		"linkerd-web":        1,
+		"linkerd-identity":   1,
 		"linkerd-prometheus": 1,
-		"linkerd-controller": 1,
+		"linkerd-web":        1,
```
siggy: 👌

PR #2530, opened 2019-03-20
Restructure inject resource config struct

ihcsim:
This PR separates out the restructuring work done in #2500, where the `ResourceConfig` struct is restructured better encapsulate and separate a parent resource related fields from those of a pod.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -73,45 +73,50 @@ var injectableKinds = []string{
 	k8s.StatefulSet,
 }
 
-// objMeta provides a generic struct to parse the names of Kubernetes objects
-type objMeta struct {
-	*metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
-}
-
 // ResourceConfig contains the parsed information for a given workload
 type ResourceConfig struct {
-	configs       *config.All
-	nsAnnotations map[string]string
-	meta          metav1.TypeMeta
-	obj           runtime.Object
-	workLoadMeta  *metav1.ObjectMeta
-	podMeta       objMeta
-	podLabels     map[string]string
-	podSpec       *v1.PodSpec
-
+	configs                *config.All
+	nsAnnotations          map[string]string
 	destinationDNSOverride string
 	identityDNSOverride    string
 	proxyOutboundCapacity  map[string]uint
+
+	workload struct {
+		obj      runtime.Object
+		meta     *metav1.ObjectMeta
+		metaType metav1.TypeMeta
+	}
+
+	pod struct {
+		// Meta is the pod's metadata. It's exported so that the YAML marshaling
+		// will work in the ParseMeta() function.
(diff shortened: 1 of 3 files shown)


pkg/inject/inject.go
```diff
@@ -63,53 +63,59 @@ var injectableKinds = []string{
...
+	workload struct {
+		obj      runtime.Object
+		meta     *metav1.ObjectMeta
+		metaType metav1.TypeMeta
+	}
+
+	pod struct {
+		// Meta is the pod's metadata. It's exported so that the YAML marshaling
+		// will work in the ParseMeta() function.
+		Meta   *metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
alpeb: Oh.. having the `Meta` inside `pod` plays the same role `objMeta` did for parsing the yaml, but this is much nicer :100:

pkg/inject/inject.go
```diff
@@ -448,7 +454,7 @@ func (conf *ResourceConfig) injectProxyInit(patch *Patch) {
 		Image:                    conf.taggedProxyInitImage(),
 		ImagePullPolicy:          conf.proxyInitImagePullPolicy(),
 		TerminationMessagePolicy: v1.TerminationMessageFallbackToLogsOnError,
-		Args:                     conf.proxyInitArgs(),
+		Args: conf.proxyInitArgs(),
```
alpeb: oops seems your editor is not playing nice, and the linter is not liking it...

pkg/inject/inject.go
```diff
@@ -63,53 +63,59 @@ var injectableKinds = []string{
...
+	config := &ResourceConfig{
 		globalConfig:          globalConfig,
 		proxyConfig:           proxyConfig,
-		podLabels:             map[string]string{k8s.ControllerNSLabel: globalConfig.GetLinkerdNamespace()},
 		proxyOutboundCapacity: map[string]uint{},
 	}
+
+	config.pod.Meta = &metav1.ObjectMeta{}
+	config.pod.labels = map[string]string{k8s.ControllerNSLabel: globalConfig.GetLinkerdNamespace()}
+	return config
```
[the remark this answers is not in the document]
ihcsim: `pod` is a nested anonymous struct, so the way to initialize it directly is a bit verbose:
```
config := &ResourceConfig{
  // ...
  pod: struct {
    Meta   *metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
    labels map[string]string
    spec   *v1.PodSpec
  }{
    Meta:   &metav1.ObjectMeta{},
    labels: map[string]string{k8s.ControllerNSLabel: globalConfig.GetLinkerdNamespace()},
  },
}
```

ihcsim:
@alpeb Thanks for the feedback.

@klingerf Since this isn't a big change, I'm going to merge this. I'll like to continue with #2500, with this change included. If you have any post-merge feedback, we can address it in the other PR.

PR #2535, opened 2019-03-21
Check the cluster's config for install & inject

### LINKED ISSUE #2531: rfc: Changing install & inject to support identity configuration
## Problem

Linkerd's new identity implementation expects that proxies are configured with trust anchors and basic identity configuration at inject-time. During installation, we must include this information in generated configuration that is stored in the control plane; and during inject, we must fetch this configuration to configure proxies.

Currently, a configuration is generated during installation, but this configuration is only read by controller pods and not by the CLI. The CLI, on the other hand, is configured entirely by CLI flags (and their defaults) to produce a set of template values used to configure proxies (for both the control plane and other injected pods).

### `install`

When `install` is invoked, it attempts to determine whether a control plane already exists in the currently configured kubernetes cluster at the CLI-specified control plane namespace.

If a configuration already exists, then the `--overwrite` flag must be specified for install.

Aside from the process global configuration (like `controlPlaneNamespace`), all of the `installOptions` will NOT be initialized with defaults. Instead, they will be empty and their values will only be considered as overrides.

Instead, the install generates a `public.config.All` with defaults specified, though notably lacking any identity configuration.

Then, the install option override and this default config are used to produce `installValues`.  First, this process updates the config types from the overrides, (including consuming/generating an identity configuration). These are all fed into an `installValues` that   satisfies the control plane templates.

#### `upgrade`

We'll probably want to add a new `upgrade` command, or add an `--upgrade` to `install`, that uses an existing controller configuration instead of generating a new one. That should happen as a follow-up, though.

### `inject`

When `inject` is invoked, it attempts to determine whether a control plane already exists in the currently configured kubernetes cluster at the CLI-specified control plane namespace.

If a configuration does not already exists, then the `--disable-identity` flag must be specified, otherwise the command fails.

Similarly, `injectOptions` are only to be considered as overrides to the controller-provided configuration. Once the config is overridden from the options, the config is used to produce `injectValues` that can be used to configure the proxy for the given resources.

olix0r:
The introduction of identity in 0626fa37 created new state in the
control plane's configuration that must be considered when re-installing
the control plane or when injecting pods.

This change alters `install` to fail if it would seem to conflict with
an existing installation. This behavior may be disabled with the
`--ignore-cluster` flag.

Furthermore, `inject` now _requires_ that it can fetch a configuration
from the control plane in order to operate. Otherwise the
`--ignore-cluster` and `--disable-identity` flags must be specified.

This change does not actually instrument pods to use identity yet---it
lays the framework for proxy identity without changing the test fixture
output (besides a change to how identity HA is configured).

Fixes #2531

### DIFF
--- a/chart/templates/grafana.yaml
+++ b/chart/templates/grafana.yaml
@@ -46,7 +46,7 @@ spec:
         {{.Values.CreatedByAnnotation}}: {{.Values.CliVersion}}
     spec:
       volumes:
-      - name: {{.Values.GrafanaVolumeName}}
+      - name: data
         emptyDir: {}
       - name: grafana-config
         configMap:
@@ -65,10 +65,10 @@ spec:
           containerPort: 3000
         env:
         - name: GF_PATHS_DATA
-          value: /{{.Values.GrafanaVolumeName}}
+          value: /data
         volumeMounts:
-        - name: {{.Values.GrafanaVolumeName}}
-          mountPath: /{{.Values.GrafanaVolumeName}}
+        - name: data
+          mountPath: /data
         - name: grafana-config
           mountPath: /etc/grafana
           readOnly: true
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -9,6 +11,7 @@ import (
 
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/linkerd/linkerd2/controller/gen/config"
+	"github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/inject"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	log "github.com/sirupsen/logrus"
@@ -26,15 +29,17 @@ const (
 )
 
 type injectOptions struct {
+	ignoreCluster, disableIdentity bool
 	*proxyConfigOptions
 }
 
 type resourceTransformerInject struct {
-	configs
+	configs *config.All
+
 	proxyOutboundCapacity map[string]uint
 }
 
-func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, conf configs) int {
+func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, conf *config.All) int {
 	return transformInput(inputs, errWriter, outWriter, resourceTransformerInject{
 		configs: conf,
 	})
@@ -42,7 +47,10 @@ func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, conf confi
 
 func newInjectOptions() *injectOptions {
 	return &injectOptions{
-		proxyConfigOptions: newProxyConfigOptions(),
+		ignoreCluster: false,
+
+		// No proxy config overrides.
+		proxyConfigOptions: &proxyConfigOptions{},
 	}
 }
 
@@ -79,20 +87,32 @@ sub-folders, or coming from stdin.`,
 				return err
 			}
 
-			conf := injectOptionsToConfigs(options)
+			configs, err := options.fetchConfigsOrDefault()
+			if err != nil {
+				return err
+			}
 
-			exitCode := uninjectAndInject(in, stderr, stdout, conf)
+			options.overrideConfigs(configs)
+			exitCode := uninjectAndInject(in, stderr, stdout, configs)
 			os.Exit(exitCode)
 			return nil
 		},
 	}
 
 	addProxyConfigFlags(cmd, options.proxyConfigOptions)
+	cmd.PersistentFlags().BoolVar(
+		&options.ignoreCluster, "ignore-cluster", options.ignoreCluster,
+		"Ignore the current Kubernetes cluster when checking for existing cluster configuration (default false)",
+	)
+	cmd.PersistentFlags().BoolVar(
+		&options.disableIdentity, "disable-identity", options.disableIdentity,
+		"Disables resources from participating in identity",
+	)
 
 	return cmd
 }
 
-func uninjectAndInject(inputs []io.Reader, errWriter, outWriter io.Writer, conf configs) int {
+func uninjectAndInject(inputs []io.Reader, errWriter, outWriter io.Writer, conf *config.All) int {
 	var out bytes.Buffer
 	if exitCode := runUninjectSilentCmd(inputs, errWriter, &out, conf); exitCode != 0 {
 		return exitCode
@@ -101,7 +121,7 @@ func uninjectAndInject(inputs []io.Reader, errWriter, outWriter io.Writer, conf 
 }
 
 func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Report, error) {
-	conf := inject.NewResourceConfig(rt.global, rt.proxy)
+	conf := inject.NewResourceConfig(rt.configs)
 	if len(rt.proxyOutboundCapacity) > 0 {
 		conf = conf.WithProxyOutboundCapacity(rt.proxyOutboundCapacity)
 	}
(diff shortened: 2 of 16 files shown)


olix0r:
Note: the VolumeName fields were never configured, so they've been hardcoded (like other resource names that don't need to be configured).

Also, i've consolidated usage of configs around `*config.All`

olix0r:
CI failure is due to lints... i'm returning error messages I want shown to the user, so I'm using sentence capitalization; but the lint doesn't like that. Should i `//nolint`? Or change the error handling?

ihcsim:
> Should i `//nolint`? Or change the error handling?

My thought is to `//nolint` your stuff, so the rest of the code base can remain consistently lower case.

olix0r:
The alternative is that we just print to stderr and exit here instead of returning an error... Or that we return a specialized error type.

siggy:
The intent of requiring all errors begin with a lower case is so that they may be passed around in different contexts and concatenated together. At the point of logging, recommend prepending with a bit of capitalized context. A bit more detail:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

olix0r:
Thanks @siggy. Do you have a suggestion for how I should handle these specific cases?

siggy:
@olix0r I don't have the full context in front of me right now, but something like this pattern:

```go
values, configs, err := options.validateAndBuild()
if err != nil {
	log.Errorf("Option validation failure: %s", err)
	return err
}

func (options *installOptions) validate() error {
...
	return fmt.Errorf("unable to connect to a Kubernetes cluster to check for configuration. If this expected, use the --ignore-cluster flag.")
}
```

... or, if this is due to the `RunE` automatically printing the error returned, I would have assumed it would prepend that message with something similar?

olix0r:
For context, I'm trying to achieve what @grampelberg described in https://github.com/linkerd/linkerd2/issues/2531#issuecomment-475009724:

```
# prior to install
$ cat app.yaml | linkerd inject
Control plane is not running, please run linkerd install first.
$ cat app.yaml | linkerd inject --disable-identity
Control plane is not running, please run linkerd install first.
$ cat app.yaml | linkerd inject --disable-identity --ignore-cluster
deployment "vote-bot" injected

# first-time install
$ linkerd install | kubectl apply -f -
...
deployment.extensions/linkerd-web created
$ cat app.yaml | linkerd inject --verbose -
used defaults:
    linkerd-cni-enabled: true

# second time install / upgrade
$ linkerd install | kubectl apply -f -
You are already running a control plane. Please run linkerd upgrade if you would like to update.
$ linkerd upgrade | kubectl apply -f -
...
deployment.extensions/linkerd-web configured
$ linkerd install --ignore-cluster
...
deployment.extensions/linkerd-web configured
```

I'm also fine just punting this detail to a follow-up issue.

chart/templates/grafana.yaml
```diff
@@ -46,7 +46,7 @@ spec:
         {{.Values.CreatedByAnnotation}}: {{.Values.CliVersion}}
     spec:
       volumes:
-      - name: {{.Values.GrafanaVolumeName}}
+      - name: data
```
klingerf: 👍 Thanks for removing unnecessary variables.

cli/cmd/inject_test.go
```diff
@@ -48,21 +48,21 @@ func testUninjectAndInject(t *testing.T, tc testCase) {
 }
 
 func TestUninjectAndInject(t *testing.T) {
-	defaultConfig := newConfig()
-	defaultConfig.global.Version = "testinjectversion"
+	_, defaultConfig, _ := testInstallOptions().validateAndBuild()
```
ihcsim: Any reasons why skipping the error check here and in the next two calls? The tests still seem to pass with the checks.

cli/cmd/inject_util.go
```diff
@@ -17,11 +16,6 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-type configs struct {
-	global *config.Global
-	proxy  *config.Proxy
-}
```
klingerf: Nice catch -- glad we are switching to `config.All`

cli/cmd/install.go
```diff
@@ -90,11 +93,13 @@ type (
 		highAvailability   bool
 		controllerUID      int64
 		disableH2Upgrade   bool
+		ignoreCluster      bool
```
klingerf: Since `ignoreCluster` is a configuration option for both inject and install, my preference is to add it to the (poorly named) `proxyConfigOptions` struct, which holds all of the other flag values that are shared between inject and install.

cli/cmd/install.go
```diff
@@ -122,15 +128,36 @@ const (
 	proxyInjectorTemplateName  = "templates/proxy_injector.yaml"
 )
 
-func newInstallOptions() *installOptions {
+func defaultInstallOptions() *installOptions {
```
klingerf: I'm ok with making this change, but all of our subcommands right now are structured to have a `<cmd>Options` struct and a `new<Cmd>Options` func that initializes the struct with default values. We should either update all of the `new<Cmd>Options` funcs in this package to instead be called `default<Cmd>Options`, or we should revert this rename in this branch and do a more comprehensive rename in a separate branch.

cli/cmd/install.go
```diff
@@ -212,157 +239,118 @@ func newCmdInstall() *cobra.Command {
...
-func render(config installConfig, w io.Writer, options *installOptions) error {
+func toPromLogLevel(level string) (prom string) {
+	switch level {
+	case "panic", "fatal":
+		prom = "error"
+	default:
+		prom = level
+	}
+	return
+}
```
klingerf: I think the use of a named return in this method is making it more complicated than it needs to be. How about:

```
func toPromLogLevel(level string) string {
	switch level {
	case "panic", "fatal":
		return "error"
	default:
		return level
	}
}
```

cli/cmd/install.go
```diff
@@ -212,157 +239,118 @@ func newCmdInstall() *cobra.Command {
...
-func validateAndBuildConfig(options *installOptions) (*installConfig, error) {
-	if err := options.validate(); err != nil {
-		return nil, err
-	}
-
-	if options.highAvailability && options.controllerReplicas == defaultControllerReplicas {
-		options.controllerReplicas = defaultHAControllerReplicas
+func (options *installOptions) validate() error {
+	if options.identityOptions == nil {
+		log.Fatal("missing identity options")
```
klingerf: We don't use `log.Fatal` in this package, since all files in the package are being run from the CLI, and the formatting is jarring when the CLI exits with `log.Fatal`. If you need to exit immediately, use `fmt.Println` followed by `os.Exit(1)` (see dashboard.go for an example of that). Otherwise, just return an error and let cobra format it correctly.

cli/cmd/inject.go
```diff
@@ -249,48 +269,77 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output
...
+		install := defaultInstallOptions()
+		return install.configs(nil), nil
+	}
+
+	api := checkPublicAPIClientOrExit()
+	return api.Config(context.Background(), &public.Empty{})
+}
+
+// overrideConfigs uses command-line overrides to update the provided configs
+func (options *injectOptions) overrideConfigs(configs *config.All) {
```
ihcsim: I think we are missing a check on image pull policy override.

cli/cmd/inject.go
```diff
@@ -26,23 +29,28 @@ const (
 )
 
 type injectOptions struct {
+	ignoreCluster, disableIdentity bool
```
klingerf: Per comment in install.go, let's move `ignoreCluster` to `proxyConfigOptions`.

cli/cmd/install.go
olix0r: This is meant to denote a programming error. Would it be better to use `panic!`?

cli/cmd/install.go
olix0r: Would it be better to call this newInstallOptionsWithDefaults?

Now that injetOptions/proxyConfigOptions do NOT come with defaults, it seems important to distinguish these.

cli/cmd/install.go
klingerf: Yes, `panic` would be better. We went through this package and painstakingly scrubbed all of the usages of `log.Fatal` a while back, so I want to be vigilant about it creeping back in.

cli/cmd/inject.go
olix0r: The issue is that we can't have the proxyConfigOptions.validate() validate this field -- it has to be done in injectOptions.validate() and installOptions.validate() independently, so it seemed better to keep that logic common to those structs. I don't really care that much though if you think it's important.

cli/cmd/inject_test.go
olix0r: we check this in another place so it seemed redundant to check everywhere, but i'll create a wrapper that does it one place.

cli/cmd/inject.go
olix0r: I see the value in combining the flag configuration, though...

olix0r:
I ended up addressing the error-handling lint by printing directly to stderr and exiting in those cases.

pkg/k8s/labels.go
```diff
@@ -160,6 +160,9 @@ const (
 	 * Component Names
 	 */
 
+	// ConfigConfigMapName is the name of hte ConfigMap containing the linkerd controller configuration.
+	ConfigConfigMapName = "linkerd-config"
```
olix0r: this isn't actually used?

olix0r: No, it's used. But there's a typo in its comment ;/

pkg/inject/inject.go
```diff
@@ -629,7 +627,8 @@ func (conf *ResourceConfig) proxyResourceRequirements() v1.ResourceRequirements
 }
 
 func (conf *ResourceConfig) proxyDestinationAddr() string {
-	dns := fmt.Sprintf("linkerd-destination.%s.svc.cluster.local", conf.globalConfig.GetLinkerdNamespace())
+	ns := conf.configs.GetGlobal().GetLinkerdNamespace()
+	dns := fmt.Sprintf("linkerd-destination.%s.svc.cluster.local", ns)
```
alpeb: Shouldn't `cluster.local` be replaced with the `identity-trust-domain` option, or are these two separate things?

olix0r: The trust domain is explicitly decoupled from DNS: we continue to connect to "svc.cluster.local" DNS names (until we accept cluster name configurations later) -- but we validate peer identity based on the identity name with the trust-domain suffix. https://github.com/linkerd/linkerd2/pull/2536 introduces peer identity configuration.

PR #2536, opened 2019-03-21
inject: Configure proxies to enable Identity

olix0r:
This change adds a new `linkerd2-proxy-identity` binary to the `proxy`
container image as well as a `linkerd2-proxy-run` entrypoint script.
    
The inject process now sets environment variables on pods to support
identity, including identity names for the destination and identity
services.
    
As the proxy starts, the identity helper creates a key and CSR in a
tmpfs. As the proxy starts it reads these files as well as a
serviceaccount token, and then provisions a certificate from
the Identity controller. The proxy's /ready endpoint will not succeed
until a certificate has been provisioned.
    
The proxy will not participate in identity with services other than the
controllers until the Destination controller is modified to provide
identities via discovery.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -281,7 +281,6 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 				conf.destinationDNSOverride = localhostDNSOverride
 			case identityDeployName:
 				conf.identityDNSOverride = localhostDNSOverride
-			default:
 			}
 		}
 
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -25,7 +25,8 @@ const (
 	// must be in absolute form for the proxy to special-case it.
 	localhostDNSOverride = "localhost."
 
-	controllerPodName = "linkerd-controller"
+	controllerDeployName = "linkerd-controller"
+	identityDeployName   = "linkerd-identity"
 
 	// defaultKeepaliveMs is used in the proxy configuration for remote connections
 	defaultKeepaliveMs = 10000
@@ -45,12 +46,21 @@ const (
 	envDestinationContext         = "LINKERD2_PROXY_DESTINATION_CONTEXT"
 	envDestinationProfileSuffixes = "LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES"
 	envDestinationSvcAddr         = "LINKERD2_PROXY_DESTINATION_SVC_ADDR"
-
-	envIdentityDisabled = "LINKERD2_PROXY_IDENTITY_DISABLED"
-	identityDisabledMsg = "Identity is not yet available"
+	envDestinationSvcName         = "LINKERD2_PROXY_DESTINATION_SVC_NAME"
 
 	// destinationAPIPort is the port exposed by the linkerd-destination service
 	destinationAPIPort = 8086
+
+	envIdentityDisabled     = "LINKERD2_PROXY_IDENTITY_DISABLED"
+	envIdentityDir          = "LINKERD2_PROXY_IDENTITY_DIR"
+	envIdentityLocalName    = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME"
+	envIdentitySvcAddr      = "LINKERD2_PROXY_IDENTITY_SVC_ADDR"
+	envIdentitySvcName      = "LINKERD2_PROXY_IDENTITY_SVC_NAME"
+	envIdentityTokenFile    = "LINKERD2_PROXY_IDENTITY_TOKEN_FILE"
+	envIdentityTrustAnchors = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS"
+
+	identityAPIPort     = 8080
+	identityDisabledMsg = "Identity is not yet available"
 )
 
 var injectableKinds = []string{
@@ -70,15 +80,17 @@ type objMeta struct {
 
 // ResourceConfig contains the parsed information for a given workload
 type ResourceConfig struct {
-	configs                *config.All
-	nsAnnotations          map[string]string
-	meta                   metav1.TypeMeta
-	obj                    runtime.Object
-	workLoadMeta           *metav1.ObjectMeta
-	podMeta                objMeta
-	podLabels              map[string]string
-	podSpec                *v1.PodSpec
+	configs       *config.All
+	nsAnnotations map[string]string
+	meta          metav1.TypeMeta
+	obj           runtime.Object
+	workLoadMeta  *metav1.ObjectMeta
+	podMeta       objMeta
+	podLabels     map[string]string
+	podSpec       *v1.PodSpec
+
 	destinationDNSOverride string
+	identityDNSOverride    string
 	proxyOutboundCapacity  map[string]uint
 }
 
@@ -263,8 +275,14 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 			return err
 		}
 
-		if v.Name == controllerPodName && v.Namespace == conf.configs.GetGlobal().GetLinkerdNamespace() {
-			conf.destinationDNSOverride = localhostDNSOverride
+		if v.Namespace == conf.configs.GetGlobal().GetLinkerdNamespace() {
+			switch v.Name {
+			case controllerDeployName:
+				conf.destinationDNSOverride = localhostDNSOverride
+			case identityDeployName:
+				conf.identityDNSOverride = localhostDNSOverride
+			default:
+			}
 		}
 
 		conf.obj = v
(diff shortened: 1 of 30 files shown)


olix0r:
```
=== PASS: all tests passed
```

test/install_test.go
```diff
@@ -56,17 +56,18 @@ var (
...
 
 		`.* linkerd-controller-.*-.* tap time=".*" level=error msg="\[.*\] encountered an error: rpc error: code = Canceled desc = context canceled"`,
 		`.* linkerd-web-.*-.* linkerd-proxy WARN trust_dns_proto::xfer::dns_exchange failed to associate send_message response to the sender`,
 
 		// prometheus scrape failures of control-plane
 		`.* linkerd-prometheus-.*-.* linkerd-proxy ERR! \[ +\d+.\d+s\] proxy={server=out listen=127\.0\.0\.1:4140 remote=.*} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: Connection refused \(os error 111\) \(address: .*:(3000|999(4|5|6|7|8))\)`,
-		`.* linkerd-prometheus-.*-.* linkerd-proxy ERR! \[ +\d+.\d+s\] proxy={server=out listen=127\.0\.0\.1:4140 remote=.*} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: operation timed out after 300ms`,
+		`.* linkerd-prometheus-.*-.* linkerd-proxy ERR! \[ +\d+.\d+s\] proxy={server=out listen=127\.0\.0\.1:4140 remote=.*} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: operation timed out after .`,
```
olix0r: should be `.*`

olix0r:
@alpeb @ihcsim i'm not particularly confident that i've addressed everything needed for autoinject here.

pkg/inject/inject.go
```diff
@@ -263,8 +275,14 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
...
 
-		if v.Name == controllerPodName && v.Namespace == conf.configs.GetGlobal().GetLinkerdNamespace() {
-			conf.destinationDNSOverride = localhostDNSOverride
+		if v.Namespace == conf.configs.GetGlobal().GetLinkerdNamespace() {
+			switch v.Name {
+			case controllerDeployName:
+				conf.destinationDNSOverride = localhostDNSOverride
+			case identityDeployName:
+				conf.identityDNSOverride = localhostDNSOverride
+			default:
```
klingerf: No need for the empty default on the switch statement here -- you can remove this line.

proxy-identity/main.go
```diff
@@ -0,0 +1,142 @@
...
+// that the wrong directory was specified incorrectly, instead of trying to
+// create or repair the directory. In practice, this directory should be tmpfs
+// so that credentials are not written to disk, so we want to be extra sensitive
+// to an incorrectly specified path.
+//
+// If the key, CSR, and/or Crt paths refer to existing files, it is assumed that
+// multiple instances of this process are running, and an error is returned.
+func checkEndEntityDir(dir string) (string, string, error) {
+	if dir == "" {
+		return "", "", errors.New("No end entity directory specified")
```
alpeb: Capitalized error strings :stuck_out_tongue:

pkg/inject/inject.go
```diff
@@ -427,14 +451,75 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
...
+			Value: "linkerd-identity.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)",
+		},
+		{
+			Name:  envDestinationSvcName,
+			Value: "linkerd-controller.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)",
+		},
+	}...)
+
+	if len(conf.podSpec.Volumes) == 0 {
+		patch.addVolumeRoot()
```
klingerf: Now that you're using this func, you can remove this comment from `pkg/inject/patch.go`:

```
// TODO use with new identity injection
//nolint:unused
```

pkg/inject/inject.go
```diff
@@ -427,14 +451,75 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
...
+	}
+	patch.addVolume(&v1.Volume{
```
klingerf: Same comment here as above for `addVolumeRoot`.

proxy-identity/main.go
```diff
@@ -0,0 +1,142 @@
...
+		return "", "", fmt.Errorf("Not a directory: %s", dir)
+	}
+	// if s.Mode().Perm()&0002 == 0002 {
+	// 	return "", "", "", fmt.Errorf("Must not be world-writeable: %s; got %s", dir, s.Mode().Perm())
+	// }
+
+	keyPath := filepath.Join(dir, "key.p8")
+	if err = checkNotExists(keyPath); err != nil {
+		log.Info(err.Error())
+	}
```
alpeb: if `keyPath` already exists, we're logging the error but not returning it from `checkEndEntityDir()`

pkg/inject/inject.go
```diff
@@ -45,12 +46,21 @@ const (
...
-	identityDisabledMsg = "Identity is not yet available"
+	envDestinationSvcName         = "LINKERD2_PROXY_DESTINATION_SVC_NAME"
 
 	// destinationAPIPort is the port exposed by the linkerd-destination service
 	destinationAPIPort = 8086
+
+	envIdentityDisabled     = "LINKERD2_PROXY_IDENTITY_DISABLED"
+	envIdentityDir          = "LINKERD2_PROXY_IDENTITY_DIR"
+	envIdentityLocalName    = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME"
+	envIdentitySvcAddr      = "LINKERD2_PROXY_IDENTITY_SVC_ADDR"
```
ihcsim: Is this supposed to be consumed by the proxy? I don't see it in the [linkerd-proxy](https://github.com/linkerd/linkerd2-proxy/search?q=LINKERD2_PROXY_IDENTITY_SVC_ADDR&unscoped_q=LINKERD2_PROXY_IDENTITY_SVC_ADDR)  repo.

pkg/inject/inject.go
```diff
@@ -45,12 +46,21 @@ const (
...
+	envIdentitySvcName      = "LINKERD2_PROXY_IDENTITY_SVC_NAME"
```
ihcsim: Likewise, I don't see this in the [proxy source](https://github.com/linkerd/linkerd2-proxy/search?q=LINKERD2_PROXY_IDENTITY_SVC_NAME&unscoped_q=LINKERD2_PROXY_IDENTITY_SVC_NAME). I brought it up because all that other env vars are in src/app/config.rs, except these two.

pkg/inject/inject.go
ihcsim: Never, I think I found [it](https://github.com/linkerd/linkerd2-proxy/blob/2064900a0f89e84ece0bae96e160d5dbee2af848/src/app/config.rs#L789).

ihcsim:
:heart: 
```
bin/linkerd logs | grep -iw "certified identity"
linkerd linkerd-controller-958566d74-chrgf linkerd-proxy INFO [     7.541253s] linkerd2_proxy::app::main Certified identity: linkerd-controller.linkerd.serviceaccount.identity.linkerd.cluster.local
linkerd linkerd-web-8d5c4fcc6-wm72m linkerd-proxy INFO [     7.573498s] linkerd2_proxy::app::main Certified identity: linkerd-web.linkerd.serviceaccount.identity.linkerd.cluster.local
linkerd linkerd-identity-cd7878bc4-txqmj linkerd-proxy INFO [     0.018148s] linkerd2_proxy::app::main Certified identity: linkerd-identity.linkerd.serviceaccount.identity.linkerd.cluster.local
linkerd linkerd-grafana-d9f89c54c-28wxh linkerd-proxy INFO [     7.604094s] linkerd2_proxy::app::main Certified identity: linkerd-grafana.linkerd.serviceaccount.identity.linkerd.cluster.local
linkerd linkerd-prometheus-7b5c6b6884-pb6sf linkerd-proxy INFO [     0.014554s] linkerd2_proxy::app::main Certified identity: linkerd-prometheus.linkerd.serviceaccount.identity.linkerd.cluster.local
linkerd linkerd-proxy-injector-67bf9bc645-qxx5k linkerd-proxy INFO [     6.058457s] linkerd2_proxy::app::main Certified identity: linkerd-proxy-injector.linkerd.serviceaccount.identity.linkerd.cluster.local
```

proxy-identity/main.go
olix0r: This initially failed, but if the proxy restarts for any reason we need to handle the case that the identity already exists gracefully, and it seems better to just reuse the credentials. Though, it still seems log-worthy in case this is unexpected.

alpeb: @olix0r if that's the case then could you please update `checkEndEntityDir()`'s godoc?

alpeb:
One last post-merge comment :grimacing: :
Currently the proxy-injector uses a self-signed certificate. Would there be any value for it to ask the identity service for a cert?

olix0r:
@alpeb No! ;)  I think we want to keep the identity service limited to mesh identity and not (yet) expand it it to address concerns other than full-mesh communication. In 2.4, I'd like to change proxy-injector to not have any knowledge of its certs at all and make that exclusively an install-time thing (so if users want to provision certs another way, they can -- but more importantly, so that cluster admins don't have to delegate webhook creation to the control plane.

proxy-identity/main.go
olix0r: I changed the log message to be clearer; and I'll update the godoc in  #2537

alpeb:
> In fact, the Identity service isn't even currently capable of providing certs that kubernetes would be able to authenticate! We explicitly decouple our certificate names from DNS, using a specific naming scheme. It would be ~impossible to configure Kubernetes to respect our mesh TLS validation rules.

Oh that's true... thanks for the detailed explanation!

PR #2538, opened 2019-03-21
install: Create linkerd-config before pods

olix0r:
Because the linkerd-config resource is created after pods that require
it, they can be started before the files are mounted, causing the pods
to restart integration tests to fail.

If we extract the config into its own template file, it can be inserted
before pods are created.

### DIFF
--- a/chart/templates/config.yaml
+++ b/chart/templates/config.yaml
@@ -1,0 +1,21 @@
+{{with .Values -}}
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: linkerd-config
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: controller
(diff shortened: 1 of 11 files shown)


alpeb:
Ah wait, an integration test failed on me. Let me check.

alpeb:
Running CLI inject with no options leads to errors because we execute `options.validate()` with empty `injectOptions` which fails the first few checks. How about doing that _after_ `options.fetchConfigsOrDefault()`?

olix0r:
Ah, I think I have a fix for this in #2536 I can pull in here.

chart/templates/config.yaml
```diff
@@ -0,0 +1,21 @@
+{{with .Values -}}
```
klingerf: 🎉

Issue #2534, opened 2019-03-21
Disable external profiles by default


olix0r:
Currently, profiles are enabled for all names; but, practically, this puts undue stress on the destination service when, in the majority of cases, the lookups will never yield results.

For now, let's change the default configuration so that this is not enabled. Ideally, I think users should explicitly _enable_ external profiles for specific domains, rather than whitelisting for all. I think it's okay that this is not a backwards-identical change.

adleong:
@dadjeibaah did some testing and found that the destination service was able to handle 10000 concurrent profile requests without really breaking a sweat.  So assuming that the proxy respects the outbound router capacity of 10000, I think we can leave external profiles enabled.

@dadjeibaah can you provide more details?

grampelberg:
From my perspective, I'm happy disabling external names for now and tackling this as part of #2191.

dadjeibaah:
Yea, the destination service can handle 10,000 concurrent profile requests with no problem. So, we can keep querying external profiles.

olix0r:
I still feel pretty strongly that we follow through with this.

PR #2547, opened 2019-03-23
Convert CLI inject proxy options to annotations

### LINKED ISSUE #2447: CLI Inject With Proxy Config Overrides Annotations

ihcsim:
The PR introduces code to convert CLI inject proxy options into config annotations. The annotations ensure that these configs are persisted in subsequent resource updates. 

Fixes #2447 

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -341,8 +341,8 @@ func (options *injectOptions) overrideConfigs(configs *config.All, overrideAnnot
 		overrideAnnotations[k8s.ProxyLogLevelAnnotation] = options.proxyLogLevel
 	}
 
+	configs.Proxy.DisableExternalProfiles = options.disableExternalProfiles
 	if options.disableExternalProfiles {
-		configs.Proxy.DisableExternalProfiles = true
 		overrideAnnotations[k8s.ProxyDisableExternalProfilesAnnotation] = "true"
 	}
 
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -318,10 +318,19 @@ func (options *injectOptions) overrideConfigs(configs *config.All, overrideAnnot
 		overrideAnnotations[k8s.ProxyOutboundPortAnnotation] = parsePort(configs.Proxy.OutboundPort)
 	}
 
-	if options.dockerRegistry != "" {
-		configs.Proxy.ProxyImage.ImageName = registryOverride(configs.Proxy.ProxyImage.ImageName, options.dockerRegistry)
-		configs.Proxy.ProxyInitImage.ImageName = registryOverride(configs.Proxy.ProxyInitImage.ImageName, options.dockerRegistry)
+	if options.proxyImage != "" {
+		configs.Proxy.ProxyImage.ImageName = options.proxyImage
+		if options.dockerRegistry != "" {
+			configs.Proxy.ProxyImage.ImageName = registryOverride(configs.Proxy.ProxyImage.ImageName, options.dockerRegistry)
+		}
 		overrideAnnotations[k8s.ProxyImageAnnotation] = configs.Proxy.ProxyImage.ImageName
+	}
+
+	if options.initImage != "" {
+		configs.Proxy.ProxyInitImage.ImageName = options.initImage
+		if options.dockerRegistry != "" {
+			configs.Proxy.ProxyInitImage.ImageName = registryOverride(configs.Proxy.ProxyInitImage.ImageName, options.dockerRegistry)
+		}
 		overrideAnnotations[k8s.ProxyInitImageAnnotation] = configs.Proxy.ProxyInitImage.ImageName
 	}
 
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
 
 	jsonpatch "github.com/evanphx/json-patch"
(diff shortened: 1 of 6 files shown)


cli/cmd/inject.go
```diff
@@ -278,71 +287,109 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
...
 	if options.proxyInboundPort != 0 {
 		configs.Proxy.InboundPort = toPort(options.proxyInboundPort)
+		overrideAnnotations[k8s.ProxyInboundPortAnnotation] = parsePort(configs.Proxy.InboundPort)
 	}
 	if options.proxyOutboundPort != 0 {
 		configs.Proxy.OutboundPort = toPort(options.proxyOutboundPort)
+		overrideAnnotations[k8s.ProxyOutboundPortAnnotation] = parsePort(configs.Proxy.OutboundPort)
 	}
 
-	if options.dockerRegistry != "" {
```
ihcsim: We still need to check for image name even if the registry option isn't provided.

cli/cmd/inject.go
```diff
@@ -278,71 +287,109 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
...
 		configs.Proxy.ProxyUid = options.proxyUID
+		overrideAnnotations[k8s.ProxyUIDAnnotation] = strconv.FormatInt(options.proxyUID, 10)
 	}
 
 	if options.proxyLogLevel != "" {
 		configs.Proxy.LogLevel = &config.LogLevel{Level: options.proxyLogLevel}
+		overrideAnnotations[k8s.ProxyLogLevelAnnotation] = options.proxyLogLevel
 	}
 
+	configs.Proxy.DisableExternalProfiles = options.disableExternalProfiles
```
ihcsim: Track this even if it's `false`, so that the `LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES` env var will be assign the correct value.

alpeb: Do you mind adding that comment in the code?

cli/cmd/inject.go
```diff
@@ -89,9 +88,14 @@ sub-folders, or coming from stdin.`,
...
 			}
+			overrideAnnotations := map[string]string{}
+			options.overrideConfigs(configs, overrideAnnotations)
 
-			options.overrideConfigs(configs)
-			exitCode := uninjectAndInject(in, stderr, stdout, configs)
+			transformer := &resourceTransformerInject{
+				configs:             configs,
+				overrideAnnotations: overrideAnnotations,
+			}
```
alpeb: How about defining this earlier, and have `overridedConfigs()` receive it as its sole argument?

cli/cmd/inject.go
klingerf: Hmm, yeah, I'm a bit confused as to why it's necessary to set this if `options.disableExternalProfiles` is false (the default). I'd expect this field to only be set if `options.disableExternalProfiles` is true.

olix0r: note that there's an open P0 to change this

klingerf: Yep, #2534 is tracking that. But all the same, I think we should only set the annotation if the flag is set to the non-default value (whatever that may be).

ihcsim: @klingerf This line only stores the value in the config struct, which is needed in the shared lib to decide on the value of `LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES`. The annotation is added in the next line, only if the flag has a non-default value.

ihcsim:
> I did find a use case that might need to be addressed.
> Say we have a POD annotated with `config.linkerd.io/proxy-memory-limit: 10Mi`
> Then we inject it with `inject --proxy-memory-limit 20Mi`.
> The annotation will get updated correctly, but the memory limit in the sidecar will get set to the old `10Mi`, because `getOverride()` only looks at the annotations from parsing the payload, disregarding the map of overriding annotations that `cli/cmd/inject.go` created.
> First thing that comes to mind to address this would be to augment `ResourceConfig` to include that map (that `cli/cmd/inject.go` would pass when calling `NewResourceConfig`), for `getOverride` to take into account.

@alpeb My thought is that the payload that is parsed by `getOverride()` will contain the correct annotations. It looks like the flag-to-annotation needs to happen sooner, before calling `GetPatch()`.

cli/cmd/inject.go
ihcsim: I prefer to keep the arguments of `overrideConfigs()` to config data. A function that reads `overrideConfigs(someKindOfTransformer)` sounds like we are overriding the transformer's config, which it isn't because the transformer has no configs of its own, and it's more a YAML-parsing receiver, than a config receiver.

ihcsim:
@klingerf @alpeb I've updated the code to support the scenario @alpeb described [above](https://github.com/linkerd/linkerd2/pull/2547#pullrequestreview-218554218). Please take a look. Thanks.

pkg/inject/inject.go
```diff
@@ -607,6 +620,17 @@ func (conf *ResourceConfig) injectObjectMeta(patch *Patch) {
...
+
+	for k, v := range conf.pod.annotations {
+		patch.addPodAnnotation(k, v)
+	}
+
+	// append any additional pod annotations to the pod's meta.
+	// for e.g., annotations that were converted from CLI inject options.
+	for annotation, value := range conf.pod.annotations {
+		conf.pod.Meta.Annotations[annotation] = value
+	}
```
alpeb: Seems to me you could merge these two loops

cli/cmd/inject.go
```diff
@@ -281,72 +293,115 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
 	return api.Config(context.Background(), &public.Empty{})
 }
 
-// overrideConfigs uses command-line overrides to update the provided configs
-func (options *injectOptions) overrideConfigs(configs *config.All) {
+// overrideConfigs uses command-line overrides to update the provided configs.
+// the overrideAnnotations map keeps track of which configs are overridden, by
+// storing the corresponding annotations and values.
+func (options *injectOptions) overrideConfigs(configs *config.All, overrideAnnotations map[string]string) {
```
olix0r: you might as well pull in my fix for #2551 here https://github.com/linkerd/linkerd2/blob/e8db3ce6fefd6b9b079c9bbc2700a67c01513e84/cli/cmd/inject.go#L286-L290

klingerf: Update: we're going to remove the `--linkerd-cni-enabled` inject flag, and instead make it install-only, so you can ignore my comment here.

Issue #2499, opened 2019-03-14
Namespace-level Configs Override Annotations


ihcsim:
### What problem are you trying to solve?
Extend the configs override annotations to work at the namespace level. This will allow proxy configs to be overridden for all workloads in a namespace, so that the users won't have to repeat the annotations on each workload.

Following our discussion in #2497, we've decided to exclude this feature from that PR, until we see a realistic use for it.

### How should the problem be solved?
Update the [`conf.getOverride()` method](https://github.com/linkerd/linkerd2/blob/468ad118f224bda78f0498c07ccec76bef0dd277/pkg/inject/inject.go#L544) to look up any namespace annotations, if no pod-level annotations are present.

### How would users interact with this feature?
Specifying the config annotations in the namespace's metadata.

PR #2555, opened 2019-03-25
Promote the shared injection check to the CLI and webhook

ihcsim:
This PR refactors the `shouldInject` logic by promoting it from the shared `pkg/inject` library to the CLI and webhook.

Performing this check earlier helps to separate the specialized logic to the CLI and webhook. Subsequent modification to this check to support configs override of existing meshed workloads will be confined to the relevant component.

This PR addresses point 1 to 3 of our [previous discussion on configs override in 2500](https://github.com/linkerd/linkerd2/pull/2500#issuecomment-474182920).

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -119,7 +119,7 @@ func uninjectAndInject(inputs []io.Reader, errWriter, outWriter io.Writer, trans
 }
 
 func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Report, error) {
-	conf := inject.NewResourceConfig(rt.configs)
+	conf := inject.NewResourceConfig(rt.configs, inject.OriginCLI)
 	if len(rt.proxyOutboundCapacity) > 0 {
 		conf = conf.WithProxyOutboundCapacity(rt.proxyOutboundCapacity)
 	}
@@ -156,11 +156,7 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	if patchJSON == nil {
 		return bytes, reports, nil
 	}
-	// TODO: refactor GetPatch() for it to return just one report item
-	if len(reports) > 0 {
-		r := reports[0]
-		log.Infof("patch generated for: %s", r.ResName())
-	}
+	log.Infof("patch generated for: %s", report.ResName())
 	log.Debugf("patch: %s", patchJSON)
 	patch, err := jsonpatch.DecodePatch(patchJSON)
 	if err != nil {
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -123,13 +123,15 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	if len(rt.proxyOutboundCapacity) > 0 {
 		conf = conf.WithProxyOutboundCapacity(rt.proxyOutboundCapacity)
 	}
-	nonEmpty, err := conf.ParseMeta(bytes)
+
+	report, err := conf.ParseMetaAndYAML(bytes)
 	if err != nil {
 		return nil, nil, err
 	}
-	if !nonEmpty {
-		r := inject.Report{UnsupportedResource: true}
-		return bytes, []inject.Report{r}, nil
+	reports := []inject.Report{*report}
+
+	if !report.Injectable() {
+		return bytes, reports, nil
 	}
 
 	conf.AppendPodAnnotations(map[string]string{
@@ -139,7 +141,7 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 		conf.AppendPodAnnotations(rt.overrideAnnotations)
 	}
 
-	p, reports, err := conf.GetPatch(bytes, inject.ShouldInjectCLI)
+	p, err := conf.GetPatch(bytes)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/cli/cmd/uninject.go
+++ b/cli/cmd/uninject.go
@@ -63,7 +63,7 @@ sub-folders, or coming from stdin.`,
 }
 
 func (rt resourceTransformerUninject) transform(bytes []byte) ([]byte, []inject.Report, error) {
-	conf := inject.NewResourceConfig(rt.configs)
+	conf := inject.NewResourceConfig(rt.configs, inject.OriginWebhook)
 
 	report, err := conf.ParseMetaAndYAML(bytes)
 	if err != nil {
--- a/cli/cmd/uninject.go
+++ b/cli/cmd/uninject.go
@@ -65,7 +65,7 @@ sub-folders, or coming from stdin.`,
 func (rt resourceTransformerUninject) transform(bytes []byte) ([]byte, []inject.Report, error) {
 	conf := inject.NewResourceConfig(rt.configs)
 
-	report, err := conf.ParseMetaAndYaml(bytes)
+	report, err := conf.ParseMetaAndYAML(bytes)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -91,7 +91,7 @@ func (resourceTransformerUninject) generateReport(reports []inject.Report, outpu
 	output.Write([]byte("\n"))
 
 	for _, r := range reports {
-		if r.Sidecar {
+		if r.Uninjected.Proxy || r.Uninjected.ProxyInit {
 			output.Write([]byte(fmt.Sprintf("%s \"%s\" uninjected\n", r.Kind, r.Name)))
 		} else {
 			if r.Kind != "" {
--- a/controller/proxy-injector/webhook_test.go
+++ b/controller/proxy-injector/webhook_test.go
@@ -41,12 +41,12 @@ var (
 
 func confNsEnabled() *inject.ResourceConfig {
 	return inject.
-		NewResourceConfig(configs).
+		NewResourceConfig(configs, inject.OriginWebhook).
 		WithNsAnnotations(map[string]string{pkgK8s.ProxyInjectAnnotation: pkgK8s.ProxyInjectEnabled})
 }
 
 func confNsDisabled() *inject.ResourceConfig {
-	return inject.NewResourceConfig(configs).WithNsAnnotations(map[string]string{})
+	return inject.NewResourceConfig(configs, inject.OriginWebhook).WithNsAnnotations(map[string]string{})
 }
 
 func TestGetPatch(t *testing.T) {
--- a/controller/proxy-injector/webhook_test.go
+++ b/controller/proxy-injector/webhook_test.go
@@ -80,28 +80,10 @@ func TestGetPatch(t *testing.T) {
 				expected: true,
 			},
 			{
-				filename: "deployment-inject-disabled.yaml",
-				ns:       nsEnabled,
-				conf:     confNsEnabled(),
-				expected: false,
-			},
-			{
-				filename: "deployment-inject-empty.yaml",
-				ns:       nsDisabled,
-				conf:     confNsDisabled(),
-				expected: false,
-			},
-			{
 				filename: "deployment-inject-enabled.yaml",
 				ns:       nsDisabled,
 				conf:     confNsDisabled(),
 				expected: true,
-			},
-			{
-				filename: "deployment-inject-disabled.yaml",
-				ns:       nsDisabled,
-				conf:     confNsDisabled(),
-				expected: false,
 			},
 		}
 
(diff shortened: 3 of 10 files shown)


cli/cmd/uninject.go
```diff
@@ -65,7 +65,7 @@ sub-folders, or coming from stdin.`,
 func (rt resourceTransformerUninject) transform(bytes []byte) ([]byte, []inject.Report, error) {
 	conf := inject.NewResourceConfig(rt.configs)
 
-	report, err := conf.ParseMetaAndYaml(bytes)
+	report, err := conf.ParseMetaAndYAML(bytes)
```
ihcsim: Use uppercase for acronym. Refer [golang wiki](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

cli/cmd/inject.go
```diff
@@ -119,15 +119,18 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
...
-	nonEmpty, err := conf.ParseMeta(bytes)
+
+	report, err := conf.ParseMetaAndYAML(bytes)
 	if err != nil {
 		return nil, nil, err
 	}
-	if !nonEmpty {
-		r := inject.Report{UnsupportedResource: true}
-		return bytes, []inject.Report{r}, nil
+	reports := []inject.Report{*report}
```
alpeb: It appears we can deal with a single `report` (instead of having to always wrap in an single-item array) everywhere except in the `inject_utils.go` functions. That got evident in on of my branches, so I decided to tackle that later in a separate PR, but you can do that here as well if you find it's a natural change :wink:

pkg/inject/report.go
```diff
@@ -47,7 +60,7 @@ func (r Report) Injectable() bool {
 
 // update updates the report for the provided resource conf.
 func (r *Report) update(conf *ResourceConfig) {
-	r.InjectDisabled = conf.pod.Meta.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled
+	r.InjectDisabled = conf.pod.meta.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled || conf.nsAnnotations[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled
```
alpeb: So this means that if a namespace is tagged with `inject:disabled` then any pod injection will be disabled even if they have `inject:enabled` explicitly set?
We could think the annotation at the namespace level being set by an admin with higher access, who would want to explicitly enable/disable the possibility of injection by less-privilege users, in which case this behavior would make sense.
OTOH, we could also think that we'd like to leave the door open for more flexibility by being able to override everything at the pod-level, including the ability to inject.
So I don't know which way we should go...

pkg/inject/inject.go
```diff
@@ -148,60 +146,38 @@ func (conf *ResourceConfig) YamlMarshalObj() ([]byte, error) {
...
-	return &r, conf.parse(bytes)
-}
 
-// ParseMeta extracts metadata from bytes.
-// It returns false if the workload's payload is empty
-func (conf *ResourceConfig) ParseMeta(bytes []byte) (bool, error) {
-	if err := yaml.Unmarshal(bytes, &conf.workload.metaType); err != nil {
-		return false, err
-	}
-	if err := yaml.Unmarshal(bytes, &conf.pod); err != nil {
```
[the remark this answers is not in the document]
ihcsim: Hmm.. that's surprising. I'd think [`parse()`](https://github.com/linkerd/linkerd2/blob/b00890144920a2166225785c3f53b5bf67e57671/pkg/inject/inject.go#L333-L340) is able to pick it up. Will investigate.

pkg/inject/report.go
ihcsim: If pod has `inject: enabled`, it will take precedence over any namespace level annotation. See [this test](https://github.com/linkerd/linkerd2/blob/b00890144920a2166225785c3f53b5bf67e57671/pkg/inject/report_test.go#L91-L100), for `report.Injectable()`. Let me know if you think there is a :beetle:  here.

alpeb: You meant [this test](https://github.com/linkerd/linkerd2/blob/b00890144920a2166225785c3f53b5bf67e57671/pkg/inject/report_test.go#L78-L89) instead, right?
It appears that `nsAnnotations` is not being used in that test suite :eyes:

pkg/inject/inject.go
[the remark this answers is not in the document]
alpeb: My bad, I was using a bad yaml in my test pod. It's working as expected :+1:

controller/proxy-injector/webhook_test.go
```diff
@@ -79,30 +79,12 @@ func TestGetPatch(t *testing.T) {
...
-				ns:       nsEnabled,
-				conf:     confNsEnabled(),
-				expected: false,
-			},
-			{
-				filename: "deployment-inject-empty.yaml",
-				ns:       nsDisabled,
-				conf:     confNsDisabled(),
-				expected: false,
-			},
```
alpeb: Is it enough to add the additional `Injectable()` test below to be able to re-enable these tests, or is it more complicated than that? If so we could track that as a separate item for the 2.3 test sprint to expedite this PR :zap:

ihcsim: These two test cases are no longer useful, as the results will always be `true` because `GetPatch()` no longer checks if injection should happen. Some of these tests need to be revisited, as they aren't testing anything webhook related, but not in this PR.

cli/cmd/inject.go
```diff
@@ -123,13 +123,15 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	if len(rt.proxyOutboundCapacity) > 0 {
 		conf = conf.WithProxyOutboundCapacity(rt.proxyOutboundCapacity)
 	}
-	nonEmpty, err := conf.ParseMeta(bytes)
+
+	report, err := conf.ParseMetaAndYAML(bytes)
```
klingerf: Now that this call returns a singular report, you can remove the `len(reports)` check on line 160 in this file, similar to the change you're making in webhook.go:

```diff
diff --git a/cli/cmd/inject.go b/cli/cmd/inject.go
index 18de30a5..c4cf0dcc 100644
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -156,11 +156,7 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	if patchJSON == nil {
 		return bytes, reports, nil
 	}
-	// TODO: refactor GetPatch() for it to return just one report item
-	if len(reports) > 0 {
-		r := reports[0]
-		log.Infof("patch generated for: %s", r.ResName())
-	}
+	log.Infof("patch generated for: %s", report.ResName())
 	log.Debugf("patch: %s", patchJSON)
 	patch, err := jsonpatch.DecodePatch(patchJSON)
 	if err != nil {
```

pkg/inject/report.go
```diff
@@ -19,23 +19,42 @@ type Report struct {
...
 		}
 	}
 
-	return Report{
+	report := &Report{
 		Kind: strings.ToLower(conf.workload.metaType.Kind),
 		Name: name,
 	}
+	if conf.pod.meta != nil && conf.pod.spec != nil {
+		report.update(conf)
```
siggy: i really like how we're only calling `report.update(` from `newReport()`. since it's side-effect-y, mind just inlining `update` into this location, and removing the function altogether?

pkg/inject/report_test.go
```diff
@@ -0,0 +1,145 @@
...
+	for i, testCase := range testCases {
+		testCase := testCase
+		t.Run(fmt.Sprintf("test case #%d", i), func(t *testing.T) {
+			resourceConfig := &ResourceConfig{}
+			resourceConfig.WithNsAnnotations(testCase.nsAnnotations)
+			resourceConfig.pod.spec = testCase.podSpec
+			resourceConfig.pod.meta = testCase.podMeta
+
+			report := newReport(resourceConfig)
+			report.update(resourceConfig)
```
siggy: i think you can remove this line, now that we're calling `update` in `newReport`?

pkg/inject/report.go
```diff
@@ -51,7 +70,7 @@ func (r *Report) Injectable() bool {
 
 // update updates the report for the provided resource conf.
 func (r *Report) update(conf *ResourceConfig) {
-	r.InjectDisabled = conf.pod.Meta.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled
+	r.InjectDisabled = conf.pod.meta.GetAnnotations()[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled || (conf.nsAnnotations[k8s.ProxyInjectAnnotation] == k8s.ProxyInjectDisabled && conf.pod.meta.GetAnnotations()[k8s.ProxyInjectAnnotation] != k8s.ProxyInjectEnabled)
```
klingerf: This is a really long line now. And I think there's an issue with the logic here, since I'm seeing the auto-injector from this branch inject pods in namespaces without the `linkerd.io/inject: enabled` annotation. It is correctly skipping pods with the `linkerd.io/inject: disabled` annotation. The issue only arises when the namespace has no `linkerd.io/inject` annotation.

ihcsim: Yes, you are right about the regression, when the `linkerd.io/inject` annotation isn't present. Sorry about the long one liner. I will break it up into a function or something.

ihcsim: The issue is that this conditional didn't account for the different implication in the absence of the `linkerd.io/inject` annotation in the webhook and the CLI. In the webhook, it implies an opt-out, while with `linkerd inject`, it implies an opt-in. I have fixed it in the `HEAD` of this branch.

pkg/inject/inject.go
```diff
@@ -61,6 +61,18 @@ const (
...
+	// YAML comes from the CLI
+	OriginCLI Origin = iota
+
+	// OriginWebhook is the value of the ResourceConfig's 'origin' field if the input
+	// YAML comes from the CLI
+	OriginWebhook
+
+	// OriginUnknown is the value of the ResourceConfig's 'origin' field if the
+	// input YAML comes from an unknown source
+	OriginUnknown
```
alpeb: Could you group these guys under a separate `const` block, for better clarity?

pkg/inject/inject.go
```diff
@@ -151,6 +169,14 @@ func (conf *ResourceConfig) AppendPodAnnotations(annotations map[string]string)
 	}
 }
 
+// CreatedByCLI returns true if the pod is annotated with linkerd.io/created-by: linkerd/cli
+func (conf *ResourceConfig) CreatedByCLI() bool {
+	if conf.pod.annotations == nil {
+		return false
+	}
+	return strings.Contains(conf.pod.annotations[k8s.CreatedByAnnotation], "")
```
alpeb: I believe this will return true iff ` conf.pod.annotations[k8s.CreatedByAnnotation]` is defined

ihcsim: Ah..sorry, that's obsolete; will remove.

pkg/inject/report.go
```diff
@@ -87,3 +83,35 @@ func checkUDPPorts(t *v1.PodSpec) bool {
...
+	// enabled   | disabled | no       | true
+	// disabled  | enabled  | yes      | false
+	// ""        | enabled  | yes      | false
+	// disabled  | disabled | no       | true
+	// ""        | disabled | no       | true
+	// disabled  | ""       | no       | true
+	// ""        | ""       | no       | true
+	//
+	// for CLI, only the 'disabled' annotation is taken into consideration
+	// for its opt-out effect
```
alpeb: Thanks for clarifying all this. To be even more extra super clear, I'd only prefix that table spelling out that it only applies to webhook injection.

pkg/inject/report.go
```diff
@@ -68,3 +83,35 @@ func checkUDPPorts(t *v1.PodSpec) bool {
...
```
siggy: to confirm my own understanding, and for others going forward, can you verify this updated table is correct, and then include it?
```
	// truth table of the effects of the inject annotation on the webhook and cli:
	//
	// origin  | namespace | pod      | inject?  | return
	// ------- | --------- | -------- | -------- | ------
	// webhook | enabled   | enabled  | yes      | false
	// webhook | enabled   | ""       | yes      | false
	// webhook | enabled   | disabled | no       | true
	// webhook | disabled  | enabled  | yes      | false
	// webhook | ""        | enabled  | yes      | false
	// webhook | disabled  | disabled | no       | true
	// webhook | ""        | disabled | no       | true
	// webhook | disabled  | ""       | no       | true
	// webhook | ""        | ""       | no       | true
	// cli     | n/a       | enabled  | yes      | false
	// cli     | n/a       | ""       | yes      | false
	// cli     | n/a       | disabled | no       | true
	//
	// for CLI, only the 'disabled' annotation is taken into consideration
	// for its opt-out effect
```

ihcsim: Thanks @siggy. That extra column makes it even clearer :+1:. And LGTM.

ihcsim:
> This change LGTM. I tested most of the permutations laid out in #2569. One thing to confirm: `config.linkerd.io` annotations are not honored at the namespace-level, correct?

Correct. We moved in out of 2.3 into https://github.com/linkerd/linkerd2/issues/2499.

PR #2562, opened 2019-03-26
install: Create auto-inject configuration

olix0r:
When reading a Linkerd configuration, we cannot determine whether autoinject should be configured.

This change adds autoinject configuration to the global config structure. Currently, this configuration is effectively boolean, determined by the presence of an empty value (versus a null).


olix0r:
depends on https://github.com/linkerd/linkerd2/pull/2561

proto/config/config.proto
```diff
@@ -16,11 +16,13 @@ message Global {
...
 
-  // null indicates TLS is disabled.
-  // Otherwise, a non-null struct indicates the equivalence
-  // of --tls=optional.
+  // If present, configures identity.
   IdentityContext identity_context = 4;
 
+  // If present, indicates that the Mutating Webhook Admission Controller should
+  // be configured to automatically inject proxies.
+  AutoinjectContext autoinject_context = 6;
```
klingerf: TIOLI I recommend capitalizing inject here:

```suggestion
  AutoInjectContext auto_inject_context = 6;
```

Issue #2176, opened 2019-01-30
Add MutatingWebhookConfiguration objects to install YAML


klingerf:
Now that #2163 has merged, the way in which the proxy-injector's MutatingWebhookConfig is provisioned is much simpler. It now works as follows:

* `linkerd install --proxy-auto-inject` emits the config for the proxy-injector pod
* Once the proxy-injector pod starts, it creates a new CA bundle in memory
* It then calls the Kubernetes API to create or update the MWC, which contains the CA bundle

This means that our install YAML does not include the MWC itself, since it's created programmatically after the proxy-injector pod starts. Ideally, however, the install YAML would include all objects that are actually installed, including the MWC. But this would require the CLI to create the bundle at install time and provide it as a secret to the linkerd-proxy-injector pod.

klingerf:
Relates to #2164.

adleong:
I think the way this should work is that at install time, Linkerd install would generate a root CA and key, create the MWC with the CA (trust root) and save the key into a secret.  When a webhook instance starts up, it should load the CA key from the secret, generate a serving key and CSR, and use the CA key to generate a serving cert.

The same approach should be used for validating webhooks and for api extensions.  This is a prerequisite to running any of these in HA mode.

Does this seem right?

adleong:
How would this interact with `linkerd upgrade`?  For upgrade to give consistent results, the root CA would have to be stored in the linkerd config configmap, or `upgrade` would need to read it from the existing MWC, right?

alpeb:
@adleong Your reasoning sounds right to me. As for `linkerd upgrade`, it appears the approach is similar to what we already have for Identity (with the difference that we have the option to not include the CA in the configmap if it's already in the MWC, like you pointed out). There's an ongoing discussion in #2679 about upgrade and secrets that might affect this issue as well.

ihcsim:
I agree that in HA mode the root CA needs to be persisted. I'm not sure about persisting the key though, because we have a [`tls` package ](https://github.com/linkerd/linkerd2/blob/1c0f1477189d7f849fc52585b3b22d852da5a737/pkg/tls/ca.go#L164) that can generate in-memory cert and private key for the CA. IMO, unless necessary, the less we write out to config maps/secrets etc., the better it is for things like upgrade.

> or upgrade would need to read it from the existing MWC

If other components will be sharing this CA, it cannot be saved in the MWC, because the MWC gets re-created everytime the webhook is restarted. See #2425.

[...]

PR #2564, opened 2019-03-26
WIP: cli: Introduce an upgrade command

olix0r:


cli/cmd/upgrade.go
```diff
@@ -0,0 +1,207 @@
...
+		// Controller configuration:
+		Namespace:              controlPlaneNamespace,
+		UUID:                   configs.GetGlobal().GetInstallationUuid(),
+		ControllerLogLevel:     options.controllerLogLevel,
+		PrometheusLogLevel:     toPromLogLevel(options.controllerLogLevel),
+		ControllerReplicas:     controllerReplicas,
+		ControllerUID:          options.controllerUID,
+		EnableHA:               options.highAvailability,
+		EnableH2Upgrade:        !options.disableH2Upgrade,
+		NoInitContainer:        configs.GetGlobal().GetCniEnabled(),
```
alpeb: Some of these options (ControllerLogLevel, EnableHA, EnableH2Upgrade) apparently will also have to be added into the Global config to avoid being overwritten during upgrade.

olix0r: I'm _slightly_ uneasy about encoding EnableHA in the config, since that _implies_ a bunch of policy that's inferred elsewhere. It would be ideal if we could just read in the existing resource requests/limits, for instance, instead of overriding.  Same goes for the number of replicas...

We're probably going to have to cut some corners to get upgrade to a useable state for the next edge; so we may want to require that some of these flags be re-set when upgrading...

Alternatively, how would anyone ever _disable_ HA during an upgrade, for instance?

olix0r: I think we should remove `EnableHA` from the `Values` struct and replace it with types corresponding to the various templated values; and then we should satisfy those Values types from config or k8s metadata...  This allows us to at least preserve resource limits/requests and replica counts without storing them in the config (where they are not a source of truth).  Things like EnableH2Upgrade probably need to go into a DestinationContext config...

cli/cmd/upgrade.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// Fetch the existing cluster configs or exit.
+	configs := &pb.All{Global: &pb.Global{}, Proxy: &pb.Proxy{}}
+	configMap, err := k.CoreV1().ConfigMaps(controlPlaneNamespace).
+		Get(k8s.ConfigConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+	uj := jsonpb.Unmarshaler{}
+	if err := uj.Unmarshal(strings.NewReader(configMap.Data["global"]), configs.Global); err != nil {
```
siggy: would like to see these "global" and "proxy" strings as constants and templatized in `config.yaml`.

cli/cmd/upgrade.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, err
+	}
+
+	return kubernetes.NewForConfig(api.Config)
+}
+
+// fetchInstallValuesFromCluster checks the kubernetes API to fetch an existing
+// linkerd configuration.
+//
+// This bypasses the public API so that we can access secrets and validate permissions.
```
siggy: we should consider modifying `fetchConfigsOrDefault` to do this as well. i'd prefer we have a single code path for retrieving configs.

cli/cmd/upgrade.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, nil, err
+	}
+	if err := uj.Unmarshal(strings.NewReader(configMap.Data["proxy"]), configs.Proxy); err != nil {
+		return nil, nil, err
+	}
+
+	// Override the configs from the command-line flags.
+	options.overrideConfigs(configs)
+	j := jsonpb.Marshaler{EmitDefaults: true}
+	globalJSON, err := j.MarshalToString(configs.GetGlobal())
```
siggy: marshaling the config to global and proxy json can be factored out with `install.go`.

going a step further, a lot of the construction of `installValues` can be factored out as well, including `installIdentityValues`.

pkg/k8s/labels.go
```diff
@@ -173,6 +173,10 @@ const (
 	// volume mounted into each proxy to store identity credentials.
 	IdentityEndEntityVolumeName = "linkerd-identity-end-entity"
 
+	// IdentityIssuerSecretName is the name assigned the temporary end-entity
+	// volume mounted into each proxy to store identity credentials.
+	IdentityIssuerSecretName = "linkerd-identity-issuer"
```
siggy: templatize this string in `identity.yaml`

cli/cmd/upgrade.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, nil, err
+	}
+
+	secret, err := k.CoreV1().Secrets(controlPlaneNamespace).
+		Get(k8s.IdentityIssuerSecretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	keyPEM := string(secret.Data["key.pem"])
```
siggy: templatize `key.pem` and `crt.pem` in `identity.yaml`

cli/cmd/upgrade.go
olix0r: The upgrade logic is different than the other uses.

pkg/k8s/labels.go
olix0r: We're not really consistent about this, and I don't feel it buys us anything.

cli/cmd/upgrade.go
olix0r: i don't think it's worth templatizing but i'll const them at least

cli/cmd/install.go
```diff
@@ -622,42 +624,48 @@ func (options *installOptions) proxyConfig() *pb.Proxy {
...
 		}
 
-		return false, err
+		fmt.Fprintln(os.Stderr, "Unable to build a Kubernetes client to check for configuration. If this expected, use the --ignore-cluster flag.")
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err)
+		os.Exit(1)
 	}
 
-	return true, nil
+	fmt.Fprintln(os.Stderr, "You are already running a control plane. If you would like to ignore its configuration, use the --ignore-cluster flag.")
```
grampelberg: Could this message be updated to something more like:

```
Linkerd has already been installed on your cluster in the linkerd namespace. Please run upgrade if you'd like to update this installation. Otherwise, use the --ignore-cluster flag.
```

cli/cmd/upgrade.go
```diff
@@ -0,0 +1,193 @@
...
+		if f := flags.Lookup(i.GetName()); f != nil && !f.Changed {
+			f.Value.Set(i.GetValue())
+			f.Changed = true
+		}
+	}
+}
+
+func (options *upgradeOptions) newK8s() (*kubernetes.Clientset, error) {
+	if options.ignoreCluster {
+		return nil, errors.New("--ignore-cluster cannot be used with upgrade")
```
alpeb: We should be able to get rid of the `ignore-cluster` flag for upgrade, which still shows in the help screen. Maybe extracting it from `options.flagSet()` and adding it inside `install.go`'s and `inject.go`'s `newCmdInstall()` ...

alpeb:
This flag appears in both `flagSet()` and `issuerFlagSet()`. Is that intended?
https://github.com/linkerd/linkerd2/blob/e813001c97ef98b13c2100982633a4b6144171c3/cli/cmd/install.go#L284

olix0r:
@alpeb it should not be in the issuer flagset, since this field is reconfigurable but the others are not

olix0r:
@alpeb `issuerFlagSet` is now `installOnlyFlagSet` and includes ignore-cluster.

grampelberg:
Works for me, I tested:

- stable -> this (via `install`)
- this -> this (no options change)
- this -> this + options (proxy + resource requests)

I turned `--proxy-auto-inject` on during the middle of this and can't figure out how to turn it off. Not a blocker, but something to write up an issue and fix later.

olix0r:
@grampelberg can you use `upgrade --proxy-auto-inject=false` ?

grampelberg:
That removes it from the output, I've not tried your `--prune` suggestion yet.

cli/cmd/install.go
```diff
@@ -622,42 +626,48 @@ func (options *installOptions) proxyConfig() *pb.Proxy {
...
-		return false, err
+		fmt.Fprintln(os.Stderr, "Unable to build a Kubernetes client to check for configuration. If this expected, use the --ignore-cluster flag.")
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err)
+		os.Exit(1)
 	}
 
 	k, err := kubernetes.NewForConfig(api.Config)
 	if err != nil {
-		return false, err
+		fmt.Fprintln(os.Stderr, "Unable to build a Kubernetes client to check for configuration. If this expected, use the --ignore-cluster flag.")
```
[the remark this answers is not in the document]
olix0r: Given that this is in install, I think it's important that we have good error messages around these things. i'm not aware of other places, aside from upgrade, where we actually build a kubernetes client in the CLI.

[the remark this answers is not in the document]
olix0r: specifically, i ran into situation where we'd get a message like `Error: EOF` followed by the usage output, which is totally inscrutable.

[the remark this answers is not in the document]
olix0r: ahh i gotcha now. yes, i'll switch up the constructor

alpeb:
> That removes it from the output, I've not tried your --prune suggestion yet.

So the suggestion was
`linkerd upgrade --proxy-auto-inject=false | k apply -f - -l linkerd.io/control-plane-component --prune`

We confirmed this works and removes the `linkerd-proxy-injector` deployment, but the mutating webhook config remains.

Apparently this doesn't affect the creation of new pods. But to be able to clean it up we need to add the `linkerd.io/control-plane-component` (tracked in #2602) and have that config be included in the install yaml (#2176)

cli/cmd/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, i := range install.GetFlags() {
+		if f := flags.Lookup(i.GetName()); f != nil && !f.Changed {
+			f.Value.Set(i.GetValue())
+			f.Changed = true
+		}
+	}
+}
+
+func (options *upgradeOptions) newK8s() (*kubernetes.Clientset, error) {
+	if options.ignoreCluster {
```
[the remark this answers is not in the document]
olix0r: because it's still present on the options structure, i've turned this into a panic (because it's a programmer error for it to be set)

cli/cmd/install.go
```diff
@@ -259,12 +285,17 @@ func (options *installOptions) flagSet(e pflag.ErrorHandling) *pflag.FlagSet {
...
+		"The amount of time to allow for clock skew within a Linkerd cluster",
+	)
 
 	return flags
 }
 
-func (options *installOptions) issuerFlagSet(e pflag.ErrorHandling) *pflag.FlagSet {
-	flags := pflag.NewFlagSet("issuer", e)
+// installOnlyFlagSet includes flags that are only accessible at install-time and not at upgrade-time.
+func (options *installOptions) installOnlyFlagSet(e pflag.ErrorHandling) *pflag.FlagSet {
```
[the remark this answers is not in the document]
olix0r: Actually, I had to go back on this, though I've renamed them a bit further. the install flag set isn't a superset of the base flag set: the "base" (now, "recordable") flag set is the set of flags that we care to serialize in the config, so they can't be combined with the install flag set, for instance.

olix0r:
CI passes but the issue isn't getting notified https://travis-ci.org/linkerd/linkerd2/jobs/514357190

PR #2571, opened 2019-03-27
Cluster DNS domain check provided.

### LINKED ISSUE #1741: Add check for cluster DNS domain
It is possible to change the cluster's TLD and that currently breaks the install (see #1720). A check should be added for whether the cluster is using `cluster.local` or not.

ZzEeKkAa:
Fixes #1741 

Cluster DNS domain checker checks if every node in the cluster has clusterDomain set to 'cluster.local'.

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -79,6 +79,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 		healthcheck.KubernetesAPIChecks,
 		healthcheck.KubernetesVersionChecks,
 		healthcheck.LinkerdVersionChecks,
+		healthcheck.ClusterDNSDomainChecks,
 	}
 
 	if options.preInstallOnly {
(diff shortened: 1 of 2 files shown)


ihcsim:
Still relevant. Once https://github.com/linkerd/linkerd2/pull/2950 (and its subsequent PRs) land, we will need to update this check to make sure `response.KubeletConfig.ClusterDomain` matches the `cluster_domain` property specified in the `linkerd_config` config map

adleong:
@ihcsim now that #2950 has merged, is this ready to be worked on?  @ZzEeKkAa are you still interested in working on this?

ihcsim:
@adleong Yes, this PR should be updated to compare cluster domain in the `KubeletConfig` with that found in the `linkerd-config` config map, which should be accessible via `hc.linkerdConfig.GetGlobal().GetClusterDomain()`.

My only concern is the call to the `/api/v1/nodes/%s/proxy/configz` endpoint in the `linkerd check` command. In a multi-stage installation, the user may not have the proper RBAC for it. Perhaps there is a way to detect the returned `err` and skip the check if it's a RBAC problem?

@ZzEeKkAa If you are still interested in working on this, it might be easier to just cherry-pick your commit to a new branch based of the current `master`. LMK what we can do to help get this merged, as this is super useful.

PR #2572, opened 2019-03-27
install: Replace EnableHA with resource values

olix0r:
This change moves resource-templating logic into a dedicated template,
creates new values types to model kubernetes resource constraints, and
changes the `--ha` flag's behavior to create these resource templates
instead of hardcoding the resource constraints in the various templates.

This is in service of https://github.com/linkerd/linkerd2/pull/2564, where
we want to be able to re-use any existing resource constraints.

### DIFF
--- a/chart/templates/proxy_injector.yaml
+++ b/chart/templates/proxy_injector.yaml
@@ -53,14 +53,8 @@ spec:
             port: 9995
           failureThreshold: 7
         {{ with .ProxyInjectorResources -}}
-        resources:
-          requests:
-            {{- with .CPU.Request}}cpu: {{.}}{{end}}
-            {{- with .Memory.Request}}memory: {{.}}{{end}}
-          limits:
-            {{- with .CPU.Limit}}cpu: {{.}}{{end}}
-            {{- with .Memory.Limit}}memory: {{.}}{{end}}
-        {{- end }}
+        {{- template "resources" . }}
+        {{ end -}}
         securityContext:
           runAsUser: {{.ControllerUID}}
       volumes:
--- a/chart/templates/proxy_injector.yaml
+++ b/chart/templates/proxy_injector.yaml
@@ -52,11 +52,14 @@ spec:
             path: /ready
             port: 9995
           failureThreshold: 7
-        {{- if .EnableHA }}
+        {{ with .ProxyInjectorResources -}}
         resources:
           requests:
-            cpu: 20m
-            memory: 50Mi
+            {{- with .CPU.Request}}cpu: {{.}}{{end}}
+            {{- with .Memory.Request}}memory: {{.}}{{end}}
+          limits:
+            {{- with .CPU.Limit}}cpu: {{.}}{{end}}
+            {{- with .Memory.Limit}}memory: {{.}}{{end}}
         {{- end }}
         securityContext:
           runAsUser: {{.ControllerUID}}
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -62,7 +62,7 @@ type (
 		IdentityResources,
 		PrometheusResources,
 		ProxyInjectorResources,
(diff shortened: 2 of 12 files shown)


web/templates/includes/base.tmpl.html
```diff
@@ -9,10 +9,10 @@
     <meta name="keywords" content="Linkerd">
     <link rel="icon" type="image/png" href="{{.Contents.PathPrefix}}dist/img/favicon.png">
     <link href="https://fonts.googleapis.com/css?family=Lato:300,400,700,900" rel="stylesheet">
-    {{ template "script-tags" . }}
```
olix0r: whoops

cli/cmd/install.go
```diff
@@ -357,7 +399,7 @@ func toPromLogLevel(level string) string {
 	}
 }
 
-func render(values *installValues, w io.Writer, configs *pb.All) error {
```
olix0r: this change isn't strictly necessary, but was helpful in the upgrade branch.

alpeb:
You can apply this to satisfy the linter:

```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 5738564d..97fc0f8c 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -62,7 +62,7 @@ type (
                IdentityResources,
                PrometheusResources,
                ProxyInjectorResources,
-               PublicApiResources,
+               PublicAPIResources,
                TapResources,
                WebResources *resources
 
@@ -372,7 +372,7 @@ func (options *installOptions) validateAndBuild() (*installValues, *pb.All, erro
                values.DestinationResources = defaultConstraints
                values.GrafanaResources = defaultConstraints
                values.ProxyInjectorResources = defaultConstraints
-               values.PublicApiResources = defaultConstraints
+               values.PublicAPIResources = defaultConstraints
                values.TapResources = defaultConstraints
                values.WebResources = defaultConstraints
```

alpeb:
... and this one:
```diff
diff --git a/chart/templates/controller.yaml b/chart/templates/controller.yaml
index f343235e..5fd6ff33 100644
--- a/chart/templates/controller.yaml
+++ b/chart/templates/controller.yaml
@@ -123,7 +123,7 @@ spec:
             path: /ready
             port: 9995
           failureThreshold: 7
-        {{ with .PublicApiResources -}}
+        {{ with .PublicAPIResources -}}
         {{- template "resources" . }}
         {{ end -}}
         securityContext:
```

cli/cmd/install.go
```diff
@@ -345,6 +363,30 @@ func (options *installOptions) validateAndBuild() (*installValues, *pb.All, erro
 		Identity:     identityValues,
 	}
 
+	if options.highAvailability {
```
grampelberg: Looking forward to the future, you *might* want to have a function in the chart that configures this. Definitely not a blocker as that can happen later =)

olix0r: yeah, we should plot out a better factoring for all of this, but i'd rather do that after we have a basic upgrade on master

grampelberg: +1 from me, just wanted folks to be aware.

chart/templates/proxy_injector.yaml
```diff
@@ -52,11 +52,14 @@ spec:
             path: /ready
             port: 9995
           failureThreshold: 7
-        {{- if .EnableHA }}
+        {{ with .ProxyInjectorResources -}}
         resources:
           requests:
-            cpu: 20m
-            memory: 50Mi
+            {{- with .CPU.Request}}cpu: {{.}}{{end}}
```
ihcsim: Curious as to why we aren't using the `resource` template here?

olix0r: oversight, good catch

PR #2573, opened 2019-03-27
Fixes #2568: Remove cni option from inject subcommand

### LINKED ISSUE #2568: Remove --linkerd-cni-enabled inject flag
~~Relates to #2540. It's now possible to override most inject options via `config.linkerd.io` annotations on the pod spec, but there's no support for a `config.linkerd.io/linkerd-cni-enabled` annotation that corresponds to running `linkerd inject --linkerd-cni-enabled`. We should add support for that in `pkg/inject`, and once #2547 merges, we should also start setting that annotation from the CLI when the `--linkerd-cni-enabled` flag is present.~~

As @olix0r points out, linkerd CNI is enabled/disabled at the cluster level (at `install`), and it doesn't make sense to modify it per workland (at `inject`). Now that inject reads the install defaults from the control plane (#2535), we no longer need to support enabling/disabling linkerd CNI on inject. Let's use this issue to track removing the `--linkerd-cni-enabled` inject flag, and make it install only. This means we'll need to move the `noInitContainer` field from the `proxyConfigOptions` struct to the `installOptions` struct.

harsh-98:
Fixes #2568

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -106,6 +106,7 @@ type (
 		highAvailability   bool
 		controllerUID      int64
 		disableH2Upgrade   bool
+		noInitContainer    bool
 		identityOptions    *installIdentityOptions
 		*proxyConfigOptions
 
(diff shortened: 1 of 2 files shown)


harsh-98:
@ihcsim review plz

Issue #2574, opened 2019-03-27
install: Persist install context with configuration


olix0r:
## Problem

When installing Linkerd, a user may override default settings, or may explicitly configure defaults.  Consider install options like `--ha --proxy-cpu-request=20m' -- the `--ha` flag sets a _new_ default value for the proxy-cpu-request, and then we override it.

When we later `upgrade` this cluster, how can we know how to configure the cluster?

We could store EnableHA and ProxyCPURequest configurations in the config, but what if, in a later upgrade, the default value changes? How can we know whether the user specified an override or just used the default?

## Proposal

To solve this, we add an `InstallContext` message onto the Global config. This message includes (at least) the CLI flags used to invoke `install`.

`upgrade` does not specify defaults for install/proxy-options fields and, instead, uses the persisted install flags to populate default values, before applying overrides from the `upgrade` invocation.

ihcsim:
The proposal sgtm. Some questions:

1. I assume the `InstallContext` will only persist the most-recent `install` CLI flags (in the event of multiple calls to `linkerd install --ignore-cluster`, for whatever reasons)
1. If there are new defaults during `upgrade`, the `linkerd-config` config map will get updated, except the `InstallContext`, right?
1. Will we be storing the linkerd version as part of the `InstallContext`? If yes, it should be modified after a successful `upgrade`, right? (I can see the benefits of using the version as a pre-upgrade check, which may be relevant for cases where users aren't allowed to skip certain mandatory intermediate version, when they want to do an one-time multi-version upgrade.)

olix0r:
@ihcsim good questions (that i was just documenting in the config.proto comments).  We'll store the CLI version and args on each install/upgrade. The UUID will only be set on install, as today.

PR #2577, opened 2019-03-27
config: Store install parameters with global config

olix0r:
When installing Linkerd, a user may override default settings, or may
explicitly configure defaults. Consider install options like `--ha
--controller-replicas=4` -- the `--ha` flag sets a new default value for
the controller-replicas, and then we override it.

When we later upgrade this cluster, how can we know how to configure the
cluster?

We could store EnableHA and ControllerReplicas configurations in the
config, but what if, in a later upgrade, the default value changes? How
can we know whether the user specified an override or just used the
default?

To solve this, we add an `InstallContext` message onto the `Global`
config.  This message includes (at least) the CLI flags used to invoke
install.

upgrade does not specify defaults for install/proxy-options fields and,
instead, uses the persisted install flags to populate default values,
before applying overrides from the upgrade invocation.

This change breaks the protobuf compability by altering the
`installation_uuid` field introduced in https://github.com/linkerd/linkerd2/commit/9c442f688575c3ee0261facc7542aa490b89c6cf.
Because this change was not yet released (even in an edge release), we
feel that it is safe to break.

Fixes https://github.com/linkerd/linkerd2/issues/2574


alpeb:
TIOLI: How do you feel about black-listing some of the Global flags from `InstallContext` (like --context, --kubeconfig). Seems like those would be used only for the current `install`/`upgrade`.

Also we still gotta figure the issue you brought in #2564: how does one for disable `--ha`, for example?

proto/config/config.proto
```diff
@@ -71,6 +74,23 @@ message IdentityContext {
   google.protobuf.Duration clock_skew_allowance = 4;
 }
 
+message InstallContext {
+  // The unique ID fr this installation. Does not change on upgrade.
```
alpeb: _for_

proto/config/config.proto
```diff
@@ -19,11 +19,14 @@ message Global {
   // If present, configures identity.
   IdentityContext identity_context = 4;
 
+  // Stores information about the last installation/upgrade.
+  //
+  // Useful for driving upgrades.
+  InstallContext install_context = 5;
```
klingerf: I could go either way on this, but the original idea behind the `linkerd-config` ConfigMap was that it would have multiple keys corresponding to multiple different types of config structs. At the moment, we have keys for "global" and "proxy", corresponding to the `Global` and `Proxy` structs in this file. Since the new config that you're adding is only relevant to install (and upgrade), I'd be inclined to put it in a separate `Install` struct, as follows:

```diff
diff --git a/proto/config/config.proto b/proto/config/config.proto
index 4e365751..a7e92a83 100644
--- a/proto/config/config.proto
+++ b/proto/config/config.proto
@@ -9,6 +9,7 @@ option go_package = "github.com/linkerd/linkerd2/controller/gen/config";
 message All {
   Global global = 1;
   Proxy proxy = 2;
+  Install install = 3;
 }
 
 message Global {
@@ -19,11 +20,6 @@ message Global {
   // If present, configures identity.
   IdentityContext identity_context = 4;
 
-  // Stores information about the last installation/upgrade.
-  //
-  // Useful for driving upgrades.
-  InstallContext install_context = 5;
-
   // If present, indicates that the Mutating Webhook Admission Controller should
   // be configured to automatically inject proxies.
   AutoInjectContext auto_inject_context = 6;
@@ -74,7 +70,7 @@ message IdentityContext {
   google.protobuf.Duration clock_skew_allowance = 4;
 }
 
-message InstallContext {
+message Install {
   // The unique ID fr this installation. Does not change on upgrade.
   string uuid = 1;
 
```

Granted, we don't have a lot of precedent around the organization of this file just yet. But if we're going to treat `Global` as a grab bag of configuration options, I'd be inclined to do away with the multiple config struct setup, and instead move the proxy config options into the global struct as well. But if we're going to try to separate out different types of a configuration, this would be a good opportunity to add a third, top-level `Install` struct.

At the end of the day, I don't have a strong preference, but I want to take whichever approach avoids an unmanageable mess of configuration options :)

olix0r: Yeah, this seems like a good idea. I'll incorporate this once i have the upgrade branch sorted out a bit further. Thanks!

olix0r: I guess the main counterpoint to this is that it will have to get added to `config.All`, so it will still be shipped around everywhere. But conceptually, I think it's probably still worth it...

grampelberg:
I got concerned about serializing the CLI flags themselves this morning. It'll make it difficult for a helm chart to create these settings. Thinking about it some more, helm would never call out to the control plane to read those settings, so it *should* be okay ... with the caveat that once you use a helm chart you can never to back (or we change the serialization strategy in the future).

So, tldr, I can't come up with a major reason not to do it and a whole bunch to do it =)

cli/cmd/install.go
```diff
@@ -206,60 +212,85 @@ func newCmdInstall() *cobra.Command {
...
+func (options *installOptions) recordFlags(flags *pflag.FlagSet) {
+	if flags == nil {
+		return
+	}
+
+	flags.VisitAll(func(f *pflag.Flag) {
+		if f.Changed {
+			switch f.Name {
+			case "ignore-cluster", "linkerd-version":
+				// Thse flags don't make sense to record.
```
alpeb: typo: These

Also how about including here as well `api-addr`, `context`, `kubeconfig` and `verbose`? Most likely these are particular to some user, so shouldn't be stored in the config shared by all.

olix0r: We only have to handle flags set by options.flagSet or their children now! This was the main rational in building FlagSet instead of appending to PersistenFlags.

I've confirmed this with manual testing.

alpeb: And what do you think about `identity-issuer-certificate-file`,  `identity-issuer-key-file` and `identity-trust-anchors-file`? I believe `linkerd upgrade` will ignore them but they don't belong in the shared config IMO.

cli/cmd/install.go
```diff
@@ -511,7 +537,19 @@ func (options *installOptions) globalConfig(identity *pb.IdentityContext) *pb.Gl
 		CniEnabled:        options.noInitContainer,
 		Version:           options.linkerdVersion,
 		IdentityContext:   identity,
-		InstallationUuid:  id,
+	}
+}
+
+func (options *installOptions) installConfig() *pb.Install {
+	installID := uuid.NewV4().String()
```
ihcsim: TIOLI: Depending on how you feel about globals... 
```
// install.go
var generateID = func() string {
  return uuid.NewV4().String()
}

func (options *installOptions) installConfig() *pb.Install {
	installID := generateID()
    ...
}

// install_test.go
generateID = func() string {
  return "deaab91a-f4ab-448a-b7d1-c832a2fa0a60"
}
```
(Personally, I'm ok using it here since it's a short-lived CLI process :stuck_out_tongue:)

cli/cmd/root.go
```diff
@@ -273,39 +275,41 @@ func registryOverride(image, registry string) string {
...
 	// Deprecated flags
-	cmd.PersistentFlags().StringVar(&options.proxyMemoryRequest, "proxy-memory", options.proxyMemoryRequest, "Amount of Memory that the proxy sidecar requests")
-	cmd.PersistentFlags().StringVar(&options.proxyCPURequest, "proxy-cpu", options.proxyCPURequest, "Amount of CPU units that the proxy sidecar requests")
-
-	cmd.PersistentFlags().MarkHidden("proxy-memory")
-	cmd.PersistentFlags().MarkHidden("proxy-cpu")
+	flags.StringVar(&options.proxyMemoryRequest, "proxy-memory", options.proxyMemoryRequest, "Amount of Memory that the proxy sidecar requests")
+	flags.StringVar(&options.proxyCPURequest, "proxy-cpu", options.proxyCPURequest, "Amount of CPU units that the proxy sidecar requests")
+	flags.MarkHidden("proxy-memory")
+	flags.MarkHidden("proxy-cpu")
```
ihcsim: I think calling `MarkHidden()` isn't necessary, as `MarkDeprecated()` will hide the flags too.

cli/cmd/install.go
olix0r: Good call. These should be omitted as well.

pkg/config/config.go
```diff
@@ -43,3 +44,46 @@ func unmarshalConfig(filepath string, msg proto.Message) error {
...
 }
+
+func unmarshal(json string, msg proto.Message) error {
+	u := jsonpb.Unmarshaler{AllowUnknownFields: true}
+	return u.Unmarshal(strings.NewReader(json), msg)
+}
+
+// FromConfigMap builds a configuration by reading a map with the keys "global"
+// and "proxy", each containing JSON values.
+func FromConfigMap(configMap map[string]string) (*pb.All, error) {
```
ihcsim: It doesn't look like this is being used here. Is it part of `upgrade`?

olix0r: yeah, i just merged this whole file from the upgrade branch...

cli/cmd/install.go
olix0r: PTAL. I moved these issuer flags onto a separate flagset so they are just never considered to be recorded (and also means we won't expose them from upgrade).

pkg/config/config.go
olix0r: I'll remove for now and re-add as it's used.

cli/cmd/install.go
olix0r: How do you feel about hanging this off of `installOptions` instead of as a global? The `cmd` package is prettttty big, so i'd rather keep this scoped somehow.

Issue #2558, opened 2019-03-25
The table obtained from linkerd top deploy/linkerd-web is not wrapping in terminal properly


gane5hvarma:
### What is the issue?
The table obtained from linkerd top deploy/linkerd-web is not wrapping in terminal properly. The issue might be not using correct function of printing table on terminal of termbox go package.
### How can it be reproduced?
linkerd -n linkerd top deploy/linkerd-web
### Logs, error output, etc
![linkerdmetrics](https://user-images.githubusercontent.com/23616236/54943982-8050b500-4f58-11e9-9739-f7eb4459bc0c.png)
![linkerd2](https://user-images.githubusercontent.com/23616236/54944049-a8401880-4f58-11e9-8e89-1ba5f351e6ba.png)

(If the output is long, please create a [gist](https://gist.github.com/) and
paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version: 1.11.7
- Cluster Environment: (GKE, AKS, kops, ...) GKE
- Host OS: Ubuntu 16
- Linkerd version: 2.2.1

gane5hvarma:
working on this issue :)

kohsheen1234:
Hey @grampelberg Can I work on this if it's been stale for long??

ihcsim:
@kohsheen1234 Thanks for your interest! Not exactly sure why https://github.com/linkerd/linkerd2/pull/2580 wasn't reviewed and merged (oops!).

@gane5hvarma Are you interested in bringing #2580 cross the finish line? (I can reopen it.) It will need to be rebased with `master`? If not, we can let @kohsheen1234 give it a try. Thanks!

gane5hvarma:
@ihcsim I'm currently occupied with other work. Hence can't spend time in the near future. Let @kohsheen1234 give it a try.

ihcsim:
@gane5hvarma Sounds good. Apologies for overlooking that PR.

@kohsheen1234 It's all yours! Please take a look at @grampelberg's comment at https://github.com/linkerd/linkerd2/pull/2580#issuecomment-477650729. If you have any questions, feel free to join us in the #contributors channel on the Linkerd slack.

kohsheen1234:
Thank you @gane5hvarma, @ihcsim . I will start working on this issue. @ihcsim Can you please assign the issue to me. :)

kohsheen1234:
Hey @ihcsim, 

[...]

PR #2580, opened 2019-03-28
wrapping issue fixed

gane5hvarma:
Signed-off-by: Ganesh Varma R <raghavag772@gmail.com>
Problem: wrapping issue
linkerd2
solution: 
linkerd10
Fixes: #2558

### DIFF
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -377,11 +377,12 @@ func getTrafficByResourceFromAPI(client pb.ApiClient, req *pb.TapByResourceReque
 
 	requestCh := make(chan topRequest, 100)
 	done := make(chan struct{})
-
+	scrollLeft := make(chan int)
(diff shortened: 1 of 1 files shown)


grampelberg:
@gane5hvarma what do you think about making it scrollable instead of wrapping? That output will be difficult if it goes below the bottom of your terminal (ending up in the same state as being too wide). It looks like you can get that to happen by listening to key commands. We'd want a legend along the top to inform users what keys to press.

gane5hvarma:
@grampelberg  sure that's a nice idea. I think scrollable makes easier to look the metrics than wrapping. I will look into termbox pkg documentation for scrollable.

gane5hvarma:
update: Scrolling left and right instead of wrapping
ld1
scroll left -
ld2
scroll right -
ld3

PR #2586, opened 2019-03-28
Adding local and cloud cross-browser testing for dashboard

scottcarol:
This PR creates a process for dashboard integration tests using WebdriverIO and SauceLabs. It includes:
- a WebdriverIO script to run the tests locally via Chromedriver - `wdio.conf.js`
- a WebdriverIO script to run the tests via SauceLabs - `wdio-sauce.conf.js`
- instructions in `TEST.md` on how to set up and run Sauce Connect proxy locally
- addition of `bin/web integration` which can be run from the command line with `cloud` or `local` as arguments 

Notes:
- For web development, the web server launched by `bin/web run` and `bin/web dev` is now 7777, not 8084, because the SauceLabs Sauce Connect proxy can only tunnel to certain ports.
- If you are reviewing this and need a SauceLabs account, ping me and I'll send you an invite ASAP!

### DIFF
--- a/TEST.md
+++ b/TEST.md
@@ -194,6 +194,68 @@ To cleanup the namespaces after the test has finished, run:
 $ bin/test-cleanup
 ```
 
+### Testing the dashboard
+
+We use WebdriverIO to test how the web dashboard looks and operates locally in
+Chrome. For cross-browser testing, we use SauceLabs, which runs simulataneous
+tests on different browsers in the cloud.
+
+Whether you are running the tests locally or in the cloud, you'll first run
+`bin/web dev` which will start up the dashboard at `localhost:7777`.
+
+To run a local WebdriverIO instance that will run the tests on a local instance
+of Chrome, run:
+
+```bash
+$ bin/web integration local
+```
+
+To run cross-browser tests via SauceLabs, you need to do a few things first:
+
+1. Sign up for a (free) SauceLabs sub-account for the account 'buoyant'. If you
+   are not a Buoyant staffer, the best way to get an account invite is to ask in
+   the [Linkerd Slack channel](https://slack.linkerd.io).
+
+2. Once you have your username and key, set them as permanent environment
+   variables. This keeps your credentials private, and means that everyone on
+   the team can run the tests via their unique login without modifying the
+   authentication credentials in the test files. Open your `~/.bash_profile`
+   file and add:
+
+```
+export SAUCE_USERNAME="your Sauce username"
+export SAUCE_ACCESS_KEY="your Sauce access key"
+```
+
+3. Now you'll [download Sauce
+   Connect](https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy), the
+   proxy server that will open a secure tunnel between a SauceLabs VM and the
+   Linkerd dashboard instance you're running on `localhost:7777`. You'll want to
+   save it in a separate directory than the rest of your development files. Once
+   you've downloaded it, navigate to its directory and run:
+
+```bash
+$ bin/sc -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
+```
+
+Wait until you see `Sauce Connect is up, you may start your tests` in your
+terminal. Open a separate terminal window and run:
+
+```bash
+$ bin/web integration cloud
+```
+
+SauceLabs will start running the tests in the cloud. If any tests fail, you'll
+immediately get the URL in your terminal window with a video of the test and
+information about what happened. The test(s) will also appear in [your SauceLabs
+archives](https://app.saucelabs.com/archives) a minute or so after they end.
+(Depending on time of day and server load, it may take longer for the tests to
+appear in the archives.)
+
+4. When you're finished, close the tunnel by pressing `CTRL-C` in the Sauce
+   Connect window. If you forget to do this, it will close on its own after a
+   few minutes.
+
 ## Writing tests
 
 To add a new test, create a new subdirectory inside the `test/` directory.
--- a/TEST.md
+++ b/TEST.md
@@ -218,7 +218,7 @@ bin/web run
 ```
 OR
 ```bash
-# with webpack
+# with webpack-dev-server
 bin/web dev
 ```
 
--- a/TEST.md
+++ b/TEST.md
@@ -196,19 +196,42 @@ $ bin/test-cleanup
 
 ### Testing the dashboard
 
-We use WebdriverIO to test how the web dashboard looks and operates locally in
-Chrome. For cross-browser testing, we use SauceLabs, which runs simulataneous
-tests on different browsers in the cloud.
+We use [WebdriverIO](https://webdriver.io/) to test how the web dashboard looks
+and operates locally in Chrome. For cross-browser testing, we use
+[SauceLabs](https://saucelabs.com/), which runs simulataneous tests on different
+browsers in the cloud.
 
-Whether you are running the tests locally or in the cloud, you'll first run
-`bin/web dev` which will start up the dashboard at `localhost:7777`.
+If you're new to the repo, make sure you've installed web dependencies via
+[Yarn](https://yarnpkg.com):
+
+```bash
+brew install yarn # if you don't already have yarn
+bin/web setup
+```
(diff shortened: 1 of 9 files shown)


TEST.md
```diff
@@ -194,6 +194,68 @@ To cleanup the namespaces after the test has finished, run:
...
+tests on different browsers in the cloud.
+
+Whether you are running the tests locally or in the cloud, you'll first run
+`bin/web dev` which will start up the dashboard at `localhost:7777`.
+
+To run a local WebdriverIO instance that will run the tests on a local instance
+of Chrome, run:
+
+```bash
+$ bin/web integration local
```
siggy: Because I'm not an astute reader, I missed the prior `bin/web dev` step and failed this command the first time. How about calling the previous command out more prominently:

Whether you are running the tests locally or in the cloud, you'll first start up the dashboard at `localhost:7777`. You can do that in one of two ways:

```bash
# standalone
bin/web run
```
OR
```bash
# with webpack
bin/web dev
```

Related, I typically omit the leading `$` if my `bash` block only contains commands and not command output, makes for easier cutting and pasting. I realize we're not consistent with this across our docs, so tioli.

TEST.md
```diff
@@ -194,6 +194,68 @@ To cleanup the namespaces after the test has finished, run:
...
+
+3. Now you'll [download Sauce
+   Connect](https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy), the
+   proxy server that will open a secure tunnel between a SauceLabs VM and the
+   Linkerd dashboard instance you're running on `localhost:7777`. You'll want to
+   save it in a separate directory than the rest of your development files. Once
+   you've downloaded it, navigate to its directory and run:
+
+```bash
+$ bin/sc -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
```
siggy: The first time I saw this command, the leading `bin/` made me assume I was running it from this repo's dir. Consider a little more guidance, for example:

```bash
SC=sc-4.5.3-osx # for example on osx
wget -O - https://saucelabs.com/downloads/$SC.zip | tar xfz - -C ~/
cd ~/$SC
bin/sc -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
```

Related, these instructions are great (and totally work!). Down the road we could consider handling the download and execution of Sauce Connect, similar to how we do things in `bin/lint` or `bin/dep`.

TEST.md
```diff
@@ -194,6 +194,68 @@ To cleanup the namespaces after the test has finished, run:
...
+```
+
+Wait until you see `Sauce Connect is up, you may start your tests` in your
+terminal. Open a separate terminal window and run:
+
+```bash
+$ bin/web integration cloud
```
siggy: I forgot I still needed to be running `bin/web run` or similar when I got to here. Related to the comment earlier, how about calling the prerequisite out as a initial section, and then delineating between Local and Cloud?

```
### Testing the dashboard
We use WebdriverIO...

...
bin/web run
...OR...
bin/web dev
...

#### Local
To run a local WebdriverIO instance that will run the tests on a local instance...

#### Cloud
To run cross-browser tests via SauceLabs...
```

TEST.md
```diff
@@ -194,6 +194,68 @@ To cleanup the namespaces after the test has finished, run:
...
+   are not a Buoyant staffer, the best way to get an account invite is to ask in
+   the [Linkerd Slack channel](https://slack.linkerd.io).
+
+2. Once you have your username and key, set them as permanent environment
+   variables. This keeps your credentials private, and means that everyone on
+   the team can run the tests via their unique login without modifying the
+   authentication credentials in the test files. Open your `~/.bash_profile`
+   file and add:
+
+```
```
siggy: formatting nit: prefix this block with `bash`, and also indent it 4 spaces to align it with the list item. similar comment for item 3 below.

bin/web
```diff
@@ -104,6 +104,16 @@ function test {
   yarn jest "$*"
 }
 
+function integration {
+  if [ $1 = "cloud" ]; then
+    cd $ROOT/web/app && \
+    ./node_modules/wdio/node_modules/.bin/wdio ./integration/wdio-sauce.conf.js
+  else
```
siggy: maybe be a bit stricter here...
```bash
function integration {
  if [[ "$#" -ne 1 || ($1 != "cloud" && $1 != "local") ]]; then
    echo "usage: bin/web integration (cloud|local)" >&2
    exit 1
  fi
  if [ $1 = "cloud" ]; then
    cd $ROOT/web/app && \
    ./node_modules/wdio/node_modules/.bin/wdio ./integration/wdio-sauce.conf.js
  else
    cd $ROOT/web/app && \
    ./node_modules/wdio/node_modules/.bin/wdio ./integration/wdio.conf.js
  fi
}
```

web/app/integration/specs/logo-redirect.js
```diff
@@ -0,0 +1,11 @@
+const assert = require('assert');
+  describe('logo link test', function() {
+    it('should redirect to the home view if logo is clicked', () => {
+    browser.url('http://localhost:7777/tap');
```
siggy: indent?

bin/web
scottcarol: Good suggestion, thanks.

TEST.md
scottcarol: Thanks for testing this out!

TEST.md
scottcarol: Hey @siggy, great feedback, thanks. I really like spelling out the process for the user so it’s as fast as possible. However I worry that by specifying a file location controlled by SauceLabs, they might change it at a later date and create confusion. What do we typically do in this type of situation?

scottcarol: And THANKS for the `geckodriver` catch - that was inadvertently left in there from previous testing. It's definitely not supposed to be in there!

web/app/integration/specs/logo-redirect.js
```diff
@@ -0,0 +1,11 @@
+const assert = require('assert');
+  describe('logo link test', function() {
+      it('should redirect to the home view if logo is clicked', () => {
```
rmars: this indentations seems a little weird. should it be

```js
  describe('logo link test', function() {
    it('should redirect to the home view if logo is clicked', () => {
      browser.url('http://localhost:7777/tap');
      const link = $('.linkerd-word-logo');
      link.click();
      const breadcrumbHeader = $('.breadcrumb-link');
      const pageTitle = breadcrumbHeader.getText();
      assert.equal(pageTitle, 'Overview');
    });
  });
```

scottcarol: weird, thanks for the catch

TEST.md
```diff
@@ -194,6 +194,93 @@ To cleanup the namespaces after the test has finished, run:
...
+# standalone
+bin/web run
+```
+OR
+```bash
+# with webpack
+bin/web dev
+```
+
+### Local
```
siggy: bump this header and the `Cloud` one below out to level 4 ;)
```
#### Local
```

TEST.md
siggy: @scottcarol Good question re: that `sc-4.5.3-osx`. I don't have a silver bullet answer. With both the `lint` and `dep` scripts (and `protoc`!), we also rely on a hard-coded version and URL. Looking at https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy, they do appear to host several older versions, so for now we just keep an eye on it. Once these tests get automated, we should know the moment that URL/version breaks.

In the ideal state, there would exist an API endpoint that would give us the latest version of the tool. `kubectl` does this:
```bash
curl -LO https://storage.googleapis.com/kubernetes-release/release/$(
  curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt
)/bin/darwin/amd64/kubectl
```
(from https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl-binary-using-curl)

Our Linkerd CLI installer is similar, though slightly more magical:
```bash
curl https://run.linkerd.io/install-edge | sh
```

scottcarol: @siggy Cool, thanks for this context.

PR #2589, opened 2019-03-28
Update tap integration test to account for always-on mTLS

### LINKED ISSUE #2519: Identity integration tests
The existing integration tests perform some basic testing with `--tls optional`. Once identity ships as part of the default install, the integration tests will automatically begin testing Identity.

Update the existing integration tests with additional validation for Identity, via metrics, etc.

Part of #2459

alpeb:
Out of all the integration tests (egress, get, stat, tap and
install_test) only in stat and tap do meshed (proxy-to-proxy) connections take
place, which we can test are 100% TLS.

For stat, #2537 already added such check for connections with the
Prometheus pod (connections to other pods are not meshed, apparently).

This PR adds such check for tap.

Fixes #2519

### DIFF
--- a/test/tap/tap_test.go
+++ b/test/tap/tap_test.go
@@ -27,6 +27,7 @@ type tapEvent struct {
 	path       string
 	httpStatus string
 	grpcStatus string
+	tls        string
 	lineCount  int
 }
 
@@ -37,6 +38,7 @@ var (
 		path:       "/buoyantio.bb.TheService/theFunction",
 		httpStatus: "200",
 		grpcStatus: "OK",
+		tls:        "true",
 		lineCount:  3,
 	}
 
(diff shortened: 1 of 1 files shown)


alpeb:
btw, I verified this test failed in Edge-19.3.2 and it passes now :+1:

alpeb:
> Perhaps here, or in a subsequent PR, consider validating `linkerd stat` output shows TLS at 100%.

@siggy [this](https://github.com/linkerd/linkerd2/commit/da0330743ffb98c8d6f6653c4b9665949599314a#diff-49adfddc11fad12047c0909b9a8f7b1d) commit took care of that. Or do you have something extra in mind?

siggy:
@alpeb Nope that's perfect, I forgot we had that test. =)

Issue #2497, opened 2019-03-14
Support Proxy Image Version Override


ihcsim:
Following @klingerf [comment](https://github.com/linkerd/linkerd2/pull/2471#discussion_r265286474), add a new `config.linkerd.io/linkerd-version` annotation to enable version override. The [`taggedProxyImage()`](https://github.com/linkerd/linkerd2/blob/e862e98d1a3c533b5eefafd72afe8925dfb49ea7/pkg/inject/inject.go#L580-L584) and [`taggedProxyInitImage()`](https://github.com/linkerd/linkerd2/blob/e862e98d1a3c533b5eefafd72afe8925dfb49ea7/pkg/inject/inject.go#L586-L590) functions should look up the value of this annotation, before falling back to the default config.

ihcsim:
Fixed by #2593

PR #2593, opened 2019-03-29
Define proxy version override annotation

### LINKED ISSUE #2551: inject: `--linkerd-version` override is not applied
```
linkerd inject strest.yml --linkerd-version=ver-reconnect-backoff  2>/dev/null |grep -c ver-reconn
0
```

ihcsim:
This PR adds a new `config.linkerd.io/linkerd-version` annotation to capture overridden proxy version, using `linkerd inject --linkerd-version`. 

Assumption: The `config.linkerd.io/linkerd-version` and `linkerd.io/proxy-version` use the same overridden value.

```diff
$ bin/go-run cli inject isim-tests/nginx/nginx.yaml --linkerd-version=isim-test  > inject-override.yaml
$ bin/go-run cli inject isim-tests/nginx/nginx.yaml  > inject.yaml
$ diff -u inject.yaml inject-override.yaml
--- inject.yaml 2019-03-29 10:25:43.379796632 -0700
+++ inject-override.yaml        2019-03-29 10:25:33.787871270 -0700
@@ -11,9 +11,10 @@
   template:
     metadata:
       annotations:
+        config.linkerd.io/linkerd-version: isim-test
         linkerd.io/created-by: linkerd/cli dev-45aa10dc-isim
         linkerd.io/identity-mode: default
-        linkerd.io/proxy-version: git-45aa10dc
+        linkerd.io/proxy-version: isim-test
       creationTimestamp: null
       labels:
         app: nginx
@@ -85,7 +86,7 @@
           value: linkerd-identity.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
         - name: LINKERD2_PROXY_DESTINATION_SVC_NAME
           value: linkerd-controller.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
-        image: gcr.io/linkerd-io/proxy:git-45aa10dc
+        image: gcr.io/linkerd-io/proxy:isim-test
         imagePullPolicy: IfNotPresent
         livenessProbe:
           httpGet:
@@ -120,7 +121,7 @@
         - "2102"
         - --inbound-ports-to-ignore
         - 4190,4191
-        image: gcr.io/linkerd-io/proxy-init:git-45aa10dc
+        image: gcr.io/linkerd-io/proxy-init:isim-test
         imagePullPolicy: IfNotPresent
         name: linkerd-init
         resources: {}
```

Fixes #2551 #2497 

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -302,7 +302,6 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
 // storing the corresponding annotations and values.
 func (options *proxyConfigOptions) overrideConfigs(configs *config.All, overrideAnnotations map[string]string) {
 	if options.linkerdVersion != "" {
-		configs.Global.Version = options.linkerdVersion
 		overrideAnnotations[k8s.ProxyVersionOverrideAnnotation] = options.linkerdVersion
 	}
 
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -303,6 +303,7 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
 func (options *proxyConfigOptions) overrideConfigs(configs *config.All, overrideAnnotations map[string]string) {
 	if options.linkerdVersion != "" {
 		configs.Global.Version = options.linkerdVersion
+		overrideAnnotations[k8s.ProxyVersionOverrideAnnotation] = options.linkerdVersion
 	}
 
 	if len(options.ignoreInboundPorts) > 0 {
--- a/cli/cmd/inject_test.go
+++ b/cli/cmd/inject_test.go
@@ -65,6 +65,9 @@ func TestUninjectAndInject(t *testing.T) {
 	defaultConfig := testInstallConfig()
 	defaultConfig.Global.Version = "testinjectversion"
 
+	overrideConfig := testInstallConfig()
+	overrideConfig.Global.Version = "override"
+
 	proxyResourceConfig := testInstallConfig()
 	proxyResourceConfig.Global.Version = defaultConfig.Global.Version
 	proxyResourceConfig.Proxy.Resource = &config.ResourceRequirements{
@@ -179,7 +182,7 @@ func TestUninjectAndInject(t *testing.T) {
 			inputFileName:    "inject_emojivoto_deployment_config_overrides.input.yml",
 			goldenFileName:   "inject_emojivoto_deployment_config_overrides.golden.yml",
 			reportFileName:   "inject_emojivoto_deployment.report",
-			testInjectConfig: defaultConfig,
+			testInjectConfig: overrideConfig,
 		},
 	}
 
--- a/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
+++ b/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
@@ -14,6 +14,7 @@ spec:
     metadata:
       annotations:
         config.linkerd.io/admin-port: "9998"
+        config.linkerd.io/linkerd-version: override
         config.linkerd.io/proxy-cpu-limit: "1"
         config.linkerd.io/proxy-cpu-request: "0.5"
         config.linkerd.io/proxy-memory-limit: 256Mi
@@ -22,7 +23,7 @@ spec:
         config.linkerd.io/skip-outbound-ports: "9999"
         linkerd.io/created-by: linkerd/cli dev-undefined
         linkerd.io/identity-mode: default
-        linkerd.io/proxy-version: testinjectversion
+        linkerd.io/proxy-version: override
       creationTimestamp: null
       labels:
         app: web-svc
@@ -102,7 +103,7 @@ spec:
           value: linkerd-identity.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
         - name: LINKERD2_PROXY_DESTINATION_SVC_NAME
           value: linkerd-controller.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
-        image: gcr.io/linkerd-io/proxy:testinjectversion
+        image: gcr.io/linkerd-io/proxy:override
         imagePullPolicy: IfNotPresent
         livenessProbe:
           httpGet:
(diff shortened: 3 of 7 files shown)


ihcsim:
@siggy I assume in your test, you added the annotation to an existing workload? If I use `linkerd inject --linkerd-version=isim-tests`, both the `linkerd.io/proxy-version` and `config.linkerd.io/linkerd-version` annotations will have the same value:
```
  template:
    metadata:
      annotations:
        config.linkerd.io/linkerd-version: isim-test
        linkerd.io/proxy-version: isim-test
```
I am not 100% sure if that's the desired behaviour. I guess it really boils down to the purpose of these annotations. I am inclined to make `linkerd.io/proxy-version` always picks up the default value, and `config.linkerd.io/linkerd-version` the overridden value. WDYT?

> Does the latter always override the former?

Yes, the function that is used to build the image name uses `config.linkerd.io/linkerd-version`, if it's available. Otherwise, it falls back to `linkerd.io/proxy-version`.

alpeb:
I think it can be a bit confusing for users to have both annotations. Let me know what you think of the following, making room for upgrade policy handling that we've alluded at some point in the past:

Leaving `linkerd.io/proxy-version` to track the version the injected proxy is using like it currently is, and instead of adding `config.linkerd.io/linkerd-version` adding something like `config.linkerd.io/upgrade-policy` whose values could be:

- `config.linkerd.io/upgrade-policy: default` (or annotation missing):
Current behavior of the auto-injector or CLI `inject`pick the latest version according to the global config.
- `config.linkerd.io/upgrade-policy: fixed` and `linkerd.io/proxy-version` set to some value:
The auto-injector will always use the value in `linker.io/proxy-version`. Same for CLI `inject`. If none of the annotations is set but we issue a CLI `inject --linkerd-version=xxx` then the annotations  `config.linkerd.io/upgrade-policy: fixed` and `linkerd.io/proxy-version: xxx` will get set.
- `config.linkerd.io/upgrade-policy: auto` (to-do for later):
When the control panel is upgraded, look for all the workloads with this annotation and upgrade them.

ihcsim:
@alpeb I like the idea of upgrade policy. But I think it might be a topic that merits further design and discussion with others like @olix0r and @grampelberg. 

Fundamentally, for this PR, I think we just wanna decide on when a user override the proxy version using the `--linkerd-version` flag, do we want to persist that info.? Based on #2551 and #2497, I think we do.

ihcsim:
>  Based on #2551 and #2497, I think we do.

To clarify, with or without update policy, if we don't encode the overridden version in the annotation, the next time the proxy injector intercepts the YAML (during an update), it will override the version with the default (if they aren't the same).

siggy:
@ihcsim Thanks for the additional detail. To clarify my set up, I started with:
```yaml
kind: Deployment
spec:
  template:
    metadata:
      annotations:
        config.linkerd.io/linkerd-version: edge-19.3.3
```

... and then after auto-injection I observed:

```yaml
kind: Pod
metadata:
  annotations:
    config.linkerd.io/linkerd-version: edge-19.3.3
    linkerd.io/created-by: linkerd/proxy-injector git-0fe5ea67
    linkerd.io/proxy-version: git-0fe5ea67
```

Is this expected?

ihcsim:
@siggy Yes, that's the expected behaviour for auto-inject; i.e. `config.linkerd.io/linkerd-version` tracks the user-provided value, while `linkerd.io/proxy-version` reads the default from the `linkerd-config` config map. In light of that, I think it makes sense to make both CLI inject and auto-inject consistent. 

In short, the reason why the `linkerd inject` is slightly different is that because it has to account for CLI flag. My next commit to this branch will fix this.

siggy:
@ihcsim I've confirmed with the latest commits that manual and auto-injection both result in the same combination of annotations, specifically:

### Manual

```yaml
$ cat ~/emojivoto-enabled.yml | bin/linkerd inject - --linkerd-version edge-19.3.3

kind: Deployment
spec:
  template:
    metadata:
      annotations:
        config.linkerd.io/linkerd-version: edge-19.3.3
        linkerd.io/created-by: linkerd/cli git-3fd90b90
        linkerd.io/proxy-version: git-3fd90b90
```

### Auto

```yaml
$ kubectl -n emojivoto-enabled get po/emoji-5f85b5d47-zdvkt -oyaml
kind: Pod
metadata:
  annotations:
    config.linkerd.io/linkerd-version: edge-19.3.3
    linkerd.io/created-by: linkerd/proxy-injector git-3fd90b90
    linkerd.io/proxy-version: git-3fd90b90
```

A more meta-question: What's the use case for having both `linkerd.io/proxy-version` and `config.linkerd.io/linkerd-version`? If `config.linkerd.io/linkerd-version` is present, we should use that. If `config.linkerd.io/linkerd-version` is not present, shouldn't we just use whatever is in the `linkerd-config` configMap at `global.version`?

ihcsim:
> What's the use case for having both linkerd.io/proxy-version and config.linkerd.io/linkerd-version?

Having both can feel a bit redundant. Our thought is that one is descriptive, while the other is an implicit override. If we decided that this should be part of the bigger update policy story, I'm fine with holding off.

@olix0r Further thoughts?

>  If config.linkerd.io/linkerd-version is present, we should use that. If config.linkerd.io/linkerd-version is not present, shouldn't we just use whatever is in the linkerd-config configMap at global.version?

That's how it works in this PR, as seen [here](https://github.com/linkerd/linkerd2/blob/3fd90b90d9139b55163aefeafdc0f4814932e7b0/pkg/inject/inject.go#L660-L665).

olix0r:
The `config.linkerd.io/proxy-version` annotation should always be considered an explicit override that is never changed from the default.  `linkerd.io/proxy-version` is always set by the injector with the value used.

This is important when re-injecting a workload: Should we update to a newer proxy version or not? We need these separate namespaces to be semantically different.

olix0r:
>  If config.linkerd.io/linkerd-version is present, we should use that. If config.linkerd.io/linkerd-version is not present, shouldn't we just use whatever is in the linkerd-config configMap at global.version?

Correct.  And the other annotation always indicates the version that was actually used.

ihcsim:
> And the other annotation always indicates the version that was actually used.

Wanna make sure I understand what you meant by _the version that was actually used_. This is referring to whatever version provided during install/upgrade and is saved in the `config.Global.Version` field, right? And this value shouldn't change often.

olix0r:
@ihcsim ah, i see the discrepancy: we have a few possible "linkerd versions". Up until now, (I think) `linkerd.io/linkerd-version` has referred to the proxy's tag. Should we split this into two distinct annotations, i.e. `linkerd.io/proxy-version` & `linkerd.io/inject-version`?

The intent here being that we need to be able to reconstitute how a specific pod was injected..

ihcsim:
Per slack discussion, the version of the injector (CLI, webhook etc.) is currently encoded in the `linkerd.io/created-by` annotation. For clarity, we'll like to separate that into a separate annotation. Also, the CLI `--linkerd-version` flag should be renamed to `--proxy-version`, which configures only the proxy; everything else, including proxy-init, gets the controller's version. We will create a separate issue for these tasks.

PR #2594, opened 2019-03-29
Disable external profiles by default

### LINKED ISSUE #2534: Disable external profiles by default

ihcsim:
Fixes #2534


ihcsim:
Per convo with @olix0r and @alpeb, we've decided to rename the flag to `--enable-external-profiles` with a better UX. Since the previous flag isn't a widely used one, we decided to replace it completely, instead of having to support the different edge scenarios. No changes were made to the corresponding proto file in this PR.

PR #2603, opened 2019-03-31
Web server updated to read the UUID from the linkerd-config ConfigMap.

### LINKED ISSUE #2567: Update web server to read the UUID from the linkerd-config ConfigMap
#2561 added the control plane install's UUID to the global `linkerd-config` ConfigMap. This means that the web server no longer needs to be explicitly passed the UUID via the `-uuid` flag, and it can instead volume mount the ConfigMap and read it from there.

For an example of how other containers volume mount the ConfigMap in the install YAML, see:

https://github.com/linkerd/linkerd2/blob/7efe385feb2afbb5e6d2918ff005d56e17ad00e9/chart/templates/proxy_injector.yaml#L41-L43

And for an example of how values are read from the global config map, see:

https://github.com/linkerd/linkerd2/blob/7efe385feb2afbb5e6d2918ff005d56e17ad00e9/controller/proxy-injector/webhook.go#L91-L94

As part of this change, we also update `web/main.go` to remove the `-uuid` flag.

KatherineMelnyk:
Fixes #2567

### DIFF
--- a/chart/templates/web.yaml
+++ b/chart/templates/web.yaml
@@ -56,12 +56,14 @@ spec:
           containerPort: 8084
         - name: admin-http
           containerPort: 9994
+        volumeMounts:
+        - name: config
+          mountPath: /var/run/linkerd/config
         image: {{.WebImage}}
         imagePullPolicy: {{.ImagePullPolicy}}
         args:
         - "-api-addr=linkerd-controller-api.{{.Namespace}}.svc.cluster.local:8085"
         - "-grafana-addr=linkerd-grafana.{{.Namespace}}.svc.cluster.local:3000"
-        - "-uuid={{.UUID}}"
         - "-controller-namespace={{.Namespace}}"
         - "-log-level={{.ControllerLogLevel}}"
         livenessProbe:
@@ -80,4 +82,8 @@ spec:
         securityContext:
           runAsUser: {{.ControllerUID}}
       serviceAccountName: linkerd-web
+      volumes:
+      - name: config
+        configMap:
+          name: linkerd-config
 {{end -}}
--- a/web/main.go
+++ b/web/main.go
@@ -11,7 +11,9 @@ import (
 
 	"github.com/linkerd/linkerd2/controller/api/public"
 	"github.com/linkerd/linkerd2/pkg/admin"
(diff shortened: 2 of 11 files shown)


chart/templates/web.yaml
```diff
@@ -56,12 +56,14 @@ spec:
           containerPort: 8084
         - name: admin-http
           containerPort: 9994
+        volumeMounts:
+        - name: config
+          mountPath: /var/run/linkerd/config
```
klingerf: Looks like one additional test fixture has been added on master since you submitted this change, and that fixture needs to be updated to pass integration tests. Can you merge the latest master into your branch, and then run:

```
go test ./cli/cmd -update
```

That should result in an update to `cli/cmd/testdata/upgrade_default.golden`.

web/main.go
```diff
@@ -11,8 +11,11 @@ import (
 
 	"github.com/linkerd/linkerd2/controller/api/public"
 	"github.com/linkerd/linkerd2/pkg/admin"
+	"github.com/linkerd/linkerd2/pkg/config"
 	"github.com/linkerd/linkerd2/pkg/flags"
 	"github.com/linkerd/linkerd2/web/srv"
+
```
klingerf: There shouldn't be a space between imports here, and the imports should be sorted alphabetically. This should fix it:

```diff
diff --git a/web/main.go b/web/main.go
index f4338c4d..9b252d3e 100644
--- a/web/main.go
+++ b/web/main.go
@@ -13,9 +13,8 @@ import (
 	"github.com/linkerd/linkerd2/pkg/admin"
 	"github.com/linkerd/linkerd2/pkg/config"
 	"github.com/linkerd/linkerd2/pkg/flags"
-	"github.com/linkerd/linkerd2/web/srv"
-
 	pkgK8s "github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/linkerd/linkerd2/web/srv"
 	log "github.com/sirupsen/logrus"
 )
 
```

KatherineMelnyk:
Sorry, just could not add sign-off in a proper way =(
I've set up gpg key, and commited with `git commit -Sm "..."` but it didn't help. Is there any right way how to add --signoff automatically?

PR #2610, opened 2019-04-01
Add edge-19.4.1 release notes

kleimkuhler:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,17 +1,18 @@
 ## edge-19.4.1
 
 * CLI
-  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured.
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured
   * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
-  * The `inject` command proxy options are now converted into config annotations. The annotations ensure that these configs are persisted in subsequent resource updates
+  * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
   * Always show open TCP connections in the CLI
-  **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command. External profiles are now disabled by default, and can be enabled with the new `--enable-external-profiles` flag.
+  * The `stat` command now always shows the number of open TCP connections
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command; external profiles are now disabled by default and can be enabled with the new `--enable-external-profiles` flag
 * Controller
   * The webhook config now reacts to changes in pods instead of deployments
 * Web UI
   * The Octopus graph now shows TCP stats if no HTTP stats are available
   * The resource detail page no longer shows blank tables if the resource only has TCP traffic
-  * Added validation to the Service Profile form (thanks @liquidslr!)
+  * Added validation to the new service profile form (thanks @liquidslr!)
 
 ## edge-19.3.3
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,17 +2,16 @@
 
 * CLI
   * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured
-  * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
   * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
   * Always show open TCP connections in the CLI
   * The `stat` command now always shows the number of open TCP connections
   * **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command; external profiles are now disabled by default and can be enabled with the new `--enable-external-profiles` flag
 * Controller
-  * The webhook config now reacts to changes in pods instead of deployments
+  * The webhook config now reacts to changes in pods instead of deployments; this means pods parented not only by Depoloyments, but any kind of workload like Daemonsets and Statefulsets can be auto-injected.
 * Web UI
   * The Octopus graph now shows TCP stats if no HTTP stats are available
   * The resource detail page no longer shows blank tables if the resource only has TCP traffic
-  * Added validation to the new service profile form (thanks @liquidslr!)
+  * Added validation to the "new service profile" form (thanks @liquidslr!)
 
 ## edge-19.3.3
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@
   * **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command; external profiles are now disabled by default and can be enabled with the new `--enable-external-profiles` flag
 * Controller
   * The auto-inject admission controller webhook is updated to watch pods creation and update events; with this change, proxy auto-injection now works for all kinds of workloads, including StatefulSets, DaemonSets, Jobs, etc
+* Proxy
+  * Some `l5d-*` informational headers have been temporarily removed from requests and responses because they could leak information to external clients
 * Web UI
   * The topology graph now shows TCP stats if no HTTP stats are available
   * The resource detail page no longer shows blank tables if the resource only has TCP traffic
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,18 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured.
+  * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,18 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured.
```
siggy: most of the previous release notes do not end with a `.`, consider removing for consistency. in the bullets with multiple sentences, maybe just split them with a `;`

kleimkuhler: Ah whoops did not mean to include a `.` there. Thanks!

scottcarol: Everything looks good to me - one styling nit, remove the period from the end of this bullet point

scottcarol: Haha should have reloaded the page, @siggy beat me to it

CHANGES.md
```diff
@@ -1,3 +1,18 @@
...
+  * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
+  * The `inject` command proxy options are now converted into config annotations. The annotations ensure that these configs are persisted in subsequent resource updates
+  * Always show open TCP connections in the CLI
```
rmars: Wording - how about: 
```
The number of open TCP connections is now shown when running `linkerd stat`
```

CHANGES.md
```diff
@@ -1,3 +1,18 @@
...
+  **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command. External profiles are now disabled by default, and can be enabled with the new `--enable-external-profiles` flag.
+* Controller
+  * The webhook config now reacts to changes in pods instead of deployments
+* Web UI
+  * The Octopus graph now shows TCP stats if no HTTP stats are available
+  * The resource detail page no longer shows blank tables if the resource only has TCP traffic
+  * Added validation to the Service Profile form (thanks @liquidslr!)
```
scottcarol: For clarity, I would call this the "New service profile" form

CHANGES.md
kleimkuhler: I agree wording should be changed. In going a little closer to my existing wording, I will probably go with `The `stat` command now always shows the number of open TCP connections`

CHANGES.md
```diff
@@ -1,3 +1,19 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured
+  * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
+  * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
+  * Always show open TCP connections in the CLI
+  * The `stat` command now always shows the number of open TCP connections
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command; external profiles are now disabled by default and can be enabled with the new `--enable-external-profiles` flag
```
siggy: sorry should have called this out in the initial review: recommend bumping this bullet to the beginning for emphasis, similar to the previous releases.

CHANGES.md
```diff
@@ -1,3 +1,18 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured.
+  * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
+  * The `inject` command proxy options are now converted into config annotations. The annotations ensure that these configs are persisted in subsequent resource updates
+  * Always show open TCP connections in the CLI
+  **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command. External profiles are now disabled by default, and can be enabled with the new `--enable-external-profiles` flag.
+* Controller
+  * The webhook config now reacts to changes in pods instead of deployments
```
alpeb: I would add a bit more detail, like "; this means we can now auto-inject pods parented not only to Deployments, but also to any kind of workload like Daemonsets and Statefulsets"

CHANGES.md
kleimkuhler: No problem! @olix0r recommend moving this to the bottom originally, but I can change to whatever we prefer. I can see how we would want to highlight the more significant changes, but I agree that we in the past have began this section highlighting break changes.

CHANGES.md
```diff
@@ -1,3 +1,19 @@
...
+  * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
+  * Always show open TCP connections in the CLI
+  * The `stat` command now always shows the number of open TCP connections
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command; external profiles are now disabled by default and can be enabled with the new `--enable-external-profiles` flag
+* Controller
+  * The webhook config now reacts to changes in pods instead of deployments
+* Web UI
+  * The Octopus graph now shows TCP stats if no HTTP stats are available
+  * The resource detail page no longer shows blank tables if the resource only has TCP traffic
+  * Added validation to the new service profile form (thanks @liquidslr!)
```
scottcarol: I would leave the quotation marks in there for clarity since the form is for creating new service profiles, it's not a new form:

`Added validation to the "new service profile" form`

kleimkuhler: Ah that would explain the capital "N" in your original comment. Thanks!

CHANGES.md
olix0r: Yeah, I don't think this is a pretty unimportant item. Much less impactful than the upgrade change, etc.

CHANGES.md
```diff
@@ -1,3 +1,19 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured
+  * The `--proxy-auto-inject` flag  from the `install` command has been completely revamped
```
alpeb: Is this making reference to #2562? If so, this means the CP remembers that flag, which I believe is implicit in the previous bullet point... If not, could you please elaborate a bit more what this change is about?

ihcsim: Yeah, I was wondering if this bullet point is necessary, as it doesn't change the functionality from a user's POV.

kleimkuhler: @alpeb Yes I was referencing that and the encompassing auto-inject changes. It was significant work, but not user facing so I will remove it.

CHANGES.md
```diff
@@ -1,3 +1,18 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured
+  * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
+  * Always show open TCP connections in the CLI
+  * The `stat` command now always shows the number of open TCP connections
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the `install` command; external profiles are now disabled by default and can be enabled with the new `--enable-external-profiles` flag
+* Controller
+  * The webhook config now reacts to changes in pods instead of deployments; this means pods parented not only by Depoloyments, but any kind of workload like Daemonsets and Statefulsets can be auto-injected.
```
wmorgan: typo: "Depoloyments"

CHANGES.md
```diff
@@ -1,3 +1,18 @@
...
+* Web UI
+  * The Octopus graph now shows TCP stats if no HTTP stats are available
```
wmorgan: Might be better to call this the dependency graph or the topology graph, since I think 🐙 graph is an internal term

CHANGES.md
```diff
@@ -1,3 +1,18 @@
+## edge-19.4.1
+
+* CLI
+  * Introduced an `upgrade` command! This allows an existing Linkerd Control Plane to be reinstalled or reconfigured
```
wmorgan: Is there a pithy summary of why/when you would need this command? Not critical but would be nice.

wmorgan: Also "control plane" is better lower-cased

CHANGES.md
```diff
@@ -1,3 +1,18 @@
...
+  * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
+  * Always show open TCP connections in the CLI
+  * The `stat` command now always shows the number of open TCP connections
```
wmorgan: Is this a duplicate of the previous item?

CHANGES.md
ihcsim: TIOLI: _The auto-inject admission controller webhook is updated to watch pods creation and update events. With this change, proxy auto-injection now works for all kinds of workload including, StatefulSets, DaemonSets, Jobs etc.

CHANGES.md
kleimkuhler: Ah yes good catch -- thanks!

CHANGES.md
kleimkuhler: I will add: "it is particularly useful for automatically reusing flags set in the previous `install` or `upgrade`"

PR #2618, opened 2019-04-02
upgrade: Do not require an 'install' config

olix0r:
Previous control plane versions do not provide an 'install' config, so
this field cannot be required.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -52,18 +52,18 @@ func FromConfigMap(configMap map[string]string) (*pb.All, error) {
 	c := &pb.All{Global: &pb.Global{}, Proxy: &pb.Proxy{}, Install: &pb.Install{}}
 
 	if err := unmarshal(configMap["global"], c.Global); err != nil {
-		return nil, fmt.Errorf("global: %s", err)
+		return nil, fmt.Errorf("invlaid global config: %s", err)
 	}
 
 	if err := unmarshal(configMap["proxy"], c.Proxy); err != nil {
-		return nil, fmt.Errorf("proxy: %s", err)
+		return nil, fmt.Errorf("invalid proxy config: %s", err)
 	}
 
-	// This instal config may not exist when upgrading from previous control plane
+	// This install config may not exist when upgrading from previous control plane
 	// versions.
-	if json := configMap["install"]; json != "" {
+	if json, ok := configMap["install"]; ok {
 		if err := unmarshal(json, c.Install); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -59,8 +59,12 @@ func FromConfigMap(configMap map[string]string) (*pb.All, error) {
 		return nil, fmt.Errorf("proxy: %s", err)
 	}
 
-	if err := unmarshal(configMap["install"], c.Install); err != nil {
-		return nil, fmt.Errorf("install: %s", err)
+	// This instal config may not exist when upgrading from previous control plane
```
kleimkuhler: Spelling of `install` but otherwise looks good

pkg/config/config.go
```diff
@@ -59,8 +59,12 @@ func FromConfigMap(configMap map[string]string) (*pb.All, error) {
...
+	// versions.
+	if json := configMap["install"]; json != "" {
```
klingerf: It seems like we'll encounter this issue again when introducing new configmap keys in the future. My preference would be to fix it more generally in the `unmarshal` func so that it won't bite us again going forward. Something like:

```diff
diff --git a/pkg/config/config.go b/pkg/config/config.go
index 6d56c524..9357bd8b 100644
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -40,6 +40,11 @@ func unmarshalFile(filepath string, msg proto.Message) error {
 }
 
 func unmarshal(json string, msg proto.Message) error {
+	// If we're trying to read a ConfigMap key that does not exist, return nil
+	// rather than returning an unmarshaling error
+	if json == "" {
+		return nil
+	}
 	// If we're using older code to read a newer config, blowing up during decoding
 	// is not helpful. We should detect that through other means.
 	u := jsonpb.Unmarshaler{AllowUnknownFields: true}
@@ -59,12 +64,8 @@ func FromConfigMap(configMap map[string]string) (*pb.All, error) {
 		return nil, fmt.Errorf("proxy: %s", err)
 	}
 
-	// This instal config may not exist when upgrading from previous control plane
-	// versions.
-	if json := configMap["install"]; json != "" {
-		if err := unmarshal(json, c.Install); err != nil {
-			return nil, fmt.Errorf("install: %s", err)
-		}
+	if err := unmarshal(configMap["install"], c.Install); err != nil {
+		return nil, fmt.Errorf("install: %s", err)
 	}
 
 	return c, nil
```

Issue #2559, opened 2019-03-25
Improve k8s access check during initialization


siggy:
## Background
When controller components initialize a k8s client:
https://github.com/linkerd/linkerd2/blob/eafd7482f56e6ba851d1f90f479f4e387b0a0e94/controller/k8s/api.go#L87

... they perform a `SelfSubjectAccessReview` check to verify they have `list` capability on the `pod` resource:
https://github.com/linkerd/linkerd2/blob/eafd7482f56e6ba851d1f90f479f4e387b0a0e94/pkg/k8s/authz.go#L74-L78

## Problem

Checking for "list pods" is at best an approximation of the controller component having the correct access, and at worse is checking for access the component may not need, while not checking for access that the component may in fact need.

## Proposal

Modify the `k8s.InitializeAPI` function to perform access checks based on the resources requested. Because the k8s client uses client-go's informers, we may assume that the component requires list/get/watch access for each of the `APIResource`'s provided to `InitializeAPI`.

Once implemented, we should be able to remove "list pods" access from controller components that do not need it, specifically the proxy injector, which was the original motivation for this ticket: https://github.com/linkerd/linkerd2/pull/2472/files#r268798782

PR #2623, opened 2019-04-02
Add validation webhook for service profiles

### LINKED ISSUE #2075: Improve validation of service profiles
The built-in [validation available to CRDs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation) is very limited.  In particular, it doesn't allow us to validate:

* that there are no unrecognized fields
* the recursive structure of `condition`s

This means that is it a common pitfall to accidentally use an incorrect field name in a condition.  If this happens, routes will not match correctly but it will be unclear why.

We can do more thorough validation by adding a validating admission webhook for service profiles.

alpeb:
Fixes #2075

(This supersedes #2508 because it was easier to redo this in master than
rebasing that PR. It addresses all the feedback received in #2508)

The new webhook endpoint is created in a new container in the
linkerd-controller pod.

The proxy injector webhook was refactored to extract-out the scaffolding
to be reused by the validation webhook (and eventually by any other
admission webhooks we'd like to create in the future). The new common
stuff now reside under controller/webhook.

The validation webhook simply sets up the endpoint server and the
webhook config using pkg/webhook, and calls the old
profiles.Validate to perform the actual validation.

I Also got rid of the unused -no-init-container flag in the proxy injector.

Todo in a follow-up PRs: remove the SP check from the CLI check.


controller/webhook/config.go
```diff
@@ -0,0 +1,81 @@
...
+// a new one
+func (c *Config) Create() (string, error) {
+	exists, err := c.Exists()
+	if err != nil {
+		return "", err
+	}
+
+	if exists {
+		log.Info("deleting existing webhook configuration")
+		if err := c.Ops.Delete(c.api); err != nil {
```
siggy: consider passing in `c.api.Client.AdmissionregistrationV1beta1()` for a more specific interface?

controller/webhook/config.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := t.Execute(buf, spec); err != nil {
+		return "", err
+	}
+
+	return c.Ops.Create(c.api, buf)
+}
+
+// Exists returns true if the webhook already exists
+func (c *Config) Exists() (bool, error) {
+	if err := c.Ops.Get(c.api); err != nil {
```
siggy: is `c.Ops.Get()` only called in the context of this `Exists()` method? if so, what about replacing `Get` with `Exists` on the `ConfigOps` interface?

controller/webhook/launcher.go
```diff
@@ -0,0 +1,64 @@
...
+	addr := flag.String("addr", ":8443", "address to serve on")
+	kubeconfig := flag.String("kubeconfig", "", "path to kubeconfig")
+	controllerNamespace := flag.String("controller-namespace", "linkerd", "namespace in which Linkerd is installed")
+	flags.ConfigureAndParse()
+
+	stop := make(chan os.Signal, 1)
+	defer close(stop)
+	signal.Notify(stop, os.Interrupt, os.Kill)
+
+	k8sAPI, err := k8s.InitializeAPI(*kubeconfig, k8s.NS, k8s.RS)
```
siggy: are `k8s.NS, k8s.RS` needed for all webhooks or just `proxy-injector`? if the latter, consider passing these in as part of the config struct.

controller/webhook/server.go
```diff
@@ -0,0 +1,152 @@
...
+		admissionReview.Response = &admissionv1beta1.AdmissionResponse{
+			UID:     admissionReview.Request.UID,
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: err.Error(),
+			},
+		}
+		return admissionReview
+	}
+	log.Infof("received admission review request %s", admissionReview.Request.UID)
```
siggy: consider logging the namespace/name rather than UID as a more user-friendly message.

curiously when i turned on debug logging, i wasn't seeing the name in the debug output, any idea what's up?
```bash
$ bin/linkerd logs --control-plane-component controller --container sp-validator
...
linkerd linkerd-controller-c689f4fd9-hpqf6 sp-validator time="2019-04-03T01:22:47Z" level=debug msg="admission request: &AdmissionRequest{UID:fd3716f1-55ae-11e9-adbd-025000000001,Kind:linkerd.io/v1alpha1, Kind=ServiceProfile,Resource:linkerd.io/v1alpha1, Resource=serviceprofiles,SubResource:,Name:,Namespace:linkerd,Operation:CREATE,UserInfo:k8s_io_api_authentication_v1.UserInfo{Username:docker-for-desktop,UID:,Groups:[system:masters system:authenticated],Extra:map[string]ExtraValue{},},Object:k8s_io_apimachinery_pkg_runtime.RawExtension{Raw:*[123
```

controller/webhook/launcher.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	k8sAPI.Sync()
+
+	go s.Start()
+	go admin.StartServer(*metricsAddr)
+
+	<-stop
+	log.Info("shutting down webhook server")
+	if err := s.Shutdown(); err != nil {
```
siggy: we're not very consistent about shutdown across our components, but i think we should move towards passing a context with a timeout into shutdown, we have this pattern today in the `web` component. this at least gives a bit more feedback in case an error is returned (and logged)
https://github.com/linkerd/linkerd2/blob/50952c813e6df45daa26bef43aee6e9f40704ec8/web/main.go#L54-L57

controller/webhook/launcher.go
```diff
@@ -0,0 +1,64 @@
...
+	config.controllerNamespace = *controllerNamespace
+	config.rootCA = rootCA
+
+	selfLink, err := config.Create()
+	if err != nil {
+		log.Fatalf("failed to create the webhook configurations resource: %s", err)
+	}
+	log.Infof("created webhook configuration: %s", selfLink)
+
+	s, err := NewServer(k8sAPI, *addr, config.WebhookServiceName, *controllerNamespace, rootCA, config.Handler)
```
siggy: if `config.Handler` only exists to get passed through `Launch()` into `NewServer()`, consider just making it a separate param on `Launch()`. similar question for `MetricsPort` and others.

controller/sp-validator/webhook_ops.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return obj.ObjectMeta.SelfLink, nil
+}
+
+// Get returns an error if the Validating webhook doesn't exist
+func (*Ops) Get(api *k8s.API) error {
+	_, err := api.Client.
+		AdmissionregistrationV1beta1().
+		ValidatingWebhookConfigurations().
+		Get(k8sPkg.SPValidatorWebhookConfigName, metav1.GetOptions{})
```
siggy: since `k8sPkg.SPValidatorWebhookConfigName` was already set in the `Config` struct during `Launch`, any way to just re-use that field? Maybe replace:
```go
type Config struct {
	WebhookConfigName string
}
```
...with:
```go
type ConfigOps interface {
	Name() string
}
```

controller/webhook/server.go
```diff
@@ -0,0 +1,152 @@
...
+
+	s := &Server{server, api, handler, controllerNamespace}
+	s.Handler = http.HandlerFunc(s.serve)
+	return s, nil
+}
+
+// Start starts the https server
+func (s *Server) Start() {
+	log.Infof("listening at %s", s.Server.Addr)
+	if err := s.ListenAndServeTLS("", ""); err != nil {
```
siggy: is `ListenAndServeTLS` eventually going to take params?

olix0r:
What happens when `--ha` is used and multiple controller instances are spawned simultaneously?

olix0r:
I think our options are:

1. Wait on this until https://github.com/linkerd/linkerd2/issues/2628 is done and proceed with this container in the controller pod.
2. Create a new deploy with a new service account with a dedicated serviceaccount that has the ClusterRoleBinding to create service accounts. This deploy should not be scaled with HA.

grampelberg:
> Create a new deploy with a new service account with a dedicated serviceaccount that has the ClusterRoleBinding to create service accounts. This deploy should not be scaled with HA.

I like this idea in general. The controller pod is getting a little large and hard to reason about. Are there any downsides?

alpeb:
@olix0r great catch :+1: 
The second option also sounds better to me. Of course #2628 will still have to be done.

grampelberg:
+1 to #2628 as well =)

controller/sp-validator/tmpl/validating_webhook_configuration.go
```diff
@@ -0,0 +1,22 @@
+package tmpl
+
+// ValidatingWebhookConfigurationSpec provides a template for a
+// ValidatingWebhookConfiguration.
```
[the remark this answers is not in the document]
alpeb: yes we will do that separately alongside with the auto-injector's in #2176. No trivial because it requires the generation of the CA Bundle (or accept it from the user as a CLI option).

controller/webhook/config.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Config contains all the necessary data to build and persist the webhook resource
+type Config struct {
+	MetricsPort         uint32
+	WebhookConfigName   string
+	WebhookServiceName  string
+	TemplateStr         string
+	Ops                 ConfigOps
+	Handler             handlerFunc
```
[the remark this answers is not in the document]
alpeb: `Ops` only deals with the CRUD operations for the webhook config.
`Handler` is focused on the implementation of the hook itself (validating the service profile yaml in this case, injecting the proxy in the proxy-injector case).

controller/webhook/launcher.go
alpeb: Yes, good catch :eye:

controller/webhook/server.go
alpeb: I've found that `Request.Name` is always sent empty by k8s. From their godoc:
```go
// Name is the name of the object as presented in the request.  On a CREATE operation, the client may omit name and
// rely on the server to generate the name.  If that is the case, this method will return the empty string.
// +optional
Name string `json:"name,omitempty" protobuf:"bytes,5,opt,name=name"`
```

siggy: 😮

controller/proxy-injector/webhook_ops.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/linkerd/linkerd2/controller/k8s"
+	k8sPkg "github.com/linkerd/linkerd2/pkg/k8s"
+	log "github.com/sirupsen/logrus"
+	arv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+// Ops satisfies the ConfigOps interface for managing MutatingWebhook configs
+type Ops struct{}
```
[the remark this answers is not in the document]
ihcsim: I think if we move all the code into one package as suggested above, we don't need an `ConfigOps` interface.

ihcsim:
> What happens when --ha is used and multiple controller instances are spawned simultaneously?

If we ever made the proxy injector a required component, we can consider having just one admission controller webhook (and rename the proxy injector to a more generic admission controller or something). The k8s repo has a [similar example](https://github.com/kubernetes/kubernetes/blob/ddf47ac13c1a9483ea035a79cd7c10005ff21a6d/test/images/webhook/main.go#L139-L148) with one webhook, but multiple handlers.

grampelberg:
@ihcsim is there a reason to keep them separate right now? Should we write up an issue to fix that in the 2.4 cycle?

ihcsim:
Right now, the proxy injector webhook is an optional component. AFAIK, this validating webhook is required. I can create an issue for it in 2.4.

grampelberg:
The webhook feels like a nice to have (though with minimal impact), I can go either way.

controller/webhook/launcher.go
alpeb: I kinda did that to avoid having to pass so many arguments, but I think I can reach a compromise in my next push :+1:

controller/sp-validator/webhook_ops.go
alpeb: Yeah, that's better :+1:

controller/webhook/server.go
alpeb: Nope. The certs are added in `s`'s setup above in the `NewServer` function.

alpeb:
@ihcsim thanks for your reviews, I'm still going through them. One comment though:

> how about just put everything into the webhook package and name the different webhook files after their component

If we go down the route of having separate deployments for the proxy-injector and the validator, it would make sense to keep them under different directories under `/controller`, don't you agree?

controller/sp-validator/tmpl/validating_webhook_configuration.go
```diff
@@ -0,0 +1,22 @@
...
+var ValidatingWebhookConfigurationSpec = `
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: {{ .WebhookConfigName }}
+webhooks:
```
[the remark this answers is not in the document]
alpeb: Could you clarify the intention? If I understand correctly this would mean that if there's a timeout or a program error the service profile will pass validation regardless.
If this is set as a parameter in the the webhook config yaml, where would the parameter value come from?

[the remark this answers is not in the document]
ihcsim: `Ignore` is the default. The other supported value is `Fail`. The result of the service profile validation is decided by the `Allow` flag in the admission response. I was thinking the value can come from a new flag added to the server.

Let's just omit this flag for now. We can re-visit this later, as it's relevant to the other webhooks too.

ihcsim:
> If we go down the route of having separate deployments for the proxy-injector and the validator, it would make sense to keep them under different directories under /controller

I don't think the kind of workloads we deploy to k8s should have any inference on the code packaging structure. Per stand-up, let's make this a TIOLI.

chart/templates/controller.yaml
```diff
@@ -187,6 +208,31 @@ spec:
         {{ end -}}
         securityContext:
           runAsUser: {{.ControllerUID}}
+      - name: sp-validator
```
klingerf: Per in-person discussion -- let's move this container to a separate deployment with it's own service account. I suggest creating a new chart file for this, something like `chart/templates/sp-validator.yaml`.

controller/webhook/server.go
```diff
@@ -0,0 +1,152 @@
...
+	if _, err := res.Write(responseJSON); err != nil {
+		http.Error(res, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func (s *Server) processReq(data []byte) *admissionv1beta1.AdmissionReview {
+	admissionReview, err := decode(data)
+	if err != nil {
+		log.Error("failed to decode data. Reason: ", err)
```
siggy: `log.Errorf("failed to decode data. Reason: %s", err)`

controller/webhook/server.go
```diff
@@ -0,0 +1,152 @@
...
+	if req.Body != nil {
+		data, err = ioutil.ReadAll(req.Body)
+		if err != nil {
+			http.Error(res, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}
+
+	if len(data) == 0 {
+		return
```
siggy: is this an error or some condition that should be logged?

chart/templates/proxy_injector.yaml
```diff
@@ -78,6 +78,7 @@ rules:
 - apiGroups: [""]
   resources: ["namespaces"]
   verbs: ["list", "get", "watch"]
+# TODO: this is for creating the webhook config. Can be removed when #2176 lands.
```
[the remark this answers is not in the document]
alpeb: It actually belongs in both places. The pod listing ability is needed more precisely because of #2559.
I tried using Helm comments (`{{- /* comment */ -}}`) but for some reason that didn't fly. I'll remove anyways these comments and I'll make a note instead :wink:

chart/templates/sp_validator.yaml
```diff
@@ -0,0 +1,110 @@
...
+    {{.ControllerComponentLabel}}: sp-validator
+  ports:
+  - name: sp-validator
+    port: 443
+    targetPort: sp-validator
+---
+kind: Deployment
+apiVersion: extensions/v1beta1
+metadata:
+  name: linkerd-sp-validator
```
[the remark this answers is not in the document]
alpeb: Good catches :+1:

controller/webhook/launcher.go
```diff
@@ -0,0 +1,68 @@
...
+	p := strconv.FormatUint(uint64(metricsPort), 10)
+	metricsAddr := flag.String("metrics-addr", ":"+p, "address to serve scrapable metrics on")
+	addr := flag.String("addr", ":8443", "address to serve on")
+	kubeconfig := flag.String("kubeconfig", "", "path to kubeconfig")
+	controllerNamespace := flag.String("controller-namespace", "linkerd", "namespace in which Linkerd is installed")
+	flags.ConfigureAndParse()
+
+	stop := make(chan os.Signal, 1)
+	defer close(stop)
+	signal.Notify(stop, os.Interrupt, os.Kill)
```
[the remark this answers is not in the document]
alpeb: oh ok, noted :+1:

controller/webhook/server.go
alpeb: I believe it's just a sanity check against empty payloads. I'll add some logging.

alpeb:
Alright, thanks for all the feedback :pray: 
The latest push applies all the suggestions, and rebases against master.

PR #2624, opened 2019-04-02
It would be nice to have a hint to next steps after doing a successfu…

christian-posta:
…l check

Not sure this is the best place to add this print statement, happy to take feedback...

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -118,6 +118,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 	}
 
 	fmt.Fprintf(w, "Status check results are %s\n", okStatus)
+	fmt.Fprintf(w, "Install with: linkerd install | kubectl apply -f -\n\n")
 
 	return nil
 }


christian-posta:
Now that I look closer, seems I really would want that print statement for the `--pre` check.. the `check` command while doing install prob wouldn't need this statement

cli/cmd/check.go
```diff
@@ -118,6 +118,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 	}
 
 	fmt.Fprintf(w, "Status check results are %s\n", okStatus)
+	fmt.Fprintf(w, "Install with: linkerd install | kubectl apply -f -\n\n")
```
klingerf: @christian-posta Thanks for submitting this! I agree that this would be a nice improvement. Assuming this message only applies to the pre-install check, I'd prefer for us to only print this message if `options.preInstallOnly` is true. Additionally, if `linkerd check --pre` was run with the `--linkerd-namespace` or `--linkerd-cni-enabled` flags set, we should also print those flags as part of the install command that we print here, since they apply to both check and install. I'm thinking something like:

```
	if options.preInstallOnly {
		installCmd := "linkerd install"
		if controlPlaneNamespace != defaultNamespace {
			installCmd += fmt.Sprintf(" --linkerd-namespace %s", controlPlaneNamespace)
		}
		if options.cniEnabled {
			installCmd += " --linkerd-cni-enabled"
		}
		installCmd += " | kubectl apply -f -"
		fmt.Fprintf(w, "\nInstall the control plane with:\n  %s\n", installCmd)
	}
```

Note that modifying the output of `linkerd check --pre` will also require updating the golden file that we use for validating the pre-check output in our integration tests, here: test/testdata/check.pre.golden.

PR #2627, opened 2019-04-03
Output a URL with instructions after running `linkerd upgrade`

### LINKED ISSUE #2575: Output upgrade instructions

Upgrading between versions is more than just the control plane. #2556 covers providing a command, but it only outputs YAML. There are breaking changes, specific upgrade steps and gotchas when going between versions.

Output a URL that provides the full upgrade instructions between versions as part of the CLI output. This way, users can go read up before then apply to the cluster and follow the steps.

scottcarol:
Fixes #2575. When a user successfully runs `linkerd upgrade` in the CLI, they will see a message after the `yaml` output pointing to a URL with full upgrade instructions.

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -16,8 +16,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-const upgradeURL = "https://linkerd.io/upgrade/"
-const upgradeInstructions = "\nYou're on your way to upgrading Linkerd! Visit this URL for installation instructions: " + upgradeURL + "\n"
+const okMessage = "You're on your way to upgrading Linkerd!\nVisit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps"
+const failMessage = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting"
 
 type (
 	upgradeOptions struct{ *installOptions }
@@ -43,7 +43,7 @@ install command.`,
 			// We need a Kubernetes client to fetch configs and issuer secrets.
 			k, err := options.newK8s()
 			if err != nil {
-				return fmt.Errorf("failed to create a kubernetes client: %s", err)
+				upgradeErrorf("failed to create a kubernetes client: %s", err)
 			}
 
 			// We fetch the configs directly from kubernetes because we need to be able
@@ -52,7 +52,7 @@ install command.`,
 			// control plane.
 			configs, err := fetchConfigs(k)
 			if err != nil {
-				return fmt.Errorf("could not fetch configs from kubernetes: %s", err)
+				upgradeErrorf("could not fetch configs from kubernetes: %s", err)
 			}
 
 			// If the install config needs to be repaired--either because it did not
@@ -80,22 +80,20 @@ install command.`,
 
 			values, err := options.buildValuesWithoutIdentity(configs)
 			if err != nil {
-				return fmt.Errorf("could not build install configuration: %s", err)
+				upgradeErrorf("could not build install configuration: %s", err)
 			}
 
 			identityValues, err := fetchIdentityValues(k, options.controllerReplicas, configs.GetGlobal().GetIdentityContext())
 			if err != nil {
-				fmt.Fprintln(os.Stderr, "Unable to fetch the existing issuer credentials from Kubernetes.")
-				fmt.Fprintf(os.Stderr, "Error: %s\n", err)
-				os.Exit(1)
+				upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s\n", err)
 			}
 			values.Identity = identityValues
 
 			if err = values.render(os.Stdout, configs); err != nil {
-				return fmt.Errorf("could not render install configuration: %s", err)
+				upgradeErrorf("could not render install configuration: %s", err)
 			}
 
-			fmt.Fprintf(os.Stderr, upgradeInstructions)
+			fmt.Fprintf(os.Stderr, "\n%s %s\n", okStatus, okMessage)
 
 			return nil
 		},
@@ -220,4 +218,11 @@ func fetchIssuer(k *kubernetes.Clientset, trustPEM string) (string, string, time
 	}
 
 	return keyPEM, crtPEM, crt.Certificate.NotAfter, nil
+}
+
+//prints the error message and quits the upgrade process
+func upgradeErrorf(format string, a ...interface{}) {
+	template := fmt.Sprintf("%s %s\n%s\n", failStatus, format, failMessage)
+	fmt.Fprintf(os.Stderr, template, a...)
+	os.Exit(1)
 }
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"time"
@@ -16,8 +17,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-const okMessage = "You're on your way to upgrading Linkerd!\nVisit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps"
-const failMessage = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting"
+const (
+	okMessage   = "You're on your way to upgrading Linkerd!\nVisit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps\n"
+	failMessage = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting\n"
+)
 
 type (
 	upgradeOptions struct{ *installOptions }
@@ -85,13 +88,18 @@ install command.`,
 
 			identityValues, err := fetchIdentityValues(k, options.controllerReplicas, configs.GetGlobal().GetIdentityContext())
 			if err != nil {
-				upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s\n", err)
+				upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s", err)
 			}
 			values.Identity = identityValues
 
-			if err = values.render(os.Stdout, configs); err != nil {
+			// rendering to a buffer and printing full contents of buffer after
+			// render is complete, to ensure that okStatus prints separately
+			var buf bytes.Buffer
+			if err = values.render(&buf, configs); err != nil {
 				upgradeErrorf("could not render install configuration: %s", err)
 			}
+
+			buf.WriteTo(os.Stdout)
 
 			fmt.Fprintf(os.Stderr, "\n%s %s\n", okStatus, okMessage)
 
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -46,7 +46,7 @@ install command.`,
 			// We need a Kubernetes client to fetch configs and issuer secrets.
 			k, err := options.newK8s()
 			if err != nil {
-				upgradeErrorf("failed to create a kubernetes client: %s", err)
+				upgradeErrorf("Failed to create a kubernetes client: %s", err)
 			}
 
 			// We fetch the configs directly from kubernetes because we need to be able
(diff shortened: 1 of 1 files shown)


cli/cmd/upgrade.go
```diff
@@ -16,6 +16,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const upgradeURL = "https://linkerd.io/upgrade/"
+const upgradeInstructions = "\nYou're on your way to upgrading Linkerd! Visit this URL for installation instructions: " + upgradeURL + "\n"
```
grampelberg: What do you think about splitting this into two lines and adding some color? As it stands, the output kinda melts together on my screen.



Also, how does this sound - `Visit this URL for further instructions`?

ihcsim: I think we can use just one `const` here. How about:

```diff
-const upgradeURL = "https://linkerd.io/upgrade/"
-const upgradeInstructions = "\nYou're on your way to upgrading Linkerd! Visit this URL for installation instructions: " + upgradeURL + "\n"
+const upgradeInstructions = "You're on your way to upgrading Linkerd! Visit this URL for installation instructions: https://linkerd.io/upgrade/"

 type (
        upgradeOptions struct{ *installOptions }
@@ -95,7 +94,7 @@ install command.`,
                                return fmt.Errorf("could not render install configuration: %s", err)
                        }

-                       fmt.Fprintf(os.Stderr, upgradeInstructions)
+                       fmt.Fprintf(os.Stderr, "\n%s\n", upgradeInstructions)

                        return nil
                },
```

cli/cmd/upgrade.go
```diff
@@ -92,6 +95,8 @@ install command.`,
 				return fmt.Errorf("could not render install configuration: %s", err)
 			}
 
+			fmt.Fprintf(os.Stderr, upgradeInstructions)
```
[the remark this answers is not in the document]
grampelberg: I love this idea, but we should have a separate message and link.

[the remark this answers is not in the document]
ihcsim: :+1: And a different color... like blood red or something.

alpeb:
What do you all think about appending an anchor with the version into that url? We could link directly to the upgrade notice for that version (if any). The downside is that if the user is upgrading from a much older version they'll require to check not only the current upgrade notice but also previous ones.

grampelberg:
@alpeb sounds great to me. I've got an outstanding issue to start versioning the docs, so we could use the anchor to redirect to the correct version.

cli/cmd/upgrade.go
scottcarol: Great idea. I used the existing color characters (√, ×) - let me know what you think.

scottcarol:
@grampelberg @alpeb I like the ⚓️ idea. I separated them out into success and failure, would that work with the doc architecture?  
`https://linkerd.io/upgrade/#nextsteps`
`https://linkerd.io/upgrade/#troubleshooting` 

Re: specific version numbers, when running `linkerd upgrade` can you specify version numbers? Or will you always get the latest upgrade?

cli/cmd/upgrade.go
```diff
@@ -16,6 +16,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const okMessage = "You're on your way to upgrading Linkerd!\nVisit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps"
+const failMessage = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting"
```
ihcsim: I think we can just combine these two lines into a `const` block:
```
const (
  okMessage = "You're on your way to upgrading Linkerd!\nVisit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps"
  failMessage = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting"
)
```

cli/cmd/upgrade.go
```diff
@@ -216,3 +219,10 @@ func fetchIssuer(k *kubernetes.Clientset, trustPEM string) (string, string, time
 
 	return keyPEM, crtPEM, crt.Certificate.NotAfter, nil
 }
+
+//prints the error message and quits the upgrade process
+func upgradeErrorf(format string, a ...interface{}) {
```
ihcsim: This function causes the process to exit on all errors. I think we should preserve the previous way of error handling, where the error is propagated back to the framework.

We should be able to use `defer` to achieve what we want with something like:
```diff
diff --git a/cli/cmd/upgrade.go b/cli/cmd/upgrade.go
index a126ab17..1936fc0a 100644
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -40,10 +40,20 @@ Note that the default flag values for this command come from the Linkerd control
 plane. The default values displayed in the Flags section below only apply to the
 install command.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			upgradeSucceeded := false
+			defer func() {
+				if !upgradeSucceeded {
+					fmt.Fprintf(os.Stderr, "\n%s %s\n", failStatus, failMessage)
+					return
+				}
+
+				fmt.Fprintf(os.Stderr, "\n%s %s\n", okStatus, okMessage)
+			}()
+
 			// We need a Kubernetes client to fetch configs and issuer secrets.
 			k, err := options.newK8s()
 			if err != nil {
-				upgradeErrorf("failed to create a kubernetes client: %s", err)
+				return fmt.Errorf("failed to create a kubernetes client: %s", err)
 			}
 
 			// We fetch the configs directly from kubernetes because we need to be able
@@ -52,7 +62,7 @@ install command.`,
 			// control plane.
 			configs, err := fetchConfigs(k)
 			if err != nil {
-				upgradeErrorf("could not fetch configs from kubernetes: %s", err)
+				return fmt.Errorf("could not fetch configs from kubernetes: %s", err)
 			}
 
 			// If the install config needs to be repaired--either because it did not
@@ -80,21 +90,22 @@ install command.`,
 
 			values, err := options.buildValuesWithoutIdentity(configs)
 			if err != nil {
-				upgradeErrorf("could not build install configuration: %s", err)
+				return fmt.Errorf("could not build install configuration: %s", err)
 			}
 
 			identityValues, err := fetchIdentityValues(k, options.controllerReplicas, configs.GetGlobal().GetIdentityContext())
 			if err != nil {
-				upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s\n", err)
+				fmt.Fprintln(os.Stderr, "Unable to fetch the existing issuer credentials from Kubernetes.")
+				fmt.Fprintf(os.Stderr, "Error: %s\n", err)
+				os.Exit(1)
 			}
 			values.Identity = identityValues
 
 			if err = values.render(os.Stdout, configs); err != nil {
-				upgradeErrorf("could not render install configuration: %s", err)
+				return fmt.Errorf("could not render install configuration: %s", err)
 			}
 
-			fmt.Fprintf(os.Stderr, "\n%s %s\n", okStatus, okMessage)
-
+			upgradeSucceeded = true
 			return nil
 		},
 	}
@@ -219,10 +230,3 @@ func fetchIssuer(k *kubernetes.Clientset, trustPEM string) (string, string, time
 
 	return keyPEM, crtPEM, crt.Certificate.NotAfter, nil
 }
-
-//prints the error message and quits the upgrade process
-func upgradeErrorf(format string, a ...interface{}) {
-	template := fmt.Sprintf("%s %s\n%s\n", failStatus, format, failMessage)
-	fmt.Fprintf(os.Stderr, template, a...)
-	os.Exit(1)
-}

```

klingerf: It was my suggestion to explicitly exit when printing "For troubleshooting help...", rather than returning an error, since in my mind, the link to the URL replaces the need for the standard cobra help output for the command, which prints all of the flags. Printing all of the flags also moves the URL that we print out of the terminal window, so I'd prefer if we did one or the other, but not both.

cli/cmd/upgrade.go
scottcarol: thanks @ihcsim!

cli/cmd/upgrade.go
ihcsim: > since in my mind, the link to the URL replaces the need for the standard cobra help output for the command, which prints all of the flags. Printing all of the flags also moves the URL that we print out of the terminal window

@scottcarol and I discussed about this earlier; definitely not a fan of the block of 'help' text in this case. If the way error is handled here can be different from the rest of the CLI, let's go with this approach of printing the URL, follow by an immediate exit. Since `upgrade` failure is a big deal, the error message and next step (i.e. go to the URL) shouldn't be mingled with the 'help' text.

ihcsim:
>Re: specific version numbers, when running linkerd upgrade can you specify version numbers? Or will you always get the latest upgrade?

It looks like I can decide which version number to upgrade/downgrade to with `linkerd upgrade --linkerd-version=edge-19.3.3`

```
$ linkerd upgrade --linkerd-version edge-19.3.3
...
kind: Deployment
metadata:
  annotations:
    linkerd.io/created-by: linkerd/cli edge-19.4.1 # original
  creationTimestamp: null
  labels:
    linkerd.io/control-plane-component: identity
  name: linkerd-identity
  namespace: linkerd
...
    spec:
      containers:
      - args:
        - identity
        - -log-level=info
        image: gcr.io/linkerd-io/controller:edge-19.3.3 # downgrade version
...
```

@olix0r Does this look right to you?

klingerf:
@grampelberg Interesting find. It also looks like `linkerd upgrade` must be writing the YAML to stdout before writing the success message to stderr, so `kubectl` gets a head start on printing its output before the success message shows up. @scottcarol, maybe we can update this PR to write to stderr first?

scottcarol:
Ok, now we're rendering values to a buffer, which prints when the render is complete. That way the `okStatus` string prints separately. Thanks @klingerf for the help.

(Unfortunately the colors won't print when piping to `kubectl` which is a Unix thing)

Outstanding question @grampelberg @ihcsim are the URLs ok? Right now they are: 
`https://linkerd.io/upgrade/#nextsteps`
`https://linkerd.io/upgrade/#troubleshooting`

grampelberg:
URLs work for me!

klingerf:
Those URLs 404 for me. Should they be https://linkerd.io/2/tasks/upgrade/#upgrade-the-data-plane for success and https://linkerd.io/2/tasks/upgrade/#step-by-step-instructions for failure?

ihcsim:
The URLs LGTM.

@grampelberg Are we setting URL redirects for these URLs?

klingerf:
Re: redirects -- hugo strips anchors when redirecting, so it won't work for https://linkerd.io/upgrade/#nextsteps to redirect to a #nextsteps anchor on a different page. We'd need to link to that page directly for the anchor to work.

By way of example, see how https://linkerd.io/2/ha/#setup redirects to https://linkerd.io/2/features/ha/, which drops the #setup anchor.

siggy:
@klingerf Re: redirects, I had assumed that we could set up `https://linkerd.io/upgrade/#foo` links similar to the checks URLs, for example: https://linkerd.io/checks/#pre-k8s-cluster-k8s

Implemented as static HTML:
https://github.com/linkerd/website/blob/master/linkerd.io/static/checks/index.html

I've found this approach useful for `linkerd check`, as it creates a stable URL that's independent of any changes in hugo or the docs.

klingerf:
@siggy Ah, that's a good approach. I wasn't aware that we were doing non-hugo redirects. Carry on.

grampelberg:
URLs look good to me. @scottcarol do you mind adding those when this PR merges?

scottcarol:
> URLs look good to me. @scottcarol do you mind adding those when this PR merges?

@grampelberg To clarify, you're saying the URLs look good as written:

success: `https://linkerd.io/upgrade/#nextsteps`
failure: `https://linkerd.io/upgrade/#troubleshooting`

and that I should add the URL forwarding to linkerd.io after this PR merges? 

Can do! There's been a lot of different URLs mentioned in this PR so I want to be sure I've got it right.

grampelberg:
@scottcarol yup, and yes please!

cli/cmd/upgrade.go
```diff
@@ -216,3 +227,10 @@ func fetchIssuer(k *kubernetes.Clientset, trustPEM string) (string, string, time
 
 	return keyPEM, crtPEM, crt.Certificate.NotAfter, nil
 }
+
+//prints the error message and quits the upgrade process
```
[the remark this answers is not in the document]
scottcarol: thanks for the tip! 👍

cli/cmd/upgrade.go
```diff
@@ -40,7 +46,7 @@ install command.`,
 			// We need a Kubernetes client to fetch configs and issuer secrets.
 			k, err := options.newK8s()
 			if err != nil {
-				return fmt.Errorf("failed to create a kubernetes client: %s", err)
+				upgradeErrorf("failed to create a kubernetes client: %s", err)
```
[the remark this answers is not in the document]
scottcarol: Good spot, will do.

PR #2635, opened 2019-04-04
Remove proxy injector's -no-init-container flag

klingerf:
In the process of reviewing linkerd/linkerd2#2573, I realized that the proxy injector no longer needs to be explicitly configured via a command line flag when linkerd-cni is enabled, since it now reads that information from the global config map. This PR removes the flag.


klingerf:
@alpeb Ah, sorry about that. I should have looked first. I'm ok closing this PR and deferring the change to your branch. Or I can merge it if you don't think it will be too conflicty. Just let me know what works best for you.

alpeb:
@klingerf no problem, you can merge this one. It shouldn't conflict too much.

Issue #2518, opened 2019-03-18
Introduce ServiceProfile test plan


siggy:
Introduce a test plan to exercise ServiceProfile features, specifically:
- for each of emojivoto and books:
  - generate service profiles
    - from tap
    - from swagger (books only)
    - from proto (emojivoto only)
  - validate retries
  - validate timeouts
  - validate request matches
  - validate metrics
  - #2517 deploy to cloud providers at scale

Part of #2459

PR #2638, opened 2019-04-04
Service profiles integration tests

dadjeibaah:
This PR adds integration tests for service profiles that test profile generation through the tap and Open API flags. This work serves as a baseline for additional work to come listed in issue #2518 . 

Related to #2518


test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,240 @@
...
+		t.Fatalf("kubectl apply command failed:\n%s", err)
+	}
+
+	// check that authors now has more than one route
+	routes, err = getRoutes("deploy/authors", "booksapp", TestHelper)
+	if err != nil {
+		t.Fatalf("routes command failed: %s\n", err)
+	}
+
+	if len(routes) <= 1 {
```
alpeb: I guess you could use here `assertExpectedRoutes` as you do below.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,240 @@
+package serviceprofiles
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/cloudflare/cfssl/log"
```
alpeb: Probably VSCode playing tricks :slightly_smiling_face:

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,240 @@
...
+			injectYAML:  "booksapp.yml",
+			deployments: []string{"webapp", "authors", "books", "traffic"},
+			deployName:  "deploy/books",
+			spName:      "books-svc",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("service profiles from tap:%s", tc.namespace), func(t *testing.T) {
+			cmd := []string{"inject", fmt.Sprintf("testdata/%s", tc.injectYAML)}
```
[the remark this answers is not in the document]
siggy: Confirming that this assumes that the namespaces for these apps have `linkerd.io/inject: enabled`? I could see value in testing both ways, though either approach would be fine for these tests.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,240 @@
...
+				t.Fatalf("kubectl apply command failed:\n%s", out)
+			}
+
+			// run routes: Expected more than default route
+			routes, err = getRoutes(tc.deployName, tc.namespace, TestHelper)
+			if err != nil {
+				t.Fatalf("routes command failed: %s\n", err.Error())
+			}
+			for _, route := range routes {
+				if len(route) <= 1 {
```
siggy: i think `len(route)` is testing the length of each route name? should this instead be...
```go
if len(routes) <= 1 {
	t.Fatalf("Expected routes for service to be greater than or equal to 1 but got %d\n", len(routes))
}
```

dadjeibaah:
@alpeb @siggy thanks for the review. I also was getting concerned about the amount of time the tests were adding to the test suite. However, the original [issue](/linkerd/linkerd2/issues/2518) reference emojivoto and booksapp. I can switch you using either `smoke_test` or `tap_application` instead, which I agree, would be better.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+			spName:         "t1-svc",
+			expectedRoutes: []string{"POST /buoyantio.bb.TheService/theFunction", "[DEFAULT]"},
+		},
+		{
+			sourceName:     "open-api",
+			namespace:      testNamespace,
+			spName:         "t3-svc",
+			deployName:     "deploy/t3",
+			expectedRoutes: []string{"DELETE /testpath", "GET /testpath", "PATCH /testpath", "POST /testpath", "[DEFAULT]"},
+		},
```
ihcsim: I thought there used to be a test case for `proto`.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+				t.Fatalf("Expected routes to have prefixes:\n%s\nbut got:\n%s",
+					strings.Join(tc.expectedRoutes, "\n"),
+					strings.Join(routes, "\n"),
+				)
+			}
+		})
+	}
+}
+
+func assertExpectedRoutes(expected, actual []string) bool {
```
ihcsim: Generally, I expect `assert` functions to either fail or pass, instead of returning boolean value.

Suggestion:
```
func assertExpectedRoutes(expected, actual []string, t *testing.T)  {
	if len(expected) != len(actual) {
		t.Errorf("mismatch routes count. Expected %d, Actual %d", len(expected), len(actual))
	}

	for _, expectedRoute := range expected {
		containsRoute := false
		for _, actualRoute := range actual {
			if strings.HasPrefix(actualRoute, expectedRoute) {
				containsRoute = true
                break
			}
		}
		if !containsRoute {
			t.Errorf("Expected route %s not found in %v", expectedRoute, actual)
		}
	}
}
```
(Doing it this way also allows you to customize the test failure reasons.)

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Run(tc.sourceName, func(t *testing.T) {
+			routes, err := getRoutes(tc.deployName, tc.namespace)
+			if err != nil {
+				t.Fatalf("routes command failed: %s\n", err)
+			}
+
+			initialExpectedRoutes := []string{"[DEFAULT]"}
+
+			if !assertExpectedRoutes(initialExpectedRoutes, routes) {
+				t.Fatalf("Expected routes to have prefixes:\n%s\nbut got:\n%s",
```
ihcsim: This error message doesn't always reflect the result of the two assertions in the `assertExpectedRoutes()` function. See suggestion below on passing `t *testing.T` into `assertExpectedRoutes()`.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+				t.Fatalf("kubectl apply command failed:\n%s", err)
+			}
+
+			routes, err = getRoutes(tc.deployName, tc.namespace)
+			if err != nil {
+				t.Fatalf("routes command failed: %s\n", err)
+			}
+
+			if !assertExpectedRoutes(tc.expectedRoutes, routes) {
+				t.Fatalf("Expected routes to have prefixes:\n%s\nbut got:\n%s",
```
ihcsim: We should make this `t.Errorf()`, so that even if one test case fails, the loop can continue with subsequent test cases.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return nil, err
+	}
+	routes := parseRouteDetails(out)
+	return routes, nil
+}
+
+func parseRouteDetails(cliOutput string) []string {
+	var cliLines []string
+	routesByDeployment := strings.SplitAfter(cliOutput, "\n")
```
ihcsim: I think you can just do `string.Split()`. The following call to `strings.TrimSpace()` is going to strip the line break character anyway.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+		if err := TestHelper.CheckDeployment(testNamespace, deploy, 1); err != nil {
+			t.Error(fmt.Errorf("Error validating deployment [%s]:\n%s", deploy, err))
+		}
+	}
+
+	testCases := []testCase{
+		{
+			sourceName:     "tap",
+			namespace:      testNamespace,
+			deployName:     "deploy/t1",
```
ihcsim: Any reasons why we aren't testing `deploy/t2` and `deploy/t3`?

test/serviceprofiles/serviceprofiles_test.go
dadjeibaah: There was, but as I was going through this I noticed that `TestServiceProfilesDeploy` already exercises this command in `install_test.go`. Do you think it would be better to move that into here?

ihcsim: ah.. :eyes: we can keep the existing one in `install_test.go`. Thanks.

test/serviceprofiles/serviceprofiles_test.go
dadjeibaah: The focus for these tests was for different ways to generate service profiles. So we want to make sure we have one deploy that generates tap data and one deploy that is capable of using a swagger file (deploy/t3, is an HTTP service, so we can add a mock swagger file for it).

Also `deploy/t2` is pretty much similar to `deploy/t1` except it always fails.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+			sourceName:     "open-api",
+			namespace:      testNamespace,
+			spName:         "t3-svc",
+			deployName:     "deploy/t3",
+			expectedRoutes: []string{"DELETE /testpath", "GET /testpath", "PATCH /testpath", "POST /testpath", "[DEFAULT]"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.sourceName, func(t *testing.T) {
```
siggy: to fix ci:
```go
tc := tc // pin
```

test/serviceprofiles/testdata/tap_application.yaml
```diff
@@ -0,0 +1,197 @@
+# slow_cooker --http-> gateway --grpc-> t1
```
siggy: what about reusing `test/tap/testdata/tap_application.yaml`?

test/serviceprofiles/testdata/tap_application.yaml
```diff
@@ -0,0 +1,197 @@
...
+#                              --grpc-> t2 always-error
+#                              --http-> t3
+#
+
+### t1 terminates gRPC requests
+---
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+    name: t1
```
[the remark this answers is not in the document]
dadjeibaah: Whoops, looks like my editor was set to 4 spaces for indents 😮

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,170 @@
...
+			if tc.sourceName == "open-api" {
+				tc.args = []string{
+					"testdata/t3.swagger",
+				}
+			}
+
+			cmd = append(cmd, tc.args...)
+			out, _, err := TestHelper.LinkerdRun(cmd...)
+			if err != nil {
+				t.Fatalf("profile command failed: %s\n", err.Error())
```
siggy: i'm getting a failure on this test (i'll dig into what's happening):
```bash
--- FAIL: TestServiceProfiles (33.18s)
    --- FAIL: TestServiceProfiles/tap (11.57s)
        serviceprofiles_test.go:106: profile command failed: exit status 1
    --- PASS: TestServiceProfiles/open-api (0.93s)
```

...mind adding a bit more output to help troubleshoot?
```go
			if err != nil {
				t.Fatalf("'linkerd %s' command failed with %s: %s\n", cmd, err.Error(), out)
			}
```

test/serviceprofiles/testdata/tap_application.yaml
dadjeibaah: I was thinking about that too, but then I thought it would be better to have a `tap_application.yaml` specifically for `serviceprofiles`. If we want to modify t* services for specific test scenarios. e.g modifying t3 to fail 50% of the time to test out and validate retries.

siggy: Makes sense, and I see you just added `--percent-failure`... mind bumping the `bb` images to `v0.0.5` to match the original version?

siggy:
fwiw i'm getting the same failure locally: https://gist.github.com/l5d-bot/1e79ad2cdbd250b787962861273485c3

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,161 @@
...
+			if tc.sourceName == "open-api" {
+				tc.args = []string{
+					"testdata/t3.swagger",
+				}
+			}
+
+			cmd = append(cmd, tc.args...)
+			out, stderr, err := TestHelper.LinkerdRun(cmd...)
+			if err != nil {
+				t.Fatalf("'linkerd %s' command failed with %s: %s\n", cmd, err.Error(), stderr)
```
siggy: this failed on my local system with:
```bash
$ go test -v ./test/serviceprofiles -integration-tests -linkerd `pwd`/bin/linkerd
=== RUN   TestServiceProfiles
=== RUN   TestServiceProfiles/tap
=== RUN   TestServiceProfiles/open-api
--- FAIL: TestServiceProfiles (35.05s)
    --- FAIL: TestServiceProfiles/tap (10.43s)
        serviceprofiles_test.go:102: 'linkerd [profile --namespace l5d-integration-serviceprofile-test t1-svc --tap deploy/t1 --tap-route-limit 5 --tap-duration 10s]' command failed with exit status 1: Error: Tap duration exceeded, try increasing --tap-duration
            Usage:
```

i'm guessing it's a timing issue between slow-cooker's `sleep 15` at startup, and the `profile --tap --tap-duration 10s` not seeing any requests.

i think you can mitigate this by increasing `--tap-duration` to `25s`, but also decrease `--tap-route-limit` to `1`, so the command should complete as soon as it sees the first request.

```diff
diff --git a/test/serviceprofiles/serviceprofiles_test.go b/test/serviceprofiles/serviceprofiles_test.go
index 34803828..c3fadc26 100644
--- a/test/serviceprofiles/serviceprofiles_test.go
+++ b/test/serviceprofiles/serviceprofiles_test.go
@@ -84,9 +84,9 @@ func TestServiceProfiles(t *testing.T) {
                                tc.args = []string{
                                        tc.deployName,
                                        "--tap-route-limit",
-                                       "5",
+                                       "1",
                                        "--tap-duration",
-                                       "10s",
+                                       "25s",
                                }
                        }
```

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+		if !containsRoute {
+			t.Errorf("Expected route %s not found in %v", expectedRoute, actual)
+		}
+	}
+}
+
+func getRoutes(deployName, namespace string) ([]string, error) {
+	cmd := []string{"routes", "--namespace", namespace, deployName}
```
siggy: not necessarily for this PR, but it may make parsing easier if you add `--output json`

PR #2644, opened 2019-04-05
install: Change --ha to set a 100m CPU request

### LINKED ISSUE #2643: Change --ha to set a 100m CPU request
When the `--ha` flag is set, we currently set a `10m` CPU request, which correspnds to 1% of a core, which isn't actually enough to keep the proxy responding to health checks if you have 100 processes on the box. Let's give ourselves a little more breathing room.

olix0r:
When the --ha flag is set, we currently set a 10m CPU request, which
correspnds to 1% of a core, which isn't actually enough to keep the
proxy responding to health checks if you have 100 processes on the box.
Let's give ourselves a little more breathing room.

Fixes #2643

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -430,7 +430,7 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
 
 	if options.highAvailability {
 		defaultConstraints := &resources{
-			CPU:    constraints{Request: "20m"},
+			CPU:    constraints{Request: "100m"},
 			Memory: constraints{Request: "50Mi"},
 		}
 		// Copy constraints to each so that further modification isn't global.
@@ -441,10 +441,10 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
 		values.TapResources = &*defaultConstraints
 		values.WebResources = &*defaultConstraints
 
-		values.IdentityResources = &resources{
-			CPU:    constraints{Request: "100m"},
-			Memory: constraints{Request: "10Mi"},
-		}
+		// The identity controller maintains no internal state, so it need not request
+		// 50Mi.
+		values.IdentityResources = &*defaultConstraints
+		values.IdentityResources.Memory = "10Mi"
 
 		values.PrometheusResources = &resources{
 			CPU:    constraints{Request: "300m"},
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -426,6 +426,15 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
 			Proxy:   proxyJSON,
 			Install: installJSON,
 		},
+
+		DestinationResources:   &resources{},
+		GrafanaResources:       &resources{},
+		IdentityResources:      &resources{},
+		PrometheusResources:    &resources{},
+		ProxyInjectorResources: &resources{},
+		PublicAPIResources:     &resources{},
+		TapResources:           &resources{},
+		WebResources:           &resources{},
 	}
 
 	if options.highAvailability {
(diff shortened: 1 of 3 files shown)


cli/cmd/install.go
```diff
@@ -442,7 +442,7 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
 		values.WebResources = &*defaultConstraints
 
 		values.IdentityResources = &resources{
-			CPU:    constraints{Request: "10m"},
+			CPU:    constraints{Request: "100m"},
 			Memory: constraints{Request: "10Mi"},
 		}
```
klingerf: Just curious, why isn't identity using `defaultConstraints`, like the rest of our go control plane components? Or put another way, should we also modify the default constraints CPU request to be 100m instead of 20m, and start using that here instead?

olix0r: yeah, we should probably update the defaults, but the identity controller maintains _no_ state, so it really doesn't need much in the way of memory reservations.

cli/cmd/install.go
```diff
@@ -441,10 +441,10 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
...
 		values.WebResources = &*defaultConstraints
 
-		values.IdentityResources = &resources{
-			CPU:    constraints{Request: "10m"},
-			Memory: constraints{Request: "10Mi"},
-		}
+		// The identity controller maintains no internal state, so it need not request
+		// 50Mi.
+		values.IdentityResources = &*defaultConstraints
+		values.IdentityResources.Memory = constraints{Request: "10Mi"}
```
klingerf: Ugh, unfortunately, this modifies the memory request for all components using `defaultConstraints`. The issue is that `&*` isn't actually creating a new pointer -- it's copying the same pointer to all of the resource fields.

This fixes it, but it sure us ugly:

```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 384a4203..cb78b0e5 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -426,6 +426,15 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
 			Proxy:   proxyJSON,
 			Install: installJSON,
 		},
+
+		DestinationResources:   &resources{},
+		GrafanaResources:       &resources{},
+		IdentityResources:      &resources{},
+		PrometheusResources:    &resources{},
+		ProxyInjectorResources: &resources{},
+		PublicAPIResources:     &resources{},
+		TapResources:           &resources{},
+		WebResources:           &resources{},
 	}
 
 	if options.highAvailability {
@@ -434,16 +443,16 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*ins
 			Memory: constraints{Request: "50Mi"},
 		}
 		// Copy constraints to each so that further modification isn't global.
-		values.DestinationResources = &*defaultConstraints
-		values.GrafanaResources = &*defaultConstraints
-		values.ProxyInjectorResources = &*defaultConstraints
-		values.PublicAPIResources = &*defaultConstraints
-		values.TapResources = &*defaultConstraints
-		values.WebResources = &*defaultConstraints
+		*values.DestinationResources = *defaultConstraints
+		*values.GrafanaResources = *defaultConstraints
+		*values.ProxyInjectorResources = *defaultConstraints
+		*values.PublicAPIResources = *defaultConstraints
+		*values.TapResources = *defaultConstraints
+		*values.WebResources = *defaultConstraints
 
 		// The identity controller maintains no internal state, so it need not request
 		// 50Mi.
-		values.IdentityResources = &*defaultConstraints
+		*values.IdentityResources = *defaultConstraints
 		values.IdentityResources.Memory = constraints{Request: "10Mi"}
 
 		values.PrometheusResources = &resources{
```

There's prob a better way to go about it.

olix0r:
@ihcsim afaiu it would be pretty difficult to force the identity pod to use that much memory, as it maintains almost no internal state (unlike most of our other controllers).

Issue #2453, opened 2019-03-06
Fix known logging errors and warnings


siggy:
#2414 introduced an integration test to ensure proxy and control-plane logs do not contain unexpected errors. This test required inclusion of a list of known errors and warnings which commonly occur:
https://github.com/linkerd/linkerd2/blob/d850b02b667dfdbbf08f4dcccf2046c5165d8209/test/install_test.go#L51-L68

This ticket is intended to track tasks that eliminate these known errors.

olix0r:
Integration tests have since changed.

PR #2647, opened 2019-04-05
Skip proxy logging errors in integration tests

siggy:
The list of known proxy log errors has been growing, and causing regular
ci failures.

Skip proxy logging errors. The tests will continue to run and report
unexpected errors, but this will not fail the tests (and ci). Also break
out the controller log errors separately, and continue to fail on those.

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -49,12 +49,15 @@ var (
 	}
 
 	// Linkerd commonly logs these errors during testing, remove these once
-	// they're addressed.
-	// TODO: eliminate these errors: https://github.com/linkerd/linkerd2/issues/2453
-	knownErrorsRegex = regexp.MustCompile(strings.Join([]string{
+	// they're addressed: https://github.com/linkerd/linkerd2/issues/2453
+	knownControllerErrorsRegex = regexp.MustCompile(strings.Join([]string{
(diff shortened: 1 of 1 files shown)


test/install_test.go
```diff
@@ -335,32 +335,45 @@ func TestLogs(t *testing.T) {
...
+					t.Errorf("No logs found for %s", name)
+				}
 
-			for _, line := range outputLines {
-				if errRegex.MatchString(line) && !knownErrorsRegex.MatchString(line) {
-					t.Errorf("Found error in %s/%s log: %s", deploy, container, line)
+				for _, line := range outputLines {
+					if errRegex.MatchString(line) && !knownErrorsRegex.MatchString(line) {
+						if proxy {
+							t.Skipf("Found proxy error in %s log: %s", name, line)
```
ihcsim: Just wanna confirm the intention here, this is going to skip all proxy errors that don't match the known regex, right?

siggy: Yeah, the intent here is to allow our integration tests to pass in the face of unexpected errors from the proxy. Once #2453 is resolved, we can re-enable this (and hopefully remove all known logging errors from this test).

This particular change enables reporting of unexpected proxy errors in the output via `t.Skipf`, but still allows the integration tests to pass. Expected logging errors do not get reported at all.

PR #2656, opened 2019-04-06
upgrade: Generate an Identity config if missing

### LINKED ISSUE #2650: linkerd-identity pod is not created on upgrade
After upgrading from Linkerd `edge-19.3.2` to `edge-19.4.1`, no `linkerd-identity` deployment is created.  This causes `linkerd check` to hang indefinitely while waiting for the `linkerd-identity` pod to become ready.

## Steps to reproduce
Using Linkerd `edge-19.3.2` CLI:

```
$ linkerd install | kubectl apply -f -
```

Using Linkerd `edge-19.4.1` CLI:

```
$ linkerd upgrade | kubectl apply -f -
$ linkerd check
```

olix0r:
When upgrading from an older cluster that has a Linkerd config but no
identity, we need to generate an identity context so that the cluster is
configured properly.

Fixes #2650

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -143,7 +143,7 @@ func setOptionsFromInstall(flags *pflag.FlagSet, install *pb.Install) {
 	}
 }
 func (options *upgradeOptions) overrideConfigs(configs *pb.All, overrideAnnotations map[string]string) {
-	options.proxyConfigOptions.overrideConfigs(configs, overrideAnnotations)
+	options.installOptions.overrideConfigs(configs, overrideAnnotations)
 
 	if options.proxyAutoInject {
 		configs.GetGlobal().AutoInjectContext = &pb.AutoInjectContext{}
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -111,11 +111,23 @@ func (options *upgradeOptions) validateAndBuild(k kubernetes.Interface, flags *p
 		upgradeErrorf("Could not build install configuration: %s", err)
 	}
 
-	identityValues, err := fetchIdentityValues(k, options.controllerReplicas, configs.GetGlobal().GetIdentityContext())
-	if err != nil {
-		upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s", err)
+	idctx := configs.GetGlobal().GetIdentityContext()
+	if idctx.GetTrustDomain() == "" || idctx.GetTrustAnchorsPem() == "" {
+		// If there wasn't an idctx, or if it doesn't specify the required fields, we
+		// must be upgrading from a version that didn't support identity, so generate it anew...
+		v, err := options.installOptions.identityOptions.genValues()
+		if err != nil {
+			upgradeErrorf("Unable to generate issuer credentials.\nError: %s", err)
+		}
+		values.Identity = v
+		configs.GetGlobal().IdentityContext = v.toIdentityContext()
+	} else {
+		v, err := fetchIdentityValues(k, options.controllerReplicas, idctx)
+		if err != nil {
+			upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s", err)
+		}
+		values.Identity = v
 	}
-	values.Identity = identityValues
 
 	return values, configs, nil
 }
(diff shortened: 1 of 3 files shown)


cli/cmd/upgrade.go
```diff
@@ -111,11 +111,23 @@ func (options *upgradeOptions) validateAndBuild(k kubernetes.Interface, flags *p
...
+	idctx := configs.GetGlobal().GetIdentityContext()
+	if idctx.GetTrustDomain() == "" || idctx.GetTrustAnchorsPem() == "" {
+		// If there wasn't an idctx, or if it doesn't specify the required fields, we
+		// must be upgrading from a version that didn't support identity, so generate it anew...
+		v, err := options.installOptions.identityOptions.genValues()
+		if err != nil {
+			upgradeErrorf("Unable to generate issuer credentials.\nError: %s", err)
+		}
+		values.Identity = v
+		configs.GetGlobal().IdentityContext = v.toIdentityContext()
```
siggy: tioli, factor out `values.Identity = v` to below the if/else

olix0r: i don't think this can actually be simplified as it is without introducing extra `var` definitions

olix0r:
@siggy that looks like an older auto-injector is reading newer configuration. are you sure that's related to this branch?

olix0r:
Specifically, what I assume is happening here is that the linkerd-config is updated before the proxy-injector is updated, and it's reading the newer config and finding new, unexpected fields.

But it would be helpful to compare the proxy-injector pod metadata and the config structure.

Also worth calling out that, in newer versions, decoding newer configs doesn't cause an error (we now ignore unknown fields)

siggy:
@olix0r I tried the same process again with this branch, and with `edge-19.4.2`, both upgrading from `edge-19.3.2`. I was unable to reproduce both times. Either I missed something in my upgrade steps, or there's a timing problem around the `proxy-injector` coming up. here's the full output from the failed upgrade:
https://gist.github.com/siggy/39b2a5871aea0e2388dfdf7c7be556c4

olix0r:
@siggy it seems pretty likely that there's a race. Perhaps something we need to understand as we revisit the install split...

cli/cmd/upgrade.go
```diff
@@ -106,16 +106,30 @@ func (options *upgradeOptions) validateAndBuild(k kubernetes.Interface, flags *p
...
+		identity, err = fetchIdentityValues(k, options.controllerReplicas, idctx)
+		if err != nil {
+			upgradeErrorf("Unable to fetch the existing issuer credentials from Kubernetes.\nError: %s", err)
+		}
 	}
 
-	identityValues, err := fetchIdentityValues(k, options.controllerReplicas, configs.GetGlobal().GetIdentityContext())
+	// Values have to be generated after any missing identity is generated,
+	// otherwise it will be missing from the generated configmap.
+	values, err := options.buildValuesWithoutIdentity(configs)
```
siggy: Confirmed successful upgrade from `edge-19.3.2` to this branch.

Related, the `autoInjectContext` is populated as `{}` via `install` but `null` via `upgrade`:
```bash
$ bin/linkerd install --ignore-cluster --proxy-auto-inject | grep autoInjectContext
..."autoInjectContext":{}}

$ bin/linkerd upgrade --proxy-auto-inject | grep autoInjectContext
..."autoInjectContext":null}
```

I think if we did something like `configs := options.configs(identityValues.toIdentityContext())`, we'd get a correctly populated `autoInjectContext` field.

olix0r: i don't want to call `proxyInjectOptions.configs` here, since i want to use the "original" config object and just update it from options (rather than generate a new config object entirely).

https://github.com/linkerd/linkerd2/commit/ccbf846e01ae5dda33c46e14b34b9dae1fd599b9 overrides the `overrideConfigs` receiver for `upgradeOptions` so that it can override the config with the autoinject settings.

This used to work but regressed in https://github.com/linkerd/linkerd2/commit/31706e541758d754ab873d6f854a0b3028713346.

cli/cmd/upgrade.go
```diff
@@ -128,6 +142,13 @@ func setOptionsFromInstall(flags *pflag.FlagSet, install *pb.Install) {
 		}
 	}
 }
+func (options *upgradeOptions) overrideConfigs(configs *pb.All, overrideAnnotations map[string]string) {
+	options.proxyConfigOptions.overrideConfigs(configs, overrideAnnotations)
```
siggy: tioli, to future-proof against changes in the install codepath:
```go
options.installOptions.overrideConfigs(configs, overrideAnnotations)
```

PR #2665, opened 2019-04-07
Updating WebdriverIO libraries for front-end integration tests

scottcarol:
This PR updates the WebdriverIO libraries used in the front-end integration tests as well as the `integration` function in `bin/web`. There should be no functionality change if you are running the tests.

Context: I've learned that some of the WebdriverIO libraries I used are from the community around `webdriverio` and have offiically-supported counterparts, prefixed with `@wdio`. I decided to switch to the "official" libraries, except in the case of `wdio-chromedriver-service` where a comparable "official" library does not exist.

### DIFF
--- a/bin/web
+++ b/bin/web
@@ -111,10 +111,10 @@ function integration {
   fi
   if [ $1 = "cloud" ]; then
     cd $ROOT/web/app && \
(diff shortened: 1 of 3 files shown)


scottcarol:
@dadjeibaah Really good point, I'll make that change in a separate PR!

PR #2666, opened 2019-04-08
Output check result as json

### LINKED ISSUE #2503: Scripting friendly output for "linkerd check"
## Feature Request

I would like to use "linkerd check --pre" and "linkerd check" in a deployment script. It would be great to have the possibility that would enable the output to be in a scripting friendly, easy to parse format (e.g.: JSON)

I would like to inform the users of the deployment script without "dumping" the whole "linkerd check" output into an error message (if and when "linkerd check" fail).

A flag could be added to the tool, for example: "-o json".

---

Users could specify the flag when using the CLI command.

alenkacz:
Fixes #2503


cli/cmd/check.go
```diff
@@ -66,6 +69,7 @@ non-zero exit code.`,
 	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for all tests to pass")
 	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	cmd.PersistentFlags().BoolVar(&options.cniEnabled, "linkerd-cni-enabled", options.cniEnabled, "When running pre-installation checks (--pre), assume the linkerd-cni plugin is already installed, and a NET_ADMIN check is not needed")
+	cmd.PersistentFlags().StringVarP(&options.namespace, "output", "o", options.output, "Output format. One of: basic, json  (default: basic)")
```
siggy: `cmd.PersistentFlags().StringVarP(&options.output`

also you can omit `(default: basic)`, as it's already printed as a default via `newCheckOptions` and cobra.

cli/cmd/check.go
```diff
@@ -126,10 +130,20 @@ func (o *checkOptions) validate() error {
 	if o.preInstallOnly && o.dataPlaneOnly {
 		return errors.New("--pre and --proxy flags are mutually exclusive")
 	}
+	if o.output != "basic" && o.output != "json" {
```
siggy: mind using the `jsonOutput` constant defined in `root.go`? also instead of "basic", i think it's ok to just use `tableOutput`, to match the other commands.

cli/cmd/check.go
```diff
@@ -172,3 +186,72 @@ func runChecks(w io.Writer, hc *healthcheck.HealthChecker) bool {
...
+
+type checkCategory struct {
+	Name   string   `json:"categoryName"`
+	Checks []*check `json:"checks"`
+}
+
+type check struct {
+	Description string      `json:"description"`
+	Hint        string      `json:"hint"`
+	Result      checkResult `json:"result"`
```
siggy: what do you think about adding json decorators to the `healthcheck.CheckResult` struct, rather than creating this similar `check` struct? that should hopefully simplify `collectJSONOutput` a bit?

added bonus, it will pick up `result.Err` for free, as i think we should include that in the json output...

```
linkerd-version
---------------
‼ cli is up-to-date
    unsupported version channel: dev-abcb8bb2-siggy
```

```json
  {
    "categoryName": "linkerd-version",
    "checks": [
      {
        "description": "can determine the latest version",
        "hint": "",
        "result": "success"
      },
      {
        "description": "cli is up-to-date",
        "hint": "https://linkerd.io/checks/#l5d-version-cli",
        "result": "warning"
      }
    ]
  },
```

cli/cmd/testdata/check_output_json.golden
```diff
@@ -0,0 +1,17 @@
+[
```
siggy: 🙏

cli/cmd/check.go
```diff
@@ -107,7 +111,7 @@ func configureAndRunChecks(w io.Writer, options *checkOptions) error {
 		RetryDeadline:         time.Now().Add(options.wait),
 	})
 
-	success := runChecks(w, hc)
+	success := runChecks(w, hc, options.output)
```
siggy: i'm realizing that when `-o json` is set, anything that isn't json should probably be written to stderr, to allow for tools to capture strictly the json.

rough (totally untested, not compiled) code on how it could work:

```go
func configureAndRunChecks(wout io.Writer, werr io.Writer, options *checkOptions) error {
...
	success := false
	w := wout
	if options.output == jsonOutput {
		success = runChecksJSON(wout, werr, hc)
		w = werr
	} else {
		success = runChecksBasic(wout, hc)	
	}
...
	fmt.Fprintf(w, "Status check results are %s\n", okStatus)

...

func runChecksJSON(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker) bool {
...
	resultJSON, err := json.MarshalIndent(outputJSON, "", "  ")
	if err == nil {
		fmt.Fprint(wout, string(resultJSON))
	} else {
		fmt.Fprintf(werr, "JSON serialization of the check result failed with %s", err)
	}

```

cli/cmd/check.go
alenkacz: good catch, thanks

cli/cmd/check.go
alenkacz: oh nice, I was not aware of that

cli/cmd/check.go
alenkacz: Hmm but how would I serialize the `"result": "success"`? I believe I need an extra struct for that anyway. I mean I can add that field to `CheckResult` but I would probably prefer to have those two structs separate. Then when you change the internal struct (CheckResult) you won't break the output for users. It's kind of API because scripts will depend on it so it should be stable at all times.

cli/cmd/check.go
alenkacz: oh you're absolutely right! Thanks!

alenkacz: I've decided to implement this a bit differently, but with equal result :)

cli/cmd/check.go
```diff
@@ -161,14 +169,93 @@ func runChecks(w io.Writer, hc *healthcheck.HealthChecker) bool {
...
+	checkWarn    checkResult = "warning"
+	checkErr     checkResult = "error"
+)
+
+func runChecksJSON(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker) bool {
+	var lastCategory healthcheck.CategoryID
+	var outputJSON []*checkCategory
+	var currentCategory *checkCategory
+
+	collectJSONOutput := func(result *healthcheck.CheckResult) {
```
ihcsim: Not 100% sure if this is equivalent to what we are trying to accomplish here, but here's a suggested implementation of `runChecksJSON()` without needing the new `checkCategory`, `check` and `checkResult` types:

```diff
diff --git a/cli/cmd/check.go b/cli/cmd/check.go
index a32b3556..13ca9f2c 100644
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -211,42 +211,15 @@ const (
 )
 
 func runChecksJSON(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker) bool {
-	var lastCategory healthcheck.CategoryID
-	var outputJSON []*checkCategory
-	var currentCategory *checkCategory
-
+	outputJSON := map[healthcheck.CategoryID][]*healthcheck.CheckResult{}
 	collectJSONOutput := func(result *healthcheck.CheckResult) {
-		if lastCategory != result.Category {
-			currentCategory = &checkCategory{
-				Name:   string(result.Category),
-				Checks: []*check{},
-			}
-			outputJSON = append(outputJSON, currentCategory)
-
-			lastCategory = result.Category
+		_, exists := outputJSON[result.Category]
+		if !exists {
+			outputJSON[result.Category] = []*healthcheck.CheckResult{}
 		}
 
 		if !result.Retry {
-			// ignore checks that are going to be retried, we want only final results
-			status := checkSuccess
-			if result.Err != nil {
-				status = checkErr
-				if result.Warning {
-					status = checkWarn
-				}
-			}
-
-			var hint string
-			if result.Err != nil && result.HintAnchor != "" {
-				hint = fmt.Sprintf("%s%s", healthcheck.HintBaseURL, result.HintAnchor)
-			}
-
-			currentCheck := &check{
-				Description: result.Description,
-				Result:      status,
-				Hint:        hint,
-			}
-			currentCategory.Checks = append(currentCategory.Checks, currentCheck)
+			outputJSON[result.Category] = append(outputJSON[result.Category], result)
 		}
 	}
```

Output:
```
{
  "kubernetes-api": [
    {
      "Category": "kubernetes-api",
      "Description": "can initialize the client",
      "HintAnchor": "k8s-api",
      "Retry": false,
      "Warning": false,
      "Err": null
    },
    {
      "Category": "kubernetes-api",
      "Description": "can query the Kubernetes API",
      "HintAnchor": "k8s-api",
      "Retry": false,
      "Warning": false,
      "Err": null
    }
  ],
  "kubernetes-version": [
    {
      "Category": "kubernetes-version",
      "Description": "is running the minimum Kubernetes API version",
      "HintAnchor": "k8s-version",
      "Retry": false,
      "Warning": false,
      "Err": null
    },
    {
      "Category": "kubernetes-version",
      "Description": "is running the minimum kubectl version",
      "HintAnchor": "kubectl-version",
      "Retry": false,
      "Warning": false,
      "Err": null
    }
  ],
  "linkerd-existence": [
    {
      "Category": "linkerd-existence",
      "Description": "control plane namespace exists",
      "HintAnchor": "l5d-existence-ns",
      "Retry": false,
      "Warning": false,
      "Err": {}
    }
  ]
}
```

cli/cmd/check.go
alenkacz: Yeah I see your point. It's not the nicest API for me though. If I got to choose what would be the easiest to consume for me as a writer of the script consuming this, I would probably go this way. And since we already have the structs in place... my choice would be to stick to them. But it's you guys's call in the end... :) I am just expressing my opinion

cli/cmd/check.go
alenkacz: nice! I removed two of those variables. I still kept the custom structs so the output is a bit different :)

cli/cmd/check.go
siggy: After playing with JSON output from both the user-facing `cmd.check` struct and marshaling directly from `healthcheck.CheckResult`, I agree with you that a user-facing version is preferable, as it will make consumption via scripts easier.

My primary concern with two sets of structs was maintenance overhead, specifically the risk of them diverging when introducing new fields/features. To that end, mind adding some comments above both `cmd.check` and `healthcheck.CheckResult` indicating how these two structs are related?

A few more changes:
- include the error message if a check fails
- only render `hint` and `error` when they are non-empty
- render a top-level `success` boolean, to allow for very simple scripting, rather than having to iterate over all the checks to validate whether it passed:
    ```json
    {
      "success": false,
      "categories": [
        {
          "categoryName": "kubernetes-api",
          "checks": [
    ```

Here's a diff encapsulating the above changes, let me know what you think:
```diff
diff --git a/cli/cmd/check.go b/cli/cmd/check.go
index 58ecb123..bfd61d51 100644
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -194,14 +194,22 @@ func runChecksTable(wout io.Writer, hc *healthcheck.HealthChecker) bool {
 	return success
 }
 
+type checkOutput struct {
+	Success    bool             `json:"success"`
+	Categories []*checkCategory `json:"categories"`
+}
+
 type checkCategory struct {
 	Name   string   `json:"categoryName"`
 	Checks []*check `json:"checks"`
 }
 
+// check is a user-facing version of `healthcheck.CheckResult`, for output via
+// `linkerd check -ojson`.
 type check struct {
 	Description string      `json:"description"`
-	Hint        string      `json:"hint"`
+	Hint        string      `json:"hint,omitempty"`
+	Error       string      `json:"error,omitempty"`
 	Result      checkResult `json:"result"`
 }
 
@@ -214,19 +222,19 @@ const (
 )
 
 func runChecksJSON(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker) bool {
-	var outputJSON []*checkCategory
+	var categories []*checkCategory
 
 	collectJSONOutput := func(result *healthcheck.CheckResult) {
 		categoryName := string(result.Category)
-		if outputJSON == nil || outputJSON[len(outputJSON)-1].Name != categoryName {
-			outputJSON = append(outputJSON, &checkCategory{
+		if categories == nil || categories[len(categories)-1].Name != categoryName {
+			categories = append(categories, &checkCategory{
 				Name:   categoryName,
 				Checks: []*check{},
 			})
 		}
 
 		if !result.Retry {
-			currentCategory := outputJSON[len(outputJSON)-1]
+			currentCategory := categories[len(categories)-1]
 			// ignore checks that are going to be retried, we want only final results
 			status := checkSuccess
 			if result.Err != nil {
@@ -236,21 +244,29 @@ func runChecksJSON(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker
 				}
 			}
 
-			var hint string
-			if result.Err != nil && result.HintAnchor != "" {
-				hint = fmt.Sprintf("%s%s", healthcheck.HintBaseURL, result.HintAnchor)
-			}
-
 			currentCheck := &check{
 				Description: result.Description,
 				Result:      status,
-				Hint:        hint,
 			}
+			if result.Err != nil {
+				currentCheck.Error = result.Err.Error()
+
+				if result.HintAnchor != "" {
+					currentCheck.Hint = fmt.Sprintf("%s%s", healthcheck.HintBaseURL, result.HintAnchor)
+				}
+			}
+
 			currentCategory.Checks = append(currentCategory.Checks, currentCheck)
 		}
 	}
 
 	result := hc.RunChecks(collectJSONOutput)
+
+	outputJSON := checkOutput{
+		Success:    result,
+		Categories: categories,
+	}
+
 	resultJSON, err := json.MarshalIndent(outputJSON, "", "  ")
 	if err == nil {
 		fmt.Fprint(wout, string(resultJSON))
diff --git a/pkg/healthcheck/healthcheck.go b/pkg/healthcheck/healthcheck.go
index 9760732c..afad9aae 100644
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -126,7 +126,9 @@ type checker struct {
 	checkRPC func(context.Context) (*healthcheckPb.SelfCheckResponse, error)
 }
 
-// CheckResult encapsulates a check's identifying information and output
+// CheckResult encapsulates a check's identifying information and output.
+// Note there exists an analogous user-facing type, `cmd.check`, for output via
+// `linkerd check -ojson`.
 type CheckResult struct {
 	Category    CategoryID
 	Description string
```

alenkacz:
@siggy ready for re-review :)

cli/cmd/check.go
```diff
@@ -216,14 +224,106 @@ func runChecks(w io.Writer, hc *healthcheck.HealthChecker) bool {
...
+	result := hc.RunChecks(collectJSONOutput)
+
+	outputJSON := checkOutput{
+		Success:    result,
+		Categories: categories,
+	}
+
+	resultJSON, err := json.MarshalIndent(outputJSON, "", "  ")
+	if err == nil {
+		fmt.Fprint(wout, string(resultJSON))
```
siggy: apologies i missed this on the last pass, mind adding a trailing newline to the output, for consistency with the other commands?
```go
fmt.Fprintf(wout, "%s\n", resultJSON)
```

cli/cmd/check.go
```diff
@@ -43,6 +46,7 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	flags.BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
+	flags.StringVarP(&options.output, "output", "o", options.output, "Output format. One of: basic, json")
```
siggy: `One of: table, json` ;)

ihcsim: @siggy Does this flag need to part of `linkerd check config`?

```
$ bin/go-run cli check config -o json
Error: unknown shorthand flag: 'o' in -o
Usage:
  linkerd check config [flags]

Examples:
  # Check that the Linkerd cluster-wide resource are installed correctly
  linkerd check config
```

siggy: @ihcsim nice catch!

@alenkacz per Ivan's comment, we recently added a `linkerd check config` subcommand, which should also benefit from your json work, you can make it available to the subcommand with:
```diff
diff --git a/cli/cmd/check.go b/cli/cmd/check.go
index 0b70d87e..a0744bc1 100644
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -46,7 +46,6 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
        flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
        flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
        flags.BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
-       flags.StringVarP(&options.output, "output", "o", options.output, "Output format. One of: basic, json")

        return flags
 }
@@ -56,6 +55,7 @@ func (options *checkOptions) checkFlagSet() *pflag.FlagSet {
        flags := pflag.NewFlagSet("check", pflag.ExitOnError)

        flags.StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
+       flags.StringVarP(&options.output, "output", "o", options.output, "Output format. One of: basic, json")
        flags.DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for all tests to pass")

        return flags
```

alenkacz:
@ihcsim @siggy ready

ihcsim:
@alenkacz @siggy Sorry for the delay. I was away last week. This works great!

PR #2667, opened 2019-04-08
Switch UUID implementation

olix0r:
The UUID implementation we use to generate install IDs is technically
not random enough for secure uses, which ours is not. To prevent
security scanners like SNYK from flagging this false-positive, let's
just switch to the other UUID implementation (Already in our
dependencies).


olix0r:
manually tested that `install --ignore-cluster` actually sets a uuid

PR #2678, opened 2019-04-09
Making `overview` page of dashboard functional on mobile devices

scottcarol:
This PR is a step towards making the Linkerd dashboard fully functional on mobile devices. 

For desktop mode, the only changes should be that the navigation icons below the horizontal line (for Docs, Community, Mailing List, Slack and GitHub) are now the same size as the icons above the line.

For mobile mode:
- all of the navigation items should be neatly lined up
- the browser defaults to a minimized sidebar if the width is below `960 px` (user can still manually toggle)
- tables horizontally scroll and are bounded by their external containers
- filter is turned off (this takes up too much room and is not practical since all of the table fields can't show) and icons removed
- the large `Namespace: ______` headers are removed because the namespace is already listed at the top of each section

For testing, I used the 'Responsive' DevTools mode in Chrome, Safari and Firefox, and also ran tests on SauceLabs with phone-sized browsers. However, testing on real devices is key and when this is merged I can ask the user who requested this feature for feedback.

There are still some views that aren't mobile-friendly, specifically `servicemesh`, which will be tackled in a later PR. 

Before screenshot (Chrome Desktop Browser on "Responsive" mode, Pixel 2):
Screen Shot 2019-04-09 at 3 21 12 PM

After screenshot (same browser):
Screen Shot 2019-04-09 at 3 20 05 PM

Note: there is a known issue in the Firefox mobile view where `ListItemIcon` is off-center, this will be addressed in a later PR.


web/app/js/components/NamespaceLanding.jsx
```diff
@@ -213,4 +227,4 @@ class NamespaceLanding extends React.Component {
   }
 }
 
-export default withContext(NamespaceLanding);
+export default withContext(withStyles(styles, { withTheme: true })(NamespaceLanding));
```
rmars: we're including `withTheme: true` here but we don't use the theme in our styles function, so this config can be omitted.

web/app/js/components/NamespaceLanding.jsx
```diff
@@ -19,6 +20,13 @@ import _isEmpty from 'lodash/isEmpty';
 import _isNil from 'lodash/isNil';
 import { friendlyTitle } from './util/Utils.js';
 import { withContext } from './util/AppContext.jsx';
+import { withStyles } from '@material-ui/core/styles';
+
+const styles = ({
```
rmars: I think styles has to be a function?
`const styles = () => ({`

web/app/js/components/Navigation.jsx
```diff
@@ -238,6 +241,15 @@ class NavigationBase extends React.Component {
     );
   }
 
+  updateWindowDimensions() {
+    let browserWidth = window.innerWidth;
+    if (browserWidth < 960) {
```
rmars: tioli consider putting this in a const somewhere `maxBrowserWidth`

scottcarol:
@grampelberg would you like to look this over?

grampelberg:
@scottcarol looking right now!

grampelberg:
This is a big improvement! There are some pages that were missed and I think we might want to rethink some things, but none of this is blocking for this specific PR as it is an immediate improvement from the current behavior. As a follow up, I'd recommend moving the menu to the top instead of the side (take a look at how bootstrap does it).

- Looks like tap/top are getting flattened (but the routes page is working).

Screenshot 2019-04-11 11 17 08

Screenshot 2019-04-11 11 17 16

- Service mesh page is funky.

Screenshot 2019-04-11 11 17 26

- When expanding resources, it isn't particularly usable.

Screenshot 2019-04-11 11 20 37

scottcarol:
Thanks @grampelberg. After merging this I'll create a new issue with your recommendations.

PR #2679, opened 2019-04-10
Integration test for the 'upgrade' command

### LINKED ISSUE #2669: upgrade integration tests
Introduce integration tests to validate `linkerd upgrade`.

The integration test should do the following:
- install last stable
- with current version, run `linkerd upgrade`
- validate with `linkerd check` the control-plane is healthy

We should modify the existing auto-injected integration test:
https://github.com/linkerd/linkerd2/blob/b454f8fbc10e239dba922afd33a4d81b36b2511e/bin/test-run#L61

### `bin/test-run`

```bash
# fresh install of current branch
run_test "$test_directory/install_test.go" || exit_code=$?

# stable install
curl https://run.linkerd.io/install | sh
$HOME/.linkerd2/bin/linkerd --proxy-auto-inject

# upgrade to current branch
run_test "$test_directory/install_test.go" "--upgrade" || exit_code=$?
```

### `test/install_test.go`

```diff
$ git diff
diff --git a/test/install_test.go b/test/install_test.go
index fd56898b..c87d233c 100644
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -102,13 +102,20 @@ func TestCheckPreInstall(t *testing.T) {
 }

 func TestInstall(t *testing.T) {
-       cmd := []string{"install",
-               "--controller-log-level", "debug",
-               "--proxy-log-level", "warn,linkerd2_proxy=debug",
-               "--linkerd-version", TestHelper.GetVersion(),
-       }
-       if TestHelper.AutoInject() {
-               cmd = append(cmd, []string{"--proxy-auto-inject"}...)
+       cmd := []string{}
+       if !TestHelper.Upgrade() {
+               cmd = append(cmd, "install",
+                       "--controller-log-level", "debug",
+                       "--proxy-log-level", "warn,linkerd2_proxy=debug",
+                       "--linkerd-version", TestHelper.GetVersion(),
+               )
+       } else {
+               cmd = append(cmd, "upgrade",
+                       "--controller-log-level", "debug",
+                       "--proxy-log-level", "warn,linkerd2_proxy=debug",
+                       "--linkerd-version", TestHelper.GetVersion(),
+                       "--proxy-auto-inject",
+               )
                linkerdDeployReplicas["linkerd-proxy-injector"] = deploySpec{1, []string{"proxy-injector"}}
        }
```

Part of #2459

ihcsim:
The extra steps to download and install an edge release increases the test duration by less than a minute, on my machine:

This branch:
```
=== PASS: all tests passed
bin/test-run `pwd`/bin/linkerd  33.16s user 4.85s system 9% cpu 6:33.27 total
```

`master`:
```
=== PASS: all tests passed
bin/test-run `pwd`/bin/linkerd  35.64s user 5.46s system 11% cpu 5:52.69 total
```

Fixes #2669 

### DIFF
--- a/bin/test-run
+++ b/bin/test-run
@@ -35,6 +35,28 @@ function run_test(){
     go test -v "$1" --linkerd="$linkerd_path" --linkerd-namespace="$linkerd_namespace" --k8s-context="$k8s_context" --integration-tests "$2"
 }
 
+function install_edge() {
+  tmp=$(mktemp -d -t l5dbin.XXX)
+  trap "rm -rf $tmp" EXIT
+
+  curl -s https://run.linkerd.io/install-edge | HOME=$tmp sh
+
+  local linkerd_path=$tmp/.linkerd2/bin/linkerd
+  $linkerd_path install --proxy-auto-inject --linkerd-namespace="$linkerd_namespace" | kubectl apply -f - > /dev/null 2>&1
+  $linkerd_path check --linkerd-namespace="$linkerd_namespace" > /dev/null 2>&1
+}
+
+function run_upgrade_test() {
+  local linkerd_namespace="$linkerd_namespace"-upgrade
+  local edge_version=$(curl -s https://versioncheck.linkerd.io/version.json | grep -o "edge-[0-9]*.[0-9]*.[0-9]*")
+
+  printf "Installing release [%s] namespace [%s]\n" "$edge_version" "$linkerd_namespace"
+  install_edge
+
+  printf "Upgrading release [%s] to [%s]\n" "$edge_version" "$linkerd_version"
+  run_test "$test_directory/install_test.go" "--upgrade-from-version=$edge_version" "--linkerd-namespace=$linkerd_namespace" || exit_code=$?
+}
+
 linkerd_path=$1
 
 if [ -z "$linkerd_path" ]; then
@@ -58,7 +80,8 @@ printf "Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\\n" "$linke
 exit_code=0
 
 run_test "$test_directory/install_test.go" || exit_code=$?
-run_test "$test_directory/install_test.go" "--proxy-auto-inject" || exit_code=$?
+
+run_upgrade_test
 for test in $(find "$test_directory" -mindepth 2 -name '*_test.go'); do
     run_test "$test" || exit_code=$?
 done
--- a/bin/test-run
+++ b/bin/test-run
@@ -31,30 +31,36 @@ function check_if_k8s_reachable(){
 }
 
 function run_test(){
-    printf "Running test [%s] %s\\n" "$(basename "$1")" "$2"
-    go test -v "$1" --linkerd="$linkerd_path" --linkerd-namespace="$linkerd_namespace" --k8s-context="$k8s_context" --integration-tests "$2"
+    filename="$1"
+    shift
+
+    printf "Running test [%s] %s\\n" "$(basename "$filename")" "$@"
+    go test -v "$filename" --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@"
 }
 
+# Install the latest edge release.
+# $1 - namespace to use for the edge release
 function install_edge() {
-  tmp=$(mktemp -d -t l5dbin.XXX)
-  trap "rm -rf $tmp" EXIT
+    tmp=$(mktemp -d -t l5dbin.XXX)
+    trap "rm -rf $tmp" EXIT
 
-  curl -s https://run.linkerd.io/install-edge | HOME=$tmp sh
+    curl -s https://run.linkerd.io/install-edge | HOME=$tmp sh > /dev/null 2>&1
 
-  local linkerd_path=$tmp/.linkerd2/bin/linkerd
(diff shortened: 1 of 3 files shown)


bin/test-run
```diff
@@ -35,6 +35,30 @@ function run_test(){
...
 
+function install_edge() {
+  set -x
+  tmp=$(mktemp -d -t l5dbin.XXX)
+  trap "rm -rf $tmp" EXIT
+
+  curl https://run.linkerd.io/install-edge | HOME=$tmp sh
+
+  local linkerd_path=$tmp/.linkerd2/bin/linkerd
+  $linkerd_path install --proxy-auto-inject --linkerd-namespace="$linkerd_namespace" | kubectl apply -f - > /dev/null 2>&1
```
siggy: i think we want to skip `--proxy-auto-inject` in the initial edge installation, and then introduce the auto injector when we do the upgrade. related, in `install_test.go`, this would also decouple the `--proxy-auto-inject` from the `TestHelper.UpgradeFrom()` setting, so the command a few lines below this becomes:

```bash
run_test "$test_directory/install_test.go" "--proxy-auto-inject" "--upgrade-from-version=$edge_version" "--linkerd-namespace=$linkerd_namespace" || exit_code=$?
```

test/install_test.go
```diff
@@ -101,18 +111,24 @@ func TestCheckPreInstall(t *testing.T) {
...
+		args = []string{
+			"--controller-log-level", "debug",
+			"--proxy-log-level", "warn,linkerd2_proxy=debug",
+			"--linkerd-version", TestHelper.GetVersion(),
+		}
+	)
+
+	if TestHelper.UpgradeFrom() != "" {
+		cmd = "upgrade"
+		args = append(args, "--proxy-auto-inject")
```
siggy: related to earlier comment, best to decouple `TestHelper.UpgradeFrom()` from `--proxy-auto-inject`

test/install_test.go
```diff
@@ -81,13 +82,22 @@ var (
 // Later tests depend on the success of earlier tests
 
 func TestVersionPreInstall(t *testing.T) {
-	err := TestHelper.CheckVersion("unavailable")
+	version := "unavailable"
+	if TestHelper.UpgradeFrom() != "" {
+		version = TestHelper.UpgradeFrom()
+	}
+
+	err := TestHelper.CheckVersion(version)
```
siggy: nice!

bin/test-run
```diff
@@ -35,6 +35,30 @@ function run_test(){
...
+  $linkerd_path check --linkerd-namespace="$linkerd_namespace" > /dev/null 2>&1
+  set +x
+}
+
+function run_upgrade_test() {
+  local linkerd_namespace="$linkerd_namespace"-upgrade
```
siggy: my bash foo is not great, does this mean that `$linkerd_namespace` becomes something like `l5d-integration-upgrade` when calling into `install_edge()`? if so, i'd prefer we use a different var name, or just pass a namespace in, that way it's clear reading through the script that `$linkerd_namespace` always means the same thing.

bin/test-run
```diff
@@ -31,8 +31,36 @@ function check_if_k8s_reachable(){
 }
 
 function run_test(){
-    printf "Running test [%s] %s\\n" "$(basename "$1")" "$2"
-    go test -v "$1" --linkerd="$linkerd_path" --linkerd-namespace="$linkerd_namespace" --k8s-context="$k8s_context" --integration-tests "$2"
+    filename="$1"
+    shift
```
ihcsim: It's like `args[1:]`.

bin/test-run
```diff
@@ -31,8 +31,36 @@ function check_if_k8s_reachable(){
...
+
+    printf "Running test [%s] %s\\n" "$(basename "$filename")" "$@"
+    go test -v "$filename" --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@"
```
ihcsim: Replaced the `"$2"` with `"$@"`, because the former converts multiple arguments into one long, space-delimited argument. For example, when running `run_test filename --linkerd-namespace=linkerd --proxy-auto-inject`, the `testutil.TestHelper` ends up treating the linkerd namespace as `linkerd --proxy-auto-inject`.

ihcsim: Also, removed. the `--linkerd-namespace="$linkerd_namespace"` argument. This should be part of `$@`, instead of relying on the global `$linkerd_namespace` variable. This was problematic for upgrade and auto-inject test cases which depend on the control plane in a different namespace.

bin/test-run
```diff
@@ -31,8 +31,36 @@ function check_if_k8s_reachable(){
...
+    curl -s https://run.linkerd.io/install-edge | HOME=$tmp sh > /dev/null 2>&1
+
+    local linkerd_path=$tmp/.linkerd2/bin/linkerd
+    local edge_namespace=${1}
+    $linkerd_path install --linkerd-namespace="$edge_namespace" | kubectl apply -f - > /dev/null 2>&1
+    $linkerd_path check --linkerd-namespace="$edge_namespace" > /dev/null 2>&1
+}
+
+function run_upgrade_test() {
+    local edge_namespace="$linkerd_namespace"-upgrade
```
siggy: with this being the only remaining place where we reference the global `linkerd_namespace`, mind just passing this in a param?

```bash
run_upgrade_test "$linkerd_namespace"-upgrade
```

PR #2685, opened 2019-04-11
add service profile integration tests for service profile metrics

### LINKED ISSUE #2518: Introduce ServiceProfile test plan

dadjeibaah:
This PR is a continuation of #2638. It adds tests that verify that retries are being performed after service profiles are added for a service and also tests to see if timeout limits for a particular route are respected. 

Fixes #2518


test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -117,7 +135,153 @@ func TestServiceProfiles(t *testing.T) {
...
+			case "retries":
+				// If the effective success rate is not equal to the actual success rate retries might already
+				// be applied so we fail the test.
+				assertion.assertFunc = func(rt *rowStat) bool { return rt.EffectiveSuccess == rt.ActualSuccess }
+				assertion.routeProperty = "Effective Success"
+				assertion.expected = "Effective Success == Actual Success"
+			case "timeouts", "budgets":
+				// If the P99 latency is greater than 500ms retries are probably happening before applying
+				// the service profile and we can't reliably test the service profile.
+				assertion.assertFunc = func(rt *rowStat) bool { return rt.LatencyP99 < 500 }
```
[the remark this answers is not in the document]
dadjeibaah: Good point. Testing for latency is pretty finicky, however, the behaviour we are testing is timeouts. The intent for this test case was to ensure that we aren't already hitting the retry timeout that we will eventually test for. My thinking here was, if we are already seeing timeouts surpassing the threshold even before applying the service profile with the intended timeout, we can't reliably prove that the service profile is doing its job after it's applied.

Another interesting thing here is with `--failure-rate=1.0`,  we will still see `effective_success` and `actual_success` be zero even after we've applied the service profile and the service is retrying requests. Similar to your suggestion, I think comparing `actual_rps` and `effective_rps` would suffice. 🤔

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -117,7 +135,153 @@ func TestServiceProfiles(t *testing.T) {
...
+				// If we get an effective success rate of less than or equal to the actual success rate requests are not
+				// being retried successfully after we applied our modified service profile.
+				assertion.assertFunc = func(rt *rowStat) bool { return rt.EffectiveSuccess > rt.ActualSuccess }
+				assertion.routeProperty = "Effective Success"
+				assertion.expected = "> Actual Success"
+				// If we get a P99 latency of less than 250ms then we aren't hitting the timeout limit
+				// set in in service profile. hello-timeouts-service and hello-budgets always fails
+				// so we expect all request latencies to be greater than or equal to the timeout set.
+			case "timeouts", "budgets":
+				assertion.assertFunc = func(rt *rowStat) bool { return rt.LatencyP99 >= 500 }
```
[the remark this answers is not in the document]
dadjeibaah: From my understanding of the initial ticket, I thought we were trying to specifically test the `timeouts` behavior. Maybe, since its unreliable we should just test that retries are being performed?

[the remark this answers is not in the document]
ihcsim: > Maybe, since its unreliable we should just test that retries are being performed?

SGTM, unless others have a way to do this. It isn't easy to reliably test timeouts without [`context`](https://golang.org/pkg/context/). So far, I can't find anything in the service profile and route code that will be useful here, based on the little that I know.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -1,16 +1,26 @@
...
+	"time"
 
+	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
 	"github.com/linkerd/linkerd2/testutil"
+	"sigs.k8s.io/yaml"
 )
 
 var TestHelper *testutil.TestHelper
 
+type rowStat struct {
```
siggy: TIOLI, you could make `jsonRouteStats` (from the `cmd` package) public and re-use that.

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -53,23 +70,23 @@ func TestServiceProfiles(t *testing.T) {
 		{
 			sourceName:     "tap",
 			namespace:      testNamespace,
-			deployName:     "deploy/t1",
+			deployName:     "deployment/t1",
```
siggy: curious: was this change necessary to match against returned data from a routes command?

test/serviceprofiles/testdata/tap_application.yaml
```diff
@@ -184,3 +184,4 @@ spec:
         ports:
         - containerPort: 9999
       restartPolicy: OnFailure
+
```
siggy: ws?

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -117,7 +134,135 @@ func TestServiceProfiles(t *testing.T) {
...
+					return fmt.Errorf(
+						"expected Effective Success to be equal to Actual Success but got: Effective [%.2f] <> Actual [%.2f]",
+						stat.EffectiveSuccess, stat.ActualSuccess)
+				}
+				return nil
+			})
+
+			profile := &sp.ServiceProfile{}
+
+			// Grab the output and convert it to a service profile object for modification
```
siggy: 🆒

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -117,7 +134,135 @@ func TestServiceProfiles(t *testing.T) {
...
+						"expected Effective Success to be greater than Actual Success but got: Effective [%f] <> Actual [%f]",
+						stat.EffectiveSuccess, stat.ActualSuccess)
+				}
+				return nil
+			})
+		})
+	}
+}
+
+func assertRouteStat(assertion *routeStatAssertion, t *testing.T, assertFn func(stat *rowStat) error) {
```
siggy: tioli: since `routeStatAssertion` is only used in one place, i'd probably just pass `upstream`, `downstream`, and `namespace` directly into `assertRouteStat`

test/serviceprofiles/serviceprofiles_test.go
dadjeibaah: Yea, I had to use the full `deployment` name in order to get stuff back. I can make a change to the routes command to accept both `deploy` and `deployment`

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -137,25 +301,26 @@ func assertExpectedRoutes(expected, actual []string, t *testing.T) {
...
-func parseRouteDetails(cliOutput string) []string {
-	var cliLines []string
-	routesByDeployment := strings.Split(cliOutput, "\n")
-	for _, routes := range routesByDeployment {
-		routes = strings.TrimSpace(routes)
-		if routes != "" && !strings.HasPrefix(routes, "ROUTE") {
-			cliLines = append(cliLines, routes)
-		}
+	if isWideOutput {
+		cmd = append(cmd, "-owide")
```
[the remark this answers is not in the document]
dadjeibaah: Ah nice! thanks for pointing that out, I had no idea.

Issue #2630, opened 2019-04-03
Proxy starts after inject then connection refused


vic3lord:
## Bug Report

I have injected the proxy into one specific app (most apps work) and the proxy fails to connect to the service

### What is the issue?

linkerd2_proxy::app::errors unexpected error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:8080)

### How can it be reproduced?

I couldn't find a way to reproduce this error on other services, this service specifically is a Go service that publishes data to Google Pubsub (gRPC) which is external to the mesh

### Logs, error output, etc

linkerd2_proxy::app::errors unexpected error: an error occurred trying to connect: Connection refused (os error 111) (address: 127.0.0.1:8080)

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version
√ is running the minimum kubectl version

linkerd-existence
-----------------
√ control plane namespace exists
√ controller pod is running
√ can initialize the client
√ can query the control plane API

linkerd-api
-----------
√ control plane pods are ready
√ control plane self-check
√ [kubernetes] control plane can talk to Kubernetes
√ [prometheus] control plane can talk to Prometheus
√ no invalid service profiles

linkerd-version
---------------
√ can determine the latest version
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date
√ control plane and cli versions match

Status check results are √
```

### Environment

- Kubernetes Version: 1.12.5-gke.10
- Cluster Environment: GKE
- Host OS: Google COS
- Linkerd version: edge-19.4.1

grampelberg:
That error message makes me wonder what the headers are. You wouldn't happen to have a tcpdump would you?

vic3lord:
It's distroless image but I will try to do some tcpdump even though this exact service worked just fine with linkerd `edge-19.3.1` and when upgrading I had to uninject so it looks like it's a linkerd-proxy issue..

LMK if I can provide you with more details and I will try to run tcpdump later on

grampelberg:
Hmm, if it was working with `edge-19.3.1` that invalidates my assumption there. `tcpdump` probably won't get everything I was hoping. Is there anything interesting in the logs if you bump the level up?

olix0r:
[...]

Issue #2585, opened 2019-03-28
Proxy restarted due to readiness/liveness probe failure


siggy:
## Background

Using the test-scale environment introduced in #2578, observed pod restarts due to failed liveness and readiness checks in `linkerd-proxy`.

### Environment

- AKS
- 15 node x Standard DS3 v2 (4 vcpus, 14 GB memory)
- K8s v1.11.8

### Deploy

```bash
bin/test-scale `pwd`/target/cli/darwin/linkerd
```

### Observe restarted pods

```bash
$ kubectl --all-namespaces=true get po | grep lifecycle | grep bb-broadcast | grep -v ' 0 '
l5d-scale-lifecycle-1   bb-broadcast-687d4448d8-8j57x           2/2       Running       1          10h
l5d-scale-lifecycle-2   bb-broadcast-687d4448d8-4w7mt           2/2       Running       2          10h
l5d-scale-lifecycle-2   bb-broadcast-687d4448d8-nzvvh           2/2       Running       1          10h
l5d-scale-lifecycle-3   bb-broadcast-687d4448d8-5fmrs           2/2       Running       1          10h
l5d-scale-lifecycle-3   bb-broadcast-687d4448d8-g7rbv           2/2       Running       1          10h
l5d-scale-lifecycle-3   bb-broadcast-687d4448d8-w6sb5           2/2       Running       2          10h
l5d-scale-lifecycle-4   bb-broadcast-687d4448d8-f7dz6           2/2       Running       1          10h
l5d-scale-lifecycle-5   bb-broadcast-687d4448d8-dt9wm           2/2       Running       1          10h
l5d-scale-lifecycle-5   bb-broadcast-687d4448d8-h7726           2/2       Running       1          10h
l5d-scale-lifecycle-5   bb-broadcast-687d4448d8-z2vrs           2/2       Running       1          10h
```

```bash
$ kubectl -n l5d-scale-lifecycle-4 describe po/bb-broadcast-687d4448d8-f7dz6
...
Events:
  Type     Reason     Age               From                                Message
  ----     ------     ----              ----                                -------
  Warning  Unhealthy  3m (x35 over 8h)  kubelet, aks-nodepool1-35313189-13  Liveness probe failed: Get http://10.244.5.70:4191/metrics: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
[...]

PR #2689, opened 2019-04-11
Update proxy to fix a connection starvation issue

### LINKED ISSUE #2581: Live/ready probes timing out on application container
### What is the issue?
After upgrading to `19.3.3` we started seeing intermittent pod restarts. The liveness probe was timing out and k8s was killing the pod. It seems like the proxy was blocking the request getting through to our application container. The proxy would log the following:

`WARN [ 28995.080301s] proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::transport::addr_info::linux failed to read SO_ORIGINAL_DST: Os { code: 2, kind: NotFound, message: "No such file or directory" }`

`ERR! [ 28995.080743s] proxy={server=in listen=0.0.0.0:4143 remote=10.0.1.1:46102} linkerd2_proxy::proxy::http::router could not recognize request`

### How can it be reproduced?
We haven't been able to reproduce the whole thing, but we have recognised patterns. Only pods receiving traffic from our ingress controller are affected. We're using traefik so this seems very similar to this issue https://github.com/linkerd/linkerd2/issues/1861

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
$ linkerd check
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane self-check
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √
```

### Environment

- Kubernetes Version: 1.12.5
- Cluster Environment: GKE
- Host OS: cos
- Linkerd version: edge-19.3.3

olix0r:
In https://github.com/linkerd/linkerd2-proxy/pull/233, we fixed an issue
in the proxy where, when the proxy performed TLS discovery (on inbound
connections), detection on a slow or idle connection could block all
other connections from being accepted on the listener.

This would frequently cause liveness probe failures on application
containers, causing pods to go into CrashLoopBackoff.

Fixes #2581 #2585 #2630

### DIFF
--- a/bin/docker-build-proxy
+++ b/bin/docker-build-proxy
@@ -14,6 +14,6 @@ rootdir="$( cd $bindir/.. && pwd )"
 . $bindir/_tag.sh
 
 # Default to a pinned commit SHA of the proxy.
-PROXY_VERSION="${PROXY_VERSION:-b634c3a}"
+PROXY_VERSION="${PROXY_VERSION:-c96dd3a}"
 
 docker_build proxy "$(head_root_tag)" $rootdir/Dockerfile-proxy --build-arg PROXY_VERSION=$PROXY_VERSION


olix0r:
pending https://travis-ci.org/linkerd/linkerd2-proxy/jobs/518918675

olix0r:
integration tests ran before the binary was available

PR #2696, opened 2019-04-15
Get rid of references to minikube in docs

### LINKED ISSUE #2600: Move away from minikube
Move away from minikube for local development 
Possible replacements:
- [kind](https://github.com/kubernetes-sigs/kind) 
- Docker-for-Desktop 's inbuilt k8s environment

Both have their own pros and cons:

#### Pros:
- Should be helpful for Conformance Validation #1096 with latest k8s releases 
- Current default environment used in prow/testing jobs for k8s addons like [ingress-nginx](https://github.com/kubernetes/ingress-nginx) and istio
- Supports latest k8s versions, one of the first to support 1.14 release
- Is a [CNCF certified conformant Kubernetes installer](https://landscape.cncf.io/selected=kind)
- Supports Windows in addition to MacOS and Linux
#### Cons:
- Requires Docker-for-Desktop 
- Need to load images into the `kind cluster` after building. 
   Here's the one-liner I use:
 ```sh
docker images  --filter "dangling=false" | grep gcr.io/linkerd-io | \ 
awk ' {printf "%s:%s\n", $1, $2}' | xargs -n 1 kind load docker-image
```
#### Pros:
- One less dependency 
- No need to load images after building  
- One click k8s cluster creation through GUI 
#### Cons:
- Runs a not so recent v1.10.11 node
- Windows support?

What do you folks think? I will be happy to implement either

cc @grampelberg @olix0r

### PR DESCRIPTION
**Pull request by epk**
Closes #2600 

Signed-off-by: Aditya Sharma <git@adi.run>

### CODE REVIEW

File: BUILD.md [hunk 1]
```diff
@@ -99,22 +99,25 @@ linkerd2_components
...
 - [Web](#web): Development of the Linkerd2 Dashboard.
 
 ### Comprehensive
 
 This configuration builds all Linkerd2 components in Docker images, and deploys
-them onto Minikube. This setup most closely parallels our recommended production
+them onto local kubernetes cluster. This setup most closely parallels our recommended production
 installation, documented at https://linkerd.io/2/getting-started/.
 
 These commands assume a working
```
grampelberg: Mind just linking to the k8s cluster install docs? That way we're not picking a specific one.

TEST.md
```diff
@@ -160,14 +160,15 @@ You can also test a locally-built version of the `linkerd` CLI. Note, however,
 that this requires that you build the corresponding Linkerd docker images and
 publish them to a docker registry that's accessible from the Kubernetes cluster
 where you're running the tests. As a result, local testing mostly applies to
-[minikube](https://github.com/kubernetes/minikube), since you can build the
-images directly into minikube's local docker registry, as described below.
+[Kubernetes for Docker](https://blog.docker.com/2018/01/docker-mac-kubernetes/)
```
grampelberg: Same comment as above, let's just link to the options from the k8s docs.

BUILD.md
paranoidaditya: Sorry for the delay in getting back to this. 
I am not 100% sure about what you are referring to by `k8s cluster install docs`?

grampelberg: https://kubernetes.io/docs/setup/

PR #2697, opened 2019-04-15
Introduce upgrade --from-manifests flag

### LINKED ISSUE #2629: RFC: load upgrade configuration from kubernetes manifest

We use the gitops method for maintaining our clusters and deploying services. This involves writing kubernetes manifests (or kustomize, or helm charts) to a config repo and ArgoCD applies them to the cluster specified.

For example:

```text
infrastructure
  \__ linkerd
         \__ azure
                \__ dev-ea
                \__ qa-scu
                \__ prod-scu
```

each folder under the ```azure``` folder is a specific cluster and contains the output of `linkerd install --ha --proxy-auto-inject > install.yaml` as well as an ingress for the web-ui, etc.

With the new upgrade mechanism (and identity), we can no longer do a `linkerd install --ha --proxy-auto-inject > install.yaml` (even with `--ignore-cluster`) and expect identity to stay happy (due to install creating new keys, etc).

If we do a `linkerd upgrade > install.yaml` it will take the config from whichever cluster is my current kubernetes context. Therefore if I wish to upgrade ```prod-scu``` and I'm not paying attention, it may write config values from ```dev-ea```.

The addition of a `--from-manifest` flag for the `linkerd upgrade` command. Instead of reading the config from the `linkerd-config` configmap, it reads it from the manifest created by a prior `linkerd install --flags... > install.yaml`

Nothing else on the linkerd side comes to mind. We could probably write some tooling that does the upgrade in a way that mitigates the risk outlined above.

```linkerd upgrade --from-manifests install.yaml > install.yaml```

### PR DESCRIPTION
**Pull request by siggy**
The `linkerd upgrade` command read the control-plane's config from
Kubernetes, which required the environment to be configured to connect
to the appropriate k8s cluster.

Introduce a `linkerd upgrade --from-manifests` flag, allowing the user
to feed the output of `linkerd install` into the upgrade command.

Fixes #2629

Relates to #2637.

## Notes

Note that the first commit, 5fb1c0c, contains all the code changes. The second commit simply renames `test_helper.go` to `fake.go`.

### Install flow

```bash
linkerd install > install_manifests.yaml
kubectl apply -f install_manifests.yaml
linkerd upgrade --from-manifests install_manifests.yaml
```

### Validate `linkerd upgrade` command works the same with and without `--from-manifests`

```bash
linkerd install | kubectl apply -f -

diff <(
  bin/linkerd upgrade
) <(
  kubectl -n linkerd get cm/linkerd-config secrets/linkerd-identity-issuer -oyaml |
    bin/linkerd upgrade --from-manifests -
)
```


siggy:
@jon-walton if you have a moment, please take this out for a spin, thanks!

jon-walton:
awesome, that works :+1: thanks! 🎉

cli/cmd/upgrade.go
```diff
@@ -22,12 +22,16 @@ const (
...
-)
+type upgradeOptions struct {
+	manifests string
+	*installOptions
+}
 
 func newUpgradeOptionsWithDefaults() *upgradeOptions {
-	return &upgradeOptions{newInstallOptionsWithDefaults()}
+	return &upgradeOptions{
+		"",
```
olix0r: tioli: named params might be easier to read here

grampelberg:
@olix0r because they show up in the manifest?

ihcsim:
Just wanna confirm that we are ok with writing the issuer's secret to disk.

siggy:
@ihcsim Re: writing secrets to disk, I think that's a good question for @grampelberg, and if the answer is "yes", maybe deserves treatment in our docs.

olix0r:
@grampelberg right. Ideally, we'd encrypt that part of the manifest and have `--from-manifest` decrypt the secrets

ihcsim:
fwiw, I think all #2629 cares about is the `linkerd-config` config map.

alpeb:
Another possibility would be to add a note in the Upgrade CLI docs, telling people they can store all the resources in disk, but discouraging keeping the Secret one. Then we could have `fetchIdentityValues()` not fail when the Secret is not found when using `--from-manifests`. In my brief testing this doesn't break anything.
One downside I can see is when trying to do `kubectl apply --prune` with resource files in disk that don't match what there is in the cluster, one has to be careful not to delete the Secret in the cluster.

jon-walton:
This is an interesting point.

However, when using a gitops workflow, a `linkerd install --flags > install.yaml` would end up in git and deployed by the deployment controller (in my case, ArgoCD). if the secret is kept out of the manifest, we'll need linkerd to create the secret and store it in a k8s secret object at runtime

grampelberg:
Is there another option to writing the secret to disk? It'd need to live somewhere ...

From a docs perspective, it is pretty easy to suggest piping through `gpg` or the like before committing the file.

ihcsim:
> if the secret is kept out of the manifest, we'll need linkerd to create the secret and store it in a k8s secret object at runtime

The (identity) secret is created during installation, and remained untouched, in the cluster, during an upgrade.

siggy:
Going to merge this as-is. I agree storing secrets in the clear on disk is not great, but also agree we can guide users via docs on best practices, such as @grampelberg's `gpg` approach.

grampelberg:
Can we open an issue on the website repo to track writing up a blurb about managing the manifest?

siggy:
@grampelberg filed linkerd/website#275

PR #2701, opened 2019-04-16
Release notes for stable-2.3.0

siggy:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -103,8 +103,6 @@ upgrading to this release, please see the
   * Updated the `mutatingwebhookconfiguration` so that it is recreated when the
     proxy injector is restarted, so that the MWC always picks up the latest
     config template during version upgrade
-  * Updated the auto-inject admission controller webhook to respond to UPDATE
-    events for deployment workloads
   * Updated destination service to return TLS identities only when the
     destination pod is TLS-aware and is in the same controller namespace
   * Lessen klog level to improve security
@@ -121,7 +119,6 @@ upgrading to this release, please see the
   * **Fixed** a connection starvation issue where TLS discovery detection on
     slow or idle connections could block all other connections from being
     accepted on the inbound listener of the proxy
-  * The `CommonName` field on CSRs is now set to the proxy's identity name
   * Some `l5d-*` informational headers have been temporarily removed from
     requests and responses because they could leak information to external
     clients
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ Among the many improvements to the web dashboard, we've added a Community page
 to surface news and updates from linkerd.io.
 
 For more details, see the announcement blog post:
-https://blog.linkerd.io/2019/04/16/announcing-linkerd-2-3/
+https://linkerd.io/2019/04/16/announcing-linkerd-2.3/
 
 To install this release, run: `curl https://run.linkerd.io/install | sh`
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,7 +22,7 @@ To install this release, run: `curl https://run.linkerd.io/install | sh`
 `--tls=optional`, manually delete the `linkerd-ca` deployment after upgrading.
 Also, `--single-namespace` mode is no longer supported. For full details on
 upgrading to this release, please see the
-[upgrade instructions](https://linkerd.io/2/upgrade/#upgrade-notice-stable-2-3-0).
+[upgrade instructions](https://linkerd.io/2/tasks/upgrade/#upgrade-notice-stable-2-3-0).
 
 **Special thanks to**: @codeman9, @harsh-98, @huynq0911, @KatherineMelnyk,
 @liquidslr, @paranoidaditya, @Pothulapati, @TwinProduction, and @yb172!
@@ -456,7 +456,7 @@ To install this release, run: `curl https://run.linkerd.io/install | sh`
 
 **Upgrade notes**: The default behavior for proxy auto injection and service
 profile ownership has changed as part of this release. Please see the
-[upgrade instructions](https://linkerd.io/2/upgrade/#upgrade-notice-stable-2-2-0)
+[upgrade instructions](https://linkerd.io/2/tasks/upgrade/#upgrade-notice-stable-2-2-0)
 for more details.
 
 **Special thanks to**: @alenkacz, @codeman9, @jonrichards, @radu-matei, @yeya24,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,195 @@
+## stable-2.3.0
+
+### Announcing Linkerd 2.3 🎈
+
+This stable release introduces a new TLS Identity system into the default
+Linkerd installation, replacing `tls=optional` and the `linkerd-ca` controller.
+Now, proxies generate ephemeral private keys into a tmpfs directory and
+dynamically refresh certificates, authenticated by Kubernetes ServiceAccount
+tokens, via the newly-introduced Identity controller.
+
+Now, all meshed HTTP communication is private and authenticated by default.
+
+In the CLI, this release includes 2 new commands:
+* `linkerd metrics` fetches metrics directly from a Linkerd proxy
+* `linkerd upgrade` upgrades your existing Linkerd installation
+
+Among the many improvements to the Web UI, we've added a Community page to
+surface news and updates from linkerd.io.
+
+For more details, see the announcement blog post:
+https://blog.linkerd.io/2019/04/16/announcing-linkerd-2-3/
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: The `linkerd-ca` controller is deprecated in favor of the
+`linkerd-identity` controller. If you had previously installed Linkerd with
+`tls=optional`, you will need to manually delete `linkerd-ca` after upgrading.
+Also, `single-namespace` mode is no longer supported. For full details on
+upgrading to this release, please see the
+[upgrade instructions](https://linkerd.io/2/upgrade/#upgrade-notice-stable-2-3-0).
+
+**Special thanks to**: @codeman9, @harsh-98, @huynq0911, @KatherineMelnyk,
+@liquidslr, @paranoidaditya, @Pothulapati, @TwinProduction, and @yb172!
+
+**Full release notes**:
+
+* CLI
+  * Eliminated false-positive vulnerability warnings related to go.uuid
+  * Removed TLS metrics from the `stat` command; this is in preparation for
+    surfacing identity metrics in a clearer way
+  * **Breaking Change:** The `--linkerd-cni-enabled` flag has been removed from
+    the `inject` command; CNI is configured at the cluster level with the
+    `install` command and no longer applies to the `inject` command
+  * Introduced an `upgrade` command! This allows an existing Linkerd control
+    plane to be reinstalled or reconfigured; it is particularly useful for
+    automatically reusing flags set in the previous `install` or `upgrade`
+  * The `inject` command proxy options are now converted into config
+    annotations; the annotations ensure that these configs are persisted in
+    subsequent resource updates
+  * The `stat` command now always shows the number of open TCP connections
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the
+    `install` command; external profiles are now disabled by default and can be
+    enabled with the new `--enable-external-profiles` flag
+  * Changed `install` to accept or generate an issuer Secret for the Identity
+    controller
+  * Changed `install` to fail in the case of a conflict with an existing
+    installation; this can be disabled with the `--ignore-cluster` flag
+  * Changed `inject` to require fetching a configuration from the control plane;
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,4 +1,195 @@
...
+    config template during version upgrade
+  * Updated the auto-inject admission controller webhook to respond to UPDATE
+    events for deployment workloads
+  * Updated destination service to return TLS identities only when the
+    destination pod is TLS-aware and is in the same controller namespace
+  * Lessen klog level to improve security
+  * Updated control-plane components to query Kubernetes at startup to determine
+    authorized namespaces and if ServiceProfile support is available
+  * Modified the stats payload to include the following TCP stats:
+    `tcp_open_connections`, `tcp_read_bytes_total`, `tcp_write_bytes_total`
```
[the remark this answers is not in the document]
siggy: hmm, i'm inclined to keep this in, as it's a public-api change, which i think in theory we want to be a user-facing supportable api.

siggy:
@alpeb i like the idea of sorting by importance, another change incoming...

CHANGES.md
```diff
@@ -1,4 +1,192 @@
...
+  * Fixed permissions check for CRDs
+  * Included kubectl version check as part of `linkerd check` (thanks @yb172!)
+  * Added TCP stats to the stat command, under the `-o wide` and `-o json` flags
+  * Updated `linkerd check` to ensure hint URLs are displayed for RPC checks
+  * Implemented `--proxy-cpu-limit` and `--proxy-memory-limit` for setting the
+    proxy resources limits (`--proxy-cpu` and `--proxy-memory` were deprecated in
+    favor of `proxy-cpu-request` and `proxy-memory-request`) (thanks @TwinProduction!)
+  * Updated the `inject` and `uninject` subcommands to issue warnings when
+    resources lack a `Kind` property (thanks @Pothulapati!)
+  * Unhid the `install-cni` command and its flags, and tweaked their descriptions
```
rmars: tioli: `exposed` instead of `unhid`?

CHANGES.md
```diff
@@ -1,4 +1,192 @@
...
+    responses
+  * Fixed issue with proxy falling back to filesystem polling due to improperly
+    sized inotify buffer
+* Web UI
+  * The Overview page in the Linkerd dashboard now renders appropriately when
+    viewed on mobile devices
+  * Removed TLS columns from the dashboard tables; this is in preparation for
+    surfacing identity metrics in a clearer way
+  * The topology graph now shows TCP stats if no HTTP stats are available
+  * The resource detail page no longer shows blank tables if the resource only
```
rmars: tioli: maybe `Improved table display on the resource detail page for resources with TCP-only traffic`

CHANGES.md
```diff
@@ -1,4 +1,189 @@
+## stable-2.3.0
+
+### Announcing Linkerd 2.3 🎈
+
+This stable release introduces a new TLS Identity system into the default
+Linkerd installation, replacing `tls=optional` and the `linkerd-ca` controller.
```
klingerf: I recommending using `--tls=optional` to indicate that it's a flag. That's how flags are handled in the individual bullet points below.

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+https://blog.linkerd.io/2019/04/16/announcing-linkerd-2-3/
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: The `linkerd-ca` controller is deprecated in favor of the
+`linkerd-identity` controller. If you had previously installed Linkerd with
+`tls=optional`, you will need to manually delete `linkerd-ca` after upgrading.
+Also, `single-namespace` mode is no longer supported. For full details on
+upgrading to this release, please see the
+[upgrade instructions](https://linkerd.io/2/upgrade/#upgrade-notice-stable-2-3-0).
```
scottcarol: Right now when I go to this URL, I just get the "Upgrading Linkerd" page which starts with `Upgrade notice: stable-2.2.0` - will this be changing before this goes out?

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+Linkerd installation, replacing `tls=optional` and the `linkerd-ca` controller.
+Now, proxies generate ephemeral private keys into a tmpfs directory and
+dynamically refresh certificates, authenticated by Kubernetes ServiceAccount
+tokens, via the newly-introduced Identity controller.
+
+Now, all meshed HTTP communication is private and authenticated by default.
+
+In the CLI, this release includes 2 new commands:
+* `linkerd metrics` fetches metrics directly from a Linkerd proxy
+* `linkerd upgrade` upgrades your existing Linkerd installation
```
klingerf: This command won't work for upgrading from `stable-2.2.1` to `stable-2.3.0`, right? For users upgrading from 2.2, we're still recommending that they use `linkerd install`? Maybe we could note that here and link to an upgrade guide that explains in more detail?

CHANGES.md
siggy: ah nice catch, i should have left a note that this URL, and the blog post, aren't live yet, will do that now...

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+    for all kinds of workloads, including StatefulSets, DaemonSets, Jobs, etc
+  * Service profile validation is now performed via a webhook endpoint; this
+    prevents Kubernetes from accepting invalid service profiles
+  * Changed the default CPU request from `10m` to `100m` for HA deployments;
+    this will help some intermittent liveness/readiness probes from failing due
+    to tight resource constraints
+  * Updated destination service to return TLS identities only when the
+    destination pod is TLS-aware and is in the same controller namespace
+  * Lessen klog level to improve security
+  * Updated control-plane components to query Kubernetes at startup to determine
```
scottcarol: Question, should there be a dash between `control` and `plane` here? (I know it was taken from a previous release)

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+Among the many improvements to the Web UI, we've added a Community page to
+surface news and updates from linkerd.io.
+
+For more details, see the announcement blog post:
+https://blog.linkerd.io/2019/04/16/announcing-linkerd-2-3/
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: The `linkerd-ca` controller is deprecated in favor of the
+`linkerd-identity` controller. If you had previously installed Linkerd with
```
klingerf: Deprecated implies to me that there's still legacy support, but that's not the case. How about:

> The `linkerd-ca` controllers has been removed in favor of the `linkerd-identity` controller.

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+
+In the CLI, this release includes 2 new commands:
+* `linkerd metrics` fetches metrics directly from a Linkerd proxy
+* `linkerd upgrade` upgrades your existing Linkerd installation
+
+Among the many improvements to the Web UI, we've added a Community page to
+surface news and updates from linkerd.io.
+
+For more details, see the announcement blog post:
+https://blog.linkerd.io/2019/04/16/announcing-linkerd-2-3/
```
siggy: URL not yet live

CHANGES.md
siggy: URL not yet live

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: The `linkerd-ca` controller is deprecated in favor of the
+`linkerd-identity` controller. If you had previously installed Linkerd with
+`tls=optional`, you will need to manually delete `linkerd-ca` after upgrading.
```
klingerf: I'd phrase this as "manually delete the `linkerd-ca` deployment after upgrading."

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+Also, `single-namespace` mode is no longer supported. For full details on
```
klingerf: `single-namespace` => `--single-namespace`

CHANGES.md
klingerf: Also, `tls=optional` => `--tls=optional`

CHANGES.md
siggy: good point! i see 2 occurrences of `control-plane` and 56 of `control plane` so i think we know which one wins

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+  * Introduced the `linkerd metrics` command for fetching proxy metrics
+  * **Breaking Change:** The `--linkerd-cni-enabled` flag has been removed from
+    the `inject` command; CNI is configured at the cluster level with the
+    `install` command and no longer applies to the `inject` command
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the
+    `install` command; external profiles are now disabled by default and can be
+    enabled with the new `--enable-external-profiles` flag
+  * **Breaking change** Removed the `--api-port` flag from the `inject` and
+    `install` commands, since there's no benefit to running the control plane's
+    destination API on a non-default port (thanks, @paranoidaditya)
```
klingerf: In previous stable releases, we've removed all of these parenthetical thank yous, in favor of the "Special thanks to" section that you have above.

CHANGES.md
klingerf: On second thought, since you have a section below for "Upgrade notes", I'd recommend removing this paragraph entirely. I'm not sure it's worth calling out the two new CLI commands separately, especially since one of them is unusable for stable installs.

CHANGES.md
```diff
@@ -1,4 +1,189 @@
+## stable-2.3.0
+
+### Announcing Linkerd 2.3 🎈
```
klingerf: For previous stable releases, we've only included this subheading in the release notes on github, and not in CHANGES.md. I'd recommend removing it from here.

CHANGES.md
```diff
@@ -49,17 +234,28 @@ UUID from the `linkerd-config` ConfigMap!
...
 
 * CLI
-  * Introduced an `upgrade` command! This allows an existing Linkerd control plane to be reinstalled or reconfigured; it is particularly useful for automatically reusing flags set in the previous `install` or `upgrade`
-  * The `inject` command proxy options are now converted into config annotations; the annotations ensure that these configs are persisted in subsequent resource updates
+  * Introduced an `upgrade` command! This allows an existing Linkerd control
+    plane to be reinstalled or reconfigured; it is particularly useful for
+    automatically reusing flags set in the previous `install` or `upgrade`
+  * The `inject` command proxy options are now converted into config
+    annotations; the annotations ensure that these configs are persisted in
+    subsequent resource updates
```
klingerf: Thanks for fixing!

CHANGES.md
```diff
@@ -1,4 +1,189 @@
...
+  * **Breaking Change:** The `--linkerd-cni-enabled` flag has been removed from
+    the `inject` command; CNI is configured at the cluster level with the
+    `install` command and no longer applies to the `inject` command
+  * **Breaking Change** Removed the `--disable-external-profiles` flag from the
+    `install` command; external profiles are now disabled by default and can be
+    enabled with the new `--enable-external-profiles` flag
+  * **Breaking change** Removed the `--api-port` flag from the `inject` and
+    `install` commands, since there's no benefit to running the control plane's
+    destination API on a non-default port (thanks, @paranoidaditya)
+  * Removed the `--tls=optional` flag from the `linkerd install` command, since
```
klingerf: I'd also prefix this bullet point with `**Breaking change**`, for consistency with the three bullet points that precede it, which also describe flags being removed.

CHANGES.md
siggy: yeah i had noticed that... i like to overdo with thanking folks from the community.

CHANGES.md
```diff
@@ -1,4 +1,183 @@
...
+  * **Breaking change** Removed the `--api-port` flag from the `inject` and
+    `install` commands, since there's no benefit to running the control plane's
+    destination API on a non-default port (thanks, @paranoidaditya)
+  * **Breaking change** Removed the `--tls=optional` flag from the
+    `linkerd install` command, since TLS is now enabled by default
+  * Changed `install` to accept or generate an issuer Secret for the Identity
+    controller
+  * Changed `install` to fail in the case of a conflict with an existing
+    installation; this can be disabled with the `--ignore-cluster` flag
+  * Added the ability to adjust the Prometheus log level
```
wmorgan: Not a blocker, but would be nice to point to the flag or mechanism used to do this

CHANGES.md
```diff
@@ -1,4 +1,183 @@
...
+  * The `stat` command now always shows the number of open TCP connections
+  * Removed TLS metrics from the `stat` command; this is in preparation for
+    surfacing identity metrics in a clearer way
+  * Exposed the `install-cni` command and its flags, and tweaked their descriptions
+  * Eliminated false-positive vulnerability warnings related to go.uuid
+* Controller
+  * Added a new public API endpoint for fetching control plane configuration
+  * **Breaking change** Removed support for running the control plane in
+    single-namespace mode, which was severely limited in the number of features
+    it supported due to not having access to cluster-wide resources
```
wmorgan: Can we add a note about how we'll address this use in a better way in a later release?

PR #2702, opened 2019-04-16
Split the `linkerd-version` CLI flag into `control-plane-version` and `proxy-version`

### LINKED ISSUE #2614: Deprecate/hide the --linkerd-version flag
During development, it's important to be able to change the proxy version to not match the controller's version.  It is making increasingly less sense, however, that we should expect a given CLI version to install arbitrary control-plane versions.

The CLI should support a `--proxy-version` flag that configures ONLY the proxy image tag (not even the proxy-init image tag). All other image tags should be pinned by the installer.

We may want to maintain the `--linkerd-version` flag for development, but it should be hidden and possibly renamed. This would apply ONLY to install & upgrade.

ihcsim:
This PR replaces the CLI `linkerd-version` flag with a new flag named `proxy-version`. When provided during install, upgrade or inject, this flag sets the proxy image tag to the provided value. The original `linkerd-version` flag is marked deprecated. If provided, it has no effects on the YAML output.

All the control plane components' and proxy init's image tags are now pinned to `version.Version`. My thought is to not provide any CLI flags to override them, because:

1. The override can be done via `kubectl set image`
1. The only way to pass this value to the proxy init container spec is either via annotation, the config proto, or the `ResourceConfig` struct, none of which feel clean, just to support a dev-only flag.

The CNI `linkerd-version` flag isn't changed; it refers to the CNI image tag.

How I tested it:
```
# install the control plane with an older proxy version
bin/linkerd install --proxy-auto-inject --proxy-version=edge-19.4.3 | kubectl apply -f -

# expect proxy version to be persisted as a global property in the linkerd-config config map
kubectl -n linkerd describe configmap linkerd-config

# check the control plane components YAML; expect proxy to be edge-19.4.3
kubectl -n linkerd get deploy -o yaml | less

# try with emojivoto; expect proxy to be edge-19.4.3
curl -sL https://run.linkerd.io/emojivoto.yml | bin/linkerd inject | less

# expect proxy to be edge-19.4.4
curl -sL https://run.linkerd.io/emojivoto.yml | bin/linkerd inject --proxy-version=edge-19.4.4 - | less

# upgrade the control plane, but use proxy version edge-19.4.4
bin/linkerd upgrade --proxy-version=edge-19.4.4 | kubectl apply -f -

# expect proxy version edge-19.4.4 to be persisted as a global property
kubectl -n linkerd describe configmap linkerd-config

# upgrade control plane and proxy to HEAD
bin/linkerd upgrade | kubectl apply -f -

# expect latest dev proxy version to be persisted as a global property
k -n linkerd describe configmap linkerd-config
```

Fixes #2614 


cli/cmd/root.go
```diff
@@ -175,6 +175,7 @@ func getSuccessRate(success, failure uint64) float64 {
 // corresponding flags added in the addProxyConfigFlags func later in this file.
 type proxyConfigOptions struct {
 	linkerdVersion         string
+	proxyVersion           string
```
siggy: can we remove `linkerdVersion` from this struct?

cli/cmd/root.go
```diff
@@ -285,8 +290,10 @@ func (options *proxyConfigOptions) flagSet(e pflag.ErrorHandling) *pflag.FlagSet
 	// Deprecated flags
 	flags.StringVar(&options.proxyMemoryRequest, "proxy-memory", options.proxyMemoryRequest, "Amount of Memory that the proxy sidecar requests")
 	flags.StringVar(&options.proxyCPURequest, "proxy-cpu", options.proxyCPURequest, "Amount of CPU units that the proxy sidecar requests")
+	flags.StringVarP(&options.linkerdVersion, "linkerd-version", "", options.linkerdVersion, "Tag to be used for Linkerd images")
```
siggy: related to previous comment, remove?

test/inject/inject_test.go
```diff
@@ -249,19 +249,22 @@ func patchDeploy(in string, name string, annotations map[string]string) (string,
 	return applyPatch(in, patchJSON)
 }
 
-func removeBuildAnnotations(in string) (string, error) {
-	patchJSON := []byte(`[
-		{"op": "remove", "path": "/spec/template/metadata/annotations/linkerd.io~1created-by"},
-		{"op": "remove", "path": "/spec/template/metadata/annotations/linkerd.io~1proxy-version"}
-	]`)
+func useTestImageTag(in string) (string, error) {
```
siggy: 👍

cli/cmd/root.go
olix0r: I think it's valuable to leave this, especially for development. but i don't think it should be visible by default

cli/cmd/testdata/inject_gettest_deployment.good.golden.yml
```diff
@@ -132,7 +132,7 @@ spec:
         - "2102"
         - --inbound-ports-to-ignore
         - 4190,4191
-        image: gcr.io/linkerd-io/proxy-init:testinjectversion
```
olix0r: this seems like a regression?

cli/cmd/install.go
```diff
@@ -591,7 +592,7 @@ func (options *installOptions) globalConfig(identity *pb.IdentityContext) *pb.Gl
 		LinkerdNamespace:  controlPlaneNamespace,
 		AutoInjectContext: autoInjectContext,
 		CniEnabled:        options.noInitContainer,
-		Version:           options.linkerdVersion,
+		Version:           options.proxyVersion,
```
olix0r: This isn't correct. the global version should be the installed control plane version. It sounds like we may need the ProxyConfig to get a Version field...

cli/cmd/inject.go
```diff
@@ -309,9 +309,9 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
 // the overrideAnnotations map keeps track of which configs are overridden, by
 // storing the corresponding annotations and values.
 func (options *proxyConfigOptions) overrideConfigs(configs *config.All, overrideAnnotations map[string]string) {
-	if options.linkerdVersion != "" {
-		configs.Global.Version = options.linkerdVersion
-		overrideAnnotations[k8s.ProxyVersionOverrideAnnotation] = options.linkerdVersion
+	if options.proxyVersion != "" {
+		configs.Global.Version = options.proxyVersion
```
olix0r: again, I think this is conflating things...

cli/cmd/root.go
siggy: i don't think it's used anymore?
```bash
$ bin/go-run cli install --linkerd-version edge-1.2.3 | grep edge-1.2.3
Flag --linkerd-version has been deprecated, use --proxy-version instead
```

olix0r: Ah, I think that's wrong then. We still need a way to override the linkerd version. If we want to call it `control-plane-version` or something, that's fine with me.

ihcsim: Yeah, I did rename it to `control-plane-version` prior to submitting this PR, then back out of it, because of the changes I had to make to get it into the [proxy-init image tag](https://github.com/linkerd/linkerd2/blob/8323e104fbf6921e3c9b038acd96be3f33b5a860/pkg/inject/inject.go#L643).

Before I go ahead and include this change, can we not just use something like `kubectl edit` or `kubectl set image` to change the control plane version, for development purposes?

ihcsim: Actually, if we're gonna include the control plane version in the global config, then this won't be a problem. I can fix this in my next commit.

ihcsim:
I split the `--linkerd-version` flag into `--proxy-version` and `--control-plane-version`.

`--control-plane-version`:

1. Available only for `install` and `upgrade`
1. Used as the tag for the controller, web, grafana and proxy-init images
1. Maps to the `config.Global.Version` proto field
1. During install, the protobuf field defaults to `version.Version`, unless overridden by the `--control-plane-version` field
1. This field isn't recorded in the `config.Install` config

`--proxy-version`:

1. Available for `install`, `inject` and `upgrade`.
1. Used as the tag for the proxy image
1. Maps to the `config.Proxy.ProxyVersion` proto field
1. During install, this protobuf field defaults to`version.Version`, unless it's overridden by the `--proxy-version` field
1. This field isn't recorded in the `config.Install` config
1. It can be overridden at runtime via the `config.linkerd.io/proxy-version` annotation

LMK if I missed anything.

cli/cmd/upgrade.go
```diff
@@ -141,6 +141,9 @@ func (options *upgradeOptions) validateAndBuild(k kubernetes.Interface, flags *p
 	if options.proxyAutoInject {
 		configs.GetGlobal().AutoInjectContext = &pb.AutoInjectContext{}
 	}
+	if options.controlPlaneVersion != "" {
+		configs.Global.Version = options.controlPlaneVersion
```
alpeb: Mind using `GetGlobal()` here to remain consistent with the other parts of this method?

pkg/inject/inject_test.go
```diff
@@ -17,7 +17,7 @@ import (
 type expectedProxyConfigs struct {
 	image                      string
 	imagePullPolicy            corev1.PullPolicy
-	imageVersion               string
+	proxyVersion               string
```
siggy: mind validating the control-plane / init-image version as well?

```diff
diff --git a/pkg/inject/inject_test.go b/pkg/inject/inject_test.go
index 78eac0d0..b98aa831 100644
--- a/pkg/inject/inject_test.go
+++ b/pkg/inject/inject_test.go
@@ -34,6 +34,7 @@ type expectedProxyConfigs struct {
 	readinessProbe             *corev1.Probe
 	destinationProfileSuffixes string
 	initImage                  string
+	initVersion                string
 	initImagePullPolicy        corev1.PullPolicy
 	initArgs                   []string
 	inboundSkipPorts           string
@@ -69,7 +70,7 @@ func TestConfigAccessors(t *testing.T) {
 
 	globalConfig := &config.Global{
 		LinkerdNamespace: "linkerd",
-		Version:          "default",
+		Version:          "control-plane-version",
 	}
 
 	configs := &config.All{Global: globalConfig, Proxy: proxyConfig}
@@ -150,6 +151,7 @@ func TestConfigAccessors(t *testing.T) {
 				},
 				destinationProfileSuffixes: "svc.cluster.local.",
 				initImage:                  "gcr.io/linkerd-io/proxy-init",
+				initVersion:                "control-plane-version",
 				initImagePullPolicy:        corev1.PullPolicy("Always"),
 				initArgs: []string{
 					"--incoming-proxy-port", "5000",
@@ -214,6 +216,7 @@ func TestConfigAccessors(t *testing.T) {
 				},
 				destinationProfileSuffixes: ".",
 				initImage:                  "gcr.io/linkerd-io/proxy-init",
+				initVersion:                "control-plane-version",
 				initImagePullPolicy:        corev1.PullPolicy("IfNotPresent"),
 				initArgs: []string{
 					"--incoming-proxy-port", "6000",
@@ -374,6 +377,13 @@ func TestConfigAccessors(t *testing.T) {
 				}
 			})
 
+			t.Run("proxyInitVersion", func(t *testing.T) {
+				expected := testCase.expected.initVersion
+				if actual := resourceConfig.proxyInitVersion(); expected != actual {
+					t.Errorf("Expected: %v Actual: %v", expected, actual)
+				}
+			})
+
 			t.Run("proxyInitImagePullPolicy", func(t *testing.T) {
 				expected := testCase.expected.initImagePullPolicy
 				if actual := resourceConfig.proxyInitImagePullPolicy(); expected != actual {
```

cli/cmd/upgrade_test.go
```diff
@@ -15,11 +15,15 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-const upgradeVersion = "TEST-VERSION"
+const (
+	upgradeProxyVersion        = "TEST-VERSION"
+	upgradeControlPlaneVersion = "dev-undefined"
```
siggy: mind making these more specific to the test?

```go
const (
	upgradeProxyVersion        = "UPGRADE-PROXY-VERSION"
	upgradeControlPlaneVersion = "UPGRADE-CONTROL-PLANE-VERSION"
)
```

cli/cmd/testdata/upgrade_default.golden
```diff
@@ -1653,7 +1653,7 @@ spec:
         - 4190,4191
         - --outbound-ports-to-ignore
         - "443"
-        image: gcr.io/linkerd-io/proxy-init:TEST-VERSION
+        image: gcr.io/linkerd-io/proxy-init:dev-undefined
```
siggy: related to other comment, i'd expect these to become `UPGRADE-CONTROL-PLANE-VERSION`. seeing `dev-undefined` makes me assume this field was never set.

cli/cmd/inject_test.go
```diff
@@ -63,13 +63,13 @@ func testInstallConfig() *pb.All {
 
 func TestUninjectAndInject(t *testing.T) {
 	defaultConfig := testInstallConfig()
-	defaultConfig.Global.Version = "testinjectversion"
+	defaultConfig.Proxy.ProxyVersion = "testinjectversion"
```
siggy: mind setting these to something more specific?

```go
defaultConfig.Global.Version = "test-inject-control-plane-version"
defaultConfig.Proxy.ProxyVersion = "test-inject-proxy-version"
```

cli/cmd/upgrade_test.go
```diff
@@ -184,9 +188,9 @@ metadata:
...
+    {"linkerdNamespace":"linkerd","cniEnabled":false,"version":"install-control-plane-version","identityContext":{"trustDomain":"cluster.local","trustAnchorsPem":"-----BEGIN CERTIFICATE-----\nMIIBYDCCAQegAwIBAgIBATAKBggqhkjOPQQDAjAYMRYwFAYDVQQDEw1jbHVzdGVy\nLmxvY2FsMB4XDTE5MDMwMzAxNTk1MloXDTI5MDIyODAyMDM1MlowGDEWMBQGA1UE\nAxMNY2x1c3Rlci5sb2NhbDBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABAChpAt0\nxtgO9qbVtEtDK80N6iCL2Htyf2kIv2m5QkJ1y0TFQi5hTVe3wtspJ8YpZF0pl364\n6TiYeXB8tOOhIACjQjBAMA4GA1UdDwEB/wQEAwIBBjAdBgNVHSUEFjAUBggrBgEF\nBQcDAQYIKwYBBQUHAwIwDwYDVR0TAQH/BAUwAwEB/zAKBggqhkjOPQQDAgNHADBE\nAiBQ/AAwF8kG8VOmRSUTPakSSa/N4mqK2HsZuhQXCmiZHwIgZEzI5DCkpU7w3SIv\nOLO4Zsk1XrGZHGsmyiEyvYF9lpY=\n-----END CERTIFICATE-----\n","issuanceLifetime":"86400s","clockSkewAllowance":"20s"},"autoInjectContext":null}
```
ihcsim: `global.version` and `proxy.proxyVersion` pick up the version test values used in `testInstallOptions()`, instead of just `dev-undefined`.

PR #2716, opened 2019-04-17
Don't use spinner in cli when run without a tty

### LINKED ISSUE #2700: Don't use spinner in linkerd-cli when run without a tty
Our CI machines do not play nice with terminal control characters. 
image

Don't dynamically update the line as linkerd is running its checks when run without a TTY. See how docker solved it here: https://github.com/moby/moby/issues/8342

We have band-aid'ed the problem using sed (`linkerd check | sed -e "s/\x1b\[.\{1,5\}m//g"`) but it is ugly and we now have to look at the exit code of the first command in the pipeline (since sed will always exit with a zero exit code).

There should be no change from the user's perspective, but without a TTY linkerd should not print control characters.

output?

siggy:
In some non-tty environments, the `linkerd check` spinner can render
unexpected control characters.

Disable the spinner when run without a tty.

Fixes #2700

check-tty

### DIFF
--- a/Dockerfile-proxy
+++ b/Dockerfile-proxy
@@ -11,7 +11,7 @@ RUN (proxy=$(bin/fetch-proxy $PROXY_VERSION) && \
     echo "$version" >version.txt)
 
 ## compile proxy-identity agent
-FROM gcr.io/linkerd-io/go-deps:44063d94 as golang
+FROM gcr.io/linkerd-io/go-deps:4120b3aa as golang
 WORKDIR /go/src/github.com/linkerd/linkerd2
 ENV CGO_ENABLED=0 GOOS=linux
 COPY pkg/flags pkg/flags
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -450,12 +450,12 @@
   version = "v0.0.9"
 
 [[projects]]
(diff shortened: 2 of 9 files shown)


siggy:
Note: we may be able to back this change out when https://github.com/briandowns/spinner/issues/53 is resolved.

siggy:
@dwj300 Let us know if this branch solves your use case.

siggy:
@alpeb whoah cool thanks for testing with ACME. if it's not much trouble, mind just sharing a couple notes/commands on how you did that?

alpeb:
@siggy Sure :slightly_smiling_face:  I did this in Ubuntu:
```bash
# Install "Plan 9 From User Space", using the root user
apt-get install xorg-dev
cd /user/local
git clone https://github.com/9fans/plan9port plan9

# change ownership and get back to your normal user
chown -R youruser:youruser plan9
cd plan9
./INSTALL
export PLAN9=/usr/local/plan9
export PATH=$PATH:$PLAN9/bin

# prepare the script to run under Acme
echo "linkerd check --pre > run"
chmod +x run

# install the linkerd CP and quickly afterwards run our "run" command
linkerd install | k apply -f -
acme

# in Acme's UI, middle-click with the mouse on the "run" script
# and you should see the first screenshot I pasted above

Issue #2540, opened 2019-03-21
Add support for `config.linkerd.io/disable-identity` annotation in auto-inject


klingerf:
#2535 added a `--disable-identity` flag for CLI inject, but there's no equivalent for disabling identity in the auto-injector. We should add support for a `config.linkerd.io/disable-identity` annotation that allows disabling identity per workload. We should also consider a `--disable-identity` flag for `linkerd install`, which globally disables identity for all workloads. cc @olix0r

alpeb:
The part for `linkerd upgrade`  has been merged. Still have to do the same in `linkerd install`. The merged branch contains some notes to take into account for `linkerd install`.

alpeb:
Turns out we need identity for the upcoming tap work. #2783 tracks the segregation of mTLS in the data plane vs establishing identity against the control plane.

PR #2717, opened 2019-04-17
Add config.linkerd.io/disable-identity annotation

alpeb:
First part of #2540

We'll tackle support for `--disable-identity` in `linkerd install` in a separate PR.

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -30,11 +30,6 @@ const (
 	udpDesc            = "pod specs do not include UDP ports"
 )
 
-type injectOptions struct {
-	disableIdentity bool
-	*proxyConfigOptions
-}
-
 type resourceTransformerInject struct {
 	configs               *config.All
 	overrideAnnotations   map[string]string
@@ -45,15 +40,8 @@ func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, transforme
 	return transformInput(inputs, errWriter, outWriter, transformer)
 }
 
-func newInjectOptions() *injectOptions {
-	return &injectOptions{
-		// No proxy config overrides.
-		proxyConfigOptions: &proxyConfigOptions{},
-	}
-}
-
 func newCmdInject() *cobra.Command {
-	options := newInjectOptions()
+	options := &proxyConfigOptions{}
 
 	cmd := &cobra.Command{
 		Use:   "inject [flags] CONFIG-FILE",
@@ -102,7 +90,7 @@ sub-folders, or coming from stdin.`,
 		},
 	}
 
-	flags := options.proxyConfigOptions.flagSet(pflag.ExitOnError)
+	flags := options.flagSet(pflag.ExitOnError)
 	flags.BoolVar(
 		&options.disableIdentity, "disable-identity", options.disableIdentity,
 		"Disables resources from participating in TLS identity",
@@ -283,7 +271,7 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output 
 	output.Write([]byte("\n"))
 }
 
-func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
+func (options *proxyConfigOptions) fetchConfigsOrDefault() (*config.All, error) {
 	if options.ignoreCluster {
 		if !options.disableIdentity {
 			return nil, errors.New("--disable-identity must be set with --ignore-cluster")
@@ -299,9 +287,6 @@ func (options *injectOptions) fetchConfigsOrDefault() (*config.All, error) {
 		return nil, err
 	}
 
-	if options.disableIdentity {
-		config.Global.IdentityContext = nil
-	}
 	return config, nil
 }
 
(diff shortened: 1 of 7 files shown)


cli/cmd/inject.go
```diff
@@ -30,11 +30,6 @@ const (
 	udpDesc            = "pod specs do not include UDP ports"
 )
 
-type injectOptions struct {
-	disableIdentity bool
-	*proxyConfigOptions
-}
-
```
alpeb: `--disable-identity` will be an option shared by both `linkerd inject` and `linkerd install`, so I moved it into `proxyConfigOptions` and we no longer need a separate `injectOptions` struct.

olix0r: Mmm... I'm not sure that this would be safe to apply on upgrade. Almost certainly would produce a broken cluster.

alpeb: `linkerd upgrade --disable-identity` is not allowed. Or you meant something else?

olix0r: Oh, I'm confusing things. As long as this doesn't cause disable-identity to be included in the upgrade help output, I'm happy.

test/inject/testdata/injected_default.golden
```diff
@@ -11,6 +11,7 @@ spec:
   template:
     metadata:
       annotations:
+        config.linkerd.io/disable-identity: "true"
```
siggy: for my own understanding, how does this interact with `linkerd.io/identity-mode: disabled` ?

alpeb: if we have `config.linkerd.io/disable-identity: "true"` (set manually by the user or by the system through `linkerd inject --disable-identity`) then the system will also set `linkerd.io/identity-mode: disabled`. 
Which is a little redundant but the reasoning is similar to what we had in #2593 for having both annotations for proxy-version, in that one is an explicit override present in a spec that might not have been injected yet, and the other is the actual config value used when the injection is performed.

siggy: Ok thanks for the explanation. Naively this does feel a bit redundant. Something to keep in mind as we go along.

cli/cmd/inject.go
ihcsim: To confirm the implication behind adding `--disable-identity` to `install`, in this set up,

1. The identity service is still installed in the control plane
1. By default, all meshed services will not participate in identity
1. The only way for any meshed services to participate in identity is via the new `config.linkerd.io/disable-identity=false` annotation
1. It won't be possible to enable global identity again during upgrade, because it's not a supported flag at the moment.

Am I right?

olix0r: One thing to call out here -- if control plane proxies are not instrumented with identity, then no other pods will be able to use identity.  So `--disable-identity` at install-time should probably set a flag on the proxy config to configure the default injection policy, but we probably should avoid applying this policy to the control plane itself.

alpeb: @ihcsim @olix0r Those are great points, thanks. I'll keep them in mind when doing the install part.
Adding a new flag for the default injection policy will facilitate re-enabling it in `linkerd update`, if we want to implement that at some point.

olix0r: @alpeb and an alternative, also viable approach, would be to let `install --disable-identity` disable CP proxy identity; and if `disable-identity: false` is specified, injection would fail.

cc @grampelberg

ihcsim:
@alpeb Sorry, I missed the unit test. Would you mind updating the `pkg/inject/inject_test.go` to test the new accessor? Thanks!

```diff
diff --git a/pkg/inject/inject_test.go b/pkg/inject/inject_test.go
index 4dac5943..95404aec 100644
--- a/pkg/inject/inject_test.go
+++ b/pkg/inject/inject_test.go
@@ -15,6 +15,7 @@ import (
 )
 
 type expectedProxyConfigs struct {
+	identityContext            *config.IdentityContext
 	image                      string
 	imagePullPolicy            corev1.PullPolicy
 	imageVersion               string
@@ -69,6 +70,7 @@ func TestConfigAccessors(t *testing.T) {
 	globalConfig := &config.Global{
 		LinkerdNamespace: "linkerd",
 		Version:          "default",
+		IdentityContext:  &config.IdentityContext{},
 	}
 
 	configs := &config.All{Global: globalConfig, Proxy: proxyConfig}
@@ -83,6 +85,7 @@ func TestConfigAccessors(t *testing.T) {
 				Template: corev1.PodTemplateSpec{
 					metav1.ObjectMeta{
 						Annotations: map[string]string{
+							k8s.ProxyDisableIdentityAnnotation:        "true",
 							k8s.ProxyImageAnnotation:                  "gcr.io/linkerd-io/proxy",
 							k8s.ProxyImagePullPolicyAnnotation:        "Always",
 							k8s.ProxyInitImageAnnotation:              "gcr.io/linkerd-io/proxy-init",
@@ -169,6 +172,7 @@ func TestConfigAccessors(t *testing.T) {
 				},
 			},
 			expected: expectedProxyConfigs{
+				identityContext: &config.IdentityContext{},
 				image:           "gcr.io/linkerd-io/proxy",
 				imagePullPolicy: corev1.PullPolicy("IfNotPresent"),
 				imageVersion:    "default",
@@ -240,6 +244,13 @@ func TestConfigAccessors(t *testing.T) {
 				t.Fatal(err)
 			}
 
+			t.Run("identityContext", func(t *testing.T) {
+				expected := testCase.expected.identityContext
+				if actual := resourceConfig.identityContext(); !reflect.DeepEqual(expected, actual) {
+					t.Errorf("Expected: %+v Actual: %+v", expected, actual)
+				}
+			})
+
 			t.Run("proxyImage", func(t *testing.T) {
 				expected := testCase.expected.image
 				if actual := resourceConfig.proxyImage(); expected != actual {
```

PR #2719, opened 2019-04-18
Introduce install stages

siggy:
This change introduces two named parameters for `linkerd install`, split
by privilege:
- `linkerd install config`
  - Namespace
  - ClusterRoles
  - ClusterRoleBindings
  - CustomResourceDefinition
  - ServiceAccounts
- `linkerd install control-plane`
  - ConfigMaps
  - Secrets
  - Deployments
  - Services

Comprehensive `linkerd install` is still supported.

TODO:
- `linkerd check` support
- `linkerd upgrade` support
- integration tests

Part of #2337


siggy:
Per https://github.com/linkerd/linkerd2/issues/2337#issuecomment-484667758 I'm considering scoping stage 1 strictly to cluster-level resources:

- `linkerd install config`
  - Namespace
  - ClusterRoles
  - ClusterRoleBindings
  - CustomResourceDefinition
- `linkerd install control-plane`
  - ConfigMaps
  - ServiceAccounts
  - Secrets
  - Deployments
  - Services

olix0r:
My expectation is the linkerd control plane admin will have to care about the issuer credentials; not the cluster admin.

cli/cmd/install.go
```diff
@@ -129,32 +131,16 @@ type (
...
-	prometheusConfigTemplateName    = "templates/prometheus-config.yaml"
-	prometheusTemplateName          = "templates/prometheus.yaml"
-	grafanaConfigTemplateName       = "templates/grafana-config.yaml"
-	grafanaTemplateName             = "templates/grafana.yaml"
-	resourcesTemplateName           = "templates/_resources.yaml"
-	serviceprofileTemplateName      = "templates/serviceprofile.yaml"
-	proxyInjectorConfigTemplateName = "templates/proxy_injector-config.yaml"
-	proxyInjectorTemplateName       = "templates/proxy_injector.yaml"
-	spValidatorConfigTemplateName   = "templates/sp_validator-config.yaml"
-	spValidatorTemplateName         = "templates/sp_validator.yaml"
```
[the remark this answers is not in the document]
siggy: Since these strings only occur in one place, I prefer to just hard-code them. This way I can look at one place in the file rather than two to get my head around what's happening.

siggy:
@alpeb 

> Per your last comment, if you limit the first stage to just cluster-level resources, you'd increase the chance of getting warning events during the second stage.

I think this is ok for now (at least not worse than current state). Per https://github.com/linkerd/linkerd2/issues/2337#issuecomment-484667758, Kubernetes is designed to handle not everything being available during startup, and I'm increasingly feeling that we should not try to control/solve deployment ordering.

> Also users might start entering config flags in the first stage, so we'll have to persist them in the main ConfigMap in that stage. This also means we'll have to update that config with flags passed during the second stage. We can't fail in the second stage if there's already a ConfigMap, so most likely we'll have to introduce a new field to distinguish between finished installs and configed-only ones.

I've rebased the whole PR to make the first stage cluster-level resources only, and also added some code to disable most flags for the first stage.

cli/cmd/install.go
```diff
@@ -248,7 +251,34 @@ func newCmdInstall() *cobra.Command {
...
+	}
+
+	// only a few flags are available for config stage
+	if stage == configStage && flags != nil {
+		var err error
+		flags.VisitAll(func(f *pflag.Flag) {
+			if f.Changed {
+				switch f.Name {
+				// TODO: remove "proxy-auto-inject" when it becomes default
+				case "linkerd-namespace", "proxy-auto-inject":
```
alpeb: I think we should add to this list `ignore-cluster` and all the global config flags.

alpeb: I meant all the "global flags" (--api-addr, etc)

alpeb:
If not through some sort of state field in the global ConfigMap, what's the plan for ensuring `linkerd install control-plane` gets run after `linkerd instal config`? I assume listing ClusterRoles and ClusterRoleBindings is not available under simple namespace-level privileges, so we could only maybe check the existence of the linkerd namespace, which wouldn't be a very limited check...

cli/cmd/install.go
siggy: The `flags` var passed into this function comes from `installOptions.recordableFlagSet()`, which does not include the "global flags".

```bash
$ bin/go-run cli install config --api-addr foo > /dev/null && echo $?
0
$ bin/go-run cli install config --admin-port 123 > /dev/null && echo $?
Error: flag not available for config stage: --admin-port
Usage:
```

... that said, i'm realizing that there is also `installOnlyFlagSet`, none of which should be set during `linkerd install config`, i have pushed a new commit to reflect this. i don't think `linkerd install config --ignore-cluster` should be supported, as it has no effect on the output.

siggy:
> If not through some sort of state field in the global ConfigMap, what's the plan for ensuring linkerd install control-plane gets run after linkerd instal config? I assume listing ClusterRoles and ClusterRoleBindings is not available under simple namespace-level privileges, so we could only maybe check the existence of the linkerd namespace, which wouldn't be a very limited check...

@alpeb I think we can manage a lot of this via docs. I'd like to ensure users can run these commands in any order, just to see the output. While the behavior of `linkerd install` and `linkerd upgrade` are now reliant on k8s state, I'd like to minimize that surface area.

cli/cmd/install.go
alpeb: > `linkerd install config --ignore-cluster` should be supported, as it has no effect on the output.

`linkerd install ` and `linkerd install config` both check if linkerd already exists in the cluster, so I was thinking on allowing that flag in both cases for consistency.

siggy: oops sorry i mean `linkerd install config --ignore-cluster` should _not_ be supported (i think you read it the way i intended anyway ;) ) ...

nice catch! i failed to realize that `linkerd install config` would error out if a cluster already existed, and not providing an `--ignore-cluster` flag leaves the user completely unable to render that output.

dadjeibaah:
I kinda think we may need to evaluate the user experience more when a user tries to install the control plane without running `linkerd install config`. When I run the command, I get a ton of 
```bash
Error from server (NotFound): error when creating "STDIN": namespaces "linkerd" not found
```
Which may be a little shocking when running `install control-plane`. WDYT?

siggy:
> Which may be a little shocking when running install control-plane. WDYT?

@dadjeibaah Your comments and @alpeb's are making me think we need to guide the user a bit more via the cli. Maybe we can have `linkerd install control-plane` return an error if the namespace (and maybe RBAC?) does not exist, and provide a flag to just get the YAML if they really want it:

```bash
$ linkerd install control-plane
Error: 'linkerd' namespace does not exist, run 'linkerd install config' first, or 'linkerd install control-plane ignore-cluster' to get the output anyway.

$ linkerd install control-plane --ignore-cluster
---
kind: Service
...
```
@grampelberg thoughts?

alpeb:
It's not clear which flags are available to each stage. You think it would be too much to have these be totally different commands, like `install-config` and `install-control-plane`? Keeping of course the full `install` command, still with a note explaining you can separate the install using those other two commands.

siggy:
@alpeb good points re: which flags work for which commands. Creating different commands would make that clearer. I think another approach could be to set these up as proper Cobra subcommands of `install`.

Either way, I'd like to get this baseline functionality merged as-is, to minimize merge conflicts with ongoing install/inject work. I've filed #2729 to track this.

siggy:
@alpeb @dadjeibaah @grampelberg latest commit causes `linkerd install control-plane` to fail is namespace does not exist:

```bash
$ bin/go-run cli install control-plane
Failed to find the Linkerd control-plane namespace. If this expected, use the --ignore-cluster flag.
Error: The "linkerd" namespace does not exist

$ bin/go-run cli install control-plane --ignore-cluster
---
kind: ConfigMap
...
```

cli/cmd/install.go
```diff
@@ -717,6 +744,31 @@ func exitIfClusterExists() {
...
+// exitIfNamespaceDoesNotExist checks the kubernetes API to determine if the
+// control-plane namespace exists, and returns an error if it does not.
+//
+// This is useful when running `linkerd install control-plane`, where the
+// namespace must exist, but `linkerd-config` should not.
+func exitIfNamespaceDoesNotExist() {
+	hc := newHealthChecker(
+		[]healthcheck.CategoryID{healthcheck.KubernetesAPIChecks},
+		time.Time{},
+	)
```
alpeb: TIOLI: we might gain some consistency by reusing this same minimalistic healthchecker in `exitIfClusterExists`. That might need some new `hc.CheckConfigMap` method to replace the direct call we're making in `existIfClusterExists`.

siggy: Agree, I looked at refactoring this a bit further, then backed it out in the interest of getting this PR through. I captured your comment, and related requirements for healthcheck and client creation via #2735.

adleong:
The check that the Linkerd namespace exists before running `linkerd install control-plane` feels similar to other pre-command checks (e.g. checking that the control plane is reachable before running stat) so I'd call the flag something like `--skip-checks`.  If I understand correctly, `--ignore-cluster` is specifically for overriding the linkerd config when running install over an existing install.

cli/cmd/install.go
```diff
@@ -717,6 +744,31 @@ func exitIfClusterExists() {
...
+
+	success := hc.RunChecks(exitOnError)
+	if !success {
+		fmt.Fprintln(os.Stderr, "Failed to connect to Kubernetes. If this expected, use the --ignore-cluster flag.")
+		os.Exit(1)
+	}
+
+	err := hc.CheckNamespace(context.Background(), controlPlaneNamespace, true)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Failed to find the Linkerd control-plane namespace. If this expected, use the --ignore-cluster flag.")
```
adleong: Probably our first suggestion should be to run `linkerd install config`.  Maybe something like: "Failed to find control-plane namespace: %s.  Run `linkerd install config -l %s | kubectl apply -f -` to create it (this requires cluster administration permissions.)  See <website docs link> for more information.  Or use <flag> to proceed anyway."

cli/cmd/install.go
```diff
@@ -717,6 +746,34 @@ func exitIfClusterExists() {
...
+// namespace must exist, but `linkerd-config` should not.
+func exitIfNamespaceDoesNotExist() {
+	hc := newHealthChecker(
+		[]healthcheck.CategoryID{healthcheck.KubernetesAPIChecks},
+		time.Time{},
+	)
+
+	success := hc.RunChecks(exitOnError)
+	if !success {
+		fmt.Fprintln(os.Stderr, "Failed to connect to Kubernetes. If this expected, use the --ignore-cluster flag.")
```
adleong: I think this message needs to be updated as well

cli/cmd/public_api.go
```diff
@@ -1,5 +1,8 @@
 package cmd
 
+// TODO: this file should probably move out of `./cli/cmd` into something like
+// `./cli/publicapi` or `./cli/pkg`
```
adleong: maybe create an issue and link to it in this TODO?

siggy: filed #2735 yesterday to track, have updated the comment for reference.

PR #2698, opened 2019-04-15
Introduce basic UI testing for Linkerd dashboard: Sidebar (#2475)

drholmie:
Deals with UI testing for sidebar in Linkerd dashboard.
Made simple UI test to make sure UI elements of the sidebar are correct and in order.
Increased timeouts in wdio*.conf.js files to allow for longer running tests (>1 minute) and handle slow connections.
Added sauce-service to wdio-sauce.conf.js .
Tested locally and on cloud.

Issue #2475, opened 2019-03-08
Automated integration testing for the dashboard


admc:
It would be very helpful if we had an integration test suite that was easy to run for the linkerd2 dashboard. It appears that webdriver.io is the most polished project for building that suite, and we have a free open source account with lots of concurrency to run it on our build using Sauce Labs.

I would start with each of the pages in the sidebar, and build a test for each component on that page to validate that normal happy path interactions are working as expected.

This would require some setup steps, like injecting emojivoto so that there is real traffic we can validate.

grampelberg:
A mostly priority sorted list of pages:

- [ ] Sidebar
- [ ] Overview
- [ ] Resources
    - [ ] Deployments
    - [ ] Pods
    - [ ] Namespaces
    - [ ] Daemon Sets
    - [ ] Stateful Sets
- [ ] Tap
- [ ] Top
- [ ] Top Routes
- [ ] Service Mesh

grampelberg:
@rmars where do you think this should live in the repo?

admc:
A hint for running on Sauce is that to hit any web property running on a local port from the cloud service needs to use the Sauce Connect tunnel: https://wiki.saucelabs.com/display/DOCS/Sauce+Connect+Proxy

That probably isn't a concern until we want to think about running them across multiple platforms and or in CI.

admc:
In CI we should use the webdriver.io sauce service, locally - whatever is faster for building good tests (webdriver, chromedriver, etc)

rmars:
I think it could live in `web/app/integration` since we have the unit tests alongside components in `web/app/js` and we have test fixtures in `web/app/test` (or even `web/integration` could work)

drholmie:
Hey there! I have been working on this issue a bit. (https://github.com/drholmie/webdrive-tests). Thanks for clarifying where this should live on the repo! So far I think I have covered the tests for sidebar(they check if all buttons are working, and if links are broken. Do tell if I've missed anything). Will begin work on individual pages now.

PR #2720, opened 2019-04-19
Introduce basic UI testing for Linkerd dashboard: Daemonset (#2475) #2698

drholmie:
Made simple UI test to make sure table headers of daemonset page are correct and in order
Increased timeouts in wdio*.conf.js files to allow for longer running tests (>1 minute) and handle slow connections
Added sauce-service to wdio-sauce.conf.js for webdriverio commands to run correctly on sauce labs
Tested locally and on cloud

### DIFF
--- a/web/app/integration/specs/daemonset/daemonset-table.js
+++ b/web/app/integration/specs/daemonset/daemonset-table.js
@@ -1,0 +1,16 @@
+const assert = require('assert');
+    describe('table check', function() {
+    it('should check the page for proper table structure', () => {
+        browser.url('http://127.0.0.1:7777/daemonsets');
+        browser.waitUntil(() => {return $('table').isExisting();
+          }, 10000, 'expected table to exist');
+        http_headers=["Namespace", "Daemon Set","Meshed", "Success Rate", "RPS", "P50 Latency", 
+        "P95 Latency", "P99 Latency", "Grafana"];
+        tcp_headers=["Namespace","Daemon Set","Meshed","Connections", "Read Bytes / sec","Write Bytes / sec", 
+"Grafana"];
+        const http=$$("table")[0].$("thead").$$("th").map(item=>item.getText());
+        assert(http.join('')==http_headers.join(''));
+        const tcp=$$("table")[1].$("thead").$$("th").map(item=>item.getText());
+        assert(tcp.join('')==tcp_headers.join(''));
+    });
+});
--- a/web/app/integration/specs/daemonset/daemonset-table.js
+++ b/web/app/integration/specs/daemonset/daemonset-table.js
@@ -1,16 +1,19 @@
 const assert = require('assert');
-    describe('table check', function() {
-    it('should check the page for proper table structure', () => {
-        browser.url('http://127.0.0.1:7777/daemonsets');
-        browser.waitUntil(() => {return $('table').isExisting();
-          }, 10000, 'expected table to exist');
-        http_headers=["Namespace", "Daemon Set","Meshed", "Success Rate", "RPS", "P50 Latency", 
-        "P95 Latency", "P99 Latency", "Grafana"];
-        tcp_headers=["Namespace","Daemon Set","Meshed","Connections", "Read Bytes / sec","Write Bytes / sec", 
-"Grafana"];
-        const http=$$("table")[0].$("thead").$$("th").map(item=>item.getText());
-        assert(http.join('')==http_headers.join(''));
-        const tcp=$$("table")[1].$("thead").$$("th").map(item=>item.getText());
-        assert(tcp.join('')==tcp_headers.join(''));
-    });
+
+describe('table check', function() {
+  it('should check the page for proper table structure', () => {
+    browser.url('http://127.0.0.1:7777/daemonsets');
+    browser.waitUntil(() => {
+      return $('table').isExisting();
+    }, 10000, 'expected table to exist');
+    httpHeaders=["Namespace", "Daemon Set","Meshed", "Success Rate", "RPS", "P50 Latency","P95 Latency", "P99 Latency", "Grafana"];
+    tcpHeaders=["Namespace","Daemon Set","Meshed","Connections", "Read Bytes / sec","Write Bytes / sec","Grafana"];
+    const pageTables=$$("table");
+    const httpTable=pageTables[0].$("thead");
+    const tcpTable=pageTables[1].$("thead");
+    const http=httpTable.$$("th").map(item=>item.getText());
+    assert(http.join('')===httpHeaders.join(''));
+    const tcp=tcpTable.$$("th").map(item=>item.getText());
+    assert(tcp.join('')===tcpHeaders.join(''));
+  });
 });
--- a/web/app/integration/wdio-sauce.conf.js
+++ b/web/app/integration/wdio-sauce.conf.js
@@ -5,19 +5,16 @@ exports.config = {
   service: ['sauce'],
   sauceConnect: true,
   specs: [
-      './integration/specs/**/*.js'
+    './integration/specs/**/*.js'
   ],
-    suites: {
-    sidebar: [
-        './integration/specs/sidebar/*.js'
-    ],
+  suites: {
     daemonset: [
-        './integration/specs/daemonset/*.js'
+      './integration/specs/daemonset/*.js'
     ]
-},
+  },
   // Patterns to exclude.
   exclude: [
-      // 'path/to/excluded/files'
+    // 'path/to/excluded/files'
   ],
   maxInstances: 10,
   capabilities: [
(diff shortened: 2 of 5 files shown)


web/app/integration/specs/daemonset/daemonset-table.js
```diff
@@ -0,0 +1,16 @@
+const assert = require('assert');
+    describe('table check', function() {
+    it('should check the page for proper table structure', () => {
+        browser.url('http://127.0.0.1:7777/daemonsets');
+        browser.waitUntil(() => {return $('table').isExisting();
+          }, 10000, 'expected table to exist');
+        http_headers=["Namespace", "Daemon Set","Meshed", "Success Rate", "RPS", "P50 Latency",
```
dadjeibaah: We use camel case for variables, so instead of `http_headers` we use `httpHeaders`

web/app/integration/specs/daemonset/daemonset-table.js
```diff
@@ -0,0 +1,16 @@
...
+    describe('table check', function() {
+    it('should check the page for proper table structure', () => {
+        browser.url('http://127.0.0.1:7777/daemonsets');
+        browser.waitUntil(() => {return $('table').isExisting();
+          }, 10000, 'expected table to exist');
+        http_headers=["Namespace", "Daemon Set","Meshed", "Success Rate", "RPS", "P50 Latency", 
+        "P95 Latency", "P99 Latency", "Grafana"];
+        tcp_headers=["Namespace","Daemon Set","Meshed","Connections", "Read Bytes / sec","Write Bytes / sec", 
+"Grafana"];
+        const http=$$("table")[0].$("thead").$$("th").map(item=>item.getText());
```
dadjeibaah: I think this `$("table")` could be DRY'd up a little bit. Might I suggest placing `$("table")` in a variable like 
```js
const pageTables = $("table")
```

web/app/integration/specs/daemonset/daemonset-table.js
```diff
@@ -0,0 +1,16 @@
...
+        assert(http.join('')==http_headers.join(''));
+        const tcp=$$("table")[1].$("thead").$$("th").map(item=>item.getText());
+        assert(tcp.join('')==tcp_headers.join(''));
```
dadjeibaah: in JS the preferred comparison operator is `===` as opposed to `==`. Check out [this](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons) stackoverflow post on the reason's why.

web/app/integration/wdio-sauce.conf.js
```diff
@@ -2,10 +2,19 @@ exports.config = {
...
   key: process.env.SAUCE_ACCESS_KEY,
+  service: ['sauce'],
   sauceConnect: true,
   specs: [
-      './integration/specs/*.js'
+      './integration/specs/**/*.js'
   ],
+    suites: {
+    sidebar: [
+        './integration/specs/sidebar/*.js'
```
dadjeibaah: I noticed that this commit includes 
```js
    sidebar: [
        './integration/specs/sidebar/*.js'
    ],
```

I don't see that directory in the integration directory. Was this added by mistake?

web/app/integration/wdio-sauce.conf.js
```diff
@@ -2,10 +2,19 @@ exports.config = {
...
+    ],
+    daemonset: [
+        './integration/specs/daemonset/*.js'
```
dadjeibaah: According to our [.eslintrc](https://github.com/linkerd/linkerd2/blob/master/web/app/.eslintrc), This line, and a couple of other lines in this file, need to be indented with two spaces instead of four
```diff
diff --git a/web/app/integration/wdio-sauce.conf.js b/web/app/integration/wdio-sauce.conf.js
index ca23c4c6..211ea9cd 100644
--- a/web/app/integration/wdio-sauce.conf.js
+++ b/web/app/integration/wdio-sauce.conf.js
@@ -5,24 +5,24 @@ exports.config = {
   service: ['sauce'],
   sauceConnect: true,
   specs: [
-      './integration/specs/**/*.js'
+    './integration/specs/**/*.js'
   ],
-    suites: {
+  suites: {
     sidebar: [
-        './integration/specs/sidebar/*.js'
+      './integration/specs/sidebar/*.js'
     ],
     daemonset: [
-        './integration/specs/daemonset/*.js'
+      './integration/specs/daemonset/*.js'
     ]
-},
+  },
   // Patterns to exclude.
   exclude: [
-      // 'path/to/excluded/files'
+    // 'path/to/excluded/files'
   ],
   maxInstances: 10,
   capabilities: [
-    {browserName: 'firefox', platform: 'Windows 10', version: '60.0'},
-    {browserName: 'chrome', platform: 'OS X 10.13', version: '69.0'}
+    { browserName: 'firefox', platform: 'Windows 10', version: '60.0' },
+    { browserName: 'chrome', platform: 'OS X 10.13', version: '69.0' }
   ],
   bail: 0,
   baseUrl: 'http://localhost',
@@ -31,7 +31,7 @@ exports.config = {
   connectionRetryCount: 3,
   framework: 'mocha',
   mochaOpts: {
-      ui: 'bdd',
-      timeout: 6000000
+    ui: 'bdd',
+    timeout: 6000000
   }
 }

```

web/app/integration/wdio-sauce.conf.js
```diff
@@ -17,12 +26,12 @@ exports.config = {
   ],
   bail: 0,
   baseUrl: 'http://localhost',
-  waitforTimeout: 10000,
-  connectionRetryTimeout: 90000,
+  waitforTimeout: 1000000,
+  connectionRetryTimeout: 90000000,
```
dadjeibaah: These timeouts seem a little high. `waitforTimeout` looks like it waits for 16 minutes, while `connectionRetryTimeout` waits for 25 hours. I'm wondering if you were running into timing issues with the test? 

I think we could set these to something much lower, or maybe use the default values supplied by webdriverIO. We would like to make sure we aren't spending too much time running tests since the goal is to have tests run as quickly as possible to get feedback quickly when we encounter a bug in the UI.

web/app/package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "@material-ui/core": "3.6.1",
     "@material-ui/icons": "3.0.1",
+    "@wdio/sauce-service": "^5.7.15",
```
dadjeibaah: This should be moved into `devDependencies` rather than here since we don't need to run this module production. Also, it looks like we only use this module for testing so `devDependencies` is a more appropriate place to have this.

web/app/integration/specs/daemonset/daemonset-table.js
drholmie: Oh, I see. Will make the changes to my other test files as well. Thank you for the information.

web/app/integration/specs/daemonset/daemonset-table.js
drholmie: Oh my! JS sure has it's quirks. Will change tests to the appropriate equality symbol. Thank you for pointing this out!

web/app/integration/wdio-sauce.conf.js
drholmie: Ah, this PR was actually meant to follow an older one(https://github.com/linkerd/linkerd2/pull/2698). I had made suites to allow for testing of various pages individually. I realized my error in submitting multiple PRs. I apologize for that. Will change this too.

web/app/integration/wdio-sauce.conf.js
drholmie: Oh this is strange. I don't remember changing anything under capabilities, but yes I'll make the appropriate indents.

web/app/integration/wdio-sauce.conf.js
drholmie: Ah, my apologies, forgot to remove some of the zeros. The default values tend to be too low for successful running of tests, especially when using browser.waitUntil(), browser.pause() on sauce labs (and on slower connections) so I added a few(too many) extra zeros to fix it. I'll change those values to make them more suitable. So sorry about that. Thank you for pointing it out!

web/app/package.json
drholmie: Alright! will change that.
Unfortunately currently going through a stressful week in college [finals and project submissions begin from next week :) ]. So I'll try to do the required changes(including the ones in the aforementioned comments) whenever I can get the time. So sorry about that. 
Thank you for reviewing my PR!

dadjeibaah:
Looks like this was closed by accident

drholmie:
Hey there! @dadjeibaah finished making the changes. Hope everything is in order. Thanks for reopening it!

scottcarol:
Thanks @drholmie! Most of our engineering team is prepping for KubeCon next week so our response time is a little delayed. I'll definitely take a look at this early next week!

scottcarol:
Closing after conversation with @drholmie, who will submit a new PR after this

Issue #2711, opened 2019-04-17
Introduce a new CLI inject `--advanced` flag


ihcsim:
Following #2710 , the `inject` command's default behaviour will be changed to only output the annotated YAML. To preserve the ability to output fully injected YAML, introduce a new `--advanced` CLI flag. 

This flag will be used as a 'break the glass' measure, to support users who want to manually inspect or modify the YAML. Auto injection and config override aren't supported for these injected YAMLs.

Part of #2576

alpeb:
Closed via #2721

Issue #2715, opened 2019-04-17
Make proxy injector a required component of the control plane


ihcsim:
Update the installation process to make the proxy injector a required component.

* [x] Remove the CLI `--proxy-auto-inject` flag
* [x] Deprecate the `auto_inject_context` protobuf field
* [x] Test `upgrade`

Part of #2576 

Relevant #2164 #2337

Issue #2576, opened 2019-03-27
Consolidate Differences In Runtime YAML of Auto-Inject And Manual-Injected Workloads


ihcsim:
With #2472, the proxy injector webhook now mutates the pod spec directly. This results in discrepancies between the YAML output generated by the `linkerd inject` command and that by the webhook.

At runtime, the output of `kubectl describe [deploy|sts|job|rc|dameonset]` of an auto-injected workload will not have the proxy specs, because the webhook modifies the pod specs, instead of its owner's pod template. The specs of the live pod becomes the only source of truth, which can be confusing for support purposes.

For 2.4, one way to consolidate that this is, as @grampelberg and @alpeb suggested, to devolve the `linkerd inject` to just adding linkerd annotations.

cc @grampelberg

grampelberg:
As a project, we've relied on showing users exactly what is happening when they do something. By moving to auto-injecting pods instead of deployments, we've moved away from that a little bit. It feels like there needs to be a preview command that shows the actual YAML for the pods as a preview (`linkerd inject --preview`?) as well as unifying the two different places that containers can sit.

alpeb:
`linkerd inject --preview` sounds like a great idea to me!

A few things to take into account with this approach though:
- The auto injector will always have to be installed (bye bye `--proxy-auto-inject` :+1:  ). This means we can't support k8s versions lower than 1.9 (I think), which should be ok, no? Also maybe we can create a new `--pre` check to ensure the mutatingwebhook plugin is enabled.
- CLI `uninject` will now just set the `linkerd.io/inject: disabled` annotation on the parent workload. That'll trigger the re-creation of the pod and the auto-injector will ignore it. If the pod is an orphan, the same annotation will be set on the pod's meta, but `apply` might not trigger the `UPDATE` event in the webhook, so we might have to instruct users to use `annotate`. This still needs to be double checked.
- To reman consistent, we will no longer inject the proxies into the control plane pods during install; that'll happen as a side effect when they get applied. Is that ok?

grampelberg:
>  can't support k8s versions lower than 1.9 

I updated support to 1.10+ as part of 2.2.

>  If the pod is an orphan

Orphan pods are weird, can you update the containers in them? Are they restarted?

> To reman consistent, we will no longer inject the proxies into the control plane pods during install; that'll happen as a side effect when they get applied. 

Is there a bootstrapping issue we'd run into here? The MWC wouldn't be running and so none of the control plane pods would be injected?

[...]

PR #2721, opened 2019-04-19
New `linkerd inject` default and advanced modes

### LINKED ISSUE #2710: Update CLI inject to only output annotated YAML
Change the CLI inject's default behaviour to only output a workload's original YAML, plus Linkerd annotations. The injection of the proxy spec will be handled by the proxy injector service.

The list of Linkerd annotations to be included are:

* `linkerd.io/inject: enabled`
* `linkerd.io/created-by`
* `config.linkerd.io/*` for any proxy flags provided by the user

Part of #2576 

Relevant #2164 #2337

alpeb:
Fixes #2710 and #2711 (made more sense to tackle both at once)

This changes the default behavior of `linkerd inject` to not inject the
proxy but just the `linkerd.io/inject: enabled` annotation for the
auto-injector to pick it up (regardless of any namespace annotation).

A new `--advanced` mode was added, which behaves as before, injecting
the proxy in the command output.

The unit tests are running with `--advanced` to avoid any changes in the
fixtures.

I still need to add a new test for non-advanced mode and fix the
integration tests.

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -97,7 +97,7 @@ sub-folders, or coming from stdin.`,
 	flags := options.flagSet(pflag.ExitOnError)
 	flags.BoolVar(
 		&advancedOption, "advanced", advancedOption,
-		"Add the proxy container in the ouput (as opposed to just adding the annotations for the auto-injector to do it) (default false)",
+		"Include the proxy sidecar container spec in the YAML output (the auto-injector won't pick it up, so config annotations aren't supported) (default false)",
 	)
 	flags.BoolVar(
 		&options.disableIdentity, "disable-identity", options.disableIdentity,
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -44,7 +44,7 @@ func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, transforme
 
 func newCmdInject() *cobra.Command {
 	options := &proxyConfigOptions{}
-	var advancedOption, enableDebugSidecar bool
+	var manualOption, enableDebugSidecar bool
 
 	cmd := &cobra.Command{
 		Use:   "inject [flags] CONFIG-FILE",
@@ -83,7 +83,7 @@ sub-folders, or coming from stdin.`,
 			options.overrideConfigs(configs, overrideAnnotations)
 
 			transformer := &resourceTransformerInject{
-				injectProxy:         advancedOption,
+				injectProxy:         manualOption,
 				configs:             configs,
 				overrideAnnotations: overrideAnnotations,
 				enableDebugSidecar:  enableDebugSidecar,
@@ -96,7 +96,7 @@ sub-folders, or coming from stdin.`,
 
 	flags := options.flagSet(pflag.ExitOnError)
 	flags.BoolVar(
-		&advancedOption, "advanced", advancedOption,
+		&manualOption, "manual", manualOption,
 		"Include the proxy sidecar container spec in the YAML output (the auto-injector won't pick it up, so config annotations aren't supported) (default false)",
 	)
 	flags.BoolVar(
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -31,6 +31,7 @@ const (
 )
 
 type resourceTransformerInject struct {
+	injectProxy           bool
 	configs               *cfg.All
 	overrideAnnotations   map[string]string
 	proxyOutboundCapacity map[string]uint
@@ -43,7 +44,7 @@ func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, transforme
 
 func newCmdInject() *cobra.Command {
 	options := &proxyConfigOptions{}
-	var enableDebugSidecar bool
+	var advancedOption, enableDebugSidecar bool
 
 	cmd := &cobra.Command{
 		Use:   "inject [flags] CONFIG-FILE",
@@ -82,6 +83,7 @@ sub-folders, or coming from stdin.`,
 			options.overrideConfigs(configs, overrideAnnotations)
 
 			transformer := &resourceTransformerInject{
+				injectProxy:         advancedOption,
 				configs:             configs,
 				overrideAnnotations: overrideAnnotations,
 				enableDebugSidecar:  enableDebugSidecar,
@@ -93,6 +95,10 @@ sub-folders, or coming from stdin.`,
 	}
 
 	flags := options.flagSet(pflag.ExitOnError)
+	flags.BoolVar(
+		&advancedOption, "advanced", advancedOption,
+		"Add the proxy container in the ouput (as opposed to just adding the annotations for the auto-injector to do it) (default false)",
+	)
 	flags.BoolVar(
 		&options.disableIdentity, "disable-identity", options.disableIdentity,
 		"Disables resources from participating in TLS identity",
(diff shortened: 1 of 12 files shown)


pkg/inject/inject.go
```diff
@@ -187,12 +192,19 @@ func (conf *ResourceConfig) ParseMetaAndYAML(bytes []byte) (*Report, error) {
...
-		conf.injectObjectMeta(patch)
-		conf.injectPodSpec(patch)
+		if len(conf.pod.meta.Annotations) == 0 {
+			patch.addPodAnnotationsRoot()
+		}
+		if injectProxy {
+			conf.injectObjectMeta(patch)
+			conf.injectPodSpec(patch)
+		}
+		conf.injectPodAnnotations(patch)
```
ihcsim: I think the call to `inject.injectPodAnnotations()` needs to happen earlier, because it populates the `conf.pod.meta.Annotations` map used to override the configs in `conf.injectPodSpec()`.

alpeb: Good point :+1:

alpeb: This also means there's a gap in the unit tests to be filled :mag:

alpeb:
> We shouldn't necessarily block merging this on #2715, but I think we should block an edge release on this until #2715 merges.

Yeah I was kinda waiting for #2723 to be finish before tackling #2715, but it seems multi-stage install still needs quite a few more pieces. I'll do #2715 on Monday, should be a quickie.

As for the empty "proxy-version", how can I reproduce that?

siggy:
> As for the empty "proxy-version", how can I reproduce that?
My mistake, I was diffing `linkerd inject -` against `bin/go-run cli inject --advanced -`. The behavior I reported already exists on master.

cli/cmd/inject.go
```diff
@@ -54,6 +55,7 @@ func newInjectOptions() *injectOptions {
 
 func newCmdInject() *cobra.Command {
 	options := newInjectOptions()
+	var advancedOption bool
```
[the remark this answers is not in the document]
alpeb: Right. And also it's just used in `newCmdInject` since it doesn't need to be validated or persisted elsewhere, so I figured I'd just make it a local var.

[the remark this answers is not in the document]
dadjeibaah: Ahh makes sense. Thanks for the explanation.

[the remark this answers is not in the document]
alpeb: btw your suggestion about `rt.injectProxy` makes sense; I'll do that in the next push :+1:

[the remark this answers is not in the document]
alpeb: Actually while implementing that change I realized uninject.go also makes use of `inject.NewResourceConfig` and the resulting `ResourceConfig` but just for parsing the workload (it doesn't call `conf.GetPatch`). It is only when `conf.GetPatch` is called that we perform the injection, so I'd rather leave that argument there in the only function that uses it, and try keeping `ResourceConfig` as succinct as possible.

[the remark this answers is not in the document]
dadjeibaah: Good point. I didn't realize that same object was used for `uninject` as well having the argument separate makes more sense.

pkg/inject/inject.go
alpeb: There were no tests for CLI option overrides. I augmented inject_test.go to take that into account and that did catch the problem you alluded to.

alpeb:
My latest push contains a new test for CLI config overrides, one new test for non-advanced mode, and adds `--advanced` there where injection without the auto-injector is made. In #2715 the auto-injector will become mandatory, so we might adjust the integration tests accordingly by removing that flag wherever possible.

cli/cmd/inject.go
```diff
@@ -93,6 +95,10 @@ sub-folders, or coming from stdin.`,
 	}
 
 	flags := options.flagSet(pflag.ExitOnError)
+	flags.BoolVar(
+		&advancedOption, "advanced", advancedOption,
+		"Add the proxy container in the ouput (as opposed to just adding the annotations for the auto-injector to do it) (default false)",
```
ihcsim: typo: _ouput_

Can we include the implication of using this flag (i.e. the proxy injector will skip the YAML output, hence, auto-injection and config override won't work)? How about _Include the proxy sidecar container spec in the YAML output. Note that proxy configuration annotations aren't supported._?

adleong:
The flag name of "advanced" is not very descriptive of what this flag actually does (although the flag help text does a good job of explaining it).  What do you think of something like "manual" instead of "advanced"?  "Advanced" sounds aspirational; as if you should strive to become an advanced user so that you can use inject in advanced mode.  But, in reality, no one should be using this flag unless they have a very specific usecase for it.

adleong:
The existence of this flag means that we can't assume that meshed pods have been injected by the auto-injector and we still have to support both ways that a pod can become meshed.  Is the plan to eventually phase this flag out?  That would yield a smaller surface area to support and let us make more assumptions about meshed pods.

alpeb:
I do like the `--manual` name better.

>  Is the plan to eventually phase this flag out? That would yield a smaller surface area to support and let us make more assumptions about meshed pods.

I believe not. The idea with `--manual` is to support advanced cases where we need the injected yaml before it gets sent to the cluster, like in CD pipelines where something like kustomize would modify some proxy parameter.

> So without the --advanced flag, all the proxy flags (--admin-port, --control-port) will have no effects, right? I wonder if we should just add a comment in those flags description like (advanced mode only)?

They do have an effect, being transformed into override annotations.

adleong:
I may be missing some context here but by my reading of https://github.com/linkerd/linkerd2/issues/2576 the original problem that this is trying to solve is that there's a discrepancy between the specs of manually vs automatically injected deployments.  I'm unclear on how changing the default behavior to automatic fixes this.

ihcsim:
Essentially, we are forcing all workloads to be auto-injected by the proxy injector. This will make all CLI injected-workload consistent with the 2.3 auto-inject workload, where the source of truth lies in the pod YAML, not the parent's YAML.

The CLI `manual` mode exists (as @grampelberg puts in) a break-the-glass measure. The implication there is that workload will bypass the mutating webhook, hence, pod mutation and config annotations aren't supported.

> That would yield a smaller surface area to support

Both the CLI `manual` mode and proxy injector use the same code path to inject the proxy. The difference lies in the "API boundary" (string vs. k8s admission types). So in a way, the surface area is small.

>  and let us make more assumptions about meshed pods.

Let us know if you think we have overlooked anything.

cli/cmd/inject.go
alpeb: I think it's important to mention the auto-injector, so we could put something like
```
Include the proxy sidecar container spec in the YAML output (the auto-injector won't pick it up, so config annotations aren't supported) (default false)
```
A bit long still...

adleong:
Thanks for the additional context, @ihcsim!  This is very helpful.

My concern about surface area is that when debugging (or even when developing features) there isn't a consistent source of truth for the injected YAML because it may be in the parent's YAML if manual injection was used or in the pod YAML otherwise.  This is essentially the problem statement from #2576.

But this change does make it so that the vast majority of users will use automatic injection which may make support easier in most cases.  We should just keep in mind that keeping manual injection around means that we always have to be aware that there are two potential sources of truth.

cli/cmd/inject.go
ihcsim: Or we can just create a ticket in [website](https://github.com/linkerd/website) to document it.

Issue #2671, opened 2019-04-09
Listen for ConfigMap updates to mitigate timing issues during install


siggy:
## Problem statement

The `linkerd upgrade` command may deploy and restart new control-plane components before an updated `linkerd-config` ConfigMap is available. Because the control-plane components read `linkerd-config` via a volume mount at startup time, any subsequent updates to the ConfigMap go unnoticed.

## Proposal

Modify the control-plane components to poll the ConfigMaps, or listen for changes via Kubernetes. Also validate the ConfigMap is the expected version.

## Open questions

Upcoming multi-stage install work may render this work moot, if we can guarantee that ConfigMaps are always deployed in a stage prior to dependent pods starting up:
https://github.com/linkerd/linkerd2/pull/2656#issuecomment-480980518

ihcsim:
> i thought the problem we saw was that the proxy-autoinjector too eagerly loaded a newer config than it knew how to support 

I bet that it happens right at the start of the [inject handler](https://github.com/linkerd/linkerd2/blob/abcb8bb22cab1245471c7926bf9c08777b34f960/controller/proxy-injector/webhook.go#L24-L27), before the injector even gets a chance to decide whether to skip the pod mutation or not.

siggy:
In some cases we are hot-loading, in others we are reading at startup time:
https://github.com/search?q=MountPathGlobalConfig&type=Code
https://github.com/search?q=MountPathInstallConfig&type=Code
https://github.com/search?q=MountPathProxyConfig&type=Code

The goal of this ticket is to identity and fix the failure modes where there is a mismatch between the ConfigMap version and the control-plane component version, typically manifesting during install/upgrade when things are changing.

olix0r:
> The goal of this ticket is to identity and fix the failure modes where there is a mismatch between the ConfigMap version and the control-plane component version, typically manifesting during install/upgrade when things are changing.

Ok, good. I think we explicitly do NOT want to hot-reload configs in some places, then... This needs to be audited on a case-by-case basis. Specifically, the identity service should NOT be able to hot-reload it's trust roots, etc

alpeb:
We're currently hot-reloading in the proxy auto-injector and the gRPC `Config` endpoint, and not hot-reloading everywhere else (Destination, Identity and Web services). That setup sounds correct to me.

[...]

PR #2723, opened 2019-04-19
Introduce multi-stage upgrade

siggy:
`linkerd install` supports a 2-stage install process, `linkerd upgrade`
did not.

Add 2-stage support for `linkerd upgrade`. Also exercise multi-stage
functionality during upgrade integration tests.

Part of #2337

Depends on #2719

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -18,8 +18,10 @@ import (
 )
 
 const (
-	okMessage   = "You're on your way to upgrading Linkerd!\nVisit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps\n"
-	failMessage = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting\n"
+	okMessage           = "You're on your way to upgrading Linkerd!"
+	controlPlaneMessage = "Don't forget to run `linkerd upgrade control-plane`!"
+	visitMessage        = "Visit this URL for further instructions: https://linkerd.io/upgrade/#nextsteps"
+	failMessage         = "For troubleshooting help, visit: https://linkerd.io/upgrade/#troubleshooting\n"
 )
 
 type upgradeOptions struct {
@@ -117,6 +119,10 @@ install command.`,
 			buf.WriteTo(os.Stdout)
 
 			fmt.Fprintf(os.Stderr, "\n%s %s\n", okStatus, okMessage)
+			if stage == configStage {
+				fmt.Fprintf(os.Stderr, "%s\n", controlPlaneMessage)
+			}
+			fmt.Fprintf(os.Stderr, "%s\n\n", visitMessage)
 
 			return nil
 		},
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -39,21 +39,37 @@ func newCmdUpgrade() *cobra.Command {
 	flags := options.recordableFlagSet()
 
 	cmd := &cobra.Command{
-		Use:   "upgrade [flags]",
-		Short: "Output Kubernetes configs to upgrade an existing Linkerd control plane",
+		Use:       fmt.Sprintf("upgrade [%s|%s] [flags]", configStage, controlPlaneStage),
+		Args:      cobra.OnlyValidArgs,
+		ValidArgs: []string{configStage, controlPlaneStage},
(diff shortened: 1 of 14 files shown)


cli/cmd/upgrade.go
```diff
@@ -39,21 +39,37 @@ func newCmdUpgrade() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if options.ignoreCluster {
 				panic("ignore cluster must be unset") // Programmer error.
 			}
 
+			stage, err := validateArgs(args, flags, nil)
+			if err != nil {
+				return err
+			}
+
```
dadjeibaah: Just noticed this now, when trying to test `--from-manifests` after @alpeb's comment. It seems, on line 79 of this file, that we are using a testing artifact to create a Kubernetes clientset. Is this intended? Wasn't clear on how its being used.

siggy: Good question! (Also I forgot about @alpeb's original question re: allowing `linkerd config --from-manifests`, have pushed a new commit to prevent that.)

Re: `k8s.NewFakeClientSetsFromManifests` on line 79, this is intended. Adding `--from-manifests` required reading k8s yaml from stdin (or files), and processing them in lieu of k8s resources returned from the k8s api. We already had some test code doing this to mock out k8s clients. The change in #2697 moved that test code into a new `fake.go` file, to support `--from-manifests`. Leveraging `fake.go` allows most of the upgrade code to be agnostic of where the k8s resources came from, since it's just using a k8s client.

cli/cmd/upgrade.go
```diff
@@ -96,18 +127,12 @@ install command.`,
...
-		"Read config from a Linkerd install YAML rather than from Kubernetes",
-	)
-
 	cmd.PersistentFlags().AddFlagSet(flags)
+	cmd.PersistentFlags().AddFlagSet(upgradeOnlyFlags)
 	return cmd
 }
 
-func (options *upgradeOptions) validateAndBuild(k kubernetes.Interface, flags *pflag.FlagSet) (*installValues, *pb.All, error) {
+func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Interface, flags *pflag.FlagSet) (*installValues, *pb.All, error) {
```
ihcsim: I don't think `stage` is needed in this function; should be able to assign it to `values` after the call to `validateAndBuild()`.

alpeb:
I verified `--from-manifests` is now only available in the second stage :+1: 

> there is a chance that after running `linkerd upgrade config`, the control plane is left in limbo if the controller and other components aren't restarted

Probably an extra message in stderr alongside the existing `"You're on your way to upgrading Linkerd!"` could do the trick, like `"Don't forget to run linkerd upgrade control-plane!"` (in bold if possible?)

> I wonder how we can let users know when the need for linkerd upgrade config is needed, which would only be occasionally. First thing that occurs to me is hard-coding some boolean const alongside version.Version indicating that. Then linkerd upgrade control-plane would just issue a warning in stderr about it.

A different take to ensure the second stage is ran after the first would be to add an annotation into linkerd's namespace with the version string. The second stage would retrieve the namespace and ensure it has the proper version. This could also apply to `linkerd install`, through a refactor of `exitIfNamespaceDoesNotExist()`. This also means both stages would always need to be run. An extra `linkerd check` could be added to ensure the namespace's version would match the control plane's.

ihcsim:
> Probably an extra message in stderr alongside the existing "You're on your way to upgrading Linkerd!" could do the trick, like "Don't forget to run linkerd upgrade control-plane!" (in bold if possible?)

Yeah, the scenario I was thinking about is _hey, your control plane is broken, until you run `linkerd upgrade control-plane` and restart the controller_

siggy:
I agree there's more we could do here around version checking during an upgrade. Some of that will happen via multi-stage support in `linkerd check`, and I think most of it will happen as we consider versioning as part of #2671.

cli/cmd/upgrade.go
siggy: I considered this initially as it would have minimized the change and preserved `validateAndBuild` signature. I opted to set `stage` in this function keep all the code that creates an `installValues` in one place. The alternative is that `validateAndBuild` creates a nearly-complete `installValues` that the caller must know to set `stage` after the fact.

PR #2733, opened 2019-04-22
Make the auto-injector required and removed proxy-auto-inject flag

### LINKED ISSUE #2715: Make proxy injector a required component of the control plane

alpeb:
Fixes #2715 

~~Still to-do: refactor the integration tests~~


chart/templates/proxy_injector.yaml
```diff
@@ -80,4 +79,3 @@ spec:
     targetPort: proxy-injector
 ---
```
siggy: realizing this was already an issue, but now that it's default, mind removing the trailing `---` from this file, for consistency with the other templates?

proto/config/config.proto
```diff
@@ -21,10 +21,6 @@ message Global {
 
   // If present, configures identity.
   IdentityContext identity_context = 4;
-
-  // If present, indicates that the Mutating Webhook Admission Controller should
-  // be configured to automatically inject proxies.
-  AutoInjectContext auto_inject_context = 6;
```
siggy: i think we should keep this as a deprecated placeholder, so as not to break newer components/cli reading an older config?
```proto
AutoInjectContext auto_inject_context = 6 [deprecated=true];
```

proto/config/config.proto
```diff
@@ -63,9 +59,6 @@ message ResourceRequirements {
   string limit_memory = 4;
 }
 
-// Currently, this is basically a boolean.
-message AutoInjectContext {}
```
siggy:
```proto
message AutoInjectContext { option deprecated = true; }
```

proto/config/config.proto
alpeb: I thought it would be cleaner to just remove it, given the unmarshaller is set to ignore unknown fields, so it actually doesn't brake. You agree?

ihcsim: My first thought is also to mark this as deprecated. Per the [grpc doc](https://developers.google.com/protocol-buffers/docs/proto3#reserved), if later on we assign field `6` to another field, it will break older clients which still depend on field `6` to be the `auto_inject_context`. If we really want to remove it, we can use `reserved` to ensure `6` doesn't get recycled.

alpeb: Field numbers are used to identify the fields in the protobuf message binary format, but we're not storing the config in the protobuf binary format; we're storing them in JSON, and we're only using protobuf to map the JSON into the go structs. From my understanding, such mapping is done by matching fields labels and order doesn't matter. One can test this by manually shifting order on some of the proxy config fields in the ConfigMap, restarting the injector, and checking the behavior is the same.

alpeb: After further discussion over Slack, we realized `linkerd inject` does receive that config via a binary message, so we gotta deprecate that field instead of just removing it.

ihcsim:
This branch still works for me, after the rebase :+1:.

Sorry, I missed this in my first `grep`. Can you remove this TODO too? Thanks.
```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 6e5caabb..e2d57703 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -921,12 +921,7 @@ func validateArgs(args []string, flags *pflag.FlagSet, installOnlyFlags *pflag.F
 		var err error
 		combinedFlags.VisitAll(func(f *pflag.Flag) {
 			if f.Changed {
-				switch f.Name {
-				// TODO: remove "proxy-auto-inject" when it becomes default
-				case "proxy-auto-inject":
-				default:
-					err = fmt.Errorf("flag not available for config stage: --%s", f.Name)
-				}
+				err = fmt.Errorf("flag not available for config stage: --%s", f.Name)
 			}
 		})
 		if err != nil {
```

cli/cmd/install.go
```diff
@@ -933,18 +920,14 @@ func validateArgs(args []string, flags *pflag.FlagSet, installOnlyFlags *pflag.F
...
-				case "proxy-auto-inject":
-				default:
-					err = fmt.Errorf("flag not available for config stage: --%s", f.Name)
-				}
+				invalidFlags = append(invalidFlags, f.Name)
 			}
 		})
-		if err != nil {
+		if len(invalidFlags) > 0 {
+			err := fmt.Errorf("flags not available for config stage: --%s", strings.Join(invalidFlags, ", --"))
```
siggy: 👍

test/install_test.go
```diff
@@ -268,30 +264,15 @@ func TestInject(t *testing.T) {
 
 	prefixedNs := TestHelper.GetTestNamespace("smoke-test")
 
-	if TestHelper.AutoInject() {
-		out, err = testutil.ReadFile("testdata/smoke_test.yaml")
-		if err != nil {
-			t.Fatalf("failed to read smoke test file: %s", err)
-		}
-		err = TestHelper.CreateNamespaceIfNotExists(prefixedNs, map[string]string{
```
siggy: not necessarily in this PR, but consider testing manual injection as well. if we're going to continue supporting it, we should exercise cli injection somewhere in our integration tests.

alpeb: Agreed. I added #2744 to track that.

Issue #2741, opened 2019-04-24
Service profile integration tests seem to be flakey.


dadjeibaah:
According to https://travis-ci.org/linkerd/linkerd2/jobs/523726427#L1050 It looks like service profile integration tests are failing inconsistently. The tests run by checking to see that we have the expected routes from the `linkerd routes` command. It could be that when the routes command is ran, the CLI does not have the expected routes listed for a particular deployment before the test timeout completes. Which sometimes causes the tests to fail.

alpeb:
This is the api controller log from before `routes` is only returning `[DEFAULT]` until right after it starts working:
```
time="2019-04-24T17:28:11Z" level=debug msg="Serving POST /api/v1/TopRoutes" req.Form="map[]" req.Method=POST req.URL=/api/v1/TopRoutes
time="2019-04-24T17:28:11Z" level=debug msg="TopRoutes request: selector:<resource:<namespace:\"l5d-integration-serviceprofile-test\" type:\"deployment\" name:\"t1\" > > time_window:\"1m\" none:<> "                                                                         
time="2019-04-24T17:28:11Z" level=debug msg="no Service Profile found for 't1-svc.l5d-integration-serviceprofile-test.svc.cluster.local' -- using default"                                                                                                                     
time="2019-04-24T17:28:11Z" level=debug msg="Query request:\n\thistogram_quantile(0.99, sum(irate(route_response_latency_ms_bucket{deployment=\"t1\", direction=\"inbound\", dst=~\"(t1-svc.l5d-integration-serviceprofile-test.svc.cluster.local)(:\\\\d+)?\", namespace=\"l5d-integration-serviceprofile-test\"}[1m])) by (le, dst, rt_route))"
time="2019-04-24T17:28:11Z" level=debug msg="Query request:\n\tsum(increase(route_response_total{deployment=\"t1\", direction=\"inbound\", dst=~\"(t1-svc.l5d-integration-serviceprofile-test.svc.cluster.local)(:\\\\d+)?\", namespace=\"l5d-integration-serviceprofile-test\"}[1m])) by (rt_route, dst, classification)"
[...]

PR #2724, opened 2019-04-19
Report known logging errors in test output

siggy:
The integration tests check container logs for errors. When an error is
encountered that matches a list of expected errors, it was hidden, and
the test passed.

Modify the integration tests to report known errors in logs via
`t.Skipf`.

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

PR #2742, opened 2019-04-24
skip serviceprofile tests

dadjeibaah:
Related to #2741 Skipping these tests for now until we figure out why these tests are flakey

### DIFF
--- a/test/serviceprofiles/serviceprofiles_test.go
+++ b/test/serviceprofiles/serviceprofiles_test.go
@@ -73,6 +73,7 @@ func TestServiceProfiles(t *testing.T) {
 
 	for _, tc := range testCases {
 		tc := tc // pin
+		t.Skipf("Skipping TestServiceProfiles/%s", tc.sourceName)
 		t.Run(tc.sourceName, func(t *testing.T) {
 			routes, err := getRoutes(tc.deployName, tc.namespace, []string{})
 			if err != nil {
@@ -140,6 +141,7 @@ func TestServiceProfileMetrics(t *testing.T) {
 		)
 
 		t.Run(tc, func(t *testing.T) {
+			t.Skipf("Skipping TestServiceProfileMetrics/%s", tc)
 			out, stderr, err := TestHelper.LinkerdRun("inject", "--manual", testYAML)
 			if err != nil {
 				t.Errorf("'linkerd %s' command failed with %s: %s\n", "inject", err.Error(), stderr)


test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -73,6 +73,7 @@ func TestServiceProfiles(t *testing.T) {
 
 	for _, tc := range testCases {
 		tc := tc // pin
+		t.Skipf("Skipping TestServiceProfiles/%s", tc.sourceName)
```
siggy: rather than skipping the whole test, what about replacing some or all of the `t.Fatalf`'s below with `t.Skipf`? is there a particular case that's failing so we can only skip a single check?

test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -140,6 +141,7 @@ func TestServiceProfileMetrics(t *testing.T) {
 		)
 
 		t.Run(tc, func(t *testing.T) {
+			t.Skipf("Skipping TestServiceProfileMetrics/%s", tc)
```
siggy: same comment as above.

test/serviceprofiles/serviceprofiles_test.go
dadjeibaah: Based on [these](https://travis-ci.org/linkerd/linkerd2/jobs/523726427#L1050) integration logs, it looks like all the tests were failing for some reason. The failures are attributed to when service profiles are applied but do not show up immediately in the `linkerd routes` command. Since the command is pretty much used in each test, I thought disabling it this way might be better.

ihcsim: Have you tried restarting the CI job to see if that helps? Looks like the log tests caught (and skipped) a bunch of timeout errors prior to the service profile tests.

dadjeibaah: @adleong initially encountered this issue first, and then tried restarting but didn't seem to self-correct. Weird that the some tests were skipped, which tests were they @ihcsim?

ihcsim: The `TestLogs()` test function skips, but outputs, unexpected errors. I was just surprised to see the number of timeout errors in your integration logs; possibly indicating an unhealthy cluster.

Looks like the integration test just [passed](https://gist.github.com/l5d-bot/3be5b6475d9bff94a448f79b8aba425a)?

dadjeibaah: The integration tests that just passed are for this brunch which skips the service profile tests entirely. However, I do see `TestLogs()` skipping tests on this branch as well. I too suspect that the cluster is out of wack, pods are not being scheduled or something. it's a bummer that these tests depends on how healthy a cluster is. I'd need to spend some time thinking about a more reliable way to test this functionality 🤔

alpeb:
I can reliable reproduce the test failures on my local only while using code after (and including) this [commit](3de16d47be61bf5b845e61e8752717b6f0e55370). It does take a very long while for routes to get updated. I'm digging through logs...

test/serviceprofiles/serviceprofiles_test.go
siggy: Based on [that test log](https://travis-ci.org/linkerd/linkerd2/jobs/523726427#L1050), it looks like we could instead replace the `t.Errorf`'s on lines 243 (`assertRouteStat`) and 250 + 266 (`assertExpectedRoutes`) with `t.Skipf`'s. This way, we're still exercising the `linkerd profile` command (and that at least `linkerd routes` does not return an error). The intent here is to skip the narrowest set of testing possible.

siggy: Ah, FWIW re: `TestLogs()`, we recently enabled rendering of all log failures via #2724, previously we were just hiding them, so things are not necessarily worse, it's just the logging issues are more visible. =)

dadjeibaah:
Closing this since the tests are working as intended.

PR #2748, opened 2019-04-25
[WIP] Update service profile CRD to v1beta2

adleong:


alpeb:
The integration tests are passing now because we first install the latest git version, then we install the latest Edge to test an upgrade from it. When installing the Edge we're sorta attempting to "downgrade" the CRD, which fails, but everything else gets installed:
```
# running this with the latest git CRD
bin/go-run cli upgrade | kubectl apply -f -
namespace/linkerd created
configmap/linkerd-config created
serviceaccount/linkerd-identity created
clusterrole.rbac.authorization.k8s.io/linkerd-linkerd-identity unchanged
clusterrolebinding.rbac.authorization.k8s.io/linkerd-linkerd-identity unchanged
service/linkerd-identity created
secret/linkerd-identity-issuer created
deployment.extensions/linkerd-identity created
serviceaccount/linkerd-controller created
clusterrole.rbac.authorization.k8s.io/linkerd-linkerd-controller unchanged
clusterrolebinding.rbac.authorization.k8s.io/linkerd-linkerd-controller unchanged
service/linkerd-controller-api created
service/linkerd-destination created
deployment.extensions/linkerd-controller created
serviceaccount/linkerd-web created
service/linkerd-web created
deployment.extensions/linkerd-web created
serviceaccount/linkerd-prometheus created
clusterrole.rbac.authorization.k8s.io/linkerd-linkerd-prometheus unchanged
clusterrolebinding.rbac.authorization.k8s.io/linkerd-linkerd-prometheus unchanged
service/linkerd-prometheus created
deployment.extensions/linkerd-prometheus created
configmap/linkerd-prometheus-config created
serviceaccount/linkerd-grafana created
service/linkerd-grafana created
deployment.extensions/linkerd-grafana created
configmap/linkerd-grafana-config created
serviceaccount/linkerd-sp-validator created
clusterrole.rbac.authorization.k8s.io/linkerd-linkerd-sp-validator unchanged
clusterrolebinding.rbac.authorization.k8s.io/linkerd-linkerd-sp-validator configured
service/linkerd-sp-validator created
deployment.extensions/linkerd-sp-validator created
The CustomResourceDefinition "serviceprofiles.linkerd.io" is invalid: status.storedVersions[0]: Invalid value: "v1alpha2": must appear in spec.versions
```
So the Edge will be ran with the CRD from git, which in this case doesn't hurt, and the `l5d-integration` isn't affected.
I think this is fine for now.
It would be great if, in the future, we could test upgrades from different versions, each test in a totally fresh environment, probably as a set of separate tests outside the integration tests.

adleong:
My primary concern is the upgrade from previous edge to next edge (or similarly from the previous stable to the next stable).  I've tested this upgrade path manually and works as I would expect: the controller pod fails to start and goes into crash loop backoff until the CRD is upgraded and then starts successfully.  After the upgrade, service profiles that had been previously created at version v1alpha1 continue to work.

However, I have not been able to get integration tests passing on minikube or on my dev cluster in GKE.  Unsure if this is an issue with insufficient resources or if there's some other problem.  I could use help testing this and getting confidence that users won't run into problems on upgrade.

ihcsim:
@adleong The way you tested the upgrade path and the observed behaviour sounds right to me. The integration test is odd because essentially, we have two control planes sharing the same global resources (RBAC, CRD, MWC etc.). AFAICT, most (if not all) of the integration tests only target the `l5d-integration` control plane. I think there are no reasons why we can't restructure the tests to use only one control plane, from installing an older edge (using the dev-only  `--control-plane-version` flag), to upgrading it, then executing the remaining test suites.

> I have not been able to get integration tests passing on minikube

Same here. I find chances are succeeding are higher with I give the (kvm2) VM at least 3G of RAM.

siggy:
Re: the shared global resources (RBAC, CRD, MWC etc.), I think we should be intentional about whether we support multiple control-planes in one cluster, and if so, we should test for that. Separately, I think it's also important that, for every PR, we test a fresh install as well as an upgrade from a previous version.

adleong:
If our integration tests are exercising an upgrade across versions of a global resource (like a CRD) then I think we're always going to run into issues with our tests being multi-tenant.  Imagine test1 is testing the upgrade path and sets the CRD version to v1alpha1; this will break any concurrently running test which is expecting the CRD version to be v1alpha2.  Since CRDs are shared, I don't think there's any safe way for these tests to run simultaneously in the same cluster.

dadjeibaah:
When running tests on minikube, the ones that fail are related to version checks for some reason. Not quite sure why that is the case.
```bash
--- FAIL: TestVersionPreInstall (0.06s)
    install_test.go:92: Version command failed
        Expected server version [edge-19.4.4], got:
        Client version: git-fe07d0f7
        Server version: stable-2.3.0
...
--- FAIL: TestVersionPostInstall (5.08s)
    install_test.go:192: Version command failed
        Expected server version [git-fe07d0f7], got:
        Client version: git-fe07d0f7
        Server version: unavailable
```

adleong:
A few potential options are:
* force integration tests to run in series and delete global resources after each run
* have each integration test spin up its own dedicated cluster for the duration of the test

In the short term, however, I think we may want to revert the change to the CRD to give ourselves more time to figure out how to proceed.  CRD versioning is an important problem that we will definitely need to solve though.

siggy:
notes for reference, not suggesting we do anything at this moment:
> force integration tests to run in series and delete global resources after each run

l5d-bot runs tests serially, so cleaning up global resources is an easy change. travis on the other hand does not run integration tests serially, so would require more work

> have each integration test spin up its own dedicated cluster for the duration of the test

this could work for both l5d-bot and travis, it may be more up-front work than the first option

siggy:
I'm onboard with merging this PR, or rolling back the ServiceProfile schema change.

While I agree versioning CRDs needs further consideration, we definitely will have to update CRD versions at some point. It may be beneficial to roll forward with this PR, just to gain some experience with the process and its effects.

siggy:
on l5d-bot:

```bash
$ kubectl get crd/serviceprofiles.linkerd.io -oyaml
...
spec:
  version: v1alpha1
  versions:
  - name: v1alpha1
    served: true
    storage: false
  - name: v1alpha2
    served: true
    storage: true
```

adleong:
https://github.com/linkerd/linkerd2/issues/2753

PR #2755, opened 2019-04-25
Release notes for Edge-19.4.5

alpeb:


CHANGES.md
```diff
@@ -1,3 +1,66 @@
+## edge-19.4.5
+
+**Significant Update**
+
+As of this edge release the proxy auto-injector component is always installed.
+To have the auto-injector inject a pod you still have the option to manually
+add the `linkerd.io/inject: enable` annotation into the pod spec, or at the
+namespace level to have all your pods be injected by default.
+With this release the behaviour of the `linkerd inject` command changes, as
```
scottcarol: We need a stylesheet, but in the past we have always used the non-British "behavior" spelling

alpeb: Bloody hell, I keep on making those kind of mistakes! :gb:

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+require namespace-level privileges. This also applies to the `linkerd upgrade`
+command.
+
+* CLI
+  * **Breaking Change** Removed the `--proxy-auto-inject` flag, as the
+    auto-injector is now always installed
+  * **Breaking Change** Replaced the `--linkerd-version` flag with the
+    `--proxy-version` flag in the `linkerd install` and `linkerd upgrade`
+    commands, which allows setting the version for the injected sidecar image,
+    without changing the image versions for the control planel
```
scottcarol: panel (typo)

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+add the `linkerd.io/inject: enable` annotation into the pod spec, or at the
+namespace level to have all your pods be injected by default.
+With this release the behaviour of the `linkerd inject` command changes, as
+now it won't output the injected sidecar, but instead it will just add the
+annotation to defer the injection to the auto-injector.
+For advanced CI/CD pipelines where you'd like to keep the old behaviour, you can
+use the new `--manual` flag to output the injected sidecar as before.
+
+Another important update is the introduction of install stages. You still have
+the old `linkerd install` command, but now it can be broken into `linkerd install config`
```
siggy: nit: wrap `linkerd install config` to the next line

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+    `--proxy-version` flag in the `linkerd install` and `linkerd upgrade`
+    commands, which allows setting the version for the injected sidecar image,
+    without changing the image versions for the control planel
+  * Introduced install stages: `linkerd install config` and
+    `linkerd install control-plane`
+  * Introduced upgrade stages: `linkerd upgrade config` and
+    `linkerd upgrade control-plane`
+  * Introduced a new `--from-manifests` flag to `linkerd upgrade` allowing
+    manually feeding a previously saved output of `linkerd install` into the
+    command, instead of requiring to connect to the cluster to fetch the config
```
scottcarol: would this be better as "instead of requiring a connection to the cluster" ?

CHANGES.md
siggy: `planel`

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+  * Introduced a new `--manual` flag to `linkerd inject` to output the proxy
+    sidecar container spec
+  * Introduced a new `--enable-debug-sidecar` option to `linkerd inject`, that
+    injects a debug sidecar to inspect traffic to and from the meshed pod
+  * Added a new check for unschedulable pods and PSP issues (thanks, @liquidslr!)
+  * Disabled the spinner in `linkerd check` when running without a tty
```
scottcarol: i have no idea, but should this be 'TTY' instead of tty? or should we spell it out?

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+    Deployment to avoid warnings when installing the auto-injector
+    (thanks, @dwj300!)
+
+* Controller
+  * Added a profiling endpoint for all the control plane components, to better
+    assist debugging efforts
+  * Fixed bug in the auto-injector, where sporadically the pod workload owner
+    wasn't properly determined, which would result in erroneous stats
+  * Added support for a new `config.linkerd.io/disable-identity` annotation to
+    opt-out of identity for a specific pod
```
scottcarol: I think this would just be "opt out" with no dash

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+
+* Web UI
+  * Added the Font Awesome stylesheet locally; this allows seeing both Font
+    Awesome and Material-UI sidebar icons consistently with no/limited Internet
```
scottcarol: i would rephrase as "this allows both Font Awesome and Material-UI sidebar icons to display consistently with no/limited internet access"

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+    (thanks again, @liquidslr!)
+
+* Internal
```
dadjeibaah: I don't think we need to highlight the new tests we added since these are not direct production level concerns for users.

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+  * Introduced a new `--enable-debug-sidecar` option to `linkerd inject`, that
+    injects a debug sidecar to inspect traffic to and from the meshed pod
+  * Added a new check for unschedulable pods and PSP issues (thanks, @liquidslr!)
+  * Disabled the spinner in `linkerd check` when running without a tty
+  * Ensured the ServiceAccount for the auto-injector is created before its
+    Deployment to avoid warnings when installing the auto-injector
+    (thanks, @dwj300!)
+
+* Controller
+  * Added a profiling endpoint for all the control plane components, to better
```
dadjeibaah: How about "Added Go [pprof](https://golang.org/pkg/net/http/pprof/) HTTP endpoints to all control plane components' admin servers to better assist debugging efforts"

CHANGES.md
```diff
@@ -1,3 +1,66 @@
+## edge-19.4.5
+
+**Significant Update**
+
+As of this edge release the proxy auto-injector component is always installed.
```
ihcsim: I think we should use the actual component name i.e. proxy injector.

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+To have the auto-injector inject a pod you still have the option to manually
```
ihcsim: _To have the proxy injector auto-injects a pod_

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+add the `linkerd.io/inject: enable` annotation into the pod spec, or at the
+namespace level to have all your pods be injected by default.
+With this release the behaviour of the `linkerd inject` command changes, as
+now it won't output the injected sidecar, but instead it will just add the
```
ihcsim: How about _command changes, where the proxy sidecar container YAML is no longer included in its output by default_?

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+annotation to defer the injection to the auto-injector.
```
ihcsim: _annotations_.

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+For advanced CI/CD pipelines where you'd like to keep the old behaviour, you can
```
ihcsim: I think we should omit the _advanced CI/CI pipelines_, as the `--manual` flag will be relevant to other uses cases like helm, kustomize etc. How about something like, _For use cases that require the full injected YAML to be output, a new `--manual` flag has been added_?

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+Another important update is the introduction of install stages. You still have
+the old `linkerd install` command, but now it can be broken into `linkerd install config`
+which installs the resources that require cluster-level privileges, and
+`linkerd install control-plane` that continues with the resources that only
+require namespace-level privileges. This also applies to the `linkerd upgrade`
+command.
+
+* CLI
+  * **Breaking Change** Removed the `--proxy-auto-inject` flag, as the
+    auto-injector is now always installed
```
ihcsim: Prefer _proxy injector_ since that's the component's actual name.

CHANGES.md
```diff
@@ -1,3 +1,66 @@
...
+  * **Breaking Change** Replaced the `--linkerd-version` flag with the
+    `--proxy-version` flag in the `linkerd install` and `linkerd upgrade`
+    commands, which allows setting the version for the injected sidecar image,
```
ihcsim: I think we should be more specific about which sidecar, since we can also now inject debug sidecar container. So _injected proxy sidecar image_?

CHANGES.md
ihcsim: Also, _still have the option_ sounds optional. For auto-inject to work, that annotation must be added. How about _you can manually add the_?

CHANGES.md
scottcarol: It's my least favourite kind of mistake because it's just a nit!

CHANGES.md
alpeb: It is TTY according to Wikipedia; I'll change it :+1

Issue #1981, opened 2018-12-13
GKE Private Clusters Cannot Proxy


joshkurz:
## Bug Report

GKE Private Clusters cannot proxy requests via `kubectl proxy` from the looks of it. This breaks `linkerd dashboard` command and the `linkerd check` command. I don't think this is a linkerd issue after looking into it further. I cannot proxy any services on GKE Private Clusters.

### What is the issue?

`linkerd dashboard`
`linkerd check`

both of these fail, because the proxy times out.

### How can it be reproduced?

Create a GKE private cluster and install linkerd.

### Logs, error output, etc

(If the output is long, please create a [gist](https://gist.github.com/) and
paste the link here.)

#### `linkerd check` output

```text
NYJKurzMBP:linkerd2 joshua.kurz$ linkerd check --verbose
kubernetes-api: can initialize the client..................................[ok]
kubernetes-api: can query the Kubernetes API...............................[ok]
kubernetes-api: is running the minimum Kubernetes API version..............[ok]
linkerd-api: control plane namespace exists................................[ok]
linkerd-api: control plane pods are ready..................................[ok]
DEBU[0000] Expecting API to be served over [https://35.237.232.208/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/]
linkerd-api: can initialize the client.....................................[ok]
DEBU[0000] Making gRPC-over-HTTP call to [https://35.237.232.208/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/SelfCheck] []
DEBU[0005] Error invoking [https://35.237.232.208/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/SelfCheck]: Post https://35.237.232.208/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/SelfCheck: context deadline exceeded
linkerd-api: can query the control plane API...............................[FAIL] -- Post https://35.237.232.208/api/v1/namespaces/linkerd/services/linkerd-controller-api:http/proxy/api/v1/SelfCheck: context deadline exceeded

```

### Environment

- Kubernetes Version:

```
kubectl version
[...]

PR #2757, opened 2019-04-26
Use port-forwarding for linkerd CLIs

jackprice:
Hi,

Private k8s clusters, such as the private GKE clusters offered by Google
Cloud, cannot be reached throught the current API proxy method. (see #1981)

This merge switches to using the port forwarding feature already developed and I believe implemented for the dashboard.

I'm not sure how to go about getting this tested, perhaps someone can advise?


siggy:
@jackprice After a bit more discussion in linkerd slack, and taking some inspiration from @samcgardner's work in #2776, I've modified your PR a bit to make `NewPortForward()` use an ephemeral by default, and have the dashboard code specifically request it's own default of 50750.

The diff I'm sharing causes a slight behavior change, where `linkerd dashboard` no longer falls back to an ephemeral port if 50750 is in use. This new behavior is consistent with how `kubectl proxy` behaves, so I'm not very concerned. I've left a TODO in the code in case we want to address it, but for your PR I don't think it's necessary. Here's the new error message when this occurs:

```bash
$ bin/go-run cli dashboard
Error running port-forward: Unable to listen on any of the requested ports: [{50750 8084}]
Check for `linkerd dashboard` running in other terminal sessions, or use the `--port` flag.
```

My earlier concerns around the port forwards not cleaning up their connection were primarily motivated in cases where this code could be called via a long-running server, which today is not the case. I've updated the comment on `NewPortForward()` to reflect this.

Here's the diff off of your branch. If you apply it, I think we can get this merged pretty quickly, thanks! 
https://gist.github.com/siggy/1c4671bf413fbe7dce687d5a68a47973

For reference the full branch is at https://github.com/linkerd/linkerd2/compare/siggy/2757-integration:

jackprice:
@siggy thanks for the comments!

I've applied the diff to my branch and everything still seems to work as expected. Would you prefer I squash everything into one commit?

siggy:
@jackprice No need to squash, we'll do that when we merge. I'll give this PR another pass, stay tuned.

siggy:
@jackprice i hit another snag when running integration tests:
```bash
$ DOCKER_TRACE=1 bin/docker-build && bin/test-run `pwd`/bin/linkerd
...
Running test [install_test.go] --linkerd-namespace=l5d-integration
=== RUN   TestVersionPreInstall
--- FAIL: TestVersionPreInstall (0.20s)
    install_test.go:94: Version command failed
        Unexpected error: exit status 1
        Client version: git-4b4ab8df
```

This test failure was due to a behavior change `rawPublicAPIClient`, where it now returns an error because it's unable to establish a port-forward when the public-api is not available. This causes `linkerd version` to exit, rather than returning the expected:
```bash
$ linkerd version
Client version: edge-19.4.5
Server version: unavailable
```

Here's another commit to apply onto your PR, it modifies `configureAndRunVersion()` to more gracefully handle errors, and also includes a test to validate this behavior change:
https://gist.github.com/siggy/031038ac42afb9afea732e6575fc3041

I'm running integration tests again, should be done in about 30 minutes.

Issue #2751, opened 2019-04-25
Move ServiceProfile generated code into pkg


siggy:
## Background

The ServiceProfile generated [client](https://github.com/linkerd/linkerd2/tree/cd37d3f0f50600ce21644f9ae3884b23867a5555/controller/gen/client) and [api](https://github.com/linkerd/linkerd2/tree/cd37d3f0f50600ce21644f9ae3884b23867a5555/controller/gen/apis/serviceprofile) code lives under `./controller/gen` primarily that is where all generated code is centralized. As non-controller components, namely the CLI, depend on this code, it makes less sense to have it there.

## Proposal

Move ServiceProfile generated code into `./pkg`.

This would allow us to move [`NewSpClientSet`](https://github.com/linkerd/linkerd2/blob/cd37d3f0f50600ce21644f9ae3884b23867a5555/controller/k8s/clientset.go) into [`pkg/k8s`](https://github.com/linkerd/linkerd2/blob/cd37d3f0f50600ce21644f9ae3884b23867a5555/pkg/k8s/k8s.go), which would then allow us to make `k8s.GetConfig` private.

This is a continuation of #2747.

klingerf:
We have the same issue for all of the protobuf bindings in `controller/gen` too. Files in the web, pkg, and cli dirs all import protobuf from `controller/gen`, which doesn't seem right.

Based on the layout of the [sample codegen project](https://github.com/kubernetes/sample-controller/tree/master/pkg), maybe we could:

* Move `controller/gen/apis` to `pkg/apis`
* Move `controller/gen/client` to `pkg/generated`
* Move `controller/gen/common` to `pkg/pb/common`
* Move `controller/gen/config` to `pkg/pb/config`
* Move `controller/gen/controller` to `pkg/pb/controller`
* Move `controllergen/public` to `pkg/pb/public`

(note that I previously requested we put all of the codegen code in the `controller/gen` directory, since I was concerned about mixing generated and non-generated code in `pkg`, but on reflection I'm not actually concerned about that)

alpeb:
Will be addressing this one after #2786 lands, to avoid merge conflicts.

PR #2759, opened 2019-04-26
Introduce k8s apiextensions support

siggy:
CustomResourceDefinition parsing and retrieval is not available via
client-go's `kubernetes.Interface`, but rather via a separate
`k8s.io/apiextensions-apiserver` package.

Introduce support for CustomResourceDefintion object parsing and
retrieval. This change facilitates retrieval of CRDs from the k8s API
server, and also provides CRD resources as mock objects.

Also introduce a `NewFakeAPI` constructor, deprecating
`NewFakeClientSets`. Callers need no longer be concerned with discreet
clientsets (for k8s resources vs. CRDs vs. (eventually)
ServiceProfiles), and can instead use the unified `KubernetesAPI`.

Part of #2337, in service to multi-stage check.

### DIFF
--- a/pkg/k8s/api.go
+++ b/pkg/k8s/api.go
@@ -9,6 +9,7 @@ import (
 	"github.com/linkerd/linkerd2/pkg/prometheus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/version"
@@ -22,9 +23,11 @@ import (
 var minAPIVersion = [3]int{1, 10, 0}
 
 // KubernetesAPI provides a client for accessing a Kubernetes cluster.
+// TODO: support spClient
 type KubernetesAPI struct {
 	*rest.Config
 	kubernetes.Interface
+	Apiextensions apiextensionsclient.Interface // for CRDs
 }
 
 // NewAPI validates a Kubernetes config and returns a client for accessing the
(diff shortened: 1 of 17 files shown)


pkg/k8s/api.go
```diff
@@ -22,9 +23,11 @@ import (
 var minAPIVersion = [3]int{1, 10, 0}
 
 // KubernetesAPI provides a client for accessing a Kubernetes cluster.
+// TODO: support spClient
```
olix0r: Is this something that has to happen in this PR?

If not, can this be more descriptive? What's `spClient`? Are there blockers? What's the benefit?

pkg/k8s/fake.go
```diff
@@ -19,29 +22,64 @@ import (
...
+
+	return &KubernetesAPI{
+		Interface:     client,
+		Apiextensions: apiextClient,
+	}, nil
+}
+
+// NewFakeClientSets provides mock Kubernetes ClientSets.
+// TODO: make this private once KubernetesAPI (and NewFakeAPI) supports spClient
+func NewFakeClientSets(configs ...string) (kubernetes.Interface, apiextensionsclient.Interface, spclient.Interface, error) {
```
dadjeibaah: What are your thoughts on making `NewFakeClientSets` return `kubernetesAPI` instead of the two clientsets? Are there instances where we would want to just use the `NewFakeClientSets` instead of `NewFakeAPI`.

pkg/k8s/api.go
siggy: `spClient` == Service Profile ClientSet. It doesn't have to happen in this PR, and maybe shouldn't, because I think we should move ServiceProfile code from `controller` into `pkg` (#2751).

I'll update the comment with all this context.

pkg/k8s/fake.go
siggy: I agree replacing `kubernetes.Interface` and `apiextensionsclient.Interface` with a single `kubernetesAPI` is a good intermediate step. I'm going to leave this for now, because this function still needs to return an `spclient.Interface`, since `kubernetesAPI` does not yet support it. Once we do (via #2751), we can make this function private, and everyone can call `NewFakeAPI`.

PR #2764, opened 2019-04-29
Fix check/dashboard failing from one pod when HA

### LINKED ISSUE #2554: check/dashboard fails when any linkerd pod is down
Even when running with --ha with redundant controllers, `linkerd dashboard` fails to start when a single pod is down:

```
NAME                                  READY     STATUS             RESTARTS   AGE
linkerd-controller-695dcbdcf7-8ssrn   4/4       Running            18         15h
linkerd-controller-695dcbdcf7-9bsvp   1/4       CrashLoopBackOff   292        21h
linkerd-controller-695dcbdcf7-p2pzx   4/4       Running            0          17h
linkerd-grafana-579444c8bd-752pp      2/2       Running            0          21h
linkerd-identity-6c86f797-4b59v       2/2       Running            0          21h
linkerd-identity-6c86f797-g4qdz       2/2       Running            0          21h
linkerd-identity-6c86f797-m4lh6       2/2       Running            0          21h
linkerd-prometheus-6f44484f65-29629   0/2       Evicted            0          20h
linkerd-prometheus-6f44484f65-2qctn   0/2       Evicted            0          21h
linkerd-prometheus-6f44484f65-c9mt8   0/2       Evicted            0          19h
linkerd-prometheus-6f44484f65-kl2w7   2/2       Running            4          16h
linkerd-web-6c8644d47-4fcj8           2/2       Running            0          21h
```

```
:; linkerd check --wait 10s 
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
× control plane pods are ready
    The "linkerd-controller" pod's "destination" container is not ready
    see https://linkerd.io/checks/#l5d-api-control-ready for hints

Status check results are ×
```

This logic should probably only warn if ALL instances are not down.

siggy:
The `linkerd check` and `linkerd dashboard` commands validate control
plane pods are up via the `LinkerdAPIChecks` category of checks. These
checks will fail if a single pod is not ready, even in HA mode.

Modify the underlying `validateControlPlanePods` check to return
successful if at least one pod per control plane component is ready.

Fixes #2554

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -903,6 +903,7 @@ func validateControlPlanePods(pods []corev1.Pod) error {
 			if containersReady {
 				// at least one pod has all containers ready
 				ready = true
+				break
 			}
 		}
 		if !ready {
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -842,8 +842,15 @@ func (hc *HealthChecker) validateServiceProfiles() error {
 	return nil
 }
 
-func getPodStatuses(pods []corev1.Pod) map[string][]corev1.ContainerStatus {
-	statuses := make(map[string][]corev1.ContainerStatus)
+// getPodStatuses returns a map of all Linkerd container statuses:
+// component =>
+//   pod name =>
+//     container statuses
+// "controller" =>
+//   "linkerd-controller-6f78cbd47-bc557" =>
+//     [destination status, public-api status, ...]
+func getPodStatuses(pods []corev1.Pod) map[string]map[string][]corev1.ContainerStatus {
+	statuses := make(map[string]map[string][]corev1.ContainerStatus)
 
 	for _, pod := range pods {
 		if pod.Status.Phase == corev1.PodRunning && strings.HasPrefix(pod.Name, "linkerd-") {
@@ -853,9 +860,9 @@ func getPodStatuses(pods []corev1.Pod) map[string][]corev1.ContainerStatus {
 			if len(parts) >= 4 {
 				name := strings.Join(parts[1:len(parts)-2], "-")
 				if _, found := statuses[name]; !found {
-					statuses[name] = make([]corev1.ContainerStatus, 0)
+					statuses[name] = make(map[string][]corev1.ContainerStatus)
 				}
-				statuses[name] = append(statuses[name], pod.Status.ContainerStatuses...)
+				statuses[name][pod.Name] = pod.Status.ContainerStatuses
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


siggy:
@grampelberg per https://github.com/linkerd/linkerd2/issues/2554#issuecomment-478653315, this PR may not do exactly what we want. it makes both dashboard and check pass if at least one pod of each component is up. i can modify things so check and dashboard behave differently. alternatively, i left a TODO in the code to indicate that ideally the check could pass but issue a warning. this would be a bigger change to how checks work, but i think it would be useful. lmkwyt.

unrelated, @harsh-98 i just saw on the ticket you had offered to work on this. my apologies for swooping in. i was on an airplane with no wifi, and only had visibility into the title of the issue, and not the contents.

grampelberg:
This PR's behavior is good enough for merge, I agree that it'd be nice to have a little different behavior though in the future.

@siggy I like where you're going with the TODO, are there any other checks that can benefit from that right now? I'm hesitant to prioritize the work to the top of the queue if we only have one use case right now.

siggy:
> @siggy I like where you're going with the TODO, are there any other checks that can benefit from that right now? I'm hesitant to prioritize the work to the top of the queue if we only have one use case right now.

surveying the other checks, nothing else jumps out that would immediately benefits from a dynamic warning<->fatal change.

pkg/healthcheck/healthcheck.go
```diff
@@ -873,16 +880,34 @@ func validateControlPlanePods(pods []corev1.Pod) error {
 	}
```
alpeb: While we're at it, I think we can move "proxy-injector" into the slice above now that it's a required component.

alpeb: (:point_up: this was in reference to the `names` slice in `validateControlPlanePods()`)

siggy: hmm, i intentionally left it as-is because i didn't want newer CLIs to hang when checking an older control-plane that doesn't have a `proxy-injector`. let me know what you think.

alpeb: That's a good point I didn't consider. Let's leave it as is for some more time then :+1:

pkg/healthcheck/healthcheck.go
```diff
@@ -873,16 +880,34 @@ func validateControlPlanePods(pods []corev1.Pod) error {
...
+			for _, container := range containers {
+				if !container.Ready {
+					// TODO: Save this as a warning, allow check to pass but let the user
+					// know there is at least one pod not ready. This might imply
+					// restructuring health checks to allow individual checks to return
+					// either fatal or warning, rather than setting this property at
+					// compile time.
+					err = fmt.Errorf("The \"%s\" pod's \"%s\" container is not ready", pod,
+						container.Name)
+					containersReady = false
```
ihcsim: TIOLI: we should be able to `break` here.

Issue #2709, opened 2019-04-17
Expose identity of edges in the dashboard


scottcarol:
### What problem are you trying to solve?

See #2708

### How should the problem be solved?

For dashboard users, exposing identity of edges in an "Identity" table. In the future, there could be a dedicated "Identity" view page, but initially this feature will be only a table.

### Any alternatives you've considered?

See #2708

### How would users interact with this feature?

An `Identity` table will appear in the `Overview` view of the dashboard for each namespace, and in every resource detail view. 

The table will contain these columns (as depicted above and in the CLI as `linkerd edges identity`):
    - SRC
    - DST
    - Client
    - Server
    - Message/Reason (We will experiment with the best way to show this for readability/usefulness. Exposing the entire reason may be hard to parse, but obscuring it in favor of a tooltip creates more work for the user.)

![grambp](https://user-images.githubusercontent.com/2289389/56306199-c41d8100-60f6-11e9-8369-a9785b3c667e.png)

This issue builds off ideas in #2605 and #2606.

scottcarol:
@grampelberg here’s what I have so far for the Edges table.

The default `Overview` view is by namespace, but requests to the `edges` API endpoint must specify one of 6 resource types:
- daemonsets
- deployments
- jobs
- pods
- replicationcontrollers
- statefulsets

In the below screenshot from the `Overview` view for the `emojivoto` namespace, I query for pod edges and include a column for resource type. Alternatively I could make multiple queries and pull in all available edges. 

Or, I could only have the Edges table only in the views for the 6 resource types specified above. What do you think?

![Screen Shot 2019-06-12 at 9 31 56 AM](https://user-images.githubusercontent.com/2289389/59370515-881e2b00-8cf7-11e9-912f-799246ef1064.png)

grampelberg:
Let's add the edges table to each of the resource detail views and only there for now. WDYT? 

What do you think about opening up two new issues? The overview page needs some love as it is really deployment based right now and should be more general. The resource list pages are also a little questionably valuable (though the namespace one is super useful and buried).

scottcarol:
I think adding the edges table to the resource detail views is a good plan. In that case I'll omit the column specifying resource name.

Re: two new issues, one issue is adding the edges table to each resource detail view, and one issue is rethinking the overview page to be more general/helpful? Yes I can do that.

[...]

PR #2770, opened 2019-04-29
Add --disable-identity flag to `linkerd install` and `linkerd upgrade`

alpeb:
In #2717 we implemented `--disable-identity` in `linkerd inject`, adding the annotation `config.linkerd.io/disable-identity: true` which made the proxy injector to not set up identity for the injected proxy.

This PR expands that by enabling `--disable-identity` in `linkerd install [config|control-plane]` and `linkerd upgrade [config|control-plane]`.

Using that flag has these effects:
- The `linkerd-identity` pod is not installed
- All injected pods (including the CP pods) won't use identity (they'll all have the annotation`linkerd.io/identity-mode: disabled`)

Re: second part of #2540

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -208,7 +208,6 @@ func newCmdInstall() *cobra.Command {
 	// the configuration in validateAndBuild.
 	flags := options.recordableFlagSet()
 	installOnlyFlags := options.installOnlyFlagSet()
-	disableIdentityFlag := options.disableIdentityFlagSet()
 
 	cmd := &cobra.Command{
 		Use:       fmt.Sprintf("install [%s|%s] [flags]", configStage, controlPlaneStage),
@@ -232,7 +231,7 @@ control-plane.`,
   linkerd install config -l linkerdtest | kubectl apply -f -
   linkerd install control-plane -l linkerdtest | kubectl apply -f -`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			stage, err := validateArgs(args, flags, installOnlyFlags, disableIdentityFlag)
+			stage, err := validateArgs(args, flags, installOnlyFlags)
 			if err != nil {
 				return err
 			}
@@ -245,7 +244,7 @@ control-plane.`,
 				exitIfNamespaceDoesNotExist()
 			}
 
-			values, configs, err := options.validateAndBuild(stage, flags, disableIdentityFlag)
+			values, configs, err := options.validateAndBuild(stage, flags)
 			if err != nil {
 				return err
 			}
@@ -258,16 +257,15 @@ control-plane.`,
 
 	// Some flags are not available during upgrade, etc.
 	cmd.PersistentFlags().AddFlagSet(installOnlyFlags)
-	cmd.PersistentFlags().AddFlagSet(disableIdentityFlag)
 
 	return cmd
 }
 
-func (options *installOptions) validateAndBuild(stage string, flags ...*pflag.FlagSet) (*installValues, *pb.All, error) {
+func (options *installOptions) validateAndBuild(stage string, flags *pflag.FlagSet) (*installValues, *pb.All, error) {
 	if err := options.validate(); err != nil {
 		return nil, nil, err
 	}
-	options.recordFlags(flags...)
+	options.recordFlags(flags)
 
 	var identityValues *installIdentityValues
 	if !options.disableIdentity {
@@ -335,6 +333,8 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.controlPlaneVersion, "control-plane-version", "", options.controlPlaneVersion, "(Development) Tag to be used for the control plane component images")
 	flags.MarkHidden("control-plane-version")
 
+	options.addDisableIdentityFlag(flags)
+
 	return flags
 }
 
@@ -372,17 +372,12 @@ func (options *installOptions) installOnlyFlagSet() *pflag.FlagSet {
 	return flags
 }
 
-func (options *installOptions) recordFlags(flags ...*pflag.FlagSet) {
-	if len(flags) == 0 {
+func (options *installOptions) recordFlags(flags *pflag.FlagSet) {
+	if flags == nil {
 		return
 	}
 
-	combinedFlags := pflag.NewFlagSet("combined-flags", pflag.ExitOnError)
-	for _, f := range flags {
-		combinedFlags.AddFlagSet(f)
-	}
-
-	combinedFlags.VisitAll(func(f *pflag.Flag) {
+	flags.VisitAll(func(f *pflag.Flag) {
 		if f.Changed {
 			switch f.Name {
 			case "ignore-cluster", "control-plane-version", "proxy-version":
@@ -931,8 +926,8 @@ func validateArgs(args []string, flags ...*pflag.FlagSet) (string, error) {
 		combinedFlags.VisitAll(func(f *pflag.Flag) {
 			if f.Changed {
 				switch f.Name {
-				case "ignore-cluster":
-					// allow `linkerd install config --ignore-cluster`
+				case "ignore-cluster", "disable-identity":
+					// flags available in `linkerd install config` and `linkerd upgrade config`
 				default:
 					invalidFlags = append(invalidFlags, f.Name)
 				}
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -56,6 +56,7 @@ type (
 		ControllerUID            int64
 		EnableH2Upgrade          bool
 		NoInitContainer          bool
+		DisableIdentity          bool
 
 		Configs configJSONs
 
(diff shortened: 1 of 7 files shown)


cli/cmd/install.go
```diff
@@ -207,6 +208,7 @@ func newCmdInstall() *cobra.Command {
 	// the configuration in validateAndBuild.
 	flags := options.recordableFlagSet()
 	installOnlyFlags := options.installOnlyFlagSet()
+	disableIdentityFlag := options.proxyConfigOptions.disableIdentityFlagSet()
```
siggy: you could just do `options.disableIdentityFlagSet()` (though i could see reasons for making the convention either way)

cli/cmd/install.go
```diff
@@ -542,7 +555,13 @@ func (values *installValues) render(w io.Writer, configs *pb.All) error {
 		files = append(files, []*chartutil.BufferedFile{
 			{Name: "templates/_resources.yaml"},
 			{Name: "templates/config.yaml"},
-			{Name: "templates/identity.yaml"},
+		}...)
+		if !values.DisableIdentity {
```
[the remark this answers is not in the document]
alpeb: Sounds good. I'll do the same in `identityd-rbac.yml`. But it should be enough this, no?
```
{{- if and (not .DisableIdentity) (.Identity) }}
```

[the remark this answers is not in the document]
siggy: I'm not sure about `.Identity.Issuer` vs. `.Identity`, and had err'd on the side of changing it. ;)

cli/cmd/install.go
alpeb: `options` is of type `*installOptions` in `install.go`.
`disableIdentityFlagSet()` is also called from `inject.go`:
https://github.com/linkerd/linkerd2/blob/ac6bd5eb6a2ca1e1539a949e53a8a8b0b507e830/cli/cmd/inject.go#L47
where `options` is of type `*proxyConfigOptions`. So I'm forced to use the most narrow common type.

siggy: i may be misunderstanding, since `installOptions` has an in-lined `*proxyConfigOptions` member, you should be able to just call `options.disableIdentityFlagSet()` as-is, without specifying `proxyConfigOptions`?

alpeb: You're actually right :grin:

ihcsim: I'd think we want this flag to be recorded. Otherwise, during upgrade, [identity will be re-introduced](https://github.com/linkerd/linkerd2/blob/85a7c885d47f17635b8cc53ee97f724fd5dedbeb/cli/cmd/upgrade.go#L175-L176). Recording this flag will allow us to distinguish what `if idctx == nil` mean during upgrade, i.e. upgrading from a version that didn't support identity vs. identity is disabled.

ihcsim: Looks like this flag is already recorded:
```
   install: |
     {"uuid":"8dcc640d-76b4-4ef9-953b-977a042b15b8","cliVersion":"git-f12d824c","flags":[{"name":"disable-identity","value":"true"}]}
```

I wonder if we should add it to the recordable flagset? Then it will also be available for upgrade. That way, a user will be able to enable identity during upgrade.

alpeb: I haven't really explored the option of enabling identity during an upgrade. All the `identity-issuer-*` and `identity-trust-*` flags are disabled in `linkerd upgrade`, I guess because it's presumed that was already set during `linkerd install`. This also entails the question whether `--disable-identity=true` issued during an upgrade would cause problems. If so we could only allow enabling identity during upgrades and not disabling. But that's something I'd rather do in a separate PR.

ihcsim:
Shall we update the dashboard to not show the identity deployment, when identity is disabled?

Screenshot_20190430_094852

cli/cmd/install.go
ihcsim: > All the identity-issuer-* and identity-trust-* flags are disabled in linkerd upgrade

Right, because during upgrade, those identity assets shouldn't change. Otherwise, the existing data plane will be broken. Upgrade will either generate new ones (if none exist), or re-use existing ones.

scottcarol:
@ihcsim That's a good question. Because identity is enabled by default, I think we should make sure the user knows that they have disabled it b/c this is unusual (and possibly temporary) behavior. 

What about a text bubble that says "disabled" that would appear next to the Identity deployment in that table? Below is what the "meshed" text bubble looks like, perhaps something like this to the right of "Identity"?

Screen Shot 2019-04-30 at 10 27 11 AM

ihcsim:
@scottcarol Sure, what you suggested makes sense. Shall we include this change as part of #2709? Also, not sure how/if this will affect #2709, keep in mind that `linkerd install --disable-identity` essentially, disables TLS identity entirely, while `linkerd inject --disable-identity` will disable identity for that particular workload.

scottcarol:
@ihcsim I think it would make sense to live in #2709. @grampelberg what are your thoughts on UX and which GitHub issue it should be part of?

Re: changes in `install` vs. `inject`... when you say "workload" do you mean identity will be disabled for the specific resource? What does "workload" mean in this case? Thanks and I will most likely be pinging you about it when I get to that part of the project.

ihcsim:
> when you say "workload" do you mean identity will be disabled for the specific resource? What does "workload" mean in this case? 

Yes, "workload" is one of the [k8s resource categories](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/), referring to things like deployments, pods, statefulsets etc.

scottcarol:
@ihcsim Thanks -- good to know.

alpeb:
My last commit adds support for `--disable-identity` into `linkerd upgrade [config|control-plane]`.

After having installed with `linkerd install --disable-identity` the identity pod is not installed and the control plane pods have no identity set up.
Afterwards, upgrades behave so:
- `linkerd upgrade`: I confirmed identity is not set during upgrade
- `linkerd upgrade --disable-identity=false`: The identity pod is installed and the other control plane pods are reinstalled with identity enabled.

Note that running `linkerd upgrade --disable-identity` on a control plane that was installed **with** identity enabled throws an error, to avoid disabling identity on a cluster with injected pods that are already relying on it.

@olix0r you expressed some concerns in this [comment](https://github.com/linkerd/linkerd2/pull/2717#discussion_r276452360). Does this approach make sense to you?

olix0r:
I still think there's a Simplicity argument against supporting `upgrade --disable-identity` at all... It's a big enough change to the operation mode that it doesn't seem like something we should necessarily target for graceful upgrade/downgrade; and it doesn't seem particularly valuable to users (we want to really really really encourage this feature to be left on).

I should also note that I suspect as a part of the tap-hardening work, tap will also become contingent on identity, and so we'll have to go through a similar dance with tap as well.

But, if folks really want this, this sounds like a reasonable set of behavior.

ihcsim:
@alpeb and I had a chat earlier regarding whether it's worth doing `upgrade --disable-identity`. The reason why it came up is because upgrade always (re-)enables identity, despite of previous `install --disable-identity`. If this is acceptable, we don't need `upgrade --disable-identity`. But I assume that if a user has reasons to use `install --disable-identity`, they won't want an upgrade to automatically re-enable identity.

I prefer that we **don't** support `--disable identity` for both `install` and `upgrade`, as it makes things more complicated, and AFAIK, we haven't had any users seriously asking for it.

@grampelberg Any thoughts on supporting `install --disable-identity`/`upgrade --disable-identity` or not?

grampelberg:
@scottcarol I would like to keep identity from being a chip on the main page. That's partially why we removed the %TLS column and moved it into the edges pages as part of #2709. Basically, I want to discourage disabling identity as much as possible. I'd support adding a check to `linkerd check --proxy` that warns when you've disabled identity though. That feels like a closer match to intent (auditing vs viewing metrics). Happy to chat more if I'm not making any sense =)

@ihcsim @alpeb is there any reason we should allow users to disable identity as a supported configuration for the control plane itself? I'm getting the feeling that it is not something we want to support moving forward. I've chatted with some users who want to disable identity in their workloads. So far the lack of service account tokens in those workloads is the best reasoning I've heard.

scottcarol:
@grampelberg My comment about the UX was related to @ihcsim's screenshot [higher on the page](https://github.com/linkerd/linkerd2/pull/2770#issuecomment-488037477) - my understanding is that we've talked about what the user will see when identity is enabled, but not when it's disabled. In the scenario Ivan screenshotted, should we communicate anything special to the user to signify that they've disabled identity?

grampelberg:
@scottcarol we probably should (though see my reservations about making it a supported configuration at all). It feels like a better place to surface this is during `check --proxy` as a warning.

alpeb:
Based on offline discussion, we decided to not support disabling identity globally through `linkerd install` nor `linkerd upgrade` for the time being, as it could generate quite some extra complexity, just to cater a theoretical requirement that might not really be needed by end-users.

user-name-is-taken:
This feature would have been useful for me. I'm sitting behind a corporate firewall that requires all tls traffic to be intercepted by the company's MITM cert. Doing this via a flag seems a lot less error prone from my point of view than injecting an attribute.

Issue #2712, opened 2019-04-17
Move tap service out of the controller pod


adleong:
In anticipation of tap becoming an APIExtension so that requests to it may be authenticated and authorized, we should split the tap service out of the controller pod.

* The tap pod should serve the TapByResource route of the public API (protobuf over HTTP)
* The public-api service should return a message on the TapByResource route indicating that this has moved.
* The linkerd tap CLI command should make requests to the linkerd-tap service instead of the public-api
* The web process should use a protobuf over HTTP client to make calls to the tap service.

dadjeibaah:
As of #2773, Tap is now in its on pod, however, it is not configured to serve tap events using protobuf over HTTP. That work is being done in this [diff](https://github.com/linkerd/linkerd2/compare/dad/tap-pod-2)

kleimkuhler:
@dadjeibaah Are you still making progress with the branch you linked to above? I'd like to get a better idea of where things stand with respect to the tap service being ready to serve tap events, and what still needs to be wrapped up!

dadjeibaah:
Thanks for checking in @kleimkuhler! The branch I linked above is a rough PR that has tap set up to use the k8s `apiservice`. It also has the CLI stream tap events from the new `apiservice`. Although this needs to be tested thoroughly. The next piece that needs to be implemented is to fix tap in the UI dashboard.

kleimkuhler:
Closed in [tap hardening project](https://github.com/orgs/linkerd/projects/8)

Issue #1930, opened 2018-12-05
Add SecurityContext to the various containers installed, except the initContainer


codeman9:
### What problem are you trying to solve?

Our Pod Security Policy explicitly denies containers running as root. Since the Dockerfiles for the various containers needed during linkerd install do not specify a user, it is assumed that they will run as root. When I try to install linkerd into our cluster, I get an error that says that containers cannot runAsRoot. The `initContainer` that needs root and `NET_ADMIN` will still be an issue for us until #1887 is resolved.

### How should the problem be solved?

Adding a securityContext with a runAsUser value to the various containers is one way to solve this problem. If any of the required containers do actually need to run as root, then I'm not sure how to get around that. From what I can tell this change doesn't affect the installation.

### Any alternatives you've considered?

I'm not super familiar with alternatives other than updating the Dockerfiles to run as a particular user.

### How would users interact with this feature?

Users will not be affected by this change.

codeman9:
#1929

klingerf:
Fixed by #1929.

PR #2773, opened 2019-04-30
Move tap service into its own pod.

dadjeibaah:
This PR moves the tap service to its own pod in the control plane in order to lay the foundation to authorize tap requests using Kubernetes' APIExtension API. This PR makes tap work the same way it already does currently, however, this PR adds a new HTTPS server in the tap service that will eventually be used to authorize tap requests. 

In a follow up PR, the new HTTPS server will eventually set up a Tap CRD with the server registering itself to allow it to work with Kubernetes aggregation layer.

Relates #2712

### DIFF
--- a/chart/templates/tap-rbac.yaml
+++ b/chart/templates/tap-rbac.yaml
@@ -1,0 +1,40 @@
+{{with .Values -}}
+---
+###
+### Tap RBAC
+###
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: linkerd-{{.Namespace}}-tap
+rules:
+- apiGroups: [""]
+  resources: ["pods", "services", "replicationcontrollers", "namespaces"]
+  verbs: ["list", "get", "watch"]
+- apiGroups: ["extensions", "apps"]
+  resources: ["daemonsets", "deployments", "replicasets", "statefulsets"]
+  verbs: ["list", "get", "watch"]
+- apiGroups: ["extensions", "batch"]
+  resources: ["jobs"]
+  verbs: ["list" , "get", "watch"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: linkerd-{{.Namespace}}-tap
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: linkerd-{{.Namespace}}-controller
+subjects:
+- kind: ServiceAccount
+  name: linkerd-tap
+  namespace: {{.Namespace}}
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-tap
+  namespace: {{.Namespace}}
+{{end -}}
--- a/chart/templates/tap.yaml
+++ b/chart/templates/tap.yaml
@@ -1,0 +1,72 @@
+{{with .Values -}}
+---
+###
+### Tap
+###
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: linkerd-tap
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: tap
+  annotations:
+    {{.CreatedByAnnotation}}: {{.CliVersion}}
+spec:
+  type: ClusterIP
+  selector:
+    {{.ControllerComponentLabel}}: tap
+  ports:
+  - name: grpc
+    port: 8088
+    targetPort: 8088
+---
+kind: Deployment
+apiVersion: extensions/v1beta1
+metadata:
+  name: linkerd-tap
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: tap
+  annotations:
+    {{.CreatedByAnnotation}}: {{.CliVersion}}
+spec:
+  replicas: {{.ControllerReplicas}}
+  template:
+    metadata:
+      labels:
+        {{.ControllerComponentLabel}}: tap
+      annotations:
+        {{.CreatedByAnnotation}}: {{.CliVersion}}
+    spec:
+      serviceAccountName: linkerd-tap
+      containers:
+      - name: tap
+        ports:
+        - name: grpc
+          containerPort: 8088
+        - name: admin-http
+          containerPort: 9998
+        image: {{.ControllerImage}}
+        imagePullPolicy: {{.ImagePullPolicy}}
+        args:
+        - "tap"
+        - "-controller-namespace={{.Namespace}}"
+        - "-log-level={{.ControllerLogLevel}}"
+        livenessProbe:
+          httpGet:
+            path: /ping
+            port: 9998
+          initialDelaySeconds: 10
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 9998
+          failureThreshold: 7
+        {{ with .TapResources -}}
+        {{- template "resources" . }}
+        {{ end -}}
+        securityContext:
+          runAsUser: {{.ControllerUID}}
+{{end -}}
--- a/cli/cmd/testdata/install-sp_default.golden
+++ b/cli/cmd/testdata/install-sp_default.golden
@@ -122,3 +122,15 @@ spec:
       method: GET
       pathRegex: /public/img/.*
 ---
+apiVersion: linkerd.io/v1alpha1
+kind: ServiceProfile
+metadata:
+  name: linkerd-tap.linkerd.svc.cluster.local
+  namespace: linkerd
+spec:
+  routes:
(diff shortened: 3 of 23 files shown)


dadjeibaah:
@siggy thanks for the insight on this change, you bring up a good point on making sure this still works on upgrade. I would have to think a little bit in making sure that there is some fallback in the follow-up PR for the CLI when we try to tap resources.

In the meantime, I think this PR is safe to merge once I address all your comments and the failing tests.

controller/tap/server.go
```diff
@@ -426,18 +432,53 @@ func (s *server) translateEvent(orig *proxy.TapEvent) *public.TapEvent {
 	return ev
 }
 
+func newHTTPServer(k8sAPI *k8s.API, controllerNamespace string) *http.Server {
+	var (
+		aggregationLayerConfigMapName = "extension-apiserver-authentication"
+	)
+
+	cm, err := k8sAPI.Client.CoreV1().
```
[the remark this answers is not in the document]
dadjeibaah: I tried to include that but it didn't seem to work. I think that its because k8sAPI is not a `kubernetes.Interface` which is an interface we create. It looks like this file only has `*k8s.API` available in `NewServer`.

controller/tap/server.go
```diff
@@ -426,18 +432,53 @@ func (s *server) translateEvent(orig *proxy.TapEvent) *public.TapEvent {
...
+		ConfigMaps("kube-system").
+		Get(aggregationLayerConfigMapName, metav1.GetOptions{})
+
+	if err != nil {
+		log.Fatalf("Failed to load [%s] config: %s", aggregationLayerConfigMapName, err)
```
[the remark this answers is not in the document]
dadjeibaah: Good point. I think if we can't create this server, for now, we want to fail silently. When the second PR comes along, this would mean that there may be an issue with the cluster and we want the pod to restart. I will include tests to this to capture the initial behavior.

cli/cmd/testdata/install-sp_default.golden
```diff
@@ -122,3 +122,15 @@ spec:
       method: GET
       pathRegex: /public/img/.*
 ---
+apiVersion: linkerd.io/v1alpha1
+kind: ServiceProfile
+metadata:
+  name: linkerd-tap-api.linkerd.svc.cluster.local
```
olix0r: this should be `linkerd-tap`, no?

cli/installsp/template.go
```diff
@@ -124,4 +124,16 @@ spec:
     condition:
       method: GET
       pathRegex: /public/img/.*
+---
+apiVersion: linkerd.io/v1alpha1
+kind: ServiceProfile
+metadata:
+  name: linkerd-tap-api.{{.Namespace}}.svc.cluster.local
```
olix0r: errr, THIS should be `linkerd-tap`, shouldn't it?

cli/cmd/testdata/install-sp_default.golden
dadjeibaah: Ahh forgot to change that, good catch!

controller/cmd/tap/main.go
```diff
@@ -52,6 +52,11 @@ func main() {
 		server.Serve(lis)
 	}()
 
+	go func() {
+		log.Println("starting HTTPS server on :8081")
```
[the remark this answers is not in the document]
olix0r: I think I'd prefer that the TLS server change be made in a separate PR, if possible. But, whenever we do that, can we make this port configurable via the CLI?

controller/tap/server.go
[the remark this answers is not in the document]
siggy: Ah my mistake! This is an `API` from `controller/k8s`, I thought it was a `KubernetesAPI` from `pkg/k8s`. Perhaps this is a hint to myself that we could consolidate these at some point.

test/install_test.go
```diff
@@ -18,9 +18,9 @@ type deploySpec struct {
 	containers []string
 }
 
-//////////////////////
-///   TEST SETUP   ///
-//////////////////////
+// ////////////////////
```
siggy: intentional?

test/install_test.go
```diff
@@ -59,26 +61,26 @@ var (
...
 
 		// prometheus scrape failures of control-plane
 		`.* linkerd-prometheus-.*-.* linkerd-proxy ERR! \[ +\d+.\d+s\] proxy={server=out listen=127\.0\.0\.1:4140 remote=.*} linkerd2_proxy::proxy::http::router service error: an error occurred trying to connect: .*`,
 	}, "|"))
 )
 
-//////////////////////
-/// TEST EXECUTION ///
-//////////////////////
+// ////////////////////
```
siggy: same?

test/install_test.go
olix0r: TIOLI: I _much_ prefer simplifying this so it's less brittle.

Something like
```
//
// Test Setup
//
```

Should create just as much visual spacing, but is easier to change, reformatting, etc.

dadjeibaah: Definitely not intentional. It must have been my editor being overly helpful. I'll revert this.

chart/templates/tap-rbac.yaml
```diff
@@ -0,0 +1,40 @@
...
+### Tap RBAC
+###
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: linkerd-{{.Namespace}}-tap
+rules:
+- apiGroups: [""]
+  resources: ["pods", "endpoints", "services", "replicationcontrollers", "namespaces"]
```
klingerf: I'm pretty sure the tap server doesn't deal with the `endpoints` API, so you can remove it from this list.

chart/templates/tap.yaml
```diff
@@ -0,0 +1,76 @@
...
+          failureThreshold: 7
+        {{ with .TapResources -}}
+        {{- template "resources" . }}
+        {{ end -}}
+        securityContext:
+          runAsUser: {{.ControllerUID}}
+      volumes:
+      - name: config
+        configMap:
+          name: linkerd-config
```
klingerf: You can safely remove the `volumes` field in this spec, since you're not mounting this volume via a `volumeMount` field in the tap container spec.

chart/templates/tap-rbac.yaml
dadjeibaah: Thanks! wasn't quite sure how I could determine which resources were used by each control plane component, but I think the code below is a good indicator. https://github.com/linkerd/linkerd2/blob/f758a9e4282b2e181fcd06ecb20a44f215ae7e91/controller/cmd/tap/main.go#L27

klingerf: Yep, that's how I'd determine it

chart/templates/tap.yaml
```diff
@@ -0,0 +1,72 @@
...
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 9998
+          failureThreshold: 7
+        {{ with .TapResources -}}
+        {{- template "resources" . }}
+        {{ end -}}
+        securityContext:
+          runAsUser: {{.ControllerUID}}
```
adleong: Just curious: why do we need to run the controller as a specific UID and does that still apply to tap running in its own pod?

test/stat/stat_test.go
```diff
@@ -78,11 +78,12 @@ func TestCliStatForLinkerdNamespace(t *testing.T) {
 				"linkerd-prometheus":     "1/1",
 				"linkerd-proxy-injector": "1/1",
 				"linkerd-sp-validator":   "1/1",
+				"linkerd-tap":            "1/1",
 				"linkerd-web":            "1/1",
 			},
 		},
 		{
-			args: []string{"stat", "po", "-n", TestHelper.GetLinkerdNamespace(), "--from", "deploy/linkerd-controller"},
+			args: []string{"stat", "po", "-n", TestHelper.GetLinkerdNamespace(), prometheusPod, "--from", "po/" + controllerPod},
```
adleong: I'm not sure what's going on here, why is the prometheus pod added to this command?

chart/templates/tap.yaml
dadjeibaah: Good question. This was merely a result of a copy/pasta but I dug through git history to see why we added this change and the reason for it was for this issue https://github.com/linkerd/linkerd2/issues/1930. Reading through it, I think we still would like the tap pod to run with this UID since it technically is a controller component.

test/stat/stat_test.go
dadjeibaah: This test was created to test that the controller was sending traffic to Prometheus initially but now, with the Linkerd tap pod in the mix, this test fails because the controller sends traffic to both Prometheus and linkerd tap. The traffic to the Linkerd tap pod is dependent on whether other tests were exercising the Tap API, which made this test brittle because we would sometimes see the stat command display stats for the tap pod and other times it wouldn't

PR #2781, opened 2019-05-02
Fix Service Profile CRD check to handle k8s errors

### LINKED ISSUE #2780: Control plane k8s client initialization fails on ServerResources request
## Summary

User reported this error on public-api startup:
```
Failed to initialize K8s API: unable to retrieve the complete list of server APIs: external.metrics.k8s.io/v1beta1: the server could not find the requested resource
```

Source of error is this `Discovery().ServerResources()` request in `ServiceProfilesAccess()`:
https://github.com/linkerd/linkerd2/blob/e5d2460792f04aa0c6a8206db836ce38bbd3f11b/pkg/k8s/authz.go#L56-L59
...which calls into client-go:
https://github.com/kubernetes/client-go/blob/8d9ed539ba3134352c586810e749e58df4e94e4f/discovery/discovery_client.go#L252

## Proposal

The intent of `ServiceProfilesAccess()` is to validate that the ServiceProfile CRD is installed and accessible by the control plane. When a failure occurs in `Discovery().ServerResources()` for individual GroupVersion resources, it still returns a valid `[]*metav1.APIResourceList` for the GroupVersions that succeeded, along with a `ErrGroupDiscoveryFailed` error for the GroupVersions that failed.

Modify `ServiceProfilesAccess()` to gracefully handle `ErrGroupDiscoveryFailed`.

siggy:
`ServiceProfilesAccess()`, called by control plane components at
startup, would fail if it encountered an `ErrGroupDiscoveryFailed` from
a GroupVersion request. This error is mostly innocuous, as it returns an
error if any GroupVersion fails. `ServiceProfilesAccess()` only needs to
validate ServiceProfiles are available.

Modify `ServiceProfilesAccess()` to specifically request the
ServiceProfile GroupVersion. Also add Discovery object
(`APIResourceList`) support to `NewFakeClientSets`.

Fixes #2780

### DIFF
--- a/pkg/k8s/authz_test.go
+++ b/pkg/k8s/authz_test.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/k8s/authz_test.go
```diff
@@ -58,3 +59,37 @@ subjects:
...
+	if err != nil {
+		t.Fatalf("NewFakeAPI error: %s", err)
+	}
+
+	err = ServiceProfilesAccess(api)
+	// RBAC SSAR request failed, but the Discovery lookup succeeded
+	if !reflect.DeepEqual(err, errors.New("not authorized to access serviceprofiles.linkerd.io")) {
+		t.Fatalf("unexpected error: %s", err)
+	}
+}
```
alpeb: For my own understanding, how would one manage testing a positive result? I guess we'd have to add into `fakeResources` the RBAC to access serviceprofiles like we have in `controller-rbac.yaml`, but then how's the user of `NewFakeAPI` bound to those privileges?

siggy: Right, that's the general idea. I haven't yet determined which mock resources we need enable a successful `SelfSubjectAccessReview` request.

This applies to every test in this file. I referenced the issue in a TODO when I introduced `authz_test.go`:
https://github.com/linkerd/linkerd2/blob/00a94be07375eb24ef9a15594fd9bc3518a221b1/pkg/k8s/authz_test.go#L15

PR #2786, opened 2019-05-03
Refactor destination service

adleong:
This is a major refactor of the destination service.  The goals of this refactor are to simplify the code for improved maintainability.  In particular:

* Remove the "resolver" interfaces.  These were a holdover from when our decision tree was more complex about how to handle different kinds of authorities.  The current implementation only accepts fully qualified kubernetes service names and thus this was an unnecessary level of indirection.
* Moved the endpoints and profile watchers into their own package for a more clear separation of concerns.  These watchers deal only in Kubernetes primitives and are agnostic to how they are used.  This allows a cleaner layering when we use them from our gRPC service.
* Renamed the "listener" types to "translator" to make it more clear that the function of these structs is to translate kubernetes updates from the watcher to gRPC messages.

TODO before merging:
* re-add the discovery API
* fix lints
* add comments

TODO in followup PR:
* Add endpoint weights from a pod annotation


controller/api/destination/endpoint_translator.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+	}
+
+	et.log.Debugf("Sending destination no endpoints: %+v", u)
+	if err := et.stream.Send(u); err != nil {
+		et.log.Errorf("Failed to send address update: %s", err)
+	}
+}
+
+func (et *endpointTranslator) toAddr(address watcher.Address) (*net.TcpAddress, error) {
```
siggy: this method doesn't need to hang off `endpointTranslator`. is it for consistency with `toWeightedAddr`?

controller/api/destination/profile_translator.go
```diff
@@ -0,0 +1,35 @@
...
+		pt.stream.Send(&profiles.DefaultServiceProfile)
+		return
+	}
+	destinationProfile, err := profiles.ToServiceProfile(profile)
+	if err != nil {
+		pt.log.Error(err)
+		return
+	}
+	pt.log.Debugf("Sending profile update: %+v", destinationProfile)
+	pt.stream.Send(destinationProfile)
```
siggy: `stream.Send()` returns an error, should `Update()` pass it up? or at least log it?

olix0r: When would this error?

AFAIU, this would only error when a client has disconnected, so logging seems inappropriate.

controller/cmd/destination/main.go
```diff
@@ -20,7 +20,7 @@ func main() {
 	addr := flag.String("addr", ":8086", "address to serve on")
 	metricsAddr := flag.String("metrics-addr", ":9996", "address to serve scrapable metrics on")
 	kubeConfigPath := flag.String("kubeconfig", "", "path to kube config")
-	k8sDNSZone := flag.String("kubernetes-dns-zone", "", "The DNS suffix for the local Kubernetes zone.")
+	_ = flag.String("kubernetes-dns-zone", "", "The DNS suffix for the local Kubernetes zone.")
```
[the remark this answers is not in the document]
adleong: yup, thanks for the reminder

controller/api/destination/server.go
```diff
@@ -38,153 +43,180 @@ type server struct {
...
+		service,
+		stream,
+		log,
+	)
+
+	s.endpoints.Subscribe(service, port, translator)
+	defer s.endpoints.Unsubscribe(service, port, translator)
+
+	select {
+	case <-s.shutdown:
```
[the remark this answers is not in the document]
adleong: good question.  returning nil will gracefully complete the stream

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,480 @@
...
+	// listeners may be subscribed to a servicePublisher.  servicePublisher maintains the
+	// current state of the address set and publishes diffs to all listeners when
+	// updates come from either the endpoints API or the service API.
+	servicePublisher struct {
+		id     ServiceID
+		log    *logging.Entry
+		k8sAPI *k8s.API
+
+		ports map[Port]*portPublisher
+		// All access to the servicePublisher and its portPublishers is explicitly synchronized by
```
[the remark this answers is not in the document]
adleong: I went with one mutex for an entire servicePublisher which guards all operations on its portPublishers.  Most updates to a servicePublisher will affect all of its portPublishers (e.g. an endpoints update) so it seemed simpler to lock them all at once.  This is contrast to how each servicePublisher has a distinct mutex since updates coming into the endpointsWatcher will usually only affect a single servicePublisher.

controller/api/destination/watcher/profile_watcher.go
```diff
@@ -0,0 +1,182 @@
...
+		profiles      map[ProfileID]*profilePublisher
+		profilesMu    sync.RWMutex // This mutex protects modifcation of the map itself.
+
+		log *logging.Entry
+	}
+
+	profilePublisher struct {
+		profile   *sp.ServiceProfile
+		listeners []ProfileUpdateListener
+		// All access to the profilePublisher is explicitly synchronized by this mutex.
```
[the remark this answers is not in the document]
adleong: ProfileUpdateListener is an interface, implemented by profileTranslator.  I don't think it requires any synchronization

controller/api/destination/endpoint_translator.go
adleong: that's right.  these were originally one method which I split into two and it seemed reasonable to keep them together by hanging both off of endpointTranslator

controller/api/destination/server.go
```diff
@@ -38,153 +43,180 @@ type server struct {
...
-		}
+	return nil
+}
 
-		s.log.Debugf("ServicePorts[%s]: %+v", serviceID, discoverySP)
-		rsp.ServicePorts[serviceID.String()] = &discoverySP
+func nsFromToken(ctx string) string {
+	// ns:<namespace>
+	parts := strings.Split(ctx, ":")
+	if len(parts) == 2 && parts[0] == "ns" {
```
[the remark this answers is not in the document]
adleong: I'm not sure, we may be considering the token as optional which means that "" is a valid value.

controller/api/destination/server.go
```diff
@@ -193,30 +225,30 @@ func getHostAndPort(dest *pb.GetDestination) (string, int, error) {
 		var err error
 		port, err = strconv.Atoi(hostPort[1])
 		if err != nil {
-			err = fmt.Errorf("Invalid port %s", hostPort[1])
-			log.Error(err)
-			return "", 0, err
+			return "", 0, fmt.Errorf("Invalid port %s", hostPort[1])
```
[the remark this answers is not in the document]
adleong: :+1: will add a log upstream in Get and GetProfile where this error is returned to

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,480 @@
...
+	if !ok {
+		return
+	}
+	sp.unsubscribe(port, listener)
+
+	return
+}
+
+func (ew *EndpointsWatcher) addService(obj interface{}) {
+	service := obj.(*corev1.Service)
```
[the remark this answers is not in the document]
adleong: to be extra defensive, yes; but I guess we're just trusting that k8s isn't going to send us objects of the wrong type.  this is how it is on master 🤷‍♂

controller/api/destination/endpoint_translator.go
```diff
@@ -0,0 +1,171 @@
...
+func newEndpointTranslator(
+	controllerNS string,
+	identityTrustDomain string,
+	enableH2Upgrade bool,
+	service watcher.ServiceID,
+	stream pb.Destination_GetServer,
+	log *logging.Entry,
+) *endpointTranslator {
+	log = log.WithFields(logging.Fields{
+		"component": "endpoint-converter",
```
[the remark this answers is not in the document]
adleong: good catch

controller/api/destination/watcher/endpoints_watcher.go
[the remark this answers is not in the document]
olix0r: It's a programmer error, not a user/input/state error to get the wrong type here, so it should definitely panic/fail hard and not be handled gracefully.

controller/api/destination/server.go
```diff
@@ -7,23 +7,28 @@ import (
...
 
 	pb "github.com/linkerd/linkerd2-proxy-api/go/destination"
-	"github.com/linkerd/linkerd2/controller/api/util"
+	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
 	discoveryPb "github.com/linkerd/linkerd2/controller/gen/controller/discovery"
 	"github.com/linkerd/linkerd2/controller/k8s"
-	"github.com/linkerd/linkerd2/pkg/addr"
 	"github.com/linkerd/linkerd2/pkg/prometheus"
-	log "github.com/sirupsen/logrus"
+	logging "github.com/sirupsen/logrus"
```
olix0r: we call this import `log` in every use. I don't think it should be different in this module. If you feel strongly about the `logging` change, I'd prefer we make that its own change on the whole project.

controller/api/destination/server.go
```diff
@@ -7,23 +7,28 @@ import (
...
-	log *log.Entry
-}
+type (
+	server struct {
+		endpoints *watcher.EndpointsWatcher
+		profiles  *watcher.ProfileWatcher
+
+		enableH2Upgrade     bool
+		controllerNS        string
+		identityTrustDomain string
```
olix0r: All of these things are endpoints-specific, right? Can this configuration be pushed into the EndpointsWatcher?

controller/api/destination/server.go
```diff
@@ -38,153 +43,180 @@ type server struct {
...
-			}
-
-			discoverySP.PortEndpoints[port] = &podAddrs
-		}
+	return nil
+}
 
-		s.log.Debugf("ServicePorts[%s]: %+v", serviceID, discoverySP)
-		rsp.ServicePorts[serviceID.String()] = &discoverySP
+func nsFromToken(ctx string) string {
```
olix0r: This logic is profiles-specific. Can this be moved behind the ProfilesWatcher interface?

controller/api/destination/server.go
```diff
@@ -193,30 +225,30 @@ func getHostAndPort(dest *pb.GetDestination) (string, int, error) {
...
-) (streamingDestinationResolver, error) {
-	k8sDNSZoneLabels := []string{}
-	if k8sDNSZone != "" {
-		var err error
-		k8sDNSZoneLabels, err = splitDNSName(k8sDNSZone)
-		if err != nil {
-			return nil, err
-		}
+func (s *server) getServiceAndPort(dest *pb.GetDestination) (watcher.ServiceID, watcher.Port, error) {
+	if dest.Scheme != "k8s" {
```
olix0r: Let's please not perpetuate this logic. I'd suggest that we just ignore scheme checks until we have a reason the check them.

controller/api/destination/server.go
```diff
@@ -193,30 +225,30 @@ func getHostAndPort(dest *pb.GetDestination) (string, int, error) {
...
-	log.Infof("Built k8s name resolver")
-
-	return k8sResolver, nil
+	host, port, err := s.getHostAndPort(dest)
+	if err != nil {
+		return watcher.ServiceID{}, 0, err
+	}
+	domains := strings.Split(host, ".")
+	// S.N.svc.cluster.local
+	if len(domains) != 5 {
```
olix0r: Shouldn't you actually check the suffix?

controller/api/destination/server.go
adleong: I did this so that I can have variables named `log`.  What do you call `log` if you can't call it `log`?

controller/api/destination/server.go
adleong: no, they're configured on the destination server by the flags:
* `enable-h2-upgrade`
* `controller-namespace`
* and the global identity context

controller/api/destination/server.go
olix0r: idiomatic go would probably call them `l` or something...  in general i've noticed this pattern _especially_ when the field/var name would just be the type/module name. But i'd prefer to do whatever we do elsewhere in the codebase.

adleong: `logrus.Entry` is not used elsewhere in the project, except for one usage in `cni-plugin` where the logrus import is left as `logrus`

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,488 @@
...
+	if pp.exists {
+		if len(pp.pods) > 0 {
+			listener.Add(pp.pods)
+		} else {
+			listener.NoEndpoints(true)
+		}
+	} else {
+		listener.NoEndpoints(false)
+	}
+	pp.listeners = append(pp.listeners, listener)
```
siggy: related to comment on previous review around mutexes (https://github.com/linkerd/linkerd2/pull/2786#discussion_r280855725), if we're guarding access to `portPublisher`'s from a higher-level struct, mind adding comments around which `portPublisher` methods are not thread safe? i'm concerned about future changes that may call these methods and be unaware of that context (for example, when grabbing state for a debug endpoints request).

siggy:
@adleong the most recent integration test failure was fixed via #2797, merging master should get you past this.

controller/api/destination/endpoint_translator_test.go
```diff
@@ -0,0 +1,272 @@
...
+			"serviceaccount":        "serviceaccount-name",
+			"control_plane_ns":      "linkerd",
+		}
+		if !reflect.DeepEqual(actualAddedAddress1MetricLabels, expectedAddedAddress1MetricLabels) {
+			t.Fatalf("Expected global metric labels sent to be [%v] but was [%v]", expectedAddedAddress1MetricLabels, actualAddedAddress1MetricLabels)
+		}
+	})
+
+	t.Run("Sends TlsIdentity when enabled", func(t *testing.T) {
+		expectedTLSIdentity := &pb.TlsIdentity_DnsLikeIdentity{
```
scottcarol: TIOLI, I'm curious about the different capitalization of "TlsIdentity" vs "TLSIdentity" which is a little confusing when reading through. Is this intentional?

controller/api/destination/profile_translator.go
```diff
@@ -0,0 +1,323 @@
...
+				"failed to parse duration for route '%s' in service profile '%s' in namespace '%s': %s",
+				route.Name,
+				profile.Name,
+				profile.Namespace,
+				err,
+			)
+			timeout = defaultRouteTimeout
+		}
+	}
+	ret := pb.Route{
```
scottcarol: instead of creating `ret` could you return a `&pb.Route` directly as you did in line 87?

```
	return &pb.Route{
		Condition:       cond,
		ResponseClasses: rcs,
		MetricsLabels:   map[string]string{"route": route.Name},
		IsRetryable:     route.IsRetryable,
		Timeout:         toDuration(timeout),
	}, nil
```

controller/api/destination/server.go
```diff
@@ -38,185 +41,141 @@ type server struct {
...
-
-	log.Infof("Built k8s name resolver")
+	return nil
+}
 
-	return k8sResolver, nil
+func (s *server) Endpoints(ctx context.Context, params *discoveryPb.EndpointsParams) (*discoveryPb.EndpointsResponse, error) {
+	s.log.Debugf("serving endpoints request")
+
+	// servicePorts := e.getState()
```
scottcarol: Did you mean to leave the commented-out section in this file?

controller/api/destination/server.go
```diff
@@ -38,185 +41,141 @@ type server struct {
...
-	listener := newEndpointListener(stream, s.k8sAPI.GetOwnerKindAndName, s.enableH2Upgrade, s.controllerNS, s.identityTrustDomain)
+	primary, secondary := newFallbackProfileListener(translator)
 
-	resolverCanResolve, err := s.resolver.canResolve(host, port)
-	if err != nil {
-		return fmt.Errorf("resolver [%+v] found error resolving host [%s] port [%d]: %v", s.resolver, host, port, err)
-	}
-	if !resolverCanResolve {
-		return fmt.Errorf("cannot find resolver for host [%s] port [%d]", host, port)
+	if dest.GetContextToken() != "" {
```
klingerf: It's hard to see it in the context of this diff, but if the context token exists, won't we end up calling subscribe twice, once with the context token, and once without it? I think that will result in sending duplicate updates to the proxy? Should it instead be:

```diff
diff --git a/controller/api/destination/server.go b/controller/api/destination/server.go
index 3fad31a9..13f36d3a 100644
--- a/controller/api/destination/server.go
+++ b/controller/api/destination/server.go
@@ -125,11 +125,11 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 	if dest.GetContextToken() != "" {
 		s.profiles.Subscribe(dest.GetPath(), dest.GetContextToken(), primary)
 		defer s.profiles.Unsubscribe(dest.GetPath(), dest.GetContextToken(), primary)
+	} else {
+		s.profiles.Subscribe(dest.GetPath(), "", secondary)
+		defer s.profiles.Unsubscribe(dest.GetPath(), "", secondary)
 	}
 
-	s.profiles.Subscribe(dest.GetPath(), "", secondary)
-	defer s.profiles.Unsubscribe(dest.GetPath(), "", secondary)
-
 	select {
 	case <-s.shutdown:
 	case <-stream.Context().Done():
```

controller/api/destination/server.go
klingerf: What's going on with the commented out code in this function? It looks like it's no longer implemented. Won't that break the `linkerd endpoints` command? I see in your PR description that you list "re-add the discovery API" as a TODO before merging. Is it referring to the implementation of this endpoint?

controller/api/destination/watcher/profile_watcher.go
```diff
@@ -0,0 +1,234 @@
...
+		Namespace: profile.Namespace,
+		Name:      profile.Name,
+	}
+
+	pw.profilesMu.RLock()
+	publisher, ok := pw.profiles[id]
+	pw.profilesMu.RUnlock()
+	if ok {
+		publisher.update(nil)
+	}
```
scottcarol: Would there be an instance when `!ok` needs further action? Or can we assume that the profile was already deleted in that case?

controller/api/destination/watcher/profile_watcher.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (pp *profilePublisher) subscribe(listener ProfileUpdateListener) {
+	pp.mutex.Lock()
+	defer pp.mutex.Unlock()
+
+	pp.listeners = append(pp.listeners, listener)
+	listener.Update(pp.profile)
+}
+
+// unsubscribe returns true iff the listener was found and removed.
```
scottcarol: iff -> if

controller/api/destination/endpoint_translator_test.go
adleong: I believe that go style is to capitalize the entire acronym (like `actualTLSIdentity`) in identifiers, unless they are at the start of of a private identifier (like `tlsDisablePod`).  However, the `TlsIdentity` field and related methods are code-generated from protobuf and therefore don't follow this language specific convention.

controller/api/destination/profile_translator.go
adleong: nice catch, thanks

controller/api/destination/server.go
adleong: Thanks for bringing this up.  The plan is to remove this implementation while we figure out the right long term solution.  I'll remove the commented out code.

controller/api/destination/watcher/profile_watcher.go
adleong: If the profile isn't present, we've already reached the desired state so no further action should be necessary.

controller/api/destination/endpoint_translator.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/linkerd/linkerd2-proxy-api/go/net"
+	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
+	"github.com/linkerd/linkerd2/pkg/addr"
+	"github.com/linkerd/linkerd2/pkg/k8s"
+	logging "github.com/sirupsen/logrus"
+)
+
+const defaultWeight uint32 = 10000
+
+// endpointTranslator statisfies EndpointUpdateListener and translates updates
```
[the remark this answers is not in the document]
adleong: 🤦‍♂ thanks

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,497 @@
...
+	// PodSet is a set of pods, indexed by IP.
+	PodSet = map[PodID]Address
+
+	// EndpointsWatcher watches all endpoints and services in the Kubernetes
+	// cluster.  Listeners can subscribe to a particular service and port and
+	// endpointsWatcher will publish the address set and all future changes for
+	// that service:port.
+	EndpointsWatcher struct {
+		publishers   map[ServiceID]*servicePublisher
+		publishersMu sync.RWMutex // This mutex protects modifcation of the map itself.
```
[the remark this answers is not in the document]
adleong: this is a really nice pattern, thanks!

controller/api/destination/watcher/profile_watcher.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return ProfileID{}, err
+	}
+	service, _, err := GetServiceAndPort(authority)
+	if err != nil {
+		return ProfileID{}, err
+	}
+	name := ProfileID{
+		Name:      host,
+		Namespace: service.Name,
```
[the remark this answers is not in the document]
adleong: Yes, and actually, I think this is a serious bug in the profile fallback behavior.  I'll add a test to fix the fact that there's no test coverage of this fallback logic.

We don't require context tokens because for each request we do two lookups: one with the context token (to look up the profile in the client's namespace) and one without the context token (to look up the profile in the server's namespace).

controller/api/destination/server.go
adleong: It is intentional that we create two subscriptions here.  The fallbackProfileListener consumes updates from both subscriptions (via the `primary` and `secondary` listeners) and decides which updates to send to the proxy.  I'll add a comment here explaining this behavior.

controller/api/destination/server.go
adleong: That's right.  However, the plan is now to roll forward with this change, leaving this discovery API unimplemented and decide on a more long term solution later.  I will remove this commented code.

controller/api/destination/watcher/profile_watcher.go
adleong: Old mathie habits die hard (https://www.merriam-webster.com/dictionary/iff).  I'll update this to human English.

controller/api/destination/endpoint_translator_test.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+)
+
+func makeEndpointTranslator(t *testing.T) (*mockDestinationGetServer, *endpointTranslator) {
+	mockGetServer := &mockDestinationGetServer{updatesReceived: []*pb.Update{}}
+	translator, err := newEndpointTranslator(
+		"linkerd",
+		"trust.domain",
+		false,
+		"service-name.service-ns.svc.cluster.local",
```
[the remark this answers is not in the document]
adleong: It doesn't necessarily have to, although in reality I think it always would.  This is the authority that is being looked up.  i.e. this is the key that the proxy would query to the destination service.  There's nothing about our API that requires that the pods that the destination service responds with have to be in the same namespace as the query (even though I believe this is always true in Kubernetes).

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,497 @@
...
+	}
+	ew.log.Infof("Stopping watch on endpoint [%s:%d]", id, port)
+
+	ew.publishersMu.RLock()
+	sp, ok := ew.publishers[id]
+	ew.publishersMu.RUnlock()
+	if !ok {
+		ew.log.Errorf("Cannot unsubscribe from unknown service [%s:%d]", id, port)
+	}
+	sp.unsubscribe(port, listener)
```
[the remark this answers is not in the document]
adleong: good catch

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,497 @@
...
+
+func (ew *EndpointsWatcher) deleteService(obj interface{}) {
+	service := obj.(*corev1.Service)
+	id := ServiceID{
+		Namespace: service.Namespace,
+		Name:      service.Name,
+	}
+
+	ew.publishersMu.RLock()
+	sp, ok := ew.publishers[id]
```
ihcsim: I notice that we never delete `ew.publishers[id]`. What is the rationale behind this design?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -0,0 +1,497 @@
...
+	// servicePublisher represents a service along with a port number.  Multiple
+	// listeners may be subscribed to a servicePublisher.  servicePublisher maintains the
+	// current state of the address set and publishes diffs to all listeners when
+	// updates come from either the endpoints API or the service API.
+	servicePublisher struct {
+		id     ServiceID
+		log    *logging.Entry
+		k8sAPI *k8s.API
+
+		ports map[Port]*portPublisher
```
ihcsim: Just a nit: I notice that there are multiple instances of port publishers being called `port` in this file. AFAICT, they are not quite the same thing, right? I found some of these statements confusing:

* `for srcPort, port := range sp.ports`
* `port.updatePort(newTargetPort)`
* `sp.ports[srcPort] = port`

where `port` is actually a `portPublisher` struct.

controller/api/destination/watcher/endpoints_watcher.go
adleong: We could delete the publisher once the number of subscribers drops to 0, but note that it would be re-created as soon as there is an update to that endpoints.  We also want to allow subscriptions to services that don't exist, so that clients can subscribe and wait for a service to be created.

Therefore we only really want to delete a publisher if it has no subscribers AND is for a service that does not currently exist.  Getting this logic exactly right may be tricky so I'd like to punt on this for now.  This is technically leaky, but should only matter when there is a high rate of service churn with new unique names.

Issue #2789, opened 2019-05-03
Adding alternative AWS ACM Private Certificate Authority as an optional CA


khappucino:
## Feature Request
We would like the ability to delegate CSR request handling to the AWS ACM Private Certificate Authority of our choice.

### What problem are you trying to solve?
We need to control the creation of every certificate generated by the Linkerd control plane.

### How should the problem be solved?
We plan on utilizing the Issuer Interface provided by Linkerd2.
We should update the CLI to choose which CA to use.  
When choosing to use the AWS ACMPCA we should provide the ARN:// and the Region.  
We should also allow the user to choose the issuance validity period. AWS ACMPCA seems to have a minimum issuance of a single day.  
There is an option for ABSOLUTE issuance time but that feature does not seem to be implemented yet in aws-sdk-go.  

~We also need to update the linkerd2-proxy, rustls, and webpki. The root cause of this change is that webpki strictly enforces the SignatureAlgorithm Null parameter bytes. The ACM Private Certificate authority leaves off these extra bytes and this causes the linkerd2-proxy to reject the certificates generated by ACMPCA.~  
~The golang crypto, openssl, and various other libraries seem to have relaxed requirements on this.~  
~https://github.com/briansmith/webpki/issues/97 and~   ~https://github.com/Nordstrom/webpki/tree/0.19.1+seanmonstar+cody~  
~It is not clear if the author of webpki will merge our changes so we will probably fork it for now.~  

~Rustls needs to updated to point to our version of webpki.~  
**Update: Apparently AWS ACM Private Certificate Authority has made an update regarding the issue that missing bytes that we found.  The implication is that we do not need to make the linkerd2-proxy, rustls, and webpki changes mentioned earlier in this issue**

What do you want to happen? Add any considered drawbacks.  
We have added a package that handles the delegation to aws under this merge request https://github.com/linkerd/linkerd2/pull/2790 
The main drawback is that until AWS alters their ACM PCA we will need to use a forked version of webpki/rustls  

### Any alternatives you've considered?
We have not considered any other solutions  

Is there another way to solve this problem that isn't as good a solution?

### How would users interact with this feature?
The users would choose to use the AWS ACM Private Certificate Authority by adding extra CLI options.  
The users would need to set the following:  
	--identity-ca-type <1>   (0 is default linkerd2 ca, 1 is the aws acmpca ca)  
    --identity-issuance-lifetime <1 day>  
    --identity-ca-region <AWS Region>  
[...]

PR #2790, opened 2019-05-03
Adding an alternative CA (#2789)

khappucino:
Relevant issue:  https://github.com/linkerd/linkerd2/issues/2789

We created a new package called pcadelegate, this is used to delegate CSR signing to AWS ACM Private Certificate authority. 

We also modified the CLI so that you can specify whether or use the default CA or delegate to AWS ACM Private Certificate Authority.  The CLI parameters are used in the Identity/main.go to determine which CA to use.

~The Docker base was modified to install ca-certs since we cannot communicate with AWS API without validating their Certificates.~  ~This intent of this change was already added previously to Dockerfile-proxy in commit 40e1b53ea94ea96098288c3f1562c5abb79d9a57.~ See update below.

There are changes required in linkerd2-proxy, rustls, and webpki.

linkerd2-proxy - https://github.com/Nordstrom/linkerd2-proxy/tree/custom_webpki_rustls
This is just a change to the Cargo.toml pointing to the forked versions of rustls and webpki.

rustls - https://github.com/Nordstrom/rustls/tree/0.15.1-rsa_nonull
This is just a change to the Cargo.toml pointing to the forked version of webpki.

webpki - https://github.com/Nordstrom/webpki/tree/0.19.1+seanmonstar+cody
This is a change to how Certificate Signature Algorithms are validated, the webpki is very strict and will not allow certificates that are missing the 'Null {}' bytes in the SignatureAlgorithmParameters.
The AWS ACMPCA certificates are missing these 'Null {}' bytes in the certificates they generate.
Relevant links:
briansmith/webpki#97
https://github.com/Nordstrom/webpki/tree/0.19.1+seanmonstar+cody

**UPDATE: 
Looks like AWS changed the SHA256 missing bytes so we don't need a custom linkerd2-proxy as noted by @codeman9 
We did have to add CA-Certs to the controller/Dockerfile in order for the Identity Service to trust AWS**


alpeb:
Hi @khappucino, thanks a lot for this! I'll follow up with a review, but in the meantime I have one quick question: you said "The Docker base was modified to install ca-certs" but it seems that got lost during a merge?

khappucino:
@alpeb

good catch,. I think that change got lost in the shuffle.  I will update the branch. Thanks

codeman9:
Just did some new testing and it looks like AWS has fixed their missing NULL bytes issue. This means that we no longer need a custom build of the linkerd-proxy to support this requirement.

olix0r:
Excellent news, @codeman9! Thanks for the update.

cpretzer:
Closing this after chatting with @codeman9 

We will revisit the requirements once the `cert-manager` integration is complete

PR #2796, opened 2019-05-06
Install MWC and VWC During Control Plane Installation

### LINKED ISSUE #2176: Add MutatingWebhookConfiguration objects to install YAML

ihcsim:
This PR introduces changes to install the proxy injector and profile validator MWC and VWC during the control plane installation.

## Highlights

1. Introduce the chart to install a new `Secret` named `linkerd-ca`
    1. In multi-stage install, this is added to the `control-plane` stage
1. Change the `proxy-injector` and `sp-validator` to:
    1. Mount the `linkerd-ca` as a read-only volume
    1. Use the CA `cert.pem` and `key.pem` to generate a new x509 TLS cert (via the `pkg/tls` library)
1. Preserve the `cert.pem` and `key.pem` of the `linkerd-ca` secret during upgrade (similar mechanism as `linkerd-config`)
1. Introduce a new function to parse CA pem-encoded cert and key, converting the bytes into `tls.CA`.

This PR doesn't affect the existing identity mTLS.

There are no CLI flags for users to override the CA trust with custom cert and private key.

## Test Cases

1. CLI inject, CLI `manual` inject and auto-inject
1. Profile validation
1. Upgrade from previous older versions which don't have the new CA secret
1. In HA mode with multiple proxy injector and profile validator replicas
1. Multi-stage install

## ToDo

1. Not sure about the CA renewal workflow. Currently, the CA is created during the control plane installation only. CA renewal with `install --ignore-cluster` won't work for as it will affect the identity TLS assets.
    1. Once the CA is renewed, the proxy injector and profile validator must be restarted to pick up the new CA assets.
1. Fix the upgrade `--from-manifests` integration tests.

Fixes #2176


cli/cmd/install.go
```diff
@@ -188,6 +197,18 @@ func newInstallOptionsWithDefaults() *installOptions {
...
+
+		generateCATrust: func() (*caTrustValues, error) {
+			root, err := tls.GenerateRootCAWithDefaults(caCommonName)
+			if err != nil {
+				return nil, fmt.Errorf("failed to generate root certificate for control plane CA: %s", err)
+			}
+
+			return &caTrustValues{
+				CertPEM: root.Cred.Crt.EncodeCertificatePEM(),
+				KeyPEM:  root.Cred.EncodePrivateKeyPEM(),
```
alpeb: nit: mind keeping the same order as the struct definition?

chart/templates/sp_validator.yaml
```diff
@@ -71,7 +93,7 @@ spec:
         securityContext:
           runAsUser: {{.ControllerUID}}
       volumes:
-      - name: config
-        configMap:
-          name: linkerd-config
```
alpeb: Nice catch :+1:

cli/cmd/install.go
```diff
@@ -141,6 +148,8 @@ const (
 	defaultIdentityTrustDomain        = "cluster.local"
 	defaultIdentityIssuanceLifetime   = 24 * time.Hour
 	defaultIdentityClockSkewAllowance = 20 * time.Second
+
+	caCommonName = "ca.linkerd.cluster.local"
```
siggy: curious about this name, should it be namespace-specific? also is there a reason this is different from most `*.svc.cluster.local` names, or is this convention?

cli/cmd/install.go
```diff
@@ -340,6 +361,12 @@ func (options *installOptions) validateAndBuild(stage string, flags *pflag.FlagS
 	values.Identity = identityValues
 	values.stage = stage
 
+	caTrust, err := options.generateCATrust()
```
siggy: mind moving `values.stage = stage` below the `CATrust` block (or somewhere more prominent, as it affects so much behavior)

pkg/tls/ca.go
```diff
@@ -262,3 +263,40 @@ func (v *Validity) Window(t time.Time) (time.Time, time.Time) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	keyDecoded, _ := pem.Decode(keyPEM)
+	if keyDecoded == nil {
+		return nil, fmt.Errorf("CA private key must be PEM-encoded")
+	}
+
+	key, err := x509.ParseECPrivateKey(keyDecoded.Bytes)
```
siggy: similar question, possible to re-use `tls.DecodePEMKey`?

pkg/tls/ca.go
```diff
@@ -262,3 +263,40 @@ func (v *Validity) Window(t time.Time) (time.Time, time.Time) {
...
+//
+// Both the CA cert and private key must be PEM-encoded. Otherwise, an error is
+// returned.
+func ParseRootCA(rootCAPEM, keyPEM []byte) (*CA, error) {
+	rootCADecoded, _ := pem.Decode(rootCAPEM)
+	if rootCADecoded == nil {
+		return nil, fmt.Errorf("root CA must be PEM-encoded")
+	}
+
+	rootCA, err := x509.ParseCertificate(rootCADecoded.Bytes)
```
siggy: possible to re-use `tls.DecodePEMCertificates`?

cli/cmd/install.go
```diff
@@ -69,6 +69,8 @@ type (
 		WebResources *resources
 
 		Identity *installIdentityValues
+
+		CATrust *caTrustValues
```
siggy: i'm a little out of context, can you explain the motivation for separate `KeyPEM`/`CertPem` values for the MWC/VWC and the identify system values in `installIdentityValues`? or, taking this a step further, re-using the `linkerd-identity-issuer` secret?

chart/templates/proxy_injector.yaml
```diff
@@ -4,6 +4,24 @@
 ### Proxy Injector
 ###
 ---
+apiVersion: admissionregistration.k8s.io/v1beta1
```
siggy: as a (recently-created) convention, we typically keep cluster-level resources in the `linkerd install config`. everything in `linkerd install control-plane` is scoped to the namespace. what's the motivation for including it in the `control-plane` stage?

ihcsim:
> For some reason only one of the pods gets used. 

That's surprising, because the `Service` in front of the proxy injector replicas should load balance the traffic between the replicas. I'll need to do some more testing on it.

> You also need to add to the to-do list the removal of the code for generating the old configs dynamically. 

Yeah, good point. I am kinda hesitant at first, because in the past, we did have users encountering outdated MWC after upgrading the version (before `linkerd upgrade` was introduced). With the introduction of `linkerd upgrade`, I think we can remove this code as you suggested.

cli/cmd/install.go
ihcsim: Yeah, I didn't put much thought into this name. LMK what's most appropriate. I basically follow the identity issuer's name [convention](https://github.com/linkerd/linkerd2/blob/5ece3430eb72d8286214e599997f3f703f943fd3/cli/cmd/install.go#L875); i.e. without the `.svc` (which makes sense, since there isn't a `Service` with this name).

siggy: I think following the identity issuer's name makes sense. Note that the name varies by namespace.

cli/cmd/install.go
ihcsim: Based on my brief chat with @olix0r, we prefer to keep them separated, because they exist for different purposes. `installIdentityValues`, as we all know, is used for data plane mTLS and consist of fields like issuance life time, clock skew allowances etc. that aren't relevant to the CA. The CA is used to sign certs that will be used by webhooks and taps to authenticate with the k8s API Server.

pkg/tls/ca.go
ihcsim: Good :eyes:. Actually, I can use [`tls.ReadPEMCreds()`](https://github.com/linkerd/linkerd2/blob/3d58c9d86e075d6a20675f2c00bac147106c601f/pkg/tls/cred.go#L139) in the webhook, and remove this new method altogether.

alpeb:
Re HA, for some reason k8s continues to use only one webhook, but if you kill it, the other one starts receiving requests, so this should be fine :+1:

ihcsim:
@alpeb Yeah, it's very strange that the k8s api server seems to only send requests to one replica. AFACIT, the Service` is the common `ClusterIP` type.

If anything, it's good that even when some of the replicas entered into a crash loop, auto-inject still works.
```
k -n linkerd get po -l linkerd.io/control-plane-component=proxy-injector
NAME                                      READY   STATUS             RESTARTS   AGE
linkerd-proxy-injector-5f499c5bff-km9nk   1/2     CrashLoopBackOff   6          6m53s
linkerd-proxy-injector-5f499c5bff-pjjxn   1/2     CrashLoopBackOff   6          8m26s
linkerd-proxy-injector-5f964c49c4-5kq6f   2/2     Running            0          53m


$ k -n linkerd logs linkerd-proxy-injector-5f964c49c4-5kq6f proxy-injector   
...
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-24"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-24"
time="2019-05-08T03:50:44Z" level=info msg="received admission review request 74b7336b-7144-11e9-ab03-ac2175b64ffe"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-23"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-23"
time="2019-05-08T03:50:44Z" level=info msg="received admission review request 74b9e11e-7144-11e9-ab03-ac2175b64ffe"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-18"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-18"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-11"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-11"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-25"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-25"
time="2019-05-08T03:50:44Z" level=info msg="received admission review request 74b96859-7144-11e9-ab03-ac2175b64ffe"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-19"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-19"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-5"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-9"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-9"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-21"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-21"
time="2019-05-08T03:50:44Z" level=info msg="received pod/nginx-16"
time="2019-05-08T03:50:44Z" level=info msg="patch generated for: pod/nginx-16"
```

ihcsim:
> ```shell
> linkerd linkerd-proxy-injector-56459d9d47-6q4r8 proxy-injector 2019/05/08 10:55:11 http: TLS handshake error from 127.0.0.1:43100: remote error: tls: bad certificate
> linkerd linkerd-proxy-injector-56459d9d47-6q4r8 proxy-injector 2019/05/08 10:55:12 http: TLS handshake error from 127.0.0.1:43104: remote error: tls: bad certificate
> linkerd linkerd-proxy-injector-56459d9d47-6q4r8 proxy-injector 2019/05/08 10:55:12 http: TLS handshake error from 127.0.0.1:43116: remote error: tls: bad certificate
> linkerd linkerd-proxy-injector-56459d9d47-6q4r8 proxy-injector 2019/05/08 10:55:13 http: TLS handshake error from 127.0.0.1:43130: remote error: tls: bad certificate
> ```

That's a regression from the change to move the MWC and VWC to the `config` stage. The CA cert and key are generated during both stages. Prior to the change, this has no effects on the `config` stage, because all the CA assets are only injected in the same `control-plane` stage. Will fix.

ihcsim:
@ver @adleong To recap, this PR creates a new `linkerd-ca` secret during install. It has two fields, which correspond to the CA cert and private key. These are used by the webhooks to self-sign its leaf certificate. The CA cert is also duplicated in the MWC's and VWC's `caBundle`, which the k8s api server uses to verify the self-signed certs served by the webhooks. The concern that @siggy raised was around the duplicate CA cert in two resources.

@ver IIUC, you are saying that we don't need to persist the CA in the secret. During install and upgrade, we just (re-)create the CA cert, private key and self-signed certs. The implication is that the MWC and VWC must be re-created during upgrade, to use the new CA assets. Is that the high level idea?

adleong:
Thanks for the recap, @ihcsim.  My understanding is that on startup, each instance of the webhook needs to generate a signed certificate.  To do this, it needs the CA certificate and CA key.  This can be achieved by storing the CA cert and key in a secret (or by storing the CA key in a secret and providing the CA cert in an env var).

If, as @olix0r suggests, the CA key is thrown away after install, how does a new instance of the webhook get a signed certificate?  Or are you suggesting that the webhook certificate should be generated and signed during install and all webhook replicas should share that cert?

olix0r:
I guess my point is that we don't need a _CA_ at all... we only need a self-signed certificate. _If_ we want to generate a root CA to generate the leaf certificates, then I don't see any value in retaining the root certificate.

During upgrade, we can just read the self-signed cert & key, as we do for the issuer certs. If you ever want to create new self-signed certs for these pods, we just generate a new key.

Maybe I'm getting hung up on use of the term "CA" -- are we just generating a self-signed cert? or are we generating a root CA that then signs leaf certificates? If the former, then I don't think there's anything to do but to store the keypair. If the latter, there's no need to store the root CA key; but also there's no real need to have the level of indirection, imo.

olix0r:
After a little thought, my suggestion is that we generate a key & self-signed cert _for each webhook_. If all of the webhooks share a single CA and have to access the root key, then that is effectively equivalent to having them all share a keypair (trust-wise). I don't think the level of indirection that a shared CA provides grants us any security benefits.

adleong:
My assumption was that each webhook replica would have a distinct leaf keypair, but after talking with you, I see there's not really any value in that.

So I think where we're landing with this is that at install, we should generate a (self-signed) certificate and key at startup, put the certificate in the webhook config as the "CA", and distribute the certificate and key to the webhook instances.

ihcsim:
Superseded by #2806.

Issue #2713, opened 2019-04-17
Integration test for reading Kubernetes events


siggy:
A class of Kubernetes issues, such as #2694, are not currently caught in our integration tests. Introduce an integration test that reviews Kubernetes events post-install.

alpeb:
We're consistently getting k8s warnings after install, for failed readiness probes on proxies seemingly because they won't succeed until the main container is not ready. Possible ways to fix:
- appropriately setting the injected proxy's `initialDelaySeconds` depending on the main container (e.g. Prometheus takes 30s for its probe to succeed)
- not use readiness probes in proxies. Proxies aren't added to Endpoints so I think it makes sense to forgo them

siggy:
> appropriately setting the injected proxy's initialDelaySeconds depending on the main container (e.g. Prometheus takes 30s for its probe to succeed)

i think i'd rather we have some failed probes on startup, and come up as quickly as possible, rather than setting `initialDelaySeconds`.

> not use readiness probes in proxies. Proxies aren't added to Endpoints so I think it makes sense to forgo them

i think it's worthwhile to explore removing this. if it does not cause unexpected failure, i'd be in favor of it.

for a bit more context, the goal of this particular ticket is just to wire up the integration tests to Kubernetes events, and report anomalies, similar to our logging tests.

alpeb:
@klingerf proved me in #2798 we can't just remove the probes. I'll most likely then just consider those warnings as expected, as alluded to by @siggy

PR #2798, opened 2019-05-06
Remove readiness probe from sidecar proxies

alpeb:
Proxies aren't referred in any service/endpoints resource, so there's
no reason for them to expose a readiness probe.

Also the probes as they're currently declared are usually failing during
install because the main container is usually ready afterwards (e.g.
Prometheus probe takes 30s), which causes warnings events in k8s.

Part of #2713

### DIFF
--- a/cli/cmd/testdata/inject-filepath/expected/injected_nginx.yaml
+++ b/cli/cmd/testdata/inject-filepath/expected/injected_nginx.yaml
@@ -97,11 +97,6 @@ spec:
           name: linkerd-proxy
         - containerPort: 4191
           name: linkerd-admin
-        readinessProbe:
-          httpGet:
-            path: /ready
-            port: 4191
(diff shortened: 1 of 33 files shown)


klingerf:
I don't think it's safe to remove the readiness check for the linkerd-proxy container. A pod won't be considered Ready until all of its containers have passed their readiness checks. If we remove the proxy readiness check, then a pod could start receiving requests before the proxy has started, but after iptables rules have been rewritten, and in that scenario the requests will fail.

alpeb:
@klingerf Thanks for the warning, you're right! I wasn't thinking in terms of the whole pod readiness...

olix0r:
It may be the case that we can remove the _liveness_ probe for the proxy (I'm not sure); but the readiness probe is definitely useful so that readiness is coupled with certificate provisioning.

PR #2803, opened 2019-05-08
Introduce pre-install healthcheck for clock skew

### LINKED ISSUE #2762: detect clock skew in `linkerd check --pre`
A user reported this error from the linkerd-identity pod when installing Linkerd 2.3 on minikube:

> Failed to verify issuer credentials for 'identity.linkerd.cluster.local' with trust anchors: x509: certificate has expired or is not yet valid" in linkerd-identity pod.

This was due to the clock of their minikube VM being wrong. Can we detect this as part of `linkerd check`?

matej-g:
**Subject**
Introduce pre-install healthcheck for clock skew

**Problem**
Recently, a user has reported a failed Linkerd install, which was manifesting itself as an error when verifying certificate in linkerd-identity pod. After investigation, the user discover that the issue was caused by clock skew between his system and the Kubernetes node (Minikube in this case). The problem in this scenario is that the user is not aware of the clock skew, as this problem manifests itself as a different issue (i.e. invalid / expired certificate).

**Solution**
To alleviate the issue, an additional pre-install healthcheck for clock skew was suggested. This PR implements such a healthcheck in following way:
- It queries the Kubernetes API for node status(es) and takes the last heartbeat timestamp received from the node as the determining value
- It compares the timestamp with system time of machine running the pre-install check
- If these two time values differ by more than 70 seconds (60 seconds for the presumed maximal interval for node heartbeat + currently 10 second clock skew as allowed by default by TLS package) for any node, fail the check and return names of the failing nodes

For details please follow the thread in the issue comments

**Validation**
This change has been tested with 1) Minikube v1.0.0 node running in VM and 2) local test cluster composed of 2 nodes (running Kuberenets v1.13; the system times have been manipulated to test different scenarios)

Fixes #2762

Signed-off-by: Matej Gera <matejgera@gmail.com>

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -14,6 +14,7 @@ import (
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/profiles"
+	"github.com/linkerd/linkerd2/pkg/tls"
 	"github.com/linkerd/linkerd2/pkg/version"
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
@@ -102,6 +103,14 @@ const (
 // point to. Each check adds its own `hintAnchor` to specify a location on the
 // page.
 const HintBaseURL = "https://linkerd.io/checks/#"
+
+// AllowedClockSkew sets the allowed skew in clock synchronization
+// between the system running inject command and the node(s), being
+// based on assumed node's heartbeat interval (<= 60 seconds) plus default TLS
+// clock skew allowance.
+//
+// TODO: Make this default value overridiable, e.g. by CLI flag
+const AllowedClockSkew = (60 * time.Second) + tls.DefaultClockSkewAllowance
 
 var (
 	retryWindow    = 5 * time.Second
@@ -324,6 +333,13 @@ func (hc *HealthChecker) allCategories() []category {
 					hintAnchor:  "pre-k8s",
 					check: func(context.Context) error {
 						return hc.checkCanCreate(hc.ControlPlaneNamespace, "", "v1", "configmaps")
+					},
+				},
+				{
+					description: "no clock skew detected",
+					hintAnchor:  "pre-k8s-clock-skew",
+					check: func(context.Context) error {
+						return hc.checkClockSkew()
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


matej-g:
So I see the lint fails straightaway, strangely enough it did not flag any issues on my local, although I ran it without `--verbose`, added fix

matej-g:
@klingerf Thanks for the comment, added the patch

pkg/healthcheck/healthcheck.go
```diff
@@ -988,6 +1004,43 @@ func (hc *HealthChecker) checkNetAdmin() error {
...
+
+	nodeList, err := hc.kubeAPI.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodeList.Items {
+		for _, condition := range node.Status.Conditions {
+			// we want to check only KubeletReady condition and only execute if the node is ready
+			if condition.Type == "Ready" && condition.Status == "True" {
```
siggy: mind using:
```go
if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
```

pkg/healthcheck/healthcheck.go
```diff
@@ -988,6 +1004,43 @@ func (hc *HealthChecker) checkNetAdmin() error {
...
+				ts := condition.LastHeartbeatTime.Time
+				if (time.Since(ts) > AllowedClockSkew) || (time.Since(ts) < -AllowedClockSkew) {
```
siggy: tioli:
```go
since := time.Since(condition.LastHeartbeatTime.Time)
if (since > AllowedClockSkew) || (since < -AllowedClockSkew) {
```
(also i can't think of an duration-type-safe-way to validate magnitude with a single comparison :p )

pkg/healthcheck/healthcheck.go
```diff
@@ -988,6 +1004,43 @@ func (hc *HealthChecker) checkNetAdmin() error {
...
+					clockSkewNodes = append(clockSkewNodes, node.Name)
+				}
+			}
+		}
+	}
+
+	errMsg := ""
+	if len(clockSkewNodes) > 0 {
+		errMsg = fmt.Sprintf("Clock skew detected for node(s): %s", strings.Join(clockSkewNodes, ", "))
```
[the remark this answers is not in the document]
matej-g: Yeah, I was aware of the error convention, but seen other errors string also started with capital letter, so I think I implicitly went with it. Anyways, changing it to be conform.

matej-g:
@siggy Thanks for the review! I incorporated the comments, hopefully everything is now OK.

matej-g:
Right, forgot to update test case but fixed, CI should succeed now

PR #2807, opened 2019-05-09
Support for resources opting-out of tap

### LINKED ISSUE #2778: Add inject flag for disabling tap
Add a `--disable-tap` flag to inject which prevents the pod from being tapped by instructing the proxy to not expose a tap server.

* Introduce a proxy environment variable to disable the tap server
* Add an inject flag to set the proxy env var
* Also set an annotation on the pod to indicate that tap has been disabled on that pod so that the control plane knows to exclude that pod from any taps

alpeb:
Fixes #2778

This one is just for the control-plane side of things.
The env var injected into the sidecar proxy was labeled `LINKERD2_PROXY_TAP_DISABLED`.

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -104,6 +104,10 @@ sub-folders, or coming from stdin.`,
 		"Disables resources from participating in TLS identity",
 	)
 	flags.BoolVar(
+		&options.disableTap, "disable-tap", options.disableTap,
+		"Disables resources from from being tapped",
+	)
+	flags.BoolVar(
 		&options.ignoreCluster, "ignore-cluster", options.ignoreCluster,
 		"Ignore the current Kubernetes cluster when checking for existing cluster configuration (default false)",
 	)
@@ -379,6 +383,10 @@ func (options *proxyConfigOptions) overrideConfigs(configs *cfg.All, overrideAnn
 	if options.disableIdentity {
 		configs.Global.IdentityContext = nil
 		overrideAnnotations[k8s.ProxyDisableIdentityAnnotation] = "true"
+	}
+
+	if options.disableTap {
+		overrideAnnotations[k8s.ProxyDisableTapAnnotation] = "true"
 	}
 
 	// keep track of this option because its true/false value results in different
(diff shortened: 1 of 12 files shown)


cli/cmd/inject.go
```diff
@@ -103,6 +103,10 @@ sub-folders, or coming from stdin.`,
 		&options.disableIdentity, "disable-identity", options.disableIdentity,
 		"Disables resources from participating in TLS identity",
 	)
+	flags.BoolVar(
+		&options.disableTap, "disable-tap", options.disableTap,
+		"Disables resources from from being tapped",
+	)
```
siggy: until the proxy supports it, should we do `flags.MarkHidden("disable-tap")` ? maybe reference a Github issue in a comment for the proxy change? or, since the control-plane is also filtering, maybe this is ok?

related: does this mean both the control-plane and the proxy will independently guard against tapping?

controller/tap/server.go
```diff
@@ -68,7 +68,7 @@ func (s *server) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_T
 		}
 
 		for _, pod := range podsFor {
-			if pkgK8s.IsMeshed(pod, s.controllerNamespace) {
+			if pkgK8s.IsMeshed(pod, s.controllerNamespace) && !pkgK8s.IsTapDisabled(pod) {
```
siggy: tioli, the `if len(pods) == 0` below returns a pretty generic `no pods found for...` message. consider returning a more descriptive message if pods were filtered out due to tap disabling.

test/tap/tap_test.go
```diff
@@ -112,6 +112,23 @@ func TestCliTap(t *testing.T) {
 		}
 	})
 
+	t.Run("tap a disabled deployment", func(t *testing.T) {
```
siggy: 🙏

cli/cmd/inject.go
alpeb: As is, tapping gets disabled through the control plane API, but that doesn't disallow someone hitting directly the proxy. I think it makes sense to mark the flag hidden in the meantime.

alpeb: Created #2811 to track proxy-side work

Issue #2708, opened 2019-04-17
Create `linkerd edges identity` CLI command and subcommand for exposing identity


scottcarol:
### What problem are you trying to solve?

Currently, users have no easy way of knowing whether connections between Linkerd proxies have a known identity, and if not, why.

The new TLS Identity system now enables TLS by default when a user installs Linkerd, and for each edge (*edge* refers to the connection between two nodes, in this case two Linkerd proxies) we can surface data for the identity of each proxy, or the reason identity is not known. 

To make this information actionable by users, we need to surface it in the CLI. This will help users verify that Linkerd proxies have a known identity, and debug why a connection may not be secured with TLS.

### How should the problem be solved?

Create a new CLI command called `linkerd edges` to expose information about connections between Linkerd proxies. Initially it will have just one subcommand.

`linkerd edges identity` is a subcommand comparable to `linkerd stat` that will surface the known identities for the source and destination of each edge. If identities are not known, it will print a reason.

Ultimately, every use case outlined in `linkerd stat -h` should have a `linkerd edges identity` counterpart, although the first area to tackle will be `pods`.

### Any alternatives you've considered?

Currently, identity information is being surfaced in Linkerd's metrics and it is possible for a user to cobble together information with Prometheus queries. However this is not quick and it takes prior knowledge of labels and where to find the data.

Instead of a  new `edges` command, identity could be surfaced as part of the normal `stat` query, but this could be confusing. Adding `edges` sets the stage for more functionality to be built later around identity. 

### How would users interact with this feature?

 Below is an example of CLI output. The column names are:
    - SRC: the source of the connection
    - DST: the destination of the connection
    - CLIENT: name of K8s service account identity for the SRC.
    - SERVER: name of K8s service account identity for DST.
    - MSG: message (reason) explaining why there is no identity for SRC/DST, in human-readable format.

```bash
$ linkerd -n emojivoto edges identity po
 SRC                                     DST                      CLIENT   SERVER   MSG
 web-56955474f7-hsqnn                    emoji-596866d567-6wl6s   web      emoji
 gke-foobar-default-pool-5f6c4d44-7jjh   web-56955474f7-hsqnn                       No Authority In HTTP Request
 emoji-596866d567-6wl6s                  172.217.5.110                              Not Provided By Remote
 emoji-596866d567-6wl6s                  mongo-1                                    TCP
$ linkerd -n booksapp edges identity deploy
 SRC                  DST       CLIENT    SERVER    MSG
[...]

Issue #2783, opened 2019-05-02
Replace --disable-identity with --disable-tls


adleong:
Linkerd inject supports a `--disable-identity` flag which prevents the Linkerd proxy from establishing an identity and thus prevents the Linkerd proxy from doing TLS on data plane traffic.  However, the lack of identity prevents the proxy from securely serving administrative services such as tap and metrics.

In order to be able to serve these admin services over TLS, the Linkerd proxy should *always* establish an identity.  I suggest we remove the `--disable-identity` flag.  

To support the use-case where data plane traffic should not be encrypted with TLS, I propose we add a `--disable-tls` flag which prevents the Linkerd proxy from doing TLS on data plane traffic.  However, communication with the control plane should always be secured with TLS, even when `--disable-tls` is set.

ihcsim:
If I follow what @grampelberg was saying yesterday, we can extend the skip ports functionality to the namespace level, instead of introducing another inject flag. That way, identity is still on by default, but user decides which ports don't require TLS.

olix0r:
I'm -1 on calling it disable-tls. I think it would be more accurately called `disable-mesh-traffic-identity` or something less vague

alpeb:
@ihcsim That'd be effectively removing out of the mesh the connections going through that port.

If people ask for a non mTLS'd mesh, it looks like we do need to separate a pod's ability of establishing identity against the control plane from the ability to perform secured connections with other members of the mesh.

sbvitok:
Any news? We need non mTLS'd mesh.

hawkw:
@sbvitok 
> Any news? We need non mTLS'd mesh.

You can get that now using `disable-identity`, if you don't mind also not having mTLS for tap, metrics, etc.

sbvitok:
> @sbvitok
> 
> > Any news? We need non mTLS'd mesh.
> 
> You can get that now using `disable-identity`, if you don't mind also not having mTLS for tap, metrics, etc.

You suggest using service mesh (linkerd), without service mesh?
We need service mesh, with tap, metrics, etc. and without TLS in data plane traffic.

hawkw:
> > @sbvitok
> > > Any news? We need non mTLS'd mesh.
> > 
> > 
> > You can get that now using `disable-identity`, if you don't mind also not having mTLS for tap, metrics, etc.
> 
> You suggest using service mesh (linkerd), without service mesh?
[...]

PR #2808, opened 2019-05-10
Adding an edges command to the CLI

scottcarol:
This PR adds an `edges` command to the CLI and continues work on #2708. When you specify one of six resource types and the name of the namespace, it will call the API and return the data in a table. 

It takes these resource types:
```
deployment
pod
replicationcontroller
daemonset
statefulset
job
```

To test:
1. Install emojivoto and inject Linkerd
`curl -sL https://run.linkerd.io/emojivoto.yml | linkerd inject - | kubectl apply -f -`

2. Run the public API locally (this should start HTTP server on `:8085`)
`bin/go-run controller/cmd/public-api -log-level debug`

3. Port forward Prometheus
`kubectl -n linkerd port-forward $(kubectl --namespace=linkerd get po --selector=linkerd.io/control-plane-component=prometheus -o jsonpath='{.items[*].metadata.name}') 9090:9090`

4. Run `edges` in the terminal, communicating directly with the control plane at `:8085`
`bin/go-run cli edges po -n emojivoto --api-addr localhost:8085`

Result should be:
```bash
SRC                                             DST              CLIENT             SERVER   MSG
web-57b7f9db85-6kjs7         emoji-646ddcc5f9-b6phk       web.emojivoto    emoji.emojivoto     -
web-57b7f9db85-6kjs7        voting-689f845d98-llnst       web.emojivoto   voting.emojivoto     -
vote-bot-7466ffc7f7-c8vk8      web-57b7f9db85-6kjs7   default.emojivoto      web.emojivoto     -
```

*Known issue - if you run `bin/go-run cli edges daemonset -n emojivoto --api-addr localhost:8085`
there is a table row with a blank `src` and `dst` even though there are no actual daemonsets returned. This is due to the Prometheus query and should be fixed in the controller in a separate PR.

### DIFF
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -1,0 +1,297 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/linkerd/linkerd2/controller/api/util"
+	pb "github.com/linkerd/linkerd2/controller/gen/public"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+type indexedEdgeResults struct {
+	ix   int
+	rows []*pb.Edge
+	err  error
+}
+
+func newCmdEdges() *cobra.Command {
+	options := newStatOptions()
+
+	cmd := &cobra.Command{
+		Use:   "edges [flags] (RESOURCETYPE)",
+		Short: "Display connections between resources, and the identity of their associated Linkerd proxies (if known)",
+		Long: `Display connections between resources, and the identity of their associated Linkerd proxies (if known).
+
+  The RESOURCETYPE argument specifies the type of resource to display edges within. A namespace must be specified.
+
+  Examples:
+  * deploy
+  * ds
+  * job
+	* po
+	* rc
+	* sts
+
+  Valid resource types include:
+  * daemonsets
+  * deployments
+  * jobs
+  * pods
+  * replicationcontrollers
+  * statefulsets
+
+If no resource name is specified, displays edges within all resources of the specified RESOURCETYPE`,
+		Example: `  # Get all edges between pods in the test namespace.
+  linkerd edges deploy -n test`,
+		Args:      cobra.MinimumNArgs(1),
+		ValidArgs: util.ValidTargets,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			reqs, err := buildEdgesRequests(args, options)
+			if err != nil {
+				return fmt.Errorf("error creating metrics request while making edges request: %v", err)
+			}
+
+			// The gRPC client is concurrency-safe, so we can reuse it in all the following goroutines
+			// https://github.com/grpc/grpc-go/issues/682
+			client := checkPublicAPIClientOrExit()
+			c := make(chan indexedEdgeResults, len(reqs))
+			for num, req := range reqs {
+				go func(num int, req *pb.EdgesRequest) {
+					resp, err := requestEdgesFromAPI(client, req)
+					if err != nil {
+						fmt.Println(err)
+					}
+
+					rows := edgesRespToRows(resp)
+
+					c <- indexedEdgeResults{num, rows, err}
+				}(num, req)
+			}
+
+			totalRows := make([]*pb.Edge, 0)
+			i := 0
+			for res := range c {
+				if res.err != nil {
+					return res.err
+				}
+				totalRows = append(totalRows, res.rows...)
+				if i++; i == len(reqs) {
+					close(c)
+				}
+			}
+
+			output := renderEdgeStats(totalRows, options)
+			_, err = fmt.Print(output)
+
+			return err
+		},
+	}
+
+	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace of the specified resource")
+	cmd.PersistentFlags().StringVarP(&options.outputFormat, "output", "o", options.outputFormat, "Output format; one of: \"table\" or \"json\"")
+	return cmd
+}
+
+func buildEdgesRequests(resources []string, options *statOptions) ([]*pb.EdgesRequest, error) {
+	targets, err := util.BuildResources(options.namespace, resources)
+	if err != nil {
+		return nil, err
+	}
+
+	requests := make([]*pb.EdgesRequest, 0)
+	for _, target := range targets {
+		err = options.validate(target.Type)
+		if err != nil {
+			return nil, err
+		}
+
+		requestParams := util.EdgesRequestParams{
+			ResourceType: target.Type,
+			Namespace:    options.namespace,
+		}
+
+		req, err := util.BuildEdgesRequest(requestParams)
+		if err != nil {
+			return nil, err
+		}
+		requests = append(requests, req)
+	}
+	return requests, nil
+}
+
+func edgesRespToRows(resp *pb.EdgesResponse) []*pb.Edge {
+	rows := make([]*pb.Edge, 0)
+	if resp != nil {
+		for _, edgeTable := range resp.GetOk().Edges {
+			rows = append(rows, edgeTable)
+		}
+	}
+	return rows
+}
+
+func requestEdgesFromAPI(client pb.ApiClient, req *pb.EdgesRequest) (*pb.EdgesResponse, error) {
+	resp, err := client.Edges(context.Background(), req)
+	if err != nil {
+		return nil, fmt.Errorf("Edges API error: %+v", err)
+	}
+	if e := resp.GetError(); e != nil {
+		return nil, fmt.Errorf("Edges API response error: %+v", e.Error)
+	}
+	return resp, nil
+}
+
+func renderEdgeStats(rows []*pb.Edge, options *statOptions) string {
+	var buffer bytes.Buffer
+	w := tabwriter.NewWriter(&buffer, 0, 0, padding, ' ', tabwriter.AlignRight)
+	writeEdgesToBuffer(rows, w, options)
+	w.Flush()
+
+	return renderEdges(buffer, &options.statOptionsBase)
+}
+
+type edgeRowStats struct {
+	src    string
+	dst    string
+	client string
+	server string
+	msg    string
+}
+
+func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *statOptions) {
+	edgeTables := make(map[string]*edgeRowStats)
+	if len(rows) != 0 {
+		for _, r := range rows {
+			key := string(r.Dst.Name + r.Src.Name)
+
+			edgeTables[key] = &edgeRowStats{
+				client: r.ClientId,
+				server: r.ServerId,
+				msg:    r.NoIdentityMsg,
+				src:    r.Src.Name,
+				dst:    r.Dst.Name,
+			}
+		}
+	}
+	switch options.outputFormat {
+	case tableOutput:
+		if len(edgeTables) == 0 {
+			fmt.Fprintln(os.Stderr, "No edges found.")
+			os.Exit(0)
+		}
+		printSingleEdgeTable(edgeTables, w)
+	case jsonOutput:
+		printEdgesJSON(edgeTables, w)
+	}
+}
+
+// returns the length of the longest src name
+func srcWidth(stats map[string]*edgeRowStats) int {
+	maxLength := 0
+	for _, row := range stats {
+		if len(row.src) > maxLength {
+			maxLength = len(row.src)
+		}
+	}
+	return maxLength
+}
+
+func printSingleEdgeTable(edges map[string]*edgeRowStats, w *tabwriter.Writer) {
+	// template for left-aligning the src column
+	srcTemplate := fmt.Sprintf("%%-%ds", srcWidth(edges))
+
+	headers := []string{
+		fmt.Sprintf(srcTemplate, "SRC"),
+	}
+	headers = append(headers, []string{
+		"DST",
+		"CLIENT",
+		"SERVER",
+		"MSG",
+	}...)
+
+	headers[len(headers)-1] = headers[len(headers)-1] + "\t" // trailing \t is required to format last column
+
+	fmt.Fprintln(w, strings.Join(headers, "\t"))
+
+	sortedKeys := sortEdgesKeys(edges)
+	for _, key := range sortedKeys {
+		values := make([]interface{}, 0)
+		templateString := srcTemplate + "\t%s\t%s\t%s\t%s\t\n"
+
+		if edges[key].msg == "" {
+			edges[key].msg = "-"
+		}
+
+		values = append(values, []interface{}{
+			edges[key].src,
+			edges[key].dst,
+			edges[key].client,
+			edges[key].server,
+			edges[key].msg,
+		}...)
+
+		fmt.Fprintf(w, templateString, values...)
+
+	}
+}
+
+func renderEdges(buffer bytes.Buffer, options *statOptionsBase) string {
+	var out string
+	switch options.outputFormat {
+	case jsonOutput:
+		out = buffer.String()
+	default:
+		// strip left padding on the first column
+		out = string(buffer.Bytes()[padding:])
+		out = strings.Replace(out, "\n"+strings.Repeat(" ", padding), "\n", -1)
+	}
+
+	return out
+}
+
+func sortEdgesKeys(stats map[string]*edgeRowStats) []string {
+	var sortedKeys []string
+	for key := range stats {
+		sortedKeys = append(sortedKeys, key)
+	}
+	sort.Strings(sortedKeys)
+	return sortedKeys
+}
+
+type edgesJSONStats struct {
+	Src    string `json:"src"`
+	Dst    string `json:"dst"`
+	Client string `json:"client_id"`
+	Server string `json:"server_id"`
+	Msg    string `json:"no_tls_reason"`
+}
+
+func printEdgesJSON(edgeTables map[string]*edgeRowStats, w *tabwriter.Writer) {
+	// avoid nil initialization so that if there are not stats it gets marshalled as an empty array vs null
+	entries := []*edgesJSONStats{}
+
+	for _, row := range edgeTables {
+		entry := &edgesJSONStats{
+			Src:    row.src,
+			Dst:    row.dst,
+			Client: row.client,
+			Server: row.server,
+			Msg:    row.msg}
+		entries = append(entries, entry)
+	}
+
+	b, err := json.MarshalIndent(entries, "", "  ")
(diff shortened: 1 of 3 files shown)


cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+	rows []*pb.Edge
+	err  error
+}
+
+func newCmdEdges() *cobra.Command {
+	options := newStatOptions()
+
+	cmd := &cobra.Command{
+		Use:   "edges [flags] (RESOURCETYPE)",
+		Short: "Display connections between resources, and the identity of their associated Linkerd proxies (if known)",
```
siggy: i'd shorten this up. while we're not perfectly adhering to 80 chars for the default cli help output, this line is significantly longer than the other commands.

maybe something like:
```
Display connections between resources, and Linkerd proxy identities
```

```
Available Commands:
  check       Check the Linkerd installation for potential problems
  completion  Output shell completion code for the specified shell (bash or zsh)
  dashboard   Open the Linkerd dashboard in a web browser
  edges       Display connections between resources, and the identity of their associated Linkerd proxies (if known)
  endpoints   Introspect Linkerd's service discovery state
  get         Display one or many mesh resources
  help        Help about any command
  inject      Add the Linkerd proxy to a Kubernetes config
  install     Output Kubernetes configs to install Linkerd
  install-cni Output Kubernetes configs to install Linkerd CNI (experimental)
  install-sp  Output Kubernetes configs to install Linkerd Service Profiles
  logs        Tail logs from containers in the Linkerd control plane
  metrics     Fetch metrics directly from Linkerd proxies
  profile     Output service profile config for Kubernetes
  routes      Display route stats
  stat        Display traffic stats about one or many resources
  tap         Listen to a traffic stream
  top         Display sorted information about live traffic
  uninject    Remove the Linkerd proxy from a Kubernetes config
  upgrade     Output Kubernetes configs to upgrade an existing Linkerd control plane
  version     Print the client and server version information
```

cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+type indexedEdgeResults struct {
+	ix   int
+	rows []*pb.Edge
+	err  error
+}
+
+func newCmdEdges() *cobra.Command {
+	options := newStatOptions()
```
siggy: we might want to consider using a new `edgesOptions` struct:
```go
type edgesOptions struct {
	namespace    string
	outputFormat string
}
```

`statOptions` supports a number of fields that are unused by `edges`. the underlying `statOptionsBase` is close, but it's still a superset of what we need. alternatively, you could create a struct underneath `statOptionsBase`:
```go
type optionsBase struct {
	namespace    string
	outputFormat string
}
type statOptionsBase struct {
    optionsBase
	timeWindow   string
}
```

another motivation for not re-using `statOptions` is that the output format validation will succeed on `-owide` even though `edges` doesn't support it:
```bash
$ bin/linkerd edges -n linkerd deploy -owide
panic: runtime error: slice bounds out of range
```

cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+
+  Valid resource types include:
+  * daemonsets
+  * deployments
+  * jobs
+  * pods
+  * replicationcontrollers
+  * statefulsets
+
+If no resource name is specified, displays edges within all resources of the specified RESOURCETYPE`,
```
siggy: probably can remove this sentence, since we don't support things like:
```bash
linkerd edges deploy/my-deploy
```

cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+		}
+	}
+	switch options.outputFormat {
+	case tableOutput:
+		if len(edgeTables) == 0 {
+			fmt.Fprintln(os.Stderr, "No edges found.")
+			os.Exit(0)
+		}
+		printSingleEdgeTable(edgeTables, w)
```
siggy: i think we either want to explicitly support multiple resource types (similar to `stat`), or disallow it:
```bash
$ bin/linkerd edges -n linkerd deploy/ po/
SRC                                                                   DST                                                                     CLIENT                                                                     SERVER   MSG
linkerd-controller-86ddf787bb-wz44v   linkerd-prometheus-595b874c57-zx8nw   linkerd-controller.linkerd.serviceaccount.identity.linkerd.cluster.local   linkerd-prometheus.linkerd.serviceaccount.identity.linkerd.cluster.local     -
linkerd-controller                                     linkerd-prometheus   linkerd-controller.linkerd.serviceaccount.identity.linkerd.cluster.local   linkerd-prometheus.linkerd.serviceaccount.identity.linkerd.cluster.local     -
```

```bash
$ bin/linkerd stat -n linkerd deploy/ po/
NAME                                         MESHED   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TCP_CONN
deploy/linkerd-controller                       1/1   100.00%   0.9rps           1ms           8ms          10ms         10
deploy/linkerd-grafana                          1/1   100.00%   0.3rps           1ms           2ms           2ms          2
...

NAME                                         MESHED   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TCP_CONN
po/linkerd-controller-86ddf787bb-wz44v          1/1   100.00%   0.9rps           1ms           8ms          10ms         10
po/linkerd-grafana-c46ccbbd7-4r2x7              1/1   100.00%   0.3rps           1ms           2ms           2ms          2
...
```

cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+		},
+	}
+
+	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace of the specified resource")
+	cmd.PersistentFlags().StringVarP(&options.outputFormat, "output", "o", options.outputFormat, "Output format; one of: \"table\" or \"json\"")
+	return cmd
+}
+
+func buildEdgesRequests(resources []string, options *statOptions) ([]*pb.EdgesRequest, error) {
+	targets, err := util.BuildResources(options.namespace, resources)
```
siggy: unless we're planning to support things like `deploy/my-deploy`, we probably want to validate that `target.Name` is empty, somewhere around here.

cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+
+func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *statOptions) {
+	edgeTables := make(map[string]*edgeRowStats)
+	if len(rows) != 0 {
+		for _, r := range rows {
+			key := string(r.Dst.Name + r.Src.Name)
+
+			edgeTables[key] = &edgeRowStats{
+				client: r.ClientId,
+				server: r.ServerId,
```
siggy: are the identities consistently of the form `resource.namespace.serviceaccount.identity.linkerd.cluster.local`. if so, can we drop the `.serviceaccount.identity.linkerd.cluster.local`, it makes the table output wrap on my laptop screen:
Screen Shot 2019-05-10 at 12 39 44 PM

cli/cmd/edges.go
```diff
@@ -0,0 +1,297 @@
...
+  * daemonsets
+  * deployments
+  * jobs
+  * pods
+  * replicationcontrollers
+  * statefulsets
+
+If no resource name is specified, displays edges within all resources of the specified RESOURCETYPE`,
+		Example: `  # Get all edges between pods in the test namespace.
+  linkerd edges deploy -n test`,
```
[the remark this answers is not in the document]
scottcarol: good catch, thanks!!

cli/cmd/edges.go
scottcarol: Thanks, I was trying to figure out how to express that simply - your suggestion is much more clear.

cli/cmd/edges.go
scottcarol: I was attempting to re-use `statOptions` in order to DRY things up but you're right, the validation error is 👎 . Good point.

cli/cmd/edges.go
scottcarol: Yup, that's the consistent format of the identities.

controller/api/util/api_utils.go
```diff
@@ -220,6 +227,32 @@ func BuildStatSummaryRequest(p StatsSummaryRequestParams) (*pb.StatSummaryReques
 	return statRequest, nil
 }
 
+// BuildEdgesRequest builds a Public API EdgesRequest from a
+// EdgesRequestParams.
+func BuildEdgesRequest(p EdgesRequestParams) (*pb.EdgesRequest, error) {
```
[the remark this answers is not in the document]
scottcarol: I was mirroring the format of `stat.go` as I built this file so that's why I put `BuildEdgesRequest` in `api_utils.go`. I haven't looked into the web implementation yet. Would it make more sense to have `BuildEdgesRequest` in `edges.go`?

cli/cmd/edges.go
scottcarol: Thanks for catching that. Multiple resource types seems like something that would be 👍 to support and would take just a few tweaks.

alpeb:
@scottcarol thanks for the detailed testing steps :+1: 
It's working fine for me in the regular scenario:
```
$ bin/go-run cli edges po -n emojivoto                                                
SRC                                             DST                                                            CLIENT                                                           SERVER   MSG
web-8bcffdc9-h9pf8           emoji-6bf9f47bd5-l4svc       web.emojivoto.serviceaccount.identity.linkerd.cluster.local    emoji.emojivoto.serviceaccount.identity.linkerd.cluster.local     -
web-8bcffdc9-h9pf8          voting-75ccdd49c4-wcwtd       web.emojivoto.serviceaccount.identity.linkerd.cluster.local   voting.emojivoto.serviceaccount.identity.linkerd.cluster.local     -
vote-bot-59646d6cf9-z7jk7        web-8bcffdc9-h9pf8   default.emojivoto.serviceaccount.identity.linkerd.cluster.local      web.emojivoto.serviceaccount.identity.linkerd.cluster.local     -
```
Then I added the `config.linkerd.io/disable-identity: "true"` annotation to the `web` deployment and I redeployed again. Then I got an empty response:
```
$ bin/go-run cli edges po -n emojivoto                                                                                                                                                                                                                                         
No edges found.
```
We should be getting the same as before with a msg explaining identity has been disabled. But maybe Prometheus doesn't have enough info in this case? I'll dig around.

On the other hand you don't have `replicasets` in you list of supported resource types. Is that intended?

scottcarol:
@alpeb Thanks for testing and for catching that - let me try on my end as well with `disable-identity` and see what the issue is.

Yes, not supporting `replicasets` is intentional - when @klingerf is back on Monday I will clarify with him the reason why, I know `stat` also does not support `replicasets` although both `edges` and `stat` support `replicationcontrollers`.

alpeb:
Ok I think my test didn't make any sense...
All connections go through `web` and I disabled identity in that pod. `linkerd edges` rightfully ignores anything that doesn't have identity, so nothing is shown.

What would make sense is not to disable identity, but to disable tls, and then see how `linkerd edges` reports it. But that won't happen until #2783 is addressed.

scottcarol:
Hey @alpeb, currently `linkerd edges` is just for two-way edges, i.e. a query from one pod to another pod. However a later stage of this command will include non-meshed traffic, like prometheus, tap, and health checks, which will not have an identity (this is why there is a `msg` column - in Prometheus the label is `no_tls_reason`). 

I think you are bringing up a very good question and it's good to think about `linkerd edges` in the context of disabled identity -- can you clarify the steps you followed to disabled identity for the `web` deployment so I can look at the Prometheus data?

alpeb:
@scottcarol one way to do it through the CLI is by re-injecting using `--disable-identity`, on an already deployed (and injected, doesn't matter) web deployment:
```
kubectl -n emojivoto get deployments web -oyaml | bin/go-run cli inject --disable-identity - | kubectl apply -f -
```

scottcarol:
Oof, accidentally rebased in my local branch after making this PR and thought force pushing would fix, but instead it messed up my commit history. Sorry about that. You can view the code differences from my original PR and my new commit [by viewing 'Changes from 1 commit'](https://github.com/linkerd/linkerd2/pull/2808/commits/617f54f73e0d9e46a0dd95c3a4e0df30867557b5)

cli/cmd/edges.go
scottcarol: On second thought I'd like to merge a simpler command that only accepts one resource type, then go from there. I changed `newCmdEdges` to accept only one resource argument with `cobra. ExactArgs(1)`.

alpeb: @scottcarol If you're only accepting one resource type for now, then you might wanna validate against having `all` being passed, as it's currently returning something like this:
```
$ bin/go-run cli edges -n emojivoto all
SRC   DST              CLIENT          SERVER   MSG
            default.emojivoto   web.emojivoto     -
```

cli/cmd/edges.go
```diff
@@ -0,0 +1,325 @@
...
+	for _, row := range stats {
+		if len(row.src) > maxLength {
+			maxLength = len(row.src)
+		}
+	}
+	return maxLength
+}
+
+func printSingleEdgeTable(edges map[string]*edgeRowStats, w *tabwriter.Writer) {
+	// template for left-aligning the src column
```
rmars: it seems like the headers are right aligned except for the first column:
```
$ linkerd edges po -n linkerd
SRC                                                                   DST                       CLIENT                       SERVER   MSG
linkerd-controller-5dc5958dbc-s4w2d   linkerd-prometheus-7c6d8fd47b-jhz8v   linkerd-controller.linkerd   linkerd-prometheus.linkerd     -
```
I'd expect the text column headers (and text column values) to be left aligned.

cli/cmd/edges.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+	err = validateEdgesOutputFormat(options)
+	if err != nil {
+		return nil, err
+	}
+
+	requests := make([]*pb.EdgesRequest, 0)
+	for _, target := range targets {
+		if target.Name != "" {
+			target.Name = "" // validating that no resource name is specified in the request
```
rmars: I think it'd be nicer to make a validation error here to let the user know that they can't get edges for a specific resource name, it might be confusing to drop this silently and return results anyway

cli/cmd/edges.go
```diff
@@ -0,0 +1,325 @@
...
+			}
+		}
+	}
+	switch options.outputFormat {
+	case tableOutput:
+		if len(edgeTables) == 0 {
+			fmt.Fprintln(os.Stderr, "No edges found.")
+			os.Exit(0)
+		}
+		printSingleEdgeTable(edgeTables, w)
```
rmars: just curious: why is this method called `printSingleEdgeTable`? is it because we're showing one side of a potentially bidirectional edge?

cli/cmd/edges.go
scottcarol: When I run `linkerd routes deploy -n emojivoto` the first column is left-aligned but all subsequent columns are right-aligned including text columns. I do see that `linkerd top` and `linkerd endpoints` left-align all text columns. Is our style generally to left-align text? If so I can go ahead and change that.

Screen Shot 2019-05-14 at 12 53 21 PM

cli/cmd/edges.go
scottcarol: Great question 👁. In the future you should be able to run `linkerd edges` for multiple resource types and `printSingleEdgeTable` will be called each time, a la `stat.go`, but this specific PR doesn't support that yet. For clarity I'll change it to `printEdgeTable`.

cli/cmd/edges.go
rmars: yea, iirc the style should be to left align text columns, and right align numbers.

scottcarol: Ok cool, thanks for the spot 🔍

cli/cmd/edges.go
scottcarol: @alpeb I believe I've addressed this in my latest commit. Thanks for the thorough tests of this feature! 🙏

scottcarol:
In my latest commit I formatted the columns so they are all left-justified. I would appreciate feedback on whether I did this in the most elegant #golang way.

cli/cmd/edges.go
```diff
@@ -0,0 +1,360 @@
...
+			if err != nil {
+				return fmt.Errorf("Error creating edges request: %s", err)
+			}
+
+			// The gRPC client is concurrency-safe, so we can reuse it in all the following goroutines
+			// https://github.com/grpc/grpc-go/issues/682
+			client := checkPublicAPIClientOrExit()
+			c := make(chan indexedEdgeResults, len(reqs))
+			for num, req := range reqs {
+				go func(num int, req *pb.EdgesRequest) {
```
siggy: these concurrent requests are intended for multiple-resource commands like `linkerd edges deploy/ pod/`. if we're planning to support that at some point, i'm ok leaving it here. if not, better to just have `buildEdgesRequests` return a single request, and make that request synchronously.

cli/cmd/edges.go
```diff
@@ -0,0 +1,360 @@
...
+			edges[key].server,
+			edges[key].msg,
+		}...)
+
+		fmt.Fprintf(w, templateString, values...)
+
+	}
+}
+
+func renderEdges(buffer bytes.Buffer, options *edgesOptions) string {
```
siggy: tioli: consider re-using `renderStats`.

it would require changing its signature to:
```go
func renderStats(buffer bytes.Buffer, outputFormat string) string {
```
... but i think that's a good change to make anyway.

cli/cmd/edges.go
```diff
@@ -0,0 +1,360 @@
...
+			Dst:    row.dst,
+			Client: row.client,
+			Server: row.server,
+			Msg:    row.msg}
+		entries = append(entries, entry)
+	}
+
+	b, err := json.MarshalIndent(entries, "", "  ")
+	if err != nil {
+		log.Error(err.Error())
```
siggy: i realize we're not consistent about printing errors in the cli (`log.Error` vs. `fmt.Print(os.Stderr...`), but i'd prefer the latter, as logging is more of a server thing, and being explicit about stdout vs. stderr in the cli is a good thing:
```go
fmt.Fprintf(os.Stderr, "Error marshalling JSON: %s", err)
```
(@klingerf and i had a good exchange in some long ago PR i can't find, where he convinced me of this position)

cli/cmd/edges.go
scottcarol: Yes, planning on supporting multiple-resource commands in the next iteration of this feature, so going to leave this in for now.

cli/cmd/edges.go
scottcarol: Great idea. Flagging for a separate PR.

cli/cmd/edges.go
scottcarol: Makes sense 👍

scottcarol:
Quick note that #2825 should solve the issue that @alpeb and @siggy saw when querying for a resource type not present in a specific namespace.

Issue #2809, opened 2019-05-10
web: build errors on bin/web setup for fsevents using node 12


cpretzer:
## Bug Report
Build errors occur when running `bin/web setup` with node 12 on MacOS Mojave.

**NB**: The messages indicate that the module is optional and the errors can be ignored, however the dashboard will crash. I'll file a separate issue for that.

### What is the issue?
When following the [web build instructions](https://github.com/linkerd/linkerd2/blob/master/BUILD.md#web) using node 12 or higher, the build errors in [this related gist](https://gist.github.com/cpretzer/824b2c4b0e08474613ff71c84ebad5e0) occur, because the webpack-dev-server defined in 3.1.14 depends on a version of [chokidar](https://github.com/paulmillr/chokidar) which has a dependency on an incompatible version of fsevents (1.2.4).

The build process attempts to retrieve the version of fsevents v1.2.4 for node 12 (v72) which is not available. The build process tries build the library because the file does not exist on the fsevents CDN. See the errors in the error section below.

### How can it be reproduced?
These steps assume that homebrew is installed.

Install node version 12 or higher and follow the steps in the [build guide for the web](https://github.com/linkerd/linkerd2/blob/master/BUILD.md#web)

```
brew install node
```

verify node version 12
```
node --version
```

clone the repository, install yarn, and run setup
```
git clone https://github.com/linkerd/linkerd2
brew install yarn
bin/web setup
```

### Logs, error output, etc
```
node-pre-gyp http GET https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.4/fse-v1.2.4-node-v72-darwin-x64.tar.gz
node-pre-gyp http 404 https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.4/fse-v1.2.4-node-v72-darwin-x64.tar.gz
```

The underlying error comes from nan:
```
In file included from ../fsevents.cc:6:
In file included from ../../nan/nan.h:223:
In file included from ../../nan/nan_new.h:189:
../../nan/nan_implementation_12_inl.h:356:37: error: too few arguments to function call, expected 2, have 1
  return v8::StringObject::New(value).As<v8::StringObject>();
         ~~~~~~~~~~~~~~~~~~~~~      ^
/Users/user/.node-gyp/12.1.0/include/node/v8.h:5380:3: note: 'New' declared here
  static Local<Value> New(Isolate* isolate, Local<String> value);
  ^
```

[...]

PR #2818, opened 2019-05-11
2817 enable shorthands for lodash

### LINKED ISSUE #2817: web: "iteratee is not a function" error when using Node 12
## Bug Report
Building the dashboard with Node 12 results in a JavaScript error from the BaseTable component.

This is related to #2809 

### What is the issue?
BaseTable.jsx uses `lodash/sortBy` and using Node 12 appears to pull in a different version or cause a stricter check on function parameters.

### How can it be reproduced?
These instructions assume that the developer has a kubernetes cluster and linkerd is installed there.

Start with a clean environment.
```
rm -rf web/app/node_modules
```

Run the web setup
```
bin/web setup
```

Run the dev server (also works with the `run` command
```
bin/web dev
```

Open the dashboard in the browser:
http://localhost:7777

### Logs, error output, etc
The complete console output is [here](https://gist.github.com/cpretzer/ab2e716c4b41556df78cfc31814da857)

#### `linkerd check` output

kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane self-check
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √

### Environment

- Kubernetes Version:
v1.10.11

Kubernetes on Docker for Mac

- Host OS:
macOS Mojave

- Linkerd version:
stable-2.3.0

### Possible solution
Set the `state.orderBy` at line 57 to an empty array if `this.props.defaultOrderBy` is undefined:

```
  constructor(props) {
    super(props);
    this.state = {
      order: this.props.defaultOrder || "asc",
      orderBy: this.props.defaultOrderBy || [],
      filterBy: ""
    };
    this.handleFilterInputChange = this.handleFilterInputChange.bind(this);
    this.handleFilterToggle = this.handleFilterToggle.bind(this);
  }
```

cpretzer:
Problem:
The parameter to sortBy is not properly interpreted as a function, and the `iteratee is not a function` error occurs.

Solution
Set `shorthand: true` in webpack.config.js

Validation
I tested with both node 10 and 12, including tests with `bin/web run` and `bin/web dev` for each version of node.

I have questions about the knock-on effects of setting shorthands to true, specifically around the size of the file that is generated. I ask that one or more of the reviewers validate that this is the right change.

It's possible that this fix as well as the fix for #2809 are not the optimal fixes.

Fixes #2817 

Notes:
This branch includes a merge from pretzer/2809-update-webpack-dev-server-version


scottcarol:
Hey @cpretzer thanks so much for this PR! 👍 Can you give me instructions on replicating the `iteratee is not a function` error? Is this happening when you are attempting to sort on the dashboard or are you seeing it before then? Thanks!

scottcarol:
Oops, I just saw that you had detailed output [here](https://github.com/linkerd/linkerd2/issues/2817). Let me try to replicate this on my end!

web/app/package.json
```diff
@@ -68,7 +68,7 @@
     "webpack": "4.20.2",
     "webpack-bundle-analyzer": "^3.0.3",
     "webpack-cli": "3.0.7",
-    "webpack-dev-server": "3.1.14"
+    "webpack-dev-server": "^3.2.0"
```
rmars: can we continue with the version pinning already here? we can pin to `3.3.1` rather than `^`

scottcarol:
Ah, thanks for the clarification on bundle size @rmars, I was comparing the actual size of `yarn.lock` in the master branch (401kb) vs this branch (421kb) in the terminal, but it makes sense to look at the whole bundle.

web/app/package.json
cpretzer: No problem. I'll change to pin the version to 3.3.1.

cpretzer:
@rmars The latest commit removes the lodash rules as well as the plugin

PR #2820, opened 2019-05-14
Added controller component labels to webhook configurations

### LINKED ISSUE #2602: Make sure all the components are labeled with "linkerd.io/control-plane-component"
At least I know of the proxy-injector's mutating webhook config which is not labelled so.

This allows to act on all Linkerd resources as a whole through label selection.

Pothulapati:
Fixes #2602 

This adds the `linkerd.io/control-plane-component` label to the proxy-injector and sp-validator webhook Congurations.

The following query would work after this
`kubectl get MutatingWebhookConfigurations,ValidatingWebhookConfigurations  -l 'linkerd.io/control-plane-component in (proxy-injector,sp-validator)'`


Pothulapati:
Ah Right! I will keep an eye an eye out on the PR, and submit a PR regarding labels, once it merges. :)

Issue #2819, opened 2019-05-13
proxy-injector fails with a restricted psp


codeman9:
### What is the issue?
We are noticing that the proxy-injector in our cluster is unable to successfully complete injection depending on the PSP that is in place. We are using the linkerd-cni plugin and the linkerd install command is as follows:
`linkerd install-cni --proxy-uid 12102 -l tram-linkerd` (our dest-cni-bin-dir and dest-cni-net-dir might be different than the default)
`linkerd install --ha --linkerd-cni-enabled --proxy-uid 12102 -l tram-linkerd`

### How can it be reproduced?
Set up PSP as follows:
restricted-psp
```yaml
apiVersion: extensions/v1beta1
kind: PodSecurityPolicy
metadata:
  name: restricted
spec:
  allowPrivilegeEscalation: false
  fsGroup:
    ranges:
    - max: 65535
      min: 10001
    rule: MustRunAs
  readOnlyRootFilesystem: true
  requiredDropCapabilities:
  - ALL
  runAsUser:
    rule: MustRunAsNonRoot
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    ranges:
    - max: 65535
      min: 10001
    rule: MustRunAs
  volumes:
  - configMap
  - emptyDir
  - projected
  - secret
  - downwardAPI
  - persistentVolumeClaim
```

permissive-psp
```yaml
apiVersion: extensions/v1beta1
kind: PodSecurityPolicy
metadata:
  name: permissive
spec:
  allowPrivilegeEscalation: true
  allowedCapabilities:
  - '*'
  fsGroup:
    rule: RunAsAny
  hostIPC: true
  hostNetwork: true
  hostPID: true
  hostPorts:
  - max: 65535
    min: 0
  privileged: true
  runAsUser:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - '*'
```
When you have those in place, create roles and rolebindings that use the PSPs:
```yaml
apiVersion: v1
items:
- apiVersion: rbac.authorization.k8s.io/v1
  kind: Role
  metadata:
    name: hello-psp-permissive
    namespace: hello
  rules:
  - apiGroups:
    - policy
    resourceNames:
    - permissive
    resources:
    - podsecuritypolicies
    verbs:
    - use
- apiVersion: rbac.authorization.k8s.io/v1
  kind: Role
  metadata:
[...]

PR #2821, opened 2019-05-14
add the allowPrivilegeEscalation property to the security context who…

codeman9:
…se value depends on whether or not CNI is enabled.

Resolves: #2819 


codeman9:
Doing some more testing on this to ensure that it works as expected with our cluster. Hold on merging for a bit.

codeman9:
After doing some more testing, there are 2 other properties that need to be set on the container in order for the pods to start up and become injected on our cluster: `requiredDropCapabilities` must drop "ALL" and `readOnlyRootFilesystem` must be "true".

hercynium:
Just FYI, I tried this patch myself and it worked perfectly for me.

ihcsim:
@codeman9 @hercynium Sorry for the delay. I have been running a few tests using both the CNI and non-CNI setups, on GKE. What I found is that we can safely set `allowPrivilegeEscalation: false` and `readOnlyRootFilesystem: true` for both setups. For _capabilities_, instead of using the `Drop: ALL` rules which won't work if a psp uses different `Drop` rules, we can copy the capabilities of the primary container (injected by the psp mutator) into the proxy and proxy-init (for non-CNI) containers, for both setups.

Would you mind trying out this patch to see if it works for you? You should be able to just `git apply` it either to this branch of `master`. The only real change happens in `pkg/inject/inject.go` and the rest of are just unit test fixes.

[patch.txt](https://github.com/linkerd/linkerd2/files/3262644/patch.txt)

hercynium:
> Would you mind trying out this patch to see if it works for you? You should be able to just `git apply` it either to this branch of `master`. The only real change happens in `pkg/inject/inject.go` and the rest of are just unit test fixes.

Will try later tonight.

hercynium:
Applying the patch to master works, but had some trouble deploying it
`$ k logs -n linkerd linkerd-sp-validator-5fb9fbfd7f-ffp8w sp-validator
time="2019-06-06T20:58:25Z" level=info msg="running version dev-39b89420-scaffidi"
time="2019-06-06T20:58:25Z" level=fatal msg="failed to read TLS secrets: open /var/run/linkerd/tls/key.pem: no such file or directory"`

Tried applying to stable-2.3.2 but doesn't apply cleanly.

Will look closer a bit later tonight.

hercynium:
@ihcsim - I got the build and deployment working. It was as you hinted - I needed to use the same version of the linkerd CLI utility to do the deployment to make it work.

The patch works perfectly in my cluster. I tested several different apps without any issue, both with an annotation on the namespace and an annotation on the pods in a deployment. This is a cluster for integration testing, deployed with the same settings and restrictions as our production clusters. I also tested manual injection via `linkerd inject ...`

Tomorrow morning I'l try to coax my linkerd POC guinea-pigs to redeploy their stuff to see if there are any regressions.

codeman9:
@ihcsim This patch works for me too in my testing. I will update this PR to include this patch. Awesome!

codeman9:
The DCO failures are not my commits...unclear what to do there.

ihcsim:
@codeman9 I'm not sure why GH thinks that this PR is trying to merge 39 commits into `master` :laughing: AFAICT, those commits are already in `master`. It might be easier if you just create a new PR based off the current `master` `HEAD`, with my patch.

codeman9:
Ok, I will do that.

codeman9:
Closing this PR as it is replaced with: #2911

PR #2824, opened 2019-05-15
Split proxy-init into separate repo

### LINKED ISSUE #2563: Split proxy-init into separate repo
The proxy-init container rarely changes. Yet, we build it on every build, publish it on every release, etc. Furthermore, because the proxy version we inject with needs to match the init version, it complicates testing proxy versions.

Why don't we split this code out into a separate repo (i.e. linkerd/proxy-init) and depend on a less-frequently-changing version during inject? This streamlines the normal build, release, and development workflows.

alpeb:
Fixes #2563

The new repo is https://github.com/linkerd/linkerd2-proxy-init, and I
tagged the latest there `stable-2.3.0`. The plan would be to tag it each
time it is updated, using the same tag of the linkerd2 repo that
required that change. I think that helps keeping track of which
proxy-init version is required by which linkerd2 version. Let me know if
you have a different opinion!

Here, I've removed the `/proxy-init` dir and pinned the injected
proxy-init version to `stable-2.3.0` in the injector code and tests.

`/cni-plugin` depends on proxy-init, so I updated the import paths
there, and could verify CNI is still working (there is some flakiness
but unrelated to this PR).

For consistency, I added a `--init-image-version` flag to `linkerd
inject` along with its corresponding override config annotation. If
deemed too much, we can hide it (along with `--init-image`).


olix0r:
> i think i have a slight preference for versioning the proxy-init repo separately from the main linkerd2 repo.

I have a strong preference to keep them separate ;)  Can we just call the proxy-init 1.0.0?

siggy:
> i did not realize until just now that linkerd-cni has a code-level Go dependency on proxy-init. this could make further development in that area more painful.

reflecting on this a bit, given how infrequently the proxy-init code changes, this is probably not a big deal.

> I have a strong preference to keep them separate ;) Can we just call the proxy-init 1.0.0?

agree, let's tag things as `v1.0.0` (per https://github.com/golang/go/wiki/Modules#modules)

alpeb:
Thanks for the feedback!
I've tagged https://github.com/linkerd/linkerd2-proxy-init to `v1.0.0`, but I have no permissions to push the docker image into google cloud, which will make the integration tests fail.
Can anyone assist?

proto/config/config.proto
```diff
@@ -43,6 +43,8 @@ message Proxy {
   bool disable_external_profiles = 12;
 
   string proxy_version = 13;
+
+  string proxy_init_image_version = 14;
```
siggy: on line 19 of this file:
```diff
diff --git a/proto/config/config.proto b/proto/config/config.proto
index 81dfd645..19e3320f 100644
--- a/proto/config/config.proto
+++ b/proto/config/config.proto
@@ -16,7 +16,7 @@ message Global {
   string linkerd_namespace = 1;
   bool cni_enabled = 2;

-  // Control plane and proxy-init version
+  // Control plane version
   string version = 3;

   // If present, configures identity.
```

alpeb: Good :eye: !

test/install_test.go
```diff
@@ -83,6 +83,7 @@ var (
 		`MountVolume.SetUp failed for volume .* : couldn't propagate object cache: timed out waiting for the condition`,
 		`Readiness probe failed: HTTP probe failed with statuscode: 50(2|3)`,
 		`(Liveness|Readiness) probe failed: Get http://.*: dial tcp .*: connect: connection refused`,
+		`Failed to update endpoint .*: Operation cannot be fulfilled on endpoints ".*": the object has been modified; please apply your changes to the latest version and try again`,
```
siggy: i noticed another event from a broken l5d-bot run, mind adding:
```diff
diff --git a/test/install_test.go b/test/install_test.go
index b55eea5c..41d15e84 100644
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -81,7 +81,7 @@ var (

        knownEventWarningsRegex = regexp.MustCompile(strings.Join([]string{
                `MountVolume.SetUp failed for volume .* : couldn't propagate object cache: timed out waiting for the condition`,
-               `Readiness probe failed: HTTP probe failed with statuscode: 50(2|3)`,
+               `(Liveness|Readiness) probe failed: HTTP probe failed with statuscode: 50(2|3)`,
                `(Liveness|Readiness) probe failed: Get http://.*: dial tcp .*: connect: connection refused`,
        }, "|"))
```

pkg/version/version.go
```diff
@@ -12,6 +12,12 @@ import (
 // DO NOT EDIT
 var Version = undefinedVersion
 
+// ProxyInitVersion is the pinned version of the proxy-init, from
+// https://github.com/linkerd/linkerd2-proxy-init
+// This has to be kept in sync with the constraint version for
+// github.com/linkerd/linkerd2-proxy-init in /Gopkg.toml
+var ProxyInitVersion = "v1.0.0"
```
ihcsim: AFAICT, this is currently hard-coded, right? I wonder if we should update the [`docker build`](https://github.com/linkerd/linkerd2/blob/bd4c2788fa19f31b0fb2f32254ac6fd52c737565/bin/_docker.sh#L45-L49) command to inject this value at build time. Something like
```
docker build  -ldflags "-X version.ProxyInitVersion=${PROXY_INIT_VERSION}"
```

pkg/inject/inject.go
```diff
@@ -717,10 +717,10 @@ func (conf *ResourceConfig) proxyVersion() string {
 }
 
 func (conf *ResourceConfig) proxyInitVersion() string {
-	if version := conf.configs.GetGlobal().GetVersion(); version != "" {
-		return version
+	if override := conf.getOverride(k8s.ProxyInitImageVersionAnnotation); override != "" {
```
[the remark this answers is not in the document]
alpeb: Good call :+1: I'll fix that.

pkg/version/version.go
alpeb: Yeah, it's hard-coded. I see how it can be cleaner to parametrize this kind of things, but this one will hardly ever change. Where would we be setting `PROXY_INIT_VERSION`?

ihcsim: Probably in `bin/docker-build` and maybe `.travis.yml`, with hard-coded default if the `PROXY_INIT_VERSION` environment variable is empty. Yeah, I agree that since this doesn't change often, what we have here is fine, unless others have different opinion on it.

PR #2825, opened 2019-05-15
Adding a not-nil check to the Prometheus query for edges

scottcarol:
This PR adds a not-nil check to the generated Prometheus query in the API endpoint for `edges`. 

Previously the queries generated were:
```
count(response_total{direction="inbound", namespace="emojivoto"}) by (replicaset, client_id)
count(response_total{direction="outbound", namespace="emojivoto"}) by (replicaset, dst_replicaset, server_id, no_tls_reason)
```

With these queries, if there were no `replicaset` metrics Prometheus would skip that label and return data for the other labels. 

The new queries generated should be:

```
count(response_total{replicaset!="", direction="inbound", namespace="emojivoto"}) by (replicaset, client_id)
count(response_total{replicaset!="", direction="outbound", namespace="emojivoto"}) by (replicaset, dst_replicaset, server_id, no_tls_reason)
```

If there is no `replicaset` data, no data will be returned. This should fix the issue in #2808 where `bin/linkerd edges -n linkerd sts` was returning data when it should have been empty.

### DIFF
--- a/controller/api/public/edges.go
+++ b/controller/api/public/edges.go
@@ -62,8 +62,13 @@ func (s *grpcServer) getEdges(ctx context.Context, req *pb.EdgesRequest) ([]*pb.
 	}
 	resourceType := string(labelNames[1]) // skipping first name which is always namespace
 	labels := promQueryLabels(req.Selector.Resource)
-	labelsOutbound := labels.Merge(promDirectionLabels("outbound"))
-	labelsInbound := labels.Merge(promDirectionLabels("inbound"))
+	labelsOutbound := fmt.Sprint(labels.Merge(promDirectionLabels("outbound")))
+	labelsInbound := fmt.Sprint(labels.Merge(promDirectionLabels("inbound")))
+
+	// checking that data for the selected resource type exists
+	resourceExistsCheck := fmt.Sprintf(`%s!="",`, resourceType)
+	labelsOutbound = labelsOutbound[:1] + resourceExistsCheck + labelsOutbound[1:]
+	labelsInbound = labelsInbound[:1] + resourceExistsCheck + labelsInbound[1:]
 
 	inboundQuery := fmt.Sprintf(inboundIdentityQuery, labelsInbound, resourceType)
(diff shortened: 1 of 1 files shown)


controller/api/public/edges.go
```diff
@@ -62,8 +62,13 @@ func (s *grpcServer) getEdges(ctx context.Context, req *pb.EdgesRequest) ([]*pb.
...
 	labels := promQueryLabels(req.Selector.Resource)
-	labelsOutbound := labels.Merge(promDirectionLabels("outbound"))
-	labelsInbound := labels.Merge(promDirectionLabels("inbound"))
+	labelsOutbound := fmt.Sprint(labels.Merge(promDirectionLabels("outbound")))
+	labelsInbound := fmt.Sprint(labels.Merge(promDirectionLabels("inbound")))
+
+	// checking that data for the selected resource type exists
+	resourceExistsCheck := fmt.Sprintf(`%s!="",`, resourceType)
+	labelsOutbound = labelsOutbound[:1] + resourceExistsCheck + labelsOutbound[1:]
+	labelsInbound = labelsInbound[:1] + resourceExistsCheck + labelsInbound[1:]
```
rmars: :sob: I don't know if this would be any less weird/confusing, but we could do what the prometheus library does with stringifying label sets, but also add in `!=` capabilities...

```
func CustomToString(l LabelSet, resourceType string) string {
    lstrs := make([]string, 0, len(l))
    for l, v := range l {
        lstrs = append(lstrs, fmt.Sprintf("%s=%q", l, v))
    }
  lstrs = append(lstrs, fmt.Sprintf(`%s!="",`, resourceType))

    sort.Strings(lstrs)
    return fmt.Sprintf("{%s}", strings.Join(lstrs, ", "))
}
```

scottcarol:
Thanks @rmars for the help with the logic on this one. After I made this PR, `linkerd edges` was merged to master. I'm closing this PR and opening a new one from a branch that contains the `linkerd edges` command so reviewers can see the different CLI behavior.

PR #2826, opened 2019-05-15
Validate that data exists for resource type in Prometheus

scottcarol:
Logic for this PR is credit @rmars who modeled her code on the Prometheus library. 

This PR adds a check to the Prometheus queries for edges, in order to verify that data for a specified label name exists. Previous queries generated: 
```
count(response_total{direction="inbound", namespace="emojivoto"}) by (replicaset, client_id)
count(response_total{direction="outbound", namespace="emojivoto"}) by (replicaset, dst_replicaset, server_id, no_tls_reason)
```
With these queries, if there were no `replicaset` metrics Prometheus would skip that label and return data for the other labels. 

The new queries generated should be:

```
count(response_total{replicaset!="", direction="inbound", namespace="emojivoto"}) by (replicaset, client_id)
count(response_total{replicaset!="", direction="outbound", namespace="emojivoto"}) by (replicaset, dst_replicaset, server_id, no_tls_reason)
```

To check in the CLI, run these steps on `master` and on this branch:
1. Install emojivoto and inject Linkerd
`curl -sL https://run.linkerd.io/emojivoto.yml | linkerd inject - | kubectl apply -f -`

2. Run the public API locally (this should start HTTP server on :8085)
`bin/go-run controller/cmd/public-api -log-level debug`

3. Port forward Prometheus
`kubectl -n linkerd port-forward $(kubectl --namespace=linkerd get po --selector=linkerd.io/control-plane-component=prometheus -o jsonpath='{.items[*].metadata.name}') 9090:9090`

4. Run edges in the terminal, communicating directly with the control plane at :8085
`bin/go-run cli edges rc -n emojivoto --api-addr localhost:8085`

The incorrect response is:
```bash
SRC   DST   CLIENT              SERVER             MSG
            default.emojivoto   web.emojivoto      -
```

The correct response should be:
```bash
No edges found.
```

### DIFF
--- a/controller/api/public/prometheus.go
+++ b/controller/api/public/prometheus.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"sort"
+	"strings"
 	"time"
 
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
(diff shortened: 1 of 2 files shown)


controller/api/public/prometheus.go
```diff
@@ -116,6 +118,20 @@ func promDstQueryLabels(resource *pb.Resource) model.LabelSet {
 	return set
 }
 
+// insert a not-nil check into a LabelSet to verify that data for a specified
+// label name exists. due to the `!=` this must be inserted as a string. the
+// structure of this code is taken from the Prometheus labelset.go library.
+func customToString(l model.LabelSet, labelName string) string {
```
rmars: nit: I think it should also be made clearer from the name of this fn that it's generating the prometheus label query string with exclusions. Wordier but more descriptive suggestion: `generateLabelStringWithExclusion`

scottcarol: How about `generateLabelSetStringWithNotNilCheck`?

rmars: fine by me!

PR #2829, opened 2019-05-16
Update CHANGES.md for edge-19.5.3 release

rmars:


CHANGES.md
```diff
@@ -1,3 +1,20 @@
...
+* CLI
+  * **New** Added a `linkerd edges` command that shows TLS status
+    between src and dst resources
+  * Tap can now be disabled by using the `--disable-tap` flag
+  * Introduced pre-install healthcheck for clock skew (thanks, @matej-g!)
+* Controller
+  * Added Controller Component Labels to webhooks (thanks, @Pothulapati!)
+  * Moved the tap service into its own pod
+* Proxy
+  * Upgraded mio to 0.6.17
```
olix0r: Prefer: Fix an epoll notification issue that could cause excessive CPU usage

CHANGES.md
```diff
@@ -1,3 +1,20 @@
...
+  * Allow disabling tap by setting an env var (thanks, @zaharidichev!)
+  * Added tests to ensure that outbound TLS works even when there is no verified
```
olix0r: not user facing, omit

CHANGES.md
```diff
@@ -1,3 +1,20 @@
...
+    certificate for the proxy yet (thanks, @zaharidichev!)
+  * Fall back to ORIG_DST when no endpoints exist
```
olix0r: not user facing, please omit

CHANGES.md
```diff
@@ -1,3 +1,20 @@
+## edge-19.5.3
+
+* CLI
+  * **New** Added a `linkerd edges` command that shows TLS status
+    between src and dst resources
+  * Tap can now be disabled by using the `--disable-tap` flag
```
alpeb: I would be more specific on this, saying it can be disabled for specific pods during injection by ways of the `--disable-tap`  flag, or its corresponding `config.linkerd.io/disable-tap` annotation.

CHANGES.md
```diff
@@ -1,3 +1,20 @@
...
+  * Introduced pre-install healthcheck for clock skew (thanks, @matej-g!)
+* Controller
+  * Added Controller Component Labels to webhooks (thanks, @Pothulapati!)
```
alpeb: ... to the webhook config resources

CHANGES.md
```diff
@@ -1,3 +1,20 @@
+## edge-19.5.3
+
+* CLI
+  * **New** Added a `linkerd edges` command that shows TLS status
```
scottcarol: I think we're trying to differentiate between TLS and identity in general, so I think it would be better to say "Added a `linkerd edges` command that shows connections between src and dst resources and the identities of their Linkerd proxies. This will help users verify that Linkerd proxies have a known identity."

scottcarol: Update, recommendation from @klingerf: "Added a `linkerd edges` command that shows the source and destination name and identity for proxied connections, to assist in debugging" I think that's more clear.

rmars: thanks!

CHANGES.md
```diff
@@ -1,3 +1,21 @@
...
+    `--disable-tap` flag, or by using the `config.linkerd.io/disable-tap`
+    annotation
+  * Introduced pre-install healthcheck for clock skew (thanks, @matej-g!)
+* Controller
+  * Added Controller Component Labels to the webhook config resources (thanks,
+    @Pothulapati!)
+  * Moved the tap service into its own pod
+* Proxy
+  * Fix an epoll notification issue that could cause excessive CPU usage
+  * Allow disabling tap by setting an env var (thanks, @zaharidichev!)
```
hawkw: Can we phrase this so that it begins with "add", i.e. "Added the ability to disable tap by..."?

hawkw: Or "Add LINKERD2_PROXY_TAP_DISABLED env var which disable tap when set" if my prior suggestion was too passive voice?

PR #2835, opened 2019-05-20
Modify upgrade integration test to upgrade stable

siggy:
In #2679 we introduced an upgrade integration test. At the time we only
supported upgrading from a recent edge. Since that PR, a stable build
was released supporting upgrade.

Modify the upgrade integration test to upgrade from the latest stable
rather than latest edge. This fulfills the original intent of #2669.

Also add some known k8s event warnings.

### DIFF
--- a/bin/test-run
+++ b/bin/test-run
@@ -38,32 +38,32 @@ function run_test(){
     go test -v "$filename" --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@"
 }
 
-# Install the latest edge release.
-# $1 - namespace to use for the edge release
-function install_edge() {
+# Install the latest stable release.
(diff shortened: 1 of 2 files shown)


ihcsim:
@dadjeibaah I think it's ok to limit the integration test to just test stable -> stable for now, to cover typical user's production upgrade scenario. There may be some values in covering the edge -> stable scenario (not so sure about stable -> edge), but we will likely run into more test failures related to cluster-scoped resources; i.e. depending on if we run stable -> stable or edge -> stable upgrade first, the test failures might be different.

dadjeibaah:
@ihcsim that makes sense. Testing for those other use cases are more aligned with edge testing and its not the typical production use case.

PR #2837, opened 2019-05-20
Adding unit test for CLI edges command

scottcarol:
This PR adds a unit test for the new `linkerd edges` CLI command. It also removes two lines from `validateEdgesRequestInputs` in `edges.go` that were causing the function to skip one of the validation checks (discovered in the process of writing the test).

### DIFF
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -112,11 +112,7 @@ func validateEdgesRequestInputs(targets []pb.Resource, options *edgesOptions) er
 			return fmt.Errorf("Edges cannot be returned for a specific resource name; remove %s from query", target.Name)
 		}
 		switch target.Type {
-		case "authority":
-			return fmt.Errorf("Resource type is not supported: %s", target.Type)
-		case "service":
-			return fmt.Errorf("Resource type is not supported: %s", target.Type)
-		case "all":
+		case "authority", "service", "all":
 			return fmt.Errorf("Resource type is not supported: %s", target.Type)
 		}
 	}
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -118,8 +118,6 @@ func validateEdgesRequestInputs(targets []pb.Resource, options *edgesOptions) er
 			return fmt.Errorf("Resource type is not supported: %s", target.Type)
 		case "all":
 			return fmt.Errorf("Resource type is not supported: %s", target.Type)
-		default:
-			return nil
 		}
 	}
 
--- a/cli/cmd/edges_test.go
+++ b/cli/cmd/edges_test.go
@@ -1,0 +1,152 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/linkerd/linkerd2/controller/api/public"
+)
+
+type edgesParamsExp struct {
+	options      *edgesOptions
+	resSrc       []string
+	resDst       []string
+	resClient    []string
+	resServer    []string
+	resMsg       []string
+	resourceType string
+	file         string
+}
+
+var (
+	resSrc = []string{
+		"web-57b7f9db85-297dw",
+		"web-57b7f9db85-297dw",
+		"vote-bot-7466ffc7f7-5rc4l",
+	}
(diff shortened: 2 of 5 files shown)


scottcarol:
Thanks @dadjeibaah! Dennis and I talked about this offline - since the `edges` command is modeled in part on `stat`, I modeled this test on `cli/cmd/stat_test.go` which has each test case in a separate code block. However as Dennis pointed out we use the `testCase` struct style in many of the CLI tests (such as `cli/cmd/inject_test.go`) and it does DRY up the test.

Personally the tests in separate blocks are easier for me when I'm debugging since I can easily see what was passed in, and it's easy to add/delete blocks. Would like to know others' opinions when they're back from traveling especially @rmars and @alpeb who have worked on `stat_test.go`.

cli/cmd/edges.go
```diff
@@ -118,8 +118,6 @@ func validateEdgesRequestInputs(targets []pb.Resource, options *edgesOptions) er
 			return fmt.Errorf("Resource type is not supported: %s", target.Type)
 		case "all":
 			return fmt.Errorf("Resource type is not supported: %s", target.Type)
-		default:
```
ihcsim: This `switch` can be simplified to:
```
switch target.Type {
  case "authority", "service", "all":
    return fmt.Errorf("Resource type is not supported: %s", target.Type)
}
```
Or just an `if` statement.

cli/cmd/edges_test.go
```diff
@@ -0,0 +1,152 @@
...
+	resSrc       []string
+	resDst       []string
+	resClient    []string
+	resServer    []string
+	resMsg       []string
+	resourceType string
+	file         string
+}
+
+var (
```
ihcsim: AFAICT, these variables are only used in `TestEdges(t *testing.T)`, right? I'd prefer to make them local variables inside the test function. Otherwise, these global variables are accessible by everything within the `cmd` package.

cli/cmd/edges_test.go
```diff
@@ -0,0 +1,152 @@
...
+		}, t)
+	})
+
+	t.Run("Returns an error if outputFormat specified is not table or json", func(t *testing.T) {
+		options.outputFormat = wideOutput
+		args := []string{"pod"}
+		expectedError := "--output currently only supports table and json"
+
+		_, err := buildEdgesRequests(args, options)
+		if err == nil || err.Error() != expectedError {
```
ihcsim: TIOLI:

```go
expectedError := fmt.Errorf("--output currently only supports table and json")
if _, err := buildEdgesRequests(args, options); !reflect.DeepEqual(expectedError, err) {
  t.Fatalf("Expected error [%s] instead got [%s]", expectedError, err)
}
```

cli/cmd/edges.go
scottcarol: 🙌 good point, thanks!

cli/cmd/edges_test.go
scottcarol: yup, these are only local variables. will move them inside.

cli/cmd/edges_test.go
```diff
@@ -0,0 +1,153 @@
...
+			"web.emojivoto.serviceaccount.identity.linkerd.cluster.local",
+			"web.emojivoto.serviceaccount.identity.linkerd.cluster.local",
+			"default.emojivoto.serviceaccount.identity.linkerd.cluster.local",
+		}
+		resServer = []string{
+			"emoji.emojivoto.serviceaccount.identity.linkerd.cluster.local",
+			"voting.emojivoto.serviceaccount.identity.linkerd.cluster.local",
+			"web.emojivoto.serviceaccount.identity.linkerd.cluster.local",
+		}
+		resMsg = []string{"", "", ""}
```
rmars: tioli: may as well have some messages in there to test both blank and not blank rendering?

scottcarol: Thanks @rmars, currently the `edges` command is only returning meshed traffic with known identity, once it has the capacity to return requests that have no identity I will add to the test.

PR #2842, opened 2019-05-22
Add debug container annotation

### LINKED ISSUE #2840: --enable-debug-sidecar only works with --manual
The `--enable-debug-sidecar` flag does not set annotations, so this can't be used with proxy-injector:

```
:; linkerd inject --enable-debug-sidecar app.yml 2>/dev/null | grep config.linkerd.io 
```

This works, however, when `--manual` is specified.

This should work with auto injection.

ihcsim:
This  PR adds a new `config.linkerd.io/debug` annotation. When added to the pod's metadata, the proxy injector will inject the debug container into the workload's containers list. This ensures that `linkerd inject --enable-debug-sidecar` works in an auto-inject setup.

The reason why this new annotation is needed it's because when using `linkerd install`, the `pkg/inject` library will only inject Linkerd annotations into the workload YAML. So even though the [`conf.debugSidecar` field is set by the CLI](https://github.com/linkerd/linkerd2/blob/bd4c2788fa19f31b0fb2f32254ac6fd52c737565/cli/cmd/inject.go#L138-L140), the `inject.injectPodSpec()` function is never called. Once the workload YAML got picked up by the proxy injector, `conf.debugSidecar` is already nil, since it's a different, new `conf` object.

Fixes #2840 

### DIFF
--- a/cli/cmd/testdata/inject_emojivoto_deployment_debug.golden.yml
+++ b/cli/cmd/testdata/inject_emojivoto_deployment_debug.golden.yml
@@ -13,6 +13,7 @@ spec:
   template:
     metadata:
       annotations:
+        config.linkerd.io/debug: "true"
         linkerd.io/created-by: linkerd/cli dev-undefined
         linkerd.io/identity-mode: default
         linkerd.io/proxy-version: test-inject-proxy-version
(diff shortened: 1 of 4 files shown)


ihcsim:
@alpeb Thanks for the review. Re the annotation name, is the concern that this might be mistaken for the debug level? I think `debug` is fine (or `debug-mode`?), just as we use `inject`, not `inject-proxy` or `inject-sidecar`.

ihcsim:
@alpeb @dadjeibaah Thanks again for the review. I fixed a mistake in my code where I left out the call to the `WithDebugSidecar()` method. 

I think for any future changes to this `ResourceConfig` struct, we should consider revisiting its structure and usages. At the moment, depending on which component is calling it, the object gets modified in the `cli`, `controller/proxy-injector`, `controller/sp-validator` and `pkg/inject` packages. And there are no simple ways to tell which `With*` methods have been called; hence, leading to issue like #2840.

ihcsim:
Integration test passed on my machine.
```
=== PASS: all tests passed
```

PR #2853, opened 2019-05-28
Added labels to webhook configurations in charts/

### LINKED ISSUE #2602: Make sure all the components are labeled with "linkerd.io/control-plane-component"

Pothulapati:
Fixes #2602 

Added Controller component labels for webhook configurations.


ihcsim:
Oppss.. I missed this by a minute. Not a big deal, but the next time we are here, we should use the `{{.ControllerComponentLabel}}` template variable.

PR #2854, opened 2019-05-28
Generate the debug container spec in the shared library

ihcsim:
This commit refactors the changes introduced by #2842 where the debug container spec is created in the `cli` and `pkg` packages. This change follows the existing pattern of annotating the YAML in the CLI code, and injecting the sidecar spec in the shared library.

Test cases:
```
$ curl https://run.linkerd.io/emojivoto.yml | bin/linkerd inject --enable-debug-sidecar --manual -|k apply -f -

$ curl https://run.linkerd.io/emojivoto.yml | bin/linkerd inject --enable-debug-sidecar -|k apply -f -
```
Expect debug container to be injected.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -145,18 +145,6 @@ func NewResourceConfig(configs *config.All, origin Origin) *ResourceConfig {
 	return config
 }
 
-// WithDebugSidecar enriches ResourceConfig with a debug sidecar for a resource
-func (conf *ResourceConfig) WithDebugSidecar() *ResourceConfig {
-	conf.debugSidecar = &corev1.Container{
-		Name:                     k8s.DebugSidecarName,
-		ImagePullPolicy:          conf.proxyImagePullPolicy(),
-		Image:                    fmt.Sprintf("%s:%s", k8s.DebugSidecarImage, conf.configs.GetGlobal().GetVersion()),
-		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
-	}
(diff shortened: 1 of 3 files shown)


pkg/inject/inject.go
```diff
@@ -427,16 +414,15 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
 	}
 
 	if v := conf.pod.meta.Annotations[k8s.ProxyEnableDebugAnnotation]; v != "" {
-		debugEnabled, err := strconv.ParseBool(v)
+		debug, err := strconv.ParseBool(v)
 		if err != nil {
 			log.Warnf("unrecognized value used for the %s annotation: %s", k8s.ProxyEnableDebugAnnotation, v)
-			debugEnabled = false
+			debug = false
```
dadjeibaah: I think we don't need to have this here since `strconv.ParseBool(v)` already sets debug to false if it fails.

dadjeibaah: Also, what's the reasoning for changing the variable name from `debugEnabled` to `debug`?

olix0r: > I think we don't need to have this here since strconv.ParseBool(v) already sets debug to false if it fails.

I don't think we want to rely on that behavior.. In general, when `err` is non-nil, you shouldn't assume any semantics about the returned value.

ihcsim: > Also, what's the reasoning for changing the variable name

I meant to update that in my last PR, to match the annotation.

Issue #2103, opened 2019-01-17
document (or remove) Authorities table


wmorgan:
The "Authorities" table can cause confusion in users. I propose either a) adding a tooltip that says something along the lines of "The name that the service was called as. Equivalent to the Host header for HTTP/1.1 requests, or the :authority psuedo-header for HTTP/2 requests", or b) removing this section entirely.

grampelberg:
@scottcarol I'd love to make some progress on this issue in the next release or so. Want to do some brainstorming about how to make it better? @adleong has some great ideas.

scottcarol:
@grampelberg Can do - @adleong I'll ping you.

Issue #2855, opened 2019-05-28
Support SMI traffic split API


adleong:
Linkerd should implement the [SMI traffic split API](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md) by keeping a watch on the traffic split CRD and routing traffic accordingly.

wxdao:
Will it support A/B testing? i.e traffic split based on headers/cookies.

grampelberg:
@wxdao not currently, please open an issue in https://github.com/deislabs/smi-spec to track adding that functionality to TrafficSplit.

adleong:
Fixed by #2931

PR #2856, opened 2019-05-28
Rename "Authorities" table in dashboard to "Services" for readability

scottcarol:
Currently the `Authorities` table in the dashboard can be confusing due to users not necessarily knowing what `authority` and `authority name` means, and/or how to act on this information. #2103 suggests documenting or removing that table entirely. 

@adleong suggested that one solution could be renaming the table to `Services` and displaying the service name instead of the authority name. We could see if this is useful before resorting to removing the table.

This is a small change which 
- renames the `Authorities` table to `Services`. (NOTE: `authority`, `isAuthorityTable`, etc. are retained as internal value names, as this is mostly a cosmetic change to try for better UX.)
- instead of authority name `emoji-svc.emojivoto.svc.cluster.local:8080` the `Services` table displays service name `emoji-svc`  
- renames the "Resources" sidebar view from Authorities to Services

Before, in namespace `emojivoto`:
Screen Shot 2019-05-28 at 11 00 24 AM

After, in namespace `emojivoto`:
Screen Shot 2019-05-28 at 10 55 03 AM

### DIFF
--- a/web/app/js/components/MetricsTable.jsx
+++ b/web/app/js/components/MetricsTable.jsx
@@ -128,7 +128,9 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink, isTcpTab
       let nameContents;
       if (resource === "namespace") {
         nameContents = <PrefixedLink to={"/namespaces/" + d.name}>{d.name}</PrefixedLink>;
-      } else if (!d.added || isAuthorityTable) {
+      } else if (isAuthorityTable) {
+        nameContents = getResourceDisplayName(d).split('.')[0];
+      } else if (!d.added) {
         nameContents = getResourceDisplayName(d);
       } else {
         nameContents = (
--- a/web/app/js/components/MetricsTable.jsx
+++ b/web/app/js/components/MetricsTable.jsx
@@ -128,9 +128,7 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink, isTcpTab
       let nameContents;
       if (resource === "namespace") {
         nameContents = <PrefixedLink to={"/namespaces/" + d.name}>{d.name}</PrefixedLink>;
-      } else if (isAuthorityTable) {
-        nameContents = getResourceDisplayName(d).split('.')[0];
-      } else if (!d.added) {
+      } else if (!d.added || isAuthorityTable) {
         nameContents = getResourceDisplayName(d);
       } else {
         nameContents = (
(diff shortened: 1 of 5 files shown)


olix0r:
To be clear, this will only ever show Kubernetes services, right? I.e., where the Authorities section could show data for arbitrary hosts like api.github.com, the Services menu will show nothing. (I think that's a fine trade-off, I just want to understand the specifics)

web/app/js/components/MetricsTable.jsx
```diff
@@ -128,7 +128,9 @@ const columnDefinitions = (resource, showNamespaceColumn, PrefixedLink, isTcpTab
       let nameContents;
       if (resource === "namespace") {
         nameContents = <PrefixedLink to={"/namespaces/" + d.name}>{d.name}</PrefixedLink>;
-      } else if (!d.added || isAuthorityTable) {
+      } else if (isAuthorityTable) {
+        nameContents = getResourceDisplayName(d).split('.')[0];
```
adleong: What happens for non-service authorities?  For example, when Prometheus calls directly to a pod ip, would this display the first octect of the IP address?  e.g. if the authority is `5.6.7.8:99` then would this display `5` as the service name?  

One idea would be to only reformat names with hosts that end in `svc.cluster.local` and keep all other names as-is.

adleong: Or omit non-`svc.cluster.local` entries entirely.

scottcarol:
Update, after conversation with @olix0r @adleong, I am temporarily going to hide the `Authorities` table in the dashboard instead of making this change, and work on a new view incorporating data from #2855 in the next few weeks. Will push a new commit reflecting this.

scottcarol:
Update: the original commit had renamed the "Authorities" table to "Services" and displayed the service name. The latest commit reverts those changes and instead simply hides the Authorities table in:
* Overview page of Dashboard (there is usually one in each Namespace section)
* Namespace view
* `Resources` sidebar does not have `Authorities` link (an `Authorities` page will still be rendered at `http://localhost:7777/authorities`

This will solve the original problem of #2103 (confusing and undocumented "Authorities" table). The next step is to create mocks for a dashboard view of the soon-to-come traffic split info which will incorporate the Authorities data in a more easily-understandable way.

scottcarol:
Hey @alpeb, thanks for taking a look! Yes, this is just hiding the specific "Authorities" table and "Authorities" view in the dashboard due to the table being confusing for users, to make way for a better view of service/authority data when we implement the SMI traffic split API (#2855). 

Re: Grafana, I'd prefer to leave as-is since those graphs are customizable by users and I want them to have the authorities data if they already digging through the data, but there could definitely be new Grafana graphs of the traffic split view as well when that happens.

PR #2858, opened 2019-05-28
Make JSON output of `linkerd edges` consistent

scottcarol:
This PR updates `printEdgesJSON` to iterate over sorted keys instead of iterating over a map. This will make the ordering of the resulting JSON consistent and allow the associated unit test to pass. For reference, look at `printEdgeTable` which also iterates over sorted keys.

### DIFF
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -181,7 +181,6 @@ func renderEdgeStats(rows []*pb.Edge, options *edgesOptions) string {
 }
 
 type edgeRow struct {
-	key    string
 	src    string
 	dst    string
 	client string
@@ -207,7 +206,6 @@ func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *edgesOpti
 	edgeRows := []edgeRow{}
 	if len(rows) != 0 {
 		for _, r := range rows {
-			key := r.Src.Name + r.Dst.Name
 			clientID := r.ClientId
 			serverID := r.ServerId
 			msg := r.NoIdentityMsg
@@ -225,7 +223,6 @@ func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *edgesOpti
 			}
 
 			row := edgeRow{
-				key:    key,
 				client: clientID,
 				server: serverID,
 				msg:    msg,
@@ -255,7 +252,9 @@ func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *edgesOpti
 
 	// sorting edgeRows by key for alphabetical listing
 	sort.Slice(edgeRows, func(i, j int) bool {
-		return edgeRows[i].key < edgeRows[j].key
+		keyI := edgeRows[i].src + edgeRows[i].dst
+		keyJ := edgeRows[j].src + edgeRows[j].dst
+		return keyI < keyJ
 	})
 
 	switch options.outputFormat {
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -329,17 +329,18 @@ type edgesJSONStats struct {
 	Msg    string `json:"no_tls_reason"`
 }
 
-func printEdgesJSON(edgeTables map[string]*edgeRow, w *tabwriter.Writer) {
+func printEdgesJSON(rows map[string]*edgeRow, w *tabwriter.Writer) {
 	// avoid nil initialization so that if there are not stats it gets marshalled as an empty array vs null
 	entries := []*edgesJSONStats{}
 
-	for _, row := range edgeTables {
+	sortedKeys := sortEdgesKeys(rows)
+	for _, key := range sortedKeys {
 		entry := &edgesJSONStats{
-			Src:    row.src,
-			Dst:    row.dst,
-			Client: row.client,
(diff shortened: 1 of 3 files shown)


scottcarol:
Thanks @adleong, these are good points. Let me see if I can reduce the number of conversions going on here.

scottcarol:
Hey @adleong, thanks again for thoughtful feedback! In looking more into this, it seems like there are a few things going on:

* in `writeEdgesToBuffer` for each `r (*Edge)` we create a key that is the SRC name plus the DST name - there may be many rows with the same SRC (or DST) name but there will only one row for a particular SRC-DST pairing, so that’s the naming formula I chose
* we want those keys to be consistently ordered and alphabetical so in `sortEdgesKeys` we create an ordered slice of keys
* when rendering our table/JSON, we range through each item in the slice so that we’re building in the right order, and for each of those keys, we reference the key/value pairs in our original map 

So, we are not converting the map to a slice, we are creating a new slice from the map. We use both of them when rendering our tables. 

I think the map structure is pretty useful, and so is the slice. If we do stop using the map structure we lose the benefit of our key/value pairs and would need to restructure a lot.

I agree with your idea of DRYing up by only calling `sortEdgesKeys` once, and passing those sorted keys into the render functions, and I did that in my latest commit, let me know what you think.

cli/cmd/edges.go
```diff
@@ -206,7 +206,7 @@ func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *edgesOpti
 	edgeTables := make(map[string]*edgeRow)
 	if len(rows) != 0 {
 		for _, r := range rows {
-			key := r.Dst.Name + r.Src.Name
+			key := r.Src.Name + r.Dst.Name
```
scottcarol: Note, I switched these around because alphabetizing by `SRC` makes more sense in the table output since `SRC` is the left-most column.

ihcsim:
(IIUC) I'm wondering instead of sorting the slice of keys, then using them to order the tables, can you just sort the `rows` slice right before you call `writeEdgesToBuffer()`? You can actually implement the `sort.Interface` interface on a custom type based on the `[]*pd.Edges` type.

Here's a TIOLI:
```diff
diff --git a/cli/cmd/edges.go b/cli/cmd/edges.go
index a43a7d18..43193d79 100644
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -33,6 +33,30 @@ type indexedEdgeResults struct {
 	err  error
 }
 
+type edges []*pb.Edge
+
+// Len returns the number edges in the collection
+func (e edges) Len() int {
+	return len(e)
+}
+
+// Less returns true if the element with index i should sort before the element
+// with index j
+func (e edges) Less(i, j int) bool {
+	var (
+		k1 = e[i].Src.Name + e[i].Dst.Name
+		k2 = e[j].Src.Name + e[j].Dst.Name
+	)
+	return k1 < k2
+}
+
+// Swap swaps the elements with indexes i and i
+func (e edges) Swap(i, j int) {
+	tmp := e[j]
+	e[j] = e[i]
+	e[i] = tmp
+}
+
 func newCmdEdges() *cobra.Command {
 	options := newEdgesOptions()
 
@@ -92,7 +116,9 @@ func newCmdEdges() *cobra.Command {
 				}
 			}
 
-			output := renderEdgeStats(totalRows, options)
+			sorted := edges(totalRows)
+			sort.Sort(sorted)
+			output := renderEdgeStats(sorted, options)
 			_, err = fmt.Print(output)
 
 			return err
```

scottcarol:
Thanks @ihcsim, just talked to @adleong and am working on simplifying, your feedback is super helpful as well! 💯

adleong:
I think you can also just use [sort.Slice](https://golang.org/pkg/sort/#Slice) which should be easier than implementing `sort.Interface`.

scottcarol:
Thanks @adleong, I'm using `sort.Slice` now

cli/cmd/edges.go
```diff
@@ -246,22 +250,27 @@ func writeEdgesToBuffer(rows []*pb.Edge, w *tabwriter.Writer, options *edgesOpti
 			if len(msg) > maxMsgLength {
 				maxMsgLength = len(msg)
 			}
-
 		}
 	}
+
+	// sorting edgeRows by key for alphabetical listing
+	sort.Slice(edgeRows, func(i, j int) bool {
+		return edgeRows[i].key < edgeRows[j].key
```
adleong: TIOLI: since this is the only place that `key` is used, you could omit adding it as a field on the struct and just compute it here directly.  This has the advantage of keeping all the sorting logic together in one place.

scottcarol: Do you mean something like this? 
```
	sort.Slice(edgeRows, func(i, j int) bool {
		keyI := edgeRows[i].src + edgeRows[i].dst
		keyJ := edgeRows[j].src + edgeRows[j].dst
		return keyI < keyJ
	})
```
This would mean recalculating `keyI` and `keyJ` multiple times but it prevents passing `key` down unnecessarily. 👍

adleong: Yup!  Unless we expect the size of `edgeRows` to be large, I wouldn't worry about extra calculations

PR #2865, opened 2019-05-29
Known event warnings should hold just the message part, not the reason nor t…

alpeb:
…he involved object name

This was causing integration tests to fail sporadically.

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -85,7 +85,7 @@ var (
 		`(Liveness|Readiness) probe failed: Get http://.*: dial tcp .*: connect: connection refused`,
 		`(Liveness|Readiness) probe failed: Get http://.*: read tcp .*: read: connection reset by peer`,
 		`(Liveness|Readiness) probe failed: Get http://.*: net/http: request canceled \(Client\.Timeout exceeded while awaiting headers\)`,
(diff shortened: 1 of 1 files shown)


scottcarol:
👍 This looks good @alpeb, for my own understanding, would you mind explaining why the previous known event warnings were causing the integration tests to fail sporadically?

alpeb:
@scottcarol The culprit event warning was:
```
FailedToUpdateEndpoint linkerd-web Failed to update endpoint l5d-integration-auto-git-013d2a16-upgrade/linkerd-web: Operation cannot be fulfilled on endpoints "linkerd-web": the object has been modified; please apply your changes to the latest version and try again
```
which appears to be benign, so it was included in the list of _expected_ event warnings. But such list should only contain the message portion of that warning, which is what starts with "Failed to update..." (the "reason" being "FailedToUpdateEndpoint" and the "object name" being "linkerd-web").

We output the entire thing in the test logs, but only care about the message part:
https://github.com/linkerd/linkerd2/blob/031cd9d4baa821c8b8ea4781eb557b57b371ecb6/test/install_test.go#L542-L547

scottcarol:
Thanks for the explanation! 👏

Issue #2553, opened 2019-03-24
destination controller panics during shutdown


olix0r:
Running an AKS cluster, I noticed one of my controller instances in crashloopbackoff; and it appears that during shutdown the controller just panics:
```
:; kubectl logs linkerd-controller-695dcbdcf7-9bsvp  -c destination -n linkerd
time="2019-03-24T20:54:37Z" level=info msg="running version edge-19.3.3"
time="2019-03-24T20:54:37Z" level=info msg="Built k8s name resolver"
time="2019-03-24T20:54:37Z" level=info msg="waiting for caches to sync"
time="2019-03-24T20:54:37Z" level=info msg="caches synced"
time="2019-03-24T20:54:37Z" level=info msg="starting admin server on :9996"
time="2019-03-24T20:54:37Z" level=info msg="starting gRPC server on :8086"
time="2019-03-24T20:54:37Z" level=info msg="Establishing watch on endpoint linkerd-prometheus.linkerd:9090" component=endpoints-watcher
time="2019-03-24T20:54:37Z" level=info msg="Establishing watch on endpoint linkerd-grafana.linkerd:3000" component=endpoints-watcher
time="2019-03-24T20:54:37Z" level=info msg="Establishing watch on endpoint emoji-svc.emojivoto:8080" component=endpoints-watcher
time="2019-03-24T20:54:37Z" level=info msg="Establishing watch on endpoint linkerd-controller-api.linkerd:8085" component=endpoints-watcher
time="2019-03-24T20:54:37Z" level=info msg="Establishing watch on endpoint voting-svc.emojivoto:8080" component=endpoints-watcher
time="2019-03-24T20:55:16Z" level=info msg="shutting down gRPC server on :8086"
panic: close of closed channel

goroutine 147 [running]:
github.com/linkerd/linkerd2/controller/api/destination.(*profileListener).Stop(0xc0009f94c0)
	/go/src/github.com/linkerd/linkerd2/controller/api/destination/profile_listener.go:39 +0x2f
github.com/linkerd/linkerd2/controller/api/destination.(*fallbackChildListener).Stop(0xc000572790)
	/go/src/github.com/linkerd/linkerd2/controller/api/destination/fallback_profile_listener.go:65 +0x37
github.com/linkerd/linkerd2/controller/api/destination.(*profileEntry).unsubscribeAll(0xc000b9c090)
[...]

PR #2867, opened 2019-05-30
Fixed controller panic due to dup listener #2553

tammert:
Fixes (or at least provides a workaround) for #2553.

~~There is something weird happening in the primary/backup listener logic, it seems like both listeners are the same. Because of this, ranging over and then stopping the listeners causes a closed channel (from the first duplicate) to be closed again in the second duplicate. This causes the panic.~~

~~I've added a TODO in the newFallbackProfileListener() func, might be good for someone more experienced (maybe original author @adleong) to see if my observations are correct.~~

EDIT: nevermind, I was confused for a bit. The problem was with both the primary and backup listener having a pointer to the same parent `fallbackProfileListener`, which had `.Stop()` called on it for both the primary and backup. Since `.Stop()` closes a channel, this resulted in closing that channel twice, which caused the panic.

Signed-off-by: Sebastiaan Tammer <sebastiaantammer@gmail.com>

### DIFF
--- a/controller/api/destination/fallback_profile_listener.go
+++ b/controller/api/destination/fallback_profile_listener.go
@@ -39,12 +39,12 @@ func newFallbackProfileListener(listener profileUpdateListener) (profileUpdateLi
 	}
 
 	primary := primaryProfileListener{
-		fallbackChildListener{
+		fallbackChildListener: fallbackChildListener{
 			parent: &fallback,
 		},
 	}
 	backup := backupProfileListener{
-		fallbackChildListener{
+		fallbackChildListener: fallbackChildListener{
 			parent: &fallback,
 		},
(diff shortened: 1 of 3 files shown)


tammert:
@olix0r this PR is ready for review, fixes #2553.

adleong:
Hi @tammert, we just recently merged a very substantial refactor to this code.  In particular, the `Stop` method has been removed completely.  It is possible this issue was fixed as a side-effect of that change.  Please take a look at the current code and let me know what you think.

tammert:
@adleong yea, the code differs quite a lot from what I was working on! This PR is definitely not valid anymore, and I'm pretty sure #2553 is no longer valid either :). I'm closing the PR.

PR #2868, opened 2019-05-30
proxy: Update to linkerd/linkerd2-proxy#ed32e496

olix0r:
This changes the proxy's routing behavior so that, when the control plane does not resolve a destination, the proxy forwards request with minimal additional routing logic.

Furthermore, this fixes a bug in the proxy's HPACK codec that could cause requests with very large header values to hang indefinitely.

    b3dcc6e0 Use the proxy's log formatting in tests (linkerd/linkerd2-proxy#258)
    1c91a398 Rewrite the destination client and remove DNS fallback (linkerd/linkerd2-proxy#259)
    ed32e496 Update h2 to v0.1.21 (linkerd/linkerd2-proxy#261)

### DIFF
--- a/bin/docker-build-proxy
+++ b/bin/docker-build-proxy
@@ -23,7 +23,7 @@ validate_go_deps_tag $dockerfile
 ) >/dev/null
 
 # Default to a pinned commit SHA of the proxy.
-PROXY_VERSION="${PROXY_VERSION:-3e0e00c}"
+PROXY_VERSION="${PROXY_VERSION:-ed32e49}"
 
 tag="$(head_root_tag)"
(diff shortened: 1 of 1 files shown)


olix0r:
> lgtm, but when it's merged I think using "Update Proxy SHA to latest master" is clearer as the title ￼

@scottcarol fwiw, i disagree. "proxy SHA to latest master" is something that will end up in the commit log once a week or so, and doesn't distinguish the update. I prefer to reference the commit that we're updating against so it's easy to know what the change was from the commit log.

hawkw:
The other nice thing about referencing the specific commit is that if it's referred to as `linkerd2/linkerd2-proxy#(SHA)`, GitHub will turn it into a link automatically.

PR #2869, opened 2019-05-30
Stable 2.3.1 Release Notes

dadjeibaah:
This stable release adds a number of proxy stability improvements.

To install this release, run: `curl https://run.linkerd.io/install | sh`

**Special thanks to**: @zaharidichev and @11Takanori!

**Full release notes**:

* Proxy
  * Changed the proxy's routing behavior so that, when the control plane
    does not resolve a destination, the proxy forwards the request with minimal
    additional routing logic
  * Fixed a bug in the proxy's HPACK codec that could cause requests with
    very large header values to hang indefinitely
  * Replaced the fixed reconnect backoff with an exponential one (thanks,
    @zaharidichev!)
  * Fixed an issue where requests could be held indefinitely by the load balancer
  * Added a dispatch timeout that limits the amount of time a request can be
    buffered in the proxy
  * Removed the limit on the number of concurrently active service discovery
    queries to the destination service
  * Fixed an epoll notification issue that could cause excessive CPU usage
  * Added the ability to disable tap by setting an env var (thanks,
    @zaharidichev!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 ## stable-2.3.1
+
 This stable release adds a number of proxy stability improvements.
 
 To install this release, run: `curl https://run.linkerd.io/install | sh`
@@ -6,22 +7,23 @@ To install this release, run: `curl https://run.linkerd.io/install | sh`
 **Special thanks to**: @zaharidichev and @11Takanori!
 
 **Full release notes**:
+
 * Proxy
   * Changed the proxy's routing behavior so that, when the control plane
-    does not resolve a destination, the proxy forwards request with minimal
+    does not resolve a destination, the proxy forwards the request with minimal
     additional routing logic
   * Fixed a bug in the proxy's HPACK codec that could cause requests with
-    very large header values to hang indefinitely.
+    very large header values to hang indefinitely
   * Replaced the fixed reconnect backoff with an exponential one (thanks,
       @zaharidichev!)
   * Fixed an issue where load balancers can become stuck
   * Added a dispatch timeout that limits the amount of time a request can be
     buffered in the proxy
   * Removed the limit on the number of concurrently active service discovery
-    queries to the Destination service
+    queries to the destination service
   * Fix an epoll notification issue that could cause excessive CPU usage
   * Added the ability to disable tap by setting an env var (thanks,
-      @zaharidichev!)
+    @zaharidichev!)
 
 ## edge-19.5.3
 
@@ -54,7 +56,7 @@ To install this release, run: `curl https://run.linkerd.io/install | sh`
   * Added a dispatch timeout that limits the amount of time a request can be
     buffered in the proxy
   * Removed the limit on the number of concurrently active service discovery
-    queries to the Destination service
+    queries to the destination service
 
 Special thanks to @zaharidichev for adding end to end tests for proxies with
 TLS!
@@ -1357,7 +1359,7 @@ formerly hosted at github.com/runconduit/conduit.
   * Update branding to reference Linkerd throughout
   * The CLI is now called `linkerd`
 * Production Readiness
-  * Fix issue with Destination service sending back incomplete pod metadata
+  * Fix issue with destination service sending back incomplete pod metadata
   * Fix high CPU usage during proxy shutdown
   * ClusterRoles are now unique per Linkerd install, allowing multiple instances
     to be installed in the same Kubernetes cluster
@@ -1509,7 +1511,7 @@ Kubernetes-aware observability and debugging.
 * Service Discovery
   * The proxy now uses the [trust-dns] DNS resolver. This fixes a number of DNS
     correctness issues.
-  * The Destination service could sometimes return incorrect, stale, labels for an
+  * The destination service could sometimes return incorrect, stale, labels for an
     endpoint. This has been fixed!
 
 [trust-dns]: https://github.com/bluejekyll/trust-dns
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,11 @@ To install this release, run: `curl https://run.linkerd.io/install | sh`
 
 **Full release notes**:
 * Proxy
+  * Changed the proxy's routing behavior so that, when the control plane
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,28 @@
+## stable-2.3.1
+This stable release adds a number of proxy stability improvements.
```
klingerf: Need an extra new line here between the header and the first paragraph.

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+
+**Full release notes**:
+* Proxy
+  * Changed the proxy's routing behavior so that, when the control plane
+    does not resolve a destination, the proxy forwards request with minimal
+    additional routing logic
+  * Fixed a bug in the proxy's HPACK codec that could cause requests with
+    very large header values to hang indefinitely.
+  * Replaced the fixed reconnect backoff with an exponential one (thanks,
+      @zaharidichev!)
```
klingerf: You can remove 2 leading spaces from this line, to match to formatting of the other bullet points.

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+  * Fixed an issue where load balancers can become stuck
+  * Added a dispatch timeout that limits the amount of time a request can be
+    buffered in the proxy
+  * Removed the limit on the number of concurrently active service discovery
+    queries to the Destination service
+  * Fix an epoll notification issue that could cause excessive CPU usage
+  * Added the ability to disable tap by setting an env var (thanks,
+      @zaharidichev!)
```
klingerf: Same goes for this line.

CHANGES.md
```diff
@@ -1,3 +1,28 @@
+## stable-2.3.1
+This stable release adds a number of proxy stability improvements.
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Special thanks to**: @zaharidichev and @11Takanori!
+
+**Full release notes**:
+* Proxy
```
klingerf: Let's add an additional new line before this first bullet point as well.

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+  * Changed the proxy's routing behavior so that, when the control plane
+    does not resolve a destination, the proxy forwards request with minimal
```
admc: should 'request' be plural?

scottcarol: should this be `proxy forwards the request`?

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+    additional routing logic
+  * Fixed a bug in the proxy's HPACK codec that could cause requests with
+    very large header values to hang indefinitely.
```
scottcarol: you can remove period from the end of this sentence

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+  * Replaced the fixed reconnect backoff with an exponential one (thanks,
+      @zaharidichev!)
+  * Fixed an issue where load balancers can become stuck
+  * Added a dispatch timeout that limits the amount of time a request can be
+    buffered in the proxy
+  * Removed the limit on the number of concurrently active service discovery
+    queries to the Destination service
```
admc: total nit, but Capitalization of "Destination service" seems strange, maybe either both caps or neither? If that is indeed correct, feel free to ignore.

scottcarol: I was curious about this too as I'm drafting edge release notes. Right now there are 3 "Destination service" mentions and 3 "destination service" mentions so Dennis I guess you are the tie-breaker? :)

dadjeibaah: This is too much pressure haha. No caps feels much better?

dadjeibaah:
Thanks for the feedback folks, changes coming up in the next commit.

CHANGES.md
```diff
@@ -1,3 +1,30 @@
...
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Special thanks to**: @zaharidichev and @11Takanori!
+
+**Full release notes**:
+
+* Proxy
+  * Changed the proxy's routing behavior so that, when the control plane
+    does not resolve a destination, the proxy forwards the request with minimal
```
hawkw: IMO "forwards the request to its original destination" better describes the behaviour than "forwards the request with minimal additional routing logic".

CHANGES.md
```diff
@@ -1,3 +1,30 @@
...
+  * Fixed a bug in the proxy's HPACK codec that could cause requests with
+    very large header values to hang indefinitely
+  * Replaced the fixed reconnect backoff with an exponential one (thanks,
+      @zaharidichev!)
+  * Fixed an issue where load balancers can become stuck
+  * Added a dispatch timeout that limits the amount of time a request can be
+    buffered in the proxy
+  * Removed the limit on the number of concurrently active service discovery
+    queries to the destination service
+  * Fix an epoll notification issue that could cause excessive CPU usage
```
hawkw: This is in present tense, while it looks like all the other bullet points are in past-tense. 
```suggestion
  * Fixed an epoll notification issue that could cause excessive CPU usage
```

CHANGES.md
dadjeibaah: I think "forwards the request with minimal additional routing logic" works because it implies there is more going on behind the scenes w.r.t routing and that the behavior has now been fixed.

CHANGES.md
```diff
@@ -1,3 +1,30 @@
...
+
+* Proxy
+  * Changed the proxy's routing behavior so that, when the control plane
+    does not resolve a destination, the proxy forwards the request with minimal
+    additional routing logic
+  * Fixed a bug in the proxy's HPACK codec that could cause requests with
+    very large header values to hang indefinitely
+  * Replaced the fixed reconnect backoff with an exponential one (thanks,
+      @zaharidichev!)
+  * Fixed an issue where load balancers can become stuck
```
olix0r: Perhaps: _Fixed an issue where requests could be held indefinitely by the load balancer_ ?

"stuck" is a bit squishy

dadjeibaah: Sounds great. Thanks for the feedback.

PR #2870, opened 2019-05-30
Ensure log level is lowercase (#2823)

### LINKED ISSUE #2823: Improve controller log level validation
The `linkerd install` command performs validation of the `--controller-log-level` flag to ensure only valid log levels are passed in.  However, Prometheus has stricter requirements on the log level string and requires that it be lowercase.  Passing `DEBUG` as the value passes Linkerd's validation, but the Prometheus container fails with:

```
$ k logs -n linkerd linkerd-prometheus-cd49d97f9-27mqq prometheus
Error parsing commandline arguments: unrecognized log level "DEBUG"
```

We should tighten the validation to require that the log level be lowercase, or else convert it to lowercase before passing to Prometheus.

d-baranowski:
linkerd accepts correct log level in upper case and they are then passed
into prometheus which only allows lowecase log level with error

Error parsing commandline arguments: unrecognized log level "DEBUG"

More strict validation

Updated the unit tests

Fixes #2823

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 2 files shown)


d-baranowski:
Ready for review I think :)

jonrichards:
The logging library handles this case by forcing the log level to lower case: [https://github.com/sirupsen/logrus/blob/4f5fd631f16452fbd023813c1eb7dbd67130cb0c/logrus.go#L26](https://github.com/sirupsen/logrus/blob/4f5fd631f16452fbd023813c1eb7dbd67130cb0c/logrus.go#L26). Maybe the same approach could be done for `controllerLogLevel` when sending to Prometheus?

klingerf:
👍 to @jonrichards comments -- I'd prefer to lowercase the string automatically, rather than return an error for uppercase strings.

d-baranowski:
I'm on it.

d-baranowski:
I updated the buildValuesWithoutIdentity function to ensure that PrometheusLogLevel is converted to lower case. The value of ControllerLogLevel remains unchanged. Is that ok?

d-baranowski:
?

PR #2874, opened 2019-05-31
HA for proxy-injector and sp-validator

### LINKED ISSUE #2873: Run multiple replicas of the proxy-injector and sp-validator in HA mode
Now that #2806 has shipped, the proxy-injector and sp-validator processes should be stateless, which means that we can run more than one replica for each of those deployments when running the control plane in HA mode. Rather than adding separate install flags for configuring the proxy-injector and sp-validator replicas, I'd prefer for us to just run the same number of replicas as we do for the controller deployment, as determined by the existing `--controller-replicas` and `--ha` flags.

Pothulapati:
Fixes #2873 

This PR also adds `ControllerComponentLabel` template variable for webhook configurations.

@ihcsim @klingerf

### DIFF
--- a/bin/docker-build-proxy
+++ b/bin/docker-build-proxy
@@ -23,7 +23,7 @@ validate_go_deps_tag $dockerfile
 ) >/dev/null
 
 # Default to a pinned commit SHA of the proxy.
-PROXY_VERSION="${PROXY_VERSION:-3e0e00c}"
+PROXY_VERSION="${PROXY_VERSION:-ed32e49}"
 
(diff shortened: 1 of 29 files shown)


alpeb:
Hi Tarun, can I ask you to please rebase instead of merge, whenever you'd like to bring your branch in-sync with master from now on? It's a little easier to review like that, and when this gets merged back to master, it gets merged as a single commit.

Pothulapati:
Oops Sorry! Will definitely rebase from now on!

PR #2737, opened 2019-04-23
Implement fallback logic for owner ref lookups

siggy:
The proxy-injector retrieves owner information when injecting pods. For
pods created via deployments, this requires a Pod -> ReplicaSet ->
Deployment lookup. There is a race condition where the injection happens
before the k8s informer client has indexed the new ReplicaSet.

If a ReplicaSet informer lookup initially fails, retry one time via a
get request. Also introduce logging to record the failure/retry, and
tests to validate `GetOwnerKindAndName` works with and without informer
indexing.

Fixes #2731

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

## Log of failure/retry

```bash
kubectl -n linkerd logs -f deploy/linkerd-proxy-injector proxy-injector
...
time="2019-04-23T06:26:30Z" level=info msg="received admission review request bb27118b-6590-11e9-a508-025000000001"
time="2019-04-23T06:26:30Z" level=error msg="failed to retrieve replicaset from indexer emojivoto-enabled/vote-bot-dc4bd6984: replicaset.apps \"vote-bot-dc4bd6984\" not found"
time="2019-04-23T06:26:30Z" level=info msg="successfully recovered replicaset via k8s get request: emojivoto-enabled/vote-bot-dc4bd6984"
time="2019-04-23T06:26:30Z" level=info msg="received pod/vote-bot-dc4bd6984-"
time="2019-04-23T06:26:30Z" level=info msg="patch generated for: pod/vote-bot-dc4bd6984-"
```

Issue #2731, opened 2019-04-22
auto-injector setting inconsistent parent label


siggy:
## Summary

Using the proxy auto-injector, some pods are getting a `linkerd.io/proxy-replicaset` label while others are getting `linkerd.io/proxy-deployment`.

### Install control-plane

```bash
linkerd install --proxy-auto-inject` # stable-2.3.0
```

### Inject annotations and deploy

Note: annotations injected via #2721

```bash
curl https://run.linkerd.io/emojivoto.yml | bin/linkerd inject - | kubectl apply -f -
```

### Observe labels

```bash
$ kubectl -n emojivoto describe po/emoji-646ddcc5f9-8t7tr
Name:           emoji-646ddcc5f9-8t7tr
Namespace:      emojivoto
Node:           docker-for-desktop/192.168.65.3
Start Time:     Mon, 22 Apr 2019 13:37:34 -0700
Labels:         app=emoji-svc
                linkerd.io/control-plane-ns=linkerd
                linkerd.io/proxy-deployment=emoji
                pod-template-hash=2028877195
Annotations:    linkerd.io/created-by=linkerd/proxy-injector stable-2.3.0
                linkerd.io/identity-mode=default
                linkerd.io/inject=enabled
                linkerd.io/proxy-version=stable-2.3.0
Status:         Running
```

```bash
$ kubectl -n emojivoto describe po/web-57b7f9db85-qg7gn
Name:           web-57b7f9db85-qg7gn
Namespace:      emojivoto
Node:           docker-for-desktop/192.168.65.3
Start Time:     Mon, 22 Apr 2019 13:37:35 -0700
Labels:         app=web-svc
                linkerd.io/control-plane-ns=linkerd
                linkerd.io/proxy-replicaset=web-57b7f9db85
                pod-template-hash=1363958641
Annotations:    linkerd.io/created-by=linkerd/proxy-injector stable-2.3.0
                linkerd.io/identity-mode=default
                linkerd.io/inject=enabled
                linkerd.io/proxy-version=stable-2.3.0
Status:         Running
```

```bash
$ kubectl -n emojivoto get po -o jsonpath='{range .items[*]}{.metadata.labels}{"\n"}{end}'
map[linkerd.io/proxy-deployment:emoji pod-template-hash:194161803 app:emoji-svc linkerd.io/control-plane-ns:linkerd]
[...]

PR #2881, opened 2019-06-03
Resync API when informer index returns Not Found

### LINKED ISSUE #2738: evaluate forced k8s informer resync
#2737 fixed an issue (#2731) where the proxy-injector's k8s informer index was out of sync with the state of k8s. The fix in that PR was to fallback to a GET request.

While the proxy-injector's case is a bit unique, in that it's by definition operating on k8s resources in the process of being created, there still may be room for improvement in how the control-plane manages it's informer indexes. Ideally upon a not found error from the informer, we could force and block on a resync of the index, and do it across all informer lookups in the control-plane.

Misc notes:
https://github.com/linkerd/linkerd2/pull/2737#issuecomment-485871889
https://github.com/linkerd/linkerd2/pull/2737#issuecomment-485974192

alpeb:
Fixes #2738 and followup to #2737

When an informer index returns Not Found for a given resource, force an
API resync and try the retrieval one more time.

Also added an API function for retrieving ReplicaSets.

This tested well when retrieving ReplicaSets that were not yet indexed,
in the proxy injector (#2731)

Note: As an alternative I did try to do this in the cache index itself
by replacing the ThreadSafeStore with another Store that would do the
resync inside the Get() method, but I couldn't find a way to replace the
Store, given the private visibility of some of these structs in
go-client. Anyways, it's probably better to do this kind of
customization more explicitly on the calling side.

### DIFF
--- a/controller/k8s/api.go
+++ b/controller/k8s/api.go
@@ -174,6 +174,7 @@ func NewAPI(k8sClient kubernetes.Interface, spClient spclient.Interface, resourc
 
 // Sync waits for all informers to be synced.
 func (api *API) Sync() {
+	// It doesn't matter if informers were already started
 	api.sharedInformers.Start(nil)
 	api.spSharedInformers.Start(nil)
 
@@ -340,18 +341,7 @@ func (api *API) GetOwnerKindAndName(pod *corev1.Pod) (string, string) {
 
 	parent := ownerRefs[0]
 	if parent.Kind == "ReplicaSet" {
-		rs, err := api.RS().Lister().ReplicaSets(pod.Namespace).Get(parent.Name)
-		if err != nil {
-			log.Warnf("failed to retrieve replicaset from indexer, retrying with get request %s/%s: %s", pod.Namespace, parent.Name, err)
-
-			// try again with a get request
-			rs, err = api.Client.AppsV1beta2().ReplicaSets(pod.Namespace).Get(parent.Name, metav1.GetOptions{})
-			if err != nil {
-				log.Errorf("failed to get replicaset from k8s %s/%s: %s", pod.Namespace, parent.Name, err)
-			} else {
-				log.Debugf("successfully recovered replicaset via k8s get request: %s/%s", pod.Namespace, parent.Name)
-			}
-		}
+		rs, err := api.getRSs(pod.Namespace, parent.Name)
 		if err != nil || len(rs.GetOwnerReferences()) != 1 {
 			return strings.ToLower(parent.Kind), parent.Name
 		}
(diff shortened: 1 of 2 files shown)


ihcsim:
Per slack convo, we will put this PR on hold for now. Since the `api.Sync()` is a blocking op, we will need to better understand the implication of invoking it whenever a resource isn't found. Other alternatives to consider include, reducing the hard coded resync duration, resyncing only relevant resource(s), are there resources that should be resynced more often (see [`informers.WithCustomResyncConfig`](https://godoc.org/k8s.io/client-go/informers#WithCustomResyncConfig), can we use a resource version to determine if it's stale etc.

controller/k8s/api.go
```diff
@@ -174,6 +174,7 @@ func NewAPI(k8sClient kubernetes.Interface, spClient spclient.Interface, resourc
 
 // Sync waits for all informers to be synced.
 func (api *API) Sync() {
```
siggy: since we're calling `Sync()` at non-startup times now, consider adding a comment that this method may kill the process if sync fails. we could go a step further and have it return an error rather than exit, but maybe not worth it. also consider adding a log line (warning? info?) to help us better understand how often this is happening.

(un)related: is it safe to call this method concurrently?

controller/k8s/api.go
```diff
@@ -600,6 +606,18 @@ func (api *API) getRCs(namespace, name string) ([]runtime.Object, error) {
 	return objects, nil
 }
 
+func (api *API) getRSs(namespace, name string) (*appsv1beta2.ReplicaSet, error) {
```
siggy: 👍

controller/tap/server_test.go
```diff
@@ -240,7 +240,7 @@ status:
 
 			// TODO: mock out the underlying grpc tap events, rather than waiting an
 			// arbitrary time for request to timeout.
-			ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+			ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
```
siggy: intentional?

alpeb: Yea, that's for one of those tests where the tap request is asking for a pod that doesn't exist, so we're hitting Sync(), which takes a bit more of time. With less than 200ms we hit a timeout error.

controller/k8s/api.go
alpeb: Re concurrency safety, the underlying functions there are the `HasSynced` functions, which at least are locked per sharedInformer:

https://github.com/kubernetes/client-go/blob/7de88b14dcc8f8ebdc2953e776666387205cb592/tools/cache/shared_informer.go#L283-L291

alpeb:
Thank you all for the feedback.

As suggested by @ihcsim I did try refactoring all this so that we only send `WaitForCacheSync` the syncCheck relevant for each situation. Strangely enough the performance remained similar, and in some tests it got a bit worse. The unit tests are particularly good to catch those delays since lots of cases deal with the resource not-found scenario (also seen in the change I had to do to server_test.go above). 

I can get behind @klingerf's suggestion. What do you think @siggy, given the initial intention of #2738?

siggy:
@alpeb I'm onboard with @klingerf's suggestion. I'd also be onboard with dropping this PR altogether. The initial intent of #2738 was motivated by https://github.com/linkerd/linkerd2/pull/2737#issuecomment-485871889. I think @klingerf's suggestion (and the existing implementation) correctly make an explicit distinction between cached and non-cached reads. There are specific cases (ReplicaSet creation) where we want non-cached reads, but implicit fallback for all reads feels like a bandaid, and adds complexity.

alpeb:
Sounds good to me :+1: 
I've created #2972 for tracking the new approach.

PR #2897, opened 2019-06-04
Add support for TrafficSplits

adleong:
Add support for querying TrafficSplit resources through the common API layer.  This is done by depending on the TrafficSplit client bindings from smi-sdk-go.


controller/k8s/clientset.go
```diff
@@ -21,3 +22,16 @@ func NewSpClientSet(kubeConfig string) (*spclient.Clientset, error) {
 	config.WrapTransport = prometheus.ClientWithTelemetry("sp", wt)
 	return spclient.NewForConfig(config)
 }
+
+// NewTsClientSet returns a Kubernetes TrafficSplit client for the given
+// configuration.
+func NewTsClientSet(kubeConfig string) (*tsclient.Clientset, error) {
```
alpeb: I think we can extract the common code from the two functions in this file

controller/k8s/api.go
```diff
@@ -106,23 +112,47 @@ func InitializeAPI(kubeConfig string, resources ...APIResource) (*API, error) {
 			break
 		}
 	}
-	return NewAPI(k8sClient, spClient, resources...), nil
+
+	// TrafficSplits
+	var tsClient *tsclient.Clientset
+	for _, res := range resources {
+		if res == SP {
```
alpeb: I believe this should be `TS`

controller/k8s/api.go
```diff
@@ -298,6 +332,14 @@ func (api *API) SPAvailable() bool {
 	return api.sp != nil
 }
 
+// TS provides access to a shared informer and lister for TrafficSplits.
+func (api *API) TS() spinformers.ServiceProfileInformer {
+	if api.sp == nil {
+		panic("SP informer not configured")
+	}
+	return api.sp
+}
```
alpeb: s/sp/ts

pkg/k8s/fake.go
```diff
@@ -52,15 +55,16 @@ func NewFakeAPIFromManifests(readers []io.Reader) (*KubernetesAPI, error) {
 
 // NewFakeClientSets provides mock Kubernetes ClientSets.
 // TODO: make this private once KubernetesAPI (and NewFakeAPI) supports spClient
-func NewFakeClientSets(configs ...string) (kubernetes.Interface, apiextensionsclient.Interface, spclient.Interface, error) {
+func NewFakeClientSets(configs ...string) (kubernetes.Interface, apiextensionsclient.Interface, spclient.Interface, tsclient.Interface, error) {
```
dadjeibaah: nit: It feels like `NewFakeClientSets`'s return types need to be compressed to a struct or meta interface. This will also help when we add new fake clients down the road.

pkg/k8s/fake.go
```diff
@@ -5,8 +5,11 @@ import (
 	"io"
 	"strings"
 
+	tsclient "github.com/deislabs/smi-sdk-go/pkg/gen/client/split/clientset/versioned"
+	tsfake "github.com/deislabs/smi-sdk-go/pkg/gen/client/split/clientset/versioned/fake"
 	spclient "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
 	spfake "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned/fake"
+
```
dadjeibaah: Extra space

controller/k8s/clientset.go
adleong: Hilariously, because error handling has to be repeated at every call layer, this refactor actually makes the code *longer*: 

```
--- a/controller/k8s/clientset.go
+++ b/controller/k8s/clientset.go
@@ -5,33 +5,41 @@ import (
        spclient "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
        "github.com/linkerd/linkerd2/pkg/prometheus"
+       "k8s.io/client-go/rest"

        // Load all the auth plugins for the cloud providers.
        _ "k8s.io/client-go/plugin/pkg/client/auth"
 )

+func newConfig(kubeConfig string, telemetryName string) (*rest.Config, error) {
+       config, err := k8s.GetConfig(kubeConfig, "")
+       if err != nil {
+               return nil, err
+       }
+
+       wt := config.WrapTransport
+       config.WrapTransport = prometheus.ClientWithTelemetry(telemetryName, wt)
+       return config, nil
+}
+
 // NewSpClientSet returns a Kubernetes ServiceProfile client for the given
 // configuration.
 func NewSpClientSet(kubeConfig string) (*spclient.Clientset, error) {
-       config, err := k8s.GetConfig(kubeConfig, "")
+       config, err := newConfig(kubeConfig, "sp")
        if err != nil {
                return nil, err
        }

-       wt := config.WrapTransport
-       config.WrapTransport = prometheus.ClientWithTelemetry("sp", wt)
        return spclient.NewForConfig(config)
```

I still think it's a good idea though.

Issue #2852, opened 2019-05-28
Support setting FailurePolicy of proxy injector webhook


annismckenzie:
## Feature Request
Support specifying the failure policy of the proxy injector webhook during `linkerd install` and `linkerd upgrade`.

### What problem are you trying to solve?
We'd like to prevent deployments of apps, especially when things go wrong (node on which the proxy injector webhook runs is partitioned from the rest of the cluster and a failing node's pods being rescheduled onto healthy nodes), to result in non-meshed applications.

### How should the problem be solved?
Optionally allow configuring the failure policy of the proxy injector webhook during the control plane installation/upgrade.

### Any alternatives you've considered?

We're using kustomize so we can just patch the webhook configuration manifest now that #2806 has landed but not everyone does so it'd be nice to be able to pass it.

### How would users interact with this feature?
```
linkerd install --proxy-injector-failure-policy=fail | kubectl apply -f -
```

olix0r:
I wonder if there's a good reason not to _always_ set this failure policy? It seems like the most desirable behavior, at least when `--ha` is used for installation.

annismckenzie:
Indeed, that should be the default in the HA case!

grampelberg:
I'm +1 to @olix0r's suggestion to turn it on by default `--ha`. For now, I'd prefer to keep from adding one more configuration option as a separate item though.

alpeb:
I'm having some trouble implementing this. Given all the resources are persisted at once during install, the order in which they are ready is not deterministic. From my testing, it's frequent that the the webhook config is persisted before the proxy injector. So we end up with a chicken-and-egg situation: the proxy injector can't start because the webhook config is forcing it to go through the proxy injector itself which isn't ready yet.

This only happens when the failure policy is `Fail`. And no matter how the yamls are arranged (e.g. putting the webhook config at last), this happens sometimes.

The only way around this that I can think of now would be to have an extra command (or `install` sub-command) to set this policy _after_ the control plane is ready.

olix0r:
@alpeb I suppose the alternative would be for the proxy-injector to be installed as part of the phase-0 (cluster role bindinsg etc)? But that sounds likely to be complicated as well.

I agree that this probably makes most sense as something that is configured after/outside of installation. Perhaps only as a part of upgrade? But we'll need to be especially sensitive to the various ways a cluster can get into a stuck state if it loses all of the proxy injectors...

alpeb:
[...]

PR #2900, opened 2019-06-05
Fix HA during upgrade

alpeb:
If we have a Linkerd installation with HA, and then we do `linkerd
upgrade` without specifying `--ha`, the replicas will get set back to 1,
yet the resource requests will keep their HA values.

Desired behavior: `linkerd install --ha` adds the `ha` value into the
linkerd-config, so it should be used during upgrade even if `--ha` is
not passed to `linkerd upgrade`.
Note we still can do `linkerd upgrade --ha=false` to disable HA.

This is a prerequesite to address #2852


alpeb:
I've just pushed a new test that I confirmed failed before this fix :star:

PR #2902, opened 2019-06-06
New `linkerd upgrade` flag for setting mutating webhook failure policy

### LINKED ISSUE #2852: Support setting FailurePolicy of proxy injector webhook

alpeb:
Fixes #2852

Added the `--failure-policy` flag for `linkerd upgrade`, that allows
setting the mutating webhook failure policy.

There needs to be at least one proxy injector pod running when the
failure policy is changed to 'Fail': if there's none and the policy is
changed, when the proxy injector tries to start it won't be allowed
because the webhook config will force it to go through an already
existing proxy injector and fail if there's none.

So the policy can be changed to 'Fail` only when:
- Linkerd was already installed with HA
- The CLI version matches the control plane version (so there are no
changes to the proxy-injector that would cause it to be redeployed)
- The control plane is healthy

Note that these checks are performed only the first time that flag is
set.

Another change: when doing `linkerd install --ha` the failure policy for
the validating webhook config is set to Fail.

Still to do: tests.

### DIFF
--- a/chart/templates/proxy_injector-rbac.yaml
+++ b/chart/templates/proxy_injector-rbac.yaml
@@ -70,6 +70,7 @@ webhooks:
       namespace: {{ .Namespace }}
       path: "/"
     caBundle: {{ b64enc .ProxyInjector.CrtPEM }}
+  failurePolicy: {{.ProxyInjector.FailurePolicy}}
   rules:
   - operations: [ "CREATE" , "UPDATE" ]
     apiGroups: [""]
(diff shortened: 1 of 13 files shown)


alpeb:
Closing given there's a much simpler approach, as explained in this [comment](https://github.com/linkerd/linkerd2/issues/2852#issuecomment-499635352).

PR #2903, opened 2019-06-06
Upgrading packages and updating package.json to reflect current versions 

scottcarol:
Recent `package.json` upgrades (#2818) triggered warnings about `react-router` and `react-router-dom`'s deprecated use of `require("history/FILENAME")`. This PR upgrades `react-router` and `react-router-dom` so that the warnings are not triggered.

To test: 
1. From `master`, run `bin/web setup` then `bin/web test`. (You can also run `bin/web dev` and look at your dashboard console.) The tests will pass but you will see multiple warnings like the below:
```
Please use require("history").createBrowserHistory instead of require("history/createBrowserHistory"). Support for the latter will be removed in the next major release.
```
2. Check out this branch, run `bin/web setup` then `bin/web test`. The tests will pass with no console warnings.


scottcarol:
After a closer look, the `console.log` warning addressed in this PR is related to package changes that happened with `yarn upgrade`. No need for more reviews until I push a new commit.

scottcarol:
This new commit
* fixes security issue #2908 in dashboard
* updates and pins package version numbers in `package.json` to reflect the actual versions in `yarn.lock`
* pins `react-iframe` to `1.7.16` and `jest` to `23.6.0` - in both cases, there are later versions but they include breaking changes. will make separate issues to update these libraries.
* modifies `webpack.config.js` to work with the updated `css-loader` library (related to #2908)

To test:
run `bin/web setup` and `bin/web dev`, and all tests (unit and integration) - all pages should render correctly including `Community` which is currently not rendering in `master`

**Asset size changes:** -1 KiB 
`$ NODE_ENV=production yarn webpack`
Master branch:
```
main (902 KiB)
      main.8583df7fe17a8dc7c018.js
```
Current branch:
```
main (901 KiB)
      main.84168178a272a6228340.js
```

web/app/package.json
```diff
@@ -7,40 +7,40 @@
...
-    "babel-plugin-transform-react-remove-prop-types": "^0.4.21",
+    "babel-plugin-transform-react-remove-prop-types": "0.4.24",
     "babel-preset-env": "1.7.0",
     "babel-preset-react-app": "3.1.1",
-    "babel-runtime": "^6.26.0",
-    "chromedriver": "^2.46.0",
+    "babel-runtime": "6.26.0",
+    "chromedriver": "2.46.0",
     "clean-webpack-plugin": "1.0.0",
-    "css-loader": "0.28.7",
```
rmars: 👀

Issue #2888, opened 2019-06-03
Add namespaceSelector that matches control plane namespace to MWC


grampelberg:
### What problem are you trying to solve?

When multiple MWC mutators are running in a cluster, each one needs to know what pods it is authoritative for.

### How should the problem be solved?

Each MWC will have a label selector in this form:  `"namespaceSelector": { "matchLabels": { "config.linkerd.io/control-plane": "ns-of-proxy-injector" }}` . 

This will match all the workloads under any namespace that has the label `config.linkerd.io/control-plane: ns-of-proxy-injector`

### Open questions

- This feature only works at the namespace level. Meaning each namespace that gets added must have `config.linkerd.io/control-plane: namespace`. In its labels. How do we solve this for namespaces that were added before Linkerd was installed?

- What happens we a MWC doesn't have a namespaceSelector and there are multiple MWC. we will then have pod specs sent to all webhooks and will put the cluster in a bad state. One solution might be to check if the MWC doesn't have a selector and if there are multiple MWC, we reject the admission request with some error message.

dadjeibaah:
Marking this as related to #2909. The idea is to add a `namespaceSelector` in the MWC to get this working properly.

ihcsim:
@grampelberg @dadjeibaah @alpeb After catching up on the to-date comments, I will like to propose we reconsider using the MWC's namespace selector to define the proxy injectors' authoritative scope. My primary concern lies in potentially conflating the implication of the presence or absence of the MWCs' namespace selectors.

Can we simplify the workflow to:

1. A user installs multiple control planes with `linkerd install -l <ns-*>`
1. `linkerd check` reports a warning that there are multiple MWCs in the k8s cluster, linking to the doc with more info on the implication of multiple MWCs.
1. The doc explains that either this is the byproduct of an upgrade (see https://github.com/linkerd/website/issues/336), or it's part of the user's multi control plane setup.
1. The proxy injector is updated to compare its namespace with the `config.linkerd.io/control-plane` annotation in the workload's YAML to decide if it's responsible for injecting the given workload.
1. The `linkerd inject` code is updated to inject the `config.linkerd.io/control-plane` annotation (or just reuse the existing `linkerd.io/control-plane-ns` annotation).
[...]

PR #2905, opened 2019-06-06
add namespace name to MWC

### LINKED ISSUE #2887: Add namespace to MWC

When running multiple control planes on a single cluster, the MWC resources are named identically and clobber each other.

Match the other components naming schema and add the control plane namespace to the name so that each can coexist.

dadjeibaah:
When installing multiple control planes, the `mutatingwebhookconfiguration` of the first control plane gets overwritten by any subsequent control plane install. This is caused by the fixed name given to the `mutatingwebhookconfiguration` manifest at install time.

This PR adds in the namespace to the manifest so that there is a unique configuration for each control plane.

Fixes #2887

### DIFF
--- a/chart/templates/proxy_injector-rbac.yaml
+++ b/chart/templates/proxy_injector-rbac.yaml
@@ -59,7 +59,7 @@ data:
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: linkerd-proxy-injector-webhook-config
+  name: linkerd-{{ .Namespace }}-proxy-injector-webhook-config
   labels:
     {{ .ControllerComponentLabel }}: proxy-injector
 webhooks:
--- a/chart/templates/sp_validator-rbac.yaml
+++ b/chart/templates/sp_validator-rbac.yaml
@@ -53,7 +53,7 @@ data:
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: linkerd-sp-validator-webhook-config
+  name: linkerd-{{ .Namespace }}-sp-validator-webhook-config
   labels:
     {{ .ControllerComponentLabel }}: sp-validator
 webhooks:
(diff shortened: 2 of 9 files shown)


alpeb:
If you have multiple MWCs they will all get triggered when a pod is created in any namespace. I think one prerequisite of multi-tenancy is to have user namespaces (e.g. emojivoto) have a label that declares what control plane namespace they're using. Then you can add a `namespaceSelector` into the MWC so that it only applies to user namespaces with a label that matches the appropriate linkerd namespace.. makes sense?

What I've just mentioned wouldn't be BC, so we might need to consider a new install flag or something to explicitly support multi-tenancy.

Btw, I've already started using `namespaceSelector`, but for a different purpose, in #2906

alpeb:
> What I've just mentioned wouldn't be BC, so we might need to consider a new install flag or something to explicitly support multi-tenancy.

... or have the `namespaceSelector` condition match also when that label doesn't exist (`DoesNotExist`)

dadjeibaah:
Yea, that makes sense @alpeb, talked with @grampelberg offline as well and he described the same scenario. So it sounds like #2888 needs to go out first and then this change should probably immediately follow.

dadjeibaah:
@alpeb, thinking about this more I think I understand the situation better. I am still a little fuzzy on how admission controllers work but If we run `linkerd inject`, do control plane namespaces annotations get added before or after the API request object flows through admission controller chain? If that is the case, all we would need to do is add an annotation in the webhook configuration only and then change the injector service to only mutate pod specs that have the matching label.

Now, I feel like I just explained exactly what you are describing 🤔.

ihcsim:
If the intention is to implement #2888 in another PR, this change LGTM. As @alpeb pointed out, the change in this PR (73bcfb405f0567967ba8a21e2a6fe8d92b719da4) alone doesn't solve the multi MWCs problem. 

The other thing to be aware of is that, during upgrade, this change will cause a new MWC to be created alongside, instead of replacing, the previous `linkerd-proxy-injector-webhook-config` MWC. We will need to include in the 2.4 upgrade notes (https://github.com/linkerd/website/issues/336), together with https://github.com/linkerd/linkerd2/issues/2912 to make sure the pre-2.4 MWC is manually deleted.

PR #2907, opened 2019-06-06
Edge 19.6.1 release notes

hawkw:
## edge-19.6.1

* CLI
  * Fixed an issue where, when Linkerd is installed with `--ha`, running
    `linkerd upgrade` without `--ha` will disable the high availability
    control plane
  * Added a `--init-image-version` flag to `linkerd inject` to override the
    injected proxy-init container version
* Controller
  * Added multiple replicas for the `proxy-injector` and `sp-validator`
    controllers when run in high availability mode (thanks to @Pothulapati!)
* Proxy
  * Fixed a memory leak that can occur if an HTTP/2 request with a payload
    ends before the entire payload is sent to the destination
* Internal
  * Moved the proxy-init container to a separate `linkerd/proxy-init` Git
    repository

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,21 @@
+## edge-19.5.4
+
+* CLI
+  * Fixed an issue where, when Linkerd is installed with `--ha`, running
+    `linkerd upgrade` without `--ha` will disable the high availability
+    control plane
+  * Added a `--init-image-version` flag to `linkerd inject` to override the
+    injected proxy-init container version
+* Controller
+  * Added multiple replicas for the `proxy-injector` and `sp-validator`
+    controllers when run in high availability mode (thanks to @Pothulapati!)
+* Proxy
+  * Fixed a memory leak that can occur if an HTTP/2 request with a payload
+    ends before the entire payload is sent to the destination
+* Internal
+  * Moved the proxy-init container to a separate `linkerd/proxy-init` Git
+    repository
+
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,21 @@
+## edge-19.5.4
+
+* CLI
+  * Fixed an issue where, when Linkerd is installed with `--ha`, running
+    `linkerd upgrade` without `--ha` will disable the high availability
```
scottcarol: I was a little confused about "will disable the high availability control plane", do you mean that is NOW the behavior or it was the previous behavior? Should the verbs be "when Linkerd was installed... ...would disable..."?

CHANGES.md
```diff
@@ -1,3 +1,21 @@
...
+    control plane
+  * Added a `--init-image-version` flag to `linkerd inject` to override the
+    injected proxy-init container version
+* Controller
+  * Added multiple replicas for the `proxy-injector` and `sp-validator`
+    controllers when run in high availability mode (thanks to @Pothulapati!)
+* Proxy
+  * Fixed a memory leak that can occur if an HTTP/2 request with a payload
```
scottcarol: maybe switch "can" to "could" since it's now in the past?

CHANGES.md
hawkw: Yeah, I think this should be past tense. Thanks!

Issue #2767, opened 2019-04-29
Admission Controllers don't run over the proxy and init injected containers. Check consequences.


alpeb:
At the moment the proxy injector injects the proxy and init containers, most of the Admission Controllers have already run (on the owner ref (Deployment, StatefulSet, etc) ) and they won't run again against the injected containers.

This has forced us to, during injection:
1) Do the job of the ServiceAccount MWC, by manually mounting the token mount
2) Do the job of the LimitRanger by manually setting the resource requests/limits on the init container (and forcing people to set them for the proxy through injection flags/annotations) #2763

This issue is to track if there are other possible issues with other MWCs.

grampelberg:
@alpeb is there a way to change the ordering at all?

alpeb:
@grampelberg I couldn't find any docs specifying this, so after lots of k8s code scavenging I could confirm the order is hard-coded:

https://github.com/kubernetes/kubernetes/blob/5268f69405251a4a74130fa903e055a59071179a/pkg/kubeapiserver/options/plugins.go#L61-L94

The mutating webhook runs almost at last.

ihcsim:
Other relevant issues:

https://github.com/kubernetes/kubernetes/issues/55435
https://github.com/kubernetes/kubernetes/issues/65716
https://github.com/kubernetes/kubernetes/issues/64333

annismckenzie:
Because @alpeb just commented on my PR in our fork I'm gonna link to it here because it's relevant: alfatraining/linkerd2#1 – disabling `AutomountServiceAccountToken` in the pod spec will break the sidecar.

alpeb:
I went through the list of admission controller plugins in the [k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/), and these are the plugins that can cause problems because of the proxy injector ignoring them:

## PSP
The proxy injector can't work with PSP restrictions. During the mutating phase, after the containers are injected the PSP plugin won't be triggered so it won't add its restrictions into the containers. Then during the validation phase the same PSP plugin won't find those restrictions and it will deny the creation of the pod.

## LimitRanger and ResourceQuota
Already handled in #2763 by making sure we have resource requests and limits set on the injected containers

## ServiceAccount
[...]

PR #2911, opened 2019-06-07
Correctly set securityContext values on injection

codeman9:
The patch provided by @ihcsim applies correct values for the securityContext during injection, namely: `allowPrivilegeEscalation = false`, `readOnlyRootFilesystem = true`, and the capabilities are copied from the primary container. Additionally, the proxy-init container securityContext has been updated with appropriate values.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -426,13 +426,21 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
 		}
 	}
 
-	proxyUID := conf.proxyUID()
+	var (
+		proxyUID                 = conf.proxyUID()
+		allowPrivilegeEscalation = false
+		readOnlyRootFilesystem   = true
+	)
 	sidecar := corev1.Container{
 		Name:                     k8s.ProxyContainerName,
 		Image:                    conf.taggedProxyImage(),
 		ImagePullPolicy:          conf.proxyImagePullPolicy(),
 		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
-		SecurityContext:          &corev1.SecurityContext{RunAsUser: &proxyUID},
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: &allowPrivilegeEscalation,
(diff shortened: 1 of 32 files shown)


ihcsim:
@codeman9 Thanks! I'm gonna request extra pairs of eyes to review this. 

For the record, I have uploaded the PSP I used to test both the CNI and non-CNI set up to this [gist](https://gist.github.com/ihcsim/d94f4cda94c1cce1662e12a68499c400).

ihcsim:
Fixes https://github.com/linkerd/linkerd2/issues/2819

pkg/inject/inject.go
```diff
@@ -599,24 +615,41 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
...
+	capabilities := &corev1.Capabilities{}
+	if conf.pod.spec.Containers != nil && len(conf.pod.spec.Containers) > 0 {
+		if sc := conf.pod.spec.Containers[0].SecurityContext; sc != nil && sc.Capabilities != nil {
+			capabilities = sc.Capabilities
+		}
+	}
+	if capabilities.Add == nil {
+		capabilities.Add = []corev1.Capability{}
+	}
+	capabilities.Add = append(capabilities.Add, corev1.Capability("NET_ADMIN"))
```
afirth: Unfortunately this seems to break proxy-init. Adding NET_RAW to the primary container's capabilities works around it. Should we add NET_RAW here? #2959

ihcsim: Can confirm that `NET_RAW` is needed when either the PSP or primary container security context has a `Drop: ALL` capability rule.

For posterity, the reason why we have to copy the capabilities of the primary container into the proxy-init and proxy sidecars is because by the time the proxy injector receives the admission request, the primary container would have been mutated by the PSP admission controller to include any default `requiredDropCapabilities` rules. The PSP validation which happens after expects the rules to be present in all the containers' security context. AFACIT, the validation is performing a exact string values match.

All these will become moot in k8s 1.15 per https://github.com/linkerd/linkerd2/issues/2767#issuecomment-497538600

PR #2919, opened 2019-06-11
Remove UPDATE event from the mutating webhook config

### LINKED ISSUE #2889: Remove UPDATE operation from proxy-injector webhook
From earlier testing, it appears it's not possible to inject the proxy sidecar during a pod UPDATE event. The only mention I could find in k8s' docs about this was in the PSP docs
https://kubernetes.io/docs/concepts/policy/pod-security-policy/
"...because pod mutations are disallowed during update operations"

We need to double check this is really the case, and if so, remove that entry from `proxy_injector-rbac.yaml`

alpeb:
Fixes #2889


grampelberg:
👍

PR #2923, opened 2019-06-11
Fix BaseTable ignoring sorting order

rmars:
I noticed that the tables weren't maintaining sort order, and were completely ignoring the `defaultOrderBy` BaseTable prop. This branch fixes sorting and incorporates `defaultOrderBy` if provided.

Before:
Look at the tables, for example in the Service Mesh page, the meshed resource table, or in Top Routes. 
Note that both these tables have a default sort order, but do not appear sorted.

Screen Shot 2019-06-11 at 10 47 45 AM
Screen Shot 2019-06-11 at 10 47 49 AM

After:
These tables should be sorted by their default order.
Screen Shot 2019-06-11 at 11 14 30 AM
Screen Shot 2019-06-11 at 11 14 34 AM

### DIFF
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -16,7 +16,6 @@ import Tooltip from '@material-ui/core/Tooltip';
 import Typography from '@material-ui/core/Typography';
 import _find from 'lodash/find';
 import _get from 'lodash/get';
-import _isEmpty from 'lodash/isEmpty';
 import _isNil from 'lodash/isNil';
 import _orderBy from 'lodash/orderBy';
 import classNames from 'classnames';
@@ -90,10 +89,6 @@ class BaseTable extends React.Component {
   generateRows = (tableRows, tableColumns, order, orderBy, filterBy) => {
     let rows = tableRows;
     let col = _find(tableColumns, d => d.dataIndex === orderBy);
-    let {defaultOrderBy} = this.props;
-    if (_isEmpty(orderBy) && !_isEmpty(defaultOrderBy)) {
-      orderBy = defaultOrderBy;
-    }
 
     if (orderBy && col.sorter) {
       rows = _orderBy(rows, row => col.sorter(row), order);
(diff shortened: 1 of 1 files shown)


web/app/js/components/BaseTable.jsx
```diff
@@ -89,8 +90,13 @@ class BaseTable extends React.Component {
   generateRows = (tableRows, tableColumns, order, orderBy, filterBy) => {
     let rows = tableRows;
     let col = _find(tableColumns, d => d.dataIndex === orderBy);
+    let {defaultOrderBy} = this.props;
```
scottcarol: TIOLI: instead of declaring `defaultOrderBy` here, you could add it to line 180 and pass it in to `generateRows()`

web/app/js/components/BaseTable.jsx
```diff
@@ -89,8 +90,13 @@ class BaseTable extends React.Component {
...
+    if (_isEmpty(orderBy) && !_isEmpty(defaultOrderBy)) {
```
scottcarol: is there an advantage to using `if (_isEmpty(orderBy) && !_isEmpty(defaultOrderBy))` instead of `if (!orderBy && defaultOrderBy)` here? 

since ‘orderBy’ and ‘defaultOrderBy’ are always strings and would never contain boolean values, do we need lodash? (and if there is an advantage, should we be using this in line 98 as well)?

web/app/js/components/BaseTable.jsx
```diff
@@ -89,8 +90,13 @@ class BaseTable extends React.Component {
...
+      orderBy = defaultOrderBy;
+    }
+
     if (orderBy && col.sorter) {
-      rows = _sortBy(rows, row => col.sorter(row));
+      rows = _orderBy(rows, row => col.sorter(row), order);
```
scottcarol: super smart to use `_orderBy` instead of `_sortBy`, TIL!

web/app/js/components/BaseTable.jsx
rmars: I think rather than having a very long list of arguments to generateRows, I'd prefer to access things in the function where they're used. The parent render prop doesn't make use of defaultOrderBy at all.

web/app/js/components/BaseTable.jsx
rmars: I realized that `orderBy` is being set by `defaultOrderBy` in the constructor, and using `_orderBy` was enough to fix this bug, so I'm removing this check :)

but you're right, we don't need the isEmpty here if those are always strings!

PR #2930, opened 2019-06-12
Allow configuration of cluster base domain

arminbuerkle:
PR for #1720

Adds a global `--cluster-domain` flag to linkerd which allows configuring linkerd in a non `cluster.local` setup.

You can test it by running
```
./bin/linkerd install --cluster-domain=mycluster.local
```

I tested it with `go test -cover -race ./...` as well as deployed it into our own cluster.

---
I still need to test whether upgrading your cluster works since the `cluster-domain` flag is now required in the global config and set to `cluster.local` by default.

Changing the value however is something that needs a reinstall.

---
I'll squash the commits when the PR looks.


olix0r:
Thanks @arminbuerkle! As one first bit of feedback, I think we'll need to start configuring the proxy's discovery suffixes:

https://github.com/linkerd/linkerd2-proxy/blob/1a52a5e6cd6d0084e5e4d928f95a08bead9c04f9/src/app/config.rs#L350-L351

The injector is going to have to be updated to support setting `LINKERD2_PROXY_DESTINATION_GET_SUFFIXES` and `LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES`.

Because this branch is so large, it is likely to require a substantial amount of testing and review; and in that time it may run into conflicts with master as new features are merged in the next week or two.

I wonder if this can be broken into smaller PRs so that we can more easily ship tested pieces of this?  For instance, if you can propose changes to the various controller components individually, and then follow with changes to the install flow that support setting non-default cluster names... We find that this incremental approach works best when adding these types of cross-cutting features to the project.

arminbuerkle:
Hi @olix0r,

i started splitting this MR into smaller ones starting with #2950

The controller part still has a lot of changes, mostly auto generated though (proto, .golden, updating tests so it doesn't use `cluster.local` as default).

AFAIK [`LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES` is already set](https://github.com/alfatraining/linkerd2/blob/add-cluster-domain/pkg/inject/inject.go#L481) on the go side. 

I can add `LINKERD2_PROXY_DESTINATION_GET_SUFFIXES` to the injector, though i cannot tell if there are changes required in rust since i'm not fluent in it.

The value will probably be the same as for profile suffixes, since it needs to change when external profiles are enabled?

olix0r:
@arminbuerkle thanks! That sounds correct. No changes should be necessary on the Rust side, as that configuration is already honored.

wmorgan:
Thanks @arminbuerkle, we will take a look soon!

arminbuerkle:
@wmorgan thanks!

I'll try and keep this branch updated as well as the ones which do the same changes but split up into several PRs starting with #2950 

I use this one here currently for my local cluster and it works afaik 😄

ihcsim:
@arminbuerkle Thanks again for taking this on. It's going to help out a number of people in the community. To confirm, you're going to split this PR into multiple smaller PRs, starting with #2950, right? If that's case, is it ok if we close this one?

arminbuerkle:
@ihcsim yes. I'm going to close this though i'll keep this branch updated as good as i can if anyone wants to build it themselve and try.

As i said before we're running this on two of our own clusters with different domains.

The branches for the other PRs are already created and linked above, though #2950 needs to be merged first since the others depend on that.

ihcsim:
Appreciate it. We will try to get to #2950 soon as we know it affects you and a number of other users.

PR #2931, opened 2019-06-12
Add traffic splitting to destination profiles

adleong:
This change implements the DstOverrides feature of the destination profile API (aka traffic splitting).

We add a TrafficSplitWatcher to the destination service which watches for TrafficSplit resources and notifies subscribers about TrafficSplits for services that they are subscribed to.  A new TrafficSplitAdaptor then merges the TrafficSplit logic into the DstOverrides field of the destination profile.

As part of this change, we also install the TrafficSplit CRD as part of Linkerd's installation.


controller/gen/apis/serviceprofile/v1alpha1/types.go
```diff
@@ -27,8 +28,9 @@ type ServiceProfile struct {
 
 // ServiceProfileSpec specifies a ServiceProfile resource.
 type ServiceProfileSpec struct {
-	Routes      []*RouteSpec `json:"routes"`
-	RetryBudget *RetryBudget `json:"retryBudget,omitempty"`
+	Routes       []*RouteSpec   `json:"routes"`
+	RetryBudget  *RetryBudget   `json:"retryBudget,omitempty"`
+	DstOverrides []*WeightedDst `json:"dstOverrides,omitempty"`
```
alpeb: Doesn't this need a corresponding change to the ServiceProfile CRD?

chart/templates/trafficsplit-crd.yaml
```diff
@@ -0,0 +1,29 @@
...
+  group: split.smi-spec.io
+  version: v1alpha1
+  scope: Namespaced
+  names:
+    kind: TrafficSplit
+    shortNames:
+      - ts
+    plural: trafficsplits
+    singular: trafficsplit
+  additionalPrinterColumns:
```
alpeb: Note that `additionalPrinterColumns` is only available as of k8s v1.11, so we'll have to bump our minimal required version.

controller/gen/apis/serviceprofile/v1alpha1/types.go
alpeb: ... well I guess not given we don't care about the bare-bones validation the CRD provides

controller/api/destination/traffic_split_adaptor.go
```diff
@@ -0,0 +1,63 @@
...
+	if tsa.split != nil {
+		overrides := []*sp.WeightedDst{}
+		for _, backend := range tsa.split.Spec.Backends {
+			dst := &sp.WeightedDst{
+				Authority: fmt.Sprintf("%s.%s.svc.cluster.local:%d", backend.Service, tsa.split.Namespace, tsa.port),
+				Weight:    backend.Weight,
+			}
+			overrides = append(overrides, dst)
+		}
+		merged.Spec.DstOverrides = overrides
```
alpeb: This means any rules manually added into a `ServiceProfile` will get overridden by the corresponding `TrafficSplit` resource. Why not instead of going through service profiles use a separate gRPC endpoint for listening to changes from `TrafficSplit` only? That would also avoid having people install the service's service profile in order to use traffic splitting...

alpeb: I realize now no SP is actually required as `tsa.profile` can be nil, so we're just leveraging the `ServiceProfile` struct. That's cool.

alpeb:
@adleong it seems this privilege is missing:
```diff
diff --git a/chart/templates/controller-rbac.yaml b/chart/templates/controller-rbac.yaml
index 23567ca9..f205a262 100644
--- a/chart/templates/controller-rbac.yaml
+++ b/chart/templates/controller-rbac.yaml
@@ -21,6 +21,9 @@ rules:
 - apiGroups: ["linkerd.io"]
   resources: ["serviceprofiles"]
   verbs: ["list", "get", "watch"]
+- apiGroups: ["split.smi-spec.io"]
+  resources: ["trafficsplits"]
+  verbs: ["list"]
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1beta1
```

controller/api/destination/server.go
```diff
@@ -154,5 +179,5 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 
 func (s *server) Endpoints(ctx context.Context, params *discoveryPb.EndpointsParams) (*discoveryPb.EndpointsResponse, error) {
 	s.log.Debugf("serving endpoints request")
-	return nil, fmt.Errorf("Not implemented")
+	return nil, status.Error(codes.Unimplemented, "Not implemented")
```
klingerf: 👍

alpeb:
This is _pretty_ **awesome** :cowboy_hat_face: 
I'm testing with emojivoto and it's working great so far.

One question came up: when a service doesn't have a service profile associated but it does have a traffic split for it, `GetProfile()` is gonna return a default service profile with some `retryBudget`. This doesn't have any effect, as there are no routes marked with `isRetryable`, right?

adleong:
@alpeb that's correct.  Even when a service has no service profile AND no traffic split, the destination service will still serve a default destination profile to the proxy (which includes the default retry budget).

controller/api/destination/traffic_split_adaptor.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func (tsa *trafficSplitAdaptor) Update(profile *sp.ServiceProfile) {
+	tsa.profile = profile
+	tsa.publish()
+}
+
+func (tsa *trafficSplitAdaptor) UpdateTrafficSplit(split *ts.TrafficSplit) {
+	if tsa.split != split {
```
klingerf: This is a pointer comparison, which I'm guessing might not be what you want. Maybe use `reflect.DeepEqual`?

adleong:
@klingerf your understanding of the double update is correct.  this only happens on initial subscription and when there is already an existing traffic split.  we could add some de-duping logic to prevent this, but I don't think it's a big deal.

controller/api/destination/traffic_split_adaptor.go
adleong: This was a quick and dirty check to avoid sending traffic split updates when the traffic split is updated from `nil` to `nil`.  This helps to avoid duplicate updates in most cases (although there are still duplicate updates in some cases, as you described).  I'll tweak this to be more explicit about what it's doing.

chart/templates/trafficsplit-crd.yaml
```diff
@@ -0,0 +1,29 @@
+{{with .Values -}}
+---
+###
+### TrafficSplit CRD
+### Copied from https://github.com/deislabs/smi-sdk-go/blob/cea7e1e9372304bbb6c74a3f6ca788d9eaa9cc58/crds/split.yaml
+###
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
```
ihcsim: Mind labelling this new CRD? We labelled all the global resources in #2971. Thanks!
```
  labels:
```

adleong: Is the controller namespace an appropriate thing to label on the CRD?  CRDs are cluster wide.  If there were multiple control planes installed, they would all share a single traffic split CRD.

ihcsim: For now, we are disabling multi control plane setup. All the other global resources, including the service profile CRD, are labelled. The idea is to be able to find all Linkerd global and namespaced resources by using this label, for the purpose of determining if a control plane is installed. Previously, we rely only on checking for the existence of the `linkerd-config` config map, which is unreliable.

Issue #2885, opened 2019-06-03
endpoints: Use the public Destination APIs


olix0r:
The `linkerd endpoints` is tightly coupled to the internal representation of the destination service, without much concrete benefit.

I propose the following:

* The `linkerd endpoints` CLI takes a list of (fully qualified) authorities, e.g. _app.ns.svc.cluster.local:1234_
* The CLI uses the public Destination.Get method to describe the state of the service.

PR #2937, opened 2019-06-13
WIP: Refactor public api client.go to more easily add separate clients

alpeb:
The plan for #2885 is to have `linkerd endpoints` hit the
`Destination.Get` gRPC endpoint. Given this will probably be leveraged
from the dashboard as well, it made sense to do like with the other
endpoints and serve it through the public api http server and from there
forward to the Destination service.

The easy way would have been to just add the Destination endpoints into
/controller/api/public/client.go and complete the plumbing.

But I wanted to experiment using a separate new client
/controller/api/destination/client.go to try keeping things separate.

This implied extracting out of /controller/api/public some things that I
moved up one level into /controller/api to be reused by any client under
that directory.

Now both the public API and the Destination clients use a new
`portforward.Init()` method that captures the common logic for building
the port-forwarding, and then each client builds a separate
`grpcOverHTTPClient` struct that only implements the methods relevant to
each client.

I also created /controller/api/stream_client.go with the struct and
common methods needed for our gRPC streaming endpoints: Tap and
Destination.Get.

On the gRPC server side of things, one could make a similar refactoring
to avoid grouping together all the endpoints, but I skipped that for
this PR, and just did the normal plumbing.

/cli/cmd/endpoints.go changed `requestEndpointsFromAPI()` just to be
able to test the new Destination endpoint for a particular emoji
authority (you can test with just `linkerd endpoints`). No processing of
the output has been done yet.

Let me know what you think. Was this too much?

### DIFF
--- a/controller/api/public/http_server.go
+++ b/controller/api/public/http_server.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	destinationPb "github.com/linkerd/linkerd2-proxy-api/go/destination"
+	"github.com/linkerd/linkerd2/controller/api"
 	healthcheckPb "github.com/linkerd/linkerd2/controller/gen/common/healthcheck"
 	discoveryPb "github.com/linkerd/linkerd2/controller/gen/controller/discovery"
 	tapPb "github.com/linkerd/linkerd2/controller/gen/controller/tap"
@@ -27,6 +29,7 @@ var (
 	selfCheckPath     = fullURLPathFor("SelfCheck")
 	endpointsPath     = fullURLPathFor("Endpoints")
 	edgesPath         = fullURLPathFor("Edges")
+	destGetPath       = fullURLPathFor("Get")
 	configPath        = fullURLPathFor("Config")
 )
 
@@ -62,6 +65,8 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		h.handleSelfCheck(w, req)
 	case endpointsPath:
 		h.handleEndpoints(w, req)
+	case destGetPath:
+		h.handleDestGet(w, req)
 	case edgesPath:
 		h.handleEdges(w, req)
 	case configPath:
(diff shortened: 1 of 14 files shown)


pkg/k8s/portforward.go
```diff
@@ -162,6 +163,46 @@ func (pf *PortForward) Run() error {
 	return fw.ForwardPorts()
 }
 
+// Init creates and runs a port-forward connection, waits for it to be ready, and returns
+// its URL and http client.
+func (pf *PortForward) Init(controlPlaneNamespace string, kubeAPI *KubernetesAPI) (*url.URL, *http.Client, error) {
```
klingerf: 👍 I like moving this logic into `portforward.go`, even if we only end up calling this from one place.

But it looks like you're not using the `controlPlaneNamespace` param that's being passed in, and we're already passing `kubeAPI` when `NewPortForward` is called. If we persist that param on the `PortForward` struct, then you won't need to pass it here.

I also wonder if we could make this func more generic so that we could use it in `cli/cmd/metrics.go` and `cli/cmd/dashboard.go`.

controller/api/public/http_server.go
```diff
@@ -279,6 +306,21 @@ func (s tapServer) Send(msg *pb.TapEvent) error {
 	return nil
 }
 
+type destinationServer struct {
+	tapServer
```
klingerf: It feels strange to me that `destinationServer` embeds `tapServer`, since the destination and tap endpoints are unrelated. It might make sense to instead provide a `StreamServer` struct (a corollary to the `StreamClient` struct that you're creating in this branch), and then embed that struct in both `destinationServer` and `tapServer`.

pkg/k8s/portforward.go
alpeb: @klingerf I've created a separate PR for the port-forwarding refactoring: #2976
Looking forward for your comments :wink:

alpeb:
I've created #2990 to address #2885 without the client refactoring from this PR.

alpeb:
Closing this one in favor of #2990

PR #2941, opened 2019-06-13
Add edge-19.6.2 release notes

kleimkuhler:
## edge-19.6.2

* CLI
  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
    `NET_ADMIN` capability is omitted from the control plane's PSP
* Controller
  * Fixed security context values so that injection does not fail on
    restricted PSPs (thanks @codeman9!)
  * Fixed webhook failure policy in order to avoid timing issues in HA mode
  * Introduced control plane's PSP and RBAC resources into Helm templates;
    these policies are only in effect if the PSP admission controller is
    enabled
  * Fixed MWC namespace value so that when installing multiple control planes,
    there is a unique configuration for each one
  * Removed `UPDATE` operation from proxy-injector webhook because pod
    mutations are disallowed during update operations
* Proxy
  * The `l5d-override-dst` header is now honored for inbound service profiles
  * Added the prefix of metric reports in log lines
  * Fixed metrics so that response errors are properly counted
  * Changed requests to services with no endpoints to fail faster
* Web UI
  * Fixed dashboard behavior that caused incorrect table sorting

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@
   * Removed `UPDATE` operation from proxy-injector webhook because pod
     mutations are disallowed during update operations
 * Proxy
-  * The `l5d-override-dst` header is now honored for inbound service profile
+  * The `l5d-override-dst` header is now used for inbound service profile
     discovery
   * Include errors in `response_total` metrics
   * Changed the load balancer to require that Kubernetes services are resolved
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,27 @@
+## edge-19.6.2
+
+* CLI
+  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
+    `NET_ADMIN` capability is omitted from the control plane's PSP
+* Controller
+  * Fixed security context values so that injection does not fail on
+    restricted PSPs (thanks @codeman9!)
+  * Fixed webhook failure policy in order to avoid timing issues in HA mode
+  * Introduced control plane's PSP and RBAC resources into Helm templates;
+    these policies are only in effect if the PSP admission controller is
+    enabled
+  * Fixed MWC namespace value so that when installing multiple control planes,
+    there is a unique configuration for each one
+  * Removed `UPDATE` operation from proxy-injector webhook because pod
+    mutations are disallowed during update operations
+* Proxy
+  * The `l5d-override-dst` header is now honored for inbound service profiles
+  * Added the prefix of metric reports in log lines
+  * Fixed metrics so that response errors are properly counted
+  * Changed requests to services with no endpoints to fail faster
+* Web UI
+  * Fixed dashboard behavior that caused incorrect table sorting
+
 ## edge-19.5.4
 
 * CLI
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
   * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
     `NET_ADMIN` capability is omitted from the CNI-enabled control plane's PSP
 * Controller
+  * Default to least-privilege security context values for the proxy container
+    so that auto-inject does not fail on restricted PSPs (thanks @codeman9!)
   * Defined least privilege default security context values for the proxy
     container so that auto-injection does not fail on (thanks @codeman9!)
   * Default the webhook failure policy to `Fail` in order to account for
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,8 +4,8 @@
   * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
     `NET_ADMIN` capability is omitted from the CNI-enabled control plane's PSP
 * Controller
-  * Fixed security context values so that injection does not fail on
-    restricted PSPs (thanks @codeman9!)
+  * Defined least privilege default security context values for the proxy
+    container so that auto-injection does not fail on (thanks @codeman9!)
   * Default the webhook failure policy to `Fail` in order to account for
     unexpected errors during auto-inject; this ensures uninjected applications
     are not deployed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,11 +2,13 @@
 
 * CLI
   * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
-    `NET_ADMIN` capability is omitted from the control plane's PSP
+    `NET_ADMIN` capability is omitted from the CNI-enabled control plane's PSP
 * Controller
   * Fixed security context values so that injection does not fail on
     restricted PSPs (thanks @codeman9!)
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+  * Introduced control plane's PSP and RBAC resources into Helm templates;
+    these policies are only in effect if the PSP admission controller is
+    enabled
+  * Fixed MWC namespace value so that when installing multiple control planes,
+    there is a unique configuration for each one
+  * Removed `UPDATE` operation from proxy-injector webhook because pod
+    mutations are disallowed during update operations
+* Proxy
+  * The `l5d-override-dst` header is now honored for inbound service profiles
+  * Added the prefix of metric reports in log lines
```
olix0r: not user facing

hawkw: This is fairly unlikely to be user-facing IMO

CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+  * Fixed webhook failure policy in order to avoid timing issues in HA mode
+  * Introduced control plane's PSP and RBAC resources into Helm templates;
+    these policies are only in effect if the PSP admission controller is
+    enabled
+  * Fixed MWC namespace value so that when installing multiple control planes,
+    there is a unique configuration for each one
+  * Removed `UPDATE` operation from proxy-injector webhook because pod
+    mutations are disallowed during update operations
+* Proxy
+  * The `l5d-override-dst` header is now honored for inbound service profiles
```
hawkw:
```suggestion
  * The `l5d-override-dst` header is now honored for inbound service profile discovery
```
might be a little bit clearer IMO?

CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+  * Added the prefix of metric reports in log lines
+  * Fixed metrics so that response errors are properly counted
```
hawkw: TIOLI: what do you think of 
```suggestion
  * Fixed response errors not being properly counted in metrics
```

CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+  * Changed requests to services with no endpoints to fail faster
```
olix0r: This is phrased awkwardly.  Perhaps: _Changed the load balancer to require that kubernetes services are resolved via the control plane._

CHANGES.md
olix0r: _Fixed a bug where errors were not counted in `response_total` metrics_ ?

CHANGES.md
hawkw: Oh, I like that. @kleimkuhler and I bikeshedded over this a bit offline & had a hard time coming up with a concise description, but yours is much better IMO.

CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+    `NET_ADMIN` capability is omitted from the control plane's PSP
+* Controller
+  * Fixed security context values so that injection does not fail on
+    restricted PSPs (thanks @codeman9!)
+  * Fixed webhook failure policy in order to avoid timing issues in HA mode
+  * Introduced control plane's PSP and RBAC resources into Helm templates;
+    these policies are only in effect if the PSP admission controller is
+    enabled
+  * Fixed MWC namespace value so that when installing multiple control planes,
+    there is a unique configuration for each one
```
ihcsim: User action is required here, if they are using the `linkerd upgrade` command to upgrade the latest edge. After the upgrade, they will see two mutating webhook configurations:

```
k get mutatingwebhookconfigurations.admissionregistration.k8s.io
NAME                                    CREATED AT
linkerd-proxy-injector-webhook-config   2019-06-13T22:22:55Z
linkerd-linkerd-proxy-injector-webhook-config 2019-06-13T22:22:55Z

```
They will need to manually remove the `linkerd-proxy-injector-webhook-config` mwc. This isn't relevant if they are doing a fresh installation of the edge release.

CHANGES.md
kleimkuhler: _Include errors in `response_total` metrics_?

I'd like to avoid "fixed a bug ..." bullets

CHANGES.md
```diff
@@ -1,3 +1,27 @@
+## edge-19.6.2
+
+* CLI
+  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
+    `NET_ADMIN` capability is omitted from the control plane's PSP
+* Controller
+  * Fixed security context values so that injection does not fail on
+    restricted PSPs (thanks @codeman9!)
+  * Fixed webhook failure policy in order to avoid timing issues in HA mode
```
ihcsim: The reason for that change is to avoid un-injected applications from being deployed, if the proxy injector fails as a result of unexpected errors (like network timeout etc.). Something like:

_Default the webhook failure policy to "fail", to account for unexpected errors during auto-proxy injection. This ensures un-injected applications aren't deployed._

CHANGES.md
```diff
@@ -1,3 +1,27 @@
+## edge-19.6.2
+
+* CLI
+  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
+    `NET_ADMIN` capability is omitted from the control plane's PSP
```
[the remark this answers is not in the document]
kleimkuhler: Oh no problem thanks for the clarification on that!

CHANGES.md
kleimkuhler: Is this included in docs? I think this information is better conveyed through docs rather than the release notes

ihcsim: sgtm. yeah, probably not needed for an edge release.

CHANGES.md
```diff
@@ -1,3 +1,30 @@
...
+    are not deployed
+  * Introduced control plane's PSP and RBAC resources into Helm templates;
+    these policies are only in effect if the PSP admission controller is
+    enabled
+  * Fixed MWC namespace value so that when installing multiple control planes,
+    there is a unique configuration for each one
+  * Removed `UPDATE` operation from proxy-injector webhook because pod
+    mutations are disallowed during update operations
+* Proxy
+  * The `l5d-override-dst` header is now honored for inbound service profile
```
hawkw: TIOLI, but personally I think 
```suggestion
  * The `l5d-override-dst` header is now used for inbound service profile
```
might be better

CHANGES.md
```diff
@@ -1,3 +1,30 @@
+## edge-19.6.2
+
+* CLI
+  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
+    `NET_ADMIN` capability is omitted from the CNI-enabled control plane's PSP
+* Controller
+  * Fixed security context values so that injection does not fail on
```
ihcsim: Can we change this to something like _Defined least privilege default security context values for the proxy container so that auto-injection does not fail on..._?

kleimkuhler: Wording a bit differently but sure!

PR #2945, opened 2019-06-14
travis integration test cleanup

dadjeibaah:
After #2887, integration tests will now leave stale validating and mutating webhook configurations since a new one is generated with a namespace name.

This PR adds a couple of clean up steps to the `test-cleanup` script to remove configs and also adds steps in the `.travis.yml` to clean up configs that were generated by Linkerd CLI's prior to #2887.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,7 +161,11 @@ jobs:
 
       script:
         - |
-          # Run integration tests.
+          version="$(./linkerd version --client --short | tr -cd '[:alnum:]-')"
+          # Cleanup previous integration test artifacts if present
+          ./bin/test-cleanup linkerd
+          kubectl delete mutatingwebhookconfigurations linkerd-proxy-injector-webhook-config
+          kubectl delete validatingwebhookconfigurations linkerd-sp-validator-webhook-config
           version="$(./linkerd version --client --short | tr -cd '[:alnum:]-')"
           ./bin/test-run `pwd`/linkerd linkerd-$version
         - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,8 +161,8 @@ jobs:
 
       script:
         - |
+          # Run integration tests.
           version="$(./linkerd version --client --short | tr -cd '[:alnum:]-')"
-          # Cleanup previous integration test artifacts if present
           ./bin/test-cleanup linkerd-$version
           ./bin/test-run `pwd`/linkerd linkerd-$version
         - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ branches:
 stages:
   - name: test
   - name: docker-deploy
-    #if: type != pull_request
+    if: type != pull_request
   - name: integration-test
-   # if: type != pull_request
+    if: type != pull_request
 
 jobs:
   include:
@@ -111,7 +111,7 @@ jobs:
       after_success:
         - bin/docker-push-deps
         - bin/docker-push $LINKERD_TAG
-        # - bin/docker-retag-all $LINKERD_TAG master && bin/docker-push master
+        - bin/docker-retag-all $LINKERD_TAG master && bin/docker-push master
         - target/cli/linux/linkerd install --control-plane-version=$LINKERD_TAG |tee linkerd.yml
         - kubectl -n linkerd apply -f linkerd.yml --prune --selector='linkerd.io/control-plane-component'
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -161,7 +161,11 @@ jobs:
 
       script:
         - |
-          # Run integration tests.
+          version="$(./linkerd version --client --short | tr -cd '[:alnum:]-')"
+          # Cleanup previous integration test artifacts if present
+          ./bin/test-cleanup linkerd
+          kubectl delete mutatingwebhookconfigurations linkerd-proxy-injector-webhook-config
+          kubectl delete validatingwebhookconfigurations linkerd-sp-validator-webhook-config
           version="$(./linkerd version --client --short | tr -cd '[:alnum:]-')"
```
alpeb: You can get rid of this line, given it's already in 164

.travis.yml
```diff
@@ -161,7 +161,11 @@ jobs:
 
       script:
         - |
-          # Run integration tests.
+          version="$(./linkerd version --client --short | tr -cd '[:alnum:]-')"
+          # Cleanup previous integration test artifacts if present
+          ./bin/test-cleanup linkerd
```
alpeb: I think we should do the same as in `after_script`:
```
./bin/test-cleanup linkerd-$version
```
as the intention would be to get rid of garbage from previous test integration runs that didn't finish and thus couldn't run `after_script`

dadjeibaah: I think that should take care of that since `./bin/test-cleanup` will delete all namespaces with the `linkerd-*` prefix This includes `linkerd-$version`. So that should clean up namespaces that have been left behind by other tests.

ihcsim: I agree with @alpeb that the integration test should only delete what it creates. AFAIK, the integration test doesn't create the `linkerd` namespace, right? If one exists, it's probably a dev namespace; i.e. others might run the integration test alongside their WIP dev stuff.

bin/test-cleanup
```diff
@@ -24,6 +24,14 @@ if ! clusterroles=$(kubectl --context=$k8s_context get clusterroles -oname | gre
   echo "no clusterroles found for [$linkerd_namespace]" >&2
 fi
 
-if [[ $namespaces || $clusterrolebindings || $clusterroles ]]; then
-  kubectl --context=$k8s_context delete --wait=false $namespaces $clusterrolebindings $clusterroles
+if ! webhookconfigs=$(kubectl --context=$k8s_context get mutatingwebhookconfigurations -oname | grep -E  "/linkerd-$linkerd_namespace(-|$)"); then
```
ihcsim: Not sure if this and the `if` in line 31 are necessary, since you are calling `kubectl delete` on all of them anyway at line 36. Do they just echo a warning?

bin/test-cleanup
```diff
@@ -24,6 +24,14 @@ if ! clusterroles=$(kubectl --context=$k8s_context get clusterroles -oname | gre
...
+  echo "no mutatingwebhookconfigurations found for [$linkerd_namespace]" >&2
+fi
+
+if ! validatingconfigs=$(kubectl --context=$k8s_context get validatingwebhookconfigurations -oname | grep -E  "/linkerd-$linkerd_namespace(-|$)"); then
+  echo "no validatingwebhookconfigurations found for [$linkerd_namespace]" >&2
+fi
+
+if [[ $namespaces || $clusterrolebindings || $clusterroles || $webhookconfigs || $validatingconfigs ]]; then
```
ihcsim: This can be replaced with `kubectl delete --ignore-not-found=false`. I didn't check when this option was introduced though. Mine is v1.13.0.

.travis.yml
dadjeibaah: Ah! I think I get it now. @ihcsim @alpeb  you are right in that this would interfere with other linkerd installs that are running on the same cluster.  Thanks for the explanation @ihcsim!

bin/test-cleanup
dadjeibaah: I added these lines to be consistent with the other cluster level resources and it appears to be done so that we log warnings if those resources aren't found.

ihcsim:
I took a look at the l5d-bot cluster, but couldn't find the cluster role that was causing the integration test to fail. The travis integration test [run](https://travis-ci.org/linkerd/linkerd2/builds/546928164?utm_source=github_status&utm_medium=notification) for this branch had. Maybe just fix up the comment out lines in .travis.yaml and merge it?

dadjeibaah:
@ihcsim thanks for taking a look, I'll fix up the commented out lines and get this merged.

PR #2920, opened 2019-06-11
Introduce Control Plane's PSP and RBAC resources into Helm templates

ihcsim:
This PR updates `linkerd install` to always install the control plane's PSP and RBAC resources. These policies are only in-effect if the PSP admission controller is enabled. Only the Helm templates are changed.

Test cases:
```
bin/linkerd version --client
Client version: git-de1b9cf8

# expect linkerd-linkerd-control-plane PSP, linkerd-psp role and linkerd-psp role-binding
# to be created
bin/linkerd install| [less | kubectl apply -f - ]

# multi-stage install
# expect linkerd-linkerd-control-plane PSP, linkerd-psp role and linkerd-psp role-binding 
# to be created during config stage
bin/linkerd install config| [less | kubectl apply -f - ]
bin/linkerd install control-plane [less | kubectl apply -f - ]

# upgrade
linkerd version --client
Client version: edge-19.6.1

# expect linkerd-linkerd-control-plane PSP, linkerd-psp role and linkerd-psp role-binding
# to be created
bin/linkerd upgrade| [ less | kubectl apply -f - ]

# cni on GKE; nodes must be ubuntu since cos has read-only root filesystem
# expect linkerd-linkerd-cni PSP, linkerd-cni role and linkerd-cni role binding to be created
bin/linkerd install-cni --dest-cni-bin-dir=/home/kubernetes/bin|k apply -f -

# expect linkerd-linkerd-control-plane PSP, linkerd-psp role and linkerd-psp role-binding 
# to be created
bin/linkerd install --linkerd-cni-enabled | [ less | kubectl apply -f -]

# cni with multi-stage install on GKE
# expect linkerd-linkerd-control-plane PSP to be created w/o NET_ADMIN capability
bin/linkerd install config --linkerd-cni-enabled| kubectl apply -f -

# expect control plane components to be installed with proxy-init init containers
bin/linkerd install control-plane --linkerd-cni-enabled| kubectl apply -f -
```

Signed-off-by: Ivan Sim <ivan@buoyant.io>

Fixes https://github.com/linkerd/linkerd2/issues/2892

Issue #2849, opened 2019-05-24
linkerd check should check for PodSecurityPolicy admission controller


siggy:
## Background

Linkerd user @jamessthompson reported a failing `linkerd check --pre`, due to PodSecurityPolicy:

```bash
$ linkerd check --pre --verbose
...
pre-kubernetes-capability
-------------------------
× has NET_ADMIN capability
    found 7 PodSecurityPolicies, but none provide NET_ADMIN
    see https://linkerd.io/checks/#pre-k8s-cluster-net-admin for hints
```

Despite the failed check, Linkerd successfully installed. This is likely due to the PodSecurityPolicy admission controller not running.

## Reproduction

```bash
gcloud beta container clusters create psp-test --enable-pod-security-policy
```

## Proposal

Modify `linkerd check --pre` to validate whether the `PodSecurityPolicy` is running, in addition to check the actual `PodSecurityPolicy`'s.

If the admission controller is not running, but `PodSecurityPolicy`'s still exist that would disallow `NET_ADMIN`, issue a warning.

Additionally, if the admission controller is running, and no `PodSecurityPolicy`'s are found, this check should fail. This is not the case in the current implementation.

Relevant code: https://github.com/linkerd/linkerd2/blob/bd4c2788fa19f31b0fb2f32254ac6fd52c737565/pkg/healthcheck/healthcheck.go#L985-L988

ihcsim:
@jamessthompson Can you take a look at the Linkerd deployments to see whether the `kubernetes.io/psp: ` annotation has been added. If the PSP admission controller is running, and the service account has the right RBAC to `use` the policy, this annotation will show you the policy that is in-effect.

Also speaking of RBAC, I think you need to make sure all the Linkerd service accounts have the correct RBAC to `use` the PSP policy(ies), because when it comes to non-pod workloads like `Deployment`, the service accounts are the ones creating the pods. See example [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#service-account-permissions).

jamessthompson:
That annotation does not exist which matches what we determined yesterday.  This ticket is more about providing clarity during installs for people like me that have never used the PSP admissions controller but have 3rd party charts installed that added PSPs.  I thought I had a broken install and reached out on slack looking for an example of a valid linkerd PSP.

alpeb:
[...]

PR #2958, opened 2019-06-18
Make NET_ADMIN check a warning, add PSP check

### LINKED ISSUE #2884: NET_ADMIN capability issue with linkerd --pre check
### What is the issue?
On a freshly installed cluster in AKS. Ran `linkerd check --pre`. All checks pass except for "pre-kubernetes-capability" See output below.

FWIW, the linkerd install ends up working and seems to be functioning correctly. 

#### `linkerd check` output

```text
pre-kubernetes-capability
-------------------------
× has NET_ADMIN capability
    found 6 PodSecurityPolicies, but none provide NET_ADMIN
    see https://linkerd.io/checks/#pre-k8s-cluster-net-admin for hints
```

### Environment

- Kubernetes Version: 1.12.8
- Cluster Environment: AKS
- Host OS: Ubuntu
- Linkerd version: stable-2.3.1

siggy:
`linkerd check` validates whether PSP's exist, and if the caller has the
`NET_ADMIN` capability. This check was previously failing if `NET_ADMIN`
was not found, even in the case where the PSP admission controller was
not running. Related, `linkerd install` now includes a PSP, so
`linkerd check` should also validate that the caller can create PSP's.

Modify the `NET_ADMIN` check to warn, but not fail, if PSP's are found
but the caller does not have `NET_ADMIN`. Update the warning message to
mention that this is only a problem if the PSP admission controller is
running (and will only be a problem during injection, since #2920
handles control plane installation by adding its own PSP).

Also introduce a check to validate the caller can create PSP's.

Fixes #2884, #2849

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -310,6 +310,13 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
+					description: "can create PodSecurityPolicies",
+					hintAnchor:  "pre-k8s",
+					check: func(context.Context) error {
+						return hc.checkCanCreate(hc.ControlPlaneNamespace, "policy", "v1beta1", "podsecuritypolicies")
+					},
+				},
+				{
 					description: "can create ServiceAccounts",
 					hintAnchor:  "pre-k8s",
 					check: func(context.Context) error {
@@ -352,6 +359,7 @@ func (hc *HealthChecker) allCategories() []category {
 				{
 					description: "has NET_ADMIN capability",
 					hintAnchor:  "pre-k8s-cluster-net-admin",
+					warning:     true,
 					check: func(context.Context) error {
 						return hc.checkNetAdmin()
 					},
@@ -408,7 +416,7 @@ func (hc *HealthChecker) allCategories() []category {
 			checkers: []checker{
 				{
 					description: "control plane components ready",
-					hintAnchor:  "l5d-existence-psp",
+					hintAnchor:  "l5d-existence-psp", // needs https://github.com/linkerd/website/issues/272
 					fatal:       true,
 					check: func(context.Context) error {
 						controlPlaneReplicaSet, err := hc.kubeAPI.GetReplicaSets(hc.ControlPlaneNamespace)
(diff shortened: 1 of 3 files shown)


grampelberg:
With the control plane PSP, `NET_ADMIN` isn't required for `install` but it is still required for workloads right? Maybe there should be a check in `inject`?

grampelberg:
Another idea, as we're moving towards auto-inject being the one true way (tm), and auto-inject can return errors. What about checking to see if the attached service account has a PSP that includes `NET_ADMIN` and returning an error if it doesn't (assuming CNI isn't there).

This all feels like 100% a separate issue and shouldn't block the merge of this specific PR as most of my confusion seems to have stemmed from me making an invalid assumption about what the check was actually validating.

siggy:
@grampelberg @ihcsim Thanks for the feedback re: proxy injector checking for PSP, I've filed #2968 to track.

I've put the `NET_ADMIN` check back into this PR, so the overall diff is now smaller. Since `NET_ADMIN` is still required for injection, I thought it prudent to continue validating this in `linkerd check`. To solve #2884 and #2849 specifically, I've changed the `NET_ADMIN` check into a warning, to inform the user that there may be a problem if the PSP admission controller is running (if it's not running, it's a non-issue).

ihcsim:
I'm not sure if it's necessary to check for `NET_ADMIN` during `linkerd check --pre`, since `linkerd install` and `linkerd install config` will create the control plane PSP with that capability. I wonder if adding back this check will lead users to think that they must have a PSP with `NET_ADMIN` prior to installation. 

Also, did your change leave out the CNI-enabled condition for this check?

siggy:
@ihcsim You're right that `linkerd install` and `linkerd install config` will work regardless of whether the `NET_ADMIN` check passes, since the control plane creates a PSP anyway. I left the `NET_ADMIN` check in place because that capability is still required for proxy injection. To account for this, I've changed the `NET_ADMIN` check from an error to a warning (`linkerd check` still succeeds), and updated the message to read:

```
pre-kubernetes-capability
-------------------------
‼ has NET_ADMIN capability
    found 1 PodSecurityPolicies, but none provide NET_ADMIN, proxy injection will fail if the PSP admission controller is running
    see https://linkerd.io/checks/#pre-k8s-cluster-net-admin for hints

...
Status check results are √
```

Re: the cni check, I've left the existing behavior, where this command skips the `NET_ADMIN` check:
```
linkerd check --linkerd-cni-enabled --pre
```

PR #2963, opened 2019-06-18
Add sideEffects field to Webhooks

### LINKED ISSUE #2850: Admission controller not supporting --server-dry-run
Kubernetes >=1.12 added the new `--server-dry-run` flag (alpha in 1.12, beta in 1.13), which allows to test manifests against the apiserver and all the admission controllers. This is a great utility to fully test the integrity of a deployment as part CI and avoid any pitfalls caused by multiple admission controllers.

Furthermore the `--server-dry-run` can be used along with `--output`/`-o` to retrieve the computed result.

The admission controller should support `--server-dry-run`. An introductory post can be found here: https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/
If I understand the issue correctly only a minimal change to linkerd's `admissionregistration.k8s.io/v1beta1.Webhook` is necessary, it needs to specify the `sideEffects`-field to indicate that it does not have side-effects on dry-runs. 
If there are side-effects I am unaware of (e.g. admission controller creates tls identities), the admission controller needs to disable the side-effects for workloads marked as _dry-run_.

The user can then use the `--server-dry-run` to test deployments utilizing the linkerd admision controller, e.g.:
```bash
# test deployment
kubectl apply --server-dry-run -f my-deployment.yaml
# retrieve computed output
kubectl apply --server-dry-run -f my-deployment.yaml -o yaml
```

Pothulapati:
Fixes #2850 

This PR adds the `sideEffects` field to the webhooks by default. As sideEffects field is supported from kubernetes version ge 1.12, the PR also bumps the minSupported version for `check` to 1.12

For users, to run on clusters < 1.12, they can use the `--omit-webhook-side-effects` to omit the `sideEffects` during installation, and this is flag is persisted during upgrades.

@ihcsim @grampelberg


cli/cmd/install.go
```diff
@@ -119,15 +120,16 @@ type (
...
-		identityOptions     *installIdentityOptions
+		controlPlaneVersion    string
+		controllerReplicas     uint
+		controllerLogLevel     string
+		highAvailability       bool
+		controllerUID          int64
+		disableH2Upgrade       bool
+		noInitContainer        bool
+		skipChecks             bool
+		OmitWebhookSideEffects bool
```
ihcsim: This doesn't need to be exported.

cli/cmd/install.go
```diff
@@ -435,6 +438,10 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 		&options.identityOptions.clockSkewAllowance, "identity-clock-skew-allowance", options.identityOptions.clockSkewAllowance,
 		"The amount of time to allow for clock skew within a Linkerd cluster",
 	)
+	flags.BoolVar(
+		&options.OmitWebhookSideEffects, "omit-webhook-side-effects", options.OmitWebhookSideEffects,
+		"Omit the sideEffects flag in the webhook manifests, to run on Kubernetes versions less than 1.12",
```
ihcsim: I think we should emphasize to the users that if they are on k8s pre-1.12, they must provide this flag, during install and upgrade. How about, _Omit the sideEffects flag in the webhook manifests. This flag must be provided during install or upgrade for Kubernetes pre 1.12_

cli/cmd/install.go
Pothulapati: Ah! Totally missed this! Sorry!

Pothulapati:
Updated. :)  @ihcsim

ihcsim:
Thanks!

I am using this branch to do some testing on `upgrade` atm. I notice that we need to update the `config.Global` protobuf to include this new field too. Otherwise, pre-1.12 users will have to keep providing the `--omit-webhook-side-effect` flag every time they perform an upgrade.

Here's a diff for the needed change:
```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 53c4f406..359bfab8 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -761,10 +761,11 @@ func (options *installOptions) configs(identity *pb.IdentityContext) *pb.All {
 
 func (options *installOptions) globalConfig(identity *pb.IdentityContext) *pb.Global {
 	return &pb.Global{
-		LinkerdNamespace: controlPlaneNamespace,
-		CniEnabled:       options.noInitContainer,
-		Version:          options.controlPlaneVersion,
-		IdentityContext:  identity,
+		LinkerdNamespace:       controlPlaneNamespace,
+		CniEnabled:             options.noInitContainer,
+		Version:                options.controlPlaneVersion,
+		IdentityContext:        identity,
+		OmitWebhookSideEffects: options.omitWebhookSideEffects,
 	}
 }
 
diff --git a/proto/config/config.proto b/proto/config/config.proto
index 19e3320f..5ee8975d 100644
--- a/proto/config/config.proto
+++ b/proto/config/config.proto
@@ -23,6 +23,8 @@ message Global {
   IdentityContext identity_context = 4;
 
   AutoInjectContext auto_inject_context = 6 [deprecated=true];
+
+  bool omitWebhookSideEffects = 7;
 }
 
 message Proxy {
```
I omitted the unit tests diff because they are too long. You can update them with `go test ./cli/... --update`.

ihcsim:
@Pothulapati Thanks for the quick turnaround.

I discovered a tricky issue in `upgrade`. If I install the control plane with the `--omit-webhook-side-effects=true` option (pretending I'm on k8s 1.11), and then try to upgrade it with `--omit-webhook-side-effects=false` (say, now I'm on k8s 1.15), the `config.Global.OmitWebhookSideEffects` config becomes out-of-sync.

To reproduce:
```sh
bin/linkerd version
Client version: git-7e06dde6
Server version: unavailable

# omit the webhook side effects field
bin/linkerd install --omit-webhook-side-effects=true |k apply -f -

# these are correct
k -n linkerd get validatingwebhookconfigurations.admissionregistration.k8s.io linkerd-sp-validator-webhook-config -ojsonpath='{.webhooks[*].sideEffects}'
Unknown%
k -n linkerd get mutatingwebhookconfigurations.admissionregistration.k8s.io linkerd-proxy-injector-webhook-config -ojsonpath='{.webhooks[*].sideEffects}'
Unknown%
k -n linkerd get cm linkerd-config -ojsonpath='{.data.global}' | jq . | grep -i omitwebhooksideeffects
  "omitWebhookSideEffects": true

# do an upgrade to enable the field
bin/linkerd upgrade --omit-webhook-side-effects=false|k apply -f -

# this is out-of-sync; it should be false
k -n linkerd get cm linkerd-config -ojsonpath='{.data.global}' | jq . | grep -i omitwebhooksideeffects
  "omitWebhookSideEffects": true

# these are correct
k -n linkerd get mutatingwebhookconfigurations.admissionregistration.k8s.io linkerd-proxy-injector-webhook-config -ojsonpath='{.webhooks[*].sideEffects}'
None%
k -n linkerd get validatingwebhookconfigurations.admissionregistration.k8s.io linkerd-sp-validator-webhook-config -ojsonpath='{.webhooks[*].sideEffects}'
None%
```

We just need to make sure that `upgrade` picks up the new value of the flag:
```diff
diff --git a/cli/cmd/upgrade.go b/cli/cmd/upgrade.go
index ee5c43e9..94f0e430 100644
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -218,6 +218,7 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
 		configs.GetGlobal().Version = options.controlPlaneVersion
 	}
 	configs.GetInstall().Flags = options.recordedFlags
+	configs.GetGlobal().OmitWebhookSideEffects = options.omitWebhookSideEffects
 
 	var identity *installIdentityValues
 	idctx := configs.GetGlobal().GetIdentityContext()
```

Pothulapati:
Understood and updated :)
 @ihcsim

Pothulapati:
@ihcsim dry-run dosen't effect anything with emojivoto because, it is a deployment and our webhooks mutate only pods as in dry-run the deployment isnt actually run, `dry-run` without `sideEffects:None` dosent fail, try with a pod deployment with linkerd injection label, it will fail when `dry-run` is enabled and `sideEffects:None` is not present :)

ihcsim:
@Pothulapati Ha! Thanks for the tips. You are right. The `--server-dry-run` is in-effect only with pod workloads, because the `Deployment` is never persisted by the api server. Hence, our proxy injector isn't triggered, as no pods are created.

E.g. on a PSP-enabled cluster with limit range defined, I can use
```
kubectl apply --server-dry-run -f my-pod.yaml -o yaml
```
to see the final injected YAML of my pod. Also, with service profile validator, I can see the `sp-validator` picks up the admission requests in dry run mode, instead of rejecting it.

PR #2964, opened 2019-06-18
Set logrus timestamps on public-api and web startup

rmars:
We can have logrus print full timestamps instead of the time elapsed since starting in the controller go processes. I think this might be more readable/helpful than the current display. Let me know what you think!

Before:
Screen Shot 2019-06-18 at 1 57 59 PM

After:
Screen Shot 2019-06-18 at 1 42 42 PM

### DIFF
--- a/controller/cmd/public-api/main.go
+++ b/controller/cmd/public-api/main.go
@@ -29,6 +29,10 @@ func main() {
 	ignoredNamespaces := flag.String("ignore-namespaces", "kube-system", "comma separated list of namespaces to not list pods from")
 	flags.ConfigureAndParse()
 
+	// set log timestamps
+	formatter := &log.TextFormatter{FullTimestamp: true}
+	log.SetFormatter(formatter)
+
 	stop := make(chan os.Signal, 1)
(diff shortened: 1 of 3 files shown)


olix0r:
We should either change this for all of the controller pods or none of them...

alpeb:
Even without the changes in the branch, when using `kubectl logs` all our containers logs have already a timestamp format, for some reason.
When ran locally through `go run` we do need to explicit the formatting like in this branch. Trying to figure why...

alpeb:
Ok got it. When a TTY is detected, the default format is elapsed timestamp. When there's no TTY (our scratch containers) the default format is full timestamps.

So I think we still can do
```
	formatter := &log.TextFormatter{FullTimestamp: true}
	log.SetFormatter(formatter)
```
to get a uniform formatting regardless of the environment, but probably better to do it in `pkg/flags/flags.go`'s `ConfigureAndParse()` which is called by all our servers' main.go's

rmars:
ooooooo thanks @alpeb! that's useful. I'll move it to `ConfigureAndParse()`

PR #2965, opened 2019-06-18
Add Edges table to resource detail view of dashboard

### LINKED ISSUE #2709: Expose identity of edges in the dashboard

scottcarol:
This PR adds an Edges table to the resource detail view if one or more edges are returned going to/from that specific resource. A tooltip icon to the right of the table title explains what "Edges" data is. If a specific resource is not part of any edges, the Edges table should not show.

You can check the data by running `linkerd edges`.

Screen Shot 2019-06-18 at 2 27 32 PM

Resolves #2709.

### DIFF
--- a/web/app/js/components/EdgesTable.jsx
+++ b/web/app/js/components/EdgesTable.jsx
@@ -31,7 +31,7 @@ const edgesColumnDefinitions = (PrefixedLink, namespace, type) => {
       filter: d => d.dst.name,
       render: d => {
         // check that the destination is a k8s resource with a name we can link to
-        if (namespace && type && d.dst && d.src.name) {
+        if (namespace && type && d.dst && d.dst.name) {
           return (
             <PrefixedLink to={"/namespaces/" + namespace + "/" + type + "s/" + d.dst.name}>
               {d.dst.name}
--- a/web/app/js/components/EdgesTable.jsx
+++ b/web/app/js/components/EdgesTable.jsx
@@ -71,8 +71,8 @@ const edgesColumnDefinitions = (PrefixedLink, namespace, type) => {
   ];
 };
 
-const tooltipText = "Edges show the source, destination name and identity " +
-  "for proxied connections. If no identity is known, a message is displayed.";
+const tooltipText = `Edges show the source, destination name and identity
+  for proxied connections. If no identity is known, a message is displayed.`;
 
 class EdgesTable extends React.Component {
   static propTypes = {
--- a/web/app/js/components/EdgesTable.jsx
+++ b/web/app/js/components/EdgesTable.jsx
@@ -7,7 +7,7 @@ import { withContext } from './util/AppContext.jsx';
 const edgesColumns = [
   {
     title: "Source",
-    dataIndex: "name",
+    dataIndex: "source",
     isNumeric: false,
     filter: d => d.src.name,
     render: d => d.src.name,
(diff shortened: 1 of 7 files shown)


grampelberg:
@scottcarol can the source/dst names be links? client/server would be nice too, but I don't think we have a way to display service accounts yet.

scottcarol:
@grampelberg yup, can do.

web/app/js/components/ResourceDetail.jsx
```diff
@@ -14,15 +15,19 @@ import TopRoutesTabs from './TopRoutesTabs.jsx';
...
 import _isNil from 'lodash/isNil';
 import _merge from 'lodash/merge';
 import _reduce from 'lodash/reduce';
+import { processEdges } from './util/EdgesUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 
 // if there has been no traffic for some time, show a warning
 const showNoTrafficMsgDelayMs = 6000;
+// resource types supported when querying API for edge data
+const possibleResourceType = ["daemonset", "deployment", "job", "pod", "replicationcontroller", "statefulset"];
```
rmars: nit: can we call this "edgeDataAvailableType" or `edgeDataAvailable` or something? `possible` seems kind of vague

web/app/js/components/ResourceDetail.jsx
```diff
@@ -347,6 +365,18 @@ export class ResourceDetailBase extends React.Component {
           title="TCP"
           isTcpTable={true}
           metrics={this.state.podMetrics} />
+
+        {
+          noEdges ? null :
```
rmars: if you're not using `noEdges` anywhere else, might as well just directly say `_isEmpty(this.state.edges)` here

web/app/js/components/EdgesTable.jsx
```diff
@@ -0,0 +1,113 @@
...
+      dataIndex: "message",
+      isNumeric: false,
+      filter: d => d.noIdentityMsg,
+      render: d => d.noIdentityMsg,
+      sorter: d => d.noIdentityMsg
+    }
+  ];
+};
+
+const tooltipText = "Edges show the source, destination name and identity " +
```
dadjeibaah: Super minor nit: You could do use backticks for this string. This will allow you to split the line if its too long without having to string concatenate.
```diff
diff --git a/web/app/js/components/EdgesTable.jsx b/web/app/js/components/EdgesTable.jsx
index ebf46541..48bdfb58 100644
--- a/web/app/js/components/EdgesTable.jsx
+++ b/web/app/js/components/EdgesTable.jsx
@@ -71,8 +71,8 @@ const edgesColumnDefinitions = (PrefixedLink, namespace, type) => {
   ];
 };
 
-const tooltipText = "Edges show the source, destination name and identity " +
-  "for proxied connections. If no identity is known, a message is displayed.";
+const tooltipText = `Edges show the source, destination name and identity 
+  for proxied connections. If no identity is known, a message is displayed.`;
 
 class EdgesTable extends React.Component {
   static propTypes = {
```

Issue #2959, opened 2019-06-18
Proxy-init fails if primary container drops capability NET_RAW


afirth:
## Bug Report
In edge 19.6.2, If a pod has DROP: [ALL] the init container vomits

### What is the issue?
It seems that the init container requires NET_RAW as well as NET_ADMIN @ https://github.com/linkerd/linkerd2/pull/2911/files#diff-2e89e743609d90f7ab63ec3498f78c8bR627

### How can it be reproduced?

GKE, 19.6.2
`linkerd install`
`helm template nginx-ingress | linkerd inject | k apply -f -`

### Logs, error output, etc
```
    initContainerStatuses:
    - containerID: docker://7db0b2af9b82572b36e394eb06683031978833852aa9dd360c24d73f6cf90d79
      image: gcr.io/linkerd-io/proxy-init:v1.0.0
      imageID: docker-pullable://gcr.io/linkerd-io/proxy-init@sha256:b88af208664c1c82a55867e03f41d22f564cddf873e1524c7a854296371a6eff
      lastState:
        terminated:
          containerID: docker://df461543658bcc4864e770a033fabe5106617a9ce2343c5f3cb523e8e5fd8936
          exitCode: 1
          finishedAt: "2019-06-18T14:07:11Z"
          message: |+
            2019/06/18 14:07:11 Tracing this script execution as [1560866831]
            2019/06/18 14:07:11 State of iptables rules before run:
            2019/06/18 14:07:11 > iptables -t nat -vnL
            2019/06/18 14:07:11 < iptables v1.6.0: can't initialize iptables table `nat': Permission denied (you must be root)
            Perhaps iptables or your kernel needs to be upgraded.

            2019/06/18 14:07:11 Aborting firewall configuration
            Error: exit status 3
----
      name: linkerd-init
      resources:
        limits:
          cpu: 100m
          memory: 50Mi
        requests:
          cpu: 10m
          memory: 10Mi
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          add:
          - NET_BIND_SERVICE
          - NET_ADMIN
          drop:
          - ALL
        privileged: false
        procMount: Default
        readOnlyRootFilesystem: true
        runAsNonRoot: false
        runAsUser: 0
```
### Possible solution
Adding NET_RAW to the primary container's capabilities allows the container to start

alpeb:
Thanks for the issue @afirth 
[...]

PR #2969, opened 2019-06-19
Add `NET_RAW` capability to the proxy-init container

ihcsim:
Add the `NET_RAW` capability to the proxy-init container. This ensures that if a PSP or the primary container's security context contains the `Drop: All` capability rule, the proxy-init container won't fail.

Also, updated the control plane PSP to match that defined in https://github.com/linkerd/website/issues/94.

Test cases can be found [here](https://gist.github.com/ihcsim/d94f4cda94c1cce1662e12a68499c400). Also tested with multi-stage installation.

Fixes https://github.com/linkerd/linkerd2/issues/2959.

### DIFF
--- a/chart/templates/psp.yaml
+++ b/chart/templates/psp.yaml
@@ -14,6 +14,7 @@ spec:
   {{- if not .NoInitContainer }}
   allowedCapabilities:
   - NET_ADMIN
+  - NET_RAW
   {{- end}}
   seLinux:
     rule: RunAsAny
--- a/chart/templates/psp.yaml
+++ b/chart/templates/psp.yaml
@@ -16,18 +16,32 @@ spec:
   - NET_ADMIN
   - NET_RAW
   {{- end}}
+  requiredDropCapabilities:
+  - ALL
+  hostNetwork: false
+  hostIPC: false
+  hostPID: false
   seLinux:
-    rule: RunAsAny
-  supplementalGroups:
     rule: RunAsAny
   runAsUser:
     rule: RunAsAny
+  supplementalGroups:
+    rule: MustRunAs
+    ranges:
+    - min: 1
+      max: 65535
   fsGroup:
-    rule: RunAsAny
+    rule: MustRunAs
+    ranges:
+    - min: 1
+      max: 65535
   volumes:
   - configMap
   - emptyDir
   - secret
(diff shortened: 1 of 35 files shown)


pkg/inject/inject.go
```diff
@@ -618,13 +622,14 @@ func (conf *ResourceConfig) injectProxyInit(patch *Patch, saVolumeMount *corev1.
 	capabilities := &corev1.Capabilities{}
 	if conf.pod.spec.Containers != nil && len(conf.pod.spec.Containers) > 0 {
 		if sc := conf.pod.spec.Containers[0].SecurityContext; sc != nil && sc.Capabilities != nil {
-			capabilities = sc.Capabilities
+			capabilities.Add = sc.Capabilities.Add
+			capabilities.Drop = sc.Capabilities.Drop
```
siggy: curious about this change: do we want to guarantee that we're _only_ copying the `Add` and `Drop` fields? if not, i'd revert this.

ihcsim: Yup, just the `Add` and `Drop`, as they are the only exported fields.

ihcsim: Also, previously, it ended by copying references to the slides. The end result is that, as  the sidecars' security context is updated, the primary container's is updated too :scream:.

chart/templates/psp.yaml
```diff
@@ -14,19 +14,34 @@ spec:
...
+    ranges:
+    - min: 1
+      max: 65535
   volumes:
   - configMap
   - emptyDir
   - secret
+  - projected
+  - downwardAPI
+  - persistentVolumeClaim
```
alpeb: Any particular reason these three types of volumes were added? I couldn't find where in the CP they're required.

ihcsim: Just going by the list in https://github.com/linkerd/website/issues/94. They are part of the [recommended minimum set](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems).

alpeb:
This tested fine in my local with k8s 1.14 :+1: 
 I'll give it a try with k8s 1.15 to see the PSP controller actually trying to mutate the sidecars and see how it goes :pray: :prayer_beads:

PR #2972, opened 2019-06-20
Have `GetOwnerKindAndName` be able to skip the cache

### LINKED ISSUE #2738: evaluate forced k8s informer resync

alpeb:
Refactored `GetOwnerKindAndName` so it can optionally skip the
shared informer cache and instead hit the k8s API directly.
Useful for the proxy injector, when the pod's replicaset got just
created and might not be in ready in the cache yet.

Fixes #2738, supersedes #2881

### DIFF
--- a/controller/k8s/api.go
+++ b/controller/k8s/api.go
@@ -369,8 +369,10 @@ func (api *API) GetObjects(namespace, restype, name string) ([]runtime.Object, e
 
 // GetOwnerKindAndName returns the pod owner's kind and name, using owner
 // references from the Kubernetes API. The kind is represented as the Kubernetes
-// singular resource type (e.g. deployment, daemonset, job, etc.)
-func (api *API) GetOwnerKindAndName(pod *corev1.Pod) (string, string) {
+// singular resource type (e.g. deployment, daemonset, job, etc.).
+// If skipCache is false we use the shared informer cache; otherwise we hit the
+// Kubernetes API directly.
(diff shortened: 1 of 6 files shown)


controller/k8s/api.go
```diff
@@ -382,18 +384,14 @@ func (api *API) GetOwnerKindAndName(pod *corev1.Pod) (string, string) {
...
 			rs, err = api.Client.AppsV1beta2().ReplicaSets(pod.Namespace).Get(parent.Name, metav1.GetOptions{})
-			if err != nil {
-				log.Errorf("failed to get replicaset from k8s %s/%s: %s", pod.Namespace, parent.Name, err)
-			} else {
-				log.Debugf("successfully recovered replicaset via k8s get request: %s/%s", pod.Namespace, parent.Name)
-			}
+		} else {
+			rs, err = api.RS().Lister().ReplicaSets(pod.Namespace).Get(parent.Name)
 		}
+
```
siggy: mind logging the error in case err != nil ?

PR #2976, opened 2019-06-20
Simplify port-forwarding code

alpeb:
Simplifies the establishment of a port-forwarding by moving the common logic into `PortForward.Init()`.

Stemmed from this [comment](https://github.com/linkerd/linkerd2/pull/2937#discussion_r295078800).

### DIFF
--- a/cli/cmd/metrics.go
+++ b/cli/cmd/metrics.go
@@ -160,6 +160,7 @@ func getMetrics(
 		return nil, err
 	}
 
+	defer portforward.Stop()
 	if err = portforward.Init(); err != nil {
 		fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
 	}
--- a/cli/cmd/metrics.go
+++ b/cli/cmd/metrics.go
@@ -160,17 +160,9 @@ func getMetrics(
 		return nil, err
 	}
 
-	defer portforward.Stop()
-
-	go func() {
-		err := portforward.Run()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
-			portforward.Stop()
-		}
-	}()
-
-	<-portforward.Ready()
+	if err = portforward.Init(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
+	}
 
 	metricsURL := portforward.URLFor("/metrics")
 	resp, err := http.Get(metricsURL)
--- a/controller/api/public/client.go
+++ b/controller/api/public/client.go
@@ -252,27 +252,7 @@ func NewExternalClient(controlPlaneNamespace string, kubeAPI *k8s.KubernetesAPI)
 		return nil, err
 	}
 
-	log.Debugf("Starting port forward on [%s]", apiURL)
-
-	wait := make(chan error, 1)
-
-	go func() {
-		if err := portforward.Run(); err != nil {
-			wait <- err
-		}
-
-		portforward.Stop()
-	}()
-
-	select {
-	case <-portforward.Ready():
-		log.Debugf("Port forward initialised")
-
-		break
-
-	case err := <-wait:
-		log.Debugf("Port forward failed: %v", err)
-
+	if err = portforward.Init(); err != nil {
 		return nil, err
 	}
 
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
@@ -160,6 +161,33 @@ func (pf *PortForward) Run() error {
 	}
 
 	return fw.ForwardPorts()
+}
+
+// Init creates and runs a port-forward connection
+func (pf *PortForward) Init(wait ...chan struct{}) error {
+	log.Debug("Starting port forward")
+
+	failure := make(chan error, 1)
+
+	go func() {
+		if err := pf.Run(); err != nil {
+			failure <- err
+		}
+		if len(wait) > 0 {
+			close(wait[0])
+		}
+	}()
+
+	select {
+	case <-pf.Ready():
+		log.Debug("Port forward initialised")
+		break
+	case err := <-failure:
+		log.Debugf("Port forward failed: %v", err)
+		return err
+	}
+
+	return nil
 }
 
 // Ready returns a channel that will receive a message when the port-forward
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -138,9 +138,9 @@ func newPortForward(
 	}, nil
 }
 
-// Run creates and runs the port-forward connection.
-// It blocks upon a successful connection; otherwise it returns an error.
-func (pf *PortForward) Run() error {
+// run creates and runs the port-forward connection.
+// When the connection is established it blocks until Stop() is called.
+func (pf *PortForward) run() error {
 	transport, upgrader, err := spdy.RoundTripperFor(pf.config)
 	if err != nil {
 		return err
@@ -166,18 +166,23 @@ func (pf *PortForward) Run() error {
 
 // Init creates and runs a port-forward connection.
 // This function blocks until the connection is established, in which case it returns nil.
+// It's the caller's responsibility to call Stop() to finish the connection.
 func (pf *PortForward) Init() error {
 	log.Debugf("Starting port forward to %s %d:%d", pf.url, pf.localPort, pf.remotePort)
 
-	failure := make(chan error, 1)
+	failure := make(chan error)
 
 	go func() {
-		if err := pf.Run(); err != nil {
+		if err := pf.run(); err != nil {
 			failure <- err
 		}
 
-		if _, ok := <-pf.stopCh; ok {
-			close(pf.stopCh)
+		select {
+		case <-pf.GetStop():
+			// stopCh was closed, do nothing
+		default:
+			// pf.run() returned for some other reason, close stopCh
+			pf.Stop()
 		}
 	}()
 
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -186,11 +186,11 @@ func (pf *PortForward) Init() error {
 		}
 	}()
 
-	// The `select` statement below depends on one of two outcomes from `pf.Run()`:
-	// 1) Succeed and block, causing a receive on `<-pf.Ready()`
+	// The `select` statement below depends on one of two outcomes from `pf.run()`:
+	// 1) Succeed and block, causing a receive on `<-pf.readyCh`
 	// 2) Return an err, causing a receive `<-failure`
 	select {
-	case <-pf.Ready():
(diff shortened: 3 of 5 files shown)


pkg/k8s/portforward.go
```diff
@@ -162,6 +163,33 @@ func (pf *PortForward) Run() error {
 	return fw.ForwardPorts()
 }
 
+// Init creates and runs a port-forward connection
```
adleong: It would be good to document the semantics of what and when this returns and what the wait channels are used for.  Is it the case that this method blocks until the port-forward is ready (at which point it returns nil) or until the port-forward fails before becoming ready (at which point it returns an err)?  And the wait channels are notified (closed) when the port-forward terminates successfully or unsuccessfully? 

The semantics here depend on the semantics of pf.Run so it would be good to document those as well on the Run method.

pkg/k8s/portforward.go
```diff
@@ -162,6 +163,33 @@ func (pf *PortForward) Run() error {
...
+	log.Debug("Starting port forward")
+
+	failure := make(chan error, 1)
+
+	go func() {
+		if err := pf.Run(); err != nil {
+			failure <- err
+		}
+		if len(wait) > 0 {
+			close(wait[0])
```
adleong: Why do we only close the first wait channel?

cli/cmd/metrics.go
```diff
@@ -160,17 +160,9 @@ func getMetrics(
...
-	go func() {
-		err := portforward.Run()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
-			portforward.Stop()
-		}
-	}()
-
-	<-portforward.Ready()
+	if err = portforward.Init(); err != nil {
```
adleong: why do we no longer need to call `Stop` here?

pkg/k8s/portforward.go
```diff
@@ -162,6 +163,33 @@ func (pf *PortForward) Run() error {
 	return fw.ForwardPorts()
 }
 
+// Init creates and runs a port-forward connection
+func (pf *PortForward) Init(wait ...chan struct{}) error {
+	log.Debug("Starting port forward")
```
adleong: should these logs include more context like the local and remote addresses of the forward?

pkg/k8s/portforward.go
alpeb: @adleong the plan was to have `Init` receive one or no args. Anyways, I'll use @siggy's suggestion to leverage the `stopCh` channel instead.

cli/cmd/metrics.go
alpeb: `Stop` there was to finish the port-forwarding in case of error. Now that's handled inside `Init`. The new comments in the upcoming push should make it clearer.

pkg/k8s/portforward.go
alpeb: Sounds like a good idea :+1:

alpeb:
Thanks for the feedback!

My last push gets rid of that new `wait` channel, and adds the comments suggested by @siggy and some other extra comments to make the blocking behavior clearer.

pkg/k8s/portforward.go
```diff
@@ -138,6 +139,7 @@ func newPortForward(
 }
 
 // Run creates and runs the port-forward connection.
+// It blocks upon a successful connection; otherwise it returns an error.
```
adleong: I'm a little unclear on the wording here.  Does this mean it blocks until the connection is destroyed?

pkg/k8s/portforward.go
```diff
@@ -162,6 +164,37 @@ func (pf *PortForward) Run() error {
...
+	log.Debugf("Starting port forward to %s %d:%d", pf.url, pf.localPort, pf.remotePort)
+
+	failure := make(chan error, 1)
+
+	go func() {
+		if err := pf.Run(); err != nil {
+			failure <- err
+		}
+
+		if _, ok := <-pf.stopCh; ok {
```
adleong: I'm having trouble figuring out the semantics of stopCh.  Is it the case that this channel is written to when the port-forward fails and the channel is closed when the port-forward is explicitly stopped?  

This line of code seems to wait until the channel is closed or written to, and then closes the channel if it hasn't been.  What is the purpose of this?

What happens if the channel fails after being stopped?  Could the portforward attempt to write a message to the closed channel and panic?

cli/cmd/metrics.go
adleong: if I'm reading the old code correctly, it looks like this was closing the port-forward after all bytes were finished reading from the response below.  Don't we still want to close the port-forward after we're finished reading?

ihcsim: I think `Stop` is still needed here. Otherwise, `pf.stopCh` will not be closed, implying that the connection opened by `pf.Run()` (specifically, [`fw.ForwardPorts()`](https://godoc.org/k8s.io/client-go/tools/portforward#PortForwarder.ForwardPorts)) won't be closed. `cli/cmd/dashoard.go` still calls `Stop` on `SIGINT`.

pkg/k8s/portforward.go
ihcsim: I wonder if this will end up competing with the `select` statement in `portforward.ForwardPorts()`, as seen [here](https://github.com/kubernetes/client-go/blob/2c6e35a5b9cf292580c2d8ecb7d141572997d3cd/tools/portforward/portforward.go#L230-L234).

pkg/k8s/portforward.go
```diff
@@ -162,6 +164,37 @@ func (pf *PortForward) Run() error {
...
 }
 
+// Init creates and runs a port-forward connection.
+// This function blocks until the connection is established, in which case it returns nil.
+func (pf *PortForward) Init() error {
+	log.Debugf("Starting port forward to %s %d:%d", pf.url, pf.localPort, pf.remotePort)
+
+	failure := make(chan error, 1)
+
+	go func() {
```
ihcsim: To confirm my understanding, the `Init()` method will exit when the connection is ready, but this goroutine will block (because `portforward.ForwardPorts()` is [blocking](https://github.com/kubernetes/client-go/blob/2c6e35a5b9cf292580c2d8ecb7d141572997d3cd/tools/portforward/portforward.go#L230-L234)) and live on until a message is sent to `pf.stopCh`, right? Unless the callers call `defer pf.Stop()` (panic etc.), I think this can cause goroutine leaks.

controller/api/public/client.go
```diff
@@ -252,27 +252,7 @@ func NewExternalClient(controlPlaneNamespace string, kubeAPI *k8s.KubernetesAPI)
...
-	log.Debugf("Starting port forward on [%s]", apiURL)
-
-	wait := make(chan error, 1)
-
-	go func() {
-		if err := portforward.Run(); err != nil {
-			wait <- err
-		}
-
-		portforward.Stop()
```
ihcsim: This is probably still needed to make sure `pf.stopCh` receives a message to signal the termination of the connection.

pkg/k8s/portforward.go
siggy: Yeah I think `Init()` should be documented that it's the caller's responsibility to call `Stop()`. In `cli/cmd/metrics.go`, this should be a trivial change. `controller/api/public/client.go` looks more involved, as there's an expectation that the portforward continues running after `NewExternalClient` returns. Since `NewExternalClient` is only ever called from the CLI, I'm less concerned about a resource leak, and would be ok deferring that change for a subsequent PR.

cli/cmd/metrics.go
alpeb: You're right, I was leaking `stopCh`. Will add an explicit call to `Stop()` on the calling side.

pkg/k8s/portforward.go
alpeb: Yeah that's right. I'll reword the comment to make it clearer.

pkg/k8s/portforward.go
alpeb: Right, `stopCh` is only used as a broadcast mechanism to signal we want the connection to end (by closing the channel). You can see in client-go's [portforward.go](https://github.com/kubernetes/client-go/blob/2c6e35a5b9cf292580c2d8ecb7d141572997d3cd/tools/portforward/portforward.go) its type is always `<-chan` so it doesn't get written to. For the same reason, there is no competition when reading from it; once it's closed all its readers will unblock.

@siggy That's a nice idiom. I had the same intention but I got the meaning of `_, ok := <- ch` wrong.

pkg/k8s/portforward.go
alpeb: I'll add `Stop()` to metrics.go.
According to @ihcsim explanation, client.go was already leaking. To fix this we'll have to surface something from portforward into `RunChecks()` that can signal the end of the check in order to close the channel. Agreed can be done in a separate PR.

controller/api/public/client.go
alpeb: If I understand correctly, line 264 was never reached because `portforward.Run()` above blocks until `Stop()` is called.

alpeb:
Great feedback, pointed out some misconceptions I had :sweat_smile:
Here's another round, which also "privatizes" the `run` function.

pkg/k8s/portforward.go
```diff
@@ -162,6 +164,42 @@ func (pf *PortForward) Run() error {
...
+		if err := pf.run(); err != nil {
+			failure <- err
+		}
+
+		select {
+		case <-pf.GetStop():
+			// stopCh was closed, do nothing
+		default:
+			// pf.run() returned for some other reason, close stopCh
```
adleong: Is this expected to ever happen?  Should we log/warn here?

siggy: This can happen if the connection unexpectedly closes:
https://github.com/kubernetes/client-go/blob/2c6e35a5b9cf292580c2d8ecb7d141572997d3cd/tools/portforward/portforward.go#L232
... though I wouldn't be opposed to logging here.

pkg/k8s/portforward.go
```diff
@@ -162,6 +164,42 @@ func (pf *PortForward) Run() error {
...
+			pf.Stop()
+		}
+	}()
+
+	// The `select` statement below depends on one of two outcomes from `pf.Run()`:
+	// 1) Succeed and block, causing a receive on `<-pf.Ready()`
+	// 2) Return an err, causing a receive `<-failure`
+	select {
+	case <-pf.Ready():
```
siggy: sorry i missed this on the last review, i think you can just do `<-pf.readyCh` and remove the `Ready()` method.

pkg/k8s/portforward.go
alpeb: Yeah I confirmed this is already being logged by k8s:
```bash
$ bin/go-run cli dashboard
Linkerd dashboard available at:
http://127.0.0.1:50750
Grafana dashboard available at:
http://127.0.0.1:50750/grafana
Opening Linkerd dashboard in the default browser
[27967:27986:0626/101432.351720:ERROR:browser_process_sub_thread.cc(221)] Waited 3 ms for network service
Opening in existing browser session.

# I'm killing Minkube here...

E0626 10:14:37.566713   27828 portforward.go:224] lost connection to pod
E0626 10:14:37.566770   27828 portforward.go:224] lost connection to pod
```

pkg/k8s/portforward.go
alpeb: Good :eye: , `Ready()` is not being used anywhere else :+1:

Issue #2974, opened 2019-06-20
Add context to UUID warning when running web server locally 


scottcarol:
#2668 allowed the web server to run without install config, simply warning when the UUID isn't available. However the log message can be confusing if someone using `bin/web dev` to debug a dashboard issue doesn't have any context: 

```
WARN[0000] failed to load uuid from install config: failed to read config file: open /var/run/linkerd/config/install: no such file or directory
```

I'd like to add `disregard warning if running web server locally` to the log message, however, does this add too much noise to the warning? @klingerf @rmars

klingerf:
@scottcarol That's totally fine by me! It should only printing in development mode anyway, assuming everything is working as expected.

scottcarol:
Thanks @klingerf - it prints with both `bin/web dev` and `bin/web run`, is that expected? I will add that clarification to the log message.

klingerf:
Yep -- that's expected.

PR #2977, opened 2019-06-21
Adding context to log warning

scottcarol:
Adds context to UUID warning when running web server locally. See #2974 for context.

### DIFF
--- a/web/main.go
+++ b/web/main.go
@@ -40,7 +40,7 @@ func main() {
 
 	installConfig, err := config.Install(pkgK8s.MountPathInstallConfig)
 	if err != nil {
-		log.Warnf("(disregard warning if running in development mode) failed to load uuid from install config: %s", err)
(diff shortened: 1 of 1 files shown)


web/main.go
```diff
@@ -40,7 +40,7 @@ func main() {
 
 	installConfig, err := config.Install(pkgK8s.MountPathInstallConfig)
 	if err != nil {
-		log.Warnf("failed to load uuid from install config: %s", err)
+		log.Warnf("(disregard warning if running in development mode) failed to load uuid from install config: %s", err)
```
rmars: This lgtm; was playing around with placing the explanation at the end, cause it feels slightly weird to me to have a warning that begins with "disregard warning".

TIOLI:
```
log.Warnf("failed to load uuid from install config: [%s] (disregard warning if running in development mode)", err)
```

scottcarol: That's a good solution!

PR #2985, opened 2019-06-21
Add release notes for edge-19.6.3

kleimkuhler:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,28 @@
+## edge-19.6.3
+
+* CLI
+  * Added a check for Pod Security Policy (PSP) permissions to `linkerd check`
+  * Changed `linkerd check` to only warn (not error) if PSPs are found, but
+    the caller does not have the `NET_ADMIN` capability
+* Controller
+  * Improved timestamp formatting in logs
+  * Default the mutating and validating webhook configurations `sideEffects` 
+    property to `None` to indicate that the webhooks have no side effects on
+    external resources
+* Proxy
+  * Added the `NET_RAW` capability to the proxy-init container to prevent it
+    from failing when the primary container has certain security contexts
+  * Fixed the handling of HTTP2 requests that don't have an `:authority`
+  * Introduced eager eviction to the router cache so that idle services are
+    removed even when the cache is not full 
+* Web UI
+  * Removed the "debug" page from the Linkerd dashboard while the functionality
+    of that page is being redesigned
+  * Reduced the dashboard package size and number of files loaded by importing
+    fewer icons
+  * Added an Edges table to the resource detail view that shows the source,
+    destination name, and identity for proxied connections
+
 ## edge-19.6.2
 
 * CLI
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,25 +1,21 @@
 ## edge-19.6.3
 
 * CLI
-  * Added a check for Pod Security Policy (PSP) permissions to `linkerd check`
-  * Changed `linkerd check` to only warn (not error) if PSPs are found, but
-    the caller does not have the `NET_ADMIN` capability
+  * Updated `linkerd check` to validate the caller can create
+    `PodSecurityPolicy` resources
 * Controller
-  * Improved timestamp formatting in logs
   * Default the mutating and validating webhook configurations `sideEffects` 
     property to `None` to indicate that the webhooks have no side effects on
-    external resources
+    other resources (thanks @Pothulapati!)
 * Proxy
-  * Added the `NET_RAW` capability to the proxy-init container to prevent it
-    from failing when the primary container has certain security contexts
-  * Fixed the handling of HTTP2 requests that don't have an `:authority`
-  * Introduced eager eviction to the router cache so that idle services are
-    removed even when the cache is not full 
+  * Added the `NET_RAW` capability to the proxy-init container to be compatible
+    with `PodSecurityPolicy`s that use `drop: all`
+  * Fixed the proxy rejecting HTTP2 requests that don't have an `:authority`
+  * Improved idle service eviction to reduce resource consumption for clients
+    that send requests to many services
 * Web UI
-  * Removed the "debug" page from the Linkerd dashboard while the functionality
+  * Removed the "Debug" page from the Linkerd dashboard while the functionality
     of that page is being redesigned
-  * Reduced the dashboard package size and number of files loaded by importing
-    fewer icons
   * Added an Edges table to the resource detail view that shows the source,
     destination name, and identity for proxied connections
 


CHANGES.md
```diff
@@ -1,3 +1,28 @@
+## edge-19.6.3
+
+* CLI
+  * Added a check for Pod Security Policy (PSP) permissions to `linkerd check`
+  * Changed `linkerd check` to only warn (not error) if PSPs are found, but
```
hawkw: Nit/TIOLI: How about
```suggestion
  * Changed `linkerd check` to warn (rather than error) if PSPs are found, but
```

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+    the caller does not have the `NET_ADMIN` capability
+* Controller
+  * Improved timestamp formatting in logs
+  * Default the mutating and validating webhook configurations `sideEffects` 
+    property to `None` to indicate that the webhooks have no side effects on
+    external resources
+* Proxy
+  * Added the `NET_RAW` capability to the proxy-init container to prevent it
```
hawkw: Does this go in the proxy heading, or should there be a separate proxy-init heading? What have we done previously?

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+    from failing when the primary container has certain security contexts
+  * Fixed the handling of HTTP2 requests that don't have an `:authority`
```
hawkw: How about
```suggestion
  * Fixed the proxy rejecting HTTP2 requests that don't have an `:authority`
```

CHANGES.md
```diff
@@ -1,3 +1,28 @@
+## edge-19.6.3
+
+* CLI
+  * Added a check for Pod Security Policy (PSP) permissions to `linkerd check`
+  * Changed `linkerd check` to only warn (not error) if PSPs are found, but
+    the caller does not have the `NET_ADMIN` capability
+* Controller
+  * Improved timestamp formatting in logs
```
olix0r: AIUI this only impacted development

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+  * Default the mutating and validating webhook configurations `sideEffects` 
+    property to `None` to indicate that the webhooks have no side effects on
+    external resources
+* Proxy
+  * Added the `NET_RAW` capability to the proxy-init container to prevent it
+    from failing when the primary container has certain security contexts
+  * Fixed the handling of HTTP2 requests that don't have an `:authority`
+  * Introduced eager eviction to the router cache so that idle services are
```
hawkw: This seems like it could be made a little less wordy somehow? What about something like 
> Changed the proxy to eagerly evict idle services from router caches, even when they are not full

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+    removed even when the cache is not full 
+* Web UI
+  * Removed the "debug" page from the Linkerd dashboard while the functionality
+    of that page is being redesigned
+  * Reduced the dashboard package size and number of files loaded by importing
+    fewer icons
```
scottcarol: I think you can omit the bullet starting with "reduced the dashboard package size" as it's a small reduction in package size and not going to affect users

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+
+* CLI
+  * Added a check for Pod Security Policy (PSP) permissions to `linkerd check`
+  * Changed `linkerd check` to only warn (not error) if PSPs are found, but
+    the caller does not have the `NET_ADMIN` capability
+* Controller
+  * Improved timestamp formatting in logs
+  * Default the mutating and validating webhook configurations `sideEffects` 
+    property to `None` to indicate that the webhooks have no side effects on
+    external resources
```
ihcsim: Sorry, my bad. I think we should replace _external_ with _other_ here, as the k8s doc seems to imply _all_ resources. Also, this is a contribution from Tarun.

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+* Proxy
+  * Added the `NET_RAW` capability to the proxy-init container to prevent it
+    from failing when the primary container has certain security contexts
+  * Fixed the handling of HTTP2 requests that don't have an `:authority`
+  * Introduced eager eviction to the router cache so that idle services are
+    removed even when the cache is not full 
+* Web UI
+  * Removed the "debug" page from the Linkerd dashboard while the functionality
```
scottcarol: Nit but "Debug" is capitalized in the dashboard so I'd capitalize here

CHANGES.md
```diff
@@ -1,3 +1,28 @@
+## edge-19.6.3
+
+* CLI
+  * Added a check for Pod Security Policy (PSP) permissions to `linkerd check`
+  * Changed `linkerd check` to only warn (not error) if PSPs are found, but
+    the caller does not have the `NET_ADMIN` capability
```
olix0r: This isn't very clear to me... Can we summarize these two points as:

* Added `PodSecurityPolicy` warnings to `linkerd check`.

Is that an accurate summary of the change?

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+* Controller
+  * Improved timestamp formatting in logs
+  * Default the mutating and validating webhook configurations `sideEffects` 
+    property to `None` to indicate that the webhooks have no side effects on
+    external resources
+* Proxy
+  * Added the `NET_RAW` capability to the proxy-init container to prevent it
+    from failing when the primary container has certain security contexts
```
ihcsim: I think a bit more specifics here will let users know whether this is relevant to them or not. So, _when the primary container has a `drop: ALL` capability security context_

olix0r: replace "to prevent it from failing when ..." with "to satisfy `PodSecurityPolicy` requirements"

CHANGES.md
```diff
@@ -1,3 +1,28 @@
...
+  * Fixed the handling of HTTP2 requests that don't have an `:authority`
+  * Introduced eager eviction to the router cache so that idle services are
+    removed even when the cache is not full
```
olix0r: Let's make this a little more user-facing...

* Improved idle service eviction to reduce resource consumption for clients that send requests to many services.

Or something like that?

CHANGES.md
olix0r: per @ihcsim, perhaps make this "to be compatible with `PodSecurityPolicy`s that use `drop: all`" ?

CHANGES.md
ihcsim: The task was to update `linkerd check` to ensure the caller has the RBAC permissions to create the control plane's PSP.

ihcsim: The 2nd bullet point is ok, but it isn't _the caller does not have_, it's _the existing PSPs do not have_. Thanks for calling it out. I missed that.

olix0r: Perhaps:

* Updated `linkerd check` to validate the caller can create `PodSecurityPolicy` resources.

CHANGES.md
kleimkuhler: I don't think we need to worry about introducing a separate heading for this. It could also go under "Controller", but I will probably keep it here.

CHANGES.md
kleimkuhler: I'll consolidate these two bullet points to the above. Thanks for the clarification

Issue #2954, opened 2019-06-17
Disable installation of multiple control planes in 2.4


ihcsim:
For 2.4, we've decided to not support multi control plane setup. The `linkerd install` and `linkerd check --pre` commands will be updated to error out if existing Linkerd global resources are detected on the k8s cluster.

The following is a list of items to be implemented:

- Label all ~global~ resources (cluster role, cluster role binding, namespace, service accounts, CRD) with the existing `linkerd.io/control-plane-ns` label
- Update `linkerd install` and `linkerd check --pre` to query by this label. If the result is non-empty, an error will be returned
- `linkerd check` should check the mutating webhook config to make sure the service's namespace (defined in the config) matches that of the control plane. Further testing require wit multi-stage install, as it might not have the proper RBAC to access globally scoped resources like the webhook config.

grampelberg:
Anything to pay attention to with 2.3 -> 2.4 upgrades?

ihcsim:
2.3 -> 2.4 upgrades will definitely be an important test case. AFAICT, it won't break anything, since we are just adding a new label to all previously unlabelled resources. I think it's a safe to assume that users running `linkerd upgrade` have RBAC permission to `edit` all Linkerd global resources.

PR #2987, opened 2019-06-24
Introduce global resources checks to install and multi-stage install

ihcsim:
This PR introduces the following two new checks to `pkg/healthcheck/healthcheck.go`:

`LinkerdPreInstallGlobalResourcesChecks` which checks for the existence of Linkerd global resources when user runs `linkerd check --pre`. The check fails if any of the Linkerd cluster roles, cluster role bindings, mutating webhook configuration, validating webhook configuration, custom resource definition or pod security policy exist. This error will point to new troubleshooting documentation informing the user that multi control plane setup isn't supported.

`LinkerdConfigNotExistsChecks` which checks if the `linkerd-config` config map exists. The check fails if the config map already exists. This error will point to new troubleshooting documentation informing the user can't install a control plane in the specified namespace, unless overridden with the `--ignore-cluster` option.

The checks use the `linkerd.io/control-plane-ns` label to select the resources. See https://github.com/linkerd/linkerd2/pull/2971.

The following are updates to install and multi-stage install:

Previously, `linkerd install` checks for existence of the `linkerd-config` config map to determine if the control plane installation is permitted. This is now changed to check for the existence of global resources, to ensure user can't install the control plane into a different namespace.

Previously, `linkerd install control-plane` checks for the existence of the specified control plane namespace to determine if `linkerd install config` has already been run. This is now changed to check for the existence of global resources that are installed by `linkerd install config`.

Previously, the `linkerd install` code calls the Kubernetes API directly. This is now changed to delegate the API interaction to `pkg/healthcheck/healthcheck.go`.

Test cases can be found [here](https://gist.github.com/ihcsim/1747c05f43e455a4e8f807b6d1b25434).

Open questions:

* The usage of the `linkerd.io/control-plane-ns` label won't work if the existing control plane is of an older version, since the older global resources aren't labelled. Do we need to support this use case?

Fixes https://github.com/linkerd/linkerd2/issues/2954.

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -155,9 +155,8 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		checks = append(checks, healthcheck.LinkerdConfigChecks)
 
 		if stage != configStage {
-			checks = append(checks,
-				healthcheck.LinkerdControlPlaneExistenceChecks,
-				healthcheck.LinkerdAPIChecks)
+			checks = append(checks, healthcheck.LinkerdControlPlaneExistenceChecks)
+			checks = append(checks, healthcheck.LinkerdAPIChecks)
 
 			if options.dataPlaneOnly {
 				checks = append(checks, healthcheck.LinkerdDataPlaneChecks)
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -227,15 +227,8 @@ func runChecksTable(wout io.Writer, hc *healthcheck.HealthChecker) bool {
 		}
 
 		fmt.Fprintf(wout, "%s %s\n", status, result.Description)
-		if err := result.Err; err != nil {
-			errMsg := fmt.Sprintf("%s", err)
-
-			// indent multi-line error message
-			if n := strings.Count(errMsg, "\n"); n > 0 {
-				errMsg = strings.Replace(errMsg, "\n", "\n    ", n-1)
-			}
-			fmt.Fprintf(wout, "    %s\n", strings.TrimSpace(errMsg))
-
+		if result.Err != nil {
+			fmt.Fprintf(wout, "    %s\n", result.Err)
 			if result.HintAnchor != "" {
 				fmt.Fprintf(wout, "    see %s%s for hints\n", healthcheck.HintBaseURL, result.HintAnchor)
 			}
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -227,8 +227,15 @@ func runChecksTable(wout io.Writer, hc *healthcheck.HealthChecker) bool {
 		}
 
 		fmt.Fprintf(wout, "%s %s\n", status, result.Description)
-		if result.Err != nil {
-			fmt.Fprintf(wout, "    %s\n", result.Err)
+		if err := result.Err; err != nil {
+			errMsg := fmt.Sprintf("%s", err)
+
+			// indent multi-line error message
+			if n := strings.Count(errMsg, "\n"); n > 0 {
+				errMsg = strings.Replace(errMsg, "\n", "\n    ", n-1)
+			}
+			fmt.Fprintf(wout, "    %s\n", strings.TrimSpace(errMsg))
+
 			if result.HintAnchor != "" {
 				fmt.Fprintf(wout, "    see %s%s for hints\n", healthcheck.HintBaseURL, result.HintAnchor)
 			}
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -145,7 +145,9 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 	}
 
 	if options.preInstallOnly {
-		checks = append(checks, healthcheck.LinkerdPreInstallChecks)
+		checks = append(checks,
+			healthcheck.LinkerdPreInstallChecks,
+			healthcheck.LinkerdPreInstallConfigChecks)
 		if !options.cniEnabled {
 			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
 		}
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -147,7 +147,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 	if options.preInstallOnly {
 		checks = append(checks,
 			healthcheck.LinkerdPreInstallChecks,
-			healthcheck.LinkerdPreInstallConfigChecks)
+			healthcheck.LinkerdPreInstallGlobalResourcesChecks)
 		if !options.cniEnabled {
 			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
 		}
@@ -155,8 +155,9 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		checks = append(checks, healthcheck.LinkerdConfigChecks)
 
 		if stage != configStage {
-			checks = append(checks, healthcheck.LinkerdControlPlaneExistenceChecks)
-			checks = append(checks, healthcheck.LinkerdAPIChecks)
+			checks = append(checks,
+				healthcheck.LinkerdControlPlaneExistenceChecks,
+				healthcheck.LinkerdAPIChecks)
 
 			if options.dataPlaneOnly {
 				checks = append(checks, healthcheck.LinkerdDataPlaneChecks)
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -23,8 +23,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/helm/pkg/chartutil"
 	"k8s.io/helm/pkg/proto/hapi/chart"
@@ -251,7 +249,7 @@ resources for the Linkerd control plane. This command should be followed by
   # Install Linkerd into a non-default namespace.
   linkerd install config -l linkerdtest | kubectl apply -f -`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return installRunE(options, configStage, parentFlags)
+			return installRunE(options, configStage, parentFlags, globalResourcesExist)
 		},
 	}
 
@@ -282,11 +280,18 @@ control plane. It should be run after "linkerd install config".`,
   # Install Linkerd into a non-default namespace.
   linkerd install control-plane -l linkerdtest | kubectl apply -f -`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if !options.skipChecks {
-				exitIfNamespaceDoesNotExist()
+			// check if global resources exist to determine if the `install config`
+			// stage succeeded
+			exists := globalResourcesExist()
+			if !exists && !options.skipChecks {
+				fmt.Fprintf(os.Stderr,
+					"Failed to find required control-plane namespace: %s. Run \"linkerd install config -l %s | kubectl apply -f -\" to create it (this requires cluster administration permissions).\nSee https://linkerd.io/2/getting-started/ for more information. Or use \"--skip-checks\" to proceed anyway.\n",
+					controlPlaneNamespace, controlPlaneNamespace,
+				)
+				os.Exit(1)
 			}
 
-			return installRunE(options, controlPlaneStage, flags)
+			return installRunE(options, controlPlaneStage, flags, linkerdConfigConfigMapExists)
 		},
 	}
 
@@ -330,7 +335,7 @@ control plane.`,
   # Installation may also be broken up into two stages by user privilege, via
   # subcommands.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return installRunE(options, "", flags)
+			return installRunE(options, "", flags, globalResourcesExist)
 		},
 	}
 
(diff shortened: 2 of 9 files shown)


cli/cmd/install.go
```diff
@@ -834,61 +838,32 @@ func (options *installOptions) proxyConfig() *pb.Proxy {
...
-	}
+	hc := healthcheck.NewHealthChecker(checks, &healthcheck.Options{
+		ControlPlaneNamespace: controlPlaneNamespace,
+		KubeConfig:            kubeconfigPath,
+	})
 
-	fmt.Fprintln(os.Stderr, "Linkerd has already been installed on your cluster in the linkerd namespace. Please run upgrade if you'd like to update this installation. Otherwise, use the --ignore-cluster flag.")
-	os.Exit(1)
+	notExist := hc.RunChecks(func(result *healthcheck.CheckResult) {})
+	return !notExist
```
scottcarol: I got a bit confused here with the `notExist` variable. 

When we run `hc.RunChecks()` here does a `false` return mean that global resources already exist? And that is why you are returning `!notExist`? I have the same question about `linkerdConfigConfigMapExists()`.

A longer but more explicit way could be:

```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 701d63ee..c078f2a7 100644 func globalResourcesExist() bool {
+       globalResourcesAlreadyExist := false // default value
+
        checks := []healthcheck.CategoryID{
                healthcheck.LinkerdPreInstallGlobalResourcesChecks,
        }
@@ -848,8 +850,13 @@ func globalResourcesExist() bool {
                KubeConfig:            kubeconfigPath,
        })

-       notExist := hc.RunChecks(func(result *healthcheck.CheckResult) {})
-       return !notExist
+       checksPass := hc.RunChecks(func(result *healthcheck.CheckResult) {})
+
+       if !checksPass {
+               globalResourcesAlreadyExist = true
+       }
+
+       return globalResourcesAlreadyExist
 } 
```

Or could you simplify by returning the result of `hc.RunChecks()` directly and changing the boolean wherever you are expecting that result?

pkg/healthcheck/healthcheck.go
```diff
@@ -382,38 +452,109 @@ func (hc *HealthChecker) allCategories() []category {
...
+					},
+				},
+				{
+					description: "'linkerd-config' config map does not exist",
+					hintAnchor:  "l5d-linkerd-config-not-exists",
+					fatal:       true,
+					check: func(context.Context) error {
+						// to successfully install the control plane, the `linkerd-config`
+						// config map must not exist.
+						if err := hc.checkLinkerdConfigConfigMap(); !kerrors.IsNotFound(err) {
```
siggy: since `checkLinkerdConfigConfigMap` is namespace-specific, can we remove the `l5d-control-plane-ns-exists` check above?

taking this a step further, since `LinkerdConfigNotExistsChecks` is not used by the `linkerd check` command (and thus does not require all the category/hintanchor/check()/etc machinery that health checkers provide, it may be more straightforward to simply implement the `linkerd-config` check in terms of the existing `fetchConfigs` function:

```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 701d63ee..49bd09e3 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -23,6 +23,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/helm/pkg/chartutil"
 	"k8s.io/helm/pkg/proto/hapi/chart"
@@ -853,17 +854,14 @@ func globalResourcesExist() bool {
 }
 
 func linkerdConfigConfigMapExists() bool {
-	checks := []healthcheck.CategoryID{
-		healthcheck.LinkerdConfigNotExistsChecks,
+	kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, 0)
+	if err != nil {
+		return false
 	}
 
-	hc := healthcheck.NewHealthChecker(checks, &healthcheck.Options{
-		ControlPlaneNamespace: controlPlaneNamespace,
-		KubeConfig:            kubeconfigPath,
-	})
+	_, err = fetchConfigs(kubeAPI)
 
-	notExists := hc.RunChecks(func(result *healthcheck.CheckResult) {})
-	return !notExists
+	return !kerrors.IsNotFound(err)
 }
 
 func (idopts *installIdentityOptions) validate() error {
diff --git a/pkg/healthcheck/healthcheck.go b/pkg/healthcheck/healthcheck.go
index bf8222d3..ec27354f 100644
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -8,8 +8,6 @@ import (
 	"strings"
 	"time"
 
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
-
 	"github.com/linkerd/linkerd2/controller/api/public"
 	spclient "github.com/linkerd/linkerd2/controller/gen/client/clientset/versioned"
 	healthcheckPb "github.com/linkerd/linkerd2/controller/gen/common/healthcheck"
@@ -64,12 +62,6 @@ const (
 	// to determine if a control plane is already installed.
 	LinkerdPreInstallGlobalResourcesChecks CategoryID = "pre-linkerd-global-resources"
 
-	// LinkerdConfigNotExistsChecks adds a series of checks to determine
-	// if the `linkerd-config` config map already exist. In order to install a
-	// control plane in a namespace, this config map must not exist in that
-	// namespace.
-	LinkerdConfigNotExistsChecks CategoryID = "linkerd-config-config-map-not-exists"
-
 	// LinkerdConfigChecks enabled by `linkerd check config`
 
 	// LinkerdConfigChecks adds a series of checks to validate that the Linkerd
@@ -505,41 +497,6 @@ func (hc *HealthChecker) allCategories() []category {
 				},
 			},
 		},
-		{
-			id: LinkerdConfigNotExistsChecks,
-			checkers: []checker{
-				{
-					description: "control plane namespace exists",
-					hintAnchor:  "l5d-control-plane-ns-exists",
-					fatal:       true,
-					check: func(context.Context) error {
-						if hc.kubeAPI == nil {
-							var err error
-							hc.kubeAPI, err = k8s.NewAPI(hc.KubeConfig, hc.KubeContext, requestTimeout)
-							if err != nil {
-								return err
-							}
-						}
-
-						return hc.CheckNamespace(hc.ControlPlaneNamespace, true)
-					},
-				},
-				{
-					description: "'linkerd-config' config map does not exist",
-					hintAnchor:  "l5d-linkerd-config-not-exists",
-					fatal:       true,
-					check: func(context.Context) error {
-						// to successfully install the control plane, the `linkerd-config`
-						// config map must not exist.
-						if err := hc.checkLinkerdConfigConfigMap(); !kerrors.IsNotFound(err) {
-							return fmt.Errorf("found existing 'linkerd-config' config map")
-						}
-
-						return nil
-					},
-				},
-			},
-		},
 		{
 			id: LinkerdControlPlaneExistenceChecks,
 			checkers: []checker{
```

cli/cmd/install.go
```diff
@@ -346,11 +350,11 @@ control plane.`,
 	return cmd
 }
 
-func installRunE(options *installOptions, stage string, flags *pflag.FlagSet) error {
-	if !options.ignoreCluster {
-		// TODO: consider cobra.SilenceUsage, so we can return errors from
-		// `RunE`, rather than calling `os.Exit(1)`
-		exitIfClusterExists()
+func installRunE(options *installOptions, stage string, flags *pflag.FlagSet, preInstallCheck func() bool) error {
+	controlPlaneExists := preInstallCheck()
```
siggy: `controlPlaneExists` isn't quite accurate, as `preInstallCheck()` has different meanings depending on the caller.

consider moving the call to `preInstallCheck()` (and related `!options.ignoreCluster` logic) out of `installRunE`. given that we only have 3 cases (`install`, `install config`, `install control-plane`), and each is sufficiently different, i think it's helpful to put this logic explicitly into the respective `newCmdInstall*()` functions.

grampelberg:
@siggy I think it is time for an `uninstall` command that is consumed by `kubectl delete`. We can add that to the warn text and to the website.

pkg/healthcheck/healthcheck.go
```diff
@@ -841,8 +991,11 @@ func expectedServiceAccountNames() []string {
 	}
 }
 
-func (hc *HealthChecker) checkClusterRoles() error {
-	crList, err := hc.kubeAPI.RbacV1().ClusterRoles().List(metav1.ListOptions{})
+func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
+	options := metav1.ListOptions{
+		LabelSelector: k8s.ControllerNSLabel,
+	}
```
alpeb: It might be worth moving this into a global var given it's used multiple times.

pkg/healthcheck/healthcheck.go
ihcsim: The reason why `l5d-control-plane-ns-exists` is needed is to guard against cases like:
```
linkerd install config | kubectl apply -f -
linkerd install control-plane -l linkerd-dev | kubectl apply -f -
```
Without the namespace check, we won't be able to distinguish between the `linkerd-config` config map isn't created yet, and the namespace just doesn't exist.

I think moving this logic the `fetchConfigs` makes sense to me.

pkg/healthcheck/healthcheck.go
ihcsim: At this point, I prefer to keep this close to where it's used. As this is a big file, I'm reluctant to make this variable (which is used only by a handful of methods) a global, and am willing to live with a bit of redundancy.

ihcsim:
@alpeb Appreciate the feedback!

> While testing install, it was indeed clear the necessity to see the full list of offending resources.

:+1: I will surface those information.

> #2954 mentions an additional check for the webhook config's service namespace; will that be tackled in a separate PR?

Probably not going to do it, as it requires the user running `linkerd check` to have cluster-scoped permissions, which will defeat the whole purpose of multi-stage install.

cli/cmd/install.go
```diff
@@ -162,6 +162,10 @@ const (
 	defaultIdentityTrustDomain        = "cluster.local"
 	defaultIdentityIssuanceLifetime   = 24 * time.Hour
 	defaultIdentityClockSkewAllowance = 20 * time.Second
+
+	errMsgGlobalResourcesExist           = "Can't install the Linkerd control plane in the '%s' namespace. Global resources from an existing Linkerd installation detected:\n%s\nRemove these resources, or use the --ignore-cluster flag to overwrite them."
+	errMsgLinkerdConfigConfigMapNotFound = "Can't install the Linkerd control plane in the '%s' namespace. Reason: %s.\nIf this is expected, use the --ignore-cluster flag to continue the installation."
+	errMsgGlobalResourcesMissing         = "Can't install the Linkerd control plane in the '%s' namespace. The required Linkerd global resources are missing.\nIf this is expected, use the --skip-checks flag to continue the installation."
```
siggy: these error messages need a trailing newline, either here, or where they are printed.

pkg/healthcheck/healthcheck.go
```diff
@@ -898,10 +1009,79 @@ func (hc *HealthChecker) checkCustomResourceDefinitions() error {
...
+func checkResources(resourceName string, objects []runtime.Object, expectedNames []string, shouldExist bool) error {
+	if !shouldExist {
+		if len(objects) > 0 {
+			resources := ""
+			for _, obj := range objects {
+				m, err := meta.Accessor(obj)
+				if err != nil {
+					return err
+				}
+				resources += fmt.Sprintf("- %s/%s\n", resourceName, m.GetName())
```
siggy: i'd avoid format-specific string rendering here, as this code is used by at least `linkerd check -o json`:

```
        {
          "description": "no ClusterRoleBindings exist",
          "hint": "https://linkerd.io/checks/#pre-l5d-existence",
          "error": "- ClusterRoleBindings/linkerd-linkerd-controller\n- ClusterRoleBindings/linkerd-linkerd-identity\n- ClusterRoleBindings/linkerd-linkerd-prometheus\n- ClusterRoleBindings/linkerd-linkerd-proxy-injector\n- ClusterRoleBindings/linkerd-linkerd-sp-validator\n- ClusterRoleBindings/linkerd-linkerd-tap\n",
          "result": "error"
        },
```

the newline-delimited list of resources looks really good when rendered as text, but i think it's more clear to just return an error string list of resources, which will render appropriately in all contexts. something like:
```
ClusterRoleBindings found but should not exist: linkerd-linkerd-controller linkerd-linkerd-identity linkerd-linkerd-prometheus linkerd-linkerd-proxy-injector linkerd-linkerd-sp-validator linkerd-linkerd-tap
```

if we wanted the best of both worlds, we probably should modify these APIs to return a structured object that enumerates the missing resource strings, but that's probably best done in a subsequent PR.

cli/cmd/install.go
```diff
@@ -834,61 +850,51 @@ func (options *installOptions) proxyConfig() *pb.Proxy {
...
-			return
+	hc := healthcheck.NewHealthChecker(checks, &healthcheck.Options{
+		ControlPlaneNamespace: controlPlaneNamespace,
+		KubeConfig:            kubeconfigPath,
+	})
+
+	errMsgs := ""
+	hc.RunChecks(func(result *healthcheck.CheckResult) {
+		if result.Err != nil {
+			errMsgs += fmt.Sprintf("%s", result.Err)
```
siggy: i think this will just concatenate error strings together?

consider:
```go
errMsgs := []string{}
hc.RunChecks(func(result *healthcheck.CheckResult) {
	if result.Err != nil {
		errMsgs = append(errMsgs, result.Err.Error())
	}
})

if len(errMsgs) > 0 {
	return errors.New(strings.Join(errMsgs, ", "))
}
```

cli/cmd/check.go
```diff
@@ -225,8 +227,15 @@ func runChecksTable(wout io.Writer, hc *healthcheck.HealthChecker) bool {
 		}
 
 		fmt.Fprintf(wout, "%s %s\n", status, result.Description)
-		if result.Err != nil {
-			fmt.Fprintf(wout, "    %s\n", result.Err)
+		if err := result.Err; err != nil {
+			errMsg := fmt.Sprintf("%s", err)
+
+			// indent multi-line error message
```
siggy: per other comments around multi-line error messages, i think this can be backed out.

pkg/healthcheck/healthcheck.go
```diff
@@ -801,6 +888,18 @@ func (hc *HealthChecker) PublicAPIClient() public.APIClient {
 	return hc.apiClient
 }
 
+func (hc *HealthChecker) checkLinkerdConfigConfigMap(shouldExist bool) error {
+	cm, err := hc.kubeAPI.CoreV1().ConfigMaps(hc.ControlPlaneNamespace).Get(k8s.ConfigConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		if !shouldExist && kerrors.IsNotFound(err) {
```
siggy: is this `shouldExist` check redundant now that `checkResources` handles it?

ihcsim: I added it here to account for the `not found` error returned by the k8s API. Yes, you are right that we don't need to perform the `shouldExist` check here. I'll change it to:

```go
if err != nil && !kerrors.IsNotFound(err) {
  return err
}
```
And let `checkResources` do the `shouldExist` check.

PR #2988, opened 2019-06-24
linkerd inject from remote URL

### LINKED ISSUE #2983: inject: support remote URLs in file argument
It's common to run something like:

```
curl https://gist.githubusercontent.com/.../example.yml |linkerd inject - |kubectl apply -f -
```

The inject command could behave more like kubectl and accept remote URLs in place of local files like:
```
linkerd inject https://gist.githubusercontent.com/.../example.yml |kubectl apply -f -
```

Pothulapati:
Fixes #2983 

```
linkerd inject http://raw.githubusercontent.com/BuoyantIO/emojivoto/master/emojivoto-daemonset.yml --manual | kubectl apply -f -
```
Should work with this.

@olix0r @ihcsim 

### DIFF
--- a/cli/cmd/inject_util.go
+++ b/cli/cmd/inject_util.go
@@ -3,8 +3,11 @@ package cmd
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


cli/cmd/inject_util.go
```diff
@@ -139,6 +142,23 @@ func read(path string) ([]io.Reader, error) {
...
 	if path == "-" {
 		in = append(in, os.Stdin)
+	} else if isValidURL(path) {
+		var resp *http.Response
+		resp, err = http.Get(path)
+		if err != nil {
+			return nil, err
+		}
+
+		if resp.StatusCode == 404 {
```
ihcsim: I think we should return an error if the status isn't a 200. See the similar code in kubectl [here](https://github.com/kubernetes/cli-runtime/blob/92e6ccd8a070991c43024c9ef9989b68b47452c1/pkg/resource/visitor.go#L280-L282).

Pothulapati:
@ihcsim Updated :)

Issue #2953, opened 2019-06-17
Redesign `Overview` page in Dashboard


scottcarol:
The `Overview` page in the Dashboard could be improved and made more useful.

Currently it is an accordion-style list of namespaces where each namespace, when expanded, shows tables for Deployments, Pods and TCP if available.

Notes to consider:
* Education. There are some features and/or data we offer that might not be apparent to users (identity, edges, traffic split, upcoming Services view), are there ways to use the Overview space to teach users how and why to see different tables?
* #2838 In multi-tenancy clusters with Linkerd there could be 50+ namespaces, making the accordion list impractical

scottcarol:
cc @grampelberg

scottcarol:
Another note, the sidebar should be considered part of this project. We could allow the user to choose a specific namespace, so that instead of scrolling through every deployment in every namespace (below), it would only be deployments in the selected namespace. 

![Screen Shot 2019-06-18 at 5 11 39 PM](https://user-images.githubusercontent.com/2289389/59727828-38a29800-91ec-11e9-9dce-246302dbed6f.png)

scottcarol:
Replaced by #3400

PR #2995, opened 2019-06-25
Improve UI for Edges table in dashboard

scottcarol:
This PR improves the UI for the Edges table in the dashboard as discussed with @wmorgan and @grampelberg 

Previous table:
Screen Shot 2019-06-25 at 3 07 22 PM

New table:
Screen Shot 2019-06-25 at 3 03 14 PM

New table if `noIdentityMsg` is present - explanatory message will show on hover:
Screen Shot 2019-06-25 at 3 05 20 PM

If no edges are found, an empty table will display. We discussed displaying a special message such as "No Edges Found" but as shown in below screenshot we aren't doing that for any other table. @grampelberg do you want to treat the "Edges" table differently than all other tables?

Screen Shot 2019-06-25 at 2 22 18 PM


scottcarol:
Context on what's returned from the current `edges` API endpoint (#2793):

"...We are currently filtering out edges that are one-sided (i.e. health probes, Prometheus, tap) and also edges where we cannot definitively match the inbound/outbound queries together. That will be the next step."

grampelberg:
- I still like the tooltip, but we can tackle putting tooltips on everything in a separate issue/PR.
- I'd like *all* the tables to have a decent empty state, but that can be a separate issue/PR as well.
- As we've got more room now, WDYT about adding a `namespace` column?
- Looking at the TCP table above, it is using `Pod` as the name, WDYT about that instead of `Name`? Could conditionally change the name based on what the detail page is.

scottcarol:
@grampelberg sounds good, I created issues #2997 and #2998 for tooltips and empty table message.

`Namespace` column sounds good.

Re: using `Pod` instead of `Name` for the DST/SRC name, there could be an instance of it *not* being a Pod/other k8s resource type, correct? That's not wired up yet in the `edges` command but that could be theoretically possible.

grampelberg:
That’s a good argument, sold me on keeping it `Name`.

scottcarol:
@grampelberg After investigating, when `/api/public.edges.go` queries Prometheus, it is only for edges within a particular namespace. When I run `linkerd edges po -n emojivoto` it generates 2 prometheus queries and then matches up the results by `pod` and `dst_pod`:

```bash
count(response_total{direction="inbound", namespace="emojivoto", pod!=""}) by (pod, client_id)
count(response_total{direction="outbound", namespace="emojivoto", pod!=""}) by (pod, dst_pod, server_id, no_tls_reason)
```

So on the dashboard, the `Namespace` column would only ever be the namespace of the resource detail page you're on. I can still add it for consistency, and when we expand the `edges` query that could change. What do you think?

grampelberg:
Is there a reason it is namespace scoped? Seems like a primary use case to me.

scottcarol:
After chatting with @grampelberg my plan is to get this UI PR approved and merged, and then modify the edges API endpoint so that it is not restricted by namespace and returns namespace info (with associated updates to the CLI and dashboard).

web/app/js/components/EdgesTable.jsx
```diff
@@ -1,113 +1,118 @@
...
 
-const tooltipText = `Edges show the source, destination name and identity
-  for proxied connections. If no identity is known, a message is displayed.`;
+const generateEdgesTableTitle = edges => {
+  let title = "Edges";
+  let identity = "";
+  if (edges.length > 0) {
+    edges[0].direction === "INBOUND" ? identity = edges[0].serverId : identity = edges[0].clientId;
+    identity = identity.split('.')[0] + '.' + identity.split('.')[1];
+    title = title + " (Identity: " + identity + ")";
```
rmars: tioli: might be easier to read as
```
title = `${title} (Identity: ${identity})`;
```

web/app/js/components/EdgesTable.jsx
```diff
@@ -1,113 +1,118 @@
...
   ];
 };
 
-const tooltipText = `Edges show the source, destination name and identity
-  for proxied connections. If no identity is known, a message is displayed.`;
+const generateEdgesTableTitle = edges => {
+  let title = "Edges";
+  let identity = "";
+  if (edges.length > 0) {
+    edges[0].direction === "INBOUND" ? identity = edges[0].serverId : identity = edges[0].clientId;
```
rmars: we don't use identity outside the if, so this would be simpler to not declare it on line 75, and declare it here as 
```
let identity = edges[0].direction === "INBOUND" ? edges[0].serverId : edges[0].clientId;
```

web/app/js/components/EdgesTable.jsx
```diff
@@ -1,113 +1,118 @@
...
-      title: "Client",
-      dataIndex: "client",
+      title: "Identity",
+      dataIndex: "identity",
       isNumeric: false,
-      filter: d => d.clientId,
-      render: d => d.clientId.split('.')[0] + '.' + d.clientId.split('.')[1],
-      sorter: d => d.clientId
+      filter: d => d.identity,
+      render: d => d.identity ? d.identity.split('.')[0] + '.' + d.identity.split('.')[1] : null,
```
rmars: TIOLI:
```
`${d.identity.split('.')[0]}.${d.identity.split('.')[1]}`
```

scottcarol: Much cleaner, great suggestion 🙌

scottcarol:
Hey @alpeb thanks for the UI suggestions! 
Re: icon size, I chose to have a larger icon to highlight the "Secured" status, but we could also shrink it down to be more in line with Grafana/other icons. @grampelberg do you have an opinion?

Current:

Screen Shot 2019-06-26 at 4 07 33 PM


Smaller:

Screen Shot 2019-06-26 at 4 01 14 PM

Re: identity label, I'm hesitant to introduce new font sizes and styles into the table without a larger design plan for it, but I agree it could be differentiated better. This could be something we could look at for #2953.

alpeb:
@scottcarol I actually had in mind having a minimal column width, regardless of the icon size. The icon does feel pretty lonely using up all that space :slightly_smiling_face:

scottcarol:
@alpeb Ah thanks for the clarification, I interpreted your comment as being about the icon's size! Yes, I see your point about the column width 👍

PR #3013, opened 2019-06-28
edge-19.6.4

adleong:
## edge-19.6.4

This release adds support for the SMI [Traffic Split](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md) API.  Creating a TrafficSplit resource will cause Linkerd to
split traffic between the specified backend services.  Please see [the spec](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
for more details.

* CLI
  * Added a check to install to prevent installing multiple control planes into
    different namespaces
  * Added support for passing a URL directly to linkerd inject
  * Added a `--all-namespaces` flag to `linkerd edges`
* Controller
  * Added support for the SMI TrafficSplit API which allows users to define
    traffic splits in TrafficSplit custom resources
* Proxy
* Web UI
  * Improved UI for Edges table in dashboard by changing column names, adding a
    "Secured" icon and showing an empty Edges table in the case of no returned
    edges

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,8 @@
 ## edge-19.6.4
 
 This release adds support for the SMI [Traffic Split](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
-API.  Creating a TrafficSplit resource will cause Linkerd to split traffic
-between the specified backend services.  Please see [the spec](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
+API. Creating a TrafficSplit resource will cause Linkerd to split traffic
+between the specified backend services. Please see [the spec](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
 for more details.
 
 * CLI
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@ for more details.
     into different namespaces
   * Added support for passing a URL directly to `linkerd inject` (thanks
     @Pothulapati!)
-  * Added a `--all-namespaces` flag to `linkerd edges`
+  * Added the `--all-namespaces` flag to `linkerd edges`
 * Controller
   * Added support for the SMI TrafficSplit API which allows users to define
     traffic splits in TrafficSplit custom resources
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,14 +6,14 @@ between the specified backend services.  Please see [the spec](https://github.co
 for more details.
 
 * CLI
-  * Added a check to install to prevent installing multiple control planes into
-    different namespaces
-  * Added support for passing a URL directly to linkerd inject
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,24 @@
+## edge-19.6.4
+
+This release adds support for the SMI [Traffic Split](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
+API.  Creating a TrafficSplit resource will cause Linkerd to split traffic
+between the specified backend services.  Please see [the spec](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
+for more details.
+
+* CLI
+  * Added a check to install to prevent installing multiple control planes into
```
scottcarol: I would backtick `install`

CHANGES.md
```diff
@@ -1,3 +1,24 @@
...
+    different namespaces
+  * Added support for passing a URL directly to linkerd inject
```
scottcarol: `linkerd inject`

CHANGES.md
```diff
@@ -1,3 +1,24 @@
...
+  * Added a `--all-namespaces` flag to `linkerd edges`
```
scottcarol: The tiniest nit but... I think it should be `an` and not `a`

CHANGES.md
```diff
@@ -1,3 +1,24 @@
...
+* Controller
+  * Added support for the SMI TrafficSplit API which allows users to define
+    traffic splits in TrafficSplit custom resources
+* Proxy
```
scottcarol: this is blank?

admc: Do we normally areas like "Proxy" when there are no updates? Should we add something about traffic split support in the proxy?

CHANGES.md
ihcsim: Shout out to Tarun (at Pothulapati).

CHANGES.md
adleong: good catch.  traffic split spans controller and proxy but I decided to put it in the controller section.  I don't think we need it in both.  I'll remove the empty proxy bullet.

CHANGES.md
kleimkuhler: Still tiny, but it seems like we have usually referred to new flags with `the`!

CHANGES.md
```diff
@@ -1,3 +1,24 @@
+## edge-19.6.4
+
+This release adds support for the SMI [Traffic Split](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
+API.  Creating a TrafficSplit resource will cause Linkerd to split traffic
+between the specified backend services.  Please see [the spec](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
```
kleimkuhler: Looks like extra space before "Please ..."

PR #3025, opened 2019-07-02
Allow `edges` to handle requests to one resource from multiple namespaces

scottcarol:
This PR fixes a bug in the `edges` command where if src_resources from two different namespaces sent requests to the same dst_resource, the original src_identity was overwritten.

Previously, the `inbound` and `outbound` requests were paired by destination resource. For example, an `inbound` request where `deployment="foo"` was matched to an `outbound` request with `dst_deployment="foo"`. The match key was `foo`.

Now, requests are matched by both `dst_resource` and `src_namespace` so requests from different namespaces are not overwritten. 

///////////

To test, install emojivoto with Linkerd, and then install the `vote-bot` deployment of emojivoto into the `default` namespace (with Linkerd). On master, `linkerd edges deploy --all-namespaces` will result in the below.  Both `vote-bot` connections have the same client ID (incorrect).
```bash
SRC                  DST                  CLIENT                       SERVER                       MSG
linkerd-controller   linkerd-tap          linkerd-controller.linkerd   linkerd-tap.linkerd          -
linkerd-web          linkerd-controller   web.emojivoto                linkerd-controller.linkerd   -
vote-bot             web                  default.default              web.emojivoto                -
vote-bot             web                  default.default              web.emojivoto                -
web                  emoji                web.emojivoto                emoji.emojivoto              -
web                  voting               web.emojivoto                voting.emojivoto             -
```

testing with this branch should result in: 
```bash
SRC                  DST                  CLIENT                       SERVER                       MSG
linkerd-controller   linkerd-tap          linkerd-controller.linkerd   linkerd-tap.linkerd          -
linkerd-web          linkerd-controller   linkerd-web.linkerd          linkerd-controller.linkerd   -
vote-bot             web                  default.emojivoto            web.emojivoto                -
vote-bot             web                  default.default              web.emojivoto                -
web                  emoji                web.emojivoto                emoji.emojivoto              -
web                  voting               web.emojivoto                voting.emojivoto             -
```

### DIFF
--- a/controller/api/public/edges.go
+++ b/controller/api/public/edges.go
@@ -103,7 +103,7 @@ func processEdgeMetrics(inbound, outbound model.Vector, resourceType, selectedNa
 			// format of clientId is id.namespace.serviceaccount.cluster.local
 			clientIDSlice := strings.Split(string(clientID), ".")
 			srcNs := clientIDSlice[1]
-			key := model.LabelValue(dstResource + srcNs)
+			key := model.LabelValue(fmt.Sprintf("%s.%s", dstResource, srcNs))
 			dstIndex[key] = sample.Metric
 		}
 	}
@@ -112,7 +112,7 @@ func processEdgeMetrics(inbound, outbound model.Vector, resourceType, selectedNa
 		dstResource := sample.Metric[model.LabelName(resourceReplacementOutbound)]
 		srcNs := sample.Metric[model.LabelName("namespace")]
 
-		key := dstResource + srcNs
+		key := model.LabelValue(fmt.Sprintf("%s.%s", dstResource, srcNs))
 		if _, ok := srcIndex[key]; !ok {
 			srcIndex[key] = []model.Metric{}
 		}
--- a/controller/api/public/edges.go
+++ b/controller/api/public/edges.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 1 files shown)


controller/api/public/edges.go
```diff
@@ -96,14 +97,22 @@ func processEdgeMetrics(inbound, outbound model.Vector, resourceType, selectedNa
...
 		// information
-		if _, ok := sample.Metric[model.LabelName("client_id")]; ok {
-			key := sample.Metric[model.LabelName(resourceReplacementInbound)]
+		if clientID, ok := sample.Metric[model.LabelName("client_id")]; ok {
+			dstResource := string(sample.Metric[model.LabelName(resourceReplacementInbound)])
+
+			// format of clientId is id.namespace.serviceaccount.cluster.local
+			clientIDSlice := strings.Split(string(clientID), ".")
+			srcNs := clientIDSlice[1]
+			key := model.LabelValue(dstResource + srcNs)
```
klingerf: Let's use a delimiter here to separate the name and the namespace, so there's no ambiguity as to which part of the string is the name, and which part is the namespace. I recommend using a ".", since that's no a valid character for names or namespaces. How about something like:

```suggestion
			key := model.LabelValue(fmt.Sprintf("%s.%s", dstResource, srcNs))
```

controller/api/public/edges.go
```diff
@@ -96,14 +97,22 @@ func processEdgeMetrics(inbound, outbound model.Vector, resourceType, selectedNa
...
 			dstIndex[key] = sample.Metric
 		}
 	}
 
 	for _, sample := range outbound {
-		key := sample.Metric[model.LabelName(resourceReplacementOutbound)]
+		dstResource := sample.Metric[model.LabelName(resourceReplacementOutbound)]
+		srcNs := sample.Metric[model.LabelName("namespace")]
+
+		key := dstResource + srcNs
```
klingerf: And then same comment here -- I'd use a "." separator. Also, the `key` var here is a string which is being coerced to a `model.LabelValue` when the map lookup happens, but for consistency with how the key is built above, I recommend explicitly typing the key a `model.LabelValue` as well.

```suggestion
		key := model.LabelValue(fmt.Sprintf("%s.%s", dstResource, srcNs))
```

controller/api/public/edges.go
scottcarol: 👍 Great idea, this will make it easier to read if debugging later.

PR #3031, opened 2019-07-02
release notes for edge-19.7.1

hawkw:
## edge-19.7.1

* CLI
  * Added more descriptive output to the `linkerd check` output for control
    plane ReplicaSet readiness
  * **Breaking change** Renamed `config.linkerd.io/debug` annotation to
    `config.linkerd.io/enable-debug-sidecar`, to match the
    `--enable-debug-sidecar` CLI flag that sets it
  * Fixed a bug in `linkerd edges` that caused incorrect identities to be
    displayed when requests were sent from two or more namespaces
* Controller
  * Added the `linkerd.io/control-plane-ns` label to the SMI Traffic Split CRD
* Proxy
  * Fixed proxied HTTP/2 connections returning 502 errors when the upstream
    connection is reset, rather than propagating the reset to the client
  * Changed the proxy to treat unexpected HTTP/2 frames as stream errors rather
    than connection errors

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,15 +1,15 @@
 ## edge-19.7.1
 
 * CLI
-  * Added more descriptive output to the `linkerd check` for control plane
-    ReplicaSet readiness
+  * Added more descriptive output to the `linkerd check` output for control
+    plane ReplicaSet readiness
   * **Breaking change** Renamed `config.linkerd.io/debug` annotation to
     `config.linkerd.io/enable-debug-sidecar`, to match the
     `--enable-debug-sidecar` CLI flag that sets it
   * Fixed a bug in `linkerd edges` that caused incorrect identities to be
     displayed when requests were sent from two or more namespaces
 * Controller
-  * Addws the `linkerd.io/control-plane-ns` label to the SMI Traffic Split CRD
+  * Added the `linkerd.io/control-plane-ns` label to the SMI Traffic Split CRD
 * Proxy
   * Fixed proxied HTTP/2 connections returning 502 errors when the upstream
     connection is reset, rather than propagating the reset to the client
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,20 @@
+## edge-19.7.1
+
+* CLI
+  * Added more descriptive output to the `linkerd check` for control plane
+    ReplicaSet readiness
+  * Renamed `config.linkerd.io/debug` annotation to
+    `config.linkerd.io/enable-debug-sidecar`, to match the CLI flag that sets it
+  * Fixed a bug in `linkerd edges` that caused incorrect identities to be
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,20 @@
+## edge-19.7.1
+
+* CLI
+  * Added more descriptive output to the `linkerd check` for control plane
+    ReplicaSet readiness
+  * Renamed `config.linkerd.io/debug` annotation to
+    `config.linkerd.io/enable-debug-sidecar`, to match the CLI flag that sets it
```
klingerf: Is this a breaking change? In the past we've doing something like:

```
  * **Breaking Change:** Renamed the `config.linkerd.io/debug` annotation to...
```

CHANGES.md
```diff
@@ -1,3 +1,20 @@
...
+  * Fixed a bug in `linkerd edges` that caused incorrect identities to be
+    displayed when requests were sent from two or more namespaces
+* Controller
+  * Add the `linkerd.io/control-plane-ns` label to the SMI Traffic Split CRD
```
klingerf: `Add` => `Added`

CHANGES.md
kleimkuhler: It may be helpful to say "to match the `--enable-debug-sidecar` flag that sets it" incase a user reading this is not aware of that flag already.

hawkw: Ah, yeah, good call -- @ihcsim, is this breaking?

CHANGES.md
```diff
@@ -1,3 +1,21 @@
+## edge-19.7.1
+
+* CLI
+  * Added more descriptive output to the `linkerd check` for control plane
```
wmorgan: the `linkerd check` **command** ?

CHANGES.md
ihcsim: Yes, it is. Thanks for calling it out.

CHANGES.md
```diff
@@ -1,3 +1,21 @@
...
+    ReplicaSet readiness
+  * **Breaking change** Renamed `config.linkerd.io/debug` annotation to
+    `config.linkerd.io/enable-debug-sidecar`, to match the
+    `--enable-debug-sidecar` CLI flag that sets it
+  * Fixed a bug in `linkerd edges` that caused incorrect identities to be
+    displayed when requests were sent from two or more namespaces
+* Controller
+  * Addws the `linkerd.io/control-plane-ns` label to the SMI Traffic Split CRD
```
wmorgan: Typo

CHANGES.md
hawkw: cool, updated!

CHANGES.md
hawkw: I think this should be "the linkerd check _output_"?

PR #3032, opened 2019-07-03
Add release notes for stable-2.4.0

kleimkuhler:
## stable-2.4.0

This release adds traffic splitting functionality, support for the Kubernetes
Service Mesh Interface (SMI), graduates high-availability support out of
experimental status, and adds a tremendous list of other improvements,
performance enhancements, and bug fixes.

Linkerd's new traffic splitting feature allows users to dynamically control the
percentage of traffic destined for a service. This powerful feature can be used
to implement rollout strategies like canary releases and blue-green deploys.
Support for the [Service Mesh Interface](https://smi-spec.io) (SMI) makes it
easier for ecosystem tools to work across all service mesh implementations.

Along with the introduction of optional install stages via the `linkerd install
config` and `linkerd install control-plane` commands, the default behavior of
the `linkerd inject` command only adds annotations and defers injection to the
always-installed proxy injector component.

Finally, there have been many performance and usability improvements to the
proxy and UI, as well as production-ready features including: 
* A new `linkerd edges` command that provides fine-grained observability into
  the TLS-based identity system
* A `--enable-debug-sidecar` flag for the `linkerd inject` command that improves
  debugging efforts

Linkerd recently passed a CNCF-sponsored security audit! Check out the in-depth
report [here](https://github.com/linkerd/linkerd2/blob/master/SECURITY_AUDIT.pdf).

To install this release, run: `curl https://run.linkerd.io/install | sh`

**Upgrade notes**: Use the `linkerd upgrade` command to upgrade the control
plane. This command ensures that all existing control plane's configuration and
mTLS secrets are retained. For more details, please see the [upgrade
instructions](https://linkerd.io/2/tasks/upgrade/#upgrade-notice-stable-2-4-0) for more details.

**Special thanks to**: @alenkacz, @codeman9, @dwj300, @jackprice, @liquidslr
@matej-g, @Pothulapati, @zaharidichev, 

**Full release notes**:

* CLI
  * **Breaking Change** Removed the `--proxy-auto-inject` flag, as the proxy
    injector is now always installed
  * **Breaking Change** Replaced the `--linkerd-version` flag with the
    `--proxy-version` flag in the `linkerd install` and `linkerd upgrade`
    commands, which allows setting the version for the injected proxy sidecar
    image, without changing the image versions for the control plane
  * Introduced install stages: `linkerd install config` and `linkerd install
    control-plane`
  * Introduced upgrade stages: `linkerd upgrade config` and `linkerd upgrade
    control-plane`
  * Introduced a new `--from-manifests` flag to `linkerd upgrade` allowing
    manually feeding a previously saved output of `linkerd install` into the
    command, instead of requiring a connection to the cluster to fetch the
    config
  * Introduced a new `--manual` flag to `linkerd inject` to output the proxy
    sidecar container spec
  * Introduced a new `--enable-debug-sidecar` flag to `linkerd inject`, that
    injects a debug sidecar to inspect traffic to and from the meshed pod
  * Added a new check for unschedulable pods and PSP issues (thanks,
    @liquidslr!)
  * Disabled the spinner in `linkerd check` when running without a TTY
  * Ensured the ServiceAccount for the proxy injector is created before its
    Deployment to avoid warnings when installing the proxy injector (thanks,
    @dwj300!)
  * Added a `linkerd check config` command for verifying that `linkerd install
    config` was successful
  * Improved the help documentation of `linkerd install` to clarify flag usage
  * Added support for private Kubernetes clusters by changing the CLI to connect
    to the control plane using a port-forward (thanks, @jackprice!)
  * Fixed `linkerd check` and `linkerd dashboard` failing when any control plane
    pod is not ready, even when multiple replicas exist (as in HA mode)
  * **New** Added a `linkerd edges` command that shows the source and
    destination name and identity for proxied connections, to assist in
    debugging
  * Tap can now be disabled for specific pods during injection by using the
    `--disable-tap` flag, or by using the `config.linkerd.io/disable-tap`
    annotation
  * Introduced pre-install healthcheck for clock skew (thanks, @matej-g!)
  * Added a JSON option to the `linkerd edges` command so that output is
    scripting friendly and can be parsed easily (thanks @alenkacz!)
  * Fixed an issue when Linkerd is installed with `--ha`, running `linkerd
    upgrade` without `--ha` will disable the high availability control plane
  * Fixed an issue with `linkerd upgrade` where running without `--ha` would
    unintentionally disable high availability features if they were previously
    enabled
  * Added a `--init-image-version` flag to `linkerd inject` to override the
    injected proxy-init container version
  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
    `NET_ADMIN` capability is omitted from the CNI-enabled control plane's PSP
  * Updated `linkerd check` to validate the caller can create
    `PodSecurityPolicy` resources
  * Added a check to `linkerd install` to prevent installing multiple control
    planes into different namespaces avoid conflicts between global resources
  * Added support for passing a URL directly to `linkerd inject` (thanks
    @Pothulapati!)
  * Added more descriptive output to the `linkerd check` output for control
    plane ReplicaSet readiness
  * Refactored the `linkerd endpoints` to use the same interface as used by the
    proxy for service discovery information
  * Fixed a bug where `linkerd inject` would fail when given a path to a file
    outside the current directory
  * Graduated high-availability support out of experimental status
  * Modified the error message for `linkerd install` to provide instructions for
    proceeding when an existing installation is found
* Controller
  * Added Go pprof HTTP endpoints to all control plane components' admin servers
    to better assist debugging efforts
  * Fixed bug in the proxy injector, where sporadically the pod workload owner
    wasn't properly determined, which would result in erroneous stats
  * Added support for a new `config.linkerd.io/disable-identity` annotation to
    opt out of identity for a specific pod
  * Fixed pod creation failure when a `ResourceQuota` exists by adding a default
    resource spec for the proxy-init init container
  * Fixed control plane components failing on startup when the Kubernetes API
    returns an `ErrGroupDiscoveryFailed`
  * Added Controller Component Labels to the webhook config resources (thanks,
    @Pothulapati!)
  * Moved the tap service into its own pod
  * **New** Control plane installations now generate a self-signed certificate
    and private key pair for each webhook, to prepare for future work to make
    the proxy injector and service profile validator HA
  * Added the ` config.linkerd.io/enable-debug-sidecar` annotation allowing the
    `--enable-debug-sidecar` flag to work when auto-injecting Linkerd proxies
  * Added multiple replicas for the `proxy-injector` and `sp-validator`
    controllers when run in high availability mode (thanks to @Pothulapati!)
  * Defined least privilege default security context values for the proxy
    container so that auto-injection does not fail (thanks @codeman9!)
  * Default the webhook failure policy to `Fail` in order to account for
    unexpected errors during auto-inject; this ensures uninjected applications
    are not deployed
  * Introduced control plane's PSP and RBAC resources into Helm templates; these
    policies are only in effect if the PSP admission controller is enabled
  * Removed `UPDATE` operation from proxy-injector webhook because pod mutations
    are disallowed during update operations
  * Default the mutating and validating webhook configurations `sideEffects`
    property to `None` to indicate that the webhooks have no side effects on
    other resources (thanks @Pothulapati!)
  * Added support for the SMI TrafficSplit API which allows users to define
    traffic splits in TrafficSplit custom resources
  * Added the `linkerd.io/control-plane-ns` label to all Linkerd resources
    allowing them to be identified using a label selector
  * Added Prometheus metrics for the Kubernetes watchers in the destination
    service for better visibility
* Proxy
  * Replaced the fixed reconnect backoff with an exponential one (thanks,
    @zaharidichev!)
  * Fixed an issue where load balancers can become stuck
  * Added a dispatch timeout that limits the amount of time a request can be
    buffered in the proxy
  * Removed the limit on the number of concurrently active service discovery
    queries to the destination service
  * Fix an epoll notification issue that could cause excessive CPU usage
  * Added the ability to disable tap by setting an env var (thanks,
    @zaharidichev!)
  * Changed the proxy's routing behavior so that, when the control plane does
    not resolve a destination, the proxy forwards the request with minimal
    additional routing logic
  * Fixed a bug in the proxy's HPACK codec that could cause requests with very
    large header values to hang indefinitely
  * Fixed a memory leak that can occur if an HTTP/2 request with a payload ends
    before the entire payload is sent to the destination
  * The `l5d-override-dst` header is now used for inbound service profile
    discovery
  * Added errors totals to `response_total` metrics
  * Changed the load balancer to require that Kubernetes services are resolved
    via the control plane
  * Added the `NET_RAW` capability to the proxy-init container to be compatible
    with `PodSecurityPolicy`s that use `drop: all`
  * Fixed the proxy rejecting HTTP2 requests that don't have an `:authority`
  * Improved idle service eviction to reduce resource consumption for clients
    that send requests to many services
  * Fixed proxied HTTP/2 connections returning 502 errors when the upstream
    connection is reset, rather than propagating the reset to the client
  * Changed the proxy to treat unexpected HTTP/2 frames as stream errors rather
    than connection errors
  * Fixed a bug where DNS queries could persist longer than necessary
  * Improved router eviction to remove idle services in a more timely manner
  * Fixed a bug where the proxy would fail to process requests with obscure
    characters in the URI
* Web UI
  * Added the Font Awesome stylesheet locally; this allows both Font Awesome and
    Material-UI sidebar icons to display consistently with no/limited internet
    access (thanks again, @liquidslr!)
  * Removed the Authorities table and sidebar link from the dashboard to prepare
    for a new, improved dashboard view communicating authority data
  * Fixed dashboard behavior that caused incorrect table sorting
  * Removed the "Debug" page from the Linkerd dashboard while the functionality
    of that page is being redesigned
  * Added an Edges table to the resource detail view that shows the source,
    destination name, and identity for proxied connections
  * Improved UI for Edges table in dashboard by changing column names, adding a
    "Secured" icon and showing an empty Edges table in the case of no returned
    edges
* Internal
  * Known container errors were hidden in the integration tests; now they are
    reported in the output without having the tests fail
  * Fixed integration tests by adding known proxy-injector log warning to tests
  * Modified the integration test for `linkerd upgrade` in order to test
    upgrading from the latest stable release instead of the latest edge and
    reflect the typical use case
  * Moved the proxy-init container to a separate `linkerd/proxy-init` Git
    repository


stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
+## stable-2.4.0
+
+This stable release introduces adds support for SMI [Traffic Split](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md),
```
alpeb: "introduces adds"

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+fine-grained observability into the TLS-based identity system via the new
+`linkerd edges` command, default automatic proxy injection, installation stages,
+and a host of performance and usability improvements in the proxy and Web UI.
+
+Creating an SMI [Traffic Split](https://github.com/deislabs/smi-spec/blob/master/traffic-split.md)
+resource will allow Linkerd to split traffic between the specified backend
+services. The `linkerd edges` command provides a thorough breakdown of the name
+and identity for proxied connections, as well as an exciting new "Edges" table
```
alpeb: "The `linkerd edges` command provides a thorough breakdown of the name
and identity for proxied connections"
That sounds like the connections are named and identified, which might not be totally accurate. Probably needs a rephrase like ".. provides a thorough breakdown of the proxied connections, with the names and identity of each resource on both sides of the connection"

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+in the Web UI!
+
+Significant changes to the installation lifecycle have introduced default
+automatic proxy injection, and optional installation stages via the `linkerd
+install config` and `linkerd install control-plane` commands which separate the
+concerns for cluster-level and namespace-level privileges.
+
```
alpeb: I think the automatic proxy injection require a bit more detail, being such a big change on the UX. From the edge-19.4.5 changelog:
```
With this release the behavior of the linkerd inject command changes, where
the proxy sidecar container YAML is no longer included in its output by
default, but instead it will just add the annotations to defer the injection to
the proxy injector.
For use cases that require the full injected YAML to be output, a new
--manual flag has been added.
```

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+    resource spec for the proxy-init init container
+  * Fixed control plane components failing on startup when the Kubernetes API
+    returns an `ErrGroupDiscoveryFailed`
+  * Added Controller Component Labels to the webhook config resources (thanks,
+    @Pothulapati!)
+  * Moved the tap service into its own pod
+  * **New** Control plane installations now generate a self-signed certificate
+    and private key pair for each webhook, to prepare for future work to make
+    the proxy injector and service profile validator HA
+  * Added a debug container annotation, allowing the `--enable-debug-sidecar`
```
ihcsim: Consider specifying the annotation as you did for `disable-identity`; it's `config.linkerd.io/enable-debug-sidecar`.

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+  * Introduced control plane's PSP and RBAC resources into Helm templates; these
+    policies are only in effect if the PSP admission controller is enabled
+  * Removed `UPDATE` operation from proxy-injector webhook because pod mutations
+    are disallowed during update operations
+  * Default the mutating and validating webhook configurations `sideEffects`
+    property to `None` to indicate that the webhooks have no side effects on
+    other resources (thanks @Pothulapati!)
+  * Added support for the SMI TrafficSplit API which allows users to define
+    traffic splits in TrafficSplit custom resources
+  * Added the `linkerd.io/control-plane-ns` label to the SMI Traffic Split CRD
```
ihcsim: All the other global and namespace-scoped resources are labelled too. I think we can just say the label is added to all Linkerd resources, allowing Linkerd resources to be identified using label selector.

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+  * Defined least privilege default security context values for the proxy
+    container so that auto-injection does not fail on (thanks @codeman9!)
+  * Default the webhook failure policy to `Fail` in order to account for
+    unexpected errors during auto-inject; this ensures uninjected applications
+    are not deployed
+  * Introduced control plane's PSP and RBAC resources into Helm templates; these
+    policies are only in effect if the PSP admission controller is enabled
+  * Removed `UPDATE` operation from proxy-injector webhook because pod mutations
+    are disallowed during update operations
+  * Default the mutating and validating webhook configurations `sideEffects`
```
Pothulapati: Adding the `sideEffects` flag by default, breaks linkerd installations running on cluster versions < 1.12, `--omit-webhook-side-effects` can be used to install in those cases. 
I am guessing, A mention regarding this would be useful for users. WDYT?

Pothulapati: As this field is added by default, It breaks linkerd installations on clusters versions < 1.12, `--omit-webhook-side-effects` can be used for installation in those cases.
I am guessing, A mention about this would be useful. WDYT?

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+    @Pothulapati!)
+  * Moved the tap service into its own pod
+  * **New** Control plane installations now generate a self-signed certificate
+    and private key pair for each webhook, to prepare for future work to make
+    the proxy injector and service profile validator HA
+  * Added a debug container annotation, allowing the `--enable-debug-sidecar`
+    flag to work when auto-injecting Linkerd proxies
+  * Added multiple replicas for the `proxy-injector` and `sp-validator`
+    controllers when run in high availability mode (thanks to @Pothulapati!)
+  * Defined least privilege default security context values for the proxy
```
wmorgan: What does defined mean here? Also spurious "on", I think?

stable-2.4.0.md
ihcsim: I am going to make a mention of the k8s pre-1.12 compatibility issue in the upgrade notes. Not sure if we need to mention it here. If anything, we should mention that k8s 1.12 is now the official minimum supported version.

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+    `--disable-tap` flag, or by using the `config.linkerd.io/disable-tap`
+    annotation
+  * Introduced pre-install healthcheck for clock skew (thanks, @matej-g!)
+  * Added a JSON option to the `linkerd edges` command so that output is
+    scripting friendly and can be parsed easily (thanks @alenkacz!)
+  * Fixed an issue when Linkerd is installed with `--ha`, running `linkerd
+    upgrade` without `--ha` will disable the high availability control plane
+  * Added a `--init-image-version` flag to `linkerd inject` to override the
+    injected proxy-init container version
+  * Added the `--linkerd-cni-enabled` flag to the `install` subcommands so that
```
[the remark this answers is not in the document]
ihcsim: It's more like a bug fix. Previously, `linkerd install --linkerd-cni-enabled` works as expected. But not for the multi install subcommands, i.e. `linkerd install config --linkerd-cni-enabled` and `linkerd install control-plane --linkerd-cni-enabled` didn't work. The fix ensures the flag is honoured by the subcommands.

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+    `NET_ADMIN` capability is omitted from the CNI-enabled control plane's PSP
+  * Updated `linkerd check` to validate the caller can create
+    `PodSecurityPolicy` resources
+  * Added a check to `linkerd install` to prevent installing multiple control
```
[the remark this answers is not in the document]
ihcsim: Good point. The reason behind disabling multiple control plane setup is to avoid conflicts between global resources such as the mutating and validating webhook configurations and CRDs.

stable-2.4.0.md
ihcsim: The main difference in `inject` UX now is that previously, `linkerd inject` outputs the injected YAML blob, but now, it only outputs the annotated YAML (without the proxy YAML). The previous behaviour can be reproduced by using `linkerd inject --manual`. I am on the fence with whether to include this in the release notes. We have doc for this.

stable-2.4.0.md
```diff
@@ -0,0 +1,188 @@
...
+Among the many performance and usability improvements to the proxy and Web UI,
+debugging efforts have been accompanied by a new a new `--enable-debug-sidecar`
+flag for the `linkerd inject` command and more accurate metric collection of
+proxied connections.
+
+TODO: Update blog post link. For more details, see the announcement [blog post](https://linkerd.io/blog/)
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: TODO: Add summary of upgrade notes link. Please see the
```
ihcsim: When https://github.com/linkerd/website/pull/396 is merged and published, the upgrade notes will be at https://linkerd.io/2/tasks/upgrade/#upgrade-notice-stable-2-4-0

stable-2.4.0.md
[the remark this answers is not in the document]
wmorgan: Come to think of it, I'm drafting the announcement blost post right now. It would be good to sync w/this first paragraph and make sure we're telling the same story.

kleimkuhler:
The last open TODO is adding the link for the blog post; it is ready-ready otherwise.

CHANGES.md
```diff
@@ -1,3 +1,206 @@
+## stable-2.4.0
+
+This release adds traffic splitting functionality, support for the Kubernetes
+Service Mesh Interface (SMI), graduates high-availability support out of
+experimental status, and adds a tremendous list of other improvements,
+performance enhancements, and bug fixes
```
wmorgan: Missing period

CHANGES.md
```diff
@@ -1,3 +1,206 @@
...
+
+Linkerd's new traffic splitting feature allows users to dynamically control the
+percentage of traffic destined for a service. This powerful feature can be used
+to implement rollout strategies like canary releases and blue-green deploys.
+
+Support for the [Service Mesh Interface](https://smi-spec.io) (SMI) makes it
```
wmorgan: I would just merge the first sentence of this paragraph into the previous one, and kill the Flagger reference -- doesn't feel right for release notes.

PR #3039, opened 2019-07-05
Fix linkerd inject url handling

### LINKED ISSUE #3037: `linkerd inject` doesn't work when supplying a full path to a file outside the current dir
`linkerd inject /foo/bar/baz.yml` returns:
```
Error: Get /foo/bar/baz.yml: unsupported protocol scheme ""
```

I believe this bug was introduced in #2988 in edge-19.6.4

CC @Pothulapati

siggy:
PR #2988 introduced support for `linkerd inject` with a URL parameter.
Some edge cases such as `/foo/bar/baz.yml` will return without error
from `url.ParseRequestURI`, even though we know this is a valid local
filename.

Modify inject's URL check to specifically look for a valid `http[s]://`
prefix. If that check succeeds, we will still rely on `http.Get` for URL
validation, as it calls `url.Parse` under the hood.

Fixes #3037

### DIFF
--- a/cli/cmd/inject_util.go
+++ b/cli/cmd/inject_util.go
@@ -6,9 +6,9 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/linkerd/linkerd2/pkg/inject"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


siggy:
@Pothulapati I really did try to add a test file, but the `read` function depends on filesystem and network and did not want to abstract that away for such a small change. 😄

siggy:
closing in favor of #3038

PR #3047, opened 2019-07-08
Tweak the vertical positioning of arrows in octopus graph

klingerf:
This branch fixes two small issues that impacted the height of the arrows drawn in the octopus graph. The code to calculate the height of each arrow was not taking into account the 10px curve that we draw at the top and bottom of each arrow. This caused the arrows to be 20px longer than they should be. This was not noticeable in most cases, however, because the box height that we were using to calculate arrow height did not take into account 24px of vertical padding around the boxes. So often these two issues canceled each other out, but with more nodes on the graph, it looked like this:

Screen Shot 2019-07-08 at 1 05 30 PM

Whereas with the fix from this branch, it looks like this:

Screen Shot 2019-07-08 at 1 05 34 PM

Note that when we have 7 or more inbound or outbound boxes, we're aggregating them into a single box. It previously looked like this:

Screen Shot 2019-07-08 at 1 13 53 PM

And now it looks like this:

Screen Shot 2019-07-08 at 1 13 56 PM

It's a bit unfortunate that the overflow box is taller than it needs to be, but that seems preferable to adding support for dynamically sized arrows based on the box content.

### DIFF
--- a/web/app/js/components/Octopus.jsx
+++ b/web/app/js/components/Octopus.jsx
@@ -39,10 +39,12 @@ const styles = () => ({
     marginRight: "auto"
   },
   centerNode: {
-    width: "244px"
+    width: "244px",
+    height: `${baseHeight - 24}px` // grid alignment adds 24px vertical padding
   },
   neighborNode: {
-    width: "220px"
+    width: "220px",
+    height: `${baseHeight - 24}px` // grid alignment adds 24px vertical padding
   }
 });
 class Octopus extends React.Component {
--- a/web/app/js/components/util/OctopusArms.jsx
+++ b/web/app/js/components/util/OctopusArms.jsx
@@ -4,7 +4,7 @@ import grey from '@material-ui/core/colors/grey';
 const strokeOpacity = "0.7";
 const arrowColor = grey[500];
 
-export const baseHeight = 220; // the height of the neighbor node box
+export const baseHeight = 242; // the height of the neighbor node box plus padding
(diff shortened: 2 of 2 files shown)


alpeb:
This tested fine for me :+1: 
I wasn't aware we had a limit of 6 boxes plus one for overflows... and speaking of the overflow box size, it seems it's equipped to show only 5 entries, which makes for a total of 11 inbound resources, and any other inbound resource is not visible. Perhaps something to address in a different PR (not sure if a simple scrollbar inside that box would be usable).

alpeb:
... well, this is how it would look like with `overflow-y: auto`. It ain't that ugly
image

klingerf:
Sorry for the delay here, and thanks for the reviews. Given the issues brought up, I'd like to try to figure out a way to support a variable-height box / arrow for the collapsed neighbors node, prior to this branch being merged. But I'm not actively working on that. I'm happy to close this PR out until that functionality is available.

scottcarol:
Hey @klingerf, thanks for your work on this and for highlighting (and solving!) the original issue of arrow positioning. One solution for the collapsed neighbors node could be to pass in a boolean prop to each node called `variableHeight`. I'm happy to take the baton on this one and continue what you started. 🏁 🏃

scottcarol:
As discussed with @klingerf, feedback on this PR raised some other issues that need to be addressed. Closing this PR and creating an issue for future work 📆

PR #3056, opened 2019-07-10
Introduce Cluster Heartbeat cronjob

### LINKED ISSUE #2961: Version usage reporting

It is extremely valuable to know what versions of Linkerd are currently in use and the environments they're being used in. This helps prioritize testing, backports and gives a picture of what is important to users.

As part of install, add a `CronJob` that issues a request to `versioncheck.linkerd.io` every 24 hours with the following query params:

#### P0

- [x] #3056 Linkerd version
- [x] #3056 Cluster UUID
- [x] #3056 Kubernetes version
- [x] #3056 Installation time

#### P1

- [x] #3238 CPU/Memory for prometheus and destination
  - max mem over all available time
  - p95 cpu over all available time
- [x] #3238 CPU/Memory for the data plane proxies

#### P2

- [ ] Memory histogram of the data plane proxies
- [ ] Histogram of latency introduced by the proxy

#### P3

- [x] #3059 Number of injections
- [ ] CPU/memory usage of each container in the control plane
- [ ] Histogram of cpu usage of the data plane proxies
- [ ] Histogram of RPS by proxy

siggy:
`linkerd check`, the web dashboard, and Grafana all perform version
checks to validate Linkerd is up to date. It's common for users to
seldom execute these codepaths. This makes it difficult to identify what
versions of Linkerd are currently in use and what environments it is
being run in, which helps prioritize testing and backports.

Introduce a `heartbeat` CronJob to the default Linkerd install. The
cronjob executes every 24 hours, starting from 5 minutes after
`linkerd install` is run.

Example check URL:
https://versioncheck.linkerd.io/version.json?
  install-time=1562761177&
  k8s-version=v1.15.0&
  meshed-pods=8&
  total-rps=3&
  source=heartbeat&
  uuid=cc4bb700-3314-426a-9f0f-ec588b9df020&
  version=git-b97ee9f7

Fixes #2961

### DIFF
--- a/chart/templates/heartbeat-rbac.yaml
+++ b/chart/templates/heartbeat-rbac.yaml
@@ -1,0 +1,48 @@
+{{with .Values -}}
+---
+###
+### Heartbeat RBAC
+###
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: linkerd-heartbeat
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get"]
+  resourceNames: ["linkerd-config"]
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["get"]
+  resourceNames: ["{{.Namespace}}"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: linkerd-heartbeat
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+roleRef:
+  kind: Role
+  name: linkerd-heartbeat
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: linkerd-heartbeat
+  namespace: {{.Namespace}}
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-heartbeat
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: heartbeat
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+{{- end}}
--- a/chart/templates/heartbeat-rbac.yaml
+++ b/chart/templates/heartbeat-rbac.yaml
@@ -16,10 +16,6 @@ rules:
   resources: ["configmaps"]
   verbs: ["get"]
   resourceNames: ["linkerd-config"]
-- apiGroups: [""]
-  resources: ["namespaces"]
-  verbs: ["get"]
-  resourceNames: ["{{.Namespace}}"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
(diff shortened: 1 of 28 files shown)


chart/templates/heartbeat-rbac.yaml
```diff
@@ -0,0 +1,48 @@
...
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get"]
+  resourceNames: ["linkerd-config"]
+- apiGroups: [""]
+  resources: ["namespaces"]
```
adleong: Could we omit this privilege if we used the creation date of the configmap to determine the age of the install instead of the creation date of the namespace?

chart/templates/heartbeat.yaml
```diff
@@ -0,0 +1,42 @@
...
+          containers:
+          - name: heartbeat
+            image: {{.ControllerImage}}
+            imagePullPolicy: {{.ImagePullPolicy}}
+            args:
+            - "heartbeat"
+            - "-prometheus-url=http://linkerd-prometheus.{{.Namespace}}.svc.cluster.local:9090"
+            - "-controller-namespace={{.Namespace}}"
+            - "-log-level={{.ControllerLogLevel}}"
+            {{- include "resources" .HeartbeatResources | indent 4 | trimPrefix "    " }}
```
adleong: My go template foo is weak, can you explain this a bit?  why does it look different from the resource includes in other, uh.... resources?

siggy:
@adleong 
> Do we want to expose any configuration as install flags? Perhaps the ability to change the schedule or disable the cronjob entirely?

I'm not sure I see a use-case for changing the schedule. As it only impacts collection of this data, I think it's preferable for the schedules to be the same.

Disabling the cronjob is a valid use case, though I think user's could just remove it from the install yaml.

> Is the Linkerd proxy injected into the cronjob pod? Do we want it to be?

It's not injected, I don't think we want it to be, as a k8s CronJob is just a type of Job, which would exhibit the sidecar shutdown problem.

grampelberg:
I'd like for users to be able to opt-out of this, but that feels like we might as well exclude the CronJob entirely from `install` output. For anyone who *really* wants to modify the configuration, the kustomize method should work well enough for them.

chart/templates/heartbeat-rbac.yaml
siggy: Great idea! Updated.

chart/templates/heartbeat.yaml
siggy: Yeah... this is an unfortunate side effect of how the [_resources.yaml template](https://github.com/linkerd/linkerd2/blob/f535c2d3d2f0da76bd62f3cbc1e39994a773ed21/chart/templates/_resources.yaml#L2) has a hard-coded indentation. This works for every other component because they are all under:
```yaml
spec:
  template:
    spec:
      containers:
      - resources:
```
...but the heartbeat cronjob is structured as:
```yaml
spec:
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - resources:
```
... the extra 2 levels motivates the `indent 4` (and then trimming the extra ws on the previous line)

a better fix would be to make `_resources.yaml` agnostic to indentation level, but i decided that was out of scope for this PR.

grampelberg: You're looking for `nindent` - https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#using-the-include-function

siggy: filed #3119

PR #3064, opened 2019-07-10
Sort slices to avoid flakeyness in tests

adleong:
The `TestGetServicesFor` is flaky because it compares two slices of services which are in a non-deterministic order.

To make this deterministic, we first sort the slices by UID.  We also introduce a `byUID` type to consolidate this with similar sorting logic in another test.


controller/k8s/test_helper.go
```diff
@@ -33,10 +34,30 @@ func NewFakeAPI(configs ...string) (*API, error) {
...
+
+func (s byUID) Len() int { return len(s) }
+
+func (s byUID) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 
-func (s byPod) Len() int { return len(s) }
+func (s byUID) Less(i, j int) bool { return s[i].GetUID() < s[j].GetUID() }
 
-func (s byPod) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+func podByUID(pods []*corev1.Pod) byUID {
```
adleong: This is necessary because it is not possible to treat a `[]*corev1.Pod` as a `[]hasUID` even though `Pod` implements `hasUID`.  I can only assume that this design decision arose from a deep hatred for humankind.

adleong:
@jonathanbeber you may be interested in reviewing this

controller/k8s/test_helper.go
jonathanbeber: that's really strange for me, I found some explanation [here](https://stackoverflow.com/questions/12990338/cannot-convert-string-to-interface/12990540#12990540) and [here](https://stackoverflow.com/a/12994852/5164435) to give some peace for my mind. Do you think, is it worth to comment something here explaining this decision?

siggy: what about this pattern...
https://github.com/linkerd/linkerd2/blob/bdf5b46d093a3ff45f52b9538fee046d2449890e/controller/api/public/grpc_server_test.go#L33-L45

controller/k8s/api_test.go
```diff
@@ -879,8 +879,8 @@ status:
 				t.Fatalf("api.GetPodsFor() unexpected error, expected [%s] got: [%s]", exp.err, err)
 			}
 
-			sort.Sort(byPod(pods))
-			sort.Sort(byPod(k8sResultPods))
+			sort.Sort(podByUID(pods))
+			sort.Sort(podByUID(k8sResultPods))
```
siggy: i think this will only sort the new slice created by `podByUID`, and leave the original ones, that get passed into `DeepEqual`, unchanged.

jonathanbeber: you're right, since it returns a new array, the `sort.Sort` function just sort it and let the original one as it's.

jonathanbeber: I could see test running for multiple times without flakiness with these changes: https://gist.github.com/jonathanbeber/33adb4bb051142eb2a30e24e938933ba

adleong:
Thanks for catching this @siggy and @jonathanbeber!

PR #3065, opened 2019-07-10
Add pod status to linkerd check

### LINKED ISSUE #2877: Show pod status in linkerd check
When there is a problem bringing up controller pods, Linkerd check will wait at this step:

```
| controller pod is running -- waiting for check to complete
```

It can be unclear if I need to continue waiting, or if there is some underlying problem I need to correct.  By looking at `kubectl get pods` I can see that my controller pods are in `Init:ImagePullBackOff` which tells me the image can't be found.  By including that pod status in the check message, I will know immediately that I need to take action.

Similarly, seeing the pod status during the "waiting for controller pods to become ready" step can be useful as well, since it can show if the pods are in CrashLoopBackoff.

adleong:
When waiting for controller pods to be created or become ready, `linkerd check` doesn't offer any hints as to whether there has been an error (such as an ImagePullBackoff).

We add pod status to the output to make this more immediately obvious.

Sample output:

```
linkerd-existence
-----------------
√ control plane components ready
√ no unschedulable pods
- controller pod is running -- linkerd-controller-6d5754cc77-2g59s status is Init:ErrImagePull
```

```
linkerd-existence
-----------------
√ control plane components ready
√ no unschedulable pods
/ controller pod is running -- linkerd-controller-fc99df966-7sm7x status is CrashLoopBackOff
```

Fixes #2877


grampelberg:
The wait messages are a little weird now:

```
control plane pods are ready -- The "linkerd-prometheus-5cd5f895b6-572xz" pod's "prometheus" container is not ready 
```

I’m not sure I have a good suggestion though.

adleong:
@grampelberg do you think it would be an improvement to just condense it like:

`control plane pods are ready -- linkerd-prometheus-5cd5f895b6-572xz:prometheus is not ready`

Not sure if the colon for indicating the container of a pod is clear or if there's any pre-existing notation for this...

siggy:
FWIW we intentionally removed these kinds of messages in #2087 (in service to #1471 and related tickets).

The intent is that it's confusing to the user to display error messages if `linkerd check` is waiting/retrying, and still on its way to successful completion.

I agree it's valuable to quickly surface failures to the user, rather than waiting for `linkerd check` to timeout. For those use cases, we have `--wait=0` and `--verbose`.

adleong:
Thanks @siggy, this is really helpful context.  I think that now that we have a spinner for retries rather than outputting a new line for each retry, having the current status in that line is really helpful.  As a user, I'd like to know what I'm waiting for so that I can know if it's appropriate to continue to wait (e.g. if the init container is running or the image is pulling) or if I should cancel (e.g. if the pod is in ImagePullBackoff).  

The current state of affairs is that I wait until it feels like it's been taking too long, and then I run `k get pods`, which is pretty annoying.

Let me see if I can adjust the text to make sure that it's clear that check is waiting/retrying and hasn't given up.

adleong:
I've tweaked the output so that the check description is no longer displayed during retries.  For example, it looks like this while retrying:

```
linkerd-api
-----------
- The "linkerd-controller-6d79bcd547-ggrvd" pod's "public-api" container is not ready
```

and then this after it has succeeded

```
linkerd-api
-----------
```

This avoids confusing messages like "control plane pods are ready -- xxx is not ready".

I played with a few different wordings to indicate that this state is not final, but in the end, the thing that was most compelling for me was just the spinning spinner.  The fact that the spinner is going indicates to me that `linkerd check` is still trying and that the status message is not final.

Let me know what you think.

scottcarol:
It worked for me! 
Screen Shot 2019-07-16 at 10 56 20 AM
 
However since I don't have `linkerd check` memorized, I didn't know exactly what I was looking at. I prefer the original UI in this PR where you had the descriptive message AND the status, because you're giving context to the user for what that status means. Have you thought of printing the status below the descriptor as we do for warnings?

```bash
\ control plane pods are ready
      linkerd-controller-7d9b45f5b9-fhz86 container public-api is not ready
```

and then keep success the same 
```bash
```

adleong:
Thanks for testing, @scottcarol!

My hesitation around your suggestion is that it reads a little contradictorily: "control plane pods are ready -- [a control plane pod] is not ready" which is confusing.

My hope was that the message: "[Pod] container [container] is not ready" was self-explanatory and could stand alone.

pkg/k8s/api.go
```diff
@@ -133,7 +133,7 @@ func (kubeAPI *KubernetesAPI) GetReplicaSets(namespace string) ([]appsv1.Replica
 // GetPodStatus receives a pod and returns the pod status, based on `kubectl` logic.
 // This logic is imported and adapted from the github.com/kubernetes/kubernetes project:
 // https://github.com/kubernetes/kubernetes/blob/33a3e325f754d179b25558dee116fca1c67d353a/pkg/printers/internalversion/printers.go#L558-L640
-func GetPodStatus(pod *corev1.Pod) string {
+func GetPodStatus(pod corev1.Pod) string {
```
scottcarol: For my own understanding, what was the reason for the change here?

scottcarol:
To me, the spinner makes it clear that "control plane pods are ready" is not true yet. But I don't have a strong preference! A different way to be more explicit would be to add `po/` to the pod name so that the user can glance at it and know that a pod check is going on.

pkg/k8s/api.go
adleong: It's a really boring reason.  In `checkControllerRunning` we process a slice of Pods (not Pod pointers) and wish to call this function.  We could simply pass a reference to the pod into this function and it would work fine, but our linter doesn't like that we'd be taking the reference to a loop variable and complains about it.

scottcarol: Cool, thanks for the explanation.

PR #3066, opened 2019-07-10
Add a flag to install-cni command to configure iptables wait flag

cpretzer:
In #2970 there are error messages in the kubelet log which indicate that the xtables is locked by another process and the cni-plugin cannot add an entry to iptables.

This commit adds a flag named `--use-wait-flag` to the `linkerd install-cni` command in order to allow the use of the `-w` to be configurable.

It's worth noting that we don't know what is locking xtables in #2970. There's a good chance that this fix merely exposes another issue or causes the pods to start slowly because the CNI is waiting for the xtables lock.


cli/cmd/install-cni-plugin.go
```diff
@@ -120,6 +123,11 @@ assumes that the 'linkerd install' command will be executed with the
 	cmd.PersistentFlags().StringVar(&options.logLevel, "cni-log-level", options.logLevel, "Log level for the cni-plugin.")
 	cmd.PersistentFlags().StringVar(&options.destCNINetDir, "dest-cni-net-dir", options.destCNINetDir, "Directory on the host where the CNI configuration will be placed.")
 	cmd.PersistentFlags().StringVar(&options.destCNIBinDir, "dest-cni-bin-dir", options.destCNIBinDir, "Directory on the host where the CNI plugin binaries reside.")
+	cmd.PersistentFlags().BoolVar(
+		&options.useWaitFlag,
+		"use-wait-flag",
+		options.useWaitFlag,
+		"Configures the CNI plugin to use the \"-w\" flag for the iptables command.")
```
alpeb: Maybe put in the description the default value (false) in parenthesis?

pkg/version/version.go
```diff
@@ -16,7 +16,7 @@ var Version = undefinedVersion
 // https://github.com/linkerd/linkerd2-proxy-init
 // This has to be kept in sync with the constraint version for
 // github.com/linkerd/linkerd2-proxy-init in /Gopkg.toml
-var ProxyInitVersion = "v1.0.0"
+var ProxyInitVersion = "87ab4bd22a6e45ae74ff05efca7c7bed5b5a5cce"
```
alpeb: This will break installations without `--linkerd-cni-enabled` because there's no init container image published with this tag. I think we should leave it to `v1.0.0` while this PR gets tested, and after the changes in the `linkerd2-proxy-init` repo get published under a new version tag we can set it here, although that wouldn't be strictly necessary since those changes only apply to the CNI script, not to the init containers.

cpretzer: Released [linkerd2/proxy-init v1.1.0](https://github.com/linkerd/linkerd2-proxy-init/releases) to address this.

cli/cmd/install-cni-plugin.go
cpretzer: Added default value to help in [this commit](https://github.com/linkerd/linkerd2/commit/43cb598ca5a23b736a02cf3a68521a6ef9228222)

cli/cmd/install-cni-plugin.go
```diff
@@ -127,7 +127,7 @@ assumes that the 'linkerd install' command will be executed with the
 		&options.useWaitFlag,
 		"use-wait-flag",
 		options.useWaitFlag,
-		"Configures the CNI plugin to use the \"-w\" flag for the iptables command.")
+		"Configures the CNI plugin to use the \"-w\" flag for the iptables command. Default value is false")
```
alpeb: Nit: the standard we follow in other commands is to have the default in parenthesis
`Configures the CNI plugin to use the \"-w\" flag for the iptables command (default false)`

alpeb:
@cpretzer Besides my nit above, could you also please fix the fixtures in the unit tests? They're still referring to proxy-init v1.0.0.
Running `go test ./... -update` will automatically make some of the needed changes, and then you can check with git diff and verify if those were indeed the changes you wanted. Some other changes (like in webhook_test.go) will need a manual fix.

ihcsim:
Before merging this PR into `master`, is it possible to create a patched image of the CNI with this change, and get the user to retry it, to see if adding the `-w` flag resolves #2970?

cli/cmd/install-cni-plugin.go
cpretzer: Thanks for letting me know about the standard. I've updated the description

cpretzer:
@ihcsim, I can create that image.

@tiurikov, will you be able to test the image?

alpeb:
@cpretzer Could you also please apply this patch? This adds a label to the PSP resource that `linkerd install-cni` creates, that will help `bin/test-cleanup` remove it when needed. This change also requires updating the test fixtures one more time :grimacing: 

```
diff --git a/cli/cmd/install-cni-plugin.go b/cli/cmd/install-cni-plugin.go
index 8a5c73d5..c082f5de 100644
--- a/cli/cmd/install-cni-plugin.go
+++ b/cli/cmd/install-cni-plugin.go
@@ -16,19 +16,20 @@ import (
 )
 
 type installCNIPluginConfig struct {
-       Namespace           string
-       CNIPluginImage      string
-       LogLevel            string
-       InboundPort         uint
-       OutboundPort        uint
-       IgnoreInboundPorts  string
-       IgnoreOutboundPorts string
-       ProxyUID            int64
-       DestCNINetDir       string
-       DestCNIBinDir       string
-       CreatedByAnnotation string
-       CliVersion          string
-       UseWaitFlag         bool
+       Namespace                string
+       ControllerNamespaceLabel string
+       CNIPluginImage           string
+       LogLevel                 string
+       InboundPort              uint
+       OutboundPort             uint
+       IgnoreInboundPorts       string
+       IgnoreOutboundPorts      string
+       ProxyUID                 int64
+       DestCNINetDir            string
+       DestCNIBinDir            string
+       CreatedByAnnotation      string
+       CliVersion               string
+       UseWaitFlag              bool
 }
 
 type cniPluginOptions struct {
@@ -150,19 +151,20 @@ func validateAndBuildCNIConfig(options *cniPluginOptions) (*installCNIPluginConf
        }
 
        return &installCNIPluginConfig{
-               Namespace:           controlPlaneNamespace,
-               CNIPluginImage:      options.taggedCNIPluginImage(),
-               LogLevel:            options.logLevel,
-               InboundPort:         options.inboundPort,
-               OutboundPort:        options.outboundPort,
-               IgnoreInboundPorts:  strings.Join(ignoreInboundPorts, ","),
-               IgnoreOutboundPorts: strings.Join(ignoreOutboundPorts, ","),
-               ProxyUID:            options.proxyUID,
-               DestCNINetDir:       options.destCNINetDir,
-               DestCNIBinDir:       options.destCNIBinDir,
-               CreatedByAnnotation: k8s.CreatedByAnnotation,
-               CliVersion:          k8s.CreatedByAnnotationValue(),
-               UseWaitFlag:         options.useWaitFlag,
+               Namespace:                controlPlaneNamespace,
+               ControllerNamespaceLabel: k8s.ControllerNSLabel,
+               CNIPluginImage:           options.taggedCNIPluginImage(),
+               LogLevel:                 options.logLevel,
+               InboundPort:              options.inboundPort,
+               OutboundPort:             options.outboundPort,
+               IgnoreInboundPorts:       strings.Join(ignoreInboundPorts, ","),
+               IgnoreOutboundPorts:      strings.Join(ignoreOutboundPorts, ","),
+               ProxyUID:                 options.proxyUID,
+               DestCNINetDir:            options.destCNINetDir,
+               DestCNIBinDir:            options.destCNIBinDir,
+               CreatedByAnnotation:      k8s.CreatedByAnnotation,
+               CliVersion:               k8s.CreatedByAnnotationValue(),
+               UseWaitFlag:              options.useWaitFlag,
        }, nil
 }
 
diff --git a/cli/install/cni-template.go b/cli/install/cni-template.go
index 8951a48c..4662eb62 100644
--- a/cli/install/cni-template.go
+++ b/cli/install/cni-template.go
@@ -30,6 +30,8 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: linkerd-{{.Namespace}}-cni
+  labels:
 spec:
   allowPrivilegeEscalation: false
   fsGroup:
```

alpeb:
I also missed labeling the other resources in cni-template.go:
```
diff --git a/cli/install/cni-template.go b/cli/install/cni-template.go
index 8951a48c..44d50bb0 100644
--- a/cli/install/cni-template.go
+++ b/cli/install/cni-template.go
@@ -56,6 +56,8 @@ kind: Role
 metadata:
   name: linkerd-cni
   namespace: {{.Namespace}}
+  labels:
 rules:
 - apiGroups: ['extensions', 'policy']
   resources: ['podsecuritypolicies']
@@ -68,6 +70,8 @@ kind: RoleBinding
 metadata:
   name: linkerd-cni
   namespace: {{.Namespace}}
+  labels:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
@@ -76,6 +80,8 @@ subjects:
 - kind: ServiceAccount
   name: linkerd-cni
   namespace: {{.Namespace}}
+  labels:
 ---
 # Include a clusterrole for the linkerd CNI DaemonSet,
 # and bind it to the linkerd-cni serviceaccount.
@@ -83,6 +89,8 @@ kind: ClusterRole
 metadata:
   name: linkerd-cni
+  labels:
 rules:
 - apiGroups: [""]
   resources: ["pods", "nodes", "namespaces"]
@@ -92,6 +100,8 @@ apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
   name: linkerd-cni
+  labels:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
@@ -107,6 +117,8 @@ apiVersion: v1
 metadata:
   name: linkerd-cni-config
   namespace: {{.Namespace}}
+  labels:
 data:
   incoming_proxy_port: "{{.InboundPort}}"
   outgoing_proxy_port: "{{.OutboundPort}}"
@@ -154,6 +166,7 @@ metadata:
   namespace: {{.Namespace}}
   labels:
     k8s-app: linkerd-cni
   annotations:
     {{.CreatedByAnnotation}}: {{.CliVersion}}
 spec:
```

alpeb:
Turns out with these new labels, after the cni DaemonSet is installed we can't install linkerd:
```$ linkerd install|k apply -f -
Unable to install the Linkerd control plane. It appears that there is an existing installation:

clusterrole.rbac.authorization.k8s.io/linkerd-cni
clusterrolebinding.rbac.authorization.k8s.io/linkerd-cni
podsecuritypolicy.policy/linkerd-linkerd-cni

If you are sure you'd like to have a fresh install, remove these resources with:

    linkerd install --ignore-cluster | kubectl delete -f -

Otherwise, you can use the --ignore-cluster flag to overwrite the existing global resources.
```
I'd like to keep these labels for `bin/test-cleanup` to do its job. We could fix this by having more restricted label selectors in the checks for `LinkerdPreInstallGlobalResourcesChecks`.

Gopkg.lock
```diff
@@ -449,15 +449,15 @@
...
   name = "github.com/linkerd/linkerd2-proxy-init"
   packages = [
     "cmd",
     "iptables",
   ]
   pruneopts = ""
-  revision = "458ed04ebd1d23db106200b49afb65fc7f6db47b"
-  version = "v1.0.0"
+  revision = "8d2e800f589f066cfc32a391736b5fe72c90e711"
+  version = "1.1.0"
```
alpeb: @cpretzer Could you please remove the `1.1.0` tag from the `linkerd2-proxy-init` repo and replace it with `v1.1.0` so we keep the same naming scheme? After that, running `bin/dep ensure` should automatically update that tag here.

cpretzer: @alpeb, yes, I'll make that change today

alpeb:
@cpretzer After having run `linkerd install-cni |kubectl apply -f -` if I run install it fails with an error:
```bash
$ linkerd install --linkerd-cni-enabled
Unable to install the Linkerd control plane. It appears that there is an existing installation:

clusterrole.rbac.authorization.k8s.io/linkerd-cni
clusterrolebinding.rbac.authorization.k8s.io/linkerd-cni
podsecuritypolicy.policy/linkerd-linkerd-cni

If you are sure you'd like to have a fresh install, remove these resources with:

    linkerd install --ignore-cluster | kubectl delete -f -

Otherwise, you can use the --ignore-cluster flag to overwrite the existing global resources.
```
The problem is the labels (`linkerd.io/control-plane-ns`) I asked you to add to all the resources created by `linkerd install-cni`.
`linkerd install` thinks there's already a linkerd installed in the cluster and so it fails.

One option to fix this could be to add an extra label into those resources and modify the functions in the healthchecker so it skips the resources having that label:
https://github.com/linkerd/linkerd2/blob/d64a2f368920c8cb269dd3ca88c98742bd18ebda/pkg/healthcheck/healthcheck.go#L1013-L1029

The purpose of that label that you added was just so that `bin/test-cleanup` could find those resources and delete them. So a less complex solution could be to change that label to something else, like `linkerd.io/cni-resource: true` and add a few lines into `bin/test-cleanup` to delete those particular resources. Let me know what you think.

Also, I saw `Gopk.lock` and `Gopkg.toml` got readded in one of your commits. Those files should no longer be needed after we moved to use go modules.

cpretzer:
@alpeb thanks for the review. I'll get these items fixed today

cpretzer:
@alpeb I'm thinking about expanding on the solution that you mention above and pass the `installOptions` from `install.go` to `healthcheck.go`. That way the HealthChecker can conditionally check whether to skip the CNI resources.

The added benefit to this is that if the CNI resources are there and shouldn't be, then we can display an error message in that case.

What do you think?

cli/cmd/install-cni-plugin.go
```diff
@@ -120,6 +124,11 @@ assumes that the 'linkerd install' command will be executed with the
 	cmd.PersistentFlags().StringVar(&options.logLevel, "cni-log-level", options.logLevel, "Log level for the cni-plugin.")
 	cmd.PersistentFlags().StringVar(&options.destCNINetDir, "dest-cni-net-dir", options.destCNINetDir, "Directory on the host where the CNI configuration will be placed.")
 	cmd.PersistentFlags().StringVar(&options.destCNIBinDir, "dest-cni-bin-dir", options.destCNIBinDir, "Directory on the host where the CNI plugin binaries reside.")
+	cmd.PersistentFlags().BoolVar(
+		&options.useWaitFlag,
+		"use-wait-flag",
+		options.useWaitFlag,
+		"Configures the CNI plugin to use the \"-w\" flag for the iptables command. (default false)")
```
alpeb: Minor nit: command descriptions shouldn't end in dot. While you're at it, would you please also remove the period from the descriptions for `--cni-log-level`, `--dest-cni-bin-dir` and `--dest-cni-net-dir`?

cpretzer: @alpeb will do thanks

pkg/healthcheck/healthcheck.go
```diff
@@ -1020,7 +1027,16 @@ func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
 	}
 
 	objects := []runtime.Object{}
+
 	for _, item := range crList.Items {
+
+		if hc.Options.NoInitContainer {
+			label := item.ObjectMeta.Labels[linkerdCniResourceLabel]
+			if label == "true" {
+				continue
```
alpeb: TIOLI: Given this is repeated in the functions below, I'd DRY it into a separate function...

alpeb:
Thanks @cpretzer . The approach looks fine to me :+1: I left a couple of comments about minor things above.
It still looks like there are some issues though. First I'm gonna try publishing the `gcr.io/linkerd-io/proxy-init:v1.1.0` image to see if that fixes the non-cni scenario.

alpeb:
After having published the the new proxy-init image the integration tests are working fine locally.
You can rebase from master and maybe will have to run `bin/update-go-deps-shas` again, and l5d-bot should be happy again :robot:

PR #3069, opened 2019-07-11
Fix upgrade tests conflicting with integration

siggy:
Integration tests on master broke following the 2.4 release, caused by
the recent disabling of multi control-plane support, coupled with the
upgrade integration test (which now upgrades from 2.4 to current sha).

The integration tests do the following:
1. install the current sha
2. test the current sha
3. install the latest stable in an `upgrade` namespace
4. in the `upgrade` namespace, upgrade from stable to latest sha
5. test the upgraded installation

Step 3 breaks because `linkerd install` with stable-2.4 will fail if
existing global resources (from step 1) are present.

For now, modify the integration tests to do the following:
1. install the latest stable in an `upgrade` namespace
2. in the `upgrade` namespace, upgrade from stable to latest sha
3. test the upgraded installation
4. upon successful step 3, remove all related resources
5. install the current sha
6. test the current sha

### DIFF
--- a/bin/test-run
+++ b/bin/test-run
@@ -35,7 +35,7 @@ function run_test(){
     shift
 
     printf "Running test [%s] %s\\n" "$(basename "$filename")" "$@"
-    go test -v "$filename" --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@"
+    go test "$filename" --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@"
 }
 
 # Install the latest stable release.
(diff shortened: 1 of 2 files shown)


bin/test-run
```diff
@@ -88,8 +88,16 @@ printf "Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\\n" "$linke
 
 exit_code=0
 
-run_test "$test_directory/install_test.go" --linkerd-namespace=$linkerd_namespace || exit_code=$?
 run_upgrade_test "$linkerd_namespace"-upgrade
+# if run_upgrade_test succeeded, delete the installation to prepare for subsequent
+# full integration tests.
+# TODO: Consider running the the upgrade tests and normal integration tests on
```
alpeb: nanonit: `the the`

bin/test-run
```diff
@@ -88,8 +88,16 @@ printf "Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\\n" "$linke
 
 exit_code=0
 
-run_test "$test_directory/install_test.go" --linkerd-namespace=$linkerd_namespace || exit_code=$?
 run_upgrade_test "$linkerd_namespace"-upgrade
```
alpeb: missing `|| exit_code=$?` that will get used in the next `if` below?

siggy: `exit_code` was getting set in the `run_upgrade_test`, but i've moved it to here, probably clearer.

PR #3071, opened 2019-07-11
Add bin/git-commit-proxy-version

olix0r:
Each time we update the proxy from the linkerd2-proxy repo, we make the
change slightly differently. The bin/git-commit-proxy-version does all the
steps needed to update the proxy version up to and including making a
commit to this repo.

The proxy version is now stored in a .proxy-version file and is
consumed directly by Dockerfile-proxy, which both simplifies the
Dockerfile and the update process.

This script formats commit messages and emits output as follows:

```
commit c05198a851f69bdc7007974a0ef1f4c01c98d0ce (HEAD -> ver/proxy-update)
Author: Oliver Gould <ver@buoyant.io>
Date:   Thu Jul 11 17:23:05 2019 +0000

    proxy: Update to linkerd/linkerd2-proxy#3a3ec3b

    * linkerd/linker2-proxy#0cc58cd fallback: Clarify fallback layering (linkerd/linkerd2-proxy#288)
    * linkerd/linker2-proxy#b71349a Replace `log` and `env-logger` with `tracing` and `tracing-fmt` (linkerd/linkerd2-proxy#277)
    * linkerd/linker2-proxy#3a3ec3b Use a constant-time load balancer (linkerd/linkerd2-proxy#266)

diff --git a/.proxy-version b/.proxy-version
index f81f40de..d7faa12d 100644
--- a/.proxy-version
+++ b/.proxy-version
@@ -1 +1 @@
-05b012d
+3a3ec3b
```


olix0r:
> can the script check for a clean git repo, maybe even create a new branch? i was making some local modifications to test this thing and ended up committing my own changes (along with a proxy update) to your branch locally.

That's surprising. Since the [`git commit`](https://github.com/linkerd/linkerd2/pull/3071/files#diff-61af07d1f7f9f9dfd9503eaa54c5ef6dR34) references a single file, it _should_ only commit that file and not other staged changes... I'll see if I can repro

olix0r:
I've updated the script to resolve the provided proxy revision, so now symbolic names like `master` can be used when updating the proxy.

We also now open an editor so the user can add notes to the commit if necessary.

The script has been renamed to reflect the fact that it interacts with git.

hawkw:
> I've updated the script to resolve the provided proxy revision, so now symbolic names like `master` can be used when updating the proxy.

This is awesome!

bin/git-commit-proxy-version
```diff
@@ -0,0 +1,44 @@
...
+    >>"$tmp/commit.template"
+
+cd "$rootdir"
+echo "$new_proxy_version" >"$rootdir/.proxy-version"
+
+git commit --file="$tmp/commit.template" --edit -- "$rootdir/.proxy-version"
+
+rm -rf "$tmp"
+
+git --no-pager show HEAD
```
hawkw: TIL that `--no-pager` is a thing!

bin/git-commit-proxy-version
```diff
@@ -0,0 +1,44 @@
...
+old_proxy_version="$(tr -d '\n' <"$rootdir/.proxy-version")"
+if [ -z "$old_proxy_version" ]; then
+    echo "missing .proxy-version" >&2
+    exit 1;
+fi
+
+# Checkout the linkerd2-proxy repo to resolve the new proxy version to a SHA
+# and obtain the commit log since the prior version.
+tmp=$(mktemp -d -t l2-proxy.XXX)
+git clone --depth=100 https://github.com/linkerd/linkerd2-proxy "$tmp"
```
hawkw: It seems pretty unlikely, but i suppose this _could_ fail weirdly if the new revision is over 100 commits old? I doubt that will be a common occurance though.

bin/git-commit-proxy-version
```diff
@@ -0,0 +1,44 @@
...
+cd "$tmp"
+
+sha="$(git rev-parse "${new_proxy_revision}")"
+new_proxy_version="${sha:0:7}"
+
+echo "proxy: Update proxy to ${new_proxy_revision}" >"$tmp/commit.template"
+echo "" >>"$tmp/commit.template"
+
+git log --pretty=oneline --abbrev-commit --reverse "${old_proxy_version}..${new_proxy_version}" | \
+sed -E -e 's,^,* linkerd/linker2-proxy#,' \
```
hawkw: Typo:
```suggestion
sed -E -e 's,^,* linkerd/linkerd2-proxy#,' \
```

bin/git-commit-proxy-version
olix0r: Yeah... don't think we need to worry about being particularly robust. We could try to rev-parse the old sha to be sure...

PR #3078, opened 2019-07-11
Update ServiceProfile CRD to version v1alpha2 and remove validation

adleong:
The openAPIV3Schema validation in the ServiceProfiles CRD is very limited in what it can validate and is obviated by more sophisticated validation done by the validating admission controller.  Therefore, we would like to remove the openAPIV3Schema validation to reduce the size and complexity of the CRD object.

To do so, we must also bump the version of the ServiceProfile custom resource from v1alpha1 to v1alpha2.  This ensures that when the controller is upgraded, it will attempt to watch the v1alpha2 resource.  If it cannot (because, for example, the controller pod started before the ServiceProfile CRD was updated and therefore the v1alpha2 version does not exist) then it will go into a crash loop backoff until it can.  This essentially means that the controller will wait for the CRD to be upgraded to include v1alpha2 before it will start.  

Bumping the version is necessary because if we did not, it would be possible for the controller to start before the CRD is updated (removing the validation).  In this case, when the CRD is edited, the controller will lose its list watch on ServiceProfiles and will stop getting updates.

The upgrade process will look like this: consider a cluster running stable-2.4 with an existing v1alpha1 service profile:

```
$ linkerd version
Client version: stable-2.4.0
Server version: stable-2.4.0
$ kubectl get sp/authors.default.svc.cluster.local -o yaml
...
$ linkerd routes svc/authors
ROUTE                       SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
DELETE /authors/{id}.json   authors   100.00%   0.6rps          45ms          50ms          50ms
...
```

Now we upgrade Linkerd:

```
$ bin/linkerd upgrade | kubectl apply -f -
...
$ bin/linkerd version
Client version: dev-7062b915-alex
Server version: dev-7062b915-alex
```

Linkerd can still read/watch the v1alpha1 service profiles:

```
$ kubectl get sp/authors.default.svc.cluster.local -o yaml
...
$ bin/linkerd routes svc/authors
ROUTE                       SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
DELETE /authors/{id}.json   authors   100.00%   0.5rps          47ms          91ms          98ms
...
```

But any new service profiles will be created under v1alpha2:

```
$
curl https://run.linkerd.io/booksapp/books.swagger | bin/linkerd profile --open-api - books | k apply -f -
...
$ k get sp/books.default.svc.cluster.local -o yaml
...
$ bin/linkerd routes svc/books
ROUTE                     SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
DELETE /books/{id}.json     books   100.00%   1.1rps          18ms          29ms          30ms
...
```


controller/sp-validator/tmpl/validating_webhook_configuration.go
```diff
@@ -0,0 +1,22 @@
+package tmpl
+
+// ValidatingWebhookConfigurationSpec provides a template for a
+// ValidatingWebhookConfiguration.
+var ValidatingWebhookConfigurationSpec = `
```
siggy: i don't see this used?

adleong:
Good call on `check`.  https://github.com/linkerd/website/pull/435

PR #3085, opened 2019-07-15
A Slightly More Restrictive PSP

### LINKED ISSUE #3074: Custom PSP for Linkerd

It is great that linkerd 2.4.0 introduces a custom built PSP for its components. My cluster is even a little more restrictive than what is provided, though. For example, my cluster requires that `runAsUser` is `MustRunAsNonRoot` and that the fsGroups/supplementalGroups are limited to UIDs between 10001 and 65535. When I make these changes to the linkerd custom provided PSP, linkerd seems to work fine.

I'd like to be able to point the linkerd components at my own cluster's PSP without creating a new one. I understand that the PSP I use would need to meet the requirements for linkerd.

FWIW, I use linkerd with the CNI configuration only and haven't tested the changes above with a non-CNI setup.

Perhaps an install flag `--psp-name` or similar where I could input my custom PSP name. This would also be used during `linkerd check` to ensure it was restrictive/permissive enough.

The cool thing about the linkerd install process is that there is always the ability to edit the `.yaml` that is output by the `linkerd install` command where I can delete the provided PSP and update the Roles and RoleBindings.

Alternatively, changing the custom provided PSP to what I need is always an option. :)

codeman9:
Resolves #3074 with a slightly more restrictive PSP. Needs to be tested on a non-CNI cluster.


ihcsim:
Per our _slack_ convo, we will need to update the PSP Helm template to account for the non-CNI setup, since the `proxy-init` init container is run as the `root` user. The following `diff` works a PSP-enabled, non-CNI GKE cluster. Following this change, you can use `go test ./cli/cmd/... -update` to update the failing unit tests.

```diff
diff --git a/chart/templates/psp.yaml b/chart/templates/psp.yaml
index feaf64fe..3d83e7b9 100644
--- a/chart/templates/psp.yaml
+++ b/chart/templates/psp.yaml
@@ -26,17 +26,31 @@ spec:
   seLinux:
     rule: RunAsAny
   runAsUser:
+    {{- if .NoInitContainer }}
     rule: MustRunAsNonRoot
+    {{- else }}
+    rule: RunAsAny
+    {{- end }}
   supplementalGroups:
     rule: MustRunAs
     ranges:
+    {{- if .NoInitContainer }}
     - min: 10001
       max: 65535
+    {{- else }}
+    - min: 1
+      max: 65535
+    {{- end }}
   fsGroup:
     rule: MustRunAs
     ranges:
+    {{- if .NoInitContainer }}
     - min: 10001
       max: 65535
+    {{- else }}
+    - min: 1
+      max: 65535
+    {{- end }}
   volumes:
   - configMap
```

PR #3087, opened 2019-07-15
Read dstOverrides directly from service profiles

adleong:
Allow `dstOverrides` to be specified directly on a service profile as an alternative to using the SMI traffic split API.  This traffic split object:

```
kind: TrafficSplit
metadata:
  name: books-error-inject
spec:
  service: books
  backends:
  - service: books
    weight: 80m
  - service: error-injector
    weight: 20m
```

is equivalent to this service profile:

```
kind: ServiceProfile
metadata:
  name: books.default.svc.cluster.local
  namespace: default
spec:
  dstOverrides:
  - service: error-injector
    weight: 20m
  - service: books
    weight: 80m
```

If a service has both a traffic split and a service profile that defines dstOverrides, the dstOverrides in the service profile are used and the traffic split is ignored.

### DIFF
--- a/controller/api/destination/profile_translator.go
+++ b/controller/api/destination/profile_translator.go
@@ -97,7 +97,7 @@ func toDstOverrides(dsts []*sp.WeightedDst) []*pb.WeightedDst {
 	pbDsts := []*pb.WeightedDst{}
 	for _, dst := range dsts {
 		pbDst := &pb.WeightedDst{
-			Authority: dst.Authority,
+			Authority: dst.Service,
 			// Weights are expressed in decimillis: 10_000 represents 100%
 			Weight: uint32(dst.Weight.MilliValue() * millisPerDecimilli),
 		}
--- a/controller/api/destination/traffic_split_adaptor.go
+++ b/controller/api/destination/traffic_split_adaptor.go
@@ -2,10 +2,13 @@ package destination
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 
 	ts "github.com/deislabs/smi-sdk-go/pkg/apis/split/v1alpha1"
 	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
 	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha1"
+	log "github.com/sirupsen/logrus"
 )
 
 // trafficSplitAdaptor merges traffic splits into service profiles, encoding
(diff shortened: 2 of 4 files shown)


adleong:
@alpeb I'll update the template example to include a dstOverride.  Did you have any other changes in mind?  The validation uses the struct as a guide for unmarshalling so I don't think the field name needs to be accounted for.

I'm open to friendlier names.  We can pick whatever we think works best for Linkerd.  No need to necessarily match the SMI terminology here.  `split` feels a bit vague to me.

alpeb:
@adleong you're right, that property is already accounted for. Persisting the service profile example from this issue returned a validation error and I assumed that was the cause, but the problem is just that `spec.routes` is missing.

alpeb:
So far this has tested well for me. I also tried edge cases like starting with just an SP without a TS, or having both an SP and a TS and deleting the SP for the TS to be picked up :+1:

adleong:
@grampelberg what do you think about this change?  From an internal architecture point of view it feels very natural because SMI traffic splits get converted into destination profiles under the hood anyway.

But it does increase the API surface area and makes it so there are two different ways of creating a traffic split.  It also may complicate future work related to a `linkerd get trafficsplit` command since traffic splits configured directly on the service profile would not have a corresponding SMI traffic split resource.

grampelberg:
Originally, I had assumed that we would have a simple controller which takes TrafficSplit and rewrites to ServiceProfiles. I'm not a big fan of having two different places to configure it. What happens when someone changes the ServiceProfile setting but leaves TrafficSplit the same?

adleong:
Service profile take precedence over traffic splits.

grampelberg:
I don't particularly like adding extra config knobs that we and users need to reason about. I don't have any concrete issues though, from the product perspective adding a SMI-free version doesn't really get us anything ... but it also doesn't really hurt either.

tldr, 🤷‍♂

adleong:
Aw, I wanted you to tell me no so I could blame someone else for taking away my candy 😜 

But I share your misgivings.  In general I'm a little unclear on how tightly coupled Linkerd and SMI should be and how deeply integrated SMI primitive should be in Linkerd as opposed to "native" Linkerd primitives like the service profile.  The fact that, for example, retries and traffic splits get configured in two different CRDs feels weird.

I'm going to close this for now until our vision about this comes into better focus.  We should keep this solution in mind for the future, though.

PR #3096, opened 2019-07-17
Review tap hardening RFC (drafting for review approval)

kleimkuhler:
# Summary

The goal of the Tap Hardening project is to restrict the set of users by
integrating with native Kubernetes RBAC. The current proposal is for the Tap
Service to become an [Extension API server](https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/). This will allow requests to be
both authorized and authenticated through the Kubernetes API. The TapByResource
route currently served by the public API will be served by the APIService object
on the now separate Tap Pod.

The data plane proxies will also begin authenticating and authorizing Tap
requests. The proxy’s Tap server will require a TLS connection and the client
must be the Tap Controller. All other authenticated clients will be rejected.

# Motivation

The Tap feature is open to any user that has access to the cluster. Tap requests
made through the CLI and the Web Deployment are neither authenticated or
authorized. The TapByResource route is served by the public API and will start
Tap connections to any of the resources requested.

When the Linkerd proxies for requested resources receive tap requests, the Tap
server also does not authenticate or authorize the requesting client. If it can
open a connection over TLS it will, but it is not required.

# Details

## Control Plane

The Tap Controller becoming an Extension API server will allow requests to be
authenticated and authorized through an [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/). The Tap APIService
will claim the TapByResource route in the Kubernetes API. This will allow the
aggregation layer to proxy anything sent to that API path to the Tap extension
apiserver on the Tap Pod. Tap requests made by the CLI and Web Deployment will
therefore be proxied through the aggregation layer to the APIService.

The Kubernetes API will authenticate the requesting user and authorize the
rights for the TapByResource API path via RBAC. The Tap extension apiserver will
authenticate the proxied request from the Kubernetes apiserver and then execute
the request.

## Data Plane

The proxy's Tap server will begin requiring connections over TLS and authorizing
requests made only from the Tap Controller. Changes will be required both on
outbound and inbound tap requests.

### Outbound Tap Requests

On outbound tap requests made by the Tap Controller, requests will continue
being made directly to the resource's Pod IP. A new header will be added to
these requests that will require connecting to a specified server identity. The
presence of this header will allow the proxy to make a TLS connection to the
requested resource. It will also verify that the requested resource is in-fact
the identity expected.

In order to get the identity of requested resource(s), the Tap Controller will
make a request to the Kubernetes API for the expected identities of the
resources' Pod IPs. It will use those identities as header values in the tap
requests.

### Inbound Tap Requests

Inbound Tap requests are received by the Tap server running in the resource's
proxy container. When proxies are injected, a new environment variable will be
added that has the expected identity value of the Tap Controller. The Tap server
will begin authorizing clients--accepting requests from the expected Tap Service
identity and rejecting all others.

Once the Tap Controller is making outbound requests with the the required
identity header and the Tap server is able to validate incoming requests are
made from the Tap Controller, the Tap server will stop accepting plaintext
connections.

# Drawbacks

The proposed design is an overall benefit for securing Tap on a cluster. It will
introduce additional resources and permissions for properly authenticating
clients through the CLI and Web Deployment. If a user is authorized to tap
cluster resources it should not change that experience.

# Alternatives (or lack thereof)

- The proposed design for the control plane uses Kubernetes security primitives
  which is a big benefit. Setting up an aggregation layer that proxies requests
  through the Kubernetes API allows users to be authenticated through RBAC,
  rather than implementing a separate Linkerd ACL and associated RBAC.

# Unresolved Questions

- Should the proxy Tap server continue serving protobuf over HTTP or move to
  only serving HTTP?
- When should the code path on the public API for TapByResource be removed?
- How will the new resources required for authorizing request made to the Tap
  Service affect install/upgrade/uinstall lifecycle?
- What is the desired behavior in HA?

# Possible Sub-projects

- [x] The Tap Service should be moved out of the controller Pod into it's own in
  preparation for becoming an extension apiserver
- [ ] Communication between the Tap Controller and Tap servers is secured
    * Tap Controller queries the Kubernetes API for the identities of the
      requested resources
    * The returned identities are used as header values in the requests made by
      the Tap Controller
    * Proxies are injected with the identity of the Tap Controller as an
      environment variable
    * The Tap servers on the proxies validate TLS connections are made only by
      the Tap Controller
- [ ] The Tap Controller becomes an extension apiserver
    * The Tap Controller generates a CA to include in the APIService object
    * The Tap Controller serves HTTPS which is a requirement for the aggregator
      layer
    * CLI and Web Deployment talk to the Kubernetes API server instead of the
      public API
    * RBAC lifecycle works for a user in the CLI and the Web ServiceAccount
        * The Web ServiceAccount is authorized through RBAC
        * Users through the CLI are authorized through RBAC
- [ ] Cleanup existing Tap infrastructure that would be deprecated
    * Remove GRPC server from the Tap Service
    * Remove existing public API Tap infrastructure
    * Update documentation

### DIFF
--- a/rfc/tap-hardening.md
+++ b/rfc/tap-hardening.md
@@ -1,15 +1,15 @@
 # Summary
 
-The goal of the Tap Hardening project is to restrict the set of clients that can
-tap resources on the cluster. The current proposal is for the Tap Service to
-become an Extension API server. This will allow requests to be both authorized
-and authenticated through the Kubernetes API. The TapByResource route currently
-served by the public API will be served by the APIService object on the now
-separate Tap Pod.
+The goal of the Tap Hardening project is to restrict the set of users by
+integrating with native Kubernetes RBAC. The current proposal is for the Tap
+Service to become an [Extension API server](https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/).
+This will allow requests to be both authorized and authenticated through the
+Kubernetes API. The TapByResource route currently served by the public API will
+be served by the APIService object on the now separate Tap Pod.
 
 The data plane proxies will also begin authenticating and authorizing Tap
 requests. The proxy’s Tap server will require a TLS connection and the client
-must be the Tap Service. All other authenticated clients will be rejected.
+must be the Tap Controller. All other authenticated clients will be rejected.
 
 # Motivation
 
@@ -24,12 +24,9 @@ open a connection over TLS it will, but it is not required.
 
 # Details
 
-The control plane component of Tap will be referred to as the Tap Controller.
-The data plane component of Tap will be referred to as the Tap server.
-
 ## Control Plane
 
-The Tap Service becoming an Extension API server will allow requests to be
+The Tap Controller becoming an Extension API server will allow requests to be
 authenticated and authorized through an [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/).
 The Tap APIService will claim the TapByResource route in the Kubernetes API.
 This will allow the aggregation layer to proxy anything sent to that API path to
@@ -39,39 +36,41 @@ APIService.
 
 The Kubernetes API will authenticate the requesting user and authorize the
 rights for the TapByResource API path via RBAC. The Tap extension apiserver will
-authenticate the proxied request from the Kubernetes apiserver and then
-authorize the request from the original user.
+authenticate the proxied request from the Kubernetes apiserver and then execute
+the request.
 
 ## Data Plane
 
 The proxy's Tap server will begin requiring connections over TLS and authorizing
-requests made only from the Tap Service. Changes will be required both on
+requests made only from the Tap Controller. Changes will be required both on
 outbound and inbound tap requests.
 
 ### Outbound Tap Requests
 
-On outbound tap requests made by the Tap Service, requests will continue being
-made directly to the resource's Pod IP. A new header will be added to these
-requests that will require connecting to a specified server identity. The
+On outbound tap requests made by the Tap Controller, requests will continue
+being made directly to the resource's Pod IP. A new header will be added to
+these requests that will require connecting to a specified server identity. The
 presence of this header will allow the proxy to make a TLS connection to the
 requested resource. It will also verify that the requested resource is in-fact
 the identity expected.
 
-In order to get the identity of requested resource(s), the Tap Service will make
-a request to the Kubernetes API for the expected identities of the resources'
-Pod IPs. It will use those identities as header values in the tap requests.
+In order to get the identity of requested resource(s), the Tap Controller will
+make a request to the Kubernetes API for the expected identities of the
+resources' Pod IPs. It will use those identities as header values in the tap
+requests.
 
 ### Inbound Tap Requests
 
 Inbound Tap requests are received by the Tap server running in the resource's
 proxy container. When proxies are injected, a new environment variable will be
-added that has the expected identity value of the Tap Service. The Tap server
+added that has the expected identity value of the Tap Controller. The Tap server
 will begin authorizing clients--accepting requests from the expected Tap Service
 identity and rejecting all others.
 
-Once the Tap Service is making outbound requests with the the required identity
-header and the Tap server is able to validate incoming requests are made from
-the Tap Service, the Tap server will stop accepting plaintext connections.
+Once the Tap Controller is making outbound requests with the the required
+identity header and the Tap server is able to validate incoming requests are
+made from the Tap Controller, the Tap server will stop accepting plaintext
+connections.
 
 # Drawbacks
 
@@ -100,18 +99,18 @@ cluster resources it should not change that experience.
 
 - [x] The Tap Service should be moved out of the controller Pod into it's own in
   preparation for becoming an extension apiserver
-- [ ] Communication between the Tap Service and Tap servers is secured
-    * Tap Service queries the Kubernetes API for the identities of the requested
-      resources
+- [ ] Communication between the Tap Controller and Tap servers is secured
+    * Tap Controller queries the Kubernetes API for the identities of the
+      requested resources
     * The returned identities are used as header values in the requests made by
-      the Tap Service
-    * Proxies are injected with the identity of the Tap Service as an
+      the Tap Controller
+    * Proxies are injected with the identity of the Tap Controller as an
       environment variable
     * The Tap servers on the proxies validate TLS connections are made only by
-      the Tap Service
-- [ ] The Tap Service becomes an extension apiserver
-    * The Tap Service generates a CA to include in the APIService object
-    * The Tap Service serves HTTPS which is a requirement for the aggregator
+      the Tap Controller
+- [ ] The Tap Controller becomes an extension apiserver
+    * The Tap Controller generates a CA to include in the APIService object
+    * The Tap Controller serves HTTPS which is a requirement for the aggregator
       layer
     * CLI and Web Deployment talk to the Kubernetes API server instead of the
       public API
--- a/rfc/tap-hardening.md
+++ b/rfc/tap-hardening.md
@@ -1,0 +1,99 @@
+# Summary
+
+The goal of the Tap Hardening project is to restrict the set of clients that can
+tap resources on the cluster. The current proposal is for the Tap Service to
+become an Extension API server. This will allow requests to be both authorized
+and authenticated through the Kubernetes API. The TapByResource route currently
+served by the public API will be served by the APIService object on the now
+separate Tap Pod.
+
+The data plane proxies will also begin authenticating and authorizing Tap
+requests. The proxy’s Tap server will require a TLS connection and the client
+must be the Tap Service. All other authenticated clients will be rejected.
+
+# Motivation
+
+The Tap feature is open to any user that has access to the cluster. Tap requests
+made through the CLI and the Web Deployment are neither authenticated or
+authorized. The TapByResource route is served by the public API and will start
+Tap connections to any of the resources requested.
+
+When the Linkerd proxies for requested resources receive tap requests, the Tap
+server also does not authenticate or authorize the requesting client. If it can
+open a connection over TLS it will, but it is not required.
+
+# Details
+
+The control plane component of Tap will be referred to as the Tap Controller.
+The data plane component of Tap will be referred to as the Tap server.
+
+## Control Plane
+
+The Tap Service becoming an Extension API server will allow requests to be
+authenticated and authorized through an [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/).
+The Tap APIService will claim the TapByResource route in the Kubernetes API.
+This will allow the aggregation layer to proxy anything sent to that API path to
+the Tap extension apiserver on the Tap Pod. Tap requests made by the CLI and Web
+Deployment will therefore be proxied through the aggregation layer to the
+APIService.
+
+The Kubernetes API will authenticate the requesting user and authorize the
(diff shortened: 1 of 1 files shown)


rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
+# Summary
+
+The goal of the Tap Hardening project is to restrict the set of clients that can
```
grampelberg: Maybe pedantic, but `restrict the set of users by integrating with native Kubernetes RBAC`.

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+tap resources on the cluster. The current proposal is for the Tap Service to
+become an Extension API server. This will allow requests to be both authorized
```
grampelberg: Might want to link to a definition of this.

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+The Tap APIService will claim the TapByResource route in the Kubernetes API.
+This will allow the aggregation layer to proxy anything sent to that API path to
+the Tap extension apiserver on the Tap Pod. Tap requests made by the CLI and Web
+Deployment will therefore be proxied through the aggregation layer to the
+APIService.
+
+The Kubernetes API will authenticate the requesting user and authorize the
+rights for the TapByResource API path via RBAC. The Tap extension apiserver will
+authenticate the proxied request from the Kubernetes apiserver and then
+authorize the request from the original user.
```
grampelberg: Technically, the tap controller won’t be doing any authz right?

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+The Tap Service becoming an Extension API server will allow requests to be
+authenticated and authorized through an [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/).
+The Tap APIService will claim the TapByResource route in the Kubernetes API.
+This will allow the aggregation layer to proxy anything sent to that API path to
+the Tap extension apiserver on the Tap Pod. Tap requests made by the CLI and Web
+Deployment will therefore be proxied through the aggregation layer to the
+APIService.
+
+The Kubernetes API will authenticate the requesting user and authorize the
+rights for the TapByResource API path via RBAC. The Tap extension apiserver will
```
grampelberg: I’d love to see an example `Role`.

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+open a connection over TLS it will, but it is not required.
+
+# Details
+
+The control plane component of Tap will be referred to as the Tap Controller.
+The data plane component of Tap will be referred to as the Tap server.
+
+## Control Plane
+
+The Tap Service becoming an Extension API server will allow requests to be
```
grampelberg: Tap Controller?

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+outbound and inbound tap requests.
+
+### Outbound Tap Requests
+
+On outbound tap requests made by the Tap Service, requests will continue being
+made directly to the resource's Pod IP. A new header will be added to these
+requests that will require connecting to a specified server identity. The
+presence of this header will allow the proxy to make a TLS connection to the
+requested resource. It will also verify that the requested resource is in-fact
+the identity expected.
```
grampelberg: The controller will always know the identity before making the request (pod-service-account) and the server will always know the identity of the requestor (linkerd-tap). Why does there need to be a header?

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+proxy container. When proxies are injected, a new environment variable will be
+added that has the expected identity value of the Tap Service. The Tap server
+will begin authorizing clients--accepting requests from the expected Tap Service
+identity and rejecting all others.
+
+Once the Tap Service is making outbound requests with the the required identity
+header and the Tap server is able to validate incoming requests are made from
+the Tap Service, the Tap server will stop accepting plaintext connections.
+
+# Drawbacks
```
grampelberg: Do we have any idea what happens by default? Will `cluster-admin` users just get permission or do we need to ship the control plane with a new set of bindings for everyone?

rfc/tap-hardening.md
```diff
@@ -0,0 +1,99 @@
...
+cluster resources it should not change that experience.
+
+# Alternatives (or lack thereof)
+
+- The proposed design for the control plane uses Kubernetes security primitives
+  which is a big benefit. Setting up an aggregation layer that proxies requests
+  through the Kubernetes API allows users to be authenticated through RBAC,
+  rather than implementing a separate Linkerd ACL and associated RBAC.
+
+# Unresolved questions
```
grampelberg: How will streaming work? Will the dashboard need websockets?

grampelberg: How will it work from the dashboard? Is delegation out of scope for now?

rfc/tap-hardening.md
kleimkuhler: I agree a good thing to be explicit about when the design is focused on using Kubernetes primitives!

rfc/tap-hardening.md
kleimkuhler: I was explicit about this because the documentation leads me to believe the extension apiserver still does some authorization of the request proxied from the Kubernetes apiserver: https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#authentication-flow

> 4. Extension apiserver: authorize the request from the original user

I removed that part, but I can get a better idea of what exactly happens there.

rfc/tap-hardening.md
kleimkuhler: This is going to be a follow-up for me.

rfc/tap-hardening.md
kleimkuhler: The proxy will not attempt to do service discovery on outbound requests made to IPs. The role of the header is to both to set _and_ require the identity of the server.

rfc/tap-hardening.md
grampelberg: In my testing of smi-metrics, the only authz I did was validating the incoming connection from the api server. I was doing some pretty basic RBAC though.

rfc/tap-hardening.md
kleimkuhler: Without the header, the proxy will continue to making plaintext connections to the requested resource's proxy.

rfc/tap-hardening.md
kleimkuhler: I've added this as a question. I'm not sure on an answer to this right now.

rfc/tap-hardening.md
kleimkuhler: I've added this as an unresolved question.

rfc/tap-hardening.md
grampelberg: As I understand it, the API format has implications for what kinds of RBAC `Role`s can be used. We’ll need to make sure the API served by the tap controller matches up for our goals. Ideally, I’d like to allows tap requests for specific pods, deployments and such.

rfc/tap-hardening.md
grampelberg: Oh yeah! Thank you. Would you mind adding that to the doc (the IP based piece is important).

kleimkuhler: Sure!

rfc/tap-hardening.md
```diff
@@ -0,0 +1,125 @@
...
+server also does not authenticate or authorize the requesting client. If it can
+open a connection over TLS it will, but it is not required.
+
+# Details
+
+## Control Plane
+
+The Tap Controller becoming an Extension API server will allow requests to be
+authenticated and authorized through an [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/). The Tap APIService
+will claim the TapByResource route in the Kubernetes API. This will allow the
```
adleong: What is the route/path, exactly?

rfc/tap-hardening.md
```diff
@@ -0,0 +1,125 @@
...
+aggregation layer to proxy anything sent to that API path to the Tap extension
```
adleong: The creation of the APIService resource should probably follow the same pattern as the webhooks use which is described in https://github.com/linkerd/linkerd2/pull/2806

i.e. The cert, key, and APIService resource are generated at install time and the cert and key are loaded by the tap controller when it starts up.

adleong:
I suggest we use an API structure as follows:

The URL would be: `GET /apis/tap.linkerd.io/v1alpha1/watch/namespaces/<NS>/<RESOURCE>/<NAME>/tap`
For example: `GET /apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/pod/linkerd-web-844dbd7fb4-9qk9t/tap` to tap an individual pod or `/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployment/linkerd-controller/tap` to tap a deployment.

This will cause the aggregator to check that the user has RBAC privileges to "watch" the tap subresource of the specified resource.  Here is an example of a Role which grants permission to tap all pods:

```
kind: ClusterRole
metadata:
  name: tap
rules:
- apiGroups:
  - tap.linkerd.io
  resources:
  - deployment/tap
  verbs:
  - watch
```

And here's one that allows the user to tap a specific deployment:

```
kind: ClusterRole
metadata:
  name: tap
rules:
- apiGroups:
  - tap.linkerd.io
  resources:
  - deployment/tap
  resourceNames:
  - linkerd-grafana
  verbs:
  - watch
```

The body of the request would be the protobuf encoded TapByResource request.  It is the responsibility of the tap controller to verify that the resource in the URL matches the requested resource in the protobuf request message.  If these do not match, the tap controller must reject the request, otherwise a user could initiate a tap for a resource that they do not have RBAC permissions for.

Here is an example of response from the aggregator when you attempt to make a request that you do not have RBAC permissions for:

```
$ curl -v localhost:8001/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployment/linkerd-controller/tap
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8001 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8001 (#0)
> GET /apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployment/linkerd-controller/tap HTTP/1.1
> Host: localhost:8001
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 403 Forbidden
< Audit-Id: f8224424-7915-4aac-a758-715694b8cdad
< Content-Length: 450
< Content-Type: application/json
< Date: Fri, 19 Jul 2019 21:41:48 GMT
< X-Content-Type-Options: nosniff
<
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "deployment.tap.linkerd.io \"linkerd-controller\" is forbidden: User \"obama@buoyant.io\" cannot watch resource \"deployment/tap\" in API group \"tap.linkerd.io\" in the namespace \"linkerd\"",
  "reason": "Forbidden",
  "details": {
    "name": "linkerd-controller",
    "group": "tap.linkerd.io",
    "kind": "deployment"
  },
  "code": 403
* Connection #0 to host localhost left intact
}⏎
```

grampelberg:
Sounds good to me! I'm questioning whether having a `tap` subresource is required, but I guess it is pretty similar to port forward.

kleimkuhler:
Work tracked in https://github.com/orgs/linkerd/projects/8 and closed with the release of [stable-2.5.0](https://github.com/linkerd/linkerd2/releases/tag/stable-2.5.0)

PR #3101, opened 2019-07-17
Release notes for Edge-19.7.4

alpeb:
* CLI
  * Made the `linkerd routes` command traffic-split aware; consider resources
    which are members of a leaf service with non-zero weight to be members of
    the apex service for the purpose of determining which routes to display
  * Added pod status to the output of the `linkerd stat`command (thanks
    @jonathanbeber!)
  * Fixed `linkerd stat` wrong "meshed" count when resources share the same
    label selector for pods (thanks @jonathanbeber!)
  * Improved output of the `linkerd edges` command and added a new `-o wide`
    flag to show the identity of the client and server if known
  * Added a new check to the `linkerd check --pre` command validating that if
    PSP is enabled, the NET_RAW capability is available

* Proxy
  * Improved performance by using a constant-time load balancer
  * Added a new `/proxy-log-level` endpoint to update the log level at runtime

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * CLI
   * Made the `linkerd routes` command traffic-split aware
+  * Fixed bug the `linkerd upgrade config` command that was causing it to crash
   * Added pod status to the output of the `linkerd stat`command (thanks
     @jonathanbeber!)
   * Fixed incorrect "meshed" count in `linkerd stat` when resources share the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,7 @@
 ## edge-19.7.4
 
 * CLI
-  * Made the `linkerd routes` command traffic-split aware; it now considers
-    resources which are members of a leaf service with non-zero weight to be
-    members of the apex service for the purpose of determining which routes to
-    display
+  * Made the `linkerd routes` command traffic-split aware
   * Added pod status to the output of the `linkerd stat`command (thanks
     @jonathanbeber!)
   * Fixed `linkerd stat` wrong "meshed" count when resources share the same
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 ## edge-19.7.4
 
 * CLI
-  * Made the `linkerd routes` command traffic-split aware; it now consider
+  * Made the `linkerd routes` command traffic-split aware; it now considers
     resources which are members of a leaf service with non-zero weight to be
     members of the apex service for the purpose of determining which routes to
     display
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,7 +14,7 @@
     PSP is enabled, the NET_RAW capability is available
 
 * Proxy
-  * Improve performance by using a constant-time load balancer
+  * Improved performance by using a constant-time load balancer
   * Added a new `/proxy-log-level` endpoint to update the log level at runtime
 
 ## stable-2.4.0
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,22 @@
...
+  * Fixed `linkerd stat` wrong "meshed" count when resources share the same
+    label selector for pods (thanks @jonathanbeber!)
+  * Improved output of the `linkerd edges` command and added a new `-o wide`
+    flag to show the identity of the client and server if known
+  * Added a new check to the `linkerd check --pre` command validating that if
+    PSP is enabled, the NET_RAW capability is available
+
+* Proxy
+  * Improve performance by using a constant-time load balancer
+  * Added a new `/proxy-log-level` endpoint to update the log level at runtime
```
hawkw: Hmm, it seems like the _behaviour_ of the endpoint maybe ought to be documented somewhere (though the changelog is probably the wrong place)

CHANGES.md
```diff
@@ -1,3 +1,22 @@
+## edge-19.7.4
+
+* CLI
+  * Made the `linkerd routes` command traffic-split aware; consider resources
```
scottcarol: is "consider resources" referring to the behavior of the command? I might say "it considers resources" (as long as that's correct).

CHANGES.md
```diff
@@ -1,3 +1,22 @@
...
+    which are members of a leaf service with non-zero weight to be members of
+    the apex service for the purpose of determining which routes to display
+  * Added pod status to the output of the `linkerd stat`command (thanks
+    @jonathanbeber!)
+  * Fixed `linkerd stat` wrong "meshed" count when resources share the same
+    label selector for pods (thanks @jonathanbeber!)
+  * Improved output of the `linkerd edges` command and added a new `-o wide`
```
scottcarol: to be clear, the previous default behavior was to show the identity and not namespaces. now the namespace of the source and the destination are shown by default, and you can use `-o wide` to show identity. I might rephrase as:

```
Added namespace information to the `linkerd edges` command output and a new `-o wide` flag that shows the identity of the client and server if known
```

CHANGES.md
alpeb: Good catch :+1:

CHANGES.md
alpeb: Can you create a PR in the website repo to track that?

CHANGES.md
wmorgan: What about just eliding that part of the sentence? Feels overly implementation-detail-y for release notes.

CHANGES.md
```diff
@@ -1,3 +1,23 @@
+## edge-19.7.4
+
+* CLI
+  * Made the `linkerd routes` command traffic-split aware; it now considers
```
wmorgan: TIOLI: I would just remove everything from ";" afterwards since it feels too implementation-specific-y

CHANGES.md
```diff
@@ -1,3 +1,23 @@
...
+    resources which are members of a leaf service with non-zero weight to be
+    members of the apex service for the purpose of determining which routes to
+    display
+  * Added pod status to the output of the `linkerd stat`command (thanks
+    @jonathanbeber!)
+  * Fixed `linkerd stat` wrong "meshed" count when resources share the same
```
kleimkuhler: Wording nit: "Fixed wrong "meshed" count in `linkerd stat` when ..."

kleimkuhler: "incorrect" maybe? More just moving `linkerd stat` after the thing being fixed.

PR #2950, opened 2019-06-17
Allow custom cluster domain in controller components

arminbuerkle:
Splits #2930 into smaller components: Part 1, controller components.

While it's still a lot of changes, most of them are auto generated (proto, .golden) or updating tests so a different domain than `cluster.local` is used.

---

I added the `cluster.local` default domain to the `cli/cmd/` pkg, so the `.golden` files will directly add it to their global configs. Otherwise they'd get initialised with an empty string: `{..., "clusterDomain":""}`

PR #3113, opened 2019-07-20
Add support for stateful sets

### LINKED ISSUE #2266: Can't Reach StatefulSet Pods Via Stable Network ID
### What is the issue?
A meshed client isn't able to reach a meshed statefulset pod via the pod's [stable network ID](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id). This works if the client isn't meshed.

### How can it be reproduced?
Deploy a meshed statefulset nginx
```
cat <<EOF | linkerd inject - | kubectl apply -f -
kind: Service 
apiVersion: v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  clusterIP: None
  selector:
    app: nginx
  ports:
  - name: http
    port: 80
    targetPort: http

---
kind: StatefulSet
apiVersion: apps/v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 3
  serviceName: nginx
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx
        ports:
        - containerPort: 80
          name: http
EOF
```

Use a `curl` pod to reach nginx:
```
$ kubectl run curl --image=appropriate/curl --restart=Never --command -- sleep 3600

# this works
$ kubectl exec curl -- curl --silent nginx-0.nginx.default.svc.cluster.local
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
...
```

Now meshed the `curl` pod and notice that it can't reach the nginx pod:
```
$ kubectl run curl --image=appropriate/curl --restart=Never --dry-run -o yaml --command -- sleep 3600 | linkerd inject - | kubectl apply -f -

# stuck
$ kubectl  exec curl -c curl -- curl --silent nginx-0.nginx.default.svc.cluster.local

# this works
$ kubectl exec curl -c curl -- curl --silent nginx.default
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
...
```

### Logs, error output, etc
Proxy logs from the `curl` client:
```
WARN admin={bg=resolver} linkerd2_proxy::control::destination::background::destination_set Destination.Get stream errored for NameAddr { name: "nginx-0.nginx.default.svc.cluster.local", port: 80 }: Grpc(Status { code: Unknown, message: "resolver [&{k8sDNSZoneLabels:[] controllerNamespace:linkerd endpointsWatcher:0xc4205e9f20 profileWatcher:0xc4203f2ae0}] found error resolving host [nginx-0.nginx.default.svc.cluster.local] port [80]: not a service: nginx-0.nginx.default.svc.cluster.local" })
WARN admin={bg=resolver} linkerd2_proxy::control::destination::background::destination_set Destination.Get stream errored for NameAddr { name: "nginx-0.nginx.default.svc.cluster.local", port: 80 }: Grpc(Status { code: Unknown, message: "resolver [&{k8sDNSZoneLabels:[] controllerNamespace:linkerd endpointsWatcher:0xc4205e9f20 profileWatcher:0xc4203f2ae0}] found error resolving host [nginx-0.nginx.default.svc.cluster.local] port [80]: not a service: nginx-0.nginx.default.svc.cluster.local" })
```

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ controller pod is running
√ can initialize the client

linkerd-api
-----------

linkerd-service-profile
-----------------------
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date
```

### Environment

- Kubernetes Version: v1.13.2
- Cluster Environment: Minikube
- Host OS: Ubuntu 16.04
- Linkerd version: edge-19.2.2

adleong:
We add support for looking up individual pods in a stateful set with the destination service.  This allows Linkerd to correctly proxy requests which address individual pods.  The authority structure for such a request is `<pod-name>.<service>.<namespace>.svc.cluster.local:<port>`.

Fixes #2266 


controller/api/destination/watcher/k8s.go
```diff
@@ -50,26 +50,39 @@ func getHostAndPort(authority string) (string, Port, error) {
...
 	}
 	domains := strings.Split(host, ".")
-	// S.N.svc.cluster.local
-	if len(domains) != 5 {
-		return ServiceID{}, 0, fmt.Errorf("Invalid k8s service %s", host)
-	}
 	suffix := []string{"svc", "cluster", "local"}
-	for i, subdomain := range domains[2:] {
+	n := len(domains)
+	if n < 5 {
```
siggy: should this also check for `n > 6` ? that way we could ditch the the last return statement.

controller/api/destination/watcher/endpoints_watcher_test.go
```diff
@@ -357,6 +357,88 @@ spec:
...
+  namespace: ns
+  ownerReferences:
+  - kind: ReplicaSet
+    name: rs-1
+status:
+  phase: Running
+  podIP: 172.17.0.20`,
+			},
+			authority:                        "name1-3.name1.ns.svc.cluster.local:5959",
+			expectedAddresses:                []string{"172.17.0.20:5959"},
```
siggy: 🙏

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -393,6 +418,9 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) PodSe
 	for _, subset := range endpoints.Subsets {
 		resolvedPort := pp.resolveTargetPort(subset)
 		for _, endpoint := range subset.Addresses {
+			if pp.hostname != "" && pp.hostname != endpoint.Hostname {
```
siggy: why the `pp.hostname != ""` check? if `pp.hostname` == "" and `endpoint.Hostname` == "foo" this check will fail, is that the intent?

adleong:
They're referred to as hostnames in the Kubernetes documentation as well: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id

I agree that it's not the best name, but I'd prefer to match the kubernetes terminology.

controller/api/destination/watcher/k8s.go
adleong: Yes, but I think we'd need to make the `n == 6` case into the `else` block.  I think the current way is clearer.

controller/api/destination/watcher/endpoints_watcher.go
adleong: If the portPublisher's hostname is unset (empty string) that means that no hostname was specified.  In other words, this is a regular service lookup and not a stable pod id lookup.  Therefore, we do not want to skip any of the endpoints.

ihcsim: So does that mean `pp.hostname == "" && endpoint.Hostname == "foo"` will never happen? And if it does, then there is a bug elsewhere? Or can we just do `if pp.hostname != endpoint.Hostname` here?

adleong: `pp.hostname` corresponds to the subscriber's query whereas `endpoint.Hostname` corresponds to the hostname of the pod.  So, for example, if we did a destination lookup for `foo.default.svc.cluster.local` where `foo` is a service which selects a stateful set, then `pp.hostname` would be empty (because we are querying the entire service, not a particular stable id) but each pod in the service would have a non-empty `endpoint.Hostname`.

In this case, we want to select all the pods, and therefore do not want to hit the `continue`.

adleong:
This conflicted non-trivially with https://github.com/linkerd/linkerd2/pull/2950.  Since authority parsing is now more complex and depends on the cluster domain, I have pulled that logic up out of the watchers and into the destination server.  This required changing the watcher interfaces to accept service and profile IDs rather than raw authorities.  

@arminbuerkle: would you like to review this?

controller/cmd/destination/main.go
```diff
@@ -43,24 +43,27 @@ func main() {
 		log.Fatalf("Failed to listen on %s: %s", *addr, err)
 	}
 
+	global, err := config.Global(consts.MountPathGlobalConfig)
```
arminbuerkle: I'd probably move this into the `destination` package into the `NewServer` function.

We don't need to and don't want to set a different cluster domain than what's set in the config so calling it here give us no benefit.

By moving it you reduce the amount of parameters required for `NewServer`.

---
Since `config.Global` is needed for the trust domain too, i'd also move that into the `destination` package and pass `*disableIdentity ` instead of `trustDomain` to `NewServer`

controller/api/destination/watcher/k8s.go
```diff
@@ -50,25 +50,41 @@ func getHostAndPort(authority string) (string, Port, error) {
 // GetServiceAndPort is a utility function that destructures an authority into
 // a service and port.  If the authority does not represent a Kubernetes
 // service, an error is returned.  If no port is specified in the authority,
-// the HTTP default (80) is returned as the port number.
-func GetServiceAndPort(authority string) (ServiceID, Port, error) {
+// the HTTP default (80) is returned as the port number.  If the authority
+// is a pod DNS name then the pod hostname is also returned as the 3rd return
+// value.  See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/.
+func GetServiceAndPort(authority string, clusterDomain string) (ServiceID, Port, string, error) {
```
arminbuerkle: `GetServiceAndPort` now returns a `ServiceID`, `Port`, `string` and `error`.

---

It looks inconsistent that the pod hostname does not have a type while for example port does 

---

In addition returning a hostname from a `GetServiceAndPort` function seems confusing. IMO it's time to rename the function. Maybe `ParseAuthority`? Similar to go's `url.Parse`?

controller/api/destination/watcher/k8s.go
```diff
@@ -50,25 +50,41 @@ func getHostAndPort(authority string) (string, Port, error) {
...
+	n := len(domains)
+	// S.N.{suffix}
+	if n < 2+len(suffix) {
+		return ServiceID{}, 0, "", fmt.Errorf("Invalid k8s service %s", host)
 	}
-
-	// Needs to have `".svc."` to be a valid k8s svc
-	if domains[2] != "svc" {
-		return ServiceID{}, 0, fmt.Errorf("Invalid k8s service %s", host)
+	for i, subdomain := range domains[n-len(suffix):] {
```
arminbuerkle: If i'm not missing anything you could rewrite this loop into:

```go
// svc.{clusterDomain} != 'host' suffix
if !strings.HasSuffix(host, "svc."+clusterDomain) {
	return ServiceID{}, 0, "", fmt.Errorf("Invalid k8s service %s", host)
}
```
which is easier to understand.

controller/api/destination/watcher/k8s.go
```diff
@@ -50,25 +50,41 @@ func getHostAndPort(authority string) (string, Port, error) {
...
-		Namespace: domains[1],
+
+	if n == 2+len(suffix) {
+		// <service>.<namespace>.<suffix>
+		service := ServiceID{
+			Name:      domains[0],
+			Namespace: domains[1],
+		}
+		return service, port, "", nil
+	} else if n == 3+len(suffix) {
```
arminbuerkle: There's no `else if` needed since you're returning in the previous `if` statement.

If you only use an `if` you can save on indentations and make it easier to understand.

controller/cmd/destination/main.go
adleong: I think I prefer to keep the config reading in the main and pass the specific values into `NewServer`.  This reduces coupling between the server implementation and the config, and makes the server easier to test.

controller/api/destination/watcher/k8s.go
adleong: Great suggestion, thanks.

Issue #3114, opened 2019-07-20
Build: Support Multi Arch


aliariff:
### What problem are you trying to solve?

Right now, if we want to build Linkerd to be able to work with another architecture for example like the ARM by running `bin/docker-build` inside Raspberry Pi, it will not build correctly(the resulting image is not runnable inside the Raspberry Pi) because of the existing image that is pulled for runtime image and also builder image is not supporting multi-arch[1]  therefore it will not run in another architecture.

![image](https://user-images.githubusercontent.com/6806035/61580779-c50ae800-ab15-11e9-828d-3a36a7c3dd5c.png)

https://github.com/linkerd/linkerd2/blob/f535c2d3d2f0da76bd62f3cbc1e39994a773ed21/bin/docker-build-base#L17-L23

https://github.com/linkerd/linkerd2/blob/f535c2d3d2f0da76bd62f3cbc1e39994a773ed21/bin/docker-build-go-deps#L17-L24

### How should the problem be solved?

It would be great if we can customize the target build architecture.
I am proposing to add a new environment key called `BUILD_ARCH` so that we can define in what architecture the resulting image should be based on.

The behavior should be like this:
if `BUILD_ARCH` NOT set: just do as current behavior do.

if `BUILD_ARCH` is set:
- Add suffix to the tag based on the BUILD_ARCH value ex: BUILD_ARCH=arm -> `gcr.io/linkerd-io/go-deps:4c8f4294.arm`, `gcr.io/linkerd-io/base:2019-02-19.01.arm`. With this addition of a suffix, the code will build the image instead of pulling it from the registry because it does not exist.
- Add `--platform=(BUILD_ARCH value)` when running the docker build[2], with this setting the docker will pull the base images with the defined platform.

Another improvement that I propose is to remove the hardcoded image string in the Dockerfile and move it as an argument, so we can put it in one single place that generates the tag(with or without suffix) and always add `--build-arg` when running the docker build.
```
docker build ... \
        --build-arg RUNTIME_IMAGE=repo:base_image_tag \
        BUILDER_IMAGE=repo:go_deps_tag
```

`base_image_tag` and `go_deps_tag` will generate the tag.
The output ex: `base_image_tag` = `base:2019-02-19.01`
The output ex: `go_deps_tag` = `go-deps:4c8f4294.arm`

### Any alternatives you've considered?

[...]

Issue #1165, opened 2018-06-20
Support for ARM based architectures?


jeroenjacobs79:
Hi,

I'm experimenting with a Kubernetes cluster that runs on aarch64 architecture. I'm unable to install Conduit:

```
curl https://run.conduit.io/install | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1353  100  1353    0     0   2142      0 --:--:-- --:--:-- --:--:--  2140
Downloading conduit-0.4.4-linux...
Conduit was successfully installed 🎉

Copy /root/.conduit/bin/conduit into your PATH.  Then run

    conduit install | kubectl apply -f -

to deploy Conduit to Kubernetes.  Once deployed, run

    conduit dashboard

to view the Conduit UI.
Visit conduit.io for more information.
[root@testserver home]# export PATH=$PATH:$HOME/.conduit/bin
[root@testserver home]# conduit
-bash: /root/.conduit/bin/conduit: cannot execute binary file
[root@testserver home]# 

```

The binary that was installed seems to be for x64 architecture:

```
 readelf -h conduit 
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x456880
  Start of program headers:          64 (bytes into file)
  Start of section headers:          456 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         7
  Size of section headers:           64 (bytes)
  Number of section headers:         13
  Section header string table index: 3
```

Are there plans to support other architectures besides x64?

grampelberg:
It looks like you're only looking to have aarch64 support in the CLI? Or on the cluster side as well?

grampelberg:
Does this work for you?

[conduit cli](https://www.dropbox.com/s/iz8zhu95yh9yu9q/conduit?dl=0)

vielmetti:
Looks like this closed because of being stale - can we get a current status?

olix0r:
[...]

PR #3115, opened 2019-07-21
Add support for build in multi arch

aliariff:
This PR only means to address the needs of building Linkerd manually from local with different architecture. It will not make Linkerd officially support any other architecture than amd64, because it does not yet utilize `docker manifest` to support multi-arch images[1].

Changes:
- Introduce new env key `DOCKER_BUILD_ARCH` to set the output docker images architecture
- If `DOCKER_BUILD_ARCH` set, it will create the repositories with the suffix name of the arch, ex: `gcr.io/linkerd-io/base-arm`, `gcr.io/linkerd-io/grafana-arm64`, `gcr.io/linkerd-io/web-arm`
- Use the SHA method to version the `base` image same like the `go-deps`
- Remove hard-coded image name, instead, we use docker ARG and pass the image name dynamically, the arguments called `RUNTIME_IMAGE` and `BUILDER_IMAGE`
- Reuse `go-deps` image to do cross-compile to other architecture, therefore a new docker ARG called `GOARCH` introduced
- Remove the tag validation features, because it is not needed anymore, we already remove hard-coded image name and use docker ARG instead
- Add more documentation

What happens after this is merged:
- It WILL create a new base image with the tag is using SHA, called `gcr.io/linkerd-io/base:44134b1b` according to the SHA of current `Dockerfile-base` file
- It WILL create a new `go-deps` image because of the changes `Dockerfile-go-deps` file, and it called `gcr.io/linkerd-io/go-deps:60254700`(if nothing change with the `Gopkg.lock`)

- It WILL NOT create a repo with different arch name and push it to the registry because we don't set any `DOCKER_BUILD_ARCH` in `.travis.yml`, so it will behave as usual, only push the normal amd64 image

Tested manually:
:white_check_mark: Grafana
Screen Shot 2019-07-22 at 16 01 46

:white_check_mark: Debug
Screen Shot 2019-07-22 at 16 01 05

:white_check_mark: Web
Screen Shot 2019-07-22 at 16 00 09

:white_check_mark: CNI-Plugin
Screen Shot 2019-07-22 at 15 58 58

:white_check_mark: Controller
Screen Shot 2019-07-22 at 15 52 50

:white_check_mark: CLI-Bin
Screen Shot 2019-07-22 at 15 50 22

:x: Proxy (Future work to support Rust cross compile)
Screen Shot 2019-07-22 at 16 03 45

Future works in another PR:
- Support `linkerd2-proxy` that using Rust to do cross-compile in order to fully support other architecture
- Support multi-arch images officially by build and push other architectures like `arm` or `arm64` to the registry and link them using `docker manifest`. This is should be done after making the `linkerd2-proxy` works in another architecture

References:
[1] https://docs.docker.com/engine/reference/commandline/manifest/

Related issues:
#3114 
#1165


grampelberg:
@aliariff interested in getting this polished up so that we can merge it?

aliariff:
@grampelberg
Do you think this is still relevant for now? because earlier you said that there is a blocker which is the hardware and it is necessary to have it so that the integration test can be performed.

grampelberg:
@aliariff most definitely, I guess I should have used something more specific than "polished up". We'd love to work with you to get a checklist together on what would be required for this functionality to land in Linkerd. With a checklist, we'll be able to divide and conquer!

aliariff:
@grampelberg 
Yes sure, I think my idea at that time when I create this PR is just to introduce a way to create docker images in multiple arch, this is a good start to make Linkerd support multi-arch officially. 
But of course, as you say even after it is possible to create the docker images in another arch like `arm`, it is necessary to do an integration test for that arch.

So, in short, there is 2 part that we need to do.
1. Make it possible to create the docker images in different arch
2. Prepare the environment that we want to officially support (ex: `arm`, `arm64`) & integrate it with the current CI/CD and also pushed the created images to the GCR repo after the test succeeds.

What do you think?

grampelberg:
Hmm, that sounds right to me. It feels like there's a couple steps after 2 around release process and making sure all the artifacts are up to snuff.

It appears that this PR does step 1? Or does it do more than that?

aliariff:
@grampelberg 

Yes, this PR only means to address the first part.

richliu:
One question, 

in file cni-plugin/Dockerfile, it will download a jq binary file 
RUN curl -kL -o $(which jq) https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64

but this file is for AMD64 arch, so, I need to compile a ARM64 binary. 

Do you have some idea to solve that ?

grampelberg:
@richliu Maybe add some logic to check the arch and download the right binary?

richliu:
> @richliu Maybe add some logic to check the arch and download the right binary?

There is no official jq Linux arm64 binary on jq's website. I am not sure why it doesn't use distribution's default jq. probably version issue ? 
Download version is 1.6 
and ubuntu 18.04 is 1.5

grampelberg:
Honestly, that sounds fine. Give it a try and see if that works.

aliariff:
Closed. More thorough design will follow in the RFC repo.

PR #3135, opened 2019-07-24
Add RSA support to TLS libraries

### LINKED ISSUE #3131: Add RSA support to TLS libraries
Refactor the TLS libraries relied upon by the `proxy-injector` and `sp-validor` webhooks to have them work with RSA as well (they currently only deal with EC)

alpeb:
Fixes #3131

Wrapped private keys into either `PrivateKeyEC` or `PrivateKeyRSA` to
provide different certificate matching logic and marshaling depending on
the block type.

You can test having an RSA cert for the proxy injector by applying this
patch:

```diff
$ diff -u chart/templates/proxy_injector-rbac.yaml ~/tmp/proxy_injector-rbac.yaml
--- chart/templates/proxy_injector-rbac.yaml    2019-07-24 14:34:43.570616936 -0500
+++ /home/alpeb/tmp/proxy_injector-rbac.yaml    2019-07-24 13:41:03.150285099 -0500
@@ -1,4 +1,5 @@
 {{with .Values -}}
+{{- $ca := genCA "linkerd-proxy-injector.linkerd.svc" 365 -}}
 ---
 ###
 ### Proxy Injector RBAC
@@ -60,8 +61,8 @@
     {{ .CreatedByAnnotation }}: {{ .CliVersion }}
 type: Opaque
 data:
-  crt.pem: {{ b64enc .ProxyInjector.CrtPEM }}
-  key.pem: {{ b64enc .ProxyInjector.KeyPEM }}
+  crt.pem: {{ b64enc $ca.Cert }}
+  key.pem: {{ b64enc $ca.Key }}
 ---
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: MutatingWebhookConfiguration
@@ -81,7 +82,7 @@
       name: linkerd-proxy-injector
       namespace: {{ .Namespace }}
       path: "/"
-    caBundle: {{ b64enc .ProxyInjector.CrtPEM }}
+    caBundle: {{ b64enc $ca.Cert }}
   failurePolicy: {{ .WebhookFailurePolicy }}
   rules:
   - operations: [ "CREATE" ]
```

This will replace the logic to generate the cert with a call to Helm's
`genCA`, which uses RSA.

### DIFF
--- a/pkg/tls/codec.go
+++ b/pkg/tls/codec.go
@@ -52,7 +52,7 @@ func encode(buf *bytes.Buffer, blk *pem.Block) {
 func DecodePEMKey(txt string) (GenericPrivateKey, error) {
 	block, _ := pem.Decode([]byte(txt))
 	if block == nil {
-		return nil, errors.New("Not PEM-encoded")
+		return nil, errors.New("not PEM-encoded")
 	}
 	switch block.Type {
 	case "EC PRIVATE KEY":
@@ -60,15 +60,15 @@ func DecodePEMKey(txt string) (GenericPrivateKey, error) {
 		if err != nil {
 			return nil, err
 		}
-		return PrivateKeyEC{k}, nil
+		return privateKeyEC{k}, nil
 	case "RSA PRIVATE KEY":
 		k, err := x509.ParsePKCS1PrivateKey(block.Bytes)
 		if err != nil {
 			return nil, err
 		}
-		return PrivateKeyRSA{k}, nil
+		return privateKeyRSA{k}, nil
 	default:
-		return nil, fmt.Errorf("Unsupported block type: '%s'", block.Type)
+		return nil, fmt.Errorf("unsupported block type: '%s'", block.Type)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/tls/cred.go
```diff
@@ -12,9 +13,25 @@ import (
 )
 
 type (
+	// PrivateKeyEC wraps an EC private key
+	PrivateKeyEC struct {
```
siggy: can `PrivateKeyEC` and `PrivateKeyRSA` be private?

pkg/tls/codec.go
```diff
@@ -48,16 +48,28 @@ func encode(buf *bytes.Buffer, blk *pem.Block) {
...
+		}
+		return PrivateKeyEC{k}, nil
+	case "RSA PRIVATE KEY":
+		k, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		return PrivateKeyRSA{k}, nil
+	default:
+		return nil, fmt.Errorf("Unsupported block type: '%s'", block.Type)
```
siggy: nit: errors should not be capitalized... https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/tls/cred.go
alpeb: Good catch :+1:

pkg/tls/codec.go
alpeb: Indeed. Note to self: Logs can be capitalized, error strings not.

PR #3148, opened 2019-07-26
Allow setting custom cluster domain in service profiles

arminbuerkle:
Continue of #2950.

I decided to check for the `clusterDomain` in the config map in web server main for the same reasons as as pointed out here https://github.com/linkerd/linkerd2/pull/3113#discussion_r306935817

It decouples the server implementations from the config.

### DIFF
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -100,17 +100,17 @@ func newCmdProfile() *cobra.Command {
 			}
 
 			if options.template {
-				return profiles.RenderProfileTemplate(options.namespace, options.name, os.Stdout)
+				return profiles.RenderProfileTemplate(options.namespace, options.name, defaultClusterDomain, os.Stdout)
 			} else if options.openAPI != "" {
-				return profiles.RenderOpenAPI(options.openAPI, options.namespace, options.name, os.Stdout)
+				return profiles.RenderOpenAPI(options.openAPI, options.namespace, defaultClusterDomain, options.name, os.Stdout)
 			} else if options.tap != "" {
 				k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, 0)
 				if err != nil {
 					return err
 				}
-				return profiles.RenderTapOutputProfile(k8sAPI, options.tap, options.namespace, options.name, options.tapDuration, int(options.tapRouteLimit), os.Stdout)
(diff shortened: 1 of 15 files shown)


cli/cmd/profile.go
```diff
@@ -99,13 +99,13 @@ func newCmdProfile() *cobra.Command {
 			}
 
 			if options.template {
```
adleong: Should we be reading the cluster domain from the config object?

arminbuerkle: Yes and no:

I think there's two different use cases here:

1. I want to create my default service profile with the cluster domain read from the config (default)
2. I want to create a custom service profile with the cluster domain overwritten from cmd args

Allowing both uses cases at the same time shouldn't be a problem. I'd implement that behaviour in the (possible) next PR which introduces the cli flag to actually set a custom cluster domain.

grampelberg: Why would you want to create a custom service profile with a different domain from the control plane?

arminbuerkle: When you have multiple clusters, each with a different domain you don't want to switch into different kubectl contexts to create the appropriate service profile.

Instead it'd be easier to pass this as a cli flag so you can generate your service profile and then add it to source control.

So mostly out of convenience.

adleong: @arminbuerkle I see what you're saying, but I think it will ultimately be confusing if the Linkerd CLI's behavior doesn't match with the current kubectl context.  I'd prefer to maintain the assumption that the Linkerd CLI should always work with relation to the current kubectl context.

arminbuerkle: @adleong I understand, although i wouldn't agree. I think having both options can only be beneficial especially when the default option (when `--cluster-domain` cli arg is _NOT_ set) generates a service profile from the cluster config.

---

That being said: I implemented your suggestion in  8420bb3f51e0085e8aee2c3d1d3f0b6a40aac16f 

Overwriting it from CLI wouldn't be part of this PR anyways.

cli/cmd/profile.go
```diff
@@ -98,14 +101,30 @@ func newCmdProfile() *cobra.Command {
 				return err
 			}
 
+			var k kubernetes.Interface
+			k, err = k8s.NewAPI(kubeconfigPath, kubeContext, 0)
```
adleong: any reason not to just do `k, err :=` here instead of declaring k on the previous line?

arminbuerkle:
Just rebased to get the changes from https://github.com/linkerd/linkerd2/pull/3187 in.

Could someone take a look so we can get this merged?

arminbuerkle:
@ihcsim yes. CLI option and setting the proxy container env variable.

PR #3149, opened 2019-07-26
Remove unused argument

alpeb:
Removed unused argument in the `GetPatch()` function in `pkg/inject/inject.go`


alpeb:
@siggy Also the linter didn't catch this [unnecessary global scope](https://github.com/linkerd/linkerd2/pull/3135#discussion_r306997469) :man_shrugging:

PR #3150, opened 2019-07-26
Bump golangci-lint to 1.17.1

siggy:
Also add `bodyclose` linter and remove `exclude-use-default: false`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,16 +1,15 @@
 # This file specifies which linters golangci-lint should run.
 #
 # For descriptions of all available linters, run:
-# ./.golangci-lint-1.15.0 linters
+# ./.golangci-lint-1.17.1 linters
 # or browse to:
 # https://github.com/golangci/golangci-lint#supported-linters
 
 run:
   deadline: 5m
-issues:
-  exclude-use-default: false
 linters:
   enable:
+  - bodyclose
   - deadcode
   - depguard
   - goconst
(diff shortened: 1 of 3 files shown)


adleong:
Is there any specific motivation for bumping this, or is this just to stay modern?  After reading the golangci-lint docs I still have no idea what `exclude-use-default` does.

siggy:
@adleong Primarily to stay modern, also `bodyclose` is a nice addition.

We had original set `exclude-use-default: false` to ensure we were running a default `golint` linter. After a bit of git history spelunking, I'm going to add it back:
https://github.com/linkerd/linkerd2/pull/2239#discussion_r256675986

siggy:
@adleong Most explicit location I can see for the linter version is at:
https://github.com/golangci/golangci-lint/blob/v1.17.1/go.mod#L23

I have not yet tried this, but here's some info on VS Code integration, ideally this could ensure that VS Code is linting via the same version of golangci-lint:
https://github.com/golangci/golangci-lint#editor-integration

adleong:
Looks like it uses its own forked formatter here: https://github.com/golangci/golangci-lint/blob/v1.17.1/go.mod#L20

Looking at the git history of that repo: https://github.com/golangci/gofmt/commits/master

It looks like it was forked from this commit of goimports in the go tools repo:

https://github.com/golang/tools/tree/059bec968c61383b574810040ba9410712de36c5/cmd/goimports

So we probably want to standardize on using `goimports` built from revision `059bec9`

PR #3154, opened 2019-07-28
Add `l5d-require-id` header to Tap requests

kleimkuhler:
### Summary

In order for Pods' tap servers to start authorizing tap clients, the tap
controller must open TLS connections so that it can identity itself to the
server.

This change introduces the use of `l5d-require-id` header on outbound tap
requests.

### Details

When tap requests are made by the tap controller, the `Authority` header is an
IP address. The proxy does not attempt to do service discovery on such requests
and therefore the connection is over plaintext. By introducing the
`l5d-require-id` header the proxy can require a server name on the connection.
This allows the tap controller to identity itself as the client making tap
requests. The name value for the header can be made from the Pod Spec and tap
request, so the change is rather minimal.

#### Proxy Changes

* Update h2 to v0.1.26
* Properly fall back in the dst_router (linkerd/linkerd2-proxy#291)

### Testing

Unit tests for the header have not been added mainly because [no test
infrastructure currently exists](https://github.com/linkerd/linkerd2/blob/065c2218588093831dc2f8f49516e85e4a71b896/controller/tap/server_test.go#L241) to mock proxy requests. After talking with
@siggy a little about this, it makes to do in a separate change at some point
when behavior like this cannot be reliably tested through integration tests
either.

Integration tests do test this well, and will continue to do once
linkerd/linkerd2-proxy#290 lands.


kleimkuhler:
This is not dependent on #3155

controller/tap/server.go
```diff
@@ -105,8 +107,16 @@ func (s *server) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_T
 	}
 
 	for _, pod := range pods {
+		// create the expected pod identity from the pod spec
+		name := fmt.Sprintf("%s.%s.serviceaccount.identity.%s.cluster.local", pod.Spec.ServiceAccountName, req.Target.Resource.Namespace, s.controllerNamespace)
+		log.Debugf("requiring pod name={%s}", name)
```
adleong: a bit more context in this log line would be helpful.  something like "initiating tap request to %s with required id %s"

.proxy-version
```diff
@@ -1 +1 @@
-1c0d100
+9faa7ca
```
adleong: can you list the proxy changes that this picks up in the description of this commit?

controller/tap/server.go
```diff
@@ -105,8 +107,16 @@ func (s *server) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_T
 	}
 
 	for _, pod := range pods {
+		// create the expected pod identity from the pod spec
+		name := fmt.Sprintf("%s.%s.serviceaccount.identity.%s.cluster.local", pod.Spec.ServiceAccountName, req.Target.Resource.Namespace, s.controllerNamespace)
```
siggy: we're not consistent around this, but going forward recommend using getters when accessing protobuf structs, protects against nil pointers:
```go
req.GetTarget().GetResource().GetNamespace()
```

.proxy-version
kleimkuhler: When I squash this PR, I will list out the changes!

kleimkuhler: I just updated the PR comment so that you can see what it will look like.

PR #3155, opened 2019-07-28
Inject Tap service name into proxy PodSpec

kleimkuhler:
### Summary

In order for Pods' tap servers to start authorizing tap clients, the tap server
must be able to check client names against the expected tap service name.

This change injects the `LINKERD2_PROXY_TAP_SVC_NAME` into proxy PodSpecs.

### Details

The tap servers on the individual resources being tapped should be able to
verify that the client is the tap service. The `LINKERD2_PROXY_TAP_SVC_NAME` is
now injected as an environment variable in the proxies so that it can check this
value against the client name of the TLS connection. Currently, this environment
will go unused. There is an open PR (linkerd2-proxy#290) to use this variable in
the proxy, but this is *not* dependent on that merging first. 

Note: The variable is not injected if tap is disabled.

### Testing

Test output has been updated with the newly injected environment variable.

### DIFF
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -601,7 +601,7 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
 	}...)
 
 	// The tap service name depends on the `_l5d_ns` and `_l5d_trustdomain`
-	// variables set above, so ordering is significant.
+	// variables set above, so ordering is significant
 	if !conf.tapDisabled() {
 		sidecar.Env = append(sidecar.Env,
 			corev1.EnvVar{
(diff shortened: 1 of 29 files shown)


kleimkuhler:
This is not dependent on #3154

pkg/inject/inject.go
```diff
@@ -599,6 +600,15 @@ func (conf *ResourceConfig) injectPodSpec(patch *Patch) {
 		},
 	}...)
 
+	if !conf.tapDisabled() {
```
siggy: i think this should move up to near the other `conf.tapDisabled()` check, for readability, and also because there's an early return on line 557 in the case where identity is disabled.

kleimkuhler:
@siggy After our discussion, I added a comment to clarify that the order of setting `tapSvcName` is significant because it depends on `_l5d_ns` and `_l5d_trustdomain`.

I opened #3157 to note on the proxy work to take place to couple spawning the tap server only when identity has _not_ been disabled. There will be a similar issue for control plane as well.

PR #3161, opened 2019-07-30
Common structs for Helm

alpeb:
This PR adds `pkg/inject/template-values.go` which contains the structs
needed to set the helm variables during injection. They can also be
leveraged during installation:

```go
package inject

type (
        // Image contains the details to define a container image
        Image struct {
                Name       string
                PullPolicy string
                Version    string
        }

        // Port contains all the port-related setups
        Port struct {
                Admin               int32
                Control             int32
                Inbound             int32
                Outbound            int32
                IgnoreInboundPorts  string
                IgnoreOutboundPorts string
        }

        // Constraints wraps the Limit and Request settings for computational resources
        Constraints struct {
                Limit   string
                Request string
        }

        // Resources represents the computational resources setup for a given container
        Resources struct {
                CPU    Constraints
                Memory Constraints
        }

        // Identity contains the fields to set the identity variables in the proxy
        // sidecar container
        Identity struct {
                TrustDomain  string
                TrustAnchors string
        }

        // Proxy contains the fields to set the proxy sidecar container
        Proxy struct {
                Component             string
                ClusterDomain         string
                DisableIdentity       bool
                EnableExternalProfile bool
                HighAvailability      bool
                Identity              *Identity
                Image                 Image
                LogLevel              string
                ControlPlaneNamespace string
                Port                  Port
                UID                   int64
                ResourceRequirements  *Resources
        }

        // ProxyInit contains the fields to set the proxy-init container
        ProxyInit struct {
                Image                Image
                Port                 Port
                UID                  int64
                ResourceRequirements *Resources
        }
)
```

Signed-off-by: Alejandro Pedraza <alejandro@buoyant.io>

### DIFF
--- a/pkg/inject/template-values.go
+++ b/pkg/inject/template-values.go
@@ -1,0 +1,91 @@
+package inject
+
+type (
+	// Values contains the top-level elements in the Helm charts
+	Values struct {
+		Namespace        string
+		ClusterDomain    string
+		HighAvailability bool
+		Identity         *Identity
+
+		Proxy     *Proxy
+		ProxyInit *ProxyInit
+	}
+
+	// Proxy contains the fields to set the proxy sidecar container
+	Proxy struct {
+		Capabilities          *Capabilities
+		Component             string
+		DisableIdentity       bool
+		DisableTap            bool
+		EnableExternalProfile bool
+		Image                 *Image
+		LogLevel              string
+		SAMountPath           *SAMountPath
+		Ports                 *Ports
+		Resources             *Resources
+		UID                   int64
+	}
+
+	// ProxyInit contains the fields to set the proxy-init container
+	ProxyInit struct {
+		Capabilities        *Capabilities
+		IgnoreInboundPorts  string
+		IgnoreOutboundPorts string
+		Image               *Image
+		SAMountPath         *SAMountPath
+		Resources           *Resources
+	}
+
+	// DebugContainer contains the fields to set the debugging sidecar
+	DebugContainer struct {
+		Image *Image
+	}
+
+	// Image contains the details to define a container image
+	Image struct {
+		Name       string
+		PullPolicy string
+		Version    string
+	}
+
+	// Ports contains all the port-related setups
+	Ports struct {
+		Admin    int32
+		Control  int32
+		Inbound  int32
+		Outbound int32
+	}
+
+	// Constraints wraps the Limit and Request settings for computational resources
+	Constraints struct {
+		Limit   string
+		Request string
+	}
+
+	// Capabilities contains the SecurityContext capabilities to add/drop into the injected
+	// containers
+	Capabilities struct {
+		Add  []string
+		Drop []string
+	}
+
+	// SAMountPath contains the details for ServiceAccount volume mount
+	SAMountPath struct {
+		Name      string
+		MountPath string
+	}
+
+	// Resources represents the computational resources setup for a given container
+	Resources struct {
+		CPU    Constraints
+		Memory Constraints
+	}
+
+	// Identity contains the fields to set the identity variables in the proxy
+	// sidecar container
+	Identity struct {
+		TrustAnchorsPEM string
+		TrustDomain     string
+	}
+)


ihcsim:
Thanks for putting this together. Per _slack_ convo, I removed some of the redundant nested variables from `values.yaml`, which relied on YAML anchors to inherit their values from top-level variables. This approach is error-prone because e.g. `helm install --set HighAvailability=true` will only override the values of `{{.HighAvailability}}`, but not `{{.Proxy.HighAvailability}}`.

Here's the list of structs that correspond to the YAML in my `values.yaml`:
```go
// top level variables 
type InjectValues struct {
	Namespace        string
	ClusterDomain    string
	HighAvailability bool

	Proxy     *Proxy
	ProxyInit *ProxyInit
}

type Proxy struct {
	Capabilities          *Capabilities
	Component             string
	DisableIdentity       bool
	DisableTap            bool
	EnableExternalProfile bool
	Image                 *Image
	Identity              *Identity
	LogLevel              string
	MountPaths            []*MountPath
	Ports                 *Ports
	Resources             *Resources
	UID                   string
}

type ProxyInit struct {
	Capabilities        *Capabilities
	IgnoreInboundPorts  string
	IgnoreOutboundPorts string
	Image               *Image
	MountPaths          []*MountPath
	Resources           *Resources
}

type Capabilities struct {
	Add  []string
	Drop []string
}

type Image struct {
	Name       string
	Version    string
	PullPolicy string
}

type Identity struct {
	Issuer struct {
		CrtPEM string
	}

	TrustDomain string
}

type MountPath struct {
	Name      string
	MountPath string
}

type Resources struct {
	CPU    Constraints
	Memory Constraints
}

type Constraints struct {
	Limit   string
	Request string
}

type Ports struct {
	Admin    int
	Control  int
	Inbound  int
	Outbound int
}
```
LMK if this works.

alpeb:
Looks great @ihcsim :+1: 
I've pushed a change with your modifications, plus some of my own: 
- I moved `Identity` from `Proxy` to `InjectValues`, according to `values.yaml`
- In both `Proxy` and `ProxyInit` I replaced `MountPaths  []*MountPath`  with `SAMountPath *SAMountPath` given we're only adding a mount for the ServiceAccount so no need of a slice here
- I created a separate `Issuer` struct to be able to instantiate that piece

pkg/inject/template-values.go
```diff
@@ -0,0 +1,91 @@
...
+		Resources             *Resources
+		UID                   int64
+	}
+
+	// ProxyInit contains the fields to set the proxy-init container
+	ProxyInit struct {
+		Capabilities        *Capabilities
+		IgnoreInboundPorts  string
+		IgnoreOutboundPorts string
+		Image               Image
```
siggy: curious: why is `Image` a pointer in `Proxy` but not in `ProxyInit`?

pkg/inject/template-values.go
```diff
@@ -0,0 +1,91 @@
...
+		SAMountPath           *SAMountPath
+		Ports                 *Ports
+		Resources             *Resources
+		UID                   int64
+	}
+
+	// ProxyInit contains the fields to set the proxy-init container
+	ProxyInit struct {
+		Capabilities        *Capabilities
+		IgnoreInboundPorts  string
```
siggy: can should this be an array?

pkg/inject/template-values.go
alpeb: Just an oversight on my part

pkg/inject/template-values.go
alpeb: yeah we might. WDYT @ihcsim ?

ihcsim: The value of this variable eventually gets mapped into the `args` of the proxy-init container. From a rendering perspective, I think it should remain as string. Otherwise, wont' we end up with something like:

```
  initContainers:
  - args:
    ...
    - --inbound-ports-to-ignore
    - [4190,4191]
    - --outbound-ports-to-ignore
    - ["443"] 
```

ihcsim:
I made a mistake in the `Identity` struct. Instead of `Identity.Issuer.CrtPEM`, the data plane should be using `Identity.TrustAnchorsPEM`. Although they both have the [same default values](https://github.com/linkerd/linkerd2/blob/9a672dd5a9c436336a8b536856c56042bcf4ab51/cli/cmd/install.go#L1013-L1020), the former is used on the identity service (via `Secret`), whereas the trust anchor is used in the proxy's `LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS` env var, which I believe is the one needed here.

alpeb:
I've just pushed the changes according to the previous message. I've also changed the name of `injectValues` to `Values`, to avoid stuttering since we're already in the `inject` package.
@ihcsim please let me know when you're ready for me to merge this into `isim/helm-charts` :star:

Issue #3162, opened 2019-07-30
Tap controller validates resource in request body and URL are equal


kleimkuhler:
### Motivation

The aggregation layer will authorize that a user has the required privileges for the requested resource in the tap request URL. If the user is authorized, then the request will be passed to the tap controller.

Once the tap controller receives the authorized tap request, it unpacks the request body and taps the resources specified by those parameters. There is a possibility that these resources could be different, so the tap controller should first make sure that they are the same.

If the requested body resources are different from the request URL, then the request should be failed

siggy:
Fixed via #3167.

PR #3167, opened 2019-07-31
Introduce tap APIService, update `linkerd tap`

### LINKED ISSUE #2725: Make the tap service an APIExtension
Depends on https://github.com/linkerd/linkerd2/issues/2712

The tap service should be registered as an APIExtension so that it can ensure that all traffic that it receives has been authenticated and authorized by the API aggregation layer.  This allows to use RBAC to control which users and service accounts may access tap data.

On startup, the tap service should generate a serving CA, private key, and certificate and give the CA when registering as an APIExtension by posting to the ApiregistrationV1 API.  (Note that generating this material on startup means that the tap service may not run multiple replicas (HA mode).  As part of https://github.com/linkerd/linkerd2/issues/2176, this can be resolved by generating this material at install time and saving the CA key in a secret.)

The tap server should read the `extension-apiserver-authentication` configmap to get the client CA to use to validate the identity of calling clients.  This should be used to only allow tap requests from the aggregator.  This ensures all requests have been authorized.

By serving the tap TAP on `POST /apis/tap.linkerd.io/v1alpha1/namespaces/<ns>/taps` we cause the aggregator to check that the authenticated user has "create" permission on the "taps" resource in the "`<ns>`" namespace.

siggy:
The Tap Service enabled tapping of any meshed pod, regardless of user
privilege.

This change introduces a new Tap APIService. Kubernetes provides
authentication and authorization of Tap requests, and then forwards
requests to a new Tap APIServer, which implements a Kubernetes
aggregated API server. The Tap APIServer authenticates the client TLS
from Kubernetes, and authorizes the user via a SubjectAccessReview.

The `linkerd tap` command now makes requests against the new APIService.

The Tap APIService implements three Kubernetes-style endpoints:
GET  /apis/tap.linkerd.io/v1alpha1
POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:ns/tap
POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:ns/:res/:name/tap

Users authorize to the new `tap.linkerd.io/v1alpha1` via RBAC. Only the
`watch` verb is supported. Access is also available via subresources
such as `deployments/tap` and `pods/tap`.

This change introduces the following resources into the default Linkerd
install:
- Global
  - APIService/v1alpha1.tap.linkerd.io
  - ClusterRoleBinding/linkerd-linkerd-tap-auth-delegator
- `linkerd` namespace:
  - Secret/linkerd-tap-tls
- `kube-system` namespace:
  - RoleBinding/linkerd-linkerd-tap-auth-reader

Tasks not covered by this PR:
- `linkerd top`
- `linkerd dashboard`
- `linkerd profile --tap`
- removal of the unauthenticated tap controller

Fixes #2725, #3162

This PR is based on https://github.com/linkerd/linkerd2/compare/dad/tap-pod-2

### DIFF
--- a/bin/test-cleanup
+++ b/bin/test-cleanup
@@ -40,6 +40,18 @@ if ! customresourcedefinitions=$(kubectl --context=$k8s_context get customresour
   echo "no customresourcedefinitions found" >&2
 fi
 
-if [[ $namespaces_controlplane || $namespaces_dataplane || $clusterrolebindings || $clusterroles || $webhookconfigs || $validatingconfigs || $podsecuritypolicies || $customresourcedefinitions ]]; then
-  kubectl --context=$k8s_context delete $namespaces_controlplane $namespaces_dataplane $clusterrolebindings $clusterroles $webhookconfigs $validatingconfigs $podsecuritypolicies $customresourcedefinitions
+if ! apiservices=$(kubectl --context=$k8s_context get apiservices -l linkerd.io/control-plane-ns -oname); then
+  echo "no apiservices found" >&2
+fi
+
+if [[ $namespaces_controlplane || $namespaces_dataplane || $clusterrolebindings || $clusterroles || $webhookconfigs || $validatingconfigs || $podsecuritypolicies || $customresourcedefinitions || $apiservices ]]; then
+  kubectl --context=$k8s_context delete $namespaces_controlplane $namespaces_dataplane $clusterrolebindings $clusterroles $webhookconfigs $validatingconfigs $podsecuritypolicies $customresourcedefinitions $apiservices
+fi
+
+echo "cleaning up rolebindings in kube-system namespace in k8s-context [${k8s_context}]"
+if ! rolebindings=$(kubectl --context=$k8s_context -n kube-system get rolebindings -l linkerd.io/control-plane-ns -oname); then
+  echo "no rolebindings found in the kube-system namespace" >&2
+fi
+if [[ $rolebindings ]]; then
+  kubectl --context=$k8s_context delete $rolebindings -n kube-system
 fi
--- a/chart/templates/tap-rbac.yaml
+++ b/chart/templates/tap-rbac.yaml
@@ -38,6 +38,22 @@ subjects:
   name: linkerd-tap
   namespace: {{.Namespace}}
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: linkerd-{{.Namespace}}-tap-auth-delegator
+  labels:
+    {{.ControllerComponentLabel}}: tap
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:auth-delegator
+subjects:
+- kind: ServiceAccount
+  name: linkerd-tap
+  namespace: {{.Namespace}}
+---
 kind: ServiceAccount
 apiVersion: v1
 metadata:
@@ -46,4 +62,53 @@ metadata:
   labels:
     {{.ControllerComponentLabel}}: tap
     {{.ControllerNamespaceLabel}}: {{.Namespace}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: linkerd-{{.Namespace}}-tap-auth-reader
+  namespace: kube-system
+  labels:
+    {{.ControllerComponentLabel}}: tap
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
+- kind: ServiceAccount
+  name: linkerd-tap
+  namespace: {{.Namespace}}
+---
+kind: Secret
+apiVersion: v1
+metadata:
+  name: linkerd-tap-tls
+  namespace: {{ .Namespace }}
+  labels:
+    {{.ControllerComponentLabel}}: tap
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+  annotations:
+    {{ .CreatedByAnnotation }}: {{ .CliVersion }}
+type: Opaque
+data:
+  crt.pem: {{ b64enc .Tap.CrtPEM }}
+  key.pem: {{ b64enc .Tap.KeyPEM }}
+---
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1alpha1.tap.linkerd.io
+  labels:
+    {{.ControllerComponentLabel}}: tap
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+spec:
+  group: tap.linkerd.io
+  version: v1alpha1
+  groupPriorityMinimum: 1000
+  versionPriority: 100
+  service:
+    name: linkerd-tap
+    namespace: {{.Namespace}}
+  caBundle: {{ b64enc .Tap.CrtPEM }}
 {{end -}}
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -80,6 +80,7 @@ type (
 		Identity         *installIdentityValues
 		ProxyInjector    *proxyInjectorValues
 		ProfileValidator *profileValidatorValues
+		Tap              *tapValues
 	}
 
 	configJSONs struct{ Global, Proxy, Install string }
@@ -112,6 +113,10 @@ type (
 	}
 
 	profileValidatorValues struct {
+		*tlsValues
+	}
+
+	tapValues struct {
 		*tlsValues
 	}
 
@@ -231,6 +236,7 @@ func newInstallOptionsWithDefaults() *installOptions {
 			return id.String()
 		},
 
+		// TODO: rename to generateTLS or something
 		generateWebhookTLS: func(webhook string) (*tlsValues, error) {
 			root, err := tls.GenerateRootCAWithDefaults(webhookCommonName(webhook))
 			if err != nil {
@@ -428,6 +434,12 @@ func (options *installOptions) validateAndBuild(stage string, flags *pflag.FlagS
 		return nil, nil, err
 	}
 	values.ProfileValidator = &profileValidatorValues{profileValidatorTLS}
+
+	tapTLS, err := options.generateWebhookTLS(k8s.TapServiceName)
+	if err != nil {
+		return nil, nil, err
+	}
+	values.Tap = &tapValues{tapTLS}
 
 	values.stage = stage
 
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -263,6 +263,12 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
 	}
 	values.ProfileValidator = &profileValidatorValues{profileValidatorTLS}
 
+	tapTLS, err := fetchWebhookTLS(k, k8s.TapServiceName, options)
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not fetch existing tap secret: %s", err)
+	}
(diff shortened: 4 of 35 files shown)


controller/tap/apiserver.go
```diff
@@ -0,0 +1,342 @@
...
+		a.log.Error(err)
+		renderJSONError(w, err, http.StatusBadRequest)
+		return
+	}
+
+	a.log.Debugf("SubjectAccessReview: namespace: %s, resource: %s, name: %s, user: %s, group: %s",
+		namespace, resource, name, req.Header.Get(a.usernameHeader), req.Header[a.groupHeader],
+	)
+
+	err := pkgK8s.ResourceAuthzForUser(
```
adleong: Is this subject access review necessary? are there requests which could get past the aggregator but fail here?

pkg/protohttp/protohttp.go
```diff
@@ -131,24 +143,77 @@ func deserializePayloadFromReader(reader *bufio.Reader) ([]byte, error) {
...
+	if err != nil {
+		return fmt.Errorf("error unmarshalling array of [%d] bytes error: %v", len(messageAsBytes), err)
+	}
+
+	return nil
+}
+
+// TapReqToURL converts a TapByResourceRequest protobuf object to a URL for use
+// with the Kubernetes tap.linkerd.io APIService.
+func TapReqToURL(req *pb.TapByResourceRequest) string {
```
adleong: This feels pretty specific to the tap apiservice and a bit out of place in this file.

controller/tap/apiserver.go
siggy: I haven't been able to craft a request that gets by the aggregator but fails this SAR, but I'm not convinced it's not possible. This is the recommended pattern from Kubernetes:
https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#extension-apiserver-authorizes-the-request

pkg/protohttp/protohttp.go
siggy: I agree, but ended up putting it here for a few reasons:
- both `/cli/cmd` and `/controller/tap` depend on it, so it should be somewhere in `/pkg`
- `protohttp` already depends on `/controller/gen/public`, so putting `TapReqToURL` here does not add any new edges to our dependency tree
- it relates to the business of protobuf <-> http conversion

Given all this, the only other option would be to put that one function into a new package under `/pkg`, and that new package would still depend on `/controller/gen/public` (which is a more general issue to be fixed in #2751). Once #2751 completes, I'd be more open to a reorg of this code.

siggy:
@dadjeibaah The Tap Public API endpoint still serves `linkerd top`, `linkerd profile --tap` and `linkerd dashboard`. Once we get all those moved over, we can disable the public API endpoint. This work is tracked in #3171.

siggy:
@grampelberg filed #3177 to track tap ClusterRole

cli/cmd/install.go
```diff
@@ -231,6 +236,7 @@ func newInstallOptionsWithDefaults() *installOptions {
 			return id.String()
 		},
 
+		// TODO: rename to generateTLS or something
```
kleimkuhler: Is this TODO for the PR or the future?

cli/cmd/upgrade.go
```diff
@@ -263,6 +263,12 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
 	}
 	values.ProfileValidator = &profileValidatorValues{profileValidatorTLS}
 
+	tapTLS, err := fetchWebhookTLS(k, k8s.TapServiceName, options)
```
kleimkuhler: Possibly rename `fetchWebhookTLS` now as well?

controller/tap/apiserver.go
adleong: I'm pretty sure this is the exact same SAR check that the aggregator does.  So doing it here as well feels redundant.  Removing it also allows us to remove the auth-delegator role binding.

controller/cmd/tap/main.go
```diff
@@ -45,17 +49,46 @@ func main() {
...
 	}
 
+	_, port, err := net.SplitHostPort(lis.Addr().String())
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	// TODO: remove the network hop in favor of APIServer calling TapByResource
+	// directly.
+	tapClient, cc, err := tap.NewClient("127.0.0.1:" + port)
```
kleimkuhler: This is interesting. Why does there need to be this hop here introducing an "intermediate" client?

controller/tap/apiserver.go
```diff
@@ -0,0 +1,342 @@
...
+
+// /apis/tap.linkerd.io/v1alpha1
+// this is required for `kubectl api-resources` to work
+func (a *apiServer) handleAPIResourceList(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(a.apiResList)
+}
+
+// /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/tap
+// /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/:resource/:name/tap
```
adleong: It's also possible to make tap requests for a resource type without a resource name.  E.g. `bin/linkerd tap --as trudeau@buoyant.io -n linkerd deploy`

This results in a URL like `/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployments//tap` (note the empty name segment).  Interestingly, the aggregator seems to treat the empty segment as if it's not there: `X-Forwarded-Uri:[/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployments/tap]` and authorizes the user's ability to watch all deployments/tap, which is what we want.  

So, long story short, this all works correctly.  It's just a bit weird that we send along these non-standard URLs.  It might be more "correct" to completely omit the empty segment.

controller/tap/apiserver.go
```diff
@@ -0,0 +1,342 @@
...
+	tapReq := public.TapByResourceRequest{}
+	err = protohttp.HTTPRequestToProto(req, &tapReq)
+	if err != nil {
+		err = fmt.Errorf("Error decoding Tap Request proto: %s", err)
+		a.log.Error(err)
+		protohttp.WriteErrorToHTTPResponse(w, err)
+		return
+	}
+
+	url := protohttp.TapReqToURL(&tapReq)
```
kleimkuhler: Oh this is convenient for checking the protobuf object!

controller/tap/apiserver.go
siggy: Yeah, I was aware of the `deployments//tap` and pleasantly surprised it all just works. I'm inclined to leave it in, as `deployments//tap` makes clear to tap all deployments, while `deployments/tap` could mean either tap all deployments, or tap a single deployment called `tap`.

bin/test-cleanup
```diff
@@ -40,6 +40,18 @@ if ! customresourcedefinitions=$(kubectl --context=$k8s_context get customresour
...
+
+if [[ $namespaces_controlplane || $namespaces_dataplane || $clusterrolebindings || $clusterroles || $webhookconfigs || $validatingconfigs || $podsecuritypolicies || $customresourcedefinitions || $apiservices ]]; then
+  kubectl --context=$k8s_context delete $namespaces_controlplane $namespaces_dataplane $clusterrolebindings $clusterroles $webhookconfigs $validatingconfigs $podsecuritypolicies $customresourcedefinitions $apiservices
+fi
+
+echo "cleaning up rolebindings in kube-system namespace in k8s-context [${k8s_context}]"
+if ! rolebindings=$(kubectl --context=$k8s_context -n kube-system get rolebindings -l linkerd.io/control-plane-ns -oname); then
+  echo "no rolebindings found in the kube-system namespace" >&2
+fi
+if [[ $rolebindings ]]; then
```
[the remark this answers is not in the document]
siggy: Yup.

chart/templates/tap-rbac.yaml
```diff
@@ -46,4 +62,53 @@ metadata:
   labels:
     {{.ControllerComponentLabel}}: tap
     {{.ControllerNamespaceLabel}}: {{.Namespace}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
siggy: `RoleBinding/linkerd-linkerd-tap-auth-reader`, in the `kube-system` namespace, gives access to `cm/extension-apiserver-authentication` in the `kube-system` namespace, which gives us client TLS and header info, for authn and authz of incoming requests from Kubernetes to our apiserver. More details in the comment above `apiServerAuth`, also at: 
https://kubernetes.io/docs/tasks/access-kubernetes-api/#kubernetes-apiserver-client-authentication

`ClusterRoleBinding/linkerd-linkerd-tap-auth-delegator` gives the TAP APIServer SubjectAccessReview privileges. See `handleTap` for that implementation, also for more details:
https://kubernetes.io/docs/tasks/access-kubernetes-api/#extension-apiserver-authorizes-the-request

cli/cmd/install.go
siggy: For the future, `generateWebhookTLS` doesn't make sense for non-webhooks, but I'm trying to limit to size of this PR.

controller/cmd/tap/main.go
siggy: It's not strictly necessary, was just a quick and dirty way to wire up the APIServer to the Tap Controller.

controller/tap/apiserver.go
adleong: Actually... I wonder how the Linkerd CLI is even able to make these requests and have them get through at all...

This is what happens when I try to make the request manually:

```
$ curl -v -X POST http://localhost:8001/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployments//tap
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8001 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8001 (#0)
> POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployments//tap HTTP/1.1
> Host: localhost:8001
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Location: /apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/deployments/tap
< Date: Wed, 31 Jul 2019 22:52:12 GMT
< Content-Length: 0
<
* Connection #0 to host localhost left intact
```

adleong: Also: `WATCH deployments/tap` watches the deployment named tap, which is not a resource supported by `tap.linkerd.io`.  To tap the deplyoment named tap you'd need to do `WATCH deployments/tap/tap`.

siggy:
This PR now also addresses #3172

controller/tap/apiserver.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return s, lis, nil
+}
+
+// ServeHTTP handles all routes for the APIServer.
+func (a *apiServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	a.log.Debugf("ServeHTTP(): %+v", req)
+
+	// if `requestheader-allowed-names` was empty, allow any CN
```
adleong: I'm gonna be pedantic about this comment here because this is potentially confusing.  in local development we won't be providing a client cert at all so there will be no client CN.  (and our TLS config doesn't require a client cert, it only validates one if it is provided)

controller/cmd/tap/main.go
siggy: yup, filed #3171 to track, and linked to this comment.

controller/tap/apiserver.go
siggy: A bit more context, from https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#kubernetes-apiserver-client-authentication

```
The connection must be made using a client certificate whose CN is one of those listed in --requestheader-allowed-names. Note: You can set this option to blank as --requestheader-allowed-names="". This will indicate to an extension apiserver that any CN is acceptable.
```

PR #3173, opened 2019-07-31
Introduce `linkerd --as` flag for impersonation

siggy:
Similar to `kubectl --as`, global flag across all linkerd subcommands
which sets a `ImpersonationConfig` in the Kubernetes API config.

```bash
$ linkerd stat deploy --as foo
Cannot find Linkerd: configmaps "linkerd-config" is forbidden: User "foo" cannot get resource "configmaps" in API group "" in the namespace "linkerd"
```

```bash
$ linkerd check --as foo
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-config
--------------
× control plane Namespace exists
    namespaces "linkerd" is forbidden: User "foo" cannot get resource "namespaces" in API group "" in the namespace "linkerd"
    see https://linkerd.io/checks/#l5d-existence-ns for hints

Status check results are ×
```

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -171,6 +171,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		DataPlaneNamespace:    options.namespace,
 		KubeConfig:            kubeconfigPath,
 		KubeContext:           kubeContext,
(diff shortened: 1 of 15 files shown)


grampelberg:
Holy cow, this is awesome!

PR #3175, opened 2019-07-31
Cscott/demo ts metrics

scottcarol:
Now that Linkerd supports trafficsplit, we would like to return traffic split metrics in the UI so a user can easily get metrics about their trafficsplits and leaf services. 

Ideally we could use the `stat` API endpoint to return trafficsplit metrics if the user types `linkerd stat ts`. This is consistent with the current user experience and utilizes the complex and powerful `stat` endpoint. The downside is that the information we want to return (a list of trafficsplits, their apexes and leaf services, with metrics for each leaf service) is different than the information we currently return with `stat`, and we risk making the `stat` endpoint confusing with multiple `if` statements.

The alternative is to create a completely new command for traffic split metrics. The positives of a new command would be that we could build on this over time and tailor the endpoint to trafficsplits. The negatives are that users would have to learn a new command, and we would be recreating a lot of the logic of the `stat` endpoint but not closely enough to re-use the actual code.

This draft PR is a proof of concept way to get `trafficsplit` information from the `stat` API endpoint. It is meant to get high-level feedback for:

* how `api/public/stat_summary.go` could deal with a request from `linkerd stat ts`  
* protobuf changes
* using regex to match the leaf service name to authority name
* architecture of the data returned

**How to Run** 

1. Create a trafficsplit for an application in your `default` namespace -- [files and directions here if needed](https://github.com/scottcarol/trafficsplit)
2. Confirm trafficsplit was created
```bash
kubectl get ts
```
3. Port forward Prometheus:
```bash
kubectl -n linkerd port-forward $(kubectl --namespace=linkerd get po --selector=linkerd.io/control-plane-component=prometheus -o jsonpath='{.items[*].metadata.name}') 9090:9090
```
4. Run the go server for public-api
```bash
~/linkerd2 $ go run controller/cmd/public-api/main.go --kubeconfig ~/.kube/config --prometheus-url http://localhost:9090 --log-level debug
``` 
5. Save these sample files to your linkerd2 directory: [stat-ts.go](https://raw.githubusercontent.com/scottcarol/trafficsplit/master/stat-ts.go) (queries stat endpoint for trafficsplit info) and [stat-deploy.go](https://raw.githubusercontent.com/scottcarol/trafficsplit/master/stat-deploy.go) (queries stat endpoint for deployment info)
6. Run the files and compare
```bash
$ go run stat-ts.go
$ go run stat-deploy.go
```


scottcarol:
Meant to make a demo PR, closing this one

PR #3186, opened 2019-08-02
Update `linkerd top` to use Tap APIService

### LINKED ISSUE #3168: Update `linkerd top` to use the new Tap APIService
Follow up from #3167

siggy:
PR #3167 introduced a Tap APIService, and migrated `linkerd tap` to it.

This change migrates `linkerd top` to the new Tap APIService. It also
addresses a `panic: close of closed channel` issue, where two go
routines could both call `close(done)` on exit.

Fixes #3168

### DIFF
--- a/cli/tap/tap.go
+++ b/cli/tap/tap.go
@@ -3,9 +3,9 @@ package tap
 import (
 	"bufio"
 	"bytes"
-	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 
 	"github.com/golang/protobuf/proto"
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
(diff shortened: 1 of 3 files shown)


cli/tap/tap.go
```diff
@@ -0,0 +1,56 @@
...
+		fmt.Sprintf("%s%s", k8sAPI.Host, url),
+		bytes.NewReader(reqBytes),
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	httpRsp, err := client.Do(httpReq)
+	if err != nil {
+		log.Debugf("Error invoking [%s]: %v", "taps", err)
```
kleimkuhler: Is `"taps"` supposed to be the resources requested for tapping? I'm not sure if you intended to replace this with `req.GetTarget().GetResource()`--or something like that.

siggy: This is a good catch, it should have been printing the URL. This also got me thinking the URL construction could be a bit more robust. Updated.

Issue #3127, opened 2019-07-24
Refactor CLI Install Command To Work With 2.5 Helm Chart


ihcsim:
Refactor `linkerd install` to work with the new control plane Helm chart. The main changes should be instead of relying on the existing `inject` code to inject the proxy spec, we will have to rely on Helm's library to inject the proxy's partial template into the control plane template. Ensure that config overrides via the CLI options are honored by the partials.

Depends on https://github.com/linkerd/linkerd2/issues/3126.

PR #3195, opened 2019-08-05
Refactor proxy injection to use Helm charts

### LINKED ISSUE #3128: Refactor Proxy Injection To Work With 2.5 Helm Chart
Refactor the proxy injection logic in the `pkg/inject/inject.go` file to work with the new Helm chart. The main changes will be instead of relying on the `k8s.io/api/*` packages to build the proxy and proxy-init spec in Go, the spec will be read from the new partials and rendered by the Helm's library to produce the full injected YAML.

This task includes creating the `patch` chart.

Tests should cover the auto injection, and `linkerd inject` and `linkerd inject --manual` scenarios.

Depends on https://github.com/linkerd/linkerd2/issues/3126.

alpeb:
Fixes #3128

## New `patch` chart
A new chart `/charts/patch` was created, that generates the JSON patch
payload that is to be returned to the k8s API when doing the injection
through the proxy injector, and it's also leveraged by the `linkerd
inject --manual` CLI.

## Virtual File System changes
The VFS was used by `linkerd install` to access the old chart under
`/chart`. Now the proxy injection also uses the Helm charts to generate
the JSON patch (see above) so we've moved the VFS from `cli/static` to a
new common place under `/pkg/charts/static`, and the new root for the VFS is
now `/charts`.

`linkerd install` hasn't yet migrated to use the new charts (that'll
happen in #3127), so the only change in that regard was the creation of
`/charts/chart` which is a symlink pointing to `/chart` that
`install.go` now uses, so that the VFS contains both the old and new
charts, as a temporary measure.

You can see that `/bin/Dockerfile-bin`, `/controller/Dockerfile` and
`/bin/build-cli-bin` do now `go generate` pointing to the new location
(and the `go generate` annotation was moved from `/cli/main.go` to
`pkg/charts/static/templates.go`).

The symlink trick doesn't work when building the binaries through
Docker, so `/bin/Dockerfile-bin` replaces the symlink with an actual
copy of `/chart`.

Also note that in `/controller/Dockerfile` we now need to include the
`prod` tag in `go install` like we do in `/bin/Dockerfile-bin` so that
the proxy injector does use the VFS instead of the local file system.

## Other changes
- The common logic to parse a chart has been moved from `install.go` to
`/pkg/charts/util.go`.
- The special ENV var in the proxy for "outbound router capacity" that
only applies to the Prometheus pod is now handled directly in the proxy
partial and all the associated go code could be removed.
- The `patch.go` lib for generating the JSON patch in go along
with its tests `patch_test.go` are no longer needed.
- Lots of functions in `/pkg/inject/inject.go` got removed/simplified
with their logic being moved into the charts themselves. As a
consequence lots of things in `inject_test.go` became irrelevant.
- Moved `template-values.go` from `/pkg/inject` to `pkg/charts` as that
contains the go structs representation of the chart variables that
will be leveraged in #3127.

## Testing
Don't forget to run `/bin/helm.sh` whenever you make changes to charts
;-)


cli/cmd/install.go
```diff
@@ -752,73 +745,24 @@ func (values *installValues) render(w io.Writer, configs *pb.All) error {
...
-		t := path.Join(renderOpts.ReleaseOptions.Name, tmpl.Name)
-		if _, err := buf.WriteString(renderedTemplates[t]); err != nil {
-			return err
-		}
-	}
-
-	// Skip outbound port 443 to enable Kubernetes API access without the proxy.
-	// Once Kubernetes supports sidecar containers, this may be removed, as that
-	// will guarantee the proxy is running prior to control-plane startup.
-	configs.Proxy.IgnoreOutboundPorts = append(configs.Proxy.IgnoreOutboundPorts, &pb.Port{Port: 443})
```
alpeb: This is now handled directly in the chart

pkg/inject/inject.go
```diff
@@ -18,54 +19,14 @@ import (
...
-	envIdentityLocalName    = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME"
-	envIdentitySvcAddr      = "LINKERD2_PROXY_IDENTITY_SVC_ADDR"
-	envIdentitySvcName      = "LINKERD2_PROXY_IDENTITY_SVC_NAME"
-	envIdentityTokenFile    = "LINKERD2_PROXY_IDENTITY_TOKEN_FILE"
-	envIdentityTrustAnchors = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS"
-
-	identityAPIPort = 8080
-
-	envTapDisabled = "LINKERD2_PROXY_TAP_DISABLED"
-	envTapSvcName  = "LINKERD2_PROXY_TAP_SVC_NAME"
```
alpeb: All these consts are now in the charts

controller/proxy-injector/webhook_test.go
```diff
@@ -71,25 +71,21 @@ func TestGetPatch(t *testing.T) {
 			filename string
 			ns       *corev1.Namespace
 			conf     *inject.ResourceConfig
-			expected bool
 		}{
 			{
 				filename: "pod-inject-empty.yaml",
 				ns:       nsEnabled,
 				conf:     confNsEnabled(),
-				expected: true,
```
alpeb: `expected` is true in all the test cases, so I removed it.

alpeb:
@siggy my last push fixes things so that dependencies are taken care of automatically without forcing the build environments to have `helm` to bring them up-to-date.

But in order to install linkerd through helm (`helm install charts/linkerd2`), given that doesn't go through our go code, people do need to run `helm dep up charts/linkerd2` beforehand, for development. In prod that won't be necessary when the linkerd2 chart gets published in the official repo, and people would be able to just issue `helm install stable/linkerd2`.

cli/cmd/testdata/install_output.golden
```diff
@@ -792,6 +792,8 @@ spec:
           value: ns:$(_pod_ns)
         - name: LINKERD2_PROXY_IDENTITY_DISABLED
           value: disabled
+        - name: LINKERD2_PROXY_TAP_SVC_NAME
+          value: linkerd-tap.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
```
siggy: non-actionable observation: i would have expected `LINKERD2_PROXY_TAP_SVC_NAME` to already exist here prior to this PR, but glad to see it's here now.

pkg/charts/util.go
```diff
@@ -0,0 +1,94 @@
+package charts
```
siggy: naming nit: maybe just call this file `charts.go`? prefer to avoid `util` where possible.

pkg/charts/util.go
```diff
@@ -0,0 +1,94 @@
...
+		t := path.Join(renderOpts.ReleaseOptions.Name, tmpl.Name)
+		if _, err := buf.WriteString(renderedTemplates[t]); err != nil {
+			return bytes.Buffer{}, err
+		}
+	}
+
+	return buf, nil
+}
+
+func (chart *Chart) filesReader(dir string, files []*chartutil.BufferedFile) error {
```
siggy: any reason not to just declare this function as:
```go
func filesReader(dir string, files []*chartutil.BufferedFile) error
```

alpeb: good catch, that was my oversight :+1:

alpeb:
> mind adding a section to `BUILD.md` and/or `TEST.md` around when to use commands like `bin/helm.sh` and `helm dep up charts/linkerd2`? maybe documenting the typical dev flow of modifying templates and testing a `helm install` command?

Thanks for the feedback! I've created #3199 in the Helm project to track Helm docs effort :+1:

ihcsim:
@alpeb Opppss... looks like GH deleted this branch after I merged #3146, because this branch is set to be merged into the other feature branch. Can you reopen this PR? (GH doesn't offer me the option to reopen this PR.)

alpeb:
This PR has been replaced with #3200

PR #3203, opened 2019-08-06
Introduce tap-admin ClusterRole, web privs flag

### LINKED ISSUE #3177: Consider ClusterRole for tap.linkerd.io
PR #3167 introduced a TAP APIService, authorized via RBAC. Aside from cluster-admins, tapping resources now requires new explicit RBAC privileges.

Consider introducing a ClusterRole to give users tap privileges.

Context: https://github.com/linkerd/linkerd2/pull/3167#pullrequestreview-269306856

siggy:
The web dashboard will be migrating to the new Tap APIService, which
requires RBAC privileges to access.

Introduce a new ClusterRole, `linkerd-linkerd-tap-admin`, which gives
cluster-wide tap privileges. Also introduce a new ClusterRoleBinding,
`linkerd-linkerd-web-admin` which binds the `linkerd-web` service
account to the new tap ClusterRole. This ClusterRoleBinding is enabled
by default, but may be disabled via a new `linkerd install` flag
`--restrict-dashboard-privileges`.

Fixes #3177


cli/cmd/install.go
```diff
@@ -491,6 +494,10 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 		&options.omitWebhookSideEffects, "omit-webhook-side-effects", options.omitWebhookSideEffects,
 		"Omit the sideEffects flag in the webhook manifests, This flag must be provided during install or upgrade for Kubernetes versions pre 1.12",
 	)
+	flags.BoolVar(
+		&options.restrictDashboardPrivileges, "restrict-dashboard-privileges", options.restrictDashboardPrivileges,
+		"Restrict the Linkerd Dashboard's default privileges to disallow Tap.",
```
alpeb: Minuscule nit: no other help description finishes with a dot :stuck_out_tongue_winking_eye:

alpeb:
Oh I did found an issue afterwards: the new flag is exposed in `linkerd install control-plane` and not in `linkerd install config` but I think it should be the other way around...

ihcsim:
Installation works great with both `linkerd` and `helm`, besides the CLI option brought up in https://github.com/linkerd/linkerd2/pull/3203#issuecomment-519160268.

Some questions:

1. Why not use `RoleBinding` for `linkerd-linkerd-web-admin`?
1. Will the `linkerd-controller` needs this permission too?
1. Why is `--restrict-dashboard-privileges` necessary? Can we just default to installing the `linkerd-linkerd-web-admin` the `ClusterRoleBinding`?

siggy:
@alpeb Nice catch re: this flag not being on `linkerd install config`. I've pushed an update that puts the flag on both stages, since it affects the RBAC resource in `linkerd install config`, but also affects the flags stored in `linkerd-config` as part of `linkerd install control-plane`. It's not ideal that users must be careful to set this flags on both stages, but we were already in that situation with the `linkerd-cni-enabled` flag, and I've updated the code to couple the behavior of these two flags and make it explicit what's going on.

siggy:
@ihcsim Good questions re: the rbac stuff.

> 1. Why not use `RoleBinding` for `linkerd-linkerd-web-admin`?

My understanding is that since the `tap.linkerd.io` group contains non-namespaced resources., we need ClusterRoleBinding:
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding

> 2. Will the `linkerd-controller` needs this permission too?

If the `linkerd-controller`'s Public API was calling the new Tap APIService, we would, but with the Tap APIService, our clients (CLI, web) no longer go through the Public API for Tap. Instead it's client -> K8s API -> TAP ApiServer (in the tap pod).

> 3. Why is `--restrict-dashboard-privileges` necessary? Can we just default to installing the `linkerd-linkerd-web-admin` the `ClusterRoleBinding`?

We do default to installing `linkerd-linkerd-web-admin`, but thought it prudent to offer users a more locked-down approach option.

Issue #2998, opened 2019-06-25
Add a "No Info Found" label to tables in dashboard if they are empty


scottcarol:
### What problem are you trying to solve?

Currently when a table in the dashboard has no rows to display, it shows just the title and column titles. This could be confusing since it's not clear if something's gone wrong or if there is just no data found for that topic. 

![Screen Shot 2019-06-25 at 2 22 18 PM](https://user-images.githubusercontent.com/2289389/60139392-968e2d00-9762-11e9-9bd3-745b2d94ee09.png)

### How should the problem be solved?

Display a friendly message like "No inbound resources found" or "No edges found" somewhere on each table if empty.

cc @grampelberg

klingerf:
Alternatively, we might want to consider omitting the Edges table altogether if there's no edge info. That's more consistent with how we handle the Inbound and Outbound tables on this page. But I could also see the value in displaying a "no edges" message if it would help the user understand why there's no edge data.

scottcarol:
@klingerf The original dashboard behavior for the Edges table was to omit it altogether if no edges were found, but my current PR #2995 changes this, after @grampelberg made the point that "No edges found" is also valuable information and worth pointing out explicitly.

Currently we still display Inbound/Outbound tables even if they are empty - see blank table for Outbound below taken from `localhost:7777/namespaces/linkerd/pods/linkerd-prometheus-fd4dcc6fd-cs82g` 

![Screen Shot 2019-06-26 at 11 31 59 AM](https://user-images.githubusercontent.com/2289389/60205510-38b32100-9806-11e9-9127-87d6e093ab19.png)

klingerf:
@scottcarol Ahhhh, sorry, I was mistaken. Am not sure where I got the idea that we were omitting those tables. Ok, this all makes sense to me. Thanks for the explanation!

scottcarol:
I still think this could be a good improvement!

grampelberg:
I've got a POC empty state table in the mockups =)

scottcarol:
Closed in favor of #3611

Issue #455, opened 2018-02-26
Add mock timers for timing-dependent tests


hawkw:
- [x] Create a facade that abstracts over timer implementations
- [x] Modify existing code to use the timer facade
- [ ] Create a mock timer that implements the timer facade
- [ ] Rewrite time-dependent tests to use the mock timer instead

Probably depends on #412.

hawkw:
Some progress on this: 01aa4ff906f4d050b484eafad853f39cf327f413.

I've gotten everything ready for injecting mock timers, but there isn't yet a mock timer implementation, and I haven't rewritten the tests to use mock timers yet. Currently hitting a bug in some of the tests that I'm trying to work out.

olix0r:
related: https://github.com/tokio-rs/tokio/pull/368

olix0r:
@hawkw what is the state of this?

hawkw:
@olix0r the required Tokio changes have landed, and I think we should be able to rewrite our tests to use the Tokio mock timer support. I haven't been working on this, however, as it hasn't been prioritized and I'm not sure how big an impact it will have --- most of the flaky tests are flaky because they depend on _ordering_ rather than time.

PR #3208, opened 2019-08-07
Update web server to use tap APIService

kleimkuhler:
### Motivation

PR #3167 introduced the tap APIService and migrated `linkerd tap` to use it.
Subsequent PRs (#3186 and #3187) updated `linkerd top` and `linkerd profile
--tap` to use the tap APIService. This PR moves the web's Go server to now also
use the tap APIService instead of the public API. It also ensures an error
banner is shown to the user when unauthorized taps fail via `linkerd top`
command in *Overview* and *Top*, and `linkerd tap` command in *Tap*.

### Details

The majority of these changes are focused around piping through the HTTP error
that occurs and making sure the error banner generated displays the error
message explaining to view the tap RBAC docs.

`httpError` is now public (`HTTPError`) and the error message generated is short
enough to fit in a control frame (explained [here](https://github.com/linkerd/linkerd2/blob/kleimkuhler%2Fweb-tap-apiserver/web/srv/api_handlers.go#L173-L175)).

### Testing

The error we are testing for only occurs when the linkerd-web service account is
not authorzied to tap resources. Unforutnately that is not the case on Docker
For Mac (assuming that is what you use locally), so you'll need to test on a
different cluster. I chose a GKE cluster made through the GKE console--not made
through cluster-utils because it adds cluster-admin.

Checkout the branch locally and `bin/docker-build` or `ares-build` if you have
it setup. It should produce a linkerd with the version `git-04e61786`. I have
already pushed the dependent components, so you won't need to `bin/docker-push
git-04e61786`.

Install linkerd on this GKE cluster and try to run `tap` or `top` commands via
the web. You should see the following errors:

### Tap

web-tap-unauthorized

### Top

web-top-unauthorized


grampelberg:
A `Websocket error: internal error` sounds like we're broken and it isn't expected. Can we have a specific message for auth issues? Like `Missing authorization, see https://linkerd.io/tap-rbac to remedy`? It'd be nice if the link was clickable and formatted that way as well.

@siggy should this be a specific section on that page for the dashboard? We'll need to explain the web service account and such.

web/srv/api_handlers.go
```diff
@@ -168,10 +170,34 @@ func (h *handler) handleAPITopRoutes(w http.ResponseWriter, req *http.Request, p
...
+		log.Debugf("tap error: %s", err.Error())
+		msg = err.Error()[:123]
+	}
+
+	return msg
+}
+
 func websocketError(ws *websocket.Conn, wsError int, msg string) {
-	ws.WriteControl(websocket.CloseMessage,
+	err := ws.WriteControl(websocket.CloseMessage,
```
siggy: i'd modify `websocketError` to take an `err error` rather than `msg string`, and then call `createTapErrorMessage` from within this function, otherwise other calls to `websocketError` still could hit that character limit.

web/srv/api_handlers.go
```diff
@@ -168,10 +170,34 @@ func (h *handler) handleAPITopRoutes(w http.ResponseWriter, req *http.Request, p
...
+			msg = fmt.Sprintf("tap authorization failed, visit %s for more information", tap.TapRbacURL)
+		} else {
+			msg = httpErr.WrappedError.Error()[:123]
+		}
+	} else {
+		log.Debugf("tap error: %s", err.Error())
+		msg = err.Error()[:123]
+	}
+
+	return msg
```
siggy: i think it's useful to return the code from non-StatusForbidden `HTTPError`'s, rather than just the `Wrapped` error. you can treat non-StatusForbidden `HTTPError`'s and all other non-HTTPError's the same:

```go
// websocket.maxControlFramePayloadSize - 2 bytes for payload size
const maxControlFrameMsgSize = 123

func createTapErrorMessage(err error) string {
	log.Debugf("tap error: %s", err)

	if httpErr, ok := err.(protohttp.HTTPError); ok && httpErr.Code == http.StatusForbidden {
		return fmt.Sprintf("tap authorization failed, visit %s for more information", tap.TapRbacURL)
	}

	msg := err.Error()
	if len(msg) > maxControlFrameMsgSize {
		return msg[:maxControlFrameMsgSize]
	}

	return msg
}
```

siggy:
@grampelberg Once #3203 ships the dashboard should _just work_, and this error will be more exception than rule, but I do think some special treatment here is warranted.

Recommend in increasing order of effort (and descending order of whether we should include it for 2.5, I think we should just do the first two):
1. Document `linkerd-web` RBAC management, filed #455 to track.
1. Update the `Websocket error: internal error...` message to be more RBAC-specific.
1. Render the error message with a clickable URL.
1. Render the error message differently on the dashboard from other error messages.

web/srv/api_handlers.go
kleimkuhler: Good catch! I'll make this change and the others above.

kleimkuhler:
@siggy Thanks for the comments I addressed them all above.

@grampelberg I changed the error message so that it now only displays: `Missing authorization, visit https://linkerd.io/tap-rbac to remedy` as shown in the picture below. I worked with @scottcarol for a little bit to try and get the error banner to render the URL as a link, but ultimately was unable to after trying to get [react-linkify](https://www.npmjs.com/package/react-linkify) to render it. As @siggy mentioned in the above comment, we can address this in follow-up PRs in time for a stable 2.5.

The more significant change in this commit is that tap errors now return a `websocket.ClosePolicyViolation` status code so that we can branch on that [here](https://github.com/linkerd/linkerd2/blob/5850121da6f0db97694757c5e35e8d44c4136d7f/web/app/js/components/Tap.jsx#L110-L111) and not display the `Websocket close error [${e.code}: ${wsCloseCodes[e.code]}]` prefix. Branching on a unique code was chosen over branching on a regex such as looking for `unauthorized` in the reason.

`ClosePolicyViolation` was chosen because in the HTTP standard it notes:

> This is a generic status code, used when codes 1003 and 1009 are not suitable.

kleimkuhler:
It looks like this line in the CI logs indicates something in `protohttp` is failing: `FAIL	github.com/linkerd/linkerd2/pkg/protohttp [build failed]`. I changed the `HTTPError` type name in here to make it public and added a comment, but nothing else. Any idea what's up? I don't see this error locally.

siggy:
@kleimkuhler re: the ci failure, locally i get:
```bash
$ go test ./...
...
# github.com/linkerd/linkerd2/pkg/protohttp [github.com/linkerd/linkerd2/pkg/protohttp.test]
pkg/protohttp/protohttp_test.go:106:26: undefined: httpError
pkg/protohttp/protohttp_test.go:153:16: undefined: httpError
...
```

kleimkuhler:
Ah jeez thank you. I was looking for build issues with `go build` and not thinking it was just in the tests... It is fixed now

kleimkuhler:
Yep just fixed that lint. Sounds good!

kleimkuhler:
@scottcarol Oh awesome thanks for looking more into that! I'll get this change in now.

kleimkuhler:
Ok the link is clickable now!

linkify-error

scottcarol:
Your updated branch works for me, the error message was rendered and linked properly. 👍 

I'm still seeing that l5d-bot is failing, but it failed in a different place this time:
```
Checking if there is a Kubernetes cluster available...Starting to serve on 127.0.0.1:8080

Failed to connect to Kubernetes cluster
cleaning up control-plane namespaces in k8s-context []
cleaning up data-plane namespaces in k8s-context []
no control-plane namespaces found
no data-plane namespaces found
no clusterrolebindings found
no clusterroles found
no mutatingwebhookconfigurations found
no validatingwebhookconfigurations found
no podsecuritypolicies found
no customresourcedefinitions found
no apiservices found
cleaning up rolebindings in kube-system namespace in k8s-context []
```

kleimkuhler:
Hm yea not sure why it says it failed. It [passed](https://travis-ci.org/linkerd/linkerd2/builds/569412148?utm_source=github_status&utm_medium=notification) CI this time.

kleimkuhler:
Okay I ran the integration tests locally and everything passed. I'm going to merge this.

```
➜  linkerd2 git:(kleimkuhler/web-tap-apiserver) bin/test-run $PWD/target/cli/darwin/linkerd l5d-integration
==================RUNNING ALL TESTS==================
Testing Linkerd version [git-96c7c435] namespace [l5d-integration] k8s-context []
Test script: [install_test.go] Params: [-failfast --upgrade-from-version=stable-2.4.0 --linkerd-namespace=l5d-integration-upgrade]
ok      command-line-arguments  128.514s
Test script: [install_test.go] Params: [-failfast --linkerd-namespace=l5d-integration]
ok      command-line-arguments  117.239s
Test script: [serviceprofiles_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  56.900s
Test script: [egress_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  19.422s
Test script: [inject_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  12.854s
Test script: [get_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  16.492s
Test script: [tap_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  43.629s
Test script: [stat_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  10.234s
Test script: [routes_test.go] Params: [--linkerd-namespace=l5d-integration]
ok      command-line-arguments  0.820s

=== PASS: all tests passed
```

scottcarol:
Thanks for doing the extra tests! 💪

adleong:
This is awesome!

I may have missed the boat on the high level UX discussion, but what do folks think about replacing the content of the "Live requests" and  "Tap" sections with a friendly message rather than using error banners?  Especially since the error banner can be dismissed and then you're left with a UI that doesn't work.  I think replacing the section entirely communicates "This feature is disabled" rather than "This feature is broken".  WDYT?

scottcarol:
@adleong I agree that dashboard-wide we could do a better job explaining the reasons why no data exists and whether or not that is a concern. This aligns with #2998 which came out of conversation with @grampelberg about what to do if no data is found (currently in the dashboard we show an empty table if there are no rows to display, with no additional message).

adleong:
Something along the lines of:

Screen Shot 2019-08-08 at 11 05 35 AM

grampelberg:
I personally prefer @adleong's suggestion. Should we get a ticket to track that or just update #2998?

scottcarol:
Is it possible for a dashboard user to have Top/Tap access for certain traffic streams and not others? 

If the user would be blocked from all traffic streams when using the dashboard and we could check with just one query, the above suggestion seems simple to implement. 

If the user could have access to some streams but not others, we'd need to allow the user to formulate their desired query before telling them "yes" or "no." I think a separate ticket would make sense for this while keeping in mind #2998. Happy to make the ticket later today.

grampelberg:
I would like to always fire off the request and handle the 403 response in-panel instead of as part of a popup.

kleimkuhler:
Wrote up the issue here: #3220

PR #3209, opened 2019-08-07
Fix bug in service profile name generation

alpeb:
Followup to #3148

Wrong args order in call to `profiles.RenderOpenAPI` was generating an
invalid service profile name.


alpeb:
Thanks @siggy. Integration tests did pass in my machine :computer:

PR #3210, opened 2019-08-08
Fix invalid `l5d-require-id` for some tap requests

siggy:
PR #3154 introduced an `l5d-require-id` header to Tap requests. That
header string was constructed based on the TapByResourceRequest, which
includes 3 notable fields (type, name, namespace). For namespace-level
requests (via commands like `linkerd tap ns linkerd`), type ==
`namespace`, name == `linkerd`, and namespace == "". This special casing
for namespace-level requests yielded invalid `l5d-require-id` headers,
for example: `pd-sa..serviceaccount.identity.linkerd.cluster.local`.

Fix `l5d-require-id` string generation to account for namespace-level
requests. The bulk of this change is tap unit test updates to validate
the fix.


controller/tap/server.go
```diff
@@ -82,33 +86,36 @@ func (s *server) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_T
...
 	if err != nil {
 		return apiUtil.GRPCError(err)
 	}
 
 	for _, pod := range pods {
 		// create the expected pod identity from the pod spec
-		ns := req.GetTarget().GetResource().GetNamespace()
+		ns := res.GetNamespace()
+		if res.GetType() == pkgK8s.Namespace {
+			ns = res.GetName()
```
siggy: this is the actual fix.

controller/tap/server.go
```diff
@@ -50,14 +50,18 @@ func (s *server) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_T
...
+	res := req.GetTarget().GetResource()
+	if res == nil {
+		return status.Error(codes.InvalidArgument, "TapByResource received nil target Resource")
+	}
+	if req.GetMaxRps() == 0.0 {
 		req.MaxRps = defaultMaxRps
 	}
 
-	objects, err := s.k8sAPI.GetObjects(req.Target.Resource.Namespace, req.Target.Resource.Type, req.Target.Resource.Name)
+	objects, err := s.k8sAPI.GetObjects(res.GetNamespace(), res.GetType(), res.GetName())
```
kleimkuhler: Thanks for switching these over!

controller/tap/server_test.go
```diff
@@ -195,63 +261,135 @@ metadata:
...
+			if err != nil {
+				t.Fatalf("Invalid port: %s", port)
+			}
+
+			_, fakeGrpcServer := newGRPCTapServer(uint(tapPort), "controller-ns", k8sAPI)
+
+			k8sAPI.Sync()
+
+			err = fakeGrpcServer.TapByResource(&exp.req, &stream)
+			if !reflect.DeepEqual(err, exp.err) {
```
kleimkuhler: Ah this is good to see. Thanks for adding this

PR #3215, opened 2019-08-08
Check for 403 status code when preparing tap error

### LINKED ISSUE #3213: All tap errors render with missing error prefix in dashboard
PR #3208 changed the status code returned from `tap.Reader` errors to `1008: Policy Violation`. This was so that in the UI, we could branch on this error code and only display `Missing authorization, visit ...` without the prefix `Websocket close error [ ... ] reason` as shown [here](https://github.com/linkerd/linkerd2/pull/3208/files#diff-a05fea277c48fcafd6375c1922d0defaR110).

`tap.Reader` can return errors unrelated to authorization such as no resources found for a certain selection (e.g. statefulsets). These errors now render only as `no pods for statefulset/` as show in the picture below.

no-statefulsets

We may still want the `Websocket close error` prefix on these errors, so we should possibly revert the specific branch for `1008: Policy Violation`.

kleimkuhler:
Fixes #3213

### DIFF
--- a/web/srv/api_handlers.go
+++ b/web/srv/api_handlers.go
@@ -180,6 +180,7 @@ func (h *handler) handleAPITopRoutes(w http.ResponseWriter, req *http.Request, p
 func createTapErrorMessage(err error) string {
 	log.Debugf("tap error: %s", err.Error())
 
+	// TODO: reconcile this cast and 403 check with the one in handleAPITap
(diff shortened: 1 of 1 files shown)


scottcarol:
Was just about to say 🚢 on this one after testing as an unauthorized and an authorized user!

This is a good fix for the current UX, which will likely change as discussed in #3208.

PR #3216, opened 2019-08-08
Add changes for `edge-19.8.2`

scottcarol:


CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+  * Introduced parent resources for all tappable subresources in the tap
+    APIService, enabling commands such as `kubectl auth can-i`
+* Controller
+  * Introduced a new ClusterRole, `linkerd-linkerd-tap-admin`, which gives
+    cluster-wide tap privileges. Also introduced a new ClusterRoleBinding,
+    `linkerd-linkerd-web-admin`, which binds the `linkerd-web` service account
+    to the new tap ClusterRole. This ClusterRoleBinding is enabled by default,
+    but may be disabled via a new `linkerd install` flag,
+    `--restrict-dashboard-privileges`
+  * Added the ability to set a custom cluster domain in service profiles (thanks
```
ihcsim: This feature isn't completed yet. There isn't a way for users to explicitly specify custom cluster domain, until the associated CLI option is added. Maybe we can omit it for now.

CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+  * Moved `linkerd top` and `linkerd profile --tap` to the new tap APIService
+  * Introduced parent resources for all tappable subresources in the tap
+    APIService, enabling commands such as `kubectl auth can-i`
+* Controller
+  * Introduced a new ClusterRole, `linkerd-linkerd-tap-admin`, which gives
+    cluster-wide tap privileges. Also introduced a new ClusterRoleBinding,
+    `linkerd-linkerd-web-admin`, which binds the `linkerd-web` service account
+    to the new tap ClusterRole. This ClusterRoleBinding is enabled by default,
+    but may be disabled via a new `linkerd install` flag,
+    `--restrict-dashboard-privileges`
```
ihcsim: I think we should make a mention of the effects of this option. Seems like certain things on the web UI won't work with `--restrict-dashboard-privileges=true`. I will defer to @siggy or @grampelberg for wording.

CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+using the `helm install` and `helm delete` commands. Proxy injection also now
+uses Helm charts.
+
+No changes were made to the existing `linkerd install` behavior.
+
+For detailed installation steps using Helm, see the notes for [PR #3146]
+(https://github.com/linkerd/linkerd2/pull/3146).
+
+* CLI
+  * Moved `linkerd top` and `linkerd profile --tap` to the new tap APIService
```
siggy: maybe more user-facing:
```
`linkerd top` and `linkerd profile --tap` now require `tap.linkerd.io` RBAC privileges, see https://linkerd.io/tap-rbac for more info
```

CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+  * Introduced parent resources for all tappable subresources in the tap
+    APIService, enabling commands such as `kubectl auth can-i`
```
siggy: similar user-facing thought:
```
Modified `tap.linkerd.io` APIService to enable usage in `kubectl auth can-i` commands
```

CHANGES.md
siggy: Agree about focusing on the end-user effect. Consider moving mention of `--restrict-dashboard-privileges` to the CLI section, something like:
```
Introduced `--restrict-dashboard-privileges` flag to `linkerd install` command, to restrict the Linkerd Dashboard's default privileges to disallow Tap
```

CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+    `linkerd-linkerd-web-admin`, which binds the `linkerd-web` service account
+    to the new tap ClusterRole. This ClusterRoleBinding is enabled by default,
+    but may be disabled via a new `linkerd install` flag,
+    `--restrict-dashboard-privileges`
+  * Added the ability to set a custom cluster domain in service profiles (thanks
+    @arminbuerkle!)
+  * Removed successfully completed `linkerd-heartbeat` jobs from pod listing in
+    the linkerd control plane to streamline `get po` output (thanks
+    @Pothulapati!)
+* Proxy
```
siggy: the most recent proxy bump i see in master is from Aug 1 (f146041c25f79b9203ca9c3750fabe0c14b5c7df), expected?

hawkw: The changes mentioned here would need us to be on at least linkerd/linkerd2-proxy@f9dd4f8e28b9a6d8cdbe8094368495d94441b741

CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+  * Fixed a bug in `tap` which had caused the tap output to hang when reaching
```
hawkw: What about "Fixed a bug in `tap` which caused the tap output to hang after 100 events"? The limit mentioned here is kind of an implementation detail.

CHANGES.md
```diff
@@ -1,3 +1,44 @@
...
+
+This edge release introduces the new Linkerd control plane Helm chart, named
+linkerd2. Helm users can now install and remove the Linkerd control plane by
+using the `helm install` and `helm delete` commands. Proxy injection also now
+uses Helm charts.
+
+No changes were made to the existing `linkerd install` behavior.
+
+For detailed installation steps using Helm, see the notes for [PR #3146]
+(https://github.com/linkerd/linkerd2/pull/3146).
```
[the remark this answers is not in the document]
scottcarol: yes, good catch!

CHANGES.md
scottcarol: Thanks for the suggestions @ihcsim @siggy. To clarify, should we keep the background information about the new ClusterRole in the `Controller section`, or delete that portion entirely and only mention this in the `CLI` section?

CHANGES.md
scottcarol: Just talked with @olix0r and I will not bump the proxy before this edge release, due to an integration test failure. I am removing all of the `Proxy` bullet points from these release notes as they all occur after the current proxy version.

CHANGES.md
siggy: Ah good point. Yeah I think it's fine to mention the new ClusterRole in the `Controller section`. It's relevant because if users want to give themselves (or their serviceaccounts) tap privileges, they can use that new `linkerd-linkerd-tap-admin` role.

scottcarol:
Just pushed changes, would you mind giving the `Controller` and `CLI` changes one more look @siggy @ihcsim?

CHANGES.md
```diff
@@ -1,3 +1,38 @@
+## edge-19.8.2
+
+**Significant Update**
+
+This edge release introduces the new Linkerd control plane Helm chart, named
+linkerd2. Helm users can now install and remove the Linkerd control plane by
```
siggy: maybe wrap the helm chart name, `linkerd2`?

CHANGES.md
```diff
@@ -1,3 +1,38 @@
...
+    info
+  * Modified `tap.linkerd.io` APIService to enable usage in `kubectl auth can-i`
+    commands
+  * Introduced `--restrict-dashboard-privileges` flag to `linkerd install`
+    command, to restrict the dashboard's default privileges to disallow tap
+* Controller
+  * Introduced a new ClusterRole, `linkerd-linkerd-tap-admin`, which gives
+    cluster-wide tap privileges. Also introduced a new ClusterRoleBinding,
+    `linkerd-linkerd-web-admin`, which binds the `linkerd-web` service account
+    to the new tap ClusterRole `--restrict-dashboard-privileges`
```
siggy: dangling `--restrict-dashboard-privileges`

scottcarol: Ok thanks - so the text of that bullet point should end in `ClusterRole`

CHANGES.md
```diff
@@ -1,3 +1,38 @@
+## edge-19.8.2
+
+**Significant Update**
```
wmorgan: I didn't really understand this last time and I definitely don't understand it this time... what does this signify? I think we should take it out.

Issue #3143, opened 2019-07-25
Incorporate Helm Charts Validation Into Dev Workflow


ihcsim:
08/07: To ensure build consistency and continuous charts validation, we propose the following additions to our existing dev workflow:

* Download and pin the Helm CLI with the `bin/helm.sh` script
* Ensure the `bin/helm.sh` script is invoked during the local and CI build
* Add unit test to diff the output of `helm template` with a set of golden files
* Introduce integration test to test the `helm install` command

----
07/24: Add a shell script to perform the following steps:

* install, check Helm client version
* `helm dependency up`
* `helm lint`
* `helm test # to be evaluated`
* `helm install [--dep-up] [--verify] --dry-run # maybe not necessary? It requires `tiller` to be installed`
* Inject the build version into the `appVersion` property of `Chart.yaml`
* `helm package # maybe not necessary`

grampelberg:
@ihcsim is this just for testing or release as well?

ihcsim:
Definitely the testing phase as a start. Haven't thought too much about the releasing and publishing process yet.

ihcsim:
See #3218 and #3223.

ihcsim:
Fixed by #3218, #3223 and #3248.

ihcsim:
Chart release and publication work is tracked under #3256.

Issue #3199, opened 2019-08-06
Document Helm commands in dev workflow


alpeb:
Update BUILD.md and TEST.md to detail the workflow to follow when testing Helm template modifications, like using `bin/helm.sh` and other Helm commands.

alpeb:
Now that #3211 has been merged, the instructions should point to `bin/helm` and `bin/helm.sh` will be removed.

PR #3218, opened 2019-08-08
Bash function to setup Helm

alpeb:
This introduces the `install_helm` bash function that installs the
`helm` binary under `/target`, and optionally Tiller in the current
cluster.

To be consumed by the Helm helper scripts for development, and by
integration tests.

Ref #3143

Signed-off-by: Alejandro Pedraza <alejandro@buoyant.io>


bin/_helm.sh
```diff
@@ -0,0 +1,34 @@
...
+
+set -eu
+
+# Checks if target/helm exists and it installs it if it doesn't.
+# Use `install_helm true` to also install Tiller.
+install_helm() {
+    bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+    rootdir="$( cd $bindir/.. && pwd )"
+
+    if [ ! -f $rootdir/target/helm ]; then
```
siggy: name the executable with the version number, similar to protoc, that way when we update the version in this script, it'll re-download.

bin/_helm.sh
```diff
@@ -0,0 +1,34 @@
...
+        if [ "$(uname -s)" = "Darwin" ]; then
+            os=darwin
+        else
+            os=linux
```
siggy: maybe switch on arch as well?
```bash
if [ "$(uname -s)" = "Darwin" ]; then
    os=darwin
    arch=amd64
else
    os=linux
    case $(uname -m) in
        x86_64) arch="amd64" ;;
        arm)    dpkg --print-architecture | grep -q "arm64" && arch="arm64" || arch="arm" ;;
    esac
fi
```

bin/_helm.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+set -eu
+
+# Checks if target/helm exists and it installs it if it doesn't.
+# Use `install_helm true` to also install Tiller.
```
siggy: `install_helm` without params fails with `bash: $1: unbound variable`, just confirming that's expected.

alpeb: Good catch, I though I had that covered :-(

alpeb:
@siggy thanks for the feedback; all points agreed upon and implemented :+1:

ihcsim:
The new script looks good :+1:. I think we can remove the existing `bin/helm.sh` and update `travis.yaml` to call this script. Optionally, we can also tackle #3199 as part of this branch by updating `BUILD.md`?

bin/helm
```diff
@@ -0,0 +1,35 @@
...
+        cd "$tmp"
+        curl -Lsf -o "./helm.tar.gz" "$helmcurl"
+        tar zf "./helm.tar.gz" -x "$targetdir"
+        chmod +x "$targetdir/helm"
+    )
+    mv "$tmp/$targetdir/helm" "$helmbin"
+    rm -rf "$tmp"
+fi
+
+$helmbin "$@"
```
siggy: nit: newline at end of file (my editor will automatically end up doing this eventually)

alpeb:
I'm merging this one, and punting the dev workflow notes and pending .travis.yml work to #3199 and #3230

bin/helm
alpeb: oops I forgot to push this one before merging. Will take care of it in #3223

PR #3222, opened 2019-08-08
Remove redundant `HTTPError` cast check in web server

kleimkuhler:

### DIFF
--- a/web/srv/api_handlers.go
+++ b/web/srv/api_handlers.go
@@ -22,7 +22,7 @@ import (
 )
 
 // Control Frame payload size can be no bigger than 125 bytes. 2 bytes are
-// reserverd for the status code when formatting the message.
+// reserved for the status code when formatting the message.
 const maxControlFrameMsgSize = 123
 
 type (
--- a/web/srv/api_handlers.go
+++ b/web/srv/api_handlers.go
@@ -21,6 +21,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// Control Frame payload size can be no bigger than 125 bytes. 2 bytes are
+// reserverd for the status code when formatting the message.
 const maxControlFrameMsgSize = 123
 
 type (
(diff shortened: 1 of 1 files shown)


web/srv/api_handlers.go
```diff
@@ -219,7 +216,7 @@ func (h *handler) handleAPITap(w http.ResponseWriter, req *http.Request, p httpr
 	}
 
 	if messageType != websocket.TextMessage {
-		websocketError(ws, websocket.CloseUnsupportedData, errors.New("MessageType not supported"))
+		websocketError(ws, websocket.CloseUnsupportedData, errors.New("messageType not supported"))
```
kleimkuhler: Removed the capitalization for correct style

web/srv/api_handlers.go
```diff
@@ -243,7 +240,8 @@ func (h *handler) handleAPITap(w http.ResponseWriter, req *http.Request, p httpr
 			// socket with `ClosePolicyViolation` status code so that the error
 			// renders without the error prefix in the banner
 			if httpErr, ok := err.(protohttp.HTTPError); ok && httpErr.Code == http.StatusForbidden {
-				websocketError(ws, websocket.ClosePolicyViolation, err)
+				msg := fmt.Sprintf("Missing authorization, visit %s to remedy", tap.TapRbacURL)
+				websocketError(ws, websocket.ClosePolicyViolation, errors.New(msg))
```
siggy:
```
err := fmt.Errorf("missing authorization, visit %s to remedy", tap.TapRbacURL)
websocketError(ws, websocket.ClosePolicyViolation, err)
```

web/srv/api_handlers.go
```diff
@@ -21,6 +21,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// Control Frame payload size can be no bigger than 125 bytes. 2 bytes are
+// reserverd for the status code when formatting the message.
```
siggy: i think the 2 bytes are for payload size:
https://github.com/gorilla/websocket/blob/6a67f44b69004415802d2e678e483065b8fff6c8/conn.go#L1159-L1162

```
// Control Frame payload size can be no bigger than 125 bytes. 2 bytes are
// reserved for the payload size when formatting the message.
```

kleimkuhler: @siggy `FormatCloseMessage` takes the status code we are closing the socket with and the message. It copies the close code into the first two bytes of the buffer, and then the message into the rest. That message is what Gorilla validates to be no longer than 125 bytes.

Maybe I'm missing something, but it doesn't seem like is related to the size of the payload?

web/srv/api_handlers.go
kleimkuhler: @siggy I like this change, but I think "Missing" should still be capitalized. We display this error with no prefix in the banner, so it's always going to be the first word and may look weird lowercase.

siggy: Agree that capitalizing messages in the banner is preferable. I don't think the code at this layer should be concerned about that. Better to consistently return well-formed Go error messages, and let the presentation layer decide how to render them.

web/srv/api_handlers.go
siggy: Ah you're right, I misread `FormatCloseMessage`.

web/srv/api_handlers.go
kleimkuhler: Yep that is a good point. I'll remove the capitalization and let the open issue on how we want to present this error handle that!

PR #3223, opened 2019-08-09
Add integration test for `helm install`

alpeb:
Ref #3143

Also added `exit_on_err` in `bin/test-run`, for DRYing code for failing
fast after tests (we weren't doing that if install_test.go failed during
the non-upgrade).

Note this adds 3 mins of testing in my laptop (out of a total of 13
mins). Would be great if we could consider running some of these in
parallel clusters in the next integration tests infra revamp :wink:


alpeb:
@siggy Seems we've got RBAC issues with l5d-bot's cluster when running this test. The Helm install is performed by the Tiller pod, which runs under the `kube-system:default` service account, which doesn't have enough privileges. Could you please take a look? I believe setting that SA on par with the `default:default` SA should do the trick. We can also easily have Tiller use a different SA if you deem necessary.

alpeb:
I'm made some changes to simplify error handling. LMKWYT :ear:

bin/test-run
```diff
@@ -83,15 +83,19 @@ function run_upgrade_test() {
 }
 
 function run_helm_test() {
+    kubectl --context=$k8s_context create ns $tiller_namespace
+    kubectl --context=$k8s_context create clusterrolebinding tiller-cluster-admin --clusterrole=cluster-admin --serviceaccount=${tiller_namespace}:default
```
siggy: maybe this is better:
`tiller-cluster-admin` => `${tiller_namespace}:tiller-cluster-admin`

bin/test-run
```diff
@@ -80,17 +74,54 @@ function run_upgrade_test() {
     run_test "$test_directory/install_test.go" -failfast --upgrade-from-version=$stable_version --linkerd-namespace=$stable_namespace
 }
 
+function run_helm_test() {
+    (
+        set -e
+        kubectl --context=$k8s_context create ns $tiller_namespace
+        kubectl --context=$k8s_context create clusterrolebinding ${tiller_namespace}:tiller-cluster-admin --clusterrole=cluster-admin --serviceaccount=${tiller_namespace}:default
+        $helm_path --kube-context=$k8s_context --tiller-namespace=$tiller_namespace init --wait
+        $helm_path --kube-context=$k8s_context --tiller-namespace=$tiller_namespace dependency update $rootdir/charts/linkerd2
```
ihcsim: `$rootdir/charts/linkerd2` can be replaced with `$helm_chart`.

bin/test-run
```diff
@@ -105,16 +136,19 @@ printf "Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\\n" "$linke
...
+#run_upgrade_test "$linkerd_namespace"-upgrade
+#exit_on_err "can't upgrade to version $linkerd_version"
 cleanup
 
-# run the rest of the test suite. `cleanup` is not called if this test failed
-# so that there is a chance to debug the problem
-run_test "$test_directory/install_test.go" -failfast --linkerd-namespace=$linkerd_namespace || exit_code=$?
+run_helm_test
+exit_on_err "error testing Helm"
+cleanup
```
ihcsim: Probably not needed.

bin/test-run
```diff
@@ -105,16 +136,19 @@ printf "Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\\n" "$linke
 # 3. if failed, exit script to avoid leaving behind stale resources which will
 # fail subsequent tests. `cleanup` is not called if this test failed so that
 # there is a chance to debug the problem
-run_upgrade_test "$linkerd_namespace"-upgrade || exit_code=$?
-if [ $exit_code -ne 0 ]; then
-  printf "\\n=== FAIL: Can't upgrade to version $linkerd_version\\n"
-  exit $exit_code
-fi
+#run_upgrade_test "$linkerd_namespace"-upgrade
```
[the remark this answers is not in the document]
alpeb: whoops. Thanks for catching that!

bin/test-run
alpeb: As explained over slack, this is needed because `cleanup` blocks until all the resources are gone, unlike `helm delete`.

PR #3229, opened 2019-08-12
Update install and upgrade code to use the new helm charts

### LINKED ISSUE #3127: Refactor CLI Install Command To Work With 2.5 Helm Chart

ihcsim:
This PR updates the `linkerd install` and `linkerd upgrade` commands to use the new Helm chart.

The following commands were executed to install and upgrade the control plane. Their YAML output are also diff'ed with that of `edge-19.8.2`:
```
bin/linkerd install
bin/linkerd install --ha
bin/linkerd install config
bin/linkerd install control-plane
bin/linkerd install-cni
bin/linkerd install --linkerd-cni-enabled
bin/linkerd install --ignore-cluster --admin-port=7000 --control-port=7001 --controller-log-level=warn --controller-uid=9009 --disable-h2-upgrade=true --enable-external-profiles=true --identity-clock-skew-allowance=60s --identity-issuance-lifetime=3600s --identity-issuer-key-file=fake-tls/key.pem --identity-trust-anchors-file=fake-tls/crt.pem --identity-issuer-certificate-file=fake-tls/crt.pem --identity-trust-domain=cluster.local --image-pull-policy=Always --inbound-port=7002 --omit-webhook-side-effects=false --outbound-port=7003 --proxy-cpu-limit=2 --proxy-cpu-request=1 --proxy-log-level=debug --proxy-memory-limit=256M --proxy-memory-request=64M --proxy-uid=3000 --skip-inbound-ports=3000,3001 --skip-outbound-ports=4000,4001 --registry=gcr.io/runconduit --restrict-dashboard-privileges=true

# inject, auto-inject and manual inject with emojivoto
bin/linkerd inject https://run.linkerd.io/emojivoto.yml 
bin/linkerd inject --manual https://run.linkerd.io/emojivoto.yml 

# upgrade from stable-2.4.0
bin/linkerd upgrade
```

Fixes #3127


charts/linkerd2/templates/proxy-injector-rbac.yaml
```diff
@@ -61,8 +61,8 @@ metadata:
 type: Opaque
 data:
   {{ $ca := genCA (printf "linkerd-proxy-injector.%s.svc" .Namespace) 365 -}}
-  crt.pem: {{ b64enc $ca.Cert }}
-  key.pem: {{ b64enc $ca.Key }}
+  crt.pem: {{ ternary (b64enc $ca.Cert) (b64enc .ProxyInjector.CrtPEM) (empty .ProxyInjector.CrtPEM) }}
```
ihcsim: During upgrade, `.ProxyInjector.CrtPEM` and `.ProxyInjector.KeyPEM` will be assigned the values of the existing cert and key, so that they can be re-used. This change is also added to the service profile validator and tap RBAC templates.

Also, the `.ProxyInjector.CrtPEM` and `.ProxyInjector.KeyPEM` variables are needed to pin the test fixtures in the install and upgrade tests. Otherwise, Helm will always regenerate the cert/key pairs on every test run. If needed, we can revisit whether it's necessary to use Helm to auto-generate the TLS secrets.

charts/linkerd2/templates/sp-validator.yaml
```diff
@@ -41,8 +41,6 @@ spec:
   selector:
     matchLabels:
       {{.ControllerComponentLabel}}: sp-validator
-      {{.ControllerNamespaceLabel}}: {{.Namespace}}
```
ihcsim: Service profile validator already has its own selectors in 2.4. Since these selectors are immutable, adding new match labels here will cause upgrade to fail. Hence, I have to revert this change. The same revert is also added to the proxy injector template.

cli/cmd/install.go
```diff
@@ -71,54 +74,24 @@ type (
...
 		ProxyInjector    *proxyInjectorValues
 		ProfileValidator *profileValidatorValues
 		Tap              *tapValues
+		Proxy            *proxyValues
+		ProxyInit        *proxyInitValues
 	}
 
 	configJSONs struct{ Global, Proxy, Install string }
 
-	resources   struct{ CPU, Memory constraints }
```
ihcsim: These are all moved to the common package at `pkg/charts/template-values.go`.

cli/cmd/install.go
```diff
@@ -142,9 +115,8 @@ type (
 		recordedFlags []*pb.Install_Flag
 
 		// function pointers that can be overridden for tests
-		generateUUID       func() string
-		generateWebhookTLS func(webhook string) (*tlsValues, error)
```
ihcsim: During install, the webhook TLS secrets will be generated by Helm functions. During an upgrade, these secrets will be preserved.

cli/cmd/install.go
```diff
@@ -413,7 +377,7 @@ func (options *installOptions) validateAndBuild(stage string, flags *pflag.FlagS
 	if err := options.validate(); err != nil {
 		return nil, nil, err
 	}
-	options.handleHA()
```
ihcsim: See changes in the `buildValuesWithoutIdentity()` function.

cli/cmd/install.go
```diff
@@ -1071,24 +1126,3 @@ func (idvals *installIdentityValues) toIdentityContext() *pb.IdentityContext {
...
-
-func webhookCommonName(webhook string) string {
-	return fmt.Sprintf("%s.%s.svc", webhook, controlPlaneNamespace)
-}
-
-func webhookSecretName(webhook string) string {
-	return fmt.Sprintf("%s-tls", webhook)
-}
-
-func verifyWebhookTLS(value *tlsValues, webhook string) error {
```
ihcsim: Move to `upgrade.go`, since they are now only relevant during upgrade.

charts/linkerd2/templates/heartbeat.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
     {{.CreatedByAnnotation}}: {{default (printf "linkerd/helm %s" .LinkerdVersion) .CliVersion}}
 spec:
   schedule: "{{.HeartbeatSchedule}}"
+  successfulJobsHistoryLimit: 0
```
alpeb: :+1:

pkg/charts/charts.go
```diff
@@ -39,7 +39,7 @@ func (chart *Chart) Render() (bytes.Buffer, error) {
 		{Name: "charts/partials/templates/_capabilities.tpl"},
 	}
 	if err := filesReader("", partialsFiles); err != nil {
-		return bytes.Buffer{}, nil
+		return bytes.Buffer{}, err
```
alpeb: :blush:

pkg/charts/defaults.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// for non-HA, only interested in values.yaml
+	for _, valuesFile := range valuesFiles {
+		if valuesFile.Name == helmDefaultValuesFile {
+			values, err := chartutil.ReadValues(valuesFile.Data)
+			if err != nil {
+				return nil, err
+			}
+			return setDefaults(values, ha)
+		}
```
alpeb: Those two loops look very similar; how about consolidating with something like this?
```diff
iff --git a/pkg/charts/defaults.go b/pkg/charts/defaults.go
index a713b6fd..9dfb0dd7 100644
--- a/pkg/charts/defaults.go
+++ b/pkg/charts/defaults.go
@@ -1,7 +1,6 @@
 package charts
 
 import (
-       "fmt"
        "time"
 
        "k8s.io/helm/pkg/chartutil"
@@ -79,32 +78,19 @@ func ReadDefaults(chartDir string, ha bool) (*DefaultValues, error) {
                return nil, err
        }
 
-       // if ha, merge values.yaml and values-ha.yaml
-       if ha {
-               values := chartutil.Values{}
-               for _, valuesFile := range valuesFiles {
+       values := chartutil.Values{}
+       for _, valuesFile := range valuesFiles {
+               // if ha, merge values.yaml and values-ha.yaml
+               if ha || valuesFile.Name == helmDefaultValuesFile {
                        v, err := chartutil.ReadValues(valuesFile.Data)
                        if err != nil {
                                return nil, err
                        }
                        values.MergeInto(v)
                }
-
-               return setDefaults(values, ha)
-       }
-
-       // for non-HA, only interested in values.yaml
-       for _, valuesFile := range valuesFiles {
-               if valuesFile.Name == helmDefaultValuesFile {
-                       values, err := chartutil.ReadValues(valuesFile.Data)
-                       if err != nil {
-                               return nil, err
-                       }
-                       return setDefaults(values, ha)
-               }
        }
 
-       return nil, fmt.Errorf("Can't find default Helm values file at %s", chartDir)
+       return setDefaults(values, ha)
 }
```

pkg/charts/defaults.go
```diff
@@ -0,0 +1,265 @@
...
+)
+
+const (
+	helmDefaultValuesFile   = "values.yaml"
+	helmDefaultHAValuesFile = "values-ha.yaml"
+)
+
+// DefaultValues contain all the default variables defined in the values.yaml
+// and values-ha.yaml.
+type DefaultValues struct {
```
alpeb: It seems there is overlap between this struct, `charts.Values` and `installValues` used in `install.go`. The other structs in `charts.Values` already reflect the contents of `values.yaml` (Proxy, init, ports, etc). What do you think about completing `charts.Values` so it completely reflects `values.yaml` and use that everywhere we need to pass values to Helm? This wouldn't affect the injection side of things as that only uses the stuff in `charts.Values`  it needs.

cli/cmd/install.go
```diff
@@ -71,54 +74,24 @@ type (
...
-
-	tlsValues struct {
-		KeyPEM, CrtPEM string
-	}
+	installIdentityValues  charts.Identity
+	proxyInjectorValues    charts.ProxyInjector
+	profileValidatorValues charts.ProfileValidator
+	tapValues              charts.Tap
+	proxyValues            charts.Proxy
+	proxyInitValues        charts.ProxyInit
```
siggy: curious: why redefine these types rather than using the `charts.*` variants?

cli/cmd/install.go
```diff
@@ -190,40 +161,50 @@ Otherwise, you can use the --ignore-cluster flag to overwrite the existing globa
...
 		restrictDashboardPrivileges: false,
 		proxyConfigOptions: &proxyConfigOptions{
 			proxyVersion:           version.Version,
 			ignoreCluster:          false,
-			proxyImage:             defaultDockerRegistry + "/proxy",
-			initImage:              defaultDockerRegistry + "/proxy-init",
+			proxyImage:             defaults.ProxyImageName,
+			initImage:              defaults.ProxyInitImageName,
 			initImageVersion:       version.ProxyInitVersion,
 			dockerRegistry:         defaultDockerRegistry,
```
siggy: curious: why do fields like `initImageVersion` and `dockerRegistry` not come from `charts.ReadDefaults()`? what's the distinction between a default hard-coded here vs. `charts.ReadDefaults()`

cli/cmd/upgrade.go
```diff
@@ -300,27 +315,19 @@ func repairInstall(generateUUID func() string, install *pb.Install) {
...
-		value, err = options.generateWebhookTLS(webhook)
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		value = &tlsValues{
-			KeyPEM: string(secret.Data["key.pem"]),
-			CrtPEM: string(secret.Data["crt.pem"]),
-		}
+	value = &charts.TLS{
```
siggy: you can remove the `var value` above in favor of `value := &charts.TLS{` here.

charts/linkerd2/values.yaml
```diff
@@ -4,7 +4,6 @@
 
 ClusterDomain: &cluster_domain cluster.local
 EnableH2Upgrade: true
-EnablePodAntiAffinity: false
 ImagePullPolicy: &image_pull_policy IfNotPresent
 
 # control plane version. See Proxy section for proxy version
```
siggy: unrelated to this PR, should this be `edge-19.8.2`, or something else?

pkg/charts/defaults.go
```diff
@@ -0,0 +1,265 @@
...
+		ProxyInitMemoryRequest:           proxyInitResourcesMemory["Request"].(string),
+		ProxyLogLevel:                    proxy["LogLevel"].(string),
+		ProxyMemoryLimit:                 proxyResourcesMemory["Limit"].(string),
+		ProxyMemoryRequest:               proxyResourcesMemory["Request"].(string),
+		ProxyOutboundPort:                uint(proxyPorts["Outbound"].(float64)),
+		ProxyUID:                         int64(proxy["UID"].(float64)),
+		WebhookFailurePolicy:             defaults["WebhookFailurePolicy"].(string),
+	}
+
+	if ha {
```
siggy: in the non-ha case above, i liked how your grouped all the `defaults.Table()` calls together and the the assignment into `defaultValues` at the end. would you mind making this consistent with that?

pkg/charts/template-values.go
```diff
@@ -87,5 +90,36 @@ type (
...
+		CrtExpiry           time.Time
+		TLS                 *TLS
+	}
+
+	// ProxyInjector has all the proxy injector's Helm variables
+	ProxyInjector struct {
+		*TLS
+	}
+
+	// ProfileValidator has all the profile validator's Helm variables
```
siggy: similar question to earlier, what's the intent of wrapping `TLS` in a new struct like `ProfileValidator`?

pkg/charts/defaults.go
ihcsim: Good :eyes:! I think the `if ha || valuesFile.Name == helmDefaultValuesFile` is probably not needed. We can just merge all `values*.yaml` into the `values` struct.

charts/linkerd2/values-ha.yaml
```diff
@@ -17,15 +17,18 @@ DestinationResources: *controller_resources
...
+  Memory:
+    Limit: 250Mi
+    Request: 10Mi
 
 # grafana configuration
-GrafanaResources: *controller_resources
+GrafanaResources:
+  CPU: *controller_resources_cpu
+  Memory:
+    Limit: 1024Mi
```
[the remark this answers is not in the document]
ihcsim: Yes, it has already been bumped in [`master`](https://github.com/linkerd/linkerd2/blob/b4b2a44299093cf5ad24f046ce3d4bf423b1cf68/cli/cmd/install.go#L710).

pkg/charts/defaults.go
ihcsim: The thought did cross my mind, but I decided against it for now, mainly because I don't want to deal with multiple layers of type casting. The default values read from the `values*.yaml` will be of primitive types. In order to get them into `charts.Values`, I'll have to convert them into custom types defined in the `pkg/charts` package, only to have them reconverted back to primitive types to be rendered in the templates. I feel we might end up adding more code, than saving.

IMO, it's also reasonable to keep `charts.Values` as an inject-only struct. The context it is used (by its surrounding code) is different from that of the default values. I think at one point, we chatted about whether to name it as `charts.InjectValues`.

If we feel like it's a worthwhile refactoring effort, perhaps we can track it in a separate PR.

cli/cmd/install.go
ihcsim: Good questions. `initImageVersion` and `ProxyVersion` remain as-is because their values need to be updated by the `version` package across builds. These values aren't returned by `charts.ReadDefaults()`, because the static versions in the `values.yaml` and `values-ha.yaml` files are only relevant to `helm install`.

`dockerRegistry` remains as-is because it isn't configurable by Helm user. (Hence, it isn't specified in the `values.yaml` file.) The reason being the way we structured the `installValues` in our original Helm template. E.g. on [`master`](https://github.com/linkerd/linkerd2/blob/b4b2a44299093cf5ad24f046ce3d4bf423b1cf68/chart/templates/controller.yaml#L75), only `{{ControllerImage}}` is configurable by the `install` code, with the docker registry baked into that variable.

I'm open to suggestions on how to make this better and clearer.

cli/cmd/install.go
ihcsim: Initially, I did this with `cli.installIdentityValues` to avoid having to change the signature of multiple install and upgrade unexported functions. I can take a look again to see if it makes sense to removed these types.

charts/linkerd2/values.yaml
ihcsim: Yes, it should. We are yet to figure out a way to inject the new version numbers into these static files (`values.yaml` and `Chart.yaml`) as part of the build.

pkg/charts/defaults.go
alpeb: If we always merge, doesn't that imply that the HA values will always override the non-HA ones?

pkg/charts/defaults.go
ihcsim: I think your intuition is right. I'll like to give this some more thought post-2.5, as the amount of code change will be non-trivial (with the removal of `installValues`). I've created https://github.com/linkerd/linkerd2/issues/3237 to track the refactor work.

pkg/charts/defaults.go
ihcsim: Not quite, because `values-ha.yaml` is only added when `ha` is `true`.

```
	if ha {
		valuesFiles = append(valuesFiles, &chartutil.BufferedFile{
			Name: helmDefaultHAValuesFile,
		})
	}
```

alpeb: Ah indeed :+1:

cli/cmd/install.go
ihcsim: I've created #3237 to revisit this, post 2.5. Ideally, the `pkg/charts` library should provide a clean interface for the CLI and webhook to work with these values types.

pkg/charts/defaults.go
alpeb: Maybe I'm misunderstanding the problem, but take this for example:
```go
for _, valuesFile := range valuesFiles {
		valuesStruct := &Values{}
		if err := yaml.Unmarshal(valuesFile.Data, valuesStruct); err != nil {
			fmt.Printf("error: %s\n", err)
		}
		fmt.Printf("valuesStruct: %#v\n", valuesStruct)
		fmt.Printf("valuesStruct.Issuer.ClockSkewAllowance: %s\n", valuesStruct.Identity.Issuer.ClockSkewAllowance)
		v, err := chartutil.ReadValues(valuesFile.Data)
		if err != nil {
			return nil, err
		}
		values.MergeInto(v)
	}
```
Which outputs:
```
valuesStruct: &charts.Values{Namespace:"linkerd", ClusterDomain:"cluster.local", HighAvailability:false, Identity:(*charts.Identity)(0xc00056d830), Proxy:(*charts.Proxy)(0xc00004d740), ProxyInit:(*charts.ProxyInit)(0xc0002f7780)}
valuesStruct.Issuer.ClockSkewAllowance: 20s
```
So we could replace the call `chartutil.ReadValues(valuesFile.Data)` and do the unmarshalling ourselves with  `yaml.Unmarshal(valuesFile.Data, valuesStruct)` which gives us a well-formed `charts.Values` (pending adding the missing fields), instead of having to deal with the flat `chartutil.Values` map...

PR #3240, opened 2019-08-13
Remove tap from public API and associated test infrastructure

kleimkuhler:
### Summary

After the addition of the tap APIServer, all the logic related to tap in the public API no longer needs to be there. The servers and clients that are created but not used, as well as all the old testing infrastrucure related to tap can be removed.

This deprecates TapByResource and therefore required an update to the protobuf files with `bin/protoc-go.sh`. While the change to deprecate this method was extremely small, a lot of protobuf fils were updated in the process. These changes to the code and protobuf files should probably remain coupled since `TapByResource` is officially deprecated in the public API, but a majority of the additions/deletions are related to those files.

This draft passes `go test` as well as a local run of the integration tests.


kleimkuhler:
The diff size in this PR will go down significantly since it depends on #3241

proto/controller/tap.proto
```diff
@@ -8,5 +8,5 @@ option go_package = "github.com/linkerd/linkerd2/controller/gen/controller/tap";
 
 service Tap {
   rpc Tap(public.TapRequest) returns (stream public.TapEvent) { option deprecated = true; }
-  rpc TapByResource(public.TapByResourceRequest) returns (stream public.TapEvent) {}
```
adleong: Is this actually deprecated?  Isn't this the interface that the tap pod serves? (albeit as proto-over-http instead of gRPC)

siggy:
you'll need to rebase on master again, and then update https://github.com/linkerd/linkerd2/blob/4d01e3720e6d113efc36df52c515f9ad1406bf28/charts/linkerd2/templates/controller.yaml#L83

proto/controller/tap.proto
siggy: As of this PR, it's true that the `TapByResource` endpoint in Tap's gRPC server is still accepting requests. In a followup PR, the plan is to remove that gRPC server.

Current request flow:
CLI -> K8s API Server -> Tap API Server -> Tap gRPC Server -> Proxy

The plan:
CLI -> K8s API Server -> Tap API Server -> Proxy

adleong: Maybe this is splitting hairs but I would say this grpc service definition here defines the contract of the Tap API server.  i.e. it takes a TapByResourceRequest as input (encoded as proto-over-http) and returns a stream of TapEvents (encoded as proto-over-http).

Even though none of the grpc bindings generated from this IDL are used, I still wouldn't call it deprecated because it describes the server's interface.

siggy: It's true that there still exists a server endpoint that takes a proto-over-http `TapByResourceRequest` and returns `TapEvent`'s, but I don't think it accurately describes the server's interface, since it's not a gRPC server, and there is not going to be an endpoint called `TapByResource`.

I agree about documenting our API contracts. In that case, we would document the HTTP interface for the TAP APIService. This could be a starting point:
https://github.com/linkerd/linkerd2/blob/4d01e3720e6d113efc36df52c515f9ad1406bf28/controller/tap/handlers.go#L72-L104

kleimkuhler: @adleong Does this make sense or did you have any more questions about it?

adleong: Yeah.  I guess this is weird either way.  At the least, we should add comments to `TapByResourceRequest` and `TapEvent` to indicate that these structs are used by the tap API, since these structs are otherwise not used in any non-deprecated gRPC service.

kleimkuhler: @adleong Good point--I added some comments that clarify they are only used by the tap APIServer right now. In the future when we are able to re-expose this to authorized users through the gRPC interface, the comments can be updated.

PR #3241, opened 2019-08-13
Update protobuf files

kleimkuhler:
### Summary

Changes from `bin/protoc-go.sh`

An existing [draft PR](https://github.com/linkerd/linkerd2/pull/3240) has a majority of its changes related to protobuf file
updates. In order to separate these changes out into more related components,
this PR updates the generated protobuf files so that #3240 can be rebased off
this and have a more manageable diff.


kleimkuhler:
The diff size in this PR will go down significantly since it depends on #3241

kleimkuhler:
There are no `.proto` changes with this PR; this is just after running `bin/protoc-go.sh`. #3240 has the related `.proto` changes, but after rebased off this branch, its overall diff is much smaller.

siggy:
I think it's good to err on the side of more, smaller PRs rather than single big ones. The goal being easily understandable changes, that can each standalone, pass tests, and be merged to master. Minimizing diff side is a nice side effect of this, but not the primary goal.

All that said, these changes in the `.pb.go` files are bigger than expected, and not related to your diff (I just ran `bin/proto-go.sh` on master and got similarly big changes. It may be related to the recent move to Go Modules. I'll dig in a bit.

siggy:
Looks like we've been dirtying the repo in our Go unit tests in Travis:
https://github.com/linkerd/linkerd2/blob/master/.travis.yml#L37-L39

Issue #3242, opened 2019-08-13
Expand automatic mTLS to all meshed traffic


olix0r:
# moreTLS

## Problem

Linkerd automatically establishes mutual TLS between Kubernetes services that communicate via HTTP. However, this only applies when clients send traffic to a Kubernetes `Service`. When traffic is sent directly between pods, for instance as Prometheus does to scrape metrics, or when non-HTTP protocols are used, Linkerd cannot currently establish trusted, private communication between the pods. This needs to change.

## Goals

1. Automatically make **all** communication private between pods participating in a Linkerd mesh.
2. Provide an audit trail of the TLS status of all connections into or out of the mesh.

### Non-goals

* Implement authorization policies.
* Fail-closed.

## Proposal

* Extend the destination service to support endpoint resolution by IP address;
* Modify the proxy to perform endpoint resolution on IP addresses;
* Introduce TCP resolution into the proxy's outbound TCP forwarding, bypassing kube-proxy.
* Enhance TCP metrics with endpoint metadata, including the SNI name of opaque TLS connections, to support auditability.

#### Destination Service

The `Destination.Get` API is extended to resolve names in the form `IP:PORT`:

1. When the given IP refers to a `Service` IP, its constituent endpoints are returned as they would be if the service's name were used instead of its IP.
2. When the given IP refers to a Pod IP, its metadata is returned.
3. Otherwise, an endpoint is resolved with no additional metadata.

##### Start resolving IP address destinations via the destination service

When the proxy attempts to send an HTTP request without a `Host` header or with such a header referring to an authority containing an IP address instead of a hostname, the proxy will attempt to resolve the IP address through the control plane as it would resolve any Kubernetes service name.

##### Introduce a TCP balancer into outbound TCP forwarding

When the proxy attempt to forward a outbound TCP connection, it will first resolve the original dst address through the `Destination.Get` API and balance connections over the returned set of pods.

##### Expose per-dst TCP stats

The proxy's TCP metrics should be extended to include per-destination metadata so that the TLS status of each outbound TCP connection includes the controller-provided metadata, including TLS status.

grampelberg:
This sounds like a good incremental step to me. I like tackling policy later.

PR #3244, opened 2019-08-13
destination: Reorganize authority-parsing

olix0r:
In preparation for #3242, the destination controller will need to
support a broader set of valid authorities including IP addresses.

This change modifies the destination controller's authority-parsing code
so that the is-this-a-kubernete-service-name decision is decoupled from
parsing of authorities into their consituent parts.

The `Get` API now explicitly handles IP address names, though it
currently fails all such resolutions.


olix0r:
I also changed all references from "pod hostname" to "instance ID" -- host/hostname is a very confusing term in the context of this code.

controller/api/destination/server.go
```diff
@@ -253,48 +284,53 @@ func getHostAndPort(authority string) (string, watcher.Port, error) {
...
-// parseServiceAuthority is a utility function that destructures an authority
-// into a service, port, and optionally a pod hostname.  If the authority does
-// not represent a Kubernetes service, an error is returned.  If no port is
+type InstanceID = string
+
+// parseK8sServiceName is a utility that destructures a Kubernetes serviec hostname into its constituent components.
+//
+// If the authority does not represent a Kubernetes service, an error is returned.
+//
+// If no port is
```
adleong: This function doesn't return a port

olix0r:
Holding off on merging this until stable-2.5.0 is tagged (hopefully tomorrow)

PR #3247, opened 2019-08-13
Label and cleanup helm after integration tests

siggy:
When helm integration tests fail, `bin/test-run` exits prior to calling
`helm_cleanup`, leaving behind a helm namespace and clusterrolebinding.

Update `bin/test-cleanup` to delete any remaining helm resources.


siggy:
@ihcsim Your understanding is correct. It's not ideal that we have cleanup code in two places (`helm_cleanup` in `bin/test-run`, and `bin/test-cleanup`), but it's workable for now.

PR #3248, opened 2019-08-14
Update the Helm build script

ihcsim:
This PR updates the `bin/helm.sh` script to use the pinned version of the Helm CLI, by calling the new `bin/helm` script. The `bin/helm.sh` script is also renamed to `bin/helm-build`.

The Linkerd version in the `charts/linkerd2/values.yaml` file is also updated to the next edge release, `edge-19.8.3`.


charts/linkerd2/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: "v1"
-appVersion: edge-19.8.1
+appVersion: edge-19.8.3
```
siggy: Re: this `appVersion`, and `LinkerdVersion` defined in `values.yaml`, what's the process around changing these? i'm reluctant to set them to versions that have not yet been released, as we won't necessarily know what it's called (`edge-19.8.3`, `edge-19.9.1`, `stable-2.5`, etc).

What about keeping this value as the latest released version, and then rev-ing it when we do release notes / tagging?

bin/helm-build
```diff
@@ -0,0 +1,24 @@
...
+bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rootdir="$( cd $bindir/.. && pwd )"
+
+$bindir/helm lint $rootdir/charts/partials
+
+$bindir/helm dep up $rootdir/charts/linkerd2
+$bindir/helm dep up $rootdir/charts/patch
+$bindir/helm lint --set Identity.TrustAnchorsPEM="fake-trust" --set Identity.Issuer.TLS.CrtPEM="fake-cert" --set Identity.Issuer.TLS.KeyPEM="fake-key" --set Identity.Issuer.CrtExpiry="fake-expiry-date" $rootdir/charts/linkerd2
+
+# bump the Linkerd version in the chart
```
siggy: are there use cases where we'd want to run `helm-build` but not bump the version? for example, if i run `helm-build` on my local checkout of this branch, i get:

```diff
-appVersion: edge-19.8.3
+appVersion: git-bc86453f
...
-LinkerdVersion: &linkerd_version edge-19.8.3
+LinkerdVersion: &linkerd_version git-bc86453f
```

...which i think we don't want? what about decoupling `helm-build` from the version bumps? we could even make version bumps just a documented part of our release process.

charts/linkerd2/Chart.yaml
grampelberg: I would recommend producing the helm chart as a build artifact instead of linking folks to the repo itself. That would allow for updating the app version dynamically and providing what it is based off the tag. Feels like that's part of the work to get to a stable chart as well.

bin/helm-build
ihcsim: Yeah, my initial thought for this and `appVersion` in `Chart.yaml` was to make the version bump a manual step, until we publish the chart. I felt like there was no point of putting a lot of effort into this until we get a chance to iron out our chart publication process. This was a last-minute, minimum-effort addition, which I am happy to remove.

PR #3249, opened 2019-08-14
Add linkerd config subcommand

adleong:
After changing the Linkerd settings on an existing cluster by using `linkerd upgrade` it can be difficult to know what the current cluster settings are.  Reading the configmap is kind of annoying to and it doesn't format well:

```
kubectl -n linkerd get cm/linkerd-config -ojsonpath="{}.data"
map[apiVersion:v1 data:map[global:{"linkerdNamespace":"linkerd" [...]
```

We add a `linkerd config` subcommand that prints the linkerd configmap nicely in json or yaml format.

```
$ linkerd config
global:
  autoInjectContext: null
  clusterDomain: ""
  cniEnabled: false
  [...]
```


grampelberg:
I really don't like this. It is probably okay for now, but it feels like we're papering over the issue that there are some bits which aren't k8s "native".

Why'd we go for serializing the data like this instead of using something a little more yaml-y? Should we schedule some cleanup work there?

adleong:
It's also possible that I'm missing a better workflow.  How do you all inspect the linkerd config?

adleong:
I can grab a specific config like this:

```
$ kubectl -n linkerd get cm/linkerd-config -ojsonpath="{.data.proxy}"
{"proxyImage":{"imageName":"gcr.io/linkerd-io/proxy","pullPolicy":"IfNotPresent"},"proxyInitImage":{"imageName":"gcr.io/linkerd-io/proxy-init","pullPolicy":"IfNotPresent"},"controlPort":{"port":4190},"ignoreInboundPorts":[],"ignoreOutboundPorts":[],"inboundPort":{"port":4143},"adminPort":{"port":4191},"outboundPort":{"port":4140},"resource":{"requestCpu":"","requestMemory":"","limitCpu":"","limitMemory":""},"proxyUid":"2102","logLevel":{"level":"warn,linkerd2_proxy=info"},"disableExternalProfiles":true,"proxyVersion":"stable-2.4.0","proxyInitImageVersion":"v1.0.0"}
```

But I guess because that data is just a string, I can't jsonpath any deeper into that structure.  And I can't change the output format into yaml.

adleong:
I can get the behavior that I want with a script like this:

```bash
GLOBAL=$(kubectl -n linkerd get cm/linkerd-config -ojsonpath="{.data.global}")
INSTALL=$(kubectl -n linkerd get cm/linkerd-config -ojsonpath="{.data.install}")
PROXY=$(kubectl -n linkerd get cm/linkerd-config -ojsonpath="{.data.proxy}")
echo "{\"global\":$GLOBAL,\"install\":$INSTALL,\"proxy\":$PROXY}" | jq .
```

but it would be nice to make this more easily available to Linkerd users somehow...

ihcsim:
Not sure if the version of `kubectl` matters here, with kubectl 1.15

```
kubectl -n linkerd get cm linkerd-config -ojsonpath="{.data.proxy}"
```

pretty prints the JSON:
```
{
  "proxyImage":{
    "imageName":"gcr.io/linkerd-io/proxy",
    "pullPolicy":"IfNotPresent"
  },
  "proxyInitImage":{
    "imageName":"gcr.io/linkerd-io/proxy-init",
    "pullPolicy":"IfNotPresent"
  },
  "controlPort":{
    "port": 4190
  },
  "ignoreInboundPorts":[],
  "ignoreOutboundPorts":[],
  "inboundPort":{
    "port": 4143
  },
  "adminPort":{
    "port": 4191
  },
  "outboundPort":{
    "port": 4140
  },
  "resource":{
    "requestCpu": "",
    "limitCpu": "",
    "requestMemory": "",
    "limitMemory": ""
  },
  "proxyUid": 2102,
  "logLevel":{
    "level": "warn,linkerd2_proxy=info"
  },
  "disableExternalProfiles": true,
  "proxyVersion": "git-159a8621",
  "proxyInitImageVersion": "v1.0.0"
}
```

ihcsim:
Actually, never mind. I just noticed that my control plane was installed by Helm, where the config map was pretty-printed beforehand. Sorry for the noise.

olix0r:
> Actually, never mind. I just noticed that my control plane was installed by Helm, where the config map was pretty-printed beforehand. Sorry for the noise.

Is it feasible to pretty-fmt the JSON as we write the config outside of helm?

olix0r:
Before merging this, we should consider whether we want to dedicate the `linkerd config` to this forever, or whether we'd ever want that command to do something different (like setting config, for instance); and perhaps that would motivate making another set of subcommand under config.

But, I think I would prefer not to add this command as a one-off, and to start with a bit more design about how we _want_ the config workflow to work...

adleong:
That's fine with me, I can go back to using my little script.  Mostly I wanted to get people thinking about this.  If it was confusing and difficult for me to inspect my Linkerd config, it is also probably confusing and difficult for other users.

olix0r:
@adleong I do think we should turn this into an issue though so that we can have a longer design discussion. I think it would even be reasonable to put together an RFC if you have time to think through a proposal!

grampelberg:
Agreed!

PR #3250, opened 2019-08-14
Remove unused code from inject

alenkacz:
This method was unused and I believe it's not really needed because the same kinds are present in `getFreshWorkloadObj` which is used in `parse` which means that no objects that are not injectable will get proxy injected.


alpeb:
Good find @alenkacz :+1:  Thank you very much! :ship: :shipit:

PR #3257, opened 2019-08-14
Make address that controller serves tap on not configurable

kleimkuhler:
### Summary

Instead of a larger refactor that completely removes the gRPC server hop that happens from the APIServer, we will keep that for future work so that a gRPC interface can be exposed to authorized users.

This change instead makes the address that the controller serves tap on not configurable. It will now always be served on localhost so that it is not accidently exposed to the public.

### DIFF
--- a/controller/cmd/tap/main.go
+++ b/controller/cmd/tap/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"crypto/tls"
 	"flag"
-	"net"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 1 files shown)


olix0r:
@adleong Good point! We hadn't considered that...

I think the options are:
1. Add 8088 to the inbound skip-ports so that the proxy never intercepts this traffic. Probably the simplest, though a bit hacky.
2. Modify Tap's HTTP server to take something that accepts the tap interface (not a concrete impl) and just pass it the server impl without binding a port for it all. Probably most-correct, though I'm unsure of the code implications.

kleimkuhler:
Addressed in #3276

PR #3265, opened 2019-08-15
Add changes for `edge-19.8.3`

siggy:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,28 +1,26 @@
 ## edge-19.8.3
 
-This edge release introduces a new `linkerd stat ts` subcommand, enabling
-retrieval of TrafficSplit metrics. It also introduces a
-`Kubernetes cluster monitoring` Grafana dashboard.
+This edge release introduces a new `linkerd stat trafficsplits` subcommand, to
+show traffic split metrics. It also introduces a "Kubernetes cluster monitoring"
+Grafana dashboard.
 
 * CLI
-  * Added TrafficSplit metrics via `linkerd stat ts` subcommand
+  * Added traffic split metrics via `linkerd stat trafficsplits` subcommand
   * Fixed `linkerd uninject` not removing `linkerd.io/inject: enabled`
     annotations
   * Fixed `linkerd stat -h` example commands (thanks @ethan-daocloud!)
+* Controller
+  * Removed Tap from the Public API, now only available via authenticated TAP
+    APIServer
 * Proxy
   * Added `request_handle_us` histogram to measure proxy overhead
-  * Updated Tap server to require identity
-  * Formatted log records consistently
-  * Fixed tap response streams never ending
-  * Fixed Destination service resource leak by hanging up destination
-    resolutions when dropped
-  * Authority label reflects logical dst
-  * Replaced futures_watch with tokio::sync::watch
-  * Adopted Rust 2018
-  * Adopted std::convert::TryFrom
-  * Removed futures-mpsc-lossy
+  * Updated the Tap server to only admit requests from the control plane's Tap
+    controller
+  * Fixed a bug where tap would stop streaming after a short amount of time
+  * Fixed a bug that could cause the proxy to leak service discovery resolutions
+    to the Destination controller
 * Web UI
-  * Added `Kubernetes cluster monitoring` Grafana dashboard with cluster and
+  * Added "Kubernetes cluster monitoring" Grafana dashboard with cluster and
     containers metrics
 * Internal
   * Updated `linkerd install` and `linkerd upgrade` to use Helm charts for
@@ -30,8 +28,6 @@ retrieval of TrafficSplit metrics. It also introduces a
   * Pinned Helm tooling to `v2.14.3`
   * Added Helm integration tests
   * Added container CPU and memory usage to `linkerd-heartbeat` requests
-  * Removed Tap from the Public API, now only available via authenticated TAP
-    APIServer
   * Removed unused inject code (thanks @alenkacz!)
 
 ## edge-19.8.2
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+    annotations
+  * Fixed `linkerd stat -h` example commands (thanks @ethan-daocloud!)
+* Proxy
+  * Added `request_handle_us` histogram to measure proxy overhead
+  * Updated Tap server to require identity
+  * Formatted log records consistently
+  * Fixed tap response streams never ending
+  * Fixed Destination service resource leak by hanging up destination
+    resolutions when dropped
+  * Authority label reflects logical dst
```
olix0r: I would remove this proxy message, as well as those below.

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+`Kubernetes cluster monitoring` Grafana dashboard.
+
+* CLI
+  * Added TrafficSplit metrics via `linkerd stat ts` subcommand
+  * Fixed `linkerd uninject` not removing `linkerd.io/inject: enabled`
+    annotations
+  * Fixed `linkerd stat -h` example commands (thanks @ethan-daocloud!)
+* Proxy
+  * Added `request_handle_us` histogram to measure proxy overhead
+  * Updated Tap server to require identity
```
olix0r: Updated the Tap server to only admit requests from the control plane's Tap controller

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+  * Formatted log records consistently
```
olix0r: omit

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+  * Fixed tap response streams never ending
```
olix0r: Fixed a bug that could prevent taps from completing

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+  * Fixed Destination service resource leak by hanging up destination
+    resolutions when dropped
```
olix0r: Fixed a bug that could cause the proxy to leak service discovery resolutions to the Destination service

olix0r: better: Destination _controller_

CHANGES.md
adleong: "Updated Tap server to only accept tap requests from the tap controller" would be a more functional description of the change

CHANGES.md
adleong: Fixed a bug where tap would stop streaming after a short amount of time

CHANGES.md
adleong: This should be in the past tense.

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+  * Authority label reflects logical dst
+  * Replaced futures_watch with tokio::sync::watch
```
adleong: Not sure if any of these next 4 bullet points are user-facing enough to mention.

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+* Web UI
+  * Added `Kubernetes cluster monitoring` Grafana dashboard with cluster and
+    containers metrics
+* Internal
+  * Updated `linkerd install` and `linkerd upgrade` to use Helm charts for
+    templating
+  * Pinned Helm tooling to `v2.14.3`
+  * Added Helm integration tests
+  * Added container CPU and memory usage to `linkerd-heartbeat` requests
+  * Removed Tap from the Public API, now only available via authenticated TAP
```
adleong: Should this go into a `Controller` section?

CHANGES.md
```diff
@@ -1,3 +1,39 @@
+## edge-19.8.3
+
+This edge release introduces a new `linkerd stat ts` subcommand, enabling
+retrieval of TrafficSplit metrics. It also introduces a
```
wmorgan: s/enabling retrieval of TrafficSplit metrics/to show traffic split metrics/.

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+`Kubernetes cluster monitoring` Grafana dashboard.
```
wmorgan: Feels like more of a " than a `

CHANGES.md
```diff
@@ -1,3 +1,39 @@
...
+
+* CLI
+  * Added TrafficSplit metrics via `linkerd stat ts` subcommand
```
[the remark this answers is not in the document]
wmorgan: I think we should, and let's call it "traffic split" instead of TrafficSplit here too please

CHANGES.md
wmorgan: Roll 'em up

CHANGES.md
```diff
@@ -1,3 +1,35 @@
...
+show traffic split metrics. It also introduces a "Kubernetes cluster monitoring"
+Grafana dashboard.
+
+* CLI
+  * Added traffic split metrics via `linkerd stat trafficsplits` subcommand
+  * Fixed `linkerd uninject` not removing `linkerd.io/inject: enabled`
+    annotations
+  * Fixed `linkerd stat -h` example commands (thanks @ethan-daocloud!)
+* Controller
+  * Removed Tap from the Public API, now only available via authenticated TAP
```
wmorgan: How about: Removed old non-authenticated endpoints for `linkerd tap`, which is now only available via the authenticated API?

CHANGES.md
```diff
@@ -1,3 +1,35 @@
...
+    APIServer
+* Proxy
+  * Added `request_handle_us` histogram to measure proxy overhead
+  * Updated the Tap server to only admit requests from the control plane's Tap
```
wmorgan: How are we capitalizing tap? I see three different versions in these notes

CHANGES.md
siggy: Actually both that wording and your suggestion are not quite true. They will be in stable-2.5. For now changing this to:
`Removed unauthenticated Tap from the Public API`

wmorgan: Just... make the capitalization make me happy

PR #3194, opened 2019-08-05
Check for Namespace level config override annotations

Pothulapati:
Fixes #2499 

This PR makes the `proxy-injector` check for Namespace Level Override Config annotations after checking the Pod level override config annotations before going with the default values. 

I tried with a Cluster and it considers the annotations of the namespace before going with the defaults.

@ihcsim @grampelberg 

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

PR #3272, opened 2019-08-15
Add changes for `edge-19.8.4`

siggy:
Depends on #3276

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+## edge-19.8.4
+
+This edge release is a release candidate for `stable-2.5`.
+
+* CLI
+  * Introduced `--use-wait-flag` to the `linkerd install-cni` command, to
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,14 @@
+## edge-19.8.4
+
+This edge release is a release candidate for `stable-2.5`.
+
+* CLI
+  * Introduced `--use-wait-flag` to the `linkerd install-cni` command, to
+    configure the CNI plugin to use the `-w` flag for `iptables` commands
+* Controller
+  * Disabled the tap gRPC server listener. All tap requests now require RBAC
+    authentication and authorization
```
alpeb: #3194 got missed in the last edge, so you might wanna add it here. It's about supporting the proxy config override annotations at the namespace level.

siggy: Good catch. I'm going to add this under `edge-19.8.3` since that's when the feature was introduced.

PR #3275, opened 2019-08-16
Add support for a node selector in the Helm chart.

### LINKED ISSUE #2851: Support NodeSelector for linkerd control plane
## Feature Request

Support configuring the NodeSelector for the linkerd control plane during `linkerd install` and `linkerd upgrade`.

We use dedicated node pools for deploying specific workloads into. We'd like to control were the linkerd control plane gets deployed on, to avoid linkerd using node pools dedicated for specific kind of workloads (i.g. ingress, data processing nodes, etc).

Optionally allow to configure a node selector for the control plane installation.

Alternatively I can modify the resulting manifests from linkerd install, but this means we have to have an extra (potentially manual) step during install / upgrade.

```bash
linkerd install --node-selector=node-role.kubernetes.io/linkerd=true | kubectl apply -f -
```

bmcstdio:
This PR adds support for specifying a node selector for control-plane components in the Helm chart, as requested in #2851.

Closes #2851.

### DIFF
--- a/charts/linkerd2/templates/_nodeselector.tpl
+++ b/charts/linkerd2/templates/_nodeselector.tpl
@@ -1,0 +1,4 @@
+{{- define "linkerd.node-selector" -}}
+nodeSelector:
+{{- toYaml .NodeSelector | trim | nindent 2 }}
+{{- end -}}
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -53,6 +53,7 @@ spec:
         {{.ControllerNamespaceLabel}}: {{.Namespace}}
         {{- include "partials.proxy.labels" .Proxy | nindent 8}}
     spec:
+      {{- include "linkerd.node-selector" . | nindent 6 }}
       {{- if .EnablePodAntiAffinity -}}
       {{- $local := dict "Component" "controller" "Label" .ControllerComponentLabel -}}
       {{- include "linkerd.pod-affinity" $local | nindent 6 -}}
--- a/charts/linkerd2/values.yaml
+++ b/charts/linkerd2/values.yaml
@@ -132,3 +132,8 @@ LinkerdNamespaceLabel: linkerd.io/is-control-plane
 # - The external tool needs to create the namespace with the label:
 #     linkerd.io/is-control-plane: "true"
 InstallNamespace: true
+
+# Node selection constraints for control-plane components
+# https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector.
+NodeSelector:
+  beta.kubernetes.io/os: linux
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -122,6 +122,7 @@ var (
 		"templates/_affinity.tpl",
 		"templates/_config.tpl",
 		"templates/_helpers.tpl",
+		"templates/_nodeselector.tpl",
 		"templates/config.yaml",
 		"templates/identity.yaml",
 		"templates/controller.yaml",
--- a/charts/linkerd2/templates/grafana.yaml
+++ b/charts/linkerd2/templates/grafana.yaml
@@ -113,6 +113,7 @@ spec:
         {{.ControllerNamespaceLabel}}: {{.Namespace}}
         {{- include "partials.proxy.labels" .Proxy | nindent 8}}
     spec:
+      {{- include "linkerd.node-selector" . | nindent 6 }}
       containers:
       - env:
         - name: GF_PATHS_DATA
--- a/charts/linkerd2/templates/heartbeat.yaml
+++ b/charts/linkerd2/templates/heartbeat.yaml
@@ -27,6 +27,7 @@ spec:
           annotations:
             {{.CreatedByAnnotation}}: {{default (printf "linkerd/helm %s" .LinkerdVersion) .CliVersion}}
         spec:
+          {{- include "linkerd.node-selector" . | nindent 10 }}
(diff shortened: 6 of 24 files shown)


grampelberg:
Now that the helm chart is a real installation method, I kinda like leaving this as a helm only feature and not wiring it through to `install`. WDYT @ihcsim @alpeb. Does that work for your use case @bmcstdio?

bmcstdio:
@grampelberg I don't really have a use case for this at the moment, I am just new to Linkerd and felt like getting to know the codebase and contributing by looking at open issues. 🙂

grampelberg:
Awesome! Thank you for the PR =) We've got a bunch of awesome swag for first time contributors if you're up for filling out the [survey](https://docs.google.com/forms/d/19GLBNfu8XnIETrwEI3ijoWlocAiDp3-j7CW3l4cTnyc/viewform?edit_requested=true).

bmcstdio:
Awesome, done - thank you! 🙂

alpeb:
I agree with @grampelberg in that we should discourage the addition of options to the CLI unless they cover some obvious and often requested case, and users should instead be directed down the Kustomize or Helm chart road for endless customization options :scroll:

charts/linkerd2/templates/controller.yaml
```diff
@@ -72,6 +72,10 @@ spec:
         {{.ControllerNamespaceLabel}}: {{.Namespace}}
         {{- include "partials.proxy.labels" .Proxy | nindent 8}}
     spec:
+      {{- if .NodeSelector }}
```
grampelberg: It'd be nice to DRY this up by having it be a template function. Check out how `_affinity.tpl` is doing it. Should collapse all of this down to a single call.

cli/cmd/install.go
```diff
@@ -419,6 +422,9 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 		&options.highAvailability, "ha", options.highAvailability,
 		"Enable HA deployment config for the control plane (default false)",
 	)
+	flags.StringVar(&options.nodeSelectorString, "node-selector", options.nodeSelectorString,
```
grampelberg: I would rather not expose this functionality through `install`.

charts/linkerd2/values.yaml
```diff
@@ -124,3 +124,5 @@ ProxyInjectDisabled: disabled
 ControllerComponentLabel: linkerd.io/control-plane-component
 ControllerNamespaceLabel: linkerd.io/control-plane-ns
 LinkerdNamespaceLabel: linkerd.io/is-control-plane
+
+NodeSelector: {}
```
grampelberg: Mind adding a comment here to explain what it is and what it does?

cpretzer:
@bmcstdio have you tried the helm installation process yet?

If you have, then are you okay with adding these options only for helm install, rather than as `install` and `upgrade` options?

bmcstdio:
@cpretzer sorry, I didn't have time to test it yet—and haven't had the time to resume work on this either.

charts/linkerd2/templates/controller.yaml
bmcstdio: Done!

charts/linkerd2/values.yaml
bmcstdio: Sure, done!

cli/cmd/install.go
bmcstdio: Sure, removed!

bmcstdio:
@cpretzer @grampelberg I've made the requested changes, please let me know what your thoughts are! :)

charts/linkerd2/templates/_nodeselector.tpl
```diff
@@ -0,0 +1,8 @@
+{{- define "linkerd.node-selector" }}
+nodeSelector:
+{{- if .NodeSelector }}
+{{- toYaml .NodeSelector | trim | nindent 2 }}
+{{- else }}
+  beta.kubernetes.io/os: linux
```
grampelberg: WDYT about just setting the value to this and getting rid of the if/else?

bmcstdio: I agree :+1:

bmcstdio:
> For extra credit, adding a test would be amazing.

Could you please guide me through that?

grampelberg:
@ihcsim you wouldn't happen to have a pointer to one of the tests?

ihcsim:
@grampelberg @bmcstdio The unit tests [are failing](https://github.com/linkerd/linkerd2/pull/3275/checks?check_run_id=221387315) for this PR. If you run `go test ./cli/... --update`, it should automatically update all the test fixtures for you. (Note that `go test ./cli/... --update` will fail the first time, but if you re-run the same command again, the tests will pass.)

btw, the node selector isn't a HA-only option, right?

cpretzer:
@bmcstdio thanks for working on this even while you're busy. I'd love to get this pulled in and I'm running the `test` commands that @ihcsim mentioned.

If all goes well, I'll update this ticket and it should be an easy change for you to rebase and commit.

bmcstdio:
> btw, the node selector isn't a HA-only option, right?

@ihcsim right, I don't think it is/should be.

---

@cpretzer @grampelberg @ihcsim I've rebased against `master` and fixed the unit tests.

bmcstdio:
@alpeb @cpretzer @grampelberg @ihcsim can we get this merged then, please? 🙂

bmcstdio:
@alpeb @cpretzer @grampelberg @ihcsim awesome - can we get this merged then, please? 🙂

PR #3276, opened 2019-08-16
Remove gRPC tap server listener from controller

kleimkuhler:
### Summary

As an initial attempt to secure the connection from clients to the gRPC tap
server on the tap Pod, the tap `addr` only listened on localhost.

As @adleong pointed out #3257, this was not actually secure because the inbound
proxy would establish a connection to localhost anyways.

This change removes the gRPC tap server listener and changes `TapByResource`
requests to interface with the server object directly.

From this, we know that all `TapByResourceRequests` have gone through the tap
APIServer and thus authorized by RBAC.

### Details

[NewAPIServer](https://github.com/linkerd/linkerd2/blob/ef90e0184f238cbe79987a84f36d4eb91cbcda46/controller/tap/apiserver.go#L25-L26) now takes a [GRPCTapServer](https://github.com/linkerd/linkerd2/blob/f6362dfa805de9a009188014256ecd66e7dc3bfc/controller/tap/server.go#L33-L34) instead of a `pb.TapClient` so that
`TapByResource` requests can interact directly with the [TapByResource](https://github.com/linkerd/linkerd2/blob/f6362dfa805de9a009188014256ecd66e7dc3bfc/controller/tap/server.go#L49-L50) method.

`GRPCTapServer.TapByResource` now makes a private [grpcTapServer](https://github.com/linkerd/linkerd2/blob/ef90e0184f238cbe79987a84f36d4eb91cbcda46/controller/tap/handlers.go#L373-L374) that satisfies
the [tap.TapServer](https://godoc.org/github.com/linkerd/linkerd2/controller/gen/controller/tap#TapServer) interface. Because this interface is satisfied, we can interact
with the tap server methods without spawning an additional listener.

### DIFF
--- a/controller/tap/apiserver.go
+++ b/controller/tap/apiserver.go
@@ -22,8 +22,13 @@ type apiServer struct {
 }
 
 // NewAPIServer creates a new server that implements the Tap APIService.
-// func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, client pb.TapClient, disableCommonNames bool) (*http.Server, net.Listener, error) {
-func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, tapServer *server, disableCommonNames bool) (*http.Server, net.Listener, error) {
+func NewAPIServer(
+	addr string,
+	cert tls.Certificate,
+	k8sAPI *k8s.API,
+	grpcTapServer *GRPCTapServer,
+	disableCommonNames bool,
+) (*http.Server, net.Listener, error) {
 	clientCAPem, allowedNames, usernameHeader, groupHeader, err := apiServerAuth(k8sAPI)
 	if err != nil {
 		return nil, nil, err
@@ -43,7 +48,7 @@ func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, tapServer 
 		k8sAPI:         k8sAPI,
 		usernameHeader: usernameHeader,
 		groupHeader:    groupHeader,
-		tapServer:      *tapServer,
+		grpcTapServer:  *grpcTapServer,
 		log:            log,
 	}
 
--- a/controller/tap/apiserver.go
+++ b/controller/tap/apiserver.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
-	pb "github.com/linkerd/linkerd2/controller/gen/controller/tap"
 	"github.com/linkerd/linkerd2/controller/k8s"
 	"github.com/linkerd/linkerd2/pkg/prometheus"
 	"github.com/sirupsen/logrus"
@@ -23,7 +22,8 @@ type apiServer struct {
 }
 
 // NewAPIServer creates a new server that implements the Tap APIService.
-func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, client pb.TapClient, disableCommonNames bool) (*http.Server, net.Listener, error) {
+// func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, client pb.TapClient, disableCommonNames bool) (*http.Server, net.Listener, error) {
+func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, tapServer *server, disableCommonNames bool) (*http.Server, net.Listener, error) {
 	clientCAPem, allowedNames, usernameHeader, groupHeader, err := apiServerAuth(k8sAPI)
 	if err != nil {
 		return nil, nil, err
@@ -43,7 +43,7 @@ func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, client pb.
 		k8sAPI:         k8sAPI,
 		usernameHeader: usernameHeader,
 		groupHeader:    groupHeader,
-		tapClient:      client,
+		tapServer:      *tapServer,
 		log:            log,
 	}
 
--- a/controller/tap/apiserver.go
+++ b/controller/tap/apiserver.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
+	"github.com/linkerd/linkerd2/controller/gen/controller/tap"
 	"github.com/linkerd/linkerd2/controller/k8s"
 	"github.com/linkerd/linkerd2/pkg/prometheus"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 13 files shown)


controller/tap/server.go
```diff
@@ -30,23 +29,24 @@ const requireIDHeader = "l5d-require-id"
...
 
-type (
-	server struct {
-		tapPort             uint
-		k8sAPI              *k8s.API
-		controllerNamespace string
-	}
-)
+// GRPCTapServer describes the gRPC server implementing tap.Tap_TapByResourceServer
+type GRPCTapServer struct {
```
kleimkuhler: I decided to make this public because of lint warnings when the type is returned outside the package in [main.go](https://github.com/linkerd/linkerd2/blob/f6362dfa805de9a009188014256ecd66e7dc3bfc/controller/cmd/tap/main.go#L49-L50)

controller/tap/handlers.go
```diff
@@ -363,3 +343,38 @@ func renderJSONError(w http.ResponseWriter, err error, status int) {
 	w.WriteHeader(status)
 	w.Write(rsp)
 }
+
+type serverStream struct {
```
kleimkuhler: The implementation here reflects a previous revision of the tap server in [stable-2.4.0](https://github.com/linkerd/linkerd2/blob/53e589890ddb0283be794e71ebe7cba250b02a11/controller/api/public/http_server.go#L317-L318)

go.mod
```diff
@@ -60,6 +60,7 @@ require (
 	github.com/projectcalico/libcalico-go v1.7.3
 	github.com/prometheus/client_golang v0.9.2
 	github.com/prometheus/common v0.0.0-20181126121408-4724e9255275
+	github.com/runconduit/conduit v18.9.1+incompatible
```
adleong: whoa, how did this sneak in here??

controller/tap/server.go
```diff
@@ -30,23 +29,24 @@ const requireIDHeader = "l5d-require-id"
...
 const defaultMaxRps = 100.0
 
-type (
-	server struct {
-		tapPort             uint
-		k8sAPI              *k8s.API
-		controllerNamespace string
-	}
-)
+// GRPCTapServer describes the gRPC server implementing tap.Tap_TapByResourceServer
```
adleong: It's worth considering if we want to continue having this implement the TapByResource gRPC interface.  Since we only call these methods directly, it's not strictly necessary to continue to implement that interface.  Continuing to implement the interface would make it easier in the future if we ever were to split this back out into a gRPC server again.  On the other hand, breaking that interface would allow us to greatly simplify this code and remove the need for the ServerStream boilerplate.

My preference would be to break the interface and simplify the code, but it's worth considering the tradeoffs.

controller/tap/server.go
```diff
@@ -30,23 +29,24 @@ const requireIDHeader = "l5d-require-id"
...
 
-func (s *server) Tap(req *public.TapRequest, stream pb.Tap_TapServer) error {
+// Tap is deprecated, use TapByResource.
+func (s *GRPCTapServer) Tap(req *public.TapRequest, stream pb.Tap_TapServer) error {
 	return status.Error(codes.Unimplemented, "Tap is deprecated, use TapByResource")
 }
 
-func (s *server) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_TapByResourceServer) error {
+// TapByResource taps all resources matched by the request object.
+func (s *GRPCTapServer) TapByResource(req *public.TapByResourceRequest, stream pb.Tap_TapByResourceServer) error {
```
adleong: if we were to break the gRPC interface, this could be simplified by taking a TapEvent channel instead of a TapByResourcesServer.

controller/tap/apiserver.go
```diff
@@ -23,7 +22,13 @@ type apiServer struct {
 }
 
 // NewAPIServer creates a new server that implements the Tap APIService.
-func NewAPIServer(addr string, cert tls.Certificate, k8sAPI *k8s.API, client pb.TapClient, disableCommonNames bool) (*http.Server, net.Listener, error) {
+func NewAPIServer(
+	addr string,
+	cert tls.Certificate,
+	k8sAPI *k8s.API,
+	grpcTapServer *GRPCTapServer,
```
olix0r: can this depend on an interface? or does it need to depend on the concrete GRPCTapServer type?

controller/tap/server.go
adleong: @olix0r do you have thoughts about maintaining this interface?

olix0r: @adleong I'm sorry, there are like 3 interfaces at play here. Which interface are we talking about changing? What uses it?

adleong: I think this comment should say `pb.TapServer` instead of `tap.Tap_TapByResourceServer`

adleong: GRPCTapServer implements the TapServer interface which is autogenerated from the (deprecated) protobuf service definition.  Since this is not used as a gRPC server, nothing uses this interface.  

My suggestion is that we change the signature of TapByResource to be more ergonomic (for example by taking a channel instead of a `pb.Tap_TapByResourceServer`) but this would mean that we no longer implement the TapServer interface.

go.mod
kleimkuhler: Wow now clue. It's gone!

controller/tap/handlers.go
```diff
@@ -363,3 +343,38 @@ func renderJSONError(w http.ResponseWriter, err error, status int) {
...
+	w   protohttp.FlushableResponseWriter
+	req *http.Request
+	log *logrus.Entry
```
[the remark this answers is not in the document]
kleimkuhler: @siggy I meant to include this in the `severStream` creation:

```
tapServer := tapByResourceServer{serverStream{w: flushableWriter, req: req, log: h.log}}
```

Thoughts on that option?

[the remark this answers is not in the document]
siggy: sure that lgtm.

controller/tap/apiserver.go
kleimkuhler: I changed this to expect a `pb.TabServer` instead of a concrete type.

controller/tap/server.go
kleimkuhler: In order to keep the existing gRPC interface for use in the future, we're going to keep these around.

controller/tap/server.go
kleimkuhler: Comment address [here](https://github.com/linkerd/linkerd2/pull/3276#discussion_r314851135)

controller/tap/handlers.go
```diff
@@ -363,3 +344,37 @@ func renderJSONError(w http.ResponseWriter, err error, status int) {
 	w.WriteHeader(status)
 	w.Write(rsp)
 }
+
+// serverStream and tapByResourceServer provide functionality that satisfy the
```
siggy: you can remove mention of `tapByResourceServer`

PR #3291, opened 2019-08-20
Merge the CLI 'installValues' type with Helm 'Values' type

### LINKED ISSUE #3237: Revisit the different 'Values' types in the CLI install code
Following #3229, the CLI code and Helm library (`pkg/charts`) added 3 `*Values` types to capture CLI options and Helm variables. The purpose of this issue is to revisit these types and come up with a clean, reusable interface in the `pkg/charts` library, so that the CLI and webhook can consume these types.

`cmd.installValues` - this struct consists of fields that are interpolated in the control plane Helm chart when the user runs the `linkerd install` command. It consists of both configurable and non-configurable fields. They are primarily primitive types suitable for rendering the Helm templates.

`charts.DefaultValues` - this struct consists of fields that are mapped directly to the `charts/linkerd2/values.yaml` and `charts/linkerd2/values-ha.yaml` files. These default values replaced the previously hard-coded defaults found in the CLI code. Any options provided to the `linkerd install` command will override these values. They are primarily primitive types suitable for rendering the Helm templates.

`charts.Values` - this struct consists of fields that correspond to proxy and proxy-init properties that are interpolated in the `partials` chart. These fields are mainly relevant to proxy injection, depended by the `pkg/inject` library. This struct is made up of a number of custom types.

ihcsim:
This PR merges the `cmd.installValues` type with the `charts.Values` type. This merge ensures that all the variables and defaults defined in the CLI and Helm chart are found in one place (i.e. `pkg/charts/values.go`). It also moves many of the Helm-related code out of the CLI package.

Also, added unit test to validate the YAML output generated by the Helm renderer, using defaults from the `values.yaml` file.

Fixes #3237.


cli/cmd/install_helm_test.go
```diff
@@ -0,0 +1,207 @@
...
+			{Name: "templates/config.yaml"},
+			{Name: "templates/identity.yaml"},
+			{Name: "templates/controller.yaml"},
+			{Name: "templates/heartbeat.yaml"},
+			{Name: "templates/web.yaml"},
+			{Name: "templates/prometheus.yaml"},
+			{Name: "templates/grafana.yaml"},
+			{Name: "templates/proxy-injector.yaml"},
+			{Name: "templates/sp-validator.yaml"},
+			{Name: "templates/tap.yaml"},
```
alpeb: What do you think about extracting this list from here and from `render` in `install.go` and putting it in some common place under the `cmd` package? Would of course require some helper function to account for the different sets of templates those two callers require, and also the different types (`pb.Template` vs (`chartutil.BufferedFile`)

pkg/charts/values.go
```diff
@@ -0,0 +1,250 @@
...
+func NewValues(ha bool) (*Values, error) {
+	chartDir := fmt.Sprintf("%s/", helmDefaultChartDir)
+	v, err := readDefaults(chartDir, ha)
+	if err != nil {
+		return nil, err
+	}
+
+	v.CliVersion = k8s.CreatedByAnnotationValue()
+	v.ProfileValidator = &ProfileValidator{TLS: &TLS{}}
+	v.Proxy.Component = k8s.Deployment // only Deployment workloads are injected
```
alpeb: How does this work? Isn't `Proxy.Component` supposed to be something like `linkerd-*`?

cli/cmd/install_helm_test.go
```diff
@@ -0,0 +1,207 @@
...
+		goldenFileName = "install_helm_output.golden"
+		ha             = false
+		namespace      = "linkerd-dev"
+	)
+
+	// read the control plane chart and its defaults from the local folder.
+	// override certain defaults with pinned values.
+	// use the Helm lib to render the templates.
+	// the golden file is generated using the following `helm template` command:
+	// helm template --set Identity.TrustAnchorsPEM="test-crt-pem" --set Identity.Issuer.TLS.CrtPEM="test-crt-pem" --set Identity.Issuer.TLS.KeyPEM="test-key-pem" charts/linkerd2  --set Identity.Issuer.CrtExpiry="Jul 30 17:21:14 2020" --set ProxyInjector.KeyPEM="test-proxy-injector-key-pem" --set ProxyInjector.CrtPEM="test-proxy-injector-crt-pem" --set ProfileValidator.KeyPEM="test-profile-validator-key-pem" --set ProfileValidator.CrtPEM="test-profile-validator-crt-pem" --set Tap.KeyPEM="test-tap-key-pem" --set Tap.CrtPEM="test-tap-crt-pem" --set LinkerdVersion="linkerd-version"  > cli/cmd/testdata/install_helm_output.golden
```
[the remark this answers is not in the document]
ihcsim: Yeah, I have to manually reorder the `helm template` YAML output. I *think* it's because `renderutil.Render()` returns a `map`. So whenever there is a `for` loop either in our code or the Helm lib to iterate that map, the order of the YAML changes. Hence, I'm not able to use something like the [FakeClient.InstallReleaseFromChart()](https://godoc.org/k8s.io/helm/pkg/helm#FakeClient.InstallReleaseFromChart) to generate the YAML in the test (i.e. less code for us), because the ordering will always be [different](https://github.com/helm/helm/blob/df5ca23a418199c1b2af3abfc420f8339d0b9b8e/pkg/helm/fake.go#L384-L390).

I added this comment to keep a record of all the `--set` options used to generate the golden file.

cli/cmd/install_helm_test.go
ihcsim: Good idea. Will adding the filepath to a string slice work for now, and let the callers decide what types they need?

At first, I was hoping to move all these templates references to `charts.Chart`, as it makes sense that a `Chart` instance knows where to find its templates. But then, that will require more code change in `Chart` to support multi-stage templates and some of the `inject` code too. So I decided not to go down that route for now.

pkg/charts/values.go
ihcsim: ha... good :eyes:. I think this is obsolete code from the previous code. The `Proxy.Component` will always be set in the chart templates. E.g., in the [`controller.yaml`](https://github.com/linkerd/linkerd2/blob/d4cd8add3aea39c40a8c7e68fbbe144b2105c5c1/charts/linkerd2/templates/controller.yaml#L46). I can remove it.

I just did a test with `linkerd install` and `helm install`; both of them produce:
```
  labels:
    linkerd.io/control-plane-component: proxy-injector
    linkerd.io/proxy-deployment: linkerd-proxy-injector
```

Issue #3253, opened 2019-08-14
Better logging for why injection is skipped


charlie-hadden:
### What problem are you trying to solve?

Whilst setting up linkerd for the first time I was seeing the following in the proxy-injector logs:

```
time="2019-08-13T14:53:59Z" level=info msg="received admission review request 2ef3cde0-bdda-11e9-bbf7-762a488f8e4f"
time="2019-08-13T14:54:00Z" level=info msg="received pod/production-6cd4bff898-"
time="2019-08-13T14:54:00Z" level=info msg="skipped pod/production-6cd4bff898-"
```

In my case the actual issue was that I had the annotations set on my deployment rather than the pods themselves, but I only found that out after discussing it in slack.

### How should the problem be solved?

Adding some logging around _why_ the injection got skipped would have saved me quite a bit of time. Just a note that it was missing the annotation would have been great.

grampelberg:
@alpeb @ihcsim any ideas on ways to get this a little more "native". While the first pass should be to add logging, WDYT about emitting a k8s event for every injection with a reason that is attached to the pod and parent resources about why injection happened or didn't? It feels like that'd be an easy way to spot whether the injector is being called at all, if the event isn't there then it didn't get called.

alpeb:
Yep, we definitely should start doing that.

>that is attached to the pod and parent resources

Something like the event's `source` being the linkerd-proxy-injector, the `involvedObject` the pod being injected, and the `related` object the pod's parent workload?

grampelberg:
I know almost nothing about k8s events. I would like `kubectl describe deploy/foobar` to have something there from `linkerd-proxy-injector` that says "skipped adding proxy because of X" or "successfully added linkerd proxy".

Looking at the schema I'm getting back from `-o yaml`, it looks like there's also a `reportingComponent` and `reportingInstance` that I'm not sure how that falls out in the `kubectl` UI.

ihcsim:
Feels like we can do more with capturing this as metrics than events, per https://github.com/linkerd/linkerd2/issues/3060.

grampelberg:
@ihcsim metrics won't be a part of `kubectl describe` though, right?

wmorgan:
@alpeb how are you planning to solve this?

alpeb:
> @alpeb how are you planning to solve this?

[...]

PR #2386, opened 2019-02-26
Lessen klog for security

olix0r:
We currently set klog to maximum verbosity when debug logging is
enabled. This causes control plane components, however, to log their
serviceaccount tokens, leaking secret information into logs.

By setting the klog level to 8, we avoid this logging.

Fixes #2383

PR #3315, opened 2019-08-22
Stop ignoring client-go log entries

alpeb:
Pipe klog output into logrus. Not doing this avoids us from seeing
client-go log entries, for some reason I don't understand.

To enable, `--controller-log-level` must be `debug`.

This was discovered while trying to debug sending events for #3253.

I added an integration test that fails when this piping is not in place.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -44,6 +44,8 @@ func setLogLevel(logLevel string) {
 		flag.Set("stderrthreshold", "INFO")
 		flag.Set("logtostderr", "true")
 		flag.Set("v", "6") // At 7 and higher, authorization tokens get logged.
+		// pipe klog entries to logrus
+		klog.SetOutput(log.StandardLogger().Writer())
 	}
 }
 
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -489,6 +489,8 @@ func TestCheckProxy(t *testing.T) {
 func TestLogs(t *testing.T) {
 	controllerRegex := regexp.MustCompile("level=(panic|fatal|error|warn)")
 	proxyRegex := regexp.MustCompile(fmt.Sprintf("%s (ERR|WARN)", k8s.ProxyContainerName))
+	clientGoRegex := regexp.MustCompile("client-go@")
+	hasClientGoLogs := false
 
 	for deploy, spec := range linkerdDeployReplicas {
 		deploy := strings.TrimPrefix(deploy, "linkerd-")
@@ -537,9 +539,15 @@ func TestLogs(t *testing.T) {
 							}
 						}
 					}
+					if clientGoRegex.MatchString((line)) {
+						hasClientGoLogs = true
+					}
 				}
 			})
 		}
+	}
+	if !hasClientGoLogs {
+		t.Errorf("Didn't find any client-go entries")
 	}
 }
 


olix0r:
We should validate that this logging does not introduce the logging of (Kubernetes) secrets when set to its most verbose level... This was a problem in the past, and I seem to recall that the lack of logging here was a saving grace preventing us from logging auth credentials.

pkg/flags/flags.go
```diff
@@ -44,6 +44,8 @@ func setLogLevel(logLevel string) {
 		flag.Set("stderrthreshold", "INFO")
 		flag.Set("logtostderr", "true")
 		flag.Set("v", "6") // At 7 and higher, authorization tokens get logged.
+		// pipe klog entries to logrus
+		klog.SetOutput(log.StandardLogger().Writer())
```
siggy: FWIW i tried setting `-v=10` on the public-api container (which worked!) but didn't seem to affect the log output. i think because we're doing `flag.Set("v", "6")` after calling `flag.Parse()`, we're overriding the value.
```yaml
containers:
- args:
  - public-api
  - -prometheus-url=http://linkerd-prometheus.linkerd.svc.cluster.local:9090
  - -controller-namespace=linkerd
  - -log-level=debug
  - -v=10
```

```bash
$ kubectl -n linkerd logs -f linkerd-controller-59db74787d-w44fz public-api
time="2019-08-23T01:19:39Z" level=info msg="running version git-42ee2e2f"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.103678       1 merged_client_builder.go:122] Using in-cluster configuration"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.113756       1 round_trippers.go:438] POST https://10.96.0.1:443/apis/authorization.k8s.io/v1/selfsubjectaccessreviews 201 Created in 7 milliseconds"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.115704       1 round_trippers.go:438] GET https://10.96.0.1:443/apis/linkerd.io/v1alpha2?timeout=32s 200 OK in 0 milliseconds"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.119568       1 round_trippers.go:438] POST https://10.96.0.1:443/apis/authorization.k8s.io/v1/selfsubjectaccessreviews 201 Created in 3 milliseconds"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.120051       1 merged_client_builder.go:122] Using in-cluster configuration"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.120639       1 merged_client_builder.go:122] Using in-cluster configuration"
time="2019-08-23T01:19:39Z" level=info msg="waiting for caches to sync"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.121941       1 reflector.go:131] Starting reflector *v1.Pod (10m0s) from pkg/mod/k8s.io/client-go@v0.0.0-20181213151034-8d9ed539ba31/tools/cache/reflector.go:95"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.122504       1 reflector.go:169] Listing and watching *v1.Pod from pkg/mod/k8s.io/client-go@v0.0.0-20181213151034-8d9ed539ba31/tools/cache/reflector.go:95"
time="2019-08-23T01:19:39Z" level=info msg="I0823 01:19:39.122762       1 reflector.go:131] Starting reflector *v1beta2.ReplicaSet (10m0s) from pkg/mod/k8s.io/client-go@v0.0.0-20181213151034-8d9ed539ba31/tools/cache/reflector.go:95"
```

alpeb:
@olix0r I confirmed authorization tokens are not leaked into the log. As @siggy pointed out above, klog's log level is hard-coded to be either 0 or 6 (when `--controller-log-level` is `debug`). The behavior is still as described on #2386: only at levels 7 or up are those tokens visible.

pkg/flags/flags.go
alpeb: So klog's level is 0 or 6 depending on whether `--controller-log-level` is `debug` or not. You think it's worth not hard-coding it to 6 if `-v` is explicitly set?

olix0r:
@alpeb thanks for confirming!

pkg/flags/flags.go
siggy: Sorry, yeah leave the hard-coded override as-is. When I said `(which worked!)`, I meant the public-api successfully booted rather than error out with `flag provided but not defined: -foo`.

PR #3316, opened 2019-08-23
Have the proxy-injector emit events upon injection/skipping injection

### LINKED ISSUE #3253: Better logging for why injection is skipped

alpeb:
Fixes #3253

Have the proxy-injector emit an event whenever a injection happens, or
when injection is skipped for some reason (also added that reason into
the proxy-injector logs). The level is associated to the parent workload
(it can't be associated to the pod because at this point the pod hasn't
been persisted).

The event recorder was setup at the `webhook/server.go` level and passed
to the proxy-injector's `Inject` function. The sp-validator thus also
has access to the event recorder, but for now it's not using it.

Related changes:

- Refactored `api.GetOwnerKindAndName()` to have it return a more
generic object.
- Refactored `report.Injectable()` to also have it return the reason why
a workload is not injectable.

### DIFF
--- a/controller/cmd/proxy-injector/main.go
+++ b/controller/cmd/proxy-injector/main.go
@@ -8,8 +8,9 @@ import (
 
 func main() {
 	webhook.Launch(
-		[]k8s.APIResource{k8s.NS, k8s.RS},
+		[]k8s.APIResource{k8s.NS, k8s.Deploy, k8s.RC, k8s.RS, k8s.Job, k8s.DS, k8s.SS, k8s.Pod},
 		9995,
 		injector.Inject,
+		"linkerd-proxy-injector",
 	)
 }
--- a/controller/proxy-injector/webhook.go
+++ b/controller/proxy-injector/webhook.go
@@ -12,12 +12,21 @@ import (
 	log "github.com/sirupsen/logrus"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+)
+
+const (
+	eventTypeSkipped  = "InjectionSkipped"
+	eventTypeInjected = "Injected"
 )
 
 // Inject returns an AdmissionResponse containing the patch, if any, to apply
 // to the pod (proxy sidecar and eventually the init container to set it up)
 func Inject(api *k8s.API,
 	request *admissionv1beta1.AdmissionRequest,
+	recorder record.EventRecorder,
 ) (*admissionv1beta1.AdmissionResponse, error) {
 	log.Debugf("request object bytes: %s", request.Object.Raw)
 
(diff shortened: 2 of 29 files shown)


controller/proxy-injector/webhook.go
```diff
@@ -53,8 +62,23 @@ func Inject(api *k8s.API,
...
-	if !report.Injectable() {
-		log.Infof("skipped %s", report.ResName())
+	var parent *runtime.Object
+	if ownerRef := resourceConfig.GetOwnerRef(); ownerRef != nil {
+		objs, err := api.GetObjects(request.Namespace, ownerRef.Kind, ownerRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		if len(objs) == 0 {
+			return nil, fmt.Errorf("couldn't retrieve parent object %s-%s-%s", request.Namespace, ownerRef.Kind, ownerRef.Name)
```
siggy: do we want to just log this error and continue with injection, rather than returning? the `if parent != nil` checks below make me think we can gracefully handle this?

controller/proxy-injector/webhook.go
```diff
@@ -53,8 +62,23 @@ func Inject(api *k8s.API,
...
+		}
+		parent = &objs[0]
+	}
+
+	if injectable, reason := report.Injectable(); !injectable {
+		if parent != nil {
+			recorder.Eventf(*parent, v1.EventTypeNormal, eventTypeSkipped, "Linkerd sidecar proxy injection skipped: %s", reason)
```
siggy: if there's no parent, should we record the event with the object itself?

pkg/inject/inject.go
```diff
@@ -342,8 +348,9 @@ func (conf *ResourceConfig) parse(bytes []byte) error {
 		conf.pod.meta = &v.ObjectMeta
 
 		if conf.ownerRetriever != nil {
-			kind, name := conf.ownerRetriever(v)
-			switch kind {
+			conf.workload.ownerRef = conf.ownerRetriever(v)
+			name := conf.workload.ownerRef.Name
+			switch strings.ToLower(conf.workload.ownerRef.Kind) {
```
siggy: non-actionable feedback (unrelated to this PR) to think about going forward: all of the `strings.ToLower` to match `Kind` to our `k8s.*` constants isn't ideal.

controller/cmd/proxy-injector/main.go
```diff
@@ -8,8 +8,9 @@ import (
 
 func main() {
 	webhook.Launch(
-		[]k8s.APIResource{k8s.NS, k8s.RS},
+		[]k8s.APIResource{k8s.NS, k8s.Deploy, k8s.RC, k8s.RS, k8s.Job, k8s.DS, k8s.SS, k8s.Pod},
```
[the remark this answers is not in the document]
alpeb: This is the `Eventf` signature:
```
Eventf(object runtime.Object, eventtype, reason, messageFmt string, args ...interface{})
```
The first arg is an object reference to the pod's parent workload (Deployment, Job, Statefulset, etc) that I need to retrieve from the k8s' api. This also explains why I needed to augment the rbac permissions for the proxy-injector.

controller/proxy-injector/webhook.go
alpeb: Agreed :+1:

controller/proxy-injector/webhook.go
alpeb: At this point in the persistence workflow the pod hasn't been persisted. If one tries to associate the event with the pod, the api will error out saying it can't find the object.

pkg/inject/inject.go
alpeb: Yeah, I'll see if I can see how they handle that in the k8s source code.

PR #3318, opened 2019-08-23
Release notes for edge-19.8.6

cpretzer:

Includes release notes and changes to edge version in Chart.yml and values.yml files.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,15 +1,16 @@
 ## edge-19.8.6
 
 * CLI
-  * Fixed missing `clusterDomain` in render RenderTapOutputProfile
-  * Added `LINKERD2_PROXY_DESTINATION_GET_SUFFIXES` env variable (thanks @arminbuerkle!)
-  * Merged the CLI 'installValues' type with Helm 'Values' type 
-* Controller 
-  * Added requirement for Go `1.12.9` for controller builds to include security fixes
+  * Fixed custom cluster domain support for tap profiles 
+    (thanks @arminbuerkle!) 
 * Web UI
   * Added a `Linkerd Namespace` Grafana dashboard, allowing users to view
-    historical data for a given namespace, similar to CLI output for `linkerd stat deploy -n myNs` (thanks @bourquep!)
-
+    historical data for a given namespace, similar to CLI output for 
+    `linkerd stat deploy -n myNs` (thanks @bourquep!)
+* Internal
+  * Added requirement for Go `1.12.9` for controller builds to include 
+    security fixes
+    
 ## stable-2.5.0
 
 This release adds [Helm support](https://linkerd.io/2/tasks/install-helm/),
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,32 @@
+## edge-19.8.6
+
+* CLI
+  * Fixed missing `clusterDomain` in render RenderTapOutputProfile
+  * Added LINKERD2_PROXY_DESTINATION_GET_SUFFIXES env variable (thanks @arminbuerkle!)
+  * Merged the CLI 'installValues' type with Helm 'Values' type 
+    * Renamed template-values.go
+    * Defined new constructor of charts.Values type
+    * Moved all Helm values related code to the pkg/charts package
+    * Bumped dependency
+    * Used '/' in filepath to remain compatible with VFS requirement
+    * Added unit test to verify Helm YAML output
+    * Added unit test for Helm YAML validation (HA)
+* Controller 
+  * Added requirement for go 1.12.9 for controller builds
+  * Reorganized authority-parsing code so that the is-this-a-kubernete-service-name decision is decoupled from parsing
+    of authorities into their consituent parts
+* Web UI
+  * Added a "Linkerd Namespace" Grafana dashboard  
+* Internal
+  * Added a unit test for the `edges` API endpoint. To maintain a consistent order for testing, the returned rows in 
+    api/public/edges.go are now sorted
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+    * Added unit test to verify Helm YAML output
+    * Added unit test for Helm YAML validation (HA)
+* Controller 
+  * Added requirement for go 1.12.9 for controller builds
+  * Reorganized authority-parsing code so that the is-this-a-kubernete-service-name decision is decoupled from parsing
+    of authorities into their consituent parts
+* Web UI
+  * Added a "Linkerd Namespace" Grafana dashboard  
+* Internal
+  * Added a unit test for the `edges` API endpoint. To maintain a consistent order for testing, the returned rows in
```
scottcarol: You can omit the new unit test as it's not user-facing -- I actually think you could delete the entire `Internal` section for this edge release but I would defer to @siggy and others

admc: In our review of the 2.5 release notes we agreed that for edges we want user facing stuff, and anything internal/architectural that might affect an edge adopters experience. Adding unit tests doesn't seem important to call out to me.

CHANGES.md
```diff
@@ -1,3 +1,32 @@
+## edge-19.8.6
+
+* CLI
+  * Fixed missing `clusterDomain` in render RenderTapOutputProfile
+  * Added LINKERD2_PROXY_DESTINATION_GET_SUFFIXES env variable (thanks @arminbuerkle!)
+  * Merged the CLI 'installValues' type with Helm 'Values' type
```
scottcarol: All of the indented bullets below this line should be deleted as those are from individual commits within the PR mentioned on line 6. We generally want to summarize the entire PR and describe why it is useful, in a single bullet point. 

And a high-level note, I am not sure if the specific PR listed on line 6 is user-facing enough to include - I would defer to @alpeb

CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+    * Bumped dependency
+    * Used '/' in filepath to remain compatible with VFS requirement
+    * Added unit test to verify Helm YAML output
+    * Added unit test for Helm YAML validation (HA)
+* Controller 
+  * Added requirement for go 1.12.9 for controller builds
+  * Reorganized authority-parsing code so that the is-this-a-kubernete-service-name decision is decoupled from parsing
+    of authorities into their consituent parts
+* Web UI
+  * Added a "Linkerd Namespace" Grafana dashboard
```
scottcarol: This feature was thanks to `@bourquep` 😄 

I would add a few words about why this was useful, maybe:

Added a “Linkerd Namespace” Grafana dashboard, allowing users to view historical data for a given namespace, similar to CLI output for `linkerd stat deploy -n myNs`

CHANGES.md
```diff
@@ -1,3 +1,32 @@
+## edge-19.8.6
+
+* CLI
+  * Fixed missing `clusterDomain` in render RenderTapOutputProfile
+  * Added LINKERD2_PROXY_DESTINATION_GET_SUFFIXES env variable (thanks @arminbuerkle!)
```
siggy: similar to @scottcarol's comment around line 6, combine lines 4 and 5 as they're from the same PR, and summarize into a more user-facing comment, something like:
```
Fixed custom cluster domain support for tap profiles
```

CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+  * Merged the CLI 'installValues' type with Helm 'Values' type 
+    * Renamed template-values.go
+    * Defined new constructor of charts.Values type
+    * Moved all Helm values related code to the pkg/charts package
+    * Bumped dependency
+    * Used '/' in filepath to remain compatible with VFS requirement
+    * Added unit test to verify Helm YAML output
+    * Added unit test for Helm YAML validation (HA)
+* Controller 
+  * Added requirement for go 1.12.9 for controller builds
```
scottcarol: I would capitalize "Go" and add "to include security fixes" to this bullet point to show users why this is important.

CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+  * Reorganized authority-parsing code so that the is-this-a-kubernete-service-name decision is decoupled from parsing
```
scottcarol: Think you can omit this.

CHANGES.md
siggy: i'd move this to internal, and note it applies to the whole codebase, beyond just the controller, something like
```
Require go 1.12.9 to address HTTP/2 vulnerabilities
```

CHANGES.md
```diff
@@ -1,3 +1,32 @@
...
+    of authorities into their consituent parts
```
siggy: you can remove this

CHANGES.md
siggy: yeah agree you can remove all these items under internal.

cpretzer: Thanks for the comments, I've removed the `Internal` section

CHANGES.md
cpretzer: Great additional context 👍

CHANGES.md
cpretzer: done

CHANGES.md
cpretzer: After chatting with @siggy I've added the entry for Go 1.12.9 under the Internal section because it deals with security, which users should be aware of

CHANGES.md
alpeb: That's correct, that's a 100% implementation detail that doesn't change the UX, so it shouldn't go in the release notes.

PR #3333, opened 2019-08-28
Add trafficsplit to dashboard

### LINKED ISSUE #3261: TrafficSplit dashboard integration

With TrafficSplit visibility landing as part of #3106, it should be added to the dashboard as well. This should include both a detail and overview page and match the other existing resources, such as deployments.

## How should the problem be solved?

Here's some mockups.

### Sidebar

Screenshot 2019-08-21 13 33 21

### Namespace List View

Nest the table from the list view below.

### List View

Screenshot 2019-08-21 13 58 52

### Detail View

This view should only contain the octopus and table, there should not be any other information here.

Screenshot 2019-08-21 13 40 22

Screenshot 2019-08-21 13 49 26

### Out of Scope

- Adding any information to the overview screen.
- Adding links or tables to any other list/details views (excluding namespaces).

scottcarol:
Fixes #3261. 

This PR adds trafficsplit data to the dashboard.

To test, install a trafficsplit. You can `kubectl apply` [booksapp-trafficsplit.yml](https://github.com/BuoyantIO/booksapp) in the Booksapp repo.

Note: Our current Octopus graph always displays the selected resource in the center, whether or not neighbors are to the left or right. In the case of trafficsplit there are never upstream neighbors to display, so it looks off-center. This is consistent with all other Octopus graphs in which there are no upstream neighbors.

Also, I left the P50 and P95 columns in the metrics table but it's quite wide, so we can remove if desired.

Screen Shot 2019-08-27 at 7 02 43 PM


alpeb:
> Not sure if related to this specific branch, but I'm seeing incomplete octopus graphs on the deploy pages

@siggy Yeah that's a bug that got into master, that's being fixed by #3331

web/app/js/components/util/MetricUtils.jsx
```diff
@@ -171,6 +176,14 @@ export const processMultiResourceRollup = rawMetrics => {
       return;
     }
 
+    // the Stat API returns trafficsplit data if queried with resource_type=all.
+    // trafficsplit leaves cannot be "upstream" or "downstream" of other
+    // resources, so we do not process trafficsplit data if the requested
+    // resourceType is not "trafficsplit" or "all".
+    if (resourceType !== "all" && resourceType !== "trafficsplit" && table.podGroup.rows[0].tsStats) {
+      return;
```
siggy: are we returning early here because the presence of `table.podGroup.rows[0].tsStats` implies that this row only has trafficsplit data and nothing else (`BasicStats` or `TcpStats`)? if so, mind adding that to the comment, and consider moving the `table.podGroup.rows[0].tsStats` to the front of the `if`, to signify its significance? or, if i'm missing something else, please enlighten me. =)

web/app/js/components/util/MetricUtils.jsx
```diff
@@ -112,8 +112,12 @@ const processStatTable = table => {
   let rows = _compact(table.podGroup.rows.map(row => {
     let runningPodCount = parseInt(row.runningPodCount, 10);
     let meshedPodCount = parseInt(row.meshedPodCount, 10);
+    let rowKey = `${row.resource.namespace}-${row.resource.type}-${row.resource.name}`;
+    if (row.resource.type === "trafficsplit" && row.tsStats) {
```
siggy: if only one of these two things is true, is that expected? or a bug?

web/app/js/components/TrafficSplitDetail.jsx
```diff
@@ -0,0 +1,70 @@
...
+// the octopus graph.
+const formatLeaves = resourceRsp => {
+  let leaves = processSingleResourceRollup(resourceRsp, "trafficsplit");
+  _each(leaves, leaf => {
+    leaf.name = leaf.tsStats.leaf;
+  });
+  return leaves;
+};
+export default class TrafficSplitDetail extends React.Component {
+    static propTypes = {
```
siggy: nit: indent

web/app/js/components/ResourceDetail.jsx
```diff
@@ -298,6 +304,15 @@ export class ResourceDetailBase extends React.Component {
 
     let showNoTrafficMsg = resourceIsMeshed && (Date.now() - lastMetricReceivedTime > showNoTrafficMsgDelayMs);
 
+    if (isTrafficSplit) {
+      return (
+        <TrafficSplitDetail
+          resourceType={resourceType}
+          resourceName={resourceName}
+          resourceMetrics={resourceMetrics}
+          resourceRsp={resourceRsp} />
```
siggy: tioli: this `TrafficSplitDetail` block and the `<div>` below feel a bit incongruous, as the former leverages a new component while the latter is in-lined. also, a lot of the code immediately above `TrafficSplitDetail` is not relevant if `isTrafficSplit` is true. consider breaking out the `div` below into a new component. this would have the added bonus of moving a lot of the code that is only relevant to the `!isTrafficSplit` case into that new component.

web/app/js/components/util/MetricUtils.jsx
scottcarol: Great question. Because trafficsplit metrics are now being returned from the Stat API endpoint, the current `processMultiResourceRollup` can erroneously return trafficsplit leaf services as "upstream" or "downstream" of a selected resource in the Octopus graph.

This line checks if the statTable contains trafficsplit data and returns early if the selected resource is not "all" or "trafficsplit", since any other resource should not include trafficsplit metrics.

I updated the comment to reflect this. The better longterm solution would be to refactor `processMultiResourceRollup` so that `upstream` and `downstream` are handled separately. I'd like to do that in a separate PR.

web/app/js/components/util/MetricUtils.jsx
scottcarol: You're right, `row.resource.type === "trafficsplit"` and `row.tsStats` would only occur together so I can remove one of the checks.

web/app/js/components/ResourceDetail.jsx
scottcarol: I see your point. My goal was to change `ResourceDetail` as little as possible, so I built the `TrafficSplitDetail` component as an extension of `ResourceDetail`. But this does mean that there is a lot of code in `ResourceDetail` that is not relevant to trafficsplits. Let me experiment with reorganizing to see if it’s more clear.

scottcarol:
Thanks @grampelberg for the review. Can you confirm which commit you were on? I pushed `e110352` about 45 minutes ago and believed that this fixed it.

scottcarol:
@adleong Thanks for the reviews. My rationale for leaving the `Traffic Split` column in when you're on the specific trafficsplit view is to mirror the output of `linkerd stat ts` which lists the trafficsplit name for every leaf row. 

Since this is a slightly different kind of resource table (a row for each leaf, not a row for each traffic split) I wanted to err on the side of more info, not less. @grampelberg WDYT?

adleong:
My thinking is that the `stat ts` command may list multiple different traffic split resources in the same table, whereas the "Leaf Services" table on the traffic split detail page will only ever have one traffic split.

scottcarol:
@adleong That's a good point. The bonus of removing the traffic split name is removing a column in a very wide table. There will be a slight inconsistency with `linkerd stat ts/authors-split` which does print the split name but I think that's fine.

PR #3339, opened 2019-08-28
Introduce Github Actions CI workflow

siggy:
The existing Travis CI setup requires additional integrations and
permissions with Github, and also lacks some flexibility around job
dependency management.

Introduce a new CI workflow based on Github Actions. This initial
workflow performs the same CI work that Travis does, and will iniitially
run in parallel:
- Go unit tests
- JS unit tests
- Go lint
- Validate Go deps
- Integration tests (deep, upgrade, helm)

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -15,8 +15,6 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v1
-    - name: ENV
-      run: env | sort
     - name: Validate go deps
       run: |
         . bin/_tag.sh
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,223 @@
+name: CI
+
+on:
+  pull_request: {}
+  push:
+    branches:
+    - master
+    - refs/tags/edge-[0-9]?.[0-9]?.[0-9]?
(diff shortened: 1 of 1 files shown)


siggy:
@alpeb Re: Dependencies between jobs, I'd like to avoid introducing dependencies unless a job actually depends on a previous job. The goal is to minimize the total time for all jobs to complete, and also make clear what jobs depends on other jobs. While `validate_go_deps` takes only ~8s, `go_lint` takes ~4m, and `go_unit_tests` takes ~5m.

Fortunately the Github PR and Actions UIs provide feedback when individual jobs fail, prior to the full workflow completing, so the user gets feedback asap.

For reference, right now the critical path is:
[docker_build | kind_setup] -> kind_integration -> kind_cleanup

siggy:
@ihcsim Agree good to highlight the `same workspace filesystem` feature between steps in a job. We are taking advantage of this feature in the `Docker SSH setup`, but the more, the better.

siggy:
Filed a related issue to Github Community forums:
https://github.community/t5/GitHub-API-Development-and/Github-Actions-Inconsistent-repo-checkouts-across-jobs/td-p/30258

Sort of relates to this issue in actions/checkout:
https://github.com/actions/checkout/issues/15

PR #3341, opened 2019-08-29
Fix up grammar mistake

tanjunchen:


grampelberg:
@tanjunchen it looks like you've pulled in a lot of changes outside of your branch. You might want to rebase and cherry pick to get rid of the other unassociated changes.

There are also some things like `thatproxy` which seem like a bad search and replace?

PR #3350, opened 2019-08-29
update helm chart for 19.8.7

hawkw:
Hadn't realised this was also part of the release process now. My bad!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,6 @@
 ## edge-19.8.7
 
-* Web UI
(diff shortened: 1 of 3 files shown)


hawkw:
My assumption is that this test failure (https://travis-ci.org/linkerd/linkerd2/jobs/578547856#L316) is transient --- is that correct?

ihcsim:
That's correct. It looks like the ssh tunnelling failed, and it's unrelated to the change in this PR.

PR #3368, opened 2019-09-03
Fixing row keys for edges

scottcarol:
While testing trafficsplit data I found two issues in `EdgesUtils` which were
leading to React console warnings for duplicate row keys. This PR
ensures that there are no duplicate keys for `edges` table rows.

### DIFF
--- a/web/app/js/components/util/EdgesUtils.jsx
+++ b/web/app/js/components/util/EdgesUtils.jsx
@@ -1,7 +1,6 @@
 import PropTypes from 'prop-types';
 import _each from 'lodash/each';
 import _isEmpty from 'lodash/isEmpty';
-import _startsWith from 'lodash/startsWith';
 
 export const processEdges = (rawEdges, resourceName) => {
   let edges = [];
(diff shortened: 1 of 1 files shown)


web/app/js/components/util/EdgesUtils.jsx
```diff
@@ -13,15 +12,15 @@ export const processEdges = (rawEdges, resourceName) => {
       return;
     }
     // check if any of the returned edges match the current resourceName
-    if (_startsWith(edge.src.name, resourceName)) {
+    if (edge.src.name === resourceName) {
```
siggy: any idea why we were using `_startsWith` in the first place?

scottcarol: Good question, thanks for asking! I believe that as I began writing `EdgesUtils` for #2965, I originally assigned each row in `rawEdges` a `key` and was using `_startsWith` to check if that `key` started with the given `resourceName`. As I simplified the logic, I no longer filtered by `key`, but I didn’t catch the error. This happened before I put up the PR so the original record is lost to history.

PR #3372, opened 2019-09-04
GitHub Actions, kind, integration test logs fixes

siggy:
PR #3339 introduced a GitHub Actions CI workflow. Booting 6 clusters
simultaneously (3x Github Actions + 3x Travis) exhibits some transient
failures.

Implement fixes in GitHub Actions and integration tests to address kind
cluster creation and testing:
- Retry kind cluster creation once.
- Retry log reading from integration k8s clusters once.
- Add kind cluster creation debug logging.
- Add a GitHub Actions status badge to the top of `README.md`.

Screen Shot 2019-09-03 at 6 31 27 PM

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -131,7 +131,9 @@ jobs:
       run: |
         TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
         export KIND_CLUSTER=github-$TAG-${{ matrix.integration_test }}
-        bin/kind create cluster --name=$KIND_CLUSTER --wait=1m
+        # retry cluster creation once in case of port conflict or kubeadm failure
+        bin/kind create cluster --name=$KIND_CLUSTER --wait=2m --loglevel debug ||
+          bin/kind create cluster --name=$KIND_CLUSTER --wait=2m --loglevel debug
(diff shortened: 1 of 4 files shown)


.github/workflows/workflow.yml
```diff
@@ -131,7 +131,9 @@ jobs:
       run: |
         TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
         export KIND_CLUSTER=github-$TAG-${{ matrix.integration_test }}
-        bin/kind create cluster --name=$KIND_CLUSTER --wait=1m
+        # retry cluster creation once in case of port conflict or kubeadm failure
+        bin/kind create cluster --name=$KIND_CLUSTER --wait=2m --loglevel debug ||
+          bin/kind create cluster --name=$KIND_CLUSTER --wait=2m --loglevel debug
```
ihcsim: If the first cluster creation fails, will the cluster remain on the host? Or does `kind` perform auto cleanup? If it doesn't clean up after itself, the second attempt to create the cluster will fail, right? In that case, I wonder if we can just do a `docker restart $KIND_CLUSTER-control-plane`?

siggy: Yeah, `kind` cleans up after itself if cluster creation fails.

PR #3387, opened 2019-09-05
Update to client-go v12.0.0, forked stern

### LINKED ISSUE #3382: Upgrade client-go
Our version of client-go is about to be EOLed and it's time to upgrade. Unfortunately, this is not  straightforward, since our dependency on `stern` can't easily be upgraded... (@siggy has more information on that)

siggy:
The repo depended on an old version of client-go. It also depended on
stern, which itself depended on an old version of client-go, making
client-go upgrade non-trivial.

Update the repo to client-go v12.0.0, and also replace stern with a
fork.

This fork of stern includes the following changes:
- updated to use Go Modules
- updated to use client-go v12.0.0
- fixed log line interleaving:
  - https://github.com/wercker/stern/issues/96
  - based on:
    - https://github.com/oandrew/stern/commit/8723308e46b408e239ce369ced12706d01479532

Fixes #3382


pkg/prometheus/prometheus.go
```diff
@@ -7,11 +7,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/grpc"
+	"k8s.io/client-go/transport"
```
dadjeibaah: It looks like this package is coming from k8s' client-go library but we are using this in the prometheus package. What is the reasoning behind switching this from the custom type `WrapTransport` to the k8s provided transport? I would think that we'd still want to keep using `WrapTransport` so we have control over the type.

go.mod
```diff
@@ -68,24 +64,26 @@ require (
...
-	k8s.io/apimachinery v0.0.0-20181127105237-2b1284ed4c93
-	k8s.io/client-go v0.0.0-20181213151034-8d9ed539ba31
+	k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
+	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 	k8s.io/helm v2.12.2+incompatible
 	k8s.io/klog v0.3.2
 	sigs.k8s.io/yaml v1.1.0
 )
+
+replace github.com/wercker/stern => github.com/siggy/stern v0.0.0-20190905052711-472ce9d16dab
```
alpeb: @siggy why not have the fork live under the linkerd org?

cli/cmd/logs.go
```diff
@@ -97,10 +99,36 @@ func (o *logsOptions) toSternConfig(controlPlaneComponents, availableContainers
 	if err != nil {
 		return nil, err
 	}
+
+	// Based on stern/cmd/cli.go
```
dadjeibaah: It's unfortunate that this fix has not been merged to stern yet :(. The more we patch this the more I feel like we should just implement this feature ourselves.

adleong: should this comment link to a github issue in the upstream stern repo?  I don't know what this is fixing.

cli/cmd/logs.go
```diff
@@ -97,10 +99,36 @@ func (o *logsOptions) toSternConfig(controlPlaneComponents, availableContainers
...
+	t := "{{color .PodColor .PodName}} {{color .ContainerColor .ContainerName}} {{.Message}}"
+	if o.noColor {
+		t = "{{.PodName}} {{.ContainerName}} {{.Message}}"
+	}
+	funs := map[string]interface{}{
```
adleong: is this a typo for `funcs`?

pkg/prometheus/prometheus.go
adleong: I agree that having this prometheus package depend on client-go is a bit weird.

go.mod
siggy: i opted to put it under my personal account due to concerns around making the fork look like a project that linkerd is officially supporting (for other developers to depend on).

if we as an org are comfortable with those possible implications, i can totally move it, but i was concerned it would send the wrong message.

pkg/prometheus/prometheus.go
siggy: curiously, this change was motivated because our `prometheus.WrapTransport` type was no longer compatible with client-go's `config.WrapTransport` type:
```
pkg/k8s/api.go:48:23: cannot use prometheus.ClientWithTelemetry("k8s", wt) (type prometheus.WrapTransport) as type "k8s.io/client-go/transport".WrapperFunc in assignment
pkg/k8s/api.go:48:56: cannot use wt (type "k8s.io/client-go/transport".WrapperFunc) as type prometheus.WrapTransport in argument to prometheus.ClientWithTelemetry
```

config.go in client-go changed from `v10.0.0` to `v12.0.0`:
```diff
-WrapTransport func(rt http.RoundTripper) http.RoundTripper
+WrapTransport transport.WrapperFunc
```

...with this in `v12.0.0`:
```go
type WrapperFunc func(rt http.RoundTripper) http.RoundTripper
```

i agree the client-go dependency in our prometheus package is not great, i'll find a way to remove that.

cli/cmd/logs.go
siggy: this particular comment is not a fix, simply borrowing code from stern, and i wanted to call it out.

this does bring up a good point around the fixes that were necessary to workaround the log line interleaving issue. i've updated the PR with a comment around that part of the code.

cli/cmd/logs.go
siggy: it is intended to mean `functions`, it comes from https://github.com/oandrew/stern/commit/8723308e46b408e239ce369ced12706d01479532

go.mod
olix0r: if we're going to have the project depend on forks, i'm slightly more comfortable being owned by the linkerd org. i'm not concerned about branding implications, since there are good tools to highlight the repos we care about

siggy: sounds good, i'll move it over to the linkerd org.

PR #3393, opened 2019-09-06
Add prefixes to "partials" and "patch" Helm charts (#3392)

### LINKED ISSUE #3392: Please rename "partials" and "path" Helm charts

1. If one needs to publish Linkerd helm charts to a local registry or cache them in a local git repo, these names are not clear:
- partials
- patch

2. Also I found that it is very hard to find search entries for "partials" and especially for "patch" in the project code without the prefix.

Please add prefixes same as for the main chart:

- linkerd2-partials
- linkerd2-patch

Nothing clever comes to my mind.

For users everything will be the same. Taking into account that [the documentation](https://linkerd.io/2/tasks/install-helm/) is pretty simple, users like me can not understand why the `patch` chart exists and how to use it. :(

KIVagant:
Fixes #3392

Changes chart names so that they can be pushed in any Helm chart repository without name collisions.

Signed-off-by: Eugene Glotov <kivagant@gmail.com>


kivagant-ba:
Please help me to test this.

KIVagant:
Please help me to test this.

PR #3401, opened 2019-09-07
Update Prometheus config to keep only needed cadvisor metrics

ihcsim:
This PR updates the Prometheus config to keep only metrics relevant to containers resource usage, as needed by the Grafana dashboard and the `heartbeat` service.

With this change, we see a significant drop in the number of time series data ingested by the `kubernetes-nodes-cadvisor` job:

stable-2.5:
before-changes

git-5446b6a2:
after-changes

There is also a drop in the amout of memory required by Prometheus on start up:

```
✗ linkerd version                                                                                                      
Client version: stable-2.5.0
Server version: stable-2.5.0

✗ k -n linkerd top po
NAME                                      CPU(cores)   MEMORY(bytes)
...
linkerd-prometheus-58d4c4b7b4-t4277      362m         903Mi
...

✗ bin/linkerd version                                                                                                  (isim-dev-prom2/default)
Client version: git-5446b6a2
Server version: git-5446b6a2

✗ k -n linkerd top po
NAME                                     CPU(cores)   MEMORY(bytes)
...
linkerd-prometheus-6b9868dc49-wt4hs      27m          147Mi
...
```
Here's a sample comparison of the Prometheus' working set memory between `stable-2.5` and this feature branch, after running 500 emojivoto pods, the Linkerd dashboard and a few queries from the Prometheus console:

stable-2.5:
before-changes-memory

git-5446b6a2:
after-changes-memory

I notice that the difference in the memory usage is dependent on the promql queries invoked, where the more expensive the queries are, the more noticeable the difference.

### DIFF
--- a/charts/linkerd2/templates/prometheus.yaml
+++ b/charts/linkerd2/templates/prometheus.yaml
@@ -54,6 +54,9 @@ data:
       - source_labels: [__name__]
         regex: '(container|machine)_(cpu|memory|network|fs)_(.+)'
         action: keep
+      - source_labels: [__name__]
+        regex: 'container_memory_failures_total' # unneeded large metric
+        action: drop
       relabel_configs:
       - action: labelmap
         regex: __meta_kubernetes_node_label_(.+)
--- a/charts/linkerd2/templates/prometheus.yaml
+++ b/charts/linkerd2/templates/prometheus.yaml
@@ -50,13 +50,6 @@ data:
 
       kubernetes_sd_configs:
       - role: node
-      metric_relabel_configs:
-      - source_labels: [__name__]
-        regex: '(container|machine)_(cpu|memory|network|fs)_(.+)'
-        action: keep
-      - source_labels: [__name__]
-        regex: 'container_memory_failures_total' # unneeded large metric
-        action: drop
       relabel_configs:
       - action: labelmap
         regex: __meta_kubernetes_node_label_(.+)
@@ -66,6 +59,13 @@ data:
         regex: (.+)
         target_label: __metrics_path__
         replacement: /api/v1/nodes/$1/proxy/metrics/cadvisor
+      metric_relabel_configs:
+      - source_labels: [__name__]
+        regex: '(container|machine)_(cpu|memory|network|fs)_(.+)'
+        action: keep
+      - source_labels: [__name__]
+        regex: 'container_memory_failures_total' # unneeded large metric
+        action: drop
 
     - job_name: 'linkerd-controller'
       kubernetes_sd_configs:
--- a/charts/linkerd2/templates/prometheus.yaml
+++ b/charts/linkerd2/templates/prometheus.yaml
@@ -50,6 +50,10 @@ data:
 
       kubernetes_sd_configs:
       - role: node
+      metric_relabel_configs:
+      - source_labels: [__name__]
+        regex: '(container|machine)_(cpu|memory|network|fs)_(.+)'
+        action: keep
       relabel_configs:
       - action: labelmap
         regex: __meta_kubernetes_node_label_(.+)
--- a/cli/cmd/testdata/install_control-plane.golden
+++ b/cli/cmd/testdata/install_control-plane.golden
@@ -806,6 +806,13 @@ data:
 
       kubernetes_sd_configs:
       - role: node
+      metric_relabel_configs:
+      - source_labels: [__name__]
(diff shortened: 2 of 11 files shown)


ihcsim:
After running my clusters for 3 days with 500 emojivoto bots , the cluster with Prometheus was deployed w/o this change ran OOM, while the cluster with this change is still alive.

stable-2.5:
before-changes

git-5446b6a2:
after-changes

charts/linkerd2/templates/prometheus.yaml
```diff
@@ -50,6 +50,13 @@ data:
 
       kubernetes_sd_configs:
       - role: node
+      metric_relabel_configs:
```
siggy: since `metric_relabel_configs` happen after `relabel_configs` are applied, mind moving this below that section?

charts/linkerd2/templates/prometheus.yaml
```diff
@@ -50,6 +50,13 @@ data:
...
+      - source_labels: [__name__]
+        regex: '(container|machine)_(cpu|memory|network|fs)_(.+)'
+        action: keep
+      - source_labels: [__name__]
+        regex: 'container_memory_failures_total' # unneeded large metric
+        action: drop
```
siggy: what do you think about limiting the metrics strictly to what we show in the dashboard? /cc @grampelberg 

```bash
$ curl -s "http://localhost:9090/api/v1/query?query=count(%7Bjob%3D%22kubernetes-nodes-cadvisor%22%7D)%20by(__name__)" |
  jq -r '.data.result[] | "\(.value[1]):\(.metric.__name__)" ' |
  sort -r -n |
  xargs -I {} sh -c '
    echo {} |
    cut -d ":" -f2 |
    grep -q -f /dev/stdin grafana/dashboards/kubernetes.json; echo $?: {}
  '
0: 62:container_memory_working_set_bytes
1: 62:container_memory_usage_bytes
1: 62:container_memory_swap
1: 62:container_memory_rss
1: 62:container_memory_max_usage_bytes
1: 62:container_memory_mapped_file
1: 62:container_memory_failcnt
1: 62:container_memory_cache
1: 62:container_cpu_user_seconds_total
0: 62:container_cpu_usage_seconds_total
1: 62:container_cpu_system_seconds_total
1: 62:container_cpu_load_average_10s
1: 30:container_fs_writes_total
1: 30:container_fs_reads_total
1: 27:container_fs_writes_bytes_total
1: 27:container_fs_reads_bytes_total
1: 17:container_network_transmit_packets_total
1: 17:container_network_transmit_packets_dropped_total
1: 17:container_network_transmit_errors_total
0: 17:container_network_transmit_bytes_total
1: 17:container_network_receive_packets_total
1: 17:container_network_receive_packets_dropped_total
1: 17:container_network_receive_errors_total
0: 17:container_network_receive_bytes_total
1: 4:container_fs_writes_merged_total
1: 4:container_fs_write_seconds_total
0: 4:container_fs_usage_bytes
1: 4:container_fs_sector_writes_total
1: 4:container_fs_sector_reads_total
1: 4:container_fs_reads_merged_total
1: 4:container_fs_read_seconds_total
0: 4:container_fs_limit_bytes
1: 4:container_fs_io_time_weighted_seconds_total
1: 4:container_fs_io_time_seconds_total
1: 4:container_fs_io_current
1: 4:container_fs_inodes_total
1: 4:container_fs_inodes_free
1: 2:container_cpu_cfs_throttled_seconds_total
1: 2:container_cpu_cfs_throttled_periods_total
1: 2:container_cpu_cfs_periods_total
1: 1:up
1: 1:scrape_series_added
1: 1:scrape_samples_scraped
1: 1:scrape_samples_post_metric_relabeling
1: 1:scrape_duration_seconds
0: 1:machine_memory_bytes
0: 1:machine_cpu_cores
```

(un)related: i just noticed that we're collecting `container_memory_usage_bytes` but grafana references `container_memory_usage`, so that dashboard may need an update (https://grafana.com/grafana/dashboards/315 is 2 years old). this is probably for another PR, but noting it here in case we do remove every metric above that is not referenced in the dashboard.

ihcsim: Yup, I like the idea of keeping only those metrics that the dashboard needs; just wasn't sure if someone will ever want to query for those other cadvisor-related metrics, say in a federated environment, for alerts etc.

PR #3407, opened 2019-09-09
Put the destination controller into its own deployment

### LINKED ISSUE #3268: Split the destination controller into its own deployment
Since the destination controller is a fairly critical component, it should be moved into a distinct deployment (out of the linkerd-controller deployment) so that it is isolated from other controllers.

This was motivated by a situation where misbehaving clients overwhelmed the destination controller, putting it into crash loop backoff. This unfortunately impacts the availability of other components, like the stat command -- and being unable to use these components to diagnose control plane issues is _extra_ difficult.

alpeb:
Fixes #3268

Now that linkerd-controller pod ends up with just the public-api container, we might
wanna consider renaming it in a followup PR.


alpeb:
@ihcsim great! thanks for circling back.

PR #3408, opened 2019-09-09
Introduce healthcheck dependencies

alenkacz:

Submitted based on discussion with @siggy 

Currently HealthChecking code has few flaws that make it hard to use - one of them is that there is implicit dependencies between checks that might cause HC to be initialized in a way that a check would always fail e.g. because kubernetes client was not initialized by running a different check before.

This PR proposes a way how to make those dependencies explicit.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -240,9 +240,10 @@ type CheckResult struct {
 type checkObserver func(*CheckResult)
 
 type category struct {
-	id       CategoryID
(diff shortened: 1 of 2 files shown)


alenkacz:
hmm ok, I did not see that coming :D I should have ... I'll have to think about it more then :)

adleong:
@alenkacz are you still interested in working on this?

alenkacz:
Nah let's close it for now, I think I need to approach it differently. I can always open another one :)

Issue #3262, opened 2019-08-14
Expose arbitrary headers through tap


grampelberg:
### What problem are you trying to solve?

Currently, when running `tap`, users are able to see a few choice details. The selection is small and has been restricted for security concerns. With tap moving to an RBAC model, these concerns are less of an issue. It is extremely helpful to see arbitrary headers as part of the request/response flow.

Tap should be expanded to show all request, response and trailer headers.

### How should the problem be solved?

There are no user facing changes for this proposal. Instead, headers should be added to the tap protobuf. As #3390 recommends a 1:1 dump of the protobuf for JSON output, the headers will automatically start showing up. This should be an unstructured list of labels similar to `EndpointMeta`.

- Include all headers, for http1 and http2 including the pseudo headers.
- Normalize header names by lowercasing them.

### Out of scope

- Proxy side filtering

grampelberg:
Related #3035.

hawkw:
> 
>   * Include all headers, for http1 and http2 including the pseudo headers.

What's the motivation for including HTTP/2 pseudo-headers in the headers part of the tap event?

The tap events [already include pseudo-header fields](https://github.com/linkerd/linkerd2-proxy-api/blob/ddbc3a4f7f8b0058801f896d27974d19ee98094c/proto/tap.proto#L142-L145) in the structured data. Furthermore, this means that pseudo-headers will show up in the headers part of the tap event for HTTP/2 requests, while the equivalent values for HTTP/1 will show up in the tap event's structured data, but _not_ in the headers, which seems inconsistent?

>   * Normalize header names by lowercasing them.
>  
FWIW, the proxy already does this :)

adleong:
> request, response and trailer headers.

To be clear, this means request headers, request trailers, response headers, and response trailers?

adleong:
> Include all headers, for http1 and http2 including the pseudo headers.

Binary valued headers too?

hawkw:
> > request, response and trailer headers.
> 
> To be clear, this means request headers, request trailers, response headers, and response trailers?

Request stream ends don't currently emit tap events (as I pointed out here https://github.com/linkerd/linkerd2-proxy-api/pull/33#issuecomment-529696679). In order to display request trailers, we would also need to add a new tap event for request body ends.

> > Include all headers, for http1 and http2 including the pseudo headers.
> 
> Binary valued headers too?

[...]

PR #3410, opened 2019-09-09
Add TapEvent headers and trailers to the tap protobuf

kleimkuhler:
### Motivation

In order to expose arbitrary headers through tap, headers and trailers should be
read from the linkerd2-proxy-api `TapEvent`s and set in the public `TapEvent`s.
This change should have no user facing changes as it just prepares the events
for JSON output in linkerd/linkerd2#3390

### Solution

The public API has been updated with a headers field for
`TapEvent_Http_RequestInit_` and `TapEvent_Http_ResponseInit_`, and trailers
field for `TapEvent_Http_ResponseEnd_`.

These values are set by reading the corresponding fields off of the proxy's tap
events.

The proto changes are equivalent to the proto changes proposed in
linkerd/linkerd2-proxy-api#33

Closes #3262

Signed-off-by: Kevin Leimkuhler <kleimkuhler@icloud.com>

PR #3411, opened 2019-09-10
Add JSON output to tap command

### LINKED ISSUE #3390: Add json output to tap

It is a little tough to filter/read the current tap output. As headers are being added to tap, the output is starting to get difficult to consume. Take a peek at #3262 for an example. It would be nice to have some more machine readable output that can be sliced and diced with tools such as `jq`.

```bash
$ linkerd -n linkerd tap deploy/linkerd-web -o json | jq .
{
    "source": {
        "ip": "10.4.1.42",
        "port": 55370
    },
    "source_meta": {
        "kind": "deployment",
        "namespace": "linkerd",
        "name": "linkerd-prometheus"
    },
    "destination": {
        "ip": "10.4.1.41",
        "port": 9994
    },
    "destination_meta": {
        "kind": "deployment",
        "namespace": "linkerd",
        "name": "linkerd-web"
    },
    "route_meta": {
        "route": "/api/"
    },
    "proxy_direction": "inbound",
    "http": {
        "request_init": {
            "id": {
                "base": 1234,
                "stream": 5678
            },
            "method": {
                "registered": "GET"
            },
            "scheme": {
                "registered": "HTTP"
            },
            "authority": "linkerd-web.linkerd.svc.cluster.local:9994",
            "path": "/api/foobar"
        }
    }
}
{
    "source": {
        "ip": "10.4.1.42",
        "port": 55370
    },
    "source_meta": {
        "kind": "deployment",
        "namespace": "linkerd",
        "name": "linkerd-prometheus"
    },
    "destination": {
        "ip": "10.4.1.41",
        "port": 9994
    },
    "destination_meta": {
        "kind": "deployment",
        "namespace": "linkerd",
        "name": "linkerd-web"
    },
    "route_meta": {
        "route": "/api/"
    },
    "proxy_direction": "inbound",
    "http": {
        "response_init": {
            "id": {
                "base": 1234,
                "stream": 5678
            },
            "since_request_init": {
                "seconds": 0,
                "nanos": 1000
            },
            "http_status": 200
        }
    }
}
{
    "source": {
        "ip": "10.4.1.42",
        "port": 55370
    },
    "source_meta": {
        "kind": "deployment",
        "namespace": "linkerd",
        "name": "linkerd-prometheus"
    },
    "destination": {
        "ip": "10.4.1.41",
        "port": 9994
    },
    "destination_meta": {
        "kind": "deployment",
        "namespace": "linkerd",
        "name": "linkerd-web"
    },
    "route_meta": {
        "route": "/api/"
    },
    "proxy_direction": "inbound",
    "http": {
        "response_end": {
            "id": {
                "base": 1234,
                "stream": 5678
            },
            "since_request_init": {
                "seconds": 0,
                "nanos": 1000
            },
            "since_request_init": {
                "seconds": 0,
                "nanos": 2000
            },
            "response_bytes": 1024,
            "eos": {
                "reset_error_code": 0
            }
        }
    }
}
```

Note: this is a JSON dump of the existing tap protobuf.

kleimkuhler:
### Motivation

It is a little tough to filter/read the current tap output. As headers are being
added to tap, the output is starting to get difficult to consume. Take a peek at
#3262 for an example. It would be nice to have some more machine readable output
that can be sliced and diced with tools such as jq.

### Solution

A new output option has been added to the `linkerd tap` command that returns the
JSON encoding of tap events.

The default output is line oriented; `-o wide` appends the request's target
resource type to the tap line oriented tap events.

**Default**:
```
➜  linkerd2 git:(kleimkuhler/tap-json-output) linkerd -n linkerd tap deploy/linkerd-web
req id=5:0 proxy=in  src=10.1.6.146:36976 dst=10.1.6.148:9994 tls=not_provided_by_remote :method=GET :authority=10.1.6.148:9994 :path=/metrics
rsp id=5:0 proxy=in  src=10.1.6.146:36976 dst=10.1.6.148:9994 tls=not_provided_by_remote :status=200 latency=3366µs
end id=5:0 proxy=in  src=10.1.6.146:36976 dst=10.1.6.148:9994 tls=not_provided_by_remote duration=132µs response-length=1505B
```

**Wide**:
```
➜  linkerd2 git:(kleimkuhler/tap-json-output) linkerd -n linkerd tap deploy/linkerd-web -o wide
req id=6:0 proxy=in  src=10.1.0.1:35394 dst=10.1.6.148:9994 tls=not_provided_by_remote :method=GET :authority=10.1.6.148:9994 :path=/ping dst_res=deploy/linkerd-web dst_ns=linkerd
rsp id=6:0 proxy=in  src=10.1.0.1:35394 dst=10.1.6.148:9994 tls=not_provided_by_remote :status=200 latency=1442µs dst_res=deploy/linkerd-web dst_ns=linkerd
end id=6:0 proxy=in  src=10.1.0.1:35394 dst=10.1.6.148:9994 tls=not_provided_by_remote duration=88µs response-length=5B dst_res=deploy/linkerd-web dst_ns=linkerd
```

`-o json` performs a JSON encoding of each tap event using `"encoding/json"`.
This encodes the entire `TapEvent` protobuf as JSON and prints it to the
console.

**JSON**
*Edit 1: Updated header values to `string` when valid*
*Edit 2: Updated header values to be `oneof`*
*Edit 3: Changed tap JSON marshaler to `protobuf/jsonpb`*
```
{
  "source": {
    "ip": {
      "ipv4": 176820176
    },
    "port": 55294
  },
  "sourceMeta": {
    "labels": {
      "daemonset": "ip-masq-agent",
      "namespace": "kube-system",
      "pod": "ip-masq-agent-ltt4f",
      "serviceaccount": "ip-masq-agent",
      "tls": "not_provided_by_remote"
    }
  },
  "destination": {
    "ip": {
      "ipv4": 171704338
    },
    "port": 9994
  },
  "destinationMeta": {
    "labels": {
      "control_plane_ns": "linkerd",
      "deployment": "linkerd-web",
      "namespace": "linkerd",
      "pod": "linkerd-web-5967994c7d-dmlnw",
      "pod_template_hash": "5967994c7d",
      "serviceaccount": "linkerd-web"
    }
  },
  "routeMeta": {
    "labels": {
    }
  },
  "proxyDirection": "INBOUND",
  "http": {
    "requestInit": {
      "id": {
        "base": 0,
        "stream": "0"
      },
      "method": {
        "registered": "GET"
      },
      "scheme": null,
      "authority": "10.60.0.18:9994",
      "path": "/ping",
      "headers": {
        "headers": [
          {
            "name": "host",
            "valueStr": "10.60.0.18:9994"
          },
          {
            "name": "user-agent",
            "valueStr": "kube-probe/1.14+"
          },
          {
            "name": "accept-encoding",
            "valueStr": "gzip"
          }
        ]
      }
    }
  }
}
{
  "source": {
    "ip": {
      "ipv4": 176820176
    },
    "port": 55294
  },
  "sourceMeta": {
    "labels": {
      "daemonset": "calico-node",
      "namespace": "kube-system",
      "pod": "calico-node-vwgfx",
      "serviceaccount": "calico-sa",
      "tls": "not_provided_by_remote"
    }
  },
  "destination": {
    "ip": {
      "ipv4": 171704338
    },
    "port": 9994
  },
  "destinationMeta": {
    "labels": {
      "control_plane_ns": "linkerd",
      "deployment": "linkerd-web",
      "namespace": "linkerd",
      "pod": "linkerd-web-5967994c7d-dmlnw",
      "pod_template_hash": "5967994c7d",
      "serviceaccount": "linkerd-web"
    }
  },
  "routeMeta": {
    "labels": {
    }
  },
  "proxyDirection": "INBOUND",
  "http": {
    "responseInit": {
      "id": {
        "base": 0,
        "stream": "0"
      },
      "sinceRequestInit": "0.000595751s",
      "httpStatus": 200,
      "headers": {
        "headers": [
          {
            "name": "date",
            "valueStr": "Thu, 12 Sep 2019 18:13:32 GMT"
          },
          {
            "name": "content-length",
            "valueStr": "5"
          },
          {
            "name": "content-type",
            "valueStr": "text/plain; charset=utf-8"
          }
        ]
      }
    }
  }
}
{
  "source": {
    "ip": {
      "ipv4": 176820176
    },
    "port": 55294
  },
  "sourceMeta": {
    "labels": {
      "daemonset": "ip-masq-agent",
      "namespace": "kube-system",
      "pod": "ip-masq-agent-ltt4f",
      "serviceaccount": "ip-masq-agent",
      "tls": "not_provided_by_remote"
    }
  },
  "destination": {
    "ip": {
      "ipv4": 171704338
    },
    "port": 9994
  },
  "destinationMeta": {
    "labels": {
      "control_plane_ns": "linkerd",
      "deployment": "linkerd-web",
      "namespace": "linkerd",
      "pod": "linkerd-web-5967994c7d-dmlnw",
      "pod_template_hash": "5967994c7d",
      "serviceaccount": "linkerd-web"
    }
  },
  "routeMeta": {
    "labels": {
    }
  },
  "proxyDirection": "INBOUND",
  "http": {
    "responseEnd": {
      "id": {
        "base": 0,
        "stream": "0"
      },
      "sinceRequestInit": "0.000644509s",
      "sinceResponseInit": "0.000048758s",
      "responseBytes": "5",
      "eos": null,
      "trailers": {
        "headers": [
        ]
      }
    }
  }
}
```

Currently, `json` does not attempt to decode `[]byte` values, so the `Value`
fields remain base64 encoded. Decoding of these values into UTF8 could be
attempted, but they may not always be valid.

Closes #3390


kleimkuhler:
@grampelberg @olix0r This is opened as a draft because of the open question [here](https://github.com/linkerd/linkerd2-proxy-api/pull/33#discussion_r322547771).

If we want to decode base64 values, this could be done in a follow-up PR or added to this. If a value is valid UTF8, then it should probably be formatted that way. If a value is not valid UTF8, how should it be displayed?

grampelberg:
The two here that really pop out:

- It'd be nice to have real IP addresses.
- WDYT about trying utf-8 and if the decode fails, putting `--binary--`? `--base64` could be an option if folks want the entire output. Alternatively, if there was a convenient `jq` oneliner to do something similar, that'd probably be okay for now.

hawkw:
@grampelberg 
>    * WDYT about trying utf-8 and if the decode fails, putting `--binary--`? `--base64` could be an option if folks want the entire output. Alternatively, if there was a convenient `jq` oneliner to do something similar, that'd probably be okay for now.

It looks like `jq` can decode base64 with [`@base64d`](https://stedolan.github.io/jq/manual/#Formatstringsandescaping), but I agree that it would be nicer to handle the decoding in the Linkerd CLI --- it's certainly surprising to see the output with all header values base64'd.

adleong:
+100 on decoding those IP addresses.

What do you think of rendering the `value` field with the UTF-8 decoded string value if possible, otherwise rendering a `value-binary` field with an encoding of the binary value.  Do folks like base64 as a binary encoding better than just hex bytes?  I find hex bytes much easier to read, but maybe that's just me.

kleimkuhler:
Okay so after some further discussion, the linkerd2-proxy-api change proposed in linkerd/linkerd2-proxy-api#33 will continue treating header values as `bytes`. The proxy should not attempt to decode these values from the requests and responses.

The public API change proposed in this PR however will treat header values as `string`. This way, the control plane can attempt to base64 decode header values--dropping the header from the JSON encoded output if the base64 decoding fails. In this situation, we can just log that a decoding failure occurred.

grampelberg:
> The public API change proposed in this PR however will treat header values as string. This way, the control plane can attempt to base64 decode header values--dropping the header from the JSON encoded output if the base64 decoding fails. In this situation, we can just log that a decoding failure occurred.

I assume you mean *encode*. This suggests that the client will then need to base64 decode everything, utf-8 encode it and decide what to do if the utf-8 encoding fails?

adleong:
I don't think base64 comes into play anywhere here (except maybe as an under-the-hood implementation detail of the grpc transport?).  The proxy sends header values as a `bytes` typed protobuf field.  The tap controller attempts to decode that field as UTF-8.  If successful, it sends the header value to the CLI through the public API as a `string` typed protobuf field.  If UTF-8 decoding fails, the header is dropped and not reported to the CLI.

kleimkuhler:
The header values are now rendered as `string` if they consist entirely of valid UTF-8 encodings. This PR is rebased off of #3410 and should be merged in after.

Tests have been added and this is now ready for review.

scottcarol:
This is a great addition! Mind adding this to the `Flags` list that show when you input `linkerd tap -h` docs? Would be on line 148 of `tap.go`, currently 

``` -o, --output string         Output format. One of: wide`

**Comment by kleimkuhler** [comment]
> I'm fine with merging this as is to move forward, but I think we need to address the usability of the json output before we can consider this done.

@adleong I'm not sure I understand if you are suggesting further changes to this PR right now, or beginning a discussion for future work.

I agree with your points, and this PR does introduce a difference in how the protobuf is rendered into JSON: nested fields via `oneof` messages (IP addresses) and more readable fields via specific fields names and populating only one (headers).

Would you prefer this change to render header values as `oneof` messages and have a follow-up that addresses the readability as a whole?

**Comment by scottcarol** [comment]
For my own understanding, why is the capitalization different for different key/value pairs? For example:
```
"destination": {
    "ip": {
      "Ip": {
        "Ipv4": 167839439
      }
```

**Comment by kleimkuhler** [comment]
@scottcarol The difference in capitalization when the protobuf is rendered comes from the generated structs. If a field has a tag that contains the `json` key, it follows that naming.

For example:
```
type TapEvent struct {
	Source          *TcpAddress             `protobuf:"bytes,1,opt,name=source,proto3" json:"source,omitempty"`
```

The `Source` field of `TapEvent` has a `json` field in the tag that determines it should be rendered with a lowercase `source`.

The fields that are capitalized do not have a `json` key in their field tags, so I assume it defaults to rendering camel case?

As for why a field does not have a tag specifying JSON behavior, it looks like once you no longer have user defined types (such as `oneof`) it drops that field...

File: cli/cmd/tap.go [hunk 1]
```diff
@@ -292,6 +302,16 @@ func renderTapEvent(event *pb.TapEvent, resource string) string {
 	}
 }
 
+// renderTapEventJSON renders a Public API TapEvent to a string in JSON format.
+func renderTapEventJSON(event *pb.TapEvent, _ string) string {
+	e, err := json.MarshalIndent(event, "", "  ")
```
klingerf: Go's built-in JSON serializer doesn't work very well with Protobuf structs, so the Go implementation of protobuf actually provides its own serializer, here:

https://godoc.org/github.com/golang/protobuf/jsonpb#Marshaler

That marshaler is more configurable, and the JSON objects are flatter than those produced with the built-in serializer. Linkerd's web server is using it to send JSON payloads back to the web client, here:

https://github.com/linkerd/linkerd2/blob/8270ba363cc9c1750d63da2f745ad8c86d69c13f/web/srv/api_handlers.go#L36

And here:

https://github.com/linkerd/linkerd2/blob/8270ba363cc9c1750d63da2f745ad8c86d69c13f/web/srv/api_handlers.go#L62-L65

kleimkuhler:
@klingerf This is awesome and should address the concerns @adleong brought up about the difference in how fields are rendered.

I'd like to handle this in a few steps since these can all be done separately.

I'm going to update the public API change proposed in #3410 so that `Headers`'s value field is now a `oneof`. That way we don't have a `value` and `value_bin` field.

I'm going to rebase this PR to add a JSON output option to the `tap` command, but not worry about the marshaler that we use. This PR will just introduce a JSON output format.

I'm then going to follow-up these changes to address the new issue (#3423) that moves `tap.go` to use `jsonpb` instead of `json`.

adleong:
Unfortunately, this doesn't help with the way that the IP address values themselves are formatted (as a fixed32 instead of a dot delimited octects).

kleimkuhler:
Closing in favor of #3434

PR #3292, opened 2019-08-20
Last changes before submitting to the Helm incubator

alpeb:
- Added missing OWNERS and README files
- Added maintainers section to  Chart.yaml
- Changed NOTES.txt so it points to the installation of the CLI
- Set the proxy-init version to v1.1.0 in values.yaml

Ref #3256

PR #3415, opened 2019-09-10
Add chart_deploy into workflow.yml

alpeb:
Ref linkerd/website#502

Adds a new final step into `workflow.yml` to generate and push the Helm
artifacts.

This depends on the changes made to `bin/helm-build` in #3292 so this
branch is based on that one till it gets merged.


.github/workflows/workflow.yml
```diff
@@ -366,7 +366,8 @@ jobs:
         dir="${CLOUDSDK_INSTALL_DIR:-${HOME}}/google-cloud-sdk"
         (
           . bin/_gcp.sh ;
-          install_gcloud_kubectl "$dir"
+          install_gcloud "$dir"
+          gcloud components install kubectl
```
siggy: we have kubectl available via `bin/kubectl`, so this line is probably not necessary. (i had left it in `_gcp.sh` simply because i didn't want to change too many things when switching to github actions).

if you are calling scripts that expect `kubectl` in the path (like `bin/test-run` and `get_k8s_ctx`), we have this pattern:
https://github.com/linkerd/linkerd2/blob/9bb7b6f119fe3a78f48440241f01947ea3be2791/.github/workflows/workflow.yml#L174

.github/workflows/workflow.yml
```diff
@@ -466,7 +468,8 @@ jobs:
         echo "$CLOUD_SDK_SERVICE_ACCOUNT_KEY" > .gcp.json
         dir="${CLOUDSDK_INSTALL_DIR:-${HOME}}/google-cloud-sdk"
         (. bin/_gcp.sh ;
-          install_gcloud_kubectl "$dir"
+          install_gcloud "$dir"
```
siggy: nit unrelated to your change, mind adding a newline above `bin/_gcp.sh`?
```
        (
          . bin/_gcp.sh ;
          install_gcloud "$dir"
```

.github/workflows/workflow.yml
alpeb: ah cool. I'll remove that then :+1:

alpeb:
@siggy I could finally test this successfully, by setting a fake tag like `edge-1.2.3`, and triggering the build multiple times until `github.ref` decided to use that tag instead of something else. I verified the Helm artifacts were properly uploaded into GCS :+1:, corresponding to that fake tag.

PR #3432, opened 2019-09-14
Fix bad request in the top routes tab on empty fields

pierDipi:
When clicking on the start button on the top routes tab without
filling the namespace and resource dropdown buttons will cause
a bad request error.

Disabling the `Start` button if the `namespace` and the `resource type`
field of the query object of the state are empty strings will solve
the problem.

Signed-off-by: Pierangelo Di Pilato <pierangelodipilato@gmail.com>

### DIFF
--- a/web/app/js/components/TopRoutes.jsx
+++ b/web/app/js/components/TopRoutes.jsx
@@ -208,9 +208,7 @@ class TopRoutes extends React.Component {
               <Button
                 color="primary"
                 variant="outlined"
-                disabled={this.state.requestInProgress ||
-                  this.state.query.namespace === '' ||
-                  this.state.query.resource_type === ''}
+                disabled={this.state.requestInProgress || !this.state.query.namespace || !this.state.query.resource_type}
                 onClick={this.handleBtnClick(true)}>
               Start
               </Button>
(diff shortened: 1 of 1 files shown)


grampelberg:
@pierDipi great catch!

scottcarol:
Hi @pierDipi, thanks for the great PR and the catch. Good 👁! 

Minor nit - we can simplify a tiny bit and keep this consistent with `TapQueryForm` by checking for `!this.state.query.namespace` instead of `this.state.query.namespace === ''`. The full check would read:

```
disabled={this.state.requestInProgress || !this.state.query.namespace || !this.state.query.resource_type}
```

Mind making that small change? Then this should be good to merge 🚀

pierDipi:
Done!

@grampelberg @scottcarol Great project! I hope I will be able to help you soon again.

grampelberg:
@pierDipi we'd love to send you some swag, fill out [the form](https://docs.google.com/forms/d/e/1FAIpQLSdXA7kAl6oyUQ4DFRjiDc2biWlvkhNirrMwsXaQ8V3Uh8Yj5A/viewform) if you're interested!

PR #3433, opened 2019-09-15
proxy: Update to v2.71.0

olix0r:
Now that proxy releases are published to GitHub as named releases,
the `fetch-proxy` and `git-commit-proxy-version` scripts need to be
updated to support the new home for release artifacts and the tagging
scheme used in the linkerd2-proxy repo (namely that tags are of the form
`release/v2.N.P` to properly disambiguate from other tags in the repo.


olix0r:
We won't be able to merge this until we actually have a proxy release published

bin/git-commit-proxy-version
```diff
@@ -3,38 +3,48 @@
...
 rootdir="$( cd $bindir/.. && pwd )"
 
 old_proxy_version="$(tr -d '\n' <"$rootdir/.proxy-version")"
 if [ -z "$old_proxy_version" ]; then
-    echo "missing .proxy-version" >&2
+    echo "Missing .proxy-version" >&2
     exit 1
 fi
 
+# We temporarily have to handle aa SHA version.
```
adleong: aa typo

adleong: Do we still have to handle SHA versions, as of this commit?

olix0r: Eh the code still does because it had to generate this diff... It doesn't really have to after this, though, so I'll touchup.

bin/git-commit-proxy-version
```diff
@@ -3,38 +3,48 @@
...
+old_proxy_rev="$old_proxy_version"
+if echo "$old_proxy_version" | grep '^v2.'; then
+    old_proxy_rev="release/$old_proxy_version"
```
ihcsim: If this is the older proxy revision scheme, shouldn't it be `linkerd/`, instead of `release/`?

bin/git-commit-proxy-version
```diff
@@ -3,38 +3,48 @@
...
 tmp=$(mktemp -d -t l2-proxy.XXX)
 git clone --depth=100 https://github.com/linkerd/linkerd2-proxy "$tmp"
 cd "$tmp"
 
-sha="$(git rev-parse --verify --quiet "${new_proxy_revision}")"
-new_proxy_version="${sha:0:7}"
-
-if ! git rev-parse --verify --quiet "${old_proxy_version}" ; then
+if ! git rev-parse --verify --quiet "${old_proxy_rev}" ; then
     echo "The old proxy-version ${old_proxy_version} does not exist in the last 100 proxy commits." >&2
```
ihcsim: I think this should now be `${old_proxy_rev}`.

bin/git-commit-proxy-version
olix0r: No, "old" here means "prior" (as in, the one we read from .proxy-version)

bin/git-commit-proxy-version
olix0r: It could, but my thought is that this value is that in `.proxy-version`, so it's what the user interfaces with.

bin/git-commit-proxy-version
olix0r: Also, we'll likely continue to add linkerd2 tags on the proxy repo so it's clear how versions correlate (i.e. the tag release/v2.72.0 will likely be identical to linkerd2/edge-19.9.3)

PR #3440, opened 2019-09-17
Refactor stat summary promql 'group-by all' code

### LINKED ISSUE #3399: N+1 queries to prometheus
### What is the issue?

When the `public-api` service receives a request to `StatSummary`, there can be a list of resources. This ends up regularly containing many resource types such as deployments and pods. Each resource type fires off its own query, creating a 10x increase in the number of requests that are issued to prometheus.

See [stat_summary.go](https://github.com/linkerd/linkerd2/blob/master/controller/api/public/stat_summary.go#L107) for the code that is creating the n+1 requests.

ihcsim:
This PR refactors the code around supporting `linkerd stat all` by:

1. Changing the PromQL to use the `by(namespace, deployment, pod, service, trafficsplit, authority, daemonset, statefulset)` group by condition, replacing the existing group by individual type conditions. 
1. Replacing the multiple goroutines that issue multiple concurrent queries, with a single goroutine to issue the new query.
1. Updating the code that builds the `StatTable` rows.
1. Adding some unit tests to provide better visibility into the generation of the labels, group-by conditions and map keys that are used to combine the k8s stats and Prometheus stats.

In addition to passing the existing unit and integration tests, the output of the following commands are compared with that from `edge-19.9.3`:
```sh
$ linkerd stat all
$ linkerd stat all -n linkerd
$ linkerd stat all -n emojivoto
$ linkerd stat all -A
$ linkerd stat deploy -n emojivoto
$ linkerd stat po -n emojivoto
$ linkerd stat job -n emojivoto-job
$ linkerd stat sts -n emojivoto-sts
$ linkerd stat ns
$ linkerd stat authority
$ linkerd stat authority -A
$ linkerd stat ts -n emojivoto-ts
$ linkerd stat ns --from ns/linkerd
$ linkerd stat ns --from ns/emojivoto-ts
$ linkerd stat po --to deploy/linkerd-prometheus --to-namespace linkerd -A
$ linkerd stat po --from deploy/web --from-namespace emojivoto -A
```

The `diff` between the `edge-19.9.3` PromQL and those generated by this PR can be found in this [gist](https://gist.github.com/ihcsim/b65b8a2027c655981b50e78715b542d4), organized by the different CLI commands. (TL;DR the number of queries generated by `linkerd stat all [-n <namespace] [-A]` is significantly reduced. The queries generated by `linkerd stat `resource-type` remains unchanged.)

Fixes #3399


adleong:
Thanks, @ihcsim!  There's a lot changing here so it's a bit intimidating to review.  Is it possible to split this into smaller pieces?  In particular separating:
* changes to the prometheus queries we're making (with examples in the description of the old queries vs the new queries)
* changes to the ways we're making the queries (parallel vs serial)
* organizational refactors that don't change behavior

Or at the least, describing in more detail which changes fall into which of the above categories.

scottcarol:
Hey @ihcsim, this is looking good! Are you planning on pushing more changes or would you like this to be reviewed as it is currently?

controller/api/public/stat_summary.go
```diff
@@ -92,50 +95,65 @@ func (s *grpcServer) StatSummary(ctx context.Context, req *pb.StatSummaryRequest
...
+		return nil, util.GRPCError(status.Errorf(codes.Unimplemented, "unimplemented resource type: %s", kind))
+	}
 
-	var resourcesToQuery []string
-	if req.Selector.Resource.Type == k8s.All {
-		resourcesToQuery = k8s.StatAllResourceTypes
-	} else {
-		resourcesToQuery = []string{req.Selector.Resource.Type}
+	// get stats for individual workload kind
+	if isK8sWorkloadKind(kind) {
```
alpeb: Maybe there's some DRYness to be gained here if we reuse `getStatsForAllKinds()` instead of `s.k8sResourceQuery()`? The former could receive the list of kinds to query, which would be just `kind` in this case, and `k8s.StatAllWorkloadKinds` in `kind == k8s.All`'s case...

controller/api/public/prometheus.go
```diff
@@ -64,27 +65,49 @@ func (s *grpcServer) queryProm(ctx context.Context, query string) (model.Vector,
...
+		return names
+	}
 
-	if resource.Type != k8s.Namespace {
-		names = append(names, promResourceType(resource))
+	if resource.Type == k8s.All {
+		for _, kind := range k8s.StatAllKinds {
+			clone := proto.Clone(resource).(*pb.Resource)
+			clone.Type = kind
+			names = append(names, promResourceType(clone))
```
alpeb: What do you think about providing an alternative to `promResourceType()`  that just receives the kind? That way we avoid the cloning here and below.

controller/api/public/prometheus.go
```diff
@@ -64,35 +64,53 @@ func (s *grpcServer) queryProm(ctx context.Context, query string) (model.Vector,
...
-		names = append(names, promResourceType(resource))
-	} else if resource.Type != k8s.Namespace {
-		names = append(names, "dst_"+promResourceType(resource))
+		return append(names, promResourceType(resource.Type))
 	}
-	return names
+
+	if resource.Type == k8s.All {
+		for _, kind := range k8s.StatAllWorkloadKinds {
+			if isNonK8sResourceQuery(kind) {
```
ihcsim: If kind is authority, don't add the `dst_` prefix to the label.

controller/api/public/prometheus.go
```diff
@@ -64,35 +64,53 @@ func (s *grpcServer) queryProm(ctx context.Context, query string) (model.Vector,
...
 func promGroupByLabelNames(resource *pb.Resource) model.LabelNames {
 	names := model.LabelNames{namespaceLabel}
+	if resource.Type == k8s.Namespace {
+		return names
+	}
 
-	if resource.Type != k8s.Namespace {
-		names = append(names, promResourceType(resource))
+	if resource.Type == k8s.All {
+		for _, kind := range k8s.StatAllWorkloadKinds {
```
ihcsim: This loop only iterates through k8s workload kinds like deployment, statefulset, job etc. Traffic split and authority kinds are handled outside of `k8s.All`. See changes below: https://github.com/linkerd/linkerd2/blob/2cf17da3d39befd5cc9531be708fdf73d3fff06a/controller/api/public/stat_summary.go#L107-L135

controller/api/public/stat_summary.go
```diff
@@ -92,50 +97,46 @@ func (s *grpcServer) StatSummary(ctx context.Context, req *pb.StatSummaryRequest
...
-	statTables := make([]*pb.StatTable, 0)
-
-	var resourcesToQuery []string
-	if req.Selector.Resource.Type == k8s.All {
-		resourcesToQuery = k8s.StatAllResourceTypes
-	} else {
-		resourcesToQuery = []string{req.Selector.Resource.Type}
+	kind := req.GetSelector().GetResource().GetType()
+	if !isSupportedKind(kind) {
+		return nil, util.GRPCError(status.Errorf(codes.Unimplemented, "unimplemented resource type: %s", kind))
```
ihcsim: Previously, we rely on [`k8sAPI.Objects()`](https://github.com/linkerd/linkerd2/blob/49d39e5a123ecf40e12d51f28d4185eb1bf7512f/controller/api/public/stat_summary.go#L162) to detect unsupported resource type. This change allows us to fail early, without creating any additional goroutines.

controller/api/public/stat_summary.go
```diff
@@ -585,24 +620,32 @@ func processPrometheusMetrics(req *pb.StatSummaryRequest, results []promResult,
...
-		Type: req.GetSelector().GetResource().GetType(),
-		Name: string(metric[groupBy[len(groupBy)-1]]),
+// metricToKey generates a key which is used to match the metric stats we
+// queried from prometheus with the k8s object stats we queried from k8s
+func metricToKey(kind string, outboundFrom bool, metric model.Metric) rKey {
+	kindLabel := model.LabelName(k8s.KindToStatsLabel(kind, outboundFrom))
+	namespaceLabel := model.LabelName(k8s.KindToStatsLabel(k8s.Namespace, outboundFrom))
+
+	if _, exists := metric[kindLabel]; !exists {
+		log.Debugf(debugTxtMismatchLabel, "label", kindLabel, metric)
```
ihcsim: I made this `Debugf` because if there is a mismatch between the keys used in the k8s stats and Prometheus stats maps, it's more than likely a programming error. There is nothing a user can do about it, but to file GH issue.

ihcsim:
@siggy I chatted with @adleong about splitting up this PR. Unfortunately, most of the changes in this PR is related to combining the workload queries into a single `group-by`. I don't think a separate `group by` only PR will be significantly smaller. Functions like `getStatMetrics()`, `buildRequestLabel()`, `metricToKey()` and the unexported functions they called in `prometheus.go` are shared between the authority/traffic split and k8s resources queries. To separate this into multiple PRs, I found myself having to add some temporary code to support both the existing and new ways of querying and outputting the result, and the context of why certain changes are needed is also lost.

controller/api/public/prometheus.go
alpeb: But `k8s.StatAllWorkloadKinds` doesn't contain `Authority`, so this will always be false, won't it?

controller/api/public/stat_summary.go
alpeb: From what I can tell, the tcp family of metrics (`tcp_read_bytes`, `tcp_write_bytes`, `tcp_open_connections`) don't have `dst_*` labels, so they raise these errors. Maybe they deserve special treatment here.

controller/api/public/prometheus.go
ihcsim: Yes, you are right. The `Authority` case is already handled in linnke 90 above, so this is not needed. Thanks.

controller/api/public/stat_summary.go
ihcsim: Thanks for pointing this out. These two checks don't make sense because, e.g.:

```
# given this query
sum(increase(tcp_read_bytes_total{direction="inbound", namespace="emojivoto"}[1m])) by (namespace, daemonset, statefulset, k8s_job, deployment, replicationcontroller, pod, service)

# ... the response looks like

Element | Value
------- | --
{k8s_job="voting-job",namespace="emojivoto",pod="voting-job-jj5ms"}       | 7323.6
{k8s_job="web-job",namespace="emojivoto",pod="web-job-fpn8z"}             | 283312.8
{deployment="emoji",namespace="emojivoto",pod="emoji-697b575bd9-rctbl"}   | 145273.2
{deployment="voting",namespace="emojivoto",pod="voting-6b4bf7494b-m7767"} | 7274.4
{deployment="web",namespace="emojivoto",pod="web-559684dbc5-8v45z"}       | 283203.6
{k8s_job="emoji-job",namespace="emojivoto",pod="emoji-job-rjzjk"}         | 142485.59999999998
```
So in the case of `k8s.All`, this function will be called by the `processPrometheusMetrics()` function as it iterates through the different kinds with `metric` that belongs to other workload kinds. I'm removing these `Debugf` statements.

adleong:
Do we have any sense of the performance implications of this change?  Is there a measurable performance difference (in terms of total latency, CPU utilization, and/or memory utilization) from making one large query instead of several smaller queries?

ihcsim:
@adleong No official data points. In my last test setup, the difference isn't huge. I am thinking we can put this on the back burner for now, until more users have a chance to try out the new dashboard. I am also a bit concerned about merging this PR, when https://github.com/linkerd/linkerd2/issues/3562 remains open. (I assume that part of the dashboard relies on the code in this PR.)

PR #3442, opened 2019-09-18
Fix injector timeout under high load

### LINKED ISSUE #3358: Investigate into proxy injector admission timeout
When injecting a workload with many replicas, many of the replicas appear to be uninjected during the admission phase. We have reasons to believe that this is caused by the [default 30 sec timeout](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#mutatingwebhook-v1beta1-admissionregistration-k8s-io) imposed on the webhook call, by the k8s API server. At present, the 30 seconds interval is the maximum allowed values. (See https://github.com/kubernetes/kubernetes/issues/60914).

We need to investigate into:

* if the timeout applies to each admission request, or the entire workload creation 
* if there are any bottlenecks within the proxy injector code that is causing the timeout
* if the k8s API server behave as expected, when under load

To reproduce:

* Deploy a nginx deployment with 1000 replicas

Relevant API server logs:
```
W0830 17:53:24.316506       1 dispatcher.go:136] Failed calling webhook, failing open linkerd-proxy-injector.linkerd.io: failed calling webhook "linkerd-proxy-injector.linkerd.io": Post https://linkerd-proxy-injector.linkerd.svc:443/?time
out=30s: context deadline exceeded (Client.Timeout exceeded while awaiting headers)                                                                                                                                                          
E0830 17:53:24.316536       1 dispatcher.go:137] failed calling webhook "linkerd-proxy-injector.linkerd.io": Post https://linkerd-proxy-injector.linkerd.svc:443/?timeout=30s: context deadline exceeded (Client.Timeout exceeded while awaiti
ng headers)
```

Relevant #3290.

alpeb:
Fixes #3358

When retrieving a pod owner, we were hitting the k8s API directly because
at injection time the informer might not have been informed about the
existence of the parent object.
Under a large number of injection requests this ended up in the k8s API requests
being throttled, the proxy-injector getting blocked and the webhook requests
timing out.

Now we'll hit the shared informer first, and hit the k8s API only when
the informer doesn't return anything. After a few injection requests for
the same owner, the informer should have been updated.

## Testing

Scaling an emoji deployment to 1000 replicas, and after waiting for a
couple of minutes:

Before:
```bash
# a portion of the pods doesn't get injected
$ kubectl-n emojivoto get po | grep ./1 | wc -l
109

kubectl -n kube-system logs -f kube-apiserver-minikube | grep
failing.*timeout
.... (lots of errors)
```

After:
```bash
# all the pods get injected
$ kubectl -n emojivoto get po | grep ./1 | wc -l
0

kubectl -n kube-system logs -f kube-apiserver-minikube | grep
failing.*timeout
# no errors
```

### DIFF
--- a/controller/k8s/api_test.go
+++ b/controller/k8s/api_test.go
@@ -16,10 +16,10 @@ import (
 )
 
 // newAPI constructs a mock controller/k8s.API object for testing
-// enableInformers: forces informer indexing, enabling informer lookups
+// retry: forces informer indexing, enabling informer lookups
 // resourceConfigs: resources available via the API, and returned as runtime.Objects
 // extraConfigs:    resources returned as runtime.Objects
-func newAPI(enableInformers bool, resourceConfigs []string, extraConfigs ...string) (*API, []runtime.Object, error) {
+func newAPI(retry bool, resourceConfigs []string, extraConfigs ...string) (*API, []runtime.Object, error) {
 	k8sConfigs := []string{}
 	k8sResults := []runtime.Object{}
 
@@ -39,7 +39,7 @@ func newAPI(enableInformers bool, resourceConfigs []string, extraConfigs ...stri
 		return nil, nil, fmt.Errorf("NewFakeAPI returned an error: %s", err)
 	}
 
-	if enableInformers {
+	if retry {
 		api.Sync()
 	}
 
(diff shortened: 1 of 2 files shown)


controller/k8s/api_test.go
```diff
@@ -974,19 +974,19 @@ metadata:
...
 		tt := tt // pin
-		for _, enableInformers := range []bool{
+		for _, retry := range []bool{
 			false,
 			true,
 		} {
-			enableInformers := enableInformers // pin
-			t.Run(fmt.Sprintf("%d/enableInformers:%t", i, enableInformers), func(t *testing.T) {
-				api, objs, err := newAPI(enableInformers, []string{tt.podConfig}, tt.extraConfigs...)
+			retry := retry // pin
```
KIVagant: Just if you have time, could you explain what the pinning does? Thank you.

alpeb: The call below (`t.Run(fmt.Sprintf("%d/retry:%t", i, retry), func(t *testing.T)`...) uses `retry` inside the closure in that second argument. If we don't "pin" that variable, when those two closures run it's possible they will use the current value of `retry`, which in this case would be the last value of the array, `true`. If we "pin" it, then each closure will instead grab a copy of that value, a different copy for each iteration.
I've created a simpler example [here](https://play.golang.org/p/W0imMglYA-U).

KIVagant: So I got it right and it's to avoid race conditions but just the name of the var within the closure visibility scope is the same as the name of the var created outside of the visibility scope. And Golang allows to define both of them together.

KIVagant: Actually this is amazing comparing to other languages even though it looks tricky first. Thank you for the explanation.

PR #3451, opened 2019-09-19
Update CHANGES.md for edge-19.9.3

alpeb:
## edge-19.9.3

* Helm
  * Allow disabling namespace creation during install (thanks @KIVagant!)
* CLI
  * Added a new `json` output option to the `linkerd tap` command
* Controller
  * Fixed proxy injector timeout during a large number of concurrent injections
  * Separated the destination controller into its own separate pod
  * Updated Prometheus config to keep only needed `cadvisor` metrics,
    substantially reducing the number of time-series stored in most clusters
* Web UI
  * Fixed bad request in the top routes tab on empty fields (thanks @pierDipi!)
* Proxy
  * Fixes to the client's backoff logic
  * Added 587 to the list of ports to disable protocol detection (thanks @brianstorti!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,14 +6,15 @@
   * Added a new `json` output option to the `linkerd tap` command
 * Controller
   * Fixed proxy injector timeout during a large number of concurrent injections
-  * Separated the destination controller into its own separate pod
+  * Separated the destination controller into its own separate deployment
   * Updated Prometheus config to keep only needed `cadvisor` metrics,
     substantially reducing the number of time-series stored in most clusters
 * Web UI
   * Fixed bad request in the top routes tab on empty fields (thanks @pierDipi!)
 * Proxy
   * Fixes to the client's backoff logic
-  * Added 587 to the list of ports to disable protocol detection (thanks @brianstorti!)
+  * Added 587 (SMTP) to the list of ports to ignore in protocol detection (bound
+    to server-speaks-first protocols) (thanks @brianstorti!)
 
 ## edge-19.9.2
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,17 @@
+## edge-19.9.3
+
+* Helm
+  * Allow disabling namespace creation during install (thanks @KIVagant!)
+* Controller
+  * Fixed proxy injector timeout during a large number of concurrent injections
+  * Separated the destination controller into its own separate pod
+  * Updated Prometheus config to keep only needed cadvisor metrics
+* Web UI
+  * Fixed bad request in the top routes tab on empty fields (thanks @pierDipi!)
+* Proxy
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,17 @@
...
+  * Allow disabling namespace creation during install (thanks @KIVagant!)
+* Controller
+  * Fixed proxy injector timeout during a large number of concurrent injections
+  * Separated the destination controller into its own separate pod
+  * Updated Prometheus config to keep only needed cadvisor metrics
+* Web UI
+  * Fixed bad request in the top routes tab on empty fields (thanks @pierDipi!)
+* Proxy
+  * Fixes to the client's backoff logic
+  * Add 587 to the list of ports to disable protocol detection
```
olix0r: fwiw, this was an external PR too https://github.com/linkerd/linkerd2-proxy/pull/350, if you want to call out the author

CHANGES.md
```diff
@@ -1,3 +1,17 @@
+## edge-19.9.3
+
+* Helm
+  * Allow disabling namespace creation during install (thanks @KIVagant!)
+* Controller
+  * Fixed proxy injector timeout during a large number of concurrent injections
+  * Separated the destination controller into its own separate pod
+  * Updated Prometheus config to keep only needed cadvisor metrics
```
olix0r: this is okay, but maybe clearer wrt user impact as: "Updated Prometheus config to keep only needed cadvisor metrics, substantially reducing the number of time-series stored in most clusters"

CHANGES.md
adleong: might be nice to call out why.  something to do with smtp?

CHANGES.md
```diff
@@ -1,3 +1,20 @@
+## edge-19.9.3
+
+* Helm
+  * Allow disabling namespace creation during install (thanks @KIVagant!)
+* CLI
+  * Added a new `json` output option to the `linkerd tap` command
+* Controller
+  * Fixed proxy injector timeout during a large number of concurrent injections
+  * Separated the destination controller into its own separate pod
```
ihcsim: s/pod/deployment

CHANGES.md
alpeb: Yeah, just another common port for SMPT.  @adleong does this look fine for you?
```
Added 587 (SMTP) to the list of ports disabled by default (bound to server-speaks-first protocols)
```

adleong: :+1:

CHANGES.md
```diff
@@ -1,3 +1,21 @@
+## edge-19.9.3
+
+* Helm
+  * Allow disabling namespace creation during install (thanks @KIVagant!)
```
scottcarol: I think this should be "allowed"

PR #3457, opened 2019-09-20
Make Identity use GRPC Server with Prom Metrics

Pothulapati:
All the control plane components use `prometheus.NewGrpcServer()` which initialises the grpc server with some grpc metrics which are useful.

This also gives a single place where other handlers (Ex OC Grpc handler) can be added directly.

### DIFF
--- a/controller/cmd/identity/main.go
+++ b/controller/cmd/identity/main.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"github.com/linkerd/linkerd2/pkg/prometheus"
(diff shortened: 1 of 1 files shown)


controller/cmd/identity/main.go
```diff
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"github.com/linkerd/linkerd2/pkg/prometheus"
+
```
olix0r: prefer to omit the newline so it sorts with other imports

PR #3458, opened 2019-09-21
Add script to load images into kind cluster

kleimkuhler:
## Summary

[kind](https://github.com/kubernetes-sigs/kind) has been a helpful tool for running local Kubernetes clusters and
testing linkerd builds. Once images are built with `bin/docker-build`, the
images must be loaded into the kind cluster.

This script should be run after `bin/docker-build` and will load the images into
the specified kind cluster.

Example:
```
$ bin/docker-build
$ kind get clusters # show available clusters to load images on to
kleimkuhler
$ bin/kind-load kleimkuhler
$ ./target/cli/linux/linkerd install | kubectl apply -f -
```

### DIFF
--- a/bin/kind-load
+++ b/bin/kind-load
@@ -2,12 +2,8 @@
 
 set -eu
 
-if [ $# -eq 1 ]; then
-    cluster="${1:-}"
-else
-    echo "usage: $(basename $0) cluster-name" >&2
-    exit 64
-fi
+# If cluster name is unset or null, default to $USER.
+cluster="${1:-$USER}"
 
 bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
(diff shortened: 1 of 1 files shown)


alpeb:
Thanks @kleimkuhler , I'll be using this extensively. Since we already have `bin/kind` what do you think about leveraging that in this script?

bin/kind-load
```diff
@@ -0,0 +1,25 @@
...
+tag="$(head_root_tag)"
+org_repo="gcr.io/linkerd-io"
+
+kind load docker-image $org_repo/proxy:$tag --name $cluster
+kind load docker-image $org_repo/controller:$tag --name $cluster
+kind load docker-image $org_repo/web:$tag --name $cluster
+kind load docker-image $org_repo/cni-plugin:$tag --name $cluster
+kind load docker-image $org_repo/debug:$tag --name $cluster
+kind load docker-image $org_repo/cli-bin:$tag --name $cluster
+kind load docker-image $org_repo/grafana:$tag --name $cluster
```
siggy: a bit more succinct:
```bash
for img in controller grafana proxy web ; do
  bin/kind load docker-image $org_repo/$img:$tag --name=$cluster
done
```

also, `cli-bin`, `cni-plugin`, and `debug` are not required for a normal installation, and are also large images which slow down this process. how about making them optional (or not including them at all)?

bin/kind-load
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -eu
+
+if [ $# -eq 1 ]; then
+    cluster="${1:-}"
```
siggy: what do you think supporting a default? `kind` itself uses `default` as the cluster name when it's not provided. personally i use `$USER`.

bin/kind-load
kleimkuhler: Oh good call yea I'll remove the non-required ones.

kleimkuhler:
@siggy Thanks for the suggestions.

@alpeb I added an execution of `bin/kind` to ensure kind is installed on the user's cluster.

PR #3461, opened 2019-09-23
Trace Control Plane components using OC

Pothulapati:
This PR uses OpenCensus Libraries especially `ocgrpc` & `ochttp` to trace the major grpc and http calls. 

We introduce two new flags with the defaults:

`trace-collector`:          This specified the URL at wihch these spans are sent.
`sampling-probability`: This specifies the rate at which sampling has to be done. 0.1 imples to collect 1 in every 10 (0 implies none, and 1.0 includes all spans)

These installation flags are also part of the helm chart values. The probability will default to 0, i.e to collect no spans. 
These flags are added to the command args in the `linkerd/charts/templates` folder.

These values can be overriden by 
```
linkerd install --trace-collector "oc-tracing:5434" --trace-probability 1
```
As the installation flags are stored in the `linkerd-config` cm Upgrades, etc work as expected.

**How does this work:**

- We use a new function `flags.AddTraceFlags(cmd)` which adds the above mentioned flags and returns their values to be used.
- We also use `util.InitialiseTracing("linkerd-destination", *traceCollector, *probabilisticSamplingRate)`, which registers exporters and also initialises the trace config as per the flags.

Both of these functions are used in the `main` functions of the components.

**Components Traced:**

- All the server side components i.e `linkerd-web`, `linkerd-destination`, `linkerd-tap`, `linkerd-public-api`, `linkerd-identity` that use `prometheus.newGrpcServer()` will have the `ocgrpc` handler added, which automatically sends the request spans.
- As `linkerd-public-api` and `linkerd-web` components have http handlers, they are also traced using `ochttp` handler at the server side. Along with `public-API's` http client to pass the context further.
- Prometheus queries are also traced whenever  `queryProm()` is called.

**Questions**

- As grpc calls especically that of `Identity`, `Destination` are not adding any contextual information.  Are the initial request spans useful? If not, we can remove them. Contextual information is added by using interceptors which would be another PR.
- Is the Information in the prometheus query spans enough? If not, what can be added?

**What doesn't work:**
- The spans from `linkerd-web -> linkerd-tap` are seen as induvidual spans i.e spearate traces because the http client from the `linkerd-web` component uses the k8s transport for communication, so we cant attach the ochttp transport.
 

To be reviewed by, @adleong @grampelberg @ihcsim @alpeb

### DIFF
--- a/controller/api/public/prometheus.go
+++ b/controller/api/public/prometheus.go
@@ -49,9 +49,7 @@ func (s *grpcServer) queryProm(ctx context.Context, query string) (model.Vector,
 
 	_, span := trace.StartSpan(ctx, "query.prometheus")
 	defer span.End()
-	span.Annotate([]trace.Attribute{
-		trace.StringAttribute("queryString", query),
-	}, "Prometheus Query Details")
+	span.AddAttributes(trace.StringAttribute("queryString", query))
 	// single data point (aka summary) query
 	res, err := s.prometheusAPI.Query(ctx, query, time.Time{})
 	if err != nil {
--- a/controller/api/public/prometheus.go
+++ b/controller/api/public/prometheus.go
@@ -8,12 +8,11 @@ import (
 	"strings"
 	"time"
 
-	"go.opencensus.io/trace"
-
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/prometheus/common/model"
 	log "github.com/sirupsen/logrus"
+	"go.opencensus.io/trace"
 )
 
 type promType string
--- a/controller/api/public/prometheus.go
+++ b/controller/api/public/prometheus.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"go.opencensus.io/trace"
+
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/prometheus/common/model"
@@ -46,6 +48,11 @@ func extractSampleValue(sample *model.Sample) uint64 {
 func (s *grpcServer) queryProm(ctx context.Context, query string) (model.Vector, error) {
 	log.Debugf("Query request:\n\t%+v", query)
 
+	_, span := trace.StartSpan(ctx, "query.prometheus")
+	defer span.End()
+	span.Annotate([]trace.Attribute{
+		trace.StringAttribute("queryString", query),
+	}, "Prometheus Query Details")
 	// single data point (aka summary) query
 	res, err := s.prometheusAPI.Query(ctx, query, time.Time{})
 	if err != nil {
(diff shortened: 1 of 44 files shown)


Pothulapati:
**Linkerd-Web** (which talks to the `public-api`)

web-1
web-2
web-3

**Linkerd-Destination**
destination

**Linkerd-Identity**

identity-1

controller/api/public/prometheus.go
```diff
@@ -46,6 +47,11 @@ func extractSampleValue(sample *model.Sample) uint64 {
 func (s *grpcServer) queryProm(ctx context.Context, query string) (model.Vector, error) {
 	log.Debugf("Query request:\n\t%+v", query)
 
+	_, span := trace.StartSpan(ctx, "query.prometheus")
+	defer span.End()
+	span.Annotate([]trace.Attribute{
```
adleong: This ends up showing as a log event in Jaeger rather than a tag.  Not a big deal either way, but it's a bit surprising.  I would definitely have expected this data to be a tag.

pkg/util/trace.go
```diff
@@ -0,0 +1,23 @@
+package util
+
+import (
+	"log"
+
+	"contrib.go.opencensus.io/exporter/ocagent"
+	"go.opencensus.io/trace"
+)
+
+// InitialiseTracing initialises trace, exporter and the sampler
```
adleong: We prefer the "Initialize" spelling

pkg/util/trace.go
```diff
@@ -0,0 +1,23 @@
...
+func InitialiseTracing(serviceName string, address string, probability float64) {
+	oce, err := ocagent.NewExporter(
+		ocagent.WithInsecure(),
+		ocagent.WithAddress(address),
```
adleong: What happens when the address flag is not provided?  It looks like the default value of "" is provided here, which seems to turn off tracing as we would want.  A bit surprising that this doesn't trigger the fatal error below though.  A safer thing might be to just return right away if the address is "".

pkg/util/trace.go
```diff
@@ -0,0 +1,23 @@
...
+		ocagent.WithServiceName(serviceName))
+	if err != nil {
+		log.Fatalf("Failed to create ocagent-exporter: %v", err)
```
adleong: I don't think we want to panic if we can't create an exporter.  Logging at Error level should be fine.

adleong:
It looks like you have some failing unit tests because the output manifests changed.  You can run the tests with the `-update` flag to automatically update the expected manifests.

controller/api/public/prometheus.go
Pothulapati: Moved it as a tag

pkg/util/trace.go
adleong: just for consistency's sake, I think we should spell this method as "InitializeTracing"

Pothulapati:
Also, when the `trace-probability` >= 1, it sends all spans and dosent break. It dosent send any spans when its negative as expected.

PR #3464, opened 2019-09-23
Increase the Grafana dashboards refresh interval

### LINKED ISSUE #3456: Increase Grafana refresh interval from 5s to 10s
This should help to reduce the CPU and memory load on Prometheus.

Relevant #3258.

ihcsim:
This PR increases the refresh interval of all workload dashboards from 5s to 10s. For dashboards that query for cluster-level or non-Linkerd resources, I set the refresh interval to 1m, following the default of the Prometheus and Prometheus benchmark dashboards.

See Grafana [docs](https://grafana.com/docs/reference/dashboard/#json-fields) for field description.

Fixes #3456.

### DIFF
--- a/grafana/dashboards/authority.json
+++ b/grafana/dashboards/authority.json
@@ -1171,7 +1171,7 @@
       "type": "text"
     }
   ],
-  "refresh": "10s",
+  "refresh": "1m",
   "schemaVersion": 18,
   "style": "dark",
   "tags": [
--- a/grafana/dashboards/authority.json
+++ b/grafana/dashboards/authority.json
@@ -1171,7 +1171,7 @@
       "type": "text"
     }
   ],
-  "refresh": "5s",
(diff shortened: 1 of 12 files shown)


ihcsim:
@grampelberg I was **just** discussing with @alpeb if we should make it 1m for all dashboards, to see a difference. It does sound like a reasonable tradeoff between resource usage and data staleness. I will apply that change.

alpeb:
I was checking whether this interval had anything to do with the `datasources.jsonData.timeInterval` property in Grafana's `datasource.yaml` config file (currently set at 5s), and it turns out that's just for specifying the resolution of the returned series (e.g. a sample every 5s).

Tapping `linkerd-prometheus` with Grafana opened proved me that the `refresh` property modified in this PR does indeed control how frequently the Prometheus API query endpoints get called :+1:

PR #3467, opened 2019-09-24
Redesign of sidebar

### LINKED ISSUE #3400: Simplify dashboard sidebar

The dashboard's sidebar has a couple current problems:

- It is a little cluttered.
- Information is hidden by default.
- Filtering is difficult.
- Load on prometheus is excessive.

The Kubernetes dashboard does a good job of solving these problems by:

- Listing the possible resources types, not showing every resource.
- Filtering to namespaces by default, instead of showing everything at once.
- Not using trees to hide information.

Towards this end, I've thrown a mockup together of how the sidebar could look. Check out [the repo](https://github.com/grampelberg/linkerd2-mockups/tree/master/sidebar) to run it yourself, or check out the screenshots:

Screenshot 2019-09-12 11 58 22

- Namespaces shows the traffic aggregated by namespace and should be the default view.
- Nodes doesn't exist yet, so ignore that for now.
- Control plane is the same as the service mesh view currently. That'll get refreshed into something a little different later.
- Dropdown shows the available namespaces. There is an `All Namespaces` option that should not be the default. See further down for a screenshot.
- All workloads go to the same page as `All` right now, without the namespace column.
- Ingress, service, service accounts, service profiles do not exist yet and can be ignored.

Screenshot 2019-09-12 11 58 46

- Tools are a 1:1 with what's there now. It would be interesting to scope these to the namespace selector, but let's not do that yet.
- About is a new page that has the version and whether the system is up to date.

Screenshot 2019-09-12 11 59 06

### Examples of Kubernetes dashboard

Screenshot 2019-09-06 15 23 07

Screenshot 2019-09-06 15 23 27

scottcarol:
Fixes #3400. 

Simplifies dashboard. Major changes are:
* Workloads are viewed by namespace.
* Individual resources are not listed on the sidebar.
* User can select "All Namespaces".
* If user is looking at `/namespaces/linkerd/pods`, which shows a list of all pods in `linkerd`, and selects the `emojivoto` namespace from the sidebar, they should see `/namespaces/emojivoto/pods`. (Their view should not change if they are looking at resource detail page.)
* Sidebar collapses completely with narrow width.
* New Kubernetes SVG icons for workloads.

Since this redesign involves changes to routing and paths, help is wanted in confirming that all of the data display pages are intact.

**Note on Context in React**
Although we already use React’s [Context API](https://reactjs.org/docs/context.html) to pass information down throughout the app, we are using it in a new way because the user can now update the `selectedNamespace` via the `Namespace` button, which is propagated to all components. 

In order to update context from the bottom up, I restructured `index.js` to pass `<AppContext.Provider value={this.state}>`
into the app. Previously we passed a static `context` object. 

Passing `{this.state}` into the app as context allows us to pass down `updateNamespaceInContext()`.

I am moderately sure I structured this in the correct way, and I modeled it on the React docs, but I would love feedback on this. I was getting an `eslint-plugin-react` `no-unused-state` error on line 44 of `index.js` when I structured state like so:
```
 this.state = {
      ...appData,
      api: ApiHelpers(pathPrefix),
      productName: "linkerd",
    };
```
I found [a Github issue](https://github.com/yannickcr/eslint-plugin-react/issues/1910) about this but there didn't seem to be a clear resolution, since this is the pattern explicitly suggested by the React Context docs. To work around the lint error I put the different state variables on different lines but I am researching a better solution.


alpeb:
This looks awesome to me :+1:  :1st_place_medal: 

I haven't looked at the code yet, so I only have two visual remarks/questions:
- The k8s icons look much smaller than the others, is that intended?
- I did like how the sidebar used to collapse into just icons. Is that really not worth the space and so it was decided to completely collapse instead?

scottcarol:
Thanks @alpeb! 

Re: icon sizes, no, not intentional. We're now using 3 different libraries for our icons (Material, Font Awesome and K8s). I'll work on making the sizes uniform.

Re: sidebar, when I talked with @grampelberg his preference was to have the sidebar visible or invisible, no middle collapsed version. My sense is now that we are using so many sidebar icons, AND we are letting the user pick a namespace, the collapsed dashboard starts to make less sense (is a user going to know what the K8s "Pods" icon is without a title?). Also certain items like the "Choose Namespace" button would have to disappear/change.

web/app/js/index.js
```diff
@@ -32,29 +31,62 @@ if (proxyPathMatch) {
...
-                render={props => <Navigation {...props} ChildComponent={NamespaceLanding} />} />
+              <Redirect exact from={`${pathPrefix}/`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/overview`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/deployments`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/trafficsplits`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/daemonsets`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/statefulsets`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/jobs`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/replicationcontrollers`} to={`${pathPrefix}/namespaces`} />
+              <Redirect exact from={`${pathPrefix}/pods`} to={`${pathPrefix}/namespaces`} />
```
alpeb: Why not instead of redirecting all the legacy links to `/namespaces`, redirect to their new locations? E.g.
```
<Redirect exact from={`${pathPrefix}/deployments`} to={`${pathPrefix}/namespaces/all/deployments`} />
```

web/app/js/components/util/ApiHelpers.jsx
```diff
@@ -127,7 +127,7 @@ const ApiHelpers = (pathPrefix, defaultMetricsWindow = '1m') => {
     // Traffic Performance Summary. This retrieves stats for the given resource.
     let resourceUrl = '/api/tps-reports?resource_type=' + type;
 
-    if (_isEmpty(namespace)) {
+    if (_isEmpty(namespace) || namespace === "all") {
```
alpeb: For consistency, I think `urlForResourceNoStats` below also requires the addition of `|| namespace === "all"`

web/app/js/index.js
siggy: Building on that idea, there's a weird edge case here where a cluster could have a namespace called `all`. To handle this, perhaps the legacy links should become the true `all` pages:
`/deployments` => deployments in all namespaces
`/namespaces/all/deployments` => deployments in the `all` namespace

web/app/js/index.js
```diff
@@ -32,29 +31,62 @@ if (proxyPathMatch) {
...
+
               <Route
-                path={`${pathPrefix}/servicemesh`}
+                path={`${pathPrefix}/controlPlane`}
```
siggy: mixed case

web/app/js/index.js
scottcarol: These are great comments, thanks @siggy @alpeb for thinking about this.

My thought on URLs: Since we are now querying the API by specific namespace (with a voluntary `all` option), we want to formulate our URLs to teach the users that. I like having an `all` in the URL to explicitly communicate what the user is looking at. 

I checked the K8s dashboard and URLs are not as user-friendly but they do always explicitly state the namepace - here's the `Jobs` workload page URL when I have `all namespaces` selected: `/proxy/#!/job?namespace=_all`

I agree that it might make sense to redirect `/deployments/` to `/namespaces/all/deployments` instead of redirecting to the home view (`/namespaces`) as I'd initially thought, I can see how that could be confusing.

scottcarol:
@siggy good 👁 re: namespace dropdown not updating, I would like @grampelberg to weigh in on this. Originally I was emulating the K8s dashboard, which does not update if you have it open and then add a namespace. However we can definitely add a recurring call to check for new namespaces (which will not ping Prometheus).

Re: looking at an emojivoto pod detail page, that's interesting because I had the opposite expectation 😆 I thought it was startling to navigate away from a resource detail page. However as I think about it, I think your expectation makes more sense and I'm 👍 with navigating to the `/namespaces/linkerd` view upon namespace change... @grampelberg WDYT? 

👍 re: icons and `Namespaces` highlight.

web/app/js/index.js
scottcarol: @siggy I just re-read and see what you mean re: a namespace explicitly named `all`. Since you could name a namespace absolutely anything, I see what you mean. I like the URL pattern of `/namespaces/all/deployments` but if it couldn't handle every possible namespace, we should probably go back to `/deployments`.

web/app/js/components/Navigation.jsx
```diff
@@ -260,54 +316,140 @@ class NavigationBase extends React.Component {
   }
 
   render() {
-    const { classes, ChildComponent, ...otherProps } = this.props;
+    const { api, classes, selectedNamespace, ChildComponent, ...otherProps } = this.props;
+    const { namespaces, anchorEl } = this.state;
```
siggy: should this be `const { namespaces, anchorEl, drawerOpen } = this.state;` for consistency?

web/app/js/index.js
```diff
@@ -32,29 +31,62 @@ if (proxyPathMatch) {
...
+
+    this.state = {
+      ...appData
+    };
 
-const theme = createMuiTheme(dashboardTheme);
+    this.state.api = ApiHelpers(pathPrefix);
+    this.state.pathPrefix = pathPrefix;
+    this.state.productName = "Linkerd";
+    this.state.selectedNamespace = selectedNamespace;
```
siggy: curious: in other places in the code i see us doing `this.state = this.getInitialState(...` or `this.state = {...`. is this a relevant convention?

web/app/js/index.js
scottcarol: One more idea, we could have the default path be `/namespaces/all_namespaces/deployments` since you can't have an `_` in a K8s namespace name.

siggy: @scottcarol Yup! I see the k8s dashboard uses `_all`, that would work fine, or `all_namespaces`.

grampelberg: My preference is `_all` like the k8s dashboard.

grampelberg:
> Re: looking at an emojivoto pod detail page

I would like to copy how the k8s dashboard does this:

Screenshot 2019-09-24 13 42 21

Selecting `yes` will send you to the overview page for the new namespace.

web/app/js/index.js
scottcarol: Sounds good, this is my understanding of our conversation, thanks all:

* list of pods in all namespaces:       `/namespaces/_all/pods`
* list of pods in linkerd namespace: `/namespaces/linkerd/pods`
* `/pods` will redirect to                    `/namespaces/_all/pods`

web/app/js/index.js
scottcarol: hey @siggy, see my note in the original PR message about the lint error vs the React Context docs. i think there must be a better solution than defining these variables line by line, but I'm prioritizing the UI in this PR.

scottcarol:
@alpeb @siggy @adleong @grampelberg I believe I've incorporated all of the routing changes requested. 

I can do a lot to simplify `Navigation.jsx`, including pulling out the dialog box into its own component, but for your re-review, I'd primarily like to know whether the UI flow and routing is behaving as you'd expect. 

I will create a separate issue for UI polish and component simplification, and tackle that next!

grampelberg:
I've just got one request with this PR - could we default to `default` for the namespaces instead of `all namespaces`?

There's some other stuff that I'd like cleaned up (breadcrumbs come to mind), but lemme open up issues to get that cleaned up separately.

web/app/js/components/util/ApiHelpers.jsx
```diff
@@ -143,7 +143,7 @@ const ApiHelpers = (pathPrefix, defaultMetricsWindow = '1m') => {
     // Traffic Performance Summary. This retrieves (non-Prometheus) stats for the given resource.
     let resourceUrl = '/api/tps-reports?skip_stats=true&resource_type=' + type;
 
-    if (_isEmpty(namespace)) {
+    if (_isEmpty(namespace) || namespace === "_all") {
```
alpeb: Should this be really changed to `_all`? I see calls to `urlForResourceNoStats("all")` in `Tap.jsx`, `Top.jsx` and `TopRoutes.jsx`.
There are also calls to `urlForResource(..., "all")` above from `Namespace.jsx` and `ResourceDetail.jsx`

scottcarol:
@grampelberg thanks for the catch, will restore the default namespace to `default`.

web/app/js/components/util/ApiHelpers.jsx
scottcarol: `urlsForResourceNoStats()` and `urlsForResource()` both take two arguments, `type` and `namespace`.  

`Tap`, `Top` and `TopRoutes` pass in `"all"` for `type`. The sidebar redesign relates to the second `namespace` argument, where you can pass in `"_all"` to query the API for all namespaces. Agree that this is confusing, I'll rework this.

web/app/js/components/Navigation.jsx
```diff
@@ -260,62 +359,174 @@ class NavigationBase extends React.Component {
...
+            <Dialog
+              open={showNamespaceChangeDialog}
+              onClose={this.handleClose}
+              aria-labelledby="form-dialog-title">
+              <DialogTitle id="form-dialog-title">
+                Change namespace?
+              </DialogTitle>
+              <DialogContent>
+                <DialogContentText>
+              The resource you are viewing is in a different namespace than the namespace you have selected. Do you want to change the namespace from { selectedNamespace } to { newNamespace }?
```
siggy: personally i didn't find this confirmation necessary. i was expecting that by changing the namespace while on a resource detail page, i would be taken to `/namespace/foo`. also it this was not expected, a user should be able to hit the back button.

scottcarol: Thanks for the feedback. @grampelberg suggested the dialog in order to emulate the K8s dashboard. Once we get it in front of some users we can see what the preference is, I'm happy either way.

alpeb:
@scottcarol I thought this was related to what @siggy reported, but I still see it after the last push: Right after opening the dashboard (`http://localhost:50750/namespaces`), selecting a different namespace from the dropdown doesn't get reflected on the right pane. I can see the pane gets refreshed, but with the same Namespaces screen.

scottcarol:
@alpeb do you mean the view you see when you are at `http://localhost:50750/namespaces`? This view will always display information about all namespaces, no matter what the selected namespace is. Or are you seeing incorrect information when you are at `http://localhost:50750/namespaces/someNamespace`?

alpeb:
> @alpeb do you mean the view you see when you are at `http://localhost:50750/namespaces`? This view will always display information about all namespaces, no matter what the selected namespace is.

Yes, I would expect that selecting a namespace from the dropdown would take me to the namespace summary screen (like `http://localhost:50750/namespaces/someNamespace`), but it remains at `http://localhost:50750/namespaces`.

When at `http://localhost:50750/namespaces/someNamespace` I do see someNamespace's summary screen and selecting another namespace from the dropdown takes me to that new namespace summary screen, as expected.

scottcarol:
@alpeb @siggy thanks for the reviews. At present the `/namespaces` behavior is working as intended, which is to continue to show the list of all namespaces until the user selects a specific workload to view. (This is how the K8s dashboard does it - totally open to changing though, I see how it could be confusing.) My preference is to merge this in and get some user feedback on the existing flow.

PR #3470, opened 2019-09-24
The linkerd proxy does not work with headless services

### LINKED ISSUE #3308: Can't connect to headless service
### What is the issue?

Linkerd proxy returns `503 Service Unavailable` but should return answer from external service instead.

### How can it be reproduced?
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: repro
  annotations:
    linkerd.io/inject: enabled
---
apiVersion: v1
kind: Service
metadata:
  name: some-external-svc
  namespace: repro
spec:
  ports:
  - port: 9753
    protocol: TCP
    targetPort: 80
---
apiVersion: v1
kind: Endpoints
metadata:
  name: some-external-svc
  namespace: repro
subsets:
- addresses:
  - ip: 1.1.1.1
  ports:
  - port: 80
```

Attach a shell in an alpine repro pod with the following cmd: `kubectl run repro -n repro --rm -it --image=alpine --restart=Never --generator=run-pod/v1`

Install curl in pod with `apk add curl`

Request some-external-svc with curl: `curl -IXGET some-external-svc:9753`

### Logs, error output, etc

Response from linkerd proxy: `503 Service Unavailable`

Expect answer with status code <500 from external server.

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client                                                                                                                                                                  √ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version                                                                                                                                              √ is running the minimum kubectl version

linkerd-config
--------------
√ control plane Namespace exists                                                                                                                                                             √ control plane ClusterRoles exist

linkerd-existence
-----------------
√ 'linkerd-config' config map exists                                                                                                                                                         √ control plane replica sets are ready
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane pods are ready                                                                                                                                                               √ control plane self-check
√ [kubernetes] control plane can talk to Kubernetes                                                                                                                                          √ [prometheus] control plane can talk to Prometheus                                                                                                                                          √ no invalid service profiles

linkerd-version
---------------
√ can determine the latest version                                                                                                                                                           √ cli is up-to-date                                                                                                                                                                          
control-plane-version
---------------------
√ control plane is up-to-date                                                                                                                                                                √ control plane and cli versions match                                                                                                                                                       
Status check results are √
```

### Environment

- Kubernetes Version: 1.13.5
- Cluster Environment: AKS
- Host OS: Ubuntu
- Linkerd version: stable-2.5.0

### Possible solution

It is possible to get it to work if I pass a host header along with the target ip (eg. `Host:1.1.1.1`). However this would negate the need for an headless service. The point of which is to provide a common cluster internal name for the external service, so that the other cluster deployments does't need to know about the external ip.

### Additional context

This scenario works just fine without the linkerd proxy.

JohannesEH:
The linkerd proxy does not work with headless services (i.e. endpoints not referencing a pod).

Changed endpoints_watcher to also return endpoints with no targetref. Changed endpoint_translator to handle addresses with no associated pod.

Fixes #3308

Ran tests in minikube verifying that the proxy handles headless services correctly both in cases with and without port-remapping.

Signed-of-by: Johannes Hansen <johannesh1980@gmail.com>

### DIFF
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -47,7 +47,23 @@ func newEndpointTranslator(
 func (et *endpointTranslator) Add(set watcher.PodSet) {
 	addrs := []*pb.WeightedAddr{}
 	for _, address := range set {
-		wa, err := et.toWeightedAddr(address)
+		var (
+			wa  *pb.WeightedAddr
+			err error
+		)
+		if address.Pod != nil {
+			wa, err = et.toWeightedAddr(address)
+		} else {
+			// handling address with no associated pod, such as those
+			// belonging to headless services
+			var addr *net.TcpAddress
+			addr, err = et.toAddr(address)
+			wa = &pb.WeightedAddr{
+				Addr:   addr,
+				Weight: defaultWeight,
+			}
+		}
+
 		if err != nil {
 			et.log.Errorf("Failed to translate endpoints to weighted addr: %s", err)
 			continue
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -421,7 +421,7 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) PodSe
 				continue
 			}
 			if endpoint.TargetRef == nil {
-				id := PodID{
+				id := ServiceID{
 					Name: strings.Join([]string{
 						endpoints.ObjectMeta.Name,
 						endpoint.IP,
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -3,6 +3,7 @@ package watcher
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/linkerd/linkerd2/controller/k8s"
@@ -420,7 +421,18 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) PodSe
 				continue
 			}
 			if endpoint.TargetRef == nil {
-				pp.log.Warnf("Endpoint missing TargetRef: %+v", endpoint)
+				id := PodID{
+					Name: strings.Join([]string{
+						endpoints.ObjectMeta.Name,
+						endpoint.IP,
+						fmt.Sprint(resolvedPort),
+					}, "-"),
+					Namespace: endpoints.ObjectMeta.Namespace,
+				}
+				pods[id] = Address{
+					IP:   endpoint.IP,
+					Port: resolvedPort,
+				}
 				continue
 			}
 			if endpoint.TargetRef.Kind == "Pod" {
--- a/controller/api/destination/watcher/endpoints_watcher_test.go
+++ b/controller/api/destination/watcher/endpoints_watcher_test.go
@@ -138,6 +138,7 @@ status:
 				"172.17.0.12:8989",
 				"172.17.0.19:8989",
 				"172.17.0.20:8989",
+				"172.17.0.21:8989",
 			},
 			expectedNoEndpoints:              false,
 			expectedNoEndpointsServiceExists: false,


JohannesEH:
I have to admit. I'm pretty unsure if this change can have unforeseen consequences. Specifically regarding the test case I altered. Is it a normal thing that endpoints keep ip addresses around if a pod goes down? What will the linkerd proxy do in this case? 

Secondly it seems "dirty" to return a podset from the function if no pods exists. Maybe the code should use another abstraction.

Also if I need to write more tests please let me know.

I have had very limited time to setup the project, code and test. So I accept if this change needs to be scrapped after review due to my lack of understanding of the codebase.

olix0r:
Thanks @JohannesEH. We really appreciate you taking the time to submit a fix and for providing helpful context about the change!

JohannesEH:
I've tried getting a linkerd test environment up and running with minikube on my windows machine through wsl, but had no luck getting it to work. So I'm not sure if I can take this change any further as I would essentially be coding "in the dark".

ihcsim:
Looks like the panic is caused by an `Address` struct built without the `Pod` field, as part of the [new conditional](https://github.com/linkerd/linkerd2/pull/3470/files#diff-23f192ea6ae079b7f0816af58a49a0d7R432-R435) added in this PR. And the panic happens inside the `toWeightedAddr` function.

It seems to me that we shouldn't be calling the `toWeightedAddr` function for endpoints (addresses) without an associated pods. @adleong does this sound right?

@JohannesEH The following diff works for me, when tested with a headless service. Can you verify if it works for you?
```diff
diff --git a/controller/api/destination/endpoint_translator.go b/controller/api/destination/endpoint_translator.go
index 9ed2087e..5e337163 100644
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -47,7 +47,23 @@ func newEndpointTranslator(
 func (et *endpointTranslator) Add(set watcher.PodSet) {
 	addrs := []*pb.WeightedAddr{}
 	for _, address := range set {
-		wa, err := et.toWeightedAddr(address)
+		var (
+			wa  *pb.WeightedAddr
+			err error
+		)
+		if address.Pod != nil {
+			wa, err = et.toWeightedAddr(address)
+		} else {
+			// handling address with no associated pod, such as those
+			// belonging to headless services
+			var addr *net.TcpAddress
+			addr, err = et.toAddr(address)
+			wa = &pb.WeightedAddr{
+				Addr:   addr,
+				Weight: defaultWeight,
+			}
+		}
+
 		if err != nil {
 			et.log.Errorf("Failed to translate endpoints to weighted addr: %s", err)
 			continue
diff --git a/controller/api/destination/watcher/endpoints_watcher.go b/controller/api/destination/watcher/endpoints_watcher.go
index c3095fd1..c145fd12 100644
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -421,7 +421,7 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) PodSe
 				continue
 			}
 			if endpoint.TargetRef == nil {
-				id := PodID{
+				id := ServiceID{ // I think using ServiceID here makes it clearer that we aren't dealing with pods
 					Name: strings.Join([]string{
 						endpoints.ObjectMeta.Name,
 						endpoint.IP,
```

YAML I used for testing:
```
apiVersion: v1
kind: Service
metadata:
  name: some-external-svc
spec:
  clusterIP: None
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
---
apiVersion: v1
kind: Endpoints
metadata:
  name: some-external-svc
subsets:
- addresses:
  - ip: 52.86.78.202 # postman-echo.com
  ports:
  - port: 80
```

Using a `curl` container:
```
✗ k run curl --image=appropriate/curl --restart=Never --generator=run-pod/v1 --command -- sleep 3600 
✗ k exec -it curl -- /bin/sh 
/ # curl -4 -v "some-external-svc/get?foo1=bar1&foo2=bar2"
*   Trying 52.86.78.202...
* TCP_NODELAY set
* Connected to some-external-svc (52.86.78.202) port 80 (#0)
> GET /get?foo1=bar1&foo2=bar2 HTTP/1.1
> Host: some-external-svc
> User-Agent: curl/7.59.0
> Accept: */*
>
< HTTP/1.1 200 OK
< content-type: application/json; charset=utf-8
< date: Wed, 02 Oct 2019 20:55:58 GMT
< etag: W/"127-b7yb7plbgUmi0PtqWNltq5LT/yI"
< server: nginx
< set-cookie: sails.sid=s%3AWhDZkwCRgiSzOwXrPCyiN1zWep-ORrIM.osc4%2BVy%2Fe0XNXhmsSWGSCTnTwoJ1BtI58DVbUrU4nQg; Path=/; HttpOnly
< vary: Accept-Encoding
< content-length: 295
<
* Connection #0 to host some-external-svc left intact
{"args":{"foo1":"bar1","foo2":"bar2"},"headers":{"x-forwarded-proto":"https","host":"some-external-svc","accept":"*/*","l5d-dst-canonical":"some-external-svc.temp.svc.cluster.local:80","user-agent":"curl/7.59.0","x-forwarded-port":"80"},"url":"https://some-external-svc/get?foo1=bar1&foo2=bar2"}/

```

JohannesEH:
@ihcsim thanks!

I finally got the test environment working on my windows machine. Basically I had some problems because I fetched the source on windows leading to some bash scripts having CRLF line endings instead of LF (Maybe I should add a note contributing.md to clarify the gotchas on windows). Also, the hyper-v minikube setup script is running kubernetes version 1.8.0 and should be updated to 1.12+. Finally the mkube script is looking for minikube in "Program Files (x86)" instead of "Program Files" which was the minikube install folder on my machine. With all that out of the way I was able to build and run the proxy. 

I applied your diff and ran some tests. It works in the simple case with port targetPort being the same, but not with port remapping port != targetPort. I'm not sure how or if this should be forwarded to the linkerd proxy? Maybe it's handled by kube_proxy?

JohannesEH:
@ihcsim I also have some problems getting port remapping on headless services to work without linkerd so I'm not sure the issues lies with the proxy.

According to this https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-mapping-external-services (scenario 3) it should work.

It might be my minikube that have some problem.

ihcsim:
@JohannesEH Have you tried [Docker Desktop for Windows](https://docs.docker.com/docker-for-windows/kubernetes/)? I don't have much experience developing on Windows, but wonder if you will have better luck with Docker Desktop.

JohannesEH:
@ihcsim yup I'm using that. However all bash build scripts needs something like wsl to execute. But it wasn't that bad once I got I working, it just took some time. But honestly my plan is to redo my work machine and setup dual boot with windows and some Linux distro. I would much rather do k8s/docker/go development on Linux.

JohannesEH:
@adleong or @ihcsim Will you review/merge this after the 2.6 release or is there something more I can do to move this along?

ihcsim:
@JohannesEH Yeah, this PR will go into 2.7. btw, I haven't had a chance to look into the port != targetPort case you mentioned. Have you had a chance to investigate further.

JohannesEH:
@ihcsim Cool. :) Regarding the error I reported with port remapping I wasn't able to repro the problem in my following tests. So I wrote of the initial test result as a fluke.

Issue #3263, opened 2019-08-14
Launch proxy as part of CNI setup


grampelberg:
### What problem are you trying to solve?

Kubernetes has somewhat lacking sidecar support. When an application with a sidecar starts, and issues a request to an external source more quickly than the proxy starts, it is possible for there to be connection failures. It is difficult for an application to determine whether this is because the proxy has taken time to start or something more fundamental is wrong. In particular, it requires application developers to write custom code in preparation for being meshed. While there will be a native solution to this soon ([issue](https://github.com/kubernetes/enhancements/issues/753)), it is unclear when this will land and how long it will take to be adopted by all the providers of Kubernetes.

In addition, the sidecar pattern forces developers to at least be aware of the proxy running. It must be added to the pod's specification and get started. When moving into more secure enterprise environments that are dictating configuration, operations prefers to keep the details of the proxies to themselves. This includes keeping the versions and settings external from something that developers can observe.

### How should the problem be solved?

There is already a CNI plugin for Linkerd. This moves the iptables concerns from a developer to the operations team and is primarily used to remove NET_ADMIN permissions to make more secure environments. The plugin can run arbitrary processes and is setup/torndown as part of the pod's lifecycle. 

The CNI plugin should be improved to also start and manage the proxy's lifecycle alongside setting up iptables. This would result in there no longer being a sidecar in the pod, init containers sending connections through the proxy and having policy applied and removing the proxy's concerns from developers.

zaharidichev:
## CNI proxy life cycle

### The current state of things
We currently use a CNI plug-in which is essentially a daemon set that installs a CNI binary that gets called every time a pod is created. This plug-in is currently responsible for setting the IP tables rules that are needed to pass traffic through the proxy. This effectively removes the need for init containers. It does not however remove the concern of having a proxy running in the pod. So form the perspective of the developer, the proxy needs to be in the pod, needs to be injected through the inject command or needs to be added through the auto injector feature of linkerd. 

### The "ideal" scenario
In an ideal scenario, this cni plug-in will also handle the creation/shutdown of the proxy itself as part of the network setup, removing the need for any kind of other injection. This pretty much makes the existence of the proxy invisible to developers. It also removes the need for auto injection, init containers or manual injection as every pod that is added to the cluster will be configured to work with a proxy. 

### Benefits
- Remove concerns of adding proxy containers into pods
- Remove the need for init containers
- Remove the need for NET_ADMIN privileges in pods
[...]

PR #3471, opened 2019-09-24
WIP: Managing proxy lifecycle as part of CNI setup

zaharidichev:
This PR is not in ready state but I decided to nevertheless push for the sake of triggering the discussion around the cni work. 

**What does this actually do?** 
As you can read in #3263, there is the idea that the proxy itself can be started as part of the CNI setup. The advantages of this are discussed there in more detail. In essence this tries to bring the proxy into the "network layer" part of k8s, making it less of a concern for developers. I looked at some examples of other people attempting to do similar things as well as plenty of k8s kubelet code and decided that using CRI is appropriate here.

**How does it actually work?**
The way the CNI plugin used to work was to install a small binary on ever k8s node. This binary is ran whenever a pod is created, before the pod actually  gets scheduled. This was used to run all the necessary setup that needs to happen on the network level (setting up ip tables, etc), effectively removing the need for init containers. Now instead of that this plugin also takes care of starting the proxy and managing its life cycle. For the purpose of that, I have split the plugin into two parts: 

- the binary - this is an executable that gets ran whenever a pod is created
- the proxy scheduler - a daemon exposing a rest api, through which proxy instances can be started and stopped for particular pod sandbox. 

So what happens when the CNI plugin ADD command is called is that  apart from setting up the ip tables for the pod, it also fires a HTTP request to the proxy scheduler with all the necessary data to setup a proxy (sandbox id, pod ip, etc). 

When the scheduler receives this request, it queries the k8s API to get all the linkerd config bits that are needed for setting up the proxy. This includes things like trust anchors, port configuration, resources (if any), namespaces, etc. The scheduler then starts the proxy container in the correct namespace with the correct configuration. It is supposed to wait for the proxy to be available, through querying its readiness endpoints and only then allow the rest of the pod setup process to complete. 

Note that the pod does not actually gets scheduled until the CNI cmd execution is done. From then on the scheduler is supposed to check periodically whether the proxy container is running and restart it if needed. The reason for all that is the fact that this proxy container is not not really visible to the kubelet itself, so we cannot rely on k8s to actually take care of restart. When the pod goes away, it calls the cni plugin del command and that takes care of terminating the proxy container for this pod.

**How to try it**  
- Get this branch
- Build the docker images
- Install kind 0.2 (more on that later)
- Upload the build images to your k8s cluster 
- Run `linkerd install-cni --scheduler-bind-port=8999 --cni-log-level=debug  | kubectl apply -f -`
- Run `linkerd install --linkerd-cni-enabled | kubectl apply -f -`
- Wait for all linkerd components to come up
- Run `curl -sL https://run.linkerd.io/emojivoto.yml  | kubectl apply -f -`
- If you want you can take a look at the proxy-scheduler logs to see whats happening

After the containers are started, you can go to any of them and hit the `/metrics` endpoints of the proxy to see that traffic is actually flowing through it. You can also try out the emoji app to convince yourself that it is working. Dashbords, Prometheus metrics, tap, top, stat, etc **are not working at that point**

**Things that are broken**
First of all, this build will most likely be broken, as I have not taken the care to fix the tests exercising the generation of yaml files for the install-cni command (as this will be changed quite a lot anyway). Secondly, because now the proxy is invisible to k8s, a lot of the functionality that relies on it actually being present in the pod spec would not work. I think this is why Prometheus metrics acquisition does not work as well as tap and all the rest. Effectively from the viewpoint of the control plane these pods are not `meshed`.  This can be fixed however. 

**Things that are not implemented yet**
- [ ] A good set of unit + integration tests
- [ ] Monitoring the status of running proxies and restarting when needed
- [ ] Terminating proxy containers
- [ ] Pulling proxy images if not present
- [ ] Improved logging for the proxy scheduler
- [ ] Waiting for the proxy container to be available and the proxy fully initialized before the rest of the pod is initialized
- [ ] Migration of cni template to help chart
- [ ] Support for different container runtimes (currently it works only for docker, thereby restricting me to kind 0.2, as they transitioned to containerd in their later versions)
- [ ] Ensure this all works on a number of different environments
- [ ] Probably a bunch of other things that I did not think about

PS: This is my first piece of Go code, my first exposure to the internals of k8s let alone CRI itself and my first pr to the control plane, so approach every decision I made with a healthy dose of doubt...


grampelberg:
@zaharidichev mind closing this one out?

zaharidichev:
Potentially revisit in the future

PR #3476, opened 2019-09-25
make public-api use the right destination address

Pothulapati:

PR https://github.com/linkerd/linkerd2/pull/3407 moved `destination` into a separate component. The `public-api` component was not using the latest address thus breaking `linkerd endpoints ` which interacts with the `public-api`.

Passing the address explicitely, fixes the issue!

@alpeb

### DIFF
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -61,6 +61,7 @@ spec:
       - args:
         - public-api
         - -prometheus-url=http://linkerd-prometheus.{{.Namespace}}.svc.{{.ClusterDomain}}:9090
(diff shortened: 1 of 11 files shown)


grampelberg:
Is there a test we can add to make sure this doesn't happen again in the future?

alpeb:
@grampelberg we need at least one minimal integration test for `linkerd endpoints`.
`linkerd edges` is also currently lacking an integration test.

grampelberg:
@alpeb mind creating issues for both (if they don't exist) and marking as p0 in 2.7?

alpeb:
Done: #3477 and #3478

Issue #3472, opened 2019-09-24
Enable tracing on a per-pod basis


grampelberg:
### What problem are you trying to solve?

The proxy is now able to emit spans. This needs to be configured on a per-proxy basis.

### How should the problem be solved?

Add two annotations:

- `config.linkerd.io/trace-collector: oc-collector.tracing:55678` - translates to `LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR ` for the proxy.
- `config.experimental.linkerd.io/trace-collector-service-account: default` - should be `default` by default and optional. Convert to the linkerd identity string (`default.tracing.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)`) and populate `LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME`

### Out of scope

- Wiring this up to `inject`.
- Wiring this up to `install`.
- Discovery of the service account being used by the collector service's pods.

Issue #3468, opened 2019-09-24
Allow trace collector address to be configured via flag or helm variable


adleong:
https://github.com/linkerd/linkerd2-proxy/pull/354 added environment variables: `LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR` and `LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME` which enable the Linkerd proxy to participate in traces and to emit spans to the provided collector.  We should allow these environment variables to be set from the CLI in the normal ways.

grampelberg:
@adleong this seems like a blocker for 2.6?

adleong:
yes, I think so

adleong:
So far I've been testing by using manual injection and editing the output to add:

```
       - name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME
          value: default.tracing.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
        - name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR
          value: oc-collector.tracing.svc.cluster.local:55678
```

ihcsim:
With edge-19.9.3, Helm users can enable tracing by running:
```
$ helm install [--name=linkerd2] --set Proxy.Trace.CollectorSvcAddr="oc-collector.tracing.svc.cluster.local:55678" --set Proxy.Trace.CollectorSvcAccount="default.tracing" --set Identity.Issuer.CrtExpiry="<cert-expiry-date>" --set-file Identity.TrustAnchorsPEM=<path-to-trust-pem-file> --set-file Identity.Issuer.TLS.CrtPEM=<path-to-cert-pem-file> --set-file Identity.Issuer.TLS.KeyPEM=<path-to-key-pem-file> linkerd2 linkerd-edge/linkerd2
```
Notice that I have to specify the namespace of the service account, as there isn't a way to derive it from the service name. I think this is a detail we can add to the Linkerd Helm documentation.

Pothulapati:
I'm taking this up, 

The plan is to have, `linkerd inject --proxy-collector-svc-address --proxy-collector-svc-account` flags add relevant annotations to the workloads.

grampelberg:
@Pothulapati let's try and match the annotations:

- `--trace-collector`
- `--trace-collector-service-account`

@adleong @ihcsim any good ideas on how to move the `config.alpha.linkerd.io` space forward to the CLI arguments?

PR #3481, opened 2019-09-26
Introduce Tracing Annotations

ihcsim:
This PR introduces two proxy config annotations named `config.linkerd.io/trace-collector` and `config.alpha.linkerd.io/trace-collector-service-account`, to support per-pod tracing. These annotations are translated into two new proxy environment variables.

Note that these annotations work on both the pod template and namespace levels.

The proxy injector logs and events have been updated to include `tracing enabled` events.
```
linkerd-proxy-injector-667958849d-pkvkn proxy-injector time="2019-09-26T04:32:23Z" level=info msg="patch generated for: pod/voting-855d545684-"
linkerd-proxy-injector-667958849d-pkvkn proxy-injector time="2019-09-26T04:32:23Z" level=info msg="received admission review request 93017758-75b4-4aac-8721-c9c37d11b043"
linkerd-proxy-injector-667958849d-pkvkn proxy-injector time="2019-09-26T04:32:23Z" level=info msg="received pod/web-79694f4896-"
linkerd-proxy-injector-667958849d-pkvkn proxy-injector time="2019-09-26T04:32:23Z" level=info msg="tracing enabled for deployment/web: remote service=oc-collector.tracing:55678, service account=tracing.tracing"


Events:
  Type    Reason             Age   From                    Message
  ----    ------             ----  ----                    -------
  Normal  ScalingReplicaSet  83s   deployment-controller   Scaled up replica set emoji-697b575bd9 to 1
  Normal  Injected           83s   linkerd-proxy-injector  Linkerd sidecar proxy injected
  Normal  Tracing            83s   linkerd-proxy-injector  Tracing Enabled
```

Fixes https://github.com/linkerd/linkerd2/issues/3472.

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -68,6 +68,14 @@ env:
 - name: LINKERD2_PROXY_TAP_SVC_NAME
   value: linkerd-tap.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
 {{ end -}}
+{{ if .Proxy.Trace -}}
+{{ if .Proxy.Trace.CollectorSvcAddr -}}
+- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR
+  value: {{ .Proxy.Trace.CollectorSvcAddr }}
+- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME
+  value: {{ .Proxy.Trace.CollectorSvcAccount }}.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
+{{ end -}}
+{{ end -}}
 image: {{.Proxy.Image.Name}}:{{.Proxy.Image.Version}}
 imagePullPolicy: {{.Proxy.Image.PullPolicy}}
 livenessProbe:
--- a/controller/proxy-injector/webhook.go
+++ b/controller/proxy-injector/webhook.go
@@ -20,6 +20,7 @@ import (
 const (
 	eventTypeSkipped  = "InjectionSkipped"
 	eventTypeInjected = "Injected"
+	eventTypeTracing  = "Tracing"
 )
 
 // Inject returns an AdmissionResponse containing the patch, if any, to apply
@@ -108,6 +109,9 @@ func Inject(api *k8s.API,
 
 	if parent != nil {
 		recorder.Event(*parent, v1.EventTypeNormal, eventTypeInjected, "Linkerd sidecar proxy injected")
+		if report.TracingEnabled {
+			recorder.Event(*parent, v1.EventTypeNormal, eventTypeTracing, "Tracing Enabled")
+		}
 	}
 	log.Infof("patch generated for: %s", report.ResName())
 	log.Debugf("patch: %s", patchJSON)
--- a/pkg/charts/values_test.go
+++ b/pkg/charts/values_test.go
@@ -85,6 +85,10 @@ func TestNewValues(t *testing.T) {
 					Request: "",
 				},
 			},
+			Trace: &Trace{
+				CollectorSvcAddr:    "",
+				CollectorSvcAccount: "default",
+			},
 			UID: 2102,
 		},
 
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -2,6 +2,7 @@ package inject
 
 import (
 	"encoding/json"
+	"fmt"
 	"regexp"
 	"sort"
 	"strconv"
@@ -55,6 +56,7 @@ var (
 		k8s.ProxyVersionOverrideAnnotation,
 		k8s.ProxyIgnoreInboundPortsAnnotation,
 		k8s.ProxyIgnoreOutboundPortsAnnotation,
+		k8s.ProxyTraceCollectorSvcAddr,
 	}
 )
 
@@ -497,6 +499,11 @@ func (conf *ResourceConfig) injectPodSpec(values *patch) {
 	}
 
 	values.AddRootVolumes = len(conf.pod.spec.Volumes) == 0
+
+	if trace := conf.trace(); trace != nil {
+		log.Infof("tracing enabled for %s/%s: service name=%s, service account=%s", conf.workload.metaType.Kind, conf.workload.Meta.Name, trace.CollectorSvcAddr, trace.CollectorSvcAccount)
+		values.Proxy.Trace = trace
+	}
 }
 
 func (conf *ResourceConfig) injectProxyInit(values *patch) {
@@ -537,6 +544,32 @@ func (conf *ResourceConfig) serviceAccountVolumeMount() *corev1.VolumeMount {
 		}
 	}
 	return nil
+}
+
+func (conf *ResourceConfig) trace() *charts.Trace {
+	var (
+		svcAddr    = conf.getOverride(k8s.ProxyTraceCollectorSvcAddr)
+		svcAccount = conf.getOverride(k8s.ProxyTraceCollectorSvcAccount)
+	)
+
+	if svcAddr == "" {
+		return nil
+	}
+
+	splits := strings.Split(svcAddr, ".")
+	if len(splits) <= 1 {
+		log.Errorf("Fail to start tracing due to malformed collector service address %s. Expected service address to use the <svc-name>:<ns> format", svcAddr)
+		return nil
+	}
+
+	if delimiter := strings.Index(splits[1], ":"); delimiter != -1 {
+		splits[1] = splits[1][:delimiter]
+	}
+
+	return &charts.Trace{
+		CollectorSvcAddr:    svcAddr,
+		CollectorSvcAccount: fmt.Sprintf("%s.%s", svcAccount, splits[1]),
+	}
 }
 
 // Given a ObjectMeta, update ObjectMeta in place with the new labels and
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -556,19 +556,19 @@ func (conf *ResourceConfig) trace() *charts.Trace {
 		return nil
 	}
 
-	splits := strings.Split(svcAddr, ".")
-	if len(splits) <= 1 {
-		log.Errorf("fail to enable tracing due to malformed collector service address %s. Expected service address to use the <svc-name>:<ns> format", svcAddr)
-		return nil
-	}
+	hostAndPort := strings.Split(svcAddr, ":")
+	hostname := strings.Split(hostAndPort[0], ".")
 
-	if delimiter := strings.Index(splits[1], ":"); delimiter != -1 {
-		splits[1] = splits[1][:delimiter]
+	var ns string
+	if len(hostname) == 1 {
+		ns = conf.workload.Meta.Namespace
+	} else {
+		ns = hostname[1]
 	}
 
 	return &charts.Trace{
 		CollectorSvcAddr:    svcAddr,
-		CollectorSvcAccount: fmt.Sprintf("%s.%s", svcAccount, splits[1]),
+		CollectorSvcAccount: fmt.Sprintf("%s.%s", svcAccount, ns),
 	}
 }
 
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -28,6 +28,8 @@ const (
 	proxyInitResourceRequestMemory = "10Mi"
 	proxyInitResourceLimitCPU      = "100m"
 	proxyInitResourceLimitMemory   = "50Mi"
+
+	traceDefaultSvcAccount = "default"
 )
 
 var (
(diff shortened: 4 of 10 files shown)


pkg/inject/inject.go
```diff
@@ -539,6 +546,32 @@ func (conf *ResourceConfig) serviceAccountVolumeMount() *corev1.VolumeMount {
...
+		svcAccount = conf.getOverride(k8s.ProxyTraceCollectorSvcAccount)
+	)
+
+	if svcAddr == "" {
+		return nil
+	}
+
+	splits := strings.Split(svcAddr, ".")
+	if len(splits) <= 1 {
+		log.Errorf("fail to enable tracing due to malformed collector service address %s. Expected service address to use the <svc-name>:<ns> format", svcAddr)
```
alpeb: Expected should be `<svc-name>.<ns>`, right?

controller/proxy-injector/webhook.go
```diff
@@ -108,6 +109,9 @@ func Inject(api *k8s.API,
 
 	if parent != nil {
 		recorder.Event(*parent, v1.EventTypeNormal, eventTypeInjected, "Linkerd sidecar proxy injected")
+		if report.TracingEnabled {
```
grampelberg: Should we have a disabled event as well? I'm not 100% sure it is needed.

pkg/inject/inject.go
alpeb: Also isn't it possible for `config.linkerd.io/trace-collector`'s value to have more than one dot?

pkg/charts/values_test.go
```diff
@@ -85,6 +85,10 @@ func TestNewValues(t *testing.T) {
 					Request: "",
 				},
 			},
+			Trace: &Trace{
+				CollectorSvcAddr:    "",
+				CollectorSvcAccount: "default",
```
grampelberg: It looks like you're expecting the namespace further down, so this won't work?

pkg/inject/inject.go
```diff
@@ -539,6 +546,32 @@ func (conf *ResourceConfig) serviceAccountVolumeMount() *corev1.VolumeMount {
 	return nil
 }
 
+func (conf *ResourceConfig) trace() *charts.Trace {
+	var (
+		svcAddr    = conf.getOverride(k8s.ProxyTraceCollectorSvcAddr)
+		svcAccount = conf.getOverride(k8s.ProxyTraceCollectorSvcAccount)
```
alpeb: Shouldn't this be defaulted to `default` if the annotation is missing?

ihcsim: The default is read from `values.yaml`, like all other defaults.

controller/proxy-injector/webhook.go
ihcsim: Just like inject/skip, I think it will be easier for users to determine if tracing is enabled by looking at events than the proxy injector logs. Happy to remove if you don't think it's needed.

pkg/inject/inject.go
alpeb: I think you're missing a step. `conf.getOverride()` only checks for annotations. It seems to me you need to add the `Trace` struct into the protobuf def, and then if `conf.getOverride()` doesn't return anything, default to something like `conf.configs.GetProxy().GetTrace().GetCollectorSvcAddr()` and `conf.configs.GetProxy().GetTrace().GetCollectorSvcAccount()`

pkg/charts/values_test.go
ihcsim: This test is to validate the defaults read from the `values.yaml` file. The logic that appends the namespace is in the `pkg/inject` package. The test in `pkg/inject/inject_test.go` will require the namespace suffix to be added as seen [here](https://github.com/linkerd/linkerd2/pull/3481/files#diff-61c5e7cb5fc7a736c4ba737d5ad5f4ebR142).

controller/proxy-injector/webhook.go
ihcsim: Sorry, completely misread your comment. Yeah, I agree we don't need to add `disable` events, as I like to reserve events for unusual stuff.

pkg/inject/inject.go
ihcsim: The reason why I didn't update the protobuf is because everything in the pb def (i.e. all the configuration in the `linkerd-config` config map) is meant to apply to *all* proxies, and be persisted across upgrades. Since #3472 is limited to (on demand) enabling tracing per workload, I think we don't need to add these to the pb def.

In this case, if `conf.getOverride()` returns an empty string, implying that no tracing annotations are added, we assume tracing is disabled.

alpeb: Ok, in that case what are the `Trace` properties in `values.yaml` for, since we only care for the annotation values?

pkg/inject/inject.go
ihcsim: It is possible, but this check only logs an error if the provided service address only has the service name, with no namespace or namspace + cluster domain suffix.

pkg/inject/inject.go
```diff
@@ -539,6 +546,32 @@ func (conf *ResourceConfig) serviceAccountVolumeMount() *corev1.VolumeMount {
...
+		return nil
+	}
+
+	splits := strings.Split(svcAddr, ".")
+	if len(splits) <= 1 {
+		log.Errorf("fail to enable tracing due to malformed collector service address %s. Expected service address to use the <svc-name>:<ns> format", svcAddr)
+		return nil
+	}
+
+	if delimiter := strings.Index(splits[1], ":"); delimiter != -1 {
```
adleong: What is this doing?  Stripping the port off?  Is there a more clear way to do this, perhaps before splitting the domain into segments?

pkg/inject/inject.go
adleong: Do we want to support a single segment service name if the service is in the same namespace as the pod?

charts/partials/templates/_proxy.tpl
```diff
@@ -68,6 +68,14 @@ env:
 - name: LINKERD2_PROXY_TAP_SVC_NAME
   value: linkerd-tap.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
 {{ end -}}
+{{ if .Proxy.Trace -}}
+{{ if .Proxy.Trace.CollectorSvcAddr -}}
+- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR
+  value: {{ .Proxy.Trace.CollectorSvcAddr }}
+- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME
+  value: {{ .Proxy.Trace.CollectorSvcAccount }}.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
```
adleong: I found it a bit hard to follow that this variable contains both the service account name and namespace.  I think it would be more clear if there was a template variable for the entire identity string and the value for the whole identity string was composed in the injector code *OR* if there were two separate template variables for service account name and service account namespace.

pkg/inject/inject.go
ihcsim: Ha.. you are right. It was a leftover from pre- a4b17184add0b3f6b9579a459e32e9ad3aea1bdc.

ihcsim: Ha.. you are right. It was a leftover from pre- a4b17184add0b3f6b9579a459e32e9ad3aea1bdc.

pkg/inject/report.go
```diff
@@ -77,6 +78,7 @@ func newReport(conf *ResourceConfig) *Report {
 		report.HostNetwork = conf.pod.spec.HostNetwork
 		report.Sidecar = healthcheck.HasExistingSidecars(conf.pod.spec)
 		report.UDP = checkUDPPorts(conf.pod.spec)
+		report.TracingEnabled = conf.pod.meta.Annotations[k8s.ProxyTraceCollectorSvcAddr] != "" || conf.nsAnnotations[k8s.ProxyTraceCollectorSvcAddr] != ""
```
adleong: Feel free to ignore for now, but just a forward looking question:

Are reports always based just on the annotations?  If we, in the future, support setting default tracing values in the linkerd-config, is it possible to report that tracing is enabled even when it's not on the annotations?

pkg/inject/inject.go
alpeb: If the there's more than one dot in that address then the splitting over the next lines might not work as expected. I'll let you discuss with @adleong over his suggestions ;-)

pkg/inject/report.go
ihcsim: Yes, if down the road, we have other ways to enable tracing, we will need to account for that here.

pkg/inject/inject.go
ihcsim: Per slack chat, we will keep these as they are needed for https://github.com/linkerd/linkerd2/issues/3468.

pkg/inject/inject.go
ihcsim: Good point. I forgot that the `.ns` suffix is optional if the `Service` is in the same namespace.

ihcsim: Good catch on the subsequent call to `strings.Split()`.

ihcsim: Since the namespace that will be appended to the service account below is derived from the service address, a single segment service name isn't going to work. We will need to add this to the docs.

adleong: if the service namespace isn't provided, isn't it assumed to be the same as the pod?  i.e. if a pod in the `foo` namespace specifies the `bar` service, we would assume that means the `bar` service in the `foo` namespace and also use `foo` as the namespace for the service account?

charts/partials/templates/_proxy.tpl
ihcsim: For some reasons, I thought that `<serviceaccount>.<ns>` is the fully qualified name of a service account, but turns out it's something like `<group>:<ns>:<serviceaccount>`. I'm gonna keep the `.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)` as-is to remain consistent with the rest of the `*_SVC_NAME` in this template, and might just change the `CollectorSvcAccount` variable name to something more reflective. Any suggestions on variable name?

pkg/inject/inject.go
```diff
@@ -497,6 +501,11 @@ func (conf *ResourceConfig) injectPodSpec(values *patch) {
 	}
 
 	values.AddRootVolumes = len(conf.pod.spec.Volumes) == 0
+
+	if trace := conf.trace(); trace != nil {
+		log.Infof("tracing enabled for %s/%s: remote service=%s, service account=%s", conf.workload.ownerRef.Kind, conf.workload.ownerRef.Name, trace.CollectorSvcAddr, trace.CollectorSvcAccount)
```
alpeb: When running `linkerd inject --manual` against some uninjected payload with the trace annotation, this is throwing a segfault because `conf.workload.ownerRef` is nil.

pkg/charts/values_test.go
adleong: I'm trying to understand this.  Does the injector use values.yaml as the defaults for the Values struct?  If so, I think this is a problem.  For any pods that don't have the annotation, these default values would be used and plugged into the proxy partial, leading to a `LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME` of `default.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)` which is missing the namespace.

adleong: Wouldn't it be more appropriate for the Trace struct to be absent from Values.yaml?  So that the default value of `Trace` is nil?

pkg/inject/inject.go
```diff
@@ -539,6 +548,36 @@ func (conf *ResourceConfig) serviceAccountVolumeMount() *corev1.VolumeMount {
 	return nil
 }
 
+func (conf *ResourceConfig) trace() *charts.Trace {
+	var (
+		svcAddr    = conf.getOverride(k8s.ProxyTraceCollectorSvcAddr)
```
adleong: just curious why these are var definitions instead of `:=` assignments

pkg/inject/inject.go
ihcsim: Good catch! Thanks. We got bitten by this before; we can't tell which variables in the config are non-nil, depending on which CLI commands instantiate it.

pkg/inject/inject.go
ihcsim: Just a habit of grouping variables together.

pkg/charts/values_test.go
ihcsim: The injector doesn't use `values.yaml` during inject time. The `values.yaml` file is only relevant for `install`, when we work on https://github.com/linkerd/linkerd2/issues/3468. The reason for not defaulting `Trace` to nil here is because I'm hoping that this file is the only place where we will specify the default constant. 

> leading to a LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME of default.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain) which is missing the namespace.

The tracing environment variables won't be added if the `CollectorSvcAddr` is empty.

ihcsim: This change to the `values.yaml` is a leftover from pre- a4b17184add0b3f6b9579a459e32e9ad3aea1bdc. Since I assume we will tackle #3468 very soon, I figure it's a small enough change to be left alone.

adleong: I can confirm that when the annotation is absent, the environment variables are also absent.

PR #3486, opened 2019-09-26
Revert upgrade to buster based on CNI test failure after merge

cpretzer:
Signed off by "Charles Pretzer" <charles@buoyant.io>

Tested locally and there were no failures.


codeman9:
Ah yes. We ran into something like this with our aws proxy. This page describes what happened: https://wiki.debian.org/nftables we ended up reverting buster to using iptables-legacy.

PR #3488, opened 2019-09-27
Check extension apiserver authentication

### LINKED ISSUE #3343: Check for extension server certificate

It appears that not all clusters have `requestheader-client-ca-file`. With tap and soon other APIServices, that is required. Check should notify the user when that is missing as part of `check --pre`.

bmcstdio:
This PR closes #3343 by implementing a check for `--requestheader-client-ca-file` than runs as part of `linkerd check --pre`. I am also taking on this opportunity to reduce code duplication a bit. 🙂

### DIFF
--- a/pkg/k8s/constants.go
+++ b/pkg/k8s/constants.go
@@ -1,0 +1,8 @@
+package k8s
+
+const (
+	// ExtensionAPIServerAuthenticationConfigMapName is the name of the ConfigMap where authentication data for extension API servers is placed.
+	ExtensionAPIServerAuthenticationConfigMapName = "extension-apiserver-authentication"
+	// ExtensionAPIServerAuthenticationRequestHeaderClientCAFileKey is the key that contains the value of the "--requestheader-client-ca-file" flag.
+	ExtensionAPIServerAuthenticationRequestHeaderClientCAFileKey = "requestheader-client-ca-file"
+)
--- a/pkg/k8s/util.go
+++ b/pkg/k8s/util.go
@@ -1,0 +1,24 @@
+package k8s
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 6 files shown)


wmorgan:
Thanks for submitting this @bmcstdio! We're in the middle of scrambling for 2.6 so it may take us a few days to take a look at this. Thanks in advance for your patience.

bmcstdio:
@wmorgan sure thing, no worries! :)

pkg/k8s/constants.go
```diff
@@ -0,0 +1,8 @@
+package k8s
+
+const (
+	// ExtensionAPIServerAuthenticationConfigMapName is the name of the ConfigMap where authentication data for extension API servers is placed.
+	ExtensionAPIServerAuthenticationConfigMapName = "extension-apiserver-authentication"
```
adleong: Many constants live in `pkg/k8s/labels.go`, even constants which are not labels.  I'd recommend putting these in that file.

pkg/k8s/util.go
```diff
@@ -0,0 +1,24 @@
+package k8s
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ExtensionAPIServerConfigMapResource returns the string representation of a ConfigMap resource, meant to be used in unit tests.
+func ExtensionAPIServerConfigMapResource(data map[string]string) string {
```
adleong: I can see how this cleans up a bunch of duplication in the tests but I think I'd still prefer putting the manifest directly inline instead of a `map[string]string`.  The tests are going to be verbose anyway and I think it's helpful to see the manifest files.

grampelberg:
@bmcstdio since we've not heard from you in awhile, I'm going to close this out. Please reopen if you're still interested in polishing it off! Looks like you're almost there!

PR #3495, opened 2019-09-29
Trace Control plane Components with OC

Pothulapati:
This PR uses OpenCensus Libraries especially `ocgrpc` & `ochttp` to trace the major grpc and http calls. 

We introduce two new flags with the defaults:

`trace-collector`:          This specified the URL at wihch these spans are sent.
`sampling-probability`: This specifies the rate at which sampling has to be done. 0.1 imples to collect 1 in every 10 (0 implies none, and 1.0 includes all spans)

**How does this work:**

- We use a new function `flags.AddTraceFlags(cmd)` which adds the above mentioned flags and returns their values to be used.
- We also use `util.InitialiseTracing("linkerd-destination", *traceCollector, *probabilisticSamplingRate)`, which registers exporters and also initialises the trace config as per the flags.
Both of these functions are used in the `main` functions of the components.
- `prometheus.NewGrpcServer()` called by components to create a grpc server with telemetry, also now adds grpc request traces using `ocgrpc` handler.
- `prometheus.WithTelemetry()` which is used to instrument HTTP server components, now also adds `ochttp` trace handler.
- All the components, that have `client.go` which are used by other components to call these components are instrumented to use `ochttp` transport to propogate the trace headers.

**Components Traced:**

- All the server side components i.e `linkerd-web`, `linkerd-destination`, `linkerd-tap`, `linkerd-public-api`, `linkerd-identity`
- The prometheus queries in the `linkerd-public-api` component are also traced.

**Questions**

- As grpc calls especically that of `Identity`, `Destination` are not adding any contextual information.  Are the initial request spans useful? If not, we can remove them. Contextual information is added by using interceptors which would be another PR.
- Is the Information in the prometheus query spans enough? If not, what can be added?

**What doesn't work:**
- The spans from `linkerd-web -> linkerd-tap` are seen as induvidual spans i.e spearate traces because the http client from the `linkerd-web` component uses the k8s transport for communication, so we cant attach the ochttp transport.
 

To be reviewed by, @adleong @grampelberg @ihcsim @alpeb

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -33,6 +33,15 @@ func ConfigureAndParse(cmd *flag.FlagSet, args []string) {
 	maybePrintVersionAndExit(*printVersion)
 }
 
+// AddTraceFlags adds the traceCollector and ProbabilisticSamplingRate flags
+// to the flagSet and returns their pointers for usage
+func AddTraceFlags(cmd *flag.FlagSet) (*string, *float64) {
+	traceCollector := cmd.String("trace-collector", "", "Enables OC Tracing with the specified endpoint as collector")
+	probabilitySampling := cmd.Float64("sampling-probability", 0, "The probabilistic sampling rate for the traces")
+
+	return traceCollector, probabilitySampling
+}
+
 func setLogLevel(logLevel string) {
 	level, err := log.ParseLevel(logLevel)
 	if err != nil {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -33,13 +33,12 @@ func ConfigureAndParse(cmd *flag.FlagSet, args []string) {
 	maybePrintVersionAndExit(*printVersion)
 }
 
-// AddTraceFlags adds the trace-collector and sampling-probability flags
+// AddTraceFlags adds the trace-collector flag
 // to the flagSet and returns their pointers for usage
-func AddTraceFlags(cmd *flag.FlagSet) (*string, *float64) {
+func AddTraceFlags(cmd *flag.FlagSet) *string {
 	traceCollector := cmd.String("trace-collector", "", "Enables OC Tracing with the specified endpoint as collector")
-	probabilitySampling := cmd.Float64("sampling-probability", 0, "The probabilistic sampling rate for the traces")
 
-	return traceCollector, probabilitySampling
+	return traceCollector
 }
 
 func setLogLevel(logLevel string) {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -33,7 +33,7 @@ func ConfigureAndParse(cmd *flag.FlagSet, args []string) {
 	maybePrintVersionAndExit(*printVersion)
 }
 
-// AddTraceFlags adds the traceCollector and ProbabilisticSamplingRate flags
+// AddTraceFlags adds the trace-collector and sampling-probability flags
 // to the flagSet and returns their pointers for usage
 func AddTraceFlags(cmd *flag.FlagSet) (*string, *float64) {
 	traceCollector := cmd.String("trace-collector", "", "Enables OC Tracing with the specified endpoint as collector")
--- a/pkg/prometheus/prometheus.go
+++ b/pkg/prometheus/prometheus.go
@@ -89,7 +89,7 @@ func init() {
 	)
 }
 
-// NewGrpcServer returns a grpc server pre-configured with prometheus interceptors, and OcGrpc Stats handler
+// NewGrpcServer returns a grpc server pre-configured with prometheus interceptors and OcGrpc Stats handler
 func NewGrpcServer() *grpc.Server {
 	server := grpc.NewServer(
 		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
@@ -102,7 +102,7 @@ func NewGrpcServer() *grpc.Server {
 	return server
 }
 
-// WithTelemetry instruments the HTTP server with prometheus, and Trace handlers
+// WithTelemetry instruments the HTTP server with prometheus and Trace handlers
 func WithTelemetry(handler http.Handler) http.Handler {
 	return &ochttp.Handler{
 		Handler: promhttp.InstrumentHandlerDuration(serverLatency,
--- a/pkg/prometheus/prometheus.go
+++ b/pkg/prometheus/prometheus.go
@@ -3,11 +3,11 @@ package prometheus
 import (
 	"net/http"
 
-	"go.opencensus.io/plugin/ocgrpc"
-
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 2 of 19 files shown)


pkg/flags/flags.go
```diff
@@ -33,6 +33,15 @@ func ConfigureAndParse(cmd *flag.FlagSet, args []string) {
 	maybePrintVersionAndExit(*printVersion)
 }
 
+// AddTraceFlags adds the traceCollector and ProbabilisticSamplingRate flags
```
kleimkuhler: Nit: It looks like the flag names are `trace-collector` and `sampling-probability` as you mentioned in your PR description. If this comment is talking about the flags, it could be changed to use the flag names rather than the current names!

pkg/prometheus/prometheus.go
```diff
@@ -87,23 +89,26 @@ func init() {
 	)
 }
 
-// NewGrpcServer returns a grpc server pre-configured with prometheus interceptors
+// NewGrpcServer returns a grpc server pre-configured with prometheus interceptors, and OcGrpc Stats handler
```
kleimkuhler: Nit: No need for the comma here 😝

pkg/prometheus/prometheus.go
```diff
@@ -87,23 +89,26 @@ func init() {
...
 	grpc_prometheus.Register(server)
 	return server
 }
 
-// WithTelemetry instruments the HTTP server with prometheus
-func WithTelemetry(handler http.Handler) http.HandlerFunc {
-	return promhttp.InstrumentHandlerDuration(serverLatency,
-		promhttp.InstrumentHandlerResponseSize(serverResponseSize,
-			promhttp.InstrumentHandlerCounter(serverCounter, handler)))
+// WithTelemetry instruments the HTTP server with prometheus, and Trace handlers
```
kleimkuhler: Same comment about the comma!

pkg/flags/flags.go
```diff
@@ -33,6 +33,15 @@ func ConfigureAndParse(cmd *flag.FlagSet, args []string) {
 	maybePrintVersionAndExit(*printVersion)
 }
 
+// AddTraceFlags adds the trace-collector and sampling-probability flags
+// to the flagSet and returns their pointers for usage
+func AddTraceFlags(cmd *flag.FlagSet) (*string, *float64) {
+	traceCollector := cmd.String("trace-collector", "", "Enables OC Tracing with the specified endpoint as collector")
+	probabilitySampling := cmd.Float64("sampling-probability", 0, "The probabilistic sampling rate for the traces")
```
grampelberg: Let's not add the ability to change sampling for the control plane. We can add it later and I suspect we'll want it at 100% most of the time.

pkg/util/trace.go
```diff
@@ -0,0 +1,24 @@
...
+
+// InitializeTracing initiates trace, exporter and the sampler
+func InitializeTracing(serviceName string, address string, probability float64) {
+	if address != "" {
+		oce, err := ocagent.NewExporter(
+			ocagent.WithInsecure(),
+			ocagent.WithAddress(address),
+			ocagent.WithServiceName(serviceName))
+		if err != nil {
+			log.Errorf("Couldn't create a OC Agent exporter:%s", err)
```
ihcsim: If `err != nil`, I don't think we want to call `trace.RegisterExporter(oce)` later, right? (I don't think `log.Errorf()` stops the execution.) [`oce` will be `nil`](https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/a8a6f458bbc1d5042322ad1f9b65eeb0b69be9ea/ocagent.go#L98-L104), which might be [problematic for `trace.RegisterExporter()`](https://github.com/census-instrumentation/opencensus-go/blob/3b5a343282fe4b4fccdb0f24cbd1d7169d20858a/trace/export.go#L53). I think we should just return `err` and let `main` deals with it.

pkg/flags/flags.go
Pothulapati: yeah, this will also allow us to have lesser configuration.

grampelberg: @adleong is there a reason you wanted the sampling here? Feels like extra, unneeded configuration that we'd have to support.

adleong: I'd be kinda tempted to leave this flag in here but not expose it through the install flag, just in case?  But I'm also fine taking it out altogether.

grampelberg: I can be talked into just hiding it, worried it'll become stale after awhile though. @Pothulapati it is up to you, my preference would be removing it entirely, but if that's a bunch of work you could just hide the flag.

Pothulapati: Removed it. :p

Pothulapati: Sampling Config has also been removed in #3539

pkg/util/trace.go
```diff
@@ -0,0 +1,27 @@
...
+		if err != nil {
+			return err
+		}
+		trace.RegisterExporter(oce)
+		trace.ApplyConfig(trace.Config{
+			DefaultSampler: trace.AlwaysSample(),
+		})
+		return nil
+	}
+	return errors.New("collector address is empty")
```
ihcsim: Cross-referencing #3539, doesn't `address == ""` imply that tracing is disabled?

pkg/util/trace.go
```diff
@@ -0,0 +1,27 @@
+package util
```
ihcsim: Curious, why not just `pkg/trace`?

pkg/util/trace.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"errors"
+
```
kleimkuhler: Looks like a newline got into the imports! Can this please be removed real quick before a merge?

pkg/util/trace.go
ihcsim: Can we just not call this function in all the `main`s if `*traceCollector` is empty?

pkg/util/trace.go
Pothulapati: Nothing specific, just thought it was a utility. moving it to `pkg/trace`

Pothulapati:
@ihcsim I updated the code to not warn when the collector is empty. 
I don't know how this relates with the config PR, we are using the `enabled` flag there for that, not the `collectorAddress` as we want a default value in helm pointing to `linkerd-collector`.

ihcsim:
@Pothulapati Thanks for the fixes! One last question on the changes to `controller/api/public/prometheus.go`: does that mean `trace.StartSpan()` will always be called even without a collector deployed?

The reason why I brought up https://github.com/linkerd/linkerd2/pull/3539 is because in the new `_trace.tpl` template, if `{{.Enabled}}` is `false`, then the `-trace-collector` argument isn't added, implying that it's empty in the control plane components, further implying that tracing is disabled, leading back to my comment above.

Pothulapati:
@ihcsim Yeah, Irrespective of the collector's endpoint, the prometheus span is started. Even if the prometheus span is started, the tracing logic tries to send those spans, but dosent error out when it can't. I don't think it should be a problem. maybe, @adleong can add something here.

Ah Yes, right. If enabled is false, the value is empty which dosent initialise tracing.

PR #3496, opened 2019-09-30
Prevent '--cluster-domain' from being changed.

### LINKED ISSUE #3454: Disable custom cluster domain changes during control plane upgrade
PR #3360 added a new option to the `linkerd install` command, to support custom cluster domain. The `upgrade` command inherits this option, because it shares the same set of flags with `install`. Since upgrading the trust domain isn't supported, we should fail the upgrade process if the `--cluster-domain` option is used during upgrade, to avoid breaking mTLS.

bmcstdio:
This PR aims at preventing `--cluster-domain` from being changed during `linkerd upgrade`. I am not sure this is all that is necessary, but it can probably be at least a good start. 🙂 Closes #3454.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -454,7 +454,6 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 		"Omit the sideEffects flag in the webhook manifests, This flag must be provided during install or upgrade for Kubernetes versions pre 1.12",
 	)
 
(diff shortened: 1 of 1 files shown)


ihcsim:
@bmcstdio What you said earlier about using `installOnlyFlagSet` is correct. As long as the cluster domain is persisted as part of the `global` config, which it does, we don't need to persist it as an `install` config. See @alpeb comment at https://github.com/linkerd/linkerd2/issues/3454#issuecomment-536606950. Mind making the change in this PR?

bmcstdio:
@ihcsim sure thing, I've just pushed the change (cc @alpeb).

PR #3503, opened 2019-09-30
change notes for edge-19.9.5

scottcarol:


CHANGES.md
```diff
@@ -1,3 +1,13 @@
+## edge-19.9.5
+
+This edge release is a release candidate for `stable-2.6`.
+
+* CLI
+  * Added headers to the JSON output for `linkerd tap`
+* Controller
+  * Added support for a node selector in the Helm chart, so users can choose
```
ihcsim: I will put this under the `Helm` section, because there are no changes to the controller code.

CHANGES.md
```diff
@@ -1,3 +1,13 @@
+## edge-19.9.5
+
+This edge release is a release candidate for `stable-2.6`.
+
+* CLI
+  * Added headers to the JSON output for `linkerd tap`
```
kleimkuhler: I know most readers will understand the meaning of "headers", but I think something like:

> Added request and response headers to the JSON output option for `linkerd tap`

provides a little more detail about what headers have actually been added (and where to look!)

CHANGES.md
ihcsim: I think it will be clearer with, _Added node selector constraints, so users can control which nodes the control plane is deployed to_.

PR #3510, opened 2019-10-01
Add Tilt for better contributors' UX

ihcsim:
I'll like to get some feedback on how we feel about adding Tilt as an alternate development workflow. The goal is to use Tilt to ensure a seamless and easy "getting started" experience for the community, where they don't have to worry about any of the scripts in the `bin` folder.

In addition to insulating the community from our existing CI shell scripts, Tilt can:

1. Auto retag Docker images from `gcr.io/linkerd-io/*` to a repository managed by the users. This is beneficial for community members who need to test advance features like PSP, network policies etc., that aren't easily configurable on local k8s clusters.
1. Generate [snapshots](https://docs.tilt.dev/snapshots) of a developer's local control plane, where build and runtime error logs can be efficiently shared with the maintainers over (free) TiltCloud.

I have updated the `BUILD.md` with more information on how the file watchers are set up and which containers can benefit from the "live updates" feature.

With the `tilt run` command, Tilt:

1. Builds the local CLI
1. Uses the newly built local CLI to run `linkerd install` to generate the control plane YAML
   1. Additional installation options can be provided using the `tilt_options.json`
   1. The proxy version is pinned using the `tilt_options.json`
1. Set up file watchers on the `cli`, `controller`, `pkg`, `web`, `grafana`, `charts` and `proto` folders
1. Only trigger rebuilds when initiated from the Tilt dashboard. This can be configured to always rebuild on every keystroke.

Some pending issues:

1. Tilt continues to stream logs from terminated pods. See https://github.com/windmilleng/tilt/issues/2263.
1. I haven't quite figured out how to handle changes to the Go modules dependencies yet


olix0r:
My main concern with this is that it's Yet Another place where we describe the control plane's manifests; and its subtle and a bit odd that we have to update `bin/_tilt` every time we add/alter control plane components. Ideally, it seems like it would be preferable to _generate_ this spec from some source of truth (i.e. our helm templates)... like if one could do something like `linkerd install --ignore-cluster | tilt-gen`, we wouldn't have to maintain this spec in parallel.

I won't block this if this is actually useful for the broader development team, but it seems like a high maintenance cost to me.

ihcsim:
@olix0r Thanks for the feedback! @dbentley said he's gonna take a look at this.

ihcsim:
I'm gonna closing this draft PR for now, and resuming further discussion in https://github.com/ihcsim/linkerd2-tilt.

PR #3513, opened 2019-10-02
If tap source IP matches many running pods then only show the IP

### LINKED ISSUE #3103: linkerd appears to mislabel incoming traffic from GKE ingresses
### What is the issue?
Inbound traffic to deployments in the "default" namespace is, somehow, being marked as coming from the fluentd-gcp pod in the kube-system namespace in Google Kubernetes Engine clusters.

image

The "live calls" view generates a tap link with a source of the fluentd-gcp pod and a destination of the deployment in question:

```
linkerd tap pod/fluentd-gcp-v3.1.1-svgcw --namespace kube-system --to deployment/auth --to-namespace default --path /ping
```

...but no traffic is ever seen in the tap.  (To be clear the lack of output in the tap is, I believe, _correct_: fluentd-gcp is the log aggregator to stackdriver for GKE clusters, and would not be sending REST queries to our microservices.)

### How can it be reproduced?

Slightly elided, here are our ingress, service and deployment configs:

```
---
kind: Ingress
metadata:
  annotations:
    ingress.gcp.kubernetes.io/pre-shared-cert: ingress-cert-1
    kubernetes.io/ingress.global-static-ip-name: auth-address
  name: auth
  namespace: default
spec:
  backend:
    serviceName: auth
    servicePort: service-port
---
apiVersion: v1
kind: Service
metadata:
  annotations: {}
  name: auth
  namespace: default
spec:
  externalTrafficPolicy: Cluster
  ports:
  - name: service-port
    nodePort: 32163
    port: 54323
    protocol: TCP
    targetPort: pod-port
  selector:
    app: auth
  sessionAffinity: None
  type: NodePort
---
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: '8'
  labels:
    app: auth
  name: auth
  namespace: default
spec:
  progressDeadlineSeconds: 2147483647
  replicas: 1
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      app: auth
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 0
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: auth
    spec:
      containers:
        image: gcr.io/project-id/image:version
        imagePullPolicy: IfNotPresent
        name: auth
        ports:
        - containerPort: 80
          name: pod-port
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /ping
            port: pod-port
            scheme: HTTP
          initialDelaySeconds: 5
          periodSeconds: 5
          successThreshold: 1
          timeoutSeconds: 1
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
```

### Logs, error output, etc

I don't have anything to add here but let me know if there are specific linkerd pod logs that would be helpful to get.

#### `linkerd check` output

```text
$ linkerd check
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-config
--------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane self-check
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √
```

### Environment

- Kubernetes Version:
```
Client Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.9-gke.7", GitCommit:"b6001a5d99c235723fc19342d347eee4394f2005", GitTreeState:"clean", BuildDate:"2019-06-24T19:37:31Z", GoVersion:"go1.10.8b4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.7-gke.25", GitCommit:"d4c79083ab6dea5d26ef4ed8d50b145268349bc3", GitTreeState:"clean", BuildDate:"2019-06-22T16:10:31Z", GoVersion:"go1.10.8b4", Compiler:"gc", Platform:"linux/amd64"}
```
- Cluster Environment: GKE
- Host OS: Google COS
- Linkerd version:
```
Client version: stable-2.4.0
Server version: stable-2.4.0
```

alpeb:
When an unmeshed source ip matched more than one running pod, tap was
showing the names for all those pods, even though they didn't necessary
originate the connection. This could be reproduced when using a pod
network add-on such as Calico.

With this change, if more than one running pod matches the source, only
use the IP as source without attaching any of the pods' metadata.

Fixes #3103 

#### Tested using GKE with Network Policy enabled (Calico), with emojivoto without vote-bot, but the readiness probe hitting the web deployment:

Before:
Screenshot from 2019-10-02 08-22-02

After:
Screenshot from 2019-10-02 12-12-25

### DIFF
--- a/controller/tap/server.go
+++ b/controller/tap/server.go
@@ -603,17 +603,24 @@ func (s *GRPCTapServer) hydrateIPLabels(ip *public.IPAddress, labels map[string]
 	return nil
 }
 
-// resourceForIP returns the pod or node corresponding to a given IP address.
+// resourceForIP returns the node or pod corresponding to a given IP address.
 //
-// If multiple pods exist with the same IP address, this may be because some
-// are terminating and the IP has been assigned to a new pod. In this case, we
-// select the running pod, if only one currently exists. If multiple exist, then
-// we return the node whose internal IP matches (if there's only one).
-//
-// If no pods or nodes were found for the provided IP address, it returns nil. Errors are
-// returned only in the event of an error searching the indices.
+// First it checks if the IP corresponds to a Node's internal IP and returns the
+// node if that's the case. Otherwise it checks the running pods that match the
+// IP. If exactly one is found, it's returned. Otherwise it returns nil. Errors
+// are returned only in the event of an error searching the indices.
 func (s *GRPCTapServer) resourceForIP(ip *public.IPAddress) (runtime.Object, error) {
 	ipStr := addr.PublicIPToString(ip)
+
+	nodes, err := s.k8sAPI.Node().Informer().GetIndexer().ByIndex(ipIndex, ipStr)
+	if err != nil {
+		return nil, err
+	}
+	if len(nodes) == 1 {
+		log.Debugf("found one node at IP %s", ipStr)
+		return nodes[0].(*corev1.Node), nil
+	}
+
 	pods, err := s.k8sAPI.Pod().Informer().GetIndexer().ByIndex(ipIndex, ipStr)
 	if err != nil {
 		return nil, err
@@ -621,9 +628,6 @@ func (s *GRPCTapServer) resourceForIP(ip *public.IPAddress) (runtime.Object, err
 
 	if len(pods) == 1 {
 		log.Debugf("found one pod at IP %s", ipStr)
-		// It's safe to cast elements of `objs` to a `Pod`s here (and in the
-		// loop below). If the object wasn't a pod, it should never have been
-		// indexed by the indexing func in the first place.
 		return pods[0].(*corev1.Pod), nil
 	}
 
@@ -637,14 +641,6 @@ func (s *GRPCTapServer) resourceForIP(ip *public.IPAddress) (runtime.Object, err
 					ipStr,
 					len(pods),
 				)
-
-				nodes, err := s.k8sAPI.Node().Informer().GetIndexer().ByIndex(ipIndex, ipStr)
-				if err != nil {
-					return nil, err
-				}
-				if len(nodes) == 1 {
-					return nodes[0].(*corev1.Node), nil
-				}
 				return nil, nil
 			}
 			singleRunningPod = pod
--- a/controller/tap/server.go
+++ b/controller/tap/server.go
@@ -612,19 +612,21 @@ func (s *GRPCTapServer) podForIP(ip *public.IPAddress) (*corev1.Pod, error) {
 		return objs[0].(*corev1.Pod), nil
 	}
 
+	var singleRunningPod *corev1.Pod
 	for _, obj := range objs {
 		pod := obj.(*corev1.Pod)
 		if pod.Status.Phase == corev1.PodRunning {
-			// Found a running pod with this IP --- it's that!
-			log.Debugf("found running pod at IP %s", ipStr)
-			return pod, nil
+			if singleRunningPod != nil {
+				log.Warnf(
+					"could not uniquely identify pod at %s (found %d pods)",
+					ipStr,
+					len(objs),
+				)
+				return nil, nil
+			}
+			singleRunningPod = pod
 		}
 	}
 
-	log.Warnf(
-		"could not uniquely identify pod at %s (found %d pods)",
-		ipStr,
-		len(objs),
-	)
-	return nil, nil
+	return singleRunningPod, nil
 }
--- a/controller/tap/server.go
+++ b/controller/tap/server.go
@@ -23,11 +23,12 @@ import (
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 )
 
 const requireIDHeader = "l5d-require-id"
-const podIPIndex = "ip"
+const ipIndex = "ip"
 const defaultMaxRps = 100.0
 
 // GRPCTapServer describes the gRPC server implementing pb.TapServer
@@ -510,7 +511,8 @@ func NewGrpcTapServer(
 	clusterDomain string,
 	k8sAPI *k8s.API,
 ) *GRPCTapServer {
-	k8sAPI.Pod().Informer().AddIndexers(cache.Indexers{podIPIndex: indexPodByIP})
+	k8sAPI.Pod().Informer().AddIndexers(cache.Indexers{ipIndex: indexByIP})
+	k8sAPI.Node().Informer().AddIndexers(cache.Indexers{ipIndex: indexByIP})
 
 	return newGRPCTapServer(tapPort, controllerNamespace, clusterDomain, k8sAPI)
 }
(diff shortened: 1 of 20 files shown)


adleong:
I'm a bit confused as to how the screenshot relates to the code change.  Is tap data really used in the octopus graph?  I thought that was purely based on metrics data.  Also, if I'm reading the old code correctly, it looks like it would return the first running pod that it finds with the ip address, not a list or set of pods.  Am I missing something?

alpeb:
@adleong Correct, the Octopus (at least the unmeshed resources part) is fed from tap data. 
This is where the Octopus component is rendered:
https://github.com/linkerd/linkerd2/blob/9e287ace3b178601b47bd4cf4a3e14b2b519dbf4/web/app/js/components/ResourceDetail.jsx#L342-L346
That `unmeshedSources` property comes from here:
https://github.com/linkerd/linkerd2/blob/9e287ace3b178601b47bd4cf4a3e14b2b519dbf4/web/app/js/components/TopModule.jsx#L265-L271

You're also right about the old code behavior. Using my example screenshot above, all those pods (ip-mask, calico, etc) have the same IP `10.128.0.11`. The tap backend asks for the first running pod for that IP, each time a tap event is processed. But apparently that first running pod returned is not always the same; it could be any of the many living in that IP, so they start accumulating in the UI. Makes sense?

grampelberg:
Specifically, this problem arises when there are pods that use the host network on a node. Traffic external to the cluster has the IP address of the local node. When any pods with host networking have been scheduled to that node (calico, fluentd), they have the same IP address as the node. As we're looking up simply pod IP addresses, it ends up selecting pods that are not associated.

The midterm solution here would be to look at nodes first and then pods. That way, we'd label it as a node. The long term solution would be to be aware that it is external traffic and label it that way.

alpeb:
@grampelberg I like the idea of explicitly labeling as "node" traffic. That'd imply granting tap RBAC access to the `nodes` resource to figure the ip. Tap already has a ton of privileges anyway.

Note that I can only observe this situation when there's a network add-on like Calico installed.
If not, any pod with `hostNetwork: true` will indeed take the node's IP, but connections originating from it, or from other unmeshed sources, appear to k8s's API as comming from some other IP that I can't figure out what it corresponds to. For example in a node with a pod CIDR `10.52.0.0/24` that ip is `10.52.0.1`. Still digging...

alpeb:
I could reproduce that non-Calico scenario I described above in Minikube. That IP (10.1.0.1 in this case) turns out to be the ipam gateway in the cni config file:
```bash
$  cat /etc/cni/net.d/k8s.conf 
{
  "cniVersion": "0.3.0",
  "name": "rkt.kubernetes.io",
  "type": "bridge",
  "bridge": "mybridge",
  "mtu": 1460,
  "addIf": "true",
  "isGateway": true,
  "ipMasq": true,
  "ipam": {
    "type": "host-local",
    "subnet": "10.1.0.0/16",
    "gateway": "10.1.0.1",
    "routes": [
      {
        "dst": "0.0.0.0/0"
      }
    ]
  }
}
```
I think it'll be enough for now to just figure the node's ip, and return its metadata if there's a match, and return just the ip otherwise.

alpeb:
The last push changes this so if a node's internal IP matches, it is displayed like so:
image
If no node matches, we proceed to look for a matching pod. If exactly one running pod matches we return it. Otherwise we return just the IP.

controller/tap/server.go
```diff
@@ -568,63 +580,72 @@ func (s *GRPCTapServer) hydrateEventLabels(ev *public.TapEvent) {
...
-		// It's safe to cast elements of `objs` to a `Pod`s here (and in the
-		// loop below). If the object wasn't a pod, it should never have been
-		// indexed by the indexing func in the first place.
-		return objs[0].(*corev1.Pod), nil
+		return pods[0].(*corev1.Pod), nil
 	}
 
-	for _, obj := range objs {
+	var singleRunningPod *corev1.Pod
+	for _, obj := range pods {
```
ihcsim: Will this cause the dashboard to output non-deterministic graphs? E.g., on first load, if there is only one pod that is running, this will return a non-nil `singleRunningPod`. But on refreshing the dashboard if more than one pods are running, then the previous pod won't show up in the dashboard, right?

Now that we explicitly handle node internal IP, will it still be possible for multiple running pods to have the same IP? Is it bad to just issue a warning with the names of the pods if `len(pods) > 0`, and just `return nil, nil`?

alpeb:
> Is the screenshot below indicating that there is a pod with hostNetwork: true talking to my controller? I'd read it as the node is sending some traffic to my controller though.

It could be that, or the kubelet in the node sending probes for example. Unless we start looking into the requests themselves, we can't tell apart those two scenarios. I'm open to suggestions on how to maybe make that more explicit, but I think the current PR at least makes things a little clearer than they were before, saying that "something in that node" is sending traffic, instead of showing random pods that didn't necessarily originate the traffic.

controller/tap/server.go
alpeb: > Will this cause the dashboard to output non-deterministic graphs? E.g., on first load, if there is only one pod that is running, this will return a non-nil singleRunningPod. But on refreshing the dashboard if more than one pods are running, then the previous pod won't show up in the dashboard, right?

Yes. On the first load the pod would have been interpreted as a regular unmeshed pod, and on the second run that assumption would have been proven wrong. In your second run those pods would have `hostNetwork: true` and their IP wouldn't match the node's, which is a possible scenario depending the network config (see [comment above](https://github.com/linkerd/linkerd2/pull/3513#issuecomment-538145228)).

> Now that we explicitly handle node internal IP, will it still be possible for multiple running pods to have the same IP?

Multiple running pods with the same IP is indeed the case when they all have `hostNetwork: true`. 

> Is it bad to just issue a warning with the names of the pods if len(pods) > 0, and just return nil, nil?

What would the dashboard show in that case? Currently in this branch, if that happens and the IP matches the node, we just say "something in that node" is sending unmeshed traffic. If the IP doesn't match the node, we say "IP/1.2.3.4"

ihcsim: > those pods would have hostNetwork: true and their IP wouldn't match the node's

Ah.. somehow k8s API sees it as the CNI gateway IP (e.g. in minikube), instead of the pod IP?

> Multiple running pods with the same IP is indeed the case when they all have hostNetwork: true.

If there are multiple pods sharing the same host IP, wouldn't the change in [`controller/tap/server.go`](https://github.com/linkerd/linkerd2/blob/9a4d97675303bfbf2db60fa75859cb847f2264a8/controller/tap/server.go#L619-L622) be `true` and return?

> What would the dashboard show in that case? 

Not sure tbh :stuck_out_tongue:. Ideally, it will be nice to separate pods using host network from say kubelet traffic. But yeah, we don't need to tackle this here. We can check with @grampelberg if we want to open a separate issue for it.

grampelberg: I think we're getting about as good as we can with this method. It'll be interesting to flesh out something more comprehensive, but I think we need to look at it more deeply first.

ihcsim:
Thanks for the additional unit tests. Still LGTM!

PR #3519, opened 2019-10-03
Make config mount path configurable

adleong:
The base config mount path where Linkerd will look for config files is hardcoded.  This makes it very difficult to run the control plane locally since you need to create or download the Linkerd configmap files to that specific location in your local filesystem.

We make this configurable with a `-config-mount-path` flag which indicates where the config files should be loaded from.  We also add a `fetch-config` script which uses kubectl to download the config files from your cluster to your local filesystem for local development.

### DIFF
--- a/controller/cmd/tap/main.go
+++ b/controller/cmd/tap/main.go
@@ -26,9 +26,10 @@ func Main(args []string) {
 	kubeConfigPath := cmd.String("kubeconfig", "", "path to kube config")
 	controllerNamespace := cmd.String("controller-namespace", "linkerd", "namespace in which Linkerd is installed")
 	tapPort := cmd.Uint("tap-port", 4190, "proxy tap port to connect to")
-	tlsCertPath := cmd.String("tls-cert", pkgK8s.MountPathTLSCrtPEM, "path to TLS Cert PEM")
-	tlsKeyPath := cmd.String("tls-key", pkgK8s.MountPathTLSKeyPEM, "path to TLS Key PEM")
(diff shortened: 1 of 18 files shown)


controller/cmd/tap/main.go
```diff
@@ -26,9 +26,10 @@ func Main(args []string) {
 	kubeConfigPath := cmd.String("kubeconfig", "", "path to kube config")
 	controllerNamespace := cmd.String("controller-namespace", "linkerd", "namespace in which Linkerd is installed")
 	tapPort := cmd.Uint("tap-port", 4190, "proxy tap port to connect to")
-	tlsCertPath := cmd.String("tls-cert", pkgK8s.MountPathTLSCrtPEM, "path to TLS Cert PEM")
-	tlsKeyPath := cmd.String("tls-key", pkgK8s.MountPathTLSKeyPEM, "path to TLS Key PEM")
+	tlsCertPath := cmd.String("tls-cert", pkgK8s.TLSCrtPEM(pkgK8s.DefaultMouthPathBase), "path to TLS Cert PEM")
+	tlsKeyPath := cmd.String("tls-key", pkgK8s.TLSKeyPEM(pkgK8s.DefaultMouthPathBase), "path to TLS Key PEM")
```
siggy: what do you think about updating `bin/fetch-config` to also retrieve these?

adleong:
@siggy great suggestion, thanks!  that's much simpler: https://github.com/linkerd/linkerd2/pull/3530

adleong:
Closed in favor of #3530

PR #3522, opened 2019-10-03
Change notes for edge-19.10.1

scottcarol:


wmorgan:
Shouldn't this be 10.1?

scottcarol:
@wmorgan good catch, will change

wmorgan:
Personally I would love for September to have a few more weeks this year though.

PR #3529, opened 2019-10-04
CHANGES.md and values.yaml for stable-2.6.0

scottcarol:
change notes for stable-2.6.0

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,76 @@
+## stable-2.6.0
+
+This release introduces distributed tracing support, adds request and response
+headers to `linkerd tap`, dramatically improves the performance of the dashboard
+on large clusters, adds traffic split visualizations to the dashboard, adds a
+public Helm repo, and many more improvements!
+
+For more details, see the announcement blog post: TODO
+https://linkerd.io/2019/08/20/announcing-linkerd-2.6/
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: Please see the [upgrade
+instructions](https://linkerd.io/2/tasks/upgrade/#upgrade-notice-stable-2-6-0).
+
+**Special thanks to**: @alenkacz, @arminbuerkle, @bmcstdio, @bourquep,
+@brianstorti, @kevtaylor, @KIVagant, @pierDipi, and @Pothulapati!
+
+**Full release notes**:
+
+* CLI
+  * Added a new `json` output option to the `linkerd tap` command, which exposes
+    request and response headers
+  * Added a public Helm repo - for full installation instructions, see our [Helm
+    documentation](https://linkerd.io/2/tasks/install-helm/).
+  * Added an `--address` flag to `linkerd dashboard`, allowing users to specify
+    a port-forwarding address (thanks @bmcstdio!)
+  * Added node selector constraints to Helm installation, so users can control
+    which nodes the control plane is deployed to (thanks @bmcstdio!)
+  * Added a `--cluster-domain` flag to the `linkerd install` command that allows
+    setting a custom cluster domain (thanks @arminbuerkle!)
+  * Added a `--disable-heartbeat` flag for `linkerd install | upgrade` commands
+  * Allowed disabling namespace creation when installing Linkerd using Helm
+    (thanks @KIVagant!)
+  * Improved the error message when the CLI cannot connect to Kubernetes (thanks
+    @alenkacz!)
+* Controller
+  * Updated the Prometheus config to keep only needed `cadvisor` metrics,
+    substantially reducing the number of time-series stored in most clusters
+  * Introduced `config.linkerd.io/trace-collector` and
+    `config.alpha.linkerd.io/trace-collector-service-account` pod spec
+    annotations to support per-pod tracing
+  * Instrumented the proxy injector to provide additional metrics about
+    injection (thanks @Pothulapati!)
+  * Added Kubernetes events (and log lines) when the proxy injector injects a
+    deployment, and when injection is skipped
+  * Fixed a workload admission error between the Kubernetes apiserver and the HA
+    proxy injector, by allowing workloads in a namespace to be omitted from the
+    admission webhooks phase using the `config.linkerd.io/admission-webhooks:
+    disabled` label (thanks @hasheddan!)
+  * Fixed proxy injector timeout during a large number of concurrent injections
+  * Added support for disabling the heartbeat cronjob (thanks @kevtaylor!)
+* Proxy
+  * Added distributed tracing support
+  * Decreased proxy Docker image size by removing bundled debug tools
+  * Added 587 (SMTP) to the list of ports to ignore in protocol detection (bound
+    to server-speaks-first protocols) (thanks @brianstorti!)
+* Web UI
+  * Redesigned dashboard navigation so workloads are now viewed by namespace,
+    with an "All Namespaces" option, in order to increase dashboard speed
+  * Added Traffic Splits as a resource to the dashboard, including a Traffic
+    Split detail page
+  * Added a `Linkerd Namespace` Grafana dashboard, allowing users to view
+    historical data for a given namespace, similar to CLI output for `linkerd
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+and response headers to `linkerd tap`, dramatically improves the performance of
+the dashboard on large clusters, adds traffic split visualizations to the
+dashboard, adds a public Helm repo, and many more improvements!
+
+For more details, see the announcement blog post: TODO
+https://linkerd.io/2019/08/20/announcing-linkerd-2.6/
+
+To install this release, run: `curl https://run.linkerd.io/install | sh`
+
+**Upgrade notes**: Use the `linkerd upgrade` command to upgrade the control
```
wmorgan: Does this release also require rolling the data plane? cc @ihcsim

CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+    annotations to support per-pod tracing
+  * Instrumented the proxy injector to provide additional metrics about
+    injection (thanks @Pothulapati!)
+  * Added support for `config.linkerd.io/admission-webhooks: disabled` label on
+    namespaces so that the pod creation events in these namespaces are ignored
+    by the proxy injector; this fixes situations in HA deployments where the
+    proxy-injector is installed in `kube-system` (thanks @hasheddan!)
+  * Added support for disabling the heartbeat cronjob (thanks @kevtaylor!)
+* Proxy
+  * Added experimental distributed tracing support
```
wmorgan: Remove experimental

CHANGES.md
```diff
@@ -1,3 +1,77 @@
+## stable-2.6.0
+
+This release introduces experimental distributed tracing support, adds request
```
wmorgan: I don't think there's any good reason to call this experimental. cc @grampelberg

CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+  * Added support for disabling the heartbeat cronjob (thanks @kevtaylor!)
+* Proxy
+  * Added experimental distributed tracing support
+  * Decreased proxy Docker image size by removing bundled debug tools
+  * Added 587 (SMTP) to the list of ports to ignore in protocol detection (bound
+    to server-speaks-first protocols) (thanks @brianstorti!)
+* Web UI
+  * Redesigned dashboard navigation so workloads are now viewed by namespace,
+    with an "All Namespaces" option, in order to increase dashboard speed
+  * Added trafficsplits as a resource to the dashboard, including a trafficsplit
```
wmorgan: IMO this should either be "traffic splits", the two English words, or `TrafficSplit`s, if we want to talk about the K8s objects

CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+  * Added a public Helm repo - this release can be installed with: `helm repo add
+    linkerd https://helm.linkerd.io/stable && helm install linkerd/linkerd2`
+  * Added an `--address` flag to `linkerd dashboard`, allowing users to specify
+    a port-forwarding address (thanks @bmcstdio!)
+  * Added node selector constraints to Helm installation, so users can control
+    which nodes the control plane is deployed to (thanks @bmcstdio!)
+  * Added a `--cluster-domain` flag to the `linkerd install` command that allows
+    setting a custom cluster domain (thanks @arminbuerkle!)
+  * Added a `--disable-heartbeat` flag for `linkerd` `install | upgrade` commands
+  * Fixed the `linkerd endpoints` command to use the correct Destination API
```
Pothulapati: This was a change in controller and can be left out, as this was caused by separating out the destination component, which happened part of a previous edge release. @alpeb

CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+  * Introduced `config.linkerd.io/trace-collector` and
+    `config.alpha.linkerd.io/trace-collector-service-account` pod spec
+    annotations to support per-pod tracing
+  * Instrumented the proxy injector to provide additional metrics about
+    injection (thanks @Pothulapati!)
+  * Added support for `config.linkerd.io/admission-webhooks: disabled` label on
+    namespaces so that the pod creation events in these namespaces are ignored
+    by the proxy injector; this fixes situations in HA deployments where the
+    proxy-injector is installed in `kube-system` (thanks @hasheddan!)
+  * Added support for disabling the heartbeat cronjob (thanks @kevtaylor!)
```
wmorgan: Is this a dup of the `--disable-heartbeat` flag in the CLI?

scottcarol: It's part of the feature but I wanted to keep this in to call out the contributor

CHANGES.md
scottcarol: I see, thanks @Pothulapati!

CHANGES.md
ihcsim: Yes, it does. I think we should just link this to the upgrade notes in the docs. See https://github.com/linkerd/website/pull/547. Once the PR is merged, the official link should be https://linkerd.io/2/tasks/upgrade/#upgrade-notice-stable-2-6-0.

ihcsim:
We should include these:

```
Added Kubernetes events (and log lines) when the proxy injector injects a
deployment, and when injection is skipped
Fixed proxy injector timeout during a large number of concurrent injections
```

The first allows users to use `kubectl describe` and `kubectl get ev` to see proxy injection events. The second is relevant to a few user-raised issues on large clusters.

CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+**Special thanks to**: @alenkacz, @arminbuerkle, @bmcstdio, @bourquep,
+@brianstorti, @kevtaylor, @KIVagant, @pierDipi, and @Pothulapati!
+
+**Full release notes**:
+
+* CLI
+  * Added a new `json` output option to the `linkerd tap` command, which exposes
+    request and response headers
+  * Added a public Helm repo - this release can be installed with: `helm repo add
+    linkerd https://helm.linkerd.io/stable && helm install linkerd/linkerd2`
```
ihcsim: The actual `helm install` command is a bit longer. I think we should provide a link to https://linkerd.io/2/tasks/install-helm/.

CHANGES.md
```diff
@@ -1,3 +1,77 @@
...
+    substantially reducing the number of time-series stored in most clusters
+  * Introduced `config.linkerd.io/trace-collector` and
+    `config.alpha.linkerd.io/trace-collector-service-account` pod spec
+    annotations to support per-pod tracing
+  * Instrumented the proxy injector to provide additional metrics about
+    injection (thanks @Pothulapati!)
+  * Added support for `config.linkerd.io/admission-webhooks: disabled` label on
+    namespaces so that the pod creation events in these namespaces are ignored
+    by the proxy injector; this fixes situations in HA deployments where the
+    proxy-injector is installed in `kube-system` (thanks @hasheddan!)
```
ihcsim: This should be _where all the proxy injector replicas are unavailable_.

CHANGES.md
adleong: I think at least some of how tracing is configured is subject to change.  For example, I believe the `config.alpha.linkerd.io/trace-collector-service-account` annotation was named with `alpha` to indicate that it's going away soon.  @ihcsim or @grampelberg can you confirm?

CHANGES.md
scottcarol: Thanks @ihcsim, I updated to this, can you confirm this is correct?

> Added support for `config.linkerd.io/admission-webhooks: disabled` label on
    namespaces so that the pod creation events in these namespaces are ignored
    by the proxy injector; this fixes situations in HA deployments where all of
    the proxy injector replicas are unavailable (thanks @hasheddan!)

ihcsim: Feels a bit wordy. What about this:

> Fixed a workload admission error between the k8s API server and the HA proxy injector, by allowing workloads in namespace to be omitted from the admission webhooks phase using the `config.linkerd.io/admission-webhooks: disabled` label. (thanks @hasheddan!)

scottcarol:
@ihcsim @wmorgan @adleong @siggy Thanks for the change notes feedback, I've pushed the latest version. I removed `experimental` from the tracing sections but can restore if needed.

PR #3530, opened 2019-10-04
Fallback to defaults when config cannot be loaded

adleong:
When running the destination controller locally, the Linkerd config files which are typically mounted from a configmap are not available.  To facilitate local development, we fall back to default values in this case instead of failing to start up.

### DIFF
--- a/controller/cmd/destination/main.go
+++ b/controller/cmd/destination/main.go
@@ -53,9 +53,8 @@ func Main(args []string) {
 	if *disableIdentity {
 		log.Info("Identity is disabled")
 	} else {
-		if err != nil {
-			trustDomain = global.GetIdentityContext().GetTrustDomain()
-		} else {
+		trustDomain = global.GetIdentityContext().GetTrustDomain()
+		if err != nil || trustDomain == "" {
 			trustDomain = "cluster.local"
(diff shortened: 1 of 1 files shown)


controller/cmd/destination/main.go
```diff
@@ -48,18 +48,24 @@ func Main(args []string) {
...
 	if *disableIdentity {
 		log.Info("Identity is disabled")
 	} else {
-		trustDomain = global.GetIdentityContext().GetTrustDomain()
+		if err != nil {
+			trustDomain = global.GetIdentityContext().GetTrustDomain()
+		} else {
+			trustDomain = "cluster.local"
+			log.Warnf("failed to load trust domain from global config: [%s] (falling back to %s)", err, trustDomain)
+		}
```
siggy: switch `err != nil` check.

going a step further to match the `clusterDomain` block below:
```go
trustDomain = global.GetIdentityContext().GetTrustDomain()
if err != nil || trustDomain == "" {
	trustDomain = "cluster.local"
	log.Warnf("failed to load trust domain from global config: [%s] (falling back to %s)", err, trustDomain)
}
```

adleong: good catch, thanks

PR #3534, opened 2019-10-06
Correct definition of Less function in CLI's metrics command(#3533)

### LINKED ISSUE #3533: CLI: typo in `Less` function in cmd/metrics.go
The definition of function `Less` in the `cli/cmd/metrics.go` is as follows:
``` go
func (s byResult) Less(i, j int) bool {
	return s[i].pod < s[i].pod
}
```
It looks like it's a typo in the line `s[i].pod < s[i].pod` and the correct definition should be(replace the second i with j)

``` go
func (s byResult) Less(i, j int) bool {
	return s[i].pod < s[j].pod
}
```

srv-twry:
Fixes #3533


wmorgan:
Lol

PR #3535, opened 2019-10-07
Add Kurio as adopter

ekapratama93:
There is no turning back now.


wmorgan:
@ekapratama93 Welcome aboard!

PR #3539, opened 2019-10-07
Control Plane Trace configuration

### LINKED ISSUE #3487: Tracing configuration

There are a couple different ways to configure tracing with competing concerns. This issue strives to make each piece explicit.

#### Control Plane

```bash
linkerd [install|upgrade] --control-plane-trace-collector oc-collector.tracing:55678
```

This should:

- Configure each control plane component (eg `linkerd-web`).
- Set `config.linkerd.io/trace-collector` for each proxy used by the control plane.
- Set `-sampling-probability=1` for each control plane component. The assumption is that if tracing is happening on the control plane, someone is actively debugging an issue and would like to see everything.

This should not:

- Configure tracing for anything else in the data plane.

#### Data Plane

```bash
linkerd inject --trace-collector oc-collector.tracing:55678
```

This should:

- Add the `config.linkerd.io/trace-collector` annotation to each spec being passed in.

This should not:

- Configure anything else except the specs that are passed in.

Note: for anyone wanting a default configuration, the annotation should be added at the namespace level.

### Open Questions

- What happens when there's an annotation at the namespace level and a user runs `linkerd inject` on a resource that will live in that namespace? Do they compose?
- In the long term, I would really prefer that linkerd does discovery of the collector's service account. This proposal does not discuss how that is done as it isn't happening today.
- @Pothulapati has mentioned that there's a chicken/egg problem with sending traces to a collector injected with linkerd before the control plane is installed. What are the implications here?

### Out of Scope

- Dynamically enabling tracing for any control plane component or data plane proxy.
- Allow updating control plane injection via `linkerd inject`, see #3396.

Pothulapati:
Fixes #3487 

This PR adds the Control Plane Trace configuration, to make the Control Plane components emit spans along with the proxy spans to the collector. 

This adds the following Trace Configuration values, 
```
# Control Plane Trace Configuration
ControlPlaneTrace:
  Enabled: false
  ProxyTrace:
    CollectorSvcAddr: linkerd-collector.linkerd-tracing:55678 # leave blank to disable tracing
    CollectorSvcAccount: linkerd-collector.linkerd-tracing
```

Installation is done by,
```
linkerd install --control-plane-tracing | k apply -f -
```
This will enable `ControlPLaneTrace.Enabled` flag and also configures the control-plane proxies to use the default values as specified in the config.

For helm installation, Users will have to set the `Proxy.Trace` values to the values as specified above.

Once the installation is done, the collector and jaeger can be installed by running 
```
kubectl apply -f https://gist.githubusercontent.com/Pothulapati/245842ce7f319e8bcd02521460684d6f/raw/39bacaa657255eca59de768fefef10da2835e966/linkerd-tracing.yaml
```

The BUILD.md docs will be updated, with the above manifest to let users install the tracing components once it is enabled.

**Note**

This PR can only be merged after https://github.com/linkerd/linkerd2/pull/3495


Pothulapati:
Examples of Control Plane Component traces with the proxy traces:

Screenshot from 2019-10-04 01-29-23
Screenshot from 2019-10-04 01-33-47

BUILD.md
```diff
@@ -142,6 +142,23 @@ bin/linkerd -n emojivoto stat deployments
...
+
+This will configure all the components to send the traces at `linkerd-collector.linkerd-tracing:55678`
+
+```bash
+
+# install Linkerd with tracing
+linkerd install --control-plane-tracing | kubectl apply -f -
+
+# install OpenCensus collector and Jaeger collector to collect traces
+kubectl apply -f kubectl apply -f https://gist.githubusercontent.com/Pothulapati/245842ce7f319e8bcd02521460684d6f/raw/39bacaa657255eca59de768fefef10da2835e966/linkerd-tracing.yaml
```
adleong: should one of these `kubectl apply -f` be `linkerd inject`?

adleong:
What do you think of having the tracing components installed into the linkerd namespaces instead of a separate `linkerd-tracing` namespace?  Then the collector address would be: `linkerd-collector.linkerd` or actually: `linkerd-collector.{control plane namespace}`.

BUILD.md
Pothulapati: as the namespace has the inject annotation already. `linkerd inject` is not needed. sorry

Pothulapati:
@adleong The problem I think would be, As the `Controlplane-Namespace` is configurable, we can't use `linkerd` by default as you know.

I don't think we can use `{{.Namespace}}` templated variable in `values.yaml` where the traces values are need. even if its possible, I don't think its right to template the values file. @ihcsim

adleong:
@Pothulapati for example, the web deployment template has this line

```
 - -api-addr=linkerd-controller-api.{{.Namespace}}.svc.{{.ClusterDomain}}:8085
```

where the controller service name is hardcoded, but the namespace and cluster domain are templated.  I think we can do the same thing for the collector service.

Pothulapati:
@adleong Updated the PR, to add the linkerd namespace from the linkerd install go code rather than templates. I did not prefer helm templating, because the `SvcAccount` value to be templated, requires it to be placed at `_config,tpl` which aint ideal. This can be done by doing some if-else stuff though. But From the start, control-plane tracing through helm wasn't given priority based on my convo with @grampelberg and @ihcsim as many users wouldnt use it. Happy to change, if there are concerns.

Enabling the Control plane traces through helm require users to add relevant complete values (i.e svcaddr & svcAccount) both at `ControlPlaneTrace` & `Proxy.Trace` parts of the YAML manually.

pkg/charts/values.go
```diff
@@ -74,6 +75,12 @@ type (
 	// ConfigJSONs is the JSON encoding of the Linkerd configuration
 	ConfigJSONs struct{ Global, Proxy, Install string }
 
+	// ControlPlaneTrace contains fields to enable Control Plane Components Tracing
+	ControlPlaneTrace struct {
+		Enabled    bool
+		ProxyTrace *Trace
```
adleong: why is this field called ProxyTrace?  This controls traces emitted directly from the controller, right?

cli/cmd/install.go
```diff
@@ -695,6 +702,16 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*cha
 	installValues.ProxyInit.IgnoreInboundPorts = strings.Join(inboundPortStrs, ",")
 	installValues.ProxyInit.IgnoreOutboundPorts = strings.Join(outboundPortStrs, ",")
 
+	if options.controlPlaneTrace {
```
adleong: I find this block of code very confusing since we do a mutation of `installValues.ControlPlaneTrace.ProxyTrace.CollectorSvcAddr` to append the namespace and port.  And we do a few more relative mutations which are hard to understand.

Does the `installValues.ControlPlaneTrace.ProxyTrace.CollectorSvcAccount` not end with a namespace?  Does `installValues.Proxy.Trace.CollectorSvcAddr` end with a namespace?

What might be simpler would be to change the chart to something like:

```
- -trace-collector=linkerd-collector.{{.Namespace}}.svc.{{.ClusterDomain}}:55678
```

and have the `ControlPlaneTrace` field in Values.go simply be a boolean.  Does that seem reasonable?

ihcsim: I think what's happening here is that we are trying to copy the tracing config from the control plane components to the tracing env vars of the control plane proxies.

I agree that since this is mainly used for control plane debugging, we have liberty over how the config is structured. So something like this makes sense to me:

```
# values.yaml
ControlPlaneTracing: false

# charts/partials/templates/_tracing.tpl
{{ define "partials.linkerd.tracing" -}}
- -trace-collector=linkerd-collector.{{.Namespace}}.svc.{{.ClusterDomain}}:55678
{{- end }}

# charts/partials/templates/_proxy.tpl
# if the whitespaces started to cause malformed YAML, try use `helm template` to debug or DM me
{{ if .ControlPlaneTracing -}}
- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR
  value: linkerd-collector.{{.Namespace}}.svc.{{.ClusterDomain}}:55678
- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME
  value: linkerd-collector.{{.Namespace}}.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
{{ else if .Proxy.Trace -}}
{{ if .Proxy.Trace.CollectorSvcAddr -}}
- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR
  value: {{ .Proxy.Trace.CollectorSvcAddr }}
- name: LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME
  value: {{ .Proxy.Trace.CollectorSvcAccount }}.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
{{ end -}}
{{ end -}}
```

BUILD.md
```diff
@@ -142,6 +142,23 @@ bin/linkerd -n emojivoto stat deployments
 bin/linkerd -n emojivoto tap deploy voting
 ```
 
+#### Deploying Control Plane components with Tracing
```
ihcsim: @grampelberg Any thoughts on if BUILD.md is the right place for this? Feels like it should be in `website`.

pkg/charts/values.go
Pothulapati: Yes, but this implies the svcAccount and svcAddress needed by the Control-plane proxies.

BUILD.md
grampelberg: For now, it should be in `BUILD.md` as this is for folks actively doing development. Once we've got a more general demo tracing control plane solution, it should move over to the website.

Pothulapati:
Updated.

Thanks @adleong @ihcsim

BUILD.md
```diff
@@ -142,6 +142,23 @@ bin/linkerd -n emojivoto stat deployments
...
+
+Control Plane components have the `trace-collector` flag used to enable [Distributed Tracing](https://opentracing.io/docs/overview/what-is-tracing/) for development purposes. It can be enabled globally i.e Control plane components and their proxies by using the `--control-plane-tracing` installation flag.
+
+This will configure all the components to send the traces at `linkerd-collector.linkerd-tracing:55678`
```
adleong: I think this should be `linkerd-collector.linkerd:55678`

Pothulapati: Updated that to `linkerd-collector.{{.Namespace}}.svc.{{.ClusterDomain}}:55678`

adleong:
I think this is ready to merge once the conflicts have been resolved

ihcsim:
@Pothulapati Let's do an integration test run on this branch before merging.

ihcsim:
```
=== PASS: all tests passed
```

Issue #3537, opened 2019-10-07
2.7 control plane upgrade causes 503 errors in the data plane


ihcsim:
After upgrading the stable-2.5 control plane to the latest RC, the injected services started to show 503 errors. This is the result of a recent work to migrate the `destination` container into its own `Deployment`. While restarting all the pods in the data plane (so that the proxies picked up the new spec) resolve the problem, this may not be feasible for large cluster. A workaround may require users to temporarily uninject the services, to avoid downtime.

To reproduce:

1. Install a stable-2.5 control plane
1. Inject and deploy emojivoto 
1. Inject and deploy curl
1. From within the curl pod, run `curl -v web-svc.emojivoto/api/list`
   1. Expect a 200 response
1. Upgrade the control plane to the latest RC. Do not restart the data plane
1. Re-run the `curl -v web-svc.emojivoto/api/list` command
   1. Notice a 503 error
   1. Restarting the emojivoto **and** curl at this point will resolve the problem

Relevant logs from the `web-svc` pod:
```
WARN [   539.134884s] linkerd-destination.linkerd.svc.cluster.local:8086 linkerd2_proxy::proxy::reconnect connect error to ControlAddr { addr: Name(NameAddr { name: "linkerd-destination.linkerd.svc.cluster.local", port: 8086 }), identity: Some("linkerd-controller.linkerd.serviceaccount.identity.linkerd.cluster.local") }: received corrupt message
...
WARN [   835.640294s] resolve{authority=voting-svc.emojivoto.svc.cluster.local:8100}: linkerd2_proxy_resolve::destination::resolution Destination.Get stream error: grpc-status: Unknown, grpc-message: "the request could not be dispatched in a timely fashion"
WARN [   835.640378s] resolve{authority=emoji-svc.emojivoto.svc.cluster.local:8100}: linkerd2_proxy_resolve::destination::resolution Destination.Get stream error: grpc-status: Unknown, grpc-message: "the request could not be dispatched in a timely fashion"
WARN [   839.660914s] linkerd2_proxy::app::profiles error fetching profile for emoji-svc.emojivoto.svc.cluster.local:8100: Status { code: Unknown, message: "the request could not be dispatched in a timely fashion" }
WARN [   839.664641s] linkerd2_proxy::app::profiles error fetching profile for voting-svc.emojivoto.svc.cluster.local:8100: Status { code: Unknown, message: "the request could not be dispatched in a timely fashion" }
[...]

PR #3540, opened 2019-10-07
Re-add the destination container to the controller spec

ihcsim:
This PR re-adds the `destination` container back to the `controller` to ensure backward compatibile upgrade. It retains the previous selector of the `linkerd-destination` service, and adds a new `linkerd-dst` service to front the new `Destination` deployment.

This fix is necessary to avoid data plane downtime during an upgrade to 2.6. 

In summary, after a 2.5 control plane is upgraded to 2.6, 

all existing 2.5 proxies will continue to send their requests to these 2.5 resources:
Deployment: `linkerd-controller`
Service: `linkerd-destination`
Service account: `linkerd-controller`

then, on restart the proxies will be upgraded to 2.6, in which cases all requests will be sent to these 2.6 resources:
Deployment: `linkerd-destination`
Service: `linkerd-dst`
Service account: `linkerd-destination`

Fixes https://github.com/linkerd/linkerd2/issues/3537.

### DIFF
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -81,7 +81,7 @@ spec:
       - args:
         - public-api
         - -prometheus-url=http://linkerd-prometheus.{{.Namespace}}.svc.{{.ClusterDomain}}:9090
-        - -destination-addr=linkerd-destination.{{.Namespace}}.svc.{{.ClusterDomain}}:8086
+        - -destination-addr=linkerd-dst.{{.Namespace}}.svc.{{.ClusterDomain}}:8086
         - -controller-namespace={{.Namespace}}
         - -log-level={{.ControllerLogLevel}}
         image: {{.ControllerImage}}:{{default .LinkerdVersion .ControllerImageVersion}}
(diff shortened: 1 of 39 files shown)


charts/linkerd2/templates/controller.yaml
```diff
@@ -91,6 +110,38 @@ spec:
         volumeMounts:
         - mountPath: /var/run/linkerd/config
           name: config
+      - args:
```
adleong: I think we should also update line 84 in this file to point the public-api at the new linkerd-dst service

ihcsim:
@adleong Unfortunately, changing the service, service account and deployment name isn't as straightforward. It causes a series of integration test failures. I ended up having to make changes to these tests, to a point where I'm not 100% sure if I'm fixing or "altering" the tests.
```
test/endpoints/endpoints_test.go
test/get/get_test.go
test/install_test.go
test/routes/routes_test.go
test/stat/stat_test.go
```
E.g. the [log test](https://github.com/linkerd/linkerd2/blob/c5d68ecb1611aa3e09269c9b9101b71662982403/test/install_test.go#L507-L529) relies on the deployment name and `control-plane-component` label to be the same. If I changed the deployment name, it causes further downstream test failures. And the label is used as a Prometheus label, so not something I want to change here.

Unless we have more time to investigate these test failures, I prefer not renaming these resources now.

PR #3545, opened 2019-10-08
Add the ability to do destination lookups by service cluster ip

### LINKED ISSUE #3444: destination: support resolving service IP addresses
The destination service should `Get` request with a `path` in the form of `IP:PORT`; and when this IP is that of a kubernetes Service IP, serve that service's metadata via a resolution (i.e. as if it were called by its proper name).

This will primarily allow opaque TCP traffic that is destined for a kubernetes service to participate in mesh identity.

adleong:
Fixes #3444 

This change adds support for the destination service to resolve Get requests which contain a service clusterIP as the `Path` parameter.  It returns the stream of endpoints, just as if `Get` had been called with the service's authority.

We do this by creating a `IPWatcher` which wraps the `EndpointsWatcher` and supports lookups by clusterIP.   `IPWatcher` maintains a mapping up clusterIPs to service ids and translates subscriptions to an IP address into a subscription to the service id using the underlying `EndpointsWatcher`.  

Since the service name is no longer always infer-able directly from the input parameters, we restructure `EndpointTranslator` and `PodSet` so that we propagate the service name from the endpoints API response.

For ease of testing, this PR is branched off of #3530


adleong:
Sample output:

```
$
go run controller/script/destination-client/main.go -path 192.168.54.78:80 -method get -addr localhost:8086
INFO[0000] Add:
INFO[0000] labels: map[namespace:emojivoto service:web-svc]
INFO[0000] - 10.0.0.138:8080
INFO[0000]   - labels: map[control_plane_ns:linkerd deployment:web pod:web-98c6f7954-qpqlh pod_template_hash:98c6f7954 serviceaccount:web]
INFO[0000]   - protocol hint: H2
INFO[0000]
```

Notice the path input of `192.168.54.78:80` (the service cluster IP) and the output of `10.0.0.138:8080` (the pod IP).

adleong:
Closed in favor of #3595

PR #3547, opened 2019-10-09
Responsive dashboard sidebar behavior and refactoring Navigation component

scottcarol:
This PR fixes #3467 with browser routing tests coming in a separate PR. It only focuses on the sidebar behavior for mobile, tablet and desktop views, and changes navbar color to the color of the linkerd.io website as suggested by @grampelberg.
Works in the latest Chrome, Safari and Firefox.

**Testing Instructions**

Clone @grampelberg's repo here: https://github.com/grampelberg/linkerd2-mockups
Run the mockups demo, note behavior when browser window is in desktop, mobile and tablet size
Compare to behavior in this branch

**New and Old Screenshots**

This Branch:
Dashboard desktop, mobile size and mobile size w/sidebar
Screen Shot 2019-10-16 at 11 55 37 AM

Screen Shot 2019-10-16 at 11 55 52 AM

Screen Shot 2019-10-16 at 11 56 00 AM

Previous Branch:
Dashboard desktop, mobile size and mobile size w/sidebar
Screen Shot 2019-10-16 at 11 57 09 AM

Screen Shot 2019-10-16 at 11 57 17 AM

Screen Shot 2019-10-16 at 11 57 24 AM

### DIFF
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -4,16 +4,11 @@ import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
 import Badge from '@material-ui/core/Badge';
 import BreadcrumbHeader from './BreadcrumbHeader.jsx';
 import Button from '@material-ui/core/Button';
-import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
-import Dialog from '@material-ui/core/Dialog';
-import DialogActions from '@material-ui/core/DialogActions';
-import DialogContent from '@material-ui/core/DialogContent';
-import DialogContentText from '@material-ui/core/DialogContentText';
-import DialogTitle from '@material-ui/core/DialogTitle';
 import Divider from '@material-ui/core/Divider';
 import Drawer from '@material-ui/core/Drawer';
 import EmailIcon from '@material-ui/icons/Email';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import Hidden from '@material-ui/core/Hidden';
 import IconButton from '@material-ui/core/IconButton';
 import LibraryBooksIcon from '@material-ui/icons/LibraryBooks';
 import { Link } from 'react-router-dom';
@@ -22,6 +17,7 @@ import ListItemText from '@material-ui/core/ListItemText';
 import Menu from '@material-ui/core/Menu';
 import MenuItem from '@material-ui/core/MenuItem';
 import MenuList from '@material-ui/core/MenuList';
+import NamespaceConfirmationModal from './NamespaceConfirmationModal.jsx';
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactRouterPropTypes from 'react-router-prop-types';
@@ -29,7 +25,6 @@ import Toolbar from '@material-ui/core/Toolbar';
 import Typography from '@material-ui/core/Typography';
 import Version from './Version.jsx';
 import _maxBy from 'lodash/maxBy';
-import classNames from 'classnames';
 import { faBars } from '@fortawesome/free-solid-svg-icons/faBars';
 import { faCloud } from '@fortawesome/free-solid-svg-icons/faCloud';
 import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons/faExternalLinkAlt';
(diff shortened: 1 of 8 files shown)


grampelberg:
Looks perfect in chrome!

scottcarol:
This PR fixes #3467 with browser routing tests coming in a separate PR.
Changes navbar color to the color of the `linkerd.io` website as suggested by @grampelberg.
Works in the latest Chrome, Safari and Firefox.

Known issue is SVG render issue in Safari, in which Linkerd logo does not include the proper gradient in mobile view when the sidebar is expanded, then closed. I created a workaround in which the logo gradient is hard-coded as a solid color when drawer closes, but will create an issue and continue researching.

scottcarol:
Following up on the Safari issue: I decided to trigger a browser reload after the mobile pop-up sidebar is closed. This ensures that the Linkerd logo is always displayed correctly and consistently. The reload should not be noticeable to most users, and opening/minimizing the mobile sidebar is an edge use case. 

My previous solution of rendering the logo in Safari without the gradient seemed like worse UX due to the slight change in logo color upon opening/closing the mobile sidebar. Safari `13.0.2` appears to have the gradient rendering issue even with sample code from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient).

web/app/js/components/Navigation.jsx
```diff
@@ -4,16 +4,11 @@ import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
 import Badge from '@material-ui/core/Badge';
 import BreadcrumbHeader from './BreadcrumbHeader.jsx';
 import Button from '@material-ui/core/Button';
-import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
```
scottcarol: Note: I pulled out the sidebar code into a `const` called `drawer` which is why so many lines are changed. Most of the diff lines are because of slightly different tabbing due to restructuring of file.

scottcarol:
To test this PR:
1. Clone @grampelberg's repo here: https://github.com/grampelberg/linkerd2-mockups
2. Run the mockups demo, note behavior when browser window is in desktop, mobile and tablet size
3. Compare to behavior in this branch

scottcarol:
@grampelberg I requested you as a reviewer - you gave a 👍  to an earlier version of this PR, if you could re-run this branch and confirm that the mobile behavior matches your mockups, would love a 🚢

scottcarol:
Thanks for taking a look @alpeb -- I should have titled this PR better, it is specifically for the responsive dashboard sidebar menu behavior. That's a good observation re: clicking on the phone sidebar menu, I think that makes sense to change. 

The other mobile/responsive issues you pointed out are already in existence due to the fact that we've never prioritized designing a truly mobile Linkerd experience. This PR doesn't address those. We may be getting some help with that soon.

scottcarol:
@alpeb In looking again at this branch, the issue with automatically closing the sidebar after the user clicks is that we don't always know when the user is "done" with the sidebar. A user might click the namespace selection button to switch the namespace, then click a workload title like "Pods" to navigate to a resource page. For this reason, my preference is to let the user control when to open and close that sidebar.

alpeb:
Any way to have it behave depending on the case? :grimacing:

scottcarol:
@alpeb We definitely have the ability to make it stay open if the user clicks on the namespace change button, and close it when they click on a specific workload! 

I want to confirm with @grampelberg before making this change, to make sure it matches what he's thinking with the mockups. Also if we make this change, we should decide whether it's for tablet view or just mobile view.

grampelberg:
> @alpeb We definitely have the ability to make it stay open if the user clicks on the namespace change button, and close it when they click on a specific workload!

Yeah! I think that's a great idea.

> Also if we make this change, we should decide whether it's for tablet view or just mobile view.

What's the difference?

scottcarol:
Cool, will make the change. Great idea @alpeb 

I'll treat mobile/tablet the same, but wanted to explicitly call out that we're making that decision, since in the future we may want to differentiate the three viewing experiences (i.e. does a mobile user really need to see the `Control Plane` view, etc.).

grampelberg:
> in the future we may want to differentiate the three viewing experiences

Perfect, I've been spending a bunch of time thinking about our tables and how they scale/scroll and are usable with the mockups. So far I've been assuming two break points, but maybe three would make my life a little bit easier. At the bare minimum, I think it makes sense to just hide some things at small screen sizes (octopus comes to mind).

scottcarol:
@grampelberg For me, the tablet/desktop functionality blends together because I typically have a lot of browser windows open and they may not all be full-size, depending on what else I'm doing on my computer. When I'm using my browser from my mobile phone I have very different goals and expectations -- but I'm curious if this would be different for Linkerd users.

scottcarol:
@alpeb Do you mind opening an issue of the single pod view and share screenshots, along with which browser you're using? I couldn't immediately replicate on my end.

alpeb:
> @alpeb Do you mind opening an issue of the single pod view and share screenshots, along with which browser you're using? I couldn't immediately replicate on my end.

No problemo. I've opened #3601 :+1:

PR #3549, opened 2019-10-09
Add inject support to namespaces

### LINKED ISSUE #3255: auto-proxy support for namespaces

Now that auto-proxy works, please teach `linkerd inject` how to annotate a namespace like:

`kubectl get ns/default -o yaml | linkerd inject -`

So that anything added to said namespace magically gets linkerd mesh support!

Currently, we are running a `kubectl patch`, but this is not as clean.

This is as of linkerd v2.4 stable.

mayankshah1607:
Fixes #3255 

This PR adds support for `linkerd inject` to annotate namespaces as follows : 
`kubectl get ns/default -o yaml | linkerd inject -`
Anything added to the said namespace gets linkerd support.

- Added method `AppendNsAnnotation()` for `ResourceConfig` objects
- Added method `IsNamespace()` for `ResourceConfig` objects - checks if a given config is of Kind namespace
- Updated `linkerd install` golden files
- Updated `inject.go` to make namespace annotations come in through patching - https://github.com/Pothulapati/linkerd2/commit/b49fcdc096c9fd70e959d2ab34f7618155c8e10b
- Relevant unit tests have been added with `overrideAnnotations` as well

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -145,18 +145,27 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 		return bytes, reports, nil
 	}
 
+	//Add auto-proxy support for namespace configs
+	if conf.IsNamespace() {
+		// Namespace config annotations get appended here
+		conf.AppendNsAnnotation(k8s.ProxyInjectAnnotation, k8s.ProxyInjectEnabled)
+	}
+
 	if rt.injectProxy {
 		conf.AppendPodAnnotation(k8s.CreatedByAnnotation, k8s.CreatedByAnnotationValue())
 	} else {
 		// flag the auto-injector to inject the proxy, regardless of the namespace annotation
-		conf.AppendPodAnnotation(k8s.ProxyInjectAnnotation, k8s.ProxyInjectEnabled)
+		// this can also be a namespace resource
+		conf.AppendAnnotation(k8s.ProxyInjectAnnotation, k8s.ProxyInjectEnabled)
 	}
 
 	if len(rt.overrideAnnotations) > 0 {
-		conf.AppendPodAnnotations(rt.overrideAnnotations)
+		conf.AppendAnnotations(rt.overrideAnnotations)
 	}
 
+	// Need help from this part
 	patchJSON, err := conf.GetPatch(rt.injectProxy)
+	fmt.Println(string(patchJSON))
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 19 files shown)


pkg/inject/inject.go
```diff
@@ -110,6 +101,18 @@ type ResourceConfig struct {
 	}
 }
 
+// Workload contains the information for a given conf
+type Workload struct {
```
Pothulapati: As we can check the `kind` directly by 
```go
//IsWorkloadNamespace checks if a given config is of type namespace
func (conf *ResourceConfig) IsWorkloadNamespace() bool {
	return strings.ToLower(conf.workload.metaType.Kind) == k8s.Namespace
}
```

I don't see a point in making `workload` public and taking all this hassle.

cli/cmd/inject.go
```diff
@@ -165,7 +171,9 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	}
 	log.Infof("patch generated for: %s", report.ResName())
 	log.Debugf("patch: %s", patchJSON)
+
```
Pothulapati: can you remove these unnecessary white spaces?

pkg/inject/inject.go
```diff
@@ -869,3 +920,8 @@ func sortedKeys(m map[string]string) []string {
 
 	return keys
 }
+
+//IsNamespace checks if a given config is of type namespace
+func (conf *ResourceConfig) IsNamespace() bool {
```
Pothulapati: I dont think we need this anymore. We can use the `IsWorkloadNamespace()` everywhere as both of these functions are doing the same.

Pothulapati:
This Looks good :100: 

@ihcsim can you take a look?

Pothulapati:
Hmm, @ihcsim that would make the logic simple, b
ut a command like `linkerd inject --proxy-Image XYZ namespace.yaml` would not add the config annoations. 
This may create confusion now that the config annotations are honoured by the proxy-injector during injection.

mayankshah1607:
I'll work on these changes and open another PR.

cpretzer:
@mayankshah1607 do you mind pushing your commits so that we can take a look at the code where the namespace is `nil`?

mayankshah1607:
Hi @cpretzer . I have pushed my commits - https://github.com/mayankshah1607/linkerd2/commit/24f0c0cae76ca159fead9be32c127bf90cea3ca3

mayankshah1607:
Fixed it. The problem was that `conf.workload.obj` itself was `nil` because `conf.ParseMetaYAML()` was not configured to return a namespace kind ResourceConfig. 
Everything works as expected now.

Considering all the factors discussed above, I have opened a new PR as this one has too many unnecessary file changes dealing with the `GetPatch()` method. Hence, I shall be closing this one.

Please review my new PR - https://github.com/linkerd/linkerd2/pull/3607

Issue #3531, opened 2019-10-04
cert-manager integration


grampelberg:
## What problem are you trying to solve?

While linkerd-proxy manages certificate rotation for the actual proxy certs, the core certificate used by the built-in CA (identity) is not managed today. In addition, it can be difficult to create your own external trust root and keys.

### How should the problem be solved?

Many k8s clusters today use [cert-manager](https://github.com/jetstack/cert-manager) as a way to create and refresh certificates. It would be nice to use an existing community standard to solve certificate creation and rotation. To do this, a CLI option can be added:

```bash
linkerd install --self-managed-certs
```

This would assume `linkerd-identity-issuer` is already created in the `linkerd` namespace. If it is not, `install` should fail. The content of this secret should be:

```yaml
apiVersion: v1
kind: Secret
type: kubernetes.io/tls
metadata:
    name: linkerd-identity-issuer
    namespace: linkerd
data:
    ca.crt:  {{ }}
    tls.crt: {{ }}
    tls.key: {{ }}
```

Note:

- The secret generated by cert-manager is a little bit different than the one that is there right now.
- Identity will need to reload the secret as it is rotated by cert-manager.
- `ca.crt` is the content of `linkerd-config.data.global.identityContext.trustAnchorsPem`.
- Documentation will need to explain the cert-manager configuration to get this all setup.

### Open Questions

- Kubernetes allows CSRs and has a root CA cert. Does it make sense to just use all that machinery instead?

### Example

After setting cert-manager up, you can get the cert created by adding the following to your cluster:

```yaml
apiVersion: v1
kind: Secret
type: kubernetes.io/tls
metadata:
  name: linkerd-trust-anchor
  namespace: linkerd
data:
  tls.crt:  {{ .Identity.TrustAnchorsPEM }}
  tls.key: {{ trust anchor key }}
---
apiVersion: certmanager.k8s.io/v1alpha1
kind: Issuer
metadata:
  name: linkerd-trust-anchor
  namespace: linkerd
spec:
  ca:
    secretName: linkerd-trust-anchor
---
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
metadata:
  name: linkerd-identity-issuer
  namespace: linkerd
spec:
  secretName: linkerd-identity-issuer
  duration: 24h
  renewBefore: 1h
  issuerRef:
    name: linkerd-trust-anchor
    kind: Issuer
  commonName: identity.linkerd.cluster.local
  isCA: true
[...]

PR #3552, opened 2019-10-09
Add support for --identity-external-issuer=true

zaharidichev:
This PR adds support for using cert-manager as the source of issuer certs. When we specify 
`--identity-external-issuer=true` the cert, key and trust anchors are loaded from the k8s secret that should already be present. As cert manager rolls certificates, the identity service monitors the path where the secret is mounted and automatically refreshes them. 


cli/cmd/install.go
```diff
@@ -359,6 +362,12 @@ control plane.`,
 				os.Exit(1)
 			}
 
+			shouldExist := options.identityOptions.identityExternalIssuer
+			if err := errIfLinkerdIdentityissuerSecretMissing(shouldExist); err != nil && !options.ignoreCluster {
```
alpeb: You could wrap that condition with `if !options.ignoreCluster && shouldExist` and then have `errIfLinkerdIdentityissuerSecretMissing()` have no args and only check when `shouldExist` is true, to avoid the network calls when `shouldExist` is false.

cli/cmd/install.go
```diff
@@ -458,6 +467,11 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.controlPlaneVersion, "control-plane-version", "", options.controlPlaneVersion, "(Development) Tag to be used for the control plane component images")
 	flags.MarkHidden("control-plane-version")
 
+	flags.BoolVar(
+		&options.identityOptions.identityExternalIssuer, "identity-external-issuer", options.identityOptions.identityExternalIssuer,
+		"Whether to use an external identity issuer (false by default)",
```
alpeb: Nit: the standard followed by other commands is `(default false)`

cli/cmd/install.go
alpeb: As a matter of fact, I'm unable to run a naked `linkerd install` because it's forcing me to have the namespace already. This is probably also why the integration tests are failing.

alpeb:
Ref #3531

cli/cmd/install.go
```diff
@@ -973,6 +1031,82 @@ func (idopts *installIdentityOptions) genValues() (*charts.Identity, error) {
...
+
+	externalIssuerData, err := loadExternalIssuerData()
+	if err != nil {
+		return nil, err
+	}
+
+	creds, err := tls.ValidateAndCreateCreds(externalIssuerData.issuerCrt, externalIssuerData.issuerKey)
+
+	if err != nil {
+		log.Fatalf("Failed to read CA from %s: %s", consts.IdentityIssuerSecretName, err)
```
alpeb: For some reason, `log.Fatalf` isn't producing any output in the CLI so when there's an error the command just fails with an empty output. I would replace this (and the one below) with something like `return nil, fmt.Errorf("Failed to read CA from %s: %s", consts.IdentityIssuerSecretName, err)`

controller/identity/creds_watcher.go
```diff
@@ -0,0 +1,97 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA from %s: %s", fscw.issuerPath, err)
+	}
+
+	if err := creds.Crt.Verify(fscw.roots, fscw.expectedName); err != nil {
+		return nil, fmt.Errorf("failed to verify issuer credentials for '%s' with trust anchors: %s", fscw.expectedName, err)
+	}
+
+	log.Infof("Loaded issuer cert:\nCert: %s\nKey: %s", creds.EncodeCertificatePEM(), creds.EncodePrivateKeyPEM())
```
alpeb: I don't know if we did this elsewhere, but better not log the private key IMO

ihcsim:
@zaharidichev Might wanna branch off the linkerd/linkerd2 repo, instead of a fork. There are a few things that CI doesn't run on forks, including Docker build and integration test, even though they show up with :heavy_check_mark:.

controller/identity/creds_watcher.go
```diff
@@ -0,0 +1,97 @@
...
+	return tls.NewCA(*creds, fscw.validity), nil
+}
+
+// StartWatching starts watching the filesystem for cert updates
+func (fscw *FsCredsWatcher) StartWatching() error {
```
ihcsim: I wonder if we are trying to do too much in this method. How about instead of sending `*tls.CA` to the channel, we just send it an "event", and then let `identity.Service` do most of the TLS-related stuff? 

I think there might be a way to make this and the changes to `pkg/identity/services.go` simpler. I am experimenting with it in this [gist](https://gist.github.com/ihcsim/8580c88f94fffc0c345fc852e4bb38ae). The initializing and switching of the `issuer` object will be interesting.

zaharidichev: Looked at the diff. I though about something similar in the beginning. We can go both ways but my initial idea was that this makes the service harder to test in case we want to add some extensive testing to it. 

Since the service is has quite a bit of logic in there that has to do with issuing certs, we might want to exercise some of these scenarios by having it call into `tls.CA` created from different certs. If that is the case, we will have to put the issuer certs on disk and have different ones for every test case while pinging the service some dummy event. Supplying the `tls.CA` via the channel allows us to mock it in a much easier way I think. 

Wrt to initializing and swapping the `tls.CA`, I dont see much of a problem as long as the lock is held. Maybe I am missing smth here? In any way, we can go either direction. Do not really have a strong opinion.

zaharidichev: Whats more interesting to me however is all the error handling in case we have errors from the error channel or invalid issuer certs written to the file system

cli/cmd/install.go
```diff
@@ -196,9 +198,10 @@ func newInstallOptionsWithDefaults() (*installOptions, error) {
 			enableExternalProfiles: defaults.Proxy.EnableExternalProfiles,
 		},
 		identityOptions: &installIdentityOptions{
-			trustDomain:        defaults.Identity.TrustDomain,
-			issuanceLifetime:   issuanceLifetime,
-			clockSkewAllowance: clockSkewAllowance,
+			trustDomain:            defaults.Identity.TrustDomain,
+			issuanceLifetime:       issuanceLifetime,
+			clockSkewAllowance:     clockSkewAllowance,
+			identityExternalIssuer: false,
```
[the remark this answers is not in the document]
zaharidichev: was trying to avoid updating all the golden files in the case when `identityExternalIssuer=false` but yes you are right - much cleaner :)

cli/cmd/upgrade.go
```diff
@@ -369,7 +370,15 @@ func fetchIdentityValues(k kubernetes.Interface, idctx *pb.IdentityContext) (*ch
 	}, nil
 }
 
-func fetchIssuer(k kubernetes.Interface, trustPEM string) (string, string, time.Time, error) {
+func fetchIssuer(k kubernetes.Interface, trustPEM string, externalIssuer bool) (string, string, time.Time, error) {
+	crtName := k8s.IdentityIssuerCrtName
```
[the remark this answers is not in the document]
zaharidichev: Yes, whats even worse is that there is the possibility for this to change over time and if some people are running older/newer versions of cert-manager the names of their secret keys might be different. So I wonder whether we should make all of this configurable ?

controller/identity/creds_watcher.go
ihcsim: The benefit of moving the TLS code out of the CredsWatcher is that we can convert it into a generic file watcher package, say `pkg/watcher/file`.

In terms of testing the `identity.Service`, I think certain unexported functions (like `loadCredentials` if we choose to move it out of the file watcher) can be mock by using function pointers. There are some examples in `cli/cmd/install.go`. Testing `Certify()` will be difficult even in its current state.

> If that is the case, we will have to put the issuer certs on disk and have different ones for every test case while pinging the service some dummy event. 

Sounds like a good integration test. We can create a separate issue for that, as I assume there will be quite a bit of new test code.

> Whats more interesting to me however is all the error handling in case we have errors from the error channel or invalid issuer certs written to the file system

Good point. My first thought is first of all, the file watcher shouldn't worry about how to handle it; the identity service should. Then, the old cert shouldn't be rotated and the identity service shouldn't go down. As to how we notify the users in addition to generate logs, we can consider publishing k8s events. I don't know if we need to tackle this here, but controller/proxy-injector/webhook.go has some good examples on recording k8s events.

cli/cmd/upgrade.go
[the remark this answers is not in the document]
ihcsim: I think keeping it the way you have it is fine. Making it configurable can be a double-edged sword because the last thing we want is to support cases where users can accidentally provide some unsupported secret keys in the k/v pairs. If cert-manager changed their API, then it's a conscious effort for us and our users to upgrade.

controller/identity/creds_watcher.go
zaharidichev: These are all great points. Thanks!

cli/cmd/install_test.go
```diff
@@ -187,7 +187,7 @@ func TestRender(t *testing.T) {
 	}
 }
 
-func testInstallOptions() (*installOptions, error) {
+func testInstallOptions(addCertDataOptions bool) (*installOptions, error) {
```
adleong: what about always including the certDataOptions but manually unsetting fields in the returned options when necessary?

cli/cmd/upgrade.go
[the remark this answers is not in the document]
adleong: rather than passing a boolean in, can we just check for both filenames?

olix0r:
@adleong 

> It would be really great if we did not have to persist this flag value. If think if we check for both file names, as in my comment above, we will no longer need to persist it.

Can you elaborate on this?

To my reasoning: if it's specified on the commandline, it should be persisted.

ihcsim:
> you MUST do a multistage install:

Not necessarily. Some users already create their own `linkerd` namespace outside of `linkerd install`. (Hence, we have the `{{.InstallNamespace}}` variable in the `values.yaml`.)

> linkerd install config (to create the linkerd namesapce)
> external issuer create secret in linkerd namespace
> linkerd install control-plane reads the issuer secret
> Does that sound right?

Yes.

> Why does the flag need to apply to install config? 

Yes, you are right. Looking closer at the Helm templates, the flag isn't used until the `linkerd install control-plane` phase. cc @zaharidichev

adleong:
> Not necessarily.

Ah, let me rephrase: simply doing a single step install with `linkerd install --identity-external-issue=true` will not work.

>Can you elaborate on this?

As far as I can tell, the `identity-external-issuer` enables the following behavior:
* The issuer secret is NOT created
* The trust root is copied from the issuer secret to the Linkerd config

These are both install behaviors and the `identity-external-issuer` flag doesn't change any behaviors after install.  Thus there is no reason to persist it.  This is great because it creates a very clear loosely coupled boundary between the external issuer and Linkerd.

The exception to this is that the names of the cert and key files that cert-manager creates are different from the ones that Linkerd creates.  Rather than using the `identity-external-issuer` to determine which file name to read, we could check both filenames.

ihcsim:
> Thus there is no reason to persist it. 

My thought around this is that, from a support perspective, there are times when I'll like to know how the control plane is installed, along with the install options used. With this flag persisted, I can tell for sure at one point in an older release, the user had decided to use cert-manager, which will be helpful with certs-related issues, as per https://github.com/linkerd/linkerd2/issues/3578#issuecomment-542088523, it sounds like cert-manager has some sort of revert/reconciliation mechanism, if the user tries to override the managed cert.

zaharidichev:
@ihcsim I see. I thinkg @adleong had some very good points around the lack of need for persisting that as we can deduce even during upgrade whether cert-manager was used by simply looking at the contents of the secret. That being said though, you comments also make sense. So should we persist it after all or not. I guess it does not hurt.  Should I modify it to include it in the `recordableFlagSet` ?

olix0r:
@adleong ok, i see your point. I'm not 100% sold that it's worth the optimization. I don't see any harm in making the persisted flags explicitly reflect the install mechanism (i.e. for diagnostics, etc). Is there a reason that persisting the flag is harmful?

adleong:
Persisting the flag is not harmful but I think the abstraction boundary is more clear if we don't.  i.e. Linkerd simply acts on the issuer secret without knowing how it was created.

And as @zaharidichev said, we can reverse engineer if cert-manager was used by looking at the file names.

zaharidichev:
@adleong @ihcsim added some more changes along the comments you have. Currently I do not persist the flag. Let me know if we want to change that. Also shuffled some things around in the identitiy service. Btw, why is my CI not running on this. Is it because the PR has conflicts ?

adleong:
> why is my CI not running on this.

I... have no idea.  Try pushing a new commit?

siggy:
@zaharidichev yup it's the conflicts. ci merges your PR into master for testing.

ihcsim:
@zaharidichev Let's not worry about persisting the flag for now. I am more keen on getting this tested, and merged so that users can try it out in the next edge. If we see the needs to add it in the future, we just need to change its flagset to `recordableFlagSet` (because we don't need it for `linkerd install config` anymore).

> Btw, why is my CI not running on this

In addition to the conflicts, as a reminder, because your branch is on a fork, certain CI steps (like Docker build and integration tests) won't run even though GitHub Action still gives you the :heavy_check_mark:.

cli/cmd/upgrade.go
```diff
@@ -259,6 +261,7 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
 		return nil, nil, fmt.Errorf("could not build install configuration: %s", err)
 	}
 	values.Identity = identity
+	values.InstallNamespace = !identity.Issuer.External
```
ihcsim: Why is this necessary? My understanding of our earlier conversation around omitting the persistence of the `--identity-external-issuer` flag is that its value isn't needed during upgrade. If during upgrade, the code needs to make decisions based on whether an issuer is external, then I don't think what the `isExternal()` function below is doing is reliable.

zaharidichev: >@ihcsim I see. I think @adleong had some very good points around the lack of need for persisting that as we can deduce even during upgrade whether cert-manager was used by simply looking at the contents of the secret.

Well if this is not set, then an attempt to include the namespace in the produced output will be made, right? Just like I noted here, this can be deduced without persisting. If thats unreliable, lets persist it then. The `isExternal()` function is doing pretty much the same thing that the identity service is doing when looking at the contents of the secret. 

I dont have the full context around the potential use and edge cases here, so its quite difficult for me to decide for sure. But why would it be unreliable? If the secret contains contents from cert manager and the flag is set to not use external issuer, it will not work at all after all. Again, up to you what we do here.

ihcsim: > Well if this is not set, then an attempt to include the namespace in the produced output will be made, right? 

My thought is that `kubectl apply` will leave it alone, unless there are some differences. But yes, if the users created their own `linkerd` namespace, then `linkerd upgrade` should leave it alone, like what you did here.

> lets persist it then.

Sure. Thanks. Sorry for the back-and-forth.

zaharidichev:
closing in favor of #3600

PR #3554, opened 2019-10-09
Fix sorting of metricsResult

bvwells:
Fix sorting of metricsResult. Currently element with index i never sorts before the element with index j.


bvwells:
Just noticed that this is already fixed in #3554!

Issue #3279, opened 2019-08-16
Introduce Go source file to record tools dependencies


siggy:
Some `go generate` commands introduce dependencies into `go.mod`, that are then removed by `go mod tidy`. We work around this through a few steps:
https://github.com/linkerd/linkerd2/blob/6567206b53a29a6c6ddcb5f814546e8da966032b/BUILD.md#updating-docker-dependencies

Instead, introduce source files to explicit record the dependencies:
https://github.com/go-modules-by-example/index/tree/5ec250b4b78114a55001bd7c9cb88f6e07270ea5/010_tools

siggy:
Fixed via #3555

PR #3555, opened 2019-10-10
Switch from using golangci fmt to using goimports

adleong:
CI currently enforcing formatting rules by using the fmt linter of golang-ci-lint which is invoked from the bin/lint script.  However it doesn't seem possible to use golang-ci-lint as a formatter, only as a linter which checks formatting.  This means any formatter used by your IDE or invoked manually may or may not use the same formatting rules as golang-ci-lint depending on which formatter you use and which specific revision of that formatter you use.  

In this change we stop using golang-ci-lint for format checking.  We introduce `tools.go` and add goimports to the `go.mod` and `go.sum` files.  This allows everyone to easily get the same revision of goimports by running `go install -mod=readonly golang.org/x/tools/cmd/goimports` from inside of the project.  We add a step in the CI workflow that uses goimports via the `bin/fmt` script to check formatting.

Some shell gymnastics were required in the `bin/fmt` script to work around some limitations of `goimports`:
* goimports does not have a built-in mechanism for excluding directories, and we need to exclude the vendor director as well as the generated Go sources
* goimports returns a 0 exit code, even when formatting errors are detected


adleong:
CI is failing due to an intentional formatting error to demonstrate this change.  Once this PR has been reviewed and approved, I will fix the formatting error and then merge.

bin/fmt
```diff
@@ -0,0 +1,20 @@
...
+
+# go list will exclude the vendor directory.  We use grep to exclude generated
+# code.
+dirs=$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)
+# go list will list all subdirectories and goimports acts recursively.  This
+# results in certain files being reported multiple time.  Therefore, we must
+# dedup them.
+files=$(goimports -l $dirs | sort | uniq)
+
+if [ ! -z "$files" ]; then
```
klingerf: You could use the `-n` test flag here, which is equivalent to `! -z`.

controller/api/destination/watcher/profile_watcher.go
```diff
@@ -19,7 +19,7 @@ type (
 	// publish the service profile and all future changes for that profile.
 	ProfileWatcher struct {
 		profileLister splisters.ServiceProfileLister
-		profiles      map[ProfileID]*profilePublisher
+		profiles map[ProfileID]*profilePublisher // <-- intentional formatting error to test CI
```
klingerf: 🔪 🔪 🔪

tools.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
+
+package tools
+
+import (
+	_ "golang.org/x/tools/cmd/goimports"
```
klingerf: I didn't quite understand why this file has to exist in perpetuity, since `go.mod` has already been modified to pick up the goimports dependency, but this seems to be the recommended best practice in [this thread](https://github.com/golang/go/issues/25922#issuecomment-402918061).

tools.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
+
+package tools
```
siggy: Mind adding `vfsgen` to this file? That should satisfy #3279 and then allow us to delete the special handling of `go mod tidy`:

```diff
diff --git a/tools.go b/tools.go
index 0a53843a..8ac42785 100644
--- a/tools.go
+++ b/tools.go
@@ -3,5 +3,6 @@
 package tools

 import (
+       _ "github.com/shurcooL/vfsgen"
        _ "golang.org/x/tools/cmd/goimports"
 )
```

In BUILD.md:

 ```bash
 go mod tidy
-bin/build-cli-bin # adds back dependencies specific to `go generate` commands
 bin/update-go-deps-shas
 ```

tools.go
adleong: I don't understand this either.  I'm trying to follow best practices but I'm having such a hard time finding any documentation for the go modules behavior that it's very hard to tell what this does if it's necessary.  Let me do some experimentation...

.github/workflows/workflow.yml
```diff
@@ -98,6 +98,23 @@ jobs:
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v1
+    - name: Format
+      env:
+        GITCOOKIE_SH: ${{ secrets.GITCOOKIE_SH }}
+        # prevent OOM
+        GOGC: 20
+      run: |
+        echo "$GITCOOKIE_SH" | bash
```
siggy: i don't think `GITCOOKIE_SH` or `GOGC` should be necessary. `GITCOOKIE_SH` mitigates rate-limiting issues when downloading go dependencies, and `GOGC` is a special tuning to prevent golangci-lint from OOM'ing.

bin/fmt
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -eu
+
+go install -mod=readonly golang.org/x/tools/cmd/goimports
```
siggy: i may not have my paths set up correctly, but this doesn't work out of the box in my env:
```bash
$ bin/fmt
bin/fmt: 13: bin/fmt: goimports: not found
```

rather than relying on (and modifying) the user's Go env, can you follow the pattern in `bin/lint` and other scripts to download the `goimports` binary into `./target/bin`?

i think it's fine to leave instructions in `BUILD.md` to enable folks to set up their env to use our blessed version of goimports, but this script should be less env-dependent.

.github/workflows/workflow.yml
adleong: Would we not need to potentially worry about downloading go dependencies when running `go install -mod=readonly golang.org/x/tools/cmd/goimports`?

bin/fmt
adleong: The `bin/lint` tool is downloaded differently, it downloads an install shell script and pipes it to sh which creates a binary in the current working directory.  Since we rely on go modules to install the correct version of this tool, I think we're stuck with whatever the behavior of `go install` is.

We could potentially run `mv $GOPATH/bin/goimports ${targetbin}/.goimports` after install, but I'm not sure if that's better.  

Does this work on your machine if you change the the `goimports` invocation to `$GOPATH/bin/goimports`?

.github/workflows/workflow.yml
siggy: Probably less a concern as I wouldn't expect that install command to download many deps, but yeah I guess it doesn't hurt to leave `GITCOOKIE_SH` in for now. FWIW we should be able to remove this in Go 1.13.

bin/fmt
siggy: It's true `bin/lint` is a bit different given the pipe to shell pattern, but we're pretty consistent about wrapping tool downloading and execution this way:

```bash
$ grep "target/bin" bin/*
bin/docker:targetbin="$( cd $bindir/.. && pwd )"/target/bin
bin/goimports:targetbin="${rootdir}"/target/bin
bin/helm:targetbin="$( cd $bindir/.. && pwd )"/target/bin
bin/kind:targetbin="$( cd $bindir/.. && pwd )"/target/bin
bin/kubectl:targetbin="$( cd $bindir/.. && pwd )"/target/bin
bin/lint:targetbin="${rootdir}"/target/bin
bin/protoc:targetbin="$( cd $bindir/.. && pwd )"/target/bin
```

We can still leverage `go.mod` versioning while installing `goimports` into a specific location, by temporarily setting `GOBIN`. Recommend breaking out installing `goimports` from  `bin/fmt` into a new `bin/goimports` script:

`bin/goimports`:

```bash
#!/bin/bash

set -eu

cd "$(pwd -P)"

bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
rootdir="$( cd $bindir/.. && pwd )"
targetbin="${rootdir}"/target/bin

GOBIN=$targetbin go install -mod=readonly golang.org/x/tools/cmd/goimports

goimportsbin=${targetbin}/goimports

exec $goimportsbin "$@"
```

...and then `bin/fmt` can simply call `bin/goimports`:

```diff
diff --git a/bin/fmt b/bin/fmt
index b3e0eabf..b9ea373e 100755
--- a/bin/fmt
+++ b/bin/fmt
@@ -2,19 +2,17 @@

 set -eu

-go install -mod=readonly golang.org/x/tools/cmd/goimports
-
 # go list will exclude the vendor directory.  We use grep to exclude generated
 # code.
 dirs=$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)
 # go list will list all subdirectories and goimports acts recursively.  This
 # results in certain files being reported multiple time.  Therefore, we must
 # dedup them.
-files=$(goimports -l $dirs | sort | uniq)
+files=$(bin/goimports -l $dirs | sort | uniq)

 if [ ! -z "$files" ]; then
   for file in $files; do
-    goimports -d "$file"
+    bin/goimports -d "$file"
   done
   exit 64
 fi
```

bin/goimports
```diff
@@ -0,0 +1,17 @@
...
+
+cd "$(pwd -P)"
+
+bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rootdir="$( cd "$bindir/.." && pwd )"
+targetbin="${rootdir}"/target/bin
+goimportsbin=${targetbin}/goimports
+
+# install goimports if it does not exist
+if [ ! -f "$goimportsbin" ]; then
```
siggy: i think if we're going to do this check (which is great to avoid the download/installation hit), we should version the binary somehow, otherwise when we bump `golang.org/x/tools/cmd/goimports` in `go.mod`, this script won't pick it up.

if that's too much of a pain, we can just install every time.

```bash
$ ls -al target/bin
total 514560
drwxr-xr-x  9 siggy  staff       288 Aug 29 11:38 .
drwxr-xr-x  8 siggy  staff       256 Aug 21 15:34 ..
-rwxr-xr-x  1 siggy  staff  60235692 Jul 25 14:27 .docker-19.03.1
-rwxr-xr-x  1 siggy  staff  22921872 Aug 21 12:05 .golangci-lint-1.17.1
-rwxr-xr-x  1 siggy  staff  36022520 Aug 21 12:10 .kind-v0.4.0
-rwxr-xr-x  1 siggy  staff  38215872 Aug 29 11:39 .kind-v0.5.1
-rwxr-xr-x  1 siggy  staff  48591088 Aug 26 11:13 .kubectl-v1.15.3
-rwxr-xr-x  1 siggy  staff  47149640 Jul 30 09:34 helm-v2.14.3
-rwxr-xr-x  1 siggy  staff   9117784 Jun 15  2018 protoc-3.6.0
```

PR #3560, opened 2019-10-11
Dockerfile-base - remove explicit apt cache clean up

ScottBrenner:
Proposing removing the explicit apt cache clean up in `Dockerfile-base` - it already happens automatically!

From the [official Dockerfile best practices](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#run) (scroll down a bit):
> The official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

### DIFF
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -11,5 +11,4 @@ RUN apt-get update \
         dnsutils \
         iptables \
         jq \
-        nghttp2 \
-    && rm -rf /var/lib/apt/lists/*
+        nghttp2


ScottBrenner:
@cpretzer Mind reviewing?

olix0r:
@ScottBrenner Can you share the output of `docker image ls <image>`?

The current image clocks in at 141MB.

My understanding is that `apt-get clean` is run _before_ we do anything with the image. Because we run `apt-get udpate`, we download a bunch of new cache information, and so it's necessary to remove it to keep the image svelt.

cpretzer:
@ScottBrenner thanks for taking the time to work on optimizing the linkerd builds.

I built the image using the `Dockerfile-base` modified in this PR and it resulted in a larger image:

`gcr.io/linkerd-io/base               scottbrenner                 dbc4e550bb99        45 seconds ago      158MB`

When I built the image using the unmodified `Dockerfile-base`, the output is:
 `gcr.io/linkerd-io/base               scottbrenner                 64d53721dfe4        18 seconds ago      141MB`

The best practices document you link does state that dependent images will remove the `/var/lib/apt/lists/*` directories, and we can see that [here](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L236).

As I think about this being the base image for all of the `linkerd` images, I consider that explicit call to remove these directories has more benefit because the amount of data that is transferred to build the images is smaller. 

The net result of the dependent image is the same, so I like the efficiency of having a smaller base image, rather than relying on the implicit call to remove the directories when the dependent image is built.

ScottBrenner:
Yes, just came to this conclusion ☝️ as well
```
$ docker image ls
no-explicit                                               latest                  1a8c427c584b        16 seconds ago       158MB
with-explicit                                             latest                  9bbfbca5ac2b        About a minute ago   141MB
```

Nevermind!

PR #3572, opened 2019-10-14
Clean username before using as docker image tag

### LINKED ISSUE #3570: Build: Linkerd install fails if the username contains an underscore
## Bug Report
I have been trying to setup Linkerd for local development on my Arch Linux machine following the [comprehensive development configuration](https://github.com/linkerd/linkerd2/blob/master/BUILD.md#comprehensive).

According to the guide building the docker images using `DOCKER_TRACE=1 bin/mkube bin/docker-build` runs fine. 
However, on attempting to install Linkerd using `bin/linkerd install | kubectl apply -f -` leads to failure with `Error: dev-55e4fd18-srv_twry is not a valid version` message.

### What is the issue?
I was able to track down the issue to the `bin/_tag.sh` script. Basically, the issue is that my username on the system is `srv_twry` which contains an underscore. The docker images are generated with the username concatenated with the SHA hash of the `HEAD` commit. As a result, the image name also contains an underscore.

The `linkerd install` commands validates the image name and doesn't allow underscores in the image name, hence the error. https://github.com/linkerd/linkerd2/blob/1039d82547388f361160b419e30d0b6b2051dc36/cli/cmd/install.go#L541 

### How can it be reproduced?
Change your username to have an underscore and follow the comprehensive development configuration instructions.

### Logs, error output, etc
```
Error: dev-55e4fd18-srv_twry is not a valid version
```

#### `linkerd check` output
N/A

### Environment

- Kubernetes Version: 1.16
- Cluster Environment: Minikube
- Host OS: Arch Linux
- Linkerd version: N/A

### Possible solution
1. Force me to change my username: Please don't :)
2. Remove underscores and other forbidden characters from the username before using it in the tag.
3. Allow underscores in the version names.

srv-twry:
Fixes #3570

Reference: https://stackoverflow.com/a/20007439/6748052

### DIFF
--- a/bin/_tag.sh
+++ b/bin/_tag.sh
@@ -24,7 +24,8 @@ head_root_tag() {
     if clean_head ; then
         clean_head_root_tag
     else
-        echo "dev-$(git_sha_head)-$(echo $USER | sed "s/[^[:alpha:].-]//g")"
+        name="$(echo $USER | sed "s/[^[:alnum:].-]//g")"
+        echo "dev-$(git_sha_head)-$name"
     fi
 }
 
--- a/bin/_tag.sh
+++ b/bin/_tag.sh
@@ -24,7 +24,7 @@ head_root_tag() {
     if clean_head ; then
         clean_head_root_tag
     else
-        echo "dev-$(git_sha_head)-$USER"
+        echo "dev-$(git_sha_head)-$(echo $USER | sed "s/[^[:alpha:].-]//g")"
     fi
 }
 


bin/_tag.sh
```diff
@@ -24,7 +24,7 @@ head_root_tag() {
     if clean_head ; then
         clean_head_root_tag
     else
-        echo "dev-$(git_sha_head)-$USER"
+        echo "dev-$(git_sha_head)-$(echo $USER | sed "s/[^[:alpha:].-]//g")"
```
adleong: This might be cleaner to split onto it's own line:
```
name="$(...)"
echo echo "dev-$(git_sha_head)-$name"
```

adleong: What do you think of using `[:alnum:]` to allow numbers as well?

srv-twry: Yes, valid points. I'll incorporate your suggestions. Thanks.

PR #3579, opened 2019-10-15
CLI command to fetch control plane metrics

### LINKED ISSUE #3116: linkerd command to get control plane metrics
The `linkerd metrics` command allows fetching the proxy metrics from any injected pod (including pods in the control plane) but there is not a corresponding way to get metrics directly from the control plane containers, such as the destination controller.  

This can be done manually like so:

```bash
kubectl -n linkerd port-forward deploy/linkerd-controller 9996 &
curl localhost:9996/metrics
```

but it would be nice to have a single `linkered` subcommand for this.

srv-twry:
The CLI command `diagnostics` will fetch the control plane metrics directly from containers.

Sample output: https://gist.github.com/srv-twry/8e5f649edfa8bd34dcee286cbf913d5d

Fixes #3116

### DIFF
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/linkerd/linkerd2/pkg/k8s"
@@ -69,7 +70,7 @@ func newCmdDiagnostics() *cobra.Command {
 			// ensure we can connect to the public API before fetching the diagnostics.
 			checkPublicAPIClientOrRetryOrExit(time.Now().Add(options.wait), true)
 
-			resCount := 0
+			var wg sync.WaitGroup
 			resultChan := make(chan diagnosticsResult)
 			for _, d := range deployments.Items {
 				pods, err := getPodsFor(k8sAPI, controlPlaneNamespace, "deploy/"+d.Name)
@@ -86,8 +87,8 @@ func newCmdDiagnostics() *cobra.Command {
 					}
 
 					for i := range containers {
-						resCount++
-						go func(container corev1.Container) {
+						wg.Add(1)
+						go func(container corev1.Container, wg *sync.WaitGroup) {
 							bytes, err := getDiagnostics(k8sAPI, pod, container, verbose)
 
 							resultChan <- diagnosticsResult{
@@ -96,23 +97,32 @@ func newCmdDiagnostics() *cobra.Command {
 								metrics:   bytes,
 								err:       err,
 							}
-						}(containers[i])
+
+							defer wg.Done()
+						}(containers[i], &wg)
 					}
 				}
 			}
 
 			var results []diagnosticsResult
-			timer := time.NewTimer(options.wait)
-			timedOut := false
 
+			c := make(chan struct{})
+			go func() {
+				defer close(c)
+				wg.Wait()
+			}()
+
+			done := false
 			for {
 				select {
 				case result := <-resultChan:
 					results = append(results, result)
-				case <-timer.C:
-					timedOut = true
+				case <-c:
+					done = true // completed normally
+				case <-time.After(options.wait):
+					done = true // timed out
 				}
-				if len(results) == resCount || timedOut {
+				if done {
 					break
 				}
 			}
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -1,0 +1,207 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	adminHTTPPortName string = "admin-http"
+)
+
+type diagnosticsResult struct {
+	pod       string
+	container string
+	metrics   []byte
+	err       error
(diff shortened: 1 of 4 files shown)


cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,205 @@
...
+			timedOut := false
+
+			for {
+				select {
+				case result := <-resultChan:
+					results = append(results, result)
+				case <-timer.C:
+					timedOut = true
+				}
+				if len(results) == resCount || timedOut {
```
srv-twry: `resCount` stores the count of the total number of containers. Any suggestions on better ways to do this?

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,205 @@
...
+								metrics:   bytes,
+								err:       err,
+							}
+						}(containers[i])
+					}
+				}
+			}
+
+			var results []diagnosticsResult
+			timer := time.NewTimer(50 * time.Second)
```
srv-twry: Any suggestions on time limit?

srv-twry:
@grampelberg @adleong 
I have done some work on this and wanted to know if the PR is heading in the right direction or I am totally in the wrong direction.

Also, I have added some inline questions. Hoping to get some suggestions for them.

cli/cmd/diagnostics.go
grampelberg: Take a look at `--wait` for `check`, let's do something similar to that here.

adleong:
Thanks @srv-twry!  The overall approach here looks good!  I was thinking that this could be part of the `linkerd metrics` command rather than a brand new command.  For example, `linkerd -n linkerd metrics deploy/linkerd-controller` gets the proxy metrics for the linkerd controller.  What do you think about a `--controller` flag which changes this to get the controller metrics instead?  E.g. `linkerd -n linkerd metrics deploy/linkerd-controller --controller`

@grampelberg wdyt?

By making this part of the `metrics` command, you can share some code with the existing metrics code and avoid duplication, especially around things like timeouts.

I like your design of iterating through the list of containers and port looking for ports named admin-http.

cli/cmd/diagnostics.go
adleong: I think that [waitgroups](https://gobyexample.com/waitgroups) are often used for this kind of thing.

srv-twry:
@adleong Thanks for the suggestions.

> I was thinking that this could be part of the linkerd metrics command rather than a brand new command. 

My initial thought was to do the same but @grampelberg suggested to make it a separate command(check his proposal on the linked issue).

adleong:
Ahh, my apologies for joining the discussion now rather than during the initial design on the issue.  My expectation was that the `--controller` flag would look for all control plane containers within the specified resource.  That means that `linkerd metrics -n emojivoto deploy --controller` would return zero results and `linkerd metrics -A -l app=foo --controller` would also return zero results.

My motivations for the suggestion are:
* Simplify the implementation by sharing common code with the existing `linkerd metrics` command
* Allow easier selection of specific control plane components or containers without needing to post process the metrics dump.  e.g. `linkerd metrics -n linkerd deploy/linkerd-destination --controller`

grampelberg:
> Simplify the implementation by sharing common code with the existing linkerd metrics command

Common code should be shared no matter whether it is a separate command or not.

> without needing to post process the metrics dump

I don't understand this one. Why don't we do some brainstorming today, I'm still not a fan but it sounds like I'm missing some things.

adleong:
@grampelberg that's a good point.  I was optimizing for the case when you know exactly which component you want to query, but you're right that it's more common to want to dump everything from the control plane.

@srv-twry sorry for all the back and forth!  I'm on board with the `diagnostics` command as you have it here.  However, I think that `metrics.go` and `diagnostics.go` can be refactored to share a lot of common code.

srv-twry:
@adleong No problem! 

>  However, I think that metrics.go and diagnostics.go can be refactored to share a lot of common code.

Yeah sure. In fact, I was planning to do it after receiving confirmation that I was on the right track :)

grampelberg:
Thanks @srv-twry !

srv-twry:
@adleong Thanks for being patient with this. I was caught up in other stuff and hence wasn't able to devote much time on this. I am back on this now.

The command works now. I have also incorporated your suggestion to use wait groups in order to implement timeouts. Please make sure to check out the linked Github gist for a sample output.

However, I need some pointers regarding code duplication between the metrics and the diagnostics command. I am relatively new to Go and hence am facing some difficulties. I am not able to figure out where and how I could extract code out from both the commands to a separate file in order to remove code duplication. For eg., the handling of timeouts is done in different ways in the commands and hence no scope for metrics command unless you want me to change the way it is being done in the `metrics` command as well.

adleong:
No problem, @srv-twry, I'm happy to give you a few pointers.  

A first step might be to update the `NewProxyMetricsForward` function in pkg/k8s/portforward.go to accept the port name as an argument rather than hardcoding it to ProxyAdminPortName.

You could then update the `getMetrics` function in metrics.go to also accept a port name.

Finally, you could create a function which, given a list of pods and a port name, returns all the metrics for those pods.  All the waitgroup logic would be encapsulated here.

There's a bit more to figure out regarding containers: you'll actually either need to pass around a list of containers rather than pods, or you can make the assumption that port names between proxy and non-proxy containers will never conflict.  I'll leave it to you to figure out what makes sense there.

adleong:
@srv-twry are you still interested in working on this?

srv-twry:
@adleong I am but unfortunately, I am very busy for the next month or so. If it's urgent, then it can be taken up by anyone else.

Thanks

adleong:
In that case, I'm going to close this issue for now since no one is actively working on it.  Once you have some time, feel free to open a new PR or re-open this one.  Looking forward to it!

PR #3581, opened 2019-10-15
Move CNI template to helm

zaharidichev:
This PR moves the cni installation logic from cni-template.go to a helm template. It also simplifies some of the logic around how configuration is being supplied to the plugin. 

Fixes: #2921


grampelberg:
I'm a big +1 to having a separate chart. We should (as a general rule) be keeping individual charts as small as possible and introducing dependencies between them.

zaharidichev:
@alpeb Thanks for pointing that out, was not aware. Will split it up!

zaharidichev:
@alpeb Currently I am reusing the values.go for the cni template. Do we want to split that up as well and have `cni_values.go` with a corresponding `yaml` containing defaults?

alpeb:
> @alpeb Currently I am reusing the values.go for the cni template. Do we want to split that up as well and have `cni_values.go` with a corresponding `yaml` containing defaults?

Yeah, I think it would make sense to have all cleanly separated :+1: People could get confused by those CNI values in the default `values.yaml`

zaharidichev:
@alpeb I split up the CNI parts further and added some tests around that. Let me know if that looks better

zaharidichev:
@alpeb Thanks a lot! If I move it from a fork branch to an internal one the tests should be ran automatically, right ?

alpeb:
> @alpeb Thanks a lot! If I move it from a fork branch to an internal one the tests should be ran automatically, right ?

I'm afraid not. The CNI integration tests only run when merging to master, or when tagging master.

alpeb:
It took me a while to run the integration tests because I had a misconfigured docker install :grimacing: 

They are throwing some errors, could you please address those? It's actually not a big deal to run those tests after all, and they don't even require a cluster.
First do `bin/docker-build-cni-plugin` and then `go test -v github.com/linkerd/linkerd2/cni-plugin/test -integration-tests`.
That'll likely complain that the docker image is missing, I think because they're designed to run in master. To run from the branch, just tag the `gcr.io/linkerd-io/cni-plugin` image with the tag in the error message. After addressing that you'll see the real errors.

Feel free to ping me if you need any assistance.

alpeb:
@zaharidichev this tested well using the CLI, but in order to do it through Helm, a new value in linkerd's `values.yaml` file, something like `ProxyInit.Enabled` will need to be added for the main chart to skip the init containers. I think that can be tackled in a separate PR.

zaharidichev:
@alpeb  there is already `Values.NoInitContainer`, which is not included in the values.yaml file but it defaults to false.  Why is there te need for `ProxyInit.Enabled` in that case?

alpeb:
You're right, I was thrown off by the absence of `NoInitContainer` in `values.yaml`.
After creating some docs explaining how to go about with Helm and CNI, we can add it there, pointing to the docs.

This tested fine in Minikube using the following commands:
```
(Minikube requires UseWaitFlag=true)
helm install -f ~/k8s/helm/linkerd-overrides.yaml --set UseWaitFlag=true charts/linkerd2-cni
helm install -f ~/k8s/helm/linkerd-overrides.yaml --set NoInitContainer=true --set InstallNamespace=false charts/linkerd
```

grampelberg:
Aw man, I hate having PRs fall through the cracks. This looks good to me, @adleong or @ihcsim mind taking a look so we can get two 👍 ?

zaharidichev:
@grampelberg yes its been a while :) the refactor I did here will align nicely with some other work I am doing atm so that's where my desire to get it merged comes from

cli/cmd/main_test.go
```diff
@@ -56,6 +56,8 @@ func diffTestdata(t *testing.T, path, actual string) {
 		return
 	}
 
+	println(actual)
```
ihcsim: Is this for debugging purposes only?

ihcsim:
I just noticed the unresolved conflicts and fork. Please run the integration test on this fork, once the conflicts are resolved, before merging into `master`. Thanks.

PR #3589, opened 2019-10-16
Keep old releases in Helm repo index

### LINKED ISSUE #3561: Helm chart 2.5 disappeared?
Hi, I was using happily Linkerd2 2.5 helm chart. Now I want to simulate upgrade from 2.5 to 2.6 locally, but helm says `chart "linkerd2" matching version "2.5.0" not found in linkerd index.` Can you add the helm chart back on track, pls?

alpeb:
When building the Helm repo index file, keep the references to the old
releases. Also rename and keep the old index file in case
something goes wrong when generating the new one.

Fixes #3561

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -519,12 +519,12 @@ jobs:
     - name: Edge Helm chart creation and upload
       if: startsWith(github.ref, 'refs/tags/edge')
       run: |
-        gsutil rsync cp gs://helm.linkerd.io/edge/index.yaml target/helm/index-pre.yaml
+        gsutil cp gs://helm.linkerd.io/edge/index.yaml target/helm/index-pre.yaml
         bin/helm-build package
         gsutil rsync target/helm gs://helm.linkerd.io/edge
     - name: Stable Helm chart creation and upload
       if: startsWith(github.ref, 'refs/tags/stable')
       run: |
-        gsutil rsync cp gs://helm.linkerd.io/stable/index.yaml target/helm/index-pre.yaml
+        gsutil cp gs://helm.linkerd.io/stable/index.yaml target/helm/index-pre.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yml
```diff
@@ -516,15 +516,15 @@ jobs:
...
         . "$dir/path.bash.inc"
-    - name: Helm chart creation
-      if: startsWith(github.ref, 'refs/tags')
-      run: |
-        bin/helm-build package
-    - name: Edge Helm chart upload
+    - name: Edge Helm chart creation and upload
       if: startsWith(github.ref, 'refs/tags/edge')
       run: |
+        gsutil rsync cp gs://helm.linkerd.io/edge/index.yaml target/helm/index-pre.yaml
```
adleong: what is the `cp` subcommand to `rsync`?  I can't find it on https://cloud.google.com/storage/docs/gsutil/commands/rsync and it does not work for me locally.

alpeb: Good catch. It should be just `gsutil cp`.

alpeb:
> I'm guessing that the helm charts themselves are already versioned and present, but prior to this change, the index file only pointed to the most recent release? So this change merges the previous index file with the current one so that all versions are referenced?

Correct.

> It looks like only the previous index.yaml is downloaded, whereas the current and previous indices are uploaded. This means that exactly one previous index version will ever be stored. Is that intended?

No, unless I'm mistaking somewhere. 
- The github action downloads `index.yaml` (with `gsutil cp`)
- `bin/helm-build package` renames it as `index-pre-$version.yaml`
- The github action uploads the merged `index.yaml` and that `index-pre-$version.yaml`. The older `index-pre-$version.yaml` files with the previous `$version` should remain in that google storage folder.

PR #3597, opened 2019-10-17
Increase `kubectl wait` timeout after `helm delete`

alpeb:
The `kubectl wait` after `helm delete` is timing out because it's taking
slightly more than 40s


alpeb:
Not ready for review yet. Might not be needed after all...

alpeb:
Apparently this won't be needed.

PR #3600, opened 2019-10-18
Cert manager support

zaharidichev:
This pr introduces the ability to use external cert manager via the --identitiy-external-issuer=true flag upon install. Currently the flag is persisted so its easier to figure out which kind of secret is being used during upgrade. The slight complication comes from the fact that cert-manager is storing the data in the secret using different keys than our own implementation. Whether we persist this flag or not is a matter of further discussion I guess but for now you can just go ahead and test this change manually (there are also unit tests) by following these steps: 

**Install Cert Manager**
1. `kubectl create namespace cert-manager`
2. `kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v0.11.0/cert-manager.yaml`

**Create cert manager resources.** 
1. Create the [root certs](https://linkerd.io/2/tasks/generate-certificates/) and create a secret with them

```
kubectl create secret tls linkerd-trust-anchor \
   --cert=ca-crt.pem \
   --key=ca-key.pem \
   --namespace=linkerd
```

2.  Create a cert manager issuer: 
```
apiVersion: cert-manager.io/v1alpha2
kind: Issuer
metadata:
  name: linkerd-trust-anchor
  namespace: linkerd
spec:
  ca:
    secretName: linkerd-trust-anchor
```

3. Create a cert manager certificate: 
```
apiVersion: cert-manager.io/v1alpha2
kind: Certificate
metadata:
  name: linkerd-identity-issuer
  namespace: linkerd
spec:
  secretName: linkerd-identity-issuer
  duration: 1h
  renewBefore: 55m
  issuerRef:
    name: linkerd-trust-anchor
    kind: Issuer
  commonName: identity.linkerd.cluster.local
  isCA: true
  keyAlgorithm: ecdsa
  usages:
  - cert sign
  - crl sign
  - server auth
  - client auth
```

**Inspect the setup**
You can take a look at your secret: 
`kubectl get secret linkerd-identity-issuer -o yaml -n linkerd`

Now, you can install linkerd with proxy configured with a short issuance lifetime to see the whole process in action: 
[bin/linkerd install  --identity-external-issuer=true --controller-log-level=debug --identity-issuance-lifetime=2m  | kubectl apply -f -](url)

With this configuration cert manager rolls the certificate every approximately every five minutes. Proxy sends a CSR every two minutes.  You can take a look at the logs of `identitiy-service`, `cert-manager` as well as the proxies to make sure that all certs are being updated accordingly.


cli/cmd/upgrade.go
```diff
@@ -381,14 +393,18 @@ func fetchIssuer(k kubernetes.Interface, trustPEM string) (string, string, time.
 	if err != nil {
 		return "", "", time.Time{}, err
 	}
+	if externalIssuer {
```
adleong: Similar to what the identity service does, we can easily check the for the existence of different keys.  e.g.

```
keyPEM, ok := secret.Data[k8s.IdentityIssuerKeyName]
if !ok {
    keyPEM = secret.Data[k8s.IdentityIssuerKeyNameExternal]
}
```

cli/cmd/upgrade.go
```diff
@@ -246,7 +253,7 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
 		}
 		configs.GetGlobal().IdentityContext = toIdentityContext(identity)
 	} else {
-		identity, err = fetchIdentityValues(k, idctx)
+		identity, err = fetchIdentityValues(k, idctx, isIssuerExternal(flags))
```
adleong: Is this configuration loaded from the linkerd config?  i.e. are the flags populated from the config map? 

This question might be irrelevant if we can get rid of this boolean as per my other comment.

zaharidichev: Yes, they are coming from the config map

cli/cmd/upgrade.go
zaharidichev: Indeed this is true, and I used to do that right [here](https://github.com/linkerd/linkerd2/pull/3600/commits/470c86a615bf7b01c07b38f67729bbec692ebfaf#diff-7596a8e804160c4bfa0dca95355e8c81R376). But then I read [this here](https://github.com/linkerd/linkerd2/pull/3552#discussion_r336326810) and decided to persist the flag and use it during upgrade. Ironically though, I am not using it in the identity service and I am doing pretty much the opposite of what I am doing here. I am getting lost in all the back and forth. My final opinion is that we have two options: 

1. Never persist it and rely on deducing that from the format of the secret.
2. Persist it and pass a bool everywhere where this decision needs to be made (including as an arg to the identitiy service). 

I am fine with any of these as long as we all agree on one.

pkg/identity/service.go
```diff
@@ -69,6 +154,14 @@ func Register(g *grpc.Server, s *Service) {
 
 // Certify validates identity and signs certificates.
 func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.CertifyResponse, error) {
+	svc.issuerMutex.Lock()
```
ihcsim: I am not sure about locking the entire function. Will it work if we create a `getIssuerSafe() *tls.Issuer` method and perform the lock there?

cli/cmd/upgrade.go
```diff
@@ -192,6 +192,13 @@ func upgradeRunE(options *upgradeOptions, stage string, flags *pflag.FlagSet) er
 	return nil
 }
 
+func isIssuerExternal(flags *pflag.FlagSet) bool {
+	if f := flags.Lookup("identity-external-issuer"); f != nil {
```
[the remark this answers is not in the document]
ihcsim: Oppss.. just saw your recent comment on this. My bad.

pkg/identity/service.go
```diff
@@ -56,9 +62,88 @@ type (
...
+	credentials, err := svc.loadCredentials()
+	if err != nil {
+		return err
+	}
+	svc.updateIssuer(credentials)
+	return nil
+}
+
+func (svc *Service) updateIssuer(newIssuer tls.Issuer) {
+	svc.issuerMutex.Lock()
```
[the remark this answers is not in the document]
ihcsim: Actually nvm, the lock is needed here because of the read op in `Certify()`.

cli/cmd/install.go
```diff
@@ -359,6 +361,14 @@ control plane.`,
 				os.Exit(1)
 			}
 
+			shouldExist := options.identityOptions.identityExternalIssuer
+			if !options.ignoreCluster && shouldExist {
```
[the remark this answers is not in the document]
zaharidichev: Yes, we do want it to continue to work... Pff.... so now I understand what you meant when you said that there are quite few install/upgrade permutations that all need to work and lick together. So really there are only a few options to go about this. 


1. We can ditch the idea of fetching the trust anchors from the secret and always provide it via `--identity-trust-anchors-file`
2. When we have `--ignore-cluster && --identity-external-issuer` we **require** that `--identity-trust-anchors-file` is provided. 

If this is not too much cognitive luggage for the user, I would go with option 2. This will solve the problem and allow us to still benefit from fetching the trust-anchors from the secret when cluster does not need to be ignored. WDYT ?

cli/cmd/install.go
```diff
@@ -359,6 +361,14 @@ control plane.`,
...
+				if err := errIfLinkerdIdentityIssuerSecretMissing(); err != nil {
```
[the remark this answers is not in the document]
zaharidichev: You mean we just remove this `errIfLinkerdIdentityIssuerSecretMissing()` check and leave it to the `readExternallyManaged()` to act as an implicit check as well ?  Makes sense to me

pkg/identity/service.go
zaharidichev: It will work, but now that I look at it, I think we can use `RWMutex` and just do a `RLock/RUnlock` to eliminate any potential contention in this block that you might be worried about.

controller/cmd/identity/main.go
```diff
@@ -37,6 +38,11 @@ func Main(args []string) {
 		"/var/run/linkerd/identity/issuer",
 		"path to directory containing issuer credentials")
 
+	externalIssuerPathCrt := filepath.Join(*issuerPath, k8s.IdentityIssuerCrtNameExternal)
```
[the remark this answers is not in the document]
zaharidichev: Good point! I already tried that one BUT this necessitates copying the `k8s` package into the proxy container so the `proxy-identitiy` can be built. This on top requires some other packages required by `k8s` to be copied as well. So decided that its not worth it to introduce this dependency just so I can reduce the number of fields the identitiy service takes.

cli/cmd/install.go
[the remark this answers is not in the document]
ihcsim: My thought is that the output of `--identity-external-issuer --ignore-cluster` should be the same as `--identity-external-issuer`, i.e. without the namespace and `linkerd-identity-issuer` secret, ignoring whether the namespace and secret actually exists in the cluster. We can confirm with @grampelberg tomorrow.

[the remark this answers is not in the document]
zaharidichev: Yes, I get it and agree. As I said, the two possible way I see to make this happen is to make it do we require a trustAnchor file when we have `--identity-external-issuer --ignore-cluster` because in this case we dont fail because of a "check" but because we dont have anywhere to get the trust anchors from (which lives in the secret by default).

controller/cmd/identity/main.go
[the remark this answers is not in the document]
ihcsim: Ah.. I see what you mean. Makes me wanna move those new constants from `pkg/k8s` to `pkg/identity` :wink: But yeah, it's super minor.

pkg/identity/service.go
ihcsim: Good call on `RWMutex`. Will it be used to lock just the `service.issuer` field or the `Certify` function?

zaharidichev: The lock does not lock a field, it locks a critical section. In this case there will be nothing waiting (being blocked) on read, despite the fact that the whole certify function is a critical section. Except for when there is concurrent write performed by the update function which in practice happens so infrequently (how often issuer certs are rotated?) and is such a quick operation that it should not matter in practice.  Does that make sense?

zaharidichev:
@ihcsim Addressed some more comments and implemented k8s events publishing.

cli/cmd/upgrade.go
adleong: My preference continues to be that we do not persist it, but I agree with you that it's more important that we be consistent.

cli/cmd/install.go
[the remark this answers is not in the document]
ihcsim: Per post standup discussion, we won't be supporting `linkerd install --identity-external-issuer --ignore-cluster`.

ihcsim:
@zaharidichev WDYT of always outputting the namespace YAML during `install` and `upgrade`? @grampelberg pointed out that even though users are required to pre-create the `linkerd` namespace, `linkerd install --identity-external-issuer` can still output the namespace YAML, which will save users the effort of manually labeling and annotating the pre-created namespace. 

It will also ensure that `linkerd install --ignore-cluster | kubectl delete -f -` will work, regardless of an external issuer or not. In the external issuer case, `kubectl delete` will delete the entire `linkerd` namespace, including the pre-created cert-manager's CA issuer and secret.

pkg/identity/service.go
```diff
@@ -56,9 +67,92 @@ type (
...
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA from disk: %s", err)
+	}
+
+	if err := creds.Crt.Verify(svc.trustAnchors, svc.expectedName); err != nil {
+		return nil, fmt.Errorf("failed to verify issuer credentials for '%s' with trust anchors: %s", svc.expectedName, err)
+	}
+
+	log.Infof("Loaded issuer cert: %s", creds.EncodeCertificatePEM())
```
ihcsim: Maybe either set this to `Debugf` or just remove it? Currently, my logs look like this:
```
linkerd-identity-7d95b7477-rvqgp identity time="2019-10-22T21:03:07Z" level=info msg="updated identity issuer"
linkerd-identity-7d95b7477-rvqgp identity time="2019-10-22T21:05:39Z" level=info msg="Loaded issuer cert: -----BEGIN CERTIFICATE-----\nMIIBrDCCAVKgAwIBAgIRANG6lRv4YowiYhSyl/AAMjkwCgYIKoZIzj0EAwIwKTEn\nMCUGA1UEAxMeaWRlbnRpdHkubGlua2VyZC5jbHVzdGVyLmxvY2FsMB4XDTE5MTAy\nMjIxMDQ1OFoXDTE5MTAyMjIyMDQ1OFowQDEVMBMGA1UEChMMY2VydC1tYW5hZ2Vy\nMScwJQYDVQQDEx5pZGVudGl0eS5saW5rZXJkLmNsdXN0ZXIubG9jYWwwWTATBgcq\nhkjOPQIBBggqhkjOPQMBBwNCAATZZJmCsTRlJ2dXbFuZcomPxwXduLuFQir8mZM3\nJ1FNLOsZCJW8tj0AHtgy2zmYcCd5U3ul6PavQa7tUzyjahobo0QwQjAOBgNVHQ8B\nAf8EBAMCAqQwDwYDVR0TAQH/BAUwAwEB/zAfBgNVHSMEGDAWgBRTBvulIY9M/owA\nlSNIgPgiUQG30TAKBggqhkjOPQQDAgNIADBFAiEAttkcAJM2BbO8y1Lgv/MN1ueR\nD1BGHvtMvWeENTN95BsCIB7ozpWURXOxksXxa7MzcpJGgjMPzvYuz8QF+1MEqwuZ\n-----END CERTIFICATE-----\n"
linkerd-identity-7d95b7477-rvqgp identity time="2019-10-22T21:05:39Z" level=info msg="updated identity issuer"
linkerd-identity-7d95b7477-rvqgp identity time="2019-10-22T21:06:41Z" level=info msg="Loaded issuer cert: -----BEGIN CERTIFICATE-----\nMIIBqzCCAVKgAwIBAgIRAIhJ4cXXxjT3yRg1PEf6PxIwCgYIKoZIzj0EAwIwKTEn\nMCUGA1UEAxMeaWRlbnRpdHkubGlua2VyZC5jbHVzdGVyLmxvY2FsMB4XDTE5MTAy\nMjIxMDU0OFoXDTE5MTAyMjIyMDU0OFowQDEVMBMGA1UEChMMY2VydC1tYW5hZ2Vy\nMScwJQYDVQQDEx5pZGVudGl0eS5saW5rZXJkLmNsdXN0ZXIubG9jYWwwWTATBgcq\nhkjOPQIBBggqhkjOPQMBBwNCAATDonFjW90A6tiLgBX/T4m+UsQe/ey97h+SNgEp\nEzixkMKLYqlrzjmuiqyrGgdwVRCk0+9uqptFrTCbRp8g8sbAo0QwQjAOBgNVHQ8B\nAf8EBAMCAqQwDwYDVR0TAQH/BAUwAwEB/zAfBgNVHSMEGDAWgBRTBvulIY9M/owA\nlSNIgPgiUQG30TAKBggqhkjOPQQDAgNHADBEAiBQtAEPgtMjwKQacOysnGP13S01\nR18xxhXo+GaUfpN28AIgYTMvKzZ0enLxELMgTDknOntx8XTpIhac8yiAMYcvCnQ=\n-----END CERTIFICATE-----\n"
```

pkg/identity/service.go
```diff
@@ -56,9 +67,92 @@ type (
...
+		select {
+		case <-issuerEvent:
+			if err := svc.Initialize(); err != nil {
+				message := fmt.Sprintf("skipping issuer update as certs could not be read from disk: %s", err)
+				log.Warn(message)
+				svc.recordEvent(v1.EventTypeWarning, eventTypeSkipped, message)
+			} else {
+				message := "updated identity issuer"
+				log.Infof(message)
+				svc.recordEvent(v1.EventTypeNormal, eventTypeUpdated, message)
```
ihcsim: Nit: Can we use `strings.ToUpper(message)` to make the output consistent with the rest of the k8s events? This is what it currently looks like on my cluster, where the `IssuerUpdated` looks a bit different from the rest:
```
⚡   k -n linkerd get ev | grep -i issuer                                                                        
10m         Normal    CertificateIssued   certificaterequest/linkerd-identity-issuer-860701636   Certificate fetched from issuer successfully
5m41s       Normal    CertificateIssued   certificaterequest/linkerd-identity-issuer-860701636   Certificate fetched from issuer successfully
41s         Normal    CertificateIssued   certificaterequest/linkerd-identity-issuer-860701636   Certificate fetched from issuer successfully
10m         Normal    GeneratedKey        certificate/linkerd-identity-issuer                    Generated a new private key
41s         Normal    Requested           certificate/linkerd-identity-issuer                    Created new CertificateRequest resource "linkerd-identity-issuer-860701636"
41s         Normal    Issued              certificate/linkerd-identity-issuer                    Certificate issued successfully
37s         Normal    IssuerUpdated       deployment/linkerd-identity                            updated identity issuer
```
Also, on line 95 above.

cli/cmd/upgrade.go
ihcsim: Let's chat about this at tomorrow's standup. I am starting to wonder if we need the flag at all. It really boils down to are we integrating with the secret structure of an external issuer, or are we just enabling the usage of a pre-created secret of type [`SecretTypeTLS`](https://github.com/kubernetes/api/blob/816a9b7df6780e786246b6e46caadaed8d17e34c/core/v1/types.go#L5490-L5495)? 

Essentially, what cert-manager does is create a secret of that type (just like `kubectl create secret tls my-tls --cert=cert.pem --key=key.pem`, and insert an additional `ca.crt` in there (along with some additional labels and annotations). I think supporting the [`SecretTypeTLS`] secret provides a cleaner integration, because it's a k8s primitive. But it will require us to not depends on it for the trust root.

ihcsim:
So far, everything on this branch is working well for me. `linkerd [install|upgrade]` works as expected. The trust anchor in the config map matches that in the secret generated by cert-manager. Cert auto-rotation and reconciliation (when the secret is manually deleted) also work as expected. I'm gonna try out `helm [install|upgrade]` next.

Also, the `--identity-external-issuer` flag is currently available to the `upgrade` command. I think this should be disabled, because if users decided to use this during upgrade, they will also need to update the trust anchor in the config map. I think the `--from-manifest` option will be a better fit.

cli/cmd/upgrade.go
zaharidichev: Ummm, yes you are right this is what cert-manager does. I guess we really need to aks ourselves whether we want the benefit of having the ca.crt in the secret. I think its quite useful in fact and if users are using cert-manager specifically then its a good thing to have. Supporting the `SecretTypeTLS` as a cleaner integration path is quite nice especially considering the fact that the format of our secret that we generate by the cli is different than the format of `SecretTypeTLS`, so we might want to change that as well. But isnt all of this a bit out of scope. Not sure, we need to discuss.

zaharidichev:
@ihcsim 

> Also, the --identity-external-issuer flag is currently available to the upgrade command. I think this should be disabled, because if users decided to use this during upgrade, they will also need to update the trust anchor in the config map. I think the --from-manifest option will be a better fit.

Would that essentially mean moving the flag from `recordableFlagSet` to `installPersistentFlagSet` ?

ihcsim:
@zaharidichev Per post standup conversation, 

1. We will extend the [`IdentityContext`](https://github.com/linkerd/linkerd2/blob/0f9ea553d2ab9b985e6d9666e95dae2260c26210/proto/config/config.proto#L73) to persist the external issuer flag
1. Although the `--identity-issuer-identity` flag will remain a boolean type, the persisted value should be a string type, so that any future changes (i.e. supporting other certificate managers) won't require a change to the protobuf definition. IMO, for now, even just a `true` string seems reasonable.
1. The external issuer flag doesn't need to be part of `recordableFlagSet` anymore. It should be an install-only flag.

cc @adleong @grampelberg @olix0r

adleong:
One implication is that we should never guess the TLS file names, but should instead use the persisted value to determine which files to load.

I'd prefer a more meaningful string value than "true".  I'd consider something like "linkerd" or "linkerd-legacy" for the Linkerd generated issuer and something like "cert-manager", "tls", "kuberenetes.io/tls", or something like that for the cert manager generated issuer.

grampelberg:
> I'd prefer a more meaningful string value than "true".

If we're going this direction, I think it starts to make sense that the CLI flag is a string and has three values for now:

- `auto` - we generate everything, errors out if `--identity-*` is used.
- `manual` - errors out if `--identity-*` is *not* used.
- `use-secret-with-these-two-files` - uses `linkerd-identity-issuer`, errors out if `--identity-*` is used.

I've picked a long name for the third one because I can't come up with a good name and it should be changed.

adleong:
May I suggest "external" for the third option

adleong:
Going forward it may make sense to simplify this by removing the `manual` mode and removing all of the `--identity` flags, and replacing them with documentation explaining how to take your PEMs and put them into an issuer secret.  i.e. how to be your own cert manager.

grampelberg:
> May I suggest "external" for the third option

I'm not a huge fan of `external` because we're doing this so that it can be changed structurally to begin with.

> replacing them with documentation explaining how to take your PEMs and put them into an issuer secret

I got excited about this one, then thought about the damage it'd do to documentation. I'm not a big fan, but we can look again once everything lands.

ihcsim:
>  so that it can be changed structurally to begin with.

IIUC, almost feels like this should be expressed differently from whether `linkerd install` need to output the issuer secret YAML or not. I'm gonna take a stab here:

`linkerd install`  translates to: 

* `IdentityContext.Issuer=auto`
* `IdentityContext.Issuer.Scheme=linkerd.io/tls`

`linkerd install --identity-*` translates to:

* `IdentityContext.Issuer=manual`
* `IdentityContext.Issuer.Scheme=linkerd.io/tls`

`linkerd install --identity-external-issuer`  translates to:

* `IdentityContext.Issuer=external` 
* `IdentityContext.Issuer.Scheme=[kubernetes.io/tls | etc... ]`

We can add an `--identity-issuer-scheme` to `install` later. As long as we represent it as a string type, we can expand the list as we see fit.

adleong:
Thanks, @ihcsim!  Do we need both `IdentityContext.Issuer` and `IdentityContext.Issuer.Scheme`?  It feels a bit redundant.  Just having `IdentityContext.Issuer.Scheme` should be sufficient, right?

zaharidichev:
I reworked it a bit to store the Scheme in `IdentitiyContext` and just use that. Have not touched the flags but I think this is relatively easy to change if we want to in another PR. Now that I gave it a bit more thought, I really like this approach with the scheme string value.

ihcsim:
@adleong Not sure tbh. At this point, I'm going by whatever gives us the most flexibility down the road. I see them conveying different things. Will we want to support the case where users want to load their own l5d format secret from a remote source? Or a k8s TLS secret from a file?

Also, if we think `IdentityContext.Issuer.Scheme` is the right approach, maybe we can just use shorthand like `l5d`, `k8s` etc.

@grampelberg Thoughts?

grampelberg:
I think all of this can change after the merge. My current preference is to get this in and then address the CLI flags later.

cli/cmd/install.go
```diff
@@ -906,6 +914,19 @@ func errIfLinkerdConfigConfigMapExists() error {
 	return fmt.Errorf("'linkerd-config' config map already exists")
 }
 
+func checkFiles(files []string) error {
```
adleong: TIOLI: should this be called `checkFilesExist`?

controller/cmd/identity/main.go
```diff
@@ -102,27 +109,70 @@ func Main(args []string) {
...
 	}
 
-	svc := identity.NewService(v, ca)
+	// Create K8s event recorder
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{
+		Interface: k8sAPI.CoreV1().Events(controllerNS),
+	})
+	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: componentName})
+	deployment, err := k8sAPI.ExtensionsV1beta1().Deployments(controllerNS).Get(componentName, v1machinary.GetOptions{})
```
adleong: it sucks that we have to lookup our deployment in order to be able to record events about it.  what a weird API 😬

charts/linkerd2/README.md
```diff
@@ -96,6 +96,7 @@ The following table lists the configurable parameters of the Linkerd2 chart and
 |`ControllerReplicas`                  | Number of replicas for each control plane pod                                                   |`1`|
 |`ControllerUID`                       | User ID for the control plane components                                                        |`2103`|
 |`Identity.Issuer.ClockSkewAllowance`  | Amount of time to allow for clock skew within a Linkerd cluster                                 |`20s`|
+|`Identity.Issuer.External`            | Whether to use external issuer for managing certificates                                        |`false`|
```
adleong: I think this needs to be updated to Identity.Issuer.Scheme, right?

charts/linkerd2/templates/identity-rbac.yaml
```diff
@@ -16,6 +16,12 @@ rules:
 - apiGroups: ["authentication.k8s.io"]
   resources: ["tokenreviews"]
   verbs: ["create"]
+- apiGroups: ["extensions"]
+  resources: ["deployments"]
+  verbs: ["get"]
```
adleong: this makes me sad

controller/cmd/identity/main.go
zaharidichev: Yeah,  I was a bit perplexed by that as well... its not great, especially due to the additional permissions required...

PR #3602, opened 2019-10-18
Replace actions/checkout with actions/upload/download

alpeb:
So far just a test, please don't review

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -8,16 +8,20 @@ on:
     tags:
     - "*"
 
-# Jobs by event type and dependencies:
+## Jobs by event type and dependencies:
 
-# Unit tests for every master/tag push and PR
+# Run as a preliminary before starting the other jobs:
+#
+# download_src
+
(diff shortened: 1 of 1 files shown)


alpeb:
@siggy this is ready for review. I've removed all the `action/checkout` entries except for the one at the top. It ended up in quite a lot of repetitions, but I don't know if there's anything we can do about it.  Let me know what you think.

alpeb:
@siggy this is ready for review again. The tests were green!

alpeb:
Notes for reviewers:

This replaces the multiple calls to `actions/checkout` as a workaround to actions/checkout#27
With this PR, there's a new `download_src` job at the beginning that calls `actions/checkout`, tars the `.git` directory and then uploads it into the builds' artifact storage using `actions/upload-artifact`.

Ever other job depends on `download_src`. Now instead of calling `actions/checkout` again, they use `actions/download-artifact` to download the tarball, untar it into the runner's temporary storage dir, and clone from it into the workspace.

PR #3603, opened 2019-10-18
Add direct edges integration test

adleong:
Add an integration test which exercises the behavior when one meshed pod connects to another meshed pod by pod ip address.

The current behavior is that the Linkerd proxy will not do any lookup against the destination service for this kind of connection and will proxy directly to the SO_ORIG_DST.  This means that it will not have the identity metadata necessary to TLS the connection, and the connection will not be present in the `linkerd edges` command output.  This test validates that behavior.

The purpose of this test is to set the stage for future work which will allow the Linkerd proxy to TLS this type of connection and display it in `linkerd edges`.  The assertions in this test will be updated as part of that work.

This test will be run as part of the integration test suite.  It can also be run directly:

```
go test --failfast --mod=readonly test/install_test.go   --linkerd=(pwd)"/bin/linkerd" --k8s-context="$CTX" --integration-tests
go test -v --mod=readonly test/edges/edges_test.go  --linkerd=(pwd)"/bin/linkerd" --k8s-context="$CTX" --integration-tests
```

### DIFF
--- a/test/edges/edges_test.go
+++ b/test/edges/edges_test.go
@@ -3,8 +3,10 @@ package edges
 import (
 	"bytes"
 	"html/template"
+	"io/ioutil"
 	"os"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/linkerd/linkerd2/testutil"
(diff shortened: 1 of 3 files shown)


test/edges/edges_test.go
```diff
@@ -45,3 +47,90 @@ func TestEdges(t *testing.T) {
 		t.Errorf("Expected output:\n%s\nactual:\n%s", b.String(), out)
 	}
 }
+
+// TestDirectEdges deploys a terminus and then generates a load generator which
+// sends traffic directly to the pod ip of the terminus pod. Traffic which is
+// addressed this way (as opposed to using the service name) does not show up
+// in the `linkerd edges` command. This test should be updated once
+// `linkerd edges` is updated to support this kind of traffic.
```
scottcarol: For my own understanding, this will likely be a short-lived comment as you are actively working on this, correct?

adleong: Laying the groundwork for it, yeah.  The feature is targeted for 2.7

PR #3623, opened 2019-10-23
build-cli-bin: POSIX compatible & handle whitespace paths

joakimr-axis:
Getting the directory where the script resides can easily be done
without bash-specific functionality, and hence the script can be POSIX
compatible. Also adding the missing pieces for handling paths with
whitespaces.

Change-Id: Ie2e867929be0322e476342438d9cf4a3d36f58f1

### DIFF
--- a/bin/build-cli-bin
+++ b/bin/build-cli-bin
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -eu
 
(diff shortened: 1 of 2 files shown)


joakimr-axis:
> bin/build-cli-bin: line 9: realpath: command not found

Oh, being part of coreutils on Linux systems I made a bad assumption (and not verifying with my non-Linux virtual machines) realpath (or rather coreutils) was sort of ubiquitous. Please let me get back with a non-realpath variant that is not bash-specific.

joakimr-axis:
> Please let me get back with a non-realpath variant that is not bash-specific.

Now there is an updated patchset that does not use realpath.

PR #3625, opened 2019-10-23
fetch-proxy: Make POSIX compatible

joakimr-axis:
Getting the directory where the script resides can easily be done
without bash-specific functionality, and hence the script can be POSIX
compatible.

Change-Id: I30bd69dccbc950bdce3dc5da4bea279305a7b1f9

### DIFF
--- a/bin/fetch-proxy
+++ b/bin/fetch-proxy
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # If the first argument to this script is "latest" or unset, it fetches the
-# latest proxy binary via build.l5d.io/linkerd2-proxy/latest.txt. If it's set to
+# latest proxy binary from the linkerd2-proxy github releases. If it's set to
 # a commit sha from the master branch of the linkerd2-proxy repo, it will fetch
 # the binary matching that sha instead.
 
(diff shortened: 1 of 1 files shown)


bin/fetch-proxy
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # If the first argument to this script is "latest" or unset, it fetches the
 # latest proxy binary via build.l5d.io/linkerd2-proxy/latest.txt. If it's set to
```
adleong: While you're updating this file, it looks like this comment is actually out of date.  Would you mind updating it while you're here?  This script no longer users `build.l5d.io` but instead fetches the latest proxy release from the linkerd2-proxy github releases.

joakimr-axis: No problem, I'll see to that.

joakimr-axis: Done.

bin/fetch-proxy
```diff
@@ -1,18 +1,18 @@
-#!/bin/bash
+#!/bin/sh
 
 # If the first argument to this script is "latest" or unset, it fetches the
-# latest proxy binary via build.l5d.io/linkerd2-proxy/latest.txt. If it's set to
+# latest proxy binary from the linkerd2-proxy github releases. If it's set to
 # a commit sha from the master branch of the linkerd2-proxy repo, it will fetch
```
adleong: I believe the first argument can be set to a proxy version number such as `v2.76.0` but not a commit sha.

joakimr-axis: Done!

Issue #3083, opened 2019-07-12
Dashboard: Missing DNS Rebinding attack protection


admc:
From our external security audit.

### What is the issue?

It was found that the Linkerd dashboard service, as well as the Grafana dashboard, are vulnerable to a DNS Rebinding attack. This allows an attacker to have a victim visiting an attacker-controlled website. The attacker will then be able to perform requests and retrieve responses on the victim’s behalf.

The attack exploits the fact that the dashboard does not restrict the Host header for incoming requests. By setting up a script on a domain with a very low TTL value and then quickly rebinding its IP address of localhost, the script will then “bypass” the restrictions normally set by SOP on the browser.

### How can it be reproduced?

- Run Linkerd dashboard service with port 3000 (linkerd dashboard --port 3000)
- Go to http://rebind.it:8080/manager.html and change the target port to 3000.
- Click on “Start Attack” and wait for a minute.
- The response from the dashboard will be returned, indicating success of a DNS rebinding attack

### Additional context

It is recommended to provide an option to fixate the dashboard domain. Grafana, for example, provides an enforce_domain option to ensure that redirection to the correct domain happens consistently.

olix0r:
It seems that this is complicated by our port-forwarding logic: clients may choose arbitrary ports for the api service. We must allow for an arbitrary number of ports, as clients may wish to connect to multiple cluster dashboards simultaneously.

If we wanted to hardcode a check in the web service, we could test that the host header matched `127.0.0.1:\d+`, etc; or, we could determine a fixed port range that we serve on (say, 100 ports) to limit the attack surface area.

But even if we do this, it's not going to be possible to configure grafana to participate in such a scheme, since its configuration only supports a single fixed hostname.

Generally, it seems that it would be a great feature for Linkerd to enforce in the mesh itself -- that pods only receive requests to a blessed set of service names. This would help to provide DNS rebinding protection at the infrastructure level! But, this is complicated by the way in which [kubectl port-forwards skip the linkerd proxy](https://github.com/linkerd/linkerd2/issues/2352), so our dashboard is typically exposed without the benefit an inbound proxy to enforce this sort of policy.

In summary, in order to remediate this -- whether in the mesh or via a frontend HTTP router for the control plane -- we're likely going to have to alter how the web service is exposed to clients via the command-line.

olix0r:
[...]

Issue #3100, opened 2019-07-17
linkerd dashboard should support --address


swgriffith:
## Feature Request
Add support for the --address flag for linkerd dashboard, as supported by kubectl port-forward and kubectl proxy. 

### What problem are you trying to solve?
An assumed binding to 127.0.0.1 doesn't always work. For example, if you're running from within Windows Subsystem for Linux V2 you're actually running within a micro VM on the host, and if you run linkerd dashboard from your WSL ubuntu (or other distro) you'll appear to be hosting the dashboard on 127.0.0.1, but that's actually the localhost of the micro vm not the host machine. In that case what you really wanted was for it to be exposed on the eth0 interface of the micro vm. 

### How should the problem be solved?
It appears that linkerd dashboard is just using port forwarding in k8s. If I run the following from by WSL2 Ubuntu environment I can reach the dashboard:

kubectl port-forward -n linkerd svc/linkerd-web --address 172.22.212.183 8080:8084

I'd recommend a minor update to the cmd to enable a --address parameter, consistent with kubectl proxy and kubectl port-forward.

What do you want to happen? Add any considered drawbacks.

### Any alternatives you've considered?

For me the alternative would be to avoid WSL, which isnt ideal from a productivity perspective, or write my own script to handle the port forwards manually rather than using the linkerd dashboard command.

### How would users interact with this feature?

linkerd dashboard --address 172.22.212.183

Linkerd dashboard available at:
http://172.22.212.183:50750
Grafana dashboard available at:
http://172.22.212.183:50750/grafana
Opening Linkerd dashboard in the default browser

grampelberg:
I’ve ran into this a couple times with my chromebook as well, I wish `port-forward` and `proxy` were easier to pass config options down to.

bmcstdio:
I've opened #3274 in an attempt to implement this.

PR #3629, opened 2019-10-23
Add DNS rebinding attack protection

alpeb:
## Changes to `linkerd dashboard`

Creates a new web server listening on http://$options.host:$options.port
(host and port can be set as CLI params) that forwards the connection to
the old k8s port-forwarder listening on http://$options.host:$targetPort
(targetPort is now fixed). Any request received with the Host header
different than $options.host:$options.port will be rejected, thus
preventing DNS rebinding attacks.
This local web server will also rewrite the Host header to something
like "linkerd-web.linkerd.svc.cluster.local:8084" that is now enforced
by the linkerd-web service.

Note that the other alternative mentioned in #3083 to have k8s'
port-forwarding do the header rewrites was explored, but the
port-forwarding code doesn't expose any APIs that we could leverage.

## Changes to `linkerd install`

- Attack prevention is enabled by default. A new option (visible in
`linkerd (install|upgrade) [control-plane]`)
`--disable-dns-rebinding-protection` (with its corresponding Helm value
`DisableDNSRebindingProtection`) where added.

## Changes to the `linkerd-web` service

- The service will reject any request not having a Host header like
`linkerd-web.linkerd.svc.cluster.local:8084`

## How to verify the attach protection works

- Run `linkerd dashboard --port 3000`
- Go to http://rebind.it:8080/manager.html and change the target port to
3000
- Click on “Start Attack” and wait for a minute.
- The response from the dashboard will be returned, showing an 'Invalid
Host header' message returned by the dashboard. If the attack would have
succeeded then the dashboard's html would be shown instead.

## Grafana

I didn't had to add the `enforce_domain` config in Grafana because when
using `linkerd dashboard` the requests for Grafana are proxied through
the `linkerd-web` service which is now protected. This means that, unlike
the dashboard, exposing Grafana directly doesn't require headers rewriting.
But while trying to expose Grafana I noticed we need to modify its
`root_url` config (even before this PR). Something to add to the docs.

## Additional Notes

- `linkerd dashboard` remains backwards compatible with older control
planes whose `linkerd-web` doesn't enforce the Host header.
- Why would one want to disable this protection? If you're exposing the
dashboard directly you won't be able to easily set up the Host header
accordingly. Also when exposing through Ingress (gotta update those docs
to explain the header-rewrites that would be needed).

Ref #3083


adleong:
Some open questions:

* Do we need to add this hostname validation to any other services?
* Does linkerd-web have any callers other than `linkerd dashboard` and potentially an ingress that the user has set up?
* Is it a problem that callers must use the fully qualified domain name?
* If we add `localhost` et al. to the linkerd-web hostname whitelist, do we still need the local web server?
  * Making the whitelist configurable works around the ingress issue nicely because users can simply add whatever explicit hostnames they need to the whitelist

adleong:
Somewhat relevant prior art: `kubectl proxy` supports this flag:

```
--accept-hosts='^localhost$,^127\.0\.0\.1$,^\[::1\]$': Regular expression for hosts that the proxy should accept.
```

linkerd-web could support something similar.

alpeb:
Thank you all for your feedback. Here are some answers. I think there's room for more discussion around this solution.

> My preference would be to remove --disable-dns-rebinding-protection and only allow disabling this through a helm install.

:+1:

> Not all ingress controllers allow host header rewriting, could we use l5d-dst-override as well?

Makes sense. So this means having the proxy do the Host header overriding, right?

> Do we need to add this hostname validation to any other services?

Like discussed offline, that would be good. Although It'd forbid accessing the other services directly through kubectl port-forwarding or proxy; the user could set up port-forwarding but have its http client rewrite the Host header just like we're doing here. But I don't know if there's a case for accessing those services directly.

> Does linkerd-web have any callers other than linkerd dashboard and potentially an ingress that the user has set up?

I can only think of tap when it inspects linkerd-web, but it's sending the proper Host header. Ingress would need to rewrite that header. The `l5d-dst-override` idea above should do the trick.

> Is it a problem that callers must use the fully qualified domain name?

Only if their http clients don't support header rewriting.

> If we add localhost et al. to the linkerd-web hostname whitelist, do we still need the local web server?
Making the whitelist configurable works around the ingress issue nicely because users can simply add whatever explicit hostnames they need to the whitelist

> Somewhat relevant prior art: kubectl proxy supports this flag:
--accept-hosts='^localhost$,^127\.0\.0\.1$,^\[::1\]$': Regular expression for hosts that the proxy should accept.
linkerd-web could support something similar.

We could. And if the user relies on something like `linkerd dashboard --address=1.2.3.4` (see #3100) then they'd have to add that address into that whitelist as well.

That said, I got convinced about the "properness" of enforcing a Host header that matches the pod identity. I think that warrants more discussion.

grampelberg:
> Makes sense. So this means having the proxy do the Host header overriding, right?

I think it means that we check both headers in `linkerd-web`. Looking at the XHR spec, you can add your own headers though, so it would only be slightly better than we have now for security.

Let's make sure there is a very clear error that comes back when the header is wrong, directs them to documentation and explains what is going on.

alpeb:
After some discussion offline, the point was raised that we should focus on just linkerd-web and the broader issue of tying this with identity can be tackled more generally later on in a separate issue. So the whitelist approach should be enough, without introducing the local webserver. I'll open a new PR that replaces this one.

I still have one question though. In order to still support `linkerd dashboard --address`, that whitelist needs to be configurable during install/upgrade. Would it be fine to surface this at both the `linkerd install|upgrade` and Helm levels?

grampelberg:
>  Would it be fine to surface this at both the linkerd install|upgrade and Helm levels?

WDYT about hard coding it for now? Folks who really care can use kustomize and I'm not convinced we need the added complexity yet.

alpeb:
Superseded by #3644

PR #3632, opened 2019-10-24
docker*: POSIX compatible and whitespace path handling

joakimr-axis:
Make docker scripts POSIX compatible and handle paths with whitespaces.

Change-Id: I58707a28d42723bc8782e1a8d72487ac4f3080c3


joakimr-axis:
Errors revealed in CI that was not triggered during local testing.

Issue #2899, opened 2019-06-05
Websocket problem when exposing linkerd dashboard through ambassador


pablomfc:
Dears,

I'm trying to expose the Linkerd (2.3.1) dashboard through  ambassador (0.70.1) using the configuration as follow:

  annotations:
    getambassador.io/config: |-
      ---
      apiVersion: ambassador/v1
      kind: Mapping
      name: linkerd-mapping
      host: "linkerd.example.net"
      prefix: /
      service: linkerd-web.linkerd.svc.cluster.local:8084
      add_request_headers:
        l5d-dst-override: linkerd-web.linkerd.svc.cluster.local:8084

Dashboard works fine, but not the tap feature, which presents web-socket errors.

I've checked some examples at https://linkerd.io/2/tasks/exposing-dashboard/ but didn't help me
to translate into ambassador rules.
I've looked also at  https://www.getambassador.io/user-guide/websockets-ambassador/

Is there a way to make it works with ambassador ?

Thanks for the great application ! 
Pablo Costa

--------------^ Click "Preview" for a better view!

We primarily use GitHub as an issue tracker. For any questions you may have,
please check out the resources below.

---

- Mailing List: https://lists.cncf.io/g/cncf-linkerd-users

- Slack: https://slack.linkerd.io

- Discourse: https://discourse.linkerd.io/

- Twitter: https://twitter.com/linkerd

grampelberg:
Are you running in AWS with ELBs? They don't support websockets, you'll need to change to ALBs.

pablomfc:
Thank you very much Thomas !

Your hint guided me to the main issue. I'm using ELB, but the problem was with ambassador configuration itself.

I've installed ambassador using the helm/chart. 

The right configuration for service.beta.kubernetes.io/aws-load-balancer-backend-protocol and use_proxy_proto must be "tcp" and "True" respectively, but It was set  as "http" and "False" 

Working  ambassador chart values.yaml:

`namespace:
  name: ambassador

service:

  https:
    enabled: true
    port: 443
    targetPort: 8080

  annotations:
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:xxxxxxxx
    service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
    service.beta.kubernetes.io/aws-load-balancer-internal:  0.0.0.0/0
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: "tcp"
[...]

PR #3633, opened 2019-10-24
Support Dashboard replicas (#2899)

KIVagant:
This PR makes possible to increase the amount of web dashboard replicas.

Follows up #2899

Signed-off-by: Eugene Glotov <kivagant@gmail.com>


go.sum
```diff
@@ -433,6 +433,7 @@ golang.org/x/tools v0.0.0-20190506145303-2d16b83fe98c/go.mod h1:RgjU9mgBXZiqYHBn
 golang.org/x/tools v0.0.0-20190524140312-2c0ae7006135/go.mod h1:RgjU9mgBXZiqYHBnxXauZ1Gv1EHHAz9KjViQ78xBX0Q=
 golang.org/x/tools v0.0.0-20191009213438-b090f1f24028 h1:uYpiJNFQR6EGzWevFLoFS3mfOVJdQce29o6eRxtYqYA=
 golang.org/x/tools v0.0.0-20191009213438-b090f1f24028/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
+golang.org/x/tools v0.0.0-20191024074452-7defa796fec0 h1:G/HlsVBRpvScWBPC2N5F3t6KORPLQD6n5k2nPsWt6DU=
```
KIVagant: No idea where the change came from

adleong:
Thanks for submitting this, @KIVagant!  I've got a few questions as I try to understand the issue:

Is sessionAffinity irrelevant when there is only one replica?  If so, should we just default the sessionAffinity to ClientIP all of the time?

In what types of scenarios would you need to change the sessionAffinity to something else?  Are these scenarios uncommon enough that we don't need to add a flag for configuring it? (users can always use something like kustomize if they really need to change it)

KIVagant:
With only one replica sessionAffinity changes really nothing because there's no real balancing. But yes, overall it should be enabled by default. I decided not to set it by default to ClientIP to not affect any existing installations. If not only me believe the sessionAffinity setting is useful for the dashboard service, I can enable it by default.

adleong:
Unless there's some downside to having sessionAffinity as ClientIP when there is only one replica, I would say let's make it the default.

KIVagant:
@adleong ,  sessionAffinity and sessionAffinityConfig are enabled by default.

grampelberg:
I don't think this is harmful, but I'm also pretty sure it won't actually fix anything. `port-forward` is always to a specific pod, so that would skip all this. For anything going through an ingress controller, they'll use endpoints and the stickiness would be required there.

adleong:
This is a good point.  Which ingress are you testing this with, @KIVagant?

KIVagant:
@grampelberg , it is very possible that I missed or confused something, there is a lot of settings involved. I'm using ALB Ingress Controller which talks to [oauth2_proxy](https://github.com/pusher/oauth2_proxy) which in its turn talks to Linkerd dashboard. The ALB Ingress Controller supports stickness options, for example `stickness.enabled=true,stinkyness.ib_cookie.duration_seconds=1200`. So it sticks to the oauth2_proxy pod(s) but I _feel_ like oauth2_proxy may have some problems if Linkerd dashboard has 2 or more pods but the service does not have the enabled sessionAffinity. I will try to make more tests.

grampelberg:
@KIVagant I hadn't even thought about `oauth2_proxy`! Mind sharing your config for that? I bet this'll fix that exact use case.

KIVagant:
The only setting they have is called [`-proxy-websockets`](https://pusher.github.io/oauth2_proxy/configuration) and I set it to `true`. Together with the stickness settings for ALB it works for one pod. But I'm not sure if it still the same for 2 and more pods (don't have the access to the cluster atm, will check tomorrow).

KIVagant:
I did some testing. Prerequisites: stickness enabled for the ALB. The dashboard scaled to 3 replicas.

Here's the results:

Pod A:
image

Pod B:
image

After `sessionAffinity` is enabled, external requests from `po/oauth2-proxy-linkerd` disappeared from `linkerd tap pod/linkerd-web-A --namespace linkerd` but continued reaching the Pod B. 

So from one side, stickness is working only between the ALB to the oauth2-proxy. From the other side, I don't see any errors in Linkerd dashboard when the sessionAffinity is disabled. Taking this into account I would say this PR is not important unless someone report about a problem.

KIVagant:
Btw I can extract the `Replicas` option to another PR.

grampelberg:
I don't mind it going in, but I'd like to remove all configuration for `SessionAffinity`. It'll probably just be easier to have a very small PR for changing the replicas count.

KIVagant:
Affinity removed

KIVagant:
> No spaces to match overall style, and the default is not really required

@alpeb , done

go.sum
kivagant-ba: Removed

KIVagant:
Thank you for merging the PR and helping me to analyze the use case. The result is tiny but for me it was a good example to practice with the repo and with Golang tests. Sorry for the noise produced by this.

grampelberg:
Thank you for the contribution! It was super helpful to think things through on our side as well =)

PR #3634, opened 2019-10-24
Fix whitespace path handling in docker (build) scripts

joakimr-axis:
Handling of whitespace paths was not fully implemented; this patch adds
the missing pieces. Also, only use bash where bash-specific
functionality is used/needed.

### DIFF
--- a/bin/_docker.sh
+++ b/bin/_docker.sh
@@ -1,20 +1,20 @@
 set -eu
 
-bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 
-. $bindir/_log.sh
+. "$bindir"/_log.sh
 
 # TODO this should be set to the canonical public docker registry; we can override this
 # docker regsistry in, for instance, CI.
-export DOCKER_REGISTRY="${DOCKER_REGISTRY:-gcr.io/linkerd-io}"
+export DOCKER_REGISTRY=${DOCKER_REGISTRY:-gcr.io/linkerd-io}
 
 # When set, causes docker's build output to be emitted to stderr.
-export DOCKER_TRACE="${DOCKER_TRACE:-}"
+export DOCKER_TRACE=${DOCKER_TRACE:-}
 
 docker_repo() {
-    repo="$1"
+    repo=$1
 
-    name="$repo"
+    name=$repo
     if [ -n "${DOCKER_REGISTRY:-}" ]; then
         name="$DOCKER_REGISTRY/$name"
     fi
@@ -26,23 +26,23 @@ docker_build() {
     repo=$(docker_repo "$1")
     shift
 
-    tag="$1"
+    tag=$1
     shift
 
-    file="$1"
+    file=$1
     shift
 
     extra="$@"
 
-    output="/dev/null"
+    output=/dev/null
     if [ -n "$DOCKER_TRACE" ]; then
-        output="/dev/stderr"
+        output=/dev/stderr
     fi
 
-    rootdir="$( cd $bindir/.. && pwd )"
+    rootdir=$( cd "$bindir"/.. && pwd )
 
     log_debug "  :; docker build $rootdir -t $repo:$tag -f $file $extra"
-    docker build $rootdir \
+    docker build "$rootdir" \
         -t "$repo:$tag" \
         -f "$file" \
         $extra \
@@ -53,22 +53,22 @@ docker_build() {
 
 docker_pull() {
     repo=$(docker_repo "$1")
-    tag="$2"
+    tag=$2
     log_debug "  :; docker pull $repo:$tag"
     docker pull "$repo:$tag"
 }
 
 docker_push() {
     repo=$(docker_repo "$1")
-    tag="$2"
+    tag=$2
     log_debug "  :; docker push $repo:$tag"
     docker push "$repo:$tag"
 }
 
 docker_retag() {
     repo=$(docker_repo "$1")
-    from="$2"
-    to="$3"
+    from=$2
+    to=$3
     log_debug "  :; docker tag $repo:$from $repo:$to"
     docker tag "$repo:$from" "$repo:$to"
     echo "$repo:$to"
--- a/bin/_docker.sh
+++ b/bin/_docker.sh
@@ -1,6 +1,7 @@
 set -eu
 
-bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
+#bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
+bindir=$( cd "${0%/*}" && pwd )
 
 . "$bindir"/_log.sh
 
--- a/bin/_docker.sh
+++ b/bin/_docker.sh
@@ -1,7 +1,6 @@
 set -eu
 
-#bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
-bindir=$( cd "${0%/*}" && pwd )
+bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 
 . "$bindir"/_log.sh
 
--- a/bin/_tag.sh
+++ b/bin/_tag.sh
@@ -52,7 +52,7 @@ validate_tag() {
     sha=$1
     shift
 
-    dockerfile_tag=$(grep -oe $image':[^ ]*' "$file") || true
+    dockerfile_tag=$(grep -oe "$image"':[^ ]*' "$file") || true
     deps_tag="$image:$sha"
     if [ -n "$dockerfile_tag" ] && [ "$dockerfile_tag" != "$deps_tag" ]; then
         echo "Tag in "$file" does not match source tree:"
--- a/bin/_tag.sh
+++ b/bin/_tag.sh
@@ -7,9 +7,9 @@ git_sha_head() {
 }
 
 go_deps_sha() {
-    bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-    rootdir="$( cd $bindir/.. && pwd )"
-    cat $rootdir/go.mod $rootdir/Dockerfile-go-deps | shasum - | awk '{print $1}' |cut -c 1-8
+    bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
+    rootdir=$( cd "$bindir"/.. && pwd )
+    cat "$rootdir/go.mod" "$rootdir/Dockerfile-go-deps" | shasum - | awk '{print $1}' |cut -c 1-8
 }
 
 clean_head() {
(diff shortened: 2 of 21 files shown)


bin/_tag.sh
```diff
@@ -24,37 +24,37 @@ head_root_tag() {
     if clean_head ; then
         clean_head_root_tag
     else
-        name="$(echo $USER | sed "s/[^[:alnum:].-]//g")"
+        name=$(echo $USER | sed 's/[^[:alnum:].-]//g')
```
adleong: shellcheck suggests this cool syntax here:

```
name=${USER//[^[:alnum:].-]/}
```

bin/_tag.sh
```diff
@@ -24,37 +24,37 @@ head_root_tag() {
...
-    image="$1"
+    image=$1
     shift
 
-    sha="$1"
+    sha=$1
     shift
 
-    dockerfile_tag=$(grep -oe $image':[^ ]*' $file) || true
+    dockerfile_tag=$(grep -oe $image':[^ ]*' "$file") || true
```
adleong: should $image be double quoted here?

bin/docker-build
```diff
@@ -1,22 +1,22 @@
-#!/bin/bash
+#!/bin/sh
 
 set -eu
 
 if [ $# -ne 0 ]; then
-    echo "no arguments allowed for $(basename $0), given: $@" >&2
+    echo "no arguments allowed for ${0##*/}, given: $@" >&2
```
adleong: I think this should be `$*` instead of `$@`

bin/docker-build-base
```diff
@@ -5,19 +5,19 @@
 set -eu
 
 if [ $# -ne 0 ]; then
-    echo "no arguments allowed for $(basename $0), given: $@" >&2
+    echo "no arguments allowed for ${0##*/}, given: $@" >&2
```
adleong: same thing here `$*`

bin/docker-build-base
```diff
@@ -5,19 +5,19 @@
...
     exit 64
 fi
 
-bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-rootdir="$( cd $bindir/.. && pwd )"
+bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
```
adleong: just curious, are there things in this file that rely on bash?

bin/docker-build-cli-bin
```diff
@@ -3,26 +3,26 @@
 set -eu
 
 if [ $# -ne 0 ]; then
-    echo "no arguments allowed for $(basename $0), given: $@" >&2
+    echo "no arguments allowed for ${0##*/}, given: $@" >&2
```
adleong: ditto

bin/docker-build-cli-bin
```diff
@@ -3,26 +3,26 @@
...
 
 (
-    $bindir/docker-build-go-deps
+    "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-tag="$(head_root_tag)"
-docker_build cli-bin $tag $dockerfile --build-arg LINKERD_VERSION=$tag
+tag=$(head_root_tag)
+docker_build cli-bin $tag "$dockerfile" --build-arg LINKERD_VERSION=$tag
```
adleong: should $tag be double quoted here?

bin/docker-build-cni-plugin
```diff
@@ -3,17 +3,17 @@
 set -eu
 
 if [ $# -ne 0 ]; then
-    echo "no arguments allowed for $(basename "$0"), given: $*" >&2
+    echo "no arguments allowed for ${0##*/}, given: $@" >&2
```
adleong: ditto

bin/docker-build-cni-plugin
```diff
@@ -22,4 +22,4 @@ validate_go_deps_tag "$dockerfile"
     "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-docker_build cni-plugin "$(head_root_tag)" "$dockerfile"
+docker_build cni-plugin $(head_root_tag) "$dockerfile"
```
adleong: why remove the quotes from head_root_tag?

bin/docker-build-controller
```diff
@@ -3,23 +3,23 @@
...
 
 (
-    $bindir/docker-build-go-deps
+    "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-tag="$(head_root_tag)"
-docker_build controller $tag $dockerfile --build-arg LINKERD_VERSION=$tag
+tag=$(head_root_tag)
+docker_build controller $tag "$dockerfile" --build-arg LINKERD_VERSION=$tag
```
adleong: does $tag need to be double quoted?

bin/docker-build-debug
```diff
@@ -3,22 +3,20 @@
...
+dockerfile="$rootdir"/Dockerfile-debug
 
 (
     "$bindir"/docker-build-base
 ) >/dev/null
 
-tag="$(head_root_tag)"
-
-docker_build debug $tag $dockerfile
+docker_build debug $(head_root_tag) "$dockerfile"
```
adleong: why remove the quotes from head_root_tag?

bin/docker-build-grafana
```diff
@@ -3,16 +3,16 @@
...
-. $bindir/_docker.sh
-. $bindir/_tag.sh
+. "$bindir"/_docker.sh
+. "$bindir"/_tag.sh
 
-dockerfile=$rootdir/grafana/Dockerfile
+dockerfile="$rootdir/grafana/Dockerfile"
 
-docker_build grafana "$(head_root_tag)" $dockerfile
+docker_build grafana $(head_root_tag) "$dockerfile"
```
adleong: ditto

bin/docker-pull-binaries
```diff
@@ -3,32 +3,30 @@
...
 
 workdir="$rootdir/target/release"
 mkdir -p "$workdir"
 
-shorttag="$tag"
-if [ "${tag:0:1}" == "v" ]; then
-  shorttag="${tag:1}"
-fi
+# create shorttag where eventual initial v in tag is stripped
+shorttag=${tag#v}
```
adleong: cool

bin/_tag.sh
joakimr-axis: The only place where I see validate_tag being called is on line 72 in _tag.sh, where the value is gcr.io/linkerd-io/go-deps. Besides, I believe whitespaces are not allowed in that context, is it? That's why I figured $image would not be expected to contain whitespaces (=no quotes necessary).

bin/docker-build-cni-plugin
joakimr-axis: If you look at the head_root_tag function, it will never return whitespace (and if it did, it would not be a valid tag for use with docker build at line 45 in _docker.sh where it would end up).

bin/docker-build-controller
joakimr-axis: Hmm ... besides whitespaces, are there any allowed characters in tags that would in fact need quotes not to interfere with the shell?

bin/docker-build-debug
joakimr-axis: Here the behavior is the very same as it was. At assignment, the double quotes are redundant since
test=$(echo foo bar)
and
test="$(echo foo bar)"
will both end up with test having the value 'foo bar'.
And since tag was used only in one place, I found it unnecessary to have it in a separate variable and replace the unquoted $tag on the docker_build line with $(head_root_tag). So it does exactly the same thing as before, but with fewer lines and characters.
Surely, that leaves the question on if the tag should be quoted or not, and I believe the tag is not allowed to contain any characters that would interfere with the shell unquoted anyway?

bin/docker-pull-binaries
joakimr-axis: POSIX builtins FTW! :)

bin/_tag.sh
adleong: I think you're right, but it seems like a good practice/standard to always quote variables in this type of scenario to be safe.  Is there any reason not to?

To be perfectly honest, I don't know very much about shell scripting and I lean *very* heavily on https://www.shellcheck.net/ for best practices.  So even if it doesn't make a difference in this case, I feel a lot more comfortable if I can make shellcheck happy.

joakimr-axis: I guess it is sort of a philosophical question; one side prefers fail-early principles and do not quote where content that should not be quote is, while the other side says pass on whatever in a safe manner and leave it for the end consumer to decide what is ok an what is not. To me, as long as there is a clear understanding of why and why not something was quoted, I see benefits with any of those approaches.
I will update the patchset with your suggestion!

bin/_docker.sh
```diff
@@ -1,20 +1,21 @@
 set -eu
 
-bindir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+#bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
```
adleong: did you mean to remove this commented out line?

bin/docker-build-cli-bin
```diff
@@ -3,26 +3,26 @@
...
 
 (
-    $bindir/docker-build-go-deps
+    "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-tag="$(head_root_tag)"
-docker_build cli-bin $tag $dockerfile --build-arg LINKERD_VERSION=$tag
+tag=$(head_root_tag)
+docker_build cli-bin "$tag" "$dockerfile" --build-arg LINKERD_VERSION=$tag
```
adleong: the $tag variable appears twice in this line and I think both should be quoted

bin/docker-build-controller
```diff
@@ -3,23 +3,23 @@
...
 
 (
-    $bindir/docker-build-go-deps
+    "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-tag="$(head_root_tag)"
-docker_build controller $tag $dockerfile --build-arg LINKERD_VERSION=$tag
+tag=$(head_root_tag)
+docker_build controller "$tag" "$dockerfile" --build-arg LINKERD_VERSION=$tag
```
adleong: ditto

bin/docker-build-proxy
```diff
@@ -3,24 +3,23 @@
...
 (
-    $bindir/docker-build-go-deps
+    "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-tag="$(head_root_tag)"
-docker_build proxy "$tag" "$dockerfile" \
-  --build-arg "LINKERD_VERSION=$tag"
+tag=$(head_root_tag)
+docker_build proxy "$tag" "$dockerfile" --build-arg LINKERD_VERSION=$tag
```
adleong: ditto

bin/docker-build-web
```diff
@@ -3,23 +3,23 @@
...
 
 (
-    $bindir/docker-build-go-deps
+    "$bindir"/docker-build-go-deps
 ) >/dev/null
 
-tag="$(head_root_tag)"
-docker_build web $tag $dockerfile --build-arg LINKERD_VERSION=$tag
+tag=$(head_root_tag)
+docker_build web "$tag" "$dockerfile" --build-arg LINKERD_VERSION=$tag
```
adleong: ditto

bin/docker-images
```diff
@@ -2,22 +2,22 @@
...
         --format "{{printf \"%-16s %-10s\" \"$1\" \"${2}\"}} {{.Size | printf \"%6s\"}}  {{.ID}}  {{.CreatedAt}}" \
         "${repo}:${2}"
 }
 
 tag=$(head_root_tag)
 
-for img in cli-bin cni-plugin controller debug grafana proxy web  ; do
-    docker_image "$img" "$tag"
+for img in cli-bin cni-plugin controller debug grafana proxy web ; do
+    docker_image $img $tag
```
adleong: ditto

bin/_docker.sh
joakimr-axis: No, I did not, thanks for observing it. Funny enough, my test builds as well as the CI did work (since the bindir is set in the scripts that source _docker.sh). But it was not intended to be commented out. I will fix this.

PR #3641, opened 2019-10-25
Check stats of deployed apps before and after linkerd upgrade

### LINKED ISSUE #3544: Add integration test to ensure backward compatible upgrade
Following https://github.com/linkerd/linkerd2/issues/3537, we should update the `upgrade` integration test to check the application's health and success rates, to ensure that the newer version of the control plane doesn't break the data plane. To ensure zero downtime upgrade, the application should remain healthy before and after its restart.

zaharidichev:
This PR adds a stat checks to the upgrade integration test to ensure the deployed apps did not break. This is done by looking at the SuccessRate. 

Fixes #3544 

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -151,58 +151,39 @@ func TestCheckPreInstall(t *testing.T) {
 	}
 }
 
-// Gathers success stats from the test app
-func ensureTestAppRunning(t *testing.T, namespace string) {
-	args := []string{"stat", "deploy", "-n", namespace, "-t", "60s"}
-	err := TestHelper.RetryFor(60*time.Second, func() error {
-		out, stderr, err := TestHelper.LinkerdRun(args...)
-		if err != nil {
-			fmt.Println(stderr)
-			t.Fatalf("Unexpected stat error: %s\n%s", err, out)
-		}
-
-		rowStats, err := testutil.ParseRows(out, 1, 8)
+func exerciseTestAppEndpoint(endpoint, namespace string) error {
+	testAppURL, err := TestHelper.URLFor(namespace, "web", 8080)
+	if err != nil {
+		return err
+	}
+	for i := 0; i < 30; i++ {
+		_, err := TestHelper.HTTPGetURL(testAppURL + endpoint)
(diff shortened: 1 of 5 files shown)


test/install_test.go
```diff
@@ -151,6 +151,64 @@ func TestCheckPreInstall(t *testing.T) {
 	}
 }
 
+// Gathers success stats from the apps installed in the test-tap namespace
+// Used to ensure upgrade process went smoothly
+func ensureTAppsRunning(t *testing.T) {
+	args := []string{"stat", "deploy", "-n", TestHelper.GetTestNamespace("tap-test"), "-t", "60s"}
+	err := TestHelper.RetryFor(20*time.Second, func() error {
```
adleong: If I understand this correctly, a 20 second timeout seems super aggressive.  It will take some time or the app to come up and to start processing traffic.  Plus there's up to a 10 second delay from the prometheus scrape interval.

zaharidichev: @adleong That's very useful context. What would you consider a more suitable value?

adleong: Not sure. I'd recommend experimenting to see how long it takes for these stats to become available.

zaharidichev: I see now that this is why my integration test is failing with **No traffic found**

alpeb:
@zaharidichev I tried having the new test fail in the scenario described in #3537 but using your new `upgrade_test.yaml`: I installed 2.5.0 stable, installed that app, upgraded to edge-19.10.1, and then checked its stats, which showed fine.
I _think_ the proxy caches the Destination queries, so after the upgrade no newer queries are performed and the failure doesn't occur. The app would have to perform a request to a different destination to trigger that.
Let me know what you think.

zaharidichev:
@alpeb I changed things up a bit. Now we are injecting the app via a shell script so it is injected with the version of linkerd we are upgrading to. Also changed the app and how we test it. We only rely on hitting its endpoints now. The results are: 

2.6.0 -> HEAD - works
2.5.0 -> edge-19.10.1 - breaks because destination service is not reachable anymore

PR #3642, opened 2019-10-25
Move table check output to stderr

memory:
Assertion: friendly human-readable output with clever animations is
never meant to be consumed by pipelines, and will cause unwanted
behavior when it does.  

Once merged, this PR will send all spinners etc to stderr; if people
want a pipeline they can (and will) use the JSON output option.

Signed-off-by: Nathan Mehl <memory@blank.org>

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -191,22 +191,22 @@ func runChecks(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker, ou
 	if output == jsonOutput {
 		return runChecksJSON(wout, werr, hc)
 	}
-	return runChecksTable(wout, hc)
+	return runChecksTable(werr, hc)
 }
 
-func runChecksTable(wout io.Writer, hc *healthcheck.HealthChecker) bool {
+func runChecksTable(werr io.Writer, hc *healthcheck.HealthChecker) bool {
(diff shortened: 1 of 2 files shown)


adleong:
Hey @memory!  Can you tell me a bit more about the motivation for this change? You are correct that using JSON output option is appropriate when using pipelines, so what's wrong with using stdout for the animations when in non-JSON mode?

memory:
I chatted about this a bit with @grampelberg over slack today.  The 5s summary is that I had built myself a kustomize generator that ran (essentially) `linkerd check && linkerd install` and kustomize spat blood in a most unhelpful manner because it was trying to parse `check`'s output as a kubernetes resource definition.  (And even better it was completely swallowing the output so it took forever to find it.)

This is probably a case of "doctor it hurts when I do this!" and if that's the consensus I'm fine with it.  But kustomize is currently the official solution for automating customized linkerd installations and it's a <i>bear</i> to debug when it gets unexpected input so here we are. :)

grampelberg:
@memory I think I'm of the opinion to close this one out. I really don't want to subvert what `stderr` is actually supposed to suggest and it isn't hard to do an output redirect in the integration.

That said, we definitely need to make this customization easier for you on the prometheus side.

PR #3644, opened 2019-10-28
DNS rebinding protection for the dashboard

### LINKED ISSUE #3083: Dashboard: Missing DNS Rebinding attack protection

alpeb:
Fixes #3083 and replacement for #3629

This adds a new parameter to the `linkerd-web` container `enforcedHost`
that establishes the regexp that the Host header must enforce, otherwise
it returns an error.

This parameter will be hard-coded for now, in `linkerd-web`'s deployment
yaml.

Note this also protects the dashboard because that's proxied from
`linkerd-web`.

Also note this means the usage of `linkerd dashboard --address` will
require the user to change that parameter in the deployment yaml (or
have Kustomize do it).

## How to verify the attach protection works

- Run `linkerd dashboard`
- Go to http://rebind.it:8080/manager.html and change the target port to
50750
- Click on “Start Attack” and wait for a minute.
- The response from the dashboard will be returned, showing an 'Invalid
Host header' message returned by the dashboard. If the attack would have
succeeded then the dashboard's html would be shown instead.

### DIFF
--- a/charts/linkerd2/templates/web.yaml
+++ b/charts/linkerd2/templates/web.yaml
@@ -64,7 +64,7 @@ spec:
         - -controller-namespace={{.Namespace}}
         - -log-level={{.ControllerLogLevel}}
         {{- $host := replace "." "\\." (printf "linkerd-web.%s.svc.%s" .Namespace .ClusterDomain) }}
-        - -enforced-host=^(localhost|127\.0\.0\.1|{{ $host }}|\[::1\]):\d+$
+        - -enforced-host=^(localhost|127\.0\.0\.1|{{ $host }}|\[::1\])(:\d+)?$
         {{- include "partials.linkerd.trace" . | nindent 8 -}}
         image: {{.WebImage}}:{{default .LinkerdVersion .ControllerImageVersion}}
         imagePullPolicy: {{.ImagePullPolicy}}
--- a/charts/linkerd2/templates/web.yaml
+++ b/charts/linkerd2/templates/web.yaml
@@ -63,7 +63,8 @@ spec:
         - -grafana-addr=linkerd-grafana.{{.Namespace}}.svc.{{.ClusterDomain}}:3000
         - -controller-namespace={{.Namespace}}
         - -log-level={{.ControllerLogLevel}}
-        - -enforced-host=^(localhost|127\.0\.0\.1|\[::1\]):\d+$
+        {{- $host := replace "." "\\." (printf "linkerd-web.%s.svc.%s" .Namespace .ClusterDomain) }}
+        - -enforced-host=^(localhost|127\.0\.0\.1|{{ $host }}|\[::1\]):\d+$
         {{- include "partials.linkerd.trace" . | nindent 8 -}}
         image: {{.WebImage}}:{{default .LinkerdVersion .ControllerImageVersion}}
         imagePullPolicy: {{.ImagePullPolicy}}
--- a/charts/linkerd2/templates/web.yaml
+++ b/charts/linkerd2/templates/web.yaml
@@ -63,6 +63,7 @@ spec:
         - -grafana-addr=linkerd-grafana.{{.Namespace}}.svc.{{.ClusterDomain}}:3000
         - -controller-namespace={{.Namespace}}
         - -log-level={{.ControllerLogLevel}}
+        - -enforced-host=^(localhost|127\.0\.0\.1|\[::1\]):\d+$
         {{- include "partials.linkerd.trace" . | nindent 8 -}}
         image: {{.WebImage}}:{{default .LinkerdVersion .ControllerImageVersion}}
         imagePullPolicy: {{.ImagePullPolicy}}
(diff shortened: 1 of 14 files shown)


alpeb:
Good call @adleong :+1:  I've just pushed a fix.

web/srv/server.go
```diff
@@ -44,6 +46,10 @@ type (
 
 // this is called by the HTTP server to actually respond to a request
 func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if !s.reHost.MatchString(req.Host) {
+		http.Error(w, "Invalid Host header", http.StatusNotFound)
```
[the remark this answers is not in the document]
alpeb: @grampelberg where should we put that doc? As a section of https://linkerd.io/2/reference/cli/dashboard/ or as a new page under reference or features?

alpeb:
I've just updated the error message, and created linkerd/website#571 to track the doc changes :+1:

web/srv/server.go
```diff
@@ -44,6 +48,15 @@ type (
 
 // this is called by the HTTP server to actually respond to a request
 func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if !s.reHost.MatchString(req.Host) {
+		error := fmt.Sprintf(`It appears that you are trying to reach this service with a host of '%s'.
+This does not match /^(localhost|127\.0\.0\.1|linkerd-web\.linkerd\.svc\.cluster\.local|\[::1\]):\d+$/ and has been denied for security reasons.
+Please see https://linkerd.io/dns-rebinding for an explanation of what is happening and how to fix it.`,
+			html.EscapeString(req.Host))
+
+		http.Error(w, error, http.StatusNotFound)
```
adleong: I think status code 400 might be more appropriate since this seems like it would fall under "deceptive request routing"

web/srv/server.go
```diff
@@ -44,6 +48,15 @@ type (
 
 // this is called by the HTTP server to actually respond to a request
 func (s *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if !s.reHost.MatchString(req.Host) {
+		error := fmt.Sprintf(`It appears that you are trying to reach this service with a host of '%s'.
+This does not match /^(localhost|127\.0\.0\.1|linkerd-web\.linkerd\.svc\.cluster\.local|\[::1\]):\d+$/ and has been denied for security reasons.
```
adleong: Should this message include the actual regex instead of the default one?

charts/linkerd2/templates/web.yaml
```diff
@@ -63,6 +63,8 @@ spec:
         - -grafana-addr=linkerd-grafana.{{.Namespace}}.svc.{{.ClusterDomain}}:3000
         - -controller-namespace={{.Namespace}}
         - -log-level={{.ControllerLogLevel}}
+        {{- $host := replace "." "\\." (printf "linkerd-web.%s.svc.%s" .Namespace .ClusterDomain) }}
+        - -enforced-host=^(localhost|127\.0\.0\.1|{{ $host }}|\[::1\]):\d+$
```
adleong: should the port be optional in this regex?

web/main.go
```diff
@@ -73,7 +75,13 @@ func main() {
 		}
 	}
 
-	server := srv.NewServer(*addr, *grafanaAddr, *templateDir, *staticDir, uuid, *controllerNamespace, clusterDomain, *reload, client, k8sAPI)
+	reHost, err := regexp.Compile(*enforcedHost)
```
adleong: is the empty string a valid regex?  does it match everything?  I would expect omitting the --enforced-host flag to effectively disable this protection.

charts/linkerd2/templates/web.yaml
alpeb: Indeed!

web/main.go
alpeb: Yes the empty string matches everything. So omitting `--enforced-host` makes `enforcedHost` default to the empty string, which allows any `Host` header.

web/srv/server.go
alpeb: Yes, good catch :+1:

PR #3646, opened 2019-10-28
Add tests for dashboard's new namespace-focused navigation

scottcarol:
This PR adds two tests to test the dashboard's new navigation and routing patterns:
`/web/app/integration/specs/url-routing.js`
`/web/app/integration/specs/namespace-select.js`

It deletes the now-obsolete `logo-redirect.js` test.

To make testing easier, this PR adds an `id` to some of the elements rendered in the browser. 

It also updates the `webdriverio` packages we rely on and adds the `chai` assertion library.

You must have a working version of the dashboard running on `localhost:7777` (this opens with `bin/web dev`). The location of the dashboard can be changed in `wdio.conf.js`

This PR does not change any functionality of the dashboard. 

**Note** Our SauceLabs tunnel connection documented in `TEST.md` is not working (`bin/web integration cloud`), I'm making a separate ticket for this. At this time, these tests should be run with `bin/web integration local`

To test:
You need to have Chrome installed, no specific version. From `/linkerd2`:
```
$ bin/web setup # download updated dependencies
$ bin/web dev # start the dashboard
$ bin/web integration local # webdriverio will open Chrome and click through the tests.
```


admc:
Delivered @scottcarol some feedback about trying to avoid using ID's as lookup selectors (as it's hard to be sure that only one copy of a component exists in the DOM), and to try to use waitFor type commands instead of pauses when possible.

scottcarol:
Got great feedback from @admc on improving tests, once I incorporate the changes I will push the updated branch and it will be ready for review again.

scottcarol:
Just pushed the updated branch which uses the `react$$` API instead of ID selectors and `waitUntil()` instead of browser pauses. Ready for review.

web/app/integration/specs/namespace-select.js
```diff
@@ -0,0 +1,114 @@
+const assert = require('assert');
```
dadjeibaah: Hmm, I am not quite sure how this ends up using the chai assertion library we imported in `package.json`. Is there some underlying code that does some magic for these tests to use chai under the hood?

scottcarol: Thanks for the review! Chai is the assertion library that is used in conjunction with Mocha. We initialize it in the [before hook of wdio.conf.js](https://webdriver.io/docs/frameworks.html). ⭐️

Issue #3271, opened 2019-08-15
traffic split integration test


siggy:
Our integration tests should deploy a TrafficSplit resource and execute some variants of `linkerd stat trafficsplits`.

Issue #3562, opened 2019-10-11
Traffic split shows no stats when filtered by namespace


stefanprodan:
### What is the issue?
Traffic split shows no metrics in dashboard or with `linkerd stat ts -n some-ns` while `linkerd stat ts --all-namespaces` does.

### How can it be reproduced?

Create a traffic split and generate some traffic.

### Logs, error output, etc

<img width="1173" alt="Screenshot 2019-10-11 at 14 58 37" src="https://user-images.githubusercontent.com/3797675/66649824-cab92b00-ec37-11e9-8e87-63d5fce8b084.png">

```
$ linkerd stat ts -n test-linkerd
NAME      APEX      LEAF              WEIGHT   SUCCESS   RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
podinfo   podinfo   podinfo-canary         0         -     -             -             -             -
podinfo   podinfo   podinfo-primary      100         -     -             -             -             -
```

```
$ linkerd stat ts --all-namespaces
NAMESPACE      NAME       APEX       LEAF               WEIGHT   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
dev            backend    backend    backend-canary          0         -        -             -             -             -
dev            backend    backend    backend-primary       100   100.00%   1.0rps           3ms          15ms          19ms
dev            cache      cache      cache-canary            0         -        -             -             -             -
dev            cache      cache      cache-primary         100   100.00%   1.0rps           2ms          15ms          19ms
dev            database   database   database-canary         0         -        -             -             -             -
dev            database   database   database-primary      100   100.00%   1.0rps           1ms           2ms           3ms
dev            frontend   frontend   frontend-canary         0         -        -             -             -             -
dev            frontend   frontend   frontend-primary      100   100.00%   1.0rps           5ms          25ms          29ms
test-linkerd   podinfo    podinfo    podinfo-canary          0         -        -             -             -             -
test-linkerd   podinfo    podinfo    podinfo-primary       100   100.00%   0.3rps           1ms           2ms           2ms
```

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version
√ is running the minimum kubectl version

linkerd-config
--------------
√ control plane Namespace exists
√ control plane ClusterRoles exist
√ control plane ClusterRoleBindings exist
[...]

PR #3649, opened 2019-10-29
Traffic split integration test

zaharidichev:
This pr adds an integration test that ensures the CLI stat ts command reflects properly the traffic routing done by the traffic split resource. 

Fixes: #3271


scottcarol:
@zaharidichev, what's the best way to test this PR? 
`$ bin/docker-build`
`$ go test -v ./test -integration-tests -linkerd /path/to/linkerd`
then run this new test?

zaharidichev:
@scottcarol So you need to have a cluster that has the images uploaded on it (the ones you have built with `bin/docker-build`). Then you do: 

`go test -v ./test -integration-tests -linkerd /path/to/linkerd` - that runs the install tests and gets linkerd on to the cluster

`go test -v ./test/trafficsplit -integration-tests  -linkerd /path/to/linkerd` - that should run the trafficsplit integration tests that are in this PR. 

They are already run by the CI though. Hope that helps :)

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	if actual.weight != expected.weight && expected.weight != "" {
+		return fmt.Errorf("expected weight '%s' for leaf %s, got '%s'", expected.weight, expected.leaf, actual.weight)
+	}
+
+	if actual.success != expected.success && expected.success != "" {
+		return fmt.Errorf("expected success '%s' for leaf %s, got '%s'", expected.success, expected.leaf, actual.success)
+	}
+
+	if actual.rps != expected.rps && expected.rps != "" {
```
adleong: rps will never match exactly, but we can make sure that if expected rps is 0 then actual rps must be 0 and if expected rps is non-zero then actual rps must be non-zero.

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,247 @@
...
+		tsResourceFile := "testdata/traffic_splitter.yaml"
+		tsResource, err := testutil.ReadFile(tsResourceFile)
+
+		if err != nil {
+			t.Fatalf("Cannot read traffic split resource: %s, %s", tsResourceFile, err)
+		}
+
+		out, err := TestHelper.KubectlApply(tsResource, prefixedNs)
+		if err != nil {
+			t.Fatalf("Failed to create traffic split resource: %s\n %s", err, out)
```
scottcarol: see above comment

test/trafficsplit/trafficsplit_test.go
scottcarol: I agree with @adleong that it would be helpful to have some checks around RPS being non-zero. 

Also, although we can't match RPS, we are able to calculate the relative percentages of RPS expected, from the weights. Could we compare the RPS of each leaf? If there is a `500m/500m` split we could divide `leaf1` by `leaf2` and check to see that the result was between `9.9` and `1.1`. Would this be practical, or too vulnerable to flakiness?

If the above idea is practical, we could test additional traffic ratios, such as `200m/800m`.

The reason I'm thinking about this is because of the conversation in #3562 - what additional checks can we add to this test to reflect an actual traffic split scenario?

scottcarol:
When I run this test multiple times, after the first time I see:
```
FAIL: TestTrafficSplitCli/ensure_traffic_is_sent_to_one_backend_only (40.18s)
        trafficsplit_test.go:202: expected success '-' for leaf backend-svc, got '0.00%'
```
I believe this is because old Prometheus stats are still available, since the traffic split application has not been deleted. Is it desirable to delete the traffic split application at the end of the test, so that you can run it multiple times in a row without running `bin/test-cleanup`? Or is it expected that you'd only be able to run it once?

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,247 @@
...
+			}
+			return nil
+		})
+
+		if err != nil {
+			t.Fatal(err.Error())
+		}
+	})
+
+	t.Run("create traffic split resource", func(t *testing.T) {
```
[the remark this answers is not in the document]
zaharidichev: @scottcarol Nice catch. This is a left over. For a bit of context of how I started this. Initially I was not creating a traffic split resource in the main yaml and was just creating an equally weighted after that as a separate step, ensuring that this is later reflected by the change in traffic. Decided that modification of weights is a better test scenario though. Will change the comments and file names :)

test/trafficsplit/trafficsplit_test.go
zaharidichev: @scottcarol I think although this is a good idea, it probably is trying to be a bit too clever and might introduce flakiness. The purpose of the test as its written is to ensure that there is traffic flowing to both services. But since one of them is an always failing one, it might very well be the case that its RPS  might be a bit different (just a guess). But to illustrate this with an example. After I modify the split resource weights to be equal, this is what I get as an output at some point: 

NAME                    APEX          LEAF          WEIGHT   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
```
backend-traffic-split   backend-svc   backend-svc     500m   100.00%   0.4rps           8ms          17ms          20ms
backend-traffic-split   backend-svc   failing-svc     500m     0.00%   0.2rps           6ms          27ms          30ms
```

zaharidichev:
@scottcarol Wrt to deleting the app that can easily be done by simply deleting the whole `l5d-integration-trafficsplit-test` namespace. However I do not  really see this pattern in the rest of the tests. I think they are supposed to be ran just once. Adding a cleanup step to the test itself will most likely prolong the runtime. Not sure its worth it, but might be  wrong. What do you think @adleong

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,262 @@
...
+		if expected.rps != ZeroRps && actual.rps == ZeroRps {
+			return fmt.Errorf("expected rps '%s' for leaf %s, got '0.0rps'", expected.rps, expected.leaf)
+		}
+	} else {
+		if actual.rps != expected.rps && expected.rps != "" {
+			return fmt.Errorf("expected rps '%s' for leaf %s, got '%s'", expected.rps, expected.leaf, actual.rps)
+		}
+	}
+
+	if actual.latencyP50 != expected.latencyP50 && expected.latencyP50 != "" {
```
adleong: we never compare latencies so I think we can remove this

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if actual.weight != expected.weight && expected.weight != "" {
+		return fmt.Errorf("expected weight '%s' for leaf %s, got '%s'", expected.weight, expected.leaf, actual.weight)
+	}
+
+	if actual.success != expected.success && expected.success != "" {
+		return fmt.Errorf("expected success '%s' for leaf %s, got '%s'", expected.success, expected.leaf, actual.success)
+	}
+
+	if expected.rps != "-" {
```
adleong: I'm having trouble following the nested ifs with non-equals.  I think this is saying...

if expected.rps is "" then pass
if expected.rps is "-" then actual must be "-"
if expected.rps is "0.0rps" then actual must be "0.0rps"
if expected.rps is a non-zero number, then actual must be a non-zero number

Is that right?  Can these be re-structured to be more straightforward?

test/trafficsplit/trafficsplit_test.go
scottcarol: Yes, it's probably not practical to compare theoretical/real RPS. If it was a priority in the future to measure RPS changes, we *could* calculate the general trend in RPS over time and report whether each leaf's trend was positive or negative when switching from `200m/300m` to `500m/0m`, for example. This would be slightly more specific than measuring for zeroRPS. Completely unnecessary for this test and PR though!

test/trafficsplit/trafficsplit_test.go
zaharidichev: They are there so we can ensure that when there is no traffic observed at all, we have a placeholder such as: 

```
expectedFailingSvcOutput := &statTsRow{
  name:       "backend-traffic-split",
  apex:       "backend-svc",
  leaf:       "backend-svc",
  weight:     "0",
  success:    "-",
  rps:        "-",
  latencyP50: "-",
  latencyP95: "-",
  latencyP99: "-",
} 
```
We can get rid of that if you think its not worth comparing it.

test/trafficsplit/trafficsplit_test.go
zaharidichev: I really miss pattern matching in cases like this one

zaharidichev:
> I'm not sure what assumptions the other tests make about if they can be run multiple times. However, it seems reasonable that this one could, because the first step is to set the traffic split object to the initial state. We could use a shorter -t time window on the stat command to ensure we're only looking at traffic metrics from after the most recent change to the traffic split.
> 
> e.g.
> 
>     Apply traffic split
>     Wait 30s
>     linkerd stat -t 30s

@adleong 
That seemed like a good idea until I realized that it will not work. The problem is that initially when traffic has never gone through the service all stats are filled with the placeholder `-`. Later if traffic has been observed, it does not matter whether this traffic falls outside of your -t window, you get 0.0 values for the stats. I think this is actually not right from UX perspective because a value of 0ms for the latency is not the same as "there was no latency observed at all because there was no traffic"...  So for example you end up with the following output: 
```

NAME                    APEX          LEAF          WEIGHT   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
backend-traffic-split   backend-svc   backend-svc     500m   100.00%   1.0rps           6ms          10ms          10ms
backend-traffic-split   backend-svc   failing-svc        0     0.00%   0.0rps           0ms           0ms           0ms
```

instead of 
```
NAME                    APEX          LEAF          WEIGHT   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
backend-traffic-split   backend-svc   backend-svc     500m   100.00%   1.0rps           6ms          10ms          10ms
backend-traffic-split   backend-svc   failing-svc        0     -          -           -           -             -  

```

adleong:
> Later if traffic has been observed, it does not matter whether this traffic falls outside of your -t window, you get 0.0 values for the stats.

That seems like a bug.  I think your integration test just found a bug 🎊

zaharidichev:
@adleong Yes this does not look right. If you are okey with this PR, lets go ahead with it. I will open an issue to fix this bug and change the test accordingly.

adleong:
My preference would be to fix the bug you found first so that we don't have to check in an incorrect test.  Tests don't have to be on master to be useful 😄

zaharidichev:
@adleong Added the fix for this rendering issue.  Feel free to take a look. I think this makes more sense now.

cli/cmd/stat.go
```diff
@@ -310,6 +315,7 @@ func writeStatsToBuffer(rows []*pb.StatTable_PodGroup_Row, w *tabwriter.Writer,
 				tcpOpenConnections: r.GetTcpStats().GetOpenConnections(),
 				tcpReadBytes:       getByteRate(r.GetTcpStats().GetReadBytesTotal(), r.TimeWindow),
 				tcpWriteBytes:      getByteRate(r.GetTcpStats().GetWriteBytesTotal(), r.TimeWindow),
+				hasRequestData:     statHasRequestData(r.Stats),
```
adleong: Would it be simpler to make this `if r.Stats != nil && statHasRequestData(r.Stats)` on line 308 instead of adding a new boolean field to this struct?  That way, `stats[key].rowStats` will be nil in the case where there is no traffic and the existing empty template code path will be used.

zaharidichev: Yes I did it like that initially, but then I realized that very similar logic needs to be applied to the [routes command](https://github.com/linkerd/linkerd2/pull/3649/files#diff-e87f0b2931093d08df1cea47d36797a6R204-R224). And there the difference is that there is no `nil` added to the list you are iterating over. So instead of forcefully adding a nil there or skipping the row altogether, which is not an option, I decided to unify the logic between the two places a bit and rely on the bool. I can change it to rely on a `nil` being present if you think its better. Your call!

adleong: Ah, I see what you mean.  The routes row handling code seems either too coupled, or not coupled enough to the stat row code 😂 

Given that, I think my preference would be to handle this with a nil for stat and do something different for routes.  At the time when we are rendering the row, we can check the counts to see if there are no requests and render the blank line instead.  I don't think we need to persist that boolean in the struct.

zaharidichev: I edited it and took the bool approach to the routes part only. Hope that is better now

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+
+		statRows[row.leaf] = row
+
+	}
+	return statRows, nil
+}
+
+func statTrafficSplit(from string, ns string) (map[string]*statTsRow, error) {
+	time.Sleep(31 * time.Second)
```
adleong: I'd consider moving this sleep out into the main body of the test, or renaming this function to make it more clear that this will block and wait for old stats to fall out of the window.

test/trafficsplit/trafficsplit_test.go
```diff
@@ -0,0 +1,263 @@
...
+			t.Error(err)
+		}
+
+		if err := TestHelper.CheckDeployment(prefixedNs, deploy, 1); err != nil {
+			t.Error(fmt.Errorf("Error validating deployment [%s]:\n%s", deploy, err))
+		}
+	}
+
+	t.Run("ensure traffic is sent to one backend only", func(t *testing.T) {
+		err := TestHelper.RetryFor(40*time.Second, func() error {
```
adleong: Food for thought, are the sleep and the retryFor redundant?  Is it possible to just use the retryFor to keep trying until the old stats fall out of the window?

zaharidichev:
@adleong Will try out your suggestions. Regarding the mixed up commits... The cert rotation test was initially a PR against this PR and the intention was to merge this first. However the opposite happened and the cert rotation branch got merged into this branch here. Once this is approved, I will rebase and merge them as two separate commits which should net the same result as if they were separate PRs against master. Retrospectively, I should have merged them in the order I originally intended to...

PR #3650, opened 2019-10-29
Fix whitespace path handling in non-docker (build) scripts

joakimr-axis:
Handling of whitespace paths was not fully implemented; this patch adds
the missing pieces. Also, only use bash where bash-specific
functionality is used/needed.

Change-Id: Ia5846443a8029c7aa039cb1cce4411438dfef54b


bin/mkube
```diff
@@ -8,8 +8,7 @@
 # If we're running under WSL then we have to use the Windows native Minikube
 # as the Linux version doesn't work in WSL. Assume WSL is Microsoft's only
 # Linux distro.
-uname -r | grep "Microsoft" > /dev/null
-if  [ $? -ne 0 ]; then
+if uname -r | grep Microsoft > /dev/null; then
```
zaharidichev: I think this logic needs to be reversed. Note the `if  [ $? -ne 0 ]; then` you have deleted. Now under MacOS for example `MINIKUBE_EXE` gets evalauted to `  MINIKUBE_EXE="${MINIKUBE_EXE:-/mnt/c/Program Files (x86)/Kubernetes/Minikube/minikube.exe}"` at least for me.

go.sum
```diff
@@ -504,9 +504,11 @@ k8s.io/client-go v0.0.0-20181213151034-8d9ed539ba31/go.mod h1:7vJpHMYJwNQCWgzmNV
 k8s.io/client-go v0.0.0-20190620085101-78d2af792bab h1:E8Fecph0qbNsAbijJJQryKu4Oi9QTp5cVpjTE+nqg6g=
 k8s.io/client-go v0.0.0-20190620085101-78d2af792bab/go.mod h1:E95RaSlHr79aHaX0aGSwcPNfygDiPKOVXdmivCIZT0k=
 k8s.io/code-generator v0.0.0-20181116203124-c2090bec4d9b/go.mod h1:MYiN+ZJZ9HkETbgVZdWw2AsuAi9PZ4V80cwfuf2axe8=
+k8s.io/code-generator v0.0.0-20190612205613-18da4a14b22b h1:p+PRuwXWwk5e+UYvicGiavEupapqM5NOxUl3y1GkD6c=
 k8s.io/code-generator v0.0.0-20190612205613-18da4a14b22b/go.mod h1:G8bQwmHm2eafm5bgtX67XDZQ8CWKSGu9DekI+yN4Y5I=
 k8s.io/component-base v0.0.0-20190620085130-185d68e6e6ea/go.mod h1:VLedAFwENz2swOjm0zmUXpAP2mV55c49xgaOzPBI/QQ=
 k8s.io/gengo v0.0.0-20190116091435-f8a0810f38af/go.mod h1:ezvh/TsK7cY6rbqRK0oQQ8IAqLxYwwyPxAX1Pzy0ii0=
+k8s.io/gengo v0.0.0-20190128074634-0689ccc1d7d6 h1:4s3/R4+OYYYUKptXPhZKjQ04WJ6EhQQVFdjOFvCazDk=
```
zaharidichev: Why did these end up changed?

bin/_test-run.sh
```diff
@@ -42,143 +42,143 @@ function upgrade_integration_tests() {
...
     fi
     if [ ! -x "$linkerd_path" ]; then
-        printf "\\n[%s] does not exist or is not executable\\n" "$linkerd_path"
+        printf '\\n[%s] does not exist or is not executable\\n' "$linkerd_path"
         exit 1
     fi
     exit_code=0
     "$linkerd_path" version --client > /dev/null 2>&1
-    exit_on_err "error running linkerd version command"
-    printf "[ok]\\n"
```
zaharidichev: Seems like changing from "" to '' messes up the new line  char interpretation and now this actually gets rendered as \n instead of producing a new line in my terminal.

go.sum
joakimr-axis: I wonder the same. Thanks for noticing. Should not be changed. Will fix.

bin/_test-run.sh
joakimr-axis: Should be printf '[ok]\n' and not printf '[ok]\\n'. Will fix.

bin/mkube
joakimr-axis: You are right. Will fix.

bin/test-clouds
```diff
@@ -57,7 +61,7 @@ done
 linkerd_path=$1
 check_linkerd_binary
 
-printf "\nKicking off tests for:\n- $AKS\n- $DO\n- $EKS\n- $GKE\n\n"
+printf '\nKicking off tests for:\n- %s\n- %s\n- %s\n- %s\n\n' "$AKS" "$DO" "$EK" "$GKE"
```
alpeb: You missed the `S` in `$EKS`

joakimr-axis: Thank you for noticing, fixed now!

Issue #3651, opened 2019-10-29
I18N for the dashboard


grampelberg:
### What problem are you trying to solve?

The dashboard should be usable in languages other than english.

### How should the problem be solved?

The dashboard should be i18n'd so that there are strings that can be handed out to others for translation and then loaded based on a browser's language.

scottcarol:
The below comment is by @cynthia-sg from #3757: 

//////////////////////

> * What are the major options available for internationalizing React? Beyond the `react-i18next` library, are there any good options? What are the positives and negatives about using them?

Almost all libraries use the same approach: they provide higher-order components, which injects internationalization functions for handling messages, dates, numbers and currencies via React’s context features. The most popular options available for internationalizing React are (the code samples are trying to keep at minimum the verbose boilerplate added to the current code):

#### react-intl

This is the most popular option. It’s part of [FormatJS](https://formatjs.io). This library provides React components and an API to format dates, numbers, and strings, including pluralization and handling translations. One of their cool features is the option to display dates relative to "now" (perhaps we can replace `date-fns`).

It was created by `Yahoo`, but from the end of 2018 Yahoo isn't providing a lot of support for this library.

**Use**: We need to wrap the component with the `injectIntl` HOC and then we can get the translations ready only using `intl` function.

```javascript
import { injectIntl } from "react-intl";

const Component = ({ intl }) => (
  <Tipography>
    { intl.formatMessage({ id: "This is a sample text." }) }
  <Tipography>
);

export default injectIntl(Component);
```

#### react-i18next

This library is based on `i18next`, which works well with small projects thanks to its simplicity (the use of `namespaces` and `nesting` is pretty useful to keep organised the data); but thinking big, `i18next` gives you the option to separate translations into multiple files and to load them on demand.

**Use**: We need to wrap our component with the `withTranslation` HOC passing our namespace (we can use a common one by default) and then we can get the translations ready only using `t` function.

```javascript
import { withTranslation } from "react-i18next";

const Component = ({ t }) => (
  <Tipography>
    { t("This is a sample text.") }
  <Tipography>
);

export default withTranslation()(Component);
```

#### linguiJS

This library is in many ways very similar to `react-intl`: both libraries use the same syntax and they also have very similar API. The library is compatible with both JSON and the PO translation file formats.

**Use**: We need to wrap our component with the `withI18n` and then we can get the translations ready only using `t` function imported from `@lingui/macro`.

```javascript
import { withI18n } from "@lingui/react"
[...]

PR #3656, opened 2019-10-31
Add `linkerd check` to dashboard

### LINKED ISSUE #3613: Allow user to run `linkerd check` from the dashboard and see results
Running `linkerd check` in the CLI is a handy diagnostic tool. We can allow users to run `linkerd check` from the web dashboard from the `/controlplane` view.

Control Plane view should now have:
1. A button titled "Run Linkerd Check" that, when clicked runs `check`
2. After running check, display the output as displayed in the CLI.

```
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-config
--------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane self-check
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √
```

tegioz:
`linkerd check` can now be run from the `/controlplane` view by clicking on the button added below the service mesh details table. Once the checks results are received, they are displayed in a modal in a similar way to the CLI output.

Closes #3613 

check-button

linkerd-check_01

linkerd-check_02

### DIFF
--- a/charts/linkerd2/templates/web-rbac.yaml
+++ b/charts/linkerd2/templates/web-rbac.yaml
@@ -5,6 +5,82 @@
 ###
 ---
 {{- if not .RestrictDashboardPrivileges }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: linkerd-web-check
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: web
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["namespaces", "configmaps",]
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["serviceaccounts", "pods"]
+  verbs: ["list"]
+- apiGroups: ["apps"]
+  resources: ["replicasets"]
+  verbs: ["list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: linkerd-{{.Namespace}}-web-check
+  labels:
+    {{.ControllerComponentLabel}}: web
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+rules:
+- apiGroups: ["rbac.authorization.k8s.io"]
+  resources: ["clusterroles", "clusterrolebindings"]
+  verbs: ["list"]
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["list"]
+- apiGroups: ["admissionregistration.k8s.io"]
+  resources: ["mutatingwebhookconfigurations", "validatingwebhookconfigurations"]
+  verbs: ["list"]
+- apiGroups: ["policy"]
+  resources: ["podsecuritypolicies"]
+  verbs: ["list"]
+- apiGroups: ["linkerd.io"]
+  resources: ["serviceprofiles"]
+  verbs: ["list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: linkerd-web-check
+  namespace: {{.Namespace}}
+  labels:
+    {{.ControllerComponentLabel}}: web
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+roleRef:
+  kind: Role
+  name: linkerd-web-check
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: linkerd-web
+  namespace: {{.Namespace}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: linkerd-{{.Namespace}}-web-check
+  labels:
+    {{.ControllerComponentLabel}}: web
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+roleRef:
+  kind: ClusterRole
+  name: linkerd-{{.Namespace}}-web-check
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: linkerd-web
+  namespace: {{.Namespace}}
+---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
--- a/go.mod
+++ b/go.mod
@@ -58,7 +58,6 @@ require (
 	github.com/soheilhy/cmux v0.1.4 // indirect
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.3
-	github.com/stretchr/testify v1.3.0
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5 // indirect
 	github.com/wercker/stern v0.0.0-20190705090245-4fa46dd6987f
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -58,6 +58,7 @@ require (
 	github.com/soheilhy/cmux v0.1.4 // indirect
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.3
+	github.com/stretchr/testify v1.3.0
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5 // indirect
 	github.com/wercker/stern v0.0.0-20190705090245-4fa46dd6987f
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
--- a/web/app/css/styles.css
+++ b/web/app/css/styles.css
@@ -1,5 +1,5 @@
 /* material-ui uses Roboto */
-@import url("https://fonts.googleapis.com/css?family=Montserrat:400,700|Roboto:100,300,400");
+@import url("https://fonts.googleapis.com/css?family=Montserrat:400,700|Roboto:100,300,400|Roboto+Mono");
 
 :root {
   --linkblue: #2196f3;
--- a/web/app/js/components/CheckModal.jsx
+++ b/web/app/js/components/CheckModal.jsx
@@ -114,11 +114,11 @@ const Results = ({title, results, classes}) => {
         const resultType = getResultType(result.Err, result.Warning);
 
         return (
-          <Grid container direction="row" key={result.Description}>
+          <Grid container direction="row" alignItems="flex-start" key={result.Description}>
             <Grid item>
               <Icon type={resultType} classes={classes} />
             </Grid>
-            <Grid item>
+            <Grid item xs>
               <Typography className={classes.result} color="inherit" data-i18n={`${title}_${index + 1}`}>
                 {result.Description}
               </Typography>
--- a/web/app/js/components/CheckModal.jsx
+++ b/web/app/js/components/CheckModal.jsx
@@ -158,7 +158,7 @@ Icon.propTypes = {
   type: PropTypes.oneOf(['success', 'error', 'warning']).isRequired,
 };
 
-export class CheckModal extends React.Component {
+class CheckModal extends React.Component {
   constructor(props) {
     super(props);
 
--- a/web/app/js/components/CheckModal.jsx
+++ b/web/app/js/components/CheckModal.jsx
@@ -1,0 +1,272 @@
+import Button from '@material-ui/core/Button';
+import CheckIcon from '@material-ui/icons/Check';
+import Chip from '@material-ui/core/Chip';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import CloseIcon from '@material-ui/icons/Close';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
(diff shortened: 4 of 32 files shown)


web/srv/api_handlers.go
```diff
@@ -39,6 +41,14 @@ var (
...
 		WriteBufferSize: maxMessageSize,
 	}
+
+	// Checks whose description matches the following regexp won't be included
+	// in the handleApiCheck output. In the context of the dashboard, some
+	// checks like cli or kubectl versions ones may not be relevant.
+	//
+	// TODO(tegioz): use more reliable way to identify the checks that should
+	// not be displayed in the dashboard (hint anchor is not unique).
+	excludedChecksRE = regexp.MustCompile(`(?i)cli|(?i)kubectl`)
```
tegioz: Hi @scottcarol,

To avoid introducing changes to the `healthcheck` package I'm omitting the checks not relevant to the dashboard on the dashboard web server using a regular expression over the checker description. I'm not very happy with this though.

I think a better solution would be to introduce a new option in `healthcheck.Options` to indicate the health checker instance will be used for the dashboard, as well as a new flag to `healthcheck.checker` to mark individual checks to be skipped if the health checker instance is meant to be used for the dashboard. That way unneeded tests wouldn't even be run and no unreliable filtering would be needed on the dashboard web server. Happy to make this change if you guys are happy with that approach (or happy to implement any other solution you prefer).

web/app/css/styles.css
```diff
@@ -1,5 +1,5 @@
 /* material-ui uses Roboto */
-@import url("https://fonts.googleapis.com/css?family=Montserrat:400,700|Roboto:100,300,400");
+@import url("https://fonts.googleapis.com/css?family=Montserrat:400,700|Roboto:100,300,400|Roboto+Mono");
```
cynthia-sg: Would it be possible to add `Roboto Mono` font to get a `terminal` aspect for the modal?

web/app/js/components/CheckModal.jsx
```diff
@@ -0,0 +1,311 @@
...
+              direction="row"
+              justify="space-between"
+              alignItems="center">
+              <Grid item>
+                Linkerd Check
+              </Grid>
+
+              {success !== undefined &&
+                <Grid item>
+                  <Chip
```
cynthia-sg: I was wondering if I can add a new not required prop to `util/Chip.jsx` for the icon to reuse this component on this PR.

web/app/css/styles.css
scottcarol: @cynthia-sg, that sounds good to me. 👍

web/app/js/components/CheckModal.jsx
scottcarol: Hi @cynthia-sg, what is the unrequired prop you would add to the `Chip` component? Can you paste in an example of the diffed code?

scottcarol:
@tegioz @cynthia-sg, this is looking good. I have some UI feedback and suggestions, but I first want to sit down with @grampelberg tomorrow morning to make sure we agree on those suggestions. Once I talk with him I will add more review comments. ✔️

web/app/js/components/CheckModal.jsx
cynthia-sg: One approach could be this: 

```
import Chip from '@material-ui/core/Chip';
import React from 'react';

const styles = theme => ({
  good: {
    color: theme.status.dark.good,
    border: '1px solid ' + theme.status.dark.good
  },
  warning: {
    color: theme.status.dark.warning,
    border: '1px solid ' + theme.status.dark.warning,
  },
  bad: {
    color: theme.status.dark.danger,
    border: '1px solid ' + theme.status.dark.danger,
  }
});

function SimpleChip(props) {
  const { classes, label, type, icon } = props;

  return (
    <Chip
      className={classes[type]}
      label={label}
      icon={icon}
      variant="outlined" />
  );
}

SimpleChip.propTypes = {
  classes: PropTypes.shape({}).isRequired,
  icon: PropTypes.element,
  label: PropTypes.string.isRequired,
  type: PropTypes.string.isRequired,
};

SimpleChip.defaultProps = {
  icon: undefined,
};

export default withStyles(styles, { withTheme: true })(SimpleChip);
```

Another approach to use always same icons could be: 

```
import Check from '@material-ui/icons/Check';
import Chip from '@material-ui/core/Chip';
import Close from '@material-ui/icons/Close';
import PriorityHigh from '@material-ui/icons/PriorityHigh';
import React from 'react';

const styles = theme => ({
  good: {
    color: theme.status.dark.good,
    border: '1px solid ' + theme.status.dark.good
  },
  warning: {
    color: theme.status.dark.warning,
    border: '1px solid ' + theme.status.dark.warning,
  },
  bad: {
    color: theme.status.dark.danger,
    border: '1px solid ' + theme.status.dark.danger,
  },
  icon: {
    border: 0,
    marginLeft: "10px",
  },
});

const Icon = ({type, classes}) => {
  const className = `${classes[type]} ${classes.icon}`;

  return (
    <React.Fragment>
      {(() => {
        switch (type) {
          case "good":
            return <Check className={className} fontSize="small" />;
          case "bad":
            return <Close className={className} fontSize="small" />;
          case "warning":
            return <PriorityHigh className={className} fontSize="small" />;
          default:
            null;
        }
      })()}
    </React.Fragment>
  );
};

Icon.propTypes = {
  classes: PropTypes.shape({}).isRequired,
  type: PropTypes.string.isRequired,
};

function SimpleChip(props) {
  const { classes, label, type, icon } = props;

  return (
    <Chip
      className={classes[type]}
      label={label}
      icon={icon ? <Icon classes={classes} type={type} /> : undefined}
      variant="outlined" />
  );
}

SimpleChip.propTypes = {
  classes: PropTypes.shape({}).isRequired,
  icon: PropTypes.bool.isRequired,
  label: PropTypes.string.isRequired,
  type: PropTypes.string.isRequired,
};

SimpleChip.defaultProps = {
  icon: false,
};

export default withStyles(styles, { withTheme: true })(SimpleChip);
```

go.mod
```diff
@@ -58,6 +58,7 @@ require (
 	github.com/soheilhy/cmux v0.1.4 // indirect
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.3
+	github.com/stretchr/testify v1.3.0
```
ihcsim: I am a bit hesitant to introduce `testify` as a dependency here. Looking at the changes in `web/srv/api_handlers_test.go`, I don't think we are gaining much, since `require.NoError()` is just [`if err != nil`](https://github.com/stretchr/testify/blob/a88bf7aab8f390daf0a71952568272f6638b319e/assert/assertions.go#L1225-L1227) and `assert.Equal()` is just `==`.

@scottcarol WDYT?

tegioz: I've only used some basic helpers on that test, but the `assert` and `require` packages contain [a lot](https://godoc.org/github.com/stretchr/testify/assert) of helpful helpers that may save a lot of boilerplate when writing tests. Even in this small example it's saving a few `if` blocks, making tests very easy to read. It'd be great if you could consider introducing it to the project :)

scottcarol: @tegioz Could you open an issue proposing using this dependency? That way other Go developers on the project could chime in with feedback. For this specific PR, let's omit `testify` as a dependency, and then if we do have consensus on adding that dependency to our codebase, we can systematically update the tests. 😃

tegioz: Sure, no problem 👍

web/srv/api_handlers.go
tegioz: Moved to #3670

tegioz:
> Thanks for your work so far on this @tegioz @cynthia-sg!

Thanks @scottcarol!

> 
> I have some big-picture feedback and some smaller UI feedback. Once we have the UI/UX finalized, I can give detailed code review.
> 
> Big-picture:
> 
> * As the feature is now, a Dialog opens once `check` is complete, to display the results. Could we instead open the Dialog and stream `check` into the Dialog as it is running, to feel more like the CLI tool? Let me know your thoughts about this; we could merge the PR as is, and have that be a follow-up ticket, if you prefer.

Yes, that sounds good. We're happy to do it as part of this PR or in a new one, whatever works best for you. The rest of the feedback in this comment has already been addressed. Maybe you can give it a try to check how it looks now and decide based on that if you prefer to merge it as it is or wait until the check results are streamed. We'll update the screenshots at the top right now.

> * Currently we are ranging through `Object.keys(results)` to display the returned `check` results. However, we are planning to move to an internationalized (i18n) dashboard (#3651) where text strings are available for different languages/locales. Could the results instead display with unique classnames to be ready for an i18n library [like this](https://github.com/i18next/react-i18next)?
> 
> More specific feedback:
> 
> * If a check fails, the test should report the helptext/message, the way the CLI does:
>   Screen Shot 2019-11-04 at 1 15 20 PM
>   we can even have a link to the specific documentation if returned.

Done.

> * Users should be able to re-run the check from the Dialog.

Done.

> * To be consistent with our other usage of `<Chip>`, that component should just read `success` or `failure` without an `√` or `X`, in the same way that we do for a pod detail page:
>   Screen Shot 2019-11-04 at 12 52 06 PM

Done.

cynthia-sg:
> * Currently we are ranging through `Object.keys(results)` to display the returned `check` results. However, we are planning to move to an internationalized (i18n) dashboard (#3651) where text strings are available for different languages/locales. Could the results instead display with unique classnames to be ready for an i18n library [like this](https://github.com/i18next/react-i18next)?
> 

Hi, @scottcarol, I'm not sure what do you mean about `unique classnames` for i18n. I have added for now:
```
data-i18n={`${title}_${index + 1}`}
``` 
 Please, let me know your thoughts about this approach.

scottcarol:
@tegioz @cynthia-sg I think the UI looks great. We can create a separate issue for a streaming `check` once this PR is merged. I'll take a more detailed look at the code next.

web/srv/api_handlers.go
```diff
@@ -310,3 +320,35 @@ func (h *handler) handleAPIEdges(w http.ResponseWriter, req *http.Request, p htt
...
+		*healthcheck.CheckResult
+		ErrMsg  string `json:",omitempty"`
+		HintURL string `json:",omitempty"`
+	}
+
+	results := make(map[healthcheck.CategoryID][]*CheckResult)
+
+	collectResults := func(result *healthcheck.CheckResult) {
+		if result.Retry || excludedChecksRE.MatchString(result.Description) {
+			return
```
klingerf: I just realized that the checks that you're skipping here could be failing, and that would cause `RunChecks` to return `false`. But if the only failed checks are the ones that were skipped, that will cause us to display all passing checks in the UI, but with a chip that says that checks are failing. Instead of relying on the return value of `RunChecks`, I think you need to instead calculate success manually inside this function, based only on the checks that are not being skipped.

tegioz: Good point @klingerf, thanks!

tegioz:
Hi @klingerf 

I have one question, maybe you can help me.

To reproduce the issue about the CA certificates you posted I tested the branch running the dashboard web server inside my dev k8s cluster instead of running it locally. When running the checks from the dashboard, before being able to see the error you saw, I got this one:

Screenshot 2019-11-08 at 13 30 10

I realised I hadn't created any `ClusterRole` and `ClusterRoleBinding` to give the `ServiceAccount` used by the web pod permissions to run the checks directly. Just to proceed with my original goal, I granted permissions to everything to that pod, saw the error you reported and committed the fix you suggested. All good on that end.

But I was wondering why didn't you experience the same issue. Did you run the dashboard in a different way with different credentials? 

I'm thinking that I probably should identify the permissions required to run the checks from the web pod directly, create a `ClusterRole` for them and bind it to the web `ServiceAccount`. Does that sound right or am I missing something?

klingerf:
Hi @tegioz, that's a good catch! I think my local dev k8s cluster doesn't have RBAC enforcement enabled by default, which is why I didn't see those errors. I agree that we'll need to grant the `linkerd-web` service account all of the privileges required to run those checks. Right now it's only granted the `linkerd-linkerd-tap-admin` cluster role, but it will need an additional role that covers the privileges for check.

tegioz:
Thanks @klingerf, I will take care of it 👍

charts/linkerd2/templates/web-rbac.yaml
```diff
@@ -5,6 +5,82 @@
 ###
 ---
 {{- if not .RestrictDashboardPrivileges }}
```
klingerf: Should we change the description for the flag that controls this option? Right now it reads:

```
      --restrict-dashboard-privileges             Restrict the Linkerd Dashboard's default privileges to disallow Tap
```

Maybe we should rephrase to "... disallow Tap and Check" or something like that.

tegioz: Done, thanks!

scottcarol:
@tegioz - @alpeb and I are looking into why the tests are failing, this may be a Github Actions issue. Otherwise everything looks good and ready to merge. 👍

tegioz:
Awesome, thanks @scottcarol 👍

scottcarol:
Merging now - @adleong, thanks for your comments re: #3670 - that issue is currently more in an "idea/brainstorm" state than a "definitely do this" state so I'm merging this. I just marked #3670 as `needs/design` to clarify this.

alpeb:
Hi @cynthia-sg, I just wanted to ping you about an issue I found when testing this functionality, that I documented in #3703 . Let me know if you can't reproduce the problem.

cynthia-sg:
> Hi @cynthia-sg, I just wanted to ping you about an issue I found when testing this functionality, that I documented in #3703 . Let me know if you can't reproduce the problem.

Thanks, @alpeb , will look into it soon

Issue #3543, opened 2019-10-08
Remove the destination container from the controller pod (again)


adleong:
#3268 describes removing the destination container out of the controller pod.  This was implemented in #3407.  However, in order to facilitate graceful upgrade, we added the destination container back into the controller in #3540 in addition to keeping the dedicated destination deployment.  This allows proxies to gracefully upgrade from using the destination container in the controller to the destination deployment.  This graceful upgrade will happen when users upgrade to stable-2.6.

After 2.6 is released, we should remove the destination container from the controller pod again since it will be unused once the upgrade to 2.6 is complete.

PR #3661, opened 2019-11-01
Remove destination container from controller

zaharidichev:
Now that we are going to release 2.7, we can remove the destination container from the controller again. 

Fixes: #3543 

### DIFF
--- a/test/routes/routes_test.go
+++ b/test/routes/routes_test.go
@@ -39,7 +39,8 @@ func TestRoutes(t *testing.T) {
 		c int
 	}{
 		{"linkerd-controller-api", 9},
-		{"linkerd-destination", 4},
+		{"linkerd-destination", 1},
+		{"linkerd-dst", 3},
 		{"linkerd-grafana", 12},
 		{"linkerd-identity", 2},
 		{"linkerd-prometheus", 5},
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -23,25 +23,6 @@ spec:
     port: 8085
     targetPort: 8085
 ---
-kind: Service
-apiVersion: v1
-metadata:
-  name: linkerd-destination
-  namespace: {{.Namespace}}
-  labels:
-    {{.ControllerComponentLabel}}: controller
-    {{.ControllerNamespaceLabel}}: {{.Namespace}}
-  annotations:
-    {{.CreatedByAnnotation}}: {{default (printf "linkerd/helm %s" .LinkerdVersion) .CliVersion}}
-spec:
-  type: ClusterIP
-  selector:
-    {{.ControllerComponentLabel}}: controller
-  ports:
-  - name: grpc
-    port: 8086
-    targetPort: 8086
----
 {{ $_ := set .Proxy "WorkloadKind" "deployment" -}}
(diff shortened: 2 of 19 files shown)


alpeb:
The problem still remains, that proxies that haven't restarted since pre-2.6 will fail when the old destination service is finally removed here, correct? The idea was just to give clusters a little more time, until 2.7?

grampelberg:
@alpeb the issue would be anyone skipping a stable or edge between the two right? I'd rather get this taken out early and make sure folks are okay than wait until the last second.

alpeb:
@grampelberg Even if people haven't skipped releases on the control plane, what matters is whether their proxies haven't been restarted since they installed a version older than 2.6.
I'm ok with removing this, just clarifying we'll still need to have a word about it in the upgrade notes.

grampelberg:
We should definitely call it out in the upgrade notes, good call.

ihcsim:
Yeah, the `Service` will need to remain as `linkerd-dst`. Otherwise, once the control plane is upgraded to 2.7, the 2.6 data plane will experience downtime, until it is restarted. If we want to make everything consistent, we can attempt to change the service account, deployment, rbac etc. to `linkerd-dst`, as @adleong suggested in https://github.com/linkerd/linkerd2/pull/3540#pullrequestreview-298442546. I didn't do it in https://github.com/linkerd/linkerd2/pull/3540 because of the number of integration test failures that needed to be fixed right before the release goes out the door.

> We should definitely call it out in the upgrade notes, good call.

Yeah, we need to emphasize in the upgrade doc that the upgrade process isn't completed, until the data plane is rolled, regardless of which release.

zaharidichev:
@ihcsim Thanks for the exaplanation. That makes perfect sense. I see where my naming mistake is now.

> Yeah, we need to emphasize in the upgrade doc that the upgrade process isn't completed, until the data plane is rolled, regardless of which release.

The only thing I could not quite get is whether this needs to happen in this PR or its part of something else?

zaharidichev:
Ahh, need to fix the integration test.

alpeb:
> Yeah, we need to emphasize in the upgrade doc that the upgrade process isn't completed, until the data plane is rolled, regardless of which release.

Great point. To be perfectly clear, to avoid downtime, the data plane rollout needs to occur _before_  the upgrade while still running 2.6 in the control plane before upgrading to 2.7.
But that only applies if the user came from pre-2.6 to 2.6 and thus their data plane is in a "transitional" state (hitting `Destination` inside `linkerd-controller`).
If they installed for the first time in 2.6 then there's no need of a rollout.
And if they are in pre-2.6 they need to go through 2.6 before upgrading to 2.7 to also avoid downtime.
Too much detail? :-P

test/routes/routes_test.go
```diff
@@ -39,7 +39,8 @@ func TestRoutes(t *testing.T) {
 		c int
 	}{
 		{"linkerd-controller-api", 9},
-		{"linkerd-destination", 4},
+		{"linkerd-destination", 1},
```
adleong: what does this instance of `linkerd-destination` correspond to?

zaharidichev: This simply does a count on the number of string occurrences. The deployment is still named `linkerd-destination`. This is why it shows up.

PR #3664, opened 2019-11-01
Improve how empty tables are displayed in dashboard

### LINKED ISSUE #3611: Display an "Empty State" Component for dashboard tables with no data
*Problem*: Currently, if there is no data to display in a table, we simply show the table headers without explicitly saying there is no data to return, or explaining why. 

empty-state-table

*Solution*: @grampelberg is designing an empty state view. Before he has created mockups, we can create a placeholder card component that reads "No data to display", that will eventually be styled and include content. This should display within every type of table (resource view, tap, top, etc.)

empty-state

cynthia-sg:
Currently, if there is no data to display in a table in the dashboard, only the table headers are shown.

This commit introduces an empty table view that is displayed when there is no data available. The component introduced can be customized with a custom text if desired.

emptyTable-1

emptyTable-2

Closes #3611


scottcarol:
This is a really creative approach and absolutely gives the look and feel of the mockups. However, #3611 specified a a [card](https://material-ui.com/components/cards/) component, in order to have the freedom to implement different layout/design combinations in the future for this empty state table. Could this PR be updated so that the component is a card instead of a  `<TableCell>`?

cynthia-sg:
Sorry, @scottcarol, I lost `card` concept from ticket definition. I have pushed some changes to use `Card` component. Please, let me know your feedback.

PR #3658, opened 2019-10-31
update changelog for edge-19.10.5

hawkw:
## edge-19.10.5

This edge release adds support for integrating Linkerd's public-key
infrastructure with an external certificate issuer such as [`cert-manager`],
adds distributed tracing support to the Linkerd control plane, and adds
protection against DNS rebinding attacks to the web dashboard. In addition, it
includes several improvements to the Linkerd CLI.

* CLI
  * Added a new `--identity-external-issuer` flag to `linkerd install` that
    configures Linkerd to use certificates issued by an external certificate
    issuer (such as `cert-manager`)
  * Added support for injecting a namespace to `linkerd inject` (thanks
    @mayankshah1607!)
  * Added checks to `linkerd check --preinstall` ensuring Kubernetes Secrets
    can be created and accessed
  * Fixed `linkerd tap` sometimes displaying incorrect pod names for unmeshed
    IPs that match multiple running pods
* Controller
  * Added support for using trust anchors from an external certificate issuer
    (such as `cert-mananger`) to the `linkerd-identity` service
* Web UI
  * Added `Host:` header validation to the `linkerd-web` service, to protect
    against DNS rebinding attacks
* Internal
  * Added new `--trace-collector` and `--trace-collector-svc-account` flags to
    `linkerd inject` that configures the OpenCensus trace collector used by
    proxies in the injected workload (thanks @Pothulapati!)
  * Added a new `--control-plane-tracing` flag to `linkerd install` that enables
    distributed tracing in the control plane (thanks @Pothulapati!)
  * Added distributed tracing support to the control plane (thanks
    @Pothulapati!)

Also, thanks to @joakimr-axis for several fixes and improvements to internal
build scripts!

PR #3665, opened 2019-11-01
rename `edge-19.10.5` to `edge-19.11.1`

hawkw:
The intention was to release `edge-19.10.5` yesterday, but due to CI
issues, the release had to be delayed until today. Since yesterday was
October 31st, the release's version would have been `edge-19.10.5`;
however, today is November 1st, so the version number is now
`edge-19.11.1`. Unfortunately, I only realised this *immediately after*
I merged #3658...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-## edge-19.10.5
+## edge-19.11.1
 
 This edge release adds support for integrating Linkerd's PKI with an external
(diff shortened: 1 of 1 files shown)


olix0r:
i'm fine with leaving it as 19.10.5, since it started yestdrday

hawkw:
@olix0r okay, we'll continue with 19.10.5, then!

Issue #3614, opened 2019-10-21
Add Cron Jobs and Replica Sets to dashboard


scottcarol:
The Kubernetes dashboard has views for these workloads:
```
Cron Jobs
Daemon Sets
Deployments
Jobs
Pods
Replica Sets
Replication Controllers
Stateful Sets
```
The Linkerd dashboard does not show data for `Cron Jobs` or `Replica Sets`. We should add these workload types to the sidebar, list them by namespace and return stats for them if possible.

tegioz:
Similar to #3584

tegioz:
Similar to #3585

PR #3666, opened 2019-11-01
Make resources definition available to dashboard

### LINKED ISSUE #3615: Fetch the resource definition when on dashboard detail view
In the future, the resource detail page of the dashboard will be redesigned, and the dashboard will also show service profiles. 

We can prepare for this by fetching the resource definition when on a resource detail page. This should be the same definition returned if you query kubectl:
```
kubectl get po/podname -oyaml
```

This should not (yet) be shown in the UI, but can be wired through to the `ResourceDetail` component and added to state as as `resourceDefinition`.

tegioz:
This is some groundwork in preparation for the resource detail page redesign of the dashboard. The resource definition in yaml format has been wired through to the `ResourceDetail` component state.

Closes #3615

### DIFF
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -84,7 +84,11 @@ export class ResourceDetailBase extends React.Component {
       resourceIsMeshed: true,
       pendingRequests: false,
       loaded: false,
-      error: null
+      error: null,
+      resourceDefinition: null,
+      // queryForDefinition is set to false now due to we are not currently using
+      // resource definition. This can change in the future
+      queryForDefinition: false,
     };
   }
 
@@ -152,6 +156,11 @@ export class ResourceDetailBase extends React.Component {
         )
       ];
 
+    if (this.state.queryForDefinition) {
+      // definition for this resource
+      apiRequests.push(this.api.fetchResourceDefinition(resource.namespace, resource.type, resource.name));
+    }
+
     if (_indexOf(edgeDataAvailable, resource.type) > 0) {
       apiRequests = apiRequests.concat([
         this.api.fetchEdges(resource.namespace, resource.type)
@@ -161,12 +170,19 @@ export class ResourceDetailBase extends React.Component {
     this.api.setCurrentRequests(apiRequests);
 
     Promise.all(this.api.getCurrentPromises())
-      .then(([resourceRsp, podListRsp, podMetricsRsp, upstreamRsp, downstreamRsp, edgesRsp]) => {
+      .then(results => {
+        const [resourceRsp, podListRsp, podMetricsRsp, upstreamRsp, downstreamRsp, ...rsp] = [...results];
         let resourceMetrics = processSingleResourceRollup(resourceRsp, resource.type);
         let podMetrics = processSingleResourceRollup(podMetricsRsp, resource.type);
         let upstreamMetrics = processMultiResourceRollup(upstreamRsp, resource.type);
         let downstreamMetrics = processMultiResourceRollup(downstreamRsp, resource.type);
-        let edges = processEdges(edgesRsp, this.state.resource.name);
+        let resourceDefinition = this.state.queryForDefinition ? rsp[0] : this.state.resourceDefinition;
+
+        let edges = [];
+        if (_indexOf(edgeDataAvailable, resource.type) > 0) {
+          const edgesRsp = rsp[rsp.length - 1];
+          edges = processEdges(edgesRsp, this.state.resource.name);
+        }
 
         // INEFFICIENT: get metrics for all the pods belonging to this resource.
         // Do this by querying for metrics for all pods in this namespace and then filtering
@@ -233,7 +249,8 @@ export class ResourceDetailBase extends React.Component {
           loaded: true,
           pendingRequests: false,
           error: null,
-          unmeshedSources: this.unmeshedSources // in place of debouncing, just update this when we update the rest of the state
+          unmeshedSources: this.unmeshedSources, // in place of debouncing, just update this when we update the rest of the state
+          resourceDefinition, // eslint-disable-line react/no-unused-state
         });
       })
       .catch(this.handleApiError);
--- a/web/app/js/components/util/ApiHelpers.test.jsx
+++ b/web/app/js/components/util/ApiHelpers.test.jsx
@@ -303,7 +303,7 @@ describe('ApiHelpers', () => {
       expect(fetchStub.args[0][0]).toEqual('/api/check');
     });
   });
-  
+
   describe('fetchResourceDefinition', () => {
     it('fetches the resource definition from the api', () => {
       const [namespace, type, name] = ["namespace", "type", "name"];
@@ -311,7 +311,7 @@ describe('ApiHelpers', () => {
       api.fetchResourceDefinition(namespace, type, name);
 
       expect(fetchStub.calledOnce).toBeTruthy;
-      expect(fetchStub.args[0][0]).toEqual(`/api/namespaces/${namespace}/${type}/${name}`);
+      expect(fetchStub.args[0][0]).toEqual(`/api/resource-definition?namespace=${namespace}&resource_type=${type}&resource_name=${name}`);
     });
   });
 });
--- a/web/app/js/components/util/ApiHelpers.test.jsx
+++ b/web/app/js/components/util/ApiHelpers.test.jsx
@@ -303,4 +303,15 @@ describe('ApiHelpers', () => {
       expect(fetchStub.args[0][0]).toEqual('/api/check');
     });
   });
+  
+  describe('fetchResourceDefinition', () => {
+    it('fetches the resource definition from the api', () => {
+      const [namespace, type, name] = ["namespace", "type", "name"];
+      api = ApiHelpers();
+      api.fetchResourceDefinition(namespace, type, name);
+
+      expect(fetchStub.calledOnce).toBeTruthy;
+      expect(fetchStub.args[0][0]).toEqual(`/api/namespaces/${namespace}/${type}/${name}`);
+    });
+  });
 });
--- a/web/srv/api_handlers.go
+++ b/web/srv/api_handlers.go
@@ -21,6 +21,8 @@ import (
 	"github.com/linkerd/linkerd2/pkg/protohttp"
 	"github.com/linkerd/linkerd2/pkg/tap"
 	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
 )
 
 // Control Frame payload size can be no bigger than 125 bytes. 2 bytes are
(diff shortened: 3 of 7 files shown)


web/app/js/components/ResourceDetail.jsx
```diff
@@ -149,7 +149,9 @@ export class ResourceDetailBase extends React.Component {
         // downstream resources of this resource (meshed traffic only)
         this.api.fetchMetrics(
           `${this.api.urlsForResource("all")}&from_name=${resource.name}&from_type=${resource.type}&from_namespace=${resource.namespace}`
-        )
+        ),
+        // definition for this resource
+        this.api.fetchResourceDefinition(resource.namespace, resource.type, resource.name),
```
cynthia-sg: Hi @scottcarol,

I’ve chained the new resource definition api call to the other ones in that component, so that all of them are called the same way. But the response of this new one is a bit large and maybe we don’t need to call it that often. What do you think?

Cheers

scottcarol:
@tegioz @cynthia-sg I'm looking forward to reviewing this one! Do you mind making the JS unit tests pass before I dig into this (see Github Actions results)? You can run the tests from `/linkerd2/`
```
$ bin/web setup # to fetch dependencies if needed
$ bin/web test
```

web/srv/server.go
```diff
@@ -165,6 +165,7 @@ func NewServer(
 	server.router.GET("/api/tap", handler.handleAPITap)
 	server.router.GET("/api/routes", handler.handleAPITopRoutes)
 	server.router.GET("/api/edges", handler.handleAPIEdges)
+	server.router.GET("/api/namespaces/:namespace/:kind/:name", handler.handleAPIResourceDefinition)
```
scottcarol: I would use `resource-type` instead of `kind` here

web/app/js/components/util/ApiHelpers.test.jsx
```diff
@@ -293,4 +293,15 @@ describe('ApiHelpers', () => {
       expect(url).toEqual('/api/tps-reports?resource_type=sts&all_namespaces=true&tcp_stats=true');
     })
   });
+
+  describe('fetchResourceDefinition', () => {
```
scottcarol: thanks for adding this to the test! ⭐️

web/app/js/components/ResourceDetail.jsx
```diff
@@ -227,6 +230,7 @@ export class ResourceDetailBase extends React.Component {
           upstreamMetrics,
           downstreamMetrics,
           edges,
+          resourceDefinition, // eslint-disable-line react/no-unused-state
```
scottcarol: to clarify, is this `eslint-disable` because we are fetching the resourceDefinition but not actually displaying it anywhere?

web/app/js/components/ResourceDetail.jsx
scottcarol: Great question @cynthia-sg! Having this code in the `ResourceDetail` component is great for testing. Because the goal of this project is to prepare for future work but not to display any data, I agree with you that we shouldn't make unnecessary queries. 

What about adding `queryForDefinition: false` to the state, and check for `true` before adding `fetchResourceDefinition` to the `apiRequests` array? That way your code is ready to run once we're utilizing the data. Open to other suggestions as well!

web/srv/api_handlers.go
```diff
@@ -310,3 +312,42 @@ func (h *handler) handleAPIEdges(w http.ResponseWriter, req *http.Request, p htt
 	}
 	renderJSONPb(w, result)
 }
+
+func (h *handler) handleAPIResourceDefinition(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
+	namespace := p.ByName("namespace")
+	kind := p.ByName("kind")
```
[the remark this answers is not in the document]
tegioz: Sure, done 👍

web/srv/api_handlers.go
```diff
@@ -310,3 +312,46 @@ func (h *handler) handleAPIEdges(w http.ResponseWriter, req *http.Request, p htt
...
+	namespace := p.ByName("namespace")
+	resourceType := p.ByName("resource-type")
+	resourceName := p.ByName("resource-name")
+
+	var resource interface{}
+	var err error
+	options := metav1.GetOptions{}
+	switch resourceType {
+	case k8s.CronJob:
+		resource, err = h.k8sAPI.BatchV1beta1().CronJobs(namespace).Get(resourceName, options)
```
tegioz: Added CronJob support too in preparation for #3614 @scottcarol

web/app/js/components/ResourceDetail.jsx
cynthia-sg: Yes, I added it only to avoid linter error due to `resourceDefinition` is not being used anywhere.

web/app/js/components/ResourceDetail.jsx
```diff
@@ -149,9 +150,14 @@ export class ResourceDetailBase extends React.Component {
         // downstream resources of this resource (meshed traffic only)
         this.api.fetchMetrics(
           `${this.api.urlsForResource("all")}&from_name=${resource.name}&from_type=${resource.type}&from_namespace=${resource.namespace}`
-        )
+        ),
       ];
 
+    if (!this.state.queryForDefinition) {
```
scottcarol: I think this should check if `queryForDefinition` equals `true`. If it's `false` we don't need to fetch the definition. Same with line 178.

web/app/js/components/ResourceDetail.jsx
```diff
@@ -227,13 +245,15 @@ export class ResourceDetailBase extends React.Component {
...
+          resourceDefinition, // eslint-disable-line react/no-unused-state
           lastMetricReceivedTime,
           isTcpOnly,
           isTrafficSplit,
           loaded: true,
           pendingRequests: false,
           error: null,
-          unmeshedSources: this.unmeshedSources // in place of debouncing, just update this when we update the rest of the state
+          unmeshedSources: this.unmeshedSources, // in place of debouncing, just update this when we update the rest of the state
+          queryForDefinition: true,
```
scottcarol: the value of `queryForDefinition` shouldn't have changed here from the initial state, should it? do we need this line?

web/app/js/components/ResourceDetail.jsx
```diff
@@ -161,12 +167,24 @@ export class ResourceDetailBase extends React.Component {
...
-      .then(([resourceRsp, podListRsp, podMetricsRsp, upstreamRsp, downstreamRsp, edgesRsp]) => {
+      .then(results => {
+        const [resourceRsp, podListRsp, podMetricsRsp, upstreamRsp, downstreamRsp, ...rsp] = [...results];
         let resourceMetrics = processSingleResourceRollup(resourceRsp, resource.type);
         let podMetrics = processSingleResourceRollup(podMetricsRsp, resource.type);
         let upstreamMetrics = processMultiResourceRollup(upstreamRsp, resource.type);
         let downstreamMetrics = processMultiResourceRollup(downstreamRsp, resource.type);
-        let edges = processEdges(edgesRsp, this.state.resource.name);
+
+        let resourceDefinition = undefined;
```
scottcarol: I'm a little uncomfortable setting the variable to `undefined` here, since if `queryForDefinition` is `false`, it will never be defined. Can we set `resourceDefinition` to `null` up in the `getInitialState()` function and avoid this line?

web/app/js/components/ResourceDetail.jsx
```diff
@@ -84,7 +84,8 @@ export class ResourceDetailBase extends React.Component {
       resourceIsMeshed: true,
       pendingRequests: false,
       loaded: false,
-      error: null
+      error: null,
+      queryForDefinition: false,
```
scottcarol: could you add a note above this line that explains that we are setting the variable to false now because we are not displaying the resource definition, but that this can change in the future?

scottcarol:
This PR is looking good 👍 I made a few comments but I think this is almost ready to merge. @adleong would you mind reviewing the Go changes?

web/app/js/components/ResourceDetail.jsx
cynthia-sg: Done

web/app/js/components/ResourceDetail.jsx
cynthia-sg: Done

web/app/js/components/ResourceDetail.jsx
cynthia-sg: Not, I'll remove the line

scottcarol:
Hi @cynthia-sg @tegioz can you address the merge conflicts in this PR? Thanks!

tegioz:
> Hi @cynthia-sg @tegioz can you address the merge conflicts in this PR? Thanks!

Done!

tegioz:
Thanks @scottcarol. I've just rebased from master and pushed again.

web/srv/api_handlers.go
```diff
@@ -358,3 +360,46 @@ func (h *handler) handleAPICheck(w http.ResponseWriter, req *http.Request, p htt
...
+
+func (h *handler) handleAPIResourceDefinition(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
+	namespace := p.ByName("namespace")
+	resourceType := p.ByName("resource-type")
+	resourceName := p.ByName("resource-name")
+
+	var resource interface{}
+	var err error
+	options := metav1.GetOptions{}
+	switch resourceType {
```
adleong: should we be able to fetch ServiceProfiles also?

web/srv/server.go
```diff
@@ -173,6 +173,7 @@ func NewServer(
 	server.router.GET("/api/routes", handler.handleAPITopRoutes)
 	server.router.GET("/api/edges", handler.handleAPIEdges)
 	server.router.GET("/api/check", handler.handleAPICheck)
+	server.router.GET("/api/namespaces/:namespace/:resource-type/:resource-name", handler.handleAPIResourceDefinition)
```
adleong: This API where we simply can look up resources by name seems good but it's hard to know whether it will be sufficient without knowing the use-cases.

For example, when visiting a service detail page, we might also want to look up the service profile for that service.  And possibly all the service profiles for that service from other namespaces.  Or we may want to look up all traffic splits which have this service as an apex or leaf.  Such lookups would require a more specialized API.

dadjeibaah: We probably should change this URL scheme so it is consistent with the other API paths. So it would look something like this:
```diff
diff --git a/web/srv/server.go b/web/srv/server.go
index 9f0c05b1..dff692ff 100644
--- a/web/srv/server.go
+++ b/web/srv/server.go
@@ -173,7 +173,7 @@ func NewServer(
 	server.router.GET("/api/routes", handler.handleAPITopRoutes)
 	server.router.GET("/api/edges", handler.handleAPIEdges)
 	server.router.GET("/api/check", handler.handleAPICheck)
-	server.router.GET("/api/namespaces/:namespace/:resource-type/:resource-name", handler.handleAPIResourceDefinition)
+	server.router.GET("/api/resource-definitions", handler.handleAPIResourceDefinition)
 
 	// grafana proxy
 	server.router.DELETE("/grafana/*grafanapath", handler.handleGrafana)
```

This would mean we would use query parameters to pass resource information to the handler. See full diff above.

web/srv/api_handlers.go
```diff
@@ -358,3 +360,46 @@ func (h *handler) handleAPICheck(w http.ResponseWriter, req *http.Request, p htt
 		"results": results,
 	})
 }
+
+func (h *handler) handleAPIResourceDefinition(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
+	namespace := p.ByName("namespace")
```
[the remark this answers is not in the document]
dadjeibaah: We also should add in some validation that captures the case where we might supply empty strings in the params.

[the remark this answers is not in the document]
tegioz: Hi @dadjeibaah,

Thanks for your feedback :)

There is a reasoning behind it, yes. IMHO the route using the path params does a better job explicitly identifying the concrete resource. Parameters specified in the path are not optional and this route makes that clear to the reader without requiring to dig deeper in the code, making it easier to understand how to use the API. To me, query parameters would be a better fit for filtering, sorting or any other *optional* parameters. In the other API routes, query parameters are used mostly for filtering, not to identify a particular resource, so they are a great match. I didn't see it as a consistency issue as to me they were different use cases. And as a bonus, we avoid the extra validation required for empty parameters as shown in the diff.

There are quite a few routes using path parameters in [server.go](https://github.com/linkerd/linkerd2/blob/65d5778b9393cd41c001b5023cf02a0d7626d03c/web/srv/server.go#L125-L152) so I thought it'd be ok to use it if the use case was appropriate.

Said that, I'm happy to change it if you still think it's more correct the other way.

Thanks!

grampelberg:
@adleong I want to have access to the entire resource's YAML (status and spec) for the detail pages. I'll probably want it for the list views as well but don't have anything as concrete there yet.

web/srv/api_handlers.go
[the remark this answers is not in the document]
dadjeibaah: @tegioz, thanks for the explanation, I totally agree that using path parameters makes more sense here and that we can save an extra step in validation by using path parameters instead. I think if we were redesigning the URL scheme for the dashboard, this would be a great opportunity to also change all other `handleAPI...` methods to use path parameters as well. 

However, my biggest concern is that since all `webapp` routes are path parameters and all API related paths use query parameters, we could potentially deviate from this pattern which may lead to more confusion when refactoring/adding to `server.go` down the road. IMO, I think we should keep things consistent and use the query param style as suggested.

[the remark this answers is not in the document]
tegioz: No problem, done 👍

web/srv/server.go
tegioz: API has been updated to `/api/resource-definition` and uses now query string params instead of path params 👍

PR #3667, opened 2019-11-01
Add filter to namespace select button

### LINKED ISSUE #3610: Allow user to filter list of namespaces in dashboard selection menu
*Problem*: If a user's cluster has more than 5-6 namespaces, it may be time-consuming to find the desired namespace after opening the namespace selection menu.

Screen Shot 2019-10-21 at 11 26 42 AM

*Solution*: Add a text filter to the namespace selection dropdown, as the [Kubernetes dashboard does](https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/). 

Below screenshots show functionality only - pick styles, sizing, etc. to fit in with the existing Linkerd dashboard.

1. 
Screen Shot 2019-10-21 at 10 57 48 AM

2. 
Screen Shot 2019-10-21 at 10 58 12 AM

scottcarol:
Fixes #3610 

Adds a text input to the namespace selection button of the dashboard to allow users to filter the list of namespaces.

Moved some filter logic to `Utils.js` since it is now used by both `BaseTable` and `Navigation` components.

Fixed a bug where using `*` as a wildcard in the filter input, as in `li*kerd`, was not returning `linkerd` as a match.

Namespace select button before:
Screen Shot 2019-11-01 at 1 04 50 PM

Namespace select button after:
Screen Shot 2019-11-01 at 1 03 59 PM
Screen Shot 2019-11-01 at 1 04 09 PM

### DIFF
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -75,7 +75,7 @@ class BaseTable extends React.Component {
   };
 
   handleFilterInputChange = event => {
-    this.setState({ filterBy: event.target.value });
+    this.setState({ filterBy: regexFilterString(event.target.value) });
   }
 
   handleFilterToggle = () => {
@@ -95,11 +95,12 @@ class BaseTable extends React.Component {
       let filteredRows = rows.filter(row => {
         return columnsToFilter.some(col => {
           let rowText = col.filter(row);
-          return rowText.match(regexFilterString(filterBy));
+          return rowText.match(filterBy);
         });
       });
       rows = filteredRows;
     }
+
     return rows;
   }
 
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -20,6 +20,7 @@ import _get from 'lodash/get';
 import _isNil from 'lodash/isNil';
 import _orderBy from 'lodash/orderBy';
 import classNames from 'classnames';
+import { regexFilterString } from './util/Utils.js';
 import { withStyles } from '@material-ui/core/styles';
 
 const styles = theme => ({
@@ -74,9 +75,7 @@ class BaseTable extends React.Component {
   };
 
   handleFilterInputChange = e => {
-    let input = e.target.value.replace(/[^A-Z0-9/.\-_]/gi, "").toLowerCase();
-    let swapWildCard = /[*]/g; // replace "*" in input with wildcard
-    let filterBy = new RegExp(input.replace(swapWildCard, ".+"), "i");
+    let filterBy = regexFilterString(e.target.value);
     if (filterBy !== this.state.filterBy) {
       this.setState({ filterBy });
     }
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -90,15 +90,16 @@ class BaseTable extends React.Component {
     if (orderBy && col.sorter) {
       rows = _orderBy(rows, row => col.sorter(row), order);
     }
-    let columnsToFilter = tableColumns.filter(col => col.filter);
-    let filteredRows = rows.filter(row => {
-      return columnsToFilter.some(col => {
-        let rowText = col.filter(row);
-        return rowText.match(regexFilterString(filterBy));
+    if (filterBy) {
+      let columnsToFilter = tableColumns.filter(col => col.filter);
+      let filteredRows = rows.filter(row => {
+        return columnsToFilter.some(col => {
+          let rowText = col.filter(row);
+          return rowText.match(regexFilterString(filterBy));
+        });
       });
-    });
-    rows = filteredRows;
-
+      rows = filteredRows;
+    }
     return rows;
   }
 
--- a/web/app/js/components/BaseTable.jsx
+++ b/web/app/js/components/BaseTable.jsx
@@ -74,11 +74,8 @@ class BaseTable extends React.Component {
     this.setState({ order, orderBy });
   };
 
-  handleFilterInputChange = e => {
-    let filterBy = regexFilterString(e.target.value);
-    if (filterBy !== this.state.filterBy) {
-      this.setState({ filterBy });
-    }
+  handleFilterInputChange = event => {
+    this.setState({ filterBy: event.target.value });
   }
 
   handleFilterToggle = () => {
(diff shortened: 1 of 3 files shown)


scottcarol:
Note: This is going to conflict with #3646 - whichever merges first, I'll update the unmerged branch to incorporate those changes.

web/app/js/components/util/Utils.js
```diff
@@ -115,6 +115,15 @@ export const toClassName = name => {
...
 };
 
+/*
+ Create regex string from user input for a filter
+*/
+export const regexFilterString = input => {
+  input = input.replace(/[^A-Z0-9/.\-_*]/gi, "").toLowerCase();
+  let swapWildCard = /[*]/g; // replace "*" in input with wildcard
+  return new RegExp(input.replace(swapWildCard, ".+"), "i");
+};
```
cpretzer: Nice work making this reusable! 👍

web/app/js/components/Navigation.jsx
```diff
@@ -356,7 +368,12 @@ class NavigationBase extends React.Component {
 
   render() {
     const { api, classes, selectedNamespace, ChildComponent, ...otherProps } = this.props;
-    const { namespaces, anchorEl, showNamespaceChangeDialog, newNamespace, mobileSidebarOpen } = this.state;
+    let { namespaces, namespaceFilter, namespaceFilterInput, anchorEl, showNamespaceChangeDialog, newNamespace, mobileSidebarOpen } = this.state;
```
grampelberg: This is totally a style thing, but I like to keep everything under 80 chars so that it is readable. Looks like there are some jsfmt tools that work just like gofmt to do that for you.

web/app/js/components/util/Utils.js
```diff
@@ -115,6 +115,15 @@ export const toClassName = name => {
   return _lowerCase(name).replace(/[^a-zA-Z0-9]/g, "_");
 };
 
+/*
+ Create regex string from user input for a filter
+*/
+export const regexFilterString = input => {
+  input = input.replace(/[^A-Z0-9/.\-_*]/gi, "").toLowerCase();
```
grampelberg: I'm not sure everywhere this is being used, but wouldn't it be easier to do something like:

```
RegExp.escape = text => text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');

return new RegExp(`^.*${RegExp.escape(input)}.*`, "i");
```

web/app/js/components/Navigation.jsx
```diff
@@ -356,7 +368,12 @@ class NavigationBase extends React.Component {
...
+    if (namespaceFilter !== '') {
```
grampelberg: With my proposed change to the filter regex, you wouldn't need this if statement anymore.

web/app/js/components/Navigation.jsx
```diff
@@ -297,6 +302,13 @@ class NavigationBase extends React.Component {
     this.props.history.push(`/namespaces/${this.state.newNamespace}`);
   }
 
+  handleFilterInputChange = e => {
+    let namespaceFilter = regexFilterString(e.target.value);
```
grampelberg: I believe that if you just store the value itself in here, you don't need `namespaceFilterInput` or the `if` statement further down. I assume you're doing this to reduce regex compile time?

web/app/js/components/Navigation.jsx
scottcarol: Good point - I use VSCode to softwrap my JS at 80 chars so didn't realize these lines were getting so long. Will make a ticket to update systematically.

web/app/js/components/Navigation.jsx
scottcarol: Good call!

web/app/js/components/util/Utils.js
scottcarol: I incorporated some of this Regex feedback and will create a new ticket to test and possibly improve the Regex even more.

web/app/js/components/Navigation.jsx
```diff
@@ -401,14 +413,24 @@ class NavigationBase extends React.Component {
             open={this.state.namespaceMenuOpen}
             keepMounted
             onClose={this.handleNamespaceMenuClick}>
+
+            <MenuItem>
+              <InputBase
+                id="namespace-filter-textfield"
+                value={namespaceFilter}
```
cynthia-sg: What do you think about highlighting the selected namespace in the menu? Perhaps, from the user experience point of view, is useful.

```
<MenuItem
    selected={formattedNamespaceName === "All Namespaces"}
    value="all"
    onClick={e => this.handleNamespaceChange(e, "_all")}>
    All Namespaces
</MenuItem>

{namespaces.map(ns => (
    <MenuItem
    selected={ns.name === formattedNamespaceName}
    onClick={e => this.handleNamespaceChange(e, ns.name)}
    key={ns.name}>
        {ns.name}
    </MenuItem>
))}
```

menu

web/app/js/components/Navigation.jsx
```diff
@@ -356,7 +364,11 @@ class NavigationBase extends React.Component {
 
   render() {
     const { api, classes, selectedNamespace, ChildComponent, ...otherProps } = this.props;
-    const { namespaces, anchorEl, showNamespaceChangeDialog, newNamespace, mobileSidebarOpen } = this.state;
+    let { namespaces, namespaceFilter, anchorEl, showNamespaceChangeDialog,
+      newNamespace, mobileSidebarOpen } = this.state;
+    namespaces = namespaces.filter(ns => {
+      return ns.name.match(regexFilterString(namespaceFilter));
```
cynthia-sg: I think it would be better to avoid calling `regexFilterString` into the `filter` method:

```
const formattedNamespaceFilter = regexFilterString(namespaceFilter);
namespaces = namespaces.filter(ns => {
    return ns.name.match(formattedNamespaceFilter);
});
```

... or much better, call `regexFilterString` only to change the input and save the `formatted` namespaceFilter version into the state. This way we avoid building the regular expression multiple times.

```
handleFilterInputChange = event => {
    this.setState({ namespaceFilter: event.target.value,  formattedNamespaceFilter: regexFilterString(event.target.value)});
}
```

web/app/js/components/BaseTable.jsx
```diff
@@ -99,12 +95,11 @@ class BaseTable extends React.Component {
       let filteredRows = rows.filter(row => {
         return columnsToFilter.some(col => {
           let rowText = col.filter(row);
-          return rowText.match(filterBy);
+          return rowText.match(regexFilterString(filterBy));
```
cynthia-sg: Please, check my comment below.

web/app/js/components/Navigation.jsx
scottcarol: Thanks for the idea! Having a separate `formattedNamespace` variable was actually part of my original PR but I omitted it after some review feedback. I like your solution of calculating the regex when saving it into the state, and have restored the variable. 👍

web/app/js/components/Navigation.jsx
scottcarol: Interesting idea -- I'm going to leave as is for now, but if the button is redesigned in the future this may be good to suggest.

PR #3668, opened 2019-11-02
Update uninject command to handle namespaces

### LINKED ISSUE #3648: Update uninject command to handle namespaces
Following https://github.com/linkerd/linkerd2/pull/3607, the `linkerd inject` command can now annotate namespaces with the `linkerd.io/inject: enabled` annotation. The issue involves updating the `linkerd uninject` command to parse namespace YAML, and remove the `linkerd.io/inject: enabled` annotation, if present.

cc @mayankshah1607 @alpeb

mayankshah1607:
Fixes #3648
* Add support for `uninject` command to uninject namespace configs
* Add relevant unit tests in `cli/cmd/uninject_test.go`


pkg/inject/uninject.go
```diff
@@ -58,6 +74,18 @@ func (conf *ResourceConfig) uninjectPodSpec(report *Report) {
 	t.Volumes = volumes
 }
 
+func uninjectNamespace(ns *v1.Namespace, report *Report) {
```
Pothulapati: Hey, Can we directly use the `uninjectObjectMeta` ,method and just pass in `ns.ObjectMeta`, Any Problem doing that way?

Because, the already defined function, even removes the proxy config labels and also helps us re-use code.

pkg/inject/uninject.go
```diff
@@ -1,8 +1,11 @@
 package inject
 
 import (
+	"errors"
 	"strings"
 
+	"sigs.k8s.io/yaml"
+
```
Pothulapati: removing this line, would just sort out the pkg's in alphabetical order.

pkg/inject/uninject.go
zaharidichev: +1 For reusing the uninjectObjectMeta function if possible

mayankshah1607: @Pothulapati @zaharidichev  thanks for your suggestion! Initially, I was under the impression that we'd necessarily have to typecast to `v1.Namespace` to get the annotations from `conf`. I was able to simply pass the namespace `conf` as `con thanks for your suggestion! Initially, I was under the impression that we'd necessarily have to typecast to `v1.Namespace` to get the annotations from `conf`. I was able to simply pass the namespace `conf` as `conf.workload.Meta` (i.e without having to perform typecasting) to `uninjectObjectMeta()` and carry out the uninject process as expected! This also helped me reduce the number of lines of code

pkg/inject/uninject.go
```diff
@@ -11,6 +14,19 @@ import (
 // Uninject removes from the workload in conf the init and proxy containers,
 // the TLS volumes and the extra annotations/labels that were added
 func (conf *ResourceConfig) Uninject(report *Report) ([]byte, error) {
+	if conf.IsNamespace() {
+		ns, ok := conf.workload.obj.(*v1.Namespace)
+		if !ok {
+			return nil, errors.New("can't uninject namespace. Type assertion failed")
+		}
+		uninjectNamespace(ns, report)
```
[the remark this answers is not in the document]
mayankshah1607: Done!

pkg/inject/uninject.go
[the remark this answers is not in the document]
mayankshah1607: Done. Check the latest commit

PR #3675, opened 2019-11-05
Fixed bad identity string for target pod in tap

### LINKED ISSUE #3506: Tap not returning any data, missing content-type header in logs
### What is the issue?

Using the tap CLI (or via dashboard) seems to return no data for any deployment. 
For example when running:

`linkerd -n linkerd tap deploy/linkerd-controller`

The process just sits there and does not print anything. This is the same for deployments in the default namespace also.

Everything else seems to work as expected.

### How can it be reproduced?

`linkerd -n linkerd tap deploy/linkerd-controller`

### Logs, error output, etc

```
linkerd-tap-7f9f9c9fcf-ljc6f tap time="2019-10-01T15:09:41Z" level=info msg="Tapping 1 pods for target: {Namespace:linkerd Type:deployment Name:linkerd-controller XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}"
linkerd-tap-7f9f9c9fcf-ljc6f tap time="2019-10-01T15:09:41Z" level=info msg="Establishing tap on 10.88.9.37:4190"
linkerd-tap-7f9f9c9fcf-ljc6f tap time="2019-10-01T15:09:44Z" level=error msg="[10.88.9.37] encountered an error: rpc error: code = Unavailable desc = Service Unavailable: HTTP status code 503; transport: missing content-type field"
linkerd-tap-7f9f9c9fcf-ljc6f linkerd-proxy WARN [  5732.041959s] proxy={server=out listen=127.0.0.1:4140 remote=10.88.7.171:34880} linkerd2_proxy::app::errors request aborted because it reached the configured dispatch deadline
```

#### `linkerd check` output

```text
$ linkerd check
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-config
--------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane self-check
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √
```

### Environment

- Kubernetes Version: 1.13.2
- Cluster Environment: Openstack
- Host OS: RedHat Enterprise 7.5
- Linkerd version: 19.9.5 (edge)

### Additional context

* selinux disabled
* tried deleting the tap pods, and letting them recreate - no difference

alpeb:
Fixes #3506

Was using the cluster domain instead of the trust domain, which results
in an error when those domains differ.

### DIFF
--- a/controller/tap/server.go
+++ b/controller/tap/server.go
@@ -36,7 +36,6 @@ type GRPCTapServer struct {
 	tapPort             uint
 	k8sAPI              *k8s.API
 	controllerNamespace string
-	clusterDomain       string
 	trustDomain         string
 }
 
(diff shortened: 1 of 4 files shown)


controller/tap/server.go
```diff
@@ -37,6 +37,7 @@ type GRPCTapServer struct {
 	k8sAPI              *k8s.API
 	controllerNamespace string
 	clusterDomain       string
```
adleong: is clusterDomain actually used for anything?  does it need to be part of this struct?

alpeb: Good catch :+1: I'll remove that.

PR #3682, opened 2019-11-06
use appsv1 api in identity

### LINKED ISSUE #3681: Linkerd edge fails to run on Kubernetes v1.16.*
## Bug Report

Linkerd install fails running on Kubernetes clusters running `v1.16.2`, because `Deployment` API's were depreciated in 1.16 and #3600 added an API Call using `Extensions` to record events.
@zaharidichev 

### What is the issue?

Linkerd `identity` pod fails to run, as it tries to use the depreciated API.

### How can it be reproduced?

`linkerd install | k apply -f -`

### Logs, error output, etc

The identity pod logs show and falls into a restart loop
```
time="2019-11-06T17:14:49Z" level=fatal msg="Failed to construct k8s event recorder: the server could not find the requested resource"
```

### Environment

- Kubernetes Version: 1.16.2
- Host OS:
- Linkerd version:

### Possible solution

Use the `AppsV1` APIGroup to get deployments like every other call.

Pothulapati:

Fixes #3681 

Uses `appsv1` APIGroup to perform API calls.

@zaharidichev @ihcsim 

Thanks @droidnoob for the find! :)

### DIFF
--- a/charts/linkerd2/templates/identity-rbac.yaml
+++ b/charts/linkerd2/templates/identity-rbac.yaml
@@ -16,7 +16,7 @@ rules:
 - apiGroups: ["authentication.k8s.io"]
   resources: ["tokenreviews"]
   verbs: ["create"]
-- apiGroups: ["extensions"]
+- apiGroups: ["apps"]
   resources: ["deployments"]
   verbs: ["get"]
 - apiGroups: [""]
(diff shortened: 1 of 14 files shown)


zaharidichev:
That's a great find! Did not expect that to happen!

Pothulapati:
Yes, https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

I was expecting we had an integration test with cluster version `1.16`, Maybe we should now

ihcsim:
```
=== PASS: all tests passed
```

PR #3687, opened 2019-11-07
Add CronJobs and ReplicaSets to dashboard and CLI

### LINKED ISSUE #3584: Implement CronJob view

The dashboard does not currently show CronJobs. There should be a section added to the dashboard for this as well as a corresponding grafana dashboard.

TBD

### LINKED ISSUE #3585: Implement ReplicaSet dashboard

It should be possible to see replicasets as well as the deployments that own them. There should be a corresponding grafana dashboard.

TBD

### LINKED ISSUE #3614: Add Cron Jobs and Replica Sets to dashboard

### LINKED ISSUE #3630: "linkerd inject" does not work on cronjobs?
### What is the issue?
Trying to use `linkerd inject` to annotate a cronjob yaml and it  returns "!! no supported object found"

### How can it be reproduced?
Create a cronjob yaml, apply it, then:

`kubectl get cronjobs.batch mssql-backup -o yaml | linkerd inject  -`

### Logs, error output, etc
(apologies for the debugging code in the sample)

```
$ kubectl get cronjobs.batch mssql-backup -o yaml | linkerd inject  -
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  annotations:
      {"apiVersion":"batch/v1beta1","kind":"CronJob","metadata":{"annotations":{},"name":"mssql-backup","namespace":"se"},"spec":{"concurrencyPolicy":"Forbid","failedJobsHistoryLimit":1,"jobTemplate":{"spec":{"template":{"spec":{"containers":[{"args":["echo \"hello world\"; /opt/mssql-tools/bin/sqlcmd -S server -U sa -P password -I -Q \"EXEC sp_msForEachDB 'PRINT ''?'' ';\" | while read dbname; do echo \"*** Backing up $dbname ***\"; /opt/mssql-tools/bin/sqlcmd -S server -U sa -P password -I -Q \"BACKUP DATABASE master TO DISK='/mssql-backup/backup/$dbname/$dbname-$(date +%Y-%m-%d.%s).bak' WITH STATS, COMPRESSION;\" ; done; echo \"*** End of $dbname backup ***\"; echo \"\"; sleep 60; echo \" *** NORMAL TERMINATION *** \";"],"command":["bash","-c"],"image":"microsoft/mssql-tools","name":"mssql-backup"}],"nodeSelector":{"kubernetes.io/os":"linux"},"restartPolicy":"Never","securityContext":{"fsGroup":0,"runAsUser":0}}}}},"schedule":"*/5 * * * *","successfulJobsHistoryLimit":3}}
  creationTimestamp: "2019-10-23T20:03:48Z"
  name: mssql-backup
  namespace: se
  resourceVersion: "2061027"
  selfLink: /apis/batch/v1beta1/namespaces/se/cronjobs/mssql-backup
  uid: 399f4b60-f5d0-11e9-a92d-72a7270cd27e
spec:
  concurrencyPolicy: Forbid
  failedJobsHistoryLimit: 1
  jobTemplate:
    metadata:
      creationTimestamp: null
    spec:
      template:
        metadata:
          creationTimestamp: null
        spec:
          containers:
          - args:
            - echo "hello world"; /opt/mssql-tools/bin/sqlcmd -S server
              -U sa -P password -I -Q "EXEC sp_msForEachDB
              'PRINT ''?'' ';" | while read dbname; do echo "*** Backing up $dbname
              ***"; /opt/mssql-tools/bin/sqlcmd -S server -U sa -P password
              -I -Q "BACKUP DATABASE master TO DISK='/mssql-backup/backup/$dbname/$dbname-$(date
              +%Y-%m-%d.%s).bak' WITH STATS, COMPRESSION;" ; done; echo "*** End of
              $dbname backup ***"; echo ""; sleep 60; echo " *** NORMAL TERMINATION
              *** ";
            command:
            - bash
            - -c
            image: microsoft/mssql-tools
            imagePullPolicy: Always
            name: mssql-backup
            resources: {}
            terminationMessagePath: /dev/termination-log
            terminationMessagePolicy: File
          dnsPolicy: ClusterFirst
          nodeSelector:
            kubernetes.io/os: linux
          restartPolicy: Never
          schedulerName: default-scheduler
          securityContext:
            fsGroup: 0
            runAsUser: 0
          terminationGracePeriodSeconds: 30
  schedule: '*/5 * * * *'
  successfulJobsHistoryLimit: 3
  suspend: false
status:
  active:
  - apiVersion: batch/v1
    kind: Job
    name: mssql-backup-1571861700
    namespace: se
    resourceVersion: "2061025"
    uid: f8bae702-f5d1-11e9-a92d-72a7270cd27e
  lastScheduleTime: "2019-10-23T20:15:00Z"
---

‼ no supported objects found

cronjob "mssql-backup" skipped
```
#### `linkerd check` output

```$ linkerd check
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-config
--------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-api
-----------
√ control plane self-check
√ no invalid service profiles

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √

$ linkerd version
Client version: stable-2.6.0
Server version: stable-2.6.0
```

### Environment

- Kubernetes Version: 1.14.6
- Cluster Environment: AKS
- Host OS: AKS Linux (Ubuntu?)
- Linkerd version: 2.6-stable

### Additional context
This is actually to add the annotation to end up "disable"ing the auto-proxy-inject until we figure out how to ask the sidecar to exit when the main container is done.

tegioz:
CronJobs and ReplicaSets are now visible in the dashboard (please see screenshots attached). A new entry for each workload has been added to the navigation menu. From there it's possible to open the view listing each kind of resource and accessing their details as with any other workload. It's also possible to use the tap/top tools in the dashboard on CronJobs and ReplicaSets.

Some other points addressed: 

- A new Grafana dashboard has been added for each kind of resource
- CLI tools like metrics/tap/top can now work with CronJobs
- Update CLI help to include CronJobs (and ReplicaSets where missing)
- Inject now supports CronJobs
- Add support for CronJobs and ReplicaSets in K8S API
- Allow access to CronJobs in RBAC templates

Closes #3614 
Closes #3630
Closes #3584 
Closes #3585

rs1

rs2

cj1

cj2

tap_cj

top_cj

tap_rs

top_rs

grafana_rs

grafana_cronjob

### DIFF
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -28,7 +28,7 @@ import { withContext } from './util/AppContext.jsx';
 // if there has been no traffic for some time, show a warning
 const showNoTrafficMsgDelayMs = 6000;
 // resource types supported when querying API for edge data
-const edgeDataAvailable = ["daemonset", "deployment", "job", "pod", "replicationcontroller", "statefulset"];
+const edgeDataAvailable = ["cronjob", "daemonset", "deployment", "job", "pod", "replicaset", "replicationcontroller", "statefulset"];
 
 const getResourceFromUrl = (match, pathPrefix) => {
   let resource = {
--- a/web/app/js/components/util/SvgWrappers.jsx
+++ b/web/app/js/components/util/SvgWrappers.jsx
@@ -953,3 +953,82 @@ export const statefulSetIcon = (
     </g>
   </svg>
 );
+
+export const cronJobIcon = (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="24px"
+    height="24px"
+    version="1.1"
+    viewBox="0 0 18.035 17.5">
+    <g fillOpacity="1">
+      <g
+        stroke="none"
+        strokeDasharray="none"
+        strokeMiterlimit="4"
+        strokeWidth="0">
+        <path
+          fill="#757575"
+          strokeOpacity="1"
+          d="M-6.85 4.272a1.12 1.11 0 00-.428.109l-5.852 2.796a1.12 1.11 0 00-.606.753l-1.444 6.282a1.12 1.11 0 00.152.85 1.12 1.11 0 00.064.089l4.05 5.037a1.12 1.11 0 00.876.417l6.496-.001a1.12 1.11 0 00.875-.417l4.049-5.038a1.12 1.11 0 00.216-.939L.152 7.93a1.12 1.11 0 00-.605-.753L-6.307 4.38a1.12 1.11 0 00-.542-.109z"
+          transform="translate(-.993 -1.174) matrix(1.01489 0 0 1.01489 16.902 -2.699)" />
+        <path
+          fill="#fff"
+          fillRule="nonzero"
+          d="M-6.852 3.818a1.181 1.172 0 00-.452.115l-6.18 2.951a1.181 1.172 0 00-.638.795l-1.524 6.63a1.181 1.172 0 00.16.9 1.181 1.172 0 00.067.093l4.276 5.317a1.181 1.172 0 00.924.44h6.858a1.181 1.172 0 00.923-.44L1.837 15.3a1.181 1.172 0 00.228-.99L.54 7.677a1.181 1.172 0 00-.64-.795l-6.178-2.95a1.181 1.172 0 00-.573-.115zm.003.455a1.12 1.11 0 01.542.108l5.853 2.795a1.12 1.11 0 01.606.753l1.446 6.281a1.12 1.11 0 01-.216.94l-4.05 5.037a1.12 1.11 0 01-.875.417l-6.496.001a1.12 1.11 0 01-.875-.417l-4.05-5.037a1.12 1.11 0 01-.064-.088 1.12 1.11 0 01-.152-.851l1.444-6.281a1.12 1.11 0 01.605-.753l5.853-2.797a1.12 1.11 0 01.429-.108z"
+          baselineShift="baseline"
+          color="#000"
+          direction="ltr"
+          display="inline"
+          enableBackground="accumulate"
+          fontFamily="Sans"
+          fontSize="medium"
+          fontStretch="normal"
+          fontStyle="normal"
+          fontVariant="normal"
+          fontWeight="normal"
+          letterSpacing="normal"
+          overflow="visible"
+          textAnchor="start"
+          textDecoration="none"
+          transform="translate(-.993 -1.174) matrix(1.01489 0 0 1.01489 16.902 -2.699)"
+          style={{
+            lineHeight: "normal",
+            InkscapeFontSpecification: "Sans",
+            WebkitTextIndent: "0",
+            textIndent: "0",
+            WebkitTextAlign: "start",
+            textAlign: "start",
+            WebkitTextDecorationLine: "none",
+            textDecorationLine: "none",
+            WebkitTextTransform: "none",
+            textTransform: "none",
+            marker: "none"
+          }}
+          visibility="visible"
+          wordSpacing="normal"
+          writingMode="lr-tb" />
+      </g>
+      <g fill="#fff">
+        <path
+          fillRule="nonzero"
+          stroke="none"
+          strokeDasharray="1.418, 1.418"
+          strokeDashoffset="23.045"
+          strokeLinecap="butt"
+          strokeLinejoin="round"
+          strokeMiterlimit="10"
+          strokeOpacity="1"
+          strokeWidth="0.709"
+          d="M11.673 3.96v2.143h2.202V3.96zm0 3.084v.546c.258-.06.526-.097.803-.097.497 0 .97.106 1.4.295v-.744zm-6.284.033V9.22h2.203V7.077zm3.142 0V9.22h.928c.31-.52.75-.955 1.275-1.258v-.885zm-3.163 3.117v2.143h2.203v-2.143zm3.152 0v2.143h.707c-.17-.411-.265-.86-.265-1.33 0-.28.037-.551.1-.813z"
+          opacity="1"
+          transform="translate(-.993 -1.174) translate(-.578 .775)" />
+        <path
+          strokeWidth="0.32"
+          d="M12.608 7.94a3.21 3.21 0 00-3.2 3.2c0 1.76 1.44 3.2 3.2 3.2 1.76 0 3.2-1.44 3.2-3.2 0-1.76-1.44-3.2-3.2-3.2zm1.344 4.543l-1.664-1.024V9.54h.48v1.664l1.44.864z"
+          clipPath="url(#b)"
+          transform="translate(-.993 -1.174) translate(-.578 .775)" />
+      </g>
+    </g>
+  </svg>
+);
--- a/charts/linkerd2/templates/controller-rbac.yaml
+++ b/charts/linkerd2/templates/controller-rbac.yaml
@@ -15,7 +15,7 @@ rules:
   resources: ["daemonsets", "deployments", "replicasets", "statefulsets"]
   verbs: ["list", "get", "watch"]
 - apiGroups: ["extensions", "batch"]
-  resources: ["jobs"]
+  resources: ["cronjobs", "jobs"]
   verbs: ["list" , "get", "watch"]
 - apiGroups: [""]
   resources: ["pods", "endpoints", "services", "replicationcontrollers", "namespaces"]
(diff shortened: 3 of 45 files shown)


tegioz:
Thanks @adleong!

Sure, I've used these:

**Replica set:**

```yaml
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: rs1
spec:
  replicas: 2
  selector:
    matchLabels:
      app: test1
  template:
    metadata:
      labels:
        app: test1
    spec:
      containers:
      - name: web-server
        image: nginx:latest
```

**Cron job:**

```yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: cronjob1
spec:
  schedule: "*/5 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: web-server
            image: nginx:latest
          restartPolicy: OnFailure
```

**Helper pod:**

( to generate some traffic to the `replicaset` and `cronjob` pods http server)

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: pod1
spec:
  containers:
  - name: busybox
    image: busybox
    args:
    - sleep
    - "1000000"
```

I've applied them injecting this way:

```
bin/build-cli-bin
bin/linkerd inject pod1.yml --manual | kubectl apply -f -
bin/linkerd inject rs1.yml --manual | kubectl apply -f -
bin/linkerd inject cj1.yml --manual | kubectl apply -f -
```

I've also updated the deployments impacted by the changes in my dev environment (`controller/public_api`, `tap`, `proxy-injector` and `grafana`) so that they use images built from this branch. Before doing this I updated some cluster roles (`linkerd-linkerd-controller`, `linkerd-linkerd-tap` and `inkerd-linkerd-proxy-injector`) to include `cronjobs`.

```
DOCKER_TRACE=1 bin/mkube bin/docker-build-controller
DOCKER_TRACE=1 bin/mkube bin/docker-build-grafana
kubectl -n linkerd set image deployment/linkerd-controller public-api=gcr.io/linkerd-io/controller:git-xxx
kubectl -n linkerd set image deployment/linkerd-tap tap=gcr.io/linkerd-io/controller:git-xxx
kubectl -n linkerd set image deployment/linkerd-proxy-injector proxy-injector=gcr.io/linkerd-io/controller:git-xxx
kubectl -n linkerd set image deployment/linkerd-grafana grafana=gcr.io/linkerd-io/grafana:git-xxx
```

Hope this helps. Please let me know if you need anything else.

alpeb:
Up for discussion:

Given injected pods from jobs won't finish because the sidecar will still be running, are we sure we want to add CronJobs to the list of supported workloads?
For frequently ran CronJobs this will leave lots of pods around consuming resources.

Note that in k8s 1.17 this won't be an issue anymore. Would it make sense to delay a bit this feature till then?

grampelberg:
@alpeb we already inject jobs, feels like we should complete the story instead of having it be half done.

alpeb:
Ok it turns out injection through the webhooks always works, regardless of the owning workload, including (Cron)Jobs. OTOH, the CLI `linkerd inject` was not happy with CronJobs, which this addresses, so I'm ok with bringing the CLI up to par :+1:

alpeb:
Even before this change, when a Deployment triggered the creation and injection of a Pod, the event was being associated to the Deployment, bypassing the intermediary ReplicaSet.
When a stand-alone ReplicaSet (with no Deployment owner) triggers the creation and injection of a Pod, the event is being correctly associated to the Replicaset.

And now when a CronJob triggers the creation and injection of a Pod, the event is being associated to the intermediary Job, instead of the CronJob. I believe the original parent (CronJob) is the one that should receive the notification. Would that make sense? CC @grampelberg

grampelberg:
@alpeb it makes sense to me. I'd honestly love it if the event happened on every resource up the chain. I vaguely remember there being a good reason why that didn't make sense for Deployments though ...

alpeb:
@grampelberg I don't remember that reason... What I remember is that we couldn't do that at the Pod level because it hadn't been persisted at that point of the admission workflow.

@tegioz As a first approach, for persisting at the CronJob level, the `GetOwnerKindAndName` function could be expanded to retrieve the `Job`'s parent just like we do for `ReplicaSet`.

Then adding the event at the intermediary parent level might require a little deeper refactoring, having that function return a slice of the ancestry, and modify the callers accordingly. Might be worth exploring in a separate PR.

tegioz:
> @tegioz As a first approach, for persisting at the CronJob level, the `GetOwnerKindAndName` function could be expanded to retrieve the `Job`'s parent just like we do for `ReplicaSet`.

Thanks @alpeb , I will take care of it 👍

alpeb:
@tegioz thanks for the last commit, that worked fine for me :+1: 

I stumbled however on another issue. A few minutes after installing, a the `linkerd-heartbeat` cronjob is triggered. It gets injected with the proxy, and it fails, probably because the main container starts before the proxy, and thus it's unable to reach the network.

To avoid the cronjob to be injected during installation, I think you'll have to add some logic at the beginning of `transform`
https://github.com/linkerd/linkerd2/blob/005bc0ca8ffc45474b18d3c4be76c7804aba1acd/cli/cmd/inject.go#L140
that checks whether we're dealing with the particular case of `linkerd-heartbeat` and if so, just return the original bytes:
```go
return bytes, reports, nil
```
One way of doing that could be refactoring `IsControlPlaneComponent`
https://github.com/linkerd/linkerd2/blob/005bc0ca8ffc45474b18d3c4be76c7804aba1acd/pkg/inject/inject.go#L877-L880
renaming it just `ControlPlaneComponent` and have it return the component's name, so you can check on the calling side whether it's `heartbeat`.

tegioz:
Good catch! I'll look into it, thanks 👍

tegioz:
Hi @alpeb

Do you think adding this annotation `linkerd.io/inject: disabled` to the heartbeat cronjob pod template spec could work? I've tried an installation adding it to the `heartbeat.yaml` template and the pod doesn't seem to be injected anymore. The job is launched and it completes successfully. But I'm not sure if I'm missing something..

Thanks!

alpeb:
@tegioz adding the annotation does indeed do to the trick. However, I think this surfaces a deeper problem when trying to inject during install, that neither does my proposal above address. So I took the liberty of pushing #3774, which clarifies a little the code and fixes this issue as well. If/when that gets merged, you can rebase to bring that fix here.

tegioz:
Awesome, thanks @alpeb!

scottcarol:
Hey @tegioz, checking in on this one - is this PR ready to re-review?

tegioz:
Hi @scottcarol

Yes, it is. I need to rebase once #3774 is merged (more details above), but that shouldn’t cause you any problems while reviewing.

Thanks!

web/app/js/components/util/SvgWrappers.jsx
```diff
@@ -953,3 +953,82 @@ export const statefulSetIcon = (
...
+          transform="translate(-.993 -1.174) translate(-.578 .775)" />
+        <path
+          strokeWidth="0.32"
+          d="M12.608 7.94a3.21 3.21 0 00-3.2 3.2c0 1.76 1.44 3.2 3.2 3.2 1.76 0 3.2-1.44 3.2-3.2 0-1.76-1.44-3.2-3.2-3.2zm1.344 4.543l-1.664-1.024V9.54h.48v1.664l1.44.864z"
+          clipPath="url(#b)"
+          transform="translate(-.993 -1.174) translate(-.578 .775)" />
+      </g>
+    </g>
+  </svg>
+);
```
scottcarol: Mind fixing the newline error?

web/app/js/components/ResourceDetail.jsx
```diff
@@ -28,7 +28,7 @@ import { withContext } from './util/AppContext.jsx';
 // if there has been no traffic for some time, show a warning
 const showNoTrafficMsgDelayMs = 6000;
 // resource types supported when querying API for edge data
-const edgeDataAvailable = ["daemonset", "deployment", "job", "pod", "replicationcontroller", "statefulset"];
+const edgeDataAvailable = ["daemonset", "deployment", "job", "pod", "replicationcontroller", "statefulset", "replicaset", "cronjob"];
```
scottcarol: This is a tiny nit, but in general in the JS code we have the resources alphabetized for readability - could you alphabetize these as well?

alpeb:
Hi @tegioz, you may rebase this one off of master, now that #3774 has been merged :wink:

tegioz:
Thanks @alpeb! Done 👍

PR #3689, opened 2019-11-07
Added check for TapAPI service

### LINKED ISSUE #3462: Add check for Tap APIService

`Linkerd check` is very useful rn as it does comprehensive tests on most of the control-plane integration points. It would be really great to have one for the `Tap APIService` too, as we have been seeing some issues around APIServices on some clusters.

We may check for 

- Use k8s API and check the status of the APIService
- Talk to the APIService and see if it fails

### What do you want to happen? Add any considered drawbacks.

.. to make Linkerd check more awesome?

Users can use the check command and also see the status of the APIService

droidnoob:
Fixes #3462
Added a checker using `kube-aggregator` client
@Pothulapati

Signed-off-by: Ananthakrishnan <kannan4mi3@gmail.com>

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1015,6 +1015,14 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkAPIService(linkerdTapAPIServiceName)
 					},
 				},
+				{
+					description: "tap api service is running",
+					hintAnchor:  "#",
+					warning:     true,
+					check: func(ctx context.Context) error {
+						return hc.checkAPIService(linkerdTapAPIServiceName)
+					},
+				},
 			},
 		},
 		{
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1812,14 +1812,16 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 	}
 
 	available := false
+	var errorMessage string
 	for _, condition := range apiStatus.Status.Conditions {
 		if condition.Status == "True" {
 			available = true
 			break
 		}
+		errorMessage = condition.Message
 	}
 	if !available {
-		return fmt.Errorf("%s service is not available", serviceName)
+		return fmt.Errorf(errorMessage)
 	}
 	return nil
 }
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1837,7 +1837,7 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 			}
 			errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
 		}
-		errorMessage = fmt.Sprintf("%s:%s", condition.Reason, condition.Message)
+		errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
 	}
 	if available {
 		return errors.New(errorMessage)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1009,7 +1009,7 @@ func (hc *HealthChecker) allCategories() []category {
 				},
 				{
 					description: "tap api service is running",
-					hintAnchor:  "#",
+					hintAnchor:  "l5d-tap-api",
 					warning:     true,
 					check: func(ctx context.Context) error {
 						return hc.checkAPIService(linkerdTapAPIServiceName)
@@ -1806,9 +1806,13 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 		return err
 	}
 
-	_, err = apiServiceClient.APIServices().Get(serviceName, metav1.GetOptions{})
-	if err != nil {
+	apiStatus, err := apiServiceClient.APIServices().Get(serviceName, metav1.GetOptions{})
+	if err != nil && len(apiStatus.Status.Conditions) == 0 {
 		return err
+	}
+
+	if apiStatus.Status.Conditions[0].Status != "True" {
+		return fmt.Errorf("%s service is not available", serviceName)
 	}
 	return nil
 }
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1017,7 +1017,7 @@ func (hc *HealthChecker) allCategories() []category {
 				},
 				{
 					description: "tap api service is running",
-					hintAnchor:  "#",
+					hintAnchor:  "l5d-tap-api",
 					warning:     true,
 					check: func(ctx context.Context) error {
 						return hc.checkAPIService(linkerdTapAPIServiceName)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1811,19 +1811,13 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 		return err
 	}
 
-	available := false
-	var errorMessage string
 	for _, condition := range apiStatus.Status.Conditions {
 		if condition.Type == "Available" {
-			available = true
 			if condition.Status == "True" {
 				return nil
 			}
-			errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
+			return fmt.Errorf("%s: %s", condition.Reason, condition.Message)
 		}
-	}
-	if available {
-		return errors.New(errorMessage)
 	}
 
 	return fmt.Errorf("%s service not available", linkerdTapAPIServiceName)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1818,7 +1818,7 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 			available = true
 			break
 		}
-		errorMessage = condition.Message
+		errorMessage = fmt.Sprintf("%s:%s", condition.Reason, condition.Message)
 	}
 	if !available {
 		return fmt.Errorf(errorMessage)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1826,7 +1826,7 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 			available = true
 			break
 		}
-		errorMessage = fmt.Sprintf("%s:%s", condition.Reason, condition.Message)
+		errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
 	}
 	if !available {
 		return fmt.Errorf(errorMessage)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1015,22 +1015,6 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkAPIService(linkerdTapAPIServiceName)
 					},
 				},
-				{
-					description: "tap api service is running",
-					hintAnchor:  "l5d-tap-api",
-					warning:     true,
-					check: func(ctx context.Context) error {
-						return hc.checkAPIService(linkerdTapAPIServiceName)
-					},
-				},
-				{
-					description: "tap api service is running",
-					hintAnchor:  "l5d-tap-api",
-					warning:     true,
-					check: func(ctx context.Context) error {
-						return hc.checkAPIService(linkerdTapAPIServiceName)
-					},
-				},
 			},
 		},
 		{
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1837,6 +1837,7 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 			}
 			errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
 		}
+		errorMessage = condition.Message
 	}
 	if available {
 		return errors.New(errorMessage)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1837,7 +1837,7 @@ func (hc *HealthChecker) checkAPIService(serviceName string) error {
 			}
 			errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
 		}
-		errorMessage = condition.Message
+		errorMessage = fmt.Sprintf("%s:%s", condition.Reason, condition.Message)
 	}
 	if available {
 		return errors.New(errorMessage)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -34,6 +34,7 @@ import (
 	yamlDecoder "k8s.io/apimachinery/pkg/util/yaml"
 	k8sVersion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
+	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -729,6 +735,14 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.validateServiceProfiles()
 					},
 				},
+				{
+					description: "tap api service is running",
+					hintAnchor:  "#",
```
adleong: This will require a corresponding change in the website repo to add the hint help.

```
‼ tap api service is running
    apiservices.apiregistration.k8s.io "v1alpha1.tap.linkerd.io" not found
    see https://linkerd.io/checks/## for hints
```

pkg/healthcheck/healthcheck.go
```diff
@@ -1370,6 +1383,19 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
 	return hc.checkCanPerformAction("get", namespace, group, version, resource)
 }
 
+func (hc *HealthChecker) checkAPIService(serviceName string) error {
+	apiServiceClient, err := apiregistrationv1client.NewForConfig(hc.kubeAPI.Config)
+	if err != nil {
+		return err
+	}
+
+	_, err = apiServiceClient.APIServices().Get(serviceName, metav1.GetOptions{})
```
adleong: Would it be helpful to check the `Available` status condition as well?

droidnoob: `Available` status is in `Conditions` of type `[]v1.APIServiceCondition`.  I'm assuming there will be only one element in it. Is that correct?

pkg/healthcheck/healthcheck.go
droidnoob: I don't know what needs to be in the hint help. If you can provide me with the content I'll add a PR to that repo. For now I'll add a hint anchor

pkg/healthcheck/healthcheck.go
adleong: I don't know, but we can always iterate over the list of conditions until we find the `Available` condition.

adleong:
@Pothulapati do you have any suggestions for the hint text?

pkg/healthcheck/healthcheck.go
```diff
@@ -1370,6 +1383,30 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+
+	available := false
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Status == "True" {
+			available = true
+			break
+		}
+	}
+	if !available {
+		return fmt.Errorf("%s service is not available", serviceName)
```
Pothulapati: Hey, The APIStatusCondition also has `Reason` and `Message` as `<Reason>: <Message>`, Can you also return them when its not available?

droidnoob: I've made the changes

Pothulapati:
@adleong Now that the error message explains itself, I don't see what we could mention at the hint website as there could be multiple reasons for the failure, I think we can have a simpler hint explaining the Tap APIService,

Example Failures:
```
‼ tap api service is running
    ServiceNotFound:service/linkerd-tap in "linkerd" is not present
    see https://linkerd.io/checks/#l5d-tap-api for hints
```
Tap APIService is the endpoint for the Linkerd Tap functionality. It requires the underlying tap service to work correctly, and pass health checks.

@ihcsim can also add somethings here.

pkg/healthcheck/healthcheck.go
```diff
@@ -1370,6 +1383,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+	}
+
+	available := false
+	var errorMessage string
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Status == "True" {
+			available = true
+			break
+		}
+		errorMessage = fmt.Sprintf("%s:%s", condition.Reason, condition.Message)
```
Pothulapati: Can we keep this `%s: %s`? to make them similar with other errors.

droidnoob: Yes!! Please check now

droidnoob:
@zaharidichev I'll check it out 👍 
Maybe I'll add it as another PR?

droidnoob:
@zaharidichev @Pothulapati what should be the test `k8sConfig` be? https://github.com/linkerd/linkerd2/blob/master/pkg/healthcheck/healthcheck_test.go#L378

Should it be something similar to this https://github.com/linkerd/linkerd2/blob/015ea9e17aa5673f90b2fe3b397a46aff6e9eba6/charts/linkerd2/templates/tap.yaml

zaharidichev:
@droidnoob hmmm, you can mock the actual apiServiceClient directly I think, and not rely on the config. I think your best bet is to take a look at the `apiServiceClient` implementation. If you want to mock it, I think it will be best to make it a part of `HealthChecker`. 

FYI: The yaml you are looking at is the unrendered chart. You would need a rendered one.

droidnoob:
@zaharidichev  I'm currently swamped with work. So can't do the healthcheck_test right now. I've resolved the merge conflicts. Please review the changes.

pkg/healthcheck/healthcheck.go
```diff
@@ -1361,6 +1374,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+	}
+
+	available := false
+	var errorMessage string
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Status == "True" {
+			available = true
+			break
+		}
+		errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
```
zaharidichev: @droidnoob looking at this again, I am wondering whether this logic is very robust. So as per the comments in the k8s api this `Conditions` list is a list of objects: `APIServiceCondition describes the state of an APIService at a particular point`.  Here you are assuming that if you see a condition with status `True` its all good. But is it? What if there was a condition right after it that was generated at a later point in time and was with status `Unknown`. Moreover, is the order of these statuses in the list guaranteed to be time dependent? 

I think the more sane way to do is to get the condition with the highest `LastTransitionTime` and check that one. @adleong @Pothulapati Does that make sense. I have not spent too much time with the k8s API

droidnoob: @zaharidichev I was following what @adleong suggested me to do. https://github.com/linkerd/linkerd2/pull/3689#discussion_r344300551

@adleong Should there be changes like @zaharidichev mentioned?

adleong: my understanding of this object is that there may be a list of condition objects.  each condition has a type and and status.  I think we should be checking that the condition with type "Available" has status == "True".  In other words, we should iterate through the list of conditions until we find the `Type == Available` condition and check that that condition has `Status == "True"`. 

Does that make sense?

zaharidichev: Yes it makes perfect sense. I was thinking that this is simply a list of conditions that have been transited to ordered in a chronological order. But I am wrong I guess

pkg/healthcheck/healthcheck.go
```diff
@@ -1361,6 +1374,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+	var errorMessage string
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Type == "Available" && condition.Status == "True" {
+			available = true
+			break
+		}
+		errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
+	}
+	if !available {
+		return fmt.Errorf(errorMessage)
```
zaharidichev: By the looks of it we only care about the Available condition indeed. If that is the case, I think we should be looking only for it. It might be the case that this is the only condition that is in there alltogether but we can make our intent clearer by doing something like: 

```
func ensureAvailable(service *v1.APIService) error {
	for _, condition := range service.Status.Conditions {
		if condition.Type == "Available" {
			if condition.Status == "True" {
				return nil
			}
			return fmt.Errorf("%s: %s", condition.Reason, condition.Message)
		}
	}
	return fmt.Errorf("cannot determine availability of %s", service.Name)
}
```

@droidnoob @Pothulapati what do you think?

Pothulapati: Makes total sense :)

droidnoob: 👍

zaharidichev: @droidnoob you might also want to rebase against master and resolve any conflicts before going further. Almost there :)

zaharidichev:
@droidnoob are you still keen on bringing that across the finish line?

droidnoob:
@zaharidichev Yes I am

pkg/healthcheck/healthcheck.go
droidnoob: Looking at this again, wouldn't this return an error if `condition.Status` is not `True` without checking all the conditions?

zaharidichev: Well, yes but arent we interested only in the `Available` condition anyway? We only care whether its status is True. Correct me if I am wrong

droidnoob: There might be multiple `Available` conditions. So we need to check all the conditions till the status is `True`

zaharidichev: Ahh, alright. If that is indeed the case (and it seems so) feel free to ignore my comment.

zaharidichev:
@droidnoob let me know if I can help :)

droidnoob:
> @droidnoob let me know if I can help :)

I think required changes have been made. If there are additional changes please feel free to add
@zaharidichev

pkg/healthcheck/healthcheck.go
```diff
@@ -1582,6 +1595,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
 }
 
+func (hc *HealthChecker) checkAPIService(serviceName string) error {
+	apiServiceClient, err := apiregistrationv1client.NewForConfig(hc.kubeAPI.Config)
+	if err != nil {
+		return err
+	}
+
+	apiStatus, err := apiServiceClient.APIServices().Get(serviceName, metav1.GetOptions{})
+	if err != nil && len(apiStatus.Status.Conditions) == 0 {
```
adleong: What's the purpose of checking the Conditions length here?  If there is any kind of error at all with this Get, shouldn't we return it?

pkg/healthcheck/healthcheck.go
```diff
@@ -1582,6 +1595,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+	}
+
+	available := false
+	var errorMessage string
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Type == "Available" && condition.Status == "True" {
+			available = true
+			break
+		}
+		errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
```
adleong: Should we restructure the above `if` statement so that we only set this errorMessage if we find an "Available" condition with status not equal to "True"?  I don't think we want to set this message if we find conditions of other types.

pkg/healthcheck/healthcheck.go
```diff
@@ -1582,6 +1595,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+		return err
+	}
+
+	apiStatus, err := apiServiceClient.APIServices().Get(serviceName, metav1.GetOptions{})
+	if err != nil && len(apiStatus.Status.Conditions) == 0 {
+		return err
+	}
+
+	available := false
+	var errorMessage string
```
adleong: I think we should have a default errorMessage that will be used if no conditions of type "Available" are found.  As it is written, this would use the nil string as the error message if the conditions list is empty, for example.

pkg/healthcheck/healthcheck.go
```diff
@@ -1582,6 +1595,32 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Type == "Available" && condition.Status == "True" {
+			available = true
+			break
+		}
+		errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
+	}
+	if !available {
+		return fmt.Errorf(errorMessage)
```
adleong: No formatting is happening in this error message so we can use `errors.New` instead of `fmt.Errorf`.

zaharidichev:
@droidnoob can you rebase against master?

droidnoob:
@zaha

> @droidnoob can you rebase against master?

I did rebase against master. I guess the extra commits came after that

zaharidichev:
Hm, probably. I just tried rebasing your branch against the latest master and it gives me a minor conflict in  `pkg/healthcheck/healthcheck.go`. Probably that is what confuses the diff

droidnoob:
> Hm, probably. I just tried rebasing your branch against the latest master and it gives me a minor conflict in `pkg/healthcheck/healthcheck.go`. Probably that is what confuses the diff

I did the rebase again. Forgot to force push

pkg/healthcheck/healthcheck.go
```diff
@@ -1787,6 +1800,35 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+	if err != nil {
+		return err
+	}
+
+	available := false
+	var errorMessage string
+	for _, condition := range apiStatus.Status.Conditions {
+		if condition.Type == "Available" {
+			available = true
+			if condition.Status == "True" {
```
zaharidichev: naming this `available` implies that the service is available, but as we can see it might not be. So why noy something like `availableStatusFound` or smth like that ?

pkg/healthcheck/healthcheck.go
```diff
@@ -1787,6 +1800,35 @@ func (hc *HealthChecker) checkCanGet(namespace, group, version, resource string)
...
+				return nil
+			}
+			errorMessage = fmt.Sprintf("%s: %s", condition.Reason, condition.Message)
```
adleong: If we just return this error immediately here, we don't need the `errorMessage` or `available` variables.

zaharidichev: @adleong I was under the impression that we need to check all conditions as there might be multiple available conditions. See https://github.com/linkerd/linkerd2/pull/3689#discussion_r364840296 @droidnoob, thats correct, right?

droidnoob: @zaharidichev I was under that impression. If there's only going to be one `available` condition then we can return right away.

zaharidichev: Yes I was also under that impression after the comment here https://github.com/linkerd/linkerd2/pull/3689#discussion_r364840296. @adleong yeah ignore that. We can just return right away. Makes total sense

adleong:
It looks like you need to update the expected output in https://github.com/linkerd/linkerd2/blob/master/test/testdata/check.golden to include your new check.

droidnoob:
> It looks like you need to update the expected output in https://github.com/linkerd/linkerd2/blob/master/test/testdata/check.golden to include your new check.

@adleong  I have already added the check https://github.com/linkerd/linkerd2/blob/6d32cf4c7bdf6a297c23f63aaa707c29480d3d60/test/testdata/check.golden

zaharidichev:
@droidnoob The chec output is not in the correct order. I guess you have shuffled things around since the last time. Also  you need to add it to the golden file for the check --proxy comman. Here is a diff to make it easier 

```diff
diff --git a/test/testdata/check.golden b/test/testdata/check.golden
index c147ae45..c22570ce 100644
--- a/test/testdata/check.golden
+++ b/test/testdata/check.golden
@@ -44,9 +44,9 @@ linkerd-api
 -----------
 √ control plane pods are ready
 √ control plane self-check
-√ tap api service is running
 √ [kubernetes] control plane can talk to Kubernetes
 √ [prometheus] control plane can talk to Prometheus
+√ tap api service is running

 linkerd-version
 ---------------
diff --git a/test/testdata/check.proxy.golden b/test/testdata/check.proxy.golden
index 8ef02e41..2e6a1546 100644
--- a/test/testdata/check.proxy.golden
+++ b/test/testdata/check.proxy.golden
@@ -50,6 +50,7 @@ linkerd-api
 √ control plane self-check
 √ [kubernetes] control plane can talk to Kubernetes
 √ [prometheus] control plane can talk to Prometheus
+√ tap api service is running

 linkerd-version
 ---------------
```

In fact its always good to run te integrations tests locally to make sure everything passes

PR #3696, opened 2019-11-11
Add checks for issuer certificate validation

### LINKED ISSUE #3431: Validate certificates

It is way too easy to build the certificates wrong. Either because of a bad algorithm, misformating the actual data or time drift between generation and installation.

Add a `check` that validates the cert can be used and has been generated correctly from the control plane secret itself.

zaharidichev:
This PR adds the following tls related checks to the check cmd: 

The successful output looks as follows. Before I give it a final refactor (as its a bit messy now) and fix all the unit tests, lets agree on he format and the scope of these checks. So opinions on how that looks are more than welcome :) 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
√ issuer cert has valid lifetime

linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
√ all data plane trust roots have valid lifetimes
√ control plane issuer certificate can be verified with data plane trust roots

```

#### linkerd-mtls
These checks are concerned only to the control plane or otherwise put, the certs that are in `linkerd-identity-issuer` and the roots that are in the `IdentityContext.TrustAnchors`

- √ certificate config is valid - this one looks at the config map and the linkerd-identity-issuer secret. It ensured that the secret is there and is in the correct form. Namely if our linkerd config indicates that we are using a secret of type `linkerd.io/tls`, it ensures that we have the corresponding data keys in the secret that are expected for this scheme. 
- √ control plane is using supported trust roots - this check gets the trust roots from the config map and for each root cert (as there might be more than one) it checks the algorithm constraints. We need to have `ECDSA P-256` for the roots. The check is fatal. If any of the root certs in this bundle are not compliant with our constraints, we fail and provide details as to what is wrong.
- √ identity is using supported issuer certificate - this check does the same as the above one but instead looks at the cert that is in the `identity-issuer-secret.` This check is fatal as well.
- √ all control plane trust roots have valid lifetimes - this check looks at all root certs that are in the config map for linkerd and verified that their `NotBefore` and `NotAfter` are fine. The check will fail if any of the roots is expired, listing the first seven bytes of the cert's signature + the subject common name + the reasoning why its not valid (not valid yet OR not valid anymore). If any of the certs is at more than 90% of its lifetime this check will instead simply warn, listing the roots that are close to expiry and issuing an event to the k8s event bus. So whether this check is fatal or not depends on the circumstances. 
- √ issuer cert has valid lifetime - same as above this check ensures that the certificate in the linkerd-identity-issuer is valid time-wise and will error or warn accordingly

#### linkerd-data-plane-mtls
These checks are concerned with all the meshed pods in the cluster. Essentially these validate whether the roots that the proxies of these meshed pods are valid and whether they work with the issuer certificates that the control plane is configured to work with

- √ data plane is using supported trust roots - this one gets all the meshed pods and for every pods checks all root certs that the proxy container is configured with. It ensures that we use the correct signature, public key algorithms and curve type. If any of the roots fail, it will list the root signature, the subject common name and the pods that this cert is residing in. It will also list stats about how many pods are affected (i.e. 5/500 pods, where 500 is the number of meshed pods that we have)
- √ all data plane trust roots have valid lifetimes - much like the control plane tls check, this check looks at all roots for all proxies and issues and produces an error or warning depending on the time validity of the cert. In both cases it will provide a list of roots that are expired/going to expire soon and the pods where they reside in. It will also output stats for the number of affected pods. In case of a warning (cert at more than 90% of lifetime) it will also issue a warning event on the K8s bus. 
- √ control plane issuer certificate can be verified with data plane trust roots - this is probably the check that delivers the ultimate guarantee that thing are ok. Essentially what it does it goes through the roots of all proxies and tries to validate the control plane issuer certificate (stored in `linkerd-identity-issuer`) with the roots. If any of that fails, it produces a list of pods, whose roots cannot verify the issuer certificate and fails. The rationale behind that is simply that we dont really care that much how many certs are in the root bundles of the proxies or whether the env var matches exactly what is in the config map (`IdentityContext.TrustAnchors`), as there will be times (during cert rotation) when these will not match one for one. The thing we care about is that it works. 

Phew... 

Now lets look at how our failure scenarios look like and what information we get out of the check: 

- When we have `linkerd.io/tls` secret type but we don't have the appropriate content in the `linkerd-identity-issuer` map: 

```
linkerd-mtls
------------
× certificate config is valid
    key crt.pem containing the issuer certificate needs to exist in secret linkerd-identity-issuer if --identity-external-issuer=false
    see https://linkerd.io/checks/#l5d-crt-config-is-valid for hints

Status check results are ×
```

- When unsupported trust root is used:

```
linkerd-mtls
------------
√ certificate config is valid
× control plane is using supported trust roots
    Invalid roots:
      * 30818702413131 identity.linkerd.cluster.local must use P-256 curve for public key, instead P-521 was used
    see https://linkerd.io/checks/#l5d-supported-certs-type for hints

Status check results are ×
```

- When unsupported identity issuer cert is used: 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
× identity is using supported issuer certificate
    issuer certificate must use P-256 curve for public key, instead P-521 was used
    see https://linkerd.io/checks/#l5d-supported-certs-type for hints
```

- When (at least one) of the control plane trust root certs is about to expire (warning) (more than 90% of the way there): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
‼ all control plane trust roots have valid lifetimes
    Roots expiring soon:
      * 3044022019354d identity.linkerd.cluster.local will expire on 2019-12-06T10:41:07Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints
√ issuer cert has valid lifetime
```

- When (at least one) of the control plane trust root certs is expired (fatal): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
× all control plane trust roots have valid lifetimes
    Invalid roots:
      * 3044022019354d identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T10:41:07Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```

- When the issuer certificate is about to expire (warning): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
‼ issuer cert has valid lifetime
    issuer certiifcate will expire on 2019-12-06T10:45:14Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints
```

- When the issuer certificate is expired (fatal): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
× issuer cert has valid lifetime
    issuer certiifcate is not valid anymore. Expired on 2019-12-06T10:45:14Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```

- When some of the roots in some of the proxies are not in the correct format (either unparsable or wrong algorithm): 

```
linkerd-data-plane-mtls
-----------------------
× data plane is using supported trust roots
    Invalid roots:
      * Unparsable anchors for pod linkerd-grafana-5fcb5f44b9-z7hqd
      * 308187024200f9 identity.linkerd.cluster.local must use P-256 curve for public key, instead P-521 was used in pod linkerd-tap-768d79f567-qxrsp
    Affected pods: 2/11
    see https://linkerd.io/checks/#l5d-supported-certs-type for hints

Status check results are ×
```

- When certs int he proxies root anchors are about to expire. Obviously that will give information about all the different roots
across all proxies, so it provides quite a lot of visibility into what is actually going on.  

```
linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
‼ all data plane trust roots have valid lifetimes
    Roots expiring soon:
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-controller-844df9c9d5-vxsw2
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-destination-67b9f958d-qm4gx
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-grafana-86c957cc7d-jd4fr
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-identity-6b798cb557-5r4k4
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-prometheus-64c75d7659-kbghf
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-proxy-injector-b5d8d5984-rl2km
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-sp-validator-5fb587cf57-qsqsj
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-tap-7947f75966-h8c4r
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-web-587d69c8b4-p5qsm
    Affected pods: 9/9
    see https://linkerd.io/checks/#l5d-certs-rotation for hints
√ control plane issuer certificate can be verified with data plane trust roots
```

- When certs in the proxy root anchors have expired (fatal): 

```
linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
× all data plane trust roots have valid lifetimes
    Expired roots:
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-controller-844df9c9d5-vxsw2
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-destination-67b9f958d-qm4gx
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-grafana-86c957cc7d-jd4fr
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-identity-6b798cb557-5r4k4
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-prometheus-64c75d7659-kbghf
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-proxy-injector-b5d8d5984-rl2km
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-sp-validator-5fb587cf57-qsqsj
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-tap-7947f75966-h8c4r
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-web-587d69c8b4-p5qsm
    Affected pods: 9/9
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```

- Finally when for some reason everything passes but there are one or more proxies that have roots that just don't work with the issuer certs at all, we can diagnose that as well: 

```
linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
√ all data plane trust roots have valid lifetimes
× control plane issuer certificate can be verified with data plane trust roots
    Roots in pods cannot verify control plane issuer cert:
      * linkerd-grafana-69ff587876-6xqqb
      * linkerd-tap-dfc85d948-r6zlv
    Affected pods: 2/11
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```

Fixes #3431 


pkg/tls/ca.go
```diff
@@ -249,7 +251,12 @@ func createTemplate(
 }
 
 // Window returns the time window for which a certificate should be valid.
-func (v *Validity) Window(t time.Time) (time.Time, time.Time) {
+func (v *Validity) Window() (time.Time, time.Time) {
+	t := v.CreatedAt
+	if t == nil {
+		now := time.Now()
+		t = &now
+	}
```
olix0r: Doesn't this change the semantics of this function substantially? Why?

pkg/issuercerts/issuercerts.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if algo != x509.ECDSA {
+		return nil, fmt.Errorf("the required public key algorithm is %s, instead %s was used", x509.ECDSA, algo)
+	}
+
+	if starts.After(time.Now()) {
+		return nil, fmt.Errorf("certificate not valid before: %s", starts.Format(time.RFC3339))
+	}
+
+	if ends.Before(time.Now()) {
```
olix0r: AFAIU, we're checking the credentials aginst the linkerd CLI's clock. Is that sufficient?

zaharidichev: Might be wrong here, but assuming a --pre check has passed and concluded that there is [no clock skew](https://github.com/linkerd/linkerd2/blob/4b6254b52ef84c42fb72b9473fe9cd7ae0721f8b/pkg/healthcheck/healthcheck.go#L1416), wouldnt it be sufficient?

cli/cmd/install_test.go
```diff
@@ -302,6 +302,44 @@ func TestValidate(t *testing.T) {
...
 
+	t.Run("Validates the issuer certs upon install", func(t *testing.T) {
+
+		testCases := []struct {
+			crtFilePrefix string
+			expectedError string
+		}{
+			{"expired", "failed to verify issuer certs stored on disk: certificate not valid anymore. Expired at: 1990-01-01T01:01:11Z"},
+			{"not-valid-yet", "failed to verify issuer certs stored on disk: certificate not valid before: 2100-01-01T01:00:51Z"},
+			{"wrong-domain", "failed to verify issuer certs stored on disk: invalid credentials: x509: certificate is valid for wrong.linkerd.cluster.local, not identity.linkerd.cluster.local"},
```
adleong: should we have a test case here for when the cert is valid?

pkg/healthcheck/healthcheck_test.go
```diff
@@ -1774,6 +1781,215 @@ data:
...
+				ends:   time.Now().AddDate(1, 0, 0),
+			}
+		}
+
+		if testCase.schemeInConfig == "" && testCase.tlsSecretScheme == "" {
+			testCase.schemeInConfig = k8s.IdentityIssuerSchemeLinkerd
+			testCase.tlsSecretScheme = k8s.IdentityIssuerSchemeLinkerd
+		}
+
+		t.Run(fmt.Sprintf("%d", id), func(t *testing.T) {
```
adleong: should the test name use `testCase.checkDescription`?

pkg/healthcheck/healthcheck_test.go
```diff
@@ -1774,6 +1781,215 @@ data:
...
+			hc := NewHealthChecker([]CategoryID{}, &Options{})
+			hc.ControlPlaneNamespace = "linkerd"
+
+			issuerData := createIssuerData(testCase.certificateDNSName, testCase.lifespan.starts, testCase.lifespan.ends)
+			config := getFakeConfig(testCase.tlsSecretScheme, testCase.schemeInConfig, issuerData)
+
+			/*			o.heartbeatSchedule = fakeHeartbeatSchedule
```
adleong: remove commented out code?

pkg/healthcheck/healthcheck_test.go
```diff
@@ -1774,6 +1781,215 @@ data:
...
+						o.identityOptions.crtPEMFile = filepath.Join("testdata", "crt.pem")
+						o.identityOptions.keyPEMFile = filepath.Join("testdata", "key.pem")
+						o.identityOptions.trustPEMFile = filepath.Join("testdata", "trust-anchors.pem")
+			*/
+
+			if testCase.fileNameToSave != "" {
```
adleong: should this defaulting behavior be moved up to be with the rest of the defaulting behavior outside of the `t.Run`?

pkg/healthcheck/healthcheck_test.go
```diff
@@ -1774,6 +1781,215 @@ data:
...
+				crtFile := path.Join("/Users/zaharidichev/work/linkerd2/testCerts/testData/", testCase.fileNameToSave+"-crt.pem")
```
adleong: this file paths need to be fixed up

pkg/healthcheck/healthcheck_test.go
```diff
@@ -1774,6 +1781,215 @@ data:
...
+				keyFile := path.Join("/Users/zaharidichev/work/linkerd2/testCerts/testData/", testCase.fileNameToSave+"-key.pem")
+				anchors := path.Join("/Users/zaharidichev/work/linkerd2/testCerts/testData/", testCase.fileNameToSave+"-trust-anchors.pem")
+
+				ioutil.WriteFile(crtFile, []byte(issuerData.IssuerCrt), 0777)
```
adleong: actually, I don't even see where these files are used.  is this needed?

pkg/healthcheck/healthcheck_test.go
```diff
@@ -1774,6 +1781,215 @@ data:
...
+			}
+
+			var err error
+			hc.kubeAPI, err = k8s.NewFakeAPI(config...)
+			if err != nil {
+				t.Fatalf("Unexpected error: %q", err)
+			}
+
+			hc.issuerCerts, err = hc.checkIssuerCertsValidity()
+			if !reflect.DeepEqual(err, testCase.expectedErr) {
```
adleong: can you do a string comparison of the `err.Error()` rather than a deep equals?

pkg/issuercerts/issuercerts.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/linkerd/linkerd2/pkg/k8s"
+	consts "github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/linkerd/linkerd2/pkg/tls"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const кeyMissingError = "key %s containing the %s needs to exist in secret %s if --identity-external-issuer=%v"
```
adleong: it looks like this variable name starts with a cyrillic ka instead of a latin k

pkg/healthcheck/healthcheck_test.go
zaharidichev: oh, yes I used the test to generate the test files used in the install cmd tests. Must have left that there by mistake.

pkg/healthcheck/healthcheck_test.go
zaharidichev: Was trying to be consistent with [this](https://github.com/linkerd/linkerd2/blob/master/pkg/healthcheck/healthcheck_test.go#L1757), although it does not make sense to me that the description is used only for code documentation purposes. Should we change it for all tests in this file?

adleong: seems weird to me.  I'd prefer to have the description in there but :shrug: not a big deal

pkg/healthcheck/healthcheck.go
```diff
@@ -1022,6 +1047,46 @@ func (hc *HealthChecker) checkLinkerdConfigConfigMap() (string, *configPb.All, e
...
+func (hc *HealthChecker) checkIssuerCertsValidity() (*tls.Cred, error) {
+	_, configPB, err := FetchLinkerdConfigMap(hc.kubeAPI, hc.ControlPlaneNamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	idctx := configPB.Global.IdentityContext
+	var data *issuercerts.IssuerCertData
+
+	if idctx.Scheme == k8s.IdentityIssuerSchemeLinkerd {
```
adleong: if the scheme property is empty (which will be the case for previous linkerd installations) we should default to assuming the scheme is Linkerd.  E.g. when running `check` on my stable-2.6 cluster I get this error:

```
× issuer certs are valid
    key ca.crt containing the trust anchors needs to exist in secret linkerd-identity-issuer if --identity-external-issuer=true
    see https://linkerd.io/checks/#l5d-issuer-certs-are-valid for hints
```

because I don't have a scheme and we're falling through to the else block here.

zaharidichev: Ahh, very good catch, thanks!

zaharidichev: Added a test for that as well

zaharidichev:
@adleong if you have any ideas about additional checks we can do, feel free to suggest

pkg/issuercerts/issuercerts.go
```diff
@@ -0,0 +1,125 @@
...
+	creds, err := tls.ValidateAndCreateCreds(ic.IssuerCrt, ic.IssuerKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA: %s", err)
+	}
+
+	algo := creds.Certificate.PublicKeyAlgorithm
+	starts := creds.Certificate.NotBefore
+	ends := creds.Certificate.NotAfter
+
+	if algo != x509.ECDSA {
```
ihcsim: It feels like all these check code should be in `pkg/healthcheck`.

zaharidichev:
@ihcsim Thanks a bunch for the review. I did another round of refactoring and shuffled things around a bit as well as addressed some of your comments. Here is some details around my line of thinking: 

> I am a bit hesitant about this struct because of the implicit coupling to the cert-manager's internal representation of a TLS secret. Can we just use the corev1.Secret type?

I do not get the `core1.Secret` type reference. This is just a map of string to bytes. `IssuerData` is the parsed data. How can I use the secret type instead? 

> The LoadIssuerDataFromFiles function can remain in the cli/cmd and just return (string, string, string)

Maybe I cam a junkie for concrete types but I would prefer to use something that describes itself a bit more clearly. But anyway, got rid of the  IssuerData. 

> The FetchIssuerData, FetchExternalIssuerData and BuildCreds functions can go into pkg/healthcheck/issuer_secret.go

So this is where it gets interesting. I wanted to have a new package for a few reasons:
1. This functionality is not really a healthcheck only thing as it is now used for: 
    - Loading and validation during install.
    - Loading and validation when the identity service picks a new issuer cert.
    - I intend to use it for loading and validation during the upgrade process in #3771  (need to rebase against this current one first)
2. If I put all this into the tls package, then this becomes a transitive dependency of proxy-identity, which requires us to copy things such as the k8s package to the proxy docker container build, which brings even more dependencies. Why bloat things up if we can avoid it.

ihcsim:
My main concern with the `IssuerData` type is that it only makes sense in the context of `cert-maanger`. Having it in our code causes an abstraction leak. I see that you have removed it, but that didn't make life easier (sorry) because we are trying to share code between the CLI and healthcheck.

WDYT of this approach of exporting new functions, `NewCredsFromSecret` and `NewCredsFromFiles`, that construct and check the creds? 

(Some of the unit tests in `pkg/healthcheck/healthcheck_test.go` are failing because I think initially, the mock configs generated by `getFakeConfig` doesn't have `secret.Type` defined, so certain code path isn't tested. We can fix them later if you think this is a reasonable approach.)

```diff
diff --git a/pkg/issuercerts/issuer_secret.go b/pkg/issuercerts/issuer_secret.go
index 67576166..c0e01524 100644
--- a/pkg/issuercerts/issuer_secret.go
+++ b/pkg/issuercerts/issuer_secret.go
@@ -126,3 +126,85 @@ func BuildAndCheckCreds(anchors, crt, key, dnsName string) (*tls.Cred, error) {
 
 	return creds, nil
 }
+
+func NewCredsFromSecret(api *k8s.KubernetesAPI, controlPlaneNamespace, dnsName string) (*tls.Cred, *corev1.Secret, error) {
+	secret, err := api.CoreV1().Secrets(controlPlaneNamespace).Get(k8s.IdentityIssuerSecretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var creds *tls.Cred
+	switch secret.Type {
+	case corev1.SecretTypeTLS:
+
+		creds, err = NewCredsFromFiles(k8s.IdentityIssuerTrustAnchorsNameExternal, corev1.TLSCertKey, corev1.TLSPrivateKeyKey, dnsName)
+
+	case k8s.IdentityIssuerSchemeLinkerd:
+		creds, err = NewCredsFromFiles("", k8s.IdentityIssuerCrtName, k8s.IdentityIssuerKeyName, dnsName)
+
+	default:
+		return nil, nil, fmt.Errorf("unsupported secret type: %s", secret.Type)
+	}
+
+	return creds, secret, err
+}
+
+func NewCredsFromFiles(anchorPath, keyPath, crtPath, dnsName string) (*tls.Cred, error) {
+	var (
+		anchors []byte
+		err     error
+	)
+	if anchorPath != "" {
+		anchors, err = ioutil.ReadFile(anchorPath)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	key, err := ioutil.ReadFile(keyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	crt, err := ioutil.ReadFile(crtPath)
+	if err != nil {
+		return nil, err
+	}
+
+	creds, err := tls.ValidateAndCreateCreds(string(crt), string(key))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA: %s", err)
+	}
+
+	algo := creds.Certificate.PublicKeyAlgorithm
+	starts := creds.Certificate.NotBefore
+	ends := creds.Certificate.NotAfter
+
+	if algo != x509.ECDSA {
+		return nil, fmt.Errorf("the required public key algorithm is %s, instead %s was used", x509.ECDSA, algo)
+	}
+
+	if starts.After(time.Now()) {
+		return nil, fmt.Errorf("certificate not valid before: %s", starts.Format(time.RFC3339))
+	}
+
+	if ends.Before(time.Now()) {
+		return nil, fmt.Errorf("certificate not valid anymore. Expired at: %s", ends.Format(time.RFC3339))
+	}
+
+	roots, err := tls.DecodePEMCertPool(string(anchors))
+	if err != nil {
+		return nil, err
+	}
+
+	if err := creds.Verify(roots, dnsName); err != nil {
+		return nil, fmt.Errorf("invalid credentials: %s", err)
+	}
+
+	return creds, nil
+}

diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 7bf2b429..365e34ff 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -1013,32 +1013,25 @@ func (idopts *installIdentityOptions) genValues() (*charts.Identity, error) {
 }
 
 func (idopts *installIdentityOptions) readExternallyManaged() (*charts.Identity, error) {
-
 	kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, 0)
 	if err != nil {
 		return nil, fmt.Errorf("error fetching external issuer config: %s", err)
 	}
 
-	anchors, crt, key, err := issuercerts.FetchExternalIssuerData(kubeAPI, controlPlaneNamespace)
-	if err != nil {
-		return nil, err
-	}
-	_, err = issuercerts.BuildAndCheckCreds(anchors, crt, key, idopts.issuerName())
-
+	_, secret, err := issuercerts.NewCredsFromSecret(kubeAPI, controlPlaneNamespace, idopts.issuerName())
 	if err != nil {
 		return nil, fmt.Errorf("failed to verify issuer certs stored in %s: %s", consts.IdentityIssuerSecretName, err)
 	}
 
 	return &charts.Identity{
 		TrustDomain:     idopts.trustDomain,
-		TrustAnchorsPEM: anchors,
+		TrustAnchorsPEM: string(secret.Data[k8s.IdentityIssuerTrustAnchorsNameExternal]),
 		Issuer: &charts.Issuer{
 			Scheme:             string(corev1.SecretTypeTLS),
 			ClockSkewAllowance: idopts.clockSkewAllowance.String(),
 			IssuanceLifetime:   idopts.issuanceLifetime.String(),
 		},
 	}, nil
-
 }
 
 // readValues attempts to read an issuer configuration from disk
@@ -1046,20 +1039,19 @@ func (idopts *installIdentityOptions) readExternallyManaged() (*charts.Identity,
 //
 // The identity options must have already been validated.
 func (idopts *installIdentityOptions) readValues() (*charts.Identity, error) {
-	anchors, key, crt, err := issuercerts.LoadIssuerDataFromFiles(idopts.keyPEMFile, idopts.crtPEMFile, idopts.trustPEMFile)
+	creds, err := issuercerts.NewCredsFromFiles(idopts.trustPEMFile, idopts.keyPEMFile, idopts.crtPEMFile, idopts.issuerName())
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to verify issuer certs stored on disk: %s", err)
 	}
 
-	creds, err := issuercerts.BuildAndCheckCreds(anchors, crt, key, idopts.issuerName())
-
+	anchors, err := ioutil.ReadFile(idopts.trustPEMFile)
 	if err != nil {
-		return nil, fmt.Errorf("failed to verify issuer certs stored on disk: %s", err)
+		return nil, err
 	}
 
 	return &charts.Identity{
 		TrustDomain:     idopts.trustDomain,
-		TrustAnchorsPEM: anchors,
+		TrustAnchorsPEM: string(anchors),
 		Issuer: &charts.Issuer{
 			Scheme:              consts.IdentityIssuerSchemeLinkerd,
 			ClockSkewAllowance:  idopts.clockSkewAllowance.String(),
diff --git a/pkg/healthcheck/healthcheck.go b/pkg/healthcheck/healthcheck.go
index c18f0b64..b2a7cb7e 100644
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1057,34 +1057,17 @@ func (hc *HealthChecker) checkIssuerCertsNotExpiringTooSoon() error {
 func (hc *HealthChecker) checkIssuerCertsValidity() (*tls.Cred, error) {
 	_, configPB, err := FetchLinkerdConfigMap(hc.kubeAPI, hc.ControlPlaneNamespace)
 	if err != nil {
-		return nil, err
-	}
-
-	idctx := configPB.Global.IdentityContext
-	var anchors, crt, key string
-
-	if idctx.Scheme == "" || idctx.Scheme == k8s.IdentityIssuerSchemeLinkerd {
-		crt, key, err = issuercerts.FetchIssuerData(hc.kubeAPI, hc.ControlPlaneNamespace)
-		anchors = idctx.TrustAnchorsPem
-	} else {
-		anchors, crt, key, err = issuercerts.FetchExternalIssuerData(hc.kubeAPI, hc.ControlPlaneNamespace)
-		// ensure trust anchors in config matches whats in the secret
-		if err == nil && idctx.TrustAnchorsPem != anchors {
-			errFormat := "IdentityContext.TrustAnchorsPem does not match %s in %s"
-			err = fmt.Errorf(errFormat, k8s.IdentityIssuerTrustAnchorsNameExternal, k8s.IdentityIssuerSecretName)
-		}
-	}
-
-	if err != nil {
 		return nil, err
 	}
 
 	issuerDNS := fmt.Sprintf("identity.%s.%s", hc.ControlPlaneNamespace, configPB.Global.IdentityContext.TrustDomain)
-	creds, err := issuercerts.BuildAndCheckCreds(anchors, crt, key, issuerDNS)
-
+	creds, _, err := issuercerts.NewCredsFromSecret(hc.kubeAPI, hc.ControlPlaneNamespace, issuerDNS)
 	if err != nil {
 		return nil, err
 	}
+
 	return creds, nil
 }
 
diff --git a/pkg/healthcheck/healthcheck_test.go b/pkg/healthcheck/healthcheck_test.go
index 97d1742e..812e89c7 100644
--- a/pkg/healthcheck/healthcheck_test.go
+++ b/pkg/healthcheck/healthcheck_test.go
@@ -1792,6 +1792,7 @@ metadata:
 data:
   crt.pem: %s
   key.pem: %s
+type: linkerd.io/tls
 ---
 `, base64.StdEncoding.EncodeToString([]byte(issuerCrt)), base64.StdEncoding.EncodeToString([]byte(issuerKey))))
 	} else {
@@ -1806,6 +1807,7 @@ data:
   ca.crt: %s
   tls.crt: %s
   tls.key: %s
+type: kubernetes.io/tls
 ---
 `, base64.StdEncoding.EncodeToString([]byte(trustAnchors)), base64.StdEncoding.EncodeToString([]byte(issuerCrt)), base64.StdEncoding.EncodeToString([]byte(issuerKey))))
 	}
@@ -1940,6 +1942,7 @@ func TestValidateIssuerCert(t *testing.T) {
 		}
 
 		t.Run(fmt.Sprintf("%d", id), func(t *testing.T) {
+			t.Skip()
 			hc := NewHealthChecker([]CategoryID{}, &Options{})
 			hc.ControlPlaneNamespace = "linkerd"
 
diff --git a/pkg/identity/service.go b/pkg/identity/service.go
index 562e609b..53b52036 100644
--- a/pkg/identity/service.go
+++ b/pkg/identity/service.go
@@ -106,14 +106,7 @@ func (svc *Service) Run(issuerEvent <-chan struct{}, issuerError <-chan error) {
 }
 
 func (svc *Service) loadCredentials() (tls.Issuer, error) {
-
-	key, crt, err := issuercerts.LoadIssuerCrtAndKeyFromFiles(svc.issuerPathKey, svc.issuerPathCrt)
-
-	if err != nil {
-		return nil, fmt.Errorf("failed to read CA from disk: %s", err)
-	}
-	creds, err := issuercerts.BuildAndCheckCreds(svc.trustAnchors, crt, key, svc.expectedName)
-
+	creds, err := issuercerts.NewCredsFromFiles("", svc.issuerPathKey, svc.issuerPathCrt, svc.expectedName)
 	if err != nil {
 		return nil, err
 	}
```

ihcsim:
> . I wanted to have a new package for a few reasons:

That's fine. And yes, `pkg/tls` isn't the right place for it because those are tls and crypto related code. Also, I notice we have been avoiding changing the `Dockerfile-proxy` file, but in general, if it's the right thing to do, I prefer to change it over hacking around it.

zaharidichev:
@ihcsim Thanks for putting that together! I think there is something corrupt with the patch file. I get :

```bash
➜  linkerd2 git:(zd/validate-certs) ✗ git apply ivan.diff
error: corrupt patch at line 87
```

That is when I try to apply it agaisnt the last commit on this branch. Not sre why that is the case. But anywy, looked at the diff itself and might be wrong (hard to read it when not applied), but I think this is not going to work. So the `NewCredsFromSecret` func seems to be using the k8s api to simply fetch the secret and look at its type, right? And then delegate to `NewCredsFromFiles`. That will surely work if your secrets are mounted on disk (identity servie).  Why doing two things at once?  Why fetching the secret AND attempting to read the files from disk? This seems a bit confusing. In a similar fashion: 

```diff
 func (idopts *installIdentityOptions) readExternallyManaged() (*charts.Identity, error) {
-
 	kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, 0)
 	if err != nil {
 		return nil, fmt.Errorf("error fetching external issuer config: %s", err)
 	}
 
-	anchors, crt, key, err := issuercerts.FetchExternalIssuerData(kubeAPI, controlPlaneNamespace)
-	if err != nil {
-		return nil, err
-	}
-	_, err = issuercerts.BuildAndCheckCreds(anchors, crt, key, idopts.issuerName())
-
+	_, secret, err := issuercerts.NewCredsFromSecret(kubeAPI, controlPlaneNamespace, idopts.issuerName())
 	if err != nil {
 		return nil, fmt.Errorf("failed to verify issuer certs stored in %s: %s", consts.IdentityIssuerSecretName, err)
 	}
 
 	return &charts.Identity{
 		TrustDomain:     idopts.trustDomain,
-		TrustAnchorsPEM: anchors,
+		TrustAnchorsPEM: string(secret.Data[k8s.IdentityIssuerTrustAnchorsNameExternal]),
 		Issuer: &charts.Issuer{
 			Scheme:             string(corev1.SecretTypeTLS),
 			ClockSkewAllowance: idopts.clockSkewAllowance.String(),
 			IssuanceLifetime:   idopts.issuanceLifetime.String(),
 		},
 	}, nil
-
 }
```

Will this work? This is (by the looks of it) attempting to fetch the k8s secret, gets its type and then read files from disk. I feel this `NewCredsFromSecret` will return an error here if the files are not on disk. Am I missing something major here? 

Anyway, I am fine with whatever approach as long as it works, is not scattered all over the place, is composable and is not confusing.

ihcsim:
@zaharidichev Thanks for exploring this with me. Per slack discussion, to avoid further increasing the change scope in this PR, let's stick with your previous approach. You are right in that some of these existing code are called by multiple places with different context; essentially, we traded DRY-ness for fan-ins. Sorry for the noise.

ihcsim:
I notice that after an issue certificate expired, `linkerd check --proxy` reports the following error:
```
linkerd-api
-----------
√ control plane self-check
    Error calling Prometheus from the control plane: server_error: server error: 503
```
Prometheus logs show:
```
bin/linkerd logs --control-plane-component=prometheus
...
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy ERR! [   519.909393s] rustls::session TLS alert received: Message {
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy     typ: Alert,
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy     version: TLSv1_2,
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy     payload: Alert(
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy         AlertMessagePayload {
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy             level: Fatal,
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy             description: BadCertificate,
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy         },
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy     ),
linkerd-prometheus-776b4fbb95-rflfh linkerd-proxy }
...
```

These are my steps to reproduce the issue:

First install Linkerd. FWIW, my issuer cert has a 5-minute validity and trust root has a 3-days validity:
```
$ bin/linkerd install --identity-trust-anchors-file=ca.crt --identity-issuer-certificate-file=issuer.crt --identity-issuer-key-file=issuer.key --identity-issuance-lifetime=1m | k apply -f -

$ bin/linkerd check --proxy
√ issuer certs are valid
‼ issuer certs do not expire sooner than 60 days from now
    certificate is expiring sooner than 60 days from now
    see https://linkerd.io/checks/#l5d-issuer-certs-do-not-expire-sooner-than-60-days for hints
√ data plane proxies certificate match CA

# wait for a few minutes **after** the cert expired
$ bin/linkerd check --proxy
linkerd-api
-----------
√ control plane self-check
    Error calling Prometheus from the control plane: server_error: server error: 503
```

LMK if you can produce this on your side.

zaharidichev:
cc @ihcsim @adleong @grampelberg 

So I made a few changes here namely to provide the check command with the ability to give users more information as to what exactly is wrong with the certs. This will be quite helpful in cases such as the one described in this issue: #3777 So there is a new checke category now and its concen is only certificates. These are the checks that are contained in it: 

- **certificate config is valid(fatal)** - this pretty much ensures that the issuer certificates are in the correct format. SO if we are configured to use a certain tls scheme, it will check the contents of the map to ensure nothing is missing and keys are named as they should be named
- **data plane proxies certificate match CA(warning)** - this is an old check that I moved to this category
- **no trust roots expire sooner than 60 days from now(warning)** - this is an interesting one. What we store in trustRootAnchors as far as Itake it might be multiple roots. What we actually care about is that our issuer certs can be verified (build a chain) to at least one of these roots. So we cannot assue that there is only one root and check it for expiry. Therefore we simply issue a warning if any of the roots expire in less than 60 days. 
- **issuer certificate matches algorithm requirements(fatal)** - here the constraints on the algorithms are checked so we dont end up with unpleasant situations such as the one described in #3777
- **issuer certificate is not expired(fatal)** - we check explicitly here whether our issuerCa has expired and if it has, we fail the check.
- **issuer certificate validates with trust roots(fatal)** - This is where we ensure that the issuer cert can be verified with the roots. Note that is there is no such root or it has expired, this check will fail.

So all in all, I think this gives us a pretty good description of what is happening wrt to certificates and if there is a problem, it can be discovered quickly by running the checks command. Let me know what you think.

zaharidichev:
Now with respect to our last conversation around behaviour of linkerd when certs expire. Here is what the results of this pr are. 

**When it is all good:** 
```
linkerd-certs
-------------
√ certificate config is valid
√ data plane proxies certificate match CA
√ no trust roots expire sooner than 60 days from now
√ issuer certificate matches algorithm requirements
√ issuer certificate is not expired
√ issuer certificate validates with trust roots
```

**When issuer cert has expired but root has not:** 
Identity service fails to serve CSRs (as it should) and produces the following logs: 

`time="2019-12-04T15:38:03Z" level=error msg="Could not process CSR because of CA cert validation failure: x509: certificate has expired or is not yet valid"`

Output of checks is: 

```
linkerd-certs
-------------
√ certificate config is valid
√ data plane proxies certificate match CA
√ no trust roots expire sooner than 60 days from now
√ issuer certificate matches algorithm requirements
× issuer certificate is not expired
    certificate not valid anymore. Expired at: 2019-12-04T15:37:55Z
    see https://linkerd.io/checks/#l5d-issuer-cert-not-expired for hints
```

**When the root expires but the issuer does not:** 
Initially because expiry of the root is soon, we get a warning: 
```
linkerd-certs
-------------
√ certificate config is valid
√ data plane proxies certificate match CA
‼ no trust roots expire sooner than 60 days from now
    one or more root certificate is expiring sooner than 60 days from now
    see https://linkerd.io/checks/#l5d-issuer-certs-do-not-expire-sooner-than-60-days for hints
√ issuer certificate matches algorithm requirements
√ issuer certificate is not expired
√ issuer certificate validates with trust roots
```

Then after it expires we can observe the identitiy service failing to process CSRs: 
time="2019-12-04T15:45:51Z" level=error msg="Could not process CSR because of CA cert validation failure: x509: certificate has expired or is not yet valid"

Also we can see in he checks now that the issuer ca cert cannot be validated with the roots provided (as the only one in there that is valid has expired): 

```
linkerd-certs
-------------
√ certificate config is valid
√ data plane proxies certificate match CA
‼ no trust roots expire sooner than 60 days from now
    one or more root certificate is expiring sooner than 60 days from now
    see https://linkerd.io/checks/#l5d-issuer-certs-do-not-expire-sooner-than-60-days for hints
√ issuer certificate matches algorithm requirements
√ issuer certificate is not expired
× issuer certificate validates with trust roots
    x509: certificate has expired or is not yet valid
    see https://linkerd.io/checks/#l5d-issuer-certs-are-valid-for-root for hints
```

I think this part should solve #3779

Let me know what you think :)

grampelberg:
How is the trust root matching working? I'd like to make sure that's informational when users are updating the trust root.

zaharidichev:
It will provide a list of pods that have the contents of their `LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS` not matching (one for one) what is in the linkerd config map.

grampelberg:
@zaharidichev is there a way to tell if it matches vs if it would even work? Also, WDYT about having a count as well as a list? eg. 10/30 pods do not match.

zaharidichev:
Adding a count is not hard. There is a way to check whether it will work. However, I think that this is performed in the checks that come after that. Ie, if all pods roots match whats in the config map, then we verify the certs in he config map with the roots in the config map. Isn't that enough ?

grampelberg:
> Ie, if all pods roots match whats in the config map, then we verify the certs in he config map with the roots in the config map. Isn't that enough ?

I was thinking of the situation where you've added a new cert during rotation. It won't match, but the old proxies will continue to work.

zaharidichev:
@grampelberg you are right, will change it accordingly

zaharidichev:
So I took all the suggestion by @ihcsim  and @grampelberg  and restructured the checks into two categories, adding quite a bit of detail to each one of them. The successful output looks as follows: 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
√ issuer cert has valid lifetime

linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
√ all data plane trust roots have valid lifetimes
√ control plane issuer certificate can be verified with data plane trust roots

```

### What does each check do:

#### linkerd-mtls
These checks are concerned only to the control plane or otherwise put, the certs that are in `linkerd-identity-issuer` and the roots that are in the `IdentityContext.TrustAnchors`

- √ certificate config is valid - this one looks at the config map and the linkerd-identity-issuer secret. It ensured that the secret is there and is in the correct form. Namely if our linkerd config indicates that we are using a secret of type `linkerd.io/tls`, it ensures that we have the corresponding data keys in the secret that are expected for this scheme. 
- √ control plane is using supported trust roots - this check gets the trust roots from the config map and for each root cert (as there might be more than one) it checks the algorithm constraints. We need to have `ECDSA P-256` for the roots. The check is fatal. If any of the root certs in this bundle are not compliant with our constraints, we fail and provide details as to what is wrong.
- √ identity is using supported issuer certificate - this check does the same as the above one but instead looks at the cert that is in the `identity-issuer-secret.` This check is fatal as well.
- √ all control plane trust roots have valid lifetimes - this check looks at all root certs that are in the config map for linkerd and verified that their `NotBefore` and `NotAfter` are fine. The check will fail if any of the roots is expired, listing the first seven bytes of the cert's signature + the subject common name + the reasoning why its not valid (not valid yet OR not valid anymore). If any of the certs is at more than 90% of its lifetime this check will instead simply warn, listing the roots that are close to expiry and issuing an event to the k8s event bus. So whether this check is fatal or not depends on the circumstances. 
- √ issuer cert has valid lifetime - same as above this check ensures that the certificate in the linkerd-identity-issuer is valid time-wise and will error or warn accordingly

#### linkerd-data-plane-mtls
These checks are concerned with all the meshed pods in the cluster. Essentially these validate whether the roots that the proxies of these meshed pods are valid and whether they work with the issuer certificates that the control plane is configured to work with

- √ data plane is using supported trust roots - this one gets all the meshed pods and for every pods checks all root certs that the proxy container is configured with. It ensures that we use the correct signature, public key algorithms and curve type. If any of the roots fail, it will list the root signature, the subject common name and the pods that this cert is residing in. It will also list stats about how many pods are affected (i.e. 5/500 pods, where 500 is the number of meshed pods that we have)
- √ all data plane trust roots have valid lifetimes - much like the control plane tls check, this check looks at all roots for all proxies and issues and produces an error or warning depending on the time validity of the cert. In both cases it will provide a list of roots that are expired/going to expire soon and the pods where they reside in. It will also output stats for the number of affected pods. In case of a warning (cert at more than 90% of lifetime) it will also issue a warning event on the K8s bus. 
- √ control plane issuer certificate can be verified with data plane trust roots - this is probably the check that delivers the ultimate guarantee that thing are ok. Essentially what it does it goes through the roots of all proxies and tries to validate the control plane issuer certificate (stored in `linkerd-identity-issuer`) with the roots. If any of that fails, it produces a list of pods, whose roots cannot verify the issuer certificate and fails. The rationale behind that is simply that we dont really care that much how many certs are in the root bundles of the proxies or whether the env var matches exactly what is in the config map (`IdentityContext.TrustAnchors`), as there will be times (during cert rotation) when these will not match one for one. The thing we care about is that it works. 

Phew... 

Now lets look at how our failure scenarios look like and what information we get out of the check: 

- When we have `linkerd.io/tls` secret type but we don't have the appropriate content in the `linkerd-identity-issuer` map: 

```
linkerd-mtls
------------
× certificate config is valid
    key crt.pem containing the issuer certificate needs to exist in secret linkerd-identity-issuer if --identity-external-issuer=false
    see https://linkerd.io/checks/#l5d-crt-config-is-valid for hints

Status check results are ×
```


- When unsupported trust root is used:

```
linkerd-mtls
------------
√ certificate config is valid
× control plane is using supported trust roots
    Invalid roots:
      * 30818702413131 identity.linkerd.cluster.local must use P-256 curve for public key, instead P-521 was used
    see https://linkerd.io/checks/#l5d-supported-certs-type for hints

Status check results are ×
```

- When unsupported identity issuer cert is used: 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
× identity is using supported issuer certificate
    issuer certificate must use P-256 curve for public key, instead P-521 was used
    see https://linkerd.io/checks/#l5d-supported-certs-type for hints
```


- When (at least one) of the control plane trust root certs is about to expire (warning) (more than 90% of the way there): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
‼ all control plane trust roots have valid lifetimes
    Roots expiring soon:
      * 3044022019354d identity.linkerd.cluster.local will expire on 2019-12-06T10:41:07Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints
√ issuer cert has valid lifetime
```

- When (at least one) of the control plane trust root certs is expired (fatal): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
× all control plane trust roots have valid lifetimes
    Invalid roots:
      * 3044022019354d identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T10:41:07Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```


- When the issuer certificate is about to expire (warning): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
‼ issuer cert has valid lifetime
    issuer certiifcate will expire on 2019-12-06T10:45:14Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints
```

- When the issuer certificate is expired (fatal): 

```
linkerd-mtls
------------
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
× issuer cert has valid lifetime
    issuer certiifcate is not valid anymore. Expired on 2019-12-06T10:45:14Z
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```


- When some of the roots in some of the proxies are not in the correct format (either unparsable or wrong algorithm): 

```
linkerd-data-plane-mtls
-----------------------
× data plane is using supported trust roots
    Invalid roots:
      * Unparsable anchors for pod linkerd-grafana-5fcb5f44b9-z7hqd
      * 308187024200f9 identity.linkerd.cluster.local must use P-256 curve for public key, instead P-521 was used in pod linkerd-tap-768d79f567-qxrsp
    Affected pods: 2/11
    see https://linkerd.io/checks/#l5d-supported-certs-type for hints

Status check results are ×
```


- When certs int he proxies root anchors are about to expire. Obviously that will give information about all the different roots
across all proxies, so it provides quite a lot of visibility into what is actually going on.  

```
linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
‼ all data plane trust roots have valid lifetimes
    Roots expiring soon:
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-controller-844df9c9d5-vxsw2
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-destination-67b9f958d-qm4gx
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-grafana-86c957cc7d-jd4fr
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-identity-6b798cb557-5r4k4
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-prometheus-64c75d7659-kbghf
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-proxy-injector-b5d8d5984-rl2km
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-sp-validator-5fb587cf57-qsqsj
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-tap-7947f75966-h8c4r
      * 3046022100cfa3 identity.linkerd.cluster.local will expire on 2019-12-06T11:05:48Z  in pod linkerd-web-587d69c8b4-p5qsm
    Affected pods: 9/9
    see https://linkerd.io/checks/#l5d-certs-rotation for hints
√ control plane issuer certificate can be verified with data plane trust roots
```


- When certs in the proxy root anchors have expired (fatal): 

```
linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
× all data plane trust roots have valid lifetimes
    Expired roots:
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-controller-844df9c9d5-vxsw2
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-destination-67b9f958d-qm4gx
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-grafana-86c957cc7d-jd4fr
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-identity-6b798cb557-5r4k4
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-prometheus-64c75d7659-kbghf
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-proxy-injector-b5d8d5984-rl2km
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-sp-validator-5fb587cf57-qsqsj
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-tap-7947f75966-h8c4r
      * 3046022100cfa3 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-06T11:05:48Z in pod linkerd-web-587d69c8b4-p5qsm
    Affected pods: 9/9
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```


- Finally when for some reason everything passes but there are one or more proxies that have roots that just don't work with the issuer certs at all, we can diagnose that as well: 

```
linkerd-data-plane-mtls
-----------------------
√ data plane is using supported trust roots
√ all data plane trust roots have valid lifetimes
× control plane issuer certificate can be verified with data plane trust roots
    Roots in pods cannot verify control plane issuer cert:
      * linkerd-grafana-69ff587876-6xqqb
      * linkerd-tap-dfc85d948-r6zlv
    Affected pods: 2/11
    see https://linkerd.io/checks/#l5d-certs-rotation for hints

Status check results are ×
```

zaharidichev:
On another note, the events that are issued at the moment are of the following form: 

```
➜  ~ kubectl get events --field-selector reason=LinkerdCertsExpiringSoon -n linkerd
LAST SEEN   TYPE      REASON                     OBJECT                        MESSAGE
64s         Warning   LinkerdCertsExpiringSoon   deployment/linkerd-identity   Some of the TLS certificates that Linkerd is configured with are expiring soon
➜  ~
```
Currently whatever is about to expire will issue the same event. We can provide much more detail in the message or at least make them different for roots and issuer certs. Up to you

zaharidichev:
@ihcsim @grampelberg At that point the feedback that I am mostly looking for is around [this description](https://github.com/linkerd/linkerd2/pull/3696#issuecomment-562537902). When that is all agreed upon, I will apply some more refactoring and fix/rewrite unit and integration tests

ihcsim:
@zaharidichev Great work on extending the completeness of the checks. A few questions to confirm my understanding:

Both `linkerd check` and `linkerd check --proxy` run the `linkerd-mtls` and `linkerd-data-plane-mtls` checks. Initially, I was thinking we can separate out the two checks. What are the benefits of running them in both commands?

> listing the first seven bytes of the cert's signature 

Is this to help users to identify which certificates within the bundle is failing? I wonder if we should use the certificate serial number instead.

>  and issuing an event to the k8s event bus.

I haven't look into the code yet, but is this event generated by the CLI while it's calling the `healthcheck` package? 

>  it will provide a list of roots that are expired/going to expire soon and the pods where they reside in

For clusters with many pods (one user has at least 3000 meshed pods on his cluster), this will clutter the `check` output. Any thoughts on how we should handle this? Can we group this by namespace and/or limit the row counts in the output to say 10?

adleong:
Regarding the category name `linkerd-mtls`: we call this feature "identity" elsewhere in the product.  To be consistent, should this be `linkerd-identity`?

"certificate config is valid": does this check that the issuer certificate validates against the trust root bundle?

"all control plane trust roots have valid lifetimes": I notice this is now 90% instead of 60 days.  For year long certs, this is about 36 days instead of 60.  Is this enough of a warning?

"linkerd-data-plane-mtls": I assume this category is only included when `--proxy` is used?

"(i.e. 5/500 pods, where 500 is the number of meshed pods that we have)": I think this points to a larger issue around how the --proxy checks are organized which is out of scope for this PR.  

"control plane issuer certificate can be verified with data plane trust roots": This feels a little bit backwards to me.  It does test the setup almost end-to-end but it seems backwards that this is a data plane check when it's the data plane checking the validity of the control plane.

zaharidichev:
@adleong 

> Regarding the category name linkerd-mtls: we call this feature "identity" elsewhere in the product. To be consistent, should this be linkerd-identity?

Yes that makes sense. 

> "certificate config is valid": does this check that the issuer certificate validates against the trust root bundle?

It does not. I was thinking of including it  and now that you bring it up, I will do so. What it checks at the moment is that the certs and key are parsable and that they have the correct algo requirements. Also it checks that the secret is in the correct format (i.e. if its a cert manager type scheme it will check for the corresponding keys and vice versa)

> "all control plane trust roots have valid lifetimes": I notice this is now 90% instead of 60 days. For year long certs, this is about 36 days instead of 60. Is this enough of a warning?
I think when we last spoke with @olix0r and @grampelberg we settled on this value. Its easy to change so I am not too attached. 

> "linkerd-data-plane-mtls": I assume this category is only included when --proxy is used?

Should be, yes.

> "(i.e. 5/500 pods, where 500 is the number of meshed pods that we have)": I think this points to a larger issue around how the --proxy checks are organized which is out of scope for this PR.
Just out of curiosity, what is the issue? 

> "control plane issuer certificate can be verified with data plane trust roots": This feels a little bit backwards to me. It does test the setup almost end-to-end but it seems backwards that this is a data plane check when it's the data plane checking the validity of the control plane.

The way I see it is that it is a data plane, because it checks every proxy. If 5 of the proxies pass this check and only one fails, the moral is that the particular proxy that is failing the check has the wrong roots in there. I guess its just a matter of perspective. Not sure really.

zaharidichev:
@ihcsim 

> Both linkerd check and linkerd check --proxy run the linkerd-mtls and linkerd-data-plane-mtls checks. Initially, I was thinking we can separate out the two checks. What are the benefits of running them in both commands?

Yes, I left it like that so I can run them more easily while testing it. Will separate them for sure.

> I wonder if we should use the certificate serial number instead.

We can, I guess as long as its not 1000 chars long :) 

> I haven't look into the code yet, but is this event generated by the CLI while it's calling the healthcheck package?

Correct, its generated by the CLI but the owner of the event is the identity service. I think that is what we talked about.

> For clusters with many pods (one user has at least 3000 meshed pods on his cluster), this will clutter the check output. Any thoughts on how we should handle this? Can we group this by namespace and/or limit the row counts in the output to say 10?

Yeah, I agree on the clutterness. I think its a tough call on what would be the best way to summerize that. Frankly I lean towards limiting the number displayed pods instead of namespaces. I think that this will give better information to the user. So we can limit it to 10 lets say and provide a command line option that allows tweaking that number so people can see as much as they want. How does that sound?

cli/cmd/upgrade.go
```diff
@@ -438,6 +438,18 @@ type issuerData struct {
 	exp time.Time
 }
 
+func verifyCreds(creds *tls.Cred, trustAnchors, dns string) error {
```
ihcsim: I think there is an opportunity to reuse your new code from `pkg/issuercerts` in `readIssuer()` and `fetchIssuer()`, while getting rid of the `issuerData` struct and `verifuCreds()` function. I recommend we do the clean up in a separate PR.

pkg/healthcheck/sidecar.go
```diff
@@ -7,7 +7,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// HasExistingSidecars returns true if the pod spec already has the proxy init
+// HasExistingSidecars returns true if the name spec already has the proxy init
```
ihcsim: The previous usage of _pod spec_ sounds better to me.

grampelberg:
Overall, this looks great to me and should catch most issues here. Just a couple comments:

Should we have a flag to select only these sections of check? As part of a root update `check` is going to get run a *lot*. It'd be nice to just have it check the cert side of things, but I'm mostly thinking about the docs.

> Correct, its generated by the CLI but the owner of the event is the identity service. I think that is what we talked about.

I would prefer not to change state on the cluster from the CLI. Could we either skip this or have it be a cronjob? It seems like we'll need to get better with what object to associate it with as well.

ihcsim:
@zaharidichev I think the changes in `pkg/healthcheck/healthcheck.go` alone has grown quite a bit. Can we split this up into 2 or 3 PRs in this order?

1. A PR to include the new `pkg/issuercerts` package and the CLI changes around it. We can do some work here to make the exported functions work with `upgrade` too. See this [gist](https://gist.github.com/ihcsim/8d5cd19ff813c411061ea279b45e421a) for example.
1. Then another PR to include the `healthcheck` changes
1. Then optionally, a PR for the `pkg/identity` stuff (I don't know if we gained much by changing this tbh)

adleong:
Based on https://docs.google.com/document/d/148CPU0bUbz2cPFQ8Sj4UDOrU34QdqSDCll3hiXvb_NY/edit# here's where I think we are on this checks:

```
√ certificate config is valid
√ control plane is using supported trust roots
√ identity is using supported issuer certificate
√ all control plane trust roots have valid lifetimes
√ issuer cert has valid lifetime
```

I think the above checks are all simply checking the validity of the certificates and we would not expect any of them to fail at any point during certificate rotation.  I believe these are all fatal errors.  I think we can split these into a separate PR and then merge them basically as is.  One change is that the "issuer cert has valid lifetime" should just check for the certs actually being expired and return an error. (as we discussed above, these checks should also verify that the issuer validates against the trust root)

In a separate PR, I think we should have a check which looks for if the issuer cert or trust root is expiring "soon" and returns a warning.

In a separate PR, I think we should have a `--proxy` check which verifies that the data plane proxies have identical trust bundles to the trust bundle config map.  I think checking for them being identical is simpler and more to the point than attempting to validate the issuer against the proxies.  In conjunction with the checks above which verify that the config map trust root is valid, this should be sufficient.  The call to action is also more clear: restarting the pod will always resolve this.

In a separate PR, we need to add a gate to `linkerd upgrade` when the `--identity-issuer-certificate-file` is used to ensure that the new issuer validates against the trust root of all the proxies.  Additionally, when the `--identity-trust-anchors` flag is used, we need to validate that the issuer validates against the new trust root.

adleong:
I've suggested splitting this up into smaller PRs to make it easier to review and get merged.  But please use your best judgement if you think certain parts of this make more sense to split up or consolidate.

pkg/healthcheck/healthcheck.go
```diff
@@ -198,6 +216,21 @@ func IsCategoryError(err error, categoryID CategoryID) bool {
 	return false
 }
 
+// CheckRunResult Represents the result of running an individual checker
+type CheckRunResult struct {
```
adleong: in the long term, I really like this approach.  In fact, I even suggested something similar on another PR.  But to keep this diff size down, I don't think we actually need to do this here.  I think we can keep the "error" and "warning" checks separate.

pkg/healthcheck/sidecar.go
zaharidichev: That was an automatic refactor gone bad...

zaharidichev:
@grampelberg @adleong @ihcsim 

I am closing this in favour of: 

#3810
#3811
#3813 

Its best if we review/merge them in this particular order. I am currently working on another that adds some validation to the `upgrade` command as per the comment Alex made here:

> In a separate PR, we need to add a gate to linkerd upgrade when the --identity-issuer-certificate-file is used to ensure that the new issuer validates against the trust root of all the proxies. Additionally, when the --identity-trust-anchors flag is used, we need to validate that the issuer validates against the new trust root.

Issue #3592, opened 2019-10-16
Update the dashboard to Material 4


scottcarol:
The dashboard is currently using Material-UI `3.6.1` but the [latest version](https://material-ui.com/getting-started/installation/) is `4.5.1`. Update the dashboard, ensuring views and functionality are consistent.

scottcarol:
Closed by #3710

PR #3698, opened 2019-11-12
Fix Octopus graph with multiple nodes

### LINKED ISSUE #3577: Octopus Graph Polish
As raised in #3047, there are a few issues with the octopus graphs if there are multiple nodes displayed:

- Vertical positioning of arrows is not consistent when there are multiple nodes in the graph (see #3047 for screenshot)
- The octopus graph can only display 6 nodes, with an additional box for "overflow nodes". The styling, sizing and positioning of the "overflow node box" (as discussed in above issue) needs to be improved. Any arrow positioning changes should take this into account.

cynthia-sg:
This PR fixes issues with Octopus graphs components when some cards have different sizes: 
- Fix vertical position of arrows
- Improve collapsedNeighbors card title (based on #3047)
- Support for custom `maxNumNeighbors`, now we can use any neighbours number greater than or less than 6 passing it through prop (by default we are using 6) -> 8 is the result of 6 `resourceCard` + `unmeshedResources` card + `collapsedNeighbors` card
- Tested on Safari, Chrome, Firefox and Edge (for Mac)

Closes #3577

octopus_01
octopus_02
octopus_03
octopus_04
octopus_05

### DIFF
--- a/web/app/js/components/Octopus.jsx
+++ b/web/app/js/components/Octopus.jsx
@@ -1,4 +1,4 @@
-import { OctopusArms, baseHeight } from './util/OctopusArms.jsx';
+import { OctopusArms, inboundAlignment } from './util/OctopusArms.jsx';
 import { displayName, metricToFormatter } from './util/Utils.js';
 
 import Card from '@material-ui/core/Card';
@@ -6,14 +6,13 @@ import CardContent from '@material-ui/core/CardContent';
 import Grid from '@material-ui/core/Grid';
 import PropTypes from 'prop-types';
 import React from 'react';
+import RootRef from '@material-ui/core/RootRef';
 import { StyledProgress } from './util/Progress.jsx';
 import Table from '@material-ui/core/Table';
 import TableBody from '@material-ui/core/TableBody';
 import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
 import Typography from '@material-ui/core/Typography';
-import _ceil from 'lodash/ceil';
-import _floor from 'lodash/floor';
 import _get from 'lodash/get';
 import _isEmpty from 'lodash/isEmpty';
 import _isNil from 'lodash/isNil';
@@ -21,11 +20,8 @@ import _size from 'lodash/size';
 import _slice from 'lodash/slice';
 import _sortBy from 'lodash/sortBy';
 import _take from 'lodash/take';
-import _times from 'lodash/times';
 import { getSuccessRateClassification } from './util/MetricUtils.jsx';
 import { withStyles } from "@material-ui/core/styles";
-
-const maxNumNeighbors = 6; // max number of neighbor nodes to show in the octopus graph
 
 const styles = () => ({
   graphContainer: {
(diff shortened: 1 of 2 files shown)


scottcarol:
@cynthia-sg, looking forward to reviewing this! For my own understanding, could you explain why you decided to use the `RootRef` component and why the `createRef` API is a good choice here? Are there any alternatives? I want to be careful around introducing more complexity to the graph here unless it's 100% necessary.

cynthia-sg:
To fix properly the arrow, we need to get the correct height and position relative to the parent for all cards. Unfortunately I haven’t been able to achieve this in a different way using Material UI 3 because:

- prop `ref` is not available for `Grid` component 
- the `ref` for `Card` component is forwarded to the main element 

Anyway we can remove this `extra` component after updating the library #3592 

About `createRef`, I used it only for simplicity, but if you prefer I can use callback ref as I do for cards.

web/app/js/components/Octopus.jsx
```diff
@@ -92,6 +103,16 @@ class Octopus extends React.Component {
     return display;
   }
 
+  refCallback = (element, index, isOutbound) => {
```
scottcarol: could this be named something more descriptive, for readability? Am I correct in thinking that `refCallback` adds the element to the appropriate array (`downstreamRefs` or `upstreamRefs`) so that the spacing will be correct?

web/app/js/components/Octopus.jsx
```diff
@@ -115,22 +136,24 @@ class Octopus extends React.Component {
...
-
-            <Progress variant="determinate" value={resource.successRate * 100} />
-
-            <Table padding="dense">
-              {showTcp ? this.renderTCPStats(resource) : this.renderHttpStats(resource)}
-            </Table>
-          </CardContent>
-        </Card>
-      </Grid>
+      <RootRef rootRef={el => type === "neighbor" ? this.refCallback(el, index, isOutbound) : undefined} key={resource.type + "-" + resource.name}>
```
scottcarol: There is a lot of logic in this line - is it possible to calculate `rootRef` before we get to the `return` statement, for easier debugging?

alpeb:
Thanks @cynthia-sg. I'm trying to test the new behavior (the code changes are above my React level :-P)
I'm testing using the emojivoto app, looking at the web pod, and I found a small issue. When I scale the vote-bot deployment to 7 replicas, it's behaving as expected. But then when I scale it back to 1 an orphaned arrow remains:
image
I'm not sure if this is related to this change. If not, it could be addressed in a separate PR.

alpeb:
Also (likely unrelated to this change as well), it looks like the `Unmeshed` node is not being cleaned up. For example if I deploy an unmeshed vote-bot pod, wait for it to appear in the topology graph, and then undeploy it, it remains in the graph, unlike other meshed nodes that are removed from the graph as soon as traffic from them stops.

web/app/js/components/Octopus.jsx
```diff
@@ -43,23 +39,38 @@ const styles = () => ({
...
+    paddingTop: "10px",
+  },
 });
 class Octopus extends React.Component {
   static defaultProps = {
     neighbors: {},
     resource: {},
-    unmeshedSources: []
+    unmeshedSources: [],
+    maxNumNeighbors: 6,
```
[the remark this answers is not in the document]
cynthia-sg: Really I was thinking only to have it as a new feature (perhaps we need it in the future, not only to display more than 8, if not to reduce the `maxNumNeighbors` if required). But I'll leave it like it is now :)

cynthia-sg:
Good catch, @alpeb ! Thanks for your feedback.

cynthia-sg:
> Also, what does the `extra` in your comment below mean?

`RootRef` could be removed after updating `Material UI`

scottcarol:
This is likely because I just viewed your updated branch with a slightly different workload setup -- when I view the dashboard, the line between the center node and the downstream nodes, and the line to the middle downstream node, looks a little disjointed. It's not a blocker for this PR, but pointing it out in case there is a simple way to line up those two lines in this case (?). Otherwise this looks good to merge!

@alpeb let me know if it is working on your end!

(zoomed out)
Screen Shot 2019-12-04 at 11 53 31 AM
 

(close up)
Screen Shot 2019-12-04 at 11 45 37 AM

alpeb:
Thanks @cynthia-sg for the quick fix regarding the orphaned arrow; I confirmed it's working fine now :+1: 
I'm still seeing though the unmeshed sources not being removed from the graph when undeployed. Is that something to be addressed in a separate PR?

scottcarol:
@alpeb Yes let's address this in a separate PR and get these styling changed merged in. Thanks for observing that issue, I'll create a separate ticket.

cynthia-sg:
> Thanks @cynthia-sg for the quick fix regarding the orphaned arrow; I confirmed it's working fine now 👍
> I'm still seeing though the unmeshed sources not being removed from the graph when undeployed. Is that something to be addressed in a separate PR?

Thanks @alpeb. IMHO it would be better to address the issue in a separate PR when this one is not changing the logic about displaying or not the unmeshed sources.

Issue #3224, opened 2019-08-09
Increase the comprehensiveness of check --pre


grampelberg:
## What problem are you trying to solve?

`check` is awesome and provides a way for users to reason about whether linkerd can run effectively on their clusters. Unfortunately, it doesn't test for somethings. For example, while users can create `ClusterRoles`, it is possible that they don't have the broad permissions required to actually create these with the correct permissions required. 

### How should the problem be solved?

For `--pre`, it would provide better results to do something similar to what `kubectl apply --server-dry-run` does.

grampelberg:
As part of this, it would be nice to add a check to validate that the injector is operating correctly. This could simply try and create a pod, and see if it was correctly injected. The `--server-dry-run` *should* go through the injector and give some information back for that. See #3290 and #2973 for two examples of this failure.

Pac23:
can i take this as per slack discussion ?

grampelberg:
Go for it!

Pac23:
> Hi @Pac23
> 
> I was about to start working on this issue and I was wondering if you were still interested in it.
> 
> Thanks!

i found some of the bits of the solution to this issue slightly difficult as a i learned go only couple of weeks back,please take this.

tegioz:
Thanks @Pac23!

PR #3708, opened 2019-11-13
Add check to verify injector is operating correctly

tegioz:
A new check has been added to validate that the injector is operating correctly. It creates a pod using dry-run and checks if the pod was injected successfully. 

Mentioned in a comment in #3224

### DIFF
--- a/charts/linkerd2/templates/web-rbac.yaml
+++ b/charts/linkerd2/templates/web-rbac.yaml
@@ -102,6 +102,36 @@ subjects:
   name: linkerd-web
   namespace: {{.Values.namespace}}
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: linkerd-web-check-injector
+  namespace: default
+  labels:
+    {{.Values.ControllerComponentLabel}}: web
+    {{.Values.ControllerNamespaceLabel}}: {{.Values.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["create"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: linkerd-web-check-injector
+  namespace: default
+  labels:
+    {{.Values.ControllerComponentLabel}}: web
+    {{.Values.ControllerNamespaceLabel}}: {{.Values.Namespace}}
+roleRef:
+  kind: Role
+  name: linkerd-web-check-injector
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: linkerd-web
+  namespace: {{.Values.Namespace}}
+---
 {{- end}}
 kind: ServiceAccount
 apiVersion: v1
--- a/charts/linkerd2/templates/web-rbac.yaml
+++ b/charts/linkerd2/templates/web-rbac.yaml
@@ -102,36 +102,6 @@ subjects:
   name: linkerd-web
   namespace: {{.Values.namespace}}
 ---
-apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
-metadata:
-  name: linkerd-web-check-injector
-  namespace: default
-  labels:
-    {{.Values.ControllerComponentLabel}}: web
-    {{.Values.ControllerNamespaceLabel}}: {{.Values.Namespace}}
-rules:
-- apiGroups: [""]
-  resources: ["pods"]
-  verbs: ["create"]
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
-metadata:
-  name: linkerd-web-check-injector
-  namespace: default
-  labels:
-    {{.Values.ControllerComponentLabel}}: web
-    {{.Values.ControllerNamespaceLabel}}: {{.Values.Namespace}}
-roleRef:
-  kind: Role
-  name: linkerd-web-check-injector
-  apiGroup: rbac.authorization.k8s.io
-subjects:
-- kind: ServiceAccount
-  name: linkerd-web
-  namespace: {{.Values.Namespace}}
----
 {{- end}}
 kind: ServiceAccount
 apiVersion: v1
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -24,7 +24,6 @@ type checkOptions struct {
 	namespace       string
 	cniEnabled      bool
 	output          string
-	dryRun          bool
 }
 
 func newCheckOptions() *checkOptions {
@@ -36,7 +35,6 @@ func newCheckOptions() *checkOptions {
 		namespace:       "",
 		cniEnabled:      false,
 		output:          tableOutput,
-		dryRun:          true,
 	}
 }
 
@@ -48,7 +46,6 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	flags.BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
-	flags.BoolVar(&options.dryRun, "dry-run", options.dryRun, "Include checks that rely on k8s dry-run (they may require special write permissions)")
 
 	return flags
 }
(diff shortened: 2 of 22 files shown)


tegioz:
Hi @grampelberg

This PR should cover the second part of #3224.

charts/linkerd2/templates/web-rbac.yaml
```diff
@@ -25,6 +25,19 @@ rules:
...
+metadata:
+  name: linkerd-web-check-injector
+  namespace: default
+  labels:
+    {{.ControllerComponentLabel}}: web
+    {{.ControllerNamespaceLabel}}: {{.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["create"]
```
adleong: could this rule just be added to the `linkerd-web-check` role rather than adding a new role?

tegioz: The role that rule belongs to is attached to the `default` namespace whereas `linkerd-web-check` role is attached to the namespace where Linkerd is installed. Please note that `linkerd-web-check` role is gone after the rebase I've just done from master. Those permissions were nicely merged into `linkerd-web` role in #3694.

I used the default namespace instead the one Linkerd is installed to because the latter had the annotation `linkerd.io/inject: disabled`, so the pod created during the check was never injected.

charts/linkerd2/templates/web-rbac.yaml
```diff
@@ -103,6 +103,36 @@ subjects:
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: linkerd-web-check-injector
+  namespace: default
```
zaharidichev: I wonder whether it makes sense for this to be configurable to allow users to specify a different "test" namespace

pkg/healthcheck/healthcheck.go
```diff
@@ -1445,6 +1454,45 @@ func (hc *HealthChecker) checkClockSkew() error {
...
+						"name":  "busybox",
+						"image": "busybox",
+					},
+				},
+			},
+		},
+	}
+	injectedPod, err := hc.kubeAPI.DynamicClient.
+		Resource(schema.GroupVersionResource{Version: "v1", Resource: "pods"}).
+		Namespace("default").
```
zaharidichev: Same as above comment

tegioz:
Hi @adleong 

> ```
> ‼ injector is operating correctly
>     error creating pod to check injector: the dryRun alpha feature is disabled
> ```
> 
> Can we detect if dryRun is available and skip this check if it is not?

I was thinking that maybe checks could have a `requiredFeatures` field that we could check when running the checks. That way we could skip them easily if the required features are not enabled. But I haven't found yet anything in the k8s api to check what feature gates are enabled in the cluster.

```go
{
    description: "injector is operating correctly",
    requiredFeatures: []featuregate.Feature{
        features.DryRun,
    },
}
```

Another option would be to do something similar but based on the Kubernetes cluster version. The problem is this doesn't seem very reliable because users can disable feature gates and we wouldn't be covering when features are in alpha either.

Going back to the first solution, if we can't find a way to check for the feature gates enabled, we could build something ourselves to check for the ones the checks depend of. In this case it'd be just to check `DryRun`. I could try to just use it attempting to create a pod and based on the error returned track the feature as enabled or disabled. This functionality could be packaged as something like `FeatureGateChecker`.

The check is marked as a `warning` anyway, so it won't make the checks fail. And the error returned is informing the user why it didn't work, which can lead him to enable that feature gate (he wouldn't know if we just skip it). Also, the minimum k8s version supported is `1.12`, in which `DryRun` is alpha. Starting `1.13` it should be beta and be enabled by default, so eventually this may not even be a problem at all.

charts/linkerd2/templates/web-rbac.yaml
tegioz: Good point @zaharidichev, that could be interesting.

grampelberg: I would prefer to keep `check` as zero config right now. I think that if someone does a skip they know what they're doing (and we can address that through the docs).

grampelberg: As an added note, this check is primarily for users who have api-servers that cannot call into the cluster because of firewall rules (GKE private clusters, unique EKS deploys). A `default` check will verify that and be good enough (tm) for most folks.

grampelberg:
Oh man, this is so awesome! There are quite a few users who will really appreciate this check.

> This functionality could be packaged as something like FeatureGateChecker.

I think having this functionality will be pretty great. We're starting to run into features that are only enabled on modern clusters which we'd like to use and can't because we have to stick to the lowest common denominator right now. That said, I think we should tackle that as a separate PR and spend some time thinking about what we'd like to do. I'm also a little unconvinced it is possible, last time I played around with this, it required getting the api-server flags themselves and it wasn't possible to do in an automated fashion.

> Starting 1.13 it should be beta and be enabled by default, so eventually this may not even be a problem at all.

I'm good bumping our required version up to 1.13 as 1.14 is the earliest version supported by k8s itself right now.

alpeb:
Also please rebase against master one more time to have those tests pass.

grampelberg:
@adleong I thought we'd split this into two pieces - checking connectivity and actually running a test. I'm in agreement with you that we shouldn't be modifying cluster state from the CLI currently. Let's chat in person to get a better grasp on what's going on here.

tegioz:
Hi @adleong 

I've created a new checks category called `LinkerdControlPlaneDryRunChecks` and added the injector check to it as suggested in #3701.

At the moment it's enabled by default for the CLI, but users can opt-out using `--dry-run=false`. Happy to leave it disabled by default if that makes more sense. On the dashboard side, I've included this new category but we can remove it if that fits better.

If this approach works I can do something similar for #3701. 

Thanks!

charts/linkerd2/templates/web-rbac.yaml
```diff
@@ -102,6 +102,36 @@ subjects:
   name: linkerd-web
   namespace: {{.Values.Namespace}}
 ---
+apiVersion: rbac.authorization.k8s.io/v1
```
adleong: if the web is not including the injector dry run check, it should no longer need these RBAC permissions, correct?

cli/cmd/check.go
```diff
@@ -46,6 +48,7 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	flags.BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
+	flags.BoolVar(&options.dryRun, "dry-run", options.dryRun, "Include checks that rely on k8s dry-run (they may require special write permissions)")
```
adleong: @grampelberg if we bump our minimum supported K8s version, do you think we should still include this escape hatch to disable dry run checks?  If so, this flag should also apply to the dry run check added in https://github.com/linkerd/linkerd2/pull/3701

grampelberg: WDYT about removing the escape hatch for now? If a bunch of folks run into issues with it, it is easy to put back in and in the meantime we don't have to maintain the code unless it is absolutely required.

adleong: works for me

charts/linkerd2/templates/web-rbac.yaml
tegioz: It's included at the moment by default in both the CLI and the dashboard (and can be disabled in the CLI by the user using the new flag). But it's easy to remove now if you prefer this one off in the dashboard.

cli/cmd/check.go
tegioz: 👍 Will remove it then, thanks.

charts/linkerd2/templates/web-rbac.yaml
adleong: yeah, I think we should remove it from the dashboard in order to avoid needing to give "create" permissions to web

tegioz: Done 👍

grampelberg:
I'd like to know a little bit better where this *is* supported. If everyone handles it fine except AKS, then let's make this an info check - eg "could not dry run on this cluster, you'll want to manually verify that the injector is running".

If the majority of the cloud providers cannot do a dry run for this object type, let's either drop the check entirely or figure out a different way to implement it.

tegioz:
`DryRun` is enabled by default for Amazon EKS clusters.

https://docs.aws.amazon.com/eks/latest/userguide/eks-ug.pdf

Issue #3617, opened 2019-10-21
Update dashboard to latest React


scottcarol:
The dashboard is currently on React `16.5.0`. React can be updated to `16.9.0`.

Updating to the latest React means that we will need to stop using from the `react-url-query` library, which we are using to encode/decode URLs in the Tap and Top Routes views. 

`react-url-query` utilizes the `componentWillMount` lifecycle method which will no longer be supported by React.

The creator of the react-url-query created a [use-query-params](https://github.com/pbeshai/use-query-params#readme) hook which we can use instead.

tzsword:
🙋🏻‍♂️

scottcarol:
Hey tzsword! 👋 Glad you're interested in working on the dashboard! This issue is already assigned, but are you interested in working on another React ticket? #2915 and #2996 are available.

tzsword:
@scottcarol 👋 i will try #2915

scottcarol:
@cynthia-sg I will not be able to get to this until after KubeCon so if you would like this issue it's yours 😄

PR #3710, opened 2019-11-13
Update dashboard to Material 4

### LINKED ISSUE #3592: Update the dashboard to Material 4

cynthia-sg:
This PR updates Material UI to version 4.7.1 and Material UI icons to 4.5.1. Some of the breaking changes fixed are: 
- `theme.spacing.unit` usage is deprecated
- spacing API changes -> from `[0, 8, 16, 24, 32, 40]` to `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`
- `numeric` prop in TableCell is deprecated
- `dense` mode has been moved to `size` property. The result (please, check screenshot) doesn’t look like the current table look, so we are not using `size` and we have styled the tableHead a little bit
- `useNextVariants` can be removed safely
- `Snackbar` has changed the default transition from Slide to Grow
-  Rename the `headlineMapping` property to `variantMapping`
- `Select` component - value:  this prop is required when the native prop is false (default). Set to an empty string '' if you don't want any of the available options to be selected. 
- Tested on Safari, Chrome, Firefox and Edge (for Mac)

**NOTE**: Material UI 4 requires the minimum version `react@^16.8.0`, so this PR has to be merged after #3617

Differences before fixing tables and navigation menu:
MUI_4_00
MUI_4_00a

Table using size="small":
MUI_4_01

Some screenshots with the final result:
MUI_4_02
MUI_4_03
MUI_4_04
MUI_4_05


cynthia-sg:
@scottcarol, tests are not passing due to React has to be upgraded at least to 16.8.0 version.

scottcarol:
@cynthia-sg let us know when this is ready to review. Thanks!

cynthia-sg:
**NOTE about tests**: 

After updating Material UI, some tests started failing raising several errors: 
#### Error 1

```
Getting Enzyme Internal Error: unknown node with tag 14
```

Solved updating the `enzyme-adapter-react-16` library to version 1.15.0.

#### Error 2

```
Method “simulate” is only meant to be run on a single node. 0 found instead.

—

Expected <BrowserRouter> to contain "A new version (2.3.4) is available." but it did not.
    Expected HTML: "A new version (2.3.4) is available."
    Actual HTML: ""
```

These errors were showing up when the content was wrapped with a `Hidden` component, because it wasn’t being rendered when the component was mounted (more details here: https://github.com/airbnb/enzyme/issues/2179).

The following solutions have been applied in the tests to address this problem:

- Add tests in children components where the `Hidden` wrap doesn’t apply, reducing the need to inspect the children rendered markup from the parent.
- Use `shallow` instead of `mount` to render the components hierarchy. When using shallow, the problem described above with the wrapped Hidden component is not occurring as all components are being rendered.
- To not remove some other existing tests, only the state and props of some children components is tested, instead of testing their events and methods. The first point in this list should take care of testing those remaining bits though.

#### Some other details to keep in mind

While I was trying to fix the tests with `FilterListIcon`, I saw that I was not able to `find` the icon out of the `Hidden` component neither. So, I was investigating a little this issue and I found another bug:

```
https://github.com/mui-org/material-ui/issues/15928
https://github.com/airbnb/enzyme/issues/2146
```

I think we are doing an incorrect use of `toBeDefined()` method: when we are checking `expect(component.find("Child").toBeDefined()`, this check is going to be always true due to `find` method returns an element `ShallowWrapper` which wraps the found nodes (independently of the existing nodes number).

scottcarol:
Hi @cynthia-sg, thanks for the detailed description of the test fixes.

To clarify, are the unit tests changes you added to `BaseTable.test.js`, `MetricsTable.test.js`, and the `enzyme` update, directly related to which version of Material-UI we are running? Or could they be added to the tests using our current version of Material-UI? If they could be added now, I would prefer those to be a separate PR and merged first. If not, then I agree they do belong in this PR.

cynthia-sg:
> Hi @cynthia-sg, thanks for the detailed description of the test fixes.
> 
> To clarify, are the unit tests changes you added to `BaseTable.test.js`, `MetricsTable.test.js`, and the `enzyme` update, directly related to which version of Material-UI we are running? Or could they be added to the tests using our current version of Material-UI? If they could be added now, I would prefer those to be a separate PR and merged first. If not, then I agree they do belong in this PR.

Hi @scottcarol, the tests can be changed in a separate PR (only I'll need to change a little bit the new test for `BaseTable`. So, I'm going to create that new PR with `enzyme-adapter-react-16` update and all the tests, and I'll remove from this PR all tests changes (although this means that this PR is not going to pass the tests until the new PR is merged and rebased here).

cynthia-sg:
> Thanks so much for doing this work! It will be great to have this library updated. 💪
> 
> In general things look great. There were two main differences I noticed between this branch and `master` -- if those can be fixed, I think we should be good to go:
> 
> 1. The main content, on the right of the sidebar, is no longer centered in its container.
> 
> `master`:
> Screen Shot 2019-12-12 at 12 24 02 PM
> 
> this branch:
> Screen Shot 2019-12-12 at 12 24 09 PM
> 
> 1. If table column titles are longer than one line, there is now a space between the first line and second line, for example, `Success Rate`.
> 
> `master`:
> master
> 
> this branch:
> current

Done 

Screenshot 2019-12-13 at 14 25 53

scottcarol:
Thanks for changing that! I noticed one more difference -- the shadow effect around the table seems to have changed -- see the top, left and right side. Is it possible to restore that? 

`master`
Screen Shot 2019-12-13 at 4 08 13 PM


this branch:
Screen Shot 2019-12-13 at 4 08 22 PM

cynthia-sg:
> Thanks for changing that! I noticed one more difference -- the shadow effect around the table seems to have changed -- see the top, left and right side. Is it possible to restore that?
> 
> `master`
> Screen Shot 2019-12-13 at 4 08 13 PM
> 
> this branch:
> Screen Shot 2019-12-13 at 4 08 22 PM

Good catch! I missed this:

```
Migration 3 to 4 (Material UI):
[Paper] Reduce the default elevation. Change the default Paper elevation to match the Card and the Expansion Panel:
-<Paper />
+<Paper elevation={2} />
This affects the ExpansionPanel as well.
```

scottcarol:
This looks 👍 and is mergeable once @cynthia-sg rebases -- thanks @cynthia-sg!

web/app/js/components/Navigation.test.jsx
```diff
@@ -167,11 +183,15 @@ describe('Namespace Select Button', () => {
       </BrowserRouter>
     );
 
-    const button = component.find("Button");
+    const button = component.find("button");
     expect(button).toIncludeText("emojivoto");
   });
 
   it('opens the Namespace Selection menu if button is clicked', () => {
+    // Material UI - Popover needs a proper anchorEl if NOVE_ENV is different to "test"
```
cynthia-sg: `Popover.js` - Material UI
```javascript
/**
   * This is the DOM element, or a function that returns the DOM element,
   * that may be used to set the position of the popover.
   */
  anchorEl: (0, _utils.chainPropTypes)(_propTypes.default.oneOfType([_propTypes.default.object, _propTypes.default.func]), function (props) {
    if (props.open && (!props.anchorReference || props.anchorReference === 'anchorEl')) {
      var resolvedAnchorEl = getAnchorEl(props.anchorEl);
      var containerWindow = (0, _ownerWindow.default)(resolvedAnchorEl);

      if (resolvedAnchorEl instanceof containerWindow.Element) {
        var box = resolvedAnchorEl.getBoundingClientRect();

        if (process.env.NODE_ENV !== 'test' && box.top === 0 && box.left === 0 && box.right === 0 && box.bottom === 0) {
          return new Error(['Material-UI: the `anchorEl` prop provided to the component is invalid.', 'The anchor element should be part of the document layout.', "Make sure the element is present in the document or that it's not display none."].join('\n'));
        }
      } else {
        return new Error(['Material-UI: the `anchorEl` prop provided to the component is invalid.', "It should be an Element instance but it's `".concat(resolvedAnchorEl, "` instead.")].join('\n'));
      }
    }

    return null;
  }),
```

scottcarol: Mind changing `NOVE_ENV` to `NODE_ENV`?

I would change the whole comment to be: 
```
// Material-UI v4 requires that the popover needs a valid anchorEl if NODE_ENV is not "test"
```

cynthia-sg: `Popover.js` - Material UI
```javascript
/**
   * This is the DOM element, or a function that returns the DOM element,
   * that may be used to set the position of the popover.
   */
  anchorEl: (0, _utils.chainPropTypes)(_propTypes.default.oneOfType([_propTypes.default.object, _propTypes.default.func]), function (props) {
    if (props.open && (!props.anchorReference || props.anchorReference === 'anchorEl')) {
      var resolvedAnchorEl = getAnchorEl(props.anchorEl);
      var containerWindow = (0, _ownerWindow.default)(resolvedAnchorEl);

      if (resolvedAnchorEl instanceof containerWindow.Element) {
        var box = resolvedAnchorEl.getBoundingClientRect();

        if (process.env.NODE_ENV !== 'test' && box.top === 0 && box.left === 0 && box.right === 0 && box.bottom === 0) {
          return new Error(['Material-UI: the `anchorEl` prop provided to the component is invalid.', 'The anchor element should be part of the document layout.', "Make sure the element is present in the document or that it's not display none."].join('\n'));
        }
      } else {
        return new Error(['Material-UI: the `anchorEl` prop provided to the component is invalid.', "It should be an Element instance but it's `".concat(resolvedAnchorEl, "` instead.")].join('\n'));
      }
    }

    return null;
  }),
```

PR #3716, opened 2019-11-14
Added the ability to define Identity.External in order to use an external secret for issuer certs

yarinm:
Let me know if further clarifications are needed here, this is pretty straight forward.

The use case here is the need to define the issuer secret using certs-manager / other way.

### DIFF
--- a/charts/linkerd2/templates/identity.yaml
+++ b/charts/linkerd2/templates/identity.yaml
@@ -4,7 +4,7 @@
 ###
 ### Identity Controller Service
 ###
-{{ if and (.Identity.Issuer) (eq .Identity.Issuer.Scheme "linkerd.io/tls") -}}
+{{ if and (not .Identity.External) (.Identity.Issuer) (eq .Identity.Issuer.Scheme "linkerd.io/tls") -}}
 ---
 kind: Secret
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


yarinm:
Hey @zaharidichev I wasn't aware that the scheme changes,

I think the current flag isn't very clear on what it does. And I wanted to remove the secret altogether and not change its scheme. we can consolidate the scheme value to be toggled by the new boolean flag. WDYT?

zaharidichev:
@yarinm There is a [PR with documentation](https://github.com/linkerd/website/pull/574) around that feature which will be merged an up on the website quite soon. 

The reason we chose a scheme represented as a `string` instead of a `bool` is to give ourselves a bit more flexibility and be able to add different kind of secret formats down the road. In that sense, since we have the scheme now that can be used to control things, I am not sure that adding a `bool` would give us any kind of tangible benefit. Or perhaps I am wrong? 

To put it another way. Is there something that you are trying to achieve at the moment that cannot be achieved by setting the Scheme but would be possible if you had a bool there?

yarinm:
You're right it's not needed, the scheme will suffice for me. I'm not using cert-manager at the moment, but using the kuberentes.io/tls scheme is good enough as I can just create the secret myself using our secrets management system.

zaharidichev:
@yarinm Good to know all that will work! Let us know if you find any quirks that need fixing :)

PR #3718, opened 2019-11-14
Remove redundant serviceprofile check

alenkacz:
This is obsolete as there's already a validating webhook in place.


alenkacz:
I think the test failure is unrelated..

PR #3722, opened 2019-11-14
Temporarily disable kind cleanup step

alpeb:
This is temporary, to debug this problem:
https://github.com/linkerd/linkerd2/runs/302886802#step:7:599


alpeb:
I'm forced to do this in #3718 already, so this one can be closed.

PR #3727, opened 2019-11-14
Disable TestCliStatForLinkerdNamespace integration test

adleong:
https://github.com/linkerd/linkerd2/pull/3693 caused the proxy to start resolving private IP addresses with the destination service.  However, the destination service does not support IP lookups and returns failures for these lookups.  This negatively affects the destination service success rate and can cause this test to fail.  We disable this test for now until the destination service supports IP lookups.

### DIFF
--- a/test/stat/stat_test.go
+++ b/test/stat/stat_test.go
@@ -1,10 +1,7 @@
 package get
 
 import (
-	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


alpeb:
Ah you need to comment out `validateRowStats` as well for the linter to pass...

Issue #3332, opened 2019-08-27
Fix flakey TestGetProfiles


siggy:
in travis:
https://travis-ci.org/linkerd/linkerd2/jobs/577582791#L749-L753

locally:
```bash
$ while go test -count 1 github.com/linkerd/linkerd2/controller/api/destination; do ls > /dev/null ; done
ok  	github.com/linkerd/linkerd2/controller/api/destination	0.586s
...
ok  	github.com/linkerd/linkerd2/controller/api/destination	0.602s
time="2019-08-27T16:06:13-07:00" level=error msg="A request match must have a field set" test=0x1ec8140
time="2019-08-27T16:06:13-07:00" level=error msg="A response match must have a field set" test=0x1ec8140
time="2019-08-27T16:06:13-07:00" level=error msg="Range maximum cannot be smaller than minimum" test=0x1ec8140
time="2019-08-27T16:06:13-07:00" level=info msg="waiting for caches to sync"
time="2019-08-27T16:06:13-07:00" level=info msg="caches synced"
time="2019-08-27T16:06:13-07:00" level=info msg="waiting for caches to sync"
time="2019-08-27T16:06:13-07:00" level=info msg="caches synced"
time="2019-08-27T16:06:13-07:00" level=info msg="Establishing watch on endpoint [ns/name1:8989]" component=endpoints-watcher test=0x1ec8140
time="2019-08-27T16:06:13-07:00" level=warning msg="failed to retrieve replicaset from k8s ns/rs-1: replicaset.apps \"rs-1\" not found"
time="2019-08-27T16:06:13-07:00" level=info msg="Stopping watch on endpoint [ns/name1:8989]" component=endpoints-watcher test=0x1ec8140
time="2019-08-27T16:06:13-07:00" level=info msg="waiting for caches to sync"
time="2019-08-27T16:06:13-07:00" level=info msg="caches synced"
time="2019-08-27T16:06:13-07:00" level=info msg="waiting for caches to sync"
time="2019-08-27T16:06:13-07:00" level=info msg="caches synced"
time="2019-08-27T16:06:13-07:00" level=info msg="Establishing watch on service ns/name1" component=traffic-split-watcher test=0x1ec8140
time="2019-08-27T16:06:13-07:00" level=info msg="Establishing watch on profile other/name1.ns.svc.mycluster.local" component=profile-watcher test=0x1ec8140
[...]

PR #3730, opened 2019-11-15
Unit and integration test fixups

adleong:
- Added cleanup step at the end of all integration tests.
- Disable external_issuer_integration_tests in cloud_tests due to
  namespace issue. Running this via `kind` tests is sufficient for now.
- Set a flakey test to `Skip`, relates to #3332.

### DIFF
--- a/bin/_test-run.sh
+++ b/bin/_test-run.sh
@@ -62,11 +62,13 @@ function deep_integration_tests() {
 
     run_test "$(go list $test_directory/.../...)" --linkerd-namespace=$linkerd_namespace
     exit_on_err "error during deep tests"
+    cleanup
 }
 
 function custom_domain_integration_tests() {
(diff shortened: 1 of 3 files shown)


siggy:
A note on the `docker_deploy` step failing on `bin/docker-push-deps`. For some reason the `~/.ssh/known_hosts` was either not being generated or getting blown away at some point. A couple of debug commands confirm the file is empty, and subsequent ssh commands fail because of it:
```bash
$ cat ~/.ssh/known_hosts
$ ssh -v github@$DOCKER_ADDRESS ls -al
...
debug1: read_passphrase: can't open /dev/tty: No such device or address
```

Full log: https://github.com/linkerd/linkerd2/pull/3730/checks?check_run_id=304493671#step:6:47

PR #3731, opened 2019-11-15
Add check to ensure kube-system has the needed annotations (HA)

### LINKED ISSUE #3721: linkerd check should fail if kube-system doesn't have annotations/labels to disable pod injection
## Feature Request

if the kube-system namespace doesn't have the following:

```yaml

  annotations:
    linkerd.io/inject: disabled
  labels:
    config.linkerd.io/admission-webhooks: disabled
```

`linkerd check` should fail.

I had a failure scenario similar to https://github.com/linkerd/linkerd2/issues/3606 (I was destructively testing a cluster by terminating & disabling the kubelet service and powering off the nodes).

Here's the journey of about 30 mins of troubleshooting, are you sat comfortably?

<hr />

```
NAME    STATUS     ROLES   AGE   VERSION
aks-0   NotReady   agent   35h   v1.14.8
aks-1   Ready      agent   36h   v1.14.8
aks-2   NotReady   agent   28h   v1.14.8
aks-3   NotReady   agent   28h   v1.14.8
aks-4   NotReady   agent   28h   v1.14.8
```

```
× can initialize the client
    error upgrading connection: error dialing backend: dial tcp 10.183.0.4:10250: i/o timeout
    see https://linkerd.io/checks/#l5d-existence-client for hints
```

☝️ that's a prometheus node exporter 🤷‍♂  

```
replicaset-controller  Error creating: Internal error occurred: failed calling webhook "linkerd-proxy-injector.linkerd.io": Post https://linkerd-proxy-injector.linkerd.svc:443/?timeout=30s: dial tcp 172.16.1.84:443: i/o timeout
```

☝️  all replica/daemon sets are saying something like that

```
linkerd-proxy-injector   ClusterIP   172.16.1.84
```

the service ip is correct, so dns is somewhat working 😁 

```
linkerd-proxy-injector   10.183.0.153:8443,10.183.1.150:8443,10.183.1.70:8443
```

the endpoint ips are correct for the pods and the proxy injector pods are up

```
 k -n linkerd logs linkerd-proxy-injector-58bcfcfbd5-nmhkb proxy-injector
Error from server: Get https://[node-name]:10250/containerLogs/linkerd/linkerd-proxy-injector-58bcfcfbd5-nmhkb/proxy-injector: dial tcp 10.183.0.117:10250: i/o timeout
```

the *kube-system* namespace didn't have `config.linkerd.io/admission-webhooks: disabled` or `linkerd.io/inject: disabled` and their pods can now restart

after a few minutes, everything else came back online

zaharidichev:
Adds a check to ensure kube-system namespace has `linkerd.io/inject: disabled`  and `linkerd.io/inject: disabled`

FIxes #3721

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -128,6 +128,10 @@ const (
 	// from LinkerdVersionChecks, so those checks must be added first.
 	LinkerdDataPlaneChecks CategoryID = "linkerd-data-plane"
 
+	// LinkerdHaChecks adds checks to validate that the HA configuration
+	// is correct. These checks are no ops if linkerd is not in HA mode
+	LinkerdHaChecks CategoryID = "linkerd-ha-checks"
+
 	// linkerdCniResourceLabel is the label key that is used to identify
 	// whether a Kubernetes resource is related to the install-cni command
 	// The value is expected to be "true", "false" or "", where "false" and
@@ -593,14 +597,6 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkPodSecurityPolicies(true)
 					},
 				},
-				{
-					description: "pod injection disabled on kube-system (HA only)",
-					hintAnchor:  "l5d-injection-disabled",
-					warning:     true,
-					check: func(context.Context) error {
-						return hc.checkHAMetadataPresentOnKubeSystemNamespace()
-					},
-				},
 			},
 		},
 		{
@@ -973,6 +969,19 @@ func (hc *HealthChecker) allCategories() []category {
 							}
 						}
 						return nil
+					},
+				},
+			},
+		},
+		{
+			id: LinkerdHaChecks,
+			checkers: []checker{
+				{
+					description: "pod injection disabled on kube-system",
+					hintAnchor:  "l5d-injection-disabled",
+					warning:     true,
+					check: func(context.Context) error {
+						return hc.checkHAMetadataPresentOnKubeSystemNamespace()
 					},
 				},
 			},
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -220,6 +220,16 @@ func IsCategoryError(err error, categoryID CategoryID) bool {
 	return false
 }
 
+// SkipError is returned by a check in case this check needs to be ignored.
+type SkipError struct {
+	Reason string
+}
+
+// Error satisfies the error interface for SkipError.
+func (e *SkipError) Error() string {
+	return e.Reason
+}
+
 type checker struct {
 	// description is the short description that's printed to the command line
 	// when the check is executed
@@ -273,10 +283,9 @@ type CheckResult struct {
 type CheckObserver func(*CheckResult)
 
 type category struct {
-	id        CategoryID
-	checkers  []checker
-	enabled   bool
-	shouldRun func() bool
+	id       CategoryID
+	checkers []checker
+	enabled  bool
 }
 
 // Options specifies configuration for a HealthChecker.
@@ -982,11 +991,13 @@ func (hc *HealthChecker) allCategories() []category {
 					hintAnchor:  "l5d-injection-disabled",
 					warning:     true,
 					check: func(context.Context) error {
-						return hc.checkHAMetadataPresentOnKubeSystemNamespace()
+						if hc.isHA() {
+							return hc.checkHAMetadataPresentOnKubeSystemNamespace()
+						}
+						return &SkipError{Reason: "not run for non HA installs"}
 					},
 				},
 			},
-			shouldRun: hc.isHA,
 		},
 	}
 }
@@ -1029,32 +1040,29 @@ func (hc *HealthChecker) RunChecks(observer CheckObserver) bool {
 
 	for _, c := range hc.categories {
 		if c.enabled {
-			if c.shouldRun == nil || c.shouldRun() {
-				for _, checker := range c.checkers {
-					checker := checker // pin
-					if checker.check != nil {
-						if !hc.runCheck(c.id, &checker, observer) {
-							if !checker.warning {
-								success = false
-							}
-							if checker.fatal {
-								return success
-							}
+			for _, checker := range c.checkers {
+				checker := checker // pin
+				if checker.check != nil {
+					if !hc.runCheck(c.id, &checker, observer) {
+						if !checker.warning {
+							success = false
 						}
-					}
-
-					if checker.checkRPC != nil {
-						if !hc.runCheckRPC(c.id, &checker, observer) {
-							if !checker.warning {
-								success = false
-							}
-							if checker.fatal {
-								return success
-							}
+						if checker.fatal {
+							return success
 						}
 					}
 				}
 
+				if checker.checkRPC != nil {
+					if !hc.runCheckRPC(c.id, &checker, observer) {
+						if !checker.warning {
+							success = false
+						}
+						if checker.fatal {
+							return success
+						}
+					}
+				}
 			}
 		}
 	}
@@ -1067,6 +1075,10 @@ func (hc *HealthChecker) runCheck(categoryID CategoryID, c *checker, observer Ch
 		ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
 		defer cancel()
 		err := c.check(ctx)
+		if se, ok := err.(*SkipError); ok {
+			log.Debugf("Skipping check: %s. Reason: %s", c.description, se.Reason)
+			return true
+		}
 		if err != nil {
 			err = &CategoryError{categoryID, err}
 		}
@@ -1099,6 +1111,10 @@ func (hc *HealthChecker) runCheckRPC(categoryID CategoryID, c *checker, observer
 	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
 	defer cancel()
 	checkRsp, err := c.checkRPC(ctx)
+	if se, ok := err.(*SkipError); ok {
+		log.Debugf("Skipping check: %s. Reason: %s", c.description, se.Reason)
+		return true
+	}
 	if err != nil {
 		err = &CategoryError{categoryID, err}
 	}
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1581,7 +1581,7 @@ func (hc *HealthChecker) checkHAMetadataPresentOnKubeSystemNamespace() error {
 
 	val, ok := ns.Labels[k8s.AdmissionWebhookLabel]
 	if !ok || val != "disabled" {
-		return fmt.Errorf("kube-system namespace needs to have %s: disabled if HA mode is enabled", k8s.AdmissionWebhookLabel)
+		return fmt.Errorf("kube-system namespace needs to have the label %s: disabled if HA mode is enabled", k8s.AdmissionWebhookLabel)
 	}
 
 	return nil
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -593,6 +593,14 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkPodSecurityPolicies(true)
 					},
 				},
+				{
+					description: "pod injection disabled on kube-system (HA only)",
+					hintAnchor:  "l5d-injection-disabled",
+					warning:     true,
+					check: func(context.Context) error {
+						return hc.checkHAMetadataPresentOnKubeSystemNamespace()
+					},
+				},
 			},
 		},
 		{
@@ -1265,6 +1273,15 @@ func (hc *HealthChecker) checkClusterRoleBindings(shouldExist bool) error {
 	return checkResources("ClusterRoleBindings", objects, hc.expectedRBACNames(), shouldExist)
 }
 
+func (hc *HealthChecker) isHA() bool {
+	for _, flag := range hc.linkerdConfig.GetInstall().GetFlags() {
+		if flag.GetName() == "ha" && flag.GetValue() == "true" {
+			return true
+		}
+	}
+	return false
+}
+
 func (hc *HealthChecker) isHeartbeatDisabled() bool {
 	for _, flag := range hc.linkerdConfig.GetInstall().GetFlags() {
 		if flag.GetName() == "disable-heartbeat" && flag.GetValue() == "true" {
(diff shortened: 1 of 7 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -600,6 +600,14 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkPodSecurityPolicies(true)
 					},
 				},
+				{
+					description: "pod injection disabled on kube-system (HA only)",
+					hintAnchor:  "l5d-existence-psp",
```
grampelberg: Shouldn't the hint be something like `l5d-injection-disabled`?

pkg/healthcheck/healthcheck.go
```diff
@@ -1353,6 +1370,27 @@ func (hc *HealthChecker) checkCanPerformAction(verb, namespace, group, version,
 	)
 }
 
+func (hc *HealthChecker) checkHAMetadataPresentOnKubeSystemNamespace() error {
+	if hc.isHA() {
```
grampelberg: It is probably out of scope for this PR, but should we move the HA part to the check definition itself?

pkg/healthcheck/healthcheck.go
```diff
@@ -600,6 +600,14 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkPodSecurityPolicies(true)
 					},
 				},
+				{
+					description: "pod injection disabled on kube-system (HA only)",
```
grampelberg: Does it make sense to introduce a HA section? There'd only be one check there right now, so it might be a little weird.

zaharidichev: I toyed with this idea and reached the same conclusion. We can do it, if you think it makes sense though.

pkg/healthcheck/healthcheck.go
zaharidichev: This will make testing the case when we do not have ha enabled a bit cumbersome I think

pkg/healthcheck/healthcheck.go
adleong: could we programmatically add this check to the list of checks only when HA is enabled?  or is the config not yet available at that point?

pkg/healthcheck/healthcheck.go
```diff
@@ -1353,6 +1370,27 @@ func (hc *HealthChecker) checkCanPerformAction(verb, namespace, group, version,
...
+			return err
+		}
+
+		val, ok := ns.Labels[k8s.AdmissionWebhookLabel]
+		if !ok || val != "disabled" {
+			return fmt.Errorf("kube-system namespace needs to have %ss: disabled if HA mode is enabled", k8s.AdmissionWebhookLabel)
+		}
+
+		val, ok = ns.Annotations[k8s.ProxyInjectAnnotation]
+		if !ok || val != "disabled" {
```
adleong: is this actually required?  the documentation only mentions the `config.linkerd.io/admission-webhooks` label.  https://linkerd.io/2/features/ha/

pkg/healthcheck/healthcheck.go
zaharidichev: Yes, unfortunately the config is not there at that point.

pkg/k8s/labels.go
```diff
@@ -246,6 +246,9 @@ const (
 	// TapServiceName is the name of the tap APIService
 	TapServiceName = "linkerd-tap"
 
+	// AdmissionWebhookLabel indicates whether admission webhooks are enabled for a namespace
+	AdmissionWebhookLabel = ProxyConfigAnnotationsPrefix + "/admission-webhooks"
```
alpeb: TIOLI, or a separate PR: You could update `namespace.yaml`, `proxy-injector-rbac.yaml` and `sp-validator-rbac.yaml`  with this new constant...

zaharidichev:
@alpeb I thought I fixed that but obviously I am wrong. Will fix it shortly!

pkg/healthcheck/healthcheck.go
```diff
@@ -1353,6 +1379,22 @@ func (hc *HealthChecker) checkCanPerformAction(verb, namespace, group, version,
...
+func (hc *HealthChecker) checkHAMetadataPresentOnKubeSystemNamespace() error {
+	if hc.isHA() {
+		ns, err := hc.kubeAPI.CoreV1().Namespaces().Get("kube-system", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		val, ok := ns.Labels[k8s.AdmissionWebhookLabel]
+		if !ok || val != "disabled" {
+			return fmt.Errorf("kube-system namespace needs to have %ss: disabled if HA mode is enabled", k8s.AdmissionWebhookLabel)
```
alpeb: It appears one extra `s` sneaked in

pkg/healthcheck/healthcheck_test.go
```diff
@@ -2124,6 +2124,90 @@ metadata:
...
+		},
+		{
+			"passes when HA is enabled and namespace has required metadata",
+			getConfigAndKubeSystemNamespace(true, "config.linkerd.io/admission-webhooks: disabled"),
+			nil,
+		},
+		{
+			"fails when HA and admission hooks are enabled",
+			getConfigAndKubeSystemNamespace(true, "config.linkerd.io/admission-webhooks: enabled"),
+			errors.New("kube-system namespace needs to have config.linkerd.io/admission-webhookss: disabled if HA mode is enabled"),
```
alpeb: extra `s` here, and in the next element as well

zaharidichev:
@adleong @alpeb  I actually made the ha checks category dynamic. So it only works if we are in HA mode. I think that works out better Glad to know what you think

pkg/healthcheck/healthcheck_test.go
```diff
@@ -2124,6 +2126,95 @@ metadata:
...
+			var err error
+			hc.kubeAPI, _ = k8s.NewFakeAPI(tc.k8sConfigs...)
+			_, hc.linkerdConfig, err = hc.checkLinkerdConfigConfigMap()
+			if err != nil {
+				t.Fatalf("Unexpected error: %q", err)
+			}
+
+			hc.addCheckAsCategory("l5d-injection-disabled", LinkerdHaChecks, "pod injection disabled on kube-system", func() bool {
+				return hc.isHA()
+			})
```
alpeb: This might also be simplified by passing `hc.isHA`

pkg/healthcheck/healthcheck.go
```diff
@@ -110,6 +110,10 @@ const (
 	// from LinkerdVersionChecks, so those checks must be added first.
 	LinkerdDataPlaneChecks CategoryID = "linkerd-data-plane"
 
+	// LinkerdHaChecks adds checks to validate that the HA configuration
+	// is correct. These checks are no ops if linkerd is not in HA mode
+	LinkerdHaChecks CategoryID = "linkerd-ha-checks"
```
ihcsim: Nit: Can we uppercase the `Ha` in the variable name? See Go convention [here](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

pkg/healthcheck/healthcheck.go
```diff
@@ -251,9 +255,10 @@ type CheckResult struct {
 type CheckObserver func(*CheckResult)
 
 type category struct {
-	id       CategoryID
-	checkers []checker
-	enabled  bool
+	id        CategoryID
+	checkers  []checker
+	enabled   bool
```
ihcsim: This feels a bit redundant. What are the benefits of introducing `shouldRun`? Is it possible to determine the HA condition on the caller side? AFAICT, it has been that the clients (`install`, `check`, `public_api`) decide which checks to include, after determining the check categories based on some conditions.

zaharidichev: So I guided myself only by the check.go use of this class but the way I see it is that the decisions of which checks to include is based solely on the command line option. To make that depend on for example whether the installation is HA or not, we need to fetch the config map in advanced and potentially fail or default to false if we cannot get the config map. But this is already part of the check sequence tself and in fact comes later than some other checks. Namely: 


```
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-config
--------------

linkerd-existence
-----------------
√ 'linkerd-config' config map exists.  <----- this here
``` 

So requiring to have the config map upfront in order to understand what checks we need to include, while some of the checks checking whether this config map exists at all renders the whole thing a bit leaky in my mind. Feel free to disagree and we can always change it :)

adleong: this isn't the first time I've wanted to conditionally skip a check depending on the current state.  what do you think about having a special error value called skip?  if a check wishes to be skipped, it can return this value.  the check runner can check the error, and skip over any checks that have returned skip.

ihcsim: > So requiring to have the config map upfront in order to understand what checks we need to include, while some of the checks checking whether this config map exists at all renders the whole thing a bit leaky in my mind. 

I'm not sure that's an issue though, because the CLI and some controllers already knows how to look up the configmap. I'd assume this HA check happens after the `linkerd-existence` check referred above.
```
⚡  grep -rniI fetchlinkerdconfigmap *                                         
cli/cmd/install.go:896: _, _, err = healthcheck.FetchLinkerdConfigMap(kubeAPI, controlPlaneNamespace)
cli/cmd/upgrade.go:214: _, configs, err := healthcheck.FetchLinkerdConfigMap(k, controlPlaneNamespace)
cli/cmd/profile.go:108:                 _, configs, err := healthcheck.FetchLinkerdConfigMap(k8sAPI, controlPlaneNamespace)
cli/cmd/install-sp.go:30:                       _, configs, err := healthcheck.FetchLinkerdConfigMap(k8sAPI, controlPlaneNamespace)
controller/heartbeat/heartbeat.go:25:   cm, _, err := healthcheck.FetchLinkerdConfigMap(kubeAPI, controlPlaneNamespace)
web/main.go:80: cm, _, err := healthcheck.FetchLinkerdConfigMap(k8sAPI, *controllerNamespace)
```

adleong: @ihcsim if we go that route, we can also clean up some awkward coupling where the linkerdConfig is set by one of the checks, imposing an implicit ordering constraint.  If we're going to look up the config in the check.go we should set it on the healthcheck from the beginning.

adleong: I agree with @ihcsim's suggestion of loading to configmap when we're building the list of checks instead of during the linkerd-config check.  This lets include the LinkerdHAChecks category only if the install is HA.

zaharidichev: Alright, this  makes sense +1 for the cleanup of checks fetching and setting the config map!

pkg/healthcheck/healthcheck.go
```diff
@@ -865,6 +870,20 @@ func (hc *HealthChecker) allCategories() []category {
...
+					description: "pod injection disabled on kube-system",
+					hintAnchor:  "l5d-injection-disabled",
+					warning:     true,
+					check: func(context.Context) error {
+						return hc.checkHAMetadataPresentOnKubeSystemNamespace()
+					},
+				},
+			},
+			shouldRun: hc.isHA,
+		},
```
cpretzer: This is a cool construct!

pkg/healthcheck/healthcheck.go
adleong: With this refactor, we no longer need the `shouldRun` field, right?

adleong: @zaharidichev is this correct?

zaharidichev: this is correct, I will update the code to reflect that

zaharidichev: @adleong @ihcsim Sorry for the back and forth on this one, but taking another look at it, I have some concerns. We can surely fetch the linkerd config map from the get go which pretty much entails the following: 
1. Initialize the kube config API
2. Query the config map
3. Parse it and determine whether we are in HA or not. 

Now that all makes sense but in doing so we get ahead of ourselves in terms of checks. Because the checks as they go are something like: 
- can initialize the client - tries to initialize the kubeApi ( and sets the kubeApi var in the HealthChecker struct)
- can query the Kubernetes API - checks whether the k8s cluster cna actually be queried
- linkerd-config' config map exists - fetches and sets the config map, etc reporting an error if it cannot do so. 

So if we fetch the config map from the get go we are effectivelly bypassing these couple of checks? What happens if we get an error? do we fail right away with an output format that is different that what a failed check would print?  Or do we fail silently so we let the actual checks run and report the problems in the usual format. In the latter case we are doing the same thing twice... 

So all in all it does not seem to me that moving the config map fetching and as a consequence the kube Api intialization out of the checks and thereby remove the checks dependency results in the akwardness being reduced. After all these two things are checks in themselves... Not sure what your thoughts on this are.

adleong: Yup, that sure is awkward.

I think the simplest thing to do is probably to have a special error struct called `skipped` which checks can return when they want to be skipped.  The run checks code would need to handle this error properly and not print a line for checks that return this value.  This ends up being very similar to having a `shouldRun` but in a more backwards compatible way and keeps all the logic for a check in one place.

zaharidichev: @adleong I swtiched to a custom error type. I agree that looks a lot neater. Feel free to take another look:)

pkg/healthcheck/healthcheck_test.go
```diff
@@ -2124,6 +2171,93 @@ metadata:
...
+  install: |
+    {"cliVersion":"dev-undefined","flags":[{"name":"ha","value":"%v"}]}`, ha),
+		fmt.Sprintf(`
+apiVersion: v1
+kind: Namespace
+metadata:
+  creationTimestamp: null
+  labels:
+    %s
+    linkerd.io/is-control-plane: "true"
```
alpeb: I believe you can get rid of this label

pkg/healthcheck/healthcheck_test.go
```diff
@@ -2124,6 +2171,93 @@ metadata:
...
+		},
+		{
+			"passes when HA is enabled and namespace has required metadata",
+			getConfigAndKubeSystemNamespace(true, "config.linkerd.io/admission-webhooks: disabled"),
+			"l5d-injection-disabled pod injection disabled on kube-system",
+		},
+		{
+			"fails when HA and admission hooks are enabled",
+			getConfigAndKubeSystemNamespace(true, "config.linkerd.io/admission-webhooks: enabled"),
+			"l5d-injection-disabled pod injection disabled on kube-system: kube-system namespace needs to have config.linkerd.io/admission-webhooks: disabled if HA mode is enabled",
```
alpeb: I would precise here that it needs to have the *label* `config.linkerd.io/admission-webhooks: disabled`

PR #3737, opened 2019-11-18
Update dashboard to React 16.11.0

### LINKED ISSUE #3617: Update dashboard to latest React

cynthia-sg:
**Problem**

The current version of React we use doesn’t support some features (i.e. hooks) that we’d like to use.

**Solution**

This PR updates `react` and `react-dom` to version 16.11.0, and `react-router` and `react-router-dom` to version 5.1.2. 

The following breaking changes have been fixed as part of the upgrade:

- Change deprecated `componentWillUpdate` to `componentDidUpdate`
- Replace `react-url-query` library by `use-query-params` due to the deprecation of some React lifecycle methods as suggested on #3617. This required some changes in: `Tap`, `Top`, `TapQueryForm` and `TopRoutes`.

**Validation**

To test, please run `bin/web setup` to update dependencies

Closes #3617 
Signed-off-by: Cintia Sanchez Garcia cynthiasg@icloud.com

### DIFF
--- a/web/app/js/components/Navigation.jsx
+++ b/web/app/js/components/Navigation.jsx
@@ -189,7 +189,7 @@ class NavigationBase extends React.Component {
     window.addEventListener("resize", this.updateWindowDimensions);
   }
 
-  componentWillUpdate() {
+  componentDidUpdate() {
     if (this.props.history) {
       this.props.checkNamespaceMatch(this.props.history.location.pathname);
     }
--- a/web/app/js/components/Tap.jsx
+++ b/web/app/js/components/Tap.jsx
@@ -1,4 +1,4 @@
-import { UrlQueryParamTypes, addUrlProps } from 'react-url-query';
+import { StringParam, withQueryParams } from 'use-query-params';
 import { WS_ABNORMAL_CLOSURE, WS_NORMAL_CLOSURE, WS_POLICY_VIOLATION, emptyTapQuery, processTapEvent, setMaxRps, wsCloseCodes } from './util/TapUtils.jsx';
 
 import ErrorBanner from './ErrorBanner.jsx';
@@ -17,7 +17,7 @@ import { groupResourcesByNs } from './util/MetricUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 
 const urlPropsQueryConfig = {
-  autostart: { type: UrlQueryParamTypes.string }
+  autostart: StringParam,
 };
 
 class Tap extends React.Component {
@@ -293,7 +293,7 @@ class Tap extends React.Component {
           resourcesByNs={this.state.resourcesByNs}
           authoritiesByNs={this.state.authoritiesByNs}
           updateQuery={this.updateQuery}
-          query={this.state.query} />
+          currentQuery={this.state.query} />
 
         <TapEventTable
           resource={this.state.query.resource}
@@ -303,4 +303,4 @@ class Tap extends React.Component {
   }
 }
 
-export default addUrlProps({ urlPropsQueryConfig })(withContext(Tap));
+export default withQueryParams(urlPropsQueryConfig, withContext(Tap));
--- a/web/app/js/components/TapQueryForm.jsx
+++ b/web/app/js/components/TapQueryForm.jsx
@@ -164,15 +164,18 @@ class TapQueryForm extends React.Component {
     };
 
     let shouldScopeAuthority = name === "namespace";
+    let newQueryValues = {};
 
     return event => {
       let formVal = event.target.value;
       state.query[name] = formVal;
+      newQueryValues[name] = formVal;
 
       if (!_isNil(scopeResource)) {
         // scope the available typeahead resources to the selected namespace
         state.autocomplete[scopeResource] = this.state.resourcesByNs[formVal];
         state.query[scopeResource] = `namespace/${formVal}`;
+        newQueryValues[scopeResource] = `namespace/${formVal}`;
       }
 
       if (shouldScopeAuthority) {
@@ -181,12 +184,13 @@ class TapQueryForm extends React.Component {
 
       this.setState(state);
       this.props.updateQuery(state.query);
-      this.handleUrlUpdate(state.query);
+      this.handleUrlUpdate(newQueryValues);
     };
   }
 
-  // Each time state.query is updated, this method calls the equivalent
-  // method to reflect the update in url query params.
+  // Each time state.query is updated, this method calls setQuery provided
+  // by useQueryParams HOC to partially update url query params that have
+  // changed
   handleUrlUpdate = query => {
     this.props.setQuery({ ...query });
   }
--- a/web/app/js/components/TapQueryForm.jsx
+++ b/web/app/js/components/TapQueryForm.jsx
@@ -1,4 +1,4 @@
-import { UrlQueryParamTypes, addUrlProps } from 'react-url-query';
+import { StringParam, withQueryParams } from 'use-query-params';
 import {
   defaultMaxRps,
   emptyTapQuery,
@@ -25,13 +25,11 @@ import React from 'react';
 import Select from '@material-ui/core/Select';
 import TextField from '@material-ui/core/TextField';
 import Typography from '@material-ui/core/Typography';
-import _each from 'lodash/each';
 import _flatten from 'lodash/flatten';
 import _isEmpty from 'lodash/isEmpty';
 import _isEqual from 'lodash/isEqual';
 import _isNil from 'lodash/isNil';
 import _map from 'lodash/map';
-import _mapValues from 'lodash/mapValues';
 import _merge from 'lodash/merge';
 import _noop from 'lodash/noop';
 import _omit from 'lodash/omit';
@@ -39,7 +37,6 @@ import _pick from 'lodash/pick';
 import _some from 'lodash/some';
 import _startCase from 'lodash/startCase';
 import _uniq from 'lodash/uniq';
-import _upperFirst from 'lodash/upperFirst';
 import _values from 'lodash/values';
 import { withStyles } from '@material-ui/core/styles';
 
@@ -48,9 +45,10 @@ const getResourceList = (resourcesByNs, ns) => {
   return resourcesByNs[ns] || _uniq(_flatten(_values(resourcesByNs)));
 };
 
-const urlPropsQueryConfig = _mapValues(tapQueryProps, () => {
-  return { type: UrlQueryParamTypes.string };
-});
+const urlPropsQueryConfig = {};
+for (let value in tapQueryProps) {
+  urlPropsQueryConfig[value] = StringParam;
+}
 
 const styles = theme => ({
   root: {
@@ -91,11 +89,13 @@ class TapQueryForm extends React.Component {
   static propTypes = {
     classes: PropTypes.shape({}).isRequired,
     cmdName: PropTypes.string.isRequired,
+    currentQuery: tapQueryPropType.isRequired,
     enableAdvancedForm: PropTypes.bool,
     handleTapClear: PropTypes.func,
     handleTapStart: PropTypes.func.isRequired,
     handleTapStop: PropTypes.func.isRequired,
     query: tapQueryPropType.isRequired,
+    setQuery: PropTypes.func.isRequired,
     tapIsClosing: PropTypes.bool,
     tapRequestInProgress: PropTypes.bool.isRequired,
     updateQuery: PropTypes.func.isRequired
@@ -135,7 +135,7 @@ class TapQueryForm extends React.Component {
   constructor(props) {
     super(props);
 
-    let query = _merge({}, props.query, _pick(this.props, Object.keys(tapQueryProps)));
+    const query = _merge({}, props.currentQuery, _pick(props.query, Object.keys(tapQueryProps)));
     props.updateQuery(query);
 
     let advancedFormExpanded = _some(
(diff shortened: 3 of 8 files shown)


scottcarol:
Hi @cynthia-sg, thanks for making this PR. Before the review process starts, it would be helpful to summarize the changes you made and why. What major changes are happening in the dashboard code due to the upgraded React? How should this affect work moving forward? Because we have many people working on the dashboard, including external contributors, it's helpful to explain the thought process behind making these changes. Thank you!

scottcarol:
It may be helpful to check out `CONTRIBUTING.md` for our documentation about submitting a pull request. With large PRs like this, sharing as much explanation as possible makes it easier to review, and helps educate the entire contributor community about what is going on. 👍

web/app/js/components/TopRoutes.jsx
```diff
@@ -19,29 +19,27 @@ import TopRoutesModule from './TopRoutesModule.jsx';
...
+import _pickBy from 'lodash/pickBy';
 import _uniq from 'lodash/uniq';
-import _upperFirst from 'lodash/upperFirst';
 import { groupResourcesByNs } from './util/MetricUtils.jsx';
 import { tapResourceTypes } from './util/TapUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 import { withStyles } from '@material-ui/core/styles';
 
-const topRoutesQueryProps = {
+let topRoutesQueryProps = {
```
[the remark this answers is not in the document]
cynthia-sg: This was a mistake.

web/app/js/components/TopRoutes.jsx
```diff
@@ -333,11 +307,11 @@ class TopRoutes extends React.Component {
               query={cliQueryToDisplay}
               resource={toResourceName(query, "resource_type", "resource_name")} />
             }
-          { !this.state.requestInProgress || !this._isMounted ? null : <TopRoutesModule query={this.state.query} /> }
+          { !this.state.requestInProgress || !this._isMounted ? null : <TopRoutesModule query={_pickBy(query)} /> }
```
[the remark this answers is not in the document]
cynthia-sg: For cleaning undefined values from `query` object. If you do not provide the predicate function to `pickBy`, the method will automatically check checked whether a value is truthy or not.

web/app/js/components/TapQueryForm.jsx
```diff
@@ -112,11 +100,12 @@ class TapQueryForm extends React.Component {
       let resourcesByNs = props.resourcesByNs;
       let authoritiesByNs = props.authoritiesByNs;
       let namespaces = Object.keys(resourcesByNs).sort();
-      let resourceNames  = getResourceList(resourcesByNs, state.query.namespace);
-      let toResourceNames = getResourceList(resourcesByNs, state.query.toNamespace);
-      let authorities = getResourceList(authoritiesByNs, state.query.namespace);
+      let resourceNames  = getResourceList(resourcesByNs, props.query.namespace);
```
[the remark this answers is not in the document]
cynthia-sg: I saw that we were maintaining `query` values into props and state and, having in mind that we need to update the parent state, I thought that avoiding unnecessary renders to update both across the component was a good improvement :)

Most of the changes on this file are regarding to avoid duplicity in state and props `query`.

web/app/js/components/Navigation.jsx
```diff
@@ -189,7 +189,7 @@ class NavigationBase extends React.Component {
     window.addEventListener("resize", this.updateWindowDimensions);
   }
 
-  componentWillUpdate() {
```
cynthia-sg: This method is deprecated, they recommend: `Typically, this method can be replaced by componentDidUpdate(). If you were reading from the DOM in this method (e.g. to save a scroll position), you can move that logic to getSnapshotBeforeUpdate().`

web/app/js/components/TapQueryForm.jsx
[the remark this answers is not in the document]
scottcarol: Hi @cynthia-sg, thanks for the explanation. I agree that it's a good change to avoid unnecessary renders, great 👁! 

Because these changes are not directly related to updating React, I would prefer to make those changes in a separate PR. That way, this PR will be solely focused on updating React and the fewest lines of code possible are changed. We'll be able to review it and merge it in more quickly, and there is less chance of introducing errors.

After it's been merged, if you could create a new issue for other improvements to the component, explaining what the changes were and the benefits, that would be great 👍

web/app/js/components/TapQueryForm.jsx
```diff
@@ -91,11 +89,13 @@ class TapQueryForm extends React.Component {
   static propTypes = {
     classes: PropTypes.shape({}).isRequired,
     cmdName: PropTypes.string.isRequired,
+    currentQuery: tapQueryPropType.isRequired,
```
cynthia-sg: This `prop` name has been changed due to `withQueryParams ` HOC injects the props `query` and `setQuery` into the component based on the config.

web/app/js/components/TapQueryForm.jsx
```diff
@@ -183,14 +181,14 @@ class TapQueryForm extends React.Component {
...
   }
 
   // Each time state.query is updated, this method calls the equivalent
-  // onChange method to reflect the update in url query params. These onChange
-  // methods are automatically added to props by react-url-query.
-  handleUrlUpdate = (name, formVal) => {
-    this.props[`onChange${_upperFirst(name)}`](formVal);
+  // method to reflect the update in url query params.
+  handleUrlUpdate = query => {
+    this.props.setQuery({ ...query });
```
cynthia-sg: `setQuery` by default only does partial updates (with setting replaceIn as the default). So we don't need now to call it every time that one `query` value is updated.

web/app/js/components/TopRoutes.jsx
```diff
@@ -62,24 +61,23 @@ class TopRoutes extends React.Component {
       PrefixedLink: PropTypes.func.isRequired,
     }).isRequired,
     classes: PropTypes.shape({}).isRequired,
-    query: topRoutesQueryPropType,
+    query: topRoutesQueryPropType.isRequired,
+    setQuery: PropTypes.func.isRequired,
   }
-  static defaultProps = {
```
cynthia-sg: The injected `query` prop is replacing the defaultProps to an object with undefined values (if query string value is not defined of course)

web/app/js/components/TopRoutes.jsx
```diff
@@ -62,24 +61,23 @@ class TopRoutes extends React.Component {
...
-    query: {
+
+  constructor(props) {
+    super(props);
+    this.api = this.props.api;
+
+    const query = _merge({
```
cynthia-sg: We need to send by default all values to empty strings to avoid issues with mandatory prop `value` for `SelectInput` component

web/app/js/components/TopRoutes.jsx
```diff
@@ -160,12 +158,9 @@ class TopRoutes extends React.Component {
   }
 
   // Each time state.query is updated, this method calls the equivalent
-  // onChange method to reflect the update in url query params. These onChange
-  // methods are automatically added to props by react-url-query.
+  // method to reflect the update in url query params.
```
cynthia-sg: Please, check comment above.

web/app/js/components/TopRoutes.jsx
```diff
@@ -39,9 +37,10 @@ const topRoutesQueryProps = {
 };
 const topRoutesQueryPropType = PropTypes.shape(topRoutesQueryProps);
 
-const urlPropsQueryConfig = _mapValues(topRoutesQueryProps, () => {
-  return { type: UrlQueryParamTypes.string };
-});
+let topRoutesQueryConfig = {};
```
scottcarol: Here you are defining the variable with `let`, but line 48 of `TapQueryForm` uses the same logic with `const`. Even though object keys are not protected by `const` and therefore can be changed/added/updated, I think it's better to define with `let` in both cases, for consistency and to be clear that the variable values will change.

web/app/js/components/TapQueryForm.jsx
scottcarol: Thanks for the explanation!

web/app/js/components/TapQueryForm.jsx
```diff
@@ -48,9 +45,10 @@ const getResourceList = (resourcesByNs, ns) => {
   return resourcesByNs[ns] || _uniq(_flatten(_values(resourcesByNs)));
 };
 
-const urlPropsQueryConfig = _mapValues(tapQueryProps, () => {
-  return { type: UrlQueryParamTypes.string };
-});
+const urlPropsQueryConfig = {};
```
scottcarol: See other note

web/app/js/components/TapQueryForm.jsx
```diff
@@ -183,14 +181,14 @@ class TapQueryForm extends React.Component {
 
       this.setState(state);
       this.props.updateQuery(state.query);
+      this.handleUrlUpdate(state.query);
```
scottcarol: For my own understanding, can you explain why you moved `this.handleUrlUpdate()` out of the `if (!_isNil(scopeResource)) {` block that starts on line 172? The previous code only ran `handleUrlUpdate()` if the `scopeResource` is not nil. This branch is running it in all cases.

cynthia-sg: Due to `setQuery` accepts more than one value and replace them in the current `query` prop, we don't need to call `handleUrlUpdate()` every time that some value changes. For that, we are going to call it only once with all values. Although, we only need to pass `formVal` and `scopeResource` (if not nil) values to `handleUrlUpdate()`. If you prefer this, I can change it.

scottcarol: Thanks for the explanation -- I think passing `formVal` and `scopeResource` to `handleUrlUpdate()` instead of the entire `state.query` makes it easier for others to figure out what's going on in that function, so I agree with making that change 👍

web/app/js/components/TapQueryForm.jsx
```diff
@@ -183,14 +181,14 @@ class TapQueryForm extends React.Component {
...
     };
   }
 
   // Each time state.query is updated, this method calls the equivalent
-  // onChange method to reflect the update in url query params. These onChange
-  // methods are automatically added to props by react-url-query.
-  handleUrlUpdate = (name, formVal) => {
-    this.props[`onChange${_upperFirst(name)}`](formVal);
+  // method to reflect the update in url query params.
```
dadjeibaah: I think we can reword this to be a little more descriptive something like:
```diff
diff --git a/web/app/js/components/TapQueryForm.jsx b/web/app/js/components/TapQueryForm.jsx
index 592533ba..47e69504 100644
--- a/web/app/js/components/TapQueryForm.jsx
+++ b/web/app/js/components/TapQueryForm.jsx
@@ -185,8 +185,9 @@ class TapQueryForm extends React.Component {
     };
   }
 
-  // Each time state.query is updated, this method calls the equivalent
-  // method to reflect the update in url query params.
+  // Each time state.query is updated, this method calls setQuery provided
+  // by useQueryParams HOC to partially update url query params that have
+  // changed
   handleUrlUpdate = query => {
     this.props.setQuery({ ...query });
   }


```
Or something similar.

dadjeibaah: Note that we have to include the same comment in `TapRoutes.jsx`

web/app/js/components/Tap.jsx
```diff
@@ -17,7 +17,7 @@ import { groupResourcesByNs } from './util/MetricUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
 
 const urlPropsQueryConfig = {
-  autostart: { type: UrlQueryParamTypes.string }
+  autostart: StringParam,
```
dadjeibaah: Extraneous comma after `StringParam`.

cynthia-sg: I have seen `trailing commas` in the codebase a few times (some examples below) and I personally like them because I think they make version-control diffs cleaner. However, I haven’t seen a clear convention yet as many other times it isn’t used, so I’m happy to remove it if you prefer.  

https://github.com/linkerd/linkerd2/blob/825ad51bca68c4dd518d1b92deb91e19065c63ca/web/app/js/components/Navigation.jsx#L163-L181

https://github.com/linkerd/linkerd2/blob/5d7662fd90f6352113eb1ebcef8aa7dd2ffdaf54/web/app/js/components/ErrorBanner.jsx#L15-L34

https://github.com/linkerd/linkerd2/blob/cce6183c4ab7dbe2859819f4aaa03d15a66520d4/web/app/js/components/util/CliQueryUtils.js#L4-L8

https://github.com/linkerd/linkerd2/blob/d4f3f210ce7bed885ddf5e9513c11a2a20f31d0d/web/app/js/components/util/MetricUtils.jsx#L144-L155

scottcarol: Thanks @cynthia-sg, we don't have a clear convention, as you pointed out. In this case, the original line that you replaced didn't have a trailing comma, so our preference is generally to keep it consistent to make review easier, but it's definitely a nit. 😄 

We should address this in #3667 -- if you see other inconsistencies in the code that could benefit from a formatter, please add a comment to that issue - thanks!

PR #3740, opened 2019-11-19
Fix issue with EdgesTable Grid wrapper

### LINKED ISSUE #3601: In mobile, font size when looking at a pod details is smaller
In mobile view (both portrait and landscape) the font size when in the pod details screen is smaller, both for the sidebar and the main section.

I reproduced this with Edge-19.10.3 and in #3547, using Chrome's mobile simulator

View in any section but the pod details:
image

View when looking at a pod details:
image

cynthia-sg:
Remove `Grid` wrapper around `EdgesTable` component in resource detail page.

  Fixes #3601

image

### DIFF
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -380,15 +380,11 @@ export class ResourceDetailBase extends React.Component {
           isTcpTable={true}
           metrics={this.state.podMetrics} />
 
-        <Grid container direction="column" justify="center">
-          <Grid item>
-            <EdgesTable
-              api={this.api}
-              namespace={this.state.resource.namespace}
-              type={this.state.resource.type}
-              edges={edges} />
-          </Grid>
-        </Grid>
+        <EdgesTable
+          api={this.api}
+          namespace={this.state.resource.namespace}
+          type={this.state.resource.type}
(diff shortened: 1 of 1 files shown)


web/app/js/components/ResourceDetail.jsx
```diff
@@ -380,15 +380,11 @@ export class ResourceDetailBase extends React.Component {
           isTcpTable={true}
           metrics={this.state.podMetrics} />
 
-        <Grid container direction="column" justify="center">
-          <Grid item>
```
cynthia-sg: Another solution would be to add prop `xs={12}` to Grid item, but I think that removing these extra `Grids` is better.

scottcarol:
Great fix! To test this PR, I used the Chrome DevTools "Toggle Device Toolbar" button:
Screen Shot 2019-11-25 at 3 15 58 PM

to toggle mobile view. I looked at the detail page for a Pod in `master` and see that the `Edges` table is wider than the screen, unlike other tables, which is also affecting the sidebar font size:

Screen Shot 2019-11-25 at 3 20 29 PM

then I looked at the detail page in this branch and saw that the `Edges` table is formatted properly, and the sidebar font size stays consistent.

Screen Shot 2019-11-25 at 3 17 57 PM

PR #3742, opened 2019-11-20
edge-19.11.3

zaharidichev:
## edge-19.11.3

* CLI
  * Added a check that ensures using `--namespace` and `--all-namespaces`
    results in an error as they are mutually exclusive
* Web UI
  * Fixed a bug causing the `Run Linkerd Check` button to be incorrectly
    positioned after being clicked
* Internal
  * Fixed an issue causing some control plane components to continue
    using old certificates 
  * Fixed incomplete Swagger definition of the tap api, causing benign
    error logging in the kube-apiserver
    

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,8 +7,8 @@
   * Fixed a bug causing the `Run Linkerd Check` button to be incorrectly
     positioned after being clicked
 * Internal
-  * Fixed an issue causing some control plane components to continue
-    using old certificates 
+  * Fixed an issue causing `tap`, `injector` and `sp-validator` to use 
+    old certificates after `helm upgrade` due to not being restarted
   * Fixed incomplete Swagger definition of the tap api, causing benign
     error logging in the kube-apiserver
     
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,17 @@
...
+
+* CLI
+  * Added a check that ensures using `--namespace` and `--all-namespaces`
+    results in an error as they are mutually exclusive
+* Web UI
+  * Fixed a bug causing the `Run Linkerd Check` button to be incorrectly
+    positioned after being clicked
+* Internal
+  * Fixed an issue causing some control plane components to continue
+    using old certificates
```
alpeb: Maybe you could be more specific here, saying that the tap, injector and sp-validator required to be be manually restarted when using Helm.

CHANGES.md
```diff
@@ -1,3 +1,17 @@
+## edge-19.11.3
+
+* CLI
+  * Added a check that ensures using `--namespace` and `--all-namespaces`
+    results in an error as they are mutually exclusive
+* Web UI
```
scottcarol: Hey @zaharidichev, you can delete the `Web UI` section, this only affected a certain browser version and isn't necessary to call out to users. Thanks 👍

Issue #3741, opened 2019-11-19
linkerd-proxy fails with certificates from cert-manager


piontec:
### What is the issue?

`linkerd-proxy` doesn't start when linkerd-edge is deployed with certs generated by `cert-manager`

### How can it be reproduced?

Please run the following set of commands:
```bash
kubectl create ns linkerd
kubectl -n linkerd apply -f my-certs.yaml # attached
kubectl -n linkerd get secret linkerd-identity-cert -o jsonpath="{.data['ca\.crt']}" | base64 -d > ca.crt
kubectl -n linkerd get secret linkerd-identity-cert -o jsonpath="{.data['tls\.crt']}" | base64 -d > issuer.crt
kubectl -n linkerd get secret linkerd-identity-cert -o jsonpath="{.data['tls\.key']}" | base64 -d > issuer.key
exp=$(date -d '+8760 hour' +"%Y-%m-%dT%H:%M:%SZ")
helm install -n linkerd2  linkerd-edge/linkerd2 --version=19.10.05 \
  --set-file Identity.TrustAnchorsPEM=ca.crt \
  --set-file Identity.Issuer.TLS.CrtPEM=issuer.crt \
  --set-file Identity.Issuer.TLS.KeyPEM=issuer.key \
  --set Identity.Issuer.CrtExpiry=$exp \
  --set InstallNamespace=false
```

### Logs, error output, etc

```bash
$ kubectl -n linkerd logs linkerd-identity-57998bfc4b-v94gm -c linkerd-proxy
time="2019-11-19T10:57:01Z" level=info msg="running version edge-19.10.5"
INFO [     0.002242s] linkerd2_proxy::app::main using destination service at ControlAddr { addr: Name(NameAddr { name: "linkerd-dst.linkerd.svc.cluster.local", port: 8086 }), identity: Some("linkerd-destination.linkerd.serviceaccount.identity.linkerd.cluster.local") }
INFO [     0.002296s] linkerd2_proxy::app::main using identity service at Name(NameAddr { name: "localhost.", port: 8080 })
INFO [     0.002302s] linkerd2_proxy::app::main routing on V4(127.0.0.1:4140)
INFO [     0.002306s] linkerd2_proxy::app::main proxying on V4(0.0.0.0:4143) to None
INFO [     0.002312s] linkerd2_proxy::app::main serving admin endpoint metrics on V4(0.0.0.0:4191)
[...]

PR #3746, opened 2019-11-21
Dynamically get certificates from cert-manager

piontec:
Currently, a manual process is required when installing with Helm to get the certificates from `cert-manager`. This is sub-optimal, as `cert-managers`'s task is to provide certificates for applications automatically.

This PR changes Helm's templates and hooks to first request certificates from `cert-manager`, then embed them in linkerd's configs without any interaction from the user. 

*Note:* requires `cert-manager` to be already installed in the cluster.

### DIFF
--- a/charts/linkerd2/templates/hook-pre-install-linkerd-config.yaml
+++ b/charts/linkerd2/templates/hook-pre-install-linkerd-config.yaml
@@ -10,6 +10,8 @@ metadata:
     {{.ControllerNamespaceLabel}}: {{.Namespace}}
   annotations:
     {{.CreatedByAnnotation}}: {{default (printf "linkerd/helm %s" .LinkerdVersion) .CliVersion}}
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-4"
 data:
   global: |
   {{- if .Configs -}}
--- a/charts/linkerd2/templates/hook-pre-install-namespace.yaml
+++ b/charts/linkerd2/templates/hook-pre-install-namespace.yaml
@@ -11,6 +11,8 @@ metadata:
   name: {{ .Namespace }}
   annotations:
     {{.ProxyInjectAnnotation}}: {{.ProxyInjectDisabled}}
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-10"
   labels:
     {{.LinkerdNamespaceLabel}}: "true"
     config.linkerd.io/admission-webhooks: disabled
--- a/charts/linkerd2/templates/hook-pre-install-update-config-with-cert-job.yaml
+++ b/charts/linkerd2/templates/hook-pre-install-update-config-with-cert-job.yaml
@@ -1,0 +1,123 @@
+{{with .Values -}}
+{{if .Identity -}}
+{{if and (.Identity.Issuer) (eq .Identity.Issuer.Scheme "linkerd.io/cert-manager") -}}
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: Role
+metadata:
+  name: global-config-creator
+  namespace: {{ .Namespace }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
+  labels:
+    app: global-config-creator
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  resourceNames:
+  - linkerd-config
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - linkerd-identity-issuer
+  verbs:
+  - "get"
+  - "list"
+  - "watch"
+- apiGroups:
+  - extensions
+  resources:
+  - podsecuritypolicies
+  resourceNames:
+  - linkerd-{{ .Namespace }}-control-plane
+  verbs:
+  - "use"
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: global-config-creator
+  namespace: {{ .Namespace }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
+  namespace: {{ .Namespace }}
+  labels:
+    app: global-config-creator
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: global-config-creator
+  namespace: {{ .Namespace }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-weight": "-4"
+    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
+  labels:
+    app: global-config-creator
+subjects:
+  - kind: ServiceAccount
+    name: global-config-creator
+    namespace: {{ .Namespace }}
+roleRef:
+  kind: Role
+  name: global-config-creator
+  namespace: {{ .Namespace }}
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: global-config-creator
+  namespace: {{ .Namespace }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
+  labels:
+    app: global-config-creator
+spec:
+  template:
+    spec:
+      serviceAccountName: global-config-creator
+      containers:
+      - name: global-config-creator
+        image: "quay.io/giantswarm/docker-kubectl:latest"
+        command:
+        - /bin/sh
+        - -e
+        - -x
+        - -c
+        - >-
+          while true; do 
+              export CERT=`kubectl -n linkerd get secret linkerd-identity-issuer -o jsonpath="{.data['ca\.crt']}" | base64 -d`;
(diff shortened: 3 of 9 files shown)


charts/linkerd2/templates/hook-pre-install-update-config-with-cert-job.yaml
```diff
@@ -0,0 +1,123 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: global-config-creator
+  namespace: {{ .Namespace }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
+  namespace: {{ .Namespace }}
```
zaharidichev: duplicated namespace

charts/linkerd2/templates/hook-pre-install-namespace.yaml
```diff
@@ -11,6 +11,8 @@ metadata:
   name: {{ .Namespace }}
   annotations:
     {{.ProxyInjectAnnotation}}: {{.ProxyInjectDisabled}}
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-10"
   labels:
     {{.LinkerdNamespaceLabel}}: "true"
     config.linkerd.io/admission-webhooks: disabled
```
zaharidichev: As you might have noticed our cli tools also rely on the helm chart. So if we shuffle files around, we need to make sure we are not breaking something on the cli. Currently when I compile the cli on your branch i get the error: 

```sh
➜  linkerd2 git:(use-cert-manager) ✗ bin/linkerd install --ignore-cluster
Error: open /linkerd2/templates/namespace.yaml: file does not exist
```

charts/linkerd2/templates/hook-pre-install-linkerd-config.yaml
```diff
@@ -10,6 +10,8 @@ metadata:
     {{.ControllerNamespaceLabel}}: {{.Namespace}}
   annotations:
     {{.CreatedByAnnotation}}: {{default (printf "linkerd/helm %s" .LinkerdVersion) .CliVersion}}
+    "helm.sh/hook": pre-install
```
zaharidichev: also, I am not sure we need these annotations if we are not using helm. You can check with 
`{{- if empty .CliVersion }}` to make sure that this is not a helm installation.

charts/linkerd2/templates/hook-pre-install-update-config-with-cert-job.yaml
```diff
@@ -0,0 +1,123 @@
...
+    app: global-config-creator
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  resourceNames:
+  - linkerd-config
+  verbs:
+  - "*"
```
zaharidichev: * is quite broad, can we be a bit more specific?

charts/partials/templates/_proxy.tpl
```diff
@@ -39,8 +39,16 @@ env:
 - name: LINKERD2_PROXY_IDENTITY_DIR
   value: /var/run/linkerd/identity/end-entity
 - name: LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS
+{{- if and (.Identity.Issuer) (eq .Identity.Issuer.Scheme "linkerd.io/cert-manager") }}
```
zaharidichev: FYI, we do not really support this scheme atm. The schemes we support are : 

- `linkerd.io/tls`
- `kubernetes.io/tls`

adleong:
Thanks, @piontec!  This is definitely cool.  However, it's a departure from our usual philosophy that cluster administration tasks should be performed explicitly by a user, rather than by an automated process.  If you look at the RBAC for the Linkerd components, you'll notice that (more or less) the proxy injector is the only component with write permissions.

https://linkerd.io/2/tasks/use_external_certs/#using-helm describes how Helm install works with cert manager.  The only manual aspect to this is that the `Identity.TrustAnchorsPEM` must be provided.  We certainly could use a helm pre-hook like you did here to fetch the trust anchor and populate it in the template, but we could also accomplish the same thing like so:

```
kubectl -n linkerd get secret linkerd-identity-issuer -o jsonpath="{.data['ca\.crt']}" | base64 -d > ca.cert
helm install \
  --name=linkerd2 \
  --set-file Identity.TrustAnchorsPEM=ca.cert \
  --set Identity.Issuer.Scheme=kubernetes.io/tls \
  --set InstallNamespace=false \
  linkerd/linkerd2
```

This requires one additional command at install time, but avoids all the complexity of running helm hook jobs.  Plus it makes this administrative task something that is directly invoked by a user who can validate that they're comfortable with the command before running it.

Let me know what you think!

piontec:
I think it's your call, guys. We'll use this approach at Giant Swarm, as we want the installation to be as easy as possible for a user and we're going to rely on `cert-manager` as a standard way to provide certs to applications. 

Personally, I think accepting this PR as one of the options for linkerd2 installation makes sense until it breaks something else. If you're going to support `cert-manager`'s secret format and stop there, it still means a user needs to provide CRs for `cert-manager` to create certs, then use manual commands to get CA from the cert and pass it to linkerd's chart. If the format you accept is `cert-manager` specific, why not allow the user to get the whole process automatically, as an option?

Also, just BTW, I have a question: how linkerd2 handles cert rotation now? If the values in secret change, will linkerd reload the certs automatically?

ihcsim:
@piontec Thanks again for your interest. There is definitely a tradeoff here. Do you folks have to support users who use other certificate management solutions, besides cert-manager? Just want to better understand how you solve this problem and the tradeoffs you have to consider.

Because of community requests like #2789, our current preference is to maintain a loose integration with cert-manager atm, by coupling ourselves to different secret schemes. In the case of cert-manager, it's the k8s `kubernetes.io/tls` scheme. The goal is to allow our users to bring their own issuer secret, regardless of the certificate management solutions. Eventhough the linked issue is closed, it's something that we still have in mind.

Generally, we recommend users bring their own trust root for clusters with serious workloads, so that they have full control over it. An out-of-the-box (managed or otherwise) self-signed trust root may not work for some users in the community, due to stricter security controls.

> Also, just BTW, I have a question: how linkerd2 handles cert rotation now? If the values in secret change, will linkerd reload the certs automatically?

The `identity` service will auto reload its issuer certificate if the underlying secret is changed.

zaharidichev:
> The identity service will auto reload its issuer certificate if the underlying secret is changed.

@ihcsim @piontec Just wanted to add the clarification that this is true since edge-19.10.5. So if you are on an earlier version (i.e. 2.6-stable), this auto reload functionality will not be available.

charts/linkerd2/templates/hook-pre-install-update-config-with-cert-job.yaml
```diff
@@ -0,0 +1,123 @@
...
+          while true; do 
+              kubectl -n linkerd get cm linkerd-config -o yaml --export > /tmp/config.yaml; 
+              if [ $? -eq 0 ]; then 
+                  break; 
+              fi; 
+              echo "waiting for config"; 
+              sleep 1; 
+          done; 
+          awk -v CERT="$CERT" 'BEGIN{gsub("\n", "\\n", CERT)}; {gsub("REPLACE_ME", CERT);print $0}' /tmp/config.yaml > /tmp/config-out.yaml;
+          kubectl -n linkerd replace configmap linkerd-config -f /tmp/config-out.yaml;
```
yarinm: It would be better to change the current ConfigMap definition to be named as "linkerd-config-base" and the lines mentioned here will create the "finalized" linkerd-config here.

That way we will be able to run this job periodically and achieve certificate rollout seamlessly.

In addition, at the moment every every helm upgrade will revert linkerd-config to the faulty value "REPLACE_ME". With my suggestion we won't have that kind of disruption.

WDYT?

grampelberg:
@piontec what do you think about having a separate chart for this? That way someone could add it as a dependency and get the cert-manager goodies without complicating the actual main chart. I'd be totally on board having that in-tree and adding docs for how to use it. My primary concern is bundling everything a little too tightly and that seems to solve my issues there.

piontec:
Yes, please see here https://github.com/linkerd/linkerd2/issues/3741

piontec:
I'm OK with whatever you like as a maintainer, but for me it might be hard to share the code then. Right now, the code is trying to be backwards-compatible, so until you use the new option, nothing changes.

adleong:
I think putting this in a separate chart makes the most sense for us.  This allows for the functionality without the tight coupling.

piontec:
OK, so let me know how to proceed. I have an upgraded version of this chart already, I'm just not sure how we want to keep both charts in sync and avoid code duplication (if we can)...

grampelberg:
To be clear, we mean creating a new directory which is a chart that contains all the hooks. That is setup in `requirements.yaml` and can be enabled/disabled via a value in the linkerd2 `values.yaml`. Why would that end up having syncing and code duplication issues?

piontec:
OK, I got you wrong - I thought you mean a completely unrelated helm chart.

zaharidichev:
@piontec
I think the benefit of being able to fetch the certs from the config map is pretty evident, especially  considering situations like the one described in #3880. Are you still keen on bringing that across the finish line?

adleong:
Closing due to inactivity.  Please let us know if you're interested in picking this back up again.

piontec:
Sorry for not responding so long, I had to switch my focus elsewhere. I think that our company will try to merge these changes at some time, we'll open a new PR to discuss than. Thanks!

shahriak:
Is this feature available for use?

grampelberg:
@shahriak it has been in edges for awhile now.

shahriak:
> @shahriak it has been in edges for awhile now.

I Installed the latest version of edge but I do not see a file named hook-pre-install-update-config-with-cert-job.yaml in the chart templates directory...

PR #3757, opened 2019-11-25
Add i18n support to dashboard

### LINKED ISSUE #3651: I18N for the dashboard

cynthia-sg:
**Problem**

The dashboard should be usable in languages other than English.

**Solution**

This PR adds support for multiple languages to the dashboard using the [react-i18next](https://github.com/i18next/react-i18next) library. A sample translation to Spanish has been added as well as a proof of concept.

Translation support is added at the component level, so all components with translatable text need to be updated. The changes are small though, we only need to wrap our component with the HOC provided by the library and call the translating function passing the text to be translated.

```
import { withTranslation } from 'react-i18next';
…

class Component extends React.Component {
  render() {
    const { t } = this.props;

    return (
      <p>
        {t("Lorem ipsum")}
      </p>
    );
  }
}

Component.propTypes = {
  t: PropTypes.func.isRequired,
};

export default withTranslation()(Component);  
```

The library allows splitting the texts in multiple namespaces. I’ve picked a few just to illustrate the concept, but we can adapt them as it fits better.

To add a new language we only have to:

* add to web/app/js/locale a new json file named with ISO Language Code
* import the new language file into web/app/js/i18n.js file and add to resources key to init i18n

**Validation**

Run `bin/web setup` to update dependencies
Run `bin/web build`
Open in a browser with Spanish language to see a translated version

This branch was created from the [react-update branch](https://github.com/linkerd/linkerd2/pull/3737) as it needs a more recent React version, so please check only the commit `Add i18n support to dashboard`. This may cause the PR to look bigger than it actually is.

Fixes #3651

Signed-off-by: Cintia Sanchez Garcia cynthiasg@icloud.com

i18n_01
i18n_06
i18n_05


scottcarol:
Hey @cynthia-sg, thanks for making this PR. First, could you explain the context behind this -- what you considered, trade-offs, etc.? Can you summarize the major changes that are happening? This is a pretty big PR (45 file changes) and it may be easier to review if we break it up into smaller PRs.

scottcarol:
@tegioz @cynthia-sg Thank you for your work on this!

There are some architectural decisions we need to make before we move forward with i18n. Once those decisions are made, this branch, and your work on it, will be very useful. I recognize that the initial issue, #3651, was very open-ended!

I'm going to close this PR for now, and ask that we take a step back and do some research.

Could you do some research, and add some notes in #3651, to answer the following questions:
* What are the major options available for internationalizing React? Beyond the `react-i18next` library, are there any good options? What are the positives and negatives about using them?
* How do the major projects support proper nouns vs. common nouns/phrases (i.e. "Linkerd" is a proper noun and needs no translation, vs. "You are not authorized to view this content" which needs a translation")?
* How does the K8s dashboard handle i18n? How are K8s-specific terms like Pod, Service, Replica Set, etc. handled?

Once we've agreed upon a good approach, we'll be able to move onto creating a JSON schema -- then finally, applying it to each component. Because our dashboard does not contain much text, this should be a fairly quick process -- but it's important to get the architectural decisions right before moving forward.

Thanks again, I'm looking forward to working with both of you on this!

cynthia-sg:
> * What are the major options available for internationalizing React? Beyond the `react-i18next` library, are there any good options? What are the positives and negatives about using them?

Almost all libraries use the same approach: they provide higher-order components, which injects internationalization functions for handling messages, dates, numbers and currencies via React’s context features. The most popular options available for internationalizing React are (the code samples are trying to keep at minimum the verbose boilerplate added to the current code):

#### react-intl

This is the most popular option. It’s part of [FormatJS](https://formatjs.io). This library provides React components and an API to format dates, numbers, and strings, including pluralization and handling translations. One of their cool features is the option to display dates relative to "now" (perhaps we can replace `date-fns`).

It was created by `Yahoo`, but from the end of 2018 Yahoo isn't providing a lot of support for this library.

**Use**: We need to wrap the component with the `injectIntl` HOC and then we can get the translations ready only using `intl` function.

```javascript
import { injectIntl } from "react-intl";

const Component = ({ intl }) => (
  <Tipography>
    { intl.formatMessage({ id: "This is a sample text." }) }
  <Tipography>
);

export default injectIntl(Component);
```

#### react-i18next

This library is based on `i18next`, which works well with small projects thanks to its simplicity (the use of `namespaces` and `nesting` is pretty useful to keep organised the data); but thinking big, `i18next` gives you the option to separate translations into multiple files and to load them on demand.

**Use**: We need to wrap our component with the `withTranslation` HOC passing our namespace (we can use a common one by default) and then we can get the translations ready only using `t` function.

```javascript
import { withTranslation } from "react-i18next";

const Component = ({ t }) => (
  <Tipography>
    { t("This is a sample text.") }
  <Tipography>
);

export default withTranslation()(Component);
```

#### linguiJS

This library is in many ways very similar to `react-intl`: both libraries use the same syntax and they also have very similar API. The library is compatible with both JSON and the PO translation file formats.

**Use**: We need to wrap our component with the `withI18n` and then we can get the translations ready only using `t` function imported from `@lingui/macro`.

```javascript
import { withI18n } from "@lingui/react"
import { t } from "@lingui/macro"

const Component = ({ i18n }) => (
  <Tipography>
    { i18n._(t`This is a sample text.`) }
  <Tipography>
);

export default withI18n()(Component);
```

#### FBT

This library is the runtime portion of the [Facebook localization framework](https://facebookincubator.github.io/fbt). It was open sourced this year. The library handles dates, numbers, strings and plurals and it has some interesting functionality for pronouns and enums.

**Use**: We need to wrap the existing text into a `fbt` component which is imported from the library.

```javascript
import fbt from 'fbt';

const Component = ({ i18n }) => (
  <Tipography>
    <fbt desc="component.message1">
      This is a sample text.
    <fbt />
  <Tipography>
);

export default Component;
```

#### react-polyglot

`Polyglot` provides a simple solution for interpolation and pluralization, based off of Airbnb’s experience adding I18n functionality to its apps.

**Use**:  We need to wrap our component with the `translate` and the decorator provides a prop `t` which is instance of Polyglot

```javascript
import { translate } from 'react-polyglot';

const Component = ({ t }) => (
  <Tipography>
    { t("This is a sample text.") }
  <Tipography>
);

export default translate()(Component);
```

#### ttag

This library is based on ES6 tagged templates.

**Use**: The `t` function is used almost as a simple GNU gettext function but with the possibility to embed some expressions inside template literals. There is also `jt` tag that can contain nested jsx elements

```javascript
import { t } from 'ttag';

const Component = () => (
  <Tipography>
    { t("This is a sample text.") }
  <Tipography>
);

export default Component;
```


> * How do the major projects support proper nouns vs. common nouns/phrases (i.e. "Linkerd" is a proper noun and needs no translation, vs. "You are not authorized to view this content" which needs a translation")?

If we want to be prepared for a complete internationalization, none of our visible content can be static (of course, Linkerd as proper noun has not to be translated) in order to allow us to create different versions of that content, which is dynamically loaded and adapted for each language/locale. Perhaps we can organise the translations in different `namespaces` or `keys` (if the library allows this feature).

It’s useful to consider the various places in the system where user-readable strings exist and then we can proceed to group them by:

- categories (for example: common, validations, errors…),
- pages, or
- components.

**Some ideas**:

- Dividing the translations by locale (`en.json`, `es.json`, etc.) into several files would be a great idea.
- Thinking about user experience, to provide an explicit way of choosing one of the locales (and persisting that choice) would be a good option. Perhaps, we need to see where would look better the `combo` to choose between the existing translations.
- Serve locale-specific content such as dates (the library `date-fns`, which we are using, supports internationalization) and numbers.
- We have to be careful with, for example, tables headers translated to languages, such as German, with traditionally longer strings.
- We need to be sure that all pages render correctly both in LTR (left-to-right) and RTL (right-to-left) languages, if we want to be prepared for languages such as Arabic.
- We have to be careful with string interpolation due to the same string can be translated differently in different languages depending on the context of the text.

> * How does the K8s dashboard handle i18n? How are K8s-specific terms like Pod, Service, Replica Set, etc. handled?

K8s dashboard uses [ngx-i18nsupport](https://www.npmjs.com/package/ngx-i18nsupport). Localization is handled by `Angular`, they add i18n attribute into tag surrounding new text in Angular template html files.

They have a huge list of translatable nouns/phrases (terms like `pods`, `replica sets`… are translatable, although sometimes some of these terms are not translated, for example, in French). You can, for example, check the French file [here] (https://github.com/kubernetes/dashboard/blob/03aad1255733e4ca34e6f32efee8b3c241197235/i18n/fr/messages.fr.xlf)

They support up to date these locales: `[ "en", "fr", "ja", "ko", "zh" ]`.

Issue #3734, opened 2019-11-15
Service Discovery Mirroring


grampelberg:
## Description

A common feature request for Linkerd is multi-cluster support. Users are running multiple Kubernetes clusters and struggling with connecting and securing the communication between them.

In preparation, an overview of how to solve these two problems without any service mesh whatsoever was written up - [Multicluster the Kubernetes Way with Linkerd](https://docs.google.com/document/d/1yDdd5nC348oNibvFAbxOwHL1dpFAEucpNuox1zcsOFg/edit). From this document, one of the primary difficulties identified was service discovery between clusters. While Kubernetes allows for easy discovery of services within a cluster, it is challenging to wire up public infrastructure such as DNS to make this work on a global scale.

## Details

- [Proposal](https://docs.google.com/document/d/1uzD90l1BAX06za_yie8VroGcoCB8F2wCzN0SUeA3ucw/edit#)
- [Engineering Notes](https://docs.google.com/document/d/1cMWjKpF4SFEj_3vx9Tt0YPTkDenGad84hy20LU0E38A/edit#heading=h.5d51c3niyeaf)
- [Project](https://github.com/orgs/linkerd/projects/41)

grampelberg:
This might do everything we need - https://github.com/coredns/kubernetai

zaharidichev:
Yes, all the cool things have been invented already... Looking at this more closely, I think there are (at least) two caveats: 
- Only works for headless services
- The pod ip of the service in the remote cluster needs to be reachable from the local cluster.

zaharidichev:
Also, correct me if I am wrong here, but this would  most likely not work with the destination service as these remote services will not be visible to it. So no profiles for these services, no retries, no timeouts, etc... If we want that though, we might have to modify the dst service to connect to these remote clusters and fetch data for them. Does that make sense?

grampelberg:
@zaharidichev that was the opinion I had reached as well after reading the code. It appears to assume a flat L3 network which isn't possible for most folks in my experience. The biggest downside is that because it isn't service based, it wouldn't be possible to do traffic splits across clusters - which is a major driver for the work on our side.

grampelberg:
This is historical.

----

### What problem are you trying to solve?

One of the two main challenges with trying to operate multiple k8s clusters is that service discovery is not shared between the clusters. Towards this end, services should be mirrored between different clusters so they can be addressed natively.

### How should the problem be solved?

[...]

PR #3762, opened 2019-11-26
[POC] Service mirror

zaharidichev:
This is really broad strokes but it shows in principle how service mirroring might actually work. This is based on #3734 that @grampelberg put together. At the moment this does not handle a lot of the details such as namespaces, fine grained permissions, etc and is adapted to work only on mac with kind a handful of hacks (but all this can be easily fixed). So to set this all up you can follow these steps: 

### Setting the remote cluster

- Create a config patch for kind names conf_patch.yaml: 
```yaml
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 30080
    hostPort: 30080
kubeadmConfigPatches:
- |
  apiVersion: kubeadm.k8s.io/v1beta2
  kind: ClusterConfiguration
  metadata:
    name: config
  kubernetesVersion: v1.15.3
  clusterName: "remote"
  controlPlaneEndpoint: "172.17.0.2:6443"
  apiServer:
    certSANs: [localhost, "127.0.0.1", "docker.for.mac.localhost"]
```
- Create the remote cluster `kind create cluster --name=remote --config=./conf_patch.yaml`
- Set kubectl to work with the remote cluster `export KUBECONFIG="$(kind get kubeconfig-path --name="remote")"`
- Initialize helm: 
```sh
kubectl -n kube-system create serviceaccount tiller
kubectl create clusterrolebinding tiller \
  --clusterrole=cluster-admin \
  --serviceaccount=kube-system:tiller
helm init --service-account tiller
```
- Install nginx ingress: 
```sh
helm install  stable/nginx-ingress --set controller.service.type=NodePort --set controller.service.nodePorts.http=30080  --set controller.service.nodePorts.https=30443
```
- Deploy ingress controller + two services: 
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: backend-one
spec:
  replicas: 1
  selector:
    matchLabels:
      app: backend-one
  template:
    metadata:
      labels:
        app: backend-one
    spec:
      containers:
      - name: backend-one
        image: buoyantio/bb:v0.0.5
        args:
        - terminus
        - "--h1-server-port=80"
        - "--response-text=backend-one"
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: backend-one-svc
spec:
  selector:
    app: backend-one
  ports:
  - name: http
    port: 80
    targetPort: 80
---
kind: Ingress
metadata:
  name: ingress-backend-one-svc
  annotations:
    # use the shared ingress-nginx
    kubernetes.io/ingress.class: "nginx"
spec:
  rules:
  - host: backend-one-svc.default.svc.cluster.local
    http:
      paths:
      - path: /
        backend:
          serviceName: backend-one-svc
          servicePort: 80
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: backend-two
spec:
  replicas: 1
  selector:
    matchLabels:
      app: backend-two
  template:
    metadata:
      labels:
        app: backend-two
    spec:
      containers:
      - name: backend-two
        image: buoyantio/bb:v0.0.5
        args:
        - terminus
        - "--h1-server-port=80"
        - "--response-text=backend-two"
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: backend-two-svc
spec:
  selector:
    app: backend-two
  ports:
  - name: http
    port: 80
    targetPort: 80
---
kind: Ingress
metadata:
  name: ingress-backend-two-svc
  annotations:
    # use the shared ingress-nginx
    kubernetes.io/ingress.class: "nginx"
spec:
  rules:
  - host: backend-two-svc.default.svc.cluster.local
    http:
      paths:
      - path: /
        backend:
          serviceName: backend-two-svc
          servicePort: 80
---
```
- Get the kubeconfig file of the remote cluster: 
`cat $(kind get kubeconfig-path --name="remote")` 
- Replace the `server: https://127.0.0.1:some-port` with `server:  https://docker.for.mac.localhost:some-port`
- Save the file as as remote_conf.yaml
- Base64 encode the contents of the file: `cat ./remote_conf.yaml | base64`
- Create a new file `remote-secret.yaml`: 
```
apiVersion: v1
data:
  config: <the base64 encoded remote server config>
kind: Secret
metadata:
  name: remote-secret
  namespace: srv-mirror
type: mirror.linkerd.io/kubeconfig
```

### Local cluster

- Checkout this branch  and build the cli + the controller image
- Create a local cluster with: `kind create cluster --name=local`
- Upload the controller image to the local cluster
- Set `kubectl` to work with the local cluster: `export KUBECONFIG="$(kind get kubeconfig-path --name="local")"`
- Install the service mirror component with: `bin/linkerd install-service-mirror | kubectl apply -f -`
- Create the secret that contains the configuration for the remote cluster with: `kubectl apply -f remote-secret.yaml`
- The service mirror component will create mirror services.
- Execute `kubectl describe service  backend-two-svc`
- You chould see something like: 
```
Name:              backend-two-svc
Namespace:         default
Labels:            linkerd.io/mirror-service=true
Annotations:       <none>
Selector:          <none>
Type:              ClusterIP
IP:                10.96.189.49
Port:              <unset>  80/TCP
TargetPort:        30080/TCP
Endpoints:         192.168.65.2:30080
Session Affinity:  None
Even
```
- Now install a container to try out hitting the remote service: 
```sh
cat <<EOF | kubectl apply -f -
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: blue
  namespace: default
spec:
  replicas: 1
  template:
    metadata:
      labels:
        name: blue
    spec:
      containers:
      - name: blue-website
        image: scrapinghub/httpbin:latest
        command:
        - sleep
        - "3600"
        resources:
          requests:
            cpu: 0.1
            memory: 200
EOF
```
- Get a shell to the container and execute: 
`curl http://backend-two-svc.default.svc.cluster.local`
- You should see something like: 
```json
{
  "requestUID": "in:http-sid:terminus-grpc:-1-h1:80-84655700",
  "payload": "backend-two"
}
```

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -117,6 +117,8 @@ func init() {
 	RootCmd.AddCommand(newCmdUninject())
 	RootCmd.AddCommand(newCmdUpgrade())
 	RootCmd.AddCommand(newCmdVersion())
+	RootCmd.AddCommand(newCmdInstallServiceMirror())
+
 }
 
 type statOptionsBase struct {
--- a/cli/cmd/service-mirror.go
+++ b/cli/cmd/service-mirror.go
@@ -1,0 +1,59 @@
+package cmd
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"text/template"
+
+	"github.com/linkerd/linkerd2/cli/install"
+	"github.com/linkerd/linkerd2/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+type installServiceMirrorOptions struct {
+	Namespace string
+	LinkerdVersion string
+}
+
+func newServiceMirrorOptions() *installServiceMirrorOptions {
+	return &installServiceMirrorOptions{
+		Namespace: "srv-mirror",
+		LinkerdVersion:  version.Version,
+	}
+}
+
+func newCmdInstallServiceMirror() *cobra.Command {
+	options := newServiceMirrorOptions()
+
+	cmd := &cobra.Command{
+		Use:   "install-service-mirror [flags]",
+		Short: "Output Kubernetes configs to install Linkerd Service Mirror",
+		Long:  "Output Kubernetes configs to install Linkerd Service Mirror",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return renderServiceMirror(os.Stdout, options)
+		},
+	}
+
+	cmd.PersistentFlags().StringVarP(&options.LinkerdVersion, "linkerd-version", "v", options.LinkerdVersion, "Version")
+	cmd.PersistentFlags().StringVarP(&options.Namespace, "namespace", "n", options.Namespace, "Namespace")
+	return cmd
+}
+
+
+func renderServiceMirror(w io.Writer, config *installServiceMirrorOptions) error {
+	template, err := template.New("linkerd-service-mirror").Parse(install.ServiceMirrorTemplate)
+	if err != nil {
+		return err
+	}
+	buf := &bytes.Buffer{}
+	err = template.Execute(buf, config)
+	if err != nil {
+		return err
+	}
+
+	w.Write(buf.Bytes())
+	w.Write([]byte("---\n"))
+
+	return nil
+}
--- a/cli/install/service-mirror-template.go
+++ b/cli/install/service-mirror-template.go
@@ -1,0 +1,96 @@
+package install
+
+
+const (
+	// CNITemplate provides the base template for the `linkerd install-cni-plugin` command.
+	ServiceMirrorTemplate = `### Namespace ###
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: {{.Namespace}}
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: linkerd-{{.Namespace}}-service-mirror
+  labels:
+    ControllerComponentLabel: service-mirror
+    ControllerNamespaceLabel: {{.Namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["pods", "endpoints", "services", "secrets", "namespaces"]
+  verbs: ["list", "get", "watch", "create"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: linkerd-{{.Namespace}}-service-mirror
+  labels:
+    ControllerComponentLabel: service-mirror
+    ControllerNamespaceLabel: {{.Namespace}}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: linkerd-{{.Namespace}}-service-mirror
+subjects:
+- kind: ServiceAccount
+  name: linkerd-service-mirror
+  namespace: {{.Namespace}}
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-service-mirror
+  namespace: {{.Namespace}}
+  labels:
+    ControllerComponentLabel: service-mirror
+    ControllerNamespaceLabel: {{.Namespace}}
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: linkerd-service-mirror
+  namespace: {{.Namespace}}
+spec:
+  type: ClusterIP
+  selector:
+    ControllerComponentLabel: service-mirror
+  ports:
+  - name: grpc
+    port: 8086
+    targetPort: 8086
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    ControllerComponentLabel: service-mirror
+  name: linkerd-service-mirror
+  namespace: {{.Namespace}}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      ControllerComponentLabel: service-mirror
+      ControllerNamespaceLabel: {{.Namespace}}
+  template:
+    metadata:
+      labels:
+        ControllerComponentLabel: service-mirror
+        ControllerNamespaceLabel: {{.Namespace}}
+    spec:
+      containers:
+      - args:
+        - service-mirror
+        image: gcr.io/linkerd-io/controller:{{.LinkerdVersion}}
+        name: service-mirror
+        ports:
+        - containerPort: 8086
+          name: grpc
+        - containerPort: 9996
+          name: admin-http
+        securityContext:
+          runAsUser: 2103
+      serviceAccountName: linkerd-service-mirror
+`
+)
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	servicemirror "github.com/linkerd/linkerd2/controller/cmd/service-mirror"
 	"os"
 
 	"github.com/linkerd/linkerd2/controller/cmd/destination"
@@ -34,6 +35,8 @@ func main() {
 		spvalidator.Main(os.Args[2:])
 	case "tap":
 		tap.Main(os.Args[2:])
+	case "service-mirror":
+		servicemirror.Main(os.Args[2:])
 	default:
 		fmt.Printf("unknown subcommand: %s", os.Args[1])
 		os.Exit(1)
--- a/controller/cmd/service-mirror/main.go
+++ b/controller/cmd/service-mirror/main.go
@@ -1,0 +1,241 @@
+package servicemirror
+
+import (
+	"flag"
+	"fmt"
+	"github.com/linkerd/linkerd2/controller/k8s"
+	"github.com/linkerd/linkerd2/pkg/flags"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"net"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+const (
+	mirrorSecretType = "mirror.linkerd.io/kubeconfig"
+	mirrorServiceLabel = "linkerd.io/mirror-service"
+	configKeyName = "config"
+)
+
+type RemoteClusterGatewayService  struct {
+	name string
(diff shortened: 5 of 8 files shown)


zaharidichev:
Now, with all that being said there are some questions that arise: 

1. If we are to go ahead with this, we need to think about a viable stress free way to have that working on a local dev environment. We cannot have the ingress expose a `LoadBalancer` service as its generally not natively supported on local environments. That is why I am using a `NodePort` with `Kind`. The addressing also needs to be figured out. What is the address of the remote cluster on a local machine. Is it the same as the address of the local cluster (if it was exposing an ingress controller as well).

2. Currently the ingress on the remote cluster is routing traffic based of of the Host header. Is that what we imagine happening (pardon my ignorance here). Because if that is the case, we need to carry some kind of information about the fully qualified svc name on the remote cluster and put that as the host header. Currently the way I solve this is by creating a mirror service with identical name in an identically named namespace. This might not always be possible due to permissions, etc. Can we make the proxy do that? What if we are running without a proxy, as I hear this is functionality we want to have separately from `linkerd`

3. What kind of tool chain do we envision supplying with all this? Do we want to boil everything down to a few simple commands such as `linkerd install svc-mirror`, etc or we want to make it a bit more of an advanced/manual task to set all this up? 

4. How dynamic is our configuration. Currently the mirror service watches for remote cluster config data and registers watchers for these clusters that monitor services. But what happens if the ingress service changes somehow, do we want to account for that and dynamically reconfigure our local view of the remote cluster. 

I have quite a few more thoughts about that but I think these are good questions to kick off the discussion.

grampelberg:
> If we are to go ahead with this, we need to think about a viable stress free way to have that working on a local dev environment. We cannot have the ingress expose a LoadBalancer service as its generally not natively supported on local environments. That is why I am using a NodePort with Kind. The addressing also needs to be figured out. What is the address of the remote cluster on a local machine. Is it the same as the address of the local cluster (if it was exposing an ingress controller as well).

I think we should assume there is a service with an external IP for now. If that happens to be a nodePort that someone updates manually with the external IP address that's okay. As we work through this, we *definitely* should write up docs that explain how to do it all with two kind clusters locally.

> Currently the ingress on the remote cluster is routing traffic based of of the Host header. Is that what we imagine happening (pardon my ignorance here). 

We'll need a header that rewrites from the service in one cluster (`foo-remote`) to the local one  (`foo`). If I was doing this completely independently from Linkerd, I'd do that with ingress definitions in the destination cluster.

>  Currently the way I solve this is by creating a mirror service with identical name in an identically named namespace.

We're going to want to append the cluster name to the service names that are synced to make sure there's no collision.

> This might not always be possible due to permissions, 

The mirror service is going to need permissions to create/delete services in every namespace that is sync'd. Ideally we can use a regex and subResource to keep the permissions small.

> What kind of tool chain do we envision supplying with all this? 

I would like to have the mirror component be part of the default install. Configuration would be through secrets or a specific CRD if we *really* have to.

>  But what happens if the ingress service changes somehow, do we want to account for that and dynamically reconfigure our local view of the remote cluster.

Yes, if the remote's external IP changes, we should update the endpoints for the local services.

adleong:
Some more questions to fuel discussion:

Which of the goals from https://docs.google.com/document/d/1yDdd5nC348oNibvFAbxOwHL1dpFAEucpNuox1zcsOFg/edit#heading=h.kyfu4gxqs7g1 does this approach solve?  Since the mirror endpoints do not have a TargetRef, I don't think that Linkerd will be able to TLS the connection or provide useful metrics.

I'm guessing that to add that, we would need to add support for a `linkerd.io/remote-identity` annotation.  When the destination controller resolves a service with this annotation, it can add the appropriate identity metadata to the response so that the proxy will TLS the connection.  It could also add further endpoint metadata that would indicate that this is cross-cluster traffic and would show up as prometheus labels.

Saving the current user's private token into a secret in the cluster sets off a lot of alarm bells for me.  We should consider if it's possible to generate a service account in one cluster with limited RBAC, and then transplant that service account's token to the other cluster for API use.

Better yet would be to avoid cross-cluster kubernetes API access altogether.

adleong:
As a counter proposal to service mirroring, what about this?

We create a single local service which represents all traffic bound for a remote cluster.  E.g. in the `arguello` cluster we create this service:

```
apiVersion: v1
kind: Service
metadata:
  name: balboa-cluster
  namespace: default
  annotations:
    linkerd.io/remote-identity: linkerd-gateway-default.linkerd.serviceaccount.identity.linkerd.cluster.local
  labels:
    linkerd.io/mirror-service: true
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    # Port that the remote gateway is listening on
    targetPort: 443
```
and endpoint
```
apiVersion: v1
kind: Endpoints
metadata:
  labels:
    linkerd.io/mirror-service: true
  name: balboa-cluster
  namespace: default
subsets:
- addresses:
  - ip: 23.10.11.93
  ports:
  - name: gateway
    port: 443
    protocol: TCP
```

We can then use traffic splits to redirect none, some, or all traffic from a service in the local cluster to the same service in the remote cluster.  E.g.

```
kind: TrafficSplit
metadata:
  name: foo-cross-cluster
  namespace: default
spec:
  service: foo
  backends:
  - service: foo
    weight: 500m
  - service: balboa-cluster
    weight: 500m
```

For traffic addressed to `foo.default.svc.cluster.local`, 50% will be sent to the local cluster foo service and the other 50% will be sent to the remote balboa cluster gateway.  Critically, this will not modify the authority header so that when the request reaches the balboa cluster gateway, the ingress controller can perform authority based ingress routing to send the request the `foo.default.svc.cluster.local` service in the balboa cluster.

Linkerd would still need to support the `linkerd.io/remote-identity` annotation in order to properly TLS this cross-cluster connection but this would obviate the need for service mirroring and would not require any cross-cluster API access.

grampelberg:
> Saving the current user's private token into a secret in the cluster sets off a lot of alarm bells for me. We should consider if it's possible to generate a service account in one cluster with limited RBAC, and then transplant that service account's token to the other cluster for API use.

Not the user's private token. The service account's token (which is already in a secret in the other cluster).

---

Couple questions about your proposal @adleong:

- How does it work if there isn't a local service?
- What happens when the destination cluster doesn't have the service?
- How does someone introspect what they can and can't route?
- How does this work for TCP/UDP?
- Can this work without the mesh at all? What are the fallback options?

olix0r:
@adleong Does your proposal require that there is only a single gateway per cluster? That's a non-starter, in my opinion. Policy is going to have to be tied between clients and gateways; and so I expect some environments will prefer a gateway-per-service-in-the-same-namespace approach. I don't think we want to recommend that off the bat, but we need make sure we can instrument fine-grained access control.

adleong:
Thanks, @grampelberg and @olix0r.  The problem space is starting to come into focus for me.  Let me see if I'm understanding correctly.

I think instructions in this POC save the user's kubeconfig as a secret, but I now see the description in https://github.com/linkerd/linkerd2/issues/3734 that talks about generating a kubeconfig for a service account in the remote cluster.  That makes a lot more sense.

Upon closer inspection, I see that https://github.com/linkerd/linkerd2/issues/3734 leaves the gateway configuration unspecified, and I think I understand why now.  There are a few different ways to do this: you could have a single gateway that routes based on authority, which would only work for HTTP traffic, or you could have individual gateways per service (or you could mix and match).  Either way, all the local cluster has to do is discovery the gateway for each service and use that as the mirror service's endpoint.  Should the remote services specify their gateway as an annotation?

adleong:
So all in all, here's what a user would need to do to set this up:

* Create a service account in the remote cluster, extract its kubeconfig, and save it in the local cluster
* Expose remote cluster services with externally accessible services (type loadBalancer or nodePort)
  * If using an ingress controller, configure the ingress controller to route hosts like `{svc}-balboa.{ns}.svc.cluster.local` to `{svc}.{ns}.svc.cluster.local`
* Deploy the service mirrorer in the local cluster

With all that in place, clients in the local cluster can send to `{svc}-balboa.{ns}.cluster.local` to send to the service in the remote cluster.  This will work with or without a service mesh in place.  If Linkerd *is* in place then users can use traffic splits to shift between local and remote (mirror) services.

If support for the `linkerd.io/remote-identity` annotation is added to Linkerd, then cross-cluster communication can be TLS'd as well.

If support for the `linkerd.io/mirror-service: true` annotation is added to Linkerd, then cross-cluster communication can be identified more easily in the metrics with appropriate prometheus labels.

grampelberg:
> Expose remote cluster services with externally accessible services (type loadBalancer or nodePort)

This should require creating an externally accessible service that goes to a gateway (either linkerd managed or an ingress controller). The remote services should *not* be required to have a load balancer for each one.

adleong:
More rapid fire questions that we'll need to figure out.

* What exactly are the steps to generate a Role, RoleBinding, and ServiceAccount on the remote cluster?  Should there be a `linkerd` subcommand to generate these manifests?  Should there be flags to configure which services it has permissions to list?
* What exactly are the steps to generate a kubeconfig secret from that service account?  Is this a totally manual process or is there tooling or automation that we can provide? 
* What exactly are the steps to configure the remote ingress to correctly route incoming traffic from the source cluster?  Is this configuration a manual step or is there tooling or automation we can provide?
  * What are the ingress routing rules?  How does this work for HTTP vs TCP?
* How does the service mirrorer discover the gateway endpoint for a remote service?  Should it be an annotation on that remote service?  Or should the ingress service somehow be annotated to list the services that it acts as a gateway for?
* How does the service mirrorer discovery the identity of the gateway for a remote service?

zaharidichev:
@adleong thanks for joining the discussion

> What exactly are the steps to generate a Role, RoleBinding, and ServiceAccount on the remote cluster? Should there be a linkerd subcommand to generate these manifests? Should there be flags to configure which services it has permissions to list?

I beleive all this should be packaged into linkerd subcommands as otherwise it invoves quite a bit of manual work. 

> What exactly are the steps to generate a kubeconfig secret from that service account? Is this a totally manual process or is there tooling or automation that we can provide?

There are a sequence of commans that you can use to make it less manual but it still is a non trivial operation. I think we can add automation here. [This is one version ](https://gist.github.com/ericchiang/d2a838ddad3f44436ae001a342e1001e)of the process packaged into a script, to give you a better idea of how some people do it

> How does the service mirrorer discover the gateway endpoint for a remote service? Should it be an annotation on that remote service? Or should the ingress service somehow be annotated to list the services that it acts as a gateway for?
I think it should be an annotation on that remote service rather than the other way around. This makes me wonder could there be a service that is accessible by two gateways? If so, how is that specified and where...

zaharidichev:
closing in favor of #4028

Issue #3618, opened 2019-10-21
Consolidate dashboard API requests further to speed up dashboard


scottcarol:
#3467 simplified and sped up the dashboard by reducing API requests, adding a namespace selection button and removing the sidebar's list of resources and their green/red status icons. It also expanded our use of React's `Context` API. The user's selected namespace is now shared to all components that are consuming context. We can update that shared context from within the app when the user changes the namespace.

We may be able to reduce API requests even more by leaning into the `Context` API. We should investigate each component that is making API calls to see if the calls can be further consolidated.

scottcarol:
Hey @tegioz, thanks for taking on this issue! 

Before a PR, I'd like to have a discussion, that can take place in this issue, about different options we can consider to achieve a faster dashboard, on the React/JavaScript side. 

Can you evaluate our existing architecture and API requests, and propose some options (without a PR)? This will enable us to have a conversation before making any major changes to dashboard architecture. 

Whatever we decide, due to the upcoming dashboard redesign, it may be best to wait on making these changes, to prevent doing work that will be undone with a new UI. In the meantime, I'm looking forward to your thoughts!

tegioz:
## A bit of context first

In #3258 it was pointed out that the Linkerd dashboard imposed a lot of CPU stress on Prometheus. In my following comments in this issue I'll propose some possible improvements we could apply to reduce the load on Prometheus. This is an ongoing work, I'll keep posting any new findings.

At the moment, I'm using a local minikube cluster for my tests with only the `emojivoto` app running on it. The initial goal is to start thinking about improvements that may have a positive impact at any scale.

None of the improvements identified so far are silver bullets. They may have a positive impact in some use cases, but won't apply to others. However, most of them are easy to implement, so if they help to make the life of some users easier, they may be worth considering.

tegioz:
## Improvement 1

### Avoid fetching data not needed from the dashboard

[...]

PR #3767, opened 2019-11-27
Avoid fetching data not needed in dashboard (I)

tegioz:
#### Problem

In the ResourceDetail view, when fetching data for upstream and downstream resources, we use `all` as the resource type. Up in the chain, when getting metrics from Prometheus, using `all` leads to querying Prometheus for each resource kind individually both for inbound and outbound directions. However, that view isn't really interested in all those stats, as it filters them out later on when processing the data received. When looking at a pod details, we actually only look at pods metrics. When looking at other resources, we are not interested on the pods metrics, for example.

#### Solution

With this change we only fetch the data we need for upstream and downstream resources. When the resource is a pod, only metrics for upstream and downstream pods is fetched. For any other resource, we still rely on the special resource `all`, but we ask the API to exclude the resource types we are not interested in (authority, pod and service). To make this possible, a new filter (`ExcludeFromAll`) has been added to the API.

More details about this issue and the impact of the solution [here](https://github.com/linkerd/linkerd2/issues/3618#issuecomment-556141006).

Related to #3618


scottcarol:
@tegioz, Thanks for jumping into this! I have some big-picture questions and notes about the approach, I haven’t reviewed the code line by line.

The change to `stat_summary.go` seems confusing. If `req.Selector.Resource.Type == k8s.All`, we range through every resource type and check if `ExcludeFromAll` is true for that resource -- if not, we add the resource type to the slice `resourcesToQuery`.

This means `req.Selector.Resource.Type == k8s.All` and `ExcludeFromAll` for a resource could be true *simultaneously* --- which directly contradicts the statement `req.Selector.Resource.Type == k8s.All`. 

It looks like we can make improvements to `pod` details without adding `ExcludeFromAll` to the code - is this correct? What do you think of making a separate PR with the pod-specific logic? 

And is there a way of switching from an `ExcludeFromAll` approach to an approach that would make a query without making `k8s.All` the `Resource.Type`? Let me know if you want to talk through in more detail.

tegioz:
> @tegioz, Thanks for jumping into this! I have some big-picture questions and notes about the approach, I haven’t reviewed the code line by line.
> 
> The change to `stat_summary.go` seems confusing. If `req.Selector.Resource.Type == k8s.All`, we range through every resource type and check if `ExcludeFromAll` is true for that resource -- if not, we add the resource type to the slice `resourcesToQuery`.
> 
> This means `req.Selector.Resource.Type == k8s.All` and `ExcludeFromAll` for a resource could be true _simultaneously_ --- which directly contradicts the statement `req.Selector.Resource.Type == k8s.All`.

Yes, they can be true simultaneously, that was exactly the goal. `ExcludeFromAll` is a mechanism to allow excluding specific resources only when requesting `all`, hence the name. So that piece of code reads as.. *"if all resource types are requested, check if some of them should be excluded and do it if needed"*. It's building up a slice the same way one was used before the change, except this time some types can be excluded on demand, which is what we need for our use case. If you can think of a name that expresses the intent clearer, I'd be happy to change it (I spent quite a bit time discarding some other options and I ended up liking this one the most though ;).

> It looks like we can make improvements to `pod` details without adding `ExcludeFromAll` to the code - is this correct? What do you think of making a separate PR with the pod-specific logic?

Yes, the problem could be half-fixed only addressing the case when a resource type is pod. Given that changes are minimal and that both PRs would be touching the same logic creating conflicts, I'm not sure if it's really worth it, honestly. But if you don't see it clear after my explanation in the first paragraph and you'd be more comfortable splitting, let me know and I'll create another one, it's not a problem.

> And is there a way of switching from an `ExcludeFromAll` approach to an approach that would make a query without making `k8s.All` the `Resource.Type`? Let me know if you want to talk through in more detail.

I considered a different approach as well. Instead of allowing to exclude some types, I modified the api to support requesting a list of types, which would make `all` unnecessary. That would work fine too. However that change was a bit more intrusive as that api is used in a few places, the PR was much bigger, which I was trying to avoid so that this simple change could be merged easily, and could be error prone as every time a new resource type is added we might have to add it to several places.

controller/api/util/api_utils.go
```diff
@@ -55,11 +55,12 @@ var (
...
-	Namespace     string
-	ResourceType  string
-	ResourceName  string
-	AllNamespaces bool
+	TimeWindow     string
+	Namespace      string
+	ResourceType   string
+	ResourceName   string
+	AllNamespaces  bool
+	ExcludeFromAll []string
```
zaharidichev: TIOLI: Maybe it is possible to validate the contents of this slice. Namely can we return an error if any of the to be excluded resources is not actually a valid resource name in order to avoid future headscratchers?

tegioz: Good idea, will do it, thanks 👍

scottcarol:
I appreciate your work on this @tegioz and the follow-up response to my first comment! 🙏 My overall concern is that this new flow is not going to be intuitive to someone reading through `/api/public/stat_summary.go` who does not have a lot of context. It's already a complex file used for both CLI and dashboard requests. 

If I'm seeing `Resource.Type == k8s.All` in the code, I expect to query *all* resource types. I'm not expecting that some types will be excluded. Your solution does cut down on fetching data 👏  but if `k8s.All` does not really mean `All` it begins to feel confusing to me. I don't think it's a problem of the suggested name `ExcludeFromAll`, I think it's more that we rely on "all" meaning "all" - if that makes any sense - and changing our use of that would be a much larger architectural change. 

I'm definitely open to what other folks think, and your solution totally works!

tegioz:
Hi @scottcarol 

Thanks for your feedback. Cool, let's see what other folks think. I'm happy to make any needed change :)

However, please let's keep in mind this is helping to solve a real problem with the Prometheus load and improving the overall users' experience in the dashboard. How it's done it's mostly an implementation detail, and this one particularly is very small and not intrusive. My point is that if we don't come up with a way of addressing the problem that makes everyone happy, I think we should go ahead with this one to make users happy, and improve this implementation later on.

tegioz:
As discussed above in my first reply to @scottcarol, I'm happy going that way too @adleong. That was my original attempt until I realised the change was a bit more aggressive and it may be harder to get it merged.

> > And is there a way of switching from an `ExcludeFromAll` approach to an approach that would make a query without making `k8s.All` the `Resource.Type`? Let me know if you want to talk through in more detail.
> 
> I considered a different approach as well. Instead of allowing to exclude some types, I modified the api to support requesting a list of types, which would make `all` unnecessary. That would work fine too. However that change was a bit more intrusive as that api is used in a few places, the PR was much bigger, which I was trying to avoid so that this simple change could be merged easily, and could be error prone as every time a new resource type is added we might have to add it to several places.

scottcarol:
Your related PR, #3768, has just been merged -- thank you for your work on that @tegioz! 🙏 

Because you're headed off, and there's a good architectural discussion going on here, I'm going to close this PR and open a follow-up issue that summarizes the dashboard speed improvements that you've made to date, linking to this PR as a possible next step.

tegioz:
Thanks @scottcarol, no worries 👍

PR #3768, opened 2019-11-27
Avoid fetching data not needed in dashboard (II)

tegioz:
#### Problem

Currently when viewing a pod details in the dashboard, we list all pods in the namespace and we fetch metrics for all of them every 2s. We do this to get all the pods belonging to the requested resource and display them. But in the case of a pod, that doesn't apply. So in that view, that data is not really needed. We need the requested pod metrics though, but we already fetch them in a separate query. This can lead to expensive Prometheus queries, specially in big clusters.

#### Solution

With this change we avoid listing all the pods in the namespace and fetching all their metrics when displaying a pod detail view.

More details about this issue and the impact of the solution [here](https://github.com/linkerd/linkerd2/issues/3618#issuecomment-557586629).

Related to #3618

### DIFF
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -154,7 +154,7 @@ export class ResourceDetailBase extends React.Component {
 
     let apiRequests =
       [
-        // inbound stats for this resource
+      // inbound stats for this resource
         this.api.fetchMetrics(
           `${this.api.urlsForResource(resource.type, resource.namespace, true)}&resource_name=${resource.name}`
         ),
@@ -168,6 +168,7 @@ export class ResourceDetailBase extends React.Component {
         )
       ];
 
+    // Fetch pods in a resource and their metrics (except when resource type is pod)
     if (resource.type !== "pod") {
       // list of all pods in this namespace (hack since we can't currently query for all pods in a resource)
       apiRequests.push(this.api.fetchPods(resource.namespace));
--- a/web/app/js/components/ResourceDetail.jsx
+++ b/web/app/js/components/ResourceDetail.jsx
@@ -154,15 +154,9 @@ export class ResourceDetailBase extends React.Component {
 
     let apiRequests =
       [
-      // inbound stats for this resource
+        // inbound stats for this resource
         this.api.fetchMetrics(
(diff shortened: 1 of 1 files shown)


web/app/js/components/ResourceDetail.jsx
```diff
@@ -154,16 +154,10 @@ export class ResourceDetailBase extends React.Component {
 
     let apiRequests =
       [
-      // inbound stats for this resource
+        // inbound stats for this resource
```
scottcarol: Mind removing the added whitespace?

web/app/js/components/ResourceDetail.jsx
```diff
@@ -174,6 +168,13 @@ export class ResourceDetailBase extends React.Component {
         )
       ];
 
+    if (resource.type !== "pod") {
```
scottcarol: It would be helpful to add a comment above this line explaining why you are differentiating resource types.

web/app/js/components/ResourceDetail.jsx
```diff
@@ -188,10 +189,18 @@ export class ResourceDetailBase extends React.Component {
     this.api.setCurrentRequests(apiRequests);
 
     Promise.all(this.api.getCurrentPromises())
-      .then(results => {
-        const [resourceRsp, podListRsp, podMetricsRsp, upstreamRsp, downstreamRsp, ...rsp] = [...results];
+      .then(apiResponses => {
```
scottcarol: Was there a reason you changed `results` to `apiResponses` here and below?

web/app/js/components/ResourceDetail.jsx
tegioz: Sure, no problem, done. I thought introducing a whitespace was ok when done for consistency reasons. Please note below how that comment is the only one differently indented.

https://github.com/linkerd/linkerd2/blob/c078b4ff8d6f1e40a4eabcb520b2116fa4a52454/web/app/js/components/ResourceDetail.jsx#L156-L175

web/app/js/components/ResourceDetail.jsx
tegioz: Yes, for consistency reasons with `apiRequests` and to make clearer its meaning as results was a bit too generic. But I'm happy to change it back if you prefer `results`.

PR #3769, opened 2019-11-27
Cache StatSummary responses in dashboard web server

tegioz:
#### Problem

If the dashboard is opened multiple times (by the same or different users), depending on the view opened, it’s possible that the same API stats requests are processed multiple times within the same second. Some of these requests trigger expensive queries in Prometheus, specially on big clusters.

#### Solution

This changes adds a simple caching layer to the stats API handler using the raw query as the key. Json marshalled responses are cached for up to 1.5 seconds, and they are cleaned up automatically in 5 minutes.

More details about this issue and the impact of the solution [here](https://github.com/linkerd/linkerd2/issues/3618#issuecomment-556195957).

Related to #3618

### DIFF
--- a/web/srv/api_handlers.go
+++ b/web/srv/api_handlers.go
@@ -21,14 +21,24 @@ import (
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/protohttp"
 	"github.com/linkerd/linkerd2/pkg/tap"
+	"github.com/patrickmn/go-cache"
 	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 )
 
-// Control Frame payload size can be no bigger than 125 bytes. 2 bytes are
-// reserved for the status code when formatting the message.
-const maxControlFrameMsgSize = 123
+const (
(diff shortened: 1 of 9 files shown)


web/srv/api_handlers.go
```diff
@@ -118,6 +124,20 @@ func (h *handler) handleAPIServices(w http.ResponseWriter, req *http.Request, p
 }
 
 func (h *handler) handleAPIStat(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
+	// Initialize stat cache if needed
+	if h.statCache == nil {
+		// Cached items expire in 1.5s, expired ones are cleaned up every 5m
+		h.statCache = cache.New(1500*time.Millisecond, 5*time.Minute)
```
zaharidichev: Maybe these nums can be put in a constant or made configurable somehow so they dont stay magic ones

tegioz:
I think imposing a maximum cache size is something more common in LRU caches, where you have a criteria to know with object should be discarded to leave room for a new one. In this case, when reached certain size, what would happen if a new item wanted to be added and none of the existing ones were expired?

For our use case, memory usage should be minimum and this should have no noticeable impact. Cache is cleaned up automatically every 5 minutes though, so I can't foresee any problem here. We can be a bit more aggressive and clean it up every minute anyway.

PR #3771, opened 2019-11-28
Add issuer file flags to upgrade command

### LINKED ISSUE #3677: Update the 'upgrade' command to support the identity issuer flags
In order to allow users to rotate the existing trust root and issuer's key/cert pair, the `upgrade` command need to be updated to support the following install-only flags:

```
--identity-issuer-certificate-file string   A path to a PEM-encoded file containing the Linkerd Identity issuer certificate (generated by default)
--identity-issuer-key-file string           A path to a PEM-encoded file containing the Linkerd Identity issuer private key (generated by default)
--identity-trust-anchors-file string        A path to a PEM-encoded file containing Linkerd Identity trust anchors (generated by default)
--identity-trust-domain string              Configures the name suffix used for identities. (default "cluster.local")
```

If these flags aren't provided, `upgrade` should retain the current behaviour of preserving the `linkerd-config` config map and `linkerd-identity-issuer` secret from the control plane. 

If present, then `upgrade` should only replace the trust root in the identity context of the existing `linkerd-config` config map. All other existing data within the config map should be retained. The `linkerd-identity-issuer` secret can be replaced entirely.

Relevant #3452.

zaharidichev:
This pr adds functionality to the upgrade command that allows it to overwrite the contents of `linkerd-identity-issuer` secret with certs provided by files on disk. Care is taken to ensure that these certs have the same trust roots that are currently in the config.

Fixes #3677 


cli/cmd/upgrade.go
```diff
@@ -364,30 +385,53 @@ func fetchIdentityValues(k kubernetes.Interface, idctx *pb.IdentityContext) (*ch
...
 		},
 	}, nil
 }
 
-func fetchIssuer(k kubernetes.Interface, trustPEM string, scheme string) (string, string, time.Time, error) {
+type issuerData struct {
+	key string
+	crt string
+	exp time.Time
+}
```
alpeb: Nice touch :+1:

cli/cmd/upgrade.go
```diff
@@ -236,6 +237,19 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
...
+			return nil, nil, errors.New("a certificate file must be specified if a private key is provided")
+		}
+		if options.identityOptions.keyPEMFile == "" {
+			return nil, nil, errors.New("a private key file must be specified if a certificate is provided")
+		}
+		if err := checkFilesExist([]string{options.identityOptions.crtPEMFile, options.identityOptions.keyPEMFile}); err != nil {
+			return nil, nil, err
+		}
+	}
+
```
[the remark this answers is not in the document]
zaharidichev: @alpeb, yes I certainly agree! I am doing some work to consolidate all that in #3696 so I think I will rebase this one on top of it once its merged and consolidate things further.

cli/cmd/upgrade_test.go
```diff
@@ -452,14 +459,158 @@ data:
...
+			},
 		},
 	}
 
 	for i, tc := range testCases {
 		tc := tc // pin
 		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
 			options, err := testUpgradeOptions()
+
+			if tc.processOptions != nil {
```
ihcsim: I think this should come after the `if err != nil` at line 614. No point of performing this if `testUpgradeOptions()` returns an error.

zaharidichev: Yes, good question. Missed the fact that upgrade and install are in the same package ...

cli/cmd/install.go
```diff
@@ -1047,13 +1047,14 @@ func loadExternalIssuerData() (*externalIssuerData, error) {
 	return &externalIssuerData{string(anchors), string(crt), string(key)}, nil
 }
 
-func (idopts *installIdentityOptions) verifyCred(creds *tls.Cred, trustAnchors string) error {
+// VerifyCreds verifies the validity of the provided certs
+func VerifyCreds(creds *tls.Cred, trustAnchors, dns string) error {
```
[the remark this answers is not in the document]
zaharidichev: Yes, good question. Missed the fact that upgrade and install are in the same package ...

alpeb:
@zaharidichev can you please explain why updating the trust anchor has to be performed as a separate step than updating the issuer's cert and key?

zaharidichev:
@alpeb I think this sums it up. Let me know if you see any holes in my logic: https://github.com/linkerd/website/pull/595#discussion_r351744060

alpeb:
@zaharidichev By looking at the code I realized that was a requirement, so I wondered why. Your comment explains that well.
OTOH, it seems if one wants to change just the trust root, it has to be parented to the old trust root?
```
$ step certificate create identity.linkerd.cluster.local ca.crt ca.key --profile root-ca --no-password --insecure
$ bin/go-run cli upgrade --identity-trust-anchors-file ca.crt 
× Failed to build upgrade configuration: invalid issuer credentials: x509: certificate signed by unknown authority (possibly because of "x509: ECDSA verification failure" while trying to verify candidate authority certificate "identity.linkerd.cluster.local")
```

zaharidichev:
@alpeb Yes that is my assumption. Essentially the constraint is: _You can put whatever you want in there as long as it does not break Linkerd_ Would you say that is sensible?

ihcsim:
@alpeb Both obersvations are correct: upgrading the trust root and issuer cert is separate steps, and the trust root provided to `upgrade` must be a bundle containing the old trust. Otherwise, once the proxies in the data plane received its new TLS certs signed by the new issuer cert, they won't be able to talk to each other, because they are still on the old root. I will create a issue on `website` to update the troubleshooting guide, per the CLI errors you saw.

PR #3775, opened 2019-12-01
Enable cert rotation test to work with dynamic namespaces

### LINKED ISSUE #3758: external_issuer_integration_tests hardcode namespace name
The external_issuer_integration_tests make use of static certificates that contain a fixed identity which contains the namespace of the Linkerd control plane.  Since this namespace is fixed in the certificate, it is impossible to run these tests in any other namespace.  This means that these tests must be disabled in the cloud integration tests:

https://github.com/linkerd/linkerd2/pull/3730

To make this test runnable in any environment, we could dynamically generate the certificates at the start of the test, using the configured test namespace.

zaharidichev:
This PR adds support for dynamic cert generation when running the cert rotation intergration tests. This allows to avoid baking in the namespace in the certificate CN, thereby allowing us to run these tests on the clouds. 

Fixes #3758 


testutil/test_helper.go
```diff
@@ -152,9 +153,30 @@ func (h *TestHelper) UpgradeFromVersion() string {
 
 // GetClusterDomain returns the custom cluster domain that needs to be used during linkerd installation
 func (h *TestHelper) GetClusterDomain() string {
+	if h.clusterDomain == "" {
+		return "cluster.local"
```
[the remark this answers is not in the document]
zaharidichev: Good call!

test/externalissuer/external_issuer_test.go
```diff
@@ -89,16 +91,21 @@ func verifyAppWorksBeforeCertRotation(t *testing.T) {
 }
 
 func verifyRotateExternalCerts(t *testing.T) {
-	// change issuer secret
-	secretResource, err := testutil.ReadFile("testdata/issuer_secret_2.yaml")
+	// We rotate the certificates here by simply grabbing
+	// the key and cert values from the temporary secret we have
+	// created
+	secretWithUpdatedData, err := TestHelper.KubernetesHelper.GetSecret(TestHelper.GetLinkerdNamespace(), k8s.IdentityIssuerSecretName+"-new")
+
```
ihcsim: Nit: Mind removing the extra line break, to be consistent with everywhere else? Most Go code I have seen groups the `if err != nil` immediately after the `err` declaration.

testutil/kubernetes_helper.go
```diff
@@ -54,6 +54,15 @@ func (h *KubernetesHelper) CheckIfNamespaceExists(namespace string) error {
 	return err
 }
 
+// GetSecret retrieves a Kubernetes Secret
+func (h *KubernetesHelper) GetSecret(namespace, name string) (*corev1.Secret, error) {
+	secret, err := h.clientset.CoreV1().Secrets(namespace).Get(name, metav1.GetOptions{})
```
ihcsim: TIOLI: This is the same as `return h.clientset.CoreV1().Secrets(namespace).Get(name, metav1.GetOptions{})`.

PR #3784, opened 2019-12-04
No need for `processYAML()` in `install`

alpeb:
Since `install` uses helm to do its proxy injection, there's no need to
call `processYAML`. This also fixes an issue discovered in #3687 where
we started supporting injection of cronjobs, and even though `linkerd`'s
namespace is flagged to skip automatic injection it was being injected.

This replaces #3774 as it's a much simpler approach.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -754,10 +754,8 @@ func render(w io.Writer, values *charts.Values, configs *pb.All) error {
 		return err
 	}
 
-	return processYAML(&buf, w, ioutil.Discard, resourceTransformerInject{
-		injectProxy: true,
-		configs:     configs,
-	})
+	_, err = w.Write(buf.Bytes())
+	return err
 }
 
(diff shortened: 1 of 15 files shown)


cli/cmd/install.go
```diff
@@ -754,10 +753,8 @@ func render(w io.Writer, values *charts.Values, configs *pb.All) error {
 		return err
 	}
 
-	return processYAML(&buf, w, ioutil.Discard, resourceTransformerInject{
-		injectProxy: true,
-		configs:     configs,
-	})
+	_, err = w.Write(buf.Bytes())
```
ihcsim: TIOLI: `processYAML()` ended the YAML blob with:
```
out.Write([]byte("---\n")) 
```
I think that's why all the unit tests have the extra:
```
- ---
```

alpeb: Right, `processYAML()` uses that to separate resources, even when there's only one. I think it's better to leave it out.

PR #3786, opened 2019-12-04
Added unit test for injecting debug sidecar into CP deployment

alpeb:
I realized this was missing when testing #3774 (superseded by #3784).


cli/cmd/testdata/inject_tap_deployment.input.yml
```diff
@@ -0,0 +1,246 @@
+apiVersion: extensions/v1beta1
```
ihcsim: Nit: we should use `apps/v1` here and in the output YAML.

PR #3798, opened 2019-12-09
Inject preStop hook into the proxy sidecar container to stop it last

### LINKED ISSUE #3747: Configure linkerd-proxy to ignore SIGTERM on a per-workload basis
## Feature Request

Disclaimer: 
- There was a similar question: https://github.com/linkerd/linkerd2/issues/1850 almost a year ago. I'm starting a new one but actually the problem is the same.
- There's an open issue in Kubernetes https://github.com/kubernetes/kubernetes/issues/65502 but it will take too long for the possible implementation to be shipped into production environments.

When a pod have [a preStop hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) it stops slower than its Linkerd sidecar. This eliminates the idea of the preStop hook since the pod cannot communicate with other services when Linkerd exited.

### Use case example

1. AWS ALB has a target group that sends traffic to a pod **A**.
2. `kubectl rollout restart` is called and Kubernetes starts a new pod **B** and tries to terminate the previous pod **A.**
3. [ALB continues sending some traffic](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/issues/1064) even after the instance status changed from Healthy to Draining. There's not much we can do. Also there's a special option "Deregistration delay" which was designed to keep existing connection for a while to make sure all requests were processed.
4. An application container inside pod **A** has a `preStop` hook and waits for `N` second before exit to serve the existing requests and those new requests that still come from ALB for unknown reason.
5. Linkerd exits immediately and drops all connections.

Also the problem makes impossible to use "[Slow start duration](https://aws.amazon.com/about-aws/whats-new/2018/05/application-load-balancer-announces-slow-start-support/)" setting in the ALB.

### Logs example

```
# a new pod started and Kubernetes marks the old pod as Terminating
# Linkerd immediately says in its logs: 
linkerd-proxy INFO [   919.021686s] linkerd2_proxy::signal received SIGTERM, starting shutdown

# After that the application cannot connect to monitoring and to other services anymore:
E, [2019-11-21T14:09:25.634678 #2719] ERROR -- ddtrace: [ddtrace] (/usr/local/bundle/gems/ddtrace-0.22.0/lib/ddtrace/transport.rb:215:in `log_error_once') Failed to open TCP connection to 10.5.5.8:8126 (Connection refused - connect(2) for "10.5.5.8" port 8126)

# And at the and after N seconds the app exits as it was asked. But it does not make sense.
# ALB returns many 5xx errors to clients
```

If there was a way how to inject a custom preStop hook into Linkerd container with a simple `sleep` command inside (the container probably does not have the command?..) it might solve the problem. In this case I would ask Linkerd to wait for enough long period and Kubernetes would eventually ask it to quit according [to the documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods:

> 1. one of the Pod’s containers has defined a preStop hook, it is invoked inside of the container. If the preStop hook is still running after the grace period expires, step 2 is then invoked with a small (2 second) extended grace period.
> 2. The container is sent the TERM signal. Note that not all containers in the Pod will receive the TERM signal at the same time and may each require a preStop hook if the order in which they shut down matters.

I cannot find how to actually solve the problem. As a workaround, some tunning around liveness/readiness probes and health check thresholds can be performed. And the "Deregistration delay" can be set to something close to zero to fail faster. But 5xx errors will still be there.

It can be a flag for `linkerd inject --manual`, for example `--wait-before-exit 30` that will prevent linkerd-proxy from exiting for N seconds.

KIVagant:
This commit adds support for a Graceful Shutdown technique that is used
by some Kubernetes administrators while the more perspective 
configuration is being discussed in
https://github.com/kubernetes/kubernetes/issues/65502

The problem is that RollingUpdate strategy does not guarantee that all 
traffic will be sent to a new pod _before_ the previous pod is removed.
Kubernetes inside is an event-driven system and when a pod is being 
terminating, several processes can receive the event simultaneously.
And if an Ingress Controller gets the event too late or processes it
slower than Kubernetes removes the pod from its Service, users requests
will continue flowing into the black whole. 

According [to the documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods)

> 1. If one of the Pod’s containers has defined a `preStop` hook, 
> it is invoked inside of the container. If the `preStop` hook is still 
> running after the grace period expires, step 2 is then invoked with
> a small (2 second) extended grace period.
>
> 2. The container is sent the `TERM` signal. Note that not all 
> containers in the Pod will receive the `TERM` signal at the same time
> and may each require a preStop hook if the order in which 
> they shut down matters.

This commit adds support for the `preStop` hook that can be configured 
in three forms:

1. As command line argument `--wait-before-exit-seconds` for
  `linkerd inject` command.

2. As `linkerd2` Helm chart value `Proxy.WaitBeforeExitSeconds`.

2. As `config.alpha.linkerd.io/wait-before-exit-seconds` annotation.

If configured, it will add the following preHook to the proxy container
definition:

```yaml
lifecycle:
  preStop:
    exec:
      command:
        - /bin/bash
        - -c
        - sleep {{.Values.Proxy.WaitBeforeExitSeconds}}
```

To achieve max benefit from the option, the main container should have
its own `preStop` hook with the `sleep` command inside which has
a smaller period than is set for the proxy sidecar. And none of them
must be bigger than `terminationGracePeriodSeconds` configured for the
entire pod.

An example of a rendered Kubernetes resource where
`.Values.Proxy.WaitBeforeExitSeconds` is equal to `40`:

```yaml
       # application container
        lifecycle:
          preStop:
            exec:
              command:
                - /bin/bash
                - -c
                - sleep 20

        # linkerd-proxy container 
        lifecycle:
          preStop:
            exec:
              command:
                - /bin/bash
                - -c
                - sleep 40
    terminationGracePeriodSeconds: 160 # for entire pod
```

Fixes #3747


KIVagant:
@grampelberg ptal

controller/gen/config/config.pb.go
```diff
@@ -182,6 +182,7 @@ type Proxy struct {
 	DisableExternalProfiles bool                  `protobuf:"varint,12,opt,name=disable_external_profiles,json=disableExternalProfiles,proto3" json:"disable_external_profiles,omitempty"`
 	ProxyVersion            string                `protobuf:"bytes,13,opt,name=proxy_version,json=proxyVersion,proto3" json:"proxy_version,omitempty"`
 	ProxyInitImageVersion   string                `protobuf:"bytes,14,opt,name=proxy_init_image_version,json=proxyInitImageVersion,proto3" json:"proxy_init_image_version,omitempty"`
+	WaitBeforeExitSeconds   int32                 `protobuf:"varint,10,opt,name=proxy_uid,json=waitBeforeExitSeconds,proto3" json:"wait_before_exit_seconds,omitempty"`
```
KIVagant: Whatta hell am I doing here?.. ¯\(°_o)/¯

controller/gen/config/config.pb.go
```diff
@@ -182,6 +182,7 @@ type Proxy struct {
 	DisableExternalProfiles bool                  `protobuf:"varint,12,opt,name=disable_external_profiles,json=disableExternalProfiles,proto3" json:"disable_external_profiles,omitempty"`
 	ProxyVersion            string                `protobuf:"bytes,13,opt,name=proxy_version,json=proxyVersion,proto3" json:"proxy_version,omitempty"`
 	ProxyInitImageVersion   string                `protobuf:"bytes,14,opt,name=proxy_init_image_version,json=proxyInitImageVersion,proto3" json:"proxy_init_image_version,omitempty"`
+	WaitBeforeExitSeconds   int32                 `protobuf:"varint,15,opt,name=wait_before_exit_seconds,json=waitBeforeExitSeconds,proto3" json:"wait_before_exit_seconds,omitempty"`
```
KIVagant: Could someone help me to find a documentation about the line format?

KIVagant:
How to test the PR:


1. Build and deploy all images from the branch.

2. Download https://run.linkerd.io/emojivoto.yml

3. Inject with the `--manual` flag and `--wait-before-exit-seconds 111`. Check the result.

4. Modify manually in an editor the original file and the `config.alpha.linkerd.io/wait-before-exit-seconds` annotation with random values and leave at least one Deployment without the annotation. Inject with the `--manual` flag but without the `--wait-before-exit-seconds` flag. Check the result.

```
  template:
    metadata:
      annotations:
        config.alpha.linkerd.io/wait-before-exit-seconds: "64"
```

5. Install Linkerd from the Helm chart with some `WaitBeforeExitSeconds` value. Check if injected pods have the configured sleeping period.

KIVagant:
Rebased from master

KIVagant:
Refreshed new `upgrade_overwrite_issuer.golden` and `upgrade_overwrite_trust_anchors.golden` to include default value of `waitBeforeExitSeconds`

controller/gen/config/config.pb.go
alpeb: You need to change only the file `/proto/config/config.proto`
```diff
diff --git a/proto/config/config.proto b/proto/config/config.proto
index 64b05547..3285e574 100644
--- a/proto/config/config.proto
+++ b/proto/config/config.proto
@@ -50,6 +50,8 @@ message Proxy {
   string proxy_version = 13;
 
   string proxy_init_image_version = 14;
+
+  int32 wait_before_exit_seconds = 15;
 }
 
 message Image {
```
and then run `bin/protoc-go.sh` which will regenerate the `pb.go` files automatically

cli/cmd/inject.go
```diff
@@ -100,6 +100,11 @@ sub-folders, or coming from stdin.`,
 		&manualOption, "manual", manualOption,
 		"Include the proxy sidecar container spec in the YAML output (the auto-injector won't pick it up, so config annotations aren't supported) (default false)",
 	)
+	flags.UintVar(
+		&options.waitBeforeExitSeconds, "wait-before-exit-seconds", options.waitBeforeExitSeconds,
+		"The periods during which the proxy sidecar container must stay alive while its pod is terminating. "+
+			"Must be smaller than terminationGracePeriodSeconds for the pod (default 0)",
+	)
```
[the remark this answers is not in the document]
KIVagant: TL;DR notes for myself:
- move the logic under `flagSet()` in `root.go`
- rename the option to `proxy-wait-before-exit`
- update the desc: "The periods" > "The period"
- update the desc: remove redundant "container" word 

What about the "-seconds" suffix? Worth keeping it everywhere or should I remove it and rename all the variables?

controller/gen/config/config.pb.go
KIVagant: Oh.. got it, thank you.

alpeb:
@KIVagant After talking further over this with @grampelberg we came to the conclusion we only want this to be exposed as an option to `linkerd inject` and its corresponding annotation, and not to `linkerd install` nor as config to be stored in `linkerd-config`. So some of my comments above are no longer relevant (I'll flag them as so). And it means we don't need changes to the protobuf `ProxyStruct`. These are the changes that your last push would require:
```diff
diff --git a/controller/gen/config/config.pb.go b/controller/gen/config/config.pb.go
index 4c7f7997..42472eaa 100644
--- a/controller/gen/config/config.pb.go
+++ b/controller/gen/config/config.pb.go
@@ -182,7 +182,6 @@ type Proxy struct {
        DisableExternalProfiles bool                  `protobuf:"varint,12,opt,name=disable_external_profiles,json=disableExternalProfiles,proto3" json:"disable_external_profiles,omitempty"`
        ProxyVersion            string                `protobuf:"bytes,13,opt,name=proxy_version,json=proxyVersion,proto3" json:"proxy_version,omitempty"`
        ProxyInitImageVersion   string                `protobuf:"bytes,14,opt,name=proxy_init_image_version,json=proxyInitImageVersion,proto3" json:"proxy_init_image_version,omitempty"`
-       WaitBeforeExitSeconds   int32                 `protobuf:"varint,15,opt,name=wait_before_exit_seconds,json=waitBeforeExitSeconds,proto3" json:"wait_before_exit_seconds,omitempty"`
        XXX_NoUnkeyedLiteral    struct{}              `json:"-"`
        XXX_unrecognized        []byte                `json:"-"`
        XXX_sizecache           int32                 `json:"-"`
@@ -269,13 +268,6 @@ func (m *Proxy) GetOutboundPort() *Port {
        return nil
 }
 
-func (m *Proxy) GetWaitBeforeExitSeconds() int32 {
-       if m != nil {
-               return m.WaitBeforeExitSeconds
-       }
-       return 0
-}
-
 func (m *Proxy) GetResource() *ResourceRequirements {
        if m != nil {
                return m.Resource
diff --git a/pkg/inject/inject.go b/pkg/inject/inject.go
index 9252532e..237015a9 100644
--- a/pkg/inject/inject.go
+++ b/pkg/inject/inject.go
@@ -749,7 +749,7 @@ func (conf *ResourceConfig) proxyWaitBeforeExitSeconds() int32 {
                }
        }
 
-       return conf.configs.GetProxy().GetWaitBeforeExitSeconds()
+       return 0
 }
 
 func (conf *ResourceConfig) proxyResourceRequirements() *l5dcharts.Resources {
diff --git a/pkg/inject/inject_test.go b/pkg/inject/inject_test.go
index 18ffb2a1..fbd7ae00 100644
--- a/pkg/inject/inject_test.go
+++ b/pkg/inject/inject_test.go
@@ -66,7 +66,6 @@ func TestConfigAccessors(t *testing.T) {
                LogLevel:                &config.LogLevel{Level: "info,linkerd2_proxy=debug"},
                DisableExternalProfiles: false,
                ProxyVersion:            proxyVersion,
-               WaitBeforeExitSeconds:   0,
        }
 
        globalConfig := &config.Global{
```

cli/cmd/inject.go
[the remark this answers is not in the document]
alpeb: The move to `root.go` is no longer needed according to my comment below

controller/gen/config/config.pb.go
alpeb: This is no longer needed, as per my comment below.

KIVagant:
@alpeb I cannot remember what leaded me to adding the protobuf configuration but without that, I guess, annotation won't work and also I had some problems with the default value evaluation when made a mistake in the line, so I really feel this is important to have the config line.

cli/cmd/inject.go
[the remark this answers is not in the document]
alpeb: > What about the "-seconds" suffix? Worth keeping it everywhere or should I remove it and rename all the variables?

I would remove it from the `linkerd inject` command, and accept a `duration` type, to keep it in line with other options like `--identity-clock-skew-allowance` and `--identity-issuance-lifetime`. And then when the annotation gets interpreted in `/pkg/inject/inject.go` you could translate it to seconds.
As for the Helm value `WaitBeforeExitSeconds` I would keep it as is, following the convention of manifest properties like `initialDelaySeconds`, and also because at the Helm level I believe there's no function to translate a `duration` type into an int.
Let me know if this makes sense to you.

alpeb:
My last patch above should remove all the references to the config variable. If you still have problems you can also shoot me a message in slack.

KIVagant:
Today's update:
- removed `seconds` suffix everywhere except Helm chart values where `time.Duration` is not supported;
- updated `gcr.io/linkerd-io/go-deps` (I am not sure how to do this correctly, please double check)
- added `k8s.ProxyWaitBeforeExitAnnotation` to `cli/cmd/doc.go`
- `--wait-before-exit` flag for `linkerd inject` now accepts `time.Duration` only, `uint` is not supported
- `config.alpha.linkerd.io/proxy-wait-before-exit` annotation by default expects `time.Duration` type but if a plain integer is set it will try to interpret the value as "time in seconds". Added an additional positive test case to cover this scenario.
- removed `protobuf` configuration.

Tested with the following set of annotation examples:

- `config.alpha.linkerd.io/proxy-wait-before-exit: 2m` results in `/bin/bash -c sleep 120`
- `config.alpha.linkerd.io/proxy-wait-before-exit: "64"` results in `/bin/bash -c sleep 64`
- `config.alpha.linkerd.io/proxy-wait-before-exit: "111"` results in `/bin/bash -c sleep 111`
- `config.alpha.linkerd.io/proxy-wait-before-exit: "-1"` results in missing `preStop` hook
- missed annotation results in missing `preStop` hook
- `linkerd inject --wait-before-exit 120s` results in `/bin/bash -c sleep 120`
- `linkerd inject --wait-before-exit 2m20s` results in `/bin/bash -c sleep 140`
- `linkerd inject --wait-before-exit 0` results in missing `preStop` hook same as when the 
parameter is omitted
- `linkerd inject --wait-before-exit -1` results in an error

KIVagant:
The next step would be to find a way how to avoid using bash in the hook and integrate the sleep into the proxy, but this is controversial.

KIVagant:
Rebased from master

cli/cmd/inject.go
```diff
@@ -439,6 +446,9 @@ func (options *proxyConfigOptions) overrideConfigs(configs *cfg.All, overrideAnn
 	if options.traceCollectorSvcAccount != "" {
 		overrideAnnotations[k8s.ProxyTraceCollectorSvcAccountAnnotation] = options.traceCollectorSvcAccount
 	}
+	if options.waitBeforeExit != 0 {
+		overrideAnnotations[k8s.ProxyWaitBeforeExitAnnotation] = time.DurationToSecondsString(options.waitBeforeExit) + "s"
```
alpeb: Instead of introducing a new dependency on `github.com/docker/docker/api/types/time`, you could do `options.waitBeforeExit.Seconds()`

ihcsim: s/`Seconds()`/`String()`

cli/cmd/doc.go
```diff
@@ -207,5 +207,9 @@ func generateAnnotationsDocs() []annotationDoc {
 			Name:        k8s.ProxyTraceCollectorSvcAccountAnnotation,
 			Description: "The trace collector's service account name. E.g., `tracing-service-account`. If not provided, it will be defaulted to `default`.",
 		},
+		{
+			Name:        k8s.ProxyWaitBeforeExitAnnotation,
+			Description: "The proxy sidecar will stay alive for at least the given period before receiving SIGTERM signal from Kubernetes but no longer than pod's `terminationGracePeriodSeconds`. Accepts values in two formats: time.Duration or as an unsigned integer (number of seconds). If not provided, it will be defaulted to `0`.",
```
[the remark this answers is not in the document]
KIVagant: There's one reason. When I was testing the `time.Duration` I found that without "s" at the end the value will be ignored which is confusing for end-user. When one writes annotation `linkerd/wait: 300` he probably will intuitively expect that the value will be converted to `sleep 300` but it won't.

charts/linkerd2/values.yaml
```diff
@@ -87,6 +87,16 @@ Proxy:
     CollectorSvcAddr: ""
     CollectorSvcAccount: default
   UID: 2102
+  # If not empty, the proxy sidecar will stay alive for at least the given period
+  # before receiving SIGTERM signal from Kubernetes
+  # but no longer than terminationGracePeriodSeconds value for the entire pod.
+  # Depending on the goals, the application container (to which the proxy sidecar is attached)
```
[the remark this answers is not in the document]
KIVagant: No objections, sounds good to me, thank you

cli/cmd/doc.go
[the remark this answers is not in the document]
ihcsim: I see what you mean; the existing code surrounding `conf.proxyWaitBeforeExit()` doesn't offer the opportunity to propagate any errors back to the caller. How about if we support only `int32` and convert everything into seconds (which I think is consistent with the current variable name ,`WaitBeforeExitSeconds`)? Technically, `sleep` only works with `s`, `m`, `h` and `d`, which is a subset of what `time.Duration` supports. If the description says `time.Duration` is supported, we will need to account for suffices like `ns`, `us` and `ms`.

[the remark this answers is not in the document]
KIVagant: I am totally okay with using `int32` only but now it conflicts with CLI arguments. 
Anyway, someone will get into the trap with one or another version. ¯\_(ツ)_/¯ I don't believe that we can transparently pass `time.Duration` format to bash `sleep` just as a string without accurate converting. So please choose what should be left, it is not a principal question for me.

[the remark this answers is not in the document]
adleong: Ideally we'd have both the CLI flag and the annotation in the same format.  

Is it possible for the injector to do validation of the annotation and fail to create the pod if it has this annotation in the wrong format?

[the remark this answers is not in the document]
KIVagant: Ok, so I'm returning the code to the initial state with the `-seconds` suffix which makes everything obvious for end user.

pkg/inject/inject.go
```diff
@@ -757,6 +759,22 @@ func (conf *ResourceConfig) tapDisabled() bool {
 	return false
 }
 
+func (conf *ResourceConfig) proxyWaitBeforeExit() time.Duration {
+	if override := conf.getOverride(k8s.ProxyWaitBeforeExitAnnotation); override != "" {
+		waitBeforeExit, err := time.ParseDuration(override)
+
+		// Try to parse plain integer as value in seconds if it is not formatted as Duration
+		if err != nil {
+			waitInSeconds, _ := strconv.ParseUint(override, 10, 32)
```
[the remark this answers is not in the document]
KIVagant: I'll revert the code state to the variant with `-seconds` suffix everywhere which accepts uint32. Even without logging the error, if a sleep command hasn't been enabled, anyone should quickly find the problem just from the annotation and arguments naming.

KIVagant:
Today's update:
- returned `uint` (previously it was `int32`, then `Duration`) as the format for both, the cli argument and the annotation.
- added a negative test case to test the resulting value is zero if the annotation is in a wrong format  (one scenario).
- added a log warning message if the annotation is in a wrong format.

Dockerfile-proxy
```diff
@@ -9,7 +9,7 @@ RUN (proxy=$(bin/fetch-proxy $(cat proxy-version)) && \
     mv "$proxy" linkerd2-proxy)
 
 ## compile proxy-identity agent
-FROM gcr.io/linkerd-io/go-deps:2273074d as golang
+FROM gcr.io/linkerd-io/go-deps:ab3733c1 as golang
```
KIVagant: I'm not sure if I'm doing this right.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/cyphar/filepath-securejoin v0.2.2 // indirect
 	github.com/deislabs/smi-sdk-go v0.0.0-20190610232231-f281e2121a16
 	github.com/dgrijalva/jwt-go v3.1.0+incompatible // indirect
+	github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0
```
alpeb: @KIVagant could you please remove this and run `bin/update-go-deps-shas` on more time?

KIVagant: I'm trying right now but I barely understand what I'm doing

alpeb: You introduced a dependency on the `docker` lib in your earlier commits, which resulted in this line being added. All the `Dockerfile` files contain a checksum of the `go.mod` file, so every time it changes the `Dockerfile` files need to be updated, which is done through `bin/update-go-deps-shas`. That dependency is no longer needed, so it's just a matter of manually removing that line, and then running `bin/update-go-deps-shas`.

ihcsim:
Opppss... I hit `approve` by mistake yesterday; but it looks like it's very close :wink:.

cli/cmd/doc.go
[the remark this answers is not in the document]
ihcsim: Thanks. Can you update the new option's description in this file?

cli/cmd/root.go
```diff
@@ -199,6 +199,7 @@ type proxyConfigOptions struct {
 	enableExternalProfiles   bool
 	traceCollector           string
 	traceCollectorSvcAccount string
+	waitBeforeExitSeconds    uint
```
ihcsim: Why not just make this `uint64`? Then you won't have to worry about type casting in `uintToString(v uint)` and `proxyWaitBeforeExitSeconds()`.

pkg/inject/inject.go
```diff
@@ -757,6 +759,20 @@ func (conf *ResourceConfig) tapDisabled() bool {
 	return false
 }
 
+func (conf *ResourceConfig) proxyWaitBeforeExitSeconds() (uint, error) {
+	if override := conf.getOverride(k8s.ProxyWaitBeforeExitAnnotation); override != "" {
+		waitBeforeExitSeconds, err := strconv.ParseUint(override, 10, 32)
```
ihcsim: If we use `uint64`, then we can simplify this to
```
return strcconv.ParseUnit(override, 10, 64)
```

cli/cmd/root.go
KIVagant: First, explicit uint64 is quite big. Second, uint based on system architecture which changes the focus of where the decision is made. Third, doing anything here I'm trying to follow some, let's say, traditions that were made around. I'm not a Golang expert, so I learn from the project code. And around the variable there are many `uint`s, array of `uint`s and the only one `int64`. So I picked `uint`. KISS

pkg/inject/inject.go
KIVagant: But who will ever need the sleeping period longer than 136,19251950152 years?

go.mod
KIVagant: Looks like now I've got it, thank you! :) Removed the line and the sha is the same now as in the master.

cli/cmd/root.go
ihcsim: Thanks for the additional context. Since this is a CLI command (i.e. short-lived process), I am not too concerned about a 32 vs 64 bit local variable. Your `strconv.ParseUint(override, 10, 32)` is already returning a 64-bit uint anyway.

> KISS

I think removing all the type casting you have will help us to get closer to this goal.

pkg/inject/inject.go
ihcsim: I am hoping we can explore ways to simplify the code, without needing all the type casting. The limit of the type isn't a concern here.

cli/cmd/root.go
KIVagant: Oookay :)

pkg/inject/inject.go
KIVagant: Done!

KIVagant:
Today's update:
- `uint` type replaced with `uint64`

cli/cmd/doc.go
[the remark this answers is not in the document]
alpeb: @KIVagant gentle ping about the description update :wink:

KIVagant:
Fixed the annotation description and renamed the variable from `ProxyWaitBeforeExitAnnotation` to `ProxyWaitBeforeExitSecondsAnnotation`

pkg/inject/inject.go
```diff
@@ -757,6 +759,20 @@ func (conf *ResourceConfig) tapDisabled() bool {
...
 
+func (conf *ResourceConfig) proxyWaitBeforeExitSeconds() (uint64, error) {
+	if override := conf.getOverride(k8s.ProxyWaitBeforeExitSecondsAnnotation); override != "" {
+		waitBeforeExitSeconds, err := strconv.ParseUint(override, 10, 64)
+		if nil != err {
+			log.Warnf("unrecognized value used for the %s annotation, uint64 is expected: %s",
+				k8s.ProxyWaitBeforeExitSecondsAnnotation, override)
+		}
+
+		return waitBeforeExitSeconds, err
```
alpeb: This error is being swallowed on the calling side. You could keep on logging the error here, and then do like in the other annotation parsing methods where if there's an error a default value is used (0), and the function would always return a `uint64` only.
Also, can I ask you to use additional commits instead of force-pushing? It would make reviews a lot easier. It doesn't matter if there are lots of commits because we squash them together into a single commit prior to merging with master :wink:

KIVagant:
Tested `linkerd inject --manual` for the following set of options:

- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "100" → sleep 100
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: 100 → sleep 100 # but `kubectl apply` won't work with the integer value
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "100s" → no hook
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "-100" → no hook
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "100.100" → no hook
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "999999999999999999999999" → sleep 1.8446744073709552e+19
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "18446744073709551615" → sleep 1.8446744073709552e+19 #uint64 maximum
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "4294967295" → sleep 4.294967295e+09 #uint32 maximum
- config.alpha.linkerd.io/proxy-wait-before-exit-seconds: "65535" → sleep 65535 #uint16 maximum

KIVagant:
> Just to confirm we're on the same page, the last set of tests you posted is what users should be expecting, right?

The set represents the actual behavior of this implementation. As you may see, big numbers don't work well but if the value stays below `65535` (I'm not sure about the available maximum), which is quite enough for the goals of this solution, it works as expected.

Issue #3744, opened 2019-11-20
Follow helm chart standards


nesl247:
## Feature Request
Please follow the standards set by helm.

### What problem are you trying to solve?
A lot of the helm chart is done differently than what is done by nearly every other chart.

Examples include:

* PascalCase values
* Using `.Values.Namespace` instead of `.Release.Namespace`
* Using `{{with .Values -}}`

### How should the problem be solved?

Follow helm standards.

### Any alternatives you've considered?

Nothing to consider

### How would users interact with this feature?

Release a new major version of the chart that uses the standards. When users upgrade, they will need to make changes to their installation.

grampelberg:
We've discussed this in the past, some of these are possible and some are pretty tough:

> PascalCase values

Unfortunately it will require massive refactoring in the client code as the values are serialized out of go structures that must use PascalCase on that side. I'm not opposed, it is just a minor improvement for a quite a bit of work.

> Using .Values.Namespace instead of .Release.Namespace

We can't do anything here because of how helm is managing namespaces. It'll get a little bit better with v3 and I'd love to revisit, just not something we can do until v2 is deprecated in a year.

> Using {{with .Values -}}

This is an easy change that I'm totally behind.

nesl247:
I'm not sure I understand the issue you have with the way it manages namespaces. If you put the namespace explicitly in the templates, instead of relying upon helm to automatically add it, the release namespace will work just fine.

Regarding the PascalCase values, I'm not sure I understand. Are you saying that the helm charts are being automatically built? Is there a reason for this?

grampelberg:
The core issue is that we're using the helm chart as part of the implementation of `linkerd install`. This means that we need to pass some of the go-isms down to the chart itself currently (like PascalCase and the namespaces).

nesl247:
Personally I'd advise not having that coupling. This is the first time I've heard of this being done, and certainly doesn't lead to a good first experience with the helm charts. I honestly did not trust them at all because of their non-standard design. Because of this we had not started to try out linkerd yet.

On top of this, it looks like either the coupling has required the installation of prometheus and grafana, as there doesn't appear to be any way to disable them. Or am I misunderstanding the coupling? It seems like the only way to not install them is if `linkerd install` makes them optional.

grampelberg:
> Personally I'd advise not having that coupling.

The alternative is to maintain two distinct install flows which is unfortunate for obvious reasons.

[...]

PR #3806, opened 2019-12-09
Pods with non empty securitycontext capabilities fail to be injected

alpeb:
Followup to #3744

The `_capabilities.tpl` template got its variables scope changed in
`Values.Proxy`, which caused inject to fail when security context
capabilities were detected.

Also added a unit test to catch issues when dealing with capabilities.

Discovered when testing injecting the nginx ingress controller.


ihcsim:
I think @cpretzer ran into this earlier today.

PR #3809, opened 2019-12-11
proxy: v2.80.0

olix0r:
This release fixes a bug where service profile resolutions could backup,
clogging the proxy's connection to the destination controller. Now the
proxy takes care to eagerly consume updates as they arrive. This should
resolve a large class of 503 errors related to outdated service
discovery entries.

---

* profiles: Eagerly read profiles off the wire (linkerd/linkerd2-proxy#397)
* router: Ensure that the purge task completes (linkerd/linkerd2-proxy#396)
* app-core: Add `accept` context with peer addr (linkerd/linkerd2-proxy#398)


olix0r:
going to replace this with 2.81.0 which includes additional remedies for the bug described here

Issue #3779, opened 2019-12-03
Add pre-sign check for the identity service


ihcsim:
Following https://github.com/linkerd/linkerd2/issues/3676#issuecomment-559910383 and https://github.com/linkerd/linkerd2/issues/3676#issuecomment-560359724, it looks like the `identity` service continues to sign the proxy CSR and issue new TLS certs, even after the issuer certificate is expired. 

The [`Certify()`](https://github.com/linkerd/linkerd2/blob/66a74b23a72f7ca07d907a1a7d0504c534f6bdbb/pkg/identity/service.go#L191-L200) function should be updated to check the isser certificate validity prior to signing CSR. If expired, the function should return an error.

To reproduce:

Create short-lived trust root and issuer certificates:
```
step certificate create identity.linkerd.cluster.local ca.crt ca.key --insecure --no-password --profile=root-ca --not-after=5m

step certificate create identity.linkerd.cluster.local issuer.crt issuer.key --ca=ca.crt --ca-key=ca.key --insecure --no-password --profile=intermediate-ca --not-after=5m
```

Install Linkerd with the new certs and a short issuance lifetime:
```
linkerd install --identity-trust-anchors-file=ca.crt --identity-issuer-certificate-file=iissuer.crt --identity-issuer-key-file=issuer.key --identity-issuance-lifetime=1m | k apply -f -
```

Install the emojivoto app and watch its logs after the cert expired:
```
linkerd inject  https://run.linkerd.io/emojivoto.yml | k apply -f -

k -n emojivoto logs -f emoji-786c5d657f-t2bbt linkerd-proxy 
...
ERR! [  1381.575184s] linkerd2_proxy_identity::certify Received invalid ceritficate: invalid certificate: UnknownIssuer
ERR! [  1391.584934s] linkerd2_proxy_identity::certify Received invalid ceritficate: invalid certificate: UnknownIssuer
ERR! [  1401.594683s] linkerd2_proxy_identity::certify Received invalid ceritficate: invalid certificate: UnknownIssuer
....
```
(Adding debugging statement to  [`Certify()`](https://github.com/linkerd/linkerd2/blob/66a74b23a72f7ca07d907a1a7d0504c534f6bdbb/pkg/identity/service.go#L191-L200) will show that the CSRs continued to be signed.)

zaharidichev:
That will get addressed in #3696

PR #3811, opened 2019-12-11
Check the validity of issuer certificate before processing a CSR

zaharidichev:
The PR adds a pre-sign check to te identitiy service.

Fixes: #3779

### DIFF
--- a/pkg/identity/service.go
+++ b/pkg/identity/service.go
@@ -142,6 +142,18 @@ func Register(g *grpc.Server, s *Service) {
 	pb.RegisterIdentityServer(g, s)
 }
 
+// ensureIssuerStillValid should check that the CA is still good time wise
+// and verifies just fine with the provided trust anchors
+func (svc *Service) ensureIssuerStillValid() error {
+	issuer := *svc.issuer
+	switch is := issuer.(type) {
+	case *tls.CA:
+		return is.Cred.Verify(svc.trustAnchors, svc.expectedName)
+	default:
+		return fmt.Errorf("unsupported issuer type. Expected *tls.CA, got %v", is)
+	}
+}
+
 // Certify validates identity and signs certificates.
 func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.CertifyResponse, error) {
 	svc.issuerMutex.RLock()
@@ -157,6 +169,12 @@ func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.Ce
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+
+	if err := svc.ensureIssuerStillValid(); err != nil {
+		log.Errorf("could not process CSR because of CA cert validation failure: %s", err)
+		return nil, err
+	}
+
 	if err = checkCSR(csr, reqIdentity); err != nil {
 		log.Debugf("requester sent invalid CSR: %s", err)
 		return nil, status.Error(codes.FailedPrecondition, err.Error())


zaharidichev:
@alpeb I was going to add that to the requirements to check it on install and when running linkerd check. But yes, you are right, it iw worth wiring that here as well.

pkg/identity/service.go
```diff
@@ -104,16 +106,19 @@ func (svc *Service) Run(issuerEvent <-chan struct{}, issuerError <-chan error) {
 }
 
 func (svc *Service) loadCredentials() (tls.Issuer, error) {
```
adleong: can you explain the change in this function?  I'm not clear on what the difference is between `ReadPEMCreds` followed by `Verify` vs `LoadIssuerCrtAndKeyFromFiles` followed by `VerifyAndBuildCred`.

pkg/identity/service.go
```diff
@@ -142,6 +147,25 @@ func Register(g *grpc.Server, s *Service) {
 	pb.RegisterIdentityServer(g, s)
 }
 
+// ensureIssuerStillValid should check that the CA is still good time wise
+// and verifies just fine with the provided trust anchors
+func (svc *Service) ensureIssuerStillValid() error {
+	issuer := *svc.issuer
+	switch is := issuer.(type) {
+	case *tls.CA:
+		roots, err := tls.DecodePEMCertPool(svc.trustAnchors)
```
adleong: what's the motivation behind only storing the trustAnchor as a pem encoded string and then decoding it every time we need to check it?  as opposed to storing the decoded CertPool like we were before?

pkg/identity/service.go
```diff
@@ -142,6 +147,25 @@ func Register(g *grpc.Server, s *Service) {
...
+		if err != nil {
+			return err
+		}
+		if err := is.Cred.Verify(roots, svc.expectedName); err != nil {
+			return err
+		}
+		return nil
```
adleong: this could just be `return is.Cred.Verify(roots, svc.expectedName)`

pkg/identity/service.go
zaharidichev: Yes, so I wanted to consolidate the more "bespoke" verification logic we have into `VerifyAndBuildCred`. Namely this one also checks for algorith requirements + whether the issuer is a CA (will check for that...) + provides more explicit description around potential problems with time validity. This is the reason.

pkg/identity/service.go
zaharidichev: Yes, mainly stemming from the inconvinience that we will have to encode it as a string to execute the following piece of code here: 

```
func (svc *Service) loadCredentials() (tls.Issuer, error) {
	key, crt, err := issuercerts.LoadIssuerCrtAndKeyFromFiles(svc.issuerPathKey, svc.issuerPathCrt)
	if err != nil {
		return nil, fmt.Errorf("failed to read CA from disk: %s", err)
	}

	issuerData := &issuercerts.IssuerCertData{
		TrustAnchors: svc.trustAnchors,
		IssuerCrt:    crt,
		IssuerKey:    key,
	}
```

But now that you mention it, it  indee makes more sense to store it as a pool and ecode it to a string every time we reload the certs from disk instead of parsing it from a string every time we process a CSR as the latter cna potentially be a hotter path.  WDYT ?

zaharidichev:
@alpeb I included that check [here](https://github.com/linkerd/linkerd2/pull/3810/files#diff-982c80a8bf72499d688f1e2b93bcd73fR154) And since that is called every time the issuer cert is reloaded and read from disk, this means that we should always be ending up with an issuer in the identitiy service that is CA. So no point of checking it every time we process the CSR. WDYT ?

alpeb:
> @alpeb I included that check [here](https://github.com/linkerd/linkerd2/pull/3810/files#diff-982c80a8bf72499d688f1e2b93bcd73fR154) And since that is called every time the issuer cert is reloaded and read from disk, this means that we should always be ending up with an issuer in the identitiy service that is CA. So no point of checking it every time we process the CSR. WDYT ?

Yes, that was my original suggestion, to catch that at install/cert-reload time :+1:

pkg/identity/service.go
adleong: Can any of the "bespoke" methods be removed?  It's not clear to me why we have several different methods which do basically the same thing.

zaharidichev: Certainly agree. Just added an issue for that that I will address right after these three PRs are merged as this would touch all of them most likely. Is that fine? 

https://github.com/linkerd/linkerd2/issues/3828

pkg/identity/service.go
```diff
@@ -142,6 +152,18 @@ func Register(g *grpc.Server, s *Service) {
...
 
+// ensureIssuerStillValid should check that the CA is still good time wise
+// and verifies just fine with the provided trust anchors
+func (svc *Service) ensureIssuerStillValid() error {
+	issuer := *svc.issuer
+	switch is := issuer.(type) {
+	case *tls.CA:
+		return is.Cred.Verify(svc.trustAnchorsPool, svc.expectedName)
+	default:
+		return nil
```
ihcsim: I think we should return an error here.

pkg/identity/service.go
```diff
@@ -152,6 +174,11 @@ func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.Ce
 		return nil, status.Error(codes.Unavailable, "cert issuer not ready yet")
 	}
 
+	if err := svc.ensureIssuerStillValid(); err != nil {
+		log.Errorf("Could not process CSR because of CA cert validation failure: %s", err)
```
ihcsim: Let's use lower case for the first word in the log, per Go's [code review convention](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

ihcsim:
AIUI, the only thing that is needed in this PR is the `ensureIssuerStillValid()`. I strongly recommend that we revert all the other changes, as it feels like we are trying to shoehorn the existing code to fit the new `issuercerts` package. We DRY'ed up code in some places, but introduced redundant fields (differ only in their types) in the `identity.Service` struct. `Initialize()` now have to decode the trust root (changing the behaviour of `Run()` - though benign, but unnecessary.).  Finally, we have to create a new p0 task (https://github.com/linkerd/linkerd2/issues/3828) to rework some of these code. At this point, I feel the benefits of DRY-ing these code is overshadowed by the tech debts they introduced.

zaharidichev:
@ihcsim I agree, lets do that!

PR #3820, opened 2019-12-12
Update enzyme-adapter-react-16 library

cynthia-sg:
#### Problem

Material UI 4 requires to update `enzyme-adapter-react-16` library.

#### Solution

This PR updates `enzyme-adapter-react-16` in preparation for Material UI upgrading (#3710).

The following solutions have been applied in the tests to address the issue with Material UI `Hidden` component (more details [here](https://github.com/airbnb/enzyme/issues/2179)):

* Add tests in children components where the Hidden wrap doesn’t apply, reducing the need to inspect the children rendered markup from the parent.
* Use shallow instead of mount to render the components hierarchy. When using shallow, the problem described above with the wrapped Hidden component is not occurring as all components are being rendered.
* To not remove some other existing tests, only the state and props of some children components is tested, instead of testing their events and methods. The first point in this list should take care of testing those remaining bits though.

**Note**: Before merging #3710, we’ll need to change a little bit the new test added to `BaseTable` component to find the `FilterListIcon` component and simulate the click event.

### DIFF
--- a/web/app/js/components/BaseTable.test.js
+++ b/web/app/js/components/BaseTable.test.js
@@ -82,4 +82,37 @@ describe("Tests for <BaseTable>", () => {
     expect(emptyCard).toBeDefined();
     expect(emptyCard).toHaveLength(1);
   });
+
+  it("if enableFilter is true, user can filter rows by search term", () => {
+    let extraProps = _merge({}, defaultProps, {
+      tableRows: [{
+        deployment: "authors",
+        namespace: "default",
+        key: "default-deployment-authors",
+        pods: {totalPods: "1", meshedPods: "1"}
+      },
+      {
+        deployment: "books",
+        namespace: "default",
+        key: "default-deployment-books",
+        pods: {totalPods: "2", meshedPods: "1"}
+      }],
+      tableColumns: tableColumns,
+      enableFilter: true
+    });
+
+    const component = mount(<BaseTable {...extraProps} />);
+    const enableFilter = component.prop("enableFilter");
+    const filterIcon = component.find("FilterListIcon");
+    expect(enableFilter).toEqual(true);
+    expect(filterIcon).toHaveLength(1);
+
+    filterIcon.simulate("click");
+    setTimeout(() => {
+      const input = table.find("input");
+      input.simulate("change", {target: {value: "authors"}});
+      expect(table.html()).not.toContain('books');
+      expect(table.html()).toContain('authors');
+    }, 100);
+  });
 });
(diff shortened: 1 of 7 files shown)


web/app/js/components/MetricsTable.test.js
```diff
@@ -46,24 +46,15 @@ describe('Tests for <MetricsTable>', () => {
       enableFilter: true
     });
     const component = mount(routerWrap(MetricsTable, extraProps));
-
```
scottcarol: @cynthia-sg, thanks for splitting this out into another PR! 👍As I'm reviewing this, I'm seeing some review noise, with removed line breaks, etc. that is making it hard to see what code changed. Could you take another look at the diff here before I start reviewing? Thank you!

web/app/js/components/Version.test.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
scottcarol: I think splitting this into its own file is a good idea. However `Navigation.test.jsx` still has the original tests of `Version`, with the same titles, which would make debugging difficult (for example, `renders update message when versions do not match` now occurs in two files).

web/app/js/components/BaseTable.test.js
```diff
@@ -82,4 +82,37 @@ describe("Tests for <BaseTable>", () => {
...
+      enableFilter: true
+    });
+
+    const component = mount(<BaseTable {...extraProps} />);
+    const enableFilter = component.prop("enableFilter");
+    const filterIcon = component.find("FilterListIcon");
+    expect(enableFilter).toEqual(true);
+    expect(filterIcon).toHaveLength(1);
+
+    filterIcon.simulate("click");
```
scottcarol: For my own understanding: It looks like in this test, we are simulating a click on the `filterIcon` to test the filter functionality, but in `MetricsTable.test.js` we are simply changing `filterBy` in the state to test the filter functionality. Is there a reason to test these in different ways?

scottcarol:
Quick note that merging this PR will mean the dashboard will no longer include the subdependency `es-abstract-1.14.0` and instead will include `es-abstract-1.16.0` -- this may fix the ongoing issue where Yarn sometimes reports:

`An unexpected error occurred: "https://registry.yarnpkg.com/es-abstract/-/es-abstract-1.14.0.tgz: Request failed \"404 Not Found\"".`

web/app/js/components/BaseTable.test.js
cynthia-sg: Yes, I didn't want to remove the existing tests, but I hadn't access to button from `MetricsTable` due to `Hidden` component issue, so I decided to keep the existing tests, but in these cases only the state and props of some children components is tested, instead of testing their events and methods.

scottcarol: I see, thanks for the explanation! 👍

PR #3821, opened 2019-12-12
Prevent upgrade from breaking proxies when issuer cert is overwritten

zaharidichev:
This PR adds functionality that alloes upgrade to check whether all proxies roots can be verified with the issuer cert we are upgrading to. This is done by fetching all the proxies trust anchors from the cluster and performing vrification against the credentials we are moving to. If any of the proxies anchors do not work, a report will be produced listing them. This check can be overwritten by supplying the --force flag to upgrade. 

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/linkerd/linkerd2/pkg/config"
 	"github.com/linkerd/linkerd2/pkg/issuercerts"
@@ -34,6 +35,7 @@ const (
 
 type upgradeOptions struct {
 	manifests string
+	force     bool
 	*installOptions
 
 	verifyTLS func(tls *charts.TLS, service string) error
@@ -61,6 +63,10 @@ func (options *upgradeOptions) upgradeOnlyFlagSet() *pflag.FlagSet {
 	flags.StringVar(
 		&options.manifests, "from-manifests", options.manifests,
 		"Read config from a Linkerd install YAML rather than from Kubernetes",
+	)
+	flags.BoolVar(
+		&options.force, "force", options.force,
+		"Force upgrade operation despite validation errors",
 	)
 
 	return flags
@@ -379,6 +385,32 @@ func fetchTLSSecret(k kubernetes.Interface, webhook string, options *upgradeOpti
 	return value, nil
 }
 
+func ensureIssuerCertWorksWithAllProxies(k kubernetes.Interface, cred *tls.Cred) error {
+	meshedPods, err := healthcheck.GetMeshedPodsIdentitiyData(k, "")
+	var problematicPods []string
+	if err != nil {
+		return err
+	}
+	for _, pod := range meshedPods {
+		roots, err := tls.DecodePEMCertPool(pod.Anchors)
+
+		if roots != nil {
+			err = cred.Verify(roots, "")
+		}
+
+		if err != nil {
+			problematicPods = append(problematicPods, fmt.Sprintf("* %s", pod.Name))
+		}
+	}
+
+	if len(problematicPods) > 0 {
+		errorMessageHeader := "You are attempting to use an issuer certificate which does not validate against the trust roots of the following pods:"
+		errorMessageFooter := "These pods do not have the current trust bundle and must be restarted.  Use the --force flag to proceed anyway (this will likely prevent those pods from sending or receiving traffic)."
+		return fmt.Errorf("%s\n\t%s\n%s", errorMessageHeader, strings.Join(problematicPods, "\n\t"), errorMessageFooter)
+	}
+	return nil
+}
+
 // fetchIdentityValue checks the kubernetes API to fetch an existing
 // linkerd identity configuration.
 //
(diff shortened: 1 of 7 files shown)


cli/cmd/upgrade.go
```diff
@@ -379,6 +385,32 @@ func fetchTLSSecret(k kubernetes.Interface, webhook string, options *upgradeOpti
...
+	}
+	for _, pod := range meshedPods {
+		roots, err := tls.DecodePEMCertPool(pod.Anchors)
+
+		if roots != nil {
+			err = cred.Verify(roots, "")
+		}
+
+		if err != nil {
+			problematicPods = append(problematicPods, fmt.Sprintf("* %s", pod.Name))
```
adleong: I think we need the pod namespace in here too, like in https://github.com/linkerd/linkerd2/pull/3821/files#diff-d4056ff163bcf2aeacefb2a34164563cR1445

cli/cmd/upgrade.go
```diff
@@ -379,6 +385,32 @@ func fetchTLSSecret(k kubernetes.Interface, webhook string, options *upgradeOpti
 	return value, nil
 }
 
+func ensureIssuerCertWorksWithAllProxies(k kubernetes.Interface, cred *tls.Cred) error {
```
adleong: I think the Linkerd control plane needs to be excluded from this check.  When I run through the cert rotation process, I get this error while replacing the issuer cert:

```
$ ../bin/linkerd upgrade --identity-issuer-certificate-file=./issuer-new.crt --identity-issuer-key-file=./issuer-new.key | kubectl apply -f -
× Failed to build upgrade configuration: You are attempting to use an issuer certificate which does not validate against the trust roots of the following pods:
	* linkerd-controller-8697cfcbd5-fns28
	* linkerd-destination-84979b7dd8-pgdsq
	* linkerd-grafana-7966cbb54-2rrsm
	* linkerd-identity-68c944bbd9-dv2nv
	* linkerd-proxy-injector-777cf54dd8-rw9rw
	* linkerd-sp-validator-8559d89bb7-dp4d8
	* linkerd-tap-5458484f46-nxdwz
	* linkerd-web-56db6684cb-t8dh5
These pods do not have the current trust bundle and must be restarted.  Use the --force flag to proceed anyway (this will likely prevent those pods from sending or receiving traffic).

```

It should not be necessary to manually restart the control plane proxies, since the control plane will automatically get rolled during upgrade anyway.

cli/cmd/upgrade.go
```diff
@@ -419,6 +453,18 @@ func (options *upgradeOptions) fetchIdentityValues(k kubernetes.Interface, idctx
 		return nil, err
 	}
 
+	cred, err := issuerData.VerifyAndBuildCreds("")
```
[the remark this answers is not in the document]
zaharidichev: I think this can fail for more than one reason. Not necessarily because the verification logic faile.d It might be that the issuer cert does not have the correct crypo requirements and we need to catch that as well. I have changed the error to be of the following format:
```
× Failed to build upgrade configuration: issuer certificate does not work with the provided provided roots: x509: certificate signed by unknown authority
For more information: https://linkerd.io/2/tasks/rotating_identity_trust_root/
```

Would oyu say that works?

cli/cmd/upgrade.go
zaharidichev: @adleong I think what is happening here is that you have ran these command one after the other in a short period of time and the list of these pods that is displayed here contains the pods that are currently being terminated and replaced with the new pods containing the bundled roots. If you wait a bit longer for them to go away, this will dissapear and you will be allowed to execute the command. In fact if you run the `linkerd check --proxy` cmd at that point you will see that it issues a warning that some proxies do not match the anchors in the config map and if you keep executing the command this list will gradually shrink as part of the proxy restart process. That being said, maybe we should mention it in the docs that users need to wait until the control plane is fully "upgraded" before updating the issuer command. Does that sound correct ? I am not sure we want to hide this error as it gives us extra assurance that the issuer certs work with the proxies of the control plane

adleong: ah, you're saying that the command to add the new trust root to the bundle should cause the control plane to restart?

cli/cmd/upgrade.go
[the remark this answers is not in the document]
adleong: Yup!  Having that link to the cert rotation docs is the most important thing.

cli/cmd/upgrade.go
zaharidichev: yes that is correct. You just need to wait a bit until all of that completes. These errors are just caused by the old pods still hanging around. In that case do you think we should change something ? I am not convinced we should

cli/cmd/upgrade.go
```diff
@@ -419,6 +453,18 @@ func (options *upgradeOptions) fetchIdentityValues(k kubernetes.Interface, idctx
...
+	if err != nil {
+		return nil, fmt.Errorf("issuer certificate does not work with the provided provided roots: %s\nFor more information: https://linkerd.io/2/tasks/rotating_identity_trust_root/", err)
```
ihcsim: s/provided provided/provided

cli/cmd/upgrade.go
```diff
@@ -62,6 +64,10 @@ func (options *upgradeOptions) upgradeOnlyFlagSet() *pflag.FlagSet {
 		&options.manifests, "from-manifests", options.manifests,
 		"Read config from a Linkerd install YAML rather than from Kubernetes",
 	)
+	flags.BoolVar(
+		&options.force, "force", options.force,
+		"Force upgrade operation despite validation errors",
```
ihcsim: I wonder if we can add some clarity to the description on which validation is skipped. It seems like this only skips the `ensureIssuerCertWorksWithAllProxies()` validation, even though it sounds like it will skip all validations (including those on algorithm, validity period, trust chain etc.). 

IOW, the following usage of `--force` won't allow me to just override my expired issuer cert, unless it is part of the trust root's chain, right?

```
bin/linkerd upgrade --force --identity-issuer-certificate-file=issuer.crt --identity-issuer-key-file=issuer.key
× Failed to build upgrade configuration: issuer certificate does not work with the provided provided roots: x509: certificate signed by unknown authority (possibly because of "x509: ECDSA verification failure" while trying to verify candidate authority certificate "identity.linkerd.cluster.local")
For more information: https://linkerd.io/2/tasks/rotating_identity_trust_root/
```

PR #3824, opened 2019-12-12
When running `kind load` in the build host, don't ignore `ssh`'s stderr

alpeb:
Trying to debug current CI failures...

Edit: More context:
Last week we were having lots of CI failures in different tests, where the the process was simply dying with no error message. The problem ended up being lack of disk space. The current change attempted having more output on the `kind_integration` step. It finally didn't help debugging the issue but I think it's a good change, that avoids swallowing the output of the `kind load` commands.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -354,7 +354,7 @@ jobs:
       run: |
         TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
         export KIND_CLUSTER=github-$TAG-${{ matrix.integration_test }}
-        ssh -T linkerd-docker &> /dev/null << EOF
+        ssh -T linkerd-docker > /dev/null << EOF
           # TODO: This is using the kind binary on the remote host.
           kind load docker-image gcr.io/linkerd-io/proxy-init:v1.2.0 --name=$KIND_CLUSTER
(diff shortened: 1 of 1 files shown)


alpeb:
I'm also adding this call
```
bin/docker image prune -f
```
before building the images, to keep disk usage in check.

.github/workflows/workflow.yml
```diff
@@ -270,7 +270,6 @@ jobs:
         DOCKER_HOST: ssh://linkerd-docker
       run: |
         export PATH="`pwd`/bin:$PATH"
-        bin/docker image prune -f
```
olix0r: just to confirm -- this will only remove dangling images? specifically: it will not interfere with other concurrent builds?

alpeb: Yes, those layers that have no relationship to other tagged images, in theory. In practice I observed some new weird `kind` errors appearing, so I backed out of this cleanup finally. I'll document it in the packet docs as a "maintenance item" to keep out of the build.

adleong:
@alpeb is this ready to be reviewed?  can you give more context in the description?

alpeb:
> @alpeb is this ready to be reviewed? can you give more context in the description?

Yes it's ready. I've just added more context.

Issue #3822, opened 2019-12-12
Helm Charts Clean-Up


Pothulapati:
As the plan is to use helm charts as the core-parts for Linkerd Add-Ons per #3794. It was also planned to make changes to follow [best-practices ](https://helm.sh/docs/topics/chart_best_practices/values/#naming-conventions) as a pre-requisite. 

The following tasks have been taken up for the same:

- [x] Make `linkerd2` chart use smaller-case values, support the same for `cli` and inject`
- [x] Make `linkerd2-cni` use smaller-case
- [x] Move common values out to `global` (in preparation for #3794 )

@ihcsim @grampelberg

christyjacob4:
I think this can be closed in light of the merged PRs

Pothulapati:
Yes, Thanks!

PR #3823, opened 2019-12-12
Switch to Smaller-Case in Linkerd2 and Partials Charts

Pothulapati:
First Part of #3822 

This PR
- Updates `values.yaml` and `values-ha.yaml` in `linkerd2` to smaller-case values.
- Updates helm templates in `linkerd-2`, `partials` for the same.
- Update `values.go` to serialize into lower-case.
- Update `patch` sub-chart to use lower-case, along with `inject.go` for the same.

@ihcsim @grampelberg

PR #3827, opened 2019-12-13
Switch to smaller-case values in linkerd2-cni

Pothulapati:
Second Part of #3822 

This PR
- Updates `linkerd2-cni\values.yaml` to use smaller-case along with templates.
- Updates `charts/linkerd2-cni/values.go` to serialize into smaller case.
- Update comments and readme


alpeb:
I just wanted to confirm, is the convention for acronyms to use lowercase at the beginning of a label and uppercase elsewhere? like `cniPluginImage` and `destCNIBinDir`?. Likely so given it'd be weird to have `destCniBinDir`, right?

charts/linkerd2/values.yaml
```diff
@@ -7,7 +7,7 @@ enableH2Upgrade: true
 imagePullPolicy: &image_pull_policy IfNotPresent
 
 # control plane version. See Proxy section for proxy version
-LinkerdVersion: &linkerd_version edge-19.12.2
+linkerdVersion: &linkerd_version edge-19.12.2
```
Pothulapati: Missed this change in #3823

Issue #2657, opened 2019-04-06
Pass arbitrary pod labels through as prometheus labels


nickjackson:
## Feature Request
Pass arbitrary pod labels through as prometheus labels like [these docs](https://linkerd.io/2/reference/proxy-metrics/#example ) suggest you already can.

### What problem are you trying to solve?
I'm looking to build some custom graphs that require some extra prometheus labels to be passed through. 

In particular, I want to pass through two labels that might look something like this `app=foo, version=v1.0.0`

### How should the problem be solved?
Not certain. Is it as simple as modifying the prometheus scrape config for the `linkerd-proxy` [here](https://github.com/linkerd/linkerd2/blob/65a228fe6c4a4daac6190889834692790d0d926d/chart/templates/prometheus.yaml#L198) to allow all labels through?

### Any alternatives you've considered?
So currently by default all labels prefixed with `linkerd.io/` will be passed through and remapped. I'm currently using this to get my labels through, but this is clearly not ideal.

### How would users interact with this feature?
Just add labels to the pod spec as they normal do.

olix0r:
One issue to overcome is that, i believe, prometheus cannot have labels with dots, slashes, or dashes; so we'll need some rules about translation...

siggy:
For reference, we used to have this functionality, I'd be in favor of adding it back:
https://github.com/linkerd/linkerd2/pull/1402#discussion_r207943459

matipan:
Is this issue being worked on? Since is a good first issue I would be ok with working on it 👍

siggy:
@matipan Go for it! I'd love to see this get implemented!

daxmc99:
@matipan Are you still working on this? 
@siggy Is it ok to add that line back? 

Looks like this is the place to add it 
https://github.com/linkerd/linkerd2/blob/b8f7b959d2efcef8639fb5615a339c9ff5579ca4/cli/install/template.go#L480

I am doing this personally for some metrics scraping 
https://github.com/rancher/mesh-metrics/blob/master/prometheusConfig.yaml#L99

I am a bit of loss of how to safeguard Prometheus though. Is the worry about query time increasing?

siggy:
@daxmc99 I think it's fine to add this back.

[...]

PR #3833, opened 2019-12-16
Add scraping of arbitrary pod labels

daxmc99:
This allows for users of Linkerd to leverage the Prometheus instance
deployed by the mesh for their metric needs. With support for pod labels
outside of the Linkerd metrics users are able to scrape metrics
based upon their own labels.

Addresses https://github.com/linkerd/linkerd2/issues/2657

Signed-off-by: Dax McDonald <dax@rancher.com>

### DIFF
--- a/charts/linkerd2/templates/prometheus.yaml
+++ b/charts/linkerd2/templates/prometheus.yaml
@@ -118,6 +118,20 @@ data:
       # foo=bar
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
+      # Copy all pod labels to tmp labels
+      - action: labelmap
+        regex: __meta_kubernetes_pod_label_(.+)
+        replacement: __tmp_pod_label_$1
+      # Take `linkerd_io_` prefixed labels and copy them without the prefix
+      - action: labelmap
+        regex: __tmp_pod_label_linkerd_io_(.+)
+        replacement:  __tmp_pod_label_$1
+      # Drop the `linkerd_io_` originals
+      - action: labeldrop
+        regex: __tmp_pod_label_linkerd_io_(.+)
+      # Copy tmp labels into real labels
+      - action: labelmap
+        regex: __tmp_pod_label_(.+)
 ---
 kind: Service
 apiVersion: v1
--- a/cli/cmd/testdata/install_control-plane.golden
+++ b/cli/cmd/testdata/install_control-plane.golden
@@ -1065,6 +1065,20 @@ data:
       # foo=bar
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
+      # Copy all pod labels to tmp labels
+      - action: labelmap
+        regex: __meta_kubernetes_pod_label_(.+)
+        replacement: __tmp_pod_label_$1
+      # Take `linkerd_io_` prefixed labels and copy them without the prefix
+      - action: labelmap
+        regex: __tmp_pod_label_linkerd_io_(.+)
+        replacement:  __tmp_pod_label_$1
+      # Drop the `linkerd_io_` originals
+      - action: labeldrop
+        regex: __tmp_pod_label_linkerd_io_(.+)
+      # Copy tmp labels into real labels
+      - action: labelmap
+        regex: __tmp_pod_label_(.+)
 ---
 kind: Service
 apiVersion: v1
--- a/cli/cmd/testdata/install_controlplane_tracing_output.golden
+++ b/cli/cmd/testdata/install_controlplane_tracing_output.golden
@@ -1923,6 +1923,20 @@ data:
       # foo=bar
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
+      # Copy all pod labels to tmp labels
+      - action: labelmap
+        regex: __meta_kubernetes_pod_label_(.+)
+        replacement: __tmp_pod_label_$1
+      # Take `linkerd_io_` prefixed labels and copy them without the prefix
+      - action: labelmap
+        regex: __tmp_pod_label_linkerd_io_(.+)
+        replacement:  __tmp_pod_label_$1
+      # Drop the `linkerd_io_` originals
+      - action: labeldrop
+        regex: __tmp_pod_label_linkerd_io_(.+)
+      # Copy tmp labels into real labels
+      - action: labelmap
+        regex: __tmp_pod_label_(.+)
 ---
 kind: Service
 apiVersion: v1
--- a/cli/cmd/testdata/install_custom_registry.golden
+++ b/cli/cmd/testdata/install_custom_registry.golden
@@ -1903,6 +1903,20 @@ data:
       # foo=bar
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
+      # Copy all pod labels to tmp labels
+      - action: labelmap
+        regex: __meta_kubernetes_pod_label_(.+)
+        replacement: __tmp_pod_label_$1
+      # Take `linkerd_io_` prefixed labels and copy them without the prefix
+      - action: labelmap
+        regex: __tmp_pod_label_linkerd_io_(.+)
+        replacement:  __tmp_pod_label_$1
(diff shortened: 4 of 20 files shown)


charts/linkerd2/templates/prometheus.yaml
```diff
@@ -114,10 +114,9 @@ data:
       # drop all labels that we just made copies of in the previous labelmap
       - action: labeldrop
         regex: __meta_kubernetes_pod_label_linkerd_io_proxy_(.+)
-      # __meta_kubernetes_pod_label_linkerd_io_foo=bar =>
-      # foo=bar
+      # __meta_kubernetes_pod_label_foo=bar => foo=bar
```
adleong: does this means that labels which previously had a `linkerd_io` prefix stripped off will now have the `linkerd_io` prefix in the label name?

zaharidichev: @adleong Just to get a bit more context here. What are the implications of that prefix not being stripped off?

adleong: I'm wondering if this will cause the label names to be different, breaking existing prometheus queries

daxmc99: While I believe this does cause the `linkerd_io` labels appear on the Prometheus metrics it doesn't appear to break any existing queries. I will be doing more to investigate this

zaharidichev: @daxmc99 Any results from the  investigation?

daxmc99: @zaharidichev Sorry, I was on vacation. I will be looking at this during the week.

daxmc99:
There is a change to labels introduced here: 
For a prometheus query `control_response_latency_ms_bucket`: 
Before:
```
control_response_latency_ms_bucket{addr="linkerd-dst.linkerd.svc.cluster.local:8086",control_plane_component="controller",control_plane_ns="linkerd",deployment="linkerd-controller",instance="10.42.0.9:4191",job="linkerd-proxy",le="+Inf",namespace="linkerd",pod="linkerd-controller-865847cc56-rxqvb",status_code="200",tls="true"}
```

After: 
```
control_response_latency_ms_bucket{addr="linkerd-dst.linkerd.svc.cluster.local:8086",app="emoji-svc",deployment="emoji",instance="10.42.0.23:4191",job="linkerd-proxy",le="+Inf",linkerd_io_control_plane_ns="linkerd",namespace="emojivoto",pod="emoji-d8d7d9c6b-8bbpj",pod_template_hash="d8d7d9c6b",status_code="200",tls="true"} | 1
```
 These are not the same metrics but you can see in the second one the issue with the creation of `linkerd_io_control_plane_ns`. I'm going to change this to do the `labelmap` then `labeldrop` pattern futher up in the configuration.

daxmc99:
@zaharidichev PTAL

image
This allows for the labels like  `app` to be used in queries. The screenshot is from the Prometheus targets page of the Linkerd prometheus pod

charts/linkerd2/templates/prometheus.yaml
daxmc99: Following up on this; yes this did cause the label format to change however I did not see this cause any queries to break ( I didn't see any queries that depended on this field but I might have missed one). 
I ended up doing this with plus a `labeldrop` command to remove the duplicate entries.

charts/linkerd2/templates/prometheus.yaml
```diff
@@ -118,6 +118,10 @@ data:
       # foo=bar
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
+      - action: labelmap
```
adleong: It took me a few minutes to wrap my head around these rules but I think I get it.

First we add all pod labels that start with `linkerd_io`, dropping the `linkerd_io` prefix.  So for example, the `linkerd.io/foobar` pod label becomes `foobar` as a Prometheus label.

Next we add all other pod labels.  So the pod label `bas` becomes `bas` as a Prometheus label.  Unfortunately, this also copies those `linkerd.io` labels as well so the `linkerd.io/foobar` pod label from above also becomes `linkerd_io_foobar` as a Prometheus labels.

To correct this, we now drop all labels which start with `linkerd_io_`.  

This works pretty well but it is a bit confusing and may lead to a "gotcha" in the future:

If, in the future, we have any Prometheus labels (not necessarily from pod labels) which start with `linkerd_io_`, they will get dropped.

What do you think about changing this to something like

```
      # Copy all pod labels to tmp labels
      - action: labelmap
        regex: __meta_kubernetes_pod_label_(.+)
        replacement: __tmp_pod_label_$1
      # Drop `linkerd_io_` prefix from any of those tmp labels
      - action: labelmap
        regex: __tmp_pod_label_linked_io_(.+)
        replacement:  __tmp_pod_label_$1
      # Copy tmp labels into real labels
      - action: labelmap
        regex: __tmp_pod_label_(.+)
```

Do you think that makes it clearer what is happening?

alpeb: @adleong Note this will also pick the stock labels like `linkerd.io/control-plane-ns: linkerd`, leaving us with duped labels: their complete version and their stripped version (stripped in the "foo=bar" rule above).

adleong: @alpeb you're right, I forgot that the second action there won't delete the old label that includes the prefix.  Let me experiment a bit and see if I can come up with something that works...

adleong: Ok, this is what I've come up with:

```
      # Copy all pod labels to tmp labels
      - action: labelmap
        regex: __meta_kubernetes_pod_label_(.+)
        replacement: __tmp_pod_label_$1
      # Take `linkerd_io_` prefixed labels and copy them without the prefix
      - action: labelmap
        regex: __tmp_pod_label_linked_io_(.+)
        replacement:  __tmp_pod_label_$1
      # Drop the `linkerd_io_` originals
      - action: labeldrop
        regex: __tmp_pod_label_linkerd_io_(.+)
      # Copy tmp labels into real labels
      - action: labelmap
        regex: __tmp_pod_label_(.+)
```

This behaves almost identically to what you have, @daxmc99.  The only difference is that it won't drop any existing prometheus labels that start with `linkerd_io_` because all the dropping happens on the tmp labels.  But I think what's most important is clarity.  @daxmc99 and @alpeb, do either of you have an opinion on which of these options is easier to understand?

daxmc99: I like this approach much better. A quick smoke test and this achieves the same result I was looking for.

adleong: I'm still seeing labels with the `linkerd_io_` prefix.  I think you need to add a step to drop these from the tmp labels.  See the action labeled `# Drop the `linkerd_io_` originals` in my comment above.

charts/linkerd2/templates/prometheus.yaml
```diff
@@ -118,6 +118,17 @@ data:
       # foo=bar
       - action: labelmap
         regex: __meta_kubernetes_pod_label_linkerd_io_(.+)
+      # Copy all pod labels to tmp labels
+      - action: labelmap
+        regex: __meta_kubernetes_pod_label_(.+)
+        replacement: __tmp_pod_label_$1
+      # Drop `linkerd_io_` prefix from any of those tmp labels
+      - action: labelmap
+        regex: __tmp_pod_label_linked_io_(.+)
```
alpeb: there's a typo here: `__tmp_pod_label_linked_io_`

daxmc99:
@adleong PTAL 
Screen Shot 2020-01-20 at 11 23 58 AM
This looks like it solves the issue in a more maintainable way. I tested this with Rio and it will also solve our use case.

daxmc99:
Also to aid in testing you can kubectl apply this https://gist.github.com/daxmc99/c9868ce659ec78c3ab605565327540f4 

` curl -sL https://gist.githubusercontent.com/daxmc99/c9868ce659ec78c3ab605565327540f4/raw/c491b81909d141bc86cb8a5e16163a2e24cfc038/Prometheus-proposed-config.yaml | kubectl apply -f - `

PR #3835, opened 2019-12-16
Add mutating webhook for gateway annotations

### LINKED ISSUE #3738: [Feature] Mutating webhook for ingress annotations
## Feature Request
A webhook or extension of the existing injector that will annotate ingresses automatically.

Having to manually add annotations to every ingress object.

I'm thinking a mutating webhook that will accept some configuration on install such that any ingress created within a namespace that is annotated for sidecar injection will get the necessary config annotations. A couple of things will be required:

* Some sort of configuration that specifies what type of controller, or some detection of controller type based on ingress annotations
* Possible that exclusion annotations might be necessary
* The ability to turn on/off the injector

* An external mutating webhook controller

tegioz:
#### Problem

To inject Linkerd into ingress controller's pods, users have to do some manual configuration, which usually requires adding a custom header to the outgoing request. The way this is done depends on the ingress controller used.

#### Solution

This PR adds a new mutating webhook that automatically adds the necessary Linkerd configuration to the ingress object (or any other k8s object or CRD, depending on the controller), making Linkerd even easier to setup.

At the moment only Nginx and Traefik (<=1.7) controllers are supported.

Some details to be defined:

- Global configuration for this webhook (enabled/disabled, special per-controller settings, etc)
- Injection overriding at the resource level (something like `linkerd.io/inject: disabled`)

#### Validation

Create the following `ingress` objects in a cluster and they should be automatically annotated with the Linkerd header.

**Nginx**

```yaml
kind: Ingress
metadata:
  name: tests-ingress-nginx
  annotations:
    kubernetes.io/ingress.class: "nginx"
spec:
  rules:
    - host: tests-host
      http:
        paths:
          - path: /
            backend:
              serviceName: tests-svc
              servicePort: 80
```

**Traefik**

```yaml
kind: Ingress
metadata:
  name: tests-ingress-traefik
  annotations:
    kubernetes.io/ingress.class: "traefik"
spec:
  rules:
    - host: tests-host
      http:
        paths:
          - path: /
            backend:
              serviceName: tests-svc
              servicePort: 80
```

Fixes #3738


alpeb:
Thanks @tegioz !
I personally love this approach. Given it's going to handle so many cases (5?), I think it grants having its own webhook to handle them all, separate from `linkerd-proxy-injector`.
If enabled by default, we'd have to make sure the ingress still works when the ingress/service resource has these annotations added but they ingress controller isn't yet injected.
What do other folks think of this?

alpeb:
Quick side note: `/controller/proxy-injector/webhook_ops.go` is a leftover from previous times that I'll have to remove in a separate PR. Likely, it's not needed here as well.

ihcsim:
@tegioz Thanks for putting this together.

Instead of implementing the different ingress controller annotators in Go, I wonder if this is an opportunity to utilize CRD. In my ideal world, it will be great if users don't have to wait for us to implement support for the different ingress controllers.

Maybe we can define a CRD to capture the ingress controllers config that looks like this:
```
metadata:
  name: ingresscfg.linkerd.io
  annotations:
    {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
  labels:
    {{.Values.controllerNamespaceLabel}}: {{.Values.namespace}}
spec:
  group: linkerd.io
  versions:
  - name: v1alpha1
    served: true
    storage: true
  scope: Cluster
  names:
    plural: ingresscfg
    singular: ingresscfg
    kind: IngressCfg
    shortNames:
    - icfg
  validation:
    openAPIV3Schema:
      type: object
      properties:
        spec:
          type: object
          properties:
            annotations:
              type: object
            vars:
              type: object
```
Users will create their own CRs that look like:
```
kind: IngressCfg
metadata:
  name: nginx-ingress
spec:
  annotations:
    nginx.ingress.kubernetes.io/configuration-snippet: |
      proxy_set_header l5d-dst-override $service_name.$namespace.svc.cluster.local:$service_port;
      grpc_set_header l5d-dst-override $service_name.$namespace.svc.cluster.local:$service_port;
```

AFAICT, other ingress controllers like Traefik and Ambassador may not have built-in variables like `$service_name`. (We should confirm this.) In such cases, maybe we can introduce our own variables injection mechanism:
```
kind: IngressCfg
metadata:
  name: trafik-ingress
spec:
  annotations:
    ingress.kubernetes.io/custom-request-headers: l5d-dst-override:$service_name.$namespace.svc.cluster.local:$port
  vars:
     service_name: web-svc
     namespace: emojivoto
     port: 80
```

Users will also annotate their `Ingress` to let us know which `IngressCfg` resource to use:
```
kind: Ingress
metadata:
  name: web-ingress
  namespace: emojivoto
  annotations:
    kubernetes.io/ingress.class: "nginx"
    config.linkerd.io/ingresscfg: "nginx-ingress"
```

Then the webhook will fetch the right `IngressCfg` resource, based on the provided `config.linkerd.io/ingresscfg` annotation, and mutate the `Ingress` object accordingly. I haven't done any Go-level POC on this yet, but it might work?

I like how simple the RBAC of `gateway-annotator-rbac.yaml` is. (Ultimately, the webhook only needs to deal with admission requests and responses, not actual K8s resource kinds.)

That leads me to think if we really need a second webhook for this, leading back to my previous comment at https://github.com/linkerd/linkerd2/issues/3738#issuecomment-562273377. It feels like we are introducing 7 additional resources for a relatively small service. If the proposed CRD approach works, the scope of this servce will even be smaller. (The new RBAC scope is already covered by the proxy injector's.)

grampelberg:
@ihcsim since we're just adding annotations ..... is there a k8s ecosystem project that does this for us? I assume gatekeeper can be hacked to do it, but there must be *something* that allows you to automatically add labels and annotations to workloads.

ihcsim:
> is there a k8s ecosystem project that does this for us?

Yeah, @alpeb and I were just discussing this earlier. We haven't come across any yet. What we need is an extension of [`PodPreset`](https://kubernetes.io/docs/concepts/workloads/pods/podpreset/).

> I assume gatekeeper can be hacked to do it,

I will need to check with my source :tm: to explore this. But Gatekeeper only runs at the validation phase, so I'm almost positive that this isn't doable with Gatekeeper.

tegioz:
> since we're just adding annotations ..... is there a k8s ecosystem project that does this for us?

@grampelberg please keep in mind that just adding annotations might not be enough if the plan is to support all controllers mentioned in the docs page (plus eventually new ones). 

As an example, in Traefik 2.0 headers are set in the spec of a middleware CRD that can be referenced from an `IngressRoute` (another CRD). We may need to create that middleware object if it doesn't exist, or attach it to routes in their spec.

https://docs.traefik.io/v2.0/middlewares/overview/
https://docs.traefik.io/v2.0/middlewares/headers/

Ambassador also uses a CRD for configuration since version 0.7.

https://www.getambassador.io/reference/core/crds/

But even if it was just playing with annotations, there is some custom logic around it so I think we'd need a flexible solution that allows us to do anything we need. For example, AFAIK there is no way to use variables when defining headers for Traefik. This means we can only annotate it properly if there is only a combination of service/port in the ingress backend or rules as both need to appear in the header we set. This can get complicated quickly as we need to inspect the content of the ingress on each case and add, update or even delete the header as related objects are added or updated.

tegioz:
> Instead of implementing the different ingress controller annotators in Go, I wonder if this is an opportunity to utilize CRD. In my ideal world, it will be great if users don't have to wait for us to implement support for the different ingress controllers.

The problem I see with this approach is that we won't be only dealing with `Ingress` objects. As I mentioned in my previous comment, there are already a few controllers which can be configured differently, and seeing the evolution of some of them it's likely it'll happen again. I think the effort to make something like this fully configurable for such a variety of objects using CRDs will be bigger than creating them manually in Go once.

I thought the problem behind this issue was that users at the moment need to manually do some extra configuration to make Linkerd injected ingresses work fine. If we ask them to create some CRD and annotate their ingress objects to make use of it, I think the experience it terms of ease of use won't be that much different of what they do today. I think we should aim for an approach that works out of the box with zero configuration and makes the life of a lot of users easier, probably supporting the most widely used controllers.

grampelberg:
> @grampelberg please keep in mind that just adding annotations might not be enough if the plan is to support all controllers mentioned in the docs page (plus eventually new ones).

Hmm, that's a great point. You could probably do some of that with rego, but at that point it might make more sense for us to just do it in code.

This highlights a more important part though. If this is going to support multiple different ingress with different definitions, we'll need to think through which ones make sense. For now, my vote is to only support the ingresses that allow annotations. Ambassador already has support built in, we're not sure where `IngressRoute` is going and `VirtualService` seems to be a one-off right now. Ideally, this would be pluggable and the ingress providers could take care of it.

> Maybe we can define a CRD to capture the ingress controllers config that looks like this:

I'm a -1 to this right now. For a lot of the same reasons @tegioz has mentioned.

ihcsim:
> I think we should aim for an approach that works out of the box with zero configuration and makes the life of a lot of users easier

That's a very good point. I am convinced.

> We may need to create that middleware object if it doesn't exist, or attach it to routes in their spec.

We probably don't want to couple ourselves directly to the CRDs of other ingress controllers, right? @grampelberg will the CLI addon approach works here, where support for each ingress controller is implemented as an individual mutating webhook addon? The potential downside is if I use N different types of ingress controllers, I will have N mutating webhook addons.

grampelberg:
> We probably don't want to couple ourselves directly to the CRDs of other ingress controllers, right? 

I'm not necessarily against it. We *do* need to make it so that it is really easy to contribute and/or do your own thing though.

> @grampelberg will the CLI addon approach works here, where support for each ingress controller is implemented as an individual mutating webhook addon? 

If we agree that a high level goal is to make it really easy for folks to contribute .... would that be the best solution? Feels like a lot of boilerplate and custom code that'll drift. Let's not forget that in reality the 99% use case here is nginx. While folks use the other ingress solutions, it is super rare for them to run into any problems.

ihcsim:
The reason why I suggested exploring th CLI addon approach is to decouple the control plane code base from the controller's CRDs, some of which are still in `v1alpha1`, by storing this addons in a separate GH repo that have its own release lifecycle. The community can contribute to it there, and we can expose some Linkerd packages or use kubebuilder to autogenerate the webhook code. 

If for now, we are only going to support ingress that allows annotations, it feel like a lot of work and code just for that.

grampelberg:
We should sort more of the addon story out before we go down that route. Having it be a separate repo seems to be nice from a contribution perspective. We just need to figure out how we want to structure stuff a little bit more.

It sounds like focusing on just doing annotations for now is a good scoping exercise and lets us defer some larger pieces of work that we don't have requirements quite yet for.

PR #3839, opened 2019-12-17
Move common values to global

Pothulapati:
Third Part of #3822 

This PR

- Moves common values in `values.yaml`, `values-ha.yaml` and templates to `global` field in yaml.
- Updates linkerd cli install and upgrade commands.
- Updates `pkg/inject.go` as the same templates for injection.
- Update relevant tests across both `cli` and `helm` installations.

@grampelberg @ihcsim @alpeb @zaharidichev


alpeb:
Also please update the `README.md` files :grin:

charts/linkerd2/values.yaml
```diff
@@ -2,14 +2,101 @@
...
+  clusterDomain: &cluster_domain cluster.local
+  imagePullPolicy: &image_pull_policy IfNotPresent
+
+  # control plane version. See Proxy section for proxy version
+  linkerdVersion: &linkerd_version edge-19.12.3
+
+  namespace: linkerd
+
+  # identity configuration
+  identity:
```
Pothulapati: Before fixing the docs, One concern I have is to not move `identity` to global. Only `identity.trustAnchorsPEM`, `identity.trustDomain` is needed for us. Nothing from `identity.issuer`. 

So I did some changes to move `identity.issuer` out of global, and renamed `trustAnchorsPEM`, `trustDomain` to `global.identityTrustAnchorsPEM`,`global.identityTrustDomain` (i.e out of Identity and to global). This required changes in `install.go`, etc. I took the path of using a new struct there. 

The changes can be seen in this new branch here https://github.com/Pothulapati/linkerd2/commit/0351c9b774c4b14ed5d5b7b136c101516e6e8c61 . I can push that commit here, if its acceptable and then fix the docs. 



@zaharidichev @alpeb

zaharidichev: This fragments things a bit but on the other hand make it quite DRY. One thing I have to say is I am not sure that `completeIdentity` is the right name for the struct. Maybe IdentitiyWithAnchorsAndTrustDomain (...not sure that sounds much better). I do not have any objections against the approach. In situations like these however the limitation of the go type system shine through :D

alpeb: @Pothulapati Agreed about the variables refactoring. As for the naming, I don't have a great suggestion. Not a big deal IMO, given it's a private helper struct.

Pothulapati: Updated using the helper struct.

cpretzer:
@alpeb, confirmed that it worked without setting `identity.issuer.crtExpiryAnnotation`

@Pothulapati nice work! 🚢

PR #3845, opened 2019-12-18
Increase minimum k8s version to 1.13

adleong:


alpeb:
@adleong Could you also change the `kubeVersion` value in all the `Chart.yaml` and `README.md` files under `/charts`?

PR #3847, opened 2019-12-18
Add recommended k8s labels to control-plane

### LINKED ISSUE #3816: set Kubernetes common labels on control plane components
To make it easier for tools that consume these things, Linkerd should set some of the [Kubernetes common labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for control plane components.

In particular I think we should set:
* `app.kubernetes.io/name` to the name of the component, e.g. `prometheus` or `destination`
* `app.kubernetes.io/version` to the corresponding Linkerd version
* `app.kubernetes.io/part-of` to "Linkerd"

There are some other fields but those are IMO the important ones.

siggy:
The Kubernetes docs recommend a common set of labels for resources:
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Add the following 3 labels to all control-plane workloads:
```
app.kubernetes.io/name: controller # or destination, etc
app.kubernetes.io/version: edge-X.Y.Z
```

Fixes #3816

### DIFF
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -31,6 +31,9 @@ metadata:
   annotations:
     {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
   labels:
+    app.kubernetes.io/name: controller
+    app.kubernetes.io/part-of: Linkerd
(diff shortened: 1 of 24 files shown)


siggy:
@adleong Fair question re: which resources should have these labels. For now let's err on the side of less, focusing on just the high-level resources (deployments and cronjobs). If users request these labels on other resources (service, pod, role?), we can always add them later.

PR #3857, opened 2019-12-19
Changes for edge-19.12.3

cpretzer:
## edge-19.12.3

* CLI
  * Increased the comprehensiveness of `linkerd check --pre`
  * Added TLS certificate validation to `check` and `upgrade` commands
* Controller
  * Increased minimum kubernetes version to 1.13.0
  * Added support for pod ip and service cluster ip lookups in the destination 
    service
  * Added recommended kubernetes labels to control-plane
  * Added preStop hook into the proxy sidecar container to stop it last (a huge 
    commit from @KIVagant, thanks!)
  * Added a pre-sign check to te identity service
* Web UI
  * Added logic to pause dashboard polling activity when the page is not in 
    focus
* Proxy
  * Added a timeout to release resolutions to idle balancers
  * Reclassified gRPC status codes as non-errors
* Internal
  * Updated helm charts to follow best practices using proper casing (thanks 
    @Pothulapati!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,23 @@
 ## edge-19.12.3
 
+This edge release adds support for pod IP and service cluster IP lookups, 
+improves performance of the dashboard, and makes `linkerd check --pre` perform 
+more comprehensive checks.
+
+The `--wait-before-seconds` flag has been added to allow Linkerd users to opt in
+to `preStop hooks`. The details of this change are in 
+[#3798](https://github.com/linkerd/linkerd2/pull/3798).
+
+Also, the proxy has been updated to `v2.82.0` which improves gRPC error
+classification and 
+[ensures that resolutions](https://github.com/linkerd/linkerd2/pull/3848) are 
+released when the associated balancer becomes idle.
+
+Finally, an update to follow best practices in the Helm charts has caused a
+*breaking change*. Users who have installed Linkerd using Helm must be
+certain to read the details of 
+[#3822](https://github.com/linkerd/linkerd2/issues/3822)
+
 * CLI
   * Increased the comprehensiveness of `linkerd check --pre`
   * Added TLS certificate validation to `check` and `upgrade` commands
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,26 @@
...
+* Controller
+  * Increased minimum kubernetes version to 1.13.0
+  * Added support for pod ip and service cluster ip lookups in the destination 
+    service
+  * Added recommended kubernetes labels to control-plane
+  * Added preStop hook into the proxy sidecar container to stop it last (a huge 
+    commit from @KIVagant, thanks!)
+  * Added a pre-sign check to te identity service
+* Web UI
+  * Added logic to pause dashboard polling activity when the page is not in
```
scottcarol: I would rephrase as: "Increased the speed of the dashboard by pausing network activity when the dashboard is not visible to the user"

CHANGES.md
```diff
@@ -1,3 +1,26 @@
...
+  * Increased the comprehensiveness of `linkerd check --pre`
+  * Added TLS certificate validation to `check` and `upgrade` commands
+* Controller
+  * Increased minimum kubernetes version to 1.13.0
+  * Added support for pod ip and service cluster ip lookups in the destination 
+    service
+  * Added recommended kubernetes labels to control-plane
+  * Added preStop hook into the proxy sidecar container to stop it last (a huge 
+    commit from @KIVagant, thanks!)
+  * Added a pre-sign check to te identity service
```
scottcarol: `the` identity service

CHANGES.md
```diff
@@ -1,3 +1,26 @@
...
+* CLI
+  * Increased the comprehensiveness of `linkerd check --pre`
+  * Added TLS certificate validation to `check` and `upgrade` commands
+* Controller
+  * Increased minimum kubernetes version to 1.13.0
+  * Added support for pod ip and service cluster ip lookups in the destination 
+    service
+  * Added recommended kubernetes labels to control-plane
+  * Added preStop hook into the proxy sidecar container to stop it last (a huge 
+    commit from @KIVagant, thanks!)
```
alpeb: That's opt-in, so I would be more precise and say something like "Added the `--wait-before-exit-seconds` flag to `linkerd inject` for the proxy sidecar to delay the start of its shutdown process". Not sure if we'd like to also add more context on the cases where adding that delay makes sense.

CHANGES.md
```diff
@@ -1,3 +1,26 @@
...
+  * Added a pre-sign check to te identity service
+* Web UI
+  * Added logic to pause dashboard polling activity when the page is not in 
+    focus
+* Proxy
+  * Added a timeout to release resolutions to idle balancers
+  * Reclassified gRPC status codes as non-errors
```
olix0r: Maybe _Improved error classification for gRPC services_?

CHANGES.md
```diff
@@ -1,3 +1,26 @@
...
+* Internal
+  * Updated helm charts to follow best practices using proper casing (thanks 
+    @Pothulapati!)
```
alpeb: This is an important BC breaking change that people should be aware of, so it can go under its own section `Helm` like we did back in the `edge-19.10.3` notes. Also `Helm` should be capitalized ;-)

CHANGES.md
alpeb: Maybe also worth adding a note about the performance improvements for concurrent dashboard usage? (#3769)

CHANGES.md
```diff
@@ -1,5 +1,23 @@
 ## edge-19.12.3
 
+This edge release adds support for pod IP and service cluster IP lookups, 
+improves performance of the dashboard, and makes `linkerd check --pre` perform 
+more comprehensive checks.
+
+The `--wait-before-seconds` flag has been added to allow Linkerd users to opt in
```
alpeb: should be `--wait-before-exit-seconds`

PR #3865, opened 2019-12-23
Upgraded `Helm` cli to v2.16.1

alpeb:
Needed for k8s 1.16

### DIFF
--- a/bin/helm
+++ b/bin/helm
@@ -2,7 +2,7 @@
 
 set -eu
 
-helmversion=v2.14.3
+helmversion=v2.16.1
 bindir=$( cd "${0%/*}" && pwd )
 targetbin=$( cd "$bindir"/.. && pwd )/target/bin
 helmbin=$targetbin/helm-$helmversion


grampelberg:
Should we go to v3 now?

cpretzer:
Can we support both? My guess is that some long-time helm users may not be able to migrate to v3 very quickly, and the option to use v2 or v3 would help them.

grampelberg:
> Can we support both?

Yes, but the details matter. I'd prefer to support one or the other purely from an effort perspective. We could hold off officially supporting v3 until users start running into problems. I just wanted to make sure we were thinking about it.

cpretzer:
Makes sense to me!

alpeb:
We do support both from the user's perspective. This change just determines the version against which the Helm integration tests are run, which is set in `/bin/helm`'s `helmversion` var. It would require some other changes for the tests to run against v3.
I could make those changes, to apply conditionally depending on the value of `helmversion`, so we can still run the tests against both versions.

cpretzer:
Thanks for the explanation @alpeb

Issue #3739, opened 2019-11-18
Validate incoming sidecar injection requests


scottcarol:
During the Linkerd workshop, a user was updating the `booksapp` namespace to enable automatic proxy injection:
`kubectl annotate ns booksapp linkerd.io/inject=enabled`
but accidentally typed `enable` instead.

We should notify users if incoming sidecar injection requests to `linkerd.io/inject` do not have the correct `enabled` or `disabled` value.

zaharidichev:
That sounds like a good improvement. What would the form of this notification take. Would that be just logging or emitting k8s events. Or maybe something on the website?

tegioz:
What about using a validation admission webhook? We would reject the annotation when using the key `linkerd.io/inject` but not using a valid option (`enabled`/`disabled`), letting the user know what happened.

ihcsim:
Sounds like that really [cool talk I attended at KubeCon](https://youtu.be/gMaGVHnvNfs), about Linkerd and policies :wink:.

grampelberg:
We should add this as an event to the existing stream from the proxy injector. While it'd be awesome to get a validating webhook going, I'd love to see the small incremental improvement of having something like "invalid value detected" in describe.

mayankshah1607:
Hi! I would like to take up this issue if no one's already working on it :)

grampelberg:
@mayankshah1607 sounds good! Why don't you get a draft PR up first to make sure your approach makes sense before polishing things off?

mayankshah1607:
@grampelberg I have opened a draft PR - https://github.com/linkerd/linkerd2/pull/3868

PR #3868, opened 2019-12-25
Validate incoming sidecar injection requests

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/3739

- Add condition to check if inject annotation has an invalid value
- Add warning log message and event to existing stream for proxy-injector

### DIFF
--- a/controller/proxy-injector/webhook.go
+++ b/controller/proxy-injector/webhook.go
@@ -35,7 +35,6 @@ func Inject(api *k8s.API,
 	if err != nil {
 		return nil, err
 	}
-
 	proxyConfig, err := config.Proxy(pkgK8s.MountPathProxyConfig)
 	if err != nil {
 		return nil, err
@@ -78,6 +77,23 @@ func Inject(api *k8s.API,
 		ownerKind = strings.ToLower(ownerRef.Kind)
 	}
 	proxyInjectionAdmissionRequests.With(admissionRequestLabels(ownerKind, request.Namespace, report.InjectAnnotationAt, configLabels)).Inc()
+
+	//Check if Namespace has valid inject annotation
+	if isNsAnnotationValid, val := isInjectionValid(nsAnnotations); !isNsAnnotationValid {
+		log.Warnf("invalid value \"%s\" for annotation \"%s\" detected at namespace level", val, pkgK8s.ProxyInjectAnnotation)
+		if parent != nil {
+			recorder.Eventf(*parent, v1.EventTypeWarning, eventTypeSkipped, "Invalid value \"%s\" for annotation \"%s\" detected at namespace level", val, pkgK8s.ProxyInjectAnnotation)
+		}
+	}
+
+	//Check if Pod-template has valid inject annotation
+	podAnnotations := resourceConfig.GetPodAnnotations()
+	if isPodAnnotationValid, val := isInjectionValid(podAnnotations); !isPodAnnotationValid {
+		log.Warnf("invalid value \"%s\" for annotation \"%s\" detected at pod-template %s/%s", val, pkgK8s.ProxyInjectAnnotation, request.Namespace, report.ResName())
+		if parent != nil {
+			recorder.Eventf(*parent, v1.EventTypeWarning, eventTypeSkipped, "Invalid value \"%s\" for annotation \"%s\" detected at pod-template", val, pkgK8s.ProxyInjectAnnotation)
+		}
+	}
 
 	if injectable, reasons := report.Injectable(); !injectable {
 		var readableReasons, metricReasons string
@@ -128,4 +144,14 @@ func ownerRetriever(api *k8s.API, ns string) inject.OwnerRetrieverFunc {
 		p.SetNamespace(ns)
 		return api.GetOwnerKindAndName(p, true)
 	}
+}
+
+//isInjectionValid method checks if inject annotation has a valid value - enabled/disabled
+func isInjectionValid(annotations map[string]string) (bool, string) {
+	for k, v := range annotations {
+		if k == pkgK8s.ProxyInjectAnnotation && !(v == pkgK8s.ProxyInjectEnabled || v == pkgK8s.ProxyInjectDisabled) {
+			return false, v
+		}
+	}
+	return true, ""
 }
--- a/controller/proxy-injector/fake/data/deployment-emojivoto-invalid-inject-annotation.yaml
+++ b/controller/proxy-injector/fake/data/deployment-emojivoto-invalid-inject-annotation.yaml
@@ -1,0 +1,30 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+    name: emoji
+    namespace: emojivoto
+spec:
+    replicas: 1
+    selector:
+        matchLabels:
+            app: emoji-svc
+    template:
+        metadata:
+            labels:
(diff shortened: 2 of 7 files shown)


controller/proxy-injector/webhook.go
```diff
@@ -79,6 +78,13 @@ func Inject(api *k8s.API,
 	}
 	proxyInjectionAdmissionRequests.With(admissionRequestLabels(ownerKind, request.Namespace, report.InjectAnnotationAt, configLabels)).Inc()
 
+	for k, v := range nsAnnotations {
+		if k == pkgK8s.ProxyInjectAnnotation && v != pkgK8s.ProxyInjectEnabled {
```
mayankshah1607: Would it make sense to add a condition to allow `disabled` as well?

grampelberg:
@Pothulapati mind taking a look?

controller/proxy-injector/webhook.go
```diff
@@ -79,6 +78,13 @@ func Inject(api *k8s.API,
 	}
 	proxyInjectionAdmissionRequests.With(admissionRequestLabels(ownerKind, request.Namespace, report.InjectAnnotationAt, configLabels)).Inc()
 
+	for k, v := range nsAnnotations {
```
Pothulapati: This approach checks only for ns annotations. We should check for both NS and Workload Annotations.

controller/proxy-injector/webhook.go
```diff
@@ -79,6 +78,13 @@ func Inject(api *k8s.API,
...
+		if k == pkgK8s.ProxyInjectAnnotation && v != pkgK8s.ProxyInjectEnabled {
+			log.Warnf("invalid value detected for annotation %s : %s", k, v)
+			recorder.Eventf(*parent, v1.EventTypeNormal, eventTypeSkipped, "Invalid value detected for annotation %s: %s", k, v)
```
alpeb: Please note that `parent` might be null here, so you also need to check for that.

mayankshah1607:
@Pothulapati @alpeb done! Would you mind taking a look again?

controller/proxy-injector/webhook.go
```diff
@@ -79,6 +78,23 @@ func Inject(api *k8s.API,
...
+		log.Warnf("invalid proxy injection value detected at namespace %s - \"%s\"", request.Namespace, val)
+		if parent != nil {
+			recorder.Eventf(*parent, v1.EventTypeWarning, eventTypeSkipped, "Invalid proxy injection detected at namespace : %s:%s", pkgK8s.ProxyInjectAnnotation, val)
+		}
+	}
+
+	//Check if Pod-template has valid inject annotation
+	podAnnotations := resourceConfig.GetPodAnnotations()
+	if isPodAnnotationValid, val := isInjectionValid(podAnnotations); !isPodAnnotationValid {
+		log.Warnf("invalid proxy injection value detected at pod-template %s - \"%s\"", report.ResName(), val)
```
Pothulapati: Should we also log the workload type and name. i.e `invalid injection annotation value detected at namespace/pod/name`? 
That information would be useful.

mayankshah1607: @Pothulapati with this approach, the workload type and name is being logged already because of `report.ResName` - eg; `invalid proxy injection value detected at pod-template  pod/pod-name"`. We could go ahead and prepend the namespace as well

Pothulapati: That's awesome.
It would be cool, to have the log message be in the same format at ns or workload level then like `invalid proxy injection annotation value detected at <type>/<name>"`

controller/proxy-injector/webhook.go
```diff
@@ -79,6 +78,23 @@ func Inject(api *k8s.API,
...
+		if parent != nil {
+			recorder.Eventf(*parent, v1.EventTypeWarning, eventTypeSkipped, "Invalid proxy injection value \"%s\" detected at namespace %s", val, request.Namespace)
+		}
+	}
+
+	//Check if Pod-template has valid inject annotation
+	podAnnotations := resourceConfig.GetPodAnnotations()
+	if isPodAnnotationValid, val := isInjectionValid(podAnnotations); !isPodAnnotationValid {
+		log.Warnf("invalid proxy injection value \"%s\" detected at pod-template for %s/%s", val, request.Namespace, report.ResName())
+		if parent != nil {
```
mayankshah1607: @Pothulapati WDYT now??

controller/proxy-injector/webhook.go
```diff
@@ -79,6 +78,23 @@ func Inject(api *k8s.API,
 	}
 	proxyInjectionAdmissionRequests.With(admissionRequestLabels(ownerKind, request.Namespace, report.InjectAnnotationAt, configLabels)).Inc()
 
+	//Check if Namespace has valid inject annotation
+	if isNsAnnotationValid, val := isInjectionValid(nsAnnotations); !isNsAnnotationValid {
+		log.Warnf("invalid proxy injection value \"%s\" detected at namespace \"%s\" for %s", val, request.Namespace, report.ResName())
+		if parent != nil {
+			recorder.Eventf(*parent, v1.EventTypeWarning, eventTypeSkipped, "Invalid proxy injection value \"%s\" detected at namespace %s", val, request.Namespace)
```
alpeb: This event will tied to the workload whose namespace is known to the user, so no need to explicit it here; you can just say something like "at the namespace level".
Also, I wouldn't call this `injection value` but instead `Invalid value \"%s\" for annotation \"%s\`

mayankshah1607: @alpeb yes, that makes sense! will make the required changes

mayankshah1607:
@alpeb done! Mind taking a look now?

mayankshah1607:
@alpeb oops, did not notice that one. I've fixed it. Hope everything looks good now :smile: 
Thanks for the :heavy_check_mark: !

adleong:
If I'm reading the code correctly, the previous behavior in this situation is that an event would be recorded with the text "neither the namespace nor the pod have the annotation ...".  With this change, that event text is changed to "invalid value ... for annotation ... detected at ..." and a warning message is added to the proxy-injector logs.  Do I have that right?

I'm wondering if we can isolate this change to the `disableByAnnotation` function in report.go.  This is where the logic for checking the injection annotation already lives, so it would be nice to keep all this logic in one place.  It should just be a matter of checking for an invalid injection annotation in that method and returning the new inject disabled reason.  Does this make sense?

I don't have a lot of experience in this part of the codebase, so please correct me if I've misunderstood anything.

mayankshah1607:
@adleong I think that makes sense. We could prepend this check to method `disableByAnnotation` itself, instead of introducing redundant code in `webhook.go`. Maybe I misunderstood when @grampelberg said we should have a [validating webhook going](https://github.com/linkerd/linkerd2/issues/3739#issuecomment-557602169). 
@alpeb @cpretzer @Pothulapati WDYT?

mayankshah1607:
@adleong I've opened a PR with an alternate approach suggested by you. Mind taking a look? - https://github.com/linkerd/linkerd2/pull/3909
We could close this one if the above PR seems okay! :)
@alpeb @cpretzer

adleong:
Closing in favor of https://github.com/linkerd/linkerd2/pull/3909

PR #3872, opened 2019-12-30
Fix Helm install when disabling init containers

alpeb:
To install linkerd using Helm after having installed linkerd's CNI plugin, one needs to `--set noInitContainer=true`.
But to determine whether to use init containers or not, we weren't
evaluating that, but instead `Values.proxyInit`, which is indeed null
when installing through the CLI but not when installing with Helm. So
init containers were being set despite having passed `--set
noInitContainers=true`.


alpeb:
@Pothulapati I've just merged this against master, after your #3839 got merged. This makes use of `.Values.noInitContainer` which isn't declared in `values.yaml`. I guess now we have to move that into `.Values.global`, right?

ihcsim:
To confim (for testing purposes), this actually affects `linkerd inject`, not `helm install`, right?

ihcsim:
I thought the control plane templates don't use `patch.json`.

Pothulapati:
@alpeb Yep, Makes Sense. 
I can raise a PR to move `noInitContainer` to `globals`. WDYT?

zaharidichev:
@alpeb  I have the same question as @ihcsim It looks to me that the control plane components are not affected by the patch (look [here](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/identity.yaml#L117)). I guess they can be if you do something like an inject --manual to bring in a debug container into a control plane pod. But upon install they should not be, right ?

alpeb:
Ok I stand corrected, the issue was not about control plane injection. 
This is how installing the linkerd2 chart fails currently in the latest edge:

```bash
# this part works
$ helm install charts/linkerd2-cni
 ...
# wait for the cni pod to come up, and then
$ helm install -f linkerd-overrides.yaml --set noInitContainer=true --set installNamespace=false charts/linkerd2                                                                                                                  
Error: render error in "linkerd2/templates/web.yaml": template: linkerd2/templates/web.yaml:98:12: executing "linkerd2/templates/web.yaml" at <include "partials.proxy" .>: error calling include: template: linkerd2/charts/partials/templates/_proxy.tpl:4:18: executing "partials.proxy" at <.Values.global.proxy.logLevel>: nil pointer evaluating interface {}.proxy
```
So checking for `.Values.noInitContainer` should avoid that problem.

alpeb:
Actually the previous error was caused by a stale `/charts/linkerd2/charts/partials-0.1.0.tgz`. I'll post back proper repro steps in a moment.

alpeb:
@ihcsim @zaharidichev thanks for the comments; they made me realize the approach here ended up being wrong. I'm closing this PR in favor of #3894

PR #3864, opened 2019-12-23
Upgrade `kind` to v0.6.1

alpeb:
Fixes #3852

Upgraded `/bin/kind` to pull v0.6.1.
Also have `workflow.yml` use `KUBECONFIG` explicitly for setting the
location of the config file, now that `kind get kubeconfig-path` has
been deprecated (check
https://github.com/kubernetes-sigs/kind/releases/tag/v0.6.0 for detailed
info).
Note that in the build server the kind binary for this version is
`kind-0.6.1`, leaving the `kind` binary still pointing to v0.5.1 while
this gets merged and all the PR branches get this.

PR #3875, opened 2019-12-30
Use `kind export kubeconfig` instead of `scp`

alpeb:
Followup to #3864 [comment](https://github.com/linkerd/linkerd2/pull/3864#discussion_r360976473)

Stop moving the kubeconfig file between the Github Action env and the
build server with `scp` and use instead `kind export kubeconfig`.


.github/workflows/workflow.yml
```diff
@@ -393,7 +392,7 @@ jobs:
         # Restore kubeconfig from remote docker host.
```
adleong: should this comment say that we're getting the kubeconfig from kind?

alpeb: From what I understand, `kind export kubeconfig` retrieves the config from the remote host, so the comment is accurate

Issue #3878, opened 2019-12-31
Registry override alone is ineffectual for proxyImage and proxyInitImage during inject


javaducky:
### What is the issue?
Use of the `--registry` override flag without the corresponding `--proxy-image` or `--init-image` flag has no effect on the injected `proxyImage` nor `proxyInitImage`.  Specifying a `--proxy-image` or `--init-image` override value makes the `--registry` flag unnecessary.

### How can it be reproduced?
This scenario assumes that the _Emojivoto_ demo application is installed and that _Linkerd_ has been installed without a custom registry.

```
kubectl -n emojivoto get deploy/voting -o yaml | bin/linkerd inject --registry=my.custom.registry/linkerd-io - | kubectl apply -f -
```
Inspection of the updated pod will show that the injected `proxy` and `proxy-init` images are still from the base install, i.e. `gcr.io/linkerd-io` and not the override registry of `my.custom.registry/linkerd-io`:
```
kubectl get pods -n emojivoto -l app=voting-svc -o jsonpath="{.items[*].spec..image}" | \
tr -s '[[:space:]]' '\n' | sort | uniq
```
Performing the same operation specifying the image override(s) will inject the desired image:
```
kubectl -n emojivoto get deploy/voting -o yaml | bin/linkerd inject --registry=my.custom.registry/linkerd-io --proxy-image=my.custom.registry/linkerd-io/proxy - | kubectl apply -f -

kubectl -n emojivoto get deploy/voting -o yaml | bin/linkerd inject --registry=my.custom.registry/linkerd-io --proxy-image=my.custom.registry/linkerd-io/proxy --init-image=my.custom.registry/linkerd-io/proxy-init - | kubectl apply -f -
```
Specifying the override registry with the default images seems to provide the desired result of images having the substituted registry:
```
kubectl -n emojivoto get deploy/voting -o yaml | bin/linkerd inject --registry=my.custom.registry/linkerd-io --proxy-image=gcr.io/linkerd-io/proxy --init-image=gcr.io/linkerd-io/proxy-init - | kubectl apply -f -
```

### Possible solution
The `registry` override option should automatically substitute either the default registry (gcr.io/linkerd-io) or whatever registry override was specified during install (based on the `data.install` flags) for only those images not being explicitly specified with the `--init-image` or `--proxy-image` flags.

PR #3879, opened 2019-12-31
Respect registry override during inject

adleong:
Fixes https://github.com/linkerd/linkerd2/issues/3878

If the `--registry` flag is provided to Linkerd without the `--proxy-image` or `--init-image` flags, the `--registry` flag is ignored and not applied to the existing values for the proxy or init images pulled from the configmap.

We now override the registry with the value from the `--registry` flag regardless of which other flags are provided.

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -368,17 +368,19 @@ func (options *proxyConfigOptions) overrideConfigs(configs *cfg.All, overrideAnn
 
 	if options.proxyImage != "" {
 		configs.Proxy.ProxyImage.ImageName = options.proxyImage
-		if options.dockerRegistry != "" {
-			configs.Proxy.ProxyImage.ImageName = registryOverride(configs.Proxy.ProxyImage.ImageName, options.dockerRegistry)
-		}
 		overrideAnnotations[k8s.ProxyImageAnnotation] = configs.Proxy.ProxyImage.ImageName
 	}
 
 	if options.initImage != "" {
 		configs.Proxy.ProxyInitImage.ImageName = options.initImage
-		if options.dockerRegistry != "" {
-			configs.Proxy.ProxyInitImage.ImageName = registryOverride(configs.Proxy.ProxyInitImage.ImageName, options.dockerRegistry)
-		}
+
+		overrideAnnotations[k8s.ProxyInitImageAnnotation] = configs.Proxy.ProxyInitImage.ImageName
+	}
+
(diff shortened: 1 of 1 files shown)


cli/cmd/inject.go
```diff
@@ -368,17 +368,19 @@ func (options *proxyConfigOptions) overrideConfigs(configs *cfg.All, overrideAnn
...
 	if options.initImage != "" {
 		configs.Proxy.ProxyInitImage.ImageName = options.initImage
-		if options.dockerRegistry != "" {
-			configs.Proxy.ProxyInitImage.ImageName = registryOverride(configs.Proxy.ProxyInitImage.ImageName, options.dockerRegistry)
-		}
+
+		overrideAnnotations[k8s.ProxyInitImageAnnotation] = configs.Proxy.ProxyInitImage.ImageName
+	}
+
+	if options.dockerRegistry != "" {
```
javaducky: Personally, I feel that the `registry` override is the more "general" override and should therefore be applied before the more specific `proxy-image` or `init-image` flags.  That way, the more specific can override the general.  What do you think?

adleong: The registry should only change the registry part of the full image name, leaving the `proxy-image` or `init-image` values unchanged.  In other words, it could override `gcr.io/foobar` to `my-registry.io/foobar`.  So I don't believe that it needs to be applied before the image overrides.

javaducky: Sorry if I'm really stretching my scenario into something that noone would normally do, but my concern is the following: 
```
linkerd inject --enable-debug-sidecar --registry=my.custom.registry/linkerd-io --init-image=gcr.io/linkerd-io/proxy-init
```
In this case, I my intended result would be for the following to be installed:
```
my.custom.registry/linkerd-io/proxy
my.custom.registry/linkerd-io/debug
gcr.io/linkerd-io/proxy-init
```
but the actual would result in:
```
my.custom.registry/linkerd-io/proxy
my.custom.registry/linkerd-io/debug
my.custom.registry/linkerd-io/proxy-init
```
So in reality, the only way I'd truly be able to achieve my intended setup would be to manually update the override annotations in the post-injection deployment.

Again...I'm probably going overboard with mixing registries, but I could see the possibility.  As an FYI, the registry override for debug sidecar isn't working (PR #3873) addresses that particular issue.

adleong: Ah, I see what you mean now.  Updated!

PR #3882, opened 2020-01-03
Adopt more common Javascript linting conventions

klingerf:
The current set of Javascript linting rules that we're using in this
project is outdated, and it has lead to a variety of competing styles
in the Javascript codebase.

Update the project's linting rules to match those provided by the latest
release of eslint-config-airbnb, but disable a bunch of rules that
aren't compatible with this project.

I've split this change into two commits. The first commit (7446f065) contains
the manual changes that I made to satisfy the new rules, and the second
commit (596fff42) contains all of the whitespace, quoting and commas
changes that were fixed automatically by eslint.

Signed-off-by: Kevin Lingerfelt <kl@buoyant.io>


web/app/js/components/NetworkGraph.jsx
```diff
@@ -101,72 +101,76 @@ export class NetworkGraphBase extends React.Component {
...
-        .attr("d", node =>  "M" +
-              node.source.x + " " +
-              node.source.y + " L " +
-              node.target.x + " " +
-              node.target.y);
+        .attr('d', node => `M${
+          node.source.x} ${
+          node.source.y} L ${
+          node.target.x} ${
+          node.target.y}`);
```
alpeb: Looks odd, but from my testing, it's fine...

web/app/js/components/ResourceDetail.jsx
```diff
@@ -420,7 +420,7 @@ export class ResourceDetailBase extends React.Component {
         <MetricsTable
           resource="pod"
           title="TCP"
-          isTcpTable={true}
+          isTcpTable
```
alpeb: Will this produce proper XHML?

web/app/js/components/TopEventTable.jsx
```diff
@@ -11,84 +11,84 @@ import { withContext } from './util/AppContext.jsx';
...
 ];
 
 const TopEventTable = ({ tableRows, resourceType, api }) => {
-  let columns = topColumns(resourceType, api.ResourceLink, api.PrefixedLink);
+  const columns = topColumns(resourceType, api.ResourceLink, api.PrefixedLink);
 
   return (
     <BaseTable
-      enableFilter={true}
+      enableFilter
```
alpeb: same XHTML comment as above

web/app/js/components/TopRoutes.jsx
```diff
@@ -68,19 +68,19 @@ class TopRoutes extends React.Component {
...
       to_name: '',
       to_type: '',
-      to_namespace: ''
+      to_namespace: '',
     }, _pick(props.query, Object.keys(topRoutesQueryProps)));
 
 
     this.state = {
-      query: query,
+      query,
```
alpeb: How come there's no `key: value` here?

web/app/js/components/TopRoutes.jsx
```diff
@@ -232,16 +232,16 @@ class TopRoutes extends React.Component {
...
 
             <Grid item>
-              { this.renderResourceDropdown("To Resource", "to_name", "to_type", "Target resource") }
+              { this.renderResourceDropdown('To Resource', 'to_name', 'to_type', 'Target resource') }
             </Grid>
           </Grid>
         </Grid>
         <Divider light className={classes.root} />
-        <Typography variant="caption">You can also create a new profile <ConfigureProfilesMsg showAsIcon={true} /></Typography>
+        <Typography variant="caption">You can also create a new profile <ConfigureProfilesMsg showAsIcon /></Typography>
```
alpeb: Same comment about attributes without a value

web/app/js/components/TopRoutesTable.jsx
```diff
@@ -6,58 +6,58 @@ import { metricToFormatter } from './util/Utils.js';
...
-  }
+    render: d => metricToFormatter.LATENCY(d.latency.P99),
+    sorter: d => d.latency.P99,
+  },
 ];
 
 const TopRoutesTable = ({ rows }) => (
   <BaseTable
-    enableFilter={true}
+    enableFilter
```
alpeb: ditto

web/app/js/components/TopRoutesTabs.jsx
```diff
@@ -33,22 +33,22 @@ class TopRoutesTabs extends React.Component {
...
+      namespace: query.namespace,
     };
 
     return (
       <React.Fragment>
         <TopModule
           pathPrefix={pathPrefix}
           query={topQuery}
-          startTap={true}
+          startTap
```
alpeb: :point_up:

web/app/js/components/util/withREST.jsx
```diff
@@ -96,7 +96,7 @@ const withREST = (WrappedComponent, componentPromises, options={}) => {
       Promise.all(this.api.getCurrentPromises())
         .then(data => {
           this.setState({
-            data: data,
+            data,
```
alpeb: Same comment about `key: value`

alpeb:
Btw, my review above was about the second commit resulting from the automatic fixes. I started looking at the first commit, but realized I don't have enough React chops to comment on it :disappointed:

web/app/js/components/NetworkGraph.jsx
klingerf: Hmm, yeah, I agree that that looks pretty funky. I'll collapse this all onto one line, now that it's using string interpolation.

web/app/js/components/ResourceDetail.jsx
klingerf: Yep, this was caught by the [react/jsx-boolean-value](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) rule. When setting a boolean prop to `true` in react, there's no need to explicitly pass `true` as the prop value, so this rule removes it.

web/app/js/components/TopRoutes.jsx
klingerf: This was caught by the [object-shorthand](https://eslint.org/docs/rules/object-shorthand) rule. ES6 allows a shorthand that uses the variable name as the property name if it's omitted, and this rules makes sure that's used consistently throughout.

web/app/js/components/util/TapUtils.jsx
```diff
@@ -343,9 +342,7 @@ export const srcDstColumn = (d, resourceType, ResourceLink) => {
   };
 
   let baseContent = (
-    <Link to="#" onClick={linkFn}>
-      <OpenInNewIcon fontSize="small" />
-    </Link>
+    <OpenInNewIcon fontSize="small" style={{color: "var(--linkblue)"}} onClick={linkFn} />
```
scottcarol: To confirm, you're hardcoding the style here in order not to use the `Link` component? Could we use `withStyles` instead?

web/app/js/components/util/TapUtils.jsx
```diff
@@ -386,10 +383,10 @@ export const tapLink = (d, resourceType, PrefixedLink) => {
   let toResource = "";
 
   if (_has(d.destinationLabels, resourceType)) {
-    toNamespace = d.destinationLabels.namespace,
```
scottcarol: great catch

web/app/.eslintrc
```diff
@@ -55,13 +42,47 @@
...
     "semi": [2, "always"],
     "sort-imports": [2],
     "space-before-blocks": [2, "always"],
     "space-before-function-paren": [2, "never"],
-    "strict": [2, "safe"]
+    "strict": [2, "safe"],
+
+    // disable airbnb defaults that don't work for our project
+    // TODO: fix the code and remove these overrides
+    "arrow-body-style": 0,
```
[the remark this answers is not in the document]
klingerf: Ah, yes, that's a good question. The disabled rules fall into a few different categories.

There are a few rules that just won't work in our codebase. For instance:

- no-bitwise -- I guess some codebases don't need bitwise operators (and they can be confused for logical operators), but our codebase does need them for ip translation
- no-tabs -- Some of our SVGs have tabs in them
- react/jsx-props-no-spreading -- Some of our components wrap Material UI components and use spread to forward props

And then there are a few that are just stylistic preference:

- no-else-return -- This rule doesn't allow an else clause following an if clause that calls `return`; seems weird
- no-nested-ternary - We use nested ternaries all over the place
- object-curly-newline / react/jsx-curly-newline - These rules are auto-fixable but you end up with some pretty strange indentation

And there are a couple that would be great to fix, but would require some substantial rewriting of some parts of the code base (these would be great candidates for follow-up pull requests):

- no-param-reassign -- There are bunch of places where we're reassigning the value of params passed into functions
- react/no-did-update-set-state -- We're calling setState inside componentDidUpdate, and that seems bad

And finally there are a few rules that would be easy to fix but would have made this a much noisier diff (these would also be great candidates for follow-up pull requests):

- no-use-before-define -- We need to rearrange our functions so that the functions that they call are defined first.
- react/sort-comp -- We need to rearrange our components so that the lifecycle methods appear at the top.

web/app/js/components/NetworkGraph.jsx
```diff
@@ -105,7 +93,7 @@ export class NetworkGraphBase extends React.Component {
     let graphData = this.getGraphData();
 
     // check if graph is present to prevent drawing of multiple graphs
-    if (this.svg.select("circle")._groups[0][0]) {
+    if (this.svg.select("circle")._groups[0][0]) { // eslint-disable-line
```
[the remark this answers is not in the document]
klingerf: Ahhh, this one is extraneous, actually, since I eventually decided to disable the `no-underscore-dangle` rule globally, since we rely on _ prefixes and suffixes in a few places. Will remove the comment here.

web/app/js/components/NetworkGraph.jsx
```diff
@@ -42,20 +39,11 @@ const simulation = d3.forceSimulation()
...
-    data: PropTypes.arrayOf(metricsPropType.isRequired).isRequired,
-    deployments: PropTypes.arrayOf(PropTypes.object),
-  }
-
   constructor(props) {
     super(props);
 
     // https://github.com/d3/d3-zoom/issues/32
-    d3.getEvent = (() => require("d3-selection").event).bind(this);
+    d3.getEvent = (() => require("d3-selection").event); // eslint-disable-line
```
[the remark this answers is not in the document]
klingerf: Great call! In fact I should update this to only disable the specific rule (global-require) that needs to be disabled, and that will better explain why we're skipping linting.

web/app/js/components/ResourceDetail.jsx
```diff
@@ -99,16 +87,18 @@ export class ResourceDetailBase extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (!_isEqual(prevProps.match.url, this.props.match.url)) {
+    let { match, pathPrefix, isPageVisible } = this.props;
+
+    if (!_isEqual(prevProps.match.url, match.url)) {
```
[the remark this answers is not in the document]
klingerf: `match` is labeled as a required prop for this component, so I think it's ok if this code assumes that it's present. Granted, proptype assertions aren't really binding, but I think it's safe enough to assume that required props are present.

web/app/js/components/util/TapUtils.jsx
klingerf: Yeah, this was a pretty big hack. The previous code didn't lint because it didn't pass accessibility checks. We shouldn't have been using an empty link to trigger an onclick event. We could definitely use `withStyles`, and that's how I'd fix it in a component, but this code appears in a function that returns raw JSX. We should instead convert the whole function to a component, but that felt like a bigger change. If it's ok by you I'm inclined to leave this as is.

klingerf:
@scottcarol @alpeb Thanks for all of the great feedback! I agree that it would be good to lint tests, but I'm inclined for us to tackle that in a separate branch. Right now our CI setup doesn't enforce that our test files lint cleanly. If we want to start linting our test files, we should also ensure that tests continue to lint cleanly by enforcing that in CI.

web/app/.eslintrc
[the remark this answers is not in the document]
scottcarol: Thanks for the breakdown - after you merge this, I'm happy to file a (non-urgent) issue for this comment, unless you'd prefer to do it yourself.

web/app/js/components/ResourceDetail.jsx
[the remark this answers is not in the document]
scottcarol: Fine by me.

web/app/js/components/util/TapUtils.jsx
scottcarol: Thanks for the explanation - fine by me.

scottcarol:
👍 Tackling the test file linting in a separate branch sounds good to me! This is good to merge on my end.

web/app/.eslintrc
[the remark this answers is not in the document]
klingerf: You're totally right, we should turn this into actionable issues. Prior to merging this branch, I'll reorganize this list into will fix / won't fix, and then I can open issues about the rules in the will fix section, as well as enabling linting for test files.

PR #3884, opened 2020-01-03
Add event regex to ignore in integration test

alpeb:
We were ignoring events like
```
MountVolume.SetUp failed for volume .* : couldn't propagate object cache: timed out waiting for the condition
```

but as k8s 1.16 those got replaced by more precise messages, like
```
MountVolume.SetUp failed for volume "linkerd-identity-token-cm4fn" :failed to sync secret cache: timed out waiting for the condition
MountVolume.SetUp failed for volume "prometheus-config" : failed to sync configmap cache: timed out waiting for the condition
```

This was causing sporadic CI test failures like [here](https://github.com/linkerd/linkerd2/runs/368424822#step:7:562)

So I'm including another regex for that.

Re: https://github.com/kubernetes/kubernetes/commit/96c41f8a1e1f3bee819b3e01a3fa1a89f3f55f66

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -84,7 +84,8 @@ var (
 	}, "|"))
 
 	knownEventWarningsRegex = regexp.MustCompile(strings.Join([]string{
-		`MountVolume.SetUp failed for volume .* : couldn't propagate object cache: timed out waiting for the condition`,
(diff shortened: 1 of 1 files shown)


alpeb:
@ihcsim Agreed, regexes are brittle, but does the trick for now.

Issue #2962, opened 2019-06-18
upgrade and install templates have empty strategy/status fields


cppforlife:
## Feature Request

Remove `status: {}` and `spec.strategy: {}` fields from Deployment configurations produced by install/upgrade commands.

### What problem are you trying to solve?

I am using kapp (from https://k14s.io/; I'm one of the authors for these tools) to deploy produced YAML configuration from install/upgrade commands. It looks something like this: `linkerd upgrade | kapp deploy -a linkerd -f- -c -y`. This works brilliantly with one exception. Currently both install/upgrade commands produce empty `status` and `strategy` fields, which makes kapp consider Deployments as changed since on the server side these fields were populated (example output: https://gist.github.com/cppforlife/59bce65ed38dcfa6612b661707f50f85).

To bypass this annoyance I am currently using ytt (also from k14s) to remove those fields before piping output into kapp, but ideally this would not be necessary.

```bash
linkerd upgrade | ytt -f-  -f /tmp/linkerd-patches.yml --ignore-unknown-comments=true | kapp deploy -a linkerd -f- -c -y
```

```yaml
#@ load("@ytt:overlay", "overlay")

#@overlay/match by=overlay.subset({"kind":"Deployment"}),expects="1+"
---
spec:
  #@overlay/remove
  strategy: null
#@overlay/remove
status: null
```

When empty fields are removed diff output is a lot better. Example output from 2.3.1 to 2.3.2 update: https://gist.github.com/cppforlife/114e41f964a1ddd821a38d0c08f0f508

Is there a reason why these fields are present in the configuration output? And can we remove them? I would be happy to provide a PR if removal is OK-ed.

grampelberg:
It is probably a side effect of serialization. Definitely worth removing! #2545 and #2450 are definitely related.

cppforlife:
👍 looks like it's also affecting `linkerd inject`

ihcsim:
[...]

Issue #2545, opened 2019-03-22
Linkerd inject should respect field omissions


bourquep:
### What is the issue?

When my source deployment yaml file has no `resources` key, the injected yaml ends up having `resources: {}`

### How can it be reproduced?

* Run `linkerd inject` on a deployment yaml file that has no `resources` key.
* Observe that the output yaml has a `resources: {}` key.

#### Input file example

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:1.7.9
          ports:
            - containerPort: 80
```

#### Inject output example

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: nginx
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  strategy: {}
  template:
    metadata:
      annotations:
        linkerd.io/created-by: linkerd/cli edge-19.2.5
        linkerd.io/identity-mode: disabled
        linkerd.io/proxy-version: edge-19.2.5
      creationTimestamp: null
      labels:
        app: nginx
        linkerd.io/control-plane-ns: linkerd
        linkerd.io/proxy-deployment: nginx-deployment
    spec:
      containers:
      - image: nginx:1.7.9
        name: nginx
        ports:
        - containerPort: 80
        resources: {}
      - env:
        - name: LINKERD2_PROXY_LOG
          value: warn,linkerd2_proxy=info
        - name: LINKERD2_PROXY_CONTROL_URL
          value: tcp://linkerd-destination.linkerd.svc.cluster.local:8086
        - name: LINKERD2_PROXY_CONTROL_LISTENER
          value: tcp://0.0.0.0:4190
        - name: LINKERD2_PROXY_METRICS_LISTENER
          value: tcp://0.0.0.0:4191
        - name: LINKERD2_PROXY_OUTBOUND_LISTENER
          value: tcp://127.0.0.1:4140
        - name: LINKERD2_PROXY_INBOUND_LISTENER
          value: tcp://0.0.0.0:4143
        - name: LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES
          value: .
        - name: LINKERD2_PROXY_POD_NAMESPACE
          valueFrom:
[...]

PR #3886, opened 2020-01-05
Remove empty fields from generated configs

mayankshah1607:
Fixes
- https://github.com/linkerd/linkerd2/issues/2962
- https://github.com/linkerd/linkerd2/issues/2545

### Problem
Field omissions for workload objects are not respected while marshaling to JSON.

### Solution
After digging a bit into the code, I came to realize that while marshaling, workload objects have empty structs as values for various fields which would rather be omitted. As of now, the standard library`encoding/json` does not support zero values of structs with the `omitemty` tag. The relevant issue can be found [here](https://github.com/golang/go/issues/11939). To tackle this problem, the object declaration should have _pointer-to-struct_ as a field type instead of _struct_ itself. However, this approach would be out of scope as the workload object declaration is handled by the k8s library.

I was able to find a drop-in replacement for the `encoding/json` library which supports zero value of structs with the `omitempty` tag. It can be found [here](https://github.com/clarketm/json). I have made use of this library to implement a simple filter like functionality to remove empty tags once a YAML with empty tags is generated, hence leaving the previously existing methods unaffected


pkg/inject/inject.go
```diff
@@ -283,7 +288,9 @@ func (conf *ResourceConfig) JSONToYAML(bytes []byte) ([]byte, error) {
 	if err := json.Unmarshal(bytes, obj); err != nil {
 		return nil, err
 	}
-	return yaml.Marshal(obj)
+
+	filteredJSON, _ := jsonfilter.Marshal(&obj)
```
Pothulapati: Thanks for the contribution :) 
Add a check for err.
Also, Can you move this Conversion functionality into a common function and re-use that?

mayankshah1607: @Pothulapati Sure, will do! If this approach seems fine, I could go ahead and clean up the code a bit! Just needed a thumbs-up on this kind of implementation

adleong:
It looks like https://github.com/clarketm/json pulls in commits from https://github.com/golang/go master.  Currently it's based off of https://github.com/golang/go/commit/54697702e435bddb69c0b76b25b3209c78d2120a which is the second latest commit.  So it's actually using a newer version of `encoding/json` than we are (1.13.4).  

The relevant diff is effectively this commit: https://github.com/clarketm/json/commit/6acb9d8d1d575a7961e9414c3eb3b3237bfb2bb4 which looks pretty sane to me.

So I'm pretty comfortable with adding this dependency.

pkg/inject/inject.go
```diff
@@ -177,7 +178,11 @@ func (conf *ResourceConfig) AppendPodAnnotation(k, v string) {
 
 // YamlMarshalObj returns the yaml for the workload in conf
 func (conf *ResourceConfig) YamlMarshalObj() ([]byte, error) {
-	return yaml.Marshal(conf.workload.obj)
+	j, err := jsonfilter.Marshal(conf.workload.obj)
```
adleong: This should definitely have a comment explaining why we're not using `encoding/json` here.

mayankshah1607:
@adleong @Pothulapati done! Mind taking a look now?
I've added a reusable method `getFilteredJSON()` and made it work with the existing code for `inject` / `uninject`. I've also added a comment explaining why we're not making use of `encoding/json` in the above mentioned method, as @adleong mentioned :smile:

pkg/inject/inject.go
```diff
@@ -939,5 +949,22 @@ func (conf *ResourceConfig) InjectNamespace(annotations map[string]string) ([]by
...
+		return nil, err
+	}
+	return yaml.JSONToYAML(j)
+}
+
+//getFilteredJSON method performs JSON marshaling such that zero values of
+//empty structs are respected by `omitempty` tags. We make use of a drop-in
+//replacement of the standard json/encoding library, without which empty struct values
+//present in workload objects would make it into the marshaled JSON.
+func getFilteredJSON(conf runtime.Object) ([]byte, error) {
```
adleong: I think the `getFilteredJSON` method you have have here is identical to `jsonfilter.Marshal`.  It's great that we have this comment, but I don't think the function gives us any benefit.

mayankshah1607: @adleong We'd have to be making use of the `jsonfilter.Marshal` method at multiple places. I wasn't really sure if it would be a nice idea to add such a comment at multiple places. I thought it would be a great idea to just have a reusable function that makes use of `jsonfilter.Marshal` that contributors could use in the future, instead of having to directly deal with a different version of `jso/encoding` by themselves, preserving the modularity of the codebase. 

WDYT?

adleong: Gotcha.  In that case, I think the `getFilteredJSON` function body can simply be a one-liner: `return jsonfilter.Marshal(&conf)`

mayankshah1607:
@adleong could you take a look now? I hope it looks fine :smile:

mayankshah1607:
Thank you @adleong !
@cpretzer mind taking a look do that we could get this merged?

PR #3887, opened 2020-01-06
Unify issuance lifetime name

zaharidichev:
Due to wrong snake casing, lifetime setting lifetime issuance was not reflected when installing through helm.  I think it is best to unify the spelling.


zaharidichev:
@grampelberg not that I can think of. @ihcsim can you think of some case where that would break upgrade ?

PR #3889, opened 2020-01-07
Upgraded actions/checkout to v2

### LINKED ISSUE #3881: Evaluate actions/checkout@v2
Per https://github.com/actions/checkout/issues/27#issuecomment-570397019, our incorrect SHA issue may be resolved in `actions/checkout@v2`. We should test this, and if it works, we can roll back our workaround in #3602.

alpeb:
Reverts #3602 and Fixes #3881

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -8,30 +8,28 @@ on:
     tags:
     - "*"
 
-## Jobs by event type and dependencies:
+# Jobs by event type and dependencies:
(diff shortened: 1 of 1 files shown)


alpeb:
Thanks for the review @siggy :ok_hand: 
That fetch command will indeed help us verifying this is working as intended. It should also match the `github.sha` presented in the `Validate go deps -> Dump GitHub context`  output.

PR #3897, opened 2020-01-08
proxy: v2.83.0-experimental

olix0r:
This is an experimental release that includes large changes to the
proxy's request buffering and backpressure infrastructure.

Please exercise caution before deploying this proxy version into mission
critical environments.


olix0r:
Test failure appears to be broken traffic split metrics which... is not totally shocking. I will investigate.

PR #3766, opened 2019-11-27
Enable mixed configuration of skip-[inbound|outbound]-ports

javaducky:
**Subject**
Enable mixed configuration of `skip-[inbound|outbound]-ports` using port numbers and ranges

**Problem**
The ability to configure port ranges for the `skip-[inbound|outbound]-ports` setting was requested.  The existing functionality allowed for only a comma-separated list of port numbers.

**Solution**
A new method has been created which will accept the mixed use of port numbers and ranges for the noted configuration settings.  As an example, `--skip-inbound-ports 8-10,15,17` may be provided via CLI which would allow for the configuration to be applied to ports 8, 9, 10, 15, and 17.  Unit tests have also been provided for this method.

Overall solution depends upon changes to the `linkerd2-proxy-init` with https://github.com/linkerd/linkerd2-proxy-init/pull/7.

**Validation**
Created a new unit tests to confirm desired functionality with applicable success and failure scenarios.

Fixes #3752 

Signed-off-by: Paul Balogh <javaducky@gmail.com>

PR #3903, opened 2020-01-09
Add checks for CNI plugin

### LINKED ISSUE #3804: Introduce CNI post-installation checks
Currently, after running `linkerd install-cni`, both `linkerd check` and `linkerd check --pre` return errors, because the `install-cni` command doesn't satisfy the conditions expected by `check`. We should consider either:

1. Update `linkerd check` to not fail if the CNI resources are present, and/or
1. Introduce a `linkerd check --cni` flag for CNI users to confirm that the CNI DaemonSet is healthy

If we decided to do 2, the following status info from the DaemonSet will be of interest:
```
kubectl -n linkerd describe ds linkerd-cni # on a 3-node cluster
...
Desired Number of Nodes Scheduled: 3
Current Number of Nodes Scheduled: 3
Number of Nodes Scheduled with Up-to-date Pods: 3
Number of Nodes Scheduled with Available Pods: 3
Number of Nodes Misscheduled: 0
```

Before starting this task, please confirm scope, prioritization and estimated effort with @grampelberg.

zaharidichev:
As part of the effort to remove the "experimental" label from the CNI plugin, this PR introduces a `linkerd check --cni`.

Fixes #3804

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -180,6 +180,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 
 	hc := healthcheck.NewHealthChecker(checks, &healthcheck.Options{
 		ControlPlaneNamespace: controlPlaneNamespace,
+		CNINamespace:          cniNamespace,
 		DataPlaneNamespace:    options.namespace,
 		KubeConfig:            kubeconfigPath,
 		KubeContext:           kubeContext,
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -149,6 +149,8 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		checks = append(checks, healthcheck.LinkerdPreInstallChecks)
 		if !options.cniEnabled {
 			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
+		} else {
+			checks = append(checks, healthcheck.LinkerdCniPluginChecks)
 		}
 		installManifest, err = renderInstallManifest()
 		if err != nil {
@@ -168,7 +170,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 			} else {
 				checks = append(checks, healthcheck.LinkerdControlPlaneVersionChecks)
 			}
-
+			checks = append(checks, healthcheck.LinkerdCniPluginChecks)
 			checks = append(checks, healthcheck.LinkerdHAChecks)
 		}
 	}
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -150,10 +150,10 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 	var installManifest string
 	if options.preInstallOnly {
 		checks = append(checks, healthcheck.LinkerdPreInstallChecks)
-		if !options.cniEnabled {
-			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
-		} else {
+		if options.cniEnabled {
 			checks = append(checks, healthcheck.LinkerdCNIPluginChecks)
+		} else {
+			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
 		}
 		installManifest, err = renderInstallManifest()
 		if err != nil {
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -65,6 +65,9 @@ func (options *checkOptions) validate() error {
 	if options.preInstallOnly && options.dataPlaneOnly {
 		return errors.New("--pre and --proxy flags are mutually exclusive")
 	}
+	if !options.preInstallOnly && options.cniEnabled {
+		return errors.New("--linkerd-cni-enabled can only be used with --pre")
+	}
 	if options.output != tableOutput && options.output != jsonOutput {
 		return fmt.Errorf("Invalid output type '%s'. Supported output types are: %s, %s", options.output, jsonOutput, tableOutput)
 	}
@@ -150,7 +153,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		if !options.cniEnabled {
 			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
 		} else {
-			checks = append(checks, healthcheck.LinkerdCniPluginChecks)
+			checks = append(checks, healthcheck.LinkerdCNIPluginChecks)
 		}
 		installManifest, err = renderInstallManifest()
 		if err != nil {
@@ -170,7 +173,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 			} else {
 				checks = append(checks, healthcheck.LinkerdControlPlaneVersionChecks)
 			}
-			checks = append(checks, healthcheck.LinkerdCniPluginChecks)
+			checks = append(checks, healthcheck.LinkerdCNIPluginChecks)
 			checks = append(checks, healthcheck.LinkerdHAChecks)
 		}
 	}
--- a/cli/cmd/main_test.go
+++ b/cli/cmd/main_test.go
@@ -55,6 +55,7 @@ func diffTestdata(t *testing.T, path, actual string) {
 	if actual == expected {
 		return
 	}
+	println(actual)
 	dmp := diffmatchpatch.New()
 	diffs := dmp.DiffMain(expected, actual, true)
 	diffs = dmp.DiffCleanupSemantic(diffs)
--- a/cli/cmd/main_test.go
+++ b/cli/cmd/main_test.go
@@ -55,7 +55,6 @@ func diffTestdata(t *testing.T, path, actual string) {
 	if actual == expected {
 		return
 	}
-	println(actual)
 	dmp := diffmatchpatch.New()
 	diffs := dmp.DiffMain(expected, actual, true)
 	diffs = dmp.DiffCleanupSemantic(diffs)
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -143,7 +143,6 @@ const (
 	LinkerdCNIResourceLabel = "linkerd.io/cni-resource"
 
 	linkerdCNIDisabledSkipReason = "skipping check because CNI is not enabled"
-	linkerdCNIEnabledSkipReason  = "skipping check because CNI is enabled"
 	linkerdCNIResourceName       = "linkerd-cni"
 	linkerdCNIConfigMapName      = "linkerd-cni-config"
 )
@@ -415,9 +414,6 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "control plane namespace does not already exist",
 					hintAnchor:  "pre-ns",
 					check: func(context.Context) error {
-						if hc.NoInitContainer {
-							return &SkipError{Reason: linkerdCNIEnabledSkipReason}
-						}
 						return hc.checkNamespace(hc.ControlPlaneNamespace, false)
 					},
 				},
@@ -1311,7 +1307,7 @@ func (hc *HealthChecker) runCheckRPC(categoryID CategoryID, c *checker, observer
 }
 
 func (hc *HealthChecker) controlPlaneComponentsSelector() string {
-	return fmt.Sprintf("%s in (%s),%s notin (true)", k8s.ControllerNSLabel, hc.ControlPlaneNamespace, LinkerdCNIResourceLabel)
+	return fmt.Sprintf("%s,!%s", k8s.ControllerNSLabel, LinkerdCNIResourceLabel)
 }
 
 // PublicAPIClient returns a fully configured public API client. This client is
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -132,11 +132,18 @@ const (
 	// is correct. These checks are no ops if linkerd is not in HA mode
 	LinkerdHAChecks CategoryID = "linkerd-ha-checks"
 
+	// LinkerdCniPluginChecks adds checks to validate that the CNI
+	/// plugin is installed and ready
+	LinkerdCniPluginChecks CategoryID = "linkerd-cni-plugin"
+
 	// linkerdCniResourceLabel is the label key that is used to identify
 	// whether a Kubernetes resource is related to the install-cni command
 	// The value is expected to be "true", "false" or "", where "false" and
 	// "" are equal, making "false" the default
 	linkerdCniResourceLabel = "linkerd.io/cni-resource"
+
+	cniDisabledSkipReason = "skipping check because CNI is not enabled"
+	cniEnabledSkipReason  = "skipping check because CNI is enabled"
 )
 
 // HintBaseURL is the base URL on the linkerd.io website that all check hints
@@ -166,6 +173,12 @@ var (
 		"linkerd-web",
 		"linkerd-tap",
 	}
+
+	expectedCniServiceAccountNames = []string{
+		"linkerd-cni",
+	}
+	expectedCniDsName = "linkerd-cni"
+	cniConfigMapName  = "linkerd-cni-config"
 )
 
 // Resource provides a way to describe a Kubernetes object, kind, and name.
@@ -319,6 +332,7 @@ type HealthChecker struct {
 	uuid             string
 	issuerCert       *tls.Cred
 	roots            []*x509.Certificate
+	cniDaemonSet     *appsv1.DaemonSet
 }
 
 // NewHealthChecker returns an initialized HealthChecker
@@ -404,6 +418,9 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "control plane namespace does not already exist",
 					hintAnchor:  "pre-ns",
 					check: func(context.Context) error {
+						if hc.NoInitContainer {
+							return &SkipError{Reason: cniEnabledSkipReason}
+						}
 						return hc.checkNamespace(hc.ControlPlaneNamespace, false)
 					},
 				},
@@ -500,14 +517,14 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "no ClusterRoles exist",
 					hintAnchor:  "pre-l5d-existence",
 					check: func(context.Context) error {
-						return hc.checkClusterRoles(false)
+						return hc.checkClusterRoles(false, hc.expectedRBACNames(), hc.NoInitContainer)
 					},
 				},
 				{
 					description: "no ClusterRoleBindings exist",
 					hintAnchor:  "pre-l5d-existence",
 					check: func(context.Context) error {
-						return hc.checkClusterRoleBindings(false)
+						return hc.checkClusterRoleBindings(false, hc.expectedRBACNames(), hc.NoInitContainer)
 					},
 				},
 				{
@@ -535,7 +552,7 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "no PodSecurityPolicies exist",
 					hintAnchor:  "pre-l5d-existence",
 					check: func(context.Context) error {
-						return hc.checkPodSecurityPolicies(false)
+						return hc.checkPodSecurityPolicies(false, []string{fmt.Sprintf("linkerd-%s-control-plane", hc.ControlPlaneNamespace)}, hc.NoInitContainer)
 					},
 				},
 			},
@@ -556,7 +573,7 @@ func (hc *HealthChecker) allCategories() []category {
 					hintAnchor:  "l5d-existence-cr",
 					fatal:       true,
 					check: func(context.Context) error {
-						return hc.checkClusterRoles(true)
+						return hc.checkClusterRoles(true, hc.expectedRBACNames(), hc.NoInitContainer)
 					},
 				},
 				{
@@ -564,7 +581,7 @@ func (hc *HealthChecker) allCategories() []category {
 					hintAnchor:  "l5d-existence-crb",
 					fatal:       true,
 					check: func(context.Context) error {
-						return hc.checkClusterRoleBindings(true)
+						return hc.checkClusterRoleBindings(true, hc.expectedRBACNames(), hc.NoInitContainer)
 					},
 				},
 				{
@@ -604,7 +621,7 @@ func (hc *HealthChecker) allCategories() []category {
 					hintAnchor:  "l5d-existence-psp",
 					fatal:       true,
 					check: func(context.Context) error {
-						return hc.checkPodSecurityPolicies(true)
+						return hc.checkPodSecurityPolicies(true, []string{fmt.Sprintf("linkerd-%s-control-plane", hc.ControlPlaneNamespace)}, hc.NoInitContainer)
 					},
 				},
 			},
@@ -618,6 +635,9 @@ func (hc *HealthChecker) allCategories() []category {
 					fatal:       true,
 					check: func(context.Context) (err error) {
 						hc.uuid, hc.linkerdConfig, err = hc.checkLinkerdConfigConfigMap()
+						if hc.linkerdConfig != nil {
+							hc.NoInitContainer = hc.linkerdConfig.Global.CniEnabled
+						}
 						return
 					},
 				},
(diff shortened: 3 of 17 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -1270,17 +1405,40 @@ func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
...
+	crList, err := hc.kubeAPI.RbacV1().Roles(hc.ControlPlaneNamespace).List(options)
+	if err != nil {
+		return err
+	}
+
+	objects := []runtime.Object{}
+
+	for _, item := range crList.Items {
+
+		if skipCni && hc.skipNoInitContainerResources(item.ObjectMeta.Labels) {
```
adleong: same question about skipCni here

pkg/healthcheck/healthcheck.go
```diff
@@ -1270,17 +1405,40 @@ func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
...
 		item := item // pin
 		objects = append(objects, &item)
 	}
 
-	return checkResources("ClusterRoles", objects, hc.expectedRBACNames(), shouldExist)
+	return checkResources("ClusterRoles", objects, expectedRoles, shouldExist)
 }
 
-func (hc *HealthChecker) checkClusterRoleBindings(shouldExist bool) error {
+func (hc *HealthChecker) checkRoles(shouldExist bool, expectedRoles []string, skipCni bool) error {
```
adleong: same question about expectedRoles here

cli/cmd/check.go
```diff
@@ -45,6 +46,7 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
 	flags.BoolVar(&options.cniEnabled, "linkerd-cni-enabled", options.cniEnabled, "When running pre-installation checks (--pre), assume the linkerd-cni plugin is already installed, and a NET_ADMIN check is not needed")
 	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
+	flags.BoolVar(&options.cniOnly, "cni", options.cniOnly, "Only run CNI checks, to determine if the plugin is installed and ready")
```
adleong: Can I get a high level description of the checks that are added in this PR?

Do these just check for the existence of certain resources that need to exist for CNI to work?  If so, shouldn't we always do those checks if the --linkerd-cni-enabled flag is set?  I don't understand the purpose of the `--cni` flag compared to `--linkerd-cni-enabled`.  Why does this flag disable other checks?

cli/cmd/check.go
```diff
@@ -182,7 +187,7 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		APIAddr:               apiAddr,
 		VersionOverride:       options.versionOverride,
 		RetryDeadline:         time.Now().Add(options.wait),
-		NoInitContainer:       options.cniEnabled,
+		NoInitContainer:       cniEnabled,
```
adleong: Not related to this PR, but I'll just note that this being a negative is really confusing to me, especially when we get into the double negatives.

pkg/healthcheck/healthcheck.go
```diff
@@ -701,6 +714,122 @@ func (hc *HealthChecker) allCategories() []category {
...
+				},
+				{
+					description: "cni plugin DaemonSet exists",
+					hintAnchor:  "cni-plugin-ds-exists",
+					fatal:       true,
+					check: func(context.Context) (err error) {
+						if !hc.NoInitContainer {
+							return &SkipError{Reason: cniDisabledSkipReason}
+						}
+						hc.cniDaemonSet, err = hc.kubeAPI.GetDaemonSet(hc.ControlPlaneNamespace, expectedCniDsName)
```
adleong: why do we create a dedicated method for getting the DS instead of just doing `hc.kubeAPI.CoreV1()...` directly like we do for configmaps in an above check.

cli/cmd/check.go
zaharidichev: Yes this actually got a bit more confusing that it needs to. I will revise it. Otherwise the just of the checks are: 

- Check for the resources that need to exist
- Check for the state of the deamonset (i.e. how many nodes are in the cluster vs how many nodes have the cni  pod up and running)

pkg/healthcheck/healthcheck.go
```diff
@@ -1257,7 +1386,13 @@ func (hc *HealthChecker) expectedRBACNames() []string {
...
 }
 
-func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
+func (hc *HealthChecker) expectedCniRBACNames() []string {
+	return []string{
+		"linkerd-cni",
+	}
+}
+
+func (hc *HealthChecker) checkClusterRoles(shouldExist bool, expectedRoles []string, skipCni bool) error {
```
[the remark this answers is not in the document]
zaharidichev: It gets called with the expected CNI RBAC names right here https://github.com/linkerd/linkerd2/pull/3903/files#diff-d4056ff163bcf2aeacefb2a34164563cR754

zaharidichev:
@adleong Thanks a bunch for the review. Made me realize I could simplify some things a bit. So at the moment these checks work as follows: 

- if you do a --pre check you need to specify the --linkerd-cni-enabled in order for them not to fail (skips the privilige checking, etc)
- if you do `linkerd check` if the cni is enabled the checks will be run as well.
- there is no --cni flag anymore. 

The fact that I have extra params to pass the expected RBAC names is because they defer based on whether this is a CNI check or linkerd control plane check. 
 
There is something else that needs to happen I think but I prefer that to be in a seprate PR is to be able to provide a cni-namespace flag to the check because right now the checks assume that the plugin resourceis installed in the same namespace as the control plane.

pkg/healthcheck/healthcheck.go
```diff
@@ -404,6 +418,9 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "control plane namespace does not already exist",
 					hintAnchor:  "pre-ns",
 					check: func(context.Context) error {
+						if hc.NoInitContainer {
```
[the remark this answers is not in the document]
zaharidichev: Yes this is correct. We can have a separate namespace for the CNI and that would not cause any major trouble. It will also simplify the logic right here.  Currently the `linkerd install-cni` uses the control plane namespace that is set as a flag.

pkg/healthcheck/healthcheck.go
```diff
@@ -1270,17 +1412,37 @@ func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
 
 	for _, item := range crList.Items {
 
-		if hc.skipNoInitContainerResources(item.ObjectMeta.Labels) {
+		if skipCni && isCniResource(item.ObjectMeta.Labels) {
```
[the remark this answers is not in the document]
zaharidichev: So this piece of logic was introduced some time ago because of the situation described right [here](https://github.com/linkerd/linkerd2/pull/3066#issuecomment-512479059). 

So originally this logic was just skipping resources that were CNI when CNI was enabled to check resources that are not CNI. Now however, we want to use this logic to check resources  that are not CNI when CNI is enabled and disabled, we also want to use it to check CNI resources when CNI is enabled. Therefore the ectra boolean that could be passed in depending on the check we are performing. Reading it, I see it does have some cognitive luggage indeed. Aghh... how soething so simple turned out to be so unnecessarily complicated....

pkg/healthcheck/healthcheck.go
```diff
@@ -1625,6 +1806,12 @@ func (hc *HealthChecker) checkCanCreateNonNamespacedResources() error {
 			continue
 		}
 
+		if obj.GetKind() == "Namespace" && hc.NoInitContainer {
```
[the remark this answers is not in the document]
zaharidichev: true!

pkg/healthcheck/healthcheck.go
```diff
@@ -1257,7 +1393,13 @@ func (hc *HealthChecker) expectedRBACNames() []string {
...
 }
 
-func (hc *HealthChecker) checkClusterRoles(shouldExist bool) error {
+func (hc *HealthChecker) expectedCniRBACNames() []string {
+	return []string{
+		"linkerd-cni",
+	}
+}
+
+func (hc *HealthChecker) checkClusterRoles(shouldExist bool, expectedRoles []string, skipCni bool) error {
```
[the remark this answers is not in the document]
zaharidichev: I agree the omission of the parameter. however I think having a seprate section of CNI checks that contains a check for the cluster role specific to CNI makes more sense rather than combining it into the control plane checks. Wouldn't you say so?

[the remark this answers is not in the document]
zaharidichev: In fact looking at this again, here is why I added this bool parameter and I think it is needed. So there are three cases where these methods are hit

- LinkerdPreInstallGlobalResourcesChecks - this is run during install. This needs to skip the CNI resources if cni is enabled (depends on hc.NoInitContainer)
- LinkerdConfigChecks - same as above really
- LinkerdCniChecks - here hc.NoInitContainer is true which means cni is enabled. Nevertheless we should not be skipping these resources because thats what we are checking for. 

So we cannot purely rely on the values of NoInitContainers to determine what we need to do because its interpretation changes based on the check that we are running. Therefore the extra bool.  The whole thing kind of stinks to me, I am not sure how to make it simpler so if you have any suggestions feel free to share.

pkg/healthcheck/healthcheck.go
```diff
@@ -618,6 +635,9 @@ func (hc *HealthChecker) allCategories() []category {
 					fatal:       true,
 					check: func(context.Context) (err error) {
 						hc.uuid, hc.linkerdConfig, err = hc.checkLinkerdConfigConfigMap()
+						if hc.linkerdConfig != nil {
```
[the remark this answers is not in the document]
zaharidichev: That makes total sense

pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
adleong: seems to me that this is our opportunity to change this before we take off the experimental flag.  I don't have any first hand experience with the install cni flow, but it seems like putting it in its own namespace by default (e.g. `linkerd-cni`) would simplify things.

cli/cmd/check.go
```diff
@@ -149,6 +152,8 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		checks = append(checks, healthcheck.LinkerdPreInstallChecks)
 		if !options.cniEnabled {
 			checks = append(checks, healthcheck.LinkerdPreInstallCapabilityChecks)
+		} else {
+			checks = append(checks, healthcheck.LinkerdCniPluginChecks)
```
adleong: Ok, after taking a step back, searching my soul, communing with nature, and going on a vision quest, I think I know how to simplify this.  Behold.

The `checkClusterRoles` and related methods which are part of `LinkerdPreInstallCapabilityChecks` should use `linkerd.io/control-plane-ns,!linkerd.io/cni-resource` as the label selector to select all linkerd components except the CNI ones.  This should not require checking `noInitContainer` at all since this check ONLY cares about the regular resources, not the CNI ones.  This check is run by `LinkerdPreInstallCapabilityChecks` with shouldExist=false and `LinkerdConfigChecks` with shouldExist=true.  These checks should always run, regardless of if CNI is enabled or disabled.

We should add new `checkCniRoles` and related methods.  These methods should be checked if check is run in `--pre` mode with the CNI flag enabled and also should be checked if check is run in non-pre mode and CNI mode is marked as enabled in the linkerd config.  Specifying the CNI enabled flag to a value that differs from the value in the configmap` should be a check error.  These methods should use the `linkerd.io/cni-resource` label selector to only look at CNI resources and validate that they exist.

pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
adleong: I also wouldn't complain if we renamed NoInitContainer to CniEnabled.

cli/cmd/check.go
ihcsim: We should be able to specify multiple labels in the `ListOptions.LabelSelector` where each label is separated by `,`.  See [here](https://godoc.org/k8s.io/apimachinery/pkg/labels#Parse) for the supported string format. Then we won't need all the `noInitContainer` and `skipCNI` `if` conditionals.

pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
ihcsim: Is it possible to separate the CNI check from the control plane pre checks?  IIUC, the way that this PR implements the CNI check is by using `linkerd check --pre --linkerd-cni-enabled`. The `healthcheck. LinkerdPreInstallChecks` check category will always be run, even though it has nothing to do with CNI installation.

What I mean is:

```
linkerd install-cni | k apply -f -
linkerd check --cni # only run checks under the healthcheck.LinkerdCniPluginChecks catgory
linkerd check --pre # work as before
linkerd install | k apply -f -
etc...
```

pkg/healthcheck/healthcheck.go
```diff
@@ -132,11 +132,18 @@ const (
 	// is correct. These checks are no ops if linkerd is not in HA mode
 	LinkerdHAChecks CategoryID = "linkerd-ha-checks"
 
+	// LinkerdCniPluginChecks adds checks to validate that the CNI
+	/// plugin is installed and ready
+	LinkerdCniPluginChecks CategoryID = "linkerd-cni-plugin"
```
ihcsim: Can we follow the Go code review convention on [initialism](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) here and elsewhere? So `LinkerdCNIPluginChecks`, instead of `LinkerdCniPluginChecks`? Thanks.

cli/cmd/check.go
zaharidichev: @adleong 
> Specifying the CNI enabled flag to a value that differs from the value in the configmapshould be a check error.

However specifying the value to false is the same as not specifying it at all. So should we change that bool to be a pointer to a bool (feels strange..) to cover the case when it was not set at all. Alternatively we can resort to the validation that I already have in place (the one you originally suggested) that disallows setting the `--linkerd-cni-enabled` unless you are doing a `--pre` check

pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
zaharidichev: > I also wouldn't complain if we renamed NoInitContainer to CniEnabled.

Agreed! I created #3930 for that and will do it, just dont want to stuff it into this PR as it will make reviewing harder.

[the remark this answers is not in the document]
zaharidichev: Yes I had it like that initially but @adleong comented that we can get rid of that flag and just have it run as part of the --pre checks. Do not reall have a strong opinion on that one. The way I see it is that` linkerd check --pre --linkerd-cni-enabled `pretty much shows whether you can install linkerd with cni enabled, which really consists of all the usual checks + the checks that ensure CNI is installed and in good shape...

zaharidichev:
@adleong I took a more agressive approach and just inlined the resource checks for CNI into the checks themselves. They always check for just one particular resource and are not concerned with checking for both existance and non existance so it could be a lot simpler. Also changed it up so the plugin is installed in a different namespace

pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
ihcsim: The goal of #3804 is to answer the question, "_did my CNI install correctly?_" I didn't expect `--pre` or anything else to change, [such as adding the CNI checks in the multi-stage check](https://github.com/linkerd/linkerd2/blob/748b026d31c3c4c0ebb683b7b9c977e0b4fe5ddd/cli/cmd/check.go#L176). fwiw, the person running `linkerd install-cni` may not be the same person who runs `linkerd check --pre`. IOW, just one command to tell me whether my CNI is healthy.

 Hopefully, this change in direction didn't complicate the implementation details. If it does, we should re-evaluate where we want to go with this. Thanks!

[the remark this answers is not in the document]
adleong: I don't necessarily have all the context, so I apologize if I'm missing the intent behind the original issue.  It seems to me like it makes sense to combine `linkerd check --cni` and `linkerd check --pre` into a single step which checks both.  In my mind, when you're asking the question "did my CNI install correctly?" what you *really* want to know is "is it okay to proceed with installing Linkerd?" and having the CNI correctly installed is just one part of that.  Therefore, I see the workflow as being:

```
linkerd install-cni | k apply -f -
linkerd check --pre --linkerd-cni-enabled
linkerd install --linkerd-cni-enabled | k apply -f -
linkerd check
```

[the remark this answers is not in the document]
ihcsim: The CNI health check is useful beyond the installation flow. AFAICT from this PR, the CNI check is run during:

```
linkerd check --pre --linkerd-cni-enabled
linkerd check
linkerd check --proxy
```

which is fine I guess? Personally, I still prefer a simpler one CNI command (like `linkerd check --cni`), with everything else control plane-related remains unchanged. At this point, I lean towards whichever introduces less complexity into the code. Thanks.

cli/cmd/check.go
```diff
@@ -149,6 +152,8 @@ func configureAndRunChecks(wout io.Writer, werr io.Writer, stage string, options
 		checks = append(checks, healthcheck.LinkerdPreInstallChecks)
 		if !options.cniEnabled {
```
adleong: Consider flipping the if and else blocks here to avoid the negation.

adleong: Wait, actually, don't we always want to run the LinkerdPreInstallCapabilityChecks, even when CNI is enabled?  This check will ensure that the Linkerd components (excluding the CNI ones) don't already exist.

cli/cmd/main_test.go
```diff
@@ -55,6 +55,7 @@ func diffTestdata(t *testing.T, path, actual string) {
 	if actual == expected {
 		return
 	}
+	println(actual)
```
adleong: rm debug line before merging

pkg/healthcheck/healthcheck.go
```diff
@@ -404,6 +415,9 @@ func (hc *HealthChecker) allCategories() []category {
 					description: "control plane namespace does not already exist",
 					hintAnchor:  "pre-ns",
 					check: func(context.Context) error {
+						if hc.NoInitContainer {
```
adleong: now that we're installing the CNI daemonset into a separate namespace, we can remove this check, right?

pkg/healthcheck/healthcheck.go
```diff
@@ -1152,6 +1310,10 @@ func (hc *HealthChecker) runCheckRPC(categoryID CategoryID, c *checker, observer
 	return true
 }
 
+func (hc *HealthChecker) controlPlaneComponentsSelector() string {
+	return fmt.Sprintf("%s in (%s),%s notin (true)", k8s.ControllerNSLabel, hc.ControlPlaneNamespace, LinkerdCNIResourceLabel)
```
adleong: Can this be `fmt.Sprintf("%s,!%s", k8s.ControllerNSLabel, LinkerdCNIResourceLabel)`?  Should be the same effect, but simpler

pkg/healthcheck/healthcheck.go
```diff
@@ -1883,17 +2031,3 @@ func checkControlPlaneReplicaSets(rst []appsv1.ReplicaSet) error {
 
 	return nil
 }
-
```
adleong: 😍

cli/cmd/check.go
zaharidichev: I dont think so, the `LinkerdPreInstallCapabilityChecks` checks check for NET_ADMIN and NET_RAW. We do not need these if we have CNI runnig. Correct ?

adleong: ah, I was getthing this mixed up with the check to ensure cluster scoped resources like ClusterRoles did not already exist.  nevermind

ihcsim:
@zaharidichev Thanks for working on this. Since @adleong already did a thorough job on the code, I 'm gonna skip that and just focus on testing it on GKE. 

TL;DR works great, but I am seeing an error causing the control plane to fail to start up, which I wonder if it's related to https://github.com/linkerd/linkerd2/pull/3766.

```
# start with a clean cluster
$ bin/linkerd version
Client version: git-2d798ab0
Server version: unavailable

# this works
$ bin/linkerd check --pre --expected-version=git-2d798ab0
...
linkerd-version
---------------
√ cli is up-to-date

Status check results are √

# ideally, this shouldn't fail because we use --pre to tell users whether the cluster is 
# fit for installation, but I think it's fine
bin/linkerd check --pre --linkerd-cni-enabled --expected-version=git-2d798ab0
...
linkerd-cni-plugin
------------------
× cni plugin ConfigMap exists
    configmaps "linkerd-cni-config" not found
    see https://linkerd.io/checks/#cni-plugin-cm-exists for hints

# install CNI
# this works
bin/linkerd install-cni --dest-cni-bin-dir=/home/kubernetes/bin|k apply -f -
...
k -n linkerd-cni get po                                      
NAME                READY   STATUS    RESTARTS   AGE
linkerd-cni-92kcb   1/1     Running   0          11s
linkerd-cni-l6fhm   1/1     Running   0          11s
linkerd-cni-sb5vk   1/1     Running   0          11s

# this works
bin/linkerd check --pre --linkerd-cni-enabled --expected-version=git-2d798ab0
...
linkerd-version
---------------
√ cli is up-to-date

Status check results are √
```

This is where my control plane installation failed. LMK if I'm missing something. If it's a legit bug, please create an issue for it. Thanks.
```
bin/linkerd install --linkerd-cni-enabled | k apply -f -

k -n linkerd get po                                                  
NAME                                      READY   STATUS              RESTARTS   AGE
linkerd-controller-749785ccd7-6fdx8       0/2     ContainerCreating   0          18s
linkerd-destination-76d5998446-wgq89      0/2     ContainerCreating   0          16s
linkerd-grafana-5555fbdfbb-lnvsv          0/2     ContainerCreating   0          11s
linkerd-identity-8956f5997-vcsqp          0/2     ContainerCreating   0          18s
linkerd-prometheus-658757f7fc-fp5fz       0/2     ContainerCreating   0          13s
linkerd-proxy-injector-6fd7544dd6-7cq75   0/2     ContainerCreating   0          11s
linkerd-sp-validator-68ffc65f5f-j2wnp     0/2     ContainerCreating   0          10s
linkerd-tap-67c7d7d49b-84962              0/2     ContainerCreating   0          9s
linkerd-web-7464459ff6-7wrvm              0/2     ContainerCreating   0          15s

k -n linkerd describe po linkerd-controller-749785ccd7-6fdx8
...
  Warning  FailedCreatePodSandBox  25s                 kubelet, gke-isim-dev-default-pool-0c4e32ef-9s3l  Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "ffb89dfd5a85e0aec0a7810d68ff49f5f93b796f7cb507bcd2f8952656a26a2c" network for pod "linkerd-controller-749785ccd7-6fdx8": NetworkPlugin cni failed to set up pod "linkerd-controller-749785ccd7-6fdx8_linkerd" network: linkerd-cni: failed to parse network configuration: json: cannot unmarshal number into Go struct field ProxyInit.linkerd.inbound-ports-to-ignore of type string
  Normal   SandboxChanged          14s (x12 over 54s)  kubelet, gke-isim-dev-default-pool-0c4e32ef-9s3l  Pod sandbox changed, it will be killed and re-created.
  Warning  FailedCreatePodSandBox  11s (x4 over 23s)   kubelet, gke-isim-dev-default-pool-0c4e32ef-9s3l  (combined from similar events): Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "bb38f27f0e1f27782fffdbe58cbfc33ac8d956c3ed6346327216c0763ba2bd0b" network for pod "linkerd-controller-749785ccd7-6fdx8": NetworkPlugin cni failed to set up pod "linkerd-controller-749785ccd7-6fdx8_linkerd" network: linkerd-cni: failed to parse network configuration: json: cannot unmarshal number into Go struct field ProxyInit.linkerd.inbound-ports-to-ignore of type string
```

ihcsim:
One other observation: it looks like the CNI namespace is changed from `linkerd` to `linkerd-cni`. Please add to the upgrade checklist at https://github.com/linkerd/website/issues/607 with info on how to do it. It should be just `linkerd install-cnI | k apply --prune -l <some-linkerd-label>  -f -`, before `linkerd upgrade`?

PR #3909, opened 2020-01-10
Validate incoming sidecar injection requests

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/3739

- Add method `isInjectAnnotationValid()` in `report.go` to catch invalid
inject annotations

### DIFF
--- a/pkg/inject/report.go
+++ b/pkg/inject/report.go
@@ -10,23 +10,27 @@ import (
 )
 
 const (
-	hostNetworkEnabled             = "host_network_enabled"
-	sidecarExists                  = "sidecar_already_exists"
-	unsupportedResource            = "unsupported_resource"
-	injectEnableAnnotationAbsent   = "injection_enable_annotation_absent"
-	injectDisableAnnotationPresent = "injection_disable_annotation_present"
-	annotationAtNamespace          = "namespace"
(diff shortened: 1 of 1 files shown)


mayankshah1607:
This PR is an alternative approach for https://github.com/linkerd/linkerd2/pull/3868, as suggested by @adleong . We could close https://github.com/linkerd/linkerd2/pull/3868 if we're okay with this PR's implementation

@alpeb @adleong @cpretzer @grampelberg

pkg/inject/report.go
```diff
@@ -154,6 +158,14 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
 		return podAnnotation == k8s.ProxyInjectDisabled, "", ""
 	}
 
+	if !isInjectAnnotationValid(conf.nsAnnotations) {
```
ihcsim: Instead of adding this helper function to iterate through the annotations, can we check the `podAnnotation` and `nsAnnotation` variables from line [154-155](https://github.com/linkerd/linkerd2/blob/93db0578fbff7cd61a0f924b19268d8bfa9fcfeb/pkg/inject/report.go#L154-L155) to make sure if they are non-empty, then they must be either `enabled` or `disabled`?

PR #3920, opened 2020-01-13
Add distributed tracing integration test

adleong:
This integration test roughly follows the [Linkerd guide to distributed tracing](https://linkerd.io/2019/10/07/a-guide-to-distributed-tracing-with-linkerd/).

We deploy the tracing components (oc-collector and jaeger), emojivoto, and nginx as an ingress to do span initiation.  We then watch the jaeger API and check that a trace is eventually created that includes traces from all of the data plane components: nginx, linkerd-proxy, web, voting, and emoji.


test/tracing/testdata/emojivoto.yaml
```diff
@@ -0,0 +1,201 @@
...
+kind: Deployment
+metadata:
+  creationTimestamp: null
+  name: emoji
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: emoji-svc
+  strategy: {}
```
zaharidichev: Just a nit. For good measure, you can jsut remove these empty fields as we already removed then in the other yaml files as well. Not that it would cause failures.

PR #3925, opened 2020-01-15
Enable no-use-before-define eslint rule

klingerf:
This is a follow-up to #3882, which adopted a bunch of new linting rules
in our Javascript codebase. The [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define) rule requires
moving some functions around, so I'm doing it in a separate branch.

Note that I was originally going to also enable the react/sort-comp rule
as part of this branch, but I decided that the sort ordering doesn't
work for our codebase.

Signed-off-by: Kevin Lingerfelt <kl@buoyant.io>

### DIFF
--- a/web/app/.eslintrc
+++ b/web/app/.eslintrc
@@ -71,6 +71,7 @@
     "react/jsx-filename-extension": 0,
     "react/jsx-fragments": 0,
     "react/jsx-one-expression-per-line": 0,
+    "react/sort-comp": 0,
 
     // These rules are disabled to support specific features/conventions that we use
     "no-bitwise": 0,
@@ -84,10 +85,6 @@
     // These rules represent best practices that we're not adhering to, and should be enabled
     "jsx-a11y/click-events-have-key-events": 0,
     "no-param-reassign": 0,
-    "react/no-did-update-set-state": 0,
-
-    // These rules require reordering and should be enabled in a follow-up branch
-    "no-use-before-define": 0,
-    "react/sort-comp": 0
+    "react/no-did-update-set-state": 0
   }
 }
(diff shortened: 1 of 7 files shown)


klingerf:
@alpeb For sure! The definition of the react/sort-comp rule provided by the airbnb eslint config is super customized. You can view it here:

https://github.com/airbnb/javascript/blob/6d05dd898acfec3299cc2be8b6188be542824965/packages/eslint-config-airbnb/rules/react.js#L239-L288

The sorting that they specify only makes sense if we adhere to their naming conventions for function names. For instance a function named `onClick` would sort above a function named `getInitialState`. But we don't really use consistent function names throughout the codebase, so we end up with an ordering that feels pretty arbitrary just to satisfy that rule.

One common example is that in components that poll, such as Namespace.jsx, we usually define four functions, in this order: `startServerPolling`, `stopServerPolling`, `loadFromServer`, `handleApiError`. That roughly maps to the order in which they're executed in the lifecycle of the component. But with react/sort-comp enabled, they would need to be ordered as `handleApiError`, `loadFromServer`, `startServerPolling`, `stopServerPolling`, and that doesn't make a lot of sense to me. We could add our own version of react/sort-comp that specifies an ordering that works for our existing conventions, but for now I've just disabled it.

alpeb:
@klingerf Makes perfect sense, thanks for the detailed explanation! :smile:

klingerf:
@scottcarol Whoa, good find! I've refactored the Navigation tests in a63aeba6 to make it more streamlined and add the default props. Let me know what you think.

scottcarol:
I'll also create a follow-up issue since our JS tests should not pass CI if there is a warning.

PR #3933, opened 2020-01-16
Refactor PR #3909

mayankshah1607:
This PR addresses https://github.com/linkerd/linkerd2/pull/3909#discussion_r365439172

- Remove helper function `isInjectAnnotationValid()`
- Add proxy-injection request validation to existing checks in
`report.go`

cc @ihcsim 

### DIFF
--- a/pkg/inject/report.go
+++ b/pkg/inject/report.go
@@ -158,11 +158,11 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
 		return podAnnotation == k8s.ProxyInjectDisabled, "", ""
 	}
 
-	if !isInjectAnnotationValid(conf.nsAnnotations) {
+	if !isInjectAnnotationValid(nsAnnotation) {
 		return true, invalidInjectAnnotationNamespace, ""
 	}
 
-	if !isInjectAnnotationValid(conf.pod.meta.Annotations) {
+	if !isInjectAnnotationValid(podAnnotation) {
 		return true, invalidInjectAnnotationWorkload, ""
 	}
 
@@ -180,11 +180,9 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
 	return false, "", annotationAtWorkload
 }
 
-func isInjectAnnotationValid(annotations map[string]string) bool {
-	for k, v := range annotations {
-		if k == k8s.ProxyInjectAnnotation && !(v == k8s.ProxyInjectEnabled || v == k8s.ProxyInjectDisabled) {
-			return false
-		}
+func isInjectAnnotationValid(annotation string) bool {
+	if annotation != "" && !(annotation == k8s.ProxyInjectEnabled || annotation == k8s.ProxyInjectDisabled) {
+		return false
 	}
 	return true
 }


pkg/inject/report.go
```diff
@@ -158,11 +158,11 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
 		return podAnnotation == k8s.ProxyInjectDisabled, "", ""
 	}
 
-	if !isInjectAnnotationValid(conf.nsAnnotations) {
+	if nsAnnotation != "" && !(nsAnnotation == k8s.ProxyInjectEnabled || nsAnnotation == k8s.ProxyInjectDisabled) {
```
ihcsim: Can we do just:
```
if nsAnnotation != k8s.ProxyInjectEnabled && nsAnnotation != k8s.ProxyInjectDisabled {
```

pkg/inject/report.go
```diff
@@ -158,11 +158,11 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
...
 		return true, invalidInjectAnnotationNamespace, ""
 	}
 
-	if !isInjectAnnotationValid(conf.pod.meta.Annotations) {
+	if podAnnotation != "" && !(podAnnotation == k8s.ProxyInjectEnabled || podAnnotation == k8s.ProxyInjectDisabled) {
```
ihcsim: As above, I don't think `podAnnotation != ""`  is necessary. Can we do just:
```
if podAnnotation != k8s.ProxyInjectEnabled && podAnnotation != k8s.ProxyInjectDisabled {
```

ihcsim:
@zaharidichev I am hoping these are just oneliner changes. Isn't your suggestion the same as:
```
func isInjectAnnotationValid(annotation string) bool {
  return annotation != "" && !(annotation == k8s.ProxyInjectEnabled || annotation == k8s.ProxyInjectDisabled) 
}
```
Since the conditional isn't very complicated, I think the extra inference isn't necessary.

ihcsim:
@mayankshah1607 @zaharidichev To confirm my understanding, this doesn't reject the request; it only skips the injection, right? So a user who provided an incorrect value would still see the workloads being deployed, but uninjected.

mayankshah1607:
@ihcsim yes, this approach only notifies users when the inject annotation holds an invalid value. The workload's would still be deployed but would be uninjected

mayankshah1607:
@ihcsim @zaharidichev it seems like `TestInjectable` in `report_test.go` is failing when we remove the `podAnnotation != "" ` or `nsAnnotation != ""` condition. It seems like an issue with the way these tests are written, because I don't see a flaw in removing `podAnnotation != "" ` .
In this case I think we'd have to either go with @zaharidichev suggestion.
Could you take a look?

zaharidichev:
@mayankshah1607 Left a commend on your code to point you to where the problem seems to be.

pkg/inject/report.go
```diff
@@ -158,11 +158,11 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
 		return podAnnotation == k8s.ProxyInjectDisabled, "", ""
 	}
 
-	if !isInjectAnnotationValid(conf.nsAnnotations) {
+	if !(nsAnnotation == k8s.ProxyInjectEnabled || nsAnnotation == k8s.ProxyInjectDisabled) {
 		return true, invalidInjectAnnotationNamespace, ""
 	}
 
-	if !isInjectAnnotationValid(conf.pod.meta.Annotations) {
+	if !(podAnnotation == k8s.ProxyInjectEnabled || podAnnotation == k8s.ProxyInjectDisabled) {
```
zaharidichev: Seems like your test is failing because before the assumption was that not gaing this annotation set is ok. Now you treat it like an error. The reason for this is because if you do a lookup on a map and you assign it to a value, if your map is string -> string and you have a missing key, yur result will be an empty string. You need to add a check for that to your test condition. Namely: 
`!(podAnnotation == k8s.ProxyInjectEnabled || podAnnotation == k8s.ProxyInjectDisabled) && podAnnotation != ""`

Same for `nsAnnotation`. Hope this helps.

mayankshah1607: @zaharidichev that's what we began with! @ihcsim suggested that we remove the `podAnnotation != ""` check (https://github.com/linkerd/linkerd2/pull/3933#discussion_r367562004). Seems like we'd have to stick to having that check then

zaharidichev: Thats what it seems to be. Hope I am not overlooking something, @ihcsim ?

ihcsim: Ah yes... the `podAnnotation !=""` and `nsAnnotation != ""` are needed after all, because empty string values are valid for these annotations. Thanks for pointing that out.

mayankshah1607:
@zaharidichev @ihcsim mind taking a look now?
I've made use of the helper function `isInjectAnnotationValid` - only, this time it makes use of the existing `podAnnotation` and `nsAnnotation` values. Also, code looks cleaner now :D

Issue #3505, opened 2019-10-01
Update 'check' command to validate HA


ihcsim:
Following https://github.com/linkerd/linkerd2/issues/3305#issuecomment-525033155, we'll like to update the `linkerd check` command to validate the properties of HA control planes, such as:

1. The control plane components (except, for `web`, `grafana` and `prometheus`) should have more than one replicas.
1. The replicated pods should be scheduled on different hosts.
1. The MWC and VWC's `FailurePolicy` should be set to `Fail` to prevent uninjected workloads from entering the service mesh.
1. The control plane components are assigned CPU and memory resource requirements.
1. `kube-system` has the inject skip annotation.

grampelberg:
Let's check to make sure the `kube-system` namespace has the skip annotation as well.

ihcsim:
Actually, IIRC, in order to support multi-stage installation, the `linkerd check` currently doesn't check for cluster-scoped resources, because the user running the command may not have the RBAC permissions. So checking for MWC/VWC's `FailurePolicy` and `kube-system` label won't work here.

grampelberg:
I think it is fine to do a `warning` that says we were unable to check because of RBAC.

mayankshah1607:
Hi, @grampelberg @ihcsim !
I'd like to work on this issue. Where can I start from?

grampelberg:
@mayankshah1607 the list @ihcsim has is pretty good, you'll want to:

1. Lookup whether the control plane was installed with `--ha`.
1. Introduce a new section in check for HA.
1. Add each line from the description as a separate check.

mayankshah1607:
@grampelberg @ihcsim  
This PR - https://github.com/linkerd/linkerd2/pull/3731 implements some parts of this issue like adding a new section to the checks to see if `--ha` is enabled and hence check if `kube-system` has the inject skip annotation.
I guess I'd have to wait for https://github.com/linkerd/linkerd2/pull/3731 to get merged to prevent any overlaps. Once merged, I'll open a PR that implements the checks mentioned above. Is that ok?

ihcsim:
@mayankshah1607 sgtm; I'll keep you posted.

mayankshah1607:
@ihcsim @grampelberg It seems like point 5 has been fixed in https://github.com/linkerd/linkerd2/pull/3731 . Could we remove it from this issue then?

ihcsim:
[...]

Issue #3305, opened 2019-08-22
Don't call the proxy-injector for pods in the kube-system namespace


dwj300:
### What is the issue?
On AKS, if the tunnelfront pod is not scheduled, very bad things happen with the cluster. In our case, we had an unhealthy node, which happened to contain both tunnelfront and one of our proxy-injectors. When the pods were deleted from the unhealthy node, tunnelfront could not be rescheduled since we were running linkerd in --HA mode, which sets the FailurePolicy on the MutatingAdmissionWebhook to Fail and the proxy-injector was not reachable. The proxy-injector was unable to be rescheduled on another node due to CPU resource constraints (something we are in the process of fixing now). We ran our cluster with `--HA --controller-replicas=2` since our cluster is so small, and we are currently working to identify why the second proxy-injector was either not reachable, or not scheduled.

I was wondering if either an option or a default setting to exclude the `kube-system` namespace from the admissions webhook could be added. From glancing at the docs, it seems like the following diff would would work:

```
webhooks:
- name: linkerd-proxy-injector.linkerd.io
  namespaceSelector:
    matchExpressions:
    - key: linkerd.io/is-control-plane
      operator: DoesNotExist
    - key: namespace
      operator: NotIn
      value: kube-system
```

Would this be something the team would be interested in? More than happy to send a PR :) 

### Environment

- Kubernetes Version:
- Cluster Environment: (GKE, AKS, kops, ...) AKS
- Host OS: Ubuntu 16.04
- Linkerd version: stable-2.4.0

### Additional context
Unfortunately, our DRI applied many mitigations, and as a result we do not have many logs. We have the k8s master logs that we have been digging through.

ihcsim:
@dwj300 The `webhooks.namespaceSelector` property that you referred to matches labels on the namespace. We [did consider](https://github.com/linkerd/linkerd2/issues/3104) updating the MWC to ignore `kube-system`, but decided to back out of the idea because by default, the `kube-system` namespace isn't labelled. 

[...]

PR #3942, opened 2020-01-17
cli: Update 'check' command to validate HA configuration

mayankshah1607:
This PR addresses https://github.com/linkerd/linkerd2/issues/3505

Checks are added to validate the following properties of HA:
- [x] The control plane components (except, for web, grafana and prometheus) should have more than one replicas.
- [x] ~~The replicated pods should be scheduled on different hosts.~~ (Not required anymore)
- [x] ~~The MWC and VWC's FailurePolicy should be set to Fail to prevent uninjected workloads from entering the service mesh.~~ (Not required anymore)
- [x] ~~The control plane components are assigned CPU and memory resource requirements.~~ (Not required anymore)

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1211,6 +1211,17 @@ func (hc *HealthChecker) allCategories() []category {
 						return &SkipError{Reason: "not run for non HA installs"}
 					},
 				},
+				{
+					description: "replicated control plane pods are scheduled on different hosts",
+					hintAnchor:  "l5d-control-plane-pod-hosts",
+					warning:     true,
+					check: func(ctx context.Context) error {
+						if hc.isHA() {
+							return hc.checkReplicatedPodsHosts()
+						}
+						return &SkipError{Reason: "not run for non HA installs"}
+					},
+				},
 			},
 		},
 	}
@@ -1266,6 +1277,20 @@ func (hc *HealthChecker) checkControlPlaneComponentsResourceRequirements() error
 			return fmt.Errorf("resource requirements not configured for containers %v in deployment \"%s\"", containers, conf.Name)
 		}
 
+	}
+	return nil
+}
+
+func (hc *HealthChecker) checkReplicatedPodsHosts() error {
+	for _, component := range linkerdHAControlPlaneComponents {
+		podList, err := hc.kubeAPI.CoreV1().Pods(hc.ControlPlaneNamespace).List(metav1.ListOptions{LabelSelector: getProxyDeploymentLabel(component)})
+		if err != nil {
+			return err
+		}
+
+		if ok, podName := scheduledOnDifferentHosts(podList); !ok {
+			return fmt.Errorf("pod %s may have been scheduled on the wrong host", podName)
+		}
 	}
 	return nil
 }
@@ -2177,4 +2202,28 @@ func hasResourceRequirements(conf *appsv1.Deployment) (bool, []string) {
 		return false, faulty
 	}
 	return true, faulty
+}
+
+func getProxyDeploymentLabel(deployment string) string {
+	return fmt.Sprintf("%s=%s", k8s.ProxyDeploymentLabel, deployment)
+}
+
+func scheduledOnDifferentHosts(podList *corev1.PodList) (bool, string) {
+	hosts := []string{}
+	for _, pod := range podList.Items {
+		if indexOf(hosts, pod.Status.HostIP) > -1 {
+			return false, pod.Name
+		}
+		hosts = append(hosts, pod.Status.HostIP)
+	}
+	return true, ""
+}
+
+func indexOf(slice []string, item string) int {
+	for i := range slice {
+		if slice[i] == item {
+			return i
+		}
+	}
+	return -1
 }
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/api/admissionregistration/v1beta1"
+
 	"github.com/linkerd/linkerd2/pkg/issuercerts"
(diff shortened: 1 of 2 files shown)


adleong:
Thanks for working on this, @mayankshah1607!  I'm going to wait until this is complete before giving it a full review, but if you have any questions or would like early feedback on some specific aspect of it, please let me know.

mayankshah1607:
Done! Apologies for the delay. I have added the above mentioned checks. Would you mind taking a look? @adleong @zaharidichev @ihcsim

adleong:
Thanks @mayankshah1607, this is really good work.

As part of reviewing this I've gone back and read #3505 and #3305 in more detail and I think most of these checks are not actually necessary any more.  Since the issue was filed, we added hard host anti-affinity rules to the HA components that require that Kubernetes schedule them on different hosts.  Similarly, the resource requirements and the webhook failure policies are part of the HA manifests and are automatically configured when Linkerd is installed with the --HA flag.  

With all that said, I still think the check for at least 1 available replica is still valuable.

I apologize for the poorly scoped issue and I'm sorry that I didn't notice this earlier.  Would you mind keeping the available replica check and removing the others?

As a side note, it would be interesting to have a more general purpose check that verifies that the installed resources match the manifest yamls.  This would encompass the anti-affinity configuration, the webhook failure policy, the resource requirements, and much more.  However, I suspect that this is a much more complex and involved change.

mayankshah1607:
@adleong no problem at all! :smile: 
Mind taking a look now? I have removed all the checks except the multiple replica one.
How would you feel about opening a separate issue for the general purpose check you mentioned? I could take that up with a little more detail provided :smile:

pkg/healthcheck/healthcheck.go
```diff
@@ -1152,11 +1161,36 @@ func (hc *HealthChecker) allCategories() []category {
...
+						}
+						return &SkipError{Reason: "not run for non HA installs"}
+					},
+				},
 			},
 		},
 	}
 }
 
+func (hc *HealthChecker) controlPlaneComponentsAreHA() error {
```
zaharidichev: I think this func name can be changes to reflect the fact that we are trying to establish a minimum number of available replicas.

pkg/healthcheck/healthcheck.go
```diff
@@ -1152,11 +1161,36 @@ func (hc *HealthChecker) allCategories() []category {
...
+	for _, component := range linkerdHAControlPlaneComponents {
+		conf, err := hc.kubeAPI.AppsV1().Deployments(hc.ControlPlaneNamespace).Get(component, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if conf.Status.AvailableReplicas <= 1 {
```
zaharidichev: Just to verify my understanding here @adleong refered to that as the  

> the check for at least 1 available replica is still valuable

However this effectivelly checks that there is more than one avaiable replicas and that is what it shold be? Correct? Also, since this number can change as pods come up and cross the `minReadySeconds` threshold to be considered Available, would it be worth to make that a retryable check, much like the `"data plane proxies are ready"` check ?

mayankshah1607: @zaharidichev yep, it makes sense to me to make it retryable. On it!

mayankshah1607:
@adleong @zaharidichev mind taking a look now?
- pulled changes from master
- renamed the func used in the check
- made the check retryable

zaharidichev:
@mayankshah1607 This looks great to me as well. Will run the tests tomorrow to make sure it all passes :)

zaharidichev:
@mayankshah1607 I agree with @cpretzer. Wrt to adding tests you would like to look at `healthcheck_test.go`. There are facilities there to create "fake" resources and exercise this piece of code. Let me know if you have any questions

pkg/healthcheck/healthcheck_test.go
```diff
@@ -3003,3 +3004,125 @@ func TestCniChecks(t *testing.T) {
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: %s
+  namespace: linkerd
+spec:
+  template:
+    spec:
+      containers:
+        - image: "nginx:1.7.9"
```
zaharidichev: no need to be opinionated :), you can put something more generic here ;)

pkg/healthcheck/healthcheck_test.go
```diff
@@ -3003,3 +3004,125 @@ func TestCniChecks(t *testing.T) {
...
+		case "linkerd-identity":
+			resourceDefs = append(resourceDefs, getSingleControlPlaneDef(component, replicaOptions.identity))
+		case "linkerd-sp-validator":
+			resourceDefs = append(resourceDefs, getSingleControlPlaneDef(component, replicaOptions.spValidator))
+		case "linkerd-proxy-injector":
+			resourceDefs = append(resourceDefs, getSingleControlPlaneDef(component, replicaOptions.proxyInjector))
+		case "linkerd-tap":
+			resourceDefs = append(resourceDefs, getSingleControlPlaneDef(component, replicaOptions.tap))
+		default:
+			log.Fatal("Could not find the resource")
```
zaharidichev: I think you can just do t.Fatal to avoid puling in the log dependency!

pkg/healthcheck/healthcheck_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
```
zaharidichev: no need to depend on that as there isnt really a need for logging

pkg/healthcheck/healthcheck_test.go
mayankshah1607: Just copied from k8s docs! :smile:

mayankshah1607:
@zaharidichev thanks! I've made the changes. Mind taking a look now?

PR #3947, opened 2020-01-17
attempt at a workaround for issues determining clock skew on k8s1.7

### LINKED ISSUE #3943: Clock skew check not working correctly
### What is the issue?
Linkerd --pre (randomly) fails on clock skew detection check

### How can it be reproduced?
If I were to guess why, use Kubernetes 17.0.0, made sure to check ntpd many times, also switched back and forth between ntpd and systemd-timesyncd

./clock_skew.sh is a lightly modified version of this script:
https://gist.github.com/foospidy/77c43ec2b2b71e4447f0534f7ac26798

### Logs, error output, etc
'proof'
```
/clock_skew.sh 
INFO: Clock skew on 0.0.0.27 is 0 seconds.
INFO: Clock skew on 0.0.0.51 is 1 seconds.
INFO: Clock skew on 0.0.0.161 is 0 seconds.
INFO: Clock skew on 0.0.0.113 is 0 seconds.
Everything is running on time!
```
simple capturing of node metadata
```
kubectl describe nodes | grep Ready | grep posting
  Ready                True    Fri, 17 Jan 2020 14:15:39 +0100   Fri, 17 Jan 2020 13:45:28 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:15:28 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:15:28 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:15:28 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
```

#### EDIT
Whilst verifying output for this bug report I saw a massive jump in heartbeat output
```text
 kubectl describe nodes | grep Ready | grep posting
  Ready                True    Fri, 17 Jan 2020 14:30:40 +0100   Fri, 17 Jan 2020 13:45:28 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:17 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:30:28 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:30:28 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
./clock_skew.sh 
INFO: Fri, 17 Jan 2020 14:37:53 +0100 vs Fri, 17 Jan 2020 13:37:54 +0000 Clock skew on 0.0.100.27 is 1.
INFO: Fri, 17 Jan 2020 14:37:53 +0100 vs Fri, 17 Jan 2020 13:37:54 +0000 Clock skew on 0.0.101.51 is 1.
INFO: Fri, 17 Jan 2020 14:37:54 +0100 vs Fri, 17 Jan 2020 13:37:54 +0000 Clock skew on 0.0.100.161 is 0.
INFO: Fri, 17 Jan 2020 14:37:54 +0100 vs Fri, 17 Jan 2020 13:37:54 +0000 Clock skew on 0.0.100.113 is 0.
Everything is running on time!
kubectl describe nodes | grep Ready | grep posting
  Ready                True    Fri, 17 Jan 2020 14:37:24 +0100   Fri, 17 Jan 2020 13:45:28 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:17 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:23 +0100   Fri, 17 Jan 2020 14:37:23 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:23 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
./clock_skew.sh                                   
INFO: Fri, 17 Jan 2020 14:38:13 +0100 vs Fri, 17 Jan 2020 13:38:14 +0000 Clock skew on 0.0.100.27 is 1.
INFO: Fri, 17 Jan 2020 14:38:13 +0100 vs Fri, 17 Jan 2020 13:38:13 +0000 Clock skew on 0.0.101.51 is 0.
INFO: Fri, 17 Jan 2020 14:38:14 +0100 vs Fri, 17 Jan 2020 13:38:13 +0000 Clock skew on 0.0.100.161 is -1.
INFO: Fri, 17 Jan 2020 14:38:14 +0100 vs Fri, 17 Jan 2020 13:38:14 +0000 Clock skew on 0.0.100.113 is 0.
Everything is running on time
```

And whilst investigating _that_ output it started lagging again
```
kubectl describe nodes | grep Ready | grep posting
  Ready                True    Fri, 17 Jan 2020 14:37:24 +0100   Fri, 17 Jan 2020 13:45:28 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:17 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:23 +0100   Fri, 17 Jan 2020 14:37:23 +0100   KubeletReady                 kubelet is posting ready status
  Ready                True    Fri, 17 Jan 2020 14:37:23 +0100   Fri, 17 Jan 2020 13:45:27 +0100   KubeletReady                 kubelet is posting ready status
./clock_skew.sh                                   
INFO: Fri, 17 Jan 2020 14:38:57 +0100 vs Fri, 17 Jan 2020 13:38:58 +0000 Clock skew on 0.0.100.27 is 1.
INFO: Fri, 17 Jan 2020 14:38:57 +0100 vs Fri, 17 Jan 2020 13:38:58 +0000 Clock skew on 0.0.101.51 is 1.
INFO: Fri, 17 Jan 2020 14:38:58 +0100 vs Fri, 17 Jan 2020 13:38:58 +0000 Clock skew on 0.0.100.161 is 0.
INFO: Fri, 17 Jan 2020 14:38:58 +0100 vs Fri, 17 Jan 2020 13:38:58 +0000 Clock skew on 0.0.100.113 is 0.
Everything is running on time
```

I'm using bootkube and couldn't find any override for the default heartbeat settings  is it possible the heartbeat output simply isn't reliable enough (any more) ?

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

pre-kubernetes-setup
--------------------
√ control plane namespace does not already exist
√ can create Namespaces
√ can create ClusterRoles
√ can create ClusterRoleBindings
√ can create CustomResourceDefinitions
√ can create PodSecurityPolicies
√ can create ServiceAccounts
√ can create Services
√ can create Deployments
√ can create CronJobs
√ can create ConfigMaps
× no clock skew detected
    clock skew detected for node(s): supernova-controller-0, supernova-helium-worker-0, supernova-helium-worker-1, supernova-helium-worker-2
    see https://linkerd.io/checks/#pre-k8s-clock-skew for hints

pre-kubernetes-capability
-------------------------
√ has NET_ADMIN capability
√ has NET_RAW capability

pre-linkerd-global-resources
----------------------------
√ no ClusterRoles exist
√ no ClusterRoleBindings exist
√ no CustomResourceDefinitions exist
√ no MutatingWebhookConfigurations exist
√ no ValidatingWebhookConfigurations exist
√ no PodSecurityPolicies exist

linkerd-version
---------------
‼ cli is up-to-date
    is running version 2.6.0 but the latest stable version is 2.6.1
```

### Environment

- Kubernetes Version: 1.17.0
- Cluster Environment: Packet (Bootkube)
- Host OS: Flatcar Linux
- Linkerd version: 2.6.1

### Possible solution
Checked the code (pkg/healthcheck/healthcheck.go), nothing stood out to me as to a possible cause of this. But maybe making it commandline overrideable would be nice. Now I have to ignore the check output in my terraform scripts instead

DGollings:
Subject

Problem
Clock skew not measurable in 1.7

Solution
use nodes themselves as a form of clock quorum

Validation
Added some tests

Fixes #3943

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -144,13 +144,23 @@ const (
 // page.
 const HintBaseURL = "https://linkerd.io/checks/#"
 
-// AllowedClockSkew sets the allowed skew in clock synchronization
+// AllowedNodeClockSkew sets the allowed skew in clock synchronization
 // between the system running inject command and the node(s), being
 // based on assumed node's heartbeat interval (<= 60 seconds) plus default TLS
 // clock skew allowance.
 //
 // TODO: Make this default value overridiable, e.g. by CLI flag
-const AllowedClockSkew = time.Minute + tls.DefaultClockSkewAllowance
+const AllowedNodeClockSkew = time.Minute + tls.DefaultClockSkewAllowance
+
+// AllowedTotalClockSkew sets the allowed skew in clock synchronization
+// between the system running inject command and the node(s), being
+// based on assumed node's heartbeat interval (<= 60 seconds) plus default TLS
+// clock skew allowance _plus_ keeping in account the delayed by design values retrieved
+// from the API.
+// https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/0009-node-heartbeat.md
+//
+// TODO: Make this default value overridiable, e.g. by CLI flag
+const AllowedTotalClockSkew = (5 * time.Minute) + tls.DefaultClockSkewAllowance
 
 var (
 	retryWindow    = 5 * time.Second
@@ -1691,27 +1701,51 @@ func (hc *HealthChecker) checkClockSkew() error {
 		return fmt.Errorf("unexpected error: Kubernetes ClientSet not initialized")
 	}
 
-	var clockSkewNodes []string
-
 	nodeList, err := hc.kubeAPI.CoreV1().Nodes().List(metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
 
+	// Map all the nodes we want to check
+	type _node struct {
+		Name      string
+		Timestamp time.Time
+	}
+	_nodes := make([]_node, 0)
+
 	for _, node := range nodeList.Items {
 		for _, condition := range node.Status.Conditions {
-			// we want to check only KubeletReady condition and only execute if the node is ready
 			if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
-				since := time.Since(condition.LastHeartbeatTime.Time)
-				if (since > AllowedClockSkew) || (since < -AllowedClockSkew) {
-					clockSkewNodes = append(clockSkewNodes, node.Name)
-				}
+				_nodes = append(_nodes, _node{Name: node.Name, Timestamp: condition.LastHeartbeatTime.Time})
 			}
 		}
 	}
+	// Dont continue if there's nothing to do
+	if len(_nodes) == 0 {
+		return nil
+	}
 
-	if len(clockSkewNodes) > 0 {
-		return fmt.Errorf("clock skew detected for node(s): %s", strings.Join(clockSkewNodes, ", "))
+	// Get node names
+	nodeNames := make([]string, len(_nodes))
+	for i, n := range _nodes {
+		nodeNames[i] = n.Name
+	}
+
+	// Sort by date
+	sort.Slice(_nodes, func(i, j int) bool { return _nodes[i].Timestamp.Before(_nodes[j].Timestamp) })
+
+	// Determine the skew range
+	maximumSkew := _nodes[0].Timestamp.Sub(_nodes[len(_nodes)-1].Timestamp)
+
+	// Throw error if more than AllowedClockSkew between individual nodes
+	if (maximumSkew > AllowedNodeClockSkew) || (maximumSkew < -AllowedNodeClockSkew) {
+		return fmt.Errorf("clock skew detected between node(s): %s", strings.Join(nodeNames, ", "))
+	}
+
+	// or large skew between Linkerd deployment host and last node
+	maximumSkew = time.Since(_nodes[len(_nodes)-1].Timestamp)
+	if (maximumSkew > AllowedTotalClockSkew) || (maximumSkew < -AllowedTotalClockSkew) {
+		return fmt.Errorf("large clock skew detected between your host and node(s): %s", strings.Join(nodeNames, ", "))
 	}
 
 	return nil
--- a/pkg/healthcheck/healthcheck_test.go
+++ b/pkg/healthcheck/healthcheck_test.go
@@ -414,7 +414,53 @@ status:
     status: "True"
     type: Ready`,
 			},
-			fmt.Errorf("clock skew detected for node(s): test-node"),
+			fmt.Errorf("large clock skew detected between your host and node(s): test-node"),
+		},
+		{
+			// Near identical times between nodes
+			[]string{fmt.Sprintf(`apiVersion: v1
+kind: Node
+metadata:
+  name: test-node
+status:
+  conditions:
+  - lastHeartbeatTime: "%s"
+    status: "True"
+    type: Ready`, time.Now().Format(time.RFC3339)),
+				fmt.Sprintf(`apiVersion: v1
+kind: Node
+metadata:
+  name: test-node2
+status:
+  conditions:
+  - lastHeartbeatTime: "%s"
+    status: "True"
+    type: Ready`, time.Now().Add(1*time.Second).Format(time.RFC3339)),
+			},
+			nil,
+		},
+		{
+			// Add large skew between nodes
+			[]string{fmt.Sprintf(`apiVersion: v1
+kind: Node
+metadata:
+  name: test-node
+status:
+  conditions:
+  - lastHeartbeatTime: "%s"
+    status: "True"
+    type: Ready`, time.Now().Format(time.RFC3339)),
+				fmt.Sprintf(`apiVersion: v1
+kind: Node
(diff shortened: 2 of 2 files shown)


DGollings:
Glad you like it, if this code is fine two more things could be looked at whilst integrating it all:
Documentation: its descriptive without being helpful, took me a very long time to figure out what could possibly be wrong. I assume just open a separate PR in website repo?
CLI: I assume it's not difficult to set, but would have to investigate.

zaharidichev:
@DGollings I think documentation is always helpful. And yes,  you can open a separate PR in the website repo (its probably worth to get that fully approved first, so you do not end up changing your docs PR too much). You can also add the CLI bits to this PR. Adding some more comments

adleong:
Thanks @DGollings, I just have some high level questions about this change.

What, exactly, is the situation that we're trying to guard against by doing a quorum check like this?  As I understand it, it's important that there isn't a high amount of clock skew between the CLI and the nodes because some of the TLS material is generated by the CLI and we want to make sure the expiry times are accurate.  Any high amount of skew between nodes would also manifest as skew between a node as the CLI.  So, I'm not sure what this change accomplishes.

This doesn't change the fact that we're using last heartbeat time as a proxy for node time.  Nodes may still have timestamps which vary from each other by as much as the heartbeat interval even when there is no real clock skew.

DGollings:
@adleong What you're saying is true, both pro and con. But by (new k8s) design we can't measure actual clock time (easily). This is a imho a simple way to do a sanity check for the most excessive skews, either between nodes (>1 min) or between nodes and installer (> 5 min). 

The current situation is that 4 out of 5 minutes `check --pre` will report an error. An error based on a time that can't be debugged by simply logging in and typing `date`. Throwing a warning 80% of the time isn't that much more helpful (can tell you its even more confusing). But completely disabling the check also isn't great. 

Basically, I'm suggesting a compromise. If the nodes say they're in sync it's very likely to be true. Should guard against say a completely misconfigured newly provisioned node if not a subtly out of sync node.

adleong:
Does Kubernetes provide any guarantees that heartbeat checks will be synchronized across nodes?  If not, then with a heartbeat interval of 5 minutes, the last heartbeat time could vary between nodes by up to 5 minutes.  So I don't think this compromise has the properties that we want.

That said, I agree that the current situation is a problem.  

Perhaps our best tool here is documentation.  If we made this check a warning, we could update the check docs to explain that if the amount of skew detected is less than the `node-monitor-period` then there's no problem.

DGollings:
I've only seen pretty synchronized nodes, but true. Would have to reboot one node mid heartbeat interval and see what happens.

Alternatively there's the event api that supposedly has an up to date heartbeat somewhere (every x seconds), but that's a lot more effort to figure out than this solution

-------- Original Message --------

grampelberg:
Just to suggest a potentially different tact. Can we address this post-install with logging and metrics? It is more important that folks know how to fix it than we keep it from blowing up.

DGollings:
I'm sure that would be possible, having some health endpoint also add a timestamp to its output (if it doesn't already) and comparing the various timestamps received. I don't know how `linkerd check` currently decides when a linkerd installation is successful but I;d guess adding the logic there would be a good spot.

I know for example that Cockroachdb (which needs to maintain quorum) simply starts crashing at > 500ms clock skew

It is however becoming more complex than a 'slightly more intelligent sanity check' :)

grampelberg:
Linkerd will fail to come up if the certs were generated incorrectly. The problem is that the error message gives you no way to figure out what's actually failing.

adleong:
The current behavior if the issuer certificate is out of the validity window due to clock skew is that the linkerd-identity controller's logs fill up with this error message:

```
time="2020-01-23T21:12:23Z" level=error msg="could not process CSR because of CA cert validation failure: x509: certificate has expired or is not yet valid"
```

Possible things we could do to make the problem more diagnose-able:
* Put the expiry time in the error message
* Put the "current" time in the error message
* Emit the error message as a kubernetes event

adleong:
@DGollings based on the discussion here, I think the best course of action is to close this PR and open a new one that relaxes the clock skew checks as described in https://github.com/linkerd/linkerd2/issues/3943#issuecomment-578988545

Are you still interested in working on that?

adleong:
I've also opened https://github.com/linkerd/linkerd2/issues/3980 to make certificate errors easier to diagnose when clock skew exists.

Issue #3459, opened 2019-09-22
Tap pod should consider subjectAltNames


balchua:
When verifying for `requestheader-allowed-names` in the tap apiserver consider `subjectAltNames` for cases when there are multiple names.

grampelberg:
Could you explain a little more about the situation where there are multiple names in the SAN?

balchua:
Hi sorry didn't give a bit more info here.  

For example i have `othername` in the SAN and i add it to the kube apiserver option `requestheader-allowed-names=commonname,othername` the tap pod doesn't recognize it.

Because it only checks for the allowed names in the CN.

https://github.com/linkerd/linkerd2/blob/5e51208b5deea360932bd294337683f51a4571eb/controller/tap/apiserver.go#L98

drholmie:
Hey there! Is anyone working on this? I'm interested in helping out.

wmorgan:
@drholmie I think it's all yours!

drholmie:
Awesome. Thanks for the go ahead!

javaducky:
Hope it's alright, but I noticed there hadn't been any movement on this issue for a while.  Figured I could knock it out quickly, so I threw a PR together.

grampelberg:
Thank you @javaducky!

PR #3949, opened 2020-01-18
Utilize Common Name or Subject Alternate Name for access checks (#3459)

javaducky:
**Subject**
Utilize Common Name or Subject Alternate Name for access checks (#3459)

**Problem**
When access restrictions to API server have been enabled with the `requestheader-allowed-names` configuration, only the _Common Name_ of the requestor certificate is being checked.  This check should include the use of _Subject Alternate Name_ attributes.

**Solution**
API server will now check the SAN attributes (DNS Names, Email Addresses, IP Addresses, and URIs) when determining accessibility for allowed names.

Fixes issue #3459

### DIFF
--- a/controller/tap/apiserver.go
+++ b/controller/tap/apiserver.go
@@ -95,8 +95,9 @@ func (a *apiServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		for _, cn := range a.allowedNames {
 			for _, clientCert := range req.TLS.PeerCertificates {
 				clientNames = append(clientNames, clientCert.Subject.CommonName)
-				if cn == clientCert.Subject.CommonName {
-					validCN = clientCert.Subject.CommonName
+				// Check Common Name and Subject Alternate Name(s)
+				if cn == clientCert.Subject.CommonName || isSubjectAlternateName(clientCert, cn) {
+					validCN = cn
 					break
 				}
 			}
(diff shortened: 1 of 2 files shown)


controller/tap/apiserver.go
```diff
@@ -95,8 +95,9 @@ func (a *apiServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		for _, cn := range a.allowedNames {
 			for _, clientCert := range req.TLS.PeerCertificates {
 				clientNames = append(clientNames, clientCert.Subject.CommonName)
-				if cn == clientCert.Subject.CommonName {
-					validCN = clientCert.Subject.CommonName
+				// Check Common Name and Subject Alternate Name(s)
+				if cn == clientCert.Subject.CommonName || isSubjectAlternateName(clientCert, cn) {
+					validCN = cn
```
zaharidichev: I wonder why is validCN being a string while it seems to express a bool only information. Can we change that to a bool (i.e. validCNFound) or just restructure the function?

javaducky: I guess I didn't really give it much of a look beyond including my changes! 😄 Looking at it now however, there definitely seems to be no reason to treat `validCN` as a string.  I'll restructure the method to treat that as a boolean...thanks for the suggestion!

controller/tap/apiserver.go
```diff
@@ -87,32 +87,35 @@ func NewAPIServer(
...
 
+// validate ensures that the request should be honored returning an error otherwise.
+func (a *apiServer) validate(req *http.Request) error {
 	// if `requestheader-allowed-names` was empty, allow any CN
 	if len(a.allowedNames) > 0 {
-		validCN := ""
-		clientNames := []string{}
 		for _, cn := range a.allowedNames {
 			for _, clientCert := range req.TLS.PeerCertificates {
-				clientNames = append(clientNames, clientCert.Subject.CommonName)
```
javaducky: This would duplicate each common name in the slice for each additional `cn`, so I refactored the listing of the CNs to only be done if and error will returned.

controller/tap/apiserver.go
```diff
@@ -87,32 +87,35 @@ func NewAPIServer(
 // ServeHTTP handles all routes for the APIServer.
 func (a *apiServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	a.log.Debugf("ServeHTTP(): %+v", req)
+	err := a.validate(req)
```
javaducky: Refactored the previously in-line code block to enable unit testing.

controller/tap/apiserver.go
```diff
@@ -87,32 +87,35 @@ func NewAPIServer(
...
+	if err != nil {
+		a.log.Debug(err)
+		renderJSONError(w, err, http.StatusBadRequest)
+	} else {
+		a.router.ServeHTTP(w, req)
+	}
```
zaharidichev: Dont you think since you have only an error you can use a more concise construct such as: 
```
	if err := a.validate(req); err != nil {
		a.log.Debug(err)
		renderJSONError(w, err, http.StatusBadRequest)
	}
	a.router.ServeHTTP(w, req)
```

javaducky: Updated.

PR #3951, opened 2020-01-18
Update TrafficSplit API to v1alpha2

### LINKED ISSUE #3950: Update the SMI TrafficSplit  API version to v1alpha2
In the TrafficSplit v1alpha2 the weight type was changed from `resource.Quantity` to `int` see https://github.com/deislabs/smi-sdk-go/pull/24. 

Linkerd should use the v1alpha2 API internally and the Helm chart and install YAMLs should deploy the v1alpha2 CRD.

stefanprodan:
This PR updates the TrafficSplit API to v1alpha2.

Changes:
- update Go to 1.13.5 in go mod
- update SMI package in go mod to `github.com/deislabs/smi-sdk-go v0.2.0`
- run `go mod tidy`
- pin `github.com/Azure/go-autorest v12.2.0+incompatible` to deal with autorest ambiguous import
- run `bin/update-go-deps-shas` to update dep images
- change CRD YAMLs and run `go test ./cli/cmd/... --update`
- change imports from `split/v1alpha1` to `split/v1alpha2` and code refs
- change `WeightedDst.Weight` type to int and update tests
- use a go mod compatible code-generator (Kubernetes 1.16)
- change the code-generator update script to support any location
- regenerate clientset
- add codegen verify script to CI so that the build will fail if someone changes the CRD without regenerating the clientset

Fixes #3950

Signed-off-by: stefanprodan <stefan.prodan@gmail.com>


grampelberg:
Super helpful, thank you!

stefanprodan:
Can you help me out fix the proto comparison in  stat_summary_test? I can't understand how come is `leaf:"service-2" weight:"\n"`.

controller/api/public/stat_summary.go
```diff
@@ -303,7 +303,7 @@ func (s *grpcServer) trafficSplitResourceQuery(ctx context.Context, req *pb.Stat
 
 		for _, backend := range backends {
 			name := backend.Service
-			weight := backend.Weight.String()
+			weight := string(backend.Weight)
```
adleong: I think this needs to be `weight := fmt.Sprintf("%d", backend.Weight)`, otherwise it will attempt to convert the Weight as an ascii value.

stefanprodan: Ahh thanks Alex, that fixed the tests!

stefanprodan: Ahh thanks Alex, that fixed the tests!

stefanprodan:
@adleong using fmt fixed the tests, thank you! I've updated the build docs regarding codegen.

adleong:
@stefanprodan thanks for this!  I need to spend a bit more time manually testing and thinking about how this interacts with the upgrade process and backwards compatibility and so on.  Thanks for your patience.

stefanprodan:
@adleong you may want to talk with @grampelberg about #3962

adleong:
I don't think we can accept this as-is because of the backward incompatibility.  I've taken a stab at writing a migration guide at https://github.com/linkerd/linkerd2/issues/3962.  Let me know what you think.

stefanprodan:
I hate to throw away hours of work but to implement the upgrade you described means writing a wrapper over the informer package and handle the conversion before the traffic split object is stored in the cache. I don't have bandwidth for implementing the informer stuff, this could come as a followup PR or you can close my PR.

adleong:
Closing this for now until we have time to implement #3962

Issue #3794, opened 2019-12-06
Linkerd Add-Ons


Pothulapati:
# Linkerd Add-Ons

## Goals

- To Have a consistent exerience to install stuff along with Linkerd, that add/extend functionality.
- To provide an out of box experience, with limited Knobs for popular use-cases.

## Non-Goals

- To have a scalable/maintanable experience for the add-on.
- To allow full blown configurable experience for the add-on.

All of the above cases, can always be done by using the official helm-charts/operators of those components.

## Example Add-ons

- Jaeger and OC-Collector as add-on for tracing to show the tracing functionality.
- k8s-prometheus adapter, relevant config crds as an add-on to show the Scaling based on latency use-case.
- Grafana can also be made as an add-on but a default one.
- OPA can also be configured as an add-on for policies of resources.
- SMI Metrics API Server as an add-on.

## Why not use Helm Charts as dependencies?

This can remove the hassle of having them in our code-base but introduces the following problems

- Proxy Injection into the components would be hard, as the dependency installation would happen along with Control plane.
- Gives us very limited experience to tweak configuration, and makes it very hard to configure common data across charts.

## Procedure

- There will be add-ons in `charts/add-ons/tracing`, where each chart has its own templating, values and structure.
- All charts will be mentoned as dependencies, to `linkerd`.
- Add-ons will be optional using [conditional field](https://v2.helm.sh/docs/developing_charts/ #tags-and-condition-fields-in-requirements-yaml) `global.Add-ons.<add-on>.Enabled`.
- The same `global.Add-ons.<add-on>.Enabled` can be used to toggle stuff in the `linkerd` parent chart based on the component is present or not. Ex `--control-plane-trace` flag in control-plane components can be toggled based on `global.Add-ons.tracing.Enabled`
- Each Sub-Chart dosen't have to depend on `partials` again, as `Linkerd` depends on that, and helm would make that automatically available for other siblings charts i.e `add-ons` in this case.
- Each Sub-Chart can be made configurable by using `global.Add-ons.<add-on>.property` rather than using it's own `Values.yaml`, to make configuration also available using `linkerd install`.
- Values can be passed between linkerd and sub-charts by:
        
    - `Linkerd` -> `Add-On` :
        - Used to send common properties across all sub-charts like `.Proxy`, `.Namespace`, etc.
        - This can be done by using `.Values.global.<name>`
        
 
    - `Add-On` -> `Linkerd` :
[...]

PR #3955, opened 2020-01-20
Tracing Add-on For Linkerd

Pothulapati:
Follows #3794 

This PR adds the tracing components as a Linkerd Add-On. The add-on model follows the same consistent approach both through Helm and Linkerd CLI.

The following changes are part of this PR:
- **Pre-Requisite Changes**:
   - Moved some common yaml templates like `_nodeselectors.yaml`,etc required by multiple sub-charts to `partials` for re-use. Also updating the relevant helm tests.
- **Helm Changes**:
  - A Tracing sub-chart is added with `partials` dependency (for injection). This sub-chart is also added as an optional dependency for `linkerd2` chart.
  - `Values.yaml` has been updated to include the tracing add-on config.
  -  `linkerd-values` cm has been added which stores the add-on configurations, and will only be present when there is at-least one Add-On.
- **Install Path**:
   - Updates  `installOptions` to include a `config` option, through which the configuration for the add-ons is given to the CLI.
   -  During Install, A final step has been added to read config (if given), and the default Values will be overwritten by this.
   -  If there are any add-ons present, those sub-charts are loaded and rendered in this step, by passing both `.Values.global` and the add-on values to the sub-chart.
- **Upgrade Path**:
  - The upgrade path in the final stage, now first checks if the `linkerd-values` cm exists, and updates the default addon values with the values present in `linkerd-values`, this is done to have persistent config upgrades.
   - It then checks if there is a `config` option present, and overwrite the default values further with the options from the config file. This allows users to change configuration of add-ons during upgrades.
 
The logic to Overwrite the `values` struct, is done by this fork of mergo at https://github.com/pothulapati/mergo, because of this [issue/decision](https://github.com/imdario/mergo/issues/24) from the official project.

**To try `tracing` as an add-on:**
- Create a new config file as `config.yaml`
```yaml
tracing:
  enabled: true
```
- Now Run, 
`linkerd install --config config.yaml --control-plane-tracing | kubectl apply -f -` 
and see the magic happen ;)

@ihcsim @alpeb @zaharidichev @grampelberg


Pothulapati:
**Process required to add new Add-Ons:**
- A sub-chart has to be created first at `/charts/linkerd2/add-ons/<add-on>`, and added as a dependency to the `linkerd2` sub-chart, with the condition field as `<add-on>.enabled` from the `linkerd2/values.yaml`. The `linkerd2/values.yaml` will be the file where the default config for add-ons will also be present.
- The sub-chart can optionally take a dependency on `partials` to have proxy injection by including the necessary templates.
- The `Values` struct in `pkg/charts/values.go` should have the relevant field for add-on, with the same structure as that of the one present in `Values.yaml`.
- The relevant add-on config struct will added at `pkg/charts/linkerd2/<add-on>.go`, and should also implement the `AddOn` interface, as that logic is used in the `install` and `upgrade` paths to render the add-on templates.
- Finally, The `mergeAddOnValues` and `checkAddOns` functions in `install.go` should be updated to check for the relevant newly added add-ons.

Pothulapati:
Sorry, Did not realize there was a small error with `linkerd-config` causing `identity` pod to fail, 
Fixed it.

Ready to review, now :)

pkg/charts/linkerd2/tracing.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	for _, template := range tracingControlPlaneStage {
+		files = append(files, &chartutil.BufferedFile{
+			Name: template,
+		})
+	}
+
+	return files
+}
```
alpeb: In order to avoid repeating this function into each add-on, WDYT about embedding into the `Tracing` struct another struct `BaseAddon` declared in `add-on.go` that would implement `GetFiles(configStage, controlPlaneStage []string) []*chartutil.BufferedFile` and so `GetFiles` can just be:
```go
func (tr *Tracing) GetFiles() []*chartutil.BufferedFile {
    return tr.BaseAddon.GetFiles(tracingConfigStage, tracingControlPlaneStage)
}
```

cli/cmd/install.go
```diff
@@ -1108,3 +1168,25 @@ func toIdentityContext(idvals *identityWithAnchorsAndTrustDomain) *pb.IdentityCo
...
+func mergeAddonValues(values, addonValues *l5dcharts.Values) error {
+
+	if addonValues.Tracing != nil {
+		if err := mergo.Merge(addonValues.Tracing, values.Tracing); err != nil {
+			return err
+		}
+		values.Tracing = addonValues.Tracing
+	}
+	return nil
+}
```
alpeb: Why not having the new `--config` flag serve as overrides for all the values, not just for add-ons? And if so, it might be clearer to implement that as a preliminary PR...

charts/linkerd2/templates/linkerd-values.yaml
```diff
@@ -0,0 +1,33 @@
+{{ if or (.Values.tracing.enabled) -}}
```
alpeb: Seems like the `or` here is not needed?
Also if tracing is not enabled then `$dupValues` won't be defined. Will that cause a problem down at the end of this file?

Pothulapati: This if condition is to not have `linkerd-values` cm to be deployed, when there are zero add-ons enabled. Hence, the `or` condition but once we make the `linkerd-values` common for all config, then we should be able to remove this if condition.

cli/cmd/install.go
```diff
@@ -19,6 +20,7 @@ import (
 	consts "github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/tls"
 	"github.com/linkerd/linkerd2/pkg/version"
+	"github.com/pothulapati/mergo"
```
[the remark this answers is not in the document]
Pothulapati: The custom transformer logic requires us to create a new type which looks like bool, and make it be used everywhere as we can't add transformers for native types. 
This would impose a constraint on users to use the same which would be a hard requirement :\

pkg/charts/linkerd2/tracing.go
Pothulapati: Adding a field to the `tracing` struct, would unnecessarily add more complexity as we then decode that into a yaml for `linkerd-values`. Instead I added a `defaultGetFiles` function with your suggested signature :)

cli/cmd/install.go
Pothulapati: The main reason is that, some fields in `values` like `config` etc are generated way before in the install/upgrade path but as we override them with config last in the pipeline, this would require us to again redo those values. 

Sure, we can keep this overwrite logic in the start but that would require a lot of code-changes. So, the plan is to do that next but keep it out of this PR. WDYT?

Pothulapati:
@alpeb Updated `gitignore` for the same :) Thanks
reg sub-charts in `linkerd2` 
There is no particular problem regarding that, but my opinion was that, as the sub-charts can't be re-used individually i.e without `linkerd2` parent chart, It made sense to be inside. :) but I can move it if needed.

cli/cmd/install.go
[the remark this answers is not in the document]
alpeb: Understood. We'll have to move your fork into the linkerd org, like we did for the Stern fork.

cli/cmd/install.go
alpeb: I'm not sure I follow. Doesn't this overriding logic imply that only the values present in the file provided through `--config` will get overridden, leaving all the others untouched? Why would it need to regenerate those other values? Maybe an example might help me understand :slightly_smiling_face:

Pothulapati: Ah, Sorry for not being clear.
My main concern is with this, as it is built by combining various values and is added in the `linkerd-config` cm, we will have to regenerate that agian after the override this. We could regenrate that again if it was based on `Values` struct but currently it is based on `identityInstallOptions`.
 https://github.com/linkerd/linkerd2/blob/5fc83bc1c1b81487172364548a46be80618aa4ad/pkg/charts/linkerd2/values.go#L41

.gitignore
```diff
@@ -13,3 +13,4 @@ web/app/yarn-error.log
 **/*.gogen*
 **/*.swp
 charts/*/charts/
+charts/*/add-ons/*/charts
```
alpeb: I think we can replace these two lines with
```
charts/**/charts/*
```

Pothulapati: Done.

cli/cmd/install.go
alpeb: Gotcha. Another reason to refactor this at some point in the future  so we have to rely less on the structs in install.go and leverage those in values.go more directly :wink:

charts/linkerd2/templates/linkerd-values.yaml
```diff
@@ -0,0 +1,33 @@
...
+{{- $dupValues := .Values -}}
+{{- $dupValues := omit $dupValues "partials" -}}
```
alpeb: Is this really needed? If I omit this line I still get the same result.
Could you also prepend a comment to this section to clarify what's going on?

Pothulapati: Yep Correct, It was a previous change that came it. Removed now.

Pothulapati:
> Neither do we publish the "partials" nor "patch" charts, yet they reside under /charts.

Yep, :)  Moved them to `charts` now.

Pothulapati:
@alpeb Reused similar logic, for multi-stage installation of add-ons. Also, Moved the `config` flag to `allStageFlagSet` as it would be needed for both the stages for add-ons too i.e apply `rbac`,etc during `config` and others during `control-plane`. 

Now, Multi-Stage Installation will work with add-ons.

I also added a install test with `tracing` add-on.

cli/cmd/install.go
```diff
@@ -420,6 +421,12 @@ func (options *installOptions) validateAndBuild(stage string, flags *pflag.FlagS
 	values.Global.IdentityTrustDomain = identityValues.TrustDomain
 	values.Stage = stage
 
+	// Update Configuration of Add-ons from config file
```
ihcsim: Can we leave this and the `--config` flag out for now? I understand we need a way to override addon configuration without bloating the CLI flags. I want to better understand this is an addon-only flag, or will it become something more general like Helm's `-f` option. I am also a bit hesitant about the fork repo.

For now, I think we can just use the addon defaults defined in the `values.yaml`, right?

pkg/charts/linkerd2/add-on.go
```diff
@@ -0,0 +1,29 @@
...
+
+import "k8s.io/helm/pkg/chartutil"
+
+var (
+	// AddOnChartsPath is where the linkerd2 add-ons will be present
+	AddOnChartsPath = "add-ons/"
+)
+
+// AddOn interface consists of the common functions required by add-ons to be implemented
+type AddOn interface {
```
ihcsim: Per previous slack convo, I am hesitant about this approach where new addons are required to implement this interface. Can these values be added to the `tracing` subtree in `values.yaml`, which the CLI is already parsing? IIUC, these config are just static chart name and filepaths.

pkg/charts/linkerd2/tracing.go
```diff
@@ -0,0 +1,60 @@
...
+	tracingConfigStage = []string{
+		"templates/tracing-rbac.yaml",
+	}
+
+	tracingControlPlaneStage = []string{
+		"templates/tracing.yaml",
+	}
+)
+
+// GetChartName returns the name of the add-on sub-chart
```
ihcsim: All of these look like static values that can be read from the `tracing` subtree in `values.yaml`.

cli/cmd/install.go
ihcsim: I recommend excluding this change from this PR for now. It feels like a more complicated problem than we first thought. The addon will still work without the `--config` flag, right? It will be restricted to just defaults in the `values.yaml`.

charts/add-ons/tracing/templates/tracing-rbac.yaml
```diff
@@ -0,0 +1,26 @@
+---
+###
+### linkerd-collector RBAC
+###
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-collector
+  namespace: linkerd
```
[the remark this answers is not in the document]
Pothulapati: updated.

charts/add-ons/tracing/templates/tracing.yaml
```diff
@@ -0,0 +1,223 @@
...
+### Tracing Collector Service
+###
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{printf "%s-config" .Values.collector.name}}
+  namespace: {{.Values.global.namespace}}
+  labels:
+    {{.Values.global.controllerComponentLabel}}: {{.Values.collector.name}}
```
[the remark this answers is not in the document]
Pothulapati: So, For parents chart to supply charts to sub-charts, they just have to mention the sub-chart name `tracing:` in parent's `Values.yaml` but we don't have to mention the same field again in sub-charts, as only the fields that match the chart-name field are sent to sub-charts. 
More on this https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

charts/linkerd2/templates/linkerd-values.yaml
```diff
@@ -0,0 +1,35 @@
+{{ if or (.Values.tracing.enabled) -}}
+{{- /*
```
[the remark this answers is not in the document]
Pothulapati: The problem is that, Now Imagine the same logic being put by each and every add-on into the `linkerd-values`, It would definitely cause a lot of consistency problems and a lot of unnecessary work into `linkerd-values`. We just want to have a way to get the add-on fields into `linkerd-values`

If we just go with the `duplicated` logic, then
And the installation is through `CLI` you only get the `global` once, but if it is through helm, for each add-on you get both `partials` and `global` values and each `partial` will again have its own `global` (This is the helm way of sending values to sub-charts). It makes the whole `linkerd-values` cm a mess.

However we still have to remove the Private Keys present in `global`, thus requiring processing again.
Example gist [here](https://gist.github.com/Pothulapati/f8dde2d31318f69f0e2b58e5b1d81090) for multiple add-ons  with helm install

cli/cmd/install.go
Pothulapati: Yep, Right Now, as per the code too, we only overwrite the config of add-ons, As users would want a way to toggle the add-on on/off. Without the config flag, Toggle would require each add-on having its own flag which can add unnecessary work.

And also users would definitely want a way to overwrite fields like resouce limits, values, etc without having to change templates.

The next work, would be to make the add-on cm have linkerd2 values too. 

Yep, I did not like forking the repo, but thats the only solution I came across. :\

pkg/charts/linkerd2/add-on.go
Pothulapati: Yep, We are getting the same values from `Values.yaml` but to parse them from `values.yaml` and create fields would need reflection? and I'm not sure if its a error-free way to do this. Having strict types enforced would not only help us with overwriting, and linkerd-values but also makes it error-free (which can be a problem with parsing.)

cli/cmd/install.go
Pothulapati: I can definitely try something here, if its a problem.

charts/linkerd2/templates/linkerd-values.yaml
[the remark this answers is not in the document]
ihcsim: My main concern is that this block of go template code is not unit testable, and debugging it is going to be a pain. Can you share the approach you have been using to debug this?

> The problem is that, Now Imagine the same logic being put by each and every add-on into the linkerd-values, 

With or without the go template code, we will still need to add a YAML block for each new addon components. I assume it's just going to be like:
```
data:
  values: |-
    tracing:
      collector:
  {{- toYaml $.Values.tracing.collector | trim | nindent 8}}
      jaeger:
  {{- toYaml $.Values.tracing.collector | trim | nindent 8}}
    monitoring:
    grafana:
  {{- toYaml $.Values.monitoring.grafana | trim | nindent 8}}
    prometheus:
  {{- toYaml $.Values.monitoring.prometheus | trim | nindent 8}}
```
Do you foresee something more complicated?

>  but if it is through helm, for each add-on you get both partials and global values and each partial will again have its own global 

I only see one `global` field when tested with `helm install` and `helm template`. See YAML ]here](https://gist.github.com/Pothulapati/f8dde2d31318f69f0e2b58e5b1d81090#gistcomment-3174836).

> It would definitely cause a lot of consistency problems and a lot of unnecessary work into linkerd-values. We just want to have a way to get the add-on fields into linkerd-values

Do you foresee anyone wanting to manually change this configmap? Or will they prefer to use the `--config` option?

> However we still have to remove the Private Keys present in global, thus requiring processing again.

You mean the public trust root?

pkg/charts/linkerd2/add-on.go
ihcsim: I'm hoping it can be simply:
```
tracing:
  enabled: true
  chart:
    name: tracing
    templates:
       configStage: ["templates/tracing-rbac.yaml"]
       controlPlaneStage: ["templates/tracing.yaml"]
   ...
```
Then update your [`Tracing`](https://github.com/linkerd/linkerd2/blob/d8bb871d4df0af08792fec375b35523be9d0b32b/pkg/charts/linkerd2/tracing.go#L9-L14) struct to contain these fields. And in `install.go`, they should be available in your [`values`](https://github.com/linkerd/linkerd2/blob/d8bb871d4df0af08792fec375b35523be9d0b32b/cli/cmd/install.go#L804-L807) struct.

If it works, it feels cleaner than hardcoding these values in the Go code.

ihcsim: Per our slack convo, we should be able to remove the `AddOn` interface by re-using some of the existing functions in the Helm library.

E.g., I am able to use `chartutil.Load()` to load the entire chart and all its dependencies metadata:
```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 752ea734..24807144 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -783,6 +783,17 @@ func render(w io.Writer, values *l5dcharts.Values) error {
 		}
 	}
 
+	mychart, err := chartutil.Load("charts/" + helmDefaultChartDir)
+	if err != nil {
+		return err
+	}
+
+	for _, d := range mychart.Dependencies {
+		for _, t := range d.Templates {
+			fmt.Printf("%s/%s\n", d.Metadata.Name, t.Name)
+		}
+	}
+
 	addons := checkAddons(values)
 
 	chart := &charts.Chart{
```

The output looks like:
```
partials/templates/NOTES.txt
partials/templates/_affinity.tpl
partials/templates/_capabilities.tpl
partials/templates/_debug.tpl
partials/templates/_helpers.tpl
partials/templates/_metadata.tpl
partials/templates/_nodeselector.tpl
partials/templates/_proxy-init.tpl
partials/templates/_proxy.tpl
partials/templates/_resources.tpl
partials/templates/_trace.tpl
partials/templates/_validate.tpl
partials/templates/_volumes.tpl
tracing/templates/tracing-rbac.yaml  # tracing's chart name + template file
tracing/templates/tracing.yaml  # tracing's chart name + template file
```

Does this give us what we want?

cli/cmd/install.go
ihcsim: > Without the config flag, Toggle would require each add-on having its own flag which can add unnecessary work.

Just use Helm to install addon for now. I picture we will tackle the config PR immediately after this PR.

Basically, there are 3 things I want to know:

1. Will this flag be used only for addon configs? If yes, its name should reflect that.
1. How does a user know what is the supported content of this file? Technically, they can look up the Linkerd's chart `values.yaml` in the repo. Even then, how do they know which of those variables are supported? Can we do better?
1. The logic used to merge this file with the addons' defaults defined in `values.yaml` should be unit tested.

ihcsim: > Yep, I did not like forking the repo, but thats the only solution I came across

Can you take a look to see if [`chartutil.Values.MergeInto()`](https://pkg.go.dev/k8s.io/helm@v2.16.3+incompatible/pkg/chartutil?tab=doc#Values.MergeInto) will do what you want?

pkg/charts/linkerd2/add-on.go
Pothulapati: Yep, Updated now

cli/cmd/install.go
Pothulapati: Yep, It works.
Thanks :)

pkg/charts/linkerd2/add-on.go
```diff
@@ -0,0 +1,6 @@
+package linkerd2
+
+var (
+	// AddOnChartsPath is where the linkerd2 add-ons will be present
+	AddOnChartsPath = "add-ons/"
```
ihcsim: Let's move this to the `cli` package, since it's only used there.

Pothulapati:
@ihcsim Thanks for the help :)

The tests were still failing in the Github Checks even after the changes you mentioned, because the `load` method reads the dependencies from the `charts` directory, which is ignored in the repo. So, It could not find the relevant add-on, thus failing the  install `tracing` add-on test.

To Fix this, I had to use one more method which reads dependencies from the `requirements.yaml` metadata and then use that data to get files from `charts/add-ons/<add-on>/templates`. WDYT?

Pothulapati:
@ihcsim Now the Integration tests for `check` fail, can't they find the `charts` in the same vfs that install path is able to find? O.o

alpeb:
@Pothulapati Unlike the unit tests, the integration tests build the CLI binary and then use it in isolation, so there's no access to the source files, which rightfully tests that the VFS is working fine.
`linkerd check` is the first command used in the integration tests, but you'll also see that `linkerd install` is also complaining.
You can try looking at `chartutil.Load`'s source code (which access the filesystem directly), and see at what point you can use the VFS. I see there's a call to `LoadArchive` down the call stack which receives a `io.Reader`, so maybe you can plug-in the VFS in there...

Pothulapati:
So, I updated the code to hard code add-ons so that we can avoid the loading problem. @ihcsim suggested another way of loading all the files before like a constructor. I can do this, if this way is not suggested.

I also removed the need for add-ons to have their full static types declared. Now, Add-ons can just use `map[string]interface{}` as their type and not declare any sub-fields. We use reflection internally and get the `enabled` flag and toggle based on that. 

I've created another branch to see how this code works for multiple add-ons by moving grafana as an add-on at https://github.com/Pothulapati/linkerd2/tree/multi-addons . The changes required to just add grafana as an add-on can be seen at https://github.com/Pothulapati/linkerd2/commit/cac6d41b56fc19317fa8bbe58088565ea4f5e9cb

@ihcsim @alpeb

ihcsim:
@Pothulapati Thanks for your effort with helping to minimize the tech debt in the code. Per our slack convo, I managed to get this to work by loading the chart from the VFS, into the Helm's `LoadFiles()` function, which can be used to create a "native" `Chart` instance. Take a look at this [diff](https://gist.github.com/ihcsim/b1c3b3621c717c597ef89e5c89314973). It works with your code change, while passing the unit and integration tests. I also performed some manual tests on `install`, `inject`, `check` and `upgrade`.

I am hoping this will ensure that no further code change is required, when new addons are added.

Regarding the `--config` flag, to re-iterate what we discussed earlier, let's rename it to `--addon-config` to reflect its scope (i.e. addons-only configuration). Eventually, I will like us to update its description to include something like, _"For a list of supported configuration, see https://linkerd.io/references/addon-configuration"_, which we can extend `linkerd doc` to auto-generate. We can create an issue against the `website` repo.

Also, please simplify the `linkerd-values` configmap YAML, by removing the nested go template conditionals. Those code is hard to change and debug. If we move the addon's "application configuration" to the 2nd level (e.g. `.Values.tracing.collector`, `.Values.grafana.config`, `.Values.prometheus.config` etc.), then we can avoid those duplicated variables auto-injected by Helm.

cli/cmd/install.go
```diff
@@ -542,6 +560,53 @@ func (options *installOptions) installPersistentFlagSet() *pflag.FlagSet {
 	return flags
 }
 
+// UpdateValuesFromConfig takes a values struct and updates its add-on values from the config installOption
+func (options *installOptions) UpdateAddOnValuesFromConfig(values *l5dcharts.Values) error {
```
alpeb: Can you please update the function name in the comment? :mag:

cli/cmd/install.go
```diff
@@ -542,6 +560,53 @@ func (options *installOptions) installPersistentFlagSet() *pflag.FlagSet {
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = yaml.Unmarshal(b, &bMap)
+	if err != nil {
+		return nil, err
+	}
+
+	chartutil.Values.MergeInto(aMap, bMap)
```
alpeb: This threw me off a bit, I had forgotten about this syntax in go :rofl: 
WDYT about using a more common idiom like this?
```go
var aMap, bMap chartutil.Values
...
aMap.MergeInto(bMap)
```

cli/cmd/install.go
```diff
@@ -759,6 +825,48 @@ func render(w io.Writer, values *l5dcharts.Values) error {
...
+
+	// Render for each add-on separately and attach
+	for _, dep := range dependencies {
+		chartName := dep.Metadata.Name
+		if chartName == "partials" {
+			continue
+		}
+
+		addOnValues, enabled := checkAddon(values, chartName)
+		if enabled {
```
alpeb: Can be simplified with
```go
if addOnValues, enabled := checkAddon(values, chartName); enabled {
...
}
```

cli/cmd/install.go
```diff
@@ -1108,3 +1216,18 @@ func toIdentityContext(idvals *identityWithAnchorsAndTrustDomain) *pb.IdentityCo
...
+	r := reflect.ValueOf(values)
+
+	if !reflect.Indirect(r).FieldByName(strings.Title(name)).IsNil() {
+		if reflect.Indirect(r).FieldByName(strings.Title(name)).MapIndex(reflect.ValueOf("enabled")).Interface().(bool) {
+			values, err := yaml.Marshal(reflect.Indirect(r).FieldByName(strings.Title(name)).Interface())
+			if err != nil {
+				return nil, false
+			}
+			return values, true
+		}
```
alpeb: if the addon config is missing then the value of `reflect.Indirect(r).FieldByName(strings.Title(name))` would be an empty map (zero value) to which you can safely apply `MapIndex`, so no need of the nested `if` nor the `IsNil()` check. 

You can also extract the common part of the remaining condition:
```go
		config := reflect.Indirect(r).FieldByName(strings.Title(name)); 
                if config.MapIndex(reflect.ValueOf("enabled")).Interface().(bool) {
			values, err := yaml.Marshal(config.Interface())
			if err != nil {
				return nil, false
			}
			return values, true
		}
```

pkg/charts/charts.go
```diff
@@ -24,6 +32,106 @@ type Chart struct {
...
+// It relies on LoadChart to load the files and metadata of the chart from the
+// VFS.
+func LoadDependencies(chartName string) ([]*chart.Chart, error) {
+	chart, err := LoadChart(chartName)
+	if err != nil {
+		return nil, err
+	}
+
+	return chart.Dependencies, nil
+}
```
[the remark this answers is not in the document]
ihcsim: In the patch I sent to @Pothulapati, I didn't want to call `LoadChart()` in `install.go` directly, to avoid confusion with the existing `l5dchart.Chart` struct.

pkg/charts/charts_test.go
```diff
@@ -0,0 +1,92 @@
...
+		t.Fatal("unexpected error: ", err)
+	}
+
+	expected, err := chartutil.Load(testChartRootDir)
+	if err != nil {
+		t.Fatal("unexpected error: ", err)
+	}
+
+	// compare the charts' metadata
+	if !reflect.DeepEqual(expected.Metadata, actual.Metadata) {
```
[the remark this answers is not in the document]
ihcsim: It failed the last time I tried. I suspect that's due to some dynamically-generated values in the templates.

alpeb:
@Pothulapati Also, I wasn't able to test, but it appears this new approach isn't complatible with multi-stage? One thing that occurs to me for that would be to enforce putting config-level templates under a separate `templates/config` dir...

ihcsim:
@alpeb We did some more testing yesterday, and decided that the only way to make sure the addons code works for the different install/upgrade commands and subcommands is to follow the pattern that is used for the [partials](https://github.com/linkerd/linkerd2/blob/d9956f3b35f1a083642cc1420439266476ed19d7/pkg/charts/charts.go#L75-L86), where we provide the full path inside the VFS. Let us know what you think. Thanks!

alpeb:
> @alpeb We did some more testing yesterday, and decided that the only way to make sure the addons code works for the different install/upgrade commands and subcommands is to follow the pattern that is used for the [partials](https://github.com/linkerd/linkerd2/blob/d9956f3b35f1a083642cc1420439266476ed19d7/pkg/charts/charts.go#L75-L86), where we provide the full path inside the VFS. Let us know what you think.

I liked how the previous changes allowed for so little customization to be required when adding a new add-on. So I was hoping the VFS loading logic, having being informed of the current stage, would be able to detect the stage some template belonged to given its subdir location or some other convention. Nice to have, but I would understand if that'd be too much at this point :slightly_smiling_face:

alpeb:
Oh I see all the VFS magic from `charts.go` is gone now. Alright, that looks good to me; we'll see later if being able to add addons with little-to-none changes is really required :+1:

ihcsim:
> I liked how the previous changes allowed for so little customization to be required when adding a new add-on

@alpeb Yeah, I back out of it because previously, Tarun and I thought of introducing the usage of the Helm lib to load charts in this PR. Then in a separate PR, see if we can refactor the existing charts code to use the same mechanism. But that would make it impossible to run the CLI unit tests (afaict) without requiring everyone to run `bin/helm-build`. (I didn't want to add a `command.Exec()` to the Go unit test code.) The reason is that when using the Helm lib, the `render()` function of `install` will always try to look for the addons subcharts in `charts/linkerd2/charts`, not `charts/add-ons`. There might be ways to try to bend it, but I feel it will further complicate the code.

We feel what we did here is easier to reason about, because it offers the consistency of using the same pattern as that used by the existing code to render the Linkerd, partials and CNI charts. Thanks for your help on this one!

Pothulapati:
@alpeb Thanks for the review! Updated the PR now.

As @ihcsim mentioned, These changes feel more concrete and stable as we are re-using the already present stuff. Also, Not using reflect, and writing concrete type checks make more sense!

Pothulapati:
One more thing, I think we should address is updating `helm-build` to also do `dep up` and get relevant sub-charts.

PR #3956, opened 2020-01-20
Rename noInitContainer

### LINKED ISSUE #3930: Rename NoInitContainer to CniEnabled
As NoInitContainer pretty much means that cni is enabled it makes sense to make it more clear and just rename it.

zaharidichev:
This PR renames NoInitContainer to CniEnabled to make the purpose of the parameter a bit more explicit.

Fixes #3930


grampelberg:
@zaharidichev we should call this out in the upgrade notes right?

zaharidichev:
@adleong Yes, that is correct, it only affects the helm users really. I am not too familiar with the Helm upgrade procedure but as @grampelberg noted this should be called out in the pgrade notes.

charts/linkerd2/templates/_config.tpl
```diff
@@ -1,7 +1,7 @@
 {{- define "linkerd.configs.global" -}}
 {
   "linkerdNamespace": "{{.Values.global.namespace}}",
-  "cniEnabled": {{ default false .Values.noInitContainer }},
+  "cniEnabled": {{ default false .Values.cniEnabled }},
```
alpeb: It seems that we missed prepending `global` to this one in the previous PR.

adleong:
@zaharidichev: I think you mentioned that you were going to speak with @ihcsim about how this impacts the upgrade process.  Once you do that, would you mind writing up your findings here so that I can follow along?

zaharidichev:
@adleong I did some testing and at the moment the upgrade procedure that is described in the docs does now work for upgrading from 2.6 to edge. So I submitted this issue and will investigate further: https://github.com/linkerd/website/issues/644 I however do not think that this is a blocker for getting this PR merged

PR #3953, opened 2020-01-20
Fix CNI config parsing

zaharidichev:
This PR addreses the problem introduced after #3766. It also uses an unreleased dependency of linkerd-proxy-init, so it is blocked on a new release of https://github.com/linkerd/linkerd2-proxy-init. So do not merge until then, please :) 

Fixes #3941

PR #3957, opened 2020-01-21
Fix helm list of ports not rendering correctly

zaharidichev:
There was a problem that caused helm install to not reflect the proper list of ignored inbound and outbound ports. Namely if you supply just one port, that would not get reflected.

To reproduce do a: 

```
 helm install \
       --name=linkerd2 \
       --set identity.issuer.crtExpiry=2021-01-14T14:21:43Z \
       --set-string global.proxyInit.ignoreInboundPorts="6666" \
       linkerd-edge/linkerd2
```

Check your config: 

```bash
 $ kubectl get configmap -n linkerd -oyaml | grep ignoreInboundPort
 "ignoreInboundPorts":[],
```

### DIFF
--- a/charts/linkerd2/templates/_config.tpl
+++ b/charts/linkerd2/templates/_config.tpl
@@ -30,22 +30,10 @@
     "port": {{.Values.global.proxy.ports.control}}
   },
   "ignoreInboundPorts":[
-    {{- $ports := splitList "," .Values.global.proxyInit.ignoreInboundPorts -}}
-    {{- if (and (gt (len $ports) 0) (ne (first $ports) "")) }}
-    {{- $last := sub (len $ports) 1 -}}
-    {{- range $i,$port := $ports -}}
-    {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
-    {{- end -}}
-    {{- end -}}
+    {{- include "partials.splitStringListToPorts" .Values.global.proxyInit.ignoreInboundPorts -}}
   ],
   "ignoreOutboundPorts":[
-    {{- $ports := splitList "," .Values.global.proxyInit.ignoreOutboundPorts -}}
-    {{- if (and (gt (len $ports) 0) (ne (first $ports) "")) }}
-    {{- $last := sub (len $ports) 1 -}}
-    {{- range $i,$port := $ports -}}
-    {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
-    {{- end -}}
-    {{- end -}}
+    {{- include "partials.splitStringListToPorts" .Values.global.proxyInit.ignoreOutboundPorts -}}
   ],
   "inboundPort":{
     "port": {{.Values.global.proxy.ports.inbound}}
--- a/charts/linkerd2/templates/_config.tpl
+++ b/charts/linkerd2/templates/_config.tpl
@@ -31,7 +31,7 @@
   },
   "ignoreInboundPorts":[
     {{- $ports := splitList "," .Values.global.proxyInit.ignoreInboundPorts -}}
-    {{- if gt (len $ports) 1}}
+    {{- if (and (gt (len $ports) 0) (ne (first $ports) "")) }}
     {{- $last := sub (len $ports) 1 -}}
     {{- range $i,$port := $ports -}}
     {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
@@ -40,7 +40,7 @@
   ],
   "ignoreOutboundPorts":[
     {{- $ports := splitList "," .Values.global.proxyInit.ignoreOutboundPorts -}}
-    {{- if gt (len $ports) 1}}
+    {{- if (and (gt (len $ports) 0) (ne (first $ports) "")) }}
     {{- $last := sub (len $ports) 1 -}}
     {{- range $i,$port := $ports -}}
     {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
--- a/charts/partials/templates/_helpers.tpl
+++ b/charts/partials/templates/_helpers.tpl
@@ -44,3 +44,18 @@ For example "11,22,55,44" will become "11","22","55","44"
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Splits a coma separated list into a list of port objects.
+For example "11,22,55" will become{"port":11},{"port":22},
+{"port":55}
+*/}}
+{{- define "partials.splitStringListToPorts" -}}
+{{- if gt (len .) 0 -}}
+{{- $ports := splitList "," . -}}
+{{- $last := sub (len $ports) 1 -}}
+{{- range $i,$port := $ports -}}
+{"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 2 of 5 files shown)


charts/linkerd2/templates/_config.tpl
```diff
@@ -31,7 +31,7 @@
   },
   "ignoreInboundPorts":[
     {{- $ports := splitList "," .Values.global.proxyInit.ignoreInboundPorts -}}
-    {{- if gt (len $ports) 1}}
+    {{- if (and (gt (len $ports) 0) (ne (first $ports) "")) }}
```
adleong: What are the different cases that we are expecting and how do they arise?  Are these inputs valid?

* $ports is length 0
* $ports is length 1 and contains just an empty string
* $ports is length 2 and the second element is an empty string

ihcsim:
I might be missing something (I haven't been following the recent changes to the 'ignore ports' options), but in this case, we just need to either remove the `gt (len $ports)` condition or change it to `ge`, right?

E.g., if I remove the conditional entirely, I get:
```
$ helm template charts/linkerd2 --set-file global.identityTrustAnchorsPEM=isim-tests/tls/ca.crt --set-file identity.issuer.tls.crtPEM=isim-tests/tls/issuer.crt  \
--set-file identity.issuer.tls.keyPEM=isim-tests/tls/issuer.key \
--set identity.issuer.crtExpiry="2021-01-14T14:21:43Z"\
 --set-string global.proxyInit.ignoreInboundPorts="6666"\
 --set-string global.proxyInit.ignoreOutboundPorts="7777" \
| grep -i 'ignoreinboundports\|ignoreoutboundports'
      "ignoreInboundPorts":[{"port":6666}],
      "ignoreOutboundPorts":[{"port":7777}],

$ helm template charts/linkerd2 --set-file global.identityTrustAnchorsPEM=isim-tests/tls/ca.crt \
--set-file identity.issuer.tls.crtPEM=isim-tests/tls/issuer.crt \
--set-file identity.issuer.tls.keyPEM=isim-tests/tls/issuer.key \
--set identity.issuer.crtExpiry="2021-01-14T14:21:43Z" \
--set-string global.proxyInit.ignoreInboundPorts="6666\,6667" \
--set-string global.proxyInit.ignoreOutboundPorts="7777\,7778" \
 | grep -i 'ignoreinboundports\|ignoreoutboundports'
      "ignoreInboundPorts":[{"port":6666},{"port":6667}],
      "ignoreOutboundPorts":[{"port":7777},{"port":7778}],
```
Does that look right?

Here's my diff:
```diff
diff --git a/charts/linkerd2/templates/_config.tpl b/charts/linkerd2/templates/_config.tpl
index f4c74fe5..11873f48 100644
--- a/charts/linkerd2/templates/_config.tpl
+++ b/charts/linkerd2/templates/_config.tpl
@@ -31,21 +31,17 @@
   },
   "ignoreInboundPorts":[
     {{- $ports := splitList "," .Values.global.proxyInit.ignoreInboundPorts -}}
-    {{- if gt (len $ports) 1}}
     {{- $last := sub (len $ports) 1 -}}
     {{- range $i,$port := $ports -}}
     {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
     {{- end -}}
-    {{- end -}}
   ],
   "ignoreOutboundPorts":[
     {{- $ports := splitList "," .Values.global.proxyInit.ignoreOutboundPorts -}}
-    {{- if gt (len $ports) 1}}
     {{- $last := sub (len $ports) 1 -}}
     {{- range $i,$port := $ports -}}
     {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
     {{- end -}}
-    {{- end -}}
   ],
   "inboundPort":{
     "port": {{.Values.global.proxy.ports.inbound}}
```

zaharidichev:
@ihcsim So I attempted that initially and the result produced for the `ignoreInboundPorts` and `ignoreOutboundPorts` when both of these are empty is `ignoreInboundPorts : [""]` and `ignoreOutboundPorts: [""]` respectively. Apart from tests failing because of that, I think that an empty string array should be different than an array of one empty string? Am I missing something?

charts/linkerd2/templates/_config.tpl
zaharidichev: So if by inbputs you mean what is passed to `global.proxyInit.ignoreInboundPorts`, then: 

- if its empty, its an empty string, so after split $ports will be a list of one empty string, which we filter
- if its just "", then its the same case as above
- if the length is two, then we do not do any validation to guard against an empty string not being a port.

Essentially what is happenign here is the following: 
- We fix the problem where specifying just one port will result in it not getting written to the config map
- We make sure that we have an empty `global.proxyInit.ignoreInboundPorts` we do not end up with an array that looks like `[""]` in the config map. 

I think any further validation on input is pushing the helm functionality a bit too far. WDYT? 
As you suggested in https://github.com/linkerd/linkerd2/pull/3953#discussion_r369298051, I am fine with factoring all that out into a function of its own so we can reuse it.

cli/cmd/main_test.go
```diff
@@ -55,6 +55,7 @@ func diffTestdata(t *testing.T, path, actual string) {
 	if actual == expected {
 		return
 	}
+	println(actual)
```
alpeb: I think you missed this

zaharidichev: this always gets me...

ihcsim:
@zaharidichev Ah.. you are right. Out of curiosity, do you know why we converted the slices into string in [`install.go`](https://github.com/linkerd/linkerd2/blob/60ac0d552728e9e780102b692cd22c8b9c3741f6/cli/cmd/install.go#L704-L705), instead of just passing the slice as-is? Then the Helm workflow would just be `helm install --set ignoreInboundPorts={"port1", "port2"}`.

Thanks for looking into this.

charts/linkerd2/templates/_config.tpl
ihcsim: @zaharidichev Thanks for the explanation. There is a good chance that, in the future, I will forget `{{- if (and (gt (len $ports) 0) (ne (first $ports) "")) }}` can handle all the cases @adleong pointed out.

Will it work if we check whether `Values.global.proxyInit.ignoreInboundPorts` is empty? If it is, just skip the `splitList` and `range` loop:

```
  "ignoreInboundPorts":[
    {{- if gt (len $.Values.gobal.proxyInit.ignoreInboundPorts) 0 -}}
    {{- $ports := splitList "," .Values.global.proxyInit.ignoreInboundPorts -}}
    {{- $last := sub (len $ports) 1 -}}
    {{- range $i,$port := $ports -}}
    {"port":{{$port}}}{{ternary "," "" (ne $i $last)}}
    {{- end -}}
    {{- end -}}
  ],
```

zaharidichev:
@ihcsim I think `IgnoreInboundPorts` and `IgnoreOutboundPorts` have been strings for a long time. It was just a matter of how we are constructing them. Before it was from a slice of uints (passed via the cli) I believe, but it was changed to be a slice of string to support port ranges.

charts/linkerd2/templates/_config.tpl
zaharidichev: Alright, so I rebased against #3953 (so this needs to go in first) and used your suggested approach for both fixes. Now these splitting functions are refactored in _helpers.tpl. Does that look better?

PR #3959, opened 2020-01-21
[PoC] Get cluster credentials

adleong:
This is a proof of concept to demonstrate how it's possible to get credentials for a service account from a Kubernetes cluster in the form of a Kubeconfig secret.  This secret could then be created in a different Kubernetes cluster to allow cross-cluster API requests.

Steps:

1. Create a service account called `linkerd-mirror` with RBAC to list services:

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: linkerd
---
kind: ClusterRole
metadata:
  name: linkerd-mirror
rules:
- apiGroups:
  - ""
  resources:
  - services
  verbs:
  - list
---
kind: ServiceAccount
apiVersion: v1
metadata:
  name: linkerd-mirror
  namespace: linkerd
  labels:
    linkerd.io/control-plane-component: mirror
---
kind: ClusterRoleBinding
kind: ClusterRoleBinding
metadata:
  labels:
    linkerd.io/control-plane-component: mirror
  name: linkerd-mirror
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: linkerd-mirror
subjects:
- kind: ServiceAccount
  name: linkerd-mirror
  namespace: linkerd
```

2. Run `linkerd cluster get-credentials`.  This outputs a secret containing the Kubeconfig.

3. Verify the Kubeconfig has the desired RBAC:

```
linkerd cluster get-credentials | grep kubeconfig | cut -c '15-' | base64 -D - > mirror.kubeconfig
env KUBECONFIG=mirror.kubeconfig kubectl get po # should fail
env KUBECONFIG=mirror.kubeconfig kubectl get svc # should succeed
```

4. Install the secret into another cluster

```
linkerd cluster get-credentials | kubectl --context=my-other-cluster apply -f -
```

5. Mount the secret and use the kubeconfig to make cross-cluster API calls

(left as an exercise to the reader)

### DIFF
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -1,0 +1,124 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+type clusterOptions struct {
+	namespace      string
+	serviceAccount string
+	clusterName    string
+}
+
+func newCmdCluster() *cobra.Command {
+
+	options := clusterOptions{}
+
+	clusterCmd := &cobra.Command{
+		Use:   "cluster",
+		Short: "Set up cross-cluster access",
+		Args:  cobra.NoArgs,
+	}
+
+	getCredentialsCmd := &cobra.Command{
+		Use:   "get-credentials",
+		Short: "Get cluster credentials as a secret",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			rules := clientcmd.NewDefaultClientConfigLoadingRules()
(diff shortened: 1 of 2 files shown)


cli/cmd/cluster.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			creds := corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("cluster-credentials-%s", options.clusterName),
+					Namespace: controlPlaneNamespace,
+				},
+				Data: map[string][]byte{
+					"kubeconfig": kubeconfig,
+				},
```
zaharidichev: should we have a special type of secret or some other metadata to enabled the filtering/selection of a secret that contains remote api credentials? Or wil we simply rely on the name format?

zaharidichev:
I think the verbs for the service account should also have "watch", right?

zaharidichev:
I think this does not output the secret correctly. For example if I do

```
linkerd cluster get-credentials | kubectl  -f -
error: error validating "STDIN": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false
```

We aim to output valid yaml that can be piped to apply, right?

grampelberg:
This looks pretty cool! Way easier than I was expecting.

zaharidichev:
@adleong I am closing that in favor of future work :)

PR #3960, opened 2020-01-22
[PoC] Create cluster credentials

zaharidichev:
A bit of POC expansion on the work @adleong did. Now we can do the following: 

```bash
# Create cluster and namespace
kind create cluster --name cluster-a
kubectl create namespace linkerd

# Create credentials on cluster
$ bin/linkerd cluster create-credentials | kubectl apply -f -
clusterrole.rbac.authorization.k8s.io/linkerd-mirror created
serviceaccount/linkerd-mirror created
clusterrolebinding.rbac.authorization.k8s.io/linkerd-mirror created

# Test credentials work
linkerd cluster get-credentials | grep kubeconfig | cut -c '15-' | base64 -D - > mirror.kubeconfig
export  KUBECONFIG=mirror.kubeconfig

# Accessing pods does not work
$ kubectl get pods --all-namespaces
Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:linkerd:linkerd-mirror" cannot list resource "pods" in API group "" at the cluster scope

# Accessing services works
$ kubectl get svc --all-namespaces
NAMESPACE     NAME         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)                  AGE
default       kubernetes   ClusterIP   10.96.0.1    <none>        443/TCP                  78s
kube-system   kube-dns     ClusterIP   10.96.0.10   <none>        53/UDP,53/TCP,9153/TCP   76s
```

### DIFF
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -4,9 +4,13 @@ import (
 	"fmt"
 	"strings"
 
+	v1 "k8s.io/api/rbac/v1"
+
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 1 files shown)


grampelberg:
@zaharidichev @adleong WDYT about sticking this under at `cluster` command group and having a `validate` command here as well that does the auth validation?

zaharidichev:
@grampelberg Would the auth just be given a svc-account name, pointed to the local cluster and attempt to read the secret contents and construct an api to successfully hit the remote cluster with? Something like that I guess?

grampelberg:
I would have the command:

- Check for the existence of the secrets in the local cluster.
- Try and use that kubeconfig to connect to the remote cluster.
- Validate that the RBAC is correct.

I'm wondering if there's a fourth step around fetching services and/or watching for changes. Theoretically, if the auth review is successful everything would work ... but who knows?

zaharidichev:
@grampelberg I think if the RBAC is as expected then anything else we need to worry about can be left to the service mirror component logs to report. Although if we want to be even more sure, we can check how `can-i` is implemented on the kubectl CLI and try to do something similar via the API

grampelberg:
That's what I was thinking, `can-i` for watch/get/list on services. Should we output what namespaces we can and can't mirror? I could see someone having a super sensitive namespace that they don't want any mirroring on and we'd be unable to view that.

zaharidichev:
Closing in favor of future work

PR #3966, opened 2020-01-23
Changes for edge-20.1.3

kleimkuhler:
## edge-20.1.3

* CLI
  * Introduced `linkerd check --pre --linkerd-cni-enabled`, used when the CNI
    plugin is used, to check it has been properly installed before proceeding
    with the control plane installation
  * Added support for the `--as-group` flag so that users can impersonate
    groups for Kubernetes operations (thanks @mayankshah160!)
* Controller
  * Fixed an issue where an override of the Docker registry was not being
    applied to debug containers (thanks @javaducky!)
  * Added check for the Subject Alternate Name attributes to the API server
    when access restrictions have been enabled (thanks @javaducky!)
  * Added support for arbitrary pod labels so that users can leverage the
    Linkerd provided Prometheus instance to scrape for their own labels
    (thanks @daxmc99!)
  * Fixed an issue with CNI config parsing


CHANGES.md
```diff
@@ -1,3 +1,20 @@
+## edge-20.1.3
+
+* CLI
+  * Introduced `linkerd check --cni` to ensure the CNI plugins can properly be
+    installed on the cluster
+  * Added support for the `--as-group` flag so that users can impersonate
```
ihcsim: This is an external contribution from Mayank (mayankshah160 ).

CHANGES.md
```diff
@@ -1,3 +1,20 @@
+## edge-20.1.3
+
+* CLI
+  * Introduced `linkerd check --cni` to ensure the CNI plugins can properly be
```
ihcsim: This check is used to determine the status of the CNI after its installation. IOW, not whether the cluster meets all the preconditions of the CNI installation, but more about if the CNI installation completed successfully.

CHANGES.md
```diff
@@ -1,3 +1,20 @@
...
+    installed on the cluster
```
alpeb: This ended up being `linkerd check --pre --linkerd-cni-enabled`. So this should read instead something like
```
Introduced `linkerd-check --pre --linkerd-cni-enabled`, used when the CNI plugin is used,  to check it has been properly installed before proceeding with the control plane installation
```

CHANGES.md
kleimkuhler: Ah got it thanks!

CHANGES.md
ihcsim: Good catch! FWIW, it also works with `linkerd check --linkerd-cni-enabled`.

kleimkuhler: Okay thanks for clarifying!

alpeb: @ihcsim apparently not:
```bash
$ bin/go-run cli check --linkerd-cni-enabled
Error: Validation error when executing check command: --linkerd-cni-enabled can only be used with --pre
```

adleong: right.  the post-install check will read the CNI mode from the configmap.  but for `--pre` the Linkerd configmap doesn't exist yet so it needs the `--linkerd-cni-enabled` flag

Issue #2316, opened 2019-02-18
Remove integration test dependency on httpbin.org


siggy:
The egress integration test relies on httpbin.org to test requests:
https://github.com/linkerd/linkerd2/blob/ed3fbd75f3cbee76b45d73687134e23f45b41374/test/egress/egress_test.go#L92

This means that behavior changes in that external online service can break ci:
https://travis-ci.org/linkerd/linkerd2/jobs/495002001#L7405

Propose removing the dependency on httpbin.org. One approach could be to boot our own copy of httpbin and route to it via a service with an external name.

siggy:
Related: https://github.com/postmanlabs/httpbin/issues/536

adleong:
This is causing us CI grief as the `httpbin.org` behavior is now non-deterministic regarding http vs https.

PR #3968, opened 2020-01-24
Temporarily fix flaky integration test

kleimkuhler:
*From the comment disabling the test*:

#2316

The response from `http://httpbin.org/get` is non-deterministic--returning
either `http://..` or `https://..` for GET requests. As #2316 mentions, this
test should not have an external dependency on this endpoint. As a workaround
for edge-20.1.3, temporarily disable this test and renable with one that has
reliable behavior.


kleimkuhler:
I merged master in so that #3969 would stop failing tests here. I also update the test instead of removing it to expect *either* `http` or `https` so that the test is not fully disabled.

PR #3971, opened 2020-01-24
Update edge-20.1.3 release changes

kleimkuhler:
# edge-20.1.3

An update to the Helm charts has caused a **breaking change** for users who
have installed Linkerd using Helm. In order to make the purpose of the
`NoInitContainer` parameter more explicit, it has been renamed to `CniEnabled`.

* CLI
  * Introduced `linkerd check --pre --linkerd-cni-enabled`, used when the CNI
    plugin is used, to check it has been properly installed before proceeding
    with the control plane installation
  * Added support for the `--as-group` flag so that users can impersonate
    groups for Kubernetes operations (thanks @mayankshah160!)
* Controller
  * Fixed an issue where an override of the Docker registry was not being
    applied to debug containers (thanks @javaducky!)
  * Added check for the Subject Alternate Name attributes to the API server
    when access restrictions have been enabled (thanks @javaducky!)
  * Added support for arbitrary pod labels so that users can leverage the
    Linkerd provided Prometheus instance to scrape for their own labels
    (thanks @daxmc99!)
  * Fixed an issue with CNI config parsing
* Helm
  * **Breaking change**: Renamed `NoInitContainer` parameter to `CniEnabled`
  * Fixed an issue with `helm install` where the lists of ignored inbound and
    outbound ports would not be reflected


CHANGES.md
```diff
@@ -15,6 +19,10 @@
     Linkerd provided Prometheus instance to scrape for their own labels
     (thanks @daxmc99!)
   * Fixed an issue with CNI config parsing
+* Helm
+  * **Breaking change**: Renamed `NoInitContainer` parameter to `CniEnabled`
```
zaharidichev: these start with lowercase now

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 ## edge-20.1.3
 
+An update to the Helm charts has caused a **breaking change** for users who
+have installed Linkerd using Helm. In order to make the purpose of the
+`NoInitContainer` parameter more explicit, it has been renamed to `CniEnabled`.
```
zaharidichev: lower case

CHANGES.md
```diff
@@ -1,5 +1,9 @@
...
+
 * CLI
   * Introduced `linkerd check --pre --linkerd-cni-enabled`, used when the CNI
```
zaharidichev: Might be that I am not native speaker but this sounds a bit strange to me. How about something like "Introduced CNI checks to confirm the CNI plugin is installed and ready".  On another note this check is ran not only in `--pre` stage but also in `linkerd check` **if the cni plugin is present**

kleimkuhler: @zaharidichev?

```
* Introduced CNI checks to confirm the CNI plugin is installed and ready;
  this is done through `linkerd check --pre --linkerd-cni-enabled` before installation, and
  `linkerd check` if the CNI plugin is present
```

zaharidichev: That sounds correct.

Issue #3926, opened 2020-01-15
Enable the jsx-a11y/click-events-have-key-events eslint rule


klingerf:
In #3882, we adopted more eslint rules for our Javascript codebase. The `jsx-a11y/click-events-have-key-events` rule is recommended, but it wasn't enabled as part of that PR, since it requires code changes. We should enable the rule and fix the code that fails to lint. We can enable it by removing the override that appears here:

https://github.com/linkerd/linkerd2/blob/0eabbe4173fd8a7da91d5771572bd08d85268275/web/app/.eslintrc#L85-L88

christyjacob4:
@klingerf please assign it to me

cpretzer:
@christyjacob4 looking forward to reviewing your PR for this 😄

adrijshikhar:
@christyjacob4 are you still working on this?

christyjacob4:
yea obviously @adrijshikhar

PR #3974, opened 2020-01-25
Refactoring to suppress eslint warnings react/no-did-update-set-state flag

### LINKED ISSUE #3928: Enable the react/no-did-update-set-state eslint rule
In #3882, we adopted more eslint rules for our Javascript codebase. The `react/no-did-update-set-state` rule is recommended, but it wasn't enabled as part of that PR, since it requires code changes. We should enable the rule and fix the code that fails to lint. We can enable it by removing the override that appears here:

https://github.com/linkerd/linkerd2/blob/0eabbe4173fd8a7da91d5771572bd08d85268275/web/app/.eslintrc#L85-L88

christyjacob4:
Upon enabling react/no-did-update-set-state flag in .eslintrc , a couple of warnings are raised because it is a
bad practice to use the setState() function within the componentDidUpdate() hook.

The code has been refactored to follow the eslint spec.

Fixes #3928

Signed-off-by: Christy Jacob <christyjacob4@gmail.com>

### DIFF
--- a/web/app/.eslintrc
+++ b/web/app/.eslintrc
@@ -84,7 +84,6 @@
 
     // These rules represent best practices that we're not adhering to, and should be enabled
     "jsx-a11y/click-events-have-key-events": 0,
-    "no-param-reassign": 0,
-    "react/no-did-update-set-state": 1
+    "no-param-reassign": 0
   }
 }
--- a/web/app/.eslintrc
+++ b/web/app/.eslintrc
@@ -85,6 +85,6 @@
     // These rules represent best practices that we're not adhering to, and should be enabled
     "jsx-a11y/click-events-have-key-events": 0,
     "no-param-reassign": 0,
-    "react/no-did-update-set-state": 0
+    "react/no-did-update-set-state": 1
   }
 }
--- a/web/app/js/components/Namespace.jsx
+++ b/web/app/js/components/Namespace.jsx
@@ -48,7 +48,9 @@ class Namespaces extends React.Component {
     if (!_isEqual(prevProps.match.params.namespace, params.namespace)) {
       // React won't unmount this component when switching resource pages so we need to clear state
       this.api.cancelCurrentRequests();
-      this.setState(this.getInitialState(params));
+      this.onUpdate(function callback() {
+        this.setState(this.getInitialState(params));
+      });
     }
 
     handlePageVisibility({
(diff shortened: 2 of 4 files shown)


christyjacob4:
@cpretzer Thanks for the review. However, I found that my code changes breaks some functionality. You can check it out in this video

## Without refactoring
before

## After refactoring
after

christyjacob4:
@cpretzer I found a possible fix for it 
It works by simply wrapping the setState function within another function.
```jsx
  componentDidUpdate(prevProps) {
    const { match, isPageVisible } = this.props;
    const { params } = match;
    if (!_isEqual(prevProps.match.params.namespace, params.namespace)) {
      // React won't unmount this component when switching resource pages so we need to clear state
      this.api.cancelCurrentRequests();
      this.resetState(params);
    }

    handlePageVisibility({
      prevVisibilityState: prevProps.isPageVisible,
      currentVisibilityState: isPageVisible,
      onVisible: () => this.startServerPolling(),
      onHidden: () => this.stopServerPolling(),
    });
  }

  resetState(params) {
    this.setState(this.getInitialState(params));
  }

```
This guarantees that we dont go into an infinite loop since we use the if condition to check the state. This is also the recommended method in react websites.
https://reactjs.org/docs/react-component.html#componentdidupdate

cpretzer:
thanks @christyjacob4 I'm going to pull and test today

christyjacob4:
@cpretzer @grampelberg Please check out my latest commit with the changes I had proposed

christyjacob4:
@cpretzer Thanks for the review. So can this change be merged? Also Can I start working on https://github.com/linkerd/linkerd2/issues/3926

cpretzer:
@christyjacob4 we just need one more review and can get it merged. working on that now

web/app/.eslintrc
```diff
@@ -85,6 +85,6 @@
     // These rules represent best practices that we're not adhering to, and should be enabled
     "jsx-a11y/click-events-have-key-events": 0,
     "no-param-reassign": 0,
-    "react/no-did-update-set-state": 0
+    "react/no-did-update-set-state": 1
```
klingerf: You can just delete this line, since the rule is enabled by default.

christyjacob4: Thanks for the feedback @klingerf . Ill make the change and push

Issue #3973, opened 2020-01-24
Linkerd check command should return warning when version endpoint is not reachable


sukhong:
### What is the issue?
Linkerd check command currently errors out when version endpoint is not reachable. However, we have a need to run linkerd in a private cluster with no external network connectivity.

Linkerd check command currently fails when it cannot reach version check endpoint during linkerd-version step.

This should return warning instead, to support private cluster scenario.

### How can it be reproduced?

Run linkerd check command without external network.

#### `linkerd check` output

```text
{
      "categoryName": "linkerd-version",
      "checks": [
        {
          "description": "can determine the latest version",
          "hint": "https://linkerd.io/checks/#l5d-version-latest",
          "error": "Get https://versioncheck.linkerd.io/version.json?version=stable-2.6.1\u0026uuid=453bd534-21e4-4096-ad53-a7af5b7f395a\u0026source=cli: dial tcp: lookup versioncheck.linkerd.io on 127.0.0.53:53: read udp 127.0.0.1:52288-\u003e127.0.0.53:53: i/o timeout",
          "result": "error"
        },
        {
          "description": "cli is up-to-date",
          "hint": "https://linkerd.io/checks/#l5d-version-cli",
          "error": "unsupported version channel: stable-2.6.1",
          "result": "warning"
        }
      ]
    },
    {
      "categoryName": "control-plane-version",
      "checks": [
        {
          "description": "control plane is up-to-date",
          "hint": "https://linkerd.io/checks/#l5d-version-control",
          "error": "unsupported version channel: stable-2.6.1",
          "result": "warning"
        },
        {
          "description": "control plane and cli versions match",
          "result": "success"
        }
      ]
    }
```

### Environment

- Kubernetes Version: 1.5.7
- Cluster Environment: (GKE, AKS, kops, ...) aks-engine
- Host OS: Ubuntu
- Linkerd version: 2.6.1

adleong:
@sukhong This sounds reasonable to me.  Are you interested in submitting a PR to fix this?

mayankshah1607:
@adleong @sukhong I could fix this one if that's okay

sukhong:
@mayankshah1607 I would really appreciate it if you can fix this for us :)

mayankshah1607:
Thanks @sukhong ! I've opened a PR with a fix - https://github.com/linkerd/linkerd2/pull/3975
Could you pull the changes and let me know if that works for you? :)

PR #3975, opened 2020-01-27
Update "linkerd-version" check

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/3973

- Update check command to throw a warning when it fails to determine
latest version

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1010,6 +1010,7 @@ func (hc *HealthChecker) allCategories() []category {
 				{
 					description: "can determine the latest version",
 					hintAnchor:  "l5d-version-latest",
+					warning:     true,
 					check: func(ctx context.Context) (err error) {
 						if hc.VersionOverride != "" {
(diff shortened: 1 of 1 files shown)


mayankshah1607:
@zaharidichev yes, that makes a lot more sense. I was thinking we could add a method in `version` that pings the `checkURL` for some response; if a network error is returned, we could directly throw a `SkipError` before proceeding with the check. Does that sound good?

zaharidichev:
@mayankshah1607 Can't you just use the potential errore returned by `version.GetLatestVersions` ? Seems like it will include information about the status code: https://github.com/linkerd/linkerd2/blob/master/pkg/version/channels.go#L67. Seems more reasonable than introduing a separate ping method. WDYT?

adleong:
I'm not sure there's a meaningful difference between not being able to reach the version check endpoint vs some other error coming from the version check endpoint.  Both cases seem like they should be a warning to me.

zaharidichev:
well okey then, lets simply make it a warning

PR #3976, opened 2020-01-27
Add helm upgrade integration test

zaharidichev:
In light of the breaking changes we are introducing to the Helm chart and the convoluted upgrade process (see https://github.com/linkerd/website/pull/647) an integration test can be quite helpful. This simply installs latest stable through helm install and then upgrades to the current head of the branch. 

### DIFF
--- a/bin/test-run
+++ b/bin/test-run
@@ -3,8 +3,9 @@
 # This script executes the full suite of integration tests, in series:
 # 1. upgrade_integration_tests
 # 2. helm_integration_tests
-# 3. deep_integration_tests
-# 4. external_issuer_integration_tests
+# 3. helm_upgrade_integration_tests
+# 4. deep_integration_tests
+# 5. external_issuer_integration_tests
 
 bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 
--- a/bin/test-run
+++ b/bin/test-run
@@ -17,6 +17,7 @@ printf 'Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\n' "$linker
 
 upgrade_integration_tests
 helm_integration_tests
+helm_upgrade_integration_tests
 deep_integration_tests
 external_issuer_integration_tests
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -8,8 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/linkerd/linkerd2/pkg/version"
-
 	"github.com/spf13/pflag"
 
 	"github.com/fatih/color"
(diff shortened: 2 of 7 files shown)


bin/test-run
```diff
@@ -17,6 +17,7 @@ printf 'Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\n' "$linker
 
 upgrade_integration_tests
 helm_integration_tests
+helm_upgrade_integration_tests
```
adleong: I think the comment on line 6 needs to be updated

cli/cmd/root.go
```diff
@@ -100,6 +108,7 @@ func init() {
 	RootCmd.PersistentFlags().StringArrayVar(&impersonateGroup, "as-group", []string{}, "Group to impersonate for Kubernetes operations")
 	RootCmd.PersistentFlags().StringVar(&apiAddr, "api-addr", "", "Override kubeconfig and communicate directly with the control plane at host:port (mostly for testing)")
 	RootCmd.PersistentFlags().BoolVar(&verbose, "verbose", false, "Turn on debug logging")
+	RootCmd.PersistentFlags().StringVar(&cliVersionOverride, "cli-version-override", "", "Used to override the version of the cli (mostly for testing)")
```
alpeb: WDYT about exposing this option only to `linkerd check`, alongside `--expected-version` instead of making it a global flag?

test/install_test.go
```diff
@@ -350,7 +356,32 @@ func TestInstallHelm(t *testing.T) {
...
+	var err error
+	helmTLSCerts, err = tls.GenerateRootCAWithDefaults(cn)
+	if err != nil {
+		t.Fatalf("failed to generate root certificate for identity: %s", err)
+	}
+
+	var chartToInstall string
+	args := []string{"--name", TestHelper.GetHelmReleaseName()}
+
+	if TestHelper.UpgradeHelmFromVersion() != "" {
```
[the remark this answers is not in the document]
zaharidichev: I will, but I think this needs to be kept as a separate function. Just to distingush that there might be difference in values between the latest stable and the edge

bin/test-run
```diff
@@ -16,11 +16,11 @@ init_test_run "$@"
...
 
-upgrade_integration_tests
-helm_integration_tests
+#upgrade_integration_tests
+#helm_integration_tests
 helm_upgrade_integration_tests
-deep_integration_tests
-external_issuer_integration_tests
+#deep_integration_tests
+#external_issuer_integration_tests
```
alpeb: Look like you forgot to uncomment :wink:

zaharidichev: thanks for catching that!

siggy:
@adleong We have `max-parallel: 3`, so adding another integration cluster should be ok.

PR #3979, opened 2020-01-27
Make Grafana a soft requirement

Pothulapati:
This PR aims to make grafana a soft requirement in Linkerd.

Changes Include:
- Making it optional in `web` component. If `grafana-addr` flag is empty, it is to be considered that grafana is not present, and the grafanaLinks in the MetricsTable should not be displayed.


Pothulapati:
Screenshot_2020-01-28 Linkerd

grampelberg:
@Pothulapati can we make the URL configurable as well? It'd be nice to be able to configure the grafana address and just link to that.

Pothulapati:
Yep that's the next part. 

And then the grafana add-on would just set it's endpoint to the variable :)

Pothulapati:
@alpeb Yep, My Plan is to separate the `healthChecks` for add-ons (hence the removal of `grafana` there). 
The healthChecks for add-ons would be based on `linkerd-values` cm. They would not fail or even show up in the checks if they are not present.

alpeb:
Sounds awesome! :100:

alpeb:
@Pothulapati what do you think about addressing the healthcheck changes we mentioned above (having them depend on the presence of each addon) in a separate PR, and then you can rebase this PR based off of that?

grampelberg:
@Pothulapati where'd this one get to?

Pothulapati:
@grampelberg Was waiting for #4247 and #4271 to get merged as they change the model quite a bit! Now that they are close, will start updating this!

Pothulapati:
This PR moves the grafana chart as an add-on which means that it is moved into a separate sub-chart which is added as a dependency to `linkerd2`, and config add-on is passed through `addon-config`through CLI, and `values.yaml` through helm. By using the add-on model, upgrades for these components work as these configuration is stored and is used.

**Changes Include:**
- Move grafana template files into a separate sub-chart folder, and add it as a dependency to `linkerd2` chart.
- Add `Grafana` struct which implements the add-on interface and make it a field in `Values` struct.
- Also adds `grafana` to the `linkerd-values` cm. (more on this later)
- Add `TestRenderAddOns` tests for different kinds of cases with add-ons.

**Regarding `linkerd-values` config-map:**

This config-map was only present when any of the add-on is enabled, but now that `linkerd-grafana` is enabled by default. This is made to be present in all cases.

`linkerd-values` configmap is where the configuration is stored for add-ons to have seamless upgrades. This config-map uses `_addon.tpl` file have a function which basically removes any overwritten configuration passed by the user when the add-on is disabled, and also removes `add-on.global` and `add-on.partials` fields which are added by helm automatically when add-on is present.

For Upgrades, For `l5dcharts.Values` struct first default configuration is read from `values.yaml`, which is later overwritten by the values in this configmap, and users can also further overwrite the configuration by using `--addon-config` flag.

pkg/healthcheck/healthcheck_addons.go
```diff
@@ -0,0 +1,76 @@
...
+	// so that further add-on specific checks can be ran
+	var values l5dcharts.Values
+	err = yaml.Unmarshal([]byte(cm), &values)
+	if err != nil {
+		return &SkipError{fmt.Sprintf("could not unmarshal %s config-map: %s", k8s.ValuesConfigMapName, err.Error())}
+	}
+
+	addOns, err := l5dcharts.ParseAddOnValues(&values)
+	if err != nil {
+		return &SkipError{fmt.Sprintf("could not read %s config-map: %s", k8s.ValuesConfigMapName, err.Error())}
```
alpeb, line 56: I believe `err` is enough, no need to call `err.Error()` here

alpeb: Also, why are these errors skipped?

pkg/healthcheck/healthcheck_addons.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	addOnMap := make(map[string]l5dcharts.AddOn)
+
+	for _, addOn := range addOns {
+		addOnMap[addOn.Name()] = addOn
+	}
+	hc.addOns = addOnMap
```
alpeb, line 64: You could deal directly here with `hc.addOns` and not have to go through the temp var `addonMap`

test/install_test.go
```diff
@@ -417,7 +417,7 @@ func TestCheckHelmStableBeforeUpgrade(t *testing.T) {
 		t.Fatalf("linkerd-smi-metrics SA labeling failed: %s", err)
 	}
 
-	testCheckCommand(t, "", TestHelper.UpgradeHelmFromVersion(), "", TestHelper.UpgradeHelmFromVersion())
+	testCheckCommand(t, "", TestHelper.UpgradeHelmFromVersion(), "", TestHelper.UpgradeHelmFromVersion(), false)
```
alpeb, line 420: Can you add a TODO comment explaining this is a temporary measure because 2.7.1 didn't have the `linkerd-addons` check section?

pkg/healthcheck/healthcheck_addons.go
```diff
@@ -0,0 +1,76 @@
+package healthcheck
```
alpeb, line 1: I like how this was added into a separate file, to avoid bloating more `healthcheck.go` :+1:

alpeb:
If I understood correctly, it's still pending hooking up specific healthchecks for this addon, right?

pkg/healthcheck/healthcheck_addons.go
Pothulapati: Now that `linkerd-values` is necessary, I think skipping those errors don't make sense. So, changing it.

test/install_test.go
Pothulapati: I can definitely add a TODO, but it makes for `TestCheckHelmStableBeforeUpgrade` to not check for the output, as its a before thing and checks could change in the new CLI.

Pothulapati:
@alpeb 

> Can you rename removeEmpty to something more explicit like linkerd.addons.santize-config or something like that?

done

> Also, can you condition the grafana scraping in the prometheus config on whether the addon is enabled?

added that! :)

> Can you check in what release was the linkerd-values CM introduced? Do you think it'd be feasible to give it a better name, say linkerd-config-addons?

`linkerd-values` cm was added in `edge-20.2.3` when tracing add-on was added into linkerd2. I think its safe to rename the cm. Existing users would get the default configuration back again for tracing add-on, They can use the `--addon-config` to pass the same config.

> Finally, why is that this addon's helm chart (and tracing's) depend on partials?

This is because, the components in the sub-charts should also get the proxy injected, so they use these templates, functions in partials to achieve the same just like other control-plane components.

Pothulapati:
**Using existing grafana instance:**
- Let `config.yaml` be
```yaml
global:
  grafanaUrl: "somegrafana.xyz"
grafana:
  enabled: false
```
- Apply using `linkerd install --addon-config config.yaml | k apply -f -`

Pothulapati:
This PR seems to work great but it has became rather huge than I imagined. So, I am breaking down this changes into the following Pr's:

- Refactor of `linkerd-config-addons` to make it more cleaner and reusable.
- Grafana Add-on sub-chart and relevant install and upgrade tests.
- Health checks for Grafana add-on.
- Add `grafanaUrl` to the configuration, to support bringing your own Grafana use-case.

Expect the above Prs to be opened in the following days :)

test/install_test.go
alpeb: Correct, but after 2.8 comes out, we should test that upgrading from 2.8 to whatever edge/stable comes next has a sane `linkerd check` output, thus not requiring skipping that anymore.
That being said, it's normal for this output to change from version to version, so a better long-term solution would be to have specific per-version outputs, but that'd be something for another PR :-)

alpeb:
> Existing users would get the default configuration back again for tracing add-on, They can use the `--addon-config` to pass the same config.

IIUC that'd be a manual step, so it grants a note in the release notes. Should be fine since add-ons should still be regarded as experimental for now IMO.

> This is because, the components in the sub-charts should also get the proxy injected, so they use these templates, functions in partials to achieve the same just like other control-plane components.

:+1:

Pothulapati:
Closing this because, we got all the following Grafana add-on PR's  merged!

- [x]    [ Refactor of linkerd-config-addons to make it more cleaner and reusable.](https://github.com/linkerd/linkerd2/pull/4318)
- [x]     [Make grafana optional in linkerd-web](https://github.com/linkerd/linkerd2/pull/4319)
- [x]     [Grafana Add-on sub-chart and relevant install and upgrade tests.](https://github.com/linkerd/linkerd2/pull/4320)
- [x]     [Health checks for Grafana add-on.](https://github.com/linkerd/linkerd2/pull/4321)
- [x]     [Add grafanaUrl to the configuration, to support bringing your own Grafana use-case.](https://github.com/linkerd/linkerd2/pull/4381)

Issue #3907, opened 2020-01-09
Check reports about non-valid certificate in case of \n at the end of it


StupidScience:
### What is the issue?

As I mentioned before in #3880 there is bug with certificate comparison related to `\n` symbol at the end of certificate. And I believe it was misleading in those case with root cause investigation.

### How can it be reproduced?

* Install linkerd with helm so there will be no `\n` symbol in the end trust anchor certificate but `ca.crt` in certificate that was generated by cert-manager will have this certificate. In my case it looked like:
```yaml
identity:
  trustAnchorsPEM: |
    -----BEGIN CERTIFICATE-----
    ...
    -----END CERTIFICATE-----
  issuer:
    scheme: kubernetes.io/tls
```
* Run `linkerd check`
* Got 
```text
linkerd-identity
----------------
× certificate config is valid
    IdentityContext.TrustAnchorsPem does not match ca.crt in linkerd-identity-issuer
```

### Environment

- Kubernetes Version:
- Cluster Environment: (GKE, AKS, kops, ...)
- Host OS:
- Linkerd version: `edge-19.12.3`

### Possible solution

Before comparing of certificates just encode them and decode back to string so certificates will be in the same format.

### Additional context

I can implement it and send a PR if you confirm this possible solution.

zaharidichev:
I am not sure this is technically a bug. What is actually causing that is the fact that no matter how you have specified your trust roots, cert manager will always emit a value with a newline at  the end. I think best practice is to ensure that you specify the value (in your case in your helm install process) with a newline as well. Yet, I am not super opinionated about this. @adleong WDYT?

adleong:
Is this an issue with our helm templating?  If we do `--set-file Identity.TrustAnchorsPEM=ca.crt` in a Helm install, I would expect the CA in the Linkerd configmap to exactly match contents of the `ca.crt` file.  If the process of reading the file and writing it to the configmap strips a trailing newline from the file contents, I think we should consider that a bug.

StupidScience:
@adleong looks like that problem is in `trim` [here](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/_config.tpl#L8) and it trim `\n` in the end of file.
But I think that check should handle this and check that certificates are really same.

zaharidichev:
@StupidScience I will take a deeper look into that soon enough:)

zaharidichev:
[...]

PR #3982, opened 2020-01-28
Trim space when comparing roots between the issuer secret and the config

zaharidichev:
This fix ensures that we ignore whitespace and newlines when checking that roots match between the Linkerd config map and the issuer secret (in the case of using external issue + Helm). 

Fixes: #3907

### DIFF
--- a/pkg/healthcheck/healthcheck_test.go
+++ b/pkg/healthcheck/healthcheck_test.go
@@ -2395,7 +2395,7 @@ data:
 	}
 }
 
-func getFakeConfig(secretScheme string, schemeInConfig string, issuerCerts *issuercerts.IssuerCertData) []string {
+func getFakeConfig(secretScheme string, schemeInConfig string, issuerCerts *issuercerts.IssuerCertData, configAnchorsModifier func(string) string) []string {
 
 	var resources []string
 	base64.StdEncoding.EncodeToString([]byte(issuerCerts.IssuerCrt))
@@ -2427,8 +2427,11 @@ data:
 ---
 `, base64.StdEncoding.EncodeToString([]byte(issuerCerts.TrustAnchors)), base64.StdEncoding.EncodeToString([]byte(issuerCerts.IssuerCrt)), base64.StdEncoding.EncodeToString([]byte(issuerCerts.IssuerKey))))
 	}
-
-	anchors, _ := json.Marshal(issuerCerts.TrustAnchors)
+	anchorsToEncode := issuerCerts.TrustAnchors
+	if configAnchorsModifier != nil {
+		anchorsToEncode = configAnchorsModifier(anchorsToEncode)
+	}
+	anchors, _ := json.Marshal(anchorsToEncode)
 
 	resources = append(resources, fmt.Sprintf(`
 kind: ConfigMap
@@ -2468,13 +2471,14 @@ func TestLinkerdIdentityCheck(t *testing.T) {
 		ends   time.Time
 	}
 	var testCases = []struct {
-		checkDescription   string
-		certificateDNSName string
-		lifespan           *lifeSpan
-		tlsSecretScheme    string
-		schemeInConfig     string
-		expectedOutput     []string
-		checkerToTest      string
+		checkDescription      string
+		certificateDNSName    string
+		lifespan              *lifeSpan
+		tlsSecretScheme       string
+		schemeInConfig        string
+		expectedOutput        []string
+		checkerToTest         string
+		configAnchorsModifier func(string) string
 	}{
 		{
 			checkerToTest:    "certificate config is valid",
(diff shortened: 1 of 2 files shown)


pkg/healthcheck/healthcheck_test.go
```diff
@@ -2519,6 +2522,26 @@ func TestLinkerdIdentityCheck(t *testing.T) {
...
+			},
+		},
+		{
+			checkerToTest:    "certificate config is valid",
+			checkDescription: "fails when roots in config and secret do not match",
+			tlsSecretScheme:  string(corev1.SecretTypeTLS),
+			schemeInConfig:   string(corev1.SecretTypeTLS),
+			expectedOutput:   []string{"linkerd-identity-test-cat certificate config is valid: IdentityContext.TrustAnchorsPem does not match ca.crt in linkerd-identity-issuer"},
+			configAnchorsModifier: func(anchors string) string {
+				return "nonsense"
```
alpeb: In order to actually test that extra newlines at the end of the certs don't affect things, you might wanna change this with something like `return anchors + "\n"` and make it a positive test...

pkg/healthcheck/healthcheck_test.go
```diff
@@ -2395,7 +2395,7 @@ data:
 	}
 }
 
-func getFakeConfig(secretScheme string, schemeInConfig string, issuerCerts *issuercerts.IssuerCertData) []string {
+func getFakeConfig(secretScheme string, schemeInConfig string, issuerCerts *issuercerts.IssuerCertData, configAnchorsModifier func(string) string) []string {
```
adleong: Passing in a function that modifies just the config anchors for just the configmap feels hacky and shoehorned in here.  We should consider coming back and cleaning up/refactoring these tests later.

pkg/healthcheck/healthcheck_test.go
adleong: :+1:

zaharidichev:
@adleong @alpeb  I did the following to test all this: 

First install cert manager: https://cert-manager.io/docs/installation/kubernetes/

1. Generate root cert and create anchors secret: 

```
step certificate create identity.linkerd.cluster.local ca.crt ca.key \
  --profile root-ca --no-password --insecure &&
  kubectl create secret tls \
   linkerd-trust-anchor \
   --cert=ca.crt \
   --key=ca.key \
   --namespace=linkerd
```

2.  Create an issuer
```
cat <<EOF | kubectl apply -f -
apiVersion: cert-manager.io/v1alpha2
kind: Issuer
metadata:
  name: linkerd-trust-anchor
  namespace: linkerd
spec:
  ca:
    secretName: linkerd-trust-anchor
EOF
```

3. Create a certificate resource: 

```
cat <<EOF | kubectl apply -f -
apiVersion: cert-manager.io/v1alpha2
kind: Certificate
metadata:
  name: linkerd-identity-issuer
  namespace: linkerd
spec:
  secretName: linkerd-identity-issuer
  duration: 24h
  renewBefore: 1h
  issuerRef:
    name: linkerd-trust-anchor
    kind: Issuer
  commonName: identity.linkerd.cluster.local
  isCA: true
  keyAlgorithm: ecdsa
  usages:
  - cert sign
  - crl sign
  - server auth
  - client auth
EOF
```

4. Install with helm
```
 helm install \
  --name=linkerd2 \
  --set identity.issuer.scheme=kubernetes.io/tls \
  --set installNamespace=false \
  charts/linkerd2
```

5. Run `linkerd check` and you will see: 

```
linkerd-identity
----------------
× certificate config is valid
```

If you use this branch instead, you will not encounter the problem. You can also change the number of newlines in the ca.crt file in order to test what happens when there are more than one line breaks. One particular thing to notice is that one might expect that the problem is introduced by our `trim` function in the helm chart. If that was indeed the case, simply supplying a ca.crt file that has no line breaks would fix the problem. However if you try doing that and going through this procedure, you will convince yourself that this is not the case as cert-manager adds one \n at the end of the ca.crt in the `linkerd-identity-issuer` secret. In fact it is also interesting to observe that in this scenario the values of `tls.crt` in `linkerd-trust-anchors` and `ca.crt` in the `linkerd-identity-issuer` secrets differ in the number of their trailing \n chars. 

Hope this is all useful

pkg/healthcheck/healthcheck_test.go
zaharidichev: I agree. I dont like it. How about we do the following. I introduce one more modifier function that works on the value that is in the secret. This makes it a bit less akward (symmetry) and allows us to write tests for the reverse case (more tests never hurt)

adleong: I think the problem is that we're trying to fit all of our different test cases into the same structure, even when that's not a good fit.  This leads us to passing in the ability to do arbitrary transformations on various pieces.  It's not necessary that all tests fit into a single "table".  We can split out tests that have a different structure (e.g. where the configmap trust roots and the issuer secret trust roots differ) into their own test that doesn't need to re-use the same test functions.

But none of that blocks this PR.  We can come back and clean this up after.

adleong:
Thanks for all the detail, @zaharidichev, that is very helpful!

pkg/healthcheck/healthcheck_test.go
zaharidichev: Yes that makes total sense. I also agree it will be better if split. Thanks for the tips !

PR #3987, opened 2020-01-28
Remove dependency on httpbin in egress integration test

adleong:
The egress integration test works by running a set of `bb` pods in `egress` mode which each are configured to send requests to a different combination of method and url in the `httpbin.org` domain.  We then verify certain properties of the response.

This is problematic because our integration tests rely on the availability and specific behavior of a service outside of our control.  Furthermore, using `bb` as part of this test can be problematic because `bb` sends a body with each request (including GETs) which is not handled well by all webservers.

We simplify this test by running a single `linkerd-debug` pod instead of a set of `bb` pods.  Our test can then run `curl` commands from inside this container using `kubectl exec`.  We use `linkerd.io` instead of `httpbin.org` as our egress target and we relax the check to simply verify that a non-error response is received.

Fixes https://github.com/linkerd/linkerd2/issues/2316


testutil/test_helper.go
```diff
@@ -281,6 +281,28 @@ func (h *TestHelper) RetryFor(timeout time.Duration, fn func() error) error {
 	}
 }
 
+// HTTPGetCode sends a GET request to the given URL and returns the response code.
+// It retries requests for up to 30 seconds, giving pods time to start.
+func (h *TestHelper) HTTPGetCode(url string) (int, error) {
```
kleimkuhler: Is this still needed for this PR? It looks except for the retries, the correct flags passed in to `curl` returns the response code.

test/egress/egress_test.go
```diff
@@ -57,43 +39,31 @@ func TestEgressHttp(t *testing.T) {
...
+		testName := fmt.Sprintf("Can use egress to send %s request to (%s)", methodToUse, url)
 		t.Run(testName, func(t *testing.T) {
-			expectedURL := fmt.Sprintf("%s://%s/%s", protocolToUse, dnsName, strings.ToLower(methodToUse))
-
-			url, err := TestHelper.URLFor(prefixedNs, deployName, 8080)
-			if err != nil {
-				t.Fatalf("Failed to get proxy URL: %s", err)
+			cmd := []string{
+				"-n", prefixedNs, "exec", "deploy/egress-test", "-c", "http-egress",
+				"--", "curl", "-sko", "/dev/null", "-w", "%{http_code}", "-X", methodToUse, url,
```
kleimkuhler: Cool use of `curl` flags here. Didn't know about `-w %{http_code}`.

Issue #3980, opened 2020-01-27
Make expired certs more diagnosable when clock skew exists


adleong:
One of the things that `linkerd check` checks is that the issuer certificate is valid and not expired.  It does this by comparing the expiration date on the certificate to the current time on the machine running the Linkerd CLI.  However, in the case where there is significant clock skew between the machine running the CLI and the node running the identity controller, the `linkerd check` command can succeed while the identity controller will consider the certificate to be expired.

The current behavior if the issuer certificate is out of the validity window due to clock skew is that the linkerd-identity controller's logs fill up with this error message:

```
time="2020-01-23T21:12:23Z" level=error msg="could not process CSR because of CA cert validation failure: x509: certificate has expired or is not yet valid"
```

We should make the problem more diagnose-able by doing the following:

* Put the expiry time in the error message
* Put the "current" time in the error message
* Emit the error message as a kubernetes event

mayankshah1607:
@adleong Is it okay if I take up this issue?

zaharidichev:
@mayankshah1607 go for it! Let me know if you have any questions :)

mayankshah1607:
Thanks @zaharidichev ! I have opened a draft and left a note. Mind taking a look? - https://github.com/linkerd/linkerd2/pull/3990#issuecomment-579694290

PR #3990, opened 2020-01-29
Update identity to make certs more diagnosable

mayankshah1607:
fixes https://github.com/linkerd/linkerd2/issues/3980

Update identity controller to make issuer certificates diagnosable if
cert validity is causing error

- Add expiry time in identity log message
- Add current time in identity log message
- Emit k8s event with appropriate message

On certificate expiry, the logs at `linkerd-identity` pod now look like :
```bash
time="2020-02-12T11:45:05Z" level=error msg="could not process CSR because of CA cert validation failure: x509: certificate has expired or is not yet valid - Current Time : 2020-02-12 11:45:05.461157338 +0000 UTC m=+69.039740599 - Invalid before 2020-02-12 11:42:00 +0000 UTC - Invalid After 2020-02-12 11:45:00 +0000 UTC - CSR Identity : linkerd-identity.linkerd.serviceaccount.identity.linkerd.cluster.local"
```
Event emitted for `linkerd-identity` deployment:
|Type|Reason|Age|From|Message|
|------|-----------|------|------|-------------|
|Warning|IssuerValidationFailed|5m54s|linkerd-identity|x509: certificate has expired or is not yet valid - Current Time : 2020-02-12 11:45:05.461157338 +0000 UTC m=+69.039740599 - Invalid before 2020-02-12 11:42:00 +0000 UTC - Invalid After 2020-02-12 11:45:00 +0000 UTC - CSR Identity : linkerd-identity.linkerd.serviceaccount.identity.linkerd.cluster.local|

### DIFF
--- a/pkg/identity/service.go
+++ b/pkg/identity/service.go
@@ -29,6 +29,7 @@ const (
 	EnvTrustAnchors  = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS"
 	eventTypeSkipped = "IssuerUpdateSkipped"
 	eventTypeUpdated = "IssuerUpdated"
+	eventTypeFailed  = "IssuerValidationFailed"
 )
 
 type (
@@ -113,7 +114,7 @@ func (svc *Service) loadCredentials() (tls.Issuer, error) {
 		return nil, fmt.Errorf("failed to read CA from disk: %s", err)
 	}
 
-	if err := creds.Crt.Verify(svc.trustAnchors, svc.expectedName); err != nil {
+	if err := creds.Crt.Verify(svc.trustAnchors, svc.expectedName, time.Time{}); err != nil {
 		return nil, fmt.Errorf("failed to verify issuer credentials for '%s' with trust anchors: %s", svc.expectedName, err)
 	}
 
@@ -148,7 +149,7 @@ func (svc *Service) ensureIssuerStillValid() error {
 	issuer := *svc.issuer
 	switch is := issuer.(type) {
 	case *tls.CA:
-		return is.Cred.Verify(svc.trustAnchors, svc.expectedName)
+		return is.Cred.Verify(svc.trustAnchors, svc.expectedName, time.Time{})
 	default:
 		return fmt.Errorf("unsupported issuer type. Expected *tls.CA, got %v", is)
 	}
@@ -171,7 +172,9 @@ func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.Ce
 	}
 
 	if err := svc.ensureIssuerStillValid(); err != nil {
-		log.Errorf("could not process CSR because of CA cert validation failure: %s", err)
+		log.Errorf("could not process CSR because of CA cert validation failure: %s - CSR Identity : %s", err, reqIdentity)
+		message := fmt.Sprintf("%s - CSR Identity : %s", err.Error(), reqIdentity)
+		svc.recordEvent(v1.EventTypeWarning, eventTypeFailed, message)
 		return nil, err
 	}
 
--- a/pkg/tls/cred.go
+++ b/pkg/tls/cred.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"time"
 )
 
 type (
@@ -83,13 +84,21 @@ func (crt *Crt) CertPool() *x509.CertPool {
 }
 
 // Verify the validity of the provided certificate
-func (crt *Crt) Verify(roots *x509.CertPool, name string) error {
+func (crt *Crt) Verify(roots *x509.CertPool, name string, currentTime time.Time) error {
 	i := x509.NewCertPool()
 	for _, c := range crt.TrustChain {
 		i.AddCert(c)
 	}
-	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name}
+	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name, CurrentTime: currentTime}
 	_, err := crt.Certificate.Verify(vo)
+
+	if currentTime.IsZero() {
+		currentTime = time.Now()
+	}
+
+	if crtExpiryError(err) {
+		return fmt.Errorf("%s - Current Time : %s - Invalid before %s - Invalid After %s", err, currentTime, crt.Certificate.NotBefore, crt.Certificate.NotAfter)
+	}
 	return err
 }
 
@@ -220,4 +229,13 @@ func DecodePEMCrt(txt string) (*Crt, error) {
 	}
 
 	return &crt, nil
+}
+
+func crtExpiryError(err error) bool {
+	switch v := err.(type) {
+	case x509.CertificateInvalidError:
+		return v.Reason == x509.Expired
+	default:
+		return false
+	}
 }
--- a/pkg/tls/cred_test.go
+++ b/pkg/tls/cred_test.go
@@ -2,6 +2,7 @@ package tls
 
 import (
 	"testing"
+	"time"
 )
 
 func newRoot(t *testing.T) CA {
@@ -26,7 +27,7 @@ func TestCrtRoundtrip(t *testing.T) {
 		t.Fatalf("Failed to decode PEM Crt: %s", err)
 	}
 
-	if err := crt.Verify(rootTrust, "endentity.test"); err != nil {
+	if err := crt.Verify(rootTrust, "endentity.test", time.Time{}); err != nil {
 		t.Fatal("Failed to verify round-tripped certificate")
 	}
 }
--- a/pkg/tls/cred_test.go
+++ b/pkg/tls/cred_test.go
@@ -1,6 +1,8 @@
 package tls
 
 import (
+	"crypto/x509"
+	"fmt"
 	"testing"
 	"time"
 )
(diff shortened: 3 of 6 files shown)


pkg/tls/cred.go
```diff
@@ -90,6 +91,14 @@ func (crt *Crt) Verify(roots *x509.CertPool, name string) error {
 	}
 	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name}
 	_, err := crt.Certificate.Verify(vo)
+
+	//If error due to expiry
+	if err == fmt.Errorf("x509: certificate has expired or is not yet valid") {
```
mayankshah1607: Is there a better way to check if the error is due to expiry?

mayankshah1607:
@zaharidichev @adleong left a small note; might need some suggestions on that. Also, I haven't yet tested these changes as I was not able to reproduce this error. Could you help me out with that as well??

pkg/tls/cred.go
zaharidichev: Very good question!. So the verify method returns a special type of error which is [`CertificateInvalidError`](https://github.com/golang/go/blob/master/src/crypto/x509/verify.go#L72) It looks like: 

```
type CertificateInvalidError struct {
	Cert   *Certificate
	Reason InvalidReason
	Detail string
}
```

What you can do is to check whether the error returned is of this type and if it is check the `Reason` field. If it equals to `Expired` and there is a type for that already, you can assume that this is the reason. Also this time validity of the cert can be bad not only due to expiration. It can be the case that this cert is still not valid (interesting why they named the const Expired...) Anyhow, that is all to say that maybe in your messag you should include the NotBefore as well.

zaharidichev:
@mayankshah1607 Do you want me to explain how to test that with unit tests or manually ?

mayankshah1607:
@zaharidichev I think it'll be easier for me to test this out using unit tests. Once the changes look good, I'll try to reproduce the cert expiry scenario manually and make sure it works. WDYT?

zaharidichev:
@mayankshah1607 Makes sense! The tests you probably want to take a look at are in service_test.go and cred_test.go. These will require some modification to make them test that code path. Let me know if you need help :)

mayankshah1607:
@zaharidichev I've made the required changes. Also added a test in `creds_test.go`. Everything seems to be working as expected! Could you take a look?
If things look good, I'll try and replicate the expiry situation manually on a cluster to see if the k8s event is being emitted. :smile:

pkg/tls/cred.go
```diff
@@ -90,7 +91,17 @@ func (crt *Crt) Verify(roots *x509.CertPool, name string) error {
 	}
 	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name}
 	_, err := crt.Certificate.Verify(vo)
-	return err
+
+	//If error due to expiry
+	switch err.(type) {
+	case x509.CertificateInvalidError:
+		expTime := crt.Certificate.NotAfter
```
zaharidichev: Two things that stand out. I think here you need to test whether the reason in the `CertificateInvalidError` is `Expired` as I already mentioned. This cert can be invalid for other reasons. Also, its good to include the NotBefore time as well as it could be that its just not valid yet.

pkg/tls/cred_test.go
```diff
@@ -47,3 +48,27 @@ func TestCredEncodeCertificateAndTrustChain(t *testing.T) {
...
+
+	//explicitly kill the certificate
+	crt.Certificate.NotBefore = time.Now().Add(time.Duration(100000000000))
+
+	if err := crt.Verify(rootTrust, "expired.test"); err != nil {
+		t.Logf("could not process CSR because of CA cert validation failure: certificate: %s", err)
+	} else {
+		t.Fatalf("failed to detect expired certificates")
+	}
+}
```
zaharidichev: This test is a bit strange. In your code you are emitting an event that includes the now time and the expiration time. But here you are killing the cert by moving the NotBefore date further away in time, effectivelly making it not expired but simply not valid yet.. Also you are simply testing that an error is present. I think you want to test for the specific type of error. So the following cases make sense to be covered: 
1. Test that the cert is not valid yet - NotBefore now + 10 days, notAfter now + 20 days 
2. Test that the cert is expired - NotBefore now - 20 daysm NotAfter now - 10 days
3. Make sure you get an exact error, not simply any error.

mayankshah1607: @zaharidichev Sure, I understood. I just wanted to clear my understanding a bit before I write the tests.
I looked into the `x509` library to understand what triggers the Expiry error:
```
	if now.Before(c.NotBefore) || now.After(c.NotAfter) {
		return CertificateInvalidError{c, Expired, ""}
	}
```
The Expiry error is triggered when 1) The current time is before `NotBefore` (the certificate is not valid yet), or 2) the current time is after `NotAfter` (the current time has expired).
Either of the cases will cause the certificate to be Expired. I seem to have written only the case (1). Need to cover (2) as well, right?

mayankshah1607: @zaharidichev Made the changes. Could you have a look and let me know if I've got the tests right?

pkg/identity/service.go
```diff
@@ -172,6 +173,8 @@ func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.Ce
 
 	if err := svc.ensureIssuerStillValid(); err != nil {
 		log.Errorf("could not process CSR because of CA cert validation failure: %s", err)
+		message := err.Error()
```
zaharidichev: you can just inline that in the record event invocation: 

```
svc.recordEvent(v1.EventTypeWarning, eventTypeFailed, err.Error())
```

pkg/identity/service.go
```diff
@@ -172,6 +173,8 @@ func (svc *Service) Certify(ctx context.Context, req *pb.CertifyRequest) (*pb.Ce
...
+		svc.recordEvent(v1.EventTypeWarning, eventTypeFailed, message)
```
zaharidichev: I think it makes sense to include some infomation, indicating where the CSR is coming from. If the validation fails, then we would like to know which service actually issued the CSR request. Maybe the `Identity` field of the CSR is a good source for that information

pkg/tls/cred.go
```diff
@@ -221,3 +227,10 @@ func DecodePEMCrt(txt string) (*Crt, error) {
 
 	return &crt, nil
 }
+
+func crtExpiryError(err error) bool {
+	if reflect.TypeOf(err) == reflect.TypeOf(x509.CertificateInvalidError{}) {
+		return err.(x509.CertificateInvalidError).Reason == x509.Expired
+	}
+	return false
+}
```
zaharidichev: TIOLI: you can rewrite that with a switch statement: 

```
func crtExpiryError(err error) bool {
	switch v := err.(type) {
	 case x509.CertificateInvalidError :
	 	return v.Reason == x509.Expired
	default:
		return false
	}
}
```

pkg/tls/cred_test.go
```diff
@@ -47,3 +50,68 @@ func TestCredEncodeCertificateAndTrustChain(t *testing.T) {
...
+	testCases := []struct {
+		notBefore    time.Time
+		notAfter     time.Time
+		shouldExpire bool
+	}{
+		//cert not valid yet
+		{
+			notAfter:     time.Now().AddDate(0, 0, 20),
+			notBefore:    time.Now().AddDate(0, 0, 10),
+			shouldExpire: true,
```
zaharidichev: I think the tests should be a bit more specific. Namely. The firrst two tests only check whether the error is of type `expired`. However they dont check the specific type of time constraints violation. Is it really not valid yet? I think now that you control the type of error message emitted, you can simply go ahead and test on that. You know what the error format will be so in your test case struct instead of simply having a boolean to expect upon, have an err type or a string and check the explicit error message. No need to use the `crtExpiryError`

zaharidichev:
@adleong you mentioned in the original issue that 

> the linkerd-identity controller's logs fill up with this error message

I wonder , would it make sense to put some kind of throttle somewhere, slowing doesn the rate at which these CSRs are issued in case the identitiy service fails to serve it correctly. Otherwise I would assume that we are still going to see quite a bit of log messages + k8s events?

adleong:
I think we'll see this error message once for each time a proxy in the cluster attempts to get its identity.  That can be quite a lot, but I don't see any particular reason why we need to throttle it.  Do you think there might be other useful log lines that could get buried or hidden by this one?

pkg/identity/service.go
mayankshah1607: I think the preceding line 169 does this:
```go
	reqIdentity, tok, csr, err := checkRequest(req)
	if err != nil {
		return nil, status.Error(codes.InvalidArgument, err.Error())
	}
```
We could use `reqIdentity` in the message

pkg/tls/cred_test.go
```diff
@@ -47,3 +50,68 @@ func TestCredEncodeCertificateAndTrustChain(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to decode PEM Crt: %s", err)
+	}
+
+	fakeExpiryError := x509.CertificateInvalidError{Reason: x509.Expired}
+
+	//need to remove seconds and nanoseconds for testing returned error
+	now := time.Now()
+	curTimeDate := time.Date(now.Year(), now.Month(), now.Day(), now.Hour(), now.Minute(), 0, 0, now.Location())
+
```
mayankshah1607: @zaharidichev I've updated the tests to match the returned error. However, there was a small catch. `time.Now()` when the test cases are produced and `time.Now()` when the `Verify()` method is called has a difference of seconds, causing a mismatch in the test cases. 
I had to ignore seconds and nanoseconds in both the `time.Now()` to pass the test cases.

pkg/tls/cred.go
```diff
@@ -90,6 +91,15 @@ func (crt *Crt) Verify(roots *x509.CertPool, name string) error {
 	}
 	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name}
 	_, err := crt.Certificate.Verify(vo)
+
+	if crtExpiryError(err) {
+
+		//need to remove seconds and nanoseconds for testing returned error
```
adleong: so one way to make this testable without needing to throw away precision in the actual codepath would be to define a new error struct that this method could return.  The struct would implement the error interface and would hold a Time field.  The Error() method would do the formatting that you have here.

The advantage of this is that when we call `Verify` in our tests, we can then cast the error to our known error struct, extract the Time object, and compare it to the current Time with a certain error tolerance.

mayankshah1607: That's a great idea @adleong ! I've made this change, could you take a look?

mayankshah1607:
@adleong @zaharidichev Might need some help with replicating cert expiry error manually to capture the logs and event. Any suggestions on how I can kill the certs manually on a test cluster?

pkg/tls/cred.go
zaharidichev: @adleong @mayankshah1607  Just took a look and I think there is a better way to do this. So currently the verify method looks like: 

```go
func (crt *Crt) Verify(roots *x509.CertPool, name string) error {
	i := x509.NewCertPool()
	for _, c := range crt.TrustChain {
		i.AddCert(c)
	}
	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name}
	_, err := crt.Certificate.Verify(vo)

	if _, ok := crtExpiryError(err, *crt.Certificate); ok {
		return fmt.Errorf("%s - Current Time : %s - Invalid before %s - Invalid After %s", err, time.Now(), crt.Certificate.NotBefore, crt.Certificate.NotAfter)
	}
	return err
}
```

we can change the signature to be `func (crt *Crt) Verify(roots *x509.CertPool, name string, currentTime *time.Time)`. This will allow us to actually insert this time (if not nil into ) VerifyOptions.CurrentTime. Much more predictable than trying to shave off precision and fight the wall clock. WDYT ?

zaharidichev: @mayankshah1607 Here is a [diff](https://gist.github.com/zaharidichev/0ec8d4a02fac5becf555e7718893d4a8) pointing to what I mean if you want to explore that avenue

mayankshah1607: @zaharidichev this idea too looks good. I did try something like this before, but had to discard it because of the refactoring that would follow. But if that's not an issue, I think we could very well go ahead with this approach. @adleong WDYT? I'll push these changes right away if I can have your :+1:

mayankshah1607: Although one small concern - I think the `x509.Verify` method calls `time.Now` independent of it's parent `Verify` method, which may lead to inconsistency of nanoseconds (which probably might not matter); i.e `time.Now` which actually triggers the expiry error is a few nanoseconds ahead of what's displayed on the log message.

zaharidichev: @mayankshah1607 if you are refering to this here: https://github.com/golang/go/blob/master/src/crypto/x509/verify.go#L576, the method is called only if the CurrentTime in opt is a zero value one, meaning that if you supply our own CurrentTime in the options, it would use that. Am I missing something ?

mayankshah1607: Got it! Did not strike me that we could just add the time into `opts` from the func args. :+1:

adleong: This is a good idea!

adleong:
I think the way I've tested this in the past is to [generate a certificate manually](https://linkerd.io/2/tasks/generate-certificates/) with a very short expiry (around 1 to 5 minutes) and then pass that to `linkerd install`.  The cert will be valid during install, but will expire shortly after.  We should then start seeing these errors.

pkg/tls/cred.go
```diff
@@ -83,13 +84,21 @@ func (crt *Crt) CertPool() *x509.CertPool {
 }
 
 // Verify the validity of the provided certificate
-func (crt *Crt) Verify(roots *x509.CertPool, name string) error {
+func (crt *Crt) Verify(roots *x509.CertPool, name string, currentTime time.Time) error {
 	i := x509.NewCertPool()
 	for _, c := range crt.TrustChain {
 		i.AddCert(c)
 	}
 	vo := x509.VerifyOptions{Roots: roots, Intermediates: i, DNSName: name}
```
zaharidichev: you need to pass the `currentTime` into the `VerifyOptions`

mayankshah1607: yes, wasn't yet ready for review - was just trying out something else because I ran into some rebasing errors. Anyway, its ready for review now :smile:

mayankshah1607:
@zaharidichev @adleong I have updated the tests according to @zaharidichev 's suggestion. Please have a look now. :smile: 
I've also updated the PR description and included the log message and emitted k8s event.

pkg/tls/cred.go
```diff
@@ -221,3 +230,12 @@ func DecodePEMCrt(txt string) (*Crt, error) {
 
 	return &crt, nil
 }
+
+func crtExpiryError(err error) bool {
```
zaharidichev: If not using that in another place, you can just inline it I think.

mayankshah1607:
@zaharidichev done! :)

PR #3996, opened 2020-01-29
Refactoring to suppress eslint warnings

### LINKED ISSUE #3927: Enable the no-param-reassign eslint rule
In #3882, we adopted more eslint rules for our Javascript codebase. The `no-param-reassign` rule is recommended, but it wasn't enabled as part of that PR, since it requires code changes. We should enable the rule and fix the code that fails to lint. We can enable it by removing the override that appears here:

https://github.com/linkerd/linkerd2/blob/0eabbe4173fd8a7da91d5771572bd08d85268275/web/app/.eslintrc#L85-L88

kohsheen1234:
Enabling the eslint/no-param-reassign throws some warnings with the existing code.

Made necessary changes to suppress the warnings

Fixes #3927


web/app/js/components/MetricsTable.jsx
```diff
@@ -217,7 +217,8 @@ const preprocessMetrics = metrics => {
 
   _each(tableData, datum => {
     _each(datum.latency, (value, quantile) => {
-      datum[quantile] = value;
+      const datum_ = datum;
+      datum_[quantile] = value;
```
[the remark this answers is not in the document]
kohsheen1234: @cpretzer I am not very sure of this case, but I guess it would work as expected. :)

kohsheen1234:
@Pothulapati I've done my rebase. Please have a look

web/app/js/components/MetricsTable.jsx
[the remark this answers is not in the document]
kohsheen1234: @Pothulapati and @cpretzer . There is no need to return 'datum' . I have verified this by interchanging datum by datum_ and the function works as expected.

` const preprocessMetrics = metrics => {
  const tableData = _cloneDeep(metrics);
  _each(tableData, datum_ => {
    _each(datum_.latency, (value, quantile) => {
      const datum = datum_;
      datum[quantile] = value;
    });
  });
  return tableData;
};
`
So, I do not see any need of making further changes. :)

kohsheen1234:
@Pothulapati and @cpretzer . There is no need to return 'datum' . I have verified this by interchanging datum by datum_ and the function works as expected.

` 
const preprocessMetrics = metrics => {
  const tableData = _cloneDeep(metrics);
  _each(tableData, datum_ => {
    _each(datum_.latency, (value, quantile) => {
      const datum = datum_;
      datum[quantile] = value;
    });
  });
  return tableData;
};
`

So, I do not see any need of making further changes. Also I have done my rebase, there are no conflicts any more :)

cpretzer:
@kohsheen1234 thanks this the note! 

I observed the same behavior when I tested the code and was looking to understand why it works that way. I _think_ it has to do with how variables are assigned.

Anyway, everything works, so thanks for the PR! 🙌

web/app/.eslintrc
```diff
@@ -84,6 +84,6 @@
 
     // These rules represent best practices that we're not adhering to, and should be enabled
     "jsx-a11y/click-events-have-key-events": 0,
-    "no-param-reassign": 0
+    "no-param-reassign": 1
```
alpeb: I think it's better to just delete this line to enable that rule

web/app/js/components/ErrorModal.jsx
```diff
@@ -52,20 +52,21 @@ class ErrorModal extends React.Component {
 
     const byPodAndContainer = _map(errorsByPod, (podErrors, pod) => {
       const byContainer = _reduce(podErrors.errors, (errors, err) => {
+        const errors__ = errors;
```
alpeb: nit: could you use a single underscore, like you did in your other changes?

alpeb:
Actually, I take my previous comment back :blush: 
When doing `const datum_ = datum;` the new var receives a reference, not a copy of `datum`, so mutating it does mutate the original one.
That being said, I believe creating new variables is sort of missing the spirit of the rule. I think it's fine, when the argument is an object, to allow to modify its properties. So I would suggest we relax the rule with `{ "props": true }`. What do you think?

cpretzer:
> Actually, I take my previous comment back 😊
> When doing `const datum_ = datum;` the new var receives a reference, not a copy of `datum`, so mutating it does mutate the original one.
> That being said, I believe creating new variables is sort of missing the spirit of the rule. I think it's fine, when the argument is an object, to allow to modify its properties. So I would suggest we relax the rule with `{ "props": true }`. What do you think?

👍 I think that makes sense and makes for more readable/understandable code.

kohsheen1234:
> Actually, I take my previous comment back 😊
> When doing `const datum_ = datum;` the new var receives a reference, not a copy of `datum`, so mutating it does mutate the original one.
> That being said, I believe creating new variables is sort of missing the spirit of the rule. I think it's fine, when the argument is an object, to allow to modify its properties. So I would suggest we relax the rule with `{ "props": true }`. What do you think?

By default, in the rule eslint/no-param-reassign we have, `{ "props": false}` , which means that we can directly assign values to the properties of the object which is passed as an argument. By making`{"prop":true}`, we will simply make the rule more strict instead of making it relaxed.

params

web/app/.eslintrc
kohsheen1234: @alpeb  I will do the same, thank you for the suggestion :)

web/app/js/components/ErrorModal.jsx
kohsheen1234: Surely, I'll use a single underscore. :)

alpeb:
> By default, in the rule eslint/no-param-reassign we have, `{ "props": false}` , which means that we can directly assign values to the properties of the object which is passed as an argument. By making`{"prop":true}`, we will simply make the rule more strict instead of making it relaxed.

You're right. But then I'm puzzled: why was something like `NetworkGraph.jsx` giving warnings if we were just changing object properties?

web/app/.eslintrc
kohsheen1234: By using this line `"no-param-reassign": ["error", { "props": false }] ` , we can reassign the properties of the objects that are passed as parameters.

alpeb: That looks good :+1:

kohsheen1234:
> > By default, in the rule eslint/no-param-reassign we have, `{ "props": false}` , which means that we can directly assign values to the properties of the object which is passed as an argument. By making`{"prop":true}`, we will simply make the rule more strict instead of making it relaxed.
> 
> You're right. But then I'm puzzled: why was something like `NetworkGraph.jsx` giving warnings if we were just changing object properties?

@alpeb Even I got confused, I don't know why it shows warnings. But I realized that using this line `"no-param-reassign":` ["error", { "props": false }]` does not show those warnings. 

Also, I had a doubt: After making changes by assigning values to properties of the object, I feel the dashboard isn't working as expected, though I am not sure about the same.

This is what I get after making the necessary changes. The first one is showing 0/2.

deploy

alpeb:
Thanks fine :slightly_smiling_face: 
It just means the `coredns` pod has two replicas, none of which has been injected by linkerd, which is expected since none of the pods from the kube-system namespace should be injected.

kohsheen1234:
@alpeb Thank you for the clarification. 👍  I have made changes as requested in the Pull Request. Can you just check if any further changes are required. :)

web/app/js/components/MetricsTable.jsx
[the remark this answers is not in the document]
alpeb: After having added the `"no-param-reassign": ["error", { "props": false }]` rule, you no longer need to edit this file, nor `Tap.jsx`  :wink:

web/app/js/components/TopModule.jsx
```diff
@@ -202,37 +202,38 @@ class TopModule extends React.Component {
   }
 
   indexTopResult = (d, topResults) => {
+    const topResults_ = topResults;
```
alpeb: This change is no longer required either

kohsheen1234: Yes. You are right. :) Just made the changes.

kohsheen1234:
@alpeb Can you check now. I have made the changes you told. Thank you very much for helping me with all these changes :)

web/node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
+  "systemParams": "linux-x64-57",
+  "modulesFolders": [],
+  "flags": [],
+  "linkedModules": [],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
alpeb: Could you please remove this file?

web/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
alpeb: Could you also please remove this file?

web/app/js/components/util/Utils.js
```diff
@@ -119,10 +121,11 @@ export const toClassName = name => {
  Create regex string from user input for a filter
 */
 export const regexFilterString = input => {
+  let input_ = input;
   // make input lower case and strip out unwanted characters
-  input = input.replace(/[^A-Z0-9/.\-_*]/gi, '').toLowerCase();
+  input_ = input_.replace(/[^A-Z0-9/.\-_*]/gi, '').toLowerCase();
```
alpeb: I think you can transform these into a single liner:
```javascript
let input_ = input.replace(/[^A-Z0-9/.\-_*]/gi, '').toLowerCase();
```

kohsheen1234: Done :)

web/yarn.lock
kohsheen1234: Done

web/node_modules/.yarn-integrity
kohsheen1234: Done :)

kohsheen1234:
> Thanks @kohsheen1234 !
> I've just got a few more minor comments, and then this should be ready to merge

@alpeb Thank you very much. :) . I made the new changes as well.

kohsheen1234:
> Thanks @kohsheen1234 for addressing the feedback so quickly 
> 
> As a first time contributor we'd love to send you some swag! Do you mind filling out this [form](https://docs.google.com/forms/d/19GLBNfu8XnIETrwEI3ijoWlocAiDp3-j7CW3l4cTnyc/viewform)?

Thank you very much @alpeb for your guidance. I have filled the form for some swag. :+1:

PR #4000, opened 2020-01-30
Move controlPlaneTracing helm field to globals

Pothulapati:
This already been moved into global as in https://github.com/linkerd/linkerd2/blob/a54c5b6b6553e8737a8787903ff5581ea54fcea3/charts/partials/templates/_trace.tpl#L2

but the change was missed in `values.yaml`

@alpeb


alpeb:
@Pothulapati can you please remove its unqualified version down below that same file?

Pothulapati:
:man_facepalming:  Oops Sorry @alpeb
Updated.

Issue #3635, opened 2019-10-24
Make CI workflow resilient to multiple runs on the same sha or tag


siggy:
## Background

PR #3602 guaranteed that all jobs in a CI workflow received the same copy of the repo. This made our CI workflow resilient to code changes occurring during the workflow run, both in master and the PR itself.

## Problem

Our workflow uses the sha of the repo to name global objects that are shared between jobs within a workflow. When two workflow runs occur on the same sha or tag, these two runs will collide. Specifically, in the names of the kind clusters and git repo artifacts they create.

### Collision examples

- kind cluster collision: https://github.com/linkerd/linkerd2/runs/273402477#step:5:15
- git repo artifact collision: https://github.com/linkerd/linkerd2/runs/272823368#step:2:7

## Proposal

Assign unique names to kind clusters and git repo artifacts, based on the current workflow run, in addition to the output of `bin/root-tag`.

For example, instead of:
```
TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
export KIND_CLUSTER=github-$TAG-${{ matrix.integration_test }}`
```
do something like:
```
TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
export KIND_CLUSTER=github-$TAG-${{ runner.tracking_id }}-${{ matrix.integration_test }}`
```

(I'm not sure if `runner.tracking_id` exists or is too long, but we need something unique like this, or something in the `{{job}}` context.) More info at https://help.github.com/en/github/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions.

#### kind cluster name

https://github.com/linkerd/linkerd2/blob/397970e917afef601103ed024caba54b27a742b9/.github/workflows/workflow.yml#L312-L313

#### git Repo artifact

https://github.com/linkerd/linkerd2/blob/397970e917afef601103ed024caba54b27a742b9/.github/workflows/workflow.yml#L78-L79

#### kubeconfig file

Note this filename is based on the kind cluster name, so fixing kind cluster naming should fix this automatically:
https://github.com/linkerd/linkerd2/blob/397970e917afef601103ed024caba54b27a742b9/.github/workflows/workflow.yml#L391

alpeb:
I'm wondering when two runs on the same sha can occur. Is it only when an empty commit to a branch is pushed to trigger the build, when the branch still has a build running?

ihcsim:
> I'm wondering when two runs on the same sha can occur.

[...]

PR #4001, opened 2020-01-30
Allow CI to run concurrent builds in master

### LINKED ISSUE #3911: CI can't run concurrent builds in master

CI builds on the master branch trigger a `cloud_integration` job that runs the integration tests on a GKE cluster that has already been setup beforehand that is never torn down. When more than one build attempts to run tests in that cluster they interfere with one another.
Even if the build doesn't reach the `cloud_integration` step, if the build is canceled the `cloud_cleanup` job will run regardless, interfering with another build that might still be in the `cloud_integration` step.

`cloud_integration`' s `Run integration tests` step should not be triggered until the cluster is empty. This could be achieved by prepending another step that blocks the build through a `kubectl wait` on linkerd's main resources.

A drawback is that when more than one build would be waiting on that condition they might then both unblock simultaneously when the wait is over. One way to address that would be with an algorithm like this:
```
while {
  // block until the cluster is clean
  kubectl wait --for ... --timeout=-1s

  // concurrent builds will be unblocked simultaneously, but...
  sleep random seconds

  // Concurrent builds will hit this at different moments.
  // The first to hit this will be unblocked and will immediately
  // create a resource and thus leave the other builds blocked
  // (timeout=0 means it checks once and returns)
  if (kubectl wait --for ... --timeout=0) then break
}
```
Given the `Run integration tests` step does actually run multiple tests with cluster cleanup in-between, we might need to add one special resource at the beginning that only gets clean at the very end, and use it in the two `kubectl wait` above.

- We could create separate clusters for each build run with a name associated to the commit sha like we do with the `kind` clusters. That implies using some sort of cluster API. Something to evaluate in the future, maybe as part as a more general cross-cloud testing effort.
- We could still run the integration tests concurrently in the same cluster, including the sha in the namespaces names to avoid clashes. But I believe we can't currently run multiple linkerd instances in the same cluster.

Edit: fixed timeout `kubectl wait` calls

alpeb:
Fixes #3911

Refactors the `cloud_integration` test to run in separate GKE clusters
that are created and torn down on the fly.
It leverages a new "gcloud" github action that is also used to set up
gcloud in other build steps (`docker_deploy` and `chart_deploy`).

The action also generates unique names for those clusters, based on the
git commit SHA and `run_id`, a recently introduced variable that is
unique per CI run and available to all the jobs.
This fixes part of #3635 in that CI runs on the same SHA don't interfere
with one another (in the `cloud_integration` test; still to do for
`kind_integration`).

The "gcloud" GH action is supported by `.github/actions/gcloud/index.js`
that has a couple of dependencies. To avoid having to commit
`node_modules`, after every change to that file one must run
```bash
# only needed the first time
npm i -g @zeit/ncc

cd .github/actions/gcloud
ncc build index.js
```
which generates the self-contained file
`.github/actions/gcloud/dist/index.js`.
(This last part might get easier in the future after other refactorings
outside this PR).


alpeb:
I've just realized we might be able to use a `post` hook in the `cloud_integration` test to tear down the cluster and thus get rid of the `cloud_cleanup` job. That hook is undocumented but you can see it in action for example with `actions/checkout`...

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,81 @@
...
+    await exec.exec('gcloud config set core/project', [core.getInput('gcp_project')]);
+    await exec.exec('gcloud config set compute/zone', [core.getInput('gcp_zone')]);
+    await exec.exec('gcloud auth configure-docker --quiet');
+
+    if (core.getInput('create') || core.getInput('destroy')) {
+      const name = await getClusterName();
+      if (core.getInput('create')) {
+        await exec.exec('gcloud container clusters create',
+          [name, '--cluster-version', '1.15.7-gke.23', '--num-nodes', 1, '--machine-type',
+          'n1-standard-2', '--enable-network-policy']);
```
grampelberg: Can this be moved out into configuration? Feels like we'll want to change these without modifying some strings in the source. I'd recommend at least doing:

- `--preemptible`
- `--release-channel rapid`
- `--no-enable-stackdriver-kubernetes`
- `--no-enable-basic-auth`
- `--no-enable-legacy-authorization`

Should we open a separate ticket to have this test `--ha` as well?

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,81 @@
+const core = require('@actions/core');
```
grampelberg: As this appears to just be all `exec`. Should we just do it as a bash action instead? The added indirection feels weird.

.github/actions/gcloud/index.js
alpeb: Sounds good, I'll add those options. Yeah, I think `--ha` grants a separate ticket.

.github/actions/gcloud/index.js
alpeb: I agree bash would be the right language here. But custom actions can only be built in javascript, which run directly on the Github VM, or a container that would run its own sandbox. The latter is geared more towards isolated tasks, whereas javascript actions modify the VM environment and can interact better with the other steps in the job.

I tried doing this with a bash script inside a container, that would set the gcloud, kubectl and docker configs in a directory from the VM that is mounted in the container, and then in the other job steps reconfigure those tools to point to those non-standard config locations, but it was messy and let to lots of repetition.

grampelberg: Good call, I got the wrappers all confused.

.github/actions/gcloud/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  gcp_project:
+    required: true
+  gcp_zone:
+    required: true
+  create:
+    required: false
+    description: Set to true to create a new cluster. The linkerd CLI should have been installed in the runner beforehand.
+  destroy:
+    required: false
+    description: Set to false to create a new cluster. The linkerd CLI should have been installed in the runner beforehand.
```
kleimkuhler: Since this is not required, maybe this description should be `Set to true to destroy an existing cluster`?

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,81 @@
...
+      listeners: {
+          stdout: (data) => {
+              clientVersion = data.toString().trim()
+          }
+      }
+  });
+
+  // validate CLI version matches the repo
+  if (tag != clientVersion) {
+      throw `tag ${tag} differs from clientversion ${clientVersion}`
```
kleimkuhler: nit: `.. differs from clientVersion ..` or `.. differs from client version ..`?

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+  console.log('Linkerd CLI version:', tag)
+
+  // Last part is to distinguish runs on the same sha (run-id is unique per CI run).
+  // run-id has to be provided as an input because it turns out it's not available
+  // through github.context.run_id
+  const name = `testing-${tag}-${core.getInput('run-id')}`;
```
kleimkuhler: This is cool. Glad to see `run-id` being used here to solve the naming issue we have.

.github/workflows/workflow.yml
```diff
@@ -471,34 +438,25 @@ jobs:
     - name: Checkout code
       if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags')
       uses: actions/checkout@v2
-    - name: Configure gcloud
+    - name: Install linkerd CLI
```
kleimkuhler: So it looks like `gcloud .. create/delete` only need the Linkerd CLI to be installed for the action's `getClusterName` which relies on `linkerd version --client --short`.

In `cloud_integration` this makes sense, but it's not stricly necessary to have the CLI installed for `cloud_cleanup`. Maybe instead of the descriptions requiring that a CLI is installed, they just pass in `$TAG` as an input? That way we don't need to install the CLI since we can already rely on `$TAG` having the correct value.

.github/actions/gcloud/action.yml
siggy: should `create`/`delete` just be a single string input called `action`? feels like the api should not allow the user to set both of these bools to true or both to false. also consider documenting that if these are not set, the action's default behavior is just to configure the gcloud tooling.

.github/workflows/workflow.yml
```diff
@@ -361,28 +361,11 @@ jobs:
...
-          install_gcloud "$dir"
-          gcloud components install kubectl
-          # Configure gcloud with a service account.
-          set_gcloud_config "$GCP_PROJECT" "$GCP_ZONE" "$GKE_CLUSTER"
-          # Get a kubernetes context.
-          get_k8s_ctx "$GCP_PROJECT" "$GCP_ZONE" "$GKE_CLUSTER"
-        )
-        . "$dir/path.bash.inc"
-        gcloud auth configure-docker
-        bin/kubectl version --short
```
siggy: i think the removal of this code implies we can delete `bin/_gcp.sh`?

.github/workflows/workflow.yml
siggy: Yeah agree, I think decoupling the gcloud github action from linkerd cli makes sense. It's good to validate the Linkerd CLI version matches repo version, but that could/should be done elsewhere. Taking `$TAG` as input a step further, maybe consider just having a `name` input to specify the cluster name.

.github/actions/gcloud/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  run-id:
+    required: false
+    default: ${{ github.run_id }}
```
siggy: 💥

.github/actions/gcloud/action.yml
```diff
@@ -0,0 +1,21 @@
+name: gcloud
```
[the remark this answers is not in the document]
alpeb: Definitely. Itself could also go through CI to avoid having to do the `ncc` step manually.

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,81 @@
...
+  await exec.exec(`${process.env.HOME}/.linkerd version --client --short`, [], {
+      listeners: {
+          stdout: (data) => {
+              clientVersion = data.toString().trim()
+          }
+      }
+  });
+
+  // validate CLI version matches the repo
+  if (tag != clientVersion) {
```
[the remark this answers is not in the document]
alpeb: Good catch :+1: 
`eslint` doesn't make part yet of CI. But as per the previous comment, I can add that step into the workflow that will publish this action as standalone :wink:

.github/actions/gcloud/action.yml
alpeb: Ok, an enum for this makes more sense indeed :ok_hand:

alpeb: Yea, that description was off. I'm gonna end up merging that into an `action` field, as per Siggy's suggestion below.

kleimkuhler: Yep sounds good I like that plan!

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,81 @@
...
+          CI_FORCE_CLEAN: 1
+      },
+      listeners: {
+          stdout: (data) => {
+              tag = data.toString().trim()
+          }
+      }
+  });
+
+  await exec.exec(`${process.env.HOME}/.linkerd version --client --short`, [], {
```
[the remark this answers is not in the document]
alpeb: I'll be getting rid of `getClusterName()` in my next push, but still nevertheless a good point :+1:

.github/workflows/workflow.yml
alpeb: indeed :eye:

alpeb:
Thanks @grampelberg, @kleimkuhler and @siggy for the detailed review!
I've just pushed another iteration with the following changes:
- Got rid of cloud_cleanup by using a post hook in the gcloud action
- Removed cluster naming responsibility from the gcloud action
- Added more options to the gcloud action to better control the type of cluster created, with some sane defaults
- Removed `bin/_gcp.sh`
- Consolidated `.gitignore` statements

Once we're all ok with these changes, I'll proceed to move `actions/gcloud` into its own repo with a little CI for linting and for generating `dist/index.js` automatically.

alpeb:
Note: GH is still waiting for the "Cloud cleanup" step, but that's gone. Probably the same issue #4002 was having, i.e.  "Branch Protection Rules" that need to be modified by a user with enough access.

.github/actions/gcloud/action.yml
```diff
@@ -0,0 +1,47 @@
...
+    required: false
+    default: false
+  enable_legacy_auth:
+    required: false
+    default: false
+runs:
+  using: node12
+  main: dist/index.js
+  post: dist/index.js
+  post-if: always()
```
kleimkuhler: Does this `post_if` condition ensure the action runs again and hits the `isPost` branch that deletes the cluster?

alpeb: yep, that's correct. I did test having the cluster creation job fail, and the deletion post step does get triggered.

.github/actions/gcloud/index.js
```diff
@@ -0,0 +1,110 @@
...
+  try {
+    await configure();
+    if (core.getInput('create')) {
+      if (!core.getState(isPost)) {
+        core.saveState(isPost, 'true');
+        await create();
+      } else {
+        await destroy();
+      }
+    }
```
siggy: can we avoid having this `isPost` state, and instead just call separate endpoints for `runs/main` and `runs/post`?

.github/actions/gcloud/action.yml
```diff
@@ -0,0 +1,47 @@
...
+    default: true
+  enable_stackdriver:
+    required: false
+    default: false
+  enable_basic_auth:
+    required: false
+    default: false
+  enable_legacy_auth:
+    required: false
+    default: false
```
siggy: are all these extra flags intended to make this a drop-in replacement for another package? if not, consider dropping.

alpeb: Like a replacement for an existing 3rd party gcloud GH action? Nope, just following up on a [request](https://github.com/linkerd/linkerd2/pull/4001#discussion_r373560269) from @grampelberg

.github/actions/gcloud/index.js
alpeb: I followed the pattern used by `actions/checkout`, but you're right, that's unneeded complexity. I'll see if I can split this up.

alpeb:
Ok in my last few pushes I've spun-off the "destroy" post-action and moved all the action into a separate repo https://github.com/linkerd/linkerd2-action-gcloud, that itself has the CI to build the consolidated js files and full linting.
Please give it a final look before I merge :pray:

alpeb:
> This all looks great !  
> 
> Have you been able to test most of these jobs? Possibly by disabling some `if: startsWith(github.ref, 'refs/tags')` checks?

Yeah, I've been actually testing in a fork, using PRs and merging into the fork's master, using the appropriate secrets :+1:

Issue #3978, opened 2020-01-27
`linkerd metrics` panics if port-forward fails


olix0r:
```
Error running port-forward: error upgrading connection: unable to upgrade connection: pod not found ("PODNAME_PODNS")panic: close of closed channel

goroutine 24 [running]:
github.com/linkerd/linkerd2/pkg/k8s.(*PortForward).Stop(0xc000347860)
	/linkerd-build/pkg/k8s/portforward.go:207 +0x2f
github.com/linkerd/linkerd2/cli/cmd.getMetrics(0xc000411540, 0x0, 0x0, 0x0, 0x0, 0xc00072bce0, 0x17, 0xc00072bd00, 0x12, 0xc0005357d0, ...)
	/linkerd-build/cli/cmd/metrics.go:173 +0x437
github.com/linkerd/linkerd2/cli/cmd.newCmdMetrics.func1.1(0xc000411540, 0xc000772720, 0x0, 0x0, 0x0, 0x0, 0xc00072bce0, 0x17, 0xc00072bd00, 0x12, ...)
	/linkerd-build/cli/cmd/metrics.go:109 +0x71
created by github.com/linkerd/linkerd2/cli/cmd.newCmdMetrics.func1
```

should be handled gracefully

mayankshah1607:
@grampelberg @olix0r I'd like to work on this one if that's okay :smile:

grampelberg:
@mayankshah1607 go for it!

mayankshah1607:
@olix0r @grampelberg any suggestions on how I can reproduce this bug?

ihcsim:
@mayankshah1607 I think you can fake this error with this change:
```
diff --git a/pkg/k8s/portforward.go b/pkg/k8s/portforward.go
index 8f5e79f6..1dace80d 100644
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -143,6 +143,7 @@ func newPortForward(
 // run creates and runs the port-forward connection.
 // When the connection is established it blocks until Stop() is called.
 func (pf *PortForward) run() error {
+	return fmt.Errorf("a fake port-foward error")
 	transport, upgrader, err := spdy.RoundTripperFor(pf.config)
 	if err != nil {
 		return err
```

I am reasonably sure it's because we are missing a `return` statement in this block of code in `cli/cmd/metrics.go`:
```
	if err = portforward.Init(); err != nil {
[...]

PR #4007, opened 2020-02-03
cli: handle panic caused by `linkerd metrics` port-forward failure

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/3978

- add func `recoverStop()` to recover from panic caused by `close()` in method `Stop()`

### DIFF
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -179,14 +179,6 @@ func (pf *PortForward) Init() error {
 			failure <- err
 			return
 		}
-
-		select {
-		case <-pf.GetStop():
-			// stopCh was closed, do nothing
-		default:
-			// pf.run() returned for some other reason, close stopCh
-			pf.Stop()
-		}
 	}()
 
 	// The `select` statement below depends on one of two outcomes from `pf.run()`:
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -177,6 +177,7 @@ func (pf *PortForward) Init() error {
 	go func() {
 		if err := pf.run(); err != nil {
 			failure <- err
+			return
 		}
 
 		select {
@@ -203,6 +204,7 @@ func (pf *PortForward) Init() error {
 }
 
 // Stop terminates the port-forward connection.
+// It is the caller's responsibility to call Stop even in case of errors
 func (pf *PortForward) Stop() {
 	close(pf.stopCh)
 }
--- a/cli/cmd/metrics.go
+++ b/cli/cmd/metrics.go
@@ -165,6 +165,7 @@ func getMetrics(
 	defer portforward.Stop()
 	if err = portforward.Init(); err != nil {
 		fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
+		return nil, err
 	}
 
 	metricsURL := portforward.URLFor("/metrics")


adleong:
I'm trying to understand the failure case here.  Is the issue that if `pf.run()` returns an err, we still execute the next switch statement and call `pf.Stop()` which calls close on a channel that hasn't been created?

Could we simply return from the goroutine immediately after writing the err to the failure channel?  That way we will never call `pf.Stop()` at all and don't have to try to recover the panic.

mayankshah1607:
@adleong @ihcsim seems like that was causing the error. Could you take a look now?
Also, is it a good idea to let `recoverStop()` function be there?

adleong:
Ok I looked into this in more depth.  I suggest reading the comments on https://github.com/linkerd/linkerd2/pull/2976 since there's some good discussion there about the semantics of these methods.

My takeaway is that we need to be more consistent about who's responsibility it is to close the stop channel in case of an error.  `Init()` will call `Stop()` if there is an error running the port-forward but the `getMetrics` function also calls `Stop()` (via `defer`) if there is an error.  At other invocations such as in dashboard.go and controller/api/public/client.go we don't bother closing the channel at all in case of errors.

I think the most consistent thing would be to make it clear (in the godoc) that it is the callers responsibility to call `Stop()`, even in the case of errors.  This means that we would remove the call to `Stop()` from inside of `Init()` and potentially add a call to `Stop` in the error handling of dashboard.go and controller/api/public/client.go; although properly closing the channel in these cases probably isn't super important because the process will simply exit.  See https://github.com/linkerd/linkerd2/pull/2976/files#r297399887 for a description of why this can leak goroutines, but probably doesn't matter.

ihcsim:
In general, channels closed by the sender of the data, in this case, the callers of `Init()`. I recommend removing the `recoverStop()` function, to avoid hiding unexpected panics, and if necessary, create a separate issue to revisit the channel logic in the `Init()` constructor. Otherwise, making sure we `return` in the `if err != nil` LGTM.

mayankshah1607:
@ihcsim @adleong I've made the changes and removed the `recoverStop` method as per the above explanations. For now, returning from the routine at `Init()` if `err != nil` seems to gracefully handle the panic caused due to `Stop()`.


> This means that we would remove the call to Stop() from inside of Init() and potentially add a call to Stop in the error handling of dashboard.go and controller/api/public/client.go; although properly closing the channel in these cases probably isn't super important because the process will simply exit.

@adleong Should we open another issue for this? Seems to be a bit out of the scope of this issue; as we're handling the panics caused around `getMetrics()`

pkg/k8s/portforward.go
```diff
@@ -177,6 +177,7 @@ func (pf *PortForward) Init() error {
 	go func() {
 		if err := pf.run(); err != nil {
 			failure <- err
+			return
 		}
 
 		select {
```
adleong: I think we should remove this entire select block.  Regardless of how `pf.run()` returned, we want to do nothing since it is always the caller's responsibility to call `pf.Stop()`.  And then, since we're removing the select block, we can remove the early return too.

alpeb: :point_up: that makes a lot of sense to me! It is critical indeed to clarify the responsibility here.

mayankshah1607:
@adleong @alpeb done! but looks like the integration tests are failing

pkg/k8s/portforward.go
```diff
@@ -179,14 +179,6 @@ func (pf *PortForward) Init() error {
 			failure <- err
 			return
```
alpeb: Ah you missed removing this `return` :eye:

PR #4011, opened 2020-02-04
Scroll functionality for linkerd top deploy/linkerd-web

### LINKED ISSUE #2558: The table obtained from linkerd top deploy/linkerd-web is not wrapping in terminal properly

kohsheen1234:
Table obtained from linkerd top deploy/linkerd-web is not scrollable.

Added scroll functionality for the table.

Fixes #2558

### Testing
```sh
# Build 
bin/build-cli-bin

# Run
bin/linkerd -n linkerd top deploy/linkerd-web
```

### DIFF
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -656,21 +656,10 @@ func (t *topTable) renderHeaders(scrollpos int) {
 
 func (t *topTable) tableWidthCalc() int {
 	tablewidth := 0
-	for i, col := range t.columns {
-		t.columns[i].width = runewidth.StringWidth(col.header)
-		for _, row := range t.rows {
-			cellWidth := runewidth.StringWidth(col.value(row))
-			if cellWidth > t.columns[i].width {
-				t.columns[i].width = cellWidth
-			}
-		}
-		if col.flexible {
-			tablewidth = tablewidth + columnSpacing
-		}
-		tablewidth = tablewidth + t.columns[i].width
+	for i := range t.columns {
+		tablewidth = tablewidth + t.columns[i].width + columnSpacing
 	}
-	return tablewidth
-
+	return tablewidth - columnSpacing
 }
 
 func (t *topTable) adjustColumnWidths() {
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -254,7 +254,6 @@ const (
 	headerHeight  = 4
 	columnSpacing = 2
 	xOffset       = 5
-	columnGap     = 10
 )
 
 func newTopOptions() *topOptions {
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -253,8 +253,8 @@ func newTopTable() *topTable {
 const (
 	headerHeight  = 4
 	columnSpacing = 2
-	xOffset = 5
-	columnGap = 10
+	xOffset       = 5
+	columnGap     = 10
 )
 
 func newTopOptions() *topOptions {
@@ -493,22 +493,21 @@ func pollInput(done chan<- struct{}, horizontalScroll chan int) {
 				close(done)
 				return
 			}
-			if ev.Ch == 'a' || ev.Key==termbox.KeyArrowRight {
-				horizontalScroll <- 1
-
+			if ev.Ch == 'a' || ev.Key == termbox.KeyArrowRight {
+				horizontalScroll <- xOffset
 			}
-			if ev.Ch == 'd' || ev.Key==termbox.KeyArrowLeft  {
-				horizontalScroll <- 2
+			if ev.Ch == 'd' || ev.Key == termbox.KeyArrowLeft {
+				horizontalScroll <- -xOffset
 			}
 		}
 	}
 }
 
-func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struct{},  horizontalScroll chan int) {
-	scrollpos:=0
+func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struct{}, horizontalScroll chan int) {
+	scrollpos := 0
 	ticker := time.NewTicker(100 * time.Millisecond)
 	width, _ := termbox.Size()
-	tablewidth :=table.adjustColumnWidths()
+	tablewidth := table.tableWidthCalc()
 
 	for {
 		select {
@@ -519,23 +518,22 @@ func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struc
 		case <-ticker.C:
 			termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
 			width, _ = termbox.Size()
-			tablewidth =table.adjustColumnWidths()
+			table.adjustColumnWidths()
+			tablewidth = table.tableWidthCalc()
 			table.renderHeaders(scrollpos)
 			table.renderBody(scrollpos)
 			termbox.Flush()
-		case direction:= <-horizontalScroll:
-			if (direction ==2 &&  scrollpos<0){
-					scrollpos = scrollpos + xOffset
+		case offset := <-horizontalScroll:
+			if (offset > 0 && scrollpos < 0) || (offset < 0 && scrollpos > (width-tablewidth)) {
+				scrollpos = scrollpos + offset
 			}
-			if (direction ==1 && scrollpos >  width-tablewidth){
-				scrollpos = scrollpos - xOffset
-			}
-				termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
-				width, _= termbox.Size()
-				tablewidth =table.adjustColumnWidths()
-				table.renderHeaders(scrollpos)
-				table.renderBody(scrollpos)
-				termbox.Flush()
+			termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
+			width, _ = termbox.Size()
+			table.adjustColumnWidths()
+			tablewidth = table.tableWidthCalc()
+			table.renderHeaders(scrollpos)
+			table.renderBody(scrollpos)
+			termbox.Flush()
 		}
 	}
 }
@@ -640,7 +638,7 @@ func stripPort(address string) string {
 	return strings.Split(address, ":")[0]
 }
 
-func (t *topTable) renderHeaders(scrollpos int){
+func (t *topTable) renderHeaders(scrollpos int) {
 	tbprint(0, 0, "(press q to quit)")
 	tbprint(0, 1, "(press a/LeftArrowKey to scroll left, d/RightArrowKey to scroll right)")
 	x := scrollpos
@@ -657,8 +655,26 @@ func (t *topTable) renderHeaders(scrollpos int){
 	}
 }
 
-func (t *topTable) adjustColumnWidths() int {
-	tablewidth :=0
+func (t *topTable) tableWidthCalc() int {
+	tablewidth := 0
+	for i, col := range t.columns {
+		t.columns[i].width = runewidth.StringWidth(col.header)
+		for _, row := range t.rows {
+			cellWidth := runewidth.StringWidth(col.value(row))
+			if cellWidth > t.columns[i].width {
+				t.columns[i].width = cellWidth
+			}
+		}
+		if col.flexible {
+			tablewidth = tablewidth + columnSpacing
+		}
+		tablewidth = tablewidth + t.columns[i].width
+	}
+	return tablewidth
+
+}
+
+func (t *topTable) adjustColumnWidths() {
 	for i, col := range t.columns {
 		if !col.flexible {
 			continue
@@ -670,9 +686,7 @@ func (t *topTable) adjustColumnWidths() int {
 				t.columns[i].width = cellWidth
 			}
 		}
-		tablewidth = tablewidth +  t.columns[i].width + columnGap
 	}
-	return tablewidth + columnGap + xOffset
 }
 
 func (t *topTable) renderBody(scrollpos int) {
--- a/cli/cmd/top.go
+++ b/cli/cmd/top.go
@@ -251,8 +251,10 @@ func newTopTable() *topTable {
 }
 
 const (
-	headerHeight  = 3
+	headerHeight  = 4
 	columnSpacing = 2
+	xOffset = 5
+	columnGap = 10
 )
 
 func newTopOptions() *topOptions {
@@ -407,17 +409,17 @@ func getTrafficByResourceFromAPI(k8sAPI *k8s.KubernetesAPI, req *pb.TapByResourc
 	//       processEvents() && renderTable()
 	closing := make(chan struct{}, 1)
 	done := make(chan struct{})
+	horizontalScroll := make(chan int)
 
-	go pollInput(closing)
+	go pollInput(done, horizontalScroll)
 	go recvEvents(reader, eventCh, closing)
 	go processEvents(eventCh, requestCh, done)
 
 	go func() {
 		<-closing
-		close(done)
 	}()
 
-	renderTable(table, requestCh, done)
+	renderTable(table, requestCh, done, horizontalScroll)
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


cli/cmd/top.go
```diff
@@ -1,284 +1,287 @@
 package cmd
 
 import (
-	"bufio"
```
olix0r: Thanks for submitting this improvement, @kohsheen1234 ! Before we get deeper into the review, it appears that all of the whitespace in this file (cli/cmd/top.go) was changed, so it's difficult to see what logic changed, if any. As far as I can tell, there's probably no need for this file to change in this PR at all? Can you revert the unnecessary changes to this file?  Thanks!

kohsheen1234: Sure @olix0r . Sorry for the inconvenience , those filed are changed because of my previous pull request which isn't merged yet. I will surely make the changes as required as soon as possible :)

kohsheen1234: Hey @olix0r , For reviewing the changes you can check out this link. 

https://github.com/linkerd/linkerd2/pull/4013/files#diff-a15e4fe04e9a7d4097aed3e6a41db024L658

It contains all the changes I have made for this issue. Can you can pull my latest changes and test it locally, then I will create a formal pull request again :)

olix0r: Thanks! much clearer now ;)

We're working on a release at the moment, but someone will get to reviewing this soon--hopefully in the next few days.

Thanks again for contributing to Linkerd!

cli/cmd/top.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	// "os/exec"
```
adleong: I think we can delete these lines

cli/cmd/top.go
```diff
@@ -407,8 +410,14 @@ func getTrafficByResourceFromAPI(k8sAPI *k8s.KubernetesAPI, req *pb.TapByResourc
 	//       processEvents() && renderTable()
 	closing := make(chan struct{}, 1)
 	done := make(chan struct{})
+	scrollLeft := make(chan int)
```
adleong: Since we never care about the actual data in these channels, they can be `chan struct{}` instead of `chan int`.  Just like the ones above.

cli/cmd/top.go
```diff
@@ -407,8 +410,14 @@ func getTrafficByResourceFromAPI(k8sAPI *k8s.KubernetesAPI, req *pb.TapByResourc
...
+	scrollRight := make(chan int)
+	// width := 0
```
adleong: I think we can delete these lines

cli/cmd/top.go
adleong: Alternatively, we could just use a single channel for `horizontalScroll` for both left and right events, putting in the x-offset for each event.

cli/cmd/top.go
```diff
@@ -483,33 +492,70 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
-func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struct{}) {
-	ticker := time.NewTicker(100 * time.Millisecond)
-
+func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struct{}, scrollLeft <-chan int, scrollRight <-chan int) {
+	ticker := time.NewTicker(50 * time.Millisecond)
+	tablewidth:=0
+	scrollXpos := 0
+	width := 0
+	height := 0
+	width, height = termbox.Size()
```
adleong: I think if you use `:=` here, you can delete the previous two lines.

cli/cmd/top.go
```diff
@@ -483,33 +492,70 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
 		case req := <-requestCh:
 			table.insert(req)
 		case <-ticker.C:
-			termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
-			table.adjustColumnWidths()
-			table.renderHeaders()
-			table.renderBody()
-			termbox.Flush()
+			select {
+			case <-scrollLeft:
```
adleong: I'd expect us to be reading off the scroll channels in the outer select rather than here.  That way we can accept scroll events and update the xPos at any time, instead of waiting for a render tick. As this is written, you can only scroll once per render frame, which is probably fine in practice since I don't think anyone is trying to scroll more than once every 50ms.  

But still, I think this would be cleaner if the code to update the scrollXpos was in the outer select and this timer select was solely dedicated to rendering.  This helps remove some of the code duplication here.

cli/cmd/top.go
```diff
@@ -483,33 +492,70 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
+			}
 		}
 	}
 }
 
-func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struct{}) {
-	ticker := time.NewTicker(100 * time.Millisecond)
-
+func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struct{}, scrollLeft <-chan int, scrollRight <-chan int) {
+	ticker := time.NewTicker(50 * time.Millisecond)
```
adleong: Is this ticker change so that scrolling is responsive?  I think if we make the change I suggest below, this may not be necessary.

cli/cmd/top.go
```diff
@@ -483,33 +492,70 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
 			table.insert(req)
 		case <-ticker.C:
-			termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
-			table.adjustColumnWidths()
-			table.renderHeaders()
-			table.renderBody()
-			termbox.Flush()
+			select {
+			case <-scrollLeft:
+				if scrollXpos<=0 {
```
adleong: Since this is a `<=`, this allows scrolling left once, even when xPos is 0.

cli/cmd/top.go
```diff
@@ -483,33 +492,70 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
+					scrollXpos = scrollXpos + 5
```
adleong: Would be nice to move this scroll increment value (5) into a constant.

cli/cmd/top.go
```diff
@@ -483,33 +492,70 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
+					termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
+					width, height = termbox.Size()
+					tablewidth =table.adjustColumnWidths()
+					table.renderHeaders(scrollXpos)
+					table.renderBody(scrollXpos)
+					termbox.Flush()
+				}
+			case <-scrollRight:
+				height = height +0
```
adleong: I think this can be removed.

cli/cmd/top.go
adleong: I think height is unused, so we can do something like `width, _ := termbox.Size()`

cli/cmd/top.go
```diff
@@ -614,9 +660,11 @@ func stripPort(address string) string {
 	return strings.Split(address, ":")[0]
 }
 
-func (t *topTable) renderHeaders() {
+func (t *topTable) renderHeaders(scrollXpos int) {
 	tbprint(0, 0, "(press q to quit)")
-	x := 0
+	tbprint(0, 1, "(press a/LeftArrowKey to scroll left, d/RightArrowKey to scroll right)")
```
adleong: I think you need to increase the `headerHeight` constant by 1 to accommodate for this keep the blank line separating the header from the column headers.

cli/cmd/top.go
```diff
@@ -642,16 +691,18 @@ func (t *topTable) adjustColumnWidths() {
 				t.columns[i].width = cellWidth
 			}
 		}
+
+		tablewidth = tablewidth +  t.columns[i].width + 10
```
adleong: Where do the `+10` and `+15` values come from? Should these be based on the `columnSpacing` somehow?

kohsheen1234:
Thank you for the review @adleong 

I will surely fix all the changes you have suggested to clean up the code. I do understand that are are unrelated javascript changes as well, these are due to my previous pull request which hasn't been merged yet. I will consider your advice and try to fix the issue.

Also, I guess you are right about flipping the left/right scrolling behaviour. I will change that as well. :)

cli/cmd/top.go
kohsheen1234: @adleong This is my first time working with Go. I am not sure how to set an off-set for a channel. Could you please guide me on the same. Sorry for the inconvenience.

cli/cmd/top.go
kohsheen1234: surely :)

cli/cmd/top.go
kohsheen1234: I had done it to make it more responsive, but I feel after all the changes you've suggested the ticker will not be required to change. 👍

cli/cmd/top.go
```diff
@@ -417,7 +426,7 @@ func getTrafficByResourceFromAPI(k8sAPI *k8s.KubernetesAPI, req *pb.TapByResourc
 		close(done)
```
Pothulapati: The same channel `done` is being closed twice, here and in `pollInput`, hence generating the error
```
panic: close of closed channel

goroutine 29 [running]:
github.com/linkerd/linkerd2/cli/cmd.getTrafficByResourceFromAPI.func1(0xc00004c2a0, 0xc00004c3c0)
        /home/tarun/work/linkerd2/cli/cmd/top.go:426 +0x50
created by github.com/linkerd/linkerd2/cli/cmd.getTrafficByResourceFromAPI
        /home/tarun/work/linkerd2/cli/cmd/top.go:424 +0x2dc
```

kohsheen1234: @Pothulapati Thank you for recommending changes. I have fixed this :)

cli/cmd/top.go
kohsheen1234: Done

cli/cmd/top.go
kohsheen1234: Done :)

kohsheen1234:
@adleong I have made the changes as requested. I tried running `bin/fmt` and `bin/lint` to clean up formatting and lint issues, but it shows the same error. I do not understand why it shows the same. Could you guide me on the same. Also I have done rebase, so now it does not show any unnecessary changed files. :+1:

cli/cmd/top.go
kohsheen1234: I tried to use `columnSpacing` initially, but somehow it wasn't getting the right width of the table. By printing out `columnSpacing`, I realised it's maximum value is only 2, same case with the variable `padding`. Finally by Hit and Trial I figured out that `10` works perfectly for our scenario. Hence I used the same. :)

cli/cmd/top.go
kohsheen1234: @adleong Now I have created only one channel called `horizontalScroll`, which is of type chan int. I have also created the variable` xOffset` as mentioned. :)

cli/cmd/top.go
kohsheen1234: Done

cli/cmd/top.go
adleong: No problem.  I just meant using a `chan int` which holds the amount to change the x-position by.  So whenever a left key press is detected, we would put `-5` into the channel and whenever a right key press is detected, we would put `5` into the channel (or whatever the scroll amount is).  This lets us use 1 channel instead of 2.

cli/cmd/top.go
```diff
@@ -642,16 +670,18 @@ func (t *topTable) adjustColumnWidths() {
 				t.columns[i].width = cellWidth
 			}
 		}
+		tablewidth = tablewidth +  t.columns[i].width + columnGap
```
adleong: I think your math is thrown off here because of the `continue` on line 663.  This does not take into account the width of non-flexible columns.  I suggest changing it so that instead of using continue, we just put the column resizing code inside the `if col.flexible` block.

Alternatively, we can revert the changes to this function and create a new dedicated function for calculating table width.  Either way, we should be able to get rid the `columnGap` constant.

kohsheen1234: I have made changes considering all your inputs :)

cli/cmd/top.go
```diff
@@ -506,10 +518,24 @@ func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struc
...
-			table.adjustColumnWidths()
-			table.renderHeaders()
-			table.renderBody()
+			width, _ = termbox.Size()
+			tablewidth =table.adjustColumnWidths()
+			table.renderHeaders(scrollpos)
+			table.renderBody(scrollpos)
 			termbox.Flush()
+		case direction:= <-horizontalScroll:
+			if (direction ==2 &&  scrollpos<0){
```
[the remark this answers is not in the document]
kohsheen1234: @adleong Thank you for the recommendation. I have incorporated the changes. :)

kohsheen1234:
> Thanks for making those changes. I have a few more small comments.
> 
> If you run `bin/fmt`, it will output a diff of formatting changes that are needed. You'll need to make the changes indicated in that diff.

@adleong I have made all the necessary changes. Can you please review my latest changes. I have also used `bin/fmt`to clear the formatting issues. Thank you very much for the guidance :+1:

cli/cmd/top.go
```diff
@@ -630,6 +654,25 @@ func (t *topTable) renderHeaders() {
 	}
 }
 
+func (t *topTable) tableWidthCalc() int {
+	tablewidth := 0
+	for i, col := range t.columns {
+		t.columns[i].width = runewidth.StringWidth(col.header)
```
adleong: We always call `tableWidthCalc` after calling `adjustColumnWidths` so I don't think we need to adjust the column widths here and can assume that the column widths are already correct.

cli/cmd/top.go
```diff
@@ -630,6 +654,25 @@ func (t *topTable) renderHeaders() {
...
+		for _, row := range t.rows {
+			cellWidth := runewidth.StringWidth(col.value(row))
+			if cellWidth > t.columns[i].width {
+				t.columns[i].width = cellWidth
+			}
+		}
+		if col.flexible {
```
adleong: why do we only add the column spacing when the column is flexible?

I'm comparing this to the code in `renderBody` which increments `x` by both the column width and the column spacing for each column.  Wouldn't that be the right way to calculate the total table width?

kohsheen1234: Sorry. My mistake. I have corrected it.

cli/cmd/top.go
kohsheen1234: Done :)

kohsheen1234:
> Nice cleanup! This is almost ready to merge. It looks like the latest change switched the scroll direction again so I think you want to switch it back. I also had a small question about the way the table width is calculated.

@adleong Thanks a lot. I have update the code base with the minor changes. Please have a look. :)

cli/cmd/top.go
```diff
@@ -483,20 +484,29 @@ func processEvents(eventCh <-chan pb.TapEvent, requestCh chan<- topRequest, done
...
+func pollInput(done chan<- struct{}, horizontalScroll chan int) {
 	for {
 		switch ev := termbox.PollEvent(); ev.Type {
 		case termbox.EventKey:
 			if ev.Ch == 'q' || ev.Key == termbox.KeyCtrlC {
-				closing <- struct{}{}
+				close(done)
 				return
 			}
+			if ev.Ch == 'a' || ev.Key == termbox.KeyArrowRight {
```
Pothulapati: Looks like the `a` key and right don't match here, as they are opposite in placement

cli/cmd/top.go
```diff
@@ -506,9 +516,22 @@ func renderTable(table *topTable, requestCh <-chan topRequest, done <-chan struc
...
+			tablewidth = table.tableWidthCalc()
+			table.renderHeaders(scrollpos)
+			table.renderBody(scrollpos)
+			termbox.Flush()
+		case offset := <-horizontalScroll:
+			if (offset > 0 && scrollpos < 0) || (offset < 0 && scrollpos > (width-tablewidth)) {
+				scrollpos = scrollpos + offset
+			}
+			termbox.Clear(termbox.ColorDefault, termbox.ColorDefault)
+			width, _ = termbox.Size()
```
Pothulapati: I could be wrong here but removing this line, did not break anything. I don't see why the scroll event should re-calculate the table width? WDYT?

The ticker event calculating makes sense, as new data is added, the width might change.

cli/cmd/top.go
adleong: I just tested and pressing 'a' or Right arrow scrolls the view port to the left.  I think it should be that 'a' or Right should move the view port to the right (i.e. move the text to the left)

cli/cmd/top.go
kohsheen1234: You're right, All the lines written in case `ticker.c` need not be repeated again in the case `horizontalScroll`. I have removed the same, it did not break any functionality.

cli/cmd/top.go
kohsheen1234: I have corrected this now.

kohsheen1234:
@adleong @Pothulapati Can you check now. I have made the changes recommended.

PR #4012, opened 2020-02-04
edge-20.2.1

adleong:
This edge release is a release candidate for `stable-2.7` and fixes an issue
where the proxy could consume inappropriate amounts of memory.

* Proxy
  * Fixed a bug in the proxy's logging subsystem that could cause the proxy to
    consume memory until the process is OOMKilled, especially when the proxy was
    configured to log diagnostic information
  * Fixed properly emitting `grpc-status` headers when signaling proxy errors to
    gRPC clients
* Internal
  * Updated to Rust 1.40
  * Updated certain proxy dependencies to address RUSTSEC-2019-0033,
    RUSTSEC-2019-0034, and RUSTSEC-2020-02

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,12 +5,10 @@ where the proxy could consume inappropriate amounts of memory.
 
 * Proxy
   * Fixed a bug in the proxy's logging subsystem that could cause the proxy to
-    consume memory until the process is OOMKilled, especially when the proxy was
+    consume memory until the process is OOM killed, especially when the proxy was
     configured to log diagnostic information
   * Fixed properly emitting `grpc-status` headers when signaling proxy errors to
     gRPC clients
-* Internal
-  * Updated to Rust 1.40
   * Updated certain proxy dependencies to address RUSTSEC-2019-0033,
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,19 @@
...
+where the proxy could consume inappropriate amounts of memory.
+
+* Proxy
+  * Fixed a bug in the proxy's logging subsystem that could cause the proxy to
+    consume memory until the process is OOMKilled, especially when the proxy was
+    configured to log diagnostic information
+  * Fixed properly emitting `grpc-status` headers when signaling proxy errors to
+    gRPC clients
+* Internal
+  * Updated to Rust 1.40
```
hawkw: I don't think this is a user-facing change, so I probably wouldn't put it in the changelog?

olix0r: I'd omit the Rust upgrade here, and leave it all in the proxy section...

adleong: the gRPC behavior? it's a behavior change so I think it's "user" facing

PR #4032, opened 2020-02-10
Improve cni resources labels

zaharidichev:
Now that the CNI is in a different namespace, I it makes sense to use different kind of labels to identify the resources created by the chart/cli.

### DIFF
--- a/bin/test-cleanup
+++ b/bin/test-cleanup
@@ -14,6 +14,12 @@ if [ -z "${namespaces_helm=$(kubectl --context=$k8s_context get ns -oname -l lin
   echo 'no helm namespaces found' >&2
 fi
 
+echo "cleaning up CNI namespaces in k8s-context [${k8s_context}]"
+
+if [ -z "${namespaces_cni=$(kubectl --context=$k8s_context get ns -oname -l linkerd.io/cni-resource)}" ]; then
+  echo 'no CNI namespaces found' >&2
+fi
+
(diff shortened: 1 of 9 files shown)


zaharidichev:
@ihcsim We need it in the healthchecks because we want to select the linkerd resources without the CNI ones. Does that justify having it there ?

zaharidichev:
@ihcsim if the common label is `ControllerNamespaceLabel` that is a bit misleading, isnt it? Because the CNI might be installed before (and usually is) there is the controller namespace present at all (now that we updated it to be in a seprate namespace), so not sure this label makes much sense for the CNI resources.

ihcsim:
@zaharidichev Thanks for the explanation. It does look like these two labels are usually used together in healthcheck, except for the one [data plane case](https://github.com/linkerd/linkerd2/blob/3ba66f6f9df1943fc3ef1a19d830b4b467eb8601/pkg/healthcheck/healthcheck.go#L1629). NBD - just wanna confirm my understanding.

I forgot that we moved the CNI to a different namespace :thinking:. Personally, I am ok with using the `linkerd.io/control-plane-ns` label in all the CNI resources, even though they are installed before the control plane. IMO, being able to use one label to identify all Linkerd resources during runtime is more important than having a label referring to a not-yet-installed namespace during installation. This is super low priority; I might just create an issue and tag it as a "good first issue".

PR #4033, opened 2020-02-10
Run all PRs on GH Actions VMs

kleimkuhler:
## Motivation

Currently all pushes to master branch, tags, and Linkerd org member PRs run
the `kind_integration_host` job on the same Packet host.

The means that parallel jobs spin up KinD clusters with a unique name and
sandbox the tests so that they do not clash.

This is problematic for a few reasons:
* There is a limit on the number of jobs we can run in parallel due to
  resource constraints.
* Workflow cancellation and re-runs conflict when the cancelled run deletes
  it's namespaces and the running one expects them to be present.
* There has been an observed flakiness with running multiple KinD clusters
  resulting in inconsistent timeouts and docker errors.

## Solution

This change moves all KinD integration testing to GH Actions VMs. This is
currently what forked repository workflows do.

There is no longer a `docker_pull` job as it's responsibilities has been moved
into one of the `kind_integration_tests` steps.

The renamed `kind_integration_tests` job is responsible for **all** PR
workflows and has steps specific to forked and non-forked repositories.

### Non-forked repository PRs

The Packet host is still used for building docker images as leveraging docker
layer caching is still valuable--a build can be as fast as 30 seconds compared
to ~12 minutes.

Loading the docker images into the KinD cluster on the GH Action VM is done by
saving the Packet host docker images as image archives, and loading those
directly into the local KinD cluster.

### Forked repository PRs

`docker_build` has been sped up slightly by sending `docker save` processes to
the background.

Docker layer caching cannot be leveraged since there are no SSH secrets
available, so the `artifact-upload`/`artifact-download` actions introduced in
#TODO are still used.

### Cleanup

This PR also includes some general cleanup such as:
* Some job names have been renamed to better reflect their purpose or match
  the current naming pattern.
* Environment variables are set earlier in jobs as a separate step if it is
  currently exported multiple times.
* Indentation was really bothering me because it switches back and forth
  throughout the workflow file, so lists are now indented.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -151,8 +151,8 @@ jobs:
         PROXY_INIT_IMAGE_NAME: gcr.io/linkerd-io/proxy-init:v1.3.1
         PROMETHEUS_IMAGE_NAME: prom/prometheus:v2.15.2
       run: |
-        bin/docker pull "gcr.io/linkerd-io/proxy-init:$PROXY_INIT_IMAGE_NAME"
-        bin/docker pull "prom/prometheus:$PROMETHEUS_IMAGE_NAME"
+        bin/docker pull $PROXY_INIT_IMAGE_NAME
+        bin/docker pull $PROMETHEUS_IMAGE_NAME
 
   #
   # Docker build and kind integration tests run for:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -166,12 +166,18 @@ jobs:
         mkdir -p $ARCHIVES
 
         for image in proxy controller web cni-plugin debug cli-bin grafana; do
-          docker save "$DOCKER_REGISTRY/$image:$TAG" > $ARCHIVES/$image.tar || tee save_fail &
+          docker save "$DOCKER_REGISTRY/$image:$TAG" > $ARCHIVES/$image.tar || echo "failed 'docker save'" && tee save_fail &
         done
+
+        ls
 
         # Wait for `docker save` background processes to complete. Exit early
         # if any job failed.
         wait < <(jobs -p)
+
+        echo "finished waiting.."
+        ls
+
         test -f save_fail && exit 1
     # `with.path` values do not support environment variables yet, so an
     # absolute path is used here.
@@ -264,7 +270,6 @@ jobs:
         # used to load the output into the KinD cluster. Set `DOCKER_HOST` for
         # a single command, and `docker save` the container from the Packet
         # host.
-        i=0
         for image in proxy controller web cni-plugin debug grafana; do
           kind load image-archive <(DOCKER_HOST=ssh://linkerd-docker docker save "$DOCKER_REGISTRY/$image:$TAG") || tee load_fail &
         done
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -4,9 +4,9 @@ on:
   pull_request: {}
   push:
     branches:
-      - master
+    - master
     tags:
-      - "*"
+    - "*"
 
 # Jobs by event type and dependencies:
 
@@ -257,7 +257,7 @@ jobs:
       if: github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork
       env:
         PROXY_INIT_IMAGE_NAME: gcr.io/linkerd-io/proxy-init:v1.3.1
-        PROMETHEUS_IMAGE_NAME: prom/prometheus:v2.11.1
+        PROMETHEUS_IMAGE_NAME: prom/prometheus:v2.15.2
       run: |
         # For each container, load the image archive into the KinD cluster.
         #
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -4,36 +4,28 @@ on:
   pull_request: {}
   push:
     branches:
-    - master
+      - master
     tags:
-    - "*"
+      - "*"
 
 # Jobs by event type and dependencies:
 
 # Unit tests for every master/tag push and PR
 #
-# validate_go_deps
+# go_dependencies
 # go_unit_tests
 # go_lint
 # js_unit_tests
 
-# master/tag push and linkerd org PR: Docker build and integration tests
-#
-# docker_pull
-# docker_build
-# kind_setup
-#   -> kind_integration_host
-#     -> kind_cleanup
-
-# Forked repository PR: Docker build and integration tests
+# All master/tag pushes and PRs
 #
 # docker_build
-#   -> kind_integration_github
+# -> kind_integration_tests
 
-# Docker deploy and cloud integration tests for every master/tag push
+# Docker push and cloud integration tests for every master/tag push
 #
-#       -> docker_deploy
-#         -> cloud_integration
+#     -> docker_push
+#        -> cloud_integration_tests
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -44,96 +36,91 @@ jobs:
...
     container:
       image: golang:1.13.4
     steps:
-    - name: Checkout code
-      uses: actions/checkout@v2
-    - name: Go lint
-      env:
-        GITCOOKIE_SH: ${{ secrets.GITCOOKIE_SH }}
-        # prevent OOM
-        GOGC: 20
```
alpeb: No longer needed?

.github/workflows/workflow.yml
```diff
@@ -142,363 +129,265 @@ jobs:
...
+        if: matrix.integration_test == 'custom_domain'
+        uses: engineerd/setup-kind@v0.3.0
+        with:
+          config: test/testdata/custom_cluster_domain_config.yaml
+          version: "v0.6.1"
+      - name: Load image archives into the local KinD cluster
+        if: github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork
+        env:
+          PROXY_INIT_IMAGE_NAME: gcr.io/linkerd-io/proxy-init:v1.3.1
+          PROMETHEUS_IMAGE_NAME: prom/prometheus:v2.11.1
```
alpeb: We've been on prom v2.15.2 for a while, not sure how has this been working without updating this tag...

.github/workflows/workflow.yml
kleimkuhler: I removed this on most runs of workflow tests on my fork for both the forked PRs testing as well as testing for this.

I have not run into an issue with it since, and I think it's just carried over from a time when it was required. I know it's strictly necessary to *remove*, but I have not been able to see a case where it's needed.

.github/workflows/workflow.yml
kleimkuhler: Hm yea I'm not sure. I can update it, but maybe as a separate PR? I'd like to keep a version bump out of this if possible.

kleimkuhler:
@alpeb When lists are the value of a dictionary I've tended to see them indented just like any other value would be. Also why I changed this is I've had to un-indent new lists in order to match what we currently have. Lastly pretty much any YAML documentation I look at [indents](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html) lists as well.

.github/workflows/workflow.yml
siggy: This image name is for seeding the kind cluster docker caches prior to Linkerd installation. It makes CI (and Linkerd's bootup) faster. It should be kept in-sync with Linkerd, but worst case it just slows things down.

alpeb:
@kleimkuhler it must be a kubernetes thing then. Every yaml coming out of `kubectl get -oyaml` doesn't use indentation for lists, nor have we in our test golden files or Helm templates. I've also put together [this snippet](https://play.golang.org/p/2qnSy51AiMG) that uses  https://github.com/kubernetes-sigs/yaml, the same lib that k8s and ourselves rely on.

kleimkuhler:
@alpeb Okay interesting yea I can remove the indentation then

.github/workflows/workflow.yml
```diff
@@ -183,256 +153,175 @@ jobs:
...
+
+        for image in proxy controller web cni-plugin debug cli-bin grafana; do
+          docker save "$DOCKER_REGISTRY/$image:$TAG" > $ARCHIVES/$image.tar &
+          pids[${i}]=$!
+          i=$((i+1))
+        done
+
+        # Wait for `docker save` background processes to complete.
+        for pid in ${pids[*]}; do
+          wait $pid
```
alpeb: I think you can just do one single `wait` with no args and it'll wait for all the children jobs to finish. This also avoids having to track their pids.
BUT, we avoid the option to track individual failures for each job. If we hadn't spawned background jobs, the script would fail at the first failure (since this runs using `bash -e` according to the docs). Spawning background jobs disables that fail-fast behavior (according to my tests), so we'd have to track the exit status for each job and fail whenever one of them is > 0.
We could do that by iterating through the `jobs -f` output and inside the loop do `wait -$pid || exit 1` or something like that.

.github/workflows/workflow.yml
```diff
@@ -183,256 +153,175 @@ jobs:
...
-        ssh -4 -N -L $KIND_PORT:localhost:$KIND_PORT linkerd-docker &
-        sleep 2 # Wait for ssh tunnel to come up.
-        bin/kubectl version --short # Test connection to kind cluster.
-        (
-          . bin/_test-run.sh
-          init_test_run $HOME/.linkerd
-          ${{ matrix.integration_test }}_integration_tests
-        )
+        . bin/_tag.sh
+        echo ::set-env name=TAG::$(CI_FORCE_CLEAN=1 bin/root-tag)
```
alpeb: It's possible to declare env vars at the top level, that would get shared across all jobs. That might work to avoid repeating this in `docker_build`, `kind_integration_tests` and `docker_push`

alpeb: But might be better to try that _after_ the workflow.yml split off, so we don't have to add this back

kleimkuhler: Yep in the `workflow.yml` split off I'd like to make a similar step for other jobs that set environment multiple times.

I'm not sure this can be set at the top level though since it relies on having checked out the code already. Unless you're suggesting that a step (after `actions/checkout`) can set a workflow level environment variable?

alpeb: > Unless you're suggesting that a step (after `actions/checkout`) can set a workflow level environment variable?

Yea, I know you can declare a global env var, but I'm not sure if it's modifiable from within a job.

.github/workflows/workflow.yml
kleimkuhler: I think something like this will work:

```
for image in proxy controller web cni-plugin debug cli-bin grafana; do
  docker save "$DOCKER_REGISTRY/$image:$TAG" > $ARCHIVES/$image.tar || tee save_fail &
done

# Wait for `docker save` background processes to complete.
wait < <(jobs -p)
test -f save_fail && exit 1
```

1. Each background job will create a `save_fail` file if the `docker save ..` command fails.
2. All the background jobs are waited for
3. Test for the presence of `save_fail` fail and exit if it is found.

I can test this a little more, but good find here!

.github/workflows/workflow.yml
kleimkuhler: Yea that could be helpful to set early in a workflow. After the workflow files are split I can see if there is a more global `::set-env` option, but I don't see something like that right now.

kleimkuhler:
This renamed `Go Format` to `Go format` and `Validate go deps` to `Go dependencies`, so those two required checks will not actually run. The required checks will need to be updated again.

kleimkuhler:
I added back the `docker_pull` job that runs in addition to `docker_build`.

When I changed the version of proxy-init and prometheus images, the `docker save ..` command failed in `kind_integration_tests` because the new prometheus version was not pulled down on the packet host.

That job is still helpful to to run!

alpeb:
@kleimkuhler It appears `kind_integration_tests` are not running in forks because they now require `docker_pull` which only runs in master. Perhaps it would make sense to forgo `docker_pull` and instead put the `Ensure Packet has the correct proxy-init and prometheus versions` right before `kind_integration_tests`'s `Load cli-bin image into local docker images`?

alpeb:
... OTOH I'm not sure if pulling those images is really needed after all. The `docker save` commands will fail if the images are not there, which only happens right after our weekly docker cache cleanup, and they'll be automatically pulled and be available for the following builds. How about just logging the failure and not having the build step fail?

kleimkuhler:
@alpeb Yea I agree I changed the `docker save ..` commands to just output STDERR to STDOUT but not fail the script. It should be good to go again.

PR #4034, opened 2020-02-10
Ensure we get the correct type out of Informer Deletion events

zaharidichev:
It turns out that when we define our `DeleteFunc` handler it is possible in certain unfortunate situations to get a type different than what we were expecting (we want a *Service, we get `DeletedFinalStateUnknown`). If that happens we will probably fail with some cast exception. This Pr here ensures we consider these cases and cast/extract types in safer manner. 

 Wrt to the `DeletedFinalStateUnknown`, it  is placed into a DeltaFIFO in the case where an object was deleted but the watch deletion event was missed. In this case we don't know the final "resting" state of the object, so there's a chance the included `Obj` is stale. 


zaharidichev:
@alpeb. Are you asking whether we should simply have the logic of unpacking? I guess what I was trying to achieve here is to contain the checking in one method only. That allows me to ensure that if tomorrow someone on the k8s team decides to introduce yet another wrapper to express some state that I never though of, our code would not panic with an unsafe cast but would simply skip casting and log an error. Does that make sense

controller/api/destination/watcher/k8s.go
```diff
@@ -41,3 +44,17 @@ func invalidService(authority string) InvalidService {
 func (i ID) String() string {
 	return fmt.Sprintf("%s/%s", i.Namespace, i.Name)
 }
+
+// Used by the delete handlers to make sure we are getting the
+// correct type out of the DeltaFIFO
+func extractDeletedObject(obj interface{}, expectedType reflect.Type) (interface{}, error) {
+	switch object := obj.(type) {
+	case cache.DeletedFinalStateUnknown:
```
ihcsim: It seems to be a very common pattern in the K8s code to validate the type of `object.Obj` too. I'd recommend doing the same thing. In that case, it might be easier to just not use a helper function. See

* Replicaset : https://github.com/kubernetes/kubernetes/blob/0b2636a7e705bdf7c911e12d2e8846b9e6f88704/pkg/controller/replicaset/replica_set.go#L330-L342
* Endpoint: https://github.com/kubernetes/kubernetes/blob/0b2636a7e705bdf7c911e12d2e8846b9e6f88704/pkg/controller/util/endpoint/controller_utils.go#L219-L235
* Deployment: https://github.com/kubernetes/kubernetes/blob/0b2636a7e705bdf7c911e12d2e8846b9e6f88704/pkg/controller/deployment/deployment_controller.go#L339-L356
 * Statefulset: https://github.com/kubernetes/kubernetes/blob/0b2636a7e705bdf7c911e12d2e8846b9e6f88704/pkg/controller/statefulset/stateful_set.go#L243-L259

WDYT?

zaharidichev: Good point. Should have check how they do it. I am not a programming language expert but am I the only one finding that doing so much type checking in a typed programming language is a shame :(

ihcsim: Yeah, generally a function parameter or struct field of `interface` type is useless, and it's a smell. If only there the language supports generic data types...

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -194,18 +195,22 @@ func (ew *EndpointsWatcher) addService(obj interface{}) {
...
 func (ew *EndpointsWatcher) deleteService(obj interface{}) {
-	service := obj.(*corev1.Service)
-	if service.Namespace == kubeSystem {
-		return
-	}
-	id := ServiceID{
-		Namespace: service.Namespace,
-		Name:      service.Name,
-	}
+	if serviceObj, err := extractDeletedObject(obj, reflect.TypeOf(&corev1.Service{})); err == nil {
```
adleong: I see that you're trying to factor the common code for handling the final state unknown into a common function.  Unfortunately, this introduces unnecessary use of "reflect" because Go doesn't support generics.  I'd recommend following the pattern of the k8s libraries (https://github.com/kubernetes/kubernetes/blob/0b2636a7e705bdf7c911e12d2e8846b9e6f88704/pkg/controller/replicaset/replica_set.go#L330-L342) and check the type against the final state unknown tombstone in each delete handler.

Yes, this is a lot more repeated code.  It is the Go way.  😞

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -194,18 +195,22 @@ func (ew *EndpointsWatcher) addService(obj interface{}) {
...
 
-	sp, ok := ew.getServicePublisher(id)
-	if ok {
-		sp.deleteEndpoints()
+		sp, ok := ew.getServicePublisher(id)
+		if ok {
+			sp.deleteEndpoints()
+		}
+	} else {
+		ew.log.Error(err)
```
ihcsim: Can we use the `utilruntime.HandleError()` to handle this, as seen [here](https://github.com/kubernetes/kubernetes/blob/0b2636a7e705bdf7c911e12d2e8846b9e6f88704/pkg/controller/replicaset/replica_set.go#L334-L335)? For more info of the function, see [here](https://godoc.org/k8s.io/apimachinery/pkg/util/runtime#HandleError).

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: I see, but what is the problem with using the reflect package in this particular case ?

adleong: I try to avoid reflection whenever reasonable.  According to the [Go proverbs](https://go-proverbs.github.io/):

>Clear is better than clever.
>Reflection is never clear.

zaharidichev: I have to agree, reflection is hacky. Will inline the checks and remove the tests as well. Not sure I like the Go way :(

zaharidichev: Took another look at this and with the risk of starting an argument I dont want to be a part of, here are my top reasons to still opt out for the reflection here: 

1. We seem to be using reflection in various places of the code (of course if its so bad the fact that we have done doesn't mean we should be doing it again).
2. If we are to do that we will most likely loose the ability to have meaningful tests on that logic.
3. Its hard for me to explain why the k8s codebase uses reflection in quite some spots as well but opted for code repetition in these particular places.

adleong: Can you elaborate on point 2?  If we don't have a `extractDeletedObject` function then we cannot test it --- but we should still be able to test the individual delete handlers.

zaharidichev: Ok, sure we can test the individual handlers. Its just going to be a such a non DRY piece of code... Anyway, your call. I am fine with whatever approach

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: Correct me if I am wrong but the default behaviour of that is to just log. We are using a scoped logger here (withFields) I think and that would make it hard to diagnose where this problem actually occurs (if we are using utilruntime.HandleError()) instead? Or I guess we can add the information in the error message.

ihcsim: [`utilruntime.HandleError()`](https://github.com/kubernetes/apimachinery/blob/fb45917aae8b47ab6f7c71d97863d287ee03bf6a/pkg/util/runtime/runtime.go#L82-L110) logs the error and perform some backoff. Now that I look closer at the source, it's using `klog`, which is different from our `ew.log`. Let's stick with what you have.

controller/api/destination/watcher/endpoints_watcher.go
adleong: Writing Go code is like trying to give my kids a bath.  I'd prefer to stay DRY but it never seems to work out that way...

zaharidichev: Hard to disagree :)

zaharidichev: @adleong That's it. We are drenched now. Feel free to take a second look :)

PR #4035, opened 2020-02-10
Allow custom hosts set from Helm values for enforced-host 

### LINKED ISSUE #3961: Allow custom hosts set from Helm values for "enforced-host"
## Feature Request

In the https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/web.yaml#L70 I would like to add a custom host.
```
- -enforced-host=^(localhost|127\.0\.0\.1|{{ $hostFull }}|{{ $hostAbbrev }}|\[::1\])(:\d+)?$
```

I'm trying to access Linkerd Dashboard behind a proxy which does not rewrite the `Host` header.
As [per documentation](https://linkerd.io/2/tasks/exposing-dashboard/#tweaking-host-requirement) I need to change the validation regexp that the dashboard server uses. But this regexp is hardcoded in Helm values.

A new Helm value can be introduced to support adding an additional custom host or to completely replace the rule including an ability to completely disable the `Host` header check using an empty string as it said in the documentation.

Manually change the line in a locally cached version of the chart (which I do, but it's inconvenient to maintain custom code changes and may be inappropriate for other users).

They can add a custom host to the filter when necessary.

sannimichaelse:
Allow custom hosts set from Helm values for enforced-host 

Users want to be able to specify custom hosts instead of using the default host provided in the configuration files(web.yaml and values.yaml files)

Added the option for user to set custom hosts which will be used otherwise it resolves to the default host 

Fixes #3961

Signed-off-by: Sanni Michael Tomiwa <sannimichaelse@gmail.com>


charts/linkerd2/templates/web.yaml
```diff
@@ -67,7 +67,8 @@ spec:
         - -log-level={{.Values.controllerLogLevel}}
         {{- $hostFull := replace "." "\\." (printf "linkerd-web.%s.svc.%s" .Values.global.namespace .Values.global.clusterDomain) }}
         {{- $hostAbbrev := replace "." "\\." (printf "linkerd-web.%s.svc" .Values.global.namespace) }}
-        - -enforced-host=^(localhost|127\.0\.0\.1|{{ $hostFull }}|{{ $hostAbbrev }}|\[::1\])(:\d+)?$
+        {{- $default_host :=^(localhost|127.0.0.1| $hostFull | $hostAbbrev |[::1])(:\d+)?$ }}
+        - -enforced-host={{ .Values.global.enforcedHostRegexp | default $default_host }}
```
Pothulapati: I think the code here can be written as
```
        {{- if .Values.enforcedHostRegexp }}
        - -enforced-host= {{.Values.enforcedHostRegexp}}
        {{- else -}}
        {{- $hostFull := replace "." "\\." (printf "linkerd-web.%s.svc.%s" .Values.global.namespace .Values.global.clusterDomain) }}
        {{- $hostAbbrev := replace "." "\\." (printf "linkerd-web.%s.svc" .Values.global.namespace) }}
        - -enforced-host=^(localhost|127\.0\.0\.1|{{ $hostFull }}|{{ $hostAbbrev }}|\[::1\])(:\d+)?$
        {{- end}}
```
There should be a better way to do this using `default` but I kept getting syntax errors. because of `^`, etc symbols, enclosing it into a string using `printf` did not help :\ @alpeb

charts/linkerd2/values.yaml
```diff
@@ -9,6 +9,9 @@ global:
   
   # Control Plane Trace Configuration
   controlPlaneTracing: false
+  
+  # Add custom host here
+  enforcedHostRegexp: ""
```
Pothulapati: `global` is only used for fields that are needed across the control-plane and add-ons (that are being added). So, This field can be moved outside.

charts/linkerd2/templates/web.yaml
sannimichaelse: Thanks @Pothulapati , will update this and push again

charts/linkerd2/values.yaml
sannimichaelse: Thanks, Gotten

sannimichaelse:
I think one of my commit was not signed so the DCO check has not be passing, I used several resources to try to edit the message of a particular commit but no success yet. 

https://gist.github.com/kwk/d70f20d17b18c4f3296d

https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d

Any ideas on how to solve this @Pothulapati ?

Pothulapati:
Looks like your git tree is corrupted as it has commits folks from other folks. I think that is causing those issues

I think you can reset them all or start the commit tree for this branch afresh as the changes are very small Currently and force push.

alpeb:
> Also, @alpeb I think this flag should only be for `helm` users as it feels advanced. Should we also support the field as an option through the CLI?

Definitely a Helm-only thing IMO :+1:

sannimichaelse:
> Looks like your git tree is corrupted as it has commits folks from other folks. I think that is causing those issues
> 
> I think you can reset them all or start the commit tree for this branch afresh as the changes are very small Currently and force push.

Thanks for this, how do I reset them or start the commit tree from the current branch? 

I was already thinking of creating another PR with the reviewed changes to resolve the git tree issue but closing this since its a draft PR.

Pothulapati:
Please close this PR, and just start fresh with a new branch as the changes are just two lines currently.

PR #4039, opened 2020-02-11
Separate single Actions workflow into multiple workflows

kleimkuhler:
Depends on #4033

## Motivation

If any job fails in the current GH Actions workflow, a re-run on the same
commit SHA requires re-running *all* jobs--regardless if the job already
passed in the previous run.

This can be problematic when dealing with flakiness in the integration tests.

If a test fails due to flakiness in `cloud_integration_tests`, all the unit
tests, static checks, and `kind_integration_tests` are re-run which leads to
lots of waiting and dealing with the possibility of flakiness in earlier jobs.

With this change, individual workflows can now be re-run without triggering
all other jobs to complete again first.

## Solution

`workflow.yml` is now split into:
- `static_checks.yml`
- `unit_tests.yml`
- `kind_integration.yml`
- `cloud_integration.yml`

### Workflows

`statc_checks.yml` performs checks related to dependencies, linting, and
formatting.

`unit_tests.yml` performs the Go and JS unit tests.

`kind_integration.yml` builds the images (on Packet or the GH Action VM) and
runs the integration tests on a KinD cluster. This workflow continues to run
for **all** PRs and pushes to `master` and tags.

`cloud_integration.yml` builds the images only on Packet. This is because
forked repositories do not need to trigger this workflow. It then creates a
unique GKE cluster and runs the integration tests on the cluster.

### The actual flow of work..

A forked repository or non-forked repository opening a PR triggers:
- `static_checks`
- `unit_tests`
- `kind_integration_tests`

These workflows all run in parallel and are invidivually re-runnable.

A push to `master` or tags triggers:
- `static_checks`
- `unit_tests`
- `kind_integration_tests`
- `cloud_integration_tests`

These workflows also all run in parallel, including the `docker_build` step of
both integration test workflows. This has not conflicted in testing as it
takes place on the same Packet host and just utilizes docker layer caching
well.


kleimkuhler:
In #4033 `Go Format` was renamed to `Go format` and `Validate go deps` was renamed to `Go dependencies`, so these required checks will not actually run. Those checks will be updated once it merges.

alpeb:
Thinking about the upcoming automated release work, what do you think about moving `chart_deploy` into a separate workflow that would be triggered by tags with the shape `(edge-.*|stable-*)`? We'd remove the `tags: *` trigger from all the other workflows, and so when we tag the release (after master would have gone green in the other workflows) only this new workflow would get triggered, unlike the current situation where we have to wait for everything to run again.
This new workflow would also contain the upcoming automated release jobs.

alpeb:
... to be more precise about the current situation, after the changelog changes have merged, we have to  manually cancel the build, and then tag, and it's upon this last event that the final tests are run and the release (currently just the Helm artifact) is pushed. If that last run fails, that might require pushing a fix, and some of us are familiar with the untagging/tagging business that requires :unamused: 
What I'm thinking is that we leave that first merge CI run to continue, and if it's green, tag the release to trigger the release steps.
I'm getting a little ahead into the release automation work here, but I think laying out the workflows like I described could be a good starting point.

kleimkuhler:
> what do you think about moving chart_deploy into a separate workflow ..

So this would relax the requirement that the integration tests pass for the tag (whether it's edge or stable) since the workflow would run right away instead of waiting for `cloud_integration_tests` to complete like it does currently.

However, as you also mention the the first merge would just continue, and once it's green we *know* that tagging that commit as a release is okay.

I would be okay with that change and think it makes sense! A push to master in order to release an edge or stable shouldn't require cancelling one CI run in order to let the other (tagged) workflow run. For automating releases, it would be a lot easier if the workflow was what you are suggesting and `chart_deploy` was specific to those `(edge-.*|stable-*)` conditions.

.github/workflows/cloud_integration.yml
```diff
@@ -0,0 +1,134 @@
...
+    - name: Build docker images
+      env:
+        DOCKER_TRACE: 1
+      run: |
+        export PATH="`pwd`/bin:$PATH"
+        bin/docker-build
+  docker_push:
+    name: Docker push
+    runs-on: ubuntu-18.04
+    needs: [docker_build]
```
siggy: iiuc, this could push docker images with `edge`/`stable` release tags, even if unit or integration tests are failing. i don't think this push should happen until all tests have passed.

kleimkuhler:
> iiuc, this could push docker images with edge/stable release tags, even if unit or integration tests are failing. i don't think this push should happen until all tests have passed.

@siggy and I talked about this directly and decided this okay to keep around for this change. This behavior exists in the current workflow file, and this change just persists that behavior.

The possibility of tags being pushed and integration later failing should be addressed however as part of the future release automation work.

As @alpeb mentioned in this [comment](https://github.com/linkerd/linkerd2/pull/4039#issuecomment-584791082), `chart_deploy` and other "release only" jobs should be moved to a release workflow that is responsible for pushing tagged images, deploying helm charts, etc.

siggy:
To close the loop, I'm onboard moving forward with this change, provided we ensure all tests are passing before any docker images tagged as `edge` or `stable` are pushed to a public repository. This requirement could be enforced initially via careful documentation and deploy instructions, but in the long run this should be enforced programmatically.

Also agree it would be nice if `chart_deploy` and all other release-specific steps were separated out into their own workflow. This is a secondary concern compared to ensuring a release workflow only runs after all tests have passed.

.github/workflows/cloud_integration.yml
```diff
@@ -0,0 +1,134 @@
...
+      run: |
+        export PATH="`pwd`/bin:$PATH"
+        echo "$GITCOOKIE_SH" | bash
+        version="$($HOME/.linkerd version --client --short | tr -cd '[:alnum:]-')"
+        bin/test-run $HOME/.linkerd linkerd-$version
+    - name: CNI tests
+      run: |
+        export TAG="$($HOME/.linkerd version --client --short)"
+        go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
+  chart_deploy:
```
adleong: its a bit intuitive that this is here since it's not part of the cloud integration tests.  But I guess the reason is that we only want to do the chart deploy after the cloud integration tests are successful?

adleong: *counter-intuitive

kleimkuhler: It is a little counter-intuitive but it keeps this change from really changing much about actual release process. This will all be addressed in follow-up work for release automation.

Since workflows cannot be dependent on other the results of other workflow runs (this is just a limitation of GH Actions currently), `chart_deploy` needs to be in a workflow that runs integration tests that ensures the images are safe to deploy. It's why this change cannot just separate `chart_deploy` into it's workflow, that workflow would run at the same time as all the others and it may not be okay to deploy.

kleimkuhler:
@alpeb Yep I the `.. != 'pull_reuest'` only match for `push` events right now anyways because of the `on` condition in the workflows, so I just changed it to be more obvious as well as ensure any changes that take place in the future don't accidentally trigger jobs (for example, adding additional `on` events).

Issue #2734, opened 2019-04-22
cli: Support kubectl-style resource selectors


olix0r:
The various Linkerd subcommands that operate on kubernetes resources all work somewhat differently, and none of them support Kubernetes label selection.

Anywhere where the `linkerd` CLI operates on a list of resources (`tap`, `top`, `routes`, `stat`, `metrics`, etc), `linkerd` should accept resource selections exactly like `kubectl`.

Furthermore the `--linkerd-namespace` flag should be changed so that its short flag is `-L`, so that the `-l` flag can be used for label selection (like other kubernetes-related tools).

For example, kubectl supports the following:

```
:; kubectl get -n linkerd svc/linkerd-web deploy/linkerd-web
NAME                  TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)             AGE
service/linkerd-web   ClusterIP   10.0.238.66   <none>        8084/TCP,9994/TCP   11d

NAME                                DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deployment.extensions/linkerd-web   1         1         1            1           11d
```

```
:; kubectl -n linkerd get po -l linkerd.io/control-plane-component    
NAME                                   READY     STATUS    RESTARTS   AGE
linkerd-controller-58fb7787b-kxhw5     4/4       Running   0          11d
linkerd-grafana-659c48d67-h8lx6        2/2       Running   0          11d
linkerd-identity-84fd9b78b-g9h4c       2/2       Running   0          11d
linkerd-prometheus-644698f444-zg9fn    2/2       Running   0          11d
linkerd-sp-validator-c4965d44c-8qqmb   2/2       Running   0          4d
linkerd-web-7f46b9cfcb-6c6kv           2/2       Running   0          4d
```
```
:; kubectl -n linkerd get deploy linkerd-controller linkerd-identity
NAME                 DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
linkerd-controller   1         1         1            1           11d
NAME                 DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
linkerd-identity     1         1         1            1           11d
```

`linkerd` should accept resource lists similarly.

---

`linkerd stat` accepts multiple resources, like kubectl, but does not accept label selectors:
```
:; linkerd -n linkerd stat deploy linkerd-controller linkerd-identity
[...]

PR #4040, opened 2020-02-11
cli: Add label selector flag for `stat`

mayankshah1607:
First take on https://github.com/linkerd/linkerd2/issues/2734

#### Add `kubectl` like label selector for `linkerd stat`
As opposed to earlier, `linkerd stat` command now accepts a label selector via the `--selector` (or `-l` shorthand):
```bash
$ linkerd -n emojivoto stat deploy -l test.linkerd/test
```
The above command will return stats for all deployments with the label `test.linkerd/test` in the `emojivoto` namespace.

---
Follow up PRs shall be made for the following features (as a part of https://github.com/linkerd/linkerd2/issues/2734):
- Update `tap`, `top`, `routes` commands to accept resource lists much like `stat`
- Update `tap`, `top`, `routes` to accept label selectors

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -92,7 +92,7 @@ var RootCmd = &cobra.Command{
 }
 
 func init() {
-	RootCmd.PersistentFlags().StringVarP(&controlPlaneNamespace, "linkerd-namespace", "l", defaultNamespace, "Namespace in which Linkerd is installed [$LINKERD_NAMESPACE]")
+	RootCmd.PersistentFlags().StringVarP(&controlPlaneNamespace, "linkerd-namespace", "L", defaultNamespace, "Namespace in which Linkerd is installed [$LINKERD_NAMESPACE]")
 	RootCmd.PersistentFlags().StringVarP(&cniNamespace, "cni-namespace", "", defaultCNINamespace, "Namespace in which the Linkerd CNI plugin is installed")
 	RootCmd.PersistentFlags().StringVar(&kubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests")
 	RootCmd.PersistentFlags().StringVar(&kubeContext, "context", "", "Name of the kubeconfig context to use")
--- a/cli/cmd/stat.go
+++ b/cli/cmd/stat.go
@@ -25,6 +25,7 @@ type statOptions struct {
 	fromNamespace string
 	fromResource  string
 	allNamespaces bool
+	labelSelector string
 }
 
 type indexedResults struct {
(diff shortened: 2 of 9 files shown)


mayankshah1607:
#### Need help :
From what has been implemented currently, it looks like `buildStatsSummaryRequests()` method returns a slice of `StatSummaryRequest` where each request has `Selector.LabelSelector` being read in correctly from the `-l` or `--selector` flag.
For eg:
```bash
$ linkerd stat deploy -n emojivoto -l test-label/test
```
returns a request object where `label_selector` correctly holds the value `test-label/test`.
However, the label is being ignored while fetching the stats from API - all deployments in the namespace `linkerd` are returned, and not just the ones with the specified label. I'm not really able to figure out whats happening at `requestStatsFromAPI` to diagnose this issue further.

Pothulapati:
Hey, 

Hmmm, I don't think the `linkerd-controller` api server is actually using the label selectors.

If we look at the `StatSummary` method in the `linkerd-controller` api server, which calls `k8sResourceQuery` to imply a request for kubernetes resources query. This method at https://github.com/linkerd/linkerd2/blob/9b29a915d31f33914cdb94d71489d98f7e31c9a1/controller/api/public/stat_summary.go#L162 is called to retrieve the k8s resources, but I don't see the `selectors` being used anywhere in this path.

I also see that we are using the `labelSelectors` in the `ListPods` path at https://github.com/linkerd/linkerd2/blob/9b29a915d31f33914cdb94d71489d98f7e31c9a1/controller/api/public/grpc_server.go#L104, Thus making `ListPods` tests with `LabelSelectors` pass at  https://github.com/linkerd/linkerd2/blob/9b29a915d31f33914cdb94d71489d98f7e31c9a1/controller/api/public/grpc_server_test.go#L338

I think, using the same labelSelectors logic in the `ListPods` at the `getKubernetesObjectStats` function in the `StatSummary` path should fix this.

mayankshah1607:
thanks @Pothulapati ! That fixed the problem. I've updated the controller api server to use label selectors from requests. Had to do a small amount of refactoring, but it works as expected now :smile: 
As of now, I've only updated methods relevant to `stat` to make use of label selectors, but it looks like there's a lot of other places where label selectors are being ignored.

adleong:
This looks like a good start!

mayankshah1607:
@Pothulapati @adleong oops, tried to rebase against master and the integration tests are failing. Any tips on how to fix this?

mayankshah1607:
@alpeb Thank you for pointing that out! :smile: I have updated the PR to address these changes. Also added a small helper to make things easier

mayankshah1607:
@alpeb oops, my apologies! Will keep in mind 😄 thanks for the ✅

Issue #3204, opened 2019-08-07
air gapped installation: install option like --registry for prometheus image


kellermaennchen:
## Feature Request
Add another option like the --registry option to the linkerd install CLI command which specifies from which registry the prometheus image should be pulled 
### What problem are you trying to solve?
Air gapped installation: In secure On-Prem k8s installations internet access is not an option
### How should the problem be solved?
By adding another option like the --registry option to the linkerd install CLI command

What do you want to happen? Add any considered drawbacks.
I want the prometheus image to be pulled from another (private) registry and not only from DockerHub

### Any alternatives you've considered?
Add the prometheus image to the linkerd registry ("gcr.io/linkerd-io")

Is there another way to solve this problem that isn't as good a solution?
Add the prometheus image to the linkerd registry ("gcr.io/linkerd-io")

### How would users interact with this feature?
Through the linkerd install CLI command

If you can, explain how users will be able to use this. Maybe some sample CLI
output?
linkerd install --registry some.registry --promregistry another.registry | kubectl apply -f -

grampelberg:
@ihcsim @alpeb I feel like the intent of `--registry` was to solve this exact problem, but I'm hesitant to have it override different defaults. I might be over optimizing, but I'd love to avoid adding one more configuration option. Any good ideas?

alpeb:
Sounds to me like a specific enough case to be addressed through a Helm variable or kustomize.

ihcsim:
@kellermaennchen Sounds like you have access to GCR? If you access to GCR isn't limited to only `gcr.io/linkerd-io`, will using the [GCR DockerHub mirror](https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring) be an option for you? It should automatically redirect your request for the `prom/prometheus` image to the mirror.

Alternately, there is Helm and Kustomize, as @alpeb suggested.

The reason why we are hesitant to add yet another CLI option is so that we can maintain a sane list of configuration. Let us know if any of these alternatives will work for you.

kellermaennchen:
@ihcsim No I don't have access to GCR from my k8s cluster therefore the mirror is not an option. Because of that I used the --registry option and pointed it to my local registry. This worked fine except for the prometheus image. 
If as @grampleberg stated the purpose of the --registry option was to support air gapped installations it is useless as long as you need access to DockerHub for the prometheus image. If you don't want to add another option to the CLI installation another solution could be to pull the prometheus image from the registry given by the --registry option as well.
[...]

PR #4041, opened 2020-02-11
Support for custom prometheus registry

christyjacob4:
This PR contains the proposed changes for allowing custom registries for the prometheus image.
Draft for #3204

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -176,6 +176,7 @@ func newInstallOptionsWithDefaults() (*installOptions, error) {
 		controlPlaneVersion:         version.Version,
 		controllerReplicas:          defaults.ControllerReplicas,
 		controllerLogLevel:          defaults.ControllerLogLevel,
+		prometheusImage:             defaults.PrometheusImage,
 		highAvailability:            defaults.Global.HighAvailability,
 		controllerUID:               defaults.ControllerUID,
 		disableH2Upgrade:            !defaults.EnableH2Upgrade,
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -192,7 +192,6 @@ func newInstallOptionsWithDefaults() (*installOptions, error) {
 			debugImage:             defaults.DebugContainer.Image.Name,
 			debugImageVersion:      version.Version,
 			dockerRegistry:         defaultDockerRegistry,
-			prometheusRegistry:     defaultDockerRegistry,
 			imagePullPolicy:        defaults.Global.ImagePullPolicy,
 			ignoreInboundPorts:     nil,
 			ignoreOutboundPorts:    nil,
@@ -663,7 +662,7 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*l5d
 	installValues.Global.HighAvailability = options.highAvailability
 	installValues.Global.ImagePullPolicy = options.imagePullPolicy
 	installValues.GrafanaImage = fmt.Sprintf("%s/grafana", options.dockerRegistry)
-	installValues.PrometheusImage = fmt.Sprintf("%s/prometheus", options.prometheusRegistry)
+	installValues.PrometheusImage = fmt.Sprintf("%s/%s", options.dockerRegistry, installValues.PrometheusImage)
 	installValues.Global.Namespace = controlPlaneNamespace
 	installValues.Global.CNIEnabled = options.cniEnabled
 	installValues.OmitWebhookSideEffects = options.omitWebhookSideEffects
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -41,6 +41,7 @@ type (
 		controlPlaneVersion         string
 		controllerReplicas          uint
 		controllerLogLevel          string
+		prometheusImage             string
 		highAvailability            bool
 		controllerUID               int64
 		disableH2Upgrade            bool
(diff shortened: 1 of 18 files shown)


christyjacob4:
@Pothulapati Okay I will make the requested changes. :) 
Sorry About that. I will update the PR message now :)
Thanks

christyjacob4:
@Pothulapati Can you please explain the difference b/w options & install values?

zaharidichev:
@christyjacob4 the options hold data that have been supplied by the command line arguments. The values is the data used to render the Helm chart. Thinkgs of it this way: the options are used to override some of the default values and customize what you actually get in the end. Does that answer the question?

christyjacob4:
@zaharidichev thanks for that. It makes it clearer to me now. Since I've never worked with Helm before, could you please tell me what it's used for?

Pothulapati:
@christyjacob4 Helm is like a package manager for kubernetes deployments. Applications like MySQL,etc are packaged as a helm template, which users can directly install on their cluster.

We also package Linkerd as a helm chart, which users can directly install on their cluster using helm

https://helm.sh/

https://linkerd.io/2/tasks/install-helm/

christyjacob4:
@Pothulapati Thanks so much. Much clearer now. So its sort of like a Dockerfile ?

Pothulapati:
Yep, you can think of it like that. Let's please keep the issue comments regarding the issue itself.

Feel free to use the Linkerd slack for  conversations like this.

christyjacob4:
@Pothulapati would this be a right regex to validate a prometheus image string?

```golang
prometheusImage = regexp.MustCompile(`^[a-zA-Z]+\/[a-zA-Z]+\:v[0-9]+\.[0-9]+\.[0-9]+$`)
```

cpretzer:
@christyjacob4 the progress looks good on this.

Please be sure to document the new flag as well.

christyjacob4:
@cpretzer where do you want me to add the documentation?

christyjacob4:
@cpretzer it was suggested by @Pothulapati to open a draft PR so thay its clearer to others what changes I'm making. 
The prometheusImage is the flag I want to add support for

Pothulapati:
@christyjacob4 So, The problem is that we don't maintain the `prometheus` image in the `linkerd-io` image repo (like we do with grafana). So, we can't really pre-fix the `dockerRegistry` flag to promethues as we can't support the default install (i.e repo is `linkerd-io`). Should we maintain?  IMO, we should not.In that case, we could just make this,

```diff
diff --git a/cli/cmd/install.go b/cli/cmd/install.go
index 2934d6c0..56b61e17 100644
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -653,6 +653,10 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*l5d
                options.identityOptions.replicas = options.controllerReplicas
        }

+       if options.prometheusImage != "" {
+               installValues.PrometheusImage = options.prometheusImage
+       }
+
        globalJSON, proxyJSON, installJSON, err := config.ToJSON(configs)
        if err != nil {
                return nil, err
@@ -674,7 +678,6 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*l5d
        installValues.Global.HighAvailability = options.highAvailability
        installValues.Global.ImagePullPolicy = options.imagePullPolicy
        installValues.GrafanaImage = fmt.Sprintf("%s/grafana", options.dockerRegistry)
-       installValues.PrometheusImage = fmt.Sprintf("%s/%s", options.dockerRegistry, installValues.PrometheusImage)
        installValues.Global.Namespace = controlPlaneNamespace
        installValues.Global.CNIEnabled = options.cniEnabled
        installValues.OmitWebhookSideEffects = options.omitWebhookSideEffects
```
Thus allowing users to just give a full image name (including registry name) in the flag. Should we also consider the `dockerRegistry` flag when that is also passed with `--prometheus-image`? In that case, we could have another check inside, which pre-fixes it. @cpretzer

christyjacob4:
So how would you like me to proceed on this @Pothulapati @cpretzer ?

cpretzer:
@christyjacob4 

@Pothulapati provides a good snippet of code about how to handle this case and it is congruent with the comments from the [issue](https://github.com/linkerd/linkerd2/issues/3204):

> https://github.com/linkerd/linkerd2/issues/3204#issuecomment-583892305
https://github.com/linkerd/linkerd2/blob/master/test/install_test.go#L407

> https://github.com/linkerd/linkerd2/issues/3204#issuecomment-585043355

1. We shouldn't assume a registry other than the default registry
2. If the user specifies an image using this flag, then it should contain the registry and image name

Think about it this way: let's say BigBank has an airgapped environment and they use their own docker registry. They will source the prometheus image from some location and add it to their internal registry, so the repository could be something like: `docker.bigbank.internal`

They could even rename the image if they wanted to like: `bb-prom` and include their own tag `2.15.1_rc`

In that case, their install command might be like:
`linkerd install --prometheus-image docker.bigbank.internal/bb-prom:2.15.1_rc`

For this scenario, the snippet that @Pothulapati wrote would work. Additionally, for other users like `scrappystartup.com` who don't use a custom registry, they can omit the `--prometheus-image` flag and get the official image from docker hub.

I don't see added value to having constituent parts for the `repository`, `image`, and `tag` in an airgapped environment.

Hope that helps!

Pothulapati:
This looks good but I think we should update the regex now (as we include the registry name, etc stuff in the `prometheusImage`). Without that, I get image invalid error when I run
```
./bin/go-run cli install --ignore-cluster --prometheus-image docker.bigbank.internal/bb-prom:2.15.1_rc
```

christyjacob4:
@Pothulapati I will do that now... :)

christyjacob4:
@Pothulapati Do you want me to add the regex in the root.go file or install.go?

christyjacob4:
@cpretzer @Pothulapati Are we good to go on this?

PR #4044, opened 2020-02-11
Support providing an existing secret for the identity certs

mmiller1:
Our helm CI/CD deployment processes expect for all values required for installing/upgrading charts to be supplied in a values.yaml file that is committed to source control in plain text. The requirement of providing the sensitive certificates as files to helm does not suit our use-case as we have separate processes and flows for managing Kubernetes Secrets.

This will allow a user to optionally create the secret outside of this chart, and supply the secret name as a value to the chart, which will be mounted into the identity container instead of the default. Note that there is no change to the default behavior.

Tested as follows:
```
# create a secret resource containing crt.pem & key.pem values derived from the steps outlined
# here https://linkerd.io/2/tasks/generate-certificates/
$ kubectl apply -f linkerd-identity-issuer-manual-secret.yaml
secret/linkerd-identity-issuer-manual created
$ helm install . --set-file global.identityTrustAnchorsPEM=ca.crt --set identity.issuer.crtExpiry="2021-01-16T14:50:56Z" --set identity.issuer.existingTlsSecret=linkerd-identity-issuer-manual -n linkerd
NAME:   linkerd
LAST DEPLOYED: Tue Feb 11 14:10:53 2020
NAMESPACE: default
STATUS: DEPLOYED
...
```
Then validate linkerd check & linkerd dashboard work as expected.


grampelberg:
Have you looked at https://linkerd.io/2/tasks/use_external_certs/#using-helm ?

mmiller1:
No I had not, thank you. For the time being I believe setting "identity.issuer.scheme=kubernetes.io/tls" & making sure we use the secret name linkerd already expects will unblock us. Though perhaps the ability to change the secret name being used could also be beneficial in some manor?

grampelberg:
@mmiller1 I'm hesitant to add a configuration knob unless it unlocks something specific.

mmiller1:
fair enough. thanks for the pointer to "use_external_certs"!

PR #4050, opened 2020-02-13
Linkerd CLI command to get control plane diagnostics

### LINKED ISSUE #3116: linkerd command to get control plane metrics

srv-twry:
The CLI command diagnostics will fetch the control plane metrics directly from containers.

Sample output: https://gist.github.com/srv-twry/8e5f649edfa8bd34dcee286cbf913d5d

Fixes #3116 

### DIFF
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -3,10 +3,7 @@ package cmd
 import (
 	"bytes"
 	"fmt"
-	"hash/fnv"
 	"time"
-
-	"github.com/fatih/color"
 
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/spf13/cobra"
@@ -18,33 +15,13 @@ const (
 )
 
 type diagnosticsOptions struct {
-	wait    time.Duration
-	noColor bool
+	wait time.Duration
 }
 
 func newDiagnosticsOptions() *diagnosticsOptions {
 	return &diagnosticsOptions{
-		wait:    30 * time.Second,
-		noColor: false,
+		wait: 30 * time.Second,
 	}
-}
-
-var colorList = [][2]*color.Color{
-	{color.New(color.FgHiCyan), color.New(color.FgCyan)},
-	{color.New(color.FgHiGreen), color.New(color.FgGreen)},
-	{color.New(color.FgHiMagenta), color.New(color.FgMagenta)},
-	{color.New(color.FgHiYellow), color.New(color.FgYellow)},
-	{color.New(color.FgHiBlue), color.New(color.FgBlue)},
-	{color.New(color.FgHiRed), color.New(color.FgRed)},
-}
-
-func determineColor(podName string) (podColor, containerColor *color.Color) {
-	hash := fnv.New32()
-	hash.Write([]byte(podName))
-	idx := hash.Sum32() % uint32(len(colorList))
-
-	colors := colorList[idx]
-	return colors[0], colors[1]
 }
 
 func newCmdDiagnostics() *cobra.Command {
@@ -55,8 +32,6 @@ func newCmdDiagnostics() *cobra.Command {
 		Short: "Fetch metrics directly from the Linkerd control plane containers",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			color.NoColor = options.noColor
-
 			k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
 			if err != nil {
 				return err
@@ -71,15 +46,11 @@ func newCmdDiagnostics() *cobra.Command {
 
 			var buf bytes.Buffer
 			for i, result := range results {
-				podColor, containerColor := determineColor(result.pod)
-				p := podColor.SprintFunc()
-				c := containerColor.SprintFunc()
-
-				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", p(result.pod), i+1, len(results))
+				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", result.pod, i+1, len(results))
 				if result.err != nil {
 					content += fmt.Sprintf("# ERROR %s\n", result.err)
 				} else {
-					content += fmt.Sprintf("# CONTAINER %s \n#\n", c(result.container))
+					content += fmt.Sprintf("# CONTAINER %s \n#\n", result.container)
 					content += string(result.metrics)
 				}
 				buf.WriteString(content)
@@ -91,7 +62,6 @@ func newCmdDiagnostics() *cobra.Command {
 	}
 
 	cmd.Flags().DurationVarP(&options.wait, "wait", "w", options.wait, "Time allowed to fetch diagnostics")
-	cmd.PersistentFlags().BoolVarP(&options.noColor, "no-color", "n", options.noColor, "Disable colorized output")
 
 	return cmd
 }
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -3,7 +3,10 @@ package cmd
 import (
 	"bytes"
 	"fmt"
+	"hash/fnv"
 	"time"
+
+	"github.com/fatih/color"
 
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/spf13/cobra"
@@ -15,13 +18,33 @@ const (
 )
 
 type diagnosticsOptions struct {
-	wait time.Duration
+	wait    time.Duration
+	noColor bool
 }
 
 func newDiagnosticsOptions() *diagnosticsOptions {
 	return &diagnosticsOptions{
-		wait: 30 * time.Second,
+		wait:    30 * time.Second,
+		noColor: false,
 	}
+}
+
+var colorList = [][2]*color.Color{
+	{color.New(color.FgHiCyan), color.New(color.FgCyan)},
+	{color.New(color.FgHiGreen), color.New(color.FgGreen)},
+	{color.New(color.FgHiMagenta), color.New(color.FgMagenta)},
+	{color.New(color.FgHiYellow), color.New(color.FgYellow)},
+	{color.New(color.FgHiBlue), color.New(color.FgBlue)},
+	{color.New(color.FgHiRed), color.New(color.FgRed)},
+}
+
+func determineColor(podName string) (podColor, containerColor *color.Color) {
+	hash := fnv.New32()
+	hash.Write([]byte(podName))
+	idx := hash.Sum32() % uint32(len(colorList))
+
+	colors := colorList[idx]
+	return colors[0], colors[1]
 }
 
 func newCmdDiagnostics() *cobra.Command {
@@ -32,6 +55,8 @@ func newCmdDiagnostics() *cobra.Command {
 		Short: "Fetch metrics directly from the Linkerd control plane containers",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			color.NoColor = options.noColor
+
 			k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
 			if err != nil {
 				return err
@@ -46,14 +71,16 @@ func newCmdDiagnostics() *cobra.Command {
 
 			var buf bytes.Buffer
 			for i, result := range results {
-				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", result.pod, i+1, len(results))
-				if result.container != "" {
-					content += fmt.Sprintf("# CONTAINER %s \n#\n", result.container)
-				}
-				if result.err == nil {
-					content += string(result.metrics)
-				} else {
+				podColor, containerColor := determineColor(result.pod)
+				p := podColor.SprintFunc()
+				c := containerColor.SprintFunc()
+
+				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", p(result.pod), i+1, len(results))
+				if result.err != nil {
 					content += fmt.Sprintf("# ERROR %s\n", result.err)
+				} else {
+					content += fmt.Sprintf("# CONTAINER %s \n#\n", c(result.container))
+					content += string(result.metrics)
 				}
 				buf.WriteString(content)
 			}
@@ -64,6 +91,7 @@ func newCmdDiagnostics() *cobra.Command {
 	}
 
 	cmd.Flags().DurationVarP(&options.wait, "wait", "w", options.wait, "Time allowed to fetch diagnostics")
+	cmd.PersistentFlags().BoolVarP(&options.noColor, "no-color", "n", options.noColor, "Disable colorized output")
 
 	return cmd
 }
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
 	"time"
 
@@ -43,14 +44,17 @@ func newCmdDiagnostics() *cobra.Command {
 
 			results := getMetrics(k8sAPI, pods.Items, adminHTTPPortName, options.wait, verbose)
 
+			var buf bytes.Buffer
 			for i, result := range results {
-				fmt.Printf("#\n# POD %s (%d of %d)\n# CONTAINER %s \n#\n", result.pod, i+1, len(results), result.container)
+				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n# CONTAINER %s \n#\n", result.pod, i+1, len(results), result.container)
 				if result.err == nil {
-					fmt.Printf("%s", result.metrics)
+					content += string(result.metrics)
 				} else {
-					fmt.Printf("# ERROR %s\n", result.err)
+					content += fmt.Sprintf("# ERROR %s\n", result.err)
 				}
+				buf.WriteString(content)
 			}
+			fmt.Printf("%s", buf.String())
 
 			return nil
 		},
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -46,7 +46,10 @@ func newCmdDiagnostics() *cobra.Command {
 
 			var buf bytes.Buffer
 			for i, result := range results {
-				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n# CONTAINER %s \n#\n", result.pod, i+1, len(results), result.container)
+				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", result.pod, i+1, len(results))
+				if result.container != "" {
+					content += fmt.Sprintf("# CONTAINER %s \n#\n", result.container)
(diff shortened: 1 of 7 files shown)


srv-twry:
@adleong Hi! 

I have continued from #3579 and made some changes in order to reduce duplicate code between the `metrics` and the `diagnostics` commands.

It would be very helpful if you can have a look and let me know if they seem okay to you. I'll add a few unit tests after that.

Here is a short summary of the control-flow of the two commands after these changes:

1. Each of the two gets the list of pods to get the metrics from.
2. The `getMetrics` method inside `cli/cmd/metrics_diagonals_utils.go` receives the list of pods and a `portName`. It uses the `getAllContainersWithPort` function to get the list of all the containers inside those pods which expose their metrics at `portName`.
3. For each of the containers, `getContainerMetrics` is called which create a port forward at the specified `portName` using the `NewContainerMetricsForward` function and metrics are read from their exposed endpoints.
4. Results printed by the CLI

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,162 @@
...
+					bytes, err := getContainerMetrics(k8sAPI, pod, c, emitLogs, portName)
+
+					resultChan <- metricsResult{
+						pod:       pod.GetName(),
+						container: c.Name,
+						metrics:   bytes,
+						err:       err,
+					}
+
+					defer wg.Done()
```
adleong: does this need to be a defer?  since it's at the end of the function, can't we just call `wg.Done()` directly?

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,162 @@
...
+	portName string,
+	waitingTime time.Duration,
+	emitLogs bool,
+) []metricsResult {
+	var results []metricsResult
+
+	var wg sync.WaitGroup
+	resultChan := make(chan metricsResult)
+	for i := range pods {
+		go func(pod corev1.Pod) {
```
adleong: I think it's probably overkill to use goroutines to traverse the list of pods in parallel.  All the real work is happening in `getContainerMetrics` which is already wrapped in a goroutine.  I think it would be simpler to iterate over all the pods in serial since we'll still spawn a goroutine for each container anyway.

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,162 @@
...
+					}
+
+					defer wg.Done()
+				}(containers[i], &wg)
+			}
+
+		}(pods[i])
+	}
+
+	c := make(chan struct{})
```
adleong: I think the use of a waitgroup makes this more convoluted than it needs to be.  We spawn a goroutine to watch the waitgroup, which then closes a channel, which sets a boolean.  😵 

What about a simple counter which uses sync/atomic to increment when a container metrics goroutine is spawned and decrement when it completes.  Then we can simply check that the counter is 0 to see if we are done.

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,77 @@
...
+		Short: "Fetch metrics directly from the Linkerd control plane containers",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+			if err != nil {
+				return err
+			}
+
+			timeoutSeconds := int64(30)
+			deployments, err := k8sAPI.AppsV1().Deployments(controlPlaneNamespace).List(metav1.ListOptions{TimeoutSeconds: &timeoutSeconds})
```
adleong: We don't typically specify a timeout for reads like this.  I'm not saying it's a bad idea, but if we were to start doing it, I'd like to do so in a principled and consistent way that's outside the scope of this PR.

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,77 @@
...
+			if err != nil {
+				return err
+			}
+
+			// ensure we can connect to the public API before fetching the diagnostics.
+			checkPublicAPIClientOrRetryOrExit(time.Now().Add(options.wait), true)
```
adleong: do we need this?

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,77 @@
...
+
+			var pods []corev1.Pod
```
adleong: I think this whole thing would be simpler as:

```
k8sAPI.CoreV1().Pods(controlPlaneNamespace).List(metav1.ListOptions{})
```

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,77 @@
...
+	adminHTTPPortName string = "admin-http"
+)
+
+type diagnosticsOptions struct {
+	wait time.Duration
+}
+
+func newDiagnosticsOptions() *diagnosticsOptions {
+	return &diagnosticsOptions{
+		wait: 300 * time.Second,
```
adleong: For the metrics command, the timeout is 30s.  Is that what you meant here instead of 300?

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,77 @@
...
+					continue
+				}
+
+				pods = append(pods, p...)
+			}
+
+			results := getMetrics(k8sAPI, pods, adminHTTPPortName, options.wait, verbose)
+
+			for i, result := range results {
+				fmt.Printf("#\n# POD %s (%d of %d)\n# CONTAINER %s (%d of %d)\n#\n", result.pod, i+1, len(results), result.container, i+1, len(results))
```
adleong: We don't know how many pods there are and how many containers per pod just by looking at the length of the results.  We could calculate it, but I don't know that it's particularly useful.  I'd lean towards just omitting the counts here.

cli/cmd/metrics_diagnostics_util.go
srv-twry: Yeah right.

cli/cmd/metrics_diagnostics_util.go
srv-twry: sure, makes sense.

cli/cmd/diagnostics.go
srv-twry: Yes, I intended to change it to 30s.

I had copied it from the `check` command which sets `wait` to 300 but forgot to change it. Thanks.

cli/cmd/diagnostics.go
srv-twry: umm, I am not really sure about this. 

I went through the architecture shown in the documentation [here](https://linkerd.io/2/reference/architecture/). Since the CLI talks to the Public-API, I felt that we should check if it can connect to the API or not. Maybe I am missing something?

cli/cmd/metrics_diagnostics_util.go
srv-twry: Done. It does look much cleaner. Thanks :)

cli/cmd/diagnostics.go
adleong: I think this subcommand gets everything it needs from port-forward and doesn't use the Linkerd public API at all.

srv-twry: Understood. Thanks.

srv-twry:
@Pothulapati Thanks for the review!

Yeah sure, I'll try some colouring on the pod names. Good suggestion :+1:

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,150 @@
...
+	emitLogs bool,
+) []metricsResult {
+	var results []metricsResult
+
+	resultChan := make(chan metricsResult)
+	var activeRoutines int32
+	for _, pod := range pods {
+		containers, err := getAllContainersWithPort(pod, portName)
+		if err != nil {
+			fmt.Println(err)
```
ihcsim: Generally, it's better to have a consistent error-handling mechanism. Is it possible to wrap this in a `metricsResult` instance, containing just the `pod` and `err` fields, and append it to the result slice?

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, err
+	}
+
+	defer portForward.Stop()
+	if err = portForward.Init(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
+	}
+
+	metricsURL := portForward.URLFor("/metrics")
+	bytes, err := getResponse(metricsURL)
```
ihcsim: How about just `return getResponse(metricsURL)`?

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// getResponse makes a http Get request to the passed url and returns the response/error
+func getResponse(url string) ([]byte, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	bytes, err := ioutil.ReadAll(resp.Body)
```
ihcsim: How about just `return getResponse(metricsURL)`?

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,150 @@
...
+				bytes, err := getContainerMetrics(k8sAPI, p, c, emitLogs, portName)
+
+				resultChan <- metricsResult{
+					pod:       p.GetName(),
+					container: c.Name,
+					metrics:   bytes,
+					err:       err,
+				}
+
+				atomic.AddInt32(&activeRoutines, -1)
```
ihcsim: I think we should `defer` this near the top of this goroutine. If line 123 panic, then this won't get executed and your subsequent `for` loop won't exit.

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,150 @@
...
+	portName string,
+) ([]byte, error) {
+	portForward, err := k8s.NewContainerMetricsForward(k8sAPI, pod, container, emitLogs, portName)
+	if err != nil {
+		return nil, err
+	}
+
+	defer portForward.Stop()
+	if err = portForward.Init(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
```
ihcsim: if `err != nil`, don't we want to `return`?

pkg/k8s/portforward.go
```diff
@@ -33,38 +33,25 @@ type PortForward struct {
...
 	for _, p := range container.Ports {
-		if p.Name == ProxyAdminPortName {
+		if p.Name == portName {
 			port = p
 			break
 		}
 	}
-	if port.Name != ProxyAdminPortName {
-		return nil, fmt.Errorf("no %s port found for container %s/%s", ProxyAdminPortName, pod.GetName(), container.Name)
+	if port.Name != portName {
```
ihcsim: Observation (no changes required): It doesn't feel DRY to pass both the pod and one of its container into a function, loop through the ports to check if the expected one is present, where we can just let the underlying `k8s.io/client-go/tools/portforward` package handles the error if the port is missing.

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,62 @@
...
+			if err != nil {
+				return err
+			}
+
+			pods, err := k8sAPI.CoreV1().Pods(controlPlaneNamespace).List(metav1.ListOptions{})
+			if err != nil {
+				return err
+			}
+
+			results := getMetrics(k8sAPI, pods.Items, adminHTTPPortName, options.wait, verbose)
```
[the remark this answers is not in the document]
srv-twry: Actually `getMetrics()` is also being used in the `metrics` command as well which needs to look at a different port name(`ProxyAdminPortName`). In order to make it useful for both the commands, it needs to accept the port name as a parameter.

cli/cmd/diagnostics.go
```diff
@@ -0,0 +1,62 @@
...
+
+			for i, result := range results {
+				fmt.Printf("#\n# POD %s (%d of %d)\n# CONTAINER %s \n#\n", result.pod, i+1, len(results), result.container)
```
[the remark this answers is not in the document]
srv-twry: Sure

cli/cmd/metrics_diagnostics_util.go
srv-twry: Yeah right. Thanks

cli/cmd/metrics_diagnostics_util.go
srv-twry: Yeah. Missed that. Thanks

cli/cmd/metrics_diagnostics_util.go
srv-twry: In this case, we'll have to do something like this in `cli/cmd/diagnostics`

```golang
content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", result.pod, i+1, len(results))
if result.container != "" {
    content += fmt.Sprintf("# CONTAINER %s \n#\n", result.container)
}
```

ihcsim: Sure :+1:. Or if you want to reduce the if/else, you can do:

```diff
diff --git a/cli/cmd/diagnostics.go b/cli/cmd/diagnostics.go
index 70af43fc..eed45d1c 100644
--- a/cli/cmd/diagnostics.go
+++ b/cli/cmd/diagnostics.go
@@ -47,13 +47,11 @@ func newCmdDiagnostics() *cobra.Command {
 			var buf bytes.Buffer
 			for i, result := range results {
 				content := fmt.Sprintf("#\n# POD %s (%d of %d)\n", result.pod, i+1, len(results))
-				if result.container != "" {
+				if result.err != nil {
+					content += fmt.Sprintf("# ERROR %s\n", result.err)
+				} else {
 					content += fmt.Sprintf("# CONTAINER %s \n#\n", result.container)
-				}
-				if result.err == nil {
 					content += string(result.metrics)
-				} else {
-					content += fmt.Sprintf("# ERROR %s\n", result.err)
 				}
 				buf.WriteString(content)
 			}
```

cli/cmd/metrics_diagnostics_util.go
```diff
@@ -0,0 +1,144 @@
...
+	emitLogs bool,
+) []metricsResult {
+	var results []metricsResult
+
+	resultChan := make(chan metricsResult)
+	var activeRoutines int32
+	for _, pod := range pods {
+		containers, err := getAllContainersWithPort(pod, portName)
+		if err != nil {
+			resultChan <- metricsResult{
```
ihcsim: Unfortunately, this is going to cause the sender to block, because the `resultChan` hasn't been set up to receive data yet. You can test it by changing the conditional to `if err == nil`. I think the easiest solution here is to turn `resultChan` into a buffered channel: 

```diff
diff --git a/cli/cmd/metrics_diagnostics_util.go b/cli/cmd/metrics_diagnostics_util.go
index 49dd86f0..470863a9 100644
--- a/cli/cmd/metrics_diagnostics_util.go
+++ b/cli/cmd/metrics_diagnostics_util.go
@@ -97,7 +97,7 @@ func getMetrics(
 ) []metricsResult {
 	var results []metricsResult
 
-	resultChan := make(chan metricsResult)
+	resultChan := make(chan metricsResult, len(pods))
 	var activeRoutines int32
 	for _, pod := range pods {
 		containers, err := getAllContainersWithPort(pod, portName)
```

Alternately, just spawn a goroutine to handle each pod, and don't bother with the goroutine at line 114 below.

srv-twry: Yeah changing it to a buffered channel was helpful.

Although after changing it to a buffered channel I faced another problem because of line 136 where we check if the `activeRoutines` is zero or not. After I changed the conditional to `if err == nil`, I was only getting 1 entry in the `results` slice since the `for` loop of line 129 was exiting after receiving one entry from the `resultsChan` since `activeRoutines` was still zero. 

In order to deal with it, I simply removed the `activeRoutines` variable and declared a `expectedLen` variable indicating the expected length of the results.

cli/cmd/metrics_diagnostics_util.go
srv-twry: done

cli/cmd/metrics_diagnostics_util.go
ihcsim: > After I changed the conditional to if err == nil

That is meant to reproduce the blocking problem. We still want to keep it as `if err != nil`.

srv-twry: I think you misunderstood. I'll explain the problem in detail now from the start. I do understand that we still want to keep `if err != nil`.

Consider this snippet

```golang
resultChan := make(chan metricsResult, len(pods))
var activeRoutines int32
for _, pod := range pods {
	containers, err := getAllContainersWithPort(pod, portName)
	if err != nil {
		resultChan <- metricsResult{
			pod: pod.GetName(),
			err: err,
		}
		continue
	}

	for i := range containers {
		atomic.AddInt32(&activeRoutines, 1)
		go func(p corev1.Pod, c corev1.Container) {
			defer atomic.AddInt32(&activeRoutines, -1)
			// get metrics and send to channel
		}(pod, containers[i])
	}
}

for {
	select {
	case result := <-resultChan:
		results = append(results, result)
	case <-time.After(waitingTime):
		break // timed out
	}
	if atomic.LoadInt32(&activeRoutines) == 0 {
		break
	}
}
```

Imagine if there was an error in getting the containers for all of the pods i.e. the `getAllContainersWithPort` method returns an error for each pod of the `pods` slice. This would mean that the `activeRoutines` variable stays set to 0. Now consider the `for` loop where we check if the `activeRoutines` variable is 0 or not.
In this particular case, the `if` condition will be satisfied and the loop will break just after getting one entry from the `resultsChan`.

srv-twry:
@Pothulapati I have added colours for the pod and container names in the output of the command.

[colorful_output.txt](https://github.com/linkerd/linkerd2/files/4239747/colorful_output.txt)

I am not really in favour of doing this though because the impact of this is minimal. Note that the output of the `logs` command has multiple pods interleaved among them. In the case of `diagnostics`, they are sequential.

cli/cmd/metrics_diagnostics_util.go
srv-twry: > Alternately, just spawn a goroutine to handle each pod, and don't bother with the goroutine at line 114 below.

This works well and doesn't have the above problem. Hence, I have gone for this option

ihcsim:
I agree with removing coloring the output, because per your attached output, those color unicode characters only get in the way of debugging when users send us their metrics text output. Also, that package is no longer maintained.

cli/cmd/metrics_diagnostics_util.go
ihcsim: Glad the second approach works. With the first approach, there are ways to determine if a buffered channel still has any unread elements.

PR #4052, opened 2020-02-13
Support for configuring service profile retries(x-linkerd-retryable) via openapi spec

### LINKED ISSUE #2073: add support for configuring service profile retries via openapi spec
## Feature Request

with https://github.com/linkerd/linkerd2/pull/2038 being a thing. It'll be nice is we can use openapi extensions to define a retry policy per route.

We do not manually write our openapi spec. It's generated automatically by middleware within our project. Therefore generating a retry policy would be tedious with us having to manually write a separate yaml file with the retry policies and patch it in to the service profile.

It's possible to define extensions (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#specificationExtensions) within our openapi / swagger middleware. These will appear on the operation object (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#operationObject) in the swagger.json. linkerd can look for the extension name (maybe `x-linkerd-retry`, or something more suitable) to populate the retry policy

Patching the yaml before it gets to kubectl

add an extension `x-linkerd-retry: 2` to either the `Path Item Object` or `Operation Object`. linkerd will pick this up while processing the openapi spec.

kohsheen1234:
Retry policy is manually written in yaml file and patched it into the service profile

 Added support for configuring service profile retries(x-linkerd-retryable) via openapi spec

 Fixes #2073
    
### DIFF
--- a/pkg/profiles/openapi.go
+++ b/pkg/profiles/openapi.go
@@ -73,7 +73,7 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 		path := path.Join(swagger.BasePath, relPath)
 		pathRegex := pathToRegex(path)
 		if item.Delete != nil {
-			spec := mkRouteSpec(path, pathRegex, http.MethodGet, item.Delete.Responses, item.Delete)
+			spec := mkRouteSpec(path, pathRegex, http.MethodDelete, item.Delete.Responses, item.Delete)
 			routes = append(routes, spec)
 		}
 		if item.Get != nil {
--- a/pkg/profiles/openapi.go
+++ b/pkg/profiles/openapi.go
@@ -110,7 +110,7 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 
 func mkRouteSpec(path, pathRegex string, method string, operation *spec.Operation) *sp.RouteSpec {
 	retryable := false
-	var responses = (*spec.Responses)(nil)
+	var responses *spec.Responses
 	if operation != nil {
 		retryable, _ = operation.VendorExtensible.Extensions.GetBool(xLinkerdRetryable)
 		responses = operation.Responses
--- a/pkg/profiles/openapi.go
+++ b/pkg/profiles/openapi.go
@@ -47,6 +47,10 @@ func RenderOpenAPI(fileName, namespace, name, clusterDomain string, w io.Writer)
 	return writeProfile(profile, w)
 }
 
+func getXLinkerdRetryValue(value bool, ok bool) bool {
+	return value
+}
+
 func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomain string) sp.ServiceProfile {
 	profile := sp.ServiceProfile{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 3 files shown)


pkg/profiles/openapi.go
```diff
@@ -71,31 +75,38 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 		path := path.Join(swagger.BasePath, relPath)
 		pathRegex := pathToRegex(path)
 		if item.Delete != nil {
-			spec := mkRouteSpec(path, pathRegex, http.MethodDelete, item.Delete.Responses)
+			retryable, _ := item.Delete.VendorExtensible.Extensions.GetBool("x-linkerd-retryable")
+			spec := mkRouteSpec(path, pathRegex, http.MethodGet, item.Delete.Responses, retryable)
```
adleong: I think you can clean up some of the repetition by having `mkRouteSpec` accept a `spec.Exceptions` and put the code for extracting the retryable boolean in `mkRouteSpec`.

pkg/profiles/openapi.go
```diff
@@ -71,31 +75,38 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 		path := path.Join(swagger.BasePath, relPath)
 		pathRegex := pathToRegex(path)
 		if item.Delete != nil {
-			spec := mkRouteSpec(path, pathRegex, http.MethodDelete, item.Delete.Responses)
+			retryable, _ := item.Delete.VendorExtensible.Extensions.GetBool("x-linkerd-retryable")
```
adleong: We should use a string constant for "x-linkerd-retryable" rather than inlining it

pkg/profiles/openapi_test.go
```diff
@@ -29,6 +31,9 @@ func TestSwaggerToServiceProfile(t *testing.T) {
 										},
 									},
 								},
+								VendorExtensible: spec.VendorExtensible{
+									Extensions: xLinkerdRetryable,
```
adleong: I think it's more clear if you inline the extensions here.  e.g.

```
Extensions: {"x-linkerd-retryable": true},
```

kohsheen1234:
@adleong I have made the required changes. Can you please review the changes.

pkg/profiles/openapi.go
```diff
@@ -47,6 +47,10 @@ func RenderOpenAPI(fileName, namespace, name, clusterDomain string, w io.Writer)
 	return writeProfile(profile, w)
 }
 
+func getXLinkerdRetryValue(value bool, ok bool) bool {
```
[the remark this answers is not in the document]
kohsheen1234: Done :)

pkg/profiles/openapi.go
kohsheen1234: @adleong Can you check now? I have tried to do some cleanup.

pkg/profiles/openapi.go
kohsheen1234: Sure. :)

pkg/profiles/openapi_test.go
kohsheen1234: Done 👍

pkg/profiles/openapi.go
```diff
@@ -71,31 +73,31 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 		path := path.Join(swagger.BasePath, relPath)
 		pathRegex := pathToRegex(path)
 		if item.Delete != nil {
-			spec := mkRouteSpec(path, pathRegex, http.MethodDelete, item.Delete.Responses)
+			spec := mkRouteSpec(path, pathRegex, http.MethodGet, item.Delete.Responses, item.Delete)
```
Pothulapati: This should still be `http.MethodDelete` right?

kohsheen1234: @Pothulapati Yes, you're right. I'll correct that.

pkg/profiles/openapi.go
```diff
@@ -104,11 +106,13 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 	return profile
 }
 
-func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses) *sp.RouteSpec {
+func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses, exception *spec.Operation) *sp.RouteSpec {
```
Pothulapati: I think we should add a nil check for exception here.

pkg/profiles/openapi.go
```diff
@@ -104,11 +106,13 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
...
+	xLinkerdRetryable, _ := exception.VendorExtensible.Extensions.GetBool(xLinkerdRetryable)
 	return &sp.RouteSpec{
 		Name:            fmt.Sprintf("%s %s", method, path),
 		Condition:       toReqMatch(pathRegex, method),
 		ResponseClasses: toRspClasses(responses),
+		IsRetryable:     xLinkerdRetryable,
 	}
```
Pothulapati: Should this instead be?
```suggestion
func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses, exception *spec.Operation) *sp.RouteSpec {
	retryable := false
	if exception != nil {
		retryable, _ = exception.VendorExtensible.Extensions.GetBool(xLinkerdRetryable)
	}
	return &sp.RouteSpec{
		Name:            fmt.Sprintf("%s %s", method, path),
		Condition:       toReqMatch(pathRegex, method),
		ResponseClasses: toRspClasses(responses),
		IsRetryable:     retryable,
	}
}
```

kohsheen1234: @Pothulapati I realised the same. :) Thank you very much.

pkg/profiles/openapi_test.go
```diff
@@ -29,6 +29,9 @@ func TestSwaggerToServiceProfile(t *testing.T) {
 										},
 									},
 								},
+								VendorExtensible: spec.VendorExtensible{
+									Extensions: spec.Extensions{"x-linkerd-retryable": true},
```
adleong: Let's use the xLinkerdRetryable const here

pkg/profiles/openapi.go
```diff
@@ -104,11 +106,16 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 	return profile
 }
 
-func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses) *sp.RouteSpec {
+func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses, exception *spec.Operation) *sp.RouteSpec {
```
adleong: I think we can clean this up a bit by just having it accept a `*spec.Operation` instead of both a `*spec.Responses` and a `*spec.Operation`.  We can get the responses from the `Operation` that is passed in.  

Why do you name the `Operation` parameter `exception`?  Should this be called something like `operation` or `op` instead?

kohsheen1234:
@adleong Can you have a look now. I have made the recommended changes. :)

kohsheen1234:
@adleong Can you please have a look now. :)

pkg/profiles/openapi.go
```diff
@@ -104,11 +108,18 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 	return profile
 }
 
-func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses) *sp.RouteSpec {
+func mkRouteSpec(path, pathRegex string, method string, operation *spec.Operation) *sp.RouteSpec {
+	retryable := false
+	var responses = (*spec.Responses)(nil)
+	if operation != nil {
+		retryable, _ = operation.VendorExtensible.Extensions.GetBool(xLinkerdRetryable)
```
cpretzer: @kohsheen1234 still wondering about the return value from `item.Delete.VendorExtensible.Extensions.GetBool`

Is there anything that we should do if the second value is `false`?

pkg/profiles/openapi.go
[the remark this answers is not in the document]
adleong: If the key is not present, the first value will default to false: https://github.com/go-openapi/spec/blob/master/info.go#L44

I think this is the desired behavior for us.  If the extension is absent, we will default to not being retryable.

pkg/profiles/openapi.go
```diff
@@ -104,11 +108,18 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 	return profile
 }
 
-func mkRouteSpec(path, pathRegex string, method string, responses *spec.Responses) *sp.RouteSpec {
+func mkRouteSpec(path, pathRegex string, method string, operation *spec.Operation) *sp.RouteSpec {
+	retryable := false
+	var responses = (*spec.Responses)(nil)
```
adleong: This can simply be `var responses *spec.Responses`.  Declaring a variable will automatically initialize it with the zero value (nil).

pkg/profiles/openapi.go
kohsheen1234: @cpretzer The second return value is not of concern to us, we need not do anything if the send value is `false`.

pkg/profiles/openapi.go
kohsheen1234: @adleong Thank you, this seems like a better way.:+1: I have made the change

pkg/profiles/openapi.go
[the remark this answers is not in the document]
cpretzer: 👍

pkg/profiles/openapi.go
cpretzer: LGTM, thanks for confirming

PR #4054, opened 2020-02-14
Allow custom host set from helm values 

### LINKED ISSUE #3961: Allow custom hosts set from Helm values for "enforced-host"

sannimichaelse:
Allow custom host set from helm values #3961

Users should be able to set their hosts or use deafult host if not specified in helm

Fixes #3961

Signed-off-by: Sanni Michael Tomiwa <sannimichaelse@gmail.com>

### DIFF
--- a/charts/linkerd2/values.yaml
+++ b/charts/linkerd2/values.yaml
@@ -6,7 +6,6 @@
 global:
   clusterDomain: &cluster_domain cluster.local
   imagePullPolicy: &image_pull_policy IfNotPresent
-  
   # Control Plane Trace Configuration
   controlPlaneTracing: false
 
@@ -75,6 +74,8 @@ global:
   controllerNamespaceLabel: linkerd.io/control-plane-ns
   linkerdNamespaceLabel: linkerd.io/is-control-plane
 
+# Enforced Host Validation Regular Expression
+enforcedHostRegexp: ""
 
 enableH2Upgrade: true
 
--- a/charts/linkerd2/values.yaml
+++ b/charts/linkerd2/values.yaml
@@ -6,7 +6,8 @@
 global:
   clusterDomain: &cluster_domain cluster.local
   imagePullPolicy: &image_pull_policy IfNotPresent
-  # Control Plane Trace Configuration
+
+ # control plane trace configuration
   controlPlaneTracing: false
 
   # control plane version. See Proxy section for proxy version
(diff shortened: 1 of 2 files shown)


charts/linkerd2/values.yaml
```diff
@@ -75,7 +75,8 @@ global:
   controllerNamespaceLabel: linkerd.io/control-plane-ns
   linkerdNamespaceLabel: linkerd.io/is-control-plane
 
-
+#Add custom host here
+enforcedHostRegexp: ""
```
Pothulapati: This could be
Make sure there is a new line below to separate
```
# Enforced Host Validation Regular Expression
enforcedHostRegexp: ""

```

charts/linkerd2/values.yaml
```diff
@@ -6,7 +6,6 @@
 global:
   clusterDomain: &cluster_domain cluster.local
   imagePullPolicy: &image_pull_policy IfNotPresent
-
```
alpeb: Can you please restore this empty line?

charts/linkerd2/values.yaml
alpeb: For the comment, I think we should follow the casing practice we've had in most of the rest of the file (lower-case).

sannimichaelse: Done

alpeb:
Thanks @sannimichaelse for your first contribution to Linkerd! :1st_place_medal: 
If you'd like to receive some Linkerd swag, I invite you to fill this [form](https://docs.google.com/forms/d/e/1FAIpQLSdXA7kAl6oyUQ4DFRjiDc2biWlvkhNirrMwsXaQ8V3Uh8Yj5A/viewform) :wink:

PR #4056, opened 2020-02-14
Switched figo to finleap

christianhuening:
Due to company mergers we're now finleap connect and not figo anymore. Therefore I'd like to update the URL and company name here.


grampelberg:
@christianhuening mind taking care of DCO? Check out the details of the check to understand what that means.

christianhuening:
Ah damn, sorry. I just used github edit. Am on a train now, will do later.

christianhuening:
done

PR #4058, opened 2020-02-14
bin/helm-build automatically updates version in values.yaml

alpeb:
Have the Helm charts building script (`bin/helm-build`) update the
linkerd version in the `values.yaml` files according to the tagged
version, thus removing the need of doing this manually on every release.

This is akin to the update we do in `version.go` at CLI build time.

Note that `shellcheck` is issuing some warnings about this script, but
that's on code that was already there, so that will be handled in an
followup PR.

### DIFF
--- a/bin/helm-build
+++ b/bin/helm-build
@@ -2,14 +2,14 @@
 
 set -e
 
-restoreValues() {
-    sed -i "s/$1/{version}/" charts/linkerd2/values.yaml
-    sed -i "s/$1/{version}/" charts/linkerd2-cni/values.yaml
+setValues() {
+    sed -i "s/$1/$2/" charts/linkerd2/values.yaml
+    sed -i "s/$1/$2/" charts/linkerd2-cni/values.yaml
 }
 
 showErr() {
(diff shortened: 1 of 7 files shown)


alpeb:
Still needs a bit more work to avoid breaking the CLI :grimacing:

alpeb:
Alright, ready again for review :+1:

bin/helm-build
```diff
@@ -2,8 +2,18 @@
 
 set -e
 
+restoreValues() {
```
kleimkuhler: Maybe consolidate setting and restoring the values?

```suggestion
setValues() {
    sed -i "s/$1/$2/" charts/linkerd2/values.yaml
    sed -i "s/$1/$2/" charts/linkerd2-cni/values.yaml
}
```

Then, setting and restoring would just be:

```
# Set version in Values files
setValues "{version}" $fullVersion
```

```
# Restore version in Values files
setValues $fullVersion "{version}"
```

alpeb:
> Great! Its probably also worth it to update the release process document :)

Yep, will do :+1:

bin/helm-build
alpeb: Good suggestion, will do :+1:

PR #4061, opened 2020-02-17
Update control-plane-namespace label

### LINKED ISSUE #3958: Changes to linkerd namespace spec ignored during upgrade
### What is the issue?
The recommended upgrade method: `linkerd upgrade | kubectl apply --prune -l linkerd.io/control-plane-ns=linkerd -f -` will currently ignore any changes to the `Namespace` object output from `linkerd upgrade` because the apply selector will drop said resource.

After a partial outage, we were surprised to discover that the `config.linkerd.io/admission-webhooks: disabled` label was not present on our `Namespace` object. After revisiting our previous `linkerd upgrade`, it appeared the label was present in the manifests output from the linkerd CLI, but was dropped by kubectl.

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version: 1.15.7
- Cluster Environment: GKE
- Host OS: cos
- Linkerd version: 2.6.0/2.6.1

supra08:
Upgrade command ignores changes to the namespace object

Add linkerd.io/control-plane-ns=linkerd label to the control-plane namespace

Fixes #3958

Signed-off-by: Supratik Das <rick.das08@gmail.com>


supra08:
Yeah I am a bit confused about this one. This fix requires an update of the `control-plane-namespace` from `linkerd.io/control-plane-ns` to `linkerd.io/control-plane-ns=linkerd`. Although things are not breaking, the unit tests failures are due to the `=` in the namespace label. Any way to prevent that?

supra08:
@zaharidichev please have a look.

testutil/kubernetes_helper.go
```diff
@@ -82,7 +82,7 @@ func (h *KubernetesHelper) createNamespaceIfNotExists(namespace string, annotati
 
 // CreateControlPlaneNamespaceIfNotExists creates linkerd control plane namespace.
 func (h *KubernetesHelper) CreateControlPlaneNamespaceIfNotExists(namespace string) error {
-	labels := map[string]string{"linkerd.io/is-control-plane": "true", "config.linkerd.io/admission-webhooks": "disabled"}
+	labels := map[string]string{"linkerd.io/is-control-plane": "true", "config.linkerd.io/admission-webhooks": "disabled", "linkerd.io/control-plane-ns": "linkerd"}
```
zaharidichev: Seems to me here we need to use the `namespace` variable for the name of the namespace :)

supra08: Done!

supra08:
Hi @zaharidichev I tried printing logs of the manifests and kubectl apply while testing but didn't get any thing to verify whether this was fixed. I finally just went with the fix provided by @grampelberg.  
I would be glad if you could help me with this.

zaharidichev:
@supra08 I do notthink logging will be of much use here. What I am refering to is trying to reproduce the situation that was described here https://github.com/linkerd/linkerd2/issues/3958#issue-553009554 and ensure that your change actually fixes it.

supra08:
Yeah I wasn't able to completely reproduce based on the information provided on the issue. I commented on it to get it clarified. Once it's clear I will get the tests done.

supra08:
Sorry for the delay, I misunderstood the scenario initially 😅. Here is what I inferred from the tests. 
I analyzed the output of `linkerd upgrade` before and after the label addition and made sure `linkerd.io/control-plane-ns: linkerd` is present in the stdout. 
And hence, it resulted in the `namespace/linkerd unchanged` object in the `kubectl apply` output. So this update fixes it.
Do I need to update the commit message likewise with the test?
@zaharidichev ^

ihcsim:
@supra08 Testing this issue is a bit tricky, as it requires understanding how `kubectl apply --prune` works, which still surprises me at times.

This is how you can reproduce the original bug:

1. Install Linkerd into the default `linkerd` namespace
1. Update the `charts/linkerd2/templates/namespace.yaml` with a new random label like `foo: bar`
1. Run `bin/go-run cli upgrade | less`. Notice that the namespace YAML has the new label
1. Run `bin/go-run cli upgrade | kubectl apply --prune -l linkerd.io/control-plane-ns=linkerd --dry-run -oyaml -f - | less`. Notice that the namespace YAML is missing, because `--prune` drops it
1. Run `bin/go-run cli upgrade | kubectl apply --prune -l linkerd.io/control-plane-ns=linkerd`. Notice that your new label is added to the `linkerd` namespace

Now apply your changes:

1. Run `bin/go-run cli upgrade | less`. Notice that your `foo: bar` label shows up this time
1. Run `bin/go-run cli upgrade | k apply --prune -l linkerd.io/control-plane-ns=linkerd`. Notice that your new label is added to the `linkerd` namespace

supra08:
@ihcsim thanks for the steps!
I did reproduce with my changes. `bin/go-run cli upgrade | kubectl apply --prune -l linkerd.io/control-plane-ns=linkerd -oyaml -f - | less` output contained the `Namespace` object, with the updated label along with `namespace/linkerd configured` as normal output. So I assume that it updates the namespace.

zaharidichev:
@supra08 @ihcsim I wonder if it makes sense tor try and put together an integration test that exercises precicel this scenario. We already have an upgrade integration test. Can we use that one. Should this be in this PR o follow up on you think ?

ihcsim:
> So I assume that it updates the namespace.

@supra08 if you don't mind, can you actually run it without the `-oyaml` option to make sure it really updates your namespace? Where possible, I prefer to eliminate all assumptions. Thanks.

@zaharidichev Maybe a different PR? Strictly speaking, the bug is caused by an unexpected outcome of `kubectl apply --prune`; `linkerd upgrade` did the right thing.

supra08:
Yeah sure! Actually I had a doubt, how can I view the applied labels to a namespace?
I am already getting `namespace/linkerd configured` in the `bin/go-run cli upgrade | kubectl apply --prune -l linkerd.io/control-plane-ns=linkerd -f -` output, on adding/updating a new label. I guess this means that the namespace is updated?

supra08:
After a `linkerd upgrade | k apply` , I inspected the namespace values using `kubectl describe ns linkerd`. Here is the output:
```
Name:         linkerd
Labels:       config.linkerd.io/admission-webhooks=disabled
              foo=bar
              linkerd.io/control-plane-ns=linkerd
              linkerd.io/is-control-plane=true
```
Before the command, I'd added `foo: bar` as a new label to the namespace and it reflected on the output.

PR #4070, opened 2020-02-19
Service mirror CLI 

zaharidichev:
Just putting this out here to encourage some feedback. This is based on the work that @adleong did around extracting credentials from remote clusters. I woner whether it makes sense to use go templates for rendering the credentials when creating them. Also, I wonder whether it is worth exposing a lot more parameters so you can pretty much override everything in your current k8s config context, instead of blindly copying it.

### DIFF
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -52,14 +52,29 @@ func newCmdCluster() *cobra.Command {
 	createOpts := createOptions{}
 	exportOpts := exportServiceOptions{}
 	clusterCmd := &cobra.Command{
+
 		Hidden: true,
-		Use:    "cluster manages the multicluster  setup for Linkerd",
-		Short:  "cluster manages the multicluster  setup for Linkerd",
+		Use:    "cluster [flags]",
 		Args:   cobra.NoArgs,
+		Short:  "Manages the multicluster setup for Linkerd",
+		Long: `Manages the multicluster setup for Linkerd.
+
+This command provides subcommands to manage the multicluster support
+functionality of Linkerd. You can use it to deploy credentials to
+remote clusters, extract them as well as export remote services to be
+available across clusters.`,
+		Example: `  # Create remote cluster credentials.
+  linkerd --context=cluster-a cluster create-credentials | kubectl --context=cluster-a apply -f -
+
+  # Extract mirroring cluster credentials from cluster A and install them on cluster B
+  linkerd --context=cluster-a cluster get-credentials --cluster-name=remote | kubectl apply --context=cluster-b -f -
+
+  # Export service from cluster A to be available to other clusters
+  linkerd --context=cluster-a cluster export-service --service-name=backend-svc --service-namespace=default --gateway-name=linkerd-gateway --gateway-ns=default`,
 	}
 
 	createCredentialsCommand := &cobra.Command{
-		Hidden: true,
+		Hidden: false,
 		Use:    "create-credentials",
 		Short:  "Create the necessary credentials for service mirroring",
 		Args:   cobra.NoArgs,
@@ -116,7 +131,7 @@ func newCmdCluster() *cobra.Command {
 	}
 
 	getCredentialsCmd := &cobra.Command{
-		Hidden: true,
+		Hidden: false,
 		Use:    "get-credentials",
 		Short:  "Get cluster credentials as a secret",
 		Args:   cobra.NoArgs,
@@ -216,7 +231,7 @@ func newCmdCluster() *cobra.Command {
 	}
 
 	exportServiceCmd := &cobra.Command{
-		Hidden: true,
+		Hidden: false,
 		Use:    "export-service",
 		Short:  "Exposes a remote service to be mirrored",
 		Args:   cobra.NoArgs,
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -1,0 +1,283 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/api/rbac/v1"
+
+	"github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+const (
+	tokenKey                  = "token"
+	defaultServiceAccountName = "linkerd-service-mirror"
+	defaultServiceAccountNs   = "default"
+	defaultClusterName        = "remote"
+)
+
+type (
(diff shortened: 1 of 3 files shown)


cli/cmd/cluster.go
```diff
@@ -0,0 +1,208 @@
...
+	createOpts := createOptions{}
+
+	clusterCmd := &cobra.Command{
+		Use:   "cluster manages the multicluster  setup for Linkerd",
+		Short: "cluster manages the multicluster  setup for Linkerd",
+		Args:  cobra.NoArgs,
+	}
+
+	createCredentalsCommand := &cobra.Command{
+		Use: "create-credentials", Short: "Create the necessary credentialsfor service mirroring",
```
alpeb: 2 typos: createCredentalsCommand and createCredentalsCommand

zaharidichev:
@alpeb That is correct. Maybe I should put these clarification details in the cmd description

cli/cmd/cluster.go
```diff
@@ -0,0 +1,208 @@
...
+			crbOut, err := yaml.Marshal(clusterRoleBinding)
+			if err != nil {
+				return err
+			}
+			fmt.Println(fmt.Sprintf("---\n%s---\n%s---\n%s", crOut, saOut, crbOut))
+			return nil
+		},
+	}
+
+	getCredentialsCmd := &cobra.Command{
```
grampelberg: I'm not sure I understand how get and create interact. Mind explaining that a little bit?

zaharidichev: Yes, so the `create` witll output a yaml with the service account and RBACs that allow for watching the remote cluster. This needs to be applied to the remote cluster. The get command simply extracts this information (the service account token) and outputs a config secret yaml. When this secret is `apply`-ed to the local cluster the service mirror picks it up automatically and uses the credentials to connect to the remote cluster.

cli/cmd/cluster.go
```diff
@@ -0,0 +1,208 @@
...
+					{
+						APIGroups: []string{""},
+						Resources: []string{"services"},
+						Verbs:     []string{"list", "get", "watch"},
+					},
+				},
+			}
+
+			svcAccount := corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{Name: createOpts.serviceAccount, Namespace: controlPlaneNamespace, Labels: labels},
```
alpeb: I think `create-credentials` should also expose a `--namespace` flag as `get-credentials` does, instead of defaulting to `controlPlaneNamespace` which might not be relevant in a (possible not meshed) remote cluster. Also WDYT about having `--namespace` _not_ defaulting to "linkerd", to avoid confusion, and force the user to specify something?

cli/cmd/cluster.go
grampelberg: Something like this then?

```
linkerd --context=my-remote-cluster \
    cluster create-credentials name-of-my-local-cluster | \
    kubectl --context=my-remote-cluster apply -f -
linkerd --context=my-remote-cluster \
    cluster get-credentials | \
    kubectl apply -f -
```

cli/cmd/cluster.go
grampelberg: I think the bigger question is - where should the mirror and gateway live? There are some downsides to having it live in the `linkerd` namespace but it feels like there might be more by having it live by itself. Any ideas what the tradeoffs are? Naively, it feels like everything should go into `linkerd` namespace and not be configurable, but I'm sure I'm missing something.

zaharidichev:
@grampelberg There should be tests! I will add these once we agree on everything else. Wrt to the namespace, I think we can agree to add all of this into the linkerd namespace. This feels like a natural part of Linkerd so no need to differentiate (or at least no obvious to me reason). That however brings the question of where we should install the service mirroring controller. Right now its a separate helm chart with a seprate namespace. Should we leave it like that. Note that, it is perfectly fine for the mirroring controller to live in a separate ns than the credentials. Yet.. it feels a bit odd. If we follow the practice for the cni plugin, we probably want a separate namespace. Not sure.

grampelberg:
> Right now its a separate helm chart with a seprate namespace. 

It should be a separate helm chart for sure. I don't see why we should stick it in a separate namespace though. At least, why don't we assume it is going into the linkerd control plane namespace for now and change it if someone complains.

My biggest piece of feedback is the naming, I just couldn't figure out what get/create were doing. I'm still trying to figure out what the right names should be though, not quite gotten my head around what the intent is there yet.

zaharidichev:
@grampelberg @alpeb @adleong Okey so do we agree that everything goes into the control plane namespace (credentials, secrets, as well as the service mirror)?

My only argument is that we recently moved the CNI plugin into its own namespace. Why being inconsistent here?

grampelberg:
@zaharidichev I don't remember the context on that one. My preference is to keep things simple here until it doesn't make sense anymore.

zaharidichev:
@grampelberg I think best things to do is what @alpeb suggested [right here ](https://github.com/linkerd/linkerd2/pull/4028#discussion_r382687454) and just leave the namespace out of the equation. People can specify their own namespace with via `kubectl`

grampelberg:
@zaharidichev I was more concerned with all the flags in this PR around namespaces. I think that's a reasonable statement, but what are the implications here?

zaharidichev:
@grampelberg The implications are that we do not specify anything namespace related. The commands simply output yamls. People will decide where to put these resources via their `kubectl`. Wouldn't that make sense?  And then.. there will be no need for namespace related flags in this PR.

grampelberg:
Let's do it, that sounds reasonable to me.

zaharidichev:
@grampelberg @alpeb Because the `ClusterRoleBinding` needs to refer to a namespaced `ServiceAccount` in its `subject`, I have provided the `service-account-namespace` option which defaults to "default". I have not added a namespace to the secret that is output via the `get-credentials` command and users can add that wherever they want. 

Wrt to these commands being in seprate files, I do not think this is necessary. They are sufficiently small/simple to live in one file, much like he different stags of `check` for example (check config, et )

Wrt to tests, lets just agree that this looks gooas it is and I will write the tests along with golden files.

grampelberg:
> @grampelberg @alpeb Because the ClusterRoleBinding needs to refer to a namespaced ServiceAccount in its subject, I have provided the service-account-namespace option which defaults to "default". I have not added a namespace to the secret that is output via the get-credentials command and users can add that wherever they want.

This is fine for now, I definitely want to change it though. My preference is still to tie it to the linkerd namespace. `default` is definitely the wrong setting.

cli/cmd/cluster.go
```diff
@@ -0,0 +1,284 @@
...
+
+func newCmdCluster() *cobra.Command {
+
+	getOpts := getCredentialsOptions{}
+	createOpts := createOptions{}
+	exportOpts := exportServiceOptions{}
+
+	clusterCmd := &cobra.Command{
+		Hidden: true,
+		Use:    "cluster manages the multicluster  setup for Linkerd",
```
adleong: nit: extra space

zaharidichev: I agree here. Also some more validation and testing are always better but at this point I think this is sufficient to provide the basic functionality.

zaharidichev:
@alpeb added more useful description + examples Feel free to take another look before I merge

PR #4072, opened 2020-02-19
Support for configuring service profile timeouts (x-linkerd-timeout) via openapi spec

lewiscowper:
This change is in a similar vein to #4052 (and issue #2073 which provided support for
configuring service profile retries via a vendor extension of
`x-linkerd-retryable`, when generating from an openapi specification.

This change is very similar to the final version of that pull request,
and adds a timeout value based on `x-linkerd-timeout`.

At this point I believe that if the timeout is not specified then the
default provided by linkerd of 30s will apply anyway, but won't
explicitly be reflected in the service profile, which I'm somewhat okay
with as a current state, but I think there's a potential future
improvement that the default timeout is always shown when generating
from an open api spec, but that's more to make it clear and obvious that
that timeout exists.

### DIFF
--- a/pkg/profiles/openapi.go
+++ b/pkg/profiles/openapi.go
@@ -17,6 +17,7 @@ import (
 
 const (
 	xLinkerdRetryable = "x-linkerd-retryable"
+	xLinkerdTimeout   = "x-linkerd-timeout"
 )
 
 var pathParamRegex = regexp.MustCompile(`\\{[^\}]*\\}`)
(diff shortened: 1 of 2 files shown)


pkg/profiles/openapi.go
```diff
@@ -110,16 +111,19 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
...
 		retryable, _ = operation.VendorExtensible.Extensions.GetBool(xLinkerdRetryable)
+		timeout, _ = operation.VendorExtensible.Extensions.GetString(xLinkerdTimeout)
 		responses = operation.Responses
 	}
 	return &sp.RouteSpec{
 		Name:            fmt.Sprintf("%s %s", method, path),
 		Condition:       toReqMatch(pathRegex, method),
 		ResponseClasses: toRspClasses(responses),
 		IsRetryable:     retryable,
+    Timeout:         timeout,
```
christianhuening: no idea whether the indentation gets caught by a linter ;-)

lewiscowper: huh, it looks fine in my vim. Will fix.

lewiscowper:
I force pushed to rebase my commits into one and go lint seems to be failing with a weird error that seems unrelated to this PR -- `Deadline exceeded: try increase it by passing --deadline option`. Can it be restarted?

alpeb:
It appeared to be a glitch. I triggered it again and it's green now :+1:

PR #4084, opened 2020-02-20
Exclude changes on markup files to trigger CI runs in master

### LINKED ISSUE #4082: Exclude changes on markup files to trigger CI runs in master
## Motivation

Currently when the `CHANGES.md` file gets merged into master, it triggers the Unit Tests, Static Checks and KinD Integration CI workflows. The release guide instructs to manually cancel those workflow runs before tagging the release.

## Solution

Avoid that extra step by adding a trigger exception into all the workflows, so they don't get run when only `*.md` files change, thus also avoiding CI runs for other markup file changes that don't modify linkerd's behavior. This should only affect runs on master; the workflows should still run on PRs.

alpeb:
Fixes #4082

This tested fine in a fork, under various scenarios combining:
- Modify markup file in root dir
- Modify markup file in subdir
- Modify non-markup file
- In master
- In a PR

### DIFF
--- a/.github/workflows/cloud_integration.yml
+++ b/.github/workflows/cloud_integration.yml
@@ -1,6 +1,9 @@
 name: Cloud integration
 on:
   push:
+    paths-ignore:
+    - '*.md'
+    - '**/*.md'
     branches:
     - master
     tags:
--- a/.github/workflows/kind_integration.yml
+++ b/.github/workflows/kind_integration.yml
@@ -2,6 +2,9 @@ name: KinD integration
 on:
   pull_request: {}
   push:
+    paths-ignore:
+    - '*.md'
+    - '**/*.md'
     branches:
     - master
     tags:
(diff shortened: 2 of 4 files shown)


.github/workflows/cloud_integration.yml
```diff
@@ -1,6 +1,9 @@
 name: Cloud integration
 on:
   push:
+    paths-ignore:
+    - '*.md'
```
kleimkuhler: I think `*.md` is redundant since `**/*.md` searches all files as well as directories and subdirectories. This may be GH Actions specific though?

alpeb: Turns out, according to my tests, `**/*.md` doesn't catch files in the root dir so both conditions are required :man_shrugging:

kleimkuhler: Interesting!

PR #4090, opened 2020-02-20
Add minimal release workflow

kleimkuhler:
## Motivation

A release workflow will be the only triggered workflow on `push.tags` events.

As a first step in automating the release process, it should assert that
integration tests pass once the docker images have been tagged.

Both KinD and cloud integration tests should run since they have different
sets of integration tests that they are responsible for running.

It then needs to run the `chart_deploy` job.

## Testing

This has been fully tested with a release tag push on my fork. The run can be
found [here](https://github.com/kleimkuhler/linkerd2/actions/runs/42664128)

It properly failed on `chart_deploy` because I did not want to push a test tag
helm chart.

## Solution

This workflow will:

- Build the docker images on the Packet Host
- Tag the docker images with the release tag
- Run KinD integration tests
- Run cloud integration tests
- Run `chart_deploy`

### DIFF
--- a/.github/workflows/cloud_integration.yml
+++ b/.github/workflows/cloud_integration.yml
@@ -3,8 +3,6 @@ on:
   push:
     branches:
     - master
-    tags:
-    - "*"
 jobs:
   docker_build:
     name: Docker build
@@ -72,6 +70,13 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
+    - name: Try to load cached Go modules
+      uses: actions/cache@v1
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
     - name: Install linkerd CLI
       id: install_cli
       run: |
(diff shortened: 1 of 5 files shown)


kleimkuhler:
Removed the dependency on #4082, but I don't see there being conflicts between the two.

.github/workflows/cloud_integration.yml
```diff
@@ -72,6 +70,13 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
+    - name: Try to load cached Go modules
+      uses: actions/cache@v1
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
```
alpeb: :+1:

.github/workflows/release.yml
```diff
@@ -0,0 +1,229 @@
...
+        ssh linkerd-docker docker version
+    - name: Push docker images to registry
+      env:
+        DOCKER_HOST: ssh://linkerd-docker
+      run: |
+        export PATH="`pwd`/bin:$PATH"
+        bin/docker-push-deps
+        bin/docker-push $TAG
+        bin/docker-retag-all $TAG master
+        bin/docker-push master
```
alpeb: These need to be removed

kleimkuhler: @alpeb Do they? Now that we don't run the other integration tests on `push` for markdown files, `master` tag would be behind the release tag and therefore not actually master.

alpeb: My bad, you're absolutely right!

kleimkuhler:
@siggy Good call thanks for the linking the issue. I made the suggested changes.

PR #4091, opened 2020-02-21
Add script to create release tag

kleimkuhler:
## Motivation

Creating a release tag is a manual process that is prone to error by the
release responsible member.

Additionally, the automated release project will require that a message is
included that is a copy of the recent `CHANGES.md` changes.

These steps can be scripted so that the member will just need to run a release
script.

## Solution

A `bin/create-release-tag` script will:
- Take a `$TAG` argument (maybe can remove this in the future) to use as the
  tag name
- Pull out the top section of `CHANGES.md` to use as the commit message
- Create the a tag with `$TAG` name and release changes as the message

## Example

```
$ TAG="edge-20.2.3"
$ bin/create-release-tag $TAG
$ git push $TAG
```

### DIFF
--- a/bin/create-release-tag
+++ b/bin/create-release-tag
@@ -40,4 +40,4 @@ tmp=$(mktemp -t release-commit-message.XXX.txt)
 awk '/^## (edge|stable)-[0-9]+\.[0-9]+\.[0-9]+/{n++} n==1' "$rootdir"/CHANGES.md > "$tmp"
 
 # Create an unsigned tag with the commit message.
-git tag -F "$tmp" "$new_linkerd_version"
+git tag -s -F "$tmp" "$new_linkerd_version"
--- a/bin/create-release-tag
+++ b/bin/create-release-tag
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+if [[ $# -ne 1 ]]; then
+    echo "usage: ${0##*/} (edge|stable)-xx.xx.xx" >&2
+    exit 1
+fi
+
+new_linkerd_version="$1"
+
+# Verify the tag format
+tag_format="^(edge|stable)-([0-9]+)\.([0-9]+)\.([0-9]+)$"
+if [[ $new_linkerd_version =~ $tag_format ]]; then
+    # todo: Use these values to verify the tag version increment.
+    _release_channel="${BASH_REMATCH[1]}"
+    _release_major="${BASH_REMATCH[2]}"
+    _release_minor="${BASH_REMATCH[3]}"
+    _release_patch="${BASH_REMATCH[4]}"
+else
+    echo "tag format incorrect; expected: $tag_format"
+    echo "example: edge-20.12.2, stable-2.10.1"
+    exit 1
+fi
+
+# todo: Verify the tag version increment.
+
+rootdir=$( cd "${0%/*}"/.. && pwd )
+
+# Make temporary file to save the release commit message into.
+tmp=$(mktemp -t release-commit-message.XXX.txt)
+echo $tmp
+
+# Save commit message into temporary file.
+#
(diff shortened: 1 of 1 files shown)


bin/create-release-tag
```diff
@@ -0,0 +1,44 @@
...
+
+# Verify the tag format
+tag_format="^(test|edge|stable)-([0-9]+)\.([0-9]+)\.([0-9]+)$"
+if [[ $new_linkerd_version =~ $tag_format ]]; then
+    # todo: Use these values to verify the tag version increment.
+    # release_channel="${BASH_REMATCH[1]}"
+    # release_major="${BASH_REMATCH[2]}"
+    # release_minor="${BASH_REMATCH[3]}"
+    # release_patch="${BASH_REMATCH[4]}"
+    :
```
kleimkuhler: This is a null command since the above variables are not set. I kept this structure mainly because verifying the version increment will use these variables, but there cannot be an empty `then`.

bin/create-release-tag
```diff
@@ -0,0 +1,44 @@
...
+echo $tmp
+
+# Save commit message into temporary file.
+#
+# Match each occurence of the regex and increment `n` by 1. While n == 1
+# (which is true only for the first section) print that line of `CHANGES.md`.
+# This ends up being the first section of release changes.
+awk '/^## (edge|stable)-[0-9]+\.[0-9]+\.[0-9]+/{n++} n==1' $rootdir/CHANGES.md > $tmp
+
+# Create an unsigned tag with the commit message.
```
kleimkuhler: As said by the comment, this is creating an unsigned tag object right now. `-s` requires the user to have GPG setup, and nothing about the release workflow currently relies on tags being signed.

bin/create-release-tag
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
```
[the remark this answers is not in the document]
kleimkuhler: @alpeb Just defaulted to being explicit instead of matching the other scripts. Can change them if it's a blocker.

bin/create-release-tag
```diff
@@ -0,0 +1,44 @@
...
+    exit 1
+fi
+
+# todo: Verify the tag version increment.
+
+rootdir=$( cd "${0%/*}"/.. && pwd )
+
+# Make temporary file to save the release commit message into.
+tmp=$(mktemp -t release-commit-message.XXX.txt)
+echo $tmp
```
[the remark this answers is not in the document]
kleimkuhler: Yes I did not mean to include this. This was from testing.

bin/create-release-tag
kleimkuhler: So signing a tag object with `-s` requires the user have GPG signing setup. I don't actually have this; I use the "signed-off-by" `-s` for commits but nothing for tags.

As part of this script merging, we could now require any member responsible for an edge or stable release needs setup GPG signing, but since we don't actually do anything with verifying the tag I thought that not be necessary to add right now.

So pushing unsigned tags is not changing anything about the way that tags are signed or not right now--they could be either. If we start using `-s` here then you need GPG signing setup.

bin/create-release-tag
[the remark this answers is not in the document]
alpeb: Your call, not a blocker :relaxed:

bin/create-release-tag
kleimkuhler: Okay after discussion this morning, `-s` is now added to the `git tag` command. Members responsible for releases will be responsible for being able to GPG sign the newly created tags.

kleimkuhler:
Shoot I meant to merge #4090. I don't need to revert this, but @siggy I'd still like to add your suggestions. I'm going to have a follow-up PR that adds instructions for using this script.

kleimkuhler:
I opened changes for updating the release process in our release instructions document.

PR #4095, opened 2020-02-24
Add spacing in CallToAction banner

### LINKED ISSUE #3690: Small spacing issue with the CallToAction banner on the control plane dashboard
When I install the linkerd control plane in my cluster, and open the dashboard before injecting any of my pods, the "Control Plane" page shows me a call to action that looks like this:

image

The banner is sitting flush with the headers below it, but I suspect it would be better to space it out a bit.

aliariff:
The call to action banner in the control plane page is missing some spacing.

The CSS is defined but not yet used. So the solution is to add the class name to the corresponding banner.

After its merged, the banner will have more space.

Before:
image

After:
image

Fixes #3690

### DIFF
--- a/web/app/js/components/CallToAction.jsx
+++ b/web/app/js/components/CallToAction.jsx
@@ -9,12 +9,6 @@ import { incompleteMeshMessage } from './util/CopyUtils.jsx';
 import { withStyles } from '@material-ui/core/styles';
 
 const styles = theme => ({
-  root: {
-    width: '90%',
-  },
-  button: {
-    marginRight: theme.spacing(1),
-  },
   instructions: {
     marginTop: theme.spacing(1),
     marginBottom: theme.spacing(1),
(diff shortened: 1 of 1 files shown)


Pothulapati:
This looks good, but looks like the tests were a bit flaky. Try amending the commit, so that the CI Checks are re-run?

aliariff:
@Pothulapati CI passed now.

web/app/js/components/CallToAction.jsx
```diff
@@ -38,6 +38,7 @@ const CallToAction = ({ resource, numResources }) => {
       <Typography>The service mesh was successfully installed!</Typography>
       <Stepper
         activeStep={lastStep}
+        className={classes.instructions}
```
scottcarol: This is a great fix, thanks! It looks like there are still two un-used `styles` specifications, `root` and `button`. Would you mind removing those from the component as well before we merge this in?

aliariff: Yes sure

PR #4098, opened 2020-02-25
Generate correct path regex for proto files without package name

amariampolskiy:
**Problem**

When generating linkerd service profile based on proto files, default path regex looks like this:

`/Package\.Service/Method`

When proto file is missing package name it looks like this:

`/\.Service/Method`

What it should look like:

`/Service/Method`

**Solution**

Check package name, change path regex accordingly

**Validation**

Generate service profiles based on proto files without package name.

### DIFF
--- a/pkg/profiles/proto.go
+++ b/pkg/profiles/proto.go
@@ -45,11 +45,18 @@ func protoToServiceProfile(parser *proto.Parser, namespace, name, clusterDomain 
 			pkg = typed.Name
 		case *proto.RPC:
 			if service, ok := typed.Parent.(*proto.Service); ok {
+				var path string
+				switch pkg {
+				case "":
+					path = fmt.Sprintf("/%s/%s", service.Name, typed.Name)
+				default:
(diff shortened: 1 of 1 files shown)


grampelberg:
TIL! I didn't realize you could have protos without package names.

grampelberg:
TIL! I didn't realize you could have protos without package names.

grampelberg:
TIL! I didn't realize you could have protos without package names.

grampelberg:
TIL! I didn't realize you could have protos without package names.

amariampolskiy:
Yeah, it doesn't seem like a package name is necessary for correct generation of protobuf libs.

alpeb:
Thanks @amariampolskiy  for your first contribution to Linkerd! 1st_place_medal
If you'd like to receive some Linkerd swag, I invite you to fill this [form](https://docs.google.com/forms/d/e/1FAIpQLSdXA7kAl6oyUQ4DFRjiDc2biWlvkhNirrMwsXaQ8V3Uh8Yj5A/viewform) :wink:

PR #4103, opened 2020-02-26
Service Mirroring Component (no queue)

zaharidichev:
This is a version that removes the queue and leaves it to the func handlers of the informers to call the corresponding methods synchronously. I am no convinced that this simplifies things all that much. Futrthermore: 

- Its a universal recommendation that the funcs in the handlers should be returning rather quickly, while now we start calling k8s api methods on our local and remote clusters. 
-  I was hoping that the queue approah can give us a bit more _principled_ error handling, where we can requeue and try again instead of passivelly log things.
- If we encaunter some kind of intermittent failures with processing our updates, we need to wait till the next resync (currently configured at 10 min) and hope that we can effectivelly `retry` at that point
- Not sure this approach lowers the cognitive burdon of following the code. 

Note: I have not updated the comment to reflect we are not using a queue approach. If we decide to merge that one instead, I can do that.

### DIFF
--- a/charts/linkerd2-service-mirror/README.md
+++ b/charts/linkerd2-service-mirror/README.md
@@ -18,4 +18,3 @@ The following table lists the configurable parameters of the linkerd2-service-mi
 |`namespace`                           | Service Mirror component namespace                                                |`linkerd-service-mirror`|
 |`serviceMirrorUID`                    | User id under which the Service Mirror shall be ran                               |`2103`|
 |`logLevel`                            | Log level for the Service Mirror component                                        |`info`|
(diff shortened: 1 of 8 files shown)


adleong:
Thanks for investigating this, @zaharidichev.  This diff actually looks really good to me.  The logic and flow seems to be exactly the same, but with the queue removed from the middle.  The fact that this is a net deletion of 237 lines is a good sign.

With that said, if you still feel that the queue is beneficial then I'm okay with keeping it.  I wasn't aware that informer handlers were required to return quickly.  If that's the case, it's a good argument in favor of the queue.

Feel free to close this and proceed on the main PR.

PR #4104, opened 2020-02-26
Detect changes in addresses when getting updates in endpoints watcher

zaharidichev:
While digging trhough that code to try and understand it better, I came across this TODO that seemed like a sensible fix. @adleong let me know whether that makes sense as you wrote most of that code. I tried to use the existing tests to try and write a unit test, but injecting state into the informers prooved a bit tricky tbh (like the UpdateFunc was not called). It might be fake client set.

### DIFF
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -601,15 +601,33 @@ func getTargetPort(service *corev1.Service, port Port) namedPort {
 	return targetPort
 }
 
+func addressChanged(oldAddress *Address, newAddress *Address) bool {
+	if oldAddress.Pod !=  nil && newAddress.Pod != nil {
+		// if these addresses are owned by pods we can check the resource versions
+		return oldAddress.Pod.ResourceVersion != newAddress.Pod.ResourceVersion
+	} else {
+		// in this case the identity could have changed; this can happen when for
+		// example a mirrored service is reassigned to a new gateway with a different
+		// identity and the service mirroring controller picks that and updates the
+		// identity
+		return oldAddress.Identity != newAddress.Identity
+	}
+}
+
 func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSet) {
 	// TODO: this detects pods which have been added or removed, but does not
 	// detect addresses which have been modified.  A modified address should trigger
 	// an add of the new version.
 	addAddesses := make(map[ID]Address)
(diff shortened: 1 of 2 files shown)


adleong:
I think you might be able to test this by calling `addEndpoints` directly on the watcher, similarly to how `deleteEndpoints` is called in the tests currently.

This change looks good to me, though it's worth noting that this won't have any change in behavior right now.  The data returned by the destination API about a pod is all based on immutable data (such as owner, identity, address, hash, etc) which will never change on a running pod.  The context here is that we had planned to make addr weight based on a pod annotation which could be updated on a running pod.  But that has not yet been implemented.

zaharidichev:
@adleong for some reason I was under the wrong impression that I cannot call this method as its private but now I see this i actually private to the package #GoLangNoob.... Anyway, does it make sense to move forward with this change nevertheless? If so, I can add the tests

adleong:
🤷‍♂ I think this change is correct, but not urgent.

zaharidichev:
@adleong I added some tests Let me know whether that looks correct to you. Up to you whether we want that in.

zaharidichev:
@alpeb @adleong I prefer to wait for #4160 to merge before I rebase this against master and merge as I think there will be some additional chnges needed.

zaharidichev:
@adleong @alpeb can you take another look. I have updated the code to reflect the possiblity of a remote service being reassigned to a new gateway with different identity. This  scenario is actually a lot more possible than what the scenario which triggered this initial PR.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -601,15 +601,33 @@ func getTargetPort(service *corev1.Service, port Port) namedPort {
...
+func addressChanged(oldAddress Address, newAddress Address) bool {
+	if oldAddress.Pod != nil && newAddress.Pod != nil {
+		// if these addresses are owned by pods we can check the resource versions
+		return oldAddress.Pod.ResourceVersion != newAddress.Pod.ResourceVersion
+	}
+	// in this case the identity could have changed; this can happen when for
+	// example a mirrored service is reassigned to a new gateway with a different
+	// identity and the service mirroring controller picks that and updates the
+	// identity
+	return oldAddress.Identity != newAddress.Identity
```
adleong, line 613: Should we move this return up to before the previous one?  Currently, identity overrides only exist for mirror services, but theoretically we could have identity overrides for pod-backed services too.  In this case, we would want to return true if the identity has changed, even if the pod has not.

PR #4100, opened 2020-02-25
Create linkerd/website dispatch event on release

kleimkuhler:
## Motivation

A goal of the release automation project is to automate the website publish
that publishes a new install script that uses the new release version.

linkerd/website#668 removed the hard coded versions from the repo and moved
the version update into the `make publish` command.

That workflow now needs to be triggered by a release in `linkerd2`.

## Solution

Once `kind_integration_tests` and `cloud_integration_test` pass, a job runs
using the [repository-dispatch](https://github.com/marketplace/actions/repository-dispatch) action to create a repository dispatch event in
`linkerd/website`.

This dispatch event will (current PR: linkerd/website#670) trigger the
publish workflow.

## Testing

Tested in my fork [here](https://github.com/kleimkuhler/linkerd2/actions/runs/45165789)

## Additional steps needed

A new `RELEASE_TOKEN` secret needs to be added to this repo. It should be the
personal access token of [l5d-bot](https://github.com/l5d-bot) with `repo` access.

Signed-off-by: Kevin Leimkuhler <kevin@kleimkuhler.com>

PR #4106, opened 2020-02-26
Added last release job to verify version was published in the website

alpeb:
Added the `website_publish_check` job, triggered upon edge/stable tag pushes,
that verifies that the install script at the website has indeed been
updated with the corresponding edge/stable version. It performs the
check every 5 seconds, 10 times, before giving up and failing the build
run. Tested fine in my fork :thumbsup:

This is dependent on the `website_publishwebsite_publish` job being created
in #4100 (thus the dependency failure, till #4100 gets merged).


.github/workflows/release.yml
```diff
@@ -237,6 +237,35 @@ jobs:
           ./target/release/linkerd2-cli-*-linux.sha256
           ./target/release/linkerd2-cli-*-windows.exe
           ./target/release/linkerd2-cli-*-windows.exe.sha256
+  website_publish_check:
+    name: Linkerd website publish check
+    if: startsWith(github.ref, 'refs/tags/stable') || startsWith(github.ref, 'refs/tags/edge')
```
siggy: (un)related to this PR, can we assume that this entire workflow is for edge/stable releases, and then avoid these `startsWith` lines, instead filtering at the top, something like:
```yaml
name: Release
on:
  push:
    tags:
    - "edge-*"
    - "stable-*"
```

kleimkuhler: @siggy I was thinking about this and I feel like it's not something we need to change right now. There may some situation where we'd like to publish a tag that is not edge or release related--maybe it's just for testing something.

It would still be helpful to have this workflow run so that the images are pushed to the linkerd registry after being fully tested, but not attempt to create a new GH release and update the website.

That's all a hypothetical though, so maybe it makes sense!

PR #4107, opened 2020-02-26
Add integration test to validate kubectl applies updated config

supra08:
Kubectl's dropping of config objects was tackled in [this](https://github.com/linkerd/linkerd2/pull/4061) PR. An integration test was required to ensure its effect.

An integration test which compares a known kubectl apply config with generated one on updating config with custom value is (here it is `foo: bar`) is implemented

```
go test -v ./test/upgrade -integration-tests -linkerd `pwd`/bin/linkerd 
```
gives all tests passed. It validates its success

Signed-off-by: Supratik Das <rick.das08@gmail.com>


supra08:
@ihcsim @zaharidichev please review. 
I couldn't figure out the failing test, maybe needing a little help in that.
Apart from this I could not add reviewers and the `GSoC` label, because those fields were locked.

supra08:
Yeah I get it. But the problem is `kubectl` will generate timestamps in the output. In that case, we cannot compare that to the standard premade output.
This test was just to ensure that `kubectl` is not dropping the namespace label. Some maybe just checking that the namespace config has the `linkerd.io/control-plane-ns=linkerd` is enough.

zaharidichev:
@supra08 at the moment your test is simply checking whether the label is there or not (after upgrade), right. Why dont you just do that. You kubectl got get only the labels and compare that instead of the whole object.

supra08:
Yeah sure! That would be better I guess.

PR #4111, opened 2020-02-27
Add success message for tag script

kleimkuhler:
This adds a message after running the `create-release-script` that I intended to
add as part of the initial PR. Example output:

```
❯ bin/create-release-tag $TAG tag created and signed.

tag: edge-93.1.1

To push tag, run:
    git push origin edge-93.1.1
```


alpeb:
The failed integration test should have been fixed in #4108 which has just been merged, so I'm retriggering this build :prayer_beads:

PR #4112, opened 2020-02-27
Fix `fetch-proxy` script on macos

kleimkuhler:
`sha256sum` is not installed by default. Use `openssl dgst -sha256` instead.


kleimkuhler:
@adleong Ah yea I could change it to `shasum -a 256` and then use that for all platforms.

kleimkuhler:
@adleong Unfortunately `shasum` it is not installed by default on Ubuntu [1](https://github.com/linkerd/linkerd2/pull/4112/checks?check_run_id=473731562). Changing back to the initial change.

PR #4115, opened 2020-02-27
Service mirroring tests

zaharidichev:
Unit tests that exercise most of the code in cluster_watcher.go. Essentially the whole  cluster mirroring machinary can be tought of as a function that takes remote cluster state, local cluster state, and modification events and as a result it either modifies local cluster state or issues new events onto the queue. This is what these tests are trying to model. I think this covers a lot of the logic there. Any suggestions for other edge cases are welcome.

### DIFF
--- a/controller/cmd/service-mirror/cluster_watcher_test.go
+++ b/controller/cmd/service-mirror/cluster_watcher_test.go
@@ -66,25 +66,49 @@ func runTestCase(tc *testCase, t *testing.T) {
 		localAPI.Sync(nil)
 		remoteAPI.Sync(nil)
 
-		for _, expected := range tc.expectedLocalServices {
-			actual, err := localAPI.Client.CoreV1().Services(expected.Namespace).Get(expected.Name, metav1.GetOptions{})
+		if tc.expectedLocalServices == nil {
+			// ensure the are no local services
+			services, err := localAPI.Client.CoreV1().Services(corev1.NamespaceAll).List(metav1.ListOptions{})
 			if err != nil {
-				t.Fatalf("Could not find mirrored service with name %s", expected.Name)
-			}
-
-			if err := diffServices(expected, actual); err != nil {
 				t.Fatal(err)
+			}
+			if len(services.Items) > 0 {
+				t.Fatalf("Was expecting no local services but instead found %v", services.Items)
+
+			}
+		} else {
+			for _, expected := range tc.expectedLocalServices {
+				actual, err := localAPI.Client.CoreV1().Services(expected.Namespace).Get(expected.Name, metav1.GetOptions{})
+				if err != nil {
+					t.Fatalf("Could not find mirrored service with name %s", expected.Name)
+				}
+
+				if err := diffServices(expected, actual); err != nil {
+					t.Fatal(err)
+				}
 			}
 		}
 
-		for _, expected := range tc.expectedLocalEndpoints {
-			actual, err := localAPI.Client.CoreV1().Endpoints(expected.Namespace).Get(expected.Name, metav1.GetOptions{})
+		if tc.expectedLocalEndpoints == nil {
+			// ensure the are no local endpoints
+			endpoints, err := localAPI.Client.CoreV1().Endpoints(corev1.NamespaceAll).List(metav1.ListOptions{})
 			if err != nil {
-				t.Fatalf("Could not find endpoints with name %s", expected.Name)
-			}
-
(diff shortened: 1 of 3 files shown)


controller/cmd/service-mirror/cluster_watcher_test.go
```diff
@@ -0,0 +1,1165 @@
...
+	}
+
+	if !reflect.DeepEqual(expected.Subsets, actual.Subsets) {
+		return fmt.Errorf("was expecting endpoints with subsets %v but got %v", expected.Subsets, actual.Subsets)
+	}
+
+	return nil
+}
+
+func getRemoteService(name, namespace, gtwName, gtwNs, resourceVersion string, ports []corev1.ServicePort) *corev1.Service {
```
adleong: since this isn't actually fetching anything from anywhere, should we call this `makeRemoteService` or even simply `remoteService`?

controller/cmd/service-mirror/cluster_watcher_test.go
```diff
@@ -0,0 +1,1165 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            name,
+			Namespace:       namespace,
+			ResourceVersion: resourceVersion,
+			Annotations:     annotations,
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: ports,
+		},
+		Status: corev1.ServiceStatus{},
```
adleong: Since this is the zero value, I think we can omit it.

controller/cmd/service-mirror/cluster_watcher_test.go
adleong: similar comment for other "get" functions in this file.

controller/cmd/service-mirror/cluster_watcher_test.go
```diff
@@ -0,0 +1,1165 @@
...
+func getEndpointsAsYaml(name, namespace, gtwName, gtwNs, gatewayIP string, ports []corev1.EndpointPort, t *testing.T) string {
+	ep := getEndpoints(name, namespace, gtwName, gtwNs, gatewayIP, ports)
+
+	bytes, err := yaml.Marshal(ep)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return fmt.Sprintf("%s", bytes)
+}
+
```
adleong: I might consider moving all the helper methods (everything above this line, basically) into another file so that the test cases themselves are more front-and-center in this file.  TIOLI

controller/cmd/service-mirror/cluster_watcher_test.go
```diff
@@ -0,0 +1,960 @@
...
+					Name:      "test-service-remote-to-delete",
+					Namespace: "test-namespace-to-delete",
+				},
+			},
+
+			localResources: []string{
+				mirroredServiceAsYaml("test-service-remote-to-delete", "test-namespace-to-delete", "", "", "", "", nil, t),
+				endpointsAsYaml("test-service-remote-to-delete", "test-namespace-to-delete", "", "", "", nil, t),
+			},
+		},
```
alpeb: Given `expectedLocalServices` and `expectedLocalEndpoints` is empty in this test, the loops in `runTestCase()` won't run and there is no check performed verifying if that service and endpoint are really gone. 
One way of verifying could be checking if those expected slices are empty, and if so, loop against `localResources` and verify the local API can't retrieve any service/endpoint from that slice...

controller/cmd/service-mirror/cluster_watcher_test.go
```diff
@@ -0,0 +1,960 @@
...
+			testDescription: "unregisters cluster and cleans up all mirrored resources",
+			events: []interface{}{
+				&ClusterUnregistered{},
+			},
+			localResources: []string{
+				mirroredServiceAsYaml("test-service-1-remote", "test-namespace", "", "", "", "", nil, t),
+				endpointsAsYaml("test-service-1-remote", "test-namespace", "", "", "", nil, t),
+				mirroredServiceAsYaml("test-service-2-remote", "test-namespace", "", "", "", "", nil, t),
+				endpointsAsYaml("test-service-2-remote", "test-namespace", "", "", "", nil, t),
+			},
```
alpeb: Same here, it appears no check is performed.

controller/cmd/service-mirror/cluster_watcher_test.go
zaharidichev: Thanks for catching that @alpeb !

PR #4118, opened 2020-02-28
Introduce `bin/shellcheck`, add to ci

siggy:
PR #4117 was root-caused with the help of `shellcheck`.

This change introduces a `bin/shellcheck` script, and adds it to CI. In
CI, many checks are disabled to allow it to pass. This will at least
prevent introduction of new classes of shell issue, and should motivate
re-enabling more checks over time.

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -97,7 +97,6 @@ jobs:
         ! -name docker-push-deps \
         ! -name docker-retag-all \
         ! -name _docker.sh \
-        ! -name fetch-proxy \
         ! -name fmt \
         ! -name helm-build \
         ! -name kind-load \
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -68,38 +68,52 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
     - name: shellcheck
-      # TODO: fix our shell scripts and re-enable checks
+      # TODO: Each file listed here is excluded from shellcheck because it
+      # fails. As we fix files we can remove them from this list.
+      #
+      # Once we have this list paired down signficantly, we can switch to
+      # disabling specific checks across all files, for example:
+      # bin/shellcheck bin/* --exclude=SC1000
+      #
+      # For more information on shellcheck failures:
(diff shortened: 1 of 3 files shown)


.github/workflows/static_checks.yml
```diff
@@ -58,3 +58,45 @@ jobs:
...
+  shellcheck:
+    name: shellcheck
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: shellcheck
+      # TODO: fix our shell scripts and re-enable checks
+      run: |
+        bin/shellcheck bin/* --exclude="\
```
adleong: Without doing a bunch of research, I'm not sure how to approach or understand this list of exclusions.  I assume that some of these are excluded because we would like to fix them, but don't want to do it right now.  Whereas perhaps others are excluded because we don't agree with the rule.  

I'm not sure how, but it would be great if this list of exclusions was more understandable somehow.  If a contributor wanted to help burn this down, it's now clear where to start.

adleong: Would it make sense to exclude files instead of excluding rules?  That would ensure that all new scripts are complaint with all rules and give us perhaps an easier "TODO list" of existing files to fix up.

siggy: This list of exclusions is exactly what's necessary to make `shellcheck` pass, I didn't evaluate any of the rules based on whether we _should_ use them.

I like the idea of excluding files instead, let me take a look at that.

.github/workflows/static_checks.yml
```diff
@@ -58,3 +58,59 @@ jobs:
...
+        ! -name _tag.sh \
+        ! -name test-cleanup \
+        ! -name test-clouds-cleanup \
+        ! -name test-clouds \
+        ! -name test-run \
+        ! -name _test-run.sh \
+        ! -name test-scale \
+        ! -name update-go-deps-shas \
+        ! -name web \
+        | xargs -I {} bin/shellcheck {}
```
alpeb: You can run `shellcheck -x` to allow following external sources, which is used already in `bin/helm-build` and thus that file can be removed from this list.

Issue #3871, opened 2019-12-29
stat: Suport omitting unmeshed resources


olix0r:
I frequently want to monitor a cluster with a command like `watch linkerd stat po -A`; but this includes all unmeshed (kube-system, etc) resources as well. I don't really care about unmeshed resources when I'm using Liinkerd.  I'd prefer that we provide a flag to _opt-in_ to view unmeshed resources; but I'd live with a flag that let me opt out.

olix0r:
Alternatively, if we supported label selectors, this could be a simple label selector query.

mayankshah1607:
Does https://github.com/linkerd/linkerd2/pull/4040 fix this issue?

kleimkuhler:
@mayankshah1607 This is helpful for displaying meshed pods:
```
linkerd stat -A -l linkerd.io/control-plane-ns pods
```
Unfortunately it doesn't cover all cases such as deployments since they do not have a label that can be selected on. There is an inject annotation, but those cannot be selected on.

A possibility is that we just do not display rows that have a `0` meshed pod count. It would be fairly straightforward to implement and would cover any type of resource since `0/x` meshed pods would just not be displayed. An opt-in view would show all rows regardless of number of meshed pods.

PR #4120, opened 2020-02-28
Add kubectl-style label selectors

mayankshah1607:
Second take on https://github.com/linkerd/linkerd2/issues/2734
Also, somewhat related to https://github.com/linkerd/gsoc/pull/2 , https://github.com/linkerd/linkerd2/issues/3871
This PR adds support for kubectl-style label selectors for the following cmds:
- tap
- top
- routes

Example:
```bash
# tap deployments with label sample.label=test
$ linkerd -n myns tap deploy -l sample.label=test

# display live traffic from all deployments with label sample.label=test
$ linkerd -n myns top deploy --selector sample.label=test

# display route stat from all deployments with label sample.label=test
$ linkerd -n myns routes deploy -l sample.label=test
```
> Note : Label selector for stat has been addressed in https://github.com/linkerd/linkerd2/pull/4040

### Unresolved Questions

- Do we need to add similar support for `metrics` cmd? It looks like the same can be achieved from : 
```bash
linkerd metrics -n linkerd $(
    kubectl --namespace linkerd get pod \
      --selector linkerd.io/control-plane-component=controller \
      --output name
  )
```

### TODO
A follow up PR shall be made for :
- Add support for `tap`,`top`,`routes`,`stat` to accept list of resources. Eg:
```bash
$ linkerd -n linkerd tap deploy linkerd-controller linkerd-identity
```

cc: @ihcsim

### DIFF
--- a/cli/cmd/tap.go
+++ b/cli/cmd/tap.go
@@ -32,6 +32,7 @@ type tapOptions struct {
 	authority   string
 	path        string
 	output      string
+	labelSelector string
 }
 
 type endpoint struct {
@@ -111,6 +112,7 @@ func newTapOptions() *tapOptions {
 		authority:   "",
 		path:        "",
 		output:      "",
+		labelSelector: "",
 	}
 }
 
@@ -179,6 +181,7 @@ func newCmdTap() *cobra.Command {
 				Authority:   options.authority,
 				Path:        options.path,
 				Extract:     options.output == jsonOutput,
+				LabelSelector: options.labelSelector,
 			}
 
 			err := options.validate()
@@ -218,6 +221,9 @@ func newCmdTap() *cobra.Command {
 		"Display requests with paths that start with this prefix")
 	cmd.PersistentFlags().StringVarP(&options.output, "output", "o", options.output,
 		fmt.Sprintf("Output format. One of: \"%s\", \"%s\"", wideOutput, jsonOutput))
+	cmd.PersistentFlags().StringVarP(&options.labelSelector, "selector", "l", options.labelSelector,
+		"Selector (label query) to filter on, supports '=', '==', and '!='")
+
 
 	return cmd
 }
--- a/cli/cmd/tap.go
+++ b/cli/cmd/tap.go
@@ -106,7 +106,7 @@ func newTapOptions() *tapOptions {
 		namespace:     "default",
 		toResource:    "",
 		toNamespace:   "",
-		maxRps:        100.0,
+		maxRps:        maxRps,
 		scheme:        "",
 		method:        "",
 		authority:     "",
--- a/cli/cmd/tap.go
+++ b/cli/cmd/tap.go
@@ -23,15 +23,15 @@ import (
 type renderTapEventFunc func(*pb.TapEvent, string) string
 
 type tapOptions struct {
-	namespace   string
-	toResource  string
-	toNamespace string
-	maxRps      float32
-	scheme      string
-	method      string
-	authority   string
-	path        string
-	output      string
+	namespace     string
+	toResource    string
+	toNamespace   string
+	maxRps        float32
+	scheme        string
+	method        string
+	authority     string
+	path          string
+	output        string
 	labelSelector string
 }
 
@@ -103,15 +103,15 @@ type tapEvent struct {
 
 func newTapOptions() *tapOptions {
 	return &tapOptions{
-		namespace:   "default",
-		toResource:  "",
-		toNamespace: "",
-		maxRps:      100.0,
-		scheme:      "",
-		method:      "",
-		authority:   "",
-		path:        "",
-		output:      "",
+		namespace:     "default",
+		toResource:    "",
+		toNamespace:   "",
+		maxRps:        100.0,
+		scheme:        "",
+		method:        "",
+		authority:     "",
+		path:          "",
+		output:        "",
 		labelSelector: "",
 	}
 }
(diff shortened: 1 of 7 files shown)


Pothulapati:
Hey,
Thanks for taking this up! Works Right!

Something to highlight,  For commands with `--to` flag like
` ./target/cli/linux/linkerd tap deploy -l linkerd.io/control-plane-component=web -n linkerd -ojson --to namespace/linkerd`, labelSelectors are not passed to the `destination` resource.

Another small update could be, For queries like, 
```bash
linkerd tap deploy -l linkerd.io/control-plane-component=web
Error: no pods found for deployment/
```
where there are no resources present for the given labels, i.e the resultant of `GetObjects` is empty, We could return right there with relevant error message.

mayankshah1607:
@Pothulapati thanks for pointing that out! I have updated the cmds to use the labels for destination resource.
However, returning when result of `GetObjects` is empty, is causing a lot of tests to fail, mainly because these tests were expecting some other output, and would not expect the method to necessarily return when `GetObjects` is empty. We could simply go ahead and change the expected output of these tests, but it looks like for some, it would change the meaning. WDYT?

Pothulapati:
@mayankshah1607 I actually don't think we should add the same `labelSelectors` to destination resources as most use cases wouldn't be doing {tap/top/routes} on source and destination with same labels. I was just trying to highlight that use-case. Maybe we should have a new flag? @alpeb could comment on this.

For tests where the expectation is an error, doing that should be okay but I see some tests like `FAIL: TestStatSummary/Successfully_queries_for_resource_type_'all' (0.14s)` Also failing.

alpeb:
From what I can see, the label selector that was implemented for `linkerd stat` also doesn't apply to the destination resource. Definitely it's worth a separate flag, but IMO it's best to keep this PR as minimal as possible and tackle that in a separate one.

mayankshah1607:
@alpeb @Pothulapati Makes sense to me. I'll open a separate issue where we can tackle the destination labelSelector - we can discuss the details there.

mayankshah1607:
@Pothulapati Just looked into the workings of the tests.
I don't think we should implement a return when `GetObjects` returns an empty slice. That is because tests such as `TestStatSummary/Successfully_queries_for_resource_type_'all'`  are expecting the error to be `nil` . This test must check if the all resources can be queried, but because the result is empty (which shouldn't have been a problem when the tests were return) it throws an error.
While simply changing the expected output for some tests may not be a problem, tests like the ones mentioned above would require some refactoring and may have its meaning changed. WDYT?

mayankshah1607:
@alpeb the reason why we have different label retrieving methods is that each of these accept different arguments - in case of `getLabelSelector` , the methods needs a `TapByResourceRequest` and in case of `getTopLabelSelector` we need `TopRoutesRequest`. We'd have to consider a few things before we can unify them. All of these methods are private. To make this reusable, where could we move it so that all the packages can access it? 
Is there a common interface being implemented by these ResourceRequest objects, so that we the common method can accept that as an argument and cast it into any of these ResourceRequest types?

alpeb:
You could put that function under in `/pkg/k8s/labels.go` for example. In there you could declare a new interface containing the `GetSelector()` method. That works for `StatSummaryRequest` and `TopRoutesRequest`. `TapByResourceRequest` on the other hand uses `Target` instead of `Selector`, so you'd have to rename that in the protobuf definition. But I think all that will make for a nice followup PR :ok_hand: 
Do you guys have any further comments on the `GetObjects()` discussion?

mayankshah1607:
@alpeb Thanks! Along with the adding support for resource list for `tap`,`top`, etc, I'll open a follow up PR for unifying the methods.

Also, regarding `GetObjects()`, it looks like there's no need for adding any extra conditions to check if no objects were returned, as its already being done in line 80-105.

Could you also have a look at the unresolved question in the description? I'll make changes accordingly before we can get this merged. :smile:

alpeb:
Oh indeed, `linkerd metrics` could also use a label selector :slightly_smiling_face:

mayankshah1607:
@alpeb Done! Let me know if this works. 
I had a question, though. It looks like `metrics` is built to compulsorily accept a resource name. For example `linkerd metrics deploy/web` would show the metrics from the `web` deployment. `linkerd metrics deploy` would not work as `metrics` can show the output for only 1 resource. In such a case, having label selector for `deploy` would be redundant.

So, this PR accepts label selectors which are applied at the pod level. In particular, the `getPodsFor()` method finally returns pods that includes the labels accepted from the `-l` flag.
Please let me know if this fits with the use case of the `metrics` command. :smile:

mayankshah1607:
It looks like the integration tests are failing. Not really sure why =(

alpeb:
Those are good insights :+1: 
It sounds to me having the selector only apply to pods can be confusing. The selector should apply to the same resource kind that was provided as a target in the command. But like you point out, that'd be redundant. How about lifting the requirement of providing the resource name, as long as a selector was provided? To provide an easier changelog when the next edge gets built, please do that as a separate PR, removing the routes changes from this one, and then we can merge this one.

mayankshah1607:
@alpeb Done! :)

cli/cmd/tap.go
```diff
@@ -102,15 +103,16 @@ type tapEvent struct {
...
-		maxRps:      100.0,
-		scheme:      "",
-		method:      "",
-		authority:   "",
-		path:        "",
-		output:      "",
+		namespace:     "default",
+		toResource:    "",
+		toNamespace:   "",
+		maxRps:        100.0,
```
zaharidichev: lets put that in a constant, if not too much trouble

cli/cmd/top.go
```diff
@@ -258,16 +259,17 @@ const (
...
-		scheme:      "",
-		method:      "",
-		authority:   "",
-		path:        "",
-		hideSources: false,
-		routes:      false,
+		namespace:     "default",
+		toResource:    "",
+		toNamespace:   "",
+		maxRps:        100.0,
```
zaharidichev: const

PR #4132, opened 2020-03-03
Snap package for Linkerd CLI

### LINKED ISSUE #4096: Snap package
## Feature Request

Linkerd already supports a brew package, why not a snap too?

dr0pdb:
### Subject
Building a snap package for the Linkerd CLI

### Validation
1. Installed Linkerd on a fresh Minikube cluster using the snapped CLI
2. Ran `linkerd check` to validate installation
output:
```
 srv_twry@archlinux  ~  linkerd check                       
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √

```
3. verified that Linkerd dashboard is working 

Fixes #4096 

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,22 +1,21 @@
 name: linkerd
 base: core18
-version: '0.1' # just for humans, typically '1.2+git' or '1.3.2'
+version: '2.7.0'
 summary: Command-line utility to interact with linkerd
 description: |
     Linkerd is a service mesh, designed to give platform-wide observability,
     reliability, and security without requiring configuration or code changes.
 
-grade: devel
-confinement: devmode
+grade: stable
+confinement: classic
 
 parts:
     linkerd:
         plugin: dump
         source: .
         override-pull: |
-            cd "$SNAPCRAFT_PART_SOURCE" # this line might be wrong. also maybe unneeded
-            curl -sSLO https://github.com/linkerd/linkerd2/releases/download/edge-20.3.1/linkerd2-cli-edge-20.3.1-linux
-            mv linkerd2-cli-edge-20.3.1-linux linkerd # rename binary to linkerd
+            curl -sSLO https://github.com/linkerd/linkerd2/releases/download/stable-2.7.0/linkerd2-cli-stable-2.7.0-linux
+            mv linkerd2-cli-stable-2.7.0-linux linkerd # rename binary to linkerd
             chmod +x linkerd
         build-packages:
             - curl
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,25 @@
+name: linkerd
+base: core18
+version: '0.1' # just for humans, typically '1.2+git' or '1.3.2'
+summary: Command-line utility to interact with linkerd
+description: |
+    Linkerd is a service mesh, designed to give platform-wide observability,
+    reliability, and security without requiring configuration or code changes.
+
+#icon: https://user-images.githubusercontent.com/9226/33582867-3e646e02-d90c-11e7-85a2-2e238737e859.png
+grade: devel
+confinement: devmode
+
+parts:
+    linkerd:
+        plugin: go
+        go-importpath: github.com/linkerd/linkerd2
+        source: .
+        source-type: git
+        source-tag: go1.13
+        build-packages:
+            - gcc
+
+apps:
+    linkerd:
+        command: bin/linkerd
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -13,12 +13,11 @@ parts:
     linkerd:
         plugin: dump
         source: .
-        override-pull: |
-            curl -sSLO https://github.com/linkerd/linkerd2/releases/download/stable-2.7.0/linkerd2-cli-stable-2.7.0-linux
-            mv linkerd2-cli-stable-2.7.0-linux linkerd # rename binary to linkerd
-            chmod +x linkerd
-        build-packages:
-            - curl
+        stage:
+            - target/release/linkerd2-cli-*-linux
+        override-stage: |
+            snapcraftctl stage
+            cp target/release/linkerd2-cli-*-linux $SNAPCRAFT_PRIME/linkerd
 apps:
     linkerd:
         command: linkerd
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,10 +16,10 @@ parts:
         go-importpath: github.com/linkerd/linkerd2
         source: .
         source-type: git
-        source-tag: go1.13
-        build-packages:
-            - gcc
-
+        override-build: |
+            cd /root/parts/linkerd/src/
+            ./bin/build-cli-bin
+            mv ./target/cli/linux/linkerd ~/prime/
 apps:
     linkerd:
-        command: bin/linkerd
+        command: linkerd
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -6,20 +6,20 @@ description: |
     Linkerd is a service mesh, designed to give platform-wide observability,
     reliability, and security without requiring configuration or code changes.
 
-#icon: https://user-images.githubusercontent.com/9226/33582867-3e646e02-d90c-11e7-85a2-2e238737e859.png
 grade: devel
 confinement: devmode
 
 parts:
     linkerd:
-        plugin: go
-        go-importpath: github.com/linkerd/linkerd2
+        plugin: dump
         source: .
-        source-type: git
-        override-build: |
-            cd /root/parts/linkerd/src/
-            ./bin/build-cli-bin
-            mv ./target/cli/linux/linkerd ~/prime/
+        override-pull: |
+            cd "$SNAPCRAFT_PART_SOURCE" # this line might be wrong. also maybe unneeded
+            curl -sSLO https://github.com/linkerd/linkerd2/releases/download/edge-20.3.1/linkerd2-cli-edge-20.3.1-linux
+            mv linkerd2-cli-edge-20.3.1-linux linkerd # rename binary to linkerd
+            chmod +x linkerd
+        build-packages:
+            - curl
 apps:
     linkerd:
         command: linkerd


srv-twry:
[linkerd_2.7.0_amd64.snap.zip](https://github.com/linkerd/linkerd2/files/4299221/linkerd_2.7.0_amd64.snap.zip)

The snap generated for the `Snapcraft.yaml` configuration.

To install:
1. Extract the `.snap` package out of the zip
2. Run `sudo snap install linkerd_2.7.0_amd64.snap --dangerous --classic`

Once you're done you can remove it with `sudo snap remove linkerd`

@ihcsim @grampelberg Is there any plan to integrate this in edge release workflow?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
...
+grade: stable
+confinement: classic
+
+parts:
+    linkerd:
+        plugin: dump
+        source: .
+        override-pull: |
+            curl -sSLO https://github.com/linkerd/linkerd2/releases/download/stable-2.7.0/linkerd2-cli-stable-2.7.0-linux
+            mv linkerd2-cli-stable-2.7.0-linux linkerd # rename binary to linkerd
```
srv-twry: The version number is hardcoded for now. I can inject it using environment variables if required.

ihcsim:
> @ihcsim @grampelberg Is there any plan to integrate this in edge release workflow?

This needs to be part of our CI/CD workflow.

@alpeb I am hoping we can update our CI to [build](https://github.com/marketplace/actions/snapcraft-action) this snapcraft file. What's the best way to attach the generated snap file to future GitHub releases? My understanding is that the release process (including the attachment of the CLI binaries) is now automated.

snap/snapcraft.yaml
ihcsim: We can add a step in our CI job to `sed` the release number.

alpeb:
> @alpeb I am hoping we can update our CI to [build](https://github.com/marketplace/actions/snapcraft-action) this snapcraft file. What's the best way to attach the generated snap file to future GitHub releases? My understanding is that the release process (including the attachment of the CLI binaries) is now automated.

Correct. A new step would need to be added to [this job](https://github.com/linkerd/linkerd2/blob/abbf4a4e609afc9f8395e3c56e68765ee1315fd1/.github/workflows/release.yml#L221) right before the "Create release step", using that snapcraft-action. And then the next step would pick up the generated file to upload it to Github with the other release assets.

ihcsim:
@alpeb Thanks for the pointer. From what I can see, the only way to test this job is with an edge or stable tag, right? It doesn't look like it's something that we can/should include in this PR, as it will remain untested. Maybe I can create a separate issue to update the job?

alpeb:
Yeah a separate PR should be fine. The way I've been testing this sort of thing is to follow the same release process but in a separate fork, using a trimmed down `release.yaml` file focusing on what is to be tested.

srv-twry:
> Correct. A new step would need to be added to this job right before the "Create release step", using that snapcraft-action. And then the next step would pick up the generated file to upload it to Github with the other release assets.

I think in that case, I'll have to modify this snapcraft a little bit because I am pulling the CLI binary from the Github release i.e. after the "Create release step". I'll have to modify it to pull from the local file system instead of Github

snap/snapcraft.yaml
srv-twry: Yeah we could have. We don't need it now though :)

srv-twry:
I have modified the snapcraft config to use the locally stored CLI binary. 

I saw the [release jobs](https://github.com/linkerd/linkerd2/blob/abbf4a4e609afc9f8395e3c56e68765ee1315fd1/.github/workflows/release.yml#L251) file and noticed that it stores the CLI binary at `./target/release/linkerd2-cli-*-linux`. So, the config expects the binary to be present there.

### Local Testing
In order to test this configuration locally:
1. install `snapcraft` 
2. Make sure CLI binary is present in `./target/release/linkerd2-cli-*-linux` directory
3. Run `snapcraft`. This will generate a `.snap` file.
4. Install it using [these](https://github.com/linkerd/linkerd2/pull/4132#issuecomment-595848060) instructions

alpeb:
In order to avoid having to use the `--dangerous` flag when installing this snap, we'd need to generate a key. I assume that requires creating a developer account and publishing the snap along with the public key in the snap store. @grampelberg you think that's an avenue worth pursuing?

grampelberg:
@alpeb sounds reasonable to me.

alpeb:
@srv-twry I tried uploading the generated snap into the snap store, and it was blocked because it uses "classic" confinement. We could ask for manual review and provide the reasons why we need full machine access. But what if we give a try at "strict" confinement instead? That'd require adding into the `Snapcraft.yaml` file the interfaces necessary for at least:
- Network access
- ~/.kube read access

Would you give that a try?
I can then run the full integration tests using the snap and see how it goes :+1:

srv-twry:
Sure I'll do that.

srv-twry:
@alpeb I tried to use the `strict` confinement and been facing some issues. I've opened a support discussion topic on the forum and provided the details. You might want to check it out. https://forum.snapcraft.io/t/cannot-read-kubernetes-config/16216

chris13524:
On finding the `kubectl` version, wouldn't you have to install that binary in the linkerd snap?

alpeb:
> On finding the `kubectl` version, wouldn't you have to install that binary in the linkerd snap?

Yea, you probably need to bundle `kubectl` as I don't think the snap will have access to `/usr/local/bin`. Also it appears the `home` interface doesn't provide access to hidden (dot) files so `~/.kube` is off-limits. It seems the `personal-files` interface can do the trick instead.

srv-twry:
@chris13524 @alpeb Thanks for your suggestions. I've found a [forum thread](https://forum.snapcraft.io/t/request-for-personal-files-confinement-for-k9s-popeye/10976) in which the developer is facing similar issues.

> On finding the kubectl version, wouldn't you have to install that binary in the linkerd snap?

Yeah it seems that way from the discussions in the thread I linked above.

>  It seems the personal-files interface can do the trick instead.

I am using the personal-files interface as well. It seems like the variable `$HOME/.kube` points to `~/snap/linkerd/<revision>/.kube` instead of the `~/.kube` which I expect.

I think there are a few loopholes in my understanding of how snaps work. I'll take some to read up some more and then try to fix this.

srv-twry:
@alpeb Thanks for being patient with this one. I got busy with my internship hence wasn't able to devote much time to this.

So here is what I understood from going through the conversations here: https://forum.snapcraft.io/t/request-for-personal-files-confinement-for-k9s-popeye/10976/1

1. We need to ship our own `kubectl` binary with the snap as stated [here](https://forum.snapcraft.io/t/request-for-personal-files-confinement-for-k9s-popeye/10976/26?u=srv_twry). Personally, I am not in favour of doing so. It would mean that the user will have to use `linkerd.kubectl ...` instead of `kubectl ...`. If they use `kubectl ..` they may face issues because we wouldn't be having any info about the version. This looks far from ideal in my opinion.

2. I was using the `personal-files` interface but still the program is unable to find the kubeconfig. If I explicitly provide the kubeconfig using `--kubeconfig ~/.kube/config`, linkerd is able to locate the file. So there is no access issue. The issue that Snap sets the `$HOME` environment variable to point to `~/snap/linkerd/<revision>/.kube`. In order to fix this, we'll need to create a symlink from `~/snap/linkerd/<revision>/.kube` to `~/.kube` as suggested [here](https://forum.snapcraft.io/t/request-for-personal-files-confinement-for-k9s-popeye/10976/21?u=srv_twry)

I think we should try for classic confinement especially due to the need of shipping our own version of kubectl. Also tools like [helm](https://forum.snapcraft.io/t/classic-confinement-for-existing-helm-snap/4375) and kubectl are using it.

Please let me know if you still want me to get this done using strict confinement and I'll package `kubectl` inside the snap and add the symlink script.

alpeb:
Thank you very much @srv-twry for the extra research :+1: 
The case is clear indeed in favor of remaining with classic confinement. I've just submitted the package for manual review. I'll keep you posted on how that goes :pray: :prayer_beads:

alpeb:
I've just received a notification about rejection after manual review, so I posted into their store-requests forum category as instructed:
https://forum.snapcraft.io/t/request-for-classic-confinement-for-linkerd/17461

alpeb:
@dr0pdb the classic confinement request isn't getting through...
I've dug a little deeper into this and we might give a shot again at strict confinement using the `personal-files` interface. I did verify that grants access to the real `$HOME` dir, not the snap's VM `$HOME` dir:

```
name: linkerd
base: core18
version: '2.7.1'
summary: Command-line utility to interact with linkerd
description: |
    Linkerd is a service mesh, designed to give platform-wide observability,
    reliability, and security without requiring configuration or code changes.

grade: stable
confinement: strict

parts:
    linkerd:
        plugin: dump
        source: .
        stage:
            - target/release/linkerd2-cli-*-linux
        override-stage: |
            snapcraftctl stage
            cp target/release/linkerd2-cli-*-linux $SNAPCRAFT_PRIME/linkerd

plugs:
  kube-config:
    interface: personal-files
    read:
      - $HOME/.kube

apps:
    linkerd:
        command: linkerd
        plugs:
          - kube-config
          - network
          - network-bind
```
We also need the `network` and `network-bind` interfaces to be able to use the network from within the linkerd CLI. These two interfaces are automatically connected. On the other hand, the `kube-config` interface requires manual connection:
```
snap connect linkerd:kube-config
```
As for the requirement to access `kubectl`, I'm not sure how much that's really necessary. We can start by having as a goal a successful run of `linkerd check --pre`, and having `CheckKubectlVersion()` in `kubectl.go` always return nil for now, and solve that problem later.

Another problem is that the snap doesn't have access to the `KUBECONFIG` environment variable. Apparently a snap hook needs to be set, that uses `snapctl get` to grab that setting and populate the environment variable. And the user would need to set it before triggering the snap through `snap set`.

Let me know if you're still interested in pursuing these tasks or if we should close this PR, which would be understandable given how much unforeseeable effort this is requiring. Thanks in any case!  :slightly_smiling_face:

dr0pdb:
@alpeb Hi, 

Thanks for the detailed info.

Currently, I am busy in my university exams and B.Tech. thesis work which should take another 2 weeks.

I think it would be better to close this PR. Once I get free and get some time to implement these, I'll reopen this or open another PR.

PR #4134, opened 2020-03-04
Enable dst service to work with remote gateway endpoints

zaharidichev:
The purpose of this change is to enable the destinations service to recognize remote gateway endpoints and be able to set the corresponding labels, identitiy and the rest of the data that will be needed for the proxy to handle the traffic properly. 

@adleong since you mostly wrote that code, it would be good to get your opinion on the direction of these changes. Effectivelly I want to express the idea that endpoints and hende `WeightedAddress` instaces can be the product of things different than pods. Still have to add more docs and write some more tests for this.

Description: 

The most prominant change here is that we break up the concrete `Address` type into three types: 

- `PodAddress` -  is an address/port combo that comes from Endpoints that have been created because of pods
- `HeadlessServiceAddress` - what it sounds like. These have been created because of a service with an associated Endpoints object that does not have ties to a Pod
- `RemoteGatewayAddress` - this is a subcase of `HeadlessServiceAddress`  but contains additional infromation for remote gateway identitiy and service fq name. 

Note that this PR is the first part of this change.  We still need to set these annotations when we create endpoitns in the mirroring controller. We also need to add metrics labels, etc and we need to eventually change the API to allow for carying data about the remote gateway (such as fully qualified name of the service we are targeting on the other end). These will come as a seprate smaller PRs. 

### DIFF
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/linkerd/linkerd2/controller/k8s"
+	consts "github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/prometheus/client_golang/prometheus"
 	logging "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -24,19 +25,24 @@ const (
 // https://github.com/linkerd/linkerd2/issues/2204
 
 type (
-	// Address represents an individual port on an specific pod.
+	// Address represents an individual port on a specific endpoint.
+	// This endpoint might be the result of a the existence of a pod
+	// that is targeted by this service; alternatively it can be the
+	// case that this endpoint is not associated with a pod and maps
+	// to some other IP (i.e. a remote service gateway)
 	Address struct {
 		IP        string
 		Port      Port
 		Pod       *corev1.Pod
 		OwnerName string
 		OwnerKind string
+		Identity  string
 	}
 
-	// PodSet is a set of pods, indexed by IP.
-	PodSet struct {
-		Pods   map[PodID]Address
-		Labels map[string]string
+	// AddressSet is a set of Address, indexed by ID.
+	AddressSet struct {
+		Addresses map[ID]Address
+		Labels    map[string]string
 	}
 
 	portAndHostname struct {
@@ -89,15 +95,15 @@ type (
 		k8sAPI     *k8s.API
 
 		exists    bool
-		pods      PodSet
+		addresses AddressSet
 		listeners []EndpointUpdateListener
 		metrics   endpointsMetrics
 	}
 
 	// EndpointUpdateListener is the interface that subscribers must implement.
 	EndpointUpdateListener interface {
-		Add(set PodSet)
-		Remove(set PodSet)
+		Add(set AddressSet)
+		Remove(set AddressSet)
 		NoEndpoints(exists bool)
 	}
 )
@@ -417,32 +423,32 @@ func (sp *servicePublisher) metricsLabels(port Port, hostname string) prometheus
 // portPublisher.
 
 func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
-	newPods := pp.endpointsToAddresses(endpoints)
-	if len(newPods.Pods) == 0 {
+	newAddressSet := pp.endpointsToAddresses(endpoints)
+	if len(newAddressSet.Addresses) == 0 {
 		for _, listener := range pp.listeners {
 			listener.NoEndpoints(true)
 		}
 	} else {
-		add, remove := diffPods(pp.pods, newPods)
+		add, remove := diffAddresses(pp.addresses, newAddressSet)
 		for _, listener := range pp.listeners {
-			if len(remove.Pods) > 0 {
+			if len(remove.Addresses) > 0 {
 				listener.Remove(remove)
 			}
-			if len(add.Pods) > 0 {
+			if len(add.Addresses) > 0 {
 				listener.Add(add)
 			}
 		}
 	}
 	pp.exists = true
-	pp.pods = newPods
+	pp.addresses = newAddressSet
 
 	pp.metrics.incUpdates()
-	pp.metrics.setPods(len(pp.pods.Pods))
+	pp.metrics.setPods(len(pp.addresses.Addresses))
 	pp.metrics.setExists(true)
 }
 
-func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) PodSet {
-	pods := make(map[PodID]Address)
+func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) AddressSet {
+	addresses := make(map[ID]Address)
 	for _, subset := range endpoints.Subsets {
 		resolvedPort := pp.resolveTargetPort(subset)
 		for _, endpoint := range subset.Addresses {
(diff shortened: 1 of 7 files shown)


olix0r:
@zaharidichev I don't want to create extra work, but would it be possible to get a "clean" change to the existing endpoints code without introducing the new features, and then following on with a PR that adds remote gateway support?

olix0r:
> would it be possible to get a "clean" change

Actually, on second look, the actual surface area of this change is smaller than it first appeared (thanks gofmt). If I understand correctly, this is basically introduce a polymorpish address type instead of the single concrete `Address` type? If so, i don't know a separate PR is worth it.

zaharidichev:
> If I understand correctly, this is basically introduce a polymorpish address type instead of the single concrete Address type?


@olix0r that is the just of it (for some definition of polymorpish that Go adopts .... )

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -24,19 +25,45 @@ const (
 // https://github.com/linkerd/linkerd2/issues/2204
 
 type (
-	// Address represents an individual port on an specific pod.
-	Address struct {
+
+	// Address represents an individual port on a specific endpoint.
+	// This endpoint might be associated with a Pod, a Headless service
+	// or a mirrored service that routes to a remote gateway.
+	Address interface {
```
adleong: There's nothing I love more than some good old fashioned polymorphism.  However, I'm not sure it's a good fit here.  This interface gives a false sense of encapsulation: we can use this interface in some places, but in others we have to reach inside and switch on the concrete type.

It may be simpler to deal with this as a single struct.  The `Pod`, `OwnerName`, and `OwnerKind` fields can simply be empty when they don't apply (i.e. when the address does not correspond to a pod) just like today.  And we can add an `Identity` field which can provide the identity name when the pod is not present.  I would perhaps call `FullyQualifiedRemoteName` something more aligned with how its effect such as `AuthorityOverride` or something like that.  Alternatively, we can omit this field for now since it's unused and a better name might be more obvious once we have the destination API change solidified.  The `Namespace` and `LocalServiceName` fields appear to be completely unused.  If the goal is to use them to propagate metric labels, we may want to simply have a labels map on the Address struct that we can set directly.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -24,19 +25,45 @@ const (
 // https://github.com/linkerd/linkerd2/issues/2204
 
 type (
-	// Address represents an individual port on an specific pod.
-	Address struct {
+
+	// Address represents an individual port on a specific endpoint.
+	// This endpoint might be associated with a Pod, a Headless service
```
adleong: There are some existing references to "Headless" services in various comments.  These are incorrect.  What they refer to as "headless services" are actually services which have endpoints that do not correspond to pods.  We should remove these misleading comments and variable names.

A headless service is one without a clusterIP.  Linkerd does not interact with the clusterIP of a service and does not care if a service is headless or not.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -24,19 +25,45 @@ const (
...
+	}
+
+	// HeadlessServiceAddress is a an address of a headless service
+	HeadlessServiceAddress struct {
+		IP   string
+		Port Port
+	}
+
+	// AddressSet is a set of Address, indexed by ID.
+	AddressSet struct {
```
adleong: +1 to this rename

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -584,28 +672,27 @@ func getTargetPort(service *corev1.Service, port Port) namedPort {
 	return targetPort
 }
 
-func diffPods(oldPods, newPods PodSet) (add, remove PodSet) {
-	// TODO: this detects pods which have been added or removed, but does not
-	// detect pods which have been modified.  A modified pod should trigger
+func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSet) {
+	// detect addresses which have been modified.  A modified address should trigger
```
adleong: missing TODO

controller/api/destination/watcher/ip_watcher.go
```diff
@@ -218,7 +218,7 @@ func (iw *IPWatcher) getOrNewServiceSubscriptions(clusterIP string) *serviceSubs
 			pods := []*corev1.Pod{}
 			for _, obj := range objs {
 				if pod, ok := obj.(*corev1.Pod); ok {
-					// Skip pods with HostNetwork.
+					// Skip addresses with HostNetwork.
```
adleong: "pod" is still accurate here

controller/api/destination/watcher/ip_watcher.go
```diff
@@ -244,18 +244,18 @@ func (iw *IPWatcher) getServiceSubscriptions(clusterIP string) (ss *serviceSubsc
 	return
 }
 
-func (iw *IPWatcher) podToPodSet(pod *corev1.Pod) PodSet {
+func (iw *IPWatcher) podToPodSet(pod *corev1.Pod) AddressSet {
```
adleong: rename to "podToAddressSet"

controller/api/destination/watcher/ip_watcher_test.go
```diff
@@ -239,7 +239,7 @@ status:
 			expectedError:                    false,
 		},
 		{
-			serviceType: "local services with missing pods",
+			serviceType: "local services with missing addresses",
```
adleong: I don't think the replacements in this file are correct

controller/api/public/grpc_server_test.go
```diff
@@ -30,7 +30,7 @@ type listServicesExpected struct {
 	res    pb.ListServicesResponse
 }
 
-// sort Pods in ListPodResponses for easier comparison
+// sort Addresses in ListPodResponses for easier comparison
```
adleong: I don't think this is correct

controller/k8s/api.go
```diff
@@ -301,7 +301,7 @@ func (api *API) RS() appv1informers.ReplicaSetInformer {
 	return api.rs
 }
 
-// Pod provides access to a shared informer and lister for Pods.
+// Pod provides access to a shared informer and lister for Addresses.
```
adleong: I don't think the replacements in this file are correct

controller/k8s/api_test.go
```diff
@@ -851,7 +851,7 @@ spec:
       pod-template-hash: pod2`,
 				},
 			},
-			// Deployment 2 Pods just one valid
+			// Deployment 2 Addresses just one valid
```
adleong: I don't think this is correct

test/install_test.go
```diff
@@ -401,7 +401,7 @@ func TestResourcesPostInstall(t *testing.T) {
 		}
 	}
 
-	// Tests Pods and Deployments
+	// Tests Addresses and Deployments
```
adleong: I don't think this is correct

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: > This interface gives a false sense of encapsulation: we can use this interface in some places, but in others we have to reach inside and switch on the concrete type.

Maybe its me but I do not see this as such a problem. Seems in line with trying to express the kind of addresses as a collection of ADTs. 

> t may be simpler to deal with this as a single struct. The Pod, OwnerName, and OwnerKind fields can simply be empty when they don't apply (i.e. when the address does not correspond to a pod) just like today. 

The thing I did not like about this approach (as this is how it was before) is that we are using the presence or abence of values to determine what _type_ of thing we are lookin at. Is this a headless service or a pod address, or whatever else.. 

That all being said, lets try and see how it looks without the different types.

adleong: This gets a bit philosophical, but I don't think the presence or absence of the pod field fundamentally changes the type of the address.  The way I think of it is this: if the Address has a pod, we'll use that to populate certain metadata; if it doesn't we won't.

zaharidichev: putting it that way, it makes sense.

PR #4141, opened 2020-03-06
Improve error handling in alpha stat command

adleong:
More helpful error messages when the `linkerd alpha stat` command fails.  For example, when the user is not authorized:

```
> linkerd alpha stat deploy/web -n emojivoto --as obama@buoyant.io
Error: deployments.metrics.smi-spec.io "web" is forbidden: User "obama@buoyant.io" cannot get resource "deployments" in API group "metrics.smi-spec.io" in the namespace "emojivoto"
```

When an error is encountered on the server:

```
> linkerd alpha stat deploy -n emojivoto
Error: Unauthorized client certificate. Check configuration and try again.
```

### DIFF
--- a/pkg/smimetrics/client.go
+++ b/pkg/smimetrics/client.go
@@ -151,15 +151,23 @@ func getMetricsResponse(k8sAPI *k8s.KubernetesAPI, path string, params map[strin
 	return ioutil.ReadAll(httpRsp.Body)
 }
 
-func parseTrafficMetrics(bytes []byte) (*TrafficMetrics, error) {
+func handleStatusResponse(bytes []byte) error {
 	var status Status
 	json.Unmarshal(bytes, &status)
 	if status.Kind == "Status" {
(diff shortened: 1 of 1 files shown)


pkg/smimetrics/client.go
```diff
@@ -143,23 +152,39 @@ func getMetricsResponse(k8sAPI *k8s.KubernetesAPI, path string, params map[strin
 }
 
 func parseTrafficMetrics(bytes []byte) (*TrafficMetrics, error) {
+	var status Status
+	json.Unmarshal(bytes, &status)
+	if status.Kind == "Status" {
+		return nil, errors.New(status.Message)
+	}
+
```
zaharidichev: Seems this checking logic can be factred out in its own method

adleong: Thanks to go's verbose error checking, refactoring this logic into its own method actually net *added* 7 lines of code 🙃

PR #4142, opened 2020-03-06
Add -A flag to alpha stat

adleong:
Add an `--all-namespaces` flag to `linkerd alpha stat`.  This flag ignore the value of the `--namespace` flag and looks up resources across all namespaces.

Some example usage:

```
> linkerd alpha stat po -A
NAMESPACE  NAME                                    SUCCESS        RPS  LATENCY_P50  LATENCY_P90  LATENCY_P99
default    curl                                    100.00%     0.6rps          1ms          1ms          1ms
emojivoto  emoji-ffd474b7b-nq8wc                   100.00%     2.0rps          1ms          1ms          1ms
emojivoto  vote-bot-74c4867dc6-d5j4d                90.00%     2.0rps          3ms          4ms          4ms
emojivoto  voting-6b69659f5b-6hpvx                  78.95%     0.9rps          1ms          1ms          1ms
emojivoto  web-6cfccddd6b-vrq2q                     92.86%     5.6rps          1ms          3ms          4ms
linkerd    linkerd-controller-54bbb5d485-4p9w2     100.00%     0.3rps          1ms          1ms          1ms
linkerd    linkerd-destination-69fb65c4fb-7mthj    100.00%     0.3rps          1ms          1ms          1ms
linkerd    linkerd-grafana-ffc4d969-gf5cz          100.00%     0.3rps          1ms          2ms          2ms
linkerd    linkerd-identity-6456988769-tbkx9       100.00%     0.3rps          1ms          1ms          1ms
linkerd    linkerd-prometheus-5469d5d8fd-kskc6     100.00%     2.5rps          1ms          2ms          3ms
linkerd    linkerd-proxy-injector-658f8c4cd-pfgbt  100.00%     0.3rps          1ms          1ms          1ms
linkerd    linkerd-smi-metrics-86567c5ff4-dh7rn          -     0.0rps          0ms          0ms          0ms
linkerd    linkerd-sp-validator-54c8d7dcf9-wq6jv   100.00%     0.3rps          1ms          2ms          2ms
linkerd    linkerd-tap-574b74c964-cwm6l            100.00%     0.3rps          1ms          1ms          1ms
linkerd    linkerd-web-577755788d-95slx            100.00%     0.3rps          1ms          1ms          1ms
```

```
> linkerd alpha stat po/curl --to po
FROM  TO                              SUCCESS        RPS  LATENCY_P50  LATENCY_P90  LATENCY_P99
curl  web-6cfccddd6b-vrq2q.emojivoto  100.00%     0.9rps          1ms          2ms          2ms
```

### DIFF
--- a/cli/cmd/alpha_stat.go
+++ b/cli/cmd/alpha_stat.go
@@ -14,6 +14,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
@@ -87,7 +88,7 @@ Examples:
 
 			namespace := options.namespace
 			if options.allNamespaces {
-				namespace = ""
+				namespace = corev1.NamespaceAll
 			}
 			target, err := util.BuildResource(namespace, args[0])
 			if err != nil {
--- a/cli/cmd/alpha_stat.go
+++ b/cli/cmd/alpha_stat.go
@@ -30,8 +30,9 @@ var allowedKinds = map[string]struct{}{
 }
 
 type alphaStatOptions struct {
-	namespace  string
-	toResource string
+	namespace     string
+	toResource    string
+	allNamespaces bool
 }
 
 func newCmdAlphaStat() *cobra.Command {
@@ -84,7 +85,16 @@ Examples:
 				return err
 			}
 
-			target, err := util.BuildResource(options.namespace, args[0])
+			namespace := options.namespace
+			if options.allNamespaces {
+				namespace = ""
+			}
+			target, err := util.BuildResource(namespace, args[0])
+			if target.GetName() != "" {
+				// Getting a named resource from all-namespaces is not supported.
+				// Fall back to the configured namespace.
+				target.Namespace = options.namespace
+			}
 			if err != nil {
 				return err
 			}
@@ -96,7 +106,7 @@ Examples:
 				return err
 			}
 			name := target.GetName()
-			toResource := buildToResource(options.namespace, options.toResource)
+			toResource := buildToResource(namespace, options.toResource)
 			// TODO: Lift this requirement once the API supports it.
 			if toResource != nil && toResource.GetType() != target.GetType() {
 				return errors.New("the --to resource must have the same kind as the target resource")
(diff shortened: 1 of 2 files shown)


cli/cmd/alpha_stat.go
```diff
@@ -84,7 +85,16 @@ Examples:
 				return err
 			}
 
-			target, err := util.BuildResource(options.namespace, args[0])
+			namespace := options.namespace
+			if options.allNamespaces {
+				namespace = ""
```
zaharidichev: I think it makes sense to use `corev1.NamespaceAll` here, unless you are against importing that

adleong:
I've done some more holistic thinking about how namespaces should work and I've made a few changes according to these principles:

* The `--to` resource will always look up across all namespaces.  There is no performance cost because `--to` is implemented with filtering.  This allows us to avoid adding a `--to-namespace` flag which gets confusing and unwieldy.  It also makes sense that when doing a `--to` query, we want to see all the outgoing connections from our resource, no matter what namespaces they are in.
* `-A` cannot be used with a named resource and now fails with an error message.  This is consistent with `kubectl get`.
* If the `-A` flag is used, then we add a "NAMESPACE" column.  This is consistent with `kubectl get -A`.
* Since the `--to` resource may be in any namespace, we include the namespace in its name.  e.g. "voting.emojivoto".  In my opinion this is less clunky than having a "TO-NAMESPACE" column.

adleong:
@alpeb @zaharidichev Thanks for your reviews!  Since the latest round of changes are quite significant, I'd appreciate it if you could give this a quick re-review.  🥂

cli/cmd/alpha_stat.go
```diff
@@ -242,8 +259,15 @@ func metricsToRow(metrics *smimetrics.TrafficMetrics, outbound bool) []string {
 	}
 }
 
-func buildTable(outbound bool) table.Table {
+func buildTable(outbound, allNamespaces bool) table.Table {
 	columns := []table.Column{
+		table.Column{
+			Header:    "NAMESPACE",
```
kleimkuhler: Just a heads up that this can render as:

```
❯ bin/linkerd alpha stat pods --all-namespaces
NAMESPA  NAME
linkerd  linkerd-controller-..
```

I'm not sure if `kubectl` similarly trims the column name as well.. it picks up pods in `kube-system` so `NAMESPACE` is fully rendered anyways.

adleong: Fixed, thanks

PR #4143, opened 2020-03-06
Set the published version check to sleep for 30 seconds

cpretzer:
Configures the "Check published version" step in the Release github action to wait for 30 seconds instead of 5


adleong:
Just so that it's chronicled somewhere, what is the context of this change?

alpeb:
@adleong The `release.yml` workflow waits for the website to have been updated with the latest release version and checks so in a loop. As you can see in yesterday's release run
https://github.com/linkerd/linkerd2/runs/489101463?check_suite_focus=true
the loop needs to give more time for such update to be visible.

Issue #3183, opened 2019-08-01
Handle automountServiceAccountToken


grampelberg:
### What problem are you trying to solve?

mTLS is on by default now and requires the service account token to create an identity. Some pods start with `automountServiceAccountToken: false`. This causes the pods to go into a crash loop as they're unable to fetch an identity.

### How should the problem be solved?

- If `automountServiceAccountToken: false` is set on a pod spec, inject should return an error. This should happen for any injection issue (host networking, known incompatible sidecar).
- An annotation should be added `config.linkerd.io/forceServiceAccountToken` that adds `automountServiceAccountToken: true` to the patch. This is to make it work with controllers that create pods outside of a normal user's control.

grampelberg:
I'm not sure this is the right idea. The best solution would be to use projected volumes, that is gated on figuring out whether it is generally available or not.

ihcsim:
@grampelberg Given https://github.com/linkerd/linkerd2/issues/3260, is this still relevant?

grampelberg:
@ihcsim we'll need to decide once there's some data around how many folks can use the token volumes. Definitely going to end up guiding how we fit this together.

mayankshah1607:
@ihcsim @grampelberg I'd love to take up this one if it is still relevant. :)
Seems like https://github.com/linkerd/linkerd2/issues/2843 could be related.

ihcsim:
@mayankshah1607 The only part that we can do here is to return an error and emit an event when `automountServiceAccountToken: false`, which implies that mTLS won't work. If you want to tackle this, feel free! 

The [service account token volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) is still in beta. And I don't think introducing a new annotation to modify this field behind the user's back is the right approach.

Issue #2843, opened 2019-05-23
Fail to inject when mTLS is enabled and there is no service account


csreegn:
### What is the issue?
When linkerd injects it's proxy into a pod where service account automount is disabled, **the pod fails to start up** And goes into a restart loop. The root cause is the linkerd-proxy looking for the mounted serviceaccount info but not finding it.

### How can it be reproduced?

Try to instrument a pod that has the following attribute set in it's spec:
```
automountServiceAccountToken: false
```

I encountered this issue with the official argocd manifest: https://argoproj.github.io/argo-cd/getting_started/#1-install-argo-cd

### Logs, error output, etc

On pod startup the linkerd-proxy container outputs this:
```
time="2019-05-23T15:47:08Z" level=info msg="running version dev-undefined"
time="2019-05-23T15:47:08Z" level=info msg="Using with pre-existing key: /var/run/linkerd/identity/end-entity/key.p8"
time="2019-05-23T15:47:08Z" level=info msg="Using with pre-existing CSR: /var/run/linkerd/identity/end-entity/key.p8"
ERR! [     0.000310s] linkerd2_proxy::app::config Could not read LINKERD2_PROXY_IDENTITY_TOKEN_FILE: No such file or directory (os error 2)
ERR! [     0.000354s] linkerd2_proxy::app::config LINKERD2_PROXY_IDENTITY_TOKEN_FILE="/var/run/secrets/kubernetes.io/serviceaccount/token" is not valid: InvalidTokenSource
configuration error: InvalidEnvVar
```

#### `linkerd check` output

```text
➜ linkerd check
kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version
√ is running the minimum kubectl version

linkerd-existence
-----------------
√ control plane namespace exists
√ controller pod is running
√ can initialize the client
√ can query the control plane API

linkerd-api
-----------
√ control plane pods are ready
√ control plane self-check
√ [kubernetes] control plane can talk to Kubernetes
√ [prometheus] control plane can talk to Prometheus
√ no invalid service profiles

linkerd-version
---------------
√ can determine the latest version
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date
√ control plane and cli versions match

Status check results are √
```

### Environment

- Kubernetes Version:
```bash
➜ kubectl version
[...]

PR #4145, opened 2020-03-06
Update inject to handle `automountServiceAccountToken: false`

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/3183 
Fixes https://github.com/linkerd/linkerd2/issues/2843

When a pod spec contains `automountServiceAccountToken: false` :
- `inject` throws an error
- ~webhook emits an `InjectionSkip` event with the message `Linkerd sidecar proxy injection skipped: automountServiceAccountToken set to "false"`~
```bash
$ cat deploy.yaml
kind: Deployment
apiVersion: apps/v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      automountServiceAccountToken: false
      containers:
      - name: nginx
        image: nginx
        ports:
        - name: http
          containerPort: 80

$ cat deploy.yaml | bin/go-run cli inject -
Error transforming resources: cannot enable mTLS when automountServiceAccountToken set to "false"
```

Unit tests have been added to exercise the newly added code.


ihcsim:
The output of the `linkerd inject` command looks a bit off:

```
bin/linkerd version                          
Client version: git-590100a3
Server version: git-590100a3

cat <<EOF| bin/linkerd inject -
kind: Deployment
apiVersion: apps/v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      automountServiceAccountToken: false
      containers:
      - name: nginx
        image: nginx
        ports:
        - name: http
          containerPort: 80
EOF

‼ "linkerd.io/inject: disabled" annotation set on deployment/nginx
‼ no supported objects found

deployment "nginx" skipped
```

If I piped that into `kubectl apply`, the deployment is still created as an uninjected workload. Can we return an error and not create the workload instead?

```
cat <<EOF| bin/linkerd inject - | k apply -f -     (kubernetes-admin@linkerd/default)
kind: Deployment
apiVersion: apps/v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      automountServiceAccountToken: false
      containers:
      - name: nginx
        image: nginx
        ports:
        - name: http
          containerPort: 80
EOF

kubectl get po                                           
NAME                     READY   STATUS    RESTARTS   AGE
nginx-5dd74fff98-85cv4   1/1     Running   0          10s
```

mayankshah1607:
@ihcsim From the issue discussion thread from https://github.com/linkerd/linkerd2/issues/3183
> @mayankshah1607 The only part that we can do here is to return an error and emit an event when automountServiceAccountToken: false, which implies that mTLS won't work. 

If we just throw an error and not create any workload, that would mean that the webhook wouldn't get triggered, hence not throwing an `InjectionSkip` event, right?

mayankshah1607:
@ihcsim Done! The output of `inject` now looks like this
```bash
$ cat deploy.yaml
kind: Deployment
apiVersion: apps/v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      automountServiceAccountToken: false
      containers:
      - name: nginx
        image: nginx
        ports:
        - name: http
          containerPort: 80

$ cat deploy.yaml | bin/go-run cli inject -
Error transforming resources: disabled_automount_service_account_token_account
```
As mentioned earlier, now that the workload is not created, it will not trigger the webhook to emit `InjectionSkip` events. Let me know if this works. I've also updated the tests.

pkg/inject/report.go
```diff
@@ -151,6 +153,14 @@ func (r *Report) disableByAnnotation(conf *ResourceConfig) (bool, string, string
...
 	// cli     | n/a       | disabled | no       | true
 
+	// Check https://github.com/linkerd/linkerd2/issues/3183
+	serviceAccountToken := conf.pod.spec.AutomountServiceAccountToken
+	if serviceAccountToken != nil {
+		if !*serviceAccountToken {
+			return true, disabledAutomountServiceAccountToken, ""
+		}
+	}
+
```
alpeb: I think this doesn't belong under the function `disableByAnnotation()`. Can you instead surface that as a new `Report` struct field, like `HostNetwork` and `Sidecar`? You could then set it in `newReport`, and use it in `Injectable()`.

alpeb:
> Can we return an error and not create the workload instead?

@ihcsim , when the pod has hostNetworking, or already has a sidecar, we inform the problem for this particular resource, leave it untouched, and continue our business with other resources (if passing a list through CLI). Any particular reason why this case should be handled differently?

ihcsim:
@alpeb The original requirement defined in https://github.com/linkerd/linkerd2/issues/3183 is  to error out. I just assume we want mTLS to be mandatory. As a user, I'd assume if my workload doesn't fulfill the injection requirements, the injection process errors out, and not proceed to deploy things into prod environment that can't be part of the service mesh.

mayankshah1607:
Thanks @alpeb 
I've updated the code to take the `Injectable()` path. Let me know if this looks good. 
The output of `inject` now looks like this :
```bash
$ cat deploy.yaml
kind: Deployment
apiVersion: apps/v1
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      automountServiceAccountToken: false
      containers:
      - name: nginx
        image: nginx
        ports:
        - name: http
          containerPort: 80

$ cat deploy.yaml | bin/go-run cli inject -
Error transforming resources: disabled_automount_service_account_token_account
```
`inject` just errors out when the pod spec has `automountServiceAccountToken: false`

alpeb:
> @alpeb The original requirement defined in #3183 is to error out. I just assume we want mTLS to be mandatory. As a user, I'd assume if my workload doesn't fulfill the injection requirements, the injection process errors out, and not proceed to deploy things into prod environment that can't be part of the service mesh.

#3183 also mentions:
```
If automountServiceAccountToken: false is set on a pod spec, inject should return an error. This should happen for any injection issue (host networking, known incompatible sidecar).
```
The current behavior for non-compliant pods is to issue a warning and not inject. Is the plan to change that behavior? I'm not against that, I just want for all these cases (host networking, incompatible sidecar and now disabled automoutServiceAccountToken) to behave the same and try not introducing a special case.
Or perhaps we shouldn't change behavior for now, and offer the "injection-assurance" guarantee only till that's something made explicit when we tackle policy in the future?
CC @grampelberg

ihcsim:
@alpeb SGTM either way. I am curious as to why must they behave the same?

grampelberg:
@alpeb I'm not sure I understand. If someone's explicitly asked for injection and we're unable to do it, I would like for the creation to error out (for all cases of us being unable).

alpeb:
> @alpeb SGTM either way. I am curious as to why must they behave the same?

Just for consistency. I don't see why having automountServiceAccountToken being true would be regarded as an error while the other cases just warnings.

> @alpeb I'm not sure I understand. If someone's explicitly asked for injection and we're unable to do it, I would like for the creation to error out (for all cases of us being unable).

Ok no problem. I just wanted to point that's a change in behavior. @mayankshah1607 I think that warrants a separate PR :wink:; if `Injectable()` returns false, return the appropriate error instead of just `bytes, reports, nil`. That might touch some error fixtures as well.

mayankshah1607:
Thanks @alpeb ! Any changes to be made to this PR? As mentioned earlier, when `automountServiceAccountToken: false` is set on a pod spec, `inject` simply errors out.
I'll open a follow up PR with necessary changes to ensure consistency in the behaviour of inject, i.e for hostNetworking, sidecar, etc.

adleong:
>If someone's explicitly asked for injection and we're unable to do it, I would like for the creation to error out (for all cases of us being unable).

@grampelberg would you mind creating an issue for this ^^.  Sounds like that's going to get tackled in a follow-up PR.

grampelberg:
@mayankshah1607 please create a followup issue for injection to fail in all cases we've been asked to do it and are unable.

PR #4154, opened 2020-03-09
update helm chart README with enforcedHostRegexp and controllerImageV…

lundbird:
Some variables that are available in the template but not in the readme

### DIFF
--- a/charts/linkerd2/README.md
+++ b/charts/linkerd2/README.md
@@ -85,6 +85,7 @@ The following table lists the configurable parameters of the Linkerd2 chart and 
 | Parameter                             | Description                                                                                                                                                                           | Default                              |
 |:--------------------------------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:-------------------------------------|
 | `controllerImage`                     | Docker image for the controller, tap and identity components                                                                                                                          | `gcr.io/linkerd-io/controller`       |
+| `controllerImageVersion`              | Tag for the controller container docker image                                                                                                                                         | latest version                       |
 | `controllerLogLevel`                  | Log level for the control plane components                                                                                                                                            | `info`                               |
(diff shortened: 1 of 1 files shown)


grampelberg:
@alpeb @kleimkuhler WDYT about getting something that dumps these from the helm chart and auto updates like the CLI?

alpeb:
> @alpeb @kleimkuhler WDYT about getting something that dumps these from the helm chart and auto updates like the CLI?

Definitely. Some Googling revealed [helm-docs](https://github.com/norwoodj/helm-docs) precisely for this, plus forces us to put descriptive messages into the values.yml file :+1:

alpeb:
I've created #4156 to track the automation.

charts/linkerd2/README.md
```diff
@@ -154,6 +155,7 @@ The following table lists the configurable parameters of the Linkerd2 chart and
 | `tap.keyPEM`                          | Certificate key for Tap component. If not provided then Helm will generate one.                                                                                                                                             ||
 | `webhookFailurePolicy`                | Failure policy for the proxy injector                                                                                                                                                 | `Ignore`                             |
 | `webImage`                            | Docker image for the web container                                                                                                                                                    | `gcr.io/linkerd-io/web`              |
+| `enforcedHostRegexp`                  | Host header validation regex for the dashboard. See the [Linkerd documentation](https://linkerd.io/2/tasks/exposing-dashboard) for more information                                   | `gcr.io/linkerd-io/web`              |
```
alpeb: The actual default value for this one is the empty string, which allows requests for `localhost`, `127.0.0.1`, `linkerd-web.linkerd.svc.cluster.local` and `linkerd-web.linkerd.svc` (in the latter two, the system will check for the actual namespace and cluster domain configured for this installation).
It would be great you could convey these details in the description of this entry :wink:

lundbird: Thanks for the catch. I updated the default value to an empty string.

PR #4157, opened 2020-03-09
Add alpha clients command

adleong:
We add the `linkerd alpha clients` command which displays client side metrics from each of a resource's clients.  This allows you to see who all of your clients are and see what your resource's metrics look like from your clients' point of view.  Since these metrics are measured on the client-side, they include network latency.

```
> linkerd alpha clients deploy/web -n emojivoto
FROM                TO   SUCCESS        RPS  LATENCY_P50  LATENCY_P90  LATENCY_P99
vote-bot.emojivoto  web   97.50%     2.0rps          4ms          5ms          5ms
```


adleong:
Note: due to a bug in the smi-metrics API controller, clients in different namespaces are not shown.  Once this bug is corrected in the API controller, clients from other namespaces will show up here.

cli/cmd/alpha_clients.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/spf13/cobra"
+)
+
+type alphaClientsOptions struct {
+	namespace string
+}
+
+func newCmdAlphaClients() *cobra.Command {
+	options := alphaClientsOptions{
+		namespace: "default",
```
zaharidichev: probably better to use `corev1.NamespaceDefault` instead

cli/cmd/alpha_clients.go
```diff
@@ -0,0 +1,79 @@
...
+  * deployments
+  * jobs
+  * namespaces
+  * pods
+  * replicasets
+  * replicationcontrollers
+  * statefulsets
+
+linkerd alpha clients will return a table of the requested resource's clients
+showing traffic metrics metrics to the requested resource such as request rate,
```
zaharidichev: `traffic metrics metrics` - is that right ?

PR #4159, opened 2020-03-09
Upload artifacts for all PRs

kleimkuhler:
## Motivation

#4147 adds a script for setting up a local cluster that uses the images built
from the changes introduced in a forked PR. This would be useful for all PRs.

In order to install Linkerd from a PR into a local cluster, the images still
need to be built at some point. If you happen to have SSH config setup for our
Packet host, you can pull them from there. That is not very
accessible--requiring that someone adds you as a user--so we can take a
similar approach to forked PRs.

## Solution

All PRs now make an artifact directory that is uploaded as part of the KinD
integration tests. This way, the `install-pr` script can use those images no
matter if the PR is a fork or not.

### DIFF
--- a/.github/workflows/kind_integration.yml
+++ b/.github/workflows/kind_integration.yml
@@ -38,8 +38,7 @@ jobs:
       run: |
         export PATH="`pwd`/bin:$PATH"
         bin/docker-build
-    - name: Create artifact with CLI and image archives (Forked repositories)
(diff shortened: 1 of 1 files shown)


kleimkuhler:
@adleong Yea should be cool!

It does add some time to a full CI run. The build and upload step only happens once per run, and based off this CI run it added about 1 minute and 45 seconds for both the `docker save` and `artifact-upload` steps.

I think that's a good trade off to be able to pull those artifacts and spin up local clusters immediately.

PR #4160, opened 2020-03-10
Set auth override

zaharidichev:
The changes here use the change in proxy-api to set the AuthorityOverride when that is present on the endpoints annotation. Additionally, it enables the cluster watcher to set the gateway identity data on the endpoints that it creates. This enalbes authority rewrite in the proxy as well as end to end identitiy.


go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/huandu/xstrings v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.7
 	github.com/julienschmidt/httprouter v1.2.0
-	github.com/linkerd/linkerd2-proxy-api v0.1.10
+	github.com/linkerd/linkerd2-proxy-api v0.1.12-0.20200309112816-75a7b3b20597
```
adleong: We should get the API change released before merging this.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -465,10 +466,16 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) Addre
 					Namespace: endpoints.ObjectMeta.Namespace,
 				}
 
+				var authorityOverride string
+				if fqName, ok := endpoints.Annotations[consts.RemoteServiceFqName]; ok {
+					authorityOverride = fmt.Sprintf("%s:%d", fqName, resolvedPort)
```
adleong: is this port correct?  I think `resolvedPort` is the gateway port.  I think what we actually want is the original source port from the request.

zaharidichev: hmm.. good question. Ok, correct me if I am wrong here but here is what I think: 

if we have a service such as: 

```
apiVersion: v1
kind: Service
metadata:
  name: some-service
spec:
  selector:
    app: some-service
  ports:
  - name: p1
    port: 8080
    targetPort: 8080
  - name: p2
    port: 8081
    targetPort: 8081
```

Then in this situation our target ports (this is the target port of the portPublisher type, not the endpoint) will e either 8080 or 8081. And this is ultimatelly what we want here. The real question however is, what happens if the port and the target port are different... like:

```
  - name: p1
    port: 8080
    targetPort: 9999
```

I think in this situation we want the port not the target port because... when this request hits the gateway we want to use this info to route to the service on the other cluster at 8080, not at 9999 (because there might not be 9999 at all). So I think I need to change that somehow to get the port not the targetPort. But to answer your original question, this will not be the port of the gateway. WDYT ?

adleong: Yes, I think that's right.  But to put it more simply: I think we just want to preserve the port from the request.  In other words, the port (not the target port).

zaharidichev: yes, that is correct. Will change it to account for that.

controller/api/destination/endpoint_translator.go
```diff
@@ -49,12 +49,21 @@ func (et *endpointTranslator) Add(set watcher.AddressSet) {
 		if address.Pod != nil {
 			wa, err = et.toWeightedAddr(address)
 		} else {
+
```
kleimkuhler: nit: Extra space

zaharidichev:
This.. depends on https://github.com/linkerd/linkerd2-proxy-api/pull/38 which should come with a new release for the proxy api

PR #4166, opened 2020-03-12
Update maintainers and codeowners

grampelberg:
There are a bunch of awesome maintainers that haven't been called out yet! This adds everyone who's been doing awesome work on the project to both CODEOWNERS and MAINTAINERS.md

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@ CHANGES.md @wmorgan @klingerf @olix0r @admc
 GOVERNANCE.md @klingerf @olix0r
 MAINTAINERS.md @klingerf @olix0r
 
-/.github/workflows/* @alpeb @kleimkuhler
+/.github/workflows/ @alpeb @kleimkuhler
 /charts @alpeb @zaharidichev
 
 /charts/linkerd2-cni @hemakl @zaharidichev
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -1,3 +1,25 @@
 CHANGES.md @wmorgan @klingerf @olix0r @admc
 GOVERNANCE.md @klingerf @olix0r
 MAINTAINERS.md @klingerf @olix0r
+
+/.github/workflows/* @alpeb @kleimkuhler
```
klingerf: Just curious, does this path need the wildcard? It doesn't look like we're including it for any of the other paths.

grampelberg: Not AFAIK, good call!

kleimkuhler:
(Trying not to ping everyone in this comment so not using @)

A current issue I have noticed with this is the review requirement for certain PRs. For example, Alejandro and I have separate PRs fixing parts of `go.mod` that have approvals from two Linkerd organization members:
- #4170 
- #4173

These cannot currently be merged because *both* Zahari and Hema are required to leave an approval. While I think some sort of requirement is valuable, I don't know if this is very scalable--it feels more like a blocker right now.

kleimkuhler:
> These cannot currently be merged because both Zahari and Hema are required to leave an approval

It's just one not both. It still feels like there is something we can do differently here though, since both PRs have 2 approvals from other Linkerd organization members.

adleong:
Requiring approval from the owners of a directory to make changes to that directory seems like a good policy in general.  However, it seems like the problem here is that any changes to our go deps will touch the Dockerfiles in all sub-directories.  Perhaps we could work around this by having a separate group of owners for all of the Dockerfiles...

kleimkuhler:
#4179 tracks this specific issue discussed above

PR #4167, opened 2020-03-12
Add KinD option to `install-pr` script

kleimkuhler:
## Motivation

After #4147 added the `install-pr` script, installing PRs into existing
clusters does not work if that cluster is a KinD cluster

Changing the script to be able to use KinD, and specifically automate `kind
load` would be helpful!

## Solution

The script can now be used in the following ways.

```
❯ bin/install-pr --help
Install Linkerd with the changes made in a GitHub Pull Request.

Usage:
    --context: The name of the kubeconfig context to use

    # Install Linkerd into the current cluster
    bin/install-pr 1234

    # Install Linkerd into the current KinD cluster
    bin/install-pr [-k|--kind] 1234

    # Install Linkerd into the 'kind-pr-1234' KinD cluster
    bin/install-pr [-k|--kind] --context kind-pr-1234 1234
```

The script assumes that the cluster (KinD or not) has already been created. If
the cluster is a KinD cluster, the `-k|--kind` flag should be passed.

If the `--context` flag is not passsed, the install defaults to the current
context (`kubectl config current-context`).

I also added a [`-h|--help]` option that describes how to use the script.

### DIFF
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -113,8 +113,8 @@ fi
 
 # Get the URL for downloading the artifacts archive.
 auth="Authorization: token $GITHUB_TOKEN"
-branch=$(curl -sL "$linkerd2_pulls_url/$pr" | jq -r '.head.ref')
-artifacts=$(curl -sL "$linkerd2_kind_integration_url/runs?branch=$branch" | jq -r '.workflow_runs[0].artifacts_url')
+branch=$(curl -sL -H "$auth" "$linkerd2_pulls_url/$pr" | jq -r '.head.ref')
+artifacts=$(curl -sL -H "$auth" "$linkerd2_kind_integration_url/runs?branch=$branch" | jq -r '.workflow_runs[0].artifacts_url')
 archive=$(curl -sL -H "$auth" "$artifacts" | jq -r '.artifacts[0].archive_download_url')
 
 bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
@@ -147,7 +147,7 @@ docker cp "$container_id":/out/linkerd-"$platform" "$rootdir"/target/cli/"$platf
 installed_cli=1
 
 # Load images into local docker images
-if [ $is_kind ]
+if [ $is_kind -eq 1 ]
 then
   # Create new KinD cluster
   cluster_name="pr-$pr"
@@ -186,7 +186,7 @@ fi
 
 cd -
 
-if [ $is_kind ]
+if [ $is_kind -eq 1 ]
 then
   "$bindir"/linkerd "${context_flag[@]}" check --pre
   "$bindir"/linkerd "${context_flag[@]}" install | kubectl "${context_flag[@]}" apply -f -
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -141,8 +141,8 @@ echo "### Pre checks ###"
 
 # shellcheck disable=SC2086
 (
-  set -x"$linkerd"
-  $context_flag check --pre
+  set -x
+  "$linkerd" $context_flag check --pre
 )
 
 echo "### Installing $tag ###"
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -13,27 +13,25 @@ cleanup() {
   # Remove temporary directory.
   rm -rf "$archive_tmp"
   
-  if [ "$installed_cli" -eq 1 ]
+  if [ "$installed_cli" ]
   then
     echo ""
     echo "Linkerd CLI installed!"
     echo "Usage:"
     echo "    bin/linkerd $context_flag version"
-  elif [ "$installed_cli" -eq 0 ]
-  then
+  else [ "$installed_cli" ]
     echo ""
     echo "Linkerd CLI install failed!"
   fi
 
-  if [ "$linkerd_check" -eq 1 ]
+  if [ "$linkerd_check" ]
   then
     echo ""
     echo "'linkerd check' passed!"
     echo "Usage:"
     echo "    bin/linkerd ${context_flag[*]} check"
     echo "    kubectl ${context_flag[*]} -n linkerd get pods"
-  elif [ "$linkerd_check" -eq 0 ]
-  then
+  else
     echo ""
     echo "'linkerd check' failed!"
     echo "Check again before debugging:"
@@ -42,22 +40,20 @@ cleanup() {
     echo "    kubectl ${context_flag[*]} -n linkerd get pods"
   fi
 
-  if [[ $is_kind -eq 1 && $kind_setup -eq 0 ]]
+  if [[ $is_kind && $kind_setup ]]
   then
     echo "Script failed at or before KinD cluster creation."
   fi
 }
 
-trap cleanup EXIT
-
 # Initialize variables used throughout script.
 archive_tmp=
 context_flag=
 
-installed_cli=0
-kind_setup=0
-linkerd_check=0
-is_kind=0
+installed_cli=
+kind_setup=
+linkerd_check=
+is_kind=
 
 linkerd2_pulls_url="https://api.github.com/repos/linkerd/linkerd2/pulls"
 linkerd2_kind_integration_url="https://api.github.com/repos/linkerd/linkerd2/actions/workflows/kind_integration.yml"
@@ -79,8 +75,6 @@ do
       echo ""
       echo "    # Install Linkerd into a new KinD cluster"
       echo "    bin/install-pr [-k|--kind] ####"
-      installed_cli=2
-      linkerd_check=2
       exit 0
       ;;
     -k|--kind)
@@ -95,6 +89,8 @@ do
   esac
   shift
 done
+
+trap cleanup EXIT
 
 pr=$1
 
@@ -147,7 +143,7 @@ docker cp "$container_id":/out/linkerd-"$platform" "$rootdir"/target/cli/"$platf
 installed_cli=1
 
 # Load images into local docker images
-if [ $is_kind -eq 1 ]
+if [ $is_kind ]
 then
   # Create new KinD cluster
   cluster_name="pr-$pr"
@@ -186,7 +182,7 @@ fi
 
 cd -
 
-if [ $is_kind -eq 1 ]
+if [ $is_kind ]
 then
   "$bindir"/linkerd "${context_flag[@]}" check --pre
   "$bindir"/linkerd "${context_flag[@]}" install | kubectl "${context_flag[@]}" apply -f -
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -9,63 +9,189 @@
 
 set -eo pipefail
 
+cleanup() {
+  # Remove temporary directory.
+  rm -rf "$archive_tmp"
+  
+  if [ "$installed_cli" ]
+  then
+    echo ""
+    echo "Linkerd CLI installed!"
+    echo "Usage:"
+    echo "    bin/linkerd ${context_flag[@]} version"
+  else
+    echo ""
+    echo "Linkerd CLI install failed!"
+  fi
+
+  if [ "$linkerd_check" ]
+  then
+    echo ""
+    echo "'linkerd check' passed!"
+    echo "Usage:"
+    echo "    bin/linkerd ${context_flag[@]} check"
+    echo "    kubectl ${context_flag[@]} -n linkerd get pods"
+  else
+    echo ""
+    echo "'linkerd check' failed!"
+    echo "Check again before debugging:"
+    echo "    bin/linkerd ${context_flag[@]} check --pre"
+    echo "    bin/linkerd ${context_flag[@]} check"
+    echo "    kubectl ${context_flag[@]} -n linkerd get pods"
+  fi
+
+  if [[ $is_kind && ! $kind_setup ]]
+  then
+    echo "Script failed at or before KinD cluster creation."
+  fi
+}
+
+trap cleanup EXIT
+
+# Initialize variables used throughout script.
+archive_tmp=
+context_flag=
+
+installed_cli=
+kind_setup=
+linkerd_check=
+is_kind=
+
+linkerd2_pulls_url="https://api.github.com/repos/linkerd/linkerd2/pulls"
+linkerd2_kind_integration_url="https://api.github.com/repos/linkerd/linkerd2/actions/workflows/kind_integration.yml"
+
+# Read script flags and arguments.
+while :
+do
+  case $1 in
+    -h|--help)
+      echo "Install Linkerd with the changes made in a GitHub Pull Request."
+      echo ""
+      echo "Note:"
+      echo "    If installing into an existing Kubernetes cluster, this script"
+      echo "    assumes the cluster is not a KinD cluster."
+      echo ""
+      echo "Usage:"
+      echo "    # Install Linkerd into the 'kubectl' current cluster"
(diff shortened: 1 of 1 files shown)


kleimkuhler:
As an example of the current error messaging, here is an example of `linkerd check` failing on a KinD cluster that just needed a little more time to reach a good state:

```
❯ bin/install-pr --kind 4155
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  336M    0  336M    0     0  5844k      0 --:--:--  0:00:59 --:--:-- 8350k

..

..

× no unschedulable pods
    linkerd-controller-f78f69cf4-rdkbl: 0/1 nodes are available: 1 node(s) had taints that the pod didn't tolerate.
    linkerd-identity-5fdfbcf98d-b6hnf: 0/1 nodes are available: 1 node(s) had taints that the pod didn't tolerate.
    see https://linkerd.io/checks/#l5d-existence-unschedulable-pods for hints

Status check results are ×

Linkerd CLI installed!
Usage:
    bin/linkerd --context version

'linkerd check' failed!
Check again before debugging:
    bin/linkerd --context kind-pr-4155 check --pre
    bin/linkerd --context kind-pr-4155 check
    kubectl --context kind-pr-4155 -n linkerd get pods
```

It explains the Linkerd CLI was successfully installed, but `linkerd check` failed. As suggested, rerunning `bin/linkerd --context kind-pr-4155 check` did succeed the second time.

adleong:
Now that we're building artifacts for all PRs, shouldn't this already work to install PRs from non-forked repos?  Or what change here is necessary to make that work?

bin/install-pr
```diff
@@ -9,63 +9,193 @@
 
 set -eo pipefail
 
+cleanup() {
```
adleong: this highly condition cleanup function is difficult to follow and feels like it will be annoying to maintain.  We also may not want to delete the tmp directory in the case of failures, so that we can debug.  Can we just put the messaging inline where appropriate?  We can check exit codes as we go if necessary.

bin/install-pr
```diff
@@ -9,63 +9,193 @@
...
 then
   # shellcheck disable=SC2016
-  echo 'Generate a personal access token at https://github.com/settings/tokens and set it in the $GITHUB_TOKEN env var'
+  echo 'Error: Generate a personal access token at https://github.com/settings/tokens and set it in the $GITHUB_TOKEN env var'
   exit 1
 fi
 
+# Get the URL for downloading the artifacts archive.
 auth="Authorization: token $GITHUB_TOKEN"
+branch=$(curl -sL "$linkerd2_pulls_url/$pr" | jq -r '.head.ref')
```
adleong: Is it not necessary to add the auth header for these API requests?  Even if it's not required, it might still be a good idea?  We might get higher rate-limits, for example.

bin/install-pr
```diff
@@ -9,63 +9,193 @@
...
-
-for archive in cli-bin cni-plugin controller debug grafana proxy web
-do 
-  image=$(docker load -i "$archive.tar" | sed "s/Loaded image: //")
-  docker push "$image"
-  tag=$(echo "$image" | cut -f 2 -d ":")
-done
+# Install the Linkerd CLI.
+image=$(docker load -i cli-bin.tar | sed 's/Loaded image: //')
+container_id=$(docker create "$image")
```
adleong: We already have `bin/docker-pull-binaries`, why re-create the wheel here?

bin/install-pr
```diff
@@ -9,63 +9,193 @@
...
+    exit 1
+  fi
+  
+  "$bindir"/kind create cluster --name "$cluster_name"
+
+  # Load images into cluster
+  for image in cni-plugin controller debug grafana proxy web
+  do
+    echo "Loading $image image into cluster.."
+    "$bindir"/kind load image-archive --name "$cluster_name" $image.tar || tee load_fail &
```
adleong: how long does it take to load each of these images into kind?  this parallelism machinery is really scary to me.

bin/install-pr
```diff
@@ -9,63 +9,193 @@
...
-echo "### Pre checks ###"
-
-$linkerd check --pre
-
-echo "### Installing $tag ###"
-
-$linkerd install | kubectl apply -f -
-
-$linkerd check
+if [ $is_kind ]
```
adleong: if  the context flag is just empty for the non-kind case, I think we can remove this conditional

kleimkuhler:
@adleong Yes sorry that was taken care of in #4159. That should already work.

bin/install-pr
kleimkuhler: `bin/docker-pull-binaries` doesn't point to the same location as `bin/linkerd`. Maybe those scripts can be merged so that there is only one location where binaries are installed, but I didn't think changing that should be part of this PR.

Ultimately, after this script runs and installs Linkerd, I would expect to use `bin/linkerd` and have it be the version that was installed.

adleong: we have philosophical differences then :) I expect `bin/linkerd` to build or run the code in my working tree.

bin/install-pr
kleimkuhler: I took this from [kind_integration.yml](https://github.com/linkerd/linkerd2/blob/88cafa36c6683e59a22c3cb4aa3cd5eda6c9a0fe/.github/workflows/kind_integration.yml#L144-L145) where we do the same thing.

I think when I added that it saved about a minute in CI. I think it's valuable for quickly getting an install up and running.

bin/install-pr
kleimkuhler: Yep it's not necessary but I can add it back!

bin/install-pr
kleimkuhler: Ah yep! When using this script I would expect `bin/linkerd` to point to a binary that I'm interested in testing, so I was thinking that this script should therefore put the binary there.

I don't really use the `release/` path in my workflow that much.. pretty much `bin/linkerd` or `linkerd` (the currently installed edge or stable).

I can change that back if that was the intention of two different binary locations, but it does feel a little awkward? If I install a binary that I want to test, it feels odd to have to add or type out the release path location instead of using `bin/linkerd`.

bin/install-pr
kleimkuhler: Unfortunately this is like this for a reason. If `context_flag` is empty, it's passed as an empty string option `""`. I have these separated so that that is not an issue.

If you don't quote the `${context_flag[@]}` it "works", but then `shellcheck` fails with an error.

bin/install-pr
kleimkuhler: It's not the prettiest, but it also does a good job at explaining if the script fails where it failed and what to do.

I can remove it, but it also then becomes harder to figure out where you have to manually continue the script.

Is there anything I can do that would give you less concern? In testing this, it was helpful knowing that the CLI was installed but `check` failed for some (usually Kubernetes flakiness) reason. Running `linkerd check` again after waiting a few moments would succeed.

adleong: My point is that we can print appropriate output as we go rather than tracking state and then deferring into this cleanup function.  I'll also point out that this is a script intended for developers so that shifts the priorities towards maintainability and away from fool-proof-ness, somewhat.  Personally, when I see the failure output from `linkerd check`, it's clear to me what happened (linkerd check failed) without a separate message explaining it.

bin/install-pr
adleong: I'd prefer to have `bin/linkerd` keep its current behavior of building or running the current working tree.

bin/install-pr
adleong: I took a stab at trying to clean this up, but to no avail.  Bash is insane.  We may want to just add a shellcheck exclusion for this since we know we don't have to worry about re-globbing in this case.

bin/install-pr
kleimkuhler: Sounds good!

There will still be a little bit of a dance for choosing the binary since right now this always uses the Darwin binary with `head -n 1`. This script doesn't currently work on Linux. I will keep the `case $(uname)` but go off the `release/` path.

bin/install-pr
kleimkuhler: Yep that makes sense. It's much more maintainable without `cleanup()` and the intermediate variables it was using.

bin/install-pr
kleimkuhler: Disabled these `shellcheck` checks 👍

bin/install-pr
```diff
@@ -9,63 +9,138 @@
...
+
+# Install the Linkerd CLI
+linkerd=$("$bindir"/docker-pull-binaries "$tag" | awk -v platform=$platform '$0 ~ platform')
+
+# shellcheck disable=SC2068
+"$linkerd" ${context_flag[@]} check --pre
+# shellcheck disable=SC2068
+"$linkerd" ${context_flag[@]} install | kubectl ${context_flag[@]} apply -f -
+# shellcheck disable=SC2068
+"$linkerd" ${context_flag[@]} check
```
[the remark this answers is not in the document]
kleimkuhler: Yep I can add that!

bin/install-pr
```diff
@@ -9,63 +9,138 @@
...
-
-curl -L -o archive.zip -H "$auth" "$archive_url"
-
+curl -L -o archive.zip -H "$auth" "$archive"
 unzip -o archive.zip
 
-echo "### Pushing images ###"
+# Push the Linkerd CLI
+image=$(docker load -i cli-bin.tar | sed 's/Loaded image: //')
+docker push "$image"
```
adleong: do we actually need to push the cli-bin image?

bin/install-pr
```diff
@@ -9,63 +9,138 @@
...
+    exit 1
+  fi
+  
+  "$bindir"/kind create cluster --name "$cluster_name"
+
+  # Load images into cluster
+  for image in cni-plugin controller debug grafana proxy web
+  do
+    echo "Loading $image image into cluster.."
+    "$bindir"/kind load image-archive --name "$cluster_name" $image.tar || tee load_fail &
```
adleong: why `tee` here? should this be `touch`?

bin/install-pr
```diff
@@ -9,63 +9,138 @@
...
+  wait < <(jobs -p)
+  if [ -f load_fail ]
+  then
+    echo "Loading docker images into KinD cluster failed."
+    exit 1
+  fi
+else
+  for image in cni-plugin controller debug grafana proxy web
+  do
+    docker load -i "$image.tar"
```
adleong: don't we need to push these docker images?

kleimkuhler:
@adleong @alpeb Thanks for the reviews! Alex mentioned some things in a comment yesterday that made me think a little more about the the complexity this change is adding to the script.

I think I've taken the wrong approach about what this script should be doing, and I am going to remove the cluster creation for KinD.

I'm going to add support for a `--context` flag similar to `kubectl`/`linkerd` and have the script assume that if the `-k|--kind` flag is being passed, that the KinD cluster has already been created.

This means that if you create the default KinD cluster (name=kind), then `bin/install-pr -k 1234` will just install Linkerd into that default cluster.

If you create a KinD cluster (name=pr-1234), then `bin/install-pr -k --context kind-pr-1234 123` will install Linkerd into the `kind-pr-1234` cluster.

bin/install-pr
kleimkuhler: Nope I'll remove that!

bin/install-pr
kleimkuhler: Because the images are loaded into the local docker images, the install pulls them from there instead of needing to look in the registry.

bin/install-pr
```diff
@@ -9,63 +9,146 @@
...
 
 echo "### Installing $tag ###"
 
-$linkerd install | kubectl apply -f -
-
-$linkerd check
+# shellcheck disable=SC2086
+(set -x; "$linkerd" $context_flag install | kubectl $context_flag apply -f -)
+# shellcheck disable=SC2086
+(set -x; "$linkerd" $context_flag check)
```
alpeb: You can actually just do `set -x` one more time like:
```bash
(
  set -x
  "$linkerd" $context_flag check --pre
   # shellcheck disable=SC2086
  "$linkerd" $context_flag install | kubectl $context_flag apply -f -
   # shellcheck disable=SC2086
  "$linkerd" $context_flag check
)
```

alpeb:
Also regarding the `kind-load` parallel logic, I think it's worth, in a separate PR, to update `bin/kind-load` with that, and then consume it here and in the CI scripts.

bin/install-pr
kleimkuhler: I'd like to keep the `echo "### Installing $tag ###"`, but I can wrap the commands after in parens.

kleimkuhler:
> Also regarding the kind-load parallel logic, I think it's worth, in a separate PR, to update bin/kind-load with that, and then consume it here and in the CI scripts.

@alpeb Yep I totally agree about doing this in a separate change. Originally I was thinking about moving `kind-load` into a `kind_load()` function within a new `bin/_kind.sh`, similar to `bin/_docker.sh`/`bin/_tag.sh`.

It could be made more usable such as taking some arguments so that we can use it in more places such as this and the CI script.

kleimkuhler:
> That test doesn't have a retryDeadline configured, so there are no retries. Would it make sense to set that up? That would also help scenarios like when people install linkerd and immediately call linkerd check...

@alpeb I agree it makes sense to add that default `wait` option that some of the other tests have. I can follow up with a change for that.

bin/install-pr
adleong: if installing into a non-local (cloud) cluster, the images need to be pushed to a repository that the cluster can access

bin/install-pr
```diff
@@ -9,63 +9,152 @@
...
-  docker push "$image"
-  tag=$(echo "$image" | cut -f 2 -d ":")
-done
+image=$(docker load -i cli-bin.tar | sed 's/Loaded image: //')
+tag=$(echo "$image" | cut -f 2 -d ":")
+
+if [ $is_kind ]
+then
+  # KinD context strings are created by prepending `kind-` to the cluster name
+  # shellcheck disable=SC2086
```
adleong: why disable this check?

bin/install-pr
```diff
@@ -9,63 +9,152 @@
...
+  if [ $context ]
+  then
+    name=$(echo "$context" | sed -n -E "s/(kind)-(.*)/\2/p")
+    name_flag="--name $name"
+  fi
+
+  for image in cni-plugin controller debug grafana proxy web
+  do
+    # shellcheck disable=SC2086
```
adleong: why disable this check?

adleong: I just learned about the `${var:+..}` syntax.  can that help here? http://mywiki.wooledge.org/BashFAQ/050#I_only_want_to_pass_options_if_the_runtime_data_needs_them

bin/install-pr
kleimkuhler: Ah that's right yea I was thinking about local clusters.

We should only need to `docker push` for the non-KinD case so I'll add it back for that.

bin/install-pr
kleimkuhler: I can remove this.

bin/install-pr
kleimkuhler: This is super helpful. It cleans up all the uses of these optional arguments

PR #4170, opened 2020-03-13
Bump proxy-init to v1.3.2

alpeb:
Bumped `proxy-init` version to v1.3.2, fixing an issue with `go.mod`
(linkerd/linkerd2-proxy-init#9).
This is a non-user-facing fix.


grampelberg:
@zaharidichev @hemakl this ends up touching code that you're owners for. Would you mind getting a review in so we can merge?

PR #4177, opened 2020-03-17
Add Jaeger links to the Linkerd dashboard

Pothulapati:
This PR
- Adds support for reverse proxying jaeger through linkerd-web, allowing users to access jaeger UI at `<linkerd-web>/jaeger`


Pothulapati:
jaeger links will be available in the dashboard as seen in the below image
Screenshot_2020-04-04 Linkerd

Pothulapati:
To try this PR out:

- Let `config.yaml` be 
```
tracing:
  enabled: true
```
- Install Linkerd by running
`linkerd install --control-plane-tracing --addon-config config.yaml | k apply -f -`

- Open `linkerd-web` dashboard and checkout the `jaeger` links in the MetricsTable of any component.

grampelberg:
@Pothulapati do we have the addons documented anywhere on the website?

web/app/js/components/MetricsTable.jsx
```diff
@@ -110,7 +111,7 @@ const trafficSplitDetailColumns = [
   },
 ];
 
-const columnDefinitions = (resource, showNamespaceColumn, showNameColumn, PrefixedLink, isTcpTable) => {
+const columnDefinitions = (resource, showNamespaceColumn, showNameColumn, PrefixedLink, isTcpTable, jaeger) => {
```
grampelberg, line 114: Maybe `showJaeger`?

web/app/js/components/MetricsTable.test.js
```diff
@@ -26,7 +26,7 @@ describe('Tests for <MetricsTable>', () => {
 
     expect(table).toBeDefined();
     expect(table.props().tableRows).toHaveLength(1);
-    expect(table.props().tableColumns).toHaveLength(8);
+    expect(table.props().tableColumns).toHaveLength(9);
```
grampelberg, line 29: Should there be a negative test added? eg. to test when `jaeger=false`?

web/srv/grafana_proxy.go
```diff
@@ -6,19 +6,19 @@ import (
...
-type grafanaProxy struct {
+// reverseProxy is an HTTP reverse proxy that forwards all web requests
+// containing paths prefixed  to the corresponding service. The proxy
+// strips the prefix and rewrites the Host header before sending.
+type reverseProxy struct {
 	*httputil.ReverseProxy
 }
 
-func newGrafanaProxy(addr string) *grafanaProxy {
+func newReverseProxy(addr string, prefix string) *reverseProxy {
```
grampelberg, line 16: Should this file get renamed to `reverse_proxy.go`?

web/srv/server.go
```diff
@@ -197,6 +201,15 @@ func NewServer(
 	server.router.POST("/grafana/*grafanapath", handler.handleGrafana)
 	server.router.PUT("/grafana/*grafanapath", handler.handleGrafana)
 
+	// jaeger proxy
```
grampelberg, line 204: Should there be a helper function that manages the paths for both grafana and jaeger? It is a minor optimization, my DRY senses are tingling though.

web/templates/app.tmpl.html
```diff
@@ -3,7 +3,8 @@
     data-release-version="{{.Data.ReleaseVersion}}"
     data-go-version="{{.Data.GoVersion}}"
     data-controller-namespace="{{.ControllerNamespace}}"
-    data-uuid="{{.UUID}}">
+    data-uuid="{{.UUID}}"
+    data-jaeger="{{.Jaeger}}">
```
grampelberg, line 7: Is this a `jaegerEnabled` variable?

web/app/js/components/JaegerLink.jsx
```diff
@@ -0,0 +1,37 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import _isEmpty from 'lodash/isEmpty';
+import { jaegerIcon } from './util/SvgWrappers.jsx';
+
+const JaegerLink = ({ PrefixedLink, name, namespace, resource }) => {
+  let link = '/jaeger/search?service=linkerd-proxy&tags=';
+  if (_isEmpty(namespace)) {
```
grampelberg, line 8: Minor optimization again, but WDYT about moving this into a function? I'm thinking that `to={baseLink + getQuery()` is a little more readable.

Pothulapati:
@grampelberg I have some documentation here https://github.com/linkerd/website/pull/686 specific to tracing add-on! WDYT?

web/templates/app.tmpl.html
Pothulapati: the jaeger url variable? As we need the concrete URl to use when routing. Did I understand your question correctly?

grampelberg: The variable name doesn't tell me what it is. This is the `jaegerURI` then?

grampelberg:
@Pothulapati yup, that's what I was looking for. Thanks!

web/templates/app.tmpl.html
Pothulapati: Ah, Right having a `showJaeger`makes sense, as we don't use the url anyway (because we directly proxy), but React is acting weird to pass boolean parameters. It doesn't work, and the internet agrees that its a thing. I could be wrong here! 

I can take another look if its important!

web/app/js/components/MetricsTable.test.js
Pothulapati: Actually I made the jaeger flag optional now, and added additional test for `jaeger=true`

web/srv/server.go
Pothulapati: Yep, Done!

web/app/js/components/JaegerLink.jsx
Pothulapati: done!

web/templates/app.tmpl.html
grampelberg: I don't actually care if it is a boolean, making it a string of "true" would work. It is more that that variable says what it does =)

web/app/js/components/MetricsTable.jsx
```diff
@@ -251,6 +277,7 @@ MetricsTable.propTypes = {
   metrics: PropTypes.arrayOf(processedMetricsPropType),
   resource: PropTypes.string.isRequired,
   selectedNamespace: PropTypes.string.isRequired,
+  jaeger: PropTypes.string,
```
Pothulapati, line 280: There is a bit of confusion here as `grafana` is marked as `isRequired` while `jaeger` is not. As both of these values are passed as empty strings when the arg is empty, This should not cause any problems!


The change would be to remove `isRequired` for `grafana` and update the tests as they expect grafana field to be there always.

Pothulapati: raised https://github.com/linkerd/linkerd2/issues/4349 to track

grampelberg:
@Pothulapati would you mind getting an issue to track the docs updates? @adleong has some great points!

Pothulapati:
@grampelberg @adleong 

Created https://github.com/linkerd/website/issues/720 & https://github.com/linkerd/linkerd2/issues/4351 to track them :)

grampelberg:
@zaharidichev mind giving this a review? It appears that you're a unique code owner snowflake for files that I'm not 100% sure of =)

PR #4178, opened 2020-03-17
Multicluster dashboard for traffic metrics

zaharidichev:
This change adds labels to endpoints that target remote services. It also adds a Grafana dashboard that can be used to monitor multicluster traffic. (first time making a dashboard so suggestions are welcome).

Screenshot 2020-03-17 at 17 34 34

### DIFF
--- a/grafana/dashboards/multicluster.json
+++ b/grafana/dashboards/multicluster.json
@@ -98,7 +98,7 @@
             "tableColumn": "",
             "targets": [
                 {
-                    "expr": "sum(irate(response_total{classification=\"success\", dst_remote_gateway=\"$gateway\", dst_remote_cluster=\"$cluster\", direction=\"outbound\"}[30s])) / sum(irate(response_total{dst_remote_gateway=\"$gateway\", dst_remote_cluster=\"$cluster\", direction=\"outbound\"}[30s]))",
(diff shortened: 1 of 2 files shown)


grafana/dashboards/multicluster.json
```diff
@@ -0,0 +1,968 @@
...
+            "pointradius": 5,
+            "points": false,
+            "renderer": "flot",
+            "seriesOverrides": [],
+            "spaceLength": 10,
+            "stack": false,
+            "steppedLine": false,
+            "targets": [
+                {
+                    "expr": "histogram_quantile(0.95, sum(rate(response_latency_ms_bucket{dst_remote_gateway=\"$gateway\", dst_remote_gateway=\"\", dst_remote_cluster=\"$cluster\", dst_remote_cluster=\"\", direction=\"outbound\"}[30s])) by (le, dst_remote_service))",
```
alpeb: Using `=` instead of `!=` for the empty matches, which is forbidding the graph to show data.

grafana/dashboards/multicluster.json
```diff
@@ -0,0 +1,968 @@
...
+                    "show": true
+                }
+            ],
+            "yaxis": {
+                "align": false,
+                "alignLevel": null
+            }
+        },
+        {
+            "content": "<div class=\"text-center dashboard-header\">\n  <span>INBOUND TRAFFIC BY REMOTE SERVICE</span>\n</div>",
```
alpeb: Nit: I'm not sure if it's 100% accurate to call this inbound traffic... from what I can see, the metrics on these queries are gathered on the clients calling these mirrored services, where `direction=outbound`. WDYT on calling this just "TRAFFIC BY REMOTE SERVICE"? (Also those services might have other incoming traffic not originated in the current cluster).

PR #4181, opened 2020-03-17
Upgrade golangci-lint to v1.23.8

alpeb:
This should help with some timeouts we're seeing in CI.

I fixed some new warnings found in `inject.go` and `uninject.go`.
Also we now have to explicitly disable linting `/controller/gen`.

The linter was also complaining that in `/pkg/k8s/fake.go` the
`spClient.Interface` and `tsclient.Interface` returned in the function
`newFakeClientSetsFromManifests()` aren't used, but I opted to ignore
that to leave them available for future tests.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,8 @@
 
 run:
   deadline: 5m
+  skip-dirs:
+  - controller/gen
 issues:
   exclude-use-default: false
 linters:
--- a/bin/lint
+++ b/bin/lint
@@ -2,7 +2,7 @@
 
 set -eu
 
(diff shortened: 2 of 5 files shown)


alpeb:
CI memory exploded, so I'm downgrading a bit to v1.23.0.
Others are experiencing the same issue: golangci/golangci-lint#994

alpeb:
Ok last run was saner:
```
level=info msg="Memory: 749 samples, avg is 262.2MB, max is 2567.1MB"
level=info msg="Execution took 1m23.27217011s"
```

PR #4187, opened 2020-03-19
Update changelog for edge-20.3.3

hawkw:
## edge-20.3.3

This release introduces new experimental CLI commands for querying metrics using
the Service Mesh Interface (SMI) and for multi-cluster support via service
mirroring.

If you would like to learn more about service mirroring or SMI, or are
interested in experimenting with these features, please join us in
[Linkerd Slack](https://slack.linkerd.io) for help and feedback.

* CLI
  * Added experimental `linkerd cluster` commands for managing multi-cluster
    service mirroring
  * Added the experimental `linkerd alpha clients` command, which uses the
    smi-metrics API to display client-side metrics from each of a resource's
    clients
  * Added retries to some `linkerd check` checks to prevent spurious failures
    when run immediately after cluster creation or Linkerd installation

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,22 @@
+## edge-20.3.3
+
+This release introduces new experimental CLI commands for querying metrics using
+the Service Mesh Interface (SMI) and for multi-cluster support via service
+mirroring.
+
+If you would like to learn more about service mirroring or SMI, or are
+interested in experimenting with these features, please join us in
+[Linkerd Slack](https://slack.linkerd.io) for help and feedback.
+
+* CLI
+  * Added experimental `linkerd cluster` commands for managing multi-cluster
+    service mirroring
+  * Added the experimental `linkerd alpha clients` command, which uses the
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,22 @@
+## edge-20.3.3
+
+This release introduces new experimental CLI commands for querying metrics using
+the Service Mesh Interface (SMI) and for multi-cluster support via service
```
kleimkuhler: Maybe use `.. as well as ..` instead of `.. and for ..`?

hawkw: hmm, i'm not sure..."new experimental CLI commands for querying ... as well as multi-cluster support ..." makes it sound like "multi-cluster support via service mirroring" is added in this release, rather than "_new experimental CLI commands for_ multi-cluster support via service mirroring"

kleimkuhler: Yep I see what you mean. Reading through it the first time caught my attention, but definitely doesn’t need to be changed.

PR #4192, opened 2020-03-23
Add more owners to Dockerfiles

### LINKED ISSUE #4179: Changes to Go dependencies require approval from all code owners
## Motivation

When a change requires updating Go dependencies, the Dockerfiles in all
sub-directories are changed to have new dependency SHAs.

This is an issue because the code owner for each sub-directory is required to
approve the change, even if two other code owners for other sub-directories
have already left approvals.

Current examples:
- #4170
- #4173

## Solution

@adleong proposed one option of  making a new group of code owners for all of
the Dockerfiles so that changes like the above don't require approvals from
every code owner.

adleong:
Fixes #4179 

Changes to Go dependencies will touch all Dockerfiles in the repo which requires approval from the codeowners of each subdirectory.

We revise the codeowners to add more owners for the Dockerfiles so that approval is not required from the subdirectory owners specifically.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,6 +16,8 @@ MAINTAINERS.md @klingerf @olix0r
 
 /pkg @adleong @alpeb
 
+Dockerfile* @adleong @alpeb @klingerf @olix0r @siggy
+
 # These do not have clear owners today
 # /grafana
 # /proto


kleimkuhler:
@adleong It's okay. You'll get through this.

PR #4193, opened 2020-03-23
Add missing PSP for linkerd-smi-metrics

alpeb:
The linkerd-smi-metrics ServiceAccount wasn't hooked into linkerd's PSP
resource, which resulted in the linkerd-smi-metrics ReplicaSet failing
to spawn pods:

```
Error creating: pods "linkerd-smi-metrics-574f57ffd4-" is forbidden:
unable to validate against any pod security policy: []
```

Edit: Obviously, this only happens in PSP-enabled clusters...


alpeb:
> should there be some kind of automated test for this?

It seems KinD doesn't play very well with PSP, but I'll investigate more.

PR #4199, opened 2020-03-25
charts: Using downwardAPI to mount labels to the proxy container

### LINKED ISSUE #4008: Owner Metadata for Spans.
## Feature Request

Right Now, For metrics Prometheus automatically adds the pod's labels as metric labels, giving us the capability to link grafana dashboard for that component directly in the linkerd dashboard and allowing users to jump there seamlessly.

A similar model with Jaeger would be great i.e jumping directly from the linkerd dashboard to the jaeger dashboard and showing relevant spans of the component. This would require some labels to be present that allows us to filter spans based on the selected component. 

We already send some owner Metadata, for the spans as seen in the picture.  

image
The `hostname` label allows us to categorize spans based on the pod that the spans came from, allowing us to directly specify a link with the query `hostname` label for pods. But we would also want a similar case for the `kind` and `name` of workload.

Sadly, This Information is actually not passed into the pod's env variables right now. The `hostname` is injected by k8s currently and we use that. But A similar thing is not present for workloads.

More metadata for spans.

We already find the workload, type information during injection and add them as `pod` labels.
We can use the downward API https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/ to make those labels attached as volumes for the `linkerd-proxy` to attach them as labels here.
https://github.com/linkerd/linkerd2-proxy/blob/9faff18c8a6ffabe92afcd2a47ab169b99baa1b0/linkerd/trace-context/src/layer.rs#L142

- In Open Telemetry, there is a k8s processor which attaches those labels as per https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/processor/k8sprocessor by doing a reverse pod lookup of the pod. But openTelemtry is not 1.0 yet. I'm giving this a try and will psot the results.
Further reading:
https://developers.soundcloud.com/blog/using-kubernetes-pod-metadata-to-improve-zipkin-traces

@grampelberg

Pothulapati:
Fixes #4008 

This PR updates yaml manifests to include pod's `metadata.labels` as a volume, for the `linkerd-proxy` container to use.

Changes Include
- Adds `env`, `volume` and `volumeMounts` to the manifests using downwardAPI, when tracing is enabled through `.Values.global.proxy.trace`
- Add a Inject unit test for the same
- Include `linkerd.io/workload-ns` as a label that is added during proxy injection

@grampelberg 


Pothulapati:
Traces would have the pod labels as additional attributes like 
Screenshot_2020-03-25 Jaeger UI
image

Pothulapati:
@alpeb This PR adds a new ENV variable for labels file along with a volumeMount of the pod labels. Use `tarunpothulapati/l5d-proxy` as the proxy image (along with the tracing component, and control-plane-tracing enabled) the see it working, as it includes the changes from https://github.com/linkerd/linkerd2-proxy/pull/463

Complete installation command would be `./target/cli/linux/linkerd install --ignore-cluster --addon-config config.yaml --control-plane-tracing --proxy-image tarunpothulapati/l5d-proxy --proxy-version latest  | k apply -f `

pkg/inject/inject.go
```diff
@@ -545,12 +559,6 @@ func (conf *ResourceConfig) injectPodSpec(values *patch) {
 	values.Global.IdentityTrustDomain = idctx.GetTrustDomain()
 	values.Identity = &l5dcharts.Identity{}
 
-	values.AddRootVolumes = len(conf.pod.spec.Volumes) == 0
-
```
Pothulapati, line 549: This code is moved up, as there is a return path in the middle i.e when identity is disabled (Trace should be able to be added even when identity is disabled)

charts/partials/templates/_proxy.tpl
```diff
@@ -119,8 +125,14 @@ lifecycle:
         - -c
         - sleep {{.Values.global.proxy.waitBeforeExitSeconds}}
 {{- end }}
-{{- if or (not .Values.global.proxy.disableIdentity) (.Values.global.proxy.saMountPath) }}
+{{- if or (not .Values.global.proxy.disableIdentity) (.Values.global.proxy.saMountPath) (.Values.global.proxy.trace) }}
 volumeMounts:
+{{- if (.Values.global.proxy.trace) }}
+{{- if (.Values.global.proxy.trace.collectorSvcAddr)}}
```
Pothulapati, line 131: We have to have separate checks because helm dosen't support short-cut evaluations. And checking for `.Values.global.proxy.trace.collectorSvcAddr` when `.Values.global.proxy.trace` is nil throws an error.

Pothulapati: Also, Haven't added the same volumes, etc for `controlPlaneTracing`, as I felt that these tags are net necessary for them, but I can do it if you think it is!

Pothulapati:
Sorry for pushing so many commits, Running Integration tests locally makes my laptop struct. So, Had to depend fully on CI ;)

Pothulapati:
To see the attributes on the spans;
- Install Linkerd by running 
```bash
./target/cli/linux/linkerd install --ignore-cluster --addon-config config.yaml --control-plane-tracing --proxy-image tarunpothulapati/l5d-proxy --prox-version latest | k apply -f -
```
with `config.yaml` as 
```
tracing:
  enabled: true
```
- Install emojivoto application by running
```bash
 curl -sL https://run.linkerd.io/emojivoto.yml | ./target/cli/linux/linkerd inject - | kubectl apply -f -
```
- Enable application traces by running
```bash
 kubectl -n emojivoto set env --all deploy OC_AGENT_HOST=linkerd-collector.linkerd:55678
```
- Now, Enabling proxy traces by running
```bash
 kubectl -n emojivoto patch -f https://run.linkerd.io/emojivoto.yml -p '               spec:
  template:
    metadata:
      annotations:
        linkerd.io/inject: enabled
        config.linkerd.io/trace-collector: linkerd-collector.linkerd:55678
        config.alpha.linkerd.io/trace-collector-service-account: linkerd-collector
        config.linkerd.io/proxy-log-level: trace
```


You should be able to see the traces with pod labels as attributes in jeager UI by running
```bash
k -n linkerd port-forward svc/linkerd-jaeger 16686:16686
```

Screenshot_2020-04-10 Jaeger UI

charts/partials/templates/_proxy.tpl
```diff
@@ -119,8 +125,14 @@ lifecycle:
...
+- name: podinfo
+  mountPath: var/run/linkerd/podinfo
```
alpeb, line 133: Please swap these lines to follow the same alphabetical ordering the other mounts have :wink:

charts/partials/templates/_proxy.tpl
alpeb: Not having these tags in the control plane components would mean they wouldn't get a jaegger link in the dashboard (should we go down that route), right? I think they deserve the same treatment as non-control-plane pods :-)

alpeb: On the other hand, regarding these lines, I think we can assume that `Values.global.proxy.trace` is always set as per the `values.yaml` structure. So would only need to check on `Values.global.proxy.trace.collectorSvcAddr`. Also, `patch.json` is only checking for `Values.global.proxy.trace`; if I'm not mistaking we should change that to `Values.global.proxy.trace.collectorSvcAddr`.

Pothulapati: @alpeb Found the issue that was causing `Values.global.proxy.trace` to be nil both in install and inject paths, and updated it. :)

cli/cmd/install_test.go
```diff
@@ -242,6 +243,7 @@ func TestRender(t *testing.T) {
 			if err := render(&buf, tc.values); err != nil {
 				t.Fatalf("Failed to render templates: %v", err)
 			}
+			fmt.Println(buf.String())
```
alpeb, line 246: :eyes:

alpeb:
I tried regenerating `pod.patch.json` and there are lots of indentation changes, but not as much as what you found, so it might have been your IDE that introduced them. I used the following to generate that file:
```diff
diff --git a/controller/proxy-injector/webhook_test.go b/controller/proxy-injector/webhook_test.go
index 332afd86..992c8e6d 100644
--- a/controller/proxy-injector/webhook_test.go
+++ b/controller/proxy-injector/webhook_test.go
@@ -3,6 +3,7 @@ package injector
 import (
        "encoding/json"
        "fmt"
+       "os"
        "path/filepath"
        "reflect"
        "testing"
@@ -127,6 +128,9 @@ func TestGetPatch(t *testing.T) {
                                if err != nil {
                                        t.Fatalf("Unexpected PatchForAdmissionRequest error: %s", err)
                                }
+                               f, _ := os.Create("fake/data/pod.patch.json")
+                               f.Write(patchJSON)
+                               defer f.Close()
                                actualPatch, err := unmarshalPatch(patchJSON)
                                if err != nil {
                                        t.Fatalf("Unexpected error: %s", err)
```

```bash
go test ./controller/proxy-injector -run TestGetPatch
```

charts/partials/templates/_proxy.tpl
```diff
@@ -119,8 +121,12 @@ lifecycle:
         - -c
         - sleep {{.Values.global.proxy.waitBeforeExitSeconds}}
 {{- end }}
-{{- if or (not .Values.global.proxy.disableIdentity) (.Values.global.proxy.saMountPath) }}
+{{- if or (not .Values.global.proxy.disableIdentity) (.Values.global.proxy.saMountPath) (.Values.global.proxy.trace) }}
```
alpeb, line 124: Shouldn't this be:
```
{{- if or (.Values.global.proxy.trace.collectorSvcAddr) (.Values.global.controlPlaneTracing) (not .Values.global.proxy.disableIdentity) (.Values.global.proxy.saMountPath) }}
```

Pothulapati: yep, my bad!

Pothulapati:
@alpeb My Initial thinking was to leave the control-plane components, as they already had `linkerd-io/controlplane-ns` and it would be redundant. I was planning to do some js magic to handle those cases in the dashboard. But I guess its better to be consistent, So added those labels. :)

Pothulapati:
@alpeb Looks like updating the `label selectors` for control plane components is going to hurt us for upgrades https://github.com/kubernetes/kubernetes/issues/26202#issuecomment-450523155 (Upgrade tests fail too)

But, This should not effect us for injection, as we add labels to the pod but not the workload.

alpeb:
> @alpeb Looks like updating the `label selectors` for control plane components is going to hurt us for upgrades [kubernetes/kubernetes#26202 (comment)](https://github.com/kubernetes/kubernetes/issues/26202#issuecomment-450523155) (Upgrade tests fail too)
> 
> But, This should not effect us for injection, as we add labels to the pod but not the workload.

Ok you're right. I thought replacing the deployment with changes in the label selectors would swap the entire deployment and that wouldn't be considered a mutation, but I do see the error when I attempt doing that:
```
The Deployment "linkerd-controller" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"foo":"bar", "linkerd.io/control-plane-component":"controller", "linkerd.io/control-plane-ns":"linkerd", "linkerd
.io/proxy-deployment":"linkerd-controller"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```
So this means we have to rollback doing that to the control plane components and rely on the javascript magic you mentioned before?

Pothulapati:
@alpeb Yep, Doing that. 

P.S: It feels really bad to have this blocker to update labels, as it could be a common use-case/change.

alpeb:
@Pothulapati seems like you missed including `partials.proxy.volumes.labels` in `smi-metrics.yaml`?

charts/partials/templates/_volumes.tpl
```diff
@@ -3,3 +3,12 @@ emptyDir:
   medium: Memory
 name: linkerd-identity-end-entity
 {{- end -}}
+
```
adleong, line 6: I think it would be very helpful to have some comments here describing what this is and the context for it existing.

Pothulapati:
@adleong Thanks, That's a really good suggestion!

As the changes are spread across multiple files, I added a comment in `volumes.tpl` with the full context. Feel free to comment if you have any changes in your mind! :)

Pothulapati:
@adleong I added a caution comment in `partials/metadata.tpl` :)

Issue #541, opened 2018-03-08
Proxy exposes Prometheus protobuf metrics


hawkw:
The proxy metrics endpoint (added in #540) should also expose metrics in the Prometheus [Protocol Buffers format](https://prometheus.io/docs/instrumenting/exposition_formats/#protocol-buffer-format-details). Whether to serve text metrics or protobuf should be determined based on the request's content type, with `text/plain` requests receiving text metrics and `application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited` requests receiving protobuf metrics.

hawkw:
According to the Prometheus [documentation](https://prometheus.io/docs/instrumenting/exposition_formats/):

> NOTE: Prometheus 2.0 removed support for the protocol-buffer format and only supports the text-based format.

Are we still interested in supporting the protocol buffer format? We might want to for non-Prometheus 2.0 installations, but removing support for this could potentially make our metrics implementation much more efficient (as `prost` requires a lot of things to be `Vec`s or `String`s that could otherwise be non-allocating).

PR #4200, opened 2020-03-25
Add uninstall cmd functionality to cli (#3622)

### LINKED ISSUE #3622: Provide linkerd delete command
Currently deletion of linkerd from an k8s cluster is done via the `linkerd install --ignore-cluster` command. This has worked so far but after the work carried out in #3600, some questions emerged. Namely, the previous command relies on the fact that we can generate a yaml file that is identical to the one used by the installation phase. In certain scenarios, when we cannot ignore the cluster and we need to fetch some data from it to generate the file (i.e. using --external-issuer=true and fetching the trust anchors from a secret) the standard approach simply does not work. 

I think what we need to have is a generic linkerd delete/uninstall/any-other-suitable-name command that will simply generate a yaml file that can be used to delete the linkerd resources from the k8s cluster. This command should work even if there is no connection to the cluster. 

### Open questions? 
Should it actually be a text only command that outputs the minimum yaml needed to do kubectl delete and get rid of linkerd or it should actually talk to the cluster and perform the whole deletion ? 
If its a just yaml command, does it make sense to have a separate helm chart for it?
Are there any edge cases we need to think about (i.e. having a iiexternal-issuer flag that leaved the namespace alone once #3600  is merged?)

cc @grampelberg

Matei207:
### Edit 29th March: 
PR overhauled, updated body of this msg accordingly

### What
---

* Introduce an `uninstall` command to the cli to generate resources necessary to purge/uninstall Linkerd from a Kubernetes cluster (uninstall was chosen over delete for the sake of brevity and conciseness in commands);

### How
---

* `uninstall` uses `client-go` to fetch all `cluster-wide resources` (RBAC, hooks, etc.) and the control-plane `namespace` and outputs them to stdout as yaml manifests (to be piped to `kubectl delete`
* clientset is used to grab all resources that have the `linkerd.io/control-plane-ns` label. For the `namespace` itself, the name is used instead of a label selector.
* Since `gvk` is stripped down when fetching a Kubernetes resource using `client-go` ([see here](https://github.com/kubernetes/client-go/issues/541#issuecomment-452312901)), the `marshal` function also infers the `GVK` from the object type in case it is missing.

### Test
* Test case is quite trivial, however, it ensures that empty files (as strings) are not generated, and it ensures the `apiVersion` and `kind` are correctly injected/present (without them, the pipe to `kubectl` will fail).

* I checked the results manually by following the steps below (see attached pictures):

1. Perform a fresh install through `linkerd install | kubectl apply -f -`
2. Run  `kubectl get APIService,ClusterRoleBinding,CustomResourceDefinition,MutatingWebhookConfiguration,ValidatingWebhookConfiguration,PodSecurityPolicy,Namespace  -l linkerd.io/control-plane-ns=linkerd -o yaml` -- this outputs all resources that should be deleted (except `RoleBIndings` in `kube-system` namespace which have been added as part of this PR)
3. Run `/bin/go-run cli uninstall | kubectl delete -f - `
4. Repeat step (2) -- output is now empty list

---
Screen Shot 2020-03-29 at 4 02 31 PM
Screen Shot 2020-03-29 at 4 05 22 PM

### DIFF
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -1,0 +1,293 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/ghodss/yaml"
+	"github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/spf13/cobra"
+	admissionRegistration "k8s.io/api/admissionregistration/v1beta1"
+	core "k8s.io/api/core/v1"
+	extensions "k8s.io/api/extensions/v1beta1"
+	rbac "k8s.io/api/rbac/v1"
+	apiextension "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	apiRegistration "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	yamlSep = "---\n"
+)
+
+type kubernetesResource struct {
+	runtime.TypeMeta
+	metav1.ObjectMeta `json:"metadata"`
+}
+
+func newKubernetesResource(apiVersion, kind, name string) kubernetesResource {
+	return kubernetesResource{
+		runtime.TypeMeta{
+			APIVersion: apiVersion,
+			Kind:       kind,
+		},
+		metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+}
+
+func newCmdUninstall() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "uninstall",
+		Args:  cobra.NoArgs,
+		Short: "Output Kubernetes resources to uninstall Linkerd control plane",
+		Long: `Output Kubernetes resources to uninstall Linkerd control plane.
+
+This command provides all Kubernetes namespace-scoped and cluster-scoped resources (e.g services, deployments, RBACs, etc.) necessary to uninstall Linkerd control plane.`,
+		Example: ` linkerd uninstall | kubectl delete -f -`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return uninstallRunE()
+		},
+	}
+
+	return cmd
+}
+
+func uninstallRunE() error {
+	k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+	if err != nil {
+		return err
+	}
+
+	resources, err := fetchKubernetesResources(k8sAPI)
+	if err != nil {
+		return err
+	}
+
+	for _, r := range resources {
+		if err := r.renderResource(os.Stdout); err != nil {
+			return fmt.Errorf("error rendering Kubernetes resource:%v", err)
+		}
+	}
+	return nil
+}
+
+func (r kubernetesResource) renderResource(w io.Writer) error {
+	b, err := yaml.Marshal(r)
+	if err != nil {
+		return err
+	}
+
+	_, err = w.Write(b)
+	if err != nil {
+		return err
+	}
+
+	_, err = w.Write([]byte(yamlSep))
+	return err
+}
+
+func fetchKubernetesResources(k *k8s.KubernetesAPI) ([]kubernetesResource, error) {
+	options := metav1.ListOptions{
+		LabelSelector: k8s.ControllerNSLabel,
+	}
+
+	resources := make([]kubernetesResource, 0)
+
+	clusterRoles, err := fetchClusterRoles(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch ClusterRole resources:%v", err)
+	}
+	resources = append(resources, clusterRoles...)
+
+	clusterRoleBindings, err := fetchClusterRoleBindings(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch ClusterRoleBinding resources:%v", err)
+	}
+	resources = append(resources, clusterRoleBindings...)
+
+	roleBindings, err := fetchKubeSystemRoleBindings(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch RoleBindings from kube-system namespace:%v", err)
+	}
+	resources = append(resources, roleBindings...)
+
+	crds, err := fetchCustomResourceDefinitions(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch CustomResourceDefinition resources:%v", err)
+	}
+	resources = append(resources, crds...)
+
+	apiCRDs, err := fetchAPIRegistrationResources(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch APIService CRDs:%v", err)
+	}
+	resources = append(resources, apiCRDs...)
+
+	psps, err := fetchPodSecurityPolicy(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch PodSecurityPolicy resources:%v", err)
+	}
+	resources = append(resources, psps...)
+
+	mutatinghooks, err := fetchMutatingWebhooksConfiguration(k, options)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch MutatingWebhookConfigurations:%v", err)
+	}
+	resources = append(resources, mutatinghooks...)
+
+	validationhooks, err := fetchValidatingWebhooksConfiguration(k, options)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


alpeb:
Thank you @Matei207 for all your effort. I'm gonna add some comments into the original issue #3622.

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,304 @@
...
+	nonEmptyResources := make([]string, 0)
+	for _, resource := range resources {
+		if resource != "" {
+			nonEmptyResources = append(nonEmptyResources, resource)
+		}
+	}
+	return nonEmptyResources
+}
+
+func fetchClusterRoles(k *k8s.KubernetesAPI, options metav1.ListOptions) (string, error) {
```
adleong: I notice that we have a mix of list and non-list type resources in the output, which probably works, but is a bit confusing.  I'd prefer to just have a sequence of non-list resources.  I think we can accomplish that by having all of these `fetch...` functions return a slice of `ObjectReference` or something like that.

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,304 @@
...
+	m, err := marshalKubernetesResource(list)
+	if err != nil {
+		return "", err
+	}
+	return m, nil
+}
+
+func marshalKubernetesResource(resource runtime.Object) (string, error) {
+	var gvk = resource.GetObjectKind().GroupVersionKind()
+	if gvk.Empty() {
```
adleong: I'm not sure what's going on here with populating the gvk.  Can you explain this?

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,304 @@
...
+		gvks, _, err := k8s.ObjectKinds(resource)
+		if err == nil && len(gvks) > 0 {
+			gvk = gvks[0]
+		}
+		resource.GetObjectKind().SetGroupVersionKind(gvk)
+	}
+	b, err := yaml.Marshal(resource)
```
adleong: Technically, we only need to marshal the name, namespace (if applicable), kind, and apiVersion, right?  I think we can maybe get away with just marshalling an ObjectReference instead of the entire resource.

cli/cmd/uninstall.go
Matei207: @adleong of course! I realise how unnecessarily complex this looks.

Using `client-go`, whenever we fetch a resource, the `TypeMeta` seems to be stripped from it. I haven't managed to find a reason why this happens, but going through some related issues, it seems to be an intentional design; [#541 in client-go](https://github.com/kubernetes/client-go/issues/541) is an example of an issue where this problem is brought up to light.

I found a helper function that returns the gvk by using reflection on the object type, and that's the solution I went for. Prior to this, I tried to inject the gvk myself, this seemed to be a bit cleaner, however.

If there's another solution for it, I'll get it changed asap.

cli/cmd/uninstall.go
Matei207: @adleong yes, completely right! It was an oversight on my part, was under the impression it did not work just with the kind, name and apiVersion. Will make changes.

cli/cmd/uninstall.go
Matei207: @adleong great suggestion will get it underway! Since we only need kinds, name and version, I thought maybe it's better to have our own wrapper objects around the metadata we want to marshal? In that case, all functions can return a slice of "wrapperStruct".

Matei207:
I have updated the PR with the above comments @adleong.

I have added two new structs to reduce the verbosity when marshaling. As you suggested, all of the fetch... functions (except the namespace one) now return a slice of `kubernetesResource`.

The resource itself holds the `apiVersion`, the `kind` and the `namespace` and `name` (both under a different metadata struct). The only one that needs to use the namespace is the `RoleBinding` object.

There is a bit of repetitiveness going around that I have no idea how to solve (particularly around turning list objects into kubernetesResource types), even if I would have that as a separate method that accepts an interface{}, all list objects would have to be separately cast.

I hope this is satisfactory, if not, keen on seeing what else can be improved!

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,399 @@
...
+		Name      string `json:"name"`
+		Namespace string `json:"namespace"`
+	}
+)
+
+func newCmdUninstall() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "uninstall",
+		Args:  cobra.NoArgs,
+		Short: "Output Kubernetes resources to uninstall Linkerd",
```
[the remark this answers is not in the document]
Matei207: Done, thank you.

Matei207:
Thank you for your comment, @alpeb .

I have now added an integration test like you have suggested. It occurs before `TestInstall_sp` and compares the output from `uninstall` with the golden file.

The integration tests (`custom_domain`) kept failing since before I have introduced anything new to them, not sure if this is just undefined behaviour or something I have introduced. I have rebased off master and pushed again to get rid of all of the excess commits and re-trigger the build.

If there is anything else you'd like changed, let me know.

alpeb:
The integration test failure was introduced by #4212. I'm working on a fix.

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,399 @@
...
+)
+
+type (
+	kubernetesResource struct {
+		APIVersion string             `json:"apiVersion"`
+		Kind       string             `json:"kind"`
+		Metadata   kubernetesMetadata `json:"metadata"`
+	}
+
+	kubernetesMetadata struct {
```
adleong: can we use [ObjectMeta](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) instead?

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,399 @@
...
+)
+
+const (
+	// YamlSep is used to separate manifests on stdout print
+	YamlSep = "---\n"
+)
+
+type (
+	kubernetesResource struct {
+		APIVersion string             `json:"apiVersion"`
```
adleong: Can these fields be an embedded TypeMeta struct instead?

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,399 @@
...
+		return nil, err
+	}
+
+	if list.Items == nil {
+		return nil, nil
+	}
+
+	resources := make([]kubernetesResource, len(list.Items))
+	for i, item := range list.Items {
+		v, k := getObjectGvk(item.DeepCopy()).ToAPIVersionAndKind()
```
adleong: It's annoying that the kubernetes client strips out the gvk like this.  However, we actually statically know the gvk already since we know we're dealing with cluster roles here.  We can manually set the kind and apiVersion to `ClusterRole` and `rbac.authorization.k8s.io/v1`.  And similarly for each other resource type, we know their kind and apiVersion ahead of time so we can manually fill in those fields.

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,399 @@
...
+	return resources, nil
+}
+
+func fetchClusterRoles(k *k8s.KubernetesAPI, options metav1.ListOptions) ([]kubernetesResource, error) {
+	list, err := k.RbacV1().ClusterRoles().List(options)
+	if err != nil {
+		return nil, err
+	}
+
+	if list.Items == nil {
```
adleong: I believe that if a slice is nil, the for-loop will skip right over it, so I don't think we need this special case.

cli/cmd/uninstall.go
Matei207: The embedded fields made everything much better, amazing suggestion, thank you.

cli/cmd/uninstall.go
Matei207: Done. It reads much better now, thank you.

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,288 @@
...
+	apiRegVersion1            = "apiregistration.k8s.io/v1"
+)
+
+type kubernetesResource struct {
+	runtime.TypeMeta
+	metav1.ObjectMeta `json:"metadata"`
+}
+
+func newKubernetesResource(apiVersion, kind, name string) kubernetesResource {
+	r := kubernetesResource{}
```
adleong, line 34: I think we can inline this struct initialization:

```
return kubernetesResource {
  runtime.TypeMeta {
    APIVersion: apiVersion,
    Kind: kind,
  },
  metav1.ObjectMeta {
    Name: name,
  },
}
```

Matei207: Changed. Good to know how embedded types behave in go, thought (for some reason) that wouldn't work quite well, but turns out it does 👍 .

cli/cmd/uninstall.go
```diff
@@ -0,0 +1,288 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	yamlSep                   = "---\n"
+	rbacVersion1              = "rbac.authorization.k8s.io/v1"
```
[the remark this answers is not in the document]
Matei207: Took the suggestion into account.

PR #4204, opened 2020-03-26
edge-20.3.4

adleong:
This release introduces several fixes and improvements to the CLI.

* CLI
  * Added support for kubectl-style label selectors in many CLI commands (thanks
    @mayankshah1607!)
  * Fixed the path regex in service profiles generated from proto files without
    a package name (thanks @amariampolskiy!)
  * Fixed an error when injecting Cronjobs that have no metadata
  * Relaxed the clock skew check to match the default node heartbeat interval
    on Kubernetes 1.17 and made this check a warning
  * Fixed a bug where the linkerd-smi-metrics pod could not be created on
    clusters with pod security policy enabled
* Internal
  * Upgraded tracing components to more recent versions and improved resource
    defaults (thanks @Pothulapati!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,19 @@
+## edge-20.3.4
+
+This release introduces several fixes and improvements to the CLI.
+
+* CLI
+  * Added support for kubectl-style label selectors in many CLI commands (thanks
+    @mayankshah1607!)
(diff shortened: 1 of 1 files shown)


alpeb:
#4193 fixed Linkerd in the presence of PSP. This affected edge-20.3.1 and up.

PR #4221, opened 2020-03-31
Upgrade to client-go 0.17.4 and smi-sdk-go 0.3.0

adleong:
Here we upgrade our dependencies on client-go to 0.17.4 and smi-sdk-go to 0.3.0.  Since smi-sdk-go uses client-go 0.17.4, these upgrades must be performed simultaneously.

This also requires simultaneously upgrading our dependency on linkerd/stern to a SHA which also uses client-go 0.17.4.  This keeps all of our transitive dependencies synchronized on one version of client-go.

This ALSO requires updating our codegen scripts to use the 0.17.4 version of code-generator and running it to generate 0.17.4 compatible generated code.  I took this opportunity to update our code generation script to properly use the version of code-generater from `go.mod` rather than a hardcoded SHA.


go.mod
```diff
@@ -41,28 +37,28 @@ require (
...
+	k8s.io/kube-aggregator v0.17.4
+	k8s.io/kube-openapi v0.0.0-20200121204235-bf4fb3bd569c // indirect
+	sigs.k8s.io/yaml v1.2.0
 )
 
-replace github.com/wercker/stern => github.com/linkerd/stern v0.0.0-20200316183041-1ab5375fb7e9
-
-replace k8s.io/apimachinery v0.0.0-20181127105237-2b1284ed4c93 => k8s.io/apimachinery v0.0.0-20181127025237-2b1284ed4c93
-
-replace k8s.io/code-generator v0.0.0-20181116203124-c2090bec4d9b => k8s.io/code-generator v0.0.0-20181117043124-c2090bec4d9b
```
alpeb, line 68: :+1: :+1:

PR #4226, opened 2020-04-02
Add support for Helm configuration of per-component proxy resources requests and limits

### LINKED ISSUE #4225: Add support for configuration of per-component resources configuration for proxy in Helm chart.
Currently only one cluster-wide confguration for the resources the proxy requests and is limited to are available. This is far from optimal, as the the really high values for the prometheus proxy are not needed for e.g. the web UI. 

Especially as the settings are also used for the injected proxies, this will break our cluster injecting 900+ pods.

Allow per-component setting of proxy resources based on the value of `.Values.global.proxy.component` (additionally e.g. `injected-proxy` for the configuration for the injector) in `.Values.global.proxy.perComponentResources` 

I do not know of any.

setting `.Values.global.proxy.perComponentResources` in their `values.yaml` file

cypherfox:
Add support for Helm configuration of per-component proxy resources requests and limits.

Currently the resource requests and limits for resources assigned to the proxy are the same for all proxies in a cluster. This results in the largest value for all proxies having to be set. Optimal values
may vary by an order of magnitude between components (e.g. `linkerd-prometheus` and `linkerd-web`

Solution:
Allow the setting of `.Values.global.proxy.perComponentResources.<resource name>.resources` to override the resource settings in the Helm chart.

Validation

Fixes #4225


cypherfox:
This change currently lacks two things:
* support for the proxy injector config (working on that right now)
* appropriate tests.

For the later I need some help. Is there an example I can follow on what to test with changes to the helm charts?

Also: does anybody have a better idea on how to call the section with the per component configuration? I am not all that happy with `perComponentResources`

cypherfox:
Yes, your way is cleaner. I will update the code accordingly.

cypherfox:
I have had a look at the code, and unfortunately doing it your way will introduce a potentially breaking change: 
(Prometheus is the first example I look at, but I assume the others are similar)
Currently `.Values.prometheusResources` is used (`templates/prometheus.yaml:216`). I can either add a `.Values.prometheusProxyResources` or split that up into `.Values.prometheus.resources` and `.Values.prometheus.proxy.resources`?

For all other components similar naming applies. E.g. your example: `.Values.publicAPIresources` controls the resources used by the controller pod, not the proxy used there.

I do not have any good idea on how to handle legacy configs in a clean way, and to support people updating their configurations that reference the linkerd2 chart over time.

alpeb:
Oh yeah, good point. I think it'll be fine to name these new properties like `componentProxyResources` to distinguish them from the main container resources, and thus avoid backward incompatible changes.

charts/linkerd2/templates/controller.yaml
```diff
@@ -94,6 +94,7 @@ spec:
         volumeMounts:
         - mountPath: /var/run/linkerd/config
           name: config
+      {{- $_ := set .Values.global.proxy "resources" (default .Values.global.proxy.resources .Values.controllerProxyResources) -}}
```
alpeb, line 97: Please replace the leading `{{-` with just `{{` to avoid suppressing the line break above, which is also breaking the unit tests. This applies to all the template files you modified.

charts/linkerd2/templates/prometheus.yaml
```diff
@@ -224,6 +224,7 @@ spec:
         - mountPath: /etc/prometheus
           name: prometheus-config
           readOnly: true
+      {{- $_ := set .Values.global.proxy "resources" (default .Values.global.proxy.resources .Values.prometheusProxyResources) }}
```
alpeb, line 227: Same comment as above, but here you missed the dash at the end `-}}`

charts/linkerd2/values.yaml
```diff
@@ -177,6 +218,15 @@ smiMetrics:
...
 
+
+# set resources for smi-metrics linkerd proxy
+# see global.proxy.resources for details.
+#smiMetricsProxyResources:
+
+  proxy:
+  # set resources for linkerd proxy, see global.proxy.resources for details
+  #resources:
+
```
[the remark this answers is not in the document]
cypherfox: yes.
Also: the smiMetrics shim container did not set resources. I added that as well.

charts/linkerd2/README.md
```diff
@@ -147,14 +155,22 @@ The following table lists the configurable parameters of the Linkerd2 chart and
...
 | `prometheusLogLevel`                  | Log level for Prometheus                                                                                                                                                              | `info`                               |
+| prometheusResources                   | CPU and Memory resources requested for prometheus (see `global.proxy.resources` for sub-fields)             |   |
+| prometheusProxyResources              | CPU and Memory resources requested for proxy injected into prometheus pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
+| `proxyInjectorResources`                   | CPU and Memory resources requested for the proxy injector (see `global.proxy.resources` for sub-fields)             |   |
+| `proxyInjectorProxyResources`              | CPU and Memory resources requested for proxy injected into the proxy injector pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
```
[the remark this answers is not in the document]
cypherfox: ...and also `smiMetrics`

charts/linkerd2/templates/controller.yaml
cypherfox: But there is an additional linebreak at the end of the line 97. Thus I have to eat the one in the previous line.

charts/linkerd2/templates/prometheus.yaml
cypherfox: Setting the the dashes as you propose would add the prefix spaces of each line, would it not?
I have to test that.

cypherfox:
I have force pushed in order to rebase onto the latest changes in master.  
What is the impact to you when I force push?

charts/linkerd2/templates/prometheus.yaml
cypherfox: Ah, I just noticed, I am not consistent. I will fix that.

charts/linkerd2/values.yaml
```diff
@@ -122,8 +133,19 @@ identity:
...
+# see global.proxy.resources for details.
+#identityResources:
+#identityProxyResources:
+
+
 # grafana configuration
 grafanaImage: gcr.io/linkerd-io/grafana
+# set resources for grafana linkerd proxy
+# see global.proxy.resources for details.
+#grafanaResources:
```
alpeb, line 146: "set resources for grafana and its linkerd proxy..."

charts/linkerd2/values.yaml
```diff
@@ -177,6 +219,11 @@ smiMetrics:
 
   keyPEM: |
 
+# set resources for smi-metrics linkerd proxy
+# see global.proxy.resources for details.
+#smiMetricsResources:
+#smiMetricsProxyResources:
```
alpeb, line 225: "set resources for ami-metrics and its linkerd proxy..."

charts/linkerd2/README.md
```diff
@@ -143,18 +149,32 @@ The following table lists the configurable parameters of the Linkerd2 chart and
...
 | `profileValidator.keyPEM`             | Certificate key for the service profile validator. If not provided then Helm will generate one.                                                                                                                             ||
+| `smiMetricsResources`                   | CPU and Memory resources required by SMI metrics (see `global.proxy.resources` for sub-fields)             |   |
+| `smiMetricsProxyResources`              | CPU and Memory resources required by proxy injected into SMI metrics pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
+| `spValidatorResources`                   | CPU and Memory resources required by the SP validator (see `global.proxy.resources` for sub-fields)             |   |
+| `spValidatortapProxyResources`              | CPU and Memory resources required by proxy injected into the SP validator pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
```
alpeb, line 169: Typo: spValidatortapProxyResources

charts/linkerd2/README.md
```diff
@@ -143,18 +149,32 @@ The following table lists the configurable parameters of the Linkerd2 chart and
 | `identity.issuer.scheme`              | Which scheme is used for the identity issuer secret format                                                                                                                            | `linkerd.io/tls`                     |
 | `identity.issuer.tls.crtPEM`          | Issuer certificate (ECDSA). It must be provided during install.                                                                                                                                                             ||
 | `identity.issuer.tls.keyPEM`          | Key for the issuer certificate (ECDSA). It must be provided during install.                                                                                                                                                 ||
+| `identityResources`                   | CPU and Memory resources required by the identity controller (see `global.proxy.resources` for sub-fields)             |   |
+| `identityPoxyResources`              | CPU and Memory resources required by proxy injected into identity pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
```
alpeb, line 153: Typo: identityPoxyResources

charts/linkerd2/templates/controller.yaml
alpeb: That makes sense, but for some reason having both suppressors was rendering an invalid template. Anyways after the last set of changes it's working fine :man_shrugging:

charts/linkerd2/templates/smi-metrics.yaml
```diff
@@ -78,6 +78,11 @@ spec:
           name: tls
           readOnly: true
 
+        {{- if .Values.smiMetricsResources -}}
+        {{- include "partials.resources" .Values.prometheusResources | nindent 8 }}
```
alpeb, line 82: Should be `.Values.smiMetricsResources`

cypherfox: fixed

cypherfox:
> @cypherfox I see the `actualResources` approach got reintroduced again, was that on purpose?

Yes, no. Once I better understood the integration between the helm charts and the way that the linkerd cli tool utilizes them requires a different approach than I had planed. 

*The problem:* when the chart user overwrites only a subset of the resources it will break the resource struct and the test will fail (as they should). Thus I have used the merge function to overwrite only the value set.

But when assigning that to the `.Values` tree it will propagate the changed values to the other K8s resources created per linkerd component, as linkerd maintains only a single `.Values`tree. I am not sure if this is in contrast to the way helm works, but that is irrelevant as the solution has to fit in both.

*Proposed Solution:*  have the helm chart create a deep copy of `.Values` when computing the effective resource settings per component an use that as the argument instead of `.` in call to `partials.proxy`. The partial will use the `.Values.global.proxy.resources` as in the current stable code.

I haven't had the time on thursday (yesterday was a public holiday for me) to implement that. Will get back to that on monday.

cypherfox:
Ok, the deepCopy command is not available in helm 2.16.1, but with saving the original state and copying it back afterwards it seems to work.

The test I added to check the individual values also works.

alpeb:
Note that as you merged from master, now Grafana is implemented as an add-on, not as a regular template. There is more info about addons in this PR: linkerd/website#686, but that's just for tracing as the Grafana docs aren't ready yet.
That means the `.Values` accessible in the Grafana template are just the ones under the `grafana` and `global` dictionaries in `values.yaml`. So things like `grafanaProxyResources` aren't available, and you should move that under the `grafana` section.

cypherfox:
@alpeb Please have a look. 

I have no idea what the failing CI test is telling me. Any help appreciated.

kleimkuhler:
@cypherfox Could you please try merging in master to this branch? A fix recently went in for this issue, so that should take care of the current failure you are seeing.

charts/add-ons/grafana/templates/grafana.yaml
```diff
@@ -147,7 +147,15 @@ spec:
...
           readOnly: true
+      {{- if not (empty .Values.proxyResources) }}
+      {{- $orig := .Values.global.proxy.resources }}
+      {{- $r := merge .Values.proxyResources .Values.global.proxy.resources }}
+      {{- $_ := set .Values.global.proxy "resources" $r }}
       - {{- include "partials.proxy" . | indent 8 | trimPrefix (repeat 7 " ") }}
+      {{- $_ := set .Values.global.proxy "resources" $orig }}
+      {{- else }}
+      - {{- include "partials.proxy" . | indent 8 | trimPrefix (repeat 7 " ") }}
+      {{- end }}
```
alpeb, line 158: `.Values.proxyResources` doesn't exist

pkg/charts/linkerd2/values.go
```diff
@@ -55,10 +55,21 @@ type (
...
+		SMIMetricsResources    *Resources `json:"smiMetricsResources"`
 		SPValidatorResources   *Resources `json:"spValidatorResources"`
 		TapResources           *Resources `json:"tapResources"`
 		WebResources           *Resources `json:"webResources"`
 
+		DestinationProxyResources   *Resources `json:"destinationProxyResources"`
+		IdentityProxyResources      *Resources `json:"identityProxyResources"`
+		PrometheusProxyResources    *Resources `json:"prometheusProxyResources"`
+		ProxyInjectorProxyResources *Resources `json:"proxyInjectorProxyResources"`
+		PublicAPIPRoxyResources     *Resources `json:"publicAPIProxyResources"`
```
alpeb, line 67: typo: s/PublicAPIPRoxyResources/PublicAPIProxyResources

cypherfox:
> One thing that comes to mind is to declare, in that new helper template, a dictionary with the key being the component name and the value being the appropriate var for the proxy resources override. And then override `.Values.global.proxy.resources` with the appropriate entry from that dictionary. Note that you have at hand `.Values.global.proxy.component` which is set for each component separately.

I do not think that this is good idea. While I agree on the benefit of reducing that boilerplate, injecting the proper Variable seems difficult. I understand the idea with the dict, but would try to avoid that, as it introduces a non-local association of code (when ever the name of the proxy resource field or the name of a component is changed, this dict must be updated too). Also that would make it more difficult for add-ons to set their proxy resources, as they still could not inject values into the tree without being part of that look-up.

I think the best was would be a re-write of the proxy fragment to take a values tree that only contains the required information and not gather it from all over the master `.Values` tree itself.
But that I would to be beyond the scope of this MR.

charts/add-ons/grafana/templates/grafana.yaml
```diff
@@ -147,7 +147,15 @@ spec:
         - mountPath: /etc/grafana
           name: grafana-config
           readOnly: true
+      {{- if not (empty .Values.proxy.resources) }}
```
alpeb, line 150: Since `.Values.proxy` is not necessarily defined (which is the default scenario), you'll need to check first that it's not empty, or else checking for `.Values.proxy.resources` gives a nil pointer error

alpeb:
I'm still optimist that we can reduce that boilerplate. From my testing, Helm v2.16.1 does have `deepCopy` available. [This](https://github.com/helm/helm/blob/bbdfe5e7803a12bbdf97e94cd847859890cf4050/glide.lock#L191) is the Sprig SHA used in v2.16.1, and [this](https://github.com/Masterminds/sprig/blob/e4c0945838d570720d876a6ad9b4568ed32317b4/dict.go) is the corresponding file in Sprig that implements that function. Can you check again if you can indeed leverage `deepCopy`?

cypherfox:
I have created #4602 as an alternative and changed the way the code works for `controller.yaml`.

I have installed helm 2.16.1 locally and the function `deepCopy` works in a small local test. But when running the l5d test script (https://gist.github.com/cypherfox/0af47ace95c60930d9de7816954a97af) it fails with the function not being found. The github CI pipeline fails with the same error. This was the original reason to build this copying contraption.

charts/linkerd2/README.md
```diff
@@ -189,6 +208,7 @@ The following table lists the configurable parameters for the Grafana Add-On.
 | `grafana.resources.cpu.request`     | Amount of CPU units that the gafana container requests                                                                                                                            ||
 | `grafana.resources.memory.limit`    | Maximum amount of memory that grafana container can use                                                                                                                        ||
 | `grafana.resources.memory.request`  | Amount of memory that the grafana container requests                                                                                                                               ||
+| `proxyResources`                    | Structure analog to the `resources` fields above, but overriding the resources of the linkerd proxy injected into the grafana pod.   | values in `global.proxy.resources` |
```
alpeb, line 211: s/proxyResources/grafana.proxy.resources/

charts/linkerd2/values.yaml
```diff
@@ -210,14 +252,22 @@ smiMetrics:
...
 
 grafana:
   enabled: true
   name: linkerd-grafana
   image:
     name: gcr.io/linkerd-io/grafana
+  # set resource requests and limits for grafana and its linkerd proxy respectively
+  # see global.proxy.resources in the linkerd2 chart for details.
   # resources:
+  #proxy.resources:
```
alpeb, line 270: Please remove the `#resources` line and replace `#proxy.resources` with
```
#proxy
#  resources:
```
If we use `proxy.resources`  in a single line it doesn't work, according to my tests.

charts/add-ons/grafana/values.yaml
```diff
@@ -0,0 +1,4 @@
+
+# must be present to avoid access to non-existing element of the Values tree
+proxy:
+  resources:
```
alpeb, line 4: This empty value is generating some errors when I try to override the grafana properties in the main `values.yaml`. And now that you put in place the checks for the existence of these properties, we can leave this file totally empty as it was before.

test/install_test.go
```diff
@@ -694,7 +829,7 @@ func TestLogs(t *testing.T) {
 }
 
 func TestEvents(t *testing.T) {
-	for err := range testutil.FetchAndCheckEvents(TestHelper) {
+	for _, err := range testutil.FetchAndCheckEvents(TestHelper) {
```
alpeb, line 832: Good catch :+1:

test/install_test.go
```diff
@@ -445,6 +475,111 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+		cpuLimitStr := resourceReqs.Limits.Cpu().String()
+		if cpuLimitStr != expected.cpuLimit {
+			t.Fatalf("unexpected %s CPU limit: expected %s, was %s", expected.pod, expected.cpuLimit, cpuLimitStr)
+		}
+		cpuRequestStr := resourceReqs.Requests.Cpu().String()
+		if cpuRequestStr != expected.cpuRequest {
+			t.Fatalf("unexpected %s CPU limit: expected %s, was %s", expected.pod, expected.cpuRequest, cpuRequestStr)
+		}
+	}
+}
```
alpeb, line 581: Nice tests :+1: 
But it appears you're not using `memLimit` nor `memRequest`?
Also, please replace the `t.Fatalf` calls with `testutil.AnnotatedFatalf(t, genericMsg, detailedMsg, args...)` that we're using now so that our CI reports can aggregate over these errors. `genericMsg` should be a message without details like "Unexpected CPU limit" and `detailedMsg`+`args` should be the detailed message like the ones you have.

go.mod
```diff
@@ -4,61 +4,78 @@ go 1.14
...
+// 	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.6.0
+// 	github.com/codegangsta/cli => github.com/urfave/cli v1.22.4
+// 	github.com/containerd/containerd v1.3.0-0.20190507210959-7c1e88399ec0 => github.com/containerd/containerd v1.3.0
+// 	github.com/docker/docker v1.14.0-0.20190319215453-e7b5f7dbe98c => github.com/docker/docker v1.13.0
+// 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc8
+// 	github.com/tonistiigi/fifo => github.com/containerd/fifo v0.0.0
+// 	github.com/uber-go/atomic => go.uber.org/atomic v1.6.0
+// 	github.com/wercker/stern => github.com/linkerd/stern v0.0.0-20200331220320-37779ceb2c32
+// 	google.golang.org/cloud => cloud.google.com/go v0.0.0
+// )
```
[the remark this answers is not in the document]
cypherfox: Because I forgot to press Ctrl-S :smile:

test/install_test.go
cypherfox: Added memory tests

charts/linkerd2/values.yaml
alpeb: Would you please address this before this gets merged? :slightly_smiling_face:

cypherfox: done

charts/linkerd2/README.md
```diff
@@ -146,33 +149,49 @@ their default values.
...
 | `smiMetrics.caBundle`                       | Bundle of CA certificates for the SMI metrics component. If not provided then Helm will use the certificate generated  for `smiMetrics.crtPEM`. If `smiMetrics.externalSecret` is set to true, this value must be set, as no certificate will be generated.                       ||
+| `smiMetricsResources`                       | CPU and Memory resources required by SMI metrics (see `global.proxy.resources` for sub-fields)             |   |
+| `smiMetricsProxyResources`                  | CPU and Memory resources required by proxy injected into SMI metrics pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
+| `spValidatorResources`                      | CPU and Memory resources required by the SP validator (see `global.proxy.resources` for sub-fields)             |   |
+| `spValidatortapProxyResources`              | CPU and Memory resources required by proxy injected into the SP validator pod (see `global.proxy.resources` for sub-fields)             | values in `global.proxy.resources`   |
```
Pothulapati, line 184: nit: `spValidatorProxyResources`

cypherfox: fixed

PR #4232, opened 2020-04-03
(WIP) Add option to fully externalize tls resources

### LINKED ISSUE #4203: Full automation of control plane TLS cert rotation for clusters
## Feature Request

While automating rotation of control plane TLS I noticed that there are three components that cannot be automated:

- Proxy Injector
- SP Validator
- Tap service

The certs for these credentials can be set manually through the helm chart, but the cli doesn't provide anyway to set these.

Provide full automation of all tls certs through the cli and allow automating all tls secrets through cert-manager.

Just like there's an option to externalize the issuer certificates, the same options should be provided for the three components listed above so they can be also managed by an external component like cert-manager.

When these options are provided, the generated manifests will skip creating the opaque tls secrets for  the three components above and also avoid injecting the `caBundle` for their corresponding webhooks and api service so they can be managed externally. CertManager provides options to inject the ca bundle for webhooks and api services, but I don't know of other tools besides cert manager.

The main drawback with providing these options is added complexity to the cli and the helm chart. It will also introduce a two-phase approach to installations as is done for an externally managed identity issuer.

The alternative to this option is to manually manage tls for these components through the helm chart. Once could follow the instructions described in the manual cert rotation doc, generate certs for these components and pass them to the helm [variables](https://github.com/linkerd/linkerd2/blob/stable-2.7.0/charts/linkerd2/values.yaml#L135-L153)

You're therefore left with an externally managed issuer cert, a manually managed one, or a mixture of the two.

I tried to naively replace the tls certs for these components, but the deployed application expects files that don't match a tls secret and thus cannot be updated. Updating the secret also leaves the `caBundle` in the webhook/apiservice unchanged and this breaks the api.

output?

```bash
$ kubectl apply --validate=false -f <cert-manager-manifest>.yml # Install cert manager
$ kubectl create ns linkerd
$ kubectl apply -f <cert-manager-linkerd-certs-manifest>.yml # create linkerd certificates
$ linkerd install --ha --identity-external-issuer \
        --tap-external-issuer \
        --proxy-injector-external-issuer \
        --sp-validator-external-issuer > manifest.yml
$ # Edit manifest.yml and annotate webhooks and apiserver with 
```

Edit `manifest.yml` and annotate webhooks and apiservice like so:

```yaml
...
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  name: linkerd-proxy-injector-webhook-config
  annotations: # This was added here to automate `caBundle`
    cert-manager.io/inject-ca-from: linkerd/linkerd-proxy-injector-tls
...
```

Now apply the manifest with the annotation updates

```bash
$ kubectl apply -f manifest.yml
```

This should fully automate all cert-manager components even as far as the trust anchor.

I've written up some code that attempts to do all this with kustomize, but with the annotations commented out until these features are possibly implemented in some form - https://github.com/misakwa/kustomized-linkerd

misakwa:
## Problem
Only some tls components of linkerd are currently fully automated. The identity
issuer certificate can be externalized, but the trust anchors and other cert
resources - tap, proxy injector, service profile validator, and smi metrics are not.

## Solution
This change builds on some of the initial work on identity issuer to make it
possible to use cert-manager to automate all the tls components in order to
reduce the maintenance burden of linkerd tls rotation.
The installation of linkerd would have to be a two step approach, the first to
install all the tls certificates using cert-manager. The final step is to
generate the manifests with the option to skip creating any of the tls resources.
The tap, proxy injector, service profile, and smi metrics webhooks and api
services will then have to be updated to allow cert-manager to inject the
`caBundle` into those resources.

The solution does the following in order to achieve full automation:

- [x] Update chart secret types to `kubernetes.io/tls` to keep it compatible with external issuers
- [x] Add trust anchor `ca.crt` to issuer secret to keep it consistent with external issuers
- [x] Add `tlsManager` variable to chart variables
- [x] Update chart secrets to be conditioned on `tlsManager`
- [x] Update chart `caBundle` to be conditioned on `tlsManager`
- [x] Update install cli call to set `tlsManager` using `--tls-manager`
- [x] Update upgrade cli call to use previously configured `tlsManager`
- [ ] Update testing

## Validation
Run bin/docker-build to build all the components and import the images into
minikube (or k3s).

Install cert-manager and then install the linkerd certificates that will not be
generated in the manifests.  There are kustomization manifests to install
cert-manager and the linkerd certificates:

- https://github.com/misakwa/linkerd-kustomized/tree/0.2-pre/cert-manager
- https://github.com/misakwa/linkerd-kustomized/tree/0.2-pre/certs.

Generate the linkerd installation manifests with the newly added cli option
like so:

```bash
$ linkerd install --tls-manager=cert-manager.io > install.yml
```

Edit install.yml and add annotations to inject the `caBndle` into webhooks and api services.

```yaml
...
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  name: linkerd-proxy-injector-webhook-config
    annotations:
        cert-manager.io/inject-ca-from: linkerd/linkerd-proxy-injector-tls # Add this
...
---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  name: linkerd-sp-validator-webhook-config
    annotations:
        cert-manager.io/inject-ca-from: linkerd/linkerd-sp-validator-tls # Add this
...
---
apiVersion: apiregistration.k8s.io/v1
kind: APIService
metadata:
  name: v1alpha1.tap.linkerd.io
    annotations:
        cert-manager.io/inject-ca-from: linkerd/linkerd-tap-tls # Add this
...
---
apiVersion: apiregistration.k8s.io/v1
kind: APIService
metadata:
  name: v1alpha1.metrics.smi-spec.io
    annotations:
        cert-manager.io/inject-ca-from: linkerd/linkerd-smi-metrics-tls # Add this
...
```

These edits can easily be applied with kustomize.

Apply the manifest to install linkerd:

```bash
$ kubectl apply -f install.yml
```

## Conclusion

Fixes #4203


grampelberg:
This is awesome!

alpeb:
Thanks @misakwa !
A couple of questions, without me having a full understanding of all this yet:
- When cert-manager rotates the secret, I guess it also updates the `caBundle` in the webhook/apiserver configs, correct?
- The linkerd-injector service (and the sp-validator, tap service, etc) would need to implement a watch on those secrets, in order to restart themselves when it changes, right?

misakwa:
Hi @alpeb! Thanks for taking a look. 

> When cert-manager rotates the secret, I guess it also updates the caBundle in the webhook/apiserver configs, correct ?

The cert-manager doumentation indicates it keeps it up to date, but I"ll run a quick test to confirm

> The linkerd-injector service (and the sp-validator, tap service, etc) would need to implement a watch on those secrets, in order to restart themselves when it changes, right?

You're absolutely right! I'll make some changes to add those in.

alpeb:
Thanks @misakwa. I've added a comment into #4203 to better understand the motivation behind this change.

pkg/identity/service.go
```diff
@@ -73,10 +76,34 @@ func (svc *Service) Initialize() error {
...
 	return nil
 }
 
+func (svc *Service) updateAnchors(newAnchors *x509.CertPool) {
+	svc.issuerMutex.Lock()
+	svc.trustAnchors = newAnchors
+	log.Debug("Trust anchors have been updated")
+	svc.issuerMutex.Unlock()
+}
+
```
zaharidichev, line 106: Have you tested this codepath and ensured nothing actually breaks. When that anchor gets updated, it seems nothing takes care to update the anchors that are embedded in the injected proxies. next time the proxy fires a CSR, it will fail to get a new cert, which will most likely break things in a pretty bad way.

misakwa: Hi @zaharidichev! 

You're correct. I don't have anything yet to update the injected proxies with the new anchor. I think they receive the anchor values from env vars at the time of injection. Is there a safer way to restart all the injected proxies without any downtime?

Should the proxies watch the issuer secret to pull the trust anchor instead? We could then have them be updated without any downtime like we do for identity and I recently added for the webhooks.

Thoughts?

zaharidichev: So yes, having the proxies pick up the anchors is a pretty essential part of all this. The problem with them watching the fs for the secret being updated is that you will effectivelly need to mount that secret in pretty much every pod that has a proxy injected. I am not sure this is desirable. I have to think about that a bit more.

grampelberg: As updates are not guaranteed, you'll always loose traffic which is why we've not done this for the proxies so far.

misakwa:
@zaharidichev I understand what you mean. 

If we decide to still exclude the trust anchor from some automated rotation or is too difficult to achieve at this time then I think we could do the following:

* Document the need to having a very long lived trust anchor and instruct cluster operators to use that pattern
* Update the helm chart to generate a trust anchor that lives longer than a year
* Completely remove the trust anchor from the chart an have it become an internal property of the control plane.

Even though the long lived anchor is operationally easier it makes the components appear insecure - not that they are. The cluster operator will eventually have to rotate them and possibly run into the same problem of possibly disrupting traffic to some data plane components.

I am not a big fan of the last two so I'd like to propose another option to managing the trust anchor:
* We could update the public api to return the trust anchor from the issuer secret since I've managed to normalize them - the trust anchor can be omitted from the configmap like the issuer.
* Every component can then query the public api for the trust anchor as well as the issuer certificate and update themselves if it has changed. 
* All environment variables that were used to store the trust anchor will have to be deprecated somehow or removed entirely.
* We can still use the same credential watcher to update the identity and webhook components
* The proxy will have to be updated to use the anchor value from the api call and update itself if it has changed. The proxy can use the same polling cycle it uses to renew the issuer certs to do this.

The advantages of this approach include

1. The trust anchor will now be only present in a single place and can only be retrieved through the public api thus making it easier to manage even if manually.
2. Every component can get the most recent trust anchor without having a mount of the secret or an environment variable.
3. Both externally and internally managed issuer certs and trust anchors will live in the public api as they have been, but can be cleaned from env vars and other sources.

Thoughts?

cc @zaharidichev, @grampelberg,  @alpeb

grampelberg:
> The proxy will have to be updated to use the anchor value from the api call and update itself if it has changed. The proxy can use the same polling cycle it uses to renew the issuer certs to do this.

@misakwa won't this cause downtime? The proxy is refreshing certs at 24 hours right now. If there's a trust root rotation in that 24 hour period, any new proxies started would no longer be able to talk to old proxies.

AFAIK *any* rotation of the trust root without doing it as a multi-step process will cause downtime. There's just no way to refresh all the certificates on all the proxies at the same instant.

The trust root in this situation is extremely similar to the trust roots of the internet. Looking at GitHub's (Digicert) they have an intermediate (aka issuer cert) that expires in 2028 and a trust root that expires in 2031.

I think we can make it possible to rotate the trust root but the only solution I'm aware of that's possible without downtime requires an operator to automate a dual trust root situation and then slowly roll the entire data plane to the new root (which can be done without actually restarting pods, but will require more API surface area that we should think through).

zaharidichev:
@misakwa I think we need to uphold the guarantee that the trust root is not rotated automatically. You can get into a lot of racy situations with this approach. We need to always be in a situation where our certs are valid in order to guarantee the lack of downtime (possibly through the use ofa  dual roots setup as noted by @grampelberg )

Also there is the second problem. If we do all that part of our infrastructure relies on the identitiy issuer secret being changed when we have new certs (which does not introduce a hard dependency on cert manager). However on the other hand we leave it to cert-manager to rotate that root that is in our webhook configs (which introduces a hard dependency on cert manager). We do not really want that. 

I think externalizing the tls resources is great, and we certainly want to do that. The only constraint is that we would like to keep the trust anchor out of this automatic rotation cycle and assume it does not change.

misakwa:
@grampelberg 

> @misakwa won't this cause downtime? The proxy is refreshing certs at 24 hours right now. If there's a trust root rotation in that 24 hour period, any new proxies started would no longer be able to talk to old proxies.

You're right. It will cause some downtime and requires a multi-step approach to avoid any disruption at all.

@zaharidichev I'll exclude the trust root from rotation in order to uphold that guarantee. I'd recommend making the default expiration of the trust root longer though since one year is pretty short IMHO. We should also update the documentation to include notes about this since it wasn't very obvious to me at first - I assume others might run into the same issue and questions about the trust root.

@grampelberg, @zaharidichev 
In order to make it easier to clean up usage of the trust root however, are you against centralizing its location? It is currently present in the configmap as well as env vars of a few components. I'd suggest we keep it in the issuer secret and have it only accessible from the api.

Let me know what you think.

cypherfox:
@misakwa I still cannot grasp from your code how you plan to rotate the non-identity CA certs (these are trust anchors as well). Without a multi-step approach you will always have a downtime. 

Also, a trust root validity period of one year is too long, since there is no support for revocation information and it doesn't work anyway. I would rather shorten it. Also, making it longer will only increase the pain afterwards.

grampelberg:
@misakwa that sounds sane for the issuer secret, but isn't it already centralized?

misakwa:
Hi @grampelberg 

> that sounds sane for the issuer secret, but isn't it already centralized?

I am referring to the trust anchor. It is currently available when one queries the controller api for the identity context. It is also available in the linkerd configmap, and as an environment variable.

I guess its not a big deal and we can leave that for later cleanups in order to keep this patch focused on externalizing the certs that we agreed on.

misakwa:
Hello @cypherfox 

> I still cannot grasp from your code how you plan to rotate the non-identity CA certs (these are trust anchors as well). Without a multi-step approach you will always have a downtime.

You're correct that there will be downtime for the webhook CA certs - even if very minimal. The current process of using helm to upgrade the chart will also upgrade the certs in one step, and the linkerd cli doesn't have the option to specify those certificates either. Externalizing them this way provides the maintainer with more options. They could bundle the old and new certs then wait for the components to pick them up before finally unbundling to remove the old cert.

Like its been said earlier cert-manager is not a requirement to rotation, the secrets containing the certificates just need to exist before creation/updating of the linkerd resources.

cypherfox:
> You're correct that there ~will~could be downtime for the webhook CA certs - even if very minimal. 

In a bigger cluster long enough to block pod creations. We just did that and actually had that problem.

> The current process of using helm to upgrade the chart will also upgrade the certs in one step, and the linkerd cli doesn't have the option to specify those certificates either. Externalizing them this way provides the maintainer with more options. 

I think there is a simpler way (see #4289) to take the complete rotation logic out of linkerd (and its helm chart).

The next step will be to add a way to set arbitrary annotations on the API extensions and webhooks. Which, of course, I would immediately use to tell cert-manager to inject up-to-date CA certs.

> They could bundle the old and new certs then wait for the components to pick them up before finally unbundling to remove the old cert.

Currently you cannot do that with the helm chart, as cert and caBundle are filled with the same field.

misakwa:
@cypherfox I see what you mean. You want to allow the bundle to be configured independently of the secret to allow the old and new bundle to be set there first before the secret is updated. That should allow the two-step approach to rotation that is missing from safely rotating the secret.

If you don't mind. I'll add in your changes to configure the caBundle independent of the key and crt. It should get us very close to finishing this up.
The final step after that change will be to add the annotations.

cypherfox:
@misakwa In my changes there still is the problem of reading from an APIService resource in `cli/cmd/upgrade.go`. I have asked on slack for the preferred solution.

olix0r:
@misakwa We definitely appreciate you open PR, but I worry that this PR is too large for us to merge it as it is. Practically, we'll need to break this down into smaller, targeted PRs that are merged independently. For instance, I believe #4289, which has been merged since this was opened, solves one of the problems this PR addresses.

I'm going to close this PR out for now as it's been inactive; but we'd love to help you make improvements to the project. To maximize your changes of getting things merge, I encourage you to work out your plans in an issue first so the team can help you strategize about how to get those changes merged. Thanks!

misakwa:
Thanks @olix0r.

PR #4233, opened 2020-04-03
Add liveliness checks and metrics for multicluster gateway

zaharidichev:
This PR adds liveliness checks for remote gateways. The way its is done is by maintaining probes for each gateway. The probes track whether the gateway is alive by issuing an HTTP request to an endpoint. Additionally the probes publish Prometheus metrics such as number of mirrored services per gateway and histogram latencies. 

Next steps (coming in different PRs): 

- Externalize the port, path and period through  the `get-credentials` command
- Add unit tests
- Enable Prometheus to scrape the `/metrics` endpoint
- Enable users to retreive the stats of the gateways via a CLI command (similar to `stat`)

### DIFF
--- a/charts/linkerd2/templates/prometheus.yaml
+++ b/charts/linkerd2/templates/prometheus.yaml
@@ -81,6 +81,19 @@ data:
         action: replace
         target_label: component
 
+    - job_name: 'linkerd-service-mirror'
+      kubernetes_sd_configs:
+      - role: pod
+      relabel_configs:
+      - source_labels:
+        - __meta_kubernetes_pod_label_linkerd_io_control_plane_component
+        - __meta_kubernetes_pod_container_port_name
+        action: keep
+        regex: linkerd-service-mirror;admin-http$
+      - source_labels: [__meta_kubernetes_pod_container_name]
+        action: replace
+        target_label: component
+
     - job_name: 'linkerd-proxy'
       kubernetes_sd_configs:
       - role: pod
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -438,7 +438,9 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceUpdated(ev *RemoteSe
 	serviceInfo := fmt.Sprintf("%s/%s", ev.remoteUpdate.Namespace, ev.remoteUpdate.Name)
 	rcsw.log.Debugf("Updating remote mirrored service %s/%s", ev.localService.Namespace, ev.localService.Name)
 
-	if ev.localEndpoints.Labels[consts.RemoteGatewayNameLabel] != ev.gatewayData.Name {
+	gatewayChanged := false
+	if ev.localEndpoints.Labels[consts.RemoteGatewayNameLabel] != ev.gatewayData.Name || ev.localEndpoints.Labels[consts.RemoteGatewayNsLabel] != ev.gatewayData.Namespace {
+		gatewayChanged = true
 		rcsw.probeChan <- &MirroredServiceUnpaired{
 			serviceName:      ev.localService.Name,
 			serviceNamespace: ev.localService.Namespace,
@@ -464,6 +466,22 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceUpdated(ev *RemoteSe
 			ev.localEndpoints.Annotations[consts.RemoteGatewayIdentity] = gatewayIdentity
 		} else {
 			delete(ev.localEndpoints.Annotations, consts.RemoteGatewayIdentity)
+		}
+
+		if gatewayChanged {
+			rcsw.probeChan <- &MirroredServicePaired{
+				serviceName:      ev.localService.Name,
+				serviceNamespace: ev.localService.Namespace,
+				GatewayProbeSpec: &GatewayProbeSpec{
+					clusterName:   rcsw.clusterName,
+					gatewayName:   ev.gatewayData.Name,
+					gatewayNs:     ev.gatewayData.Namespace,
+					gatewayIps:    endpointAddressesToIps(gatewayEndpoints),
+					port:          rcsw.probePort,
+					path:          rcsw.probePath,
+					periodSeconds: rcsw.probePeriodSeconds,
+				},
+			}
 		}
 
 	} else {
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -37,7 +37,7 @@ type (
 		probePort          int32
 		probePath          string
 		probePeriodSeconds int32
-		probeChan          chan<- interface{}
+		enqueueProbeEvent  func(event interface{})
 	}
 
 	// RemoteServiceCreated is generated whenever a remote service is created Observing
@@ -201,7 +201,7 @@ func NewRemoteClusterServiceWatcher(
 	probePort int32,
 	probePath string,
 	probePeriodSeconds int32,
-	probeChan chan<- interface{},
+	enqueueProbeEvent func(event interface{}),
 ) (*RemoteClusterServiceWatcher, error) {
 	remoteAPI, err := k8s.InitializeAPIForConfig(cfg, false, k8s.Svc)
 	if err != nil {
@@ -223,7 +223,7 @@ func NewRemoteClusterServiceWatcher(
 		probePort:          probePort,
 		probePath:          probePath,
 		probePeriodSeconds: probePeriodSeconds,
-		probeChan:          probeChan,
+		enqueueProbeEvent:  enqueueProbeEvent,
 	}, nil
 }
 
@@ -334,12 +334,12 @@ func (rcsw *RemoteClusterServiceWatcher) cleanupOrphanedServices() error {
 		return RetryableError{errors}
 	}
 
-	rcsw.probeChan <- &ClusterRegistered{
+	rcsw.enqueueProbeEvent(&ClusterRegistered{
 		clusterName:   rcsw.clusterName,
 		port:          rcsw.probePort,
 		path:          rcsw.probePath,
 		periodSeconds: rcsw.probePeriodSeconds,
-	}
+	})
 	return nil
 }
 
@@ -422,13 +422,13 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceDeleted(ev *RemoteSe
 	}
 
 	rcsw.log.Debugf("Successfully deleted Service: %s/%s", ev.Namespace, localServiceName)
-	rcsw.probeChan <- &MirroredServiceUnpaired{
+	rcsw.enqueueProbeEvent(&MirroredServiceUnpaired{
 		serviceName:      localServiceName,
 		serviceNamespace: ev.Namespace,
 		gatewayName:      ev.GatewayData.Name,
 		gatewayNs:        ev.GatewayData.Namespace,
 		clusterName:      rcsw.clusterName,
-	}
+	})
 	return nil
 }
 
@@ -441,13 +441,13 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceUpdated(ev *RemoteSe
 	gatewayChanged := false
 	if ev.localEndpoints.Labels[consts.RemoteGatewayNameLabel] != ev.gatewayData.Name || ev.localEndpoints.Labels[consts.RemoteGatewayNsLabel] != ev.gatewayData.Namespace {
 		gatewayChanged = true
-		rcsw.probeChan <- &MirroredServiceUnpaired{
+		rcsw.enqueueProbeEvent(&MirroredServiceUnpaired{
 			serviceName:      ev.localService.Name,
 			serviceNamespace: ev.localService.Namespace,
 			gatewayName:      ev.localEndpoints.Labels[consts.RemoteGatewayNameLabel],
 			gatewayNs:        ev.localEndpoints.Labels[consts.RemoteGatewayNsLabel],
 			clusterName:      rcsw.clusterName,
-		}
+		})
 	}
 
 	gatewayEndpoints, gatewayPort, resVersion, gatewayIdentity, err := rcsw.resolveGateway(ev.gatewayData)
@@ -469,7 +469,7 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceUpdated(ev *RemoteSe
 		}
 
 		if gatewayChanged {
-			rcsw.probeChan <- &MirroredServicePaired{
+			rcsw.enqueueProbeEvent(&MirroredServicePaired{
 				serviceName:      ev.localService.Name,
 				serviceNamespace: ev.localService.Namespace,
 				GatewayProbeSpec: &GatewayProbeSpec{
@@ -481,7 +481,7 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceUpdated(ev *RemoteSe
 					path:          rcsw.probePath,
 					periodSeconds: rcsw.probePeriodSeconds,
 				},
-			}
+			})
 		}
 
 	} else {
@@ -583,7 +583,7 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceCreated(ev *RemoteSe
 			endpointsToCreate.Annotations[consts.RemoteGatewayIdentity] = gatewayIdentity
 		}
 
-		rcsw.probeChan <- &MirroredServicePaired{
+		rcsw.enqueueProbeEvent(&MirroredServicePaired{
 			serviceName:      serviceToCreate.Name,
 			serviceNamespace: serviceToCreate.Namespace,
 			GatewayProbeSpec: &GatewayProbeSpec{
@@ -595,7 +595,7 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceCreated(ev *RemoteSe
 				path:          rcsw.probePath,
 				periodSeconds: rcsw.probePeriodSeconds,
 			},
-		}
+		})
 
 	} else {
 		rcsw.log.Warnf("Could not resolve gateway for %s: %s, skipping subsets", serviceInfo, err)
@@ -652,7 +652,7 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteGatewayDeleted(ev *RemoteGa
 func (rcsw *RemoteClusterServiceWatcher) handleRemoteGatewayUpdated(ev *RemoteGatewayUpdated) error {
 	rcsw.log.Debugf("Updating %d services due to remote gateway [%s/%s] update", len(ev.affectedServices), ev.gatewayData.Namespace, ev.gatewayData.Name)
 
-	rcsw.probeChan <- &GatewayUpdated{
+	rcsw.enqueueProbeEvent(&GatewayUpdated{
 		GatewayProbeSpec: &GatewayProbeSpec{
 			clusterName:   rcsw.clusterName,
 			gatewayName:   ev.gatewayData.Name,
@@ -662,7 +662,7 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteGatewayUpdated(ev *RemoteGa
 			path:          rcsw.probePath,
 			periodSeconds: rcsw.probePeriodSeconds,
 		},
-	}
+	})
 
 	var errors []error
 	for _, svc := range ev.affectedServices {
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -26,14 +26,18 @@ type (
 	// it can be requeued up to N times, to ensure that the failure is not due to some temporary network
 	// problems or general glitch in the Matrix.
 	RemoteClusterServiceWatcher struct {
-		clusterName     string
-		clusterDomain   string
-		remoteAPIClient *k8s.API
-		localAPIClient  *k8s.API
-		stopper         chan struct{}
-		log             *logging.Entry
-		eventsQueue     workqueue.RateLimitingInterface
-		requeueLimit    int
+		clusterName        string
+		clusterDomain      string
+		remoteAPIClient    *k8s.API
+		localAPIClient     *k8s.API
+		stopper            chan struct{}
+		log                *logging.Entry
+		eventsQueue        workqueue.RateLimitingInterface
+		requeueLimit       int
+		probePort          int32
+		probePath          string
+		probePeriodSeconds int32
+		probeChan          chan<- interface{}
 	}
 
 	// RemoteServiceCreated is generated whenever a remote service is created Observing
@@ -57,8 +61,9 @@ type (
 	// RemoteServiceDeleted when a remote service is going away or it is not
 	// considered mirrored anymore
 	RemoteServiceDeleted struct {
-		Name      string
-		Namespace string
+		Name        string
+		Namespace   string
+		GatewayData *gatewayMetadata
 	}
 
 	// RemoteGatewayDeleted is observed when a service that is a gateway to at least
@@ -187,7 +192,17 @@ func (rcsw *RemoteClusterServiceWatcher) resolveGateway(metadata *gatewayMetadat
 }
 
 // NewRemoteClusterServiceWatcher constructs a new cluster watcher
-func NewRemoteClusterServiceWatcher(localAPI *k8s.API, cfg *rest.Config, clusterName string, requeueLimit int, clusterDomain string) (*RemoteClusterServiceWatcher, error) {
+func NewRemoteClusterServiceWatcher(
+	localAPI *k8s.API,
+	cfg *rest.Config,
+	clusterName string,
+	requeueLimit int,
+	clusterDomain string,
+	probePort int32,
+	probePath string,
+	probePeriodSeconds int32,
+	probeChan chan<- interface{},
+) (*RemoteClusterServiceWatcher, error) {
 	remoteAPI, err := k8s.InitializeAPIForConfig(cfg, false, k8s.Svc)
 	if err != nil {
 		return nil, fmt.Errorf("cannot initialize remote api for cluster %s: %s", clusterName, err)
@@ -203,8 +218,12 @@ func NewRemoteClusterServiceWatcher(localAPI *k8s.API, cfg *rest.Config, cluster
 			"cluster":    clusterName,
 			"apiAddress": cfg.Host,
 		}),
-		eventsQueue:  workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
-		requeueLimit: requeueLimit,
+		eventsQueue:        workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		requeueLimit:       requeueLimit,
+		probePort:          probePort,
+		probePath:          probePath,
+		probePeriodSeconds: probePeriodSeconds,
+		probeChan:          probeChan,
 	}, nil
 }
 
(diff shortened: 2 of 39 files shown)


zaharidichev:
Ahh, som unit tests are locking up. Will address that shortly

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -0,0 +1,194 @@
...
+		worker.IncNumServices()
+		m.probeWorkers[probeKey] = worker
+		worker.Start()
+	}
+}
+
+func (m *ProbeManager) handleMirroredServiceUnpaired(event *MirroredServiceUnpaired) {
+	probeKey := probeKey(event.gatewayNs, event.gatewayName, event.clusterName)
+	worker, ok := m.probeWorkers[probeKey]
+	if ok {
```
kleimkuhler, line 110: Would it be helpful to log the case that a service unpairing event occurs, but there is not a worker for that gateway/cluster?

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -0,0 +1,194 @@
...
+			worker.Stop()
+			delete(m.probeWorkers, probeKey)
+		}
+	}
+}
+
+func (m *ProbeManager) handleGatewayUpdated(event *GatewayUpdated) {
+	probeKey := probeKey(event.gatewayNs, event.gatewayName, event.clusterName)
+	worker, ok := m.probeWorkers[probeKey]
+	if ok {
```
kleimkuhler, line 123: Similar question about logging the case that a worker does not exist for an event.

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+		metrics:               metrics,
+		log: logging.WithFields(logging.Fields{
+			"probe-key": probekey,
+		}),
+	}
+}
+
+// IncNumServices increments the number of services that are routed by the gateway
+func (pw *ProbeWorker) IncNumServices() {
+	pw.numAssociatedServices = pw.numAssociatedServices + 1
```
kleimkuhler, line 39: If we need to worry about synchronizing probe spec updates, should also do that for incrementing and decrementing the services?

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+	pw.metrics.services.Inc()
+}
+
+// DcrNumServices decrements the number of services that are routed by the gateway
+func (pw *ProbeWorker) DcrNumServices() {
+	pw.numAssociatedServices = pw.numAssociatedServices - 1
+	pw.metrics.services.Dec()
+}
+
+// UpdateProbeSpec  is used to update the probe specification when something about the gateway changes
```
kleimkuhler, line 49: extra space:
```suggestion
// UpdateProbeSpec is used to update the probe specification when something about the gateway changes
```

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+func (pw *ProbeWorker) UpdateProbeSpec(spec *GatewayProbeSpec) {
+	pw.Lock()
+	pw.probe = spec
+	pw.Unlock()
+}
+
+// Stop stops this probe worker
```
kleimkuhler, line 56: typo:
```suggestion
// Stop this probe worker
```

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+func (pw *ProbeWorker) Stop() {
+	pw.log.Debug("Stopping probe worker for")
+	close(pw.stopCh)
+}
+
+// Start starts this probe worker
```
kleimkuhler, line 62: typo:
```suggestion
// Start this probe worker
```

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (pw *ProbeWorker) doProbe() {
+	pw.RLock()
+	defer pw.RUnlock()
+
+	ipToTry := pw.pickAnIP()
+	if ipToTry == "" {
+		pw.log.Debug("No ips. Marking as unhealthy")
+		pw.metrics.alive.Set(0)
```
kleimkuhler, line 102: This (as it currently is) should not worry about closing the worker's stop channel correct? It should just mark the gateway as unhealthy but not necessarily stop probing it?

controller/cmd/service-mirror/config_watcher.go
```diff
@@ -105,63 +107,81 @@ func (rcw *RemoteClusterConfigWatcher) Stop() {
...
 	}
-	return config, clusterName, domain, nil
+
+	//TODO Externalize the port,path and period through  the get-credentials command
+	return &watchedClusterConfig{
+		apiConfig:          config,
+		clusterName:        clusterName,
+		clusterDomain:      domain,
+		probePort:          80,
+		probePath:          "/nginx-health",
```
kleimkuhler, line 184: I see this has a TODO, but should this still be a hard-coded path?

controller/cmd/service-mirror/main.go
```diff
@@ -6,17 +6,22 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/linkerd/linkerd2/pkg/admin"
+
```
kleimkuhler, line 10: nit: extra line

charts/linkerd2/templates/prometheus.yaml
```diff
@@ -81,6 +81,19 @@ data:
         action: replace
         target_label: component
 
+    - job_name: 'linkerd-service-mirror'
+      kubernetes_sd_configs:
+      - role: pod
```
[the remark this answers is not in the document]
zaharidichev: I do not think it does, because the linkerd-service-mirror is not always in the linkerd namespace

controller/cmd/service-mirror/probe_worker.go
zaharidichev: @kleimkuhler  Not sure we need to take the lock here. `IncNumServices` and as a result `numAssociatedServices` are only called/written/read from within probe_manager.go which processes events one at a time from its chan. The reason we synchronize the probe spec updates is because the latter can be modified by the probe_manager and are also read by the go routine that calls `doProbe()`. Does that sound right ? WDYT?

controller/cmd/service-mirror/probe_worker.go
zaharidichev: Yes this is correct

controller/cmd/service-mirror/config_watcher.go
zaharidichev: I was planning that to be a separate PR once we merge this as it touches the cluster CLI commands and  I did not want to make this P bigger. Is that ok ?

kleimkuhler: @zaharidichev Yep sounds good!

controller/cmd/service-mirror/probe_worker.go
kleimkuhler: Yep makes sense.

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -405,6 +438,16 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceUpdated(ev *RemoteSe
...
+	if ev.localEndpoints.Labels[consts.RemoteGatewayNameLabel] != ev.gatewayData.Name {
+		rcsw.probeChan <- &MirroredServiceUnpaired{
+			serviceName:      ev.localService.Name,
+			serviceNamespace: ev.localService.Namespace,
+			gatewayName:      ev.localEndpoints.Labels[consts.RemoteGatewayNameLabel],
+			gatewayNs:        ev.localEndpoints.Labels[consts.RemoteGatewayNsLabel],
+			clusterName:      rcsw.clusterName,
+		}
+	}
+
```
alpeb, line 450: When the remote service gets updated, you're issuing a `MirroredServiceUnpaired` event. Isn't a following `MirroredServicePaired` event required?

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+	} else {
+		start := time.Now()
+		resp, err := http.Get(fmt.Sprintf("http://%s:%d/%s", ipToTry, pw.probe.port, pw.probe.path))
```
alpeb, line 105: I don't know what's the default timeout for http.Get requests. You might wanna consider setting one manually here...

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -458,6 +501,17 @@ func remapRemoteServicePorts(ports []corev1.ServicePort) []corev1.ServicePort {
 	return newPorts
 }
 
+func endpointAddressesToIps(addrs []corev1.EndpointAddress) []string {
```
adleong, line 504: every time I read go code like this I cry sorrowful tears of yearning for `.map(_.IP)`

zaharidichev: every time I have to write code like this I feel the golang creators are actually praniking me (and you because you have to read it... )

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	*GatewayProbeSpec
+}
+
+// ClusterRegistered is emitted when a new cluster becomes registred for mirroring
+type ClusterRegistered struct {
+	clusterName   string
+	port          int32
+	path          string
+	periodSeconds int32
+}
```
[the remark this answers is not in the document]
zaharidichev: Hmmm. Yes very good catch!

controller/cmd/service-mirror/probe_worker.go
zaharidichev: I set it to 50 seconds since this is the upper bucket limit for the latency histogram of the metrics I gather

controller/cmd/service-mirror/cluster_watcher.go
zaharidichev: Yes.. that makes sense. For reference if the gateway for a particular service has changed to another gateway, we need to unpair and then pair again, potentially resulting in a new probe being created.

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,124 @@
...
+
+// IncNumServices increments the number of services that are routed by the gateway
+func (pw *ProbeWorker) IncNumServices() {
+	pw.numAssociatedServices = pw.numAssociatedServices + 1
+	pw.metrics.services.Inc()
+}
+
+// DcrNumServices decrements the number of services that are routed by the gateway
+func (pw *ProbeWorker) DcrNumServices() {
+	pw.numAssociatedServices = pw.numAssociatedServices - 1
```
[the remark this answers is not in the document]
zaharidichev: golint is shining this time: 

```
controller/cmd/service-mirror/probe_worker.go:41:2: should replace `pw.numAssociatedServices += 1` with `pw.numAssociatedServices++` (golint)
        pw.numAssociatedServices += 1
```

controller/cmd/service-mirror/config_watcher.go
```diff
@@ -105,63 +107,81 @@ func (rcw *RemoteClusterConfigWatcher) Stop() {
...
-		return nil, "", "", fmt.Errorf("secret should contain remote cluster name as annotation %s", consts.RemoteClusterNameLabel)
+		return nil, fmt.Errorf("secret should contain remote cluster name as annotation %s", consts.RemoteClusterNameLabel)
 	}
 	if !hasDomain {
-		return nil, "", "", fmt.Errorf("secret should contain remote cluster domain as annotation %s", consts.RemoteClusterDomainAnnotation)
+		return nil, fmt.Errorf("secret should contain remote cluster domain as annotation %s", consts.RemoteClusterDomainAnnotation)
 	}
-	return config, clusterName, domain, nil
+
+	//TODO Externalize the port,path and period through  the get-credentials command
```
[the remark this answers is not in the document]
zaharidichev: We had a chat with @grampelberg about that and reached the conclusion that, yes these are configured per cluster. And the most friction free way to add them in is to add them on the credentials secret. We can do that through the get-credentials command. This way we know for sure that we will always have these settings there (or some default). The alternative of adding annotations per gateway and checking what exists and what not seems like not such a great alternative. Also this brings the question of whether we might want to have different setting per gateway or they really should be the same per cluster.  I am open to suggestions.

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	"fmt"
+
+	"github.com/linkerd/linkerd2/controller/k8s"
+	consts "github.com/linkerd/linkerd2/pkg/k8s"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// NewProbeManager creates a new probe manager
+func NewProbeManager(events chan interface{}, k8sAPI *k8s.API) *ProbeManager {
```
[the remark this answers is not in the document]
zaharidichev: No because we send to it in `handleClusterRegistered`

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -0,0 +1,198 @@
...
+		worker.Start()
+	}
+}
+
+func (m *ProbeManager) handleMirroredServiceUnpaired(event *MirroredServiceUnpaired) {
+	probeKey := probeKey(event.gatewayNs, event.gatewayName, event.clusterName)
+	worker, ok := m.probeWorkers[probeKey]
+	if ok {
+		worker.DcrNumServices()
+		if worker.numAssociatedServices < 1 {
```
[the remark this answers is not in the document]
zaharidichev: The assumption is that we cannot. But we might have a bug somewhere that allows that. What do you suggest? Keep a set of associated services in the probe_worker and only decrement if this service that is getting unpaired is in there?

zaharidichev:
> These metrics seem cool! Can you explain what you mean by liveness checks? I see the code that publishes prometheus metrics based on the probes to each gateway, but I don't see anywhere that acts on that data.

Yes, so the liveliness checks do exactly that, just publish metrics. We do not actively act on these, the same way kubelet acts on the results of its liveliness checks. The idea behind that is to provide more info around the status of the gateways that we are talking to on remote clusters. This is supposed to work with dashboard + a cli command, similar to the stat one. Something like: 

Screenshot 2020-04-07 at 16 17 39

controller/cmd/service-mirror/probe_manager.go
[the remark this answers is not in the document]
adleong: I guess my question is how confident are we about this assumption?  At the very least, we should document that the contract here is that we must not get duplicate or out-of-order events.

If you trace it all the way back, these events are triggered by Kubernetes informer events.  Do informers make any guarantees about duplicate or out-of-order events?  Or does our handling of those events guarantee that events will be properly de-duped by the time they make it here?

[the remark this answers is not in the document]
zaharidichev: @adleong very good question! So yes, all the way back these events are triggered by the informer. And indeed you are right that we can get duplicates here. However, if you trace a bit forward these particular events are triggered by the cluster_watcher. The cluster watcher assumes that you might get duplicate events from the informers (i.e. if your watcher fails and you fall back to perioding listing). because of that it makes sure that we do not create or delete a service twice. So for example if you get a service created event for a service that you already have and that has not changed, it will not try to crete the service again and as a result it will not issue an event to the prober twice. Will look at it again, but as far as I remember, this is how it was designed. I will put a comment though. I think all this deserves a good dose of tests to verify that all these edge cases are handled correctly (which I will write in time). 

That being said, in my latest commit I transitioned to keeping a set of these services (well a map, as Go does not have sets...) to track the count. That alone should serve as a second gatekeeper against duplication. Does that sound correct ?

controller/cmd/service-mirror/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type probeMetricVecs struct {
+	services  *prometheus.GaugeVec
+	alive     *prometheus.GaugeVec
+	latencies *prometheus.HistogramVec
+}
+
+type probeMetrics struct {
+	services  prometheus.Gauge
```
adleong, line 19: Since we know the service name when it registers interest in the gateway, we could consider having a `service` label on this gauge.  This allows us to actually list all the services behind the gateway, instead of just knowing the number of services.

controller/cmd/service-mirror/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+	alive     prometheus.Gauge
```
adleong, line 20: in addition to having a gauge for "alive", we could also have a counter for "probes" with a label to indicate if the probe succeeded or failed.

controller/cmd/service-mirror/main.go
```diff
@@ -36,12 +40,18 @@ func Main(args []string) {
 		log.Fatalf("Failed to initialize K8s API: %s", err)
 	}
 
+	probeEvents := make(chan interface{}, probeChanBufferSize)
```
adleong, line 43: how easy is it to expose queue depth as a prometheus metric?  how can we tell if this queue is backing up?

adleong: a simple approach would be to expose the queue length `len(probeEvents)` as a gauge.

a more sophisticated approach would be to have counters of enqueues and dequeues.  The counter approach has the benefit that not only can we see queue depth by subtrating dequeues from enqueues, but we can also have a label for event type to see breakdowns by event type.

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,140 @@
...
+func (pw *ProbeWorker) NumPairedServices() int {
+	return len(pw.pairedServices)
+}
+
+// PairService increments the number of services that are routed by the gateway
+func (pw *ProbeWorker) PairService(serviceName, serviceNamespace string) {
+	svcKey := fmt.Sprintf("%s-%s", serviceNamespace, serviceName)
+	if _, ok := pw.pairedServices[svcKey]; !ok {
+		pw.pairedServices[svcKey] = true
+		pw.metrics.services.Inc()
```
adleong, line 49: do you think it would be more robust to do something like `pw.metrics.services.SetValue(len(pw.pairedServices))`?

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,140 @@
...
+	metrics        *probeMetrics
+	log            *logging.Entry
+}
+
+// NewProbeWorker creates a new probe worker associated with a particular gateway
+func NewProbeWorker(probe *GatewayProbeSpec, metrics *probeMetrics, probekey string) *ProbeWorker {
+	return &ProbeWorker{
+		RWMutex:        &sync.RWMutex{},
+		probe:          probe,
+		pairedServices: make(map[string]bool),
```
adleong, line 30: I like using `map[string]struct{}` for stuff like this.  The problem with storing a boolean value is that it's easy to make the mistake of setting the value to false when unpairing a service (rather than deleting the key) which would cause a bug.

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -0,0 +1,140 @@
...
+func (pw *ProbeWorker) Start() {
+	pw.log.Debug("Starting probe worker")
+	go pw.run()
+}
+
+func (pw *ProbeWorker) run() {
+	probeTickerPeriod := time.Duration(pw.probe.periodSeconds) * time.Second
+
+	// Introduce some randomness to avoid bursts of probes
+	time.Sleep(time.Duration(rand.Float64() * float64(probeTickerPeriod)))
```
adleong, line 85: Using time.Sleep and Ticker together is a bit confusing.  Also, this can sleep up to `probeTickerPeriod` making the time between probes potentially as long as 2*`probeTickerPeriod`.  

Consider instead a ticker with some random jitter like https://github.com/projectcalico/felix/blob/master/jitter/jittered_ticker.go

controller/cmd/service-mirror/metrics.go
zaharidichev: Hm, fairly new to the whole Prometheus thing, but is it possible/practical to have multi value labels.  This metric is 1 per gateway, not per service. So how can we do what you are describing?

controller/cmd/service-mirror/metrics.go
zaharidichev: I think a counter only goes up, right? And the number of probes is sublject to going down as well. What is the advantage of having this counter in addition the gauge. Does it give us something more? If we grab the gauges, we ultimatelly get the number of probes along with the information of whether the gateway is alive or not. 

Or  are you trying to measure the total number of attemptes to reach a remote cluster and partition them by success outcome?

controller/cmd/service-mirror/probe_worker.go
zaharidichev: Makes sense, however I think this only reflects the start point. The time between probes will be equal to `probeTickerPeriod`. It just that there is delay of when the probing starts.

zaharidichev: But fine, I will just grab this jittered ticker. Looks like a nice utility anyway.

zaharidichev:
@adleong I  incorporated your suggestions for additional metrics. Feel free to take another look :)

controller/cmd/service-mirror/metrics.go
adleong: I'm suggesting adding a label for service name. This would mean the gauge would only ever take values 1 (if the service with that name is served by that gateway) or 0 (if not).

controller/cmd/service-mirror/metrics.go
adleong: Yeah, that's right.  A total number of attempts, rather than a gauge of in-flight attempts.  This allows us to see the rate of probes over time.  For example, if for some reason probes stop happening, the alive gauge would stay at `1` and we would not be able to see that probes had stopped.  A counter would let us see this.

zaharidichev: This is done

zaharidichev:
@adleong not that $4265 introduces probe settings per gateway, I will have to think about how to handle this one here a bit better. The problem stems from the fact that we need to keep the data of what these settings are somewhere on the local cluster. The reason for that arises because... the prober tries to recrete its state on startup based on the services that are already mirrored and we want to have the correct probe settings, but we dont get any data about it once we start up. Probably will keep that as data in the local endpoints and take care to update them when something about the gateway changes (which we already do for other data such as gateway identitiy, etc.... )

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -26,14 +27,37 @@ type (
...
-		requeueLimit    int
+		clusterName        string
+		clusterDomain      string
+		remoteAPIClient    *k8s.API
+		localAPIClient     *k8s.API
+		stopper            chan struct{}
+		log                *logging.Entry
+		eventsQueue        workqueue.RateLimitingInterface
+		requeueLimit       int
+		probePort          int32
```
adleong, line 38: I think these probe fields are unused and shouldn't be part of the RCSW.  I think we decided that the probe configuration should be loaded from the gateway annotations and stored on the gateway spec, instead of being per cluster.

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -143,51 +164,29 @@ func (re RetryableError) Error() string {
...
 
 // NewRemoteClusterServiceWatcher constructs a new cluster watcher
-func NewRemoteClusterServiceWatcher(localAPI *k8s.API, cfg *rest.Config, clusterName string, requeueLimit int, clusterDomain string) (*RemoteClusterServiceWatcher, error) {
+func NewRemoteClusterServiceWatcher(
+	localAPI *k8s.API,
+	cfg *rest.Config,
+	clusterName string,
+	requeueLimit int,
+	clusterDomain string,
+	probePort int32,
```
adleong, line 185: ditto: I think these fields are unused and should be deleted.

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -26,14 +27,37 @@ type (
...
+		remoteAPIClient    *k8s.API
+		localAPIClient     *k8s.API
+		stopper            chan struct{}
+		log                *logging.Entry
+		eventsQueue        workqueue.RateLimitingInterface
+		requeueLimit       int
+		probePort          int32
+		probePath          string
+		probePeriodSeconds int32
+		enqueueProbeEvent  func(event interface{})
```
adleong, line 41: It might be more clear and concrete to hold a `*ProbeManager` instead of a totally generic `func(event interface{})`.  I don't think the generality really gets us anything useful since we only ever enqueue to a ProbeManager.

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -910,3 +962,61 @@ func (rcsw *RemoteClusterServiceWatcher) Stop(cleanupState bool) {
 	}
 	rcsw.eventsQueue.ShutDown()
 }
+
+func (rcsw *RemoteClusterServiceWatcher) extractGatewaySpec(gateway *corev1.Service) (*GatewaySpec, error) {
+	if len(gateway.Status.LoadBalancer.Ingress) == 0 {
+		return nil, errors.New("expected gateway to have at lest 1 external Ip address but it has none")
```
adleong, line 968: TIOLI: should this error message include the gateway service name/namespace?

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -910,3 +962,61 @@ func (rcsw *RemoteClusterServiceWatcher) Stop(cleanupState bool) {
...
+			IP:       ingress.IP,
+			Hostname: ingress.Hostname,
+		})
+	}
+
+	gatewayIdentity := gateway.Annotations[consts.GatewayIdentity]
+
+	var probeCfg *ProbeConfig
+
+	probePath := gateway.Annotations[consts.GatewayProbePath]
```
adleong, line 997: should we log if any of these errs are not nil?  if we move all the parsing of the ProbeConfig into a separate method that returns `(*ProbeConfig, error)` we can be a bit more idiomatic about error handling and avoid nesting if statements so deeply.

controller/cmd/service-mirror/config_watcher.go
```diff
@@ -105,63 +107,81 @@ func (rcw *RemoteClusterConfigWatcher) Stop() {
...
 
 	return nil
 }
 
-func parseRemoteClusterSecret(secret *corev1.Secret) ([]byte, string, string, error) {
+type watchedClusterConfig struct {
+	apiConfig          []byte
+	clusterName        string
+	clusterDomain      string
+	probePort          int32
```
adleong, line 159: I don't think these probe settings should be part of this cluster config anymore

controller/cmd/service-mirror/metrics.go
```diff
@@ -0,0 +1,116 @@
...
+		prometheus.GaugeOpts{
+			Name: "gateway_alive",
+			Help: "A gauge which is 1 if the gateway is alive and 0 if it is not.",
+		},
+		labelNames,
+	)
+
+	latencies := promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "gateway_request_latency_ms",
```
adleong, line 77: Should we call this `gateway_probe_latency_ms` to clarify that this is the latency of the probe requests as opposed to latency of data path requests?

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -0,0 +1,239 @@
...
+type MirroredServicePaired struct {
+	serviceName      string
+	serviceNamespace string
+	GatewaySpec
+}
+
+// ClusterNotRegistered is is emitted when the cluster is not monitored anymore
+type ClusterNotRegistered struct {
+	clusterName string
+}
```
alpeb, line 46: I couldn't find any place where this event gets emitted...

controller/cmd/service-mirror/cluster_watcher.go
zaharidichev: I wanted to keep it like that for test purposes. Namely., i want to exercise a bunch of events coming from the informer and expect that a bunch of other events will be send to the ProbeManager. And in tests, I think it might be cleaner to just provide a function that buffers them rather than mocking the ProbeManager (for some definition of Mocking that Go adopts... )

adleong: What do you think about creating a `ProbeEventSink` interface which defines just a `send` method?  This is basically exactly what you already have, but using an interface feels a bit more like idiomatic go than passing around an anonymous function.  TIOLI.

PR #4235, opened 2020-04-04
cli: error out when `linkerd inject` attempts to inject non-compliant pods

### LINKED ISSUE #4214: inject: throw an error while attempting to inject non-compliant pods
This issue has been opened after following up with the discussion on https://github.com/linkerd/linkerd2/pull/4145

### Current behaviour
At present, `linkerd inject` outputs an uninjected YAML configuration while attempting to inject non-compliant pods, i.e, pods with issues related to  hostNetwork, sidecar, udp, etc. As this output is usually piped to `kubectl`, this causes an uninjected workload to get deployed on the cluster.

### Desired behaviour
`linkerd inject` must simply error out with a meaningful message related to the non-compliance of the input config 

cc @grampelberg @alpeb @adleong

mayankshah1607:
Fixes #4214

This PR updates the behaviour of `linikerd inject` while attempting to inject pods that are non-compliant. 

As compared to before, instead of showing warnings and writing an unchanged YAML config to `stdout` (which is piped to `kubectl`; as a consequence, pods that fail the injection process are deployed onto the cluster), `inject` simply errors out by displaying some valuable message.

For example:
```bash
$ cat hostNetwork_true.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: emojivoto
spec:
  replicas: 1
  selector:
    matchLabels:
      app: web-svc
  template:
    metadata:
      labels:
        app: web-svc
    spec:
      containers:
      - env:
        - name: WEB_PORT
          value: "80"
        - name: EMOJISVC_HOST
          value: emoji-svc.emojivoto:8080
        - name: VOTINGSVC_HOST
          value: voting-svc.emojivoto:8080
        - name: INDEX_BUNDLE
          value: dist/index_bundle.js
        image: buoyantio/emojivoto-web:v3
        name: web-svc
        ports:
        - containerPort: 9100
          hostPort: 9100
          name: http
      hostNetwork: true
---

$ cat hostNetwork_true.yaml | bin/go-run cli inject - | k apply -f -

Error transforming resources: Resource with name "web" of kind "deployment" cannot be injected due to the following reasons:

× hostNetwork is enabled
```


cli/cmd/testdata/inject_emojivoto_deployment_empty_resources.golden.yml
```diff
@@ -1,163 +0,0 @@
-apiVersion: apps/v1
```
mayankshah1607, line 1: The reason why this file is expected to be empty is that if we pass a list of configs (delimited by `---`), and even if any one of the specs is non-compliant, `inject` does not output the remaining configs with the inject annotation present (that are expected to be piped to `kubectl`)
If this is non-trivial or should be regarded as unexpected behaviour, we may have to update the way `transformInput()` and `processYAML()` work, so that they may output the configs that were successfully annotated and show an error for the rest

cli/cmd/testdata/inject_emojivoto_deployment_injectDisabled.report.verbose
```diff
@@ -1,9 +1,3 @@
...
 
-√ pods do not use host networking
-√ pods do not have a 3rd party proxy or initContainer already injected
-‼ "linkerd.io/inject: disabled" annotation set on deployment/web
-‼ no supported objects found
-√ pod specs do not include UDP ports
-
-deployment "web" skipped
-
+×
```
mayankshah1607, line 3: I also happened to notice a very strange bug here. On debugging, I observed that the `Report` object that is returned by the method `newReport()` does hold a string value `"injection_enable_annotation_absent"` under the field `report.InjectDisabledReason`. However, when this report object is returned as a part of `ParseMetaAndYAML()` method and assigned to the variable `report` (in particular, we are looking at [this](https://github.com/linkerd/linkerd2/blob/be558b68694f6e64679562052869d3f7ddd68261/cli/cmd/inject.go#L154) line), `report.InjectDisabledReason` now holds an empty string.
Hence, the method `report.GetInjectFailReason()` also renders an empty string as a part of the error message due to the direct dependency on that particular field.
I'm not really sure if this behaviour would be expected, but it does look like `disableByAnnotation()` method does account for this particular test case,

cli/cmd/testdata/inject_emojivoto_deployment_empty_resources.golden.yml
Pothulapati: IMO, I don't think its non-trivial, As users would mostly group together resources in a file, and pass it to inject. Better to be pass or fail, so that users can correct rather than leaving in a confused install state. (even if we show errors for the rest)

cli/cmd/testdata/inject_emojivoto_deployment_injectDisabled.report.verbose
Pothulapati: This is pretty weird. Debugging the workflow, showed me that `transform` is being called by twice for the same yaml object once as with `origin` as `CLI` and in other-case with a different value. The `disabledByAnnotation()` of the differentValue is returning as `false`, causing it to not print the reason.

The question is why is transform for a single yaml object twice? Will try to do more debugging on this early tomorrow.

Pothulapati: Okay, Once it was `uninject` which is called before `inject` but uses the same `processYaml` functions with a different `resourceTransformer`

Pothulapati:
A lot of folks in the community (including myself) really like the verbose messages that the CLI showed.

I don't really have a problem with us showing all the warning messages, with the extra error message (with the resourcekind/resourceName) and error'ing out. 
WDYT? @alpeb

pkg/inject/report.go
```diff
@@ -130,6 +133,16 @@ func (r *Report) Injectable() (bool, []string) {
 	return true, nil
 }
 
+//GetInjectFailReason is used to generate a string that lists the reasons why the injection process
+//may have failed. This method must be used along with Injectable() to generate readable errors
```
alpeb, line 137: Nit: leading space in comments

cli/cmd/testdata/inject_emojivoto_deployment_injectDisabled.report.verbose
zaharidichev: So @Pothulapati @mayankshah1607,  inject generally is (uninject + inject) under the hood. The reason being is that you can actually do a manual inject on an already injected resource only to change something about its injected config. Does that make sense?

pkg/inject/report.go
```diff
@@ -130,6 +133,16 @@ func (r *Report) Injectable() (bool, []string) {
...
+func (r *Report) GetInjectFailReason(reasons []string) error {
+	desc := fmt.Sprintf("Resource with name \"%s\" of kind \"%s\" cannot be injected due to the following reasons:\n", r.Name, r.Kind)
+	for _, reason := range reasons {
+		desc += fmt.Sprintf("\n%s %s", failStatus, Reasons[reason])
+	}
+	return fmt.Errorf(desc)
```
zaharidichev, line 143: cant you just do errors.New() as you are not really doing any formatting here.

mayankshah1607:
@alpeb @Pothulapati I have updated the code to show the verbose on completing the injection process, and also the errors are getting accumulated into the final report instead of abruptly exiting the injection process.
A new message for `automountServiceAccountToken` has been introduced in the `generateReports()` method, hence causing the integration tests to fail. I'll update those too once we can agree that the new behaviour is desirable. :)

cli/cmd/inject.go
```diff
@@ -28,8 +28,8 @@ const (
 	injectDisabledDesc               = "pods are not annotated to disable injection"
 	unsupportedDesc                  = "at least one resource injected"
 	udpDesc                          = "pod specs do not include UDP ports"
+	automountServiceAccountTokenDesc = "pods are allowed to automount API credentials"
```
alpeb, line 31: I would prefer to be more precise here with the message mentioning service account tokens are automounted, instead of talking about API credentials.

grampelberg:
Ping @mayankshah1607, this'd be awesome to get in!

mayankshah1607:
Closing this one as I had issues with rebasing the tests files on this branch. Opened a cleaner version incorporating feedback from @Pothulapati and @alpeb  - https://github.com/linkerd/linkerd2/pull/4346

PR #4240, opened 2020-04-07
Gate SMI-Metrics behind an install flag

adleong:
This change adds a `--smi-metrics` install flag which controls if the SMI-metrics controller and associated RBAC and APIService resources are installed.  The flag defaults to false and is hidden.

We plan to remove this flag or default it to true if and when the SMI-Metrics integration graduates from experimental.


cli/cmd/install.go
```diff
@@ -521,11 +523,16 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 		&options.addOnConfig, "addon-config", options.addOnConfig,
 		"A path to a configuration file of add-ons",
 	)
+	flags.BoolVar(
+		&options.smiMetricsEnabled, "smi-metrics", options.smiMetricsEnabled,
+		"Enables installing the SMI-Metrics controller",
+	)
 
 	flags.StringVarP(&options.controlPlaneVersion, "control-plane-version", "", options.controlPlaneVersion, "(Development) Tag to be used for the control plane component images")
 	flags.StringVar(&options.smiMetricsImage, "smi-metrics-image", options.smiMetricsImage, "SMI Metrics image")
```
alpeb, line 532: I think this one needs to be hidden as well

PR #4241, opened 2020-04-08
Add gateways command, allowing the retrieval of gateway stats

zaharidichev:
This PR adds a `cluster gateways` command that allows for observing stats of remote cluster gateways as long as they are present in Prometheus. The output of the command looks like: 

```
CLUSTER   NAMESPACE   NAME              NUM SVC   ALIVE   LATENCY_P50   LATENCY_P95   LATENCY_P99
remote    default     another-gateway         1    True         275ms         385ms         397ms
remote    default     linkerd-gateway         2    True         250ms         295ms         299ms
```


controller/api/public/prometheus.go
```diff
@@ -176,7 +181,7 @@ func (s *grpcServer) getPrometheusMetrics(ctx context.Context, requestQueryTempl
 	// kick off asynchronous queries: request count queries + 3 latency queries
 	for pt, requestQueryTemplate := range requestQueryTemplates {
 		var query string
-		if pt == promTCPConnections {
+		if pt == promTCPConnections || pt == promGatewayAlive || pt == promNumMirroredServices {
```
adleong, line 184: I think this was a bit of a hack to begin with and is getting worse here.  What if we added a boolean `timeWindow` parameter to the signature of `getPrometheusMetrics` that indicates if the query expects a time window?

cli/cmd/cluster.go
```diff
@@ -285,14 +326,169 @@ available across clusters.`,
...
+)
+
+func writeGatewayRowsToBuffer(rows []*pb.GatewaysTable_Row, w *tabwriter.Writer) {
+	maxNameLength := len(gatewayNameHeader)
+	maxNamespaceLength := len(gatewayNamespaceHeader)
+	maxClusterNameLength := len(clusterNameHeader)
+
+	rowsMap := make(map[string]*pb.GatewaysTable_Row)
+
+	for _, r := range rows {
```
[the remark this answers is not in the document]
kleimkuhler: Ah cool!

controller/api/public/prometheus.go
zaharidichev: I wonder, can we jsut check whether the time window is empty or not? We should always have a default time window if the query needs one, correct ?

cli/cmd/cluster.go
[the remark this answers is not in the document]
zaharidichev: Wow @adleong that is some real engineering :)

controller/api/public/prometheus.go
adleong: makes sense to me

PR #3700, opened 2019-11-12
Add Responses to path items to satisfy kube apiserver

zaharidichev:
Adds a `Responses` field to the `/openapi/v2` endpoints to satisfy kubernetes and avoid errors being logged in kube-apiserver.

Fixes: #3361

Signed-off-by: zaharidichev <zaharidichev@gmail.com>

Issue #4209, opened 2020-03-27
APIService swagger specs


grampelberg:
### What is the issue?

It appears that some k8s clients assume APIServices expose swagger specs. If they don't get the spec, crashes happen. See:

- https://github.com/godaddy/kubernetes-external-secrets/issues/278
- #4116 

### How can it be reproduced?

- Install Linkerd and then add external secrets.
- Start GKE cluster with calico.

### Possible solution

Have the `tap` and `smi-metrics` APIServices return swagger profiles.

siggy:
It might be as simple as adding something like:
```go
router.GET("/swagger.json", handleSwagger)
```
to:
https://github.com/linkerd/linkerd2/blob/3088f404cec1c33349acfc07e6dc8aad8e0d179f/controller/tap/handlers.go#L78-L87

dad264:
This issue has caused trouble for us as we are users of [KES](https://github.com/godaddy/kubernetes-external-secrets/). Luckily, it's just in our QA environment right now. Would your recommendation be to roll back to https://github.com/linkerd/linkerd2/releases/tag/stable-2.6.1 so we can get back up and running?

grampelberg:
@dad264 the work around is to just install Linkerd after KES.

dad264:
If KES restarts, I don't believe it will start up again.

dad264:
Worked around it [like this](https://github.com/godaddy/kubernetes-external-secrets/issues/278#issuecomment-607440992)

kleimkuhler:
So I do not think this an issue about the routes that the tap API service exposes.

Between 2.6.1 and 2.7, the Linkerd tap endpoints started showing up in the response from `openapi/v2`:

```
curl localhost:port/openapi/v2 | jq .
..
    "/apis/tap.linkerd.io": {
      "get": {
            ..
        }
      }
    },
    "/apis/tap.linkerd.io/v1alpha1": {
      "get": {
            ..
        }
      }
    },
..
```

These returned endpoints do not have an `operationID` field which is not required.

This is an issue because as one of comments mentions in godaddy/kubernetes-external-secrets#278, a dependency *expects* an `operationID` to be present ([code](https://github.com/silasbw/swagger-fluent/blob/master/lib/loader.js#L223)).

To fix this, Linkerd can add an `operationID` field to tap's `/openapi/v2` response. I've confirmed this works in d314212e980c029dd577d9f43e376a33539d3ce5.

[...]

PR #4245, opened 2020-04-11
Add `operationID` field to tap openapi response

kleimkuhler:
This fixes an issue users are experiencing when upgrading from from Linkerd
2.6 to 2.7 and use the [kubernetes-external-secrets]() project.

The change introduced by #3700 resulted in the tap service showing up in the
`/openapi/v2` API response. I confirmed this with a local build.

A dependency within the project expects the `operationID` field to be present
in the swagger definition. It is optional as stated in the
[spec](https://swagger.io/docs/specification/paths-and-operations/). It's
purpose is to identify an operation and should be unique.

This change adds that field to tap service swagger spec. While this can be
fixed in the KES dependency, it certainly does not hurt to add and other
libraries may similarly expect this field.

### DIFF
--- a/controller/tap/handlers.go
+++ b/controller/tap/handlers.go
@@ -330,6 +330,7 @@ func mkPathItem(desc string) spec.PathItem {
 							},
 						},
 					},
+					ID: "tapResourceV0",
 				},
 			},
 		},


kleimkuhler:
#4209 asks for a different solution, but the issue that is being discussed on it is closed by this change. I don't know if we still need to worry about adding a `/swaggerapi` route to the tap service.

controller/tap/handlers.go
```diff
@@ -330,6 +330,7 @@ func mkPathItem(desc string) spec.PathItem {
 							},
 						},
 					},
+					ID: "tapResourceV0",
```
cpretzer, line 333: Nice fix @kleimkuhler 

Small nit: I'd put this above `Responses`, just to make it a bit more readable, but TIOLI 👍

cpretzer: forgot to mention that this works for me locally :)

mycrEEpy:
Is there a chance to cherry pick this into a 2.7.2 release?

cpretzer:
@mycrEEpy now that the is merged it will be in all edge and stable releases going forward, so it will be in 2.7.2

grampelberg:
@mycrEEpy we're not planning on a 2.7.2 right now. Doing stable releases takes a ton of effort and we've reserved those for critical proxy bugs so far (avoiding backporting any control plane changes). Is there something blocking you from picking up tomorrow's edge? 2.8 will be out in ~4 weeks as well (don't quote me on that).

PR #4247, opened 2020-04-13
Refactor AddOn Installation

Pothulapati:
Refactor AddOn Installation Code

As we add more add-ons like Grafana, Prometheus, etc. The add-ons code could become redundant and having all the add-on related stuff in `install.go` doesn't seem to be right.

Changes Include 
- A `AddOn` Interface, which should be implemented by add-ons and is used by `install.go` to remove all the add-on specific code.
- Add Unit tests for the same.

### DIFF
--- a/cli/cmd/install_helm_test.go
+++ b/cli/cmd/install_helm_test.go
@@ -153,7 +153,6 @@ func testRenderHelm(t *testing.T, chart *pb.Chart, addOnConfig string, goldenFil
 		}
 	}
 
-	fmt.Println(buf.String())
 	diffTestdata(t, goldenFileName, buf.String())
 }
 
--- a/cli/cmd/install_helm_test.go
+++ b/cli/cmd/install_helm_test.go
@@ -23,18 +23,28 @@ func TestRenderHelm(t *testing.T) {
 
 	t.Run("Non-HA mode", func(t *testing.T) {
 		ha := false
-		chartControlPlane := chartControlPlane(t, ha, "111", "222")
-		testRenderHelm(t, chartControlPlane, "install_helm_output.golden")
+		chartControlPlane := chartControlPlane(t, ha, nil, "111", "222")
+		testRenderHelm(t, chartControlPlane, "", "install_helm_output.golden")
 	})
 
 	t.Run("HA mode", func(t *testing.T) {
 		ha := true
-		chartControlPlane := chartControlPlane(t, ha, "111", "222")
-		testRenderHelm(t, chartControlPlane, "install_helm_output_ha.golden")
+		chartControlPlane := chartControlPlane(t, ha, nil, "111", "222")
+		testRenderHelm(t, chartControlPlane, "", "install_helm_output_ha.golden")
+	})
+
+	t.Run("Non-HA with add-ons mode", func(t *testing.T) {
+		ha := false
+		addOnConfig := `
+tracing:
+  enabled: true
+`
+		chartControlPlane := chartControlPlane(t, ha, []l5dcharts.AddOn{l5dcharts.Tracing{}}, "111", "222")
+		testRenderHelm(t, chartControlPlane, addOnConfig, "install_helm_output_addons.golden")
 	})
 }
 
-func testRenderHelm(t *testing.T, chart *pb.Chart, goldenFileName string) {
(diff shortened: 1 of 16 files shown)


cli/cmd/install_helm_test.go
```diff
@@ -121,10 +140,24 @@ func testRenderHelm(t *testing.T, chart *pb.Chart, goldenFileName string) {
...
+			if !exists {
+				// skip partial templates
+				continue
+			}
+			buf.WriteString("---\n# Source: " + source + "\n")
+			buf.WriteString(v)
+		}
+	}
+
+	fmt.Println(buf.String())
```
alpeb, line 156: :eyes:

pkg/charts/linkerd2/addons.go
```diff
@@ -0,0 +1,30 @@
...
+	if values.Tracing != nil {
+		if enabled, ok := values.Tracing["enabled"].(bool); !ok {
+			return nil, fmt.Errorf("invalid value for 'Tracing.enabled' (should be boolean): %s", values.Tracing["enabled"])
+		} else if enabled {
+			addOns = append(addOns, values.Tracing)
+		}
+	}
+
+	return addOns, nil
+}
```
alpeb, line 30: I agree it's fine to keep the explicit checks here instead of resorting to reflexion :+1:

pkg/charts/linkerd2/addons_test.go
```diff
@@ -0,0 +1,31 @@
+package linkerd2
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
```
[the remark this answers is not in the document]
Pothulapati: yep, makes sense

pkg/charts/linkerd2/tracing.go
```diff
@@ -0,0 +1,38 @@
+package linkerd2
+
+import (
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
Pothulapati: It's my Goland IDE adding deps automatically, not using the existing ones :p

alpeb:
Since dependencies are no longer changed, do you mind running `bin/update-go-deps-shas` one more time?

pkg/charts/linkerd2/tracing.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"k8s.io/helm/pkg/chartutil"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	tracingAddOn = "tracing"
+)
+
+// Tracing is a add-on that installs the distributed tracing
```
kleimkuhler, line 12: nit:
```suggestion
// Tracing is an add-on that installs the distributed tracing
```

pkg/charts/linkerd2/tracing.go
```diff
@@ -0,0 +1,37 @@
...
+// Values returns the configuration values that were assigned for this add-on
+func (t Tracing) Values() []byte {
+	values, err := yaml.Marshal(t)
+	if err != nil {
+		return nil
+	}
+	return values
+}
+
+//Templates returns the template files specific to this add-on
```
kleimkuhler, line 30: nit:
```suggestion
// Templates returns the template files specific to this add-on
```

pkg/k8s/api.go
```diff
@@ -221,3 +221,13 @@ func GetPodStatus(pod corev1.Pod) string {
 
 	return reason
 }
+
+// GetAddOnsConfigMap returns the data in a configmap
+func GetAddOnsConfigMap(kubeAPI kubernetes.Interface, namespace string) (map[string]string, error) {
```
kleimkuhler, line 226: I don't see anything in this function that is specific to addons. Would this be better named `GetConfigMap`?

Pothulapati: Yeah, but we ain't passing the name of the add-on cm, right? So, It's not a general function? :thinking:

kleimkuhler: So it's not the fact that we don't pass the name of the add-on configmap, it's that this function isn't specific to configmaps for *only* add-ons. If there were more configmaps in the namespace, would those be returned?

kleimkuhler: I see that `ValuesConfigMapName` is Linkerd add-on specific. Nevermind my confusion about the naming on this!

PR #4265, opened 2020-04-16
Multicluster gateway and remote setup command

zaharidichev:
This PR introduces two notable things: 

- We add a helm chart that contains setup for the remote cluster in a multicluster setting. That includes a helm chart with a reference implementation of a gateway as well as credentials for accessing the cluster remotely (by the service mirror component)
- We substitute the `cluster create-credentials` command with `cluster setup-remote`, which outputs the yaml needed for setting up the remote cluster (gateway + credentials)

### DIFF
--- a/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
+++ b/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
@@ -8,6 +8,8 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: nginx-configuration
+  annotations:
+    {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
   namespace: {{.Values.gatewayNamespace}}
 data:
   nginx.conf: |-
@@ -32,7 +34,12 @@ data:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  annotations:
+    {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
   labels:
+    app.kubernetes.io/name: gateway
+    app.kubernetes.io/part-of: Linkerd
+    app.kubernetes.io/version: {{.Values.linkerdVersion}}
     app: {{.Values.gatewayName}}
   name: {{.Values.gatewayName}}
   namespace: {{.Values.gatewayNamespace}}
@@ -44,6 +51,7 @@ spec:
   template:
     metadata:
       annotations:
+        {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}        
         linkerd.io/inject: enabled
       labels:
         app: {{.Values.gatewayName}}
@@ -54,7 +62,17 @@ spec:
             name: nginx-configuration   
       containers:
         - name: nginx
-          image: nginx:1.17
+          readinessProbe:
+            failureThreshold: 7
+            httpGet:
+              path: {{.Values.probePath}}
+              port: {{.Values.probePort}}
+          livenessProbe:
+            httpGet:
+              path: {{.Values.probePath}}
+              port: {{.Values.probePort}}
+            initialDelaySeconds: 10  
+          image: {{.Values.nginxImage}}:{{.Values.nginxImageVersion}}
           ports:
             - name: incoming-port
               containerPort: {{.Values.incomingPort}}
@@ -75,6 +93,7 @@ metadata:
     mirror.linkerd.io/probe-port: {{.Values.probePort}}
     mirror.linkerd.io/probe-period: {{.Values.probePeriodSeconds}}
     mirror.linkerd.io/probe-path: {{.Values.probePath}}
+    {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
 spec:
   ports:
   - name: incoming-port
--- a/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
+++ b/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
@@ -90,8 +90,8 @@ metadata:
   namespace: {{.Values.gatewayNamespace}}
   annotations: 
     mirror.linkerd.io/gateway-identity: {{.Values.gatewayName}}.{{.Values.gatewayNamespace}}.serviceaccount.identity.{{.Values.linkerdNamespace}}.{{.Values.identityTrustDomain}}
-    mirror.linkerd.io/probe-port: {{.Values.probePort}}
-    mirror.linkerd.io/probe-period: {{.Values.probePeriodSeconds}}
+    mirror.linkerd.io/probe-port: "{{.Values.probePort}}"
+    mirror.linkerd.io/probe-period: "{{.Values.probePeriodSeconds}}"
     mirror.linkerd.io/probe-path: {{.Values.probePath}}
     {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
 spec:
--- a/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
+++ b/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
@@ -1,0 +1,95 @@
+---
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: {{ .Values.gatewayNamespace }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nginx-configuration
+  namespace: {{.Values.gatewayNamespace}}
+data:
+  nginx.conf: |-
+    events {
+    }
+    stream {
+      server {
+          listen     {{.Values.incomingPort}};
+          proxy_pass 127.0.0.1:{{.Values.proxyOutboundPort}};
+      }
+    }
+    http {
+      server {
+          listen     {{.Values.probePort}};
+          location {{.Values.probePath}} {
+            access_log off;
+            return 200 "healthy\n";
+          }
+      }
+    }
+---    
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: {{.Values.gatewayName}}
+  name: {{.Values.gatewayName}}
+  namespace: {{.Values.gatewayNamespace}}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{.Values.gatewayName}}
+  template:
+    metadata:
+      annotations:
+        linkerd.io/inject: enabled
+      labels:
+        app: {{.Values.gatewayName}}
+    spec:
+      volumes:
+        - name: config
+          configMap:
+            name: nginx-configuration   
+      containers:
+        - name: nginx
+          image: nginx:1.17
+          ports:
+            - name: incoming-port
+              containerPort: {{.Values.incomingPort}}
+            - name: probe-port
+              containerPort: {{.Values.probePort}}             
+          volumeMounts:
+            - name: config
+              mountPath: /etc/nginx
+      serviceAccountName: {{.Values.gatewayName}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.Values.gatewayName}}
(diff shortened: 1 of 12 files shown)


charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+metadata:
+  name: nginx-configuration
+  namespace: {{.Values.gatewayNamespace}}
+data:
+  nginx.conf: |-
+    events {
+    }
+    stream {
+      server {
+          listen     {{.Values.incomingPort}};
```
grampelberg, line 18: Do we want any of this to be configurable?

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+          proxy_pass 127.0.0.1:{{.Values.proxyOutboundPort}};
+      }
+    }
+    http {
```
grampelberg, line 22: Good one!

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+  namespace: {{.Values.gatewayNamespace}}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{.Values.gatewayName}}
+  template:
+    metadata:
+      annotations:
+        linkerd.io/inject: enabled
```
grampelberg, line 47: Do we want to use the proxy partial for this instead of relying on the injector?

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+      labels:
+        app: {{.Values.gatewayName}}
+    spec:
+      volumes:
+        - name: config
+          configMap:
+            name: nginx-configuration   
+      containers:
```
grampelberg, line 55: We'll want liveness/readiness checks here.

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+        - name: nginx
+          image: nginx:1.17
```
grampelberg, line 57: Might as well make the image/tag configurable. I'd follow how we're doing it everywhere else but I dunno what the image name and tag need to be separate variables.

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+              mountPath: /etc/nginx
+      serviceAccountName: {{.Values.gatewayName}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.Values.gatewayName}}
+  namespace: {{.Values.gatewayNamespace}}
+  annotations: 
+    mirror.linkerd.io/gateway-identity: {{.Values.gatewayName}}.{{.Values.gatewayNamespace}}.serviceaccount.identity.{{.Values.linkerdNamespace}}.{{.Values.identityTrustDomain}}
```
grampelberg, line 74: This is the identity of the local gateway? Do we have a partial to put this together already? If not we probably should get one.

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
+---
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: {{ .Values.gatewayNamespace }}
```
grampelberg, line 5: I created this all in a separate namespace because it was kinda on the side at the time. Should it all just go into the control plane namespace? I could see downsides either way.

charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
```diff
@@ -0,0 +1,39 @@
+---
```
grampelberg, line 1: Should we have a version of this that is namespace specific? I'm wondering if some folks will want to pass a list of namespaces to be mirrored at install time and have this be a RoleBinding instead.

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,95 @@
...
+            access_log off;
+            return 200 "healthy\n";
+          }
+      }
+    }
+---    
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
```
grampelberg, line 35: We probably could use a partial for this, but can we get all the "standard" labels and annotations in here?

charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: {{ .Values.serviceAccountNamespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{.Values.serviceAccountName}}
```
grampelberg, line 10: I'd like to see some more labels/annotations here so that folks have some documentation around what installed this and what it is used for.

cli/cmd/cluster.go
```diff
@@ -64,7 +109,7 @@ functionality of Linkerd. You can use it to deploy credentials to
 remote clusters, extract them as well as export remote services to be
 available across clusters.`,
 		Example: `  # Create remote cluster credentials.
-  linkerd --context=cluster-a cluster create-credentials | kubectl --context=cluster-a apply -f -
+  linkerd --context=cluster-a cluster setup-remote | kubectl --context=cluster-a apply -f -
```
grampelberg, line 112: I think the previous line needs to be updated now? I'm wondering if this is the right command name, but we can change that later.

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
zaharidichev: So if we want to use the proxy partials I think it will be best to bring that into the linkerd2 chart. Or else, we need to add all the config that is exposed to the proxy, like ports, etc etc. 

My assumption here is that we install that into a cluster that already has Linkerd installed and running in the specified namespace.  @alpeb you have better understanding of Helm charts. What do you think ?

alpeb: You could make this an addon, disabled by default, and put everything under `charts/add-ons/multicluster-remote-setup`. Just like `charts/add-ons/tracing` you can require partials. The global configs in linkerd's main `values.yaml` (which include the proxy configs) will be available in your addon, without having to redeclare them in your addon's `values.yaml`.

adleong: I think a question here is if we want the gateway to be "pre-injected" with the proxy partial like the other control plane components, or if we just want to rely on the proxy injector.  I don't have a good sense for which of these options is best, but naively, using the proxy-injector seems simpler.

grampelberg: I think we've got injection disabled in `linkerd`, so there's some implications on where we want this to actually live and run.

zaharidichev:
@adleong @grampelberg Yes, so to take a step back, my idea here is that I want to keep it as simple as possible, keep it as a separate chart and not introduce a bunch of config parameter or types. The current implementation should work fine given that: 

- there is a Linkerd installation  in the specified namespace
- the automatic injector works

Is there something that I am missing that could be done better?

grampelberg:
@zaharidichev I don't have any issues merging this now as long as we get tickets to do some follow up. 

We've got injection disabled on `linkerd` (just double checked). So we'll either need to add a test to make sure they *can't* install it in that namespace or come up with another pattern.

zaharidichev:
@grampelberg I agree with most of the feedback and will address it ofcourse. I just wanted to make sure we are all fine with relying on the auto injector to avoid using the partials. (@alpeb you think that is a good approach)? With respect to upholding the namespace contraints, I intent to introduce a check that disallow all this to be installed in the linkerd namespace (just to keep things seggregated). If that makes sense, we can proceed this way :)

grampelberg:
AFAIK that's a good tradeoff for now. Keeping it out of the `linkerd` namespace has some benefits.

charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml
```diff
@@ -0,0 +1,114 @@
...
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.Values.gatewayName}}
+  namespace: {{.Values.gatewayNamespace}}
+  annotations: 
+    mirror.linkerd.io/gateway-identity: {{.Values.gatewayName}}.{{.Values.gatewayNamespace}}.serviceaccount.identity.{{.Values.linkerdNamespace}}.{{.Values.identityTrustDomain}}
+    mirror.linkerd.io/probe-port: {{.Values.probePort}}
+    mirror.linkerd.io/probe-period: {{.Values.probePeriodSeconds}}
```
alpeb, line 94: It appears k8s is not happy with having numbers as annotations values (at least in v1.15). I get the following error unless I wrap them in double quotes:
```
error: unable to decode "remote-cluster.yml": resource.metadataOnlyObject.ObjectMeta: v1.ObjectMeta.Annotations: ReadString: expects " or n, but found 8, error found in #10 byte of ...|be-port":81},"name":|..., bigger context ...|probe-period":"3","mirror.linkerd.io/probe-port":81},"name":"linkerd-gateway","namespace":"linkerd-g|...
```

alpeb:
@zaharidichev It might be worth setting the default value for `gateway-namespace` for `linkerd cluster export-service` to match the new default for the gateway namespace in this same PR, to facilitate testing :wink:

cli/cmd/cluster.go
```diff
@@ -56,96 +69,166 @@ type (
...
+			return nil
+		},
+	}
 
-			svcAccount := corev1.ServiceAccount{
-				ObjectMeta: metav1.ObjectMeta{Name: createOpts.serviceAccount, Namespace: createOpts.namespace, Labels: labels},
-				TypeMeta:   metav1.TypeMeta{Kind: v1.ServiceAccountKind, APIVersion: "v1"},
-			}
+	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "", "the name of the remote cluster")
+	cmd.Flags().StringVar(&opts.gatewayNamespace, "gateway-namespace", "", "the namespace in which the gateway resides on the remote cluster")
```
alpeb, line 158: Do you mind also setting the defaults here to match the other ones?

zaharidichev:
@alpeb I will be submitting a follow up pr to transition the export service command to output a yaml tomorrow most likely

alpeb:
@zaharidichev :cool: 
OTOH I'm still having trouble with `linkerd cluster gateways` returning a 404, when run against the local cluster. Any ideas what could be going on?

zaharidichev:
Hmm.. no. Not really. I will test it out tomorrow. Any control plane logs from the public-api pod  of the loca cluster ?

cli/cmd/cluster.go
zaharidichev: Addressed in #4307

zaharidichev:
@alpeb I noticed that this happened to me today but thre was a reason. My default kube context is set to the remote cluster (on which I am running the latest edge). So when I tried to execute the gateways comamnd it hit its api that is missing this endpoint and that is why it returned a 404. Are you sure you are not in the same situation. For example: 

```
[13:55] zahari.local:linkerd2 (zd/multicluster-gateway) | bin/linkerd --context=local cluster gateways
CLUSTER  NAMESPACE        NAME             ALIVE    NUM_SVC  LATENCY_P50  LATENCY_P95  LATENCY_P99
remote   linkerd-gateway  linkerd-gateway  True           1        275ms        385ms        397ms
[13:56] zahari.local:linkerd2 (zd/multicluster-gateway) | bin/linkerd --context=remote cluster gateways
Error: Gateways API error: HTTP error, status Code [404] (unexpected API response: 404 page not found
)
```

alpeb:
Thanks @zaharidichev after making sure I got all the pieces set up as intended, it's working impecable, thanks :+1:

Issue #3939, opened 2020-01-16
ha: anti-affnity rules are too restrictive


olix0r:
Linkerd, installed with `--ha` on a 3-node cluster, cannot be successfully upgraded:

```
:; kubectl get events -n linkerd
LAST SEEN   TYPE      REASON             OBJECT                                        MESSAGE
94s         Warning   FailedScheduling   pod/linkerd-controller-c845b4549-wbz2d        0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
94s         Warning   FailedScheduling   pod/linkerd-destination-6488b657b-5srp8       0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
94s         Warning   FailedScheduling   pod/linkerd-identity-7dc5d97cf9-rxnz2         0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
94s         Warning   FailedScheduling   pod/linkerd-proxy-injector-64d8d46794-lb69h   0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
94s         Warning   FailedScheduling   pod/linkerd-sp-validator-86bbcbfbff-sldx6     0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
94s         Warning   FailedScheduling   pod/linkerd-tap-c895b586d-ps22z               0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
```

@grampelberg suggests that this can be resolved with the following:

- define a PDB - https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget
- set progressDeadlineSeconds
- define maxUnavailable
- configure maxSurge

kleimkuhler:
@grampelberg I've been looking into this this morning and I'm not sure a PDB is the answer here. I think we either need to relax the anti-affinity rules from require to prefer so that pods can be scheduled on a node that has other control plane pods, or recommend HA deployments on at least 4 node clusters.

[...]

PR #4267, opened 2020-04-16
Add rolling update strategy to for HA deployments

kleimkuhler:
This introduces a rolling update strategy to Linkerd deployments that have
three replicas during HA deployments. This allows for at most one pod to begin
terminating before a new pod ready is ready.

This allows for upgrades to take place on three node clusters. As a pod begins
terminating, it opens up the node for the new pod to start initializing.

`.Values.enablePodAntiAffinity` was chosen as the conditional here because it
is set by the `values-ha.yaml` config on HA deployments

### DIFF
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -45,6 +45,11 @@ spec:
       {{.Values.global.controllerComponentLabel}}: controller
       {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
       {{- include "partials.proxy.labels" .Values.global.proxy | nindent 6}}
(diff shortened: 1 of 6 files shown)


kleimkuhler:
Relates to #3939

Pothulapati:
hmmm, Looks like some `ha` test golden files weren't updated in this PR, and hence those changes came up in https://github.com/linkerd/linkerd2/pull/4271/files#diff-4b8b88929c550f82a1a2ac3057981d9d as I updated the test files.

Definitely not a problem, but wondering how GH checks missed them. :thinking:

alpeb:
@Pothulapati Turns out the test is failing but it's not surfacing as a suite failure by `go test`. I'll open a PR to address that. Thanks for the find!

PR #4270, opened 2020-04-16
Use /live for liveness probe

### LINKED ISSUE #3984: Stop using the proxy's /metrics endpoint as a livenessProbe
## Background

Linkerd proxies are configured with a `livenessProbe` and a `readinessProbe`. For the `livenessProbe` we are using the proxy's `/metrics` endpoint. This endpoint is unnecessarily expensive, as we just require an endpoint to return a 200, similar to the `readinessProbe`'s `/ready` endpoint.

## Proposal

Modify the proxy's `livenessProbe` to point to a different endpoint, either a new `/liveness` endpoint, or re-use the `/ready` endpoint.

### liveness/readiness k8s config for proxy

```yaml
livenessProbe:
  httpGet:
    path: /metrics
    port: 4191
  initialDelaySeconds: 10
readinessProbe:
  httpGet:
    path: /ready
    port: 4191
  initialDelaySeconds: 2
```

### `/metrics` payload size

A typical control-plane component's proxy emits 60kb of data on the `/metrics` endpoint:
```bash
$ linkerd metrics -n linkerd $(
    kubectl --namespace linkerd get pod \
      --selector linkerd.io/control-plane-component=controller \
      --output name \
    | head -n1
  ) | wc -c
61828
```

### Kubernetes probing code

Kubernetes reads up to 10kb when probing:
https://github.com/kubernetes/kubernetes/blob/866ea74326fba45241cbccd3ca4624b68ba7022f/pkg/probe/http/http.go#L36

...and then logs a non-fatal error:
https://github.com/kubernetes/kubernetes/blob/866ea74326fba45241cbccd3ca4624b68ba7022f/pkg/probe/http/http.go#L114-L118

adleong:
Fixes #3984

We use the new `/live` admin endpoint in the Linkerd proxy for liveness probes instead of the `/metrics` endpoint.  This endpoint returns a much smaller payload.

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -85,7 +85,7 @@ image: {{.Values.global.proxy.image.name}}:{{.Values.global.proxy.image.version}
 imagePullPolicy: {{.Values.global.proxy.image.pullPolicy}}
 livenessProbe:
   httpGet:
-    path: /metrics
+    path: /live
     port: {{.Values.global.proxy.ports.admin}}
   initialDelaySeconds: 10
 name: linkerd-proxy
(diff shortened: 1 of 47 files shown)


kleimkuhler:
It looks like some additional golden files needed to be updated. They are covered #4273

PR #4272, opened 2020-04-17
Some `linkerd stat` test failures were being hidden

alpeb:
`linkerd stat` was doing an early `os.Exit(0)` when no traffic was
found, which avoided `go test` to report any test failure that ended in
that code path.

This was hiding a mismatch in the golden files for HA after the
introduction of the rolling update strategy (#4267), and the failure of
`linkerd stat trafficsplit` not returning results unless `--unmeshed` is
used. For the latter, I added the flag to the tests in order to temporarly pass
them, but the underlying issue remains to be fixed in a separate
PR.

### DIFF
--- a/cli/cmd/stat_test.go
+++ b/cli/cmd/stat_test.go
@@ -16,11 +16,6 @@ type paramsExp struct {
 
 func TestStat(t *testing.T) {
 	options := newStatOptions()
-
-	// TODO: currently `linkerd stat ts` requires the --unmeshed flag
-	tsOptions := options
-	tsOptions.unmeshed = true
-
 	t.Run("Returns namespace stats", func(t *testing.T) {
 		testStatCall(paramsExp{
 			counts: &public.PodCounts{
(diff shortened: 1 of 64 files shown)


alpeb:
@Pothulapati this also updates the golden files you updated in #4271. You can keep them there; the automatic merge shouldn't return conflicts.

cli/cmd/stat_test.go
```diff
@@ -16,6 +16,11 @@ type paramsExp struct {
 
 func TestStat(t *testing.T) {
 	options := newStatOptions()
+
+	// TODO: currently `linkerd stat ts` requires the --unmeshed flag
```
olix0r, line 20: Is this tracked somewhere?

alpeb: No, but @kleimkuhler will likely push a quick fix soon. Also I think I'll wait for that fix to merge before continuing with this PR because it appears there are other tests failing.

kleimkuhler:
@zaharidichev Just a heads up this depends on #4273 which needs one more review

PR #4273, opened 2020-04-17
Add MeshedPodCount field to TS resource rows

kleimkuhler:
The addition of the `--unmeshed` flag changed the rendering behavior of the
`stat` command so that resources with 0 meshed pods are not displayed by
default.

Rendering is based off the row's `MeshedPodCount` field which is currently not
set by `func trafficSplitResourceQuery`. This change sets that field now so
that in rendering, the trafficsplit resource is rendered in the output.

The reason for this not showing up in testing is addressed by #4272 where the
`stat` command behavior for no traffic is changed.

The following now works without `--unmeshed` flag being passed:

```
❯ bin/linkerd stat -A ts
NAMESPACE   NAME                    APEX          LEAF          WEIGHT   SUCCESS   RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
default     backend-traffic-split   backend-svc   backend-svc     500m         -     -             -             -             -
default     backend-traffic-split   backend-svc   failing-svc        0         -     -             -             -             -
```

### DIFF
--- a/controller/api/public/stat_summary.go
+++ b/controller/api/public/stat_summary.go
@@ -337,9 +337,10 @@ func (s *grpcServer) trafficSplitResourceQuery(ctx context.Context, req *pb.Stat
 					Namespace: tsStats.namespace,
 					Type:      req.GetSelector().GetResource().GetType(),
 				},
-				TimeWindow: req.TimeWindow,
-				Stats:      tsBasicStats[currentLeaf],
-				TsStats:    trafficSplitStats,
+				TimeWindow:     req.TimeWindow,
+				MeshedPodCount: 1,
+				Stats:          tsBasicStats[currentLeaf],
+				TsStats:        trafficSplitStats,
 			}
 			rows = append(rows, &row)
 		}
--- a/controller/api/public/stat_summary.go
+++ b/controller/api/public/stat_summary.go
@@ -337,10 +337,9 @@ func (s *grpcServer) trafficSplitResourceQuery(ctx context.Context, req *pb.Stat
 					Namespace: tsStats.namespace,
 					Type:      req.GetSelector().GetResource().GetType(),
 				},
-				TimeWindow:     req.TimeWindow,
-				MeshedPodCount: 1,
-				Stats:          tsBasicStats[currentLeaf],
-				TsStats:        trafficSplitStats,
+				TimeWindow: req.TimeWindow,
+				Stats:      tsBasicStats[currentLeaf],
+				TsStats:    trafficSplitStats,
 			}
 			rows = append(rows, &row)
 		}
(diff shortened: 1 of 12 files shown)


controller/api/public/stat_summary.go
```diff
@@ -337,9 +337,10 @@ func (s *grpcServer) trafficSplitResourceQuery(ctx context.Context, req *pb.Stat
 					Namespace: tsStats.namespace,
 					Type:      req.GetSelector().GetResource().GetType(),
 				},
-				TimeWindow: req.TimeWindow,
-				Stats:      tsBasicStats[currentLeaf],
-				TsStats:    trafficSplitStats,
+				TimeWindow:     req.TimeWindow,
+				MeshedPodCount: 1,
+				Stats:          tsBasicStats[currentLeaf],
+				TsStats:        trafficSplitStats,
```
kleimkuhler, line 343: (1/2) This is the significant part of this PR. The other changes are test files updates that have gotten through their initial merges

controller/api/public/test_helper.go
```diff
@@ -344,7 +344,8 @@ func GenStatTsResponse(resName, resType string, resNs []string, basicStats bool,
 					Type:      resType,
 					Name:      resName,
 				},
-				TimeWindow: "1m",
+				TimeWindow:     "1m",
+				MeshedPodCount: 1,
```
kleimkuhler, line 348: (2/2) This is the significant part of this PR. The other changes are test files updates that have gotten through their initial merges

kleimkuhler:
The other solution for this is that `linkerd stat ts` command just turns on the unmeshed flag by default.

These resources are unmeshed, so it's another reasonable solution. I chose to do use `MeshedPodCount` though because once the unmeshed flag is turned on, it would be on for everything in the `linkerd stat` command

controller/api/public/stat_summary.go
adleong: The fact that we have to set MeshedPodCount to a non-zero value in order for Traffic split rows to be rendered is super non-obvious.  At the very least this should explained in a comment here.  But I think an even clearer solution would be in to the stat command if the resource is a traffic split then we should just set the `unmeshed` option to true.  Omitting unmeshed resources is not a thing that makes sense for traffic splits.

kleimkuhler: @adleong Thanks yea I was thinking a little more about this this weekend and I agree with this as I mentioned in my [comment](https://github.com/linkerd/linkerd2/pull/4273#issuecomment-615517376). This change as it is is not very obvious and I think it would be much simpler to just set `unmeshed` to true if this is a trafficsplit resource.

alpeb: ... or more explicitly, leaving options alone, replacing
```go
if !options.unmeshed && r.GetMeshedPodCount() == 0 {
  continue
}
```
with
```go
if r.Resource.Type == k8s.TrafficSplit || (!options.unmeshed && r.GetMeshedPodCount() == 0) {
  continue
}
```

kleimkuhler: @adleong @alpeb I've add the change

PR #4279, opened 2020-04-21
Multicluster checks

zaharidichev:
Adds multicluster checks that: 

- check that a service mirror controller exists (1 only)
- check that the expected ClusterRoles, ClusterRoleBindings and ServiceAccounts for the controller exist locally
- for each secret of type `remote-kubeconfig` we check that we can parse the contents into a k8s RestConfig object, can instantiate API connection to the remote cluster AND can LIST, WATCH and GET services in all namespace on this cluster. This is done for all clusters and if there is anything wrong a list of clusters violating these constraints is printed
- checks that each of the gateways (for which we have metrics in Prometheus) is alive and if not it prints a list of problematic cluster gateways

### DIFF
--- a/charts/linkerd2-service-mirror/templates/service-mirror.yaml
+++ b/charts/linkerd2-service-mirror/templates/service-mirror.yaml
@@ -32,7 +32,7 @@ metadata:
   name: linkerd-service-mirror
   namespace: {{.Values.namespace}}
   labels:
-    {{.Values.controllerComponentLabel}}: linkerd-service-mirrors
+    {{.Values.controllerComponentLabel}}: linkerd-service-mirror
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/charts/linkerd2-service-mirror/templates/service-mirror.yaml
+++ b/charts/linkerd2-service-mirror/templates/service-mirror.yaml
@@ -32,7 +32,7 @@ metadata:
   name: linkerd-service-mirror
   namespace: {{.Values.namespace}}
   labels:
-    {{.Values.controllerComponentLabel}}: linkerd-service-mirror
+    {{.Values.controllerComponentLabel}}: linkerd-service-mirrors
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -21,6 +21,7 @@ import (
 type checkOptions struct {
 	versionOverride    string
 	preInstallOnly     bool
+	multicluster       bool
 	dataPlaneOnly      bool
 	wait               time.Duration
 	namespace          string
@@ -31,6 +32,7 @@ type checkOptions struct {
 
 func newCheckOptions() *checkOptions {
 	return &checkOptions{
+		multicluster:       false,
 		versionOverride:    "",
 		preInstallOnly:     false,
 		dataPlaneOnly:      false,
@@ -50,6 +52,7 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	flags.BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
+	flags.BoolVar(&options.multicluster, "multicluster", options.multicluster, "Run multicluster checks")
 
 	return flags
 }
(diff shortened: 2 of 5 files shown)


charts/linkerd2-service-mirror/templates/service-mirror.yaml
```diff
@@ -32,7 +32,7 @@ metadata:
   name: linkerd-service-mirror
   namespace: {{.Values.namespace}}
   labels:
-    {{.Values.controllerComponentLabel}}: linkerd-service-mirror
+    {{.Values.controllerComponentLabel}}: linkerd-service-mirrors
```
adleong, line 35: is this change intentional?

cli/cmd/check.go
```diff
@@ -50,6 +52,7 @@ func (options *checkOptions) nonConfigFlagSet() *pflag.FlagSet {
 	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
 	flags.BoolVar(&options.preInstallOnly, "pre", options.preInstallOnly, "Only run pre-installation checks, to determine if the control plane can be installed")
 	flags.BoolVar(&options.dataPlaneOnly, "proxy", options.dataPlaneOnly, "Only run data-plane checks, to determine if the data plane is healthy")
+	flags.BoolVar(&options.multicluster, "multicluster", options.multicluster, "Run multicluster checks")
```
adleong, line 55: just thinking out loud here (figuratively):

would it make sense to automatically enable these checks if the service mirror component has been installed?  I think it's still useful to have the command so that if the service mirror component failed to be installed for some reason, a user running `linkerd check --multicluster` would be informed of that.  But I think it might also be useful to automatically enable this flag so that users who have successfully installed the service mirror will still get informed of errors with their multicluster set up, even if they don't know (or remember) to add the `--multicluster` flag.

controller/cmd/service-mirror/config_watcher.go
```diff
@@ -112,74 +113,49 @@ func (rcw *RemoteClusterConfigWatcher) Stop() {
...
 		return fmt.Errorf("unable to parse kube config: %s", err)
 	}
 
 	rcw.Lock()
 	defer rcw.Unlock()
 
-	if _, ok := rcw.clusterWatchers[config.clusterName]; ok {
-		return fmt.Errorf("there is already a cluster with name %s being watcher. Please delete its config before attempting to register a new one", config.clusterName)
+	if _, ok := rcw.clusterWatchers[config.ClusterName]; ok {
+		return fmt.Errorf("there is already a cluster with name %s being watcher. Please delete its config before attempting to register a new one", config.ClusterName)
```
adleong, line 131: typo: being watcher

pkg/healthcheck/healthcheck.go
```diff
@@ -1463,6 +1546,169 @@ func FetchLinkerdConfigMap(k kubernetes.Interface, controlPlaneNamespace string)
...
+		return fmt.Errorf("Service mirror ClusterRole is missing expected policy rule")
+	}
+
+	if err := compare(expectedServiceMirrorPolicyVerbs, policyRule.Verbs); err != nil {
+		return fmt.Errorf("Service mirror ClusterRole is missing verbs: %s", err)
+	}
+
+	if err := compare(expectedServiceMirrorPolicyResources, policyRule.Resources); err != nil {
+		return fmt.Errorf("Service mirror ClusterRole is missing required resources: %s", err)
+	}
```
alpeb, line 1618: Can you avoid short-circuiting on the first error here, and instead show both errors if present?

pkg/healthcheck/healthcheck.go
```diff
@@ -1463,6 +1546,169 @@ func FetchLinkerdConfigMap(k kubernetes.Interface, controlPlaneNamespace string)
...
+		if err := hc.checkCanPerformAction(remoteAPI, "list", corev1.NamespaceAll, "", "v1", "services"); err == nil {
+			canList = true
+		}
+
+		if err := hc.checkCanPerformAction(remoteAPI, "watch", corev1.NamespaceAll, "", "v1", "services"); err == nil {
+			canWatch = true
+		}
+
+		if !canGet || !canList || !canWatch {
+			errors = append(errors, fmt.Sprintf("* cluster: [%s]: Insufficient permissions GET: %v, LIST: %v, WATCH: %v", config.ClusterName, canGet, canList, canWatch))
```
alpeb, line 1673: Nit: could you unify the error format with the one from `checkServiceMirrorLocalRBAC()`? (there you have something like `Service mirror ClusterRole is missing verbs: expected create,delete,get,list,update,watch, got create,delete,get,list,watch`

cli/cmd/check.go
zaharidichev: Yes.. that actually makes a lot of sense. Will include that modification

zaharidichev: @adleong I pushed a commit that addresses that, can you take a look.

PR #4289, opened 2020-04-24
Extend Helm chart to allow disabling secret resources for self-signed certs. 

### LINKED ISSUE #4288: In Helm chart allow disabeling the creation of certificate secrets for webhook and API server components
Currently the following secrets are created by the helm chart:
* `linkerd-proxy-injector-tls`
* `linkerd-sp-validator-tls`
* `linkerd-smi-metrics-tls`
* `linkerd-tap-tls`
These contain the certificates and private keys for us in TLS connections when configuring the respective webhooks and Kubernetes API server extensions.

The creation of these secrets makes it impossible to inject them into a cluster using tooling external to the linker helm chart.

Make the creation of the secrets by the helm chart configurable (`externalSecret: true|false`). This can be done in a way that does not change the default behaviour when deploying the chart.

**Warning**: For those users that explicitly set the private key and certificate of these TLS configs, they have to configure an additional fileld `caBundle`. This is a breaking change with regards to the Helm configuration.

A potentially beneficial side-effect is the ability to use more integrated certificate hierarchies by allowing the separate setting of a CA bundle instead of requiring the use of a self-signed certificate by making CA bundle and cert the same.

Either a direct integration with one example of this external tooling like cert-manager or full support for certificate rotation as part of linker, as described in #4232 

By setting one more of the `*.externalSecret*' configuration fields to `true` and creating the required secrets through some other means.
In that case the field `*.caBundle` must also be set.

cypherfox:
## Problem ##
The linkerd2 helm chart currently generates secrets for the certificates for the webhooks for the proxy-injector and service profile validator as well as the K8s API server extension smi-metrics and tap respectively.

## Solution ##
This change adds four configuration options (`*.externalSecret`) to the helm chart to allow disabling the creation of the following secrets:
* `linkerd-proxy-injector-tls`
* `linkerd-sp-validator-tls`
* `linkerd-smi-metrics-tls`
* `linkerd-tap-tls`

It also splits the explicit configuration between the certificate (`certPEM`) and the CA bundle (`caBundle`). For those users that currently explicitly set the certificate and private keys, **this is a breaking change**, as they now also have to configure the CA bundle.

For current users that do not configure this at all, nothing changes.

## Validation ##
The MR is currently lacking a proper test. I am working on that.

Fixes #4288

Signed-off-by: Lutz Behnke <lutz.behnke@finleap.com>


cypherfox:
Sorry, I just noted I made a design mistake with the variable scope to ensure maximum possible backward compatible. I have marked it as WIP for now.
I need to find out how to get the test to run locally. I will probably get back after the weekend.

cypherfox:
@alpeb, could you please give me a hint, I am lost:
 I do not understand why the test is failing: The error message talks about `helm check` failing (and the indicated line of the go test code supports that), but **also** about the rendering of the helm template being wrong. 

Not only is a manual rendering of the concerning template successful (`helm template linkerd2 -x templates/tap-rbac.yaml`. The CI test fails for `tap-rbac.yaml`, but I guess it would do so for the other webhooks/API server extensions as well), but the helm tests run fine on my local machine.

alpeb:
@cypherfox The test errors are likely caused by not having updated `pkg/charts/linkerd2/values.go`. The structs there should mimic what's in `values.yaml`. Their intent is to have the `linkerd install` CLI reuse `values.yaml`. Let me know if you need help updating that file.

cypherfox:
Ok, I have changed the `values.go` and fixed the test data (there was one copy error in ` cli/cmd/install_test.go` input data as well).

But to fix `cli/cmd/upgrade.go` I would need to change the way how the cluster is accessed, in order to retrieve a `APIService` resource.

Currently this uses the `kubernetes.Interface` API, which does not expose the APIService resource. I was unable to find a canonical way to access these resources in the linkerd code.

Any hints?

cypherfox:
Ok, I have add a retrieval of caBundle values and made that run. But now I have a bigger problem: **How should linkerd handle missing webhook / APIService configuration when upgrading?** 

Simply use the `crt.pem` from the secret? This will likely make things work, but might not be what the user wanted.

Currently the unit tests in `cli/cmd/upgrade_test.go` are failing, since they only define the secret, but not the appropriate webhook configus.

cypherfox:
@alpeb coud you please have a look and give me a hint on how to proceed (merge / rebase / squash) to make everything ready for merging?

cypherfox:
@alpeb could you please have another look? I finally got the tests to run.

cli/cmd/main_test.go
```diff
@@ -16,13 +16,17 @@ var (
...
+	// write any rejected test data into this path
+	rejectPath string
 )
 
 // TestMain parses flags before running tests
 func TestMain(m *testing.M) {
 	flag.BoolVar(&updateFixtures, "update", false, "update text fixtures in place")
 	prettyDiff = os.Getenv("LINKERD_TEST_PRETTY_DIFF") != ""
 	flag.BoolVar(&prettyDiff, "pretty-diff", prettyDiff, "display the full text when diffing")
+	flag.StringVar(&rejectPath, "reject-path", "", "write results for failed tests to this path (path is relative to the test location)")
```
alpeb, line 29: Can you explain the benefits of this new functionality, compared to the current diffs shown in failures? Since this is separate than what the current PR addresses, I would prefer this to be submitted in a different PR.

charts/linkerd2/templates/proxy-injector-rbac.yaml
```diff
@@ -89,7 +91,7 @@ webhooks:
       name: linkerd-proxy-injector
       namespace: {{ .Values.global.namespace }}
       path: "/"
-    caBundle: {{ ternary (b64enc $ca.Cert) (b64enc (trim .Values.proxyInjector.crtPEM)) (empty .Values.proxyInjector.crtPEM) }}
+    caBundle: {{ ternary (b64enc $ca.Cert) (b64enc (trim .Values.proxyInjector.caBundle)) (empty .Values.proxyInjector.caBundle) }}
```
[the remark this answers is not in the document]
cypherfox: Added in `smi-metrics-rbac.yaml`, `sp-validator-rbac.yaml`, `tap-rbac.yaml`,  `proxy-injector-rbac.yaml`

cli/cmd/main_test.go
cypherfox: For failed tests, it writes the generated resource definitions to a given path. This makes it possible to compare the golden file and the test result with more powerful diff tools than the pretty print.

Background: When debugging certs, the diff often is in the middle of long strings of line noise. Neither the normal nor the pretty diff output are well suited for finding the reason for the difference, as the length and type of the diff will push out the context text out of the visible scope.

alpeb: Ah that's pretty neat :+1:

alpeb:
I see that the new test `upgrade_keep_webhook_cabundle.gold` checks the cert and bundle are conserved. Can you explain the purpose of "upgrade_two_level_webhook.cert.golden"?

cypherfox:
@alpeb The `upgrade_two_level_webhook.cert.golden` tests the whole setup with actually differing CA and webhook certificates (no self-signed webhook certs). The CA cert is the same for all webhooks / API services, but the actual certificates are unique.

I have also added a small script (in `bin/`) that helps in re-creating these certs when the validity period ends (in 3 years). It could easily be extended to contain all the other certificates used in the tests, but currently only does the ones for the two-level webhook certs. It will print in a format that can directly be used in the K8s resource definitions (base64 encoded PEM concatenated into single line)

alpeb:
Merging master one more time should bring a change that will help passing CI

cypherfox:
Ok, this one worked. I removed the script and added a comment to the particular test case.

charts/linkerd2/templates/proxy-injector-rbac.yaml
```diff
@@ -89,7 +91,10 @@ webhooks:
       name: linkerd-proxy-injector
       namespace: {{ .Values.global.namespace }}
       path: "/"
-    caBundle: {{ ternary (b64enc $ca.Cert) (b64enc (trim .Values.proxyInjector.crtPEM)) (empty .Values.proxyInjector.crtPEM) }}
+{{- if and (.Values.proxyInjector.externalSecret) (empty .Values.proxyInjector.caBundle) }}
+  {{- fail "If proxyInjector.externalSecret is true then you need to provide proxyInject.caBundle" }}
```
Pothulapati, line 95: This should be `proxyInjector.caBundle`, right?

charts/linkerd2/README.md
```diff
@@ -152,12 +152,24 @@ their default values.
...
 | `profileValidator.crtPEM`                   | Certificate for the service profile validator. If not provided then Helm will generate one.                                                                                           |                                      |
 | `profileValidator.keyPEM`                   | Certificate key for the service profile validator. If not provided then Helm will generate one.                                                                                       |                                      |
+| `profileValidator.caBundle`                 | Bundle of CA certificates for service profile validator. If not provided then Helm will use the certificate generated  for `profileValidator.crtPEM`. If `profileValidator.externalSecret` is set to true, this value must be set, as no certificate will be generated.         |  |
+| `smiMetrics.enabled`                        | Enable collection of SMI metrics by setting to `true`. Default is `false`
```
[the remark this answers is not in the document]
cypherfox: I have documented all options that are present in the chart today (in part as requested by alpeb), as some where simply not documented anywhere (please see also #4226, which I did in parallel for some of the discussion). Whether these need to be expanded e.g. to indicate that they are experimental, I would consider to be beyond the scope of this ticket.

cypherfox:
> When there is an external caBundle passed i.e `externalSecret`, we are skipping the secret creation. which means that we can skip volume and volumeMounts creation in resource yaml's (like for tap https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/tap.yaml#L126), Right?

I am not quite sure I understand your question correctly. The external secrets in this MR only cover those for the self-signed root certificates that are used in the webhook / api service configurations. 

They are separate from the identity certificate and secret mechanisms. They can already be handle through an external mechanism, e.g. cert-manager as described in https://linkerd.io/2/tasks/automatically-rotating-control-plane-tls-credentials/

My MR makes it possible to safely rotate the webhook / API service certificates without a downtime.

charts/linkerd2/templates/proxy-injector-rbac.yaml
cypherfox: yes

cypherfox:
correct.

Pothulapati:
Hmmm... So I was thinking more about this. If we are not supporting `externalSecrets` through CLI, I don't think it's worth the complexity to be able to support upgrades through CLI for it. 

Removing the upgrade stuff would simplify the change a lot, while also supporting this use through Helm.

@grampelberg

cypherfox:
> Hmmm... So I was thinking more about this. If we are not supporting `externalSecrets` through CLI, I don't think it's worth the complexity to be able to support upgrades through CLI for it. Especially because our CLi upgrades don't cause downtime currently right? As we re-use existing secrets.

This is not completely correct: currently the update, without explicit cert and key (and now caBundle) for the webhooks, will trigger the generation of a new self-signed certs. This will in turn open a window of unavailability of the webhooks / api services. 
If the update that triggers this behaviour will also require side-cars to be restarted / updated it will quickly escalate into the application services becoming unavailable, as updates will not apply successfully (e.g. due to the AdmissionHook to be unavailable).

None of this has much impact on a small cluster and will quickly settle. But our production cluster takes about 20-90 minutes to even completely rotate updated side cars at the best of times. It will quickly throw things into backoff when the admission hooks fail.

cypherfox:
The main reason for me to start this MR was the ability to rotate the certs of the webhooks with the same mechanism for a compound caBundle that is documented for the indentity certs. IMO this is the only way to completely eliminate downtimes. But the CLI should have the ability to update the configration of the webhooks, but still propagate the existing caBundles.

Pothulapati:
@cypherfox We should totally take this change if the there is downtime because of this upgrades. As I see that it is mentioned https://github.com/linkerd/linkerd2/blob/d31fe45e0a32b44f1f05a08d234d5b691e126078/cli/cmd/upgrade.go#L313 that we re-use secrets TLS secrets, My expectation was that there would be no downtime. :thinking: 

If there is downtime, I'm all in this for the change! :)

cypherfox:
We have experienced this downtime when we switched from our home grown solution to the upstream chart. Because 5 people where staring at it, we kept restarting the right pods to bring it down to only 1-2 minutes and it did not stop the production, since we did the whole thing in stages.

But during the roll-over we could not have recovered any failed linkerd side-cars. We have the production services in a HA setup, so even a failing side-car or application pod would not have taken town or service, but still.

For us the biggest problem is the automatic recreation of the cert and key during each helm apply: 1-2 minutes of no new side-cars, every 10-20 minutes. The alternative is forcing the values by setting them in a `values.yaml` that can be seen by the application devs (so that they can set basically everything else to enable DevOps).

For our specific needs this MR also enables a "no-humans-allowed" policy, since we have an integration that will inject secrets that no human ever gets to see, but I would assume that this not required by everyone. 

Separating `caBundle` and cert is required for both problems and allowing the secrets to be created by somebody else enables the second use-case. And there are others like company policies that have to tie all certs into single tree (or very small copse of trees) that will be enabled by this MR as well. 

I do not see how to reduce the complexity while still supporting transparency for all those that do not need it. Alternatives like #4232 would require the implementation of a new component that will not add features beyond what this MR represents and require an increase in configuration complexity of the CLI. It will also introduce a new (internal) interface that authors of external credential managers would have to maintain. At lot of the complexity of this MR IMHO is due to the Helm/Golang code path duality of the linkerd CLI (which is already visible in #4232) . 

This MR should work transparently for the starting (linkerd CLI) user as well as the advanced (Helm-) user.

cypherfox:
> that we re-use secrets TLS secrets, My expectation was that there would be no downtime. thinking

There is no downtime if you set the values explicitly. Otherwise the CLI will trigger a re-creation of the certs and keys as it renders the Helm fragment.

alpeb:
Thanks a lot for this @cypherfox !
Please fill out this [form](https://docs.google.com/forms/d/e/1FAIpQLSdXA7kAl6oyUQ4DFRjiDc2biWlvkhNirrMwsXaQ8V3Uh8Yj5A/viewform) to receive some well deserved swag!

Issue #4197, opened 2020-03-24
Linkerd CLI should default to current namespace from kubeconfig


praseodym:
### What is the issue?

I'm using [`kubens`](https://github.com/ahmetb/kubectx) to quickly switch the default namespace that kubectl uses. This works by rewriting the current context in `~/.kube/config`. For example, executing `kubens ingress-nginx` means that `kubectl get pod nginx-ingress-controller-84dd6b84fd-gcb6x` picks up the pod from the `ingress-nginx` namespace without explicitly specifying it in every `kubectl` command.

The Linkerd CLI doesn't use this context namespace. For example, `linkerd metrics` requires the namespace to be explicitly specified:
```
> linkerd metrics pod/nginx-ingress-controller-84dd6b84fd-gcb6x
Error: pods "nginx-ingress-controller-84dd6b84fd-gcb6x" not found
[..]
> linkerd metrics pod/nginx-ingress-controller-84dd6b84fd-gcb6x -n ingress-nginx
#
# POD nginx-ingress-controller-84dd6b84fd-gcb6x (1 of 1)
#
[..]
```

### Environment

- Kubernetes Version: v1.17.2
- Cluster Environment: kind
- Host OS: Debian Bullseye
- Linkerd version: edge-20.3.3

### Possible solution

Use the namespace from the current context in `~/.kube/config` as the default in the Linkerd CLI.

Matei207:
I'd like to pick this up if nobody else is working on it.

alpeb:
@Matei207 Great! please proceed :+1:

Matei207:
Hi, I have started looking at this issue and would like to discuss a potential solution to see what the rest of the community thinks.

I've looked through the `cli` commands to see which commands accept a namespace passed as a flag, in total there should be 7 commands that would need to infer the namespace from the `kubeConfig` context.

We can get namespace of a context from a `ClientConfig`, e.g:

```
	rules := clientcmd.NewDefaultClientConfigLoadingRules()
	overrides := &clientcmd.ConfigOverrides{CurrentContext: kubeContext}
	kubeCfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
	ns, _, err := kubeCfg.Namespace()
	return ns
```

For each individual command, we change its options from the hardcoded "default" to a call to the above function, e.g:

```
return &metricsOptions{
	namespace: "default",
	pod:       "",
}
```
becomes 

```
return &metricsOptions{
[...]

PR #4291, opened 2020-04-24
Add Kubeconfig contex namespace to cli commands' options (#4197)

Matei207:
### Issue: #4197 
---

### What

According to the original issue, when using a `cli` command that operates on namespaced resources, the `-n` flag has to be used even when the resources are in the __default namespace of the Kubeconfig context__.  For example, if the currently used context by the Kubeconfig has the default namespace as `emojivoto`, doing `kubectl get pods` should list everything under the `emojivoto namespace`. Issue #4197 requested similar behaviour out of the `cli` commands. 
 

### How

* The `cli` has a set of options for each command. In implementing this feature, my first step was to compile a list of commands that act on namespaced resources and accept an `-n` flag. The command names are listed below:
```
(linkerd)
1. metrics
2. edges
3. get
4. profile
5. routes
6. top
7. tap
```

* I added a new function in `cli/cmd/root.go` called `getDefaultNamespace()` that infers the default namespace based on the currently active context. To do, it makes use of loading rules and and a `DeferredLoadingClient`. Once the loader is initialised with the loading rules, we can get access to the client configuration (i.e kubeconfig) and find the namespace associated with the currently active context (through the `Namespace()` function).
* For each command previously listed, in the `options` structure, instead of hardcoding the default namespace as `default` or `""`, I initialise it using the `getDefaultNamespace()` function.

* I thought in this case unit tests are trivial, since it would only test if a string = another string, and mostly test the functionality of `client-go`. Instead, I wrote a new function in the integration test helper that allows changing the default namespace of the currently used context. 
* I thought a full integration test might not be the best use of space, so I have instead added two new cases to the integration tests of `linkerd get` and `linkerd tap` respectively. My rationale is that if it works for these two, it will work for the rest since it's the same principle. I chose `get` and `tap` because they included more than one case each and it made the most sense (from a design perspective).

Looking forward to some feedback :) 

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	"github.com/spf13/pflag"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/fatih/color"
 	log "github.com/sirupsen/logrus"
@@ -139,7 +141,7 @@ type statOptionsBase struct {
 
 func newStatOptionsBase() *statOptionsBase {
 	return &statOptionsBase{
-		namespace:    "default",
+		namespace:    getDefaultNamespace(),
 		timeWindow:   "1m",
 		outputFormat: tableOutput,
 	}
@@ -187,6 +189,27 @@ func getSuccessRate(success, failure uint64) float64 {
 		return 0.0
 	}
 	return float64(success) / float64(success+failure)
+}
+
+// getDefaultNamespace fetches the default namespace
+// used in the current KubeConfig context
+func getDefaultNamespace() string {
+	rules := clientcmd.NewDefaultClientConfigLoadingRules()
+
+	if kubeconfigPath != "" {
+		rules.ExplicitPath = kubeconfigPath
+	}
+
+	overrides := &clientcmd.ConfigOverrides{CurrentContext: kubeContext}
+	kubeCfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
+	ns, _, err := kubeCfg.Namespace()
+
+	if err != nil {
+		log.Errorf("failed to set namespace from config context:%v", err)
+		return corev1.NamespaceDefault
+	}
+
+	return ns
 }
 
 // proxyConfigOptions holds values for command line flags that apply to both the
--- a/cli/cmd/edges.go
+++ b/cli/cmd/edges.go
@@ -23,7 +23,7 @@ type edgesOptions struct {
 
 func newEdgesOptions() *edgesOptions {
 	return &edgesOptions{
-		namespace:     "",
+		namespace:     getDefaultNamespace(),
 		outputFormat:  tableOutput,
 		allNamespaces: false,
 	}
--- a/cli/cmd/get.go
+++ b/cli/cmd/get.go
@@ -18,7 +18,7 @@ type getOptions struct {
 
 func newGetOptions() *getOptions {
 	return &getOptions{
-		namespace:     "default",
+		namespace:     getDefaultNamespace(),
 		allNamespaces: false,
 	}
 }
--- a/cli/cmd/metrics.go
+++ b/cli/cmd/metrics.go
@@ -22,7 +22,7 @@ type metricsOptions struct {
 
 func newMetricsOptions() *metricsOptions {
 	return &metricsOptions{
-		namespace: "default",
+		namespace: getDefaultNamespace(),
 		pod:       "",
 	}
 }
--- a/cli/cmd/profile.go
+++ b/cli/cmd/profile.go
@@ -27,7 +27,7 @@ type profileOptions struct {
 func newProfileOptions() *profileOptions {
 	return &profileOptions{
 		name:          "",
-		namespace:     "default",
+		namespace:     getDefaultNamespace(),
 		template:      false,
 		openAPI:       "",
 		proto:         "",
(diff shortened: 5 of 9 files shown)


grampelberg:
The context is insanely helpful, thank you @Matei207 !

Matei207:
> The context is insanely helpful, thank you @Matei207 !

Thank you for the encouragement! I seemed to have run into some issues with the integration tests, I will deal with it asap. Meanwhile, if there are any other changes that would benefit this PR let me know and I'll add them in.

cli/cmd/root.go
```diff
@@ -189,6 +191,30 @@ func getSuccessRate(success, failure uint64) float64 {
...
+		rules.ExplicitPath = kubeconfigPath
+	}
+
+	overrides := &clientcmd.ConfigOverrides{CurrentContext: kubeContext}
+	kubeCfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
+	ns, _, err := kubeCfg.Namespace()
+
+	if err != nil {
+		log.Errorf("failed to set namespace from config context:%v", err)
+	}
```
zaharidichev, line 209: Cant you just return the defaultContextNamespace and call it a day instead of doing another check for it being empty further down?

testutil/kubernetes_helper.go
```diff
@@ -294,3 +294,12 @@ func (h *KubernetesHelper) URLFor(namespace, deployName string, remotePort int)
 
 	return pf.URLFor(""), nil
 }
+
+// ChangeK8sCfgContextNamespace modifies the current KubeConfig context
+// to use the provided namespace as the default for all subsequent kubectl commands
+func (h *KubernetesHelper) ChangeK8sCfgContextNamespace(namespace string) (string, error) {
```
zaharidichev, line 300: Can't you leverage `func (h *KubernetesHelper) Kubectl(stdin string, arg ...string)` somehow ?

cli/cmd/root.go
Matei207: Can do, yes. This is how it was done initially, so I will add it back in. Alternatively, another solution that I've been hesitant on would be to just log the error and return the `ns`. If there is an error, the `ns` would be `""` which is equal to the default anyway. I thought initially this approach broke the unit tests but it doesn't. Let me know if that sounds better and I'll put it in.

testutil/kubernetes_helper.go
Matei207: Done!

cli/cmd/root.go
```diff
@@ -189,6 +191,27 @@ func getSuccessRate(success, failure uint64) float64 {
...
+
+	overrides := &clientcmd.ConfigOverrides{CurrentContext: kubeContext}
+	kubeCfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
+	ns, _, err := kubeCfg.Namespace()
+
+	if err != nil {
+		log.Errorf("failed to set namespace from config context:%v", err)
+		return defaultContextNamespace
+	}
```
zaharidichev, line 210: I start to wonder how severe this error actually is. Do we want to recover from it or we rather fail the whole command invocation right away? Is it safe to assume that if we error out here, we can simply return the "default" namespace ?

Matei207: Hi @zaharidichev  I've looked at the `Namespace()` method implementation in the `DeferredLoadingClientConfig` type to determine what will explicitly happen if an error is encountered.  The function errors out if the client config cannot be created properly (in which case it returns an empty string), the client config is created based on the rules and rule loader that are created at the start of the `getDefaultNamespace()` function. The rest of the `Namespace()`method either returns the namespace defined in the context or if the namespace is `"default"` tries to work out whether it was an implicit or explicit choice.

> Is it safe to assume that if we error out here, we can simply return the "default" namespace

I assume so, yes. If the function fails, it means for whatever reason, the config was not properly initialised/created. The single point of error when creating the client config is when the rules are loaded in. A reason for failure, in this case, would be that the config cannot be parsed or loaded from a `kubeConfig` file.

To my understanding (and skill), this is what is happening under the hood when we are creating and loading those config "rules". Based on this, I think on error it is safe to just return the `"default"` namespace. In this context (no pun intended) it seems a default is more sensible since we might not know what the underlying issues with the `kubeConfig` are in the `cli` client environment.

Tl;dr To my understanding the point of error happens when either the config or the rules are not loaded in properly, in which case I think it is sensible to just return a default. If the problem persists (i.e context namespace simply won't be inferred), the namespace can just be overridden with the `-n` flag. If instead we just fail the command invocation outright, it might be that none of the commands will work, since all `cli` command options are making a call to `getDefaultNamespace()` so options will never be initialised if it all fails.

Of course, this is just my opinion, look forward to hearing what you think :)

zaharidichev: Yes, I see. That makes sense to me.

cli/cmd/root.go
```diff
@@ -17,10 +18,11 @@ import (
...
 const (
-	defaultNamespace      = "linkerd"
-	defaultCNINamespace   = "linkerd-cni"
-	defaultClusterDomain  = "cluster.local"
-	defaultDockerRegistry = "gcr.io/linkerd-io"
+	defaultNamespace        = "linkerd"
+	defaultCNINamespace     = "linkerd-cni"
+	defaultClusterDomain    = "cluster.local"
+	defaultDockerRegistry   = "gcr.io/linkerd-io"
+	defaultContextNamespace = "default"
```
zaharidichev, line 25: you can use `corev1.NamespaceDefaul` .

Matei207: Since the constant is used in only one space, I have replaced it with `corev1.NamespaceDefault` in the function return. Think this might show intent a bit better than the solution I had initially.

Issue #4176, opened 2020-03-16
CI health and metrics


alpeb:
### What problem are you trying to solve?

- We'd like to have a global idea of how CI is performing.
- If there are recurring errors, we'd like to easily tell them apart.

### How should the problem be solved?

- [x] Modify tests in go so they properly surface errors as Github annotations when running in CI
- [x] Modify tests in javascript so they properly surface errors as Github annotations when running in CI

alpeb:
This has been addressed in the [linkerd2-ci-metrics repo](https://github.com/linkerd/linkerd2-ci-metrics). Check the readme there for a detailed description.

PR #4292, opened 2020-04-24
Go test failure message wrappers to create GH Annotations

alpeb:
First part of #4176

## Problem

Failures in go tests need to be properly formatted as Github annotations
so that we can fetch them through Github's API for aggregation and
analysis.

## Solution

A wrapper for error messages has been created in `testutil/annotations.go`.
The idea is that instead of throwing test failures like this:

```go
t.Failf("error retrieving data;\nExpected: %#v\nActual: %#v", expected,
actual)
```

We'd throw them like this:
```go
testutil.AnnotationFatalf("error retrieving data", "error retrieving data;\nExpected: %#v\nActual: %#v", expected,
actual)
```

That will continue reporting the error as before (when using `go test`
or another test runner), but as a side-effect it will also send to
stdout something like:

```
::error file=pkg/inject_test.go,line=133::error retrieving data
```
Which becomes a GH annotation, visible in the CI run summary screen.

The fist string art is used to have the GH annotation be a generic error message
that can be aggregated and counted across multiple test runs. If `testutil.Fatalf(str, args...)`
is called instead, the original error message will be used.

Note that that the output will be produced only when the env var
`GH_ANNOTATION` is set (which will when tests are triggered from a
Github Actions workflow).

Besides `testutil/annotation.go` and its accompanying unit test file,
other changes were made in other tests as examples, the plan being that
in a further PR _all_ the tests will use these wrappers.

Edit: replaced ext_err approach with annotations.go


alpeb:
Thanks @kleimkuhler. Yeah, the followup is to use that in all the go tests. And indeed, the plan is to use `WithAnn` only in those places where the original error would be too specific (say containing details such as a pod name that is different on each run), the idea being to tell how many times that type of error appeared.

pkg/healthcheck/healthcheck_test.go
```diff
@@ -2302,12 +2304,12 @@ metadata:
 		hc.kubeAPI, err = k8s.NewFakeAPI(resources...)
 		hc.ControlPlaneNamespace = "test-ns"
 		if err != nil {
-			t.Fatalf("Unexpected error: %s", err)
+			t.Fatal(testutil.Errorf("Unexpected error: %s", err).WithAnn("Unexpected error"))
```
adleong, line 2307: What do you think about having `Error`, `Errorf`, `Fatal`, and `Fatalf` methods in `testutil` that take `testing.T` as a parameter?  I think there are a few advantages to this:
* less verbose at the call site: `testutil.Fatalf(t, "Unexpected error: %s", err)`
* easier to audit for tests that don't use the wrapper: no test should call `t.Errorf` etc directly
* Possible to add `testutil.AnnotatedErrorf(t, "Unexpected error", "Unexpected error: %s", err)` style methods to avoid the `with` design pattern.

testutil/ext_err.go
```diff
@@ -0,0 +1,74 @@
...
+// as a Github annotation when the envFlag environment variable is set.
+type ExtError struct {
+	err  error
+	file string
+	line int
+	desc string
+}
+
+func newExtError(err error, desc string) ExtError {
+	_, fn, ln, ok := runtime.Caller(2)
```
adleong, line 24: TIOLI: I always read `fn` as "function" instead of "filename".

testutil/ext_err.go
```diff
@@ -0,0 +1,74 @@
...
+	if !ok {
+		panic("Couldn't recover runtime info")
+	}
+	fileName := fn[strings.LastIndex(fn, "/linkerd2/")+10:]
```
adleong, line 28: TIOLI: move "/linkerd2/" into a const and use len of that const instead of magic number 10

testutil/ext_err.go
```diff
@@ -0,0 +1,74 @@
...
+	extErr.desc = desc
+	return extErr
+}
+
+// Error ensures that the struct ExtError implements the `error` interface.
+// When ExtError.Error() is called, the underlying Error() is returned, and
+// as a side-effect (only if the envFlag environment variable is set) a Github
+// annotation string is sent to stdout
+func (extErr ExtError) Error() string {
+	if _, ok := os.LookupEnv(envFlag); ok {
```
adleong, line 70: Not sure if this matters, but this won't change during the lifetime of the process, so we only need to lookup the env once.

testutil/ext_err.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// WithAnn is to be called on a ExtError type to provide a generic
+// error description
+func (extErr ExtError) WithAnn(desc string) ExtError {
+	extErr.desc = desc
+	return extErr
+}
+
+// Error ensures that the struct ExtError implements the `error` interface.
```
adleong, line 65: Things will be a bit weird if callers do anything with this Error other than just immediately pass it to `t.Error`.  Anyone calling `.Error()` on it will cause an annotation to be printed at that time.

pkg/healthcheck/healthcheck_test.go
alpeb: The problem with that approach is that if a test fails, it'll fail inside those functions in `testutil` and the test runner won't appropriately report where the failure happened.

testutil/ext_err.go
alpeb: true :+1:

pkg/healthcheck/healthcheck_test.go
adleong: Does this help with that problem: https://golang.org/pkg/testing/#T.Helper ?

testutil/ext_err.go
alpeb: I could have it check in the call stack whether the call comes from `testing.Error`, and only in that case produce the annotation.
That being said, I share the feeling that all this is a bit hacky, but I can't figure an alternative way while conserving the default testing framework for go...

pkg/healthcheck/healthcheck_test.go
alpeb: Ah that's awesome, thanks!

testutil/ext_err.go
alpeb: Ok this should no longer be a concern if I adopt the `testutil.Fatalf()` approach you suggest above, dumping the `ExtError` struct :+1:

alpeb:
Ok, I've just pushed a changed applying @adleong suggestions. I couldn't move the env var check into a global because it would get triggered before the unit tests have a chance to change it. Let me know what you think.

PR #4294, opened 2020-04-24
Additional Jest reporter for GH Annotations

alpeb:
Second part of #4176

Added extra Jest reporter when running js tests from CI, which will send
to stdout a GH annotation for each test failure, something like:

```
::error file=web/app/js/components/Navigation.test.jsx::Navigation › checks state when versions do not match
```

See the [health metrics RFC](https://github.com/linkerd/rfc/blob/master/design/0002-ci-health-metrics.md) for more context.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -40,4 +40,4 @@ jobs:
         export PATH="$HOME/.yarn/bin:$PATH"
         export NODE_ENV=test
         bin/web
-        bin/web test --reporters=jest-dot-reporter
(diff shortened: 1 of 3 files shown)


alpeb:
> Works for me, does github handle unicode okay?

Yep, that little caret ends up surfacing well all the way into the report charts. Here's a sneak peek:
image

grampelberg:
Nice!

alpeb:
> There is not a follow-up for this correct? Passing in the `--reporters` flags will add this to all JS test failures?

Correct, this is it.

PR #4300, opened 2020-04-27
Inject LINKERD2_PROXY_DESTINATION_GET_NETWORKS proxy variable

### LINKED ISSUE #3807: Proxy injector should set destination lookup subnets
The `LINKERD2_PROXY_DESTINATION_GET_NETWORKS` environment variable controls which IP subnets Linkerd will do destination lookups for.  If unset, it currently defaults to `"10.0.0.0/8,172.16.0.0/12,192.168.0.0/16"`.

To move this defaulting logic out of the proxy and into the proxy-injector, we should have the proxy injector always set this environment variable to this value.

We do not need to make this configurable through CLI flags or annotations yet, though we may want to consider this in the future.  For now, this value can be edited manually by using `--manual` injection and editing the environment variable value in the resulting yaml.

adleong:
Fixes #3807

By setting the LINKERD2_PROXY_DESTINATION_GET_NETWORKS environment variable, we configure the Linkerd proxy to do destination lookups for authorities which are IP addresses in the private network range.  This allows us to get destination metadata including identity for HTTP requests which target an IP address in the cluster, Prometheus metrics scrape requests, for example.

This change allowed us to update the "direct edges" test which ensures that the edges command produces correct output for traffic which is addressed directly to a pod IP.

We also re-enabled the "linkerd stat" integration tests which had been disabled while the destination service did not yet support these types of IP queries.

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -4,6 +4,8 @@ env:
   value: {{.Values.global.proxy.logLevel}}
 - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
   value: {{ternary "localhost.:8086" (printf "linkerd-dst.%s.svc.%s:8086" .Values.global.namespace .Values.global.clusterDomain) (eq .Values.global.proxy.component "linkerd-destination")}}
+- name: LINKERD2_PROXY_DESTINATION_GET_NETWORKS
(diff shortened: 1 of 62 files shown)


adleong:
@kleimkuhler @zaharidichev thanks for the reviews!  getting this through CI required a few more changes since we had integration tests which were testing the old behavior and needed to be updated and re-enabled.  The description of this PR has been updated with the details.  Please take another look, thanks!

PR #4302, opened 2020-04-28
Adding means to build the runtime docker images for arm32v7 and arm64v8

### LINKED ISSUE #3114: Build: Support Multi Arch

jaerik:
Partly fixes #3114

Adding ability to build the runtime docker images for amd64, arm32v7, and arm64v8 regardless of host architecture and without the need of qemu (docker buildx). The solution take advantage the host architecture version of apt-get and dpkg to manage the target image's package needs.


olix0r:
@jaerik This is really cool, thanks! Would you mind opening a PR against the linkerd2-proxy repo with the proxy's changes (i.e. replacing the patch file in this repo)? I want to test that out in the context of the proxy's CI process.

Thanks!

jaerik:
> Would you mind opening a PR against the linkerd2-proxy repo with the proxy's changes (i.e. replacing the patch file in this repo)? I want to test that out in the context of the proxy's CI process.

@olix0r I have opened an PR [#497](https://github.com/linkerd/linkerd2-proxy/pull/497) against the linkerd2-proxy repo with the changes for the makefile.

jaerik:
I don't see the connection between the failed test and the changes I made in the code. Have I done something wrong or is the test misleading?

alpeb:
@jaerik it's ok it was just test flakiness

.proxy-init-version
```diff
@@ -0,0 +1 @@
+v1.3.2
```
joakimr-axis, line 1: Should probably be v1.3.3 now.

Dockerfile-proxy
```diff
@@ -1,15 +1,21 @@
-ARG RUNTIME_IMAGE=debian:stretch-20190812-slim
+ARG ARCH=amd64
+ARG RUNTIME_IMAGE=$ARCH/debian:buster-20200422-slim
```
joakimr-axis, line 2: Perhaps update to `buster-20200514-slim` which is what is now used for buster in other places in the codebase.

Dockerfile-proxy-init
```diff
@@ -0,0 +1,61 @@
+ARG ARCH=amd64
+ARG GO_VERSION=1.12.9
+ARG VERSION=v1.3.2
```
joakimr-axis, line 3: Should probably be v1.3.3 now.

Dockerfile-proxy-init
```diff
@@ -0,0 +1,61 @@
...
+ARG DEBIAN_VERSION=buster-20200422-slim
```
joakimr-axis, line 4: See previous buster comment.

web/Dockerfile
```diff
@@ -1,3 +1,8 @@
+ARG ARCH=amd64
+ARG RUNTIME_IMAGE=$ARCH/debian:buster-20200422-slim
```
joakimr-axis, line 2: And here

adleong:
@jaerik it looks like there's now a merge conflict on this PR that you'll need to resolve.

@olix0r now that associated linkerd2-proxy PR has merged, do you want to give this another look?

aliariff:
Hi everyone,

I want to point out something, is that the approach of creating the Dockerfile for proxy-init & proxy and perform the compilation in this repo a correct approach? 
It seems like we duplicated the logic of build the image over here, as we already have our own repo for that in `linkerd2-proxy-init` & `linkerd2-proxy`, and that repos will have their own build & release process including supporting multiple architectures.
cc: @cpretzer @alpeb

cpretzer:
hi @aliariff, you're right, the current plan is to keep the Dockerfile for `linkerd-proxy-init` and `linkerd-proxy` in their own repositories.

jaerik:
> hi @aliariff, you're right, the current plan is to keep the Dockerfile for `linkerd-proxy-init` and `linkerd-proxy` in their own repositories.

Okay, I will move these parts to separate PR in the linkerd-proxy-init repository respective the linkerd-proxy repository.

joakimr-axis:
> @olix0r changed the base branch from master to main 13 hours ago

Out of curiosity: What is this new `main` branch and how does it relate to `master`?

olix0r:
main is replacing master. master will be deleted later this week to avoid confusion

alpeb:
@jaerik Please let me know when this is again ready for review. I guess the extraction of the linkerd-proxy and linkerd-proxy-init is still pending, right?

cpretzer:
@jaerik, thanks for your work on this, it set the foundation for a larger set of changes that we made in each of the repositories: [linkerd2-proxy](https://github.com/linkerd/linkerd2-proxy/pull/593), linkerd2-proxy-init [12](https://github.com/linkerd/linkerd2-proxy-init/pull/12/files) and [15](https://github.com/linkerd/linkerd2-proxy-init/pull/12/files), and [linkerd2](https://github.com/linkerd/linkerd2/pull/4765)

I'm going to close this one for now. If there is anything in the commits above that isn't covered, please let us know with a comment! 🙏

PR #4307, opened 2020-04-29
Make export-service non side-effecting

zaharidichev:
This change alters the `linkerd cluster export-service` command to behave much like the inject command, taking a stream of yaml and   modifying it by adding the needed annotations for exporting a service. 

### DIFF
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


cli/cmd/cluster.go
```diff
@@ -362,50 +492,20 @@ func newExportServiceCommand() *cobra.Command {
...
 			}
 
-			rules := clientcmd.NewDefaultClientConfigLoadingRules()
-			rules.ExplicitPath = kubeconfigPath
-			loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, &clientcmd.ConfigOverrides{})
-			config, err := loader.RawConfig()
-			if err != nil {
-				return err
+			if len(args) < 1 {
+				return fmt.Errorf("please specify a kubernetes resource file")
```
alpeb, line 496: This is duped from right above

PR #4310, opened 2020-04-29
Hide development flags in release builds

adleong:
Certain install flags are intended to help with Linkerd development and generally are not useful (and are potentially confusing) to users.

We hide these flags in release (edge or stable) builds of the CLI but show them in all other builds.  The list of affected flags is:

* control-plane-version
* proxy-image
* proxy-version
* image-pull-policy
* init-image
* init-image-version

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -516,9 +516,8 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
 		"Enables installing the SMI-Metrics controller",
 	)
 
-	flags.StringVarP(&options.controlPlaneVersion, "control-plane-version", "", options.controlPlaneVersion, "(Development) Tag to be used for the control plane component images")
+	flags.StringVarP(&options.controlPlaneVersion, "control-plane-version", "", options.controlPlaneVersion, "Tag to be used for the control plane component images")
 	flags.StringVar(&options.smiMetricsImage, "smi-metrics-image", options.smiMetricsImage, "SMI Metrics image")
-	flags.MarkHidden("control-plane-version")
 	flags.MarkHidden("control-plane-tracing")
 	flags.MarkHidden("smi-metrics")
 	flags.MarkHidden("smi-metrics-image")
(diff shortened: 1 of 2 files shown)


olix0r:
@adleong This seems mostly useful for Linkerd developers. I don't think this is a feature we want users to try to use... I think that's the context behind it being hidden, anyway.

Can we hide it in release builds?

adleong:
It's theoretically possible by using ldflags, but would require some plumbing.

But given that we already expose quite a few developer focused flags, I don't see the harm in exposing one more.  As a future action we could think about cleaning this up and hiding all developer focused flags in release builds.

olix0r:
@adleong Heh, well, you know me.. I'd rather hide those other flags.

My way of thinking is that `install`/`upgrade` are meant to be simple interfaces for getting started, but more serious customization should be done by helm etc...

My main concern is that we have a pretty tight coupling between the install/upgrade command and the control planes they install. Very bad  things can happen if people try to use upgrade with a specified control plane version (i.e. without the newer version of the linkerd CLI). So I'd really rather not even open the possibility for having to answer those questions in Slack ;p

adleong:
Reworked this to hide a number of development flags in release builds and show them in non-release builds.

olix0r:
@adleong Thanks! I had a thought--if you think this is simpler or more ergonomic--that we could support a `LINKERD_DEV` env variable that controls whether these are shown or not. But that's in the territory of yakshave. I'm also happy to only include them when doing dev builds... Whatever folks think is most useful (but hides internal flags for most users..)

adleong:
That's an interesting idea but I think using the build version is more ergonomic than adding an env variable.

cli/cmd/install.go
```diff
@@ -516,9 +516,23 @@ func (options *installOptions) recordableFlagSet() *pflag.FlagSet {
...
+
+	// Hide developer focused flags in release builds.
+	release, err := version.IsReleaseChannel(version.Version)
+	if err != nil {
+		log.Errorf("Unable to parse version: %s", version.Version)
+	}
+	if release {
+		flags.MarkHidden("control-plane-version")
+		flags.MarkHidden("proxy-image")
+		flags.MarkHidden("proxy-version")
```
olix0r, line 530: FWIW, I think I'd be comfortable leaving `proxy-version` exposed. We ask users to use it frequently to test fixes... But I can ask them to use hidden flags, too...

adleong:
I believe this requires a review from @alpeb or @zaharidichev before it can be merged.

PR #4311, opened 2020-04-30
Add check for validating that remote clusters share the same trust an…

zaharidichev:
In order to enable TLS across cluster boundaries, we need to ensure that the issuer certs can all be verified with a common trust acnhors. In roder to ensure that all remote cluster that are monitored, share the same anchors, we add them as annotation to the remote cluster config secret. This allows us to easily compre the anchors of all currently watched clusters to the local ones. This PR makes this process part of the `check` command

### DIFF
--- a/charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
+++ b/charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
@@ -12,10 +12,13 @@ metadata:
   annotations:
     {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
 rules:
-- apiGroups:
-  - ""
-  resources: ["services", "configmaps"]
+- apiGroups: [""]
+  resources: ["services"]
   verbs: ["list", "get", "watch"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get"]
+  resourceNames: ["linkerd-config"]  
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
+++ b/charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
@@ -14,12 +14,8 @@ metadata:
 rules:
 - apiGroups:
   - ""
-  resources:
-  - services
-  verbs:
-  - list
-  - get
-  - watch
+  resources: ["services", "configmaps"]
+  verbs: ["list", "get", "watch"]
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -362,18 +362,19 @@ type HealthChecker struct {
 	*Options
 
 	// these fields are set in the process of running checks
-	kubeAPI          *k8s.KubernetesAPI
-	kubeVersion      *k8sVersion.Info
-	controlPlanePods []corev1.Pod
-	apiClient        public.APIClient
-	latestVersions   version.Channels
-	serverVersion    string
-	linkerdConfig    *configPb.All
-	uuid             string
-	issuerCert       *tls.Cred
-	trustAnchors     []*x509.Certificate
-	cniDaemonSet     *appsv1.DaemonSet
-	serviceMirrorNs  string
+	kubeAPI              *k8s.KubernetesAPI
+	kubeVersion          *k8sVersion.Info
+	controlPlanePods     []corev1.Pod
+	apiClient            public.APIClient
+	latestVersions       version.Channels
+	serverVersion        string
+	linkerdConfig        *configPb.All
+	uuid                 string
+	issuerCert           *tls.Cred
+	trustAnchors         []*x509.Certificate
+	cniDaemonSet         *appsv1.DaemonSet
+	serviceMirrorNs      string
+	remoteClusterConfigs []*sm.WatchedClusterConfig
 }
 
 // NewHealthChecker returns an initialized HealthChecker
@@ -1302,6 +1303,14 @@ func (hc *HealthChecker) allCategories() []category {
 						return hc.checkRemoteClusterGatewaysHealth(ctx)
 					},
 				},
+				{
+					description: "clusters share trust anchors",
+					hintAnchor:  "l5d-clusters-share-anchors",
+					fatal:       true,
+					check: func(ctx context.Context) error {
+						return hc.checkRemoteClusterAnchors()
+					},
+				},
 			},
 		},
 	}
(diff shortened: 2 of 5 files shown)


grampelberg:
I'm not sure I understand how this is working. Would you mind doing a step by step?

zaharidichev:
@grampelberg Sure. So the secrets that contain the remote cluster api config that we generate with `get-credentials` are not modified to also contain the trust roots that are present in the linkerd config map. because these secrets end up on the local cluster, we can actually verify that all roots of remote clusters we are mirroring match with the roots of our local linkerd installation. @alpeb is right about these edge cases though.. Maybe this check is a bit too naive..

grampelberg:
@zaharidichev I see, but that would be a point in time snapshot. If someone was to rotate the certificate on the other side, it'd stop working again, right? Would the liveness check test for this? If it did, we'd need a good error message for this specific use case as my suspicion is that it will be the biggest problem most folks run into.

adleong:
I believe that liveness checks should fail if the trust roots don't match, but it may be difficult to get a helpful error message.  From the local proxy's point of view, it will just see this as a connection refused.  

Requiring that the trust roots match exactly is a bit naive but requiring this simplifies things.  Strictly speaking, we only need the remote's issuer certificate to validate against our trust root and vice versa.  But I don't think we have a good way of reading the issuer certificate without having access to the issuer secret.

@grampelberg makes a good point about the check being "point-in-time".  If the multi-cluster service account has permission to read configmaps in the control plane namespace, then the check could use the remote credentials to fetch the remote trust root when the check is run and compare it to the local trust root.

zaharidichev:
Very good feedback! So... 

> Thinking about edge cases, would cross-cluster communication work if they had different anchors that were themselves rooted on the same root cert? Or what if one if the clusters had an anchor with a bundle containing the other cluster's cert? If so, can this check be more generic to allow those situations?

@alpeb So in this case, I think we can be a bit less naive and actually try to verify the current issuer cert (we grab it from the identity issuer) of the local cluster with the roots of the remote cluster. Since the proxy certs are issued off of the issuer cert, that should cover all cases that you mention, right ? 

@grampelberg @adleong The point in time situation you are describing is correct. So this clearly has its limitations. Granting permissions to look-up the remote cluster's linkerd config map is fine with me. That was my original idea but decided to not be so relaxed with RBACs and opt for the current approach (clearly I did not think through the edge cases). The only complication I see here is that... the local cluster would have to know the namespace in which Linkerd is installed on the remote cluster. How can we do that. Is annotation on the credentials secret a good enough approach? Clearly we would have to assume that this  Linkerd installation does not move namespaces... 

One last thing.. 

> I believe that liveness checks should fail if the trust roots don't match, but it may be difficult to get a helpful error message. From the local proxy's point of view, it will just see this as a connection refused.

@adleong As a matter of fact the service mirroring controller, which is the same component performing the checks is not injected with the proxy by default. It is a separate chart, installed with a separate command and by default is running without the proxy. So the checks would not fail.. Do we want it to be injected? Is that a requirement? WDYT?

grampelberg:
@zaharidichev The two trust roots can be different, yet still allow traffic (imagine a whole trust chain, or a bundle with swapped entries). My suggestion was to test an actual connection and examine the error that comes back (invalid certificate).

charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
```diff
@@ -14,12 +14,8 @@ metadata:
 rules:
 - apiGroups:
   - ""
-  resources:
-  - services
-  verbs:
-  - list
-  - get
-  - watch
+  resources: ["services", "configmaps"]
```
adleong, line 17: Not sure if we care, but we can scope this permission down further.  We only need permission to read configmaps in the control plane namespace.  Or, even better, we only need permission to read specifically the linkerd config configmap.

pkg/healthcheck/healthcheck.go
```diff
@@ -1651,6 +1660,63 @@ func (hc *HealthChecker) checkServiceMirrorLocalRBAC() error {
...
+			offendingClusters = append(offendingClusters, fmt.Sprintf("* %s: cannot parse trust anchors", cfg.ClusterName))
+			continue
+		}
+
+		if len(remoteAnchors) != len(localAnchors) {
+			offendingClusters = append(offendingClusters, fmt.Sprintf("* %s", cfg.ClusterName))
+			continue
+		}
+
+		for i := 0; i < len(remoteAnchors); i++ {
```
adleong, line 1704: do we need to deterministically sort the trust anchors in each bundle to make sure they appear in the same order?

charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
zaharidichev: I looked at that option. I think it would require us to have a separate Role which is namespace scoped only to the control plane namespace. It will not limit us to only the particular config map though. Was not sure if it is worth it. @alpeb wdyt?

pkg/healthcheck/healthcheck.go
zaharidichev: Yeah, I can try to sort  them or just assume they must be in the same order. If they are spelled out in the same order in the PEM they should be in the same order in the pool as well. This is just assuming that the PEMs are identical, which I though is a valid assumption.

charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
grampelberg: You can add `subResources` to scope to specific names.

pkg/healthcheck/healthcheck.go
grampelberg: Can't we just parse and do a deep compare instead? Throw them all in a hash keyed off the fingerprint and ordering doesn't matter anymore.

pkg/k8s/labels.go
```diff
@@ -369,6 +369,10 @@ const (
 	// using custom cluster domains
 	RemoteClusterDomainAnnotation = SvcMirrorPrefix + "/remote-cluster-domain"
 
+	// RemoteClusterLinkerdNamespaceAnnotation is present on the secret
+	// carrying the config of the remote cluster
+	RemoteClusterLinkerdNamespaceAnnotation = SvcMirrorPrefix + "/remote-cluster-ld5-ns"
```
kleimkuhler, line 374: Typo here?
```suggested
	RemoteClusterLinkerdNamespaceAnnotation = SvcMirrorPrefix + "/remote-cluster-l5d-ns"
```

charts/linkerd2-multicluster-remote-setup/templates/service-mirror-rbac.yaml
```diff
@@ -12,10 +12,13 @@ metadata:
...
-- apiGroups:
-  - ""
-  resources: ["services", "configmaps"]
+- apiGroups: [""]
+  resources: ["services"]
   verbs: ["list", "get", "watch"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get"]
+  resourceNames: ["linkerd-config"]
```
adleong, line 21: I think this grants permission to read a configmap in any namespace as long as it is named "linkerd-config".  This is probably fine and I don't know that we really care that much, but just for the sake of discussion: I think the most "correct" way to do this would be to also create a Role and RoleBinding in the linkerd namespace that allow access to just the linkerd-config configmap in that namespace specifically.

pkg/servicemirror/util.go
```diff
@@ -9,16 +9,19 @@ import (
...
+	ClusterDomain    string
+	LinkerdNamespace string
 }
 
 // ParseRemoteClusterSecret extracts the credentials used to access the remote cluster
 func ParseRemoteClusterSecret(secret *corev1.Secret) (*WatchedClusterConfig, error) {
 	clusterName, hasClusterName := secret.Annotations[consts.RemoteClusterNameLabel]
 	config, hasConfig := secret.Data[consts.ConfigKeyName]
 	domain, hasDomain := secret.Annotations[consts.RemoteClusterDomainAnnotation]
+	ld5Namespace, hasLd5Namespace := secret.Annotations[consts.RemoteClusterLinkerdNamespaceAnnotation]
```
kleimkuhler, line 23:
```suggested
l5dNamespace, hasL5dNamespace := secret.Annotations[consts.RemoteClusterLinkerdNamespaceAnnotation]
```

pkg/healthcheck/healthcheck.go
```diff
@@ -1701,13 +1701,18 @@ func (hc *HealthChecker) checkRemoteClusterAnchors() error {
...
 		}
 
-		for i := 0; i < len(remoteAnchors); i++ {
-			if !remoteAnchors[i].Equal(localAnchors[i]) {
+		localAnchorsMap := make(map[string]*x509.Certificate)
+		for _, c := range localAnchors {
+			localAnchorsMap[string(c.Signature)] = c
+		}
+
+		for _, remote := range remoteAnchors {
```
adleong, line 1709: this checks that all of the remote certificates are in the local bundle, but don't we also need to check that all of the local certificates are in the remote bundle?  @grampelberg suggested doing a deepEquals of the maps, does that not work?

adleong: nevermind.  I see that we're comparing the lengths above, which makes this check good enough.  A comment to that effect might be helpful.

pkg/servicemirror/util.go
zaharidichev: ooops

kleimkuhler: Ah it's `'''suggestion` not `'''suggested`.. but yes you see what I'm pointing out 😆

pkg/healthcheck/healthcheck.go
```diff
@@ -1651,6 +1660,68 @@ func (hc *HealthChecker) checkServiceMirrorLocalRBAC() error {
...
 }
 
+func (hc *HealthChecker) checkRemoteClusterAnchors() error {
+	if len(hc.remoteClusterConfigs) == 0 {
+		return &SkipError{Reason: "no remote cluster configs"}
+	}
+
+	localAnchors, err := tls.DecodePEMCertificates(hc.linkerdConfig.Global.IdentityContext.TrustAnchorsPem)
+	if err != nil {
+		return fmt.Errorf("Cannot parse local trust acnhors: %s", err)
```
alpeb, line 1670: Typo: acnhors

PR #4312, opened 2020-04-30
Probe manager events emission tests

zaharidichev:
This PR provides a set of tests that verify the cluster watcher issues correct events to the probe manager when the state of the remote clusters is modified. This required a bit of reorg of the tests to make things slightly more reusable. Namely, we now have the concept of a test environment, that describes the state of the local and remote clusters and what events are enqueued. This makes it possible to model various scenarios that can be used to test both the modifications that are performed by the service mirroring component to the local cluster as well as the events it fires to the prober. 


grampelberg:
@zaharidichev this is adding some integration tests?

zaharidichev:
@grampelberg these are unit tests. Integration tests will come after. I am still not sure how exactly we are going to do integration testing here but will have to think about that

controller/cmd/service-mirror/cluster_watcher_test_util.go
```diff
@@ -2,16 +2,721 @@ package servicemirror
...
+func (te *testEnvironment) runEnvironment(probeEventSync ProbeEventSink, watcherQueue workqueue.RateLimitingInterface) (*k8s.API, error) {
+	remoteAPI, err := k8s.NewFakeAPI(te.remoteResources...)
+	if err != nil {
+		return nil, err
+	}
+
+	localAPI, err := k8s.NewFakeAPI(te.localResources...)
+	if err != nil {
+		return nil, err
+	}
```
alpeb, line 41: :cool:

PR #4317, opened 2020-04-30
Increase timeout for Helm cleanup in integration tests

alpeb:
Tests were failing sporadically, waiting for the Helm namespace to get
cleaned up. I verified that it is getting cleaned up, but taking more
time sometimes.


kleimkuhler:
It may be worth doing 120 or 180 seconds for the timeout? It's only a timeout so it doesn't mean it will always wait that long, and if the delete has already taken more than 40 seconds, another 20 may not be enough in the rare cases.

alpeb:
> It may be worth doing 120 or 180 seconds for the timeout? It's only a timeout so it doesn't mean it will always wait that long, and if the delete has already taken more than 40 seconds, another 20 may not be enough in the rare cases.

Makes sense, will do :+1:

PR #4318, opened 2020-05-01
Refactor linkerd-config-addons configmap

Pothulapati:
Initial part of #3979, as per https://github.com/linkerd/linkerd2/pull/3979#issuecomment-621928777 Feel free to look at it, to understand how these changes integrate with grafana add-on.

This PR renames `linkerd-values` to `linkerd-config-addons` to make it more specific to add-ons (which we planned to be more general, but it makes sense to make it specific to add-ons) and to be present by default. This PR also refactors the config-map to make it easy to add more add-ons. 

**changes include:**
- Rename `linkerd-values` to `linkerd-configmap-addons`.
- Add `_addons.tpl` which adds a function that takes care of the cleaning up of the add-on fields, which removes the need to address the sub-fields individually. (which we did previously)
- Use `_addons.tpl` function in `linkerd-configmap-addons`, and make it present by default.

### DIFF
--- a/charts/linkerd2/templates/_addons.tpl
+++ b/charts/linkerd2/templates/_addons.tpl
@@ -1,11 +1,16 @@
+{- /*
+  This function does the following functions for add-on fields:
+
+  - when an add-on is enabled, it removes the `partials` and `global` sub-fields which are automatically added
+    by helm to pass configuration to sub-charts, and returns the add-on fields.
+
+  - When the add-on is disabled, only the `enabled` flag is returned as we don't need to store the other configuration
+    when an add-on is disabled.
+*/ -}}
 {{- define "linkerd.addons.sanitize-config" -}}
 {{- if .enabled -}}
 {{- $dupValues := . -}}
 {{- if kindIs "map" $dupValues -}}
-  {{/*
-  Remove "global" and "partial" keys from the add-on structs as they are added by helm
-  to propogate values.
-  */}}
   {{- if (hasKey $dupValues "global") -}}
       {{- $dupValues := unset $dupValues "global" -}}
   {{- end -}}
--- a/charts/linkerd2/templates/_addons.tpl
+++ b/charts/linkerd2/templates/_addons.tpl
@@ -1,0 +1,20 @@
+{{- define "linkerd.addons.sanitize-config" -}}
+{{- if .enabled -}}
+{{- $dupValues := . -}}
+{{- if kindIs "map" $dupValues -}}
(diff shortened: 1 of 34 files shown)


charts/linkerd2/templates/_addons.tpl
```diff
@@ -0,0 +1,25 @@
...
+  - when an add-on is enabled, it removes the `partials` and `global` sub-fields which are automatically added
+    by helm to pass configuration to sub-charts, and returns the add-on fields.
+
+  - When the add-on is disabled, only the `enabled` flag is returned as we don't need to store the other configuration
+    when an add-on is disabled.
+*/ -}}
+{{- define "linkerd.addons.sanitize-config" -}}
+{{- if .enabled -}}
+{{- $dupValues := . -}}
+{{- if kindIs "map" $dupValues -}}
```
alpeb, line 13: I think it's safe to assume this will always be the case, so you can remove this check. If it's not a map, it's better to have it fail loudly instead of ignoring.

charts/linkerd2/templates/_addons.tpl
```diff
@@ -0,0 +1,25 @@
...
+  {{- if (hasKey $dupValues "global") -}}
```
alpeb, line 14: These checks are not necessary. If `global` is not a key, the following `unset` will just return the same map.

charts/linkerd2/templates/_addons.tpl
Pothulapati: makes sense!

charts/linkerd2/templates/_addons.tpl
Pothulapati: removed! :)

Pothulapati:
@alpeb That makes sense. So, moved that into `partials` :)

alpeb:
Thanks @Pothulapati 
From my testing, it appears this
```
{{- $dupValues := . -}}
```
doesn't make a copy but a reference instead, so the global values are lost afterwards. I tested trying to output some global in `linkerd-config-addons.yaml` right after `linkerd.addons.sanitize-config` is included.
You can give a try at the `deepMustCopy` function instead. 

OTOH, it appears it's not necessary to do `unset $dupValues "partials"` because `partials` is not included in `Values.tracing`.

Pothulapati:
@alpeb Yep, The plan was to remove `<add-on>.global` as we can access the same from .Values.global` directly? Is that also being removed? :thinking:

PR #4320, opened 2020-05-01
Move grafana templates into a separate sub-chart as a add-on

Pothulapati:
Third part of #3979, as per #3979 (comment) Feel free to look at it, to understand how these changes integrate.

This PR moves the grafana templates into a separate sub-chart.

**changes include:** 
- Move grafana templates into `/charts/add-ons/grafana`, and it as a dependency to `linkerd2` chart.
- Implement the `AddOn` interface, and it as a field to `Values`.
- Add Install and update unit tests, and update golden files


cli/cmd/install.go
```diff
@@ -839,7 +837,7 @@ func render(w io.Writer, values *l5dcharts.Values) error {
 			Name:      addOn.Name(),
 			Dir:       filepath.Join(addOnChartsPath, addOn.Name()),
 			Namespace: controlPlaneNamespace,
-			RawValues: append(rawValues, addOn.Values()...),
+			RawValues: append(addOn.Values(), rawValues...),
```
Pothulapati, line 840: This change was done as we were running into weird cases because of https://stackoverflow.com/questions/35920534/why-does-append-modify-passed-slice. As rawValues containing `addon.` fields directly causing conflicts with other add-ons with similar fields.

As `append.Values()` returns a copy of the slice, this should not update the underlying addOn.Values(), and cause any weird cases!

pkg/charts/linkerd2/values_test.go
```diff
@@ -132,6 +131,11 @@ func TestNewValues(t *testing.T) {
 			Image: "deislabs/smi-metrics:v0.2.1",
 			TLS:   &TLS{},
 		},
+		Grafana: map[string]interface{}{
+			"enabled": true,
+			"name":    "linkerd-grafana",
+			"image":   "gcr.io/linkerd-io/grafana",
+		},
```
zaharidichev: I think you can put these in constant?

Pothulapati:
This PR can be tested both using the CLI and Helm. 

**Testing through CLI:**

`--addon-config` flag is used to pass the configuration of add-ons. The same format and fields from `values.yaml` can be used/overwritten. Some examples of overwriting values can be found [here](https://github.com/linkerd/linkerd2/blob/88dafe6ad6141b19fa60c8d7ff35b5da0c930d0f/cli/cmd/upgrade_test.go#L1729)

*For upgrades:*

- First, the default values are read from `values.yaml`, which are overwritten by values from `linkerd-config-addons` (if present) and which can be further overwritten by `--addon-config` to add/remote/update config of add-ons during upgrades.
- One thing to note is that, when an enabled add-on is disabled during upgrade we don't delete the already created add-on resources but any configuration w.r.t add-on applied in the control-plane components will not be present. This can be solved by using `--purge` when applying.

**Testing through Helm:**

The good thing about Helm is that we don't have to manage anything regarding state of values, etc as Helm takes care of it. All the configuration including that of add-ons is stored and re-used during upgrades, where overwriting is also allowed. 

We still need the `linkerd-config-addons` cm even when the installation is done through Helm, is to have add-on specific health-checks as they need add-on configuration as shown [here](https://github.com/linkerd/linkerd2/pull/4321/files#diff-4d95af557bb1b4c9b3c9a50e7f9dbb50R42). Also, it makes sense to have consistent configuration across installation ways.

Pothulapati:
@zaharidichev Looks like there are multiple places where common fields like `enabled`, `name`, `cpu`, etc are used. I will submit a follow up PR on updating them at all those places! WDYT?

PR #4324, opened 2020-05-01
Have the Cloud, KinD and Release workflows trigger a CI report creation

alpeb:
Third step for #4176.

Dispatch a Github event when the Cloud, KinD and Release workflows finish, that will get caught in the [linkerd2-ci-metrics repo](https://github.com/linkerd/linkerd2-ci-metrics) to trigger the build of a CI report.

### DIFF
--- a/.github/workflows/cloud_integration.yml
+++ b/.github/workflows/cloud_integration.yml
@@ -118,7 +118,6 @@ jobs:
       run: |
         export TAG="$($HOME/.linkerd version --client --short)"
         go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
-
   trigger-reports:
     name: Trigger CI reports
     runs-on: ubuntu-18.04
--- a/.github/workflows/cloud_integration.yml
+++ b/.github/workflows/cloud_integration.yml
@@ -118,3 +118,10 @@ jobs:
       run: |
         export TAG="$($HOME/.linkerd version --client --short)"
         go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
+    - name: Trigger CI reports
+      # peter-evans/repository-dispatch@v1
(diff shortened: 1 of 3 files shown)


kleimkuhler:
@alpeb Should these jobs have `if: always()` so that they always run regardless if the previous jobs before them failed?

alpeb:
> @alpeb Should these jobs have `if: always()` so that they always run regardless if the previous jobs before them failed?

Good catch :+1: I've just pushed that.

alpeb:
@kleimkuhler and @adleong Since your last reviews I changed a bit things, so I'd appreciate another review :pray: 
Now the trigger is dispatched as a separate job to avoid have it run for each matrix element.

.github/workflows/cloud_integration.yml
```diff
@@ -118,3 +118,17 @@ jobs:
       run: |
         export TAG="$($HOME/.linkerd version --client --short)"
         go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
+
```
kleimkuhler, line 121: nit: We don't have any blank lines in between jobs in the CI files

.github/workflows/kind_integration.yml
```diff
@@ -162,3 +162,17 @@ jobs:
 
         init_test_run $HOME/.linkerd
         ${{ matrix.integration_test }}_integration_tests
+
```
kleimkuhler, line 165: Same as above

alpeb:
Thanks for the reviews @kleimkuhler and @adleong .
I decided it's better to discard this PR. The CI reports end up being very taxing on the Github API rate-limiting-wise, so I think I'll just have them be triggered once per day via a cronjob. And those reports aggregate data per month, so it's fine that each day's report doesn't contain the data for the current day.

PR #4339, opened 2020-05-06
Make gateway require TLS for incoming requests

zaharidichev:
This PR introduces a few changes with the aim of making the multicluster setup a bit more secure:

- The service mirror controller will now be injected
- The proxy on the gateway will require identity on the two ports that are exposed through the gateway (the incoming port and the probes port)
- The probe worker now uses the `l5d-require-id` when doing the HTTP get requests to the gateways to  inform the outbound proxy of the expected identity of the remote peer (the gateway)
- There are two sets of probes on two different ports. One is local and can work without identity (this is for Kubelet) and the other is remote and requires identity (for the prober that is part of service mirror).

This all should guarantee that non TLSed traffic coming from outside the cluster will be banned by the proxy that is injected on the gateway pod. I have tested that and it works. No port skipping is happening anywhere

This all depends on https://github.com/linkerd/linkerd2-proxy/pull/507

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -2,7 +2,7 @@
 env:
 {{ if .Values.global.proxy.requireIdentityOnInboundPorts -}}
 - name: LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_IDENTITY
-  value: {{.Values.global.proxy.requireIdentityOnInboundPorts}}
+  value: "{{.Values.global.proxy.requireIdentityOnInboundPorts}}"
 {{ end -}}  
 - name: LINKERD2_PROXY_LOG
   value: {{.Values.global.proxy.logLevel}}
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -1,5 +1,9 @@
 {{ define "partials.proxy" -}}
 env:
+{{ if .Values.global.proxy.requireIdentityOnInboundPorts -}}
+- name: LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_IDENTITY
+  value: {{.Values.global.proxy.requireIdentityOnInboundPorts}}
+{{ end -}}  
 - name: LINKERD2_PROXY_LOG
   value: {{.Values.global.proxy.logLevel}}
 - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
--- a/controller/cmd/service-mirror/probe_worker.go
+++ b/controller/cmd/service-mirror/probe_worker.go
@@ -13,12 +13,14 @@ import (
 )
 
 const httpGatewayTimeoutMillis = 50000
+const requireIDHeader = "l5d-require-id"
 
 type probeSpec struct {
 	ips             []string
 	path            string
 	port            uint32
 	periodInSeconds uint32
+	gatewayIdentity string
 }
 
 // ProbeWorker is responsible for monitoring gateways using a probe specification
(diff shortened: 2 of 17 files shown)


zaharidichev:
> One possible solution is to define two sets of checks on separate ports. One will be for the traffic that is coming from outside the cluster( the liveliness prober) and will go through the proxy, while the other ones will be for local only traffic and will be coming from Kubelet and bypassing the proxy (we can do that with ` --skip-inbound-ports`, right? )

This is what I did, and it works pretty well actually

olix0r:
@zaharidichev This sounds like a good workaround to unblock us, but I still have some reservations about this being a generally feasible approach... We also have to take care to ensure that this healthcheck port doesn't expose any possibility to route requests or access sensitive diagnostic information.

Can you help me think through the alternatives? Can we know the node's IP at pod-creation time and allow traffic that originates from the node?

zaharidichev:
> Can you help me think through the alternatives? Can we know the node's IP at pod-creation time and allow traffic that originates from the node?

@olix0r Possibly, but as far as I take it,  pods can move nodes. I also feel it is not the best way to do all that but could not think of a much better way. Alternatively we can rely on something that is NOT an http get request at all for checking health locally. We can configure the healthcheks to actually execute a shell command on the container and check whether the nginx process is running. Now that I think about it.. this sounds like a better approach.

olix0r:
> Possibly, but as far as I take it, pods can move nodes

Not true. A deployment/replicaset creates a pod, and it's pinned to its node/ip/etc throughout the lifetime of the pod. For the pod to "move" it terminates and a new pod is scheduled on another node. The injector hooks into the pod creation event.

zaharidichev:
I see, that makes sense. So I guess we can possibly do something like that. I will investigate. What about the possibility os just checking whether the process is alive through a shell command?

grampelberg:
> Can you help me think through the alternatives? Can we know the node's IP at pod-creation time and allow traffic that originates from the node?

We can use the downward API for this.

grampelberg:
> (we can do that with --skip-inbound-ports, right? )

I'm concerned about the implications of this. AFAIK services of type `nodePort` would have IP addresses coming from the host and then skip the proxy. Anyone have a good idea on the implications of just skipping with the downward API?

zaharidichev:
@grampelberg @olix0r I have updated the description to reflect the discussion from yesterday. I went along with hat we came up with yesterday and it works well.

controller/cmd/service-mirror/probe_worker.go
```diff
@@ -13,12 +13,14 @@ import (
 )
 
 const httpGatewayTimeoutMillis = 50000
+const requireIDHeader = "l5d-require-id"
```
kleimkuhler, line 16: nit: It may be worth defining this somewhere that both this and `controller/tap` can use.

charts/partials/templates/_proxy.tpl
```diff
@@ -1,5 +1,9 @@
 {{ define "partials.proxy" -}}
 env:
+{{ if .Values.global.proxy.requireIdentityOnInboundPorts -}}
+- name: LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_IDENTITY
+  value: {{.Values.global.proxy.requireIdentityOnInboundPorts}}
```
alpeb, line 5: One problem I did find afterwards is that this value requires to be wrapped in quotes, otherwise if a single port is used here it's interpreted as an int value and it breaks the injector because a string is expected

zaharidichev: yeah very good catch. Thanks!

PR #4346, opened 2020-05-07
Update `inject` to throw an error while injecting non-compliant pods

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/4214

This PR updates the way `inject` behaves while attempting to inject non-compliant pods, i.e, pods that have issues related to host network, UDP port, sidecars and `automountServiceAccountToken`

Now instead of displaying warning messages and writing an uninjected YAML to `stdout`, `inject` throws an error. For example :
```bash
$ cat cli/cmd/testdata/inject_emojivoto_deployment_hostNetwork_true.input.yml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: emojivoto
spec:
  replicas: 1
  selector:
    matchLabels:
      app: web-svc
  template:
    metadata:
      labels:
        app: web-svc
    spec:
      containers:
      - env:
        - name: WEB_PORT
          value: "80"
        - name: EMOJISVC_HOST
          value: emoji-svc.emojivoto:8080
        - name: VOTINGSVC_HOST
          value: voting-svc.emojivoto:8080
        - name: INDEX_BUNDLE
          value: dist/index_bundle.js
        image: buoyantio/emojivoto-web:v3
        name: web-svc
        ports:
        - containerPort: 9100
          hostPort: 9100
          name: http
      hostNetwork: true
---
$ cat cli/cmd/testdata/inject_emojivoto_deployment_hostNetwork_true.input.yml | bin/go-run cli inject -

Error transforming resources: failed to inject deployment/web - pods use host networking
```


mayankshah1607:
Similarly, when the input has multiple non-compliant pods or a mix of compliant and non-compliant, the errors are accumulated and displayed as a single message and nothing is injected

```bash
$ bad.yaml | bin/go-run cli inject -

Error transforming resources: failed to inject deployment/web - pods have a 3rd party proxy or initContainer already injected, 
failed to inject deployment/web - pods use host networking
```

Find `bad.yaml` [here](https://gist.github.com/mayankshah1607/3c6669a504c1b5e42421fcbd70d73eb5)

mayankshah1607:
Ping @kleimkuhler @alpeb

cli/cmd/inject.go
```diff
@@ -503,3 +521,23 @@ func overwriteRegistry(image, newRegistry string) string {
...
+	if report.Sidecar {
+		return true, errors.New(sidecarError)
+	}
+
+	if report.UDP {
+		return true, errors.New(udpError)
+	}
+
+	return false, nil
+}
```
alpeb, line 543: I think this can be simplified by just returning `error`; nil if not, `error` if yes

cli/cmd/inject.go
```diff
@@ -28,8 +28,14 @@ const (
...
 	udpDesc                          = "pod specs do not include UDP ports"
+	automountServiceAccountTokenDesc = "pods do not have automountServiceAccountToken set to \"false\""
 	slash                            = "/"
-	automountServiceAccountTokenDesc = "cannot enable mTLS when automountServiceAccountToken set to \"false\""
+
+	// for throwing inject error
+	hostNetworkError                  = "pods use host networking"
+	sidecarError                      = "pods have a 3rd party proxy or initContainer already injected"
+	udpError                          = "pod specs include UDP ports"
+	automountServiceAccountTokenError = "pods have automountServiceAccountToken set to \"false\""
```
alpeb, line 38: You might wanna reuse the strings from `report.go` (the `Reasons` map).

cli/cmd/inject_util.go
```diff
@@ -48,11 +50,13 @@ func transformInput(inputs []io.Reader, errWriter, outWriter io.Writer, rt resou
 }
 
 // processYAML takes an input stream of YAML, outputting injected/uninjected YAML to out.
-func processYAML(in io.Reader, out io.Writer, report io.Writer, rt resourceTransformer) error {
+func processYAML(in io.Reader, out io.Writer, report io.Writer, rt resourceTransformer) []string {
 	reader := yamlDecoder.NewYAMLReader(bufio.NewReaderSize(in, 4096))
 
 	reports := []inject.Report{}
 
+	errorMessages := []string{}
```
alpeb, line 58: Given this slice is filled with calls to `err.Error()`, I think it makes more sense to return a slice of `error`.

cli/cmd/inject.go
alpeb: Even better, a `[]error` would be cooler in order to accumulate all the errors from a single pod instead of short-circuiting on the first :wink:

alpeb:
Can you also please fix the referred issue number in the description?

cli/cmd/inject_util.go
mayankshah1607: @alpeb That would indeed make the code cleaner! However, we're using the returned slice of string error message [here](https://github.com/linkerd/linkerd2/blob/fe71f52c63786c85a4456de17633fd09f47cf6f2/cli/cmd/inject_util.go#L35) to concatenate the error strings and throw that as an error message.
If we were to return a slice of strings instead, we might need another helper function that first converts the slice of `error`s to slice of `string`s and then use it in `strings.Join` to produce a final error message. WDYT? :)

alpeb: Ok how about giving a try at using wrapped errors?
https://play.golang.org/p/nqNrM8qOvST
You could have `processYAML()` return just `error`, which would really be wrapped errors, which would not need to be joined on the calling side. I *think* this might result in a more go-ish and elegant way!

mayankshah1607:
@alpeb Thanks for the tip on wrapped errors - learnt something new! :smile: 
I've updated the code according to your comments. Had to force push as I ran into some problems while rebasing. But you might want to take a look at [this](https://github.com/linkerd/linkerd2/pull/4346/commits/721730257453fcb9f53aa33ee897e06ebee1487d) commit in particular to see the new changes w.r.t to the comments on error accumulation mechanism.
Thanks :smile:

cli/cmd/inject.go
```diff
@@ -188,7 +182,7 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 		return b, reports, err
 	}
 	if b, _ := report.Injectable(); !b {
-		if e, err := shouldInjectThrowError(report); e {
+		if err := throwInjectError(report); err.Error() != "" {
```
mayankshah1607, line 185: We don't use `err != nil` here because `throwInjectError` initializes a non-nil error (an empty string) for the sake of accumulation of error messages.

alpeb:
@mayankshah1607 Thanks for having humored me with the error wrapping suggestion. After having become more comfortable with that concept I realize now it's really meant for adding context to errors and not really for _accumulating_ errors. So I take back that suggestion.
I still think however that `processYAML()` and `throwInjectErrors()` should return an array of accumulated errors (`[]error`).  LMKWYT.

alpeb:
I think you should also reintroduce the logic to properly concatenate the error messages. Currently doing `linkerd inject` for example on a resources with both `hostNetwork` and `automountServiceAccountToken` yields something like this:
```
Error transforming resources: [failed to inject deployment/emoji - [automountServiceAccountToken set to "false" hostNetwork is enabled]]
```

cli/cmd/inject.go
```diff
@@ -522,3 +534,32 @@ func overwriteRegistry(image, newRegistry string) string {
...
+
+func throwInjectError(report *inject.Report) []error {
+
+	errs := []error{}
+
+	const (
+		disabledAutomountServiceAccountToken = "disabled_automount_service_account_token_account"
+		hostNetworkEnabled                   = "host_network_enabled"
+		sidecarExists                        = "sidecar_already_exists"
+		udpPortsEnabled                      = "udp_ports_enabled"
```
alpeb, line 546: Given this function is all about `inject.Report`, I think it's better to move it into `report.go` and make it a method of that struct. That will also avoid having to repeat these consts.

mayankshah1607: @alpeb Done! You can find that refactor in [this](https://github.com/linkerd/linkerd2/pull/4346/commits/69ac7fc0b0cdd6ed2688a8ac343d03ce6f23a3f2) commit. (Had to rebase against master because some of the tests didn't seem to run)

mayankshah1607:
Pingin @zaharidichev for one more approval! :smile:

cli/cmd/inject.go
```diff
@@ -304,6 +310,12 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output
 		output.Write([]byte(fmt.Sprintf("%s %s\n", okStatus, udpDesc)))
 	}
 
+	if len(automountServiceAccountTokenFalse) > 0 {
+		output.Write([]byte(fmt.Sprintf("%s \"automountServiceAccountToken: false\" detected in %s\n", warnStatus, strings.Join(automountServiceAccountTokenFalse, ", "))))
```
ihcsim, line 314: For my own understanding, `output` writes to `stderr`, right? But I can't seem to get the CLI to output this error. What am I missing?

alpeb:
> Also, it looks like we aren't handling the case when the `automountServiceAccountToken` property is defined in the service account YAML. E.g.,

Ah I didn't think of that... I guess the best we can do is to stop ignoring ServiceAccount manifests during injection and issue a warning whenever we encounter `automountServiceAccountToken: false`. Not outright failing because the SA wouldn't necessary belong to a workload being injected.

mayankshah1607:
@alpeb @ihcsim Just to confirm my understanding - we should only error out `automountServiceAccountToken: false` when the object Kind is not a `ServiceAccount`, correct?

ihcsim:
@mayankshah1607 That's correct. I created a separate issue (https://github.com/linkerd/linkerd2/issues/4651) to deal with `ServiceAccount` kind.

cli/cmd/inject.go
mayankshah1607: Ah, I think this might be redundant. The CLI will error out before this part of the code is executed.

ihcsim: Let's remove it then. Thanks.

mayankshah1607: Done :+1:

PR #4319, opened 2020-05-01
Make Grafana optional in linkerd-web

Pothulapati:
Second part of #3979, as per #3979 (comment) Feel free to look at it, to understand how these changes integrate.

This PR makes grafana link optional in `linkerd-web`. when `--grafana-addr` is not passed to `linkerd-web`, grafana links will not showup in `linkerd-web`

**changes include:** 
- Adds a new field `grafana` to `appParams` and is propagated into js to have the check in `MetricsTable.jsx`

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

PR #4348, opened 2020-05-07
Add changes for edge-20.5.1

kleimkuhler:
## edge-20.5.1

* CLI
  * Fixed all commands to use kubeconfig's default namespace if specified
    (thanks @Matei207!)
  * Added multicluster checks to the `linkerd check` command
  * Hid development flags in the `linkerd install` command for release builds
* Controller
  * Added ability to configure Prometheus Alertmanager (more documentation and
    changes to come for this!) (thanks naseemkullah!)
* Web UI
  * Fixed TrafficSplit detail page not loading
  * Added Jaeger links to the dashboard when the tracing addon is enabled
* Proxy
  * Modified internal buffering to avoid idling out services as a request
    arrives, fixing failures for requests that are sent exactly once per
    minute--such as Prometheus scrapes


kleimkuhler:
This includes changes for #4177

CHANGES.md
```diff
@@ -1,3 +1,21 @@
...
+
+* CLI
+  * Fixed all commands to use kubeconfig's default namespace if specified
+    (thanks @Matei207!)
+  * Added multicluster checks to the `linkerd check` command
+  * Hid development flags in the `linkerd install` command for release builds
+* Controller
+  * Added ability to configure Prometheus Alertmanager (more documentation and
+    changes to come for this!) (thanks naseemkullah!)
+* Web UI
```
Pothulapati, line 11: We could also add #4319 here! But not any user facing changes unless users manually remove the `garfana-addr` field in `linkerd-web` manifest

CHANGES.md
```diff
@@ -1,3 +1,21 @@
+## edge-20.5.1
+
+* CLI
+  * Fixed all commands to use kubeconfig's default namespace if specified
+    (thanks @Matei207!)
+  * Added multicluster checks to the `linkerd check` command
+  * Hid development flags in the `linkerd install` command for release builds
+* Controller
+  * Added ability to configure Prometheus Alertmanager (more documentation and
+    changes to come for this!) (thanks naseemkullah!)
```
alpeb, line 10: Missed the @ for the user handle

alpeb: Also, added ability to configure alerting and recording rules using external configmaps, and ability to add more flags to the prometheus command.

CHANGES.md
kleimkuhler: @Pothulapati I left this out for now since it is such a specific change that needs to occur for it to be considered user facing. Once all the PRs related to making Grafana a soft requirement have merged, it will be easier to make this a point in the release notes.

CHANGES.md
kleimkuhler: @alpeb None of this is really documented or explained much in the PR. I didn't clarify more here because even if we mentioned it, I don't think we can really point users to the PR description. I can include it if you think it should be, but does that make sense?

alpeb: The PR addressed four points: alertmanager, alert rules, recording rules and command flags, but you only mentioned the first one. I think we should mention them all, without pointing them to the PR. There are some commented-out examples in `values.yaml` for those new configs, but full docs will follow along with the BYOP project.

alpeb: So to give more hints to users, you could mention those are new config values just for helm, and so people should know they can look into the chart's readme or values.yaml for more info.

kleimkuhler: Okay I split this up into two separate bullets. Let me know your thoughts

PR #4350, opened 2020-05-07
remove admin server timeouts

adleong:
The Linkerd control plane components' admin servers have an idle connection timeout of 10 seconds.  This means that they will close connections which have been idle for 10 seconds.  These components are also configured with a 10 second period for liveness checks.  This introduces a race condition where connections will be idle for approximately 10 seconds between liveness checks and can idle out, potentially causing the next liveness check to fail.

We remove the idle timeout so that the connection stays alive.

### DIFF
--- a/pkg/admin/admin.go
+++ b/pkg/admin/admin.go
@@ -28,6 +28,7 @@ func StartServer(addr string) {
 		Handler:      h,
 		ReadTimeout:  10 * time.Second,
 		WriteTimeout: 10 * time.Second,
+		IdleTimeout:  1 * time.Minute,
 	}
 
 	log.Fatal(s.ListenAndServe())
--- a/pkg/admin/admin.go
+++ b/pkg/admin/admin.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"net/http/pprof"
 	"strings"
-	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


pkg/admin/admin.go
```diff
@@ -28,6 +28,7 @@ func StartServer(addr string) {
 		Handler:      h,
 		ReadTimeout:  10 * time.Second,
 		WriteTimeout: 10 * time.Second,
+		IdleTimeout:  1 * time.Minute,
```
olix0r, line 31: Can we just unset all of these timeouts? Why are they set?

olix0r: If we unset these, we could simplify this all to `http.ListenAndServe(addr, h)`

olix0r: Here's where it was introduced https://github.com/linkerd/linkerd2/commit/682b0274b5ff4535d1c4d088650cf6d13e508b31#diff-b5ab530c184ee6c6d6fe09ad488de24aL79 -- previously we used the simpler constructor. No explanation in commit message or code.

adleong: read and write timeouts seem like reasonable safety nets for these unary APIs.  If any HTTP message takes longer than 10 seconds to read or write, we probably want to abort it.

Removing idle timeout would be fine, I think, but it doesn't appear to be possible to have a read timeout but no idle timeout (????) since an unset idle timeout defaults to the read timeout.  But as long as this is long enough and we have the liveness checks to keep the connection open, this should amount to the same thing.

olix0r: As I mentioned in the issue, I think we could set a ReadHeaderTimeout if we really want to, but I would prefer just using the defaults. In all of the cases we care about (prometheus, healthchecking), the clients will enforce some timeout.

If we do want to keep timeouts here, we need to document why we've chosen these values and what problem they solve. I don't want to cargo-cult arbitrary configuration (that has caused problems).

I'm specifically -1 on setting an IdleTimeout, though. We can clients choose how they want to manage connections.

PR #4359, opened 2020-05-08
Threadsafe buffering listener

### LINKED ISSUE #4336: Race condition in endpoint watcher tests
The endpoint watcher tests use an in-memory bufferingEndpointListener to hold endpoint updates and compares these to expected values.  However, since these updates are written from the client-go informer thread and read from the main test thread, a data race can sometimes occur, resulting in a flaky test.  

bufferingEndpointListener should be made thread-safe to protect against these data races.

zaharidichev:
Add some thread safety to endpoints watching tests

Fix #4336

### DIFF
--- a/controller/api/destination/watcher/endpoints_watcher_test.go
+++ b/controller/api/destination/watcher/endpoints_watcher_test.go
@@ -18,7 +18,7 @@ type bufferingEndpointListener struct {
 	added             []string
 	removed           []string
 	noEndpointsCalled bool
-	noEndpointsExists bool
+	noEndpointsExist  bool
 	sync.Mutex
 }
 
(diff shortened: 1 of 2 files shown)


controller/api/destination/watcher/endpoints_watcher_test.go
```diff
@@ -38,46 +41,94 @@ func addressString(address Address) string {
 	return addressString
 }
 
+func (bel *bufferingEndpointListener) Added() []string {
+	bel.Lock()
+	defer bel.Unlock()
+	return bel.added
```
adleong, line 47: Aren't slices passed by reference in go?  if so, I think this slice is still shared and can still encounter a data race.

zaharidichev: @adleong I think the last commit should fix that.

Issue #3406, opened 2019-09-09
Unbundle Prometheus and Grafana


andrew-waters:
### What problem are you trying to solve?

linkerd2 deployments come with Grafana and Prometheus bundled into the deployment. This is great for single cluster setups and PoC's but it comes at the cost of resources when an existing cluster already has prometheus running. The documentation around [exporting metrics](https://linkerd.io/2/tasks/exporting-metrics/) is useful and for the specific use case of running [thanos](https://thanos.io/), the service monitor is the preferable route (the purpose of thanos is to avoid federation, which has legitimate issues).

Below is an approximate topology of a multi cluster setup running thanos (as close as possible to the [linkerd2 supplied one](https://linkerd.io/2/reference/architecture/#prometheus)).

![linkerd (1)](https://user-images.githubusercontent.com/501976/64529654-45262080-d303-11e9-9678-b2e6efc0e6c1.png)

As you can see it should in theory be possible to remove prometheus and grafana from the linkerd deployment and maintain metric collection.

However, removing these cause errors with `linkerd check` and `linkerd dashboard` (they appear to check for the existence.

### How should the problem be solved?

I'm not entirely sure of the _scope_ of changes that are required, but adding a flag to the commands that would otherwise fail may be a useful starting point for discussion:

```bash
linkerd check --exclude=prometheus,grafana
linkerd install --exclude=prometheus,grafana
linkerd dashboard
```

Note that `linkerd dashboard` fails when prometheus isn't available which suggests there may be some core logic that is shared validation logic between the cli.

### Any alternatives you've considered?

The only alternatives are to:

 - ignore this issue and force users to install prometheus and grafana
 - avoid using the CLI

### How would users interact with this feature?

Referenced above in bash scripts.

---

**Edit:** updated diagram to have a white background for legibility

grampelberg:
The bundled prometheus is pretty tuned just for Linkerd. We're doing enough things there and with the CLI/dashboard that a central system would likely crash, especially if retention is over 6 hours. Assuming that cost can be kept in check, it feels like an important piece of the puzzle.

Grafana, on the other hand, is 100% optional. There's some ongoing work around configuration that'll make it optional and not break when it isn't installed (dashboard links for example).

andrew-waters:
That's interesting. Is linkerd using prometheus as it's own storage? The most common (recommended) tsdb lifecycle in prometheus is 2 hours so that _shouldn't_ cause issues.

[...]

PR #4362, opened 2020-05-08
Move Prometheus as an Add-On

### LINKED ISSUE #3590: Make Prometheus Pluggable
## Feature Request

The prometheus instance that comes with Linkerd is nice but for large clusters this can become a bottleneck. By allowing the prometheus instance to be configurable, we can give users the ability to re-use their existing monitoring infra with Linkerd and also use longer metric retention.

There have also been projects like [cortex](https://github.com/cortexproject/cortex), [thanos](https://github.com/thanos-io/thanos) which have been getting some traction recently to make prometheus scalabale. This would also allow users to use managed monitoring offerings like [Grafana Cloud](https://grafana.com/products/cloud/#pricing), etc

The following components use prometheus for queries right now:
- `public-api` : To serve metrics for CLI and dashboard.
- `heart-beat`
Both the components take the `--prometheus-url` flag during startup which points to `http://linkerd-prometheus.{{.Namespace}}.svc.{{.ClusterDomain}}:9090` right now.
- `Grafana`: This can be configured, by tweaking the data-sources field of `linkerd-grafana-configmap`

Users will then have to take the prometheus config and apply it to their monitoring infra i.e prometheus, cortex, etc. 

For prom instances running outside their k8s cluster i.e managed, etc,  A prometheus instace (with the linkerd-prometheus config) with [remote-write](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write) configured to the outside instance. Assuming that the outside instace is scalable, by having the linkerd components still use the outside instance can help in faster metric retrieval and longer retention.

## Solution

Having a global config to set the `prometheus-url` through helm and linkerd install command should be enough. In that case, the default prometheus installation should be disabled.

`linkerd install --prometheus-url `

## Out of Scope

- How should the outside prometheus instance be installed and configured?

related #3406 

@grampelberg @ihcsim @siggy

Pothulapati:
Fixes #3590, #3406
Implements first part of https://github.com/linkerd/rfc/pull/16, https://github.com/linkerd/linkerd2/issues/4375

This PR moves Prometheus as a add-on, thus making it optional but enabled by default. The main benefit of this work is to make `linkerd-prometheus` more configurable, and allow it to have its own life-cycle for upgrades, configuration, etc.

This work will be followed by documentation that help users configure existing Prometheus to work with Linkerd.

**Changes Include:**
- moving prometheus manifests into a separate chart at `charts/add-ons/prometheus`, and adding it as a dependency to `linkerd2`
- implement the `addOn` interface to support the same with CLI.
- include configuration in `linkerd-config-addons`


charts/add-ons/prometheus/templates/prometheus.yaml
```diff
@@ -34,7 +29,6 @@ data:
       static_configs:
       - targets: ['localhost:9090']
 
-    {{ if .Values.grafana.enabled -}}
```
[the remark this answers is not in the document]
Pothulapati: @alpeb As `grafana` is a separate add-on, and those fields are only propagated into their sub-chart.  We can't get this value in Prometheus add-on. 

I don't think prometheus would have any problem regarding the scrape-config even when there is no grafana, but I will test this out!

For other fields that would need to be accessed across multiple add-ons, we can use the `global` and have that in `linkerd-config-addons`!

Pothulapati:
**For Testing:**
- First, Apply recording rules as 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: linkerd-rules
  namespace: linkerd
data:
  alerting_rules.yml: |
    groups:
    - name: Hardware alerts
      rules:
      - alert: Node down
        expr: up{job="node_exporter"} == 0
        for: 3m
        labels:
        severity: warning
        annotations:
        title: Node {{ $labels.instance }} is down
        description: Failed to scrape {{ $labels.job }} on {{ $labels.instance }} for more than 3 minutes. Node seems down.
  recording_rules.yml: |
    groups:
    - name: example
      rules:
      - record: job:http_inprogress_requests:sum
        expr: sum(http_inprogress_requests) by (job)
```
- Let `config.yaml` be
```yaml
prometheus:
  args:
    log.format: json
  globalConfig:
    evaluation_interval: 2m
    external_labels:
      cluster: cluster-1
  scrapeConfigs:
  - job_name: 'kubernetes-nodes'
    scheme: https
    tls_config:
      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
    bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
    kubernetes_sd_configs:
    - role: node
    relabel_configs:
    - action: labelmap
      regex: __meta_kubernetes_node_label_(.+)
  alertManagers:
  - scheme: http
    static_configs:
    - targets:
      - "alertmanager.linkerd.svc:9093"
  alertRelabelConfigs:
  - action: labeldrop
    regex: prometheus_replica
  ruleConfigMapMounts:
  - name: alerting-rules
    subPath: alerting_rules.yml
    configMap: linkerd-rules
  - name: recording-rules
    subPath: recording_rules.yml
    configMap: linkerd-rules
```
- Apply it as
`linkerd install --addon-config config.yaml | k apply -f -`
- Prometheus should work, without any failures. The same configuration should be present in the Prometheus UI  which can be seen by port-forwarding!

The above configuration should also work with helm

pkg/charts/linkerd2/addons.go
```diff
@@ -35,5 +27,21 @@ func ParseAddOnValues(values *Values) ([]AddOn, error) {
 		}
 	}
 
+	if values.Prometheus != nil {
+		if enabled, ok := values.Prometheus["enabled"].(bool); !ok {
+			return nil, fmt.Errorf("invalid value for 'prometheus.enabled' (should be boolean): %s", values.Prometheus["enabled"])
```
zaharidichev, line 32: If your type is an interface here, and it fails to cast to bool, wouldn't  you want to format it with %v instead?

pkg/charts/linkerd2/addons.go
```diff
@@ -35,5 +27,21 @@ func ParseAddOnValues(values *Values) ([]AddOn, error) {
...
+		} else if enabled {
+			addOns = append(addOns, values.Prometheus)
+		}
+	}
+
+	if values.Tracing != nil {
+		if enabled, ok := values.Tracing["enabled"].(bool); !ok {
+			return nil, fmt.Errorf("invalid value for 'tracing.enabled' (should be boolean): %s", values.Tracing["enabled"])
```
zaharidichev, line 40: same here

cli/cmd/install.go
```diff
@@ -183,7 +181,7 @@ func newInstallOptionsWithDefaults() (*installOptions, error) {
 		controlPlaneVersion:         version.Version,
 		controllerReplicas:          defaults.ControllerReplicas,
 		controllerLogLevel:          defaults.ControllerLogLevel,
-		prometheusImage:             defaults.PrometheusImage,
+		prometheusImage:             defaults.Prometheus["image"].(string),
```
zaharidichev, line 184: can this cast fail ?

cli/cmd/install_addon_test.go
```diff
@@ -20,11 +21,20 @@ func TestAddOnRender(t *testing.T) {
 	withTracingAddonValues.Tracing["enabled"] = true
 	addFakeTLSSecrets(withTracingAddonValues)
 
+	withPrometheusAddOnOverwrite, err := testInstallOptions()
+	if err != nil {
+		t.Fatalf("Unexpected error: %v\n", err)
```
zaharidichev, line 26: would formatting with %s work as well ?

adleong:
@Pothulapati are you still working on this?  looks like there are a lot of merge conflicts that would need to get resolved

Pothulapati:
@adleong Yep, Started getting back on this from today! :)

charts/add-ons/prometheus/values.yaml
```diff
@@ -0,0 +1,54 @@
+image: prom/prometheus:v2.15.2
+args:
+  storage.tsdb.path: /data
+  storage.tsdb.retention.time: 6h
+  config.file: /etc/prometheus/prometheus.yml
+  log.level: debug
```
alpeb, line 6: I think `info` should be a better default

charts/add-ons/prometheus/values.yaml
```diff
@@ -0,0 +1,54 @@
...
+globalConfig:
+  scrape_interval: 10s
+  scrape_timeout: 10s
+  evaluation_interval: 10s
+scrapeConfigs:
```
alpeb, line 11: What's the reasoning behind putting some scrape configs here, and others in `prometheus.yaml`?

alpeb: Oh the ones that depend on other values are placed in the template... You could place them in the original place in the main `values.yaml` and instead of using `{{.Values.global.namespace}}` declare a yaml anchor for `namespace` and use `*namespace`.
As for `{{default .Values.global.proxyContainerName "linkerd-proxy" .Values.global.proxyContainerName}}` that custom value is such an obscure setting that we might as well get rid of it. It's not even mentioned in the chart's README.md nor in `values.yaml`.

Pothulapati:
@alpeb As you know we can't really put the defaults into `charts/linkerd2/values.yaml` as that would mean that the defaults would go into the `linkerd-config-addons`, causing upgrades to not work as expected.

Putting these values in templates might lead into conflicts when a user tries to override the same field that was set with a default value, as we do not have a way to merge them (with values in templates) but we just append them and prometheus fails starting up telling that the same value has been set twice and has conflicts.

By having these defaults in `/<add-on>/values.yaml`, we just merge them with the overriden values thus the same field will be overriden but not appended, causing it to work as expected.

charts/add-ons/prometheus/values.yaml
Pothulapati: Did not quite understand your comment about anchoring namespace in `values.yaml`
But that would mean we would have to hardcode Linkerd and the global.namespace is not propagated till here right? when users configure their own namespace using `global.namespace`.

Pothulapati: Just going to move the scrape-configs into templates as they are not conflicting, only other fields like args are i.e when they are overwritten.

Pothulapati:
**To test this PR:**
Support for most configuration like args, scrape_configs, globalConfigs, etc have been moved under `prometheus` field in `values.yaml` and they can all be configured now. Once configured they are stored under `linkerd-config-addons` and used with upgrades. 

An example full configuration file is available [here](https://gist.github.com/Pothulapati/f7c5c589b1d563546374d370d4082e3b), which can be applied by running
```bash
# Install recording and alerting rules which are used by the prometheus instance
k apply -f https://gist.githubusercontent.com/Pothulapati/00194de1dd0d819c4b853df96a8b3dfe/raw/881a5f5cd8dee4601c1ded77393ab0cbe4c459c5/linkerd-rules.yaml
# Install Linkerd
./target/cli/linux/linkerd install --addon-config ./config.yaml | k apply -f -
``` 

Once applied the configuration can be checked from the prometheus UI by port-forwarding the service (where there are options to see and check the configuration of the isntance)

Pothulapati:
This PR also has the following additional changes:
- `prometheusImage` and `prometheusLogLevel` flags have been removed as when these flags are set we would essentially want to set `prometheus.` fields which in the default case are not present as those fields are commented out in `values.yaml` and defaults are in `prometheus/values.yaml` and would cause additional complexity to create an append. This is not worth in my opinion as users can now set these values in `prometheus.image` and `promtehus.args.log.level` through the `addon-config` file.

charts/linkerd2/values.yaml
```diff
@@ -92,7 +93,6 @@ webhookFailurePolicy: Ignore
 
 # controller configuration
 controllerImage: gcr.io/linkerd-io/controller
-controllerLogLevel: &controller_log_level info
```
Pothulapati, line 95: @alpeb Though we moved this flag to be global like we should, but we don't actually have a way to use this directly in `prometheus/values.yaml` as we can't template values there. and putting that in the templates directly means that we don't have a way to merge them and appending might cause prometheus to complain just like it did with other values in configmap (I will try this out today)

alpeb:
Thanks @Pothulapati for the extended explanations :slightly_smiling_face: 

According to my understanding on how this should work, I suggest we remove the commented properties in `charts/linkerd2/values.yaml` under the prometheus section, since that's not the place to override them (also avoiding the extra work for us to keep them in sync with `charts/add-ons/prometheus/values.yaml`). And instead put a comment like this one before that section:
```
# This add-on's default property values are declared in `charts/add-ons/prometheus/values.yaml`.
# If installing/upgrading with Helm, you can override them through the usual `--set` or `-f` flags.
# Do not override them in this file!
# If installing/upgrading with linkerd's CLI, use the `--addon-config` flag.
```
And a comment like this in the header of `charts/add-ons/prometheus/values.yaml`:
```
# These are the default properties for this add-on.
# If installing/upgrading with Helm, you can override them through the usual `--set` or `-f` flags.
# Do not override them in this file!
# If installing/upgrading with linkerd's CLI, use the `--addon-config` flag.
```

Does that makes sense?

Pothulapati:
@alpeb I was thinking that with Helm, when users want to overwrite they would use `linkerd2/values.yaml` and not add-on specific `values.yaml`. As the configuration into `linkerd-config-addons` is taken from `linkerd2//values.yaml` 

This way they still get
- Checks, etc that depend on the configuration present in `linkerd-config-addons`
- Upgrades, etc would still work as its still `values.yaml` and helm would re-use that and hence provide the same helm upgrade semantics with that of

alpeb:
Ok cool. So the only thing we have to say is to not touch the `charts/add-ons/prometheus/values.yaml` file?

Pothulapati:
@alpeb Yep, Updated `/prometheus/values.yaml` accordingly.

alpeb:
@Pothulapati would you check the behavior of `linkerd upgrade --addon-overwrite`? I installed using a custom prom config passed with `--addon-config` but then `linkerd upgrade --addon-overwrite` isn't resetting to the default values...

PR #4363, opened 2020-05-08
Refactor CNI integration tests to use annotations functions

alpeb:
Followup to #4341

Replaced all the `t.Error`/`t.Fatal` calls in the integration tests with the
new functions defined in `testutil/annotations.go` as described in #4292,
in order for the errors to produce Github annotations.

This piece takes care of the CNI integration test suite.

This also enables the annotations for these and the general integration
tests, by setting the `GH_ANNOTATIONS` environment variable in the
workflows whose flakiness we're interested on catching: Kind
integration, Cloud integration and Release.

Re #4176

### DIFF
--- a/.github/workflows/cloud_integration.yml
+++ b/.github/workflows/cloud_integration.yml
@@ -6,6 +6,8 @@ on:
     - '**/*.md'
     branches:
     - master
+env:
+  GH_ANNOTATION: true
 jobs:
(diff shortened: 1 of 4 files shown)


alpeb:
@hemakl Would you like to take a quick look? This change will allow us to surface errors in the CNI integration test through Github's API, and generate reports about failures.

PR #4367, opened 2020-05-09
Add i18n to dashboard

scottcarol:
This PR adds i18n to certain views in the Linkerd dashboard, adds `es.json` with Spanish translations, and creates a process for contributing translations. 

The default language remains English and if the dashboard user's language is not included in the translations they will see English.

_Note:_ This PR does not translate **all** of the text in the dashboard - it translates table column names and text seen in the default view of the dashboard, which is about 50% of text. Once the approach is validated I will make a second PR that translates the rest of.

#### How to test the functionality:

If you are viewing the dashboard in English, nothing will change for you. To test in `es`, Firefox is recommended because it makes it easy to switch your default language.

Firefox:
* Load your [Firefox Preferences Page](about:preferences#general).
* Click on “Choose your preferred language for displaying pages.” 
* Change the default language to the language you want to test.

Testing in Chrome and Safari is more time-consuming. Both automatically display in the default system language for your computer. To change this, you must change the default setting for your machine. 

Mac: 
* Open System Preferences, go to “Language and Region” and pick a new Primary language. 
* You will be prompted to restart your machine. Be sure to remember what the “Language and Region” icon looks like so you can change it back again.

#### How to review the code:
Although this looks like a lot of text it's really just switching raw text to calling the function `t`.
```diff
         <Typography>
-          {content}
+          {t(content)}
         </Typography>
```

#### K8s context
K8s already has a ton of translations available - the [docs](https://kubernetes.io/docs/home/) are available in many different languages (dropdown on top right). Most relevant to us are the [i18n files for the K8s dashboard](https://github.com/kubernetes/dashboard/tree/master/i18n). The dashboard was built with Angular so we can’t copy their exact implementation but we can learn from the translations.

#### How the code works: 
We’re using [i18next](https://yarnpkg.com/package/i18next), a popular i18n library that includes a HOC (higher order component) called `withTranslation`. `withTranslation` wraps the React component and provides a function called `t` that takes a key and returns the value for the user’s language. If there’s no value for the user’s language, it will return English.

#### How it will change development in the Linkerd dashboard:
Text will no longer be directly included in the component. Instead the `jsx` file will include a `{t(‘keyName’)}` referencing a key/value pairing in a JSON file. 

#### How to add/edit translations in the future:
If you want to add a language, copy the `en.json` file into a new file named for the new language code. Then replace the value with the value in the new language, leaving the key intact. This approach creates easily reviewable PRs as the only file change would be to add a JSON file, and the other components remain untouched.

#### History:
This PR builds on research and translations that were done in #3757. 🙏


web/app/js/components/EmptyCard.jsx
```diff
@@ -12,12 +13,12 @@ const styles = () => ({
...
 });
 
-const EmptyCard = ({ content, classes }) => {
+const EmptyCard = ({ content, classes, t }) => {
   return (
     <Card className={classes.card} elevation={3}>
       <CardContent>
         <Typography>
-          {content}
+          {t(content)}
```
alpeb, line 21: IMO I think it's better to always try calling `t` on string literals when possible, so in this case below in `EmtpyCard.defaultProps` and in `EmptyCard`'s calling sites, so as to allow automated tools to pull all the strings and complete the json translation files.

web/app/js/components/EdgesTable.jsx
```diff
@@ -82,21 +83,21 @@ const edgesColumnDefinitions = (PrefixedLink, namespace, type, classes) => {
...
+      title = `${title} (${t('Identity')}: ${identity})`;
     }
   }
   return title;
 };
 
-const EdgesTable = ({ edges, api, namespace, type, classes }) => {
-  const edgesColumns = edgesColumnDefinitions(api.PrefixedLink, namespace, type, classes);
-  const edgesTableTitle = generateEdgesTableTitle(edges);
+const EdgesTable = ({ edges, api, namespace, type, classes, t }) => {
```
grampelberg, line 98: As this is a functional component right now, I'd recommend using the hooks instead (`useTranslation`). That way, you can use it in the sub-functions as well such as `generateEdgesTableTitle` and `edgesColumnDefinitions` without passing `t` everywhere. It has the added benefit of reducing the component wrapping!

web/app/js/components/EdgesTable.jsx
```diff
@@ -82,21 +83,21 @@ const edgesColumnDefinitions = (PrefixedLink, namespace, type, classes) => {
...
 
-const generateEdgesTableTitle = edges => {
+const generateEdgesTableTitle = (edges, t) => {
   let title = 'Edges';
   if (edges.length > 0) {
     let identity = edges[0].direction === 'INBOUND' ? edges[0].serverId : edges[0].clientId;
     if (identity) {
       identity = `${identity.split('.')[0]}.${identity.split('.')[1]}`;
-      title = `${title} (Identity: ${identity})`;
+      title = `${title} (${t('Identity')}: ${identity})`;
```
grampelberg, line 92: As this can change depending on the language, we'll want to use [interpolation](https://www.i18next.com/translation-function/interpolation) here. That'll allow folks to l10n with the order of their own language!

web/app/js/components/EdgesTable.jsx
```diff
@@ -82,21 +83,21 @@ const edgesColumnDefinitions = (PrefixedLink, namespace, type, classes) => {
   ];
 };
 
-const generateEdgesTableTitle = edges => {
+const generateEdgesTableTitle = (edges, t) => {
   let title = 'Edges';
```
grampelberg, line 87: Shouldn't this be translated as well?

web/app/js/components/EdgesTable.jsx
```diff
@@ -117,10 +118,11 @@ EdgesTable.propTypes = {
...
   type: PropTypes.string.isRequired,
+  t: PropTypes.func.isRequired,
 };
 
 EdgesTable.defaultProps = {
   edges: [],
 };
 
-export default withStyles(styles)(EdgesTable);
+export default withTranslation(['Shared'])(withStyles(styles)(EdgesTable));
```
grampelberg, line 128: I believe the standard here is `common` (at least that's what I've run into the most).

web/app/js/components/EmptyCard.jsx
```diff
@@ -12,12 +13,12 @@ const styles = () => ({
   },
 });
 
-const EmptyCard = ({ content, classes }) => {
+const EmptyCard = ({ content, classes, t }) => {
```
grampelberg, line 16: I'd recommend using hooks for functional components.

web/app/js/components/EmptyCard.jsx
grampelberg: I can't actually find any examples in the codebase where we use an empty card with non-default text? Feels like we could just simplify this component up until the extra functionality is needed. 

That said, I'm +1, the translated content should be passed down pre-translated as part of the prop.

web/app/js/components/MetricsTable.jsx
```diff
@@ -250,12 +254,12 @@ const preprocessMetrics = metrics => {
   return tableData;
 };
 
-const MetricsTable = ({ metrics, resource, showNamespaceColumn, showName, title, api, isTcpTable, selectedNamespace, grafana, jaeger }) => {
+const MetricsTable = ({ metrics, resource, showNamespaceColumn, showName, title, api, isTcpTable, selectedNamespace, grafana, jaeger, t }) => {
```
grampelberg, line 257: Another great opportunity for hooks to reduce prop passing!

web/app/js/components/Navigation.jsx
```diff
@@ -413,7 +414,7 @@ class NavigationBase extends React.Component {
         className={classes.navMenuItem}
         selected={path === normalizedPath}>
         <ListItemIcon>{icon}</ListItemIcon>
-        <ListItemText primary={title} />
+        <ListItemText primary={t(title)} />
```
grampelberg, line 417: I think we'll want to pass this in pre-translated to aide in auto-extraction.

web/app/js/components/Navigation.jsx
```diff
@@ -402,7 +403,7 @@ class NavigationBase extends React.Component {
   }
 
   menuItem(path, title, icon, onClick) {
```
grampelberg, line 405: Why isn't this a component? It seems like a perfect opportunity to pull this code out and simplify things.

web/app/js/components/QueryToCliCmd.jsx
```diff
@@ -30,13 +31,13 @@ class QueryToCliCmd extends React.Component {
...
   render = () => {
-    const { cmdName, query, resource, controllerNamespace } = this.props;
+    const { cmdName, query, resource, controllerNamespace, t } = this.props;
 
     return (
       _isEmpty(resource) ? null :
       <CardContent>
         <Typography variant="caption" gutterBottom>
-          Current {_startCase(cmdName)} query
+          {t('Current')} {_startCase(cmdName)} {t('query')}
```
grampelberg, line 40: Another opportunity for interpolation.

web/app/js/components/TapEventTable.jsx
```diff
@@ -198,9 +199,9 @@ const expandedRowRender = (d, expandedWrapStyle) => {
   );
 };
 
-const TapEventTable = ({ tableRows, resource, api }) => {
+const TapEventTable = ({ tableRows, resource, api, t }) => {
```
grampelberg, line 202: More opportunity for hooks!

scottcarol:
@alpeb @grampelberg thanks for the reviews! will rework and aim to have this updated by tomorrow EOD.

scottcarol:
Update - I let @grampelberg know that this is still in the works but it's been delayed by other work - keeping this PR open in the meantime.

ihcsim:
Closing this PR in favor of a different approach, per conversation with @scottcarol and @grampelberg.

PR #4368, opened 2020-05-11
Detect windows nodes and skip injection of proxies

### LINKED ISSUE #2882: Windows node scheduling

As of today, the proxy does not run on windows nodes. Unfortunately, injection will still happen and the pod will just not run.

- Update `inject` to be aware of windows pods and skip injection on them.
- Update `check` to warn if there are windows nodes in the cluster.

### Open questions

- Are there any ways we can surface that a pod is failing because of the proxy in a clear way?
- Should we recommend taints in the documentation?

vaniisgh:
Add a function to detect windows nodes and skip deploying proxies on them

begins implementation to fix #2882 

I have tried to find a solution to this issue but since this is my first time contributing to linkerd as well as my first contribution in golang, I was anxious to get some feedback before I went any further. I am not sure this is the solution you were looking for ... Please forgive me for the mistakes I have probably made and provide as much criticism as you feel necessary :)  
I shall try and implement the suggestions and come up with a final pull request. 

**Implementation**
A field added to the report given to the inject function, indicating when the node has a windows image, and that it should skip the injection of proxy.

The function checks for a windows node image or checks for applied taints with the field 'os'. If neither is found it concludes there are no windows node images. 

Both these are optional fields though, so I think it would be helpful to add this to the documentation if this proposed solution is adopted :)

*TO-DO*
- add documentation on taints
- provide a more elegant warning 

Signed-off-by: Vani Singh <vanisingh@live.co.uk>

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -26,7 +26,7 @@ const (
 	hostNetworkDesc                  = "pods do not use host networking"
 	sidecarDesc                      = "pods do not have a 3rd party proxy or initContainer already injected"
 	injectDisabledDesc               = "pods are not annotated to disable injection"
-	uncompatbleNodeOSDesc				 = "pods do not run on a windows node"
+	unsupportedOSDesc    			 = "pods do not run on a windows node"
 	unsupportedDesc                  = "at least one resource injected"
 	udpDesc                          = "pod specs do not include UDP ports"
 	slash                            = "/"
@@ -219,7 +219,7 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output 
 	sidecar := []string{}
 	udp := []string{}
 	injectDisabled := []string{}
-	uncompatbleNodeOS := []string{}
+	unsupportedOS := []string{}
 	warningsPrinted := verbose
 
 	for _, r := range reports {
@@ -248,7 +248,7 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output 
 		}
 
 		if r.NodeOS {
-			uncompatbleNodeOS = append(uncompatbleNodeOS, r.ResName())
+			unsupportedOS = append(unsupportedOS, r.ResName())
 			warningsPrinted = true
 		}
 
@@ -280,10 +280,10 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output 
 		output.Write([]byte(fmt.Sprintf("%s %s\n", okStatus, injectDisabledDesc)))
 	}
 
-	if len(uncompatbleNodeOS) > 0 {
-		output.Write([]byte(fmt.Sprintf("%sfollowing node image(s) is Windows %s\n", warnStatus, strings.Join(uncompatbleNodeOS, ", "))))
+	if len(unsupportedOS) > 0 {
+		output.Write([]byte(fmt.Sprintf("%sWindows is currently unsupported %s\n", warnStatus, strings.Join(unsupportedOS, ", "))))
 	} else if verbose {
-		output.Write([]byte(fmt.Sprintf("%s %s\n", okStatus, uncompatbleNodeOSDesc)))
+		output.Write([]byte(fmt.Sprintf("%s %s\n", okStatus, unsupportedOSDesc)))
 	}
 
 	if len(injected) == 0 {
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -26,6 +26,7 @@ const (
 	hostNetworkDesc                  = "pods do not use host networking"
 	sidecarDesc                      = "pods do not have a 3rd party proxy or initContainer already injected"
 	injectDisabledDesc               = "pods are not annotated to disable injection"
+	uncompatbleNodeOSDesc				 = "pods do not run on a windows node"
 	unsupportedDesc                  = "at least one resource injected"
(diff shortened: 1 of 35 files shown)


cli/cmd/inject.go
```diff
@@ -26,6 +26,7 @@ const (
 	hostNetworkDesc                  = "pods do not use host networking"
 	sidecarDesc                      = "pods do not have a 3rd party proxy or initContainer already injected"
 	injectDisabledDesc               = "pods are not annotated to disable injection"
+	uncompatbleNodeOSDesc				 = "pods do not run on a windows node"
```
grampelberg, line 29: `unsupportedOS = "pods cannot run on windows"`

cli/cmd/inject.go
```diff
@@ -272,6 +280,12 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output
 		output.Write([]byte(fmt.Sprintf("%s %s\n", okStatus, injectDisabledDesc)))
 	}
 
+	if len(uncompatbleNodeOS) > 0 {
+		output.Write([]byte(fmt.Sprintf("%sfollowing node image(s) is Windows %s\n", warnStatus, strings.Join(uncompatbleNodeOS, ", "))))
```
grampelberg, line 284: `Windows is currently unsupported`

pkg/inject/report.go
```diff
@@ -20,6 +21,7 @@ const (
 	invalidInjectAnnotationWorkload      = "invalid_inject_annotation_at_workload"
 	invalidInjectAnnotationNamespace     = "invalid_inject_annotation_at_ns"
 	disabledAutomountServiceAccountToken = "disabled_automount_service_account_token_account"
+	UncompatbleNodeOS					 = "node_os_is_windows"
```
mayankshah1607, line 24: Any reason why you've made this field public?

vaniisgh: honestly, no. Do you suggest I should leave it out?

vaniisgh:
@mayankshah1607 thank you so much! still working on adding some tests :) will certainly look into the test fixtures too!

pkg/inject/report.go
mayankshah1607: I think you might want to leave it as `unsupportedOS`. I don't see why that field should be exported.

pkg/inject/report.go
```diff
@@ -142,6 +151,24 @@ func checkUDPPorts(t *v1.PodSpec) bool {
 	return false
 }
 
+func checkNodeOS(t *v1.PodSpec) bool {
+	// Check any pods run a windows image, which is currently unsupported
+	//check the NodeSelector 
+	osimage := t.NodeSelector["kubernetes.io/os"]
+	if strings.Contains(strings.ToLower(osimage),"windows") == true {
```
mayankshah1607, line 158: There's a linting error here. You may wanna leave this as 
```go
if strings.Contains(strings.ToLower(osimage),"windows") {
```

pkg/inject/report.go
vaniisgh: ya, I see what you mean, it's unnecessary to export it. Think it will be limited to windows only and that too will go away possibly. didn't think too much about it earlier. thanks!

cli/cmd/inject.go
mayankshah1607: @vaniisgh You might also have fix the [formatting](https://github.com/linkerd/linkerd2/blob/master/BUILD.md#formatting) here and in a few other places so that the CI checks pass :)

vaniisgh: ya, I saw that :( . for some reason, my editor seems to show them differently! fixing that.

vaniisgh:
@Pothulapati I can't really see your comments ...

pkg/inject/report.go
```diff
@@ -142,6 +150,24 @@ func checkUDPPorts(t *v1.PodSpec) bool {
...
 
+func checkNodeOS(t *v1.PodSpec) bool {
+	// Check any pods run a windows image, which is currently unsupported
+	//check the NodeSelector
+	osimage := t.NodeSelector["kubernetes.io/os"]
+	if strings.Contains(strings.ToLower(osimage), "windows") {
+		return true
+	}
+	//check if Node has been tainted
+	for _, toleration := range t.Tolerations {
```
Pothulapati, line 161: From what I have been reading [here](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), Looks like they allow you to get repelled from specific nodes from being scheduled? Did I understand something wrong?

pkg/inject/report.go
```diff
@@ -142,6 +150,24 @@ func checkUDPPorts(t *v1.PodSpec) bool {
 	return false
 }
 
+func checkNodeOS(t *v1.PodSpec) bool {
```
Pothulapati, line 153: Can we rename this to something like `IsWindowsNode()`, as we are returning a bool, but not the name?
Same goes for field in `report`

pkg/inject/report.go
vaniisgh: no, you're right. I need to set this to look for taints may be, and set a tolerant... or something of that sort.

adleong:
Hi @vaniisgh, are you still interested in working on this?

vaniisgh:
hey sorry I have been  a bit busy with things at the moment... won't probably get around it for a bit ...

Issue #4357, opened 2020-05-08
install_test flakiness: failed to create containerd task


kleimkuhler:
I have ran into this multiple times over the last few days:
```
Please note: by default, Tiller is deployed with an insecure 'allow unauthenticated users' policy.
To prevent this, run `helm init` with the --tiller-tls-verify flag.
For more information on securing your installation see: https://docs.helm.sh/using_helm/#securing-your-helm-installation
"linkerd" has been added to your repositories
Test script: [install_test.go] Params: [--linkerd-namespace=l5d-integration-helm --helm-path=/home/runner/work/linkerd2/linkerd2/bin/helm --helm-chart=/home/runner/work/linkerd2/linkerd2/charts/linkerd2 --helm-stable-chart=linkerd/linkerd2 --helm-release=l5d-integration-test --tiller-ns=l5d-integration-tiller --upgrade-helm-from-version=stable-2.7.1]
--- FAIL: TestEvents (0.23s)
    install_test.go:774: Found unexpected warning events:
        Reason: [Failed] Object: [linkerd-sp-validator-646d4575d8-wv86g] Message: [Error: failed to create containerd task: failed to start io pipe copy: unable to copy pipes: containerd-shim: opening w/o fifo "/run/containerd/io.containerd.grpc.v1.cri/containers/linkerd-proxy/io/702266748/linkerd-proxy-stdout" failed: context deadline exceeded]
```

kleimkuhler:
Looks like kubernetes/kubernetes#89064 is tracking this upstream

kleimkuhler:
Closing as we have not seen this issue in CI for a while.

PR #4371, opened 2020-05-11
Flaky tests: allow up to one pod restart in integration tests

alpeb:
The restart count test has been failing sporadically due to cluster
flakiness like what's reported in #4357.

Since we run `go test` with `--failfast`, when something fails in
`test/install_test-go` we don't get to the point where `TestEvents()` is
run so we never know the reason of the pod restart (depending on whether
the flakiness continues, I might do something for `TestEvents()` to run
regardless)

### DIFF
--- a/testutil/kubernetes_helper.go
+++ b/testutil/kubernetes_helper.go
@@ -164,7 +164,7 @@ func (h *KubernetesHelper) GetConfigUID(namespace string) (string, error) {
 }
 
 // CheckPods checks that a deployment in a namespace contains the expected
-// number of pods in the Running state, and that no pods have been restarted.
+// number of pods in the Running state, and that pods have not been restarted more than once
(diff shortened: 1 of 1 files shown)


adleong:
I'm not sure how I feel about this because it really risks hiding true errors.  Do we have a good understanding of why pods are restarting?

alpeb:
In my tests, I could reproduce the restarting after many runs and in that case the culprit was what's pointed out in #4372.
But you're right, the cause can be something else that can be a legit failure caused by linkerd. We could go with my earlier suggestion of having the events test get triggered every time after a failed test in `test/install_test.go`, so we can have a better chance of knowing the cause of the restart. We'd see then that a warning like #4372's might have caused the restart and then re-run the tests? Let me know what you think.

kleimkuhler:
@alpeb Will this be closed in favor of #4410?

alpeb:
> @alpeb Will this be closed in favor of #4410?

Yeah, closing now :+1:

PR #4377, opened 2020-05-12
Add addon-overwrite flag

### LINKED ISSUE #4376: Add-On Config overwrite flag
## Feature Request

Currently, when users apply a configuration where they explicitly, add value to some configuration. 
```yaml
grafana:
  resources:
    cpu:
      limit: "1"
      request: 100m
    memory:
      limit: 250Mi
      request: 50Mi
```

Now to remove the above `resources` configuration, Users can't just apply a configuration removing the field and except it to removed, like 
```yaml
grafana:
  enabled: true
```

This was intended behaviour as we store configuration in `linkerd-config-addons`, and merge it!

Removing a field that was applied through `--addon-config`

provide a `--addon-overwrite` flag, which when applied ignores the values from `linkerd-config-addons` and applies the file passed.

There are two workarounds currently
- apply a `field: null` which would remove the field.
- delete `linkerd-config-addons` and apply the same config again but without the field.

output?
`linkerd upgrade --addon-config config.yaml --addon-overwrite`

Pothulapati:
Fixes #4376 

This PR includes `addon-overwrite` flag which allows users to skip using the config from `linkerd-config-addons`, and use `addon-config` if passed or use the default one!

This change also moves `addon-overwrite` out of `recordableFlagSet` as the flag is persistent in linkerd-config, if the file is present locally it is automatically being used with upgrades even when the user dosen't pass `addon-config` 

changes include:
- add add-on overwrite flag
- move addon-config flag out of recordableFlagSet


cli/cmd/upgrade.go
```diff
@@ -346,25 +353,26 @@ func (options *upgradeOptions) validateAndBuild(stage string, k kubernetes.Inter
...
-		rawValues, err := yaml.Marshal(values)
-		if err != nil {
-			return nil, nil, err
-		}
+	if !options.addOnOverwrite {
+		// Update Add-Ons Configuration from the linkerd-value cm
+		cmRawValues, _ := k8s.GetAddOnsConfigMap(k, controlPlaneNamespace)
+		if cmRawValues != nil {
+			//Cm is present now get the data
+			cmData := cmRawValues["values"]
```
zaharidichev, line 361: Is it safe assuming that this key is present in the map? Is it worth verifying ?

Pothulapati: It is definitely safe theoretically, but added the verification here. Also, I added the case to fail here if `values` key is not present instead of skipping as it is weird to have `linkerd-config-addons` and not have `values`. @alpeb WDYT?

cli/cmd/inject.go
```diff
@@ -443,6 +443,7 @@ func (options *proxyConfigOptions) overrideConfigs(configs *cfg.All, overrideAnn
 	}
 	if options.proxyCPULimit != "" {
 		configs.Proxy.Resource.LimitCpu = options.proxyCPULimit
+
```
alpeb, line 446: Unintended?

cli/cmd/upgrade.go
alpeb: Agreed it's better to fail in those weird cases :+1:

cli/cmd/install.go
```diff
@@ -553,6 +554,19 @@ func (options *installOptions) allStageFlagSet() *pflag.FlagSet {
...
 	)
 
+	return flags
+}
+
+// allStageFlagSetButNotRecordable returns flags usable for single and multi-stage  installs and
+// upgrades but which are not recordable. For multi-stage installs, users must set these flags consistently
+// across commands.
+func (options *installOptions) allStageFlagSetButNotRecordable() *pflag.FlagSet {
+	flags := pflag.NewFlagSet("all-stage", pflag.ExitOnError)
```
alpeb, line 564: This is the same label as in `allStageFlagSet()`. Not sure if that clashes, but better to give it a different one.

cli/cmd/install.go
```diff
@@ -553,6 +554,19 @@ func (options *installOptions) allStageFlagSet() *pflag.FlagSet {
...
+
+	// addon-config is present here, instead of allStageFlagSet to prevent this from being recorded
+	// and thus for upgrades the filename will be returned, and if the file is present locally
+	// configuration will be applied automatically, without user passing the flag explicitly
+	// which is not desired behaviour.
```
alpeb, line 569: Which filename? Also, there's a bit of redundancy on what the function doc already explains.

Pothulapati: the name of the that was passed as `--addon-config`, but update the docs to better reflect the same.

alpeb:
In the `recordFlags()` function in `install.go` there's a list of flags to ignore during persistence. Can you consider using that for `addon-config` instead of adding the new function?

Pothulapati:
@alpeb Using that makes total sense, and updated it accordingly! Thanks for the find! :)

cli/cmd/upgrade.go
```diff
@@ -67,7 +68,10 @@ func (options *upgradeOptions) upgradeOnlyFlagSet() *pflag.FlagSet {
 		&options.force, "force", options.force,
 		"Force upgrade operation even when issuer certificate does not work with the trust anchors of all proxies",
 	)
-
+	flags.BoolVar(
+		&options.addOnOverwrite, "addon-overwrite", options.addOnOverwrite,
+		"Skip add-on config from linkerd-config-addons, and overwrite with the default or passed addon-config",
```
alpeb, line 73: Given people might not be aware of the existence of `linkerd-config-addons`, could you rewrite this to something like "Overwrite (instead of merge) existing add-ons config with file in --addon-config (or reset to defaults if no new config is passed)"?

This would require also changing the `--addon-config` description to something like:
"A path to a configuration file of add-ons. If an add-on is already installed, this new config gets merged with the existing one (unless --addon-overwrite is used)"

alpeb:
I've just noticed that `--addon-config` is available in both the config and control-plane stages of `linkerd install|upgrade` yet `linkerd-config-addons` is only built during the control-plane stage. Could you please fix that in a separate PR?
Also, `--addon-overwrite` should appear only there where `--addon-config` is present. But let get this current PR merged as it is for now :wink:

Pothulapati:
@alpeb `--addon-overwrite` is present in both cases because we need the add-on configuration even for `config` stage, for any templating present in RBAC stuff. I think we already have some templating w.r.t PSP.

alpeb:
Gotcha. But `--addon-overwrite` is currently present only in `linkerd upgrade` and not in the upgrade stages, unlike `--addon-config`. So that still needs to be addressed, correct? That can be done in this PR.

Pothulapati:
@alpeb Created https://github.com/linkerd/linkerd2/issues/4454 for the same!

PR #4381, opened 2020-05-12
Add global.grafanaUrl for linking existing grafana use-case

Pothulapati:
Final part of #3979, as per #3979 (comment) Feel free to look at it, to understand how these changes integrate.

This PR adds `global.grafanaUrl` which allows users to link their existing grafana url with Linkerd. 

The assumption is that they have already added the dashboards from [the repo](https://github.com/linkerd/linkerd2/tree/master/grafana/dashboards) into their existing grafana instance. This is a fairly easy process and there have been blog posts/stuff written from outside folks already  like https://medium.com/@chris_linguine/adding-linkerd-2-x-data-to-your-main-grafana-dashboard-4371cc1fa7e3 & https://github.com/andrew-waters/linkerd2-mixin


Pothulapati:
**Using existing grafana instance:**
- Let `config.yaml` be
```yaml
global:
  grafanaUrl: "somegrafana.xyz"
grafana:
  enabled: false
```
- Apply using `linkerd install --addon-config config.yaml | k apply -f -`

alpeb:
@Pothulapati  Could you please remind me of the purpose of `linkerd-config-addons`, and why all those configs can't leave inside `linkerd-config`?

Pothulapati:
@alpeb 

`linkerd-config` already stores a lot of information that is important and necessary for Linkerd and upgrades. The expectation with add-ons was that the add-on configuration would have a lot of stuff (this can be seen in the Prometheus add-on stuff), This would make `linkerd-config` bloated easily. 

The main reasoning with `linkerd-config-addons` was that it provides a separate place specific to add-on configuration and makes `--addon-config` through CLI possible without having a hard dependency on it.(like we have with `linkerd-config`). 

Also, regarding this change We will make sure that there is no other config stored here except the ones that we need with add-ons (especially in `global`)

alpeb:
Sorry I bring this up so late in the addon story... but I'm having second thoughts about the need of `linkerd-config-addons` and thinking it might be enough to have all in our current `linkerd-config` configmap, under a separate section. I don't think it would bloat `linkerd-config`. And people aren't supposed to manually change our current `linkerd-config`, nor should they if we add addons settings there, as they should instead do that through `--addon-config` and `addon-config-override`, right?

> and makes `--addon-config` through CLI possible without having a hard dependency on it.(like we have with `linkerd-config`).

I didn't get that part, could you clarify?

So the main question here is whether it's technically necessary to have the addon config be separate. Or maybe it's not such a big deal to have another configmap?

@grampelberg you have any opinion about this?

grampelberg:
@alpeb it has been awhile, so I'm having a tough time remembering all the details. A couple things come to mind though:

- I was going for something easier to introspect as the `linkerd-config` resource is pretty tough to understand right now (though that's fixable).
- I'd like to have the components try and be separate so that we don't step on each other's toes when editing code (potentially setting us up for the addons to be in their own repo and maintained by a larger group of folks).

What's the tradeoff of using a separate ConfigMap? I'm not sure I understand the downsides enough.

charts/linkerd2/values.yaml
```diff
@@ -19,6 +19,9 @@ global:
 
   identityTrustDomain: *cluster_domain
 
+  # url of exisiting grafana instance.
```
alpeb, line 22: Typo: existing
Or better yet, could you be more explicit and say something like "url for external Grafana instance"?

Pothulapati:
@alpeb The web dashboard integration is already present, right? Also the CLI tooling works too, as it just opens at `<dashboard>/grafana`, which is already done at #4319

alpeb:
@Pothulapati Ah sorry, I was expecting to see the external url instead of localhost, forgetting there's a proxy underneath. So we're good. Did you test this yourself with an actual external Grafana, and verified the query params are passed on properly?

Pothulapati:
@alpeb  Created a new grafana instance with our image and configuration, from the manifest [here](https://gist.githubusercontent.com/Pothulapati/9f91a9542a636808e5c118d91447c2c1/raw/c799631b3b1b6f8bd9b50d670e95dd36bbf86cdb/l-grafana.yaml) and applied that with the below config and the integration works correctly.
```yaml
global:
  grafanaUrl: grafana.default:3000

grafana:
  enabled: false

```

cli/cmd/install_addon_test.go
```diff
@@ -20,11 +21,19 @@ func TestAddOnRender(t *testing.T) {
 	withTracingAddonValues.Tracing["enabled"] = true
 	addFakeTLSSecrets(withTracingAddonValues)
 
+	withExistingGrafana, err := testInstallOptions()
+	if err != nil {
+		t.Fatalf("Unexpected error: %v\n", err)
```
zaharidichev, line 26: I think you can format errors as strings, right?

cli/cmd/testdata/existing-grafana-config.yaml
```diff
@@ -0,0 +1,4 @@
+global:
+  grafanaUrl: "somegrafana.xyz"
+grafana:
+  enabled: false
```
zaharidichev, line 4: missing \n at end of file. Is that necessary ?

cli/cmd/install_addon_test.go
Pothulapati: I see that its a common pattern in the tests to use %v instead of strings!

PR #4384, opened 2020-05-12
Check that ActualSuccess is greater than 0 in ServiceProfiles test

### LINKED ISSUE #4217: Add retries integration test
Add an integration test that tests Linkerd's retries feature.  Such a test would set up an application with a service profile configured to do retries and compare the `actual` and `effective` output from the `linkerd routes -o wide` command to verify that retries were happening and being reported correctly.

kleimkuhler:
#4217 suggests a retries integration test, but this is already tested as part
of the ServiceProfiles test.

In order to fix this issue, an extra check has been added to the assertion of
the `ActualSuccess` value. It now asserts the value is both greater than 0 and
less than 100.

Closes #4217

### DIFF
--- a/test/serviceprofiles/serviceprofiles_test.go
+++ b/test/serviceprofiles/serviceprofiles_test.go
@@ -175,7 +175,7 @@ func TestServiceProfileMetrics(t *testing.T) {
 	}
 
 	assertRouteStat(testUpstreamDeploy, testNamespace, testDownstreamDeploy, t, func(stat *cmd2.JSONRouteStats) error {
-		if *stat.ActualSuccess == 0.00 || *stat.ActualSuccess == 100.00 {
+		if !(*stat.ActualSuccess > 0.00 && *stat.ActualSuccess < 100.00) {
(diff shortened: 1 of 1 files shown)


test/serviceprofiles/serviceprofiles_test.go
```diff
@@ -175,8 +175,8 @@ func TestServiceProfileMetrics(t *testing.T) {
 	}
 
 	assertRouteStat(testUpstreamDeploy, testNamespace, testDownstreamDeploy, t, func(stat *cmd2.JSONRouteStats) error {
-		if *stat.ActualSuccess == 100.00 {
-			return fmt.Errorf("expected Actual Success to be less than 100%% due to pre-seeded failure rate. But got %0.2f", *stat.ActualSuccess)
+		if *stat.ActualSuccess == 0.00 || *stat.ActualSuccess == 100.00 {
```
adleong, line 178: Might be slightly more complete to actually check that it's > 0.0 and < 100.0

PR #4395, opened 2020-05-14
Separate grafana image tag in template

Pothulapati:
Currently, `image` in grafana, has a hard dependency on the linkerd tag.

So, even if users add a config like
```yaml
grafana:
  image: my-registry/grafana
```

it is required that the user has a image *with linkerd version* in the registry.

This is obviously a pretty bad requirement, and I missed updating this. 

So, now a separate field called `tag` is added through which user can specify a tag, whose default is `*linkerd-version`, which would be the current version of the CLI/Helm-Chart. 

Also, made sure that we build `gcr.io/linkerd-io/grafana` for each version as a tag.

@alpeb 


charts/linkerd2/values.yaml
```diff
@@ -196,6 +196,7 @@ grafana:
   enabled: true
   name: linkerd-grafana
   image: gcr.io/linkerd-io/grafana
+  tag: *linkerd_version
```
alpeb, line 199: To have this be more in line with the other settings in this file, could you use `image.name` and `image.version` instead?

alpeb:
@Pothulapati If you check `helmOverridesEdge()`, which generates the set of flags to pass into `helm install`, you'll see we override both `global.linkerdVersion` and `global.proxy.image.version` apparently because anchors in Helm templates won't update all references when modifying the original value :man_shrugging: 
So to fix the test you just gotta add an override in that function for `grafana.image.version`.

PR #4397, opened 2020-05-14
Exit tests if linkerd resources exist

kleimkuhler:
## Motivation

As mentioned in the [Testing RFC](https://github.com/linkerd/rfc/blob/master/design/0003-isolated-integration-tests.md#constraints):
> The integration test setup checks require that certain conditions are
> satisfied by the given cluster. A surprising condition is that no
> pre-existing Linkerd installation resource may exist; if it does then it is
> deleted.

## Solution

`init_test_run` which runs before integration tests start will now exit the
script if any Linkerd resources exist on the cluster.

Example bad path:
```
Checking if Linkerd resources exist on cluster...
Linkerd resources exist on cluster:

pod/hello-6b6b5d644d-xrnhn
pod/hello-slow-cooker-h8xn2
pod/world-fc8f457b7-gj7wq
pod/gateway-676fd64cb9-j57k6
pod/hello-c767bf764-cbdqh
pod/hello-slow-cooker-fqmxr
pod/slow-cooker-ftxdx
pod/t1-855c678bdd-vdg96
pod/t2-76989f94d4-d5fv8
pod/t3-75c8877797-hfwgc
pod/world-6784d4f65c-cn6vl
replicaset.apps/gateway-676fd64cb9
replicaset.apps/hello-c767bf764
replicaset.apps/t1-855c678bdd
replicaset.apps/t2-76989f94d4
replicaset.apps/t3-75c8877797
replicaset.apps/world-6784d4f65c
job.batch/hello-slow-cooker
job.batch/slow-cooker

Help:
    Run [/home/kevin/Projects/linkerd/linkerd2/bin/test-cleanup]
    Specify a cluster context [/home/kevin/Projects/linkerd/linkerd2/bin/test-run /home/kevin/Projects/linkerd/linkerd2/target/cli/linux/linkerd [l5d-integration] [context]]
exit
```

Example good path:
```
Checking if Linkerd resources exist on cluster...[ok]
```


kleimkuhler:
@adleong That's a great idea but unfortunately it's not the same behavior right now.

For example if a test installs Linkerd into `linkerd` ns, pods are created and injected in `default` ns, and then `linkerd` ns is deleted, there will be pods remaining in `default` ns that have the `linkerd.io/control-plane-ns: linkerd` label.

This passes `linkerd check --pre`, but not the script.

For the test script that is good; we don't want any resources still around when starting a new test.

bin/_test-run.sh
```diff
@@ -136,16 +136,18 @@ check_if_k8s_reachable(){
...
-    cleanup
+check_if_l5d_exists() {
+    printf 'Checking if Linkerd resources exist on cluster...'
+    resources=$(kubectl --context=$k8s_context get all,clusterrole,clusterrolebinding,mutatingwebhookconfigurations,validatingwebhookconfigurations,psp,crd -l linkerd.io/control-plane-ns --all-namespaces -oname)
+    if [ ! -z "$resources" ]; then
+        printf '\nLinkerd resources exist on cluster:\n\n'
+        printf '%s\n' "$resources"
+        printf '\nHelp:\n'
+        printf "    Run: [%s/test-cleanup]\n" "$bindir"
+        printf "    Specify a cluster context: [%s/test-run %s [%s] [context]]\n" "$bindir" "$linkerd_path" "$linkerd_namespace"
```
joakimr-axis, line 147: It is more resource efficient to keep this down to one `printf` call (and then split the line in the script for maintenance readability) instead of multiple ones.

bin/_test-run.sh
```diff
@@ -136,16 +136,18 @@ check_if_k8s_reachable(){
...
 
-remove_l5d_if_exists() {
-  resources=$(kubectl --context=$k8s_context get all,clusterrole,clusterrolebinding,mutatingwebhookconfigurations,validatingwebhookconfigurations,psp,crd -l linkerd.io/control-plane-ns --all-namespaces -oname)
-  if [ ! -z "$resources" ]; then
-    printf 'Removing existing l5d installation...'
-    cleanup
+check_if_l5d_exists() {
+    printf 'Checking if Linkerd resources exist on cluster...'
+    resources=$(kubectl --context=$k8s_context get all,clusterrole,clusterrolebinding,mutatingwebhookconfigurations,validatingwebhookconfigurations,psp,crd -l linkerd.io/control-plane-ns --all-namespaces -oname)
+    if [ ! -z "$resources" ]; then
```
joakimr-axis, line 142: Please use `-n` instead of `! -z`.

bin/_test-run.sh
kleimkuhler: Thanks! This is now one `printf`

PR #4402, opened 2020-05-15
Lint all markdown files in CI

kleimkuhler:
## Motivation

linkerd/rfc#22

## Solution

Use the [markdown-lint-action](https://github.com/marketplace/actions/markdown-linting-action) to lint all `.md` files for all pull requests
and pushes to master.

This action uses the default rules outlined in [markdownlint
package](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md).

The additional rules are added are explained below:
- Ignore line length lints for code blocks
- Ignore line length lints for tables
- Allow duplicate sub-headers in sibling headers (e.g. allowing multiple ##
  Significant headers in `CHANGES.md` as long as they are part of separate
  release headers)


kleimkuhler:
This depends on #4403 which applies all the lints that this will error on.

bin/markdownlint
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -eu
+
+markdownlint_version=0.23.0
+
+if ! [ -x "$(command -v markdownlint)" ]; then
```
adleong, line 7: is there a way to verify that the installed verison of markdownlint matches the expected version here?

kleimkuhler:
@adleong I answered this [here](https://github.com/linkerd/linkerd2/pull/4403#pullrequestreview-412877049).

I added a `bin/markdownlint` script here that will use npm to install the CLI command locally--not globally. I tried downloading the GH release and using `node` to run the JS file, but it's missing a few things that prevent it from being run out of the box.

Using `npm install markdownlint@version` makes the most sense.

.github/workflows/static_checks.yml
```diff
@@ -114,3 +114,18 @@ jobs:
...
+  markdown_lint:
+    name: Markdown lint
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout code
+      # actions/checkout@v2
+      uses: actions/checkout@722adc6
+    - name: Markdown lint
+      # nosborn/github-action-markdown-cli@v1.1.1
+      uses: nosborn/github-action-markdown-cli@v1.1.1
```
adleong, line 126: rather than using this off-the-shelf action, should we just use `bin/markdownlint`?  that way we wouldn't need to worry about keeping these in sync.

bin/markdownlint
kleimkuhler: The script checks for the version now as well. If either `markdownlint` is completely missing, or `markdownlint -V` does not equal `markdownlint_version`, it is installed.

.github/workflows/static_checks.yml
kleimkuhler: Yep good idea. I changed the action to use `bin/markdownlint`

kleimkuhler: Example of it running: https://github.com/kleimkuhler/linkerd2/pull/21/checks?check_run_id=679879798

alpeb: I don't know if npm guarantees that if one vets some tool's code at some version, that code for that same version won't be changed down the line with something bad. I like using github actions pointing to a sha (and not a version tag) whenever possible because you get that guarantee. This is just a tioli comment though :-)

adleong: a SHA is more in my comfort zone too

PR #4403, opened 2020-05-15
Lint all markdown files

kleimkuhler:
## Motivation

Necessary lints for #4402

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -19,7 +19,8 @@ about testing from source can be found in the [TEST.md](TEST.md) guide.
 - [Dependencies](#dependencies)
   - [Updating protobuf dependencies](#updating-protobuf-dependencies)
   - [Updating Docker dependencies](#updating-docker-dependencies)
-  - [Updating ServiceProfile generated code](#updating-serviceprofile-generated-code)
+  - [Updating ServiceProfile generated
+    code](#updating-serviceprofile-generated-code)
 - [Helm Chart](#helm-chart)
 - [Build Architecture](#build-architecture)
 - [Generating CLI docs](#generating-cli-docs)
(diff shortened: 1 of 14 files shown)


BUILD.md
```diff
@@ -48,52 +49,14 @@ written in Go. The dashboard UI is a React application.
...
-  definitions for the data plane APIs live in the linkerd2-proxy-api repo.
+- [`linkerd2-proxy-api`](https://github.com/linkerd/linkerd2-proxy-api):
+  Protobuf definitions for the data plane APIs live in the linkerd2-proxy-api
+  repo.
 
 ## Components
 
 ![Linkerd2 Components](https://g.gravizo.com/source/svg/linkerd2_components?https%3A%2F%2Fraw.githubusercontent.com%2Flinkerd%2Flinkerd2%2Fmaster%2FBUILD.md)
 
-<details>
```
grampelberg, line 58: Did you mean to strip the graphviz out?

BUILD.md
```diff
@@ -107,7 +70,7 @@ and run Linkerd2:
 
 This configuration builds all Linkerd2 components in Docker images, and deploys
 them onto Minikube. This setup most closely parallels our recommended production
-installation, documented at https://linkerd.io/2/getting-started/.
+installation, documented at <https://linkerd.io/2/getting-started/.>
```
grampelberg, line 73: Just make it a link?

BUILD.md
kleimkuhler: I did originally. We talked about this in private channel, but it is still necessary. It was rendering without it only because the link is hard coded for the master branch.

I added it back, but had to disable some markdown rules in this file because it is now a mix of indented and fenced code block styles. I can't fence (`'''html`) this part because then it actually renders it as an HTML code block which we don't want.

kleimkuhler: I did originally. We talked about this in private channel, but it is still necessary. It was rendering without it only because the link is hard coded for the master branch.

I added it back, but had to disable some markdown rules in this file because it is now a mix of indented and fenced code block styles. I can't fence (`'''html`) this part because then it actually renders it as an HTML code block which we don't want.

kleimkuhler:
> Can you include information about how these changes were generated?

@adleong `markdownlint -f` command fixed a small percentage of these; it was not too helpful. 

The line wrappings are a mix of manual corrections and a VS code extension that I use pretty regularly for this when writing PR descriptions. It's called [rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap).

The code block formatting and link corrections are manual.

> Is there a linting tool or command that generated this diff?

I just added a `bin/markdownlint` script to #4402

That script will run the markdownlint CLI command that I used for these fixes, and that CI uses when triggered.

PR #4404, opened 2020-05-15
Use -n instead of ! -z in shell scripts

joakimr-axis:


grampelberg:
@joakimr-axis why? I'm just interested academically =)

olix0r:
I'm +1 this change. I find `-n` to be generally clearer than having to reason through negation.

joakimr-axis:
> @joakimr-axis why? I'm just interested academically =)

I'd say that @olix0r sums it up quite well below. That would be the main reason. Fewer chars, easier to read, maintain and understand => decrease risk of (human) error.
And I guess that would also be the reason shellcheck will rag down on using `! -z` (or `! -n` for that matter).

PR #4405, opened 2020-05-15
install-cni.sh: Fix shellcheck issues

joakimr-axis:
Where cat and echo are actually not needed, they have been removed.

### DIFF
--- a/cni-plugin/deployment/scripts/install-cni.sh
+++ b/cni-plugin/deployment/scripts/install-cni.sh
@@ -65,7 +65,7 @@ cleanup() {
 
   if [ -e "${CNI_CONF_PATH}" ]; then
     echo "Removing linkerd-cni config: ${CNI_CONF_PATH}"
-    CNI_CONF_DATA=$(cat "${CNI_CONF_PATH}" | jq 'del( .plugins[]? | select( .type == "linkerd-cni" ))')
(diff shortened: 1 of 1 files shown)


adleong:
Hey @joakimr-axis, thanks for helping us improve our shell scripts!  Most of our scripts are not yet conformant with shellcheck and are grandfathered in; see the list at https://github.com/linkerd/linkerd2/blob/master/.github/workflows/static_checks.yml#L71

It would be great as part of this work if you could help us remove scripts from that list.  For any script that you want to update, if you could also remove it from the list of shellcheck exclusions and make any updates that are necessary to get it to pass, that would be amazing!

joakimr-axis:
> see the list at 

I was, as you suspected, not aware of this list. Thank you for sharing!

> It would be great as part of this work if you could help us remove scripts from that list.

Yes, there is some low hanging fruit there for sure.
And even though shellcheck, like any tool, is not always perfect, it is indeed good at helping us humans not to do human mistakes.

> For any script that you want to update, if you could also remove it from the list of shellcheck exclusions and make any updates that are necessary to get it to pass, that would be amazing!

+1

joakimr-axis:
> What would it take to get `_tag.sh` removed from the shellcheck exclusions?

Not much, I reckon; with this patch and the quote patch I think there is no more issues in _tag.

PR #4406, opened 2020-05-15
Fix quotes in shellscripts

joakimr-axis:
- Add quotes where missing, to handle whitespace & c:o.
- Use single quotes for non-expansion strings.
- Fix quotes were the current would cause errors.

### DIFF
--- a/bin/_test-run.sh
+++ b/bin/_test-run.sh
@@ -26,7 +26,10 @@ init_test_run() {
   test_directory=$bindir/../test
   linkerd_version=$($linkerd_path version --client --short)
   linkerd_namespace=${2:-l5d-integration}
-  k8s_context=${3:-""}
+  k8s_context=${3:-''}
+  export linkerd_version
+  export linkerd_namespace
+  export k8s_context
 
   check_linkerd_binary
   check_if_k8s_reachable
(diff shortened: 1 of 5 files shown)


adleong:
Thanks @joakimr-axis, please see this comment about shellcheck exclusions: https://github.com/linkerd/linkerd2/pull/4405#issuecomment-629417639

bin/_tag.sh
```diff
@@ -17,7 +17,7 @@ clean_head() {
 }
 
 named_tag() {
-    tag="$(git name-rev --tags --name-only $(git_sha_head))"
+    tag="$(git name-rev --tags --name-only "$(git_sha_head)")"
```
kleimkuhler, line 20: Any reason the inner quotes need to be escaped in `validate_tag` but not here?

joakimr-axis: Are you referring to the echo parts of `validate_tag`? If so, that is since we want to print the " in the output and were they not escaped they would declare the end of the string to be echoed, which I believe was not what is intended. I expect the intent was to have that output as something like:

```
Tag in "/tmp/myfile" does not match source tree:
foo ("/tmp/myfile")
bar (source)
```

joakimr-axis:
There, now I got the last wrinkles straightened out here, it seems.

bin/_test-run.sh
```diff
@@ -141,17 +145,17 @@ Help:
...
 run_test(){
     filename=$1
     shift
 
-    printf 'Test script: [%s] Params: [%s]\n' "$(basename $filename 2>/dev/null || echo $filename )" "$*"
-    GO111MODULE=on go test --failfast --mod=readonly $filename --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@"
-    [ $? -ne 0 ] && exit 1
+    printf 'Test script: [%s] Params: [%s]\n' "${filename##*/}" "$*"
+    # Exit on failure here
+    GO111MODULE=on go test --failfast --mod=readonly "$filename" --linkerd="$linkerd_path" --k8s-context="$k8s_context" --integration-tests "$@" || exit 1
```
kleimkuhler, line 158: 👍

bin/_test-run.sh
```diff
@@ -191,45 +195,48 @@ install_stable() {
 # this branch.
 # $1 - namespace to use for the stable release
 run_upgrade_test() {
-    local stable_namespace=$1
-    local stable_version=$(latest_stable)
+    local stable_namespace
+    local stable_version
+    stable_namespace=$1
+    stable_version=$(latest_stable)
```
kleimkuhler, line 201: These still should be fine as the below I think?

```
local stable_namespace=$1
local stable_version=$(latest_stable)
```
Any reason for splitting them out?

bin/_test-run.sh
```diff
@@ -191,45 +195,48 @@ install_stable() {
...
     exit_on_err 'error setting up Helm'
 }
 
 run_helm_upgrade_test() {
     setup_helm
-    local stable_version=$(latest_stable)
-    run_test "$test_directory/install_test.go" --linkerd-namespace=$linkerd_namespace-helm \
-        --helm-path="$helm_path" --helm-chart="$helm_chart" --helm-stable-chart="linkerd/linkerd2" --helm-release=$helm_release_name --upgrade-helm-from-version="$stable_version"
+    local stable_version
+    stable_version=$(latest_stable)
```
kleimkuhler, line 219: Same question as above

bin/_test-run.sh
joakimr-axis: Yes, it is due to SC2155, "Declare and assign separately to avoid masking return values" with the details explained here:

https://github.com/koalaman/shellcheck/wiki/SC2155

kleimkuhler: I only checked `stable_namespace` which is okay as `local stable_namespace=$1`.

Makes sense that `stable_version` which is being assigned to a return value though would have the lint though.

PR #4407, opened 2020-05-15
Add HealthChecks for Tracing Add-On

Pothulapati:
This PR adds the following healthchecks which are transient i.e ran only when the relevant tracing add-on configuration is present 
- checks for service accounts of collector and Jaeger
- check for collector config map
- checks for collector and jaeger pods to be running


Pothulapati:
@alpeb Updated the code to use only safe casts :)

Pothulapati:
@alpeb @zaharidichev  Updated this PR to make multicluster checks run after the add-on tests. 
```yaml
linkerd-addons
--------------

linkerd-grafana
---------------
√ grafana pod is running

linkerd-multicluster
--------------------
√ service mirror controller is running
√ service mirror controller ClusterRoles exist
√ service mirror controller ClusterRoleBindings exist
√ service mirror controller Roles exist
√ service mirror controller RoleBindings exist
√ service mirror controller ServiceAccounts exist
√ service mirror controller has required permissions
√ service mirror controller can access remote clusters
√ clusters share trust anchors
```

As all the add-on tests are `warning=true`, even if something fails it should still run tests that come after!

PR #4409, opened 2020-05-15
Add --close-wait-timeout inject flag

### LINKED ISSUE #4276: Increase nf_conntrack_tcp_timeout_close_wait

By default, `nf_conntrack_tcp_timeout_close_wait` is set to 60 seconds. Any requests taking longer than that get lost by the kernel causing 502 gateway errors to applications. `kube-proxy` bumped this up to 3600 seconds for its rules (see https://github.com/kubernetes/kubernetes/issues/32551).

It is not possible to modify this via. the pod spec as it is marked "unsafe" in most k8s installations. However, as it is part of the network namespace, you can set it in init containers and have it persist for the life of the session. We should do this inside the `proxy-init` container and as part of setup for CNI.

adleong:
Depends on https://github.com/linkerd/linkerd2-proxy-init/pull/10

Fixes #4276 

We add a `--close-wait-timeout` inject flag which configures the proxy-init container to run with `privileged: true` and to set `nf_conntrack_tcp_timeout_close_wait`. 


charts/linkerd2/values.yaml
```diff
@@ -65,6 +65,7 @@ global:
       memory:
         limit: 50Mi
         request: 10Mi
+    closeWaitTimeoutSecs: 0
```
alpeb, line 68: Can you add an entry for this in `./charts/linkerd2/README.md`?

PR #4410, opened 2020-05-15
Flaky tests: when installation test fails, fetch logs and events

alpeb:
Re #4371

When a test fails in `./test/install_test.go`, trigger the `TestLogs`
and `TestEvents` tests in a separate process in order to output any
unexpected logs/events that might have caused the initial test failure.

For instance, currently we're sporadically experiencing pod restarts.
Instead of ignoring them, this might help provide us with the real
underlying cause.

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -19,11 +19,22 @@ import (
 ///   TEST SETUP   ///
 //////////////////////
 
-var TestHelper *testutil.TestHelper
+var (
+	TestHelper *testutil.TestHelper
+
+	// fetchLogsEvents delimits the tests for which a failure triggers running
+	// the Testlogs and TestEvents tests
+	fetchLogsEvents = false
+)
 
 func TestMain(m *testing.M) {
 	TestHelper = testutil.NewTestHelper()
-	os.Exit(m.Run())
+	code := m.Run()
+	// if the test failed, show logs and events that can potentially explain cause
+	if code != 0 && fetchLogsEvents {
+		TestHelper.RunLogsAndEventTests()
+	}
+	os.Exit(code)
 }
 
 var (
@@ -195,6 +206,8 @@ func TestInstallOrUpgradeCli(t *testing.T) {
 	if TestHelper.GetHelmReleaseName() != "" {
 		return
 	}
+	// as of this point, fetch logs and events if any test fails
+	fetchLogsEvents = true
 
 	var (
 		cmd  = "install"
@@ -357,6 +370,8 @@ func TestInstallHelm(t *testing.T) {
 	if TestHelper.GetHelmReleaseName() == "" {
 		return
 	}
+	// as of this point, fetch logs and events if any test fails
+	fetchLogsEvents = true
 
 	cn := fmt.Sprintf("identity.%s.cluster.local", TestHelper.GetLinkerdNamespace())
 	var err error
@@ -721,6 +736,9 @@ func TestCheckProxy(t *testing.T) {
 }
 
 func TestLogs(t *testing.T) {
+	// as of this point, don't fetch logs nor events if any test fails
+	fetchLogsEvents = false
+
 	controllerRegex := regexp.MustCompile("level=(panic|fatal|error|warn)")
 	proxyRegex := regexp.MustCompile(fmt.Sprintf("%s (ERR|WARN)", k8s.ProxyContainerName))
 	clientGoRegex := regexp.MustCompile("client-go@")
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -33,15 +33,15 @@ func TestMain(m *testing.M) {
 	code := m.Run()
 	// if the test failed, show logs and events that can potentially explain cause
 	if code != 0 && controlPlaneInstalled {
-		_, logs := fetchAndCheckLogs()
-		for _, err := range logs {
+		_, errs1 := fetchAndCheckLogs()
+		for _, err := range errs1 {
 			fmt.Println(err)
 		}
-		evts := fetchAndCheckEvents()
-		for _, err := range evts {
+		errs2 := fetchAndCheckEvents()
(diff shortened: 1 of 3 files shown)


alpeb:
Thanks @cpretzer :+1: The way I've been testing this is just by adding a new test that fails, say after `TestInstallHelm()`, like
```go
func TestFoo(t *testing.T) {
  testutil.Fatal(t, "This fails")
}
```

test/install_test.go
```diff
@@ -721,6 +736,9 @@ func TestCheckProxy(t *testing.T) {
 }
 
 func TestLogs(t *testing.T) {
+	// as of this point, don't fetch logs nor events if any test fails
+	fetchLogsEvents = false
```
adleong, line 740: I don't really understand why we enable and disable this in a few places.  Is this to prevent infinite recursion or something?

testutil/test_helper.go
```diff
@@ -366,6 +369,33 @@ func (h *TestHelper) HTTPGetURL(url string) (string, error) {
 	return body, err
 }
 
+// RunLogsAndEventTests runs the TestLogs and TestEvents tests in a separate
+// process
+func (h *TestHelper) RunLogsAndEventTests() {
+	// don't run for tests that were singled out
+	if h.run != "" {
```
adleong, line 376: I'm confused about this `run` field.  It seems that we only check if it's empty but never check the value.  What is it supposed to do?

testutil/test_helper.go
```diff
@@ -366,6 +369,33 @@ func (h *TestHelper) HTTPGetURL(url string) (string, error) {
...
+		return
+	}
+	args := []string{"test", "-v", ".",
+		"-integration-tests",
+		"-linkerd", h.linkerd,
+		"-run",
+	}
+	out, err := exec.Command("go", append(args, "TestLogs")...).Output()
```
adleong, line 384: Why do we invoke these tests by shelling out instead of calling the test functions directly?

test/install_test.go
alpeb: Before installing anything in the cluster, if a test fails, fetching the logs takes an eternity because it tries unsuccessfully to pull them for each container on each pod of the control plane, with a long timeout for each. So I'm only enabling fetching the logs after the tests that attempt installing the CP. And then I'm disabling that on the `TestLogs` test indeed for avoiding infinite recursion if that test fails.

testutil/test_helper.go
alpeb: From within `TestMain` there are no references to `t`, so I couldn't find a way to call a test function directly. I would be glad to find there's a better way than shelling out.

testutil/test_helper.go
alpeb: That was used to identify when we shell out, so to avoid calling `TestLogs` and `TestEvents` recursively in the child process, but we're already avoiding that with the `fetchLogsEvents` var, so I think I can get rid of this...

test/install_test.go
```diff
@@ -19,11 +19,22 @@ import (
 ///   TEST SETUP   ///
 //////////////////////
 
-var TestHelper *testutil.TestHelper
+var (
+	TestHelper *testutil.TestHelper
+
+	// fetchLogsEvents is a switch to enable calling Testlogs and TestEvents
+	// when a test fails
+	fetchLogsEvents = false
```
adleong, line 27: I think this should be descriptive rather than imperative.  i.e. it should be called `controlPlaneInstalled` or something and indicate that the control plane has been installed.

test/install_test.go
```diff
@@ -19,11 +19,22 @@ import (
...
+)
 
 func TestMain(m *testing.M) {
 	TestHelper = testutil.NewTestHelper()
-	os.Exit(m.Run())
+	code := m.Run()
+	// if the test failed, show logs and events that can potentially explain cause
+	if code != 0 && fetchLogsEvents {
```
adleong, line 34: I think the logic is clearer here if we rename the fetchLogsEvents variable to controlPlaneInstalled.  This makes it obvious that we only want to fetch logs and events if the control plane has already been installed.

alpeb:
Great feedback :+1: 
I ended up extracting the log/event retrieval log into separate functions, containing both the network calls and the regex matching. There's no longer need for shelling out, and the `fetchLogsEvents` was renamed. LMKWYT.

test/install_test.go
```diff
@@ -743,85 +770,96 @@ func TestLogs(t *testing.T) {
...
-							// report all known logging errors in the output
-							t.Logf("Found known error in %s log: %s", name, line)
+			for _, line := range outputLines {
+				if errRegex.MatchString(line) {
+					if knownErrorsRegex.MatchString(line) {
+						// report all known logging errors in the output
+						okMessages = append(okMessages, fmt.Sprintf("found known error in %s log: %s", name, line))
+					} else {
+						if proxy {
+							okMessages = append(okMessages, fmt.Sprintf("found unexpected proxy error in %s log: %s", name, line))
```
kleimkuhler, line 802: I see this is similar to the previous behavior, but any idea why we don't append to `errs` on unknown errors in the proxy?

test/install_test.go
```diff
@@ -19,11 +20,32 @@ import (
...
+	controlPlaneInstalled = false
+)
 
 func TestMain(m *testing.M) {
 	TestHelper = testutil.NewTestHelper()
-	os.Exit(m.Run())
+	code := m.Run()
+	// if the test failed, show logs and events that can potentially explain cause
+	if code != 0 && controlPlaneInstalled {
+		_, logs := fetchAndCheckLogs()
```
adleong, line 36: should we rename `logs` to `errs` since this is a `[]error`?

test/install_test.go
alpeb: According to #2647 the proxy error list kept on growing and so was causing regular CI failures.

PR #4425, opened 2020-05-18
bin/web: Fix shellcheck issues

joakimr-axis:


bin/web
```diff
@@ -121,7 +122,7 @@ integration() {
 }
 
 main() {
-  setup
+  setup ''
```
olix0r, line 125: I'm a little surprised that this is necessary?

joakimr-axis: Me too, but `setup` takes arguments. When none are given on this line, shellcheck will ask if you forgot to pass on the script's arguments or what is going on. The behavior of both lines above will be the same from `setup`s point of view, but shellcheck will know what we intend to do.
At least that is what I can make out of shellcheck's complaining here.

kleimkuhler: `setup()` has a line where it passes on additional arguments: `yarn "$@"`. Even though we don't have any arguments to pass in here, shellcheck warns against not being explicit.

PR #4426, opened 2020-05-18
Restrict Service mirror RBACs

### LINKED ISSUE #4399: service-mirror crashes with "failed to sync caches"
There appears to be a problem with the kubeconfig I've provided to the service-mirror, but there's not a lot to go on from the logs:

```
:; kubectl --context=k3d-multi0 logs -f -n linkerd-service-mirror linkerd-service-mirror-856454fb7b-l9frg service-mirror
time="2020-05-14T20:22:08Z" level=info msg="running version edge-20.5.2"
time="2020-05-14T20:22:08Z" level=info msg="waiting for caches to sync"
time="2020-05-14T20:22:08Z" level=info msg="caches synced"
time="2020-05-14T20:22:08Z" level=info msg="Started cluster config watcher"
time="2020-05-14T20:22:08Z" level=info msg="starting admin server on :9999"
time="2020-05-14T20:22:08Z" level=info msg="waiting for caches to sync"
time="2020-05-14T20:23:08Z" level=fatal msg="failed to sync caches"
```

The pod ends up crash-looping. Is it possible to surface a better error message? i.e. Was the connection refused? or did we fail a TLS handshake? Or???

### LINKED ISSUE #4401: service-mirror cannot require global secret access
install-service-mirror currently emits rbac like:

```yaml
kind: ClusterRole
metadata:
  name: linkerd-service-mirror
  namespace: linkerd-service-mirror
  labels:
    linkerd.io/control-plane-component: linkerd-service-mirror
rules:
- apiGroups: [""]
  resources: ["endpoints", "services", "secrets", "namespaces"]
  verbs: ["list", "get", "watch", "create", "delete", "update"]
```

Which permits the service-mirror to read and write all secrets in the cluster. This isn't a suitable default for production environments.

What secret access does this pod need?

zaharidichev:
This PR introduces a few changes that were requested after a bit of service mirror reviewing.

- we restrict the RBACs so the service mirror controller cannot read secrets in all namespaces but only in the one that it is installed in
- we unify the namespace namings so all multicluster resources are installedi n `linkerd-multicluster` on both clusters
- we provide a better error message when k8s api caches fail to sync
- fixed checks to account for changes

Fix #4401
Fix #4399


controller/cmd/tap/main.go
```diff
@@ -43,17 +43,17 @@ func Main(args []string) {
...
+		k8s.RT(k8s.DS),
+		k8s.RT(k8s.SS),
+		k8s.RT(k8s.Deploy),
+		k8s.RT(k8s.Job),
+		k8s.RT(k8s.NS),
+		k8s.RT(k8s.Pod),
+		k8s.RT(k8s.RC),
+		k8s.RT(k8s.Svc),
+		k8s.RT(k8s.RS),
+		k8s.RT(k8s.Node),
```
olix0r, line 56: Having to modify all of the constructors in this way is surprising. What motivates this? Can we instead do this transformation within `InitializeAPI`. If this really is something that needs to be called in all of these contexts, the `RT` abbreviation means nothing to me... is this something that should be understood in the broader k8s ecosystem? (The other abbreviations are canonical within k8s afaiu).

olix0r:
Generally, it's a little hard for me to contextualize all of the changes to k8s/api.go... Is it possible to split this into smaller, more targeted changes? I.e. getting the namespace change as a discreet PR, checks changes, etc.

cli/cmd/cluster.go
```diff
@@ -356,7 +354,7 @@ func newGetCredentialsCommand() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&opts.serviceAccountName, "service-account-name", defaultServiceAccountName, "the name of the service account")
-	cmd.Flags().StringVar(&opts.serviceAccountNamespace, "service-account-namespace", defaultServiceAccountNs, "the namespace in which the service account will be created")
+	cmd.Flags().StringVar(&opts.serviceAccountNamespace, "service-account-namespace", defaultMulticlusterNamespace, "the namespace in which the service account will be created")
```
alpeb, line 357: Actually the NS is hard-coded to `defaultMulticlusterNamespace` above where the Secret is created, and this param is for the NS of the SA in the remote cluster.

cli/cmd/install-service-mirror.go
```diff
@@ -92,6 +92,10 @@ func (options *installServiceMirrorOptions) validate() error {
 		return errors.New("you need to specify a namespace")
 	}
 
+	if options.namespace == controlPlaneNamespace {
+		return errors.New("you need to install the service mirror component in a namespace different than the Linkerd one")
+	}
```
alpeb, line 97: Note that for this check to be effective it requires that the user passes the `--linkerd-namespace` flag if it's not the default "linkerd". That, or having this command fetch `linkerd-config` CM.

controller/cmd/tap/main.go
zaharidichev: The reason for this is that instead of simply an enum that is encoded as an int, I want to pass a struct that contains the resource type and an optional namespace scope. The problem is that  cannot really have enum constants in go and this necessitates this constructor call here... If that wasnt the case the change would be less visible. We can make it so this is somehow handled within the `InitializeAPI` method but we would still need to pass aditional info somehow. So all in all I do not really see a good way to avoid that. RT should stand for ResourceType but clearly that is poor choice of abbreviation. I can split the change in two is that helps

cli/cmd/install-service-mirror.go
zaharidichev: So do you suggest I actually check for the presence of the conf map instead ?

alpeb: Yeah, IIUC we're coupling the service mirror controller with a linkerd control plane after all, so we could do just like with `linkerd cluster setup-remote` and check for the CM

zaharidichev: Yes, makes sense! That is what I will do

controller/cmd/tap/main.go
alpeb: How about having `initializeAPI()` build that `APIResource` using `corev1.NamespaceAll`, leaving the calling sites for all the old controllers untouched, and have a new `InitializeNamespacedAPI()` only used by the service-mirror that would receive the namespace as an argument and build `APIResource` with that namespace?

cli/cmd/cluster.go
```diff
@@ -229,13 +227,12 @@ func newSetupRemoteCommand() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&options.gatewayName, "gateway-name", options.gatewayName, "the name of the gateway")
-	cmd.Flags().StringVar(&options.gatewayNamespace, "gateway-namespace", options.gatewayNamespace, "the namespace in which the gateway will be installed")
+	cmd.Flags().StringVar(&options.namespace, "namespace", options.namespace, "the namespace in which the gateway and SA will be installed")
```
adleong, line 230: let's write our "service account" in full here

controller/k8s/api.go
```diff
@@ -38,13 +39,35 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-// APIResource is an enum for Kubernetes API resource types, for use when
+// APIResource represents a namespaces k8s api resource
```
adleong, line 42: I think I would rather that service mirror build it's own secret informer instead of trying to make the existing api.go machinery flexible enough to support this use case.

controller/cmd/tap/main.go
zaharidichev: I was thinking abou that but in our particular case we have some resources that are namespaces (secrets) and some that are not that want to share the same api. So thi will not work as we would want to specify namespace constraints per resource in this case. I say we just do it the go way and copy the code all over ;)

zaharidichev: Okey... Now that I made the change it is clear that we need to touch quite a bit less stuff than originally to make this all work. So it is worth it. Now client.go is almost untouched.  Good suggestion !

zaharidichev:
@alpeb @adleong I fixed the things you pointed out

controller/k8s/api.go
```diff
@@ -256,9 +257,10 @@ func (api *API) Sync(stopCh <-chan struct{}) {
 
 	log.Infof("waiting for caches to sync")
 	if !cache.WaitForCacheSync(ctx.Done(), api.syncChecks...) {
-		log.Fatal("failed to sync caches")
+		return errors.New("failed to sync caches")
```
alpeb, line 260: This will require changes on all the controllers calling `Sync()` that are not catching this error and will now continue running despite the API not being ready.

pkg/healthcheck/healthcheck.go
```diff
@@ -1598,6 +1628,7 @@ func (hc *HealthChecker) checkServiceMirrorController() error {
 		if controller.Status.AvailableReplicas < 1 {
 			return fmt.Errorf("Service mirror controller is not available: %s/%s", controller.Namespace, controller.Name)
 		}
+		println(fmt.Sprint(controller.Namespace))
```
alpeb, line 1631: :open_mouth:

controller/k8s/api.go
```diff
@@ -246,7 +247,7 @@ func NewAPI(
 }
 
 // Sync waits for all informers to be synced.
-func (api *API) Sync(stopCh <-chan struct{}) {
+func (api *API) Sync(stopCh <-chan struct{}) error {
```
adleong, line 250: this seems like a pretty invasive change, just so that we can print a customized error in the service mirror.  I think I'm -1 on this.

controller/cmd/service-mirror/main.go
```diff
@@ -44,11 +69,20 @@ func Main(args []string) {
...
+		log.Fatalf("Failed to initialize secrets informer: %s", err)
+	}
+
 	probeManager := NewProbeManager(k8sAPI)
 	probeManager.Start()
 
-	k8sAPI.Sync(nil)
-	watcher := NewRemoteClusterConfigWatcher(k8sAPI, *requeueLimit, &chanProbeEventSink{probeManager.enqueueEvent})
+	if err := k8sAPI.Sync(nil); err != nil {
+		log.Fatalf("Failed to sync caches for k8s api. Check your service mirror RBACs")
```
adleong, line 82: Without looking at `err` here, the RBAC thing is just a guess.  What if this fails for some other reason?  

Can we actively check RBACs before syncing by doing a `can-i` call?

controller/cmd/service-mirror/main.go
```diff
@@ -44,11 +69,17 @@ func Main(args []string) {
...
+	secretsInformer, err := initLocalSecretsInformer(k8sAPI.Client, *namespace)
+
+	if err != nil {
+		log.Fatalf("Failed to initialize secrets informer: %s", err)
+	}
+
 	probeManager := NewProbeManager(k8sAPI)
 	probeManager.Start()
 
 	k8sAPI.Sync(nil)
```
adleong, line 81: TIOLI: we could add a TODO here for adding a proactive `can-i` check to catch RBAC errors here that would not be caught by `linkerd check` in private clusters.

PR #4432, opened 2020-05-19
Remove proxy.image.version check in templates

Pothulapati:
This PR fixes using wrong case for `proxy.image.version`, during its setting which happens when it is empty, in `tracing.yaml` and `prometheus.yaml`

Considering the empty case never ran, As we would get errors otherwise for setting wrong version. Are these even useful?


alpeb:
> Considering the empty case never ran, As we would get errors otherwise for setting wrong version. Are these even useful?

Yeah, better to simplify things by removing this, IMO :+1:

PR #4440, opened 2020-05-19
Enable service mirroring to work in private networks

### LINKED ISSUE #4411: service-mirror fails when remote gateway is on a private IP
When service-mirror is configured with a remote gateway on a private (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16) network, the service-mirror's proxy attempts discovery and the destination service  responds with no identity; and the outbound requests fail because of the `l5d-require-id` header, which requires that the identity is configured (and it is not).

To fix this, I propose the following:
- We create a service for each gateway, e.g. `gateway-multi1`, `gateway-multi2`. This service has two ports (i propose gateway:4180 and healthcheck:4181) and one endpoint for each remote gateway replica.
- The service-mirror calls gateways by their name and not IP.
- The service-mirror manages endpoint readiness according to probes.
- The destination service is able to discover gateway identities (via annotation)?.

wdyt @zaharidichev @grampelberg

zaharidichev:
This change creates a gateway proxy for every gateway. This enables the probe worker to leverage the destination service functionality in order to discover the identity of the gateway. 

Fix #4411

### DIFF
--- a/charts/linkerd2-service-mirror/templates/service-mirror.yaml
+++ b/charts/linkerd2-service-mirror/templates/service-mirror.yaml
@@ -16,7 +16,7 @@ rules:
   verbs: ["list", "get", "watch", "create", "delete", "update"]
 - apiGroups: [""]
   resources: ["namespaces"]
-  verbs: ["list", "get", "watch"]
+  verbs: ["create","list", "get", "watch"]
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -169,8 +169,8 @@ func newGatewaysCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "remote", "the name of the remote cluster")
-	cmd.Flags().StringVar(&opts.gatewayNamespace, "gateway-namespace", defaultMulticlusterNamespace, "the namespace in which the gateway resides on the remote cluster")
+	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "", "the name of the remote cluster")
+	cmd.Flags().StringVar(&opts.gatewayNamespace, "gateway-namespace", "", "the namespace in which the gateway resides on the remote cluster")
 	cmd.Flags().StringVarP(&opts.timeWindow, "time-window", "t", "1m", "Time window (for example: \"15s\", \"1m\", \"10m\", \"1h\"). Needs to be at least 15s.")
 
 	return cmd
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -500,9 +500,9 @@ func remapRemoteServicePorts(ports []corev1.ServicePort) []corev1.ServicePort {
 	return newPorts
 }
 
-func (rcsw *RemoteClusterServiceWatcher) updateGatewayProxyService(spec *GatewaySpec) error {
-	localSvcProxyName := rcsw.mirroredResourceName(spec.gatewayName)
-	service, err := rcsw.localAPIClient.Svc().Lister().Services(rcsw.serviceMirrorNamespace).Get(localSvcProxyName)
+func (rcsw *RemoteClusterServiceWatcher) updateGatewayMirrorService(spec *GatewaySpec) error {
+	localServiceName := rcsw.mirroredResourceName(spec.gatewayName)
+	service, err := rcsw.localAPIClient.Svc().Lister().Services(rcsw.serviceMirrorNamespace).Get(localServiceName)
 	if err != nil {
 		return err
 	}
@@ -513,7 +513,7 @@ func (rcsw *RemoteClusterServiceWatcher) updateGatewayProxyService(spec *Gateway
 			updatedService.Annotations[consts.RemoteGatewayResourceVersionAnnotation] = spec.resourceVersion
 		}
 
-		endpoints, err := rcsw.localAPIClient.Endpoint().Lister().Endpoints(rcsw.serviceMirrorNamespace).Get(localSvcProxyName)
+		endpoints, err := rcsw.localAPIClient.Endpoint().Lister().Endpoints(rcsw.serviceMirrorNamespace).Get(localServiceName)
 		if err != nil {
 			return err
 		}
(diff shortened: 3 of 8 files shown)


charts/linkerd2-service-mirror/templates/service-mirror.yaml
```diff
@@ -16,7 +16,7 @@ rules:
   verbs: ["list", "get", "watch", "create", "delete", "update"]
 - apiGroups: [""]
   resources: ["namespaces"]
-  verbs: ["list", "get", "watch"]
+  verbs: ["create","list", "get", "watch"]
```
adleong, line 19: just to clarify, this is unrelated to this change right?  we would need namespace create permission to create mirror services if their namespace doesn't exist but we won't need it to create the gateway mirrors since those will always be in the service-mirror namespace?

cli/cmd/cluster.go
```diff
@@ -169,8 +169,8 @@ func newGatewaysCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "remote", "the name of the remote cluster")
-	cmd.Flags().StringVar(&opts.gatewayNamespace, "gateway-namespace", defaultMulticlusterNamespace, "the namespace in which the gateway resides on the remote cluster")
+	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "", "the name of the remote cluster")
```
adleong, line 172: just to clarify, leaving these blank means that the default is to look at all remote clusters?  and all gateway namespaces?

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -497,6 +500,132 @@ func remapRemoteServicePorts(ports []corev1.ServicePort) []corev1.ServicePort {
 	return newPorts
 }
 
+func (rcsw *RemoteClusterServiceWatcher) updateGatewayProxyService(spec *GatewaySpec) error {
```
adleong, line 503: calling these gateway proxies is pretty confusing to me.  can we call them "gateway mirrors" or something?

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -497,6 +500,132 @@ func remapRemoteServicePorts(ports []corev1.ServicePort) []corev1.ServicePort {
...
+	localSvcProxyName := rcsw.mirroredResourceName(spec.gatewayName)
+	service, err := rcsw.localAPIClient.Svc().Lister().Services(rcsw.serviceMirrorNamespace).Get(localSvcProxyName)
+	if err != nil {
+		return err
+	}
+
+	if service.Annotations != nil && service.Annotations[consts.RemoteGatewayResourceVersionAnnotation] != spec.resourceVersion {
+		updatedService := service.DeepCopy()
+		if updatedService.Annotations != nil {
```
adleong, line 512: probably don't nee this check because we just checked this in the original service and haven't modified the copy yet.

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -497,6 +500,132 @@ func remapRemoteServicePorts(ports []corev1.ServicePort) []corev1.ServicePort {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      localSvcProxyName,
+					Namespace: rcsw.serviceMirrorNamespace,
+					Labels: map[string]string{
+						consts.MirroredResourceLabel:  "true",
+						consts.RemoteClusterNameLabel: rcsw.clusterName,
+					},
+					Annotations: map[string]string{
+						consts.RemoteGatewayIdentity:                  spec.identity,
+						consts.RemoteGatewayResourceVersionAnnotation: spec.resourceVersion,
```
adleong, line 589: should this annotation be on the service instead of the endpoints?

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -497,6 +500,132 @@ func remapRemoteServicePorts(ports []corev1.ServicePort) []corev1.ServicePort {
...
+
+	_, err := rcsw.localAPIClient.Svc().Lister().Services(rcsw.serviceMirrorNamespace).Get(localSvcProxyName)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			serviceToCreate := &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      localSvcProxyName,
+					Namespace: rcsw.serviceMirrorNamespace,
+					Labels: map[string]string{
+						consts.MirroredResourceLabel:  "true",
```
adleong, line 565: I'm curious what we use this label for, if anything?  I ask because we now have two different kind of services that we create: mirrored services and gateway mirrors.  Anywhere that we're using this label selector will now select both kinds of services.  Is that appropriate?

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -585,6 +718,11 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteGatewayDeleted(ev *RemoteGa
 		gatewayNs:   ev.gatewayData.Namespace,
 		clusterName: rcsw.clusterName,
 	})
+
+	if err := rcsw.localAPIClient.Client.CoreV1().Services(rcsw.serviceMirrorNamespace).Delete(rcsw.mirroredResourceName(ev.gatewayData.Name), &metav1.DeleteOptions{}); err != nil {
+		rcsw.log.Errorf("Could not delete  gateway proxy %s", err)
```
adleong, line 723: extra whitespace in this string 🦅 👁️

controller/cmd/service-mirror/cluster_watcher.go
olix0r: Better yet, add a comment that clarifies what it actually is?

controller/cmd/service-mirror/cluster_watcher.go
olix0r: identities are endpoint-specific and not service-specific, unless we've special-cased this in destination?

adleong: sorry, I mean the resource version annotation.

charts/linkerd2-service-mirror/templates/service-mirror.yaml
zaharidichev: Yes this is correct

cli/cmd/cluster.go
zaharidichev: Yes, this fixes the problem you originally mentioned where gateways were not showing up

controller/cmd/service-mirror/cluster_watcher.go
zaharidichev: So, we use it in `cleanupMirroredResources` where we combine it with the cluster name. This deletes all mirrored services that needs to be deleted when we stop watching a cluster. We also use it in `mirroredServicesForGateway` but the re we combine it with some more specific annotations so there will be no conflict

PR #4441, opened 2020-05-19
Run shellcheck for all shell scripts in repository

joakimr-axis:
Currently, only the shell scripts in `./bin` are scanned by shellcheck for linkerd2 pull requests.
But the repository contains more shell scripts in other directories.

This patch aims to update the shellcheck command in static_checks.yml to search for all files with mimetype `text/x-shellscript` and feed them to shellcheck.

Certainly, this is a tad more time consuming than just scanning one directory, but still a quite fast thing to do while it prevents any new scripts to fly under the radar.

(Also, there is no need to exclude `*.nuspec` or `*.ps1` from the find command as they do not have the text/x-shellscript mimetype.)

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -68,28 +68,10 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
     - name: shellcheck
-      # TODO: Each file listed here is excluded from shellcheck because it
-      # fails. As we fix files we can remove them from this list.
-      #
-      # Once we have this list paired down signficantly, we can switch to
-      # disabling specific checks across all files, for example:
-      # bin/shellcheck bin/* --exclude=SC1000
-      #
       # For more information on shellcheck failures:
       # https://github.com/koalaman/shellcheck/wiki/Checks
       run: |
-        bin/shellcheck -x -P ./bin $(find . -type f \
-        ! -path ./bin/docker-build-proxy \
-        ! -path ./bin/fmt \
-        ! -path ./bin/lint \
-        ! -path ./bin/minikube-start-hyperv.bat \
-        ! -path ./bin/_tag.sh \
-        ! -path ./bin/test-cleanup \
-        ! -path ./bin/_test-run.sh \
-        ! -path ./bin/update-go-deps-shas \
-        ! -path ./cni-plugin/deployment/scripts/install-cni.sh \
-        ! -path ./.git/hooks/\*.sample \
-        | while read -r f; do [ "$(file -b --mime-type "$f")" = 'text/x-shellscript' ] && printf '%s\0' "$f"; done | xargs -0)
+        bin/shellcheck-all
   markdown_lint:
     name: Markdown lint
     runs-on: ubuntu-18.04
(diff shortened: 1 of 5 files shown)


.github/workflows/static_checks.yml
```diff
@@ -78,17 +78,28 @@ jobs:
...
+        ! -wholename ./bin/lint \
+        ! -wholename ./bin/_log.sh \
+        ! -wholename ./bin/minikube-start-hyperv.bat \
+        ! -wholename ./bin/_tag.sh \
+        ! -wholename ./bin/test-cleanup \
+        ! -wholename ./bin/_test-run.sh \
+        ! -wholename ./bin/update-go-deps-shas \
+        ! -wholename ./cni-plugin/deployment/scripts/install-cni.sh \
+        ! -wholename ./.git/hooks/applypatch-msg.sample \
+        ! -wholename ./.git/hooks/commit-msg.sample \
```
olix0r, line 93: Consider excluding the pattern `./.git/hooks/*.sample`?

joakimr-axis: > Consider excluding the pattern `./.git/hooks/*.sample`?

Yes, I was thinking about that but thought I'd put it up like this to see what your comments on this would be. Are the samples used? Or are they to be used? Regardless of what, should they be in this state or should they be fixed up? It's more a strategic and philosophical question IMO, for you guys to decide.

If you'd like me to add that exclusion pattern I will fix that, no problem.

olix0r: I assume that all of these files are part of some git infrastructure. We've never touched them:
```
:; git log .git/hooks
:; git rm -f .git/hooks/*.sample
fatal: pathspec '.git/hooks/applypatch-msg.sample' did not match any files
```

joakimr-axis: Ok, let's go for excluding the pattern `./.git/hooks/*.sample` then?

olix0r: @joakimr-axis Yeah, I think that's best. Thanks!

.github/workflows/static_checks.yml
```diff
@@ -78,17 +78,28 @@ jobs:
...
-        ! -name _log.sh \
-        ! -name minikube-start-hyperv.bat \
-        ! -name _tag.sh \
-        ! -name test-cleanup \
-        ! -name _test-run.sh \
-        ! -name update-go-deps-shas \
-        ! -name *.nuspec \
-        ! -name *.ps1 \
-        | xargs -I {} bin/shellcheck -x -P ./bin {}
+        bin/shellcheck -x -P ./bin $(find . -type f \
```
kleimkuhler, line 81: @joakimr-axis What are you thoughts on adding this entire command to its own `bin/shellcheck-all` (or something similar to that name?).

We have a few scripts like this already, where `bin/kind` or `bin/helm` exist, but then there is a helper script that does more with that command (e.g. `bin/helm-build`).

That way, this job could just refer to that script *and* users could run it themselves. Currently, the only way to shellcheck the entire project before pushing a change would be to copy-paste this section of the job.

joakimr-axis: I like that!

.github/workflows/static_checks.yml
joakimr-axis: Done

.github/workflows/static_checks.yml
joakimr-axis: Done

PR #4447, opened 2020-05-20
markdownlint: Use /bin/sh instead of /bin/bash

joakimr-axis:
The nice and clean markdownlint scripts use no bash-specific
functionality. Hence they could be run with /bin/sh instead. On e.g.
Debian-based systems /bin/sh is dash which has 1/10 of bash's footprint.

### DIFF
--- a/bin/markdownlint-all
+++ b/bin/markdownlint-all
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/sh -eu
 
-bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
+bindir=$( cd "${0%/*}" && pwd )
 
 "$bindir"/markdownlint ./*.md
 "$bindir"/markdownlint ./**/*.md --ignore node_modules/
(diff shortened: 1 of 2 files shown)


bin/markdownlint-all
```diff
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/sh -eu
 
-bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
+bindir=$( cd "${0%/*}" && pwd )
```
olix0r, line 3: Could this cause problems if someone adds `bin` to their PATH?

E.g.

```sh
export PATH="$PWD/bin:$PATH"
markdownlint-all
```

joakimr-axis: > Could this cause problems if someone adds `bin` to their PATH?

I did try that, in both bash and dash, and it behaves exactly the same if I use relative or absolute path, or calling `markdownlint-all` with the `linkerd2/bin` directory in `PATH`. Also regardless of which directory I am currently in when calling the command.

This behavior is also the same as what I get with the bash-specific scripts currently on master.

olix0r: Thanks!

Issue #4256, opened 2020-04-13
ADOPTERS.md change doesn't trigger CI. Also improve workflow dependencies.


alpeb:
## Main Problem

When a PR with changes to just `.md` files occurs no CI runs are triggered. This was made so because that happens when a change to `CHANGES.md` is merged which is immediately followed by a release tag which does trigger CI.

We didn't foresee though that when a change to for example solely `ADOPTERS.md` is merged then CI is not triggered and so the docker images for the latest master won't get pushed, which can confuse devs trying to deploy master in a cluster.

## Possible Solution

We could create a new CI workflow that gets triggered by changes to only `.md` files, which builds and pushes the docker images. This would require copying the `docker_build` and `docker_push` jobs that are already duplicated in `cloud_integration.yml` and `release.yml`.

A separate improvement could alleviate that duplication by having this new workflow be triggered upon all master pushes, and having this workflow dispatch an event that would in turn trigger `cloud_integration.yml` or `release.yml`, allowing removing these duped docker steps for there.

kleimkuhler:
> having this workflow dispatch an event that would in turn trigger cloud_integration.yml or release.yml

I am +1 on reducing the duplication of jobs between workflows by using the repository dispatch events! The website update job uses this, and I don't see any reason why we wouldn't be able to use it to trigger workflows within the same job.

PR #4448, opened 2020-05-20
CLI install integration test for Windows

### LINKED ISSUE #4431: ci: Test windows cli
A user in slack reported that there was an issue with the latest `linkerd` binary on Windows. AFAIK, we don't have any automated tests for this platform to, for instance, detect filesystem path problems. How hard is it to test the CLI from Windows.

Pothulapati:
Fixes #4431

This PR aims to write a integration test, which tests the install command in CLI without a cluster.


.github/workflows/release.yml
```diff
@@ -75,6 +75,56 @@ jobs:
         bin/docker-retag-all $TAG master
         bin/docker-push master
   # todo: Keep in sync with `kind_integration.yml`
+  cli_static_integration_tests:
```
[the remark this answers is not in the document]
alpeb: yeah from what I know there's no easy way to share, unless we wanna do some kind of preprocessing.
#4256 is an idea to use events to trigger workflows in sequence and extract all this duplication.

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +65,53 @@ jobs:
...
+      run: docker load < image-archives/cli-bin.tar
+    - name: Install CLI
+      run: |
+        # Copy the CLI out of the local cli-bin container.
+        container_id=$(docker create "$DOCKER_REGISTRY/cli-bin:$TAG")
+        docker cp $container_id:/out/linkerd-windows $HOME/.linkerd
+
+        # Validate the CLI version matches the current build tag.
+        [[ "$TAG" == "$($HOME/.linkerd version --short --client)" ]]
+    - name: Run CLI Integration tests
```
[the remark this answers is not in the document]
Pothulapati: Yep, it dosen't do that. So I just copied the initial code first as it would be pretty similar but on Windows and windows commands.

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,35 @@ jobs:
         name: image-archives
         path: /home/runner/archives
   # todo: Keep in sync with `release.yml`
+  cli_windows_static_integration_tests:
+    name: Integration tests (windows)
```
Pothulapati, line 72: Should we even call this an integration test?

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,35 @@ jobs:
...
+      uses: actions/checkout@722adc6
+    - name: Try to load cached Go modules
+      # actions/cache@v1.1.2
+      uses: actions/cache@70655ec
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Download image archives (Forked repositories)
```
alpeb, line 87: You can get rid of `(Forked repositories)`

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,35 @@ jobs:
...
+      # actions/download-artifact@v1
+      uses: actions/download-artifact@18f0f59
+      with:
+        name: image-archives
+    - name: Install CLI
+      run: |
+        # Copy the CLI out of the $ARCHIVES/cli folder.
+        cp image-archives\cli\windows\linkerd $HOME\linkerd.exe
```
alpeb, line 95: The code is self-explanatory so no need of the comment (also for `go test ` below).

.github/workflows/kind_integration.yml
alpeb: Given we're testing the final product, the integration test appellation doesn't clash with me :man_shrugging:

Pothulapati:
@alpeb Okay, Looks like the problem here is that the `deep integration tests` run all the tests from `\test` including the cli ones, which don't accept flags like `--k8s-context` as they are not used. I think we have the following options here.

- Move all integration tests into `\test\integration` and update that path in `_test_run.sh` which should make everything work.
- Add a condition in `_test_run.sh` to not include that file by having a if condition.

I'm currently in the favour of 1st way, as it would add more flexibility in adding more kinds of tests into `\test` i.e not only integration tests.

alpeb:
@Pothulapati Moving the integrations tests to its own separate dir sounds fine to me. I'm CCing @kleimkuhler since he's also currently working on those files.

kleimkuhler:
@alpeb Thanks! Those changes are in #4593 which is ready for review. Not sure which will merge first, but moving the location of the test files sounds good.

Pothulapati:
Created a new PR at https://github.com/linkerd/linkerd2/pull/4653 as this whole branch went out of hand with so many commits and rebases in between!

PR #4464, opened 2020-05-21
Test CI by using bin/kind instead of action

kleimkuhler:
Testing CI failures by creating kind clusters using bin script instead of GitHub action


kleimkuhler:
This is similar to how we use `bin/markdownlint-all` in the lint job instead of the GitHub action itself.

This also allows us to keep a single version location in `bin/kind`, so I think this is a good change whether it fixes the flakiness or not.

Issue #4461, opened 2020-05-21
multicluster: support per-mirror credentials


olix0r:
#4457 raises an issue I didn't notice -- `setup-remote` currently creates an SA/RBAC to be used by all remote gateways. This makes it impossible to revoke access to any one cluster. It's all-or-nothing. It also creates some unnecessary install-time-concerns.

Can we change the way that Service Accounts are allocated to do this at link-time rather than at install-time?

Basically, something like:

```
# Create new SA/RBAC for the _west_ service-mirror to discover services in the _east_ cluster.
linkerd --context=east multicluster allow west |
    kubectl apply -f - --context=east

# Get the raw SA and prepare for use in the _west_ cluster
linkerd --context=east multicluster link east |
    kubectl apply -f - --context=west
```

This could also support reusing credentials for folks who are unconcerned:
```
linkerd --context=east multicluster allow all-the-mirrors |
    kubectl apply -f - --context=east

linkerd --context=east multicluster link all-the-mirrors |
    kubectl apply -f - --context=west1
linkerd --context=east multicluster link all-the-mirrors |
    kubectl apply -f - --context=west2
```

Note that `link` should support a `--cluster-server` flag to override the value in the kubeconfig (for private networks, primarily).

olix0r:
To be clear, I think the above should replace the current `setup-remote` and `get-credentials` behavior.

grampelberg:
```yaml
# linkerd multicluster allow -h
Output Kubernetes configs to allow a mirror to connect to this cluster.

Usage:
  linkerd multicluster allow [flags] [NAME]

Examples:
    # Add a SA using the default name
    linkerd --context=east multicluster allow | kubectl --context=east apply -f -

    # Specify the SA, ClusterRole and ClusterRoleBinding names
    linkerd --context=east multicluster allow foobar | kubectl --context=east apply -f -

Flags:
    --namespace       The destination namespace for the service account. (default "linkerd-multicluster")
    --ignore-cluster   Ignore cluster configuration (default false).
```

Note: `allow` should exit if a SA of the name trying to be created already exists

```yaml
# linkerd multicluster link -h
Output Kubernetes configs to link a mirror on one cluster with another.

Usage:
  linkerd multicluster link [flags] [NAME]

Examples:
    # Get the default credentials
    linkerd --context=east multicluster link | kubectl --context=west apply -f -

    # Specify the account
    linkerd --context=east multicluster link foobar | kubectl --context=west apply -f -

[...]

PR #4466, opened 2020-05-22
Add allow and link commands

zaharidichev:
This change adds a `allow` and `link` commands, effectivelly enabling a cluster to have more than one set of credentials that allow it to be mirrored. 

Fx #4461


cli/cmd/cluster.go
```diff
@@ -387,11 +509,10 @@ func newGetCredentialsCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&opts.namespace, "namespace", defaultMulticlusterNamespace, "the namespace in which the secret will be created")
-	cmd.Flags().StringVar(&opts.serviceAccountName, "service-account-name", defaultServiceAccountName, "the name of the service account")
-	cmd.Flags().StringVar(&opts.serviceAccountNamespace, "service-account-namespace", defaultMulticlusterNamespace, "the namespace in which the svc account resides on the remote cluster")
-	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", defaultClusterName, "cluster name")
+	cmd.Flags().StringVar(&opts.namespace, "namespace", defaultMulticlusterNamespace, "The namespace for the service account")
+	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "", "cluster name")
```
adleong, line 513: This starts to get confusing because there are several names in play now: 
* the service account name, which is also called the remote cluster name, which is represents the source cluster
* the cluster name, which represents the destination cluster, (the destination cluster is often referred to as the "remote cluster").

So a typical command might be 

```
linkerd --context west cluster link east --cluster-name west | kubectl --context east apply -f -
```

Is anyone going to get all those contexts and cluster names and service account names right?  It's very confusing which things refers to which thing.

At the very least we need to differentiate between service account names and cluster names.

```
linkerd --context west cluster allow east-sa | kubectl --context west apply -f -

linkerd --context west cluster link east-sa --cluster-name west | kubectl --context east apply -f -
```

This is a little better, but still pretty confusing.  

Better yet would be to give the cluster a fixed name at multicluster install time so that this doesn't get mixed up with the service account dance at all.

```
linkerd --context west cluster install west | kubectl --context west apply -f -

linkerd --context west cluster allow east | kubectl --context west apply -f -

linkerd --context west cluster link east | kubectl --context east apply -f -
```

Pothulapati:
Running `linkerd cluster install --gateway=false` gives two `linkerd-multicluster` namespace manifests as
```yaml
tarun@tarun-Inspiron-5559:work/l5d-az-multi ‹master*›$ linkerd cluster install --gateway=false
kind: Namespace
apiVersion: v1
metadata:
  name: linkerd-multicluster

---
kind: Namespace
apiVersion: v1
metadata:
  name: linkerd-multicluster
---
``

File: cli/cmd/cluster.go [hunk 2]
```diff
@@ -107,42 +120,161 @@ func getLinkerdConfigMap() (*configPb.All, error) {
...
+				return err
+			}
+			stdout.Write(buf.Bytes())
+			stdout.Write([]byte("---\n"))
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&opts.namespace, "namespace", defaultMulticlusterNamespace, "The destination namespace for the service account.")
```
adleong, line 271: Just to be super clear about this: this flag only controls the namespace that the service account is created into.  It does NOT restrict the RBAC of the service account to this namespace.  

In other words, this really doesn't do anything different than what `multicluster install` does.  @olix0r @grampelberg does this match your expectations?  I don't really understand in what situation a user would use `multicluster allow`.

cli/cmd/cluster.go
```diff
@@ -231,29 +360,48 @@ func newSetupRemoteCommand() *cobra.Command {
...
-	opts := getCredentialsOptions{}
+func newLinkCommand() *cobra.Command {
+	opts := linkOptions{}
 
 	cmd := &cobra.Command{
 		Hidden: false,
-		Use:    "get-credentials",
-		Short:  "Get cluster credentials as a secret",
+		Use:    "link",
+		Short:  "Outputs a Kubernetes secret containing the credential that can allow a service mirror component to connect to a remote cluster",
```
adleong, line 397: Maybe something like this:

"Outputs a Kubernetes secret that allows a service mirror component to connect to this cluster"

cli/cmd/cluster.go
```diff
@@ -339,18 +457,22 @@ func newGetCredentialsCommand() *cobra.Command {
...
 			config.AuthInfos = map[string]*api.AuthInfo{
-				opts.serviceAccountName: {
+				accountName: {
 					Token: string(token),
 				},
 			}
 
 			cluster := config.Clusters[context.Cluster]
 
+			if opts.remoteClusterServer != "" {
```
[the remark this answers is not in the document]
adleong: @zaharidichev can you explain this to me?

cli/cmd/cluster.go
```diff
@@ -358,11 +510,11 @@ func newGetCredentialsCommand() *cobra.Command {
...
-	cmd.Flags().StringVar(&opts.serviceAccountName, "service-account-name", defaultServiceAccountName, "the name of the service account")
-	cmd.Flags().StringVar(&opts.serviceAccountNamespace, "service-account-namespace", defaultMulticlusterNamespace, "the namespace in which the svc account resides on the remote cluster")
-	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", defaultClusterName, "cluster name")
-	cmd.Flags().StringVar(&opts.remoteClusterDomain, "remote-cluster-domain", defaultClusterDomain, "custom remote cluster domain")
+	cmd.Flags().StringVar(&opts.namespace, "namespace", defaultMulticlusterNamespace, "The namespace for the service account")
+	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "", "Cluster name")
+	cmd.Flags().StringVar(&opts.remoteClusterDomain, "remote-cluster-domain", defaultClusterDomain, "Custom remote cluster domain")
+	cmd.Flags().StringVar(&opts.remoteClusterServer, "cluster-server", "", "Custom remote cluster domain")
+	cmd.Flags().StringVar(&opts.serviceAccountName, "service-account-name", defaultServiceAccountName, "The name of th service account associated with the credentials")
```
adleong, line 517: typo "th"

cli/cmd/cluster.go
```diff
@@ -358,11 +510,11 @@ func newGetCredentialsCommand() *cobra.Command {
...
 	}
 
-	cmd.Flags().StringVar(&opts.namespace, "namespace", defaultMulticlusterNamespace, "the namespace in which the secret will be created")
-	cmd.Flags().StringVar(&opts.serviceAccountName, "service-account-name", defaultServiceAccountName, "the name of the service account")
-	cmd.Flags().StringVar(&opts.serviceAccountNamespace, "service-account-namespace", defaultMulticlusterNamespace, "the namespace in which the svc account resides on the remote cluster")
-	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", defaultClusterName, "cluster name")
-	cmd.Flags().StringVar(&opts.remoteClusterDomain, "remote-cluster-domain", defaultClusterDomain, "custom remote cluster domain")
+	cmd.Flags().StringVar(&opts.namespace, "namespace", defaultMulticlusterNamespace, "The namespace for the service account")
+	cmd.Flags().StringVar(&opts.clusterName, "cluster-name", "", "Cluster name")
+	cmd.Flags().StringVar(&opts.remoteClusterDomain, "remote-cluster-domain", defaultClusterDomain, "Custom remote cluster domain")
```
adleong, line 515: `multicluster link` is run on the target cluster, so can't we just get the target cluster's domain from the linkerd configmap?

cli/cmd/cluster.go
olix0r: Perhaps we should call it --service-account-namespace?

adleong: To me the bigger question is: what is the purpose of this command?

If it is to be able to create a SA with namespace scoped RBAC then additional work is needed to add that functionality.

As currently written, I don't think it does anything different than what `multicluster install` does.

cli/cmd/cluster.go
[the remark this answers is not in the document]
alpeb: `--cluster-server` is for overriding the k8s api server url to avoid this:
https://github.com/olix0r/l2-k3d-multi/blob/f64de67f0a7766b04f0b288b8e0128363fd2b2ed/link.sh#L12-L19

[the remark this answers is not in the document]
adleong: Oh, this is the address of the target cluster's kubernetes API server.  Gotcha.

Can we maybe call this `--api-server-address` or something?  I had no idea what `cluster-server` meant.  The flag description "Custom remote cluster domain" is also incorrect, I think.

cli/cmd/cluster.go
zaharidichev: yes I am not exactly sure. This is mainly the result of the proposed install flow. I do not think it is a command that adds much value... ¯\_(ツ)_/¯

charts/linkerd2-multicluster/README.md
```diff
@@ -0,0 +1,39 @@
...
+|`identityTrustDomain`            | Trust domain used for identity of the existing linkerd installation                         |`cluster.local`                               |
+|`linkerdNamespace`               | The namespace of the existing Linkerd installation                                          |`linkerd`                                     |
+|`linkerdVersion`                 | Control plane version                                                                       | latest version                               |
+|`namespace`                      | Service Mirror component namespace                                                          |`linkerd-multicluster`                      |
+|`proxyOutboundPort`              | The port on which the proxy accepts outbound traffic                                        |`4140`                                        |
+|`remoteAccessServiceAccountName` | The name of the service account used to allow remote clusters to mirror local services      |`linkerd-service-mirror-remote-access-default`|
```
alpeb, line 34: should be `remoteMirrorServiceAccountName`

charts/linkerd2-multicluster/templates/service-mirror.yaml
```diff
@@ -1,3 +1,4 @@
+{{if .Values.serviceMirror -}}
 ---
 kind: Namespace
 apiVersion: v1
```
alpeb, line 4: The namespace yaml in service-mirror.yaml can be removed now that `namespace.yaml` is being included

PR #4468, opened 2020-05-22
Separate multi-cluster checks and run after add-ons

Pothulapati:
This change makes the multicluster checks run after that of add-ons. Also, separate multicluster check stuff into `healthcheck_multicluster.go` making it more cleaner.

```yaml
linkerd-addons
--------------

linkerd-grafana
---------------
√ grafana pod is running

linkerd-multicluster
--------------------
√ service mirror controller is running
√ service mirror controller ClusterRoles exist
√ service mirror controller ClusterRoleBindings exist
√ service mirror controller Roles exist
√ service mirror controller RoleBindings exist
√ service mirror controller ServiceAccounts exist
√ service mirror controller has required permissions
√ service mirror controller can access remote clusters
√ clusters share trust anchors
```

As all the add-on tests are `warning=true`, even if something fails it should still run tests that come after!


Pothulapati:
@adleong IMO, add-on checks would involve for things like grafana, prometheus, etc which could be more imp than multi-cluster (prometheus is). So, users could intuitively expect control-plane -> add-ons -> multi-cluster.

adleong:
I'm not convinced that prometheus/grafana is necessarily more important than multicluster.  In fact, multicluster itself is also an addon (of a sort).

Anyway, this seems fine. I don't see this order as any better or worse than what we had previously.

PR #4483, opened 2020-05-27
Add multicluster daisy chain check

adleong:
A mirror-service is one that has been created by the mirror service controller and resolves to a gateway in another cluster.  If a mirror service is exported (and thus mirrored into another cluster) this creates a "daisy chain" where requests can come in to the cluster through the local gateway and be immediately sent out of the cluster to a remote gateway.  If the remote gateway is in the source cluster, this can create an infinite loop.

Similarly, if an exported service routes to a mirror service by a traffic split, the same daisy chain effect occurs.

One example where this can come up is with multicluster fail-over.  If both clusters simultaneously fail-over even a portion of their traffic, a loop is created.

We add a check that detects either of the above conditions and warns of the existence of a daisy chain.

### DIFF
--- a/pkg/healthcheck/healthcheck_multicluster.go
+++ b/pkg/healthcheck/healthcheck_multicluster.go
@@ -10,6 +10,7 @@ import (
 
 	pb "github.com/linkerd/linkerd2/controller/gen/public"
 	sm "github.com/linkerd/linkerd2/pkg/servicemirror"
+	tsclient "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/split/clientset/versioned"
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/linkerd/linkerd2/pkg/k8s"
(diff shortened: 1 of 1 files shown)


pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -143,6 +144,14 @@ func (hc *HealthChecker) multiClusterCategory() category {
 					return hc.checkRemoteClusterAnchors()
 				},
 			},
+			{
+				description: "multicluster daisy chaining is avoided",
```
zaharidichev, line 148: TIOLI: I think we can rename that to something that does not imply that this is an active step that avoids daisy chaining such as: "no multicluster daisy chaining"

adleong: For clarity and consistency, checks should always be stated in the affirmative.  i.e. they should never start with "no".

zaharidichev: Makes sense

PR #4486, opened 2020-05-27
Unhide multicluster command

alpeb:


alpeb:
> LGTM! But should we just remove the `Hidden` field, instead of making it `false`?
> as with other commands we don't explicitly state it.

Makes sense. The less code, the better!

PR #4488, opened 2020-05-27
Wait for KinD nodes to be ready in CI

kleimkuhler:
Recently we have been seeing the following error in `Run integration tests` job
of the `KinD integration` workflow:

```
...
##[error]Process completed with exit code 2.
```

After some testing in a fork and changing the `linkerd check` to output to
`stout` instead of `/dev/null`, these errors show up periodically:

```
× no unschedulable pods
    linkerd-controller-7b65fdf847-rqqhn: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-destination-77b66ff5b-zg9ht: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-grafana-7c87b95798-h4qbf: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-identity-7c7f4694bf-d2wwb: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-prometheus-d479547f-j2b2l: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-proxy-injector-885845b8-xk974: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-sp-validator-7d6f48c48c-ctjkt: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-tap-74c6649ffd-x245h: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    linkerd-web-6cdf6db9f6-2bbcg: 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
    see https://linkerd.io/checks/#l5d-existence-unschedulable-pods for hints
```

This is fairly simple to reproduce locally; the kind control plane node is not
immediately ready after `bin/kind create cluster`.

The solution is to wait for all nodes to be ready. Currently, there is only ever
one node, but as this will likely change in the future this covers all cases.

So far this has only happened on forks--so while this may not be the only
issue--it is at least one.

### DIFF
--- a/bin/_test-run.sh
+++ b/bin/_test-run.sh
@@ -121,7 +121,7 @@ check_linkerd_binary(){
         exit 1
     fi
     exit_code=0
-    "$linkerd_path" version --client > /dev/null 2>&1
+    "$linkerd_path" version --client 2>&1
     exit_on_err 'error running linkerd version command'
     printf '[ok]\n'
 }
@@ -173,13 +173,13 @@ install_stable() {
     local linkerd_path=$tmp/.linkerd2/bin/linkerd
     local stable_namespace=$1
     local test_app_namespace=$stable_namespace-upgrade-test
-    $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - > /dev/null 2>&1
-    $linkerd_path check --linkerd-namespace="$stable_namespace" > /dev/null 2>&1
+    $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - 2>&1
+    $linkerd_path check --linkerd-namespace="$stable_namespace" 2>&1
 
     #Now we need to install the app that will be used to verify that upgrade does not break anything
     kubectl --context=$k8s_context create namespace "$test_app_namespace" > /dev/null 2>&1
     kubectl --context=$k8s_context label namespaces "$test_app_namespace" 'linkerd.io/is-test-data-plane'='true' > /dev/null 2>&1
-    $linkerd_path inject --linkerd-namespace="$stable_namespace" "$test_directory/testdata/upgrade_test.yaml" | kubectl --context=$k8s_context apply --namespace="$test_app_namespace" -f - > /dev/null 2>&1
+    $linkerd_path inject --linkerd-namespace="$stable_namespace" "$test_directory/testdata/upgrade_test.yaml" | kubectl --context=$k8s_context apply --namespace="$test_app_namespace" -f - 2>&1
 }
 
 # Run the upgrade test by upgrading the most-recent stable release to the HEAD of
--- a/bin/_test-run.sh
+++ b/bin/_test-run.sh
@@ -175,8 +175,8 @@ install_stable() {
     local test_app_namespace=$stable_namespace-upgrade-test
     (
         set -x
-        $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - 2>&1
-        $linkerd_path check --linkerd-namespace="$stable_namespace" 2>&1
+        "$linkerd_path" install --linkerd-namespace="$stable_namespace" | kubectl --context="$k8s_context" apply -f - 2>&1
+        "$linkerd_path" check --linkerd-namespace="$stable_namespace" 2>&1
     )
 
     #Now we need to install the app that will be used to verify that upgrade does not break anything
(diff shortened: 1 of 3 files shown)


kleimkuhler:
I used a default wait time of `300s` because that is what [setup-kind](https://github.com/engineerd/setup-kind#engineerdsetup-kind) uses. We used to this action before switching to `bin/kind`.

kleimkuhler:
This also removes the `> /dev/null` redirects on all linkerd commands in `_test-run.sh` so that the output is no longer hidden. This error and hopefully future ones will be easier to diagnose with the output.

bin/_test-run.sh
```diff
@@ -121,7 +121,7 @@ check_linkerd_binary(){
         exit 1
     fi
     exit_code=0
-    "$linkerd_path" version --client > /dev/null 2>&1
+    "$linkerd_path" version --client 2>&1
```
alpeb, line 124: Maybe it's not necessary to output stuff here given we do `exit_on_err` right next, and so we stay in line with`check_if_k8s_reachable`

bin/_test-run.sh
```diff
@@ -173,13 +173,13 @@ install_stable() {
...
-    $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - > /dev/null 2>&1
-    $linkerd_path check --linkerd-namespace="$stable_namespace" > /dev/null 2>&1
+    $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - 2>&1
+    $linkerd_path check --linkerd-namespace="$stable_namespace" 2>&1
 
     #Now we need to install the app that will be used to verify that upgrade does not break anything
     kubectl --context=$k8s_context create namespace "$test_app_namespace" > /dev/null 2>&1
     kubectl --context=$k8s_context label namespaces "$test_app_namespace" 'linkerd.io/is-test-data-plane'='true' > /dev/null 2>&1
-    $linkerd_path inject --linkerd-namespace="$stable_namespace" "$test_directory/testdata/upgrade_test.yaml" | kubectl --context=$k8s_context apply --namespace="$test_app_namespace" -f - > /dev/null 2>&1
+    $linkerd_path inject --linkerd-namespace="$stable_namespace" "$test_directory/testdata/upgrade_test.yaml" | kubectl --context=$k8s_context apply --namespace="$test_app_namespace" -f - 2>&1
```
alpeb, line 182: How about echoing something before these commands for people to know what this output corresponds to? Maybe the commands itself... I vaguely recall there was a way in bash to both echo and run something in a single go...

kleimkuhler: I'll wrap these in `( set -x ..)`. That will show the command being run.

bin/_test-run.sh
```diff
@@ -173,13 +173,19 @@ install_stable() {
     local linkerd_path=$tmp/.linkerd2/bin/linkerd
     local stable_namespace=$1
     local test_app_namespace=$stable_namespace-upgrade-test
-    $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - > /dev/null 2>&1
-    $linkerd_path check --linkerd-namespace="$stable_namespace" > /dev/null 2>&1
+    (
+        set -x
+        $linkerd_path install --linkerd-namespace="$stable_namespace" | kubectl --context=$k8s_context apply -f - 2>&1
+        $linkerd_path check --linkerd-namespace="$stable_namespace" 2>&1
+    )
```
joakimr-axis, line 180: I would like to suggest adding double quotes for `$linkerd_path` and `$k8s_context` here now that these lines are being altered in other ways anyway (there will be a merge conflict to solve for https://github.com/linkerd/linkerd2/pull/4406 either way but that's a quick update).

bin/_test-run.sh
```diff
@@ -173,13 +173,19 @@ install_stable() {
...
 
     #Now we need to install the app that will be used to verify that upgrade does not break anything
     kubectl --context=$k8s_context create namespace "$test_app_namespace" > /dev/null 2>&1
     kubectl --context=$k8s_context label namespaces "$test_app_namespace" 'linkerd.io/is-test-data-plane'='true' > /dev/null 2>&1
-    $linkerd_path inject --linkerd-namespace="$stable_namespace" "$test_directory/testdata/upgrade_test.yaml" | kubectl --context=$k8s_context apply --namespace="$test_app_namespace" -f - > /dev/null 2>&1
+    (
+        set -x
+        $linkerd_path inject --linkerd-namespace="$stable_namespace" "$test_directory/testdata/upgrade_test.yaml" | kubectl --context=$k8s_context apply --namespace="$test_app_namespace" -f - 2>&1
+    )
```
joakimr-axis, line 188: And here.

bin/_test-run.sh
kleimkuhler: Added the quotes, thanks!

PR #4489, opened 2020-05-27
Fix array handling in bin/fmt

adleong:
Quoting the list of directories passed to `goimports` was causing the list to be interpreted as a single argument which was stopping `bin/fmt` from working.

Instead, use `read` to split the list of directories into an array.

Also fix up incorrect formatting that has crept in while `bin/fmt` has been broken.


bin/fmt
```diff
@@ -1,14 +1,14 @@
...
 
 # go list will exclude the vendor directory.  We use grep to exclude generated
 # code.
-dirs=$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)
+read -ra dirs <<< "$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)"
 # go list will list all subdirectories and goimports acts recursively.  This
 # results in certain files being reported multiple time.  Therefore, we must
 # dedup them.
-files=$(bin/goimports -l "$dirs" | sort -u)
+files=$(bin/goimports -l "${dirs[@]}" | sort -u)
```
joakimr-axis, line 11: Perhaps there is no need to store the go result in the `dirs` variable, but have the `go list` command here on (current) line 11 instead?

bin/fmt
```diff
@@ -1,14 +1,14 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eu
 
 # go list will exclude the vendor directory.  We use grep to exclude generated
 # code.
-dirs=$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)
+read -ra dirs <<< "$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)"
```
joakimr-axis, line 7: Using

`'{{.Dir}}'`

instead of

`\{\{.Dir\}\}`

is perhaps easier to read (and compare with the `go help list` command output), even if the output is of course the same?

bin/fmt
adleong: I don't think that would work, since we would run into the same problem where quoting the output would treat it as a single argument.  I believe that we need to use `read` to split the output of `go list` into an array.

bin/fmt
adleong: good suggestions but I'd like to make the minimal change here to fix the script

bin/fmt
joakimr-axis: Either version (or the previous) seem not to work if the working path contains whitespace.

bin/fmt
joakimr-axis: I did elaborate somewhat to make it work with whitespace paths. Shellcheck usually recommends using mapfile for assigning array from command outputs. Using

`mapfile -t dirs < <(go list -f '{{.Dir}}' ./... | grep -v controller/gen)`

Will handle that.

bin/fmt
joakimr-axis: Using

`mapfile -t files < <(bin/goimports -l "${dirs[@]}" | sort -u)`

will handle whitespace paths properly

bin/fmt
```diff
@@ -1,14 +1,14 @@
...
 # go list will list all subdirectories and goimports acts recursively.  This
 # results in certain files being reported multiple time.  Therefore, we must
 # dedup them.
-files=$(bin/goimports -l "$dirs" | sort -u)
+files=$(bin/goimports -l "${dirs[@]}" | sort -u)
 
 if [ -n "$files" ]; then
```
joakimr-axis, line 13: With the mapfile and arrays suggested, here we would have:

if [ ${#files[@]} -gt 0 ]; then
  for file in "${files[@]}"; do
    echo "$file"
    bin/goimports -d "$@" "$file"
  done
  exit 64
fi

joakimr-axis:
Well, summarizing, the patch with the mapfile use is just a few changes:

```
diff --git a/bin/fmt b/bin/fmt
index b5e15d9f..79412ade 100755
--- a/bin/fmt
+++ b/bin/fmt
@@ -1,17 +1,18 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eu
 
 # go list will exclude the vendor directory.  We use grep to exclude generated
 # code.
-dirs=$(go list -f \{\{.Dir\}\} ./... | grep -v controller/gen)
+mapfile -t dirs < <(go list -f '{{.Dir}}' ./... | grep -v controller/gen)
 # go list will list all subdirectories and goimports acts recursively.  This
 # results in certain files being reported multiple time.  Therefore, we must
 # dedup them.
-files=$(bin/goimports -l "$dirs" | sort -u)
+mapfile -t files < <(bin/goimports -l "${dirs[@]}" | sort -u)
 
-if [ -n "$files" ]; then
-  for file in $files; do
+if [ ${#files[@]} -gt 0 ]; then
+  for file in "${files[@]}"; do
+    echo "$file"
     bin/goimports -d "$@" "$file"
   done
   exit 64

```

joakimr-axis:
...those will both address the issue targeted by this PR, pass shellcheck and work with whitespace paths.

adleong:
I don't have mapfile in my version of bash and others may not have it also.  I've updated the script to read line-by-line which I think should address the whitespace issue.

joakimr-axis:
> I don't have mapfile in my version of bash and others may not have it also. I've updated the script to read line-by-line which I think should address the whitespace issue.

I just tried the latest update here out on a directory with whitespace:

```
$ pwd
/scratch/linkerd2 whitespace
$ bin/fmt 
stat /scratch/linkerd2: no such file or directory
$
```

adleong:
Good catch with the for loop. Checking the array size is non-trivial since the `${#files[@]}` will be unset if the array is empty and we use the -u flag.  Using `-n` isn't totally correct, but should be good enough, especially since we should never have empty strings in that array.

PR #4490, opened 2020-05-27
Fix uname flags for darwin in bin/lint

adleong:
The version of `uname` on Darwin doesn't support the `-o` flag, resulting in an error message when running the `bin/lint` script. 

We add an if-branch to short-circuit the `uname-o` call if running on Darwin.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -13,7 +13,10 @@ targetbin=$rootdir/target/bin
 cd "$rootdir"
 
 exe=
-if [ "$(uname -o)" = Msys ]; then
+if [ "$(uname -s)" = Darwin ]; then
+  # Darwin's uname doesn't support the -o flag so we short circuit here.
+  :;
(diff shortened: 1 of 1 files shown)


bin/lint
```diff
@@ -13,7 +13,10 @@ targetbin=$rootdir/target/bin
 cd "$rootdir"
 
 exe=
-if [ "$(uname -o)" = Msys ]; then
+if [ "$(uname -s)" = Darwin ]; then
+  # Darwin's uname doesn't support the -o flag so we short circuit here.
+  :;
+elif [ "$(uname -o)" = Msys ]; then
```
joakimr-axis, line 19:
```
exe=
[ "$(uname -s)" != Darwin ] && [ "$(uname -o)" = Msys ] && exe=.exe
```

adleong: I think this is logically equivalent and I feel like what I have is a little easier to understand at a glance.

PR #4493, opened 2020-05-27
Fixes #4305

### LINKED ISSUE #4305: No metrics for routes when using ServiceProfile
### What is the issue?

I'm getting this output from `linkerd routes`:

```
✗ linkerd routes -n linkerd deploy/linkerd-prometheus
ROUTE                                SERVICE   SUCCESS       RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
GET /api/v1/query         linkerd-prometheus         -         -             -             -             -
GET /api/v1/query_range   linkerd-prometheus         -         -             -             -             -
GET /api/v1/series        linkerd-prometheus         -         -             -             -             -
[DEFAULT]                 linkerd-prometheus   100.00%   73.1rps          13ms          77ms          95ms
```

I've just started the dashboard up and run `linkerd install-sp | kubectl apply -f -`. I also have this output from `tap`:

```
✗ linkerd tap -n linkerd deploy/linkerd-prometheus
req id=43:43 proxy=in  src=10.4.1.5:33846 dst=10.4.1.8:9090 tls=true :method=POST :authority=linkerd-prometheus.linkerd.svc.cluster.local:9090 :path=/api/v1/query
rsp id=43:43 proxy=in  src=10.4.1.5:33846 dst=10.4.1.8:9090 tls=true :status=200 latency=2733µs
end id=43:43 proxy=in  src=10.4.1.5:33846 dst=10.4.1.8:9090 tls=true duration=44µs response-length=291B
```

### How can it be reproduced?

1. Install edge-20.4.4
1. Start dashboard
1. Run `linkerd install-sp | kubectl apply -f -`
1. Run `linkerd routes -n linkerd deploy/linkerd-prometheus`

alpeb:
Fixed SP route for `POST /api/v1/query`:

```
$ bin/linkerd routes -n linkerd deploy/linkerd-prometheus
ROUTE                                SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
GET /api/v1/query_range   linkerd-prometheus   100.00%   3.9rps           1ms           2ms           2ms
GET /api/v1/series        linkerd-prometheus   100.00%   1.1rps           1ms           1ms           1ms
POST /api/v1/query        linkerd-prometheus   100.00%   3.1rps           1ms          17ms          19ms
[DEFAULT]                 linkerd-prometheus         -        -             -             -             -
```

Also added one missing route for `linkerd-grafana`, realizing afterwards there are
many other ones missing, but not really worth adding them all.

I also removed the routes in `linkerd-controller` for the tap routes
given that's no longer handled in that service.

And the tap service SP was also removed alltogether since nothing was
getting reported.

From what I understand tap requests are now served by the k8s API,
which itself hits our tap pod through a TLS connection that the proxy doesn't inspect.
Is that correct @adleong?

### DIFF
--- a/cli/cmd/testdata/install-sp_default.golden
+++ b/cli/cmd/testdata/install-sp_default.golden
@@ -21,14 +21,6 @@ spec:
     condition:
       method: POST
       pathRegex: /api/v1/ListServices
-  - name: POST /api/v1/Tap
-    condition:
-      method: POST
-      pathRegex: /api/v1/Tap
-  - name: POST /api/v1/TapByResource
-    condition:
-      method: POST
-      pathRegex: /api/v1/TapByResource
   - name: POST /api/v1/Version
     condition:
       method: POST
@@ -61,9 +53,9 @@ metadata:
   namespace: linkerd
 spec:
   routes:
-  - name: GET /api/v1/query
+  - name: POST /api/v1/query
     condition:
-      method: GET
+      method: POST
       pathRegex: /api/v1/query
   - name: GET /api/v1/query_range
     condition:
@@ -93,6 +85,10 @@ spec:
     condition:
       method: GET
       pathRegex: /api/dashboards/uid/.*
+  - name: GET /api/dashboard/{dashboard}
+    condition:
+      method: GET
+      pathRegex: /api/dashboard/.*
   - name: GET /api/datasources/proxy/1/api/v1/series
     condition:
       method: GET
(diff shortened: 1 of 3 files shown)


grampelberg:
@alpeb is there a script we could write to just automate this? The `proto` files should be easy...

alpeb:
> @alpeb is there a script we could write to just automate this? The `proto` files should be easy...

- For `linkerd-controller` we'd have to split up `proto/public.proto` to extract the endpoints that are really tapped; some in there like `/linkerd2.public.Api/Edges` and `/linkerd2.public.Api/Gateways` are only used through a port-forward so they're never tapped.
- For `linkerd-destination` we can indeed use the proto file as-is
- For `linkerd-prometheus`, as far as I could tell, Prometheus doesn't provide a swagger file for their query API
- For `linkerd-grafana`, no swagger spec to be found either

alpeb:
I've created #4513 to track the automation aspect of this as a separate effort.

PR #4510, opened 2020-05-29
Probe manager simplification

zaharidichev:
There are a few notable things happening in this PR: 

- the probe manager has been decoupled from the cluster_watcher. Now its only responsibility is to watch for mirrored gateways beeing created and to probe them. This means that probes are initiated for all gateways no matter whether there are mirrored services being paired
- the number of paired services is derived from the existing services in the cluster rather than being published as a metric by the prober
- there are no events being exchanged between the cluster watcher and the probe manager


zaharidichev:
I will most likely now take a pass at annotation/labels namings

controller/cmd/service-mirror/probe_manager.go
```diff
@@ -2,83 +2,69 @@ package servicemirror
...
 
-// MirroredServiceUnpaired is emitted when a service is no longer mirrored
-type MirroredServiceUnpaired struct {
-	serviceName      string
-	serviceNamespace string
+// GatewayMirrorCreated is observed when a mirror of a remote gateway is created locally
+type GatewayMirrorCreated struct {
 	gatewayName      string
-	gatewayNs        string
+	gatewayNameSpace string
```
adleong, line 28: I don't think we capitalize NameSpace like this anywhere

proto/public.proto
```diff
@@ -491,11 +491,10 @@ message GatewaysTable {
     string namespace = 1;
     string name = 2;
     string cluster_name = 3;
-    uint64 paired_services = 4;
-    bool alive = 5;
-    uint64 latency_ms_p50 = 6;
-    uint64 latency_ms_p95 = 7;
-    uint64 latency_ms_p99 = 8;
+    bool alive = 4;
```
adleong, line 494: this API breaking change will cause some really nasty stuff to happen when upgrading between edges.  better to leave field tag 4 unused.

controller/api/public/gateways.go
```diff
@@ -85,10 +84,6 @@ func processPrometheusResult(results []promResult) map[string]*pb.GatewaysTable_
 				} else {
 					rows[key].Alive = true
 				}
```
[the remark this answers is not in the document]
alpeb: The CLI user may have RBAC permissions to interact with the `linkerd` namespace, but not necessarily with the `linkerd-multicluster` namespace (say this is not the same user that installed the mc controller). The controller on the other hand has guaranteed cluster access, so it might make more sense to keep this in the controller.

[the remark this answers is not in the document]
alpeb: Also this could become a one liner:
```go
rows[key].Alive = value > 0
```

cli/cmd/multicluster.go
```diff
@@ -805,6 +813,41 @@ var (
...
+	}
+
+	for _, svc := range services.Items {
+		if svc.Labels != nil {
+			clusterName := svc.Labels[k8s.RemoteClusterNameLabel]
+			gatewayName := svc.Labels[k8s.RemoteGatewayNameLabel]
+			gatewayNs := svc.Labels[k8s.RemoteGatewayNsLabel]
+			key := fmt.Sprintf("%s-%s-%s", clusterName, gatewayName, gatewayNs)
+
+			value, hasResult := results[key]
```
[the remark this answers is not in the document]
zaharidichev: ahh.. that one is good!

cli/cmd/multicluster.go
```diff
@@ -805,6 +813,41 @@ var (
...
 
+func getNumServicesMap() (map[string]uint64, error) {
+
+	results := make(map[string]uint64)
+	kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+	if err != nil {
+		return nil, err
+	}
+	matchLabels := map[string]string{
+		k8s.MirroredResourceLabel: "true",
```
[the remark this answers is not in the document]
zaharidichev: Yes, I will exclude them. Although we do not care too much as these keys will never be picked.

proto/public.proto
zaharidichev: In light of the recommendation to do all that work in the controller, this api stays the same

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -729,8 +731,10 @@ func (rcsw *RemoteClusterServiceWatcher) updateGatewayMirrorService(spec *Gatewa
 				},
 			}
 		}
-
-		_, err = rcsw.localAPIClient.Client.CoreV1().Services(rcsw.serviceMirrorNamespace).Update(updatedService)
+		
+		endpoints.Annotations[consts.RemoteGatewayIdentity]		 = spec.identity
+		
+			_, err = rcsw.localAPIClient.Client.CoreV1().Services(rcsw.serviceMirrorNamespace).Update(updatedService)
```
alpeb, line 737: Nit: formatting issue in these two lines

cli/cmd/multicluster.go
```diff
@@ -921,7 +872,7 @@ func gatewaysRowToTableRow(row *pb.GatewaysTable_Row, numSvcMap map[string]uint6
 		row.Namespace,
 		row.Name,
 		alive,
-		fmt.Sprint(numPairedServices),
+		fmt.Sprint(row.PairedServices),
```
alpeb: :-O

controller/api/public/gateways.go
```diff
@@ -52,7 +55,34 @@ func buildGatewaysRequestLabels(req *pb.GatewaysRequest) (labels model.LabelSet,
...
+
+	results := make(map[string]uint64)
+	selector := 	fmt.Sprintf("%s,!%s", k8s.MirroredResourceLabel,  k8s.MirroredGatewayLabel)
+	services, err := s.k8sAPI.Client.CoreV1().Services(corev1.NamespaceAll).List(metav1.ListOptions{LabelSelector:selector})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, svc := range services.Items {
+		if svc.Labels != nil {
```
alpeb, line 70: Given we're filtering on `k8s.MirroredResourceLabel`, wouldn't this always be non-nil?

controller/api/public/gateways.go
```diff
@@ -52,7 +55,34 @@ func buildGatewaysRequestLabels(req *pb.GatewaysRequest) (labels model.LabelSet,
...
+			clusterName := svc.Labels[k8s.RemoteClusterNameLabel]
+			gatewayName := svc.Labels[k8s.RemoteGatewayNameLabel]
+			gatewayNs := svc.Labels[k8s.RemoteGatewayNsLabel]
+			key := fmt.Sprintf("%s-%s-%s", clusterName, gatewayName, gatewayNs)
+
+			value := results[key]
+			results[key] = value + 1
```
alpeb, line 77: Can be just `results[key]++`

controller/api/public/gateways.go
```diff
@@ -3,15 +3,17 @@ package public
 import (
 	"context"
 	"fmt"
+	"github.com/linkerd/linkerd2/pkg/k8s"
```
Pothulapati, line 7: nit: Removing this line, would sort and separate out the imports!

charts/linkerd2-multicluster/templates/gateway.yaml
```diff
@@ -96,7 +96,6 @@ metadata:
   namespace: {{.Values.namespace}}
   annotations: 
     mirror.linkerd.io/gateway-identity: {{.Values.gatewayName}}.{{.Values.namespace}}.serviceaccount.identity.{{.Values.linkerdNamespace}}.{{.Values.identityTrustDomain}}
-    mirror.linkerd.io/probe-port: "{{.Values.gatewayProbePort}}"
```
adleong, line 99: is this no longer an annotation because it is now discovered from the ports in the gateway service spec?

pkg/k8s/labels.go
```diff
@@ -423,6 +435,9 @@ const (
 	// GatewayPortName is the name of the incoming port of the gateway
 	GatewayPortName = "incoming-port"
 
+	// ProbePortName is the name of the probe port of the gateway
```
adleong, line 438: Is there context or discussion somewhere about this change that I can read?  I don't really love using port names as part of the API here (or for the incoming-port).  Looking for ports with specific names feels a bit more magical and implicit as opposed to annotations which feel more explicit to me.  However, I don't feel strongly about this.

olix0r: Do we incur any code complexity due to using port names? It seems like a fine approach to me. Annotations have the downside of having to keep the two fields in-sync if you want to change it.

I might be inclined to change the port names to `linkerd-gateway` and `linkerd-gateway-probe` so it's clearer that we are in control of them and it's not arbitrary.  

I would definitely prefer that we _not_ include the `-port` suffix in these port names. We know their ports. What else could they be?

adleong: Code complexity is not an issue here.  It's more that I tend to assume that names are not important (so long as they match up where they need to), but in this case these specific names have a special meaning which feels a little bit surprising to me.  But, again, not something I feel super strongly about.

How do you want to proceed from here, @olix0r?  I can merge this PR and then prepare a follow-up which renames the ports to `linkerd-gateway` and `linkerd-gateway-probe` if that sounds good.

olix0r: @adleong sgtm

Issue #4469, opened 2020-05-22
Warnings in the CLI when KUBECONFIG is not set


alpeb:
When there's no KUBECONFIG available in the environment, a series of repeated errors are shown in the output:
```
$ KUBECONFIG= bin/go-run cli
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
ERRO[0000] failed to set namespace from config context:invalid configuration: no configuration has been provided 
linkerd manages the Linkerd service mesh.

Usage:
  linkerd [command]
...
```
This was introduced in #4291, with the call to `getDefaultNamespace()` when every command gets initialized.
One way to fix this would be call that just once in `root.go` and make the result available to all the commands (like the global flag values). And also have a more user-friendly warning like "there is no kubectl context available, please set KUBECONFIG blah blah"

Matei207:
I'll take this one on @alpeb! I'll make sure the function is called just once globally and put a bit more thought into the log message.

PR #4512, opened 2020-05-29
Add namespace global flag to hold default namespace name (#4469)

Matei207:
### What
---

* When [linkerd/linkerd2#4291](https://github.com/linkerd/linkerd2/pull/4291) was merged in, because all `cli` commands were calling the same function, they were repeatedly displaying the same error log -- see [issue #4469 ](https://github.com/linkerd/linkerd2/issues/4469).

* To fix this, I introduced a global variable `kubeNamespace` whose value is given by `getDefaultNamespace()`, the function previously called in all `cli` subcommands.

* The function should be called once, so no more than one error message should now be displayed

* This PR also changes the error message to provide a bit more context around the issue

### How
---

* Added a global `namespace` flag -- the flag binds to `kubeNamespace`, a global variable used by all subcommands

* To set the default namespace in the command options, each subcommand now assigns the value of `kubeNamespace` as a default (instead of calling the function)

* Because I added a short-form `-n` for the command, I had to edit the short-flag for `--no-color`, which is part of the logs flagset (the two can't co-exist, tests are failing around the generation of code completion i.e `cli/completions.go`).

* For the log/error message, I opted NOT to display the actual error message, but instead specify that a valid `KUBECONFIG` was not used. This is because the main point of error when using the client config loader, the error message would only add on verbosity that does not give any additional context/information (as far as I can tell)

* I wanted to also modify the alpha commands which currently use a hardcoded default namespace, I didn't edit them because I do not know what the intended functionality for those two subcommands are, however, I'd be more than happy to add them as part of this PR.

#### Other options I considered:

Instead of setting a global flag, these are the other options I have considered, and why I chose not to use them:

1. Have `kubeNamespace` as a global variable and set it in the `PreRunE` hook -> this to me was the most sensible choice, however, all command options are created before the `cmd` itself is created, meaning the `PreRunE` function actually runs after the options have been created.

2. Have a `--namespace` global flag and remove all subcommand `--namespace` flags -> to be consistent, I thought only one `--namespace` flag should exist, the global one. This approach again, does not work, the behaviour here is similar to (1): the flag is registered after the options are created, so in this case the `-n` flag does not do anything (except set a default value from the context).

3. Set `kubeNamespace` in the `init()` function instead of relying on the flag to do this -> thought this approach is not very sensible.

#### Manual test output:
```
bin/go-run cli stat deployment/web
Error: StatSummary API error: deployment.apps "web" not found

bin/go-run cli stat deployment/web -n emojivoto
web       1/1    91.15%   1.9rps          18ms          35ms          39ms          2


kubectl config set-context --namespace=emojivoto --current
bin/go-run cli stat deployments
web       1/1    93.64%   1.8rps          19ms          31ms          38ms          2

```

I'm interested to see if there is any way to avoid having the root flag and subcommand flags for `--namespace`.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -47,6 +47,7 @@ var (
 	apiAddr               string // An empty value means "use the Kubernetes configuration"
 	kubeconfigPath        string
 	kubeContext           string
+	kubeNamespace         string
 	impersonate           string
 	impersonateGroup      []string
 	verbose               bool
@@ -98,6 +99,7 @@ var RootCmd = &cobra.Command{
 }
 
 func init() {
+	RootCmd.PersistentFlags().StringVarP(&kubeNamespace, "namespace", "n", getDefaultNamespace(), "Namespace for specified resource")
 	RootCmd.PersistentFlags().StringVarP(&controlPlaneNamespace, "linkerd-namespace", "L", defaultNamespace, "Namespace in which Linkerd is installed [$LINKERD_NAMESPACE]")
 	RootCmd.PersistentFlags().StringVarP(&cniNamespace, "cni-namespace", "", defaultCNINamespace, "Namespace in which the Linkerd CNI plugin is installed")
 	RootCmd.PersistentFlags().StringVar(&kubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests")
@@ -141,7 +143,7 @@ type statOptionsBase struct {
 
 func newStatOptionsBase() *statOptionsBase {
 	return &statOptionsBase{
-		namespace:    getDefaultNamespace(),
+		namespace:    kubeNamespace,
 		timeWindow:   "1m",
 		outputFormat: tableOutput,
 	}
@@ -205,7 +207,7 @@ func getDefaultNamespace() string {
 	ns, _, err := kubeCfg.Namespace()
 
 	if err != nil {
-		log.Errorf("failed to set namespace from config context:%v", err)
+		log.Errorf("could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set")
 		return corev1.NamespaceDefault
 	}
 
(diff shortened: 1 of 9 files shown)


cli/cmd/root.go
```diff
@@ -98,6 +99,7 @@ var RootCmd = &cobra.Command{
 }
 
 func init() {
+	RootCmd.PersistentFlags().StringVarP(&kubeNamespace, "namespace", "n", getDefaultNamespace(), "Namespace for specified resource")
```
adleong, line 102: It's not clear to me how this interacts with the namespace flag created by specific commands.  For example, the `edges` command has a namespace flag which takes its default value from its options struct which is populated by the `kubeNamespace` variable which is populated here.  😵 

I think it would be more straightforward to simply populate the kubeNamespace variable directly here: `kubeNamespace = getDefaultNamespace()` and then allow individual commands to define their own namespace flag.

cli/cmd/root.go
```diff
@@ -47,6 +47,7 @@ var (
 	apiAddr               string // An empty value means "use the Kubernetes configuration"
 	kubeconfigPath        string
 	kubeContext           string
+	kubeNamespace         string
```
adleong, line 50: would it be more accurate to name this variable `defaultNamespace`? perhaps with a comment explaining that it is taken from the kube context?

cli/cmd/root.go
Matei207: That makes a lot of sense, and it is something I was worried about when I decided to go with the global namespace flag.

Think this makes sense though, I thought for some reason initialising the variable in `init` would be frowned upon a bit more than the flag. Will change it asap!

cli/cmd/root.go
Matei207: I was a bit reluctant to change it into `defaultNamespace` although it does make more contextual sense. 

There is already a `defaultNamespace` constant for the default control-plane namespace. I've changed that to be `defaultLinkerdNamespace` though, and then changed `kubeNamespace` to be `defaultNamespace` as suggested!

grampelberg:
@adleong can you take a look at this again?

Issue #4491, opened 2020-05-27
Multicluster Helm Charts


grampelberg:
What is the user flow to install the multicluster helm charts without the CLI?

alpeb:
The installation was tested fine with Helm, as @Pothulapati depicted it in linkerd/website#731.

As for the `linkerd multicluster` subcommands to be performed after the installation:
- `linkerd multicluster link`: it requires retrieving the SA token secret, so no way of doing that through Helm
- `linkerd multicluster export-service`: we might provide some kustomize resources to perform that
- `linkerd multicluster allow`: currently this renders the `remote-access-service-mirror-rbac.yaml` template from the multicluster chart. We could extract it into its own chart to allow doing this through `helm install`

Pothulapati:
@alpeb Agree with the first two cases. 

But for `linkerd mc allow`, we could already do ` helm template --generate-name . --set serviceMirror=false --set gateway=false --set remoteMirrorServiceAccountName=<cluster-name>`

PR #4519, opened 2020-05-31
charts: Support `linkerd mc allow` through Helm

Pothulapati:
As per #4491,

This change adds support for `linkerd mc allow` through Helm by running 
```
helm install linkerd2-mc-southeastasia  linkerd-edge/linkerd2-multicluster--set remoteMirrorServiceAccountName=southeastasia-1 --set gateway=false --set serviceMirror=false --set installNamespace=false --kube-context westus
```

The main issue why this was not supported previously was because of the mc namespace creation conflict. The `linkerd2-multicluster` release which is created already to install multicluster components would own the `linkerd2-multicluster` namespace, which is tried again to be created by the access cred creation helm release. By having a flag to toggle it, This can be fixed.

This install flow also goes well with the UX, as to revoke credentials the relevant helm release can just be deleted. 


Pothulapati:
@zaharidichev Right :)

alpeb:
@Pothulapati would you add an entry to the chart's readme?

PR #4526, opened 2020-06-01
Rename gateway ports

adleong:
We rename the gateway ports `incoming-port` and `probe-port` to `linkerd-gateway` and `linkerd-gateway-probe` respectively.  The `linkerd` prefix makes it more clear that these ports are special and used by the linkerd multicluster machinery.  The `port` suffix is unnecessary since it is already clear that these are ports.


Pothulapati:
Hmm, This actually fails the mc install causing
```bash
The Deployment "linkerd-gateway" is invalid: spec.template.spec.containers[0].ports[1].name: Invalid value: "linkerd-gateway-probe": must be no more than 15 characters
```

Should we revert this? @zaharidichev

zaharidichev:
@Pothulapati I think yes for now. TIL this interesting restriction

PR #4527, opened 2020-06-01
Kleimkuhler/mc term events metrics

kleimkuhler:
Change terminology from local/remote to source/target in events and metrics.

This does not change any variable, function, struct, or field names since
testing is still improving

### DIFF
--- a/controller/cmd/service-mirror/events_formatting.go
+++ b/controller/cmd/service-mirror/events_formatting.go
@@ -65,23 +65,23 @@ func (gtm gatewayMetadata) String() string {
 
 // Events for cluster watcher
 func (rsc RemoteServiceCreated) String() string {
-	return fmt.Sprintf("RemoteServiceCreated: {service: %s, gatewayData: %s}", formatService(rsc.service), rsc.gatewayData)
+	return fmt.Sprintf("TargetServiceCreated: {service: %s, gatewayData: %s}", formatService(rsc.service), rsc.gatewayData)
 }
 
 func (rsu RemoteServiceUpdated) String() string {
-	return fmt.Sprintf("RemoteServiceUpdated: {localService: %s, localEndpoints: %s, remoteUpdate: %s, gatewayData: %s}", formatService(rsu.localService), formatEndpoints(rsu.localEndpoints), formatService(rsu.remoteUpdate), rsu.gatewayData)
+	return fmt.Sprintf("TargetServiceUpdated: {sourceService: %s, sourceEndpoints: %s, targetUpdate: %s, gatewayData: %s}", formatService(rsu.localService), formatEndpoints(rsu.localEndpoints), formatService(rsu.remoteUpdate), rsu.gatewayData)
 }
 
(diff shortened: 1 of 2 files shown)


controller/cmd/service-mirror/events_formatting.go
```diff
@@ -65,23 +65,23 @@ func (gtm gatewayMetadata) String() string {
 
 // Events for cluster watcher
 func (rsc RemoteServiceCreated) String() string {
-	return fmt.Sprintf("RemoteServiceCreated: {service: %s, gatewayData: %s}", formatService(rsc.service), rsc.gatewayData)
+	return fmt.Sprintf("TargetServiceCreated: {service: %s, gatewayData: %s}", formatService(rsc.service), rsc.gatewayData)
```
adleong, line 68: I think "Remote" is actually appropriate here because the perspective is clear.  This is in the logs of the service mirror controller running in the source cluster, so it's clear that remote services are services running in the target cluster.  

Alternatively, "TargetClusterServiceCreated" could work.  To put it more generally, "source" and "target" are adjectives we use for clusters.

controller/cmd/service-mirror/metrics.go
```diff
@@ -67,7 +67,7 @@ func newProbeMetricVecs() probeMetricVecs {
 	latencies := promauto.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name: "gateway_probe_latency_ms",
-			Help: "A histogram of latencies to a remote gateway.",
+			Help: "A histogram of latencies to a target gateway.",
```
adleong, line 70: Consider "to a target cluster gateway" or "to a gateway" or "to a gateway in a target cluster"

controller/cmd/service-mirror/events_formatting.go
kleimkuhler: Good point. Since this is relative to the source cluster, I kept local/remote terminology.

kleimkuhler:
> do we want to change the dashboard labels in multicluster.json to have the sam terminology as part of this PR or we would submit a new one?

@zaharidichev Would changing all the occurrences of `dst_remote_cluster` to `dst_target_cluster` in `multicluster.json` need any additional changes in places that metric labels names are set?

adleong:
The `dst_remote_cluster` label is used in the `multicluster.json` grafana dashboard and needs to be correspondingly updated

kleimkuhler:
@adleong Yep just wanted to make sure updating in `multicluster.json` didn't need any corresponding changes in other files. It doesn't though

grafana/dashboards/multicluster.json
```diff
@@ -965,4 +965,4 @@
     "title": "Linkerd Multicluster",
     "uid": "linkerd-multicluster",
     "version": 1
-}
+}
```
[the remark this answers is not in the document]
kleimkuhler: Hm looks like VS Code formatting removes it with format-on-save. I'll add it back

Issue #4530, opened 2020-06-02
conformance validation: Add new helper method in `testutil` package


mayankshah1607:
## Problem

The `testutil` package offers a set of utility functions and helpers for the integration tests. As a part of the [Conformance Validation](https://github.com/linkerd/linkerd2-conformance/pulls) project ([RFC](https://github.com/linkerd/rfc/pull/24)), we plan on reusing these utility functions to avoid having to reinvent the wheel or make things redundant by re-writing the same helpers in a different package. How this can be done is by calling the `testutil.NewTestHelper()` method which returns a new instance of `testutil.TestHelper`, on which we call the utility functions. 

However, the problem with this approach is that the `testutil.NewTestHelper()` requires you to pass a set of flags that are only useful to the integration tests. For example, the `-integration-tests` flag, which determines if the integration tests should be run, errors out if it is not set to `true`. The `-linkerd` flag, which determines the path of the linkerd2 binary, won't serve any purpose in the conformance tests, as we plan on using Sonbuoy, which pulls up a pod, that will always look at a fixed file path that points to the binary of a particular version which will be mentioned in the [Sonobuoy plugin](https://github.com/linkerd/linkerd2-conformance/blob/66a4852af96ec9e83868ea0ad4610c89917a7cba/sonobuoy/plugin.yaml).
The `-linkerd-namespace` flag, which specifies the namespace of the control plane, is by default set to `l5d-integration`. Moreover, some of the other flags are irrelevant to conformance testing. The Conformance Validation project would most certainly benefit from reusing the `testutil` package without having to depend on these rules or flag options configured specifically for the integration tests.

## How should this be solved?

We propose to have a new helper method `testutil.NewConformanceTestHelper` which is similar to `testutil.NewTestHelper` - it returns a new instance of `testutil.TestHelper`, but instead of depending on flag options, the values of the fields in `testutil.TestHelper` shall be passed as function arguments. This way, not only will we have full control over the configuration options from the conformance testing side while being able to reuse the utility functions from `testutil` package, but will also enhance UX by hiding away irrelevant flag options. This new method shall be used by the [linkerd2-conformance](https://github.com/linkerd/linkerd2-conformance) repo. This will also ensure that the existing integration test workflow does not break.

cc @Pothulapati @grampelberg

mayankshah1607:
[...]

PR #4532, opened 2020-06-02
conformance validation: add new helper to `testutil`

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/4530

- Add new struct `GenericTestHelperOptions`

`GenericTestHelperOptions` defines the `TestHelper` options
that are used along with `NewGenericTestHelper()` method to obtain a `*TestHelper`

- Add new helper function `NewGenericTestHelper`

`NewGenericTestHelper` accepts a `*GenericTestHelperOptions` and returns a new `*TestHelper` from the options provided. This helper was created to be able to reuse the `testutil` package without any hard restrictions 

### DIFF
--- a/testutil/test_helper.go
+++ b/testutil/test_helper.go
@@ -61,12 +61,11 @@ var LinkerdDeployReplicas = map[string]deploySpec{
 	"linkerd-proxy-injector": {1, []string{"proxy-injector"}},
 }
 
-// ConformanceTestHelperOptions defines the TestHelper
-// options required for conformance validation - https://github.com/linkerd/linkerd2-conformance
-// This way the default options may be defined in the linkerd2-conformance repo,
-// and an instance can be passed to NewConformanceTestHelper() to obtain a *TestHelper
+// GenericTestHelperOptions defines the TestHelper options
+// that are used along with `NewGenericTestHelper()` method
+// to obtain a *TestHelper
 // See - https://github.com/linkerd/linkerd2/issues/4530
-type ConformanceTestHelperOptions struct {
+type GenericTestHelperOptions struct {
 	Linkerd            string
 	Namespace          string
 	UpgradeFromVersion string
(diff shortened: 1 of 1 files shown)


mayankshah1607:
@Pothulapati That makes sense! I've updated the names, mind taking a look again? I hope `NewGenericTestHelper` and `GenericTestHelperOptions` sound okay :)

mayankshah1607:
Pinging @alpeb for a other approval! :)

alpeb:
@mayankshah1607 can you try embedding structs so that there's not so much duplication between `TestHelper` and `GenericTestHelperOptions`? It's ok if you have to make all these fields public...

mayankshah1607:
@alpeb Making the fields of the `TestHelper` public would mean refactoring some amount of code as it would render the helper methods (such as `GetVersion`, `GetLinkerdNamespace`, etc ) redundant, because the fields may now be directly accessed. If the redundancy of the fields in `GenericTestHelperOptions` is an issue, how about we remove that struct and instead directly make `NewGenericTestHelper` return an instance of `TestHelper` by accepting the field values as function parameters?

This way, the `NewGenericTestHelper` may be called from anywhere by passing the `TestHelper` values as function arguments. (Similar to how `NewTestHelper` accepts flag arguments instead)

This would not only satisfy our current requirements for https://github.com/linkerd/linkerd2/pull/4448 and https://github.com/linkerd/linkerd2/issues/4530 , but would also ensure that we do not have to deal with refactoring the rest of the integration test code.
 
I've just pushed this change ([this](https://github.com/linkerd/linkerd2/pull/4532/commits/2a5d2f962f414ebd04857804a432c1116abc5c60) commit). Could you please have a look and let me know if this sounds okay? Thanks again :smile:

Pothulapati:
Ideally, we would want the fields to be public but we don't see the need for it yet. (especially considering it would take a huge refactor i.e updating fields everywhere).

IMO, we should go with this PR for now, and come back to make the fields public if its needed in the future. WDYT?

mayankshah1607:
@Pothulapati sounds cool!

PR #4533, opened 2020-06-02
Revert "Rename gateway ports"

zaharidichev:
Unfortunatelly there is a restriction for the number of chars in a port name (who knew...)

```
The Deployment "linkerd-gateway" is invalid: spec.template.spec.containers[0].ports[1].name: Invalid value: "linkerd-gateway-probe": must be no more than 15 characters

```

Reverts linkerd/linkerd2#4526


zaharidichev:
Closing this since it was addressed in #4528

PR #4534, opened 2020-06-02
Use templated value for Gateway cm name

### LINKED ISSUE #4531: Multicluster: linkerd-gateway configmap conflict with multiple gateway's
### What is the issue?

Though we have a `gatewayName` as a value in `linkerd2-multicluster` chart, we don't use it as a name for gateway configmap, Thus causing conflicts when using multiple gateways in the same cluster.

### How can it be reproduced?

Installing two gateway's through Helm

### Possible solution

The `gatewayName` can be used directly with `--config` suffix as the configmap's name. 

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version:
- Host OS:
- Linkerd version:

Pothulapati:
Fixes #4531 

This PR updates the `linkerd-gateway` cm's name to be templated. To allow multiple Gateway installations in the same cluster with different configmaps.

(Installing multiple gateways in the same cluster is possible only through Helm, as the CLI dosen't expose those commands currently.)


charts/linkerd2-multicluster/values.yaml
```diff
@@ -1,6 +1,6 @@
 controllerComponentLabel: linkerd.io/control-plane-component
 controllerImage: gcr.io/linkerd-io/controller
-controllerImageVersion: linkerdVersionValue
+controllerImageVersion: git-91a067c9
```
zaharidichev, line 3: I do not think we want that ;)

Pothulapati:
@zaharidichev Yep!

Installation of multiple gateways in the same cluster is only possible through Helm now, as these fields are not provided as CLI flags currently.

**To install multiple gateways on the same cluster using Helm:**
- Install Gateway-1 in linkerd-mc-1ns
```bash
 helm install linkerd2-gateway-1 ./charts/linkerd2-multicluster --set serviceMirror=false --set remoteMirrorServiceAccount=false --set gatewayName=linkerd-gateway-1 --set namespace=linkerd-mc-1
```
- Installing Gateway-2 in linkerd-mc-2 ns
```bash
 helm install linkerd2-gateway-2 ./charts/linkerd2-multicluster --set serviceMirror=false --set remoteMirrorServiceAccount=false --set gatewayName=linkerd-gateway-2 --set namespace=linkerd-mc-2
```

I also tested installing gateways in the same namespace, and it worked as expected.

PR #4539, opened 2020-06-03
charts: add installNamespace flag to multiCluster Go Values

Pothulapati:
#4519 didn't have the change to include the flag in Go, causing current master's `linkerd mc install` to fail.


Pothulapati:
As the change was focused specially for Helm, I tested it through Helm and was expecting CLI to work straightforwardly.

Having install Integration tests should have not made this go through! I can take that up by raising a issue.

PR #4520, opened 2020-06-01
Make labels of multicluster components consistent

zaharidichev:
This PR fixes the inconsistency in component labels.

Fix #4511

Signed-off-by: Zahari Dichev <zaharidichev@gmail.com>

PR #4540, opened 2020-06-03
Multicluster install integration test

zaharidichev:
This PR adds an install test for multicluster. This is still a draft as we need to agree on what labels to put onto the MC resources (PR: #4520) in order to be able to clean-up after ourselves. Cleanup is currently not happening.

### DIFF
--- a/.github/workflows/kind_integration.yml
+++ b/.github/workflows/kind_integration.yml
@@ -68,7 +68,7 @@ jobs:
   kind_integration_tests:
     strategy:
       matrix:
-        integration_test: [deep, upgrade_stable, upgrade_edge, helm, helm_upgrade, uninstall, custom_domain, external_issuer]
+        integration_test: [deep, upgrade_stable, upgrade_edge, helm, helm_upgrade, uninstall, custom_domain, external_issuer, multicluster]
     needs: [docker_build]
     name: Integration tests (${{ matrix.integration_test }})
     runs-on: ubuntu-18.04
--- a/.github/workflows/kind_integration.yml
+++ b/.github/workflows/kind_integration.yml
@@ -68,7 +68,7 @@ jobs:
   kind_integration_tests:
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


zaharidichev:
Oh and for the tests to succeed this needs to get merged: #4539

alpeb:
Would be extra-:cool: to also include a test for mc install through helm :wink:

adleong:
Since we decided to take no action on #4520, this should be unblocked now, right?

zaharidichev:
> Would be extra-🆒 to also include a test for mc install through helm 😉

@alpeb Agreed. how do you suggest we organize that? Do we want to always install multicluster during the installation tests and do that for both CLI and Helm. Otherwise we will have to to have the following options: 

- cli install
- cli install with mc
- helm install
- helm install with mc

So two additional clusters spun up. Not sure if it is worth the trouble. We can just bake it in an always install MC and just have it either as CLI or Helm installation. WDYT ?

alpeb:
@zaharidichev Agreed, the simpler we can keep the integration test matrix (7 clusters being spawned currently), the better. Doesn't hurt to run the entire integration test suite with MC on IMO.

.github/workflows/kind_integration.yml
```diff
@@ -68,7 +68,7 @@ jobs:
   kind_integration_tests:
     strategy:
       matrix:
-        integration_test: [deep, upgrade, helm, helm_upgrade, uninstall, custom_domain, external_issuer]
+        integration_test: [deep, upgrade, helm, helm_upgrade, uninstall, custom_domain, external_issuer, multicluster]
```
kleimkuhler, line 71: You'll need to add this `release.yaml` line 81 as well!

zaharidichev:
@alpeb @Pothulapati I have changed this to not introduce new clusters but install the muklticluster components as part of most of the tests. This tests the helm installation codepath as well. Feel free to take another look

PR #4543, opened 2020-06-03
Run linkerd-gateway as non-root

cpretzer:
Container-Optimized OS on GKE runs with a [set of read/write rules](https://cloud.google.com/container-optimized-os/docs/concepts/disks-and-filesystem) that prevent the `linkerd-gateway` from starting up.

These changes move the directories that nginx needs to write to `/tmp` and configures the `error_log` to write to `stderr`

### DIFF
--- a/charts/linkerd2-multicluster/templates/gateway.yaml
+++ b/charts/linkerd2-multicluster/templates/gateway.yaml
@@ -10,7 +10,6 @@ metadata:
 data:
   nginx.conf: |-
     error_log stderr;
-    pid /tmp/nginx.pid;
     events {
     }
     stream {
@@ -20,11 +19,6 @@ data:
        }
     }
     http {
-      client_body_temp_path /tmp/client_temp;
-      proxy_temp_path       /tmp/proxy_temp_path;
-      fastcgi_temp_path     /tmp/fastcgi_temp;
-      uwsgi_temp_path       /tmp/uwsgi_temp;
-      scgi_temp_path        /tmp/scgi_temp;
       server {
           listen     {{.Values.gatewayProbePort}};
           location {{.Values.gatewayProbePath}} {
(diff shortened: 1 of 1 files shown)


zaharidichev:
thanks a bunch @cpretzer  Will have never found that out.. I will run some manual test tomorrow but otherwise this LGTM

alpeb:
@cpretzer I've been running tests in GKE the last few days using the default image type `cos` and I haven't hit this issue.... is there another piece that might be causing that?

cpretzer:
It's entirely possible that it's my config. I'll look into it

cpretzer:
thanks @alpeb you're right, it turns out that my cluster was provisioned with some restrictions.

I've removed the configurations for the `/tmp` volume in both the deployment and the nginx config.

The `securityContext` change for the deployment remains.

cpretzer:
Added an `nginx` configuration that ensures only the `/health` location returns a response, denying all other requests

zaharidichev:
@cpretzer I have tested that wil Oliver's script and I the gateway fails to start: 
```
$ kubectl --context=k3d-east logs linkerd-gateway-647f5c4c45-prr6t nginx  -n linkerd-multicluster
2020/06/05 06:48:42 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

Am I doing something wrong ?

cpretzer:
That's the same error that I was getting before adding the changes to use the /tmp directory. I'll take a look

grampelberg:
When you run as non-root, you're hitting issues with the image having root/root be the owner of `/var`. You'll need:

```
pid        /tmp/nginx.pid;

http {
    client_body_temp_path /tmp/client_temp;
    proxy_temp_path       /tmp/proxy_temp_path;
    fastcgi_temp_path     /tmp/fastcgi_temp;
    uwsgi_temp_path       /tmp/uwsgi_temp;
    scgi_temp_path        /tmp/scgi_temp;
...
}
```

PR #4552, opened 2020-06-04
multicluster: checks for misconfigured mirror services

### LINKED ISSUE #4541: Add checks for misconfigured export services

When manually changing the annotations on exported services, you can change the gateway name. It is possible that this gateway either doesn't exist or is misconfigured. Currently `check` appears to be healthy, but no traffic is happening.

- On the source cluster, there should be a failure if a mirrored service does not have any endpoints.
- On the target cluster, there should be an error if an exported service is referencing a gateway that does not exist.

## Open Questions

- Should we validate the target cluster gateway's configuration as well?
- Should `check` fail mirrored services in the source cluster when the gateway is unreachable or misconfigured? We could improve the help text but I'm not sure it is worth the extra effort.

Pothulapati:
Fixes #4541 

This PR adds the following checks
-  if a mirrored service has endpoints. (This includes gateway mirrors too).
-  if an exported service is referencing a gateway that does not exist.

### DIFF
--- a/pkg/healthcheck/healthcheck_multicluster.go
+++ b/pkg/healthcheck/healthcheck_multicluster.go
@@ -152,6 +152,22 @@ func (hc *HealthChecker) multiClusterCategory() category {
 					return hc.checkDaisyChains()
 				},
 			},
+			{
+				description: "all mirrored services have endpoints",
+				hintAnchor:  "l5d-multicluster-services-endpoints",
+				warning:     true,
+				check: func(ctx context.Context) error {
+					return hc.checkIfMirroredServicesHaveEndpoints()
+				},
+			},
+			{
+				description: "gateways exists for all exported services",
+				hintAnchor:  "l5d-multicluster-gateways-exist",
+				warning:     true,
+				check: func(ctx context.Context) error {
+					return hc.checkifGatewaysExistforExportedServices()
+				},
+			},
 		},
 	}
 }
@@ -488,6 +504,57 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
 			return fmt.Errorf("Some gateways are not alive:\n\t%s", strings.Join(deadGateways, "\n\t"))
 		}
 		return &VerboseSuccess{Message: strings.Join(aliveGateways, "\n")}
+	}
+	return &SkipError{Reason: "not checking muticluster"}
+}
+
+func (hc *HealthChecker) checkIfMirroredServicesHaveEndpoints() error {
+	if hc.Options.ShouldCheckMulticluster {
+
+		var servicesWithNoEndpoints []string
+		mirroredServices, err := hc.kubeAPI.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{LabelSelector: k8s.MirroredResourceLabel})
+		if err != nil {
+			return err
+		}
+
+		for _, svc := range mirroredServices.Items {
+			// Check if there is a relevant end-point
+			_, err := hc.kubeAPI.CoreV1().Endpoints(svc.Namespace).Get(svc.Name, metav1.GetOptions{})
+			if err != nil {
+				servicesWithNoEndpoints = append(servicesWithNoEndpoints, fmt.Sprintf("%s.%s of cluster: [%s], gateway: [%s/%s]", svc.Name, svc.Namespace, svc.Labels[k8s.RemoteClusterNameLabel], svc.Labels[k8s.RemoteGatewayNsLabel], svc.Labels[k8s.RemoteGatewayNameLabel]))
+			}
+		}
+
+		if len(servicesWithNoEndpoints) > 0 {
+			return fmt.Errorf("Some mirrored services do not have endpoints:\n\t%s", strings.Join(servicesWithNoEndpoints, "\n\t"))
+		}
+		return nil
+	}
+	return &SkipError{Reason: "not checking muticluster"}
+}
+func (hc *HealthChecker) checkifGatewaysExistforExportedServices() error {
+	if hc.Options.ShouldCheckMulticluster {
+
+		var nonExistentGateways []string
+		exportedServices, err := hc.kubeAPI.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		for _, svc := range exportedServices.Items {
+			if serviceExported(svc) {
+				// Check if there is a relevant gateway
+				_, err := hc.kubeAPI.CoreV1().Services(svc.Annotations[k8s.GatewayNsAnnotation]).Get(svc.Annotations[k8s.GatewayNameAnnotation], metav1.GetOptions{})
+				if err != nil {
+					nonExistentGateways = append(nonExistentGateways, fmt.Sprintf("%s.%s gateway used with service %s.%s", svc.Annotations[k8s.GatewayNameAnnotation], svc.Annotations[k8s.GatewayNsAnnotation], svc.Name, svc.Namespace))
+				}
+			}
+		}
+
+		if len(nonExistentGateways) > 0 {
+			return fmt.Errorf("Some gateways do not exist:\n\t%s", strings.Join(nonExistentGateways, "\n\t"))
+		}
+		return nil
 	}
 	return &SkipError{Reason: "not checking muticluster"}
 }
--- a/pkg/healthcheck/healthcheck_multicluster.go
+++ b/pkg/healthcheck/healthcheck_multicluster.go
@@ -159,7 +159,7 @@ func (hc *HealthChecker) multiClusterCategory() []category {
 					},
 				},
 				{
-					description: "all mirrored services have endpoints",
+					description: "all mirror services have endpoints",
 					hintAnchor:  "l5d-multicluster-services-endpoints",
 					warning:     true,
 					check: func(ctx context.Context) error {
@@ -175,7 +175,7 @@ func (hc *HealthChecker) multiClusterCategory() []category {
 					},
 				},
 				{
-					description: "matching gateway mirrors of all mirrored services exist",
+					description: "matching gateway mirrors of all mirror services exist",
 					hintAnchor:  "l5d-multicluster-mirror-gateways",
 					warning:     true,
 					check: func(ctx context.Context) error {
@@ -183,7 +183,7 @@ func (hc *HealthChecker) multiClusterCategory() []category {
 					},
 				},
 				{
-					description: "remote: all gateways have external Ips",
+					description: "remote: all gateways have external IPs",
 					hintAnchor:  "l5d-multicluster-gateways-exist",
 					warning:     true,
 					check: func(ctx context.Context) error {
@@ -219,7 +219,7 @@ func (hc *HealthChecker) multiClusterCategory() []category {
 					},
 				},
 				{
-					description: "all gateways have external Ips",
+					description: "all gateways have external IPs",
 					hintAnchor:  "l5d-multicluster-gateways-exist",
 					warning:     true,
 					check: func(ctx context.Context) error {
@@ -711,19 +711,19 @@ func (hc *HealthChecker) checkIfAllMirrorServicesHaveGatewayMirrors() error {
 
 func (hc *HealthChecker) checkIfGatewaysHaveExternalIP() error {
 
-	gatewaysWithNoExternalIps, err := getGatewaysWithNoExternalIps(hc.kubeAPI)
+	gatewaysWithNoExternalIPs, err := getGatewaysWithNoExternalIPs(hc.kubeAPI)
 	if err != nil {
 		return err
 	}
-	if len(gatewaysWithNoExternalIps) > 0 {
-		return fmt.Errorf("Some gateways don't have external Ips:\n\t%s", strings.Join(gatewaysWithNoExternalIps, "\n\t"))
+	if len(gatewaysWithNoExternalIPs) > 0 {
+		return fmt.Errorf("Some gateways don't have external IPs:\n\t%s", strings.Join(gatewaysWithNoExternalIPs, "\n\t"))
 	}
 	return nil
 }
 
-func getGatewaysWithNoExternalIps(api *k8s.KubernetesAPI) ([]string, error) {
+func getGatewaysWithNoExternalIPs(api *k8s.KubernetesAPI) ([]string, error) {
 
-	var gatewaysWithNoExternalIps []string
+	var gatewaysWithNoExternalIPs []string
 	services, err := api.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
 	if err != nil {
 		return nil, err
@@ -733,12 +733,12 @@ func getGatewaysWithNoExternalIps(api *k8s.KubernetesAPI) ([]string, error) {
 		if gatewayService(svc) {
 			// check if there is an external IP for the gateway service
 			if len(svc.Status.LoadBalancer.Ingress) <= 0 {
-				gatewaysWithNoExternalIps = append(gatewaysWithNoExternalIps, fmt.Sprintf("%s.%s", svc.Name, svc.Namespace))
+				gatewaysWithNoExternalIPs = append(gatewaysWithNoExternalIPs, fmt.Sprintf("%s.%s", svc.Name, svc.Namespace))
 			}
 		}
 	}
 
-	return gatewaysWithNoExternalIps, nil
+	return gatewaysWithNoExternalIPs, nil
 }
 
 func (hc *HealthChecker) checkRemoteIfGatewaysHaveExternalIP() error {
@@ -761,14 +761,14 @@ func (hc *HealthChecker) checkRemoteIfGatewaysHaveExternalIP() error {
 			continue
 		}
 
-		gatewaysWithNoExternalIps, err := getGatewaysWithNoExternalIps(remoteAPI)
+		gatewaysWithNoExternalIPs, err := getGatewaysWithNoExternalIPs(remoteAPI)
 		if err != nil {
 			offendingClusters = append(offendingClusters, err.Error())
 			continue
 		}
 
-		if len(gatewaysWithNoExternalIps) > 0 {
-			offendingClusters = append(offendingClusters, fmt.Sprintf("%s:\n\t\t%s", cfg.ClusterName, strings.Join(gatewaysWithNoExternalIps, "\n\t\t")))
+		if len(gatewaysWithNoExternalIPs) > 0 {
+			offendingClusters = append(offendingClusters, fmt.Sprintf("%s:\n\t\t%s", cfg.ClusterName, strings.Join(gatewaysWithNoExternalIPs, "\n\t\t")))
 		}
 	}
 
@@ -874,12 +874,12 @@ func (hc *HealthChecker) checkifGatewaysExistforExportedServices() error {
 
 func getNonExistentGateways(api *k8s.KubernetesAPI) ([]string, error) {
 	var nonExistentGateways []string
-	exportedServices, err := api.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
+	services, err := api.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
-	for _, svc := range exportedServices.Items {
+	for _, svc := range services.Items {
 		if serviceExported(svc) {
 			// Check if there is a relevant gateway
 			_, err := api.CoreV1().Services(svc.Annotations[k8s.GatewayNsAnnotation]).Get(svc.Annotations[k8s.GatewayNameAnnotation], metav1.GetOptions{})
--- a/pkg/healthcheck/healthcheck_multicluster.go
+++ b/pkg/healthcheck/healthcheck_multicluster.go
@@ -642,8 +642,8 @@ func (hc *HealthChecker) checkIfMirroredServicesHaveEndpoints() error {
 
 		for _, svc := range mirroredServices.Items {
 			// Check if there is a relevant end-point
-			_, err := hc.kubeAPI.CoreV1().Endpoints(svc.Namespace).Get(svc.Name, metav1.GetOptions{})
-			if err != nil {
+			endpoint, err := hc.kubeAPI.CoreV1().Endpoints(svc.Namespace).Get(svc.Name, metav1.GetOptions{})
+			if err != nil || len(endpoint.Subsets) == 0 {
 				servicesWithNoEndpoints = append(servicesWithNoEndpoints, fmt.Sprintf("%s.%s of cluster: [%s], gateway: [%s/%s]", svc.Name, svc.Namespace, svc.Labels[k8s.RemoteClusterNameLabel], svc.Labels[k8s.RemoteGatewayNsLabel], svc.Labels[k8s.RemoteGatewayNameLabel]))
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +507,54 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+
+		if len(servicesWithNoEndpoints) > 0 {
+			return fmt.Errorf("Some mirrored services do not have endpoints:\n\t%s", strings.Join(servicesWithNoEndpoints, "\n\t"))
+		}
+		return nil
+	}
+	return &SkipError{Reason: "not checking muticluster"}
+}
+func (hc *HealthChecker) checkifGatewaysExistforExportedServices() error {
+	if hc.Options.ShouldCheckMulticluster {
```
zaharidichev, line 536: So what are you checking here, the target or the source cluster? I am not sure this is correct because: 
- gateway mirrors will be created in the namnespace where the mirror controller lives not in the namespace where they live on the target cluster
- not sure that using the `GatewayNameAnnotation` will work because we append the name of the cluster at the end (for gateway mirrors)

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -152,6 +152,22 @@ func (hc *HealthChecker) multiClusterCategory() category {
 					return hc.checkDaisyChains()
 				},
 			},
+			{
+				description: "all mirrored services have endpoints",
```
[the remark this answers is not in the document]
Pothulapati: Sure!

pkg/healthcheck/healthcheck_multicluster.go
Pothulapati: This check is done on the target cluster, to see if there is a gateway service associated with each exported service. We do this by getting the exported service and reading the `gateway-name` & `gateway-namespace` annotation. I did not actually see the cluster-name being appended here (maybe it is on hte source cluster).

Example exported service

```yaml
 kubectl --context westus -n emojivoto get svc voting-svc -oyaml                   apiVersion: v1
kind: Service
metadata:
  annotations:
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{"mirror.linkerd.io/gateway-name":"linkerd-gateway","mirror.linkerd.io/gateway-ns":"linkerd-multicluster"},"creationTimestamp":"2020-06-03T12:06:51Z","name":"voting-svc","namespace":"emojivoto","resourceVersion":"483451","selfLink":"/api/v1/namespaces/emojivoto/services/voting-svc","uid":"0be02047-63a5-4f12-bb5f-7082a01b577f"},"spec":{"clusterIP":"10.0.247.27","ports":[{"name":"grpc","port":8080,"protocol":"TCP","targetPort":8080},{"name":"prom","port":8801,"protocol":"TCP","targetPort":8801}],"selector":{"app":"voting-svc"},"sessionAffinity":"None","type":"ClusterIP"},"status":{"loadBalancer":{}}}
    mirror.linkerd.io/gateway-name: linkerd-gateway
    mirror.linkerd.io/gateway-ns: linkerd-multicluster
  creationTimestamp: "2020-06-03T12:06:51Z"
  name: voting-svc
```

zaharidichev: Ahh I get it now. However in my mind this is a bit confusing. I think the whole multicluster checks need to be relevant only in the context on one cluster target or source. Now we have checks that are checking both the soruce and the target cluster. In my mind we can simplify that by sayin that we only run checks on the source cluster and we check every target cluster that we have config for for the things we want to check. Does that make sense? Similar to how we go about the trust root anchors, we can iterate through all target clusters we are paired with, grab the config, instantiate a k8s api and check whether it is all good on the other side. I think we are fine because we have permission to look at services on target clusters. Does that make sense or you think its  not such a good idea? @grampelberg WDYT ?

Pothulapati: That totally makes sense. My only concern being, If I own the target cluster, I can't directly check if there are any configured services?

Apart from that I'm okay with the proposed change, and let @grampelberg  decide! :)

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -152,6 +152,54 @@ func (hc *HealthChecker) multiClusterCategory() category {
...
```
adleong, line 156: let's call these "mirror services" instead of "mirrored services".

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -152,6 +152,54 @@ func (hc *HealthChecker) multiClusterCategory() category {
...
+			{
+				description: "all gateway mirrors have endpoints",
+				hintAnchor:  "l5d-multicluster-gateways-endpoints",
+				warning:     true,
+				check: func(ctx context.Context) error {
+					return hc.checkIfGatewayMirrorsHaveEndpoints()
+				},
+			},
+			{
+				description: "matching gateway mirrors of all mirrored services exist",
```
adleong, line 172: This description was really hard for me to understand.  What about "all mirror services have a gateway mirror"?

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -152,6 +152,54 @@ func (hc *HealthChecker) multiClusterCategory() category {
...
+				hintAnchor:  "l5d-multicluster-mirror-gateways",
+				warning:     true,
+				check: func(ctx context.Context) error {
+					return hc.checkIfAllMirrorServicesHaveGatewayMirrors()
+				},
+			},
+			{
+				description: "all gateways have external Ips",
```
adleong, line 180: I think we should capitalize the P in IPs

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +539,178 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+		var servicesWithNoEndpoints []string
+		selector := fmt.Sprintf("%s, !%s", k8s.MirroredResourceLabel, k8s.MirroredGatewayLabel)
+		mirroredServices, err := hc.kubeAPI.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{LabelSelector: selector})
+		if err != nil {
+			return err
+		}
+
+		for _, svc := range mirroredServices.Items {
+			// Check if there is a relevant end-point
+			_, err := hc.kubeAPI.CoreV1().Endpoints(svc.Namespace).Get(svc.Name, metav1.GetOptions{})
```
adleong, line 555: Should we also check that the endpoints object is not empty?  i.e. that it contains an address?

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +539,178 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+		var gatewayMirrorsWithNoEndpoints []string
+		gatewayServices, err := hc.kubeAPI.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{LabelSelector: k8s.MirroredGatewayLabel})
+		if err != nil {
+			return err
+		}
+
+		for _, svc := range gatewayServices.Items {
+			// Check if there is a relevant end-point
+			_, err := hc.kubeAPI.CoreV1().Endpoints(svc.Namespace).Get(svc.Name, metav1.GetOptions{})
+			if err != nil {
```
adleong, line 581: ditto  checking for non-empty

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +539,178 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+			// Check if the relevant gateway mirror exists
+			gatewayMirrorName := svc.Labels[k8s.RemoteGatewayNameLabel] + "-" + svc.Labels[k8s.RemoteClusterNameLabel]
+			_, err := hc.kubeAPI.CoreV1().Services(svc.Labels[k8s.RemoteGatewayNsLabel]).Get(gatewayMirrorName, metav1.GetOptions{})
+			if err != nil {
+				nonExistentGatewayMirrors = append(nonExistentGatewayMirrors, fmt.Sprintf("%s.%s", gatewayMirrorName, svc.Annotations[k8s.RemoteGatewayNsLabel]))
+			}
+		}
+
+		if len(nonExistentGatewayMirrors) > 0 {
+			return fmt.Errorf("Some mirrored gateway services of mirrored services don't exist:\n\t%s", strings.Join(nonExistentGatewayMirrors, "\n\t"))
```
adleong, line 615: I'd switch this around to be clearer: "Some mirror services do not have a gateway mirror"

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +539,178 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+			if gatewayService(svc) {
+				// check if there is an external IP for the gateway service
+				if len(svc.Status.LoadBalancer.Ingress) <= 0 {
+					gatewaysWithNoExternalIps = append(gatewaysWithNoExternalIps, fmt.Sprintf("%s.%s", svc.Name, svc.Namespace))
+				}
+			}
+		}
+
+		if len(gatewaysWithNoExternalIps) > 0 {
+			return fmt.Errorf("Some gateways don't have external Ips:\n\t%s", strings.Join(gatewaysWithNoExternalIps, "\n\t"))
```
adleong, line 641: IPs

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +539,178 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+		}
+	}
+
+	return false
+}
+func (hc *HealthChecker) checkifGatewaysExistforExportedServices() error {
+	if hc.Options.ShouldCheckMulticluster {
+
+		var nonExistentGateways []string
+		exportedServices, err := hc.kubeAPI.CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{})
```
adleong, line 695: this variable name is misleading since this is all services, not just the exported ones.

pkg/healthcheck/healthcheck_multicluster.go
adleong, line 180: It might be nice to somehow make it clearer that the above 3 checks are for the cluster as a source and the next 3 checks are for the cluster as a target.  Not sure what the best way to convey that is, though.

pkg/healthcheck/healthcheck_multicluster.go
adleong: As discussed today, I think doing any checks on the remote cluster is overkill.  However, we should check the health of the local cluster as a source (check that mirror services and mirror gateways are valid) AND as a target (check that exported services and gateways are valid).

pkg/healthcheck/healthcheck_multicluster.go
Pothulapati: Hmmm, but that would not express that the gateway mirror actually exists? Mirror services can have a remote gateway name as annotations that does not exist :thinking:

pkg/healthcheck/healthcheck_multicluster.go
Pothulapati: Hmm, Added an extra check here, as `len(endpoints.Subsets) == 0` but should the warning also be explicit that the endpoints is present but the `Subsets` is empty? 

I currently included that in the no endpoints found case.

Pothulapati:
As discussed yesterday, Now the multicluster checks are separated into `source` and `target` categories. All the checks that were present previously are moved into the source category. Tests that deal with gateways are considered as `target`.

As `source` checks are triggered only if a service-mirror deployment is present, `target` checks are triggered when there is at-least one `linkerd-gateway` deployment present.

When `--multicluster` flag is filed, The check suite will try to run both `source` and `target` checks.

@grampelberg

pkg/healthcheck/healthcheck_multicluster.go
Pothulapati: Hmmmm, :thinking: I remember us concluding that we should do both so that users on the source cluster can also understand directly if there are any problems with the target clusters that they access (as its effect the client services in their cluster).

I updated the code to perform these checks on both sides, but happy to revisit if this was not intended.

pkg/healthcheck/healthcheck_multicluster.go
adleong: perhaps "all mirror services have a valid gateway mirror".  a gateway that doesn't exist wouldn't be considered "valid"

pkg/healthcheck/healthcheck_multicluster.go
adleong: I think that's okay

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -394,8 +498,42 @@ func (hc *HealthChecker) checkDaisyChains() error {
 	return &SkipError{Reason: "not checking muticluster"}
 }
 
+func (hc *HealthChecker) checkClusterGatewaysHealth() error {
+
+	var offendingGateways []string
+	options := metav1.ListOptions{
+		LabelSelector: hc.gatewayComponentsSelector(),
```
adleong, line 505: I don't think we should be relying on this label selector since this couples the check to the specific implementation of the linkerd-gateway.  Instead, I'd consider swapping the order of this check and the next one:

First check that all exported services reference a valid gateway.  If there are 1 or more exported services then enable TargetCluster checking.  

Next, check that all the gateways referenced by the exported services are healthy.  To check health, I think it's sufficient to ensure that the gateway service has non-empty endpoints.  I don't think we need to fetch the deployment at all (in fact, we shouldn't be assuming that the gateway is a deployment, it could be e.g. a daemonset)

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +629,300 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
 	}
 	return &SkipError{Reason: "not checking muticluster"}
 }
+
+func (hc *HealthChecker) checkIfMirroredServicesHaveEndpoints() error {
```
adleong, line 633: MirrorServices

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -451,15 +589,15 @@ func (hc *HealthChecker) checkRemoteClusterConnectivity() error {
 		}
 
 		if len(errors) > 0 {
-			return fmt.Errorf("Problematic clusters:\n\t%s", strings.Join(errors, "\n\t"))
+			return fmt.Errorf("Problematic clusters:\n\t%s", strings.Join(errors, "\n\t\t"))
```
adleong, line 592: Will this print the first problematic cluster indented with one tab but all following clusters with two tabs?

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +629,300 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+		if err != nil {
+			return err
+		}
+
+		for _, svc := range mirrorServices.Items {
+			// Check if the relevant gateway mirror exists
+			gatewayMirrorName := svc.Labels[k8s.RemoteGatewayNameLabel] + "-" + svc.Labels[k8s.RemoteClusterNameLabel]
+			_, err := hc.kubeAPI.CoreV1().Services(svc.Labels[k8s.RemoteGatewayNsLabel]).Get(gatewayMirrorName, metav1.GetOptions{})
+			if err != nil {
+				nonExistentGatewayMirrors = append(nonExistentGatewayMirrors, fmt.Sprintf("%s.%s", gatewayMirrorName, svc.Annotations[k8s.RemoteGatewayNsLabel]))
```
adleong, line 700: If the mirror service is missing the labels then I think this will render as "-." which looks weird.

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +629,300 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+			}
+		}
+
+		if len(nonExistentGatewayMirrors) > 0 {
+			return fmt.Errorf("Some mirror services do not have a valid gateway mirror:\n\t%s", strings.Join(nonExistentGatewayMirrors, "\n\t"))
```
adleong, line 705: This sounds like it will list the mirror services which don't have valid gateway mirrors but instead it lists the nonExistentGatewayMirrors.  I think listing the problematic mirror services is more useful.

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -491,3 +629,300 @@ func (hc *HealthChecker) checkRemoteClusterGatewaysHealth(ctx context.Context) e
...
+	return hasGtwName
+}
+
+func (hc *HealthChecker) checkIfAllGatewaysHaveCorrectPortNames() error {
+	gatewaysWithMisConfiguredPorts, err := getGatewaysWithMisConfiguredPorts(hc.kubeAPI)
+	if err != nil {
+		return err
+	}
+	if len(gatewaysWithMisConfiguredPorts) > 0 {
+		return fmt.Errorf("Some gateway's have misconfigured ports:\n\t%s", strings.Join(gatewaysWithMisConfiguredPorts, "\n\t"))
```
adleong, line 792: no need for apostrophe in "gateways"

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -54,108 +59,207 @@ var expectedServiceMirrorRemoteClusterPolicyVerbs = []string{
...
+				{
+					description: "remote: gateways exists for all exported services",
+					hintAnchor:  "l5d-multicluster-exported-gateways",
+					warning:     true,
+					check: func(ctx context.Context) error {
+						return hc.checkRemoteifGatewaysExistforExportedServices()
+					},
+				},
+				{
+					description: "remote: all gateways have correct port names exposed",
```
adleong, line 202: I think all of these remote checks can be collapsed into a single check that checks the health of exported service on the remote cluster.  For each exported service on the remote cluster we should check that:

* the gateway service it points should exist
* that gateway should have an external IP
* that gateway should have the correct ports defined
* that gateway should have non-empty endpoints

Doing this all in one check makes logical sense to me and should simplify the code since we won't need to repeatedly fetch the list of gateways.

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -54,108 +59,207 @@ var expectedServiceMirrorRemoteClusterPolicyVerbs = []string{
...
-				hintAnchor:  "l5d-multicluster-daisy-chaining",
-				warning:     true,
-				check: func(ctx context.Context) error {
-					return hc.checkDaisyChains()
+		},
+		{
+			id: LinkerdMulticlusterTargetChecks,
+			checkers: []checker{
+				{
+					description: "all cluster gateways are alive",
```
adleong, line 215: My same suggestion about combining the gateway checks applies here as well.  We can do a single check that verifies that each exported service has a valid gateway (where "valid" means that it exists, has an external IP, has the right ports, has non-empty endpoints)

Pothulapati:
@adleong Those are great suggestions!

I've made the following changes:

- Combine all the checks to be done on Gateways (found through exported services) to consider a gateway valid i.e ( should exist, have externalIP, and relevant ports) into `checkGateways` and will be used as `checkRemoteGateways` & `checkLocalGateways` by source and target respectively.
- Target cluster gateways checks are only if there is at-least one exported service present, through `checkIfTargetCluster`

One thing to note is that, we can't check for gateway endpoints through source cluster. and hence that check is not included in `checkGateways` as the creds don't give access to `endpoints` resource. This can be performed in the target cluster by having a separate check.

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -54,102 +58,160 @@ var expectedServiceMirrorRemoteClusterPolicyVerbs = []string{
...
+		{
+			id: LinkerdMulticlusterTargetChecks,
+			checkers: []checker{
+				{
+					check: func(ctx context.Context) error {
+						return hc.checkIfTargetCluster()
+					},
+				},
+				{
+					description: "all  cluster gateways are valid",
```
adleong, line 195: too many spaces between "all" and "cluster"

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -54,102 +58,160 @@ var expectedServiceMirrorRemoteClusterPolicyVerbs = []string{
...
+					warning:     true,
+					check: func(ctx context.Context) error {
+						if hc.TargetCluster {
+							return hc.checkLocalGateways()
+						}
+						return &SkipError{Reason: "not checking target cluster"}
+					},
+				},
+				{
+					description: "all  cluster gateways have endpoints",
```
adleong, line 206: too many spaces between "all" and "cluster"

pkg/healthcheck/healthcheck_multicluster.go
```diff
@@ -394,8 +456,138 @@ func (hc *HealthChecker) checkDaisyChains() error {
...
+			offendingClusters = append(offendingClusters, fmt.Sprintf("%s:\n\t%s", cfg.ClusterName, offendingCluster))
+		}
+	}
+	if len(offendingClusters) > 0 {
+		return fmt.Errorf("Some gateways have problems:\n\t%s", strings.Join(offendingClusters, "\n\t"))
+	}
+	return nil
+}
+
+func checkGateways(api *k8s.KubernetesAPI) (string, error) {
```
adleong, line 509: this could use a comment explaining what it returns

adleong: I think this method would be a lot easier to understand if it simply returned `[]error` instead of `(string, error)`.  

I don't think we need to build up separate slices of different error types (misconfigured ports, nonexistant gateways, etc) we can just append to a single error slice.

adleong:
I've pushed to this PR with my suggestions above.  PTAL.  For reference, here is some sample output for various error cases:

```
× all target cluster gateways are alive
    Some gateways are not alive:
  * cluster: [], gateway: [linkerd-multicluster/linkerd-gateway]
    see https://linkerd.io/checks/#l5d-multicluster-target-gateways-alive for hints

‼ all mirror services have endpoints
    Some mirror services do not have endpoints:
  voting-svc-gke.emojivoto mirrored from cluster [gke] (gateway: [linkerd-multicluster/linkerd-gateway])
    see https://linkerd.io/checks/#l5d-multicluster-services-endpoints for hints

‼ all gateway mirrors have endpoints
    Some gateway mirrors do not have endpoints:
  linkerd-gateway-gke.linkerd-multicluster mirrored from cluster [gke]
    see https://linkerd.io/checks/#l5d-multicluster-gateways-endpoints for hints

‼ remote: all referenced gateways are valid
      * gke: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway that does not exist: linkerd-gateway.linkerd-multicluster
  * gke-two: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway that does not exist: linkerd-gateway.linkerd-multicluster
    see https://linkerd.io/checks/#l5d-multicluster-gateways-exist for hints

‼ remote: all referenced gateways are valid
      * gke: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway that does not exist: linkerd-gateway.linkerd-multicluster
  * gke-two: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway that does not exist: linkerd-gateway.linkerd-multicluster

‼ remote: all referenced gateways are valid
      * gke: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway with no external IP: linkerd-gateway.linkerd-multicluster
  * gke-two: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway with no external IP: linkerd-gateway.linkerd-multicluster

‼ all cluster gateways are valid
      Exported service voting-svc.emojivoto references a gateway that does not exist: linkerd-gateway.linkerd-multicluster
    see https://linkerd.io/checks/#l5d-multicluster-gateways-exist for hints

‼ remote: all referenced gateways are valid
      * gke: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway that is missing port mc-gateway: linkerd-gateway.linkerd-multicluster
  * gke-two: remote cluster has invalid gateways:
    Exported service voting-svc.emojivoto references a gateway that is missing port mc-gateway: linkerd-gateway.linkerd-multicluster
    see https://linkerd.io/checks/#l5d-multicluster-gateways-exist for hints

‼ all cluster gateways are valid
      Exported service voting-svc.emojivoto references a gateway that is missing port mc-gateway: linkerd-gateway.linkerd-multicluster
    see https://linkerd.io/checks/#l5d-multicluster-gateways-exist for hints

‼ all cluster gateways are valid
    Some gateway services do not have endpoints:
  linkerd-gateway.linkerd-multicluster
    see https://linkerd.io/checks/#l5d-multicluster-gateways-exist for hints
```

Note also that I have filed a separate issue for the inconsistent indentation: https://github.com/linkerd/linkerd2/issues/4574

PR #4554, opened 2020-06-04
Fix load balancer missing ip race condition

### LINKED ISSUE #4550: Race between gateway IP and service export
## What is the issue?

If a service is mirrored before a gateway gets its public IP address, it appears the endpoints do not get updated in the source cluster for the mirrored service. Looking at the service-mirror logs, it also appears that probes don't get started either.

```bash
for ctx in east; do
    kubectl --context=${ctx} apply \
      -k "github.com/linkerd/website/multicluster/${ctx}/?ref=grampelberg/multi-split"
    kubectl --context=${ctx} -n test \
      rollout status deploy/podinfo || break
    kubectl --context=${ctx} apply \
        -f https://www.getambassador.io/yaml/aes-crds.yaml
    kubectl --context=${ctx} wait \
        --for condition=established --timeout=90s crd -lproduct=aes
    curl https://www.getambassador.io/yaml/aes.yaml | \
        linkerd --context=${ctx} inject - | \
        kubectl --context=${ctx} apply -f -
    kubectl --context=${ctx} -n ambassador wait \
        --for condition=available --timeout=90s deploy -lproduct=aes
    cat <<EOF | kubectl --context=${ctx} apply -f -
---
apiVersion: getambassador.io/v2
kind: Module
metadata:
  name: ambassador
  namespace: test
spec:
  config:
    add_linkerd_headers: true
---
apiVersion: getambassador.io/v2
kind: Host
metadata: 
  name: wildcard
  namespace: test
spec:
  hostname: "*"
  selector:
    matchLabels:
      nothing: nothing
  acmeProvider:
    authority: none
  requestPolicy:
    insecure:
      action: Route
---
apiVersion: getambassador.io/v2
kind: Mapping
metadata:
  name: podinfo-mapping
  namespace: test
spec:
  prefix: /
  host: podinfo
  service: podinfo:9898
EOF
    kubectl --context=${ctx} -n ambassador patch svc ambassador --type='json' -p='[
        {"op":"add","path":"/spec/ports/-", "value":{"name": "incoming-port", "port": 4180}},
        {"op":"add","path":"/spec/ports/-", "value":{"name": "probe-port", "port": 8877}}
    ]'
    kubectl --context=${ctx} -n ambassador patch svc ambassador -p='
metadata:
    annotations:
        mirror.linkerd.io/gateway-identity: ambassador.ambassador.serviceaccount.identity.linkerd.cluster.local
        mirror.linkerd.io/multicluster-gateway: "true"
        mirror.linkerd.io/probe-path: /ambassador/v0/check_ready
        mirror.linkerd.io/probe-period: "3"
        mirror.linkerd.io/probe-port: "8877"
'
    kubectl --context=${ctx} -n test patch svc podinfo -p='
metadata:
    annotations:
        mirror.linkerd.io/gateway-name: ambassador
        mirror.linkerd.io/gateway-ns: ambassador
'
done
```

### LINKED ISSUE #4553: Missing gateway name and inconsitent num services in `gateways` command
During my testing on some of @grampelberg issues I discovered that we get gateways output such as: 
```
$ bin/linkerd --context=source mc  gateways

CLUSTER  NAMESPACE             NAME             ALIVE    NUM_SVC  LATENCY_P50  LATENCY_P95  LATENCY_P99
         linkerd-multicluster  linkerd-gateway  True           0        150ms        195ms        199ms
```

This is not correct and has been introduced in #4527 where the metrics label was not changed in prometheus.go

zaharidichev:
This PR fixes a problem where a mirror service is not created (and it should) when its paired to a gateway that still does not have ready external IP addresses

Fix #4550

### DIFF
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -491,11 +491,16 @@ func (rcsw *RemoteClusterServiceWatcher) handleRemoteServiceCreated(ev *RemoteSe
 	if err == nil {
 		// only if we resolve it, we are updating the endpoints addresses and ports
 		rcsw.log.Debugf("Resolved gateway [%v:%d] for %s", gatewaySpec.addresses, gatewaySpec.incomingPort, serviceInfo)
-		endpointsToCreate.Subsets = []corev1.EndpointSubset{
-			{
-				Addresses: gatewaySpec.addresses,
-				Ports:     rcsw.getEndpointsPorts(ev.service, int32(gatewaySpec.incomingPort)),
-			},
+
+		if len(gatewaySpec.addresses) > 0 {
+			endpointsToCreate.Subsets = []corev1.EndpointSubset{
+				{
+					Addresses: gatewaySpec.addresses,
+					Ports:     rcsw.getEndpointsPorts(ev.service, int32(gatewaySpec.incomingPort)),
+				},
+			}
+		} else {
+			rcsw.log.Warnf("gateway for %s: %s does not have ready addresses, skipping subsets", serviceInfo, err)
 		}
 		serviceToCreate.Annotations[consts.RemoteGatewayResourceVersionAnnotation] = gatewaySpec.resourceVersion
 		if gatewaySpec.identity != "" {


zaharidichev:
@grampelberg  In order to validate: 

# Install service mirror on source
bin/linkerd --context=source mc install  --gateway=false  --service-mirror-credentials=false --log-level=debug | kubectl --context=source apply -f -

# Install credentials only on target
bin/linkerd --context=target mc install  --service-mirror=false --gateway=false  --service-mirror-credentials=true --log-level=debug | kubectl --context=target apply -f -

# Allow source to see target
bin/linkerd --context=target  multicluster link --cluster-name=target | kubectl --context=source apply -f -

# Deploy this on the target cluster
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: backend-one
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: backend-one
  template:
    metadata:
      labels:
        app: backend-one
      annotations:
        linkerd.io/inject: enabled  
    spec:
      containers:
      - name: backend-one
        image: buoyantio/bb:v0.0.5
        args:
        - terminus
        - "--h1-server-port=8888"
        - "--response-text=hello-remote-1"
        ports:
        - containerPort: 8888
---
apiVersion: v1
kind: Service
metadata:
  name: backend-one-svc
  namespace: default
spec:
  selector:
    app: backend-one
  ports:
  - name: http
    port: 8888
    targetPort: 8888
```

# Now we are going to install the gateway and export the service while the gateway does not have an IP yet and see what happens

bin/linkerd --context=target mc install  --service-mirror=false --gateway=true  --service-mirror-credentials=false --log-level=debug | kubectl --context=target apply -f -
kubectl --context=target get svc backend-one-svc -oyaml | bin/linkerd --context=target multicluster export-service - | kubectl --context=target apply -f -

You will see the mirror service created without endpoints, as well as the gateway mirror. You shold also see the gateway being unhealthy for a while. Once the the LB gets an IP, things will converge to normal.

alpeb:
When the target gateway receives its public IP `createOrUpdateService()` gets called. I see that ends up calling `updateAffectedServices()` which retrieves the mirrored services to update their endpoints. Why wasn't this working before, and how does not creating the endpoint subsets is supposed to fix things now?

zaharidichev:
@olix0r It is not. I will add one.

@alpeb  Here is what happend when @grampelberg was testing. 

- A gateway was created (no IP yet)
- a service was exported
- the servicem mirror sees there is a new gateway and creates a gateeway mirror for it without endpoints subsets as  there are no IPs.. So far so good
- the service mirror sees a service was exported, resolved its gateway to the one mentioned above and tries to create a mirror service with the subsets. However it fails to check that the addresses are empty so it tries to create Endpoints with Subsets that have empty addresses. This fails a k8s api does not allow that. so the service is not mirrored now..
- now on the source cluster we have a gateway mirror and 0 services  mirrored because the service could not be mirrored...
- the gateway gets updated with the ip, the endpoints of the gateway mirror get updated as well but no services are updated because there are no mirrored services for this gateway...

This is invalid state. So what we do is that we guard against empty addresses now on both gateway mirrors and service mirrors and if athe addresses are empty we have null endpoints. 
Interesting thing is that @grampelberg mentioned that he removed and added the export annotation on the service on the target cluster and that fixed things. That makes sense. Alternatively this situation would have been fixed during the next resync (10m time). 

Does that make sense?

alpeb:
Gotcha, I wasn't aware of that k8s call failure. Makes sense. I'll test a bit more before :heavy_check_mark: this one up

adleong:
Two questions:

1. When the subset is empty the endpoints object will fail to be created, but won't the mirror service object still be created?  And so wouldn't that service still be associated with the gateway such that when the gateway is updated with a valid IP the endpoints will be created?

 2. After this change, this situation will lead to the Subsets array being empty (which is valid?) as opposed to the Subsets array containing one subset with an empty list of addresses (which is invalid?).  Is that correct?

zaharidichev:
@adleong 


> When the subset is empty the endpoints object will fail to be created, but won't the mirror service object still be created? And so wouldn't that service still be associated with the gateway such that when the gateway is updated with a valid IP the endpoints will be created?

So no, if we fail to create the endpoints we go back and delete the service we have created as well just to clean things up because we assume that if the endpoints did not get created due to an error, such as a k8s API error it does not make much sense to have the service hanging around. So when this whole initial failure happened there was no service and no endpoints created so nothing to update. Maybe that is not the moest correct behaviour but I figured I want to clean up after myself if I fail midthrough. 

> After this change, this situation will lead to the Subsets array being empty (which is valid?) as opposed to the Subsets array containing one subset with an empty list of addresses (which is invalid?). Is that correct?

Yes that is correct.

Issue #4556, opened 2020-06-04
Upgrading from edge version does not set Grafana control plane version corrrectly


kleimkuhler:
This only happens when the server version is edge. If the server version is stable, this does not happen.

## Reproduction:

```bash
❯ linkerd version
Client version: edge-20.6.1
```

```
❯ linkerd install |kubectl apply -f -
```

Use new version of linkerd. In this example I just pull the current master:

```
❯ bin/docker-pull-binaries master
..
❯ ./target/release/linkerd2-cli-master-darwin version
Client version: git-108b383a
Server version: edge-20.6.1
```

Now, upgrade the control plane and examine the output.

```
❯ ./target/release/linkerd2-cli-master-darwin upgrade control-plane | rg 'version: edge-20.6.1'
        version: edge-20.6.1
```

## Expected

This should display `git-654839e6`. The `grep` command should not have matched
anything

```yaml
..
data:
  values: |-
    global:
      grafanaUrl:
    grafana:
      enabled: true
      image:
        name: gcr.io/linkerd-io/grafana
        version: edge-20.6.1
      name: linkerd-grafana
    tracing:
      enabled: false
..
```

kleimkuhler:
This doesn't happen when upgrading with manifests:

```
kubectl get -n linkerd cm/linkerd-config secrets/linkerd-identity-issuer -o yaml |./target/release/linkerd2-cli-git-654839e6-darwin upgrade control-plane --from-manifests -
```

```yaml
..
data:
  values: |-
    global:
      grafanaUrl:
    grafana:
      enabled: true
      image:
        name: gcr.io/linkerd-io/grafana
        version: git-654839e6
      name: linkerd-grafana
    tracing:
      enabled: false
..
```

kleimkuhler:
This should be okay. The difference between it being stable and edge is from the add-on side of things. Upgrading from edge and passing the `--addon-overwrite` fixes this.

PR #4557, opened 2020-06-04
Add integration test for upgrading from edge

### LINKED ISSUE #4471: Can't upgrade Linkerd from edge 20.5.3 to 20.5.4 due to render error
### What is the issue?
When upgrading from 20.5.3 to 20.5.4, I am getting the following error after running `linkerd upgrade`:
```
× Could not render upgrade configuration: render error in "grafana/templates/grafana.yaml": template: grafana/templates/grafana.yaml:124:24: executing "grafana/templates/grafana.yaml" at <.Values.image.name>: can't evaluate field name in type interface {}
```

### How can it be reproduced?
```
curl https://run.linkerd.io/install-edge | sh -
linkerd upgrade --context $CLUSTER --ha | kubectl --context $CLUSTER apply -f -
```

### Logs, error output, etc

See above

#### `linkerd check` output

```text
$ linkerd check
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
‼ control plane is up-to-date
    is running version 20.5.3 but the latest edge version is 20.5.4
    control plane running edge-20.5.3 but cli running edge-20.5.4

linkerd-ha-checks
-----------------

linkerd-addons
--------------
√ linkerd-config-addons config map exists

linkerd-grafana
---------------
√ grafana pod is running

Status check results are √
```

### Environment

- Kubernetes Version: v1.17.0
- Cluster Environment: (GKE, AKS, kops, ...) AKS
- Host OS: Mac
- Linkerd version: 20.5.3

### Possible solution
Looks like it targets this line in `grafana.yaml`: `       image: {{.Values.grafanaImage}}:{{default .Values.global.linkerdVersion .Values.controllerImageVersion}}`

kleimkuhler:
This adds an integration test for upgrading from the latest edge to the current
build.

Closes #4471 

### DIFF
--- a/bin/test-run
+++ b/bin/test-run
@@ -2,12 +2,12 @@
 
 # This script executes the full suite of integration tests, in series:
 # 1. upgrade_stable_integration_tests
-# 1. upgrade_edge_integration_tests
-# 2. helm_integration_tests
-# 3. helm_upgrade_integration_tests
-# 4. uninstall_integration_tests
-# 5. deep_integration_tests
-# 6. external_issuer_integration_tests
+# 2. upgrade_edge_integration_tests
+# 3. helm_integration_tests
+# 4. helm_upgrade_integration_tests
+# 5. uninstall_integration_tests
+# 6. deep_integration_tests
+# 7. external_issuer_integration_tests
 
 bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 
--- a/bin/test-run
+++ b/bin/test-run
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 # This script executes the full suite of integration tests, in series:
-# 1. upgrade_integration_tests
+# 1. upgrade_stable_integration_tests
+# 1. upgrade_edge_integration_tests
 # 2. helm_integration_tests
 # 3. helm_upgrade_integration_tests
 # 4. uninstall_integration_tests
@@ -18,7 +19,8 @@ init_test_run "$@"
 printf '==================RUNNING ALL TESTS==================\n'
 printf 'Testing Linkerd version [%s] namespace [%s] k8s-context [%s]\n' "$linkerd_version" "$linkerd_namespace" "$k8s_context"
 
-upgrade_integration_tests
+upgrade_stable_integration_tests
+upgrade_edge_integration_tests
 helm_integration_tests
 helm_upgrade_integration_tests
 uninstall_integration_tests
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -235,7 +235,7 @@ func TestInstallOrUpgradeCli(t *testing.T) {
 		}
 
 		// prepare for stage 2
-		args = append([]string{"control-plane"}, args...)
+		args = append([]string{"control-plane --addon-overwrite"}, args...)
 	}
 
 	exec := append([]string{cmd}, args...)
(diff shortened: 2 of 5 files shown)


bin/test-run
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 # This script executes the full suite of integration tests, in series:
-# 1. upgrade_integration_tests
+# 1. upgrade_stable_integration_tests
+# 1. upgrade_edge_integration_tests
```
alpeb, line 5: Nit: numbering

test/install_test.go
```diff
@@ -235,7 +235,7 @@ func TestInstallOrUpgradeCli(t *testing.T) {
 		}
 
 		// prepare for stage 2
-		args = append([]string{"control-plane"}, args...)
+		args = append([]string{"control-plane", "--addon-overwrite"}, args...)
```
alpeb, line 238: I think we shouldn't add this flag as it doesn't reflect the usual upgrade scenario. Without it the upgrade fails from edge-20.5.3 like described in #4471, but it's green without it using the current latest edge.

kleimkuhler: I don't think this works for any edge right now. I ran into this issue when upgrading from `edge-20.6.1` as well. I originally opened an issue for it, but closed it (#4556).

alpeb: Ah sorry, I was testing in a different branch... I confirmed it's failing upgrading from Friday's edge as well. From what I can see, the resulting upgrades with and without `--from-manifests` are different. The former isn't touching Grafana's version while the latter is. @Pothulapati can you lend a hand to see what's going on?

kleimkuhler: This was updated to get `configmaps/linkerd-config-addons` if it exists.

alpeb:
This might depend on #4571 in order to pass, so I'm merging that one now.

test/install_test.go
```diff
@@ -247,20 +247,29 @@ func TestInstallOrUpgradeCli(t *testing.T) {
...
+		resources = append(resources, "configmaps/"+k8s.ConfigConfigMapName, "secrets/"+k8s.IdentityIssuerSecretName)
+
+		// If `linkerd-config-addons` exists, add it to the resources to get
+		_, err := TestHelper.Kubectl("", "--namespace", TestHelper.GetLinkerdNamespace(), "get", "configmaps/"+k8s.AddOnsConfigMapName)
+		if err != nil {
+			resources = append(resources, "configmaps/"+k8s.AddOnsConfigMapName)
+		}
+
+		args := append([]string{"--namespace", TestHelper.GetLinkerdNamespace(), "get"}, resources...)
+		args = append(args, "-o yaml")
```
alpeb, line 260: Maybe the test is failing also because `-o yaml` is used in a single arg?

kleimkuhler: Yes definitely it. Thanks!

grampelberg:
@kleimkuhler what do we need to do to get this merged?

kleimkuhler:
@grampelberg It needs to be admin merged and then the expected checks need to be updated in GitHub settings.

This PR changes a check name from `Integration tests (upgrade)` to `Integration test (upgrade_stable)`. All checks have passed, it just expects a test that is changed by this PR.

grampelberg:
@admc can you help out?

PR #4562, opened 2020-06-05
Tweak MC log levels

### LINKED ISSUE #4546: Surface probe failures
### What is the issue?

I'm trying to get my own gateway up, it isn't configured quite right yet. `check` is passing in the source cluster but nothing's working. Once I turned my logs up to debug for `service-mirror`, I saw that the probe was failing. Could we:

- Have check fail.
- Issue k8s events against the source cluster's gateway service (in `linkerd-multicluster`).
- Log probe failures at info.

zaharidichev:
This PR just modifies the log levels on the probe and cluster watchers to emit in INFO what they would emit in DEBUG. I think it makes sense as we need that information to track problems. The only difference is that when probing gateways we only log if the probe attempt was unsuccessful. 

Fix #4546

### DIFF
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -289,7 +289,7 @@ func (rcsw *RemoteClusterServiceWatcher) cleanupOrphanedServices() error {
 					// something went wrong with deletion, we need to retry
 					errors = append(errors, err)
 				} else {
-					rcsw.log.Debugf("Deleted service %s/%s as part of GC process", srv.Namespace, srv.Name)
+					rcsw.log.Infof("Deleted service %s/%s as part of GC process", srv.Namespace, srv.Name)
 				}
 			} else {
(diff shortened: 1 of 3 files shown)


controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -289,7 +289,7 @@ func (rcsw *RemoteClusterServiceWatcher) cleanupOrphanedServices() error {
 					// something went wrong with deletion, we need to retry
 					errors = append(errors, err)
 				} else {
-					rcsw.log.Debugf("Deleted service %s/%s as part of GC process", srv.Namespace, srv.Name)
+					rcsw.log.Infof("Deleted service %s/%s as part of GC process", srv.Namespace, srv.Name)
```
kleimkuhler, line 292: While you're changing this, could we make this more clear about what GC process means?

```suggestion
					rcsw.log.Infof("Deleted service %s/%s while cleaning up mirror services", srv.Namespace, srv.Name)
```

Issue #4501, opened 2020-05-28
service topologies: Add EndpointSlice support to destination service


Matei207:
#### **[Link to RFC](https://github.com/linkerd/rfc/pull/23)**

Problem Statement
==============
A prerequisite for service topology is supporting `EndpointSlices`. EndpointSlices as a feature automatically splits a service's endpoints into (by default) 100-endpoint slice resources; apart from making resource files smaller, it also includes new topology labels, such as the host and zone an endpoint is in.

Implementing this will allow the `destination svc` to collect the topological information of an endpoint — one of the three core necessities needed to support service topology. To support endpoint slices, the resource must be watched (in a similar fashion to how endpoints are currently watched and collected), and a check must be made when the `destination svc` is initialised to ensure the resource exists in the cluster (currently it sits behind a feature gate).

Acceptance Criteria
===============

*  Have a check that determines whether `EndpointSlices` resources are supported in the cluster (i.e ensure feature gate is active)
*  An `EndpointSlices` watcher is created and registers all `EndpointSlice` resources in the cache
*  To register `EndpointSlices` there should be new handler functions created
*  All cached `EndpointSlices` are added to their respective services
* `EndpointSlices` should identify services by the service's namespace and the "service" label defined in their spec and not by name (`EndpointSlices` have generated names since there is a 1 — * type relationship between a service and its slices)
* Every `address set` that belongs to a service should now retain topological information associated with it
*  New unit test scenarios with `EndpointSlice` resources should be added to:
    1. `endpoints_watcher_test.go` 
    2.  `ip_watcher_test.go`

### Concerns/Questions
---
- At the moment, the initial idea is to either have an `EndpointSlice` watcher or an `Endpoints` watcher but not both. As I go along, I am interested to see if the two can be reconciled and used at the same time instead — if there are any benefits to it.

PR #4563, opened 2020-06-05
[POC] Destination Service EndpointSlice support

Matei207:
### Issue: #4501 

After a few attempts at adding support for `EndpointSlice` resources, I came up with this poc to show one way of doing this.

### EndpointSlice access
---

* The first prerequisite to supporting `EndpointSlices` is to make sure that the cluster has access to the resource, since it sits behind a feature gate. To do this, we are using a similar check to `ServiceProfile`. First, we check to see what resources are registered with the api-server under the group-version `discovery/v1beta1`. Second, we check to see if `kind: EndpointSlice` is a registered api resource, if it is, we return from the function with no error.

* When we create a new `EndpointWatcher`, we check if the cluster has any `EndpointSlice`. If we get an error, we start the `Endpoints` shared informer, otherwise, we start an `EndpointSlice` shared informer.

### Difference between EndpointSlice and Endpoints
---

* There are a couple of differences between the two resources, namely around what labels, names and addresses they contain.
* `Endpoints` contain ALL endpoints of a service. An endpoint contains a _subset_ -- a group of addresses with common ports. The set of all endpoints is the union of all subsets.
```
subsets:
- addresses:
  - ip: 172.17.0.10
  - ip: 172.17.0.9
  ports:
   - name: http
      port: 8080
   - name: metrics
      port: 9090

# Both IP addresses have the common set of ports {http, metrics}
# So our endpoints in this case are Addresses x Ports in this subset
```
* `EndpointSlices` ARE a subset. Instead of grouping all subsets into a single resource, the slice contains only a set of addresses with common parts.
```
endpoints:
- addresses:
  - 172.17.0.10
     # extra stuff
     # e.g topology, conditions
- addresses:
  - 172.17.0.9
     # more extra stuff
ports:
   - name: http
      port: 8080
   - name: metrics
      port: 9090

# Here, we have a set of endpoints for the slice, where each endpoint has an address, a targetRef
# topology, etc
# If ports was included in a subset (so each subset has a different port set)
# a slice only has one set of ports corresponding to its endpoint addresses
```
* **Other notable differences**:
  * Slices have a prefix (svc name), the rest of their name is randomly generated
  * Slices have a label with the service name
  * This means 1 service ------ * slices instead of a 1:1 relationship (like with Endpoints)
  * **EndpointSlices are dual-stack, an endpoint can have both IPv4 and IPv6** (which is why addresses is a set, and not a field like it is in `Endpoints` with `ip:`

* Because of these differences, `EndpointSlices` as resources need to be processed a bit differently than `Endpoints`. Will go more in-depth in the next sections.

### Handler functions
---

* When implementing new handler functions for `EndpointSlice`, I tried to reuse as much code as possible. Not everything can be reused because of the type system, where applicable I tried to reuse logic by doing type assertions.

* **Add**:
  * Most type assertions have been done here
  * When adding a new `EndpointSlice` in my mind there was one thing that needs to be addressed and is different from `Endpoints`. Specifically, only update those ports whose addresses are in the slice.
  * For example, if an `Endpoints` resource has 2 subsets, one which includes addresses associated with 8080 and one with addresses associated with 6000, all ports would have to be updated when a change to the `Endpoints` object is made, since E = Subset1 U Subset2. A slice IS a subset, however, so we want to update only the `portPublisher` that has a targetPort of 8080, if our slice contains addresses associated with 8080. I'm not sure if this would be a case often met in a real production cluster, I've personally never seen an `Endpoints` with more than 1 subset, but I thought mathematically this should hold (?)
   * To check, I added a function `isTargetPortInSlice` that checks whether the current `portPublisher`'s targetPort is part of the slice, if it is, update the endpoint.
   * Another key difference is when building `AddressSets` for a port. Because each address has a `conditions` field associated with it, we only want to add an address when it is ready (to avoid duplicate ADDs being sent to the proxy).

* **Delete**:
   * No type assertions here because it was too complicated to get it working (due to the tombstones), so I just decided to create its own functions.
   * The key difference here is similar to `Add` -- when we want the delete the addresses of a `portPublisher`, only delete those addresses which were part of the slice, since other slices can be in existence it does not make sense to delete all endpoints.
  * Again, we make use of `isTargetPortInSlice`. If the addressSet is empty when we delete the slice addresses, we do what `noEndpoints()` does -- set the existence to false and tear down the `portPublisher` (along with its listener).

### Misc
---

I'm thinking that the resource access for `EndpointSlices` should be done before on the server-side and passed as an argument to `NewEndpointsWatcher` to make it easier to write tests. 

I planned on adding new tests for the actual PR where we can check conditions such as ensuring a slice's addresses have been completely updated, a slice that does not correspond to a `portPublisher` does not change anything, etc.

When we create a new portPublisher we make another check for each resource to use when we populate it. If we have slices, we will populate the port with it. Not sure what to do when a port is updated, however.

I have tested this with the destination client script and it seems to be working fine. I have tested if `EndpointSlices` can be deleted, updated, added and that `NoEndpoints` is returned when all pods in a slice have been deleted (to test if the address set is empty the functionality is similar to `noEndpoints`).

Let me know if I've missed anything!

### DIFF
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -509,6 +509,9 @@ func (sp *servicePublisher) newPortPublisher(srcPort Port, hostname string) *por
 		}
 		if err == nil {
 			for _, es := range esList {
+				if !isTargetPortInSlice(port.targetPort, es.Ports) {
+					continue
+				}
 				port.updateEndpoints(es)
 			}
 		}
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -11,14 +11,18 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	logging "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	dv1beta1 "k8s.io/api/discovery/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 )
 
 const (
-	kubeSystem = "kube-system"
-	podIPIndex = "ip"
+	kubeSystem        = "kube-system"
+	podIPIndex        = "ip"
+	endpointSliceKind = "EndpointSlice"
 
 	// metrics labels
 	service                = "service"
@@ -145,12 +149,21 @@ func NewEndpointsWatcher(k8sAPI *k8s.API, log *logging.Entry) *EndpointsWatcher 
 		UpdateFunc: func(_, obj interface{}) { ew.addService(obj) },
 	})
 
-	k8sAPI.Endpoint().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    ew.addEndpoints,
-		DeleteFunc: ew.deleteEndpoints,
-		UpdateFunc: func(_, obj interface{}) { ew.addEndpoints(obj) },
-	})
-
+	if err := endpointSliceAccess(k8sAPI.Client); err != nil {
+		ew.log.Debugf("%v\nWatching Endpoints resources")
+		k8sAPI.Endpoint().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    ew.addEndpoints,
+			DeleteFunc: ew.deleteEndpoints,
+			UpdateFunc: func(_, obj interface{}) { ew.addEndpoints(obj) },
+		})
+	} else {
+		ew.log.Debugf("Watching EndpointSlice resources")
+		k8sAPI.ES().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    ew.addEndpoints,
+			DeleteFunc: ew.deleteEndpointSlice,
+			UpdateFunc: func(_, obj interface{}) { ew.addEndpoints(obj) },
+		})
+	}
 	return ew
 }
 
@@ -240,18 +253,24 @@ func (ew *EndpointsWatcher) deleteService(obj interface{}) {
 }
 
 func (ew *EndpointsWatcher) addEndpoints(obj interface{}) {
-	endpoints := obj.(*corev1.Endpoints)
-	if endpoints.Namespace == kubeSystem {
-		return
-	}
-	id := ServiceID{
-		Namespace: endpoints.Namespace,
-		Name:      endpoints.Name,
+
+	switch res := obj.(type) {
+	case *corev1.Endpoints:
+		if res.Namespace == kubeSystem {
+			return
+		}
+		id := ServiceID{res.Namespace, res.Name}
+		sp := ew.getOrNewServicePublisher(id)
+		sp.updateEndpoints(res)
+	case *dv1beta1.EndpointSlice:
+		if res.Namespace == kubeSystem {
+			return
+		}
+		id := ServiceID{res.Namespace, res.Labels[dv1beta1.LabelServiceName]}
+		sp := ew.getOrNewServicePublisher(id)
+		sp.updateEndpoints(res)
 	}
 
-	sp := ew.getOrNewServicePublisher(id)
-
-	sp.updateEndpoints(endpoints)
 }
 
 func (ew *EndpointsWatcher) deleteEndpoints(obj interface{}) {
@@ -280,6 +299,35 @@ func (ew *EndpointsWatcher) deleteEndpoints(obj interface{}) {
 	sp, ok := ew.getServicePublisher(id)
 	if ok {
 		sp.deleteEndpoints()
+	}
+}
+
+func (ew *EndpointsWatcher) deleteEndpointSlice(obj interface{}) {
+	es, ok := obj.(*dv1beta1.EndpointSlice)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			ew.log.Errorf("couldn't get object from DeletedFinalStateUnknown %#v", obj)
+		}
+		es, ok = tombstone.Obj.(*dv1beta1.EndpointSlice)
+		if !ok {
+			ew.log.Errorf("DeletedFinalStateUnknown contained object that is not an EndpointSlice %#v", obj)
+			return
+		}
+	}
+
+	if es.Namespace == kubeSystem {
+		return
+	}
+
+	id := ServiceID{
+		Namespace: es.Namespace,
+		Name:      es.Labels[dv1beta1.LabelServiceName],
+	}
+
+	sp, ok := ew.getServicePublisher(id)
+	if ok {
+		sp.deleteEndpointSlice(es)
 	}
 }
 
@@ -319,14 +367,15 @@ func (ew *EndpointsWatcher) getServicePublisher(id ServiceID) (sp *servicePublis
 /// servicePublisher ///
 ////////////////////////
 
-func (sp *servicePublisher) updateEndpoints(newEndpoints *corev1.Endpoints) {
+func (sp *servicePublisher) updateEndpoints(newResource interface{}) {
 	sp.Lock()
 	defer sp.Unlock()
 	sp.log.Debugf("Updating endpoints for %s", sp.id)
 
 	for _, port := range sp.ports {
-		port.updateEndpoints(newEndpoints)
+		port.updateEndpoints(newResource)
 	}
+
 }
 
 func (sp *servicePublisher) deleteEndpoints() {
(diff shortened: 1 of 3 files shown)


controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +923,43 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
 	}
 	return
 }
+
+func endpointSliceAccess(k8sClient kubernetes.Interface) error {
```
zaharidichev, line 927: does it actually make sense to retru nan error from here. From what I see you just want to know whether you can use slices or not, right? I think a bool should suffice there WDYT ?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +923,43 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
+	}
+
+	return fmt.Errorf("%s resource not found", endpointSliceKind)
+}
+
+func isIPv6Address(address string) bool {
+	return strings.Count(address, ":") >= 2
+}
+
+func isTargetPortInSlice(targetPort namedPort, slicePorts []dv1beta1.EndpointPort) bool {
```
zaharidichev, line 949: TIOLI: We can save on writing that for loop twice I guess: 

```
func isTargetPortInSlice(targetPort namedPort, slicePorts []dv1beta1.EndpointPort) bool {
	for _, p := range slicePorts {
		switch targetPort.Type {
		case intstr.Int:
			if *p.Port == targetPort.IntVal {
				return true
			}
		case intstr.String:
			if *p.Name == targetPort.StrVal {
				return true
			}
		}
	}
	return false
}
```

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -483,6 +603,87 @@ func metricLabels(endpoints *corev1.Endpoints) map[string]string {
...
+					Identity:          es.Annotations[consts.RemoteGatewayIdentity],
+					AuthorityOverride: authorityOverride,
+				}
+			}
+			continue
+		}
+
+		if endpoint.TargetRef.Kind == "Pod" {
+			for _, addr := range endpoint.Addresses {
+				if isIPv6Address(addr) {
```
zaharidichev, line 656: I think instead of doing so you can simply check the `AddressType` on the EndpointsSlice object. A slice will only have addresses of one type in its endpoints unless the documentation is incorrect.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -483,6 +603,87 @@ func metricLabels(endpoints *corev1.Endpoints) map[string]string {
...
+			continue
+		}
+
+		// Every endpoint has addresses that correspond to the EndpointSlice addressType field
+		// Consumers must handle different types of addresses
+		// An "addresses" set must contain at least 1 but no more than 100 addresses
+		if endpoint.TargetRef == nil {
+			serviceName := es.Labels[dv1beta1.LabelServiceName]
+			for _, addr := range endpoint.Addresses {
+				if isIPv6Address(addr) {
```
zaharidichev, line 627: Hm, I think this might cause some trouble here and there. So currently an EndpointsSlice has an address type defined such as: 

```
const (
	// AddressTypeIP represents an IP Address.
	// This address type has been deprecated and has been replaced by the IPv4
	// and IPv6 adddress types. New resources with this address type will be
	// considered invalid. This will be fully removed in 1.18.
	// +deprecated
	AddressTypeIP = AddressType("IP")
	// AddressTypeIPv4 represents an IPv4 Address.
	AddressTypeIPv4 = AddressType(v1.IPv4Protocol)
	// AddressTypeIPv6 represents an IPv6 Address.
	AddressTypeIPv6 = AddressType(v1.IPv6Protocol)
	// AddressTypeFQDN represents a FQDN.
	AddressTypeFQDN = AddressType("FQDN")
)
```

This means that the address can actually be a domain name like `google.com`. In this case this function will return false: 

```
func isIPv6Address(address string) bool {
	return strings.Count(address, ":") >= 2
}
```

Not sure what happens next but I think it is not what we want. Nevertheless the fact that there is an address type that can be a concrete netwrok address (IPv4 or IPv6) OR a domain name is a great feature that I think will solve some of the headscratchers we are having here at the moment

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -483,6 +602,79 @@ func metricLabels(endpoints *corev1.Endpoints) map[string]string {
...
+					continue
+				}
+				ownerKind, ownerName := pp.k8sAPI.GetOwnerKindAndName(pod, false)
+				addresses[id] = Address{
+					IP:        addr,
+					Port:      resolvedPort,
+					Pod:       pod,
+					OwnerName: ownerName,
+					OwnerKind: ownerKind,
+				}
```
zaharidichev, line 667: This loggic here seems quite similar to the one in `endpointsToAddresses`. Can we extract it somehow?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -483,6 +602,79 @@ func metricLabels(endpoints *corev1.Endpoints) map[string]string {
...
+		// and once when ready
+		if !*endpoint.Conditions.Ready {
+			continue
+		}
+
+		// Every endpoint has addresses that correspond to the EndpointSlice addressType field
+		// Consumers must handle different types of addresses
+		// An "addresses" set must contain at least 1 but no more than 100 addresses
+		if endpoint.TargetRef == nil {
+			serviceName := es.Labels[dv1beta1.LabelServiceName]
```
zaharidichev, line 624: What happens if this name is empty? Are we guaranteed that there will be a value for this label?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +914,38 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
+
+	if res.GroupVersion == gv {
+		for _, apiRes := range res.APIResources {
+			if apiRes.Kind == endpointSliceKind {
+				return true
+			}
+		}
+	}
+
+	return false
```
zaharidichev, line 933: I think we can move that somewhere else indeed. Not sure it is worth it before we actually need it outside of this file.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -459,13 +563,28 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
...
+	var resLabels, resAnnotations map[string]string
+	switch res := resource.(type) {
+	case *corev1.Endpoints:
+		{
+			serviceName, ns = res.Name, res.Namespace
+			resLabels, resAnnotations = res.Labels, res.Annotations
+		}
+	case *dv1beta1.EndpointSlice:
+		{
+			serviceName, ns = res.Labels[dv1beta1.LabelServiceName], res.Namespace
```
zaharidichev, line 577: Again, not sure what we need to do in the case when this is not present. I kind of feel this is an invalid scenario and we should just skip processing the slice. Is this the only way to associate an `EndpointSlice` with a service? I assume so. If this label is non present.. then I guess this is an endpoint that is an "orphan" we cannot really do anything with it, so why go through the trouble of processing it at all. Correct me if I am wrong

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +914,38 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
 	}
 	return
 }
+
+func endpointSliceAccess(k8sClient kubernetes.Interface) bool {
+	gv := dv1beta1.SchemeGroupVersion.String()
+	res, err := k8sClient.Discovery().ServerResourcesForGroupVersion(gv)
+	if err != nil {
```
zaharidichev, line 921: maybe we can log the error here

zaharidichev:
@Matei207 As already mentioned I think you are on the right track. Lets abstract ourselves from how thi sall will work with multicluster as there are a few unknowns there at the moment. Instead we can focus on simply suporting the slices as if they were Endpoints and go with that for now. I think this is the right approach to integrate this new feature.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Labels are not the only way of associating an endpointslice to a service. According to the [endpointslice docs](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/), ownership is established by an ownership reference and/or a label. 

e.g 

```yaml
  ownerReferences:
  - apiVersion: v1
    blockOwnerDeletion: true
    controller: true
    kind: Service
    name: web-svc
    uid: f8d63059-f3ef-4a6f-9bfd-d9742ddb70de
```

There is no resource creation validation happening, so theoretically endpoints can be orphaned if they are created manually without specifying either a label or an ownership ref to the service. While it is unlikely to happen, think there should be a safeguard in place. 

To cope with this, I will add a check to the `addEndpoints` function, a slice that:
* has an addressType != IPv4 (so IPv6 or FQDN)
* or does not have an ownership ref
* or does not have a service label

will not be processed.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Yup, sounds good. Will change this.

PR #4567, opened 2020-06-08
Use buster for base and web images too

joakimr-axis:
Requires setting iptables-legacy as the iptables provider.

Cred should go to @erikwilson and @codeman9 for suggesting the use of iptables-legacy.
https://github.com/rancher/k3s/issues/116#issuecomment-526272383
https://github.com/linkerd/linkerd2/pull/3486#issuecomment-536110931

### DIFF
--- a/Dockerfile-base
+++ b/Dockerfile-base
@@ -3,7 +3,7 @@
 # This means that all Linkerd containers share a common set of tools, and furthermore, they
 # are highly cacheable at runtime.
 
-FROM debian:stretch-20190812-slim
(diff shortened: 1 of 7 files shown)


olix0r:
We still have https://github.com/linkerd/linkerd2-proxy-init/pull/3 open -- it should be updated.

cpretzer:
yep, I still have that in a branch. I think I need to use `iptables-legacy` for it as well

Awesome work @joakimr-axis

PR #4569, opened 2020-06-08
Add HA flags for HA installations using helm (#4568)

### LINKED ISSUE #4568: HA installation from helm not have HA status
If install the Linkerd HA using helm, then the indication that the linkerd is deployed as HA is skipped:

```
# helm repo add linkerd https://helm.linkerd.io/stable
# helm fetch --untar linkerd/linkerd2
# exp=$(date -d '+43800 hour' +"%Y-%m-%dT%H:%M:%SZ")
# helm install --namespace=linkerd --name=linkerd2 --set-file global.identityTrustAnchorsPEM=ca.crt --set-file identity.issuer.tls.crtPEM=issuer.crt --set-file identity.issuer.tls.keyPEM=issuer.key --set identity.issuer.crtExpiry=$exp --set installNamespace=false -f linkerd2/values-ha.yaml linkerd/linkerd2
# linkerd check --verbose
DEBU[0001] Skipping check: pod injection disabled on kube-system. Reason: not run for non HA installs
```

MemberIT:
Subject:
Add HA flags for HA installations using helm.

Problem:
If install the Linkerd HA using helm, then the indication that the linkerd is deployed as HA is skipped.

Solution:
Added HA flags if `controllerReplicas` is 3 or more.

Fixes #4568

### DIFF
--- a/charts/linkerd2/templates/_config.tpl
+++ b/charts/linkerd2/templates/_config.tpl
@@ -66,15 +66,12 @@
 {{- end -}}
 
 {{- define "linkerd.configs.install" -}}
-{{- if ge (.Values.controllerReplicas | int) 3 -}}
 {
   "cliVersion":"{{ .Values.global.linkerdVersion }}",
+{{- if ge (.Values.controllerReplicas | int) 3 }}
   "flags":[{"name":"ha","value":"true"}]
-}
-{{- else -}}
-{
-  "cliVersion":"{{ .Values.global.linkerdVersion }}",
+{{- else }}
   "flags":[]
+{{- end }}
 }
-{{- end -}}
 {{- end -}}
--- a/charts/linkerd2/templates/_config.tpl
+++ b/charts/linkerd2/templates/_config.tpl
@@ -66,8 +66,15 @@
 {{- end -}}
 
 {{- define "linkerd.configs.install" -}}
+{{- if ge (.Values.controllerReplicas | int) 3 -}}
+{
+  "cliVersion":"{{ .Values.global.linkerdVersion }}",
+  "flags":[{"name":"ha","value":"true"}]
+}
+{{- else -}}
 {
   "cliVersion":"{{ .Values.global.linkerdVersion }}",
   "flags":[]
 }
+{{- end -}}
 {{- end -}}
--- a/cli/cmd/testdata/install_helm_output_ha.golden
+++ b/cli/cmd/testdata/install_helm_output_ha.golden
@@ -929,7 +929,7 @@ data:
   install: |
     {
       "cliVersion":"linkerd-version",
-      "flags":[]
+      "flags":[{"name":"ha","value":"true"}]
     }
 ---
 # Source: linkerd2/templates/identity.yaml


grampelberg:
@MemberIT mind taking care of [DCO](https://github.com/linkerd/linkerd2/pull/4569/checks?check_run_id=749853735)?

MemberIT:
> @MemberIT mind taking care of [DCO](https://github.com/linkerd/linkerd2/pull/4569/checks?check_run_id=749853735)?
Fix it

charts/linkerd2/templates/_config.tpl
```diff
@@ -66,8 +66,15 @@
 {{- end -}}
 
 {{- define "linkerd.configs.install" -}}
+{{- if ge (.Values.controllerReplicas | int) 3 -}}
+{
+  "cliVersion":"{{ .Values.global.linkerdVersion }}",
+  "flags":[{"name":"ha","value":"true"}]
```
Pothulapati, line 72: I could be missing something here, but can we add a if case for only the `flags` field?

MemberIT:
> @MemberIT Thanks for the PR, This tested fine but not sure if the controlPlaneReplicas >=3 is the right condition for HA but looks like the only one we have got!

@Pothulapati I thought about it and couldn't find a better solution.

charts/linkerd2/templates/_config.tpl
MemberIT: You are right, I corrected it.

alpeb:
What do you guys think about introducing a new `HighAvailability` boolean property in the `values.yaml` and `values-ha.yaml` files, false in the former and true in the latter, to use that in the condition instead of relying on the replicas number?

Pothulapati:
Agree with @alpeb. That feels like a way better approach!

MemberIT:
> What do you guys think about introducing a new `HighAvailability` boolean property in the `values.yaml` and `values-ha.yaml` files, false in the former and true in the latter, to use that in the condition instead of relying on the replicas number?

@alpeb 
Introducing such a boolean property `HighAvailability` was my first thought, but there are two implementations with corresponding disadvantages:
1) We replace the `controllerReplicas` with `HighAvailability`, and if `HighAvailability` is true, then the `replicas` in the deployment is 3. HA is controlled by one variable, but the flexibility is lost if someone needs to set `replicas: 5` or `replicas: 7` in deployment .
2) We save `controllerReplicas` and add `HighAvailability`, but then check anyway that `controllerReplicas` is greater than or equal to 3 when `HighAvailability` is true and HA is controlled by two parameters, which increases complexity.

The current solution seemed to me the most optimal, it does not introduce additional complexity and retains flexibility.

grampelberg:
I prefer having `--ha` implemented as a values file in the helm situation. Helm is by definition more configurable and I don't want to have to sort out conflicts between configuration.

alpeb:
> 2\. We save `controllerReplicas` and add `HighAvailability`, but then check anyway that `controllerReplicas` is greater than or equal to 3 when `HighAvailability` is true and HA is controlled by two parameters, which increases complexity.

Installing linkerd in HA mode through Helm means using the `values-ha.yaml` file. We should still keep on using `controllerReplicas` in case the user wants to tweak the number of replicas he wants, but the `ha` flag persisted in the `linkerd-config` configmap should only depend on the new `HighAvailability` value.

adleong:
The linked issue is really about not skipping the `pod injection disabled on kube-system` check.  Since that's what we care about, I think we should actually be checking the webhookFailurePolicy instead of the replica counts.

Think about it this way: if the replica counts were 1 but the webhookFailurePolicy was "Fail" then we would still not want to skip this check.

ihcsim:
@MemberIT What do you think of @adleong suggestion at https://github.com/linkerd/linkerd2/pull/4569#issuecomment-643500764? Still up for bringing this PR across the finishing line?

ihcsim:
@MemberIT I am closing this PR due to inactivity. Feel free to re-open when you are ready.

Issue #4498, opened 2020-05-28
service topologies: Add node locality token to pod injection template


Matei207:
### Problem Statement
---
In order to make use of service topology, `destination svc` needs to calculate the priority of an endpoint. To give contextual information, we need to send the name of the source node to infer its topology labels. This feature makes use of the `Kubernetes` downward API to set the `nodeName` as an environment variable in the `podSpec` at injection time.

This change will allow the proxy to obtain the `nodeName` as an environment variable by modifying the proxy injection template in `linkerd/templates`.

### Acceptance Criteria
---
* `_proxy.tpl` needs to include a new environment variable (`LINKERD2_PROXY_SOURCE_NODE` or something similar)
* The new environment variable should be set in the podSpec automatically at injection time by making use of the downward API.

PR #4570, opened 2020-06-08
service topologies: Add nodeName to proxy partial (#4498)

Matei207:
### What
---

* Small PR to help with getting the locality of the source node in the context of service topologies, for a bit more background into why this is necessary, the issue (#4498) contains most information. 

* Updated the proxy template partial to do 2 things:
   1. Using the downward API, get the name of the node the pod is on (`spec.nodeName`)
   2. Set the nodeName as an env variable (`LINKERD2_PROXY_SOURCE_NODE`)

* Updated all cli test data using `go test ./cli/cmd/... --update`

* **Manual test:**
  - Tested on `minikube`
  - Build docker images and push to minikube repo

```sh
$ k exec -n emojivoto -it web-8d985c8fb-tq77t -c linkerd-proxy -- bash -c env | grep LINKERD2_PROXY_SOURCE
Output: LINKERD2_PROXY_SOURCE_NODE=minikube
```

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -3,7 +3,7 @@ env:
 {{ if .Values.global.proxy.requireIdentityOnInboundPorts -}}
 - name: LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_IDENTITY
   value: "{{.Values.global.proxy.requireIdentityOnInboundPorts}}"
-{{ end -}}  
+{{ end -}}
 - name: LINKERD2_PROXY_LOG
   value: {{.Values.global.proxy.logLevel}}
 - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
@@ -21,7 +21,7 @@ env:
 {{ if .Values.global.proxy.isGateway -}}
 - name: LINKERD2_PROXY_INBOUND_GATEWAY_SUFFIXES
   value: {{printf "svc.%s." .Values.global.clusterDomain}}
-{{ end -}}  
+{{ end -}}
 - name: LINKERD2_PROXY_DESTINATION_GET_SUFFIXES
   value: {{printf "svc.%s." .Values.global.clusterDomain}}
 - name: LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES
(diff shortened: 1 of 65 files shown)


grampelberg:
@Pothulapati we were doing something like this to get extra information into jaeger weren't we?

Pothulapati:
@grampelberg Yep, We do use the downward APi to get pod labels but it is optional and attaches those labels as a volume and that path as a env for proxy to read from

https://github.com/linkerd/linkerd2/blob/master/charts/patch/templates/patch.json#L70

https://github.com/linkerd/linkerd2/blob/master/charts/partials/templates/_proxy.tpl#L36

olix0r:
Per comments on the API PR... I think we should encode all of this information in the `CONTEXT_TOKEN` -- the intent of this is exactly so that the injector can share information with the destination service to inform discovery.  My main question is whether we should migrate that field to being structured (i.e. protobuf/json). Basically: if we're going to encode more data in this field, how do we manage that going forward?

Matei207:
From what I can gather, migrating the contextual token to a structured way (protobuf) wouldn't impact the performance that much, considering the volume of data is relatively small for the destination look-ups.

From an organisational perspective:
  * A nested structure would make it easier to understand the structure of the request and make it easier to extend in the future if more information needs to be piggy-backed on the contextual token.
  * A flat structure would be easier to deal with now and would require fewer changes to be made in the context of service topology, however, I'm not incredibly excited about comma-delimited <k,v> pairs being sent in the same string/field.

Not sure if I'm missing any other critical/obvious details. 

Any thoughts on this @grampelberg @adleong?

adleong:
This sounds like it's headed the right direction.  Perhaps we should get an issue open where we can explore the detailed of the context token format.

One quick note is that if we want to populate the context token from the downward API then I think we're limited to a data format that we can populate with variable substitution.  For example, I can't think of a simple way to take fields from the downward API (such as node name) and encode them into protobuf directly in the manifest yml.

adleong:
Hey @Matei207, since the plan is to use the context token for this, should we close this PR?

Matei207:
Thank you for pointing that out @adleong! Will be closing this PR in favour of using the context token to send locality information for service topology.

PR #4571, opened 2020-06-08
Removes grafana.image.version field

Pothulapati:
Fixes https://github.com/linkerd/linkerd2/issues/4556

Currently, whenever we perform upgrades, the configuration from `linkerd-config-addons` is used and overrides the default upgrades. This prevents grafana image tag from being upgraded, when `grafana.image.version` is present in `linkerd-config-addons`

The fix is to remove `grafana.image.version` field, so that users we don't use the value from `linkerd-config-addons` (which prevents us to upgrade to the latest image).

This means that, users who want to get their own Grafana Image should have the image tagged with the control-plane version.

This does not effect the bring your own Grafana image use-case. 

Future Concrete Solution:
- Do not build grafana image for each tag, just like we don't build other add-on images like jaeger, etc. Thus making Controlplane upgrades separate from grafana image.

### DIFF
--- a/charts/add-ons/grafana/templates/grafana.yaml
+++ b/charts/add-ons/grafana/templates/grafana.yaml
@@ -121,7 +121,7 @@ spec:
       - env:
         - name: GF_PATHS_DATA
           value: /data
-        image: {{.Values.image.name}}:{{default .Values.global.controllerImageVersion .Values.image.version}}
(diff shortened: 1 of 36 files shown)


Pothulapati:
Raised https://github.com/linkerd/linkerd2/issues/4572

PR #4581, opened 2020-06-09
Fix yaml in linkerd-config-addons when providing grafanaUrl

alpeb:
Put back space after `grafanaUrl` label in `linkerd-config-addons.yaml`
to avoid breaking the yaml parsing.

```
$ linkerd check
...
linkerd-addons
--------------
‼ 'linkerd-config-addons' config map exists
    could not unmarshal linkerd-config-addons config-map: error
    unmarshaling JSON: while decoding JSON: json: cannot unmarshal
    string into Go struct field Values.global of type linkerd2.Global
```
This was added in #4544 to avoid having the configmap being badly formatted.

So this PR fixes the yaml, but then if we don't set `grafanaUrl` the
configmap format gets messed up, but apparently that's just a cosmetic
problem:

```
apiVersion: v1
data:
  values: "global:\n  grafanaUrl: \ngrafana:\n  enabled: true\n
  image:\n    name:
      gcr.io/linkerd-io/grafana\n  name: linkerd-grafana\n  resources:\n
      cpu:\n      limit:
          240m\n    memory:\n      limit: null\ntracing:\n  enabled:
          false"
          kind: ConfigMap
```

### DIFF
--- a/charts/linkerd2/templates/linkerd-config-addons.yaml
+++ b/charts/linkerd2/templates/linkerd-config-addons.yaml
@@ -21,7 +21,7 @@ metadata:
 data:
   values: |-
     global:
-      grafanaUrl: {{.Values.global.grafanaUrl}}
+      grafanaUrl: "{{.Values.global.grafanaUrl}}"
     grafana:
       {{- include "linkerd.addons.sanitize-config" .Values.grafana}}
     tracing:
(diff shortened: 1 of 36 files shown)


charts/linkerd2/templates/linkerd-config-addons.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
 data:
   values: |-
     global:
-      grafanaUrl:{{.Values.global.grafanaUrl}}
```
olix0r, line 24: tioli-- should we quote this so it shows up as `""` when empty?

alpeb: That actually fixes the formatting issue :+1:

PR #4588, opened 2020-06-11
Resolve gateway hostnames into IP addresses

### LINKED ISSUE #4582: multicluster doesn't work on EKS
When attempting to link a source cluster to a target EKS cluster using the `linkerd multicluster link` command, the service mirror controller in the source cluster logs this error:

```msg="Error processing RemoteGatewayCreated: {gatewaySpec: GatewaySpec: {gatewayName: linkerd-gateway, gatewayNamespace: linkerd-multicluster, clusterName: data-dev, addresses: [[]], incomingPort: 4143, resourceVersion: 6142270, identity: linkerd-gateway.linkerd-multicluster.serviceaccount.identity.linkerd.cluster.local, probeConfig: ProbeConfig: {path: /health, port: 4181, periodInSeconds: 3}}} 
(will retry): Inner errors:\n\tEndpoints \"linkerd-gateway-[xxx]\" is invalid: [subsets[0].addresses[0].ip: Invalid value: \"\": must be a valid IP address, (e.g. 10.9.8.7), 
subsets[0].addresses[0].hostname: Invalid value [...]
```

and fails to create the endpoints object for the gateway mirror.

This happens because the service mirror controller attempts to copy the `hostname` and `IP` fields from the gateway service's ingress struct to the EndpointAddress struct of the gateway mirror's endpoints object.  In EKS, the ingress struct of the `linkerd-gateway` service is populated with a `hostname` and no `IP`.  However, the `IP` field of EndpointAddress is mandatory.

adleong:
Fixes #4582 

When a target cluster gateway is exposed as a hostname rather than with a fixed IP address, the service mirror controller fails to create mirror services and gateway mirrors for that gateway.  This is because we only look at the IP field of the gateway service.

We make two changes to address this problem:
 
First, when extracting the gateway spec from a gateway that has a hostname instead of an IP address, we do a DNS lookup to resolve that hostname into an IP address to use in the mirror service endpoints and gateway mirror endpoints.

Second, we schedule a repair job on a regular (1 minute) to update these endpoint objects.  This has the effect of re-resolving the DNS names every minute to pick up any changes in DNS resolution.

### DIFF
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -39,6 +39,7 @@ type (
 		log                    *logging.Entry
 		eventsQueue            workqueue.RateLimitingInterface
 		requeueLimit           int
+		repairPeriod           time.Duration
 	}
 
 	// ProbeConfig describes the configured probe on particular gateway (if presents)
@@ -178,6 +179,7 @@ func NewRemoteClusterServiceWatcher(
 	cfg *rest.Config,
 	clusterName string,
 	requeueLimit int,
+	repairPeriod time.Duration,
 	clusterDomain string,
 ) (*RemoteClusterServiceWatcher, error) {
 	remoteAPI, err := k8s.InitializeAPIForConfig(cfg, false, k8s.Svc)
@@ -198,6 +200,7 @@ func NewRemoteClusterServiceWatcher(
 		}),
 		eventsQueue:  workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
 		requeueLimit: requeueLimit,
+		repairPeriod: repairPeriod,
 	}, nil
 }
 
@@ -1078,7 +1081,7 @@ func (rcsw *RemoteClusterServiceWatcher) Start() error {
 	go rcsw.processEvents()
 
 	go func() {
-		ticker := time.NewTicker(1 * time.Minute)
+		ticker := time.NewTicker(rcsw.repairPeriod)
 		for {
 			select {
 			case <-ticker.C:
--- a/controller/cmd/service-mirror/cluster_watcher.go
+++ b/controller/cmd/service-mirror/cluster_watcher.go
@@ -3,11 +3,14 @@ package servicemirror
 import (
 	"errors"
 	"fmt"
+	"net"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/linkerd/linkerd2/controller/k8s"
 	consts "github.com/linkerd/linkerd2/pkg/k8s"
+	"github.com/prometheus/client_golang/prometheus"
 	logging "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


controller/cmd/service-mirror/cluster_watcher_mirroring_test.go
```diff
@@ -262,6 +264,12 @@ func TestRemoteServiceUpdatedMirroring(t *testing.T) {
 }
 
 func TestRemoteGatewayUpdatedMirroring(t *testing.T) {
+
+	linkerdIP, err := net.ResolveIPAddr("ip", "linkerd.io")
+	if err != nil {
+		t.Fatal(err)
+	}
```
olix0r, line 271: This definitely isn't a blocker -- but this test requires internet connectivity, which is generally something I think we want to avoid.

I think we could avoid this if we hid https://github.com/linkerd/linkerd2/pull/4588/files#diff-d826c2a036f7464e72c567626b5adce9R1150 behind adding a `resolve` strategy onto `RemoteClusterServiceWatcher`.  By default we use an impl that actually hits the network; but in tests we could mock it.

olix0r: Or, could we use `localhost` in the test?

grampelberg:
@spaghettifunk check it out! Would love to know if this fixes your issues.

spaghettifunk:
Worked like a charm 🎉 🎉  Thank you very much for this fix!

controller/cmd/service-mirror/cluster_watcher.go
```diff
@@ -1121,9 +1148,16 @@ func (rcsw *RemoteClusterServiceWatcher) extractGatewaySpec(gateway *corev1.Serv
 
 	var gatewayEndpoints []corev1.EndpointAddress
 	for _, ingress := range gateway.Status.LoadBalancer.Ingress {
+		ip := ingress.IP
+		if ip == "" {
+			ipAddr, err := net.ResolveIPAddr("ip", ingress.Hostname)
```
zaharidichev, line 1153: Just to verify. What happens here if this name resolution returns an IPv6 address and there is no dual stack feature enabled on this cluster. Would that still work ? If not, should we make sure this is not an IPv6 addr ?

adleong: To be honest, I have no idea what the state of our ipv6 support is in Linkerd.

olix0r: Are we actually making IPv4-specific assumptions here? We need to document this at the very least. It's debt.

adleong: I don't think there are any ipv4-specific assumptions here.  `ResolveIPAddr` can return either ipv4 or ipv6 addresses and the `String()` implementation will render them correctly.  So we support ipv6 here to whatever extent that kubernetes does.

PR #4591, opened 2020-06-12
Update function signature of `testutil.NewGenericTestHelper`

mayankshah1607:
This PR is a follow-up for https://github.com/linkerd/linkerd2/pull/4532

Once merged, `testutil.NewGenericTestHelper` will

- accept `version`, `httpClient` and `k8sContext` as parameters
- initialize the `KubernetesHelper` field in the returned instance of `*TestHelper`

Cc: @Pothulapati @alpeb

### DIFF
--- a/testutil/test_helper.go
+++ b/testutil/test_helper.go
@@ -71,6 +71,7 @@ func NewGenericTestHelper(
 	namespace,
 	upgradeFromVersion,
 	clusterDomain,
+	k8sContext,
 	helmPath,
 	helmChart,
 	helmStableChart,
@@ -78,8 +79,8 @@ func NewGenericTestHelper(
 	externalIssuer,
 	uninstall bool,
 	httpClient http.Client,
-) *TestHelper {
-	return &TestHelper{
+) (*TestHelper, error) {
+	h := &TestHelper{
 		linkerd:            linkerd,
 		version:            version,
 		namespace:          namespace,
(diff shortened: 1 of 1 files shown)


testutil/test_helper.go
```diff
@@ -76,9 +77,11 @@ func NewGenericTestHelper(
 	helmReleaseName string,
 	externalIssuer,
 	uninstall bool,
+	httpClient http.Client,
 ) *TestHelper {
 	return &TestHelper{
```
Pothulapati, line 82: Looks like there is also `KubernetesHelper` variable in `TestHelper`. Maybe we can use that somewhere in the code using the returned value from `NewKubernetesHelper()`?

olix0r:
@mayankshah1607 Would you mind updating the PR title to describe what this change does? Consider how this description will fit in the output of `git log` later -- try to make it easy for people to understand what is changing in the summary, and then the description should explain why and how. Thanks!

testutil/test_helper.go
mayankshah1607: Ah, nice catch - was only considering setting the private fields through this method, but it certainly makes sense to initialize the `KubernetesHelper` field as well. That would mean that `NewGenericTestHelper` would now also have to accept `k8sContext` as a parameter and also return the error returned from `NewKubernetesHelper`.

I have pushed the changes incorporating the above. PTAL! :smile:

testutil/test_helper.go
```diff
@@ -91,7 +95,16 @@ func NewGenericTestHelper(
 		clusterDomain:  clusterDomain,
 		externalIssuer: externalIssuer,
 		uninstall:      uninstall,
+		httpClient:     httpClient,
 	}
+
+	kubernetesHelper, err := NewKubernetesHelper(k8sContext, h.RetryFor)
```
Pothulapati, line 101: As the `kubernetesHelper` is public, and so is `NewKubernetesHelper()`, why can't we just take it as a argument and assign it directly, rather than using this function directly. Any upsides to this?

mayankshah1607: Done!

PR #4593, opened 2020-06-12
Isolate tests by cluster and make run interface simpler

kleimkuhler:
## Summary

Change the default behavior of integration tests to be isolated by cluster.
Additionally, make running one or all tests easier than the current process.

These changes are explained more in the [Testing
RFC](https://github.com/linkerd/rfc/blob/master/design/0004-isolated-integration-tests.md)

## Changes

This is a script used only by Linkerd developers, but there is a lot of useful
usage examples and explanations in `bin/tests --help` output:

```
Run Linkerd integration tests.

Optionally specify one of the following tests: [upgrade helm helm-upgrade uninstall deep external-issuer]

Usage:
    tests [--images] [--images-host ssh://linkerd-docker] [--name test-name] [--skip-kind-create] /path/to/linkerd

Examples:
    # Run all tests in isolated clusters
    tests /path/to/linkerd

    # Run single test in isolated clusters
    tests --name test-name /path/to/linkerd

    # Skip KinD cluster creation and run all tests in default cluster context
    tests --skip-kind-create /path/to/linkerd

    # Load images from tar files located under the 'image-archives' directory
    # Note: This is primarly for CI
    tests --images /path/to/linkerd

    # Retrieve images from a remote docker instance and then load them into KinD
    # Note: This is primarly for CI
    tests --images --images-host ssh://linkerd-docker /path/to/linkerd

Available Commands:
    --name: the argument to this option is the specific test to run
    --skip-kind-create: skip KinD cluster creation step and run tests in an existing cluster.
    --images: (Primarily for CI) use 'kind load image-archive' to load the images from local .tar files in the current directory.
    --images-host: (Primarily for CI) the argument to this option is used as the remote docker instance from which images are first retrieved (using 'docker save') to be then loaded into KinD. This command requires --images.
```

### Run all tests

Old:

```bash
bin/test-run $PWD/bin/linkerd
```

New:

```bash
bin/tests $PWD/bin/linkerd
```

### Run single test (upgrade for example):

Current:

```bash
. bin/_test-run.sh
init_test_run $PWD/bin/linkerd
upgrade_integration_tests
```

New:

```bash
bin/tests --name upgrade $PWD/bin/linkerd
```

### Run tests in isolated KinD clusters

Current: Not possible without running single tests in newly created clusters
manually

New:

```bash
bin/tests $PWD/bin/linkerd
```

### Run tests in isolated namespaces on an existing cluster

Old:

```bash
bin/test-run $PWD/bin/linkerd
```

New:

```bash
bin/tests --skip-kind-create $PWD/bin/linkerd
```

## CI

`kind_integration` has been updated so that it does not create a KinD cluster as
part of its test setup.

`cloud_integration` passes the `--skip-kind-create` flag so that the tests are
run serially in a non-KinD cluster.


kleimkuhler:
This changes the names of tests with underscores to hyphens. The 3 `Expected` checks have passed if look for their hyphen equivalent (e.g. `Integration tests (custom_domain)` -> `Integration tests (custom-domain)`)

kleimkuhler:
@joakimr-axis Let me know if you have any feedback on this! It touches a lot of the testing scripts that you've had an eye on.

.github/workflows/kind_integration.yml
```diff
@@ -125,35 +125,11 @@ jobs:
...
-        # Fetch images from the Packet host and load them into the local KinD cluster
-        bin/kind-load --images --images-host ssh://linkerd-docker
-
-        # Load proxy-init and prometheus images into KinD while it is
-        # available. Allow these commands to fail since they will be cached
-        # for the next run.
-        kind load image-archive <(DOCKER_HOST=ssh://linkerd-docker docker save $PROXY_INIT_IMAGE_NAME) 2>&1 || true
-        kind load image-archive <(DOCKER_HOST=ssh://linkerd-docker docker save $PROMETHEUS_IMAGE_NAME) 2>&1 || true
-    - name: Load image archives into the local KinD cluster (Forked repositories)
+        bin/tests --images --images-host ssh://linkerd-docker --name ${{ matrix.integration_test }} $HOME/.linkerd
```
joakimr-axis, line 131: If $HOME would contain whitespaces, there would be a failure here.

`"$HOME/.linkerd"` or `"$HOME"/.linkerd` would be the safest choice eliminating unlikely corner case issues, IMO.

TEST.md
```diff
@@ -94,110 +94,93 @@ bin/shellcheck -x bin/*
...
-
-That will create multiple namespaces in your Kubernetes cluster:
-
-```bash
-$ kubectl get ns | grep l5d-integration
-l5d-integration                  Active    4m
-l5d-integration-egress-test      Active    2m
-l5d-integration-get-test         Active    1m
-...
+bin/tests $PWD/bin/linkerd
```
joakimr-axis, line 178: `bin/tests "$PWD/bin/linkerd"` or `bin/tests "$PWD"/bin/linkerd`

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+# Override CI's `set -e` default, so we can catch errors manually and display
+# proper messages
+set +e
+
+##### Test setup helpers #####
+
+export test_names=(upgrade helm helm-upgrade uninstall deep external-issuer)
+
+handle_input() {
+  export images=""
```
joakimr-axis, line 12: Better use single quotes `''` here since there is nothing to be expanded.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+        echo "    # Retrieve images from a remote docker instance and then load them into KinD"
+        echo "    # Note: This is primarly for CI"
+        echo "    ${0##*/} --images --images-host ssh://linkerd-docker /path/to/linkerd"
+        echo ""
+        echo "Available Commands:"
+        echo "    --name: the argument to this option is the specific test to run"
+        echo "    --skip-kind-create: skip KinD cluster creation step and run tests in an existing cluster."
+        echo "    --images: (Primarily for CI) use 'kind load image-archive' to load the images from local .tar files in the current directory."
+        echo "    --images-host: (Primarily for CI) the argument to this option is used as the remote docker instance from which images are first retrieved (using 'docker save') to be then loaded into KinD. This command requires --images."
+        exit 0
```
joakimr-axis, line 51: The shellscripting folks tend to suggest using one echo command instead of multiple ones. No that a modern Threadripper CPU would stagger, but it is a relatively big performance difference between one echo command and multiple ones. Hence it would be something like

```
echo "Run Linkerd integration tests.

Optionally specify one of the following tests: [${test_names[*]}]

Usage:
    ${0##*/} [--images] [--images-host ssh://linkerd-docker] [--name test-name] [--skip-kind-create] /path/to/linkerd
[etc.]
"
```
which is also easier to read and maintain.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+        ;;
+      --images)
+        images=1
+        ;;
+      --images-host)
+        images_host=$2
+        if [ -z "$images_host" ]; then
+          echo "Error: the argument for --images-host was not specified"
```
joakimr-axis, line 59: Nothing is expanded here, static string, better to use single quotes.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+          exit 1
+        fi
+        shift
+        ;;
+      --name)
+        test_name=$2
+        if [ -z "$test_name" ]; then
+          echo "Error: the argument for --name was not specified"
```
joakimr-axis, line 67: Single quotes

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+        skip_kind_create=1
+        ;;
+      *)
+        break
+    esac
+    shift
+  done
+
+  if [ "$images_host" ] && [ -z "$images" ]; then
+    echo "Error: --images-host needs to be used with --images" >&2
```
joakimr-axis, line 82: Single quotes.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+    exit 1
+  fi
+
+  export linkerd_path="$1"
+  if [ -z "$linkerd_path" ]; then
+    echo "Error: path to linkerd binary is required"
```
joakimr-axis, line 88: Single quotes.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+    echo "Help:"
+    echo "     ${0##*/} -h|--help"
+    echo "Basic usage:"
+    echo "     ${0##*/} /path/to/linkerd"
```
joakimr-axis, line 92: See previous suggestion to use single echo command instead.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+test_setup() {
+  bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
+  export bindir
+
+  export test_directory="$bindir"/../test
+
+  check_linkerd_binary
+}
+
+check_linkerd_binary(){
```
joakimr-axis, line 106: In the other locations you have `() {` and here it is `(){`. Not that it matters for functionality, but it is inconsistent. :)

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  printf 'Checking the linkerd binary...'
```
joakimr-axis, line 107: Using `echo` would be better, since it takes less resources than `printf` and no `printf`-specific functionality is used here.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  local name=$1
+  "$bindir"/kind delete cluster --name "$name" 2>&1
+}
+
+cleanup_cluster() {
+  "$bindir"/test-cleanup "$context" > /dev/null 2>&1
+  exit_on_err 'error removing existing Linkerd resources'
+}
+
+check_if_k8s_reachable(){
```
joakimr-axis, line 147: `() {` ?

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  printf 'Checking if there is a Kubernetes cluster available...'
+  exit_code=0
+  kubectl --context="$context" --request-timeout=5s get ns > /dev/null 2>&1
+  exit_on_err 'error connecting to Kubernetes cluster'
+  printf '[ok]\n'
```
joakimr-axis, line 152: Better with `echo` here

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+}
+
+check_if_l5d_exists() {
+  printf 'Checking if Linkerd resources exist on cluster...'
```
joakimr-axis, line 156: Better with `echo` here

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  resources=$(kubectl --context="$context" get all,clusterrole,clusterrolebinding,mutatingwebhookconfigurations,validatingwebhookconfigurations,psp,crd -l linkerd.io/control-plane-ns --all-namespaces -oname)
+  if [ -n "$resources" ]; then
+    printf '
+Linkerd resources exist on cluster:
+\n%s\n
+Help:
+    Run: [%s/test-cleanup]' "$resources" "$bindir"
+    exit 1
+  fi
+  printf '[ok]\n'
```
joakimr-axis, line 167: Better with `echo` for line 167

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  if [ -z "$skip_kind_create" ]; then
+    delete_cluster "$name"
+  else
+    cleanup_cluster
+  fi
+}
+
+get_test_config() {
+  local name=$1
+  config=""
```
joakimr-axis, line 192: `config=''`

Nothing is to be expanded

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  case $name in
+    cluster-domain)
+      config="cluster-domain"
```
joakimr-axis, line 195: Better with single quotes, nothing is to be expanded

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+      ;;
+    *)
+      config="default"
```
joakimr-axis, line 198: `config='default'`

Nothing is to be expanded

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
...
+  )
+  exit_on_err 'install_stable() - installing stable failed'
+
+  (
+    set -x
+    "$linkerd_path" check 2>&1
+  )
+  exit_on_err 'install_stable() - linkerd check failed'
+
+  #Now we need to install the app that will be used to verify that upgrade does not break anything
```
joakimr-axis, line 239: `# Now`

bin/test-cleanup
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/local/bin/bash
```
joakimr-axis, line 1: Why `/usr/local`?

bin/tests
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
```
joakimr-axis, line 3: Is this file sourced in a way where the BASH_SOURCE[0] is needed, or could the simpler

`$( cd "${0%/*}" && pwd )`

be used?

bin/tests
```diff
@@ -0,0 +1,27 @@
...
+
+# shellcheck source=_test-helpers.sh
+. "$bindir"/_test-helpers.sh
+handle_input "$@"
+
+if [ -n "$test_name" ]; then
+  config=$(get_test_config "$test_name")
+  start_test "$test_name" "$config"
+else
+  printf '==================RUNNING ALL TESTS==================\n'
```
joakimr-axis, line 13: Better use `echo` here since no `printf`-specifics are used.

bin/tests
```diff
@@ -0,0 +1,27 @@
...
+
+  for test_name in "${test_names[@]}"; do
+    config=$(get_test_config "$test_name")
+    start_test "$test_name" "$config"
+  done
+
+  if [ $exit_code -eq 0 ]; then
+  printf '\n=== PASS: all tests passed\n'
```
joakimr-axis, line 21: Better use echo here since no printf-specifics are used.

bin/tests
```diff
@@ -0,0 +1,27 @@
...
+  else
+    printf '\n=== FAIL: at least one test failed\n'
```
joakimr-axis, line 23: Better use echo here since no printf-specifics are used.

joakimr-axis:
> @joakimr-axis Let me know if you have any feedback on this! It touches a lot of the testing scripts that you've had an eye on.

Sure thing! I have added a few suggestions now!

bin/test-cleanup
kleimkuhler: Whoops I didn't mean to include this in the changes!

bin/tests
kleimkuhler: This is how `bindir` is created in all the scripts. I'd rather not change it in a single place here since I'm not sure if that would continue to work on all shells. If it does continue to work, that can be a separate change at some point.

bin/_test-helpers.sh
kleimkuhler: I'm going to keep this `printf`. We use the fact that it doesn't start a new line, which the `[ok]` relies on below so that it prints on the same line.

bin/tests
joakimr-axis: There is in fact 22 script files in the `bin `dir today where it is not, but done this POSIX way.
In some cases where bash is used and sourced in different ways, the BASH_SOURCE is needed (and there are 25 such script files in the `bin `dir), although I do not think this file is one of those cases.

bin/_test-helpers.sh
kleimkuhler: And this is my same response for all the below suggestions about changing `printf` to`echo`!

joakimr-axis: Ah, then `printf`-specific functionality _is_ indeed used. My bad. :)

bin/tests
kleimkuhler: Yep looks like it's a big mix. I'm going to keep this as is for now still. It's not related to this actual change and is how it was before. If it is actually better to change this, I think that can be done everywhere all at once in a single PR.

joakimr-axis: I guess not a big mix, but using Bash functionality where it is needed (in some of those occurrences not? and POSIX where there is no need for extra bells & whistles. Regardless of what I agree that your strategy of handling that separately is good.

bin/_test-helpers.sh
```diff
@@ -0,0 +1,318 @@
+#!/bin/bash
+
+# Override CI's `set -e` default, so we can catch errors manually and display
+# proper messages
+set +e
+
+##### Test setup helpers #####
+
+export test_names=(upgrade helm helm-upgrade uninstall deep external-issuer)
```
alpeb, line 9: isn't `cluster-domain` missing from this list?

kleimkuhler: Yep this on purpose right now, but I realize it affects the output for available tests when running `bin/tests --help`.

It's missing the same way it is currently missing in `bin/test-run`; it's used only on cloud integration where the cluster hasn't been configured with a custom cluster domain.

We use it in the kind integration tests where we can specify it with `bin/tests --test cluster-domain` and make sure it runs on a cluster that has been configured properly.

I'll change it so that `cluster-domain` renders in the help output, but it shouldn't run on the cloud integration tests.

alpeb:
Also I've just noticed the helm upgrade test is currently failing silently in master because the `latest_stable` function is no longer defined. Depending on how long will the current PR going to get merged, it might be worth fixing that in a separate PR to have it run for this week's edge and next week's stable.

bin/_test-helpers.sh
alpeb: You meant it's only used in `kind_integration.yml` right? Might also be valuable to be available in local testing as well.

kleimkuhler: Yep only used in `kind_integration.yml`.

I'd like to avoid adding more flags to `bin/tests`, but maybe makes sense to somehow indicate to run (or not run) certain tests? `cloud_integration.yml` would then use that to indicate it should *not* run `cluster-domain` test.

kleimkuhler:
@alpeb Good catch! I opened #4622

TEST.md
```diff
@@ -94,110 +94,93 @@ bin/shellcheck -x bin/*
 The `test/` directory contains a test suite that can be run to validate Linkerd
 functionality via a series of end-to-end tests.
 
-### Prerequisites
+### Prerequisites for default behavior
 
-The integration test suite operates on your currently configured Kubernetes
-cluster. Prior to running the test suite, verify that:
+The integration tests will configure their own KinD clusters by default. There
```
adleong, line 99: isn't `kind` a prerequisite?  What about resource requirements?  Is `kind` running on docker-for-desktop on a 4 year old macbook going to cut it?

kleimkuhler: > isn't kind a prerequisite? What about resource requirements?

The script uses `bind/kind` which downloads `kind` if it is not present. You don't need `kind` in your `PATH` to use the script.

> Is kind running on docker-for-desktop on a 4 year old macbook going to cut it?

I haven't had any issues with running `kind` locally on my 2015 macbook. If anything, Docker For Mac with Kubernetes turned on seems to use a lot more resources than `kind` ever does; I rarely use Docker For Mac these days because of that.

Also keep in mind we already use `kind` on the GitHub Action runners which I think are 2-core machines with 7 GB of ram. Knowing we don't have issues on those, I think we're good to assume running this locally will be fine as well.

PR #4597, opened 2020-06-14
Support parsing PKCS8 encoded private keys

alex-berger:
Tools like cert-manager might encode private keys in PKCS8 format instead of PKCS1
in which case linkerd would fail as it cannot parse PKCS8 encoded private keys.

With this commit support for parsing PKCS8 encoded private keys is added to linkerd,
allowing it to read ECDSA and RSA keys encoded in PKCS8.

Unit tests have been added to test the private key parsing.

This commit addresses https://github.com/jetstack/cert-manager/issues/2942.

Signed-off-by: Alexander Berger <alex.berger@nexxiot.com>

### DIFF
--- a/pkg/tls/codec.go
+++ b/pkg/tls/codec.go
@@ -3,10 +3,12 @@ package tls
 import (
 	"bytes"
 	"crypto/ecdsa"
+	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"reflect"
 )
 
 // === ENCODE ===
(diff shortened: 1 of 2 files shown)


olix0r:
Before we merge this, we should check that a cluster created with an RSA key actually works, or we should introduce checks in the appropriate places to prevent cluster misconfiguration.

Perhaps we should even add an integration test for this?

adleong:
Attempting to use an RSA pkcs8 key will cause `linkerd install` to abort with this error:

```
Error: failed to verify issuer certs stored on disk: must use ECDSA for public key algorithm, instead RSA was used
```

olix0r:
@adleong OK, thanks. I wonder how far we get with a Helm install that uses RSA keys. I don't think that's a blocker, though -- but it would be useful to understand so that we can pattern-match user-reported issues.

zaharidichev:
@adleong I think I am getting confused. So currently we can support RSA private keys but we cannot support RSA public ones, correct?

olix0r:
@zaharidichev I don't think we should actually support RSA keys in the identity system at this point. My main understanding of this is that it allows cert manager to create ecdsa credentials that the identity controller can use.

adleong:
It looks like CI is stuck.  @kleimkuhler could this be because of https://github.com/linkerd/linkerd2/pull/4557 ?  @alex-berger you may need to merge latest master to pick up that change and unstick CI.

kleimkuhler:
@adleong Yep exactly! Once master is merged in, CI should run those expected tests.

PR #4600, opened 2020-06-14
Upgrade Grafana to 7.0.3

cpretzer:
# Upgrade Grafana to 7.0.3

Grafanalabs recently released version 7 which includes performance improvements and additional functionality and this pull request updates the Dockerfile to use `grafana/grafana:7.0.3` to take advantage of the improvements and functionality.

## Validation
Built the Dockerfile and pushed an image to my dockerhub account, then changed the `linkerd-grafana` image version in an existing Linkerd 2 installation.

Once the new pod was running, opened the dashboard and checked the grafana dashboards. Everything looks good, and further validation is appreciated.

### DIFF
--- a/charts/add-ons/grafana/templates/grafana.yaml
+++ b/charts/add-ons/grafana/templates/grafana.yaml
@@ -121,6 +121,10 @@ spec:
       - env:
         - name: GF_PATHS_DATA
           value: /data
+        # Force using the go-based DNS resolver instead of the OS' to avoid failures in some environments
+        # see https://github.com/grafana/grafana/issues/20096
+        - name: GODEBUG
+          value: netdns=go
         image: {{.Values.image.name}}:{{default .Values.global.linkerdVersion .Values.global.controllerImageVersion}}
         imagePullPolicy: {{.Values.global.imagePullPolicy}}
         livenessProbe:
--- a/cli/cmd/testdata/install_addon_control-plane.golden
+++ b/cli/cmd/testdata/install_addon_control-plane.golden
@@ -2597,6 +2597,10 @@ spec:
       - env:
         - name: GF_PATHS_DATA
           value: /data
+        # Force using the go-based DNS resolver instead of the OS' to avoid failures in some environments
+        # see https://github.com/grafana/grafana/issues/20096
+        - name: GODEBUG
+          value: netdns=go
         image: gcr.io/linkerd-io/grafana:install-control-plane-version
         imagePullPolicy: IfNotPresent
         livenessProbe:
--- a/cli/cmd/testdata/install_control-plane.golden
+++ b/cli/cmd/testdata/install_control-plane.golden
@@ -2139,6 +2139,10 @@ spec:
       - env:
         - name: GF_PATHS_DATA
           value: /data
+        # Force using the go-based DNS resolver instead of the OS' to avoid failures in some environments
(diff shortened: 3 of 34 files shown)


alpeb:
@cpretzer this is erroring for me. When clicking on the Grafana link for the linkerd namespace in the dashboard, I get this:
image
One thing I see is that in that first failed request we have `series?match%5B%5D` while in the previous version the brackes were unencoded (`series?match[]`). Not sure if that's the issue...

cpretzer:
@alpeb I tested again with `bin/install-pr 4600` and the grafana dashboard displayed correctly.

The requests are being made with encoded brackets, so I think that's not an issue.
Screen Shot 2020-06-15 at 10 08 34

I was able to reproduce the display and errors that you saw when I `ctrl+c` the `linkerd dashboard` command. Is it possible that the dashboard command wasn't running when you tested?

Screen Shot 2020-06-15 at 10 09 28
Screen Shot 2020-06-15 at 10 09 37

alpeb:
From my tests, Grafana isn't able to communicate with Prometheus:
```
t=2020-06-15T21:38:38+0000 lvl=eror msg="Data proxy error" logger=data-proxy-log userId=0 orgId=1 uname= path=/api/datasources/proxy/1/api/v1/query_range remote_addr=127.0.0.1 referer=http://localhost:50750/grafana/ error="http: proxy error: dial tcp: lookup linkerd-prometheus.linkerd.svc.cluster.local: no such host"
```
Would love to have a third tester for this!

cpretzer:
@alpeb thanks for continuing to review. This is really strange.

I checked the instance that we looked at on our call today and the logs don't have that error, or any others.

A third tester would make me very happy as well 😄

alpeb:
Okey... In version 6.4, Grafana switched their base image from Ubuntu to Alpine. This caused some DNS resolving trouble for some users. It appears that only happens in some scenarios like mine (I'm running desktop Linux). Although I couldn't reproduce the problem in GKE nor AKS. I didn't test EKS but grafana/grafana#20096 reports such failure with that provider. So as of Grafana 6.5 they're providing two images, one based on Alpine and another on Ubuntu grafana/grafana#20178. The latter solves this issue, but it's a larger image. Another fix, that worked for me too, is to keep the default Alpine-based one, but add an environment variable that forces using the go DNS resolver instead of the OS's. I'm proposing that change below.

alpeb:
This is the proposed change:
```diff
diff --git a/charts/add-ons/grafana/templates/grafana.yaml b/charts/add-ons/grafana/templates/grafana.yaml
index ac80b6e3..5d875741 100644
--- a/charts/add-ons/grafana/templates/grafana.yaml
+++ b/charts/add-ons/grafana/templates/grafana.yaml
@@ -121,6 +121,10 @@ spec:
       - env:
         - name: GF_PATHS_DATA
           value: /data
+        # Force using the go-based DNS resolver instead of the OS' to avoid failures in some environments
+        # see https://github.com/grafana/grafana/issues/20096
+        - name: GODEBUG
+          value: netdns=go
         image: {{.Values.image.name}}:{{default .Values.global.linkerdVersion .Values.global.controllerImageVersion}}
         imagePullPolicy: {{.Values.global.imagePullPolicy}}
         livenessProbe:
```

cpretzer:
thanks @alpeb! I'm adding this change now.

Great find!

PR #4602, opened 2020-06-15
(WIP) Deep copy for per component resources

cypherfox:
This is a variant to #4226, with the aim to use deepCopy instead of more complex copying of helm subtrees for generating the proper values.


alpeb:
It turns out the reason of the failure is that linkerd is still using and old Helm dependency. You should be able to upgrade it safely to v2.16.8 [here](https://github.com/linkerd/linkerd2/blob/d10ed2aa5e4fa12936fc73c4be377ae2a58bfa53/go.mod#L57) to be able to make use of `deepCopy`.

cypherfox:
I have updated the version of the go dependency, the function still is not known. I cannot find another place where a specific helm version.

alpeb:
It turns out there's a hard-coded dependency towards an older sprig library [here](https://github.com/linkerd/linkerd2/blob/fe71ef04b04c632ccfd6218aa3492ac83c87b573/go.mod#L8). If you delete that line and do `go mod tidy` again it should automatically grab the the proper version required by the newer helm dependency.

charts/linkerd2/templates/controller.yaml
```diff
@@ -100,7 +100,11 @@ spec:
         volumeMounts:
         - mountPath: /var/run/linkerd/config
           name: config
-      - {{- include "partials.proxy" . | indent 8 | trimPrefix (repeat 7 " ") }}
+      {{- $tree := deepCopy . }}      
+      {{- if not (empty .Values.publicAPIProxyResources) }}
+      {{- $r := merge .Values.publicAPIProxyResources .Values.global.proxy.resources }}
+      {{- $_ := set $tree.Values.global.proxy "resources" $r }}
```
alpeb, line 106: There's a missing `{{- end }}` here that is preventing the chart to render properly.

cypherfox:
@alpeb if you are fine with the changes (especially the dependency changes), I would merge them back into #4226 and close this one.

go.mod
```diff
@@ -4,61 +4,76 @@ go 1.14
...
+	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.6.0
+	github.com/codegangsta/cli => github.com/urfave/cli v1.22.4
+	github.com/containerd/containerd v1.3.0-0.20190507210959-7c1e88399ec0 => github.com/containerd/containerd v1.3.0
+	github.com/docker/docker v1.14.0-0.20190319215453-e7b5f7dbe98c => github.com/docker/docker v1.13.0
+	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc8
+	github.com/tonistiigi/fifo => github.com/containerd/fifo v0.0.0
+	github.com/uber-go/atomic => go.uber.org/atomic v1.6.0
+	github.com/wercker/stern => github.com/linkerd/stern v0.0.0-20200331220320-37779ceb2c32
+	google.golang.org/cloud => cloud.google.com/go v0.0.0
+)
```
alpeb, line 79: This change might be an artifact from the previous changes. If you revert it back to 
```
replace github.com/wercker/stern => github.com/linkerd/stern v0.0.0-20200331220320-37779ceb2c32
```
then `go mod tidy` should leave this part alone (also gotta run `bin/update-go-deps-shas` one more time of course).

cypherfox:
Closed as the change is merged into #4226

PR #4608, opened 2020-06-16
Add destination-get-networks option

zaharidichev:
In #4585 we are observing an issue where a loop is encountered when using nginx ingress. The problem is that the outbound proxy does a dst lookup on the IP address which happens to be the very same address the ingress is listening on.

In order to avoid situations like that this PR introduces a way to modify the set of networks for which the proxy shall do IP based discovery. The change introduces a helm flag `.Values.global.proxy.destinationGetNetworks` that can be used to modify this value. There are two ways a user can affect the this setting: 

- setting the `destinationGetNetworks` field in values during a Helm install, which changes the default on all injected pods
- using an annotation ` config.linkerd.io/proxy-destination-get-networks` for injected workloads to override this value

Note that this setting cannot be tweaked through the `install` or `inject` command

Fix: #4585


cli/cmd/root.go
```diff
@@ -226,6 +227,7 @@ type proxyConfigOptions struct {
 	debugImageVersion             string
 	dockerRegistry                string
 	imagePullPolicy               string
+	destinationGetNetworks        []string
```
olix0r, line 230: I think for now we should omit this from the CLI flags, leaving it only as an annotation or helm setting.

zaharidichev: Okey, we can do that. Would that mean that we do not want it in the config map or this is not related. @alpeb WDYT ?

olix0r: I think it would be preferable to omit it, but I think it's ultimately an implementation detail.

zaharidichev:
Ok @olix0r I have removed that flag from the CLI but by the looks of it it makes sense to leave that setting in the config map. This allows us to leave the Helm flag as well as override what the default is via an annotation. Moreover if we leave it out of the cli _for now_, this allows us to easily expose it whenever we want. Let me know if all that works.

zaharidichev:
@Pothulapati Yes I guess this is the byproduct of trating empty strings the same way we treat the NON presence of a value. Go's type system is all sorts of great.....

Pothulapati:
@zaharidichev The empty string case works well for me now. But when I do `-oyaml`, I get the following 
```yaml
  - env:
    - name: LINKERD2_PROXY_LOG
      value: warn,linkerd=info
    - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
      value: linkerd-dst.linkerd.svc.cluster.local:8086
    - name: LINKERD2_PROXY_DESTINATION_GET_NETWORKS
    - name: LINKERD2_PROXY_CONTROL_LISTEN_ADDR
      value: 0.0.0.0:4190
```

How about having a if condition to only set if not empty in the helm yaml templating to make it more cleaner? :thinking: like we do for other env's

zaharidichev:
@Pothulapati I think that makes sense yes

zaharidichev:
@Pothulapati I addressed your feedback :)

cli/cmd/inject.go
```diff
@@ -437,6 +437,12 @@ func (options *proxyConfigOptions) overrideConfigs(configs *cfg.All, overrideAnn
 		overrideAnnotations[k8s.ProxyRequireIdentityOnInboundPortsAnnotation] = strings.Join(options.requireIdentityOnInboundPorts, ",")
 	}
 
+	if len(options.destinationGetNetworks) > 0 {
```
adleong, line 440: since there is no inject flag for this, this code is unreachable, right?

cli/cmd/install.go
```diff
@@ -778,6 +779,7 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*l5d
 	installValues.SMIMetrics.Enabled = options.smiMetricsEnabled
 
 	installValues.Global.Proxy = &l5dcharts.Proxy{
+		DestinationGetNetworks: strings.Join(options.destinationGetNetworks, ","),
```
adleong, line 782: this is basically just here for consistency, right?  In practice this will always be the default value.

zaharidichev: Yes, correct

zaharidichev:
@adleong I tweaked the PR description

PR #4610, opened 2020-06-16
Fix install-pr script

kleimkuhler:
`bin/kind-load --images` expects images to be located in `image-archives/`.

This fixes the `install-pr` script to unzip the images into that directory.

### DIFF
--- a/test/install_test.go
+++ b/test/install_test.go
@@ -655,7 +655,7 @@ func TestInject(t *testing.T) {
 }
 
 func TestServiceProfileDeploy(t *testing.T) {
-	bbProto, err := TestHelper.HTTPGetURL("https://raw.githubusercontent.com/BuoyantIO/bb/55b78f210c98a7eb431c3f555cf0a406f0b12edf/api.proto")
+	bbProto, err := TestHelper.HTTPGetURL("https://raw.githubusercontent.com/BuoyantIO/bb/v0.0.5/api.proto")
 	if err != nil {
 		testutil.AnnotatedFatalf(t, "unexpected error",
 			"unexpected error: %v %s", err, bbProto)
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -80,7 +80,7 @@ cd "$dir" || exit
 echo "### Downloading images ###"
 
 curl -L -o archive.zip -H "$auth" "$archive"
-unzip -o archive.zip
+unzip -o archive.zip -d image-archives/
 
 echo "### Loading images into Docker ###"
 
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -84,7 +84,7 @@ unzip -o archive.zip -d image-archives/
 
 echo "### Loading images into Docker ###"
 
-image=$(docker load -i cli-bin.tar | sed 's/Loaded image: //')
+image=$(docker load -i image-archives/cli-bin.tar | sed 's/Loaded image: //')
 tag=$(echo "$image" | cut -f 2 -d ":")
 
 if [ $is_kind ]
@@ -99,7 +99,7 @@ then
 else
   for image in cni-plugin controller debug grafana proxy web
   do
-    image=$(docker load -i "$image.tar" | sed 's/Loaded image: //')
+    image=$(docker load -i "image-archives/$image.tar" | sed 's/Loaded image: //')
     docker push "$image"
   done
 fi


kleimkuhler:
@cpretzer 👍

alpeb:
I think `bin/kind-load` is failing because it uses `TAG=${TAG:-$(head_root_tag)}` so it assumes we're under the source directory but in this case we're in a temporary dir. I see the `fatal: not a git repository (or any of the parent directories): .git` error as seen below, although it ends up working, probably because the PR I tested with had its images already pushed in gcr.io

```
$ GITHUB_TOKEN=xxx bin/install-pr -k 4610                                                                                                                                                                                                    
### Downloading images ###                                                                                                                                                                                                                                                        
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                                                                                                                                                                                                   
                                 Dload  Upload   Total   Spent    Left  Speed                                                                                                                                                                                                     
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0                                                                                                                                                                                                    
100  358M    0  358M    0     0  6054k      0 --:--:--  0:01:00 --:--:-- 7268k                                                                                                                                                                                                    
Archive:  archive.zip                                                                                                                                                                                                                                                             
  inflating: image-archives/cli-bin.tar                                                                                                                                                                                                                                           
  inflating: image-archives/cni-plugin.tar                                                                                                                                                                                                                                        
  inflating: image-archives/controller.tar                                                                                                                                                                                                                                        
  inflating: image-archives/debug.tar                                                                                                                                                                                                                                             
  inflating: image-archives/grafana.tar                                                                                                                                                                                                                                           
  inflating: image-archives/proxy.tar                                                                                                                                                                                                                                             
  inflating: image-archives/web.tar                                                                                                                                                                                                                                               
### Loading images into Docker ###                                                                                                       
fatal: not a git repository (or any of the parent directories): .git    
fatal: not a git repository (or any of the parent directories): .git           
kind v0.8.1 go1.14.2 linux/amd64         
/home/alpeb/src/prs/4610/linkerd2 
```

test/install_test.go
```diff
@@ -655,7 +655,7 @@ func TestInject(t *testing.T) {
 }
 
 func TestServiceProfileDeploy(t *testing.T) {
-	bbProto, err := TestHelper.HTTPGetURL("https://raw.githubusercontent.com/BuoyantIO/bb/master/api.proto")
+	bbProto, err := TestHelper.HTTPGetURL("https://raw.githubusercontent.com/BuoyantIO/bb/55b78f210c98a7eb431c3f555cf0a406f0b12edf/api.proto")
```
adleong, line 658: this might look nicer if there's a release tag we can target instead of an arbitrary looking sha

cpretzer: Good idea. changed to the latest tagged version: `https://raw.githubusercontent.com/BuoyantIO/bb/v0.0.5/api.proto`

cpretzer:
> I think `bin/kind-load` is failing because it uses `TAG=${TAG:-$(head_root_tag)}` so it assumes we're under the source directory but in this case we're in a temporary dir. I see the `fatal: not a git repository (or any of the parent directories): .git` error as seen below, although it ends up working, probably because the PR I tested with had its images already pushed in gcr.io

@alpeb I see this, too. WDYT of passing the `image` and `tag` from `install-pr` to `kind-load`?

Then `kind-load` can check whether those values have been passed as parameters. If they have, then do nothing, if they haven't, then resolve them using `head_root_tag`.

Probably should do this in another PR

kleimkuhler:
@cpretzer I think there are simpler changes we can make so this error doesn't happen! I'd prefer to not have these errors showing up as part of this change 😬.

As it was pointed out, `kind-load` fails because it always sets `TAG` [here](https://github.com/linkerd/linkerd2/blob/5c104ebec6d590c22b62fe62b7a0919a5d3d8407/bin/kind-load#L66).

As I'm looking at, I'm realizing we don't actually need to set `TAG` if we are loading from an `images-host` as done [here](https://github.com/linkerd/linkerd2/blob/5c104ebec6d590c22b62fe62b7a0919a5d3d8407/bin/kind-load#L77-L78). It's actually incorrect! The images we are saving are not related to `TAG` at all since they are not part of the working tree; they should probably just be saved as their name (`cli-bin`, `controller`, etc.).

I think a simpler change to fix is this just set `TAG` if we are not loading from a directory:

```
if [ -z "$images"]; then
  TAG=${TAG:-$(head_root_tag)}
fi
```

And line 77 would stop using `TAG`:

```
DOCKER_HOST=$images_host docker save "$DOCKER_REGISTRY/$img:$TAG" > image-archives/$img.tar
```

PR #4627, opened 2020-06-18
test-cleanup: Make populate_array() bash 3-friendly

### LINKED ISSUE #4621: Test cleanup does not work on MacOS
### What is the issue?
The test cleanup script does not work on a mac

### How can it be reproduced?
Run bin/test-cleanup

### Logs, error output, etc
```
$bin/test-cleanup
cleaning up control-plane namespaces in k8s-context []
bin/test-cleanup: line 14: local: -n: invalid option
local: usage: local name[=value] ...
```

joakimr-axis:
Legacy versions of bash (used in e.g. Mac OS) do not have the [nameref](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameters.html) functionality.
This patch replaces the use of that in the `populate_array` function and uses a bash 3-friendly way of handing this instead.

([Kubernetes](https://github.com/kubernetes/kubernetes) developers will recognize this bash 3-friendly way from [kube::util::read-array](https://github.com/kubernetes/kubernetes/blob/d8febccacfc9d51a017be9531247689e0e36df04/hack/lib/util.sh#L755-L770) in the Kubernetes code base.)

Fixes #4621

### DIFF
--- a/bin/test-cleanup
+++ b/bin/test-cleanup
@@ -15,7 +15,7 @@ Usage: ${FUNCNAME[0]} TARGET_ARRAY RESOURCE_TYPE LINKERD_LABEL (FRIENDLY_NAME)" 
   # shellcheck disable=SC2034
   local i=0
   unset -v "$1"
-  while IFS= read -r "$1[i++]"; do :; done <<< "$(kubectl --context="$k8s_context" get "$2" -oname -l "linkerd.io/$3")"
+  while IFS= read -r "$1[i++]"; do :; done <<< "$(kubectl --context="$k8s_context" get -A "$2" -oname -l "linkerd.io/$3")"
   # ensure that the last element isn't empty
   eval "[[ \${$1[--i]} ]]" || unset "$1[i]" # ensures last element isn't empty
   # inform user if no results were returned from command
--- a/bin/test-cleanup
+++ b/bin/test-cleanup
@@ -11,11 +11,15 @@ Usage: ${FUNCNAME[0]} TARGET_ARRAY RESOURCE_TYPE LINKERD_LABEL (FRIENDLY_NAME)" 
     return
   }
 
-  local -n res=$1
-  while IFS= read -r line; do res+=("$line"); done <<< "$(kubectl --context="$k8s_context" get "$2" -oname -l "linkerd.io/$3")"
-  if [ -z "${res[*]}" ]; then
-    echo "no ${4:-$2} found" >&2
-  fi
+  # We are using the variable i even if shellsheck does not see it
+  # shellcheck disable=SC2034
+  local i=0
+  unset -v "$1"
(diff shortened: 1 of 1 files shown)


joakimr-axis:
> @joakimr-axis I tried that after running an install test and I am afraid that it still does not work for me

The check for empty results is not working properly. I'm on it.

joakimr-axis:
@zaharidichev, please try it out now. At least the github CI pipeline seems happy with the fix I did since your last comment.

zaharidichev:
@joakimr-axis I wonder whether #4628 is an alternative for the changes happening here ?

joakimr-axis:
> This now works well for me! Thanks a lot! Works on both older and newer versions of bash.

Thanks for your verification!

bin/test-cleanup
```diff
@@ -11,11 +11,15 @@ Usage: ${FUNCNAME[0]} TARGET_ARRAY RESOURCE_TYPE LINKERD_LABEL (FRIENDLY_NAME)"
...
-  local -n res=$1
-  while IFS= read -r line; do res+=("$line"); done <<< "$(kubectl --context="$k8s_context" get "$2" -oname -l "linkerd.io/$3")"
-  if [ -z "${res[*]}" ]; then
-    echo "no ${4:-$2} found" >&2
-  fi
+  # We are using the variable i even if shellsheck does not see it
+  # shellcheck disable=SC2034
+  local i=0
+  unset -v "$1"
+  while IFS= read -r "$1[i++]"; do :; done <<< "$(kubectl --context="$k8s_context" get "$2" -oname -l "linkerd.io/$3")"
```
kleimkuhler, line 18: I noticed while testing this that a Linkerd rolebinding ends up in `kube-system` and is not removed:

```
❯ kubectl get -A rolebindings 
NAMESPACE     NAME                                                ROLE                                                  AGE
..
kube-system   linkerd-linkerd-tap-auth-reader                     Role/extension-apiserver-authentication-reader        103s
..
```

This is an existing issue so can we just fix it here by checking all namespaces?

```suggestion
  while IFS= read -r "$1[i++]"; do :; done <<< "$(kubectl --context="$k8s_context" get -A "$2" -oname -l "linkerd.io/$3")"
```

kleimkuhler:
@joakimr-axis Thanks for the attribution! I think you'll need to `git commit --amend` though and add a `Signed-off-by` as well so that the DCO check passes. You'll probably need to do a force push afterwards.

joakimr-axis:
> @joakimr-axis Thanks for the attribution! I think you'll need to `git commit --amend` though and add a `Signed-off-by` as well so that the DCO check passes. You'll probably need to do a force push afterwards.

Yup. It is so strange that manual thing is needed when I explicitly press the button to take in the suggested change and thus of course approve it.

PR #4628, opened 2020-06-18
Improve shellscript portability by using /bin/env

joakimr-axis:
Using `/bin/env` increases portability for the shell scripts (and often using `/bin/env` is requested by e.g. Mac users). This would also facilitate testing scripts with different Bash versions via the Bash containers, as they have bash in `/usr/local` and not `/bin`. Using `/bin/env`, there is no need to change the script when testing. (I assume the latter was behind https://github.com/linkerd/linkerd2/pull/4593/files/c301ea214b7ccf8d74d7c41cbf8c4cc05fea7d4a#diff-ecec5e3a811f60bc2739019004fa35b0, which would not happen using `/bin/env`.)

### DIFF
--- a/bin/_docker.sh
+++ b/bin/_docker.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eu
 
--- a/bin/_log.sh
+++ b/bin/_log.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 set -eu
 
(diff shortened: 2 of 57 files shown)


joakimr-axis:
> I'd be okay with this. I think we should still merge #4627 if a backwards compatible solution exists.

I totally agree with that.

> If we choose not to merge #4627 in favor of this, there is still a step where all Linkerd devs on MacOS will need to install their own version of bash for things to work properly.

Yes, and it is much better if they can just run it out-of-the-box.

PR #4641, opened 2020-06-19
Post-2.8.0 integration test cleanup

alpeb:
We had some code for testing upgrades from pre-2.8.0 stables that took
care of creating the non-existent `linkerd-smi-metrics` SA, which is no
longer necessary.


alpeb:
Ah right, thanks!

PR #4653, opened 2020-06-23
Add a Windows Linkerd cli Test

### LINKED ISSUE #4431: ci: Test windows cli

Pothulapati:
Fixes #4431 

This PR adds a new cli test to see if installation yamls are correctly
generated even on windows, this is important because of all the file
path difference between windows and Linux, and if any code uses a wrong
format might cause the chart generation commands to fail on windows.

This creates a separate workflow for both release, and PR tests, and
downloads the windows binary and runs the test.

Also, all the exisiting integration tests are moved in to
`/tests/integration` to separate from `/test/cli` as this test does not fall
under integration tests category


.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,33 @@ jobs:
         name: image-archives
         path: /home/runner/archives
   # todo: Keep in sync with `release.yml`
+  cli_windows_static_integration_tests:
+    name: Integration tests (windows)
```
Pothulapati, line 72: can rename this to CLI tests too, as we did the same separation in `/test/` too

Pothulapati: Updated

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,33 @@ jobs:
...
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Download image archives
+      # actions/download-artifact@v1
+      uses: actions/download-artifact@18f0f59
+      with:
+        name: image-archives
+    - name: Install CLI
+      run: |
+        cp image-archives\cli\windows\linkerd $HOME\linkerd.exe
```
alpeb, line 94: Nit: I think we can save ourselves this step by feeding `image-archives\cli\windows\linkerd` directly into `--linkerd` below (IMO the smaller these build files are, the better :wink: )

test/cli/cli_install_static_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if !*runTests {
+		exit(0, "cli tests not enabled: enable with -cli-tests")
+	}
+
+	if *linkerd == "" {
+		exit(1, "-linkerd flag is required")
+	}
+
+	TestHelper = testutil.NewGenericTestHelper(*linkerd, "", "l5d", "", "", "", "", "", "", false, false, *http.DefaultClient, testutil.KubernetesHelper{})
+	os.Exit(testutil.Run(m, TestHelper))
```
alpeb, line 38: `testutil.Run()` attempts to fetch the logs from the cluster whenever there's a failure. Since there's no cluster here, you can just do the old `os.Exit(m.Run())` instead.

.github/workflows/kind_integration.yml
```diff
@@ -50,6 +50,9 @@ jobs:
           docker save "$DOCKER_REGISTRY/$image:$TAG" > $ARCHIVES/$image.tar || tee save_fail &
         done
 
+        # save cli's as artifacts
+        cp -r ./target/cli $ARCHIVES/cli
```
alpeb, line 54: We end up with two copies of the cli binaries uploaded as artifacts, the other copy being inside the docker images. But we do this so we don't have to instantiate a docker image in order to retrieve them below in the windows test (like we do in the kind tests below), right?

.github/workflows/kind_integration.yml
Pothulapati: Hmm, The main reason for doing this was to add `.exe` without which windows acts weirdly. Should we just rename it?

.github/workflows/kind_integration.yml
Pothulapati: Yeah, Loading linux build docker images don't seem to work on windows. They err out mentioning that its a different ref format.

.github/workflows/kind_integration.yml
alpeb: Oh I see. In that case you can leave it as is :+1:

test/integration/testdata/custom_cluster_domain_config.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Cluster
```
kleimkuhler, line 1: I think this is a duplicate of what is in `test/integration/configs/cluster-domain.yaml`

Pothulapati: Yep, It is. So, I updated `release.yaml` to use `test/integration/configs/cluster-domain.yaml` but I don't see a similar test in `kind_integration.yaml`. So, are we running these custom domain tests only for release? :)

kleimkuhler: Opened #4668 to fix this separately from this change!

.github/workflows/release.yml
```diff
@@ -75,6 +75,33 @@ jobs:
...
+      uses: actions/checkout@722adc6
+    - name: Try to load cached Go modules
+      # actions/cache@v1.1.2
+      uses: actions/cache@70655ec
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Download image archives
```
Pothulapati, line 94: @alpeb @kleimkuhler Just realised we dont actually push artifacts to Github during release right? which means that this would fail?

alpeb: Shoot, that's right :-(

kleimkuhler: Yes that's correct we don't push artifacts in the release workflow so this would be an issue..

I think this could do something similar to `cloud_integration` where it installs from the locally built images: https://github.com/linkerd/linkerd2/blob/4372ed56dd39f79ba4fa8286e3f95d4383786f5d/.github/workflows/cloud_integration.yml#L89-L90?

.github/workflows/release.yml
```diff
@@ -75,6 +75,32 @@ jobs:
...
+      uses: actions/cache@70655ec
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Install linkerd CLI
+      id: install_cli
+      run: |
+        $image="gcr.io/linkerd-io/cli-bin:$TAG"
```
kleimkuhler, line 97: Each job needs to set it's own environment variables. `$TAG` hasn't been set in `windows_static_cli_tests` yet, so you'll need another copy of the `Set environment variables from scripts` step in this job!

.github/workflows/release.yml
```diff
@@ -75,6 +75,32 @@ jobs:
...
+    - name: Try to load cached Go modules
+      # actions/cache@v1.1.2
+      uses: actions/cache@70655ec
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Install linkerd CLI
+      id: install_cli
```
kleimkuhler, line 95: I don't think you need this `id` field.

```suggestion
```

I think we only need this field when a future step needs to reference a previous one. For example, `Create GKE cluster` uses `install_cli` step ID which is set in a previous step.

.github/workflows/release.yml
Pothulapati: Ah, @alpeb and I were expecting them to be present throughout the workflow and he is testing it. We previously went with the main tag as its also pushed everytime but then switched over to $TAG.

The problem with running is the script is that its a windows machine and that bash script would not work! :\

alpeb: @Pothulapati note though that [you can use bash in windows as well](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell). Worth a shot if the TAG env var is not available :wink:

alpeb: I tested in a fork using this minimal workflow:
```yaml
name: Release
on:
  push:
    branches:
    - master
env:
  GH_ANNOTATION: true
  TAG: ""
jobs:
  set_env_var:
    name: Setting TAG
    runs-on: ubuntu-18.04
    steps:
    - name: Checkout code
      # actions/checkout@v2
      uses: actions/checkout@722adc6
    - name: Set environment variables from scripts
      run: |
        . bin/_tag.sh
        echo ::set-env name=TAG::$(CI_FORCE_CLEAN=1 bin/root-tag)
  get_env_var:
    name: Getting TAG (linux)
    needs: [set_env_var]
    runs-on: ubuntu-18.04
    steps:
    - name: Check TAG using bash
      shell: bash
      run: |
        echo "TAG:"
        echo $TAG
  windows_static_cli_tests:
    name: Getting TAG (windows)
    needs: [set_env_var]
    runs-on: windows-latest
    steps:
    - name: Check TAG using bash
      shell: bash
      run: |
        echo "TAG:"
        echo $TAG
    - name: Check TAG using PS
      run: |
        echo "TAG:"
        echo $env:TAG
    - name: Checkout code
      # actions/checkout@v2
      uses: actions/checkout@722adc6
    - name: Run bin/root-tag
      shell: bash
      run: bin/root-tag
```
Results [here](https://github.com/alpeb/linkerd2/runs/812705443?check_suite_focus=true).

As @Pothulapati noticed before, global env vars are only for static values; attempting to change them via `echo ::set-env name=TAG::...` doesn't work. That only changes the value for the current job and the value isn't carried over the following jobs, regardless of them using linux or windows (PS or bash).
However, if we set the shell to be `bash`, we can run `bin/root-tag` to retrieve the tag!

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,39 @@ jobs:
...
+      # actions/download-artifact@v1
+      uses: actions/download-artifact@18f0f59
+      with:
+        name: image-archives
+    - name: Install linkerd CLI
+      shell: bash
+      run: |
+        echo ::set-env name=TAG::$(CI_FORCE_CLEAN=1 bin/root-tag)
+        echo $TAG
+        image="gcr.io/linkerd-io/cli-bin:main"
```
kleimkuhler, line 97: Have that other docker issue going on, but will want to use `$TAG` here!

```suggestion
        image="gcr.io/linkerd-io/cli-bin:$TAG"
```

Pothulapati: yeah, totally. Just want to check if these commands actually work! wanted to switch to $TAG once I see things work.

Pothulapati:
Updated both `release.yaml` & `kind_integration.yaml` to be the same by adding create and upload `artifacts` to `release.yaml`.

.github/workflows/release.yml
```diff
@@ -75,6 +106,33 @@ jobs:
...
+      uses: actions/download-artifact@18f0f59
+      with:
+        name: image-archives
+    - name: Install CLI
+      run: |
+        cp image-archives\cli\windows\linkerd $HOME\linkerd.exe
+    - name: Run CLI Integration tests
+      run: |
+        go test --failfast --mod=readonly ".\test\cli" --linkerd=$HOME\linkerd.exe --cli-tests -v
   kind_integration_tests:
```
alpeb, line 136: Can you please leave the `Keep in sync with `kind_integration.yml` comment right above `kind_integration_tests` as well?

.github/workflows/release.yml
```diff
@@ -37,6 +41,33 @@ jobs:
...
+          docker save "$DOCKER_REGISTRY/$image:$TAG" > $ARCHIVES/$image.tar || tee save_fail &
+        done
+
+        # save cli's as artifacts
+        cp -r ./target/cli $ARCHIVES/cli
+
+        # Wait for `docker save` background processes to complete. Exit early
+        # if any job failed.
+        wait < <(jobs -p)
+        test -f save_fail && exit 1 || true
```
alpeb, line 60: IIUC we only need the CLI image, so you can simplify all this with a single `docker save` and there's no need to do process waiting business.

Pothulapati:
@alpeb Yep! That makes total sense. Updated the code to not have a dependency on the docker images.

.github/workflows/kind_integration.yml
```diff
@@ -65,6 +68,30 @@ jobs:
...
+          ${{ runner.os }}-go-
+    - name: Download image archives
+      # actions/download-artifact@v1
+      uses: actions/download-artifact@18f0f59
+      with:
+        name: image-archives
+    - name: Run CLI Integration tests
+      run: |
+        go test --failfast --mod=readonly ".\test\cli" --linkerd=$PWD\image-archives\linkerd-windows.exe --cli-tests -v
   kind_integration_tests:
```
alpeb, line 95: Can you add "# todo: Keep in sync with `release.yml`" back here (while leaving it for `windows_static_cli_tests`).

PR #4657, opened 2020-06-23
go-run: Move temporary binary into `target` directory

olix0r:
The `bin/go-run` script generates a temporary binary, stored in the root
of the repository.

This change moves it into `target/` so that is included in the
.dockerignore, and so that the repo can be cleaned easily by removing
the `target/` directory.


adleong:
This change should also remove it from .gitignore

PR #4663, opened 2020-06-24
EndpointSlice support (#4501)

Matei207:
### What
---

* This PR aims to introduce support for `EndpointSlices` the K8s resource succeeding `Endpoints`
* The changes are around adding the basic logic to replace `Endpoints` with `EndpointSlices` on the `EndpointsWatcher` side of discovery service
* The PoC linked at the top should contain most information on how this has been thought out, along with the initial feedback received

### How
---

* The first step has been to add support of the EndpointSlice resource to `k8s/api.go`, as well as a method that checks whether the cluster has the EndpointSlice resource registered. (EndpointSlice sits behind a feature gate)
* The next step has been to modify and/or create handler functions to process the resource in the destination service.
* *Add*: when adding a new EndpointSlice resource, we are reusing the same methods previously used by the `Endpoints` resource by introducing type assertions. 
   - There are not many considerations and differences between the two, however, the fact that all EndpointSlices make up an Endpoints resource needed to be taken into account when adding/removing addresses associated with a service and port (`portPublisher: AddressSet`). 
    - `EndpointSlices` can be orphaned, but only through human error. Whenever an EndpointSlice is created by K8s it will either have a label with the service name or an owner reference to that service (or both).
  - `EndpointSlices` have a `Ready` condition for their endpoints, endpoints that are not ready are not processed.
  - `EndpointSlices` can have an addressType of: IPv4, IPv6 and host/FQDN. After the feedback on the poc I've decided to only allow IPv4 endpoints to be processed at this time, dual-stack requires a bit more diligence and experimentation.
  - Furthermore, `EndpointSlices` as a type contains many pointers (e.g Port numbers as pointers). 
  - Last but not least, `EndpointSlices` have a generated name based on the associated service, to get a service ID we need to look at the label OR owner reference.
* *Remove*: when removing an EndpointSlice resource, we are using completely different methods than the Endpoints counterpart. When adding, it is easier to use type assertions because of certain similarities on how the resources are processed; deletions are processed entirely different. Notable in this case is that we can no longer delete all addresses associated with a svc:port, simply because an EndpointSlice would not delete all endpoints associated but an Endpoints resource would (ES1 U ES2 ... U ESN = Endpoints).
* Other changes come from breaking stuff up and *attempting* to refactor.
* Tests have been added, mainly by replicating what has already been done and considering additional cases. Unit test scenarios covered:
   - Local services w/ EndpointSlice
   - Local services w/ missing addresses
   - Local services w/ no endpoints
   - Services that do not exist
   - External service
   - Stateful sets
   - EndpointSlices with no label (should not be processed, can't get service ID)
   - EndpointSlices with an addressType != IPv6
   - Deletion

### Feedback and open questions:

* In the spirit of learning and improving the quality of this PR, I would very much appreciate feedback related to the code quality (and quality of tests) as well as feedback on the performance of the code. Because of the differences between the two resources, you'll notice a lot of loops spawned, if there are any corners I can cut, let me know. I am also interested to see if there are any architectural decisions that can be improved.

* I do have some open questions as well:
  1. We are checking whether an EndpointSlice does not have an owner (which can be a label or an owner reference). For simplicity sake, should we not check the owner reference? Whenever a new port publisher is created, all of its endpoints are updated -- in this case, we rely on a label selector to query all endpoint slices associated with the port publisher's service. If an endpoint does not have a label but has an owner, the port publisher will not pick it up anyway, so I suppose relying on owner reference should not be done. If an EndpointSlice does not have a label then it should not be processed. (note by "a label" I'm referring to the service name label specifically).
 2. When I manually tested the code, I noticed a slight drop in performance -- not sure if this is introduced by this PR, or it's a K8s thing. The noticed delay is when an EndpointSlice is added.
 3. When an EndpointSlice is added, we add the new addresses to the port publisher's existing address set. What are some thoughts on this, should we consolidate all slices or is it an unnecessary step I took because of a misunderstanding.

Thank you for going through this and be as critical as possible, please!

### DIFF
--- a/controller/api/destination/watcher/endpoints_watcher.go
+++ b/controller/api/destination/watcher/endpoints_watcher.go
@@ -7,11 +7,13 @@ import (
 	"sync"
 
 	"github.com/linkerd/linkerd2/controller/k8s"
-	consts "github.com/linkerd/linkerd2/pkg/k8s"
+	pkgK8s "github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/prometheus/client_golang/prometheus"
 	logging "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/cache"
 )
@@ -122,7 +124,9 @@ type (
 var endpointsVecs = newEndpointsMetricsVecs()
 
 // NewEndpointsWatcher creates an EndpointsWatcher and begins watching the
-// k8sAPI for pod, service, and endpoint changes.
+// k8sAPI for pod, service, and endpoint changes. An EndpointsWatcher will
+// watch on Endpoints or EndpointSlice resources, depending on cluster configuration.
+//TODO: Allow EndpointSlice resources to be used once opt-in functionality is supported.
 func NewEndpointsWatcher(k8sAPI *k8s.API, log *logging.Entry) *EndpointsWatcher {
 	ew := &EndpointsWatcher{
 		publishers: make(map[ServiceID]*servicePublisher),
@@ -145,12 +149,22 @@ func NewEndpointsWatcher(k8sAPI *k8s.API, log *logging.Entry) *EndpointsWatcher 
 		UpdateFunc: func(_, obj interface{}) { ew.addService(obj) },
 	})
 
-	k8sAPI.Endpoint().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    ew.addEndpoints,
-		DeleteFunc: ew.deleteEndpoints,
-		UpdateFunc: func(_, obj interface{}) { ew.addEndpoints(obj) },
-	})
-
+	if !pkgK8s.EndpointSliceAccess(k8sAPI.Client) {
+		// ew.log.Debugf("Cluster does not have EndpointSlice access:%v", err)
+		ew.log.Debugf("Watching Endpoints resources")
+		k8sAPI.Endpoint().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    ew.addEndpoints,
+			DeleteFunc: ew.deleteEndpoints,
+			UpdateFunc: func(_, obj interface{}) { ew.addEndpoints(obj) },
+		})
+	} else {
+		ew.log.Debugf("Watching EndpointSlice resources")
+		k8sAPI.ES().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    ew.addEndpointSlice,
+			DeleteFunc: ew.deleteEndpointSlice,
+			UpdateFunc: ew.updateEndpointSlice,
+		})
+	}
 	return ew
 }
 
@@ -240,17 +254,17 @@ func (ew *EndpointsWatcher) deleteService(obj interface{}) {
 }
 
 func (ew *EndpointsWatcher) addEndpoints(obj interface{}) {
-	endpoints := obj.(*corev1.Endpoints)
+	endpoints, ok := obj.(*corev1.Endpoints)
+	if !ok {
+		ew.log.Errorf("error processing endpoints resource, got %#v expected *corev1.Endpoints", obj)
+		return
+	}
+
 	if endpoints.Namespace == kubeSystem {
 		return
 	}
-	id := ServiceID{
-		Namespace: endpoints.Namespace,
-		Name:      endpoints.Name,
-	}
-
+	id := ServiceID{endpoints.Namespace, endpoints.Name}
 	sp := ew.getOrNewServicePublisher(id)
-
 	sp.updateEndpoints(endpoints)
 }
 
@@ -280,6 +294,84 @@ func (ew *EndpointsWatcher) deleteEndpoints(obj interface{}) {
 	sp, ok := ew.getServicePublisher(id)
 	if ok {
 		sp.deleteEndpoints()
+	}
+}
+
+func (ew *EndpointsWatcher) addEndpointSlice(obj interface{}) {
+	newSlice, ok := obj.(*discovery.EndpointSlice)
+	if !ok {
+		ew.log.Errorf("error processing EndpointSlice resource, got %#v expected *discovery.EndpointSlice", obj)
+		return
+	}
+
+	if newSlice.Namespace == kubeSystem {
+		return
+	}
+
+	id, err := getEndpointSliceServiceID(newSlice)
+	if err != nil {
+		ew.log.Errorf("Could not fetch resource service name:%v", err)
+		return
+	}
+
+	sp := ew.getOrNewServicePublisher(id)
+	sp.addEndpointSlice(newSlice)
+}
+
+func (ew *EndpointsWatcher) updateEndpointSlice(oldObj interface{}, newObj interface{}) {
+	oldSlice, ok := oldObj.(*discovery.EndpointSlice)
+	if !ok {
+		ew.log.Errorf("error processing EndpointSlice resource, got %#v expected *discovery.EndpointSlice", oldObj)
+		return
+	}
+	newSlice, ok := newObj.(*discovery.EndpointSlice)
+	if !ok {
+		ew.log.Errorf("error processing EndpointSlice resource, got %#v expected *discovery.EndpointSlice", newObj)
+		return
+	}
+
+	if newSlice.Namespace == kubeSystem {
+		return
+	}
+
+	id, err := getEndpointSliceServiceID(newSlice)
+	if err != nil {
+		ew.log.Errorf("Could not fetch resource service name:%v", err)
+		return
+	}
+
+	sp, ok := ew.getServicePublisher(id)
+	if ok {
+		sp.updateEndpointSlice(oldSlice, newSlice)
+	}
+}
+
+func (ew *EndpointsWatcher) deleteEndpointSlice(obj interface{}) {
+	es, ok := obj.(*discovery.EndpointSlice)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			ew.log.Errorf("couldn't get object from DeletedFinalStateUnknown %#v", obj)
+		}
+		es, ok = tombstone.Obj.(*discovery.EndpointSlice)
+		if !ok {
+			ew.log.Errorf("DeletedFinalStateUnknown contained object that is not an EndpointSlice %#v", obj)
+			return
+		}
+	}
+
+	if es.Namespace == kubeSystem {
+		return
+	}
+
+	id, err := getEndpointSliceServiceID(es)
+	if err != nil {
+		ew.log.Errorf("Could not fetch resource service name:%v", err)
+	}
+
+	sp, ok := ew.getServicePublisher(id)
+	if ok {
+		sp.deleteEndpointSlice(es)
 	}
 }
 
@@ -323,7 +415,6 @@ func (sp *servicePublisher) updateEndpoints(newEndpoints *corev1.Endpoints) {
 	sp.Lock()
 	defer sp.Unlock()
 	sp.log.Debugf("Updating endpoints for %s", sp.id)
-
 	for _, port := range sp.ports {
 		port.updateEndpoints(newEndpoints)
 	}
@@ -333,9 +424,35 @@ func (sp *servicePublisher) deleteEndpoints() {
 	sp.Lock()
 	defer sp.Unlock()
 	sp.log.Debugf("Deleting endpoints for %s", sp.id)
-
 	for _, port := range sp.ports {
 		port.noEndpoints(false)
+	}
+}
+
+func (sp *servicePublisher) addEndpointSlice(newSlice *discovery.EndpointSlice) {
+	sp.Lock()
+	defer sp.Unlock()
+	sp.log.Debugf("Adding EndpointSlice for %s", sp.id)
+	for _, port := range sp.ports {
+		port.addEndpointSlice(newSlice)
+	}
+}
+
+func (sp *servicePublisher) updateEndpointSlice(oldSlice *discovery.EndpointSlice, newSlice *discovery.EndpointSlice) {
+	sp.Lock()
+	defer sp.Unlock()
+	sp.log.Debugf("Updating EndpointSlice for %s", sp.id)
+	for _, port := range sp.ports {
+		port.updateEndpointSlice(oldSlice, newSlice)
+	}
+}
+
+func (sp *servicePublisher) deleteEndpointSlice(es *discovery.EndpointSlice) {
+	sp.Lock()
+	defer sp.Unlock()
+	sp.log.Debugf("Deleting EndpointSlice for %s", sp.id)
+	for _, port := range sp.ports {
+		port.deleteEndpointSlice(es)
 	}
 }
 
@@ -411,12 +528,29 @@ func (sp *servicePublisher) newPortPublisher(srcPort Port, hostname string) *por
 		metrics:    endpointsVecs.newEndpointsMetrics(sp.metricsLabels(srcPort, hostname)),
 	}
 
-	endpoints, err := sp.k8sAPI.Endpoint().Lister().Endpoints(sp.id.Namespace).Get(sp.id.Name)
-	if err != nil && !apierrors.IsNotFound(err) {
-		sp.log.Errorf("error getting endpoints: %s", err)
-	}
-	if err == nil {
-		port.updateEndpoints(endpoints)
+	if !pkgK8s.EndpointSliceAccess(sp.k8sAPI.Client) {
+		// sp.log.Debugf("No EndpointSlice access, using endpoints:%v", err)
+		endpoints, err := sp.k8sAPI.Endpoint().Lister().Endpoints(sp.id.Namespace).Get(sp.id.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			sp.log.Errorf("error getting endpoints: %s", err)
+		}
+		if err == nil {
+			port.updateEndpoints(endpoints)
+		}
+	} else {
+		sp.log.Debugf("Using EndpointSlice")
+		matchLabels := map[string]string{discovery.LabelServiceName: sp.id.Name}
+		selector := k8slabels.Set(matchLabels).AsSelector()
+
+		sliceList, err := sp.k8sAPI.ES().Lister().EndpointSlices(sp.id.Namespace).List(selector)
+		if err != nil && !apierrors.IsNotFound(err) {
+			sp.log.Errorf("error getting endpointSlice list: %s", err)
+		}
+		if err == nil {
+			for _, slice := range sliceList {
+				port.addEndpointSlice(slice)
+			}
+		}
 	}
 
 	return port
@@ -451,21 +585,92 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
 			}
 		}
 	}
-	pp.exists = true
 	pp.addresses = newAddressSet
-
+	pp.exists = true
 	pp.metrics.incUpdates()
 	pp.metrics.setPods(len(pp.addresses.Addresses))
 	pp.metrics.setExists(true)
 }
 
-func metricLabels(endpoints *corev1.Endpoints) map[string]string {
-	labels := map[string]string{service: endpoints.Name, namespace: endpoints.Namespace}
+func (pp *portPublisher) addEndpointSlice(slice *discovery.EndpointSlice) {
+	newAddressSet := pp.endpointSliceToAddresses(slice)
+	for id, addr := range pp.addresses.Addresses {
+		newAddressSet.Addresses[id] = addr
+	}
 
-	gateway, hasRemoteGateway := endpoints.Labels[consts.RemoteGatewayNameLabel]
-	gatewayNs, hasRemoteGatwayNs := endpoints.Labels[consts.RemoteGatewayNsLabel]
-	remoteClusterName, hasRemoteClusterName := endpoints.Labels[consts.RemoteClusterNameLabel]
-	serviceFqn, hasServiceFqn := endpoints.Annotations[consts.RemoteServiceFqName]
+	add, _ := diffAddresses(pp.addresses, newAddressSet)
+	if len(add.Addresses) > 0 {
+		for _, listener := range pp.listeners {
+			listener.Add(add)
+		}
+	}
+
+	pp.addresses = newAddressSet
+	pp.exists = true
+	pp.metrics.incUpdates()
+	pp.metrics.setPods(len(pp.addresses.Addresses))
+	pp.metrics.setExists(true)
+}
+
+func (pp *portPublisher) updateEndpointSlice(oldSlice *discovery.EndpointSlice, newSlice *discovery.EndpointSlice) {
+	updatedAddressSet := AddressSet{
+		Addresses: make(map[ID]Address),
+		Labels:    pp.addresses.Labels,
+	}
+
+	for id, address := range pp.addresses.Addresses {
+		updatedAddressSet.Addresses[id] = address
+	}
+
+	oldAddressSet := pp.endpointSliceToAddresses(oldSlice)
+	for id := range oldAddressSet.Addresses {
+		delete(updatedAddressSet.Addresses, id)
+	}
+
+	newAddressSet := pp.endpointSliceToAddresses(newSlice)
+	for id, address := range newAddressSet.Addresses {
+		updatedAddressSet.Addresses[id] = address
+	}
+
+	add, remove := diffAddresses(pp.addresses, updatedAddressSet)
+	for _, listener := range pp.listeners {
+		if len(remove.Addresses) > 0 {
+			listener.Remove(remove)
+		}
+		if len(add.Addresses) > 0 {
+			listener.Add(add)
+		}
+	}
+
+	pp.addresses = updatedAddressSet
+	pp.exists = true
+	pp.metrics.incUpdates()
+	pp.metrics.setPods(len(pp.addresses.Addresses))
+	pp.metrics.setExists(true)
+}
+
+func metricLabels(resource interface{}) map[string]string {
+	var serviceName, ns string
+	var resLabels, resAnnotations map[string]string
+	switch res := resource.(type) {
+	case *corev1.Endpoints:
+		{
+			serviceName, ns = res.Name, res.Namespace
+			resLabels, resAnnotations = res.Labels, res.Annotations
+		}
+	case *discovery.EndpointSlice:
+		{
+			serviceName, ns = res.Labels[discovery.LabelServiceName], res.Namespace
+			resLabels, resAnnotations = res.Labels, res.Annotations
+		}
+	}
+
+	labels := map[string]string{service: serviceName, namespace: ns}
+
+	gateway, hasRemoteGateway := resLabels[pkgK8s.RemoteGatewayNameLabel]
+	gatewayNs, hasRemoteGatwayNs := resLabels[pkgK8s.RemoteGatewayNsLabel]
+	remoteClusterName, hasRemoteClusterName := resLabels[pkgK8s.RemoteClusterNameLabel]
+	serviceFqn, hasServiceFqn := resAnnotations[pkgK8s.RemoteServiceFqName]
 
 	if hasRemoteGateway && hasRemoteGatwayNs && hasRemoteClusterName && hasServiceFqn {
 		// this means we are looking at Endpoints created for the purpose of mirroring
@@ -483,6 +688,61 @@ func metricLabels(endpoints *corev1.Endpoints) map[string]string {
 	return labels
 }
 
+func (pp *portPublisher) endpointSliceToAddresses(es *discovery.EndpointSlice) AddressSet {
+	addresses := make(map[ID]Address)
+	resolvedPort := pp.resolveESTargetPort(es.Ports)
+	if resolvedPort == Port(0) {
+		return AddressSet{}
+	}
+	serviceID, err := getEndpointSliceServiceID(es)
+	if err != nil {
+		pp.log.Errorf("Could not fetch resource service name:%v", err)
+	}
+
+	for _, endpoint := range es.Endpoints {
+		if endpoint.Hostname != nil {
+			if pp.hostname != "" && pp.hostname != *endpoint.Hostname {
+				continue
+			}
+		}
+		if endpoint.Conditions.Ready != nil && !*endpoint.Conditions.Ready {
+			continue
+		}
+
+		if endpoint.TargetRef == nil {
+			for _, IPAddr := range endpoint.Addresses {
+				var authorityOverride string
+				if fqName, ok := es.Annotations[pkgK8s.RemoteServiceFqName]; ok {
+					authorityOverride = fmt.Sprintf("%s:%d", fqName, pp.srcPort)
+				}
+
+				identity := es.Annotations[pkgK8s.RemoteGatewayIdentity]
+				address, id := pp.newServiceRefAddress(resolvedPort, IPAddr, serviceID.Name, es.Namespace)
+				address.Identity, address.AuthorityOverride = authorityOverride, identity
+				addresses[id] = address
+			}
+
+			continue
+		}
+
+		if endpoint.TargetRef.Kind == "Pod" {
+			for _, IPAddr := range endpoint.Addresses {
+				address, id, err := pp.newPodRefAddress(resolvedPort, IPAddr, endpoint.TargetRef.Name, endpoint.TargetRef.Namespace)
+				if err != nil {
+					pp.log.Errorf("Unable to create new address:%v", err)
+					continue
+				}
+				addresses[id] = address
+			}
+		}
+
+	}
+	return AddressSet{
+		Addresses: addresses,
+		Labels:    metricLabels(es),
+	}
+}
+
 func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) AddressSet {
 	addresses := make(map[ID]Address)
 	for _, subset := range endpoints.Subsets {
@@ -491,47 +751,28 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) Addre
 			if pp.hostname != "" && pp.hostname != endpoint.Hostname {
 				continue
 			}
-			if endpoint.TargetRef == nil {
-				id := ServiceID{
-					Name: strings.Join([]string{
-						endpoints.ObjectMeta.Name,
-						endpoint.IP,
-						fmt.Sprint(resolvedPort),
-					}, "-"),
-					Namespace: endpoints.ObjectMeta.Namespace,
-				}
 
+			if endpoint.TargetRef == nil {
 				var authorityOverride string
-				if fqName, ok := endpoints.Annotations[consts.RemoteServiceFqName]; ok {
+				if fqName, ok := endpoints.Annotations[pkgK8s.RemoteServiceFqName]; ok {
 					authorityOverride = fmt.Sprintf("%s:%d", fqName, pp.srcPort)
 				}
 
-				addresses[id] = Address{
-					IP:                endpoint.IP,
-					Port:              resolvedPort,
-					Identity:          endpoints.Annotations[consts.RemoteGatewayIdentity],
-					AuthorityOverride: authorityOverride,
-				}
+				identity := endpoints.Annotations[pkgK8s.RemoteGatewayIdentity]
+				address, id := pp.newServiceRefAddress(resolvedPort, endpoint.IP, endpoints.Name, endpoints.Namespace)
+				address.Identity, address.AuthorityOverride = identity, authorityOverride
+
+				addresses[id] = address
 				continue
 			}
+
 			if endpoint.TargetRef.Kind == "Pod" {
-				id := PodID{
-					Name:      endpoint.TargetRef.Name,
-					Namespace: endpoint.TargetRef.Namespace,
-				}
-				pod, err := pp.k8sAPI.Pod().Lister().Pods(id.Namespace).Get(id.Name)
+				address, id, err := pp.newPodRefAddress(resolvedPort, endpoint.IP, endpoint.TargetRef.Name, endpoint.TargetRef.Namespace)
 				if err != nil {
-					pp.log.Errorf("Unable to fetch pod %v: %s", id, err)
+					pp.log.Errorf("Unable to create new address:%v", err)
 					continue
 				}
-				ownerKind, ownerName := pp.k8sAPI.GetOwnerKindAndName(pod, false)
-				addresses[id] = Address{
-					IP:        endpoint.IP,
-					Port:      resolvedPort,
-					Pod:       pod,
-					OwnerName: ownerName,
-					OwnerKind: ownerKind,
-				}
+				addresses[id] = address
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +938,49 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func isValidSlice(es *dv1beta1.EndpointSlice) bool {
+	if es.AddressType != dv1beta1.AddressTypeIPv4 {
```
adleong, line 960: I know that more work is required to properly support IPv6 but do we need to add code to skip non-IPv4 addresses?  This just seems like one more thing we'll need to remove when we properly support IPv6.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +938,49 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
+	if !ok && len(es.OwnerReferences) == 0 {
+		return false
+	} else if len(es.OwnerReferences) == 0 && serviceName == "" {
+		return false
+	}
+
+	return true
+}
+
+func getEndpointSliceSVCName(es *dv1beta1.EndpointSlice) (string, error) {
```
adleong, line 974: Could we save some work elsewhere by having this return a ServiceID instead of a string?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -319,14 +385,26 @@ func (ew *EndpointsWatcher) getServicePublisher(id ServiceID) (sp *servicePublis
...
+	switch res := newResource.(type) {
+	case *corev1.Endpoints:
+		sp.log.Debugf("Updating endpoints for %s", sp.id)
+		for _, port := range sp.ports {
+			port.updateEndpoints(newResource)
+		}
+	case *dv1beta1.EndpointSlice:
+		sp.log.Debugf("Updating endpoint slice for %s", sp.id)
+		for _, port := range sp.ports {
+			if !isTargetPortInSlice(port.targetPort, res.Ports) {
```
adleong, line 401: Technically different EndpointSlice objects for a service could have different ports defined, but this seems strange to me and I'm not sure exactly what it means.  My assumption is that EndpointsSlices would be sliced up by address, meaning that each EndpointSlice for a service would have the same ports defined but a different set of addresses.

Does this match your understanding?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -411,12 +502,27 @@ func (sp *servicePublisher) newPortPublisher(srcPort Port, hostname string) *por
 		metrics:    endpointsVecs.newEndpointsMetrics(sp.metricsLabels(srcPort, hostname)),
 	}
 
-	endpoints, err := sp.k8sAPI.Endpoint().Lister().Endpoints(sp.id.Namespace).Get(sp.id.Name)
-	if err != nil && !apierrors.IsNotFound(err) {
-		sp.log.Errorf("error getting endpoints: %s", err)
-	}
-	if err == nil {
-		port.updateEndpoints(endpoints)
+	if !sp.k8sAPI.EndpointSliceAccess() {
```
adleong, line 505: nit: I'd flip the branches of this if-else to get rid of the negation in the condition.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -451,38 +571,130 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
...
+}
 
-	pp.metrics.incUpdates()
-	pp.metrics.setPods(len(pp.addresses.Addresses))
-	pp.metrics.setExists(true)
+func (pp *portPublisher) updateEndpointSlicesSet(endpointslice *dv1beta1.EndpointSlice) {
+	newAddressSet := pp.endpointSliceToAddresses(endpointslice)
+	if len(newAddressSet.Addresses) == 0 {
+		for _, listener := range pp.listeners {
+			listener.NoEndpoints(true)
```
adleong, line 581: I'm not sure the logic is correct here for dealing with individual EndpointSlice objects.  If a single EndpointSlice object has no addresses, I think there still could be other EndpointSlice objects for that service that do have addresses.  So I don't think a NoEndpoints update is appropriate.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -451,38 +571,130 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
...
+		}
+	} else {
+		add, remove := diffAddresses(pp.addresses, newAddressSet)
```
adleong, line 584: It looks like when we see an EndpointSlice add or update, we are treating that EndpointSlice as the entire set of addresses when actually there could be other EndpointSlice objects which exist for that service.

I think what we need to do is slightly more complex:
* For EndpointSlice adds, we can update `pp.addresses` by adding the addresses in the EndpointSlice
* For updates, we probably want to start by removing all the addresses in the old EndpointSlice from `pp.addresses` and then adding the ones in the new EndpointsSlice

In either case, we'd want to do a diff of the previous `pp.addresses` to the current `pp.addresses` to determine which updates to send to the listener.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -565,6 +806,21 @@ func (pp *portPublisher) updatePort(targetPort namedPort) {
 	}
 }
 
+func (pp *portPublisher) deleteEndpointSlice(es *dv1beta1.EndpointSlice) {
+	addrSet := pp.endpointSliceToAddresses(es)
+	for id := range pp.addresses.Addresses {
+		if _, found := addrSet.Addresses[id]; found {
+			delete(pp.addresses.Addresses, id)
```
adleong, line 813: don't we also need to send a `Remove` update to the listener with the deleted addresses?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -23,11 +25,11 @@ const (
 	// metrics labels
 	service                = "service"
 	namespace              = "namespace"
-	targetGatewayNamespace = "target_gateway_namespace"
-	targetGateway          = "target_gateway"
-	targetCluster          = "target_cluster"
-	targetService          = "target_service"
-	targetServiceNamespace = "target_service_namespace"
+	remoteGatewayNamespace = "remote_gateway_namespace"
```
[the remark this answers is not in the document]
Matei207: yup, this happened before I opened the PR. I thought it would be a good idea to rebase the changes from master, then I remembered it had changed to main, rebased both and did not notice the new additions until I opened the PR.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -11,7 +11,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	logging "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	dv1beta1 "k8s.io/api/discovery/v1beta1"
```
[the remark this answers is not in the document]
Matei207: changed, thank you. it does sound better!

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -144,13 +146,22 @@ func NewEndpointsWatcher(k8sAPI *k8s.API, log *logging.Entry) *EndpointsWatcher
...
 		UpdateFunc: func(_, obj interface{}) { ew.addService(obj) },
 	})
-
-	k8sAPI.Endpoint().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    ew.addEndpoints,
-		DeleteFunc: ew.deleteEndpoints,
-		UpdateFunc: func(_, obj interface{}) { ew.addEndpoints(obj) },
-	})
-
+	acc := k8sAPI.EndpointSliceAccess()
```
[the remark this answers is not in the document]
zaharidichev: But does that actually make sense? As per K8s docs: 

>  Note: Although EndpointSlices may eventually replace Endpoints, many Kubernetes components still rely on Endpoints. For now, enabling EndpointSlices should be seen as an addition to Endpoints in a cluster, not a replacement for them.

I think if the feature is enabled we shoud support slices _as well_

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: +1 on that I think one of the main benefits of slices is that you can have multiple objects for a service. So this is certainly a valid case.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -451,38 +571,130 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
...
+			if pp.hostname != "" && pp.hostname != *endpoint.Hostname {
+				continue
+			}
+		}
+		if !*endpoint.Conditions.Ready {
+			continue
+		}
+
+		if endpoint.TargetRef == nil {
+			serviceName, err := getEndpointSliceSVCName(es)
```
zaharidichev, line 659: We can move that out of the loop is guess. No point of doing it for every endpoint as it will not change.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -565,6 +806,21 @@ func (pp *portPublisher) updatePort(targetPort namedPort) {
...
+		if _, found := addrSet.Addresses[id]; found {
+			delete(pp.addresses.Addresses, id)
+		}
+	}
+
+	if len(pp.addresses.Addresses) > 0 {
+		return
+	}
+
+	pp.noEndpoints(false)
```
zaharidichev, line 821: Hm, is this here correct? If a slice is deleted, you are calling noEndpoints(false). Correct me if I am wrong but this should send an update saying that the svc_name:port does not exist anymore. If you delete an endpoint slice does that mean that we should be sending exists=false? @adleong wdyt ? Also, does that really mean that no Endpoints need to be send. There might be other endpoints in other slices. I have the feeling I might be getting all this wrong. Have to think about that a bit.

zaharidichev: So from the comments of the proxy api: 

```
// If the service does not exist then the controller must send
// `no_endpoints{exists: false}` ASAP when a client subscribes or when the
// service stops existing. If the service exists and has endpoints available
// then the controller must send `add` that lists all (or at least a large
// number) of the endpoints for the service. If and only if the service exists
// but does not have any endpoints available then the controller SHOULD send
// `no_endpoints{exists: true}` when a client subscribes. In other words, the
// `no_endpoints` message must only be sent when there are *no*endpoints for
// the service.
```

zaharidichev: Yes I think there are two problems here: 

- we cannot inform the proxy that there are no endpoints simply because a slice was deleted because there might be other slices that are associated with this service
- we certainly cannot inform the proxy that this service has no associated endpoints because of the above as well.

controller/api/destination/watcher/endpoints_watcher.go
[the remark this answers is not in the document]
Matei207: @zaharidichev it's an interesting point, and it's something I considered in the first drafts of the RFC. You can check out the very short conversation [here](https://github.com/linkerd/rfc/pull/23#discussion_r425328745).

Another thing to note is that @adleong asked in the RFC how the kube-proxy handles this situation. I looked at the Service Topology implementation PR, and it seems in kube-proxy they do a similar check to see which informer to fire up. You can check out how they instantiate the "watcher config" [here](https://github.com/m1093782566/kubernetes/blob/31d623b4d74ceadfad101d36a4710d4406fcf144/cmd/kube-proxy/app/server.go#L681).

I am not sure what the general consensus is, I suppose at the very least if we use both we would have to reconciliate between the two resources, I can see this leading to some race conditions and complications down the line. It might also be harder to migrate to EndpointSlices once the feature graduates (?). Either way, I am open to making all of the necessary changes, so let me know what you think!

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +938,49 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
+			if *p.Name == targetPort.StrVal {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func isValidSlice(es *dv1beta1.EndpointSlice) bool {
```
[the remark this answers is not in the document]
Matei207: I think that's a great idea! Changed.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: That actually makes a lot of sense, changed! Thank you!

controller/api/destination/watcher/endpoints_watcher.go
Matei207: The probability of an EndpointSlice with an IPv6 address type (or FQDN for that matter) should be quite low. I don't think any would be processed at the moment since IPv6 networking in Kubernetes is not supported out-of-the-box and itself sits behind feature gates, so I can remove this.

If anybody feels like this should be added back in, let me know and I'll handle it.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Sounds like a great idea, changed!

controller/api/destination/watcher/endpoints_watcher.go
Matei207: this fixed itself (in a way) after I changed the return type for `EndpointSliceAccess`

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Thanks for the detailed comment @zaharidichev it all makes sense! I was a bit confused around the `NoEndpoints` concept.

I wanted to send a `NoEndpoints` message only if the port publisher has no addresses registered anymore, however. The way I thought this was:
* becase a svc:port might have loads of addresses split across different slices, go through the current addresses for that port publisher and remove them.
* if at the end we do not have any addresses, send a no endpoints message.

I tried to do this with the line above:
```
if len(pp.addresses.Addresses) > 0 -- return out of the function
```

perhaps this is not the most intuitive way to do though, any thoughts?

controller/api/destination/watcher/endpoints_watcher.go
Matei207: this completely eluded me, added it in, thanks!

controller/api/destination/watcher/endpoints_watcher.go
Matei207: that is totally correct, I think I might have confused myself at the time. Thank you for spotting this, I have replaced the above with a guard clause to return out of the function in case the new address set has a length of zero. Would this need any other special consideration? wdyt

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Hm, that's interesting. Since I have started working on EndpointSlices, ports have been a major headache to deal with and there's quite a bit of confusion around it from my side.

The way I have been thinking about this has been based on the documentation on Endpoints in client-go. Specifically:
```
// EndpointSubset is a group of addresses with a common set of ports. The
// expanded set of endpoints is the Cartesian product of Addresses x Ports.
// For example, given:
//   {
//     Addresses: [{"ip": "10.10.1.1"}, {"ip": "10.10.2.2"}],
//     Ports:     [{"name": "a", "port": 8675}, {"name": "b", "port": 309}]
//   }
// The resulting set of endpoints can be viewed as:
//     a: [ 10.10.1.1:8675, 10.10.2.2:8675 ],
//     b: [ 10.10.1.1:309, 10.10.2.2:309 ]
```

They say an Endpoints object is the union of all subsets, where a subset is defined as a group of addresses with common ports. By this logic, I assumed an EndpointSlice is simply a subset -- a group of addresses (within a given limit, default 100 of them) that use the same ports. I have rarely encountered any endpoints with more than a subset, however, and replicating it has been a bit difficult, so I never saw my assumptions play out in minikube.

The only mention of subsets in EndpointSlices is:
```
// EndpointSlice represents a subset of the endpoints that implement a service.
// For a given service there may be multiple EndpointSlice objects, selected by
// labels, which must be joined to produce the full set of endpoints.
```

What do you think? Is there any easy way to test our assumptions or am I missing something very obvious?
Also, just to confirm: my assumption all along has been that EndpointSlices are sliced based on the common ports they share, so different slices would hold addresses that match that respective slice's port.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: This answers a few questions I had! Thank you both for your comments, I think this has been the trickiest part of adding support for this feature so far. Will change things around based on the feedback 👍

Matei207:
Thank you for the comments and feedback @zaharidichev @adleong. I have incorporated most of the suggested changes, for some I'm waiting to hear back from you and to see what the general consensus is (e.g Watcher firing up both the slice and endpoints informers at the same time). I will be changing the logic when adding/updating a single endpointslice to reflect all feedback, this might take a bit longer.

Let me know what you think of the changes I made, as well as any other questions, concerns etc. Thanks again!

controller/api/destination/watcher/endpoints_watcher.go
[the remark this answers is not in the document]
adleong: @Matei207 I think you're right.  Even if the cluster supports both Endpoints and EndpointSlices, I think it makes sense to use *only* EndpointSlices so that we don't need to try to reconcile two sources of truth.

controller/api/destination/watcher/endpoints_watcher.go
adleong: After thinking about this some more, I think you're right about this.

controller/api/destination/watcher/endpoints_watcher.go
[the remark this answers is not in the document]
zaharidichev: Ok yes, I agree if kube-proxy does that there is not point of trying to be smarter about it.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -240,18 +254,32 @@ func (ew *EndpointsWatcher) deleteService(obj interface{}) {
...
+			if err != nil {
+				ew.log.Errorf("Could not fetch service name for EndpointSlice [%s/%s]:%v", res.Namespace, res.Name, err)
+				return
+			}
+			sp = ew.getOrNewServicePublisher(id)
+		}
+	}
 
-	sp.updateEndpoints(endpoints)
+	sp.updateEndpoints(obj)
```
zaharidichev, line 282: Because Golang is amazing, you can pass some arbitrary interface here in this function that is NOT `discovery.EndpointSlice` or `corev1.Endpoints` and skip on both of the match arms, which would leave your sp uninitialized and you will git an NPE. I think you want a default case here that handles that  somehow.

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: I think it will be clearer if you do something like: 

```
svcExists := len(pp.addresses.Addresses) > 0
pp.noEndpoints(svcExists)
```

zaharidichev: But all that brings an interesting for me question. Namely does the fact that a service has no endpoints (and hence noe addresses) mean that it should be non existent for the purposes of discovery? I see that in `deleteEndpoints` we used to fire a NoEndpoints(false).  I guess this was designed like that for a reason. Coming from some previous discussions around services that do not have any endpoints and are there just as proxies (for mc purrposes), I think it is a valid scenario to do a discovery on a service that does not have any endpoints.. or is it not? @adleong wdyt? 

@Matei207  Note, this is not directly related to this PR. I am just thinking out loud. Your approach is correct for the time being.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -681,3 +935,51 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
 	}
 	return
 }
+
+func isTargetPortInSlice(targetPort namedPort, slicePorts []discovery.EndpointPort) bool {
+	for _, p := range slicePorts {
+		switch targetPort.Type {
+		case intstr.Int:
+			if *p.Port == targetPort.IntVal {
```
zaharidichev, line 943: What happens if this is nil? The docs say something intersting that I am not sure I can interpret correctly: 

```
	// ports specifies the list of network ports exposed by each endpoint in
	// this slice. Each port must have a unique name. When ports is empty, it
	// indicates that there are no defined ports. When a port is defined with a
	// nil port value, it indicates "all ports". Each slice may include a
	// maximum of 100 ports.
	// +optional
	// +listType=atomic
```

All ports?... not sure what they mean here. Any idea?

controller/api/destination/watcher/endpoints_watcher.go
Matei207: > I think it will be clearer if you do something like (...)

That's an amazing suggestion, thank you! It reads much better.

> But all that brings an interesting for me question (...) does the fact that a service has no endpoints (and hence no addresses) mean that it should be non-existent for the purposes of discovery?

No, at least I don't think so. This is a very good question, based on the test cases and the general functionality that is provided by the destination service, I think in the case of a service that does not have any endpoints available, listeners should still be able to subscribe to the service and receive a "NoEndpoints" message. Since on subscription the client (proxy) receives a stream of updates, it's only fair that a service be available for discovery even if no endpoints (addresses) exist; my assumption is that there might be a stream of updates propagated to the client at some point in the future (?), pods can come up (and disappear) at any time.

So @zaharidichev I agree with that it is valid to do a discovery on a service that does not have any endpoints, based on what I've seen so far.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Yup, you're completely right. Will change this asap.

I thought we would always get either a `discovery.EndpointSlice` or `corev1.Endpoints` since the function is registered to the handler, but perhaps it's better not to take any chances. I'll edit it so that each match arm has a separate call to updateEndpoints.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: @zaharidichev I looked into this and I couldn't find anything that makes sense. I'm a bit puzzled by it too, in all honesty.

I had to do some work around the nil port today, it's not related to the description but it might offer some insight. In my case, whenever an endpointslice had all of its associated pods, all of the addresses (i.e endpoints) and ports automatically got deleted and set to nil (because they use pointers instead of empty lists *sigh*).

I suppose in this case, an EndpointSlice with a nil `ports` field, targets all targetPorts of the service. So, going back on a previous explanation, an `EndpointSlice` is an endpoints subset. An endpoints subset is a grouping of addresses with  common ports. This means every slice (assuming none of the slices hit their endpoint capacity limit) would have a different set of ports, an EndpointSlice with nil ports then would mean ALL ports of the service? Perhaps for the rare case when a pod exposes all ports? Still, not sure why it has to be nil, but it might make sense? Lemme know what you think.

Matei207:
@adleong @zaharidichev sorry for the slight delay on this, but I ran into some bumps along the way. To start, I think (99% sure) I covered every piece of feedback that has been initially left on this PR, I think it's finally ready to be checked again. 

There are some changes that I've made that I will document here to make it easier to follow the differences. There are also some bits and pieces that I am pretty sure you will flag down, so I will comment directly on them to let you know why they're in there in the first place.

### Changes
---

* **Handler functions**: 
     - remove type assertions, by the end, the type assertions made it very difficult to follow what was going on, even for me, turns out they weren't saving much space in the file either so I went back to the original way code was organised
     - based on the feedback around updating/adding EndpointSlices, I thought separate methods for updating slices were needed, adding functions almost the same, updates function a bit differently, overall I thought a bit harder about the logic here and ironed out some pesky bugs

* **RBAC & Access**: 
   - Integration tests kept failing, that's because I forgot to add the roles, added new roles for endpointslices resources (discovery.k8s.io apigroup). Re-generated test files.
   - Cluster access function seemed to be a bit wonky. This is partly because I am using **v1.18.3** for Kubernetes (see next section). I suspect this is what made the integration tests fail as well (coupled with the RBAC rules). Basically, what happened here was that the `EndpointSlice` feature was set to FALSE, but the api server still had the resource registered... I thought the easiest way to fix this is to make sure the resource is registered in the cluster and there are some slices already in existence. This seemed to fix it, maybe there's a better solution out there?

* **K8s changes**: feature flag got broken up into separate pieces for `EndpointSlices` -- tl;dr one is for the kube-proxy, one is for the resource itself, additionally `EndpointSlices` will be on by default from `v1.18`. They're moving quite fast with it, and I'm not sure how this will impact people who are not familiar with `EndpointSlices`, perhaps in this case it's better to try and have both at the same time -- @adleong @zaharidichev 
You can read more about it [here](https://github.com/kubernetes/website/pull/19316).

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -673,11 +977,63 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
 		Labels:    newAddresses.Labels,
 	}
 	remove = AddressSet{
 		Addresses: removeAddresses,
 	}
 	return
 }
+
+func isTargetPortInSlice(targetPort namedPort, slice *discovery.EndpointSlice) bool {
+	if slice.Ports == nil && slice.Endpoints == nil {
```
Matei207, line 990: Sort of an edge case, let:
`web` -- a service
`pod-1` -- web-pod-1
`pod-2` -- web-pod-2
`s` -- slice for `web`: {`pod-1-IP`, `pod-2-IP`}

When we execute
`kubectl delete pod-1, pod-2`

Then we expect `NoEndpoints` for that slice. In actuality, if we don't include this check, we will be left with an endpoint that does not exist, but the application thinks exists.
1. Pod-1 gets deleted first -- s updates itself (removes pod-1-ip) --> dest-service sends a remove for pod-1-ip
2.1. Pod-2 gets deleted second -- s updates itself (removes pod-2-ip) --> s now has NIL PORT and NIL Endpoints => `isTargetPortInSlice` returns false (without this line) since the port is nil, the target port of the service is definitely not in there.
2.2. Pod-2 gets deleted second -- s updates itself (removes pod-2-ip) --> s now has NIL PORT and NIL Endpoints => `isTargetPortInSlice` returns true, the slice is processed, dest-service sends a remove for pod-2-ip.

Hope it makes sense.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -451,38 +572,173 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
...
 	pp.addresses = newAddressSet
+	pp.exists = true
+	pp.metrics.incUpdates()
+	pp.metrics.setPods(len(pp.addresses.Addresses))
+	pp.metrics.setExists(true)
+}
+
+func (pp *portPublisher) updateEndpointSlice(oldSlice *discovery.EndpointSlice, newSlice *discovery.EndpointSlice) {
+	newAddressSet, oldAddressSet := pp.endpointSliceToAddresses(newSlice), pp.endpointSliceToAddresses(oldSlice)
+	if len(newAddressSet.Addresses) == 0 && len(oldAddressSet.Addresses) == 0 {
```
Matei207, line 614: This is a very special case that I've come across when I was playing around. When we don't have any endpoints left and we update a slice with nil ports and nil endpoints with a slice that has nil ports and nil endpoints, the dest-service spams `NoEndpoints`. This checks avoids that.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -451,38 +572,173 @@ func (pp *portPublisher) updateEndpoints(endpoints *corev1.Endpoints) {
...
+		if len(remove.Addresses) > 0 {
+			listener.Remove(remove)
+		}
+		if len(add.Addresses) > 0 {
+			listener.Add(add)
+		}
+	}
+
+	pp.addresses = newAddressSet
+	if len(pp.addresses.Addresses) == 0 {
```
Matei207, line 639: If after updating we end up with no addresses (1 service, 1 slice, oldSlice = 1 IP address, newSlice = nil ports, nil endpoints, for example) we should send a `NoEndpoints`. So far this seems to work well, looking for some feedback here if this makes sense logically?

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -411,12 +493,30 @@ func (sp *servicePublisher) newPortPublisher(srcPort Port, hostname string) *por
...
+	if err != nil {
+		sp.log.Debugf("No EndpointSlice access, using endpoints:%v", err)
+		endpoints, err := sp.k8sAPI.Endpoint().Lister().Endpoints(sp.id.Namespace).Get(sp.id.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			sp.log.Errorf("error getting endpoints: %s", err)
+		}
+		if err == nil {
+			port.updateEndpoints(endpoints)
+		}
+	} else {
```
Matei207, line 527: While I agree that we should only have 1 informer of type Endpoints or 1 of type EndpointSlice, since EndpointSlice will be on by default, perhaps it's a good idea we always update and populate new publishers with Endpoints?

We can eliminate an extra step in the processing, the whole port gets updated/instantiated at once, and makes it easier to understand. We are not using informers here just the client, thoughts?

controller/api/destination/watcher/endpoints_watcher.go
adleong: In the Endpoints code, if we get an update with 0 addresses, we know that all addresses have been deleted so we send a NoEndpoints which I think is effectively a shorthand for "remove all".  In the the case of EndpointSlices the logic is more complicated because an updated slice with 0 addresses doesn't mean that there are no addresses for the whole service.  I think we can do ourselves a favor here and just not use the `NoEndpoints` shortcut at all and just rely on plain Adds and Removes for managing things.

controller/api/destination/watcher/endpoints_watcher.go
adleong: With my suggestion below to not use NoEndpoints, this won't be an issue.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -565,6 +850,24 @@ func (pp *portPublisher) updatePort(targetPort namedPort) {
...
+	remove := make(map[ID]Address)
+	for id := range pp.addresses.Addresses {
+		if _, found := addrSet.Addresses[id]; found {
+			delete(pp.addresses.Addresses, id)
+			remove[id] = addrSet.Addresses[id]
+		}
+	}
+
+	for _, listener := range pp.listeners {
+		listener.Remove(AddressSet{remove, addrSet.Labels})
```
adleong, line 864: Do we need to build the `remove` map at all?  I think extraneous removes (removes for addresses not already in the load balancer) will be ignored.  We should just be able to use addrSet here.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -655,16 +959,16 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
 	// TODO: this detects pods which have been added or removed, but does not
 	// detect addresses which have been modified.  A modified address should trigger
 	// an add of the new version.
-	addAddesses := make(map[ID]Address)
+	addAddresses := make(map[ID]Address)
```
adleong, line 962: 😱 thank you!

controller/api/destination/watcher/endpoints_watcher.go
adleong: I think if we push the port handling logic down into `resolveESTargetPort` we can avoid these kinds of edge cases.

controller/api/destination/watcher/endpoints_watcher.go
adleong: >The port number of the endpoint. If this is not specified, ports are not restricted and must be interpreted in the context of the specific consumer.

It sounds a bit underspecified.  Let's not worry too much about this.  We can come back and adjust this behavior if we run into a situation where it's problematic.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -673,11 +977,63 @@ func diffAddresses(oldAddresses, newAddresses AddressSet) (add, remove AddressSe
...
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func getEndpointSliceServiceID(es *discovery.EndpointSlice) (ServiceID, error) {
+	err := isValidSlice(es)
```
adleong, line 1011: up to you, but since this is the only place this is called, you could inline it here and that would remove some of the error handling boilerplate.

controller/api/destination/watcher/endpoints_watcher.go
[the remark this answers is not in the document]
Matei207: Yup, should be done now. Sorry it took a bit longer than expected.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -283,6 +292,66 @@ func (ew *EndpointsWatcher) deleteEndpoints(obj interface{}) {
 	}
 }
 
+func (ew *EndpointsWatcher) addEndpointSlice(obj interface{}) {
+	es := obj.(*discovery.EndpointSlice)
```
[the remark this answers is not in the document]
Matei207: Done. Added typesafe casts to `addEndpointSlice`, `updateEndpointSlice` and `addEndpoints`. The delete handlers have more logic around it and it doesn't make sense to do it here. Lmk if you want me to though!

controller/api/destination/watcher/endpoints_watcher.go
Matei207: That sounds good. The topology data slipped my mind, will opt for consistency and make all relevant changes. Thank you!

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Oh! This makes so much sense then. I've been under the impression constantly when testing that when all addresses are removed the `no endpoints` message should be sent. This is why I was so focused on putting it in there. Thank you for clarifying, it's more intuitive just doing it with adds and removes I think so will remove it (the no endpoints call).

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Yup, on second thought there is no need to build the `remove` map, so not sure why that was in place. Changed it around!

pkg/k8s/authz.go
```diff
@@ -105,6 +107,39 @@ func ServiceProfilesAccess(k8sClient kubernetes.Interface) error {
...
+			if apiRes.Kind == "EndpointSlice" {
+				return checkEndpointSlicesExist(k8sClient)
+			}
+		}
+	}
+
+	return errors.New("EndpointSlice resource not found")
+}
+
+func checkEndpointSlicesExist(k8sclient kubernetes.Interface) error {
```
adleong, line 130: I've been thinking a bit more about this and I think it would be a good idea to have the EndpointSlices functionality be opt-in.  This has a few advantages:

* it seems safer to let users opt-in when they know it has been enabled on their cluster rather than us trying to guess based on the existence of an EndpointSlices object
* it allows to us to release this sooner since it is explicitly opt-in and experimental
* we can change the default behavior in the future (perhaps once kube-proxy uses EndpointSlices by default)
* If we discover issues, this gives us a mechanism to fall back to Endpoints mode.

In practice, I see this as being an install flag (and helm variable) that enables EndpointSlices mode.  We should also include some validation that prevents EndpointSlices mode from being enabled on a cluster without the EndpointSlices API resource type.

In the interest of getting this merged and rolling forward, I think the safest thing to do would be to comment out the body of the EndpointSliceAccess function and hard-code it to return false, effectively making all of the EndpointSlice logic unreachable.  Then, in a separate PR, we can add the install flag which enables EndpointsSlice mode.  What do you think?

Matei207: @adleong it makes sense, and I think it's a good decision. I have been doubting the use of the access function lately, especially after I've read that in K8s v1.8 `EndpointSlices` are going to be switched on by default (but not in-use by kube-proxy).

Realistically, as much as I have tried to be thorough with both the unit tests and the manual tests, this needs to be played around with for a while so we can see where unconventional edge cases might be; I'm all for hardcoding the function to return a false and leaving it out for now.

Will make the change and maybe open up an issue in the next few days to get the ball rolling on how to enable them. So far I like the idea of an install flag, can maybe make it part of an edge release and hide the command (like the `alpha stat` cmd perhaps?).

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -283,6 +298,81 @@ func (ew *EndpointsWatcher) deleteEndpoints(obj interface{}) {
...
+	oldSlice, ok := oldObj.(*discovery.EndpointSlice)
+	if !ok {
+		ew.log.Errorf("error processing EndpointSlice resource, got %#v expected *discovery.EndpointSlice", oldObj)
+		return
+	}
+	newSlice, ok := newObj.(*discovery.EndpointSlice)
+	if !ok {
+		ew.log.Errorf("error processing EndpointSlice resource, got %#v expected *discovery.EndpointSlice", oldObj)
+		return
+	}
```
zaharidichev, line 330: I think you can refactor that logic in a separate method since you are repeating it here and in add as well. Also you are formatting with `%#v` while using `%T` in the add method

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -283,6 +298,81 @@ func (ew *EndpointsWatcher) deleteEndpoints(obj interface{}) {
...
+
+	if newSlice.Namespace == kubeSystem {
+		return
+	}
+
+	id, err := getEndpointSliceServiceID(newSlice)
+	if err != nil {
+		ew.log.Errorf("Could not fetch service name for EndpointSlice [%s/%s]:%v", newSlice.Namespace, newSlice.Name, err)
```
zaharidichev, line 338: Not sure there is a need to chain errors like that. Why not simply format the error you return from `getEndpointSliceServiceID` Seems like if you include the slice name and namespace it would give us all the info we need.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -541,6 +780,58 @@ func (pp *portPublisher) endpointsToAddresses(endpoints *corev1.Endpoints) Addre
...
+		Pod:       pod,
+		OwnerName: ownerName,
+		OwnerKind: ownerKind,
+	}
+
+	return addr, id, nil
+}
+
+func (pp *portPublisher) resolveESTargetPort(slicePorts []discovery.EndpointPort) Port {
+	if slicePorts == nil {
```
zaharidichev, line 818: WDYT about making that a bit more explicit. You can return a nil or an error instead of treating `Port(0)` as a special value. Is there any reason why you chose this approach?

controller/api/destination/watcher/endpoints_watcher.go
Matei207: This is a great suggestion! I have made the changes. 

* In addition to moving the `ns/name` identifiers in `getEndpointSliceServiceID`, I have also modified `isValidSlice` to return a less verbose error, so it can be chained in a better way in `getEndpointSliceServiceID`
* When we process the slice, the error message in my opinion should say that the slice could not be processed, it might be more helpful in the debugging process to know that's where the error was thrown (speaking from experience, at least).

Thank you, let me know if you see them done/structured in a different way and will make the changes :)

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Hmm, I looked at this based on your suggestion, but I'm not sure how well it can be extracted out into a separate function without losing context or having the same amount of lines in. I've left some thoughts below :) 

* If we extract out just the type casting, we would not save much in terms of space since the new function would still have to return an error or a bool in case the obj is not a slice; if we would not do that we can't exit early from the handler method.
* Although add and update work very similarily, for `add` we use `getOrNewServicePublisher`, for `update` we use `getServicePublisher`. This subtle difference makes it harder to extract the code.
* A third alternative is to create a function which returns a `*discovery.EndpointSlice` resource and a `ServiceID`, this is the approach I took. If the function does not fit well, I can take it out, however.

Let me know what you think.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Nothing that I can think of other than I thought it might be consistent. When there isn't a defined requested port, we return `Port(0)`, so thought it would add the context we need. I think `nil` is a good idea though! Error might be a bit too much in this instance, I would rather return a nil since this function is used just once in the whole file when we translate the slice addresses.

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: If that is the case, it makes sense to leave it as  is

controller/api/destination/watcher/endpoints_watcher.go
zaharidichev: Yes I think simply returning nil makes sense here. Error is too much

Matei207: Because of the type though it will have to return `Port(0)`. The underlying data type for `Port` is `uint32`, so we can't have a nil return type, unfortunately.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Cool, will roll back the changes in this case

adleong:
Now that https://github.com/linkerd/linkerd2/pull/4696 has merged, a merge or rebase of main should significantly shrink this diff.

controller/api/destination/watcher/endpoints_watcher.go
adleong: In a more sane programming language this would return an `Option[Port]`, but alas.  Using `Port(0)` as a special zero value that indicates that the port does not exist seems like a reasonably go-idiomatic thing to do.

Matei207:
> Now that #4696 has merged, a merge or rebase of main should significantly shrink this diff.

Rebased and pushed, only three files in the diff now! Thank you :)

Matei207:
If it is not introduced by this PR, I would be more than happy to have a look at it and see what can be done after the service topologies RFC is wrapped up! Let me know what you think the best approach would be in this case :)

PR #4666, opened 2020-06-25
cli: support url with addon-config flag

### LINKED ISSUE #4664: Support Urls with --addon-config flag
## Feature Request

Currently `--addon-config` takes a file path, but just like `linkerd inject` it should be able to take a URL and perform the same `install` or `upgrade`

Provide a more easier way for users to apply `--addon-config` without having to download file and apply.

same like `linkerd inject `

output?

Pothulapati:
Fixes #4664 

This PR makes `--addon-config` to also read from URL's.
Adds a new readFile func which can read from a file path or a url

**To Test:**

```bash
/bin/go-run cli install --ignore-cluster --addon-config https://gist.githubusercontent.com/Pothulapati/b8ed9472dac445f50b104365a54f8c37/raw/d84d52329f5941edacdd809958a3cfbc17a077b6/addon-config.yaml | code -       
```

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -1,12 +1,10 @@
 package cmd
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net/http"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


cli/cmd/install.go
```diff
@@ -1261,3 +1263,33 @@ func toIdentityContext(idvals *identityWithAnchorsAndTrustDomain) *pb.IdentityCo
 		Scheme:             idvals.Identity.Issuer.Scheme,
 	}
 }
+
+// Read a resource file from a path or URL
+func readFile(path string) ([]byte, error) {
+	if isValidURL(path) {
+		resp, err := http.Get(path)
```
zaharidichev, line 1270: I wonder whether we can extract the logic that is here https://github.com/linkerd/linkerd2/blob/main/cli/cmd/inject_util.go#L152 into something that we can reuse right here.

cli/cmd/install.go
```diff
@@ -1261,3 +1263,33 @@ func toIdentityContext(idvals *identityWithAnchorsAndTrustDomain) *pb.IdentityCo
...
+		}
+		resp.Body.Close()
+		return buf.Bytes(), nil
+	}
+
+	in, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	return in, nil
```
ihcsim, line 1293: This can be rewritten as `return ioutil.ReadFile(path)`.

cli/cmd/install.go
ihcsim: That's a good idea. Sounds like we might even be able to reuse the entire `read()` function.

PR #4675, opened 2020-06-26
Make uncheduble pods check warning only

### LINKED ISSUE #3940: ha: CLI checks are too aggressive
I have a running control plane, where all running instances are healthy, but there are unscheduled pods in a new control plane deployment:

```
:; linkerd stat po -n linkerd
Cannot find Linkerd: linkerd-controller-c845b4549-wbz2d: 0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
    linkerd-destination-6488b657b-5srp8: 0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
    linkerd-identity-7dc5d97cf9-rxnz2: 0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
    linkerd-proxy-injector-64d8d46794-lb69h: 0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
    linkerd-sp-validator-86bbcbfbff-sldx6: 0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
    linkerd-tap-c895b586d-ps22z: 0/3 nodes are available: 3 node(s) didn't match pod affinity/anti-affinity, 3 node(s) didn't satisfy existing pods anti-affinity rules.
```

This is a bit clunky. Notably, none of these deployments are necessary to serve a stat request. And furthermore, they all have running pods! Why should stat care about any of this? I suspect the same goes for other commands.

zaharidichev:
Currently commands that need access to the public api are executing the `LinkerdControlPlaneExistenceChecks` This set of checks includes one that specifically checks that there is no unscheduable pods. In fact in order to run commands like `stat` and `edge` we do not need to meet that requirement. 

This change relaxes all this by makind the `no unschedulable pods` a warning only check.  Fixes #3940

### DIFF
--- a/cli/cmd/public_api.go
+++ b/cli/cmd/public_api.go
@@ -50,13 +50,13 @@ func checkPublicAPIClientOrRetryOrExit(retryDeadline time.Time, apiChecks bool) 
 		checks = append(checks, healthcheck.LinkerdAPIChecks)
 	}
 
-	hc := newHealthChecker(checks, retryDeadline)
+	hc := newHealthChecker(checks, retryDeadline, true)
 
 	hc.RunChecks(exitOnError)
 	return hc.PublicAPIClient()
 }
 
-func newHealthChecker(checks []healthcheck.CategoryID, retryDeadline time.Time) *healthcheck.HealthChecker {
+func newHealthChecker(checks []healthcheck.CategoryID, retryDeadline time.Time, publicAPIClientChecks bool) *healthcheck.HealthChecker {
 	return healthcheck.NewHealthChecker(checks, &healthcheck.Options{
 		ControlPlaneNamespace: controlPlaneNamespace,
 		KubeConfig:            kubeconfigPath,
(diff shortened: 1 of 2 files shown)


zaharidichev:
@adleong yes that is a much simpler solution indeed.

zaharidichev:
@grampelberg brings a valid point [here](https://github.com/linkerd/linkerd2/issues/4680#issuecomment-650291814): 

> I'd like for stat to only check for its dependencies (eg prometheus and one 

That being said, should we maybe put together a separate bit of code that is more generic and lives outsides of the healtcheck logic that can check for different things depending on the cli command being invoked (and only be run for CLI commands). I propose that because there are number of CLI commands that hit this bit of code while instantiating the public api and they have different dependencies: 

- edges
- endpoints
- get
- inject
- multicluster
- routes
- stat


For example, I do not think inject needs... a prometheus instance... It needs to get the config which it does via the public api (so it needs the controller). This is all explicitly. However, it also needs the injector running inless its a manual injection. So yes.. not sure how deep we should go here. Maybe it deserves a design discussion

Issue #4572, opened 2020-06-08
Support grafana.image.version field


Pothulapati:
### How should the problem be solved?

PR #4571 removes the `grafana.image.version` field, thus restricting using our own image tag. This is caused because of the tight coupling between the linkerd version. This prevents grafana image tag from being upgraded, when grafana.image.version is present in `linkerd-config-addons`
### What problem are you trying to solve?

Ideally there should not be a tight dependency between add-on upgrades and control plane upgrades. 

What do you want to happen? Add any considered drawbacks.

Support image overwriting without running into the upgrade problem

### Any alternatives you've considered?

Is there another way to solve this problem that isn't as good a solution?

### How would users interact with this feature?

If you can, explain how users will be able to use this. Maybe some sample CLI
output?

Pothulapati:
So, I was thinking about this today.

Currently with Add-Ons, we store the default configuration in `values.yaml` and that is also stored directly in `linerd-config-addons` as a configmap. This was done because to have a central place where add-on config is present for things like check to use, and also users to update. 

The problem is now when we perform upgrades, we can't really change those values as the updated default values are overridden by the current values in `linkerd-config-addons` and these could be the older default values

## Possible Solution

Without changing much, this can be solved by making values like `image`, etc which need updates from the project **should not be present in `<add-on>:` in `values.yaml` by default** as they are used directly in cm instead they should be present in the template or another field in value. This prevents the default value from going into the cm, thus making it over-writable. This also allows users to specify `add-on.image.version` which is considered as the one to use when explicitly set.

alpeb:
Sounds like a proper solution to me :+1:

alpeb:
Would be very useful to exercise this kind of overriding in an upgrade integration test (easier if done in the existing one) :wink:

Pothulapati:
We added support for this in #4682

PR #4682, opened 2020-06-29
Use add-on level values.yaml for chart generation

Pothulapati:
Part of #4572 

We have been planning to move defaults into add-on specific `values.yaml`.

This change adds add-on level `values.yaml` to be part of the helm template to be used for
rendering.

As there are values specified in `add-on` level `values.yaml`, there are no changes needed for golden test files. 

This PR will have follow up PR's which move defaults into `values.yaml` for `grafana` and `tracing` charts.


Pothulapati:
Yep @alpeb, Was planning to separate this change. So, that the PR can focus only moving prometheus into a sub-chart.

Issue #4679, opened 2020-06-26
Unable to have ignoreOutboundPorts in helm chart propagate to injected pods


lundbird:
### What is the issue?
Using a helm3 install of linkerd with the `values.global.proxyInit.ignoreOutboundPorts` set will not have the values propagated to injected pods.

### How can it be reproduced?

Installing linkerd with several ignoreOutboundPorts set.
```
helm install linkerd2 \                                                                                   
  --set-file global.identityTrustAnchorsPEM=ca.crt \
  --set-file identity.issuer.tls.crtPEM=issuer.crt \
  --set-file identity.issuer.tls.keyPEM=issuer.key \
  --set identity.issuer.crtExpiry=$exp \
--set-string 'global.proxyInit.ignoreOutboundPorts=442\,443\,444\,445' \
linkerd/linkerd2
```
from any of the linkerd control plane pods:
```   
    - --outbound-ports-to-ignore
    - 443,442,443,444,445
```
from linkerd-config configmap:
```
      "ignoreOutboundPorts":[{"port":442},{"port":443},{"port":444},{"port":445}],
```
from any injected pods which are unable to start
```
    - --outbound-ports-to-ignore
    - ',,,'
```

Additionally an installation with one outbound port set:
```
helm install linkerd2 \                                                                                   
  --set-file global.identityTrustAnchorsPEM=ca.crt \
  --set-file identity.issuer.tls.crtPEM=issuer.crt \
  --set-file identity.issuer.tls.keyPEM=issuer.key \
  --set identity.issuer.crtExpiry=$exp \
--set-string 'global.proxyInit.ignoreOutboundPorts=443
linkerd/linkerd2
```
from linkerd-config configmap:
```
      "ignoreOutboundPorts":[{"port":443}]
```

Results in no `--outbound-ports-to-ignore` flag in injected pods.

#### `linkerd check` output
All clean

### Environment
Reproduced by setting up a new cluster in Kind, installing linkerd with the helm command, injecting namespace `default`, creating an nginx deployment in the default namespace, then verifying proxyInit args.

- Kubernetes Version: 1.14.10
- Cluster Environment: (GKE, AKS, kops, ...) Kind
- Host OS: MacOS
- Linkerd version: 2.8.1-stable

### Possible solution
[...]

PR #4688, opened 2020-06-30
Fix `splitStringListToPorts` port range object rendering

zaharidichev:
The `splitStringListToPorts` helm function is currently incorrectly formating a list of ports as an array of Port objects that look ike `{"port" : 555}`.  The config map protobuf representation however expects that the `ignoreOutboundPorts` and `ignoreInboundPorts` fields are are list of `PortRange` objects (`{"portRange" : 555}`). 

This was causing the injector to return an empty string when trying to parse a `PortRange` object resulting in the ports not getting set correctly when injecting workloads. Note that this is happening only with helm installations as this is when we are actually using a helm template for outputting the config map. 

To fix that the `splitStringListToPorts` helm function is changed to format the objects as the json representation of PortRange and is renamed to `splitStringListToPortRanges`

Fix: #4679

### DIFF
--- a/charts/linkerd2/templates/_config.tpl
+++ b/charts/linkerd2/templates/_config.tpl
@@ -30,10 +30,10 @@
     "port": {{.Values.global.proxy.ports.control}}
   },
   "ignoreInboundPorts":[
-    {{- include "partials.splitStringListToPorts" .Values.global.proxyInit.ignoreInboundPorts -}}
+    {{- include "partials.splitStringListToPortRanges" .Values.global.proxyInit.ignoreInboundPorts -}}
   ],
   "ignoreOutboundPorts":[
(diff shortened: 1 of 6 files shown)


zaharidichev:
@alpeb how do you feel about using the helm template for creating the config map even when installing through the CLI?

alpeb:
> @alpeb how do you feel about using the helm template for creating the config map even when installing through the CLI?

Yes!
So I see that in `config.yaml` we're taking separate paths depending on the existence of `.Values.configs.global` for example, which seems is only set for the CLI. So we should stop setting that. And instead, make sure all the values referred to in `_config.tpl` are actually set by the CLI...

adleong:
@zaharidichev ping :)

PR #4693, opened 2020-07-01
Use cniPluginVersion with Helm for linkerd2-cni

### LINKED ISSUE #4691: linkerd-cni helm deployment fails without a values.yaml file
### What is the issue?

linkerd-cni helm chart is incomplete, and won't deploy without a values.yaml file

The annotation `linkerd.io/created-by:` does not get set

### How can it be reproduced?

helm install linkerd2-cni linkerd2/linkerd2-cni --debug --dry-run

### Logs, error output, etc

Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [unknown object type "nil" in DaemonSet.metadata.annotations.linkerd.io/created-by, unknown object type "nil" in DaemonSet.spec.template.metadata.annotations.linkerd.io/created-by]
helm.go:84: [debug] error validating "": error validating data: [unknown object type "nil" in DaemonSet.metadata.annotations.linkerd.io/created-by, unknown object type "nil" in DaemonSet.spec.template.metadata.annotations.linkerd.io/created-by]

#### `linkerd check` output

N/A

### Environment

- Linkerd version: testing with 2.8.1, using helm 3

### Possible solution

provide a values.yaml file

```
linkerd2-cni:
  cliVersion: linkerd/helm stable-2.8.1
```

Pothulapati:
Fixes #4691 

As `cliVersion` field is not present when the installation is done through Helm, default to `cniVersionValue` when it is not present.


cli/cmd/testdata/install_cni_helm_default_output.golden
```diff
@@ -0,0 +1,204 @@
...
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: linkerd-cni
+  namespace: linkerd-cni
+  labels:
+    k8s-app: linkerd-cni
+    linkerd.io/cni-resource: "true"
+  annotations:
+    linkerd.io/created-by: linkerd/cli dev-undefined
```
Pothulapati, line 144: Hmmm, From what I understood, `dev-undefined` is our fallback when there is no version tag is specified. and whats where we fallback when `cliVersion` is empty but should we be okay with having tests using this? especially the helm ones as there we can't fall back to this?

PR #4695, opened 2020-07-01
[POC] service topologies: add nodeName to context token

Matei207:
### PoC for [linkerd/linkerd2#4498](https://github.com/linkerd/linkerd2/issues/4498)
**N.B**:  *most of the information in the ticket/issue/card related to implementation details is not as relevant following the [first PR](https://github.com/linkerd/linkerd2/pull/4570)*

### What
---

* Proof of concept to show how to add a pod's nodeName as part of the `LINKERD2_PROXY_DESTINATION_CONTEXT` environmental variable. Following a conversation on this topic (*see first PR*), I tried to structure the context token in json format.
* `LINKERD2_PROXY_DESTINATION_CONTEXT` receives a json string with the field & field values (obtained through downward API, hardcoded, etc.). The proxy should read the json string in and pass it over the wire through gRPC.
* On the server side (i.e destination service), we keep track of the structure of the token and unmarshal the string into the applicable fields.

As part of this PR, I changed the server to log out the fields of the structure when a call of type `GetDestination` is made. I also modified the client script to test it out at first, all log clippings annexed at the end of the PR are from the destination service running in minikube.

Think this is a good way to piggyback contextual information on the already present `DESTINATION_CONTEXT_TOKEN`. I'm not sure what the performance implications are necessarily, but this would allow the token to be extended over time. For the purpose of simplicity, I don't think the `proxy-api` should be extended. I have also looked into it, and splitting it into additional fields on the `proxy-api` side shouldn't have any performance improvements (and can have the complete opposite effect depending on the chosen way of structuring the token).

Finally, I think this approach should preserve the opacity and separation of concerns in the proxy. Since the destination service is aware of all of the contextual bits and bobs associated with a service/pod, having a structure of the token inside makes sense to me. On the other hand, by simply treating the token as a json string from proxy all the way through the wire, it will not have any concept of what's inside the token, other than its type. Hope this makes sense.

Let me know what you think of the approach and if it fits with what you had in mind
@olix0r @adleong @grampelberg 

---

```bash
# Pod exec to check if env variables have been properly set
$ k exec linkerd-destination-88f8bb878-jrqbk -n linkerd -it -c linkerd-proxy -- bin/bash -c "env | grep nodeName"
_pod_nodeName=minikube
LINKERD2_PROXY_DESTINATION_CONTEXT={"ns":"linkerd","nodeName":"minikube"}
```

```bash
# Destination Service logs, no other apps were running in minikube at the time
# so the client in this case is another linkerd component
$ k logs <dest-service>
time="2020-07-01T17:22:36Z" level=debug msg=Context token: [ogToken={\"ns\":\"linkerd\",\"nodeName\":\"minikube\"}\n] [tokenStruct={linkerd minikube}] [ns=linkerd], [nodeName=minikube]

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -41,13 +41,13 @@ env:
   valueFrom:
     fieldRef:
       fieldPath: metadata.namespace
-- name: LINKERD2_PROXY_DESTINATION_CONTEXT
-  value: |
-    {"ns":"$(_pod_ns)","nodeName":"$(_pod_node)"}
-- name: _pod_node
+- name: _pod_nodeName
   valueFrom:
     fieldRef:
-      fieldPath: spec.nodeName
+      fieldPath: spec.nodeName      
+- name: LINKERD2_PROXY_DESTINATION_CONTEXT
+  value: |
+    {"ns":"$(_pod_ns)","nodeName":"$(_pod_nodeName)"}
 {{ if eq .Values.global.proxy.component "linkerd-prometheus" -}}
 - name: LINKERD2_PROXY_OUTBOUND_ROUTER_CAPACITY
   value: "10000"
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -42,7 +42,12 @@ env:
     fieldRef:
       fieldPath: metadata.namespace
 - name: LINKERD2_PROXY_DESTINATION_CONTEXT
-  value: ns:$(_pod_ns)
+  value: |
+    {"ns":"$(_pod_ns)","nodeName":"$(_pod_node)"}
+- name: _pod_node
+  valueFrom:
+    fieldRef:
+      fieldPath: spec.nodeName
 {{ if eq .Values.global.proxy.component "linkerd-prometheus" -}}
 - name: LINKERD2_PROXY_OUTBOUND_ROUTER_CAPACITY
   value: "10000"
--- a/controller/api/destination/server.go
+++ b/controller/api/destination/server.go
@@ -6,6 +6,8 @@ import (
 	"strconv"
 	"strings"
 
+	"encoding/json"
+
 	pb "github.com/linkerd/linkerd2-proxy-api/go/destination"
 	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
 	"github.com/linkerd/linkerd2/controller/k8s"
@@ -100,6 +102,12 @@ func (s *server) Get(dest *pb.GetDestination, stream pb.Destination_GetServer) e
 		log,
 	)
 
+	var token contextToken
+	b := dest.GetContextToken()
+	if err := json.Unmarshal([]byte(b), &token); err != nil {
+		log.Debugf("Error: %v", err)
+	}
+	log.Debugf("Context token: [ogToken=%s] [tokenStruct=%s] [ns=%s], [nodeName=%s]", dest.GetContextToken(), token, token.Ns, token.NodeName)
 	// The host must be fully-qualified or be an IP address.
 	host, port, err := getHostAndPort(dest.GetPath())
 	if err != nil {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by grampelberg** [comment]
@adleong @olix0r can you check and see how this looks? Let's chat if there needs to be another approach!

File: charts/partials/templates/_proxy.tpl [hunk 1]
```diff
@@ -41,8 +41,13 @@ env:
...
     fieldRef:
       fieldPath: metadata.namespace
+- name: _pod_nodeName
+  valueFrom:
+    fieldRef:
+      fieldPath: spec.nodeName      
 - name: LINKERD2_PROXY_DESTINATION_CONTEXT
-  value: ns:$(_pod_ns)
+  value: |
+    {"ns":"$(_pod_ns)","nodeName":"$(_pod_nodeName)"}
```
adleong, line 50: +1 this format looks good to me

controller/api/destination/server.go
```diff
@@ -100,6 +102,12 @@ func (s *server) Get(dest *pb.GetDestination, stream pb.Destination_GetServer) e
 		log,
 	)
 
+	var token contextToken
+	b := dest.GetContextToken()
+	if err := json.Unmarshal([]byte(b), &token); err != nil {
+		log.Debugf("Error: %v", err)
+	}
+	log.Debugf("Context token: [ogToken=%s] [tokenStruct=%s] [ns=%s], [nodeName=%s]", dest.GetContextToken(), token, token.Ns, token.NodeName)
```
adleong, line 110: logging the token all of these different ways is probably overkill.  I think we can either log just the raw token, or the parsed struct.

controller/api/destination/server.go
```diff
@@ -100,6 +102,12 @@ func (s *server) Get(dest *pb.GetDestination, stream pb.Destination_GetServer) e
 		log,
 	)
 
+	var token contextToken
+	b := dest.GetContextToken()
+	if err := json.Unmarshal([]byte(b), &token); err != nil {
```
adleong, line 107: I'm thinking about the upgrade story and I think we need to be able to accept tokens in either format: `ns:<namespace>` or json.  During an upgrade, the destination controller will get upgraded first but the proxies will still be running the previous version and will still be sending context tokens in the old format.

controller/api/destination/server.go
```diff
@@ -100,6 +102,12 @@ func (s *server) Get(dest *pb.GetDestination, stream pb.Destination_GetServer) e
...
+		log.Debugf("Error: %v", err)
```
adleong, line 108: This is a serious error so we should log it at error level.  We should also either return the error (causing the API request to fail) or explicitly set the contextToken to a default empty state.

controller/api/destination/server.go
```diff
@@ -237,6 +245,11 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 /// util ///
 ////////////
 
+type contextToken struct {
+	Ns       string `json:"ns,omitempty"`
+	NodeName string `json:"nodeName,omitempty"`
+}
+
 func nsFromToken(token string) string {
```
adleong, line 253: I think we need to remove the `nsFromToken` function and instead just use the `token.Ns` field.

controller/api/destination/server.go
Matei207: Apologies for this! I left this in so that the logs I added in the PR description make a bit more sense. It was more to be let you (the reviewers) visualise the changes and manual tests a bit better. Should have been a bit more explicit, sorry! Will make sure we log a bit better in the actual implementation.

controller/api/destination/server.go
Matei207: Amazing suggestion, I was a bit short sighted and did not think about backwards compatibility that much. Thank you!

controller/api/destination/server.go
Matei207: Hm, I have looked at this again. Wouldn't it make sense to keep the function in if the token is not in json format? The (rough) idea I have in mind is:

```go
func parseToken(token string) contextToken {
  if !isValidJSON(token) {
    ns := nsFromToken(token)
    return contextToken{ns}
  }
  unmarshal json
  ...
  return ctxToken
}
```

adleong: This is purely a stylistic choice at this point, but personally I'd inline the `nsFromToken` logic into `parseToken`.  The reason being that the name `nsFromToken` suggests that that function is the authoritative way to get a namespace from a token.

Matei207:
Closing this now in favour of an actual PR. Thank you for the feedback!

PR #4696, opened 2020-07-01
service topologies: add Kubernetes/API EndpointSlice support

Matei207:
Based on the [EndpointSlice PR](https://github.com/linkerd/linkerd2/pull/4663), this is just the k8s/api support for endpointslices to shorten the first PR.

* Adds CRD
* Adds functions that check whether the cluster has EndpointSlice access
* Adds discovery & endpointslice informers to api.

### DIFF
--- a/charts/linkerd2/templates/destination-rbac.yaml
+++ b/charts/linkerd2/templates/destination-rbac.yaml
@@ -26,6 +26,9 @@ rules:
 - apiGroups: ["split.smi-spec.io"]
   resources: ["trafficsplits"]
   verbs: ["list", "get", "watch"]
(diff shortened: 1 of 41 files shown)


Matei207:
**Update**: based on the discussion with @adleong in [PR#4663](https://github.com/linkerd/linkerd2/pull/4663#discussion_r449193678) I've changed the return type as well as commented out the two functions responsible for checking whether EndpointSlices can be used.

The main assumption we are making is that EndpointSlices should be opt-in for now, as a result, until a mechanism to enable them on a user-by-user basis is put in, the idea is to only use Endpoints (but have the support for the slices in).

PR #4709, opened 2020-07-03
Fix deep integration test

kleimkuhler:
This fixes the deep integration test which currently only calls `run_test` for
`edges` integration test.

This occurs because `run_test "${tests[@]}"` will pass an entire array of
filenames when `run_test` only expects *one* filename.

The solution is to loop through `tests` and call `run_test` for each file.


bin/_test-helpers.sh
```diff
@@ -339,10 +339,12 @@ run_uninstall_test() {
 }
 
 run_deep_test() {
-  run_test "$test_directory/install_test.go" --multicluster
+  run_test "$test_directory/install_test.go"
```
alpeb, line 342: Tests are run with `--multicluster`  to test the installation (and uninstallation) of the multicluster components in `install_test.go`

kleimkuhler: Ah yep didn't mean to remove this!

PR #4717, opened 2020-07-03
Gateway Metrics in Dashboard

### LINKED ISSUE #4601: Support Gateways in the dashboard
## Feature Request

Currently, There is no way to get gateways information in the Linkerd2 dashboard. As this information is exposed through the `public-api`, It should not be hard to have a `Gateways` tab where we can show the same information from the CLI using a `MetricsTable`

Remote gateways information in the dashboard

Add a `Gateways` Tab into the Linkerd2 dashboard?

There is the CLI already for this information, but as this info is fetched from public-api, It should be a pretty straight forward PR.

output?

tharun208:
Added Gateway Component in Linkerd Dashboard
Updated MetricsTable to support gateway metrics

Fixes #4601


zaharidichev:
@tharun208 This is great. I wonder though whether a lot of this will change significantly once the changed @adleong has planned land ?

controller/api/util/api_utils.go
```diff
@@ -116,6 +117,14 @@ type TapRequestParams struct {
 	LabelSelector string
 }
 
+// GatewayRequestParams conttains parameters that are used to build a
+// GatewayRequest
+type GatewayRequestParams struct {
+	RemoteClusterName string
```
adleong, line 123: is this unused?

controller/api/util/api_utils.go
```diff
@@ -268,6 +277,28 @@ func BuildEdgesRequest(p EdgesRequestParams) (*pb.EdgesRequest, error) {
...
+	window := defaultMetricTimeWindow
+	if p.TimeWindow != "" {
+		_, err := time.ParseDuration(p.TimeWindow)
+		if err != nil {
+			return nil, err
+		}
+		window = p.TimeWindow
+	}
+	multiClusterNameSpace := p.GatewayNamespace
+	if multiClusterNameSpace == "" {
```
adleong, line 292: if unspecified, I think we want to leave this blank which should return gateways in all namespaces

web/srv/api_handlers.go
```diff
@@ -438,3 +438,22 @@ func (h *handler) handleAPIResourceDefinition(w http.ResponseWriter, req *http.R
...
 	w.Write(resourceDefinition)
 }
+
+func (h *handler) handleAPIGateways(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	requestParams := util.GatewayRequestParams{
+		TimeWindow:        req.FormValue("window"),
+		GatewayNamespace:  req.FormValue("gatewayNamespace"),
+		RemoteClusterName: req.FormValue("remoteClusterName"),
+	}
+	gatewayRequest, err := util.BuildGatewayRequest(requestParams)
```
adleong, line 448: I don't think we really need a brand new struct and helper method for building the api request.  The only properties we need are the time window and (optionally) the gateway namespace so we should be able to construct the request inline here easily enough.

controller/api/util/api_utils.go
tharun208: it is used

web/srv/api_handlers.go
tharun208: to make it consistent with the code with other handler's. I added this, but one more reason for adding this is that in future if we try to get a particular gateway from a namespace

adleong: I'm a big fan consistency but I think in this case it's more valuable to simplify and to reduce boilerplate.  If we need more functionality in the future we can always add it at that time.

tharun208: sure @adleong will do the changes

controller/api/util/api_utils.go
tharun208: it is used in `handleAPIGateways` handler.

adleong: can you show me exactly where in the `handleAPIGateways` handler?  I see it being set on line 446 but as far as I can tell, the `requestParams` object is only ever passed to BuildGatewayRequest which does not use the RemoteClusterName field.

tharun208: fixed it

web/srv/api_handlers.go
tharun208: > sure @adleong will do the changes, but I am wondering the `timeWindow` validation should be done at the helper method. So, I don't want to pollute the `handler` by putting all the logic in a single place.
> 
> And if I add the `timeWindow` validation at the handler level, then I need to introduce one more constant `defaultTimeWindow`.And, I don't want this @adleong

adleong: I don't think the validation is complex enough to require a separate function.  This code is more readable to me to have the time window validation and defaulting in-line here in the handler.

tharun208: @adleong, can I remove both the `BuildGatewayRequest` for now ?

adleong: yes, that's what I'm suggesting.

tharun208: @adleong removed it

tharun208:
@adleong , are there any other changes need to be done?

web/srv/api_handlers.go
```diff
@@ -438,3 +438,31 @@ func (h *handler) handleAPIResourceDefinition(w http.ResponseWriter, req *http.R
 	w.Header().Set("Content-Type", "text/yaml")
 	w.Write(resourceDefinition)
 }
+
+func (h *handler) handleAPIGateways(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	requestParams := util.GatewayRequestParams{
```
adleong, line 443: Thanks for updating and removing the helper function!  However, I was getting at is that we can remove this params structure entirely and just construct the request directly.  Sorry if that wasn't clear.  What I had in mind was something like this:

```
	window := req.FormValue("window")
        if window == "" {
                window = "1m"
        }
	_, err := time.ParseDuration(window)
	if err != nil {
 		renderJSONError(w, err, http.StatusInternalServerError)
		return
	}
	gatewayRequest := &pb.GatewaysRequest{
		TimeWindow:        window,
		GatewayNamespace:  req.FormValue("gatewayNamespace"),
		RemoteClusterName: rreq.FormValue("remoteClusterName"),
	}
```

PR #4718, opened 2020-07-04
Enable overriding timeout for DNS canonization through annotation.

### LINKED ISSUE #4660: Support annotation to override default timeouts for Linkerd-proxy DNS canonicalization
Currently, Linkerd proxy provides `LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT` environment variable that allows the default 100ms timeout to be overridden.  https://github.com/linkerd/linkerd2-proxy/pull/175
However there is no [proxy configuration](https://linkerd.io/2/reference/proxy-configuration/) or annotation to override this environment value during helm or customized installation or even using release pipeline

we should probably provide linkerd config annotation that can override environment variables like `LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT`

`config.linkerd.io/proxy-dns-canonicalize-timeout : "200"`

Alternatively we should consider providing CLI flag too   `--proxy-dns-canonicalize-timeout`

For example, we can add annotation to our proxy config, 
```
             annotations:
                   config.linkerd.io/proxy-dns-canonicalize-timeout : "200"
```
or can also be done manually using linkerd CLI
`kubectl get deployment ... | linkerd inject  --proxy-dns-canonicalize-timeout : "200" | kubectl apply -`

ArthurSens:
Currently, Linkerd proxy provides LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT environment variable that allows the default 100ms timeout to be overridden.

However, it is not possible to override it with annotations. We should provide linkerd config annotation that can override environment variables like LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT

- setting the `DNSCanonicalizeTimeoutSeconds` field in values during a Helm install, which changes the default on all injected pods
- using an annotation ` config.linkerd.io/proxy-dns-canonicalize-timeout` for injected workloads to override this value

Fixes #4660 

Signed-off-by: arthursens <arthursens2005@gmail.com>


ArthurSens:
@cpretzer I'm using #4608 and #4620 as references, I can't really see if I'm doing something wrong with the code or if the test I'm doing is not valid.

I'm running the commands below, expecting that the injected yaml would contain the variable `LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT`:

```shell
minikube delete && minikube start
DOCKER_TRACE=1 bin/mkube bin/docker-build
bin/linkerd install | kubectl apply -f -
bin/linkerd inject --manual  https://run.linkerd.io/emojivoto.yml | grep -i DNS
```
I get nothing as the return of the last command

charts/linkerd2/templates/_config.tpl
```diff
@@ -62,7 +62,8 @@
     "pullPolicy":"{{.Values.debugContainer.image.pullPolicy}}"
   },
   "debugImageVersion": "{{.Values.debugContainer.image.version}}",
-  "destinationGetNetworks": "{{.Values.global.proxy.destinationGetNetworks}}"
+  "destinationGetNetworks": "{{.Values.global.proxy.destinationGetNetworks}}",
+  "DNSCanonicalizeTimeoutSeconds": "{{.Values.global.proxy.DNSCanonicalizeTimeoutSeconds}}"
```
zaharidichev, line 66: I think we would want lowercase here, WDYT ?

proto/config/config.proto
```diff
@@ -55,6 +55,8 @@ message Proxy {
   string debug_image_version = 16;
    
   string destination_get_networks = 17;
+
+  uint64 DNS_canonicalize_timeout_seconds = 18;
```
zaharidichev, line 59: Should this be lowercase ?

ArthurSens:
Hi @Pothulapati thanks for the review! I've made the changes you suggested, but I still wasn't able to get it working. In fact, I can't even run `bin/linkerd install ...` now, because it fails to create deployments with the linkerd-proxy container

```
Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found 1, error found in #10 byte of ...|,"value":100},{"name|..., bigger context ...|LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT","value":100},{"name":"LINKERD2_PROXY_CONTROL_LISTEN_ADDR","|...
```

I see that it is expecting a string, but I don't see a reason for that... it should be possible to create Environment Variable as integers, right?




@zaharidichev I'll try to add a new test as soon as I get it working :sweat:

cpretzer:
@ArthurSens the `value` for an `env` is expected to be a [string](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#envvar-v1-core), and the parser is interpreting the number as an integer, which is why you see that error.

To test this, I pulled your branch and modified values.yaml to append an 's', like this:
```
- name: LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT
  value: {{.Values.global.proxy.DNSCanonicalizeTimeoutSeconds}}ms
```

Do you think appending the duration specifier is the right approach? The alternative is to always specify the value with the duration specifier in the values.yaml file and as an annotation, i.e. `config.linkerd.io/proxy-dns-canonicalize-timeout: 100ms`

Another thing to consider is that the proxy logic [parses](https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/src/env.rs#L294) the specified value as a duration in [milliseconds](https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/src/env.rs#L203), so setting the [default value](https://github.com/linkerd/linkerd2/pull/4718/files#diff-9d329f800b46936a44723914cdd2cf48R57) in values.yaml would use the specifier `ms` and instead of `s`.

What do you think of setting this value in milliseconds instead of seconds?

ArthurSens:
> @ArthurSens the `value` for an `env` is expected to be a [string](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#envvar-v1-core), and the parser is interpreting the number as an integer, which is why you see that error.
> 
> To test this, I pulled your branch and modified values.yaml to append an 's', like this:
> 
> ```
> - name: LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT
>   value: {{.Values.global.proxy.DNSCanonicalizeTimeoutSeconds}}ms
> ```
> 
> Do you think appending the duration specifier is the right approach? The alternative is to always specify the value with the duration specifier in the values.yaml file and as an annotation, i.e. `config.linkerd.io/proxy-dns-canonicalize-timeout: 100ms`
> 
> Another thing to consider is that the proxy logic [parses](https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/src/env.rs#L294) the specified value as a duration in [milliseconds](https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/src/env.rs#L203), so setting the [default value](https://github.com/linkerd/linkerd2/pull/4718/files#diff-9d329f800b46936a44723914cdd2cf48R57) in values.yaml would use the specifier `ms` and instead of `s`.
> 
> What do you think of setting this value in milliseconds instead of seconds?

Oh nice, that makes things a bit clearer. I was using [waitBeforeExitSeconds](https://github.com/linkerd/linkerd2/pull/4718/files#diff-9d329f800b46936a44723914cdd2cf48R54) as a reference to create the new duration attribute at values.yaml. 

Since the time unit is declared when setting this annotation, I think `dnsCanonicalizeTimeout` should be enough as its name, without Seconds or Milliseconds in the end. WDYT?

cpretzer:
> Since the time unit is declared when setting this annotation, I think dnsCanonicalizeTimeout should be enough as its name, without Seconds or Milliseconds in the end. WDYT?

Yeah, that sounds reasonable. Let's be sure that the documentation is clear that the value is milliseconds.

cpretzer:
Closing this per comment from @ArthurSens in #4660

Issue #4375, opened 2020-05-12
Prometheus add-on parent issue


Pothulapati:
This is a parent issue of all the changes that will be done to make prometheus as an add-on. 

- [x] Move prometheus into a sub-chart and as a add-on
- [x] Relax or remove existing component dependencies on prometheus and for checks.
- [x] Add add-on specific checks, that see if the prometheus components are running correctly.
- [x] Add `global.prometheusUrl` for BYOP case., along with validation in helm
- [x] Add [check tooling](https://github.com/Pothulapati/rfc/blob/prom/design/0003-byop.md#linkerd-cli-tooling) for the BYOP case.
- [x] Bring your own prometheus linkerd website task
- [ ] Test Plan

kleimkuhler:
@Pothulapati Correct me if I'm wrong but this can be closed now? The final thing would be the test plan (tracked in #5041) and that currently passes. That issue is only remaining open until I have retested with the RC.

Pothulapati:
@kleimkuhler Yep, you are right. Only the testing with RC is left.

PR #4724, opened 2020-07-07
Remove/Relax prometheus related checks

Pothulapati:
Implements second part of https://github.com/linkerd/rfc/pull/16, https://github.com/linkerd/linkerd2/issues/4375

Now that prometheus is an add-on (i.e #4362 ), There can be cases where prometheus is
disabled at which the check should show a warning but not fail. This
decouples the tight depedency.

This changes the following checks:

- Removes serviceAccount and pod checks in the CLI.
- Relaxes `linkerd-api` checks to only check for prometheus access when
the URL is not empty. This should work seamlessly with external
prometheus as that URL will be passed and it performs the same
check.
- Updates `queryProm` function to perform a check, to not crash but
report a failure to the dashboard and cli


alpeb:
How feasible would it be to move those checks to its own section under `healthcheck_addons.go` so prometheus gets own checks section in the `linkerd check output`?

Pothulapati:
@alpeb Yep, That's the plan :)

Pothulapati:
@alpeb Updated the controller instance to fail gracefully when there is no prometheus, works as expected both with CLI and as well as the dashboard.

PR #4727, opened 2020-07-07
Add support for profile lookups by IP address

### LINKED ISSUE #3916: destination: Support profile lookups by IP address
Now that `Destination.Get` supports lookups by IP address, we should entertain supporting IP lookups for `GetProfile` as well.

I imagine this would work by mapping the IP to a kubernetes service and lookup as we already do today; and that the default profile would be returned for all IPs that cannot be discovered.

kleimkuhler:
## Motivation

Closes #3916

This adds the ability to get profiles for services by IP address.

### Change in behavior

When the destination server receives a `GetProfile` request with an IP address,
it now tries to map that IP address to a service.

If the IP address maps to an existing service, then the destination server
returns the profile stream subscribes for updates to the _service_--this is the
existing behavior. If the IP changes to a new service, the stream will still
send updates for the first service the IP address corresponded to since that is
what it is subscribed to.

If the IP address does not map to an existing service, then the destination
server returns the profile stream but does not subscribe for updates. The stream
will receive one update, the default profile.

### Solution

This change uses the `IPWatcher` within the destination server to check for what
services an IP address correspond to. By adding a new method `GetSvc` to
`IPWatcher`, the server now calls this method if `GetProfile` receives a request
with an IP address.

## Testing

Install linkerd on a cluster and get the cluster IP of any service:

```bash
❯ kubectl get -n linkerd svc/linkerd-tap -o wide
NAME          TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)            AGE   SELECTOR
linkerd-tap   ClusterIP   10.104.57.90   <none>        8088/TCP,443/TCP   16h   linkerd.io/control-plane-component=tap
```

Run the destination server:

```bash
❯ go run controller/cmd/main.go destination -kubeconfig ~/.kube/config
```

Get the profile for the tap service by IP address:

```bash
❯ go run controller/script/destination-client/main.go -method getProfile -path 10.104.57.90:8088
INFO[0000] retry_budget:{retry_ratio:0.2  min_retries_per_second:10  ttl:{seconds:10}} 
INFO[0000]
```

Get the profile for an IP address that does not correspond to a service:

```bash
❯ go run controller/script/destination-client/main.go -method getProfile -path 10.256.0.1:8088
INFO[0000] retry_budget:{retry_ratio:0.2  min_retries_per_second:10  ttl:{seconds:10}} 
INFO[0000]
```

You can add and remove settings for the service profile for tap and get updates.


controller/api/destination/server.go
```diff
@@ -164,16 +164,23 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
+	var path string
 
-	service, _, err := parseK8sServiceName(host, s.clusterDomain)
-	if err != nil {
-		log.Debugf("Invalid service %s", dest.GetPath())
-		return status.Errorf(codes.InvalidArgument, "invalid service: %s", err)
+	if ip := net.ParseIP(host); ip != nil {
+		svc, exists := s.ips.GetSvc(ip.String())
+		if exists {
+			path = fmt.Sprintf("%s.%s.svc.%s", svc.Name, svc.Namespace, s.clusterDomain)
```
adleong, line 173: we also need to set the `service` variable here, right?

controller/api/destination/server.go
```diff
@@ -164,16 +164,23 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
+		} else {
+			return status.Errorf(codes.InvalidArgument, "IP address %s is not a service", ip.String())
```
adleong, line 175: InvalidArgument feels a bit weird, but I'm not sure what a better error would be.

It's worth noting that this works differently than looking up a service that doesn't exist by authority.  In that case we immediately send an update with the default service profile and then we'll keep the stream open and if a service profile with that name ever gets created we'll send an update.

In this case, however, if we lookup an IP that doesn't correspond to a service, we immediately fail the stream instead of sending back the default service profile at all.

@olix0r: is that how you would expect this to work from the proxy's perspective?  Would it be preferable to send the default service profile and then leave the stream open, never sending any further updates on it?

controller/api/destination/watcher/ip_watcher.go
```diff
@@ -104,6 +104,15 @@ func (iw *IPWatcher) Unsubscribe(clusterIP string, port Port, listener EndpointU
 	ss.unsubscribe(port, listener)
 }
 
+// GetSvc returns the service that corresponds to an IP address if one exists.
+func (iw *IPWatcher) GetSvc(clusterIP string) (*ServiceID, bool) {
+	ss, exists := iw.getServiceSubscriptions(clusterIP)
```
adleong, line 109: I think this will work because we create an serviceSubscription whenever there is a service added event.  However, it still feels a bit brittle and indirect.  A more direct approach would be to do something like `iw.k8sAPI.Svc().Informer().GetIndexer().ByIndex(podIPIndex, clusterIP)` to get the service directly.

kleimkuhler: I see what you mean. I originally did it this way, but changed it to how it currently is since it ends up fairly similar to `getOrNewServiceSubscriptions`.

I think it's pretty safe to assume that the informers event handler will be triggered in a timely manner--and call `getOrNewServiceSubscriptions`, but see why this may be a more direct approach.

controller/api/destination/server.go
```diff
@@ -164,16 +164,50 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
+		} else {
+			primary, secondary := newFallbackProfileListener(translator)
+
+			if dest.GetContextToken() != "" {
+				err = s.profiles.Subscribe(nil, primary)
+				if err != nil {
+					log.Warnf("Failed to subscribe to profile %s: %s", path, err)
+					return err
+				}
+				defer s.profiles.Unsubscribe(nil, primary)
```
zaharidichev, line 187: I think the error that Unsubsribe can sometimes return will be muted when calling it via defer. Maybe better to leave the function without a return value and just log on error.

controller/api/destination/server.go
kleimkuhler: @adleong I change it so that when an IP does not map to a service, a profile stream is still returned that receives one update, the default profile. This stream will not receive any more updates since we do not have a service to subscribe to. This does match more with what Oliver mentioned in the original issue, so I think that is what he would lean towards anyways.

controller/api/destination/watcher/ip_watcher.go
kleimkuhler: This method has been changed since your review. On top checking the actual indexer, it now handles an IP that does _or_ does not map to a service and errors. The existence of a service that it maps to is used by the calling function to determine if it should subscribe to a service or not.

controller/api/destination/server.go
kleimkuhler: I also updated the PR description with this new behavior and added an example of testing it.

controller/api/destination/server.go
kleimkuhler: Yes you are correct here that errors that occur in `Unsubscribe` will fail silently.

I tested this by inserting `return fmt.Errorf("this is an error")` in `Unsubscribe`, and then subscribed to and quit a profile stream.

I expected to see this error in the server, but nothing occurred.

This occurs in the other places that we unsubscribe in this file, so I lean towards making this change separately.

controller/api/destination/watcher/ip_watcher.go
```diff
@@ -104,6 +106,27 @@ func (iw *IPWatcher) Unsubscribe(clusterIP string, port Port, listener EndpointU
 	ss.unsubscribe(port, listener)
 }
 
+// GetSvc returns the service that corresponds to an IP address if one exists.
+func (iw *IPWatcher) GetSvc(clusterIP string) (*ServiceID, bool, error) {
```
adleong, line 110: I think this would be fine to just return a `(*ServiceID, error)` because `exists == false` exactly when `serviceID == nil`.

controller/api/destination/watcher/profile_watcher.go
```diff
@@ -127,53 +127,76 @@ func (pw *ProfileWatcher) deleteProfile(obj interface{}) {
...
+	publisher, ok := pw.getProfilePublisher(&id)
 	if ok {
 		publisher.update(nil)
 	}
 }
 
-func (pw *ProfileWatcher) getOrNewProfilePublisher(id ProfileID, profile *sp.ServiceProfile) *profilePublisher {
-	pw.Lock()
-	defer pw.Unlock()
+func (pw *ProfileWatcher) getOrNewProfilePublisher(id *ProfileID, profile *sp.ServiceProfile) *profilePublisher {
```
adleong, line 136: it looks like all the changes in this file are due to switching ProfileID to a pointer and properly handling nil.  Why did this need to change?  As far as I can tell we shouldn't ever be passing `nil` into subscribe.  (all of the instances in server.go pass in a reference to a specific struct).

kleimkuhler: These changes can be removed. I simplified this earlier today, and with that simplification we don't need to pass `nil` profiles anymore. Thanks for catching that!

controller/api/destination/server.go
```diff
@@ -164,16 +164,35 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
-		return status.Errorf(codes.InvalidArgument, "invalid service: %s", err)
+			return nil
+		}
+	} else {
+		service, _, err = parseK8sServiceName(host, s.clusterDomain)
+		if err != nil {
+			log.Debugf("Invalid service %s", dest.GetPath())
+			return status.Errorf(codes.InvalidArgument, "invalid service: %s", err)
+		}
+		path = dest.GetPath()
```
olix0r, line 195: This could probably use a little more commentary/narrative, as the rest of this function has.

controller/api/destination/server.go
```diff
@@ -164,16 +164,35 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
+			case <-s.shutdown:
+			case <-stream.Context().Done():
+				log.Debugf("GetProfile(%+v) cancelled", dest)
+			}
 
-	service, _, err := parseK8sServiceName(host, s.clusterDomain)
-	if err != nil {
-		log.Debugf("Invalid service %s", dest.GetPath())
-		return status.Errorf(codes.InvalidArgument, "invalid service: %s", err)
+			return nil
```
olix0r, line 187: So, Thinking a bit more. I do think we're going to have to watch for updates. Just because a service isn't in our index doesn't mean it's not about to show up there. That doesn't have to be a hard blocker -- we can merge this and be incremental. But I think we should support the service being known by the app before it is known by the controller...

This is also jogging some ideas for things we may want to do in a followup re: supporting returning all profiles for a pod:ip target.

kleimkuhler:
> Cool, I think we can ship this as-is with a TODO or a follow up issue to make the mapping from ip to service dynamic.

#4743 is the follow-up issue for this

PR #4728, opened 2020-07-07
Helm integration deep tests

alpeb:
This creates a new integration test target that launches the deep suite,
using a linkerd instance installed through Helm.

I've added a `global.proxyInit.ignoreInboundPorts=1234,5678` override
during install and enhanced the injection test to catch problems like
what we saw in #4679. So this PR will fail that test until #4688 merges.


test/integration/inject/inject_test.go
```diff
@@ -299,6 +302,49 @@ func TestInjectAutoPod(t *testing.T) {
...
+	falsy := false
+	zero := int64(0)
+	expectedInitContainer := v1.Container{
+		Name:  k8s.InitContainerName,
+		Image: "gcr.io/linkerd-io/proxy-init:" + version.ProxyInitVersion,
+		Args: []string{
+			"--incoming-proxy-port", "4143",
+			"--outgoing-proxy-port", "4140",
+			"--proxy-uid", "2102",
+			"--inbound-ports-to-ignore", "4190,4191,1234,5678",
```
kleimkuhler, line 315: We should only expect `4190` and `4191` right?

alpeb: I'm testing the `global.proxyInit.ignoreInboundPorts=1234,5678` override I added to `install_test.go`. I'll add a comment.

PR #4733, opened 2020-07-08
(WIP) Make api service secrets compatible to cert-manager

cypherfox:
## Problem ##
 Currently the secrets for the proxy-injector, sp-validator webhooks and tap API service are using the `Opaque` secret type and linkerd-specific field names. This makes it impossible to use cert-manager (https://github.com/jetstack/cert-manager) to provisions and rotate the secrets for these services.

## Solution ##
Convert the secrets defined in the linkerd2 helm charts and the controller use the `kubernetes.io/tls` format instead. This format is used for secrets containing the generated secrets by cert-manager.

## Validation ##
Run the suite of unit and integration tests.


cypherfox:
closed in favor of #4737 in order to clean up the git history.

PR #4737, opened 2020-07-09
make api service secrets compatible to cert manager

cypherfox:
## Problem ##

Currently the secrets for the proxy-injector, sp-validator webhooks and tap API service are using the Opaque secret type and linkerd-specific field names. This makes it impossible to use cert-manager (https://github.com/jetstack/cert-manager) to provisions and rotate the secrets for these services.

## Solution ##

Convert the secrets defined in the linkerd2 helm charts and the controller use the kubernetes.io/tls format instead. This format is used for secrets containing the generated secrets by cert-manager.

## Validation ##

Run the suite of unit and integration tests.


cypherfox:
This change currently cannot be fixed, as it lacks the needed update path.

## Background 
Currently the three relevant secrets (`proxy-injector-tls`, `sp-validator-tls`, and `tap-apiservice-tls`) are of the type `Opaque` and hold the data in non-standard field names. But changing the type of the secrets would required deleting them, as the type of an secret is immutable.

Deleting the secret at the point in time when the need for an upgrade is detected has two problems: there is a time-frame where no secret is available. If any webhook/ApiService components need to be restarted at this time, it would lead to a downtime for the webhook. Also it will make the update non-atomic: the update may fail to create new secrets after the old once have already been deleted.

## Possible Solution 
create new secrets with a new name. But this will require improved migration logic, which is not there yet.

See also Slack conversion for more info: https://linkerd.slack.com/archives/CMS5ZD60Y/p1594016108377100

alpeb:
@cypherfox The best solution IMO is the name-changing strategy. During upgrade the pods will get rolled out, so the new ones will start to use the new secrets with no downtime. We'd just add some instructions in the upgrade notes about the deletion of the old secrets after the upgrade is done. Do you have any other concerns regarding this plan?

cypherfox:
The update notes should also need to indicate that those that use the external secrets need to create new ones with the new name (and possibly the old content) BEFORE updating.

But yes, as this is hopefully a one-time event, simply changing the name and documenting the manual steps required should be fine.

ihcsim:
@cypherfox What's the status of this PR? Are you able to get the `helm upgrade` integration test to pass? Let us know if you need help.

cypherfox:
@ihcsim was away on vacation for a few days, will check when I get some time to get this cleaned up.

ihcsim:
@alpeb @zaharidichev PTAL when you get the chance. This change will allow users to set up cert-manager to manage the webhook secrets. Thanks!

cli/cmd/upgrade.go
```diff
@@ -658,6 +696,14 @@ func webhookSecretName(webhook string) string {
 	return fmt.Sprintf("%s-tls", webhook)
 }
 
+func webhookK8sSecretName(webhook string) string {
+	if webhook == "smi-metrics" {
+		return "smi-metrics-tls"
+	}
```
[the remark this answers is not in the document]
cypherfox: done

cli/cmd/upgrade.go
```diff
@@ -507,6 +507,44 @@ func fetchTLSSecret(k *k8s.KubernetesAPI, webhook string, options *upgradeOption
...
+	secret, err := k.CoreV1().
+		Secrets(controlPlaneNamespace).
+		Get(webhookK8sSecretName(webhook), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	value := &charts.TLS{
+		KeyPEM: string(secret.Data["tls.key"]),
+		CrtPEM: string(secret.Data["tls.crt"]),
```
zaharidichev, line 520: maybe worth checking these keys exist here?

zaharidichev (commented):
Thanks a lot @cypherfox! This looks fine to me, apart from a minor nit. An upgrade test sounds like a great addition! Keep in mind this PR is related to #4924, so there will be some follow up work depending on which one merges first. This ofcourse is not a blocker.

cypherfox:
@alpeb, @zaharidichev I can do the update test in the next few days, no prob.

cli/cmd/upgrade.go
```diff
@@ -507,6 +507,50 @@ func fetchTLSSecret(k *k8s.KubernetesAPI, webhook string, options *upgradeOption
...
+
+	if err := options.verifyTLS(value, webhook); err != nil {
+		return nil, err
+	}
+
+	return value, nil
+}
+
+// try to fetch K8sTLSSecret, but revert to TLSSecret if the former fails
+func fetchOrUpdateTLSSecret(k *k8s.KubernetesAPI, webhook string, options *upgradeOptions) (*charts.TLS, error) {
```
adleong, line 541: It looks like this just fetches the TLS data from the secret in one format or another.  Let's rename this to not use the word "Update" since this doesn't update anything.

`fetchTLSSecertWithFallback`? Or call this `fetchTLSSecret` and rename the existing `fetchTLSSecret` to `fetchLinkerdTLSSecret`?

adleong (commented):
Apart from a naming concern, this looks good to me.

As described in https://github.com/linkerd/linkerd2/issues/4813, we're planning to regenerate these certificates during a `linkerd upgrade`.  That means that the code you added in here in `upgrade.go` for fetching the TLS data during upgrade will probably go away soon.  But this shouldn't affect the cert-manager case since we won't render the secret at all if these certs are externally managed.

It's also worth noting that the webhooks do not reload their TLS configs which means that they will need to be restarted to pick up any changes made by cert-manager.

cypherfox:
@adleong I will change the name tomorrow morning.

As discussed in the earliest comments above, the renaming is only  a means to the end of changing the type of the K8s secrets. I am fine with the renaming logic going away as long as the sercret type stays `kubernetes.io/tls`. This is required for cert-manager compatibility. 

The restart is fine, as I will have to solve the CA bundling anyway (IIRC cert-manager doesn't do that yet)

zaharidichev:
@cypherfox Is this ready for a final look? I see you wanted to do some renaming beforehand?

cypherfox:
It is till lacking a proper test. I have made time in my current sprint and will be working on it full time.

zaharidichev (commented):
Great work @cypherfox This looks good to me! You just need to rebase against main and get the latest changes as the update process is failing due to images named incorrectly (note we moved to ghcr.io).

cli/cmd/upgrade_test.go
```diff
@@ -330,6 +330,60 @@ func TestUpgradeOverwriteTracingAddon(t *testing.T) {
...
+
+	upgrade, err := renderUpgrade(t, install, upgradeOpts, upgradeFlags)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedManifests := parseManifestList(expected)
+	upgradeManifests := parseManifestList(upgrade.String())
+	for id, diffs := range diffManifestLists(expectedManifests, upgradeManifests) {
+		for _, diff := range diffs {
+			fmt.Errorf("Unexpected diff in %s:\n%s", id, diff.String())
```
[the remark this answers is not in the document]
cypherfox: I still need to write a proper test for the name change functionality. I commented this out in order to work on the changed health check.
Update for image names coming up.

zaharidichev:
@cypherfox whenever I upgrade from 2.8.1 to this version I get `Back-off pulling image "gcr.io/linkerd-io/grafana:git-a1af8f32"`.

adleong:
@zaharidichev did you do a `bin/docker-build` and `bin/docker-push` from this branch?  that should make the grafana image available.

cli/cmd/testdata/upgrade_grafana_enabled_disabled.yaml
```diff
@@ -0,0 +1,2872 @@
+---
```
adleong, line 1: where did this file come from?  is this leftover from a merge?

adleong (commented):
Looks great to me, but I think there's an extra file in here somehow.

pkg/healthcheck/healthcheck.go
```diff
@@ -1031,7 +1036,10 @@ func (hc *HealthChecker) allCategories() []category {
 						}
 						cert, err := hc.fetchCredsFromSecret(tapTLSSecretName)
 						if err != nil {
```
alpeb, line 1038: TIOLI: How about checking against `kerrors.IsNotFound(err)` instead of `nil` (like you did in `upgrade.go`) to avoid swallowing an error that might be different than the secret not being there? (Like RBAC issues, etc).

alpeb (approved):
Thanks @cypherfox , this looks good to me, barring @adleong comment on the new file.
I just have one small TIOLI comment.

zaharidichev (approved):
Thanks a lot for the hard work @cypherfox! LGTM

cli/cmd/testdata/upgrade_grafana_enabled_disabled.yaml
cypherfox: not mine. Probably a merge error.

adleong (approved):

alpeb:
@cypherfox Looks like another merge beat us and is creating a conflict. Would you mind fixing it? I'll keep an eye open for it so I can merge this PR right away and include it in this week's Edge :wink:

christianhuening:
congratz and thanks @cypherfox ! 🥇

chris-vest:
Thanks for this! Would it be possible to get some documentation on how to use this feature?

cypherfox:
@chris-vest Will do as soon as we get this working internally. I have a helm chart and some integration tests for it, but still need to update our internal ops documentation. I hope to get some of that back into the l5d docs.

chris-vest:
@cypherfox Great, thank you!

PR #4739, opened 2020-07-09
Fetch proxy with specific arch

aliariff:
@cpretzer @olix0r 

Problem:
The `linkerd2-proxy` starting to release artifacts in multi-architecture (https://github.com/linkerd/linkerd2-proxy/pull/593). The naming of the release artifacts now included the architecture name. The current proxy fetcher in `linkerd2` repo does not follow that naming convention.

Solution:
Adjust the proxy fetcher to retrieve the `amd64` version for now.

Note:
This PR must be merged after https://github.com/linkerd/linkerd2-proxy/pull/593. And a new release tag will be published for the proxy and adjustment in file `.proxy-version` need to be made.


Dockerfile-proxy
```diff
@@ -1,7 +1,7 @@
 ARG RUNTIME_IMAGE=debian:buster-20200514-slim
 
 FROM debian:buster-20200514-slim as fetch
-RUN apt-get update && apt-get install -y ca-certificates curl
+RUN apt-get update && apt-get install -y ca-certificates curl jq
```
olix0r, line 4: Is this change actually needed? Where does `jq` get invoked?

aliariff: It is called when we don't specify which version to retrieve.
Ref: https://github.com/linkerd/linkerd2/blob/main/bin/fetch-proxy#L16

olix0r:
@aliariff You can update `.proxy-version` to reference `v2.104.1` and, once that release is complete, this branch should pass CI.

PR #4740, opened 2020-07-09
service topologies: add install flag for EndpointSlices

Matei207:
## How
---

* **What**: Make EndpointSlices opt-in through helm values/cli install values.

* **Helm**: update the values file to include a global `enableEndpointSliceMode` value, along with comments explaining both what the value is for, and that slices are still "experimental". I have also added a conditional in the destination rbac template, so if slices are not enabled rbac is not rendered -- my rationale here is that we shouldn't have any slice specific resources created if they are not going to be in use.
* **CLI**: first instance was to add the flag and the values so it can be parsed. Another thing that I did for the cli was to add the boolean in the config map (more reasoning for this in a later section). The `enableEndpointSliceMode` flag takes after `cniEnabled` and `smiMetrics.Enabled`.
* **Destination Service**: the approach I went with is storing the flag value in the configmap, each component mounts the config in a volume -- when the destination service is started up, we pass the flag value to `NewEndpointsWatcher`. This way, only the `EndpointsWatcher` part of the destination service will know of the flag and condition existence. Unfortunately, this also meant that I had to put the value in the struct, and add it as an argument to a few functions so it can be passed down the call stack.

* **Tests**: I have not added any unit tests yet for the flag. One more important point of consideration is the changes in tests for the destination service since `EndpointsWatcher` now needs the `EndpointSlice` boolean when created, I had to change the tests to reflect this. Tests that had nothing to do with slices had the watcher initialised with a false, while tests that make use of slices have it set to true. Note that setting just the `endpointsSliceEnabled` to true will not mean endpoint slices can be used, slices, in order to be used, have to:
   - be enabled;
   - be registered on the cluster as a resource;
   - exist in the cluster.

## Approaches considered
---
* There are a few ways to do this. I came up with three approaches and then asked @alpeb to give me his opinion, given his extensive experience with flags and the cli.

* **Approach 1**: my initial idea was to use the `linkerd-configmap` in `EndpointSliceAccess`, it looks a lot like the approach I ended up using. Initially, I considered to pass down the controller namespace, in the `EndpointSliceAccess` function I would have fetched the configmap, check the value and return early in case `EndpointSlices` are disabled. This is an infeasible solution though; I found getting the controller namespace passed to the function is quite hard -- ultimately the solution would have looked like what it does now.

* **Approach 2**: the second approach I thought of was to not use a configmap and instead rely on rbac. In `EndpointSliceAccess` I would pass a (hardcoded) component name, since rbac is not namespaced and we know the name ahead of time we can hard code the values and get the rbac for our component through `client-go`. If the rbac does not include permissions for our component to list/get slices, then we can't use them. The downside to this is that it looks a bit hacky in nature, also, it would mean every time we want to use slices somewhere we have to change that components rbac meaning more PRs and a bit of tech debt (?).

* **Approach 3**: use a global value, read the boolean from the configmap, set the global value to the boolean in the `EndpointsWatcher`. Passed on this because global state is not exciting.

### Feedback and notes:
---

* If you can think of any other (and better) ways to do this or of any architectural decisions that seem more sound, please let me know and I'll change it around. Given the possibilities, I think this solution is the most feasible, but I could have easily missed something.
* To keep the diff smaller, I'll enable everything once we sort out the flag, that's why functions are still commented out or hardcoded to return false.


controller/cmd/destination/main.go
```diff
@@ -75,11 +67,30 @@ func Main(args []string) {
 		}
 	}
 
+	enableEndpointSlices := global.GetEndpointSliceEnabled()
+
```
adleong, line 71: I think we should call `EndpointSliceAccess` and error out eagerly if `enableEndpointsSlices` is set but we don't have access.

pkg/k8s/authz.go
```diff
@@ -109,7 +109,11 @@ func ServiceProfilesAccess(k8sClient kubernetes.Interface) error {
 // access to EndpointSlice resources.
 //TODO: Uncomment function and change return type once EndpointSlices
 // are supported and made opt-in through install flag
-func EndpointSliceAccess(k8sClient kubernetes.Interface) bool {
+func EndpointSliceAccess(k8sClient kubernetes.Interface, enableEndpointSlice bool) bool {
```
adleong, line 112: let's leave the second parameter off of this method.  This should just check if we have access.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -149,7 +151,7 @@ func NewEndpointsWatcher(k8sAPI *k8s.API, log *logging.Entry) *EndpointsWatcher
 		UpdateFunc: func(_, obj interface{}) { ew.addService(obj) },
 	})
 
-	if !pkgK8s.EndpointSliceAccess(k8sAPI.Client) {
+	if !pkgK8s.EndpointSliceAccess(k8sAPI.Client, ew.enableEndpointSlices) {
```
adleong, line 154: if we check for access in the main method as I suggest below, we can remove all of these checks in this file and instead treat the `ew.enableEndpointSlices` as authoritative.

charts/linkerd2/values.yaml
```diff
@@ -15,6 +15,11 @@ global:
 
   namespace: linkerd
 
+  # enables the use of EndpointSlice informers for the destination service;
+  # EndpointSliceMode should be set to true only if EndpointSlice feature gate is on;
+  # the feature is still experimental.
+  enableEndpointSliceMode: false
```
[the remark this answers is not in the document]
Matei207: I think it reads much better! :)

cli/cmd/install.go
```diff
@@ -49,6 +49,7 @@ type (
 		disableH2Upgrade            bool
 		disableHeartbeat            bool
 		cniEnabled                  bool
+		endpointSliceModeEnabled    bool
```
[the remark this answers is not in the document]
Matei207: Yup, I have added it in. I ended up creating a different function to test it, since we have to initialise a k8s client to check if the resource type exists. The only point of concern is if I make the validation in the right place -- let me know if it should be done in a different place/phase.

controller/api/destination/watcher/endpoints_watcher.go
Matei207: Amazing suggestion, it reads much better and makes matters less complicated. It also simplifies writing tests; I have uncommented and split up the unit test cases for EndpointSlices since they are no longer coupled to `EndpointSliceAccess`.

Matei207:
I like @Pothulapati's approach, and I think he is right; it will be easier to ensure backwards compatibility when `EndpointSlices` will be the default if we don't have any values in the cm. 

The only downside that I can think of at this point in time is having to wire up each component to use the `--enable-endpoint-slice` program argument if they want to make use of `EndpointSlices`, but that doesn't seem nearly as much of a hassle as deprecating the value in the protocol buffer (as Tharun pointed out to me!).

adleong:
A flag to the destination controller was my first thought too!  But I think the current approach where the destination controller reads the value out of the configmap has some advantages:

* Even if we were to make this a destination controller flag, we would still need to put it in the Linkerd configmap so that the destination manifest would get generated correctly by `linkerd upgrade`.
* Since it has to be in the configmap and the destination controller already reads the configmap, to get the trust domain and cluster domain, it just makes sense to read the endpoint slices value at the same way.

cli/cmd/install.go
```diff
@@ -434,6 +436,10 @@ func (options *installOptions) validateAndBuild(stage string, flags *pflag.FlagS
 		return nil, nil, err
 	}
 
+	if err = validateEndpointSlicesFeature(options.enableEndpointSlices); err != nil {
+		return nil, nil, errors.New("--enableEndpointSlice enabled but resource non-existent in cluster")
```
alpeb, line 440: This is hiding the error returned by `validateEndpointSlicesFeature()`, which might not necessarily be that the resource is non-existent.

Matei207: Hi @alpeb  thanks for the comments! I haven't used helm much so the `charts/linkerd2/templates/_config.tpl` file escaped me! Will run some manual tests with helm installations to make sure everything works as intended 👍 

For `validateEndpointSlicesFeature()` I have changed the returned error by chaining `--enableEndpointSlice true not supported:` with whatever error `validateEndpointSlicesFeature` returns. I could also take another crack at `validateEndpointSlicesFeature()` and return some more customised errors instead of chaining them if you think that'd be a better approach.

alpeb: Thanks @Matei207 chaining errors like that works for me :+1:

Matei207:
**Update**: after a few failed integration tests, @alpeb kindly pointed out to me that there was a missing comma in the configmap which made installing/updating through helm error out.

Also, as a result of of the `EndpointSlice` k8s controller being a bit flaky (both in the integration test and another case with multicluster) it was suggested I add a known warning to the CI so we can ignore the error in the event that installation fails due to the k8s slice controller.

controller/cmd/destination/main.go
Matei207: I've resolved this, after we initialise the K8s.API if `enableEndpointSlices=true` but the resource is not registered we have a call to `log.Fatal`

cli/cmd/install.go
```diff
@@ -690,6 +699,19 @@ func (options *installOptions) validate() error {
 	return nil
 }
 
+func validateEndpointSlicesFeature(enableEndpointSlices bool) error {
+	if !enableEndpointSlices {
```
adleong, line 703: generally I prefer to let the calling code control things like directly rather than passing instructions into the function.  In this case, I'd remove the enableEndpointsSlices argument and instead call this function like:


```
if enableEndpointSlices {
  if err := validateEndpointsSlicesFeature(); err != nil {
    return ....
  }
}
```

Matei207: I can see why that's a better choice :) Thanks for signalling this, I've changed the code, squashed the commits and pushed.

controller/cmd/destination/main.go
adleong: I think we probably want to move the EndpointSliceAccess check up to before initializing the k8s API.  If EndpointSlices are not present on the cluster then initializing the k8s API will hang at "Waiting for caches to sync" instead of getting to the fatal log line.

PR #4741, opened 2020-07-09
update changelog for edge-20.7.2

hawkw:
This edge release moves Linkerd's bundled Prometheus into an add-on.
This makes the Linkerd Prometheus more configurable, gives it a separate
upgrade lifecycle from the rest of the control plane, and allows users
to disable the bundled Prometheus instance. In addition, this release
includes fixes for several issues, including a regression where the
proxy would fail to report OpenCensus spans.

* Prometheus is now an optional add-on, enabled by default
* Custom tolerations can now be specified for control plane resources
  when installing with Helm (thanks @DesmondH0!)
* Evicted data plane pods are no longer considered to be failed by
  `linkerd check --proxy`, fixing an issue where the check would be
  retried indefinitely as long as evicted pods are present
* Fixed a regression where proxy spans were not reported to OpenCensus
* Fixed a bug where the proxy injector would fail to render skipped port
  lists when installed with Helm
* Internal improvements to the proxy for lower latencies under high
  concurrency
* Thanks to @Hellcatlk and @surajssd for adding new unit tests and
  spelling fixes!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,32 +5,22 @@
 This edge release moves Linkerd's bundled Prometheus into an add-on. This makes
 the Linkerd Prometheus more configurable, gives it a separate upgrade lifecycle
 from the rest of the control plane, and allows users to disable the bundled
-Prometheus instance. In addition, the Destination service will now use the new
-Kubernetes `EndpointSlice` resource when available. Finally, this release
-includes fixes for several issues, including a regression where the proxy would
-fail to report OpenCensus spans.
+Prometheus instance. In addition, this release includes fixes for several
+issues, including a regression where the proxy would fail to report OpenCensus
+spans.
 
-* CLI
-  * Evicted data plane pods are no longer considered to be failed by `linkerd
-    check --proxy`
-* Controller
-  * Prometheus is now an add-on! As an add-on, Linkerd's bundled Prometheus is
-    still enabled by default, but it is now optional, and can be disabled or
-    configured
-  * The Destination service will now use the Kubernetes `EndpointSlice` resource
-    on clusters that support it (k8s v1.16+), improving performance (thanks
-    @Matei207!)
-* Internal
-  * Thanks to @Hellcatlk and @surajssd for adding new unit tests and spelling
-    fixes!
-* Helm
-  * Added support for custom tolerations for control plane resources (thanks 
-    @DesmondH0!)
-  * Fixed a bug where the proxy injector would fail to render skipped port lists
-    when installed with Helm
-* Proxy
-  * Fixed a regression where proxy spans were not reported to OpenCensus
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,5 +1,37 @@
 # Changes
 
+## edge-20.7.2
+
+This edge release moves Linkerd's buindled Prometheus into an add-on. This makes
```
olix0r, line 5:
```suggestion
This edge release moves Linkerd's bundled Prometheus into an add-on. This makes
```

CHANGES.md
```diff
@@ -1,5 +1,37 @@
...
+the Linkerd Prometheus more configurable, gives it a separate upgrade lifecycle
+from the rest of the control plane, and allows users to disable the bundled
+Prometheus instance. In addition, the Destination service will now use the new
+Kubernetes `EndpointSlice` resource when available. Finally, this release
```
olix0r, line 9: Is this actually true? I thought we merged support but it's disabled by configuration.

CHANGES.md
```diff
@@ -1,5 +1,37 @@
...
+  * Thanks to @Hellcatlk and @surassjd for adding new unit tests and spelling
+    fixes!
+* Helm
+  * Added support for custom tolerations for control plane resources (thanks 
+    @DesmondH0!)
+  * Fixed a bug where the proxy injector would fail to render skipped port lists
+    when installed with Helm
+* Proxy
+  * Fixed a regression where proxy spans were not reported to OpenCensus
+  * Internal improvements for lower latencies under high concurrency
```
olix0r: Look at recent release notes for preferred style

CHANGES.md
hawkw: I may have misinterpreted the PR description for #4501. In that case, I'm guessing this should be removed from the changelog, and added in whatever release actually enables `EndpointSlice` support?

CHANGES.md
```diff
@@ -1,5 +1,27 @@
 # Changes
 
+## edge-20.7.2
+
+This edge release moves Linkerd's bundled Prometheus into an add-on. This makes
+the Linkerd Prometheus more configurable, gives it a separate upgrade lifecycle
+from the rest of the control plane, and allows users to disable the bundled
+Prometheus instance. In addition, this release includes fixes for several
```
alpeb, line 8: Prometheus is an add-on now, and the goal is indeed to be able to use a an external instance or to not use a prometheus instance at all, but these two options are not implemented yet, so you might wanna leave the "allow users to disable" out for now.

hawkw: Do you think it should be changed to "will allow" or something, or just removed?

alpeb: "will allow" sounds good to me :+1:

PR #4742, opened 2020-07-09
CNI add support for priorityClassName

### LINKED ISSUE #2981: Priority/Priorityclass for linkerd pods
## Feature Request
Non zero priority for linkerd pods

Some clusters have eager processes that will try to keep the nodes busy in all situation. When that happens, linkerd pods (currently installed with a priority = 0) will easily become unresponsive, endangering the behaviour of meshed pods/services (which can easily be all of them)

One of:
- be able to specify a priority or (and?)  priorityClass when installing linkerd on a k8s cluster
- by default use the maximum value possible ([1 billion](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass)) 

Default to max + override possible for other use cases I couldn't think of out there

@grampelberg mentionned kustomize as a potential solution, in which case it might deserve a documentation sample

linkerd upgrade --priorityclass=mypriorityclass

output?
N/A

alex-berger:
As requested in #2981 one should be able to optionally define a priorityClassName for the linkerd2 pods.

With this PR support for priorityClassName is added to the CNI plugin helm chart as well as to the
cli command for installing the CNI plugin. Furthermore, a `--create-namespace` flag is introduce to
allow to skip the namespace creation for setups where the namespace already exists.

Implements part of #2981. Note, this PR only adds priorityClassName to the CNI plugin and therefore does not 
yet full resolve/fix #2981. 

Signed-off-by: alex.berger@nexiot.ch <alex.berger@nexiot.ch>

### DIFF
--- a/charts/linkerd2-cni/templates/cni-plugin.yaml
+++ b/charts/linkerd2-cni/templates/cni-plugin.yaml
@@ -183,6 +183,9 @@ spec:
         beta.kubernetes.io/os: linux
       hostNetwork: true
       serviceAccountName: linkerd-cni
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
+      {{- end }}
       containers:
       # This container installs the linkerd CNI binaries
       # and CNI network config file on each node. The install
--- a/charts/linkerd2-cni/templates/cni-plugin.yaml
+++ b/charts/linkerd2-cni/templates/cni-plugin.yaml
@@ -17,7 +17,7 @@ limitations under the License.
 This file was inspired by
 1) https://github.com/istio/cni/blob/c63a509539b5ed165a6617548c31b686f13c2133/deployments/kubernetes/install/manifests/istio-cni.yaml
 */ -}}
-{{- if .Values.createNamespace -}}
+{{- if .Values.installNamespace -}}
 kind: Namespace
 apiVersion: v1
 metadata:
--- a/cli/cmd/install-cni-plugin.go
+++ b/cli/cmd/install-cni-plugin.go
@@ -39,7 +39,7 @@ type cniPluginOptions struct {
 	destCNIBinDir       string
 	useWaitFlag         bool
 	priorityClassName   string
-	createNamespace     bool
+	installNamespace    bool
 }
 
 func (options *cniPluginOptions) validate() error {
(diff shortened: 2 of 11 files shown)


Pothulapati:
Hey.

This looks good and the `priority-class-name` works both through CLI and Helm. 

Weirdly, setting  `--create-namespace` to false, does not skip emitting the namespace object. It does work with Helm directly. Looks like it has something to do with being overidden by the defaults. Will take a look on this later today!
Also, Can you update the PR description about the new `--create-namespace` flag?

alex-berger:
> Weirdly, setting `--create-namespace` to false, does not skip emitting the namespace object. It does work with Helm directly. Looks like it has something to do with being overidden by the defaults. Will take a look on this later today!
> Also, Can you update the PR description about the new `--create-namespace` flag?

@Pothulapati, I cannot reproduce that. On my machine on commit `42ffae` if I run `./bin/go-run cli install-cni --create-namespace=false` it will skip the namespace object as expected (see output below):

```
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: linkerd-linkerd-cni-cni
  labels:
    linkerd.io/cni-resource: "true"
spec:
  allowPrivilegeEscalation: false
  fsGroup:
    rule: RunAsAny
  hostNetwork: true
  runAsUser:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - hostPath
  - secret
---
...
```

cli/cmd/install-cni-plugin.go
```diff
@@ -102,7 +104,8 @@ assumes that the 'linkerd install' command will be executed with the
 	cmd.PersistentFlags().StringVar(&options.cniPluginImage, "cni-image", options.cniPluginImage, "Image for the cni-plugin")
 	cmd.PersistentFlags().StringVar(&options.logLevel, "cni-log-level", options.logLevel, "Log level for the cni-plugin")
 	cmd.PersistentFlags().StringVar(&options.destCNINetDir, "dest-cni-net-dir", options.destCNINetDir, "Directory on the host where the CNI configuration will be placed")
-	cmd.PersistentFlags().StringVar(&options.destCNIBinDir, "dest-cni-bin-dir", options.destCNIBinDir, "Directory on the host where the CNI plugin binaries reside")
+	cmd.PersistentFlags().StringVar(&options.priorityClassName, "priority-class-mame", options.priorityClassName, "Pod priorityClassName for CNI daemonset's pods")
```
Matei207, line 107: there's a typo in name

charts/linkerd2-cni/templates/cni-plugin.yaml
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 This file was inspired by
 1) https://github.com/istio/cni/blob/c63a509539b5ed165a6617548c31b686f13c2133/deployments/kubernetes/install/manifests/istio-cni.yaml
 */ -}}
-
+{{- if .Values.createNamespace -}}
```
Matei207, line 20: It seems to me that before the change the namespace was automatically created. What do you think about changing the logic here, something like:

```
{{- if not .Values.skipNamespace}}
 kind: Namespace
 ...
```
 
In addition, this would mean changing the flag name from 'createNamespace' to something like `skip-namespace-creation` or `omit-namespace`. WDYT @alex-berger @Pothulapati

cli/cmd/install-cni-plugin.go
alex-berger: Fixed

Pothulapati:
@alex-berger Can you merge with the latest main to make the tests pass? :)

PR #4744, opened 2020-07-09
Fix upgrade integration test

alpeb:
An inappropriate variable reuse resulted in the failure of the test for
upgrading using manifests. This only happened when the upgrade was
retried a second time (when there's a discrepancy in the heartbeat cron
schedule, which is benign).


adleong:
For my go-education, what is the bug here?  My understanding is that the `args :=` line creates a new variable which shadows the one in the outer scope.  Is the outer scoped `args` variable getting updated somehow?  How is this possible?

alpeb:
> For my go-education, what is the bug here? My understanding is that the `args :=` line creates a new variable which shadows the one in the outer scope. Is the outer scoped `args` variable getting updated somehow? How is this possible?
> 
> e.g.
> 
> https://play.golang.org/p/QCBBt0Rz3KH

In this case what goes after `if out != upgradeFromManifests {` wants to reuse `args` as it was set before it got overridden with `args := `

PR #4751, opened 2020-07-11
add fish shell completion

### LINKED ISSUE #4208: Add fish shell completions
## Feature Request
Add fish shell completions. Currently only bash/zsh are supported.

Adding completions to the very popular fish shell

Add support for `linkerd completion fish`.

WLun001:
fixes #4208


WLun001:
Not sure how to fix the failed CI checks. Could be change of dependencies? Please advise

adleong:
@WLun001 that's right.  since you updated the go dependencies, you'll need to run the `bin/update-go-deps-shas` script to update the Dockerfiles to use the new go dependency images.

cli/cmd/completion.go
```diff
@@ -29,15 +29,22 @@ func newCmdCompletion() *cobra.Command {
   source <(linkerd completion zsh)
 
   # zsh on osx / oh-my-zsh
-  linkerd completion zsh > "${fpath[1]}/_linkerd"`
+  linkerd completion zsh > "${fpath[1]}/_linkerd"
+
+  # Fish:
```
kleimkuhler, line 34: Small suggestion, but just keeping it consistent with the above help text

```suggestion
  # fish
```

cli/cmd/completion.go
```diff
@@ -29,15 +29,22 @@ func newCmdCompletion() *cobra.Command {
...
+  linkerd completion fish | source
+
+  # fish  To load completions for each session, execute once:
```
kleimkuhler, line 37:
```suggestion
  # To load fish shell completions for each session, execute once:
```

PR #4752, opened 2020-07-12
conformance validation: make `checkOutput` public

mayankshah1607:
### Problem

The conformance validation project needs a way to validate the output of `linkerd check`. A very simple way to do this would be to capture the output of `check` and compare it against some predefined `.golden` file which holds the desired output, much like the integration tests. The  issue with this is that for conformance validation, users are allowed to specify the version of Linkerd CLI to use for performing the tests, and it is possible for different versions to have different sets of checks. It is also not feasible to have a `.golden` file with the desired check output for each stable and edge version.

### Solution

Capture the JSON output of the check results (`linkerd check -o json`), and unmarshal it to an instance of `CheckOutput`. We can now iterate through this object to see if any `Check.Result` holds `"error"`.

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -296,7 +296,7 @@ type CheckCategory struct {
 	Checks []*Check `json:"checks"`
 }
 
-// check is a user-facing version of `healthcheck.CheckResult`, for output via
+// Check is a user-facing version of `healthcheck.CheckResult`, for output via
 // `linkerd check -o json`.
 type Check struct {
 	Description string      `json:"description"`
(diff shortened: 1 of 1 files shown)


olix0r:
Hmmm... one of the challenges here is that this makes it all part of a public API, and we won't be able to easily know when we're breaking conformance with changes to this code... And we don't make any effort to version this semantically so that other repos can explicitly depend on it...

How bad would it be to copy/paste this for now?

mayankshah1607:
@olix0r Ah, that makes sense. Didn't really give it that much thought.
> How bad would it be to copy/paste this for now?

Not at all, copy paste would work just fine. In fact that's how we're doing it right now. Should I close this one then?

ihcsim:
@mayankshah1607 Can we close this then?

PR #4755, opened 2020-07-13
grafana: move default values into values file

Pothulapati:
This PR moves default values into add-on specific values.yaml thus
allowing us to update default values as they would not be present in
`linkerd-config-addons` cm.

This PR also has the following changes:
- Remove `grafana.name` field, as it adds unnecessary confusion while
not being used a lot.
- Adds `grafana.image.tag` field


charts/add-ons/grafana/values.yaml
```diff
@@ -0,0 +1,2 @@
+image:
+  name: gcr.io/linkerd-io/grafana
```
alpeb, line 2: This is no longer used in the template, so might wanna remove it here?

alpeb: Sorry, I confused the name of the image with the resource name.

alpeb: OTOH it looks like `image.tag` is missing here

alpeb: Oh I didn't see the nested `default` in the template. So `image.tag` will only be used to override the tag coming from the controller version or linkerd version. And it's already documented in both the readme file and the main values.yaml. Good :+1:

charts/linkerd2/README.md
```diff
@@ -201,6 +201,7 @@ The following table lists the configurable parameters for the Grafana Add-On.
 | `grafana.enabled`                     | Flag to enable grafana instance to be installed                                                                                                                                                | `true`
 | `grafana.name`                | Name of the grafana instance Service                                                                                                                                                 | `linkerd-grafana`                             |
```
alpeb, line 202: You can get rid of this line

charts/add-ons/grafana/templates/grafana.yaml
```diff
@@ -128,7 +128,7 @@ spec:
         # see https://github.com/grafana/grafana/issues/20096
         - name: GODEBUG
           value: netdns=go
-        image: {{.Values.image.name}}:{{default .Values.global.linkerdVersion .Values.global.controllerImageVersion}}
+        image: {{.Values.image.name}}:{{ default (default .Values.global.linkerdVersion .Values.global.controllerImageVersion) .Values.image.tag}}
```
ihcsim, line 131: > .Values.image.tag

Should this be `.Values.grafana.image.tag`?

cli/cmd/testdata/grafana_overwrite.yaml
```diff
@@ -1,5 +1,5 @@
 grafana:
   enabled: true
-  name: linkerd-grafana-overwrite
   image:
     name: linkerd-image-overwrite
+    tag: image-tag-overwrite
```
ihcsim, line 5: Where does `image-tag-overwrite` come from?

cli/cmd/testdata/install_custom_registry.golden
```diff
@@ -2629,7 +2626,7 @@ spec:
         # see https://github.com/grafana/grafana/issues/20096
         - name: GODEBUG
           value: netdns=go
-        image: my.custom.registry/linkerd-io/grafana:install-control-plane-version
+        image: gcr.io/linkerd-io/grafana:install-control-plane-version
```
ihcsim, line 2629: Any idea why this is changed? The `install_test.go` is still using the `my.custom.registry` override as seen [here](https://github.com/linkerd/linkerd2/blob/46d22f8b04b10fd99dd16a9aec840a3d3c6bea9c/cli/cmd/install_test.go#L218-L227).

pkg/healthcheck/healthcheck_addons.go
```diff
@@ -53,9 +56,15 @@ func (hc *HealthChecker) addOnCategories() []category {
 					check: func(context.Context) error {
 						if grafana, ok := hc.addOns[l5dcharts.GrafanaAddOn]; ok {
 							name, err := getString(grafana, "name")
-							if err != nil {
+							if err != nil && err != errorKeyNotFound {
```
ihcsim, line 59: It will be helpful for the error message to include which key is missing. Would wrapping `errorKeyNotFound` work?

```
diff --git a/pkg/healthcheck/healthcheck_addons.go b/pkg/healthcheck/healthcheck_addons.go
index c880fd14..a824ca8f 100644
--- a/pkg/healthcheck/healthcheck_addons.go
+++ b/pkg/healthcheck/healthcheck_addons.go
@@ -76,7 +76,7 @@ func (hc *HealthChecker) addOnCategories() []category {
 					check: func(context.Context) error {
 						if grafana, ok := hc.addOns[l5dcharts.GrafanaAddOn]; ok {
 							name, err := getString(grafana, "name")
-							if err != nil && err != errorKeyNotFound {
+							if err != nil && errors.Is(err, errorKeyNotFound) {
 								return err
 							}
 
@@ -283,7 +283,7 @@ func getString(i interface{}, k string) (string, error) {
 
 	v, ok := m[k]
 	if !ok {
-		return "", errorKeyNotFound
+		return "", fmt.Errorf("%w: %s", errorkeyNotFound, k)
 	}
 
 	res, ok := v.(string)
```

charts/add-ons/grafana/templates/grafana.yaml
Pothulapati: @ihcsim As its the sub-chart, we don't actually need `.grafana` for referencing values as helm makes sure we do that, and for CLI we manually do the same.

cli/cmd/testdata/grafana_overwrite.yaml
Pothulapati: This comes from the fact that `image.tag` being overwritten, will be saved in `linkerd-config-addons` and thus being useful for upgrades.

cli/cmd/testdata/install_custom_registry.golden
Pothulapati: Yeah, This was because we removed using cli flags, to propogate stuff to add-on fields. This means that users who would want to change to a custom registry ( in this case, or use a different image name) would need to update the field using `--addon-config` through `grafana.image.name`

So, In this case, users would require to update that individually. Do you think we should make sure we propogate configuration changes to add-ons if the flag is a global one (like the custom registry one here)?

ihcsim: What is the current behaviour for other global variables? If they are inherited by the add-ons, I think we should do the same here.

Pothulapati: @ihcsim We definitely follow the `global` fields for add-ons too, but My question was around the CLI flags? Currently, Its a bit confusing whether to support them or not. as users can always update fields using add-on level configuration. Should we propogate the configuration through the CLI flags too?

ihcsim: Are any other "global" flags repeated in the add-on config? I'd assume add-on config only affects the add-ons, not the entire control plane. If we want to deprecate a CLI flag, we will have to mark it as so, not just remove them.

pkg/healthcheck/healthcheck_addons.go
Pothulapati: @ihcsim I tried the above wrapping, but then it wasn't recognised as a errorKeyNotFound for us to compare. :thinking: 

Also, Updated the code to use `errors.Is` instead of `==`

charts/linkerd2/values.yaml
```diff
@@ -249,9 +249,10 @@ smiMetrics:
 
 grafana:
   enabled: true
-  name: linkerd-grafana
-  image:
-    name: gcr.io/linkerd-io/grafana
+  # name: linkerd-grafana
```
ihcsim, line 252: If this is no longer needed, we should remove it.

cli/cmd/testdata/grafana_overwrite.yaml
ihcsim: ah.. so this is an input file.

pkg/healthcheck/healthcheck_addons.go
```diff
@@ -53,9 +56,15 @@ func (hc *HealthChecker) addOnCategories() []category {
 					check: func(context.Context) error {
 						if grafana, ok := hc.addOns[l5dcharts.GrafanaAddOn]; ok {
 							name, err := getString(grafana, "name")
-							if err != nil {
+							if err != nil && !errors.Is(err, errorKeyNotFound) {
 								return err
 							}
+
+							if err == errorKeyNotFound {
```
ihcsim, line 63: Does `if errors.Is(err, errorKeyNotFound)` work here?

PR #4756, opened 2020-07-14
prometheus: add add-on checks

Pothulapati:
Implements third part of https://github.com/linkerd/rfc/pull/16, https://github.com/linkerd/linkerd2/issues/4375
As linkerd-prometheus is optional now, the checks can also be separated
and should be only work when the prometheus add-on is installed.

PR #4724 already removes the existing checks, and here we separate
the checks.

This is done by re-using the add-on check code.

### DIFF
--- a/pkg/healthcheck/healthcheck_addons.go
+++ b/pkg/healthcheck/healthcheck_addons.go
@@ -18,13 +18,16 @@ const (
 	// LinkerdGrafanaAddOnChecks adds checks related to grafana add-on components
 	LinkerdGrafanaAddOnChecks CategoryID = "linkerd-grafana"
 
+	// LinkerdPrometheusAddOnChecks adds checks related to Prometheus add-on components
+	LinkerdPrometheusAddOnChecks CategoryID = "linkerd-prometheus"
+
 	// LinkerdTracingAddOnChecks adds checks related to tracing add-on components
 	LinkerdTracingAddOnChecks CategoryID = "linkerd-tracing"
 )
 
 var (
 	// AddOnCategories is the list of add-on category checks
-	AddOnCategories = []CategoryID{LinkerdAddOnChecks, LinkerdGrafanaAddOnChecks, LinkerdTracingAddOnChecks}
+	AddOnCategories = []CategoryID{LinkerdAddOnChecks, LinkerdPrometheusAddOnChecks, LinkerdGrafanaAddOnChecks, LinkerdTracingAddOnChecks}
 )
 
 // addOnCategories contain all the checks w.r.t add-ons. It is strongly advised to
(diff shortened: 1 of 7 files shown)


pkg/healthcheck/healthcheck_addons.go
```diff
@@ -44,6 +47,54 @@ func (hc *HealthChecker) addOnCategories() []category {
...
+						}
+						return &SkipError{Reason: "prometheus add-on not enabled"}
+					},
+				},
+				{
+					description: "prometheus add-on config map exists",
+					warning:     true,
+					check: func(context.Context) error {
+						if _, ok := hc.addOns[l5dcharts.PrometheusAddOn]; ok {
+							_, err := hc.kubeAPI.CoreV1().ConfigMaps(hc.ControlPlaneNamespace).Get("linkerd-prometheus-config", metav1.GetOptions{})
```
zaharidichev, line 68: Can't you just return the err here ?

pkg/healthcheck/healthcheck_addons.go
```diff
@@ -44,6 +47,54 @@ func (hc *HealthChecker) addOnCategories() []category {
...
+				{
+					description:         "prometheus pod is running",
+					warning:             true,
+					retryDeadline:       hc.RetryDeadline,
+					surfaceErrorOnRetry: true,
+					check: func(context.Context) error {
+						if _, ok := hc.addOns[l5dcharts.PrometheusAddOn]; ok {
+							// populate controlPlanePods to get the latest status, during retries
+							var err error
+							hc.controlPlanePods, err = hc.kubeAPI.GetPodsByNamespace(hc.ControlPlaneNamespace)
```
zaharidichev, line 86: do we really need to do that here?

Pothulapati: Hmmm? Updating `hc.controlPlanePods` ?

Yes because then we would be waiting indefinitely for the control-plane pod to become ready as we are not updating the status of the pods in real time. :thinking:

test/integration/install_test.go
```diff
@@ -397,7 +397,7 @@ func TestCheckHelmStableBeforeUpgrade(t *testing.T) {
 		t.Skip("Skipping as this is not a helm upgrade test")
 	}
 
-	testCheckCommand(t, "", TestHelper.UpgradeHelmFromVersion(), "", TestHelper.UpgradeHelmFromVersion())
+	testCheckCommand(t, "", TestHelper.UpgradeHelmFromVersion(), "", TestHelper.UpgradeHelmFromVersion(), false)
```
alpeb, line 400: I see this is what we did when the add-on checks were introduced in the past, but given this new parameter is only set to false here, I think it might be clearer that instead of modifying the `testCheckCommand` function along with all its calling points, we just comment out this call here, with a to-do saying this should be uncommented when the next stable after 2.8.1 comes out.

alpeb: @Pothulapati My suggestion was to comment out this `testCheckCommand()` call because it's a no-op in this case, and so not having to modify its definition nor the other places where it's called from.

Pothulapati: @alpeb but that would mean we would be totally skipping the check tests here (which can fail). My thinking with the `compareOutput` flag is that we would only not check the output but atleast have a way to check if the command failed.

alpeb: Currently, if `compareOutput` is false then `testCheckCommand()` doesn't do anything besides initializing some vars.

PR #4761, opened 2020-07-16
Add sidecar container support for linkerd-prometheus helm chart

memory:
### Subject

Add sidecar container support for linkerd-prometheus helm chart

### Problem

Exporting data from prometheus to external systems (e.g. cloudwatch, stackdriver, datadog) requires running an external process to follow the prometheus write-ahead log (WAL) and process the data found herein.  Presently, doing this requires installing linkerd with a combination of the default installer and kustomize: the helm installation method offers no way to add containers to the prometheus pod.

### Solution

This PR adds a `sidecarContainers: []` key to the prometheus add-on section: each list element is a pod container definition.

### DIFF
--- a/charts/add-ons/prometheus/templates/prometheus.yaml
+++ b/charts/add-ons/prometheus/templates/prometheus.yaml
@@ -231,6 +231,9 @@ spec:
       securityContext:
         fsGroup: 65534
       containers:
+      {{- if .Values.sidecarContainers -}}
+      {{- toYaml .Values.sidecarContainers | trim | nindent 6 }}
+      {{- end}}
       - args:
         {{- range $key, $value := .Values.args}}
         - --{{ $key }}{{ if $value }}={{ $value }}{{ end }}
--- a/charts/linkerd2/README.md
+++ b/charts/linkerd2/README.md
@@ -229,6 +229,7 @@ The following table lists the configurable parameters for the Prometheus Add-On.
 | `prometheus.resources.memory.request`  | Amount of memory that the prometheus container requests                                                                                                                               ||
 | `prometheus.ruleConfigMapMounts`             | Alerting/recording rule ConfigMap mounts (sub-path names must end in `_rules.yml` or `_rules.yaml`)                                                                                   | `[]`                                 |
 | `prometheus.scrapeConfigs`             | A scrape_config section specifies a set of targets and parameters describing how to scrape them.                                                        | `[]`                                 |
+| `prometheus.sidecarContainers`         | A sidecarContainers section specifies a list of secondary containers to run in the prometheus pod e.g. to export data to non-prometheus systems | `[]`                                 |
 
 Most of the above configuration match directly with the official Prometheus
 configuration which can be found [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration)
--- a/charts/linkerd2/values.yaml
+++ b/charts/linkerd2/values.yaml
@@ -298,6 +298,33 @@ prometheus:
   # - name: recording-rules
   #   subPath: recording_rules.yml
   #   configMap: linkerd-prometheus-rules
+  ###
+  ### Sidecar containers allow access to the prometheus data directory,
+  ### e.g. for exporting data to non-prometheus systems.
+  # sidecarContainers:
+  # - name: sidecar
+  #   image: gcr.io/myproject/stackdriver-prometheus-sidecar
+  #   lifecycle:
+  #     # n.b. only valid on k8s >=1.19
+  #     type: Sidecar
+  #   imagePullPolicy: always
+  #   command:
+  #   - /bin/sh
+  #   - -c
+  #   - |
+  #     exec /bin/stackdriver-prometheus-sidecar \
+  #       --stackdriver.project-id=myproject \
+  #       --stackdriver.kubernetes.location=us-central1 \
+  #       --stackdriver.kubernetes.cluster-name=mycluster \
+  #       --prometheus.wal-directory=/data/wal \
+  #       --log.level=info
+  #     volumeMounts:
+  #     - mountPath: /data
+  #       name: data
+  #   ports:
+  #   - name: foo
+  #     containerPort: 9091
+  #     protocol: TCP
   ### WARNING: persistence is experimental and has not been tested/vetted by the Linkerd team.
   ### As such, please refer to https://linkerd.io/2/tasks/exporting-metrics/ for the recommended approach to metrics data retention.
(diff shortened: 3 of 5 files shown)


charts/add-ons/prometheus/templates/prometheus.yaml
```diff
@@ -231,6 +231,34 @@ spec:
       securityContext:
         fsGroup: 65534
       containers:
+      {{- range .Values.sidecarContainers }}
```
Pothulapati, line 234: I was thinking why not let the user define the full container info in `values.yaml` and we just put that here without having to deal with each individual field. :thinking:

memory: I went back and forth on this a lot, tbh, and ended up coming down on the side of ensuring that the /data and config mounts were always there, but I think you're right: we can just put that in the example.

memory:
Possibly I'm confused: I thought that the test/golden files in cli/cmd/testdata/ are there to test the output of `linkerd install` / `linkerd update`, neither of which would be changed by alterations to the helm chart?

cpretzer:
hi @memory this tripped me up a bit when I first started looking at it, too. The `install` and `update` commands use the [helm charts](https://github.com/linkerd/linkerd2/blob/main/cli/cmd/install.go#L843) when generating the yaml.

It's some well designed code which allows reuse of the helm templates for helm and CLI installation. 😄

memory:
Side note: my initial cut at this failed the tests because I gave a nonexistent example docker image in the golden test data. This seems suboptimal.

charts/linkerd2/README.md
```diff
@@ -229,6 +229,7 @@ The following table lists the configurable parameters for the Prometheus Add-On.
 | `prometheus.resources.memory.request`  | Amount of memory that the prometheus container requests                                                                                                                               ||
 | `prometheus.ruleConfigMapMounts`             | Alerting/recording rule ConfigMap mounts (sub-path names must end in `_rules.yml` or `_rules.yaml`)                                                                                   | `[]`                                 |
 | `prometheus.scrapeConfigs`             | A scrape_config section specifies a set of targets and parameters describing how to scrape them.                                                        | `[]`                                 |
+| `prometheus.sidecarContainers`         | A sidecarContainers section specifies a list of secondary containers to run in the prometheus pod e.g. to export data to non-prometheus systems | `[]`                                 |
```
Pothulapati, line 232: Should we call the field just `containers`, as there is nothing special about them? :thinking:

charts/linkerd2/values.yaml
```diff
@@ -298,6 +298,33 @@ prometheus:
...
+  ###
+  ### Sidecar containers allow access to the prometheus data directory,
+  ### e.g. for exporting data to non-prometheus systems.
+  # sidecarContainers:
+  # - name: sidecar
+  #   image: gcr.io/myproject/stackdriver-prometheus-sidecar
+  #   lifecycle:
+  #     # n.b. only valid on k8s >=1.19
+  #     type: Sidecar
+  #   imagePullPolicy: always
```
Pothulapati, line 310:
```suggestion
  #   imagePullPolicy: Always
```

This configuration actually failed, so thought it would be better to have something that works atleast as a example :)

charts/linkerd2/README.md
memory: maybe `extraContainers`?  just `containers` seems like it's suggesting that you _need_ to define something, which you don't.

cpretzer: `bonusContainers`! 😄 

Does prometheus have a standardized way of referring to these processes? If so, that seems like the right pattern to follow.

Pothulapati:
@memory can you sign the DCO by following https://github.com/linkerd/linkerd2/pull/4761/checks?check_run_id=900071375

charts/linkerd2/README.md
memory: To the best of my knowledge it does not.  However, _kubernetes_ officially calls them sidecars (cf https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/#how-pods-manage-multiple-containers) so I'm going to assert moderately strongly that is the terminology we should use. :)

cpretzer: That's a well-reasoned point and makes sense to me. WDYT @Pothulapati?

Pothulapati: Yep. Makes sense for me too!

charts/linkerd2/values.yaml
```diff
@@ -298,6 +298,30 @@ prometheus:
...
+  #   - |
+  #     exec /bin/stackdriver-prometheus-sidecar \
+  #       --stackdriver.project-id=myproject \
+  #       --stackdriver.kubernetes.location=us-central1 \
+  #       --stackdriver.kubernetes.cluster-name=mycluster \
+  #       --prometheus.wal-directory=/data/wal \
+  #       --log.level=info
+  #     volumeMounts:
+  #     - mountPath: /data
+  #       name: data
```
alpeb, line 320: This section has an extra indentation

cli/cmd/testdata/prom-config.yaml
```diff
@@ -35,3 +35,26 @@ prometheus:
     configMap: linkerd-prometheus-rules
   remoteWrite:
   - url: http://cortex-service.default:9009/api/prom/push
+  sidecarContainers:
+  - name: sidecar
+    lifecycle:
+      # n.b. only valid on k8s >=1.19
+      type: Sidecar
```
alpeb, line 42: We can leave this type as an example, but would you remove that comment? Sidecars won't make it to 1.19 (when they will is anybody's guess unfortunately). Otherwise we'll have to edit this out later on :wink:

charts/linkerd2/values.yaml
memory: Oops, indeed it does.  Fixed.

cli/cmd/testdata/prom-config.yaml
memory: Sadness; removed.

memory:
Okay, the upgrade-edge integration test appears to be failing on the prometheus install, which seems likely to be related to this PR, but the output does not suggest any obvious culprit.  Any hints?

Pothulapati:
@memory It was actually another PR which got merged causes this, and hence there are other PR's failing for the same reason. @alpeb has raised a fix with #4797

memory:
Aha! Well in that case, I'm going to assert that _this_ PR is mergeable as-is. :)

PR #4764, opened 2020-07-17
tracing: move defaults to add-on specific values.yaml

Pothulapati:
This PR moves the default values from `linkerd2/values.yaml` to
`add-ons/tracing/values.yaml`, thus making them not appear in
`linkerd-config-addons` and giving us the capability to change defaults
during upgrades.

This PR, also updates web to use `linkerd-jaeger` which is the default
jaeger instance name


charts/linkerd2/templates/web.yaml
```diff
@@ -72,7 +72,7 @@ spec:
         - -grafana-addr={{.Values.grafana.name}}.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:3000
         {{- end}}
         {{- if .Values.tracing.enabled }}
-        - -jaeger-addr={{.Values.tracing.jaeger.name}}.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:16686
+        - -jaeger-addr=linkerd-jaeger.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:16686
```
alpeb, line 75: Why hard-coding `linkerd-jaeger` if it's still declared in `values.yaml`?

charts/linkerd2/values.yaml
```diff
@@ -312,11 +312,11 @@ prometheus:
...
-    image: jaegertracing/all-in-one:1.17.1
-    # resources:
+  # collector:
+  #   name: linkerd-collector
+  #   image: omnition/opencensus-collector:0.1.11
+  #   resources:
+  # jaeger:
+  #   name: linkerd-jaeger
+  #   image: jaegertracing/all-in-one:1.17.1
+  #   resources:
```
kleimkuhler, line 322: These can be removed

charts/linkerd2/templates/web.yaml
Pothulapati: yeah but its a add-on level value and hence control-plane components can't actually know these values, especially through Helm?

charts/linkerd2/values.yaml
Pothulapati: hmmm? :thinking: We actually comment values that can be allowed for overriden right? just like we did for Prometheus https://github.com/linkerd/linkerd2/blob/main/charts/linkerd2/values.yaml#L268

charts/linkerd2/templates/web.yaml
alpeb: Ok, OTOH if `jaeger.name` gets overridden at the add-on level, it won't work with the web component because the names won't match, right? If so, it might make more sense to hard-code it at the add-on level as well (and the same would apply for the other addons names).

charts/linkerd2/values.yaml
kleimkuhler: Ah yep I didn't realize that! Thanks

charts/linkerd2/templates/web.yaml
Pothulapati: Yeah, so should we remove the `jaeger.name` and `collector.name` fields in this PR itself?

alpeb: I think so, yeah.

cpretzer:
@Pothulapati this is installing fine for me locally, but I'm not sure what is causing this failure: https://github.com/linkerd/linkerd2/pull/4764/checks?check_run_id=928137900

What is the expected value of `.Values.global.proxy.component`? I suspect that it's trying to compare the string `"linkerd-destination"` with nil, but I haven't been able to confirm that.

charts/linkerd2/values.yaml
```diff
@@ -340,11 +340,11 @@ prometheus:
...
-  collector:
-    name: linkerd-collector
-    image: omnition/opencensus-collector:0.1.11
-    # resources:
-  jaeger:
-    name: linkerd-jaeger
-    image: jaegertracing/all-in-one:1.17.1
-    # resources:
+  # collector:
+  #   name: linkerd-collector
```
alpeb, line 344: `name` here (and under `jaeger`) should be removed not that they're being hard-coded

ihcsim:
@Pothulapati Is this still relevant in light of your new minimal control plane RFC?

Pothulapati:
This is mostly related to the state management of add-ons, which is being changed in the newer RFC. So, Closing this!

PR #4765, opened 2020-07-17
Migrate CI to docker buildx and other improvements

alpeb:
## Motivation
- Improve build times in forks. Specially when rerunning builds because of some flaky test.
- Start using `docker buildx` to pave the way for multiplatform builds.

## Performance improvements
These timings were taken for the `kind_integration.yml` workflow when we merged and rerun the lodash bump PR (#4762)

Before these improvements:
- when merging: `24:18`
- when rerunning after merge (docker cache warm): `19:00`
- when running the same changes in a fork (no docker cache): `32:15`

After these improvements:
- when merging: `25:38`
- when rerunning after merge (docker cache warm): `19:25`
- when running the same changes in a fork (docker cache warm): `19:25`

As explained below, non-forks and forks now use the same cache, so the important take is that forks will always start with a warm cache and we'll no longer see long build times like the `32:15` above.
The downside is a slight increase in the build times for non-forks (up to a little more than a minute, depending on the case).

## Build containers in parallel
The `docker_build` job in the `kind_integration.yml`, `cloud_integration.yml` and `release.yml` workflows relied on running `bin/docker-build` which builds all the containers in sequence. Now each container is built in parallel using a matrix strategy.

## New caching strategy
CI now uses `docker buildx` for building the container images, which allows using an external cache source for builds, a location in the filesystem in this case. That location gets cached using actions/cache, using the key `{{ runner.os }}-buildx-${{ matrix.target }}-${{ env.TAG }}` and the restore key `${{ runner.os }}-buildx-${{ matrix.target }}-`.

For example when building the `web` container, its image and all the intermediary layers get cached under the key `Linux-buildx-web-git-abc0123`. When that has been cached in the `main` branch, that cache will be available to all the child branches, including forks. If a new branch in a fork asks for a key like `Linux-buildx-web-git-def456`, the key won't be found during the first CI run, but the system falls back to the key `Linux-buildx-web-git-abc0123` from `main` and so the build will start with a warm cache (more info about how keys are matched in the [actions/cache docs](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key)).

## Packet host no longer needed
To benefit from the warm caches both in non-forks and forks like just explained, we're required to ditch doing the builds in Packet and now everything runs in the github runners VMs.
As a result there's no longer separate logic for non-forks and forks in the workflow files; `kind_integration.yml` was greatly simplified but `cloud_integration.yml` and `release.yml` got a little bigger in order to use the actions artifacts as a repository for the images built. This bloat will be fixed when support for [composite actions](https://github.com/actions/runner/blob/users/ethanchewy/compositeADR/docs/adrs/0549-composite-run-steps.md) lands in github.

## Local builds
You still are able to run `bin/docker-build` or any of the `docker-build.*` scripts. And to make use of buildx, run those same scripts after having set the env var `DOCKER_BUILDKIT=1`. Using buildx supposes you have installed it, as instructed [here](https://github.com/docker/buildx).

## Other
- A new script `bin/docker-cache-prune` is used to remove unused images from the cache. Without that the cache grows constantly and we can rapidly hit the 5GB limit (when the limit is attained the oldest entries get evicted).
- The `go-deps` dockerfile base image was changed from `golang:1.14.2` (ubuntu based) to `golang-1:14.2-alpine` also to conserve cache space.

## Known issues to be addressed in a followup PR
- Most dockerfiles rely on the `go-deps` base image at a hard-coded tag, that they retrieve from the gcr registry. Whenever that base image changes, it gets rebuilt prior to building the other images. Now that we're using the docker-container driver for buildx, it can't use the local cache for retrieving the `go-deps` image just built (see docker/buildx#156). So changes to `go-deps` will break the build.


.github/workflows/release.yml
```diff
@@ -14,37 +10,52 @@ on:
...
         DOCKER_TRACE: 1
       run: |
         export PATH="`pwd`/bin:$PATH"
-        bin/docker-build
-    - name: Create artifact with Windows CLI binary
+        docker buildx create --driver docker-container --use
+        bin/docker-build-${{ matrix.target }}
+        # changes generate new images while the existing ones don't get removed
+        # so we manually do that to avoid bloating the cache
+        bin/docker-cache-prune
```
aliariff, line 48: @alpeb 
Create its own step for the cache pruning? what do you think?

.github/workflows/release.yml
```diff
@@ -14,37 +10,52 @@ on:
...
+          ${{ runner.os }}-buildx-${{ matrix.target }}-
     - name: Build docker images
       env:
-        DOCKER_HOST: ssh://linkerd-docker
         DOCKER_TRACE: 1
       run: |
         export PATH="`pwd`/bin:$PATH"
-        bin/docker-build
-    - name: Create artifact with Windows CLI binary
+        docker buildx create --driver docker-container --use
```
aliariff, line 44: what about using existed actions to setup the buildx https://github.com/crazy-max/ghaction-docker-buildx
the action is set up with builtin capability to perform multi-arch build (Ref: https://github.com/crazy-max/ghaction-docker-buildx/blob/master/src/main.ts#L27)

aliariff:
@cpretzer @alpeb 
I will continue working on the multi-arch images on top of this.
This PR is a good introduction to moving on to `buildx`.
I am excited to see this merged. 🚀

cpretzer:
Thanks @aliariff this is looking good. Can't wait to start testing it

.github/workflows/release.yml
```diff
@@ -68,21 +79,23 @@ jobs:
...
-      run: |
-        mkdir -p ~/.ssh/
-        touch ~/.ssh/id && chmod 600 ~/.ssh/id
-        echo "${{ secrets.DOCKER_SSH_CONFIG }}"  > ~/.ssh/config
-        echo "${{ secrets.DOCKER_PRIVATE_KEY }}" > ~/.ssh/id
-        echo "${{ secrets.DOCKER_KNOWN_HOSTS }}" > ~/.ssh/known_hosts
-        ssh linkerd-docker docker version
     - name: Push docker images to registry
       env:
         DOCKER_HOST: ssh://linkerd-docker
```
aliariff, line 101: forget to remove this

cpretzer:
I think the checks need to be updated in GitHub settings to remove `Docker build` since that check is named/handled differently with these improvements

Dockerfile-go-deps
```diff
@@ -5,7 +5,7 @@
 #
 # When this file is changed, run `bin/update-go-deps-shas`.
 
-FROM golang:1.14.2
+FROM golang:1.14.2-alpine
```
kleimkuhler, line 8: Is this related to the change?

.github/workflows/release.yml
alpeb: You definitely should make use of that action when you tackle multi-arch builds. For this PR, it's more practical to just use this command.

.github/workflows/release.yml
alpeb: Sure, that'll distinguish its output from the build step output :+1:

Dockerfile-go-deps
alpeb: Yeah the `actions/cache` space is very limited (5GB) so I'm reducing the base layer size by a few hundred megs doing this.

alpeb:
Thanks @aliariff @cpretzer @kleimkuhler for the reviews! :taco: 
I'll push an update in a few.

> I think the checks need to be updated in GitHub settings to remove `Docker build` since that check is named/handled differently with these improvements

Yes, I'll ask Adam help with that before getting this merged :+1: 

> What are your thoughts on maybe combining these jobs so that we don't need to go through the whole process of starting up a new VM and downloading the artifacts which can take ~2 mins. We should be able to push right from `docker_build` save a few mins on the process.

That definitely makes sense!

Dockerfile-go-deps
kleimkuhler: Sounds good

alpeb:
Ok the next set of changes are in:

- In `cloud_integration.yml`, have `docker_build` do the docker pushing as well, removing the `docker_push` job.
- In `release.yml` the `docker_push` job was removed as well, but we still have to upload the CLI windows binary into the build artifacts because the `windows_static_cli_tests` job requires it and it can't use the docker image.
- In `release.yml` the `kind_integration_tests` were simplified as well to use the pushed image instead of the artifacts.
- Separation of the cache pruning into a separate step
- Other minor simplifications

.github/workflows/cloud_integration.yml
```diff
@@ -8,72 +8,62 @@ on:
...
-    - name: Checkout code
-      # actions/checkout@v2
-      uses: actions/checkout@722adc6
-    - name: Set environment variables from scripts
-      run: |
-        . bin/_tag.sh
-        echo ::set-env name=TAG::$(CI_FORCE_CLEAN=1 bin/root-tag)
+        docker buildx create --driver docker-container --use
+        bin/docker-build-${{ matrix.target }}
     - name: Configure gcloud
```
kleimkuhler, line 45: I don't think we want `Configure gcloud` as part of the matrix that this job is.

I think this will lead to each target of the matrix having a `Configure gcloud` step. Ideally, the `Push docker images to registry` remains part of this job, but then `Configure cloud` is a separate job that is dependent on `Docker build`.

aliariff: @kleimkuhler 

But how we can push the image if we don't run `Configure gcloud` beforehand?

kleimkuhler: @aliariff Yep I was thinking this step was responsible for cluster creating as well which is not the case, so this should be fine!

.github/workflows/release.yml
```diff
@@ -14,84 +10,75 @@ on:
 env:
   GH_ANNOTATION: true
```
aliariff, line 11: missing `DOCKER_BUILDKIT: 1` here

.github/workflows/release.yml
```diff
@@ -139,41 +125,21 @@ jobs:
...
+        CMD="$PWD/target/release/linkerd2-cli-$TAG-linux"
+        echo "::set-env name=CMD::$CMD"
+        echo "::set-output name=tag::$TAG"
+    - name: Run integration tests
       run: |
-        # Copy the CLI out of the local cli-bin container.
-        container_id=$(docker create "$DOCKER_REGISTRY/cli-bin:$TAG")
-        docker cp $container_id:/out/linkerd-linux $HOME/.linkerd
-
+        bin/docker-pull-binaries $TAG
```
aliariff, line 134: this `$TAG` is not set

.github/workflows/release.yml
alpeb: Good catch :+1:

cpretzer:
Nice work @alpeb and great reviews @kleimkuhler @aliariff 

Do we need to be strategic about whether we merge this or #4773 first?

cpretzer:
@alpeb thanks for the quick turnaround on getting #4773 merged into this. That will help @aliariff move to the next phase of  his GSoC project.

Do we all agree that removing the `Docker build` required check is the correct way to address the check that hasn't completed? I think that it was replaced by the `Kind Integration/Docker build` checks.

alpeb:
> Do we all agree that removing the `Docker build` required check is the correct way to address the check that hasn't completed? I think that it was replaced by the `Kind Integration/Docker build` checks.

Yes, I'll have that removed, and after a few weeks till the PRs start getting this change, we'll add the newly required checks :+1:

PR #4766, opened 2020-07-17
Add changes for edge-20.7.3

kleimkuhler:
## edge-20.7.3

This edge release introduces an install flag for EndpointSlices. With this flag,
endpoint slices can be used as a resource in the destination service instead of
the endpoints resource.

* Introduce CLI and Helm install flag for EndpointSlices (thanks @Matei207!)
* Internal improvements to the CI process for testing Helm installations

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Changes
 
+## edge-20.7.3
+
+This edge release introduces an install flag for EndpointSlices. With this flag,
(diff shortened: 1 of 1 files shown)


alpeb:
@kleimkuhler The `--enable-endpoint-slices` was introduced but it's not yet operational. If you use it, `linkerd install --enable-endpoint-slices` throws an error. So maybe you can rephrase that as preliminary work was done to support EndpointSlices.

PR #4769, opened 2020-07-20
Fix build status badge in README

siggy:
The GitHub Actions build status badge was referencing an old workflow
named `CI`, which always shows red, and is no longer used.

Fix the build status badge to reference the `Release` workload.

Also slightly reformat the matrix build yaml, as the list has grown a
bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ specific language governing permissions and limitations under the License.
 
 <!-- refs -->
 [github-actions]: https://github.com/linkerd/linkerd2/actions
-[github-actions-badge]: https://github.com/linkerd/linkerd2/workflows/Release/badge.svg
+[github-actions-badge]: https://github.com/linkerd/linkerd2/workflows/Cloud%20integration/badge.svg
 [cncf]: https://www.cncf.io/
 [coc]: https://github.com/linkerd/linkerd/wiki/Linkerd-code-of-conduct
 [getting-started]: https://linkerd.io/2/getting-started/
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ specific language governing permissions and limitations under the License.
 
 <!-- refs -->
 [github-actions]: https://github.com/linkerd/linkerd2/actions
-[github-actions-badge]: https://github.com/linkerd/linkerd2/workflows/CI/badge.svg
+[github-actions-badge]: https://github.com/linkerd/linkerd2/workflows/Release/badge.svg
```
kleimkuhler, line 97: How about referencing Cloud Integration workflow? That runs on all merges to main.

Release will only run on commits that are tagged. There is a possibility that an untagged commit breaks Cloud Integration, but it would still show as succeeding because it is looking at some previous Release workflow.

siggy: `Cloud Integration` seems like a more appropriate workflow to show a badge for. Part of the reason for selecting `Release` was because the name of the workflow appears in the badge. Just confirming we're ok with a badge that says `Cloud Integration`:
Screen Shot 2020-07-20 at 4 48 51 PM

kleimkuhler: Ah yep that is a good thing to note. I'm okay with that though. I think having a `Cloud Integration` label that is reflective of the current commit in main is better than `Release` that may reflect something older. Thanks for pointing that out!

PR #4771, opened 2020-07-21
service topologies: change format of context token

Matei207:
**PoC**: #4695 

### What
---

* Implemented this feature according to the feedback received in the last PR (proof of concept).
* This PR changes the structure of the context token used by the proxy when doing service discovery from `ns:<namespace>` to JSON format.
* Additionally, this PR also adds the `nodeName` of the pod as a field in the context token JSON, to be used by the destination service to infer the source of traffic when taking the topology-preference of services into account.

### How
---
* Change `_proxy.tpl` partial:
  - to use the downward API to get the `nodeName` from the spec of the pod (after it has been scheduled);
  - to have a JSON format value for `LINKERD2_PROXY_DESTINATION_CONTEXT` environment variable (by using the special character `|`).

* Change tests to reflect new format:
  - updated golden files;
  - json patch files for the `proxy-injector`.

* Most of the code changes have been done in the `destination/server.go` file to support the new format of the contextual token, as well as to ensure there is backwards compatibility (@adleong pointed out how necessary this is in the PoC PR). A couple of points here:
  - Added a new function `parseContextToken` to parse the json-formatted token as a string and unmarshal it into a `contextualToken` structure;
  - Added new tests that test `GetProfile` works as expected with the new format (old format is tested by old tests);
  - Added new tests that test `parseContextToken` behaves correctly: unmarshals valid JSON, returns namespace name if token uses old format, returns an empty structure when JSON is invalid and old format is not used.

* The new context token has a structure that (in my opinion) makes it easily extendable by making the necessary changes in the `struct` on the destination service side, as well as in the proxy template partial. Its originally desired opacity should still be enforced -- the proxy has no notion of what is inside this token, only that its type is a string, as such, it gets read in the proxy as a string (from the env variable), gets passed over the wire as a string and is only deserialized in its structure when it reaches the destination service (where its type & functionality is known and used to provide service discovery).

**Note**: *I didn't do any service topology work on the `destination/server.go` because it's out of the "context" of this PR; the nodeName part of the token will be used in subsequent PRs, this was just to change the format and add backwards compatibility in the destination svc*.

**Feedback**: would love some feedback on:
* Stylistic quality of the code;
* Format of the token (incl. how it looks in the logs/over-the-wire, there are a lot of backslashes!);
* Backwards compatibility part -- more specifically error handling and how we use the old format in case JSON is invalid.

**Tests**:

Apart from automated tests, I manually tested this by:
*  building, installing the new control plane in minikube and first checking if the environment variable in a proxy container is set to the new context token format:
```sh
$ k exec linkerd-destination-dd44b4cc5-dm9pc -n linkerd -it -c linkerd-proxy -- bin/bash -c "env | grep LINKERD2_PROXY_DESTINATION_CONTEXT" LINKERD2_PROXY_DESTINATION_CONTEXT={"ns":"linkerd", "nodeName:"minikube"}
```
* analysing the logs from the destination pod to make sure context token has correct format:
```sh
time="2020-07-21T17:58:46Z" level=debug msg="GetProfile(path:\"linkerd-identity.linkerd.svc.cluster.local:8080\"  context_token:\"{\\\"ns\\\":\\\"linkerd\\\", \\\"nodeName\\\":\\\"minikube\\\"}\\n\")" addr=":8086" component=server remote="127.0.0.1:41260"
```

**I still need to test**:
* creating `ServiceProfiles` and testing out the namespace can be used in its new format
* upgrade the control plane from an edge/stable version to the new version (which should show how backwards compatibility is preserved)
will update PR message after these tests

### DIFF
--- a/controller/api/destination/server.go
+++ b/controller/api/destination/server.go
@@ -1,6 +1,7 @@
 package destination
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"strconv"
@@ -225,7 +226,9 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 	// up to the fallbackProfileListener to merge updates from the primary and
 	// secondary listeners and send the appropriate updates to the stream.
 	if dest.GetContextToken() != "" {
-		profile, err := profileID(path, dest.GetContextToken(), s.clusterDomain)
+		ctxToken := s.parseContextToken(dest.GetContextToken())
+
+		profile, err := profileID(path, ctxToken, s.clusterDomain)
 		if err != nil {
 			log.Debugf("Invalid service %s", path)
 			return status.Errorf(codes.InvalidArgument, "invalid profile ID: %s", err)
@@ -239,7 +242,7 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 		defer s.profiles.Unsubscribe(profile, primary)
 	}
 
-	profile, err := profileID(path, "", s.clusterDomain)
+	profile, err := profileID(path, contextToken{}, s.clusterDomain)
 	if err != nil {
 		log.Debugf("Invalid service %s", path)
 		return status.Errorf(codes.InvalidArgument, "invalid profile ID: %s", err)
@@ -264,17 +267,29 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 /// util ///
 ////////////
 
-func nsFromToken(token string) string {
-	// ns:<namespace>
-	parts := strings.Split(token, ":")
-	if len(parts) == 2 && parts[0] == "ns" {
-		return parts[1]
-	}
-
-	return ""
+type contextToken struct {
+	Ns       string `json:"ns,omitempty"`
+	NodeName string `json:"nodeName,omitempty"`
 }
 
-func profileID(authority string, contextToken string, clusterDomain string) (watcher.ProfileID, error) {
+func (s *server) parseContextToken(token string) contextToken {
+	ctxToken := contextToken{}
+	if err := json.Unmarshal([]byte(token), &ctxToken); err != nil {
+		// if json is invalid, means token can have ns:<namespace> form
+		parts := strings.Split(token, ":")
+		if len(parts) == 2 && parts[0] == "ns" {
+			s.log.Warnf("context token %s using old token format", token)
+			ctxToken = contextToken{
+				Ns: parts[1],
+			}
+		} else {
+			s.log.Errorf("context token %s is invalid: %s", token, err)
+		}
+	}
+	return ctxToken
+}
+
+func profileID(authority string, ctxToken contextToken, clusterDomain string) (watcher.ProfileID, error) {
 	host, _, err := getHostAndPort(authority)
 	if err != nil {
 		return watcher.ProfileID{}, fmt.Errorf("invalid authority: %s", err)
@@ -287,8 +302,8 @@ func profileID(authority string, contextToken string, clusterDomain string) (wat
 		Name:      fmt.Sprintf("%s.%s.svc.%s", service.Name, service.Namespace, clusterDomain),
 		Namespace: service.Namespace,
 	}
-	if contextNs := nsFromToken(contextToken); contextNs != "" {
-		id.Namespace = contextNs
+	if ctxToken.Ns != "" {
+		id.Namespace = ctxToken.Ns
 	}
 	return id, nil
 }
--- a/controller/api/destination/server_test.go
+++ b/controller/api/destination/server_test.go
@@ -243,6 +243,84 @@ func TestGetProfiles(t *testing.T) {
 		}
 	})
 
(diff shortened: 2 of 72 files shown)


Matei207:
Integration tests are failing because of the `upgrade` stage, which is very similar to the manual stage that I have mentioned in the testing phase of the PR description. One suggestion would be to add these messages to known events or alternatively, re-think the error handling. I'm not sure which is better to do this so as to preserve the quality (& integrity) of tests and provide helpful logging.

controller/api/destination/server.go
```diff
@@ -264,17 +270,26 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
+
+func parseContextToken(token string) (contextToken, error) {
+	ctxToken := contextToken{}
+	if err := json.Unmarshal([]byte(token), &ctxToken); err != nil {
+		// if json is invalid, means token can have ns:<namespace> form
+		parts := strings.Split(token, ":")
+		if len(parts) == 2 && parts[0] == "ns" {
+			ctxToken.Ns = parts[1]
+		}
+		return ctxToken, err
```
adleong, line 286: I think we want to return nil for the second argument here.  We failed to parse as json, but successfully parsed in the old format.

adleong: In fact, under that framing, this method never returns an error and can return just one return argument.

controller/api/destination/server.go
```diff
@@ -225,7 +226,12 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 	// up to the fallbackProfileListener to merge updates from the primary and
 	// secondary listeners and send the appropriate updates to the stream.
 	if dest.GetContextToken() != "" {
-		profile, err := profileID(path, dest.GetContextToken(), s.clusterDomain)
+		ctxToken, err := parseContextToken(dest.GetContextToken())
+		if err != nil {
+			log.Errorf("error parsing context token %s as json:%s", dest.GetContextToken(), err)
```
adleong, line 231: Ah, I see: you're returning the error in order to log that we have received an old format token.  I think this is confusing because the `parseContextToken` method is actually successful, so I don't think it should return an error in this case.  Instead, I would have the `parseContextToken` method itself log when it gets a token in the old format.  I'd also probably just make that log at the warning level since getting old format tokens doesn't cause any problems (but users will want to upgrade to the new format eventually).

Matei207: @adleong this is a fair point, I've read the previous comments as well. I have made some changes to log both when the JSON/context token is invalid (as an error) and when a warning when the format of the token is old. I wanted to avoid checking if the JSON is valid since it's a pretty expensive operation, but that's also an alternative to make it more cleaner if the current way is not very straightforward. I'm thinking the alternative would be:
```go
func parseContextToken(token string) {
  ctxToken := &contextToken
  if !isValidJSON(token) {
      // process old format
   }

   // unamarshal

  return ctxToken
}
```

Matei207:
**Update:** I had to change the integration test data related to manual injection and also add the warn log to known messages so the integration test does not fail. Adding this here so they can be reviewed as they're not modified by the `--update` flag in `go test`.

controller/api/destination/server.go
```diff
@@ -264,17 +267,27 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
...
 
-	return ""
+func (s *server) parseContextToken(token string) contextToken {
+	ctxToken := contextToken{}
+	if err := json.Unmarshal([]byte(token), &ctxToken); err != nil {
+		// if json is invalid, means token can have ns:<namespace> form
+		parts := strings.Split(token, ":")
+		if len(parts) == 2 && parts[0] == "ns" {
+			s.log.Warnf("context token %s using old token format", token)
+			ctxToken.Ns = parts[1]
```
adleong, line 282: This is probably fine but I don't like relying on the assumption that when `Unmarshal` returns `err` that it won't have written anything to `ctxToken`.  i.e. I don't like assuming that `ctxToken.NodeName` is still the empty value.  This is probably paranoia, but to be super safe I prefer:

```
ctxToken = contextToken {
    Ns: parts[1],
}
```

controller/api/destination/server_test.go
```diff
@@ -186,6 +186,7 @@ func TestGet(t *testing.T) {
 	})
 }
 
+//TODO: write test where json is fucked and where we won't get a ns token if we split up the string
```
kleimkuhler, line 189: May want to clean this comment up a bit 😬

Matei207: That was not supposed to pushed, oops. 😳

controller/api/destination/server.go
Matei207: @adleong thank you, that makes a lot of sense, I've made all necessary changes.

I prefer a more defensive programming style, eliminates some of the work that might have to be done in the future to fix issues that slip through the cracks. The suggestion is very appreciated!

Matei207:
> This looks good! I tested with service profiles as well, and both the response and destination logs had no errors.

Thank you for the review and for testing it out, it is very appreciated!

PR #4773, opened 2020-07-21
CI: Remove need of `bin/update-go-deps-shas`

alpeb:
This is a followup to #4765

Got rid of the `go-deps` image and instead added something similar on top of all the Dockerfiles dealing with `go`, as a first stage for those Dockerfiles. That continues to serve as a way to pre-populate go's build cache, which speeds up the builds in the subsequent stages. That build should in theory be rebuilt automatically only when `go.mod` or `go.sum` change, and now we don't require running `bin/update-go-deps-shas`. That script was removed along with all the logic elsewhere that used it, including the `go_dependencies` job in the `static_checks.yml` github workflow.

The list of modules preinstalled was moved from `Dockerfile-go-deps` to a new script `bin/install-deps`. I couldn't find a way to generate that list dynamically, so whenever a slow-to-compile dependency is found, we have to make sure it's included in that list.

Although this simplifies the dev workflow, note that the real motivation behind this was a limitation in buildx's `docker-container` driver that forbids us from depending on images that haven't been pushed to a registry, so we have to resort to building the dependencies as a first stage in the Dockerfiles. See the final section at #4765 for more details.


bin/docker-push-base
```diff
@@ -6,8 +6,5 @@ bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 
 # shellcheck source=_docker.sh
 . "$bindir"/_docker.sh
-# shellcheck source=_tag.sh
```
aliariff, line 9: need to change some script in github workflow which still call `bin/docker-push-deps`
ex: https://github.com/linkerd/linkerd2/blob/main/.github/workflows/release.yml#L91

aliariff: oh nvm, you try to merge to `alpeb/buildx` I though to `main`.

PR #4782, opened 2020-07-22
CI: Remove Base image

aliariff:
@cpretzer @alpeb 

Followup to #4765 

- Removing the dependency from `base` image. Same reason with #4773 
- Cleanup some leftover/obsolete documentation in `BUILD.md`


Dockerfile-debug
```diff
@@ -1,9 +1,20 @@
...
+    nghttp2 \
     tcpdump \
     iproute2 \
     lsof \
     conntrack \
     tshark && \
     rm -rf /var/lib/apt/lists/*
+
+# Switching base image from stretch to buster, we need to use iptables-legacy
+# if no changes to how iptables is invoked are done.
```
alpeb, line 16: Can you please remove this comment? No need to explain the details about the stretch-to-buster upgrade, plus the second sentence doesn't make sense. It can be replaced with something like  "We still rely on old iptables-legacy syntax"
(Same thing for `cni-plugin/Dockerfile`)

cni-plugin/Dockerfile
```diff
@@ -14,9 +14,17 @@ COPY controller controller
 COPY cni-plugin cni-plugin
 RUN CGO_ENABLED=0 GOOS=linux go build -o /go/bin/linkerd-cni -v -mod=readonly ./cni-plugin/
 
-FROM gcr.io/linkerd-io/base:2020-06-08.01
+FROM debian:buster-20200514-slim
 WORKDIR /linkerd
-RUN curl -kL -o $(which jq) https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    iptables \
+    jq && \
```
kleimkuhler, line 21: Do we not need `curl`, `dnsutils`, `iptables`, `nghttp2` here?

aliariff: from @alpeb, `cni-plugin` only need `iptables`

kleimkuhler: 👍

cpretzer:
@alpeb is `Go dependencies` another check that we can remove?

alpeb:
> @alpeb is `Go dependencies` another check that we can remove?

Yep, I did ask Adam :-)

cpretzer:
Awesome, TY!

cni-plugin/Dockerfile
alpeb: Those are only debugging tools that were provided to the Debug container for convenience, but anyways let me give CNI a test to make sure everything is working as expected...

alpeb: I confirmed CNI is working as expected :+1:

kleimkuhler: @alpeb Thanks 🙂

PR #4783, opened 2020-07-23
Helm chart - linkerd2-collector : enable jaeger receiver

### LINKED ISSUE #4778: Enable jaeger receiver in collector configuration in helm chart

I want to use the jaeger receiver of OpenCensus Collector but it is not enabled when installing linkerd with the Helm chart. Only opencensus & zipkin are enabled

Just enable the receiver in the configuration and declare the port in service.

```
receivers:
      jaeger:
        jaeger-thrift-http-port: 14268
```

olivierboudet:
This PR just enable the jaeger receiver in the configmap of linkerd2-collector.

Questions : 
- As the zipkin and opencensus receiver are always enabled, I just enabled also the jaeger receiver. Perhaps this must be an opt-in ?
- I changed the chart version of partials and tracing add-ons, is it the correct way or the version is set by CI as it is for linkerd chart version ?

Fixes #4778

Signed-off-by: Olivier Boudet <o.boudet@gmail.com>

### DIFF
--- a/charts/linkerd2/requirements.lock
+++ b/charts/linkerd2/requirements.lock
@@ -12,4 +12,4 @@ dependencies:
   repository: file://../add-ons/tracing
   version: 0.1.0
 digest: sha256:d2428770ae7d5134c5af6521c78a4c5f95da4c75f21bdea0f74fad6ab6e2e044
-generated: "2020-06-24T11:07:53.924602129Z"
+generated: "2020-07-23T14:47:05.936935407+02:00"
--- a/charts/add-ons/tracing/templates/tracing.yaml
+++ b/charts/add-ons/tracing/templates/tracing.yaml
@@ -20,6 +20,8 @@ data:
         port: 55678
       zipkin:
         port: 9411
+      jaeger:
+        jaeger-thrift-http-port: 14268
     queued-exporters:
       jaeger-all-in-one:
         num-workers: 4
@@ -51,6 +53,10 @@ spec:
     port: 9411
     protocol: TCP
     targetPort: 9411
+  - name: jaeger
+    port: 14268
+    protocol: TCP
+    targetPort: 14268
   selector:
     {{.Values.global.controllerComponentLabel}}: {{.Values.collector.name}}
 ---
@@ -110,6 +116,7 @@ spec:
         ports:
         - containerPort: 55678
         - containerPort: 9411
+        - containerPort: 14268
         readinessProbe:
           httpGet:
             path: /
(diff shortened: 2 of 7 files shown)


olivierboudet:
> I don't think we need to change versions here, as the parent level linkerd version is the one we rely on and we just bundle sub-charts with them, without maintaining any chart versions here.

ok thanks, I have reverted the version change.

ihcsim:
@olivierboudet I am just wondering instead of updating the Linkerd chart every time a new receiver is needed, can we just provide documentation on how to use tools like Kustomize to do this? Looking at the changes to `charts/add-ons/tracing/templates/tracing.yaml`, this should be easy to accomplish with Kustomize. LMKWYT.

olivierboudet:
> @olivierboudet I am just wondering instead of updating the Linkerd chart every time a new receiver is needed, can we just provide documentation on how to use tools like Kustomize to do this? Looking at the changes to `charts/add-ons/tracing/templates/tracing.yaml`, this should be easy to accomplish with Kustomize. LMKWYT.

I agree that it is not ideal to update the chart each time, especially that not all users will need all configured receivers.
But instead of asking users to use a new tool (I personnally does not use Kustomize at the moment), I would prefer to have a configurable chart.

Perhaps with something like that :
```
tracing:
  collector:
    extraReceivers:
    - jaeger:
        jaeger-thrift-http-port: 14268
    extraPorts:
      - name: jaeger
        port: 14268
        protocol: TCP
        targetPort: 14268
```

ihcsim:
@olivierboudet Or just turn the entire `data.linkerd-collector-config` into a Helm variable? @Pothulapati WDYT?

olivierboudet:
Actually I proposed an "extra" variable to avoid that a user override a piece of configuration needed by linkerd itself, especially after a chart upgrade which add a line in configuration.

ihcsim:
@Pothulapati Thoughts?

ihcsim:
@olivierboudet Mind fixing up the conflicting files? Then we should be able to merge this. Thanks.

olivierboudet:
> @olivierboudet Mind fixing up the conflicting files? Then we should be able to merge this. Thanks.

Thanks, I have fixed the conflicts.

Pothulapati (approved):
LGTM!

charts/linkerd2/requirements.lock
```diff
@@ -12,4 +12,4 @@ dependencies:
   repository: file://../add-ons/tracing
   version: 0.1.0
 digest: sha256:d2428770ae7d5134c5af6521c78a4c5f95da4c75f21bdea0f74fad6ab6e2e044
-generated: "2020-06-24T11:07:53.924602129Z"
+generated: "2020-07-23T14:47:05.936935407+02:00"
```
Pothulapati, line 15: As `requirements.yaml` did not change in this PR, wonder if this is relevant/needed? :thinking:

PR #4785, opened 2020-07-23
edge-20.7.4

zaharidichev:
## edge-20.7.4

This edge release adds support for the new Kubernetes [EndpointSlice] resource
to the Destination controller. Using the EndpointSlice API is more efficient
for the Kubernetes control plane than using the Endpoints API. If the cluster
supports EndpointSlices (a beta feature in Kubernetes 1.17), Linkerd can be
installed with `--enable-endpoint-slices` flag to use this resource rather
than the Endpoints API.

* Added fish shell completions to the `linkerd` command (thanks @WLun001!)
* Enabled the support for EndpointSlices (thanks @Matei207!)
* Separated prometheus checks and made them runnable only when the add-on
  is enabled

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Changes
 
+## edge-20.7.4
+
+This edge release brings support for EndpointSlices.
+
+* Added fish shell completions to the `linkerd` command (thanks @WLun001!)
+* Enabled the support for EndpointSlices (thanks @Matei207!)
+* Separated prometheus checks and made them runnable only when there the add-on
+  is enabled
+
 ## edge-20.7.3
 
 * Add preliminary support for EndpointSlices which will be usable in future
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,12 @@
 
 ## edge-20.7.4
 
-This edge release brings support for EndpointSlices.
+This edge release adds support for the new Kubernetes [EndpointSlice] resource
+to the Destination controller. Using the EndpointSlice API is more efficient
+for the Kubernetes control plane than using the Endpoints API. If the cluster
+supports EndpointSlices (a beta feature in Kubernetes 1.17), Linkerd can be
+installed with `--enable-endpoint-slices` flag to use this resource rather
+than the Endpoints API.
 
 * Added fish shell completions to the `linkerd` command (thanks @WLun001!)
 * Enabled the support for EndpointSlices (thanks @Matei207!)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,16 +2,17 @@
 
 ## edge-20.7.4
 
-This edge release adds support for the new Kubernetes [EndpointSlice] resource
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,5 +1,14 @@
 # Changes
 
+## edge-20.7.4
+
+This edge release brings support for EndpointSlices.
+
+* Added fish shell completions to the `linkerd` command (thanks @WLun001!)
+* Enabled the support for EndpointSlices (thanks @Matei207!)
+* Separated prometheus checks and made them runnable only when there the add-on
```
admc, line 9: I think we have an extra "there" in here:

"Separated prometheus checks and made them runnable only when the add-on is enabled"

CHANGES.md
```diff
@@ -1,5 +1,14 @@
 # Changes
 
+## edge-20.7.4
+
+This edge release brings support for EndpointSlices.
```
hawkw, line 5: It might be nice to have a little more discussion of what endpoint slices are, what this change means for the user, etc. Maybe:
```suggestion
This edge release adds support for the new Kubernetes [EndpointSlice] resource to
the Destination controller. Using the EndpointSlice API is more efficient for the 
Kubernetes control plane than using the Endpoints API. If the cluster supports 
EndpointSlices (a beta feature in Kubernetes 1.17), Linkerd can be installed with 
```console
linkerd install --enable-endpoint-slices
```
to use this resource rather than the Endpoints API.

[EndpointSlice](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)
```

File: CHANGES.md [hunk 3]
```diff
@@ -1,5 +1,19 @@
 # Changes
 
+## edge-20.7.4
+
+This edge release adds support for the new Kubernetes [EndpointSlice] resource
```
hawkw, line 5: Looks like this link needs a reference. Let's add

```markdown
[EndpointSlice](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)
```
to the end of the changelog entry.

I tried to include the link reference in my suggested changes https://github.com/linkerd/linkerd2/pull/4785#discussion_r459646336 but it looks like GitHub screwed it up, because I also used a fenced code block _within_ my suggested code, so it ended the suggestion block too early. :/

CHANGES.md
```diff
@@ -1,5 +1,19 @@
...
+to the Destination controller. Using the EndpointSlice API is more efficient
+for the Kubernetes control plane than using the Endpoints API. If the cluster
+supports EndpointSlices (a beta feature in Kubernetes 1.17), Linkerd can be
+installed with `--enable-endpoint-slices` flag to use this resource rather
+than the Endpoints API.
+
+* Added fish shell completions to the `linkerd` command (thanks @WLun001!)
+* Enabled the support for EndpointSlices (thanks @Matei207!)
+* Separated prometheus checks and made them runnable only when the add-on
```
hawkw, line 14: nit: i think this should be 
```suggestion
* Separated Prometheus checks and made them runnable only when the add-on
```

Issue #4612, opened 2020-06-16
Support minimal installation footprint


olix0r:
For the Linkerd data plane to function correctly, it only needs (1) the proxy injection admission controller, (2) the identity service, and (3) the destination service.  Other components like the Web UI, Tap, Grafana, Prometheus, etc are purely optional.

I propose the following:

1. These optional components be moved into a "linkerd-viz" extension (analogous to the linkerd-multicluster extension). CRDs, API Extensions, RBAC, etc, that are only required by these components should only be installed when this extension is installed.
2. `linkerd install` should install this extension by default, but should provide a flag to opt-out of installing it. Ideally `linkerd viz install` can be used to install this component post-hoc.
3. CLI commands that interact with services in this extension (tap, stat, etc) should fail gracefully when the extension is not installed.

### Questions:

* How to upgrade from 2.8?

Pothulapati:
On top of my mind, I see that the following changes are required from the exisiting model to make this possible.
- Obviously, The first requirement would be to separate web, tap, Prometheus into individual add-ons.
- Support Individual Install/Uninstall of add-ons. Currently, The add-ons go hand in hand with control-plane install but we want a model like the current multi-cluster model where you can install and uninstall it without effecting the control-plane components. This looks possible with the current model as all these add-ons are sub-charts currently, and if we can populate all the existing `Values` struct but leave the control-plane chart while rendering could make this possible.
- To provide packages like `linkerd-viz`, etc we can have a `viz-values.yaml` which just configures the required components and disables others, and this file can be used by the CLI if the viz component is to be installed. The same can also be used with Helm.
- Currently `linkerd-config-addons` is where all the add-on config goes, and is used by the check command. We can have a similar thing with stat, tap, etc where we first check if the relevant component is installed and try running those commands only then.

The above are the assumptions/thinking based on the current add-on model. But if we need a more better way to do the above stuff, I'm happy to discuss and consider them. :)

alpeb:
I agree we should continue with the add-ons model to achieve this kind of modularity. Currently that approach is not very user-friendly as it requires writing yaml to be passed to the `--addon-config` flag at install/upgrade time. But having a separate subcommand to install these might allow us to surface those config items as separate CLI flags as well, no longer having the `linkerd install` flag explosion fear.

As for upgrading from 2.8, I only see these flag values from `linkerd install` that would get moved into the viz add-ons:
```
- controller-log-level
- controller-replicas
- controller-uid
- prometheus-image
[...]

PR #4793, opened 2020-07-24
move linkerd-web as a add-on

Pothulapati:
This PR moves linkerd-web into a separate chart called dashboard and
thus making it optional. This follows the same approach as that of
grafana and prometheus.


cli/cmd/install.go
```diff
@@ -791,7 +789,6 @@ func (options *installOptions) buildValuesWithoutIdentity(configs *pb.All) (*l5d
 	installValues.HeartbeatSchedule = options.heartbeatSchedule()
 	installValues.RestrictDashboardPrivileges = options.restrictDashboardPrivileges
 	installValues.DisableHeartBeat = options.disableHeartbeat
-	installValues.WebImage = fmt.Sprintf("%s/web", options.dockerRegistry)
```
Pothulapati, line 794: This image name can be overriden by using `--addon-config` and through `values.yaml` with Helm. but when a user applies the `--registry` flag, Is the expectation that the image registry value is overriden even for add-ons? We can make sure its propagated to add-ons if thats the case. @alpeb

charts/add-ons/dashboard/templates/web.yaml
```diff
@@ -68,12 +68,10 @@ spec:
...
         - -grafana-addr={{.Values.global.grafanaUrl}}
-        {{- else if .Values.grafana.enabled }}
-        - -grafana-addr={{.Values.grafana.name}}.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:3000
-        {{- end}}
-        {{- if .Values.tracing.enabled }}
-        - -jaeger-addr={{.Values.tracing.jaeger.name}}.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:16686
+        {{- else}}
+        - -grafana-addr=linkerd-grafana.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:3000
         {{- end}}
+        - -jaeger-addr=linkerd-jaeger.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:16686
```
Pothulapati, line 74: but as we are removing the condition and passing that variable all the time, This means that users will see the `jaeger` icon irrespective of the add-on being installed which is not great. wondering if there's a better way to do this. :thinking: 

@alpeb

cli/cmd/install.go
alpeb: Yeah I guess that is the expectation, for `--registry` to apply to all the images, including add-ons.

charts/add-ons/dashboard/templates/web.yaml
alpeb: This stems from the inability of add-ons to access each other's Helm configs, right?
The web component currently mounts `linkerd-config`; we could have it mount `linkerd-config-addons` as well, and override the grafana/jaeger flags based off of that.

Pothulapati: @alpeb Great Suggestion!
Updated the PR regarding the same!

web/main.go
```diff
@@ -57,6 +59,33 @@ func main() {
...
 
+	values, err := config.AddOns(pkgK8s.MountPathAddOnsConfig)
+	if err != nil {
+		log.Warnf("failed to load values config: [%s]", err)
+	}
+
+	addOns, err := l5dcharts.ParseAddOnValues(values)
+	if err != nil {
+		log.Warnf("failed to parse add-on values: [%s]", err)
+	}
```
alpeb, line 70: How do you feel about moving this logic into `config.Addons()`?

web/main.go
```diff
@@ -57,6 +59,33 @@ func main() {
...
+
+	for _, addOn := range addOns {
+		if addOn.Name() == l5dcharts.GrafanaAddOn && *grafanaAddr == "" {
+			*grafanaAddr = fmt.Sprintf("linkerd-grafana.%s.svc.%s:3000", *controllerNamespace, clusterDomain)
+		}
+
+		if addOn.Name() == l5dcharts.TracingAddOn && *jaegerAddr == "" {
+			*jaegerAddr = fmt.Sprintf("linkerd-jaeger.%s.svc.%s:16686", *controllerNamespace, clusterDomain)
```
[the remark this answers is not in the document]
Pothulapati: `l5dcharts.ParseAddOnValues()` actually gives back a list of **enabled** add-ons. So, No need to check if they are enabled again.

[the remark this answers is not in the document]
alpeb: :+1:

Pothulapati:
@alpeb Yep, That was my plan!

Pothulapati:
Minimal Control Plane effort is being worked as a separate effort (with an RFC) along with a slightly different approach. So, Closing this! See #4612

PR #4794, opened 2020-07-24
Build ARM docker images

aliariff:
@cpretzer @alpeb @kleimkuhler 
These changes enable us to build ARM images.

# Changes:
- Add a new env key `DOCKER_MULTIARCH` and `DOCKER_PUSH`. When set, it will build multi-arch images and push them to the registry. See https://github.com/docker/buildx/issues/59 for why it must be pushed to the registry.
- Usage of `crazy-max/ghaction-docker-buildx ` is necessary as it already configured with the ability to perform cross-compilation (using QEMU) so we can just use it, instead of manually set up it.
- Usage of `buildx` now make default global arguments. (See: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope)

# Follow-up:
- Releasing the CLI binary file in ARM architecture. The docker images resulting from these changes already build in the ARM arch. Still, we need to make another adjustment like how to retrieve those binaries and to name it correctly as part of Github Release artifacts.


aliariff:
@alpeb 
So far the resulting images that come from this, is working fine in RPi cluster (without putting `GOARCH ` in `bin/install-deps `). Is this because maybe the different behavior of `go install` and `go build`?

cpretzer:
Tested the Dockerfile changes and the images are building for me locally. I chatted with @aliariff and the [build portions](https://github.com/aliariff/linkerd2/pull/6/checks?check_run_id=907757945) of the GitHub actions succeed.

alpeb:
I had added some inline comments to the changeset, but I've just deleted them since all that will change given my previous text wall :grimacing:

aliariff:
@alpeb 
I think your approach is the other way around from what I am doing here. But the basic idea of cross-compilation is the same.
I utilize the `buildx build --platform ....` the reason I use `FROM --platform=$BUILDPLATFORM ` in the Dockerfile is for reusing the image so that we can perform cross-compilation. in this case, the image for the `go-deps` and `fetch` will only the `linux/amd64` version. So from that same image `linux/amd64` it will perform cross-compilation to another arch.
And for the runtime image, it will be based on the target platform, in this case, we specify when calling `buildx build --platform linux/amd64,linux/arm64,....`
Because `By default, the target platform of the build request is used` ([ref](https://docs.docker.com/engine/reference/builder/#from)).
that's why we need to explicitly write `BUILDPLATFORM`.

aliariff:
this is what happens when we dont specify the `BUILDPLATFORM`
image
the go-deps part will try to pull 3 different images for every arch.

but then when we specify it.
image
it changed to only once pull the golang image, and the other arch will reuse that image from cache.

bin/_docker.sh
```diff
@@ -53,11 +59,18 @@ docker_build() {
       if [ -n "$DOCKER_BUILDKIT_CACHE" ]; then
         cache_params="--cache-from type=local,src=${DOCKER_BUILDKIT_CACHE} --cache-to type=local,dest=${DOCKER_BUILDKIT_CACHE},mode=max"
       fi
-      log_debug "  :; docker buildx $rootdir $cache_params --load -t $repo:$tag -f $file $*"
+
+      output_params="--load"
+      if [ -n "$DOCKER_MULTIARCH" ]; then
+        output_params="--push --platform $SUPPORTED_ARCHS"
+      fi
```
alpeb, line 66: This has implicit the need of pushing when building for more than one platform (as required per docker/buildx#59). Can you make this more explicit? Like using `--push` only if `DOCKER_PUSH` is set, and `--platform` if `DOCKER_MULTIARCH` is set. And before that, failing with an error if the latter is set while the former isn't (with a comment linking to the buildx issue explaining why we need to do this).

bin/_docker.sh
```diff
@@ -53,11 +59,18 @@ docker_build() {
...
+
+      log_debug "  :; docker buildx $rootdir $cache_params $output_params -t $repo:$tag -f $file $*"
       # shellcheck disable=SC2086
       docker buildx build "$rootdir" $cache_params \
-          --load \
+          $output_params \
           -t "$repo:$tag" \
+          -t "$repo:main" \
```
alpeb, line 73: Since we'll be doing this for release tags only, you can drop the `main` tag from here.

Dockerfile-proxy
```diff
@@ -1,19 +1,21 @@
...
+ARG BUILDPLATFORM=linux/amd64
 
 # Precompile key slow-to-build dependencies
-FROM golang:1.14.2-alpine as go-deps
+FROM --platform=$BUILDPLATFORM golang:1.14.2-alpine as go-deps
 WORKDIR /linkerd-build
 COPY go.mod go.sum ./
 COPY bin/install-deps bin/
 RUN go mod download
 RUN ./bin/install-deps
```
alpeb, line 10: This stage will be ran separately for each arch and `go install` will use that arch, so this is fine. What do you think though about using `ARG TARGETARCH` and passing that to `install-deps`, so it's clearer what's going on, like you did in the stages below?

aliariff: It is possible, but then it will result in 3 different `go-deps` layers. It will get reused on their arch only, not across the arch. That is my understanding. LMKWYT

alpeb: Isn't that stage already producing 3 different layers? This is an extract of the output I see currently:
```
=> CACHED [linux/amd64 go-deps 2/6] WORKDIR /linkerd-build                                                                                                                                                                               0.0s
 => CACHED [linux/amd64 go-deps 3/6] COPY go.mod go.sum ./                                                                                                                                                                                0.0s
 => CACHED [linux/amd64 go-deps 4/6] COPY bin/install-deps bin/                                                                                                                                                                           0.0s
 => CACHED [linux/amd64 go-deps 5/6] RUN go mod download                                                                                                                                                                                  0.0s 
 => CACHED [linux/amd64 go-deps 6/6] RUN ./bin/install-deps                                                                                                                                                                               0.0s 
 => CACHED [linux/amd64 golang 1/7] WORKDIR /linkerd-build                                                                                                                                                                                0.0s 
 => CACHED [linux/amd64 golang 2/7] COPY pkg/flags pkg/flags                                                                                                                                                                              0.0s 
 => CACHED [linux/amd64 golang 3/7] COPY pkg/tls pkg/tls                                                                                                                                                                                  0.0s 
 => CACHED [linux/amd64 golang 4/7] COPY pkg/version pkg/version                                                                                                                                                                          0.0s 
 => CACHED [linux/amd64 go-deps 2/6] WORKDIR /linkerd-build                                                                                                                                                                               0.0s 
 => CACHED [linux/amd64 go-deps 3/6] COPY go.mod go.sum ./                                                                                                                                                                                0.0s
 => CACHED [linux/amd64 go-deps 4/6] COPY bin/install-deps bin/                                                                                                                                                                           0.0s 
 => CACHED [linux/amd64 go-deps 5/6] RUN go mod download                                                                                                                                                                                  0.0s
 => CACHED [linux/amd64 go-deps 6/6] RUN ./bin/install-deps                                                                                                                                                                               0.0s
 => CACHED [linux/amd64 golang 1/7] WORKDIR /linkerd-build                                                                                                                                                                                0.0s
 => CACHED [linux/amd64 golang 2/7] COPY pkg/flags pkg/flags                                                                                                                                                                              0.0s
 => CACHED [linux/amd64 golang 3/7] COPY pkg/tls pkg/tls                                                                                                                                                                                  0.0s
 => CACHED [linux/amd64 golang 4/7] COPY pkg/version pkg/version                                                                                                                                                                          0.0s
 => [linux/amd64 golang 5/7] RUN CGO_ENABLED=0 GOOS=linux GOARCH=arm go build -mod=readonly ./pkg/...                                                                                                                                     5.0s
 => [linux/amd64 golang 5/7] RUN CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -mod=readonly ./pkg/...                                                                                                                                   4.9s
 => CACHED [linux/amd64 go-deps 2/6] WORKDIR /linkerd-build                                                                                                                                                                               0.0s
 => CACHED [linux/amd64 go-deps 3/6] COPY go.mod go.sum ./                                                                                                                                                                                0.0s
 => CACHED [linux/amd64 go-deps 4/6] COPY bin/install-deps bin/                                                                                                                                                                           0.0s
 => CACHED [linux/amd64 go-deps 5/6] RUN go mod download                                                                                                                                                                                  0.0s
 => CACHED [linux/amd64 go-deps 6/6] RUN ./bin/install-deps
```
`go-deps` is running 3 times, with the same base `linux/amd64`, but I think each run is for each different arch. Which is fine since the ensuing `golang` stage requires a `go` cache built for each one of the archs.

aliariff:
After discussion with @alpeb 
We decided to use `GOARCH` in `go-deps` stage.
The reason is that when we don't pass the arch argument in that stage, in `golang` stage which is later stage based on `go-deps` to perform cross-compilation will only get faster on `amd64` arch, and the other arch compilation takes a long time because the `go-deps` is base on `amd64` only.
And yes, this will result in the cache layers getting bigger because `go-deps` stage now will have three different layers. We will address that issue on separate PR.

aliariff:
Another discussion with @alpeb,
Conclude that we will not use separate cache key for the multi-arch images. Because `GitHub will remove any cache entries that have not been accessed in over 7 days`. We know that most of the use case for the `release.yml` workflow is running weekly when releasing the `edge` version. Therefore there is no point in using separate cache key for multi-arch because, in the end, it is going to be used again in a week, and Github already removes those cache entries.

Ref: https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

.github/workflows/release.yml
```diff
@@ -34,15 +34,25 @@ jobs:
       uses: actions/cache@b820478
       with:
         path: ${{ env.DOCKER_BUILDKIT_CACHE }}
-        key: ${{ runner.os }}-buildx-${{ matrix.target }}-${{ env.TAG }}
+        key: ${{ runner.os }}-buildx-multiarch-${{ matrix.target }}-${{ env.TAG }}
         restore-keys: |
+          ${{ runner.os }}-buildx-multiarch-${{ matrix.target }}-
           ${{ runner.os }}-buildx-${{ matrix.target }}-
```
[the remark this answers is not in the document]
aliariff: This is key already removed. Could you refresh again the PR?

.github/workflows/cloud_integration.yml
```diff
@@ -33,28 +33,25 @@ jobs:
       uses: actions/cache@b820478
       with:
         path: ${{ env.DOCKER_BUILDKIT_CACHE }}
-        key: ${{ runner.os }}-buildx-${{ matrix.target }}-${{ env.TAG }}
+        key: ${{ runner.os }}-buildx-multiarch-${{ matrix.target }}-${{ env.TAG }}
         restore-keys: |
+          ${{ runner.os }}-buildx-multiarch-${{ matrix.target }}-
           ${{ runner.os }}-buildx-${{ matrix.target }}-
```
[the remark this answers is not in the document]
cpretzer: disregard this. I wrote this a couple of weeks ago and didn't submit it

.github/workflows/release.yml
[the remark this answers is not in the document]
cpretzer: disregard this. I wrote this a couple of weeks ago and didn't submit it

bin/_docker.sh
```diff
@@ -53,10 +62,33 @@ docker_build() {
...
+
+        output_params="--platform $SUPPORTED_ARCHS"
+        if [ -n "$DOCKER_PUSH" ]; then
+          output_params+=" --push"
+        else
+          echo "Error: env DOCKER_PUSH=1 is missing"
+          echo "When building the multi-arch images it is required to push the images to the registry"
+          echo "See https://github.com/docker/buildx/issues/59 for more details"
+          exit 1
+        fi
```
kleimkuhler, line 85: Is there a specific reason `DOCKER_PUSH` exists if it's always needed when `DOCKER_MULTIARCH` is set?

For example, if we only set `DOCKER_MULTIARCH=1` in `release.yml`, then we would just always append ` --push` to `output_params`.

From what I can tell, we don't check for `DOCKER_PUSH` in any other case, so it seems like it's a variable that has to be set for no specific reason.

aliariff: the reason is to make it to be more explicit, as @alpeb suggested in his [comment](https://github.com/linkerd/linkerd2/pull/4794#discussion_r461736683)

kleimkuhler: Ah okay that makes sense. I'm good with being more explicit 👍

PR #4795, opened 2020-07-24
Add support for service-mirror selectors

### LINKED ISSUE #4481: Service Selectors

Currently, when you'd like a service exported to another cluster, you add an annotation. The source service-mirror watches all services and then filters based off the annotation. This has a couple downsides:

- There's no way for a source to filter what services they'd like in the local cluster. This has implications around cluster ips and slowing down iptables updates. It also could be too much for a small dev cluster.
- There's an implicit assumption that exporting a service is what actually grants access. Currently, access is based on the trust anchor and technically allows full access to services that haven't been exported.

Instead of using an annotation on the target side, let's switch the ownership of what services are moved to the source side. This removes the assumption that exporting services is related to policy and simultaneously provides the tools required for a source cluster to manage what is ending up on their cluster.

The requirements for this are:

- A default selector is added at install time.
- Support `matchLabels` and `matchExpressions`
- Multiple selector sets can be added as an `OR`.
- Users can update the selector list at any time.
- Users can introspect what services are being matched.

Note that this effectively removes the `export-service` command as `kubectl label` does the same thing.

#### ConfigMap

Following the pattern set by Prometheus and Grafana, the selectors should be stored as a `ConfigMap`:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: service-selector
  namespace: linkerd-multicluster
data:
  selectorConfiguration: |-
    - matchLabels:
        config.linkerd.io/multicluster: true
    - matchExpressions:
        - key: config.linkerd.io/multicluster
          operator: NotIn
          values:
            - true
    - matchLabels:
        foo: bar
      matchExpressions:
        - {key: environment, operator: NotIn, values: [dev]}
```

We should mount this as a volume and reload the service-mirror on change. The output should be part of the logs and any errors parsing should cause service-mirror to output the error and crash.

#### RBAC

This solution primarily provides control to the source side of the equation. If the target cluster would like to restrict what services are being made available, it can be done by changing the RBAC output from `linkerd multicluster allow`.

### Open Questions

- Should there be implicit labels like cluster name and namespace added automatically?

On the source cluster:

```bash
kubectl -n linkerd-multicluster edit configmap service-selector
```

On the target cluster:

```bash
kubectl -n default label svc my-service arbitrary-label=foobar
```

adleong:
Fixes #4481 

We add support for the source cluster to now specify which services from the target cluster should be exported by using a 
label selector rather than looking for the `mirror.linkerd.io/gateway` and `mirror.linkerd.io/gateway-ns` annotations.  Further context for this change is described in https://github.com/linkerd/rfc/pull/31

The `Link` resource spec now has a `selector` field which contains a [label selector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#labelselector-v1-meta).  All services in the target cluster which match this selector will be mirrored.  This selector can be set by using the `--selector` flag to the `multicluster link` command and follows the same format as the `--selector` flag does in kubectl.  The selector can alternatively be set by editing the Link resource directly.  If the selector flag to `multicluster link` is left unspecified, it defaults to the selector: `mirror.linkerd.io/exported`.

Some example usage:

1. Using the default selector:

```
linkerd --context=target multicluster link --cluster-name=foo | kubectl --context=source apply -f -
kubectl --context=target -n emojivoto label svc/web-svc mirror.linkerd.io/exported=true
# the web-svc service is now mirrored to the source cluster
```

2. Using a custom selector

```
linkerd --context=target multicluster link --cluster-name=foo -l exportme=please | kubectl --context=source apply -f -
kubectl --context=target -n emojivoto label svc/web-svc exportme=please
# the web-svc service is now mirrored to the source cluster
```

### DIFF
--- a/pkg/multicluster/link.go
+++ b/pkg/multicluster/link.go
@@ -1,6 +1,7 @@
 package multicluster
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
@@ -41,6 +42,7 @@ type (
 		GatewayPort                   uint32
 		GatewayIdentity               string
 		ProbeSpec                     ProbeSpec
+		Selector                      metav1.LabelSelector
 	}
 )
 
@@ -121,6 +123,18 @@ func NewLink(u unstructured.Unstructured) (Link, error) {
 		return Link{}, err
 	}
 
+	selector := metav1.LabelSelector{}
+	if selectorObj, ok := specObj["selector"]; ok {
+		bytes, err := json.Marshal(selectorObj)
+		if err != nil {
+			return Link{}, err
+		}
+		err = json.Unmarshal(bytes, &selector)
+		if err != nil {
+			return Link{}, err
+		}
+	}
+
 	return Link{
 		Name:                          u.GetName(),
 		Namespace:                     u.GetNamespace(),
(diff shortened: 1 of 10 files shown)


pkg/multicluster/link.go
```diff
@@ -121,6 +123,18 @@ func NewLink(u unstructured.Unstructured) (Link, error) {
 		return Link{}, err
 	}
 
+	selector := metav1.LabelSelector{}
```
zaharidichev, line 126: Maybe worth putting a comment that this piece of code is to exercise the serde rountrip

PR #4797, opened 2020-07-24
Fixed `linkerd check` not finding Prometheus

alpeb:
## The Problem

`linkerd check` run right after install is failing because it can't find the Prometheus Pod.

## The Cause

The "control plane pods are ready" check used to verify the existence of all the control plane pods, blocking until all the pods were ready.

Since #4724, Prometheus is no longer included in that check because it's checked separately as an add-on. An unintended consequence is that when the ensuing "control plane self-check" is triggered, Prometheus might not be ready yet and the check fails because it doesn't do retries.

## The Fix

The "control plane self-check" uses a gRPC call (it's the only check that does that) and those weren't designed with retries in mind.

This PR adds retry functionality to the `runCheckRPC()` function, making sure the final output remains the same

It also temporarily disables the `upgrade-edge` integration test because after installing edge-20.7.4 `linkerd check` will fail because of this.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1038,8 +1038,11 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
-					description:         "control plane self-check",
-					hintAnchor:          "l5d-api-control-api",
+					description: "control plane self-check",
+					hintAnchor:  "l5d-api-control-api",
+					// to avoid confusing users with a prometheus readiness error, we only show
+					// "waiting for check to complete" while things converge. If after the timeout
+					// it still hasn't converged, we show the real error (a 503 usually).
 					surfaceErrorOnRetry: false,
 					fatal:               true,
 					retryDeadline:       hc.RetryDeadline,
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1038,10 +1038,11 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
-					description:   "control plane self-check",
-					hintAnchor:    "l5d-api-control-api",
-					fatal:         true,
-					retryDeadline: hc.RetryDeadline,
+					description:         "control plane self-check",
+					hintAnchor:          "l5d-api-control-api",
+					surfaceErrorOnRetry: false,
+					fatal:               true,
+					retryDeadline:       hc.RetryDeadline,
 					checkRPC: func(ctx context.Context) (*healthcheckPb.SelfCheckResponse, error) {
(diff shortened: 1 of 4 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -1038,10 +1038,11 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
-					description:   "control plane self-check",
-					hintAnchor:    "l5d-api-control-api",
-					fatal:         true,
-					retryDeadline: hc.RetryDeadline,
+					description:         "control plane self-check",
+					hintAnchor:          "l5d-api-control-api",
+					surfaceErrorOnRetry: false,
```
Pothulapati, line 1043: Any reason this has to be false? As its hard to understand which check we are waiting for with just `\ waiting for check to complete`?

Pothulapati: I get the following output, by making it `true`
```
- Error calling Prometheus from the control plane: server_error: server error: 503
```
This seems better than the waiting for check right?

pkg/healthcheck/healthcheck.go
```diff
@@ -1353,49 +1354,75 @@ func (hc *HealthChecker) runCheck(categoryID CategoryID, c *checker, observer Ch
...
+				checkResult.Retry = time.Now().Before(c.retryDeadline)
+				// only show the waiting message during retries,
+				// and send the underlying error on the last try
+				if !c.surfaceErrorOnRetry && checkResult.Retry {
+					checkResult.Err = errors.New("waiting for check to complete")
+				}
+				observer(checkResult)
+			} else if !checkResult.alreadyObserved(observedResults) {
+				observer(checkResult)
+			}
```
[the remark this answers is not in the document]
alpeb: The control plane self-check (which is the only check that is using `checkRPC`) returns two responses, one for the connectivity check against k8s and another for the connectivity check against prometheus. This inner `for` might have already run in a previous try (outer `for`) with a success against k8s and a failure against prom, and we don't wanna report the success again. Without this check `linkerd check` would show something like:
```
linkerd-api
-----------
√ control plane self-check
√ tap api service is running
```

pkg/healthcheck/healthcheck.go
alpeb: As you know, prometheus takes a long while to become available, so users that run `linker check` right after install will always see that error. I thought it was less scary for them to just let them know we're waiting for the check. If prometheus doesn't come up by the time-out, the real error is shown:
```
linkerd-api
-----------
√ control plane self-check
    Error calling Prometheus from the control plane: server_error: server error: 503

Status check results are ×
```

pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
kleimkuhler: Ah cool thanks for the explanation!

PR #4799, opened 2020-07-27
publicAPI: move public-api into a add-on

Pothulapati:
This PR movies the public-api component into its own add-on, thus making
it optional.

This follows the same model as other add-ons, except that
it defines its own struct i.e `publicApi` with fields rather than being
a `map[string]interface{}` so that these fields can be used in code to
be able to be overridden from options.


charts/add-ons/publicApi/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
```
alpeb, line 1: This file requires to be cleaned up and completed like the other addons `Chart.yaml` files.

pkg/charts/linkerd2/addons.go
```diff
@@ -43,5 +43,11 @@ func ParseAddOnValues(values *Values) ([]AddOn, error) {
 		}
 	}
 
+	if values.PublicAPI != nil {
+		if values.PublicAPI.Enabled {
+			addOns = append(addOns, values.PublicAPI)
+		}
```
alpeb, line 49: Why is this different than the other addon checks above?

Pothulapati:
Minimal Control Plane effort is being worked as a separate effort (with an RFC) along with a slightly different approach. So, Closing this! See #4612

PR #4800, opened 2020-07-27
conformance validation: move `tap_test.go` test helpers to `testutil`

mayankshah1607:
The `tap` integration tests currently make use of 2 useful helper functions - `validateExpected` and `tap`, and struct `tapEvent`. This PR moves these into the `testutil` package so that they can be reused for [conformance validation](https://github.com/linkerd/linkerd2-conformance) and to ensure consistency between the 2 repos.

- move `tap` to `testutil.Tap`
- move `validateExpected` to `testutil.ValidateExpected`
- move struct `tapEvent` to `testutil.TapEvent`

Cc: @Pothulapati @ihcsim @alpeb

### DIFF
--- a/testutil/tap.go
+++ b/testutil/tap.go
@@ -1,0 +1,89 @@
+package testutil
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+// TapEvent represents a tap event
+type TapEvent struct {
+	Method     string
+	Authority  string
+	Path       string
+	HTTPStatus string
+	GrpcStatus string
+	TLS        string
+	LineCount  int
+}
+
+// Tap executes a tap command and converts the command's streaming output into tap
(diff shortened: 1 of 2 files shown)


testutil/tap.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+// TapEvent represents a tap event
+type TapEvent struct {
+	Method     string
```
Pothulapati, line 11: are we planning to use these fields directly, and hence we need to make them public here?

mayankshah1607: Yep. We may have to create expected tap events that we'll check against the received tap events using the `ValidateExpected` method. Plus the `tap_test.go` also need to create instances of `TapEvent` for assertion.

PR #4802, opened 2020-07-27
Add multicluster unlink command

### LINKED ISSUE #4707: Add multicluster unlink command
Add a `linkerd multicluster unlink` command which uninstalls an individual multicluster link.

This command requires a `--cluster-name` flag, just like `linkerd multicluster link` does.  

This command outputs manifests of all resources associated with the Link so that they can be piped to `kubectl delete`.

However, a naive implementation may result in a race condition where mirror services are deleted before the service mirror controller and the service mirror controller re-creates them before itself getting deleted.

A more sophisticated approach would be for the `unlink` command to send a signal (such as an HTTP request to the admin port) to the service mirror controller telling it to clean up all of the resources that it manages (mirror services, gateway mirror, and their endpoints) and the block until this is complete.  Only once this cleanup is complete will the `unlink` command then output manifests for the Link resource, the service-mirror-controller, the cluster credentials secret, and the associated RBAC (i.e. all manifests created by `linkerd multicluster link`).

adleong:
Fixes #4707 

In order to remove a multicluster link, we add a `linkerd multicluster unlink` command which produces the yaml necessary to delete all of the resources associated with a `linkerd multicluster link`.  These are:
* the link resource
* the service mirror controller deployment
* the service mirror controller's RBAC
* the probe gateway mirror for this link
* all mirror services for this link

This command follows the same pattern as the `linkerd uninstall` command in that its output is expected to be piped to `kubectl delete`.  The typical usage of this command is:

```
linkerd --context=source multicluster unlink --cluster-name=foo | kubectl --context=source delete -f -
```

This change also fixes the shutdown lifecycle of the service mirror controller by properly having it listen for the shutdown signal and exit its main loop.

A few alternative designs were considered:

I investigated using owner references as suggested [here](https://github.com/linkerd/linkerd2/issues/4707#issuecomment-653494591) but it turns out that owner references must refer to resources in the same namespace (or to cluster scoped resources).  This was not feasible here because a service mirror controller can create mirror services in many different namespaces.

I also considered having the service mirror controller delete the mirror services that it created during its own shutdown.  However, this could lead to scenarios where the controller is killed before it finishes deleting the services that it created.  It seemed more reliable to have all the deletions happen from `kubectl delete`.  Since this is the case, we avoid having the service mirror controller delete mirror services, even when the link is deleted, to avoid the race condition where the controller and CLI both attempt to delete the same mirror services and one of them fails with a potentially alarming error message.


cli/cmd/multicluster.go
```diff
@@ -422,7 +424,7 @@ func newLinkCommand() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "link",
-		Short: "Outputs a Kubernetes secret that allows a service mirror component to connect to this cluster",
+		Short: "Outputs resources that allow another cluster to mirror services from this one",
```
kleimkuhler, line 427: 👍

cli/cmd/multicluster.go
```diff
@@ -634,6 +636,101 @@ func newLinkCommand() *cobra.Command {
...
+			clusterRole := newKubernetesResource(rbac.SchemeGroupVersion.String(), "ClusterRole", fmt.Sprintf("linkerd-service-mirror-access-local-resources-%s", opts.clusterName))
+			clusterRoleBinding := newKubernetesResource(rbac.SchemeGroupVersion.String(), "ClusterRoleBinding", fmt.Sprintf("linkerd-service-mirror-access-local-resources-%s", opts.clusterName))
+			role := newNamespacedKubernetesResource(rbac.SchemeGroupVersion.String(), "Role", fmt.Sprintf("linkerd-service-mirror-read-remote-creds-%s", opts.clusterName), opts.namespace)
+			roleBinding := newNamespacedKubernetesResource(rbac.SchemeGroupVersion.String(), "RoleBinding", fmt.Sprintf("linkerd-service-mirror-read-remote-creds-%s", opts.clusterName), opts.namespace)
+			serviceAccount := newNamespacedKubernetesResource(corev1.SchemeGroupVersion.String(), "ServiceAccount", fmt.Sprintf("linkerd-service-mirror-%s", opts.clusterName), opts.namespace)
+			serviceMirror := newNamespacedKubernetesResource(appsv1.SchemeGroupVersion.String(), "Deployment", fmt.Sprintf("linkerd-service-mirror-%s", opts.clusterName), opts.namespace)
```
kleimkuhler, line 686: Not a blocker, but most of these resource prefixes occur in this file or `healthcheck_multicluster.go`. It may be worth making these constants in a single location.

PR #4803, opened 2020-07-28
Add i18n library to Linkerd dashboard

scottcarol:
Part of #3651 

This PR adds the [LinguiJS project](https://github.com/lingui/js-lingui) to the Linkerd dashboard for i18n and translation. This is an open source project that is well documented and updated, with an active community. After evaluating several libraries this seemed like a lightweight option with easy CLI tools, making it easy for other contributors to learn and use.

This PR ONLY changes 2 React components. The purpose is for reviewers to evaluate the use of the library, adding a language, editing translations, compiling, etc. After this approach is validated a second PR will go up which adds translations for all of the remaining components. 

### Viewing the dashboard

 1. Check out this branch and run `bin/web dev` or `bin/web run`. 
 1. With your browser language set to `en` open http://localhost:7777/namespaces/linkerd/replicationcontrollers
 1. If you have the standard Linkerd setup you should see:
    Screen Shot 2020-07-27 at 4 24 10 PM

 1. Change your browser language to `es`. You should see:
    Screen Shot 2020-07-27 at 4 24 43 PM

### Adding a translation

1. Open `Namespace.jsx` in your code editor.
1. Apply this diff. Note importing `Trans` and wrapping text in it.
  ```diff
  diff --git a/web/app/js/components/Namespace.jsx b/web/app/js/components/Namespace.jsx
  index 65f32e77..c33a33e8 100644
  --- a/web/app/js/components/Namespace.jsx
  +++ b/web/app/js/components/Namespace.jsx
  @@ -7,6 +7,7 @@ import NetworkGraph from './NetworkGraph.jsx';
   import PropTypes from 'prop-types';
   import React from 'react';
   import Spinner from './util/Spinner.jsx';
  +import { Trans } from '@lingui/macro';
   import _filter from 'lodash/filter';
   import _get from 'lodash/get';
   import _isEmpty from 'lodash/isEmpty';
  @@ -147,7 +148,7 @@ class Namespaces extends React.Component {
           {!error ? null : <ErrorBanner message={error} />}
           {!loaded ? <Spinner /> : (
             <div>
  -            {noMetrics ? <div>No resources detected.</div> : null}
  +            {noMetrics ? <div><Trans>No resources detected.</Trans></div> : null}
               {
                 _isEmpty(deploymentsWithMetrics) ? null :
                 <NetworkGraph namespace={ns} deployments={metrics.deployment} />
  ```
3. From `web/app` run `yarn extract`
1. It should indicate that you have 1 missing translation in each file
  Screen Shot 2020-07-27 at 4 28 51 PM
1. Open `locales/es/messages.json` and `locales/en/messages.json` and fill in the blanks.
1. Run `yarn extract` again. It should say you have 0 missing translations.
1. Run `yarn compile`.
1. Load http://localhost:7777/namespaces/blabhablhab from an `en` browser and an `es` browser. You should see the translations you put in. If your browser is set to a different language, only the English key will show.

### Adding a locale

1. From `web/app` run `yarn add-locale fr` then `yarn extract`
1. Open `locales/fr/messages.json` and add sample values
1. Add `import catalogFr from './locales/fr/messages.js’;` to `index.js`
1. Run `yarn compile`
1. Load dashboard with your browser set to `fr`

Signed-off-by: Carol Scott <carol@buoyant.io>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ tmp.discovery
 **/node_modules
 web/web
 web/app/dist
+web/app/js/locales/_build
+web/app/js/locales/**/*.js
 web/app/yarn-error.log
 vendor
 **/*.gogen*
--- a/bin/web
+++ b/bin/web
@@ -55,7 +55,7 @@ dev() {
 
 build() {
   cd "$ROOT"/web/app
-  yarn compile && yarn webpack
+  yarn lingui compile && yarn webpack
 }
 
 get-pod() {
--- a/bin/web
+++ b/bin/web
@@ -55,7 +55,7 @@ dev() {
 
 build() {
   cd "$ROOT"/web/app
-  yarn lingui compile && yarn webpack
+  yarn run lingui compile && yarn webpack
 }
 
 get-pod() {
--- a/bin/web
+++ b/bin/web
@@ -55,7 +55,7 @@ dev() {
 
 build() {
   cd "$ROOT"/web/app
-  yarn run lingui compile && yarn webpack
+  yarn lingui compile && yarn webpack
 }
 
 get-pod() {
--- a/bin/web
+++ b/bin/web
@@ -55,7 +55,7 @@ dev() {
 
 build() {
   cd "$ROOT"/web/app
-  yarn webpack
+  yarn compile && yarn webpack
 }
 
 get-pod() {
--- a/web/app/js/components/EmptyCard.jsx
+++ b/web/app/js/components/EmptyCard.jsx
@@ -1,7 +1,7 @@
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
-import PropTypes from 'prop-types';
 import React from 'react';
+import { Trans } from '@lingui/macro';
 import Typography from '@material-ui/core/Typography';
 import { withStyles } from '@material-ui/core/styles';
 
@@ -12,24 +12,16 @@ const styles = () => ({
   },
 });
 
-const EmptyCard = ({ content, classes }) => {
+const EmptyCard = ({ classes }) => {
   return (
     <Card className={classes.card} elevation={3}>
       <CardContent>
         <Typography>
-          {content}
+          <Trans>No data to display</Trans>
         </Typography>
       </CardContent>
     </Card>
   );
-};
-
-EmptyCard.propTypes = {
-  content: PropTypes.string,
-};
-
-EmptyCard.defaultProps = {
-  content: 'No data to display',
 };
 
 export default withStyles(styles)(EmptyCard);
--- a/web/app/js/components/util/CopyUtils.jsx
+++ b/web/app/js/components/util/CopyUtils.jsx
@@ -1,23 +1,20 @@
 import React from 'react';
+import { Trans } from '@lingui/macro';
 import Typography from '@material-ui/core/Typography';
 
 /*
 * Instructions for adding resources to service mesh
 */
 export const incompleteMeshMessage = name => {
-  if (name) {
-    return (
-      <Typography variant="body2">
-        Add {name} to the k8s.yml file<br /><br />
-        Then run <code>linkerd inject k8s.yml | kubectl apply -f -</code> to add it to the service mesh
-      </Typography>
-    );
-  } else {
-    return (
-      <Typography variant="body2">
-        Add one or more resources to the k8s.yml file<br /><br />
-        Then run <code>linkerd inject k8s.yml | kubectl apply -f -</code> to add them to the service mesh
-      </Typography>
-    );
-  }
+  const unspecifiedResources = <Trans>one or more resources</Trans>;
+  const inject = <code>linkerd inject k8s.yml | kubectl apply -f -</code>;
+
+  return (
+    <Typography variant="body2">
+      <Trans>
+        Add {name || unspecifiedResources } to the k8s.yml file<br /><br />
+        Then run {inject} to add it to the service mesh
+      </Trans>
+    </Typography>
+  );
 };
--- a/web/app/js/index.js
+++ b/web/app/js/index.js
@@ -9,6 +9,7 @@ import AppContext from './components/util/AppContext.jsx';
 import Community from './components/Community.jsx';
 import CssBaseline from '@material-ui/core/CssBaseline';
 import Gateway from './components/Gateway.jsx';
+import { I18nProvider } from '@lingui/react';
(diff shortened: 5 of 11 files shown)


.gitignore
```diff
@@ -8,6 +8,8 @@ tmp.discovery
 **/node_modules
 web/web
 web/app/dist
+web/app/js/locales/_build
```
scottcarol, line 11: From [docs](https://lingui.js.org/guides/excluding-build-files): 
`extract` command creates temporary message catalogs per each source file. Also,`compile` command generates compiled message catalogs from source ones. *All these files can be safely ignored from VCS and linters.*

bin/web
```diff
@@ -55,7 +55,7 @@ dev() {
 
 build() {
   cd "$ROOT"/web/app
-  yarn webpack
+  yarn compile && yarn webpack
```
scottcarol, line 58: we run this to generate compiled message catalogs

web/app/js/components/EmptyCard.jsx
```diff
@@ -12,24 +12,16 @@ const styles = () => ({
...
 });
 
-const EmptyCard = ({ content, classes }) => {
+const EmptyCard = ({ classes }) => {
   return (
     <Card className={classes.card} elevation={3}>
       <CardContent>
         <Typography>
-          {content}
+          <Trans>No data to display</Trans>
```
scottcarol, line 20: This is an example of translating a static text string.

Also, we were never passing in a prop `content` to this component, so we could remove the propTypes checks.

web/app/js/components/util/CopyUtils.jsx
```diff
@@ -1,23 +1,20 @@
...
-        Then run <code>linkerd inject k8s.yml | kubectl apply -f -</code> to add them to the service mesh
-      </Typography>
-    );
-  }
+  const unspecifiedResources = <Trans>one or more resources</Trans>;
+  const inject = <code>linkerd inject k8s.yml | kubectl apply -f -</code>;
+
+  return (
+    <Typography variant="body2">
+      <Trans>
```
scottcarol, line 14: This is an example of translating a block of text containing a dynamic text value.

web/app/js/index.js
```diff
@@ -72,9 +75,13 @@ class App extends React.Component {
   }
 
   render() {
+    const catalogs = { en: catalogEn, es: catalogEs };
+
     return (
       <AppContext.Provider value={this.state}>
-        <AppHTML />
+        <I18nProvider language={window.navigator.language} catalogs={catalogs}>
```
scottcarol, line 82: Detecting user's language from browser

web/app/js/index.js
```diff
@@ -72,9 +75,13 @@ class App extends React.Component {
   }
 
   render() {
+    const catalogs = { en: catalogEn, es: catalogEs };
```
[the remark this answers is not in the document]
scottcarol: Good point, I updated it so that we are only passing the necessary catalog into `I18nProvider`.

I tried taking it one step further and dynamically loading the specific catalog we needed into `index.js`:
```
const selectedCatalog = require(`./locales/${selectedLocale}/messages.js`);
```
but I hit the [no-dynamic-require](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-dynamic-require.md) es-lint rule, which seems to be best practices, so I am continuing to load all of the possible catalogs into this file.

web/app/package.json
```diff
@@ -90,5 +95,10 @@
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
       "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js"
     }
+  },
+  "scripts": {
+    "add-locale": "lingui add-locale",
+    "extract": "lingui extract",
+    "compile": "lingui compile"
```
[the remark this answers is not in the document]
scottcarol: Yes, that's a good point. To be more explicit, I updated the commands so they include `lingui`.

scottcarol:
@alpeb Thanks for taking a look! 

Based on your feedback around generic locale fallback support, I took a look in the `js-lingui` repo and they recommend using the [locales-detector](https://github.com/fabulator/locales-detector) library, so I added this. 

If my browser locales are `es-AR, fr-CA`, we create an array of languages: `['es-AR', 'es', 'fr-CA', 'fr']` and return the first match to an existing catalog (and default to the 'en' catalog if there are no matches). 

If this updated branch makes sense to you, I'll pull out the logic into its own `util` file and add tests.

web/app/package.json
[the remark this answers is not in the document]
alpeb: In that case the `scripts` section is no longer necessary, is it? Or are you leaving that there just to document the available commands?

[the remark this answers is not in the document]
scottcarol: You're totally right, it's not needed now. Just removed, and the `BUILD.md` should be the file with the documentation of commands. I can add documentation in the follow-up PR I will make after this.

alpeb:
Nice, this fixes the case when say the browser uses `es-AR` and if that locale is missing it falls back to `es` :+1: 

Do you know if there's a chance to have this fallback mechanism operate at the single phrase level?
E.g. I have a complete catalog `locales/es/messages.js` with all the entries for standard Spanish, and I also have `locales/es_AR/messages.js` but that only contains overrides for phrases that differ from the standard Spanish. So if an entry is not there, it should fallback to the entry from the `es` locale. 
Maybe if lingui doesn't support that, the right workflow in that case could be having users copy all the entries from `es` and overwrite what they need in `es-AR`...

scottcarol:
Good question @alpeb:
> Do you know if there's a chance to have this fallback mechanism operate at the single phrase level?

When we run `lingui add-locale my-LOCALE` and then `lingui extract` it creates a new JSON file with all of the existing keys and empty values. If we want to use the `lingui extract` command to call out anything we need to fix/add, we will need to have complete keys/values for each translation file. Otherwise any empty values would be listed as "missing" in the command output. I'm thinking there is a benefit for each translation file to stand independently, since it might be hard to communicate which values are actually missing and which values are intentionally blank, but I'll look back at the docs.

scottcarol:
@alpeb I can't immediately find best practices on individual default fallback values, but if it's ok with you, we could merge this PR, in a follow-up PR add translations for `es` and `en`, and then decide on our process for adding `locale-SPECIFIC` formats.

PR #4810, opened 2020-07-30
Fix tap filter authority

### LINKED ISSUE #4697: Dashboard: Tap filter - Authority drop down not populating
### What is the issue?
Not able to configure tap with authority options. There is an authority dropdown that isn't populated. Should probably be an input box and not a dropdown if it's not intended to auto populate.

#### `linkerd check` output

```
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

linkerd-ha-checks
-----------------

linkerd-addons
--------------

linkerd-grafana
---------------
√ grafana pod is running

Status check results are √
```

### Environment

- Kubernetes Version: 1.17.5
- Cluster Environment: AWS
- Host OS: Ubuntu
- Linkerd version: 2.8.1

### Possible solution
Convert dropdown to input box.

tharun208:
The tap component is calling fetch metrics with `skip_stats` and authority
service type is not sent. So, authority dropdown is not getting populated.

Removed `skip_stats` query param from tap metrics

Fixes #4697

Signed-off-by: Tharun <rajendrantharun@live.com>

### DIFF
--- a/web/app/js/components/Tap.jsx
+++ b/web/app/js/components/Tap.jsx
@@ -262,11 +262,12 @@ class Tap extends React.Component {
     });
 
     const url = this.api.urlsForResourceNoStats('all');
-    this.api.setCurrentRequests([this.api.fetchMetrics(url)]);
+    const authorityUrl = this.api.urlsForResource('authority');
+    this.api.setCurrentRequests([this.api.fetchMetrics(url), this.api.fetchMetrics(authorityUrl)]);
     this.serverPromise = Promise.all(this.api.getCurrentPromises())
-      .then(([rsp]) => {
-        const { resourcesByNs, authoritiesByNs } = groupResourcesByNs(rsp);
-
+      .then(rsp => {
+        const { resourcesByNs } = groupResourcesByNs(rsp[0]);
+        const { authoritiesByNs } = groupResourcesByNs(rsp[1]);
         this.setState({
           resourcesByNs,
           authoritiesByNs,


Pothulapati:
I just tested this out and the dashboard was definitely not as responsive as it was previously. I think we have two options here
- Find a way to load authority without doing the full stats load.
- Allow Authority option to be a field rather than a list, so that atleast users can enter their own values. just like they do with CLI right now.

WDYT? @cpretzer

tharun208:
> I just tested this out and the dashboard was definitely not as responsive as it was previously. I think we have two options here
> 
> * Find a way to load authority without doing the full stats load.
> * Allow Authority option to be a field rather than a list, so that atleast users can enter their own values. just like they do with CLI right now.
> 
> WDYT? @cpretzer

@Pothulapati Instead can we make a separate call with `resource_type` as authority ?.

ihcsim:
> Instead can we make a separate call with resource_type as authority ?.

@tharun208 Let's give that a try. We definitely don't want the dashboard performance to degrade. Thanks!

tharun208:
> > Instead can we make a separate call with resource_type as authority ?.
> 
> @tharun208 Let's give that a try. We definitely don't want the dashboard performance to degrade. Thanks!

@ihcsim , can we include the `authority` in the stats table response ? . It is a non k8's resource, so it is ignored i think

ihcsim:
@tharun208 Authority is included as part of the [stats `all` type](https://github.com/linkerd/linkerd2/blob/6467b709daf19cff85d6ff71fab3ae10feac1e97/pkg/k8s/k8s.go#L74). I think the issue is that `api.urlsForResourceNoStats('all');` sets `skip_stats=true`, which causes the [relevant code](https://github.com/linkerd/linkerd2/blob/6467b709daf19cff85d6ff71fab3ae10feac1e97/controller/api/public/stat_summary.go#L373) in the backend `nonK8sResourceQuery()` function to be skipped. WDYT? We can try removing that `SkipStats` conditional from the backend function, and see if it affects the dashboard performance. Also, we will need to confirm there are no side effects on the `linkerd stat` CLI.

cc @alpeb who made the original code change.

alpeb:
I like @tharun208 's idea (IIUC) of issuing a separate additional call like `/api/tps-reports?resource_type=authority&all_namespaces=true`.
That'll result in 4 prom queries, vs 40 if we use `this.api.urlsForResource('all')`

ihcsim:
@tharun208 WDYT of @alpeb suggestion?

tharun208:
> @tharun208 WDYT of @alpeb suggestion?

yes, I am working on it

Pothulapati (approved):
This works great! Also, I don't see any kind of lag in Dashboard!

Issue #4818, opened 2020-07-30
[2.8.1] CLI Standard Error: "could not set namespace from kubectl; ..."


masters3d:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**: When I upgrade  (or install from scratch) using the CLI to linkerd 2.8.1 I start getting the following error: 

```
2020-07-30T22:47:11.4898911Z ##[error]Bash wrote one or more lines to the standard error stream.
2020-07-30T22:47:11.4909905Z ##[error]time="2020-07-30T22:44:51Z" level=error msg="could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set"
2020-07-30T22:47:11.4911647Z ##[error]time="2020-07-30T22:44:53Z" level=error msg="could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set"
2020-07-30T22:47:11.4912834Z ##[error]time="2020-07-30T22:47:03Z" level=error msg="could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set"
```

**What you expected to happen**: no error.

**How to reproduce it (as minimally and precisely as possible)**: Use 2.8.1 CLI to install/upgrade linkerd on AKS 1.18.02
```bash
# install command
$linkerd2_target_bin install --kubeconfig="$KUBECONFIG_FILE_NAME" --ha --controller-replicas=2 | $KUBECTL_BIN --kubeconfig="$KUBECONFIG_FILE_NAME" apply -f -

# upgrade command
$linkerd2_target_bin upgrade --kubeconfig="$KUBECONFIG_FILE_NAME" --ha --controller-replicas=2 2>/dev/null | $KUBECTL_BIN --kubeconfig="$KUBECONFIG_FILE_NAME" apply -f -
```

**Anything else we need to know?**:

``` Logs
2020-07-30T22:44:51.9647484Z time="2020-07-30T22:44:51Z" level=error msg="could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set"
2020-07-30T22:44:53.2489277Z time="2020-07-30T22:44:53Z" level=error msg="could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set"
2020-07-30T22:44:54.0942312Z Running linkerd version: Server version: stable-2.7.1
2020-07-30T22:44:54.0943203Z Linkerd attempting to upgrade Server version: stable-2.7.1 to stable-2.8.1
[...]

PR #4824, opened 2020-07-31
Change log output for failed namespace lookup

cpretzer:
## What
This PR changes the log output when retrieving the namespace from the `kubeconfig` during install and upgrade operations

## Why
As reported in #4818, if an error occurs while retrieving the namespace set for the current context, the error log message that is written doesn't print the underlying error nor does the message indicate that the `default` namespace will be used instead.

## How
Change the log message to `warn` level, print the underlying error from `client-go`, and report that the `default` namespace will be used

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -207,7 +207,8 @@ func getDefaultNamespace() string {
 	ns, _, err := kubeCfg.Namespace()
 
 	if err != nil {
-		log.Errorf("could not set namespace from kubectl context: ensure a valid KUBECONFIG path has been set")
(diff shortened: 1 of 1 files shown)


Matei207:
This reads much better and is more informative than what I had originally put in there. Puts things into perspective for me, thank you!

cpretzer:
> I think we should make this message multiline instead of separating it into two `Warnf`s.

@kleimkuhler had that same thought initially, and changed it because I convinced myself it was more readable. Just updated with your suggestion 😄

cpretzer:
@kleimkuhler TIL about backticks preserving line breaks. Thanks! Update committed

PR #4837, opened 2020-08-04
Temporarily disable job `psscript-analyzer` in static checks

alpeb:
The job started failing consistently today with:
```
##[error]devblackops/github-action-psscriptanalyzer/v2/action.yml (Line:
30, Col: 9): Unexpected value ''
##[error]devblackops/github-action-psscriptanalyzer/v2/action.yml (Line:
30, Col: 9): Unexpected value ''
##[error]System.ArgumentException: Unexpected type 'NullToken'
encountered while reading 'outputs'. The type 'MappingToken' was
expected.
```

It seems it's something in Github that changed today that is clashing
with the `devblackops/github-action-psscriptanalyzer` action.

I've raised devblackops/github-action-psscriptanalyzer#12

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -49,22 +49,22 @@ jobs:
       # https://github.com/koalaman/shellcheck/wiki/Checks
       run: |
         bin/shellcheck-all
-  psscript-analyzer:
-    name: PSScriptAnalyzer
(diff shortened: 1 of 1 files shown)


cpretzer:
ah, disregard my last question. I just saw the devblackops issue you raised

PR #4840, opened 2020-08-05
Add multicluster uninstall command

### LINKED ISSUE #4708: Add a multicluster uninstall command
Add a `linkerd multicluster uninstall` command which is used to uninstall the multicluster add-on.  

The command should first validate that there are no `Link` resources in the cluster.  If any Links exist, the user should be advised to `linkerd multicluster unlink` them first.  Once validation is successful, the command should output all manifests created by `linkerd multicluster install` so that they may be piped to `kubectl delete`.  This includes:

* the linkerd-multicluster namespace and all its contents
* the Link CRD
* all cluster scoped service mirror controller RBAC
* the remote access service account and associated RBAC

adleong:
Fixes #4708 

Adds a `linkerd multicluster uninstall` command which outputs the manifests required to uninstall the mutlicluster components.  This command first checks that no links exist and advises that any links must be removed with `linkerd multicluster unlink` before proceeding.  Typical usage is:

```
linkerd multicluster uninstall | kubectl delete -f -
```


cli/cmd/multicluster.go
```diff
@@ -409,6 +409,92 @@ func newMulticlusterInstallCommand() *cobra.Command {
...
+				Files:     files,
+			}
+			buf, err := chart.RenderNoPartials()
+			if err != nil {
+				return err
+			}
+			stdout.Write(buf.Bytes())
+			stdout.Write([]byte("---\n"))
+
+			return nil
```
kleimkuhler, line 489: I noticed this is the entirety of `newMulticlusterInstallCommand`.

I don't think this is a blocker for me, but it may be worth having this be a helper function that both commands call.

I could go either way on keeping on keeping it duplicated; it's easy to follow what is happening in both commands without having to jump to a helper function, but introduces parts of code we'll need to keep in sync.

cli/cmd/multicluster.go
```diff
@@ -409,6 +409,92 @@ func newMulticlusterInstallCommand() *cobra.Command {
...
+				return err
+			}
+
+			links, err := mc.GetLinks(k.DynamicClient)
+			if err != nil {
+				return err
+			}
+
+			if len(links) > 0 {
+				err := []string{"Please unlink the following clusters before uninstalling multicluster:"}
```
zaharidichev, line 448: That is really nice

PR #4841, opened 2020-08-05
Release ARM CLI artifacts

aliariff:
@cpretzer @alpeb @kleimkuhler 

# Changes
- Only push single docker image arch `amd64` for `cli-bin`. Inside that image, it will consist of multiple architecture binary files. It used the multi-stage build pattern to produce those binaries. [Ref](https://medium.com/@tonistiigi/advanced-multi-stage-build-patterns-6f741b852fae)
- Add the arch name on the binary filename.

# TODO
- [x] Adjust the installer script in https://github.com/linkerd/website

# Note
- Installer script for brew formula does not need a change, because it builds the binary directly instead of download the existing binary. [Brew Formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/linkerd.rb)

Manual tested in a local fork. https://github.com/aliariff/linkerd2/releases/tag/stable-test

### DIFF
--- a/.github/workflows/cloud_integration.yml
+++ b/.github/workflows/cloud_integration.yml
@@ -80,7 +80,7 @@ jobs:
       id: install_cli
       run: |
         TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
-        CMD="$PWD/target/release/linkerd2-cli-$TAG-linux"
+        CMD="$PWD/target/release/linkerd2-cli-$TAG-linux-amd64"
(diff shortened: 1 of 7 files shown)


alpeb:
Sorry @aliariff , my last comment was utter nonsense! I had seemingly previously built a single arch image and that's what the script used. After clearing my local cache it's working fine. I'll test now the release in my fork and report back.

alpeb:
Before merging this one up, we'll wait for linkerd/website#785 to get merged first.

Issue #4790, opened 2020-07-23
smi-metrics in Linkerd


Pothulapati:
## Bug Report

[smi-metrics](https://github.com/servicemeshinterface/smi-metrics) was included with Linkerd2, to support multi-tenant metrics and was marked as a external feature.

### What is the issue?

There are definitely not many users for this feature, as it was disabled by default. This has not been updated recently, and might have even become stale. Should we continue supporting this or remove it?

### Possible solution

Remove it so that we dont have to keep maintaining the templates

@adleong

adleong:
Thanks for filing this @Pothulapati!

We should remove the SMI metrics resources from the main Linkerd2 chart and CLI as well as remove the experimental metrics commands which consume this API.  We should also add an integration test which installs https://github.com/servicemeshinterface/smi-metrics on top of Linkerd and ensures that it works as intended.  This will help us ensure that the SMI metrics integration will continue to work correctly.

PR #4844, opened 2020-08-06
Integration test for smi-metrics

Pothulapati:
part of #4790 

This PR adds a integration test which installs SMI-Metrics and performs
queries and matches the reply with a regex query. Both, resource and 
Edge queries will be performed.

Currently, As the smi-metrics helm chart is not published publicly it is
stored locally but publishing is currently being worked on and this
PR will be updated to pull that chart before being merged.


test/integration/smimetrics/smi-metrics_test.go
```diff
@@ -0,0 +1,107 @@
...
+			filePath: "testdata/resources.golden",
+		},
+		{
+			queryURL: fmt.Sprintf("/apis/metrics.smi-spec.io/v1alpha1/namespaces/%s/deployments/linkerd-destination/edges", TestHelper.GetLinkerdNamespace()),
+			filePath: "testdata/edges.golden",
+		},
+	}
+
+	timeout := 50 * time.Second
+	err = TestHelper.RetryFor(timeout, func() error {
```
adleong, line 74: do we want to put the retry inside the loop so that a single failed test case doesn't cause all previous test cases to be re-run?

test/integration/smimetrics/smi-metrics_test.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+
+			out, err := TestHelper.Kubectl("", queryArgs...)
+			if err != nil {
+				return fmt.Errorf("failed to query smi-metrics URL %s: %s\n%s", tc.queryURL, err, out)
+			}
+
+			// check resources output
+			vars := struct{ Ns string }{TestHelper.GetLinkerdNamespace()}
+			tpl := template.Must(template.ParseFiles(tc.filePath))
```
adleong, line 89: instead of using golden template files, what if we added the smi-metrics repo as a dependency to pull in the TrafficMetricsList struct.  We can then unmarshal the response as a TrafficMetricsList and assert that it unmarshals without errors and do some asserts on the values in the struct (e.g. that request counts are >0 or something like that)

Pothulapati: Hmm, I don't think it would give any benefits doing that, and also adds more problems on knowing the value boundaries, as sometimes negative values are also returned  https://github.com/servicemeshinterface/smi-metrics/issues/8

In the current approach we just match them with regex :thinking:

adleong: Ok but what are we trying to test?  Testing that the response deserializes successfully as a TrafficMetricsList (perhaps with the correct number of elements in the list) seems much less brittle than an big regex template.

Pothulapati: @adleong Hmm, That makes sense especially, as it might reduce any type of flakiness because of additional/less edges (we did not see this on multiple runs, but helps us be more concrete).

Will update this code to get the `TrafficMetricsList` into a struct and have test-cases based on number of edges.

bin/_test-helpers.sh
```diff
@@ -333,11 +333,11 @@ run_uninstall_test() {
 
 run_deep_test() {
   local tests=()
-  export helm_path="$bindir"/helm
+  setup_helm
```
alpeb, line 336: Sorry, I meant to suggest moving `export helm_path="$bindir"/helm` to `start_test`, not to `setup_helm`. Otherwise calling `setup_helm` here causes building the chart and showing some unwanted output in the (non-helm) deep tests...

adleong:
Nice, I like this approach of unmarshalling the response as a TrafficMetrics or TrafficMetricsList.  :+1:

alpeb (approved):
LGTM :+1:

test/integration/smimetrics/smi-metrics_test.go
```diff
@@ -0,0 +1,168 @@
...
+	// Add the SMI Helm repo
+	args := []string{
+		"repo",
+		"add",
+		"smi",
+		"https://servicemeshinterface.github.io/smi-metrics/",
+	}
+
+	if stdout, stderr, err := TestHelper.HelmRun(args...); err != nil {
+		testutil.AnnotatedFatalf(t, "'helm repo add' command failed\n%s\n%s\n%v", stdout, stderr, err)
```
adleong, line 45: I love that we have this integration test but I am a bit worried about issues with the SMI chart taking down our CI.  It would be nice to have some escape hatch so that if the SMI chart is having issues we can bypass this test.  Perhaps if the Helm installation of the SMI chart fails we should log an annotated warning and abort this integration test.

Do you think that would be visible enough?  Where are the annotated warning surfaced? @alpeb

alpeb: The warning annotation is shown in the CI run's main page (as soon as you click on a run in the Actions tab). Another option is to download the chart (as a tar.gz file) and add it as a test artifact.

Pothulapati: Updated the PR to have the chart in the file system. So that we don't have to an external dependency in CI

alpeb (approved):

adleong (approved):
LGTM.  Does it make sense to rename the `smi-metrics.tgz` to a filename that includes the version?  And include instructions in a readme.md or in a comment saying where this file came from and how to download a new version?

Issue #4828, opened 2020-08-01
The CNI Plugin does not set iptables rules


cpretzer:
### What is the issue?
@jvandermark has been helpful in debugging the CNI plugin and provided these gists[[1](https://gist.github.com/jvandemark/ae2a72b203ad25780b68dfb4d993d80d)][[2](https://gist.github.com/jvandemark/3153e0976b416857b1b9de209af4a51b)]

The error was initially discovered on AKS, and I have reproduced it there as well as on KinD. When the CNI plugin runs, this error occurs:

`invalid character 'T' looking for beginning of value`

I believe that the "T" is from the output that is written as a result of [this change](https://github.com/linkerd/linkerd2-proxy-init/commit/a38ea177aa5bd6b4ddfe036cc041193a519aa613#diff-b90cadcd0928c1e490272f4761a52bacR60).

### How can it be reproduced?
Using `edge-20.7.5` run:
`linkerd install-cni | kubectl apply -f -`
`linkerd install --linkerd-cni-enabled | kubectl apply -f -`

### Logs, error output, etc
In the kubelet logs, you will see:
```Aug 01 22:44:27 kind-control-plane kubelet[732]: E0801 22:44:27.405985     732 remote_runtime.go:105] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to setup network for sandbox "dce453f897022d5c499d8d921f5caf471f713c454284679bb31c147cd27c7617": invalid character 'T' looking for beginning of value```

#### `linkerd check` output

- Linkerd check never finishes

```text
kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version
√ is running the minimum kubectl version

linkerd-existence
-----------------
√ 'linkerd-config' config map exists
√ heartbeat ServiceAccount exist
√ control plane replica sets are ready
√ no unschedulable pods
| linkerd-controller-6cdc88f55f-xs97t status is ContainerCreating
```

### Environment

- Kubernetes Version:  v1.18.2
- Cluster Environment: KinD
- Host OS: macOS
- Linkerd version: edge-20.7.5

### Possible solution
Revert the change from `fmt.Printf` to `log.Printf` in `iptables.go`

### Additional context

The pods are all stuck in ContainerCreating

```text
NAME                                      READY   STATUS              RESTARTS   AGE
[...]

PR #4849, opened 2020-08-06
edge-20.8.1

adleong:
This edge adds multi-arch support to Linkerd! Our docker images now support the
amd64, arm64, and arm architectures.

Note: This edge release contains a known issue where the Linkerd-CNI plugin
does not function correctly. We advise against using the Linkerd-CNI plugin
with this edge release. For more information, see [the issue](https://github.com/linkerd/linkerd2/issues/4828).

* Multicluster
  * Added a multicluster unlink command for removing multicluster links
  * Improved multicluster checks to be more informative when the remote API is
    not reachable
* Proxy
  * Enabled a multi-threaded runtime to substantially improve latency especially
    when the proxy is serving requests for many concurrent connections
* Other
  * Fixed an issue where the debug sidecar image was missing during upgrades
    (thanks @javaducky!)
  * Updated all control plane plane and proxy container images to be multi-arch
    to support amd64, arm64, and arm (thanks @aliariff!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,22 @@
 # Changes
 
+## edge-20.8.1
+
+This edge adds multi-arch support to Linkerd! Our docker images now support the
+amd64, arm64, and arm architectures.
+
+* Multicluster
+  * Added a multicluster unlink command for removing multicluster links
+  * Improved multicluster checks to be more informative when the remote API is
+    not reachable
+* Proxy
+  * Enabled a multi-threaded runtime to substantially
+* Other
+  * Fixed an issue where the debug sidecar image was missing during upgrades
+    (thanks @javaducky!)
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,5 +1,22 @@
...
+
+This edge adds multi-arch support to Linkerd! Our docker images now support the
+amd64, arm64, and arm architectures.
+
+* Multicluster
+  * Added a multicluster unlink command for removing multicluster links
+  * Improved multicluster checks to be more informative when the remote API is
+    not reachable
+* Proxy
+  * Enabled a multi-threaded runtime to substantially
```
hawkw, line 13: to substantially _what_?

adleong: it's a mystery

Issue #4848, opened 2020-08-06
External Grafana url does not work with `https://`


MaT1g3R:
### What is the issue?

We are trying to install linkerd with the helm chart. Setting `global.grafanaUrl` to `https://grafana.example.com` does not work.

### How can it be reproduced?

Set `global.grafanaUrl` to a proper URL including the protocol.

### Logs, error output, etc

```
http: proxy error: dial tcp: lookup https://grafana.example.com: no such host
```

### Possible solution

Seting `global.grafanaUrl` to only include the hostname does work (e.g. `grafana.example.com`)

The documentation should be updated to reflect this, or the config value should be changed to `global.grafanaHostname`

ihcsim:
The proposed fix at https://github.com/linkerd/linkerd2/issues/4838#issuecomment-670509215 might address too. Meanwhile, this variable should be added to the chart README.

hetii:
The same issue here as I try to use grafana included by rancher2 monitoring. 
Now instead `no such host` error I'm able to proxy into grafana but just partial as I see
`If you're seeing this Grafana has failed to load its application files`
This is due the fact that my grafana try to load all additional css etc files without /grafana/ prefix.
Any clue about this issue?

PR #4852, opened 2020-08-07
Document workaround for external Grafana URL

MaT1g3R:
Fixes https://github.com/linkerd/linkerd2/issues/4848

Signed-off-by: Peijun Ma <peijun.ma@protonmail.com>


ihcsim:
@MaT1g3R Thanks for the PR. I think the more complete fix is in https://github.com/linkerd/linkerd2/pull/4857. My suggestion is that we include this doc change in that PR.

@Pothulapati WDYT?

Pothulapati:
@MaT1g3R #4857 has been closed because it does not actually makes sense and adds a lot of extra maintaince burden. This looks good to be merged. Can you re-base to make the tests pass?

MaT1g3R:
@Pothulapati For sure

Pothulapati:
@alpeb Hmm, That makes sense, especially because as the user can't set the URl of a external grafana instance directly without changing its proxy settings, this does not seem to add a lot of value. Also, considering that our min control plane RFC (which is currently in works) is definitely going to change this configuration. WDYT?

cc @ihcsim

MaT1g3R:
I think having the ability to click the Grafana icon on the Linkerd UI and taking you to your own Grafana instance provides value, as we don't really want to run 2 separate Grafana instances but still want to see the dashboards.

ihcsim:
@MaT1g3R I am closing this PR for now. @Pothulapati has an upcoming RFC with design on the minimal control plane feature, which will affect this. Once @Pothulapati is ready to share the RFC, he can share the link with you and feel free to chime in there.

As for the original issue at https://github.com/linkerd/linkerd2/issues/4848, sounds like you resolved it by not including the `https` prefix in the URL, right?

PR #4857, opened 2020-08-10
Allow external redirection for Grafana

### LINKED ISSUE #4838: Grafana links not updated when leveraging external grafana
### What is the issue?
Using the helm-chart I have specified following values:
```yaml
global:
  cniEnabled: true
  grafanaUrl: https://externalurl.com
enforcedHostRegexp: '^linkerd\.(int|qa|stagin|prod)*\.io$'
installNamespace: false
grafana:
  enabled: false
```
On opening the page-source of the linkerd-web we do see the html body being updated as (check label `data-grafana`):
```

  <!doctype html>
  <html lang="en">
  <head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width">
  <title>Linkerd</title>
  <meta name="description" content="Linkerd">
  <meta name="keywords" content="Linkerd">
  <link rel="icon" type="image/png" href="/dist/img/favicon.png">
  <link href="https://fonts.googleapis.com/css?family=Lato:300,400,700,900" rel="stylesheet">
     
  <script type="text/javascript" src="/dist/index_bundle.js" async></script>
   
   
  </head>
   <body>
    
   <div class="main" id="main"
   <data-release-version="stable-2.8.1"
   <data-go-version="go1.14.2"
   <data-controller-namespace="linkerd"
   <data-uuid="abd18360-3053-45fe-b9b7-e55001ec4c44"
   <data-grafana="https://externalurl.com"
   <data-jaeger="">
    
   </div>
    
   </body>
   </html>
 
```
however the grafana link still points to the old dashboard (copying element from html - check `href="/grafana/dashboard/db/linkerd-namespace?var-namespace=int"`):
```
<a target="_blank" href="/grafana/dashboard/db/linkerd-namespace?var-namespace=int">&nbsp;&nbsp;<svg version="1.1" id="Layer_1" x="0px" y="0px" width="15px" height="15px" viewBox="0 0 351 365" xml:space="preserve"><g id="Layer_1_1_"></g><linearGradient id="SVGID_1_grafana" gradientUnits="userSpaceOnUse" x1="175.5" y1="445.4948" x2="175.5" y2="114.0346"><stop offset="0" style="stop-color: rgb(255, 241, 0);"></stop><stop offset="1" style="stop-color: rgb(240, 90, 40);"></stop></linearGradient><path class="grafana-st0" d=".."></path></svg></a>
```

### How can it be reproduced?

update helm-chart values file as:
```yaml
global:
  cniEnabled: true
  grafanaUrl: https://externalurl.com
enforcedHostRegexp: '^linkerd\.(int|qa|stagin|prod)*\.io$'
installNamespace: false
grafana:
  enabled: false
```
and check service `linkerd-web`

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version: 1.16
- Cluster Environment: EKS
- Host OS: 5.4.38-17.76.amzn2.x86_64
- Linkerd version: stable-2.8.1

### Possible solution

Prepend all the links to grafana in linkerd2 web UI with external grafana endpoints

### LINKED ISSUE #4848: External Grafana url does not work with `https://`

Pothulapati:
Fixes #4838

Currently, For Grafana, the web component acts as a proxy to
the given grafana isntance (could be a external one too, using
global.grafanaURL). For proxying, Grafana have to be
configured to allow requests from a proxy and this also this makes
grafana not accept requests for normal URLs. When a user is using a
external grafana instance, this defy's the purpose as the external
grafana instance can't accept normal requests.

This PR updates the web component with a new flag called `--redirect`
which when configured redirects instead of proxying.


adleong:
This ties into the larger question of how these components (web and grafana) should be divided up or bundled into addons.

To be honest, I don't see a lot of value in being able to link to an external grafana in our web dashboard.  I don't really see what the use case is there.

The way that I'd expect this to work in the long term is that the Linkerd viz system would be configured with a `enableGrafana` boolean which controls if the bundled grafana is installed or not.  If set to true we can include links in the web dashboard to our bundled grafana which are proxied through web.  If set to false, we can simply omit the links altogether.

TLDR: If users have their own grafana, that's great, but it's not clear to me why we would link to it from the Linkerd web dashboard.

Pothulapati:
@adleong Hmmm, I agree with you especially as `--grafana-addr` is not required anymore and not giving that address just skips the link altogether.

This seems like a unnecessary configuration maintenance burden! Closing this!

PR #4860, opened 2020-08-10
Refactor upgrade tests to remove use of golden files

adleong:
The upgrade tests were failing due to hardcoded certificates which had expired.  Additionally, these tests contained large swaths of yaml that made it very difficult to understand the semantics of each test case and even more difficult to maintain.

We greatly improve the readability and maintainability of these tests by using a slightly different approach.  Each test follows this basic structure:

* Render an install manifest
* Initialize a fake k8s client with the install manifest (and sometimes additional manifests)
* Render an upgrade manifest
* Parse the manifests as yaml tree structures
* Perform a structured diff on the yaml tree structured and look for expected and unexpected differences

The install manifests are generated dynamically using the regular install flow.  This means that we no longer need large sections of hardcoded yaml in the tests themselves.  Additionally, we now asses the output by doing a structured diff against the install manifest.  This means that we no longer need golden files with explicit expected output.

All test cases were preserved except for the following:

* Any test cases related to multiphase install (config/control plane) were not replicated.  This flow doesn't follow the same pattern as the tests above because the install and upgrade manifests are not expected to be the same or similar.  I also felt that these tests were lower priority because the multiphase install/upgrade feature does not seem to be very popular and is a potential candidate for deprecation.
* Any tests involving upgrading from a very old config were not replicated.  The code to generate these old style configs is no longer present in the codebase so in order to test this case, we would need to resort to hardcoded install manifests.  These tests also seemed low priority to me because Linkerd versions that used the old config are now over 1 year old so it may no longer be critical that we support upgrading from them.  We generally recommend that users upgrading from an old version of Linkerd do so by upgrading through each major version rather than directly to the latest.


cli/cmd/upgrade_test.go
```diff
@@ -19,2386 +25,869 @@ const (
...
-  proxy: |
-    {"proxyImage":{"imageName":"gcr.io/linkerd-io/proxy","pullPolicy":"IfNotPresent"},"proxyInitImage":{"imageName":"gcr.io/linkerd-io/proxy-init","pullPolicy":"IfNotPresent"},"controlPort":{"port":4190},"ignoreInboundPorts":[],"ignoreOutboundPorts":[],"inboundPort":{"port":4143},"adminPort":{"port":4191},"outboundPort":{"port":4140},"resource":{"requestCpu":"100m","requestMemory":"20Mi","limitCpu":"1","limitMemory":"250Mi"},"proxyUid":"2102","logLevel":{"level":"warn,linkerd=info"},"disableExternalProfiles":true,"debugImage":{"imageName":"gcr.io/linkerd-io/debug","pullPolicy":"IfNotPresent"},"debugImageVersion":"edge-19.4.1","destinationGetNetworks":"DST-GET-NETWORKS"}
-  install: |
-    {"cliVersion":"edge-19.4.1","flags":[]}`,
-				`
-kind: Secret
-apiVersion: v1
-metadata:
-  name: linkerd-identity-issuer
-  namespace:
```
alpeb, line 211: is this supposed to be empty?

alpeb: Github is not picking up the right lines. I'm referring to the code around line 210 in `upgrade_test.go`:
```go
} else if pathMatch(diff.path, []string{"metadata", "annotations", "linkerd.io/identity-issuer-expiry"}) {                                                                                                 
} else {  
```

adleong: Yes.  I added a comment to help clarify.

adleong:
Thanks, @alpeb, those are good observations.  I feel like having a explicit understanding of what you're testing and expecting is a prerequisite to a good test.  My goal was to provide enough helpers to make the tests themselves concise and expressive.  Moving away from table-driven-testing also gives us a lot more flexibility and lets each test bring in as much or as little complexity as is required to test that case.

I've moved the manifest parsing and diffing code into a separate file, as you suggested.  These kinds of algorithms require a lot of test cases to check all the different edge cases and I'd prefer not to block on that.

`testOptionsAndFlags` is invoked at the start of each test to allocation options structs used in the test.  I don't think caching these will work since each test will be mutating them.  Do you have reason to believe that this is slow?  As far as I know, it should just be some struct allocations.  I'd expect it to be very fast.

PR #4861, opened 2020-08-11
Triger kube-system HA check based on webhook failure policy

### LINKED ISSUE #4754: HA: Relaxing webhook failure policy still triggers check warnings
## Bug Report
When linkerd is installed in HA mode we have to add  `config.linkerd.io/admission-webhooks=disabled` label in `kube-system `namespace to prevent linkerd to get in the way of system critical workload. However every time I add the label, AKS removes all annotations added in this namespace as AKS team manage the namespace.  The other option  is to change webhook policy in values-ha.yaml file from `fail` to `Ignore`. However After loosing linkerd webhook policy by editing the `linkerd-proxy-injector-webhook-config` mutating webhook configuration, Linkerd check still thinks that the `config.linkerd.io/admission-webhooks =disable `annotation is still needed. 

### What is the issue?
When linkerd is installed in `HA mode` we have to add  `config.linkerd.io/admission-webhooks=disabled` in `kube-system` namespace to prevent linkerd to get in the way of system critical workload.  In HA mode, the admission webhooks use a stricter policy, to ensure uninjected workloads can't enter the cluster. We can also loosen the webhook policy by editing the `linkerd-proxy-injector-webhook-config` mutating webhook configuration, and set the `failurePolicy` to `Ignore`. After change is applied `linkerd check ` still thinks that the `config.linkerd.io/admission-webhooks=disable `annotation is still needed

### How can it be reproduced?
1.	Install linkerd in **HA Mode**
2.	Look for the **MutatingWebookConfiguration** resource in the [value-ha.YAML](https://github.com/linkerd/linkerd2/blob/aea541d6f9bf1b2c2b759918f93b8bf3b3ec5343/charts/linkerd2/values-ha.yaml#L59)  and set it to `ignore`
3.	Apply the change
4.	Run `linkerd check `
After change is applied` linkerd check` still thinks admission webhooks disable annotation is needed
```
‼ pod injection disabled on kube-system
    kube-system namespace needs to have the label config.linkerd.io/admission-webhooks: disabled if HA mode is enabled
    see https://linkerd.io/checks/#l5d-injection-disabled for hints
```
5.	Checked the policy in webhook `**ConfigMap**`, `linkerd-proxy-injector-webhook-config` to make sure, it is set to Ignore

### Logs, error output, etc

```
‼ pod injection disabled on kube-system
    kube-system namespace needs to have the label config.linkerd.io/admission-webhooks: disabled if HA mode is enabled
    see https://linkerd.io/checks/#l5d-injection-disabled for hints
```

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
‼ cli is up-to-date
    is running version 2.7.1 but the latest stable version is 2.8.1

control-plane-version
---------------------
‼ control plane is up-to-date
    is running version 2.7.1 but the latest stable version is 2.8.1

linkerd-ha-checks
-----------------
‼ pod injection disabled on kube-system
    kube-system namespace needs to have the label config.linkerd.io/admission-webhooks: disabled if HA mode is enabled
    see https://linkerd.io/checks/#l5d-injection-disabled for hints

Status check results are √
```

### Environment

- Kubernetes Version:   **v1.16.9**
- Cluster Environment:  **AKS**
- Host OS:  **Ubuntu 16.04** 
- Linkerd version:  **2.7.1**

### Possible solution
I think ` Linkerd check` code might need to check if either `admission-webhooks` policy is set to Ignore or `kube-system` namespace labeled with `config.linkerd.io/admission-webhooks: disabled`

zaharidichev:
This PR changes the HA check that verifies that the `config.linkerd.io/admission-webhooks=disabled` is present on kube-system to be enabled only when the failure policy for the proxy injector webhook is set to `Fail`. This allows users to skip this check in cases when the label is removed because the namespace is managed by the cloud provider like in the case described in #4754

Fix #4754

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1901,7 +1901,7 @@ func (hc *HealthChecker) checkHAMetadataPresentOnKubeSystemNamespace() error {
 
 	val, ok := ns.Labels[k8s.AdmissionWebhookLabel]
 	if !ok || val != "disabled" {
-		return fmt.Errorf("kube-system namespace needs to have the label %s: disabled if HA mode is enabled", k8s.AdmissionWebhookLabel)
+		return fmt.Errorf("kube-system namespace needs to have the label %s: disabled if injector webhook failure policy is Fail", k8s.AdmissionWebhookLabel)
 	}
 
 	return nil
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -23,6 +23,7 @@ import (
 	"github.com/linkerd/linkerd2/pkg/tls"
 	"github.com/linkerd/linkerd2/pkg/version"
 	log "github.com/sirupsen/logrus"
+	admissionRegistration "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


zaharidichev:
@adleong Yes, I was just waiting for the PR that fixes the certs test to merge so I can rebase. Tests fixed now

PR #4862, opened 2020-08-11
Use existing /end-entity in proxy-identity

kleimkuhler:
## Motivation

I was creating a proxy image with a pre-loaded `csr.der` and `key.p8` so that I
could test mocking identity (linkerd/linkerd2-mock-dst#10).

The proxy would start up fine, but the identity was not being verified properly.
This issue was hidden by the fact that I was using the mock identity server, so
it was verifying anything without actually checking the `csr.der`.

The actual issue is that upon starting the proxy image, we overwrite the
`csr.der` and `key.p8` even if they already exist.

This is a problem because the files I pre-loaded the image with were required to
work the pre-loaded signed certificate I also had generated.

The proxy was ending up in a state where it's signed certificate was created
from a *different* `csr.der`, and this of course fails.

## Solution

We no longer overwrite the `csr.der` and `key.p8` if they are present.

If both the `csr.der` and `key.p8` exist, we use those and do not generate new
ones.

If either of the files do not exist, we generate a new key and CSR.

I've confirmed this works with the setup I am trying to create in
olix0r/l2-proxy-harness#2.

### DIFF
--- a/proxy-identity/main.go
+++ b/proxy-identity/main.go
@@ -45,7 +45,7 @@ func main() {
 	}
 
 	if exists {
-		log.Infof("Using with pre-existing CSR: %s and key: %s", keyPath, csrPath)
+		log.Infof("Using with pre-existing CSR: %s and key: %s", csrPath, keyPath)
 		return
 	}
 
--- a/proxy-identity/main.go
+++ b/proxy-identity/main.go
@@ -52,6 +52,9 @@ func main() {
 			log.Fatalf("failed to read key file: %s", err)
 		}
 		k, err := tls.DecodeDERKey(keyb)
+		if err != nil {
+			log.Fatalf("failed to decode key file: %s", err)
+		}
 		key = k.Signer()
 	} else {
 		key, err = generateAndStoreKey(keyPath)
--- a/proxy-identity/main.go
+++ b/proxy-identity/main.go
@@ -45,7 +45,7 @@ func main() {
 	}
 
 	if exists {
-		log.Infof("Using with pre-existing CSR and key")
+		log.Infof("Using pre-existing CSR and key")
 		return
 	}
 
--- a/proxy-identity/main.go
+++ b/proxy-identity/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"crypto"
 	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/x509"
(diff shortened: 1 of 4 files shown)


olix0r:
@kleimkuhler I don't think this is desirable behavior in general though, is it? It seems that the case where we want to start with given keys is special; and in those cases we shouldn't be running the identity wrapper at all. Would it be possible to change our development docker image to avoid CSR generation altogether?

olix0r:
Or can we do something that modifies only https://github.com/linkerd/linkerd2/blob/main/proxy-identity/run-proxy.sh?

kleimkuhler:
@olix0r I assumed this was desirable behavior based off these [comments](https://github.com/linkerd/linkerd2/blob/8b891560f069704911a24de208e46f5637024aea/proxy-identity/main.go#L89-L90).

This is definitely a special case, but I think it is surprising that those files are overwritten even when they are already present.

With it being a special case, I think changes to `run-proxy.sh` could also make sense even if they achieve the same thing. 

Maybe just remove the comment saying the files will be used if already exist? If the changes instead went in to `run-proxy.sh`, are you thinking we would still check for file existence there, or go off a new environment variable or something?

olix0r:
@kleimkuhler that's a good point... i'm trying to make sense of what's going on here ;)

Does that comment even make any sense in the context of that function? It doesn't create those files, so why is it even there? (Not your fault, this seems like something I wrote...)

I think this change is fine from a feature point of view. It seems fine to reuse existing credentials. We should probably emit a warning.

proxy-identity/main.go
```diff
@@ -35,7 +36,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	keyPath, csrPath, err := checkEndEntityDir(*dir)
+	keyPath, keyExists, csrPath, csrExists, err := checkEndEntityDir(*dir)
```
olix0r, line 39: we should probably do this is all with a single exists flag equivalent to `keyExists && csrExists`. If the key doesn't exist and we generate a new one, it won't match the CSR, and vice versa.

kleimkuhler:
@olix0r Yep all good! So the comment kind of makes sense because we acknowledge with those log lines that we found existing files, but it seems like we are just missing the part where we use them; this is what I'm trying to add with this change.

proxy-identity/main.go
kleimkuhler: Ah good point yea I'll change that.

kleimkuhler: This removed a lot of the additional changes that checked the validity of the pre-existing key.

kleimkuhler:
Proxy startup with pre-existing CSR and key looks like this now:

```
time="2020-08-12T00:33:34Z" level=info msg="running version git-e7b9aa29"
time="2020-08-12T00:33:34Z" level=info msg="Found pre-existing key: /end-entity/key.p8"
time="2020-08-12T00:33:34Z" level=info msg="Found pre-existing CSR: /end-entity/csr.der"
time="2020-08-12T00:33:34Z" level=info msg="Using with pre-existing CSR: /end-entity/key.p8 and key: /end-entity/csr.der"
[     0.10586072s] DEBUG ThreadId(01) linkerd2_app: building app
```

proxy-identity/main.go
```diff
@@ -75,30 +91,34 @@ func loadVerifier(pem string) (verify x509.VerifyOptions, err error) {
...
+	keyExists := false
 	if err = checkNotExists(keyPath); err != nil {
 		log.Infof("Using with pre-existing key: %s", keyPath)
+		keyExists = true
 	}
 
 	csrPath := filepath.Join(dir, "csr.der")
+	csrExists := false
 	if err = checkNotExists(csrPath); err != nil {
-		log.Infof("Using with pre-existing CSR: %s", keyPath)
```
olix0r, line 98: I know you're just copying the above log line... but the word _with_ in these messages breaks my brain ;p

kleimkuhler: Ha no you're totally right I removed it!

adleong:
I'm a bit confused about the approach here.  If the idea is to create a proxy image which includes pregenerated key material, why run the proxy-identity script at all?  Shouldn't the proxy just be able to start up directly?

My understanding is that the whole point of the proxy-identity script is to generate these files.  So it's not surprising to me that it overwrites the files if they already exist.

kleimkuhler:
@adleong When I initially came across this as the issue I was experiencing, reading through the code and the comment I referred to above led me think this was a bug with the proxy-identity.

I was pre-loading files and surprised to see based off the code that they were being overwritten.

Now that I'm looking at this as _adding_ the ability to not overwrite the files, I could go either way on where we solve this.

Oliver already mentioned this is a special case, and I don't really see any users needing to do this for actual use. For that reason, if we had a `LINKERD2_PROXY_MOCK_IDENTITY` environment variable (or something like that) that could be set by the proxy harness, we could check for that in the proxy-identity script and just skip generation completely.

kleimkuhler:
After some further conversation I am closing this so that it can be solved with build-args in the proxy.

There are some changes in here that are still worth having, but I'll open in a separate PR since the primary reason for this change is no longer relevant.

PR #4863, opened 2020-08-12
Diff generated code from proto files

### LINKED ISSUE #4669: Generate protobuf in CI to ensure no changes
We should ensure that generated protobuf files don't diverge by regenerating them in CI and failing if the files have changed.

zaharidichev:
This PR adds a static check that ensures the generated files from the proto definitions have not changed. 

Fix #4669

### DIFF
--- a/bin/protoc-diff
+++ b/bin/protoc-diff
@@ -4,19 +4,19 @@ set -eu
 
 # Install unzip if not already installed (Linux only)
 system=$(uname -s)
-if [ "$system" = Linux ] && [ $(dpkg-query -W -f='${Status}' unzip 2>/dev/null | grep -c "ok installed") -eq 0 ];
-then
+if [ "$system" = Linux ] && ! [ -x "$(command -v unzip)" ]; then
   apt-get update
   apt-get install unzip
 fi
 
 bin/protoc-go.sh
 
-dir_dirty=$(git status -s)
+dir_dirty=$(git diff HEAD)
 if [ -z "$dir_dirty" ]; then
   echo "Protobuf definitions match generated code"
   exit 0
 else
-  echo "Protobuf definitions diverge from generated code"
+  echo "Protobuf definitions diverge from generated code:"
+  echo "$(git status)"
   exit 64
 fi
--- a/bin/protoc-diff
+++ b/bin/protoc-diff
@@ -1,12 +1,21 @@
 #!/usr/bin/env sh
 
 set -eu
-dir_dirty=$(git status --untracked-files=no --porcelain)
+
+# Install unzip if not already installed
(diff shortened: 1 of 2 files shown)


bin/protoc-diff
```diff
@@ -0,0 +1,22 @@
...
+system=$(uname -s)
+if [ "$system" = Linux ] && [ $(dpkg-query -W -f='${Status}' unzip 2>/dev/null | grep -c "ok installed") -eq 0 ];
+then
+  apt-get update
+  apt-get install unzip
+fi
+
+bin/protoc-go.sh
+
+dir_dirty=$(git status -s)
```
kleimkuhler, line 15: This is going to pick up untracked files. Funny enough, I had untracked files when testing and this was failing and I didn't see why right away.

What we  probably should check for here is just if any files have changed with `git diff HEAD`.

adleong: if the protoc generates new bindings files which aren't already checked into git, they would be untracked files, right?  Isn't that exactly what we want to catch?  Since this is intended to be run in CI, I don't think there would be any other untracked files hanging around.

bin/protoc-diff
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
+
+set -eu
+
+# Install unzip if not already installed (Linux only)
+system=$(uname -s)
+if [ "$system" = Linux ] && [ $(dpkg-query -W -f='${Status}' unzip 2>/dev/null | grep -c "ok installed") -eq 0 ];
```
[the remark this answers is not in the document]
adleong: what about just `! which -s unzip`?

bin/protoc-diff
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+bin/protoc-go.sh
+
+dir_dirty=$(git status -s)
+if [ -z "$dir_dirty" ]; then
+  echo "Protobuf definitions match generated code"
+  exit 0
+else
+  echo "Protobuf definitions diverge from generated code"
```
adleong, line 20: do you think it's worth printing the output of `git status` so we can see where the differences are?

bin/protoc-diff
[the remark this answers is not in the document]
kleimkuhler: I added a similar check to `markdownlint` script and came across this SO answer on why `command` is better than `which` for this situation: https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script

bin/protoc-diff
kleimkuhler: Yea that's true. This was also coming from the fact that we use `diff-index` (which I think `diff` is just recommend now) in `_tag.sh`. Similar situation so thought I'd suggest it. Definitely not a blocker.

bin/protoc-diff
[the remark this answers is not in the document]
adleong: TIL

PR #4867, opened 2020-08-12
Choose the right architecture in bin/linkerd script

kleimkuhler:
## Motivation

I noticed after a local build of linkerd, `bin/linkerd version` was printing a
different git version than I expected.

I realized `bin/linkerd` (on Linux) was still pointing to
`./target/cli/linux/linkerd` even though we now build and put images into
`./target/cli/linux-amd64/linkerd`.

## Solution

We now select the correct architecture directory based off `uname -m`.

## Continued issues

If a binary does not exist in the path returned by `bin/linkerd`, we attempt to
build it with `bin/build-cli-bin`. I don't think this script is up to date with
multi-arch builds.

For example if I was on `arm64`, I don't think `build-cli-bin` would properly
build a put a binary in `./target/cli/linux-amd64/`.

### DIFF
--- a/bin/linkerd
+++ b/bin/linkerd
@@ -10,7 +10,29 @@ system=$(uname -s)
 if [ "$system" = Darwin ]; then
   bin=$rootdir/target/cli/darwin/linkerd
 elif [ "$system" = Linux ]; then
-  bin=$rootdir/target/cli/linux/linkerd
+  arch=$(uname -m)
+  case $arch in
+    x86_64)
+      arch=amd64
+      ;;
+    armv8*)
+      arch=arm64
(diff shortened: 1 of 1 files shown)


kleimkuhler:
Created #4868 for the continued issue.

kleimkuhler:
@alpeb So `bin/linkerd` is already broken by the fact that `bin/docker-build` (without `LINKERD_LOCAL_BUILD_CLI` set) puts the binary in `linux-amd64`.

This at least fixes that, it's just that it will remain broken for developers who are on Linux but not `x86_64` archs (which is uncommon right now).

PR #4869, opened 2020-08-12
Fix spelling

MaT1g3R:
- Fix spelling: trunacte
- Fix spelling: gatewy


adleong:
Looks like this is a subset of https://github.com/linkerd/linkerd2/pull/4872

jsoref:
Yeah, this is the subset that gets pulled in when one tries to deploy it.

PR #4876, opened 2020-08-13
Fix build-cli-bin

### LINKED ISSUE #4868: Update bin/build-cli-bin to build binary based off architecture
## Feature Request

`bin/build-cli-bin` could benefit from the multi-arch builds we are now doing in
CI. Developers on `arm64` for example could expect that running
`bin/build-cli-bin` would produce a binary in `./target/cli/linux-arm64`.

When using the `bin/linkerd` script, if a binary does not exist in the expected
path then it attempts to build one in `bin/build-cli-bin`. On some
architectures, this will not produce a binary in the correct location.

The script should be updated to build the CLI based off the current
architecture.

--

Users on supported architectures would have a seemless experience using
`bin/linkerd`.

aliariff:
@kleimkuhler @cpretzer @alpeb 

Fixes #4868

Related to https://github.com/linkerd/linkerd2/pull/4867

### DIFF
--- a/bin/build-cli-bin
+++ b/bin/build-cli-bin
@@ -49,10 +49,10 @@ esac
 (
     cd "$rootdir"
     cd "$(pwd -P)"
-		target=target/cli/$host_platform/linkerd
-		if [ -n "$arch" ]; then
-			target=target/cli/$host_platform-$arch/linkerd
-		fi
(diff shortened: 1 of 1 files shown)


aliariff:
@alpeb yap that's the plan, after both of the PR is merged, then the duplication will be addressed

PR #4878, opened 2020-08-13
Upgrade to Helm v3 sdk

### LINKED ISSUE #4874: Move to Helm v3 sdk in CLI
## Feature Request

Currently, we use Helm `v2.16.8`. Helm v2 is moving to its end of lifecycle and wont be supported anymore. Its best to move to Helm v3 internally sdk. Irrespective of #4833, Its best to move the current state of things to Helm v3 so that its lets us experiment with the idea of using Helm in the CLI internally, and also may help clean up the code-base a bit as the Helm Go SDK has been refactored in V3 for cases like us, to be able to perform most of the helm cmd operations.

Pothulapati:
Fixes #4874

This branch upgrades Helm sdk from v2 to v3 *without any functionaly
changes*, just replacing types with newer API's.

This should not effect our current support for Helm v2 as we did not
change any of the underlying templates(which work with Helm v2). This
works because we did not use any of the API's that read the Chart
metadata (which are the only ones changed from v2 to v3) and currently
manually load files and pass it to the sdk.

This PR should provide a initial point to start moving to the newer 
Helm v3 API's especially for the upgrade workflow thus allowing us to make
Linkerd CLI more simpler.


cli/cmd/multicluster.go
```diff
@@ -20,6 +20,9 @@ import (
 	"github.com/linkerd/linkerd2/pkg/version"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	chartloader "helm.sh/helm/v3/pkg/chart/loader"
```
alpeb, line 24: :eyes:

Pothulapati:
As we are generating our own secret and not using helm functions internally(makes sense to not add unnecessary complexity as the APIs are not simpler and straightforward for our use-case), This update is not needed!

adleong (approved):
I'm very interested in @alpeb's review of this since he's much more familiar with the details of Helm.  This all looks totally reasonable to me.

By upgrading to v3 here we get access to better helper functions for manipulating a Values struct, is that right?  E.g. for setting a specific value like from a `--set` flag?

Pothulapati:
@adleong Yes, We might get it from `v2` too but a lot of things changed in v3 and hence making the move to v3 would be useful so that we don't have to change much in the future!

alpeb (approved):
This looks pretty benign to me :+1:

PR #4880, opened 2020-08-13
Remove duplication on arch detection

aliariff:
@kleimkuhler @alpeb @cpretzer 

# Changes
- Move the os architecture detection to another file, avoid duplication

### DIFF
--- a/bin/_os.sh
+++ b/bin/_os.sh
@@ -32,13 +32,13 @@ os() {
           arch=$arch
           ;;
         *)
-          echo "unknown architecture: $arch" >&2
+          echo "unsupported architecture: $arch" >&2
           exit 1
           ;;
       esac
       ;;
     *)
-      echo "unknown os: $os" >&2
+      echo "unsupported os: $os" >&2
       exit 1
       ;;
   esac
--- a/bin/_os.sh
+++ b/bin/_os.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -eu
+
+os() {
+  os=$(uname -s)
+  arch=""
+  case $os in
+    CYGWIN* | MINGW64*)
+      os=windows
+      ;;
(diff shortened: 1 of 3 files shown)


bin/_os.sh
```diff
@@ -0,0 +1,51 @@
...
+          arch=arm64
+          ;;
+        armv*)
+          arch=arm
+          ;;
+        amd64|arm64)
+          arch=$arch
+          ;;
+        *)
+          echo "unknown architecture: $arch" >&2
```
kleimkuhler, line 35: I know I worded this differently before, but maybe "unsupported" is better here? We know the architecture, it's just not supported.

```suggestion
          echo "unsupported architecture: $arch" >&2
```

bin/_os.sh
```diff
@@ -0,0 +1,51 @@
...
+          exit 1
+          ;;
+      esac
+      ;;
+    *)
+      echo "unknown os: $os" >&2
```
kleimkuhler, line 41: Similar comment about using "unsupported"

bin/_os.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -eu
+
+os() {
+  os=$(uname -s)
+  arch=""
```
kleimkuhler, line 7: Do we need to set this if it's only used in the `Linux` case and gets set right away by `arch=$(unman -m)`?

aliariff: Yes this one is necessary. Because if we remove it, when it is not Linux, then it will throw error `arch: unbound variable` on `if [ -n "$arch" ]; then`

kleimkuhler: Oh okay yea I wasn't sure if `-n` would evaluate to false when it wasn't set _or_ unbound. Thanks!

Issue #4879, opened 2020-08-13
destination: Always return dst_overrides for existing services


olix0r:
Currently, the destination service (in GetProfile repsonses) only sets dst_overrides when a traffic split is present. Instead, it should always return at least one dst_override when the service exists. The proxy can then use this response as a hint to indicate whether it should perform a destination.Get: if no dst_overrides are present, it is assumed that the target is not a Service and endpoint resolution will be skipped.

kleimkuhler:
Closed by #4890

PR #4890, opened 2020-08-18
Always return destination overrides for services

kleimkuhler:
## Motivation

#4879

## Solution

When no traffic split exists for services, return a single destination override
with a weight of 100%.

Using the destination client on a new linkerd installation, this results in the
following output for `linkerd-identity` service:

```
❯ go run controller/script/destination-client/main.go -method getProfile -path linkerd-identity.linkerd.svc.cluster.local:8080
INFO[0000] retry_budget:{retry_ratio:0.2 min_retries_per_second:10 ttl:{seconds:10}} dst_overrides:{authority:"linkerd-identity.linkerd.svc.cluster.local.:8080" weight:100000} 
INFO[0000]
```

### DIFF
--- a/controller/api/destination/traffic_split_adaptor.go
+++ b/controller/api/destination/traffic_split_adaptor.go
@@ -2,10 +2,12 @@ package destination
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
 	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha2"
 	ts "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/split/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 // trafficSplitAdaptor merges traffic splits into service profiles, encoding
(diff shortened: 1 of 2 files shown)


controller/api/destination/traffic_split_adaptor.go
```diff
@@ -64,11 +66,19 @@ func (tsa *trafficSplitAdaptor) publish() {
...
-	if tsa.profile == nil && tsa.split == nil {
-		tsa.listener.Update(nil)
 	} else {
-		tsa.listener.Update(&merged)
+		// If there is no traffic split, always return a destination override
+		// so that it's known the host is a service.
+		overrides := []*sp.WeightedDst{}
+		id := strings.Split(tsa.id.String(), "/")
+		dst := &sp.WeightedDst{
+			Authority: fmt.Sprintf("%s.%s.svc.%s.:%d", id[1], tsa.id.Namespace, tsa.clusterDomain, tsa.port),
```
zaharidichev, line 75: I think indexing into this array deserves to be a bit more safe?

controller/api/destination/traffic_split_adaptor.go
```diff
@@ -64,11 +66,19 @@ func (tsa *trafficSplitAdaptor) publish() {
...
+			// Weights are expressed in decimillis: 10_000 represents 100%
+			Weight: resource.MustParse("10000m"),
```
adleong, line 77: It doesn't really matter, since there's only one value, but this doesn't match the comment.  10_000m = 10 = 100_000 decimillis = 1000%

To set this to 100% you would want it to be "1000m" (or can you just do "1"?)

controller/api/destination/traffic_split_adaptor.go
```diff
@@ -64,11 +66,19 @@ func (tsa *trafficSplitAdaptor) publish() {
...
+		}
+		overrides = append(overrides, dst)
+		merged.Spec.DstOverrides = overrides
```
adleong, line 80: why not just `meged.Spec.DstOverrdies = []*sp.WeightedDst { dst }`?

PR #4897, opened 2020-08-20
Add ARM64 Integration Test

aliariff:
@cpretzer @alpeb 

# Changes
- Add `arm64_integration_tests` job in the release workflow
- Skip tracing test, because the OpenCensus collector and Jaeger images do not support ARM yet
- Skip `helm-upgrade` and `upgrade-stable` test because the Linkerd stable version does not support ARM yet
- Bump the version of the images that used for testing purposes to version that support ARM

# Note
- Before merging this, it is required to provision an ARM64 machine and also setup Kubernetes on it. Then set a new secret in Github called `ARM64_KUBECONFIG` with the kubeconfig credential from the ARM64 machine.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -188,11 +188,56 @@ jobs:
       run: |
         export TAG="$($CMD version --client --short)"
         go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
+  arm64_integration_tests:
+    name: ARM64 integration tests
+    runs-on: ubuntu-18.04
+    needs: [docker_build]
+    steps:
+    - name: Checkout code
+      # actions/checkout@v2
+      uses: actions/checkout@722adc6
+    - name: Try to load cached Go modules
+      # actions/cache@v1.1.2
+      uses: actions/cache@70655ec
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+    - name: Install linkerd CLI
+      id: install_cli
+      run: |
+        TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
+        CMD="$PWD/target/release/linkerd2-cli-$TAG-linux-amd64"
+        bin/docker-pull-binaries $TAG
+        $CMD version --client
+        # validate CLI version matches the repo
+        [[ "$TAG" == "$($CMD version --short --client)" ]]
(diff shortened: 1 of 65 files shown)


bin/_test-helpers.sh
```diff
@@ -284,6 +284,11 @@ run_upgrade-edge_test() {
 # Run the upgrade-stable test by upgrading the most-recent stable release to the
 # HEAD of this branch.
 run_upgrade-stable_test() {
+  if [ -n "$ARM_TEST" ]; then
```
cpretzer, line 287: What do you think of naming this something more descriptive like `RUN_ARM_TEST`

.github/workflows/release.yml
```diff
@@ -188,11 +188,56 @@ jobs:
...
+      run: |
+        TAG="$(CI_FORCE_CLEAN=1 bin/root-tag)"
+        CMD="$PWD/target/release/linkerd2-cli-$TAG-linux-amd64"
+        bin/docker-pull-binaries $TAG
+        $CMD version --client
+        # validate CLI version matches the repo
+        [[ "$TAG" == "$($CMD version --short --client)" ]]
+        echo "Installed Linkerd CLI version: $TAG"
+        echo "::set-env name=CMD::$CMD"
+        echo "::set-output name=tag::$TAG"
```
alpeb, line 218: `id: install_cli` along with `echo "::set-output name=tag::$TAG"` is used so that another step can refer to the tag using `steps.install_cli.output.tag`. It's not used here, so you can remove that id and this statement.

.github/workflows/release.yml
```diff
@@ -188,11 +188,56 @@ jobs:
...
+      env:
+        RUN_ARM_TEST: 1
+      run: bin/tests --skip-kind-create "$CMD"
+    - name: CNI tests
+      run: |
+        export TAG="$($CMD version --client --short)"
+        go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
+    - name: Test cleanup
+      if: ${{ always() }}
+      run: bin/test-cleanup
```
alpeb, line 235: The VM gets discarded when the jobs finishes, so there's no point on manually cleaning things up.

aliariff: I think we plan to reuse the same machine, we are not going to provision the machine everytime release happen, or do we?

alpeb: gotcha, makes sense :+1:

cpretzer: Yep, that's right, because we would have to install k3s every time, which would dramatically increase test times.

PR #4898, opened 2020-08-20
Introduce support for authenticated docker registries using imagePullSecrets, Fixes #4413

### LINKED ISSUE #4413: Adding support for Docker Private Virtual Registries like Jfrog Artifactory for Airgap systems.
Jfrog Artifactory or other private artifact registries for docker container are common in airgap systems, where the Linkerd chart referencing docker images, cannot be downloaded from the default upstream urls ( example: gcr.io/linkerd-io/proxy ).

Also these private registries are configured with authentication, for which the current implementation has no support.

By adding `imagePullSecrets: <registry-secret>` parameter in the values.yaml (which is not implemented and therefore needs code change, the airgap systems can smoothly install the chart. 

The implementation should assume the registry secret is already available with the namespace.

User has to configure the value for `imagePullSecrets` in Linkerd Chart's values.yaml, and change the image reference of docker images in the values.yaml to point to their Jfrog Registry so that the charts can directly be deployed. 

The other way to support such airgap system is to download the chart in zip format, use helm package command to package it after code changes as shared above, and deploy it in Jfrog Artifactory as part of local chart repository. This is a tedious process.

User will supply the `ImagePullSecrets` parameter to configure their artifactory registry, along with changing the reference of docker images in the values.yaml

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

c-n-c:
Introduce support for authenticated docker registries using imagePullSecrets 

Problem: Private Docker Registries are not supported for the moment as detailed in issue #4413

Solution: Every Service Account of linkerd subcomponents are Attached with imagePullSecrets,
which in turn can then pulls the docker images from authenticated private registries using them.
The imagePullSecret is configured in global.imagePullSecret parameter of values.yaml like

imagePullSecret:
  - name: <name-of-private-registry-secret-resource>

Fixes #4413

Signed-off-by: Nilakhya <nilakhya@hotmail.com>


charts/add-ons/grafana/templates/grafana-rbac.yaml
```diff
@@ -11,3 +11,4 @@ metadata:
   labels:
     {{.Values.global.controllerComponentLabel}}: grafana
     {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
+{{ include "partials.image-pull-secrets" . }}
```
alpeb, line 14: Please add use `{{- include "partials.image-pull-secrets" . }}` to suppress line-feeds/spaces (this is what's making the tests fail)

charts/linkerd2-cni/templates/cni-plugin.yaml
```diff
@@ -56,6 +56,7 @@ metadata:
   namespace: {{.Values.namespace}}
   labels:
     {{.Values.cniResourceLabel}}: "true"
+{{ include "partials.image-pull-secrets" . }}
```
alpeb, line 59: `charts/linkerd2-cni` is a separate chart that doesn't receive the global config from `charts/linkerd2/values.yml`. In order to use `global.imagePullSecrets` in this chart, you need to declare it in `charts/linkerd2-cni/values.yaml`.

charts/linkerd2/values.yaml
```diff
@@ -92,6 +92,13 @@ global:
   linkerdNamespaceLabel: linkerd.io/is-control-plane
   workloadNamespaceLabel: linkerd.io/workload-ns
 
+  #
+  ## Define your image pull secrets for private docker registries like the following
+  #imagePullSecrets:
```
alpeb, line 97: In this comment, please clarify that this is for adding the ImagePullSecret to each of the control plane service accounts (and not adding the ImagePullSecret to each Pod).

charts/add-ons/grafana/templates/grafana-rbac.yaml
```diff
@@ -11,3 +11,7 @@ metadata:
   labels:
     {{.Values.global.controllerComponentLabel}}: grafana
     {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
+{{- if .Values.global.imagePullSecrets }}
```
Pothulapati, line 14: because `partials` is also a sub-chart for add-ons, `{{- include "partials.image-pull-secrets" . }}` should work directly

Pothulapati (approved):
I tested this out by pushing images into the new GitHub package registry.

If you are trying to do the same, Make sure to push and pull the images into `ghcr.io` instead of `docker.pkg.github.com` as the later act weird as per https://github.com/containerd/containerd/issues/3291#issuecomment-685099010

charts/add-ons/grafana/templates/grafana-rbac.yaml
c-n-c: I tried this, but were failing tests so on a later commit i  introduced this.

charts/linkerd2-cni/values.yaml
```diff
@@ -19,3 +19,15 @@ priorityClassName: ""
...
+
+#
+## For Private docker registries, authentication is needed.
+# If the control plane service images are pulled from a
+# protected docker registry, define pull secrets as follows:
+#
+#imagePullSecrets:
+#    - name: my-private-docker-registry-login-secret
+#
+# The pull secrets are applied to the respective service accounts
```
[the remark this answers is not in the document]
c-n-c: I just revamped the comments as per @alpeb in his last review.

charts/partials/templates/_pull-secrets.tpl
```diff
@@ -0,0 +1,6 @@
+{{- define "partials.image-pull-secrets" -}}
+{{- if .Values.global.imagePullSecrets }}
```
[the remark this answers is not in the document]
c-n-c: maintaining homogenousity

charts/linkerd2-cni/templates/cni-plugin.yaml
```diff
@@ -57,6 +57,10 @@ metadata:
   namespace: {{.Values.namespace}}
   labels:
     {{.Values.cniResourceLabel}}: "true"
+{{- if .Values.imagePullSecrets }}
```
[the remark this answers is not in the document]
c-n-c: As per @alpeb in last review,  cni plugin does not import partials, so i included it

charts/add-ons/prometheus/templates/prometheus-rbac.yaml
```diff
@@ -39,3 +39,7 @@ metadata:
   labels:
     {{.Values.global.controllerComponentLabel}}: prometheus
     {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
+{{- if .Values.global.imagePullSecrets }}
```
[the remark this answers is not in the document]
c-n-c: I had to as i shared above, the tests start failing otherwise

charts/linkerd2-cni/values.yaml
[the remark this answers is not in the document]
alpeb: I agree with @Pothulapati that the description is too verbose. My suggestion was just that you clarified that these are ServiceAccount ImagePullSecrets, and not Pod ImagePullSecrets. Also you can provide a commented example like so:

```
  imagePullSecrets: []
  # - name: my-private-docker-registry-login-secret                                                                 
```

charts/add-ons/tracing/values.yaml
```diff
@@ -1 +1 @@
-# Default values for tracing.
\ No newline at end of file
+# Default values for tracing.
```
alpeb, line 1: Would you please revert this change?

charts/add-ons/prometheus/values.yaml
```diff
@@ -13,3 +13,4 @@ globalConfig:
   scrape_interval: 10s
   scrape_timeout: 10s
   evaluation_interval: 10s
+
```
alpeb, line 16: Same here

charts/add-ons/grafana/templates/grafana-rbac.yaml
alpeb: Just using `{{- include "partials.image-pull-secrets" . }}` works for me... Can you push it just with that, and so we can see what's the actual failure in CI?

alpeb (commented):
In order to have this work from the install CLI and not just Helm, you need to add this to `./pkg/charts/linkerd2/values.go` under the `Global` struct:
```
ImagePullSecrets         []map[string]string `json:"imagePullSecrets""`
```

charts/add-ons/prometheus/values.yaml
c-n-c: @apleb done

charts/add-ons/tracing/values.yaml
c-n-c: @apleb done

charts/add-ons/grafana/templates/grafana-rbac.yaml
c-n-c: @apleb, @Pothulapati  done

Pothulapati:
@c-n-c I think we are in the right direction with defining `image-pull-secrets` tpl again for `cni`, as its config structure is different i.e it does not have `global` while others do. So, It makes sense to have a separate tpl for CNI. I also see that the PR is failing because of `gofmt`, Can you please run gofmt on top of the project so that it follows the required structure?

@alpeb Hmm. even though we are adding that property to `Values` it's still not usable through the CLI, right? :thinking: unless we add a relevant CLI `installOption` field.

alpeb:
> @alpeb Hmm. even though we are adding that property to `Values` it's still not usable through the CLI, right?  unless we add a relevant CLI `installOption` field.

Correct not usable in the CLI, but we still want the CLI to pick the default values set in `values.yaml`. So if in the future we decide to change any default in `values.yaml`, the CLI will pick the change as well, even if there's no flag for it. This can be tested by changing the default, recompiling the CLI, and running `linkerd install` again to see the change was picked up.

alpeb:
@c-n-c please reset `actual.Global.ImagePullSecrets = nil` in `TestNewValues` for the unit test to pass :wink:

alpeb (approved):
Thanks @c-n-c , this looks good to me.

alpeb:
@zaharidichev do you mind giving a look at this one?

zaharidichev (approved):
LGTM

alpeb:
Thanks @c-n-c :+1: 
I took the liberty of fixing the merge conflicts before merging to main.

PR #4903, opened 2020-08-21
add omitempty json field tag to Values struct

Pothulapati:
This PR adds `omitempty` json feild tag to `Values` struct so that they are not marshalled out as null values. This should not change anything but will come to use once we marshal `Values` to store state(part of min control plane)


Pothulapati:
@adleong Yep, Totally! I was trying to see what works and what does not, had to push this up so as to see what CI feels about this! :)

Pothulapati:
Closing in favor of #4911

Issue #4914, opened 2020-08-26
Parent issue for Simplifying State Management


Pothulapati:
This is a parent issue to track the current work happening on Simplifying State Management as per https://github.com/linkerd/rfc/pull/35

---
- [ ]  Add a new linkerd-config-overrides secret
- [ ] Update the control-plane components to be able to get most configuration through
flags, and remove the dependency on the current linkerd cm.
- [ ] Remove the generation of `linkerd-config` cm.
- [ ] Update the upgrade flow to use the secrets, and give it priority over the
current `linkerd-config`.
- [ ] Update the check workflow to understand the current cluster state from the
new format.
- [ ] Update Add-ons individually to have its own configuration resource, and
update CLI for upgrades and checks accordingly.

PR #4915, opened 2020-08-26
Remove dependency of linkerd-config for control plane components

Pothulapati:
Part of #4914 

This PR removes the dependency of mounting `linkerd-config` into control
plane components by making all that information passed through CLI
flags.`


controller/cmd/destination/main.go
```diff
@@ -43,23 +43,16 @@ func Main(args []string) {
...
-	trustDomain := ""
 	if *disableIdentity {
 		log.Info("Identity is disabled")
 	} else {
-		trustDomain = global.GetIdentityContext().GetTrustDomain()
-		if err != nil || trustDomain == "" {
-			trustDomain = "cluster.local"
-			log.Warnf("failed to load trust domain from global config: [%s] (falling back to %s)", err, trustDomain)
+		if err != nil || *trustDomain == "" {
+			log.Warnf("failed to load trust domain from global config: [%s] (falling back to %s)", err, *trustDomain)
```
adleong, line 50: this warning should be updated since we don't load from global config anymore.

controller/cmd/identity/main.go
```diff
@@ -42,6 +40,13 @@ func Main(args []string) {
 	addr := cmd.String("addr", ":8080", "address to serve on")
 	adminAddr := cmd.String("admin-addr", ":9990", "address of HTTP admin server")
 	kubeConfigPath := cmd.String("kubeconfig", "", "path to kube config")
+	controllerNS := cmd.String("controller-namespace", "", "namespace of the linkerd control plane")
```
adleong, line 43: I think this is just specifies which namespace to emit kubernetes events to, right?  We could just call the flag `--namespace`.  or maybe even get the current namespace from the downward api.

controller/cmd/identity/main.go
```diff
@@ -42,6 +40,13 @@ func Main(args []string) {
...
+	identityScheme := cmd.String("identity-scheme", "", "scheme of the identity")
```
adleong, line 44: this description is not very helpful.  can we have a flag help that is more informative?

controller/cmd/destination/main.go
```diff
@@ -43,23 +43,16 @@ func Main(args []string) {
...
 		}
 	}
 
-	clusterDomain := global.GetClusterDomain()
-	if err != nil || clusterDomain == "" {
-		clusterDomain = "cluster.local"
-		log.Warnf("failed to load cluster domain from global config: [%s] (falling back to %s)", err, clusterDomain)
+	if err != nil || *clusterDomain == "" {
+		log.Warnf("failed to load cluster domain from global config: [%s] (falling back to %s)", err, *clusterDomain)
```
adleong, line 55: ditto

controller/cmd/identity/main.go
```diff
@@ -2,33 +2,31 @@ package identity
...
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	v1machinery "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
 )
 
 // TODO watch trustAnchorsPath for changes
```
adleong, line 32: I think these TODOs no longer apply

adleong:
It looks like the golden files need to be updated.

Additionally there are some unit test failures related to the way that clock skew allowance is rendered during upgrade.  It looks like this has revealed an existing bug where when the clock skew allowance is read from the configmap during upgrade, it is rendered as a protobuf duration instead of as a go duration:

```
IssuanceLifetime:    idctx.GetIssuanceLifetime().String(),
```

I think we need to first convert it to a go duration in order to render it correctly.

cli/cmd/upgrade.go
```diff
@@ -587,15 +588,25 @@ func (options *upgradeOptions) fetchIdentityValues(k kubernetes.Interface, idctx
 		}
 	}
 
+	clockSkewDuration, err := ptypes.Duration(idctx.GetClockSkewAllowance())
```
Pothulapati, line 591: This was causing the upgrade tests to fail, as protobuf's `duration` is different from `Duration` in golang, we have to manually convert that using this func, if we are using the value directly. I think this wasn't a problem previously because we were again converting it back into protobuf and using it only there.

cli/cmd/upgrade_test.go
```diff
@@ -186,6 +189,11 @@ func TestUpgradeOverwriteIssuer(t *testing.T) {
 			if id == "ConfigMap/linkerd-config" {
 				continue
 			}
+
+			if id == "Deployment/linkerd-identity" {
+				// skip checking identity as `identity-trust-anchor-pem` is overridden, which is passed as a CLI flag
```
adleong, line 194: if we wanted to take this a step further, we could verify that that the path is `{spec, containers, *, spec, args, *}` or whatever and that the `diff.b` is `issuerCerts.caCrt`.

controller/cmd/identity/main.go
Pothulapati: Hmm, I went with `--controller-namespace` as every other component also has it, and helps with consistency?

Pothulapati:
Updated this PR to remove the `linkerd-config` dependency even for proxy-injector.

cli/cmd/upgrade_test.go
Pothulapati: Great suggestion! Done :+1:

controller/api/public/grpc_server.go
```diff
@@ -231,22 +223,6 @@ func (s *grpcServer) SelfCheck(ctx context.Context, in *healthcheckPb.SelfCheckR
 	return response, nil
 }
 
-func (s *grpcServer) Config(ctx context.Context, req *pb.Empty) (*configPb.All, error) {
```
Pothulapati, line 234: This API is also removed, as there will be no `linkerd-config` to directly retrieve and give out configuration

controller/cmd/proxy-injector/main.go
```diff
@@ -1,19 +1,147 @@
...
+	proxyInitImagePullPolicy := cmd.String("proxy-init-image-pull-policy", "", "linkerd proxy init image's pull policy")
+	controlPort := cmd.Uint64("control-port", 0, "proxy port to use for control")
+	ignoreInboundPorts := cmd.String("ignore-inbound-ports", "", "ports and/or port ranges (inclusive) that should skip the proxy and send directly to the application")
+	ignoreOutboundPorts := cmd.String("ignore-outbound-ports", "", "outbound ports and/or port ranges (inclusive) that should skip the proxy")
+	inboundPort := cmd.Uint64("inbound-port", 0, "proxy port to use for inbound traffic")
+	adminPort := cmd.Uint64("admin-port", 0, "proxy port to serve metrics on")
+	outboundPort := cmd.Uint64("outbound-port", 0, "proxy port to use for outbound traffic")
+	proxyUID := cmd.Int64("proxy-uid", 0, "run the proxy under this user ID")
+	proxyLogLevel := cmd.String("proxy-log-level", "", "log level for the proxy")
+	disableExternalProfiles := cmd.Bool("disable-external-profiles", false, "disable service profiles for non-Kubernetes services")
```
adleong, line 44: should invert this flag to make it affirmative? `enable-external-profiles`?

controller/cmd/proxy-injector/main.go
```diff
@@ -1,19 +1,147 @@
...
+	proxyVersion := cmd.String("proxy-version", "", "tag to be used for the Linkerd proxy images")
+	proxyInitImageVersion := cmd.String("proxy-init-image-version", "", "linkerd init container image version")
+	debugImageName := cmd.String("debug-image", "", "linkerd debug container image name")
+	debugImagePullPolicy := cmd.String("debug-image-pull-policy", "", "docker image pull policy for debug image")
+	debugImageVersion := cmd.String("debug-image-version", "", "linkerd debug container image version")
+	destinationGetNetworks := cmd.String("destination-get-networks", "", "network ranges for which the Linkerd proxy does destination lookups by IP address")
+	proxyLogFormat := cmd.String("proxy-log-format", "", "log format (`plain` or `json`) for the proxy")
+	outBoundConnectTimeout := cmd.String("outbound-connect-timeout", "", "maximum time allowed for the proxy to establish an outbound TCP connection")
```
adleong, line 52: weird capitalization of outbound

go.mod
```diff
@@ -54,7 +54,7 @@ require (
 	golang.org/x/sys v0.0.0-20200124204421-9fbb57f87de9 // indirect
 	golang.org/x/tools v0.0.0-20191009213438-b090f1f24028
 	google.golang.org/grpc v1.29.1
-	google.golang.org/protobuf v1.24.0
+	google.golang.org/protobuf v1.25.0
```
adleong, line 57: why do we need to upgrade this?

Pothulapati: Not needed, but happened automatically as I installed the latest version of `protoc-gen-go`. So reverted

controller/cmd/proxy-injector/main.go
Pothulapati: Yeah, makes sense as we are using `enable-external-profiles` in the CLI too.

cli/cmd/testdata/install_addon_control-plane.golden
```diff
@@ -1065,6 +1052,41 @@ spec:
...
+        - -identity-trust-anchors-pem=LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJZRENDQVFlZ0F3SUJBZ0lCQVRBS0JnZ3Foa2pPUFFRREFqQVlNUll3RkFZRFZRUURFdzFqYkhWemRHVnkKTG14dlkyRnNNQjRYRFRFNU1ETXdNekF4TlRrMU1sb1hEVEk1TURJeU9EQXlNRE0xTWxvd0dERVdNQlFHQTFVRQpBeE1OWTJ4MWMzUmxjaTVzYjJOaGJEQlpNQk1HQnlxR1NNNDlBZ0VHQ0NxR1NNNDlBd0VIQTBJQUJBQ2hwQXQwCnh0Z085cWJWdEV0REs4ME42aUNMMkh0eWYya0l2Mm01UWtKMXkwVEZRaTVoVFZlM3d0c3BKOFlwWkYwcGwzNjQKNlRpWWVYQjh0T09oSUFDalFqQkFNQTRHQTFVZER3RUIvd1FFQXdJQkJqQWRCZ05WSFNVRUZqQVVCZ2dyQmdFRgpCUWNEQVFZSUt3WUJCUVVIQXdJd0R3WURWUjBUQVFIL0JBVXdBd0VCL3pBS0JnZ3Foa2pPUFFRREFnTkhBREJFCkFpQlEvQUF3RjhrRzhWT21SU1VUUGFrU1NhL040bXFLMkhzWnVoUVhDbWlaSHdJZ1pFekk1RENrcFU3dzNTSXYKT0xPNFpzazFYckdaSEdzbXlpRXl2WUY5bHBZPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
+        - -identity-scheme=linkerd.io/tls
+        - -omit-webhook-side-effects=false
+        - -cluster-domain=cluster.local
+        - -proxy-image=gcr.io/linkerd-io/proxy
+        - -proxy-image-pull-policy=IfNotPresent
+        - -proxy-init-image=gcr.io/linkerd-io/proxy-init
+        - -proxy-init-image-pull-policy=IfNotPresent
+        - -control-port=4190
+        - -ignore-inbound-ports=
```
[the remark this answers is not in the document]
Pothulapati: @adleong Good Question! I did some experimentation, and it actually takes it as a empty string and does not apply the default values (though are default values are also ""). Updated the string args to use a function which checks and emits only if the arg is non-empty.

charts/partials/templates/_helpers.tpl
```diff
@@ -59,3 +59,16 @@ For example "11,22,55" will become{"portRange":11},{"portRange":22},
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Returns an arg, only if the field value is not empty.
+This has to be used only with strings, as false is also considered as empty.
+*/}}
+{{- define "partials.CheckArgIfEmpty" -}}
```
adleong, line 67: I think naming this `FlagIfNotEmpty` would be more clear

adleong (approved):

alpeb:
Nit: can you use `list` instead of `tuple`? They're equivalent but only the former appears in the docs.

adleong:
I think this will be simpler if we do https://github.com/linkerd/linkerd2/issues/5008 first.  Then we can use that in inject instead of loading the protobuf config.

Pothulapati:
@adleong :+1: waiting for the new configMap PR's to merge first! :)

adleong:
As we discussed, I think it would be helpful to narrow this PR to just focus on inject or to split the inject portions of this PR into a separate PR.

cli/cmd/upgrade_test.go
```diff
@@ -202,6 +204,13 @@ func TestUpgradeOverwriteIssuer(t *testing.T) {
 				// Trust root has changed.
 				continue
 			}
+
+			if id == "Deployment/linkerd-identity" || id == "Deployment/linkerd-proxy-injector" {
+				if !pathMatch(diff.path, []string{"spec", "template", "spec", "containers", "*", "args", "*"}) || strings.TrimPrefix(diff.b.(string), "-identity-trust-anchors-pem=") != issuerCerts.ca {
```
alpeb, line 209: - If there's a diff outside the container args then fail
- If there's a diff in some of the containers arg, check if that arg's value differs from `issuerCerts.ca`. What if the diff is not for that particular arg?

controller/api/public/grpc_server.go
Pothulapati: As this `Config` returns the older format which we do not intend to support, so removing this in the PR should be the right solution here.

alpeb (approved):
This looks ok to me, I just had that test question above :+1:

adleong (commented):
This looks pretty good.

I notice that we're using `partials.FlagIfNotEmpty` in most places to skip rendering the flag if it's empty.  But, will these values ever be empty?  I believe they all should have default values from the chart or are required values (such as the trust anchors PEM).  I think it probably makes more sense to assume that all defaulting logic has already been applied by the time rendering happens and that all of these values have been populated.  This way we're not spreading the defaulting logic between the values generation and the control plane component flags.

Pothulapati:
@adleong That makes sense, and most fields used here are not empty. Also, The reason for adding that helper check function was for proxy-injector, as it had fields that could be `nil` but now it's not needed anymore as that will be a separate PR.

go.sum
```diff
@@ -259,6 +259,7 @@ github.com/golang/protobuf v1.4.0-rc.1.0.20200221234624-67d41d38c208/go.mod h1:x
 github.com/golang/protobuf v1.4.0-rc.2/go.mod h1:LlEzMj4AhA7rCAGe4KMBDvJI+AwstrUpVNzEA03Pprs=
 github.com/golang/protobuf v1.4.0-rc.4.0.20200313231945-b860323f09d0/go.mod h1:WU3c8KckQ9AFe+yFwt9sWVRKCVIyN9cPHBJSNnbL67w=
 github.com/golang/protobuf v1.4.0/go.mod h1:jodUvKwWbYaEsadDk5Fwe5c77LiNKVO9IDvqG2KuDX0=
+github.com/golang/protobuf v1.4.1 h1:ZFgWrT+bLgsYPirOnRfKLYJLvssAegOj/hgyMFdJZe0=
```
adleong, line 262: any reason why this needs to be added?

adleong (approved):
I think the change to `go.sum` can be reverted.  Once that is done, I think this is good to go.

go.sum
Pothulapati: This is `go.sum` which is edited/used by only the go compiler, and this was done automatically. probably because some code paths do not depend on the protobuf pkg anymore. Also, Running `go mod tidy` on this branch removed some stuff.

PR #4916, opened 2020-08-26
edge-20.8.4

zaharidichev:
## edge-20.8.4

* Fixed a problem causing the `enable-endpoint-slices` flag to not be persisted
  when set via `linkerd upgrade` (thanks @Matei207!)
* Removed SMI-Metrics templates and experimental sub-commands
* Use `--frozen-lockfile` to avoid accidental update of dashboard JS
  dependencies in CI (thanks @tharun208!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Changes
 
+## edge-20.8.4
+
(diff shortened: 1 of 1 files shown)


zaharidichev:
Thanks for the LGTMs. Will update the proxy version and carry on with the edge tomorrow

PR #4923, opened 2020-08-28
Make destination and identity services headless

### LINKED ISSUE #4912: Make proxy controller services headless
Following https://github.com/linkerd/linkerd2-proxy/commit/6962794b1ef3e0d17a6ece16e581bffb1dda4060, the proxy now supports SRV-based controller discovery. In order to take advantage of this, the control plane's services should be exposed as headless services, so that no load balancer IPs are allocated (as implemented in https://github.com/linkerd/linkerd2/commit/17e6490bf09d4848381eff112e3fca495f47ecfe).

Before merging this change, we should test that an upgrade can be performed gracefully. Specifically:

1. If a cluster uses load balancer IP services (i.e. as in the current edge release), `linkerd upgrade` applies cleanly. ISTR that it's not necessarily possible to modify services with LB IPs...
2. Following a `linkerd upgrade` older proxies that do not perform SRV lookups continue to work as expected. I.e., we cannot require a full, lock-step data plane upgrade with this service change.

If we can't meet either of these requirements, we'll need to consider a more staged/incremental upgrade plan (probably introducing _new_ services).

---
Fixes https://github.com/linkerd/linkerd2/issues/4624 https://github.com/linkerd/linkerd2/issues/4674

zaharidichev:
With the proxy changes introduced in https://github.com/linkerd/linkerd2-proxy/pull/594, we can resolve control plane components via SRV records. For that purpose, we need to change the identity and dst services to be headless (i.e. clusterIp: None). The problem is that the clusterIp field is immutable. For that reason, this PR adds two new services `linkerd-dst-headless` and `linkerd-identity-headless`.  We should keep the linkerd-identitiy and linkerd-dst. The plan is to (potentially) switch back the names as we remove the non headless services in a later release.

Note: for this PR to be green, we need the a new version of the proxy

Fix #4912


adleong:
What's the long term upgrade story?  Something like this:

Step 1: When upgrading to 2.9, we add the headless services and switch the proxies to use them
Step 2: When upgrading to 2.10, we remove the unused cluster ip services
Step 3: When upgrading to 2.11, we re-add the services with their original names (but now headless) and switch the proxies to use them
Step 4: When upgrading to 2.12, we remove the unused services with the "headless" names

Alternatively, we could skip the 3rd and 4th step and just keep the "headless" names forever.  Or we could prompt users to manually delete the unused services after the upgrade is complete, which lets us skip steps 2 and 4.

Thoughts?

zaharidichev:
@adleong My plan was to do the steps outlined and end up where we are name wise. However, I am also fine with just skipping 3 and 4 and leaving the services named as headless. I guess we have the optionality to decide that at a later stage. What do you think is better ?

olix0r:
Related https://github.com/linkerd/linkerd2/pull/4932

adleong:
Seems reasonable to me.  These services aren't really user-facing so if the names are a little awkward, it's not the end of the world.

alpeb (approved):
LGTM :+1:

PR #4924, opened 2020-08-28
Perform TLS checks on injector, sp validator and tap

### LINKED ISSUE #4889: Extend the 'check' command to validate webhooks certificate validity period
Some users have recently reported downtime due to errors caused by the expiry of the webhooks certificates and CA bundles. We should extend the `linkerd check` command to check the validity period of the webhook certs, similar to the existing checks for the issuer cert as seen [here](https://github.com/linkerd/linkerd2/blob/ac2bfb387b9fcfa2ff1a29dda4e0ab74a47ea60f/pkg/healthcheck/healthcheck.go#L979-L1000).

The certs and CA bundles of the webhooks are found in the following objects:

1. `secret/linkerd-proxy-injector-tls`
1. `secret/linkerd-sp-validator-tls`
1. `secret/linkerd-tap-tls`
1. `mutatingwebhookconfiguration/linkerd-proxy-injector-webhook-config`
1. `validatingwebhookconfiguration/linkerd-sp-validator-webhook-config`
1. `apiservice/v1alpha1.tap.linkerd.io`

zaharidichev:
This PR adds checks for the certificats of Tap, Sp Validator and Proxy injector. Still need to make tests pass and figure out some details around tha naming of things. Good news is we reuse a lot of the code here

UPDATE: 

Simplified the checks and now the output looks like: 

```
linkerd-identity
----------------
√ certificate config is valid

```

Fix #4889

### DIFF
--- a/test/integration/testdata/check.golden
+++ b/test/integration/testdata/check.golden
@@ -40,6 +40,15 @@ linkerd-identity
 √ issuer cert is valid for at least 60 days
 √ issuer cert is issued by the trust anchor
 
+linkerd-webhooks-tls
+--------------------
+√ proxy-injector webhook has valid cert
+√ sp-validator webhook has valid cert
+
+linkerd-tap-tls
+---------------
+√ tap has valid cert
+
 linkerd-api
 -----------
 √ control plane pods are ready
--- a/test/integration/testdata/check.golden
+++ b/test/integration/testdata/check.golden
@@ -40,14 +40,11 @@ linkerd-identity
 √ issuer cert is valid for at least 60 days
 √ issuer cert is issued by the trust anchor
 
-linkerd-webhooks-tls
---------------------
+linkerd-webhooks-and-apisvc-tls
+-------------------------------
+√ tap API server has valid cert
 √ proxy-injector webhook has valid cert
 √ sp-validator webhook has valid cert
-
-linkerd-tap-tls
----------------
-√ tap has valid cert
 
 linkerd-api
 -----------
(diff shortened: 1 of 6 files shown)


adleong:
Some sample output in the PR description would be helpful in making this easier to review.

test/integration/testdata/check.golden
```diff
@@ -35,10 +35,37 @@ linkerd-identity
...
-√ issuer cert is using supported crypto algorithm
-√ issuer cert is within its validity period
-√ issuer cert is valid for at least 60 days
-√ issuer cert is issued by the trust anchor
+√ cert is using supported crypto algorithm
+√ cert is within its validity period
+√ cert is valid for at least 60 days
+√ cert is issued by the trust anchor
+
+linkerd-tap-tls
```
adleong, line 43: This feels a bit verbose for the successful case.  What about condensing this into a single check for each webhook that does all of the validation and returns the specific failure in the case of an error?  Something like this:

```
webhooks
----------
√ sp-validator webhookd has valid cert
```

A side-benefit of this is that we can fetch and use the certificate within the same check which means we no longer need to store it on the health check struck.

zaharidichev: Can the tap TLS check be grouped under webhooks. Not too sure about that one

adleong: `webhooks and apiservices` 🤣 

Not sure about the naming, but it makes conceptual sense to me to group these together.  They are each isolated trust chains with a caBundle given to the kubernetes API.

adleong (approved):

Pothulapati (approved):
LGTM! Just to confirm, This should work with external issuer certs too right? as they generate the secrets in that case and we still validate them?

zaharidichev:
@Pothulapati I think it will require a tiny bit of a change to get it working with external issuers as the type of secret is different..

PR #4941, opened 2020-09-03
Use latest config secret for CLI functionality

Pothulapati:
This PR updates the current CLI functions to use the latest config overrides secret
resource i.e #4911 . This is done by replacing the FetchLinkerdConfigMap
with a newer FetchLinkerdConfiugration which
- if the newer secret is present,  returns the full values i.e default + overridden
- otherwise, converts the older ConfigMap i.e `config.All` into a Values struct and returns it.

This does not touch the install/upgrade flow!
This also moves the mergeRaw functionality outside to the values, to be reused.


alpeb (commented):
@Pothulapati was the `healthcheck.FetchLinkerdConfigMap()` call in `heartbeat.go` left till the uuid issue gets resolved?

adleong:
I've been thinking about how to access the configuration in the case of a helm install.  We could certainly access the helm release secret to get the values but relying on the internal helm representation is potentially brittle.  The representation could change between helm versions without warning since this is not a public api.

Instead, we can save the fully hydrated values into a secret of our own so that we can own the representation.  This is redundant with what helm does, but it's under our control.

In total we would have:

`Secret/linkerd-config-overrides`: specially rendered by the `linkerd install` CLI command.  contains only the overridden values.  not present for helm installs.  only used by `linkerd upgrade`.

`Secret/linkerd-config`: part of the Linkerd2 chart and is therefore rendered by both `linkerd install` and by Helm install.  contains the full values.  only used by the CLI when needing to know the runtime configuration (e.g. check, etc.)

Open question: is it possible to define this secret as a helm template?  What does `b64enc` do if you give it a dict (the full values context)?  Does it convert the dict to yaml and then b64 encode that?  Is there some way to accomplish this?

Pothulapati:
Closing in favor of #5023, as this went through multiple iterations.

PR #4944, opened 2020-09-04
Fixed helm cleanup in `./bin/test-cleanup`

alpeb:
`./bin/test-cleanup` was trying to remove the
resources with the label `linkerd.io/is-test-helm` which we're not
using. Instead, we simply call `helm delete` on the appropriate helm
releases.

This is required for a clean cleanup after the ARM integration test, whose
cluster is just cleaned by this script at the end and is not torn down.

### DIFF
--- a/bin/test-cleanup
+++ b/bin/test-cleanup
@@ -3,6 +3,7 @@
 set -eu
 
 k8s_context=${1:-""}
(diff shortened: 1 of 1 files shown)


Pothulapati (approved):
LGTM, :+1:  on using the same stuff that performed the install

ihcsim (approved):

PR #4951, opened 2020-09-09
tracing: Move default values into addon-chart

Pothulapati:
This branch updates the tracing add-on's values into their own chart's values.yaml
(just like grafana and prometheus). This prevents them from being saved into
`linkerd-config-addons` where only the overridden values are stored. Thus allowing
us to change the defaults.

This also
-  Updates the check command to fall back to default values, if there are no
overridden name fields.
- Updates jaeger to `1.19.2`


charts/linkerd2/templates/web.yaml
```diff
@@ -72,7 +72,7 @@ spec:
         - -grafana-addr=linkerd-grafana.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:3000
         {{- end}}
         {{- if .Values.tracing.enabled }}
-        - -jaeger-addr={{.Values.tracing.jaeger.name}}.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:16686
+        - -jaeger-addr=linkerd-jaeger.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:16686
```
alpeb, line 75: Sorry I can't recall how this works...  The drawback of moving the values to the addon's `values.yaml` file is that they're no longer visible in the linkerd2 charts like `.Values.global.namespace` here? If that's the case shouldn't we remove the option to set the Jaeger name given it's gonna be hard-coded here?

Pothulapati: Yep, That makes sense and we removed `.name` fields for grafana too for the same reason.

alpeb (approved):

ihcsim (approved):

PR #4952, opened 2020-09-09
tests: Add CNI deep integration tests

### LINKED ISSUE #3944: Add an integration test for CNI
It should be quite easy to add an integration test for CNI. In light of some recently uncovered problems (#3941), this will be quite useful. What we can do is follow the same logic we have for testing the external cert issuer and apply it to the CNI plugin, effectivelly spinning a separate cluster.

Pothulapati:
Fixes #3944

This PR adds a new test, called cni which installs the Linkerd CNI
plugin and performs the current integration tests on top, thus
validating various Linkerd features when CNI is enabled

This is different from the CNI integration tests that we run in
cloud integration which perform the CNI level integration tests.


alpeb (commented):
Thanks for taking on this @Pothulapati . I agree it's important to run the full deep tests when using CNI :+1: 
Just one small nit: can you call the test `cni-deep` to emphasize it runs the full suite, just like `deep` and `helm-deep`?

OTOH, I have experienced a lot of flakiness when testing CNI with KinD. Maybe this is what we're observing in the failures below?

alpeb:
Ok I remember now using `linkerd install-cni --use-wait-flag` makes the cluster converge to an ok state after a few pod restarts.

test/integration/install_test.go
```diff
@@ -163,6 +163,40 @@ func TestRetrieveUidPreUpgrade(t *testing.T) {
...
+func TestInstallCNIPlugin(t *testing.T) {
+	if !TestHelper.CNI() {
+		return
+	}
+
+	// install the CNI plugin in the cluster
+	var (
+		cmd  = "install-cni"
+		args = []string{
+			"--dest-cni-net-dir", "/home/kubernetes/bin",
```
alpeb, line 175: Can you please add a comment explaining why this is needed?

alpeb:
I can advance further in the tests if I don't include the `--dest-cni-net-dir` flag. Also `linkerd check` will require a new golden file given the appearance now of the "linkerd-cni-plugin" section.

zaharidichev:
@Pothulapati Last time I tried this approach I had some problems with the fact that the newer version of the kind network plugin (which I think is what is being used in the kind version running on CI) is a bit simplistic and does not support chaining of CNI plugins. That might have changed though. My latest conclusion was that we need to have this test be a Cloud integration test rather than a Kind one. I think this is what @alpeb is refering to when saying that he observed flakiness

test/integration/inject/inject_test.go
```diff
@@ -268,10 +268,10 @@ func TestInjectAutoAnnotationPermutations(t *testing.T) {
 				if containers[0].Name != k8s.ProxyContainerName && containers[1].Name != k8s.ProxyContainerName {
 					testutil.Fatalf(t, "expected %s container in pod %s/%s, got %+v", ns, pods[0].GetName(), k8s.ProxyContainerName, containers[0])
 				}
-				if len(initContainers) != 1 {
+				if len(initContainers) != 1 && !TestHelper.CNI() {
 					testutil.Fatalf(t, "expected 1 init container for pod %s/%s, got %d", ns, pods[0].GetName(), len(initContainers))
 				}
-				if initContainers[0].Name != k8s.InitContainerName {
+				if initContainers[0].Name != k8s.InitContainerName && !TestHelper.CNI() {
```
alpeb, line 274: You might wanna do  `!TestHelper.CNI()` to short-circuit and avoid getting a panic when checking `initContainers[0]`

Pothulapati:
@zaharidichev I was able to get CNI working with Kind with default configs. This can be verified as `initContainer` checks had to be skipped while running the deep integration tests on top.

But as you mentioned, I also see the importance of running these CNI integration tests on a cloud provider so that we can have more concrete conformance but currently we run cloud integration tests only for push i.e, not PR's. I'm guessing this should be part of the PR workflow if its added right?

alpeb (approved):
Sure, we can tackle in a separate PR a new cloud integration test running under CNI, in parallel to the regular one.
@Pothulapati I'm gonna retrigger this build a few more times, just to ensure there's no flakiness.

alpeb:
Got a little confused above. To clarify, I think it makes sense to add the extra CNI cloud integration test (GKE) in the `cloud_integration.yml` workflow, not in `kind_integration.yml`. The test that was added in the current PR should be enough for the PR workflow IMO.

Pothulapati:
@alpeb I will create a separate PR for the GKE CNI test under `cloud_integrations.yaml`

test/integration/install_test.go
```diff
@@ -163,6 +163,46 @@ func TestRetrieveUidPreUpgrade(t *testing.T) {
...
+		testutil.AnnotatedFatalf(t, "'kubectl apply' command failed",
+			"'kubectl apply' command failed\n%s", out)
+	}
+
+	// perform a linkerd check with --linkerd-cni-enabled
+	timeout := time.Minute
+	err = TestHelper.RetryFor(timeout, func() error {
+		out, stderr, err = TestHelper.LinkerdRun("check", "--pre", "--linkerd-cni-enabled")
+		if err != nil {
+			return fmt.Errorf("'linkerd check' command failed\n%s\n%s\n%vs", out, stderr, err)
```
alpeb, line 197: There's an extra `s` at the end of this message

alpeb (commented):
@Pothulapati `linkerd check` already contains a retry mechanism that shouldn't need the `TestHelper.RetryFor` that you added here. I think the problem lies actually here:
https://github.com/linkerd/linkerd2/blob/6d67b84447de9de1eb456e6ea976eaf9a3121a5e/pkg/healthcheck/healthcheck.go#L894-L910
where `hc.cniDaemonSet` is not refreshed, so `scheduled := hc.cniDaemonSet.Status.DesiredNumberScheduled` takes always the same value even if the DS status changed. So we'd need to fetch that again as the `cni plugin DaemonSet exists` test does, and probably get rid of the `hc.cniDaemonSet` property that doesn't make sense to keep as state.

Pothulapati:
@alpeb 
reg retry mechanism, I've seen that we are using the same retry thing in another place for the normal linkerd check. and hence used the same here too
https://github.com/linkerd/linkerd2/blob/main/test/integration/install_test.go#L616

> probably get rid of the hc.cniDaemonSet property that doesn't make sense to keep as state.

It feels similar to `hc.controlPlanePods` where each test loads stuff there and will have re-load if they want to read the latest data

Pothulapati:
These tests should pass once #4974 is merged

alpeb (commented):
@Pothulapati What do you think about leaving the `cni-calico-deep` test while getting rid of the `cni-deep` one? The latter would be testing CNI for the particular case of KinD's network plugin which I think nobody uses in production...

Pothulapati:
@alpeb Hmm, That makes sense i.e to remove the normal `cni-deep` test which only checks the Linkerd CNI but I was thinking as Linkerd CNI can be used without other CNI's when users would want to give `NET_ADMIN` capability to initContainers. So, It might make sense to still have it? :thinking:

Pothulapati:
@alpeb The CNI Integration **with Calico** needs the additional kind configuration applied through `./cni-calico.yaml` which disables the default kind CNI

.github/workflows/kind_integration.yml
```diff
@@ -104,6 +104,7 @@ jobs:
         - uninstall
         - upgrade-edge
         - upgrade-stable
+        - cni-calico-deep
```
alpeb, line 107: Don't forget to add this into `release.yml` as well :wink:

test/integration/install_test.go
```diff
@@ -842,8 +930,10 @@ func TestLogs(t *testing.T) {
 }
 
 func TestEvents(t *testing.T) {
-	for _, err := range testutil.FetchAndCheckEvents(TestHelper) {
-		testutil.AnnotatedError(t, "Error checking events", err)
+	if !TestHelper.Calico() {
+		for _, err := range testutil.FetchAndCheckEvents(TestHelper) {
+			testutil.AnnotatedError(t, "Error checking events", err)
+		}
```
alpeb, line 936: Unknown events appearing because of Calico?

test/integration/install_test.go
```diff
@@ -163,6 +163,86 @@ func TestRetrieveUidPreUpgrade(t *testing.T) {
...
+			return fmt.Errorf("'linkerd check' command failed\n%s\n%s\n%v", out, stderr, err)
+		}
+		return nil
+	})
+	if err != nil {
+		testutil.AnnotatedFatal(t, fmt.Sprintf("'linkerd check' command timed-out (%s)", timeout), err)
+	}
+}
+
+func TestCalicoInstall(t *testing.T) {
```
alpeb, line 207: Wouldn't it make more sense to install Calico before linkerd's CNI plugin? (thus first `TestCalicoInstall` and then `TestInstallCNIPlugin`)

test/integration/install_test.go
Pothulapati: Yep, which makes the tests fail.

test/integration/install_test.go
Pothulapati: Yep, Correct.

test/integration/install_test.go
alpeb: Can you please add a comment there explaining these  warnings are just Calico converging to a correct state so they're safe to ignore ?(assuming that's what those messages are about)

Pothulapati: Looks like `TestEvents` are not necessary after we fixed the installation order of `Calico` and `LinkerdCNI`. Thank you for the suggestion :+1:

zaharidichev (approved):
Nice work!

PR #4953, opened 2020-09-09
Push docker images to ghcr.io instead of gcr.io

alpeb:
The `cloud_integration.yml` and `release.yml` workflows were modified to
log into ghcr.io, and remove the `Configure gcloud` step which is no
longer necessary.


alpeb:
Note that besides the changes to `cloud_integration.yml` and `release.yml`, there was a change to the `upgrade-stable` integration test so that we do `linkerd upgrade --addon-overwrite` to reset the addons settings because in `stable-2.8.1` the Grafana image was pegged to `gcr.io/linkerd-io/grafana` in `linkerd-config-addons`. This will need to be mentioned in the 2.9 upgrade notes.

Also the egress integration test has a debug container that now is pegged to the `main` tag.

Besides that, the other changes are just a global search and replace (`s/gcr.io\/linkerd-io/ghcr.io\/linkerd/`).

olix0r (approved):
+1 -- i have not tested this (is it even testable before the edge goes out?)

alpeb:
This tested fine in a fork, including the release workflow :+1:

kleimkuhler (approved):
Nice!

PR #4958, opened 2020-09-10
profiles: Eliminate default timeout

olix0r:
Currently, the profiles API sets a default request timeout of 10s for
requests to a service with a ServiceProfile; but no default timeout is
set when a profile is not present. This is surprising!

This change removes the default timeout for service profile routes.


adleong (commented):
I believe this will translate to a timeout of 0s in the protobuf.  Does the proxy interpret a timeout of 0 to mean no timeout?

olix0r:
@adleong It should™️. Will confirm before taking this out of draft.

adleong (approved):

alpeb (approved):
LGTM :+1:

PR #4960, opened 2020-09-10
inject: Configure the proxy to discover profiles for unnamed services

olix0r:
The proxy performs endpoint discovery for unnamed services, but not
service profiles.

The destination controller and proxy have been updated to support
lookups for unnamed services in linkerd/linkerd2#4727 and
linkerd/linkerd2-proxy#626, respectively.

This change modifies the injection template so that the
`proxy.destinationGetNetworks` configuration enables profile
discovery for all networks on which endpoint discovery is permitted.


hawkw (approved):

kleimkuhler (approved):
👍

PR #4967, opened 2020-09-14
Address dependabot alert for node-fetch

alpeb:
Addresses [this alert](https://github.com/linkerd/linkerd2/network/alert/web/app/yarn.lock/node-fetch/closed) by upgrading lingui from v2.9.1 to v2.9.2 which drops the dependency on OpenCollective, which was the only component depending on node-fetch.

I verified the i18n tools are still working as expected.


zaharidichev (approved):
LGTM

Pothulapati (approved):
LGTM

PR #4968, opened 2020-09-14
Add `dest-cni-bin-dir` flag in install-cni

Pothulapati:
Currently, This field has to be configured to make CNI work in
GKE clusters as thats where the binaries have to be stored. This
was configurable through Helm, but the same can be allowed through
the CLI too


Pothulapati:
This is needed in GKE, as the kubelet configures the path to be `/home/kubernetes/bin`

ihcsim (approved):
Good catch! Looks like this was accidentally removed in https://github.com/linkerd/linkerd2/pull/4742/files#diff-f3a0486183a3b7b5c3b8404ccff4dbddL105

alpeb (approved):
Thanks @Pothulapati , LGTM :+1:

PR #4983, opened 2020-09-17
Delete multicluster resources in `bin/test-cleanup`

alpeb:
When some test failed in the middle of the
`./tests/integration/install_test.go` suite, multicluster resources can
be left-over, which `./bin/test-cleanup` wasn't removing.

This was affecting the ARM integration tests, that require good cleanup
since they use a non-transient cluster.


hawkw (approved):

zaharidichev (approved):
LGTM

PR #4986, opened 2020-09-18
Only run the ARM integration tests for stable releases

alpeb:


.github/workflows/release.yml
```diff
@@ -157,22 +160,26 @@ jobs:
...
     - name: CNI tests
+      if: startsWith(github.ref, 'refs/tags/stable')
       run: |
         export TAG="$($CMD version --client --short)"
         go test -cover -race -v -mod=readonly ./cni-plugin/test -integration-tests
     - name: Test cleanup
       if: ${{ always() }}
-      run: bin/test-cleanup
+      # will fail if other steps didn't run, so ignore error
+      run: bin/test-cleanup || true
```
kleimkuhler, line 182: I think this step can also be marked with [continue-on-error](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)

kleimkuhler (approved):
Looks good. I left a comment about possibly using a GH action supported way of allowing steps to fail, but I also think the `.. || true` is fine too.

Pothulapati (approved):
LGTM.

If only there was a way to skip a full workflow, instead fo marking individual steps :\

alpeb:
> If only there was a way to skip a full workflow, instead fo marking individual steps :\

Just to clarify, one can set that condition at the job level instead of repeating at each step level, but then the job wouldn't be marked as successful and then its dependencies wouldn't run :man_shrugging:

PR #4994, opened 2020-09-23
Add support for k3d in integration tests

alpeb:
KinD doesn't support setting LoadBalancer services out of the box. It can be added with some additional work, but it seems the solutions are not cross-platform.

K3d on the other hand facilitates this, so we'll be using k3d clusters for the multicluster integration test.

The current change sets the ground by generalizing some of the integration tests operations that were hard-coded to KinD.

- Added `bin/k3d` to wrap the setup and running of a pinned version of `k3d`.
- Refactored `bin/_test-helpers.sh` to account for tests to be run in either KinD or k3d.
- Renamed `bin/kind-load` to `bin/image-load` and make it more generic to load images for both KinD (default) and k3d. Also got rid of the no longer used `--images-host` option.
- Added a placeholder for the new `multicluster` test in the lists in `bin/_test-helpers.sh`. It starts by setting up two k3d clusters.


bin/_test-helpers.sh
```diff
@@ -345,6 +373,11 @@ run_uninstall_test() {
   run_test "$test_directory/uninstall/uninstall_test.go" --uninstall=true
 }
 
+run_multicluster_test() {
+  #placeholder for now
```
kleimkuhler, line 377: 👍

bin/image-load
```diff
@@ -0,0 +1,102 @@
...
+  esac
+  shift
+done
+
+bindir=$( cd "${0%/*}" && pwd )
+
+if [ -n "$k3d" ]; then
+  if [ -n "$kind" ]; then
+    echo $k3d
+    echo "Error: --kind and --k3d can't be use simultaneously" >&2
```
kleimkuhler, line 52:
```suggestion
    echo "Error: --kind and --k3d can't be used simultaneously" >&2
```

kleimkuhler (approved):
Overall looks good.

I think it'd be worth seeing if we can change everything over to k3d at some point; there is definitely some added complexity to the scripts now with handling what tool should be used to create clusters. However, I don't think this is a blocker for being able to add the multicluster test.

kleimkuhler:
I see there are few PRs lined up to merge into this *before* merging into main. Will continue to review those, but this looks good so far.

alpeb:
> I think it'd be worth seeing if we can change everything over to k3d at some point; there is definitely some added complexity to the scripts now with handling what tool should be used to create clusters. However, I don't think this is a blocker for being able to add the multicluster test.

Agreed. I'll explore separately how easy would that be with the cases that have special KinD configs.

alpeb:
#4995 was based off of this branch and it got approved, so I merged it here. So please check #4995 for the description of the additional changes.

adleong (approved):
I agree, it would be great if we didn't need to use both kind and k3d.  Can we get an issue open for standardizing on k3d (or determining if that's not possible)?

alpeb:
> I agree, it would be great if we didn't need to use both kind and k3d. Can we get an issue open for standardizing on k3d (or determining if that's not possible)?

I've created #5007 to track that migration.

PR #4996, opened 2020-09-23
Reset service-mirror component when target's k8s API is unreachable

alpeb:
When the service-mirror component can't reach the target's k8s API, the goroutine blocks and it can't be unblocked.

This was happenining specifically  in the case of the multicluster integration test (still to be pushed), where the source and target clusters are created in quick succession and the target's API service doesn't always have time to be exposed before being requested by the service mirror.

The fix consists on no longer have `restartClusterWatcher` be side-effecting, and instead return an error. If such error is not nil then the link watcher is stopped and reset after 10 seconds.

### DIFF
--- a/controller/cmd/service-mirror/main.go
+++ b/controller/cmd/service-mirror/main.go
@@ -23,6 +23,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const linkWatchRestartAfter = 10 * time.Second
+
 var (
 	clusterWatcher *servicemirror.RemoteClusterServiceWatcher
 	probeWorker    *servicemirror.ProbeWorker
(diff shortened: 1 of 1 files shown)


controller/cmd/service-mirror/main.go
```diff
@@ -107,7 +107,14 @@ main:
 							if err != nil {
 								log.Errorf("Failed to load remote cluster credentials: %s", err)
 							}
-							restartClusterWatcher(link, *namespace, creds, controllerK8sAPI, *requeueLimit, *repairPeriod, metrics)
+							err = restartClusterWatcher(link, *namespace, creds, controllerK8sAPI, *requeueLimit, *repairPeriod, metrics)
+							if err != nil {
+								// failed to restart cluster watcher; give a bit of slack
+								// and restart the link watch to give it another try
+								log.Error(err)
+								time.Sleep(10 * time.Second)
```
zaharidichev, line 115: Nit: may want to put that in a `const`

zaharidichev (approved):
LGTM

kleimkuhler:
The base branch for this should probably be `main` since it's fixing an issue, even if it is related to adding the multicluster integration test.

kleimkuhler (approved):
Looks good 👍

alpeb:
> The base branch for this should probably be `main` since it's fixing an issue, even if it is related to adding the multicluster integration test.

Yeah, the reason for doing the daisy chain PRs was for all to pass the tests. Now that this one has been reviewed and doesn't have dependencies on the changes from the parent branches, I'll merge to main and rebase the other ones :+1:

controller/cmd/service-mirror/main.go
alpeb: will do :+1:

kleimkuhler:
@alpeb Sounds good!

Issue #4992, opened 2020-09-22
Tap isn't showing the correct mTLS status


ihcsim:
### What is the issue?

When using `curl` to send multiple requests from an injected client to an injected server, `linkerd tap` always reports mTLS isn't working for the **first call**, with the `tls=no_tls_from_remote` field in its output. Subsequent calls from the same client to the same server shows that `tls=true`.

### How can it be reproduced?

Deploy the following an injected `httpbin` backend service:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: httpbin
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  namespace: httpbin
  labels:
    app: httpbin
spec:
  ports:
  - name: http
    port: 8000
    targetPort: 80
  selector:
    app: httpbin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
  namespace: httpbin
spec:
  selector:
    matchLabels:
      app: httpbin
      version: v1
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
      annotations:
        linkerd.io/inject: enabled
    spec:
      containers:
      - image: docker.io/kennethreitz/httpbin
        imagePullPolicy: IfNotPresent
        name: httpbin
        ports:
        - containerPort: 80
```

Deploy the following injected `curl` client:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: curl
spec:
  selector:
    matchLabels:
      app: curl
  template:
    metadata:
      annotations:
        linkerd.io/inject: enabled
      labels:
        app: curl
    spec:
      containers:
      - args:
        - sleep
        - "3600"
        image: appropriate/curl
        name: curl-inject
```

Tap the `httpbin` pod:

```sh
linkerd -n httpbin tap po/httpbin-d86f58cf6-chnbk
```

Tap the `curl` pod:

```sh
linked tap po/curl-85bcfccf97-qjrzk
```

Use the `curl` pod to issue a HTTP `GET` to the `httpbin` endpoint:

```sh
kubectl exec curl-85bcfccf97-qjrzk -- curl httpbin.httpbin:8000/status/200
```

Notice that the `tap` outputs on both inbound and outbound sides show mTLS isn't working:

E.g. on the inbound side,

```sh
[...]

PR #5001, opened 2020-09-24
Make DNS canonicalization timeout configurable

zaharidichev:
This PR adds a proxy config annotation that can be used to configure the DNS canonicalization timeout value for the proxy. The reason for doing that is the fact that we have recently observed sporadic timeouts during the DNS refinement step. When dns quries time out, we are unable to do a proper dst get request and establish the proper identity. Adding the `config.linkerd.io/proxy-dns-canonicalize-timeout` should provide users with a way to tweak this timeout in case they observe such failures.

Related: #4992, https://github.com/linkerd/linkerd2-proxy/pull/672


zaharidichev:
@olix0r you mentioned that the canonicalization shall be going away soon, so not sure whether it is worth providing a command line option to tweak it. In fact it might be the case we do not even need this way of tweaking it and https://github.com/linkerd/linkerd2-proxy/pull/672 is enough.

Issue #4993, opened 2020-09-23
Upgrade client-go to avoid watch reconnection issues


olix0r:
https://github.com/kubernetes/kubernetes/issues/91073 describes an issue that prevents watches from reconnecting after the tcp stream is interrupted. We should try to pull in this fix before the next stable release.

Pothulapati:
Should be fixed by https://github.com/linkerd/linkerd2/pull/5002

PR #5002, opened 2020-09-24
Bump k8s client-go to v0.19.2

### LINKED ISSUE #4191: Upgrade client-go for k8s v1.18
Kubernetes v1.18 is due out soon. However, we will need to take care to upgrade `client-go`, as many APIs have changed (to use `context` as they should have).

Before any code changes, we need to enumerate our dependencies that have transitive `client-go` dependencies to determine an upgrade path.

Pothulapati:
Fixes #4191 #4993

This PR bumps Kubernetes client-go to the latest `v0.19.2` (We had to switch directly to `1.19` because of[ this issue](https://github.com/linkerd/linkerd2/issues/4191#issuecomment-673318145)). Bumping to `v0.19.2` required upgrading to `smi-sdk-go v0.4.1`. This also depends on https://github.com/linkerd/stern/pull/5

This consists of the following changes:

- Fix `./bin/update-codegen.sh` by adding the template path to the gen commands, as it is needed after we moved to GOMOD.
- Bump all k8s related dependencies to `v0.19.2`
- Generate CRD types, client code using the latest `k8s.io/code-generator`
- Use `context.Context` as the first argument, in all code paths that touch the k8s client-go interface


go.mod
```diff
@@ -26,50 +26,45 @@ require (
...
+	k8s.io/kube-aggregator v0.18.8
 	sigs.k8s.io/yaml v1.2.0
 )
 
 replace (
 	// to avoid the `github.com/golang/protobuf/protoc-gen-go/generator` deprecation warning
 	// (see https://github.com/golang/protobuf/issues/1104)
 	github.com/grpc-ecosystem/grpc-gateway => github.com/grpc-ecosystem/grpc-gateway v1.14.8
-	github.com/wercker/stern => github.com/linkerd/stern v0.0.0-20200331220320-37779ceb2c32
+	github.com/wercker/stern => github.com/Pothulapati/stern v0.0.0-20200924090734-6fb6701edd91
```
Pothulapati, line 69: Update this once https://github.com/linkerd/stern/pull/5 is merged.

controller/cmd/public-api/main.go
```diff
@@ -46,7 +46,9 @@ func Main(args []string) {
 	defer destinationConn.Close()
 
 	k8sAPI, err := k8s.InitializeAPI(
-		*kubeConfigPath, true,
+		context.Background(),
```
alpeb, line 49: As commented elsewhere, it might make sense to use the same context as in the shutdown() call below

controller/cmd/tap/main.go
```diff
@@ -40,7 +40,9 @@ func Main(args []string) {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
 
+	ctx := context.Background()
```
alpeb, line 43: Use the same context as the call to shutdown() below?

alpeb (commented):
This looked like a lot of work indeed! :trophy: 

The integration tests are failing because it seems we lost the client-go logs when linkerd's log level is debug...

web/main.go
```diff
@@ -78,7 +78,7 @@ func main() {
 		APIAddr:               *apiAddr,
 	})
 
-	cm, _, err := healthcheck.FetchLinkerdConfigMap(k8sAPI, *controllerNamespace)
+	cm, _, err := healthcheck.FetchLinkerdConfigMap(context.Background(), k8sAPI, *controllerNamespace)
```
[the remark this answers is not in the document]
Pothulapati: That makes sense but used the same context to create a contextWithTimeout which adds a deadline and is used as per the current logic

controller/webhook/launcher.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // Launch sets up and starts the webhook and metrics servers
-func Launch(APIResources []k8s.APIResource, metricsPort uint32, handler handlerFunc, component, subcommand string, args []string) {
+func Launch(ctx context.Context, APIResources []k8s.APIResource, metricsPort uint32, handler handlerFunc, component, subcommand string, args []string) {
```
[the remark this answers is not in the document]
Pothulapati: @alpeb I don't understand why we need to defer the cancellation in `main.go`'s as we are not creating any go-routines and hence the no need for waiting and cancellation. It seems like a sequential call, and this `Launch` funciton only returns when everything is stopped based on `<-stop`, and would just return normally back to `main.go's` where they return,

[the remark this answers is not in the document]
alpeb: Good point, makes better sense :+1:

alpeb (approved):
Awesome! :1st_place_medal:

adleong (approved):

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -831,7 +832,7 @@ func (pp *portPublisher) newPodRefAddress(endpointPort Port, endpointIP, podName
 	if err != nil {
 		return Address{}, PodID{}, fmt.Errorf("unable to fetch pod %v:%v", id, err)
 	}
-	ownerKind, ownerName := pp.k8sAPI.GetOwnerKindAndName(pod, false)
+	ownerKind, ownerName := pp.k8sAPI.GetOwnerKindAndName(context.Background(), pod, false)
```
Pothulapati, line 835: This feels like an important discussion to have, 

I had to short-circuit here, as `newPodRefAdress` is used in `add*`, `update*` etc methods, which are assigned directly to `ResourceEventHandlerFuncs` at https://github.com/linkerd/linkerd2/blob/f14b7cd88059628522668bbe38328c2d6e6699dc/controller/api/destination/watcher/endpoints_watcher.go#L153
and thus have fixed func signatures. 

As the input is passed from the client-go, we don't really have a way to get propagate context from top. Any ideas here?

_This same problem applies with `ip_watcher` too_

Pothulapati (commented):
Tested this PR on 1.15 to be sure, and Integration tests passed as expected! :)

Have some final questions before I merge this

cli/cmd/inject_test.go
```diff
@@ -358,7 +360,7 @@ type injectCmd struct {
 }
 
 func testInjectCmd(t *testing.T, tc injectCmd) {
-	testConfig := testInstallConfig()
+	testConfig := testInstallConfig(context.Background())
```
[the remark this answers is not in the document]
alpeb: Looks ok to me. We're not testing anything related to context here, so should be fine to keep it simple like this.

controller/api/destination/watcher/endpoints_watcher.go
alpeb: I guess the only way would be by putting the context in the structs, although the godoc for `context` explicitly warns against that. But in this case I think it's ok, as long as it doesn't complicate things too much.

adleong: Using `context.Background()` here seems fine to me.  The events from the informer don't pass along a context and even if they did, would it even make sense to use it?  e.g. would you want a cancellation or timeout of the informer event to timeout/cancel this?  I don't think so.

Pothulapati: Agree! @adleong

PR #5005, opened 2020-09-25
Upgrade using config overrides

adleong:
This is a major refactor of the install/upgrade code which removes the config protobuf and replaces it with a config overrides secret which stores overrides to the values struct.  Further background on this change can be found here: https://github.com/linkerd/linkerd2/discussions/4966

Note: as-is this PR breaks injection.  There is work to move injection onto a Values-based config which must land before this can be merged.

A summary of the high level changes:

* the install, global, and proxy fields of linkerd-config ConfigMap are no longer populated
* the CLI install flow now follows these simple steps:
  * load default Values from the chart
  * update the Values based on the provided CLI flags
  * render the chart with these values
  * also render a Secret/linkerd-config-overrides which describes the values which have been changed from their defaults
* the CLI upgrade flow now follows these simple stesp:
  * load the default Values from the chart
  * if Secret/linkerd-config-overrides exists, apply the overrides onto the values
  * otherwise load the legacy ConfigMap/linkerd-config and use it to updates the values
  * further update the values based on the provided CLI flags
  * render the chart and the Secret/linkerd-config-overrides as above
* Helm install and upgrade is unchanged


test/integration/install_test.go
```diff
@@ -246,16 +246,22 @@ func TestInstallOrUpgradeCli(t *testing.T) {
...
-		manifests, err := TestHelper.Kubectl("", kubeArgs...)
+		kubeArgs = append([]string{"--namespace", TestHelper.GetLinkerdNamespace(), "get"}, "secrets", "-oyaml")
+		secretManifests, err := TestHelper.Kubectl("", kubeArgs...)
 		if err != nil {
 			testutil.AnnotatedFatalf(t, "'kubectl get' command failed",
-				"'kubectl get' command failed with %s\n%s\n%s", err, manifests, kubeArgs)
+				"'kubectl get' command failed with %s\n%s\n%s", err, secretManifests, kubeArgs)
 		}
 
+		manifests := configManifests + "---\n" + secretManifests
```
alpeb, line 263: :+1:

alpeb (commented):
I did a quick pass, but looking forward to dig in deep :-)

Just a couple of things I noticed so far:
- `linkerd-config-overrides` is being produced in both `config` and `control-plane` stages, making the latter stage complain.
- `UpdateAddOnValuesFromConfig` was removed, so addons values are no longer overridden by main values, which explains why `install_custom_registry.golden` isn't overriding the custom registry for Grafana.

pkg/inject/inject.go
```diff
@@ -774,73 +783,73 @@ func (conf *ResourceConfig) applyAnnotationOverrides(values *l5dcharts.Values) {
...
 		} else {
-			values.Global.Proxy.Resources.Memory.Request = res.String()
+			values.Global.Proxy.Resources.Memory.Request = override
 		}
 	}
 
-	if override, ok := conf.getOverride(k8s.ProxyCPULimitAnnotation); ok {
-		res, err := k8sResource.ParseQuantity(override)
+	if override, ok := annotations[k8s.ProxyCPULimitAnnotation]; ok {
+		_, err := k8sResource.ParseQuantity(override)
```
Pothulapati, line 805: I think we should still have this logic right i.e to convert into k8s resource quanitites and pass them into rendering?

Also, This is the reason why `inject_test.go` has to be updated with resources in different formatting.

pkg/inject/inject.go
```diff
@@ -774,73 +783,73 @@ func (conf *ResourceConfig) applyAnnotationOverrides(values *l5dcharts.Values) {
...
 			log.Warnf("%s (%s)", err, k8s.ProxyCPULimitAnnotation)
 		} else {
-			values.Global.Proxy.Resources.CPU.Limit = res.String()
+			values.Global.Proxy.Resources.CPU.Limit = override
 		}
 	}
 
-	if override, ok := conf.getOverride(k8s.ProxyMemoryLimitAnnotation); ok {
-		res, err := k8sResource.ParseQuantity(override)
+	if override, ok := annotations[k8s.ProxyMemoryLimitAnnotation]; ok {
```
Pothulapati, line 813: Looks Good. :+1: Just to confirm, This is possible because

- Flags are directly overridden on top of default Values in `applySetFlags`
- `nsAnnotations` are overridden with `pod.meta.annoations`.

So, We have a single `annotations` struct consisting of final overrides?

pkg/inject/inject.go
adleong: I could potentially go either way on this, but it seems to me if someone puts a certain string in the annotation then that's the string we should use in the template.  It's good for us to do validation so that we can warn if it's invalid, but there's not really any reason why we need to do normalization or change the format of it.

Basically, it's easier to understand what's happening here if we just use the string we're given and avoid changing it for no reason.

pkg/inject/inject.go
adleong: Yup!

pkg/k8s/fake.go
```diff
@@ -150,7 +149,7 @@ func newFakeClientSetsFromManifests(readers []io.Reader) (
 
 			switch typeMeta.Kind {
 			case "":
-				log.Warnf("Kind missing from YAML, skipping")
```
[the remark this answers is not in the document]
adleong: We have a lot of "divider" sections that look like this:

```
---
# Some section description
---
```

Emitting a warning every time we hit one of these dividers is really noisy.  It seems fine to silently skip them.

pkg/inject/inject.go
Pothulapati: Yep, That makes sense! But would be not break existing using behavior i.e values that were set without conversion by doing this?

Pothulapati (approved):
This LGTM overall.
I will do some more testing tomorrow around flags, add-ons, etc.

alpeb:
After successfully doing `linkerd install config|k apply -f `, the next stage returns an error:
```
$ bin/linkerd install control-plane
Can't install the Linkerd control plane in the 'linkerd' namespace. Reason: Secret/linkerd-config-overrides already exists.
If this is expected, use the --ignore-cluster flag to continue the installation.
```
The solution might be to have the `renderOverridesCall` in `render()` happen only for `control-plane` or full installs.

cli/cmd/install.go
```diff
@@ -304,54 +172,50 @@ control plane. It should be run after "linkerd install config".`,
...
+			if !skipChecks {
 				// check if global resources exist to determine if the `install config`
 				// stage succeeded
-				if err := errAfterRunningChecks(options); err == nil {
+				if err := errAfterRunningChecks(values.Global.CNIEnabled); err == nil {
 					if healthcheck.IsCategoryError(err, healthcheck.KubernetesAPIChecks) {
-						fmt.Fprintf(os.Stderr, errMsgCannotInitializeClient, err)
-					} else {
 						fmt.Fprintf(os.Stderr, errMsgGlobalResourcesMissing, controlPlaneNamespace)
 					}
```
alpeb, line 181: It looks like we lost the differentiation between "can't connect to k8s" errors and "there are global l5d resources missing".

cli/cmd/install.go
```diff
@@ -304,54 +172,50 @@ control plane. It should be run after "linkerd install config".`,
...
 
-			return installRunE(cmd.Context(), options, controlPlaneStage, flags)
+			return install(cmd.Context(), os.Stdout, values, flags, controlPlaneStage)
 		},
 	}
 
-	cmd.PersistentFlags().BoolVar(
-		&options.skipChecks, "skip-checks", options.skipChecks,
-		`Skip checks for namespace existence`,
-	)
```
alpeb, line 335: This wasn't add back after the refactoring, even though there's a referrence to a `skipChecks` var

alpeb (approved):
LGTM :+1: 

As an additional cleanup, possibly as a followup PR, you can consider cleaning up the `global`, `proxy` and `install` entries in `config.yaml` and consequently being able to delete `_config.tpl`

PR #4393, opened 2020-05-14
Fix kubernetes grafana dashboard (#4380)

aimbot31:
Prometheus uses a relabel rule to change pod_name to pod

Change "pod_name" to "pod" in the grafana dashboard

Run some tests on the Grafana dashboard

Fixes #4380

Signed-off-by: Florian Davasse <aimbot31@gmail.com>

PR #5012, opened 2020-09-26
Fix kubernetes grafana dashboard (#4380)

### LINKED ISSUE #4380: Grafana dashboard isn't using the good label for querying prometheus
### What is the issue?
The dashboard "Kubernetes cluster monitoring (via Prometheus) " isn't using the good label for querying prometheus.

https://github.com/linkerd/linkerd2/blob/master/grafana/dashboards/kubernetes.json

It uses `pod_name` but i do not have this label in my prometheus.

### How can it be reproduced?
I've just installed a new cluster in 1.16.8 with 3 nodes 1 master. I deployed ArgoCD, linkerd and a microservice app (bookinfo).

### Logs, error output, etc

img

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

Status check results are √

```

### Environment

- Kubernetes Version: 1.16.8
- Cluster Environment: self-hosted deployed with kubespray
- Host OS: ubuntu 18.04
- Linkerd version: Client version: stable-2.7.1
Server version: stable-2.7.1

### Possible solution

Use `pod` or `name` instead of `pod_name`.

### Additional context

This is what i got when i query `container_cpu_usage_seconds_total` in my prometheus : 
```text
container_cpu_usage_seconds_total{beta_kubernetes_io_arch="amd64",beta_kubernetes_io_os="linux",container="POD",cpu="total",id="/kubepods/besteffort/pod0efc5b94-5155-43ec-a76d-7fef302d941a/8df0af43b56d3edc499f39a92fd4b6334e4aa754f0987c3f3049f18e5c8a4b90",image="gcr.io/google_containers/pause-amd64:3.1",instance="node1",job="kubernetes-nodes-cadvisor",kubernetes_io_arch="amd64",kubernetes_io_hostname="node1",kubernetes_io_os="linux",name="k8s_POD_argocd-server-7696cd5f89-g8sj5_argocd_0efc5b94-5155-43ec-a76d-7fef302d941a_0",namespace="argocd",pod="argocd-server-7696cd5f89-g8sj5"}
```

I think the problem is the same for `container_name`, haven't got the time to check this one.

aimbot31:
Prometheus use a relabel rule that changed since 1.16

Use "pod_name" and "pod" to avoid breaking changes.
Alse use "container" and "container_name" for the
same reasons.

Run some tests on the Grafana dashboard

Fixes #4380

Signed-off-by: Florian Davasse <florian.davasse@stack-labs.com>

### DIFF
--- a/grafana/dashboards/kubernetes.json
+++ b/grafana/dashboards/kubernetes.json
@@ -1000,10 +1000,10 @@
       "steppedLine": true,
       "targets": [
         {
-          "expr": "sum (rate (container_cpu_usage_seconds_total{image!=\"\",name=~\"^k8s_.*\",kubernetes_io_hostname=~\"^$Node$\"}[1m])) by (pod_name)",
+          "expr": "sum (rate (container_cpu_usage_seconds_total{image!=\"\",kubernetes_io_hostname=~\"^$Node$\"}[1m])) by (pod_name, pod)",
           "interval": "10s",
           "intervalFactor": 1,
-          "legendFormat": "{{ pod_name }}",
+          "legendFormat": "{{ pod_name }}{{ pod }}",
           "metric": "container_cpu",
           "refId": "A",
           "step": 10
@@ -1209,11 +1209,11 @@
           "steppedLine": true,
           "targets": [
             {
-              "expr": "sum (rate (container_cpu_usage_seconds_total{image!=\"\",name=~\"^k8s_.*\",container_name!=\"POD\",kubernetes_io_hostname=~\"^$Node$\"}[1m])) by (container_name, pod_name)",
+              "expr": "sum (rate (container_cpu_usage_seconds_total{image!=\"\",container_name!=\"POD\",container!=\"POD\",kubernetes_io_hostname=~\"^$Node$\"}[1m])) by (container_name, container, pod_name, pod)",
               "hide": false,
               "interval": "10s",
               "intervalFactor": 1,
-              "legendFormat": "pod: {{ pod_name }} | {{ container_name }}",
+              "legendFormat": "pod: {{ pod_name }}{{ pod }} | {{ container_name }}{{ container }}",
               "metric": "container_cpu",
               "refId": "A",
               "step": 10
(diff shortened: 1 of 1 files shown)


alpeb (approved):
Thanks @aimbot31 , this looks great. I could verify the offending graphs now are non-empty/not-as-empty as they were before :+1: 

(Adding ref #4393 in case more context is needed for other reviewers).

grafana/dashboards/kubernetes.json
```diff
@@ -1000,10 +1000,10 @@
       "steppedLine": true,
       "targets": [
         {
-          "expr": "sum (rate (container_cpu_usage_seconds_total{image!=\"\",name=~\"^k8s_.*\",kubernetes_io_hostname=~\"^$Node$\"}[1m])) by (pod_name)",
+          "expr": "sum (rate (container_cpu_usage_seconds_total{image!=\"\",kubernetes_io_hostname=~\"^$Node$\"}[1m])) by (pod_name, pod)",
           "interval": "10s",
           "intervalFactor": 1,
-          "legendFormat": "{{ pod_name }}",
+          "legendFormat": "{{ pod_name }}{{ pod }}",
```
Pothulapati, line 1006: Is `pod_name` still a correct label? because all I can see in the dashboard is it being statically printed.




Should we just remove it? as I see `{{ pod }}` is added everywhere it is present?

Pothulapati (commented):
Thanks for taking this up! I have a small nit around `pod_name`

grafana/dashboards/kubernetes.json
aimbot31: As suggested by alpeb in #4393 I've letted `pod` and `pod_name` in order to keep things compatible.
Can you tell me what version did you used ? Have you used kind or minikube or ... ?

Thank's for you answers

Pothulapati: @aimbot31 Ahh, That makes sense! I tried this out in kind.

Do these `json` files follow the same go templating? (They kinda look like it?) If yes? maybe we can have some kind of check to print only the `pod` when `pod_name` is empty?

cc: @alpeb

alpeb: @Pothulapati That's weird since  `pod_name` should just be empty... Although I did try after merging `main` that brings a new version of Grafana. @aimbot31 could you please merge `main` to your branch so we can give it another test?

aimbot31: Yup i don't understand this because i'm not seeing pod or pod_name, just the only value filled.

aimbot31: @alpeb Done

Pothulapati: Awesome, Works now. and I don't see `pod_name` anymore.

Pothulapati (approved):
LGTM

PR #5013, opened 2020-09-26
default install of linkerd2 interferes with apiserver reliability

drewwells:
The default installation of linkerd2 uses an opt-out model. If you
consider an existing cluster with 1000s of pods and 100s of nodes, and
add linkerd2 without customizing coniguration, it's very likely to
impact cluster reliability.

This PR adds objectSelector from k8s 1.15+ that moves the
linkerd.io/inject selector from the apiservice to the apiserver
configuration. This is an important when the linkerd2 proxy injector
is down, the pods that have no interest in linkerd will not be
prevented from pod creation.

This configuration is almost identical to istio's istio/istio#16539

### DIFF
--- a/charts/linkerd2/values.yaml
+++ b/charts/linkerd2/values.yaml
@@ -166,7 +166,11 @@ heartbeatSchedule: "0 0 * * *"
 
 # proxy injector configuration
 proxyInjector:
+  enableNamespacesByDefault: false
   externalSecret: false
+  objectSelector:
+    enabled: true
+    autoInject: false
   # if empty, Helm will auto-generate these fields
   crtPEM: |
 
--- a/charts/linkerd2/templates/proxy-injector-rbac.yaml
+++ b/charts/linkerd2/templates/proxy-injector-rbac.yaml
@@ -97,7 +97,6 @@ webhooks:
     matchLabels:
       config.linkerd.io/admission-webhooks: enabled
 {{- end }}
-{{/* Object selectors are supported in k8s 1.15+ */}}
 {{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
 {{- if semverCompare "> 1.15.0-0" $kubeTargetVersion }}
 {{- if .Values.proxyInjector.objectSelector.enabled }}
@@ -113,7 +112,8 @@ webhooks:
       "linkerd.io/inject": "true"
 {{- end }}
 {{- end }}
-{{- end }}  clientConfig:
+{{- end }}
+  clientConfig:
     service:
       name: linkerd-proxy-injector
       namespace: {{ .Values.global.namespace }}
--- a/charts/linkerd2/templates/proxy-injector-rbac.yaml
+++ b/charts/linkerd2/templates/proxy-injector-rbac.yaml
@@ -83,12 +83,37 @@ metadata:
 webhooks:
 - name: linkerd-proxy-injector.linkerd.io
   namespaceSelector:
+{{- if .Values.proxyInjector.enableNamespacesByDefault }}
     matchExpressions:
+    - key: name
+      operator: NotIn
+      values:
+      - {{ .Release.Namespace }}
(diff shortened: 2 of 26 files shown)


kd7lxl (approved):

drewwells:
None of the status checks work locally. Some of them show fail instead succeed but most show checkmarks with different colors

image

charts/linkerd2/values.yaml
```diff
@@ -166,7 +166,11 @@ heartbeatSchedule: "0 0 * * *"
 
 # proxy injector configuration
 proxyInjector:
+  enableNamespacesByDefault: false
```
grampelberg, line 169: Wouldn't this disable the injector for everyone by default?

grampelberg (commented):
Couple questions:

- The docs suggest that this works only for labels. As we use annotations, wouldn't this not work?
- What happens with `objectSelector` on 1.14 clusters?

charts/linkerd2/values.yaml
drewwells: Sort of, the linkerd proxy injector mutating webhook was called for all pod creation events. By default, it would then filter to only those pods with the linkerd.io/inject:true label k/v pair. 

This was the best option before k8s 1.15. However, if the linkerd proxy service was down for any reasons, you are unable to create pods. Provisioning tools like kops create pods for the control plane. It created a chicken and egg problem where linkerd blocked the creation of kube-apiserver pod and linkerd could not receive requests b/c of a partially functioning kube-apiserver. The same situation happens if you use cert-manager to rotate linkerd certificates.

grampelberg: The chicken/egg problem is the reason we recommend skipping `kube-system` in the HA docs (as by default this isn't actually an issue).

drewwells: Those docs can be removed by ensuring a default install of linkerd does not break pod creation for the cluster.

grampelberg: Do you mean the `--ha` install? The default install of Linkerd shouldn't break pod creation on the cluster.

drewwells: Does `--ha` change the mutatingwebhook? The issue we faced is that all pod events go to proxy injector unless specific namespace labels are included. Having the proxy injector ignore specific pods is not enough to ensure pod creations work.

grampelberg: Yes, the default MWC policy is "ignore" - if the proxy injector is unreachable for any reason, the API server will just skip it and create the pods.

drewwells: That would workaround the issue but it also would lead to inconsistent application of linkerd proxies on pods which is not ideal. We want to ensure that pods that should have linkerd get them and those pods not participating (like kube-system) are not impacted by linkerd.

If the issue is with changing the default behavior, we certainly can look at just the objectSelector. I found that Istio had taken an optin approach which ensures minimal cluster impact.

grampelberg: From my 3 previous questions, we need to answer:

- How do we preserve the current behavior of it being on by default?
- What happens when we use this on a k8s 1.14 cluster?
- Does it work? The docs suggest that this is label based, and we're annotation based.

I think it is a great addition to the `--ha` configuration as it is always tough to remember to opt-out `kube-system`. We just can't change the default install behavior.

drewwells: - The existing default install of linkerd changes the cluster and is risky for anyone to use.
- Helm has k8s version flags, which is in the PR
- Object selectors do not work on annotations. Annotations are not intended to be used for selecting objects, they are for annotating the object with additional metadata. https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects So we can not provide per pod selection in linkerd as k8s intends it without also moving the inject selection criteria to a label from annotation.

drewwells:
The changes to make linkerd more friendly to clusters requires multiple API changes. I'll open a proposal so we can find ways to address these concerns.

Issue #5009, opened 2020-09-25
Use values from linkerd-config in check and inject


adleong:
Depends on https://github.com/linkerd/linkerd2/issues/5008

The `linkerd check` and `linkerd inject` commands currently get their configuration by reading the `global`, `install`, or `proxy` files in the `ConfigMap/linkerd-config` resource.  They should read their configuration from the `values` file in this resource instead.  This will simplify the codebase overall and will allow us to remove the `global`, `install`, and `proxy` files.

alpeb:
Done in #5023 (for check)  and #5036 (for inject)

PR #5015, opened 2020-09-28
Decode durations from protobuf into golang correctly

Pothulapati:
This PR updates the retrieval of durations from `linkerd-config`
to have conversion into the normal Go duration. This is needed
as the duration types in protobuf and golang are different.

This did not break anything previously, as we were directly using
these fields only for protobuf generation with upgrades, but now that
we are trying to use these fields with the `Values` go types (with the
new config format) this causes the upgrade tests to fail.

This is in sync with the encoding part here https://github.com/linkerd/linkerd2/blob/db57611a33b00a8f005a7ac65c4eeb7658189e50/cli/cmd/install.go#L1255

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang/protobuf/ptypes"
 	pb "github.com/linkerd/linkerd2/controller/gen/config"
 	charts "github.com/linkerd/linkerd2/pkg/charts/linkerd2"
 	"github.com/linkerd/linkerd2/pkg/config"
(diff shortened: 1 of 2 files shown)


Pothulapati:
This was needed both with #4915 and #4941, So hence felt its better to make a separate PR and get this merged first.

adleong:
I don't think we need to do this since we're moving over to having these components load configuration either from flags or from serialized Values, as described in https://github.com/linkerd/linkerd2/issues/5009.

The only place where we should need to load the config protobuf is in the legacy upgrade scenario.

Pothulapati:
@adleong Hmmm, This was mainly needed for tests to pass, as it renders incorrectly causing **only the diff to fail** in `TestUpgrade*` tests.

As I understand from #5005, I could not find `protobuf.Duration` -> `golang.Duration conversion` in the` legacyUpgrade` logic here https://github.com/linkerd/linkerd2/blob/bac357d812fbd451656bcbd04177148a2e4e698b/cli/cmd/upgrade_legacy.go#L156. Is the conversion not needed? :thinking:

Pothulapati:
Closing this, as it's not needed anymore

PR #5019, opened 2020-09-28
Set FQN in profile resolutions

### LINKED ISSUE #5016: destination: Set a fully_qualified_name on profile resolutions
When the proxy does resolution by IP, it may not know the name of the name of the apex service; but this is needed to make metrics consistent. In order for the proxy to know this information, the profile resolution must include such a name if it is known.

See https://github.com/linkerd/linkerd2-proxy-api/blob/7fcd01ab736d67e6f885127a1baa40f5ea0e936e/proto/destination.proto#L140-L152

kleimkuhler:
## Motivation

Closes #5016

Depends on linkerd/linkerd2-proxy-api#44

## Solution

A `profileTranslator` exists for each service and now has a new
`fullyQualifiedName` field.

This field is used to set the `FullyQualifiedName` field of
`DestinationProfile`s each time an update is sent.

In the case that no service profile exists for a service, a default
`DestinationProfile` is created and we can use the field to set the correct
name.

In the case that a service profile does exist for a service, we still use this
field to set the name to keep it consistent.

### Example

Install linkerd on a cluster and run the destination server:

```
go run controller/cmd/main.go destination -kubeconfig ~/.kube/config
```

Get the IP of a service. Here, we'll get the ip for `linkerd-identity`:

```
> kubectl get -n linkerd svc/linkerd-identity
NAME               TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
linkerd-identity   ClusterIP   10.43.161.68   <none>        8080/TCP   4h25m
```

Get the profile of `linkerd-identity` from service name or IP and note the
`FullyQualifiedName` field:

```
> go run controller/script/destination-client/main.go -method getProfile -path 10.43.161.68:8080
INFO[0000] fully_qualified_name:"linkerd-identity.linkerd.svc.cluster.local" ..
```

```
> go run controller/script/destination-client/main.go -method getProfile -path linkerd-identity.linkerd.svc.cluster.local
INFO[0000] fully_qualified_name:"linkerd-identity.linkerd.svc.cluster.local" ..
```


controller/gen/apis/serviceprofile/v1alpha2/types.go
```diff
@@ -28,9 +28,10 @@ type ServiceProfile struct {
 
 // ServiceProfileSpec specifies a ServiceProfile resource.
 type ServiceProfileSpec struct {
-	Routes       []*RouteSpec   `json:"routes"`
-	RetryBudget  *RetryBudget   `json:"retryBudget,omitempty"`
-	DstOverrides []*WeightedDst `json:"dstOverrides,omitempty"`
+	Routes             []*RouteSpec   `json:"routes"`
+	RetryBudget        *RetryBudget   `json:"retryBudget,omitempty"`
+	DstOverrides       []*WeightedDst `json:"dstOverrides,omitempty"`
+	FullyQualifiedName string         `json:"fullyQualifiedName,omitempty"`
```
adleong, line 34: isn't the name of the service profile already the fully qualified name?  can we use that instead of creating a new field for it?

kleimkuhler: @adleong Yes if a Service Profile exists for the service, then the name would already be the FQN. But, if a Service Profile does not exist for the service then we want to be able to return a default profile and we wouldn't have a name for it.

Adding this field ensures that by default, a service with or without a Service Profile always has the FQN field populated.

adleong: ah, I see that we're populating this in the traffic split adapter which seems like a really unexpected place to handle this.  

seems like we should take the FQN from the name of the service profile if it exists.  if it doesn't exist, we'll call `defaultServiceProfile()`.  If the profile translator holds the service ID (just like the traffic split adaptor does) then we can populate the FQN on the default service profile from the service ID.

controller/api/destination/traffic_split_adaptor.go
```diff
@@ -52,6 +52,8 @@ func (tsa *trafficSplitAdaptor) publish() {
 	merged := sp.ServiceProfile{}
 	if tsa.profile != nil {
 		merged = *tsa.profile
+	} else {
+		merged.Name = fmt.Sprintf("%s.%s.svc.%s", tsa.id.Name, tsa.id.Namespace, tsa.clusterDomain)
```
kleimkuhler, line 56: If we're creating an empty service profile here that will be populated with only destination overrides, we set the name based off the `trafficSplitAdaptor`.

It's set here instead of `profileTranslator` because I think it makes more sense since we are already setting one field of the service profile (destination overrides). This way, we don't need to check for empty service profile names later.

controller/api/destination/profile_translator.go
```diff
@@ -90,9 +99,10 @@ func toServiceProfile(profile *sp.ServiceProfile) (*pb.DestinationProfile, error
 		budget.Ttl = toDuration(ttl)
 	}
 	return &pb.DestinationProfile{
-		Routes:       routes,
-		RetryBudget:  budget,
-		DstOverrides: toDstOverrides(profile.Spec.DstOverrides),
+		Routes:             routes,
+		RetryBudget:        budget,
+		DstOverrides:       toDstOverrides(profile.Spec.DstOverrides),
+		FullyQualifiedName: profile.Name,
```
adleong, line 105: Thinking about this some more, since we already have the FQN on the translator anyway, we might as well just use that.  It should be the same as the ServiceProfile name but this way the logic is more consistent between the default and non-default cases.  It also means we don't need to craft a name for the merged service profile in the traffic split adaptor.

Doing this will require `toServiceProfile` to be a method on `profileTranslator` but that should be fine since it's only called from `Update`.

controller/api/destination/profile_translator.go
```diff
@@ -41,10 +49,11 @@ func (pt *profileTranslator) Update(profile *sp.ServiceProfile) {
 	pt.stream.Send(destinationProfile)
 }
 
-func defaultServiceProfile() *pb.DestinationProfile {
+func defaultServiceProfile(fullyQualifiedName string) *pb.DestinationProfile {
```
adleong, line 52: TIOLI:  I think this would make sense to be a method on `profileTranslator` which would mean we wouldn't need to pass in the FQN anymore.

controller/api/destination/profile_translator.go
kleimkuhler: Yea sounds good I think keeping it consistent is better

controller/api/destination/server.go
```diff
@@ -215,6 +216,8 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 		path = dest.GetPath()
 	}
 
+	translator = newProfileTranslator(stream, log, &service, s.clusterDomain)
```
adleong, line 219: I think you should just be able to use `:=` here and not declare translator above.

The `translator` variable that you declare in the inner scope on line 199 is already out of scope here.

go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/huandu/xstrings v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.7
 	github.com/julienschmidt/httprouter v1.2.0
-	github.com/linkerd/linkerd2-proxy-api v0.1.13
+	github.com/linkerd/linkerd2-proxy-api v0.1.14-0.20200928162511-7fcd01ab736d
```
adleong, line 29: do we need the pseudoversion looking thing on the version here?  can we just use v0.1.14?

kleimkuhler: Yep 1.14 was released now so I'll change this

adleong (approved):
Awesome!

alpeb (approved):
After having left to @adleong all the review hard work, I just can say this makes sense to me :+1:

PR #5020, opened 2020-09-29
Add values to linkerd-config

### LINKED ISSUE #5008: Add Values to linkerd-config
Linkerd stores its configuration in a `ConfigMap/linkerd-config` resource containing 3 files: `global`, `install`, and `proxy`.  Each of these files contains a protobuf defined config structure.  In an effort to simplify the install and upgrade code, we are moving away from these config structures in favor of using the Helm values directly.

Add a `values` file to the `ConfigMap/linkerd-config` resource which contains a yaml formatted representation of the full values used to render the chart.  Any private data such as the identity issuer private key should be redacted before being stored.  

This serves as an installed configuration manifest and in the future, components such as `linkerd check` and `linkerd inject` will read this configuration instead of reading the `global`, `install`, or `proxy` files (which will eventually be removed).

adleong:
Fixes #5008 

We add a `values` file to the `ConfigMap/linkerd-config` resource.  This file holds the full Values which were used to render the chart except that private data such as the identity issuer key are redacted.  This file is currently unused but will eventually be used by CLI commands such as `check` and `inject` which need to load Linkerd's configuration (as described in #5009).

This is one step in a larger effort to eventually get rid of the other files in `ConfigMap/linkerd-config`.

### DIFF
--- a/charts/linkerd2/templates/config.yaml
+++ b/charts/linkerd2/templates/config.yaml
@@ -34,3 +34,13 @@ data:
   {{- else -}}
   {{- include "linkerd.configs.install" . | nindent 4}}
   {{- end }}
+  values: |
+  {{- $values := deepCopy .Values }}
+  {{- /*
+    WARNING! All sensitive or private data such as TLS keys must be removed
+    here to avoid it being publicly readable.
+  */ -}}
+  {{- $_ := unset $values.identity.issuer.tls "keyPEM"}}
+  {{- $_ := unset $values "partials"}}
+  {{- $_ := unset $values "configs"}}
+  {{- toYaml $values | trim | nindent 4 }}
--- a/charts/linkerd2/templates/config.yaml
+++ b/charts/linkerd2/templates/config.yaml
@@ -41,6 +41,9 @@ data:
     here to avoid it being publicly readable.
   */ -}}
   {{- $_ := unset $values.identity.issuer.tls "keyPEM"}}
+  {{- $_ := unset $values.profileValidator "keyPEM"}}
+  {{- $_ := unset $values.proxyInjector "keyPEM"}}
+  {{- $_ := unset $values.tap "keyPEM"}}
   {{- $_ := unset $values "partials"}}
   {{- $_ := unset $values "configs"}}
   {{- toYaml $values | trim | nindent 4 }}
--- a/charts/linkerd2/templates/config.yaml
+++ b/charts/linkerd2/templates/config.yaml
@@ -40,10 +40,18 @@ data:
     WARNING! All sensitive or private data such as TLS keys must be removed
     here to avoid it being publicly readable.
   */ -}}
-  {{- $_ := unset $values.identity.issuer.tls "keyPEM"}}
-  {{- $_ := unset $values.profileValidator "keyPEM"}}
-  {{- $_ := unset $values.proxyInjector "keyPEM"}}
-  {{- $_ := unset $values.tap "keyPEM"}}
+  {{- if kindIs "map" $values.identity.issuer.tls -}}
+    {{- $_ := unset $values.identity.issuer.tls "keyPEM"}}
+  {{- end -}}
+  {{- if kindIs "map" $values.profileValidator -}}
+    {{- $_ := unset $values.profileValidator "keyPEM"}}
+  {{- end -}}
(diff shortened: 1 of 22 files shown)


Pothulapati:
@adleong I was actually adding this configMap as part of https://github.com/linkerd/linkerd2/pull/4941, I can remove the configMap creation and limit the PR only for the usage in the CLI. WDYT?

Also, As the same configMap is rendered with Helm and Helm attaches the `global` field and `partials` field to each add-on i.e tracing, etc for its rendering. We might also unset them to not cause unnecessary leakages. (though the information does not look private)

Hence, My version of the sanitization had the following
```
{{- $dupValues := deepCopy .Values -}}
{{- $dupValues := omit $dupValues "partials" -}}
{{- range $key, $value := $dupValues -}}
  {{- if kindIs "map" $value -}}
    {{- if and (hasKey $value "global") (ne $key "configs" ) -}}
      {{- $dupValues := set $dupValues $key (unset $value "global") -}}
    {{- end -}}
    {{- if hasKey $value "partials" -}}
      {{- $dupValues := set $dupValues $key (unset $value "partials") -}}
    {{- end -}}
    {{- if hasKey $value "KeyPEM" -}}
      {{- $dupValues := set $dupValues $key (unset $value "keyPEM") -}}
    {{- end -}}
  {{- end -}}
{{- end -}}
{{- /* Unset known private fields */ -}}
{{- if $dupValues.identity.issuer.tls.keyPEM -}}
  {{- $_ := set $dupValues.identity.issuer.tls "keyPEM" "" -}}
{{- end -}}
{{- if $dupValues.profileValidator.keyPEM -}}
  {{- $_ := set $dupValues.profileValidator "keyPEM" "" -}}
{{- end -}}
{{- if $dupValues.proxyInjector.keyPEM -}}
  {{- $_ := set $dupValues.proxyInjector "keyPEM" "" -}}
{{- end -}}
```

adleong:
> I can remove the configMap creation and limit the PR only for the usage in the CLI. WDYT?

Yeah.  In this case, I think smaller and more incremental PRs are better.

I don't understand what you mean about helm attaching fields for each addon.  I do unset the "partials" field, but I'm not sure what addons have to do with it.  Looking at an example tracing section of the values in the golden file, it looks like

```
    tracing:
      enabled: true
```

which is what I would expect.

Pothulapati:
@adleong Example from one of your golden test file (global field being present with `grafana`)
https://github.com/linkerd/linkerd2/blob/d5f04804d36922a37844848cae46cc167a172d08/cli/cmd/testdata/install_helm_output.golden#L997

adleong:
Ah, I see what you mean now, @Pothulapati.  I'd be inclined to leave this as it is for the sake of simplicity unless it causes a problems.  This will all go away once the addons are split out.

alpeb (commented):
This sounds like a nice intermediary step for moving on :+1: 

I did find that `global.linkerdVersion` and `enforcedHostRegexp` are not being reflected in the new `values` entry (only when installing through the CLI), probably because a missing field in the underlying values struct, so it might make sense to fix it in the other PR :wink:

charts/linkerd2/templates/config.yaml
```diff
@@ -34,3 +34,24 @@ data:
...
+    {{- $_ := unset $values.identity.issuer.tls "keyPEM"}}
+  {{- end -}}
+  {{- if kindIs "map" $values.profileValidator -}}
+    {{- $_ := unset $values.profileValidator "keyPEM"}}
+  {{- end -}}
+  {{- if kindIs "map" $values.proxyInjector -}}
+    {{- $_ := unset $values.proxyInjector "keyPEM"}}
+  {{- end -}}
+  {{- if kindIs "map" $values.tap -}}
+    {{- $_ := unset $values.tap "keyPEM"}}
```
alpeb, line 53: Why the additional `kindIs` checks? It doesn't fail to me if they're absent.

adleong: I hit certain tests which failed without it.  I believe when `Values.identity.issuer.tls` is absent (or nil, maybe?).  Here is the failure:

https://github.com/linkerd/linkerd2/runs/1184792150?check_suite_focus=true

alpeb (approved):
LGTM :+1:

Pothulapati (approved):
LGTM

PR #5023, opened 2020-09-30
Update check to use new linkerd-config.values

Pothulapati:
This branch updates the check functionality to read
the new `linkerd-config.values` which contains the full
Values struct showing the current state of the Linkerd
installation. (being added in #5020 )

This is done by adding a new `FetchCurrentConfiguraiton`
which first tries to get the latest, if not falls back
to the older `linkerd-config` protobuf format.`


pkg/healthcheck/healthcheck.go
```diff
@@ -1623,6 +1624,29 @@ func (hc *HealthChecker) checkCertificatesConfig(ctx context.Context) (*tls.Cred
...
+func FetchCurrentConfiguration(ctx context.Context, k kubernetes.Interface, controlPlaneNamespace string) (*corev1.ConfigMap, *l5dcharts.Values, error) {
+
+	// Get the linkerd-config values if present
+	configMap, configPb, err := FetchLinkerdConfigMap(ctx, k, controlPlaneNamespace)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if rawValues := configMap.Data["values"]; rawValues != "" {
+		var fullValues *l5dcharts.Values
```
adleong, line 1637: should `fullValues` just be a `l5dcharts.Values` instead of a pointer?  I assume that the default value for a pointer is nil, which isn't what we want.

alpeb:
Eager to see `main` merged here to give this a full try :partying_face:

pkg/healthcheck/healthcheck.go
adleong: I'm testing this and it actually seems to work.  But I'm not sure why heh....

adleong (approved):
Other than a pointer which is really confusing me, this all looks great!

alpeb (approved):
LGTM :+1:

adleong (approved):

Issue #5025, opened 2020-09-30
Expose podLabels and podAnnotations as helm chart variables


tustvold:
Most charts expose variables that can be set with arbitrary YAML that gets passed into the labels and annotations block of the pod spec. For example, [here](https://github.com/helm/charts/blob/master/incubator/aws-alb-ingress-controller/templates/deployment.yaml#L21) or [here](https://github.com/helm/charts/blob/master/stable/cluster-autoscaler/templates/deployment.yaml#L25)

My team's log shipping pipeline relies on specific labels to ship logs to the correct location and so this functionality is very important for us, but it also seems to be functionality found in most other charts.

Will shortly submit a corresponding PR for your consideration.

tustvold:
Closed by #5027

PR #5027, opened 2020-09-30
Expose podLabels and podAnnotations in helm chart

tustvold:
Closes (#5025)

I initially considered merging with the existing "partials.proxy.*" - but not only would this have required splitting the matchLabels, but also would have put the value under `.global.proxy` which I thought confusing.


zaharidichev (commented):
Thanks for the PR @tustvold. I think this change can certainly benefit from adding a test for it. Take a look at `TestRenderHelm` for reference

tustvold:
I can't work out why the test is failing in CI, what is actually failing, nor reproduce it locally...

alpeb:
> I can't work out why the test is failing in CI, what is actually failing, nor reproduce it locally...

That was confusing indeed. Turns out that before tests are run, github merges the branch with the `main` branch, so it's not _really_ testing the same code you have here. If you do the merge yourself you'll see that there are other intervening changes that recently got into `main` that require changing the output of the golden file.

cli/cmd/install_helm_test.go
```diff
@@ -43,6 +43,21 @@ tracing:
...
+		addOnConfig := `
+global:
+  podLabels:
+    foo: bar
+    fiz: buz
+  podAnnotations:
+    bingo: bongo
+    asda: fasda
+`
+		chartControlPlane := chartControlPlane(t, ha, addOnConfig, "333", "444")
```
alpeb, line 58: Do you mind changing the naming of `addonConfig` here and the argument name in the `chartControlPlane` function declaration as well, since this doesn't apply specifically to to addon configs, but any kind of `values.yml` config change? Could be named something like `configDelta`...

alpeb (approved):
Thanks @tustvold , this looks good to me :+1: 
The merge conflict looks like an easy fix.

zaharidichev (approved):
Thanks a lot. LGTM!

alpeb:
@tustvold I merged `main` again here using github's UI to solve some new conflicts, but now your new file `install_helm_output_ha_labels.golden` will need to be updated as well and I can't do that. Do you mind addressing that last bit so I can merge and close this PR?

alpeb:
Thanks again @tustvold :+1: 

Have you already received some swag from us? We'd love to send you some, just gotta fill out this [form](https://docs.google.com/forms/d/e/1FAIpQLSdXA7kAl6oyUQ4DFRjiDc2biWlvkhNirrMwsXaQ8V3Uh8Yj5A/viewform) :slightly_smiling_face:

Issue #5026, opened 2020-09-30
Allow overriding webhook namespaceSelector in helm chart


tustvold:
The current webhook configuration put in place by the helm charts has a namespaceSelector that is by default globally enabled

```
namespaceSelector:
    matchExpressions:
    - key: config.linkerd.io/admission-webhooks
      operator: NotIn
      values:
      - disabled
```

It was perhaps a rather niche situation, but whilst testing linkerd I managed to get a cluster into a chicken-and-egg situation where the admission controller wasn't running because there were no ready nodes for it to run on, but the nodes weren't becoming ready because the webhooks weren't disabled for the kube-system namespace, which meant creation of the CNI pods was blocked on the webhook, which wasn't running because none of the nodes were ready.

Whilst adding the annotation to the namespace and giving the CNI daemonset a poke resolved this situation, I'd rather explicitly enable the webhooks for specific namespaces. I am aware I may be alone in this, so I'd like to propose adding a variable to the helm chart to allow overriding the namespaceSelector used by the webhooks. I will shortly be submitting a PR doing this for your consideration.

For additional context AWS AppMesh is disabled by default - see [here](https://github.com/aws/eks-charts/blob/master/stable/appmesh-inject/templates/webhook.yaml#L23).

tustvold:
Closed by #5074

PR #5028, opened 2020-09-30
Expose namespaceSelector for admission webhooks in helm chart

tustvold:
Closes (#5026)


charts/linkerd2/templates/proxy-injector-rbac.yaml
```diff
@@ -83,11 +83,15 @@ metadata:
 webhooks:
 - name: linkerd-proxy-injector.linkerd.io
   namespaceSelector:
+    {{- if .Values.proxyInjector.namespaceSelector }}
```
grampelberg, line 86: Why not just set the default to the current value? That way there isn't an if/else statement.

grampelberg (commented):
Cool!

I'm a little surprised the tests are passing, @Pothulapati did we get rid of the golden file stuff?

grampelberg:
@drewwells does this solve your issues?

charts/linkerd2/templates/proxy-injector-rbac.yaml
tustvold: The original version did, but it caused the tests to fail :sweat_smile:

Pothulapati:
@grampelberg because this feature is opt-in? i.e the golden test files are not changed unless the `namespaceSelector` is specified?

charts/linkerd2/templates/proxy-injector-rbac.yaml
```diff
@@ -83,11 +83,15 @@ metadata:
...
+    {{- toYaml .Values.proxyInjector.namespaceSelector | trim | nindent 4 }}
```
Pothulapati, line 87: I did not run the PR yet, but As I can see `matchExpressions` already below, and from [the api](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#labelselector-v1-meta) It seems like there is only one `matchExpression` allowed? :thinking:

tustvold: I'm not sure I follow? It is just a block where the user can override what is set by default? It's no different from how you might plumb in a resources block or a set of custom sidecar containers?

charts/linkerd2/templates/proxy-injector-rbac.yaml
vijaygos: Shouldn't this also have kube-system here by default?

charts/linkerd2/templates/sp-validator-rbac.yaml
```diff
@@ -71,11 +71,15 @@ metadata:
...
   namespaceSelector:
+    {{- if .Values.profileValidator.namespaceSelector }}
+    {{- toYaml .Values.profileValidator.namespaceSelector | trim | nindent 4 }}
+    {{- else }}
     matchExpressions:
     - key: config.linkerd.io/admission-webhooks
       operator: NotIn
       values:
       - disabled
+  {{- end }}
```
drewwells, line 82:
```suggestion
    {{- end }}
```

drewwells (approved):
This is a good start that doesnt change the default contract. Ultimately well want to change the defaults so linkerd can not break pod creation ie. kube-system

charts/linkerd2/templates/sp-validator-rbac.yaml
drewwells, line 82: It would be safer to exclude your own namespace too. Once the kube-apiserver webhooks are modified, it will expect tap service to respond to pod creation events.

```suggestion
    - key: name
      operator: NotIn
       values:
       - {{ .Release.Namespace }}
     {{- end }}
```

charts/linkerd2/templates/proxy-injector-rbac.yaml
zaharidichev: > Happy to change it back as it would be my preferred solution, but it may need changes beyond just to the helm chart

I think I would prefer the default values approach. Additional changes are fine.

zaharidichev (commented):
Thanks a lot for the pr!. As always adding some tests to verify the output is as expected is highly encouraged :)

charts/linkerd2/templates/sp-validator-rbac.yaml
tustvold: I'd agree with this change, I in fact want something even stronger, but should we get that in as a separate PR?

drewwells: That sounds good to me. We could use this PR as is without the namespace change

tustvold:
Rebased and added new golden test

grampelberg:
@zaharidichev mind checking this one out again? It'd be awesome if we could get it into this week's edge!

adleong (commented):
This looks good!  Thanks for the nice fix to `coerceToTree` to properly handle arrays.  

Looks like there's another merge conflict (sorry) to resolve.

zaharidichev:
@tustvold  this looks good to me as well. Do you mind resolving the conflict so we can get it into the next edge?

adleong:
In order to make sure this change gets into our next edge release, I've gone ahead and resolved the merge conflicts here: https://github.com/linkerd/linkerd2/pull/5074

adleong:
Thanks, @tustvold!  I resolved the conflicts and merged this in https://github.com/linkerd/linkerd2/pull/5074.

PR #5032, opened 2020-10-01
add annotations and labels to service in multicluster helm chart fixe…

### LINKED ISSUE #5024: multicluster: single network with private endpoints for gateways

Two clusters running on the same network do not need to communicate via the public internet with each other.
Having large payloads can be costly for in and egress traffic.

As I am running mostly in GKE there is a concept of an internal load balancer.
You generate a static private IP and assign it to the Service of type LoadBalancer.

One solution might be to extend the helm chart for the gateway to accept service annotations like:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: some-service-name
  annotations:
    cloud.google.com/load-balancer-type: "Internal"
spec:
  type: LoadBalancer
...
```

As I am still evaluating linkerd I cannot predict any side effects which might occur either with mTLS or service mirroring.

Not yet.

As there is only a change in routing there shouldn't be an interaction change.

Best regards,
Dariusch

Dariusch:
Add annotations and labels to service in multicluster helm chart

fixes #5024

helm template renders:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: linkerd-gateway
  namespace: linkerd-multicluster
  annotations:
    mirror.linkerd.io/gateway-identity: linkerd-gateway.linkerd-multicluster.serviceaccount.identity..
    mirror.linkerd.io/probe-period: "3"
    mirror.linkerd.io/probe-path: /health
    mirror.linkerd.io/multicluster-gateway: "true"
    linkerd.io/control-plane-component: gateway
    linkerd.io/created-by: linkerd/helm linkerdVersionValue
    app: test
  labels:
    app: test
spec:
  ports:
  - name: mc-gateway
    port: 4143
    protocol: TCP
  - name: mc-probe
    port: 4181
    protocol: TCP
  selector:
    app: linkerd-gateway
  type: LoadBalancer

```

### DIFF
--- a/charts/linkerd2-multicluster/README.md
+++ b/charts/linkerd2-multicluster/README.md
@@ -13,7 +13,7 @@ linkerd2-multicluster chart and their default values.
 | Parameter                       | Description                                                                                 | Default                                      |
 |---------------------------------|---------------------------------------------------------------------------------------------|----------------------------------------------|
 |`controllerComponentLabel`       | Control plane label. Do not edit                                                            |`linkerd.io/control-plane-component`          |
-|`controllerImage`                | Docker image for the Service mirror component (uses the Linkerd controller image)           |`ghcr.io/linkerd/controller`                |
+|`controllerImage`                | Docker image for the Service mirror component (uses the Linkerd controller image)           |`ghcr.io/linkerd/controller`                  |
 |`controllerImageVersion`         | Tag for the Service Mirror container Docker image                                           |`latest version`                              |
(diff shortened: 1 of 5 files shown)


Dariusch:
Hey guys, I would appreciate some help as I am fairly new to the go language.
There seems to be something wrong with my code but I cannot figure out where the problem lies.

grampelberg:
@Pothulapati can you take a look?

Dariusch:
I am closing this draft until I have a working solution on my fork

PR #5036, opened 2020-10-02
Update Injection to use new linkerd-config.values

Pothulapati:
This PR Updates the Injection Logic (both CLI and proxy-injector)
to use `Values` struct instead of protobuf Config, part of our move
in removing the protobuf.

This does not touch any of the flags, install related code.


cli/cmd/inject.go
```diff
@@ -80,17 +81,17 @@ sub-folders, or coming from stdin.`,
 				return err
 			}
 
-			configs, err := options.fetchConfigsOrDefault()
+			values, err := options.fetchConfigsOrDefault()
```
adleong, line 84: I think if we pull in https://github.com/linkerd/linkerd2/pull/5038 then we can add `makeProxyFlags` and `makeInjectFlags` from https://github.com/linkerd/linkerd2/pull/5005 to this branch.  This lets us remove `proxyConfigOptions` altogether.

cli/cmd/inject.go
```diff
@@ -80,17 +81,17 @@ sub-folders, or coming from stdin.`,
...
 			if err != nil {
 				return err
 			}
 			overrideAnnotations := map[string]string{}
-			options.overrideConfigs(configs, overrideAnnotations)
+			options.overrideConfigs(values, overrideAnnotations)
```
adleong, line 89: see https://github.com/linkerd/linkerd2/pull/5005/files#diff-0c198b8ec41cebf0e60ade048b268139R340

I think it makes sense here to simply return the annotation overrides rather than passing them in.

cli/cmd/inject.go
```diff
@@ -336,119 +337,119 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output
...
-			return nil, err
-		}
-
-		return install.configs(nil), nil
+		return linkerd2.NewValues(false)
 	}
 
-	api := checkPublicAPIClientOrExit()
-	config, err := api.Config(context.Background(), &public.Empty{})
+	checkPublicAPIClientOrExit()
```
adleong, line 349: Do we actually need the public API client for anything?

adleong (commented):
I recommend taking a close look specifically at `pkg/inject/inject.go` and `cli/cmd/inject.go` from https://github.com/linkerd/linkerd2/pull/5005.  I suspect that you should be able to take those files from that PR pretty much as they are.

cli/cmd/inject.go
Pothulapati: It is used to perform a subset of `healthchecks` to make sure the control plane is healthy and is usually in most command paths.

pkg/inject/inject.go
```diff
@@ -572,27 +568,11 @@ func (conf *ResourceConfig) injectPodSpec(values *patch) {
...
-		values.Global.Proxy.DisableIdentity = true
-		return
-	}
-	values.Global.IdentityTrustAnchorsPEM = idctx.GetTrustAnchorsPem()
-	values.Global.IdentityTrustDomain = idctx.GetTrustDomain()
-	values.Identity = &l5dcharts.Identity{}
-
 }
 
 func (conf *ResourceConfig) injectProxyInit(values *patch) {
```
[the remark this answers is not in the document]
Pothulapati: Yep, That makes sense. Updated accordingly.

cli/cmd/install.go
```diff
@@ -398,34 +398,34 @@ control plane.`,
...
+	values, err := options.validateAndBuild(ctx, stage, flags)
 	if err != nil {
 		return err
 	}
 
 	return render(os.Stdout, values)
 }
 
-func (options *installOptions) validateAndBuild(ctx context.Context, stage string, flags *pflag.FlagSet) (*l5dcharts.Values, *pb.All, error) {
+func (options *installOptions) validateAndBuild(ctx context.Context, stage string, flags *pflag.FlagSet) (*l5dcharts.Values, error) {
```
Pothulapati, line 409: This had to be done, as the linter complains  that `pb.All` return is not being used anywhere

pkg/inject/inject.go
```diff
@@ -572,55 +501,35 @@ func (conf *ResourceConfig) injectPodSpec(values *patch) {
...
-	values.AddRootVolumes = len(conf.pod.spec.Volumes) == 0
+	if v := conf.pod.meta.Annotations[k8s.ProxyEnableDebugAnnotation]; v != "" {
+		debug, err := strconv.ParseBool(v)
+		if err != nil {
+			log.Warnf("unrecognized value used for the %s annotation: %s", k8s.ProxyEnableDebugAnnotation, v)
+			debug = false
+		}
 
-	values.Global.Proxy.Trace = &l5dcharts.Trace{}
-	if trace := conf.trace(); trace != nil {
```
Pothulapati, line 582: self comment: fix tracing

cli/cmd/testdata/inject_emojivoto_pod.golden.yml
```diff
@@ -31,6 +31,10 @@ spec:
       value: 10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
     - name: LINKERD2_PROXY_DESTINATION_PROFILE_NETWORKS
       value: 10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
+    - name: LINKERD2_PROXY_INBOUND_CONNECT_TIMEOUT
```
Pothulapati, line 34: Every golden file got these connect timeouts added, as these where part of the `Values`, and were made for control-plane only and not for user applications?

If thats the case, we can have do have another check in tpl's to make this emitted only for control-plane components :thinking:

cli/cmd/inject_test.go
```diff
@@ -61,89 +59,77 @@ func testUninjectAndInject(t *testing.T, tc testCase) {
...
 		{
 			inputFileName:    "inject_emojivoto_deployment.input.yml",
 			goldenFileName:   "inject_emojivoto_deployment.golden.yml",
 			reportFileName:   "inject_emojivoto_deployment.report",
 			injectProxy:      true,
-			testInjectConfig: defaultConfig,
-		},
-		{
-			inputFileName:    "inject_emojivoto_deployment.input.yml",
-			goldenFileName:   "inject_emojivoto_deployment_empty_version_config.golden.yml",
```
Pothulapati, line 129: These tests with empty versions were deleted as `Values` is our final config,(on which updates through flags, annotations were overridden before rendering) and hence having a test with their versions field as empty does not make sense

charts/patch/templates/patch.json
```diff
@@ -35,6 +35,7 @@
     "value": "{{$value}}"
   },
   {{- end }}
+  {{- if .Values.injection }}
```
Pothulapati, line 38: Currently, As full `Values` are rendered both in normal injection and `manual` injection, There is no way for the templates to differentiate that, as they depend on checking whether fields like `.Values.proxy`, etc are nil or not. Hence, that logic is moved into the template by adding a new `.Values.injection` which is set when it is `--manual` or through the Injector.

charts/patch/templates/patch.json
```diff
@@ -44,7 +45,7 @@
   },
   {{- end }}
   {{- end}}
-  {{- if .Values.global.proxyInit }}
+  {{- if and .Values.global.proxyInit (not .Values.global.cniEnabled) }}
```
Pothulapati, line 48: Same here, as `.Values.global.proxyInit` has values, even when the installation is through CNI, and hence we need to depend on `.global.cniEnabled` to skip InitContainer INjection

cli/cmd/install_test.go
```diff
@@ -335,6 +337,38 @@ func testInstallOptions() (*installOptions, error) {
 	return o, nil
 }
 
+func testInstallValues() (*linkerd2.Values, error) {
```
Pothulapati, line 340: This is a helper function used with tests, to fetch full `Values` struct with Cert Key pairs from `/testdata`

controller/proxy-injector/fake/data/pod-with-debug.patch.json
```diff
@@ -148,7 +156,7 @@
         },
         {
           "name": "LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES",
-          "value": "."
+          "value": "svc.cluster.local."
```
Pothulapati, line 159: This was not previously specified but now that we are using `Values` generated from `Values.yaml` which has `.svc.cluster,local` as the value, it got added.

Pothulapati (commented):
As of now, This PR does not change the flags logic from #5038, but updates the Injection logic everywhere to use `Values`.

Here are some explanations on why things changed

adleong:
I've started to look into this integration test failure.  It seems its because a `linkerd tap` command fails to get any tap events from the `gateway` deployment in the `linkerd-tap-test` namespace.  Upon further inspection, the gateway's sidecar proxy has the `LINKERD2_PROXY_TAP_DISABLED` env var enabled, which explains why it is not emitting tap data.

Looking at the test application yaml, the annotation for disabling tap is not set for the gateway deployment but it IS set for the t4 deployment.  I suspect that when we run the injection code over this list of manifests, we have some bug where annotation overrides leak out and persist from one manifest to the next.  This could explain why the disable tap annotation from the t4 deployment seems to be affecting the gateway deployment.

Pothulapati:
@adleong Yep, That's right. It's because we were using the same `Values` passed to `ResourceConfig` for rendering, which is mutated based on annotations, etc on each file. I was able to fix this, by making a copy of `Values` and using that with `Patch`

I also retrieved the previous `trace()` logic to make Tracing work, as that logic added namespace, etc if not present for `collectorSvcAccount`

zaharidichev (commented):
Solid piece of work. Will do some more manual tests before approval

adleong:
I've pushed a commit to this branch which removes the `injection` field from the patch template values.  I think this was only needed because of the previous bug that you fixed regarding re-using values.  Now that we're doing a deep copy, it doesn't seem to be necessary.

pkg/inject/inject.go
```diff
@@ -209,49 +210,45 @@ func (conf *ResourceConfig) ParseMetaAndYAML(bytes []byte) (*Report, error) {
...
 
-	if conf.destinationGetNetworks() != "" {
-		for _, network := range strings.Split(strings.Trim(conf.destinationGetNetworks(), ","), ",") {
-			if _, _, err := net.ParseCIDR(network); err != nil {
-				return nil, fmt.Errorf("cannot parse destination get networks: %s", err)
-			}
-		}
+	err = yaml.Unmarshal(copyRawValues, &copyValues)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
adleong: this won't do a deep copy.  any pointers inside the values will be copied

cli/cmd/testdata/inject_emojivoto_pod.golden.yml
adleong: since these match the defaults in the proxy, this change should be a no-op, right?  But it's more explicit and directly reflects what's in the values.yml.  Seems like a good change to me if that's true.

adleong (approved):

pkg/inject/inject.go
```diff
@@ -209,49 +209,45 @@ func (conf *ResourceConfig) ParseMetaAndYAML(bytes []byte) (*Report, error) {
...
+	// Make a copy of Values and mutate that
+	var copyValues linkerd2.Values
+	copyRawValues, err := yaml.Marshal(conf.values)
+	if err != nil {
+		return nil, err
 	}
 
-	if conf.destinationGetNetworks() != "" {
-		for _, network := range strings.Split(strings.Trim(conf.destinationGetNetworks(), ","), ",") {
-			if _, _, err := net.ParseCIDR(network); err != nil {
```
alpeb, line 218: Same here, the pod is failing at runtime instead of stopping injection if the annotation value is invalid.

alpeb (approved):
Looks good overall to me :+1: 
I just caught a couple of validation issues, but I'm approving in case this is to be dealt with later.

adleong:
Thanks, @alpeb!  I've just pushed a commit adding these validations back in.

zaharidichev (approved):
LGTM!

PR #5042, opened 2020-10-05
Fix Heartbeat mem and cpu stats

alpeb:
Since k8s 1.16 cadvisor uses the `container` label instead of
`container_name` in the prometheus metrics it exposes.
The heartbeat queries were using the latter, so they were broken
for k8s version since 1.16.

Note that the `p99-handle-us` value is still missing because the
`request_handle_us` metrics is always zero.

### DIFF
--- a/controller/heartbeat/heartbeat.go
+++ b/controller/heartbeat/heartbeat.go
@@ -18,6 +18,11 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+type containerMeta struct {
+	name model.LabelValue
+	ns   model.LabelValue
+}
+
 // K8sValues gathers relevant heartbeat information from Kubernetes
 func K8sValues(ctx context.Context, kubeAPI *k8s.KubernetesAPI, controlPlaneNamespace string) url.Values {
 	v := url.Values{}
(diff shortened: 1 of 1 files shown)


alpeb:
Raised #5043 to track the `request_handle_us` issue.

controller/heartbeat/heartbeat.go
```diff
@@ -122,7 +121,9 @@ func PromValues(promAPI promv1.API, controlPlaneNamespace string) url.Values {
 		}
 
 		// p95-cpu
-		query = fmt.Sprintf("max(quantile_over_time(0.95,rate(container_cpu_usage_seconds_total%s[5m])[24h:5m]))", containerLabels)
+		query = fmt.Sprintf("max(quantile_over_time(0.95,rate(container_cpu_usage_seconds_total%s[5m])[24h:5m]) "+
+			"or quantile_over_time(0.95,rate(container_cpu_usage_seconds_total%s[5m])[24h:5m]))",
```
zaharidichev, line 125: Was hoping there is a less verbose way, but it seems that is not the case.

zaharidichev (approved):
LGTM!

kleimkuhler (approved):
Looks good 👍

PR #5049, opened 2020-10-08
More comprehensive injection integration test

zaharidichev:
The purpose of this test is to validate that the auto injector configures the proxy and the additional containers according to the specified config.

This is done by providing a helper that can generate the desired annotations and later inspect an injected pod in order to determine that every bit of configuration has been accounted for.  This test is to provide further assurance that #5036 did not introduce any regressions


test/integration/inject/inject_test.go
```diff
@@ -53,32 +55,38 @@ func TestInjectManual(t *testing.T) {
...
-		"--proxy-uid=1337",
-		"--proxy-log-level=warn",
-		"--enable-external-profiles",
-		"testdata/inject_test.yaml",
-	}
+	}, flags...)
+
+	cmd = append(cmd, "testdata/inject_test.yaml")
 
 	out, stderr, err := TestHelper.LinkerdRun(cmd...)
```
alpeb, line 91: I believe there's a way to transform this `out` into a k8s object, from which you could extract `v1.Pod` and use `injectionValidator.ValidatePod()` on it like you do below, so we can get rid of the golden file comparison?

alpeb (commented):
This is sweet :+1: 

What do you think about using this new approach in the other tests in this file that didn't get touched? We could thus get rid of the golden files as well.

As a side note, likely outside of this PR , it would be nice to have the correspondence between annotations and flags encoded in a single place.  Currently we have annotation consts in `labels.go` and there's a list of annotations (dunno how up to date) in `cli/cmd/doc.go`.

test/integration/inject/inject_test.go
zaharidichev: Yes I will attempt that. Very good idea. Not sure why my CI is failing. Seems completely unrelated...

zaharidichev:
@alpeb I managed to remove the golden files 🎉  CI is passing for me locally. Not sure what is happening...

test/integration/inject/inject_test.go
```diff
@@ -30,67 +35,197 @@ func TestMain(m *testing.M) {
...
+	var deploy appsv1.Deployment
+	_, _, err := s.Decode([]byte(yamlString), nil, &deploy)
+	if err != nil {
+		return nil, err
+	}
+
+	return &deploy, nil
+}
+
 func TestInjectManual(t *testing.T) {
```
zaharidichev, line 50: @alpeb I wonder whether it makes sense to have this here. We are pretty much testing the same logic as in `TestInjectManualParams` but with a limited set of config params

testutil/kubernetes_helper.go
```diff
@@ -108,6 +120,20 @@ func (h *KubernetesHelper) CreateDataPlaneNamespaceIfNotExists(ctx context.Conte
 	return h.createNamespaceIfNotExists(ctx, namespace, annotations, map[string]string{"linkerd.io/is-test-data-plane": "true"})
 }
 
+// WithDataPlaneNamespace is used to create a test namespace that is deleted before the function returns
+func (h *KubernetesHelper) WithDataPlaneNamespace(ctx context.Context, testName string, annotations map[string]string, t *testing.T, helper *TestHelper, test func(t *testing.T, ns string)) {
```
alpeb, line 124: Doesn't it make more sense to declare this as a method on `TestHelper`, so you don't have to pass `*TestHelper` as an arg, and you have access to `KubernetesHelper` as a field?

test/integration/inject/inject_test.go
```diff
@@ -5,11 +5,16 @@ import (
...
 	"os"
 	"reflect"
+	"strings"
 	"testing"
+	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
```
Pothulapati, line 16: nit: Removing this line should make goimport combine all these imports.

testutil/kubernetes_helper.go
Pothulapati, line 124: Cool Hack! :+1:

testutil/kubernetes_helper.go
```diff
@@ -7,11 +7,14 @@ import (
...
 	"regexp"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
Pothulapati, line 17: unnecessary I guess?

testutil/inject_validator.go
```diff
@@ -0,0 +1,568 @@
...
+
+const enabled = "true"
+const proxyContainerName = "linkerd-proxy"
+const initContainerName = "linkerd-init"
+const debugContainerName = "linkerd-debug"
+
+// InjectValidator is used as a helper to generate
+// correct injector flags and annotations and verify
+// injected pods
+type InjectValidator struct {
```
Pothulapati, line 22: why not use `proxyOptions` here, and re-use the `overrideConfigs` here instead of having a `getFlagsAndAnnotations`? If `proxyOptions` does not contain all fields, maybe that could be a substruct? :thinking:

testutil/kubernetes_helper.go
zaharidichev: Yes it does indeed

test/integration/inject/inject_test.go
alpeb: Agreed, I don't see it adding value.

testutil/inject_validator.go
alpeb: interesting :thinking:

zaharidichev: Hmm yeah it could be done but I think it clearly serves a different purpose and apart from that `overrideConfigs` takes `values *linkerd2.Values`, so I kind of need to come up with that out of somewhere. Not sure whether it makes much sense to try and fit that here...

Pothulapati: I was expecting you to pass a `nil` there (as its not used much). But this could be unecessary work and hence resolving this suggestion.

alpeb (approved):

Pothulapati (approved):
LGTM! :shipit:

PR #5050, opened 2020-10-08
edge-20.10.2

zaharidichev:
## edge-20.10.2

This edge release adds more improvements for mTLS for all TCP traffic.
It also includes significant internal improvements to the way Linkerd
configuration is stored within the cluster.

*  Changed TCP metrics exported by the proxy to ensure that peer
   identities are encoded via the `client_id` and `server_id` labels.
*  Removed the dependency of control plane component on `linkerd-config`
*  Updated the data structure `proxy-injector` uses to derive the configuration
   used when injecting workloads

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Changes
 
+## edge-20.10.2
+
+This edge release adds more improvements for mTLS for all TCP traffic.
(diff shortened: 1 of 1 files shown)


admc (approved):
👍

olix0r (approved):

PR #5057, opened 2020-10-09
tests: Consolidate TestHelper.LinkerdRun error handling

olix0r:
Most invocations of `TestHelper.LinkerdRun` don't actually need the stderr
output except to encode it in the error message. This change introduces
`TestHelper.LinkerdRunOk` to return an error that includes the full
invoked command and error message.


test/integration/trafficsplit/trafficsplit_test.go
```diff
@@ -69,11 +69,11 @@ func parseStatTsRow(out string, expectedRowCount, expectedColumnCount int) (map[
 
 func statTrafficSplit(from string, ns string) (map[string]*statTsRow, error) {
 	cmd := []string{"stat", "ts", "--from", from, "--namespace", ns, "-t", "30s", "--unmeshed"}
-	stdOut, _, err := TestHelper.LinkerdRun(cmd...)
```
olix0r, line 72: Note that this was the case that triggered this PR -- we dropped the stderr silently. But now it's impossible to do so ;)

test/cli/cli_install_static_test.go
```diff
@@ -39,19 +39,8 @@ func TestMain(m *testing.M) {
...
-		}
-	)
-
-	exec := append([]string{cmd}, args...)
-	out, stderr, err := TestHelper.LinkerdRun(exec...)
+	_, err := TestHelper.LinkerdRunOk("install", "--ignore-cluster")
 	if err != nil {
-		testutil.AnnotatedFatalf(t, "'linkerd install' command failed",
-			"'linkerd install' command failed: \n%s\n%s", out, stderr)
+		testutil.AnnotatedFatalf(t, "'linkerd install' command failed", "%s", err)
```
alpeb, line 44: The second argument is what is shown in CI as an annotation `"'linkerd install' command failed"`, which should be a generic message with no error details, to be aggregated in the reports.
The third and fourth arguments make the error seen in the CLI. When running outside CI only that will be visible, so it might be a little confusing only seeing the error without the context ("'linkerd install' command failed")

alpeb (commented):
Sticking sdterr and the actual error makes sense to me :+1: 
I'm not sure though about the annotations refactoring.

testutil/test_helper.go
```diff
@@ -336,6 +336,16 @@ func (h *TestHelper) LinkerdRun(arg ...string) (string, string, error) {
 	return h.PipeToLinkerdRun("", arg...)
 }
 
+// LinkerdRunOk executes a linkerd command appended with the --linkerd-namespace
+// flag.
+func (h *TestHelper) LinkerdRunOk(arg ...string) (string, error) {
+	out, stderr, err := h.PipeToLinkerdRun("", arg...)
+	if err != nil {
```
adleong, line 343: I like this extra context on the error.  Can we just do this in `LinkerdRun`?  That way I don't think we need to add `LinkerdRunOk`.  Callers who don't specifically care about the stderr can safely drop the second return value since the stderr will also be present in the error if there is an error.

test/cli/cli_install_static_test.go
adleong: would it make sense to prepend the `f.Fatal` message with the annotation message?  That way we can avoid repeating ourselves at the callsite like this.

alpeb: IIRC there are cases where it made sense to make the messages slightly different, but for simplicity's sake I think we can accommodate to your suggestion :+1:

testutil/test_helper.go
olix0r: There are one or two cases where we actually want to look at error messages

test/cli/cli_install_static_test.go
olix0r: AFAICT, there's not a lot of value in including that text again. The new error message includes all of the command line args:
```
fmt.Errorf("command failed: %s\n%s\n%s", strings.Join(arg, " "), err, stderr)
```

testutil/test_helper.go
olix0r: I've renamed `LinkerdRunOk` to `LinkerdRun` and changed all other uses to PipeToLinkerdRun

test/cli/cli_install_static_test.go
alpeb: Ok, command and args should be enough for the CLI message :ok_hand:

alpeb (approved):
LGTM :+1:

adleong (approved):

PR #5059, opened 2020-10-09
Handle grafana add-on config repair

### LINKED ISSUE #5014: Upgrade command seems to have incorrect Grafana image
Upgrade command has incorrect repo for grafana image. 

```
(main) | bin/linkerd install --ignore-cluster | grep  "linkerd/grafana"
        image: ghcr.io/linkerd/grafana:git-e8f0724a
```

```
 bin/linkerd upgrade  | grep  "linkerd-io/grafana"
        image: gcr.io/linkerd-io/grafana:git-e8f0724a
```

Pothulapati:
Fixes #5014

In Grafana Add-On, Default fields i.e `grafana.image.name`, `grafana.name`
have been removed from `linkerd-config-addons` after `2.8.1`. Only
overriden values are stored in `linkerd-config-addons` as of now.
Hence, `grafana.image.name` has to be removed from
`linkerd-config-addons` unless they are overridden so that updates
to it can take place **i.e especially the move from `gcr` to `ghcr`**.

This also removes `grafana.name` field if they are set to default, as
it's been removed.

**This problem will not occur in future even if we update default values, as
default values are not stored in `linekrd-config-addons` anymore for all
add-ons.**


cli/cmd/upgrade.go
```diff
@@ -383,6 +386,42 @@ func (options *upgradeOptions) validateAndBuild(ctx context.Context, stage strin
...
+			}
+		}
+
+		// Handle removal of grafana.name field
+		name, err := healthcheck.GetString(grafana, "name")
+		if err == nil {
+			// If default, remove it as its no longer needed
+			if name == "linkerd-grafana" {
+				delete(grafana, "name")
+			}
```
Pothulapati, line 417: In the else case, That is the user has overwritten `grafana.name` to be something else. Should we warn the user here that the upgrade will change the grafana instance name back to be `linkerd-grafana`. Should that be here or in Check?

zaharidichev (approved):
LGTM!

cli/cmd/upgrade.go
```diff
@@ -383,6 +386,42 @@ func (options *upgradeOptions) validateAndBuild(ctx context.Context, stage strin
 	return values, nil
 }
 
+func repairAddOnConfig(rawValues []byte) []byte {
+
+	var values map[string]interface{}
+	yaml.Unmarshal(rawValues, &values)
+
+	// Grafana Depreciation Fix
+	// TODO: Remove this once 2.9 is released
```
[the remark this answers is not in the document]
Pothulapati: Hmmm, Removing the todo, as we might require it for all upgrades from 2.8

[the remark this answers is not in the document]
zaharidichev: I think this is the case. Unless we decide to not support upgrades that jump versions...

adleong:
This will need to be updated after merging main, but I think this is the right idea.

adleong:
Since this only affects the upgrade from pre-2.9, we should be able to isolate this logic in `upgrade_legacy.go`, right?

Pothulapati:
@adleong Good Suggestion, As this issue, is not in the latest edges that have the new upgrade workflow (because we removed these fields from `linkerd-config-addons`) and only happens with legacy.

go.sum
```diff
@@ -60,8 +60,10 @@ github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 h1:d+Bc7a5rLufV
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/agnivade/levenshtein v1.0.1/go.mod h1:CURSv5d9Uaml+FovSIICkLbAUZ9S4RqaHDIsdSBg7lM=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
+github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 h1:JYp7IbQjafoB+tBA3gMyHYHrpOtNuDiK/uB5uXxq5wM=
```
adleong, line 63: I think we can revert the changes to this file.

cli/cmd/upgrade_legacy.go
```diff
@@ -75,6 +75,13 @@ func loadStoredValuesLegacy(ctx context.Context, k *k8s.KubernetesAPI) (*charts.
 				return nil, fmt.Errorf("values subpath not found in %s configmap", k8s.AddOnsConfigMapName)
 			}
 
+			// repair Add-On configs
+			repairedCm, err := repairAddOnConfig([]byte(cmData))
+			if err == nil {
```
adleong, line 80: I think this only returns an error if there is an error during marshalling or unmarshalling.  In that case, should we log the error and/or return it?

Pothulapati: Logging seems right, as returning fails the installation flow, which seems not necessary for a Add-On fix.

adleong (approved):

PR #5060, opened 2020-10-09
bin/tests: Improve argument parsing

olix0r:
The `bin/tests` script takes command-line arguments, but it requires
that all arguments are specified before the linkerd binary path; and it
silently ignores flags that follow the linkerd binary. Furthermore,
unexpected flags may be incorrectly parsed as the linkerd binary path.

This changes argument parsing to be more flexible about ordering; and it
prints the full usage error when unexpected flags are encountered.

### DIFF
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -56,12 +56,14 @@ handle_input() {
       --images)
         images=$2
         if [ -z "$images" ]; then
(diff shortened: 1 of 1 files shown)


alpeb (approved):
Nice, this had been bothering me for a while as well :+1:

zaharidichev (approved):
I was bitten by that as well. LGTM!

Issue #5007, opened 2020-09-25
Migrate integration test infra from KinD to k3d


alpeb:
K3s is smaller and has a faster startup time than KinD. Also it's easier to expose services outside the cluster in a cross-platform way, which allowed us to setup the multicluster integration test. 

k3d/k3s support was added in #4994 for integration tests to leverage, as an alternative to KinD. We'd like to further simplify things and base all the tests in k3d. For that we need to check if the tests requiring special config (cluster-domain and cni-calico) are easily ported to k3d.

PR #5064, opened 2020-10-12
avoid waiting when creating calico cluster with kind

zaharidichev:
Currently the `--wait` flag times out when creating a calico cluster. The result is that we end up waiting for 5 minutes to simply emit a warning and continue. Instead we can check the readiness of some k8s components to ensure our cluster is up and running and avoid the delay. 


kleimkuhler (approved):
That is odd! Good spot though. May be worth looking at if this happens with k3d as well as part of #5007

alpeb (approved):
LGTM :+1:

PR #5072, opened 2020-10-13
Allows tracing ServiceAccounts to use the PSP

glitchcrab:
#### Problem

Currently the tracing deployments do not start on clusters where restricted PodSecurityPolicies are enforced.

#### Solution

This PR adds the subchart's ServiceAccounts to the `linkerd-psp` RoleBinding, thereby allowing the deployments to be satisfied.

#### Validation

Deploy the current chart to a cluster where PodSecurityPolicies must be provided. It will fail with the following:

```
message: 'pods "linkerd-jaeger-77dd585d5c-" is forbidden: unable to validate against
      any pod security policy: [spec.volumes[0]: Invalid value: "emptyDir": emptyDir
      volumes are not allowed to be used spec.initContainers[0].securityContext.runAsUser:
      Invalid value: 0: must be in the ranges: [{1000 65535}] spec.initContainers[0].securityContext.capabilities.add:
      Invalid value: "NET_ADMIN": capability may not be added spec.initContainers[0].securityContext.capabilities.add:
      Invalid value: "NET_RAW": capability may not be added]'
```

This PR mitigates this problem.

---

Signed-off-by: Simon Weald <glitchcrab-github@simonweald.com>

### DIFF
--- a/charts/linkerd2/templates/psp.yaml
+++ b/charts/linkerd2/templates/psp.yaml
@@ -84,6 +84,11 @@ roleRef:
   name: linkerd-psp
   apiGroup: rbac.authorization.k8s.io
 subjects:
+{{ if .Values.tracing.enabled -}}
+- kind: ServiceAccount
+  name: linkerd-collector
+  namespace: {{.Values.global.namespace}}
+{{ end -}}
 - kind: ServiceAccount
   name: linkerd-controller
   namespace: {{.Values.global.namespace}}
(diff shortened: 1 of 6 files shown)


glitchcrab:
I'll fix the failing test shortly

glitchcrab:
Could I get some eyes on this please? CI has failed twice now for various different reasons which don't appear to be connected to my changes - it previously failed on a different step (multicluster-related).

Pothulapati (approved):
LGTM, and it feels good to have this, as we do the same with prometheus and grafana.

One suggestion is that we don't actually support using this tracing add-on in production and would always advise users to have their own tracing infra which is independent of Linkerd

glitchcrab:
Sure, that makes sense to me. It just came up because I'm working on our own Linkerd chart based on this one and it made sense to align all the subcharts, even if they won't necessarily be used.

adleong (approved):

PR #5075, opened 2020-10-13
Rename --addon-config flag to --config

adleong:
The `--addon-config` flag allows users to supply a yaml config file which will override the Values used to install or upgrade Linkerd.  While this is useful for supplying config for addons, it can be used to configure any part of the Values struct.  Thus, we rename the flag to `--config`.

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -206,7 +206,7 @@ func makeUpgradeFlags() *pflag.FlagSet {
 	)
 	upgradeFlags.BoolVar(
 		&addOnOverwrite, "addon-overwrite", false,
-		"Overwrite (instead of merge) existing add-ons config with file in --addon-config (or reset to defaults if no new config is passed)",
+		"Overwrite add-on configuration instead of loading the existing config (or reset to defaults if no new config is specified)",
 	)
 	return upgradeFlags
 }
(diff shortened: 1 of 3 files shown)


cli/cmd/upgrade.go
```diff
@@ -206,7 +206,7 @@ func makeUpgradeFlags() *pflag.FlagSet {
 	)
 	upgradeFlags.BoolVar(
 		&addOnOverwrite, "addon-overwrite", false,
-		"Overwrite (instead of merge) existing add-ons config with file in --addon-config (or reset to defaults if no new config is passed)",
+		"Overwrite add-on configuration instead of loading the existing config (or reset to defaults if no new config is specified)",
```
olix0r, line 209: Is it still appropriate for this to be called `addon`? Should this be `--overwrite-config`?

cli/cmd/upgrade_test.go
```diff
@@ -597,7 +597,7 @@ func TestUpgradeOverwriteRemoveAddonKeys(t *testing.T) {
 
 	install := renderInstall(t, installOpts)
 
-	flagSet.Set("addon-config", filepath.Join("testdata", "grafana_enabled.yaml"))
+	flagSet.Set("config", filepath.Join("testdata", "grafana_enabled.yaml"))
 	flagSet.Set("addon-override", "true")
```
olix0r, line 601: The flag is called `addon-overwrite` but this references `addon-override`. Are these things different?

kleimkuhler (approved):

olix0r (approved):

cli/cmd/upgrade_test.go
adleong: Good catch.  Will fix in a separate PR.

PR #5077, opened 2020-10-13
Fix addon-overwrite test

adleong:
The TestUpgradeOverwriteRemoveAddonKeys was not actually verifying that the fields which should be removed were actually removed.  Thus it failed to catch an error with the test itself where the `addon-overwrite` flag was spelled incorrectly and not properly registered.  

We update the test to verify that the field is removed and fix the test by correcting the spelling of the flag and properly registering it.


olix0r (approved):

alpeb (approved):
LGTM :+1:

PR #5083, opened 2020-10-14
test: Relax regex for Endpoint Slices update error

olix0r:
We've encountered errors like the following in CI:

```
Error updating Endpoint Slices for Service linkerd/linkerd-proxy-injector: Error updating linkerd-proxy-injector-27vgh EndpointSlice for Service linkerd/linkerd-proxy-injector: endpointslices.discovery.k8s.io "linkerd-proxy-injector-27vgh" not found
```

There is a regex to prevent similar errors from failing a test, but it
is too restrictive. This change relaxes the regex to ignore all errors
of this kind ("updating Endpoint Slices for Service").


alpeb (approved):
LGTM :+1:

PR #5093, opened 2020-10-15
Add a default set of ports to bypass the proxy

olix0r:
The proxy has a default, hardcoded set of ports on which it doesn't do
protocol detection (25, 587, 3306 -- all of which are server-first
protocols). In a recent change, this default set was removed from
the outbound proxy, since there was no way to configure it to anything
other than the default set. I had thought that there was a default set
applied to proxy-init, but this appears to not be the case.

This change adds these ports to the default Helm values to restore the
prior behavior.

I have also elected to include 443 in this set, as it is generally our
recommendation to avoid proxying HTTPS traffic, since the proxy provides
very little value on these connections today. This may change in the
future, but it seems like a good default for the 2.9 release.


charts/linkerd2/values.yaml
```diff
@@ -182,10 +185,10 @@ proxyInjector:
   # Namespace selector used by admission webhook
   namespaceSelector:
     matchExpressions:
-    - key: config.linkerd.io/admission-webhooks
-      operator: NotIn
-      values:
-      - disabled
+      - key: config.linkerd.io/admission-webhooks
```
adleong, line 188: This extra indentation for lists seems unrelated to this change and also is inconsistent with how yaml lists are formatted elsewhere.

adleong (commented):
Looks like there's a fixture in an integration test  that needs to be updated.

charts/linkerd2/values.yaml
olix0r: ugh vscode

Pothulapati:
One important thing to note here is that, Whenever users overwrite these prots through a annotation/flag, we **don't append but replace the defaults with the given** and this is [why some inject golden files with skip port overrides were not updated  in this change](https://github.com/linkerd/linkerd2/blob/1c358c6246236c3c1ee62b710cb3899ac946c666/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml#L167)

Previously, this was not the case right as the proxy was doing this? Should we change this logic? or atleast have some documentation that if they override then they have to also include these common ports?

Pothulapati:
We might instead want to put this [inside the proxy-init template](https://github.com/linkerd/linkerd2/blob/d3553c59fdb6ec1824eaebc39cc7325494a45779/charts/partials/templates/_proxy-init.tpl#L10) which we already do for 4190 to make them appear, irrespective of overrides.

zaharidichev:
I think the title of the pr makes that a bit misleading as this does not only pertain to helm but to CLI  flow as well. Right ?

olix0r:
> Whenever users overwrite these prots through a annotation/flag, we don't append but replace the defaults with the given

Yeah, I think that's fine. Ports like 4190 are part of the internal implementation of Linkerd (and are configurable). These are just defaults, and I think users should have the power to change them. I don't think we should have hardcoded ports that users cannot change.

adleong (approved):
🛹 📻

grampelberg (approved):
:shipit: We should make sure to update the docs around this as it can be confusing for folks upgrading.

PR #5101, opened 2020-10-19
Fix integration tests to run better on ARM

adleong:
In order for the integration tests to run successfully on a dedicated ARM cluster, two small changes are necessary:

* We need to skip the multicluster test since this test uses two separate clusters (source and target)
* We need to properly uninstall the multicluster helm chart during cleanup.

With these changes, I was able to successfully run the integration tests on a dedicated ARM cluster.

### DIFF
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -263,6 +263,11 @@ start_kind_test() {
 }
 
 start_k3d_test() {
+  if [ -n "$RUN_ARM_TEST" ]; then
+    echo "Skipped because ARM tests run on a dedicated cluster."
+    exit 0
+  fi
+
   test_setup
   if [ -z "$skip_cluster_create" ]; then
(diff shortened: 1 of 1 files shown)


bin/_test-helpers.sh
```diff
@@ -263,6 +263,11 @@ start_kind_test() {
 }
 
 start_k3d_test() {
+  if [ -n "$RUN_ARM_TEST" ]; then
+    echo "Skipped because ARM tests run on a dedicated cluster."
+    exit 0
```
alpeb, line 268: Doesn't `exit 0` exit the script altogether and skip the tests that come after the multicluster one?

alpeb (approved):
LGTM :+1:

kleimkuhler (approved):
👍

olix0r:
We should reenable these tests on edge releases, either as a part of this change or as  separate followon

PR #5107, opened 2020-10-20
inject: Use 'quote' function in proxy template

### LINKED ISSUE #5105: `--proxy-log-level=off` produces broken yaml
Using a command like `linkerd install --proxy-log-level=off` produces broken YAML. It appears that `off` is converted to `false`:
```
Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"na|..., bigger context ...|}]},{"env":[{"name":"LINKERD2_PROXY_LOG","value":false},{"name":"LINKERD2_PROXY_LOG_FORMAT","value":|...
```

`off` is a valid log level, and it's useful to get the absolute best performance out of the proxy.

I suppose we need to add quotes to our template?

olix0r:
As described in #5105, it's not currently possible to set the proxy log
level to `off`. The proxy injector's template does not quote the log
level value, and so the `off` value is handled as `false`. Thanks, YAML.

This change updates the proxy template to use helm's `quote` function
throughout, replacing manually quoted values and fixing the quoting for
the log level value.

We also remove the default `logFormat` value, as the default is specified
in values.yaml.

Fixes #5105

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -5,7 +5,7 @@ env:
   value: "{{.Values.global.proxy.requireIdentityOnInboundPorts}}"
 {{ end -}}
 - name: LINKERD2_PROXY_LOG
-  value: {{.Values.global.proxy.logLevel}}
+  value: "{{.Values.global.proxy.logLevel | default "linkerd=info,warn" }}"
 - name: LINKERD2_PROXY_LOG_FORMAT
   value: {{.Values.global.proxy.logFormat | default "plain"}}
 - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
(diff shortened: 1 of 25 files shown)


charts/partials/templates/_proxy.tpl
```diff
@@ -5,7 +5,7 @@ env:
   value: "{{.Values.global.proxy.requireIdentityOnInboundPorts}}"
 {{ end -}}
 - name: LINKERD2_PROXY_LOG
-  value: {{.Values.global.proxy.logLevel}}
+  value: "{{.Values.global.proxy.logLevel | default "linkerd=info,warn" }}"
```
adleong, line 8: The default for this value from the chart is already "warn,linkerd=info" so this value should never be empty at this point.  I think having defaulting behavior here also just confuses things.  (for what it's worth, that's also true about the defaulting behavior below in the log format variable and probably elsewhere throughout these templates)

adleong (approved):
🌳 📴

alpeb (approved):
LGTM :+1:

PR #5109, opened 2020-10-20
Remove defunct proxy config variables

olix0r:
The proxy no longer honors DESTINATION_GET variables, as profile lookups
inform when endpoint resolution is performed.  Also, there is no longer
a router capacity limit.


adleong (approved):

alpeb (approved):
LGTM

PR #5110, opened 2020-10-20
Replace global.proxy.destinationGetNetworks with global.clusterNetworks

olix0r:
There is no longer a proxy config `DESTINATION_GET_NETWORKS`. Instead of
reflecting this implementation in our values.yaml, this changes this
variable to the more general `clusterNetworks` to emphasize its
similarity to `clusterDomain` for the purposes of discovery.


adleong (approved):
🍇 🥅 💼

kleimkuhler (approved):

PR #5111, opened 2020-10-20
Do not permit cluster networks to be overridden per-pod

olix0r:
In #5110 the `global.proxy.destinationGetNetworks` configuration is
renamed to `global.clusterNetworks` to better reflect its purpose.

The `config.linkerd.io/proxy-destination-get-networks` annotation allows
this configuration to be overridden per-workload, but there's no real use
case for this. I don't think we want to support this value differing
between pods in a cluster. No good can come of it.

This change removes support for the `proxy-destination-get-networks`
annotation.


alpeb (approved):
LGTM :+1:

Pothulapati (approved):
LGTM :shipit:

adleong (approved):

PR #5113, opened 2020-10-20
Allow setting a list of service accounts in multicluster install

### LINKED ISSUE #5098: Can't use Helm for installing multicluster access credentials
[The multicluster installation doc](https://github.com/linkerd/website/pull/782/files#diff-fae783921db4c9259ab03260024736ceb39c4f77feca6a5c46d79e1d7b98930fR343) describes a workflow for installing the RBAC to grant access for a source cluster to access the current (target) cluster through Helm:
```
helm install linkerd2-mc-source linkerd/linkerd2-multicluster --set gateway=false --set remoteMirrorServiceAccountName=source --set installNamespace=false --kube-context target
```
This reuses the same `linkerd2-multicluster` chart that was used for installing the multicluster components, stripping off the gateway. But that command results in the following error:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "links.multicluster.linkerd.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "linkerd2-mc-source": current value is "linkerd2-multicluster"
```
Even after moving the CRD into `templates/crds` and using `--skip-crds` the error persists, because apparently this way of reusing charts to generate different Helm "releases" in the same cluster is not allowed.

One possible way around this would be to separate that RBAC into its own separate chart, to be used directly in this scenario when setting up additional access, and indirectly when installing the multicluster component to provide the default access.

adleong:
Fixes #5098 

When setting up multicluster, a target cluster may wish to create multiple service accounts to be used by source clusters' service mirrors.  This allows the target cluster to individually revoke access to each of the source clusters.  When using the Linkerd CLI, this can be accomplished by running the `linkerd multicluster allow` command multiple times to create multiple service accounts.  However, there is no analogous workflow when installing with Helm.

We update the Helm templates to support interpreting the `remoteMirrorServiceAccountName` value as either a single string or a list of strings.  In the case where it is a list, we create a service account and associated RBAC for each entry in the list.


charts/linkerd2-multicluster/templates/remote-access-service-mirror-rbac.yaml
```diff
@@ -19,25 +24,25 @@ rules:
...
+  name: {{.}}
 subjects:
 - kind: ServiceAccount
-  name: {{.Values.remoteMirrorServiceAccountName}}
-  namespace: {{.Values.namespace}}
+  name: {{.}}
+  namespace: {{$.Values.namespace}}
 {{end -}}
-
+{{end -}}
```
olix0r, line 48: trailing newlines are nice

olix0r (approved):
lgtm

alpeb (commented):
Do you mind updating the multicluster README entry as well? :wink:

alpeb (approved):
LGTM barring the README update. I also added a comment into the mc docs PR referring to this.

PR #5114, opened 2020-10-20
Fixed multicluster Grafana chart

alpeb:
The graphs were empty because they were relying on the metric label `dst_target_gateway` which is no longer relevant.

When mirroring the `emoji` and `voting` services after this fix:

image


adleong (approved):

kleimkuhler (approved):
👍

PR #5117, opened 2020-10-21
Restrict controlPlaneTracing field only to control plane components

Pothulapati:
Previously, `global.controlPlaneTracing` was not available during
application injections and thus not affecting them. It was restricted to
install only.

But now that it is stored and retrieved, It causes problems when
`global.controlPlaneTracing` is enabled while no `trace-collector` value
is present or specified(through flags) during injection.

With the latest changes, This can be further be simplified by
removing `controlPlaneTracing` altogether and have trace flags 
under `install` command do the same.
This effort could be part of the minimal control plane changes.

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -39,7 +39,7 @@ env:
   value: 10000ms
 - name: LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE
   value: 10000ms
-{{ if or (.Values.global.proxy.trace.collectorSvcAddr) (.Values.global.controlPlaneTracing) -}}
+{{ if or (.Values.global.proxy.trace.collectorSvcAddr) (and (hasPrefix "linkerd-" .Values.global.proxy.component) .Values.global.controlPlaneTracing) -}}
 - name: LINKERD2_PROXY_TRACE_ATTRIBUTES_PATH
   value: /var/run/linkerd/podinfo/labels
 {{ end -}}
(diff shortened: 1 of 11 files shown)


olix0r:
Should this field be removed from the Values struct, etc?

olix0r:
Disregard my comment, I see that the field is still used.

Pothulapati:
Yeah @olix0r. I did think about removing that field altogether but then it adds unecessary changes and could be a good candidate with the control-plane chanes as these things will move around then.

adleong:
Checking the prefix of the component name to determine if a component is a control plane component and then changing the tracing configuration based on that feels brittle.

Instead, could we have the control plane component templates explicitly set the `.Values.global.proxy.trace.collectorSvcAddr` variable to the correct value (depending on if control plane tracing is enabled or not) before evaluating the proxy partial just like they do for the component name?

This allows the proxy partial to be unaware of the control plane tracing value altogether.

charts/partials/templates/_proxy.tpl
```diff
@@ -39,7 +39,7 @@ env:
   value: 10000ms
 - name: LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE
   value: 10000ms
-{{ if or (.Values.global.proxy.trace.collectorSvcAddr) (.Values.global.controlPlaneTracing) -}}
+{{ if or (.Values.global.proxy.trace.collectorSvcAddr) (and (hasPrefix "linkerd-" .Values.global.proxy.component) .Values.global.controlPlaneTracing) -}}
```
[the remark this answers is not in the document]
Pothulapati: My thinking is that if `collectorSvcAddr` is set for `controlPlaneTracing`, it could be then propogated and saved into `linkerd-config` which is used with application injections and making injections also have tracing fields set to the same.

Trying to find away to only have them set temporarily not stored.

adleong:
@Pothulapati this is what I had in mind: https://github.com/linkerd/linkerd2/compare/alex/tracing-partial

Notice that we make a copy of the values and only pass that copy into the proxy partial for the control plane deployments.  This means that we can check the value of the controlPlaneTracing field and correctly set the collector address and service account in the copy of the values which are passed to the partial.  These changes are not persisted to the linkerd-config.  

This also allows the proxy partial to be unaware of control plane tracing, instead just using the address and service account that it is given.

My branch is a proof of concept that demonstrates how this is done with the identity controller.  We would need to make this change in all control plane components.

charts/partials/templates/_trace.tpl
```diff
@@ -3,3 +3,10 @@
 - -trace-collector=linkerd-collector.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:55678
 {{ end -}}
 {{- end }}
+
+{{ define "partials.linkerd.trace.proxy" -}}
```
adleong, line 7: Can we give this a more descriptive name like "configureControlPlaneTracing" or something like that?

Pothulapati: named it `partials.controlPlaneTracing.proxy` to signify that it's changing the proxy side of tracing variables.

adleong: I don't want to block this over a naming bikeshed but this name is still pretty confusing to me.  This partial is a bit unusual in that it doesn't actually include any renderable template; instead it just sets some variables.  So when I first saw this, I was confused, thinking that it was rendering the proxy container or a control plane tracing version of the proxy container, or something like that.

I'd prefer a name that contains a word like "configure" or "set" or something else to indicate that this partial just sets values rather than rendering anything directly.

adleong (approved):

kleimkuhler (approved):
Looks good!

PR #5100, opened 2020-10-19
Small fixes to linkerd2 chart README.md

alpeb:
Fixes #5092: a couple of entries should go under `global` and fixed a typo.

PR #5119, opened 2020-10-21
Remove code refs to controllerImageVersion

alpeb:
Followup to #5100

We had both `controllerImageVersion` and `global.controllerImageVersion`
configs, but only the latter was taken into account in the chart
templates, so this change removes all the references to the former.


olix0r (approved):

adleong (approved):

Pothulapati (approved):
LGTM

PR #5120, opened 2020-10-21
Trigger ARM int tests for edge releases as well (#5073)

adleong:
Used to be triggered only for stable releases, but now that 2.9 stable
approaches let's turn it on for the upcoming RCs.


alpeb (approved):
LGTM :+1:

olix0r (approved):

PR #5122, opened 2020-10-22
cli: fix and update timeout warnings in profile cmd

### LINKED ISSUE #5121: Warnings under `linkerd profile` command
### What is the issue?

Running `linkerd profile` gives a warinng followed by the correct output.

```bash
1022 12:16:06.292679  346983 transport.go:243] Unable to cancel request for promhttp.RoundTripperFunc
error reading byte stream header: error while reading message length: net/http: request canceled (Client.Timeout or context cancellation while reading body)
kind: ServiceProfile
metadata:
  creationTimestamp: null
  name: web-svc.emojivoto.svc.cluster.local
  namespace: emojivoto
spec:
  routes:
  - condition:
      method: GET
      pathRegex: /api/list
    name: GET /api/list
  - condition:
      method: GET
      pathRegex: /api/vote
    name: GET /api/vote

```

This happened with other deployments/services under the `profile` command. This happens both in `2.8.1` and the current edges.

```bash
⛵ test in linkerd2 on  main [$?] via 🐹 v1.14.9 on ☁️  ap-southeast-1
❯ linkerd version
Client version: stable-2.8.1
Server version: stable-2.8.1

⛵ test in linkerd2 on  main [$?] via 🐹 v1.14.9 on ☁️  ap-southeast-1
❯ linkerd profile -n emojivoto web-svc --tap deploy/web --tap-duration 10s
W1022 12:22:09.102075  359480 transport.go:243] Unable to cancel request for promhttp.RoundTripperFunc
error reading byte stream header: error while reading message length: context deadline exceeded (Client.Timeout or context cancellation while reading body)
kind: ServiceProfile
metadata:
  creationTimestamp: null
  name: web-svc.emojivoto.svc.cluster.local
  namespace: emojivoto
spec:
  routes:
  - condition:
      method: GET
      pathRegex: /api/list
    name: GET /api/list
  - condition:
      method: GET
      pathRegex: /api/vote
    name: GET /api/vote
⛵ kind-kind (default) in linkerd2 on  main [$?] via 🐹 v1.14.9 on ☁️  ap-southeast-1
❯ ./bin/linkerd version
Client version: git-79893c9a
Server version: git-79893c9a
⛵ kind-kind (default) in linkerd2 on  main [$?] via 🐹 v1.14.9 on ☁️  ap-southeast-1 took 10s
❯ ./bin/linkerd profile -n linkerd-emojivoto web-svc --tap deploy/web --tap-duration 10s
error reading byte stream header: error while reading message length: context deadline exceeded (Client.Timeout or context cancellation while reading body)
kind: ServiceProfile
metadata:
  creationTimestamp: null
  name: web-svc.linkerd-emojivoto.svc.cluster.local
  namespace: linkerd-emojivoto
spec:
  routes:
  - condition:
      method: GET
      pathRegex: /api/list
    name: GET /api/list
  - condition:
      method: GET
      pathRegex: /api/vote
    name: GET /api/vote
```

### How can it be reproduced?

Run `linkerd profile`
### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version:
- Host OS:
- Linkerd version: Both 2.8.1 and latest edges

Pothulapati:
Fixes #5121

This PR has the following changes
- Updates the warning messages to the latest ones, that are emitted when the `tapDuration` deadline is crossed.
- The usage of `client.Timeout` does not seem to work correctly with k8s Client and causes `W1022
17:20:12.372780   19049 transport.go:260] Unable to cancel request for promhttp.RoundTripperFunc` to be emitted [from the Kubernetes Client.](https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/staging/src/k8s.io/client-go/transport/transport.go#L260:18) This is fixed by using `context.WithDeadline` and passing that into the HTTP Request. `tap.Render` now takes a `context` instead of a timeout

### DIFF
--- a/pkg/profiles/tap.go
+++ b/pkg/profiles/tap.go
@@ -81,7 +81,7 @@ func routeSpecFromTap(tapByteStream *bufio.Reader, routeLimit int) []*sp.RouteSp
 		if err != nil {
 			// expected errors when hitting the tapDuration deadline
 			if err != io.EOF &&
-				!strings.HasSuffix(err.Error(), "(Client.Timeout exceeded while reading body)") &&
+				!strings.HasSuffix(err.Error(), "(Client.Timeout or context cancellation while reading body)") &&
 				!strings.HasSuffix(err.Error(), "http2: response body closed") {
 				fmt.Fprintln(os.Stderr, err)
 			}
--- a/pkg/profiles/tap.go
+++ b/pkg/profiles/tap.go
@@ -2,6 +2,7 @@ package profiles
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -23,7 +24,7 @@ import (
 // RenderTapOutputProfile performs a tap on the desired resource and generates
 // a service profile with routes pre-populated from the tap data
 // Only inbound tap traffic is considered.
-func RenderTapOutputProfile(k8sAPI *k8s.KubernetesAPI, tapResource, namespace, name, clusterDomain string, tapDuration time.Duration, routeLimit int, w io.Writer) error {
+func RenderTapOutputProfile(ctx context.Context, k8sAPI *k8s.KubernetesAPI, tapResource, namespace, name, clusterDomain string, tapDuration time.Duration, routeLimit int, w io.Writer) error {
 	requestParams := util.TapRequestParams{
 		Resource:  tapResource,
 		Namespace: namespace,
@@ -35,7 +36,7 @@ func RenderTapOutputProfile(k8sAPI *k8s.KubernetesAPI, tapResource, namespace, n
 		return err
 	}
 
-	profile, err := tapToServiceProfile(k8sAPI, req, namespace, name, clusterDomain, tapDuration, routeLimit)
+	profile, err := tapToServiceProfile(ctx, k8sAPI, req, namespace, name, clusterDomain, tapDuration, routeLimit)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


alpeb (approved):
Thanks @Pothulapati. Using ctx for this does feel better than mutating the client :+1:

pkg/profiles/tap.go
```diff
@@ -81,7 +84,8 @@ func routeSpecFromTap(tapByteStream *bufio.Reader, routeLimit int) []*sp.RouteSp
 		if err != nil {
 			// expected errors when hitting the tapDuration deadline
 			if err != io.EOF &&
-				!strings.HasSuffix(err.Error(), "(Client.Timeout exceeded while reading body)") &&
+				!strings.HasSuffix(err.Error(), "(Client.Timeout or context cancellation while reading body)") &&
```
adleong, line 87: Can we use `errors.Is` here instead of string matching? https://golang.org/pkg/errors/#Is

Pothulapati: As we are wrapping the underlying error with others through the returns in the call stack, looks like we can't do this unless we used a separate error type that has this underlying i.e https://blog.golang.org/go1.13-errors

adleong: I think we can fix this by using the %w for wrapped errors in `fmt.Errrorf`.  If that's easy to do it would be nice to clean that up now while we're in here.  Otherwise, I guess we can fix in a separate PR.

Pothulapati: Looks like not all error types used here, are exposed from go std library. :\

- https://sourcegraph.com/github.com/golang/go/-/blob/src/net/http/client.go#L723:19
- https://sourcegraph.com/github.com/golang/go/-/blob/src/net/http/h2_bundle.go#L8642

I was able to find only one instance here where this could be used but there are other places in other files where this could be done. I will raise an issue and fix this in a separate PR.

adleong: I believe that this httpError implements the `net.Error` interface.  So I think we can use `errors.As` to unwrap the error into a `net.Error` and then check if it's a timeout with the `Timeout()` method on the `net.Error` interface.

Pothulapati: Hmm, We can do that. Riased https://github.com/linkerd/linkerd2/pull/5140 to work on this.

adleong (approved):

PR #5123, opened 2020-10-22
Changes for edge-20.10.4

cpretzer:

## edge-20.10.4

This edge release is a release candidate for stable-2.9.0. For the proxy, there
have been changes to improve performance, remove unused code, and configure
ports that can be ignored by default. Also, this edge release adds enhancements
to the multicluster configuration and observability, adds more translations to
the dashboardand addresses a bug in the CLI.

* Added more translations to the dashboard and validation tests
* Added support for a list of service accounts in multicluster install with
  helm
* Removed per-pod overrides for cluster network configuration, unused proxy
  variables, and renamed `global.proxy.destinationGetNetworks` to
  `global.clusterNetworks`
* Fixed an empty multicluster Grafana graph which used a deprecated label
* Added the tracing helm subchart ServiceAccounts to the `linkerd-psp`
  RoleBinding for use in environments where PodSecurityPolicy is enabled
* Enhanced EKS support by adding `100.64.0.0/10` to the set of discoverable
  networks
* Fixed a bug in the way that the `--all-namespaces` flag is handled by the
  `linkerd edges` command
* Added a default set of ports to bypass the proxy for server-first, https,
  and memcached traffic

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,17 +6,18 @@ This edge release is a release candidate for stable-2.9.0. For the proxy, there
 have been changes to improve performance, remove unused code, and configure
 ports that can be ignored by default. Also, this edge release adds enhancements
 to the multicluster configuration and observability, adds more translations to
-the dashboardand addresses a bug in the CLI.
+the dashboard, and addresses a bug in the CLI.
 
-* Added more translations to the dashboard and validation tests
-* Added support for a list of service accounts in multicluster install with
-  helm
-* Removed per-pod overrides for cluster network configuration, unused proxy
-  variables, and renamed `global.proxy.destinationGetNetworks` to
-  `global.clusterNetworks`
+* Added more Spanish translations to the dashboard and more labels that can be
+  translated
+* Added support for creating multiple service accounts when installing
+  multicluster with Helm to allow more granular revocation
+* Renamed `global.proxy.destinationGetNetworks` to `global.clusterNetworks`.
+  This is a cluster-wide setting and can no longer be overridden per-pod
 * Fixed an empty multicluster Grafana graph which used a deprecated label
-* Added the tracing helm subchart ServiceAccounts to the `linkerd-psp`
-  RoleBinding for use in environments where PodSecurityPolicy is enabled
+* Added the control plane tracing ServiceAccounts to the linkerd-psp
+  RoleBinding so that it can be used in environments where PodSecurityPolicy
+  is enabled
 * Enhanced EKS support by adding `100.64.0.0/10` to the set of discoverable
   networks
 * Fixed a bug in the way that the `--all-namespaces` flag is handled by the
@@ -171,7 +172,7 @@ topology preferences. Additionally, this release includes bug fixes for the
     for service profiles when no traffic split is present
 * Web UI
   * Fixed Tap `Authority` dropdown not being populated (thanks to @tharun208!)
-  
+
 [topology]: https://kubernetes.io/docs/concepts/services-networking/service-topology/
 
 ## edge-20.8.2
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,29 @@
 # Changes
 
+## edge-20.10.4
+
+This edge release is a release candidate for stable-2.9.0. For the proxy, there
+have been changes to improve performance, remove unused code, and configure
+ports that can be ignored by default. Also, this edge release adds enhancements
+to the multicluster configuration and observability, adds more translations to
+the dashboardand addresses a bug in the CLI.
+
+* Added more translations to the dashboard and validation tests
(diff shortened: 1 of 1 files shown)


admc (approved):
👍

CHANGES.md
```diff
@@ -1,5 +1,29 @@
...
 
+## edge-20.10.4
+
+This edge release is a release candidate for stable-2.9.0. For the proxy, there
+have been changes to improve performance, remove unused code, and configure
+ports that can be ignored by default. Also, this edge release adds enhancements
+to the multicluster configuration and observability, adds more translations to
+the dashboardand addresses a bug in the CLI.
+
+* Added more translations to the dashboard and validation tests
```
adleong, line 11: Don't think we have to mention the tests.  Can we be more specific about the translations?  We just expanded the spanish translation, right?

CHANGES.md
```diff
@@ -1,5 +1,29 @@
...
+* Added support for a list of service accounts in multicluster install with
```
adleong, line 12: "Added support for creating multiple service accounts when installing multicluster with Helm to allow more granular revocation"

CHANGES.md
```diff
@@ -1,5 +1,29 @@
...
+  helm
+* Removed per-pod overrides for cluster network configuration, unused proxy
```
adleong, line 14: "Renamed `global.proxy.destinationGetNetworks` to `global.clusterNetworks`.  This is a cluster-wide setting and can no longer be overridden per-pod"

CHANGES.md
```diff
@@ -1,5 +1,29 @@
...
+  `global.clusterNetworks`
+* Fixed an empty multicluster Grafana graph which used a deprecated label
+* Added the tracing helm subchart ServiceAccounts to the `linkerd-psp`
+  RoleBinding for use in environments where PodSecurityPolicy is enabled
+* Enhanced EKS support by adding `100.64.0.0/10` to the set of discoverable
+  networks
+* Fixed a bug in the way that the `--all-namespaces` flag is handled by the
+  `linkerd edges` command
+* Added a default set of ports to bypass the proxy for server-first, https,
+  and memcached traffic
```
hawkw, line 25: should we mention any of the proxy changes here as well? in particular, I think there was a fix for bad tap output.

hawkw, line 25: memcached is skipped _because_ it's server-first, right? the wording here doesn't make that super clear IMO

CHANGES.md
```diff
@@ -1,5 +1,29 @@
...
+the dashboardand addresses a bug in the CLI.
+
+* Added more translations to the dashboard and validation tests
+* Added support for a list of service accounts in multicluster install with
+  helm
+* Removed per-pod overrides for cluster network configuration, unused proxy
+  variables, and renamed `global.proxy.destinationGetNetworks` to
+  `global.clusterNetworks`
+* Fixed an empty multicluster Grafana graph which used a deprecated label
+* Added the tracing helm subchart ServiceAccounts to the `linkerd-psp`
```
adleong, line 18: "Added the control plane tracing ServiceAccounts to the `linkerd-psp` RoleBinding so that it can be used in environments where PodSecurityPolicy is enabled"

CHANGES.md
hawkw: also, maybe we should mention the actual port numbers that are skipped?

CHANGES.md
hawkw, line 18: nit: i think Helm should be a proper noun
```suggestion
* Added the tracing Helm subchart ServiceAccounts to the `linkerd-psp`
```

CHANGES.md
```diff
@@ -1,5 +1,29 @@
 # Changes
 
+## edge-20.10.4
+
+This edge release is a release candidate for stable-2.9.0. For the proxy, there
+have been changes to improve performance, remove unused code, and configure
+ports that can be ignored by default. Also, this edge release adds enhancements
+to the multicluster configuration and observability, adds more translations to
+the dashboardand addresses a bug in the CLI.
```
hawkw, line 9: typo:
```suggestion
the dashboard, and addresses a bug in the CLI.
```

adleong (approved):
Some non-blocking suggestions.

CHANGES.md
cpretzer: The notes said that memcached is skipped because "clients do not issue any sort of preamble that is immediately detectable." 

Is that the same as server-first?

CHANGES.md
cpretzer: Good catch, thanks!

CHANGES.md
cpretzer: The bad tap output was a fix for a regression in a past edge. I opted to leave it out since it doesn't affect the last stable version.

Should it be included in this case?

PR #5125, opened 2020-10-22
Ignore cni-calico-deep integration tests for ARM

alpeb:
... because those are tailored for kind, but we're running k3s in the arm box.

### DIFF
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -482,6 +482,11 @@ run_deep_test() {
 }
 
 run_cni-calico-deep_test() {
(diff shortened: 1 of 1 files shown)


alpeb:
Some validation:
```
$ RUN_ARM_TEST=1 bin/tests --name cni-calico-deep --skip-cluster-create /home/alpeb/.linkerd2/bin/linkerd
Checking if Linkerd resources exist on cluster...[ok]
Skipped because the cni-calico-deep test is not supported in k3s yet.
```

kleimkuhler (approved):
👍

PR #5126, opened 2020-10-22
Clear endpoint state in EndpointTranslator

adleong:
The EndpointTraslator holds the full set of endpoints that it knows about as well as a filtered set of these addresses that it has sent to clients.  It needs both of these so that it can properly compute the correct diff to send to clients when it gets updates.  When it gets a NoEndpoints message, it passes this message on to clients but does not clear its own state.  If those same endpoints are subsequently added back, it will not detect this as a diff and will fail to send the update to clients.

We now clear out the full endpoint set and the filtered set from the EndpointTranslator state when it gets a NoEndpoints message.


alpeb (approved):
Makes sense to me :+1:

cpretzer (approved):
Nice fix!

PR #5128, opened 2020-10-23
Don't skip pushing images if tag is already there

alpeb:
gcr.io and quay.io have an issue that it's not possible to update multi-arch images
(see eclipse/che#16983 and open-policy-agent/gatekeeper#665).

We're now relying on ghcr.io instead, which I verified doesn't have this
bug, so we can stop skipping these pushes.


cpretzer (approved):
In the context of using ghcr.io, this makes sense. If we end up switching back to gcr for some reason I hope they'll have fixed this by then.

kleimkuhler (approved):

PR #5130, opened 2020-10-24
Add 'linkerd.io/inject: ingress' mode

### LINKED ISSUE #5118: Support service profiles on ingress proxies.
Ingresses do not communicate with service ClusterIPs and it seems to be almost impossible to make them do that. Instead, they load balance requests over connections to each pod. Because the proxy now does discovery based on the target address (and not any header), the proxy cannot discover any service-level metadata on these requests: service profiles are not loaded, traffic splits are not applied, and the proxy does no load balancing. On one hand, this is nice because the proxy can now honor the ingress's balancing/routing choices; but, on the other hand, this is a significant regression. People rely on both service profile metadata and traffic splits between their ingress and server.

To restore the prior behavior, I propose the following:

We introduce a new proxy configuration flag that causes the proxy to operate in _http ingress mode_. In this mode the proxy behaves as follows:
* The proxy does not discover profile information before protocol detection. This effectively means that we do not support any server-first protocols through the ingress. I can't imagine how that would even work, anyway. We should support plain TCP forwarding for cases when the ingress is communicating over TLS with a backend.
* For HTTP requests, the `l5d-dst-override` header is **always** read (and **never** the `:authority` or `Host` value). If the header is not present on a request, the original dst address is used (as in the non-ingress case).
* When the header is present, the proxy does profile discovery, honors traffics splits, and does its own load balancing.
* When the header is not present, discovery is done on the TCP target and connections may be TLS'd but service metadata will not apply (unless the target includes a service's ClusterIP, but as I mention above, this is unlikely).

In the future, we may opt to change inbound logic in ingress mode (i.e. to forward to the ingress with no additional processing), but this should be considered out of scope for 2.9.

### Questions

* How does this get configured? Can the injector discover the ingressness of a workload? Or should this be configured by a workload annotation?

adleong:
Fixes #5118

This PR adds a new supported value for the `linkerd.io/inject` annotation.  In addition to `enabled` and `disabled`, this annotation may now be set to `ingress`.  This functions identically to `enabled` but it also causes the `LINKERD2_PROXY_INGRESS_MODE="true"` environment variable to be set on the proxy.  This causes the proxy to operate in ingress mode as described in #5118 

With this set, ingresses are able to properly load service profiles based on the l5d-dst-override header.

### DIFF
--- a/charts/partials/templates/_proxy.tpl
+++ b/charts/partials/templates/_proxy.tpl
@@ -32,6 +32,10 @@ env:
 - name: LINKERD2_PROXY_INBOUND_GATEWAY_SUFFIXES
   value: {{printf "svc.%s." .Values.global.clusterDomain}}
 {{ end -}}
+{{ if .Values.global.proxy.isIngress -}}
+- name: LINKERD2_PROXY_INGRESS_MODE
+  value: "true"
+{{ end -}}
 - name: LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES
   {{- $internalDomain := printf "svc.%s." .Values.global.clusterDomain }}
   value: {{ternary "." $internalDomain .Values.global.proxy.enableExternalProfiles}}
--- a/cli/cmd/testdata/install_addon.golden
+++ b/cli/cmd/testdata/install_addon.golden
@@ -868,6 +868,7 @@ data:
           version: install-proxy-version
         inboundConnectTimeout: 100ms
         isGateway: false
+        isIngress: false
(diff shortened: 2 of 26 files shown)


kleimkuhler (approved):
This looks good 👍

olix0r:
Marking this as a draft until the necessary proxy changes are available in this repo.

olix0r:
Proxy upgrade: https://github.com/linkerd/linkerd2/pull/5137

olix0r (approved):

PR #5137, opened 2020-10-26
proxy: v2.117.0

olix0r:
This release adds an 'ingress mode' to support per-request routing for
HTTP ingresses.

Additionally, the performance impact of logging should be reduced when
the proxy log level is not set to `debug` or `trace`.

---

* router: Use NewService instead of MakeService (linkerd/linkerd2-proxy#724)
* outbound: Split TCP stack into dedicated modules (linkerd/linkerd2-proxy#725)
* trace: update `tracing-subscriber` to 0.2.14 (linkerd/linkerd2-proxy#726)
* outbound: Extract HTTP and server modules (linkerd/linkerd2-proxy#727)
* outbound: Introduce 'ingress mode' (linkerd/linkerd2-proxy#728)
* Reduce tracing spans to the debug level (linkerd/linkerd2-proxy#730)


hawkw (approved):
> Additionally, the performance impact of logging should be reduced when
> the proxy log level is not set to `debug` or `trace`.

FWIW, linkerd/linkerd2-proxy#726 means that it should be reduced even when `debug` or `trace` _are_ enabled, but the other changes should reduce it even more when they aren't... :)

kleimkuhler (approved):

PR #5140, opened 2020-10-27
Use errors.Is instead of checking underlying err messages

### LINKED ISSUE #5132: Try using `errors.is` instead of checking for error messages
### What is the issue?

As underlying error messages can change, It's important that we better check on the relevant error type rather than checking on error messages.

Currently, `strings.hasSuffix` is used to match the underlying error message at multiple places.

Pothulapati:
Fixes #5132

This PR replaces the usage of `strings.hasSuffix` with `errors.Is`
wherever error messages are being checked. So, that the code is not
effected by changes in the underlying message.

### DIFF
--- a/pkg/tap/tap.go
+++ b/pkg/tap/tap.go
@@ -3,6 +3,7 @@ package tap
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -14,6 +15,11 @@ import (
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/protohttp"
 	log "github.com/sirupsen/logrus"
+)
+
+var (
+	// ErrClosedResponseBody is returned when response body is closed in http2
+	ErrClosedResponseBody = errors.New("http2: response body closed")
 )
 
 // TapRbacURL is the link users should visit to remedy issues when attempting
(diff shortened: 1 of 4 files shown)


pkg/tap/tap.go
```diff
@@ -16,6 +17,11 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var (
+	// ErrClosedResponseBody is returned when response body is closed in http2
+	ErrClosedResponseBody = errors.New("http2: response body closed")
```
adleong, line 22: It looks like we check for this error: `errors.Is(err, tap.ErrClosedResponseBody)` but we never actually return this error.  I notice this looks the same as the error from http2/transport:

https://go.googlesource.com/net/+/master/http2/transport.go#2158

But I don't think creating our own error with the same text will let us catch that one.  Unfortunately, that error is private and doesn't seem to implement any useful public interfaces, so this may be a case where string matching on the error message is the only way.

adleong (commented):
This is awesome!  Really nice cleanup, I'm very excited to eliminate string matching for errors wherever we can.

pkg/tap/tap.go
Pothulapati: @adleong 

>But I don't think creating our own error with the same text will let us catch that one

Ah you are right!

My reason for moving it into a separate type is that we can have a single place where we can update the error message (if at all it changes), instead of multiple places as it is used at a couple of places right now. I will instead create a string const type to be used.

kleimkuhler (approved):
👍

adleong (approved):

PR #5141, opened 2020-10-27
Fix some warning: redundant type from array, slice, or map composite literal

Hellcatlk:
Signed-off-by: zouyu <zouy.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/inject/inject_test.go
+++ b/pkg/inject/inject_test.go
@@ -243,9 +243,9 @@ func TestGetOverriddenValues(t *testing.T) {
 					},
 					Spec: corev1.PodSpec{
 						Containers: []corev1.Container{
-							corev1.Container{
+							{
 								Ports: []corev1.ContainerPort{
-									corev1.ContainerPort{
+									{
 										Name:          "mysql",
 										ContainerPort: 3306,
 									},


adleong:
Thanks, @Hellcatlk!  We're probably not totally consistent in this codebase about which style we use (including the struct names or not) and it would be great to standardize.  It seems that with our current golangci settings, the `bin/lint` script doesn't have an opinion on this.  Which tool are you using to see these warnings?  If we can add this as a linting rule, we can be more consistent about it going forward.

Hellcatlk:
The warnings report from `gopls`, but i don't know what lint is used by `gopls`. But gofmt can fix this:
```cmd
$ gofmt -s -d pkg/inject/inject_test.go
diff -u pkg/inject/inject_test.go.orig pkg/inject/inject_test.go
--- pkg/inject/inject_test.go.orig      2020-10-28 10:42:13.453974145 +0800
+++ pkg/inject/inject_test.go   2020-10-28 10:42:13.453974145 +0800
@@ -243,9 +243,9 @@
                                        },
                                        Spec: corev1.PodSpec{
                                                Containers: []corev1.Container{
-                                                       corev1.Container{
+                                                       {
                                                                Ports: []corev1.ContainerPort{
-                                                                       corev1.ContainerPort{
+                                                                       {
                                                                                Name:          "mysql",
                                                                                ContainerPort: 3306,
                                                                        },
```

adleong:
Ah, that makes sense!  We use `goimports` for formatting in this project (which you can invoke by running `bin/fmt`) and golangci-lint for linting (which you can invoke by running `bin/lint`).  I believe that you can configure gopls to use `goimports` as the formatting backend and then you should no longer see these as warnings.

adleong:
Hey @Hellcatlk, were you able to get your gopls set up as I described above?  That should make the warnings go away and I think we can close this PR.

PR #5144, opened 2020-10-27
charts: Do not store .component in linkerd-config

Pothulapati:
As `linkerd-config` is used with injections, It could be better to not
store the component value in `linkerd-config` which might cause any
unecessary configuration for application pods.

This is not the case currently, as there is no install specific logic in
patching anymore i.e that uses `.component`


adleong:
Even better than unsetting the component name when rendering the config would be to only set the component name on the copies used for rendering the proxy partials.  This way the component name never even gets set on the root Values in the first place.

adleong:
Looks like you introduced a change where port 443 is no longer skipped for outbound on control plane components.  Was that intentional?

adleong (approved):

charts/add-ons/grafana/templates/grafana.yaml
```diff
@@ -170,7 +170,7 @@ spec:
       - {{- include "partials.proxy" $tree | indent 8 | trimPrefix (repeat 7 " ") }}
       {{ if not .Values.global.cniEnabled -}}
       initContainers:
-      - {{- include "partials.proxy-init" . | indent 8 | trimPrefix (repeat 7 " ") }}
+      - {{- include "partials.proxy-init" $tree | indent 8 | trimPrefix (repeat 7 " ") }}
```
alpeb, line 173: `partials.proxy-init` doesn't use any of the values modified in `$tree` anymore, so you can continue passing `. `.

cli/cmd/testdata/install_controlplane_tracing_output.golden
```diff
@@ -852,7 +852,7 @@ data:
       prometheusUrl: ""
       proxy:
         capabilities: null
-        component: linkerd-controller
+        component: ""
```
alpeb, line 855: To avoid this it might be fine to remove `Component` from the `Proxy` struct in `values.go`, just as there isn't a `WorkloadKind` entry there.

Pothulapati: Yep, added this. This required an extra check in `_proxy.tpl` about the existance of `.component`

charts/add-ons/grafana/templates/grafana.yaml
Pothulapati: That makes sense! Done!

charts/partials/templates/_proxy.tpl
```diff
@@ -9,7 +18,7 @@ env:
 - name: LINKERD2_PROXY_LOG_FORMAT
   value: {{.Values.global.proxy.logFormat | quote}}
 - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
-  value: {{ternary "localhost.:8086" (printf "linkerd-dst-headless.%s.svc.%s:8086" .Values.global.namespace .Values.global.clusterDomain) (eq .Values.global.proxy.component "linkerd-destination")}}
+  value: {{ternary "localhost.:8086" (printf "linkerd-dst-headless.%s.svc.%s:8086" .Values.global.namespace .Values.global.clusterDomain) $isDestination }}
```
alpeb, line 21: One way of avoiding the new boilerplate at the beginning of this file while avoiding an error when using `eq` to compare against a potentially nil value is to do this:
```
value: {{ternary "localhost.:8086" (printf "linkerd-dst-headless.%s.svc.%s:8086" .Values.global.namespace .Values.global.clusterDomain) (eq (default .Values.global.proxy.component false) "linkerd-destination")}}
```

Pothulapati: Thanks for the suggestion! It seems to work with a slight chance that we discussed! :+1:

alpeb (approved):
Thanks @Pothulapati , LGTM :+1: :metal:

alpeb (approved):

PR #5145, opened 2020-10-27
update changelog for edge-20.10.6

hawkw:
This edge supersedes edge-20.10.5 as a release candidate for
stable-2.9.0. It adds a new `linkerd.io/inject: ingress` annotation to
support per-request routing when injecting ingress controllers.

* Added a new 'ingress mode' configuration to the proxy to support
  per-request routing for HTTP ingress controllers
* Added support for the `linkerd.io/inject: ingress` annotation to
  configure the proxy in ingress mode
* Reduced performance impact of logging in the proxy, especially when
  the `debug` or `trace` log levels are disabled
* Fixed spurious warnings logged by the `linkerd profile` CLI command

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,23 @@
 # Changes
 
+## edge-20.10.6
+
+This edge supersedes edge-20.10.5 as a release candidate for stable-2.9.0. It
+adds a new `linkerd.io/inject: ingress` annotation to support per-request
+routing when injecting ingress controllers.
+
+* Added a new 'ingress mode' configuration to the proxy to support per-request
+  routing for HTTP ingress controllers
+* Added support for the `linkerd.io/inject: ingress` annotation to configure the
+  proxy in ingress mode
+* Reduced performance impact of logging in the proxy, especially when the
+  `debug` or `trace` log levels are disabled
+* Fixed spurious warnings logged by the `linkerd profile` CLI command
+
 ## edge-20.10.5
 
 This edge supersedes edge-20.10.4 as a release candidate for stable-2.9.0. It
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,9 +1,23 @@
...
+## edge-20.10.6
+
+This edge supersedes edge-20.10.5 as a release candidate for stable-2.9.0. It
+adds a new `linkerd.io/inject: ingress` annotation to support per-request
+routing when injecting ingress controllers.
+
+* Added a new 'ingress mode' configuration to the proxy to support per-request
+  routing for HTTP ingress controllers
+* Added support for the `linkerd.io/inject: ingress` annotation to configure the
+  proxy in ingress mode
```
olix0r, line 12: We could probably drop the first point about `ingress mode` and only mention the annotation, or combine them... It's only one change from the user's point of view

olix0r (approved):

CHANGES.md
```diff
@@ -1,9 +1,23 @@
 # Changes
 
+## edge-20.10.6
+
+This edge supersedes edge-20.10.5 as a release candidate for stable-2.9.0. It
+adds a new `linkerd.io/inject: ingress` annotation to support per-request
```
adleong, line 6: Describing ingress mode as enabling "per-request routing" probably doesn't mean anything to anyone unfamiliar with the details of how the proxy works.  Perhaps we could instead describe this as a mode that allows ingresses to use service profiles and enables per-route metrics and traffic splits.

adleong (approved):

CHANGES.md
hawkw: good call, i just copied the language from the proxy's changelog

PR #5150, opened 2020-10-28
 cli: make check return SkipError when there is no prometheus configured

### LINKED ISSUE #5143: linkerd check --proxy fails when Pormetheus is disabled
### What is the issue?

When Prometheus is disabled, `linkerd check --proxy` blocks for 5 minutes and then fails.

### How can it be reproduced?

Install linkerd with Prometheus disabled:
```
$ cat config.yml
prometheus:
  enabled: false

$ linkerd install --config config.yml | kubectl apply -f -
...
$ linkerd check --proxy
...
linkerd-data-plane
------------------
√ data plane namespace exists
× data plane proxies are ready
    No prometheus instance to connect
    see https://linkerd.io/checks/#l5d-data-plane-ready for hints

Status check results are ×
```
### Environment

- Kubernetes Version: 1.18.2
- Cluster Environment: (GKE, AKS, kops, ...): KinD
- Host OS: Linux
- Linkerd version: edge-20.10.5

### Possible solution

Move the prometheus-specific tests into `healtcheck_addons.go`

Pothulapati:
Fixes #5143

Check currently keeps retrying for a bit when there is no prometheus
instnace configured i.e both linkerd and external prometheus instance
(through global.prometheusUrl).

The availability of prometheus is useful for some calls in public-api
that the check uses. This change updates the `ListPods` in `public-api` 
to still return the pods even when prometheus is not configured.

For a test that exclusively checks for prometheus metrics, we have a gate
which checks if a prometheus is configured and skips it othervise.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -2113,7 +2113,7 @@ func (hc *HealthChecker) getDataPlanePods(ctx context.Context) ([]*pb.Pod, error
 	resp, err := hc.apiClient.ListPods(ctx, req)
 	if err != nil {
 		// return a SkipError if there is no prometheus present
-		if strings.HasSuffix(err.Error(), "No prometheus instance to connect") {
+		if strings.Contains(err.Error(), public.ErrNoPrometheusInstance) {
 			return nil, &SkipError{Reason: err.Error()}
 		}
 		return nil, err
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -1241,6 +1241,16 @@ func (hc *HealthChecker) allCategories() []category {
 							return err
 						}
 
+						// Check if prometheus configured
+						prometheusValues := make(map[string]interface{})
+						err = yaml.Unmarshal(hc.linkerdConfig.Prometheus.Values(), &prometheusValues)
+						if err != nil {
+							return err
+						}
+						if !GetBool(prometheusValues, "enabled") && hc.linkerdConfig.Global.PrometheusURL == "" {
+							return &SkipError{Reason: "no prometheus instance to connect"}
+						}
+
 						return validateDataPlanePodReporting(pods)
 					},
 				},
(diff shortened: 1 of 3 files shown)


pkg/healthcheck/healthcheck.go
```diff
@@ -2113,7 +2113,7 @@ func (hc *HealthChecker) getDataPlanePods(ctx context.Context) ([]*pb.Pod, error
 	resp, err := hc.apiClient.ListPods(ctx, req)
 	if err != nil {
 		// return a SkipError if there is no prometheus present
-		if strings.HasSuffix(err.Error(), "No prometheus instance to connect") {
+		if strings.Contains(err.Error(), public.ErrNoPrometheusInstance) {
```
kleimkuhler, line 2116: After the the recent change with using `errors.Is` when possible, any reason we shouldn't do the same here?

If `ErrNoPrometheusInstance` was itself an error, we should be able to check for that here instead of using `strings.Contains`.

alpeb:
If I'm not mistaking, the `ListPods` API only uses prom to add the `Added`, `SinceLastReport` and `Uptime` properties to the returned pods. I think you still can return `ListPods` successfully without prom available, with those properties unset. In that case, the only check that needs to be skipped would be `data plane proxy metrics are present in Prometheus`.

For that particular test, what do you think about moving it into a new category `LinkerdDataPlanePrometheusAddOnChecks` in `heathcheck_addons.go`'s `addonCategories()` function? it would be added to the list of checks to make only if `options.dataPlaneOnly` is true.

adleong:
:+1: to @alpeb's suggestions

Pothulapati:
@alpeb That seems to be the right way to solving this! I updated code in `ListPods` to not fail when prometheus is avialable but instead return the podList that was gathered without any metrics.

For the prometheus check, I instead added a gate to make sure it happens only when a prometheus is available i.e either linkerd-prometheus or a external one!

alpeb (approved):
Ok, given prom can come from an external source and not just an add-on, it makes sense to not move that check into `healthcheck_addons.go` :+1: 

This looks good to me, modulo @kleimkuhler comment about error type safety :wink:

kleimkuhler (approved):
Awesome 👍

PR #5154, opened 2020-10-28
cli: add `--ingress` flag to inject cmd

Pothulapati:
This PR adds a new inject flag called `--ingress` which when enabled
adds a new annotation i.e `linkerd.io/inject: ingress`.

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -188,6 +188,10 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	}
 
 	if len(rt.overrideAnnotations) > 0 {
+		// Do not set ingress annotation if its manual
+		if val, ok := rt.overrideAnnotations[k8s.ProxyInjectAnnotation]; ok && val == k8s.ProxyInjectIngress && rt.injectProxy {
+			delete(rt.overrideAnnotations, k8s.ProxyInjectAnnotation)
+		}
 		conf.AppendPodAnnotations(rt.overrideAnnotations)
 	}
 
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -181,6 +181,9 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	}
 
 	if rt.injectProxy {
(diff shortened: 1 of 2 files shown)


adleong (approved):

kleimkuhler (approved):
Looks good; works well 👍

alpeb (commented):
Not a big deal, and a bit of a TIOLI, but using this flag with `--manual` results in the annotation being set.
I mean, `linkerd inject --manual` doesn't add an inject annotation, but `linkerd inject --manual --ingress` does, which is a little inconsistent. And that will result in a little different code path in the webhook injector: it won't inject not because of the absence of the annotation but because of the presence of the proxy.

adleong:
Really good catch, @alpeb!  I would expect `linkerd inject --manual --ingress` to set the environment variable in the proxy directly instead of setting the annotation.

Pothulapati:
> I would expect linkerd inject --manual --ingress to set the environment variable in the proxy directly instead of setting the annotation.

I think the problem is that currently it does both with manual. which makes proxy-injector react to the event but skips injection by taking a different path i.e by recognising that there is already a proxy present.

This can be fixed by having an check during injection which removes this annotation if its a manual injection and the annotaiton value is set to `ingress`.

cli/cmd/inject.go
```diff
@@ -188,6 +188,10 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	}
 
 	if len(rt.overrideAnnotations) > 0 {
+		// Do not set ingress annotation if its manual
+		if val, ok := rt.overrideAnnotations[k8s.ProxyInjectAnnotation]; ok && val == k8s.ProxyInjectIngress && rt.injectProxy {
+			delete(rt.overrideAnnotations, k8s.ProxyInjectAnnotation)
+		}
```
alpeb, line 194: You can move `delete(rt.overrideAnnotations, k8s.ProxyInjectAnnotation)` up under the existing `if injectProxy` block, and you don't need to check for the existence of the entry prior to deleting it :wink:

alpeb (approved):
LGTM! :shipit:

PR #5161, opened 2020-11-02
upgrade: set identity.issuer.crtExpiry correctly with legacy upgrades

Pothulapati:
With legacy upgrades, we can parse the cert and store the expiry
correctly instead of storing it as the default value which could be a
problem when we use that field. Currently, we do not use this field and
hence it did not cause any problems.

With Install on the latest edges, This field is correctly set and works
as expected. Thus, upgrades also hav the right value.


alpeb (approved):
Good find :+1:  LGTM!

kleimkuhler (approved):
👍

Pothulapati:
Merging this now!

Will create a separate PR for external issuers if needed.

PR #5164, opened 2020-11-02
Remove `--registry` flag from `linkerd mc install`

alpeb:
After the 2.9 multicluster refactoring, `linkerd mc install`'s only
workload installed is the nginx gateway, whose docker image is
configured through the flags `--gateway-nginx-image` and
`--gateway-nginx-image-version`. Thus there's no longer need of the
`--registry` flag, which is used OTOH by `linkerd mc link` which deploys the service mirror.


kleimkuhler (approved):

cpretzer (approved):
LGTM

PR #5166, opened 2020-11-02
upgrade: do not use external certs with upgrades

Pothulapati:
Currently, For legacy upgrades we are fetching even external certs and
using it for upgrades which contradicts the condition at
https://github.com/linkerd/linkerd2/blob/master/cli/cmd/options.go#L550
used with install and thus causing errors.

Instead we don't retrieve them with upgrades and hence they don't get
stored into the config and secrets which seems correct as we do not want
to store certs in the config and use them with upgrades when they are
created externally.

This touches only the upgrade path i.e `fetchIssuers` and would not
effect the retrievel of external certs for checks, etc.

### DIFF
--- a/cli/cmd/upgrade_legacy.go
+++ b/cli/cmd/upgrade_legacy.go
@@ -234,7 +234,9 @@ func fetchIssuer(ctx context.Context, k kubernetes.Interface, trustPEM string, s
 	)
 	switch scheme {
 	case string(corev1.SecretTypeTLS):
-		issuerData, err = issuercerts.FetchExternalIssuerData(ctx, k, controlPlaneNamespace)
+		// Do not return external issuer certs as no need of storing them in config and upgrade secrets
+		// Also contradicts condition in https://github.com/linkerd/linkerd2/blob/main/cli/cmd/options.go#L550
+		return &issuercerts.IssuerCertData{}, nil
 	default:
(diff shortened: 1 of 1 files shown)


Pothulapati:
One way to verify is to match what you get in `Values.identity` in `linkerd-config` with upgrade from legacy and a fresh install and they both seem to match.

Upgrade from `2.8.1`: _(This was failing before)_

```yaml
    identity:
      issuer:
        clockSkewAllowance: 20s
        crtExpiry: "0001-01-01T00:00:00Z"
        crtExpiryAnnotation: linkerd.io/identity-issuer-expiry
        issuanceLifetime: 24h0m0s
        scheme: kubernetes.io/tls
        tls:
          crtPEM: ""
```

Fresh install on latest edge:

```yaml
identity:
  issuer:
    clockSkewAllowance: 20s
    crtExpiry: "0001-01-01T00:00:00Z"
    crtExpiryAnnotation: linkerd.io/identity-issuer-expiry
    issuanceLifetime: 24h0m0s
    scheme: kubernetes.io/tls
    tls:
      crtPEM: ""
```


Both the above installs followed the [external issuer cert docs ](https://linkerd.io/2/tasks/automatically-rotating-control-plane-tls-credentials/)

kleimkuhler (approved):
👍

alpeb (approved):
Makes sense to me :+1: :ship: :shipit: Good :eye:

PR #5169, opened 2020-11-03
ha: Do not set a default proxy CPU limit

olix0r:
Now that the proxy can use more than one core, this behavior should be
enabled by default, even in HA mode.

This change modifies the default HA helm values to unset the cpu limit.


hawkw (approved):
no stop signs / no CPU limits!

Pothulapati (approved):
Makes sense! :shipit:

Issue #5165, opened 2020-11-02
Validate that cpu-limit actually restricts proxy CPU allocations


olix0r:
Anecdotally (in k3d) I've noticed that when setting a proxy cpu limit of "1", the proxy can still consume many cores worth of resources. The proxy detects its number of available cores via the [`num-cpus`](https://github.com/seanmonstar/num_cpus) crate, which should be cgroups-aware...

We should validate that CPU limits actually limit the proxy's thread allocations, as this could potentially have a deleterious effect on latency, in particular.

olix0r:
@grampelberg suggests that this is actually controlled by cpu _requests_ (and not necessarily limit) -- either way, we should test and validate the behavior as we write docs for it.

alpeb:
From my understanding of the [cpu management docs](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/) the number of cpus are not surfaced in cgroups unless the pod has a QoS of type "Guaranteed" and the kubelet management policy is `static` (the default being `none`).

```
$ k3d cluster create --k3s-server-arg '--kubelet-arg=kube-reserved=cpu=1' --k3s-server-arg '--kubelet-arg=cpu-manager-policy=static'
...
$ linkerd install | kubectl apply -f -
...
$  kubectl -n linkerd exec -ti linkerd-controller-xxxx -c linkerd-proxy sh
# inside the proxy container (my machine has 12 virtual CPUs)
$ cat /sys/fs/cgroup/cpuset/cpuset.cpus
0,2-5,7-11
$ exit

# now I edit the deployment so that all the containers end up with the same CPU/mem requests/limits, which turns the pod into QoS "Guaranteed"
$ kubectl -n linkerd edit deploy linkerd-controller
...
# I log again into the proxy container and cgroups exposes just one core:
$  kubectl -n linkerd exec -ti linkerd-controller-xxxx -c linkerd-proxy sh
# inside the proxy container (my machine has 12 virtual CPUs)
$ cat /sys/fs/cgroup/cpuset/cpuset.cpus
0
```

hawkw:
Hmm, looking at the docs @alpeb linked, that does appear to be the case. 

[...]

PR #5170, opened 2020-11-03
inject: Set LINKERD2_PROXY_CORES from the cpu limit

olix0r:
Per #5165, Kubernetes does not necessarily limit the proxy's access to
cores via `cgroups` when a CPU limit is set. As of #5168, the proxy now
supports a `LINKERD2_PROXY_CORES` environment configuration.

This change modifies the proxy injector to ensure that this environment
is configured from `Values.proxy.cores` Helm value and the
`config.linkerd.io/proxy-cpu-limit` annotation.


Pothulapati:
Hmmm, Why not use `values.Global.Proxy.Resources.CPU.Limit` for the same? should make this easy as if we use the same field to set the new env in the template, all the injection (from annotation) and install flow (i.e with `--proxy-cpu-limit)` would work right away. WDYT?

 With this route all of them have to be updated to take into account this new field.

alpeb:
@Pothulapati I guess that's because you wouldn't be able to implement directly in the template what `toWholeCPUCores` does...

Pothulapati:
@alpeb Ahh, That makes sense then! I'm also guessing it would be really hard to convert k8s.quantity into a whole number in Helm.

With this new variable, we should be overwriting this variable wherever `.values.Global.Proxy.Resources.CPU.Limit` is being to make inject and install overriding work.

alpeb:
> With this new variable, we should be overwriting this variable wherever `.values.Global.Proxy.Resources.CPU.Limit` is being to make inject and install overriding work.

Right, for instance `linkerd install --proxy-cpu-limit` is not affecting this new env var, not for control-plane pods nor for pods injected by the injector.

This also made me realize that for the helm install workflows, control-plane pods won't get this env var even if `global.Proxy.Resources.CPU.Limit` is explicitly set. So we end up with the same problem for CP proxies when the node has more cores than that Limit...

One not so straightforward way around that would be to expose in `values.yaml` separate `.proxy.core` entries for each CP component with anchors pointing to `global.Proxy.Resources.CPU.Limit` but then the proxy would have to deal with non-integer values. Actually if the proxy could deal with k8's Quantities directly, we could feed the env var the Limit directly like @Pothulapati suggested and all these issues would go away without additional `values.yaml` entries...

olix0r:
> With this new variable, we should be overwriting this variable wherever .values.Global.Proxy.Resources.CPU.Limit is being to make inject and install overriding work.

It's my preference that we not delay the edge release on this (and that we not make further changes between the edge and stable release, unless critical). I think it's sufficient to document this behavior, as it's only necessary when configuring via Helm and kubelet is not enforcing QoS.

The one thing we should confirm, though, is that the correct values are set when the `--proxy-cpu-limit` flag is set at install/upgrade-time.

cli/cmd/options.go
```diff
@@ -66,10 +66,11 @@ func makeInstallUpgradeFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.Fl
 					// individual flags can override the HA defaults.  This means that the HA flag must appear
 					// before the individual flags in the slice passed to flags.ApplyIfSet.
 					values.ControllerReplicas = haValues.ControllerReplicas
-					values.Global.Proxy.Cores = haValues.Global.Proxy.Cores
 					values.Global.Proxy.Resources.CPU.Request = haValues.Global.Proxy.Resources.CPU.Request
 					values.Global.Proxy.Resources.Memory.Request = haValues.Global.Proxy.Resources.Memory.Request
-					values.Global.Proxy.Resources.CPU.Limit = haValues.Global.Proxy.Resources.CPU.Limit
+					// NOTE: CPU Limits are not currently set by default HA charts.
+					//values.Global.Proxy.Cores = haValues.Global.Proxy.Cores
+					//values.Global.Proxy.Resources.CPU.Limit = haValues.Global.Proxy.Resources.CPU.Limit
```
Pothulapati, line 73: wondering what difference this makes as we are storing empty values of these variables even without this change? :thinking:

Pothulapati (approved):
Install and Inject work as expected i.e new env is also set along with the resources.

cli/cmd/options.go
olix0r: AIUI, this prevents us from overwriting values set by `--proxy-cpu-limit` when it is set before `--ha`

pkg/inject/inject_test.go
```diff
@@ -288,3 +291,27 @@ func TestGetOverriddenValues(t *testing.T) {
...
+
+func TestWholeCPUCores(t *testing.T) {
+	for _, c := range []struct {
+		v string
+		n int
+	}{
+		{v: "1", n: 1},
+		{v: "1m", n: 1},
+		{v: "1000m", n: 1},
+		{v: "1001m", n: 2},
```
Pothulapati, line 303: nit: why not round instead? so that this is `1` and not `2` so that we are not going out of the boundary :thinking:

olix0r: (1) quantity exposes a `RoundUp` and not `Round`
(2) i think if you say the proxy can use more than a core and we only allocate a single core, we'll never use the full limit, so rounding up better-reflects the user's intent, I think

alpeb (approved):
:shipit:

PR #5171, opened 2020-11-03
ha: Remove CPU limits for control plane components

olix0r:
As discussed in #5167 & #5169, Kubernetes CPU limits are not necessarily
discoverable from within the pod. This means that the control plane
processes may allocate far more threads than can actually be used by the
process given its process limits.

This change removes the default CPU limits for all control plane
components. CPU limits may still be set via Helm configuration.


alpeb (approved):
LGTM :+1:

Pothulapati (approved):
LGTM

PR #5173, opened 2020-11-03
nit: return crtExpiry even for External Certs

Pothulapati:
This change updates `FetchExternalIssuerData` to be more like
`FetchIssuerData` and return expiry correctly.

This field is currently not used anywhere and is just done for
consistentcy purposes.


olix0r (approved):
lgtm. i haven't tested it, but nothing particularly onerous

kleimkuhler (approved):
LGTM

PR #5174, opened 2020-11-03
Changes for `stable-2.9.0`

alpeb:
Only user-facing items were mentioned. There were previous edge release
notes contained a summary of a change, I preferred using that summary
instead of the more technical bullet point. Given the large list of
items, I separated into sections for easier digestion. Also, I didn't
repeat the TCP mTLS stuff (nor ARM support) below in the bullet points
as it was already well described in the summary.

## stable-2.9.0

This release extends Linkerd's zero-config mutual TLS (mTLS) support to all TCP
connections, allowing Linkerd to transparently encrypt and authenticate all TCP
connections in the cluster the moment it's installed. It also adds ARM support,
introduces a new multi-core proxy runtime for higher throughput, adds support
for Kubernetes service topologies, and lots, lots more, as described below:

* Proxy
  * Performed internal improvements for lower latencies under high concurrency
  * Reduced performance impact of logging , especially when the `debug` or
    `trace` log levels are disabled
  * Improved error handling for DNS errors encountered when discovering control
    plane addresses, which can be common during installation, before all
    components have been started, allowing linkerd to continue to operate
    normally in HA during node outages

* Control Plane
  * Added support for [topology-aware service
    routing](https://kubernetes.io/docs/concepts/services-networking/service-topology/)
    to the Destination controller; when providing service discovery updates to
    proxies the Destination controller will now filter endpoints based on the
    service's topology preferences
  * Added support for the new Kubernetes
    [EndpointSlice](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)
    resource to the Destination controller; Linkerd can be installed with
    `--enable-endpoint-slices` flag to use this resource rather than the
    Endpoints API in clusters where this new API is supported

* Dashboard
  * Added new Spanish translations (please help us translate into your
    language!)
  * Added new section for exposing multicluster gateway metrics

* CLI
  * Renamed the `--addon-config` flag to `--config` to clarify this flag can be
    used to set any Helm value
  * Added fish shell completions to the `linkerd` command

* Multicluster
  * Replaced the single `service-mirror` controller, with separate controllers
    that will be installed per target cluster through `linkerd multicluster
    link`
  * Changed the mechanism for mirroring services: instead of relying on
    annotations on the target services, now the source cluster should specify
    which services from the target cluster should be exported by using a label
    selector
  * Added support for creating multiple service accounts when installing
    multicluster with Helm to allow more granular revocation
  * Added a multicluster `unlink` command for removing multicluster links

* Prometheus
  * Moved Linkerd's bundled Prometheus into an add-on (enabled by default); this
    makes the Linkerd Prometheus more configurable, gives it a separate upgrade
    lifecycle from the rest of the control plane, and will allow users to
    disable the bundled Prometheus instance
  * The long-awaited Bring-Your-Own-Prometheus case has been finally addressed:
    added `global.prometheusUrl` to the Helm config to have linkerd use an
    external Prometheus instance instead of the one provided by default
  * Added an option to persist data to a volume instead of memory, so that
    historical metrics are available when prometheus is restarted
  * The helm chart can now configure persistent storage and limits

* Other
  * Added a new `linkerd.io/inject: ingress` annotation and accompanying
    `--ingress` flag to the `inject command, to configure the proxy to support
    service profiles and enable per-route metrics and traffic splits for HTTP
    ingress controllers
  * Changed the type of the injector and tap API secrets to `kubernetes.io/tls`
    so they can be provisioned by cert-manager
  * Changed default docker image repository to `ghcr.io` from `gcr.io`; **Users
    who pull the images into private repositories should take note of this
    change**
  * Introduced support for authenticated docker registries
  * Simplified the way that Linkerd stores its configuration; configuration is
    now stored as Helm values in the `linkerd-config` ConfigMap
  * Added support for Helm configuration of per-component proxy resources
    requests

This release includes changes from a massive list of contributors. A special
thank-you to everyone who helped make this release possible: --long
list, see file --

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,10 +10,10 @@ for Kubernetes service topologies, and lots, lots more, as described below:
 
 * Proxy
   * Performed internal improvements for lower latencies under high concurrency
-  * Reduced performance impact of logging , especially when the `debug` or
+  * Reduced performance impact of logging, especially when the `debug` or
     `trace` log levels are disabled
   * Improved error handling for DNS errors encountered when discovering control
-    plane addresses, which can be common during installation, before all
+    plane addresses; this can be common during installation before all
     components have been started, allowing linkerd to continue to operate
     normally in HA during node outages
 
@@ -40,7 +40,7 @@ for Kubernetes service topologies, and lots, lots more, as described below:
   * Added fish shell completions to the `linkerd` command
 
 * Multicluster
-  * Replaced the single `service-mirror` controller, with separate controllers
+  * Replaced the single `service-mirror` controller with separate controllers
     that will be installed per target cluster through `linkerd multicluster
     link`
   * Changed the mechanism for mirroring services: instead of relying on
@@ -54,18 +54,18 @@ for Kubernetes service topologies, and lots, lots more, as described below:
 * Prometheus
   * Moved Linkerd's bundled Prometheus into an add-on (enabled by default); this
     makes the Linkerd Prometheus more configurable, gives it a separate upgrade
-    lifecycle from the rest of the control plane, and will allow users to
+    lifecycle from the rest of the control plane, and allows users to
     disable the bundled Prometheus instance
   * The long-awaited Bring-Your-Own-Prometheus case has been finally addressed:
     added `global.prometheusUrl` to the Helm config to have linkerd use an
     external Prometheus instance instead of the one provided by default
   * Added an option to persist data to a volume instead of memory, so that
-    historical metrics are available when prometheus is restarted
+    historical metrics are available when Prometheus is restarted
   * The helm chart can now configure persistent storage and limits
 
 * Other
   * Added a new `linkerd.io/inject: ingress` annotation and accompanying
-    `--ingress` flag to the `inject command, to configure the proxy to support
+    `--ingress` flag to the `inject` command, to configure the proxy to support
     service profiles and enable per-route metrics and traffic splits for HTTP
     ingress controllers
   * Changed the type of the injector and tap API secrets to `kubernetes.io/tls`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,119 @@
 # Changes
 
+## stable-2.9.0
+
+This release extends Linkerd's zero-config mutual TLS (mTLS) support to all TCP
+connections, allowing Linkerd to transparently encrypt and authenticate all TCP
+connections in the cluster the moment it's installed. It also adds ARM support,
+introduces a new multi-core proxy runtime for higher throughput, adds support
+for Kubernetes service topologies, and lots, lots more, as described below:
+
+* Proxy
+  * Performed internal improvements for lower latencies under high concurrency
+  * Reduced performance impact of logging , especially when the `debug` or
+    `trace` log levels are disabled
+  * Improved error handling for DNS errors encountered when discovering control
+    plane addresses, which can be common during installation, before all
+    components have been started, allowing linkerd to continue to operate
+    normally in HA during node outages
+
+* Control Plane
+  * Added support for [topology-aware service
+    routing](https://kubernetes.io/docs/concepts/services-networking/service-topology/)
+    to the Destination controller; when providing service discovery updates to
+    proxies the Destination controller will now filter endpoints based on the
+    service's topology preferences
+  * Added support for the new Kubernetes
+    [EndpointSlice](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)
+    resource to the Destination controller; Linkerd can be installed with
+    `--enable-endpoint-slices` flag to use this resource rather than the
+    Endpoints API in clusters where this new API is supported
+
+* Dashboard
+  * Added new Spanish translations (please help us translate into your
+    language!)
+  * Added new section for exposing multicluster gateway metrics
+
+* CLI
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+    `--enable-endpoint-slices` flag to use this resource rather than the
+    Endpoints API in clusters where this new API is supported
+
+* Dashboard
+  * Added new Spanish translations (please help us translate into your
+    language!)
+  * Added new section for exposing multicluster gateway metrics
+
+* CLI
+  * Renamed the `--addon-config` flag to `--config` to clarify this flag can be
```
adleong, line 38: I think this point is missing some words

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+    used
+  * Added fish shell completions to the `linkerd` command to set any Helm value
```
adleong, line 40: Oh, I found the missing words!

adleong (commented):
These notes are great!

CHANGES.md
alpeb: :rofl: dang vim

adleong (approved):

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+
+This release extends Linkerd's zero-config mutual TLS (mTLS) support to all TCP
+connections, allowing Linkerd to transparently encrypt and authenticate all TCP
+connections in the cluster the moment it's installed. It also adds ARM support,
+introduces a new multi-core proxy runtime for higher throughput, adds support
+for Kubernetes service topologies, and lots, lots more, as described below:
+
+* Proxy
+  * Performed internal improvements for lower latencies under high concurrency
+  * Reduced performance impact of logging , especially when the `debug` or
```
kleimkuhler, line 13:
```suggestion
  * Reduced performance impact of logging, especially when the `debug` or
```

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+    `trace` log levels are disabled
+  * Improved error handling for DNS errors encountered when discovering control
+    plane addresses, which can be common during installation, before all
+    components have been started, allowing linkerd to continue to operate
+    normally in HA during node outages
```
kleimkuhler, line 18: Nit: I feel like this reads a little odd with all the commas, maybe something like:

```suggestion
  * Improved error handling for DNS errors encountered when discovering control
    plane addresses; this can be common during installation before all
    components have been started, allowing linkerd to continue to operate
    normally in HA during node outages
```

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+    language!)
+  * Added new section for exposing multicluster gateway metrics
+
+* CLI
+  * Renamed the `--addon-config` flag to `--config` to clarify this flag can be
+    used to set any Helm value
+  * Added fish shell completions to the `linkerd` command
+
+* Multicluster
+  * Replaced the single `service-mirror` controller, with separate controllers
```
kleimkuhler, line 43: Nit:

```suggestion
  * Replaced the single `service-mirror` controller with separate controllers
```

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+    which services from the target cluster should be exported by using a label
+    selector
+  * Added support for creating multiple service accounts when installing
+    multicluster with Helm to allow more granular revocation
+  * Added a multicluster `unlink` command for removing multicluster links
+
+* Prometheus
+  * Moved Linkerd's bundled Prometheus into an add-on (enabled by default); this
+    makes the Linkerd Prometheus more configurable, gives it a separate upgrade
+    lifecycle from the rest of the control plane, and will allow users to
```
kleimkuhler, line 57: Change the tense here to match the other points

```suggestion
    lifecycle from the rest of the control plane, and allows users to
```

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+    disable the bundled Prometheus instance
+  * The long-awaited Bring-Your-Own-Prometheus case has been finally addressed:
+    added `global.prometheusUrl` to the Helm config to have linkerd use an
+    external Prometheus instance instead of the one provided by default
+  * Added an option to persist data to a volume instead of memory, so that
+    historical metrics are available when prometheus is restarted
```
kleimkuhler, line 63:
```suggestion
    historical metrics are available when Prometheus is restarted
```

CHANGES.md
```diff
@@ -1,5 +1,119 @@
...
+  * The helm chart can now configure persistent storage and limits
+
+* Other
+  * Added a new `linkerd.io/inject: ingress` annotation and accompanying
+    `--ingress` flag to the `inject command, to configure the proxy to support
```
kleimkuhler, line 68:
```suggestion
    `--ingress` flag to the `inject` command, to configure the proxy to support
```

kleimkuhler (commented):
Notes look great! I suggested a few non-blocking changes on some bullets, but there are few that should be addressed as well.

CHANGES.md
alpeb: I'm also a big semicolon fan :slightly_smiling_face:

CHANGES.md
alpeb: Ah that's what was messing up my syntax coloring!

kleimkuhler (approved):
Awesome!

PR #5185, opened 2020-11-08
Add MC Gateway Service Annotations

### LINKED ISSUE #5182: Setting multi cluster gateway as internal load balancer
I want to be able to reach my gateway load balancer via internal VPC IP.
Add option in the multicluster chart to add an annotation to the load balancer service:
For example in AWS we will add this
`service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0`

### LINKED ISSUE #5192: cli: Reorganize diagnostic subcommand(s)
There are a few purely diagnostic CLI subcommands that are generally useful, especially for Linkerd developers, but don't achieve any high-level user-facing tasks...

- `diagnostics`  Fetch metrics directly from the Linkerd control plane containers
- `endpoints`    Introspect Linkerd's service discovery state
- `metrics`      Fetch metrics directly from Linkerd proxies

I propose that we move all of these under `linkerd diagnostics` as follows:

- `linkerd diagnostics endpoints svc/foo [port]`

The `endpoints` command should become `linkerd diagnostics endpoints`. The authority-form of resolutions is somewhat of an implementation detail and, instead of a DNS-form name, the CLI should accept a service or pod reference and a port. Service lookups should be traffic-split aware, resolving via service profiles to obtain traffic splits before resolving individual leaf services.

- `linkerd diagnostics proxy-metrics -n NS po --selector ...` 

The `linkerd metrics` command should be renamed `linkerd diagnostics proxy-metrics`. It should accept the same CLI arguments as `kubectl get` (which it does not today); and it should dump the metrics from all selected resources.

- `linkerd diagnostics controller-metrics [-n NS] [workload]`

The `linkerd diagnostics` command, as it exists today, doesn't really fit into a world where controller is split across multiple namespaces. I.e., we already have a `linkerd-multicluster` namespace that isn't supported by the command. The command should be moved to `linkerd diagnostics controller-metrics`. By default it should dump the mertrics from the core control plane components, but it should also accept an optional workload reference (pod, deployment, label selector, etc) so that the command may be used with any linkerd controllers. I think we should assume/require that there is an `admin-http` container port for all control plane components, and this should be used to determine which port is read. (Alternately, there's no reason we need these ports to vary at all -- it seems easier/better to simply standardize on using the same port across all linkerd controllers, but this isn't really a requirement, i guess).

shaikatz:
Fixes #5182

That will allow setting the load balancer as internal.

### DIFF
--- a/charts/linkerd2-multicluster/README.md
+++ b/charts/linkerd2-multicluster/README.md
@@ -13,7 +13,7 @@ linkerd2-multicluster chart and their default values.
 | Parameter                       | Description                                                                                 | Default                                      |
 |---------------------------------|---------------------------------------------------------------------------------------------|----------------------------------------------|
 |`controllerComponentLabel`       | Control plane label. Do not edit                                                            |`linkerd.io/control-plane-component`          |
-|`controllerImage`                | Docker image for the Service mirror component (uses the Linkerd controller image)           |`ghcr.io/linkerd/controller`                |
+|`controllerImage`                | Docker image for the Service mirror component (uses the Linkerd controller image)           |`ghcr.io/linkerd/controller`                  |
 |`controllerImageVersion`         | Tag for the Service Mirror container Docker image                                           |`latest version`                              |
 |`createdByAnnotation`            | Annotation label for the proxy create. Do not edit.                                         |`linkerd.io/created-by`                       |
 |`gateway`                        | If the gateway component should be installed                                                |`true`                                        |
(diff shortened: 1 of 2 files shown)


Pothulapati (approved):
Works as expected. Thanks for taking up.

@alpeb I think this will be helm only right? as it would be hard to allow that through the CLi? :thinking:

alpeb (approved):
This looks good to me :+1: 
Indeed, this would be for Helm only, which is fine, as advanced use cases like this one are supposed to be configured through Helm anyways.

adleong:
This looks good.  My only question is that if the `gatewayServiceAnnotations` field is written to the `linkerd-config` configmap, will it cause an error when this yaml is unmarshalled into a Values struct which does not have this field?

adleong (approved):
Ah, nevermind.  This is the multicluster chart, not the core Linkerd chart.  Carry on!

Issue #5193, opened 2020-11-09
cli: Remove the `install-sp` subcommand


olix0r:
The `install-sp` subcommand generates a set of service profiles for the linkerd control plane. I propose that, instead of exposing this as a core linkerd CLI subcommand, we publish helm/kustomize configuration so that users can install these CRDs if they choose to.

My main argument is that this command is _confusing_.
```
:; linkerd help install-sp
Output Kubernetes configs to install Linkerd Service Profiles.

This command installs Service Profiles into the Linkerd control plane. A
cluster-wide Linkerd control-plane is a prerequisite. To confirm Service Profile
support, verify "kubectl api-versions" outputs "linkerd.io/v1alpha2".
```

It _sounds_ like this command adds ServiceProfile functionality to Linkerd, when in reality it adds the service profiles for the linkerd controller components.

Furthermore, these service profiles are of marginal value as proxies do not apply service profiles to proxy-to-controller communication. If users feel strongly that this subcommand be preserved, I'd rather it be moved to something like `linkerd diagnostics controller-service-profiles` (see #5192). But my preference would be to publish them completely outside of the Linkerd CLI.

adleong:
:+1:

olix0r:
The verdict, per discussion, is that we should move this under `diagnostics` (see #5192)

Pothulapati:
Fixes by #5205

PR #5205, opened 2020-11-10
cli: reorganise diagnostics subcommand

### LINKED ISSUE #5192: cli: Reorganize diagnostic subcommand(s)

Pothulapati:
Fixes #5192, #5193

This PR moves `metrics`, `diagnostics`(which prints out metrics of
control-plane components), `endpoints` and `install-sp` into a new `diagnostics`
subcommand.


cli/cmd/diagnostics.go
```diff
@@ -14,29 +14,58 @@ const (
...
-func newDiagnosticsOptions() *diagnosticsOptions {
-	return &diagnosticsOptions{
+func newControlPlaneMetricsOptions() *ControlPlaneMetricsOptions {
+	return &ControlPlaneMetricsOptions{
 		wait: 30 * time.Second,
 	}
 }
 
-// newCmdDashboard creates a new cobra command `diagnostics` which contains commands to fetch control plane container's metrics
+// newCmdDiagnostics creates a new cobra command `diagnostics` which contains commands to diagnose Linkerd
```
kleimkuhler, line 33:
```suggestion
// newCmdDiagnostics creates a new cobra command `diagnostics` which contains commands to fetch Linkerd diagnostics
```

cli/cmd/diagnostics.go
```diff
@@ -14,29 +14,58 @@ const (
...
+}
+
+// newCmdControlPlaneMetrics creates a new cobra command `controlplane-metrics` which contains commands to fetch control plane container's metrics
+func newCmdControlPlaneMetrics() *cobra.Command {
+	options := newControlPlaneMetricsOptions()
 
 	cmd := &cobra.Command{
-		Use:   "diagnostics",
-		Short: "Fetch metrics directly from the Linkerd control plane containers",
+		Use:     "controlplane-metrics",
```
kleimkuhler, line 66: We tend to use a space in "control plane" in comments, as well as the fact that we use the capitalization of "ControlPlane.." everywhere. Maybe this subcommand should be `control-plane-metrics`? It's a little more verbose which isn't great.

cli/cmd/diagnostics.go
```diff
@@ -14,29 +14,58 @@ const (
...
+	}
+
+	diagnosticsCmd.AddCommand(newCmdControlPlaneMetrics())
+	diagnosticsCmd.AddCommand(newCmdEndpoints())
+	diagnosticsCmd.AddCommand(newCmdMetrics())
+	return diagnosticsCmd
+}
+
+// newCmdControlPlaneMetrics creates a new cobra command `controlplane-metrics` which contains commands to fetch control plane container's metrics
+func newCmdControlPlaneMetrics() *cobra.Command {
```
kleimkuhler, line 62: What are your thoughts on maybe moving this to a `metrics.go`, or separating into a `control-plane-metrics.go` and `proxy-metrics.go`?

It feels a little odd to have the root `diagnotics` command here with one subcommand, but the other two subcommands defined in separate files.

cli/cmd/diagnostics.go
Pothulapati: From the issue, It is mentioned to be `controller-metrics` which seems concise. Will be doing that.

cli/cmd/diagnostics.go
Pothulapati: Yeah, That makes sense. Should we instead move all sub-commands of `diagnostics` into `diagnostics.go`,  like `multicluster.go` ? :thinking:

cli/cmd/diagnostics.go
kleimkuhler: Cool yep that looks good!

cli/cmd/diagnostics.go
kleimkuhler: Alex opened #5210 which basically is what we're discussing here. Maybe for now we can move all subcommands into `diagnostics.go` and address this as part of that?

adleong: :+1: to @kleimkuhler's suggestion

cli/cmd/diagnostics.go
```diff
@@ -1,78 +1,42 @@
...
+  # Install service profiles for the control-plane components.
+  linkerd diagnostics install-sp
+  `,
 	}
 
-	cmd.Flags().DurationVarP(&options.wait, "wait", "w", options.wait, "Time allowed to fetch diagnostics")
+	diagnosticsCmd.AddCommand(newCmdControllerMetrics())
+	diagnosticsCmd.AddCommand(newCmdEndpoints())
+	diagnosticsCmd.AddCommand(newCmdMetrics())
+	diagnosticsCmd.AddCommand(newCmdInstallSP())
```
Pothulapati, line 39: Should we call the `install-sp` something else? As its a bit confusing as this installs service profiles for the control-plane components and not the service profiles crds itself

cli/cmd/diagnostics.go
```diff
@@ -1,78 +1,42 @@
...
+	diagnosticsCmd := &cobra.Command{
+		Use:     "diagnostics [flags]",
+		Aliases: []string{"dg"},
+		Args:    cobra.NoArgs,
+		Short:   "Commands used to diagnose Linkerd components",
+		Long: `Commands used to diagnose Linkerd components.
+
+This command provides subcommands to diagnose the functionality of Linkerd.`,
+		Example: `  # Get control-plane component metrics
+  linkerd diagnostics cp-metrics
```
adleong, line 23: cp-metrics doesn't seem like it's a valid abbreviation for this command.

adleong (approved):

cli/cmd/diagnostics.go
kleimkuhler, line 23: This can be updated now:

```suggestion
  linkerd diagnostics controller-metrics
```

cli/cmd/metrics.go
```diff
@@ -62,13 +62,13 @@ func newCmdMetrics() *cobra.Command {
...
-  linkerd metrics po/pod-foo-bar
+  linkerd diagnostics metrics po/pod-foo-bar
 
   # Get metrics from the web deployment in the emojivoto namespace.
-  linkerd metrics -n emojivoto deploy/web
+  linkerd diagnostics metrics -n emojivoto deploy/web
 
   # Get metrics from the linkerd-controller pod in the linkerd namespace.
-  linkerd metrics -n linkerd $(
+  linkerd diagnostics metrics -n linkerd $(
```
kleimkuhler, line 71: These should be `linkerd diagnostics proxy-metrics`.

kleimkuhler (approved):
This looks good. There are some doc updates you should change before merging, but changes make sense and work as expected.

PR #5221, opened 2020-11-12
Update proxy-init to v1.3.7

alpeb:
This upgrades both the proxy-init image itself, and the go dependency on
proxy-init as a library, which fixes CNI in k3s and any host using
binaries coming from BusyBox, where `nsenter` has an
issue parsing arguments (see rancher/k3s#1434).


Pothulapati (approved):
LGTM! :+1:

kleimkuhler (approved):

PR #5223, opened 2020-11-12
Remove logs comparisons in integration tests

alpeb:
The rare cases these were useful don't make up for the burden of
maintaing them, having different k8s version change the messages and
having unexpected warnings come up that didn't affect the final
convergence of the system.

With this we also revert the indirection added back in #4538 that
fetched unmatched warnings after a test had failed.


olix0r (approved):
Note on the commit message:

> The rare cases these were useful don't make up for the burden of

should read

> The rare cases _where these tests were_ useful...

kleimkuhler (approved):
Looks good 👍

Issue #5209, opened 2020-11-10
Linkerd stuck talking to stale address after headless service rollout


adleong:
### What is the issue?

When a meshed pod is sending requests to a headless service and the contents of that service change (e.g. from rolling the deployment), the Linkerd proxy in the client pod will be stuck attempting to connect to a stale endpoint.

This happens because when the client resolves the headless service, it will attempt to connect to an endpoint of that service (rather than a service clusterip).  The Linkerd proxy will forward the connection to that endpoint, but when the target deployment is restarted and that endpoint goes away, the Linkerd proxy does not break the connection from the client container.

### How can it be reproduced?

* Have a meshed pod send traffic to a headless service
* Restart the deployment which backs the service

### Logs, error output, etc

The client's Linkerd proxy logs:

```
[346867.852718s]  WARN ThreadId(01) outbound:accept{peer.addr=XXX.XXX.XXX.XXX:XXXX target.addr=XXX.XXX.XXX.XXX:XXXX}: linkerd2_app_core::errors: Failed to proxy request: error trying to connect: Connection refused (os error 111)
```

where the target.addr is a stale endpoint which no longer exists.

### Environment

- Linkerd version: edge-20.11.1

olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/736 and https://github.com/linkerd/linkerd2-proxy-api/pull/51 sketch out a proxy implementation that supports an `endpoint` response on profile resolutions, intended to be set when the resolved IP is known to be a k8s pod. When this field is set, further discovery is bypassed (including the balancer and traffic split router). When the resolved endpoint is neither a service nor a known pod, we should skip setting a `dst_override` so that these stacks may be bypassed.

When forwarding to a singular endpoint and not a service, we handle `io::Errors` on requests specially so that the entire service becomes defunct.

We should try to implement this resolution scheme in the destination controller to validate that this resolves this issue.

olix0r:
Fixed by #5227 & #5253

PR #5227, opened 2020-11-13
Add endpoint to GetProfile response

kleimkuhler:
Context: #5209

This updates the destination service to set the `Endpoint` field in `GetProfile`
responses.

The `Endpoint` field is only set if the IP maps to a Pod--not a Service.

Additionally in this scenario, the default Service Profile is used as the base
profile so no other significant fields are set.

### Examples

```
# GetProfile for an IP that maps to a Service
❯ go run controller/script/destination-client/main.go -method getProfile -path 10.43.222.0:9090
INFO[0000] fully_qualified_name:"linkerd-prometheus.linkerd.svc.cluster.local"  retry_budget:{retry_ratio:0.2  min_retries_per_second:10  ttl:{seconds:10}}  dst_overrides:{authority:"linkerd-prometheus.linkerd.svc.cluster.local.:9090"  weight:10000}
```

Before:

```
# GetProfile for an IP that maps to a Pod
❯ go run controller/script/destination-client/main.go -method getProfile -path 10.42.0.20
INFO[0000] retry_budget:{retry_ratio:0.2 min_retries_per_second:10 ttl:{seconds:10}}
```

After:

```
# GetProfile for an IP that maps to a Pod
❯ go run controller/script/destination-client/main.go -method getProfile -path 10.42.0.20
INFO[0000] retry_budget:{retry_ratio:0.2  min_retries_per_second:10  ttl:{seconds:10}}  endpoint:{addr:{ip:{ipv4:170524692}}  weight:10000  metric_labels:{key:"control_plane_ns"  value:"linkerd"}  metric_labels:{key:"deployment"  value:"fast-1"}  metric_labels:{key:"pod"  value:"fast-1-5cc87f64bc-9hx7h"}  metric_labels:{key:"pod_template_hash"  value:"5cc87f64bc"}  metric_labels:{key:"serviceaccount"  value:"default"}  tls_identity:{dns_like_identity:{name:"default.default.serviceaccount.identity.linkerd.cluster.local"}}  protocol_hint:{h2:{}}}
```


controller/api/destination/watcher/ip_watcher.go
```diff
@@ -5,10 +5,10 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/gogo/status"
```
adleong, line 8: Why the switch from grpc to gogo?

adleong (commented):
This looks good.

One thing I noticed here is that we'll fail to return an authorityOverride if you directly target an endpoint that has an authorityOverride but is not associated with a pod (e.g. in the case of a mirror service).  However, I don't think that ever worked to begin with since we only ever look up endpoints by their service name.  So this isn't a regression, but maybe something to keep in mind.

I also noticed an oddity of the API is that returning no FullyQualifiedName and no Endpoint (which we do for totally unknown IPs) effectively is the same as returning no FullyQualifiedName and an Endpoint with no Pod metadata on it (similar to what we do for destination.Get lookups to an unknown ip).

Having two ways to represent the same thing in the API is confusing.  Would it make sense to say that GetProfiles must always return either FullyQualifiedName or Endpoint, but the pod metadata on Endpoint is optional?

controller/api/destination/watcher/ip_watcher.go
olix0r: +1 it would be best to avoid adding a new dependency if we can avoid it. (I believe i've seen vscode just magically do this sort of thing in the past?)

olix0r:
Marking this as a draft until the api branch is merged through

controller/api/destination/watcher/ip_watcher.go
kleimkuhler: Ah sorry yea this snuck in from a VS Code auto import or something

olix0r:
> Having two ways to represent the same thing in the API is confusing. Would it make sense to say that GetProfiles must always return either FullyQualifiedName or Endpoint, but the pod metadata on Endpoint is optional?

There's a semantic difference between "i know about this target" and "i don't know anything about this target". I think it's preferable that, if the control plane really doesn't know about a target, that it indicate that by returning a default (empty) profile response. If the target refers to an endpoint that is known to the control plane, it returns an endpoint value.

controller/api/destination/watcher/ip_watcher.go
```diff
@@ -389,7 +415,7 @@ func (ss *serviceSubscriptions) unsubscribe(port Port, listener EndpointUpdateLi
 	delete(ss.listeners, listener)
 }
 
-func withPort(pods AddressSet, port Port) AddressSet {
+func WithPort(pods AddressSet, port Port) AddressSet {
```
olix0r, line 418: Note that the linter wants this to have a docstring...

Although, I'd suggest moving this to be a method on `AddressSet` so you can do something like `w.PodToAddressSet(pod).WithPort(p)`

kleimkuhler: Good call!

adleong (approved):

olix0r:
Is this behavior exercised by tests anywhere? I.e are there integration tests that use headless services or otherwise communicate via pods?  I suppose any prometheus-based metrics test relies on this behavior, but (as a followup), it might be good to instrument tests more narrowly scoped to the destination service.

kleimkuhler:
No this is it not tested anywhere right now. I looked into adding this to the destination server tests but it looks like the k8s API mock we use doesn't support IP indexers so it's not as easy as adding a test case to for the `GetProfile` call.

I'll work on adding something to the integration tests.

olix0r (approved):
lgtm

PR #5237, opened 2020-11-16
cli: don't check for SAN in root and intermediate Certs

### LINKED ISSUE #5228: Remove the reuirement of SAN for root and intermediate certs
So, The step folks [replied](https://github.com/smallstep/cli/issues/386#issuecomment-726325415) that it is not advised to use `--san`  with root and intermediate certs. But our CLI uses the `x509.verify` fn which is used to validate leaf certs not root and intermediate and hence has a strict requirement on verifying hostnames through SAN only. The code path starts [here](https://sourcegraph.com/github.com/linkerd/linkerd2@cc0edffce8ca014159665367e9fb54bb85c5517b/-/blob/pkg/tls/cred.go#L93)

We should,
- Update the cert generation docs to not have --san  with root and issuer cert generation.
- [Update the checks, etc where we validate intermediate Cert](https://sourcegraph.com/github.com/linkerd/linkerd2@cc0edff/-/blob/pkg/healthcheck/healthcheck.go#L1025) to not check for Hostname of `issuerIdentity()`  as its not needed there. (we currently do this)

_The SAN work we did with webhooks is still useful as they are used as leaf certs._

Pothulapati:
Fixes #5228

As discussed in the issue, It is not correct for root and intermediate
certs to have SAN. This PR updates the check to not verify the
intermediate issuer cert with the identity dns name (which checks with
SAN and not CN as the the `verify` func is used to verify leaf certs and
not root and intermediate certs). This PR also removes the extra san
field added in cert generation in CLI which was added previously.

### DIFF
--- a/pkg/issuercerts/issuercerts.go
+++ b/pkg/issuercerts/issuercerts.go
@@ -155,7 +155,7 @@ func CheckCertAlgoRequirements(cert *x509.Certificate) error {
 }
 
 // VerifyAndBuildCreds builds and validates the creds out of the data in IssuerCertData
-func (ic *IssuerCertData) VerifyAndBuildCreds(dnsName string) (*tls.Cred, error) {
+func (ic *IssuerCertData) VerifyAndBuildCreds() (*tls.Cred, error) {
 	creds, err := tls.ValidateAndCreateCreds(ic.IssuerCrt, ic.IssuerKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read CA: %s", err)
@@ -180,7 +180,7 @@ func (ic *IssuerCertData) VerifyAndBuildCreds(dnsName string) (*tls.Cred, error)
 		return nil, err
 	}
 
-	if err := creds.Verify(anchors, dnsName, time.Time{}); err != nil {
+	if err := creds.Verify(anchors, "", time.Time{}); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/issuercerts/issuercerts.go
```diff
@@ -180,7 +180,7 @@ func (ic *IssuerCertData) VerifyAndBuildCreds(dnsName string) (*tls.Cred, error)
 		return nil, err
 	}
 
-	if err := creds.Verify(anchors, dnsName, time.Time{}); err != nil {
+	if err := creds.Verify(anchors, "", time.Time{}); err != nil {
```
Pothulapati, line 183: Thought we could get rid of the `dnsName` argument totally here but the same func is used at other places to verify the proxy-injector's, etc leaf certificates where we do need to verify the dns name.

olix0r (approved):
Haven't tested this, but I'm generally +1 to the change.

adleong (approved):

pkg/tls/ca.go
```diff
@@ -118,7 +118,6 @@ func CreateRootCA(
 	// Configure the root certificate.
 	t := createTemplate(1, &key.PublicKey, validity)
 	t.Subject = pkix.Name{CommonName: name}
```
alpeb, line 120: Would it make sense to put here something generic like "Linkerd Root CA" instead of something specific?
Also that'd mean we could have the call `tls.GenerateRootCAWithDefaults()` in `options.go` not having to pass any name...

alpeb: After further offline discussion we decided to keep things as is :+1:

alpeb (approved):

PR #5240, opened 2020-11-17
tracing: new jaeger independent helm chart

### LINKED ISSUE #5230: Move tracing chart into jaeger package
In order to modularize the jaeger related components into a linkerd-jaeger extension, we should create a new top-level directory in the Linkerd2 repo called `jaeger` and move the tracing Helm chart there.

Behavior in the Linkerd control plane and CLI which reference the tracing chart or components need to be updated as appropriate.  For example:

* The control plane flag --control-plane-tracing should be replace with a --control-plane-opencensus-collector flag which specifies the address of an opencensus collector that the control plane should send spans to
* A JaegerURL Value should be added for configuring the link to the jaeger dashboard in the Linkerd web dashboard

Pothulapati:
Fixes #5230

This PR moves tracing into a jaeger chart with no proxy injection
templates. We still keep the dependency on partials, as we could use
common templates like resources, etc from there.


charts/jaeger/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for the jaeger add-on in Linkerd
+name: jaeger
+version: 0.1.0
+maintainers:
+  - name: Linkerd authors
+    email: cncf-linkerd-dev@lists.cncf.io
+    url: https://linkerd.io/
```
adleong, line 9: unless the Helm convention is otherwise, we should always end files with a newline

charts/jaeger/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
adleong, line 1: Can we put this chart inside the top level `jaeger` directory?  e.g. `jaeger/charts/jaeger/Chart.yaml`?

charts/jaeger/templates/tracing-rbac.yaml
```diff
@@ -0,0 +1,26 @@
+---
+###
+### linkerd-collector RBAC
+###
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-collector
```
adleong, line 9: Since this will live in the `linkerd-jaeger` namespace, I think we can simply call this `collector`

charts/jaeger/templates/tracing-rbac.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
adleong, line 1: can probably simply call this template `rbac.yaml`.

charts/jaeger/templates/tracing-rbac.yaml
```diff
@@ -0,0 +1,26 @@
...
+    namespace: {{.Values.namespace}}
+---
+###
+### linkerd-jaeger RBAC
+###
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: linkerd-jaeger
```
adleong, line 22: similarly, I think we can simply call this `jaeger`.

charts/jaeger/values.yaml
```diff
@@ -0,0 +1,15 @@
+# Default values for tracing.
+
+namespace: linkerd-jaeger
+
+collector:
+  image: 
+    name: omnition/opencensus-collector
+    version: 0.1.11
+    pullPolicy: Always
```
adleong, line 9: do we actually want the default pullPolicy to be Always for these published images?

jaeger/cmd/install.go
```diff
@@ -0,0 +1,128 @@
...
+func newCmdInstall() *cobra.Command {
+	var skipChecks bool
+
+	values, err := jaeger.NewValues()
+	if err != nil {
+		fmt.Fprint(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+
+	flags, installOnlyFlagSet := makeInstallFlags(values)
```
adleong, line 35: I think we can ditch all of this fancy flag machinery in favor of just plain cobra flags.

jaeger/cmd/install.go
```diff
@@ -0,0 +1,128 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "install [flags]",
+		Args:  cobra.NoArgs,
+		Short: "Output Kubernetes resources to install jaeger extension",
+		Long:  `Output Kubernetes resources to install jaeger extension.`,
+		Example: `  # Default install.
+  linkerd jaeger install | kubectl apply -f -
+
+  # Install Linkerd into a non-default namespace.
```
adleong, line 45: comment says "install linkerd"

jaeger/cmd/install.go
```diff
@@ -0,0 +1,128 @@
...
+  linkerd jaeger install --namespace linkerdtest | kubectl apply -f -`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !skipChecks {
+				// TODO: Add Checks for checking if linkerd exists
+				// Also check for jaeger
```
adleong, line 50: I don't think we want to error if jaeger already exists.  We want to simply overwrite the previous installation, since this is the flow for updates.

jaeger/cmd/root.go
```diff
@@ -0,0 +1,97 @@
...
+	Long:  `linkerd manages the  jaeger extension of Linkerd service mesh.`,
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// enable / disable logging
+		if verbose {
+			log.SetLevel(log.DebugLevel)
+		} else {
+			log.SetLevel(log.PanicLevel)
+		}
+
+		controlPlaneNamespaceFromEnv := os.Getenv("LINKERD_NAMESPACE")
```
adleong, line 52: why do we have this and the `linkerd-namespace` flag?

jaeger/cmd/root.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// These regexs are not as strict as they could be, but are a quick and dirty
+	// sanity check against illegal characters.
+	alphaNumDash = regexp.MustCompile(`^[a-zA-Z0-9-]+$`)
+
+	controlPlaneNamespace string
+	apiAddr               string // An empty value means "use the Kubernetes configuration"
+	kubeconfigPath        string
+	kubeContext           string
+	defaultNamespace      string // Default namespace taken from current kubectl context
```
adleong, line 33: I think this is unused.

jaeger/values/values.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+
+	return v, nil
+}
+
+// readDefaults read all the default variables from the values.yaml file.
+// chartDir is the root directory of the Helm chart where values.yaml is.
+func readDefaults(chartDir string) (*Values, error) {
+	valuesFiles := []*chartutil.BufferedFile{
```
adleong, line 45: this slice only has one element so I don't think we need a slice here.

jaeger/values/values.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	if err := charts.FilesReader(chartDir, valuesFiles); err != nil {
+		return nil, err
+	}
+
+	values := Values{}
+	for _, valuesFile := range valuesFiles {
+		var v Values
+		if err := yaml.Unmarshal(charts.InsertVersion(valuesFile.Data), &v); err != nil {
```
adleong, line 56: we never use the version placeholder in these values, so we don't need to replace it.

jaeger/values/values.go
```diff
@@ -0,0 +1,83 @@
...
+			return nil, err
+		}
+
+		var err error
+		values, err = values.merge(v)
```
adleong, line 61: there's only one values file, so no need to do any merging.

adleong (commented):
Thanks @Pothulapati, this is a great start!  I think we'll want to move everything, including the chart, into the top level `jaeger` directory.

I noticed that you left the original tracing chart alone in this PR which I think is the right approach.  There will need to be follow-up work to switch the main control plane to send traces to this collector instead and then to eventually delete the old tracing addon chart.

jaeger/charts/jaeger/templates/tracing.yaml
```diff
@@ -0,0 +1,187 @@
...
+    spec:
+      containers:
+      - command:
+        - /occollector_linux
+        - --config=/conf/collector-config.yaml
+        env:
+        - name: GOGC
+          value: "80"
+        image: {{.Values.collector.image.name}}:{{.Values.collector.image.version}}
+        imagePullPolicy: {{.Values.collector.image.imagePullPolicy}}
```
alpeb, line 95: In `values.yaml` you have `collector.image.pullPolicy`

jaeger/cmd/root.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func init() {
+	defaultNamespace = getDefaultNamespace()
+	RootCmd.PersistentFlags().StringVarP(&controlPlaneNamespace, "linkerd-namespace", "L", defaultLinkerdNamespace, "Namespace in which Linkerd is installed [$LINKERD_NAMESPACE]")
+	RootCmd.PersistentFlags().StringVar(&kubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests")
+	RootCmd.PersistentFlags().StringVar(&kubeContext, "context", "", "Name of the kubeconfig context to use")
+	RootCmd.PersistentFlags().StringVar(&impersonate, "as", "", "Username to impersonate for Kubernetes operations")
+	RootCmd.PersistentFlags().StringArrayVar(&impersonateGroup, "as-group", []string{}, "Group to impersonate for Kubernetes operations")
+	RootCmd.PersistentFlags().StringVar(&apiAddr, "api-addr", "", "Override kubeconfig and communicate directly with the control plane at host:port (mostly for testing)")
```
alpeb: I didn't see `controlPlaneNamespace` nor `apiAddr` being used. Any plans on using them later on?

alpeb (commented):
Looking good :+1: 
FYI, I'll be implementing the webhook on top of this branch.

Do you mind add completing the entries in `bin/helm-build` for this new chart?

jaeger/cmd/root.go
Pothulapati: Yeah, `controlPlaneNamespace` is removed but other k8s related flags should be useful for the check stuff.

pkg/charts/charts.go
```diff
@@ -22,14 +23,16 @@ type Chart struct {
 	Namespace string
 	RawValues []byte
 	Files     []*chartutil.BufferedFile
+	Fs        http.FileSystem
```
adleong, line 26: It looks like Fs and Dir are always combined.  Any reason we can't just turn the `Dir` field into a `http.FileSystem` so that we don't need both?  (I think `os.File` would probably be better, but that's a larger change)

jaeger/charts/jaeger/templates/rbac.yaml
```diff
@@ -0,0 +1,26 @@
...
+###
+### collector RBAC
+###
+---
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: collector
+  namespace: {{.Values.namespace}}
+  labels:
```
adleong, line 11: what is the purpose of these labels?

jaeger/charts/jaeger/values.yaml
```diff
@@ -0,0 +1,15 @@
...
+  image: 
+    name: omnition/opencensus-collector
+    version: 0.1.11
+    pullPolicy: Always
+  # resources:
+jaeger:
+  image: 
+    name: jaegertracing/all-in-one
+    version: 1.19.2
+    pullPolicy: Always
```
adleong, line 15: let's try to get newlines at the end of all files unless there's a reason not to

adleong (commented):
Since this new chart is unused by the current CLI, merging this should be a safe thing to do.  Let's prioritize getting the issues raised on this PR fixed up so that we can merge this to main before adding more functionality (such as validation, new flags, etc.). We can then work on that stuff in follow-up PRs.

jaeger/charts/jaeger/templates/rbac.yaml
Pothulapati: These are not helpful for `serviceAccounts`, but I kept them as we had them for services and deployments.

Removing them.

jaeger/values/values.go
Pothulapati: we still need a slice as the func `FileReader` expects a slices in the next lines

pkg/charts/charts.go
Pothulapati: I think its better this way, as not all code paths have to configure the filesystem field, they just use `static.Templates` which is the default i.e `/charts` and set the `Dir` based on their chart name.

If we combine them, Every `Chart` creator code-path will have to configure the `http.FileSystem` field to their chart directory. WDYT?

jaeger/charts/jaeger/templates/tracing.yaml
```diff
@@ -0,0 +1,187 @@
...
+###
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: collector-config
+  namespace: {{.Values.namespace}}
+  labels:
+    component: collector
+    namespace: {{.Values.namespace}}
```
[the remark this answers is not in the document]
Pothulapati: We removed the full `namespace` label for now, as it is not useful for anything. where as `component` is used for services x deployments.

pkg/charts/charts.go
```diff
@@ -109,12 +109,9 @@ func (c *Chart) RenderNoPartials() (bytes.Buffer, error) {
...
 
 // ReadFile updates the buffered file with the data read from disk
-func ReadFile(dir string, f *chartutil.BufferedFile) error {
+func ReadFile(fs http.FileSystem, dir string, f *chartutil.BufferedFile) error {
 	filename := dir + f.Name
-	if dir == "" {
-		filename = filename[7:]
-	}
-	file, err := static.Templates.Open(filename)
+	file, err := fs.Open(filename)
```
alpeb, line 114: Here you're reading directly from `fs`. When doing `linkerd install`, this ends up being `static.WithDefaultChart(helmDefaultChartDir)` which points directly to the "real" file system.
Before, we had `static.Templates.Open(filename)` which hits the "real" file system when `!prod` and the VFS when `prod`. So this is why it's failing the integration test (trying to hit the real FS instead of the VFS).

adleong (commented):
Lets push the CLI and chart rendering stuff out of this PR so that we can merge the chart itself.

Then we'll need to open a new PR for the chart rendering changes.  I didn't know about the VFS stuff we do here.  I think the jaeger CLI package will need a function similar to the main CLI's `static.Templates` which returns a FileSystem which may be either a VFS of it charts directory or may point to the real FS.

pkg/charts/charts.go
Pothulapati: Yep, That makes sense, As `static.Templates` is replaced by the logic to read from a VFS with code generation which our helper code is not doing similarly.

Pothulapati:
@adleong I am reverting the last commit, which involved the merging of `c,Fs` and `c.DIr` which added all this complexity.

I was trying to create a separate VFS generation specifically for `jaeger` chart by defining a `JaegerTemplates` but even that fails because of https://github.com/shurcooL/vfsgen/issues/23

For Now, I will revert the last commit and move `jaeger` chart to higher level `charts` directory so that even render works. We can raise separate PRs to fix this. WDYT?

alpeb (approved):
LGTM :+1: , looking forward for the followup PRs :wink:

adleong:
I've branched off of this PR to create https://github.com/linkerd/linkerd2/pull/5275 as a replacement which moves the jaeger charts out of the main charts directory as we discussed.  It deletes the jaeger CLI for now.

Pothulapati:
Created #5278 with only the CLI changes. Will close this once both of them are merged!

PR #5241, opened 2020-11-17
cli: Support custom namespace installation

### LINKED ISSUE #5239: Inability to specify `--linkerd-namespace` via the CLI in stable-2.9.0
### What is the issue?

When attempting to upgrade to `stable-2.9.0` by generating the YAML files using the `linkerd` command line interface, specifying a non-default namespace via the `--linkerd-namespace` no longer seems to be respected and the default `linkerd` namespace is used instead

### How can it be reproduced?

```linkerd  install --ignore-cluster --linkerd-namespace=custom-namespace```

When inspecting the output YAML, the `custom-namespace` namespace does not appear. When running the same command on `stable-2.8.0` it does.

### Environment

- OS: MacOS
- Linkerd version: stable-2.9.0

Pothulapati:
Fixes #5239 

We did a revamp of the flags logic recently, and most flags had
relevant logic and what has to be done, except for
`--linkerd-namespace`.

I'm guessing the reason we forgot this is that this specific flag is a root flag
i.e applicable under all commands.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -294,6 +294,10 @@ func install(ctx context.Context, w io.Writer, values *l5dcharts.Values, flags [
 }
 
 func render(w io.Writer, values *l5dcharts.Values, stage string) error {
+
+	// Set any global flags if present, common with install and upgrade
+	values.Global.Namespace = controlPlaneNamespace
+
 	// Render raw values and create chart config
 	rawValues, err := yaml.Marshal(values)
 	if err != nil {
--- a/cli/cmd/options.go
+++ b/cli/cmd/options.go
@@ -231,6 +231,13 @@ func makeAllStageFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.FlagSet)
 				return nil
 			}),
 
+		flag.NewStringFlag(allStageFlags, "linkerd-namespace", defaults.Global.Namespace,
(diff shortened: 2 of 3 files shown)


kleimkuhler (approved):
lgtm 👍

cpretzer (approved):
Nice work @Pothulapati 

This is working in my local env

cpretzer:
@Pothulapati Looking through the tests, I don't see one that tests a custom `linkerd-namespace` on install. Did I overlook it?

If there isn't one, we should add it.

adleong:
This adds an additional `linkerd-namespace` flag in addition to the one defined in the persistent flags in `root.go`.  I assume that the new one replaces the persistent one in the subcommands that it's added to, but this is confusing.  For example, it will mean that the CLI package global variable `controlPlaneNamespace` may not match the `globals.namespace` field in the values struct.

What about this instead: after creating a new values struct in `install.go` and `upgrade.go`, we set the global.namespace field to the `controlPlaneNamespace` (which was populated from the persistent root level flag).

Pothulapati:
Yep, That makes sense!

> after creating a new values struct in install.go and upgrade.go, we set the global.namespace field to the controlPlaneNamespace (which was populated from the persistent root level flag).

How about adding that into `render()` which is common between installl and upgrade? Considering folks would want cli flag to take highest precedence when set (instead of current stored config), It should work as expected.

adleong (approved):
Nice!

PR #5245, opened 2020-11-17
Consolidate integration tests under k3d

### LINKED ISSUE #5007: Migrate integration test infra from KinD to k3d

alpeb:
Fixes #5007

Simplified integration tests by moving all to k3d. Previously things were running in KinD, except for the multicluster tests, which implied some extra complexity in the supporting scripts.

Removed the KinD config files under `test/integration/configs`, as config is now passed as flags into the `k3d` command.

Also renamed `kind_integration.yml` to `integration_tests.yml`

Test skipping logic under ARM was also simplified.

Finally, TEST.md has now more details about the special integration tests and BUILD.md now refers to k3d instead of minikube.


cpretzer (approved):
Nice cleanup @alpeb 

LGTM :+1:

adleong (approved):

PR #5251, opened 2020-11-18
Fixes to code quality issues detected on Static Analysis

de-sh:
Hi, as discussed in https://github.com/linkerd/linkerd2/discussions/5250, I had recently been introduced to the linkerd2 codebase and decided to run it through the DeepSource static analyzer and found some interesting and concerning issues related to performance, bugs and security flaws. You can find [the report here](https://deepsource.io/gh/de-sh/linkerd2). I have fixed a few and hope to work on some more, soon.

Changes:
- Simplify boolean expressions

### DIFF
--- a/cli/cmd/routes.go
+++ b/cli/cmd/routes.go
@@ -330,7 +330,7 @@ func buildTopRoutesRequest(resource string, options *routesOptions) (*pb.TopRout
 		LabelSelector: options.labelSelector,
 	}
 
-	options.dstIsService = !(target.GetType() == k8s.Authority)
+	options.dstIsService = target.GetType() != k8s.Authority
 
 	if options.toResource != "" {
 		if options.toNamespace == "" {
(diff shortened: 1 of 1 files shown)


adleong (commented):
Hi @de-sh!

Thanks for looking into this.  I took a quick look at the report and there's definitely some interesting stuff in there.  For now, I think I'd like to keep this tooling out of the repo itself (i.e. not merge the `.deepsource.toml` file) since it's not part of the core development workflow.

As for the code changes, I'd like to consider them on an individual basis.  I'm not aware of any serious performance issues in the CLI or control plane so I'd want to avoid premature optimization.  Unless there are significant performance issues, I think we should prioritize readability of the code.

For example, the changes in `routes.go` which simplify the boolean logic seem like good changes to me.  On the other hand, I think doing a sequence of multiple `append` calls is more readable than doing all the appends at once.

de-sh:
Ok, I have reverted a few commits, any other fixes from the report you'd like me to add to this PR? If not in this PR, do let me know if it'd be a good idea to open others and what all fixes you believe need to be made urgently. :D

adleong (approved):
Great!  I don't see anything in the report that I'd say is critical.

kleimkuhler (approved):
Thanks @de-sh

PR #5256, opened 2020-11-19
Refactor webhook framework to allow webhooks define their flags

alpeb:
Pulled out of `launcher.go` the flag parsing logic and moved it into the `Main` methods of the webhooks (under `controller/cmd/proxy.injector/main.go` and `controller/cmd/sp-validator/main.go`), so that individual webhooks themselves can define the flags they want to use.

Also no longer require that webhooks have cluster-wide access.

Finally, renamed the type `webhook.handlerFunc` to `webhook.Handler` so it can be exported. This will be used in the upcoming jaeger webhook.

### DIFF
--- a/controller/cmd/proxy-injector/main.go
+++ b/controller/cmd/proxy-injector/main.go
@@ -22,7 +22,7 @@ func Main(args []string) {
 	webhook.Launch(
 		context.Background(),
 		[]k8s.APIResource{k8s.NS, k8s.Deploy, k8s.RC, k8s.RS, k8s.Job, k8s.DS, k8s.SS, k8s.Pod, k8s.CJ},
-		injector.Inject(*foobar),
+		injector.Inject,
 		"linkerd-proxy-injector",
 		*metricsAddr,
 		*addr,
--- a/controller/cmd/proxy-injector/main.go
+++ b/controller/cmd/proxy-injector/main.go
@@ -2,21 +2,30 @@ package proxyinjector
 
 import (
 	"context"
+	"flag"
+	"fmt"
 
 	"github.com/linkerd/linkerd2/controller/k8s"
 	injector "github.com/linkerd/linkerd2/controller/proxy-injector"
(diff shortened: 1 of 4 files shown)


controller/cmd/proxy-injector/main.go
```diff
@@ -2,21 +2,30 @@ package proxyinjector
...
+	"github.com/linkerd/linkerd2/pkg/flags"
 )
 
 // Main executes the proxy-injector subcommand
 func Main(args []string) {
+	cmd := flag.NewFlagSet("proxy-injector", flag.ExitOnError)
+	metricsAddr := cmd.String("metrics-addr", fmt.Sprintf(":%d", 9995), "address to serve scrapable metrics on")
+	addr := cmd.String("addr", ":8443", "address to serve on")
+	kubeconfig := cmd.String("kubeconfig", "", "path to kubeconfig")
+	flags.ConfigureAndParse(cmd, args)
```
alpeb, line 20: The upcoming jaeger webhook will define a couple of extra flags here. We'll pass them to the handler function below (`Injector.Inject`)  in this case.

controller/webhook/launcher.go
```diff
@@ -2,36 +2,33 @@ package webhook
...
+	metricsAddr string,
+	addr string,
+	kubeconfig string,
+) {
 	stop := make(chan os.Signal, 1)
 	defer close(stop)
 	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
 
-	k8sAPI, err := k8s.InitializeAPI(ctx, *kubeconfig, true, APIResources...)
+	k8sAPI, err := k8s.InitializeAPI(ctx, kubeconfig, false, APIResources...)
```
alpeb, line 31: Webhooks shouldn't require cluster-wide access.

adleong (approved):
Nice!

Pothulapati (approved):
LGTM!

PR #5266, opened 2020-11-20
Add destination service tests for the IP path

kleimkuhler:
This adds additional tests for the destination service that assert `GetProfile`
behavior when the path is an IP address.

1. Assert that when the path is a cluster IP, the configured service profile is
   returned.
2. Assert that when the path a pod IP, the endpoint field is populated in the
   service profile returned.
3. Assert that when the path is not a cluster or pod IP, the default service
   profile is returned.
4. Assert that when path is a pod IP with or without the controller annotation,
the endpoint has or does not have a protocol hint

### DIFF
--- a/controller/api/destination/server_test.go
+++ b/controller/api/destination/server_test.go
@@ -15,7 +15,8 @@ import (
 
 const fullyQualifiedName = "name1.ns.svc.mycluster.local"
 const clusterIP = "172.17.12.0"
-const podIP = "172.17.0.12"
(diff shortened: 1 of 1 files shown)


kleimkuhler:
Depends on #5267 which fixes an issue where the protocol hint is always set on endpoint responses.

adleong (approved):

alpeb (approved):
LGTM :+1:

PR #5267, opened 2020-11-20
Check correct label value when setting protocol hint

kleimkuhler:
This fixes an issue where the protocol hint is always set on endpoint responses.

We now check the right value which determines if the pod has the required label.


kleimkuhler:
A test for this has been added to #5266

olix0r (approved):

adleong (approved):

PR #5275, opened 2020-11-23
tracing: new jaeger independent helm chart

### LINKED ISSUE #5230: Move tracing chart into jaeger package

adleong:
In the interest of speed, I have branched this PR off of #5240.  As we discussed on that PR, I have moved the jaeger charts into the top level `jaeger` directory.  I have removed the jaeger CLI commands for now.

Further follow up work is required to add back the jaeger CLI and to allow it to render charts from its own charts directory (as opposed to from the main control plane's charts directory).  In particular, special attention must be given to making sure that it can load the templates from either the real filesystem (to support local development) or from a VFS (to support a standalone binary).

Fixes #5230

This PR moves tracing into a jaeger chart with no proxy injection
templates. We still keep the dependency on partials, as we could use
common templates like resources, etc from there.

Signed-off-by: Tarun Pothulapati tarunpothulapati@outlook.com


Pothulapati (approved):
LGTM

alpeb (approved):
LGTM :+1:

PR #5278, opened 2020-11-24
extension: Add new jaeger binary

Pothulapati:
Depends on #5275 

This branch adds a new jaeger binary project in the jaeger directory.
This follows the same logic as that of `linkerd install`. But as
`linkerd install` VFS logic expects charts to be present in `/charts`
directory, This command gets its own static pkg to generate its own
VFS for its chart.


jaeger/static/generate.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"github.com/linkerd/linkerd2/jaeger/static"
+	"github.com/shurcooL/vfsgen"
+	log "github.com/sirupsen/logrus"
+)
+
+func main() {
+	err := vfsgen.Generate(static.Templates, vfsgen.Options{
+		Filename:     "generated_templates.gogen.go",
```
alpeb, line 13: Nit: Can you name this something different, like `generated_jaeger_templates.gogen.go` so that the output from `bin/build-cli-bin` is a little more expressive? Currently we have:
```bash
$ bin/build-cli-bin 
writing generated_templates.gogen.go
writing generated_templates.gogen.go
target/cli/linux-amd64/linkerd-jaeger
```
(would be best if the other generated file for the linkerd binary is renamed as well...)

bin/build-cli-bin
```diff
@@ -20,5 +20,9 @@ rootdir=$( cd "$bindir"/.. && pwd )
     GO111MODULE=on go generate -mod=readonly ./pkg/charts/static # TODO: `go generate` does not honor -mod=readonly
     root_tag=$("$bindir"/root-tag)
     GO111MODULE=on CGO_ENABLED=0 go build -o $target -tags prod -mod=readonly -ldflags "-s -w -X github.com/linkerd/linkerd2/pkg/version.Version=$root_tag" ./cli
+
+    target=target/cli/$(os)/linkerd-jaeger
+    GO111MODULE=on go generate -mod=readonly ./jaeger/static # TODO: `go generate` does not honor -mod=readonly
+    GO111MODULE=on CGO_ENABLED=0 go build -o $target -tags prod -mod=readonly ./jaeger
     echo "$target"
```
alpeb, line 27: Can you also leave `echo "$target"` after the linkerd2 binary is generated right above?

alpeb (commented):
This looks good to me :+1:  I only have a few nits.

There's a typo in the `jaeger/static/tempaltes.go` file name :wink:

jaeger/cmd/root.go
```diff
@@ -0,0 +1,65 @@
...
+	Long:  `jaeger manages the jaeger extension of Linkerd service mesh.`,
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// enable / disable logging
+		if verbose {
+			log.SetLevel(log.DebugLevel)
+		} else {
+			log.SetLevel(log.PanicLevel)
+		}
+
+		controlPlaneNamespaceFromEnv := os.Getenv("LINKERD_NAMESPACE")
```
adleong, line 43: Why do we need to read the linkerd namespace from an env variable in addition to the flag?

Pothulapati: This follows the same logic of Linkerd, where users can also set the linkerd namespace using env and hence our binary also should honor that?

adleong (commented):
We'll also need to update everywhere that we call `go generate` to also generate the VFS for each bundled extension.

adleong (approved):
Nice!  When first building this I had to manually delete the `pkg/charts/static/generated_templates.gogen.go` file left over from the last time I built.  Once I did that, everything worked as expected.

jaeger/cmd/root.go
adleong: Not sure why that's there in the main CLI either, but I don't think we need to copy that behavior.  Let's not do stuff unless we know why.

alpeb (approved):
Thanks @Pothulapati , this looks good to me :ship:

PR #5281, opened 2020-11-25
Enable linkerd-proxy relabel configs additions

### LINKED ISSUE #5280: Way to exclude pods from prometheus scrape
When adding Prometheus and nginx-ingress to your mesh, they produce a huge amount of metrics, which makes Prometheus use an excessive amount of memory.

We should provide the user with the option to edit the scrape configs of the Prometheus add on, to drop metrics from specific deployments.

Remove the proxy from those targets, but I want the proxy in order to be able to take advantage of traffic splits.

Values.yaml

shaikatz:
Fix #5280

### DIFF
--- a/charts/add-ons/prometheus/templates/prometheus.yaml
+++ b/charts/add-ons/prometheus/templates/prometheus.yaml
@@ -98,6 +98,9 @@ data:
       kubernetes_sd_configs:
       - role: pod
       relabel_configs:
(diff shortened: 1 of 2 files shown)


adleong:
Hi @shaikatz, I'm going to mark this PR as a draft until we work through the questions on the associated issue (#5280)

PR #5282, opened 2020-11-25
Have webhooks refresh their certs automatically

alpeb:
Fixes partially #5272

In 2.9 we introduced the ability for providing the certs for `proxy-injector` and `sp-validator` through some external means like cert-manager, through the new helm setting `externalSecret`.
We forgot however to have those services watch changes in their secrets, so whenever they were rotated they would fail with a cert error, with the only workaround being to restart those pods to pick the new secrets.

This addresses that by first abstracting out `FsCredsWatcher` from the identity controller, which now lives under `pkg/tls`.

The webhook's logic in `launcher.go` no longer reads the certs before starting the https server, moving that instead into `server.go` which in a similar way as identity will receive events from `FsCredsWatcher` and update `Server.cert`. We're leveraging `http.Server.TLSConfig.GetCertificate` which allows us to provide a function that will return the current cert for every incoming request.

### How to test

```bash
# Create some root cert
$ step certificate create linkerd-proxy-injector.linkerd.svc ca.crt ca.key \
  --profile root-ca --no-password --insecure --san linkerd-proxy-injector.linkerd.svc

# configure injector's caBundle to be that root cert
$ cat > linkerd-overrides.yaml << EOF
proxyInjector:
  externalSecret: true
  caBundle: |
    < ca.crt contents>
EOF

# Install linkerd. The injector won't start untill we create the secret below
$ bin/linkerd install --controller-log-level debug --config linkerd-overrides.yaml | k apply -f -

# Generate an intermediatery cert with short lifespan
$ step certificate create linkerd-proxy-injector.linkerd.svc ca-int.crt ca-int.key --ca ca.crt --ca-key ca.key --profile intermediate-ca --not-after 4m --no-password --insecure --san linkerd-proxy-injector.linkerd.svc

# Create the secret using that intermediate cert
$ kubectl create secret tls \
  linkerd-proxy-injector-k8s-tls \
   --cert=ca-int.crt \
   --key=ca-int.key \
   --namespace=linkerd

# start following the injector log
$ k -n linkerd logs -f -l linkerd.io/control-plane-component=proxy-injector -c proxy-injector

# Inject emojivoto. The pods should be injected normally
$ bin/linkerd inject https://run.linkerd.io/emojivoto.yml | kubectl apply -f -

# Wait about 5 minutes and delete a pod
$ k -n emojivoto delete po -l app=emoji-svc

# You'll see it won't be injected, and something like "remote error: tls: bad certificate" will appear in the injector logs.

# Regenerate the intermediate cert
$ step certificate create linkerd-proxy-injector.linkerd.svc ca-int.crt ca-int.key --ca ca.crt --ca-key ca.key --profile intermediate-ca --not-after 4m --no-password --insecure --san linkerd-proxy-injector.linkerd.svc

# Delete the secret and recreate it
$ k -n linkerd delete secret linkerd-proxy-injector-k8s-tls
$ kubectl create secret tls \
  linkerd-proxy-injector-k8s-tls \
   --cert=ca-int.crt \
   --key=ca-int.key \
   --namespace=linkerd

# Wait a couple of minutes and you'll see some filesystem events in the injector log along with a "Certificate has been updated" entry
# Then delete the pod again and you'll see it gets injected this time
$ k -n emojivoto delete po -l app=emoji-svc

```

### DIFF
--- a/controller/webhook/server.go
+++ b/controller/webhook/server.go
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"sync"
+	"sync/atomic"
 
 	"github.com/linkerd/linkerd2/controller/k8s"
 	pkgk8s "github.com/linkerd/linkerd2/pkg/k8s"
@@ -36,8 +36,7 @@ type Server struct {
 	*http.Server
 	api       *k8s.API
 	handler   Handler
-	cert      *tls.Certificate
-	certMutex *sync.RWMutex
+	certValue atomic.Value
 	recorder  record.EventRecorder
 }
 
@@ -71,9 +70,10 @@ func NewServer(
 	})
 	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: component})
 
-	s := &Server{server, api, handler, nil, &sync.RWMutex{}, recorder}
+	var emptyCert atomic.Value
+	s := &Server{server, api, handler, emptyCert, recorder}
 	s.Handler = http.HandlerFunc(s.serve)
-	server.TLSConfig.GetCertificate = s.getCertificate()
+	server.TLSConfig.GetCertificate = s.getCertificate
 
 	if err := s.updateCert(); err != nil {
 		log.Fatalf("Failed to initialized certificate: %s", err)
@@ -101,9 +101,7 @@ func (s *Server) updateCert() error {
 	if err != nil {
 		return err
 	}
-	s.certMutex.Lock()
-	defer s.certMutex.Unlock()
-	s.cert = &cert
+	s.certValue.Store(&cert)
 	log.Debug("Certificate has been updated")
 	return nil
 }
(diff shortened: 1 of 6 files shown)


controller/webhook/server.go
```diff
@@ -31,28 +34,33 @@ type Handler func(
 // Server describes the https server implementing the webhook
 type Server struct {
 	*http.Server
-	api      *k8s.API
-	handler  Handler
-	recorder record.EventRecorder
+	api       *k8s.API
+	handler   Handler
+	cert      *tls.Certificate
```
adleong, line 39: TIOLI but storing the cert in an atomic.Value might be safer than protecting it with a mutex.  This is very similar to one of the examples given at https://golang.org/pkg/sync/atomic/#Value

controller/webhook/server.go
```diff
@@ -79,6 +119,31 @@ func (s *Server) Start() {
 	}
 }
 
+// getCertificate returns a function that provides the TLS server with the current cert
+func (s *Server) getCertificate() func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
```
adleong, line 123: it is possible to not return an anonymous function here and instead just use this function directly?  i.e. set `server.TSLConfig.GetCertificate = s.getCertificate`

controller/webhook/server.go
alpeb: Ok that's cool in that it avoids us from declaring both `cert` and its mutex, and have instead just an `atomic.Value` we can use. But the approach should be equally safe, no?

adleong: I mean safe in terms of preventing programmer error.  It's easy to accidentally write code that accesses the field without first acquiring the mutex (which can lead to a data race) or forgetting to release the mutex (which can lead to a deadlock).  It's much harder to use an atomic value wrong.

controller/webhook/server.go
alpeb: Yep, that does the trick as well :+1:

adleong (approved):

kleimkuhler (approved):
Looks good! Thanks for the testing instructions 👍

PR #5285, opened 2020-11-26
fix: change square by round brackets in cli usage manual (#3999)

### LINKED ISSUE #3999: linkerd completion command does not work as expected
## Bug Report
The `linkerd completion` command does not work as expected on zsh.

### What is the issue?
The completion scripts are generated, but sourcing them does not provide shell completion. Not sure if this error is restricted to only zsh.

### How can it be reproduced?
Run:
```bash
$ source <(linkerd completion zsh)
```
Completion does not work
### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version: 1.16
- Cluster Environment: (GKE, AKS, kops, ...) minikube
- Host OS: linux
- Linkerd version: stable-2.6

### Additional context
Could be an issue with cobra completion script generator. Check https://github.com/spf13/cobra/issues/881

jiraguha:
Signed-off-by: jpiraguha <jiraguha@gmail.com>

Fixes #3999

### DIFF
--- a/cli/cmd/completion.go
+++ b/cli/cmd/completion.go
@@ -11,25 +11,30 @@ import (
 // newCmdCompletion creates a new cobra command `completion` which contains commands for
 // enabling linkerd auto completion
 func newCmdCompletion() *cobra.Command {
-	example := `  # bash <= 3.2
+	example := `  # bash <= 3.2:
   source /dev/stdin <<< "$(linkerd completion bash)"
 
-  # bash >= 4.0
+  # bash >= 4.0:
   source <(linkerd completion bash)
 
-  # bash <= 3.2 on osx
+  # bash <= 3.2 on osx:
   brew install bash-completion # ensure you have bash-completion 1.3+
   linkerd completion bash > $(brew --prefix)/etc/bash_completion.d/linkerd
 
-  # bash >= 4.0 on osx
+  # bash >= 4.0 on osx:
   brew install bash-completion@2
   linkerd completion bash > $(brew --prefix)/etc/bash_completion.d/linkerd
 
-  # zsh
-  source <(linkerd completion zsh)
+  # zsh:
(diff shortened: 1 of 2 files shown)


jiraguha:
@olix0r pipeline fails on integration `Integration tests (external-issuer)` I can tell why... a small help is needed thanks

olix0r:
@jiraguha Thanks for the help fixing this issue!

I'm assuming that test failure is CI flakiness -- rerunning now to confirm.

olix0r (approved):
Tested locally. works.

alpeb (commented):
This tested fine in my Linux machine, as long as I followed the instructions not under "# zsh" but under "# zsh on osx / oh-my-zsh" (also from looking at [cobra's docs](https://github.com/spf13/cobra/blob/master/shell_completions.md), I only see the latter method being mentioned) . @jiraguha would you amend the instructions in `linkerd completion` for zsh so we get rid of the former? Thanks!

...By the way, I think we'll submit this as a plugin to oh-my-zsh to make this work out of the box :wink:

jiraguha:
@alpeb we should be good now, I followed the same template as `cobra`. For `oh-my-zsh` I can open an issue and a PR's for linkerd so we could collaborate to make it work. I would be happy to take the subject :-)

alpeb (approved):
Excellent, LGTM :+1: 

> For oh-my-zsh I can open an issue and a PR's for linkerd so we could collaborate to make it work. I would be happy to take the subject :-)

That'd be awesome, thanks!
Are you interested in linkerd swag? Please fill out this [form](https://docs.google.com/forms/d/e/1FAIpQLSdXA7kAl6oyUQ4DFRjiDc2biWlvkhNirrMwsXaQ8V3Uh8Yj5A/viewform) :-)

PR #5175, opened 2020-11-04
Add automatic readme generation for charts

GMarkfjard:
The current readmes for each chart is generated
manually and doesn't contain all the information available.

Utilize helm-docs to automatically fill out readme.mds
for the helm charts by pulling metadata from values.yml.

Validation done by testing locally.

Fixes #4156

PR #5293, opened 2020-11-27
extension: Separate multicluster chart and binary

### LINKED ISSUE #5257: Move multicluster charts to top-level multicluster directory
To follow the structure of a Linkerd extension, the linkerd multicluster charts should be moved into a separate top-level directory called `multicluster`.

Pothulapati:
Fixes #5257

This branch movies mc charts and cli level code to a new
top level directory. None of the logic is changed.

Also, moves some common types into `/pkg` so that they
are accessible both to the main cli and extensions.


pkg/k8s/resource.go
```diff
@@ -0,0 +1,62 @@
+package k8s
```
alpeb, line 1: WDYT about putting this under a `resource` package and remove all the appearances of the "Resource" word in the functions below? So they'd be called like `resource.Render()`, etc.

alpeb (commented):
Nice that you splitted `multicluster.go` up, and also that you extracted the resource stuff out of `uninstall.go`  :+1:  :+1: 

Maybe we can get away without using the `linkerd2` prefix in the chart directories?

Also, you missed updating `setValues()` in `bin/helm-build` :wink:

alpeb:
@Pothulapati You're planning on moving the service mirror controller code in a separate PR?

adleong:
Should we hold this PR until https://github.com/linkerd/linkerd2/pull/5175 merges?

multicluster/cmd/root.go
```diff
@@ -0,0 +1,92 @@
...
+functionality of Linkerd. You can use it to install the service mirror
+components on a cluster, manage credentials and link clusters together.`,
+		Example: `  # Install multicluster addons.
+  linkerd --context=cluster-a multicluster install | kubectl --context=cluster-a apply -f -
+
+  # Extract mirroring cluster credentials from cluster A and install them on cluster B
+  linkerd --context=cluster-a multicluster link --cluster-name=target | kubectl apply --context=cluster-b -f -`,
+	}
+
+	multiclusterCmd.PersistentFlags().StringVarP(&controlPlaneNamespace, "linkerd-namespace", "L", defaultLinkerdNamespace, "Namespace in which Linkerd is installed [$LINKERD_NAMESPACE]")
```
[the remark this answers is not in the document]
Pothulapati: we actually don't read the env variable. So, removing it!

alpeb:
> Should we hold this PR until #5175 merges?

I think we should

multicluster/static/templates.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// Templates that will be rendered by `linkerd install`. This is only used on
+// dev builds.
+var Templates http.FileSystem = http.Dir(path.Join(getRepoRoot(), "multicluster/charts"))
+
+// getRepoRoot returns the full path to the root of the repo. We assume this
+// function is only called from the `Templates` var above, and that this source
+// file lives at `pkg/charts/static`, relative to the root of the repo.
+func getRepoRoot() string {
```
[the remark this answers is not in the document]
Pothulapati: Just realized, we can use the same fn from `getRepoRoot` everywhere. So, doing that.

pkg/k8s/resource.go
Pothulapati: Hmmm, I like that way but still we can't remove the `Resource` word from the fn's right? How would they named be instead?

Pothulapati:
@alpeb Moved the service-mirror controller code in this PR too!

pkg/k8s/resource.go
alpeb:
```go
foo := resource.New()
foo = resource.NewNamespaced()
foo.Render()
```

Pothulapati: Done 👍

alpeb: To [avoid stuttering](https://blog.golang.org/package-names) I would get rid of all mentioning of "resource" in that file, besides the package name itself. So instead of `type KubernetesResource` I would put `type Kubernetes` and instead of `func (r KubernetesResource) RenderResource()` it'd be `func (r Kubernetes) Render()`. You can even use an import alias to make things extra clear: `k8sResource "github.com/linkerd/linkerd2/pkg/k8s/resource"`

Pothulapati: Good Suggestion! _(Also great blog suggestion, made things clear)_

> import alias to make things extra clear: k8sResource "github.com/linkerd/linkerd2/pkg/k8s/resource"
Does not seem that useful, so skipped this!

alpeb (approved):
Thanks @Pothulapati LGTM modulo the merge conflicts :+1:

kleimkuhler (approved):
Looks good 👍

adleong (approved):

PR #5295, opened 2020-11-30
jaeger: add check sub command

Pothulapati:
This adds a new `linkerd jaeger check` command to have checks w.r.t
jaeger extension. This is similar to that of the `linkerd check` cmd.
As jaeger is a separate package, It was a bit complex for this to work
as not all types and fields from healthcheck pkg are public, Helper
funcs were used to mitigate this.

This has the following changes:

- Adds a new `check.go` file under the jaeger extension pkg
- Moves some commonly needed funcs and types from `cli/cmd/check.go`
  and `pkg/healthcheck/health.go` into
  `pkg/healthcheck/healthcheck_output.go`.

Example Output
```bash
tarun in dev in on  kind-kind () linkerd2 on  tarun/jaeger-check [$!?] via 🐹 v1.15.4 took 2s
❯ ./bin/go-run cli jaeger check
linkerd-jaeger
--------------
√ collector service account exists
√ jaeger service account exists
√ collector config map exists
‼ collector pod is running
    collector-69cc44dfbc-8qkxw status is Pending
‼ jaeger pod is running
    jaeger-6f98d5c979-5vnpt status is Pending
√ jaeger extension pods are injected

Status check results are √

tarun in dev in on  kind-kind () linkerd2 on  tarun/jaeger-check [$!?] via 🐹 v1.15.4 took 2s
❯ ./bin/go-run cli jaeger check
linkerd-jaeger
--------------
√ collector service account exists
√ jaeger service account exists
√ collector config map exists
√ collector pod is running
√ jaeger pod is running
√ jaeger extension pods are injected

Status check results are √
```


pkg/healthcheck/healthcheck.go
```diff
@@ -2399,17 +2445,20 @@ func getPodStatuses(pods []corev1.Pod) map[string]map[string][]corev1.ContainerS
 	statuses := make(map[string]map[string][]corev1.ContainerStatus)
 
 	for _, pod := range pods {
-		if pod.Status.Phase == corev1.PodRunning && strings.HasPrefix(pod.Name, "linkerd-") {
+		if pod.Status.Phase == corev1.PodRunning {
```
adleong, line 2448: This existing logic here seems very confusing and overly complex and seems to have a lot of assumptions about the naming scheme of the Linkerd control plane baked in.  I think we should probably just rewrite these helper functions to be simpler.  But instead of doing that here, can we just leave this stuff alone for now and create more straightforward methods to use for the jaeger checks?  

All we need is a helper for determining if a container is running in a set of pods.  That should be as simple as iterating through the pods and their container statuses until we find a container with a matching name and then checking its status.

jaeger/cmd/check.go
```diff
@@ -0,0 +1,189 @@
...
+		namespace: "",
+		output:    healthcheck.TableOutput,
+	}
+}
+
+// checkFlagSet specifies flags allowed with and without `config`
+func (options *checkOptions) checkFlagSet() *pflag.FlagSet {
+	flags := pflag.NewFlagSet("check", pflag.ExitOnError)
+
+	flags.StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
```
adleong, line 119: This should be used for data plane checks, but I don't see those implemented here.

adleong (commented):
I don't see checks that the linkerd-proxy sidecar is present in for this extensions components or a check to see if data plane pods have the correct env variables injected.  Will those checks be follow ups to this PR?

pkg/healthcheck/healthcheck.go
Pothulapati: makes sense!

Pothulapati:
@adleong Yep, Plan to make the dataplane checks in a separate PR!

pkg/healthcheck/healthcheck.go
Pothulapati: Why not check for the pod Status itself? 🤔 Could be simpler. I did that rn. WDYT?

jaeger/cmd/check.go
Pothulapati: Removed it.

Should we not have a flag for user to be able to specify which ns to check for the jaeger extension pods checks? 🤔 Currently, Its dynamic (ie based on labels) as you suggested!

jaeger/cmd/check.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	return nil, fmt.Errorf("Could not find the namespace for extension %s", name)
+}
+
+func checkIfDataPlanePodsExist(pods []corev1.Pod) error {
+	for _, pod := range pods {
+		proxyContainer := false
+		for _, containerSpec := range pod.Spec.Containers {
+			if containerSpec.Name == k8s.ProxyContainerName {
+				proxyContainer = true
```
adleong, line 235: can't we just immediately return nil here?

jaeger/cmd/check.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+
+	return nil
+}
+
+// checkPodsStatus checks if the pod is in running state
+func checkPodsStatus(pods []corev1.Pod) error {
+	for _, pod := range pods {
+		fmt.Println(pod.Name, pod.Status.Phase)
```
adleong, line 250: is this leftover debugging output?

jaeger/cmd/check.go
```diff
@@ -0,0 +1,257 @@
...
+	checkers = append(checkers,
+		*healthcheck.NewChecker("collector pod is running", "", false, true, time.Time{}, false).
+			WithCheck(func(ctx context.Context) error {
+				// Check for Collector pod
+				kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+				if err != nil {
+					return err
+				}
+
+				pods, err := kubeAPI.GetPodsByNamespace(ctx, jaegerNamespace)
```
adleong, line 86: All of these checks look like they do the same thing: get a list of all pods in the namespace and ensure each of them is running.  Don't we want to select out the collector pod specifically here?

jaeger/cmd/check.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func jaegerCategory() *healthcheck.Category {
+
+	checkers := []healthcheck.Checker{}
+	checkers = append(checkers,
+		*healthcheck.NewChecker("collector service account exists", "", false, true, time.Time{}, false).
+			WithCheck(func(ctx context.Context) error {
+				// Check for Collector Service Account
+				kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
```
adleong, line 42: can we build the API client just once in the `jaegerCategory()` instead of in each check?

adleong (commented):
It would also be helpful to see sample output from running this command in the PR description.

jaeger/cmd/check.go
Pothulapati: We can do that but this way we can also report which pod does not have the proxy container :thinking: but its limited only for the first pod!

jaeger/cmd/check.go
Pothulapati: Makes sense.

Pothulapati:
@adleong Addressed the feedback.

Example Output
```bash
tarun in dev in on  kind-kind () linkerd2 on  tarun/jaeger-check [$!?] via 🐹 v1.15.4 took 2s
❯ ./bin/go-run cli jaeger check
linkerd-jaeger
--------------
√ collector service account exists
√ jaeger service account exists
√ collector config map exists
‼ collector pod is running
    collector-69cc44dfbc-8qkxw status is Pending
‼ jaeger pod is running
    jaeger-6f98d5c979-5vnpt status is Pending
√ jaeger extension pods are injected

Status check results are √

tarun in dev in on  kind-kind () linkerd2 on  tarun/jaeger-check [$!?] via 🐹 v1.15.4 took 2s
❯ ./bin/go-run cli jaeger check
linkerd-jaeger
--------------
√ collector service account exists
√ jaeger service account exists
√ collector config map exists
√ collector pod is running
√ jaeger pod is running
√ jaeger extension pods are injected

Status check results are √
```

jaeger/cmd/check.go
```diff
@@ -0,0 +1,232 @@
...
+func jaegerCategory() (*healthcheck.Category, error) {
+
+	kubeAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	checkers := []healthcheck.Checker{}
+	checkers = append(checkers,
+		*healthcheck.NewChecker("collector service account exists", "", false, true, time.Time{}, false).
```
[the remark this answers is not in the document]
alpeb: Also I kinda like better how we have things set up in `pkg/healthcheck/healthcheck.go` where you just instantiate the `Checker` struct. Using a long function call like this forces one to go dig into the declaration to see what each argument means.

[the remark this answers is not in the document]
alpeb: Shouldn't all these checks be `fatal`?

jaeger/cmd/check.go
adleong: oh, I see, this is inside the pods loop.  I think this would be clearer if we had a method that takes a pod and checks if that pod has a proxy container.  If we did it that way then we could immediately return nil from that method when a proxy contain is detected.  With that in place, this method becomes much simpler since we can immediately return an error when we encounter a pod without a proxy container.

jaeger/cmd/check.go
```diff
@@ -0,0 +1,232 @@
...
+
+		if !proxyContainer {
+			return fmt.Errorf("could not find proxy container for %s pod", pod.Name)
+		}
+	}
+
+	return nil
+}
+
+// checkPodsStatus checks if the pod is in running state
```
adleong, line 224: might want to update this comment to indicate that this checks that all of the pods are in the running state.  And perhaps change the name to describe what it does more specifically.  For example, "checkPodsRunning" or something.

pkg/k8s/api.go
```diff
@@ -155,6 +155,15 @@ func (kubeAPI *KubernetesAPI) GetPodsByNamespace(ctx context.Context, namespace
 	return podList.Items, nil
 }
 
+// GetPods returns all pods in a given namespace, with the label selector
+func (kubeAPI *KubernetesAPI) GetPods(ctx context.Context, namespace string, selector string) ([]corev1.Pod, error) {
```
adleong, line 159: Is calling

```
pods, err := kubeAPI.GetPods(ctx, ns, "foo=bar")
```

enough more ergonomic than

```
pods, err := kubeAPI.CoreV1().Pods(ns).List(ctx, metav1.ListOptions{LabelSelector: "foo=bar"})
```

to justify a layer of indirection?  I like the second one better because I can see directly what it does instead of having to look up the GetPods function to see what it does.

adleong (commented):
+1 to @alpeb's suggestions

jaeger/cmd/check.go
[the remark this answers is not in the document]
Pothulapati: > how we have things set up in pkg/healthcheck/healthcheck.go where you just instantiate the Checker struct. 

This is not possible here because those fields are not public, so not accessible here. Hence this helper function here. 🤔 To make them public would be a huge change. WDYT?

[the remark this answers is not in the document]
Pothulapati: > Shouldn't all these checks be fatal?

Currently, They are not and hence I did  the same. I think this was done so that you can get full view on whats happening in the extension. Should they fail instead? 🤔

pkg/k8s/api.go
Pothulapati: Makes sense. My reasoning was that it's a common function and could be used at multiple places,,

jaeger/cmd/check.go
Pothulapati: But we would be still needing the boolean flag to check for the proxy container right? I updated the fun to be different ones, as you suggested!

jaeger/cmd/check.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+	return nil
+}
+
+func containsProxy(pod *corev1.Pod) bool {
+	proxyContainer := false
+	for _, containerSpec := range pod.Spec.Containers {
+		if containerSpec.Name == k8s.ProxyContainerName {
+			proxyContainer = true
```
adleong, line 214: Ah, my motivation for refactoring it this way is that now you really can just `return true` here.

adleong: And then just `return false` at the end of the function if we haven't returned already.

jaeger/cmd/check.go
```diff
@@ -0,0 +1,228 @@
...
+	for _, pod := range pods {
+		pod := pod // pin
+		if !containsProxy(&pod) {
+			return fmt.Errorf("could not find proxy container for %s pod", pod.Name)
+		}
+	}
+	return nil
+}
+
+func containsProxy(pod *corev1.Pod) bool {
```
adleong, line 210: nit: if you make this take a `corev1.Pod` instead of a pointer here, you can avoid copying the pod in `checkIfDataPlanePodsExist`.

jaeger/cmd/check.go
Pothulapati: Yep, Correct! 👍

jaeger/cmd/check.go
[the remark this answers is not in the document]
alpeb: > > how we have things set up in pkg/healthcheck/healthcheck.go where you just instantiate the Checker struct.
> 
> This is not possible here because those fields are not public, so not accessible here. Hence this helper function here.  To make them public would be a huge change. WDYT?

I'm ok making those fields public now that they're used in other packages. The diff certainly will get bigger, but I think it's manageable given it's just string changes.

jaeger/cmd/check.go
```diff
@@ -0,0 +1,226 @@
...
+failure it will print additional information about the failure and exit with a
+non-zero exit code.`,
+		Example: `  # Check that the Jaeger extension is up and running
+  linkerd jaeger check`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// Get jaeger Extension Namespace
+			ns, err := getNamespaceOfExtension(jaegerExtensionName)
+			if err != nil {
+				fmt.Fprint(os.Stderr, err.Error())
```
adleong, line 130: I think this error string needs to have a newline at the end to render properly on the command line.

jaeger/cmd/check.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, ns := range namespaces.Items {
+		if ns.Labels[k8s.LinkerdExtensionLabel] == name {
+			return &ns, err
+		}
+	}
+	return nil, fmt.Errorf("Could not find the namespace for extension %s", name)
```
adleong, line 197: Let's give this error a call to action:

"Could not find the linkerd-jaeger extension. It can be installed by running `linkerd jaeger install | kubectl apply -f -`\n"

jaeger/cmd/check.go
```diff
@@ -0,0 +1,226 @@
...
+			WithCheck(func(ctx context.Context) error {
+				// Check for Jaeger Service Account
+				_, err = kubeAPI.CoreV1().ConfigMaps(jaegerNamespace).Get(ctx, "collector-config", metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				return nil
+			}))
+
+	checkers = append(checkers,
```
adleong, line 61: Should we add a retryDeadline to these checks which look for certain pods to be running so that they retry and we get the spinner behavior that the `linkerd check` command does?

jaeger/cmd/check.go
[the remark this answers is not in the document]
Pothulapati: Hmm, The Change is manageable, but it could break the API that these structs expose right? to be able to access them anywhere and that creates further expectations on future changes. I remember, there was some discussions around keeping private fields private wherever possible. 

@adleong WDYT?

jaeger/cmd/check.go
Pothulapati: Using `fmt.Fprintln` seems to help but this does not seem to be the case everywhere else.

jaeger/cmd/check.go
Pothulapati: Good Find! :)

adleong (approved):
Thanks for all the updates, @Pothulapati!  Great work!

jaeger/cmd/check.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+func newCmdCheck() *cobra.Command {
+	options := newCheckOptions()
+	cmd := &cobra.Command{
+		Use:   "check [flags]",
+		Args:  cobra.NoArgs,
+		Short: "Check the Jaeger extension for potential problems",
+		Long: `Check the jaeger extension for potential problems.
```
kleimkuhler, line 117:
```suggested
		Long: `Check the Jaeger extension for potential problems.
```

kleimkuhler (approved):
This looks good! Have some comments about spelling/formatting that should be fixed before merging, but the actual changes look good and work well.

PR #5203, opened 2020-11-10
cli: remove logs subcommand and tests

Pothulapati:
Fixes #5191

The logs command adds a external dependency that we forked to work but
does not fit within linkerd's core set of responsibilities. Hence, This
is being removed.

For capabilities like this, The Kubernetes plugin ecosystem has better
and well maintained tools that can be used.

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

PR #5202, opened 2020-11-10
cli: Remove get cmd and relevant tests

Pothulapati:
Fixes #5190

`linkerd get` is not used currently and works only for pods. This can be
removed instead as per the issue. This branch removes the command and
also the associated unit and integration tests.

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

PR #5302, opened 2020-12-01
Cherry-picks and release notes for stable-2.9.1

alpeb:
## stable-2.9.1

This stable release contains a number of proxy enhancements: better support for
high-traffic workloads, improved performance by eliminating unnecessary
endpoint resolutions for TCP traffic and properly tearing down serverside
connections when errors occur, and reduced memory consumption on proxies which
maintain many idle connections (such as Prometheus' proxy).

On the CLI and control plane sides, it relaxes checks on root and intermediate
certificates (following X509 best practices), and fixes two issues: one that
prevented installation of the control plane into a custom namespace and one
which failed to update endpoint information when a headless service was
modified. 

* Proxy:
  * Addressed some issues reported around clients seeing max-concurrency errors
    by increasing the default in-flight request limit to 100K pending requests
  * Reduced the default idle connection timeout to 5s for outbound clients and
    for inbound clients to reduce the proxy's memory footprint, especially on
      Prometheus instances
  * Fixed an issue where the proxy did not receive updated endpoint information
    when a headless service was modified
  * Added HTTP/2 keepalive PING frames
  * Removed logic to avoid redundant TCP endpoint resolution
  * Fixed an issue where serverside connections were not torn down when an error
    occurred

* CLI / Control Plane:
  * Fixed a CLI issue where the `linkerd-namespace` flag was not honored when
    passed to the `install` and `upgrade` commands
  * Updated `linkerd check` so that it doesn't attempt to validate the subject
    alternative name (SAN) on root and intermediate certificates. SANs for leaf
    certificates will continue to be validated
  * Fixed an issue in the destination service where endpoints always included a
    protocol hint, regardless of the controller label being present or not
  * No longer panic in rare cases when `linkerd-config` doesn't have an entry
    for `Global` configs (thanks @hodbn!)


alpeb:
I've pulled-in "Remove logs comparisons in integration tests" (#5223) to help with CI flakiness. As a side-effect, I had to pull-in as well "cli: Remove get cmd and relevant tests" (#5202) and "cli: remove logs subcommand and tests" (#5203), which should be fine.
The charts READMEs generation (#5175) is close to be merged, but I prefer to leave that out because as a side-effect it introduces some null values in the yamls, which should be further tested.

olix0r (approved):
Release notes look great. I'm fine merging this branch and then doing testing and potentially followup changes before tagging the release.

olix0r:
Would be good to pick up https://github.com/linkerd/linkerd2/commit/13c3aa9062bc0f698ad9ccc56eff878af8a83ace

alpeb:
I've just added here the latest proxy, and I also updated the images at `ghcr.io/alpeb`

kleimkuhler (approved):
Notes look good 👍

PR #5304, opened 2020-12-01
jaeger: Add support for override flags

Pothulapati:
This change adds flags `set`, `set-string`, `values`, `set-files`,
etc flags which are used to override the default values. This is
similar to that of Helm.

This also updates the install workflow to directly use Helm v3
pkg for chart loading and generation, without having to use
our chart type, etc.

### DIFF
--- a/jaeger/cmd/install.go
+++ b/jaeger/cmd/install.go
@@ -1,22 +1,26 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
 	"os"
+	"path"
 
 	"github.com/linkerd/linkerd2/jaeger/static"
-	jaeger "github.com/linkerd/linkerd2/jaeger/values"
 	"github.com/linkerd/linkerd2/pkg/charts"
+	partials "github.com/linkerd/linkerd2/pkg/charts/static"
+	"github.com/linkerd/linkerd2/pkg/flags"
 	"github.com/linkerd/linkerd2/pkg/healthcheck"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/spf13/cobra"
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli/values"
+	"helm.sh/helm/v3/pkg/engine"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/yaml"
 )
 
 var (
@@ -30,12 +34,7 @@ var (
 
 func newCmdInstall() *cobra.Command {
 	var skipChecks bool
-
-	values, err := jaeger.NewValues()
-	if err != nil {
-		fmt.Fprint(os.Stderr, err.Error())
-		os.Exit(1)
-	}
+	var options values.Options
 
 	cmd := &cobra.Command{
 		Use:   "install [flags]",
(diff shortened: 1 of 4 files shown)


jaeger/cmd/install.go
```diff
@@ -68,27 +67,28 @@ func newCmdInstall() *cobra.Command {
...
 
-	// TODO: Add --set flag set and also config
+	flags.AddValueOptionsFlags(cmd.Flags(), &options)
 
 	return cmd
 }
 
-func install(w io.Writer, values *jaeger.Values) error {
+func install(w io.Writer, options values.Options) error {
+	// Merge and create final set of values
```
adleong, line 76: I don't know if this comment is right.  Doesn't this just get the "overrides" yaml from the flags?  In other words, this doesn't include the default values from the chart at all.

adleong: For clarity, I'd consider naming this `valuesOverrides` or something along those lines.

jaeger/cmd/install.go
```diff
@@ -97,19 +97,52 @@ func render(w io.Writer, values *jaeger.Values) error {
...
+	vals, err := chartutil.CoalesceValues(chart, rawValues)
 	if err != nil {
 		return err
 	}
 
+	// Attach the final values into the `Values` field for rendering to work
+	finalValues := make(map[string]interface{})
+	finalValues["Values"] = vals
+
+	renderedTemplates, err := engine.Render(chart, finalValues)
```
adleong, line 132: instead of creating a named variable for the final values, I'd consider just passing a literal:

```
Render(chart, map[string]interface{} { "Values": vals })
```

adleong (commented):
This is nice!  I like it!

jaeger/cmd/install.go
```diff
@@ -68,27 +67,29 @@ func newCmdInstall() *cobra.Command {
...
-	// TODO: Add --set flag set and also config
+	flags.AddValueOptionsFlags(cmd.Flags(), &options)
 
 	return cmd
 }
 
-func install(w io.Writer, values *jaeger.Values) error {
+func install(w io.Writer, options values.Options) error {
+
+	// Create final values override
```
adleong, line 77: I'd omit the word "final" here.  This is confusing because I'm not sure what is "final" about the overrides.  (they still need to be applied on top of the defaults)

adleong (approved):

pkg/charts/charts.go
```diff
@@ -16,6 +16,29 @@ import (
 
 const versionPlaceholder = "linkerdVersionValue"
 
+var (
+	// L5dPartials is the list of templates in partials chart
+	// Keep this slice synced with the contents of /charts/partials
+	L5dPartials = []string{
```
kleimkuhler, line 22: Any specific reason we have this instead of something that lists the files under `charts/partials/templates/`? It just seems like a possibly "gotcha" for keeping this updated when we could have something that just lists files under this directory.

Pothulapati: Hmm, Interesting question. We have similar lists for **all charts**, hence I did the same here. 
I think the reason is that we may not want to be dependent on the FS for this, as it can be a real or Virtual, and in the Virtual FS case, I don't see relevant methods do that (or probably I missed something)

@alpeb WDYT?

alpeb: Yeah, `static.Templates` is an `http.FileSystem` that only inplements `Open()`. We could implement a smart `ls` function that in the case of a vfs, it would leverage the fact that in this case `static.Templates` is a function that returns a map whose keys are the file paths... I dunno, could be fun? (if so, certainly in a separate PR)

kleimkuhler: Ah okay I missed that we had this for all charts already. I thought it was just for partials, but I think if its a change we want a separate PR makes sense as well.

kleimkuhler (approved):

PR #5307, opened 2020-12-01
Allow Multicluster Service to be non LoadBalancer ServiceType

DaspawnW:
**Subject**

Allow Multicluster Service to be not LoadBalancer Service Type

**Problem**
In case a cluster is configured in such a way, that it doesn't allow LoadBalancer as ServiceType or the Cloud Policies are configured in such a way that they prohibit LoadBalancers its not possible to setup a multicluster link.

**Solution**
This MR allows to define a different ServiceType (e.g. NodePort) and set the addresses where this NodePort is accessible.


adleong (commented):
Thanks, @DaspawnW!

Adding a flag to controller the Service type of the gateway looks good to me.
However, I don't think we should add a flag for overriding the gateway address on the link command.  The whole purpose of the Link command is to fetch this address dynamically.  Instead, I'd recommend just manually creating the Link resource directly with the address that you want to use.

DaspawnW:
Hi @adleong,

thnx for the feedback. Maybe I can give here some insight why and how....

We have Kubernetes Cluster that are configured to disallow LoadBalancer type to prevent developers to simply copy stuff from the internet that maybe contains a Service type LoadBalancer. Therefore we have an Ingress Nginx that has NodePort and a LoadBalancer that points to the Ingress.

With the first part of my MR we would be able to now change the Type to NodePort. This for sure could be solved also with kustomize. The second point of the MR is as you mentioned about the Link. First of all, if you allow to change type away from NodePort to LoadBalancer its required to provide your own gatewayaddresses, otherwise you'll run into an exception, that the system can't access the loadbalancer ingress addresses. (https://github.com/linkerd/linkerd2/blob/main/cli/cmd/multicluster.go#L632)
The second point I see is that behind the link command of Linkerd CLI there is some magic (e.g. create Service Account + kubeconfig file as a secret with the SA secret) that is hard to be implemented manually. Maybe some manual instruction how to solve this is then the alternative.

adleong:
Thanks for the additional context, @DaspawnW, that's really helpful.  You're right that the `multicluster link` command does more than just create the Link resource itself.  With all that in mind, I think this change makes sense.

adleong (approved):

cli/cmd/multicluster.go
```diff
@@ -728,6 +739,7 @@ func newLinkCommand() *cobra.Command {
 	cmd.Flags().StringVar(&opts.logLevel, "log-level", opts.logLevel, "Log level for the Multicluster components")
 	cmd.Flags().StringVar(&opts.dockerRegistry, "registry", opts.dockerRegistry, "Docker registry to pull service mirror controller image from")
 	cmd.Flags().StringVarP(&opts.selector, "selector", "l", opts.selector, "Selector (label query) to filter which services in the target cluster to mirror")
+	cmd.Flags().StringVar(&opts.gatewayAddresses, "gateway-addresses", opts.gatewayAddresses, "If specified overwrites gateway addresses when gateway service is not type LoadBalancer comma separated list")
```
alpeb, line 742: Can you put "comma separated list" between parenthesis?

alpeb (commented):
Thanks @DaspawnW , looks good to me :+1: 
I've just got one tiny nit below.

cli/cmd/multicluster.go
DaspawnW: Hi @alpeb  thnx for the review :-)
I've updated the Merge Request

alpeb (approved):

alpeb:
Yesterday we pushed something to generate the Helm charts READMEs into `main`, which is why the static check in this PR is failing. Can I ask you one more thing? Just rebase against `main`,  run `bin/helm-docs` and finally push the multicluster README which should now have the entry you added into `values.yaml`. That way CI will be green again :-)

alpeb:
Whoops, you forgot to sign the commit :-)

DaspawnW:
oh boy, I think now its done :-D

alpeb:
Excellent, thanks!

PR #5311, opened 2020-12-02
Add linkerd.io/extension label

adleong:
The namespace that Linkerd extensions are installed into is configurable.  This can make it difficult to know which extensions are installed and where they are located.  We add a `linkerd.io/extension` namespace label to easily enumerate and locate Linkerd extensions.  This can be used, for example, to enable certain features only when certain extensions are installed.  All new Linkerd extensions should include this namespace label.


Pothulapati (approved):
LGTM

alpeb (approved):

PR #5333, opened 2020-12-04
Change script to work with k3d and install only CLI

kleimkuhler:
This changes the install-pr script to work with k3d.

Additionally, it now only installs the CLI; it no longer installs Linkerd on the
cluster. This was removed because most of the time when installing a Linkerd
version from a PR, some extra installation configuration is required and I was
always commenting out that final part of the script.

`--context` was changed to `--cluster` since we no longer need a context value,
only the cluster name which we are loading the images in to.

### DIFF
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -97,11 +97,9 @@ tag=$(echo "$image" | cut -f 2 -d ":")
 
 if [ $is_kind ] || [ $is_k3d ]
 then
-  distro=""
-  if [ $is_k3d ]
+  distro="k3d"
+  if [ $is_kind ]
   then
-    distro="k3d"
-  else
     distro="kind"
   fi
   "$bindir"/image-load --"$distro" --archive "$cluster"
--- a/bin/install-pr
+++ b/bin/install-pr
@@ -2,10 +2,12 @@
 
 ### Install PR ###
 #
-# This script takes a Github pull request number as an argument, downloads the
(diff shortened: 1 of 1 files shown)


adleong (approved):

bin/install-pr
```diff
@@ -87,15 +95,16 @@ echo "### Loading images into Docker ###"
...
 then
-  # KinD context strings are created by prepending `kind-` to the cluster name
-  if [ "$context" ]
+  distro=""
+  if [ $is_k3d ]
   then
-    name=$(echo "$context" | sed -n -E "s/(kind)-(.*)/\2/p")
+    distro="k3d"
+  else
+    distro="kind"
```
alpeb, line 105: tioli nit: You could do go-style here and save a couple of lines by initializing `distro="k3d"` and overriding only `if [$is_kind]`. Also this sets the default to `k3d` instead of `kind` which I believe is what most of us are using nowadays.

alpeb (approved):
LGTM :+1:  Just only one tioli/nit

bin/install-pr
kleimkuhler: Yep sounds good! `k3d` will be the "default" but we'll still need to pass the `--k3d` flag. The actual default of not passing anything is still just using an actual k8s cluster. I didn't plan to change that as part of this but maybe something to do if it seems necessary in the future.

PR #5335, opened 2020-12-04
bin/shellcheck-all was missing some files

alpeb:
`bin/shellcheck-all` identifies what files to check by filtering by the
`text/x-shellscript` mime-type, which only applies to files with a
shebang pointing to bash. We had a number of files with a
`#!/usr/bin/env sh` shebang that (at least in Ubuntu given `sh` points
to `dash`) only exposes a `text/plain` mime-type, thus they were not
being checked.

These files required just `sh` because they didn't rely on "advanced"
bash features, but it doesn't hurt having them use `bash`, IMO.

As a result I also fixed a few warnings in the files that were being
ignored.

(:taco: to @siggy for the find).

CC @joakimr-axis in case you wanna review :wink:

### DIFF
--- a/bin/helm-docs-diff
+++ b/bin/helm-docs-diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 set -eu
 
@@ -10,6 +10,6 @@ if [ -z "$dir_dirty" ]; then
   exit 0
 else
   echo "Helm-docs generated readmes diverge from current chart readmes:"
-  echo "$(git status)"
+  git status
   exit 64
 fi
--- a/bin/helm-docs-diff
+++ b/bin/helm-docs-diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 set -eu
 
@@ -10,6 +10,6 @@ if [ -z "$dir_dirty" ]; then
   exit 0
 else
   echo "Helm-docs generated readmes diverge from current chart readmes:"
-  git status
+  echo "$(git status)"
   exit 64
 fi
--- a/bin/helm-docs-diff
+++ b/bin/helm-docs-diff
@@ -6,10 +6,10 @@ bin/helm-docs
 
 dir_dirty=$(git diff HEAD)
 if [ -z "$dir_dirty" ]; then
-  echo "Helm-docs generated readmes match chart readmes."
+  echo 'Helm-docs generated readmes match chart readmes.'
   exit 0
 else
-  echo "Helm-docs generated readmes diverge from current chart readmes:"
-  echo "$(git status)"
+  echo 'Helm-docs generated readmes diverge from current chart readmes:'
+  git status
   exit 64
 fi
--- a/bin/protoc-diff
+++ b/bin/protoc-diff
@@ -13,10 +13,10 @@ bin/protoc-go.sh
 
 dir_dirty=$(git diff HEAD)
 if [ -z "$dir_dirty" ]; then
-  echo "Protobuf definitions match generated code"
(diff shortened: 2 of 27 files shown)


alpeb:
I'm making an exception after all with `bin/install-deps` and have it use `sh` because it's used in the Dockerfiles based on Alpine, which doesn't have bash, and installing it would need like 50MB extra.

adleong:
Using the mimetype seems a bit unreliable.  What about something a bit more straightforward:

```
-"$bindir"/shellcheck -x -P "$bindir" $(find "$rootdir" -type f \
-               ! -path "$rootdir"/.git/hooks/\*.sample \
-               | while read -r f; do [ "$(file -b --mime-type "$f")" = 'text/x-shellscript' ] && printf '%s\0' "$f"; done | xargs -0)
+"$bindir"/shellcheck -x -P "$bindir" $(find . -exec awk 'NR==1&&/^\#\!\/usr\/bin\/env (bash|sh)$/{print FILENAME}' {} \;)
```

This should catch everything without needing to change which shell we use.

joakimr-axis:
> These files required just `sh` because they didn't rely on "advanced"
> bash features, but it doesn't hurt having them use `bash`, IMO.

If you ask me (and you did :-) ), I would say that there is no need to use something with 10 times the footprint (bash vs. dash) if there is no good reason for it. And I really like @adleong's suggested approach to the problem at hand.

joakimr-axis:
> Using the mimetype seems a bit unreliable. What about something a bit more straightforward:
[...]

I do think we want to (still) avoid searching the .git directory and such. And involving `awk` could be avoided. To do the search-all-files-for-shebang approach (which I like), I reckon a simple (regexp) `grep` command will do:

`grep -rnsle '#!/usr/bin/env \(bash\|sh\)' *`

Combined with the shellcheck command we would then get

```
-"$bindir"/shellcheck -x -P "$bindir" $(find "$rootdir" -type f \
-               ! -path "$rootdir"/.git/hooks/\*.sample \
-               | while read -r f; do [ "$(file -b --mime-type "$f")" = 'text/x-shellscript' ] && printf '%s\0' "$f"; done | xargs -0)
+"$bindir"/shellcheck -x -P "$bindir" $(grep -rnsle '#!/usr/bin/env \(bash\|sh\)' "$rootdir"/* | xargs)
```

(If it were to be more portable, the regexp grep could be replaced with two grep commands, one for bash and one for sh.)

bin/protoc-diff
```diff
@@ -17,6 +17,6 @@ if [ -z "$dir_dirty" ]; then
   exit 0
 else
   echo "Protobuf definitions diverge from generated code:"
```
joakimr-axis, line 19: Single quotes are preferred here since nothing is to be expanded.

joakimr-axis: (also goes for line 16)

bin/helm-docs-diff
```diff
@@ -10,6 +10,6 @@ if [ -z "$dir_dirty" ]; then
   exit 0
 else
   echo "Helm-docs generated readmes diverge from current chart readmes:"
```
joakimr-axis, line 12: Single quotes are preferred here since nothing is to be expanded.

(also goes for line 9)

alpeb:
Excellent feedback! I've reverted my changes, replaced the command bin `bin/shellcheck-all` as suggested, and fixed the warnings in some scripts :+1:

kleimkuhler (approved):
Nice!

joakimr-axis:
Teamwork!!

adleong (approved):

adleong:
Thanks @joakimr-axis!  Your shell expertise is always appreciated!

PR #5339, opened 2020-12-07
extension: separate out Linkerd Viz extension

### LINKED ISSUE #5313: Split linkerd-viz components into their own chart
In order to modularize the viz related components into a linkerd-viz extension, we should create a new top-level directory in the Linkerd2 repo called viz, create a linkerd-viz chart there, and move all viz related components to that chart (such as Prometheus, grafana, the metrics-api, tap, the web dashboard, etc.)  This new chart should install these viz components into their own namespace.

The CLI should be updated to find these components in their new namespace.

To preserve the current behavior, the linkerd-viz component templates should remain pre-injected with the proxy and the `linkerd install` command should render both the core chart and the linkerd-viz chart.  In future work once a `linkerd viz install` command is available for installing the linkerd-viz chart separately, the pre-injection can be removed.

Pothulapati:
Fixes #5313 

This branch separates out the current Linkerd2 tempaltes into the
default and viz. As with this split, there is no need for `.global.`
as there are no add-ons now, that is being removed everywhere.

This moves some of the components into `viz` top level directory


Pothulapati:
This is too big to review, hence breaking this into
- Separate out components into Linkerd Viz.
- Remove `global` field from the core control-plane chart.

PR #5340, opened 2020-12-07
viz: move some components into linkerd-viz

Pothulapati:
This branch moves the grafana,prometheus,web, tap components
into a new viz chart, following the same extension model that
multi-cluster and jaeger follow.

The components in viz are not injected during install time, and
will go through the injector. The `viz install` does not have any
cli flags to customize the install directly but instead follow the Helm
way of customization by using flags such as 
`set`, `set-string`, `values`, `set-files`.

**Changes Include**
- Move `grafana`, `prometheus`, `web`, `tap` templates into viz extension.
- Remove all add-on related charts, logic and tests w.r.t CLI & Helm.
- Clean up `linkerd2/values.go` & `linkerd2/values.yaml` to not contain
 fields related to viz components.
- Update `linkerd check` Healthchecks to not check for viz components.
- Create a new top level `viz` directory with CLI logic and Helm charts.
- Clean fields in the `viz/Values.yaml` to be in the `<component>.<property>`
model. Ex: `prometheus.resources`, `dashboard.image.tag`, etc so that it is
consistent everywhere.

**Testing**

```bash

# Install the Core Linkerd Installation
./bin/linkerd install | k apply -f -

# Wait for the proxy-injector to be ready
# Install the Viz Extension
./bin/linkerd cli viz install | k apply -f -

# Customized Install
./bin/linkerd cli viz install --set prometheus.enabled=false | k apply -f -
```

What is not included in this PR:
- Move of Controller from core install into the viz extension.
- Simplification and refactoring of the core chart i.e removing `.global`, etc.


cli/cmd/install.go
```diff
@@ -64,26 +64,22 @@ var (
...
 		"templates/config.yaml",
 		"templates/identity.yaml",
-		"templates/controller.yaml",
 		"templates/destination.yaml",
 		"templates/heartbeat.yaml",
-		"templates/web.yaml",
 		"templates/proxy-injector.yaml",
 		"templates/sp-validator.yaml",
-		"templates/tap.yaml",
+		"templates/controller.yaml",
```
alpeb, line 82: Can you move it to its original location?

pkg/healthcheck/healthcheck.go
```diff
@@ -1027,121 +1024,6 @@ func (hc *HealthChecker) allCategories() []category {
...
-						}
-						if err != nil {
-							return err
-						}
-						return hc.checkCertAndAnchorsExpiringSoon(cert)
-
-					},
-				},
-				{
-					description: "proxy-injector webhook has valid cert",
```
alpeb, line 1071: This check and the ones below removed by mistake?

pkg/healthcheck/healthcheck.go
```diff
@@ -2421,7 +2281,7 @@ const running = "Running"
 func validateControlPlanePods(pods []corev1.Pod) error {
 	statuses := getPodStatuses(pods)
 
-	names := []string{"controller", "identity", "sp-validator", "web", "tap"}
```
alpeb, line 2424: Why was `controller` removed?

alpeb (commented):
Looks good on a first pass :+1: (I still need to check and test the CLI)

You can also move the `dashboard.replicas` setting in the core `values.yaml` to the viz' one. Also the `tap` section, and `webImage`, `webResources`, etc.

Pothulapati:
@alpeb Addressed all the comments! :)

charts/linkerd2/README.md
```diff
@@ -203,24 +198,19 @@ Kubernetes: `>=1.13.0-0`
 | identity.issuer.tls.crtPEM | string | `""` | Issuer certificate (ECDSA). It must be provided during install. |
 | identity.issuer.tls.keyPEM | string | `""` | Key for the issuer certificate (ECDSA). It must be provided during install |
 | installNamespace | bool | `true` | Set to false when installing Linkerd in a custom namespace. See the [Linkerd documentation](https://linkerd.io/2/tasks/install-helmcustomizing-the-namespace) for more information. |
+| linkerdVizNamespace | string | `"linkerd-viz"` | Namespace of the Linkerd Viz Extension |
```
adleong, line 201: I think we're only introducing this value temporarily since the controller isn't moving into the linkerd-viz extension in this PR.  I'd prefer to not introduce this and instead temporarily just hardcode the namespace used to discover the linkerd viz namespace.  We can remove this hardcoding when we move everything where it goes.

charts/linkerd2/templates/controller.yaml
```diff
@@ -81,8 +81,8 @@ spec:
         - -cluster-domain={{.Values.global.clusterDomain}}
         {{- if .Values.global.prometheusUrl }}
         - -prometheus-url={{.Values.global.prometheusUrl}}
-        {{- else if .Values.prometheus.enabled }}
-        - -prometheus-url=http://linkerd-prometheus.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:9090
+        {{- else }}
+        - -prometheus-url=http://linkerd-prometheus.{{.Values.linkerdVizNamespace}}.svc.{{.Values.global.clusterDomain}}:9090
```
adleong, line 85: We won't need this once the controller has moved into the linkerd-viz extension since it will be able to use the `prometheusAddr` value in that extension directly.  For now, I think we should just hardcode the namespace here.

charts/linkerd2/templates/heartbeat.yaml
```diff
@@ -48,8 +48,8 @@ spec:
             - "-log-level={{.Values.global.controllerLogLevel}}"
             {{- if .Values.global.prometheusUrl }}
             - "-prometheus-url={{.Values.global.prometheusUrl}}"
-            {{- else if .Values.prometheus.enabled }}
-            - "-prometheus-url=http://linkerd-prometheus.{{.Values.global.namespace}}.svc.{{.Values.global.clusterDomain}}:9090"
+            {{- else }}
+            - "-prometheus-url=http://linkerd-prometheus.{{.Values.linkerdVizNamespace}}.svc.{{.Values.global.clusterDomain}}:9090"
```
adleong, line 52: In a future PR, let's change this to discover the linkerd-viz namespace dynamically in the heartbeat code by using the linkerd.io/extension label instead of passing it in as a flag.  For now, let's just hardcode the namespace.

charts/partials/templates/_proxy.tpl
```diff
@@ -95,7 +95,7 @@ env:
   value: "true"
 {{ else if not .Values.global.proxy.disableIdentity -}}
 - name: LINKERD2_PROXY_TAP_SVC_NAME
-  value: linkerd-tap.$(_l5d_ns).serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
+  value: linkerd-tap.{{.Values.linkerdVizNamespace}}.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
```
adleong, line 98: This logic will eventually move to the tap-injector in the linkerd-viz extension.  For now I'd say just hardcode the namespace.

cli/cmd/dashboard.go
```diff
@@ -94,7 +94,7 @@ func newCmdDashboard() *cobra.Command {
 			portforward, err := k8s.NewPortForward(
 				cmd.Context(),
 				k8sAPI,
-				controlPlaneNamespace,
+				linkerdVizNamespace,
```
adleong, line 97: This command will eventually move to the `linkerd viz` subcommand.  Furthermore, we can dynamically detect the namespace using the linkerd.io/extensions label.  We should use that label (or, we can do that in a future PR and just hardcode it here for now)

cli/cmd/root.go
```diff
@@ -12,17 +12,19 @@ import (
...
 )
 
 const (
-	defaultLinkerdNamespace = "linkerd"
-	defaultCNINamespace     = "linkerd-cni"
-	defaultClusterDomain    = "cluster.local"
-	defaultDockerRegistry   = "ghcr.io/linkerd"
+	defaultLinkerdNamespace    = "linkerd"
+	defaultCNINamespace        = "linkerd-cni"
+	defaultLinkerdVizNamespace = "linkerd-viz"
```
adleong, line 25: I don't think we need this.

viz/charts/linkerd-viz/templates/namespace.yaml
```diff
@@ -0,0 +1,11 @@
+---
+###
+### Linkerd Viz Extension Namespace
+###
+---
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: {{ .Values.namespace }}
+  annotations:
```
adleong, line 10: Let's add the linkerd.io/extension label.

viz/charts/linkerd-viz/templates/psp.yaml
```diff
@@ -0,0 +1,32 @@
...
+  namespace: {{.Values.namespace}}
+  labels:
+    namespace: {{.Values.namespace}}
+roleRef:
+  kind: Role
+  name: linkerd-psp
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: linkerd-controller
```
adleong, line 15: I thought the linkerd-controller wasn't moving in this PR?

adleong (commented):
This looks pretty good so far!  We should make sure to enumerate all the follow up work that needs to happen after this merges:
* move controller to linkerd-viz extension
* move web, tap, and controller code to `viz` directory

adleong:
After installing the core control plane I get a check failure because prometheus isn't reachable:

```
    Error calling Prometheus from the control plane
```

We should remove this check with the intention of moving it to `linkerd viz check`.

adleong:
I noticed that the linkerd-web Role and RoleBinding still get installed into the `linkerd` namespace.  Is this intentional?

Pothulapati:
> After installing the core control plane I get a check failure because prometheus isn't reachable:
> 
> ```
> × [prometheus] control plane can talk to Prometheus
>     Error calling Prometheus from the control plane
> ```
> 
> We should remove this check with the intention of moving it to `linkerd viz check`.

This is a actually a self-check in the controller, and will be moved to `viz check` once the controller is moved into viz.

alpeb:
Following up on the discussion we started offline about the need to increase RBAC access for some components after this split...

Assuming the simplest mechanism to discover installed extensions is to just list namespaces and read their `linkerd.io/extension`label, that's fine when done from the public-api which already has that RBAC. That will work e.g. for discovering Prometheus' location. 

In other cases there isn't yet enough RBAC, like:
- HeartBeat trying to use Prometheus
- Web wanting to know if there's a Multicluster component to show its section
- Web wanting to know Jaegger's namespace in order to link back to jaegger in the dashboard

Given that linkerd's (core) namespace is provided to extensions as a parameter when installed (defaulting to `linkerd`), and that we wouldn't want to have to do the opposite (providing extension namespaces to linkerd core say through a `linkerd upgrade`), WDYT about exposing an API like `GetExtensions` in public-api returning all that info?

Also, Web currently requires `linkerd-config` access just to read its UUID. We could provide that as an API as well.

adleong:
We certainly could create a core API which returns a list of extensions, but I think this is probably over-complicating things.  Adding RBAC to read namespaces to the components that need to discover other extensions seems pretty harmless.

I assume web needs the UUID so that it can include it in version check requests.  Given the existence of the heartbeat component, it may or may not be important to have that data here.  If the UUID were passed in as a helm value, then `linkerd viz install` could fetch it from the linkerd-config and populate it correctly.  However, this value would be absent for Helm installs of the viz extension.

adleong:
For the UUID issue specifically, rather than create a brand new API, a nice solution might be to simply add the UUID to the response of the core-api's `Version` method.  This makes that method a little bit more useful and avoids needing to do any cross-namespace RBAC weirdness in order for the Web controller to get the UUID.

test/integration/smimetrics/smi-metrics_test.go
```diff
@@ -28,6 +28,8 @@ func TestMain(m *testing.M) {
 
 func TestSMIMetrics(t *testing.T) {
 
+	t.Skip("Skipped, as SMI-Metrics currently hardcodes the prometheusUrl of Linkerd which changed")
```
adleong, line 31: Ah nice catch!  We'll need to submit an update to SMI-Metrics to use the Linkerd-viz extension.

viz/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// These regexs are not as strict as they could be, but are a quick and dirty
+	// sanity check against illegal characters.
+	alphaNumDash = regexp.MustCompile(`^[a-zA-Z0-9-]+$`)
+)
+
+// NewCmdViz returns a new jeager command
+func NewCmdViz() *cobra.Command {
+	jaegerCmd := &cobra.Command{
+		Use:   "viz",
```
alpeb, line 34: still refering to jaeger

viz/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+			if !alphaNumDash.MatchString(controlPlaneNamespace) {
+				return fmt.Errorf("%s is not a valid namespace", controlPlaneNamespace)
+			}
+
+			return nil
+		},
+	}
+
+	jaegerCmd.PersistentFlags().StringVarP(&controlPlaneNamespace, "linkerd-namespace", "L", defaultLinkerdNamespace, "Namespace in which Linkerd is installed")
+	jaegerCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", defaultVizNamespace, "Namespace in which Jaeger extension is installed")
```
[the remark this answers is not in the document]
alpeb: As a matter of fact, do we have to declare all these root commands here again given they're already declared in the parent command's root file?

viz/static/templates.go
```diff
@@ -0,0 +1,15 @@
...
+package static
+
+import (
+	"net/http"
+	"path"
+
+	"github.com/linkerd/linkerd2/pkg/charts/static"
+)
+
+// Templates that will be rendered by `linkerd install`. This is only used on
```
alpeb, line 13: `linkerd viz install`

viz/cmd/install.go
```diff
@@ -0,0 +1,180 @@
...
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli/values"
+	"helm.sh/helm/v3/pkg/engine"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	templaesVIz = []string{
```
alpeb, line 27: typo

viz/cmd/install.go
```diff
@@ -0,0 +1,180 @@
...
+		if _, err := buf.WriteString(renderedTemplates[t]); err != nil {
+			return err
+		}
+	}
+
+	_, err = w.Write(buf.Bytes())
+	return err
+}
+
+func checkIfLinkerdExists(ctx context.Context) (bool, error) {
```
alpeb, line 157: Can you move this function into a library, and reuse it in jaeger?

viz/cmd/install.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	var partialFiles []*loader.BufferedFile
+	for _, template := range charts.L5dPartials {
+		partialFiles = append(partialFiles,
+			&loader.BufferedFile{Name: template},
+		)
+	}
+
+	// Load all jaeger chart files into buffer
```
alpeb, line 112: jaeger

viz/cmd/root.go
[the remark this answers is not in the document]
Pothulapati: Hmm, Interesting. I initially declared them because we were going with the separate command way of doing it, but now its just a subcommand.

They are currently not accessible because they are not public, and these are in different pkgs. Should we try doing that or keep them separate? 🤔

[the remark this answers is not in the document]
adleong: Per my suggestions above, I recommend getting rid of the `linkerd-viz-namespace` flag in the main control plane altogether.

alpeb:
What do you guys think about adding the `linkerd.io/extension` label not just to the extension namespaces but to every other resource as well?  During upgrades we need to have a common label we can pass to `--prune`:
```bash
$ linkerd viz install | kubectl apply -f --prune -l  linkerd.io/extension=linkerd-viz
```

adleong:
@alpeb I had not added the label to other resources because there wasn't really any reason to.  But the prune use-case makes sense.  I'm in favor of adding the label to all of the resources.

viz/charts/linkerd-viz/templates/prometheus.yaml
```diff
@@ -70,7 +73,7 @@ data:
       kubernetes_sd_configs:
       - role: pod
         namespaces:
-          names: ['{{.Values.global.namespace}}']
+          names: ['{{.Values.linkerdNamespace}}']
       relabel_configs:
       - source_labels:
         - __meta_kubernetes_pod_label_linkerd_io_control_plane_component
```
alpeb, line 79: Note that this scrape config wil now just scrape the core components. We have a separate section for Grafana. Worth adding separate sections for Viz and Jaeger?

viz/cmd/root.go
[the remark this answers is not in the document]
Pothulapati: Removed `linkerdVizNamespace` from the main control-plane and its helm chart.

Pothulapati:
Yep, Adding the extension label makes sense. So, Added it.

viz/charts/linkerd-viz/templates/prometheus.yaml
Pothulapati: Hmm, for grafana we already had specific metrics in mind that we felt users would find useful but I don't think we have that kind of knowledge yet with Jaeger. We can definitely replicate the same for other linkerd components in viz.

Pothulapati: I'm a bit against on having these jobs for each induvidual extension/component, etc. Instead can we have a job that can cover all linkerd components irrespective of the extension. Something like

```
    - job_name: 'linkerd-components'
      kubernetes_sd_configs:
      - role: pod
      relabel_configs:
      - source_labels:
        - __meta_kubernetes_pod_label_linkerd_io_extension
        - __meta_kubernetes_pod_label_linkerd_io_control_plane_component
        - __meta_kubernetes_pod_container_port_name
        action: keep
        regex: (.*);(.*);admin-http$
      - source_labels: [__meta_kubernetes_pod_container_name]
        action: replace
        target_label: component
```

alpeb: Ok I've created #5429 to track that effort in a separate PR, alongside with the required change in the Grafana chart.

charts/partials/templates/_pull-secrets.tpl
```diff
@@ -1,6 +1,4 @@
-{{- define "partials.image-pull-secrets" -}}
-{{- if .Values.global.imagePullSecrets }}
+{{- define "partials.image-pull-secrets"}}
 imagePullSecrets:
-{{ toYaml .Values.global.imagePullSecrets | indent 2 }}
-{{- end }}
+{{ toYaml . | indent 2 }}
 {{- end -}}
```
alpeb, line 4: Can you leave the boilerplate in the partial like it used to be, to avoid repeating it on each calling side?

alpeb:
@Pothulapati Can you also please take care of removing from the main `values-ha.yaml` the entries that no longer are relevant?

alpeb:
Seems like `RestrictDashboardPrivileges` in `values.go` can be removed :eye:

charts/partials/templates/_pull-secrets.tpl
Pothulapati: Done!

adleong (approved):
This looks good enough to merge to me.  Let's roll forward and keep fixing stuff on main.

test/integration/install_test.go
```diff
@@ -32,30 +32,32 @@ var (
...
+		{Namespace: "linkerd", Name: "linkerd-identity"},
+		{Namespace: "linkerd-viz", Name: "linkerd-prometheus"},
+		{Namespace: "linkerd-viz", Name: "linkerd-web"},
+		{Namespace: "linkerd-viz", Name: "linkerd-tap"},
+		{Namespace: "linkerd", Name: "linkerd-dst-headless"},
+		{Namespace: "linkerd", Name: "linkerd-identity-headless"},
+	}
+
+	multiclusterSvcs = []testutil.Service{
+		{Namespace: "linkerd-multicluster", Name: "linkerd-gateway"},
```
alpeb, line 60: :+1:

alpeb (approved):
Awesome, let's get this merged! :1st_place_medal: :ship: :shipit: :guitar:

PR #5347, opened 2020-12-08
bin/protoc-diff: Don't assume Debian and don't install unzip

### LINKED ISSUE #5337: bin/protoc-diff assumes Debian(-based) system
### What is the issue?
`bin/protoc-diff` checks if the host is a Linux system. If `unzip` is missing, it calls `apt-get`. This will fail for a non-Debian-based system.

### How can it be reproduced?
```shell
docker run -it --rm -v <path_to_your_linkerd2_dir>:/src/linkerd2 -w /src/linkerd2 fedora
bin/protoc-diff
```

### Logs, error output, etc
```shell
[root@f667739e8564 linkerd2]# bin/protoc-diff 
bin/protoc-diff: line 8: apt-get: command not found
[root@f667739e8564 linkerd2]#
```
#### `linkerd check` output
_Not applicable_

### Environment

- Kubernetes Version: _Not applicable_
- Cluster Environment: _Not applicable_
- Host OS: Debian Linux (bullseye)
- Linkerd version: commit 72a0ca974df2e623a3ffbff2a03abaabaab19322

### Possible solution
1. Should the script really install unzip on its host? If so:
    1. Is it supposed to be run as root? Otherwise it will fail due to permission needed for `apt-get`.
    2. A distinction is needed for the Linux breed. For Debian/Ubuntu etc. apt-get should be used. For Fedora/RedHat etc. yum. For alpine apk. Etc. etc.
2. Could it be simpler just to check for unzip and return an error message to the user if it is missing? And then let the user install it in a suitable way? Then `bin/protoc-diff` can be run as any appropriate non-root user and focus solely on its primary tasks.

### Additional context

Should really `bin/protoc-diff` do the/a `unzip` check/install at all? It does not use `unzip`, but calls `bin/protoc-go.sh` which in its turn calls `bin/protoc`, and `bin/protoc` uses `unzip`. IMO it would be more logical to leave such check/install functionality to `bin/protoc` who is the actual caller of `unzip`.

joakimr-axis:
- Do unzip check but don't install; leave installation to user
- Move unzip check to bin/protoc that actually uses unzip
- Make sure the protoc scripts can be called from any directory

Fixes #5337

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -39,6 +39,8 @@ jobs:
     container:
       image: golang:1.13.4
     steps:
+      - name: Prerequisites
+        run: apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install unzip
       - name: Checkout code
         # actions/checkout@v2
         uses: actions/checkout@722adc6
--- a/bin/protoc
+++ b/bin/protoc
@@ -15,7 +15,16 @@ protocversion=3.6.0
 protocbin=$targetbin/protoc-$protocversion
 protocurl=https://github.com/google/protobuf/releases/download/v$protocversion/protoc-$protocversion-$os-$arch.zip
 
+require_from() {
+    if ! command -v "$1" >/dev/null 2>/dev/null ; then
+        echo "Please acquire $1 from $2" >&2
+        return 1
+    fi
+}
+
 if [ ! -f "$protocbin" ]; then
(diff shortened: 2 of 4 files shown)


Pothulapati (approved):
LGTM! Thanks for working on this!

Leaving further reviews to @alpeb, who is our CI Guru! 👍

olix0r (approved):
thank you!

alpeb (approved):
Makes sense, thanks again @joakimr-axis :medal_sports:

PR #5349, opened 2020-12-08
upgrades: make webhooks restart if TLS creds are updated

### LINKED ISSUE #5231: Create jaeger-injector mutating webhook
Create a muting admission controller webhook called `jaeger-injector` which adds the appropriate tracing environment variables to the Linkerd proxy container.  

This webhook adds the LINKERD2_PROXY_TRACE_COLLECTOR_SVC_ADDR, LINKERD2_PROXY_TRACE_COLLECTOR_SVC_NAME, and LINKERD2_PROXY_TRACE_ATTRIBUTES_PATH environment variables as well as the podinfo volume to the Linkerd proxy container if it exists.  If a pod does not have a Linkerd proxy sidecar, no action should be taken.  However, this webhook should have a `IfNeeded` reinvokation policy so that it will be reinvoked if the Linkerd proxy is added.

By default, this injector should set the above env variable to the address and service account of the `oc-collector` in the linkerd-jaeger namespace.  However, if the `config.linkerd.io/trace-collector` or `config.alpha.linkerd.io/trace-collector-service-account` annotations are set on the pod or its namespace, it should use those values instead.

By default, this injector should be invoked on all newly created pods and act on them if they have a Linkerd proxy sidecar.  However, it should also be possible to configure the injector with a label selector to selectively opt in which pods will be affected.

### LINKED ISSUE #5341: `linkerd upgrade` used for changing settings (not for version upgrades) puts proxy injector in invalid state
### What is the issue?

Issuing `linkerd upgrade` CLI to update a setting, e.g. `linkerd upgrade --enable-endpoint-slices`, but not for a full version upgrade, results in the injector webhook secret being regenerated, but the injector pod doesn't get rolled-out and thus it starts failing with certificate errors and pods are no longer injected.

### How can it be reproduced?

- Install linkerd 2.9.0
- `linkerd upgrade --enable-endpoints-slices | kubectl apply -f -`
- `kubectl logs -f deploy/linkerd-proxy-injector proxy-injector`
- `linkerd inject https://run.linkerd.io/emojivoto.yml | kubectl apply -f -`

The emojivoto pods won't be injected and you'll see the following errors in the logs.

### Logs, error output, etc

```
2020/12/04 17:33:43 http: TLS handshake error from 127.0.0.1:57128: remote error: tls: bad certificate
```

#### `linkerd check` output

The check output is normal.

### Environment

- Kubernetes Version: 1.18
- Host OS: Linux
- Linkerd version: This only applies to Linkerd version 2.9

### Possible solution

This is fixed in `main` with #5282 that introduces dynamic cert reloading in the http server backing the injector. We could either back-port that into 2.9, or add some logic to avoid the CLI regenerating the secret.

### Additional context

This was found while testing #5299

Pothulapati:
Fixes #5341

Currently, we do not re-use the TLS certs during upgrades, which
means that the secrets are updated while the webhooks are still
paired with the older ones, causing the webhook requests to fail.

This can be solved by making webhooks be restarted whenever there
is a change in the certs. This can be performed by storing the hash
of the `*-rbac` file, which contains the secrets, thus making the
pod templates change whenever there is an update to the certs thus
making restarts required.

This follows the same pattern that we used with Helm


adleong (commented):
How does this work?  Isn't `proxy-injector-rbac.yaml` just the template file?  i.e. I wouldn't expect the contents of that file to change when only the Values change.

adleong:
Ah, I see, the `include` function renders the sub-template.  Pretty cool :+1:

Looks like there are some merge conflicts and integration failures to take care of.  But this approach looks good to me.

charts/linkerd2/templates/proxy-injector.yaml
```diff
@@ -33,9 +33,11 @@ spec:
   template:
     metadata:
       annotations:
-        {{- if empty .Values.global.cliVersion }}
+        {{ if kindIs "string" .Values.stage -}}
+        {{ if not (eq .Values.stage "control-plane") -}}
```
alpeb, line 37: Given `.Values.stage` won't be present for Helm installs, `checksum/config` won't be produced in that case and will break the mechanism for rolling-out the pod :-(

```bash
$ helm install linkerd -f ~/k8s/helm/linkerd-overrides.yaml charts/linkerd2
$ helm upgrade linkerd -f ~/k8s/helm/linkerd-overrides.yaml --set global.enableEndpointSlices=true charts/linkerd2
```
The last command will produce a restart only for the destination pod.
And these are the logs for the injector:
```bash
$ k -n linkerd logs -f linkerd-proxy-injector-7896f8d647-524nc proxy-injector
time="2020-12-09T22:02:34Z" level=info msg="running version dev-f7e2c84d-alpeb"
time="2020-12-09T22:02:34Z" level=info msg="waiting for caches to sync"
time="2020-12-09T22:02:35Z" level=info msg="caches synced"
time="2020-12-09T22:02:35Z" level=info msg="listening at :8443"
time="2020-12-09T22:02:35Z" level=info msg="starting admin server on :9995"
2020/12/09 22:04:41 http: TLS handshake error from 127.0.0.1:47788: remote error: tls: bad certificate
time="2020-12-09T22:04:47Z" level=info msg="Updated certificate"
```
It ends up working at the end only because of the dynamic cert reloading.

alpeb: Bringing back `{{- if empty .Values.global.cliVersion }}` in addition to the current conditions could do the trick, for always producing the checksum for Helm installs.

Pothulapati: Okay. That seems to be a problem. How about we set `stage:""` in `values.yaml` so that this is added even in helms way.
Seems to work.

```bash
tarun in dev in on  kind-kind () linkerd2 on  tarun/webhooks-upgrade [$!?] via 🐹 v1.15.4
❯ ./bin/helm upgrade linkerd2-1607576483   charts/linkerd2 \
  --set global.enableEndpointSlices=true \
  --set identity.issuer.crtExpiry=$exp
Release "linkerd2-1607576483" has been upgraded. Happy Helming!
NAME: linkerd2-1607576483
LAST DEPLOYED: Thu Dec 10 05:04:05 2020
NAMESPACE: default
STATUS: deployed
REVISION: 2
TEST SUITE: None
NOTES:
The Linkerd control plane was successfully installed 🎉

To help you manage your Linkerd service mesh you can install the Linkerd CLI by running:

  curl -sL https://run.linkerd.io/install | sh

Alternatively, you can download the CLI directly via the Linkerd releases page:

  https://github.com/linkerd/linkerd2/releases/

Looking for more? Visit https://linkerd.io/2/getting-started/

tarun in dev in on  kind-kind () linkerd2 on  tarun/webhooks-upgrade [$!?] via 🐹 v1.15.4 took 6s
❯

tarun in dev in on  kind-kind () linkerd2 on  tarun/webhooks-upgrade [$!?] via 🐹 v1.15.4 took 6s
❯ kgp -A
NAMESPACE            NAME                                         READY   STATUS            RESTARTS   AGE
kube-system          coredns-f9fd979d6-g7j6t                      1/1     Running           0          7m42s
kube-system          coredns-f9fd979d6-tkv76                      1/1     Running           0          7m42s
kube-system          etcd-kind-control-plane                      1/1     Running           0          7m52s
kube-system          kindnet-pwz9z                                1/1     Running           0          7m42s
kube-system          kube-apiserver-kind-control-plane            1/1     Running           0          7m52s
kube-system          kube-controller-manager-kind-control-plane   1/1     Running           0          7m52s
kube-system          kube-proxy-vl5mk                             1/1     Running           0          7m42s
kube-system          kube-scheduler-kind-control-plane            1/1     Running           0          7m52s
linkerd              linkerd-controller-99d7c455f-6w2dr           2/2     Running           0          2m41s
linkerd              linkerd-destination-545db4cdbb-ndz8b         0/2     PodInitializing   0          3s
linkerd              linkerd-destination-f669c68d9-zlnbq          2/2     Running           0          2m40s
linkerd              linkerd-grafana-59f85f5795-lnrsm             2/2     Running           0          2m41s
linkerd              linkerd-identity-7d5f6d797b-swcbd            2/2     Running           0          2m41s
linkerd              linkerd-prometheus-5669f85bf-xnj57           2/2     Running           0          2m40s
linkerd              linkerd-proxy-injector-5db757445b-rcvcz      2/2     Running           0          2m41s
linkerd              linkerd-proxy-injector-d85d5b5f7-vlfct       0/2     PodInitializing   0          3s
linkerd              linkerd-sp-validator-79c54c76c5-2jj6w        0/2     Init:0/1          0          3s
linkerd              linkerd-sp-validator-8547f97d45-dsbhf        2/2     Running           0          2m41s
linkerd              linkerd-tap-59bd44bc6c-fw7w4                 2/2     Running           0          2m40s
linkerd              linkerd-tap-78956c47dc-hjlcl                 0/2     Init:0/1          0          2s
linkerd              linkerd-web-65cd6786b8-2qbrg                 2/2     Running           0          2m41s
local-path-storage   local-path-provisioner-78776bfc44-jhn48      1/1     Running           0          7m42s
```

Pothulapati:
After discussing with @alpeb, We came to a conclusion to use the following condition

```
(or (empty .Values.global.cliVersion) (not (eq (.Values.stage | toString) "control-plane")
```

- `(empty .Values.global.cliVersion)` implies the Helm case, and this was present previously already.
- `(not (eq (.Values.stage | toString) "control-plane")` implies to skip adding `checksum` annotation for the `upgrade control-plane` cli sub-command, as `*-rbac.yaml` files are not available in the multi-staged install code flow.

alpeb (approved):

kleimkuhler (approved):
This looks good to me. Makes sense about what we check for when adding the annotation.

PR #5351, opened 2020-12-08
Rename proxy-mutator to jaeger-injector

adleong:
The name `proxy-mutator` is too generic.  In particular, several different linkerd extensions will have mutating webhooks which mutate the proxy sidecar, the MutatingWebhookConfiguration resource is cluster scoped, and each one needs a unique name.

We use the `jaeger-injector` name instead.  This gives us a pattern to follow for future webhooks as well (e.g. `tap-injector` etc.)


alpeb (commented):
I guess it would make sense as well to rename `jaeger/proxy-mutator` to `jaeger/injector`

testutil/kubernetes_helper.go
```diff
@@ -211,33 +211,31 @@ func (h *KubernetesHelper) CheckPods(ctx context.Context, namespace string, depl
...
 		checkedPods = []corev1.Pod{}
-		pods, err := h.clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
+		pods, err := h.GetPodsForDeployment(ctx, namespace, deploymentName)
 		if err != nil {
 			return err
 		}
 
 		var deploymentReplicas int
-		for _, pod := range pods.Items {
-			if strings.HasPrefix(pod.Name, deploymentName) {
```
adleong, line 221: Associating pods with their deployments by comparing prefixes was causing problems because the `jaeger` and `jaeger-collector` deployments share a prefix.  I changed this to reuse the GetPodsForDeployments method which relies on the deployment's label selector instead.

testutil/kubernetes_helper.go
```diff
@@ -211,33 +211,31 @@ func (h *KubernetesHelper) CheckPods(ctx context.Context, namespace string, depl
...
+					return fmt.Errorf("Container [%s] in pod [%s] in namespace [%s] is not running",
+						container.Name, pod.Name, pod.Namespace)
 				}
 			}
 		}
 
 		if deploymentReplicas != replicas {
-			return fmt.Errorf("Expected deployment [%s] in namespace [%s] to have [%d] running pods, but found [%d]",
-				deploymentName, namespace, replicas, deploymentReplicas)
+			return fmt.Errorf("Expected there to be [%d] pods in deployment [%s] in namespace [%s], but found [%d]",
```
adleong, line 237: This was the exact same error message as the one in CheckDeployment which really confused me because when I saw this message, I thought it was CheckDeployment that was failing.  I slightly re-worded this error string so the two can be distinguished.

jaeger/charts/jaeger/templates/rbac.yaml
```diff
@@ -27,7 +27,7 @@ metadata:
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: linkerd-jaeger-{{.Values.namespace}}-proxy-mutator
+  name: linkerd-jaeger-injector
 rules:
 - apiGroups: [""]
   resources: ["namespaces"]
```
alpeb, line 33: `s/Proxy Mutator RBAC/Jaeger Injector RBAC/` in the comment for this resource

alpeb (approved):
Thanks, LGTM :+1: 
I just found one place that still refers to the old nomenclature

kleimkuhler (approved):
👍

pkg/k8s/portforward.go
```diff
@@ -95,6 +99,22 @@ func NewPortForward(
 	return newPortForward(k8sAPI, namespace, podName, host, localPort, remotePort, emitLogs)
 }
 
+func getDeploymentForPod(ctx context.Context, k8sAPI *KubernetesAPI, pod corev1.Pod) (string, error) {
+	parents := pod.GetOwnerReferences()
+	if len(parents) != 1 {
+		return "", nil
```
alpeb, line 105: Better return an error with a message perhaps?

alpeb (approved):

PR #5353, opened 2020-12-09
Add jaeger uninstall command

adleong:
Add a `linkerd jaeger uninstall` command which prints the linkerd-jaeger extension resources so that they can be deleted.  This is similar to the `linkerd uninstall` command.

```
> bin/linkerd jaeger uninstall | k delete -f -
clusterrole.rbac.authorization.k8s.io "linkerd-jaeger-linkerd-jaeger-proxy-mutator" deleted
clusterrolebinding.rbac.authorization.k8s.io "linkerd-jaeger-linkerd-jaeger-proxy-mutator" deleted
mutatingwebhookconfiguration.admissionregistration.k8s.io "linkerd-proxy-mutator-webhook-config" deleted
namespace "linkerd-jaeger" deleted
```


jaeger/cmd/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	yamlSep = "---\n"
+)
+
+func newCmdUninstall() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "uninstall",
+		Args:  cobra.NoArgs,
+		Short: "Output Kubernetes resources to uninstall the Linkerd-jaeger extenion",
```
kleimkuhler, line 22:
```suggestion
		Short: "Output Kubernetes resources to uninstall the Linkerd-jaeger extension",
```

jaeger/cmd/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+		Long: `Output Kubernetes resources to uninstall the Linkerd-jaeger extenion.
```
kleimkuhler, line 23:
```suggestion
		Long: `Output Kubernetes resources to uninstall the Linkerd-jaeger extension.
```

jaeger/cmd/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+
+This command provides all Kubernetes namespace-scoped and cluster-scoped resources (e.g services, deployments, RBACs, etc.) necessary to uninstall the Linkerd-jaeger extenion.`,
```
kleimkuhler, line 25: I don't see namespace-scoped resources in the `uninstall` output. I only see the cluster-scoped resources and the `linkerd-jaeger` namespace resource. None of the services or deployments are rendered.

Based off `FetchKubernetesResources` though I wouldn't expect to see this resources. Is that expected or should that function be changed to render those resources as well?

jaeger/cmd/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+		Example: ` linkerd uninstall | kubectl delete -f -`,
```
kleimkuhler, line 26:
```suggestion
		Example: `linkerd uninstall | kubectl delete -f -`,
```

jaeger/cmd/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+	resources, err := resource.FetchKubernetesResources(ctx, k8sAPI,
+		metav1.ListOptions{LabelSelector: "linkerd.io/extension=linkerd-jaeger"},
+	)
+	if err != nil {
+		return err
+	}
+
+	for _, r := range resources {
+		if err := r.RenderResource(os.Stdout); err != nil {
+			return fmt.Errorf("error rendering Kubernetes resource:%v", err)
```
kleimkuhler, line 50:
```suggestion
			return fmt.Errorf("error rendering Kubernetes resource: %v", err)
```

kleimkuhler (commented):
My main comment is the question about what resources we expect to actually be rendered in the `uninstall` output. Right now I'm just seeing cluster-scoped resources.

kleimkuhler (approved):
Looks good!

jaeger/charts/jaeger/templates/rbac.yaml
```diff
@@ -37,6 +39,8 @@ kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: linkerd-jaeger-{{.Values.namespace}}-proxy-mutator
+  labels:
+    linkerd.io/extension: linkerd-jaeger
```
Pothulapati, line 43: I see that these are added for the cluster-wide resources. Should we also have these labels for deployments too? so that they are all easier to discover and perform these kinds of operations 🤔

kleimkuhler: Yea I think it's something good to keep in mind; it definitely wouldn't hurt at least. If this is the only operation where we care about this for now though, then having the namespace to delete is enough.

adleong: ☝️

alpeb (approved):
LGTM modulo naming nit :+1:

PR #5355, opened 2020-12-09
Fill data field when api error is not nil on web index

### LINKED ISSUE #5354: Web index does not show error if api call is failed
### What is the issue?
When the web component fails to call Linkerd API the index handler shows a blank page instead of showing an error message.

### How can it be reproduced?
Run the component with the wrong api-addr argument and try to open the index page.

### Logs, error output, etc
```ERRO[2020-12-09T05:09:44+03:30] template: app.tmpl.html:3:33: executing "content" at <.Data.ReleaseVersion>: nil pointer evaluating *public.VersionInfo.ReleaseVersion ```

#### `linkerd check` output

```kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid
    issuer certificate will expire on 2020-12-10T01:09:48Z

√ proxy-injector cert is valid for at least 60 days
√ sp-validator cert is valid for at least 60 days

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

linkerd-ha-checks
-----------------

Status check results are √
```

### Environment

- Kubernetes Version: 1.16.14
- Cluster Environment: bare-metal
- Host OS: Ubuntu-18.04
- Linkerd version: stable-2.9.0

### Possible solution
The problem can be fixed by filling the Data field of the appParams struct with appropriate values when err is not nil.

keyhana:
app.tmpl.html does not expect the Data field to be empty which only is the case when api call is successful.

This PR fills the Data field with appropriate values when api call returns an error.

Fixes #5354 

Signed-off-by: Keyhan Asadi <keyhana97@gmail.com>

### DIFF
--- a/web/srv/handlers.go
+++ b/web/srv/handlers.go
@@ -55,6 +55,11 @@ func (h *handler) handleIndex(w http.ResponseWriter, req *http.Request, p httpro
 	if err != nil {
 		params.Error = true
 		params.ErrorMessage = err.Error()
+		params.Data = &pb.VersionInfo{
+			GoVersion:      "Unavailable",
+			BuildDate:      "Unavailable",
+			ReleaseVersion: "Unavailable",
+		}
 		log.Error(err)
 	} else {
 		params.Data = version
--- a/web/srv/handlers_test.go
+++ b/web/srv/handlers_test.go
@@ -1,6 +1,7 @@
 package srv
 
 import (
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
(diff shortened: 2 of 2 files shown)


web/srv/handlers.go
```diff
@@ -55,6 +55,11 @@ func (h *handler) handleIndex(w http.ResponseWriter, req *http.Request, p httpro
 	if err != nil {
 		params.Error = true
 		params.ErrorMessage = err.Error()
+		params.Data = &pb.VersionInfo{
+			GoVersion:      "Unavailable",
+			BuildDate:      "Unavailable",
+			ReleaseVersion: "Unavailable",
```
cpretzer, line 61: Is there value in making these strings constants? There may even be a string for this already

web/srv/handlers_test.go
```diff
@@ -64,6 +65,51 @@ func TestHandleIndex(t *testing.T) {
 	}
 }
 
+func TestHandleIndexPublicAPIFail(t *testing.T) {
```
cpretzer, line 68: Great job adding a test for this!

cpretzer (commented):
@keyhana, I cloned your branch and ran the code locally. The change does prevent the blank page from being displayed, and the output in the browser looks like the attached image.

What do you think about making the error messaging more explicit by stating that the `api-addr` is unreachable? Would it make sense to add this as a check in the `linkerd check` output?

Screenshot from 2020-12-08 21-10-36

web/srv/handlers.go
keyhana: I did a quick search through the code base and did not find anything useful for this.
I'm not sure if making these consts have a great value but will do so if you think there is.

web/srv/handlers_test.go
keyhana: Thanks!

keyhana:
@cpretzer I can change the error message to something like this if that's OK.
```Error: <actual error message>, Try to check if api-addr config is set correctly and is reachable.```

I think it would be nice if we also add this as a check.

adleong (commented):
Hi @keyhana, thanks for your interest in the project!

I'm not sure that this particular change is valuable since deploying web without the public-api isn't a supported configuration.  It's also not very easy to hit this scenario by accident when using the recommended workflow (e.g. Docker and Kubernetes) so I don't think having a more informative error message for this situation is really that useful.  Even though the change itself is harmless, I'm a bit hesitant to take on code (and especially tests) that don't have a clear purpose.

With all of that said, I'd still be super excited to have you contribute to Linkerd.  I'd recommend looking at [issues tagged "good first issue"](https://github.com/linkerd/linkerd2/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) or asking in the #contributors channel in the Linkerd slack to find something valuable to work on.

keyhana:
Hello @adleong,
Thanks for your review.
You are right about deploying the web without the public-api, but I think many other things may go wrong in a healthy and complete Linkerd installation, and showing the error message can be helpful in those cases.
Also since the logic for showing the error message is already available in the HTML template file I think the current behavior of not showing it is a bug.
Nevertheless, I will respect your decision on either continuing this PR or closing it.

adleong:
Thanks for understanding, @keyhana, and I hope this doesn't discourage you from contributing.  We'd love to have your participation on the project!

PR #5357, opened 2020-12-09
charts/linkerd2: Allow setting custom priorityClassName for identity

### LINKED ISSUE #5356: In k8s cluster, linkerd-identity may get descheduled when scaling down the cluster
### What is the issue?
As there is no `priorityClassName` set for the `linkerd-identity` deployment, it may get descheduled completed if the cluster is scaled down and worker nodes are removed. This will prevent the `linkerd-proxy` from working when the cluster is scaled back up again.

### How can it be reproduced?
1. Install `linkerd2` in a Kubernetes cluster with 3 worker nodes with the High Availability configuration. There should be 2 pods for the `linkerd-identity` service
2. Scale down the cluster to only 1 worker node, this should max out the nodes resources so that it must not be able to schedule a `linkerd-identity` on that node. This may need to be down several times until the `linkerd-identity` is complete descheduled in the cluster.
3. Scale the cluster back to 3 worker nodes, now the `linkerd-proxy` containers will fail to be in a `Ready` state.

### Environment

- Kubernetes Version: 1.18
- Cluster Environment: EKS
- Host OS:
- Linkerd version: 2.9.0

### Possible solution
Allow the priorityClassName of the `linkerd-identity` deployment to be set, so the pod doesn't get completely descheduled in the cluster when scaling down.

pengux:
Fixes #5356

### DIFF
--- a/charts/linkerd2/templates/identity.yaml
+++ b/charts/linkerd2/templates/identity.yaml
@@ -157,6 +157,9 @@ spec:
       initContainers:
       - {{- include "partials.proxy-init" . | indent 8 | trimPrefix (repeat 7 " ") }}
       {{ end -}}
+      {{- if .Values.identity.priorityClassName }}
+      priorityClassName: {{ .Values.identity.priorityClassName }}
+      {{- end }}
       serviceAccountName: linkerd-identity
       volumes:
       - name: identity-issuer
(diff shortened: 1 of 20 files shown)


kleimkuhler:
Hey @pengux you'll need to run `go test ./cli/cmd/... --update` to update the CLI golden files for the unit tests to pass.

adleong:
Hey @pengux, I'm going to mark this PR as a draft until we work through the questions on the associated issue (#5356)

PR #5358, opened 2020-12-09
`linkerd install --ha` was only partially applying HA config

### LINKED ISSUE #5342: `linkerd install --ha` is only partially applying the HA config
### What is the issue?

When installing linkerd HA through the CLI (`linkerd install --ha`), most of the settings found in `charts/linkerd2/values-ha.yml` are being ignored.

Note that installing through Helm pointing to the `values-ha.yml` is working as expected.

### How can it be reproduced?

For example in HA the proxy injector and SP validator should have `failurePolicy: Fail` and instead they're getting configured with the default `Ignore`:

```
$ linkerd install --ha | grep failurePolicy
  failurePolicy: Ignore
  failurePolicy: Ignore
```

#### `linkerd check` output

Currently the only HA check performed is for verifying there are multiple replicas of the control plane pods. The replicas setting is being picked up by `linkerd install --ha` so the general problem is not detected by `linkerd check`.

### Environment

- Kubernetes Version: 1.18
- Host OS: Linux
- Linkerd version: 2.9.0

### Possible solution

The problem lies in how the `--ha` is being processed [here](https://github.com/linkerd/linkerd2/blob/72a0ca974df2e623a3ffbff2a03abaabaab19322/cli/cmd/options.go#L57-L76). Only some of the ha configs are being picked up. It should instead somehow apply all the configs, without hard-coding them in that function.

alpeb:
Fixes #5342

`values-ha.yml` contains the specific config for HA, but only the proxy
resources controller replicas settings were applied. This PR adds
EnablePodAntiafinity, WebhookFailurePolicy and all the resource settings
for the other CP pods.

Also the `--controller-replicas` flag is moved after the HA flags so it
can override the HA settings.

Finally, some comments no longer relevant were removed.

## How to test

Perform `linkerd install --ha` and make sure the values in
`values-ha.yml` are propagated correctly in the produced yaml.

## 2.9.1

After merging to `main`, this should be cherry-picked into the
`release/stable-2.9` branch.

### DIFF
--- a/cli/cmd/options.go
+++ b/cli/cmd/options.go
@@ -58,11 +58,10 @@ func makeInstallUpgradeFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.Fl
 					if err != nil {
 						return err
 					}
-					changedValues, err := (*values).Merge(*haValues)
+					*values, err = values.Merge(*haValues)
 					if err != nil {
 						return err
 					}
-					*values = changedValues
 				}
 				return nil
 			}),
--- a/cli/cmd/options.go
+++ b/cli/cmd/options.go
@@ -42,12 +42,6 @@ func makeInstallUpgradeFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.Fl
 	}
 
 	flags := []flag.Flag{
-		flag.NewUintFlag(installUpgradeFlags, "controller-replicas", defaults.ControllerReplicas,
-			"Replicas of the controller to deploy", func(values *l5dcharts.Values, value uint) error {
-				values.ControllerReplicas = value
-				return nil
-			}),
-
 		flag.NewStringFlag(installUpgradeFlags, "controller-log-level", defaults.GetGlobal().ControllerLogLevel,
 			"Log level for the controller and web components", func(values *l5dcharts.Values, value string) error {
 				values.GetGlobal().ControllerLogLevel = value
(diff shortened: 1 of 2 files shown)


cli/cmd/options.go
```diff
@@ -189,6 +203,14 @@ func makeInstallUpgradeFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.Fl
 	return flags, installUpgradeFlags, nil
 }
 
+func initResourceMaps(values ...map[string]interface{}) {
+	for _, value := range values {
+		if _, ok := value["Resources"]; !ok {
```
kleimkuhler, line 208: I think this should be?

```suggestion
		if _, ok := value["resources"]; !ok {
```

When I look at the output of this, Prometheus has the following field `Resources: {}`; I assume it's empty because it's supposed to be the `resources: ..` field.

cli/cmd/options.go
```diff
@@ -63,19 +57,39 @@ func makeInstallUpgradeFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.Fl
...
+					values.HeartbeatResources = haValues.HeartbeatResources
+
+					initResourceMaps(values.Prometheus, haValues.Prometheus)
+					values.Prometheus["resources"] = haValues.Prometheus["resources"]
+
+					values.ProxyInjectorResources = haValues.ProxyInjectorResources
+					values.WebhookFailurePolicy = haValues.WebhookFailurePolicy
+					values.SPValidatorResources = haValues.SPValidatorResources
+					values.TapResources = haValues.TapResources
+					values.WebResources = haValues.WebResources
```
kleimkuhler, line 82: Instead of individually merging these values, what are your thoughts on using `func (v Values) merge(src Values) (Values, error)`?

It needs to be made public, but it takes care of this and we don't need to worry about keeping this in sync with HA values if they are ever added or removed.

I tried this locally already and it seems to work fine. The only differences I see in the `linkerd install --ha` output (besides git version and TLS related values) are:

- The absence of the `Resources: {}` field that I mentioned in another comment
- The `linkerd-config` has some tracing values that I don't see in the output of your PR

cli/cmd/options.go
alpeb: Yeah there's something odd going on there, but I think this is the minimum change we require to fix the bug for now.

cli/cmd/options.go
alpeb: Good call, I've pushed a change using `Merge()` instead. I couldn't find though the tracing differences you mentioned, could you clarify?

kleimkuhler: Looks like they somehow got there in quickly testing in `Merge()` worked for me. I don't see them now 👍

cli/cmd/options.go
```diff
@@ -62,20 +58,21 @@ func makeInstallUpgradeFlags(defaults *l5dcharts.Values) ([]flag.Flag, *pflag.Fl
...
-					values.GetGlobal().Proxy.Resources.Memory.Request = haValues.GetGlobal().Proxy.Resources.Memory.Request
-					// NOTE: CPU Limits are not currently set by default HA charts.
-					//values.Global.Proxy.Cores = haValues.Global.Proxy.Cores
-					//values.Global.Proxy.Resources.CPU.Limit = haValues.Global.Proxy.Resources.CPU.Limit
-					values.GetGlobal().Proxy.Resources.Memory.Limit = haValues.GetGlobal().Proxy.Resources.Memory.Limit
+					changedValues, err := (*values).Merge(*haValues)
+					if err != nil {
+						return err
+					}
+					*values = changedValues
```
kleimkuhler, line 65: This can just be:

```suggestion
					*values, err = values.Merge(*haValues)
					if err != nil {
						return err
					}
```

kleimkuhler (approved):

adleong (approved):

PR #5361, opened 2020-12-10
jaeger: clean up values pkg

Pothulapati:
With this new way of chart rendering i.e using helm pkg directly
instead of using our own struct, we no longer need the `Values`
struct to be present, as all the rendering happens through
`map[string]interface{}`

This might be useful in future when we do validation of values, which
can also be done directly with out this, unless we don't want to deal
with conversions


adleong (approved):
I think this is fine for now since this is unused right now.  But I want to mention that we might consider bringing this back (or something similar) in the future if we want to programmatically manipulate the values.  Coercing a `map[string]interface{}` into a typed struct is more ergonomic and safer to work with.

kleimkuhler (approved):

adleong:
@Pothulapati I think this just needs a manual merge resolution to be happy

Pothulapati:
>want to mention that we might consider bringing this back (or something similar) in the future if we want to programmatically manipulate the values. Coercing a map[string]interface{} into a typed struct is more ergonomic and safer to work with.

Yeah, That is my thinking too. We should only golang build types, if there is a need (could be any extension)

PR #5365, opened 2020-12-10
Stop publishing the linkerd2-multicluster-link chart

### LINKED ISSUE #5348: Stop publishing the `linkerd2-multicluster-link` chart
That chart only serves as a repository of templates used by the `linkerd multicluster link` CLI command and cannot be installed by itself through `helm install` because it doesn't contain the `cluster-credentials` secret nor the `Link` CR that need to be generated dynamically.

This is just a matter of removing one line in `bin/helm-build`

alpeb:
Closes #5348

That chart generates the service mirror resources and related RBAC, but
doesn't generate the credentials secret nor the Link CR which require
go-client logic not available from sheer Helm templates.

This PR stops publishing that chart, and adds a comment to its README
about it.


adleong (approved):

kleimkuhler (approved):
lgtm!

PR #5369, opened 2020-12-10
Add missing tolerations in linkerd-cni helm chart (#5368)

### LINKED ISSUE #5368: Linkerd-cni helm chart missing tolerations
### What is the issue?

The linkerd-cni helm chart does not have the ability to add tolerations. This prevents the daemonset from being installed on tainted nodes.

### How can it be reproduced?

Taint a node and then attempt to install the linkerd-cni helm chart.

### Logs, error output, etc

paste the link here.)

#### `linkerd check` output

```text
your output here ...
```

### Environment

- Kubernetes Version: 1.18.1
- Cluster Environment: Equinix Metal (Lokomotive)
- Host OS: Flatcar
- Linkerd version: 2.9.1

### Possible solution

Add tolerations the same way the main chart has it specified.

rish-onesignal:
The linkerd-cni helm chart is missing tolerations on the daemonset. This
prevents the linkerd-cni daemonset from being installed on all intended
nodes.

We use the same template partial as used in the main linkerd helm chart
to add tolerations if specified to the linkerd-cni daemonset spec.

Fixes #5368

Signed-off-by: Rishabh Jain <rishabh@onesignal.com>

### DIFF
--- a/charts/linkerd2-cni/templates/cni-plugin.yaml
+++ b/charts/linkerd2-cni/templates/cni-plugin.yaml
@@ -184,6 +184,9 @@ spec:
       annotations:
         {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.cniPluginVersion) .Values.cliVersion}}
     spec:
+      {{- if .Values.tolerations }}
+      {{- include "linkerd.tolerations" . | nindent 6 }}
+      {{- end }}
       nodeSelector:
         beta.kubernetes.io/os: linux
       hostNetwork: true
--- a/charts/linkerd2-cni/values.yaml
+++ b/charts/linkerd2-cni/values.yaml
@@ -41,6 +41,11 @@ priorityClassName: ""
 proxyInjectAnnotation: linkerd.io/inject
 proxyInjectDisabled: disabled
 
+# -|- Tolerations section, See the 
+# [K8S documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
+# for more information
+#tolerations:
+
 #
 ## For Private docker registries, authentication is needed.
 # If the control plane service images are pulled from a


charts/linkerd2-cni/templates/cni-plugin.yaml
```diff
@@ -184,6 +184,9 @@ spec:
       annotations:
         {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.cniPluginVersion) .Values.cliVersion}}
     spec:
+      {{- if .Values.tolerations }}
+      {{- include "linkerd.tolerations" . | nindent 6 }}
```
kleimkuhler, line 188: Since we are adding a `tolerations` value to the CNI chart, I think this should just be:

```suggested
      {{- include ".Values.tolerations" . | nindent 6 }}
```

I don't think we want to use `linkerd.tolerations` here since the tolerations for the CNI plugin may be different from the rest of the Linkerd installation.

kleimkuhler (commented):
Thanks @rish-onesignal! This makes sense to add since all the extensions as well as controller components also have tolerations, but I think we want to use a different value as I commented above.

@alpeb or @Pothulapati are very familiar with the Helm charts as well though so they may have some input on this.

charts/linkerd2-cni/templates/cni-plugin.yaml
rish-onesignal: My understanding is that this will use the template partial defined here: https://github.com/linkerd/linkerd2/blob/main/charts/partials/templates/_tolerations.tpl, which essentially uses `.Values.tolerations` but does additional formatting with trim etc. Do you still want just `.Values.tolerations`?

alpeb: Correct, `linkerd.toleration` is just making reference to the partial template, so it's fine.

alpeb (approved):
Looks good to me :+1: , thanks @rish-onesignal !

charts/linkerd2-cni/templates/cni-plugin.yaml
alpeb: ... and we use a template because otherwise the `tolerations:` part would not be rendered, only its contents.

kleimkuhler: Okay cool 👍

kleimkuhler (approved):
Thanks @rish-onesignal!

alpeb:
@zaharidichev or @hemakl would you please give this one an :eye: ?

zaharidichev (approved):
LGTM

PR #5395, opened 2020-12-16
Support multi-cert CA bundles in healthcheck

### LINKED ISSUE #5292: Make `linkerd check` comprehend mulit-cert CA bundles

```
linkerd-identity
----------------
× certificate config is valid
```

cypherfox:

This change will place the comparison of the contents of `ca.cert` of the secret and the trust anchors into a separate test and support arbitrary long chains in ca.crt.

Fixes #5292


cypherfox:
Wrong commit path.

PR #5396, opened 2020-12-16
Support multi-cert CA bundles in healthcheck

### LINKED ISSUE #5292: Make `linkerd check` comprehend mulit-cert CA bundles

cypherfox:

This change will place the comparison of the contents of `ca.cert` of the secret and the trust anchors into a separate test and support arbitrary long chains in ca.crt.

Fixes #5292

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -934,7 +934,7 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
-					description: "CA certificates match",
+					description: "trust anchors in config and secret  match",
 					hintAnchor:  "l5d-identity-cert-config-valid",
 					fatal:       true,
 					check: func(ctx context.Context) (err error) {
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -934,6 +934,15 @@ func (hc *HealthChecker) allCategories() []category {
 					},
 				},
 				{
+					description: "CA certificates match",
+					hintAnchor:  "l5d-identity-cert-config-valid",
+					fatal:       true,
+					check: func(ctx context.Context) (err error) {
+						err = hc.checkTrustAnchorsMatch(ctx)
+						return
+					},
+				},
+				{
 					description: "trust anchors are using supported crypto algorithm",
 					hintAnchor:  "l5d-identity-trustAnchors-use-supported-crypto",
 					fatal:       true,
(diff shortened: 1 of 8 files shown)


cypherfox:
lint is failing with a point about a useless function param that I understand in principle, but fixing it would reduce readability and re-usability in this particular place.
Is there some form of annotation to have the linter ignore that function?

alpeb:
> lint is failing with a point about a useless function param that I understand in principle, but fixing it would reduce readability and re-usability in this particular place.
> Is there some form of annotation to have the linter ignore that function?

You can prepend the function declaration with `//nolint:unparam`

adleong (commented):
Thanks @cypherfox.  I think I'm still not totally understanding your use-case here.  

The trust root should be very long lived and not rotated regularly.  (I think in our documentation we recommend an expiry period of 10 years).  If your cert-manager certs are issued by some other PKI, then you should pass the trust roots of that PKI chain to Linkerd.

In the rare situations where a rotation of the trust root is required (for example if you lose access to the root key) then I would expect the trust bundle that Linkerd uses to differ from the trust bundle in ca.crt only temporarily while the rotation is taking place and these two should be brought back into sync afterwords.

Therefore, I expected that this new check should be a warning (not fatal) and should check that ca.crt and the Linkerd trust bundle are identical.

This means that this new check would fail temporarily during the trust root rotation but would not block the other checks from running.  It would start passing again once the rotation is complete and ca.crt and the Linkerd trust bundle are synchronized again.

cypherfox:
> Thanks @cypherfox. I think I'm still not totally understanding your use-case here.
> 
> The trust root should be very long lived and not rotated regularly. (I think in our documentation we recommend an expiry period of 10 years).

I would strongly disagree with that assertion! Even the default 1 year is too long! As a time frame for trusting the key length as well as the algorithms 10 years is a huge risk.

But much more important is the need to regularly do this manual operation, which is not only security relevant but also has a high risk of service down time if done incorrectly.

> If your cert-manager certs are issued by some other PKI, then you should pass the trust roots of that PKI chain to Linkerd.

This does not remove the need to rotate these anchors eventually.

> 
> In the rare situations where a rotation of the trust root is required (for example if you lose access to the root key) then I would expect the trust bundle that Linkerd uses to differ from the trust bundle in ca.crt only temporarily while the rotation is taking place and these two should be brought back into sync afterwords.

Even the time it takes for the l5d control plane to propagate these new certs throughout the cluster will cause a service interruption. We have observed that before. That was one of the motivations to push the cert-manager support forward.

> 
> Therefore, I expected that this new check should be a warning (not fatal) and should check that ca.crt and the Linkerd trust bundle are identical.

I agree. My biggest problem with the current behaviour is the fact that the check is marked as fatal.

> 
> This means that this new check would fail temporarily during the trust root rotation but would not block the other checks from running. It would start passing again once the rotation is complete and ca.crt and the Linkerd trust bundle are synchronized again.

I make that change to the flag, if you wish. I have separated the collection of the trust anchors  from the actual comparison of the anchors and ca.crt

cypherfox:
@adleong I have thought about this a bit more, and IMHO it makes no sense to mark the check as warning only.

Either you a) you don't have `kubernetes.io/tls` as the secret scheme, and the test will short circuit with a success. Or b) you only have a single trust anchor in the config, in which case the communication will only work with the exact same cert in `ca.crt`. Or c) there is more than one cert in the config map, and the root cert in `ca.crt`.

In either case b) or c), if this check fails to succeed, your l5d setup is broken and proxies cannot communicate. I do not think that this should only be a warning. You need to fix that for your services to work.

adleong:
Thanks for discussing this with me the other day, @cypherfox.  The additional context really helped.  Based on that discussion, I think we should just altogether eliminate the part of this check which compares `ca.crt` and the trust bundle.

cypherfox:
Closed for #5424

PR #5397, opened 2020-12-16
Add missing icon entry to jaeger chart

alpeb:
This is required for `helm lint` to pass. Its absence was what caused
the last CI edge release to fail and so we had to manually upload the
charts.


adleong (approved):

cpretzer (approved):
LGTM

PR #5399, opened 2020-12-16
Fix CLI install/upgrade overriding settings in HA

### LINKED ISSUE #5385: linkerd upgrade --ha fails with "failed to read CA: not PEM-encoded"
### What is the issue?

`linkerd upgrade --ha` fails on `edge-20.12.3`

### How can it be reproduced?

Command succeeds prior to this commit, fails with it: https://github.com/linkerd/linkerd2/commit/35612ae26823a9dfec51facc927f1b68051cd255

```bash
# works
$ git checkout 377b38f0bfb83421a3b74434aea43188810a0d01
$ go run cli/main.go upgrade --ha
---
###
### Linkerd Namespace
...

# breaks on the next commit
$ git checkout 35612ae26823a9dfec51facc927f1b68051cd255
$ go run cli/main.go upgrade --ha
Error: failed to validate issuer credentials: failed to read CA: not PEM-encoded
Usage:
...
```

#### `linkerd check` output

```bash
$ linkerd check --context aks-dive-staging
kubernetes-api
--------------
√ can initialize the client
...
    issuer certificate will expire on 2021-01-27T08:02:19Z
...
control-plane-version
---------------------
‼ control plane is up-to-date
    is running version 20.12.1 but the latest edge version is 20.12.3
    control plane running edge-20.12.1 but cli running edge-20.12.3
...
Status check results are √
```

### Environment

- Kubernetes Version: v1.18.2
- Cluster Environment: AKS
- Host OS: Linux
- Linkerd version: edge-20.12.3

### Additional context

We use cert-manager.

alpeb:
Fixes #5385

## The problems

- `linkerd install --ha` isn't honoring flags
- `linkerd upgrade --ha` is overridding existing configs silently or failing with an error
- **Upgrading HA instances from before 2.9 to version 2.9.1 results in configs being overridden silently, or the upgrade fails with an error**

## The cause

The change in #5358 attempted to fix `linkerd install --ha` that was only applying some of the `values-ha.yaml` defaults, by calling `charts.NewValues(true)` and merging that with the values built from `values.yaml` overriden by the flags. It turns out the `charts.NewValues()` implementation was by itself merging against `values.yaml` and as a result any flag was getting overridden by its default.

This also happened when doing `linkerd upgrade --ha` on an existing instance, which could result in silently overriding settings, or it could also fail loudly like for example when upgrading set up that has an external issuer (in this case the issuer cert won't be able to be read during upgrade and an error would occur as described in #5385).

Finally, when doing `linkerd upgrade` (no --ha flag) on an HA install from before 2.9 results in configs getting overridden as well (silently or with an error) because in order to generate the `linkerd-config-overrides` secret, the original install flags are retrieved from `linkerd-config` via the `loadStoredValuesLegacy()` function which then effectively ends up performing a `linkerd upgrade` with all the flags used for `linkerd install` and falls into the same trap as above.

## The fix

In `values.go` the faulting merging logic is not used anymore, so now `NewValues()` only returns the default values from `values.yaml` and doesn't require an argument anymore. It calls `readDefaults()` which now only returns the appropriate values depending on whether we're on HA or not.
There's a new function `MergeHAValues()` that merges `values-ha.yaml` into the current values (it doesn't look into `values.yaml` anymore), which is only used when processing the `--ha` flag in `options.go`.

## How to test

To replicate the issue try setting a custom setting and check it's not applied:
```bash
linkerd install --ha --controller-log level debug | grep log.level
- -log-level=info
```

## Followup

This wasn't caught because we don't have HA integration tests. Now that our test infra is based on k3d, it should be easy to make such a test using a cluster with multiple nodes. Either that or issuing `linkerd install --ha` with additional configs and compare against a golden file.

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -369,7 +369,7 @@ func runChecksJSON(wout io.Writer, werr io.Writer, hc *healthcheck.HealthChecker
 }
 
 func renderInstallManifest(ctx context.Context) (string, error) {
-	values, err := charts.NewValues(false)
+	values, err := charts.NewValues()
 	if err != nil {
 		return "", err
 	}
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -48,7 +48,7 @@ func runInjectCmd(inputs []io.Reader, errWriter, outWriter io.Writer, transforme
 }
 
 func newCmdInject() *cobra.Command {
-	defaults, err := charts.NewValues(false)
+	defaults, err := charts.NewValues()
 	if err != nil {
 		fmt.Fprint(os.Stderr, err.Error())
 		os.Exit(1)
--- a/cli/cmd/inject_test.go
+++ b/cli/cmd/inject_test.go
@@ -575,7 +575,7 @@ func TestWalk(t *testing.T) {
 }
 
 func TestProxyConfigurationAnnotations(t *testing.T) {
-	baseValues, err := linkerd2.NewValues(false)
+	baseValues, err := linkerd2.NewValues()
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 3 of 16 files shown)


adleong (approved):
:star: Nice work @alpeb!

cpretzer (approved):
:+1:

alpeb:
This fix has been included in stable-2.9.2

PR #5406, opened 2020-12-18
Add PodDisruptionBudgets to control plane (#5398)

### LINKED ISSUE #5398: Add Support for PodDisruptionBudgets in Helm Chart

When rolling the nodes of the cluster and/or scaling in it is possible for all the control plane pods to be terminated

An option to provision PodDisruptionBudgets for the control plane services as part of the helm chart

I am currently planning to configure these outside of helm using terraform, but it feels like something that could come bundled with the helm chart.

A boolean value on the helm chart

tustvold:
Closes #5398

* Add PodDisruptionBudget to controller deployments
* Add .yaml to editorconfig
* ~~Add rollingUpdate strategy to web deployment along with PodDisruptionBudget~~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,6 +29,9 @@ indent_style = space
 [*.yml]
 indent_size = 2
 
+[*.yaml]
+indent_size = 2
+
 [*.proto]
 indent_size = 2
 indent_style = space
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -21,6 +21,24 @@ spec:
   - name: http
     port: 8085
     targetPort: 8085
+{{- if .Values.enablePodAntiAffinity }}
+---
+kind: PodDisruptionBudget
+apiVersion: policy/v1beta1
+metadata:
+  name: linkerd-controller-api
+  namespace: {{.Values.global.namespace}}
+  labels:
+    {{.Values.global.controllerComponentLabel}}: controller
+    {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
+  annotations:
+    {{.Values.global.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.global.linkerdVersion) .Values.global.cliVersion}}
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      {{.Values.global.controllerComponentLabel}}: controller
+{{- end }}
 ---
 {{- $tree := deepCopy . }}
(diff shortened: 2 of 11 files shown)


charts/linkerd2/templates/web.yaml
```diff
@@ -49,6 +67,11 @@ spec:
       {{.Values.global.controllerComponentLabel}}: web
       {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
       {{- include "partials.proxy.labels" $tree.Values.global.proxy | nindent 6}}
+  {{- if .Values.enablePodAntiAffinity }}
+  strategy:
+    rollingUpdate:
+      maxUnavailable: 1
+  {{- end }}
```
kleimkuhler, line 74: I'm not sure that we want this for the web deployment. In HA mode, we don't scale the web replicas up to 3 like we do with other controller components; we keep it at 1.

This is because unlike the other controller components, it's not necessary for linkerd to continue working properly if the web deployment is unavailable during an update.

kleimkuhler (changes requested):
Overall this looks good!

We should back out the change that adds the rolling update strategy to the web deployment as I explained in a comment.

After that, I think this change is good to go.

charts/linkerd2/templates/web.yaml
adleong: I agree with @kleimkuhler and think we shouldn't add a PDB to the web deployment.  Essentially, adding a PDB would prevent scaling down the cluster if doing so would de-schedule more than one web pod.  Since web is not a critical component, I think this should not prevent scaling events.

tustvold: I've backed out this change and the addition of a PDB to the tap component as this appears to have been extracted into a separate linkerd-viz chart

kleimkuhler: Great thanks @tustvold! Pending CI this looks good 👍

tustvold:
I'm not sure how to kick the CI back into life if I even can, could someone take a look?

adleong (approved):

kleimkuhler:
@tustvold Hm yes this is odd. Could you push an empty commit (not a force push) with `git commit --allow-empty -s` and see if that fixes it?

kleimkuhler:
@tustvold If you can also merge `main` into this that'd be great! It has a picked up a few fixes that were causing CI to fail.

tustvold:
Rebased on top of current main

kleimkuhler (approved):

Issue #4459, opened 2020-05-21
cli: identity cert <pod-selection>


olix0r:
Add an `identity cert` command that returns the certificates from a selection of pods.

These certificates can be obtained by initializing an SNI TLS session to the proxy's admin port. For example:

```
:; echo | \
    openssl s_client -servername linkerd-service-mirror.linkerd-multicluster.serviceaccount.identity.linkerd.dev.k3d.example.com -connect localhost:4191 2>/dev/null | \
    openssl x509 -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 10 (0xa)
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: CN = identity.linkerd.dev.k3d.example.com
        Validity
            Not Before: May 21 18:23:35 2020 GMT
            Not After : May 22 18:24:15 2020 GMT
        Subject: CN = linkerd-service-mirror.linkerd-multicluster.serviceaccount.identity.linkerd.dev.k3d.example.com
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:ee:d4:33:af:83:5f:91:75:aa:b4:82:cd:74:21:
                    11:4d:04:86:98:19:49:30:2f:07:25:b1:18:67:a9:
                    ae:8d:9f:ce:d8:23:61:93:d0:c5:3a:98:fd:3b:9b:
                    e7:53:98:0f:0d:7a:21:33:8e:ec:87:a6:9b:68:cb:
                    e7:92:41:a0:eb
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication, TLS Web Client Authentication
            X509v3 Authority Key Identifier: 
                keyid:97:63:30:F1:04:94:17:50:CB:9F:1D:73:FB:FB:06:DC:D8:5A:0E:FD
            X509v3 Subject Alternative Name: 
                DNS:linkerd-service-mirror.linkerd-multicluster.serviceaccount.identity.linkerd.dev.k3d.example.com
    Signature Algorithm: ecdsa-with-SHA256
         30:45:02:21:00:d9:d1:c6:c1:34:79:f9:44:db:7c:9b:82:f3:
[...]

PR #5408, opened 2020-12-18
 CLI: Introduced `identity` command to fetch tls-certificates for a pod

jimil749:
CLI: Introduced `identity` command to fetch tls-certificates for a pod (#4459)

Modified and added a new cli command, which initiates a sni-tls session to the proxy's admin port and returns the certificate.

Usage:
- `linkerd identity pod/<pod-name>`   : fetches certificate from the specified pod
- `linkerd identity -n emojivoto`           : fetches certificate from all pods in the emojivoto namespace
- `linkerd identity -l app=svc/emoji`      : fetches certificate from all pods with label app=svc/emoji

Currently, I am not a 100% sure of the way I am outputting the certificates, I'll change that if needed, but overall these commands do fetch the required certificates.

Screenshot:

Screenshot from 2020-12-18 22-11-28

I would be happy to work upon any changes suggested

### DIFF
--- a/cli/cmd/identity.go
+++ b/cli/cmd/identity.go
@@ -10,8 +10,6 @@ import (
 
 	"github.com/grantae/certinfo"
 	"github.com/linkerd/linkerd2/pkg/k8s"
-	"github.com/linkerd/linkerd2/pkg/k8s/resource"
-	tls1 "github.com/linkerd/linkerd2/pkg/tls"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -66,14 +64,6 @@ func newCmdIdentity() *cobra.Command {
 				return err
 			}
 
-			resources, err := resource.FetchKubernetesResources(cmd.Context(), k8sAPI, metav1.ListOptions{LabelSelector: k8s.ControllerNSLabel})
-			if err != nil {
-				return err
-			}
-			// check if linkerd control plane exists
-			if len(resources) == 0 {
-				return fmt.Errorf("Cannot find Linkerd\nValidate the install with: linkerd check")
-			}
 			if len(args) == 0 && options.selector == "" {
 				return fmt.Errorf("Provide the pod name argument or use the selector flag")
 			}
@@ -84,7 +74,12 @@ func newCmdIdentity() *cobra.Command {
 			}
 
 			resultCerts := getCertificate(k8sAPI, pods, k8s.ProxyAdminPortName, emitLog)
-			for _, resultCert := range resultCerts {
+			if len(resultCerts) == 0 {
+				fmt.Print("Could not fetch Certificate. Ensure that the pod(s) are meshed by running `linkerd inject`")
+				return nil
+			}
+			for i, resultCert := range resultCerts {
+				fmt.Printf("\nPOD %s (%d of %d)\n\n", resultCert.pod, i+1, len(resultCerts))
 				if resultCert.err != nil {
 					fmt.Printf("\n%s", resultCert.err)
 					return nil
@@ -97,8 +92,6 @@ func newCmdIdentity() *cobra.Command {
 					return nil
 				}
 				fmt.Print(result)
-				fmt.Println("pem: |")
-				fmt.Print(tls1.EncodeCertificatesPEM(cert))
 			}
 			return nil
 		},
--- a/cli/cmd/identity.go
+++ b/cli/cmd/identity.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"net"
 	"os"
 	"regexp"
 	"sync/atomic"
@@ -13,6 +12,7 @@ import (
 
 	"github.com/linkerd/linkerd2/controller/api/util"
 	"github.com/linkerd/linkerd2/pkg/k8s"
+	tls1 "github.com/linkerd/linkerd2/pkg/tls"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,12 +31,14 @@ type Certificates struct {
 type identityOptions struct {
 	pod       string
 	namespace string
+	selector  string
 }
 
 func newIdentityOptions() *identityOptions {
 	return &identityOptions{
 		pod:       "",
 		namespace: "",
+		selector:  "",
 	}
 }
 
@@ -45,40 +47,48 @@ func newCmdIdentity() *cobra.Command {
 	options := newIdentityOptions()
 
 	cmd := &cobra.Command{
-		Use:   "identity [name]",
+		Use:   "identity [flags] (PODS)",
 		Short: "Display the certificate(s) of one or more selected pod(s)",
 		Long: `Display the certificate(s) of one or more selected pod(s).
 		
-		This command initiates a port-forward to a given pod or a set of pods, and
+		This command initiates a port-forward to a given pod or a set of pods and
 		fetches the tls certificate.
 		`,
-		Example: ` #Get certificate from pod foo-bar in the default namespace.
+		Example: ` 
+		#Get certificate from pod foo-bar in the default namespace.
 		linkerd identity foo-bar
 		
-		#Get certificate from all pods in the default namespace
-		linkerd identity
-		
-		#Get certificate from the web deployment in the emojivoto namespace
-		linkerd identity deploy/web -n emojivoto`,
-		Args: cobra.MinimumNArgs(1),
+		#Get certificate from all pods with name=nginx
+		linkerd identity -l name=nginx
+		`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
 			if err != nil {
 				return err
 			}
-			pods, err := getPods(cmd.Context(), k8sAPI, options.namespace, args[0])
+			pods, err := getPods(cmd.Context(), k8sAPI, options.namespace, options.selector, args)
 			if err != nil {
-				//fmt.Print("")
 				return err
 			}
-			// fmt.Printf("\nPod length: %d", len(pods))
 			results := getCertificate(k8sAPI, pods, k8s.ProxyAdminPortName, 30*time.Second, emitLog)
 
-			for _, result := range results {
-				// fmt.Printf("\nDEBUG")
-				// content := fmt.Sprintf("#\n# POD %s (%d of %d)\n#\n", result.pod, i+1, len(results))
+			for i, result := range results {
+				fmt.Printf("\n POD %s (%d of %d)\n\n", result.pod, i+1, len(results))
 				if result.err == nil {
-					fmt.Printf("\n%+v\n", result.Certificate)
+					fmt.Println("Version: ", result.Certificate[0].Version)
+					fmt.Println("Serial Number: ", result.Certificate[0].SerialNumber)
+					fmt.Println("Signature Algorithm: ", result.Certificate[0].SignatureAlgorithm)
+					fmt.Println("Validity")
+					fmt.Println("\t Not Before: ", result.Certificate[0].NotBefore.Format(time.RFC3339))
+					fmt.Println("\t Not After: ", result.Certificate[0].NotAfter.Format(time.RFC3339))
+					fmt.Println("Subject: ", result.Certificate[0].Subject)
+					fmt.Println("Issuer:  ", result.Certificate[0].Issuer)
+					fmt.Println("Subject Public Key Info:")
+					fmt.Println("\tPublic Key Alogrithm: ", result.Certificate[0].PublicKeyAlgorithm)
+					fmt.Println("Signature: \n", result.Certificate[0].Signature)
+					resultCerticate := tls1.EncodeCertificatesPEM(result.Certificate[0])
+
+					fmt.Printf("\n%s", resultCerticate)
 				} else {
 					fmt.Printf("\n%s\n", result.err)
 				}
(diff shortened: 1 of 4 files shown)


adleong:
This is so awesome. 😱 

Can't wait to test this out

jimil749:
@adleong, did you get the chance to review/test this?

cpretzer:
@jimil749 I had a chance to run some basic tests and this looks good! Thanks for your hard work.

One thing that I noticed is that the command doesn't write any output when Linkerd isn't installed in the cluster.

I had a new kind cluster running without Linkerd installed, and when I ran the `linkerd identity ...` command, the command returned with no output or indication that anything happened.

In contrast, the `linkerd stat deploy` and `linkerd edges deploy` commands have this output when the Linkerd control plane is not deployed:

```bash
$ linkerd stat deploy
Cannot find Linkerd: configmaps "linkerd-config" not found
```

What do you think about adding a check to make sure that linkerd is installed?

jimil749:
@cpretzer, thanks for the feedback, I'll do the needful. :)

cpretzer:
👍

jimil749:
@cpretzer I think the reason we aren't getting any output is because we do not have the proxy sidecar container injected. The `metrics` command behaves the same way as `identity` (returns with no output or any indication)! I am not using the control plane APIs to fetch the certs, I am simply establishing a sni-tls connection to the proxy's admin port to do so. The check for whether the control plane is installed or not is done while building a new public API client (in case of `stat` and `edges`), which is not required in my case. I am currently not a 100% sure of how to add the checks. I think checking whether the service is meshed i.e the proxy exists would be a better check.

cpretzer:
@jimil749 I see what you're saying. The `edges` and `stats` commands use the `checkPublicAPIClientOrExit` [function](https://github.com/linkerd/linkerd2/blob/main/cli/cmd/edges.go#L85)

Perhaps the correct approach is similar to the one used by the `uninstall` [command](https://github.com/linkerd/linkerd2/blob/main/cli/cmd/uninstall.go#L41), which checks whether the linkerd namespace exists and exits if it does not.

jimil749:
@cpretzer, Sorry for the delayed response, but I've added the check for the linkerd control plane but I wasn't able to test the command entirely, because I am having an error in installing the control plane in my cluster after fetching the latest commit, which I think is because of [this](https://github.com/linkerd/linkerd2/pull/5340), I am getting the same error as in this [comment](https://github.com/linkerd/linkerd2/pull/5340#issuecomment-747104433). Not a 100% sure, but is that the reason the tests are failing?

cpretzer:
hi @jimil749 thanks for adding the check for the namespace. We're working on getting the build fixed, and once it is, we can run the tests again.

adleong (commented):
I played with this a little bit and it's pretty great.

I like the `linkerd identity (pod-name)` mode, though I think we should drop the resource type or at least make it optional since only pods are supported.  i.e. instead of `linkerd identity pod/linkerd-destination-6887bb8f49-h8qvr` it should be `linkerd identity linkerd-destination-6887bb8f49-h8qvr`

I also like the `-l` selector mode.

I found the third mode where it prints all certificates in a namespace to be confusing and I think we can remove that.

I also looked into certificate formatting a bit more and found https://github.com/smallstep/certinfo.  I think we should use this library for formatting the certificates.

cli/cmd/identity.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// getCertificate fetches the certificates for all the pods
+func getCertificate(k8sAPI *k8s.KubernetesAPI, pods []corev1.Pod, portName string, waitingTime time.Duration, emitLog bool) []certificate {
+	var certificates []certificate
+	resultChan := make(chan certificate)
+	var activeRoutines int32
+	for _, pod := range pods {
+		atomic.AddInt32(&activeRoutines, 1)
+		go func(p corev1.Pod) {
```
adleong, line 127: I think it's fine to fetch each certificate in series rather than doing them in parallel.  This should let us simplify this and get rid of the channel.  It also happens to fix the bug where if there are no pods selected by the label selector, the select hangs.

cli/cmd/identity.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	rootPEM, err := getTrustAnchor(pod, "linkerd-proxy")
+	if err != nil {
+		return nil, err
+	}
+
+	re := regexp.MustCompile("[0-9]+")
+	res := re.FindString(url)
+
+	connURL := "localhost:" + res
```
adleong, line 209: Can you explain what's going on here?  We're extracting the port from the port-forward url and appending it onto localhost?  Isn't the port-forward url already a localhost address?

cli/cmd/identity.go
```diff
@@ -0,0 +1,295 @@
...
+
+	roots := x509.NewCertPool()
+	roots.AppendCertsFromPEM([]byte(rootPEM))
+
+	conn, err := tls.Dial("tcp", connURL, &tls.Config{
+		RootCAs:    roots,
```
adleong, line 215: Do we actually need to specify the rootCA and the serverName here?  Can we use `InsecureSkipVerify` instead?  We just want to print the certificate, not validate it.

cli/cmd/identity.go
jimil749: The thing is that the `tls.Dial` function in Go expects a domain name with port and not the entire url, and `portForward.URLFor()` gives us the URL. Just had to trim the `http://` part from the url, which I now did using the `strings.Trim`. I agree this was an overkill and changed it.

jimil749:
> 
> I like the `linkerd identity (pod-name)` mode, though I think we should drop the resource type or at least make it optional since only pods are supported. i.e. instead of `linkerd identity pod/linkerd-destination-6887bb8f49-h8qvr` it should be `linkerd identity linkerd-destination-6887bb8f49-h8qvr`
> 
Yeah, makes sense. Will do the necessary changes!
> 
> I found the third mode where it prints all certificates in a namespace to be confusing and I think we can remove that.
> 
The reason to do was this [comment](https://github.com/linkerd/linkerd2/issues/4459#issuecomment-684084431), where it is mentioned that a command like `linkerd -n ns identity` can be used to get certificates for all pods in the namespace. I will remove the command if it is not required as such
> I also looked into certificate formatting a bit more and found https://github.com/smallstep/certinfo. I think we should use this library for formatting the certificates.
>
I will take a look at the library. :+1:

cli/cmd/identity.go
jimil749: We need the serverName here, but yes if we use `InsecureSkipVerify` we can do away the the rootCA.

jimil749:
@adleong, I've addressed the changes proposed, here's the summary of the changes:
- Fetching the certificates in series rather than parallel.
- URL trim using `strings.Trim` for `tls.Dial()` (instead of extracting the portno from the URL)
- Dropped the resource type from `linkerd identity` command
- Removed the `linkerd -n ns` subcommand, which used to fetch certificates for all pods in the provided ns
- Used `grantae/certinfo` library to print the certificates in a format similar to OpenSSL
- Removed RootCAs field from `tls.Dial`

cli/cmd/identity.go
```diff
@@ -0,0 +1,231 @@
...
+		#Get certificate from all pods with name=nginx
+		linkerd identity -l name=nginx
+		`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+			if err != nil {
+				return err
+			}
+
+			resources, err := resource.FetchKubernetesResources(cmd.Context(), k8sAPI, metav1.ListOptions{LabelSelector: k8s.ControllerNSLabel})
```
adleong, line 69: I don't think we need this check that the control plane exists.

cli/cmd/identity.go
```diff
@@ -0,0 +1,231 @@
...
+				}
+				certChain := resultCert.Certificate
+				cert := certChain[len(certChain)-1]
+				result, err := certinfo.CertificateText(cert)
+				if err != nil {
+					fmt.Printf("\n%s", err)
+					return nil
+				}
+				fmt.Print(result)
+				fmt.Println("pem: |")
```
adleong, line 100: Since we're already printing the entire certificate in an openssl-like format, I don't think we also need to print the PEM encoding.

cli/cmd/identity.go
```diff
@@ -0,0 +1,231 @@
...
+				return fmt.Errorf("Provide the pod name argument or use the selector flag")
+			}
+
+			pods, err := getPods(cmd.Context(), k8sAPI, options.namespace, options.selector, args)
+			if err != nil {
+				return err
+			}
+
+			resultCerts := getCertificate(k8sAPI, pods, k8s.ProxyAdminPortName, emitLog)
+			for _, resultCert := range resultCerts {
```
adleong, line 87: Can we restore the feature where, if printing multiple certificates, it shows the pod name?

cli/cmd/identity.go
```diff
@@ -0,0 +1,231 @@
...
+			if len(args) == 0 && options.selector == "" {
+				return fmt.Errorf("Provide the pod name argument or use the selector flag")
+			}
+
+			pods, err := getPods(cmd.Context(), k8sAPI, options.namespace, options.selector, args)
+			if err != nil {
+				return err
+			}
+
+			resultCerts := getCertificate(k8sAPI, pods, k8s.ProxyAdminPortName, emitLog)
```
adleong, line 86: We should probably print an error message if no certificate are returned here.  i.e. if the user specified an un-meshed pod.

adleong (commented):
This is so awesome!  I'm really excited for this to land. 😻

jimil749:
@adleong, I made the changes as suggested. But I was not sure of what exact error message to print out when no certificates are returned. Maybe you could suggest a better message to output?

kleimkuhler (commented):
@jimil749 This is super cool thanks for getting this change up!

I've left a few smaller comments focused around formatting help text and such.

There are a few comments though that are more significant. Let me know if you have any questions about those I'd be glad to clarify a little more!

cli/cmd/identity.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func getPods(ctx context.Context, clientset kubernetes.Interface, namespace string, selector string, arg []string) ([]corev1.Pod, error) {
+	if len(arg) > 0 {
+		pod, err := clientset.CoreV1().Pods(namespace).Get(ctx, arg[0], metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		return []corev1.Pod{*pod}, nil
+	}
```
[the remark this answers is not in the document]
jimil749: Oh alright. To be honest, I had not thought about such a situation, I was under the impression that we just want to fetch cert for a single pod. I will make the required changes! :)

cli/cmd/identity.go
```diff
@@ -0,0 +1,224 @@
...
+		fmt.Fprintf(os.Stderr, "Error running port-forward: %s", err)
+		return nil, err
+	}
+
+	certURL := portForward.URLFor("")
+	return getCertResponse(certURL, pod)
+}
+
+func getCertResponse(url string, pod corev1.Pod) ([]*x509.Certificate, error) {
+	serverName, err := getServerName(pod, "linkerd-proxy")
```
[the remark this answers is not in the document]
jimil749: Just to be sure, I am supposed to add the const in `linkerd2/pkg/k8s/labels.go`, right?

cli/cmd/identity.go
```diff
@@ -0,0 +1,224 @@
...
+			}
+
+			pods, err := getPods(cmd.Context(), k8sAPI, options.namespace, options.selector, args)
+			if err != nil {
+				return err
+			}
+
+			resultCerts := getCertificate(k8sAPI, pods, k8s.ProxyAdminPortName, emitLog)
+			if len(resultCerts) == 0 {
+				fmt.Print("Could not fetch Certificate. Ensure that the pod(s) are meshed by running `linkerd inject`")
```
adleong, line 78: I think this error message is good!  It does need to have a newline at the end though.

adleong (approved):
Thanks for updating, @jimil749!

jimil749:
@kleimkuhler, I've made the required changes!

cli/cmd/identity.go
```diff
@@ -0,0 +1,227 @@
...
+
+			resultCerts := getCertificate(k8sAPI, pods, k8s.ProxyAdminPortName, emitLog)
+			if len(resultCerts) == 0 {
+				fmt.Print("Could not fetch Certificate. Ensure that the pod(s) are meshed by running `linkerd inject`\n")
+				return nil
+			}
+			for i, resultCert := range resultCerts {
+				fmt.Printf("\nPOD %s (%d of %d)\n\n", resultCert.pod, i+1, len(resultCerts))
+				if resultCert.err != nil {
+					fmt.Printf("\n%s", resultCert.err)
```
kleimkuhler, line 83: We should have a newline here after printing the error message or else (probably depending on your terminal) it indicate that it's missing. You can see that symbol at the end of output below:

```
❯ ./target/release/linkerd2-cli-git-a3b2afa6-linux-amd64 identity -n kube-system local-path-provisioner-6d59f47c7-7fnt7

POD local-path-provisioner-6d59f47c7-7fnt7 (1 of 1)


failed to find linkerd-admin port in linkerd-proxy container for given pod spec⏎ 
```

```suggestion
					fmt.Printf("\n%s\n", resultCert.err)
```

cli/cmd/identity.go
```diff
@@ -0,0 +1,227 @@
...
+					return nil
+				}
+				certChain := resultCert.Certificate
+				cert := certChain[len(certChain)-1]
+				result, err := certinfo.CertificateText(cert)
+				if err != nil {
+					fmt.Printf("\n%s", err)
```
kleimkuhler, line 90: Here as well!

kleimkuhler (commented):
This looks great! I have one more suggestion to just make sure errors are printing as expected.

After that it should be good. Sorry I did not catch this in the previous review!

kleimkuhler:
Thanks for addressing all these reviews @jimil749!

kleimkuhler (approved):

PR #5407, opened 2020-12-18
Add missing label to linkerd-config-overrides secret

alpeb:
Fixes #5385 (second bug in there)

Added missing label `linkerd.io/control-plane-ns=linkerd` that all the
control plane resources must have, that is passed to `kubectl apply
--prune`

PR #5409, opened 2020-12-18
add release notes for edge-20.12.4

hawkw:
This edge release adds support for the `config.linkerd.io/opaque-ports`
annotation on pods and namespaces, to configure ports that should skip
the proxy's protocol detection. In addition, it adds new CLI commands
related to the `linkerd-jaeger` extension, fixes bugs in the CLI
`install` and `upgrade` commands and Helm charts, and fixes a potential
false positive in the proxy's HTTP protocol detection. Finally, it
includes improvements in proxy performance and memory usage, including
an upgrade for the proxy's dependency on the Tokio async runtime.

* Added support for the `config.linkerd.io/opaque-ports` annotation on
  pods and namespaces, to indicate to the proxy that some ports should
  skip protocol detection
* Fixed an issue where `linkerd install --ha` failed to honor flags
* Fixed an issue where `linkerd upgrade --ha` can override existing
  configs
* Added missing label to the `linkerd-config-overrides` secret to avoid 
  breaking upgrades performed with the help of `kubectl apply --prune`
* Added a missing icon to Jaeger Helm chart
* Added new `linkerd jaeger check` CLI command to validate that the
  `linkerd-jaeger` extension is working correctly
* Added new `linkerd jaeger uninstall` CLI command to print the
  `linkerd-jaeger` extension's resources so that they can be piped into
  `kubectl delete`
* Fixed an issue where the `linkerd-cni` daemonset may not be installed
  on all intended nodes, due to missing tolerations to the `linkerd-cni`
  Helm chart (thanks @rish-onesignal!)
* Fixed an issue where the `tap` APIServer would not refresh its certs
  automatically when provided externally—like through cert-manager
* Changed the proxy's cache eviction strategy to reduce memory
  consumption, especially for busy HTTP/1.1 clients
* Fixed an issue in the proxy's HTTP protocol detection which could
  cause false positives for non-HTTP traffic
* Increased the proxy's default dispatch timeout to 5 seconds to
  accomodate connection pools which might open conenctions without
  immediately making a request
* Updated the proxy's Tokio dependency to v0.3

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,7 +22,7 @@ async runtime.
   `linkerd-jaeger` extension is working correctly
 * Added new `linkerd jaeger uninstall` CLI command to print the `linkerd-jaeger`
   extension's resources so that they can be piped into `kubectl delete`
-* Fixed an issue where the `linkerd-cni` daemonset may not be installed on all
+* Fixed an issue where the `linkerd-cni` daemgitonset may not be installed on all
   intended nodes, due to missing tolerations to the `linkerd-cni` Helm chart
   (thanks @rish-onesignal!)
 * Fixed an issue where the `tap` APIServer would not refresh its certs
@@ -35,6 +35,7 @@ async runtime.
   connection pools which might open conenctions without immediately making a
   request
 * Updated the proxy's Tokio dependency to v0.3
+
 ## edge-20.12.3
 
 This edge release is functionally the same as `edge-20.12.2`. It fixes an issue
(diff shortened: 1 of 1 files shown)


alpeb:
I've just merged #5407 , would you add an item for that? Something like "Added missing label to the `linkerd-config-overrides` secret to avoid breaking upgrades performed with the help of `kubectl apply --prune`"

alpeb (approved):
:ship:

olix0r (approved):

PR #5421, opened 2020-12-21
Add OpaqueTransport field to destination protocol hints

### LINKED ISSUE #5400: destination: Indicate support for connection headers
Now that the proxy has [initial support for reading connection headers](https://github.com/linkerd/linkerd2-proxy/pull/780), the outbound proxy needs a way to discover which streams should be transported with this header.

The destination service should provide an [`OpaqueHeader` protocol hint](https://github.com/linkerd/linkerd2-proxy-api/pull/53) when:
* The endpoint is injected with a proxy;
* And the endpoint includes an _opaque-port_ annotation that matches the resolved target port

The `OpaqueHeader` port value should be that of the target proxy's inbound address (i.e. 4143).

Part of #5384

kleimkuhler:
## What

When the destination service returns a destination profile for an endpoint,
indicate if the endpoint can receive opaque traffic.

## Why

Closes #5400

## How

When translating a pod address to a destination profile, the destination service
checks if the pod is controlled by any linkerd control plane. If it is, it can
set a protocol hint where we indicate that it supports H2 and opaque traffic.

If the pod supports opaque traffic, we need to get the port that it expects
inbound traffic on. We do this by getting the proxy container and reading it's
`LINKERD2_PROXY_INBOUND_LISTEN_ADDR` environment variable. If we successfully
parse that into a port, we can set the opaque transport field in the destination
profile.

## Testing

A test has been added to the destination server where a pod has a
`linkerd-proxy` container. We can expect the `OpaqueTransport` field to be set
in the returned destination profile's protocol hint.

### DIFF
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -21,7 +21,7 @@ const (
 	defaultWeight uint32 = 10000
 	// inboundListenAddr is the environment variable holding the inbound
 	// listening address for the proxy container.
-	inboundListenAddr = "LINKERD2_PROXY_INBOUND_LISTEN_ADDR"
+	envInboundListenAddr = "LINKERD2_PROXY_INBOUND_LISTEN_ADDR"
 )
 
 // endpointTranslator satisfies EndpointUpdateListener and translates updates
@@ -324,7 +324,7 @@ func toWeightedAddr(address watcher.Address, enableH2Upgrade bool, identityTrust
 				continue
 			}
 			for _, envVar := range containerSpec.Env {
-				if envVar.Name != inboundListenAddr {
+				if envVar.Name != envInboundListenAddr {
 					continue
 				}
 				addr := strings.Split(envVar.Value, ":")
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -319,6 +319,7 @@ func toWeightedAddr(address watcher.Address, enableH2Upgrade bool, identityTrust
 		// Get the inbound port from the proxy container's environment
 		// variable so that it can be set in the protocol hint.
 		var inboundPort uint32
+	loop:
 		for _, containerSpec := range address.Pod.Spec.Containers {
 			if containerSpec.Name != k8s.ProxyContainerName {
 				continue
@@ -333,6 +334,7 @@ func toWeightedAddr(address watcher.Address, enableH2Upgrade bool, identityTrust
 					log.Errorf("failed to parse inbound port for proxy container: %s", err)
 				}
 				inboundPort = uint32(port)
+				break loop
 			}
 		}
 		hint = &pb.ProtocolHint{
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -3,6 +3,8 @@ package destination
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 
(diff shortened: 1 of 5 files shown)


kleimkuhler:
Depends on linkerd/linkerd2-proxy-api#53

olix0r:
The API release has been tagged. This PR should be updated to reference https://github.com/linkerd/linkerd2-proxy-api/releases/tag/v0.1.16

controller/api/destination/endpoint_translator.go
```diff
@@ -15,7 +17,12 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-const defaultWeight uint32 = 10000
+const (
+	defaultWeight uint32 = 10000
+	// inboundListenAddr is the environment variable holding the inbound
+	// listening address for the proxy container.
+	inboundListenAddr = "LINKERD2_PROXY_INBOUND_LISTEN_ADDR"
```
olix0r, line 24: nit: perhaps `inboundListenAddrEnv` to make it clearer that it's the env variable and not the addr itself

controller/api/destination/endpoint_translator.go
```diff
@@ -299,19 +306,42 @@ func toAddr(address watcher.Address) (*net.TcpAddress, error) {
...
+			for _, envVar := range containerSpec.Env {
+				if envVar.Name != inboundListenAddr {
+					continue
+				}
+				addr := strings.Split(envVar.Value, ":")
+				port, err := strconv.ParseUint(addr[1], 10, 32)
+				if err != nil {
+					log.Errorf("failed to parse inbound port for proxy container: %s", err)
+				}
+				inboundPort = uint32(port)
```
olix0r, line 335: nit: you could `break` here to avoid looping through the rest of the environment; and similarly break after this for loop to avoid iteratating through remaining containers...

olix0r (approved):
looks good! just a few minor nits.

olix0r:
Correlated proxy changes available via https://github.com/linkerd/linkerd2/pull/5432. There are no requirements on merge order as the new proxy should work without the new annotation, and vice versa.

controller/api/destination/endpoint_translator.go
kleimkuhler: I went with `envInboundListenAddr` since we use a similar naming [here](https://github.com/linkerd/linkerd2/blob/main/pkg/identity/service.go#L29)

controller/api/destination/endpoint_translator.go
kleimkuhler: Good call.

controller/api/destination/endpoint_translator.go
```diff
@@ -299,19 +306,44 @@ func toAddr(address watcher.Address) (*net.TcpAddress, error) {
...
+				break loop
+			}
+		}
 		hint = &pb.ProtocolHint{
 			Protocol: &pb.ProtocolHint_H2_{
 				H2: &pb.ProtocolHint_H2{},
 			},
+			OpaqueTransport: &pb.ProtocolHint_OpaqueTransport{
+				InboundPort: inboundPort,
+			},
```
alpeb, line 346: Is this sent back, regardless of the `opaque-ports` annotation on the destination pod?

kleimkuhler: Yes that is correct. This field is set whether or not the endpoint should receive opaque traffic or not.

alpeb (approved):
LGTM :+1:

PR #5424, opened 2020-12-22
remove check comparing `ca.crt` field in identity issuer secret and trust anchors in config

### LINKED ISSUE #5292: Make `linkerd check` comprehend mulit-cert CA bundles

cypherfox:

This change remove the comparison between the trust anchors configured in the linkerd config map and the contents of the `ca.crt` field of the identity issuer K8s secret.

This is an alternative to MR #5396, which I will close as a result of the discussion with @adleong  

Fixes #5292 

Signed-off-by: Lutz Behnke <lutz.behnke@finleap.com>


kleimkuhler (approved):
This looks good, thanks @cypherfox!

adleong (approved):
Thanks @cypherfox!  Looks great.

PR #5433, opened 2020-12-23
viz: add render golden tests

Pothulapati:
This branch adds golden tests for the viz install. This would be
useful to track changes in render as more changes are added.


alpeb (commented):
It appears there are discrepancies in the golden files in some base64 encoded section...

olix0r:
Does this change provide an automated way to update golden files? In my opinion, this isn't really an optional feature, as these tests are otherwise extremely brittle and costly to update.

Pothulapati:
@olix0r Yep, `go test` with the `--update` flag updates these tests automatically. This is the same pattern that we followed the with the core control-plane render tests. WDYT?

olix0r:
> Yep, go test with the --update flag updates these tests automatically. 

@Pothulapati great, thanks!

pkg/charts/charts.go
```diff
@@ -175,3 +175,23 @@ func InsertVersionValues(values chartutil.Values) (chartutil.Values, error) {
 	}
 	return chartutil.ReadValues(InsertVersion([]byte(raw)))
 }
+
+// MergeMaps returns the resultant map after merging two maps
```
adleong, line 179: Can we flesh out this godoc with more details about the semantics of this method?  It should answer at least these questions:
* Are the inputs mutated?
* Which input takes precedence?
* How does this differ from other merge functions such as the one in mergo?

pkg/charts/charts.go
```diff
@@ -175,3 +175,23 @@ func InsertVersionValues(values chartutil.Values) (chartutil.Values, error) {
...
+func MergeMaps(a, b map[string]interface{}) map[string]interface{} {
+	out := make(map[string]interface{}, len(a))
+	for k, v := range a {
+		out[k] = v
+	}
+	for k, v := range b {
+		if v, ok := v.(map[string]interface{}); ok {
+			if bv, ok := out[k]; ok {
```
adleong, line 187: I think `bv` is maybe a confusing name for this variable since it's the value from input `a`.

viz/cmd/main_test.go
```diff
@@ -0,0 +1,87 @@
+package cmd
```
kleimkuhler, line 1: This whole file is pretty much the same as `cli/cmd/main_test.go`, right? Maybe it would make sense to move these functions into `pkg/` and just have the `main_test.go`s call into that?

kleimkuhler (approved):

testutil/test_helper.go
```diff
@@ -726,3 +727,58 @@ func ParseEvents(out string) ([]*corev1.Event, error) {
 
 	return events, nil
 }
+
+// DiffTestdata returns a diff with the given options
```
adleong, line 731: The comment on the `TestHelper` struct reads: `TestHelper provides helpers for running the linkerd integration tests.` which makes me feel like these unit test helper methods don't really belong in this file.

It also feels unwieldy to me to have this collection of global variables passed in to DiffTestdata everywhere.

Here's what I'd recommend:

Create a new struct in a new file called TestdataDiffer or TestdataHelper or something like that.  Have that struct hold configuration options like `prettyDiff` and `updateFixtures`.  Each `main_test.go` can then declare a single global TestdataHelper with the configuration from the flags.  Individual tests can then call `DiffTestdata` on that global TestdataHelper.

Pothulapati: Yeah, This makes total sense.

kleimkuhler: Yep good suggestion @adleong

adleong (approved):

Issue #5426, opened 2020-12-22
viz: make install wait till proxy-injector is up


Pothulapati:
## Feature Request

With #5340, Components are separated into a new viz extension, which takes a dependency on proxy injector to be up. Currently, It's the users job to make sure they are up and running.

### What problem are you trying to solve?

Make `viz install` wait till proxy-injector is up and running.

### How should the problem be solved?

Use similar semantics of that of wait with `linkerd dashboard`

PR #5434, opened 2020-12-23
viz: add a retry check for core control-plane pods before install

Pothulapati:
Fixes https://github.com/linkerd/linkerd2/issues/5426

This commit adds a new check so that `viz install` waits till
the control-plane pods are up. This follows the same pre-command
wait check model that we use for `dashboard`, etc commands.


adleong:
@Pothulapati it seems like there's 3 separate changes bundled together here:

* Remove the prometheus check in controller's SelfCheck method
* Some changes around waiting for the proxy-injector to start up
  * (I don't really understand those changes)
* Adding some conditions in linkerd-viz install to wait for the proxy-injector

It is possible to just address the first point in this PR to fix `linkerd check` and split the other changes out into a different PR?

Pothulapati:
@adleong Makes sense!

adleong (approved):

viz/cmd/install.go
```diff
@@ -155,3 +164,51 @@ func render(w io.Writer, valuesOverrides map[string]interface{}) error {
...
+
+		checkCmd := "linkerd check"
+		if controlPlaneNamespace != defaultLinkerdNamespace {
+			checkCmd += fmt.Sprintf(" --linkerd-namespace %s", controlPlaneNamespace)
+		}
+		fmt.Fprintf(os.Stderr, "Validate the install with: %s\n", checkCmd)
+
+		os.Exit(1)
+	}
+}
```
kleimkuhler, line 214: Not a blocker, but seems like something maybe worth putting in `pkg/`? This is the same as what is in `public_api.go`, so worth considering.

kleimkuhler (approved):

viz/cmd/install.go
Pothulapati: Yep, Definitely. There are other PRs having this same fn. Will update!

PR #5436, opened 2020-12-23
remove prom check for public-api self-check

Pothulapati:
Currently, public-api is part of the core control-plane where
the prom check fails when ran before the viz extension is installed.
This change comments out that check, Once metrics api is moved into
viz, maybe this check can be part of it instead or directly part of
`linkerd viz check`.


alpeb (approved):
LGTM :+1:  CI flakiness appears unrelated to this PR, I'll keep you posted about that.

kleimkuhler (approved):
👍

adleong (approved):

PR #5442, opened 2020-12-23
Update bin/web script to take into account linkerd-viz

alpeb:
The `get-pod` and `port-forward` functions continue to assume
deployments like grafana still live under the `linkerd` namespace.
This expands the definition of those functions to be able to specify the
namespace.

These changes can be solely tested by running `bin/web dev` (follow the
instructions in `BUILD.md` for the preliminaries needed).

CI is currently failing due to current instability in `main`.


adleong (commented):
Thanks @alpeb.  I'm not familiar with this script.  Does `bin/web port-forward` provide any substantial benefit over using `kubectl port-forward` directly?  Can we just get rid of this command?

alpeb:
> Thanks @alpeb. I'm not familiar with this script. Does `bin/web port-forward` provide any substantial benefit over using `kubectl port-forward` directly? Can we just get rid of this command?

It allows port-forwarding to a pod by just specifying the component name ("controller", "web", etc) and checks that that port-forward hasn't already been established. Probably not used a lot directly, but it's a dependency of `bin/web run`

adleong (approved):

olix0r:
fwiw recent kubectl versions support `kubectl port-forward -n linkerd svc/linkerd-web`

alpeb:
> fwiw recent kubectl versions support `kubectl port-forward -n linkerd svc/linkerd-web`

Yeah, I also discovered recently one's even able to use something like `deploy/linkerd-web` as a target.
Note that all the functions in this file are exposed as subcommands through some bash magic at the bottom, and since `port-forward()` is depended on elsewhere in the file, IMO it's fine to leave it as is.

olix0r (approved):

PR #5443, opened 2020-12-23
Dependabot alert fix: node-notifier upgrade

alpeb:
Upgrades `note-notifier` to `v8.0.1`, fixing [this
alert](https://github.com/linkerd/linkerd2/security/dependabot/web/app/yarn.lock/node-notifier/open).

This affects only the javascript unit tests (Jest). Nevertheless, the
dashboard was tested and no visible issues nor console errors were
detected.

CI might fail due to current `main` instability.

PR #5452, opened 2020-12-28
Update jest for node-notifier dependabot alert

cpretzer:
This updates the jest library after dependabot reported that the node-notifier dependency has a vulnerability.


alpeb (approved):
I pushed #5443 attempting the same by manually updating the yarn.lock, but it's definitely better just upgrade the Jest dependency, and have yarn.lock get updated automatically :+1:

adleong (approved):

olix0r:
FWIW https://github.com/linkerd/linkerd2/pull/5480 includes a full `yarn upgrade` (upgrading all deps), which includes this; though it primarily targets the latest chromebuilder/axios fixes.

kleimkuhler:
Addressed by #5480 which has been merged.

PR #5454, opened 2020-12-29
Added support for custom registries in the linkerd-viz helm chart

### LINKED ISSUE #5430: viz: support `--registry` flag
## Feature Request

Support a registry flag to override the registry of all images present in the viz extension. This is useful with private registries.

The core install already has a relevant flag, where that logic is done through the CLI and hence Go, But with Viz, It's important that this logic happens through Helm, so that this field is supported both through Helm and CLI.

jimil749:
Split the image `name` field in `viz/charts/linkerd-viz/values.yaml` into `name` and `registry` to support custom registries. Changed the template files accordingly.

Just like other values, the registry can now be configured via CLI via the `--set-*` flags.

Fixes #5430

Signed-off-by: Jimil Desai <jimildesai42@gmail.com>

### DIFF
--- a/viz/charts/linkerd-viz/values.yaml
+++ b/viz/charts/linkerd-viz/values.yaml
@@ -59,8 +59,10 @@ tap:
   # -- log level of the tap component
   logLevel: *log_level
   image:
+    # -- Docker registry for the grafana instance
+    registry: ghcr.io/linkerd
     # -- Docker image name for the grafana instance
-    name: ghcr.io/linkerd/controller
+    name: controller
     # -- Docker image tag for the grafana instance
     tag: *linkerd_version
   # -- Do not create a secret resource for the Tap component. If this is set to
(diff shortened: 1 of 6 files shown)


viz/charts/linkerd-viz/values.yaml
```diff
@@ -59,8 +59,10 @@ tap:
   # -- log level of the tap component
   logLevel: *log_level
   image:
+    # -- Docker registry for the grafana instance
```
alpeb, line 62: Should be "tap" (also needs to be corrected below for `name` and `tag`).

alpeb (commented):
Thanks @jimil749 :+1: 

Can you add a field with an anchor `defaultRegistry: &default_registry ghcr.io/linkerd` at the top section, so you can refer to it as `*default_regisry` on every `registry` entry to avoid repetition?

jimil749:
@alpeb, I've addressed all the changes! :)

adleong (approved):

alpeb (approved):
Thanks for the quick turnaround @jimil749 , this looks great to me :+1:

kleimkuhler:
Hi @jimil749! If you could please merge in the latest `main` it will pick up some fixes that were causing CI to fail, and then we'll be able to get this merged. Thanks!

kleimkuhler:
@jimil749 One check of CI seems to not have been kicked off properly.. not sure why. If you have a moment could you push an empty commit (`git commit --allow-empty -s`) to just trigger CI again?

jimil749:
@kleimkuhler all of the checks have passed now, so I think we can merge now! :)

kleimkuhler (approved):

PR #5458, opened 2020-12-29
Destination should return `OpaqueTransport` hint when annotation matches resolved target port

alpeb:
Followup to #5421

In the destination response, only return the `OpaqueTransport` hint if the associated pod has a `config.linkerd.io/opaque-ports` annotation that contains the resolved target port.

This was forbidding a client proxy to connect to a pod IP target; if the target pod was meshed, `OpaqueTransport` was always being set in the returned endpoint.

I also updated the unit tests, as we had an Endpoint with a port (8989) that didn't match the opaque-ports annotation (4242) so in that case no `OpaqueTransport` should be returned. And I added a new Service+Endpoint+Pod where the port matches and so `OpaqueTransport` does get returned.

For some reason this is still breaking the opaqueports integration test. Even though in my local testing I can see this in the client proxy logs:

```
DEBUG ThreadId(01) task{kind=task spawn.location=linkerd/app/src/lib.rs:384:9}:outbound:tcp.opaque:endpoint{peer.addr=10.42.0.49:8080 peer.id=Some("default.linkerd-opaque-ports-test.serviceaccount.identity.linkerd.cluster.local")}: linkerd2_app_outbound::tcp::opaque_transport: Using opaque transport header=Header { port: 8080, name: None } override_port=4143
```

The metrics in the target proxy are still showing the port is not being skipped:
```
tcp_open_total{peer="src",direction="inbound",tls="true",client_id="default.linkerd-opaque-ports-test.serviceaccount.identity.linkerd.cluster.local"} 1
```

### DIFF
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -330,7 +330,7 @@ func toWeightedAddr(k8sAPI *k8s.API, address watcher.Address, enableH2Upgrade bo
 			log.Errorf("failed getting opaque ports annotation for pod: %s", err)
 		} else {
 			for opaquePort := range opaquePorts {
-				if uint32(address.Port) == opaquePort {
+				if address.Port == opaquePort {
 					opaquePortMatches = true
 					break
 				}
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -219,9 +219,9 @@ func (et *endpointTranslator) sendClientAdd(set watcher.AddressSet) {
 			err error
 		)
 		if address.Pod != nil {
-			opaquePorts, err := getOpaquePortsAnnotations(et.k8sAPI, address.Pod)
-			if err != nil {
-				et.log.Errorf("failed getting opaque ports annotation for pod: %s", err)
+			opaquePorts, getErr := getOpaquePortsAnnotations(et.k8sAPI, address.Pod)
+			if getErr != nil {
+				et.log.Errorf("failed getting opaque ports annotation for pod: %s", getErr)
 			}
 			wa, err = toWeightedAddr(address, opaquePorts, et.enableH2Upgrade, et.identityTrustDomain, et.controllerNS, et.log)
 		} else {
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -328,34 +328,13 @@ func toWeightedAddr(address watcher.Address, opaquePorts map[uint32]struct{}, en
 				H2: &pb.ProtocolHint_H2{},
 			},
 		}
-		opaquePortMatches := false
-		for opaquePort := range opaquePorts {
-			if address.Port == opaquePort {
-				opaquePortMatches = true
-				break
-			}
-		}
-		if opaquePortMatches {
-			// Get the inbound port from the proxy container's environment
-			// variable so that it can be set in the protocol hint.
-		loop:
-			for _, containerSpec := range address.Pod.Spec.Containers {
-				if containerSpec.Name != pkgk8s.ProxyContainerName {
-					continue
-				}
-				for _, envVar := range containerSpec.Env {
-					if envVar.Name != envInboundListenAddr {
-						continue
-					}
-					addr := strings.Split(envVar.Value, ":")
-					port, err := strconv.ParseUint(addr[1], 10, 32)
-					if err != nil {
-						log.Errorf("failed to parse inbound port for proxy container: %s", err)
-					}
-					hint.OpaqueTransport = &pb.ProtocolHint_OpaqueTransport{
-						InboundPort: uint32(port),
-					}
-					break loop
+		if _, ok := opaquePorts[address.Port]; ok {
+			port, err := getInboundPort(&address.Pod.Spec)
+			if err != nil {
+				log.Error(err)
+			} else {
+				hint.OpaqueTransport = &pb.ProtocolHint_OpaqueTransport{
+					InboundPort: port,
 				}
 			}
 		}
@@ -419,4 +398,26 @@ func newEmptyAddressSet() watcher.AddressSet {
 		Labels:          make(map[string]string),
 		TopologicalPref: []string{},
 	}
+}
+
+// getInboundPort gets the inbound port from the proxy container's environment
+// variable.
+func getInboundPort(podSpec *corev1.PodSpec) (uint32, error) {
+	for _, containerSpec := range podSpec.Containers {
+		if containerSpec.Name != pkgk8s.ProxyContainerName {
+			continue
+		}
+		for _, envVar := range containerSpec.Env {
+			if envVar.Name != envInboundListenAddr {
+				continue
+			}
+			addr := strings.Split(envVar.Value, ":")
+			port, err := strconv.ParseUint(addr[1], 10, 32)
+			if err != nil {
+				return 0, fmt.Errorf("failed to parse inbound port for proxy container: %s", err)
+			}
+			return uint32(port), nil
+		}
+	}
+	return 0, fmt.Errorf("failed to find %s environment variable in any container for given pod spec", envInboundListenAddr)
 }
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -9,8 +9,9 @@ import (
 	pb "github.com/linkerd/linkerd2-proxy-api/go/destination"
 	"github.com/linkerd/linkerd2-proxy-api/go/net"
 	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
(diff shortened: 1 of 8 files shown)


kleimkuhler (commented):
It looks like the bulk of these changes are related to passing in the `k8sAPI` to `endpointTranslator` so that we can get the opaque ports annotation on a pod in `toWeightedAddr`.

I don't think this is necessary since we get the opaque ports annotation from pod directly after the call to `toWeightedAddr`. There is no reason for the ordering of that right now, so I think we can just pass the opaque ports annotation into this function and back out some of these changes.

Additionally, the test failure is real because before things were wired through in the proxy, we would expect proxy-to-proxy communication to not be TLS'd for `port_skipped` reason. Now that that works, we can change the test to expect an open TCP connection from the expected identity.

I'll push a commit to make these changes.

kleimkuhler:
`endpointTranslator` still needs the k8s API passed through because of it's use in `sendClientAdd`; we need to check pod annotations on newly added addresses so we still need a way to query the k8s API for that. 

However, I separated getting the opaque ports annotation out of `toWeightedAddr` so that it can be passed in. This simplifies the case in `GetProfile` a little.

controller/api/destination/endpoint_translator.go
```diff
@@ -306,45 +313,52 @@ func toAddr(address watcher.Address) (*net.TcpAddress, error) {
...
+					}
+					addr := strings.Split(envVar.Value, ":")
+					port, err := strconv.ParseUint(addr[1], 10, 32)
+					if err != nil {
+						log.Errorf("failed to parse inbound port for proxy container: %s", err)
+					}
+					hint.OpaqueTransport = &pb.ProtocolHint_OpaqueTransport{
+						InboundPort: uint32(port),
+					}
+					break loop
```
olix0r, line 358: tioli: Is it worth splitting this out into a helper like `getInboundPort(pod *Spec) (uint32, error)`?

kleimkuhler: I kind of lean towards leaving it, but I can move it out if that helps! We only use this logic here, so moving it it's own function doesn't remove duplication. Also, if someone is reading through this to understand how this all works and saw `getInboundPort`, I think they'd still need to jump to that definition to see exactly what was happening. So for both of those reasons, I think it's okay being here.

adleong: :+1: Any time I see labeled loops it makes me think about if breaking out into a separate function is better.

controller/api/destination/endpoint_translator.go
```diff
@@ -306,45 +313,52 @@ func toAddr(address watcher.Address) (*net.TcpAddress, error) {
...
-		for _, containerSpec := range address.Pod.Spec.Containers {
-			if containerSpec.Name != k8s.ProxyContainerName {
-				continue
+		hint = &pb.ProtocolHint{
+			Protocol: &pb.ProtocolHint_H2_{
+				H2: &pb.ProtocolHint_H2{},
+			},
+		}
+		opaquePortMatches := false
+		for opaquePort := range opaquePorts {
```
olix0r, line 332: Since `opaquePorts` is a map, why iterate through all ports. Can we not do a simple lookup like `opaquePorts[address.Port]`?

controller/api/destination/endpoint_translator.go
kleimkuhler: Sounds good! I will move it out.

controller/api/destination/endpoint_translator.go
kleimkuhler: Yep we can do that. I removed `opaquePortMatches` completely from this as well.

adleong (commented):
What do you think about updating `portPublisher.endpointsToAddresses` to copy relevant annotations (in this case the opaque port annotation) from the namespace onto each pod (merging the opaque port lists if necessary)?  This keeps all synchronous k8s lookups in one place (the watcher) and we don't need to give the translator an api client.  

It also means fewer namespace lookups (one per service instead of one per pod), though this probably isn't a big deal because of client-go caching.

olix0r:
> What do you think about updating portPublisher.endpointsToAddresses to copy relevant annotations (in this case the opaque port annotation) from the namespace onto each pod (merging the opaque port lists if necessary)?

+1 I like this idea. It's also more correct: the namespace's value can change after the pod is created, and our lookup against the namespace at runtime could produce incorrect results. If we persist it onto the workload, we can't get into that state.

kleimkuhler:
> What do you think about updating `portPublisher.endpointsToAddresses` to copy relevant annotations (in this case the opaque port annotation) from the namespace onto each pod (merging the opaque port lists if necessary)?

I like that with this change the endpoint translator would not need the k8s API.

I do want to point out though that right now, if there is an opaque ports annotation on the namespace and the pod, the pod completely overrides the namespace. This behavior was made kind of arbitrarily, but this change would change that behavior. We would now be merging the two resource annotations.

This change should then also be made in `GetProfile` where we are dealing with an IP and [get its annotations](https://github.com/linkerd/linkerd2/blob/5c41936fb18fd719d8826ec1800dba3f29c7c984/controller/api/destination/server.go#L211).

kleimkuhler:
After some further discussion, the merging of pod and namespace annotations has been removed. If there is no opaque ports annotation on the pod, then it inherits the namespace annotation. If there is no namespace annotation, then it remains unset.

controller/api/destination/watcher/endpoints_watcher.go
```diff
@@ -1081,3 +1085,22 @@ func isValidSlice(es *discovery.EndpointSlice) bool {
 
 	return true
 }
+
+// SetPodOpaquePortAnnotation ensures that if there is no opaque port
+// annotation on the pod, then it inherits the annotation from the namespace.
+// If there is also no annotation on the namespace, then it remains unset.
+func SetPodOpaquePortAnnotation(k8sAPI *k8s.API, pod *corev1.Pod, ns string) error {
+	annotation := pod.Annotations[consts.ProxyOpaquePortsAnnotation]
```
adleong, line 1093: Do we want to condition on the absence of this annotation instead of it being the empty string?  e.g. if the pod sets this annotation to the empty string, shouldn't that take priority over the namespace annotation and essentially unset all the opaque ports?

controller/k8s/api.go
```diff
@@ -1066,6 +1066,26 @@ func (api *API) GetServiceProfileFor(svc *corev1.Service, clientNs, clusterDomai
 	}
 }
 
+// GetNsAnnotationFor gets the value of an annotation for a namespace.
+func (api *API) GetNsAnnotationFor(name string, annotation string) (string, error) {
+	obj, err := api.GetObjects("", k8s.Namespace, name, labels.Everything())
```
adleong, line 1071: Why use `GetObjects` instead of just `api.NS().Lister().Get`? This avoids needing to cast.

controller/api/destination/endpoint_translator.go
```diff
@@ -216,7 +216,11 @@ func (et *endpointTranslator) sendClientAdd(set watcher.AddressSet) {
 			err error
 		)
 		if address.Pod != nil {
-			wa, err = toWeightedAddr(address, et.enableH2Upgrade, et.identityTrustDomain, et.controllerNS, et.log)
+			opaquePorts, getErr := getOpaquePortsAnnotations(address.Pod)
+			if getErr != nil {
+				et.log.Errorf("failed getting opaque ports annotation for pod: %s", getErr)
+			}
+			wa, err = toWeightedAddr(address, opaquePorts, et.enableH2Upgrade, et.identityTrustDomain, et.controllerNS, et.log)
```
adleong, line 223: I think we always call `getOpaquePortsAnnotations` prior to calling `toWeightedAddr`, right?  Why not just move the call to `getOpaquePortsAnnotation` into `toWeightedAddr`.  Then we don't need to pass the opaquePorts as an argument.

adleong (commented):
Just a few small comments, this is looking good.

controller/api/destination/endpoint_translator.go
kleimkuhler: The reason these are separated is because in [GetProfile](https://github.com/linkerd/linkerd2/blob/6afe9bef20366a87e6d9602200c830cc271ff933/controller/api/destination/server.go#L215), we use the value of `opaquePorts` later.

Separating the calls saves us a call to `getOpaquePortsAnnotations` later in that function. Before this was a little more expensive since it involved an k8s API call which is no longer the case, but it still does some string parsing and all.

controller/api/destination/watcher/endpoints_watcher.go
kleimkuhler: Yep good point.

adleong (approved):
🐱 🥛 🖌️

olix0r (approved):
awesome!

PR #5463, opened 2020-12-31
Enable setting gateway addresses from ExternalIPs or ExternalName

### LINKED ISSUE #5460: support external load balancer for multi-cluster link
## Feature Request
I am trying to link bare-metal clusters that do not have an integrated "LoadBalancer" provider but have perfectly functional external TCP load balancers.  I need to be able to do one or more of the following:

1. Tell `linkerd multicluster link` get the IP for the service from `spec.loadBalancerIP` or `spec.externalIPs` instead of `status.loadBalancer.ingress.ip`.  Both of these properties can be user defined for use with external load balancers.
2. Be able to provide a command line switch to specify the ip manually

Use multi-cluster on bare-metal with external load balancers

I've considered installing MetalLB, but it wont work with my clusters using calico without over-the-top BGP configurations.

Ideally:
`linkerd multicluster install --external-ip 10.4.3.2`

This would set the `spec.externalIPs` and  `spec.loadBalancerIP` on the service, the standard multicluster link command would then see no `status.loadBalancer.ingress.ip` and would fall-back to using the externalIP for the generated YAML.  This would be compatible with any scenario I can imagine.

Alternatively:
`linkerd --context east multicluster link --cluster-name east --ingress-ip 10.4.3.2`
This has the advantage of only a single area of change in the CLI.

### Thanks
If I was halfway decent with Go, I would offer a PR for the CLI.  I'll avoid embarrassing myself though.

### LINKED ISSUE #5462: Avoid multi-cluster LoadBalancer Service requirement

LoadBalancer type Services may not be supported or allowed in some Kubernetes clusters - for example due to:
* Being on-prem or in a lab with no "load-balancer as a service" capability
* Security concerns around exposing Services externally
* Policy that requires all incoming traffic to go via ingress proxies

Not all clusters support LoadBalancer Services but all should support ExternalName.  I have clusters where I can't use Linkerd multi-cluster because of the LoadBalancer requirement.

Support the use of ExternalName Service objects as suggested in #4590.

Linkerd multicluster already supports LoadBalancer Services where the gateway address is a hostname (#4588):

> When a target cluster gateway is exposed as a hostname rather than with a fixed IP address, the service mirror controller fails to create mirror services and gateway mirrors for that gateway. This is because we only look at the IP field of the gateway service.
> 
> We make two changes to address this problem:
> 
> First, when extracting the gateway spec from a gateway that has a hostname instead of an IP address, we do a DNS lookup to resolve that hostname into an IP address to use in the mirror service endpoints and gateway mirror endpoints.
> 
> Second, we schedule a repair job on a regular (1 minute) to update these endpoint objects. This has the effect of re-resolving the DNS names every minute to pick up any changes in DNS resolution.

Using ExternalName Services can be supported in a similar way: if the Service is of type ExternalName then get the hostname from the spec.externalName field rather than the status.loadBalancer.ingress[0].hostname field.

It looks like this should be a simple change around https://github.com/linkerd/linkerd2/blob/main/multicluster/cmd/link.go#L163 and updating some tests?

Implementing a cluster controller manager to make LoadBalancer Services work is an option where there is no external load balancer service but is a significant effort to solve this simple use case.  The security and policy issues are not easy to work around so there would likely need to be some other way to set the gateway IP, perhaps by writing a custom controller.

https://github.com/linkerd/linkerd2/issues/4590#issue-637617178

erwbgy:
Problem: LoadBalancer type Services may not be supported or allowed in some Kubernetes clusters, so fetching the gateway address from the status.loadBalancer.ingress field of the Service object doesn't work.

Solution: Enable setting the gateway address using the spec.ExternalName field of the Service object as this can be set without LoadBalancer support.

Fixes #5462

Signed-off-by: Keith Burdis <keith.burdis@gs.com>

### DIFF
--- a/multicluster/cmd/link.go
+++ b/multicluster/cmd/link.go
@@ -171,12 +171,8 @@ func newLinkCommand() *cobra.Command {
 				}
 				gatewayAddresses = append(gatewayAddresses, addr)
 			}
-			if len(gatewayAddresses) == 0 {
-				if len(gateway.Spec.ExternalIPs) > 0 {
-					gatewayAddresses = append(gatewayAddresses, gateway.Spec.ExternalIPs...)
-				} else if gateway.Spec.ExternalName != "" {
-					gatewayAddresses = append(gatewayAddresses, gateway.Spec.ExternalName)
-				}
+			if len(gatewayAddresses) == 0 && gateway.Spec.ExternalName != "" {
+				gatewayAddresses = append(gatewayAddresses, gateway.Spec.ExternalName)
 			}
 			if len(gatewayAddresses) == 0 {
 				return fmt.Errorf("Gateway %s.%s has no ingress addresses", gateway.Name, gateway.Namespace)
--- a/multicluster/cmd/link.go
+++ b/multicluster/cmd/link.go
@@ -172,6 +172,13 @@ func newLinkCommand() *cobra.Command {
 				gatewayAddresses = append(gatewayAddresses, addr)
 			}
 			if len(gatewayAddresses) == 0 {
+				if len(gateway.Spec.ExternalIPs) > 0 {
+					gatewayAddresses = append(gatewayAddresses, gateway.Spec.ExternalIPs...)
+				} else if gateway.Spec.ExternalName != "" {
+					gatewayAddresses = append(gatewayAddresses, gateway.Spec.ExternalName)
+				}
+			}
+			if len(gatewayAddresses) == 0 {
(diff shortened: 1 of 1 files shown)


adleong:
Thanks for putting this together, @erwbgy.  I've left a comment on the attached issue: https://github.com/linkerd/linkerd2/issues/5462

kleimkuhler:
It looks like there is still some discussion going on in #5462 that may affect the changes in this PR so I'm going to mark it as draft for now. Once there is a clear path forward @erwbgy feel free to make it reviewable again!

erwbgy:
Closing this in favour of #5471.

Issue #5401, opened 2020-12-17
Make service profile generation work offline


purkhusid:
Currently the service profile generation from proto/swagger requires that you have access to the k8s cluster that has Linkerd running.

The only reason for this is that currently the cluster domain is fetched from the Linkerd installation on the cluster.

I propose that the `profile` command be updated to support the `--ignore-cluster` flag
I also propose that there should be a flag `--cluster-domain` that allows the user to override the cluster domain. 

See discussion here: https://github.com/linkerd/linkerd2/discussions/5287#discussioncomment-187807

cpretzer:
@purkhusid thanks for pointing this out. We'd love your help with this, if you're up for submitting a PR

piyushsingariya:
Hi, I would like to work on this issue!

piyushsingariya:
Hi @cpretzer, I edited the `cli/cmd/profile.go ` where basically I added code to validate `--ignore-cluster`, added a new boolean variable `ignoreCluster` default to be `false` in profileOptions and now we will call a function to create a profile
and I think the step after this is to create a new file in `pkg/profiles` -> ignoreCluster.go where function(creates a profile without checking the access) will be written
this step is a bit challenging for me if you could guide me here, it would be a big help.

piyushsingariya:
Hi, @purkhusid I have started working on this. It would be helpful if you could explain what are the other flags and args which can be passed alongside the flag `--ignore-cluster`

wmorgan:
@piyushsingariya just wanted to say thanks again for fixing this. This was a pet peeve of mine.

wmorgan:
(And thanks @purkhusid for creating the issue, which I never got around to doing!)

PR #5482, opened 2021-01-06
service profile generation work offline using --ignore-cluster

piyushsingariya:
Subject
Related to issue #5401

Problem
Introduce offline profile generation to skip the access check to k8s cluster

Solution
the profile command be updated to support the --ignore-cluster flag to skip the check and generate a profile offline
Introduced a new boolean variable `ignoreCluster` to trigger offline profile generation
 
Fixes #[GitHub issue ID]
#5401


.gitignore
```diff
@@ -16,3 +16,4 @@ vendor
 **/*.swp
 **/charts/**/charts
 package-lock.json
+.vscode
```
adleong, line 19: please add a newline to the end of this file.

piyushsingariya:
@adleong Cool, I am working on this :)

cli/cmd/profile.go
```diff
@@ -94,25 +96,33 @@ func newCmdProfile() *cobra.Command {
...
-			if err != nil {
-				return err
-			}
-
-			clusterDomain := values.GetGlobal().ClusterDomain
-			if clusterDomain == "" {
-				clusterDomain = defaultClusterDomain
+			// performs an online profile generation and access-check to k8s cluster to extract
+			// clusterDomain from linkerd configuration
+			// profile generation based on tap data requires access to k8s cluster
```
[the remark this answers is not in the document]
piyushsingariya: That's a good idea, I have implemented a check to return an error if --tap and --ignore-cluster are used together.

cli/cmd/profile.go
```diff
@@ -94,25 +96,33 @@ func newCmdProfile() *cobra.Command {
...
+			if !options.ignoreCluster || options.tap != "" {
+				localk8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
```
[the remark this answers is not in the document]
piyushsingariya: Yes, I was facing a run-time error because of variable scope. Now, that you have suggested I implemented it as it would be much better to write like this

cli/cmd/profile.go
```diff
@@ -94,25 +100,33 @@ func newCmdProfile() *cobra.Command {
...
+
+				_, values, err := healthcheck.FetchCurrentConfiguration(cmd.Context(), k8sAPI, controlPlaneNamespace)
+				if err != nil {
+					return err
+				}
+
+				clusterDomain = values.GetGlobal().ClusterDomain
+				if clusterDomain == "" {
+					clusterDomain = defaultClusterDomain
+				}
```
kleimkuhler, line 129: So we already set `clusterDomain` to `defaultClusterDomain` above. Therefore, I think we should set it only if the cluster's domain is not empty:

```suggestion
				if cd := values.GetGlobal().ClusterDomain; cd != "" {
					clusterDomain = cd
				}
```

cli/cmd/profile.go
```diff
@@ -55,6 +57,10 @@ func (options *profileOptions) validate() error {
 		return errors.New("You must specify exactly one of --template or --open-api or --proto or --tap")
 	}
 
+	// service profile generation based on tap data requires access to k8s cluster
+	if options.ignoreCluster && options.tap != "" {
+		return errors.New("Service profile generation based on tap data require access check to k8s cluster. You can not use --ignore-cluster here")
```
kleimkuhler, line 62: This is a bit of a nit, but it feels a little wordy. How about something like:

```suggestion
		return errors.New(""--ignore-cluster and --tap flags are mutually exclusive; SP generation based on tap data requires access-check to k8s cluster")
```

This isn't exactly shorter, but we use similar wording [here](https://github.com/linkerd/linkerd2/blob/cbcc305b78edb9fe65fc0edee99a5722a949287b/cli/cmd/stat.go#L741-L747) and still provide a little explanation about why.

kleimkuhler (commented):
I have two smaller comments but rest is looking good. Thanks @piyushsingariya!

cli/cmd/profile.go
piyushsingariya: That's right :smile_cat: 
Working on this right now!

cli/cmd/profile.go
piyushsingariya: I felt the same while writing this :)

kleimkuhler (approved):
Thanks @piyushsingariya 👍

adleong (approved):
Great work!

piyushsingariya:
Thanks, @adleong @kleimkuhler  :smile:

purkhusid:
Awesome work @piyushsingariya!

piyushsingariya:
Thanks! @purkhusid :sweat_smile:

PR #5485, opened 2021-01-06
viz: move sub-cmds using viz extension under viz cmd

### LINKED ISSUE #5327: Move viz CLI commands to linkerd viz subcommand
Move the following viz-related commands to the `linkerd viz` subcommand:

* linkerd viz stat
* linkerd viz routes
* linkerd viz tap
* linkerd viz top
* linkerd viz dashboard
* linkerd viz edges

For now, invoking the old commands (e.g. linkerd stat) should print a warning but continue to function.

These commands should also validate that the linkerd-viz extension is installed and print a helpful message if not.

Pothulapati:
Fixes #5327

This branch moves the following commands, under the `linkerd viz`
cmd as they use the viz extension to perform the job.

- dashboard
- edges
- routes
- stat
- tap
- top

This also creates a new pkg `public-api` which fecilitates
interaction and communication with public-api to be used
across extensions.


viz/cmd/main_test.go
```diff
@@ -0,0 +1,87 @@
...
+	flag.BoolVar(&updateFixtures, "update", false, "update text fixtures in place")
+	prettyDiff = os.Getenv("LINKERD_TEST_PRETTY_DIFF") != ""
+	flag.BoolVar(&prettyDiff, "pretty-diff", prettyDiff, "display the full text when diffing")
+	flag.StringVar(&rejectPath, "reject-path", "", "write results for failed tests to this path (path is relative to the test location)")
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
+// readTestdata reads a file and returns the contents of that file as a string.
+func readTestdata(t *testing.T, fileName string) string {
```
alpeb, line 35: Can you move this, and the functions below somewhere under `testutil` to avoid the repetition with `cli/cmd/main_test.go`?

alpeb (commented):
Sweet! :muscle: 

Good call on moving the public-api functions to `pkg` :+1:  Nevertheless, in light of taking advantage of the full import path when naming packages, how about instead of `pkg/publicapi/publicapi.go` we do `pkg/public/api.go` (and thus the package would be just `api`)?

#5327 mentions having the old command show a warning but still work. I can't think of an easy way of doing that without leaving a lot of duped code around (could Cobra have something to avoid that?). IMO I think it would be enough to show a warning pointing to the new command...

Also, it seems some golden files need to be moved :wink:

Pothulapati:
> how about instead of pkg/publicapi/publicapi.go we do pkg/public/api.go (and thus the package would be just api)?

Seems like the right way to do this!

> Also, it seems some golden files need to be moved 😉

Waiting for #5433 to merge first, as that would make this easier.

> #5327 mentions having the old command show a warning but still work. I can't think of an easy way of doing that without leaving a lot of duped code around (could Cobra have something to avoid that?). IMO I think it would be enough to show a warning pointing to the new command...

We *might* be able to make the old cmd work, by exposing these `AddCmd` commands, and attaching them directly in the main CLI. Will try this out today! Adding a warning may not be that hard if this works. I'm against code duped code, unless its our only option!

cli/cmd/public_api.go
```diff
@@ -51,24 +52,12 @@ func checkPublicAPIClientOrRetryOrExit(retryDeadline time.Time, apiChecks bool)
...
-		KubeConfig:            kubeconfigPath,
-		KubeContext:           kubeContext,
-		Impersonate:           impersonate,
-		ImpersonateGroup:      impersonateGroup,
-		APIAddr:               apiAddr,
-		RetryDeadline:         retryDeadline,
-	})
-}
-
 func exitOnError(result *healthcheck.CheckResult) {
```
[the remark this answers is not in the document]
Pothulapati: Yep, in the above function. There is a repetition of this fn though. I will try to see if we can reuse.

cli/cmd/public_api.go
```diff
@@ -30,18 +30,19 @@ func rawPublicAPIClient(ctx context.Context) (pb.ApiClient, error) {
 	return public.NewExternalClient(ctx, controlPlaneNamespace, kubeAPI)
 }
 
-// checkPublicAPIClientOrExit builds a new public API client and executes default status
+// CheckPublicAPIClientOrExit builds a new public API client and executes default status
 // checks to determine if the client can successfully perform cli commands. If the
 // checks fail, then CLI will print an error and exit.
-func checkPublicAPIClientOrExit() public.APIClient {
-	return checkPublicAPIClientOrRetryOrExit(time.Time{}, false)
+func CheckPublicAPIClientOrExit(hcOptions healthcheck.Options) public.APIClient {
```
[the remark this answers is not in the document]
Pothulapati: still being used! 🤔

alpeb:
I think the integration test failure for tap occurs because it's the only test that exercises changing the default namespace for the current k8s context. It seems this refactoring doesn't include the `init()` logic from `cli/cmd/root.go` that sets `defaultNamespace` to the current context default namespace.

adleong:
I haven't tried this, but can [cmd.Find](https://godoc.org/github.com/spf13/cobra#Command.Find) be used to delegate to a subcommand?

e.g. in the body of `linkerd stat`:
```
subcmd, _, err := cmd.Find(append([]string{"viz"}, args..)
if err != nil {
  return err
}
err = subcmd.ExecuteContext(cmd.Context())
```

or something like that

adleong (commented):
Looks good once @alpeb's suggestion is addressed.

cli/cmd/root.go
```diff
@@ -99,7 +92,7 @@ var RootCmd = &cobra.Command{
 }
 
 func init() {
-	defaultNamespace = getDefaultNamespace()
+	defaultNamespace = cmd.GetDefaultNamespace(kubeconfigPath, kubeContext)
```
alpeb, line 95: Can you check if changing `kubeconfigPath`, pointing to a config with a non default ns, would get properly detected here? At this point `kubeconfigPath` is an empty string, passed by value, and I think the Cobra command below won't affect it. If so, this is an existing bug that that can be addressed in a separate PR.

cli/cmd/root.go
```diff
@@ -136,86 +123,14 @@ func init() {
...
-	}
 
-	return ns
+	// Viz Extension sub commands
+	RootCmd.AddCommand(viz.NewCmdDashboard())
+	RootCmd.AddCommand(viz.NewCmdEdges())
+	RootCmd.AddCommand(viz.NewCmdRoutes())
+	RootCmd.AddCommand(viz.NewCmdStat())
+	RootCmd.AddCommand(viz.NewCmdTap())
+	RootCmd.AddCommand(viz.NewCmdTop())
```
alpeb, line 133: In order to mark these subcommands as deprecated, you can do something like this for each one: 
```go
cmdRoutes := viz.NewCmdRoutes()
cmdRoutes.Deprecated = "use instead 'linkerd viz routes'"
RootCmd.AddCommand(cmdRoutes)
```

alpeb (commented):
Looking really nice, glad to see the change ended up being not too complex :+1:

alpeb (approved):
LGTM :+1:

cli/cmd/root.go
Pothulapati: Created https://github.com/linkerd/linkerd2/issues/5523

viz/cmd/root.go
```diff
@@ -29,6 +36,10 @@ var (
 	alphaNumDash = regexp.MustCompile(`^[a-zA-Z0-9-]+$`)
 )
 
+func init() {
+	defaultNamespace = cmd.GetDefaultNamespace(kubeconfigPath, kubeContext)
```
adleong, line 40: I think this has the same issue as https://github.com/linkerd/linkerd2/issues/5524 where `kubeconfigPath` and `kubeContext` will be empty when `init()` is called because the flags have not been parsed yet.

alpeb (approved):
I confirmed this closes #5524 :+1:

adleong (approved):

PR #5486, opened 2021-01-06
Enable dashboard test

### LINKED ISSUE #5478: Dashboard test always fails because web component is no longer installed by default
Once `linkerd viz dashboard` has been added and the web component can be reliably tested, the test should be reenabled.

kleimkuhler:
This test was never broken. My best guess is that CI was not merging with the
latest `main` as we have recently noticed, so this was an issue that was fixed
by #5458

Closes #5478


olix0r (approved):

olix0r:
Test still fails with
```
        cmd [/home/runner/.linkerd --linkerd-namespace linkerd --context= dashboard -p 52237 --show url] Timed out trying to read 4 lines
```
Is there still an issue on `main`?

alpeb (approved):

PR #5490, opened 2021-01-06
Add support for global deployment annotations

### LINKED ISSUE #5489: Add support for global deployment annotations
When secrets and configmaps change in a cluster via https://linkerd.io/2/tasks/automatically-rotating-control-plane-tls-credentials/, pods don't restart with the new secret/configmap information and therefore end up using stale data.

[Stakater/Reloader](https://github.com/stakater/Reloader) supports restarting deployments based on annotations - it watches deployments with the proper annotations and triggers a reload when it notices any changes in the configmaps/secrets the deployment are using.

Support adding custom annotations to deployments.

It's possible there are other ways to solve the reloading problem, but it would be nice to be able to add annotations to deployments so we can use the current reloader solution (and so other users can add their own desired annotations).

Users can add their desired annotations to their own values file under the `global.deploymentAnnotations` map.

gabriellemadden:
When secrets and configmaps change in a cluster via
https://linkerd.io/2/tasks/automatically-rotating-control-plane-tls-credentials/,
pods don't restart with the new secret/configmap information and
therefore end up using stale data.

The mechanism we use to restart deployments requires the use of
an annotation in the deployment metadata. This change enables users to
add their deployment annotations to the values.yaml.

This chart was deployed in a test cluster, and it was confirmed that
the annotations we added to our values.yaml file were propagated to the
deployments. Changing configmap/secret data resulted in deployments
restarting, as expected.

Fixes #5489

Signed-off-by: Gabrielle Madden <gabriellemadden333@gmail.com>

### DIFF
--- a/charts/linkerd2/README.md
+++ b/charts/linkerd2/README.md
@@ -141,6 +141,7 @@ Kubernetes: `>=1.13.0-0`
 | global.controllerLogLevel | string | `"info"` | Log level for the control plane components |
 | global.controllerNamespaceLabel | string | `"linkerd.io/control-plane-ns"` | Control plane label. Do not edit   |
 | global.createdByAnnotation | string | `"linkerd.io/created-by"` | Annotation label for the proxy create. Do not edit.  |
+| global.deploymentAnnotations | object | `{}` | Additional annotations to add to all deployments |
 | global.enableEndpointSlices | bool | `false` | enables the use of EndpointSlice informers for the destination service; enableEndpointSlices should be set to true only if EndpointSlice K8s feature gate is on; the feature is still experimental. |
 | global.grafanaUrl | string | `""` | url of external grafana instance with reverse proxy configured. |
 | global.identityTrustAnchorsPEM | string | `""` | Trust root certificate (ECDSA). It must be provided during install.  |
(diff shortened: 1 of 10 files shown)


olix0r:
Thanks for the submission, @gabriellemadden!

I'm a little unclear on what problem this solves.

Linkerd control plane components already have a standard set of metadata:
```
  annotations:
    linkerd.io/created-by: linkerd/cli edge-20.12.1
  labels:
    app.kubernetes.io/name: destination
    app.kubernetes.io/version: edge-20.12.1
    linkerd.io/control-plane-component: destination
```

Are these suitable for your needs? If not, can you help us understand where they fall short?

If you really need some custom metadata for your workloads, have you considered using an external tool like `kustomize` to modify linkerd's manifests?

adleong:
I'm marking this PR as a draft until the questions on https://github.com/linkerd/linkerd2/issues/5489 are addressed.

gabriellemadden:
Hey @olix0r, 
> Are these suitable for your needs? If not, can you help us understand where they fall short?

> If you really need some custom metadata for your workloads, have you considered using an external tool like `kustomize` to modify linkerd's manifests?

These annotations do not include the stakater annotation we are using (for obvious reasons). We need the stakater annotation in order to tell the proxy-injector pods to restart when we rotate its secrets, so it would be nice to be able to add custom annotations in.

The problem with using `kustomize` is that we are using ArgoCD to manage resources in our cluster. We wish to maintain all our infrastructure as code, which means that we have chart definitions in a repository that ArgoCD is installing automatically for us.  Modifying and then reapplying yaml doesn't work because ArgoCD will wipe the applied changes out of the cluster since the changes aren't present in the current chart definition as tracked by version control.

We have gotten around this in the past by copying the entire helm chart into the branch ArgoCD is tracking and making changes there, or by forking helm charts and hosting our own customized versions, but this becomes increasingly difficult to maintain and we'd prefer to stick to as many upstream charts as we can.

Issue #5325, opened 2020-12-03
Add linkerd viz check command


adleong:
Create a `linkerd viz check` command which checks the health of the linkerd-viz extension.  Any checks from the main `linkerd check` which relate to the linkerd-viz components should be moved here.  In particular it should check:

* verify metrics-api can call k8s API and prometheus
  * (uses the SelfCheck method in the metrics-api)
* verify tap and tap-injector webhook certificates are healthy
* in --proxy mode: check that metrics from proxies exist in prometheus
  * (uses the ListPods method in the metrics-api)

adleong:
In `--proxy` mode it should also check that all pods which have a linkerd proxy sidecar also have the appropriate tap environment variables set.

adleong:
https://github.com/linkerd/linkerd2/pull/5496 adds the `viz check` command.  However, more work is still required to add `--proxy` mode and to call the SelfCheck API.

PR #5496, opened 2021-01-08
viz: add check sub-command

Pothulapati:
This adds a new `viz check` cmd performing checks for the resources
in linkerd-viz extension. Checks include resource checks and
the health of resources, certs, etc


viz/cmd/check.go
```diff
@@ -0,0 +1,215 @@
...
+				_, err = kubeAPI.CoreV1().ConfigMaps(vizNamespace).Get(ctx, "collector-config", metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				return nil
+			}))
+
+	checkers = append(checkers,
+		*healthcheck.NewChecker("viz extension pods are running").
+			WithHintAnchor("l5d-viz-collector-running").
```
adleong, line 87: Is this anchor right?

viz/cmd/check.go
```diff
@@ -0,0 +1,215 @@
...
+			Warning().
+			WithRetryDeadline(hc.RetryDeadline).
+			SurfaceErrorOnRetry().
+			WithCheck(func(ctx context.Context) error {
+				pods, err := kubeAPI.GetPodsByNamespace(ctx, vizNamespace)
```
adleong, line 92: Does this just check that the pods which exist in the namespace are running?

Should it also check that the correct pods exist in the first place?  Or perhaps that each of the expected deployments has a non-zero number of ready replicas?

Pothulapati: Checking for the right pods to exist seemed to be the right thing.

 We can also add a another check for replica sets. WDYT?

pkg/healthcheck/healthcheck.go
```diff
@@ -1410,6 +1414,30 @@ func (hc *HealthChecker) checkCertAndAnchorsExpiringSoon(cert *tls.Cred) error {
 	return nil
 }
 
+// CheckAPIService checks the status of the given API Service and returns a error if its not running
```
alpeb, line 1417:
```suggestion
// CheckAPIService checks the status of the given API Service and returns an error if it's not running
```

viz/cmd/check.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, ns := range namespaces.Items {
+		if ns.Labels[k8s.LinkerdExtensionLabel] == name {
+			return &ns, err
+		}
+	}
+	return nil, fmt.Errorf("could not find the linkerd-viz extension. it can be installed by running `linkerd viz install | kubectl apply -f -`")
```
alpeb, line 263:
```suggestion
	return nil, fmt.Errorf("could not find the linkerd-viz extension. It can be installed by running `linkerd viz install | kubectl apply -f -`")
```

viz/cmd/check.go
```diff
@@ -0,0 +1,282 @@
...
+		Example: `  # Check that the viz extension is up and running
+  linkerd viz check`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// Get viz Extension Namespace
+			ns, err := getNamespaceOfExtension(vizExtensionName)
+			if err != nil {
+				fmt.Fprintln(os.Stderr, err.Error())
+				os.Exit(1)
+			}
```
alpeb, line 201: This is redundant with the "linkerd-viz Namespace exists"  check. That check should set a `hc.VizNamespace` as a side-effect for the other checks to use (I know it's a bit ugly to use a side-effect, but let's keep it so till we figure a better general framework).

viz/cmd/check.go
```diff
@@ -0,0 +1,282 @@
...
+					cert, err = hc.FetchCredsFromOldSecret(ctx, vizNamespace, tapOldTLSSecretName)
+				}
+				if err != nil {
+					return err
+				}
+				return hc.CheckCertAndAnchorsExpiringSoon(cert)
+			}))
+
+	checkers = append(checkers,
+		*healthcheck.NewChecker("tap api service is running").
```
alpeb, line 122:
```suggestion
		*healthcheck.NewChecker("tap API service is running").
```

alpeb (commented):
Nice code reuse! :+1:

viz/cmd/check.go
Pothulapati: That side-effect logic makes sense but adding `vizNamespace` under `hc` could be unnecessary. I can make the check update this global `vizNamespace` variable. wdyt?

alpeb: I'm actually using such a `hc.vizNamespace` field to retrieve a viz api client in my public-api split work.

Pothulapati: Hmmm. I'm okay with adding that `vizNamespace in `hc`. I can do that in this PR if there is no cocern. WDYT? @adleong

adleong: I'm not dogmatic but here are my priorities in order:

1. The check command should be performant.  If re-creating a bunch of API clients make the command feel sluggish then we shouldn't do that.  If it has no noticeable performance impact, then let's not bend over backwards to be efficient
2. We should be consistent.  Let's follow the same patterns across the different check commands
3. Side-effects can sometimes be harder to reason about.  Let's avoid them if possible, but not at the expense of the above 2 priorities.

In this case, specifically, I think the redundancy of the namespace check is okay.  Since the first namespace check is fatal, we would expect that if we even get to this check, the namespace lookup should never fail.  (of course, we should still return the error if we somehow encounter it here, just to be defensive).

alpeb (approved):
Thanks @Pothulapati , this looks good to me :+1:

viz/cmd/check.go
Pothulapati: Hmm, The current update i.e using `hc.nam|espace` with only a single namespace check that is fatal does seem to follow the priorities (especially the first two) and would also unblock Alejandros work.

adleong:
There are a few things described in https://github.com/linkerd/linkerd2/issues/5325 which still need to be added here or in a follow-up PR:

* --proxy mode
* calling the SelfCheck API

viz/cmd/check.go
```diff
@@ -0,0 +1,276 @@
...
+			WithRetryDeadline(hc.RetryDeadline).
+			SurfaceErrorOnRetry().
+			WithCheck(func(ctx context.Context) error {
+				pods, err := hc.KubeAPIClient().GetPodsByNamespace(ctx, hc.VizNamespace)
+				if err != nil {
+					return err
+				}
+
+				// Check for relevant pods to be present
+				err = healthcheck.CheckForPods(pods, []string{"linkerd-grafana", "linkerd-prometheus", "linkerd-web", "linkerd-tap"})
```
adleong, line 144: tap-injector will need to be added here once it merges

adleong (approved):

PR #5498, opened 2021-01-08
Use linkerd path in test-cleanup

kleimkuhler:
## What this fixes

When clusters are cleaned up after tests in CI, the `bin/test-cleanup` script is
responsible for clearing the cluster of all testing resources.

Right now this does not work as expected because the script uses the `linkerd`
binary instead of the Linkerd path that is passed in to the `tests` script.

There are cases where different binaries have different uninstall behavior and
the script can complete with an incomplete uninstallation.

## How it fixes

`test-cleanup` now takes a linkerd path argument. This is used to specify the
Linkerd binary that should be used when running in the `uninstall` commands.

This value is passed through from the `tests` invocation which means that in CI,
the same binary is used for running tests as well as cleaning up the cluster.

Additionally, specifying the k8s context has now moved from an argument to the
`--context` flag. This is similar to how `tests` script works because it's not
always required.

## How to use

Shown here:

``` $ bin/test-cleanup -h Cleanup Linkerd integration tests.

Usage:
    test-cleanup [--context k8s_context] /path/to/linkerd

Examples:
    # Cleanup tests in non-default context test-cleanup --context k8s_context
    /path/to/linkerd

Available Commands:
    --context: use a non-default k8s context
```

### DIFF
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -45,15 +45,14 @@ cleanup_usage() {
   echo "Cleanup Linkerd integration tests.
 
 Usage:
-    ${progname} [--context k8s_context] [--linkerd_path /path/to/linkerd]
+    ${progname} [--context k8s_context] /path/to/linkerd
 
 Examples:
-    # Cleanup tests using specified Linkerd binary
-    ${progname} --linkerd_path /path/to/linkerd
+    # Cleanup tests in non-default context
+    ${progname} --context k8s_context /path/to/linkerd
 
 Available Commands:
-    --context: use a non-default k8s context
-    --linkerd_path: use a specific Linkerd binary"
+    --context: use a non-default k8s context"
 }
 
 handle_tests_input() {
@@ -138,28 +137,29 @@ handle_cleanup_input() {
         shift
         shift
         ;;
-      --linkerd_path)
-        linkerd_path=$2
-        if [ -z "$linkerd_path" ]; then
-          echo 'Error: the argument for --linkerd_path was not specified' >&2
-          cleanup_usage "$0" >&2
-          exit 64
-        fi
-        shift
-        shift
-        ;;
       *)
         if echo "$1" | grep -q '^-.*' ; then
           echo "Unexpected flag: $1" >&2
           cleanup_usage "$0" >&2
           exit 64
         fi
+        if [ -n "$linkerd_path" ]; then
+          echo "Multliple linkerd paths specified:" >&2
+          echo "  $linkerd_path" >&2
+          echo "  $1" >&2
+          tests_usage "$0" >&2
+          exit 64
+        fi
+        linkerd_path="$1"
+        shift
         ;;
     esac
   done
 
   if [ -z "$linkerd_path" ]; then
-    linkerd_path=linkerd
+    echo "Error: path to linkerd binary is required" >&2
+    tests_usage "$0" >&2
+    exit 64
   fi
 }
 
@@ -206,7 +206,7 @@ delete_cluster() {
 }
 
 cleanup_cluster() {
-  "$bindir"/test-cleanup --linkerd_path "$linkerd_path" --context "$context" > /dev/null 2>&1
+  "$bindir"/test-cleanup --context "$context" "$linkerd_path" > /dev/null 2>&1
   exit_on_err 'error removing existing Linkerd resources'
 }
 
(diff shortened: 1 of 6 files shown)


bin/test-cleanup
```diff
@@ -2,9 +2,12 @@
 
 set -eu
 
-k8s_context=${1:-""}
 bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 
+# shellcheck source=_test-helpers.sh
+. "$bindir"/_test-helpers.sh
```
alpeb, line 8: Why adding this dependency, and instead just declare `cleanup_usage` and `handle_cleanup_input` here?

bin/_test-helpers.sh
```diff
@@ -40,8 +40,23 @@ Available Commands:
     --images: by default load images into the cluster from the local docker cache (docker), or from tar files located under the 'image-archives' directory (archive), or completely skip image loading (skip)."
 }
 
+cleanup_usage() {
+  progname="${0##*/}"
+  echo "Cleanup Linkerd integration tests.
+
+Usage:
+    ${progname} [--context k8s_context] [--linkerd_path /path/to/linkerd]
```
alpeb, line 48: Nit: If I'm not mistaking we use dashes as word separators in most script parameters, instead of underscores

bin/test-cleanup
kleimkuhler: Pretty much because we do the same thing in `tests`. Just makes those files a little more readable.

bin/_test-helpers.sh
kleimkuhler: Right! I'll change this

kleimkuhler: This is no longer a flag.

bin/test-cleanup
kleimkuhler: @alpeb Another very small reason, but we now use `check_linkerd_binary` in this file as well which is imported from `_test-helpers.sh`.

kleimkuhler:
I removed passing the Linkerd path as a flag and made it a required argument, similar to the `tests` script. I updated the PR description.

bin/_test-helpers.sh
```diff
@@ -85,14 +99,55 @@ handle_input() {
...
+          echo "Unexpected flag: $1" >&2
+          cleanup_usage "$0" >&2
           exit 64
         fi
         if [ -n "$linkerd_path" ]; then
           echo "Multliple linkerd paths specified:" >&2
           echo "  $linkerd_path" >&2
           echo "  $1" >&2
-          usage "$0" >&2
+          tests_usage "$0" >&2
```
alpeb, line 150:
```suggestion
          cleanup_usage "$0" >&2
```

bin/_test-helpers.sh
```diff
@@ -103,7 +158,7 @@ handle_input() {
 
   if [ -z "$linkerd_path" ]; then
     echo "Error: path to linkerd binary is required" >&2
-    usage "$0" >&2
+    tests_usage "$0" >&2
```
alpeb, line 161:
```suggestion
    cleanup_usage "$0" >&2
```

bin/test-cleanup
```diff
@@ -80,13 +85,13 @@ if [[ ${namespaces_controlplane[*]}   || \
 fi
 
 echo "cleaning up viz extension resources, if present [${k8s_context}]"
-linkerd viz uninstall 2> /dev/null | kubectl --context="$k8s_context" delete -f -
+"$linkerd_path" viz uninstall 2> /dev/null | kubectl --context="$k8s_context" delete -f -
 
 echo "cleaning up multicluster resources, if present [${k8s_context}]"
-linkerd mc uninstall 2> /dev/null | kubectl --context="$k8s_context" delete -f -
+"$linkerd_path" mc uninstall 2> /dev/null | kubectl --context="$k8s_context" delete -f -
```
alpeb, line 91: It appears `linkerd mc uninstall` needs to happen before the control plane is deleted, otherwise that returns 
```
$ bin/linkerd mc uninstall
Error: you need Linkerd to be installed in order to install multicluster addons
```

bin/test-cleanup
```diff
@@ -80,13 +85,13 @@ if [[ ${namespaces_controlplane[*]}   || \
...
 
 echo "cleaning up jaeger extension resources, if present [${k8s_context}]"
-linkerd jaeger uninstall 2> /dev/null | kubectl --context="$k8s_context" delete -f -
+"$linkerd_path" jaeger uninstall 2> /dev/null | kubectl --context="$k8s_context" delete -f -
```
alpeb, line 94: `linkerd jaeger uninstall` doesn't seem to have that check, but it'd be better to place this before as well.

alpeb: same for viz?

kleimkuhler: Yea I'll just move all the extension uninstalls above the control plane uninstall!

alpeb (approved):
LGTM :+1:

adleong (approved):

PR #5504, opened 2021-01-10
Report namespace and pod name when port-forward fails

### LINKED ISSUE #5457: linkerd should report the name of the pod in when portforward fails

https://github.com/linkerd/linkerd2/issues/5423#issuecomment-750480989

Currently, when port forwarding fails, the error message reports `[::1]` (localhost ipv6) `:` + (random transient local port) as source and destination:

```
linkerd-data-plane
------------------
√ data plane namespace exists
√ data plane proxies are ready
E1223 16:27:11.233873   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63481: read: connection reset by peer
E1223 16:27:11.317924   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63491: read: connection reset by peer
√ data plane proxy metrics are present in Prometheus
√ data plane is up-to-date
E1223 16:27:11.405402   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63492: read: connection reset by peer
√ data plane and cli versions match
E1223 16:27:11.490493   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63493: read: connection reset by peer
```

https://github.com/linkerd/linkerd2/blob/main/pkg/k8s/portforward.go should report the namespace+pod name

I suppose one could print out each namespace+pod+port before making each connection instead of just at the error state and then force users to collate the port themselves...

output?

```
linkerd-data-plane
------------------
√ data plane namespace exists
√ data plane proxies are ready
E1223 16:27:11.233873   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63481 for somenamespace1/somepodname5: read: connection reset by peer
E1223 16:27:11.317924   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63491 for somenamespace2/somepodname6: read: connection reset by peer
√ data plane proxy metrics are present in Prometheus
√ data plane is up-to-date
E1223 16:27:11.405402   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63492 for somenamespace3/somepodname7: read: connection reset by peer
√ data plane and cli versions match
E1223 16:27:11.490493   16623 portforward.go:385] error copying from local connection to remote stream: read tcp6 [::1]:63479->[::1]:63493 for somenamespace4/somepodname8: read: connection reset by peer
```

piyushsingariya:
Subject
Related to issue #5457 

Problem
Linkerd only reports the local port and the remote port whenever port-forwarding fails.
Linkerd could print out namespace and port if port-forwarding fails instead of just at the error state and then force users to collate the port themselves

Solution
Linkerd needs to print the namespace and the pod name.
- [x] Add two new string variables namespace and podName in `struct PortForward`
- [x] assign the values to the variables when a new Instance is being created in `func NewPortForward()`
run() function propagates the errors that occurred while port-forwarding
- [x] Format the error being returned by `ForwardPorts()` from client-go using `fmt.Errorf()` and add `namespace` and `podName` as suffix and return error

The error is being returned by ForwardPorts() from client-go https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go#L188

Fixes #5457

### DIFF
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -25,6 +25,8 @@ type PortForward struct {
 	method     string
 	url        *url.URL
 	host       string
+	namespace  string
+	podName    string
 	localPort  int
 	remotePort int
 	emitLogs   bool
@@ -140,6 +142,8 @@ func newPortForward(
 		method:     "POST",
 		url:        req.URL(),
 		host:       host,
+		namespace:  namespace,
+		podName:    podName,
 		localPort:  localPort,
 		remotePort: remotePort,
 		emitLogs:   emitLogs,
@@ -172,7 +176,12 @@ func (pf *PortForward) run() error {
 		return err
 	}
 
-	return fw.ForwardPorts()
+	err = fw.ForwardPorts()
+	if err != nil {
+		err = fmt.Errorf("%s for %s/%s", err.Error(), pf.namespace, pf.podName)
+		return err
+	}
+	return nil
 }
 
 // Init creates and runs a port-forward connection.
--- a/pkg/k8s/portforward.go
+++ b/pkg/k8s/portforward.go
@@ -178,7 +178,7 @@ func (pf *PortForward) run() error {
 
 	err = fw.ForwardPorts()
 	if err != nil {
-		err = fmt.Errorf("%s for %s/%s", err.Error(), pf.namespace, pf.podName)
+		err = fmt.Errorf("%s for %s/%s", err, pf.namespace, pf.podName)
 		return err
 	}
 	return nil


pkg/k8s/portforward.go
```diff
@@ -172,7 +176,12 @@ func (pf *PortForward) run() error {
 		return err
 	}
 
-	return fw.ForwardPorts()
+	err = fw.ForwardPorts()
+	if err != nil {
+		err = fmt.Errorf("%s for %s/%s", err.Error(), pf.namespace, pf.podName)
```
alpeb, line 181: You can just have `err` without the `.Error()` part

alpeb (approved):
Looks good to me, just a tiny nit below

pkg/k8s/portforward.go
piyushsingariya: Thanks for pointing this. I improved this :smiley:

adleong (approved):
Very nice!

Issue #5158, opened 2020-10-29
Should config passed through `--config` be authoritative?


Pothulapati:
## Bug Report

Currently, Whatever configurations passed through `--config` with upgrades act as mutations to the current state which means there will be a difference in what `--config` passed during an upgrade and what the current state is (because there could've been previous mutations).

One problem that I came across is when I want to replace all the existing config with a new config passed through `--config`. My initial suggestion was to have a similar flag like `--addon-overwrite` i.e `--overwrite` which skips retrieving existing configuration and replaces with the default + overwrite config passed through `--config`.

But based on my conversations with @olix0r, It would be better to have config passed through `--config` as authoritative and use it without mutating existing config.

### The Helm way

Obviously, It felt better to understand how Helm does this, and maybe follow a simialr methodology here. Helm, when there is no new config passed (through `--set`, `-f`), re-uses the existing config when a upgrade is performed. (This is the expected case).

But when there is extra config specified through `--set`, `--f`, It **skips retrieving the current config and considers whatever passed currently as authoritative unless `--reuse-values` flag is specified**

More on this can be found [here](https://medium.com/@kcatstack/understand-helm-upgrade-flags-reset-values-reuse-values-6e58ac8f127e)

Should we follow a similar approach to have `--config` as authoritative by default and allow mutations only when `--reuse-config` is configured?

@adleong @olix0r

adleong:
Just to be clear, this question is specifically about `linkerd upgrade`, right?  i.e. this is not relevant to `linkerd install` and not relevant to any linkerd extensions (since those do not have upgrade workflows).

It strikes me as very odd that Helm defaults to "reset-values" mode when the `--set` flag is used.  Here is what I would intuitively expect Linkerd to do:

* When `linkerd upgrade` is run without flags, the existing configuration is reused
* When `linkerd upgrade` is run with configuration flags (such as `--disable-heartbeat` for example), those mutations are applied on top of the existing configuration.  This is the "reuse-values" behavior of helm.
* Similarly, when `linkerd upgrade` is used with the `--set` flag, those mutations should be applied on top of the existing configuration.  Essentially, the configuration flags are a shorthand for `--set`.
* When `linkerd upgrade` is run with the `--values` flag (previously named `--config`), the existing configuration is discarded and the given config is applied on top of the chart defaults.  This is the "reset-values" behavior of helm.
[...]

PR #5507, opened 2021-01-11
cli: add helm customization flags to core install

### LINKED ISSUE #5506: cli: Add helm customization flags to linkerd install cmd
## Feature Request

Allow the helm way of customization through `set`, `set-string`, `values`, `set-files` flags for `linkerd install` cmd. this should follow the same extension install model

Pothulapati:
Fixes #5506

This branch adds helm way of customization through
 `set`, `set-string`, `values`, `set-files` flags for
the core control-plane install and upgrade cmds.

For this to work, the helm v3 engine rendering helpers
had to be used instead of our own wrapper type.

## Testing

```bash
tarun in dev in linkerd2 on  tarun/install-helm-flagss [$!?] via 🐹 v1.15.4 took 4s
❯ ./bin/go-run cli install --ignore-cluster --set global.namespace=l6d | grep l6d
WARN[0000] could not set namespace from kubectl context, using 'default' namespace: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
                 ensure the KUBECONFIG path  is valid
  name: l6d
    linkerd.io/control-plane-ns: l6d
  name: linkerd-l6d-identity
    linkerd.io/control-plane-ns: l6d
  name: linkerd-l6d-identity
    linkerd.io/control-plane-ns: l6d
  name: linkerd-l6d-identity
  namespace: l6d
.
.
.
```


cli/cmd/install.go
```diff
@@ -320,20 +332,61 @@ func render(w io.Writer, values *l5dcharts.Values, stage string) error {
...
+		return err
+	}
+
+	// Create values override
+	valuesOverrides, err := options.MergeValues(nil)
+	if err != nil {
+		return err
+	}
+
+	vals, err := chartutil.CoalesceValues(chart, valuesOverrides)
```
adleong, line 370: If I'm reading this correctly, this matches what I describe in https://github.com/linkerd/linkerd2/issues/5158#issuecomment-758306503.  i.e. `MergeValues` will return a set of value overrides generated from the `--set` and `--values` flags and then these overrides will be applied on top of the values loaded from the the Linkerd cluster.

cli/cmd/testdata/install_config.yaml
```diff
@@ -0,0 +1,5 @@
+global:
+  clusterDomain: example.com
+  namespace: l5d
+
+installNamespace: false
```
adleong, line 5: newline please

cli/cmd/install.go
Pothulapati: Yes, That is right as `chart.Values` contains the (default+current cluster config+CLI custom flags) while `vals` is through helm customization flags.

adleong (commented):
I think the `--config` flag, defined in `options.go` should be removed in favor of the `--values` flag.

adleong (approved):

cli/cmd/install.go
```diff
@@ -320,20 +336,61 @@ func render(w io.Writer, values *l5dcharts.Values, stage string) error {
...
-		Files:     files,
-		Fs:        static.Templates,
+	var partialFiles []*loader.BufferedFile
+	for _, template := range charts.L5dPartials {
+		partialFiles = append(partialFiles,
+			&loader.BufferedFile{Name: template},
+		)
+	}
+
+	// Load all Viz chart files into buffer
```
alpeb, line 346:
```suggestion
	// Load all chart files into buffer
```

alpeb (commented):
Ability to bring Helm configurability to the CLI: huge win!! :partying_face:

Pothulapati:
@adleong Should we have these flags for `linkerd install config`? I'm wondering if it could be confusing to users as most values in the readme would not be useful in that case.

adleong:
All stages of `linkerd install` use the same `values.yml`, they just render out different templates.  So yes, I think the flags should be available to `linkerd install`, `linkerd install config`, and `linkerd install control-plane`.

Pothulapati:
Updated!

kleimkuhler (approved):
👍

PR #5510, opened 2021-01-11
Separate observability API

### LINKED ISSUE #5312: Split "controller" into "metrics-api" and "api"
The linkerd-controller component of the control plane serves the Linkerd "public-api" which mostly contains methods for interacting with prometheus but also contains some core methods like for getting the control plane version.  This API should be split into two parts: one for interacting with prometheus and one for core methods.  The linkerd-controller component should be split up as well into two components which each serve one of these APIs.  The metrics-api component will later be moved into a linkerd-viz extension while the core api component will stay in the core control plane.

Specifically the following methods should be moved to the metrics api:

* StatSummary
* Edges
* TopRoutes
* ListPods
  * (The purpose of this API is to health check that metrics from each injected pod have been scraped and exist in prometheus.  We may want to rename and/or simplify this API to focus on this purpose.)
* SelfCheck
  *(A health check that verifies that this can successfully call the k8s API and Prometheus)

All other methods should be left in the core api.

alpeb:
Closes #5312

This is a preliminary step towards moving all the observability API into `/viz`, by first moving its protobuf into `viz/metrics-api`. This should facilitate review as the go files are not moved yet, which will happen in a followup PR. There are no user-facing changes here.

- Moved `proto/common/healthcheck.proto` to `viz/metrics-api/proto/healthcheck.prot`
- Moved the contents of `proto/public.proto` to `viz/metrics-api/proto/viz.proto` except for the `Version` Stuff.
- Merged `proto/controller/tap.proto` into `viz/metrics-api/proto/viz.proto`
- `grpc_server.go` now temporarily exposes `PublicAPIServer` and `VizAPIServer` interfaces to separate both APIs. This will get properly split in a followup.
- The web server provides handlers for both interfaces.
- `cli/cmd/public_api.go` and `pkg/healthcheck/healthcheck.go` temporarily now have methods to access both APIs.
- Most of the CLI commands will use the Viz API, except for `version`.

The other changes in the go files are just changes in the imports to point to the new protobufs.

Other minor changes:
- Removed `git add controller/gen` from `bin/protoc-go.sh`


viz/metrics-api/proto/viz.proto
```diff
@@ -0,0 +1,537 @@
...
+
+  rpc Gateways(GatewaysRequest) returns (GatewaysResponse) {}
+
+  rpc TopRoutes(TopRoutesRequest) returns (TopRoutesResponse) {}
+
+  rpc ListPods(ListPodsRequest) returns (ListPodsResponse) {}
+
+  rpc ListServices(ListServicesRequest) returns (ListServicesResponse) {}
+
+  // Superseded by `TapByResource`.
```
adleong, line 524: Not for this PR but before 2.10 we should investigate if these can be safely removed.

adleong (approved):
💯 🔥

Pothulapati (approved):
LGTM! Awesome that it would make the actual separation very simple.

One small nit would be to strip out the public naming part in Viz and just calling it `vizPb` and `vizClient` everywhere. The public part makes it a bit confusing. Does it signify anything? This could be part of another PR for sure

alpeb:
> One small nit would be to strip out the public naming part in Viz and just calling it `vizPb` and `vizClient` everywhere. The public part makes it a bit confusing. Does it signify anything? This could be part of another PR for sure

I'll keep that in mind. Naming in this PR was a bit verbose to be able to tell apart both APIs, but I'll address that in the followup PR :+1:

PR #5513, opened 2021-01-11
Enable upgrade integration tests for ARM

alpeb:
This enables the `helm-upgrade` and `upgrade-stable` integration tests,
that were disabled because the previous versions didn't have ARM
support, but now 2.9 does.


kleimkuhler (approved):

cpretzer (approved):
:+1:

PR #5517, opened 2021-01-11
Add back support for proxy resource settings

alpeb:
The last viz refactoring removed support for modifying the k8s resources
used by the proxies injected into the control plane components (values
like `tapProxyResources`, `prometheus.proxy.resources`, etc).

This adds them back using annotations, with `values.yaml` configs using a consistent naming: `tap.proxy.resources`,
`dashboard.proxy.resources`, etc.

Also fixes the tap helm template that was making reference to
`.Values.tapResources` instead of `.Values.tap.resources`.


viz/charts/linkerd-viz/values.yaml
```diff
@@ -137,6 +142,11 @@ dashboard:
       # -- Amount of memory that the web container requests
       request:
 
+  proxy:
+    # -- If set, overrides default proxy resources for the proxy injected
+    # into the tap component
```
kleimkuhler, line 147:
```suggestion
    # into the dashboard component
```

kleimkuhler (approved):
Nice good catch!

adleong (approved):

Issue #5524, opened 2021-01-12
The default namespace isn't detected by CLI commands


alpeb:
### What is the issue?

The default namespace for the current context isn't being properly detected.

### How can it be reproduced?

```bash
$ k3d cluster create cluster1
$ kubectl create ns somenamespace

# set the default namespace to somenamespace
$ kubectl config set-context k3d-cluster1 --namespace=somenamespace

# install linkerd core and viz
$ linkerd install | kubectl apply -f -
$ linkerd viz install | kubectl apply -f -

# inject and install booksapp in somenamespace
$ linkerd inject https://run.linkerd.io/booksapp.yml | kubectl apply -f -

# stats are showing fine
$ linkerd stat deploy
NAME      MESHED   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TCP_CONN
authors      1/1   100.00%   0.0rps           3ms           3ms           3ms          1
books        1/1         -        -             -             -             -          -
traffic      1/1         -        -             -             -             -          -
webapp       3/3   100.00%   0.0rps           3ms           3ms           3ms          3

# create a second cluster
$ k3d cluster create cluster2

# when pointing to the first cluster, the default namespace should be retrieved but it isn't
$ linkerd --context k3d-cluster1 stat deploy
No traffic found.
```
### Environment

- Kubernetes Version: 1.19
- Cluster Environment: k3d
- Host OS: Linux
- Linkerd version: edge-21.1.1

### Possible solution

I think the problem lies here:
https://github.com/linkerd/linkerd2/blob/40192e258ab1765944529d80a1ca2d1a7d614c74/cli/cmd/root.go#L101-L106

`getDefaultNamespace()` makes use of the global vars `kubeconfigPath` and `kubeContext` that start with an empty string as a default value, and there's no chance for Cobra to update them before `defaultNamespace` is set.
One possible solution that comes to mind would be to lazily call `getDefaultNamespace()` there where `defaultNamespace` is currently used.

alpeb:
Note this is currently being addressed in #5485

alpeb:
Fixed by #5485

PR #5530, opened 2021-01-13
Detect default ns for metrics and profile subcommands

alpeb:
Followup to #5485, fixes remaining cases for #5524

Properly detect the default namespace given `kubeConfigPath` and
`kubeContext` for the `metrics` and `profile` subcommands.

Also gets rid once and for all of the `defaultNamespace` global var.


adleong (approved):

Pothulapati:
@alpeb This logic should also be part of `linkerd identity` and `linkerd viz routes` right? It seems to be missing there.

alpeb:
> @alpeb This logic should also be part of `linkerd identity` and `linkerd viz routes` right? It seems to be missing there.

Good catch :+1:  I'll add that to `linkerd viz routes` in a moment. Although while checking `linkerd identity` I found some other inconsistencies around that, so I'll fix that in a followup PR.

alpeb:
Ok I ended up pushing to this same branch the default namespace mechanism for `linkerd identity`; the inconsistency was just a bad description in the `--namespace` flag.

Pothulapati (approved):
LGTM

PR #5532, opened 2021-01-13
Cherry-picks and release notes for stable-2.9.2

kleimkuhler:
## stable-2.9.2

This stable release fixes an issue that stops traffic to a pod when there is an
IP address conflict with another pod that is not in a running state.

It also fixes an upgrade issue when using HA that would lead to values being
overridden.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Changes
 
+## stable-2.9.2
+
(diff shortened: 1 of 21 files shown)


kleimkuhler:
I needed to force push to fix one of the cherry picked `Signed off by` trailers.

adleong (approved):

olix0r (approved):

PR #5540, opened 2021-01-14
viz: add tap-injector

### LINKED ISSUE #5326: create tap-injector webhook
Create a muting admission controller webhook called tap-injector which adds the appropriate tap environment variables to the Linkerd proxy container.

This webhook adds the LINKERD2_PROXY_TAP_SVC_NAME (and LINKERD2_PROXY_TAP_DISABLED if necessary) environment variables to the Linkerd proxy container if it exists. If a pod does not have a Linkerd proxy sidecar, no action should be taken. However, this webhook should have a IfNeeded reinvokation policy so that it will be reinvoked if the Linkerd proxy is added.

By default, this injector should be invoked on all newly created pods and act on them if they have a Linkerd proxy sidecar. However, it should also be possible to configure the injector with a label selector to selectively opt in which pods will be affected.

kleimkuhler:
## What this changes

This adds a tap-injector component to the `linkerd-viz` extension which is
responsible for adding the tap service name environment variable to the Linkerd
proxy container.

If a pod does not have a Linkerd proxy, no action is taken. If tap is disabled
via annotation on the pod or the namespace, no action is taken.

This also removes the environment variable for explicitly disabling tap through
an environment variable. Tap status for a proxy is now determined only be the
presence or absence of the tap service name environment variable.

Closes #5326

### tap-injector

The tap-injector component determines if `LINKERD2_PROXY_TAP_SVC_NAME` should be
added to a pod's Linkerd proxy container environment. If the pod satisfies the
following, it is added:

- The pod has a Linkerd proxy container
- The pod has not already been mutated
- Tap is not disabled via annotation on the pod or the pod's namespace

### LINKERD2_PROXY_TAP_DISABLED

Now that tap is an extension of Linkerd and not a core component, it no longer
made sense to explicitly enable or disable tap through this Linkerd proxy
environment variable. The status of tap is now determined only be if the
tap-injector adds or does not add the `LINKERD2_PROXY_TAP_SVC_NAME` environment
variable.

### controller image

The tap-injector has been added to the controller image's several startup
commands which determines what it will do in the cluster.

As a follow-up, I think splitting out the `tap` and `tap-injector` commands from
the controller image into a linkerd-viz image (or something like that) makes
sense.


adleong (approved):
I've tested and this works as expected.  I ran into a few related issues which I have filed follow-up issues for:

https://github.com/linkerd/linkerd2/issues/5545
https://github.com/linkerd/linkerd2/issues/5544

controller/tap-injector/webhook.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func getProxyContainerIndex(containers []corev1.Container) int {
+	for i, c := range containers {
+		if c.Name == labels.ProxyContainerName {
+			return i
+		}
+	}
+	return -1
+}
```
alpeb, line 83: Given it's also used in the jaeger injector, this is something we can move to some common place like `/controller/webhook` (which itself should be later on moved to `pkg/` as all that is used by extensions as well) (not a blocker).

viz/charts/linkerd-viz/templates/tap-injector-rbac.yaml
```diff
@@ -0,0 +1,95 @@
+###
+### Tap Injector RBAC
+###
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{.Values.namespace}}-tap-injector
```
alpeb, line 8: Given this is a cluster-level resource, and to follow the same scheme as the Jaeger injector, I don't think it's necessary to parametrize by the namespace, and instead hard-code as `linkerd-tap-injector` so it's easily identified (same comment for the resources below).

viz/charts/linkerd-viz/templates/tap-injector-rbac.yaml
```diff
@@ -0,0 +1,95 @@
...
+  labels:
+    {{.Values.extensionAnnotation}}: {{.Values.namespace}}
```
alpeb, line 10: The value should be hard-coded to `linkerd-viz` as we might be selecting elsewhere using that value regardless of the namespace used for linkerd-viz. (same comment for the resources below).

viz/charts/linkerd-viz/templates/tap-injector-rbac.yaml
```diff
@@ -0,0 +1,95 @@
...
+apiVersion: v1
+metadata:
+  name: tap-injector-k8s-tls
+  namespace: {{ .Values.namespace }}
+  labels:
+    {{.Values.extensionAnnotation}}: {{.Values.namespace}}
+    component: tap-injector
+    namespace: {{.Values.namespace}}
+  annotations:
+    {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
```
alpeb, line 56: I think this needs to be added to all the other resources in this file.

viz/charts/linkerd-viz/templates/tap-injector.yaml
```diff
@@ -0,0 +1,110 @@
...
+    app.kubernetes.io/version: {{default .Values.linkerdVersion .Values.controllerImageVersion}}
+    component: tap-injector
+    namespace: {{.Values.namespace}}
+  name: tap-injector
+  namespace: {{.Values.namespace}}
+spec:
+  replicas: {{.Values.tapInjector.replicas}}
+  selector:
+    matchLabels:
+      {{.Values.extensionAnnotation}}: linkerd-viz
```
alpeb, line 45: These are labels selectors, so the annotation doesn't belong here. Same for the template below.

viz/charts/linkerd-viz/templates/tap-injector.yaml
```diff
@@ -0,0 +1,110 @@
...
+        component: tap-injector
+        namespace: {{.Values.namespace}}
+        {{- with .Values.podLabels }}{{ toYaml . | trim | nindent 8 }}{{- end }}
+    spec:
+      {{- if .Values.tolerations -}}
+      {{- include "linkerd.tolerations" . | nindent 6 }}
+      {{- end -}}
+      {{- include "linkerd.node-selector" . | nindent 6 }}
+      {{- if .Values.enablePodAntiAffinity -}}
+      {{- $local := dict "component" "tap-injector" "label" .Values.controllerComponentLabel -}}
```
alpeb, line 72:
```suggestion
      {{- $local := dict "component" "tap-injector" "label" "component" -}}
```

viz/charts/linkerd-viz/templates/tap-injector.yaml
```diff
@@ -0,0 +1,110 @@
...
+      {{- include "linkerd.pod-affinity" $local | nindent 6 -}}
+      {{- end }}
+      containers:
+      - args:
+        - tap-injector
+        - -tap-service-name=linkerd-tap.{{.Values.namespace}}.serviceaccount.identity.$(_l5d_ns).$(_l5d_trustdomain)
+        image: {{.Values.tapInjector.image.registry}}/{{.Values.tapInjector.image.name}}:{{ default (default .Values.linkerdVersion .Values.controllerImageVersion) .Values.tapInjector.image.tag}}
```
alpeb, line 79: There's no `.Values.controllerImageVersion` in viz' `values.yaml`. I think the other resources have this bug as well.

viz/charts/linkerd-viz/templates/tap-injector.yaml
```diff
@@ -0,0 +1,110 @@
...
+        - containerPort: 9995
+          name: admin-http
+        readinessProbe:
+          failureThreshold: 7
+          httpGet:
+            path: /ready
+            port: 9995
+        {{- if .Values.tapInjector.resources -}}
+        {{- include "partials.resources" .tapInjector.resources | nindent 8 }}
+        {{- end }}
```
alpeb, line 99: If you want to use `.tapInjector.resources` then you need to use `with` instead of `if`.

viz/charts/linkerd-viz/values.yaml
```diff
@@ -106,6 +106,45 @@ tap:
...
+    # - key: runlevel
+    #   operator: NotIn
+    #   values: ["0","1"]
+  objectSelector:
+    # matchLabels:
+    #   foo: bar
+  UID: *uid
+  failurePolicy: Ignore
+  # resources:
+  # -- Do not create a secret resource for the tapInjector webhook. If this is
```
alpeb, line 131: A newline is required here for the comment to be consumed by helm-docs

alpeb (commented):
Looking good :+1: 
There still are occurrences of `disableTap` in some golden files, `_validate.tpl` and `values.go`. 
Below some remarks from a first pass.

alpeb:
> @alpeb These `disableTap`s are referencing the annotation `config.linkerd.io/disable-tap`, not the proxy container env variable `LINKERD2_PROXY_TAP_DISABLED`.

Gotcha. It's another one of those dynamic settings that reuse the `values.yaml` underlying go struct but are not supposed to be configured in `values.yaml`...

viz/charts/linkerd-viz/templates/tap-injector-rbac.yaml
kleimkuhler: I think I may just keep it on this secret. Pretty much all the RBAC templates in the Viz extension and control plane have this annotation for their secrets, but not other resources like cluster role and bindings.

If we want this annotation everywhere I'd rather make a change where we add it everywhere, instead of semi-randomly doing it.

olix0r:
Per https://github.com/linkerd/linkerd2/issues/5544#issuecomment-761104378

Does/should the tap injector set the disabled annotation on pods as they are created. I think this is mostly relevant when tap is disabled at the namespace level.

viz/charts/linkerd-viz/templates/tap-injector.yaml
kleimkuhler: Ah good catch. I'll fix this for the other components in Viz extension as part of this.

olix0r:
To summarize a conversation we had outside of GitHub: we think it's appropriate for the injector to set a pod annotation like `viz.linkerd.io/tap-enabled: true` when tap is enabled by a workload. This will allow the tap controller, etc, to avoid connecting to endpoints that do not have this setting.

pkg/k8s/labels.go
```diff
@@ -478,10 +478,17 @@ func IsMeshed(pod *corev1.Pod, controllerNS string) bool {
...
-// disabling tap
-func IsTapDisabled(pod *corev1.Pod) bool {
-	if valStr := pod.Annotations[ProxyDisableTapAnnotation]; valStr != "" {
+// IsTapDisabled returns true if a namespace or pod has an annotation for
+// explicitly disabling tap
+func IsTapDisabled(obj interface{}) bool {
+	var valStr string
+	switch resource := obj.(type) {
+	case *corev1.Pod:
+		valStr = resource.GetAnnotations()[ProxyDisableTapAnnotation]
```
Pothulapati, line 487: Hmmm, The outcome of both of these cases are same?

testutil/inject_validator.go
```diff
@@ -129,7 +138,7 @@ func (iv *InjectValidator) validateProxyContainer(pod *v1.PodSpec) error {
 	}
 
 	if iv.DisableTap {
-		if err := iv.validateEnvVar(proxyContainer, "LINKERD2_PROXY_TAP_DISABLED", enabled); err != nil {
+		if err := iv.validateNoEnvVar(proxyContainer, "LINKERD2_PROXY_TAP_SVC"); err != nil {
```
Pothulapati, line 141: 👍

viz/charts/linkerd-viz/templates/tap-injector.yaml
```diff
@@ -0,0 +1,103 @@
...
+      maxUnavailable: 1
+  {{- end }}
+  template:
+    metadata:
+      annotations:
+        {{- if empty .Values.cliVersion }}
+        checksum/config: {{ include (print $.Template.BasePath "/tap-injector-rbac.yaml") . | sha256sum }}
+        {{- end }}
+        {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
+        {{- with .Values.podAnnotations }}{{ toYaml . | trim | nindent 8 }}{{- end }}
```
Pothulapati, line 54: #5517 added support to override proxy resources for deployments. We should have that here too right? along with `tapInjector.proxy.resources` in values.yaml

Pothulapati (commented):
Changes look good, I had some nits here and there

Will be doing more testing now :}

pkg/k8s/labels.go
kleimkuhler: Yep but in both cases `resource` is a now a type—namespace or pod—that has the `GetAnnotations` method.

Without this, `obj` is just some interface.

viz/charts/linkerd-viz/templates/tap-injector.yaml
kleimkuhler: Yep good catch!

pkg/k8s/labels.go
Pothulapati: Makes sense. Can we do [a case list](https://www.dotnetperls.com/switch-go) in that case so that we don't have to repeat that solution 🤔

Pothulapati (approved):
LGTM and works as expected. Also, Tried out rolling out components in Viz to see if tap injection works and it does 🥳 

Also, I have not done any scale testing, which may not be part of this.

pkg/k8s/labels.go
kleimkuhler: Ah right good call!

kleimkuhler: This actually doesn't work. Compiler still gets an error `resource.GetAnnotations undefined (type interface{} has no field or method GetAnnotations)`

kleimkuhler:
Scale testing was an issue but was addressed by adding `watch` to the cluster role [here](https://github.com/linkerd/linkerd2/pull/5540/files#diff-189ce324e442653d34e962c3b0aaf483bfbf5e5dba99842ab6b56247acfdc42eR857)

It was also explained a little more thoroughly in #5548!

PR #5553, opened 2021-01-18
public-api/viz split 1/8: a couple more protobuf changes

alpeb:
- Moved `healthcheck.proto` back from viz to `proto/common` as it remains being used by the main `healthcheck.go` library (it was moved to viz by #5510).
- Extracted from `viz.proto` the IP-related types and put them in `/controller/gen/common/net` to be used by both the public and the viz APIs.


Pothulapati (approved):
LGTM, 👍  on moving the IP types to a common pkg

adleong:
As far as I can tell, the only place that `healthcheck.proto` is used in `healthcheck.go` is for the `checkRPC` machinery which only exists to call the `SelfCheck` API (which itself is moving to the viz API.  I would expect that `healthcheck.proto` would move to the viz extension and all of the `checkRPC` stuff would be removed from `healthcheck.go`.

I think `checkRPC` was probably overly complicated to begin with and I think only existed so that a single call to `SelfCheck` could return multiple check results.  Once this moves to the viz extension, I am perfectly okay simplifying this to be a regular check which returns a single result, just like everything else.

alpeb:
@adleong that's a great call. In #5557 I have moved the `SelfCheck` API. I'll see how easy it would be to make that change in #5557 and its downstream PRs, or as a separate effort altogether.

alpeb:
I figured the easiest would be to tackle the SelfCheck API refactoring separately, so I raised #5575 detailing the changes.

kleimkuhler (commented):
`SelfCheck` refactoring aside, these changes look good.

@alpeb Just to make sure I'm following, you won't be making the `SelfCheck` refactoring changes as part of #5557 and will instead do this separately once this whole 1-8 merges?

alpeb:
> @alpeb Just to make sure I'm following, you won't be making the `SelfCheck` refactoring changes as part of #5557 and will instead do this separately once this whole 1-8 merges?

Correct

kleimkuhler (approved):
👍

alpeb:

PR #5554, opened 2021-01-18
public-api/viz split 2/8: chart templates for new viz linkerd-metrics-api pod

alpeb:
This is based on the branch `alpeb/api-move` from #5553

The underlying go code supporting the service will be pushed in a followup PR, so tests are not expected to pass on this one.

### DIFF
--- a/viz/charts/linkerd-viz/templates/metrics-api.yaml
+++ b/viz/charts/linkerd-viz/templates/metrics-api.yaml
@@ -34,7 +34,6 @@ metadata:
     app.kubernetes.io/part-of: Linkerd
     app.kubernetes.io/version: {{default .Values.linkerdVersion .Values.controllerImageVersion}}
     component: metrics-api
-    namespace: {{.Values.namespace}}
   name: linkerd-metrics-api
   namespace: {{.Values.namespace}}
 spec:
@@ -43,7 +42,6 @@ spec:
     matchLabels:
       {{.Values.extensionAnnotation}}: linkerd-viz
       component: metrics-api
-      namespace: {{.Values.namespace}}
   template:
     metadata:
       annotations:
@@ -58,7 +56,6 @@ spec:
       labels:
         {{.Values.extensionAnnotation}}: linkerd-viz
         component: metrics-api
-        namespace: {{.Values.namespace}}
         {{- with .Values.podLabels }}{{ toYaml . | trim | nindent 8 }}{{- end }}
     spec:
       {{- if .Values.tolerations -}}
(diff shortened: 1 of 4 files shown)


adleong (commented):
This looks good to me.  I assume the reason this can't pass CI on its own is because the metrics-API docker image doesn't exist yet?

alpeb:
> This looks good to me. I assume the reason this can't pass CI on its own is because the metrics-API docker image doesn't exist yet?

Yeah, lots of missing pieces at this point. The only CI runs that make sense are the ones from the last item #5560

viz/charts/linkerd-viz/templates/metrics-api.yaml
```diff
@@ -0,0 +1,105 @@
...
+metadata:
+  annotations:
+    {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
+  labels:
+    {{.Values.extensionAnnotation}}: linkerd-viz
+    app.kubernetes.io/name: metrics-api
+    app.kubernetes.io/part-of: Linkerd
+    app.kubernetes.io/version: {{default .Values.linkerdVersion .Values.controllerImageVersion}}
+    component: metrics-api
+    namespace: {{.Values.namespace}}
```
kleimkuhler, line 37: Is the `namespace` label used anywhere? I know a few of the Viz components have this on resources, but I'm not totally sure on the purpose it serves.

This is related to the general cleanup of labels we should do across extensions anyways so not a blocker, but just curious seeing it being added here.

viz/charts/linkerd-viz/templates/metrics-api.yaml
```diff
@@ -0,0 +1,105 @@
...
+  name: linkerd-metrics-api
+  namespace: {{.Values.namespace}}
+spec:
+  replicas: {{.Values.metricsAPI.replicas}}
+  selector:
+    matchLabels:
+      {{.Values.extensionAnnotation}}: linkerd-viz
+      component: metrics-api
+      namespace: {{.Values.namespace}}
```
kleimkuhler, line 46: And if `namespace` is removed I think selecting by extension label and component name should be enough.

viz/charts/linkerd-viz/templates/metrics-api.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - args:
+        - -controller-namespace={{.Values.linkerdNamespace}}
+        - -log-level={{.Values.metricsAPI.logLevel}}
+        - -cluster-domain={{.Values.clusterDomain}}
+        {{- if .Values.prometheusUrl }}
+        - -prometheus-url={{.Values.prometheusUrl}}
+        {{- else }}
+        - -prometheus-url=http://linkerd-prometheus.linkerd-viz.svc.{{.Values.clusterDomain}}:9090
+        {{- end }}
+        image: {{.Values.metricsAPI.image.registry}}/{{.Values.metricsAPI.image.name}}:{{ default (default .Values.linkerdVersion .Values.controllerImageVersion) .Values.metricsAPI.image.tag}}
```
kleimkuhler, line 82:
```suggestion
        image: {{.Values.metricsAPI.image.registry}}/{{.Values.metricsAPI.image.name}}:{{.Values.metricsAPI.image.tag}}
```

viz/charts/linkerd-viz/templates/metrics-api.yaml
alpeb: Yeah I don't see a purpose for this either, should be removed.

viz/charts/linkerd-viz/templates/metrics-api.yaml
alpeb: Agreed.

kleimkuhler (approved):
👍

alpeb:

PR #5555, opened 2021-01-18
public-api/viz split 3/8: spin-off viz healthcheck

alpeb:
This is based on the branch `alpeb/linkerd-metrics-api` from #5554

- Created `viz/pkg/healthcheck/healthcheck.go` that wraps the original `pkg/healthcheck/healthcheck.go` while adding the `vizNamespace` and `vizAPIClient` fields which were removed from the core `healthcheck`. That way the core healthcheck doesn't have any dependencies on viz, and viz' healthcheck can now be used to retrieve viz api clients.
- The core and viz healthcheck libs are now abstracted out via the new `healthcheck.Runner` interface.
- ~~Moved to viz the `linkerd-data-plane` checks because they rely on Prometheus, but they're currently commented-out pending the wiring up of `linkerd viz check --proxy`.~~ Refactored the data plane checks so they don't rely on calling `ListPods`
- ~~The core healthcheck unit tests dealing with the viz api have been removed and will be refactored later into a new set of viz healthcheck unit tests.~~
- The checks in `viz/cmd/check.go` have been moved to `viz/pkg/healthcheck/healthcheck.go` as well, so `check.go`'s sole responsibility is dealing with command business. This command also now retrieves its viz api client through viz' healthcheck.
- ~~The `getNamespaceOfExtensions()` function has been refactored into `viz.pkg.GetVizNamespace()` as is consumed from a few other places as well.~~
- `multicluster/cmd/check.go` also now relies on viz' api because it hits the `Gateways()` endpoint that depends on Prometheus. **In a followup, we should move that api into some MC component (the gateway itself?) so multicluster can be installed without viz**
- Ditto for `multicluster/cmd/gateways.go`

Tests are not supposed to pass in this branch. As a followup I'll be pushing in a separate PR the changes to viz commands building theirs clients against the new `linkerd-metrics-api` component via the new viz' healthcheck lib.


adleong:
I think it makes sense to keep the data plane checks in the core check.  As far as I can tell, there's no good reason why these checks use the ListPods API instead of just using the kubernetes API directly.  I think if we switched to getting pods directly from k8s instead of using ListPods, we could keep the data plane checks in the core.

adleong:
Instead of moving `getNamespaceOfExtensions`, can we just replace it with `KubernetesAPI.GetNamespaceWithExtensionLabel`?  It seems like we have too many different ways of doing the same thing so we should try to standardize.

alpeb:
> I think it makes sense to keep the data plane checks in the core check. As far as I can tell, there's no good reason why these checks use the ListPods API instead of just using the kubernetes API directly. I think if we switched to getting pods directly from k8s instead of using ListPods, we could keep the data plane checks in the core.

True. The ListPods API hits Prometheus to populate the `SinceLastReport` and `Uptime` fields, but that's not used here. I'll move those checks back.

> Instead of moving getNamespaceOfExtensions, can we just replace it with `KubernetesAPI.GetNamespaceWithExtensionLabel`? 

Makes sense :+1:

pkg/healthcheck/healthcheck.go
```diff
@@ -1629,15 +1622,13 @@ func (hc *HealthChecker) KubeAPIClient() *k8s.KubernetesAPI {
...
+func (hc *HealthChecker) PublicAPIClient() public.Client {
+	return hc.apiClient
 }
 
-// VizAPIClient returns a fully configured Viz API client. This client is
-// only configured if the KubernetesAPIChecks and LinkerdAPIChecks are
-// configured and run first.
-func (hc *HealthChecker) VizAPIClient() public.VizAPIClient {
-	return hc.apiClient
+// LatestVersions returns the latest from Linkerd release channels
```
kleimkuhler, line 1629:
```suggested
// LatestVersions returns the latest versions from Linkerd release channels
```

multicluster/cmd/gateways.go
```diff
@@ -41,7 +41,14 @@ func newGatewaysCommand() *cobra.Command {
 				return err
 			}
 
-			client, err := public.NewExternalClient(cmd.Context(), controlPlaneNamespace, k8sAPI)
+			ctx := cmd.Context()
+
+			vizNs, err := k8sAPI.GetNamespaceWithExtensionLabel(ctx, "linkerd-viz")
+			if err != nil {
+				return err
```
kleimkuhler, line 48: For now should this produce a more helpful error explaining this relies on the Viz extension being installed?

kleimkuhler (commented):
This looks good! I like that the Viz healtcheck is wrapping the `pkg/` healthcheck.

alpeb:

PR #5556, opened 2021-01-18
public-api/viz split 4/8: remove linkerd-controller dependency on Prometheus

alpeb:
This is based on the branch `alpeb/viz-healthcheck` from #5555

- Removed the `global.prometheusUrl` config in the core `values.yml`.
- Removed `linkerd-controller` dependencies on Prometheus.
- Leave the Heartbeat's `-prometheus` flag hard-coded temporarily. TO-DO: have it automatically discover viz and pull Prometheus' endpoint (#5352).

Tests are not supposed to pass in this branch. More wiring will come up in the followup PRs.

PR #5557, opened 2021-01-18
public-api/viz split 5/8: move observability gRPC from linkerd-controller to viz

alpeb:
This is based on the branch `alpeb/no-more-controller-prom` from #5556

We continue removing the linkerd-controller dependencies on viz stuff, except for things required for the tap server, which can be tackled separately afterwards.

- Created a new gRPC server under `viz/metrics-api` moving prometheus-dependent functions out of the core gRPC server and into it (same thing for the accompaigning http server).
- Did the same for the `PublicAPIClient` (now called just `Client`) interface. The `VizAPIClient` interface disappears as it's enough to just rely on the viz `ApiClient` protobuf type.
- Moved the other files implementing the rest of the gRPC functions from `controller/api/public` to `viz/metrics-api` (`edge.go`, `stat_summary.go`, etc.).
- Also simplified some type names to avoid stuttering.

Tests are not supposed to pass in this branch.


alpeb:

PR #5558, opened 2021-01-18
public-api/viz split 6/8: linkerd-metrics-api bootstrap files

alpeb:
This is based on the branch `alpeb/grpc-viz` from #5557.

Here we complete the work started in #5554 ("Chart templates for new viz linkerd-metrics-api pod") by providing the Dockerfile, `main.go` command and the `bin/` scripts for building the new `linkerd-metrics-api` pod under the `linkerd-viz` namespace.

At the same time, we strip out of the public-api's `main.go` file the prometheus parameters and other no longer relevant bits.

Finally, updated tests related to linkerd-metrics-api.

Tests are not expected to pass on this one either.

PR #5559, opened 2021-01-18
public-api/viz split 7/8: linkerd-web updates

alpeb:
This is based on the branch `alpeb/linkerd-metrics-api-bootstrap` from #5558.

`linkerd-web` requires connecting with both the public-api and the viz api, so both addresses (and the viz namespace) are now provided as parameters to the container.

We're also updating some dependencies that still were pointing to the public-api and now should point to the viz api.

Tests are not expected to pass on this one either.


alpeb:

PR #5564, opened 2021-01-19
jaeger: cli and check logic cleanup

Pothulapati:
This branch cleans up some of the unecessary logic that is not
needed and thus making the check logic similar to that of other
extensions, namely viz.

Includes the following cleanups:

- Remove `namespace` flag in jaeger CLI and make the fetching logic
dynamic and use it in check and dashboard.
- Use `hc.KubeAPIClient` instead of creating our own in jaeger check.
- Move injection checks up before we run the readiness checks

This change adds a new extension namespace exist check for
jaeger.

Also, Updates integration tests to run the check commands.


test/integration/install_test.go
```diff
@@ -404,6 +404,15 @@ func TestInstallOrUpgradeCli(t *testing.T) {
...
+	// Wait for the proxy injector to be up
+	name := "linkerd-proxy-injector"
+	ns := "linkerd"
+	o, err := TestHelper.Kubectl("", "--namespace="+ns, "rollout", "status", "--timeout=120s", "deploy/"+name)
+	if err != nil {
+		testutil.AnnotatedFatalf(t, fmt.Sprintf("failed to wait for condition=available for deploy/%s in namespace %s", name, ns),
+			"failed to wait for condition=available for deploy/%s in namespace %s: %s: %s", name, ns, err, o)
+	}
+
 	// Install Linkerd Viz Extension
```
Pothulapati, line 416: Unrelated Issue: We currently don;t run the upgrade tests for viz correctly i.e staring with the stable/edge install and then installing the source. We only do source install currently

I will raise an another PR to do that

alpeb (commented):
Given that now we check first `healthcheck.KubernetesAPIChecks`, we can get rid of the `kubeAPI` creation and the namespace check inside of `RunE` that will be performed anyways as a regular check, just like we did with viz check.

Pothulapati:
@alpeb That makes sense. Updated

kleimkuhler (approved):
Looks good 👍

adleong (approved):

Pothulapati:
The explanation here is that, This PR updated the integration tests to use `viz wait` and `jaeger wait` instead off waiting on those components directly. The wait logic wasn't working correctly because the injector was taking a bit more time before its ready in `2/2` but the current check was only checking the Pod status which meant the emojivoto was being created even before `jaeger-injector` was up making the tracing tests fail.

This is fixed by updating the `CheckPodsRunnning` to use `k8s.GetPodStatus` which also had to be updated to reflect the current logic from k8s checks.

pkg/k8s/api_test.go
```diff
@@ -132,7 +132,7 @@ status:
 		},
 		{
 			desc:     "Pod has a running container",
-			expected: "Running",
+			expected: "NotReady",
```
alpeb, line 135: Nice, it seems this is more coherent given the the phase is still "Pending"

alpeb: Actually no, I was looking at the pod spec above this... is this pod's status missing something that makes it NotReady?

pkg/k8s/api_test.go
```diff
@@ -203,7 +203,7 @@ status:
 		},
 		{
 			desc:     "Pod init container terminated with signal",
-			expected: "Init:Signal:9",
+			expected: "Running",
```
alpeb, line 206: This on the other hand seems congruent with the fact that the "signal" field is no longer taken into account :+1:

pkg/k8s/api.go
```diff
@@ -192,7 +192,7 @@ func GetPodStatus(pod corev1.Pod) string {
 	for i := range pod.Status.InitContainerStatuses {
 		container := pod.Status.InitContainerStatuses[i]
 		switch {
-		case container.State.Terminated != nil && container.State.Terminated.ExitCode == 0 && container.State.Terminated.Signal == 0:
+		case container.State.Terminated != nil && container.State.Terminated.ExitCode == 0:
```
kleimkuhler, line 195: @alpeb Already commented on this above but do we want to ignore all signals now? I get ignoring `9`, but maybe others may be more significant? Not a blocker, but just curious if this is definitely what we want now.

Pothulapati:
@alpeb @kleimkuhler Instead of using `GetPodStatus` which seems to be used with statSummary, etc to give status similar as that of kubectl, I updated my `CheckPodsRunning` to check for container statuses to as that's what the use-case is here making this much simpler.

Updating `GetPodStatus` to the latest code from k8s also wasn't giving the answer that this usecase needs i.e Status as not `running` when atleast one container is not ready. Instead, just like `kubectl` it was giving `running`.

alpeb (approved):
Works for me :+1: 
Check blocks as expected when I artificially delay the readiness of the injector container using a large `initialDelaySeconds` :+1:

kleimkuhler (approved):
Looks good

PR #5567, opened 2021-01-19
Fix `linkerd metrics` resource selector

siggy:
The `linkerd metrics` command was selecting pods based on owner resource
names. If multiple owners existed with the same name (for example
`sts/web`, `deploy/web`), additional pods would be incorrectly included
in the output.

Fix the pod selector code to validate pods have owner references to the
given workload/owner.

Before:
```
$ linkerd metrics -n emojivoto deploy/web|grep POD
  # POD web-0 (1 of 3)
  # POD web-d9ffd684f-gnbcx (2 of 3)
  # POD web-fs6l7 (3 of 3)
```

After:
```
$ bin/go-run cli metrics -n emojivoto deploy/web|grep POD
  # POD web-d9ffd684f-gnbcx (1 of 1)
```

### DIFF
--- a/cli/cmd/metrics.go
+++ b/cli/cmd/metrics.go
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -137,15 +138,17 @@ func getPodsFor(ctx context.Context, clientset kubernetes.Interface, namespace s
 	}
 
 	var matchLabels map[string]string
+	var ownerUID types.UID
 	switch res.GetType() {
 	case k8s.CronJob:
 		jobs, err := clientset.BatchV1().Jobs(namespace).List(ctx, metav1.ListOptions{})
 		if err != nil {
 			return nil, err
 		}
+
 		var pods []corev1.Pod
 		for _, job := range jobs.Items {
-			if isOwner(res.GetName(), job.GetOwnerReferences()) {
+			if isOwner(job.GetUID(), job.GetOwnerReferences()) {
 				jobPods, err := getPodsFor(ctx, clientset, namespace, fmt.Sprintf("%s/%s", k8s.Job, job.GetName()))
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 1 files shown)


adleong:
Weird! I'm trying to understand this bug but I think I'm missing something.  As I understand it, this code first fetches the deployment and then grabs it's label selector and uses that to select the pods.  In your case do you have different workloads that use the same label selector?

siggy:
@adleong yup! you can repro with:

```bash
$ curl https://run.linkerd.io/emojivoto.yml | linkerd inject - | kubectl apply -f -
$ curl https://run.linkerd.io/emojivoto-daemonset.yml | linkerd inject - | kubectl apply -f -
$ curl https://run.linkerd.io/emojivoto-statefulset.yml | linkerd inject - | kubectl apply -f -
$ linkerd metrics -n emojivoto deploy/web|grep POD
  # POD web-0 (1 of 3)
  # POD web-d9ffd684f-gnbcx (2 of 3)
  # POD web-fs6l7 (3 of 3)
```

you can see the shared match labels across those 3 workloads with:

```diff
diff --git a/cli/cmd/metrics.go b/cli/cmd/metrics.go
index 95240729..f5da8d96 100644
--- a/cli/cmd/metrics.go
+++ b/cli/cmd/metrics.go
@@ -174,6 +174,8 @@ func getPodsFor(ctx context.Context, clientset kubernetes.Interface, namespace s
                matchLabels = deployment.Spec.Selector.MatchLabels
                ownerUID = deployment.GetUID()

+               fmt.Printf("matchLabels: %+v\n", matchLabels)
+
                replicaSets, err := clientset.AppsV1().ReplicaSets(namespace).List(
                        ctx,
                        metav1.ListOptions{
```

```
matchLabels: map[app:web-svc version:v11]
```

adleong:
Ah, yeah, I think Kubernetes does not like it when you re-use label selectors like that:

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#selector

> Do not overlap labels or selectors with other controllers (including other Deployments and StatefulSets). Kubernetes doesn't stop you from overlapping, and if multiple controllers have overlapping selectors those controllers might conflict and behave unexpectedly.

Basically, the controller will think it is in charge of managing all pods which match the label selector.  In this case, since the statefulset and deployment (for example) have the same label selector, they will both think they are in charge of managing the matching pods.  This can lead to them fighting and to unexpected behavior.

adleong (approved):
After some further discussion, we decided that filtering down to only the pods with owner refs back to the target resource is the least surprising behavior here.  Looks good!  Thanks, @siggy!

alpeb (approved):
LGTM :+1:

PR #5569, opened 2021-01-19
experiment: replace pointer types with values to avoid nil pointer segfaults

Pothulapati:
This commit replaces pointer types with value types in Values.go to avoid
segfaults. This change should not be as easy and should need some more
changes in the templates where these values are used but hoping
it might work and avoid us adding getters which seems
a bit complex to add for every field.

If this wont work or is too complex. Lets go with Getters.

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -359,17 +359,17 @@ func fetchConfigs(ctx context.Context) (*linkerd2.Values, error) {
 func getOverrideAnnotations(values *charts.Values, base *charts.Values) map[string]string {
 	overrideAnnotations := make(map[string]string)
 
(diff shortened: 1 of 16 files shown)


alpeb:
Thanks for investigating this :+1: 
Like you mentioned, it does seem though that the complexity would move over to templates for handling empty structs that would generate unwanted yaml... I'm afraid the most go-ish way of going about this is gonna be through go-generate, which shouldn't be too bad after all. Anyways, eager to hear about the results of this experiment!

Pothulapati:
Closing this because, `.global` has been removed. This could still be a problem with pointer types which we use at some places but this PR does not address them.

Issue #5484, opened 2021-01-06
Add support for k8s 1.19


alpeb:
It appears that with the latest edge everything works fine with k8s 1.19 except for the warnings about the resource versions deprecations, e.g. in `linkerd check` we see things like this in stderr:
```
W0106 11:09:39.204081  292603 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
```
This should be just a matter of updating the versions in the template files, along with the tests golden files.

Note that as of #5483 `bin/k3d` already installs k8s 1.19 by default.

mateiidavid:
👋  I'd like to take this on if that's ok!

kleimkuhler:
That'd be great @mateiidavid!

cpretzer:
Thanks @mateiidavid !

mateiidavid:
Hi, I started working on this yesterday. I think there may be a bit more work than initially thought. 

What I have done at first was to change all `.golden` files and all templates to use v1 instead of  v1beta1 for `apiextensions.k8s.io` and `admissionregistration.k8s.io` resources. 

To test, I tried a fresh install on my k3s cluster. Unfortunately, warnings were replaced with errors 😢:
```
error validating "STDIN": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): missing required field "versions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]; if you choose to ignore these errors, turn validation off with --validate=false
Error from server (Invalid): error when creating "STDIN": CustomResourceDefinition.apiextensions.k8s.io "serviceprofiles.linkerd.io" is invalid: [spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required, spec.versions[1].schema.openAPIV3Schema: Required value: schemas are required]
```

I think just changing the `apiVersion` on the resources might not be enough. 

[...]

PR #5573, opened 2021-01-19
Introduce OpenAPIV3 validation for CRDs

mateiidavid:
## Changes

Sorry I was late with this, it was a bit challenging to figure out what needs to be done (and how). Apologies in advance for the spam, lots of output from tests, will try to keep this short since there is also a lot of yaml to review. Based on the discussion from #5484 in order to migrate CRDs to `v1` they should have a *structural* schema associated. This is asserted throughout official docs pertaining to CRDs:

> A structural schema for CRDs in apiextensions.k8s.io/v1beta1 will not be required. But we plan to require structural schemas for every CRD created in apiextensions.k8s.io/v1

> With apiextensions.k8s.io/v1 the definition of a structural schema is mandatory for CustomResourceDefinitions. In the beta version of CustomResourceDefinition, the structural schema was optional.

I will leave some prior art links at the bottom. In short, a schema is structural if it does not contain any non-empty `type` fields aside from a few exceptions (such as x-kubernetes-int-or-string). There also one more thing to say before I go deeper, according to the [docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning) when it coms to legacy CRDS (created as v1beta1) _pruning_ is not enabled and you can store arbitrary data -- to "migrate", it is specified that `spec.preserveUnknownFields` should be set to false. However, `preserveUnknownFields` can still be present as a field in the schema and it will still be structural. [This page has more details](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning). I'm only mentioning this because you'll encounter the use of `preserveUnknownFields` in the `ServiceProfile` schema and it confused me when I first saw it.

**Changes**:

* Add validation to `TrafficSplit`
* Add validation to `Link`
* Add validation to `ServiceProfile` using kube-gen(ish)
* Validation is top level for all three CRDs

### ServiceProfiles

I tried to use `controller-gen` to re-generate the CRDs, but it kept giving a lot of `{}`. Setting a field type `{}` makes the schema nonstructural, in cases such as these, when inspecting the crd (definition), you will see something similar to:
```
  - lastTransitionTime: "2021-01-19T13:55:13Z"
    message: 'spec.versions[0].schema.openAPIV3Schema.type: Required value: must not
      be empty at the root'
    reason: Violations
    status: "True"
    type: NonStructuralSchema
```

I also tried `kubebuilder` (even though it uses controller-gen) and went down a very deep rabbit hole, not much worked so I decided to just change everything manually. I made the changes by looking at the type definitions in `controller/gen/.../types.go`. For objects such as 
* `ResponseMatch` that contain themselves (recursive types?) I did not really know how to write them and so I declared the types as `object` and added the `x-kube...-preserve-unknown-fields`. This means we can write any arbitrary blob in there but it will not work with pruning as far as I understand. 
* For ints or string I used an appropriate `x-kubernetes-int-or-string` set to true 
* For `opaquePorts` (map[ui32)struct{}) I wrote as a string array since I saw in the code somewhere it was being parsed as a string. Let me know if I should change this.
* For float32s I used `number` with a float format.

#### Test
---
I followed the [debugging-per-route-metrics tutorial](https://linkerd.io/2/tasks/books/#service-profiles). Created a booksapp service then some SPs to test the validation didn't break anything.

```sh
$ curl -sL https://run.linkerd.io/booksapp/webapp.swagger \
  | target/cli/darwin/linkerd -n booksapp profile --open-api - webapp \
  | k -n booksapp apply -f -
serviceprofile.linkerd.io/webapp.booksapp.svc.cluster.local created

$ kubectl get (..)
kind: ServiceProfile
metadata:
  annotations:
spec:
  routes:
  - condition:
      method: GET
      pathRegex: /
    name: GET /
  - condition:
      method: POST
      pathRegex: /authors
    name: POST /authors
  - condition:
      method: GET
      pathRegex: /authors/[^/]*
    name: GET /authors/{id}
....

# Add more SPs -- authors.swagger and books.swagger

$ curl -sL https://run.linkerd.io/booksapp/authors.swagger \
    | target/cli/darwin/linkerd -n booksapp profile --open-api - authors \
    | kubectl -n booksapp apply -f -
serviceprofile.linkerd.io/authors.booksapp.svc.cluster.local created
$  curl -sL https://run.linkerd.io/booksapp/books.swagger \
    | target/cli/darwin/linkerd -n booksapp profile --open-api - books \
    | kubectl -n booksapp apply -f -
serviceprofile.linkerd.io/books.booksapp.svc.cluster.local created

# Configure timeouts and check all is in order

# Before
ROUTE                     SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
DELETE /books/{id}.json     books   100.00%   0.6rps           8ms          10ms          10ms
GET /books.json             books   100.00%   1.2rps           3ms           9ms          10ms
GET /books/{id}.json        books   100.00%   2.2rps           3ms           5ms          18ms
POST /books.json            books    50.69%   2.4rps          10ms          20ms          28ms
PUT /books/{id}.json        books    64.52%   1.0rps          58ms          96ms          99ms
[DEFAULT]                   books         -        -             -             -             -

kubectl -n booksapp edit sp/books.booksapp.svc.cluster.local
serviceprofile.linkerd.io/books.booksapp.svc.cluster.local edited # added 25ms timeout

# After
ROUTE                     SERVICE   EFFECTIVE_SUCCESS   EFFECTIVE_RPS   ACTUAL_SUCCESS   ACTUAL_RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
DELETE /books/{id}.json     books             100.00%          0.6rps          100.00%       0.6rps           7ms          10ms          10ms
GET /books.json             books             100.00%          1.2rps          100.00%       1.2rps           4ms           7ms           9ms
GET /books/{id}.json        books             100.00%          2.7rps          100.00%       2.7rps           3ms           5ms          17ms
POST /books.json            books              48.92%          2.3rps           48.92%       2.3rps           9ms          19ms          20ms
PUT /books/{id}.json        books              36.96%          1.5rps           36.96%       1.5rps          47ms          92ms          98ms
[DEFAULT]                   books                   -               -                -            -             -             -             -

# EFFECTIVE SUCCESS DROPPED
```

---

### TrafficSplits

* TrafficSplits were easy, I just copied the validation from smi-sdk-client-go. I replaced the link at the top to point to a different commit sha with the updated definition. One thing to note here is that the controller & destination crash with the latest version `v1alpha3` since we are using `v1alpha1` informers. Because of this I deleted both v1alpha2 and alpha3 versions from the spec -- in that sense it differs from the smi one. I'm okay to raise an issue to update the code to use recent informers if you'd like.

#### Test
---

I followed the inject faults tutorial to make sure all is in order

```sh
# Before

NAME      MESHED   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TCP_CONN
authors      1/1   100.00%   7.3rps           3ms          23ms          33ms          6
books        1/1   100.00%   8.9rps           5ms          72ms          94ms          6
traffic      1/1         -        -             -             -             -          -
webapp       3/3   100.00%   8.3rps          16ms          70ms          94ms          9

# First try, apply this:
kind: TrafficSplit
metadata:
  name: error-split
  namespace: booksapp
spec:
  service: books
  backends:
  - service: books
    weight: 900m
  - service: error-injector
    weight: 100m —> validation error :( no strings just number (according to smi-spec)

# changed weights to 90 and 10 respectively (number type)

# Resulting routes command after applying -- SUCCESS dropped 10%
ROUTE       SERVICE   SUCCESS      RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99
[DEFAULT]     books    90.93%   6.8rps           5ms         151ms         190ms
```

---

### Link

* I had no controller-gen or smi-spec here but there aren't many fields and they aren't complicated. I made the changes.

#### Test
---

* I did not get to test it works, just that installation works and we can create a link through cli. I don't see why it wouldn't work if it can be successfully generated as a resource by the cli -- not changing information in any way just validating.

```
$ target/cli/darwin/linkerd mc install | k apply -f -
namespace/linkerd-multicluster created
configmap/linkerd-gateway-config created
...
customresourcedefinition.apiextensions.k8s.io/links.multicluster.linkerd.io created

$ target/../linkerd mc link —cluster-name east | k apply -f -
$ kubectl get crd (..) -o yaml
apiVersion: multicluster.linkerd.io/v1alpha1
kind: Link
metadata:
  annotations:
...
spec:
  clusterCredentialsSecret: cluster-credentials-east
  gatewayAddress: 172.22.0.2
  gatewayIdentity: linkerd-gateway.linkerd-multicluster.serviceaccount.identity.linkerd.cluster.local
  gatewayPort: "4143"
  probeSpec:
    path: /health
    period: 3s
    port: "4181"
  selector:
    matchExpressions:
    - key: mirror.linkerd.io/exported
      operator: Exists
  targetClusterDomain: cluster.local
  targetClusterLinkerdNamespace: linkerd
  targetClusterName: east
```

---

Hopefully after this, we can upgrade all resources to `v1` and end #5484. I mostly managed to answer all the questions I had, I think it would have been easier to just set the `spec.preserveUnknownFields` to true, but I think at some point this would have had to be done anyway? Let me know if I overlooked anything or if this isn't really needed 🤒 

*Prior art*:
 * https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
 * https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning
 * https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/
 * https://kubernetes.io/blog/2019/06/20/crd-structural-schema/
 * [smi-spec-crd](https://github.com/servicemeshinterface/smi-sdk-go/blob/d4e76b1cd7a33ead5f38d1262dd838a31c80f4e5/crds/split.yaml)
 * [apiextension-types.go](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go) -- although it's not in there, having `schema` under validation field is considered top-level. Otherwise we'd have to have it under each specific version. This is not documented anywhere (and not really common sense, it's not top level per se, so I'm 50/50 sure about this). Please let me know if you have a different take on it.

*Note*: it is very probable I made some mistakes and I wasn't thorough enough with the tests, if you could please double check it all works as expected that'd be a huge relief.

### DIFF
--- a/charts/linkerd2/templates/trafficsplit-crd.yaml
+++ b/charts/linkerd2/templates/trafficsplit-crd.yaml
@@ -1,29 +1,66 @@
 ---
 ###
 ### TrafficSplit CRD
-### Copied from https://github.com/deislabs/smi-sdk-go/blob/cea7e1e9372304bbb6c74a3f6ca788d9eaa9cc58/crds/split.yaml
-###
+### Copied from github.com/servicemeshinterface/smi-sdk-go/blob/d4e76b1cd7a33ead5f38d1262dd838a31c80f4e5/crds/split.yaml
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: trafficsplits.split.smi-spec.io
-  annotations:
-    {{.Values.global.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.global.linkerdVersion) .Values.global.cliVersion}}
-  labels:
-    {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
 spec:
   group: split.smi-spec.io
-  version: v1alpha1
   scope: Namespaced
   names:
     kind: TrafficSplit
+    listKind: TrafficSplitList
     shortNames:
-      - ts
+    - ts
     plural: trafficsplits
     singular: trafficsplit
-  additionalPrinterColumns:
-  - name: Service
-    type: string
-    description: The apex service of this split.
-    JSONPath: .spec.service
+  version: v1alpha1
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+  preserveUnknownFields: false
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          required:
+            - service
+            - backends
+          properties:
+            service:
+              description: The apex service of this split.
+              type: string
+            matches:
+              description: The HTTP route groups that this traffic split should match.
+              type: array
+              items:
+                type: object
+                required: ['kind', 'name']
+                properties:
+                  kind:
+                    description: Kind of the matching group.
+                    type: string
+                    enum:
+                      - HTTPRouteGroup
+                  name:
+                    description: Name of the matching group.
+                    type: string
+            backends:
+              description: The backend services of this split.
+              type: array
+              items:
+                type: object
+                required: ['service', 'weight']
+                properties:
+                  service:
+                    description: Name of the Kubernetes service.
+                    type: string
+                  weight:
+                    description: Traffic weight value of this backend.
+                    type: number
--- a/charts/linkerd2/templates/trafficsplit-crd.yaml
+++ b/charts/linkerd2/templates/trafficsplit-crd.yaml
@@ -2,11 +2,16 @@
 ###
 ### TrafficSplit CRD
 ### Copied from github.com/servicemeshinterface/smi-sdk-go/blob/d4e76b1cd7a33ead5f38d1262dd838a31c80f4e5/crds/split.yaml
+###
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: trafficsplits.split.smi-spec.io
+  annotations:
+    {{.Values.global.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.global.linkerdVersion) .Values.global.cliVersion}}
+  labels:
+    {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
 spec:
   group: split.smi-spec.io
   scope: Namespaced
@@ -14,7 +19,7 @@ spec:
     kind: TrafficSplit
     listKind: TrafficSplitList
     shortNames:
-    - ts
+      - ts
     plural: trafficsplits
     singular: trafficsplit
   version: v1alpha1
@@ -63,4 +68,9 @@ spec:
                     type: string
                   weight:
                     description: Traffic weight value of this backend.
-                    type: number
+                    x-kubernetes-int-or-string: true
+  additionalPrinterColumns:
+  - name: Service
+    type: string
+    description: The apex service of this split.
+    JSONPath: .spec.service
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -40,8 +40,8 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-buildx-${{ matrix.target }}-
     - name: Build docker images
-      env:
-        DOCKER_TRACE: 1
+      #env:
+      #  DOCKER_TRACE: 1
       run: |
         docker buildx create --driver docker-container --use
         bin/docker-build-${{ matrix.target }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,7 +39,7 @@ jobs:
       uses: crazy-max/ghaction-docker-buildx@552c9de
     - name: Build & Push Multi Arch Images
       env:
-        DOCKER_TRACE: 1
+        #DOCKER_TRACE: 1
         DOCKER_MULTIARCH: 1
         DOCKER_PUSH: 1
       run: |
--- a/charts/linkerd2/templates/serviceprofile-crd.yaml
+++ b/charts/linkerd2/templates/serviceprofile-crd.yaml
@@ -21,6 +21,132 @@ spec:
     served: true
     storage: true
   scope: Namespaced
+  preserveUnknownFields: false
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          description: Spec is the custom resource spec
+          required:
+          - routes
+          properties:
+            dstOverrides:
+              type: array
+              required:
+              - authority
+              - weight
+              items:
+                type: object
+                description: WeightedDst is a weighted alternate destination.
+                properties:
+                  authority:
+                    type: string
+                  weight:
+                    x-kubernetes-int-or-string: true
+                    anyOf:
+                    - type: integer
+                    - type: string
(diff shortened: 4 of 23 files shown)


olix0r:
It looks like this test is failing:

```
--- FAIL: TestTrafficSplitCli (1.08s)
196
    trafficsplit_test.go:160: 'kubectl apply' command failed
197
        deployment.apps/backend created
198
        service/backend-svc created
199
        deployment.apps/failing created
200
        service/failing-svc created
201
        deployment.apps/slow-cooker created
202
        service/slow-cooker created
203
        error: error validating "STDIN": error validating data: [ValidationError(TrafficSplit.spec.backends[0].weight): invalid type for io.smi-spec.split.v1alpha1.TrafficSplit.spec.backends.weight: got "string", expected "number", ValidationError(TrafficSplit.spec.backends[1].weight): invalid type for io.smi-spec.split.v1alpha1.TrafficSplit.spec.backends.weight: got "string", expected "number"]; if you choose to ignore these errors, turn validation off with --validate=false
```

charts/linkerd2/templates/trafficsplit-crd.yaml
```diff
@@ -14,16 +13,58 @@ metadata:
...
+              items:
+                type: object
+                required: ['service', 'weight']
+                properties:
+                  service:
+                    description: Name of the Kubernetes service.
+                    type: string
+                  weight:
+                    description: Traffic weight value of this backend.
+                    type: number
```
adleong, line 70: I think the failing test is because of this line.  In version v1alpha1, this is allowed to be a quantity string: https://github.com/servicemeshinterface/smi-spec/blob/master/apis/traffic-split/v1alpha1/traffic-split.md

mateiidavid: Thank you for pointing this out. I changed the weight to include `x-kubernetes-int-or-string`.

Initially I wanted to put this under the version itself and get rid of the top-level validation, e.g
```yaml
- name: v1alpha1
  schema:
    openAPIV3Schema:
    ...
```

however, this prompted error messages when doing `linkerd check --pre`; think resources with v1beta1 can't make use of the new versioned schema field. [Check this comment out for more info on this topic](https://github.com/kubernetes/kubernetes/issues/82443#issuecomment-529033688).

adleong (commented):
Looks great!  Thanks for your thorough work, as always, @mateiidavid!

Let's just revert the DOCKER_TRACE changes before merging.  Otherwise looks good to me.

alpeb:
@adleong I think for the time being we'll have to leave DOCKER_TRACE disabled until we figure what's going on with the runners. Otherwise CI will fail.

adleong:
Thanks for the explanation, @alpeb! :+1:

adleong (approved):

alpeb (approved):
Awesome work! :medal_military:

PR #5583, opened 2021-01-21
build: Remove the DOCKER_TRACE environment variable

olix0r:
Our build scripts hide docker's output by default and only pass through
output when DOCKER_TRACE is set. Practically everyone else tends to use
DOCKER_TRACE=1 persistently. And, recently, GitHub Actions stopped
working with `/dev/stderr`

This change removes the DOCKER_TRACE environment variable so that output
is always emitted as it would when invoking docker directly.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -121,7 +121,7 @@ Started](https://linkerd.io/2/getting-started/)
 bin/k3d cluster create
 
 # build all docker images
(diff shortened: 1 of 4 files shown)


alpeb (approved):
Makes sense :+1:

Pothulapati (approved):
LGTM

PR #5584, opened 2021-01-21
helm chart changes to run proxy-init container as non root

pankajmt:
Signed-off-by: Pankaj Tolani <pankaj.tolani@gmail.com>

So the idea is to be able to run proxy-init container iptables changes as non root with NET_RAW and NET_ADMIN capabilities. This will allow for workloads to run with a more constrained and secure PSP. Changes to the proxy-init container image are outlined here. 

Now, controller/proxy-injector binary has an embedded copy of the helm chart which gets rendered at runtime to define the securityContext for the proxy-init initContainer. This change allows for configuring the runAsNonRoot and runAsUser fields of the securityContext.

### DIFF
--- a/charts/linkerd2/values.yaml
+++ b/charts/linkerd2/values.yaml
@@ -146,6 +146,8 @@ global:
     xtMountPath:
       mountPath: /run
       name: linkerd-proxy-init-xtables-lock
+    runAsNonRoot: false
+    runAsUser: 0
 
   # -- Annotation label for the proxy create. Do not edit.
   createdByAnnotation: linkerd.io/created-by
--- a/charts/partials/templates/_proxy-init.tpl
+++ b/charts/partials/templates/_proxy-init.tpl
@@ -40,8 +40,8 @@ securityContext:
   privileged: false
   {{- end }}
   readOnlyRootFilesystem: true
-  runAsNonRoot: {{.Values.global.proxyInit.runAsNonRoot}}
-  runAsUser: {{.Values.global.proxyInit.runAsUser}}
+  runAsNonRoot: true
+  runAsUser: 76543
 terminationMessagePolicy: FallbackToLogsOnError
 {{- if or (not .Values.global.cniEnabled) .Values.global.proxyInit.saMountPath }}
 volumeMounts:
--- a/charts/partials/templates/_proxy-init.tpl
+++ b/charts/partials/templates/_proxy-init.tpl
@@ -21,11 +21,7 @@ imagePullPolicy: {{.Values.global.proxyInit.image.pullPolicy}}
 name: linkerd-init
 {{ include "partials.resources" .Values.global.proxyInit.resources }}
 securityContext:
-  {{- if .Values.global.proxyInit.closeWaitTimeoutSecs }}
   allowPrivilegeEscalation: true
-  {{- else }}
-  allowPrivilegeEscalation: false
-  {{- end }}
   capabilities:
     add:
     - NET_ADMIN
@@ -44,8 +40,8 @@ securityContext:
   privileged: false
   {{- end }}
   readOnlyRootFilesystem: true
-  runAsNonRoot: false
-  runAsUser: 0
+  runAsNonRoot: {{.Values.global.proxyInit.runAsNonRoot}}
+  runAsUser: {{.Values.global.proxyInit.runAsUser}}
 terminationMessagePolicy: FallbackToLogsOnError
 {{- if or (not .Values.global.cniEnabled) .Values.global.proxyInit.saMountPath }}
 volumeMounts:
@@ -58,5 +54,5 @@ volumeMounts:
 - mountPath: {{.Values.global.proxyInit.saMountPath.mountPath}}
   name: {{.Values.global.proxyInit.saMountPath.name}}
   readOnly: {{.Values.global.proxyInit.saMountPath.readOnly}}
-{{- end -}}  
+{{- end -}}
 {{- end -}}
--- a/charts/linkerd2/templates/psp.yaml
+++ b/charts/linkerd2/templates/psp.yaml
@@ -10,7 +10,7 @@ metadata:
   labels:
     {{.Values.global.controllerNamespaceLabel}}: {{.Values.global.namespace}}
 spec:
-  allowPrivilegeEscalation: false
+  allowPrivilegeEscalation: true
   readOnlyRootFilesystem: true
   {{- if empty .Values.global.cniEnabled }}
   allowedCapabilities:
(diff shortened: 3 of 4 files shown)


Pothulapati:
Hey @pankajmt, Thanks for raising the PR.

Can you update the PR description to reflect the problem and the solution? Otherwise it could be hard for reviewers to understand and also in the codebase after it is merged.

pankajmt:
> Hey @pankajmt, Thanks for raising the PR.
> 
> Can you update the PR description to reflect the problem and the solution? Otherwise it could be hard for reviewers to understand and also in the codebase after it is merged.

Updated. Hopefully better now.

adleong:
Before this PR can merge, we will first need to merge https://github.com/linkerd/linkerd2-proxy-init/pull/31, create a new release of the proxy-init, and then update this PR to use the new release of proxy-init.

alpeb:
@pankajmt can I ask you to post an example of such a restrictive PSP, so we have give it a test?

pankajmt:
> @pankajmt can I ask you to post an example of such a restrictive PSP, so we have give it a test?

```
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  annotations:
    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: runtime/default
    seccomp.security.alpha.kubernetes.io/defaultProfileName: runtime/default
  creationTimestamp: "2021-01-22T07:06:20Z"
  labels:
    gb/managed-by: terraform
  name: l1nkerd.psp.linkerd-cplane
  resourceVersion: "6494917"
  selfLink: /apis/policy/v1beta1/podsecuritypolicies/l1nkerd.psp.linkerd-cplane
  uid: 0f0419a0-45fa-4eb6-922b-4d46b7e3926e
spec:
  allowPrivilegeEscalation: true
  allowedCapabilities:
  - NET_ADMIN
  - NET_RAW
  allowedHostPaths:
  - pathPrefix: /dev/null
    readOnly: true
  defaultAllowPrivilegeEscalation: false
  forbiddenSysctls:
  - '*'
  fsGroup:
    ranges:
    - max: 100000
      min: 65534
    rule: MustRunAs
  readOnlyRootFilesystem: true
  requiredDropCapabilities:
  - ALL
  runAsUser:
    rule: MustRunAsNonRoot
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    ranges:
    - max: 100000
      min: 65534
    rule: MustRunAs
  volumes:
  - configMap
  - downwardAPI
  - emptyDir
  - persistentVolumeClaim
  - secret
  - projected
```

alpeb:
Sorry for the late reply, as was away for a few days ( :beach_umbrella: )

To summary my understanding one more time, all this is about having the init container run as non-root, in exchange of allowing privilege escalation, which will be focused on the iptables command via setuid, which sounds great as it establishes a very narrow perimeter as opposed to having full root access, correct?

Question: Can we safely assume every cluster can be set up to allow privilege escalation? If so, then we can safely change the defaults here, setting `runAsNonRoot: true` (and we also require `runAsUser: 76543` to be in sync with the changes in linkerd/linkerd2-proxy-init#31). 

I think we can also drop the `closeWaitTimeoutSecs` setting now. 

.. and we also need to update the PSP that ships with linkerd (`charts/linkerd2/templates/psp.yml`), setting `allowPrivilegeEscalation: true`. 

If we go down this route, we should also document the `runAsNonRoot` and `runAsUser` configs, explaining what they do, and warning that changes should be kept in sync with the proxy-init's Dockerfile.

pankajmt:
> Sorry for the late reply, as was away for a few days ( 🏖️ )
> 
> To summarize my understanding one more time, all this is about having the init container run as non-root, in exchange of allowing privilege escalation, which will be focused on the iptables command via ~setuid~ setcap, which sounds great as it establishes a very narrow perimeter as opposed to having full root access, correct?
> 
> Question: Can we safely assume every cluster can be set up to allow privilege escalation? If so, then we can safely change the defaults here, setting `runAsNonRoot: true` (and we also require `runAsUser: 76543` to be in sync with the changes in [linkerd/linkerd2-proxy-init#31](https://github.com/linkerd/linkerd2-proxy-init/pull/31)).
> 
yes. changes incorporated. 

> I think we can also drop the `closeWaitTimeoutSecs` setting now.
> 
agree but i do not know the full reasoning behind the parameter and what all testing needs to be modified inline with this change. so would like to leave out of this pr. 

> .. and we also need to update the PSP that ships with linkerd (`charts/linkerd2/templates/psp.yml`), setting `allowPrivilegeEscalation: true`.
> 
yes. done. can someone please test this if it does not as part of automation? i have not had a chance to validate this change atm. 

> If we go down this route, we should also document the `runAsNonRoot` and `runAsUser` configs, explaining what they do, and warning that changes should be kept in sync with the proxy-init's Dockerfile.
let me know where do and i can do that change too.

pankajmt:
also not sure what is failure is about 

https://github.com/linkerd/linkerd2/runs/1814202648?check_suite_focus=true

pankajmt:
and looks like some unit tests need attention too. will need to be later :-(

https://github.com/linkerd/linkerd2/runs/1814202213?check_suite_focus=true

charts/linkerd2/values.yaml
```diff
@@ -146,6 +146,8 @@ global:
     xtMountPath:
       mountPath: /run
       name: linkerd-proxy-init-xtables-lock
+    runAsNonRoot: false
+    runAsUser: 0
```
alpeb, line 150: the value here should be `76543`

charts/partials/templates/_proxy-init.tpl
```diff
@@ -44,8 +40,8 @@ securityContext:
   privileged: false
   {{- end }}
   readOnlyRootFilesystem: true
-  runAsNonRoot: false
-  runAsUser: 0
+  runAsNonRoot: true
+  runAsUser: 76543
```
alpeb, line 44: Please use the template values as in your previous commit

alpeb (commented):
About the README errors, you just need to run `bin/helm-docs` for the README files to grab the  two new helm values, and commit the automatically updated files.

alpeb:
Can you also please update `viz/charts/linkerd-viz/templates/psp.yaml`?

adleong:
Hi @pankajmt!  Just wanted to check in and see if you were still working on this and https://github.com/linkerd/linkerd2-proxy-init/pull/31 or if we should close these PRs.

adleong:
Closing due to inactivity.

Issue #5544, opened 2021-01-15
Tap controller should exclude pods with tap disabled


adleong:
If you run `linkerd viz tap` targeting a workload which has tap disabled or has not been injected by the tap-injector, the tap command simply waits and produces no output.

In some sense, this behavior is correct.  However, it can be unexpected, and users will encounter this situation more frequently now that the tap-injector is no longer part of the default proxy-injector.  

To remedy this, the tap controller can attempt to determine if a target pod is tapable by examining the environment variables of its proxy sidecar.  If a target pod is not tapable, it should be excluded from the list of tap targets.  If the list of tap targets is empty, the tap controller should return a helpful error message to the user informing them that there are no tapable pods in the target workload and to ensure that the pods exist and have been injected by the tap-injector.

Note that with https://github.com/linkerd/linkerd2/issues/5325#issuecomment-760575199 implemented, the `linkerd viz check --proxy` command can be used to diagnose this.

olix0r:
Rather than relying on environment variables, it would be preferable to use workload annotations, if possible. Can/should the tap injector set a disable annotation on pods?

PR #5602, opened 2021-01-25
viz: add data-plane and prometheus healthchecks

### LINKED ISSUE #5325: Add linkerd viz check command

Pothulapati:
Fixes #5325

This branch adds the remaining healthchecks for the viz extension
i.e

- Data-plane metrics check in prometheus
- `--proxy` mode which also checks for tap injections based
  on annotations.
  
For this, The following other changes were needed

- `Category.ID` is made public so that `--proxy` toggleness can be
allowed
- Made tap svc env key into `pkg/inject/env` so that it can be re-used for
checks


pkg/inject/env.go
```diff
@@ -0,0 +1,6 @@
+package inject
+
+const (
```
Pothulapati, line 3: There are some more of these env const keys used in jaeger injector which can be moved here. I will do this as a separate PR.

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -54,147 +73,242 @@ func (hc *HealthChecker) RunChecks(observer healthcheck.CheckObserver) bool {
...
+				WithHintAnchor("l5d-data-plane-prom").
+				Warning().
+				WithRetryDeadline(hc.RetryDeadline).
+				WithCheck(func(ctx context.Context) (err error) {
+					pods, err := hc.getDataPlanePodsFromVizAPI(ctx)
+					if err != nil {
+						return err
+					}
+
+					// TODO: Check if prometheus is present
```
Pothulapati, line 195: Is this actually important? My thinking is that, As this query goes through viz metrics-api, This also supports external prometheus use-case.

It feels okay to have this check fail, when they have viz extension installed with no default/external prometheus as most functionality is lost in that case🤔

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -40,7 +46,20 @@ type HealthChecker struct {
...
-	parentHC.AppendCategories(*hc.vizCategory())
+
+	vizCategories := hc.vizCategory()
+	checkMap := map[healthcheck.CategoryID]struct{}{}
+	for _, category := range categoryIDs {
+		checkMap[category] = struct{}{}
+	}
+
+	// Append Only the categories that aare passed
+	for i := range vizCategories {
```
dadjeibaah, line 57: I found it a little difficult to read why we needed to go over the list of `vizCategories` to append viz specific checks to the parent healthchecker. I think it might be a little clear if we create a category map from all the viz categories and then loop over `categoryIDs` to figure out what checks we'd need to append. 
```diff
diff --git a/viz/pkg/healthcheck/healthcheck.go b/viz/pkg/healthcheck/healthcheck.go
index 1a476c87..2d1fb9fe 100644
--- a/viz/pkg/healthcheck/healthcheck.go
+++ b/viz/pkg/healthcheck/healthcheck.go
@@ -47,16 +47,15 @@ func NewHealthChecker(categoryIDs []healthcheck.CategoryID, options *healthcheck
 	parentHC := healthcheck.NewHealthChecker(categoryIDs, options)
 	hc := &HealthChecker{HealthChecker: parentHC}

-	vizCategories := hc.vizCategory()
-	checkMap := map[healthcheck.CategoryID]struct{}{}
-	for _, category := range categoryIDs {
-		checkMap[category] = struct{}{}
+	vizCheckMap := map[healthcheck.CategoryID]healthcheck.Category{}
+	for _, category := range hc.vizCategories() {
+		vizCheckMap[category.ID] = category
 	}

-	// Append Only the categories that aare passed
-	for i := range vizCategories {
-		if _, ok := checkMap[vizCategories[i].ID]; ok {
-			parentHC.AppendCategories(vizCategories[i])
+	// only append incoming check categories that are also in the vizCheckMap
+	for _, id := range categoryIDs {
+		if _, ok := vizCheckMap[id]; ok {
+			parentHC.AppendCategories(vizCheckMap[id])
 		}
 	}

@@ -73,7 +72,7 @@ func (hc *HealthChecker) RunChecks(observer healthcheck.CheckObserver) bool {
 	return hc.HealthChecker.RunChecks(observer)
 }

-func (hc *HealthChecker) vizCategory() []healthcheck.Category {
+func (hc *HealthChecker) vizCategories() []healthcheck.Category {
 	return []healthcheck.Category{
 		*healthcheck.NewCategory(LinkerdVizExtensionCheck, []healthcheck.Checker{
 			*healthcheck.NewChecker("linkerd-viz Namespace exists").
```

dadjeibaah (commented):
This is looking great! Left a few suggestions for your consideration.

viz/pkg/healthcheck/healthcheck.go
Pothulapati: Yep, Makes sense. This was an unnecessary complication after the copy-pasta of the boiler-plate code from the core health-checker.

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -54,147 +72,242 @@ func (hc *HealthChecker) RunChecks(observer healthcheck.CheckObserver) bool {
...
+			pods = append(pods, pod)
+		}
+	}
+
+	return pods, nil
+}
+
+// checkForTapConfiguration checks if the tap env variable is present in the linkerd proxy
+// only for the pods with tap enabled
+func (hc *HealthChecker) checkForTapConfiguration(ctx context.Context, pods []corev1.Pod) error {
```
adleong, line 266: Note that when we implement https://github.com/linkerd/linkerd2/issues/5544, we will add an annotation to pods which have been tap-injected.  This will simplify this method since we can just look for that annotation instead of looking at the environment variables.

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -54,147 +72,242 @@ func (hc *HealthChecker) RunChecks(observer healthcheck.CheckObserver) bool {
...
+				WithCheck(func(ctx context.Context) error {
+					if hc.DataPlaneNamespace == "" {
+						// when checking proxies in all namespaces, this check is a no-op
+						return nil
+					}
+					return hc.CheckNamespace(ctx, hc.DataPlaneNamespace, true)
+				}),
+			*healthcheck.NewChecker("data-plane pods have the correct tap configuration").
+				WithHintAnchor("l5d-viz-data-plane-tap").
+				Fatal().
```
adleong, line 223: let's make this non-fatal and make it a warning.  It will be fairly normal for there to be pods which already existed before the tap-injector started up.  it will not be possible to tap those pods until they are restarted, but that's an okay state to be in as long as the tap command provides clear error messages (as is specified in https://github.com/linkerd/linkerd2/issues/5544)

adleong (commented):
If I install linkerd-viz and then look at one of the pods in the linkerd-viz namespace, it doesn't have the TAP evn variables:

```
k -n linkerd-viz get po/linkerd-grafana-5df7d9484d-xjnwg -o yaml | grep TAP
```

this is expected since it was create at the same time as the tap-injector.  however, this is not reported by the viz check command:

```
bin/linkerd viz check --proxy -n linkerd-viz
...
linkerd-viz-data-plane
----------------------
√ data plane namespace exists
√ data-plane pods have the correct tap configuration

Status check results are √
```

pkg/k8s/api.go
```diff
@@ -245,6 +245,27 @@ func GetPodStatus(pod corev1.Pod) string {
 	return reason
 }
 
+// CheckForEnv checks for the given key in the given Container env variables
+func CheckForEnv(container *corev1.Container, envKey string) bool {
+	var presence bool
+	for _, env := range container.Env {
+		if env.Name == envKey {
+			presence = true
```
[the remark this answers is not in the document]
Pothulapati: This is actually going to change, once #5608 is merged. The plan is to check only for annotation that the injector has added instead of explicitly checking for the environment variable. WDYT?

viz/pkg/healthcheck/healthcheck.go
Pothulapati: Yep.

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -54,147 +72,242 @@ func (hc *HealthChecker) RunChecks(observer healthcheck.CheckObserver) bool {
...
+					return healthcheck.CheckPodsRunning(pods, "")
+				}),
+			*healthcheck.NewChecker("can initialize the client").
+				WithHintAnchor("l5d-viz-existence-client").
+				Fatal().
+				WithCheck(func(ctx context.Context) (err error) {
+					hc.vizAPIClient, err = client.NewExternalClient(ctx, hc.vizNamespace, hc.KubeAPIClient())
+					return
+				}),
+			*healthcheck.NewChecker("data plane proxy metrics are present in Prometheus").
```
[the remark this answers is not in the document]
Pothulapati: Updated.

Pothulapati:
> this is expected since it was created at the same time as the tap-injector. however, this is not reported by the viz check command:

The logic currently explicitly skips these warnings if they are from the core cp namespace or the viz extension namespace.  As they will always error. Should we report them too? 🤔

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -40,7 +46,19 @@ type HealthChecker struct {
 func NewHealthChecker(categoryIDs []healthcheck.CategoryID, options *healthcheck.Options) *HealthChecker {
 	parentHC := healthcheck.NewHealthChecker(categoryIDs, options)
 	hc := &HealthChecker{HealthChecker: parentHC}
-	parentHC.AppendCategories(*hc.vizCategory())
+
+	vizCheckMap := map[healthcheck.CategoryID]healthcheck.Category{}
```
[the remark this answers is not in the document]
dadjeibaah: 👍  on this suggestion. I think, imo, it would improve readability if we explicitly pass in the actual categories we'd like to check for. My one comment is that we need to ensure all other sub checks follow this pattern as well. Definitely for some other PR down the line.

pkg/k8s/api.go
```diff
@@ -245,6 +245,27 @@ func GetPodStatus(pod corev1.Pod) string {
 	return reason
 }
 
+// CheckForEnv checks for the given key in the given Container env variables
+func CheckForEnv(container *corev1.Container, envKey string) bool {
```
dadjeibaah, line 249: Can we rename this method to `IsEnvPresent` or something similar?

dadjeibaah (commented):
Thanks for making those updates. I've got a couple more suggestions.

kleimkuhler:
Just a heads up that some of the existing comments on this will be addressed by #5608, which adds the annotation when tap is enabled on a pod.

adleong:
@Pothulapati I don't think we should exclude the control plane namespace or the viz namespace since this is somewhat surprising behavior, especially if the user specifically targets those namespaces.  Instead, I think we should soften the wording of the warning to make it sound less like an error.  Something like: 

```
‼ data-plane pods have tap enabled
    Some data plane pods do not have tap configured and cannot be tapped:
emoji-5f7555f8b7-zpcwg
vote-bot-558f8fffd8-m9lwr
voting-84dc67d586-4qvdw
web-6b5cc5db8-jjj4d
```

Having the list of pods less indented than the error message also looks weird.  I think we should indent that list and possibly bullet it with `*` characters as we do elsewhere.

viz/pkg/healthcheck/healthcheck.go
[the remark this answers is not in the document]
Pothulapati: Updated. But the current state is also a bit confusing for the following reason
- Now, specifying Categories for the viz hc are configured in two different ways. One way to specify the core cp checks i.e through the CategoryIDs argument and another way is for the Viz checks by directly adding them using `hc.AppendCategories`

pkg/k8s/api.go
Pothulapati: This is going to be removed once #5608 is merged as we only plan to check the newly added annotation by the injector.

Pothulapati:
@adleong Updated the formatting.

After removing the toggle based on the namespace this is the output after a normal installation, which could worry some users?
```
❯ ./bin/go-run cli viz check --proxy
kubernetes-api
--------------
√ can initialize the client

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-viz
-----------
√ linkerd-viz ConfigMaps exist
√ tap API service is running
√ can initialize the client
√ viz extension self-check

linkerd-viz-data-plane
----------------------
√ data plane namespace exists
√ data plane proxy metrics are present in Prometheus
‼ data-plane pods have the correct tap configuration
    Some data plane pods do not have tap configured and cannot be tapped:
        * emoji-696d9d8f95-9hz6p
        * linkerd-grafana-786c99fcf6-jtsj8
        * linkerd-metrics-api-7b9b896f9c-fdltb
        * linkerd-prometheus-766749f65f-65xpv
        * linkerd-tap-857b648ddc-djnpz
        * linkerd-web-54f4684d56-9rgkv
        * linkerd-controller-854c757cb-hc7sp
        * linkerd-destination-5bf94fc484-gdsbm
        * linkerd-identity-6496b9f98d-67sb5
        * linkerd-proxy-injector-58d7fb5d96-vblzk
        * linkerd-sp-validator-7775889864-7hmpb
    see https://linkerd.io/checks/#l5d-viz-data-plane-tap for hints

Status check results are √
```

dadjeibaah (approved):
Thanks for making those changes. LGTM! 📦 🚢

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -192,9 +182,123 @@ func (hc *HealthChecker) vizCategory() *healthcheck.Category {
...
+		if err != nil {
+			return err
+		}
+		// Check if Tap is disabled
+		if !k8s.IsTapDisabled(pod) && !k8s.IsTapDisabled(ns) {
+			// Check for tap-injector annotation
+			if _, contains := pod.GetAnnotations()[vizLabels.VizTapEnabled]; !contains {
+				podsWithoutTap = append(podsWithoutTap, fmt.Sprintf("* %s", pod.Name))
+			}
+		}
```
kleimkuhler, line 273: Now that we can go off the annotation, I think this can be simplified a little more.

The tap injector will not add the `tap-enabled` annotation to a pod if tap is disabled via annotation on the pod or the namespace. Therefore, we don't need to check for the `disable-tap` annotation on the namespace or the pod. We should be okay to skip getting the namespace and only check the the pod has the annotation.

Additionally, there is already a `vizLabels.IsTapEnabled` function we can use here. It also checks the value of the annotation is actually `true`.

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -192,9 +182,123 @@ func (hc *HealthChecker) vizCategory() *healthcheck.Category {
...
+			WithHintAnchor("l5d-data-plane-prom").
+			Warning().
+			WithRetryDeadline(hc.RetryDeadline).
+			WithCheck(func(ctx context.Context) (err error) {
+				pods, err := hc.getDataPlanePodsFromVizAPI(ctx)
+				if err != nil {
+					return err
+				}
+
+				// TODO: Check if prometheus is present
```
kleimkuhler, line 214: This seems like an important TODO? Maybe make a follow-up issue for this since this could possibly hang if Prometheus is disabled.

It's similar to my comment [here](https://github.com/linkerd/linkerd2/pull/5627#discussion_r566935296). If we can check the values config, we can end these checks early.

viz/pkg/healthcheck/healthcheck.go
kleimkuhler: Talked off GitHub about it, but never the part about removing the tap disabled check. We'll need that since we'd then issue warnings on pods that don't have the `tap-enabled` annotation, but do have the `disable-tap` annotation.

viz/pkg/healthcheck/healthcheck.go
kleimkuhler: This will be taken care of in a follow-up next week 👍

kleimkuhler (approved):
🤙

adleong:
@Pothulapati this looks good to me!  My suggestion to make this look less alarming to users is to adjust the wording on:

```
‼ data-plane pods have the correct tap configuration
```

Currently, it's implied that certain pods have the INCORRECT configuration whereas it's not really that anything is incorrect, it's just that those pods can't be tapped.  The wording I suggested above is `‼ data-plane pods have tap enabled`

Our examples and guides in the documentation should almost always specify a specific namespace when doing data plane checks so users are less likely to be in this situation in the first place.  This also helps trait users that checking the data plane on a specific namespace is the typical usage of this command.

adleong (approved):

PR #5608, opened 2021-01-26
viz: only tap pods that have tap explicitly enabled

### LINKED ISSUE #5544: Tap controller should exclude pods with tap disabled

kleimkuhler:
## What this changes

This allows the tap controller to inform `tap` users when pods either have tap
disabled or tap is not enabled yet.

## Why

When a user taps a resource that has not been admitted by the Viz extension's
`tap-injector`, tap is not explicitly disabled but it is also not enabled.
Therefore, the `tap` command hangs and provides no feedback to the user.

Closes #5544

## How

A new `viz.linkerd.io/tap-enabled` annotation is introduced which is
automatically added by the Viz extension's `tap-injector`. This annotation is
added to a pod when it is able to be tapped; this means that the pod and the
pod's namespace do not have the `config.linkerd.io/disable-tap` annotation
added.

When a user attempts to tap a resource, the tap controller now looks for this
new annotation; if the annotation is present on the pod then that pod is
tappable.

If the annotation is not present or tap is explicitly disabled, an error is
returned.

## UI changes

Multiple errors can now occur when trying to tap a resource:

1. There are no pods for the resource.
2. There are pods for the resource, but tap is disabled via pod or namespace
   annotation.
3. There are pods for the resource, but tap is not yet enabled because the
   `tap-injector` did not admit the resource.

Errors are now handled as shown below:

Tap is disabled:

```
❯ bin/linkerd viz tap deploy/test
Error: no pods to tap for deployment/test
pods found with tap disabled via the config.linkerd.io/disable-tap annotation
```

Tap is not enabled:

```
❯ bin/linkerd viz tap deploy/test
Error: no pods to tap for deployment/test
pods found with tap not enabled; try restarting resource so that it can be injected
```

There are a mix of pods with tap disabled or tap not enabled:

```
❯ bin/linkerd viz tap deploy/test
Error: no pods to tap for deployment/test
pods found with tap disabled via the config.linkerd.io/disable-tap annotation
pods found with tap not enabled; try restarting resource so that it can be injected
```

### DIFF
--- a/controller/tap/server.go
+++ b/controller/tap/server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 	"unicode/utf8"
 
@@ -19,6 +20,7 @@ import (
 	"github.com/linkerd/linkerd2/pkg/prometheus"
 	"github.com/linkerd/linkerd2/pkg/util"
 	pb "github.com/linkerd/linkerd2/viz/metrics-api/gen/viz"
+	vizLabels "github.com/linkerd/linkerd2/viz/pkg/labels"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -75,7 +77,8 @@ func (s *GRPCTapServer) TapByResource(req *pb.TapByResourceRequest, stream pb.Ta
 	}
 
 	pods := []*corev1.Pod{}
-	foundDisabledPods := false
+	tapDisabled := false
+	tapNotEnabled := false
 	for _, object := range objects {
 		podsFor, err := s.k8sAPI.GetPodsFor(object, false)
 		if err != nil {
@@ -85,7 +88,9 @@ func (s *GRPCTapServer) TapByResource(req *pb.TapByResourceRequest, stream pb.Ta
 		for _, pod := range podsFor {
 			if pkgK8s.IsMeshed(pod, s.controllerNamespace) {
 				if pkgK8s.IsTapDisabled(pod) {
-					foundDisabledPods = true
+					tapDisabled = true
+				} else if !vizLabels.IsTapEnabled(pod) {
+					tapNotEnabled = true
 				} else {
 					pods = append(pods, pod)
 				}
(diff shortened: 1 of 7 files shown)


controller/tap/server.go
```diff
@@ -94,13 +99,18 @@ func (s *GRPCTapServer) TapByResource(req *pb.TapByResourceRequest, stream pb.Ta
...
+		errStrings := []string{}
+		errStr := fmt.Errorf("no pods to tap for %s/%s", res.GetType(), res.GetName())
+		errStrings = append(errStrings, errStr.Error())
+		if tapDisabled {
+			errStr = fmt.Errorf("pods found with tap disabled via the %s annotation", pkgK8s.ProxyDisableTapAnnotation)
+			errStrings = append(errStrings, errStr.Error())
 		}
-		return status.Errorf(codes.NotFound, "no pods found for %s/%s", resType, resName)
+		if tapNotEnabled {
+			errStr = fmt.Errorf("pods found with tap not enabled; try restarting resource so that it can be injected")
```
Pothulapati, line 110: Once this is updated, I can update the error message to instead ask for `viz check` where the restart injector message can be displayed full list of failed injected pods.

Pothulapati (approved):
LGTM.

dadjeibaah (approved):
LGTM! 📦 🚢

PR #5610, opened 2021-01-26
tests: include viz extension in upgrade-* integration tests

Pothulapati:
Currently, The source version of viz is installed even for
upgrade integration tests. Instead, Just like the core
control-plane, the viz extension should be started with stable/edge
version and then upgraded to the source version.

This helps discover any issues between upgrades for viz extension.

Currently, This is useful only for `upgrade-edge` CLI integration tests
as the viz extension is not present in stable yet and there is no
`helm-upgrade-edge` test.


kleimkuhler:
In `linkerd-viz` healthcheck, it looks like the order of `linkerd-viz pods are injected` and `viz extension pods are running` are flipped.

Pothulapati:
@kleimkuhler I am guessing it was failing because we are doing a viz check on the edge and comparing it with `viz.golden` in source which has the order change and hence it was failing.

Now, That we have a new edge with the ordering change, My guess is that this should work.

kleimkuhler (approved):

adleong (approved):

Pothulapati:
DO NOT MERGE this yet, as there is one last thing to fix in this.

bin/_test-helpers.sh
```diff
@@ -381,6 +381,18 @@ upgrade_test() {
   fi
 
   install_version "$install_url" "$upgrade_version"
+
+  # Install Viz if the version is edge
+  # TODO: Enable this also for stable once 2.10 is out
+  if [ "$release_channel" == "edge" ]; then
```
joakimr-axis, line 387: Better use the POSIX compatible `=` instead of `==` (they are equivalent in bash).
Also, no need for double quotes for static strings, use single quotes or no quotes.

`if [ "$release_channel" = 'edge' ]; then`

adleong:
@Pothulapati this has been open for a while, what's the status of it?

alpeb (commented):
@Pothulapati can you update this now that 2.10 is out? (also installing viz with Helm :wink:  )

Pothulapati:
@alpeb Updated

Also, We do install viz with Helm integration tests [already](https://github.com/linkerd/linkerd2/blob/tarun/integration-viz/test/integration/install_test.go#L528) right?

PR #5614, opened 2021-01-26
Hide the "Gateway" sidebar link

### LINKED ISSUE #5330: web dashboard should hide gateways page when linkerd-multicluster is not installed
The linkerd-viz dashboard should display the "gateways" page only when the `linkerd-multicluster` extensions is installed.  This can be detected by looking for the `linkerd.io/extension: multicluster` namespace label.

If linkerd-multicluster is not installed, the "gateways" page should be either hidden entirely, or display a helpful message with a call to action for installed the linkerd-multicluster extension.

dadjeibaah:
This commit hides the "Gateway" sidebar link in the dashboard if the
`linkerd-extension` is not installed. If a user happens to navigate to
the Gateway page anyway, we display a CTA (Call to Action) that informs
the user that they would need to run the multicluster install command.

This change includes a new endpoint in the dashboard server; `GET
/api/extensions`. This endpoint returns a list of all currently
installed `linkerd.io/extension` and the corresponding namespace each
extension is installed in. The dashboard uses this endpoint to detect
whether it needs to hide the navigation link and whether to display the
CTA.
Screen Shot 2021-01-26 at 4 15 36 PM
Screen Shot 2021-01-26 at 4 15 53 PM

Fixes #5330

### DIFF
--- a/web/app/js/components/Gateway.jsx
+++ b/web/app/js/components/Gateway.jsx
@@ -8,7 +8,6 @@ import React from 'react';
 import Spinner from './util/Spinner.jsx';
 import { Trans } from '@lingui/macro';
 import Typography from '@material-ui/core/Typography';
-import _find from 'lodash/find';
 import _isEmpty from 'lodash/isEmpty';
 import { processGatewayResults } from './util/MetricUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
@@ -87,12 +86,11 @@ class Gateways extends React.Component {
   }
 
   checkMulticlusterExtension() {
-    this.api.setCurrentRequests([this.api.fetchExtensions()]);
+    this.api.setCurrentRequests([this.api.fetchExtension(multiclusterExtensionName)]);
 
     this.serverPromise = Promise.all(this.api.getCurrentPromises())
-      .then(([extList]) => {
-        const mcExists = _find(extList, e => e.extensionName === multiclusterExtensionName);
-        this.setState({ multiclusterExists: mcExists || false });
+      .then(([extension]) => {
+        this.setState({ multiclusterExists: !_isEmpty(extension) });
       })
       .catch(this.handleApiError);
   }
--- a/web/app/js/components/Gateway.jsx
+++ b/web/app/js/components/Gateway.jsx
@@ -1,13 +1,19 @@
 import { handlePageVisibility, withPageVisibility } from './util/PageVisibility.jsx';
+import Card from '@material-ui/core/Card';
+import CardContent from '@material-ui/core/CardContent';
 import ErrorBanner from './ErrorBanner.jsx';
 import MetricsTable from './MetricsTable.jsx';
 import PropTypes from 'prop-types';
 import React from 'react';
 import Spinner from './util/Spinner.jsx';
 import { Trans } from '@lingui/macro';
+import Typography from '@material-ui/core/Typography';
+import _find from 'lodash/find';
 import _isEmpty from 'lodash/isEmpty';
 import { processGatewayResults } from './util/MetricUtils.jsx';
 import { withContext } from './util/AppContext.jsx';
+
+const multiclusterExtensionName = 'linkerd-multicluster';
 
 class Gateways extends React.Component {
   constructor(props) {
@@ -18,6 +24,7 @@ class Gateways extends React.Component {
     this.state = {
       pollingInterval: 2000,
       metrics: {},
+      multiclusterExists: false,
       pendingRequests: false,
       loaded: false,
       error: null,
@@ -26,6 +33,7 @@ class Gateways extends React.Component {
 
   componentDidMount() {
     this.startServerPolling();
+    this.checkMulticlusterExtension();
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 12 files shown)


web/srv/api_handlers.go
```diff
@@ -441,6 +441,28 @@ func (h *handler) handleAPIResourceDefinition(w http.ResponseWriter, req *http.R
...
+	resp := []map[string]interface{}{}
+	for _, ns := range nsList.Items {
+		resp = append(resp, map[string]interface{}{
+			"extensionName": ns.GetLabels()[linkerdExtensionLabel],
+			"namespace":     ns.Name,
+		})
+	}
+
+	renderJSON(w, resp)
+}
```
kleimkuhler, line 464: So in `pkg/k8s` we already have a `GetNamespaceWithExtensionLabel` which will get the namespace for a given label value. Since this check only exists for the multicluster extension, we can use that.

We'd still need the handling here for when that method returns an error, but I don't think we need to get all namespaces and loop through them when we already know the label value we are looking for.

web/app/js/components/Navigation.jsx
```diff
@@ -318,6 +322,17 @@ class NavigationBase extends React.Component {
...
+  checkMulticlusterExtension() {
+    this.api.setCurrentRequests([this.api.fetchExtensions()]);
+
+    this.serverPromise = Promise.all(this.api.getCurrentPromises())
+      .then(([extList]) => {
+        const mcExists = _find(extList, e => e.extensionName === multiclusterExtensionName);
+        this.setState({ showGatewayLink: mcExists || false });
+      })
+      .catch(this.handleApiError);
+  }
```
kleimkuhler, line 334: This can be simplified then we use the `GetNamespaceWithExtensionLabel` method as explained below.

web/srv/api_handlers.go
dadjeibaah: That's a good point -- I think it'd be still useful to get a list of all the extensions for the frontend in case we'd like to check if other extensions apart from multicluster exist. I could change this method so that the frontend can pass the extension name in the request. If the extension exists, we will return it in the response, if it doesn't we can send a not found or an empty object. WDYT @kleimkuhler ?

adleong (commented):
When I test this I get the following error:

```
time="2021-01-29T23:46:14Z" level=error msg="namespaces is forbidden: User \"system:serviceaccount:linkerd-viz:linkerd-web\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope"
```

I think the namespace list permission needs to be  granted to web.  Probably by adding it to the `linkerd-linkerd-viz-web-check` clusterrole.

dadjeibaah:
> When I test this I get the following error:
> 
> ```
> time="2021-01-29T23:46:14Z" level=error msg="namespaces is forbidden: User \"system:serviceaccount:linkerd-viz:linkerd-web\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope"
> ```
> 
> I think the namespace list permission needs to be granted to web. Probably by adding it to the `linkerd-linkerd-viz-web-check` clusterrole.

Weird, I'm surprised I didn't run into that. Did you do a `docker-build` to test? If that's the case, that's probably why I didn't run into this. Good find.

adleong:
I tested by running:

```
bin/install-pr --kind --cluster alex 5614
linkerd2-cli-git-49299b3a-darwin upgrade | kubectl apply --prune -l linkerd.io/control-plane-ns=linkerd  -f -
linkerd2-cli-git-49299b3a-darwin viz install | k apply -f -
linkerd2-cli-git-49299b3a-darwin mc install | k apply -f -
linkerd2-cli-git-49299b3a-darwin viz dashboard
```

web/srv/api_handlers.go
kleimkuhler: Yea passing the extension name is what I'd recommend so we can use `GetNamespaceWithExtensionLabel`. Since we don't need to get any other extensions, I think solving for this case is okay. If we come to a point where we need to get lists of extensions in the future, then maybe this can change then.

adleong (approved):
Nice! 🚪

kleimkuhler (approved):
I tested this with and without multicluster installed and it works well. Code changes looks good 👍

web/app/js/components/Gateway.jsx
```diff
@@ -99,7 +116,15 @@ class Gateways extends React.Component {
...
-            {noMetrics ? <div><Trans>noResourcesDetectedMsg</Trans></div> : null}
+            {noMetrics && multiclusterExists ? <div><Trans>noResourcesDetectedMsg</Trans></div> : null}
+            {(noMetrics && !multiclusterExists) &&
+            <Card>
+              <CardContent>
+                <Typography>To view gateway stats for your mesh, install the linkerd multicluster extension by running:</Typography>
+                <br />
+                <code>linkerd multicluster install | kubectl apply -f -</code>
+              </CardContent>
+            </Card>}
```
alpeb, line 127: Do you mind setting this up with translations? I believe most, if not all, text is currently using the `<Trans>` tags, so I think it'd be good practice to continue doing that whenever new text gets added, to avoid accumulating debt in that regard. `BUILD.md` has a new section explaining how to extract these strings to include them in the lingui json language assets.

dadjeibaah: Good catch! I'll get that changed.

alpeb (approved):
Thanks @dadjeibaah , looks good to me :+1:

PR #5619, opened 2021-01-27
Helpful error when attempting `linkerd upgrade` on Helm installation

### LINKED ISSUE #5518: More helpful error when attempting to `linkerd upgrade` a Helm installation
If Linkerd is installed with Helm, it is not possible to upgrade that installation using `linkerd upgrade`; only Helm can be used to upgrade.  However, attempting to use `linkerd upgrade` on a Helm installation of Linkerd gives a confusing error:

```
bin/linkerd upgrade
Error: failed to validate issuer credentials: failed to read CA: not PEM-encoded
```

I believe this is because it is attempting (and failing) to load the CA from the (non-existent) linkerd-config-overrides secret.  Instead, if `linkerd upgrade` is run on a Helm installation where linkerd-config-overrides does not exist should return an error message that prompts the user to upgrade with Helm instead.

piyushsingariya:
**Problem**
Not possible to upgrade Linkerd using cli if Linkerd was installed with Helm
when attempting `linkerd upgrade` with Helm installation it outputs a confusing error

**Solution**
When attempting `linkerd upgrade` CLI looks for `linkerd-stored-overrides` secret using `loadStoredValues()` https://github.com/linkerd/linkerd2/blob/main/cli/cmd/upgrade.go#L241
If it is not found Linkerd assume it to be a previous version before the introduction of `linkerd-stored-overrides` and tries to lookup for default values from the Legacy linked-config config map with `loadStoredValuesLegacy()` https://github.com/linkerd/linkerd2/blob/main/cli/cmd/upgrade.go#L249

If linkerd has been installed using Helm the `linkerd-stored-overrides` do not exist in the first place. And the error we're seeing now it being output by `validateValues()` https://github.com/linkerd/linkerd2/blob/main/cli/cmd/upgrade.go#L268 where the values are validated

> bin/linkerd upgrade
> Error: failed to validate issuer credentials: failed to read CA: not PEM-encoded

So after looking up values in the legacy if still, no values exist we return a new error notifying the user to upgrade Linkerd using **Helm**

Fixes #5518

### DIFF
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -251,9 +251,9 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 			return bytes.Buffer{}, err
 		}
 	}
-	// if there is still no linkerd-config-overrides secret, Linkerd has been installed using
-	// Helm installation, where linkerd-config-overrides secret doesn't exist
-	// User must be notified to upgrade Linkerd using Helm
+	// if we fails to fetch linkerd-config-overrides secret;
+	// Linkerd maybe installed using Helm installation,
+	// where linkerd-config-overrides secret doesn't exist
 	if values == nil {
 		err := errors.New("failed to fetch current install configuration; if Linkerd was installed with Helm, users need to upgrade with Helm.")
 		return bytes.Buffer{}, err
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -251,6 +251,13 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 			return bytes.Buffer{}, err
 		}
 	}
+	// if there is still no linkerd-config-overrides secret, Linkerd has been installed using
+	// Helm installation, where linkerd-config-overrides secret doesn't exist
+	// User must be notified to upgrade Linkerd using Helm
+	if values == nil {
+		err := errors.New("Linkerd was installed with Helm Installation, upgrade Linkerd with Helm")
+		return bytes.Buffer{}, err
+	}
 
 	err = flag.ApplySetFlags(values, flags)
 	if err != nil {
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -255,7 +255,7 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 	// Helm installation, where linkerd-config-overrides secret doesn't exist
 	// User must be notified to upgrade Linkerd using Helm
 	if values == nil {
-		err := errors.New("Linkerd was installed with Helm Installation, upgrade Linkerd with Helm")
+		err := errors.New("failed to fetch current install configuration; if Linkerd was installed with Helm, users need to upgrade with Helm.")
 		return bytes.Buffer{}, err
 	}
 
--- a/cli/cmd/upgrade.go
+++ b/cli/cmd/upgrade.go
@@ -255,7 +255,7 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 	// Linkerd maybe installed using Helm installation,
 	// where linkerd-config-overrides secret doesn't exist
 	if values == nil {
-		err := errors.New("failed to fetch current install configuration; if Linkerd was installed with Helm, users need to upgrade with Helm.")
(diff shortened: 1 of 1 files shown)


cli/cmd/upgrade.go
```diff
@@ -251,6 +251,13 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 			return bytes.Buffer{}, err
 		}
 	}
+	// if there is still no linkerd-config-overrides secret, Linkerd has been installed using
+	// Helm installation, where linkerd-config-overrides secret doesn't exist
+	// User must be notified to upgrade Linkerd using Helm
+	if values == nil {
+		err := errors.New("Linkerd was installed with Helm Installation, upgrade Linkerd with Helm")
```
kleimkuhler, line 258: I'm not sure we want to say with complete certainty that Linkerd was installed with Helm at this point; maybe the installation is just in a very odd state.

It may be worth changing the messaging on this to something like:

```suggestion
		err := errors.New("failed to fetch current install configuration; if Linkerd was installed with Helm, users need to upgrade with Helm.")
```

The comment should also be updated with similar messaging.

kleimkuhler (commented):
Thanks @piyushsingariya! This change looks good, but I have a comment about the messaging we'll use in the error.

cli/cmd/upgrade.go
piyushsingariya: :sweat_smile: I am not good with writing such errors, I am replacing it, with what you have suggested

cli/cmd/upgrade.go
```diff
@@ -251,6 +251,13 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 			return bytes.Buffer{}, err
 		}
 	}
+	// if there is still no linkerd-config-overrides secret, Linkerd has been installed using
+	// Helm installation, where linkerd-config-overrides secret doesn't exist
+	// User must be notified to upgrade Linkerd using Helm
```
kleimkuhler, line 256: If you can update this comment to reflect the new error message, that'd be great!

piyushsingariya: Oops! I forgot to replace that. Thanks for notifying me about this :smiley: 
I replaced the comment with

`// if we fail to fetch linkerd-config-overrides secret;
	// Linkerd maybe installed using Helm installation,
	// where linkerd-config-overrides secret doesn't exist`

kleimkuhler (approved):
Thanks @piyushsingariya!

adleong (commented):
Hi @piyushsingariya! In my testing, this does not fix the issue.  After installing with Helm and running upgrade on the branch I still get the same error message:

```
linkerd2-cli-git-1d8cbcc7-darwin upgrade
Error: failed to validate issuer credentials: failed to read CA: not PEM-encoded
```

cli/cmd/upgrade.go
```diff
@@ -251,6 +251,13 @@ func upgrade(ctx context.Context, k *k8s.KubernetesAPI, flags []flag.Flag, stage
 			return bytes.Buffer{}, err
```
adleong, line 251: I suspect that in the Helm case, `loadStoredValuesLegacy` probably returns an error because it tries to load the legacy Linkerd config which is empty.  This results in the PEM encoding error message because the CA cert field is empty and thus cannot be decoded.

One possible approach would be to modify `loadStoredValuesLegacy` to return `nil, nil` if the legacy Linkerd config is empty.  This lets us differentiate the following situations:

1. linkerd-config-overrides exists
  * recent Linkerd CLI was used to install
  * `loadStoredValues` returns non-nil
  * we can use those values as normal
2. linkerd-config-overrides does not exist but legacy config does
  * legacy Linkerd CLI was used to install
  * `loadStoredValuesLegacy` returns non-nil
  *  we can use those values as normal
 3. linkerd-config-overrides and legacy config both do not exist
   * Helm was used to install or linkerd-config-overrides has been deleted
   * We should return an error message
   * Note that once https://github.com/linkerd/linkerd2/pull/5636 merges, we can update this error message suggesting that if Helm was not used to install, the linkerd-config-overrides secret will need to be repaired.  (But we don't need to worry about that in this PR)

alpeb:
It appears this has already been fixed by #5636 which was already merged into `main`. 
@piyushsingariya  could you please check, and if so, close this PR and the linkerd issue?

adleong:
This has been fixed by https://github.com/linkerd/linkerd2/pull/5636

Issue #5545, opened 2021-01-15
Move tap and tap-injector code to viz directory


adleong:
The `tap` and `tap-injector` controllers are currently built as subcommands of the Linkerd controller and live inside the controller docker image.  We should move the code for these controllers into the `viz` directory and move them out of the controller docker image.

adleong:
It may be convenient to bundle both `tap` and `tap-injector` into the same docker image and perhaps even into the same pod but this isn't a requirement.

PR #5621, opened 2021-01-27
viz: add viz profile command

### LINKED ISSUE #5613: core: profile command's tap source flag does not work when Viz extension is not installed
## What is the issue?

When the Viz extension is not installed, the `profile --tap` command does not
work.

The `--tap` flag outputs a SP based on tap data for the given target source, but
if tap is not running or has not injected the target source, an unhelpful error
is returned.

## How can it be reproduced?

Install linkerd without the Viz extension and try to tap any target:

```
❯ linkerd profile t1-svc --tap deploy/t1 --tap-duration 25s --tap-route-limit 1
Error: HTTP error, status Code [404] (unexpected API response: 404 page not found
)
```

## Possible solutions

The `--tap` flag could check that the Viz extension is installed through the Viz
API's `CheckClientOrExit`, but this would bring a Viz dependency into the core
install.

Maybe introduce a `viz profile` command that only uses tap? There would not be a
`--open-api` or `--proto` flags for this command since they'd belong to the
`profile` command.

Move the `profile` command to the Viz extension? This feels a little unnecessary
though since it's only the `--tap` flag that is broken.

kleimkuhler:
## What this changes

This adds a `viz profile` command that outputs a service profile based off tap
data. It is identical—but fixes—the current `profile --tap` command.

Additionally, it removes the `--tap` flag from the `profile` command since this
depends on the Viz extension being installed in order to tap a service.

## Why

The `profile --tap` command is currently broken since it depends on the Viz
extension being installed, but the `profile` command is part of the core
install.

Closes #5613

Unblocks #5545


dadjeibaah (commented):
One super minor comment but otherwise this looks and works great!

cli/cmd/profile.go
```diff
@@ -112,7 +95,6 @@ func newCmdProfile() *cobra.Command {
 			}
 			// performs an online profile generation and access-check to k8s cluster to extract
 			// clusterDomain from linkerd configuration
-			// profile generation based on tap data requires access to k8s cluster
 			if !options.ignoreCluster {
 				var err error
 				k8sAPI, err = k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
```
Pothulapati, line 100: small nit, TIOLI: `k8sAPI` no longer needed for tap generation. we can make it a local variable right?

viz/cmd/profile.go
```diff
@@ -0,0 +1,213 @@
...
+		Example: `  # Generate a profile by watching live traffic.
+  linkerd profile -n emojivoto web-svc --tap deploy/web --tap-duration 10s --tap-route-limit 5
+`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if options.namespace == "" {
+				options.namespace = pkgcmd.GetDefaultNamespace(kubeconfigPath, kubeContext)
+			}
+			options.name = args[0]
+			clusterDomain := "cluster.local"
```
Pothulapati, line 81: Better to have this is a const in `viz/cmd/root.go` so that this can be re-used? just like the core cmd.

Pothulapati (approved):
LGTM. Also, would it be useful to mention this command in `linkerd profile` help?

cli/cmd/profile.go
kleimkuhler: Yep I can remove the `var k8sAPI *k8s.KubernetesAPI` above.

viz/cmd/profile.go
kleimkuhler: Yep this is added. Replace a use of it in `metrics-api/cmd` as well.

kleimkuhler: I backed these commits out. There are docker build errors when trying to pipe this through to metrics-api and then we end up only using it in `profile.go`. Because it's only used there, it doesn't seem worth moving it to `root.go` since it's not duplicated anywhere else.

dadjeibaah (approved):
Thanks for making that change! LGTM 📦 🚢

adleong (commented):
other than the comment above when `linkerd-viz` is not installed, this works great!

pkg/profiles/openapi.go
```diff
@@ -109,7 +106,8 @@ func swaggerToServiceProfile(swagger spec.Swagger, namespace, name, clusterDomai
 	return profile
 }
 
-func mkRouteSpec(path, pathRegex string, method string, operation *spec.Operation) *sp.RouteSpec {
+// MkRouteSpec makes a service profile route from an OpenAPI operation.
+func MkRouteSpec(path, pathRegex string, method string, operation *spec.Operation) *sp.RouteSpec {
```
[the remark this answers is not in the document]
kleimkuhler: I'll probably leave this in here. It's used by `swaggerToServiceProfile` in this file quite a few times, so I think making it public in `pkg` package is okay.

viz/cmd/profile.go
```diff
@@ -0,0 +1,213 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "profile [flags] --tap resource (SERVICE)",
+		Short: "Output service profile config for Kubernetes based off tap data",
+		Long:  "Output service profile config for Kubernetes based off tap data.",
+		Example: `  # Generate a profile by watching live traffic.
+  linkerd viz profile -n emojivoto web-svc --tap deploy/web --tap-duration 10s --tap-route-limit 5
+`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
kleimkuhler: Ah good catch!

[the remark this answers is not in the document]
kleimkuhler: Updated:

```
❯ bin/linkerd viz profile -n linkerd dest --tap deploy/linkerd-destination
Cannot connect to Linkerd Viz: could not find the linkerd-viz extension
Validate the install with: linkerd viz check
```

pkg/profiles/openapi.go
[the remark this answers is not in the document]
adleong: ah, my thinking was this isn't JUST used by the openapi code, it makes sense to go in `pkg/profiles/profiles.go`.  It's a bit weird for non-openapi code to depend on a function defined in `openapi.go` (not that REALLY matters since it's in the `pkg/profiles` package, it just feels weird)

adleong (approved):

PR #5627, opened 2021-01-28
viz: make checks aware of prom and grafana being optional

### LINKED ISSUE #5618: Issue with viz routes subcommand when using existing Prometheus
### What is the issue?

It appears that several of the viz health checks are marked with both `warning` and `fatal` with respect to cluster roles and role bindings.  When Prometheus is disabled on `linkerd-viz` Helm chart install, as is the case when reusing an existing instance, the checks fail with a warning but terminate the execution of the health check chain.   In at least the case of the `linkerd viz routes` command, the health checks are utilized to create the API client.  When Prometheus is disabled, the client health check is never reached but the chain terminates on a warning and so the `CheckClientOrExit` call does not exit.  Instead it returns a `nil` client that causes a panic on use.

### How can it be reproduced?

Install `linkerd-viz` with and external Prometheus and run `linkerd viz routes`.

### Logs, error output, etc

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x237c7e9]

goroutine 1 [running]:
github.com/linkerd/linkerd2/viz/cmd.requestRouteStatsFromAPI(0x0, 0x0, 0xc0001d0370, 0xc000996880, 0x0, 0x0, 0x0, 0x0)
	/linkerd-build/viz/cmd/routes.go:109 +0x29
github.com/linkerd/linkerd2/viz/cmd.NewCmdRoutes.func1(0xc000967340, 0xc00099e7b0, 0x1, 0x3, 0x0, 0x0)
	/linkerd-build/viz/cmd/routes.go:76 +0x2da
github.com/spf13/cobra.(*Command).execute(0xc000967340, 0xc00099e780, 0x3, 0x3, 0xc000967340, 0xc00099e780)
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0x38f01a0, 0x0, 0x24ab340, 0xc000116058)
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
	/linkerd-build/cli/main.go:10 +0x2d
```

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid
    issuer certificate will expire on 2021-01-28T17:35:07Z

‼ proxy-injector cert is valid for at least 60 days
    certificate will expire on 2021-01-28T20:35:00Z
    see https://linkerd.io/checks/#l5d-webhook-cert-not-expiring-soon for hints
‼ sp-validator cert is valid for at least 60 days
    certificate will expire on 2021-01-28T20:35:00Z
    see https://linkerd.io/checks/#l5d-webhook-cert-not-expiring-soon for hints

linkerd-api
-----------

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

linkerd-ha-checks
-----------------

Status check results are √
```

### Environment

- Kubernetes Version: 1.17.14
- Cluster Environment: GKE
- Host OS: MacOS
- Linkerd version: edge-21.1.3

### Possible solution

Removing the fatal markings on the role health checks that are also marked as warnings, and removing `linkerd-prometheus` from the `CheckForPods` slice allowed the subcommand to function.

Pothulapati:
Fixes #5618

Currently, The linkerd-viz checks fail whenever external
prometheus is being used as those checks are not aware of
prometheus and grafana being optional.

This commit fixes this by making the Prometheus and Grafana
as separate checks which are not fatal.

These checks will be made dynamic in a follow-up PR.

This commit also adds some of the missing resources checks,
especially of that of the new `metrics-api` into viz checks

## Output

```bash
❯ ./bin/go-run cli viz check
kubernetes-api
--------------
√ can initialize the client

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-viz
-----------
√ tap API service is running
√ prometheus is installed and configured correctly
√ grafana is installed and configured correctly
√ can initialize the client
√ viz extension self-check

Status check results are √
```


kleimkuhler (commented):
This kind of works for me. I do see the warning:

```
‼ prometheus is installed and configured correctly
    missing ClusterRoles: linkerd-linkerd-viz-prometheus
    see https://linkerd.io/checks/#l5d-viz-prometheus for hints
```

However, if I disabled Prometheus with `linkerd viz install --set prometheus.enabled=false` this check will hang indefinitely because of `SelfCheck` at the end:

```
❯ bin/linkerd viz check
kubernetes-api
--------------
...

linkerd-existence
-----------------
...

linkerd-viz
-----------
√ viz extension self-check
- waiting for check to complete
```

This may already be an issue. It looks like if we disable Prometheus, even if we also include `---set prometheusUrl=""`, it still uses the [override value](https://github.com/linkerd/linkerd2/commit/8ac536004140b11d005878aa3369b1c2abe6ad63). So even though Prometheus is disabled and we specified an empty Prometheus URL, it still looks for the default one and hangs.

Maybe we can fix as part of this?

Pothulapati:
@kleimkuhler Yep, Good Find. Will fix it as part of this PR.

Pothulapati:
@kleimkuhler Added the resources of `tap-injector`. I had to explicitly skip those resources for web because they can be disabled when users want to restrict privileges using a flag and hence did not want to take a dependency. WDYT?

Pothulapati:
Updated this PR to also prevent passing of the `prometheus-url` when prometheus is disabled. The metrics-api is also resilient when no prometheus instance is present and replies accordingly for cmds like stat, etc that require a prom isntance

```
tarun in dev in on  kind-kind () linkerd2 on  tarun/optional-prom-checks [$?] via 🐹 v1.15.4
❯ ./bin/go-run cli viz edges po -n linkerd-viz
Error: Edges API response error: No prometheus instance to connect
Usage:
  linkerd viz edges [flags] (RESOURCETYPE)

Examples:
  # Get all edges between pods that either originate from or terminate in the test namespace.
  linkerd viz edges po -n test

  # Get all edges between pods that either originate from or terminate in the default namespace.
  linkerd viz edges po

  # Get all edges between pods in all namespaces.
  linkerd viz edges po --all-namespaces

Flags:
  -A, --all-namespaces     If present, returns edges across all namespaces, ignoring the "--namespace" flag
  -h, --help               help for edges
  -n, --namespace string   Namespace of the specified resource
  -o, --output string      Output format; one of: "table" or "json" or "wide" (default "table")

Global Flags:
      --api-addr string            Override kubeconfig and communicate directly with the control plane at host:port (mostly for testing)
      --as string                  Username to impersonate for Kubernetes operations
      --as-group stringArray       Group to impersonate for Kubernetes operations
      --cni-namespace string       Namespace in which the Linkerd CNI plugin is installed (default "linkerd-cni")
      --context string             Name of the kubeconfig context to use
      --kubeconfig string          Path to the kubeconfig file to use for CLI requests
  -L, --linkerd-namespace string   Namespace in which Linkerd is installed (default "linkerd")
      --verbose                    Turn on debug logging

```

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -165,14 +156,78 @@ func (hc *HealthChecker) vizCategory() *healthcheck.Category {
...
 
 				return healthcheck.CheckPodsRunning(pods, "")
 			}))
 
+	checkers = append(checkers,
+		*healthcheck.NewChecker("prometheus is installed and configured correctly").
+			WithHintAnchor("l5d-viz-prometheus").
+			Warning().
+			WithCheck(func(ctx context.Context) error {
+				// TODO: Skip if prometheus is disabled
```
kleimkuhler, line 172: Is there more to do for this than I'm thinking? We can just check the values config and Prometheus is not enabled then it's skipped?

kleimkuhler (approved):
Thanks for fixing the prometheus issue!

This looks good and tests as expected for me 🤙.

I left a comment about adding the TODO comments for Prometheus and Grafana being disabled. Since these are warnings now, it's not a blocker for me, but maybe make a follow-up issue so they aren't just kind of "lost in then comments".

viz/pkg/healthcheck/healthcheck.go
Pothulapati: Its harder I think as there is no values config in `linkerd-viz` 🤔

kleimkuhler: Hm yea that does make it difficult.

kleimkuhler: After some off GitHub discussion this will be covered in a follow-up early next week 👍

viz/charts/linkerd-viz/templates/metrics-api.yaml
```diff
@@ -72,8 +72,8 @@ spec:
         - -cluster-domain={{.Values.clusterDomain}}
         {{- if .Values.prometheusUrl }}
         - -prometheus-url={{.Values.prometheusUrl}}
-        {{- else }}
-        - -prometheus-url=http://linkerd-prometheus.linkerd-viz.svc.{{.Values.clusterDomain}}:9090
+        {{- else if .Values.prometheus.enabled }}
```
adleong, line 75: What's the purpose of this change? It looks like the intent here is to skip rendering the `-prometheus-url` flag if both `.Values.prometheusUrl` is unset and `.Values.prometheus.enabled` is false.  But I think we should treat that as an invalid configuration.  Perhaps we just need an else block here that aborts with an error saying that prometheus.enabled or prometheusUrl must be set.

adleong (commented):
It looks like this change just adds the prometheus and grafana checks, while the change to handle the case where these components are disabled will be coming in a follow up?  Is that correct?  Would you mind updating the title and description of this PR to reflect what's changing here?

viz/charts/linkerd-viz/templates/metrics-api.yaml
Pothulapati: This was important because when external prometheus is not set, and default is disabled we would not want the `linkerd viz check` to just run for sometime and fail. This makes it error out that there is no prometheus instance to connect.

> Perhaps we just need an else block here that aborts with an error saying that prometheus.enabled or prometheusUrl must be set.

Yep, That is the plan. I was hoping to address these Prometheus related changes in the follow-up PR.

Pothulapati:
Raised https://github.com/linkerd/linkerd2/issues/5652

kleimkuhler:
Left a comment on there, but I think we had some discussion in standup after my approval above that we may just remove the `prometheus.enabled` configuration? Basically if installing the Viz extension, you're going to get Prometheus with it since most of the commands work off its metrics anyways.

adleong:
@kleimkuhler I don't think that's right.  `prometheus.enabled` controls if a Prometheus will be installed in the linkerd-viz extension's namespace.  `prometheus.enabled` can be false if `prometheusUrl` points to a "bring-your-own" prometheus.

viz/charts/linkerd-viz/templates/metrics-api.yaml
adleong: Cool.  Let's either do the proper thing here, or revert this change and handle it in the follow-up.

kleimkuhler:
@adleong Yep current behavior makes sense and was reason for my approval above! I'm just remembering when we talked about this at standup yesterday, simplifying these configurations came up and my takeaway was removing this field and only leaving `prometheusUrl`. Maybe if not setting `prometheusUrl`, then it defaults the Linkerd Prometheus (current behavior), and if set then it should point to the BYOP?

Pothulapati:
> Maybe if not setting prometheusUrl, then it defaults the Linkerd Prometheus (current behavior), and if set then it should point to the BYOP?

IMO, That could be confusing. I think the current way of doing this is better as it is more explicit this way.

viz/charts/linkerd-viz/templates/metrics-api.yaml
Pothulapati: Addressed the else case here itself

kleimkuhler:
After thinking about this a little more I realize this could be confusing as well. I think the behavior @adleong is describing above makes sense and is definitely more intuitive.

kleimkuhler (commented):
My comment below about removing the check should be removed. When testing locally with that removed, I get this `viz check` output:

```
$ linkerd viz check
...
linkerd-viz
-----------
√ tap API service is running
‼ prometheus is installed and configured correctly
    missing ClusterRoles: linkerd-linkerd-viz-prometheus
    see https://linkerd.io/checks/#l5d-viz-prometheus for hints
√ grafana is installed and configured correctly
√ can initialize the client
√ viz extension self-check
```

This warning should be fixed by #5652 though.

viz/pkg/healthcheck/healthcheck.go
```diff
@@ -82,14 +82,14 @@ func (hc *HealthChecker) VizCategory() healthcheck.Category {
...
 			}),
 		*healthcheck.NewChecker("linkerd-viz ClusterRoleBindings exist").
 			WithHintAnchor("l5d-viz-crb-exists").
 			Fatal().
 			Warning().
 			WithCheck(func(ctx context.Context) error {
-				return healthcheck.CheckClusterRoleBindings(ctx, hc.KubeAPIClient(), true, []string{fmt.Sprintf("linkerd-%s-prometheus", hc.vizNamespace), fmt.Sprintf("linkerd-%s-tap", hc.vizNamespace)}, "")
+				return healthcheck.CheckClusterRoleBindings(ctx, hc.KubeAPIClient(), true, []string{fmt.Sprintf("linkerd-%s-tap", hc.vizNamespace), fmt.Sprintf("linkerd-%s-metrics-api", hc.vizNamespace), fmt.Sprintf("linkerd-%s-tap-auth-delegator", hc.vizNamespace), "linkerd-tap-injector"}, "")
 			}),
 		*healthcheck.NewChecker("linkerd-viz ConfigMaps exist").
```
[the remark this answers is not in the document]
Pothulapati: Right, Removed this.

Pothulapati:
> This warning should be fixed by #5652 though.

Yep, you are right. That is part of 5652

kleimkuhler (approved):
🙌

adleong (approved):

PR #5628, opened 2021-01-28
Remove extra three dashes from helm templates

tkms0106:
(Background information)
In our company we are checking the sops-encrypted Linkerd manifest into GitHub repository,
and I came across the following problem.

---

Three dashes mean the start of the YAML document (or the end of the
directive).
https://yaml.org/spec/1.2/spec.html#id2800132

If there are only comments between `---`, the document is empty.
Assume the file which include an empty document at the top of itself.

```yaml
---
# foo
---
apiVersion: v1
kind: Namespace
metadata:
  name: foo
---
# bar
---
apiVersion: v1
kind: Namespace
metadata:
  name: bar
```

When we encrypt and decrypt it with [sops](https://github.com/mozilla/sops), the empty document will be
converted to `{}`.

```yaml
{}
---
apiVersion: v1
kind: Namespace
metadata:
    name: foo
---
apiVersion: v1
kind: Namespace
metadata:
    name: bar
```

It is invalid as k8s manifest ([apiVersion not set, kind not set]).

```
error validating data: [apiVersion not set, kind not set]
```

---

I'm afraid that it's sops's problem (at least partly), but anyhow this modification is enough harmless I think.
Thank you.

### DIFF
--- a/charts/linkerd2/templates/controller-rbac.yaml
+++ b/charts/linkerd2/templates/controller-rbac.yaml
@@ -2,7 +2,6 @@
 ###
 ### Controller RBAC
 ###
----
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -2,7 +2,6 @@
 ###
 ### Controller
 ###
----
 kind: Service
 apiVersion: v1
 metadata:
--- a/charts/linkerd2/templates/destination-rbac.yaml
+++ b/charts/linkerd2/templates/destination-rbac.yaml
@@ -2,7 +2,6 @@
 ###
 ### Destination Controller Service
 ###
----
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
--- a/charts/linkerd2/templates/destination.yaml
+++ b/charts/linkerd2/templates/destination.yaml
@@ -2,7 +2,6 @@
 ###
 ### Destination Controller Service
 ###
----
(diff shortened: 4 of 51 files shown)


kleimkuhler (commented):
Hey @tkms0106 thanks for opening this change. It looks like you'll need to run `go test ./... --update` to update the golden files for unit tests to pass.

Additionally, this is kind of a nit, but it looks like in the YAML link you provided comments typically go above the `---`. Maybe we can delete the first one instead of the second? As an example, I'm looking at `example 2.7` in that document.

Finally, from a brief scan of the change we'd need to make similar changes in the Jaeger directory, and it's probably worth double checking the multicluster and viz directories as well—all of which have a corresponding Helm chart.

olix0r:
@kleimkuhler 

> Additionally, this is kind of a nit, but it looks like in the YAML link you provided comments typically go above the ---. Maybe we can delete the first one instead of the second? As an example, I'm looking at example 2.7 in that document.

Deeply nitty territory, but I think we'd want them to end up like:

```yaml
---
# bar
apiVersion: v1
kind: Namespace
metadata:
    name: bar
```

So that the comment is "part of" the correct document. If the seperator follows the comment, it will technically be at the end of the prior object.

kleimkuhler:
@olix0r Yea that makes sense. I assumed there was a reason in the doc that it had the comments above the `---`, but keeping them below seems right.

kleimkuhler (approved):
Awesome thanks for including the extensions @tkms0106!

adleong (approved):
Thanks!

kleimkuhler:
@tkms0106 Thanks for resolving the merge conflicts that happened on this!

PR #5642, opened 2021-02-01
cli: make `linkerd uninstall` fail when injected pods are present

### LINKED ISSUE #5622: `linkerd uninstall` should fail when injected resources are present
Running `linkerd uninstall` on a cluster with injected resources can leave the cluster in a bad state (i.e. because pods will be unable to discover the control plane). `linkerd uninstall` should error out with a warning when there are injected resources on the cluster (i.e. including extensions).

We should provide a `--force` flag to ignore the cluster state and run removal anyway.

Pothulapati:
Fixes #5622

This PR updates the `linkerd uninstall` cmd to check if there
are any injected pods and fails if there are any. This also
provides `--force` flag to skip this check.

pods from the control-plane namespace are skipped
so as to not error out for control-plane pods.

This PR also fixes `viz uninstall` and updates the
uninstall integration test to use it.

### DIFF
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
-	"github.com/linkerd/linkerd2/pkg/healthcheck"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/k8s/resource"
 	"github.com/spf13/cobra"
@@ -34,15 +34,22 @@ This command provides all Kubernetes namespace-scoped and cluster-scoped resourc
 			}
 
 			if !force {
-				podsList, err := k8sAPI.CoreV1().Pods("").List(cmd.Context(), metav1.ListOptions{})
+				podList, err := k8sAPI.CoreV1().Pods("").List(cmd.Context(), metav1.ListOptions{LabelSelector: k8s.ControllerNSLabel})
 				if err != nil {
 					return err
 				}
-				for _, pod := range podsList.Items {
-					// Skip core control-plane namespace
-					if pod.Namespace != controlPlaneNamespace && healthcheck.ContainsProxy(pod) {
-						return fmt.Errorf("Please uninject proxy containers before uninstalling the control-plane")
+
+				var injectedPods []string
+				for _, pod := range podList.Items {
+					// skip core control-plane namespace
+					if pod.Namespace != controlPlaneNamespace {
+						injectedPods = append(injectedPods, fmt.Sprintf("* %s", pod.Name))
 					}
+				}
+
+				if len(injectedPods) > 0 {
+					fmt.Fprintln(os.Stderr, fmt.Sprintf("Please uninject the following pods before uninstalling the control-plane:\n\t%s", strings.Join(injectedPods, "\n\t")))
+					os.Exit(1)
 				}
 			}
 
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/linkerd/linkerd2/pkg/healthcheck"
 	"github.com/linkerd/linkerd2/pkg/k8s"
@@ -40,8 +39,8 @@ This command provides all Kubernetes namespace-scoped and cluster-scoped resourc
 					return err
 				}
 				for _, pod := range podsList.Items {
-					// Skip core control-plane and extension namespaces when checking for proxies
-					if !strings.HasPrefix(pod.Namespace, "linkerd") && healthcheck.ContainsProxy(pod) {
+					// Skip core control-plane namespace
+					if pod.Namespace != controlPlaneNamespace && healthcheck.ContainsProxy(pod) {
 						return fmt.Errorf("Please uninject proxy containers before uninstalling the control-plane")
 					}
 				}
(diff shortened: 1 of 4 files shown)


cli/cmd/uninstall.go
```diff
@@ -25,18 +28,34 @@ func newCmdUninstall() *cobra.Command {
...
+			}
+
+			if !force {
+				podsList, err := k8sAPI.CoreV1().Pods("").List(cmd.Context(), metav1.ListOptions{})
+				if err != nil {
+					return err
+				}
+				for _, pod := range podsList.Items {
+					// Skip core control-plane and extension namespaces when checking for proxies
+					if !strings.HasPrefix(pod.Namespace, "linkerd") && healthcheck.ContainsProxy(pod) {
```
alpeb, line 44: I think you can tap here into the `linkerd-namespace` flag instead of relying on the hard-coded value. Also, according to the requirement in #5622, this should also check for injected pods in extensions, so a strict comparison against the linkerd namespace might be enough.

cli/cmd/uninstall.go
```diff
@@ -25,18 +28,34 @@ func newCmdUninstall() *cobra.Command {
...
+						return fmt.Errorf("Please uninject proxy containers before uninstalling the control-plane")
+					}
+				}
+			}
+
+			return uninstallRunE(cmd.Context(), k8sAPI)
 		},
 	}
 
+	cmd.Flags().BoolVarP(&force, "force", "f", force, "Enable this to perform a forced uninstall")
```
alpeb, line 54: Can you use a more explicit description, like "Force uninstall even if there exist non-control-plane injected pods"?

cli/cmd/uninstall.go
```diff
@@ -25,18 +27,34 @@ func newCmdUninstall() *cobra.Command {
...
+
+			if !force {
+				podsList, err := k8sAPI.CoreV1().Pods("").List(cmd.Context(), metav1.ListOptions{})
+				if err != nil {
+					return err
+				}
+				for _, pod := range podsList.Items {
+					// Skip core control-plane namespace
+					if pod.Namespace != controlPlaneNamespace && healthcheck.ContainsProxy(pod) {
+						return fmt.Errorf("Please uninject proxy containers before uninstalling the control-plane")
```
kleimkuhler, line 44: Instead of returning an error here, I think something like the below would be good:

```suggestion
						fmt.Fprintln(os.Stderr, "Please uninject proxy containers before uninstalling the control-plane")
						os.Exit(1)
```

That way, we are not printing the usage text since it's not really a "usage" issue here. We do this in a few other places of the CLI and probably should try to do it more. There is an open issue for it.

With this change:

```
❯ ./bin/go-run cli uninstall
Please uninject proxy containers before uninstalling the control-plane
```

kleimkuhler (commented):
This works well for me. I left a comment about handling the error case.

Something to consider is since we are looping through all pods already, could we be a little more specific about which pods (or even which namespaces) contain pods that still need to be uninjected?

This is kind of similar to the `check` change where we warn about specific pods that may not have tap or tracing configured correctly. It may be helpful here to be able to say "these are the pods that need to uninjected" instead of sending them on a search in their cluster.

cli/cmd/uninstall.go
```diff
@@ -25,18 +27,34 @@ func newCmdUninstall() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return uninstallRunE(cmd.Context())
+
+			k8sAPI, err := k8s.NewAPI(kubeconfigPath, kubeContext, impersonate, impersonateGroup, 0)
+			if err != nil {
+				return err
+			}
+
+			if !force {
+				podsList, err := k8sAPI.CoreV1().Pods("").List(cmd.Context(), metav1.ListOptions{})
```
adleong, line 37: You can use a label selector to find data plane pods specifically: 
```
podList, err := api.CoreV1().Pods(dataPlaneNamespace).List(ctx, metav1.ListOptions{LabelSelector: k8s.ControllerNSLabel})
```
similar to what is done in `healthcheck.GetMeshedPodsIdentityData`

cli/cmd/uninstall.go
adleong, line 44: We could collect the offending pods into a slice so that we can print the list of pods which need to be uninjected.

Pothulapati: Makes sense.

cli/cmd/uninstall.go
Pothulapati: This makes sense. This way we don't have to do check if there is a proxy as having this label confirms that injection happened.

cli/cmd/uninstall.go
```diff
@@ -34,16 +34,23 @@ This command provides all Kubernetes namespace-scoped and cluster-scoped resourc
...
-				for _, pod := range podsList.Items {
-					// Skip core control-plane namespace
-					if pod.Namespace != controlPlaneNamespace && healthcheck.ContainsProxy(pod) {
-						return fmt.Errorf("Please uninject proxy containers before uninstalling the control-plane")
+
+				var injectedPods []string
+				for _, pod := range podList.Items {
+					// skip core control-plane namespace
+					if pod.Namespace != controlPlaneNamespace {
+						injectedPods = append(injectedPods, fmt.Sprintf("* %s", pod.Name))
```
kleimkuhler, line 46: Could be nice including namespace here? Something like:

```
Please uninject the following pods before uninstalling the control-plane:
        * linkerd-tap-7df756764b-lqzkr
```

Turns into:

```
Please uninject the following pods before uninstalling the control-plane:
        * linkerd-viz/linkerd-tap-7df756764b-lqzkr
```

I realize we don't do this in `check --proxy` output either though so definitely not a blocker.

kleimkuhler (approved):
Awesome works well 🤙

adleong (approved):

cli/cmd/uninstall.go
Pothulapati: Yep, Good suggestion. Will raise a separate PR fixing this everywhere.

PR #5643, opened 2021-02-01
Add tracing enabled annotation

### LINKED ISSUE #5632: jaeger: Is a jaeger-enabled annotation useful?
Is it valuable to add a `jaeger-enabled` annotation for the Jaeger extension,
similar to the `tap-enabled` annotation for the Viz extension?

#5608 adds the `tap-enabled` annotation primarily so that when the tap server is
tapping a pod, it can look for this annotation and decide whether to tap or
return an error. A side effect of this is that `viz check` can look for this
annotation on pods and commmunicate to a user if the `tap-enabled` annotation is
not on a pod, then the pod cannot be tapped.

An equivalent `jaeger-enabled` annotation could prove useful for the `jaeger
check` command. If a pod does not have this annotation, then the Jaeger
extension's `jaeger-injector` has not injected the pod and we can communicate to
a user that tracing will not work for that workload.

Since right now this would only be used for `jaeger check`, it's not as
important as `tap-enabled` annotation, but it may be worth considering and I'm
interested in any further input.

kleimkuhler:
This change adds the `jaeger.linkerd.io/tracing-enabled` annotation which is
automatically added by the Jaeger extension's `jaeger-injector`.

All pods that receive this annotation have also had the required environment
variables and volume/volume mounts add by the injector.

The purpose of this annotation is that it will allow `jaeger check` to check for
the presence of this annotation instead of needing to look at the proxy
containers directly. If this annotation is not present on pods, `jaeger check`
can warn users that tracing is not configured for those pods. This is similar to
`viz check` warning users that tap is not configured—recenlty added in #5602.

Closes #5632

### DIFF
--- a/controller/tap-injector/patch.go
+++ b/controller/tap-injector/patch.go
@@ -3,7 +3,7 @@ package tapinjector
 const tpl = `[
   {
     "op": "add",
-    "path": "/metadata/annotations/{{.Annotation}}",
+    "path": "/metadata/annotations/viz.linkerd.io~1tap-enabled",
     "value": "true"
   },
   {
--- a/jaeger/injector/mutator/patch.go
+++ b/jaeger/injector/mutator/patch.go
@@ -3,7 +3,7 @@ package mutator
 const tpl = `[
   {
     "op": "add",
-    "path": "/metadata/annotations/{{.Annotation}}",
+    "path": "/metadata/annotations/jaeger.linkerd.io~1tracing-enabled",
     "value": "true"
   },
   {
--- a/jaeger/injector/mutator/patch.go
+++ b/jaeger/injector/mutator/patch.go
@@ -3,6 +3,11 @@ package mutator
 const tpl = `[
   {
     "op": "add",
+    "path": "/metadata/annotations/{{.Annotation}}",
+    "value": "true"
+  },
+  {
+    "op": "add",
     "path": "/spec/containers/{{.ProxyIndex}}/env/-",
     "value": {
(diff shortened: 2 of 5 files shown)


jaeger/injector/mutator/patch.go
```diff
@@ -1,6 +1,11 @@
 package mutator
 
 const tpl = `[
+  {
+    "op": "add",
+    "path": "/metadata/annotations/{{.Annotation}}",
```
adleong, line 6: is it worth adding a template variable and dynamic json pointer escaping logic just to reuse the constant?  I'm all for reusing constants, but it might be simpler here to just hardcode it as:

```
"path": "/metadata/annotations/jaeger.linkerd.io~1tracing-enabled",
```

kleimkuhler: Yea there are a few hoops to jump through here for something that is ultimately not going to change. I'll remove this from here as well as the tap-injector.

adleong (approved):
Nice!

controller/tap-injector/patch.go
```diff
@@ -3,7 +3,7 @@ package tapinjector
 const tpl = `[
   {
     "op": "add",
-    "path": "/metadata/annotations/{{.Annotation}}",
+    "path": "/metadata/annotations/viz.linkerd.io~1tap-enabled",
```
Pothulapati, line 6: Should this intead be `jaeger.linkerd.io~1tap-enabled` ?

Pothulapati (commented):
Works as expected, except for the one comment.

Pothulapati (approved):
LGTM. Works as expected.

PR #5647, opened 2021-02-01
Run extension checks when linkerd check is invoked

### LINKED ISSUE #5233: linkerd check should invoke extension checks
When any Linkerd extensions are installed, the `linkerd check` command should also run the checks for that extension.  For example, when the linkerd-jaeger extension is installed, `linkerd check` should also run the `linkerd jaeger check` checks.

dadjeibaah:
This change allows the linkerd check command to also run any known
linkerd extension commands that have been installed in the cluster. It
does this by first querying for any namespace that has the label
selector `linkerd.io/extension` and then runs the subcommands for either
`jaeger`, `multicluster` or `viz`. This change ignores any extension
that isn't part of the Linkerd extension suite.

Fixes #5233


adleong (commented):
Looks great, just have some suggestions on the formatting and way the output is rendered.

Here's what I see with linkerd-viz installed:

```
...
control-plane-version
---------------------
√ control plane is up-to-date
    control plane running edge-21.1.4 but cli running git-4109a954

Status check results are √
kubernetes-api
--------------
√ can initialize the client

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-viz
-----------
√ linkerd-viz ConfigMaps exist
√ tap API service is running
√ can initialize the client
√ viz extension self-check

Status check results are √
```

I think we should remove the `kubernetes-api` and `linkerd-existence` categories from the extension checks.  These are already checked by the main control plane so checking these in the extension checks is redundant and it makes the output of running the checks together like this redundant.

It looks a bit weird that we print `Status check results are √` from the main checks and then continue to print out more output.  I think this would make more sense if we used a banner or something to visually distinguish the extension checks.  Something like:

```
control-plane-version
---------------------
√ control plane is up-to-date
    control plane running edge-21.1.4 but cli running git-4109a954

Status check results are √

Linkerd extension checks
===================

linkerd-viz
-----------
...
Status check results are √
```

It would also be great to print out every extension, even if we don't have checks for that extension.  e.g.

```
Linkerd extension checks
===================

linkerd-viz
-----------
...
Status check results are √

linkerd-foobar
---------------
√ linkerd-foobar Namespace exists
Status check results are √
```

cli/cmd/check.go
```diff
@@ -211,6 +216,129 @@ func configureAndRunChecks(ctx context.Context, wout io.Writer, werr io.Writer,
...
+		return err
+	}
+
+	// no extensions to check
+	if len(namespaces) == 0 {
+		return nil
+	}
+
+	headerTxt := "Linkerd extension checks"
+	fmt.Fprintln(wout, "")
```
adleong, line 244: nit: this can just be `fmt.Fprintln(wout)`

adleong (commented):
I think we need to add some special handling to make this work in `--output json` mode:

```
    {
      "categoryName": "control-plane-version",
      "checks": [
        {
          "description": "control plane is up-to-date",
          "hint": "https://linkerd.io/checks/#l5d-version-control",
          "error": "unsupported version channel: git-6fea6c68",
          "result": "warning"
        },
        {
          "description": "control plane and cli versions match",
          "hint": "https://linkerd.io/checks/#l5d-version-control",
          "error": "control plane running git-6fea6c68 but cli running git-396368f9",
          "result": "warning"
        }
      ]
    }
  ]
}

Linkerd extension checks
========================

{
  "success": true,
  "categories": [
    {
      "categoryName": "linkerd-jaeger",
      "checks": [
        {
          "description": "linkerd-jaeger extension Namespace exists",
          "result": "success"
        },
```

One possibility would be to just skip printing the extension header in json mode, which would make this print a series of valid json objects (one for the control plane and one for each extension).  Even better would be to collect the output of each extension check and format the entire output as one valid json object or array.  This would make it much easier to process the output with json tooling like jq etc.

dadjeibaah:
> I think we need to add some special handling to make this work in `--output json` mode:
Oof. That's a good point. I agree that making the json output one json array sounds good but might be difficult since each extension check is invoked by generating and running their own cobra command. I think if we want to merge extension check json output to linkerd core check, we'd need to use each extension's healthchecker pkg to generate healthcheck output and not rely on the check commands for each extension.

I could do that in a follow up PR and we could stick with a series of valid json objects for now. But I definitely agree, it'd be a better UX to have this print one cohesive JSON object. WDYT @adleong?

adleong:
Tackling this in a follow up sounds good.  I'd like to avoid depending on the extension healthchecks and instead treat the extension checks as a black box, only interacting with their input and output, since we'll be limited to this for external extension checks.

What we could do is refactor the `NewCmdCheck` functions to take an `io.Writer` so we can pass a StringWriter into the extension checks to capture their output.  We can then construct our own json output which contains the extension json ouput as array elements or whatever.

adleong:
@dadjeibaah could you please file a follow up github issue about making the --output json mode return valid json?

adleong (approved):

dadjeibaah:
> What we could do is refactor the NewCmdCheck functions to take an io.Writer so we can pass a StringWriter into the extension checks to capture their output. 

Ah, that's a great idea! I'll file an issue to capture that.

pkg/healthcheck/healthcheck.go
```diff
@@ -442,6 +442,37 @@ func NewHealthChecker(categoryIDs []CategoryID, options *Options) *HealthChecker
...
+	k8sAPI, err := k8s.NewAPI(hc.KubeConfig, hc.KubeContext, hc.Impersonate, hc.ImpersonateGroup, RequestTimeout)
+	if err != nil {
+		return err
+	}
+	hc.kubeAPI = k8sAPI
+
+	return nil
+}
+
+// InitializeLinkerdGlobalConfig populates the linkerd config objecti in the
```
alpeb, line 458:
```suggestion
// InitializeLinkerdGlobalConfig populates the linkerd config object in the
```

cli/cmd/check.go
```diff
@@ -211,6 +216,130 @@ func configureAndRunChecks(ctx context.Context, wout io.Writer, werr io.Writer,
...
+		return err
+	}
+
+	// no extensions to check
+	if len(namespaces) == 0 {
+		return nil
+	}
+
+	if opts.output != healthcheck.JSONOutput {
+		headerTxt := "Linkerd extension checks"
```
alpeb, line 244:
```suggestion
		headerTxt := "Linkerd extensions checks"
```

test/integration/testdata/check.viz.proxy.golden
```diff
@@ -0,0 +1,22 @@
...
+√ can initialize the client
+√ viz extension self-check
+
+linkerd-viz-data-plane
+----------------------
+√ data plane namespace exists
+√ data plane proxy metrics are present in Prometheus
+√ data-plane pods have tap enabled
+
+Status check results are √
```
kleimkuhler, line 22: nit: Add newline here

kleimkuhler (commented):
This looks great! It tests well with the existing extensions as well as making a test one with the `linkerd.io/extension` label.

You should be able to remove the extra line added to the bottom of `check.config.golden`, `check.golden`, and `check.proxy.golden`.

alpeb (commented):
It appears there might be an issue when having the extension commands pick the global flags such as `--context`. Say if I install linkerd and viz in my local cluster then change the default k8s context and run `linkerd check --context originalcontext` then the core checks pass, but the viz ones fail:
```bash
$ bin/go-run cli check  --context originalcontext
...
Status check results are √                                                                                             │      --wait duration      Maximum allowed time for all tests to pass (default 5m0s)
                                                                                                                       │
Linkerd extension checks                                                                                               │Global Flags:
========================                                                                                               │      --api-addr string            Override kubeconfig and communicate directly with the control plane at host:port (mo
                                                                                                                       │stly for testing)
linkerd-viz                                                                                                            │      --as string                  Username to impersonate for Kubernetes operations
-----------                                                                                                            │      --as-group stringArray       Group to impersonate for Kubernetes operations
× linkerd-viz Namespace exists                                                                                         │      --cni-namespace string       Namespace in which the Linkerd CNI plugin is installed (default "linkerd-cni")
    could not find the linkerd-viz extension                                                                           │      --context string             Name of the kubeconfig context to use
    see https://linkerd.io/checks/#l5d-viz-ns-exists for hints                                                         │      --kubeconfig string          Path to the kubeconfig file to use for CLI requests
                                                                                                                       │  -L, --linkerd-namespace string   Namespace in which Linkerd is installed (default "linkerd")
Status check results are ×
```

alpeb (commented):
This tested pretty well :+1:

cli/cmd/check.go
```diff
@@ -211,9 +216,136 @@ func configureAndRunChecks(ctx context.Context, wout io.Writer, werr io.Writer,
 		os.Exit(1)
 	}
 
+	err = runExtensionChecks(cmd, wout, werr, options)
+	if err != nil {
+		os.Exit(1)
```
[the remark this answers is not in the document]
dadjeibaah: Ah gotcha! Missed that in the last review.

alpeb (approved):
I tested with linkerd and the viz extensions installed, plus another random namespace annotated as an extension, and then I scaled down prometheus to have the viz check fail.
`linkerd check` fails on the viz check as expected, but it doesn't continue with the other namespace check, the reason being the `os.Exit(1)` on the viz check code. So the error catching on the calling side is not effective after all. We could return the error in viz instead, but that'd cause showing the command usage. It looks like there's no easy way out of this, so I think it might be fine to leave it as it is, and address the short-circuiting later when we go down the shelling-out way...

kleimkuhler (approved):
Looks good and tests well!

PR #5651, opened 2021-02-02
Move tap from core into Viz extension

### LINKED ISSUE #5545: Move tap and tap-injector code to viz directory

kleimkuhler:
Closes #5545.

This change moves all tap and tap-injector code into the viz directory. 

The tap and tap-injector components now also use a new tap image—separating
these components from the controller image that they are currently part of. This
means the controller image has removed all its build dependencies related to
tap.

Finally, the tap Protobuf has been separated from the metrics-api and moved into
it's own `.proto` file and gen directory. This introduces a clear split between
metrics-api and tap Protobuf.

There is no change in behavior for the `viz tap` command.

#### Docker images

All the bin directory scripts should be updated to build and load the tap image.
All the CI workflows should be updated to build and push the tap image.

#### Controller and pkg directories

This is primarily deletions. Most of the deleted code in this directory is now
in the tap directory of the Viz extension.

#### viz/tap

This is the location that all the tap related code now lives in. New files are
mostly moved from the controller and pkg directories. Imports have all been
updated to point at the right locations and Protobuf.

The Protobuf here is taken from metrics-api and contains all tap-related
Protobuf.


bin/docker-build
```diff
@@ -22,4 +22,4 @@ fi
 "$bindir"/docker-build-grafana
 "$bindir"/docker-build-jaeger-webhook
 "$bindir"/docker-build-metrics-api
-
+"$bindir"/docker-build-tap
```
Pothulapati, line 25: nit: new line?

bin/protoc-go.sh
```diff
@@ -6,17 +6,18 @@ bindir=$( cd "${0%/*}" && pwd )
...
 "$bindir"/protoc -I proto --go_out=plugins=grpc,paths=source_relative:controller/gen proto/common/healthcheck.proto
 "$bindir"/protoc -I proto -I viz/metrics-api/proto --go_out=plugins=grpc,paths=source_relative:viz/metrics-api/gen viz/metrics-api/proto/viz.proto
+"$bindir"/protoc -I proto -I viz/tap/proto -I viz/metrics-api/proto --go_out=plugins=grpc,paths=source_relative:viz/tap/gen viz/tap/proto/tap.proto
 
 mv controller/gen/common/net.pb.go   controller/gen/common/net/
 mv controller/gen/public.pb.go controller/gen/public/
 mv controller/gen/common/healthcheck.pb.go   controller/gen/common/healthcheck/
 mv viz/metrics-api/gen/viz.pb.go viz/metrics-api/gen/viz/viz.pb.go
-
+mv viz/tap/gen/tap.pb.go viz/tap/gen/tap/tap.pb.go
```
Pothulapati, line 23: nit: New line

viz/tap/cmd/main.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/linkerd/linkerd2/viz/tap/injector"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("expected a subcommand")
+		os.Exit(1)
+	}
+	switch os.Args[1] {
+	case "controller":
```
Pothulapati, line 17: Could this naming be confusing? I understand that now tap is like an umbrella concept, in which injector and the API are part of. Should this be `tap-api` or `api` instead? 🤔 

My main concern is this does not perform usual characteristics like that of a controller right? i.e reconciling based on events.

viz/tap/pkg/util.go
```diff
@@ -0,0 +1,184 @@
+package pkg
```
Pothulapati, line 1: Should this be `util` instead? referencing this as `viz/tap/pkg` was really confusing as `pkg` is this higher level packages directory in `golang`

Pothulapati (commented):
Awesome Work. 👏 

Left some suggestions :}

web/srv/api_handlers.go
```diff
@@ -15,14 +15,14 @@ import (
...
 	publicPb "github.com/linkerd/linkerd2/controller/gen/public"
 	"github.com/linkerd/linkerd2/pkg/healthcheck"
 	"github.com/linkerd/linkerd2/pkg/k8s"
 	"github.com/linkerd/linkerd2/pkg/protohttp"
-	"github.com/linkerd/linkerd2/pkg/tap"
-	pb "github.com/linkerd/linkerd2/viz/metrics-api/gen/viz"
+	metricsPb "github.com/linkerd/linkerd2/viz/metrics-api/gen/viz"
 	vizUtil "github.com/linkerd/linkerd2/viz/metrics-api/util"
+	tapPb "github.com/linkerd/linkerd2/viz/tap/gen/tap"
+	"github.com/linkerd/linkerd2/viz/tap/pkg"
```
alpeb, line 25: Given there are so many pkg around, I think this deserves a more explicit alias like `tapPkg`. It will generate some naming stuttering below, but that can be fixed later.

alpeb (commented):
Looking good :+1:

bin/docker-build
kleimkuhler: As in add a newline? I think it's fine since there is no spacing between the other commands.

bin/protoc-go.sh
kleimkuhler: Again there is no spacing between the other commands so adding a newline feels a little unnecessary.

viz/tap/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY bin/install-deps bin/
+RUN go mod download
+ARG TARGETARCH
+RUN ./bin/install-deps $TARGETARCH
+
+## compile tap
+FROM go-deps as golang
+WORKDIR /linkerd-build
+COPY pkg pkg
+COPY controller controller
```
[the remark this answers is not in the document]
kleimkuhler: #5661

bin/docker-build
adleong: Previously this file ended in two newline characters.  After your change it now ends in one newline character.  I think that's fine.

viz/tap/pkg/util.go
kleimkuhler: I feel like `pkg` is okay. I see `util` naming kind of equivalent; they are both packages that have general helpers that are imported by other packages.

kleimkuhler: I split these into some better named files but still within `tap/pkg` package. Hopefully that helps in some way.

viz/tap/cmd/main.go
kleimkuhler: Yea good call. I renamed it to `api`. `tap/api` now contains all the Go files for the gRPC and API servers.

viz/tap/api/server.go
```diff
@@ -61,7 +60,7 @@ func NewAPIServer(
 	}
 
 	log := logrus.WithFields(logrus.Fields{
-		"component": "apiserver",
+		"component": "tap API server",
```
adleong: feels a bit wordy for a log context field.  How about just "tap"?

viz/tap/cmd/main.go
adleong: I've been known to say "tap controller" when talking about this.  But I agree with @Pothulapati that calling this "api" is probably more specific.

bin/image-load
```diff
@@ -79,7 +79,7 @@ fi
 "$bin" version
 
 rm -f load_fail
-for img in proxy controller web metrics-api grafana cli-bin debug cni-plugin jaeger-webhook; do
+for img in proxy controller web metrics-api grafana cli-bin debug cni-plugin jaeger-webhook tap; do
```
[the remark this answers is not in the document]
kleimkuhler: jaeger-webhook is in this list. Did you mean another image?

bin/docker-push
```diff
@@ -14,6 +14,6 @@ bindir=$( cd "${BASH_SOURCE[0]%/*}" && pwd )
 # shellcheck source=_docker.sh
 . "$bindir"/_docker.sh
 
-for img in cli-bin cni-plugin controller metrics-api debug grafana proxy web ; do
+for img in cli-bin cni-plugin controller metrics-api debug grafana proxy web tap; do
```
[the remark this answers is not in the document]
kleimkuhler: I can add it. I think we've gotten by since this script is used a lot less now after the previous CI changes.

In general, there are quite a few outdated scripts that probably don't need to be around anymore, but that's outside of this change.

bin/docker-build-tap
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
kleimkuhler: I pretty much used your exact wording in #5663 😄. Agree this would be helpful, but also separate from this change.

bin/image-load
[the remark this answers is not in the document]
adleong: Nevermind, I just didn't see it 😅

adleong (commented):
Tested and everything works as expected for me.  However, I do see a slightly concerning warning in the tap and tap-injector logs:

```
2021/02/04 00:20:17 WARNING: proto: file "tap.proto" is already registered
	previously from: "github.com/linkerd/linkerd2/viz/tap/gen/tap"
	currently from:  "github.com/linkerd/linkerd2-proxy-api/go/tap"
A future release will panic on registration conflicts. See:
https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
```

kleimkuhler:
@adleong Thanks good catch. Renamed to `viz_tap.proto`.

adleong (approved):

alpeb (approved):
LGTM !

PR #5665, opened 2021-02-04
Simplify SelfCheck API

### LINKED ISSUE #5575: Simplify SelfCheck API
Currently the SelfCheck API returns multiple responses that require special handling in the healthcheck library. It's the only gRPC call made by health checks so it still requires to be handled a little different than the other checks, but we can simplify things by:
1) Simplifying the the gRPC function and its protobuf so that `SelfCheckResponse` consists of just a status field and an error message field, instead of the array of results that it currently has.
2) In `healthcheck.go`, processing the check with a regular `Checker.check` callback and getting rid of `Checker.checkRPC`. The callback should inspect the result and eventually return an error.

As an additional simplification we can merge that protobuf into `viz.proto`

This can be addressed after #5560 merges.

alpeb:
Fixes #5575

Now that only viz makes use of the `SelfCheck` api, merged the `healthcheck.proto` into `viz.proto`.

Also removed the "checkRPC" functionality that was used for handling multiple API responses and was only used by `SelfCheck`, because the extra complexity was not granted. Revert to use the plain vanilla "check" by just concatenating error responses.

## Success Output

```bash
$ bin/linkerd viz check
...
linkerd-viz
-----------
...
√ viz extension self-check
```

## Failure Examples

Failure when viz fails to connect to the k8s api:
```bash
$ bin/linkerd viz check
...
linkerd-viz
-----------
...
× viz extension self-check
    Error calling the Kubernetes API: someerror

Status check results are ×
```

Failure when viz fails to connect to Prometheus:
```bash
$ bin/linkerd viz check
...
linkerd-viz
-----------
...
× viz extension self-check
    Error calling Prometheus from the control plane: someerror

Status check results are ×
```

Failure when viz fails to connect to both the k8s api and Prometheus:
```bash
$ bin/linkerd viz check
...
linkerd-viz
-----------
...
× viz extension self-check
    Error calling the Kubernetes API: someerror
    Error calling Prometheus from the control plane: someerror

Status check results are ×
```


viz/pkg/healthcheck/healthcheck.go
```diff
@@ -180,8 +180,25 @@ func (hc *HealthChecker) VizCategory() healthcheck.Category {
 			// "waiting for check to complete" while things converge. If after the timeout
 			// it still hasn't converged, we show the real error (a 503 usually).
 			WithRetryDeadline(hc.RetryDeadline).
-			WithCheckRPC(func(ctx context.Context) (*healthcheckPb.SelfCheckResponse, error) {
-				return hc.vizAPIClient.SelfCheck(ctx, &healthcheckPb.SelfCheckRequest{})
+			WithCheck(func(ctx context.Context) error {
+				results, err := hc.vizAPIClient.SelfCheck(ctx, &pb.SelfCheckRequest{})
```
dadjeibaah, line 184: tiny TIOL but I like to leave a little bit of whitespace around code block operations. I find that it helps in code readability:
```diff
diff --git a/viz/pkg/healthcheck/healthcheck.go b/viz/pkg/healthcheck/healthcheck.go
index e6ae2f44..fcf6b7e4 100644
--- a/viz/pkg/healthcheck/healthcheck.go
+++ b/viz/pkg/healthcheck/healthcheck.go
@@ -185,9 +185,11 @@ func (hc *HealthChecker) VizCategory() healthcheck.Category {
                                if err != nil {
                                        return err
                                }
+
                                if len(results.GetResults()) == 0 {
                                        return errors.New("No results returned")
                                }
+
                                errs := []string{}
                                for _, res := range results.GetResults() {
                                        if res.GetStatus() != pb.CheckStatus_OK {
@@ -197,6 +199,7 @@ func (hc *HealthChecker) VizCategory() healthcheck.Category {
                                if len(errs) == 0 {
                                        return nil
                                }
+
                                errsStr := strings.Join(errs, "\n    ")
                                return errors.New(errsStr)
                        }),
```

dadjeibaah (approved):
This LGTM. I left one tiny TIOLI. 📦 🚢

viz/pkg/healthcheck/healthcheck.go
alpeb: Will do :+1:

adleong (approved):
😻

PR #5666, opened 2021-02-04
bin/test-cleanup reordering

alpeb:
After #5642 `linkerd uninstall` will fail if there still are injected pods in the cluster. So uninstall should happen last.


adleong (approved):

Pothulapati (approved):
LGTM

PR #5668, opened 2021-02-04
Use full-length actions SHAs in CI

alpeb:
Github now requires that actions pined to a SHA use the full-length SHA,
otherwise CI throws a warning like:
```
actions/checkout@722adc6 looks like the shortened version of a commit
SHA. Referencing actions by the short SHA will be disabled soon. Please
see
https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions.
```


dadjeibaah (approved):
📦 🚢

adleong (approved):

Issue #5664, opened 2021-02-04
cli: Remove the --disable-tap flag from inject


kleimkuhler:
The `inject` command still supports the `--disable-tap` flag which has moved to
the Viz extension. This flag and all remaining mentions of the tap in the core
installation should be removed.

Tap is disabled with the `config.linkerd.io/disable-tap` annotation. This is a
good time to rename this to `viz.linkerd.io/disable-tap` as the naming better
reflects this belongs to the Viz extension.

If users wish to disable tap on workload after the Viz extension has been
installed, they should be annotating resources with the
`viz.linkerd.io/disable-tap` annotation. There will be no `--disable-flag`
equivalent.

mayankshah1607:
Hi! I'd love to work on this issue :)

PR #5671, opened 2021-02-05
CLI: Remove the `--disable-tap` flag from inject

mayankshah1607:
Fixes https://github.com/linkerd/linkerd2/issues/5664

- Remove `--disable-flag` from `inject`
-  Move `config.linkerd.io/disable-tap` to `viz.linkerd.io/disable-tap`

### DIFF
--- a/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
+++ b/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
@@ -12,7 +12,6 @@ spec:
     metadata:
       annotations:
         config.linkerd.io/admin-port: "9998"
-        config.linkerd.io/disable-tap: "true"
         config.linkerd.io/proxy-cpu-limit: "1"
         config.linkerd.io/proxy-cpu-request: "0.5"
         config.linkerd.io/proxy-memory-limit: 256Mi
@@ -23,6 +22,7 @@ spec:
         linkerd.io/created-by: linkerd/cli dev-undefined
         linkerd.io/identity-mode: default
         linkerd.io/proxy-version: override
+        viz.linkerd.io/disable-tap: "true"
       labels:
         app: web-svc
         linkerd.io/control-plane-ns: linkerd
--- a/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
+++ b/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
@@ -22,7 +22,6 @@ spec:
         linkerd.io/created-by: linkerd/cli dev-undefined
         linkerd.io/identity-mode: default
         linkerd.io/proxy-version: override
-        viz.linkerd.io/disable-tap: "true"
       labels:
         app: web-svc
         linkerd.io/control-plane-ns: linkerd
--- a/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml
+++ b/cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml
@@ -19,7 +19,6 @@ spec:
         config.linkerd.io/proxy-version: override
         config.linkerd.io/skip-inbound-ports: 7777,8888
         config.linkerd.io/skip-outbound-ports: "9999"
-        viz.linkerd.io/disable-tap: "true"
       labels:
         app: web-svc
     spec:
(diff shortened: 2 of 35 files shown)


mayankshah1607:
Should we also remove `--restrict-dashboard-privileges` from `install`?

alpeb:
> Should we also remove `--restrict-dashboard-privileges` from `install`?

I think that was already removed. It wasn't added back into `linkerd viz install`, but IMO that's fine given it sounds to me like a not frequently used flag. Nevertheless it was replaced by `dashboard.restrictPrivileges` in viz' values.yaml, which still is modifiable through `--set` in the CLI.

alpeb (commented):
Looking good :+1: 

By grepping for `disable.tap` it seems there are a few more places that are still making reference to the old annotation.

mayankshah1607:
Thanks @alpeb 
All instances of `config.linkerd.io/disable-tap` have been replaced

alpeb (approved):
Thanks, LGTM!
Please note you'll need to fix some conflicts with `main` before we're able to merge.

dadjeibaah (approved):
Changes look good modulo other reviewers comment. 📦 🚢

cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.golden.yml
```diff
@@ -23,6 +22,7 @@ spec:
         linkerd.io/created-by: linkerd/cli dev-undefined
         linkerd.io/identity-mode: default
         linkerd.io/proxy-version: override
+        viz.linkerd.io/disable-tap: "true"
```
kleimkuhler, line 25: I'd like to see this removed since this depends on an extension. Since we are testing the core install, this annotation should have no significance in testing.

cli/cmd/testdata/inject_emojivoto_deployment_config_overrides.input.yml
```diff
@@ -12,14 +12,14 @@ spec:
...
         config.linkerd.io/admin-port: "9998"
-        config.linkerd.io/disable-tap: "true"
         config.linkerd.io/proxy-cpu-limit: "1"
         config.linkerd.io/proxy-cpu-request: "0.5"
         config.linkerd.io/proxy-memory-limit: 256Mi
         config.linkerd.io/proxy-memory-request: 64Mi
         config.linkerd.io/proxy-version: override
         config.linkerd.io/skip-inbound-ports: 7777,8888
         config.linkerd.io/skip-outbound-ports: "9999"
+        viz.linkerd.io/disable-tap: "true"
```
kleimkuhler, line 22: Same for this!

pkg/k8s/labels.go
```diff
@@ -126,6 +126,9 @@ const (
 	// ProxyConfigAnnotationsPrefix is the prefix of all config-related annotations
 	ProxyConfigAnnotationsPrefix = "config.linkerd.io"
 
+	// VizAnnotationPrefix is the prefix for all viz related annotations
+	VizAnnotationPrefix = "viz.linkerd.io"
```
kleimkuhler, line 130: This and `ProxyDisableTapAnnotation` are specific to the Viz extension, so I don't think they belong as part of the Linkerd's `pkg` package.

In #5651, the Viz package had it's own labels added so this can removed, and `ProxyDisableTapAnnotation` should move to that file.

kleimkuhler (changes requested):
Hey @mayankshah1607 there are a few things I left comments on, as well as some further items to address:

- `_validate.tpl` should no longer validate that control plane components have tap disabled. First, this is okay to do; in fact it already is the case for the `viz install` now. Second, the core install should not be validating configurations for an extension that may not even be installed.
- `doc.go` still has a description for `ProxyDisableTapAnnotation`:
```
{
	Name:        k8s.ProxyDisableTapAnnotation,
	Description: "Disables resources from being tapped",
},
```
- `inject.go` still has logic around tap being disabled. `DisableTap` field on the [Proxy struct](https://github.com/linkerd/linkerd2/blob/535a36af7c2da5adabec601735e4ad0a936ecf2d/pkg/charts/linkerd2/values.go#L105-L126) should be safe to remove.
```
if proxy.DisableTap != baseProxy.DisableTap {
	overrideAnnotations[k8s.ProxyDisableTapAnnotation] = strconv.FormatBool(proxy.DisableTap)
}
```

Once the merge conflicts on this are resolved there may be a few more things, but I'll review again once that happens! Let me know if you have any questions and thanks for doing this 😄

mayankshah1607:
Thanks @kleimkuhler!
I have made the changes as per your comment. Let me know if I may have missed anything else that may have to be cleaned up! :)

kleimkuhler (approved):
This looks great @mayankshah1607! Thanks for taking care of this.

PR #5673, opened 2021-02-05
checks: add proxy checks for core cp and extension pods

### LINKED ISSUE #5623: `check` should imply --proxy on the relevant namespace
The control plane can't be considered healthy if its proxies are not healthy.  `linkerd check` should check the state of all proxies in the core control plane (e.g. `linkerd check --proxy -n linkerd`). 

Similarly, extension checks should validate the state of the proxies in the extension namespaces.

Pothulapati:
Fixes #5623

This PR adds proxy checks for control-plane and extension pods
when the respective checks are run. This can make sure proxies
are working correctly and are able to communicate.

Currently, The following checks are added:

- proxy status checks
- proxy certificate checks
- proxy version up-to-date checks
- proxy and CLi version match check

The last two have been added as separate checkers so that they can be
warn only, unlike the first two where they have to be fatal

This PR also simplifies the `pod running` check w.r.t jaeger extension
by combining individual deployment based checks into a single one
just like other extensions.


pkg/healthcheck/healthcheck.go
```diff
@@ -1364,6 +1370,35 @@ func (hc *HealthChecker) allCategories() []Category {
...
+	if len(outdatedPods) > 0 {
+		podList := strings.Join(outdatedPods, "\n")
+		return fmt.Errorf("Some data plane pods are not running the current version:\n%s", podList)
+	}
+	return nil
+}
+
+// CheckIfProxyVersionsMatchWithCLI checks if the latest proxy version
+// matches that of the CLI
+func CheckIfProxyVersionsMatchWithCLI(pods []corev1.Pod) error {
```
dadjeibaah, line 1392: Is there a reason why this method is not attached to `hc *HealthChecker` while `CheckProxyVersionsUpToDate` is?

pkg/healthcheck/healthcheck.go
```diff
@@ -1199,6 +1199,29 @@ func (hc *HealthChecker) allCategories() []Category {
...
+					hintAnchor:          "l5d-api-cp-proxy-healthy",
+					retryDeadline:       hc.RetryDeadline,
+					surfaceErrorOnRetry: true,
+					fatal:               true,
+					check: func(ctx context.Context) error {
+						return hc.CheckProxyHealth(ctx, hc.ControlPlaneNamespace, hc.ControlPlaneNamespace)
+					},
+				},
+				{
+					description: "control plane proxies and cli versions match",
```
dadjeibaah, line 1213: Now that we've included data plane checks by default. Do we still need the checks that are characterized under `control-plane-version`? It seems as if we are running some checks twice now based on this output:
```
linkerd-api
-----------
√ control plane proxies are healthy
‼ control plane proxies and cli versions match
    linkerd-controller-8544bdbb77-8g6ft running git-ff0042f6 but cli running dev-87f6dbc0-dennis
    see https://linkerd.io/checks/#l5d-api-cp-cli-version for hints

linkerd-version
---------------
‼ cli is up-to-date
    unsupported version channel: dev-87f6dbc0-dennis

control-plane-version
---------------------
‼ control plane is up-to-date
    unsupported version channel: git-ff0042f6
    control plane running git-ff0042f6 but cli running dev-87f6dbc0-dennis

Status check results are √
```

dadjeibaah (commented):
Changes are looking great! I had a couple of comments. For my own understanding, do we require proxy checks for the multicluster extension?

pkg/healthcheck/healthcheck.go
Pothulapati: My thinking here was that this method does need `hc` anywhere for its functionality and hence did not add 🤔

jaeger/cmd/check.go
```diff
@@ -85,48 +85,48 @@ func jaegerCategory(hc *healthcheck.HealthChecker) *healthcheck.Category {
...
+			WithHintAnchor("l5d-jaeger-proxy-healthy").
 			Fatal().
 			WithRetryDeadline(hc.RetryDeadline).
 			SurfaceErrorOnRetry().
 			WithCheck(func(ctx context.Context) error {
-				// Check for Jaeger Injector pod
-				podList, err := hc.KubeAPIClient().CoreV1().Pods(jaegerNamespace).List(ctx, metav1.ListOptions{LabelSelector: "component=jaeger-injector"})
+				return hc.CheckProxyHealth(ctx, hc.ControlPlaneNamespace, jaegerNamespace)
+			}))
+
```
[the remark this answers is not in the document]
Pothulapati: Yeah, They were causing the test failures, and hence removed them. 

Now added them back and fixing the integration tests now.

pkg/healthcheck/healthcheck.go
Pothulapati: The `control-plane-version` check actually uses the version from the config and does not depend on the proxy versions of the control-plane components. Hence there is a difference

but I do also think it's very hard to create a case where they both are different 🤔

pkg/healthcheck/healthcheck.go
dadjeibaah: Ah, I see. That makes sense. 👍

viz/cmd/check.go
```diff
@@ -56,6 +59,8 @@ code.`,
 	cmd.PersistentFlags().BoolVar(&options.proxy, "proxy", options.proxy, "Also run data-plane checks, to determine if the data plane is healthy")
 	cmd.PersistentFlags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for all tests to pass")
 	cmd.PersistentFlags().StringVarP(&options.namespace, "namespace", "n", options.namespace, "Namespace to use for --proxy checks (default: all namespaces)")
+	cmd.PersistentFlags().StringVar(&options.versionOverride, "expected-version", options.versionOverride, "Overrides the version used when checking if Linkerd is running the latest version (mostly for testing)")
```
adleong, line 62: I don't think I've ever needed to use these flags.  I think it's probably safe to get rid of them.

adleong (commented):
The category name of `linkerd-api` doesn't really make sense to me for this.  Shouldn't it be `proxies` or something like that?

```
linkerd-api
-----------
√ control plane proxies are healthy
√ control plane proxies are up-to-date
‼ control plane proxies and cli versions match
    linkerd-controller-7789879c97-hmxzg running edge-21.2.1 but cli running git-11d6c879
    see https://linkerd.io/checks/#l5d-api-cp-cli-version for hints
```

adleong (commented):
The output when I run `bin/linkerd check --proxy -n linkerd` isn't quite what I'd expect:

```
linkerd-api
-----------
√ control plane proxies are healthy
√ control plane proxies are up-to-date
‼ control plane proxies and cli versions match
    linkerd-controller-7789879c97-hmxzg running edge-21.2.1 but cli running git-11d6c879
    see https://linkerd.io/checks/#l5d-api-cp-cli-version for hints

linkerd-data-plane
------------------
√ data plane namespace exists
√ data plane proxies are ready
√ data plane is up-to-date
‼ data plane and cli versions match
    linkerd-controller-7789879c97-hmxzg running edge-21.2.1 but cli running git-11d6c879
    see https://linkerd.io/checks/#l5d-data-plane-cli-version for hints
```

I think these two sections both serve the same purpose so having them both is kind of redundant.  At the same time, the checks in both sections are slightly different (or at least the wording is different) which feels inconsistent.  

So my question is: should `linkerd check` just be equivalent to `linkerd check --proxy -n linkerd`?

Pothulapati:
> So my question is: should linkerd check just be equivalent to linkerd check --proxy -n linkerd?

Hmm, That is easier to do actually. The current approach adds separate checkers re-using the logic. We can instead just make `linkerd check` update logic to do `linkerd check --proxy -n linkerd` as you mentioned but is the wording fine in that case? i.e data-plane. Wouldn't that be confusing 🤔

adleong:
We can change the wording if we like.  For example, saying "proxies" or "proxy-health" instead of "data-plane".

alpeb:
@Pothulapati the upgrade integration tests is failing after having added the prune flag because the upgrade is performed in a two-stage manner. First `linkerd upgrade config` and then `linkerd upgrade control-plane`. The output of the latter is what's getting passed to `kubectl apply --prune`, so all the resources from the `config` are getting pruned.

pkg/healthcheck/healthcheck.go
```diff
@@ -1270,6 +1274,47 @@ func (hc *HealthChecker) allCategories() []Category {
...
+					retryDeadline:       hc.RetryDeadline,
+					surfaceErrorOnRetry: true,
+					fatal:               true,
+					check: func(ctx context.Context) error {
+						return hc.CheckProxyHealth(ctx, hc.ControlPlaneNamespace, hc.ControlPlaneNamespace)
+					},
+				},
+				{
+					description: "control plane proxies are up-to-date",
+					hintAnchor:  "l5d--cp-proxy-version",
```
kleimkuhler, line 1292:
```suggestion
					hintAnchor:  "l5d-cp-proxy-version",
```

kleimkuhler (approved):
Looks good aside from hint anchor typo!

kleimkuhler:
After #5762 changed how flags are passed through to extension checks, we weren't passing through `expected-version`, so the check output was failing when running check on extensions.

That should be fixed by c7dd607, but now it seems multicluster doesn't support that flag. Moving forward with edge for now, but we can get this in tomorrow and do another small follow-up edge.

Pothulapati:
@kleimkuhler I updated this PR by also including these flags in `mc check` but my worry is that Is adding these test flags into extensionFlags is fine? As this means that external extension checks would fail when these flags are used (just like multi-cluster, as they would not support them)?

cc: @adleong

kleimkuhler:
@Pothulapati I'm thinking that we would make these required flags specified in `EXTENSIONS.md`. On the other hand, these flags are required mainly to get the expected output in tests. Is there a way you think you could get the `.golden` file output to be what we expected without having to pipe through the the `expected-version` flag?

Pothulapati:
@kleimkuhler currently , I don't have an alternative approach but thought it was a good point to discuss

kleimkuhler:
@Pothulapati I think that makes sense! We hide other flags that should really only be used for development/testing.

adleong:
I agree, we should add support for this flag to all of the extensions, ensure that we pass it to extension checks, and update the EXTENSION docs to list it.

adleong (commented):
Ok, actually, the version checking stuff here is more confusing than I thought.  Here's the current situation, as I understand it:

In the core Linkerd, we have two checks related to version:
* Do Linkerd controllers and proxies version match the Latest released version?
* Do Linkerd controllers and proxies version match the in-use CLI version?

Since the in-use CLI might not be the latest released version, these checks are different.  Thus we have two different flags: `expected-version` which controls what we consider to be the "Latest releaed" version (used in the first check above) and `cli-version-override` which controls what we consider to be the in-use CLI version (used in the second check above).

In order for extensions to do checks the same way, we would need to pass both of these flags to each extension and each extension would need to do both of these checks.

I think we can absorb some of this complexity in Linkerd core and provide a simpler interface to the extensions.  I propose:

We change the core checks to:

* Does the in-use CLI version match the latest released version?
* Do Linkerd controllers and proxies version match the in-use CLI version?

By structuring the checks this way, the extensions only need to do one version related check:

* Do Linkerd controllers and proxies version match the in-use CLI version?

And we would only need to pass one version flag to the extensions: `expected-version` which we would set to the current in-use CLI version.

Does this make sense?

kleimkuhler:
@adleong Yes I'm +1 on this. I think making the check for in-use CLI version unique to core makes sense.

adleong:
I think that given the proximity to the stable-2.10 release, it feels risky to be changing flags around, especially changing the flag contract with extensions.

I think, therefore, it's my preference that for now we limit this change to just the core control plane.  Once 2.10 is out the door, we'll have more time to refactor this into a more holistic system that allows extensions to do proxy version checking in a principled way.

dadjeibaah (approved):
LGTM! 📦 🚢

pkg/healthcheck/healthcheck.go
```diff
@@ -1395,6 +1424,41 @@ func (hc *HealthChecker) allCategories() []*Category {
...
+	outdatedPods := []string{}
+	for _, pod := range pods {
+		proxyVersion := k8s.GetProxyVersion(pod)
+		if err := versions.Match(proxyVersion); err != nil {
+			outdatedPods = append(outdatedPods, fmt.Sprintf("\t* %s (%s)", pod.Name, proxyVersion))
+		}
+	}
+	if len(outdatedPods) > 0 {
+		podList := strings.Join(outdatedPods, "\n")
+		return fmt.Errorf("Some data plane pods are not running the current version:\n%s", podList)
```
alpeb, line 1445: This will require rewording because the check is now used in the control plane, not just the data plane. Something like just "Some proxies are not running the current version: \n%s".

alpeb:
@Pothulapati do you mind opening an issue to tracking the multicluster followup?

test/integration/install_test.go
```diff
@@ -414,7 +418,7 @@ func TestInstallOrUpgradeCli(t *testing.T) {
 	// that we intend to be delete in this stage to prevent it
 	// from deleting other resources that have the
 	// label
-	out, err = TestHelper.KubectlApplyWithArgs(out, []string{
+	cmdOut, err := TestHelper.KubectlApplyWithArgs(out, []string{
```
alpeb, line 421: :+1:

alpeb (approved):
LGTM :+1: 
One tiny nit: you can get rid of the function `ContainsOutput` in `testutil/test_helper.go` :wink:

Pothulapati:
> do you mind opening an issue to tracking the multicluster followup?

Raised https://github.com/linkerd/linkerd2/issues/6046

PR #5674, opened 2021-02-05
Fix spelling issue w/ sidecarContainers key in linkerd-viz helm chart

n-oden:
Problem:

The name of the `sidecarContainers` chart value key is misspelled
`sideCarContainers` in the default values file and the README: anyone
who uses the misspelled key will be confused and unhappy to see that
their sidecar container is not added to the pod because the Helm
template is looking for `.Values.prometheus.sidecarContainer`

Solution:

Change the first `c` character to lower-case as the template expects. :)


wmorgan:
Feels a bit complex, can you simplify this further?

alpeb (approved):
LGTM :+1:

kleimkuhler (approved):
Thanks @n-oden!

Issue #5058, opened 2020-10-09
cli: Don't print usage on API errors


olix0r:
When the linkerd control plane encounters an API error, we print a usage message:

```
            Error: StatSummary API error: server_error: server error: 503
            Usage:
              linkerd stat [flags] (RESOURCES)
            
            Examples:
              # Get all deployments in the test namespace.
              linkerd stat deployments -n test
            
              # Get the hello1 replication controller in the test namespace.
              linkerd stat replicationcontrollers hello1 -n test
            
              # Get all namespaces.
              linkerd stat namespaces
            
              # Get all inbound stats to the web deployment.
              linkerd stat deploy/web
            
              # Get all inbound stats to the pod1 and pod2 pods
              linkerd stat po pod1 pod2
            
              # Get all inbound stats to the pod1 pod and the web deployment
              linkerd stat po/pod1 deploy/web
            
              # Get all pods in all namespaces that call the hello1 deployment in the test namespace.
              linkerd stat pods --to deploy/hello1 --to-namespace test --all-namespaces
            
              # Get all pods in all namespaces that call the hello1 service in the test namespace.
              linkerd stat pods --to svc/hello1 --to-namespace test --all-namespaces
            
              # Get all services in all namespaces that receive calls from hello1 deployment in the test namespace.
              linkerd stat services --from deploy/hello1 --from-namespace test --all-namespaces
            
              # Get all trafficsplits and their leaf services.
              linkerd stat ts
            
              # Get the hello-split trafficsplit and its leaf services.
              linkerd stat ts/hello-split
            
              # Get all trafficsplits and their leaf services, and metrics for any traffic coming to the leaf services from the hello1 deployment.
              linkerd stat ts --from deploy/hello1
            
              # Get all namespaces that receive traffic from the default namespace.
              linkerd stat namespaces --from ns/default
            
              # Get all inbound stats to the test namespace.
              linkerd stat ns/test
            
            Flags:
              -A, --all-namespaces          If present, returns stats across all namespaces, ignoring the "--namespace" flag
                  --from string             If present, restricts outbound stats from the specified resource name
                  --from-namespace string   Sets the namespace used from lookup the "--from" resource; by default the current "--namespace" is used
              -h, --help                    help for stat
              -n, --namespace string        Namespace of the specified resource (default "default")
              -o, --output string           Output format; one of: "table" or "json" or "wide" (default "table")
              -l, --selector string         Selector (label query) to filter on, supports '=', '==', and '!='
[...]

PR #5678, opened 2021-02-07
Fix for CLI printing command usage for API Errors with `linkerd viz stat`

piyushsingariya:

Potential fix for #5058 

Problem: CLI prints command usage for 'linkerd viz stat' in-case of API errors
Solution: Print the error and the exit using `os.Exit(1)` to avoid Cobra printing usage

### DIFF
--- a/viz/cmd/stat.go
+++ b/viz/cmd/stat.go
@@ -193,6 +193,10 @@ If no resource name is specified, displays stats about all resources of the spec
 			for num, req := range reqs {
 				go func(num int, req *pb.StatSummaryRequest) {
 					resp, err := requestStatsFromAPI(client, req)
+					if err != nil {
+						fmt.Fprint(os.Stderr, err.Error())
+						os.Exit(1)
+					}
 					rows := respToRows(resp)
 					c <- indexedResults{num, rows, err}
 				}(num, req)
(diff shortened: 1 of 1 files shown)


kleimkuhler (approved):
Thanks @piyushsingariya this looks good for the `stat` command!

I don't think this completely closes the issue since this is just one place where this occurs, but it will fix the example in that issue which is great.

piyushsingariya:
The issue was mentioned with `stat` command so I made changes there only. What are the other command the issue concerns? :thinking: 
Or we are looking for something which can solve all such problems in one go?
> I don't think this completely closes the issue since this is just one place where this occurs, but it will fix the example in that issue which is great.

kleimkuhler:
@piyushsingariya The issue is more generally stating that there are places where a CLI command fails for a reason other than user error, and we still print the usage help text.

The `stat` command mentioned in the issue was only an example of a single place where this happens. We don't have a list of all the other places this occurs—there may be none remaining after this fix—but it is a more general issue than this specific issue in the the `stat` command.

In order to completely close the issue, I think a pass over each command double checking that errors are returned only when it's the user who has errored is necessary.

All that being said, I still think this PR as it stands is valuable to merge.

alpeb:
@piyushsingariya Given that this means we bail out early on the first error, the `err` field in the `indexedResults` struct is no longer used. Would you mind removing it?

piyushsingariya:
@alpeb `indexedResults` struct is being used to store returned responses from a range of requests to API, which is being used later on. So I made some minor changes, if we found any API error during any of the requests, it will be printed during the traversal of `indexedResults`

alpeb (approved):
Sounds good to me :+1:

PR #5690, opened 2021-02-08
Provide trust anchors as env var to identity controller

### LINKED ISSUE #3843: Allow trustAnchorsPem ca cert value to be read from Kubernetes Secrets
## Feature Request
Support for trustAnchorsPem ca cert value in linkerd-config to be read from an existing kubernetes secrets

Automating the deployment of linkerd configuration with an existing tls certificate

Allow alternate way to provide trustAnchorsPem cert.  One possible solution is to provide a key trustAnchorsPemSecret: my-namespace/secret-name in identityContext

No alternatives

### Is there another way to solve this problem that isn't as good a solution?
None

User Creates a tls certificate using external provider like cert-manager.  The secret name will then is referred in linkerd-config global -> identityContext -> trustAnchorsPemSecret

mgoltzsche:
Currently the identity controller is the only component that receives the CA certificate / trust anchors as option `-identity-trust-anchors-pem` instead of an env var.
This stops one from letting it read the trust anchors from a Secret that is managed by e.g. cert-manager.

This PR uses an env var instead of the option to provide the trust anchors. For most helm chart users this doesn't change anything. However using kustomize the helm output manifest can now be adjusted (again) so that the certificate is loaded from a ConfigMap or Secret like in [this example](https://github.com/mgoltzsche/khelm/tree/master/example/kpt/linkerd) which aims to produce a static manifest to make the installation/update more declarative and support GitOps workflows.

This PR does not provide chart options/values to specify Secrets upfront - it would introduce dependencies to other operators.

Relates to #3843, see https://github.com/linkerd/linkerd2/issues/3843#issuecomment-775516217


kleimkuhler:
@mgoltzsche I'm going to move this into draft if it's a work in progress. You'll still be able to push updates and have CI run.

Once it is review ready, feel free to update the status of it and we'll take a look!

controller/cmd/identity/main.go
```diff
@@ -39,7 +41,7 @@ func Main(args []string) {
 	controllerNS := cmd.String("controller-namespace", "", "namespace in which Linkerd is installed")
 	identityScheme := cmd.String("identity-scheme", "", "scheme used for the identity issuer secret format")
 	trustDomain := cmd.String("identity-trust-domain", "", "configures the name suffix used for identities")
-	encodedIdentityTrustAnchorPEM := cmd.String("identity-trust-anchors-pem", "", "Base64 encoded trust anchors certificate")
+	encodedIdentityTrustAnchorPEM := cmd.String("identity-trust-anchors-pem", os.Getenv(envTrustAnchors), fmt.Sprintf("Base64 encoded trust anchors certificate (%s)", envTrustAnchors))
```
adleong, line 44: To keep things simple, since we're now reading this value from an env variable, we should just get rid of the flag for it entirely.  Easier to only read this value from one place.

adleong (commented):
This seems pretty reasonable to me.  One thing to note is that if you customize this to load the trust anchors from a configmap, you'll also need to make sure to restart the identity controller if the configmap is updated, otherwise the changes will not get picked up.  But if you're customizing this manifest, you've already gone off the beaten path so I'll just assume you know what you're doing 😁

controller/cmd/identity/main.go
mgoltzsche: Done. Actually it would be nicer if the trust anchors would not need to be base64-encoded but some other tests fails when I change that. Therefore (and since I have to transform the config and trust anchors Secret within the cluster dynamically anyway within my kustomization) I leave it base64 encoded.

mgoltzsche:
In my kustomization I am going to let [reloader](https://github.com/stakater/Reloader) redeploy the identity controller whenever the trust anchors ConfigMap changes.

adleong (approved):
Thanks @mgoltzsche!

kleimkuhler:
@mgoltzsche Sorry we did not get a second review in before these merge conflicts happened. If you could please resolve those, I'll make sure to take a look once that is done.

mgoltzsche:
@kleimkuhler done.

kleimkuhler (approved):
Looks good, thanks @mgoltzsche!

PR #5692, opened 2021-02-09
Add EXTENSIONS.md

adleong:
Extensions specification for Linkerd extensions.


EXTENSIONS.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Developing Extensions
+
+The extension must be an executable file named `linkerd-name` where `name` is
+the name of the extension. The name must not be any of the built-in Linkerd
+commands (e.g. `check`) or extensions (e.g. `viz`).
+
+The extension must accept the following flags and respect them any time that
+it communicates with the Kubernetes API.  All of these flags must be accepted
+but may be ignored if they are not applicable.
```
siggy, line 27: i like that these flags are recommended but not required. we could consider a similar pattern to `kubectl plugin`, with provided libraries, or, we could even consider re-using the `kubectl` libraries/examples:
https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/#kubectl-options
https://github.com/kubernetes/cli-runtime/

also, how do we expect `linkerd` flags to interact with extension flags? what if a user sets two different `--context` values? for comparison, `kubectl` disallows all flags on itself when running plugins:
```bash
$ kubectl --context foo access-matrix
Error: flags cannot be placed before plugin name: --context
```

EXTENSIONS.md
```diff
@@ -0,0 +1,113 @@
...
+
+* `--api-addr`: Override kubeconfig and communicate directly with the control plane at host:port (mostly for testing)
+* `--context`: Name of the kubeconfig context to use
+* `--help`/`-h`: Print help message
+* `--kubeconfig`: Path to the kubeconfig file to use for CLI requests
+* `--linkerd-namespace`/`-l`: Namespace in which Linkerd is installed [$LINKERD_NAMESPACE]
+* `--verbose`: Turn on debug logging 
+
+The extension must implement these commands:
```
siggy, line 36: similar to the flags, i think these subcommands should be optional. they imply a specific type of extension that installs resources on a cluster. looking over kubectl's plugins, many (a majority?) seem to be cli-only: https://krew.sigs.k8s.io/plugins/

as a bonus, i think stateless / cli-only extensions create a lower barrier-to-entry for developers. i'd love to see an extension that reads linkerd metrics and outputs my network topology in graphviz format.

EXTENSIONS.md
```diff
@@ -0,0 +1,113 @@
...
+
+### `linkerd-name install`
+
+This command must print the Kubernetes manifests for the extension as yaml
+which is suitable to be passed to `kubectl apply -f`. These manifests must
+include a Namespace resource with the label `linkerd.io/extension=name` where
+`name` is the name of the extension. This allows Linkerd to detect installed
+extensions.
```
siggy, line 44: is the intent for linkerd to iterate over all namespaces, searching for the `linkerd.io/extension=name`? how does this set of extensions relate to the set of binaries detected via PATH?

EXTENSIONS.md
```diff
@@ -0,0 +1,113 @@
...
+
+The final line of output should be either `Status check results are √` or 
+`Status check results are ×`.
+
+In addition to the flags described above, `linkerd-name check` must accept the
+following flags:
+
+* `--namespace`/`-n`: Namespace to use for –proxy checks (default: all namespaces)
+* `--output`/`-o`: Output format. One of: basic, json
+* `--proxy`: Only run data-plane checks, to determine if the data plane is healthy
```
siggy, line 79: related to earlier comments, these `--namespace` and `--proxy` flags on the `check` subcommand feel super-specific to a certain type of extension. i could image extensions that install resources on a cluster, but do not have a proxy component.

siggy (commented):
excited for this! really appreciate the simple, unix-based approach. most of my comments focus on what parts of the linkerd+extension surface area are required vs. recommended.

EXTENSIONS.md
```diff
@@ -0,0 +1,113 @@
+# Linkerd Extensions
```
olix0r, line 1: Perhaps better to focus this as *CLI* extensions.

EXTENSIONS.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Installing Extensions
+
+To install an extension, first download the extension executable and put it
+in your PATH. The extension can then be installed into your Kubernetes cluster
+by running `linkerd <extension name> install | kubectl apply -f -`. Similarly,
+it can be uninstalled by running
+`linkerd <extension name> uninstall | kubectl delete -f -`.
+
+A full list of installed extensions can be printed by running `linkerd check`.
```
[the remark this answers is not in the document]
adleong: I will re-word this to clarify: this lists extensions that have been installed in the cluster.  I.e. it lists all namespaces with the `linkerd.io/extension` annotation.

We don't provide a mechanism for listing all extension executables on your path.  I'm not sure this is actually useful and also it will erroneously list things like `linkerd-stable-1.2.3` as extensions since they fit the naming scheme.

EXTENSIONS.md
adleong: I don't think that it's possible to set the same flag to different values on the same command.  Conceptually, the position of the flag is not important: `kubectl --context foo access-matrix` is semantically equivalent to `kubectl access-matrix --context foo`.  I believe that kubectl requires that flags come after the plugin name because of the way they do argument parsing.  It's fairly likely that we would do argument parsing the same way and therefore impose the same restriction on flag ordering.

EXTENSIONS.md
adleong: That's fair.  This suggests two distinct types of extensions: installable extensions for which install, uninstall, and check must be implemented; and stateless extensions which operate without creating any resources in the cluster.

EXTENSIONS.md
adleong: We use a label selector rather than iterating over all namespaces, but yes, that's the idea.  Installed extensions are discovered by listing all namespaces with the `linkerd.io/extension` label.  For each extension discovered in this way, we look for an executable on the path named `linkerd-<extension name>`.

EXTENSIONS.md
adleong: The `--proxy` flag indicates that you want to do data-plane checks.  If your extension doesn't have any data-plane component, it should just return an empty success.

EXTENSIONS.md
```diff
@@ -0,0 +1,121 @@
...
+scoped resources are listed.
+
+### `linkerd-name check`
+
+This command must perform any appropriate health checks for the extension
+including but not limited to checking that the extension resources exist and
+are in a healthy state. This command must exit with a status code of 0 if the
+checks pass or with a nonzero status code if they do not pass. For consistency,
+it is recommended that this command follows the same output formatting as
+`linkerd check`, e.g.
```
olix0r, line 64: As mentioned [here](https://github.com/linkerd/linkerd2/pull/5762#issuecomment-781470004), I think this might be too simplistic. I think that `linkerd` should ultimately be responsible for output formatting -- probably by _always_ specifying `--output=json` -- and the extension CLI should only be responsible for returning structured data to the main cli process.

EXTENSIONS.md
```diff
@@ -0,0 +1,126 @@
...
+but may be ignored if they are not applicable.
+
+* `--api-addr`: Override kubeconfig and communicate directly with the control
+  plane at host:port (mostly for testing)
+* `--context`: Name of the kubeconfig context to use
+* `--as`: Username to impersonate for Kubernetes operations
+* `--as-group`: Group to impersonate for Kubernetes operations
+* `--help`/`-h`: Print help message
+* `--kubeconfig`: Path to the kubeconfig file to use for CLI requests
+* `--linkerd-namespace`/`-l`: Namespace in which Linkerd is installed
```
alpeb, line 36:
```suggestion
* `--linkerd-namespace`/`-L`: Namespace in which Linkerd is installed
```

EXTENSIONS.md
```diff
@@ -0,0 +1,126 @@
...
+include a Namespace resource with the label `linkerd.io/extension=name` where
+`name` is the name of the extension. This allows Linkerd to detect installed
+extensions.
+
+### `linkerd-name uninstall`
+
+This command must print manifests for all Kubernetes resources belonging to
+this extension as yaml which is suitable to be passed to `kubectl delete -f`.
+Special care should be taken to make sure both namespace-scoped and cluster-
+scoped resources are listed.
```
alpeb, line 55: All cluster-scoped resources need to be passed, including the extension namespace. When the namespace gets deleted, the namespace-scoped resources will automatically get deleted, so there's no need to pass them.

alpeb (approved):
:guitar:

EXTENSIONS.md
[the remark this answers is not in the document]
kleimkuhler: I know this got updated on the website PR with listing by the label, but looks like it still needs to here as well.

EXTENSIONS.md
```diff
@@ -0,0 +1,125 @@
...
+A full list of installed extensions can be printed by running `linkerd check`.
+
+## Developing Extensions
+
+The extension must be an executable file named `linkerd-name` where `name` is
+the name of the extension. The name must not be any of the built-in Linkerd
+commands (e.g. `check`) or extensions (e.g. `viz`).
+
+The extension must accept the following flags and respect them any time that
+it communicates with the Kubernetes API.  All of these flags must be accepted
```
kleimkuhler, line 26:
```suggestion
it communicates with the Kubernetes API.  All of these flags must be accepted,
```

kleimkuhler (approved):
It seems like the above conversations are all pretty much resolved, especially with how the `linkerd check` command expects json output.

I have two smaller comments and one is just a nit, but otherwise looks good!

PR #5694, opened 2021-02-09
viz: support default* fields for global overrides

### LINKED ISSUE #5685: YAML anchors don't work as expected when overriding flags
### What is the issue?

We make use of a few YAML anchors in the Helm templates to avoid repetition. It turns out that only works as long as anchors refer to variables in the same document. If one sets a value (via Helm or the CLI) in a document using `-f` or through `--set`, and if that value is referred as an anchor in the main `values.yaml` file, the overridden value will be ignored.

### How can it be reproduced?

```bash
$ linkerd viz install --set defaultLogLevel=debug | ag log-level
        - -log-level=info
        - -log-level=info
        - -log-level=info

$ helm template --set defaultLogLevel=debug linkerd-edge/linkerd-viz | ag log-level
        - -log-level=info
        - -log-level=info
        - -log-level=info
```

### Environment

- Kubernetes Version: 1.19
- Host OS: Linux
- Linkerd version: edge-21.2.1

### Possible solution

Stop using anchors completely, and instead rely on the `default` function directly in the template files. Instead of using anchors for a value that we wish to reuse, we set leave it to an empty value in `values.yaml`, and then in the template file use the default value (like `defaultLogLevel`) only if the target value hasn't been overridden.

### Additional context

This is the upstream issue describing the problem: helm/helm#2336

Pothulapati:
Fixes #5685

Currently, YAML anchors are not supported through Helm Values
fields. These are required for the default* flags which are
used to override flags across components.

For them to work, YAML anchors had to be removed and rely on
default function directly in the tempalte files.


alpeb (commented):
Looks pretty good to me :+1: 

I think the images can't be pulled because the new default pull policy is `Always`, which attempts to pull the image from ghcr.io.

Also:
- Don't we need to do the same thing for the other anchored vars, i.e. `linkerdVersion` and `clusterDomain`?
- Can you complete the comments for the now-empty vars, saying something like "defaults to defaultLogLevel's value"?

Pothulapati:
> Don't we need to do the same thing for the other anchored vars, i.e. linkerdVersion and clusterDomain?

I don't think so. As these flags are never mentioned as global/ higher level flags. For the currently changed ones we do, and hence they need to act as the global overrides but for `linkerdVersion` and `clusterDomain`, we use internally yaml anchors to re-use stuff which is not external facing? We should definitely update the docs to reflect the same, if we there are instances where we are not setting all fields using the same yaml anchor... WDYT?

viz/charts/linkerd-viz/values.yaml
```diff
@@ -160,7 +168,8 @@ tapInjector:
   objectSelector:
     # matchLabels:
     #   foo: bar
-  UID: *uid
+  # UID for the tapInjector resource
```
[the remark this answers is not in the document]
Pothulapati: Instead of this we can use `@default` to specify comments as per the docs https://github.com/norwoodj/helm-docs#default-valuescolumn

viz/charts/linkerd-viz/values.yaml
```diff
@@ -61,16 +63,20 @@ metricsAPI:
   # -- number of replicas of the metrics-api component
   replicas: 1
   # -- log level of the metrics-api component
-  logLevel: *log_level
+  # @default -- defaults to defaultLogLevel
```
Pothulapati, line 66: Should we remove this `defaults to` part? 🤔 @alpeb

kleimkuhler: @Pothulapati I'm +1 for removing the `defaults to` for each of these descriptions. The column it's filling is showing the default value, so I think its a little redundant to then also say `defaults to ...`.

viz/charts/linkerd-viz/values.yaml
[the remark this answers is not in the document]
kleimkuhler: I think you missed this in the review feedback commit?

kleimkuhler (commented):
So this definitely addresses the issue, but does it totally fix it? Don't we need to make similar changes in the core install as well as each of the extensions?

At least in the core install `values.yaml`, we make similar use of anchors for log level and such.

viz/charts/linkerd-viz/values.yaml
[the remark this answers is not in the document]
Pothulapati: Righhttt. Done

alpeb:
> > Don't we need to do the same thing for the other anchored vars, i.e. linkerdVersion and clusterDomain?
> 
> I don't think so. As these flags are never mentioned as global/ higher level flags. For the currently changed ones we do, and hence they need to act as the global overrides but for `linkerdVersion` and `clusterDomain`, we use internally yaml anchors to re-use stuff which is not external facing? We should definitely update the docs to reflect the same, if we there are instances where we are not setting all fields using the same yaml anchor... WDYT?

It appears the `linkerd_version` anchor isn't used, so we can just remove that. OTOH, a user could override `clusterDomain` and `identityTrustDomain` would remain `cluster.local`, which could be surprising...

viz/charts/linkerd-viz/values.yaml
alpeb: :+1:

alpeb (commented):
I second @kleimkuhler's comment about doing the same for the other values.yaml files.

Also, some `UID` values are missing the `@default` comment.

adleong (commented):
I find yaml anchors to be confusing and have a lot of unexpected behavior, so I'm +100 on removing them everywhere if possible.

alpeb:
> It appears the `linkerd_version` anchor isn't used

I don't know what I was looking at, that's still being used :man_shrugging:

Pothulapati:
Okay. Made the following changes:
- Removed complete usage of yaml anchors from viz as it was possible.
- For the core control-plane, I had to leave `linkerd_version` and `imagePullPolicy` as they were being used with `debugContainer` which is not rendered anywhere but stored directly preventing us from removing them and using the `default` logic.

alpeb:
Can you also please check the anchors in the `values-ha.yaml` files for core and viz?

alpeb:
> Can you also please check the anchors in the `values-ha.yaml` files for core and viz?

We discussed this out-of-band and concluded anchors for resources actually make sense in there because users won't be expecting all resources to be updated when they update say `controllerResources` :+1:

alpeb (approved):
LGTM :+1:

charts/linkerd2/templates/config.yaml
```diff
@@ -25,6 +25,9 @@ data:
   {{- if kindIs "map" $values.proxyInjector -}}
     {{- $_ := unset $values.proxyInjector "keyPEM"}}
   {{- end -}}
+  {{- if (empty $values.identityTrustDomain) -}}
+  {{- $_ := set $values "identityTrustDomain" $values.clusterDomain}}
+  {{- end -}}
```
kleimkuhler, line 30: Awesome 👍

kleimkuhler (approved):
Looks great!

charts/linkerd2/values.yaml
```diff
@@ -14,7 +14,7 @@ clusterNetworks: "10.0.0.0/8,100.64.0.0/10,172.16.0.0/12,192.168.0.0/16"
 # -- Docker image pull policy
 imagePullPolicy: &image_pull_policy IfNotPresent
```
adleong, line 15: Can we remove this anchor now?

adleong: Similarly, can we remove the `&linkerd_version` anchor?

adleong: I'm not sure what you mean about the debug container not getting rendered anywhere.  Isn't it rendered from the _debug.tpl partial into the pod patch generated by the proxy injector?  Can't that partial also include the same defaulting logic as everything else?

Pothulapati: Yep,  You are right. I was under the impression that the injector depends on these values but its similar to others i.e proxy, proxy-init. So, Updated.

adleong (approved):

charts/patch/templates/patch.json
```diff
@@ -67,7 +67,7 @@
       {{- include "partials.proxy-init" . | fromYaml | toPrettyJson | nindent 6 }}
   },
   {{- end }}
-  {{- with .Values.debugContainer }}
+  {{- if .Values.debugContainer }}
```
alpeb, line 70: Why this change?

alpeb: Oh I see, the partial now needs access to all the values in order to use `default` :+1:

alpeb (approved):

PR #5702, opened 2021-02-09
Support opaque ports annotation on services

### LINKED ISSUE #5650: multicluster: Mirror should propagate opaqueness
When an outbound proxy is trying to communicate with a remote service, it needs a way to know whether the service is opaque.

Consider the following example:

* Two clusters, _east_ and _west_.
* Mirror configured in _west_ to sync services from _east_.
* A service in _east_, called _database_, with port `opaque-ports: "3306"`
* A Service & TrafficSplit in _west_, called _database_, that redirects all traffic to _database-east_.
* A client in _west_ attempts to connect to _database:3306_.

The client needs to be able to discover that the target is opaque. There is currently no means for this.

### goals

* In _west_, profile lookups for _database:3306_ OR _database-east:3306_ (via their ClusterIPs) must be marked as opaque

### proposal

* Support the `opaque-ports` annotation on `Service` resources in addition to workloads.
  * Destination service profile lookups check the service-level annotation before doing per-pod lookups.
* Mirrored services include original service's annotations

kleimkuhler:
This change introduces an opaque ports annotation watcher that will send
destination profile updates when a service or its namespace has its opaque ports
annotation change.

The user facing change introduced by this is that the opaque ports annotation is
now required on services or namespaces when using the multicluster extension.
This is because the service mirror will create mirrored services in the source
cluster, and destination lookups in the source cluster need to discover that the
workloads in the target cluster are opaque protocols.

This also fixes an existing bug where changes to the opaque port annotations on
the namespace do not update a destination client. Updates are now merged between
namespace and service — giving priority to the annotation on a service.

### Why

Closes #5650

### How

The destination server now has a new opaque ports annotation watcher. When a
client subscribes to updates for a service name or cluster IP, the `GetProfile`
method creates a profile translator stack that passes updates through resource
adaptors such as: traffic split adaptor, service profile adaptor, and now opaque
ports adaptor.

When the annotation on a namespace or service changes, the opaque ports watcher
will merge those changes — giving priority to the service annotation. This
update is passed through to the client where the `opaque_protocol` field will
either be set to true or false.

There are a few scenarios to consider that the opaque ports watcher handles:

- If both namespace and service have the annotation, the client should get the
  update for the service annotation.

  - If the annotation is removed from the service, the client should receive an
    update with the namespace annotation.
  - If the service is deleted, the stream stays open so the client should
    receive an update with the namespace annotation.
  - If the annotation is removed from the namespace there should be no update.

- If neither the namespace or the service has the annotation, the client should
  receive an update with no opaque ports — and therefore `opaque_protocol` is
  false.

  - If either the namespace or the service has the annotation added, the client
    should receive that update.

### Testing

Unit test have been added to the watcher as well as the destination server.

An integration test has been added that tests the opaque port annotation on a
service.

For manual testing, using the destination server scripts is easiest:

```
# install Linkerd

# start the destination server
$ go run controller/cmd/main.go destination -kubeconfig ~/.kube/config

# Create a service or namespace with the annotation and inject it

# get the destination profile for that service and observe the opaque protocol field
$ go run controller/script/destination-client/main.go -method getProfile -path test-svc.default.svc.cluster.local:8080
INFO[0000] fully_qualified_name:"terminus-svc.default.svc.cluster.local" opaque_protocol:true retry_budget:{retry_ratio:0.2 min_retries_per_second:10 ttl:{seconds:10}} dst_overrides:{authority:"terminus-svc.default.svc.cluster.local.:8080" weight:10000} 
INFO[0000]                                              
INFO[0000] fully_qualified_name:"terminus-svc.default.svc.cluster.local" opaque_protocol:true retry_budget:{retry_ratio:0.2 min_retries_per_second:10 ttl:{seconds:10}} dst_overrides:{authority:"terminus-svc.default.svc.cluster.local.:8080" weight:10000} 
INFO[0000]
```

### DIFF
--- a/controller/api/destination/endpoint_translator.go
+++ b/controller/api/destination/endpoint_translator.go
@@ -216,7 +216,7 @@ func (et *endpointTranslator) sendClientAdd(set watcher.AddressSet) {
 			err error
 		)
 		if address.Pod != nil {
-			opaquePorts, getErr := getOpaquePortsAnnotations(address.Pod)
+			opaquePorts, getErr := getPodOpaquePortsAnnotations(address.Pod)
 			if getErr != nil {
 				et.log.Errorf("failed getting opaque ports annotation for pod: %s", getErr)
 			}
--- a/controller/api/destination/opaque_ports_adaptor.go
+++ b/controller/api/destination/opaque_ports_adaptor.go
@@ -1,108 +1,39 @@
 package destination
 
 import (
-	"strconv"
-
 	"github.com/linkerd/linkerd2/controller/api/destination/watcher"
 	sp "github.com/linkerd/linkerd2/controller/gen/apis/serviceprofile/v1alpha2"
-	"github.com/linkerd/linkerd2/controller/k8s"
-	pkgk8s "github.com/linkerd/linkerd2/pkg/k8s"
-	"github.com/linkerd/linkerd2/pkg/util"
-	logging "github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
 )
 
-// opaquePortsAdaptor implements EndpointUpdateListener so that it can watch
-// endpoints for services. When endpoints change, it can check for the opaque
-// ports annotation on the pods and update the list of opaque ports.
-//
-// opaquePortsAdaptor also implements ProfileUpdateListener so that it can
-// receive updates from trafficSplitAdaptor and merge the service profile by
-// adding the list of opaque ports.
-//
-// When either of these implemented interfaces has an update,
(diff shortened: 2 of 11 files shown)


olix0r (approved):
Not speaking to code changes, but I spent all day testing this---actually I spent all day misconfiguring clusters, testing this was actually pretty easy---and it works as expected

kleimkuhler:
I'm closing this as it's going to change a lot after the already closed #5737. Now that services have the namespace annotations merged into them, we don't need to do anything about the namespace watching and merging going on in this PR. I will reopen with a less complex and smaller change that still addresses #5650.

Issue #5662, opened 2021-02-03
Linkerd 2.9.2 - Proxy takes long time to get certificate from identity service


RajKuni:
## Bug Report

### What is the issue?

In linkerd 2.9.2, the linkerd proxies take a long time to get ready. The CSR a proxy sends to the identity service is fulfilled in minutes rather than in seconds. This causes the actual application container that is running in the pod to fail.

We were running linkerd 2.8 prior to this on the same cluster so I'm assuming some change introduced in 2.9.2 is causing this issue? 

### How can it be reproduced?

Install linkerd 2.9.2 on a GKE cluster (v 1.16.15-gke.6000).

I tried installing linkerd via the helm chart and linkerd cli to see whether that would make any difference but it did not - same slow start-up of proxies.  

### Logs, error output, etc

When the proxy starts up, we can see the default start up logs:

```
time="2021-02-03T23:17:42Z" level=info msg="running version stable-2.9.2"                                                                                                                                          
[     0.001455s]  INFO ThreadId(01) linkerd2_proxy::rt: Using single-threaded proxy runtime                                                                                                                        
[     0.002754s]  INFO ThreadId(01) linkerd2_proxy: Admin interface on 0.0.0.0:4191                                                                                                                                
[     0.002772s]  INFO ThreadId(01) linkerd2_proxy: Inbound interface on 0.0.0.0:4143                                                                                                                              
[     0.002778s]  INFO ThreadId(01) linkerd2_proxy: Outbound interface on 127.0.0.1:4140                                                                                                                           
[     0.002783s]  INFO ThreadId(01) linkerd2_proxy: Tap interface on 0.0.0.0:4190                                                                                                                                  
[     0.002788s]  INFO ThreadId(01) linkerd2_proxy: Local identity is <redacted>.<redacted>.serviceaccount.identity.linkerd.cluster.local                                                                                  
[     0.002800s]  INFO ThreadId(01) linkerd2_proxy: Identity verified via linkerd-identity-headless.linkerd.svc.cluster.local:8080 (linkerd-identity.linkerd.serviceaccount.identity.linkerd.cluster.local)        
[     0.002806s]  INFO ThreadId(01) linkerd2_proxy: Destinations resolved via linkerd-dst-headless.linkerd.svc.cluster.local:8086 (linkerd-destination.linkerd.serviceaccount.identity.linkerd.cluster.local)      
[...]

PR #5707, opened 2021-02-10
Use fully-qualified DNS names in proxy configuration

olix0r:
Pods with unusual DNS configurations may not be able to resolve the
control plane's domain names. We can avoid search path shenanigans by
adding a trailing dot to these names.

Proposed by @adleong to address #5662.


alpeb (approved):
Neat :+1:

Pothulapati (approved):
LGTMM

PR #5713, opened 2021-02-10
Add the 'tapInjector.logLevel' value.

### LINKED ISSUE #5686: viz: logLevel parameter for tap injector
## Feature Request

Currently, there seems to be no way to set the `log-level` parameter
for the tap-injector and it would be nice to have one just like other
components, following the same semantics for overrides i.e
`<component>.logLevel`

alpeb:
Fixes #5686

Test:
```bash
$ linkerd viz install --set tapInjector.logLevel=debug | k apply -f -

// and then when creating a pod we can see debug log entries such as:

time="2021-02-10T16:19:28Z" level=debug msg="admission request: &AdmissionRequest{UID:c5e95e8d-...
```

### DIFF
--- a/viz/charts/linkerd-viz/README.md
+++ b/viz/charts/linkerd-viz/README.md
@@ -164,6 +164,7 @@ Kubernetes: `>=1.13.0-0`
 | tapInjector.image.registry | string | `"ghcr.io/linkerd"` | Docker registry for the tapInjector instance |
 | tapInjector.image.tag | string | `"linkerdVersionValue"` | Docker image tag for the tapInjector instance |
 | tapInjector.keyPEM | string | `""` | Certificate key for the tapInjector. If not provided then Helm will generate one. |
+| tapInjector.logLevel | string | defaultLogLevel | log level of the tapInjector |
 | tapInjector.namespaceSelector | string | `nil` |  |
 | tapInjector.objectSelector | string | `nil` |  |
 | tapInjector.proxy | string | `nil` |  |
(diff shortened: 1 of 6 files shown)


viz/charts/linkerd-viz/README.md
```diff
@@ -164,6 +164,7 @@ Kubernetes: `>=1.13.0-0`
 | tapInjector.image.registry | string | `"ghcr.io/linkerd"` | Docker registry for the tapInjector instance |
 | tapInjector.image.tag | string | `"linkerdVersionValue"` | Docker image tag for the tapInjector instance |
 | tapInjector.keyPEM | string | `""` | Certificate key for the tapInjector. If not provided then Helm will generate one. |
+| tapInjector.logLevel | string | defaultLogLevel | log level of the tapInjector |
```
dadjeibaah, line 167: I wonder if we should change this so that we show the actual default value. In other places where we have default log levels, we use the log level name instead of `defaulLogLevel`. WDYT?
```suggestion
| tapInjector.logLevel | string | `"info"` | log level of the tapInjector |
```

dadjeibaah (approved):
Tested it out and it works! 📦 🚢

viz/charts/linkerd-viz/README.md
alpeb: This file is auto-generated through `bin/helm-docs` which uses the comments on the `values.yaml` file. Anyways,I think it's fine to refer to other fields in the same `values.yaml` file. This goes in line with #5694, after which a change in `defaultLogLevel` will cascade down to all the other log levels, so I think it's important to reflect that in the default value.

dadjeibaah: Ahh, I gotcha. This makes total sense. I didn't realize this file was auto-generated. Thanks for the explanation!

kleimkuhler (approved):
🤙

adleong (approved):

PR #5716, opened 2021-02-10
image-load: Parallelize image loading in k3d

olix0r:
The `image-load` script invokes `k3d image load` once for reach image.
This is unnecessary, as k3d accepts a list of images on the
command-line.

This change updates the image-load script to load all images in a single
command. Anecdotally, this seems much faster.

### DIFF
--- a/bin/image-load
+++ b/bin/image-load
@@ -78,26 +78,33 @@ fi
 # the parallel executions below attempt doing so
 "$bin" version
 
-rm -f load_fail
-for img in proxy controller web metrics-api grafana cli-bin debug cni-plugin jaeger-webhook tap; do
-  printf 'Importing %s...\n' $img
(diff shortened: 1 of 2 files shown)


kleimkuhler (approved):
Nice!

adleong (approved):
In my anecdotal testing, this doesn't yield any major speed benefit (actually a slight slow down, about 3.5m -> 4m) but anyway, it seems cleaner to just pass all the images at once.

olix0r:
well, this somehow can't pass CI. so. never mind.

alpeb:
We didn't do parallel (at the shell level) image loading with k3d like we did for kind because it somehow created a conflict in k3d. Maybe the same problem arises when done via a single k3 command :man_shrugging:

PR #5721, opened 2021-02-11
CLI: add `--opaque-ports` flag to `inject`

### LINKED ISSUE #5703: Set the opaque-port config setting via cmd
When injecting linkerd into manifests that use the SSF protocol it would be nice to be able to do so via am cli flag 
just like 
```bash
linkerd inject --skip-inbound-ports=3306 --skip-outbound-ports=3306 -
```

for this setting

Via pipelines the linkerd properties would be set in the manifests already so this is not crucial  feature
You can always edit the manifest manually also after linkerd inkect and add this 
```yaml
annotations:
  config.linkerd.io/opaque-ports: 3306
```

it would be great to be able to run something similar to  
```bash
linkerd- inject --opaque-ports=3306  -
```

mayankshah1607:
Fixes #5703 

The `config.linkerd.io/opaque-ports` annotation can now be set using the `--opaque-ports` flag on `inject`

Example

```bash
$ linkerd inject /path/to/manifest.yaml --opaque-ports 3000,5000-6000,mysql
```

To do:
- [x] Add unit tests (I could do that on this PR or on a separate PR to keep this one small)


olix0r:
I haven't looked at the code change, but we should ensure that this annotation is also added to `Service` resources. Per #5185

mayankshah1607:
> @mayankshah1607 If you could add unit tests for this to the PR that'd be great — better to keep changes and tests for them together than separate.

Done, PTAL! @kleimkuhler

kleimkuhler:
@mayankshah1607 From a brief review this looks like it's on the right path. Thanks for adding the test!

As Oliver mentioned above, when the `--opaque-ports` flag is passed services will also need to get the annotation. Currently, the `inject` command does not do anything with service kind, so that will need to be changed as part of this.

It's going to make this change a little bigger, but for starters, I'd take a look at the [parse](https://github.com/linkerd/linkerd2/blob/a393c42536d569e5a9becc3b1dd8227063fba478/pkg/inject/inject.go#L343-L506) function. I got there by looking at `runInjectCmd` -> `transformInput` -> `processYAML` -> `rt.transform` -> `ParseMetaAndYAML` -> `conf.parse`.

We'll need to handle services now and prep them for adding the annotation.

Does this make sense? If you have any questions please don't hesitate to ask!

mayankshah1607:
@olix0r @kleimkuhler I have added support for `Service` in `inject`. Through this PR, a `Service` may only have annotations appended via `inject`, specifically the ones present in [`resourceTransformerInject.overrideAnnotations`](https://github.com/linkerd/linkerd2/blob/main/cli/cmd/inject.go#L42) which are set from the flags; i.e, the proxy inject annotation cannot be added on a Service using `inject`. Do we want to restrict `inject` to only add the `config.linkerd.io/opaque-ports` in such cases?

I've also modified the inject message to use `"annotated'` instead of `"injected"` while dealing with services 😉 

```bash
❯ bin/go-run cli inject svc.yaml --ignore-cluster --opaque-ports=123
apiVersion: v1
kind: Service
metadata:
  annotations:
    config.linkerd.io/opaque-ports: "123"
  name: my-service
spec:
  ports:
  - port: 80
    protocol: TCP
    targetPort: 9376
  selector:
    app: MyApp
---

service "my-service" annotated
```

cli/cmd/inject.go
```diff
@@ -319,7 +325,11 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output
 
 	for _, r := range reports {
 		if b, _ := r.Injectable(); b {
-			output.Write([]byte(fmt.Sprintf("%s \"%s\" injected\n", r.Kind, r.Name)))
+			verb := "injected"
+			if r.Kind == k8s.Service {
```
adleong, line 329: I think using the "injected" verb is fine for services too, even though we only add annotations.  For example, we already use the verb "injected" for namespaces (which also only get an annotation)

cli/cmd/inject.go
```diff
@@ -170,6 +170,11 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 
 	reports := []inject.Report{*report}
 
+	if conf.IsService() {
+		b, err := conf.AnnotateService(rt.overrideAnnotations)
```
adleong, line 174: I think we want to limit this to only set annotations which are applicable to services.  For now, this is only the opaque ports annotation.  Otherwise, calls like `linkerd inject --proxy-log-level=debug` can set annotations on the service which do not apply to service resources.

adleong (commented):
Looks pretty good, @mayankshah1607!  I left a couple of comments.

cli/cmd/uninject.go
```diff
@@ -78,6 +78,7 @@ func (rt resourceTransformerUninject) transform(bytes []byte) ([]byte, []inject.
 		output = bytes
 		report.UnsupportedResource = true
 	}
+
```
kleimkuhler, line 81: This can be removed

cli/cmd/inject.go
```diff
@@ -170,6 +170,11 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 
 	reports := []inject.Report{*report}
 
+	if conf.IsService() && rt.values.Proxy.OpaquePorts != "" {
+		b, err := conf.AnnotateService(k8s.ProxyOpaquePortsAnnotation, rt.values.Proxy.OpaquePorts)
+		return b, reports, err
+	}
```
kleimkuhler, line 176: This will need to be changed to deal with a few scenarios.

First, if we're injecting a service that does not have `rt.values.Proxy.OpaquePorts` set, then we don't branch into this and continue try and get a patch for a pod. This results in an error:

```
Error transforming resources:
add operation does not apply: doc is missing path: "/spec/template/spec/volumes/-": missing value
```

To deal with this, we should already plan to return when dealing with a service:

```suggestion
	if conf.IsService() {
		b := bytes
		if rt.values.Proxy.OpaquePorts != "" {
			b, err = conf.AnnotateService(k8s.ProxyOpaquePortsAnnotation, rt.values.Proxy.OpaquePorts)
		}
		return b, reports, err
```

Second, we can return a more helpful error message here so that we know that annotating the service failed. I'm thinking something like:

```
if err != nil {
	return nil, nil, fmt.Errorf("failed to annotate service %s: %s", report.Name, err)
}
```

Finally, this occurs before the `report.Injectable()` check below. If the service has `linker.io/inject: disabled`, we'll still annotate the service which we really shouldn't do. If inject is disabled on the service, then we should not mutate it at all.

This part is going to require changing what `func Injectable` checks. If we're dealing with a service object, we'll need to check the annotations on the object, where right now we only check annotations on a pod.

kleimkuhler (commented):
Sorry it took me a bit to look at this @mayankshah1607! I left a comment that brings up a few things that will need to be addressed. Let me know if you have any questions about that.

kleimkuhler:
@mayankshah1607 Just a heads up that with #5766 merged, there's a few things in here that were included in that change as well. It may be worth taking a look at that and making sure you understand that change as well.

cli/cmd/inject.go
kleimkuhler: Note that with #5766 merging, the `IsService` happens before the `Injectable` call ([code](https://github.com/linkerd/linkerd2/blob/edd3812f3050895fae1cc393d9d3365c405d6e21/cli/cmd/inject.go#L174-L176)).

This is okay for that change, because we never mutate the service YAML with the `inject` command. However for you change, this ordering does matter because we could be mutating the service. So we'll still want to move it below `Injectable` for this.

mayankshah1607:
Thanks for the review @kleimkuhler @adleong, and apologies for the delayed response. I have updated this PR with all the required changes. PTAL, thanks! :)

cli/cmd/testdata/inject_emojivoto_service.input.yml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: emojivoto
+spec:
+  type: ClusterIP
+  selector:
+    app: web-svc
+  ports:
+  - name: http
+    port: 80
+    targetPort: 8080
+---
```
kleimkuhler, line 14: Can you add a newline here?

pkg/inject/inject.go
```diff
@@ -942,6 +942,23 @@ func (conf *ResourceConfig) InjectNamespace(annotations map[string]string) ([]by
 	return yaml.JSONToYAML(j)
 }
 
+// AnnotateService returns a Service adds an annotation to a Service object
+// Currently, a `Service` may only need the `config.linkerd.io/opaque-ports` annotation via `inject`
+// See - https://github.com/linkerd/linkerd2/pull/5721
```
kleimkuhler, line 947: Typo in here:

```suggestion
// AnnotateService adds the given annotation to a service.
// Currently, a `Service` may only need the `config.linkerd.io/opaque-ports` annotation via `inject`
// See - https://github.com/linkerd/linkerd2/pull/5721
```

kleimkuhler (commented):
This works well! A few scenarios I tested:

1. Pass the flag and inject a service
2. Don't pass the flag and inject a service
3. Pass the flag and inject a service with `linkerd.io/inject: disabled`: In this case, we still annotate the service which is correct. Even though we may not be injecting workloads, those workloads can still be opaque protocols and we want the service to have the annotation.

We still print `service "svc" skipped` and I think it would be helpful to change that. Even though we are skipping injection, we are mutating the service object so it may be nice to at least print `service "svc" annotated`. We should still print `skipped` though when the flag is not passed.

alpeb:
#5851 wrapped this one up, in order to include it in edge-21.3.1. Thanks again @mayankshah1607 !

PR #5724, opened 2021-02-11
 Fix for CLI printing command usage for API Errors for multiple commands

piyushsingariya:

Related to #5058 

Problem: CLI prints command usage for multiple commands in-case of API errors
Solution: Print the error and the exit using os.Exit(1) to avoid Cobra printing usage


adleong (approved):
Thanks, @piyushsingariya!

alpeb (commented):
LGTM :+1: 

The integration test error for `upgrade-edge` is happening because it's upgrading from `edge-21.2.2` which had the linkerd-config data structure changed, and your source still doesn't have those changes. Merging `master` here should solve the problem.

Issue #5659, opened 2021-02-03
viz: stat command panics when extension prometheus is disabled


kleimkuhler:
Install the viz extension and bring your own Prometheus:

```
linkerd viz install --set prometheusUrl="http://prometheus.default.svc.cluster.local:9090",prometheus.enabled=false |kubectl apply -f -
```

Run the `stat` command:

```
❯ linkerd viz stat -n linkerd-viz deploy
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x17830e6]

goroutine 28 [running]:
github.com/linkerd/linkerd2/viz/cmd.requestStatsFromAPI(0x0, 0x0, 0xc000499c20, 0x0, 0xc0003063c0, 0x0)
        /linkerd-build/viz/cmd/stat.go:244 +0x26
github.com/linkerd/linkerd2/viz/cmd.NewCmdStat.func1.1(0x0, 0x0, 0xc00004c660, 0x0, 0xc000499c20)
        /linkerd-build/viz/cmd/stat.go:195 +0x4c
created by github.com/linkerd/linkerd2/viz/cmd.NewCmdStat.func1
        /linkerd-build/viz/cmd/stat.go:194 +0x2c0
```

This error does not occur if `prometheus.enabled=false` is not specified. `prometheusUrl` can be set and the `stat` command will still work.

The dashboard still has metrics so is likely an issue only with the stat command.

For reproducing, this is the Prometheus that I deploy in the default namespace:

```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: prometheus
rules:
- apiGroups: [""]
  resources: ["nodes", "nodes/proxy", "pods"]
  verbs: ["get", "list", "watch"]
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: prometheus
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: prometheus
subjects:
- kind: ServiceAccount
  name: prometheus
  namespace: default
---
kind: ServiceAccount
apiVersion: v1
metadata:
  name: prometheus
  namespace: default
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: prometheus-config
  namespace: default
data:
  prometheus.yml: |-
    global:
      evaluation_interval: 10s
      scrape_interval: 10s
[...]

PR #5731, opened 2021-02-11
Add changes for edge-21.2.2

kleimkuhler:
## edge-21.2.2

This edge release introduces support for multicluster TCP!

The `repair` command was added which will repopulate resources needed for
upgrading from a `2.9.x` installation. There will be an error message during the
upgrade process indicating that this command should be run so that users do not
need to guess.

Lastly, it contains a breaking change for Helm users. The `global` field has
been removed from the Helm chart now that it is no longer needed. Users will
need to pass in the identity certificates again—along with any other
customizations, no longer rooted at `global`.

* **Breaking change**: Removed the `Global` field from the Linkerd Helm chart
  now that it is unused because of the extension model
* Added the `repair` command which will repopulate resources needed for properly
  upgrading a Linkerd installation
* Fixed the spelling of the `sidecarContainers` key in the Viz extension Helm
  chart to match that of the template (thanks @n-oden!)
* Added the `tapInjector.logLevel` key to the Viz extension helm chart so that
  the log level of the component can be configured
* Removed the `--disable-tap` flag from the `inject` command now that tap is no
  longer part of the core installation (thanks @mayankshah1607!)
* Changed proxy configuration to use fully-qualified DNS names to avoid extra
  search paths in DNS resolutions
* Changed the `check` command to include each installed extension's `check`
  output; this allows users to check for proper configuration and installation
  of Linkerd without running a command for each extension
* Added proxy support for TCP traffic to the multicluster gateways

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## edge-21.2.2
 
+* **Breaking change**: Removed the `Global` field from the Linkerd Helm chart
+  now that it is unused because of the extension model; users doing upgrades
+  with Helm passing customized `values.yaml` will have to take this change into
+  account.
 * Added the `repair` command which will repopulate resources needed for properly
   upgrading a Linkerd installation
 * Fixed the spelling of the `sidecarContainers` key in the Viz extension Helm
@@ -15,9 +19,6 @@
 * Changed the `check` command to include each installed extension's `check`
   output; this allows users to check for proper configuration and installation
   of Linkerd without running a command for each extension
-* Removed the `Global` field from the Linkerd Helm chart now that it is unused
-  because of the extension model; this is not a breaking change and cleanup will
-  occur automatically during upgrade.
 * Added proxy support for TCP traffic to the multicluster gateways which
   completes preparation for supporting opaque traffic in multicluster
   installations; the next edge will support this end-to-end
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 ## edge-21.2.2
 
+This edge release continues to prepare support for multicluster opaque ports.
+
+Additionally, it contains a breaking change for Helm users upgrading with a
+customized `values.yaml`. If upgrading without this customization, there is no
+action to take.
+
 * **Breaking change**: Removed the `Global` field from the Linkerd Helm chart
   now that it is unused because of the extension model; users doing upgrades
   with Helm passing customized `values.yaml` will have to take this change into
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,11 +2,16 @@
 
 ## edge-21.2.2
 
-This edge release continues to prepare support for multicluster opaque ports.
+This edge release introduces a new `repair` command which will repopulate
+resources needed for upgrading from a `2.9.x` installation. There will be an
+error message during the upgrade process indicating that this command should be
+run so that users do not need to guess.
 
-Additionally, it contains a breaking change for Helm users upgrading with a
-customized `values.yaml`. If upgrading without this customization, there is no
-action to take.
+It also continues to prepare support for multicluster opaque ports.
+
+Lastly, it contains a breaking change for Helm users upgrading with a customized
+`values.yaml`. If upgrading without this customization, there is no action to
+take.
 
 * **Breaking change**: Removed the `Global` field from the Linkerd Helm chart
   now that it is unused because of the extension model; users doing upgrades
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,27 @@
 # Changes
 
+## edge-21.2.2
+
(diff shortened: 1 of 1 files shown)


kleimkuhler:
Note I did not include #5627. There is still an issue that leaves the Viz API client uninitialized which results in issues like #5659

CHANGES.md
```diff
@@ -1,5 +1,27 @@
 # Changes
 
+## edge-21.2.2
+
```
adleong, line 4: Most other edge releases have a short blurb here summarizing the most important changes at a high level.

kleimkuhler: I think a lot of the recent one have had one because they've introduced some significant changes with all the extension work, but this edge did not feel that way to me.

I can include something about the `repair` command? All the other bullets felt pretty independent and there was no major behavior change.

CHANGES.md
```diff
@@ -1,5 +1,27 @@
...
+* Removed the `--disable-tap` flag from the `inject` command now that tap is no
+  longer part of the core installation (thanks @mayankshah1607!)
+* Changed proxy configuration to use fully-qualified DNS names to avoid extra
+  search paths in DNS resolutions
+* Changed the `check` command to include each installed extension's `check`
+  output; this allows users to check for proper configuration and installation
+  of Linkerd without running a command for each extension
+* Removed the `Global` field from the Linkerd Helm chart now that it is unused
+  because of the extension model; this is not a breaking change and cleanup will
+  occur automatically during upgrade.
```
alpeb, line 20: People doing upgrades with helm passing customized values.yaml will have to take this change into account. I think this grants a mention at the top as a breaking change.

CHANGES.md
olix0r: I think we should call out TCP multicluster support up top. It's a big feature!

CHANGES.md
kleimkuhler: Moved to the top!

CHANGES.md
kleimkuhler: I kept it out because the service mirror change hasn't merged. I mentioned at the bottom though that we are preparing to support it.

kleimkuhler: Separate from opaque ports the TCP support is there so I see what you're saying. I can include that.

CHANGES.md
```diff
@@ -1,5 +1,39 @@
 # Changes
 
+## edge-21.2.2
+
+This edge release introduces a new `repair` command which will repopulate
+resources needed for upgrading from a `2.9.x` installation. There will be an
+error message during the upgrade process indicating that this command should be
+run so that users do not need to guess.
+
+It also continues to prepare support for multicluster opaque ports.
```
olix0r: We're underselling this. TCP multicluster support is a huge deal. That it doesn't yet work with opaque ports is a small implementation detail that we can comment on in the next release. Prior to this release, multicluster only supported HTTP. Now it can work with any protocol!!! Even without an opaque-ports annotation. The opaque-ports stuff just makes it _better_. This is the top-line feature in this edge release.

olix0r (approved):

CHANGES.md
```diff
@@ -1,5 +1,40 @@
...
+need to guess.
+
+Lastly, it contains a breaking change for Helm users upgrading with a customized
+`values.yaml`. If upgrading without this customization, there is no action to
+take.
+
+* **Breaking change**: Removed the `Global` field from the Linkerd Helm chart
+  now that it is unused because of the extension model; users doing upgrades
+  with Helm passing customized `values.yaml` will have to take this change into
+  account.
```
alpeb, line 20: Actually, helm installations are _always_ customized, because they need to include the issuer certs. So people will be forced to pass-in the certs again, no longer rooted at `global` (alongside any other customizations they had before, using the new structure).
Also tiny nit: `global` is all lower-cased.

CHANGES.md
kleimkuhler: Yep that's my bad. I was only thinking about this with opaque ports in mind. I included this at the top and focused more on the TCP support.

CHANGES.md
alpeb: When I mentioned issuer certs above, I should have said instead `identityTrustAnchorsPEM` which is the mandatory config that was formerly under `global`.

kleimkuhler: Okay I explained this up top.

alpeb: I'm thinking the verbiage could be something like this:

> **Breaking change for Helm**: The `values.yaml` entries for the linkerd2 Helm chart had several entries under the `global` scope, which now have been moved into the root scope. You need to pass again to `helm upgrade` any value that you have customized, using the new structure. At the very least, you'll need to provide the same `identityTrustAnchorsPEM` entry you previously used, which is always required and was moved into the root scope.

adleong (approved):

PR #5732, opened 2021-02-11
Trigger heartbeat 10 mins after install

alpeb:
Restores how this was working before it was unintentionally changed to run at midnight in 2.8

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -306,6 +306,9 @@ func install(ctx context.Context, w io.Writer, values *l5dcharts.Values, flags [
 		return err
 	}
 
+	t := time.Now().Add(10 * time.Minute).UTC()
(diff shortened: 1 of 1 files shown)


adleong (approved):

adleong:
Couple of things to keep in mind:

* This doesn't affect Helm.  Helm installs will still have the default heartbeat schedule
* This overrides the user input, so it is impossible for a CLI user to override the heartbeat schedule

alpeb:
> * This overrides the user input, so it is impossible for a CLI user to override the heartbeat schedule

Actually `install --set heartbeatSchedule="1 2 3 4"` does override that config, given the merge of helm-style flags (through `options.MergeValues(nil)`) happens aftewards

kleimkuhler (approved):
⏱

PR #5738, opened 2021-02-12
Add --skip-cluster-delete flag to bin/tests

alpeb:
... to avoid tearing down resources and the cluster itself (or clusters) after a given test finishes. This allows for example when running the multicluster test, leaving both clusters around for further inspection:
```bash
bin/tests --name multicluster --skip-cluster-delete `pwd`/target/cli/linux-amd64/linkerd
```

### DIFF
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -36,8 +36,9 @@ Examples:
 
 Available Commands:
     --name: the argument to this option is the specific test to run
-    --skip-cluster-create: skip k3d cluster creation step and run tests in an existing cluster.
-    --images: by default load images into the cluster from the local docker cache (docker), or from tar files located under the 'image-archives' directory (archive), or completely skip image loading (skip)."
+    --skip-cluster-create: skip k3d cluster creation step and run tests in an existing cluster
+    --skip-cluster-delete: if the tests succeed, don't delete the created resources nor the cluster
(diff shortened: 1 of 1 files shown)


bin/_test-helpers.sh
```diff
@@ -120,6 +126,12 @@ handle_tests_input() {
     tests_usage "$0" >&2
     exit 64
   fi
+
+  if [ -z "$test_name" ] && [ -n "$skip_cluster_delete" ]; then
+    echo "Error: must provide --name when using --skip-cluster-delete" >&2
+    tests_usage "$0" >&2
+    exit 64
+  fi
```
kleimkuhler, line 134: Good call

kleimkuhler (approved):

bin/_test-helpers.sh
kleimkuhler: Actually do we want to enforce this? We don't for `skip_cluster_create`, and we don't pass `--name` in when using it for ARM64 tests. Could be nice to have the whole cluster stick around if running all tests.

alpeb: When all the tests are run, the beginning of each test expects a clean slate otherwise it'll fail. I could make it so `--skip-cluster-delete` does clean the cluster after each test except for the last one, if you think that'd be valuable.

adleong (approved):

bin/_test-helpers.sh
kleimkuhler: Nope I think this is better then!

Issue #5736, opened 2021-02-12
proxy-injector: pods should inherit opaque ports annotation from namespaces


kleimkuhler:
During proxy injection, pods that do not have an opaque ports annotation should
inherit from their namespace if the annotation is present on the namespace.

This inheritance will happen only at injection time meaning annotation updates
to a namespace will not be updated down through to a pod. Pods will need to be
restarted in order to pick up updates from changes to the namespace.

This will remove a lot of complexity around the destination service contract for
updating destination clients about changes to opaque ports.

kleimkuhler:
Depends on #5735

PR #5739, opened 2021-02-12
Remove namespace inheritance of opaque ports annotation

### LINKED ISSUE #5735: destination: Remove opaque port annotation inheritance from namespaces
There is inconsistent behavior with resources inheriting the opaque ports
annotation from namespaces.

For example, the endpoints watcher will recalculate the inheritance for each
endpoints update, but the pod IP path of `GetProfile` is only calculated once on
the initial call—no further updates are sent. This can lead to a difference in
`OpaqueTransport` and `OpaqueProtcol` fields.

Inheritance from namespace should be completely removed in preparation for the
proxy injector merging these annotations during injection.

kleimkuhler:
This change removes the namespace inheritance of the opaque ports annotation.
Now when setting opaque port related fields in destination profile responses, we
only look at the pod annotations.

This prepares for #5736 where the proxy-injector will add the annotation from
the namespace if the pod does not have it already.

Closes #5735

### DIFF
--- a/controller/api/destination/server.go
+++ b/controller/api/destination/server.go
@@ -210,7 +210,6 @@ func (s *server) GetProfile(dest *pb.GetDestination, stream pb.Destination_GetPr
 					Namespace: pod.Namespace,
 					Name:      pod.Name,
 				}
-				err := watcher.SetPodOpaquePortAnnotation(s.k8sAPI, pod, pod.Namespace)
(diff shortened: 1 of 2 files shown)


adleong (approved):
👪 🚫💰

alpeb (approved):
LGTM :+1:

PR #5744, opened 2021-02-13
Multicluster: Uninstall multicluster without Linkerd control plane

### LINKED ISSUE #5624: multicluster extension cannot be uninstalled if Linkerd is not installed
If the main Linkerd control plane has been uninstalled, it is no longer possible to uninstall the multicluster extension.  

Uninstalling the multicluster extension should not required the main control plane.

```
bin/linkerd mc uninstall | k delete -f -
Error: you need Linkerd to be installed in order to install multicluster addons
Usage:
  linkerd multicluster uninstall [flags]
```

piyushsingariya:

Problem
If the main Linkerd control plane has been uninstalled, it is no longer possible to uninstall the multicluster extension.
```
$ bin/linkerd mc uninstall | k delete -f -
Error: you need Linkerd to be installed in order to install multicluster addons
Usage:
  linkerd multicluster uninstall [flags]
```
Solution
Fetch resources with the label `linkerd.io/extension: linkerd-multicluster` and delete them

Fixes #5624

### DIFF
--- a/multicluster/charts/linkerd-multicluster/templates/gateway.yaml
+++ b/multicluster/charts/linkerd-multicluster/templates/gateway.yaml
@@ -6,6 +6,7 @@ metadata:
   name: {{.Values.gatewayName}}-config
   labels:
     {{.Values.controllerComponentLabel}}: gateway
+    linkerd.io/extension: linkerd-multicluster
   annotations:
     {{.Values.createdByAnnotation}}: {{default (printf "linkerd/helm %s" .Values.linkerdVersion) .Values.cliVersion}}
   namespace: {{.Values.namespace}}
@@ -58,6 +59,7 @@ metadata:
     app.kubernetes.io/version: {{.Values.linkerdVersion}}
     {{.Values.controllerComponentLabel}}: gateway
     app: {{.Values.gatewayName}}
+    linkerd.io/extension: linkerd-multicluster
   name: {{.Values.gatewayName}}
   namespace: {{.Values.namespace}}
 spec:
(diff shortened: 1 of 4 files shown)


piyushsingariya:
These are the resources being deleted till now.

- [ ] namespace "linkerd-multicluster" deleted
- [ ] configmap "linkerd-gateway-config" deleted
- [ ] deployment.apps "linkerd-gateway" deleted
- [ ] service "linkerd-gateway" deleted
- [ ] serviceaccount "linkerd-gateway" deleted
- [ ] clusterrole.rbac.authorization.k8s.io "linkerd-service-mirror-remote-access-default" deleted
- [ ] serviceaccount "linkerd-service-mirror-remote-access-default" deleted
- [ ] clusterrolebinding.rbac.authorization.k8s.io "linkerd-service-mirror-remote-access-default" deleted
- [ ] customresourcedefinition.apiextensions.k8s.io "links.multicluster.linkerd.io" deleted

alpeb (commented):
`FetchKubernetesResources` returns the cluster-scoped resources with the given label, which includes the namespace. So deleting all that, including the namespace, will also get rid of the namespace-scoped resources :+1:

piyushsingariya:
This means the resources that are not being mentioned while deletion, are actually being deleted??
> `FetchKubernetesResources` returns the cluster-scoped resources with the given label, which includes the namespace. So deleting all that, including the namespace, will also get rid of the namespace-scoped resources +1

alpeb:
> This means the resources that are not being mentioned while deletion, are actually being deleted?? 
> 
> > `FetchKubernetesResources` returns the cluster-scoped resources with the given label, which includes the namespace. So deleting all that, including the namespace, will also get rid of the namespace-scoped resources +1

Correct.

adleong (approved):
Awesome!

Just something to note here: people who installed the multicluster extension at a previous version will not have this label and therefore `linkerd multicluster uninstall` will not work.  They would have to first upgrade the multicluster extension by running `linkerd multicluster install | kubectl apply -f -` to get the label, and then they would be able to run `linkerd multicluster uninstall | kubectl delete -f -`.

This can leave them in a particularly sticky situation if they don't have a working LInkerd control plane since they won't be able to upgrade, and therefore they won't be able to uninstall.  I'm not sure there's much we can easily do about this situation, however.

kleimkuhler (approved):
This works well @piyushsingariya, thanks!

PR #5745, opened 2021-02-14
viz: Upgrade material-ui version for Chrome 88 fix

### LINKED ISSUE #5612: Linkerd dashboard does not load in versions of Chrome 88+
### What is the issue?
The Linkerd dashboard does not load and displays the error message below
Screen Shot 2021-01-26 at 11 36 29 AM
It looks like this issue occurs in versions of Chrome 88 and higher. Upon further investigation, it looks like material-ui has a bug that that is only present in this version of Chrome. More details here: https://github.com/mui-org/material-ui/issues/24549 

### How can it be reproduced?
- Make sure you are using Chrome version 88 or higher
- Check out the main linkerd2 branch
- Run `bin/web dev`
- navigate to http://localhost:7777
### Logs, error output, etc

Error message is in the screenshot above. 

### Possible solution

Material-UI released https://github.com/mui-org/material-ui/releases/tag/v4.11.3 which directly fixes this issue. Linkerd's dashboard uses material-ui https://github.com/mui-org/material-ui/releases/tag/v4.9.11. We could upgrade to the latest version of Material UI to fix this issue but we should do some UI testing to make sure things work as expected between the minor upgrade.

kellycampbell:
Dashboard UI was blank when loading on Chrome 88 with an error in the console.

This upgrades material-ui from v4.9.11 to v4.11.3.

Fixes #5612

Tested on a cluster running edge-21.2.2. UI loads properly now.

Signed-off-by: Kelly Campbell <kelly.a.campbell@gmail.com>

### DIFF
--- a/web/app/package.json
+++ b/web/app/package.json
@@ -10,7 +10,7 @@
     "@fortawesome/react-fontawesome": "0.1.9",
     "@lingui/macro": "2.9.2",
     "@lingui/react": "2.9.2",
-    "@material-ui/core": "4.9.11",
+    "@material-ui/core": "4.11.3",
     "@material-ui/icons": "4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.50",
     "classnames": "2.2.6",
--- a/web/app/yarn.lock
+++ b/web/app/yarn.lock
@@ -1288,24 +1288,23 @@
     hoist-non-react-statics "3.3.0"
     prop-types "^15.7.2"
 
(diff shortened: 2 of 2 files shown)


dadjeibaah (approved):
Thanks for fixing this! Tested it out and it works for Chrome! 🙌 📦 🚢

Pothulapati (approved):
LGTM, The dashboard works as expected...

adleong (approved):

PR #5747, opened 2021-02-15
Drop special review requirements for CNI

olix0r:
We initially set special review requirements for the CNI; but in
practice this hasn't been necessary and has slowed down many reviews,
especially documentation changes.

This change removes special review requirements in favor of using the
maintainers group.


alpeb (approved):

Pothulapati (approved):
LGTM

PR #5750, opened 2021-02-15
Update debian base images to buster-20210208-slim

olix0r:
Before the upcoming stable release, we should update our base images to
use the most recent Debian images to pick up any security fixes that may
have been addressed.

This change updates all o four debian images to use the
`buster-20210208-slim` tag.


alpeb (approved):
LGTM :+1:

Pothulapati (approved):
LGTM

PR #5756, opened 2021-02-16
docker: Access container images via cr.l5d.io

olix0r:
We've created a custom domain, `cr.l5d.io` that redirects to `ghcr.io`
(using https://scarf.sh). This custom domain allows us to swap the
underlying container registry without impacting users. It also provides
us with important metrics about container usage, without surfacing PII
like IP addresses.

This change updates our Helm charts and CLIs to reference this custom
domain. The integratoin test workflow has been updated to tag images
appropriately. The release process has *not* been updated, as we can
continue to use the `ghcr.io/linkerd` registry for the purpose of
publishing images.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -29,7 +29,7 @@ jobs:
         echo "TAG=$(CI_FORCE_CLEAN=1 bin/root-tag)" >> $GITHUB_ENV
 
         . bin/_docker.sh
-        echo "DOCKER_REGISTRY=$DOCKER_REGISTRY" >> $GITHUB_ENV
+        echo "DOCKER_REGISTRY=cr.l5d.io/linkerd" >> $GITHUB_ENV
         echo "DOCKER_BUILDKIT_CACHE=${{ runner.temp }}/.buildx-cache" >> $GITHUB_ENV
     - name: Cache docker layers
       # actions/cache@v2.0.0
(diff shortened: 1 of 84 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -40,11 +40,11 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-buildx-${{ matrix.target }}-
     - name: Build docker images
-      #env:
-      #  DOCKER_TRACE: 1
       run: |
         docker buildx create --driver docker-container --use
         bin/docker-build-${{ matrix.target }}
+        # Ensure all of the images are tagged under cr.l5d.io so pulls work
+        DOCKER_REGISTRY=cr.l5d.io bin/docker-build-${{ matrix.target }}
```
alpeb, line 47: This and the change below can be discarded now that `DOCKER_REGISTRY` defaults to `cr.l5d.io`

olix0r: The GitHub secret value is unchanged, since we still publish directly to ghcr...

.github/workflows/integration_tests.yml
```diff
@@ -54,6 +52,7 @@ jobs:
         ARCHIVES: /home/runner/archives
       run: |
         mkdir -p $ARCHIVES
+        # Both tags reference the same image. Save both so they may be used interchangeably
```
alpeb, line 55:
```suggestion
```

.github/workflows/integration_tests.yml
alpeb: I think you wanna override `DOCKER_REGISTRY` to `ghcr.io/linkerd` in  `release.yml` which is where we actually push images.

alpeb (approved):
It works :+1:  But I still think the changes in `integration_tests.yml` are not needed because `bin/_docker.sh` sets `DOCKER_REGISTRY=cr.l5d.io/linkerd` at the beginning, which is then used in all the file.

kleimkuhler (approved):
🤙

PR #5660, opened 2021-02-03
Viz templates cleanup: rename .Values.extensionAnnotation to .Values.extensionLabel

alpeb:

PR #5758, opened 2021-02-16
Removed "do-not-edit" entries from values.yaml files

### LINKED ISSUE #5574: helm: Replace configurable annotation names with include
The Linkerd chart and the extension charts contain values for configuring the name of certain annotations such as `createdByAnnotation`.  In practice, the name of these annotations should never be changed.  These Helm variables are intended to simply be used as constants.

Instead, replace these variables with `include` statements which allows us to define these constants once but also does not expose them as user modifiable.

alpeb:
Fixes #5574 and supersedes #5660

- Removed from all the `values.yaml` files all those "do not edit" entries for annotation/label names, ~~moving those to subtemplate functions in the `partials` chart~~ hard-coding them in the templates instead.
- The `values.go` files got simplified as a result.
- The `created-by` annotation was also refactored into a reusable partial. This means we had to add a `partials` dependency to multicluster.


adleong (commented):
Thanks for doing this, @alpeb.

Now that I see the change, I wonder if it's even worth it to have these as centrally defined constants at all, as opposed to just inlining them.  

With code, having named constants can be very valuable because by naming the value, you're providing context to its meaning and purpose.  In this case, these constants feel like a layer of indirection that obscures more than they illuminate.  It's also confusing why some annotations are hidden behind a constant while others are not.

It's true that this gives us a central place where we can change these annotation keys in one place and have them take effect everywhere, but I'm not sure how likely it is that we'd be in a situation where we want to change just the annotation key name without needing to make any other structural changes that would require updating all of the templates.

What do you think?

alpeb:
Yep, the end result was also tickling me a bit, so thanks for the feedback. I think it'd be best to inline those annotations as you suggest. And then as I mentioned, down the line, we can revise all these metadata and see if there's room for unifying things and in that case extract entire sections into partials.

OTOH, I think it's worth keeping `partials.annotations.created-by` as that's really a function that was getting repeated all over the place. And the change to `linkerd.pod-affinity` that simplifies things a bit.

alpeb:
I've just updated this, hard-coding the annotations/labels, but leaving the created-by function

olix0r (approved):
lgtm... i think ;)

charts/linkerd2/values.yaml
```diff
@@ -131,19 +131,6 @@ proxyInit:
   xtMountPath:
     mountPath: /run
     name: linkerd-proxy-init-xtables-lock
-# -- Annotation label for the proxy create. Do not edit.
```
adleong, line 134: 😻

adleong (approved):
:100:

PR #5761, opened 2021-02-16
Report extension usage

### LINKED ISSUE #5516: heartbeat: Report extension usage
As we introduce Linkerd extensions, we should update the heartbeat reporter to report which extensions are enabled. This can be determined by reading namespaces that match the linkerd extension label.

This data is useful to understand whether the majority of clusters update to use the `linkerd-viz` extension (or whether this is viewed as purely optional), and how prevalent the `multicluster` and `jaeger` extensions are in practice.

dadjeibaah:
This change modifies `linkerd-heartbeat` to report on all linkerd
extensions being used in a cluster.

The heartbeat URL now includes extension names and a value of 
`"1"` if an extension is installed. 
```
https://versioncheck.linkerd.io/version.json?install-time=1613518301&k8s-version=v1.20.2&linkerd-jaeger=1&linkerd-multicluster=1&linkerd-viz=1
```

Fixes #5516

### DIFF
--- a/charts/linkerd2/templates/heartbeat-rbac.yaml
+++ b/charts/linkerd2/templates/heartbeat-rbac.yaml
@@ -32,6 +32,32 @@ subjects:
   name: linkerd-heartbeat
   namespace: {{.Values.namespace}}
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: linkerd-heartbeat
+  labels:
+    {{.Values.controllerNamespaceLabel}}: {{.Values.namespace}}
+rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: linkerd-heartbeat
+  labels:
+    {{.Values.controllerNamespaceLabel}}: {{.Values.namespace}}
+roleRef:
+  kind: ClusterRole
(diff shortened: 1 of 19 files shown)


alpeb:
> https://versioncheck.linkerd.io/version.json?install-time=1613518301&k8s-version=v1.20.2&linkerd-jaeger=1&linkerd-multicluster=1&linkerd-viz=1

Considering the case of third-party extensions we don't necessarily know about, we'd have something like
 ```
https://versioncheck.linkerd.io/version.json?install-time=1613518301&k8s-version=v1.20.2&linkerd-jaeger=1&linkerd-multicluster=1&linkerd-viz=1&anotherextension=1
```
The only way the server side can figure that extension would by discarding all the parameters it doesn't know about, which can be a little fragile.
What about sending a list of extensions in a single parameter instead?
```
https://versioncheck.linkerd.io/version.json?install-time=1613518301&k8s-version=v1.20.2&extensions=linkerd-jaeger,linkerd-multicluster,linkerd-viz,anotherextension
```

dadjeibaah:
@alpeb that's a good point about the fragility in using the `<extension-name>=1` pattern. As discussed offline, it might be helpful in adding a prefix to the extension name in the URL param to make it more explicit to the backend server that receives these request.

We can change this so that every extension gets an `ext-` prefix to denote that it is indeed an extension installed within the cluster.

alpeb (approved):
Thanks @dadjeibaah , this looks good to me :+1:

adleong (approved):

PR #5762, opened 2021-02-17
Add support for CLI extensions

adleong:
As described in https://github.com/linkerd/linkerd2/pull/5692, this PR adds support for CLI extensions.

Calling `linkerd foo` (if `foo` is not an existing Linkerd command) will now search the current PATH for an executable named `linkerd-foo` and invoke it with the current arguments.

* All arguments and flags will be passed to the extension command
* The Linkerd command itself will not process any flags
* To simplify parsing, flags are not allowed before the extension name

e.g. with an executable called `linkerd-foo` on my PATH:

```console
> linkerd foo install
Welcome to Linkerd foo!
Got: install
> linkerd foo --context=prod install
Welcome to Linkerd foo!
Got: --context=prod install
> linkerd --context=prod foo install
Cannot accept flags before Linkerd extension name
> linkerd bar install
Error: unknown command "bar" for "linkerd"
Run 'linkerd --help' for usage.
```

We also update `linkerd check` to invoke `linkerd <extension> check` for each extension found installed on the current cluster.  A check warning is emitted if the extension command is not found on the path.

e.g. with both `linkerd.io/extension=foo` and `linkerd.io/extension=bar` extensions installed on the cluster:

```console
> linkerd check
[...]
Linkerd extensions checks
=========================

Welcome to Linkerd foo!
Got: check --as-group=[] --cni-namespace=linkerd-cni --help=false --linkerd-cni-enabled=false --linkerd-namespace=linkerd --output=table --pre=false --proxy=false --verbose=false --wait=5m0s

linkerd-bar
-----------
‼ Linkerd extension command linkerd-bar exists

Status check results are ‼
```


cli/cmd/check.go
```diff
@@ -266,39 +269,45 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
+					if err != nil {
+						return err
+					}
+					fmt.Fprintln(wout, string(out))
+				} else {
+					fmt.Fprintln(wout, extensionCmd)
+					fmt.Fprintln(wout, strings.Repeat("-", len(extensionCmd)))
+					fmt.Fprintf(wout, "%s Linkerd extension command %s exists\n", healthcheck.WarnStatus, extensionCmd)
+					fmt.Fprintln(wout)
+					fmt.Fprintf(wout, "Status check results are %s\n", healthcheck.WarnStatus)
```
alpeb, line 302: For warning-only failures, the usual `linkerd check` returns OK for the overall success message ("Status check results are √"). So we might do the same here.

alpeb (commented):
When the extension has no command available, `linkerd check` will report the problem twice:
```
Linkerd extensions checks
=========================

linkerd-linkerd-foobar
----------------------
‼ Linkerd extension command linkerd-linkerd-foobar exists

Status check results are ‼
failed to run extensions checks: exec: "linkerd-linkerd-foobar": executable file not found in $PATH
```
Given each extension is handling its own error reporting, we could get rid of that final message (produced after having called `runExtensionChecks()` and just exit with the error code.

cli/main.go
```diff
@@ -1,13 +1,37 @@
...
+			fmt.Fprintln(os.Stderr, "Cannot accept flags before Linkerd extension name")
+			os.Exit(1)
+		}
+		path, err := exec.LookPath(fmt.Sprintf("linkerd-%s", args[0]))
+		if err == nil {
+			plugin := exec.Command(path, args[1:]...)
+			plugin.Stdin = os.Stdin
+			plugin.Stdout = os.Stdout
+			plugin.Stderr = os.Stderr
+			err = plugin.Run()
```
dadjeibaah, line 26: This section looks pretty similar to that of `check.go`. Could we DRY this up or is it not necessary?

cli/cmd/check.go
```diff
@@ -266,39 +269,45 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
-			hc.AppendCategories(*healthcheck.NewCategory(categoryID, checkers, true))
-			success := healthcheck.RunChecks(wout, werr, hc, opts.output)
-			if !success {
-				os.Exit(1)
+			extensionCmd := fmt.Sprintf("linkerd-%s", extension)
+			var path string
+			path, err = exec.LookPath(extensionCmd)
+			if err != nil {
+				if opts.output == healthcheck.JSONOutput {
+					result := map[string]interface{}{
```
dadjeibaah, line 277: I'm wondering if we should move print related work from here into `healthcheck_output.go` since other printing related work for the linkerd core checks are in that file. 🤔

dadjeibaah (commented):
So great to see this working! I left a few comments and questions.

cli/cmd/check.go
```diff
@@ -266,39 +269,45 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
+					fmt.Fprintln(wout, string(out))
+				} else {
+					fmt.Fprintln(wout, extensionCmd)
+					fmt.Fprintln(wout, strings.Repeat("-", len(extensionCmd)))
+					fmt.Fprintf(wout, "%s Linkerd extension command %s exists\n", healthcheck.WarnStatus, extensionCmd)
+					fmt.Fprintln(wout)
+					fmt.Fprintf(wout, "Status check results are %s\n", healthcheck.WarnStatus)
+				}
+			} else {
+				args := append([]string{"check"}, getGlobalFlags(cmd.Flags())...)
```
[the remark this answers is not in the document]
adleong: As described in https://github.com/linkerd/linkerd2/pull/5692, extensions will need to accept all of the global flags.

cli/cmd/check.go
adleong: yeah, that would certainly make sense, but I'm not sure exactly what the best way to structure this is.  I think ideally we'd make all of the various checkOutput structs public so that we could build the check result here in a more typesafe way.  But that's a bigger refactor than I want to tackle right now.

cli/cmd/check.go
```diff
@@ -266,39 +269,43 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
+						},
+					}
+					out, err := json.MarshalIndent(result, "", "  ")
+					if err != nil {
+						return err
+					}
+					fmt.Fprintln(wout, string(out))
+				} else {
+					fmt.Fprintln(wout, extensionCmd)
+					fmt.Fprintln(wout, strings.Repeat("-", len(extensionCmd)))
```
alpeb, line 297: The header for the extension is only being printed when the binary isn't found; this should go above instead.

cli/cmd/check.go
```diff
@@ -266,39 +269,43 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
-					APIAddr:               apiAddr,
-					VersionOverride:       opts.versionOverride,
-					RetryDeadline:         time.Now().Add(opts.wait),
-					CNIEnabled:            opts.cniEnabled,
-				})
-			hc.AppendCategories(*healthcheck.NewCategory(categoryID, checkers, true))
-			success := healthcheck.RunChecks(wout, werr, hc, opts.output)
-			if !success {
-				os.Exit(1)
+			extensionCmd := fmt.Sprintf("linkerd-%s", extension)
```
alpeb, line 272: Isn't expected that extensions be named like `linkerd-foobar`? If so, the expected command will be `linkerd-linkerd-foobar`. So maybe it's worth a bit more logic here to avoid prepending `linkerd-` if it's already there?

alpeb (commented):
There is a minor discrepancy where "Status check results are √"  is not being printed automatically (unless the external binary doing it itself), but I understand there's a mismatch between the handling of internal and external checks that makes unifying this difficult, which requires a deeper refactoring outside the scope of this PR. Perhaps also addressing internal checks through shelling-out might make sense, but that can wait.

kleimkuhler (approved):
👍

olix0r:
OK.. testing this:

1. Install linkerd
2. `k create ns linkerd-boop ; k label ns/linkerd-boop linkerd.io/extension: linkerd-boop`
3. `linkerd check`
```
...

linkerd-linkerd-boop
--------------------
‼ Linkerd extension command linkerd-linkerd-boop exists
```

This isn't at all clear what's going on... I assume this means that there's a warning that the `linkerd-linkerd-boop` command doesn't exist? But it should actually say that... The output there isn't helpful.

The other thing that's weird is `linkerd-linkerd-boop`... Should the extension label value just be `boop` instead of `linkerd-boop`? (I cribbed from the viz namespace, where the label value is `linkerd-viz`).

Running `echo "#!/bin/sh" > ~/bin/linkerd-linkerd-boop` (creating an extension file that isn't executable) -- i get the same warning message:

    ‼ Linkerd extension command linkerd-linkerd-boop exists 

It would be helpful to be more descriptive that the file exists and isn't executable...

Making it executable... `linkerd check` passes but doesn't list the extension at all? I would expect us to at least print something about the extension instead of silently ignoring it?

```
...

Linkerd extensions checks
=========================

linkerd-viz
-----------
√ tap API service is running
√ prometheus is installed and configured correctly
√ can initialize the client
√ viz extension self-check

Status check results are √
```

Without the viz extension, the output is even weirder:

```
...
Status check results are √

Linkerd extensions checks
=========================

```

(and nothing below the extension checks message)

Similarly, if the extension command fails with error (simply doing `exit 1`) no output is emitted... I'd expect the extension check to show some contextual error in this case.

With an extension script like:

```
#!/bin/sh

echo "some error message" >&2
exit 1
```

We get

```
...

Linkerd extensions checks
=========================

some error message
```

without any indication of which extension failed.

--

In general, i'm a bit skeptical of relying on the extension to only output plaintext without any handling by the wrapper process... it seems prone to very confusing/misleading output. I think, ideally, the main `linkerd` binary should handle formatting and the extension should only operate by outputing some structured output.

olix0r (approved):
I'm fine with shipping this, though I think we'll need to continue iterating on this between now and stable.

adleong:
This PR has been updated to always call extension checks with `--output=json` and parsing the resulting output.  For example an extension command like this:

```console
> cat linkerd-foo
#!/bin/sh

echo '{
  "success": true,
  "categories": [
    {
      "categoryName": "linkerd-foo",
      "checks": [
        {
          "description": "everything looks ship shape, captain!",
          "result": "success"
        },
        {
          "description": "actually, hold on...",
          "result": "warning",
          "error": "something is fishy here..."
        },
        {
          "description": "oh no!",
          "result": "error",
          "error": "abandon ship!"
        }
      ]
    }
  ]
}'

```

results in check output like:

```console
Linkerd extensions checks
=========================

linkerd-foo
-----------
√ Linkerd extension command linkerd-foo exists
√ everything looks ship shape, captain!
‼ actually, hold on...
    something is fishy here...
× oh no!
    abandon ship!

Status check results are ×
```

dadjeibaah (commented):
Works great! 📦 🚢

cli/cmd/check.go
```diff
@@ -254,58 +256,81 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
+				args := append([]string{"check"}, getExtensionCheckFlags(cmd.Flags())...)
+				plugin := exec.Command(path, args...)
+				output, err := plugin.Output()
+				if err != nil {
+					return false, err
+				}
+				extensionResults, err := healthcheck.ParseJSONCheckOutput(output)
+				if err != nil {
+
+					return false, fmt.Errorf("invalid extension check output:\n%s\n[%s]", string(output), err)
```
alpeb, line 304: I would add a hint here, like `"invalid extension check output (JSON object expected): ..."`

alpeb (commented):
Very nice :+1: 
Failures on our extensions (viz, mc, jaeger) are short-circuiting and not letting checks on other extensions continue because we exit the shell in their `check.go`'s. How about passing a hidden flag to avoid doing that?

adleong:
@alpeb great catch!  I've changed it so that even the built-in extensions get invoked by shelling out.  This makes the behavior more consistent and now if the extension check exits, we can still continue.

cli/cmd/check.go
```diff
@@ -254,103 +255,75 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
+		}
+	}
+	return success, nil
 }
 
-func getSharedFlags(lf *pflag.FlagSet, lf2 *pflag.FlagSet) []string {
+func getExtensionCheckFlags(lf *pflag.FlagSet) []string {
+	extensionFlags := []string{
+		"api-addr", "context", "as", "as-group", "kubeconfig", "linkerd-namespace", "verbose",
+		"namespace", "proxy", "wait",
```
alpeb, line 311: Not all extensions accept the same flags, and sending an unknown flag will cause an error. For example the mc check call is failing because it doesn't recognize `--proxy`.

cli/cmd/check.go
```diff
@@ -254,103 +255,75 @@ func runExtensionChecks(cmd *cobra.Command, wout io.Writer, werr io.Writer, opts
...
 		}
-		if err != nil {
-			return err
+		if err == nil {
+			plugin := exec.Command(path, args...)
+			output, _ := plugin.Output()
+			extensionResults, err := healthcheck.ParseJSONCheckOutput(output)
+			if err != nil {
+				command := fmt.Sprintf("%s %s", path, strings.Join(args, " "))
+				return false, fmt.Errorf("invalid extension check output from \"%s\" (JSON object expected):\n%s\n[%s]", command, string(output), err)
```
alpeb, line 296: If I run for example `linkerd check -L foobar` having linkerd installed in its default namespace, the core checks will fail as expected, and so will the multicluster checks with a message like:
```
failed to run extensions checks: invalid extension check output from "/home/alpeb/src/pr/5762/linkerd2/target/cli/linux-amd64/linkerd multicluster check --as-group=[] --linkerd-namespace=foobar --verbose=false --proxy=false --wait=5m0s --output=json" (JSON object expected):
```
The error message returned from multicluster won't be shown because we're not capturing stderr from the `plugin.Output()` call.
Also I think it'd be better here to not short-circuit and instead output the error message, set success to false, and continue with the other extensions.

adleong: Good point.  I've updated to handle this more gracefully by reporting the stderr if it exists, and the json parse error otherwise, and by not short circuiting in the case of an error:

```
linkerd-existence
-----------------
× 'linkerd-config' config map exists
    configmaps "linkerd-config" not found
    see https://linkerd.io/checks/#l5d-existence-linkerd-config for hints

Status check results are ×

Linkerd extensions checks
=========================

linkerd-foo
-----------
√ Linkerd extension command linkerd-foo exists
√ everything looks ship shape, captain!
‼ actually, hold on...
    something is fishy here...
× oh no!
    abandon ship!

Status check results are ×

linkerd-linkerd-multicluster
----------------------------
× Running: /Users/alexanderleong/workspace/linkerd2/target/cli/darwin/linkerd multicluster check --as-group=[] --linkerd-namespace=foobar --verbose=false --proxy=false --wait=5m0s --output=json
    Failed to fetch linkerd config: configmaps "linkerd-config" not found

    see https://linkerd.io/checks/#extensions for hints

Status check results are ×

linkerd-viz
-----------
√ tap API service is running
√ prometheus is installed and configured correctly
√ can initialize the client
√ viz extension self-check

Status check results are √
```

adleong: Specifically when reporting this kind of error, we report the category as `linkerd-linkerd-multicluster`.  Rather than adding special handling here to fix this, we should (in a separate PR) change the `linkerd.io/extension` label of the built-in extensions to remove the linkerd prefix.  Once that is changed, this will correctly report as `linkerd-multicluster`.

alpeb (approved):
Perfect :ok_hand: Thanks @adleong :medal_military:

multicluster/cmd/check.go
```diff
@@ -96,6 +96,10 @@ non-zero exit code.`,
 	}
 	cmd.Flags().StringVarP(&options.output, "output", "o", options.output, "Output format. One of: basic, json")
 	cmd.Flags().DurationVar(&options.wait, "wait", options.wait, "Maximum allowed time for all tests to pass")
+	cmd.Flags().Bool("proxy", false, "")
+	cmd.Flags().MarkHidden("proxy")
+	cmd.Flags().StringP("namespace", "n", "", "")
+	cmd.Flags().MarkHidden("namespace")
```
kleimkuhler, line 102: Maybe make a follow-up issue for supporting these?

kleimkuhler (approved):
Looks great!

multicluster/cmd/check.go
adleong: I don't think there are really any data-plane checks that the multicluster extension has to do.

PR #5766, opened 2021-02-17
Add services to proxy injector for opaque ports annotation

### LINKED ISSUE #5736: proxy-injector: pods should inherit opaque ports annotation from namespaces

### LINKED ISSUE #5737: proxy-injector: services should inherit opaque ports annotation from namespaces
This is equivalent to #5736 but for services.

The primary difference is that right now the proxy-injector does not inject
services, so the additional thing that needs to happen is the proxy-injector
will now also need to watch the cluster for service updates.

This should be done after #5736 and after stable 2.10.

kleimkuhler:
This adds namespace inheritance of the opaque ports annotation to services. 

This means that the proxy injector now watches services creation in a cluster.
When a new service is created, the webhook receives an admission request for
that service and determines whether a patch needs to be created.

A patch is created if the service does not have the annotation, but the
namespace does. This means the service inherits the annotation from the
namespace.

A patch is not created if the service and the namespace do not have the
annotation, or the service has the annotation. In the case of the service having
the annotation, we don't even need to check the namespace since it would not
inherit it anyways.

If a namespace has the annotation value changed, this will not be reflected on
the service. The service would need to be recreated so that it goes through
another admission request.

None of this applies to the `inject` command which still skips service
injection. We rely on being able to check the namespace annotations, and this is
only possible in the proxy injector webhook when we can query the k8s API.

Closes #5737

### DIFF
--- a/cli/cmd/inject.go
+++ b/cli/cmd/inject.go
@@ -169,6 +169,11 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 	}
 
 	reports := []inject.Report{*report}
+	// Injection of services depends on being able to retrieve the namespace
+	// annotations which can only occur in the proxy injector webhook.
+	if conf.IsService() {
+		return bytes, reports, nil
+	}
 
 	if rt.allowNsInject && conf.IsNamespace() {
 		b, err := conf.InjectNamespace(rt.overrideAnnotations)
@@ -195,7 +200,7 @@ func (rt resourceTransformerInject) transform(bytes []byte) ([]byte, []inject.Re
 		conf.AppendPodAnnotations(rt.overrideAnnotations)
 	}
 
-	patchJSON, err := conf.GetPatch(rt.injectProxy)
+	patchJSON, err := conf.GetPodPatch(rt.injectProxy)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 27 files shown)


cli/cmd/inject.go
```diff
@@ -318,7 +323,9 @@ func (resourceTransformerInject) generateReport(reports []inject.Report, output
 	}
 
 	for _, r := range reports {
-		if b, _ := r.Injectable(); b {
+		if r.Kind == k8s.Service {
+			output.Write([]byte(fmt.Sprintf("service \"%s\" skipped\n", r.Name)))
```
kleimkuhler, line 327: As mentioned in the description, this still results in `service "..." skipped` since we are skipping injection in the `inject` command path:

```
$ linkerd inject test/integration/testdata/smoke_test.yaml
...
deployment "smoke-test-terminus" injected
service "smoke-test-terminus-svc" skipped
deployment "smoke-test-gateway" injected
service "smoke-test-gateway-svc" skipped
```

alpeb (approved):
Looks good and tests great :+1:

alpeb:
Fixes #5737

olix0r (approved):
This looks good! easy to read.

adleong (approved):

Issue #5717, opened 2021-02-10
multicluster extension should check that the core installation is healthy


kleimkuhler:
Try to install the Viz extension immediately after core and the following is
printed:

```
❯ bin/linkerd viz install |kubectl apply -f -
Waiting for control plane to become available
...
```

This re-prints every few seconds until the core install is passing it's checks,
and then the Viz extension is installed. This is important because the core
checks are making sure identity and the proxy injector are up and running.

Try to install multicluster directly after core, and it installs immediately.
Instead, `multicluster install` should do a similar Linkerd existence check and
make sure the core install is healthy before installing the multicluster
extension.

PR #5767, opened 2021-02-18
cli: make jaeger and multicluster installs wait for core cp

Pothulapati:
This PR updates the jaeger and multicluster installs to wait
for the core control-plane to be up before moving to the rendering
logic. This prevents these components from being installed before
the injector is up and running correctly.

`--skip-checks` has been added to jaeger to skip these checks. The
same has not been added to `multicluster` as the install fails when
there is no core cp is present.

This PR also cleans up extra core cp check that we have for `viz install`

### DIFF
--- a/viz/cmd/install.go
+++ b/viz/cmd/install.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


alpeb (approved):
Very nice, great reliance on the healthcheck lib :+1:

alpeb:
Fixes #5717

viz/cmd/install.go
```diff
@@ -93,7 +76,7 @@ A full list of configurable values can be found at https://www.github.com/linker
 
 	cmd.Flags().BoolVar(
 		&skipChecks, "skip-checks", false,
-		`Skip checks for namespace existence`,
+		`Skip checks for linkerd core control-plane existence`,
```
dadjeibaah, line 79: code formatting nit: it seems to me that all other command flags are written out on a single line, however, this file has its command line flags spanning 3 lines. Should we make sure that this file matches up with other cmd files?

dadjeibaah (commented):
Looks great! I just had one minor comment.

viz/cmd/install.go
Pothulapati: Makes sense. Updating...

kleimkuhler (approved):
Awesome!

PR #5768, opened 2021-02-18
 Fix for CLI printing command usage for API Errors for multiple commands

piyushsingariya:
Signed-off-by: Piyush Singariya piyushsingariya@gmail.com

Related to #5058

Problem: CLI prints command usage for multiple commands in-case of API errors
Solution: Print the error and the exit using os.Exit(1) to avoid Cobra printing usage

### DIFF
--- a/cli/cmd/endpoints.go
+++ b/cli/cmd/endpoints.go
@@ -98,7 +98,8 @@ destination.`,
 				APIAddr:               apiAddr,
 			}), args)
 			if err != nil {
-				return fmt.Errorf("Destination API error: %s", err)
+				fmt.Fprint(os.Stderr, fmt.Errorf("Destination API error: %s", err))
+				os.Exit(1)
 			}
 
 			output := renderEndpoints(endpoints, options)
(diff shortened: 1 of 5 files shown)


piyushsingariya:
Hey @adleong @alpeb, I messed the previous PR so I opened up a fresh one.
and I think that the tests are failing because they're expecting an error in return but the changes are simply printing the error and exiting the process. :thinking:

alpeb (commented):
For the current error, it should be just a matter of replacing this message
https://github.com/linkerd/linkerd2/blob/89ec6056c02632d88026c13b89ae212d1b71e4e7/test/integration/endpoints/endpoints_test.go#L80
with
```
 if stderrOut[0] != "Destination API error: Invalid authority: foo" { 
```

piyushsingariya:
@alpeb Yes, it worked! :tada:

alpeb (approved):
Thanks @piyushsingariya , looks good to me :+1:

kleimkuhler (approved):
Looks great @piyushsingariya, thank you!

PR #5770, opened 2021-02-18
Mirror opaque port annotations on services

### LINKED ISSUE #5650: multicluster: Mirror should propagate opaqueness

kleimkuhler:
This change introduces an opaque ports annotation watcher that will send
destination profile updates when a service has its opaque ports annotation
change.

The user facing change introduced by this is that the opaque ports annotation is
now required on services when using the multicluster extension. This is because
the service mirror will create mirrored services in the source cluster, and
destination lookups in the source cluster need to discover that the workloads in
the target cluster are opaque protocols.

### Why

Closes #5650

### How

The destination server now has a new opaque ports annotation watcher. When a
client subscribes to updates for a service name or cluster IP, the `GetProfile`
method creates a profile translator stack that passes updates through resource
adaptors such as: traffic split adaptor, service profile adaptor, and now opaque
ports adaptor.

When the annotation on a service changes, the update is passed through to the
client where the `opaque_protocol` field will either be set to true or false.

A few scenarios to consider are:

  - If the annotation is removed from the service, the client should receive
    an update with no opaque ports set.
  - If the service is deleted, the stream stays open so the client should
    receive an update with no opaque ports set.
  - If the service has the annotation added, the client should receive that
    update.

### Testing

Unit test have been added to the watcher as well as the destination server.

An integration test has been added that tests the opaque port annotation on a
service.

For manual testing, using the destination server scripts is easiest:

```
# install Linkerd

# start the destination server
$ go run controller/cmd/main.go destination -kubeconfig ~/.kube/config

# Create a service or namespace with the annotation and inject it

# get the destination profile for that service and observe the opaque protocol field
$ go run controller/script/destination-client/main.go -method getProfile -path test-svc.default.svc.cluster.local:8080
INFO[0000] fully_qualified_name:"terminus-svc.default.svc.cluster.local" opaque_protocol:true retry_budget:{retry_ratio:0.2 min_retries_per_second:10 ttl:{seconds:10}} dst_overrides:{authority:"terminus-svc.default.svc.cluster.local.:8080" weight:10000} 
INFO[0000]                                              
INFO[0000] fully_qualified_name:"terminus-svc.default.svc.cluster.local" opaque_protocol:true retry_budget:{retry_ratio:0.2 min_retries_per_second:10 ttl:{seconds:10}} dst_overrides:{authority:"terminus-svc.default.svc.cluster.local.:8080" weight:10000} 
INFO[0000]
```


controller/api/destination/opaque_ports_adaptor.go
```diff
@@ -1,108 +1,39 @@
...
-// ports annotation on the pods and update the list of opaque ports.
-//
-// opaquePortsAdaptor also implements ProfileUpdateListener so that it can
-// receive updates from trafficSplitAdaptor and merge the service profile by
-// adding the list of opaque ports.
-//
-// When either of these implemented interfaces has an update,
-// opaquePortsAdaptor publishes the new service profile to the
-// profileTranslator.
 type opaquePortsAdaptor struct {
```
olix0r, line 8: Perhaps worth an updated doc comment?

controller/api/destination/opaque_ports_adaptor.go
```diff
@@ -1,108 +1,39 @@
...
-			if err != nil {
-				opa.log.Errorf("Failed to get opaque ports annotation for pod %s: %s", pod, err)
-			}
-			for port := range override {
-				ports[port] = struct{}{}
-			}
-		}
-	}
-	return ports
+func (sa *opaquePortsAdaptor) UpdateService(ports map[uint32]struct{}) {
```
olix0r, line 25: nit: Why the change from `opa` to `sa`? `opa` seems fine/good as it's at least more consistent. Unclear what `sa` references...

pkg/util/parsing.go
```diff
@@ -49,10 +35,42 @@ func ParseOpaquePorts(override string, containers []corev1.Container) []string {
...
+			if p.Name == pr {
+				return p.ContainerPort, true
+			}
+		}
+	}
+	return 0, false
+}
+
+// ParseOpaquePorts parses the opaque ports annotation into a list of ports
+func ParseOpaquePorts(override string) []string {
```
olix0r, line 53: Is this specific to Opaque Ports? Or is this really just `ParsePortRanges`

Also, from docs, it's unclear how this differs from `GetPortRanges` and why it returns a list of strings. I would expect parsing ports to produce a list of port numbers...

olix0r (commented):
Some minor notes... I'd like to give this a test, but nothing stands out as a blocker to me.

controller/api/destination/opaque_ports_adaptor.go
kleimkuhler: Yep this is from a previous iteration where I had renamed (and then un-renamed) this. I will change it back!

pkg/util/parsing.go
kleimkuhler: We don't actually use this anymore so I can remove this.

kleimkuhler: It's purpose got split up into `ParseContainerOpaquePorts` and `parseServiceOpaquePorts`.

controller/api/destination/watcher/opaque_ports_watcher.go
```diff
@@ -0,0 +1,232 @@
...
+		Namespace: svc.Namespace,
+		Name:      svc.Name,
+	}
+	opaquePorts, err := getServiceOpaquePortsAnnotation(svc)
+	if err != nil {
+		opw.log.Errorf("failed to get %s service opaque ports annotation: %s", id, err)
+		return
+	}
+	// If the service has no opaque ports there are no updates to send.
+	if len(opaquePorts) == 0 {
```
adleong, line 124: Don't we want to send an update if the opaque ports list changes from non-empty to empty?

controller/api/destination/watcher/opaque_ports_watcher.go
```diff
@@ -0,0 +1,232 @@
...
+	if !ok {
+		opw.subscriptions[id] = &svcSubscriptions{
+			opaquePorts: opaquePorts,
+			listeners:   []OpaquePortsUpdateListener{},
+		}
+		return
+	}
+	// Do not send updates if there was no change in the opaque ports; if
+	// there was, send an update to each listener.
+	if !reflect.DeepEqual(ss.opaquePorts, opaquePorts) {
```
adleong, line 139: I like to avoid reflect if at all possible.  You can write a quick function for comparing two maps by iterating through them.  It's the "go way".

controller/api/destination/watcher/opaque_ports_watcher.go
kleimkuhler: Yes! We do. Good catch.

I changed the unit tests in `opaque_ports_watcher_test.go` to have an additional `addService` before deletion which acts as an update. I added an additional test that tests this.

controller/api/destination/watcher/opaque_ports_watcher.go
kleimkuhler: Sounds good I removed it. I thought we used it in a lot more places based off searching the uses of `reflect.DeepEqual` in the code, but looks like a lot is pointer comparisons. Not sure of other places we use this for maps.

adleong (approved):
🦘 🥚

alpeb (approved):
Looks good and tests well :+1: 
I tested in multicluster, annotating the service at the target cluster, and could see how linkerd-destination on the source cluster returned a new response with the opaque ports flag true :clap: 

One question: In the integration test, we see the proxy continues to forward traffic with TLS. If there was a problem processing the opaque-ports annotation, would that metric be different?

kleimkuhler:
@alpeb Great question! I investigated this a little further and it turns out the old test for pods with the OP annotation was failing, but we were not catching that error. Now that we don't track the endpoints of a service, the old test was not picking up the OP annotation on the service's pods. However, the metric we were testing for wasn't helpful enough to catch that the test was failing.

What I did to fix these tests:

1. We now check the metrics of the slow cooker pods. Each of the slow cooker pods should no HTTP outbound requests. This is because they should discover their targets are opaque protocols, so it should be TCP.
2. I made the opaque pod service a headless service. This allows the slow cooker pod for that service to target the pod IPs which do have the annotation
3. I fixed an issue in the regex that was causing this to not catch errors anyways.

kleimkuhler:
A final change for this is adding back (and fixing) the tcp metric test on the application pod to make sure there is an inbound connection from the slow cooker. Coupled with the absence of HTTP request totals on the slow cooker, these tests should be good to go now.

alpeb:
Post-merge comment: Thanks @kleimkuhler for the comprehensive test; helped me clarify some things about opaque ports!

PR #5775, opened 2021-02-18
send serviceprofile counts in heartbeat URL

### LINKED ISSUE #5474: heartbeat: Report ServiceProfile usage
I'd like to begin deprecating the `ServiceProfile` CRD in 2.11 (to replaced by a family of more flexible traffic policy CRDs). We don't really have any insight, though, into how widely the `ServiceProfile` API is used and how impactful this change will be.

In advance of these changes, I propose that:

* The destination controller be instrumented with the total number of (unique) ServiceProfile instances currently in the cluster, e.g. `serviceprofile_instances_count`. This should be a gauge of the current number of service profile instances in the cluster that the controller is observing.
* The heartbeat reporter be updated to query/report this metric (presumably taking the maximum value from all destination controller instances).

dadjeibaah:
This change counts the number of service profiles installed in a cluster
and adds that info to the heartbeat HTTP request.

Fixes #5474

### DIFF
--- a/charts/linkerd2/templates/heartbeat-rbac.yaml
+++ b/charts/linkerd2/templates/heartbeat-rbac.yaml
@@ -42,6 +42,9 @@ rules:
 - apiGroups: [""]
   resources: ["namespaces"]
   verbs: ["list"]
+- apiGroups: ["linkerd.io"]
+  resources: ["serviceprofiles"]
+  verbs: ["list"]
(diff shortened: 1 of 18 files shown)


adleong (approved):
Nice and simple! 📽️ 🦺

Pothulapati (approved):
LGTM

PR #5779, opened 2021-02-19
tests: enable external-prometheus-deep with cleanup logic

Pothulapati:
This PR enables the temporarily disabled external-prometheus-deep
integration test. This also fixes the clean up issue by essentially
moving the external-prometheus resources into `external-prometheus`
which has the relevant annotation required to be deleted by
`bin/test-cleanup`

This PR also updates the test resource label to be `test.linkerd.io/is-test-data-plane` from
`linkerd.io/is-test-data-plane` to prevent `linkerd inject` from removing it.

### DIFF
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -7,7 +7,7 @@ set +e
 ##### Test setup helpers #####
 
 export default_test_names=(deep external-issuer helm-deep helm-upgrade uninstall upgrade-edge upgrade-stable)
-export all_test_names=(cluster-domain cni-calico-deep multicluster "${default_test_names[*]}")
+export all_test_names=(cluster-domain cni-calico-deep external-prometheus-deep multicluster "${default_test_names[*]}")
 
 tests_usage() {
   progname="${0##*/}"
--- a/bin/_test-helpers.sh
+++ b/bin/_test-helpers.sh
@@ -6,8 +6,8 @@ set +e
 
 ##### Test setup helpers #####
 
-export default_test_names=(deep external-issuer helm-deep helm-upgrade uninstall upgrade-edge upgrade-stable)
-export all_test_names=(cluster-domain cni-calico-deep external-prometheus-deep multicluster "${default_test_names[*]}")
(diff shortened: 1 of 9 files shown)


bin/_test-helpers.sh
```diff
@@ -7,7 +7,7 @@ set +e
 ##### Test setup helpers #####
 
 export default_test_names=(deep external-issuer helm-deep helm-upgrade uninstall upgrade-edge upgrade-stable)
-export all_test_names=(cluster-domain cni-calico-deep multicluster "${default_test_names[*]}")
+export all_test_names=(cluster-domain cni-calico-deep external-prometheus-deep multicluster "${default_test_names[*]}")
```
alpeb, line 10: I think it's fine to keep `external-prometheus-deep` in `default_test_names` because it doesn't require re-creating the cluster(s) unlike `cluster-domain`, `cni-calico-deep` and `multicluster`

alpeb (commented):
Thanks @Pothulapati 
I think the cluster-level resources need also to be annotated, and `bin/test-cleanup` extended to not only delete the annotated namespace, but also the cluster-level resources.

bin/test-cleanup
```diff
@@ -22,6 +22,9 @@ echo "cleaning up jaeger extension resources, if present [${k8s_context}]"
 echo "cleaning up the all namespaces labelled with linkerd.io/is-test-data-plane"
 kubectl --context="$k8s_context" delete ns -l linkerd.io/is-test-data-plane
 
+echo "cleaning up ohter cluster-scoped resources labelled with linkerd.io/is-test-data-plane"
+kubectl --context="$k8s_context" delete clusterRole,clusterRoleBindings -l linkerd.io/is-test-data-planee
```
Pothulapati, line 26: `all` could not be used here, as its only meant for `namespace-scoped` resources like pods, etc. Are there any other cluster-scoped resources that we could add? 🤔

alpeb (approved):
LGTM! :+1:

kleimkuhler (approved):
Great! 👍

PR #5791, opened 2021-02-22
Stop loading cli-bin in bin/image-load

alpeb:
The CLI binaries are never used inside a cluster so there's no need to
load the cli-bin image.

Also, I always have `LINKERD_LOCAL_BUILD_CLI=1` in my environment to avoid
building that image to begin with (takes more than 80s in my machine
with warm docker dependencies), which caused `bin/image-load` to fail
when it attempted to load it.

### DIFF
--- a/bin/image-load
+++ b/bin/image-load
@@ -79,7 +79,7 @@ fi
 "$bin" version
 
 rm -f load_fail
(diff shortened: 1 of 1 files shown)


dadjeibaah (approved):
📦 🚢

kleimkuhler (approved):
👍

PR #5799, opened 2021-02-22
Report better errors for pods with tap disabled

adleong:
Fixes https://github.com/linkerd/linkerd2/discussions/5777

When a user runs `linkerd viz check --proxy`, it will print a warning if there are any proxies which cannot be tapped.  This is a normal state of affairs after freshly installing the linkerd-viz extensions because any existing pods will need to be restarted before they can be tapped.  The check warning may lead users to falsely believe that something has gone wrong with their installation.

We remove this specific check from `linkerd viz check --proxy`.  To replace it, we improve the error output when attempting to tap a resource which is not tappable.  This gives the user actionable feedback when the tap command fails.

Old:

```console
> linkerd viz tap -n emojivoto deploy/vote-bot
no pods to tap for deployment/vote-bot
```

New:

```console
> linkerd viz tap -n emojivoto deploy/vote-bot
no pods to tap for deployment/vote-bot
1 pods found with tap not enabled:
	* vote-bot-64dd87cb87-7mcv4
restart these pods to enable tap and make them valid tap targets
```

### DIFF
--- a/test/integration/tap/tap_test.go
+++ b/test/integration/tap/tap_test.go
@@ -144,8 +144,8 @@ func TestCliTap(t *testing.T) {
 			if stderr == "" {
 				testutil.Fatal(t, "expected an error, got none")
 			}
-			expectedErr := `Error: no pods to tap for deployment/t4
-pods found with tap disabled via the viz.linkerd.io/disable-tap annotation`
+			expectedErr := `no pods to tap for deployment/t4
+1 pods found with tap disabled via the viz.linkerd.io/disable-tap annotation:`
 			split := strings.Split(stderr, "\n")
 			actualErr := strings.Join(split[:2], "\n")
 			if actualErr != expectedErr {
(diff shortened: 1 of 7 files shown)


Pothulapati:
👍🏼 on this change as this makes this warning case narrower i.e whenever a user tries to tap the pod, instead of getting this warning every time a check is performed.

Looks like the test failure is from `TestCliTap` which is expecting the older msg for disabled deployments.

kleimkuhler (approved):
I wasn't sure when first reading the error message how I felt about keeping `no pods to tap for ...`, but I think it makes sense to it there. If it just said `x pods found with tap disabled (or not enabled) ...`, it may not be as obvious that there are no valid pods to tap.

Additionally I'm still in favor of keeping a way for users to check this without running into it when using the `tap` command, but I think a follow-up issue makes sense for that.

alpeb (approved):
LGTM :+1:

PR #5803, opened 2021-02-23
Remove linkerd prefix from extension resources

dadjeibaah:
This change removes the `linkerd-` prefix on all non-cluster resources
in the jaeger and viz linkerd extensions. Removing the prefix makes all
linkerd extensions consistent in their naming.

### DIFF
--- a/jaeger/charts/linkerd-jaeger/README.md
+++ b/jaeger/charts/linkerd-jaeger/README.md
@@ -75,8 +75,6 @@ Kubernetes: `>=1.16.0-0`
 | collector.image.pullPolicy | string | `"Always"` |  |
 | collector.image.version | string | `"0.1.11"` |  |
 | collector.jaegerAddr | string | `nil` | address of the jaeger backend to send traces to |
-| createdByAnnotation | string | `"linkerd.io/created-by"` |  |
-| extensionAnnotation | string | `"linkerd.io/extension"` |  |
 | jaeger.image.name | string | `"jaegertracing/all-in-one"` |  |
 | jaeger.image.pullPolicy | string | `"Always"` |  |
 | jaeger.image.version | string | `"1.19.2"` |  |
--- a/jaeger/charts/linkerd-jaeger/README.md
+++ b/jaeger/charts/linkerd-jaeger/README.md
@@ -75,6 +75,8 @@ Kubernetes: `>=1.16.0-0`
 | collector.image.pullPolicy | string | `"Always"` |  |
 | collector.image.version | string | `"0.1.11"` |  |
 | collector.jaegerAddr | string | `nil` | address of the jaeger backend to send traces to |
+| createdByAnnotation | string | `"linkerd.io/created-by"` |  |
+| extensionAnnotation | string | `"linkerd.io/extension"` |  |
 | jaeger.image.name | string | `"jaegertracing/all-in-one"` |  |
 | jaeger.image.pullPolicy | string | `"Always"` |  |
 | jaeger.image.version | string | `"1.19.2"` |  |
--- a/viz/charts/linkerd-viz/templates/prometheus.yaml
+++ b/viz/charts/linkerd-viz/templates/prometheus.yaml
@@ -9,7 +9,7 @@ metadata:
   name: prometheus-config
   namespace: {{.Values.namespace}}
   labels:
-    {{.Values.extensionAnnotation}}: viz
+    linkerd.io/extension: viz
     component: prometheus
     namespace: {{.Values.namespace}}
   annotations:
@@ -177,7 +177,7 @@ metadata:
   name: prometheus
   namespace: {{.Values.namespace}}
   labels:
-    {{.Values.extensionAnnotation}}: viz
+    linkerd.io/extension: viz
     component: prometheus
     namespace: {{.Values.namespace}}
   annotations:
@@ -185,7 +185,7 @@ metadata:
 spec:
   type: ClusterIP
   selector:
-    {{.Values.extensionAnnotation}}: viz
+    linkerd.io/extension: viz
     component: prometheus
   ports:
   - name: admin-http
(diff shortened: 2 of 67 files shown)


jaeger/charts/linkerd-jaeger/README.md
```diff
@@ -75,6 +75,8 @@ Kubernetes: `>=1.16.0-0`
 | collector.image.pullPolicy | string | `"Always"` |  |
 | collector.image.version | string | `"0.1.11"` |  |
 | collector.jaegerAddr | string | `nil` | address of the jaeger backend to send traces to |
+| createdByAnnotation | string | `"linkerd.io/created-by"` |  |
```
adleong, line 78: These probably got added back in due to an automatic merge? I think they should be removed.

viz/charts/linkerd-viz/values.yaml
```diff
@@ -12,6 +12,8 @@ clusterDomain: cluster.local
 # @default -- clusterDomain
 identityTrustDomain: ""
 
+extensionAnnotation: linkerd.io/extension
```
adleong, line 15: I don't think we want the annotation key to be configurable.  I think we should revert this part of the change.

adleong (commented):
The rename parts of this look good but I think we don't want any of the changes which move back to using configurable annotation/label keys.  These keys should never change and should be hardcoded in the templates.

viz/charts/linkerd-viz/values.yaml
dadjeibaah: Ah ok. I can get that removed.

dadjeibaah:
@adleong, just to confirm, we still want the addition of annotation keys for jaeger for example, we just want it to be hardcoded in the helm charts, is that correct?

adleong:
Yeah, it's most important to have that label on the cluster-scoped resources (for the purposes of uninstall), and I think it's already there.  For the namespace scoped resources it's less important, but we might as well, I guess.

adleong (approved):
Looks good! 🐨

olix0r:
Spot-checking this branch I think there may be a few places that still need updating:
```
:; grep -RF 'extension: linkerd-' .
./multicluster/charts/linkerd-multicluster/templates/remote-access-service-mirror-rbac.yam    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/remote-access-service-mirror-rbac.yam    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/remote-access-service-mirror-rbac.yam    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/gateway.yaml:    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/gateway.yaml:    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/gateway.yaml:    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/gateway.yaml:    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/namespace.yaml:    linkerd.io/extension: linkerd-multicluster
./multicluster/charts/linkerd-multicluster/templates/link-crd.yaml:    linkerd.io/extension: linkerd-multicluster
```
```
#; golly:~/b/l2  (dad/naming-consistency ✔)
:; rg ExtensionName\ = 
viz/cmd/root.go
14:     ExtensionName = "viz"

jaeger/cmd/check.go
21:     JaegerExtensionName = "jaeger"

multicluster/cmd/check.go
29:     MulticlusterExtensionName = "linkerd-multicluster"

test/integration/install_test.go
87:     multiclusterExtensionName = "linkerd-multicluster"

viz/pkg/healthcheck/healthcheck.go
25:     VizExtensionName = "viz"

web/app/js/components/Navigation.jsx
47:const multiclusterExtensionName = 'linkerd-multicluster';

web/app/js/components/Gateway.jsx
15:const multiclusterExtensionName = 'linkerd-multicluster';
```

adleong:
We avoid renaming any resources in the multicluster extension here to avoid upgrade issues.  However, since the `linkerd.io/extension` label is new, there should be no concerns about updating the value of that label in the multicluster extension from linkerd-multicluster to multicluster.  So :+1: to making that change here.

dadjeibaah:
@olix0r, I was under the impression that we didn't want to modify the prefixes in the multicluster extension since it is on stable 2.9 and that we could modify anything in viz and jaeger since its not on stable yet.

olix0r:
@dadjeibaah as Alex mentioned, the label is new; so while we shouldn't update multi cluster _resources_, the label should probably be consistent.

viz/charts/linkerd-viz/templates/prometheus.yaml
```diff
@@ -294,25 +294,25 @@ spec:
...
 ---
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   labels:
-    linkerd.io/extension: linkerd-viz
+    linkerd.io/extension: viz
     app.kubernetes.io/name: prometheus
     app.kubernetes.io/part-of: Linkerd
     app.kubernetes.io/version: {{default .Values.linkerdVersion .Values.controllerImageVersion}}
```
alpeb, line 313: :point_up:

viz/charts/linkerd-viz/templates/tap.yaml
```diff
@@ -32,19 +32,19 @@ metadata:
   annotations:
     {{ include "partials.annotations.created-by" . }}
   labels:
-    linkerd.io/extension: linkerd-viz
+    linkerd.io/extension: viz
     app.kubernetes.io/name: tap
     app.kubernetes.io/part-of: Linkerd
     app.kubernetes.io/version: {{default .Values.linkerdVersion .Values.controllerImageVersion}}
```
alpeb, line 38: :eye:

alpeb (commented):
Looking good :+1: 
Just a nit around `app.kubernetes.io/version`

adleong (approved):

alpeb (approved):
:+1:

PR #5805, opened 2021-02-23
tests: Add TestOverridesSecret integration test

### LINKED ISSUE #5670: tests: check for overrides secret during upgrades
## Bug Report

To prevent things like #5646, We should make the upgrade integration tests check for changes in the secret so that we can verify whatever overrides are applied, are being stored correctly.

This can't be added as part of #5653 as the current upgrade tests are based on the edge release, and hence a separate PR can be raised once an edge is released with those changes.

Pothulapati:
Fixes #5670

This PR adds a new test under install tests called
`TestOverridesSecret` which reads the overrides secret
stored in the cluster and then checks if the relevant overrides
that are passed through helm override flags and CLi flags
are present or not.

Currently, Helm overrides flags are only available with the
edge releases and hence these flags are gated based on the version
in the install logic.


bin/_test-helpers.sh
```diff
@@ -362,7 +362,14 @@ install_version() {
 
     (
         set -x
-        "$linkerd_path" install | kubectl --context="$context" apply -f - 2>&1
+        # TODO: Use a mix of helm override flags and CLI flags and remove this condition
+        # once stable-2.10 is out
+        edge_regex='(edge)-([0-9]+\.[0-9]+\.[0-9]+)'
+        if [[ "$version" =~ $edge_regex ]]; then
+          "$linkerd_path" install --set proxyInit.ignoreInboundPorts="1234\,5678" --set controllerLogLevel=debug | kubectl --context="$context" apply -f - 2>&1
```
alpeb, line 369: can you use a mix of helm and non-helm flags here?

test/integration/install_test.go
```diff
@@ -670,6 +672,77 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
 	}
 }
 
+func TestOverridesSecret(t *testing.T) {
+
+	// Skip if Helm integration test
+	if TestHelper.IsHelm() {
+		t.Skip("Skipping as this is not a helm upgrade test")
+	}
```
alpeb, line 680: If you're introducing `TestHelper.IsHelm()`, would you update the other checks above to also us this same function?
Also, I think the code comment is not necessary :wink:

test/integration/install_test.go
```diff
@@ -349,7 +351,7 @@ func TestInstallOrUpgradeCli(t *testing.T) {
 		}
 
 		// prepare for stage 2
-		args = append([]string{"control-plane", "--addon-overwrite"}, args...)
+		args = append([]string{"control-plane", "--skip-outbound-ports", skippedOutboundPorts}, args...)
```
alpeb, line 354: Also can you add here helm-style flags as well?

alpeb (commented):
Looking good :+1: 
Can you also iterate over `overridesTree` keys to make sure there are no unintended entries? E.g. last time we discovered `stage` was sneaking in.

test/integration/install_test.go
Pothulapati: I don't think so. because this part is invoked both with `upgrade-edge` and `upgrade-stable` which would mean `upgrade-stable would fail`. Is there a way to gate this to be only for the edges?

test/integration/install_test.go
Pothulapati: I don't think we can replace the checks above. As in those cases, They want to skip only if the test is a helm upgrade test. In this case, We have to skip in all cases where Helm is involved as there is no overrides secret available with helm.

Also, updated the skip description to be more verbose.

bin/_test-helpers.sh
Pothulapati: Makes sense. Updated.

bin/_test-helpers.sh
```diff
@@ -362,7 +362,14 @@ install_version() {
 
     (
         set -x
-        "$linkerd_path" install | kubectl --context="$context" apply -f - 2>&1
+        # TODO: Use a mix of helm override flags and CLI flags and remove this condition
+        # once stable-2.10 is out
```
kleimkuhler, line 366: Can you make an issue for this? A TODO is a little easy to forget and I think this is definitely something that will be good to test as soon as the stable is out.

kleimkuhler (approved):
Left a comment about returning to this once the stable is out, but otherwise looks good!

test/integration/install_test.go
alpeb: Sorry for my vague suggestion. I mean replacing `TestHelper.GetHelmReleaseName() (=|!)= ""` with `TestHelper.IsHelm()`. And while you're at it actually remove the `GetHelmReleaseName()` method declaration and the associated `helmReleaseName` property in `test_helper.go` as they won't be used at all,

test/integration/install_test.go
alpeb: .... well, there's `TestHelper.UpgradeFromVersion()`, which could be augmented through some wrapper checking if it corresponds to an edge or a stable...

test/integration/install_test.go
Pothulapati: I removed `GetHelmReleaseName()` but was not able to remove `helm.releaseName` as the release name is used with helm upgrades, etc. So, Makes sense to have it still. WDYT?

test/integration/install_test.go
Pothulapati: Ahh, Let me try. Also, Def feels like an important test to check if things are actually persisted after upgrades.

test/integration/install_test.go
Pothulapati: Okay, I changed it to use `GetHelmReleaseName` isntead of my new `IsHelm` as it was breaking some stuff. Looks like `releaseName` is a better toggle to rely on then helm binary

kleimkuhler:
CI should be good after a `main` merge

Pothulapati:
> Can you also iterate over overridesTree keys to make sure there are no unintended entries? E.g. last time we discovered stage was sneaking in.

That definitely seems useful. Currently working on it.

test/integration/install_test.go
```diff
@@ -359,7 +363,17 @@ func TestInstallOrUpgradeCli(t *testing.T) {
...
 		// prepare for stage 2
-		args = append([]string{"control-plane", "--addon-overwrite"}, args...)
+		args = append([]string{"control-plane"}, args...)
+		edge, err := regexp.Match(`(edge)-([0-9]+\.[0-9]+\.[0-9]+)`, []byte(TestHelper.UpgradeFromVersion()))
+		if err != nil {
+			testutil.AnnotatedFatal(t, "could not match regex", err)
+		}
+
+		if edge {
+			args = append(args, []string{"--set", fmt.Sprintf("proxyInit.ignoreOutboundPorts=\"%s\"", strings.Replace("1234,5678", ",", "\\,", 1))}...)
```
alpeb, line 373:
```suggestion
			args = append(args, []string{"--set", fmt.Sprintf("proxyInit.ignoreOutboundPorts=\"%s\"", strings.Replace(skippedOutboundPorts, ",", "\\,", 1))}...)
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+	overrides := configOverridesSecret.Data["linkerd-config-overrides"]
+	overridesTree, err := tree.BytesToTree(overrides)
+	if err != nil {
+		testutil.AnnotatedFatalf(t, "could not retrieve linkerd-config-overrides",
+			"could not retrieve linkerd-config-overrides\n%s", err.Error())
+	}
+
+	// Check for fields that were added during install
+	testCases := []struct {
+		tree  tree.Tree
```
alpeb, line 717: Why parametrizing `tree` if it's always `overridesTree`?

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+	}{
+		{
+			overridesTree,
+			[]string{"controllerLogLevel"},
+			"debug",
+		},
+		{
+			overridesTree,
+			[]string{"proxyInit", "ignoreInboundPorts"},
+			"1234,5678",
```
alpeb, line 729:
```suggestion
			skippedInboundPorts,
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+	if TestHelper.UpgradeFromVersion() != "" {
+		testCases = append(testCases, []struct {
+			tree  tree.Tree
+			path  []string
+			value string
+		}{
+			{
+				overridesTree,
+				[]string{"proxyInit", "ignoreOutboundPorts"},
+				"1234,5678",
```
alpeb, line 743:
```suggestion
				skippedOutboundPorts,
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+		}...)
+	}
+
+	for _, tc := range testCases {
+		tc := tc // pin
+		t.Run(fmt.Sprintf("%s: %s", strings.Join(tc.path, "/"), tc.value), func(t *testing.T) {
+			finalValue, err := tc.tree.GetString(tc.path...)
+			if err != nil {
+				testutil.AnnotatedFatalf(t, "could not perform tree.GetString",
+					"could not perform tree.GetString\n%s", err.Error())
```
alpeb, line 754:
```suggestion
					"could not perform tree.GetString\n%s", err)
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+func TestOverridesSecret(t *testing.T) {
+
+	if TestHelper.GetHelmReleaseName() != "" {
+		t.Skip("Skipping as this is a helm test where linkerd-config-overrides is absent")
+	}
+
+	configOverridesSecret, err := TestHelper.KubernetesHelper.GetSecret(context.Background(), TestHelper.GetLinkerdNamespace(), "linkerd-config-overrides")
+	if err != nil {
+		testutil.AnnotatedFatalf(t, "could not retrieve linkerd-config-overrides",
+			"could not retrieve linkerd-config-overrides\n%s", err.Error())
```
alpeb, line 705: `.Error()` not needed
```suggestion
			"could not retrieve linkerd-config-overrides\n%s", err)
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+	}
+
+	overrides := configOverridesSecret.Data["linkerd-config-overrides"]
+	overridesTree, err := tree.BytesToTree(overrides)
+	if err != nil {
+		testutil.AnnotatedFatalf(t, "could not retrieve linkerd-config-overrides",
+			"could not retrieve linkerd-config-overrides\n%s", err.Error())
```
alpeb, line 712:
```suggestion
			"could not retrieve linkerd-config-overrides\n%s", err)
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+					strings.Join(tc.path, "/"), tc.value, finalValue)
+			}
+		})
+	}
+
+	panicIfError := func(t *testing.T, f func(...string) (string, error), path ...string) string {
+		val, err := f(path...)
+		if err != nil {
+			testutil.AnnotatedFatalf(t, err.Error(),
+				err.Error(), err.Error())
```
alpeb, line 769:
```suggestion
			testutil.AnnotatedFatalf(t, "error calling tree.GetString()", "error calling tree.GetString(): %s", err)
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+
+	t.Run("Check if any unknown fields sneaked in", func(t *testing.T) {
+		knownKeys := tree.Tree{
+			"controllerLogLevel": "debug",
+			"identity": map[string]interface{}{
+				"issuer": map[string]interface{}{},
+			},
+			"identityTrustAnchorsPEM": panicIfError(t, overridesTree.GetString, []string{"identityTrustAnchorsPEM"}...),
+			"proxyInit": map[string]interface{}{
+				"ignoreInboundPorts": "1234,5678",
```
alpeb, line 784:
```suggestion
				"ignoreInboundPorts": skippedInboundPorts,
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+			},
+		}
+
+		// Check for fields that were added during upgrade
+		if TestHelper.UpgradeFromVersion() != "" {
+			knownKeys["proxyInit"].(map[string]interface{})["ignoreOutboundPorts"] = "1234,5678"
```
alpeb, line 790:
```suggestion
			knownKeys["proxyInit"].(map[string]interface{})["ignoreOutboundPorts"] = skippedOutboundPorts
```

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+		return val
+	}
+
+	t.Run("Check if any unknown fields sneaked in", func(t *testing.T) {
+		knownKeys := tree.Tree{
+			"controllerLogLevel": "debug",
+			"identity": map[string]interface{}{
+				"issuer": map[string]interface{}{},
+			},
+			"identityTrustAnchorsPEM": panicIfError(t, overridesTree.GetString, []string{"identityTrustAnchorsPEM"}...),
```
alpeb, line 782: In this call and all the others below, you can pass a list of args instead of having to build a slice.

test/integration/install_test.go
```diff
@@ -679,6 +693,138 @@ func TestRetrieveUidPostUpgrade(t *testing.T) {
...
+
+			if tc.value != finalValue {
+				testutil.AnnotatedFatalf(t, fmt.Sprintf("Values at path %s do not match", strings.Join(tc.path, "/")),
+					"Expected value at [%s] to be [%s] but received [%s]",
+					strings.Join(tc.path, "/"), tc.value, finalValue)
+			}
+		})
+	}
+
+	panicIfError := func(t *testing.T, f func(...string) (string, error), path ...string) string {
```
alpeb, line 765: Given `f` will always be `overridesTree.GetString` you can just omit that param and call `overridesTree.GetString()` directly.

alpeb: Also, I think a better name for the function would be `extractValue` or something like that...

alpeb (commented):
This looks like it covers all the bases :+1: 
Just a bunch of nits below :grimacing:

alpeb (approved):
Thanks @Pothulapati , this looks good to me :+1:

PR #5807, opened 2021-02-23
stable-2.9.4 change notes

alpeb:
## stable-2.9.4

This stable release fixes an issue that prevented the proxy from being able to
speak HTTP/1 with older versioned proxies (announced in 2.9.3 but the fix wasn't
actually included).

* Fixed an issue that could cause the inbound proxy to fail meshed HTTP/1
  requests from older proxies (from the stable-2.8.x vintage)
* Fixed `linkerd install` command so that it can properly detect and avoid
  overwriting already installed linkerd instances from versions previous to 2.9
* Docker images are now hosted on the `cr.l5d.io` registry
* Updated base docker images to buster-20210208-slim

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,18 @@
 # Changes
 
+## stable-2.9.4
+
+This stable release fixes an issue that prevented the proxy from being able to
+speak HTTP/1 with older versioned proxies (announced in 2.9.3 but the fix wasn't
+actually included).
+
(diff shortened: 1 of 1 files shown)


adleong (approved):

kleimkuhler (approved):
👍

olix0r (approved):
Potentially worth calling out the Go 1.14.15 upgrade but also it's not really user-facing

PR #5810, opened 2021-02-23
Remove default skip ports and add to opaque ports

### LINKED ISSUE #5571: latest edge-21.1.2 still has default SSF ports set to be skipped in linkerd iptables
### What is the issue?
When testing out edge-21.1.2 as it supports server speaks first protocols it no longer makes sense to bypass common SSF ports in iptables by default ion installation . 

### How can it be reproduced?
mesh any pod in kubernetes using linkerd version 21.1.1 or higher , look at the linkerd-init pod logs 
```
-A PREROUTING -m comment --comment "proxy-init/install-proxy-init-prerouting/1611060320" -j PROXY_INIT_REDIRECT
-A OUTPUT -m comment --comment "proxy-init/install-proxy-init-output/1611060320" -j PROXY_INIT_OUTPUT
-A PROXY_INIT_OUTPUT ! -d 127.0.0.1/32 -o lo -m owner --uid-owner 2102 -m comment --comment "proxy-init/redirect-non-loopback-local-traffic/1611060320" -j PROXY_INIT_REDIRECT
-A PROXY_INIT_OUTPUT -m owner --uid-owner 2102 -m comment --comment "proxy-init/ignore-proxy-user-id/1611060320" -j RETURN
-A PROXY_INIT_OUTPUT -o lo -m comment --comment "proxy-init/ignore-loopback/1611060320" -j RETURN
-A PROXY_INIT_OUTPUT -p tcp -m multiport --dports 25,443,587,3306,11211 -m comment --comment "proxy-init/ignore-port-25,443,587,3306,11211/1611060320" -j RETURN
-A PROXY_INIT_OUTPUT -p tcp -m comment --comment "proxy-init/redirect-all-outgoing-to-proxy-port/1611060320" -j REDIRECT --to-ports 4140
-A PROXY_INIT_REDIRECT -p tcp -m multiport --dports 4190,4191,25,443,587,3306,11211 -m comment --comment "proxy-init/ignore-port-4190,4191,25,443,587,3306,11211/1611060320" -j RETURN
-A PROXY_INIT_REDIRECT -p tcp -m comment --comment "proxy-init/redirect-all-incoming-to-proxy-port/1611060320" -j REDIRECT --to-ports 4143
```
To get rid of the default port 3306 override so to test this new feature I set the annotations in my manifests to 
````
        config.linkerd.io/opaque-ports: "3306"
        config.linkerd.io/skip-inbound-ports: ""
        config.linkerd.io/skip-outbound-ports: ""
```
and then iptables looks like this 
```
-A PREROUTING -m comment --comment "proxy-init/install-proxy-init-prerouting/1611061110" -j PROXY_INIT_REDIRECT
-A OUTPUT -m comment --comment "proxy-init/install-proxy-init-output/1611061110" -j PROXY_INIT_OUTPUT
-A PROXY_INIT_OUTPUT ! -d 127.0.0.1/32 -o lo -m owner --uid-owner 2102 -m comment --comment "proxy-init/redirect-non-loopback-local-traffic/1611061110" -j PROXY_INIT_REDIRECT
-A PROXY_INIT_OUTPUT -m owner --uid-owner 2102 -m comment --comment "proxy-init/ignore-proxy-user-id/1611061110" -j RETURN
-A PROXY_INIT_OUTPUT -o lo -m comment --comment "proxy-init/ignore-loopback/1611061110" -j RETURN
-A PROXY_INIT_OUTPUT -p tcp -m comment --comment "proxy-init/redirect-all-outgoing-to-proxy-port/1611061110" -j REDIRECT --to-ports 4140
-A PROXY_INIT_REDIRECT -p tcp -m multiport --dports 4190,4191 -m comment --comment "proxy-init/ignore-port-4190,4191/1611061110" -j RETURN
-A PROXY_INIT_REDIRECT -p tcp -m comment --comment "proxy-init/redirect-all-incoming-to-proxy-port/1611061110" -j REDIRECT --to-ports 4143
```

### Environment

- Kubernetes Version:
-1.15
- GKE
- Linkerd version:
- edge-21.1.2

### Possible solution
Remove the bypassed ports in the default iptables

### LINKED ISSUE #5595: Add 5432 to default skip ports?
We've had a bunch of issues recently referencing port 5432 which is used by Amazon RDS and, I believe Postgres. If we do a stable-2.9.3, should we include this skip port? We'll still need to fold this into https://github.com/linkerd/linkerd2/issues/5571 for stable-2.10...

kleimkuhler:
This change removes the default ignored inbound and outbound ports from the
proxy init configuration.

These ports have been moved to the the `proxy.opaquePorts` configuration so that
by default, installations will proxy all traffic on these ports opaquely.

Closes #5571 
Closes #5595 


charts/linkerd2-cni/templates/cni-plugin.yaml
```diff
@@ -148,12 +148,16 @@ data:
...
+        {{- if .Values.ignoreInboundPorts }}
         "inbound-ports-to-ignore": [
           {{- include "partials.splitStringList" .Values.ignoreInboundPorts -}}
         ],
+        {{- end }}
+        {{- if .Values.ignoreOutboundPorts }}
         "outbound-ports-to-ignore": [
           {{- include "partials.splitStringList" .Values.ignoreOutboundPorts -}}
         ],
+        {{- end }}
```
alpeb, line 160: If `ignoreInboundPorts` and `ignoreOutboundPorts` are no longer used, why keeping this section?

alpeb (commented):
This is nicely simpler!

I think `_proxy-init.tpl` can be cleaned up by removing mentions to `Values.proxyInit.ignoreInboundPorts` and `Values.proxyInit.ignoreInboundPorts`, and removing the `--outbount-ports-to-ignore` section.

charts/linkerd2-cni/templates/cni-plugin.yaml
kleimkuhler: @alpeb My thought was with CNI installation if a user wants to specify these than we'd still need to populate those configurations—even thought that's not the default anymore.

alpeb: Ok, so that'd mean keeping `_proxy-init.tpl` as-is as well, otherwise why keeping those options just for CNI?

kleimkuhler: Yep so I'll keep it here as well as `_proxy-init.tpl` 👍

cli/cmd/testdata/install_controlplane_tracing_output.golden
```diff
@@ -1261,9 +1263,9 @@ spec:
         - --proxy-uid
         - "2102"
         - --inbound-ports-to-ignore
-        - "4190,4191,25,443,587,3306,11211"
+        - "4190,4191"
         - --outbound-ports-to-ignore
-        - "25,443,587,3306,11211"
+        - ",443"
```
alpeb, line 1268: The control plane deployments templates assume there is more than one port here so they prepend that unwanted comma:
```
initContainers:
       {{- if not (contains "443" ( .Values.proxyInit.ignoreOutboundPorts | toString ))}}
       {{- $_ := set $tree.Values.proxyInit "ignoreOutboundPorts" (printf "%s,443" ( .Values.proxyInit.ignoreOutboundPorts | toString )) -}}
       {{- end}}
```

adleong: could we do something like this?

```
{{- $outboundPorts := split "," .Values.proxyInit.ignoreOutboundPorts -}}
{{- $_ := set $tree.Values.proxyInit "ignoreOutboundPorts" (append $outboundPorts "443" | uniq | join ",") -}}
```

kleimkuhler: Something like that seems like a good way to solve this. There seems to be something wrong with the syntax, but hopefully can keep it close to that suggestion.

Pothulapati: I tried the following i.e without the syntax errors
```yaml
{{- $outboundPorts := splitList "," .Values.proxyInit.ignoreOutboundPorts -}}
{{- $_ := set $tree.Values.proxyInit "ignoreOutboundPorts" (append $outboundPorts "443" | uniq | join ",")}}
```

but still could not get it to remove `,` weirdly. I could narrow it down to `join` being the issue with it adding `,` when the list is only a single element (which is in our test cases)

We could instead fix this by doing
```yaml
      initContainers:
      {{- if not (contains "443" ( .Values.proxyInit.ignoreOutboundPorts | toString ))}}
      {{- $_ := set $tree.Values.proxyInit "ignoreOutboundPorts" ( ternary "443" (printf "%s,443" (.Values.proxyInit.ignoreOutboundPorts | toString )) (empty .Values.proxyInit.ignoreOutboundPorts)) -}}
      {{- end}}
```
This is def a bit complex but works.

Pothulapati:
>I think _proxy-init.tpl can be cleaned up by removing mentions to Values.proxyInit.ignoreInboundPorts and Values.proxyInit.ignoreInboundPorts, and removing the --outbount-ports-to-ignore section.

Can we do this? As that would mean we would loose all the skip ports functionality during install? I definitely don't see a use-case with skip ports except for the k8s healthchecks and UDP. Or Am I wrong in assuming that? 🤔

kleimkuhler:
> > I think _proxy-init.tpl can be cleaned up by removing mentions to Values.proxyInit.ignoreInboundPorts and Values.proxyInit.ignoreInboundPorts, and removing the --outbount-ports-to-ignore section.
> 
> Can we do this? As that would mean we would loose all the skip ports functionality during install? I definitely don't see a use-case with skip ports except for the k8s healthchecks and UDP. Or Am I wrong in assuming that? 🤔

@Pothulapati No I'm keeping those in the templates so that we keep that functionality.

Pothulapati (approved):
LGTM, except for the `,` nit

adleong (approved):

alpeb (approved):
LGTM :+1:

PR #5827, opened 2021-02-25
fix: issues affecting code quality

withshubh:
#### Description
This PR fixes a few issues that were affecting the code quality.

#### Summary of changes

- Remove unnecessary calls to fmt.Sprint
- Fix check for empty string
- Fix unnecessary calls to Printf
- Combine multiple `append`s into a single call


viz/tap/api/handlers.go
```diff
@@ -252,21 +252,7 @@ func handleAPIResourceList(w http.ResponseWriter, _ *http.Request, _ httprouter.
...
-				Verbs:      metav1.Verbs{"watch"},
-			})
-		resList.APIResources = append(resList.APIResources,
-			metav1.APIResource{
-				Name:       fmt.Sprintf("%s/tap", res.name),
-				Namespaced: res.namespaced,
-				Kind:       gvk.Kind,
-				Verbs:      metav1.Verbs{"watch"},
-			})
+		resList.APIResources = append(resList.APIResources, metav1.APIResource{Name: res.name, ShortNames: []string{res.shortname}, Namespaced: res.namespaced, Kind: gvk.Kind, Verbs: metav1.Verbs{"watch"}}, metav1.APIResource{Name: fmt.Sprintf("%s/tap", res.name), Namespaced: res.namespaced, Kind: gvk.Kind, Verbs: metav1.Verbs{"watch"}})
```
kleimkuhler, line 255: I find this harder to read than before so I'd prefer to keep separate `append`s here.

kleimkuhler (commented):
Thanks @withshubh! I'd like to keep one of these changes as-is, but everything else looks good.

We're working on the CI failures in a separate issue so this should pass once we fix the other issue.

viz/tap/api/handlers.go
withshubh: reverted :sparkles:

kleimkuhler:
@withshubh Great! Could you try merging in the latest `main`? I think the issue with CI failing should be resolved if you do that..

withshubh:
@kleimkuhler Done :sparkles:

kleimkuhler (approved):
Thanks @withshubh!

alpeb (approved):
LGTM :+1:

PR #5829, opened 2021-02-25
destination: pass opaque-ports through cmd flag

### LINKED ISSUE #5817: Pass in default set of opaque ports to destination server
After #5810 and and #5814 merge, we'll have two locations that track the default
set of opaque ports—`values.yaml` and `opaqueports/defaults.go`.

Instead of manually keeping these in sync, we should pass in the default set of
opaque ports as a flag to the destination server, and have the server use that
to return the default set of ports in destination profile responses.

Pothulapati:
Fixes #5817

Currently, Default opaque ports are stored at two places i.e
`Values.yaml` and also at `opaqueports/defaults.go`. As these
ports are used only in destination, We can instead pass these
values as a cmd flag for destination component from Values.yaml
and remove defaultPorts in `defaults.go`.

This means that if the `Values.yaml`'s opauePorts
field is overridden, That change is propagated both for injection and also
discovery like expected.


pkg/opaqueports/defaults.go
```diff
@@ -5,12 +5,4 @@ package opaqueports
...
-// Note: Keep in sync with proxy.opaquePorts in values.yaml
-var DefaultOpaquePorts = map[uint32]struct{}{
-	25:    {},
-	443:   {},
-	587:   {},
-	3306:  {},
-	5432:  {},
-	11211: {},
-}
+var DefaultOpaquePorts = map[uint32]struct{}{}
```
Pothulapati, line 8: Should we instead completely remove this pkg, and move this field as global variable in `controller`? 🤔 

I started out this way as it was the simple first step. Now trying to simplify this further.

kleimkuhler: Making it a global variable in `controller` was an issue when I first did this because we need to import it in `destination` and `watcher` packages which introduces an import cycle. I moved it into `pkg` package for this reason.

controller/cmd/destination/main.go
```diff
@@ -58,6 +64,17 @@ func Main(args []string) {
 		log.Warnf("expected cluster domain through args (falling back to %s)", *clusterDomain)
 	}
 
+	// Set opaqueports.DefaultOpaquePorts variiable
+	ports, err := parsePorts(*opaquePorts)
```
kleimkuhler, line 68: Could we figure out a way to use the parsing functions in `util`'s `parsing.go` here? We already have some separate functions for parsing opaque ports spread around that package as well as in `server.go`, and it feels unfortunate to introduce another place where we are parsing opaque ports.

controller/cmd/destination/main.go
```diff
@@ -30,6 +34,8 @@ func Main(args []string) {
 	enableEndpointSlices := cmd.Bool("enable-endpoint-slices", false, "Enable the usage of EndpointSlice informers and resources")
 	trustDomain := cmd.String("identity-trust-domain", "", "configures the name suffix used for identities")
 	clusterDomain := cmd.String("cluster-domain", "", "kubernetes cluster domain")
+	opaquePorts := cmd.String("opaque-ports", "", "configures the default opaque ports")
```
adleong, line 37: can we call this flag "default-opaque-ports"?

controller/cmd/destination/main.go
```diff
@@ -124,3 +141,35 @@ func Main(args []string) {
...
+			if err != nil {
+				return nil, err
+			}
+			opaquePorts = append(opaquePorts, uint32(port))
+		}
+	}
+	return opaquePorts, nil
+}
+
+func parsePortRanges(portsString string) []string {
```
adleong, line 159: Why do we return []string here?  Seems like we have ints and then convert them to strings, and then convert them immediately back into ints?

pkg/opaqueports/defaults.go
adleong: the cleaner thing to do would be to pass this as a constructor argument wherever it is needed instead of making it a global at all.

controller/cmd/destination/main.go
Pothulapati: Yeah, Unnecessary. Simplified this and also moved this into `util`

controller/cmd/destination/main.go
Pothulapati: Makes sense. Moved this into `util`

controller/cmd/destination/main.go
```diff
@@ -58,6 +62,17 @@ func Main(args []string) {
...
+	// Set opaqueports.DefaultOpaquePorts variiable
+	ports, err := util.ParsePorts(*opaquePorts)
+	if err != nil {
+		log.Fatalf("Failed to parse opaque Ports %s: %s", *opaquePorts, err)
+	}
+
+	for _, port := range ports {
+		opaqueports.DefaultOpaquePorts[port] = struct{}{}
+	}
+	log.Infof("Using default opaue ports: %v", ports)
```
alpeb, line 74:
```suggestion
	log.Infof("Using default opaque ports: %v", ports)
```

pkg/opaqueports/defaults.go
Pothulapati: Updated. This meant that `opaquePorts map[uint32]struct{}{}` had to be added to `server`, `endpoint_translator` and `opaque_ports_watcher` as each of them use this field induvidually

controller/cmd/destination/main.go
```diff
@@ -58,6 +61,19 @@ func Main(args []string) {
 		log.Warnf("expected cluster domain through args (falling back to %s)", *clusterDomain)
 	}
 
+	// Set opaqueports.DefaultOpaquePorts variiable
```
kleimkuhler, line 64: This comment can be updated since we don't have `DefaultOpaquePorts` anymore.

kleimkuhler (commented):
Overall this looks good. I wouldn't say it's a blocker, but what do you think about being more explicit in the endpoint translator, destination server, and opaque ports watcher this the field is the _default_ opaque ports—so changing the field name from `opaquePorts` to `defaultOpaquePorts`.

My concern is that in other places such as the opaque ports adaptor, we have an `opaquePorts` field that is updated as things change. This field being introduced though should never change after startup, and I think it would be helpful to make that obvious through the name.

pkg/util/parsing.go
```diff
@@ -10,6 +10,28 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+// ParsePorts parses the given ports string into a list of ports;
+// this includes converting port ranges into separate ports
+func ParsePorts(portsString string) ([]uint32, error) {
```
adleong, line 15: The only place that we use this function, we immediately convert the result from a `[]uint32` into a `map[uint32]struct{}`.  Should this function just return a `map[uint32]struct{}` to begin with?

adleong (approved):

kleimkuhler (approved):
Awesome thanks @Pothulapati for dealing with the naming nit 🙂. Looks good modulo Alex's comment.

pkg/util/parsing.go
Pothulapati: Makes sense. Updated.

PR #5835, opened 2021-02-26
docker: Avoid specifying TARGETARCH for await

olix0r:
When introducing the `linkerd-await` helper, we provided a default value
for `TARGETARCH`. This appears to interfere with multi-arch image
builds, causing ARM builds to fetch amd64 binaries.

Unsetting this default appears to fix this issue.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -2,8 +2,8 @@ ARG BUILDPLATFORM=linux/amd64
 
 FROM --platform=$BUILDPLATFORM docker.io/curlimages/curl:7.75.0  as await
 WORKDIR /tmp
-ARG TARGETARCH=amd64
 ARG LINKERD_AWAIT_VERSION=v0.2.1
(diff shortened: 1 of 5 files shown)


olix0r:
confirmed locally via
```
:; docker buildx build  --platform arm64 . -f controller/Dockerfile --load --progress plain --no-cache
...
#8 [await 3/3] RUN curl -fsSvLo linkerd-await https://github.com/olix0r/linkerd-await/releases/download/release%2Fv0.2.1/linkerd-await-v0.2.1-arm64 && chmod +x linkerd-await
```

we're at least fetching the correct binary now.. so I'd expect it to work

Pothulapati (approved):
LGTM, and makes sense as now we seems to use the global `TARGETARCH` and not overriding it.

kleimkuhler (approved):
Thanks this makes sense 👍

PR #5840, opened 2021-02-26
Fix Helm chart not setting the proxy-version annotation

alpeb:
In #5694 we set many Helm values to empty (like version tags), and then the
templates became responsible to filling out to proper default values. We missed
doing that for the `linkerd.io/proxy-version` annotation built in the
`_metadata.tpl` template. This fixes that, and also stops setting
`proxy.image.version` in `helm install|upgrade` in the Helm integration tests,
which is what avoided catching this error .

### DIFF
--- a/charts/linkerd2/templates/controller.yaml
+++ b/charts/linkerd2/templates/controller.yaml
@@ -72,7 +72,7 @@ spec:
     metadata:
       annotations:
         {{ include "partials.annotations.created-by" . }}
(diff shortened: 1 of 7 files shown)


kleimkuhler (approved):
Looks good!

kleimkuhler:
And thanks for fixing this in the tests. I was wondering how that got through.

Pothulapati (approved):
LGTM, Good Find.

PR #5841, opened 2021-02-27
prevent duplicate prometheus args

n-oden:
# prevent duplicate prometheus args in linkerd-viz/prometheus

# Problem

If a user specifies a `log.level` flag for linkerd-prometheus in `prometheus.args`, the template for linkerd-prometheus will generate a prometheus command where `--log.level` is specified twice (the first being directly interpolated from `prometheus.logLevel` in the template), and prometheus will crash-loop because its flags parser does not allow duplicate flags that are not arrays.

# Solution

Only fill in the `--log.level` flag from `.Values.prometheus.logLevel` if the `log.level` key is _not_ present in `.Values.prometheus.args`

# Validation

Added a test case.

### DIFF
--- a/viz/charts/linkerd-viz/templates/prometheus.yaml
+++ b/viz/charts/linkerd-viz/templates/prometheus.yaml
@@ -242,7 +242,9 @@ spec:
       {{- toYaml .Values.prometheus.sidecarContainers | trim | nindent 6 }}
       {{- end}}
       - args:
+        {{- if not (hasKey .Values.prometheus.args "log.level") }}
         - --log.level={{.Values.prometheus.logLevel | default .Values.defaultLogLevel}}
+        {{- end }}
         {{- range $key, $value := .Values.prometheus.args}}
         - --{{ $key }}{{ if $value }}={{ $value }}{{ end }}
         {{- end }}
--- a/viz/cmd/install_test.go
+++ b/viz/cmd/install_test.go
@@ -49,6 +49,15 @@ func TestRender(t *testing.T) {
 		},
 		{
 			map[string]interface{}{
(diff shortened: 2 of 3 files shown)


n-oden:
This is admittedly a corner case that view few people are likely to encounter...

...but guess who encountered it in the process of upgrading to the latest edge release in his testing environment? :)

n-oden:
(BTW while I am pleased with myself for figuring out a valid unit test for this, I think it may be a little over the top to further slow down what is already a staggeringly lengthy PR build/test cycle in order to test this one tiny bit of templating logic, so I'm fine removing the test if you'd prefer.)

n-oden:
(I am reaaaasonably certain that the test failure is not due to this PR.)

kleimkuhler:
@n-oden Is the base for this the most recent `main`? It may be worth making sure you have pulled the most recent `main` and merging it in to this branch, and then pushing again.

An issue like this should have been fixed rather recently, so I just want to make sure this branch has that fix.

n-oden:
rebased on current main/HEAD; we'll see what we see

n-oden:
@kleimkuhler the integration test is still failing the "linkerd check jaeger" step and I am definitely rebased off of main/HEAD 🤷

n-oden:
aha, re-rebasing fixed it.  good to go, I think?

kleimkuhler:
@n-oden Yes sorry about that. We have disabled that test for now while we figure out why it's failing.

kleimkuhler (approved):
Thanks looks good!

alpeb (approved):
LGTM, thanks @n-oden ! :+1:

PR #5847, opened 2021-03-01
jaeger: combine pod running checks

Pothulapati:
This PR combines the induvidual checks that check for each deployment
in running into a single check which checks for `running` status
for all the known deployments in the jaeger extension namespace.

This follows the same pattern as other extensions.


Pothulapati:
This was initially part of #5673, As it seems to take more time. Separated this change into this separate PR.

kleimkuhler (approved):
👍

alpeb (approved):

PR #5848, opened 2021-03-01
Move CP check after the readiness check

alpeb:
Moved the `can initialize client` and `can query the control plane API`
checks from the `linkerd-existence` section to the `linkerd-api` because
they required the `linkerd-controller` pod to not just be "Running" but
actually be ready.

This was causing `linkerd check` to show some port-forwarding warnings
when ran right after install.

This also allowed getting rid of the `CheckPublicAPIClientOrExit` function
and directly use `CheckPublicAPIClientOrRetryOrExit` (better naming
punted for later) which was refactored so it always runs the
`linkerd-api` checks before retrieving the client.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -106,7 +106,7 @@ jobs:
         - helm-upgrade
         - multicluster
         - uninstall
-        - upgrade-edge
+        #- upgrade-edge
         - upgrade-stable
         - cni-calico-deep
     needs: [docker_build]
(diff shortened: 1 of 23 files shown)


adleong (approved):
Moving these checks to the `linkerd-api` section makes logical sense too.  :100:

olix0r (approved):

alpeb:
Several additional fixes were required to have CI pass:

- Temporarily disabled `upgrade-edge` test because the latest edge has this readiness check issue
- Have the upgrade tests do proper prunning (stolen for @Pothulapati's #5673 :wink: )
- I was doing `kubectl wait` instead of `kubectl rollout status` for waiting for upgrades to complete
- Added missing label to tap SA (fixes #5850)
- Complete tap-injector Service selector

PR #5851, opened 2021-03-02
CLI: add `--opaque-ports` flag to `inject`

adleong:
Continuation of https://github.com/linkerd/linkerd2/pull/5721/

The `config.linkerd.io/opaque-ports` annotation can now be set using the `--opaque-ports` flag on `inject`

Example

```bash
$ linkerd inject /path/to/manifest.yaml --opaque-ports 3000,5000-6000,mysql
```

This annotation is the only one which is applied to services.


Pothulapati (approved):
LGTM, works as expected.

(Default Opaque Ports can be set using `--set opaquePorts` at install, With Helm override flags I don't think a separate CLI flag is need anymore.)

PR #5855, opened 2021-03-02
viz: support type name in tap request path

### LINKED ISSUE #5219: EOF when running linkerd tap command against a category of resources on a k3d cluster
### What is the issue?
Running the `linkerd tap` command against a control plane deployed to a k3d cluster results in an EOF error when targeting an entire category of kubernetes resources (e.g. Pod or Deployment).

The same command run against a specific resource (deploy/linkerd-prometheus, or a specific pod) works as expected.

### How can it be reproduced?
- Provision a cluster using [k3d](https://k3d.io/) (The number of nodes doesn't matter)
-- `k3d cluster create tap-test`

- Deploy the linkerd control plane
-- `linkerd install | kubectl apply -f -`

- Run tap against all resource types in the linkerd control plane
-- Pods: `linkerd tap -n linkerd po --verbose`
-- Deployments: `linkerd tap -n linkerd deploy --verbose`

### Logs, error output, etc
```
DEBU[0000] Response from [https://0.0.0.0:43867/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/pods//tap] had headers: map[Cache-Control:[no-cache, private] Content-Length:[205] Content-Type:[application/json] Date:[Thu, 12 Nov 2020 17:58:32 GMT]] 
DEBU[0000] Waiting for data...                          
error reading byte stream header: error while reading bytes from message: unexpected EOF
```

#### `linkerd check` output

```text
kubernetes-api
--------------
√ can initialize the client

kubernetes-version
------------------

linkerd-existence
-----------------
√ no unschedulable pods
√ controller pod is running
√ can initialize the client

linkerd-config
--------------

linkerd-identity
----------------
√ certificate config is valid

√ proxy-injector cert is valid for at least 60 days
√ sp-validator cert is valid for at least 60 days

linkerd-api
-----------
√ control plane self-check
√ tap api service is running

linkerd-version
---------------
√ cli is up-to-date

control-plane-version
---------------------
√ control plane is up-to-date

linkerd-prometheus
------------------
√ prometheus add-on service account exists
√ prometheus add-on config map exists
√ prometheus pod is running

linkerd-grafana
---------------
√ grafana pod is running

Status check results are √

```

### Environment

- Kubernetes Version: v1.18.9+k3s1
- Cluster Environment: k3d
- Host OS:  "Pop!_OS 20.04 LTS" (Ubuntu 20.04)
- Linkerd version: Found on `edge-20.11.1`, also happens with `stable-2.9` and `stable-2.8.1`

### Additional context
When targeting a specific deployment or pod, the command works and the output is displayed:

```bash
linkerd tap -n linkerd deploy/linkerd-prometheus
req id=0:0 proxy=in  src=10.42.0.1:47096 dst=10.42.0.17:9090 tls=no_tls_from_remote :method=GET :authority=10.42.0.17:9090 :path=/-/ready
rsp id=0:0 proxy=in  src=10.42.0.1:47096 dst=10.42.0.17:9090 tls=no_tls_from_remote :status=200 latency=833µs
end id=0:0 proxy=in  src=10.42.0.1:47096 dst=10.42.0.17:9090 tls=no_tls_from_remote duration=66µs response-length=21B
req id=0:1 proxy=out src=10.42.0.17:41954 dst=10.42.0.14:4191 tls=true :method=GET :authority=10.42.0.14:4191 :path=/metrics
rsp id=0:1 proxy=out src=10.42.0.17:41954 dst=10.42.0.14:4191 tls=true :status=200 latency=3033µs
end id=0:1 proxy=out src=10.42.0.17:41954 dst=10.42.0.14:4191 tls=true duration=260µs response-length=2628B
req id=0:2 proxy=out src=10.42.0.17:48764 dst=10.42.0.10:4191 tls=true :method=GET :authority=10.42.0.10:4191 :path=/metrics
rsp id=0:2 proxy=out src=10.42.0.17:48764 dst=10.42.0.10:4191 tls=true :status=200 latency=1847µs
end id=0:2 proxy=out src=10.42.0.17:48764 dst=10.42.0.10:4191 tls=true duration=46µs response-length=4002B
```

I've checked the logs from all the pods in the kube-system and linkerd and nothing is logged when the tap command fails.

kleimkuhler:
This changes tap requests to not have consecutive `/`s in the URL. This fixes
issues seen on k3d, and more generally creates a more explicit tap request URL.

Closes #5219

The failure  happens when there are consecutive `/`s in the URL. This occurs if
there is no specific resource specified—only the type.

For example when tapping deployments, the URL is:

```
/apis/tap.linkerd.io/v1alpha1/watch/namespaces/default/deployments//tap
```

After this change, the URL will be:

```
/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/type/pods/tap
```

And tapping a specific resource will be:

```
/apis/tap.linkerd.io/v1alpha1/watch/namespaces/linkerd/type/deployments/name/deploy-x/tap
```

### DIFF
--- a/viz/tap/api/handlers.go
+++ b/viz/tap/api/handlers.go
@@ -98,7 +98,7 @@ func initRouter(h *handler) *httprouter.Router {
 	return router
 }
 
-// createRoutes returns of list of routes that the API server should serve
+// createRoutes returns a list of routes that the API server should serve
 // depending on the type of resource given. If the resource type is anything
 // except namespace, two routes are returned: one for the resource type and
 // one for a specific resource of that type.
@@ -126,9 +126,7 @@ func (h *handler) handleTap(w http.ResponseWriter, req *http.Request, p httprout
 	path := strings.Split(req.URL.Path, "/")
 	if len(path) == 8 {
 		resource = path[5]
-	} else if len(path) == 10 {
-		resource = path[8]
-	} else if len(path) == 12 {
+	} else if len(path) == 10 || len(path) == 12 {
 		resource = path[8]
 	} else {
 		err := fmt.Errorf("invalid path: %s", req.URL.Path)
--- a/viz/tap/api/handlers.go
+++ b/viz/tap/api/handlers.go
@@ -116,8 +116,8 @@ func createRoutes(res resource) []string {
 }
 
 // POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/tap
-// POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/type/:type/tap
-// POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/type/:type/name/:name/tap
+// POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/type/{resource-type}/tap
+// POST /apis/tap.linkerd.io/v1alpha1/watch/namespaces/:namespace/type/{resource-type}/name/:name/tap
 func (h *handler) handleTap(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
 	namespace := p.ByName("namespace")
 	name := p.ByName("name")
--- a/viz/tap/api/handlers.go
+++ b/viz/tap/api/handlers.go
@@ -30,10 +30,17 @@ type handler struct {
 	log            *logrus.Entry
 }
 
-// TODO: share with api_handlers.go
-type jsonError struct {
-	Error string `json:"error"`
-}
+type (
+	// TODO: share with api_handlers.go
+	jsonError struct {
+		Error string `json:"error"`
+	}
+	resource struct {
+		name       string
+		shortname  string
+		namespaced bool
+	}
+)
 
 var (
 	gvk = &schema.GroupVersionKind{
@@ -53,11 +60,7 @@ var (
 		PreferredVersion: gvfd,
 	}
 
-	resources = []struct {
-		name       string
-		shortname  string
-		namespaced bool
-	}{
+	resources = []resource{
 		{"namespaces", "ns", false},
 		{"pods", "po", true},
 		{"replicationcontrollers", "rc", true},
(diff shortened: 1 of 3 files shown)


viz/tap/api/handlers.go
```diff
@@ -85,24 +88,36 @@ func initRouter(h *handler) *httprouter.Router {
...
 		}
-
-		router.GET(route, handleRoot)
-		router.POST(route+"/tap", h.handleTap)
 	}
-
 	return router
 }
 
+// createRoutes returns of list of routes that the API server should serve
```
alpeb, line 101:
```suggestion
// createRoutes returns a list of routes that the API server should serve
```

viz/tap/api/handlers.go
```diff
@@ -112,7 +127,9 @@ func (h *handler) handleTap(w http.ResponseWriter, req *http.Request, p httprout
 	if len(path) == 8 {
 		resource = path[5]
 	} else if len(path) == 10 {
-		resource = path[7]
+		resource = path[8]
+	} else if len(path) == 12 {
+		resource = path[8]
```
alpeb, line 132: IIUC, given `type` is now explicitly "labeled" as a query param, it might be cleaner to define `type := p.ByName("type")`. And then if `type` is empty have `resource = "namespaces"`, otherwise `resource = type`. Instead of using the `path` array.

kleimkuhler: Yep I think it'd be good to use the query param for this.

I don't think I'll change it to check if `type` is empty and assume `resource = "namespaces"`. I think it'd be good to keep the length check, and if it's `8` then `resources = "namespaces"`. And it's not `10` or `12`, then throw the error.

kleimkuhler: @alpeb I'm actually going to keep this as is. In order to use `p.ByName("type")`, the URL needs to have the `:type` parameter in the route (`typeRoute := fmt.Sprintf("/apis/%s/watch/namespaces/:namespace/type/%s", gvk.GroupVersion().String(), res.name)`).

Right now we know there routes only for the resource types we give, but if I change this to be `.../type/:type`, then we'd have to check in `handleTap` that it's a valid type. That seems like an unnecessary check to add to avoid just getting the index of the resource type.

alpeb (approved):
This tests fine for me :+1:

adleong (changes requested):
We need to understand how this interacts with the RBAC permission checking that Kubernetes does.  Since that is tied to the URL structure, I suspect that this breaks it.

kleimkuhler:
@adleong Yep I'll make sure to look into this more. I had it as something to return to since I don't think this is required for 2.10.

kleimkuhler:
Closing as I'm not actively working on this.

PR #5857, opened 2021-03-02
Helm integration tests improvements, and other flakiness fixes to CI

alpeb:
- Get rid of all the custom settings passed through `--set` during `helm install`, and instead let the defaults mechanisms in the templates to  kick-in
- Before installing `linkerd-viz` through Helm, wait on _all_ the core components to be ready (this is what might have been causing the restarts seen in CI)
- Show full output when `linkerd jaeger check` fails, and do some cleanup before triggering the tracing tests. But then I decided to temporarily disable that test till we figure out what's the deal.

### DIFF
--- a/test/integration/tracing/tracing_test.go
+++ b/test/integration/tracing/tracing_test.go
@@ -78,8 +78,7 @@ func TestTracing(t *testing.T) {
 	err = TestHelper.RetryFor(timeout, func() error {
 		out, err := TestHelper.LinkerdRun(checkCmd...)
 		if err != nil {
-			describeOut, _ := TestHelper.Kubectl("", "-n", "linkerd-jaeger", "describe", "po", "-l", "component=jaeger-injector")
-			return fmt.Errorf("'linkerd jaeger check' command failed\n%s\n%s\n%s", err, out, describeOut)
+			return fmt.Errorf("'linkerd jaeger check' command failed\n%s\n%s", err, out)
 		}
 		err = TestHelper.ValidateOutput(out, golden)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


olix0r (approved):

test/integration/tracing/tracing_test.go
```diff
@@ -42,12 +42,23 @@ func TestMain(m *testing.M) {
 //////////////////////
 
 func TestTracing(t *testing.T) {
+	t.Skip(`Temporarily skipped while we figure why it keeps on failing, only in CI ¯\_(ツ)_/¯`)
```
kleimkuhler, line 45: 👍

test/integration/tracing/tracing_test.go
```diff
@@ -42,12 +42,23 @@ func TestMain(m *testing.M) {
...
+	// cleanup cluster before proceeding
+	namespaces := []string{"smoke-test", "smoke-test-manual", "smoke-test-ann", "opaque-ports-test"}
+	for _, ns := range namespaces {
+		prefixedNs := TestHelper.GetTestNamespace(ns)
+		if err := TestHelper.DeleteNamespaceIfExists(ctx, prefixedNs); err != nil {
+			testutil.AnnotatedFatalf(t, "error deleting namespace",
+				"error deleting namespace '%s': %s", prefixedNs, err)
+		}
+	}
+
```
kleimkuhler, line 61: I don't think we used `WithDataPlaneNamespace` in some tests, so that may be way some of these stick around when they shouldn't.

kleimkuhler (approved):
Looks good!

test/integration/tracing/tracing_test.go
alpeb: Yea we show make good use of that. E.g. install_test.go should clean up all those smoke tests, but currently it's not a trivial change because other tests (stat, I think) assume they're around. Each test should build and clean up their own resources.

PR #5858, opened 2021-03-02
    Warn users about missing tap configuration in the dashboard

### LINKED ISSUE #5831: Linkerd dashboard shows tap error message if pods are not tapable
The dashboard shows an error message if pods are not tapable. It might be better to catch the error and display a friendlier message warning users that tap cannot be performed and also guide the user on how to rectify this.

image (2)

dadjeibaah:
This change modifies the dashboard so that it stops displaying an error
message when viewing tap streams from resources that don't have tap
enabled and instead, shows a warning and also guides the user on what to
do to fix it.

This change also includes updating the `<ExpansionPanel>` component
since we were getting deprecation warnings from Material-UI. The
component has been renamed to an accordion.

Fixes #5831

Screen Shot 2021-03-02 at 12 20 08 PM


web/app/js/locales/es/messages.json
```diff
@@ -155,6 +155,7 @@
   "tableTitleResponseInit": "Respuesta Inicial",
   "tableTitleTCP": "TCP",
   "tableTitleTCPMetrics": "Métricas TCP",
+  "Pods under the resource {resource} in the {namespace} namespace are missing tap configurations (restart these pods to enable tap)":"A los pods debajo del recurso {resource} en el espacio de nombres {namespace} les faltan configuraciones de toque (reinicie estos pods para habilitar el toque)",
```
alpeb, line 158: We're already using "Namespace" elsewhere in Spanish as well, because it's so prevalent and the literal translation might be confusing, IMO. (same thing regarding "tap").
```suggestion
  "Pods under the resource {resource} in the {namespace} namespace are missing tap configurations (restart these pods to enable tap)":"A los pods debajo del recurso {resource} en el namespace {namespace} les hace falta la configuración de tap (reinicie estos pods para habilitar tap)",
```

alpeb (approved):
Thanks @dadjeibaah , LGTM modulo translation nit :+1:

web/app/js/locales/es/messages.json
dadjeibaah: Great feedback! Thanks for the suggestion!

kleimkuhler (approved):
Works as expected!

PR #5861, opened 2021-03-03
Reenable disabled integration tests

alpeb:
Reenabled the `upgrade-test` and `cni-calico-deep` tracing integration tests. The former was disabled because in edge-21.2.4 the viz install wasn't waiting for the core components to be ready, and the latter was disabled because the `jaeger-webhook` image couldn't be loaded.

The reason simply was disk pressure, as evidenced by the k8s events. The fix consisted on adding a `--cleanup-docker` option to `bin/tests` which deletes the `image-archives` dir and the docker cache after the images have been loaded into k3d. That option is enacted only in the `integration_tests.yml` workflow in the deep tests, which are the ones pulling the most images.

The `release.yml` workflow doesn't require it since it pulls the images directly from the public registry.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -148,5 +148,11 @@ jobs:
         # Validate the CLI version matches the current build tag.
         [[ "$TAG" == "$($HOME/.linkerd version --short --client)" ]]
     - name: Run integration tests
+      if: !endsWith(matrix.integration_test, 'deep')
       run: |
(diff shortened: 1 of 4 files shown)


alpeb:
The `cni-calico-deep` issue appears to be fixed now; I updated the PR description with more details.

Pothulapati (approved):
LGTM, 👍🏼 on making this focused to the deep integration tests.

kleimkuhler (approved):
Looks good @alpeb; good find!

PR #5872, opened 2021-03-05
Extra privs required for CR linkerd-linkerd-viz-tap-admin

alpeb:
@dadjeibaah [pointed out](https://github.com/linkerd/website/pull/972#discussion_r588584786) while reviewing the [Enabling Tap access](https://linkerd.io/2/tasks/securing-your-cluster/#enabling-tap-access) doc that granting the `linkerd-linkerd-viz-tap` ClusterRole to a user is no longer enough for tapping, as namespace listing is now also required:

```console
$ linkerd viz tap -n linkerd deploy/linkerd-controller --as $(whoami)

Cannot connect to Linkerd Viz: namespaces is forbidden: User "dennis"   cannot list resource "namespaces" in API group "" at the cluster scope
Validate the install with: linkerd viz check
```

Adding the Namespace List privilege to that CR solves the issue.

### DIFF
--- a/viz/charts/linkerd-viz/templates/tap-rbac.yaml
+++ b/viz/charts/linkerd-viz/templates/tap-rbac.yaml
@@ -28,6 +28,9 @@ metadata:
     linkerd.io/extension: viz
     component: tap
 rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list"]
 - apiGroups: ["tap.linkerd.io"]
   resources: ["*"]
   verbs: ["watch"]
--- a/viz/cmd/testdata/install_default.golden
+++ b/viz/cmd/testdata/install_default.golden
@@ -146,6 +146,9 @@ metadata:
     linkerd.io/extension: viz
     component: tap
 rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list"]
 - apiGroups: ["tap.linkerd.io"]
   resources: ["*"]
   verbs: ["watch"]
@@ -948,7 +951,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: 7f24ffe166f8d03a8eb039ecc551f3de9ea0ddcf1cf80a69966ce3713f043b23
+        checksum/config: 2ab0ec921e7f9cb65f62dd4f2ebb77d06827c9621f473c3a9c42ad2741603c3e
         linkerd.io/created-by: linkerd/helm dev-undefined
       labels:
         linkerd.io/extension: viz
--- a/viz/cmd/testdata/install_default_overrides.golden
+++ b/viz/cmd/testdata/install_default_overrides.golden
@@ -146,6 +146,9 @@ metadata:
     linkerd.io/extension: viz
     component: tap
 rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list"]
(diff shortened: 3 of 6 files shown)


dadjeibaah (commented):
I still run into the issue when I use a different name for the `--as` flag:

```
$ go run cli/main.go viz tap -n linkerd deploy/linkerd-controller --as jimmy
Cannot connect to Linkerd Viz: namespaces is forbidden: User "jimmy" cannot list resource "namespaces" in API group "" at the cluster scope: RBAC: clusterrole.rbac.authorization.k8s.io "linkerd-linkerd-tap-admin" not found
Validate the install with: linkerd viz check
exit status 1
```
It seems like the k8s client in the viz Healthchecker is the root cause. If you modify `InitializeKubeAPIClient()` to not use the impersonated user to initialize the client, The tap request flows through and gets evaluated by the tap server and throws a 403 error.

```
go run cli/main.go viz tap -n linkerd deploy/linkerd-controller --as jimmy
HTTP error, status Code [403] (deployments.tap.linkerd.io "linkerd-controller" is forbidden: User "jimmy" cannot watch resource "deployments/tap" in API group "tap.linkerd.io" in the namespace "linkerd")exit status 1
```
I'm unclear on whether the k8s API clients needs to have the impersonated user in its initialization.

alpeb:
@dadjeibaah I'm not experiencing the same issue. This is my full list of commands, using the current branch:
```console
$ DOCKER_BUILDKIT=1 bin/docker-build
$ bin/image-load --k3d
$ bin/linkerd install|k apply -f -
$ bin/linkerd viz install|k apply -f -

# inject tap in linkerd-controller
$ k -n linkerd rollout restart deploy/linkerd-controller

$ k create clusterrolebinding   $(whoami)-tap-admin   --clusterrole=linkerd-linkerd-viz-tap-admin --user=$(whoami)
$ bin/linkerd viz tap -n linkerd deploy/linkerd-controller --as $(whoami)
req id=2:0 proxy=in  src=10.42.0.15:57782 dst=10.42.0.19:9995 tls=true :method=GET :authority=10.42.0.19:9995 :path=/metrics
rsp id=2:0 proxy=in  src=10.42.0.15:57782 dst=10.42.0.19:9995 tls=true :status=200 latency=2218µs
...
```
Can you provide a repro for your issue? Also please note the ClusteRole name has changed with respect to the old 2.9 docs: linkerd-linkerd-viz-tap-admin

alpeb:
@dadjeibaah I agree that, strictly speaking, the set of privileges required by the healtchecker and the tap API are not the same. Besides the `tap.linkerd.io` watch privilege, the former also requires the Namespace List privilege so that it can verify that the viz namespace exists, which is not really required for the operation of tap. Nevertheless, I think it's important for commands (like `linkerd viz tap`) and their check counterpart (like `linkerd viz check`) to operate under the same privileges so that `check` is able to point to problems caused by some lack of privilege that `linkerd viz tap ` might have. Let me know if this makes sense :-)

dadjeibaah:
@alpeb, yeah, that makes sense to me. i think what confused me was that the initial sentence in the error message made it seem like there was an issue with the `viz install` itself and not a permissions issue. I agree that all subcommands must have consistent privileges so I'm cool with shipping this. Thanks for clarifying! 🙏

dadjeibaah (approved):
LGTM! 📦 🚢

kleimkuhler (approved):
👍

PR #5869, opened 2021-03-04
Update BUILD.md

alpeb:
Fixes #5063

- Completed table of contents.
- Added references to viz and other extensions there where appropriate.
- Updated components chart.
- Updated build architecture chart.
- Moved `go test` paragraphs into `TEST.md`, and moved from `TEST.md` the linting paragraph.
- Group Helm paragraphs together.
- Deleted the "Go modules and dependencies" paragraph. We can assume most people are already familiar with go modules.

PR #5875, opened 2021-03-08
Only support docker buildx

alpeb:
With #5835 adding `linkerd-await` in most Dockerfiles we require
`TARGETARCH` to be set by the system, which is only available in the
BuildKit backend. Thus we drop support for plain vanilla docker and
remove all references to `DOCKER_BUILDKIT`. The developer docs are being
updated accordingly in #5869


kleimkuhler (approved):
👍

Pothulapati (approved):
LGTM

PR #5874, opened 2021-03-08
Add back ignore ports fields in values.yaml

kleimkuhler:
Add back configurations to `values.yaml` files after being removed in #5810

Signed-off-by: Kevin Leimkuhler <kevin@kleimkuhler.com>

PR #5876, opened 2021-03-08
Fix parsing of ignoreInboundPorts field

kleimkuhler:
The `ignoreInboundPorts` field is not parsed correctly when passed through the
`--set` Helm flag. This was discovered in https://github.com/linkerd/linkerd2/pull/5874#pullrequestreview-606779599.

This is happening because the value is not parsed into a string before using it
in the templating.

Before:

```
linkerd install --set proxyInit.ignoreInboundPorts="12345" |grep 12345
...
- "4190,4191,%!s(int64=12345)"
...
```

After:

```
linkerd install --set proxyInit.ignoreInboundPorts="12345" |grep 12345
...
- "4190,4191,12345"
...
```

### DIFF
--- a/charts/partials/templates/_proxy-init.tpl
+++ b/charts/partials/templates/_proxy-init.tpl
@@ -7,7 +7,7 @@ args:
 - --proxy-uid
 - {{.Values.proxy.uid | quote}}
 - --inbound-ports-to-ignore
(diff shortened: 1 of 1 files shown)


Pothulapati (approved):
LGTM, :shipit:

olix0r (approved):

dadjeibaah (approved):
📦 🚢

PR #5891, opened 2021-03-11
Fix helm-upgrade and upgrade-stable integration tests

alpeb:
Update `install_test.go` now that the upgrade test is done from 2.10.
This also implied installing viz right after core.

Refactored `HelmInstallPlain()` into `HelmCmdPlain()` to work for both
helm install and upgrade.


olix0r (approved):

dadjeibaah (approved):
Thanks for fixing this! 📦 🚢

Issue #4870, opened 2020-08-12
Intermittent 502 Bad Gateway issue when service is meshed


mbelang:
### What is the issue?

Without linkerd proxy, there are no trace of 502 bad gateway at the ingress level nor the app level
With linkerd proxy enabled on nginx ingress and or app, intermittent 502 Bad Gateway appears in the system.

I see 2 types of error from the proxy:
```
## for this one the requests actually made it through the upstream
linkerd2_app_core::errors: Failed to proxy request: connection closed before message completed
```
and
```
## for this one the requests never made it to the upstream
linkerd2_app_core::errors: Failed to proxy request: connection error: Connection reset by peer (os error 104)
```
both lead to (on ingress)
```
"GET /v1/users?ids=xxxxxxx,xxxxx HTTP/1.1" 502 0 "-" "python-requests/2.23.0" 1708 0.097 [xxxxx-app-http] [] 10.4.27.106:80 0 0.100 502 7543fb48ce22d5c6145b97daadde93d9
```
or (on an app)
```
## Here I suspect the connection timeout that could be tuned to allow more time to the proxy to establish the connection. I manage to prove that by giving more resources to datadog agent daemonset

Failed to send traces to Datadog Agent at http://10.4.87.60:8126: HTTP error status 502, reason Bad Gateway, message content-length: 0
date: Wed, 12 Aug 2020 17:58:22 GMT
```

I have eliminated the scale-up and scale-down event of ingress and app. I also ruled out the graceful termination which I gave a period of 30s on app and 40s on ingress for the sleep on the proxy.

### Logs, error output, etc

here are some linkerd-debug logs where 10.3.45.50 is the ingress controller pod and 10.3.23.252 is the upstream server pod
```text
1829200 2893.254017273    127.0.0.1 → 127.0.0.1    TCP 68 443 → 57590 [ACK] Seq=3277 Ack=1894 Win=175744 Len=0 TSval=378310645 TSecr=378310644
1829201 2893.254172427   10.3.45.50 → 127.0.0.1    HTTP 1701 GET /v1/users/xxxxxxxx?include=active_contract HTTP/1.1 
1829202 2893.254471803   10.3.45.50 → 10.3.23.252  TCP 76 47414 → 80 [SYN] Seq=0 Win=26883 Len=0 MSS=8961 SACK_PERM=1 TSval=2090260178 TSecr=0 WS=128
[...]

PR #5893, opened 2021-03-12
dest: Reduce possible response size in destination service errors

kleimkuhler:
This reduces the possible HTTP response size from the destination service when
it encounters an error during a profile lookup.

If multiple objects on a cluster share the same IP (such as pods in
`kube-system`), the destination service will return an error with the two
conflicting pod yamls.

In certain cases, these pod yamls can be too large for the HTTP response and the
destination pod's proxy will indicate that with the following error:

```
hyper::proto::h2::server: send response error: user error: header too big
```

From the app pod's proxy, this results in the following error:

```
poll_profile: linkerd_service_profiles::client: Could not fetch profile error=status: Unknown, message: "http2 error: protocol error: unexpected internal error encountered"
```

We now only return the conflicting pods (or services) names. This reduces the
size of the returned error and fixes these warnings from occurring.

Example response error:

```
poll_profile: linkerd_service_profiles::client: Could not fetch profile error=status: FailedPrecondition, message: "Pod IP address conflict: kube-system/kindnet-wsflq, kube-system/kube-scheduler-kind-control-plane", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 12 Mar 2021 19:54:09 GMT"} }
```

### DIFF
--- a/controller/api/destination/watcher/ip_watcher.go
+++ b/controller/api/destination/watcher/ip_watcher.go
@@ -123,61 +123,49 @@ func (iw *IPWatcher) Unsubscribe(clusterIP string, port Port, listener EndpointU
 // GetSvcID returns the service that corresponds to a Cluster IP address if one
 // exists.
 func (iw *IPWatcher) GetSvcID(clusterIP string) (*ServiceID, error) {
-	resource, err := getResource(clusterIP, iw.k8sAPI.Svc().Informer())
+	objs, err := iw.k8sAPI.Svc().Informer().GetIndexer().ByIndex(podIPIndex, clusterIP)
 	if err != nil {
-		return nil, err
+		return nil, status.Error(codes.Unknown, err.Error())
 	}
-	if svc, ok := resource.(*corev1.Service); ok {
-		service := &ServiceID{
-			Namespace: svc.Namespace,
-			Name:      svc.Name,
-		}
-		return service, nil
+	services := make([]*corev1.Service, 0)
+	for _, obj := range objs {
+		service := obj.(*corev1.Service)
+		services = append(services, service)
 	}
-	// Either `clusterIP` does not map to a service that the indexer is aware
-	// of, or it maps to a resource that is ignored
-	return nil, nil
+	if len(services) > 1 {
+		return nil, status.Errorf(codes.FailedPrecondition, "Cluster IP address conflict: %s/%s, %s/%s", services[0].Namespace, services[0].Name, services[1].Namespace, services[1].Name)
+	}
+	if len(services) == 0 {
+		return nil, nil
+	}
+	service := &ServiceID{
+		Namespace: services[0].Namespace,
+		Name:      services[0].Name,
+	}
+	return service, nil
 }
 
 // GetPod returns the pod that corresponds to an IP address if one exists.
 func (iw *IPWatcher) GetPod(podIP string) (*corev1.Pod, error) {
-	resource, err := getResource(podIP, iw.k8sAPI.Pod().Informer())
-	if err != nil {
-		return nil, err
-	}
-	if pod, ok := resource.(*corev1.Pod); ok {
-		return pod, nil
-	}
-	// Either `podIP` does not map to a pod that the indexer is aware of, or
-	// it maps to a resource that is ignored
-	return nil, nil
-}
-
-func getResource(ip string, informer cache.SharedIndexInformer) (interface{}, error) {
-	matchingObjs, err := informer.GetIndexer().ByIndex(podIPIndex, ip)
+	objs, err := iw.k8sAPI.Pod().Informer().GetIndexer().ByIndex(podIPIndex, podIP)
 	if err != nil {
 		return nil, status.Error(codes.Unknown, err.Error())
 	}
-
-	objs := make([]interface{}, 0)
-	for _, obj := range matchingObjs {
-		// Ignore terminated pods,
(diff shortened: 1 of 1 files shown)


olix0r (approved):
at first I was surprised that this change was so large but now I see it's so that the errors can include the service/pod name... This seems fine, but I think it would also be okay to omit the pod/service names and instead only return the IP that has duplicate entries.

kleimkuhler:
@olix0r Yes it would have been nice if I could have made `getResources` work with either resource type, but unfortunately not that easy 😬. I could remove the printing of `namespace/name` to make the diff smaller, but I like that we still print those.

Also, they aren't handled exactly the same anyways since we don't need to do a `podTerminated` check for services. Additionally, the follow-up change for this which checks `HostPort` will be another thing that makes the pod check different.

adleong:
Can you explain why `getResources` can't be made to work with either resource type?  I would expect that for each of `obj[0]` and `obj[1]`, in order to print it you could use a type switch to cast to `*corev1.Pod` or `*corev1.Service` as appropriate.

kleimkuhler:
@adleong In order to print the pod or service name, we would need a switch statement to get the name and namespace of the object. Something like:

```
var name, namespace string
switch res := obj.(type) {
case *corev1.Pod:
	name = res.Name
	namespace = res.Namespace
case *corev1.Service:
	name = res.Name
	namespace = res.Namespace
}
```

I originally went this route, but `getResource` already has specific logic for pods—the `podTerminated` function. Additionally, the follow-up change for this where we handle `hostPort` looks up nodes is also specific to just pods.

Yes it could still work in `getResource`, but I don't think continuing to add pod specific logic and switch statements to deal with printing resources to a function that is generic over the object really felt like the right path. It seemed like it was the right time to split it up.

controller/api/destination/watcher/ip_watcher.go
```diff
@@ -123,61 +123,49 @@ func (iw *IPWatcher) Unsubscribe(clusterIP string, port Port, listener EndpointU
...
-	return nil, nil
-}
-
-// GetPod returns the pod that corresponds to an IP address if one exists.
-func (iw *IPWatcher) GetPod(podIP string) (*corev1.Pod, error) {
-	resource, err := getResource(podIP, iw.k8sAPI.Pod().Informer())
-	if err != nil {
-		return nil, err
+	if len(services) > 1 {
+		return nil, status.Errorf(codes.FailedPrecondition, "Cluster IP address conflict: %s/%s, %s/%s", services[0].Namespace, services[0].Name, services[1].Namespace, services[1].Name)
```
Pothulapati, line 136: How do we know there are only 2 always? Why not `string,Join` all of them and use that string 🤔

Pothulapati (commented):
LGTM, Overall. After this update, https://github.com/linkerd/linkerd2/issues/4870 fails with the following logs:

```bash
[   144.840686s]  WARN ThreadId(01) outbound:accept{client.addr=10.244.0.19:49206 target.addr=172.18.0.2:8126}: linkerd_service_profiles::client: Could not fetch profile error=status: FailedPrecondition, message: "Pod IP address conflict: kube-system/kube-scheduler-kind-control-plane, kube-system/kube-proxy-26zgc", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 15 Mar 2021 08:32:52 GMT"} }
[   144.841287s]  WARN ThreadId(01) outbound:accept{client.addr=10.244.0.19:49206 target.addr=172.18.0.2:8126}: linkerd_service_profiles::client: Could not fetch profile error=status: FailedPrecondition, message: "Pod IP address conflict: kube-system/kube-proxy-26zgc, kube-system/kindnet-ftfck", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 15 Mar 2021 08:32:52 GMT"} }
```

controller/api/destination/watcher/ip_watcher.go
kleimkuhler: There could be more than two in this case! I just kept this error as is though, and it has only ever printed the first two conflicting pods.

It could be changed, but it's not related to this fix so I'd rather keep it as is.

Pothulapati: Yeah, With the current hostIP issue, There can be more than two in most cases.

If we are going to print those names, My suggestion would be to print them all unless if there is a reason not to. :)

kleimkuhler: I'll update this error message to provide a little more description. I don't want to list all pods because in an extreme case we could end up hitting the message limit again. I will however change it to include the number of conflicting pods:

```
Cluster IP address conflict between %d pods; first two: ...
```

That way a user can see they have for example 10 conflicting pods, and the first two give them a starting place to look.

kleimkuhler:
Error updated to include the number of conflicting pods or services:

```
Could not fetch profile error=status: FailedPrecondition, message: "found 6 pods with conflicting pod IP 172.18.0.2; first two: kube-system/kube-proxy-wr4xw, kube-system/kube-controller-manager-kind-control-plane", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 15 Mar 2021 15:52:01 GMT"} }
```

Pothulapati (approved):
LGTM!

adleong (approved):
Looks good, thanks for explaining @kleimkuhler!

PR #5900, opened 2021-03-15
cli: update err msg when control-plane already exists.

### LINKED ISSUE #5889: cli: Improve `install` error message when control plane already exists
```
:; linkerd install | k apply -f -
Can't install the Linkerd control plane in the 'linkerd' namespace. Reason: ConfigMap/linkerd-config already exists.
If this is expected, use the --ignore-cluster flag to continue the installation.
```

This error suggests that users run with `--ignore-cluster` instead of using `upgrade`... We should primarily encourage users to use `upgrade` here. I don't think we should talk about `--ignore-cluster` in this error message at all.

Pothulapati:
Fixes #5889

Currently, The error msg that is sent when control-plane already
exists suggests to use `--ignore-cluster`. We should instead
suggest users to use `linkerd upgrade`.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -54,7 +54,7 @@ If you are sure you'd like to have a fresh install, remove these resources with:
 Otherwise, you can use the --ignore-cluster flag to overwrite the existing global resources.
 `
 
-	errMsgLinkerdConfigResourceConflict = "Can't install the Linkerd control plane in the '%s' namespace. Reason: %s.\nRun `linkerd upgrade` with the relevant binary, if you are looking to upgrade Linkerd.\n"
+	errMsgLinkerdConfigResourceConflict = "Can't install the Linkerd control plane in the '%s' namespace. Reason: %s.\nRun the command `linkerd upgrade`, if you are looking to upgrade Linkerd.\n"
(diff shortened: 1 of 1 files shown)


cli/cmd/install.go
```diff
@@ -54,7 +54,7 @@ If you are sure you'd like to have a fresh install, remove these resources with:
 Otherwise, you can use the --ignore-cluster flag to overwrite the existing global resources.
 `
 
-	errMsgLinkerdConfigResourceConflict = "Can't install the Linkerd control plane in the '%s' namespace. Reason: %s.\nIf this is expected, use the --ignore-cluster flag to continue the installation.\n"
+	errMsgLinkerdConfigResourceConflict = "Can't install the Linkerd control plane in the '%s' namespace. Reason: %s.\nRun `linkerd upgrade` with the relevant binary, if you are looking to upgrade Linkerd.\n"
```
dadjeibaah, line 57: Do we need to make a reference to the "relevant binary"? Could  we say something like "run the command `linkerd upgrade` ?

dadjeibaah (approved):
LGTM. Just had one comment 📦 🚢

cli/cmd/install.go
kleimkuhler: I agree. Just saying to run the `upgrade` command should be good.

kleimkuhler (approved):

adleong (approved):

